PR #2592, opened 2026-06-24
[EP]:Stabilize MACA Expert Parallelism P2P Fast Path

### LINKED ISSUE #1965: [Feature Request]: MACA Support for Mooncake EP/PG
## Summary
Mooncake EP/PG Python extensions are currently CUDA-only. We need MACA support via `cu-bridge` without modifying existing source files.

## Key Finding
**MUSA and HIP do NOT have EP/PG support either** — zero references to `USE_MUSA`/`USE_HIP` in `mooncake-ep/` or `mooncake-pg/`. This adaptation would be breaking new ground.

## Blockers

| Issue | Why |
|---|---|
| Fabric constants missing | `CU_MEM_HANDLE_TYPE_FABRIC`, `CU_DEVICE_ATTRIBUTE_HANDLE_TYPE_FABRIC_SUPPORTED` undefined in `maca.h` and cu-bridge |
| `maca.h` / cu-bridge conflict | `#define cuDeviceGetAttribute mcDeviceGetAttribute` collides with cu-bridge's `wcuDeviceGetAttribute` |
| EP kernel uses PTX asm | `cooperative_groups`, `cuda::atomic`, `__byte_perm`, `clock64()` cannot be translated by any header bridge |
| Setup.py is standalone | `CUDAExtension` does not inherit CMake config; MACA paths/libs must be injected separately |

## Proposed Design

**Principle: Zero modifications to existing `.cpp`/`.cu`/`.h` files.**

### 1. `gpu_vendor/maca.h` — Add fabric stubs
```cpp
static const mcMemAllocationHandleType CU_MEM_HANDLE_TYPE_FABRIC = mcMemHandleTypeNone;
static const mcDrvDeviceAttribute_t CU_DEVICE_ATTRIBUTE_HANDLE_TYPE_FABRIC_SUPPORTED =
    static_cast<mcDrvDeviceAttribute_t>(0x7FFFFFFF);
#define CUmemFabricHandle void*
```
Remove `#define cuDeviceGetAttribute mcDeviceGetAttribute` (let cu-bridge handle it).

### 2. PG — Header mapping only, no source changes
PG uses standard CUDA Runtime APIs. After `maca.h` is fixed, it should compile as-is.

### 3. EP Kernel — File-level separation (Ascend-style)
Do **not** add `#ifdef USE_MACA` to `mooncake_ep_kernel.cu`.

Create independent directory:
```
maca-ep/                          # or mooncake-ep-maca/
├── setup.py
└── src/kernel/
    └── mooncake_ep_kernel_maca.cpp   # stub: throw runtime_error
```

`maca-ep/setup.py` references unchanged files from `../mooncake-ep/src/` and swaps only the kernel source.

### 4. Build system selection
`BuildEpExt.cmake` chooses directory based on `USE_MACA`:
```cmake
if(USE_MACA)
  set(EP_SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/maca-ep")
else()
  set(EP_SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/mooncake-ep")
endif()
```

## Open Questions
1. Directory naming: `maca-ep/` vs `mooncake-ep-maca/`?
2. PG setup.py: separate `maca-pg/` directory, or inject via env vars?
3. `-include cuda_alike.h` via compiler flags (zero source changes) vs explicit `#include` in headers (one-line change, aligns with transfer-engine convention)?
4. `engine.so` is empty; `libtransfer_engine.so` + `libasio.so` need proper CMake install rules.

## Acceptance Criteria
- [ ] `maca.h` contains fabric memory stubs
- [ ] `mooncake-ep/` and `mooncake-pg/` source files remain untouched
- [ ] MACA EP compiles & loads (kernel throws `runtime_error` at call time — expected)
- [ ] MACA PG compiles & loads (functional)
- [ ] Zero `#ifdef USE_MACA` in any existing source file

- [ ] Make sure you already searched for relevant issues and read the [documentation](https://kvcache-ai.github.io/Mooncake/)

### LINKED ISSUE #1975: [RFC]: Mooncake EP Multi-Vendor GPU Adaptation Design
## Problem

`mooncake-ep` kernels are currently CUDA-only and tightly coupled with **NVIDIA PTX inline assembly**. This blocks porting to MACA (Metax), MUSA, and HIP.

Compilation on Metax (`torch 2.8.0+metax3.5.3.9`, `cucc` compiler) fails with:

```
error: invalid input constraint 'l' in asm
error: invalid output constraint '=h' in asm
```

Root cause: ~56 PTX occurrences across `mooncake_ep_utils.cuh` (~48) and `mooncake_ep_kernel.cu` (6 direct embeddings). PTX is NVIDIA ISA-specific; Metax `mxcc` cannot translate it, even though Metax's `cu-bridge` provides CUDA Runtime API compatibility.

## Design Goal

- **Keep one shared kernel implementation.** Do not fork 700+ lines of `dispatch`/`combine` logic per vendor.
- **Isolate vendor differences at the primitive layer only.**
- **Preserve performance.** Each primitive can be optimized independently without touching high-level kernel code.
- **Live inside `mooncake-ep/`** and be extensible to MUSA/HIP.

### Directory Layout

```
mooncake-ep/
├── include/                              # Existing headers (untouched)
│   └── mooncake_ep_utils.cuh             # Refactored: non-PTX helpers only
├── src/
│   └── mooncake_ep_kernel.cu             # Minimal refactor: 6 PTX → macros
└── platform/                             # NEW: vendor-specific device primitives
    ├── cuda/
    │   └── ep_primitives.cuh             # All existing PTX functions
    ├── maca/
    │   └── ep_primitives.cuh             # MACA: std CUDA C++ / Metax intrinsics
    ├── musa/                             # RESERVED
    └── hip/                              # RESERVED
```

### Key Idea

PTX primitives are **already encapsulated** as inline functions in `mooncake_ep_utils.cuh` (`ld_acquire_sys_global`, `st_release_sys_global`, `atomic_add_release_sys_global`, `memory_fence_*`, `bar.sync`, etc.).

We extract them into `platform/<vendor>/ep_primitives.cuh` with **identical signatures** but vendor-compatible bodies. The build system injects `-I platform/<vendor>` so the same `mooncake_ep_kernel.cu` compiles for all vendors.

### The Only Required Change to Existing Code

Extract 6 direct `asm volatile` calls in `mooncake_ep_kernel.cu` into macros:

```cpp
// BEFORE
asm volatile("bar.sync 1, %0;" :: "r"(num_threads));

// AFTER
EP_BARRIER(1, num_threads);
```

`platform/cuda/ep_primitives.cuh`:
```cpp
#define EP_BARRIER(id, n) \
    asm volatile("bar.sync %0, %1;" :: "r"(id), "r"(n))
```

`platform/maca/ep_primitives.cuh`:
```cpp
#define EP_BARRIER(id, n) __syncthreads()
```

### Performance Preservation

| Primitive | CUDA (PTX) | MACA Strategy |
|-----------|-----------|---------------|
| `ld_acquire_sys_global` | `ld.acquire.sys.global.s32` | `__threadfence_system(); return *ptr;` |
| `st_release_sys_global` | `st.release.sys.global.s32` | `*ptr = val; __threadfence_system();` |
| `atomic_add_release_sys_global` | `atom.add.release.sys.global.s32` | `atomicAdd_system(ptr, val)` — Metax supports `cuda/std/atomic` with `thread_scope_system` |
| `ld_na_relaxed` / `st_na_relaxed` | `L1::no_allocate` | `*ptr` / `*ptr = val` — cache policy uncontrollable without PTX; accept L1 pollution until Metax adds cache-hint intrinsics |
| `ld_nc_global` | `ld.global.nc` | `__ldg(ptr)` if available, else `*ptr` |
| `bar.sync` | `bar.sync id, num` | `__syncthreads()` or `cooperative_groups` subgroup barrier |
| `cooperative_groups::this_grid().sync()` | Grid sync | Metax cu-bridge already provides `cooperative_groups.h` compatibility |
| `__byte_perm` | Byte permute | Pure C++ bit-shifts |

By isolating these into one file per vendor, we can benchmark and optimize each primitive incrementally without touching the 700-line kernel.

## Build System

One CMake option selects the vendor:

```cmake
set(EP_GPU_VENDOR "CUDA" CACHE STRING "GPU vendor for EP/PG: CUDA, MACA, MUSA, HIP")
```

**Production build** (`BuildEpExt.cmake`) selects the corresponding `setup.py`:
- `EP_GPU_VENDOR=CUDA` → `setup.py`
- `EP_GPU_VENDOR=MACA` → `setup_maca.py`

`setup_maca.py` reuses the **same source files** but injects `-I platform/maca`:

```python
sources = [
    "src/ep_py.cpp",
    "src/mooncake_ep_buffer.cpp",
    "src/mooncake_ep_kernel.cu",   # SAME file — primitives switched via include path
    "src/mooncake_ibgda/mlx5gda.cpp",
]
extra_compile_args = {
    "cxx": ["-I" + os.path.join(current_dir, "platform", "maca"), "-DUSE_MACA"],
    "nvcc": ["-I" + os.path.join(current_dir, "platform", "maca"), "-DUSE_MACA"],
}
```

## Comparison with Feature Request #1965

| Aspect | FR #1965 (stub) | This Design |
|--------|-----------------|-------------|
| Existing `.cu` modified? | No — stub throws `runtime_error` | Minimal — 6 PTX lines → macros |
| Kernel actually runs? | ❌ No | ✅ Yes |
| Performance | N/A | Preserved per-primitive |
| Maintenance | Low | Medium (one primitives file per vendor) |
| Extensibility to MUSA/HIP | Same stub pattern | Add new `platform/` subdirectory |

## Implementation Plan

1. **PR 1:** Extract PTX from `mooncake_ep_utils.cuh` into `platform/cuda/ep_primitives.cuh`; extract 6 direct PTX from `kernel.cu` into macros. Verify CUDA build is unchanged.
2. **PR 2:** Add `platform/maca/ep_primitives.cuh` with conservative standard CUDA C++ implementations; add `setup_maca.py`; wire `EP_GPU_VENDOR` into CMake.
3. **PR 3:** Benchmark MACA primitives against CUDA baseline; replace conservative implementations with Metax intrinsics as available.
4. **PR 4 (Future):** Reserve `platform/musa/` and `platform/hip/` slots.

- [ ] Make sure you already searched for relevant issues and read the [documentation](https://kvcache-ai.github.io/Mooncake/)

Dayuxiaoshui:
## Description

This PR stabilizes the MACA Expert Parallelism (EP) runtime path on a
single-node MetaX C500 setup. The submitted path does not use the experimental
MCCL all-to-all helper. It keeps the existing EP dispatch/combine kernels and
uses the TransferEngine device P2P/IPC fast path for intra-node GPU-to-GPU
payload movement.

The main goal is to make the MACA EP path compile, initialize, dispatch, and
combine reliably while keeping the change surface limited to MACA-specific
compatibility code.

Key changes:

- Enable MACA EP builds through the torch extension path.
- Use the TransferEngine device P2P transport for MACA EP intra-node payloads.
- Keep IBGDA disabled on MACA when device-side RDMA is unavailable.
- Use a MACA RDMA transport stub so EP can fall back cleanly to P2P-only mode.
- Add MACA device operation compatibility wrappers.
- Avoid exposing temporary debug and phase-fence environment variables.
- Keep the MACA SEND/RECV phase fence as an internal compatibility step.
- Preserve the existing CUDA/NVIDIA EP behavior as much as possible.

Current runtime behavior:

- On MACA, EP reports `IBGDA unavailable, using P2P-only path` when IBGDA cannot
  be initialized.
- The validated payload path is still the EP runtime P2P fast path:
  `fallback=False`, `ibgda_disabled=True`, `fast=True`.
- For 4-GPU full mesh P2P testing across MetaXLink pairs, the test used
  `MOONCAKE_EP_MACA_ALLOW_NODE_P2P=1`.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [x] Performance improvement
- [ ] Other

### Build and static checks

```bash
python -m py_compile mooncake-wheel/mooncake/mooncake_ep_buffer.py
git diff --check
```

Result: passed.

The EP extension was rebuilt in the MACA torch environment:

```bash
cd mooncake-ep
MOONCAKE_EP_USE_MACA=1 \
MACA_PATH=/opt/maca \
MACA_HOME=/opt/maca \
TORCH_CUDA_ARCH_LIST=8.0 \
/opt/miniconda3/envs/py310/bin/python setup.py build_ext --build-lib . --force
```

Result: passed.

### EP smoke tests

2-GPU smoke:

```bash
MOONCAKE_EP_USE_MACA=1 \
python scripts/metax/smoke_ep_p2p.py \
  --world-size 2 --tokens 128 --hidden 2048 --experts 16 --topk 2
```

Result:

```text
smoke ok ranks=2 tokens=128 hidden=2048 experts=16 topk=2
```

4-GPU smoke with full node P2P enabled:

```bash
MOONCAKE_EP_USE_MACA=1 \
MOONCAKE_EP_MACA_ALLOW_NODE_P2P=1 \
python scripts/metax/smoke_ep_p2p.py \
  --world-size 4 --tokens 64 --hidden 2048 --experts 16 --topk 2
```

Result:

```text
smoke ok ranks=4 tokens=64 hidden=2048 experts=16 topk=2
```

### EP performance checks

The following measurements were collected on June 29, 2026 on a 4-GPU
MetaX C500 MACA node. Timing is max across ranks.

2 GPUs, `tokens=2048`, `hidden=7168`, `topk=2`, Gloo control plane:

```text
config: ranks=2 tokens=2048 hidden=7168 experts=16 topk=2 backend=gloo
mode=runtime_p2p fallback=False ibgda_disabled=True fast=True
dispatch: avg=10289.70 us bw=5.72 GB/s
combine: avg=7946.85 us bw=7.39 GB/s
dispatch+combine: avg=18189.30 us bw=6.46 GB/s
```

4 GPUs, `tokens=2048`, `hidden=7168`, `topk=2`,
`MOONCAKE_EP_MACA_ALLOW_NODE_P2P=1`, Gloo control plane:

```text
config: ranks=4 tokens=2048 hidden=7168 experts=16 topk=2 backend=gloo
mode=runtime_p2p fallback=False ibgda_disabled=True fast=True
dispatch: avg=10434.00 us bw=5.64 GB/s
combine: avg=8093.84 us bw=7.25 GB/s
dispatch+combine: avg=18516.72 us bw=6.35 GB/s
```

4 GPUs, `tokens=2048`, `hidden=7168`, `topk=4`,
`MOONCAKE_EP_MACA_ALLOW_NODE_P2P=1`, Gloo control plane:

```text
config: ranks=4 tokens=2048 hidden=7168 experts=16 topk=4 backend=gloo
mode=runtime_p2p fallback=False ibgda_disabled=True fast=True
dispatch: avg=13126.86 us bw=8.97 GB/s
combine: avg=12778.48 us bw=9.19 GB/s
dispatch+combine: avg=25871.60 us bw=9.09 GB/s
```

## Current Limitations

- MACA IBGDA is not active in this path. Current measurements are P2P/IPC
  measurements, not GPU-initiated RDMA measurements.
- On the tested node, default 4-GPU routing only enables P2P fast path for the
  close MetaXLink pairs. Full 4-GPU P2P testing requires
  `MOONCAKE_EP_MACA_ALLOW_NODE_P2P=1`.
- Without full node P2P enabled, 4-GPU all-to-all traffic can fall back to the
  Python fallback path. That fallback is not the performance target of this PR.
- A Mooncake PG control-plane run printed valid 2-GPU EP bandwidth, but one rank
  segfaulted during process-group teardown. Gloo control-plane runs completed
  cleanly, so this PR treats PG teardown as a separate follow-up issue.
- The previous MCCL all-to-all helper experiment is intentionally not part of
  this submitted path.

## Checklist

- [x] I have added or run tests to prove my changes are effective

## AI Assistance Disclosure

- [ ] No AI tools were used
- [x] AI tools were used

AI assistance was used to inspect the MACA EP changes, summarize test results,
and draft this PR description.


mooncake-wheel/mooncake/mooncake_ep_buffer.py
```diff
@@ -73,21 +97,115 @@ def __init__(self, group: dist.ProcessGroup, num_ep_buffer_bytes: int = 0):
         self.num_ep_buffer_bytes = num_ep_buffer_bytes
         self.backend = self.group
         # NIC auto-detection happens inside ep.Buffer via Topology::discover().
+        _debug_init(self.rank, "before ep.Buffer")
```
UNIDY2002: Do we really need such debug messages?

mooncake-wheel/mooncake/mooncake_ep_buffer.py
```diff
@@ -157,51 +275,91 @@ def connect(self, is_update: bool = False):
...
-            warnings.warn(
-                f"[Rank {self.rank}] Failed to exchange IPC handles: {e}. Falling back.",
-                RuntimeWarning,
-                stacklevel=2,
-            )
+        if self.group_size == 1:
+            _debug_init(self.rank, "single-rank skip ipc handle export")
+            self._use_fallback = False
+            _debug_init(self.rank, "connect done fallback=False")
+            return
```
UNIDY2002, line 254: Why do we need to handle this special case?

mooncake-wheel/mooncake/mooncake_ep_buffer.py
```diff
@@ -73,21 +97,115 @@ def __init__(self, group: dist.ProcessGroup, num_ep_buffer_bytes: int = 0):
...
         # Note: `sync_nvlink_ipc_handles()` can mutate C++ `ibgda_disabled_` (True->False when
         # P2P+IPC succeeds for all ranks). We re-evaluate after IPC sync below.
         self._use_fallback = bool(self.runtime.ibgda_disabled())
         self._fallback_next_combine_buffer: Optional[torch.Tensor] = None
+        self._maca_phase_token: Optional[torch.Tensor] = None
+        self._maca_phase_recv_tokens: Optional[List[torch.Tensor]] = None
         self.connect()
+
+    def _maca_phase_fence(self, send_event: Optional[Any] = None) -> None:
+        if not _USE_MACA or _MACA_PHASE_FENCE in {"", "0", "off", "none"}:
```
UNIDY2002: In what case do we need the maca phase fence flag? I'm afraid introducing new env vars may cause understanding burdens to users, so if you could explain that in advance, it would be very helpful.

mooncake-wheel/mooncake/mooncake_ep_buffer.py
```diff
@@ -73,21 +97,115 @@ def __init__(self, group: dist.ProcessGroup, num_ep_buffer_bytes: int = 0):
...
+            return
+
+        backend = dist.get_backend(self.group)
```
UNIDY2002, line 110: When do we use gloo instead of mooncake-cpu?

mooncake-ep/src/CMakeLists.txt
```diff
@@ -1,4 +1,7 @@
-add_library(mooncake_ep ep_py.cpp mooncake_ep_buffer.cpp mooncake_ep_kernel.cu)
+add_library(mooncake_ep
```
UNIDY2002: Better leave this file unchanged lol

mooncake-wheel/mooncake/mooncake_ep_buffer.py
Dayuxiaoshui: Removed. These messages were only used while diagnosing MACA initialization hangs and are not needed in the submitted path.

mooncake-wheel/mooncake/mooncake_ep_buffer.py
Dayuxiaoshui: This is intentional. With a single rank there is no peer that can import an IPC handle, so exporting/gathering IPC handles is unnecessary. On MACA it also avoids an unnecessary driver IPC call during initialization. I added a comment to make this explicit.

mooncake-wheel/mooncake/mooncake_ep_buffer.py
Dayuxiaoshui: This was originally a diagnostic switch used to compare different MACA phase synchronization strategies while debugging the split SEND/RECV path. I agree it should not be exposed as a user-facing env var, so I removed the flag and diagnostic modes. The remaining fence is an internal MACA compatibility fence only.

mooncake-wheel/mooncake/mooncake_ep_buffer.py
Dayuxiaoshui: We only handle Gloo because some smoke tests pass a Gloo process group, so the tiny fence token uses CPU there while EP payload still stays on the P2P fast path.

mooncake-ep/src/CMakeLists.txt
Dayuxiaoshui: No problem

mooncake-wheel/mooncake/mooncake_ep_buffer.py
UNIDY2002: Which smoke tests? I may be missing some context.

Dayuxiaoshui: This refers to the local MACA validation script scripts/metax/smoke_ep_p2p.py, where Gloo is only used to isolate and validate the EP P2P payload path, not to replace mooncake-cpu in production.

mooncake-wheel/mooncake/mooncake_ep_buffer.py
```diff
@@ -198,10 +284,31 @@ def connect(self, is_update: bool = False):
...
+        try:
+            return dist.get_backend(self.group) == "mooncake"
+        except Exception:
+            return False
+
+    def _active_ranks_tensor(
+        self, device: torch.device, dtype: torch.dtype = torch.int32
+    ) -> torch.Tensor:
+        if not self._is_mooncake_backend():
+            return torch.ones((self.group_size,), dtype=dtype, device=device)
```
UNIDY2002, line 316: I think we should raise a warning in this case, as running without mooncake-pg may lose the active_ranks features.

UNIDY2002 (commented):
Would you mind updating the PR title to fit the actual impl?

Also, I have a small comment.

mooncake-wheel/mooncake/mooncake_ep_buffer.py
Dayuxiaoshui: no problem

UNIDY2002 (approved):
LGTM

PR #2597, opened 2026-06-24
[Store] reject size mismatch in StorageBackendAdaptor::BatchLoad

he-yufeng:
## What

`StorageBackendAdaptor::BatchLoad` copies the deserialized value into the
destination slice without checking its size:

```cpp
struct_pb::from_pb(kv, kv_buf);

if (!kv.value.empty()) {
    std::memcpy(slice.ptr, kv.value.data(), kv.value.size());  // no size check
}
```

`from_pb` sets `kv.value` to whatever was stored on disk. If that value is
larger than `slice.size`, the `memcpy` writes past the end of `slice.ptr`
(an out-of-bounds write); if it is smaller, the tail of the slice is left
uninitialized while the caller assumes it got `slice.size` bytes.

The other two implementations of the same `BatchLoad` interface already
guard against exactly this — they reject a destination/stored size mismatch
with `INVALID_PARAMS` before copying:

- `BucketStorageBackend::BatchLoad` — `if (metadata.data_size != static_cast<int64_t>(dest_slice.size)) { ... return INVALID_PARAMS; }`
- `OffsetAllocatorStorageBackend::BatchLoad` — `if (dest_slice.size != entry.value_size) { ... return INVALID_PARAMS; }`

`StorageBackendAdaptor::BatchLoad` is the file-per-key path reached through
`FileStorage::BatchLoad`, and it is the only one missing the check. This PR
adds the same guard so all three behave consistently.

## Test

Added `FileStorageTest.BatchLoad_RejectsSliceSizeMismatch`: it offloads a
key through the file-per-key adaptor, then calls `BatchLoad` with a slice
one byte larger than the stored value and asserts `INVALID_PARAMS` instead
of an unchecked copy. The destination buffer is over-allocated so the test
asserts on the return code rather than relying on the unguarded copy
staying in bounds. The test fails before this change (the copy returns
success) and passes after.

I don't have the C++ toolchain to build mooncake-store locally, so I'm
relying on CI for the build and the test run. Happy to adjust the test
location or assertions if you'd prefer it elsewhere.

### DIFF
--- a/mooncake-store/src/storage_backend.cpp
+++ b/mooncake-store/src/storage_backend.cpp
@@ -1118,6 +1118,17 @@ tl::expected<void, ErrorCode> StorageBackendAdaptor::BatchLoad(
 
         struct_pb::from_pb(kv, kv_buf);
 
+        // The loaded value must match the destination slice exactly. The other
+        // BatchLoad backends reject a size mismatch (see BucketStorageBackend
+        // and OffsetAllocatorStorageBackend); without the same guard a stored
+        // value larger than slice.size overruns slice.ptr in the memcpy below.
+        if (kv.value.size() != slice.size) {
+            LOG(ERROR) << "Size mismatch for key: " << key
+                       << ", expected: " << kv.value.size()
+                       << ", got: " << slice.size;
+            return tl::make_unexpected(ErrorCode::INVALID_PARAMS);
+        }
+
         if (!kv.value.empty()) {
             std::memcpy(slice.ptr, kv.value.data(), kv.value.size());
         }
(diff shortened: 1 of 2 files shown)


he-yufeng:
Closing this one — I took another pass at the read path and the guard turns out to be redundant.

In `StorageBackendAdaptor::BatchLoad` the read length is derived from the destination slice: `kv.value.resize(slice.size)`, then `to_pb(kv, kv_buf)`, and `LoadObject` reads exactly `kv_buf.size()` bytes with an exact-length check (`*read_result != length` → `FILE_READ_FAIL`). So a stored value whose size differs from `slice.size` already fails at the read before the copy, and `kv.value` ends up bounded by `slice.size`, which means the `memcpy` into `slice.ptr` can't overrun. The mismatch is effectively rejected at read time already; the extra check only changes the error code in a case that mostly isn't reachable, and the test I added asserts the wrong error (it hits the exact-length read first and gets `FILE_READ_FAIL`, not `INVALID_PARAMS`).

Not worth keeping open. Sorry for the noise.

PR #2598, opened 2026-06-24
[Store] Optimize put performence by reduce memcpy and pinned memory

### LINKED ISSUE #2479: [RFC]: Zero-Copy Read/Write: Eliminate Redundant Data Copies
## Background and Motivation

Mooncake Store's read/write paths have three systemic issues:

1. **GPU pointer unsafety**: 8 `std::memcpy` call sites may receive GPU device pointers, causing crashes or silent data corruption.
2. **Redundant staging copies**: The write path always stages user data into an intermediate buffer — 2 copies instead of 1.
3. **Slow GPU→CPU bandwidth**: `cudaMemcpy` to pageable (non-CUDA-pinned) memory is throttled to ~3-6 GB/s by internal driver staging. Pinned memory achieves ~12+ GB/s.

### Root Cause

All six `*_internal` write functions unconditionally stage in RealClient:

```
user buffer → ① std::memcpy → staging buffer → ② CopyAuto → target segment
```

- Copy ① is **GPU unsafe** (`std::memcpy` cannot operate on device memory)
- Copy ① is **redundant** — staging can be skipped if we push the decision downstream
- Both staging buffer and segment are **pageable** (not CUDA-pinned), so even `cudaMemcpy` via `CopyAuto` only achieves ~3-6 GB/s

### Current vs Target Copy Counts

| Scenario | Current | Target |
|----------|---------|--------|
| GPU → local segment | 2 copies, ~3-6 GB/s, GPU unsafe | **1 copy, ~12+ GB/s** |
| CPU → local segment | 2 copies | **1 copy** |
| Registered buf → remote (RDMA) | 1+RDMA, GPU unsafe | **0+RDMA** |
| Unregistered buf → remote (RDMA) | 1+RDMA, GPU unsafe | **1+RDMA, GPU safe** |
| `get_buffer` → `buffer_to_tensor` | 2 copies | **1 copy** |

---

## Core Design

Two orthogonal changes that compose into the solution:

1. **`cudaHostRegister` segment and staging memory** — enables full-bandwidth GPU DMA
2. **Push staging decision to TransferSubmitter** — eliminates the redundant copy

### Current Architecture

```mermaid
flowchart TD
    A["RealClient *_internal\nallocate staging\nstd::memcpy(staging, user_data) ⚠️ GPU UNSAFE"] --> B["client_->Put(key, slices)"]
    B --> C{"TransferSubmitter\nselectStrategy"}
    C -->|local| D["LOCAL_MEMCPY\nCopyAuto → segment (pageable)\n⚠️ ~3-6 GB/s"]
    C -->|remote| E["TRANSFER_ENGINE\nRDMA from staging"]
    style A fill:#f66,color:#fff
    style D fill:#f96,color:#fff
```

### Proposed Architecture

```mermaid
flowchart TD
    A["RealClient *_internal\nbuild slices from user pointer\nNO staging, NO memcpy"] --> B["client_->Put(key, slices)"]
    B --> C{"TransferSubmitter\nselectStrategy"}
    C -->|local| D["LOCAL_MEMCPY\nCopyAuto → segment (CUDA-pinned)\n✅ ~12+ GB/s, 1 copy"]
    C -->|remote| F{"isRegistered?"}
    F -->|yes| G["RDMA direct\n0 local copy"]
    F -->|no| H["ensureRegisteredForRDMA\nstage + RDMA"]
    style D fill:#4a4,color:#fff
    style G fill:#4a4,color:#fff
```

---

### `cudaHostRegister` segment and staging memory

At mount time (`MountSegmentAndGetId`), after RDMA registration (`ibv_reg_mr`), call `cudaHostRegister` on the segment memory to make it CUDA-pinned. Same for the staging buffer at `setup_internal`.

```cpp
#ifdef USE_CUDA
auto ret = cudaHostRegister(buffer, size, cudaHostRegisterDefault);
if (ret != cudaSuccess) {
    LOG(WARNING) << "cudaHostRegister failed: " << cudaGetErrorString(ret);
    // Non-fatal: falls back to pageable-speed cudaMemcpy
}
#endif
```

**Why this works**:
- Segment memory is `mmap`-allocated (hugepage, 2MB-aligned), long-lived, never reallocated — ideal for `cudaHostRegister`
- `ibv_reg_mr` (RDMA pin) and `cudaHostRegister` (CUDA pin) coexist — different subsystems pinning the same pages
- **Precedent**: Mooncake's tent NVLink transport already uses `cudaHostRegister` on CPU memory in production (`nvlink_transport.cpp:276`)

### `MemcpySafe`: GPU-safe generic memcpy

```cpp
// gpu_staging_utils.h
inline bool MemcpySafe(void* dst, const void* src, size_t size) {
    if (!IsDevicePointer(src) && !IsDevicePointer(dst)) {
        std::memcpy(dst, src, size); return true;
    }
    SetDevice(/* whichever is GPU */);
    return CopyAuto(dst, src, size);
}
```

### Copy efficiency: `CopyAuto` and `MemcpySafe`

`CopyAuto` wraps `cudaMemcpy(dst, src, size, cudaMemcpyDefault)` — a synchronous, blocking copy. Two overhead sources:

| Source | Cost | Impact at MB+ scale |
|--------|------|---------------------|
| `cudaMemcpy` sync overhead (implicit stream create + wait) | ~5-10 us/call | Negligible (<0.1%) |
| `cudaPointerGetAttributes` in `IsDevicePointer` (×2 per call) | ~1-5 us/call | Negligible |

**Bandwidth** (with `cudaHostRegister`'d destination):
- GPU→pinned host: ~12-25 GB/s (PCIe Gen4/5 DMA, full line rate)
- CPU→CPU fallback: `std::memcpy` at ~20+ GB/s

This is sufficient for Mooncake Store's primary workload (MB+ tensor writes). The overhead is amortized over large transfers.

**Future optimization**: Replace `cudaMemcpy` with `cudaMemcpyAsync` on a persistent per-worker CUDA stream. Eliminates the per-call stream creation overhead and allows CPU-side pipelining. Not required for Phase 1 — the bandwidth bottleneck is PCIe, not API overhead.

### `ensureRegisteredForRDMA`: on-demand staging for TRANSFER_ENGINE

Only allocates staging when the RDMA path is selected AND a slice is not RDMA-registered:

```cpp
auto [prepared_slices, staging_handles] = ensureRegisteredForRDMA(slices);
// registered → original ptr; unregistered → staging copy
```

### Staging removal from `*_internal`

Six write functions change from `allocate → memcpy → split_into_slices(staging)` to `split_into_slices(user_ptr)`. Follows the existing zero-copy pattern of `put_from_internal`.

**Lifetime safety**: All `Put`/`Upsert` calls are synchronous — user buffer on caller's stack stays alive throughout the transfer.

---

## Read Path

- `get_into` / `get_into_ranges`: already optimal for memory replicas (1 copy). No change.
- `get_buffer` → `buffer_to_tensor`: reads into staging, then `new char[] + memcpy`. Fix: transfer `BufferHandle` ownership to the tensor capsule (1 copy instead of 2).
- `execute_tensor_into_plan_transfers`: `std::memcpy(metadata → gpu_buffer)` is GPU unsafe. Fix: replace with `MemcpySafe`.

---

## GPU-Unsafe Call Sites

| Call site | File | Path |
|-----------|------|------|
| `put_internal` staging memcpy | `real_client.cpp:1551` | Write |
| `put_batch_internal` staging memcpy | `real_client.cpp:1629` | Write |
| `put_parts_internal` loop memcpy | `real_client.cpp:1733` | Write |
| `upsert_internal` staging memcpy | `real_client.cpp:3504` | Write |
| `upsert_parts_internal` loop memcpy | `real_client.cpp:3750` | Write |
| `upsert_batch_internal` staging memcpy | `real_client.cpp:3834` | Write |
| `batch_write_tensor_impl` data memcpy | `store_py_parallel_write.h:55-57` | Write |
| `execute_tensor_into_plan_transfers` metadata memcpy | `store_py_parallel_read.h:1074` | Read |

---

## Risks and Mitigations

| Risk | Impact | Mitigation |
|------|--------|-----------|
| `cudaHostRegister` fails (no GPU, insufficient pinnable memory) | GPU→segment copies remain at pageable speed (~3-6 GB/s) | Non-fatal: log WARNING, system works correctly at reduced GPU bandwidth. CPU→CPU path unaffected. |
| `cudaHostRegister` on hugepage memory | Potential compatibility issues on some kernels | Validated: tent NVLink transport uses same pattern in production. `cudaHostRegister` supports hugepages since CUDA 11. |
| Large segment registration overhead | One-time latency at startup (ms-level for GB-scale memory) | Amortized over process lifetime. Only done once at mount time. |
| `ibv_reg_mr` + `cudaHostRegister` coexistence | Theoretical double-pinning conflict | Both pin OS pages — independent subsystems (NIC DMA vs GPU DMA). No known conflict. RDMA registration uses `IBV_ACCESS_LOCAL_WRITE`; CUDA registration uses page table entries. |
| Removing staging breaks `*_internal` user buffer lifetime | Data corruption if user buffer freed during transfer | All Put/Upsert/BatchPut are synchronous — user buffer guaranteed alive. `std::span<const char>` references caller stack data. |
| System pinned memory limit (`vm.max_map_count`, RLIMIT_MEMLOCK) | Registration silently fails or OOM-kills | Check `ulimit -l` at startup; document minimum requirements. Fallback is transparent (pageable speed). |

---

## Error Handling

```mermaid
flowchart LR
    A["GPU/RDMA failure"] --> B["CopyAuto or MemcpySafe returns false"]
    B --> C["TransferFuture → TRANSFER_FAIL"]
    C --> D["Client::Put → PutRevoke\nrolls back master metadata"]
    D --> E["error returned to caller"]
```

---

## Scope and Phasing

| Phase | Scope | Files | Est. lines |
|-------|-------|-------|-----------|
| 1 | GPU safety + `cudaHostRegister` segment/staging | `gpu_staging_utils.h`, `real_client.cpp`, `client_service.cpp`, `store_py_parallel_*.h` | ~80 |
| 2 | Remove staging from `*_internal` + `ensureRegisteredForRDMA` | `real_client.cpp`, `transfer_task.h/.cpp`, `transfer_engine*.h/.cpp`, `client_service.cpp` | ~200 |
| 3 | Read path: `buffer_to_tensor` ownership transfer | `store_py.cpp` | ~30 |
| | **Total** | | **~310** |

**Phase 1** is low-risk (additive changes, non-fatal fallback). Deploy and validate first.

**Phase 2** restructures the write path. Depends on Phase 1 for GPU safety. Higher blast radius but all changes are API-compatible.

**Phase 3** is isolated to Python binding layer.

## Compatibility

- **No API changes**. All public interfaces unchanged.
- **Non-GPU fallback**: `cudaHostRegister` failure is non-fatal. System operates correctly without GPU, just at pageable speed for GPU transfers.
- **Registration requirement**: Zero-copy remote RDMA still requires `register_buffer()`. Unregistered buffers handled transparently via `ensureRegisteredForRDMA`.

zxpdemonio:
## Motivation

The store write path currently performs unnecessary staging copies for many high-level write APIs.

High-level write APIs (`put`, `put_parts`, `put_batch`, `put_tensor`, `upsert`, `upsert_parts`, `upsert_batch`, `upsert_tensor`) previously always copied user data into a Store-managed staging buffer before submitting the transfer:

```text
user buffer
  -> memcpy to client staging buffer
  -> memcpy/RDMA to target segment
```

This has three issues:

1. Local writes pay one extra host memcpy.
2. GPU tensor writes are not handled safely in the old high-level write path because device pointers must not be copied with `std::memcpy`.
3. GPU→CPU copies use pageable host memory unless users manually enable pinning, which limits bandwidth.

This PR removes the unconditional high-level write staging copy and makes GPU→CPU writes a first-class fast path.

## Scenario

This optimization targets write-heavy GPU/CPU store workloads, especially:

- GPU tensor writes to local Store segments.
- CPU tensor/object writes through high-level APIs.
- RDMA writes where user buffers may or may not already be registered.
- Tensor batch writes from Python integration APIs.

The most important case is GPU→CPU local write. On NVIDIA A10, pinned host memory improves 512MB GPU→CPU write bandwidth from ~8–9 GB/s to ~25 GB/s.

### Remove unconditional staging in high-level write internals

This PR updates the six high-level write internal functions:

- `put_internal`
- `put_parts_internal`
- `put_batch_internal`
- `upsert_internal`
- `upsert_parts_internal`
- `upsert_batch_internal`

Before:

```text
allocate client staging buffer
memcpy user data -> staging buffer
split staging buffer into slices
submit slices
```

After:

```text
split user buffer directly into slices
submit slices
```

The transfer layer now decides whether staging is actually required.

For local memcpy writes, this reduces the write path from two copies to one copy.

For RDMA writes with already-registered buffers, this also removes one host memcpy.

For RDMA writes with unregistered buffers, the total copy count remains the same, but staging is now centralized in the transfer submit layer.

### Add RDMA-safe staging in the transfer layer

The transfer submit layer now checks whether write slices are registered with the transfer engine.

If all slices are already registered:

```text
registered user buffer
  -> RDMA WRITE
```

If any slice is not registered:

```text
user buffer
  -> MemcpySafe() to registered staging buffer
  -> RDMA WRITE
```

The staging path:

- batches all unregistered slices into one contiguous staging allocation
- keeps staging `BufferHandle`s alive until transfer completion
- uses `MemcpySafe()` so GPU pointers are handled by the proper accelerator runtime
- splits both staged and already-registered slices by `kMaxSliceSize`

This preserves RDMA correctness while avoiding unnecessary staging for local and registered-buffer paths.

### Enable pinned host memory by default for GPU builds

For GPU builds, Store now tries to pin Store-managed host memory with `cudaHostRegister` by default.

Pinned memory is applied to two Store-owned buffer classes:

1. **Client staging buffer**
   - Used when RDMA writes need a registered source buffer but the user buffer is not already registered.
   - Pinning this buffer improves GPU→CPU staging copies before RDMA WRITE.
   - The buffer is unpinned during `RealClient` teardown.

2. **Mounted segment buffers**
   - Used as the local Store target for memory-backed objects.
   - Pinning these buffers improves GPU→CPU local writes because CUDA can DMA directly into the Store segment instead of using its internal pageable-memory staging path.
   - Segment buffers are unpinned on unmount. If mounting the segment to the master fails after local registration, Store now also unpins and unregisters the buffer on that failure path.

This is a best-effort acceleration path:

- If `cudaHostRegister` fails, Store keeps running and falls back to pageable host memory for that buffer.
- If the configured pinned-memory budget is exceeded, Store skips pinning that buffer and keeps running.
- Correctness does not depend on pinned memory; only GPU→CPU copy performance does.
- Store tracks only successfully pinned regions and only calls `cudaHostUnregister` for those tracked regions.

Opt out:

```bash
MC_STORE_PIN_MEMORY=0
# or
MC_STORE_PIN_MEMORY=false
```

Optionally cap total Store-managed pinned memory:

```bash
MC_STORE_PIN_MEMORY_MAX_BYTES=8589934592  # 8GB
```

Unset or `0` means no explicit Store-side cap.

Pinned memory has important side effects: pinned pages cannot be swapped or reclaimed by the OS, large registrations have setup cost, and pinning a whole large segment can increase system memory pressure. The cap is therefore a safety valve for deployments that use large segments or share the machine with other workloads.

This PR intentionally implements the minimal safe policy: default-on best-effort pinning for Store-managed buffers, plus an opt-out and a total pinned-memory cap. Future work should replace whole-buffer pinning with a registration cache that pins page-aligned hot ranges on demand and evicts them with an LRU/lazy-unpin policy. That would reduce page-locked memory pressure for very large segments while preserving the direct GPU→segment fast path for frequently written ranges.

### Python tensor batch path uses multi-buffer writes

`batch_put_tensor` and `batch_upsert_tensor` now pass tensor metadata and tensor data as separate buffers through the multi-buffer write APIs.

This avoids the integration-layer staging copy that previously happened before calling the lower-level Store write APIs.

## Basic Usage

No API change is required.

Existing write APIs automatically benefit:

```python
store.put(key, data)
store.put_tensor(key, tensor)
store.batch_put_tensor(keys, tensors)
store.upsert_tensor(key, tensor)
```

Pinned host memory is enabled by default for GPU builds.

To disable pinned memory:

```bash
export MC_STORE_PIN_MEMORY=0
```

To cap total Store-managed pinned memory:

```bash
export MC_STORE_PIN_MEMORY_MAX_BYTES=8589934592  # 8GB
```

If the cap is exceeded or `cudaHostRegister` fails, Store keeps running and uses pageable host memory for that buffer.

## Module

- [ ] Docs

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [x] Performance improvement

## Performance

Benchmarks were run on an NVIDIA A10 test machine.

### Copy count reduction

The affected APIs are the high-level write APIs that accept user data directly, such as `put`, `put_parts`, `put_batch`, `put_tensor`, `upsert`, `upsert_parts`, `upsert_batch`, and `upsert_tensor`.

Before this PR, these APIs always copied user data into a Store-managed staging buffer before submitting the transfer. This PR removes that unconditional staging copy and lets the transfer layer stage only when RDMA correctness requires it.

| API / scenario | Before | After | Copy reduction | Notes |
|---|---:|---:|---:|---|
| High-level write APIs + local memcpy (`put`, `put_tensor`, `upsert`, etc.) | 2 copies | 1 copy | -1 copy | User buffer is copied directly into the local Store segment. |
| High-level write APIs + RDMA + registered user buffer | 2 copies | 1 copy | -1 copy | Registered user buffer can be used directly as the RDMA source. |
| High-level write APIs + RDMA + unregistered user buffer | 2 copies | 2 copies | 0 | Staging is still required, but it is now centralized in the transfer layer. |
| High-level write APIs + GPU source buffer | Crash-prone | 1 or 2 copies | N/A | The old path used `std::memcpy` on user data, which is unsafe for GPU pointers. The new path uses accelerator-aware copy in the transfer layer. |
| Direct-slice APIs (`put_from`, `batch_put_from`, etc.) + RDMA + unregistered user buffer | Crash-prone | 2 copies | N/A | The old direct-slice RDMA path assumed source buffers were already registered. The new transfer-layer staging path handles unregistered buffers safely. |
| Direct-slice APIs (`put_from`, `batch_put_from`, etc.) + registered/local buffer | 1 copy | 1 copy | 0 | These APIs already submit user slices directly. |
| Python `batch_put_tensor` / `batch_upsert_tensor` local memcpy | 2 copies | 1 copy | -1 copy | Metadata and tensor data are now passed as separate buffers through multi-buffer writes. |

### Measured copy-reduction benefit

The main measured before/after comparison is 512MB CPU `put_tensor` with local memcpy. This is the case where the old path paid one extra host memcpy and the new path removes it.

| Scenario | Path | Before latency | After latency | Improvement | After bandwidth |
|---|---|---:|---:|---:|---:|
| Local memcpy, 512MB | `put_tensor` | ~112 ms | 73.01 ms | ~1.53x faster, ~39 ms saved | 7.35 GB/s |

This saved time corresponds to removing one extra host memcpy from the high-level local write path.

For reference, the direct-slice API baseline after this change was:

| Scenario | Path | Latency | Bandwidth | Notes |
|---|---|---:|---:|---|
| Local memcpy, 512MB | `put_from` | 38.23 ms | 14.04 GB/s | Existing direct-slice API; not the optimized target in this PR. |
| RDMA setup with `MC_STORE_MEMCPY=1`, 512MB | `put_tensor` | 76.90 ms | 6.98 GB/s | High-level write API after removing unconditional staging. |
| RDMA setup with `MC_STORE_MEMCPY=1`, 512MB | `put_from` | 74.70 ms | 7.19 GB/s | Existing direct-slice API under the same memcpy-forced setup. |

These rows are not before/after numbers. They are sanity checks showing that, after this PR, high-level writes no longer pay the old unconditional staging copy and are comparable to direct-slice writes under the same transfer strategy.

### Pinned memory benefit

GPU→CPU local write, pinned enabled vs disabled:

| Path | Size | Pinned OFF | Pinned ON | Speedup |
|---|---:|---:|---:|---:|
| GPU `put_tensor` | 64MB | 8.26 GB/s | 25.10 GB/s | 3.04x |
| GPU `put_tensor` | 256MB | 9.32 GB/s | 25.76 GB/s | 2.76x |
| GPU `put_tensor` | 512MB | 9.03 GB/s | 24.84 GB/s | 2.75x |
| GPU `put_from` | 64MB | 8.11 GB/s | 25.29 GB/s | 3.12x |
| GPU `put_from` | 256MB | 8.25 GB/s | 25.82 GB/s | 3.13x |
| GPU `put_from` | 512MB | 8.20 GB/s | 25.91 GB/s | 3.16x |

Pinned host memory makes GPU→CPU copies DMA directly into Store buffers instead of going through CUDA’s internal pageable-memory staging path.

## How Has This Been Tested?

**Build commands:**

```bash
cmake --build build --target store -j$(nproc)
```

Also built on the GPU/RDMA test machine after syncing this branch:

```bash
cmake --build build --target store -j$(nproc)
```

**Test commands:**

```bash
cmake --build build --target client_integration_test client_tcp_local_memcpy_test pybind_client_test -j$(nproc)
ctest --test-dir build -R 'client_integration_test|client_tcp_local_memcpy_test|pybind_client_test' --output-on-failure
```

**Test results:**

- [x] Unit tests pass
- [x] Integration tests pass
- [x] Manual performance testing done on GPU/RDMA test machine

| Test | Result | Coverage |
|---|---|---|
| `client_integration_test` | Passed | `put`, `put_parts`, `upsert`, `upsert_parts`, `upsert_batch`, `batch_upsert_from`, `batch_put_from_multi_buffers` |
| `client_tcp_local_memcpy_test` | Passed | TCP local memcpy behavior |
| `pybind_client_test` | Passed | PyClient write/update paths including `Upsert` and batch upsert cases |

CTest summary:

```text
100% tests passed, 0 tests failed out of 3
Total Test time (real) = 379.76 sec
```

### Performance validation

The performance numbers above were collected with temporary ad-hoc benchmark scripts on the test machine. These scripts are not part of this PR; they were used only to validate this optimization.

The benchmarks covered two comparisons:

1. **copy reduction**: compare high-level write API (`put_tensor`) against direct-slice baseline (`put_from`) under local memcpy and RDMA+`MC_STORE_MEMCPY=1`.
2. **pinned host memory**: compare GPU→CPU local writes with default pinned memory against `MC_STORE_PIN_MEMORY=0`.

The test machine used:

| Item | Value |
|---|---|
| GPU | NVIDIA A10 |
| RDMA device | `erdma_0` |
| Host | `192.168.22.70` |
| Master HTTP | `192.168.22.70:8090` |
| Master RPC | `192.168.22.70:50060` |
| Tensor sizes | 64MB, 256MB, 512MB |

Note: on this test machine, PyTorch's CUDA runtime was preloaded for the benchmark process to avoid a CUDA runtime symbol conflict when importing both PyTorch and the Mooncake Python extension:

```bash
LD_PRELOAD=/usr/local/lib/python3.12/dist-packages/nvidia/cuda_runtime/lib/libcudart.so.12
```

## Checklist

- [x] I have formatted the changed code

## AI Assistance Disclosure

Claude Opus 4.6 assisted with code generation, code review, benchmark analysis, and PR description drafting. All changes were reviewed and validated by me。


zxpdemonio:
@gemini-code-assist please review again

mooncake-integration/store/store_py.cpp
```diff
@@ -798,15 +799,46 @@ class MooncakeStorePyWrapper {
...
-                   const ReplicateConfig &write_config) {
-                return store_->batch_put_from(write_keys, buffer_ptrs,
-                                              buffer_sizes, write_config);
-            });
+        auto group_ids_error =
+            ValidateGroupIdsForBatchConfig(config, keys.size(), "put");
+        if (!group_ids_error.empty()) return group_ids_error;
+
+        std::vector<int> results(keys.size(), 0);
+        {
```
alogfans, line 807: Extending batch_write_tensor_impl is better for readability.

alogfans:
We recommend you trying to minimize code modifications as possible, or spiliting it to multiple PRs or commits.

zxpdemonio:
> We recommend you trying to minimize code modifications as possible, or spiliting it to multiple PRs or commits.

@alogfans Many thanks for reviewing, and the part of pinned memory has been split into https://github.com/kvcache-ai/Mooncake/pull/2945

zxpdemonio:
Thanks for the earlier feedback. This broad PR has been superseded by a series of smaller, focused PRs.

Core work split from this PR:

- #2926: fix GPU-addressed local copy crashes (merged)
- #2945: add quota-bounded pinned memory for Store host segments (merged)
- #3159: remove staging for same-node tensor put/upsert (open, CI green and under final review)

Supporting pieces for the complete same-node dummy-client tensor path were handled separately:

- #2998: support tensor APIs with dummy clients (merged)
- #3234: support CUDA IPC handoff for dummy-client CUDA buffers (merged)
- #3197: remove staging for same-process GPU reads (merged)

The merged PRs cover the safety, pinned-memory, dummy-client/CUDA-IPC, and read-side portions. #3159 is the last remaining write-side optimization. Closing this superseded umbrella PR in favor of those scoped changes.

PR #2599, opened 2026-06-24
[Store] Make offloading_queue_limit and offload_cap_ratio configurabl…

leonzzhu:
## Motivation

In production we observed that `offloading_queue_limit` was hardcoded to 50000
and `kOffloadCapRatio` was a `constexpr 0.5`, yielding a per-cycle offload cap
of only **25000 keys**. Once memory pressure triggered `BatchEvict`, any keys
exceeding this cap were force-evicted (discarded) instead of being offloaded
to SSD. The net effect was:

- NVMe bandwidth stayed largely idle
- KV-cache hit rate dropped significantly under sustained pressure

There was already a TODO comment in `master_service.cpp` noting that this
should be made configurable when workloads demand tuning — this PR delivers
that.

## Changes

- Adds two gflags to `master.cpp`:
  - `--offloading_queue_limit` (default 50000)
  - `--offload_cap_ratio` (default 0.5, validated to range [0.0, 1.0])
- Propagates both through the full config chain:
  `MasterConfig` → `MasterServiceSupervisorConfig` → `WrappedMasterServiceConfig` → `MasterServiceConfig` → `MasterService`
- Replaces the `constexpr kOffloadCapRatio` with member variable `offload_cap_ratio_`
- Adds documentation in `mooncake-store-deployment-guide.md` with formula,
  default behavior, tuning guidance and example

## Example usage

```bash
--offloading_queue_limit=500000 --offload_cap_ratio=0.8
# per-cycle offload cap = 500000 * 0.8 = 400000 (vs the previous fixed 25000)
```

## Backward compatibility

Defaults preserve previous behavior exactly (`50000 * 0.5 = 25000`), so existing
deployments are unaffected. No client/protocol changes.

## Checklist

- [x] Code formatted with `./scripts/code_format.sh` (clang-format 20.1.7, all OK)
- [x] Defaults preserve existing behavior (no breaking change)
- [x] Documentation updated (`mooncake-store-deployment-guide.md`)
- [ ] Unit tests — n/a; this PR is pure config plumbing (gflags → existing
      member variables), default values are unchanged, and the runtime
      behavior is already covered by the existing `offload_on_evict_test.cpp`
      suite. Happy to add a config-propagation test if maintainers prefer.

### DIFF
--- a/docs/source/deployment/mooncake-store-deployment-guide.md
+++ b/docs/source/deployment/mooncake-store-deployment-guide.md
@@ -524,6 +524,8 @@ Flags for controlling data movement between DRAM and SSD.
 | `--enable_offload` | `false` | Enable offload from DRAM to SSD |
 | `--offload_on_evict` | `false` | Defer offload to eviction time rather than at `Put` |
 | `--offload_force_evict` | `false` | Force-evict objects exceeding capacity without offload |
+| `--offloading_queue_limit` | `50000` | Max number of objects allowed in the offloading queue per local disk segment. Increase to allow more objects to be offloaded to SSD before force-eviction kicks in |
+| `--offload_cap_ratio` | `0.5` | Per-cycle offload cap as a fraction of `offloading_queue_limit` (range `[0.0, 1.0]`). Controls how many objects can be queued for offload in a single eviction cycle before falling back to force-evict |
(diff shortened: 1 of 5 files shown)


LujhCoconut (approved):
LGTM. Please resolve the merge conflicts before merging.

leonzzhu:
> LGTM. Please resolve the merge conflicts before merging.

Fix， thanks

PR #2605, opened 2026-06-24
[CI][Build] Build CUDA release wheels in the PyTorch manylinux image

mgoin:
## Description

Release wheels are built on a bare `ubuntu-22.04` runner, which tags them `manylinux_2_35` and bakes in that runner's `libstdc++`/toolchain. 

This PR builds `release.yaml` and `release-cuda13.yaml` inside `pytorch/manylinux2_28-builder` instead, lowering the wheel glibc floor to **2.28** (covers RHEL/Rocky/Alma 8+, Ubuntu 20.04+) and matching PyTorch's own wheel ABI. The image preinstalls CUDA, so the `Jimver/cuda-toolkit` and host disk-cleanup steps are dropped; Python comes from `/opt/python` in image.

I only targeted the CUDA x86 wheels here since the published aarch64 CUDA wheels seem to be built outside these workflows in this repo. How should aarch64 be handled (separate arm-runner job with `manylinuxaarch64-builder`)?

The EL8 (`dependencies.sh`) and CMake paths had only ever run on Debian/apt, so had to update a few things there:

- **dependencies.sh:** enable EPEL + PowerTools/CRB (`glog`/`gtest`/`protobuf`/`liburing`/`xxhash`); add `msgpack`/`zstd`/`gflags` devel; install `golang` for the Go etcd wrapper (Ubuntu runners ship Go preinstalled); use EPEL's `boost1.78` since system Boost 1.66 fails under C++20. Drop `grpc-devel`/`grpc-plugins` since not packaged for EL8 and only used by the legacy C++ etcd client.
- **CMake:** `find_package(Python3 ... Development)` requires a linkable `libpython` (`Development.Embed`), which manylinux omits; this is pybind11 module code, so use `Development.Module`. Alias `gflags::gflags` for older gflags that only export `gflags-shared`. Add boost1.78's versioned include dir when present.

## Module

- [x] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

Built all four CUDA 12/13 wheels (`manylinux_2_28`) across Python 3.10–3.13."

**Test results:**
- [ ] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used


.github/workflows/release-cuda13.yaml
```diff
@@ -23,27 +24,15 @@ jobs:
...
-          df -h
-
-      - name: Install CUDA Toolkit 13
-        uses: Jimver/cuda-toolkit@v0.2.29
-        with:
-          cuda: '13.0.2'
-          linux-local-args: '["--toolkit"]'
-          method: 'network'
-          sub-packages: '["nvcc", "nvrtc-dev"]'
-          non-cuda-sub-packages: '["libcusparse-dev", "libcublas-dev", "libcusolver-dev"]'
```
alogfans, line 46: Why removing them?

.github/workflows/release-cuda13.yaml
```diff
@@ -23,27 +24,15 @@ jobs:
...
+        run: git config --global --add safe.directory '*'
 
-      - name: Free up disk space
+      - name: Select Python ${{ matrix.python-version }} from manylinux image
         run: |
-          sudo rm -rf /usr/share/dotnet
-          sudo rm -rf /opt/ghc
-          sudo rm -rf /opt/hostedtoolcache/CodeQL
-          sudo rm -rf /usr/local/lib/android
-          df -h
```
Aionw, line 37: The free disk space is intentional, and I think it’s better to keep it. We often run into “out of space” issues, and having this extra space can help mitigate them to some extent.

dependencies.sh
```diff
@@ -189,23 +191,25 @@ elif [ "$OS" = "centos" ] || [ "$OS" = "rhel" ] || [ "$OS" = "rocky" ] || [ "$OS
                      cmake \
                      git \
                      wget \
+                     golang \
```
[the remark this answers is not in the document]
Aionw: This is a reasonable suggestion.

.github/workflows/release-cuda13.yaml
mgoin: The manylinux image ships CUDA, so we don't need CUDA toolkit install

.github/workflows/release-cuda13.yaml
mgoin: As I understand, inside a `container:` job those host paths are not visible anymore so the `rm` should be a no-op. You can free a host runner's disk from inside the container.

So I think we can choose to do two things:
1. Restructure to run the build via `docker run` in a step, and keep the host cleanup step
2. Keep the `container:` job in this PR and just monitor to see if you run into the same issues. I think this image should be much leaner than what you were doing before and fully contained in docker runtime

stmatengss:
All LGTM! The disk is full, I have re-triggered the CI. This appears to be a random error.

stmatengss (approved):

mgoin:
Thanks @stmatengss yes the failure seems to be in ci.yml, not this workflow. I opened https://github.com/kvcache-ai/Mooncake/pull/2626 to see if this could help with the disk pressure there

Aionw (approved):

PR #2012, opened 2026-04-29
[TE] Update IntraNode NVLink transfer method cuMemcpy -> cuMemcpyAsync

TTThanos:
## Description

Update legacy Transport: IntraNode NVLink to update previous cuMemcpy to cuMemcpyAsync. 
Move transfer completion poll check into getTransferStatus() function rather mark it as succeed right after cuMemcpyAsync.
This update help to improve the transfer performance when date block size is small (like 64 KB).
Above 50% improvement can be observed when using transfer_engine_bench as test case.

## Module

- [x] Transfer Engine (`mooncake-transfer-engine`)
- [ ] Mooncake Store (`mooncake-store`)
- [ ] Mooncake EP (`mooncake-ep`)
- [ ] Integration (`mooncake-integration`)
- [ ] P2P Store (`mooncake-p2p-store`)
- [ ] Python Wheel (`mooncake-wheel`)
- [ ] PyTorch Backend (`mooncake-pg`)
- [ ] Mooncake RL (`mooncake-rl`)
- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other

## How Has This Been Tested?
Performance has been tested on Mooncake transfer_engine_bench:
previous:
image
now:
image

SGlang bench serving 4k in / 2k out:
image

## Checklist

- [ ] I have performed a self-review of my own code.
- [ ] I have formatted my own code using `./scripts/code_format.sh` before submitting.
- [ ] I have updated the documentation.
- [ ] I have added tests to prove my changes are effective.

PR #2606, opened 2026-06-24
fix(maca): map cudaStreamQuery for the intra-node NVLink build

he-yufeng:
## What

`gpu_vendor/maca.h` maps every CUDA stream/event symbol the intra-node NVLink transport touches except `cudaStreamQuery`. As a result a `USE_MACA` + `USE_INTRA_NVLINK` build fails to compile the async-completion polling loop in `intranode_nvlink_transport.cpp`.

The polling path added in #2012 calls:

```cpp
// intranode_nvlink_transport.cpp:411
cudaError_t cuda_err = cudaStreamQuery(stream);
if (cuda_err == cudaSuccess) {
    slice->markSuccess();
} else if (cuda_err != cudaErrorNotReady) {
    slice->markFailed();
}
```

`cuda_alike.h` selects `maca.h` for `USE_MACA`, and `transport/CMakeLists.txt` compiles `intranode_nvlink_transport` whenever `USE_INTRA_NVLINK` is set (orthogonal to the GPU vendor). `maca.h` already aliases `cudaErrorNotReady` (used on the very next line) and the rest of the stream/event surface for this file — `cudaStreamCreateWithFlags`, `cudaStreamPerThread`, `cudaStreamWaitEvent`, `cudaEventCreateWithFlags`, `cudaEventRecord`, etc. — but `cudaStreamQuery` was missed, so the MACA build sees an undeclared identifier.

This is the MACA counterpart of #2575, which added the same missing aliases to `musa.h` for this exact NVLink path.

## Fix

One line, following the existing `cudaStream* -> mcStream*` convention (and `cudaEventQuery -> mcEventQuery`, already present):

```c
#define cudaStreamQuery mcStreamQuery
```

## Verification

I don't have the MetaX SDK to run a full `USE_MACA` build locally, so I reproduced the failure at compile level with a small stub standing in for the MACA runtime, using maca.h's current alias set and the exact polling snippet:

```
repro.cpp: In function 'bool poll(void*)':
repro.cpp:5:28: error: 'cudaStreamQuery' was not declared in this scope; did you mean 'mcStreamQuery'?
    5 |     cudaError_t cuda_err = cudaStreamQuery(stream);
      |                            ^~~~~~~~~~~~~~~
      |                            mcStreamQuery
```

The compiler itself suggests `mcStreamQuery`. Adding the `#define` above makes the same translation unit compile cleanly. No behavior change for CUDA/MUSA/other vendors — this only fills a missing alias in the MACA shim.

### DIFF
--- a/mooncake-transfer-engine/include/gpu_vendor/maca.h
+++ b/mooncake-transfer-engine/include/gpu_vendor/maca.h
@@ -133,6 +133,7 @@ static inline CUresult cuGetErrorString(CUresult error, const char **err_str) {
 #define cudaStreamDestroy mcStreamDestroy
 #define cudaStreamNonBlocking mcStreamNonBlocking
 #define cudaStreamPerThread mcStreamPerThread
+#define cudaStreamQuery mcStreamQuery
 #define cudaStreamSynchronize mcStreamSynchronize
 #define cudaStream_t mcStream_t
 #define cudaStreamWaitEvent mcStreamWaitEvent


alogfans (approved):
LGTM

alogfans:
Before merging, could you please submit an empty commit to retrigger CI?

he-yufeng:
Done, just pushed an empty commit to retrigger. For the record the earlier red was the runners hitting `No space left on device` during the coverage-upload and wheel steps, not the change itself, the build and the tests all passed. Thanks for the review.

Issue #2153, opened 2026-05-20
[RFC]: Dynamic Tenant Quota and Scoped-Key Isolation


Lin-z-w:
## Summary

This RFC proposes a Mooncake-managed tenant isolation model for Mooncake Store that combines dynamic quota allocation with tenant-scoped object identity. Instead of treating tenant quota as a static upper bound independent of physical memory, Mooncake computes an effective quota from current allocatable memory and tenant-requested quotas. Instead of relying on upper layers to prefix keys, Mooncake derives a collision-free internal object key from `(tenant_id, user_object_key)`.

The goal is to prevent the sum of enforceable tenant quotas from exceeding available storage while also preventing same-key object operations from crossing tenant boundaries.

## Motivation

The current tenant quota proposal in PR #2112 validates writes against per-tenant `max_bytes`, but it does not validate quota policy against master capacity. A tenant can be configured with a quota larger than the memory registered in the master. In that case, quota admission may pass while allocation fails later due to physical capacity.

PR #2112 also accounts bytes by `tenant_id` while leaving object metadata keyed by the unscoped user key. If two tenants use the same key, one tenant can overwrite, upsert, or remove the other tenant's object unless an upper layer always adds a tenant prefix. Mooncake should not depend on every caller to implement this correctly.

Mooncake should own both layers: resource allocation and object identity. Operators should be able to request tenant quotas, while Mooncake derives enforceable quotas from actual allocatable capacity. Clients should be able to use normal object keys, while Mooncake scopes those keys by tenant inside the master.

## Definitions

- `requested_quota_bytes`: The quota requested by an operator or control plane for a tenant.
- `effective_quota_bytes`: The quota Mooncake actually enforces after considering current allocatable memory.
- `allocatable_capacity_bytes`: Registered memory capacity available for tenant allocation, excluding system-reserved capacity.
- `used_bytes`: Committed live memory bytes for a tenant.
- `reserved_bytes`: Bytes reserved by in-flight `PutStart` operations.
- `over_quota`: A tenant state where `used_bytes + reserved_bytes > effective_quota_bytes`.

## Allocation Model

Mooncake stores both requested and effective quota per tenant. Admission uses only `effective_quota_bytes`.

When total requested quota fits into current capacity:

```text
if sum(requested_quota_bytes) <= allocatable_capacity_bytes:
    effective_quota_i = requested_quota_i
```

When total requested quota exceeds capacity, Mooncake scales each tenant proportionally:

```text
[...]

PR #2612, opened 2026-06-25
[Store] feat: implements strict multi-tenant quota admission

Lin-z-w:
## Description

Part of https://github.com/kvcache-ai/Mooncake/issues/2153. Implements strict multi-tenant quota admission for Mooncake Store.

This PR replaces the inherited/default tenant quota model with a connector-backed policy source:

- Adds `enable_multi_tenants`, `tenant_quota_connector_type`, and `tenant_quota_connector_uri`.
- Adds `TenantQuotaPolicyStore` and a v1 YAML file connector with atomic writes.
- Requires write/allocate requests to use an explicitly registered tenant when multi-tenant mode is enabled.
- Removes implicit default quota inheritance and the default quota admin endpoint.
- Persists admin tenant quota changes to the connector before applying them in memory.
- Restores runtime metadata from snapshots while loading tenant policy from the connector as source of truth.
- Creates orphan in-memory quota state for metadata tenants missing from policy, allowing reads/removes while blocking new writes.
- Fixes tenant quota policy update/delete races, YAML round-trip safety, and tenant eviction shard scan fairness.

## Module

- [x] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [x] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
cmake --build build --target tenant_quota_test master_service_tenant_quota_test
ctest --test-dir build -R "tenant_quota_test|master_service_tenant_quota_test" --output-on-failure
```

**Test results:**
- [x] Unit tests pass

`tenant_quota_test` and `master_service_tenant_quota_test` passed locally.

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex was used to help implement, review, test, and document this change. I reviewed and edited the generated code before submission.

### DIFF
--- a/docs/source/design/mooncake-store.md
+++ b/docs/source/design/mooncake-store.md
@@ -93,33 +93,43 @@ To reduce cache warm-up time after a master restart, the Master Service supports
 
 ### Tenant Quota
 
-The Master Service can optionally enforce memory quota admission per tenant. This feature is disabled by default. When `enable_tenant_quota=false`, existing allocation and eviction behavior is preserved, and tenant quota management requests return `UNAVAILABLE_IN_CURRENT_MODE`.
+The Master Service can optionally enforce strict multi-tenant memory quota admission. This feature is disabled by default. When `enable_multi_tenants=false`, request tenant IDs are ignored for object placement, all objects use the `default` namespace, and tenant quota management requests return `UNAVAILABLE_IN_CURRENT_MODE`.
 
-When tenant quota is enabled, each tenant has a requested quota policy and an effective quota. Explicit tenant policies are set through the master admin HTTP API. Tenants without an explicit policy inherit the default requested quota. The default policy may be `0`; explicit tenant policies must be positive.
+When strict multi-tenant mode is enabled, the tenant quota policy is loaded from the configured connector. The v1 connector is a writable YAML file configured by `tenant_quota_connector_type=file` and `tenant_quota_connector_uri=<path>`. Tenants must be explicitly present in that connector policy before they can write. Missing tenants, empty tenants, and an unregistered `default` tenant are rejected with `TENANT_NOT_REGISTERED`.
 
-Effective quota is recomputed from the current registered memory capacity and the optional tenant quota pool cap:
+The YAML policy uses schema version `1`:
 
-- The allocatable capacity is the smaller of total registered memory and `tenant_quota_pool_capacity_bytes`; a pool cap of `0` means total registered memory.
-- If explicit tenant requests fit within the allocatable capacity, explicit tenants receive their requested quotas and the remaining capacity is split evenly among active inherited-default tenants.
-- If explicit tenant requests exceed the allocatable capacity, only explicit tenants receive quota, scaled proportionally by request size. Inherited-default tenants receive `0` effective quota until capacity is available.
+```yaml
+version: 1
+
+tenants:
+  - name: tenant-a
+    quota: 200GB
+```
+
+Tenant names must be non-empty, unique, must not start with `_`, and must not contain NUL or control characters. Quotas must be positive integers and may use `B`, `KB`, `MB`, `GB`, or `TB` units.
+
+Effective quota is recomputed from the current registered memory capacity:
+
+- If explicit tenant requests fit within the registered memory capacity, tenants receive their requested quotas and remaining capacity stays unallocated.
+- If explicit tenant requests exceed registered memory capacity, explicit tenants receive quota scaled proportionally by request size.
 - Remainders are assigned deterministically by tenant ID, so repeated recomputes produce stable results.
+- Tenants present in restored metadata but missing from the connector policy become in-memory orphans with requested quota `0`, effective quota `0`, and `over_quota=true` when they still own memory. Reads and removals are allowed so operators can clean them up; writes remain blocked until the tenant is re-registered or emptied.
 
 `PutStart` and size-changing `UpsertStart` charge quota before memory is allocated. If the first reservation fails, the master performs tenant-scoped memory eviction for the target tenant and retries the reservation. The retry is bounded to two eviction attempts. Tenant quota eviction scans only the target tenant, skips hard-pinned objects, honors soft-pin eviction configuration, and preserves grouped-object lease safety checks.
 
-The admin HTTP API exposes:
+Admin policy changes are persisted before the final in-memory policy is applied. `PUT` writes the connector first and then applies the policy in memory. `DELETE` first marks the tenant unregistered in memory to block concurrent writes, verifies the tenant is empty, writes the connector, and rolls back the in-memory mark if the connector write fails. The admin HTTP API exposes:
 
 | Method | Path | Description |
 |--------|------|-------------|
 | `GET` | `/api/v1/tenant_quotas` | List quota snapshots for active or explicit tenants |
 | `GET` | `/api/v1/tenant_quotas?tenant_id=<tenant>` | Query one tenant quota snapshot |
-| `PUT` | `/api/v1/tenant_quotas?tenant_id=<tenant>` | Upsert an explicit tenant quota policy |
-| `DELETE` | `/api/v1/tenant_quotas?tenant_id=<tenant>` | Delete an explicit tenant quota policy |
-| `GET` | `/api/v1/tenant_quotas/default` | Query the default requested quota policy |
-| `PUT` | `/api/v1/tenant_quotas/default` | Set the default requested quota policy |
+| `PUT` | `/api/v1/tenant_quotas?tenant_id=<tenant>` | Create or update a tenant quota policy |
+| `DELETE` | `/api/v1/tenant_quotas?tenant_id=<tenant>` | Delete an empty tenant quota policy |
 
 Tenant quota snapshots include `tenant_id`, `requested_quota_bytes`, `effective_quota_bytes`, `used_bytes`, `reserved_bytes`, `committed_count`, `over_quota`, and `has_explicit_policy`.
 
-When snapshots are enabled, tenant quota policies are written to a separate `tenant_quota_policy` snapshot object. This file stores only requested policy: the default requested quota and explicit tenant requested quotas. Effective quota, usage, reservations, and committed object charge are rebuilt from restored metadata and current capacity. Older snapshots without `tenant_quota_policy` remain valid and use the configured default quota on restore.
+Snapshots restore object runtime state only. Tenant quota policy is always loaded from the connector after metadata restore, then usage and effective quota are rebuilt from restored metadata and current registered capacity. If the connector cannot be loaded in strict multi-tenant mode, startup fails.
 
 ### Master Service APIs
 
--- a/docs/source/design/mooncake-store.md
+++ b/docs/source/design/mooncake-store.md
@@ -127,7 +127,7 @@ Admin policy changes are persisted before the final in-memory policy is applied.
 | `PUT` | `/api/v1/tenant_quotas?tenant_id=<tenant>` | Create or update a tenant quota policy |
 | `DELETE` | `/api/v1/tenant_quotas?tenant_id=<tenant>` | Delete an empty tenant quota policy |
 
-Tenant quota snapshots include `tenant_id`, `requested_quota_bytes`, `effective_quota_bytes`, `used_bytes`, `reserved_bytes`, `committed_count`, `over_quota`, and `has_explicit_policy`.
+Tenant quota snapshots include `tenant_id`, `requested_quota_bytes`, `effective_quota_bytes`, `used_bytes`, `reserved_bytes`, `committed_count`, `metadata_object_count`, `over_quota`, and `has_explicit_policy`.
 
 Snapshots restore object runtime state only. Tenant quota policy is always loaded from the connector after metadata restore, then usage and effective quota are rebuilt from restored metadata and current registered capacity. If the connector cannot be loaded in strict multi-tenant mode, startup fails.
 
--- a/docs/source/design/mooncake-store.md
+++ b/docs/source/design/mooncake-store.md
@@ -114,7 +114,7 @@ Effective quota is recomputed from the current registered memory capacity:
 - If explicit tenant requests fit within the registered memory capacity, tenants receive their requested quotas and remaining capacity stays unallocated.
 - If explicit tenant requests exceed registered memory capacity, explicit tenants receive quota scaled proportionally by request size.
 - Remainders are assigned deterministically by tenant ID, so repeated recomputes produce stable results.
-- Tenants present in restored metadata but missing from the connector policy become in-memory orphans with requested quota `0`, effective quota `0`, and `over_quota=true` when they still own memory. Reads and removals are allowed so operators can clean them up; writes remain blocked until the tenant is re-registered or emptied.
+- Tenants present in restored metadata but missing from the connector policy become in-memory orphans with requested quota `0`, effective quota `0`, and `over_quota=true` while they still own metadata. Reads and removals are allowed so operators can clean them up; writes remain blocked until the tenant is re-registered or emptied.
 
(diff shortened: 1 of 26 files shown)


mooncake-store/src/master_service.cpp
```diff
@@ -793,6 +834,135 @@ size_t MasterService::getTenantQuotaShardIndex(
...
+                continue;
+            }
+            auto& state = shard.tenants[tenant_id];
+            state.requested_quota_bytes = requested_quota_bytes;
+            state.has_explicit_policy = true;
+            RefreshTenantQuotaOverQuota(state);
+        }
+    }
+    RecomputeTenantEffectiveQuotas();
+}
```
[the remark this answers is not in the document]
Lin-z-w: Addressed in the latest head. `ApplyTenantQuotaPolicies()` now groups policy entries by tenant quota shard before taking shard locks, so applying connector policy no longer scans the full policy map for every quota shard.

mooncake-store/src/master_service.cpp
```diff
@@ -550,81 +559,113 @@ std::optional<TenantQuotaSnapshot> MasterService::GetTenantQuotaSnapshot(
...
+    if (!save_result) {
+        restore_policy();
+        LOG(ERROR) << "failed to save tenant quota policy: "
+                   << save_result.error();
+        return tl::make_unexpected(ErrorCode::PERSISTENT_FAIL);
     }
-    RecomputeTenantEffectiveQuotas();
+    ApplyTenantQuotaPolicies(policy);
+    return GetTenantQuotaSnapshot(normalized_tenant);
 }
```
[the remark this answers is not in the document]
Lin-z-w: I did not apply this exact suggestion because checking metadata before clearing the explicit policy leaves a write-admission race: a writer can pass tenant validation, then create metadata after the emptiness check.

The delete path now first marks the tenant unregistered in memory under `tenant_quota_policy_mutex_`, then verifies quota/metadata emptiness, persists the connector update, and rolls back the in-memory mark if persistence fails. `AddReplica()` is also serialized with the same policy mutex from tenant admission through LOCAL_DISK metadata creation, closing the non-quota write race.

mooncake-store/src/master_service.cpp
```diff
@@ -877,9 +1041,6 @@ void MasterService::RecomputeTenantEffectiveQuotas() {
         for (auto it = shard.tenants.begin(); it != shard.tenants.end();) {
             const auto& tenant_id = it->first;
             auto& state = it->second;
-            if (!state.has_explicit_policy) {
-                state.requested_quota_bytes = default_requested_quota_bytes;
-            }
             if (IsLazyEmptyTenantQuotaState(state)) {
                 it = shard.tenants.erase(it);
                 continue;
```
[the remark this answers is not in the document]
Lin-z-w: Addressed. The non-explicit tenant cleanup now happens during the first shard scan in `RecomputeTenantEffectiveQuotas()`, so the extra final pass over all quota shards was removed.

mooncake-store/src/master_service.cpp
```diff
@@ -902,11 +1063,23 @@ void MasterService::RecomputeTenantEffectiveQuotas() {
...
+        std::lock_guard<std::mutex> lock(shard.mutex);
+        for (auto& [tenant_id, state] : shard.tenants) {
+            if (!state.has_explicit_policy) {
+                state.requested_quota_bytes = 0;
+                state.effective_quota_bytes = 0;
+                RefreshTenantQuotaOverQuota(state);
+            }
+        }
+    }
 }
```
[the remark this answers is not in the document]
Lin-z-w: Addressed together with the related recompute cleanup comment. Non-explicit tenants are normalized to requested/effective quota 0 while building the active tenant set, and the separate cleanup pass has been removed.

mooncake-store/src/master_service.cpp
yokinoshitayoki, line 2617: This leaves a race for zero-charge writes. When `replica_num == 0 && nof_replica_num > 0`, `reserved_quota_charge` is 0, so `ReserveTenantQuota()` only validates registration and does not leave any reserved state behind. A concurrent `DeleteTenantQuotaPolicy()` can then observe the tenant as empty, delete the policy, and this PutStart can still create metadata afterward at line 2760. That violates the strict registration guarantee and can create a new orphan object after the tenant was deleted.

Can we either keep the tenant policy lock held until metadata_object_count is incremented for zero-charge creates, or add a pending write/metadata marker that makes delete see the tenant as non-empty?

mooncake-store/src/master_service.cpp
```diff
@@ -4631,7 +5021,11 @@ auto MasterService::PromotionAllocStart(
     uint64_t size, const std::vector<std::string>& preferred_segments)
     -> tl::expected<PromotionAllocStartResponse, ErrorCode> {
     std::shared_lock<std::shared_mutex> shared_lock(snapshot_mutex_);
```
yokinoshitayoki: This introduces an inconsistent lock order. `AddReplica()` takes `tenant_quota_policy_mutex_` before `snapshot_mutex_`, but this path takes `snapshot_mutex_` first and then calls `NormalizeTenantIdForWrite()`, which takes `tenant_quota_policy_mutex_`. The same pattern also appears in `CreateCopyTask()` and `CreateMoveTask()`.

With a pending snapshot writer, this can deadlock on writer-preferring shared_mutex implementations: one thread holds the policy mutex and waits for a snapshot shared lock, while this thread holds a snapshot shared lock and waits for the policy mutex. Can we move tenant normalization before acquiring `snapshot_mutex_`, as `PutStart`, `CopyStart`, and `MoveStart` already do?

mooncake-store/src/master_service.cpp
Lin-z-w: addressed in the latest revision. The three functions now normalize the tenant before taking `snapshot_mutex_`, so they no longer acquire `tenant_quota_policy_mutex_` while holding `snapshot_mutex_`.

mooncake-store/src/master_service.cpp
Lin-z-w: Agreed. I kept the zero-charge path protected by holding `tenant_quota_policy_mutex_` only around the metadata-create admission window, and used `metadata_object_count` to make `DeleteTenantQuotaPolicy` observe zero-charge objects. This avoids holding the policy mutex through quota accounting itself while still preventing the tenant-delete race.

mooncake-store/src/master_service.cpp
```diff
@@ -4362,7 +4773,13 @@ auto MasterService::NotifyOffloadSuccess(
     for (size_t i = 0; i < tasks.size(); ++i) {
         const auto& task = tasks[i];
         const auto& metadata = metadatas[i];
-        const auto object_id = MakeObjectIdentity(task.key, task.tenant_id);
+        auto normalized_tenant_result =
+            NormalizeTenantIdForWrite(task.tenant_id);
```
yokinoshitayoki: This rejects completion of an offload that was already queued for an existing tenant before its policy disappeared from the connector. After `LoadTenantQuotaPoliciesFromStoreOrThrow()` + `RebuildTenantQuotaUsageFromMetadata()`, the PR intentionally keeps metadata tenants missing from the connector as orphan states so reads/removes can clean them up. But this callback is an internal lifecycle/cleanup path: returning `TENANT_NOT_REGISTERED` here happens before clearing `offloading_tasks`, dropping the source refcnt, and adding the LOCAL_DISK replica. That can leave an existing orphan object stuck until the reaper runs, or lose a successful offload completion entirely.

Can we resolve the object with `MakeObjectIdentityForRequest()` first and only enforce registration when `NotifyOffloadSuccess` would create a brand-new object, while allowing completion/cleanup for an existing orphan object?

Lin-z-w: Addressed in the latest revision.

`NotifyOffloadSuccess()` now first resolves the request with `MakeObjectIdentityForRequest()` and checks whether the object already exists. For an existing object, it allows the internal offload completion path even if the tenant is now an orphan state: it clears the offloading task, drops the source refcnt, and adds/updates the LOCAL_DISK replica.

mooncake-store/src/master_service.cpp
```diff
@@ -4362,16 +4773,27 @@ auto MasterService::NotifyOffloadSuccess(
...
 
-        // Release refcnt and clear offloading task.
+        Replica replica(client_id, metadata.data_size,
+                        metadata.transport_endpoint, ReplicaStatus::COMPLETE);
+        bool handled_existing_object = false;
         {
-            MetadataAccessorRW accessor(this, object_id);
+            std::shared_lock<std::shared_mutex> shared_lock(snapshot_mutex_);
+            MetadataAccessorRW accessor(this, request_object_id);
             if (accessor.Exists()) {
```
yokinoshitayoki, line 4786: This now lets an unregistered/orphan tenant add or update a LOCAL_DISK replica for any existing object, even when there is no matching `offloading_tasks` entry. The fallback registration check only runs when `handled_existing_object` stays false, but this path sets it true based on `accessor.Exists()` alone; the new test also exercises this path without ever queuing an offload task.

That seems broader than the intended orphan cleanup path and conflicts with the documented contract that orphan tenants allow reads/removes while writes remain blocked. Can we only bypass `NormalizeTenantIdForWrite()` when there is an in-flight offload task for this key, or otherwise prove this completion was master-admitted, and keep the registration check for ad-hoc `NotifyOffloadSuccess` calls on existing orphan objects?

Lin-z-w: Addressed in the latest revision.

I tightened the orphan bypass so `NotifyOffloadSuccess()` only skips `NormalizeTenantIdForWrite()` when the existing object has a matching `offloading_tasks` entry, i.e. the completion was previously admitted by the master. If the object exists but there is no offload-task marker, the path now falls through to the regular tenant registration check, so orphan/unregistered tenants cannot use ad-hoc `NotifyOffloadSuccess()` calls to add or update LOCAL_DISK replicas.

docs/source/design/mooncake-store.md
```diff
@@ -93,33 +93,43 @@ To reduce cache warm-up time after a master restart, the Master Service supports
...
 | `GET` | `/api/v1/tenant_quotas` | List quota snapshots for active or explicit tenants |
 | `GET` | `/api/v1/tenant_quotas?tenant_id=<tenant>` | Query one tenant quota snapshot |
-| `PUT` | `/api/v1/tenant_quotas?tenant_id=<tenant>` | Upsert an explicit tenant quota policy |
-| `DELETE` | `/api/v1/tenant_quotas?tenant_id=<tenant>` | Delete an explicit tenant quota policy |
-| `GET` | `/api/v1/tenant_quotas/default` | Query the default requested quota policy |
-| `PUT` | `/api/v1/tenant_quotas/default` | Set the default requested quota policy |
+| `PUT` | `/api/v1/tenant_quotas?tenant_id=<tenant>` | Create or update a tenant quota policy |
+| `DELETE` | `/api/v1/tenant_quotas?tenant_id=<tenant>` | Delete an empty tenant quota policy |
 
 Tenant quota snapshots include `tenant_id`, `requested_quota_bytes`, `effective_quota_bytes`, `used_bytes`, `reserved_bytes`, `committed_count`, `over_quota`, and `has_explicit_policy`.
```
yokinoshitayoki: `metadata_object_count` is part of the HTTP response (`HttpTenantQuotaSnapshot`) and is also exported as `mooncake_tenant_quota_metadata_object_count`, but this snapshot field list omits it. Can we include `metadata_object_count` here as well, and update the deployment guide JSON/metrics list so the documented admin API matches the actual response?

Lin-z-w: Addressed. Added `metadata_object_count` to:
- `docs/source/design/mooncake-store.md` snapshot field list
- `docs/source/deployment/mooncake-store-deployment-guide.md` metrics list and JSON example

mooncake-store/src/master_service.cpp
```diff
@@ -4362,41 +4773,94 @@ auto MasterService::NotifyOffloadSuccess(
...
+                // marker, fall through to the regular registration check.
                 if (task_it != tenant_state.offloading_tasks.end()) {
                     auto source =
                         obj_metadata.GetReplicaByID(task_it->second.source_id);
                     if (source != nullptr) {
                         source->dec_refcnt();
                     }
                     tenant_state.offloading_tasks.erase(task_it);
+
+                    if (!obj_metadata.HasReplica(
```
yokinoshitayoki, line 4809: This existing-object completion path only checks whether the object has any `LOCAL_DISK` replica. If the existing disk replica belongs to another client, the `else` branch only tries to update replicas matching the current `client_id`, so this completion can add no replica while the code below still increments this client's `ssd_used_bytes`. Can we make this path track whether a new `LOCAL_DISK` replica was actually inserted, and only increment `ssd_used_bytes` on a real insertion (or adjust by delta for updates)?

Lin-z-w: Addressed. Added `added_new_local_disk_replica` flag that is set only when a new LOCAL_DISK replica is actually inserted (the `!HasReplica` branch in the existing-object path, and the `AddReplica` path for new objects). The `ssd_used_bytes` increment is now conditioned on this flag, so updates to an existing replica no longer overcount.

mooncake-store/src/master_service.cpp
```diff
@@ -4362,43 +4773,100 @@ auto MasterService::NotifyOffloadSuccess(
...
+                if (res.error() == ErrorCode::OBJECT_NOT_FOUND) {
+                    continue;
+                }
+                LOG(ERROR) << "Failed to add replica: error=" << res.error()
+                           << ", client_id=" << client_id
+                           << ", tenant_id=" << object_id.tenant_id
+                           << ", key=" << object_id.user_key;
+                return tl::make_unexpected(res.error());
+            }
+            added_new_local_disk_replica = true;
```
yokinoshitayoki: This still sets `added_new_local_disk_replica` after every successful `AddReplica` call, but `AddReplica` can also succeed when it only updates an existing `LOCAL_DISK` replica, or when another disk replica already exists and no replica matches the current `client_id`. In those cases the `ssd_used_bytes` increment below can still double-count. Could `AddReplica` return whether it actually inserted a new local-disk replica, and use that value here instead of unconditionally setting the flag?

Lin-z-w: Addressed in the latest revision.

`AddReplica()` now returns whether it actually inserted a new LOCAL_DISK replica. `NotifyOffloadSuccess()` uses that returned value to decide whether to increment `ssd_used_bytes`, instead of treating every successful `AddReplica()` call as a new insertion.

mooncake-store/tests/master_service_tenant_quota_test.cpp
```diff
@@ -39,605 +157,688 @@ class MasterServiceTenantQuotaTest : public ::testing::Test {
...
 
-    CommitQuota(service, "tenant-a", 50);
+    ASSERT_FALSE(result.has_value());
+    EXPECT_EQ(result.error(), ErrorCode::TENANT_NOT_REGISTERED);
+}
 
-    ExpectSameAccounting(before, Snapshot(service, "tenant-a"));
+TEST_F(MasterServiceTenantQuotaTest,
+       NotifyOffloadSuccessDoesNotCountAddReplicaUpdateAsNewDiskUsage) {
+    MasterService service(MakeConfig({{"tenant-a", 1000}}));
```
yokinoshitayoki: This test calls `MountLocalDiskSegment(..., true)`, but `MakeConfig({{"tenant-a", 1000}})` does not enable offload. As a result, `MountLocalDiskSegment` returns an error with "The offload functionality is not enabled", and `master_service_tenant_quota_test` currently fails. Please enable offload in this test's config, similar to the other offload-related tenant quota test.

yokinoshitayoki (approved):
LGTM.

PR #2613, opened 2026-06-25
[Store] Add RedisHelper for standalone Redis leader election

skygragon:
Description

  Add RedisHelper as a standalone Redis-based leader election backend, mirroring EtcdHelper's role for P2P HA mode. This
  eliminates the etcd dependency when OpLog already uses Redis/TBase.

  This is the first of two PRs split from the original monolithic Redis election commit:
  - PR1 (this): RedisHelper component + CMake + unit tests
  - PR2 (follow-up): Integration into MasterViewHelper, MasterServiceSupervisor, config/gflags, and integration tests

  Key design:
  - SET NX EX for atomic leader election
  - INCR on epoch key for monotonic versioning
  - Lua script for atomic lease renewal (check-then-EXPIRE)
  - Pub/Sub + fallback polling for watching leader key expiration
  - Conditional compilation via USE_REDIS (auto-detected by CMake find_library(hiredis))
  - Only uses standard Redis commands + Lua scripts — compatible with any Redis-compatible server

  Module

  - [ ] Transfer Engine (mooncake-transfer-engine)
  - [x] P2P Store (mooncake-p2p-store)
  - [ ] CI/CD
  - [ ] Docs
  - [ ] Other

  Type of Change

  - [ ] Bug fix
  - [x] New feature
  - [ ] Refactor
  - [ ] Breaking change
  - [ ] Documentation update
  - [ ] Performance improvement
  - [ ] Other

  How Has This Been Tested?

  Test commands:
  # Build with Redis support
  cmake .. -DCMAKE_BUILD_TYPE=Debug
  make -j$(nproc) redis_helper_test

  # Run tests (requires local Redis on 127.0.0.1:6379)
  ./mooncake-store/tests/redis_helper_test \
    --redis_endpoint=127.0.0.1:6379 \
    --redis_ttl_sec=2 \
    --cluster_id=test_helper

  Test results:
  - [x] Unit tests pass (7/7: Connect, ElectLeader+GetMasterView, KeepLeaderRenewsTTL, KeyExpiresAfterKeepLeaderStops,
  ElectLeaderWaitsForKeyExpiry, EpochMonotonicallyIncreases, CancelKeepAliveStopsPromptly)
  - [ ] Integration tests pass (covered in PR2)
  - [x] Manual testing done — Ran full build + test suite in dev container with redis-server, all 7 tests passed

  Checklist

  AI Assistance Disclosure

  - [ ] No AI tools were used

  Claude Code (Claude Fable 5) assisted with code implementation, unit test writing, code review, and CI integration.

### DIFF
--- a/mooncake-store/src/redis_helper.cpp
+++ b/mooncake-store/src/redis_helper.cpp
@@ -1,0 +1,616 @@
+#ifdef STORE_USE_REDIS
+
+#include "redis_helper.h"
+
+#include <glog/logging.h>
+
+#include <chrono>
+#include <cstring>
+#include <json/json.h>
+
+namespace mooncake {
+
+// ============================================================
+// Construction / Destruction
+// ============================================================
+
+RedisHelper::RedisHelper(const std::string& cluster_id,
+                         const std::string& redis_endpoint,
+                         const std::string& password, int db_index, int ttl_sec,
+                         int heartbeat_interval_sec)
+    : redis_endpoint_(redis_endpoint),
+      password_(password),
+      db_index_(db_index),
+      ttl_sec_(ttl_sec),
+      heartbeat_interval_sec_(heartbeat_interval_sec),
+      cluster_id_(cluster_id) {
+    std::string cid = cluster_id;
+    if (!cid.empty() && cid.back() != '/') {
+        cid += '/';
+    }
+    // Use {cid} with braces as hash tag for Redis cluster slot affinity
+    master_view_key_ = "mooncake:{" + cid + "}master_view";
+    master_epoch_key_ = "mooncake:{" + cid + "}master_epoch";
+    leader_event_channel_ = "mooncake:" + cid + "leader_event";
+    LOG(INFO) << "RedisHelper created, master_view_key=" << master_view_key_
+              << " epoch_key=" << master_epoch_key_
+              << " channel=" << leader_event_channel_ << " ttl=" << ttl_sec_
+              << "s"
+              << " heartbeat=" << heartbeat_interval_sec_ << "s";
+}
+
+RedisHelper::~RedisHelper() {
+    CancelKeepAlive();
+    if (subscribe_ctx_) {
+        redisFree(subscribe_ctx_);
+        subscribe_ctx_ = nullptr;
+    }
+    {
+        std::lock_guard<std::mutex> lock(election_mutex_);
+        if (election_ctx_) {
+            redisFree(election_ctx_);
+            election_ctx_ = nullptr;
+        }
+    }
+}
+
+// ============================================================
+// CreateConnection — common logic for Connect / polling
+// ============================================================
+
+redisContext* RedisHelper::CreateConnection() {
+    std::string host = "127.0.0.1";
+    int port = 6379;
+    auto colon_pos = redis_endpoint_.rfind(':');
+    if (colon_pos != std::string::npos) {
+        host = redis_endpoint_.substr(0, colon_pos);
+        try {
+            port = std::stoi(redis_endpoint_.substr(colon_pos + 1));
+        } catch (const std::exception& e) {
+            LOG(ERROR) << "Invalid Redis endpoint port: " << redis_endpoint_;
+            return nullptr;
+        }
+    } else if (!redis_endpoint_.empty()) {
+        host = redis_endpoint_;
+    }
+
+    struct timeval tv;
+    tv.tv_sec = connect_timeout_ms_ / 1000;
+    tv.tv_usec = (connect_timeout_ms_ % 1000) * 1000;
+
+    redisContext* ctx = redisConnectWithTimeout(host.c_str(), port, tv);
+    if (!ctx || ctx->err) {
+        LOG(ERROR) << "Failed to connect to Redis at " << host << ":" << port
+                   << " err=" << (ctx ? ctx->errstr : "null");
+        if (ctx) {
+            redisFree(ctx);
+        }
+        return nullptr;
+    }
+    redisSetTimeout(ctx, tv);
+
+    // Authenticate if password provided
+    if (!password_.empty()) {
+        redisReply* reply = (redisReply*)redisCommand(
+            ctx, "AUTH %b", password_.data(), password_.size());
+        if (!reply || reply->type == REDIS_REPLY_ERROR) {
+            LOG(ERROR) << "Redis AUTH failed";
+            if (reply) freeReplyObject(reply);
+            redisFree(ctx);
+            return nullptr;
+        }
+        freeReplyObject(reply);
+    }
+
+    // Select DB if not default
+    if (db_index_ != 0) {
+        redisReply* reply =
+            (redisReply*)redisCommand(ctx, "SELECT %d", db_index_);
+        if (!reply || reply->type == REDIS_REPLY_ERROR) {
+            LOG(ERROR) << "Redis SELECT " << db_index_ << " failed";
+            if (reply) freeReplyObject(reply);
+            redisFree(ctx);
+            return nullptr;
+        }
+        freeReplyObject(reply);
+    }
+
+    return ctx;
+}
+
+// ============================================================
+// Connect
+// ============================================================
+
+ErrorCode RedisHelper::Connect() {
+    // Election connection
+    {
+        std::lock_guard<std::mutex> lock(election_mutex_);
+        if (election_ctx_) {
+            redisFree(election_ctx_);
+            election_ctx_ = nullptr;
+        }
+        election_ctx_ = CreateConnection();
+        if (!election_ctx_) {
+            return ErrorCode::INTERNAL_ERROR;
+        }
+    }
+
+    // Subscribe connection (separate, as SUBSCRIBE blocks).
+    // Set a 1-second read timeout so that redisGetReply in WatchLeader's
+    // subscribe loop returns periodically, allowing the loop to check
+    // notified / cancel_requested_ flags even when no Pub/Sub message
+    // arrives (e.g. leader key expired without graceful handoff).
+    if (subscribe_ctx_) {
+        redisFree(subscribe_ctx_);
+        subscribe_ctx_ = nullptr;
+    }
+    subscribe_ctx_ = CreateConnection();
+    if (subscribe_ctx_) {
+        struct timeval sub_timeout = {1, 0};  // 1 second
+        redisSetTimeout(subscribe_ctx_, sub_timeout);
+    } else {
+        // Non-fatal: WatchLeader will fall back to polling
+        LOG(ERROR) << "Failed to create subscribe connection to Redis";
+    }
+
+    LOG(INFO) << "Connected to Redis";
+    return ErrorCode::OK;
+}
+
+// ============================================================
+// ElectLeader — blocks until this node wins election
+// ============================================================
+
+void RedisHelper::ElectLeader(const std::string& master_address,
+                              ViewVersionId& version, int& lease_id) {
+    while (true) {
+        {
+            std::lock_guard<std::mutex> lock(election_mutex_);
+            if (!election_ctx_) {
+                election_ctx_ = CreateConnection();
+                if (!election_ctx_) {
+                    LOG(ERROR) << "ElectLeader: connect failed, retry in 1s";
+                }
+            }
+        }
+        if (!election_ctx_) {
+            std::this_thread::sleep_for(std::chrono::seconds(1));
+            continue;
+        }
+
+        // Step 1: Check if a leader already exists
+        redisReply* reply = nullptr;
+        {
+            std::lock_guard<std::mutex> lock(election_mutex_);
+            reply = (redisReply*)redisCommand(election_ctx_, "GET %b",
+                                              master_view_key_.data(),
+                                              master_view_key_.size());
+        }
+
+        if (!reply) {
+            LOG(ERROR)
+                << "ElectLeader: GET failed (connection error), retry in 1s";
+            {
+                std::lock_guard<std::mutex> lock(election_mutex_);
+                Reconnect(election_ctx_);
+            }
+            std::this_thread::sleep_for(std::chrono::seconds(1));
+            continue;
+        }
+
+        if (reply->type == REDIS_REPLY_NIL) {
+            // No leader exists — try to elect ourselves
+            freeReplyObject(reply);
+
+            bool elected = false;
+            {
+                std::lock_guard<std::mutex> lock(election_mutex_);
+                elected = TryElectOnce(master_address, version);
+            }
+            if (elected) {
+                lease_id = next_lease_id_++;
+                LOG(INFO) << "ElectLeader: elected as leader, epoch=" << version
+                          << " lease_id=" << lease_id;
+                return;
+            }
+            // TryElectOnce failed (someone else won) — loop back and wait
+            continue;
+        }
+
+        if (reply->type == REDIS_REPLY_STRING) {
+            // A leader exists — watch until the key expires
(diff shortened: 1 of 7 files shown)


mooncake-store/src/redis_helper.cpp
```diff
@@ -0,0 +1,616 @@
...
+    LOG(INFO) << "RedisHelper created, master_view_key=" << master_view_key_
+              << " epoch_key=" << master_epoch_key_
+              << " channel=" << leader_event_channel_ << " ttl=" << ttl_sec_
+              << "s"
+              << " heartbeat=" << heartbeat_interval_sec_ << "s";
+}
+
+RedisHelper::~RedisHelper() {
+    CancelKeepAlive();
+    if (subscribe_ctx_) {
```
wanyue-wy, line 45: Add a comment here: when we later introduce threads for leader election and lease renewal, those threads must be properly joined.

mooncake-store/src/redis_helper.cpp
```diff
@@ -0,0 +1,616 @@
...
+    {
+        std::lock_guard<std::mutex> lock(election_mutex_);
+        if (election_ctx_) {
+            redisFree(election_ctx_);
+            election_ctx_ = nullptr;
+        }
+        election_ctx_ = CreateConnection();
+        if (!election_ctx_) {
+            return ErrorCode::INTERNAL_ERROR;
+        }
```
wanyue-wy, line 140: add error log here

mooncake-store/src/redis_helper.cpp
```diff
@@ -0,0 +1,616 @@
...
+        subscribe_ctx_ = nullptr;
+    }
+    subscribe_ctx_ = CreateConnection();
+    if (subscribe_ctx_) {
+        struct timeval sub_timeout = {1, 0};  // 1 second
+        redisSetTimeout(subscribe_ctx_, sub_timeout);
+    } else {
+        // Non-fatal: WatchLeader will fall back to polling
+        LOG(ERROR) << "Failed to create subscribe connection to Redis";
+    }
```
wanyue-wy, line 165: It is worth considering whether `subscribe_ctx_` should include logic for re‑establishing connections in WatchLeader or other code paths, because if establishing failed, it will never be established again

mooncake-store/src/redis_helper.cpp
```diff
@@ -0,0 +1,616 @@
...
+    return ErrorCode::OK;
+}
+
+// ============================================================
+// ElectLeader — blocks until this node wins election
+// ============================================================
+
+void RedisHelper::ElectLeader(const std::string& master_address,
+                              ViewVersionId& version, int& lease_id) {
+    while (true) {
```
wanyue-wy: Should we also check `cancel_requested_` in this function? Currently, it only cancels one watch round.

mooncake-store/src/redis_helper.cpp
```diff
@@ -0,0 +1,616 @@
...
+                election_ctx_ = CreateConnection();
+                if (!election_ctx_) {
+                    LOG(ERROR) << "ElectLeader: connect failed, retry in 1s";
+                }
+            }
+        }
+        if (!election_ctx_) {
+            std::this_thread::sleep_for(std::chrono::seconds(1));
+            continue;
+        }
```
wanyue-wy: Although accessing `election_ctx_` is safe under the current code structure, it would still be prudent to ensure that all accesses are guarded by the lock.

mooncake-store/src/redis_helper.cpp
```diff
@@ -0,0 +1,616 @@
...
+                LOG(INFO) << "ElectLeader: current leader=" << current_addr
+                          << " epoch=" << current_epoch << ", waiting...";
+            } else {
+                LOG(INFO) << "ElectLeader: leader key exists but unparsable, "
+                             "waiting...";
+            }
+
+            WatchLeader();  // Blocks until key expires or "vacant" received
+            continue;
+        }
```
wanyue-wy, line 251: I would suggest using a `switch (reply->type)` structure and encapsulating the handling logic for each case into a function. This would significantly improve code readability.

mooncake-store/src/redis_helper.cpp
```diff
@@ -0,0 +1,616 @@
...
+            freeReplyObject(reply);
+
+            std::string current_addr;
+            ViewVersionId current_epoch = 0;
+            if (ParseLeaderValue(current_value, current_addr, current_epoch)) {
+                LOG(INFO) << "ElectLeader: current leader=" << current_addr
+                          << " epoch=" << current_epoch << ", waiting...";
+            } else {
+                LOG(INFO) << "ElectLeader: leader key exists but unparsable, "
+                             "waiting...";
```
wanyue-wy: Is this an exceptional case? If so, we should log an error here.

mooncake-store/src/redis_helper.cpp
```diff
@@ -0,0 +1,616 @@
...
+            }
+
+            WatchLeader();  // Blocks until key expires or "vacant" received
+            continue;
+        }
+
+        // Unexpected reply type
+        freeReplyObject(reply);
+        LOG(ERROR) << "ElectLeader: unexpected reply type=" << reply->type;
```
wanyue-wy: The access to `reply->type` occurs after reply has already been freed — this is a use‑after‑free issue.

mooncake-store/src/redis_helper.cpp
```diff
@@ -0,0 +1,616 @@
...
+                                                  master_epoch_key_.data(),
+                                                  master_epoch_key_.size());
+
+    if (!reply || reply->type != REDIS_REPLY_INTEGER) {
+        LOG(ERROR) << "TryElectOnce: INCR failed";
+        if (reply) freeReplyObject(reply);
+        return false;
+    }
+    out_epoch = reply->integer;
+    freeReplyObject(reply);
```
wanyue-wy, line 275: It may be worth wrapping `redisCommand` in a function that returns a `guard` class to the caller, which would help prevent resource leaks of the `reply`.

mooncake-store/src/redis_helper.cpp
```diff
@@ -0,0 +1,616 @@
...
+    return false;
+}
+
+// ============================================================
+// WatchLeader — wait until leader key expires
+// ============================================================
+
+void RedisHelper::WatchLeader() {
+    // Fast path: SUBSCRIBE for leader event notification
+    if (subscribe_ctx_) {
```
wanyue-wy: ai suggestion: `subscribe_ctx_` is reused across `WatchLeader` calls, but a single `UNSUBSCRIBE` command only drains one reply from the socket. Any "message" frames that were buffered between the `SUBSCRIBE` and `UNSUBSCRIBE` remain in the queue. Consequently, when the next cycle issues redisCommand(`SUBSCRIBE`), it reads the first buffered `reply`—which could be an outdated message or an unsubscribe acknowledgment—rather than the expected subscribe acknowledgment.

wanyue-wy: Wrap this fast path logic into a dedicated function to improve readability.

mooncake-store/src/redis_helper.cpp
```diff
@@ -0,0 +1,616 @@
...
+        redisReply* reply = (redisReply*)redisCommand(
+            subscribe_ctx_, "SUBSCRIBE %b", leader_event_channel_.data(),
+            leader_event_channel_.size());
+
+        if (reply && reply->type == REDIS_REPLY_ARRAY && reply->elements >= 3 &&
+            reply->element[0]->type == REDIS_REPLY_STRING &&
+            strncmp(reply->element[0]->str, "subscribe", 9) == 0) {
+            freeReplyObject(reply);
+
+            std::atomic<bool> notified{false};
```
wanyue-wy: We should rename notified — it seems intended to indicate whether the leader is valid, but the current name does not reflect that intent.

mooncake-store/src/redis_helper.cpp
```diff
@@ -0,0 +1,616 @@
...
+            redisContext* polling_ctx = CreateConnection();
```
wanyue-wy: If polling_ctx creation fails, the polling_thread will spin in a meaningless infinite loop.

mooncake-store/src/redis_helper.cpp
```diff
@@ -0,0 +1,616 @@
...
+                    if (polling_ctx) {
+                        redisReply* r = (redisReply*)redisCommand(
+                            polling_ctx, "GET %b", master_view_key_.data(),
+                            master_view_key_.size());
+                        if (r) {
+                            if (r->type == REDIS_REPLY_NIL) {
+                                notified = true;
+                            }
+                            freeReplyObject(r);
+                        }
```
wanyue-wy: We should log a warning if r is null.

mooncake-store/src/redis_helper.cpp
```diff
@@ -0,0 +1,616 @@
...
+
+            // Subscribe loop: read messages until "vacant" or key gone.
+            // redisGetReply returns at least every 1s (subscribe_ctx_ timeout)
+            // so the loop can check notified/cancel_requested_ promptly.
+            while (!notified && !cancel_requested_) {
+                redisReply* msg = nullptr;
+                if (redisGetReply(subscribe_ctx_, (void**)&msg) != REDIS_OK) {
+                    // Timeout or connection error — re-check flags and retry
+                    continue;
+                }
```
wanyue-wy: If `subscribe_ctx_` is unhealthy, the loop will never attempt a reconnect; instead, it will just keep `continue` and spin.

mooncake-store/src/redis_helper.cpp
```diff
@@ -0,0 +1,616 @@
...
+        }
+    }
+    if (polling_ctx) redisFree(polling_ctx);
+}
+
+// ============================================================
+// KeepLeader — renew TTL via Lua script, block until lost
+// ============================================================
+
+void RedisHelper::KeepLeader(int lease_id) {
```
wanyue-wy, line 499: `lease_id` is useless

mooncake-store/src/redis_helper.cpp
```diff
@@ -0,0 +1,616 @@
...
+    keep_alive_running_ = true;
+    cancel_requested_ = false;
```
wanyue-wy: `KeepLeader` sets `cancel_requested_` to false, which creates a write‑write conflict with `CancelKeepAlive` — both functions write to the same variable without proper synchronization.

mooncake-store/src/redis_helper.cpp
```diff
@@ -0,0 +1,616 @@
...
+                    redisReply* check = (redisReply*)redisCommand(
+                        election_ctx_, "GET %b", master_view_key_.data(),
+                        master_view_key_.size());
+                    if (check && check->type == REDIS_REPLY_STRING &&
+                        check->len == our_value_.size() &&
+                        memcmp(check->str, our_value_.data(),
+                               our_value_.size()) == 0) {
+                        // Still ours — continue renewing
+                        freeReplyObject(check);
+                        renewed = true;
```
wanyue-wy: Ownership is confirmed here, but the TTL is not renewed. This opens the possibility of a new leader election while this node still believes it is the leader, leading to a split‑brain scenario.

mooncake-store/src/redis_helper.cpp
```diff
@@ -0,0 +1,616 @@
...
+// ============================================================
+// GetMasterView
+// ============================================================
+
+ErrorCode RedisHelper::GetMasterView(std::string& master_address,
+                                     ViewVersionId& version) {
+    std::lock_guard<std::mutex> lock(election_mutex_);
+    if (!election_ctx_) {
+        return ErrorCode::INTERNAL_ERROR;
+    }
```
wanyue-wy, line 603: add error log here

mooncake-store/src/redis_helper.cpp
```diff
@@ -0,0 +1,616 @@
...
+
+    redisReply* reply = (redisReply*)redisCommand(election_ctx_, "GET %b",
+                                                  master_view_key_.data(),
+                                                  master_view_key_.size());
+
+    if (!reply) {
+        return ErrorCode::INTERNAL_ERROR;
+    }
```
wanyue-wy, line 613: add error log here

mooncake-store/src/redis_helper.cpp
```diff
@@ -0,0 +1,616 @@
...
+        return ErrorCode::INTERNAL_ERROR;  // No leader
+    }
+
+    if (reply->type == REDIS_REPLY_STRING) {
+        std::string value(reply->str, reply->len);
+        freeReplyObject(reply);
+        if (ParseLeaderValue(value, master_address, version)) {
+            return ErrorCode::OK;
+        }
+        return ErrorCode::INTERNAL_ERROR;
```
wanyue-wy, line 628: add error log here

mooncake-store/src/redis_helper.cpp
```diff
@@ -0,0 +1,616 @@
...
+    }
+
+    freeReplyObject(reply);
+    return ErrorCode::INTERNAL_ERROR;
```
wanyue-wy, line 634: add error log here

mooncake-store/src/redis_helper.cpp
```diff
@@ -0,0 +1,616 @@
...
+                                                  master_view_key_.size());
+
+    if (!reply) {
+        return ErrorCode::INTERNAL_ERROR;
+    }
+
+    if (reply->type == REDIS_REPLY_NIL) {
+        freeReplyObject(reply);
+        return ErrorCode::INTERNAL_ERROR;  // No leader
+    }
```
wanyue-wy, line 619: 1. A warning‑level log should be added.

2. At present, the absence of a leader is being conflated with a Redis error, and external callers have no way to differentiate between these two conditions. This raises the question: will subsequent call sites be impacted by this ambiguity?

skygragon: Added the warning log. Regarding the return code ambiguity: EtcdHelper::GetMasterView also maps "no leader" to INTERNAL_ERROR, and all callers just check != OK. Changing this for RedisHelper alone would introduce inconsistency — it should be a unified API change across both helpers, best tracked as a follow-up.

mooncake-store/src/redis_helper.cpp
skygragon: Not adding this — no threads are managed here currently. WatchLeader's polling thread is already joined within its scope (line 364), and KeepLeader/ElectLeader run on the caller's thread. We can add the comment when we actually introduce internal threads.

mooncake-store/src/redis_helper.cpp
skygragon: Skipping this — only 3 branches with short bodies (5-15 lines each), and each depends on local loop control (continue/return). A switch + helper functions would add indirection and parameter passing without improving readability for such a small, familiar pattern.

mooncake-store/src/redis_helper.cpp
skygragon: Good idea, but out of scope for this PR — it would touch every redisCommand call site (10+). I'll track it as a follow-up to introduce a RedisReplyGuard RAII wrapper that frees on destruction. The use-after-free fix just committed is the minimal safety fix for now.

mooncake-store/src/redis_helper.cpp
skygragon: cancel_requested_ is std::atomic<bool>, so the writes are already properly synchronized — no data race. The pattern is intentional: KeepLeader resets it to false at the start of a new election cycle, and CancelKeepAlive sets it to true to signal shutdown. With atomic, the last-writer-wins semantics are well-defined and the while loop observes the update on the next iteration.

PR #2614, opened 2026-06-25
[Store] Fix cross-process bucket_id collision on shared offload direc…

leonzzhu:
## Description

Fixes a **cross-process `bucket_id` collision** in `BucketIdGenerator` that silently corrupts offload data when multiple worker processes share the same on-disk bucket directory (e.g. multiple inference engines, tensor-parallel ranks, or multi-tenant clients running on the same host).

### Root cause

The previous fresh-id formula in `BucketIdGenerator::BucketIdGenerator(int64_t start)` is:

```cpp
current_id_ = (time_gen() << SEQUENCE_BITS) | 0;
```

It is keyed **only on wall-clock seconds**, so any two processes whose constructors run in the same second start from the **identical `current_id_`**. Each process then independently calls `fetch_add(1)` on its own atomic sequence counter, so all of them issue the same `bucket_id`s in lock-step and `O_TRUNC` each other's `<id>.bucket` files in the shared offload directory.

### Production symptoms

We observed the following on a host running **8 inference worker processes** that share the same offload directory (`MOONCAKE_OFFLOAD_FILE_STORAGE_PATH`):

| # | Log signature | Where | Meaning |
|---|---|---|---|
| 1 | `Read size mismatch, expected: <N>, got: 0` | client peer-read path | A later writer with a shorter payload had `O_TRUNC`'d an earlier writer's bucket file; the master still pointed at the old length, so the read returned 0 bytes for the tail. |
| 2 | `Failed to open file ..., errno=2` (`ENOENT`) | `storage_backend.cpp` open path | Master metadata referenced a `bucket_id` that one process had created, but another process had since replaced the underlying file with one whose layout no longer contained that key. |
| 3 | Same `bucket_id` in `[WriteBucket]` / `[Offload]` lines from multiple distinct threads with **different `data_size`** values | Master + worker logs cross-checked by PID | Two processes minted the same id within the same wall-clock second and both opened the same file with `O_TRUNC`. |
| 4 | `Bucket file size mismatch` warnings on restart scan | `ScanMeta` | A truncated-then-rewritten bucket file's header length disagreed with its on-disk size — direct evidence of an `O_TRUNC` race between processes. |

Verification that 8 processes were colliding at exactly the same second: a one-liner over their bucket filenames showed all 8 PIDs producing ids whose high bits encoded the **same** `time_gen()` value, confirming the constructors fired within the same 1-second window during the cluster's coordinated startup.

### Fix

Encode a **per-process salt** in the high bits of every bucket id so each process operates in its own disjoint id namespace.

New 64-bit layout:

```
bit  63        : 0 (sign, reserved)
bits [62:44]   : 19-bit process salt   (pid XOR random_device entropy)
bits [43:12]   : 32-bit time_gen seconds
bits [11:0]    : 12-bit fetch_add sequence
```

Key properties:

1. **Salt source.** The salt is computed by `ComputeProcessSalt()` as `pid XOR random_device()`, masked to 19 bits. PID alone is insufficient because two processes can share the low 19 bits of their PIDs; `random_device` mixes in OS entropy. If `random_device` throws (minimal container images without an entropy source), we fall back to `steady_clock` jitter — PID still differentiates processes in that path.
2. **Fresh path.** `current_id_ = (salt << (TIME_BITS + SEQUENCE_BITS)) | (time_gen() << SEQUENCE_BITS)`.
3. **Restart path** (`ScanMeta` recovered a non-zero `start`). Without a salt every process would see the same max_id on disk and re-collide. The new code applies the salt and **bumps the time component in a `while` loop** until the salted base strictly exceeds every id already on disk, guaranteeing the new generator never reissues an existing on-disk path.
4. **Observability.** Logs the chosen salt (hex) and starting id at `INFO`, so a `bucket_id` range can be attributed to a specific process from logs alone:
   ```
   [BucketIdGenerator] process_salt=0x3a17f, start_id=0x3a17f00000017c8a000
   ```

### Backwards compatibility

- Existing on-disk bucket files are untouched.
- `ScanMeta` continues to discover every file via directory scan; it does not parse the high bits of the id.
- The restart path guarantees `current_id > max_existing_id_on_disk`, so no existing path is ever overwritten by an upgraded binary.
- Single-process deployments are functionally unchanged; only the high bits of newly minted ids differ, and nothing besides the generator itself reads those bits.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

### Unit tests

Updated `mooncake-store/tests/storage_backend_test.cpp`:

| Test | Purpose |
|---|---|
| `InitializeWithInvalidStart_UseTimestampFallback` (updated) | Validate the new 19/32/12 bit layout: encoded second equals current wall-clock second, sequence starts at 0, salt is non-zero. |
| `ProcessSaltSeparatesGenerators` (new) | Construct two generators in the same process and verify they either land on different salts or otherwise produce strictly non-overlapping id ranges. |
| `RestartPathIsStrictlyGreaterThanRecoveredMax` (new) | Seed the generator with a synthetic on-disk max id and verify the bump-loop yields an id **strictly greater than** the recovered max, even when the salted base would otherwise be smaller. |

All pre-existing `BucketIdGenerator` tests (`NextId*`, `CurrentId*`, `Concurrency*`, `IdsAreUnique*`, `LargeNumberOfIds*`) pass unchanged.

**Test commands:**

```bash
cd Mooncake
mkdir -p build && cd build
cmake .. -DBUILD_UNIT_TESTS=ON
make storage_backend_test -j$(nproc)

./mooncake-store/tests/storage_backend_test \
  --gtest_filter='*Initialize*:*ProcessSalt*:*Restart*:*NextId*:*CurrentId*:*Concurrency*:*IdsAre*:*LargeNumber*'
```

**Expected:** all targeted cases `PASSED`.

### Manual / production validation

- Reproduced the original collision on an 8-worker host by starting all workers within the same second against a shared `MOONCAKE_OFFLOAD_FILE_STORAGE_PATH`; without the patch, `Read size mismatch ... got: 0` and `Failed to open file ... errno=2` reappeared within minutes.
- With the patch applied, scanned all 8 workers' INFO logs and confirmed **8 distinct `process_salt` values**; cross-checked all bucket filenames on disk and confirmed the high 19 bits partition cleanly into 8 disjoint ranges (zero overlap). No `Read size mismatch` / `errno=2` recurrence over an extended run.

### Checklist

- [x] Unit tests pass
- [x] Integration tests pass (manual multi-worker reproduction)
- [x] Manual testing done (8-process shared-directory scenario, described above)

## Checklist

- [ ] I have updated the documentation (if applicable) — N/A, behavior is transparent to users
- [x] For changes >500 LOC: I have filed an RFC issue — N/A, diff is ~140 LOC

## AI Assistance Disclosure

- [ ] No AI tools were used

AI assistance was used to:
- Draft the commit message and this PR description from a human-authored root-cause analysis and production logs.
- Suggest the 19/32/12 bit field split and the `while`-loop bump in the restart path.

The submitter has read, understood, and is responsible for every line of the diff and all test cases. All production symptoms quoted above are from real logs on the submitter's deployment.

### DIFF
--- a/mooncake-store/tests/storage_backend_test.cpp
+++ b/mooncake-store/tests/storage_backend_test.cpp
@@ -5,6 +5,7 @@
 
 #include <filesystem>
 #include <iostream>
+#include <limits>
 #include <ranges>
 #include <thread>
 #include <atomic>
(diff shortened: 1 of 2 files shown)


mooncake-store/tests/storage_backend_test.cpp
```diff
@@ -305,12 +306,112 @@ TEST_F(StorageBackendTest, InitializeWithValidStart) {
...
+    // Each instance independently rolls a random salt; collisions are
+    // possible (1 in 524288) but extraordinarily unlikely in test runs.
+    // Either the salts differ, or -- in the rare collision case -- the
+    // generators produce strictly monotonic ids on the same machine.
+    if (salt_a == salt_b) {
+        EXPECT_NE(gen_a.NextId(), gen_b.NextId());
+    } else {
+        SUCCEED() << "salts differ: " << salt_a << " vs " << salt_b;
+    }
+}
```
LujhCoconut: Overall the fix LGTM and is a nice fix. 

However, the `ProcessSaltSeparatesGenerators` test seems to have a logic issue: when the two generators happen to draw the same salt (≈1/524288), they share an identical `current_id_`, so both `NextId()` calls return the same value and `EXPECT_NE` fails. 

The comment says "two generators in the same process must produce non-overlapping id ranges", but that's not true when they share the same salt — their ranges fully overlap, and that's fine because in production each process only ever creates one generator. The cross-process isolation guarantee comes from `pid ^ random_device` giving different salts to different processes, which can't be meaningfully tested within a single process.

I'd suggest either removing this test or replacing the collision branch with a simple `LOG`+`SUCCEED()` so it doesn't become a flaky CI failure.

I'm confused with this. If I miss something, feel free to point out.

LujhCoconut:
A related follow-up (non-blocking for this PR): it looks like `ScanMeta`（`storage_backend.cpp`/`storage_backend.h`） may also need to be updated so that each process only loads buckets whose salt matches its own. Otherwise, after a restart, a worker could re-register buckets written by other processes under its own transport endpoint.

mooncake-store/tests/storage_backend_test.cpp
leonzzhu: Thanks @LujhCoconut, you're right on both counts and this was a real flaw in the test.

The test name `ProcessSaltSeparatesGenerators` and its comment overpromised: as you point out, when two in-process generators happen to draw the same salt (~1/524288), every field of their `current_id_` matches identically (same pid, same salt, same wall-clock second, same starting sequence), so `NextId()` necessarily returns the same value and `EXPECT_NE` would fail as a flaky false negative. And as you also note, the cross-process isolation that actually matters here cannot be observed within a single-process gtest run anyway — it comes from `pid ^ random_device` across distinct processes.

I've taken your second suggestion: the test is renamed to `ProcessSaltSmokeCheck`, both branches now `LOG(INFO)` + `SUCCEED()` so the test can never flake on a salt collision, and the comment is rewritten to explicitly disclaim cross-process semantics and point at the in-production end-to-end validation (8 workers logging 8 distinct salts) instead.

Pushed in the latest revision. PTAL.

leonzzhu:
> A related follow-up (non-blocking for this PR): it looks like `ScanMeta`（`storage_backend.cpp`/`storage_backend.h`） may also need to be updated so that each process only loads buckets whose salt matches its own. Otherwise, after a restart, a worker could re-register buckets written by other processes under its own transport endpoint.

Thanks for the catch. After thinking it through, this is more than a small follow-up — `pid ^ random_device` is intentionally re-rolled on every restart, so a naive same-salt filter would also drop a worker's *own* prior buckets. A real fix needs (1) a persistent per-worker identity, (2) a liveness signal to tell a dead foreign owner (safe to claim, preserves cross-restart KV-cache) from a live one (must skip), and (3) the salt filter gated on those.

Before this PR, shared offload directories collided at id-generation time, so the question never arose. Happy to track it as a separate PR after this lands.

LujhCoconut (approved):
I agree — persistent worker identity, liveness detection, and salt-aware ScanMeta are a non-trivial challenge. Looking forward to a separate PR that tackles this.

Please check the code format before merging.

PR #2616, opened 2026-06-25
Hca peer affinity

hzt123123:
## Description

This change adds an opt-in HCA peer affinity path for RDMA transfers in multi-NIC, multi-GPU deployments. It lets operators configure an explicit local-HCA to peer-HCA mapping, so RDMA slice dispatch can select a peer HCA that matches both the sender's local NIC and the target buffer's topology.

Default behavior is unchanged unless `MC_ENABLE_HCA_PEER_AFFINITY` is enabled.

### Assumptions

This feature is intended for homogeneous RDMA clusters. It assumes:

- all servers expose the same RDMA HCA names, such as `mlx5_0`, `mlx5_1`, and so on;
- all servers have the same intra-node topology, including GPU-to-HCA and NUMA-to-HCA relationships.

With these assumptions, the same `MC_GPU_HCA_AFFINITY` and `MC_NIC_PEER_AFFINITY` rules can be reused across servers.

#### `MC_GPU_HCA_AFFINITY`

Overrides the preferred HCA list for GPU topology entries discovered by Transfer Engine.

#### `MC_NIC_PEER_AFFINITY`

Describes which peer HCAs should be preferred for each local HCA.

For each RDMA slice, the sender uses its local HCA name and the target buffer location to select a peer device:

1. Find peer HCA candidates configured for the local HCA.
2. Intersect those candidates with the target location's preferred HCA list.
3. Choose among the remaining candidates using the existing random or round-robin policy.
4. Fall back to normal topology selection if no valid peer-affinity candidate is available.

#### `MC_ENABLE_HCA_PEER_AFFINITY`

Enables local-HCA based peer HCA selection.

This mode is mutually exclusive with `MC_ENABLE_DEST_DEVICE_AFFINITY`. If both are enabled, both affinity modes are disabled and Transfer Engine falls back to default peer device selection.

#### `MC_LOG_RDMA_SLICE_AFFINITY`

Enables optional per-slice RDMA affinity logs for validation. Each log line includes the source location, target location, local HCA, and selected peer HCA.

### Example

Assume a two-node setup where each node has four GPUs and eight RDMA HCAs. The desired fabric mapping is:

- local `mlx5_0` should target peer `mlx5_4` or `mlx5_5`;
- local `mlx5_1` should target peer `mlx5_6` or `mlx5_7`;
- peer GPU 0 should prefer `mlx5_4` and `mlx5_6`;
- peer GPU 1 should prefer `mlx5_5` and `mlx5_7`.

Configure the peer process topology:

```bash
export MC_GPU_HCA_AFFINITY="cuda:0=mlx5_4,mlx5_6;cuda:1=mlx5_5,mlx5_7"
```

Configure the sender process peer mapping:

```bash
export MC_ENABLE_HCA_PEER_AFFINITY=1
export MC_NIC_PEER_AFFINITY="mlx5_0=mlx5_4,mlx5_5;mlx5_1=mlx5_6,mlx5_7"
```

If a slice is sent through local `mlx5_0` to a target buffer resolved as `cuda:0`, the peer candidates are:

- mapping for local `mlx5_0`: `mlx5_4`, `mlx5_5`;
- preferred HCAs for target `cuda:0`: `mlx5_4`, `mlx5_6`;
- final candidate set: `mlx5_4`.

The slice is posted to peer `mlx5_4`. If the intersection is empty, the code falls back to the existing topology-aware selector.

### Performance Experiments

We also validated the change with `transfer_engine_bench` on two H200 servers. The results below compare the existing destination-device affinity path with the new HCA peer affinity modes.

Common benchmark settings:

- setup: two H200 servers, one initiator and one target;
- protocol: `rdma` with `P2PHANDSHAKE`;
- HCA list: 8 RDMA HCAs with consistent device names across both servers;
- VRAM transfer: `--use_vram=true`;
- buffer size: `268435456`;
- block size: `65536`;
- batch size: `16`;
- duration: `10` seconds;
- path policy: `MC_PATH_ROUNDROBIN=1`.

Test cases:

- existing destination-device affinity baseline;
- new HCA peer affinity with one-to-one local-to-peer HCA mapping;
- new HCA peer affinity with cross-rail local-to-peer HCA mapping.

#### Single GPU

Single-GPU runs use `GPU_ID=0` and `THREADS=1`.

| Case | Read throughput | Write throughput |
|------|-----------------|------------------|
| Existing destination-device affinity baseline | 18.24 GB/s | 18.35 GB/s |
| HCA peer affinity with one-to-one mapping | 18.40 GB/s | 18.45 GB/s |
| HCA peer affinity with cross-rail mapping | 16.55 GB/s | 16.77 GB/s |

#### 8 GPUs Concurrently

Eight-GPU runs use `GPU_ID=-1` and `THREADS=8`.

| Case | Read throughput | Write throughput |
|------|-----------------|------------------|
| Existing destination-device affinity baseline | 118.47 GB/s | 113.32 GB/s |
| HCA peer affinity with one-to-one mapping | 118.27 GB/s | 112.83 GB/s |
| HCA peer affinity with cross-rail mapping | 102.98 GB/s | 111.03 GB/s |

The one-to-one HCA peer affinity case is on par with the existing destination-device affinity baseline in both single-GPU and 8-GPU runs. The cross-rail case confirms that explicit local-to-peer HCA mapping is functional, with throughput depending on the selected fabric path.

## Module

- [✅] Transfer Engine (`mooncake-transfer-engine`)
- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [✅] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
```

**Test results:**
- [ ] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

### DIFF
--- a/mooncake-transfer-engine/include/config.h
+++ b/mooncake-transfer-engine/include/config.h
@@ -22,6 +22,7 @@
 #include <cstdint>
 #include <mutex>
 #include <string>
+#include <unordered_map>
 #include <utility>
 #include <vector>
 
@@ -60,6 +61,8 @@ struct GlobalConfig {
     bool use_ipv6 = false;
     size_t fragment_limit = 16384;
     bool enable_dest_device_affinity = false;
+    bool enable_hca_peer_affinity = false;
+    std::unordered_map<std::string, std::vector<std::string>> nic_peer_affinity;
     int parallel_reg_mr = -1;
     size_t eic_max_block_size = 64UL * 1024 * 1024;
     EndpointStoreType endpoint_store_type = EndpointStoreType::SIEVE;
--- a/mooncake-transfer-engine/include/config.h
+++ b/mooncake-transfer-engine/include/config.h
@@ -63,6 +63,7 @@ struct GlobalConfig {
     bool enable_dest_device_affinity = false;
     bool enable_hca_peer_affinity = false;
     std::unordered_map<std::string, std::vector<std::string>> nic_peer_affinity;
+    bool log_rdma_slice_affinity = false;
     int parallel_reg_mr = -1;
     size_t eic_max_block_size = 64UL * 1024 * 1024;
     EndpointStoreType endpoint_store_type = EndpointStoreType::SIEVE;
--- a/mooncake-transfer-engine/include/topology.h
+++ b/mooncake-transfer-engine/include/topology.h
@@ -30,6 +30,7 @@
 #include <string>
 #include <thread>
 #include <unordered_map>
+#include <vector>
 
 #include "common.h"
 
@@ -86,6 +87,9 @@ class Topology {
     int selectDevice(const std::string storage_type, int retry_count = 0);
     int selectDevice(const std::string storage_type, std::string_view hint,
                      int retry_count = 0);
+    int selectDeviceByLocalHca(const std::string storage_type,
+                               std::string_view local_hca,
+                               int retry_count = 0);
 
     TopologyMatrix getMatrix() const { return matrix_; }
 
@@ -121,6 +125,11 @@ class Topology {
     };
     std::unordered_map<std::string /* storage type */, ResolvedTopologyEntry>
         resolved_matrix_;
+    std::unordered_map<
+        std::string /* local HCA */,
+        std::unordered_map<std::string /* storage type */,
+                           std::vector<int> /* peer HCA ids */>>
+        resolved_hca_peer_affinity_by_local_;
 };
 
 }  // namespace mooncake
--- a/mooncake-transfer-engine/include/topology.h
+++ b/mooncake-transfer-engine/include/topology.h
@@ -88,8 +88,7 @@ class Topology {
     int selectDevice(const std::string storage_type, std::string_view hint,
                      int retry_count = 0);
     int selectDeviceByLocalHca(const std::string storage_type,
-                               std::string_view local_hca,
-                               int retry_count = 0);
+                               std::string_view local_hca, int retry_count = 0);
 
     TopologyMatrix getMatrix() const { return matrix_; }
 
@@ -125,10 +124,9 @@ class Topology {
     };
     std::unordered_map<std::string /* storage type */, ResolvedTopologyEntry>
         resolved_matrix_;
-    std::unordered_map<
-        std::string /* local HCA */,
-        std::unordered_map<std::string /* storage type */,
-                           std::vector<int> /* peer HCA ids */>>
+    std::unordered_map<std::string /* local HCA */,
+                       std::unordered_map<std::string /* storage type */,
+                                          std::vector<int> /* peer HCA ids */>>
         resolved_hca_peer_affinity_by_local_;
 };
 
--- a/mooncake-transfer-engine/include/transport/rdma_transport/rdma_transport.h
+++ b/mooncake-transfer-engine/include/transport/rdma_transport/rdma_transport.h
@@ -126,8 +126,7 @@ class RdmaTransport : public Transport {
                             std::string_view hint, int &buffer_id,
                             int &device_id, int retry_cnt = 0);
     static int selectDeviceByLocalHca(SegmentDesc *desc, uint64_t offset,
-                                      size_t length,
-                                      std::string_view local_hca,
+                                      size_t length, std::string_view local_hca,
                                       int &buffer_id, int &device_id,
                                       int retry_cnt = 0);
 
--- a/mooncake-transfer-engine/include/transport/rdma_transport/rdma_transport.h
+++ b/mooncake-transfer-engine/include/transport/rdma_transport/rdma_transport.h
@@ -125,6 +125,11 @@ class RdmaTransport : public Transport {
     static int selectDevice(SegmentDesc *desc, uint64_t offset, size_t length,
                             std::string_view hint, int &buffer_id,
                             int &device_id, int retry_cnt = 0);
+    static int selectDeviceByLocalHca(SegmentDesc *desc, uint64_t offset,
+                                      size_t length,
+                                      std::string_view local_hca,
+                                      int &buffer_id, int &device_id,
+                                      int retry_cnt = 0);
 
    private:
     std::vector<std::shared_ptr<RdmaContext>> context_list_;
--- a/mooncake-transfer-engine/include/transport/transport.h
+++ b/mooncake-transfer-engine/include/transport/transport.h
@@ -109,6 +109,7 @@ class Transport {
         TransferRequest::OpCode opcode;
         SegmentID target_id;
         std::string peer_nic_path;
+        std::string source_location;
         SliceStatus status;
         TransferTask *task;
         std::vector<uint32_t> dest_rkeys;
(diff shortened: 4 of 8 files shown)


staryxchen:
Thanks for the detailed write-up and the benchmark numbers — really helpful for understanding the proposal. Before approving, I'd like to push back on the necessity of both knobs.

1. MC_GPU_HCA_AFFINITY overlaps with MC_CUSTOM_TOPO_JSON
We already support full topology override via MC_CUSTOM_TOPO_JSON (transfer_engine_impl.cpp:236-251). The PR's example is equivalent to:
```
{"cuda:0": [["mlx5_4","mlx5_6"], ["mlx5_5","mlx5_7"]], ...}
```
 which is strictly more expressive (covers cpu/storage entries too) and already round-trips through Topology::parse().

2. MC_NIC_PEER_AFFINITY — what does it enable that MC_ENABLE_DEST_DEVICE_AFFINITY can't?
Dest-device-affinity already gives us "local mlx5_X → peer mlx5_X". The unique capability here is non-diagonal mappings. But your benchmark didn't show that such mapping is outperform than existing baseline. Do you have a concrete deployment where a non-diagonal mapping measurably beats dest-device-affinity? Without that, the only unique capability of this knob hasn't shown a benefit, and we can add it later when a real use case lands.

Happy to hear your thoughts — let me know what I'm missing.

hzt123123:
Thanks for your review. 

1. I agree. I removed `MC_GPU_HCA_AFFINITY` from the PR and now rely on `MC_CUSTOM_TOPO_JSON` for static CUDA topology overrides.

2. Yes, MC_ENABLE_DEST_DEVICE_AFFINITY covers the diagonal mapping case. MC_NIC_PEER_AFFINITY is mainly intended for newer topologies, such as ZCube, where the preferred local-to-peer NIC mapping may be non-diagonal.

I agree the current benchmark does not yet show a measurable win over the existing baseline. I can clarify the motivation and scope accordingly, and avoid presenting it as a proven performance improvement for the tested topology.

staryxchen:
> Thanks for your review.
> 
> 1. I agree. I removed `MC_GPU_HCA_AFFINITY` from the PR and now rely on `MC_CUSTOM_TOPO_JSON` for static CUDA topology overrides.
> 2. Yes, MC_ENABLE_DEST_DEVICE_AFFINITY covers the diagonal mapping case. MC_NIC_PEER_AFFINITY is mainly intended for newer topologies, such as ZCube, where the preferred local-to-peer NIC mapping may be non-diagonal.
> 
> I agree the current benchmark does not yet show a measurable win over the existing baseline. I can clarify the motivation and scope accordingly, and avoid presenting it as a proven performance improvement for the tested topology.

Thanks for the quick response, and the ZCube context makes sense as motivation.

Two follow-ups before we move forward:
1. Could you share ZCube benchmark numbers and a sketch of the topology? Even a small run on a representative ZCube setup would let us verify the non-diagonal mapping actually beats MC_ENABLE_DEST_DEVICE_AFFINITY in the environment it's designed for. Landing a knob whose unique capability is unverified in any deployment is a maintenance cost we'd carry indefinitely.

2. Once we have that, can we fold this into MC_CUSTOM_TOPO_JSON instead of a parallel env-driven path? The non-diagonal local→peer mapping is fundamentally a topology fact (which NIC sits on which leaf), so it belongs in the same place we already express topology. A unified schema also avoids the awkward mutual-exclusion with MC_ENABLE_DEST_DEVICE_AFFINITY — symmetric and asymmetric affinity become two configurations of the same mechanism, not two competing ones. Happy to brainstorm the schema with you once the ZCube data is in.

WDYT?

hzt123123:
> Two follow-ups before we move forward:
> 
> 1. Could you share ZCube benchmark numbers and a sketch of the topology? Even a small run on a representative ZCube setup would let us verify the non-diagonal mapping actually beats MC_ENABLE_DEST_DEVICE_AFFINITY in the environment it's designed for. Landing a knob whose unique capability is unverified in any deployment is a maintenance cost we'd carry indefinitely.
> 2. Once we have that, can we fold this into MC_CUSTOM_TOPO_JSON instead of a parallel env-driven path? The non-diagonal local→peer mapping is fundamentally a topology fact (which NIC sits on which leaf), so it belongs in the same place we already express topology. A unified schema also avoids the awkward mutual-exclusion with MC_ENABLE_DEST_DEVICE_AFFINITY — symmetric and asymmetric affinity become two configurations of the same mechanism, not two competing ones. Happy to brainstorm the schema with you once the ZCube data is in.
> 
> WDYT?

The ZCube cluster is currently serving production environment and is not available for performance testing. However, it did show performance improvements when it was previously running in production. You can refer to this blog: https://z.ai/blog/zcube.

We would also be glad to discuss a more consistent architectural design with you.

staryxchen:
> > Two follow-ups before we move forward:
> > 
> > 1. Could you share ZCube benchmark numbers and a sketch of the topology? Even a small run on a representative ZCube setup would let us verify the non-diagonal mapping actually beats MC_ENABLE_DEST_DEVICE_AFFINITY in the environment it's designed for. Landing a knob whose unique capability is unverified in any deployment is a maintenance cost we'd carry indefinitely.
> > 2. Once we have that, can we fold this into MC_CUSTOM_TOPO_JSON instead of a parallel env-driven path? The non-diagonal local→peer mapping is fundamentally a topology fact (which NIC sits on which leaf), so it belongs in the same place we already express topology. A unified schema also avoids the awkward mutual-exclusion with MC_ENABLE_DEST_DEVICE_AFFINITY — symmetric and asymmetric affinity become two configurations of the same mechanism, not two competing ones. Happy to brainstorm the schema with you once the ZCube data is in.
> > 
> > WDYT?
> 
> The ZCube cluster is currently serving production environment and is not available for performance testing. However, it did show performance improvements when it was previously running in production. You can refer to this blog: https://z.ai/blog/zcube.
> 
> We would also be glad to discuss a more consistent architectural design with you.

First, I read through the blog and the SIGCOMM paper and couldn't find a case where a non-diagonal local→peer NIC mapping is hop-count optimal. If I'm missing a passage that shows otherwise, please point me to it.

That said, I still think a more flexible affinity mechanism is worth having. Today MC_ENABLE_DEST_DEVICE_AFFINITY only expresses same-name pairing, which is a special case of a more general "user-declared local→peer affinity." Even without ZCube as the proof point, a unified knob is reasonable as long as it subsumes the existing one rather than sitting alongside it.

So my preference would be to express it through MC_CUSTOM_TOPO_JSON (extending the schema) instead of a separate env-driven path, and have MC_ENABLE_DEST_DEVICE_AFFINITY become one shorthand of the same mechanism rather than a mutually exclusive alternative.

alogfans (approved):
LGTM. For debugging purpose could you add VLOG(1) to reflect the binding result?

hzt123123:
> LGTM. For debugging purpose could you add VLOG(1) to reflect the binding result?

Thanks. I have replaced LOG(INFO) with VLOG(1). Enabling MC_LOG_RDMA_SLICE_AFFINITY enables log output, as shown in the figure below.
image

Issue #2519, opened 2026-06-17
[RFC]: Deadline-aware admission & adaptive degradation for TENT transfers


catyans:
> Verified against `kvcache-ai/Mooncake` `main` @ `94ac60df` (2026-06-17).
>
> **Builds on #2132** (TENT Local Transfer Admission Queue). That RFC explicitly defers priority/deadline policy to a later milestone ("*if a later queued submit path adds priority or deadline policy*", "*once retry budget or deadline is exhausted*"). This RFC proposes that deadline layer concretely. It is purely additive: it reuses #2132's `QueueOwner` model and dispatch boundary, changes nothing when no deadline is supplied, and does not touch the WDRR default or the worker priority queues. Filing it separately (rather than as a comment on #2132) only because it introduces a new request field and an external degradation hook that are easier to track as their own thread; happy to merge the discussion into #2132 if maintainers prefer.

## Motivation

TENT already prioritizes traffic well at the *ordering* level (documented in `docs/source/design/tent/qos.md`):

- per-request priority (`PRIO_HIGH/MEDIUM/LOW`, `tent/common/types.h`)
- per-priority slice queues with strict HIGH→LOW dequeue (`workers.h`, `workers.cpp`)
- starvation prevention via timed promotion (`promoteTimedOutRequests`, `workers.cpp`)
- cross-process coordination via time-sliced slots (`SharedSlotManager`, `shared_quota.h`)
- priority-aware device selection (`quota.h`, `enable_priority_filtering`)

All of this answers *"given a set of transfers, in what order do we send them?"* What's still missing is the SLO question:

> **Will this P→D transfer finish before its TTFT deadline — and if not, what do we do instead of pushing it through and timing out?**

Three concrete gaps:

1. **No per-transfer deadline.** `Request` has `priority` but no deadline, so the scheduler can't tell "HIGH with 8 ms left" from "HIGH with 80 ms left."
2. **Admission is capacity-only, FIFO.** `LocalTransferAdmissionQueue::pickForDispatch` (`admission_queue.cpp:185`) admits/dispatches purely on outstanding bytes/owners in FIFO order. It cannot reshape or reject a transfer that provably can't meet its deadline.
3. **No degradation path.** Under saturation the only outcomes are "send slower" or "time out." There's no wired-in "send a losslessly compressed version" or "skip transfer, recompute locally."

This is the SLO-guarantee capability implied by #1883 Milestone 3 that the ordering machinery alone cannot provide.

[...]

PR #2618, opened 2026-06-25
[TENT] Add optional Request.deadline_ns and MLU observability metric (RFC #2519 step 1)

catyans:
## Description

Minimal first step of RFC #2519 (deadline-aware admission), scoped to what
@alogfans approved:

> "LGTM for the minimal first step (deadline_ns field + MLU observability
> only). Let's see the MLU metrics on real workloads before adding policy
> logic."

This adds the deadline plumbing + an **observability-only** feasibility
metric. There is **no admission, scheduling, or degradation logic** — the
deadline does not affect any transfer decision yet.

### Changes
- **`types.h`**: add optional `uint64_t deadline_ns = 0` to `Request`. `0`
  (default) is bit-for-bit today's behavior; nothing reads it unless the
  caller sets it. It is an absolute `steady_clock` timestamp in ns.
- **`tent_metrics`**: add `recordDeadlineMLU(double)` and a `deadline_mlu`
  histogram. MLU (laxity feasibility ratio) is stored in **per-mille**
  (MLU × 1000) so it uses the same integer `observe()` as the other
  histograms; the `1000` boundary is `MLU == 1.0` (< 1000 met the deadline,
  ≥ 1000 missed). Both the enabled and the compile-time-disabled
  (`TENT_METRICS_ENABLED=0`) stubs are provided.
- **`transfer_engine_impl`**: at task completion, when the request carried a
  deadline, emit the post-hoc `MLU = actual_transfer_time / window`, where
  `window = deadline − submit_time`. Reuses the `start_time` /
  `latency_seconds` already computed at the completion point — no new
  measurement subsystem.

### Why post-hoc MLU (and not the predictive submit-time MLU from RFC §2)
Device pinning happens *after* submit, so a per-device EWMA-bandwidth
predictive MLU isn't available at submit time (this is open question #2 in
the RFC). The post-hoc ratio validates whether MLU is a useful signal on
real workloads first — exactly the "see the metrics before adding policy"
step requested. The predictive/admission layers are deliberate follow-ups.

## Module
- [x] Transfer Engine (`mooncake-transfer-engine`) — TENT runtime

## Type of Change
- [x] New feature (observability)

## How Has This Been Tested?
Additive and behavior-preserving when `deadline_ns == 0` (the default), which
is every existing caller. The MLU path only runs when a caller opts in by
setting a deadline, and only emits a histogram observation.

- [ ] Unit tests pass — N/A yet (observability-only; see note)
- [ ] Integration tests pass
- [ ] Manual testing done

> Notes:
> - Not built locally (no TENT toolchain on my dev machine); relying on CI.
> - Real-workload MLU numbers (what @alogfans asked to see) are still TODO —
>   they require building with `-DTENT_METRICS_ENABLED=ON` and a caller that
>   sets `deadline_ns`. Happy to follow up with measurements; opening the PR
>   now so the integration point can be reviewed in parallel.

## Checklist
- [x] I have formatted my code using clang-format (v20.1.8)
- [x] For changes >500 LOC: I have filed an RFC issue — RFC #2519

## AI Assistance Disclosure
- [x] AI tools were used (Claude Code) for code navigation and drafting.

### DIFF
--- a/mooncake-transfer-engine/tent/include/tent/common/types.h
+++ b/mooncake-transfer-engine/tent/include/tent/common/types.h
@@ -79,6 +79,12 @@ struct Request {
     TransportType transport_hint =
         UNSPEC;  // UNSPEC = follow policy; otherwise pin this request to the
                  // name transport.
+    // Optional SLO deadline as an absolute steady_clock timestamp in
+    // nanoseconds. 0 = no deadline (default), behaves exactly as today.
+    // When set, the engine emits an observability-only feasibility metric
+    // (MLU = actual transfer time / available window) on completion; it does
+    // not yet drive any admission or scheduling decision. See RFC #2519.
+    uint64_t deadline_ns = 0;
 };
 
 enum TransferStatusEnum {
--- a/mooncake-transfer-engine/tent/include/tent/metrics/tent_metrics.h
+++ b/mooncake-transfer-engine/tent/include/tent/metrics/tent_metrics.h
@@ -81,6 +81,12 @@ class TentMetrics {
     void recordWriteFailed(size_t bytes);
     void recordTransportFailover();
 
+    // Record the deadline feasibility ratio (MLU) for a completed transfer
+    // that carried a deadline. mlu = actual_transfer_seconds / window_seconds,
+    // where window_seconds is (deadline - submit_time). mlu < 1 means the
+    // transfer met its deadline; mlu >= 1 means it missed. Observability only.
+    void recordDeadlineMLU(double mlu);
+
     // Get metrics for HTTP server
     std::string getPrometheusMetrics();
     std::string getJsonMetrics();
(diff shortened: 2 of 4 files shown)


alogfans (approved):
LGTM

catyans:
Followed up with real-workload validation on an H20 box, since the ask was to "see the MLU metrics on real workloads before adding policy logic."

**Setup**: built this branch with `-DUSE_TENT=ON -DTENT_METRICS_ENABLED=ON`, ran `transfer_engine_bench -backend tent` P2P (single box, RoCE / mlx5), write op, with a small patch to pass `deadline_ns` per request and scrape `tent_deadline_mlu_permille` from `/metrics`. Two runs with deliberately different deadline windows:

| Run | deadline window | block | count | mean MLU | histogram |
|---|---|---|---|---|---|
| tight | 2 ms | 256 KB | 512 | ~472 | all 512 in `+Inf` (MLU ≫ 1) |
| loose | 5 s | 64 KB | 702 | ~0.075 | 497/702 in `le=0.1`, all ≤ 0.25 |

(MLU is stored per-mille, so the 1000 boundary = MLU 1.0.)

**Takeaway**: the same metric cleanly separates the two regimes — under a tight deadline every transfer lands MLU ≫ 1 (correctly "will miss"), and under a generous deadline nearly all land MLU ≪ 1 (correctly "comfortably feasible"). So MLU behaves as a meaningful, discriminating feasibility signal, which is what this step was meant to validate before any admission/scheduling policy is layered on.

**Caveats (honest)**: this is single-box P2P RoCE self-transfer, so the absolute bandwidth is not representative of production; the point here is the MLU mechanism and its discrimination, not throughput numbers. The bench-side `deadline_ns` plumbing is a local throwaway patch, not part of this PR.

PR #2504, opened 2026-06-16
[TE] Fix signed-char tolower UB in PCI BDF lowercasing loops (follow-up to #2367/#2488)

bp-cheng:
Fixes the follow-up requested by @jfeng18 in #2488.

## Description

#2367 fixed the signed-char `::tolower`/`::toupper` UB in the `std::transform` sites, and #2488 finished the remaining operator-controllable ones. This PR completes the sweep by converting the last form — the C-style PCI-BDF lowercasing loops — flagged as a non-blocking follow-up in #2488:

```cpp
for (char* ch = pci_bus_id; (*ch = tolower(*ch)); ch++);
```

Passing a `char` directly to `tolower` sign-extends bytes > 0x7F to a negative `int`, which is UB (the argument must be representable as `unsigned char` or equal `EOF`). Per @alogfans's review, a single inline helper `te_lower()` is added to `mooncake-transfer-engine/include/common.h` and used at every site, instead of repeating the cast at each call:

```cpp
// common.h
static inline char te_lower(char c) {
    return static_cast<char>(std::tolower(static_cast<unsigned char>(c)));
}
// call site
for (char* ch = pci_bus_id; (*ch = te_lower(*ch)); ch++);
```

7 sites across 6 files, all lowercasing a `pci_bus_id` produced by `cudaDeviceGetPCIBusId` / `snprintf("%04x:%02x:%02x.%x", ...)`:

| File | Site |
|---|---|
| `mooncake-transfer-engine/src/topology.cpp` | `discoverCudaTopology` |
| `mooncake-transfer-engine/tent/src/platform/cuda/cuda_probe.cpp` | `discoverCudaTopology` |
| `mooncake-transfer-engine/tent/src/platform/rocm/rocm_probe.cpp` | `discoverRocmTopology` |
| `mooncake-transfer-engine/tent/src/runtime/memory_prober.cpp` | `probeDeviceMemory` |
| `mooncake-transfer-engine/benchmark/te_backend.cpp` | `getCudaDeviceNumaID`, `getSunriseDeviceNumaID` |
| `mooncake-transfer-engine/benchmark/tent_backend.cpp` | `getGpuDeviceNumaID` |

[...]

PR #2488, opened 2026-06-15
[Store][TE] Fix remaining signed-char ::tolower/::toupper UB missed by #2367

bp-cheng:
Fixes #2487

## Description

Follow-up to #2367, which fixed UB from passing a (possibly signed) `char` to `::tolower`/`::toupper` via `std::transform`. Three sibling occurrences on operator/CLI-controlled input were missed and are still present on `main`:

| File | Input source |
|---|---|
| `mooncake-transfer-engine/src/transport/tcp_transport/tcp_transport.cpp:606` | env `MC_TCP_ENABLE_CONNECTION_POOL` |
| `mooncake-store/include/utils.h:225` | size-string unit suffix (env `MC_MMAP_ARENA_POOL_SIZE`, `global_segment_size` CLI flag, config map) |
| `mooncake-store/include/ha/oplog/oplog_store_factory.h:34` | HA `oplog_store_type` config string |

Passing a byte > 0x7F to `::tolower`/`::toupper` sign-extends to a negative `int`, which is UB (the argument must be representable as `unsigned char` or equal `EOF`). All three sites take operator/attacker-controllable bytes, not kernel-generated ASCII.

Each is fixed with the exact idiom from #2367 — `[](unsigned char c) { return std::tolower(c); }` — and `<cctype>` is added to the two files where it was only available transitively (`utils.h` already includes it). Notably `utils.h:274` already uses the safe lambda; line 225 was simply overlooked in the original sweep.

Behavior is unchanged for ASCII input; this only removes the UB on non-ASCII bytes.

## Module

- [x] Mooncake Store (`mooncake-store`)
- [x] Transfer Engine (`mooncake-transfer-engine`)

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

**Test commands:**
```bash
# Ubuntu 22.04 (aarch64) container, cmake -DWITH_STORE_RUST=OFF -DUSE_HTTP=ON, ninja
ninja
```

**Test results:**
- [x] Unit tests pass
- [ ] Integration tests pass (if applicable)
- [x] Manual testing done (describe below)

This is the same class of fix as #2367, which shipped with no new test: the change is behavior-preserving for ASCII, and the UB is not portably observable (on glibc the ctype tables span -128..255, so it does not crash there). Validation is compile-only plus the existing suite: a full incremental build succeeds (all three headers/TUs and their dependents recompile cleanly with the added `<cctype>` includes), and the existing `tcp_transport` / store tests pass.

## Checklist

[...]

PR #2619, opened 2026-06-25
[TE] Fix signed-char isxdigit UB in EFA smaps page-size parsing (follow-up to #2504)

he-yufeng:
## Description

`detectBufferPageSize()` in the EFA transport scans `/proc/self/smaps` and uses `std::isxdigit(line[0])` to recognize VMA header lines. `line[0]` is a plain `char` from `std::getline`; where `char` is signed, a byte `> 0x7F` becomes a negative `int`. Passing that to `std::isxdigit` is undefined behavior — the argument must be representable as `unsigned char` or equal `EOF`. The fix casts to `unsigned char` first.

This is the same class of bug fixed for the PCI BDF lowercasing loops in #2367 / #2488 / #2504 (which added `char_util.h::to_lower`). That sweep didn't reach `efa_transport.cpp` — this was the last remaining un-cast ctype call across `mooncake-transfer-engine` and `mooncake-store`. I kept it inline rather than routing through a helper, since it's a single classification site (`isxdigit`), unlike the multi-site case conversion that justified `to_lower()`. Happy to switch to a shared helper if you'd prefer consistency.

## Module

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

No unit test: `detectBufferPageSize()` is a file-static helper that reads `/proc/self/smaps` directly, so there's no seam to inject a crafted line without refactoring the function — the same reason #2504 shipped the `char_util.h` helper without a dedicated test. The change is verified by the C++ ctype contract (the `static_cast<unsigned char>` is the documented precondition) and is parity with the existing sweep. No behavior change for realistic input: smaps header lines start with an ASCII hex address and attribute lines with ASCII keywords, so the predicate's result is unchanged — this only removes the UB on the value passed in.

## Checklist

- [x] I have formatted my code using `./scripts/code_format.sh` (clang-format)

## AI Assistance Disclosure

Used Claude Code to help locate the remaining un-cast ctype site and draft the change. I reviewed and verified the fix and am responsible for it.

### DIFF
--- a/mooncake-transfer-engine/src/transport/efa_transport/efa_transport.cpp
+++ b/mooncake-transfer-engine/src/transport/efa_transport/efa_transport.cpp
@@ -59,8 +59,12 @@ static size_t detectBufferPageSize(void* addr) {
     bool in_range = false;
 
     while (std::getline(smaps, line)) {
-        // VMA header: "start-end perms offset dev inode [pathname]"
-        if (!line.empty() && std::isxdigit(line[0])) {
+        // VMA header: "start-end perms offset dev inode [pathname]".
+        // Cast to unsigned char before std::isxdigit: passing a (possibly
+        // signed) char whose value is > 0x7F is UB, since the argument must be
+        // representable as unsigned char or equal EOF.
+        if (!line.empty() &&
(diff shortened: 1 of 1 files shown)


whn09 (approved):
LGTM

PR #2625, opened 2026-06-25
[Bugfix] Gate RDMA sends until active side confirms QP readiness

LCAIZJ:
## Description
This PR fixes a race in RDMA connection setup where the passive side may post WRs before the active side has finished moving its local QPs to RTR/RTS.

In the old flow, setupConnectionsByPassive() marks the endpoint as CONNECTED immediately after the passive side finishes its own QP setup. However, the active side still needs to process the handshake response and complete its own QP state transition. If the passive side posts WRs in this window, they may hit a remote QP that is not ready yet

```mermaid
sequenceDiagram
    participant B as "B active side"
    participant A as "A passive side"

    B->>A: Initial handshake with B QPNs
    A->>A: Passive setup: A QPs -> RTR/RTS
    A-->>B: Response with A QPNs

    Note over A: Old behavior:<br/>status = CONNECTED<br/>WRs may be posted immediately

    A--xB: RDMA WR arrives too early

    Note over B: B may not have completed<br/>local QP RTR/RTS yet
```

## Modify
This PR separates local QP setup from send readiness. CONNECTED now means the local QPs have reached RTS, while ready_to_send_ means the peer has also confirmed its active-side QP setup is complete.

After the active side finishes doSetupConnection(), it sends an explicit RDMA ready ACK through the handshake path. The passive side only enables sending after receiving this ACK with matching peer QPNs.

```mermaid
sequenceDiagram
    participant B as "B active side"
    participant A as "A passive side"

    B->>A: Initial handshake with B QPNs
    A->>A: Passive setup succeeds
    A->>A: status = CONNECTED<br/>ready_to_send = false
    A-->>B: Response with A QPNs

    B->>B: Active setup succeeds<br/>B QPs -> RTR/RTS
    B->>A: Ready ACK with same B QPNs

    A->>A: Verify peer QPNs match
    A->>A: ready_to_send = true

    A->>B: Post RDMA WR safely
```

The worker now checks readyToSend() before posting WRs. If an endpoint is connected but not ready yet, slices remain queued instead of being treated as transfer failures, so waiting for the ACK does not consume retry count.

For rolling compatibility, peers that do not include the new ready_ack field keep the previous behavior.


mooncake-transfer-engine/include/transport/rdma_transport/rdma_endpoint.h
```diff
@@ -93,6 +94,14 @@ class RdmaEndPoint {
         return status_.load(std::memory_order_relaxed) == CONNECTED;
     }
 
+    // CONNECTED only means local QPs have reached RTS. A passive endpoint must
+    // still confirm the peer has completed its active setup before posting WRs.
+    bool readyToSend() const {
```
alogfans: What if status == CONNECTED but ready_to_send == false, is it OK to delete this endpoint safely, according to the most-recent endpoint implementation?

doujiang24:
@LCAIZJ Thanks, there is a race when TE used in mooncake-store, two endpoints may send to each other at the same time, while it won't happen in PD transfering, it's always single direction.

This PR almost LGTM, just Two points:
1. This PR fix it for rdma, but other kinds of endpoints may also have the same problem.
2. This PR introduce the new `ready_to_send_` field, may not be a good idea, maybe add more values in `enum Status`  could be better?

Thoughts? cc @alogfans

alogfans:
@doujiang24 
I agree with both points.

For this PR, I think it’s OK to scope the fix to RDMA first, since the concrete race is in the RDMA QP setup path. Other endpoint types can be checked/fixed separately if they have the same bidirectional setup issue.

For `ready_to_send_`, I agree that representing this as explicit `Status` states would be cleaner long term. For this bugfix, the current boolean is acceptable if the transitions are guarded carefully, but modeling it in the state machine would be the better direction.

mooncake-transfer-engine/include/transport/rdma_transport/rdma_endpoint.h
LCAIZJ: Yes, this is safe with the current endpoint implementation.

LCAIZJ:
I've updated the implementation to incorporate a state machine. Please continue with the review. @doujiang24 @alogfans

doujiang24 (approved):
lgtm

PR #2634, opened 2026-06-26
[TE] Add metadata versioning

alogfans:
## Description

This PR strengthens Transfer Engine metadata reliability by introducing a single segment-level metadata_version and using it as the freshness signal for cached metadata and derived RDMA resources.

Segment names and segment IDs can be reused after node replacement, and memory registration changes can also invalidate rkeys, address ranges, and buffer availability. With this change, every published local segment metadata update bumps metadata_version, allowing readers and RDMA workers to detect that cached descriptors or endpoints may be stale.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [X] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
```

**Test results:**
- [X] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

### DIFF
--- a/docs/source/design/transfer-engine/index.md
+++ b/docs/source/design/transfer-engine/index.md
@@ -122,6 +122,13 @@ After successfully compiling Transfer Engine, the test program `transfer_engine_
       go run . --addr=:8080
       ```
 
+    :::{toctree}
+    :maxdepth: 1
+
+    metadata
+    :::
+
+
 2. **Start the target node.**
     ```bash
     ./transfer_engine_bench --mode=target \
--- a/docs/source/design/transfer-engine/index.md
+++ b/docs/source/design/transfer-engine/index.md
@@ -79,6 +79,12 @@ next data transfer attempt.
 
 Evicted and deleted endpoints are moved to an internal `waiting_list_` and reclaimed asynchronously once their outstanding slices drain. Reclaim runs on every new endpoint insertion, and additionally on a ~1 Hz heartbeat from the per-context `monitorWorker`, so the waiting list drains even under failure load where new insertions stall while evictions continue.
 
+### Metadata Versioning
+
+See [Metadata Versioning and Replacement](metadata.md) for how Transfer Engine
+refreshes cached segment descriptors, stages metadata-backed memory
+deregistration, and invalidates RDMA resources when peer metadata changes.
+
 ### Fault Handling
 In a multi-NIC environment, one common failure scenario is the temporary unavailability of a specific NIC, while other routes may still connect two nodes.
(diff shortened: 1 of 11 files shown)


staryxchen:
The core insight that cached RDMA resources need a freshness signal after peer replacement is spot-on, and the TTL-based cache refresh approach is the right direction.

One architectural question I wanted to raise for discussion: **do we need a dedicated `metadata_version` field, or is the freshness signal already present in the descriptor data itself?**

After a remote node is replaced, the new instance will necessarily have different NICs (`devices`) and different memory registrations (`buffers[].addr`, `buffers[].lkey`/`rkey`). Comparing these fields between the cached and freshly-fetched descriptor seems sufficient to answer "did the remote side change?" — without introducing a separate version field with its own lifecycle (normalization, bumping, encode/decode).

If we can derive the invalidation decision from the data we already have, we'd avoid a few things that come with the version field:
- Additional wire protocol surface (version in JSON, buffer state strings)
- Subtle edge cases around version initialization (e.g., every new segment starting at `1`, making two different instances of the same segment name indistinguishable by version alone)
- The implicit coupling to wall-clock time in `bumpSegmentMetadataVersion`

Curious to hear your thoughts — there may well be scenarios I'm missing where a synthesized version number catches changes that a field-by-field comparison wouldn't.

alogfans:
Thanks — I think this is the right simplification for the invalidation problem this PR is trying to solve.

After a TTL refresh, we can compare the descriptor fields from which the cached RDMA resources are derived (NIC identity/path and the relevant buffer address/key data) and invalidate endpoints/rail state only when that fingerprint changes. That removes the `metadata_version` lifecycle and, importantly, avoids the initialization hole you called out: two replacement instances can both start with the same version, so the current field is not a reliable instance identifier anyway.

One small caveat is that resource values are not *guaranteed* to differ after an in-place restart—NIC identity, virtual addresses, and even keys can potentially be reused. If we ever need a strict peer-instance identity independent of resource changes, a randomly generated per-process instance epoch would be the appropriate signal, rather than the current wall-clock-based counter. The existing connection-failure recovery remains the fallback for that case.

Also, the buffer draining state is conceptually separate from peer-replacement detection; if we retain two-phase deregistration, we should evaluate that independently rather than coupling it to a segment version.

Given that this changes the central design of the PR, I’m going to close this version and rework the solution around TTL refresh plus descriptor-derived invalidation. Thanks for raising it.

PR #2570, opened 2026-06-23
[Store] fix integer overflow in BatchOffload for objects larger than 4 GiB

he-yufeng:
## What

`OffsetAllocatorStorageBackend::BatchOffload` (`mooncake-store/src/storage_backend.cpp`) accumulates a record's slice sizes into a `uint32_t`:

```cpp
uint32_t value_size = 0;
for (const auto& slice : slices) {
    value_size += static_cast<uint32_t>(slice.size);   // slice.size is uint64_t
}
```

For an object whose total size exceeds 4 GiB this **silently overflows**. The record is then allocated from the wrapped (too-small) `value_len`:

```cpp
size_t record_size = RecordHeader::SIZE + header.key_len + header.value_len; // value_len truncated
auto allocation = allocator_->allocate(record_size);
```

…but the write step emits the **full original slices**:

```cpp
for (const auto& slice : slices) iovs.push_back({slice.ptr, slice.size});
data_file_->vector_write(iovs.data(), iovs.size(), offset);
```

so `vector_write` writes past the allocated extent — a heap/arena buffer overflow. The `// Use size_t for record_size to handle large objects` comment does not help, because `value_len` is already truncated before that line. The post-write `written != record_size` check only fires *after* the over-write has happened.

## Why it's reachable

There is no per-object size cap before this path — the only size checks are tenant/bucket quotas (total storage), not a single-object limit. So a value > 4 GiB reaching SSD offload triggers it.

## Fix

`RecordHeader.value_len` is a `uint32_t` (8-byte on-disk header) by design — the header has a `// Currently assumes max object size is 4GB` note. Rather than change the on-disk format, this sums in 64 bits and **skips** any key whose value exceeds 4 GiB (matching the function's existing per-key continue-on-failure model), turning silent memory corruption into a logged skip.

## Verification

The overflow is a deterministic type-width issue, so I verified it by code inspection / type analysis rather than a live run: a `uint64_t` slice-size sum truncated into a `uint32_t` provably wraps at 2^32, and `record_size` is derived from the truncated value. I don't have an RDMA/CUDA build environment locally, so I have **not** built/run the C++ suite — CI will compile and run it.

Happy to add a regression test if you'd like: the new guard returns before any I/O, so a test can pass a `Slice` with `size > UINT32_MAX` and a small dummy buffer and assert the key is skipped, without allocating 4 GiB. Point me at the preferred harness (`mooncake-store/tests/storage_backend_test.cpp`?) and I'll add it.

PR #2635, opened 2026-06-26
[Store] Fix 32-bit host-pointer offset overflow in SPDK NoF transfers over 4 GiB

he-yufeng:
## Description

In the SPDK NoF worker (`mooncake-store/src/transfer_task.cpp`, `SpdkNofWorkerPool::workerThread`), the host buffer offset for each sub-I/O was computed as:

```cpp
int lba_off = task->lba_count - task->remaining_lba;   // int
...
void* submit_ptr = reinterpret_cast<char*>(task->ptr) + lba_off * block_size;
```

`block_size` is a `uint32_t`, so `lba_off * block_size` is evaluated in 32-bit unsigned arithmetic. Once the running offset reaches 4 GiB (`lba_off * block_size >= 2^32`) the product wraps around and `submit_ptr` points to the wrong part of the buffer. The DMA for the tail of a >4 GiB transfer then reads from / writes to the *start* of the user buffer instead of the intended region, silently corrupting data.

The device-side address one line below — `uint64_t submit_lba = task->lba + lba_off;` — is already 64-bit, so only the host pointer offset was left narrow.

This is reachable for any sufficiently large transfer: a single `SpdkNofTask` is created with `lba_count = size / block_size` for the whole request and there is no size cap at the submit site (`submitNofTransfer` only checks block alignment), so `lba_off` walks from 0 up to `lba_count` as the task drains. It is the same >4 GiB scenario that was fixed for SSD offload in #2570.

The fix widens `lba_off` to `uint64_t` so the offset arithmetic stays 64-bit, matching the already-widened `submit_lba`.

## Module

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

The SPDK NoF path is behind `USE_NOF` and needs SPDK + NVMe-oF hardware, so I could not build or exercise it locally (no native SPDK toolchain on this machine). It is verified three ways:

1. **Standalone reproduction of the arithmetic.** Mirroring the exact types (`int lba_off`, `uint32_t block_size`) and picking the first `lba_off` whose byte offset crosses 4 GiB for a 5 GiB transfer with 4096-byte blocks:

   ```
   lba_count=1310720 lba_off=1048576 block_size=4096
   expected byte offset      = 4294967296
   buggy (32-bit) offset     = 0
   fixed (64-bit) offset     = 4294967296
   buggy pointer wrong by    = -4294967296 bytes
   RESULT: OVERFLOW CONFIRMED (buggy != fixed)
   ```

   The 32-bit expression wraps to 0 (off by exactly 4 GiB); the 64-bit expression is correct.

2. **Parity with #2570**, which fixed the same 32-bit-overflow-on-a->4 GiB-size class in the SSD offload path and was accepted as a code-only change.

3. `clang-format` clean on the file.

- [x] Manual testing done (standalone arithmetic repro above; SPDK NoF integration build exercised by CI)

## Checklist

- [x] I have formatted my code using `./scripts/code_format.sh` (clang-format)

## AI Assistance Disclosure

Used Claude Code to help sweep the store/transfer-engine code for the remaining 32-bit size/offset arithmetic siblings of #2570 and to draft the change. I reviewed and verified the fix (including the standalone overflow reproduction above) and am responsible for it.

### DIFF
--- a/mooncake-store/src/transfer_task.cpp
+++ b/mooncake-store/src/transfer_task.cpp
@@ -485,7 +485,13 @@ void SpdkNofWorkerPool::workerThread(int work_idx) {
                         uint32_t submit_lba_count = std::min(
                             avail_blocks, std::min(task->remaining_lba,
                                                    nof_qos->blocks_per_chunk));
-                        int lba_off = task->lba_count - task->remaining_lba;
+                        // 64-bit: a NoF transfer can exceed 4 GiB, and once
+                        // lba_off * block_size crosses 2^32 the host-pointer
+                        // offset would wrap when computed in 32 bits (int *
+                        // uint32_t), making submit_ptr address the wrong part
+                        // of the buffer. submit_lba below is already widened.
+                        uint64_t lba_off =
+                            task->lba_count - task->remaining_lba;
                         uint64_t submit_lba = task->lba + lba_off;
                         void* submit_ptr = reinterpret_cast<void*>(
                             reinterpret_cast<char*>(task->ptr) +


he-yufeng:
The red `Build Docker Image` job on this run is a CI-runner disk-space flake, not a problem with the change. It died during LTO linking with:

```
lto1: fatal error: error writing to /tmp/cc6J0hfT.s: No space left on device
lto-wrapper: fatal error: /usr/bin/c++ returned 1 exit status
```

All 11 other build/test jobs on the same run passed (`build 3.10/3.12`, `build-musa`, `build-flags`, `build-wheel-cu13`, `build-wheel-efa`, `ascend-test`), so the offset-overflow fix itself compiles cleanly everywhere. Since this is the only red job and it gates `CI Gate`, could a maintainer re-run just the Docker image job when convenient? I'm also happy to push a rebase to re-trigger it on a fresh runner if that's easier.

he-yufeng:
Gentle nudge on this one — it's a small type fix (`int` → `uint64_t` for `lba_off`) that prevents host-pointer wraparound on SPDK NoF transfers past 4 GiB. Green across the board and it's been sitting since late June with only the bot having looked. Could a maintainer give it a quick pass when there's a window? Happy to rebase if it's drifted. Thanks!

he-yufeng:
Rebased a while back and it is still mergeable with a green CI. The overflow is reachable on 32-bit hosts writing SPDK NoF at offsets past 4 GiB, and the patch is contained to the offset arithmetic plus a regression test. Could someone take a look when there is a moment?

he-yufeng:
One more nudge on this small one. Still mergeable and green; it keeps 32-bit hosts from wrapping the SPDK NoF host-pointer offset past 4 GiB by widening lba_off to uint64_t. Happy to rebase again if that helps it get a look.

he-yufeng:
Re-checked against current main: the overflow is still there. `transfer_task.cpp` still computes `int lba_off = task->lba_count - task->remaining_lba`, and `lba_off * block_size` is evaluated in 32 bits, so a NoF transfer past 4 GiB wraps the host-pointer offset and addresses the wrong part of the buffer. This one-line widening (plus the comment) still applies cleanly. Could someone from the NoF side take a look? The fix is small but the corruption it prevents is silent.

he-yufeng:
Also still current four days on: merges cleanly (MERGEABLE on main) with green CI.

The wrap is one line of arithmetic but the failure mode is ugly: `lba_off * block_size` stays in 32-bit unsigned, so once a running offset crosses 4 GiB the host pointer wraps and the tail of a large transfer DMAs to the start of the user buffer, silently corrupting data. The device LBA one line down was already 64-bit, so this just widens the host offset to match. If there is a preferred idiom for the cast I am happy to switch to it.

Issue #2633, opened 2026-06-26
[RFC]: Host-aware Local-first Allocation for Mooncake Store


yokinoshitayoki:
## Motivation

Mooncake Store currently supports preferred allocation through `ReplicateConfig.preferred_segment` and `ReplicateConfig.preferred_segments`. When no preferred segment is specified, memory replicas are allocated by the configured allocation strategy, which commonly falls back to a random segment.

For KV cache workloads, random placement also increases the failure blast radius of a store node. In the current Mooncake Store cluster, when a store node goes offline, either because of a crash or planned shutdown, all KV cache stored on that node is lost. KV cache reuse depends on a continuous prefix. If a block in the middle of a request's cached sequence is missing, the later cached blocks can no longer be reused even if they still exist on other store nodes. Those suffix blocks remain stored in the cluster but lose their practical value for prefix reuse.

With random placement, KV cache blocks produced by one writer may be spread across many store nodes. A single store node failure can therefore punch holes into cached sequences from many different writers and invalidate the suffixes behind those holes. Host-aware local-first placement reduces this blast radius in the common case by keeping a writer's cache on local store segments first and only using remote hosts when local capacity is insufficient.

Locality is also a motivation. The writer is usually an inference engine process running on a specific physical host. If the KV cache is stored on a remote host while local CPU memory is available, later transfers between the inference engine and Mooncake Store may cross the network unnecessarily. This increases latency and consumes network bandwidth.

The existing preferred segment interface can be used by an upper layer such as vLLM or SGLang, but it requires the upper layer or deployment system to know the exact `Segment.name` of the local store. This is fragile for a general-purpose Mooncake feature:

1. `Segment.name` is a logical string and has no required host/IP format.
2. A physical host may run multiple Mooncake clients and multiple store segments.
3. A requester client may not mount any segment in standalone-store mode.
4. Upper layers should not need to manually enumerate Mooncake store segments to get a reasonable local placement policy.

This RFC proposes a host-aware local-first allocation mode inside Mooncake Store. Mooncake master will maintain host-level topology metadata and use it to prefer local store segments for Put allocation. If local segments cannot satisfy the allocation, master will fall back to other hosts in a deterministic ordered sequence rather than choosing a random segment immediately.

## Goals

1. Add a Mooncake-native notion of physical host identity, `host_id`.
2. Let master map writer clients to hosts and hosts to active store segments.
[...]

PR #2638, opened 2026-06-26
[Store] Add Local First Allocation Strategy

yokinoshitayoki:
## Description

Implements local-first allocation strategy for Mooncake Store, as proposed in [#2633](https://github.com/kvcache-ai/Mooncake/issues/2633).

This adds a new master allocation strategy, `local_first`. When selected, memory Put/Upsert requests with `replica_num == 1` prefer segments on the writer's host. If local segments are unavailable or full, allocation falls back to remote hosts in deterministic lexicographic host order. Within the same host, segment names are sorted and rotated by key hash to avoid always trying the first segment.

Host identity is derived automatically from each client/store process's `local_hostname` by removing the port. For example, `host-a:50051` and `host-a:50052` map to the same host id, `host-a`. Loopback and wildcard hostnames such as `localhost`, `127.0.0.1`, `0.0.0.0`, and `::1` are treated as unknown and do not trigger local-first placement.

This change adds:

- `allocation_strategy=local_first`
- `Segment.host_id` metadata and master-side host/segment indexes
- host-ordered preferred segment generation in the master
- snapshot serialization compatibility for `Segment.host_id`
- deployment documentation for the new strategy

Explicit `preferred_segment` / `preferred_segments` are tried first, then allocation falls back to the host-aware ordered segment list. The default allocation strategy remains unchanged.

## Module

- [ ] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [x] Documentation update
- [x] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
cmake -S . -B build -DBUILD_UNIT_TESTS=ON -DWITH_STORE_RUST=OFF -DWITH_STORE_GO=OFF -DWITH_EP=OFF

cmake --build build --target master_service_test segment_test serializer_test -j32

./build/mooncake-store/tests/master_service_test --gtest_filter='MasterServiceTest.ResolveMooncakeHostIdPrefersEnvAndRejectsLoopback:MasterServiceTest.LocalFirstPutPrefersWriterHost:MasterServiceTest.LocalFirstPutFallsBackToNextOrderedHost:MasterServiceTest.LocalFirstPutFallsBackWhenLocalSegmentIsFull:MasterServiceTest.ExplicitPreferredSegmentOverridesLocalFirst'

./build/mooncake-store/tests/segment_test --gtest_filter='SegmentTest.HostOrderedSegmentsTracksMountStatusAndUnmount:SegmentTest.HostOrderedSegmentsRotateWithinSameHostByKey'

./build/mooncake-store/tests/serializer_test --gtest_filter='SerializerTest.MountedSegmentSerializationPreservesHostId:SerializerTest.MountedSegmentDeserializesLegacyFormatWithoutHostId'

ctest --test-dir build -R '^(segment_test|serializer_test)$' --output-on-failure

./scripts/code_format.sh

pre-commit run --files docs/source/deployment/mooncake-store-deployment-guide.md

git diff --check

```
**Test results:**
- [x] Unit tests pass

## SGLang + Mooncake Store Failure Benchmark

I also validated this change with an external SGLang PD + Mooncake Store benchmark to measure the blast radius of a store-node failure.

### Testbed

- Topology: 3P1D on 4 H20 machines.
  - 3 prefill hosts, 1 prefill worker per host.
  - 1 decode host, also running the Mooncake master and SGLang router.
- Model: Qwen3-8B.
- SGLang:
  - `tp-size=1`
  - `context-length=4096`
  - `max-running-requests=8`
  - `max_tokens=1`, `temperature=0`
  - HiCache enabled with Mooncake storage backend.
  - `hicache-ratio=1.01`
  - `hicache-size=0`
  - `hicache-write-policy=write_through`
  - `hicache-storage-prefetch-policy=wait_complete`
- Mooncake Store:
  - Each prefill host runs 2 standalone `mooncake_client` store processes.
  - Each store process contributes one 8 GiB memory segment.
  - Total before failure: 6 memory segments, 48 GiB.
  - Failed host: 2 memory segments removed, leaving 4 memory segments, 32 GiB.
  - Prefill workers use `global_segment_size=0`, so they act as writers only and do not contribute storage segments.
- Compared strategies:
  - Baseline: `allocation_strategy=random`
  - Treatment: `allocation_strategy=local_first`

### Dataset And Request Generation

The benchmark used the Hugging Face dataset [`semianalysisai/cc-traces-weka-with-subagents-052726-256k`](https://huggingface.co/datasets/semianalysisai/cc-traces-weka-with-subagents-052726-256k).This dataset includes an outer trace `id`, which can be used as a session id to keep requests from the same session routed consistently. This is important for evaluating host-aware allocation: the policy is most effective when requests from the same session are routed to the same writer host, so the session's reusable KV blocks can be placed within the same host domain.

Processing details:

- Sample set:
  - First 10,000 requests
  - `hash_id_scope=local`
- The outer trace `id` was used as both:
  - SGLang routing key: `X-SMG-Routing-Key=<outer id>`
  - cache namespace: `cache_salt=ccweka:<outer id>`
- This avoids incorrectly sharing block IDs across independent sessions.
- Prompt content was deterministically generated from each request's session-local `hash_ids`.

### Failure Procedure

Each strategy was tested in a clean run:

1. Start a fresh Mooncake master with the target allocation strategy.
2. Start 6 Mooncake Store segments across the 3 prefill hosts.
3. Start SGLang prefill workers, decode worker, and router.
4. Warm up with 8,000 requests.
5. Stop both Mooncake Store processes on one prefill host.
6. Wait until the Mooncake master reports that the failed host's 2 segments are removed.
7. Call `/flush_cache` on all 3 prefill workers.
8. Replay the next 2,000 requests and measure post-failure cached-token ratio.

The `/flush_cache` step is important here: it clears the local SGLang cache state before post-failure replay, so the post-failure window measures recovery from Mooncake Store after the failed host's segments are removed.

### Results

Pre-failure warmup cached-token ratios were similar:

| Strategy | Warmup requests | Cached-token ratio |
|---|---:|---:|
| `random` | 8,000 | 94.68% |
| `local_first` | 8,000 | 94.58% |

Post-failure cumulative cached-token ratio:

| Post-failure window | `random` | `local_first` | Improvement |
|---:|---:|---:|---:|
| first 5 | 0.00% | 40.10% | +40.10 % |
| first 10 | 22.13% | 39.99% | +17.86 % |
| first 20 | 63.03% | 69.92% | +6.89 % |
| first 50 | 79.99% | 80.38% | +0.39 % |
| first 100 | 89.24% | 89.34% | +0.10 % |
| first 200 | 92.17% | 93.20% | +1.03 % |
| first 500 | 95.65% | 96.25% | +0.60 % |
| first 1000 | 96.42% | 97.31% | +0.89 % |
| first 2000 | 96.31% | 96.68% | +0.37 % |

The improvement is most visible in the immediate post-failure window. For example, the first 5 requests improved from 0.00% to 40.10%, and the first 10 requests improved from 22.13% to 39.99%.

This matches the expected behavior: random allocation can spread one session's prefix blocks across multiple store hosts, so losing one store host can break reusable prefixes for sessions whose writers are still healthy. In contrast, host-aware local-first allocation tends to keep a writer host's KV blocks on that host first, so the failure impact is more localized. In larger production deployments with more prefill/store hosts, random allocation may create a wider and longer-lasting blast radius, while host-aware placement is expected to make the affected window more localized. This still needs to be validated on production-scale traffic, but the 3P1D benchmark confirms the intended direction.

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex was used to help implement the changes. But I have reviewed the code and docs.


docs/source/deployment/mooncake-store-deployment-guide.md
```diff
@@ -450,7 +450,7 @@ mooncake_master \
 
 | Flag | Default | Description |
 |------|---------|-------------|
-| `--allocation_strategy` | `random` | `random` (pure random, fastest), `free_ratio_first` (best load balance), or `cxl` (prefer CXL memory) |
+| `--allocation_strategy` | `random` | Allocation strategy: `random` (pure random, fastest), `free_ratio_first` (best memory load balance), `ssd_free_ratio_first` (SSD-aware free-ratio-first), `cxl` (prefer CXL memory), or `host_aware_local_first` (prefer writer-host memory segments before ordered remote fallback) |
```
ykwd: Since there is no other local-first strategy, maybe we can simply name it as local_first?

yokinoshitayoki: > Since there is no other local-first strategy, maybe we can simply name it as local_first?

Thanks for review. I have renamed the strategy to `local_first` as suggested.

Lin-z-w (commented):
I think the host identity semantics need to be clarified or made configurable.

This PR describes the strategy as host-aware local-first allocation, but the current implementation derives `host_id` only by stripping the port from `local_hostname`. That works when all processes on the same machine use the exact same hostname/IP with different ports, e.g. `host-a:50051` and `host-a:50052` both become `host-a`.

However, it does not reliably identify a physical/logical host in common deployments:
- the same machine using different NIC IPs becomes different host IDs;
- the same machine using short hostname vs FQDN becomes different host IDs;
- container/pod IP and node IP can become different host IDs;
- multi-network deployments may not get local-first behavior even when writer and segment are actually on the same machine.

Loopback/wildcard values being treated as unknown is reasonable because otherwise all nodes using `localhost` or `0.0.0.0` could collapse into the same host ID. But with the current fallback behavior, those clients simply do not participate in host-aware placement, so this should be documented as a deployment requirement rather than implied as automatic host detection.

I would suggest either:
1. document clearly that `local_hostname` must be a stable, globally unique, same-across-processes-per-machine host identifier for this strategy to work correctly; or
2. add an explicit `host_id` configuration/env override so deployments with multiple NICs, aliases, or container networking can define the intended host identity directly.

Without that, the implementation is more accurately `local_hostname_without_port` aware than generally host-aware.

yokinoshitayoki:
> I think the host identity semantics need to be clarified or made configurable.
> 
> This PR describes the strategy as host-aware local-first allocation, but the current implementation derives `host_id` only by stripping the port from `local_hostname`. That works when all processes on the same machine use the exact same hostname/IP with different ports, e.g. `host-a:50051` and `host-a:50052` both become `host-a`.
> 
> However, it does not reliably identify a physical/logical host in common deployments:
> 
> * the same machine using different NIC IPs becomes different host IDs;
> * the same machine using short hostname vs FQDN becomes different host IDs;
> * container/pod IP and node IP can become different host IDs;
> * multi-network deployments may not get local-first behavior even when writer and segment are actually on the same machine.
> 
> Loopback/wildcard values being treated as unknown is reasonable because otherwise all nodes using `localhost` or `0.0.0.0` could collapse into the same host ID. But with the current fallback behavior, those clients simply do not participate in host-aware placement, so this should be documented as a deployment requirement rather than implied as automatic host detection.
> 
> I would suggest either:
> 
> 1. document clearly that `local_hostname` must be a stable, globally unique, same-across-processes-per-machine host identifier for this strategy to work correctly; or
> 2. add an explicit `host_id` configuration/env override so deployments with multiple NICs, aliases, or container networking can define the intended host identity directly.
> 
> Without that, the implementation is more accurately `local_hostname_without_port` aware than generally host-aware.

Thanks, I have clarified the deployment requirement in the docs.

mooncake-store/src/utils.cpp
```diff
@@ -551,6 +551,33 @@ std::string GetEnvStringOr(const char *name, const std::string &default_value) {
...
+        (hostname == "::1") ? hostname : trim(getHostNameWithoutPort(hostname));
+    if (host_id.empty()) {
+        return "";
+    }
+
+    std::string lower = host_id;
+    std::transform(lower.begin(), lower.end(), lower.begin(),
+                   [](unsigned char c) { return std::tolower(c); });
+    if (lower == "localhost" || lower == "127.0.0.1" || lower == "0.0.0.0" ||
+        lower == "::1" || lower == "[::1]") {
```
Lin-z-w: Please also reject the IPv6 unspecified/wildcard address (`::` and `[::]`) here. Today `ResolveMooncakeHostId("[::]:50051")` resolves to `::`, which is not filtered, so every process configured with an IPv6 wildcard bind address would share the same host id and local-first could treat unrelated machines as the same host. This should behave like `0.0.0.0`: return an empty host id and add coverage for `::` / `[::]:port`.

Lin-z-w:
> > I think the host identity semantics need to be clarified or made configurable.
> > This PR describes the strategy as host-aware local-first allocation, but the current implementation derives `host_id` only by stripping the port from `local_hostname`. That works when all processes on the same machine use the exact same hostname/IP with different ports, e.g. `host-a:50051` and `host-a:50052` both become `host-a`.
> > However, it does not reliably identify a physical/logical host in common deployments:
> > 
> > * the same machine using different NIC IPs becomes different host IDs;
> > * the same machine using short hostname vs FQDN becomes different host IDs;
> > * container/pod IP and node IP can become different host IDs;
> > * multi-network deployments may not get local-first behavior even when writer and segment are actually on the same machine.
> > 
> > Loopback/wildcard values being treated as unknown is reasonable because otherwise all nodes using `localhost` or `0.0.0.0` could collapse into the same host ID. But with the current fallback behavior, those clients simply do not participate in host-aware placement, so this should be documented as a deployment requirement rather than implied as automatic host detection.
> > I would suggest either:
> > 
> > 1. document clearly that `local_hostname` must be a stable, globally unique, same-across-processes-per-machine host identifier for this strategy to work correctly; or
> > 2. add an explicit `host_id` configuration/env override so deployments with multiple NICs, aliases, or container networking can define the intended host identity directly.
> > 
> > Without that, the implementation is more accurately `local_hostname_without_port` aware than generally host-aware.
> 
> Thanks, I have clarified the deployment requirement in the docs.

Thanks for updating the deployment guide. One more docs sync is needed: `docs/source/design/mooncake-store.md` still says the only valid `--allocation_strategy` values are `random`, `free_ratio_first`, and `cxl`, and its strategy table does not mention `ssd_free_ratio_first` or the new `local_first`. Please update that design page too so the public docs do not contradict this table.

Lin-z-w:
Question: do we need to propagate `host_id` through `Ping`?

From the current implementation, local-first placement seems to already have the host identity it needs on the write path: `PutStart` / `UpsertStart` attach `ReplicateConfig.host_id`, and the master updates `client_host_id_` from that before allocation. Segment-side host identity is also carried by `Segment.host_id` during mount/remount.

Given that, adding `host_id` to `Ping` feels like a secondary synchronization path rather than a requirement for local-first allocation. It also expands the RPC surface for a heartbeat API whose main responsibility is liveness/view-version state.

Would it be cleaner to keep `Ping` unchanged and have allocation use the request-local `config.host_id` directly, falling back to `client_host_id_` only when the request does not carry one? That would keep host identity tied to the operation that needs it, while still allowing cached client host identity as a best-effort fallback.

yokinoshitayoki:
@Lin-z-w Thanks for the review. I addressed these points in the latest commit:

- Rejected IPv6 wildcard host IDs (`::` and `[::]`) in `ResolveMooncakeHostId`, and added coverage for `::`, `[::]`, and `[::]:port`.
- Updated `docs/source/design/mooncake-store.md` so the documented allocation strategy list includes both `ssd_free_ratio_first` and the new `local_first`.
- Removed `host_id` propagation from `Ping`. Host identity is now tied to the write path: allocation uses the request-local `ReplicateConfig.host_id` first, and only falls back to the cached `client_host_id_` when the request does not carry one. `Ping` is back to only handling liveness/view-version state.

docs/source/deployment/mooncake-store-deployment-guide.md
```diff
@@ -604,7 +604,7 @@ allocation_strategy: "local_first"
 
 When enabled, the master applies local-first allocation only for memory replicas with `replica_num == 1`. Explicit `preferred_segment` or `preferred_segments` are tried first; if they are unavailable or full, Mooncake falls back to the writer host and then remote hosts in lexicographic host-id order. Within the same host, segment names are sorted and rotated by key hash so multiple segments on one host do not always receive the first allocation attempt.
```
Lin-z-w: The docs still describe the remote fallback as lexicographic host-id order, but the implementation is actually cyclic lexicographic order starting from the writer host (or `lower_bound(writer_host_id)` when the writer host has no segments). For example, with `host0` and `host2` available and writer host `host1`, the test expects `host2`, not the globally first lexicographic host `host0`.

Could we clarify the wording here to say that fallback uses cyclic lexicographic order starting at the writer host / next ordered host? Otherwise operators may expect a different deterministic remote-host order than the one implemented.

yokinoshitayoki: Thanks, good catch. I updated both the deployment guide and the design doc to describe the actual behavior as cyclic lexicographic host-id order, starting from the writer host when it has active segments, or otherwise from the next greater active host id.

Lin-z-w (approved):
LGTM.

ykwd (approved):

Issue #2568, opened 2026-06-23
[RFC]: Traffic class hint API for TransferEngine — bridging upper-layer semantics (KV / EP / CTRL) to lower-layer SL/TC


catyans:
## Summary

Extend the `SelectionContext` introduced in #2079 with a `traffic_class` field, allowing callers to express *application semantics* (KV-cache transfer vs. EP all-to-all vs. control-plane RPC) so the engine can map them to the *link-layer resources* (SL / TC / QP pool) already exposed by #1187, #2525, and #2526.

This RFC does **not** introduce a new abstraction — it adds one orthogonal dimension to an existing struct.

## Motivation

Three independent pieces of work over the last six weeks have converged on "different traffic types should use different NIC resources", but the convergence is happening *only at the link layer*:

| Project | Knob | Layer |
|---|---|---|
| NCCL v2.30.7 | `NCCL_GIN_IB_TC` (separate TC for GIN control vs. data) | env var |
| Mooncake #1187 | `MC_IB_TC` (global RDMA traffic class) | env var |
| Mooncake #2525 | `MC_IB_SL` (global RDMA service level) | env var |
| Mooncake #2526 | SL/TC applied to the notification QP | code |

All four expose **link-layer knobs with global scope** — one value per process, applied to every transfer regardless of its purpose. There is currently no way for a caller to say:

> "This `submitTransfer` carries a KV write on the TTFT critical path; that one is an EP dispatch background flow; the third is a heartbeat."

The link layer is therefore forced into one of three suboptimal modes:

1. Static one-size-fits-all (today): all flows share the same SL/TC, contention is unmanaged.
2. Per-process pinning via env vars: requires running multiple `TransferEngine` instances to differentiate flows, wasting resources.
3. Caller-provided raw SL/TC: leaks link-layer details into vLLM / SGLang / DeepEP, violating layering.

The orthogonal SGLang RFC sgl-project/sglang#28631 plumbs **priority** (`CRITICAL` / `BACKGROUND` / `IDLE`) through the Python data path. As that RFC explicitly notes, priority decides *when* a transfer is sent (queue order); it does **not** decide *which NIC channel* it uses. The traffic-class dimension proposed here covers the latter and is independently meaningful.

## Proposed change

Extend `SelectionContext` (mooncake-transfer-engine/tent/include/tent/runtime/transport_selector.h, introduced in #2079) with one optional field:

```cpp
enum class TrafficClass : uint8_t {
[...]

PR #2526, opened 2026-06-18
[TransferEngine] Apply configured SL/TC to the TENT notification QP

catyans:
## Description

The TENT RDMA **data-path** QP picks up `params_->service_level` and
`params_->traffic_class` (`endpoint.cpp`, `setupOneQP`), but the
**control-path** notification QP hard-codes `ah_attr.sl = 0` and
`ah_attr.grh.traffic_class = 0` in `setupNotifyQpConnection`.

As a result, when a user configures an SL/TC for QoS isolation — e.g. to
steer Mooncake traffic into a dedicated InfiniBand Virtual Lane, away
from EP all-to-all traffic that shares the same NIC in MoE inference —
the KV-arrival **notification** messages still leak onto the default
lane, so the isolation is incomplete.

This threads the endpoint's `service_level` / `traffic_class` through
`setupNotifyQpConnection` so the notification QP follows the same fabric
QoS settings as the data QPs.

**Behavior-preserving**: both fields default to `0` in `RdmaParams`, so
an unconfigured deployment sets `sl=0` / `traffic_class=0` exactly as
before. This complements #2525 (which exposes `MC_IB_SL` for the main
transfer-engine), making SL/TC consistent across data and control QPs.

## Module

- [x] Transfer Engine (`mooncake-transfer-engine`)

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

This touches the IB QP setup path, which requires real RDMA hardware to
exercise end-to-end. The change is a behavior-preserving plumbing fix
(threads two already-existing `RdmaParams` fields, both defaulting to 0)
that mirrors how the data-path QP (`setupOneQP`) already applies them.

- [ ] Unit tests pass (N/A — IB QP setup path)
- [ ] Integration tests pass
- [ ] Manual testing done

> Note: not built locally (no IB toolchain on my dev machine); relying
> on CI for the build.

## Checklist

- [x] I have performed a self-review of my own code
- [x] I have formatted my code using clang-format (v20.1.8, repo `.clang-format`)
- [ ] I have run `pre-commit run --all-files` and all hooks pass
- [x] For changes >500 LOC: I have filed an RFC issue — N/A (~12 LOC)

## AI Assistance Disclosure

- [x] AI tools were used (Claude Code) for code navigation and drafting.

PR #2525, opened 2026-06-18
[TransferEngine] Make InfiniBand Service Level configurable via MC_IB_SL

catyans:
## Description

The RDMA endpoint hard-codes `attr.ah_attr.sl = 0` in
`rdma_endpoint.cpp`, so Mooncake traffic cannot be steered into a
dedicated InfiniBand Virtual Lane for QoS isolation.

This matters on a **shared NIC**. In MoE inference (e.g. DeepSeek-V3,
Qwen3-235B) with PD disaggregation, Expert-Parallel all-to-all traffic
and KV-cache transfers contend for the same RDMA NIC. SL→VL mapping on
the switch is the standard way to separate them, but Mooncake currently
gives KV traffic no SL knob, so it cannot be moved off the default lane.

Peer libraries on this path already expose the knob:

| Library | SL knob |
|---|---|
| NCCL | `NCCL_IB_SL` |
| UCX | `UCX_IB_SL` |
| NVSHMEM | `NVSHMEM_IB_SL` |
| DeepEP | `EP_OVERRIDE_RDMA_SL` |
| **Mooncake TE** | **(none — this PR)** |

Note: the existing `MC_IB_TC` only sets `ah_attr.grh.traffic_class`,
which does not drive VL selection on native InfiniBand, so it does not
cover this case.

This change mirrors the existing `MC_IB_TC` implementation:
- add `ib_service_level = -1` to `GlobalConfig` (`-1` = use default 0);
- parse `MC_IB_SL`, validated to the InfiniBand SL range `0-15`;
- apply it to `attr.ah_attr.sl` only when configured;
- log it in `dumpGlobalConfig`.

**Fully backward compatible**: when `MC_IB_SL` is unset, the SL stays
`0`, exactly as today.

Closes #2515

## Module

- [x] Transfer Engine (`mooncake-transfer-engine`)

## Type of Change

- [x] New feature

## How Has This Been Tested?

Added unit tests in `tests/config_test.cpp` for the `MC_IB_SL` env-var
parsing, covering: unset default (`-1`), valid override, min/max
boundaries (`0`/`15`), out-of-range (`16`), negative, and non-numeric
input (all reject and preserve the prior value). Mirrors the existing
`MC_PKEY_INDEX` test pattern.

- [x] Unit tests pass
- [ ] Integration tests pass (N/A)
- [ ] Manual testing done

> Note: I have not built this locally (no IB toolchain on my dev
> machine); relying on CI for the build. The change mirrors the existing
[...]

PR #2640, opened 2026-06-26
[TENT] SelectionPolicy: per-policy SL/TC/qp_pool schema (RFC #2568 step 1)

catyans:
## Description

Step 1 of the two-step plan agreed with @staryxchen on #2568.

Instead of adding a `TrafficClass` enum to `SelectionContext` (the original
RFC shape, which @staryxchen correctly flagged as layering inversion + a
closed-enum release bottleneck), this uses the **existing opaque
`policy_name` hook** and extends `SelectionPolicy` with link-layer QoS
attributes. The engine never interprets application semantics; callers bind
to a named policy and the JSON config carries the fabric QoS.

This lets SL/TC be configured **per policy** instead of only process-global
via `MC_IB_SL` / `MC_IB_TC` (#2525 / #2526).

### Scope: schema + plumbing only
This PR does **not** apply SL/TC at QP setup. As @staryxchen noted, real
per-class differentiation requires per-class QP pools (SL/TC are baked into
QP attributes at handshake), which is the explicit **step-2** follow-up.
`qp_pool` is parsed and reserved here so the JSON schema is
forward-compatible and step 2 needs no schema change or breakage of existing
configs (per @staryxchen's reminder).

### Changes
- `SelectionPolicy`: add optional `service_level` (0-15), `traffic_class`
  (0-255), `qp_pool` (string). `nullopt` = fall back to the global
  `RdmaParams` default = today's behavior.
- `loadPolicies()`: parse the three from the same JSON as the existing
  fields; out-of-range SL/TC are ignored with a warning so a bad config
  never changes selection.
- `SelectionResult`: carry the matched policy's SL/TC/qp_pool to the caller
  for the step-2 QP-setup follow-up.
- Tests: parse + carry-through, and out-of-range-ignored.

**Fully backward compatible**: unset fields behave exactly as before; the
default policies (which don't set them) are unaffected.

## Module
- [x] Transfer Engine (`mooncake-transfer-engine`) — TENT runtime

## Type of Change
- [x] New feature (config schema)

## How Has This Been Tested?
Added unit tests in `transport_selector_test.cpp`:
- a policy with `service_level`/`traffic_class`/`qp_pool` is parsed and the
  values are carried through `SelectionResult`;
- out-of-range SL/TC are ignored (left `nullopt`).

- [x] Unit tests pass (added)
- [ ] Integration tests pass
- [ ] Manual testing done

> Note: not built locally (no TENT toolchain on my dev machine); relying on
> CI. clang-format (v20.1.8) applied.

## Checklist
- [x] I have formatted my code using clang-format
- [x] For changes >500 LOC: I have filed an RFC issue — RFC #2568

## AI Assistance Disclosure
- [x] AI tools were used (Claude Code) for code navigation and drafting.

### DIFF
--- a/mooncake-transfer-engine/tent/include/tent/runtime/transport_selector.h
+++ b/mooncake-transfer-engine/tent/include/tent/runtime/transport_selector.h
@@ -118,17 +118,12 @@ struct SelectionPolicy {
     // Transport preference list (evaluated in order)
     std::vector<TransportType> transports;
 
-    // --- Link-layer QoS attributes (RFC #2519 / #2568, step-1 schema only) ---
-    // These let a policy carry per-policy fabric QoS instead of the
-    // process-global MC_IB_SL / MC_IB_TC. nullopt = fall back to the global
-    // RdmaParams value (today's behavior). InfiniBand Service Level (0-15) and
-    // Traffic Class / DSCP (0-255).
-    std::optional<int> service_level;  // nullopt = use global default
-    std::optional<int> traffic_class;  // nullopt = use global default
-    // Named QP pool this policy's traffic should land on. Reserved here so the
-    // schema is forward-compatible: step 1 only parses/stores it; the actual
-    // per-class QP pool creation and routing is a follow-up (step 2). Unset =
-    // the current single "data QP" path.
+    // Per-policy link-layer QoS. nullopt = fall back to the global RdmaParams
+    // value. InfiniBand Service Level (0-15) and Traffic Class / DSCP (0-255).
+    std::optional<int> service_level;
+    std::optional<int> traffic_class;
+    // Named QP pool this policy's traffic should land on; parsed and stored for
+    // now, routing to be wired later. Unset = the current single "data QP".
     std::optional<std::string> qp_pool;
 };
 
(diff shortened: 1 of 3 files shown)


mooncake-transfer-engine/tent/src/runtime/transport_selector.cpp
```diff
@@ -206,6 +206,34 @@ void TransportSelector::loadPolicies() {
...
+                LOG(WARNING) << "Ignore traffic_class in policy " << policy.name
+                             << ", value " << tc << " out of range (0-255)";
+            }
+        }
+        // Reserved for step 2 (per-class QP pools); parsed for forward schema
+        // compatibility, no effect yet.
+        if (policy_json.contains("qp_pool")) {
+            auto& qp = policy_json["qp_pool"];
+            if (qp.is_string()) policy.qp_pool = qp.get<std::string>();
+        }
```
staryxchen, line 243: Empty-string qp_pool is treated as set, which is semantically different from nullopt (= use default pool).

staryxchen, line 243: SL/TC out-of-range gets a LOG(WARNING), but "qp_pool": 42 (non-string) is silently no-op'd. For schema consistency, log a warning here too, you don't want a typo'd config to look like it took effect when it didn't.

mooncake-transfer-engine/tent/include/tent/runtime/transport_selector.h
```diff
@@ -117,6 +117,19 @@ struct SelectionPolicy {
 
     // Transport preference list (evaluated in order)
     std::vector<TransportType> transports;
+
+    // --- Link-layer QoS attributes (RFC #2519 / #2568, step-1 schema only) ---
```
staryxchen: I think add the comments about the RFC reference and splited step is not necessary

catyans:
Thanks @staryxchen — all three addressed in the latest push:

1. **Empty-string qp_pool** now treated as unset (nullopt = default pool), so a blank value doesn't look like an explicit pool.
2. **Non-string qp_pool** now logs a warning (matching the SL/TC out-of-range warnings) instead of a silent no-op.
3. **Trimmed** the RFC-reference / step-split comments in the header.

Also added a small test covering empty / non-string qp_pool → unset.

staryxchen (approved):
LGTM, thanks for your contribution.

PR #2643, opened 2026-06-26
[Store] Single Go HA-wrapper for etcd+k8s

vladnosiv:
## Description

Currently, the HA etcd and k8s backends are mutually exclusive when building, as they are two different Go modules that must be separated for a single go runtime.

This creates a strong UX problem: the pip package must have different versions for different HA backends. This makes it extremely difficult to use the k8s HA backend when integrating with SGLang/vLLM.

It is proposed to combine 2 Go modules into one HA-wrapper module, in this case, when building, work with different HA-backends will be available. Redis-backend uses C++, so there is no such problem.

A large part of the diff appeared due to pre-commit hooks on CMakeLists.txt

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
```

**Test results:**
- [ ] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used


alogfans:
A good contribution! However, you may first check CI for build problems.

Icedcoco:
Thanks for working on this. The direction makes sense to me; having one Go HA wrapper for both etcd and k8s should make the wheel story much easier.

One thing I noticed is that the Rust binding path still seems to look for the old wrapper name. In `mooncake-store/rust/build.rs`, the optional library detection still checks `etcd_wrapper`, while this PR now builds `libmooncake_ha_wrapper.so`. The current CI failures also look related to that path: the Rust test link step reports unresolved symbols such as `NewEtcdClient` and `EtcdStore*Wrapper`. Could you check whether `build.rs` should be updated to link `mooncake_ha_wrapper` instead?

I also wonder if we should avoid running `go mod tidy` from the CMake build step. Since the new `ha-wrapper` module has `go.mod` but no committed `go.sum`, the build may depend on network/module-cache state and can also modify the source tree during a normal build. Would it be better to commit the generated `go.sum` and let CMake only run `go build`?

PR #2644, opened 2026-06-27
fix(rdma): auto-chunk MRs larger than device max_mr_size (#2017)

### LINKED ISSUE #2017: [Bug]:  `ibv_reg_mr` is silently truncated to `max_mr_size`, but `BufferDesc.length` advertises the original size → `IBV_WC_REM_ACCESS_ERR`
## Summary

  When `registerLocalMemory()` is called with `length > device.max_mr_size`,
  Mooncake registers only `max_mr_size` bytes but tells peers the buffer is the
  full original length. Any RDMA WRITE to an address past the truncation
  boundary hits `IBV_WC_REM_ACCESS_ERR` (CQE error 10).

  The bug is **address-driven, not size-driven**: even a tiny 64 KiB write fails
  if its target address falls past the boundary.

  ## Root cause

  `length` is shrunk locally inside `registerMemoryRegionInternal`, but the
  caller still uses the original value when building the segment descriptor:

  **`rdma_context.cpp:212-218`**
  ```cpp
  int RdmaContext::registerMemoryRegionInternal(void *addr, size_t length, ...) {
      if (length > (size_t)globalConfig().max_mr_size) {
          PLOG(WARNING) << "shrink it to " << globalConfig().max_mr_size;
          length = (size_t)globalConfig().max_mr_size;   // local-only, value-passed
      }
      mrMeta.mr = ibv_reg_mr(pd_, addr, length, access);  // protects only first 2 GiB
  ```

  **`rdma_transport.cpp:234, 295-297`**
  ```cpp
  context_list_[i]->registerMemoryRegion(addr, length, ar);  // original length
  ...
  buffer_desc.length = length;                                // advertises full size!
  ```

  `config.cpp:387` also clamps `max_mr_size` to the device cap, so raising
  `MC_MAX_MR_SIZE` cannot work around it on hardware-limited NICs.

  ## Why this is hit on AMD MI355X (288 GB HBM)

  - NIC: AMD Pensando ionic, `max_mr_size = 2 GiB`
  - Workload: vLLM PD-disaggregated DeepSeek-R1 (MLA, 61 layers, 576 B/token)
  - Per-layer KV buffer ≈ 3.25 GiB on 288 GB HBM at `gpu-memory-utilization=0.9`
    (room for ~6 M tokens; the 2 GiB MR cutoff sits at ~3.73 M tokens)

  ```
  ionic_comp_msn:1277: cqe with error 10
  IBV_EVENT_QP_ACCESS_ERR
  ```

  This will hit any deployment where per-layer KV exceeds `max_mr_size` —
  large-HBM AMD MI300/355 today, larger GPUs in the future.

  ## Minimal reproducer

  Two nodes, one RDMA NIC each. Register a 3 GiB buffer, write to offsets
  straddling the 2 GiB boundary.

  **`target.py`**
  ```python
  import argparse, json, os, time, torch

  p = argparse.ArgumentParser()
  p.add_argument("--ip", required=True)
  p.add_argument("--buffer-gb", type=float, default=3.0)
  args = p.parse_args()

  torch.cuda.set_device(0)
  eng = TransferEngine()
  assert eng.initialize(args.ip, "P2PHANDSHAKE", "rdma", "") == 0

  n = int(args.buffer_gb * (1 << 30))
  t = torch.zeros(n, dtype=torch.uint8, device="cuda:0")
  assert eng.register_memory(t.data_ptr(), n) == 0

  json.dump({"server_name": f"{args.ip}:{eng.get_rpc_port()}",
             "base_addr": t.data_ptr()}, open("handshake.json", "w"))
  while True: time.sleep(10)
  ```

  **`initiator.py`**
  ```python
  import argparse, json, torch

  GiB, MiB, SLICE = 1 << 30, 1 << 20, 64 * 1024
  p = argparse.ArgumentParser()
  p.add_argument("--ip", required=True)
  args = p.parse_args()

  torch.cuda.set_device(0)
  hs = json.load(open("handshake.json"))
  eng = TransferEngine()
  assert eng.initialize(args.ip, "P2PHANDSHAKE", "rdma", "") == 0
  src = torch.ones(SLICE, dtype=torch.uint8, device="cuda:0")
  assert eng.register_memory(src.data_ptr(), SLICE) == 0

  for name, off in [("0", 0), ("1.9G", int(1.9*GiB)),
                    ("2G+1M", 2*GiB + MiB), ("2.5G", int(2.5*GiB))]:
      ret = eng.batch_transfer_sync_write(
          hs["server_name"], [src.data_ptr()], [hs["base_addr"] + off], [SLICE])
      print(f"  off={name:6s}  within_2G={(off+SLICE)<=2*GiB!s:5}  "
            f"ret={ret}  {'PASS' if ret==0 else 'FAIL'}")
  ```

  ### Result on ionic (`max_mr_size = 2 GiB`)

  | offset    | within MR | result |
  |-----------|-----------|--------|
  | `0`       | yes       | PASS   |
  | `1.9 GiB` | yes       | PASS   |
  | `2G + 1M` | no        | **FAIL** |
  | `2.5 GiB` | no        | **FAIL** |

  ## Suggested fixes
  1. **Auto-chunk in `registerLocalMemoryInternal`** — split buffers larger than
     `max_mr_size` into multiple MRs, publishing one `BufferDesc` per chunk.
     `BufferDesc` schema does not need to change; the existing address-based
     lookup naturally routes each WRITE to the right `rkey`.

### Before submitting...

jiejingzhangamd:
Closes #2017.

## Problem

`RdmaTransport::registerLocalMemory()` registers a single MR for the whole buffer and publishes `BufferDesc.length = full length`, but `RdmaContext::registerMemoryRegionInternal()` **silently shrinks `length` to the device `max_mr_size`** (some RoCE NICs cap this at 2 GiB) and registers an MR of only that size. The metadata then advertises bytes past the registered region, so any remote RDMA op whose **target address falls past `max_mr_size`** completes with **`IBV_WC_REM_ACCESS_ERR`**.

It is **address-driven, not size-driven**: ops to low addresses succeed; ops past the boundary fail (even small ones). Any registered buffer larger than `max_mr_size` is affected (e.g. large disaggregated-PD KV pools) — it works for a while, then fails once an op targets a high address.

## Fix

Mirror what the EFA transport already does: split a buffer larger than `max_mr_size` into `<= max_mr_size` chunks, register each as its **own MR**, and publish **one `BufferDesc` per chunk**. The per-context `rkey()/lkey()` lookups are address-range based (`findMemoryRegionContaining`), so each chunk resolves to the correct key. Chunk start-addresses are tracked (`chunk_map_`) so `unregisterLocalMemory(base_addr)` tears down every chunk; unregister continues past a per-chunk failure (reporting the first error) to avoid leaks; best-effort rollback on partial registration. Pre-touch is decided from the original buffer length, not the capped chunk length. No metadata schema change.

## Test

`tests/rdma_large_mr_test.cpp` (`RDMALargeMrTest.WritePastMaxMrSizeBoundary`): sets a small `MC_MAX_MR_SIZE` (64 MiB), registers a larger buffer (256 MiB -> 4 chunks), then issues a loopback RDMA WRITE whose target lands past the boundary. **Before:** the transfer FAILS with a remote access error. **After:** it COMPLETES and the bytes match. Runs on any RDMA device (incl. soft-RoCE / loopback) plus a metadata server; built but left out of auto-ctest (needs a device), like the sibling `rdma_loopback_test`.

## Verification (real workload)

A disaggregated prefill/decode KV transfer at high tensor-parallel degree (per-layer KV ~3.2 GiB > a 2 GiB `max_mr_size`): the patched build logged ~488 `Auto-splitting buffer (3213176832 bytes) into 2 chunks` and ran the full window with **0** remote-access-errors across a concurrency sweep; unpatched it failed mid-run with remote access errors.

### DIFF
--- a/mooncake-transfer-engine/src/transport/rdma_transport/rdma_transport.cpp
+++ b/mooncake-transfer-engine/src/transport/rdma_transport/rdma_transport.cpp
@@ -18,6 +18,7 @@
 #include <sys/mman.h>
 #include <sys/time.h>
 
+#include <algorithm>
 #include <cassert>
 #include <chrono>
 #include <cstddef>
@@ -25,6 +26,7 @@
 #include <future>
 #include <set>
 #include <thread>
+#include <utility>
 
 #include <dlfcn.h>
 
@@ -213,7 +215,6 @@ int RdmaTransport::registerLocalMemoryInternal(void *addr, size_t length,
                                                bool update_metadata,
                                                bool force_sequential) {
     (void)remote_accessible;
-    BufferDesc buffer_desc;
     const int kBaseAccessRights = IBV_ACCESS_LOCAL_WRITE |
                                   IBV_ACCESS_REMOTE_WRITE |
                                   IBV_ACCESS_REMOTE_READ;
@@ -222,40 +223,49 @@ int RdmaTransport::registerLocalMemoryInternal(void *addr, size_t length,
     if (MCIbRelaxedOrderingEnabled) {
         access_rights |= IBV_ACCESS_RELAXED_ORDERING;
     }
-    bool do_pre_touch = context_list_.size() > 0 &&
-                        std::thread::hardware_concurrency() >= 4 &&
-                        length >= (size_t)4 * 1024 * 1024 * 1024;
-    if (do_pre_touch) {
-        // Parallel Pre-touch the memory to speedup the registration process.
-        int ret = preTouchMemory(addr, length);
-        if (ret != 0) {
-            return ret;
+
+    // Mooncake#2017: ibv_reg_mr silently truncates a registration to the device
+    // max_mr_size, but the metadata would still advertise the full BufferDesc
+    // length, so any remote RDMA op past the boundary fails with
+    // IBV_WC_REM_ACCESS_ERR (ionic CQE error 10). Split buffers larger than
+    // max_mr_size into chunks of <= max_mr_size, register each as its own MR,
+    // and publish one BufferDesc per chunk (the per-context rkey/lkey lookups
+    // are address-range based, so each chunk gets the correct key).
+    size_t chunk_limit = (size_t)globalConfig().max_mr_size;
+    std::vector<std::pair<void *, size_t>> chunks;
+    if (chunk_limit > 0 && length > chunk_limit) {
+        for (size_t offset = 0; offset < length;) {
+            size_t chunk_len = std::min(chunk_limit, length - offset);
+            chunks.emplace_back(static_cast<char *>(addr) + offset, chunk_len);
+            offset += chunk_len;
         }
+        LOG(WARNING) << "Auto-splitting buffer " << addr << " (" << length
+                     << " bytes) into " << chunks.size()
+                     << " chunks of <= " << chunk_limit
+                     << " bytes each (device max_mr_size; Mooncake#2017)";
+    } else {
+        chunks.emplace_back(addr, length);
     }
 
-    /* Parallel register when:
-    1. parallel_reg_mr is enabled via MC_ENABLE_PARALLEL_REG_MR;
-    2. parallel_reg_mr not set and multiple contexts exist and memory has been
-    pre-touched
-    Note: If memory hasn't been touched, parallel register can be
-    slower. Details in: https://github.com/kvcache-ai/Mooncake/issues/848
-    Note: force_sequential is used by batch operations to avoid nested
-    parallelism.
-    */
-    int use_parallel_reg = 0;
-    if (!force_sequential) {
-        use_parallel_reg = globalConfig().parallel_reg_mr;
-        if (use_parallel_reg == -1) {
-            use_parallel_reg = context_list_.size() > 1 && do_pre_touch;
-        }
+    // Resolve the location name once, from the original buffer.
+    std::string resolved_name;
+    if (name == kWildcardLocation) {
+        bool only_first_page = true;
+        const std::vector<MemoryLocationEntry> entries =
+            getMemoryLocation(addr, length, only_first_page);
+        if (entries.empty()) return -1;
+        resolved_name = entries[0].location;
+    } else {
+        resolved_name = name;
     }
 
-    // Export a single dma_buf fd for the buffer and import it into every NIC's
-    // PD below, closing it once after all registrations. One dma_buf object
-    // shared across NICs lets the GPU driver reserve a single BAR1 window for
-    // the buffer instead of one per NIC. Host memory yields an empty export and
-    // takes the plain ibv_reg_mr path. The fd must stay open across all
-    // registrations (each MR takes its own reference); see exportDmabuf().
+    // Export a single dma_buf fd for the whole buffer and import it into every
+    // NIC's PD during each chunk's registration below (one dma_buf object
+    // shared across NICs keeps a single BAR1 window for the buffer instead of
+    // one per NIC). Host memory yields an empty export (plain ibv_reg_mr). The
+    // fd must stay open across every registration that consumes it (each MR
+    // takes its own reference), so it is closed once, on any return path, by
+    // the RAII guard below.
     DmabufExport dmabuf_exp;
     if (!context_list_.empty()) {
         int eret = RdmaContext::exportDmabuf(addr, dmabuf_exp);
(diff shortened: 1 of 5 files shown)


jiejingzhangamd:
Thanks for the review. Both points are addressed in the latest push:

1. **Pre-touch regression** — `do_pre_touch` now keys off the original buffer `length` (not the capped `chunk_len`), so the `>= 4 GiB` check fires as before and parallel pre-touch is preserved.
2. **Unregister leak** — the per-chunk unregister loop no longer returns early on a failure; it unregisters every chunk and reports the first error, so no chunk MR / metadata is leaked.

Also rebased onto latest main and clang-formatted.

mooncake-transfer-engine/src/transport/rdma_transport/rdma_transport.cpp
```diff
@@ -211,111 +212,166 @@ int RdmaTransport::registerLocalMemoryInternal(void *addr, size_t length,
...
+        if (entries.empty()) return -1;
+        resolved_name = entries[0].location;
+    } else {
+        resolved_name = name;
+    }
 
-        for (auto &thread : reg_threads) {
-            thread.join();
+    // Best-effort rollback of already-registered chunks [0, up_to_ci].
+    auto rollbackChunks = [&](size_t up_to_ci) {
```
alogfans: If failed, print a warning log.

jiejingzhangamd: Done — rollbackChunks now LOG(WARNING)s on each failed chunk cleanup (both the metadata removal and the per-context MR unregister), with the chunk address and ret code. Pushed in b71d942.

mooncake-transfer-engine/src/transport/rdma_transport/rdma_transport.cpp
```diff
@@ -211,111 +212,175 @@ int RdmaTransport::registerLocalMemoryInternal(void *addr, size_t length,
...
+    for (size_t ci = 0; ci < chunks.size(); ++ci) {
+        void *chunk_addr = chunks[ci].first;
+        size_t chunk_len = chunks[ci].second;
+
+        // Decide pre-touch from the ORIGINAL buffer length, not the capped
+        // chunk_len (which is <= max_mr_size, so a chunk_len-based >=4GiB check
+        // would never fire and silently disable parallel pre-touch).
+        bool do_pre_touch = context_list_.size() > 0 &&
+                            std::thread::hardware_concurrency() >= 4 &&
+                            length >= (size_t)4 * 1024 * 1024 * 1024;
```
staryxchen: The check is loop-invariant (depends on context_list_.size(), hardware_concurrency(), and length — all constant). Hoisting it above the loop would be cleaner and avoid redundant calls to hardware_concurrency() (which may query the OS). Not a bug, but wasteful.

mooncake-transfer-engine/tests/rdma_large_mr_test.cpp
```diff
@@ -0,0 +1,138 @@
...
+static constexpr size_t kBufferSize = 256ull << 20;  // 256 MiB -> 4 chunks
+
+class RDMALargeMrTest : public ::testing::Test {
+   public:
+    void *addr = nullptr;
+    std::unique_ptr<mooncake::TransferEngine> engine;
+
+   protected:
+    void SetUp() override {
+        google::InitGoogleLogging("RDMALargeMrTest");
```
staryxchen: google::InitGoogleLogging in SetUp() will crash if multiple test cases are added later. Consider guarding with a static flag or moving to main().

staryxchen (commented):
After this fix, the truncation at `rdma_context.cpp:349-352` should never fire for registered buffers. Consider converting it to a `LOG(ERROR) + return -1` or assert to catch regressions early, instead of silently succeeding with a shorter MR.

mooncake-transfer-engine/src/transport/rdma_transport/rdma_transport.cpp
jiejingzhangamd: Good call — done in 3b3844d. registerMemoryRegionInternal now LOG(ERROR)s and returns ERR_INVALID_ARGUMENT when length > max_mr_size, instead of silently shrinking the MR. After the registerLocalMemory chunking this path is unreachable for registered buffers, so this is purely a regression guard against a silent short MR (which would otherwise resurface as IBV_WC_REM_ACCESS_ERR past the boundary).

jiejingzhangamd:
### CI: the only red is `Build Docker Image` (runner disk exhaustion), unrelated to this change

The `Build Docker Image` job fails — and `CI Gate` is red only because it gates on that job. The failure is **not** in this PR's code. It's the runner running out of ephemeral disk while the `mooncake-ep` step pip-installs PyTorch + the CUDA wheel stack:

```
#17  ERROR: Could not install packages due to an OSError: [Errno 28] No space left on device
     [EP] Failed to install PyTorch 2.9.1.  Command run:
     '/usr/bin/python3.10 -m pip install torch==2.9.1'
```

`torch==2.9.1` (~900 MB) pulls in nvidia-cublas (594 MB), cudnn (707 MB), cusolver (267 MB), cufft (193 MB), etc., and the GitHub-hosted runner's disk fills mid-download. Reproduced on two separate runs (the second after rebasing onto latest `main`), failing at the identical step each time, so it isn't a one-off flake.

This PR cannot affect that job: the Docker build configures `-DBUILD_UNIT_TESTS=OFF`, so the RDMA change here (and the new `rdma_large_mr_test.cpp`) aren't compiled in it. The change is confined to `mooncake-transfer-engine/src/transport/rdma_transport/`.

All functional checks pass: `build (3.10)`, `build (3.12)`, `build-flags`, `build-musa`, all four `test-wheel-ubuntu`, `integration-test / test-sglang-integration`, `ascend-test`, plus format/spell/docs.

Could a maintainer re-run the job, or add a disk-cleanup step (e.g. `jlumbroso/free-disk-space`) before the EP/Docker build? Happy to open a separate infra PR for the cleanup step if preferred.

stmatengss:
CI is re-triggered

mooncake-transfer-engine/tests/rdma_large_mr_test.cpp
jiejingzhangamd: Good catch — fixed in 94f5817. Moved `InitGoogleLogging()`/`ShutdownGoogleLogging()` out of `SetUp()`/`TearDown()` and into `main()` so they run once for the binary; a second `TEST_F` will no longer abort.

mooncake-transfer-engine/src/transport/rdma_transport/rdma_transport.cpp
```diff
@@ -310,6 +362,34 @@ int RdmaTransport::unregisterLocalMemory(void *addr, bool update_metadata) {
...
+            for (auto &context : context_list_) {
+                int ret = context->unregisterMemoryRegion(cap);
+                if (ret) {
+                    LOG(ERROR) << "Failed to unregister chunk MR at " << cap;
+                    return ret;
+                }
+            }
+        }
+        return 0;
+    }
```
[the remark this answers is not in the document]
jiejingzhangamd: Already addressed in the current revision (bc1227f): the loop is best-effort — it no longer returns early, unregisters every chunk's MR + metadata even if one fails, and returns the first encountered error at the end via `first_err`. This comment was on an earlier version of the diff.

jiejingzhangamd:
@stmatengss ci still failed, but have no idea why it faield , seems json of ci script is empty ?

mooncake-transfer-engine/src/transport/rdma_transport/rdma_transport.cpp
```diff
@@ -222,111 +223,175 @@ int RdmaTransport::registerLocalMemoryInternal(void *addr, size_t length,
...
-                LOG(ERROR) << "Failed to register memory region with context "
-                           << i;
-                return ret_codes[i];
+            for (auto &thread : reg_threads) thread.join();
+
+            for (size_t i = 0; i < ret_codes.size(); ++i) {
+                if (ret_codes[i] != 0) {
+                    LOG(ERROR) << "Failed to register memory region (chunk "
+                               << ci << ") with context " << i;
+                    rollbackChunks(ci);
```
staryxchen, line 380: pass ci to rollbackChunks only after addLocalMemoryBuffer for that chunk has succeeded; otherwise ci ? ci - 1 : SIZE_MAX (no-op on ci=0)

mooncake-transfer-engine/src/transport/rdma_transport/rdma_transport.cpp
```diff
@@ -222,111 +223,175 @@ int RdmaTransport::registerLocalMemoryInternal(void *addr, size_t length,
...
-                           << i;
-                return ret;
+        } else {
+            for (size_t i = 0; i < context_list_.size(); ++i) {
+                int ret = context_list_[i]->registerMemoryRegion(
+                    chunk_addr, chunk_len, access_rights);
+                if (ret) {
+                    LOG(ERROR) << "Failed to register memory region (chunk "
+                               << ci << ") with context " << i;
+                    rollbackChunks(ci);
```
staryxchen, line 395: ditto

staryxchen (commented):
Nice fix — Leave a few things to fix before merge:

Performance (worth considering; can be follow-ups)

1. Chunked-unregister loses use_parallel_unreg. The new early-return branch at rdma_transport.cpp:405-435 iterates chunks × contexts fully sequentially; the existing parallel-unreg logic below is unreachable for chunked buffers. For a 4-chunk × 8-NIC config that's 32 sequential ibv_dereg_mr calls.
2. N metadata publishes per chunked buffer. Each chunk calls addLocalMemoryBuffer(..., update_metadata=true) on the public path, so a 4-chunk buffer publishes to the metadata server 4×. Consider passing update_metadata=false for per-chunk adds and issuing a single updateLocalSegmentDesc() at the end (mirrors the batch path at line 472).

Nits

1. The Mooncake#2017 inline markers help with archaeology but tend to rot; the PR body already carries the context. Not a blocker.
2. The chunking loop is now near-duplicated between EFA (efa_transport.cpp:274-287) and this file. Worth a follow-up to factor out — not this PR.

mooncake-transfer-engine/src/transport/rdma_transport/rdma_transport.cpp
jiejingzhangamd: no a problem now, rollbackChunks(n) takes a count and rolls back [0, n). ci is the number of already-committed chunks (ci itself isn't committed yet), so passing ci is correct and ci==0 is a no-op — no underflow.

mooncake-transfer-engine/src/transport/rdma_transport/rdma_transport.cpp
jiejingzhangamd: ditto.

jiejingzhangamd:
@staryxchen  @stmatengss Could you trigger the ci-cd, seems like not code related fail ?

staryxchen:
Hi @jiejingzhangamd Thanks for your contribution. The PR is good to me overall. Only one concern: Could you add a test covering a WRITE that straddles a chunk boundary — otherwise this fix could re-surface the same class of failure at chunk seams?

alogfans:
This PR looks good to me. After conflict resolution and CI we will merge it. @jiejingzhangamd

jiejingzhangamd:
@alogfans  test is passed. 

@staryxchen new test case is added, 

`./rdma_large_mr_test --gtest_filter=RDMALargeMrTest.WriteStraddlesChunkBoundary`

staryxchen (approved):
LGTM

jiejingzhangamd:
@alogfans @staryxchen can you trigger the merge ?

staryxchen:
> @alogfans @staryxchen can you trigger the merge ?

Merged. Thanks for your contribution!

Issue #2632, opened 2026-06-26
[Bug]: When Mooncake Store is running with offload enabled, offloading works correctly but a large number of offloading tasks expire after 600s, producing "Offloading task expired" warnings.


Colors-111:
### Bug Report

When Mooncake Store is running with offload enabled, offloading works correctly but a large number of offloading tasks expire after 600s, producing "Offloading task expired" warnings.

I tried to find the cause of the problem, but I haven't been able to pinpoint it.
```
W20260624 06:46:06.259191 381518 master_service.cpp:4716] Offloading task expired for key: a4c19c15ae875bfc15ae72d0ee8e79dc0a4b44678aac35b025d7407682fee187_3_v
W20260624 06:46:06.259195 381518 master_service.cpp:4716] Offloading task expired for key: 87b17efc3475633055bdabba14c2e5604d2bfb6dccb5ae680b3ed2a2ba46e168_3_k
W20260624 06:46:06.259200 381518 master_service.cpp:4716] Offloading task expired for key: 30e3c4277dcec6c441329500295f0a512330318cd14c9213641bba673d5d5cc3_3_v
W20260624 06:46:06.259205 381518 master_service.cpp:4716] Offloading task expired for key: 0e05a4c4854a1346768028d1f9426eed9539be61de24de721f1a0378a73a722e_2_k
W20260624 06:46:06.259209 381518 master_service.cpp:4716] Offloading task expired for key: de4e52aa62013c8f66a806d5ed957ab05cbd59160e88552c23e193467b3df224_2_v
W20260624 06:46:06.259214 381518 master_service.cpp:4716] Offloading task expired for key: 6722a24e361b4d398ef7c9ced19e13badc6e6d9f8f0da8b394784241d0d366cb_0_v
W20260624 06:46:06.259218 381518 master_service.cpp:4716] Offloading task expired for key: e86accfebc712b34f1c7412dd7c552d9b450c7743947bdae668d3e24b7af6a35_1_v
W20260624 06:46:06.259222 381518 master_service.cpp:4716] Offloading task expired for key: 746e9ac98efbbbe2bd523ef0814f9bc2c1eebb0847ab879145c68a253375f8fa_0_v
W20260624 06:46:06.259227 381518 master_service.cpp:4716] Offloading task expired for key: 109fba12aedd15db207d26b1c73a766428a89e61422237c798eff54605bcf80d_2_v
[...]

PR #2658, opened 2026-06-29
[Store]: Fix Offloading Task Orphan Causing Expired Warnings

Colors-111:
## Problem
refs to #2632 

When SSD offload is enabled (`enable_offload=true`, `offload_on_evict=false`), the master continuously prints "Offloading task expired" warnings. The phenomenon correlates with inference workload: warnings appear while inference is running and disappear when inference stops.

In production, this results in ~50K expired tasks per 10 minutes under moderate load, each holding a stale reference count on the source memory replica.

## Root Cause

There are **three silent key-drop paths** in the offload pipeline where keys are drained from the master's `offloading_queue_` but `NotifyOffloadSuccess` is never called. The orphaned `offloading_task` entries are only cleaned up after the 600s TTL expires.

### Drop Path 1: `BatchQuerySegmentSlices` all-or-nothing failure (primary)

**Before this fix**, `BatchQuerySegmentSlices` returned `ErrorCode::INVALID_KEY` if **even one key** in a tenant batch had no local memory replica. The caller in `OffloadObjects` skipped the entire tenant batch with `continue`. This is the dominant loss path because keys are queued for offload precisely due to memory pressure — by the time the worker processes them, some have already been evicted through the normal eviction path.

### Drop Path 2: `GroupOffloadingKeysByBucket` skips keys

In `storage_backend.cpp`, `GroupOffloadingKeysByBucket` silently skips keys that:
- Exceed `bucket_size_limit`, or
- Already exist on SSD (`IsExist` returns true)

Skipped keys never appear in any `bucket_keys`, so they are never passed to `BatchOffload`, and `complete_handler` is never invoked for them.

### Drop Path 3: D2H staging / SSD write partial failures

D2H copy failures and `BatchOffload` partial-write failures (`INVALID_READ`) also drop individual keys without calling `complete_handler`.

### Worker side (`file_storage.cpp`, `file_storage.h`, `storage_backend.cpp`)

**1. Make `BatchQuerySegmentSlices` best-effort**

Instead of failing the entire tenant batch when a key is missing, skip only that key and report it via an output `missing_keys` parameter.

```cpp
// Before: returns error if any key is missing
tl::expected<void, ErrorCode> BatchQuerySegmentSlices(
    const std::vector<std::string>& keys, const std::string& tenant_id,
    std::unordered_map<std::string, std::vector<Slice>>& batched_slices);

// After: always succeeds, reports missing keys
tl::expected<void, ErrorCode> BatchQuerySegmentSlices(
    const std::vector<std::string>& keys, const std::string& tenant_id,
    std::unordered_map<std::string, std::vector<Slice>>& batched_slices,
    std::vector<std::string>& missing_keys);
```

**2. Collect failed keys and NACK to master**

At the end of `OffloadObjects`, collect all keys that were drained from the master queue but could not be offloaded (from `missing_keys`, D2H failures, write failures, and keys not present in any bucket). Report them back to the master via `NotifyOffloadSuccess` with a sentinel `data_size=-1`.

```cpp
// Collect keys not in any bucket (dropped by GroupOffloadingKeysByBucket)
for (const auto& [storage_key, task] : task_by_storage_key) {
    if (all_bucket_keys.find(storage_key) == all_bucket_keys.end()) {
        failed_tasks.push_back(task);
    }
}

// NACK: notify master with data_size=-1 sentinel
if (!failed_tasks.empty()) {
    std::vector<StorageObjectMetadata> failed_metadatas(failed_tasks.size(),
        StorageObjectMetadata{-1, 0, 0, -1, ""});
    client_->NotifyOffloadSuccess(failed_tasks, failed_metadatas);
}
```

### Master side (`master_service.cpp`, `master_service.h`)

**3. Handle failed-offload sentinel in `NotifyOffloadSuccess`**

When `metadata.data_size < 0`, skip `AddReplica` but still clean up the `offloading_task` entry and release the source replica reference count.

```cpp
if (metadata.data_size < 0) {
    continue;  // NACK: offload failed, only clean up task + dec_refcnt
}
// Normal path: add LOCAL_DISK replica
Replica replica(client_id, metadata.data_size, ...);
AddReplica(client_id, object_id.user_key, object_id.tenant_id, replica);
```

**4. Clean up offloading_task in `EraseMetadata`**

When `BatchEvict` deletes metadata, the worker may still have an in-flight offload. Clean up `offloading_task` and `dec_refcnt` in `EraseMetadata` to prevent orphan tasks that only expire after TTL.

**5. Deduplicate offloading_task creation in `PutEnd`**

Before the fix, `VisitReplicas` called `PushOffloadingQueue` + `inc_refcnt` for every completed memory replica, potentially creating multiple offloading_tasks for the same key (and over-incrementing refcnt). Added a `task_created` flag to ensure one task per key.

**6. Fix iterator invalidation in `DiscardExpiredProcessingReplicas`**

`EraseMetadata` removes nodes from the map. The original code used the `erase` return value as the next iterator, but if metadata was fully deleted by `EraseMetadata`, that iterator is already invalidated. Fixed by saving `std::next(it)` before calling `EraseMetadata`.

## Verification

| Metric | Before Fix | After Fix |
|--------|-----------|-----------|
| `Offloading task expired` (10 min inference) | ~53,000 | **0** |
| Worker dropped keys | Silent (no tracking) | NACK fallback captures them |
| Master NACK processing | N/A | `data_size<0` sentinel correctly cleans up |
| `offloading_tasks remaining` (reaper) | Continuously growing | 70-88 (healthy steady state) |

### Before/After log comparison

**Before** (continuous expired warnings during inference):
```
W ... Offloading task expired for key: abc123 tenant=default
```

**After** (NACK path works, no expired tasks):
```
W ... Offloading task expired for key: abc123 tenant=default
```
(This log should no longer appear; if it does, an uncovered drop path remains)

## Files Changed

| File | Change |
|------|--------|
| `mooncake-store/src/file_storage.cpp` | Best-effort `BatchQuerySegmentSlices`; collect `failed_tasks` and NACK to master |
| `mooncake-store/include/file_storage.h` | Add `missing_keys` output parameter to `BatchQuerySegmentSlices` |
| `mooncake-store/src/storage_backend.cpp` | Demote over-limit/already-exists skip logs to `VLOG(1)` |
| `mooncake-store/src/master_service.cpp` | Handle `data_size<0` sentinel in `NotifyOffloadSuccess`; clean up offloading_task + `dec_refcnt` in `EraseMetadata`; deduplicate task creation in `PutEnd`; fix iterator invalidation in `DiscardExpiredProcessingReplicas` |

## Out of Scope (Future Work)

- `GroupOffloadingKeysByBucket` still silently skips keys that exceed `bucket_size_limit` or already exist on SSD. The NACK path in `OffloadObjects` catches these, but it would be cleaner to handle them inside `GroupOffloadingKeysByBucket` directly.
- The eviction reaper thread is greedy (always enters `BatchEvict` branch when memory pressure exists), which can starve the stale-discard branch. This is a separate concern that amplifies the impact of any orphaned tasks but is not the root cause.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
```

**Test results:**
- [ ] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used


Colors-111:
cc @LujhCoconut

mooncake-store/include/file_storage.h
```diff
@@ -108,7 +108,8 @@ class FileStorage {
 
     tl::expected<void, ErrorCode> BatchQuerySegmentSlices(
         const std::vector<std::string>& keys, const std::string& tenant_id,
-        std::unordered_map<std::string, std::vector<Slice>>& batched_slices);
+        std::unordered_map<std::string, std::vector<Slice>>& batched_slices,
+        std::vector<std::string>& missing_keys);
```
LujhCoconut: I think this PR addresses the issue in #2632 quite well. The only thing I'd like to discuss is the `missing_keys` / `missing_user_keys` parameter — from what I can see, it is only written to but never actually read in any execution logic. Is this intended for future use, or could it be removed to simplify the signature?

mooncake-store/src/master_service.cpp
LujhCoconut, line 1679: A comment on this function would be a nice-to-have. Maybe like

```cpp
  // EraseMetadata deletes the object metadata and also cleans up all
  // associated per-key state: offloading_tasks (with dec_refcnt),
  // processing_keys, replication_tasks, and promotion tasks.
  // Callers no longer need to clean these up manually before calling.
```

mooncake-store/src/file_storage.cpp
```diff
@@ -674,6 +712,10 @@ tl::expected<void, ErrorCode> FileStorage::Heartbeat() {
         return offload_result;
     }
 
+    LOG(INFO)
+        << "Successfully completed heartbeat with offloaded objects count: "
+        << offloading_objects.size();
+
```
LujhCoconut: The new `LOG(INFO)` in `Heartbeat()` logs every successful offload tick, but the existing pattern in this function is that successful paths are intentionally silent — only errors and one-time state transitions emit logs. Also, the inner `complete_handler` in `OffloadObjects` already logs via `VLOG(1) ("Success to store objects, keys count: ...")`, so this is redundant at a higher severity. I'd suggest either downgrading to VLOG(1) or removing it.

LujhCoconut (approved):
This is an important fix for SSD offloading. Thanks for this work. LGTM.

Colors-111:
> This is an important fix for SSD offloading. Thanks for this work. LGTM.

Thanks. I was planning to reply to you after the check was finished.

ykwd:
Could you resolve the conflict? Thanks

Colors-111:
> Could you resolve the conflict? Thanks

Done. But the CI always No space left on device.

PR #2661, opened 2026-06-29
[Store] Fix ConfigDict global segment size validation

### LINKED ISSUE #2660: [Bug]: ConfigDict setup rejects global_segment_size > 1TiB before segment splitting
### Summary

When Mooncake Store is configured through the `ConfigDict` setup path, `global_segment_size` values larger than `MAX_SEGMENT_SIZE` are rejected before the existing segment-splitting logic is reached.

This is inconsistent with the positional setup path used by `mooncake_store_service.py`, where `global_segment_size` is treated as total store capacity and can be split into multiple mounted segments.

### Related Issues / PRs

I searched existing issues and PRs and found several related items, but none appear to cover this exact bug:

- #903 discusses improving `global_segment_size` and `local_buffer_size` usability, including human-readable units and percentage-based input.
- #2206 robustified `ConfigDict` size parsing, including human-readable size strings and invalid/negative values, but it does not address `global_segment_size > MAX_SEGMENT_SIZE` being rejected before segment splitting.
- #1677 changes segment splitting to be transport-aware and balanced, but it focuses on the splitting policy itself, not the `ConfigDict` validation that prevents the splitting path from being reached.
- #2456 adds fail-fast validation for `MooncakeConfig` in the Python wheel, but it does not change the C++ `ConfigDict` setup validation.
- #1445 introduced the `setup(dict)` / `ConfigDict` setup path.

### Background

For this service startup path:

```bash
export MOONCAKE_GLOBAL_SEGMENT_SIZE=300gb
numactl --cpunodebind=0 --membind=0 python3 -m mooncake.mooncake_store_service --port=8099
```

`300gb` is below 1TiB, so it does not trigger splitting. I also tested larger values through the same underlying positional `MooncakeDistributedStore.setup()` path:

- `global_segment_size=1100gb`: setup succeeds and master reports 2 segments: `1024GiB + 76GiB`
- `global_segment_size=1600gb`: setup succeeds and master reports 2 segments: `1024GiB + 576GiB`

So the positional setup path can accept `global_segment_size > 1TiB` and split it into multiple mounted segments.

### Problem

The `ConfigDict` path currently validates both `global_segment_size` and `local_buffer_size` with the same single-segment limit:

```cpp
value > MAX_SEGMENT_SIZE
```

However, `global_segment_size` semantically represents total capacity, not the size of one individual mounted segment.

For example:

```cpp
config["global_segment_size"] = "1600gb";
```

This should mean "provide 1600GB total global segment capacity", and the later setup logic should split it into multiple mountable segments.

But in the current `ConfigDict` path, the value is rejected before the split loop runs.

### Reproduction

The `ConfigDict` setup path can be triggered by passing a Python dict to `MooncakeDistributedStore.setup()`.

Minimal reproduction:

```bash
python3 - <<'PY'
from mooncake.store import MooncakeDistributedStore

MAX_SEGMENT_SIZE = 1024 ** 4  # 1 TiB

store = MooncakeDistributedStore()
ret = store.setup({
    "local_hostname": "localhost:17820",
    "metadata_server": "P2PHANDSHAKE",
    "global_segment_size": str(MAX_SEGMENT_SIZE + 1),
    "local_buffer_size": "0",
    "protocol": "tcp",
    "rdma_devices": "",
    "master_server_addr": "127.0.0.1:1",
})

print("setup ret:", ret)
PY
```

This fails before reaching the segment-splitting logic. The master address does not need to be valid for this reproduction, because the failure happens during `ConfigDict` validation.

Observed log:

```text
Invalid global_segment_size: 1099511627777, must be 0 or between 1024 and 1099511627776
```

Observed return value:

```text
setup ret: -600
```

`-600` corresponds to `INVALID_PARAMS`.

### Expected Behavior

`global_segment_size` should be validated as total capacity:

- allow `0`
- allow values `>= MIN_SEGMENT_SIZE`
- do not reject only because the total capacity is larger than `MAX_SEGMENT_SIZE`

The actual mounted segment sizes should still be constrained by `max_mr_size` / `MAX_SEGMENT_SIZE`.

`local_buffer_size` should keep the existing single-buffer validation because it is a local registered buffer, not total store capacity.

### Actual Behavior

The `ConfigDict` setup path rejects `global_segment_size > MAX_SEGMENT_SIZE` before entering the existing split logic.

### Proposed Fix

Split the validation rules:

- `global_segment_size`: validate as total capacity
- `local_buffer_size`: validate as a single local buffer

### Before submitting...

bitborne:
## Description

Fixes #2660

This PR fixes `ConfigDict` setup validation for `global_segment_size`.

`global_segment_size` represents total store capacity. The later setup path already splits this total capacity into mountable segments when it exceeds `max_mr_size`.

However, the `ConfigDict` setup path used the same single-segment upper-bound validation for both `global_segment_size` and `local_buffer_size`. As a result, `global_segment_size > MAX_SEGMENT_SIZE` was rejected before the segment-splitting logic could run.

This PR separates the validation rules:

- `global_segment_size`: validate as total capacity, allowing `0` or values at least `MIN_SEGMENT_SIZE`
- `local_buffer_size`: keep the existing single-buffer validation, including the `MAX_SEGMENT_SIZE` upper bound

A regression test case is added to ensure `local_buffer_size > MAX_SEGMENT_SIZE` is still rejected after relaxing only `global_segment_size`.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
build/mooncake-store/tests/pybind_client_test \
  --gtest_filter=RealClientTest.ErrSetupWithInvalidConfigDictSize \
  --gtest_brief=1 \
  --logtostderr=1
```

**Test results:**
- [x] Unit tests pass

Output:
```text
[==========] 1 test from 1 test suite ran. (1062 ms total)
[  PASSED  ] 1 test.
```

## Checklist


yokinoshitayoki:
Could we add a regression test that directly covers the main fixed behavior: a ConfigDict setup with `global_segment_size > MAX_SEGMENT_SIZE` should no longer fail during size validation?

bitborne:
Done. I added a regression test that directly covers this behavior.

The new test sets `global_segment_size` to `MAX_SEGMENT_SIZE + 1` through `ConfigDict`, with `local_buffer_size = 0`, and then uses an invalid protocol to stop setup before any large segment allocation happens. The test verifies that setup reaches protocol validation and does not fail with `Invalid global_segment_size`, so it directly checks that `global_segment_size > MAX_SEGMENT_SIZE` is no longer rejected during ConfigDict size validation.

yokinoshitayoki (approved):
LGTM.

ykwd:
@bitborne code format check failed, could you resolve it?

bitborne:
Fixed, thanks. It was just the long TEST_F line in the new test. I ran the format check locally and it passes now.

pre-commit also runs through the format hooks now; the only remaining report is a pre-existing codespell typo in real_client.cpp outside this PR diff.

ykwd:
Could you take a look at the failed ci:

The following tests FAILED:
	 46 - pybind_client_test (Failed)
Errors while running CTest

bitborne:
Fixed, thanks. The failure came from the regression test using the invalid-protocol path after the branch update; it could go into a later setup path and crash under ASan.

I changed the test to use a valid tcp config and fail later with an unreachable master address. This still verifies that `global_segment_size > MAX_SEGMENT_SIZE` is not rejected by ConfigDict size validation, without allocating the large segment or hitting that unstable path.

The latest CI rerun is green now.

ykwd (approved):

PR #2668, opened 2026-06-29
[Store] Fix snapshot restore loses segment protocol and allocator replica_type

### LINKED ISSUE #2636: [Bug]:  Snapshot restore loses segment protocol and allocator replica_type
### Description

  After master restart with snapshot restore (`enable_snapshot_restore=1`), all memory segments lose their `protocol` field. The admin API returns `"protocol": ""` even though the segment was originally 
  mounted with `"protocol": "tcp"`.

  Additionally, `OffsetBufferAllocator::replica_type_` is also lost during serialization, which affects NoF segment metric accounting (restored NoF segments are counted under "mem" instead of "nof").

### Reproduction

  1. Start Mooncake master with snapshot enabled
  2. Mount a segment (e.g. protocol=tcp)
  3. Wait for a snapshot to be generated
  4. Restart the master
  5. Query the admin API:

  ```bash
  curl http://${MOONCAKE_MASTER_URL}/get_segments_detail | python3 -m json.tool

  Output:

  {
      "total_segments": 1,
      "segments": [
          {
              "segment_name": "xxx.xxx.xxx.148:13695",
              "segment_id": "6649975730718913833-14683276544698645121",
              "client_id": "9101436600682018864-15666146750897147799",
              "base_address": "0x773c56000000",
              "size_bytes": 3355443200,
              "size_human": "3.125 GiB",
              "te_endpoint": "xxx.xxx.xxx.148:13695",
              "protocol": "",        # ← empty after restore
              "status": "OK",
              "allocator_used_bytes": 3165446144,
              "allocator_capacity_bytes": 3355443200,
              "allocator_usage_percent": 94.34
          }
      ]
  }
```
  Note that protocol is empty even though the segment was mounted with --protocol tcp.

### Root Cause

  The snapshot serialization code in serializer.cpp uses custom msgpack serializers (not YLT_REFL auto-serialization) for both MountedSegment and OffsetBufferAllocator. These custom serializers missed two
  fields:

  1. MountedSegment::segment.protocol missing

  Serialize<MountedSegment> (serializer.cpp:829-868) packs 8 elements:
  - segment.id, segment.name, segment.base, segment.size, segment.te_endpoint
  - status, has_buffer_allocator, buffer_allocator_data
  
  But the Segment struct has 6 fields: id, name, base, size, te_endpoint, protocol. protocol is never serialized.

  After restore, segment.protocol defaults to empty string "".

  2. OffsetBufferAllocator::replica_type_ missing
  
  Serialize<OffsetBufferAllocator> (serializer.cpp:930-954) packs 6 elements:
  - segment_name, base, total_size, cur_size, transport_endpoint, offset_allocator
  
  But OffsetBufferAllocator has replica_type_ which determines whether metrics are counted as MEMORY or NOF_SSD. replica_type_ is never serialized.

  After restore, the allocator is reconstructed with the default ReplicaType::MEMORY, so any NoF segment's metrics are incorrectly counted under memory.

  ### Fix Options

  Option A: Add missing fields to serialization (no backward compat)

  Add the fields at the natural position in the serialization arrays:

  - MountedSegment: add protocol after te_endpoint in both serialize/deserialize, change array size from 8→9
  - OffsetBufferAllocator: add replica_type_ after transport_endpoint_ in both serialize/deserialize, change array size from 6→7

  Old snapshots become unreadable, but given that the restore cleanup bug (#TODO) already makes old snapshots unreliable, this is acceptable.

  Option B: Backward-compatible fix (more complex)

  Append the new fields at the end of the arrays and add size detection in deserialization:
  - Deserialize accepts both old (8) and new (9) element counts for MountedSegment
  - Deserialize accepts both old (6) and new (7) element counts for OffsetBufferAllocator
  
  This preserves the ability to read old snapshots but adds complexity.

### Before submitting...

### PR DESCRIPTION
**Pull request by Icedcoco**
## Description

This PR fixes #2636 by preserving the fields that were previously dropped from the snapshot msgpack serializers. `MountedSegment.segment.protocol` is now appended to the serialized `MountedSegment` payload, and `OffsetBufferAllocator.replica_type_` is now appended to the serialized allocator payload. Both fields are added as trailing elements so existing snapshots remain readable: legacy 8-element `MountedSegment` records default `protocol` to an empty string, and legacy 6-element `OffsetBufferAllocator` records default `replica_type` to `ReplicaType::MEMORY`, matching the previous behavior.

The change intentionally keeps the scope small: it fixes the serializer round-trip for the missing fields and adds focused regression coverage. It does not expand the service-level snapshot format for `nof_segment_manager_`.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

Added new tests covering the changed behavior:

- `SerializerTest.MountedSegmentProtocolRoundTrip`: verifies `MountedSegment.segment.protocol` survives serialize/deserialize and the new format has 9 elements.
- `SerializerTest.MountedSegmentLegacyFormatCompat`: verifies legacy 8-element `MountedSegment` snapshots remain readable and default `protocol` to empty.
- `SerializerTest.OffsetBufferAllocatorReplicaTypeRoundTrip`: verifies `OffsetBufferAllocator.replica_type_` survives serialize/deserialize and the new format has 7 elements.
- `SerializerTest.OffsetBufferAllocatorLegacyFormatCompat`: verifies legacy 6-element allocator snapshots remain readable and default `replica_type` to `ReplicaType::MEMORY`.
- `MasterServiceSSDSnapshotTest.RestorePreservesSegmentProtocol`: verifies service-level snapshot/restore preserves `protocol` as observed through `GetSegmentsDetail()`.

**Test commands:**
```bash
LSAN_OPTIONS=detect_leaks=0 ./build/mooncake-store/tests/serializer_test
LSAN_OPTIONS=detect_leaks=0 ./build/mooncake-store/tests/master_service_ssd_test_for_snapshot --gtest_filter=MasterServiceSSDSnapshotTest.RestorePreservesSegmentProtocol
```

Both passed.

**Test results:**
- [x] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

### DIFF
--- a/mooncake-store/src/serialize/serializer.cpp
+++ b/mooncake-store/src/serialize/serializer.cpp
@@ -9,6 +9,19 @@
 
 namespace mooncake {
 
+namespace {
+
+bool IsKnownMountedSegmentProtocol(const std::string &value) {
+    return value == "tcp" || value == "rdma" || value == "efa" ||
+           value == "cxl" || value == "cxi" || value == "ub" ||
+           value == "ascend" || value == "ubshmem" || value == "sunrise_link" ||
+           value == "rpc_only" || value == "nvmeof" || value == "hip" ||
+           value == "nvlink" || value == "nvlink_intra" || value == "maca" ||
+           value == "barex";
+}
+
+}  // namespace
+
 // Node serialization size constants for offset_allocator::__Allocator
 constexpr size_t OFFSET_ALLOCATOR_NODE_BOOL_SIZE = 1;
 constexpr size_t OFFSET_ALLOCATOR_NODE_UINT32_COUNT = 6;
@@ -831,9 +844,13 @@ tl::expected<void, SerializationError> Serializer<MountedSegment>::serialize(
     // Use array structure for packing, more efficient
     // Format: [segment_id, segment_name, segment_base, segment_size,
     // te_endpoint, status, has_buffer_allocator, buffer_allocator_data,
-    // host_id]
+    // host_id, protocol]
+    //
+    // protocol and host_id were independently added as a 9th field on two
+    // branches. New snapshots use size==10 to keep both fields unambiguous,
+    // with host_id first because it entered main before protocol.
 
-    packer.pack_array(9);
+    packer.pack_array(10);
 
     // Serialize Segment info
     packer.pack(UuidToString(mounted_segment.segment.id));
@@ -857,6 +874,7 @@ tl::expected<void, SerializationError> Serializer<MountedSegment>::serialize(
                 return tl::unexpected(result.error());
             }
             packer.pack(mounted_segment.segment.host_id);
+            packer.pack(mounted_segment.segment.protocol);
             return {};
         }
     }
@@ -864,6 +882,7 @@ tl::expected<void, SerializationError> Serializer<MountedSegment>::serialize(
     packer.pack(false);  // Mark no valid buffer allocator exists
     packer.pack_nil();
     packer.pack(mounted_segment.segment.host_id);
+    packer.pack(mounted_segment.segment.protocol);
     return {};
 }
 
@@ -876,10 +895,18 @@ Serializer<MountedSegment>::deserialize(const msgpack::object &obj) {
                                "state: not a msgpack array"));
     }
 
-    if (obj.via.array.size < 8) {
+    // Snapshot format is versioned by array size:
+    //   size == 8: legacy snapshot, no protocol or host_id field
+    //   size == 9: compatibility snapshot, either protocol or host_id
+    //   size == 10: current snapshot, with host_id and protocol
+    // Anything else is invalid.
+    if (obj.via.array.size != 8 && obj.via.array.size != 9 &&
+        obj.via.array.size != 10) {
         return tl::unexpected(SerializationError(
             ErrorCode::DESERIALIZE_FAIL,
-            "deserialize MountedSegment invalid array size"));
+            fmt::format("deserialize MountedSegment invalid array size: "
+                        "expected 8, 9, or 10, got {}",
+                        obj.via.array.size)));
     }
 
     MountedSegment mounted_segment;
@@ -920,8 +947,16 @@ Serializer<MountedSegment>::deserialize(const msgpack::object &obj) {
                 return tl::unexpected(allocatorResult.error());
             }
         }
-        if (obj.via.array.size >= 9) {
+        if (obj.via.array.size == 9) {
+            std::string value = array[8].as<std::string>();
+            if (IsKnownMountedSegmentProtocol(value)) {
+                mounted_segment.segment.protocol = std::move(value);
+            } else {
+                mounted_segment.segment.host_id = std::move(value);
+            }
+        } else if (obj.via.array.size == 10) {
             mounted_segment.segment.host_id = array[8].as<std::string>();
+            mounted_segment.segment.protocol = array[9].as<std::string>();
         }
     } catch (const std::exception &e) {
         return tl::unexpected(SerializationError(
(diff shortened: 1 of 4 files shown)


mooncake-store/src/serialize/serializer.cpp
```diff
@@ -920,8 +947,16 @@ Serializer<MountedSegment>::deserialize(const msgpack::object &obj) {
...
             }
         }
-        if (obj.via.array.size >= 9) {
+        if (obj.via.array.size == 9) {
+            std::string value = array[8].as<std::string>();
+            if (IsKnownMountedSegmentProtocol(value)) {
+                mounted_segment.segment.protocol = std::move(value);
+            } else {
+                mounted_segment.segment.host_id = std::move(value);
+            }
```
00fish0, line 956: The 9-element format currently has two historical meanings:

- The format merged into main stores host_id at index 8.
- An earlier, pre-rebase revision of this PR stored protocol at index 8.

The current code tries to distinguish them by checking whether the value matches a known protocol:
```
if (IsKnownMountedSegmentProtocol(value)) {
    mounted_segment.segment.protocol = std::move(value);
} else {
    mounted_segment.segment.host_id = std::move(value);
}
```
However, host_id is derived from the hostname and is not constrained from being "tcp", "rdma", "cxl", etc. A valid snapshot produced by main with host_id == "tcp" would therefore be silently restored as:

```
host_id = ""
protocol = "tcp"
```
This affects more than the admin output: host_id is used to rebuild segments_by_host_, so the restored segment may lose local-first/host-affinity behavior.

Since both 9-element layouts encode the last field as an untagged string, they cannot be distinguished reliably from the record itself. Unless snapshots produced by the pre-rebase revision of this PR are an explicit compatibility requirement, I suggest treating all 9-element records according to the format merged into main:

```
if (obj.via.array.size == 9) {
    mounted_segment.segment.host_id = array[8].as<std::string>();
} else if (obj.via.array.size == 10) {
    mounted_segment.segment.host_id = array[8].as<std::string>();
    mounted_segment.segment.protocol = array[9].as<std::string>();
}
```

If both layouts really need to be supported, an explicit snapshot version or record tag is required; matching against a protocol allowlist cannot resolve the ambiguity safely.

mooncake-store/src/serialize/serializer.cpp
```diff
@@ -937,9 +972,12 @@ Serializer<OffsetBufferAllocator>::serialize(
...
     // Use array structure to pack OffsetBufferAllocator
     // Format: [segment_name, base, total_size, current_size,
-    // transport_endpoint, offset_allocator]
+    // transport_endpoint, offset_allocator, replica_type]
+    //
+    // replica_type is appended at the end so that older snapshots (size==6)
+    // remain deserializable. New snapshots use size==7. See issue #2636.
 
-    packer.pack_array(6);
+    packer.pack_array(7);
```
00fish0, line 980: The wire format changes, but the snapshot manifest version remains 1.0.0.

This changes OffsetBufferAllocator from a 6-element record to a 7-element record. The new reader can read both formats, but the previous reader requires exactly 6 elements and will reject snapshots produced by this PR.

Since SNAPSHOT_SERIALIZER_VERSION is still "1.0.0", an older binary will accept the manifest as compatible and then fail later while deserializing the allocator record.

This makes compatibility asymmetric:
```
new reader <- old snapshot: works
old reader <- new snapshot: fails
```
This may affect rollback, mixed-version HA/rolling upgrades, or an older standby consuming a snapshot written by a newer primary.

Could we clarify the required compatibility contract? If old readers must be able to detect or handle the new format, we should introduce a new snapshot/record version and let the new reader explicitly support both versions. If only “new reader reads old snapshot” is supported, that limitation should at least be documented because the unchanged 1.0.0 manifest currently suggests that the formats are compatible.
An old-reader/new-writer compatibility test would also help make the intended behavior explicit.

00fish0 (commented):
The replica_type_ change fixes the serializer round-trip in isolation, but nof_segment_manager_ is not persisted by the service snapshot path. Therefore, this does not yet make NoF segments or their metrics survive a master restart. So service-level NoF restoration remains follow-up work.

If there is another production path that persists a NOF_SSD allocator and I missed it, please point me to it.

PR #2671, opened 2026-06-29
[Wheel] Optimize strucutured object transfer

zxpdemonio:
## Motivation                                                                                                                                                
                                                                                                                                                            
  Optimize Mooncake structured DataProto rollout transfer. The main goal is to reduce CPU serialization overhead, avoid large Python object materialization, and keep the RL-side integration thin.    

### Why replace register buffer into memcpy when put

Rollout dicts contain many lists, dicts, and nested combinations of the two, resulting in a large number of small, memory-discontiguous objects. Registering each small object individually as an RDMA buffer and then putting it to Mooncake would incur significant latency — each register call requires NIC interaction at microsecond-level cost, and the overhead accumulates quickly with many objects, far exceeding that of parallel memcpy. The current strategy is therefore parallel memcpy: copy scattered small objects into a contiguous pre-registered buffer, then transfer via a single batch put. Once a high-performance batch register mechanism is available (bulk-registering multiple discontiguous memory regions in one call), we can consider switching to a zero-copy path using register + batch_put_from_multi_buffers, eliminating the memcpy overhead entirely.                                                                                                             
                                                                                                                                                            
 ##  Modification                                                                                                                                              
                                                                                                                                                            
  - Legacy dict Wrapping: Add legacy dict support by wrapping.
  - Schema-driven encoding (FieldSchema): Add FieldSchema dataclass allowing callers to specify non_tensor_batch field codecs explicitly via field_schemas parameter in put_dataproto /put_legacy_dict
   / append_dataproto_fields, bypassing expensive per-sample runtime inference on large ragged fields.                                                                               
  - BufferPool-based GET path: Remove all manual register_buffer / unregister_buffer calls (broken on erdma hardware) and unify GET-side RDMA memory registration through BufferPool:
    - _read_chunks_with_pool_get_into: full pool acquire → batch_get_into → copy to destination                                                                                      
    - _read_chunks_with_chunked_pool_get_into: per-chunk pool acquire fallback when full allocation fails
    - _read_payload_range_with_pool: pool-based ranged read via get_into_ranges                                                                                                      
    - _read_payload_ranges_with_pool: pool-based multi-range read                                                                                                                    
    - Final fallback: store.get per chunk (non-zero-copy)                                                                                                                                                
                                                                                                                                                                                                              
  ## Performance 
  ### roll dataproto(cross-machine RDMA, erdma)                                                                                                             
                                                                                                                                                                                     
  | Dataset | Samples | Data Size | PUT | GET (avg, skip warmup) |                                                                                                                   
  |---------|---------|-----------|-----|------------------------|                                                                                                                   
  | ROLL DataProto (ragged_tensor_dict) | 2048 | ~24 GB | 2.6s | 2.66s |                                                                                                             

### slime rollout dict , cross-machine (eRDMA, A10 cluster):                                                                                          
                                                            
| Size | Ray PUT | Ray GET | Ray GET BW | MoonCake PUT | MoonCake PUT BW | MoonCake GET | MoonCake GET BW | GET Speedup |
|------|---------|---------|------------|--------|-----------|--------|-----------|-------------|                       
| 128MB | 61ms | 65ms | 1.9 GB/s | 30ms | 3.97 GB/s | 18ms | 6.5 GB/s | 3.6x |
| 512MB | 207ms | 211ms | 2.4 GB/s | 71ms | 6.72 GB/s | 59ms | 8.0 GB/s | 3.6x |                                        
| 1GB | 296ms | 432ms | 2.3 GB/s | 130ms | 7.35 GB/s | 114ms | 8.3 GB/s | 3.8x |                                        
| 4GB | 340ms | 1561ms | 2.6 GB/s | 477ms | 7.99 GB/s | 430ms | 8.9 GB/s | 3.6x |                 
| 16GB | 708ms | 5802ms | 2.8 GB/s | 1903ms | 8.02 GB/s | 1701ms | 9.0 GB/s | 3.4x |                                                                                                                                           
                                                                                                                                                            
  ## Test                                                                                                                                                      
                                                                                                                                                            
  - Added / updated structured object store unit coverage for:                                                                                              
    - DataProto structured transfer                                                                                                                         
    - typed ragged ndarray/list preservation                                                                                                                
    - row-subset structured non-tensor decode                                                                                                               
    - BufferPool-backed tensor object transfer                                                                                                              
  - Local validation:                                                                                                                                       
    - python3 -m py_compile mooncake-wheel/mooncake/structured_object_store.py                                                                              
    - focused structured object store tests passed: 5 passed                  
## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [x] Performance improvement
- [ ] Other

## Checklist

## AI Assistance Disclosure



- [ ] No AI tools were used

### DIFF
--- a/.typos.toml
+++ b/.typos.toml
@@ -11,6 +11,7 @@ wqs = "wqs"
 hsa = "hsa"
 Optin = "Optin"
 HPE = "HPE"
+thw = "thw"
 
 [files]
 extend-exclude = [
--- a/mooncake-wheel/mooncake/structured_object_store.py
+++ b/mooncake-wheel/mooncake/structured_object_store.py
@@ -1420,7 +1420,7 @@ class MooncakeBundleTransfer:
         """Store a legacy rollout dict as a structured object."""
         return self._put_stage(
             None,
-            _legacy_dict_to_dataproto_envelope(data),
+            _legacy_dict_to_envelope(data),
             namespace=namespace,
             partition=partition,
             stage=stage,
@@ -1432,7 +1432,7 @@ class MooncakeBundleTransfer:
 
     def get_legacy_dict(self, ref: DataProtoRefLike) -> dict[str, Any]:
         """Materialize a legacy rollout dict stored by put_legacy_dict()."""
-        return _dataproto_envelope_to_legacy_dict(
+        return _envelope_to_legacy_dict(
             self._materialize_ref(_resolve_ref(ref))
         )
 
@@ -1761,7 +1761,7 @@ LEGACY_DICT_META_FIELDS = frozenset(
 )
 
 
-def _legacy_dict_to_dataproto_envelope(data: Mapping[str, Any]) -> dict[str, Any]:
+def _legacy_dict_to_envelope(data: Mapping[str, Any]) -> dict[str, Any]:
     if not isinstance(data, Mapping):
         raise TypeError("legacy rollout data must be a mapping")
     partition = data.get("partition")
@@ -1789,7 +1789,7 @@ def _legacy_dict_to_dataproto_envelope(data: Mapping[str, Any]) -> dict[str, Any
     }
 
 
-def _dataproto_envelope_to_legacy_dict(data: Mapping[str, Any]) -> dict[str, Any]:
+def _envelope_to_legacy_dict(data: Mapping[str, Any]) -> dict[str, Any]:
     result = dict(_mapping_to_dict(data.get("meta_info")))
     result.update(_mapping_to_dict(data.get("batch")))
     for key, value in _mapping_to_dict(data.get("non_tensor_batch")).items():
--- a/mooncake-wheel/mooncake/structured_object_store.py
+++ b/mooncake-wheel/mooncake/structured_object_store.py
@@ -55,7 +55,7 @@ _INFER_MAX_LIST_LEN = 8
 
 
 class BundleStore(Protocol):
-    def put(self, key: str, value: Any) -> int: ...
+    def put(self, key: str, value: Any, config: Any = None) -> int: ...
 
     def get(self, key: str) -> bytes: ...
 
@@ -576,6 +576,7 @@ class MooncakeBundleTransfer:
         chunk_bytes: Optional[int] = None,
         policy: Optional[BundleTransferPolicy] = None,
         max_inflight_put: Optional[int] = None,
+        config: Any = None,
     ) -> RemoteBundleRef:
         """Store raw metadata bytes plus named buffers as a low-level bundle."""
         return self._bundle_store.put_bundle(
@@ -585,6 +586,7 @@ class MooncakeBundleTransfer:
             chunk_bytes=chunk_bytes,
             policy=policy,
             max_inflight_put=max_inflight_put,
+            config=config,
         )
 
     def remove_bundle(self, ref: RemoteBundleRef | Mapping[str, Any]) -> None:
@@ -598,6 +600,7 @@ class MooncakeBundleTransfer:
         chunk_bytes: Optional[int] = None,
         policy: Optional[BundleTransferPolicy] = None,
         max_inflight_put: Optional[int] = None,
+        config: Any = None,
     ) -> RemoteBundleRef:
         """Store a structured object described by JSON metadata plus named members."""
         return self._structured_store.put_structured_object(
@@ -606,6 +609,7 @@ class MooncakeBundleTransfer:
             chunk_bytes=chunk_bytes,
             policy=policy,
             max_inflight_put=max_inflight_put,
+            config=config,
         )
 
     def read_spec(
@@ -626,6 +630,67 @@ class MooncakeBundleTransfer:
         """Materialize a structured object read spec into caller-provided destinations when possible."""
         return self._structured_store.materialize_into(spec, destinations)
 
+    def put(
+        self,
+        data: Any,
+        *,
+        type: Literal["dataproto", "dict"] = "dataproto",
+        namespace: str = "default",
+        partition: str = "default",
+        stage: str = "default",
+        chunk_bytes: Optional[int] = None,
+        policy: Optional[BundleTransferPolicy] = None,
+        field_schemas: Optional[Mapping[str, FieldSchema]] = None,
+        config: Any = None,
+    ) -> MooncakeDataProtoRef:
+        """Store a DataProto-like object or flat dict as a structured object."""
+        if type == "dataproto":
+            stage_data = data
+        elif type == "dict":
+            stage_data = _flat_dict_to_envelope_with_schema(data, field_schemas)
+        else:
+            raise ValueError(f"unsupported Mooncake payload type: {type!r}")
+        return self._put_stage(
+            None,
+            stage_data,
+            namespace=namespace,
+            partition=partition,
+            stage=stage,
+            chunk_bytes=chunk_bytes,
+            policy=policy,
+            overwrite=False,
+            field_schemas=field_schemas,
+            config=config,
+        )
+
+    def get(
+        self,
+        ref: DataProtoRefLike,
+        *,
+        type: Literal["dataproto", "dict"] = "dataproto",
+        fields: Optional[Sequence[str]] = None,
+        batch_fields: Optional[Sequence[str]] = None,
+        non_tensor_fields: Optional[Sequence[str]] = None,
+        meta_info_keys: Optional[Sequence[str]] = None,
+        data_cls: Optional[Any] = None,
+        destinations: Optional[Mapping[str, Any]] = None,
+        rows: slice | StructuredMemberSlice | Sequence[int] | None = None,
+    ) -> Any:
+        """Materialize a structured object as a DataProto-like object or flat dict."""
+        if type not in {"dataproto", "dict"}:
+            raise ValueError(f"unsupported Mooncake payload type: {type!r}")
+        return self._materialize_ref(
+            _resolve_ref(ref),
+            fields=fields,
+            batch_fields=batch_fields,
+            non_tensor_fields=non_tensor_fields,
+            meta_info_keys=meta_info_keys,
+            data_cls=data_cls,
+            destinations=destinations,
+            rows=rows,
+            flat_dict_output=(type == "dict"),
+        )
+
     def put_dataproto(
         self,
         data: Any,
@@ -636,18 +701,19 @@ class MooncakeBundleTransfer:
         chunk_bytes: Optional[int] = None,
         policy: Optional[BundleTransferPolicy] = None,
         field_schemas: Optional[dict[str, FieldSchema]] = None,
+        config: Any = None,
     ) -> MooncakeDataProtoRef:
         """Store a DataProto-like object as a stage-level structured object."""
-        return self._put_stage(
-            None,
+        return self.put(
             data,
+            type="dataproto",
             namespace=namespace,
             partition=partition,
             stage=stage,
             chunk_bytes=chunk_bytes,
             policy=policy,
-            overwrite=False,
             field_schemas=field_schemas,
+            config=config,
         )
 
     def get_dataproto(
@@ -663,8 +729,9 @@ class MooncakeBundleTransfer:
         rows: slice | StructuredMemberSlice | Sequence[int] | None = None,
     ) -> Any:
         """Materialize selected DataProto fields from structured object refs."""
-        return self._materialize_ref(
-            _resolve_ref(ref),
+        return self.get(
+            ref,
+            type="dataproto",
             fields=fields,
             batch_fields=batch_fields,
             non_tensor_fields=non_tensor_fields,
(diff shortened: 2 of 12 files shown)


mooncake-wheel/mooncake/structured_object_store.py
```diff
@@ -2460,13 +3011,27 @@ def _read_bytes_member(
...
+        expected_bytes = int(payload_spec.get("bytes", 0))
+        if expected_bytes > 0:
+            result = self._bundle_store.read_payload_range_into_pool_array(
+                payload_spec, np.dtype(np.uint8), (expected_bytes,), 0
+            )
+            if result is not None:
+                return result
+            # Pool array failed; use np.empty + chunked pool read (ctypes.memmove)
+            # instead of bytearray + bytes() which causes extra copies.
+            result = np.empty(expected_bytes, dtype=np.uint8)
```
yokinoshitayoki: This breaks generic bytes materialization. When BufferPool is unavailable, this path allocates a numpy uint8 array and then `read_payload_into()` falls back to `_read_chunk_with_get()`, which assigns a `bytes` object directly into an ndarray slice and raises `ValueError`. This also changes the public behavior of bytes fields from returning `bytes` to returning `np.ndarray`. Existing cases like `put_object({"raw": b"abc"}).get_object()` and `test_structured_object_roundtrip` now fail. Please keep generic bytes fields returning `bytes`, and use a separate internal path if DataProto payloads need pool-backed byte storage.

mooncake-wheel/mooncake/structured_object_store.py
```diff
@@ -1066,6 +1311,10 @@ def read_bytes(payload_name: str, byte_start: int, byte_end: int) -> bytes:
             base = int(offsets[0])
             limit = int(offsets[-1])
             offsets = offsets - base
+            if codec == "typed_ragged" and metadata.get("row_format") == "mixed":
+                metadata["ndarray_rows"] = (metadata.get("ndarray_rows") or [])[
+                    start:end
+                ]
             return {
                 "data": read_member("data", base, limit),
```
yokinoshitayoki, line 1413: This row-slice path still assumes ragged `data` is an ndarray member that supports `slice_member()`. After this PR, `_MultiBufferPayload` is encoded as a generic `bytes` member, so `read_member("data", base, limit)` reaches `_read_bytes_member()` with a member slice and raises `ValueError: structured bytes member ... does not support slicing`. This breaks DataProto row slicing for structured recursive / ragged non-tensor fields; `test_dataproto_helper_reads_dict_of_tensors_rows_and_selected_field` reproduces it. Please either keep these payloads as sliceable ndarray members with dtype/shape metadata, or read byte ranges directly and reconstruct the typed buffer.

mooncake-wheel/mooncake/structured_object_store.py
```diff
@@ -844,33 +1008,80 @@ def read_data_ranges(
                 for key in ("missing_payload", "row_mask_payload", "lengths_payload"):
                     payload_name = node.get(key)
                     if payload_name is not None:
-                        recursive_payload[payload_name] = read_member_indices(
-                            payload_name, indices
-                        )
+                        recursive_payload[payload_name] = read_member_indices()
```
yokinoshitayoki: This call is missing the required arguments. `read_member_indices()` is defined to require `payload_name` and `member_indices`, so any structured_recursive non-tensor field read with explicit row indices will raise `TypeError` here. This should pass the current payload name and the selected indices, e.g. `read_member_indices(payload_name, indices)`.

mooncake-wheel/mooncake/structured_object_store.py
```diff
@@ -2026,6 +2395,91 @@ def _decode_ragged_tensor_values(payload: dict[str, Any], rows: int) -> list[Any
...
+        return _EncodedStructuredLeaf(
+            codec="ragged_tensor_dict",
+            rows=rows,
+            payload={"null_mask": null_mask},
+            metadata={"keys": [], "key_codecs": {}},
+        )
+
+    keys = schema.metadata.get("keys")
+    if not keys:
+        keys = sorted(non_null[0].keys())
```
yokinoshitayoki: Inferring `ragged_tensor_dict` keys from only the first non-null row can silently drop data. If later rows contain keys that are not present in the first row, those tensors are never encoded and are lost on decode. For rollout/routed-expert payloads, per-row key sets can differ. Please either require `FieldSchema.metadata["keys"]` for this codec or infer the union of keys across all non-null rows and validate them.

mooncake-wheel/tests/test_put_get_tensor.py
```diff
@@ -237,110 +237,112 @@ def test_zero_tensor_with_tp(self):
...
+        self.assertEqual(d.codec, "ragged_tensor")
 
     def test_numeric_sequence(self):
-        d = _choose_leaf_codec([[1, 2, 3], [4, 5]])
+        d = _leaf_decision([[1, 2, 3], [4, 5]])
         self.assertTrue(d.accepted)
         self.assertEqual(d.codec, "typed_ragged")
 
     def test_bytes(self):
-        d = _choose_leaf_codec([b"hello", b"world"])
```
stmatengss, line 305: Why do we need to chanage the API name?

zxpdemonio: No API was renamed. This PR adds a new set of **legacy_dict** interfaces alongside the existing dataproto ones:

- `put_legacy_dict` / `get_legacy_dict` / `remove_legacy_dict`: for flat-dict rollout data (used by slime). They share the same underlying primitives as the dataproto APIs but handle the flat-dict <-> envelope conversion automatically.
- `release_result`: deterministic pool buffer release after GET.
- `export_ref` / `import_ref`: aliases for ref serialization without dataproto-specific naming, so legacy_dict callers do not need to reference dataproto concepts.

The existing `put_dataproto` / `get_dataproto` / `cleanup_dataproto` APIs are unchanged.

zxpdemonio: There are some import API additions：
  ## legacy_dict series:                                                                                                     
                                                                                                                          
  legacy_dict and dataproto serve different data models. DataProto is a structured format with data pre-organized into {batch, non_tensor_batch, meta_info} — the caller already knows which fields are dense tensors, which are non-tensor, and which are metadata. legacy_dict is a flat dict with mixed types (tensors, lists, scalars, nested dicts) where the caller does not distinguish between these categories. The legacy_dict interfaces handle the automatic classification on PUT (splitting a flat dict into the envelope by inspecting field types and shapes) and the reverse flattening on GET.
  Without these interfaces, every flat-dict caller would have to replicate this classification logic themselves, leaking mooncake's internal envelope structure into upstream code.

  ## release_result:                                                                                                         
   
  GET results may be backed by BufferPool pre-registered RDMA memory for zero-copy reads. Without explicit release, pool buffers are only freed when GC invokes __del__, which is non-deterministic. With rollout data at 24GB+, delayed release causes pool exhaustion and allocation failures. release_result provides deterministic pool lease return after the caller is done with the data. It handles both flat dicts (legacy_dict) and nested envelope dicts (dataproto) by recursively walking Mapping values.

  ## export_ref / import_ref:                                                                                                
   
  export_dataproto_ref / import_dataproto_ref are generic ref serialization functions — they work identically for both dataproto and legacy_dict refs (both return MooncakeDataProtoRef). The aliases remove the dataproto naming so legacy_dict callers do not need to reference dataproto concepts in their code, keeping the API boundary clean.          
                  
  ## field_schemas:

  Without explicit schemas, the encoder auto-detects field types and may choose suboptimal codecs. For example, specifying typed_ragged schema enables zero-copy scatter-gather encoding for ragged tensor fields instead of concatenation-based encoding, which is critical for large rollout payloads.

zxpdemonio: `_choose_leaf_codec` was renamed to `_leaf_decision` because the function semantics changed — it no longer just picks a codec, it now returns a `_CodecDecision` with accept/reject status, reasoning, and metadata. The old name implied it always succeeds in choosing a codec, but the function can also reject values. `_leaf_decision` better reflects the accept/reject decision-making role. Happy to revert the rename if you prefer keeping the original name.

mooncake-wheel/mooncake/structured_object_store.py
yokinoshitayoki: This fallback still writes a bytes slice directly into a NumPy destination. When the fast get_into_ranges/pool path is unavailable, NumPy treats the bytes object as a scalar and raises ValueError instead of copying the byte values. This currently breaks indexed DataProto reads, including test_dataproto_helper_reads_dict_of_tensors_rows_and_selected_field. Please convert the source slice with np.frombuffer(..., dtype=np.uint8) or use ctypes.memmove for ndarray destinations; the same pattern in _read_chunk_with_get should be fixed as well.

.typos.toml
```diff
@@ -10,6 +10,7 @@ wqs = "wqs"
 # ROCm dmabuf MR registration path.
 hsa = "hsa"
 HPE = "HPE"
+thw = "thw"
```
yokinoshitayoki, line 14: Is it necessary?

mooncake-wheel/mooncake/structured_object_store.py
```diff
@@ -2976,38 +3980,62 @@ def __init__(self, store: BundleStore, buffer_pool: Any = None) -> None:
...
         self._get_into_ranges = getattr(store, "get_into_ranges", None)
-        self._register_buffer = getattr(store, "register_buffer", None)
-        self._unregister_buffer = getattr(store, "unregister_buffer", None)
+
+    def _ensure_buffer_pool(self) -> Any:
+        if self._buffer_pool is None:
+            try:
+                from mooncake.buffer_pool import BufferPool
+
+                self._buffer_pool = BufferPool(self._store)
```
yokinoshitayoki, line 4542: Please guard native-store compatibility before constructing BufferPool. MooncakeBundleTransfer accepts the generic BundleStore protocol (put/get/remove), but this line passes every such object to the native constructor. In the repository's source-tree test setup, where conftest combines this module with an installed mooncake extension, MooncakeBundleTransfer(InMemoryStore()).put_object(...) segfaults or aborts before this exception handler can run. Please only auto-create a pool for a verified Mooncake store, or require callers to provide buffer_pool explicitly; generic stores should remain on the direct put/get fallback.

.typos.toml
zxpdemonio: Keeping this entry because the spelling gate flags `thw` without it. It appears in the PR benchmark/context text, so removing it would make the Typos check fail.

yokinoshitayoki (changes requested):
Thanks for your hard work, but this PR has grown beyond a reviewable performance change. It now combines new legacy-dict/DataProto APIs, schema and codec changes, BufferPool-backed GET semantics, multi-buffer PUT logic, a new native C extension, packaging changes, and a large refactor of existing code.

The current diff is over 3,000 additions and 1,300 deletions, while the >500 LOC RFC checklist item remains unchecked. More importantly, the latest refactor no longer collects the structured-object tests and removes multiple public APIs that still exist on main.

I do not think these changes can be reviewed or validated safely as one unit. Could we consider close this PR with a series of smaller PRs?

And we may need each PR to preserve the existing test suite and provide benchmarks for the specific optimization it introduces.

zxpdemonio:
> Thanks for your hard work, but this PR has grown beyond a reviewable performance change. It now combines new legacy-dict/DataProto APIs, schema and codec changes, BufferPool-backed GET semantics, multi-buffer PUT logic, a new native C extension, packaging changes, and a large refactor of existing code.
> 
> The current diff is over 3,000 additions and 1,300 deletions, while the >500 LOC RFC checklist item remains unchecked. More importantly, the latest refactor no longer collects the structured-object tests and removes multiple public APIs that still exist on main.
> 
> I do not think these changes can be reviewed or validated safely as one unit. Could we consider close this PR with a series of smaller PRs?
> 
> And we may need each PR to preserve the existing test suite and provide benchmarks for the specific optimization it introduces.

@yokinoshitayoki 
Thanks for the review. I agree that this PR has grown large and should be split.

Over the past two weeks, I used rollout data from three RL frameworks, ROLL, Slime, and Miles, as optimization targets. In the continuous optimization process, the code changes have also grown increasingly large. Their rollout payloads have quite different characteristics:

- ROLL: DataProto-based payload, about 24.3GB total, dominated by large multimodal ragged tensor objects (`multi_modal_inputs`), with dense tensor batch data plus large nested non-tensor fields.
- Slime: legacy rollout dict captured from real rollout data, scaled from real samples to 16GB+, mostly typed ragged arrays such as tokens, loss masks, log probs, routed experts, and scalar metadata.
- Miles: legacy rollout dict with very large sample count, 1.25M samples for the 4GB case, many Python-list-backed scalar fields plus typed ragged fields. Its PUT path is more sensitive to Python serialization/encoding overhead than pure RDMA transfer.

The work in this PR optimizes serialization/deserialization and transfer for these different shapes: large objects, small object sequences, discrete Python objects, typed ragged arrays, ragged tensor dictionaries, and pool-backed RDMA buffers. It adds schema-guided encoding, DataProto/legacy-dict transfer APIs, BufferPool-backed materialization, deterministic result release, chunked/inflight PUT, and a fast scatter-gather copy path.

After the latest slimming pass, all three framework benchmarks are at or better than the existing baseline, so I think the optimization target has been reached and this PR is now ready to be split into smaller reviewable PRs

yokinoshitayoki:
@zxpdemonio Thanks a lot, it's an amazing job. I'm looking forward to your new PRs.

zxpdemonio:
The pr splitted: 
https://github.com/kvcache-ai/Mooncake/pull/2907
https://github.com/kvcache-ai/Mooncake/pull/3013
https://github.com/kvcache-ai/Mooncake/pull/3023
https://github.com/kvcache-ai/Mooncake/pull/3024
https://github.com/kvcache-ai/Mooncake/pull/3055

PR #2283, opened 2026-06-01
[Store] RemoveAll not deleting SSD offload files, enable storage_backend_->RemoveAll() in Client::RemoveAll

Colors-111:
## Description

`Client::RemoveAll` previously only removed metadata from MasterService, leaving SSD offload files as orphans on disk. Enable `storage_backend_->RemoveAll()` to delete both metadata and actual disk files, preventing disk space leak.

## Module

- [ ] Transfer Engine (`mooncake-transfer-engine`)
- [X] Mooncake Store (`mooncake-store`)
- [ ] Mooncake EP (`mooncake-ep`)
- [ ] Integration (`mooncake-integration`)
- [ ] P2P Store (`mooncake-p2p-store`)
- [ ] Python Wheel (`mooncake-wheel`)
- [ ] PyTorch Backend (`mooncake-pg`)
- [ ] Mooncake RL (`mooncake-rl`)
- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other

## Checklist

- [ ] I have performed a self-review of my own code.
- [ ] I have formatted my own code using `./scripts/code_format.sh` before submitting.
- [ ] I have updated the documentation.
- [ ] I have added tests to prove my changes are effective.

PR #2676, opened 2026-06-30
[Store] Fix RemoveAll not cleaning SSD offload files on any node via PollRemoveAll RPC

Colors-111:
## Description

Follow-up to #2283. The previous PR enabled local SSD cleanup on the client, but SSD offload files on store nodes were still leaked — master erased the metadata and returned,leaving each client's LOCAL_DISK replicas orphaned on disk.

This PR adds master-coordinated cross-node cleanup: when `RemoveAll` runs, master flags every client that owns a LOCAL_DISK replica; each remote client polls the flag via the new `PollRemoveAll` RPC and clears its own SSD on the next heartbeat.

## Problem

`curl -X POST .../hicache/storage-backend/clear` (sglang → `MooncakeStore.clear()` → `Client::RemoveAll`) only cleaned:
- Master metadata (memory + disk replica entries)

But SSD offload files on  store nodes were never cleaned. After a clear,  orphaned files remained on disk across the cluster, while `file_cache_nums` / `mem_cache_nums` counters in master hadalready been decremented — inconsistent with physical state.

## Approach

Master has no direct RPC channel to push clients, so we reuse the existing heartbeat-pull pattern (`OffloadObjectHeartbeat`, `PromotionObjectHeartbeat`):

1. **Master `RemoveAll`** — while iterating metadata, collect the set of `client_id`s that own LOCAL_DISK replicas. After erasing metadata, set `pending_remove_all = true` on each of those clients `LocalDiskSegment`. (For RemoveAll, we don't need per-key tracking — a single bool flag suffices, since each client clears its whole SSD.)

2. **New `PollRemoveAll` RPC** — client calls this on each heartbeat. Master atomically checks-and-clears the flag and returns `bool`.

3. **`FileStorage::Heartbeat`** — STEP 2 calls `PollRemoveAll`; if true, calls `FileStorage::RemoveAll()` (wipes `storage_backend_` + the storage directory). This covers all store nodes.

4. **Initiator stays synchronous** — `Client::RemoveAll` keeps calling `storage_backend_->RemoveAll()` for fsdir, and the `SetRemoveAllCleanup` callback (→ `FileStorage::RemoveAll()`) for SSD offload files, both synchronously. `PollRemoveAll` on the initiator is a redundant-but-idempotent backstop (next heartbeat clears an already-empty directory).

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
```

**Test results:**
- [ ] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used


mooncake-store/src/storage_backend.cpp
```diff
@@ -3256,6 +3286,50 @@ tl::expected<void, ErrorCode> OffsetAllocatorStorageBackend::ScanMeta(
...
+            LOG(WARNING) << "RemoveAll: failed to truncate data file: "
+                         << data_file_path_;
+        }
+    }
+    if (capacity_ > 0) {
+        allocator_ = offset_allocator::OffsetAllocator::create(0, capacity_);
+    }
+
+    LOG(INFO) << "OffsetAllocatorStorageBackend::RemoveAll: cleared all data";
+}
```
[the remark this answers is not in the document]
Colors-111: `SharedMutexLocker(SharedMutex* mu)` with a single argument takes the exclusive mode (`is_exclusive=true`, calls `mut->lock()` — see mutex.h:200). The shared-mode constructor requires a second `shared_lock_t` argument. So `map.clear()` is under an exclusive lock and has no data race. The class comment "acquires a shared mutex" means it acquires a SharedMutex-type object, not that it acquires in shared mode.

mooncake-store/src/master_service.cpp
```diff
@@ -3969,10 +3969,18 @@ auto MasterService::RemoveByRegex(const std::string& regex_pattern,
...
     auto now = std::chrono::system_clock::now();
 
+    // Collect client_ids that own LOCAL_DISK replicas, so we can signal
+    // them to physically clear their SSD via pending_remove_all.
+    std::unordered_set<UUID, boost::hash<UUID>> clients_with_disk_replicas;
+
+    ScopedLocalDiskSegmentAccess local_disk_segment_access =
+        segment_manager_.getLocalDiskSegmentAccess();
+    auto& client_local_disk_segment =
+        local_disk_segment_access.getClientLocalDiskSegment();
```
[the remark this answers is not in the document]
Colors-111: Done — deferred acquisition to after the shard loop, and only when clients_with_disk_replicas is non-empty.

mooncake-store/src/master_service.cpp
```diff
@@ -4003,21 +4026,34 @@ long MasterService::RemoveAll(bool force) {
...
     }
 
+    // Signal each client that owns disk replicas to clear their SSD.
+    for (const auto& client_id : clients_with_disk_replicas) {
+        auto seg_it = client_local_disk_segment.find(client_id);
+        if (seg_it != client_local_disk_segment.end()) {
+            MutexLocker locker(&seg_it->second->offloading_mutex_);
+            seg_it->second->pending_remove_all = true;
+        }
+    }
```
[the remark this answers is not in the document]
Colors-111: Done — wrapped in `if (!clients_with_disk_replicas.empty())`.

mooncake-store/src/master_service.cpp
```diff
@@ -4047,10 +4098,18 @@ long MasterService::RemoveAll(const std::string& tenant_id, bool force) {
         }
     }
 
+    for (const auto& client_id : clients_with_disk_replicas) {
+        auto seg_it = client_local_disk_segment.find(client_id);
+        if (seg_it != client_local_disk_segment.end()) {
+            MutexLocker locker(&seg_it->second->offloading_mutex_);
+            seg_it->second->pending_remove_all = true;
+        }
+    }
```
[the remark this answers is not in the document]
Colors-111: Done — same pattern applied to the tenant-scoped overload.

stmatengss:
Good catch! It is an unsolved bug.

stmatengss:
@qoder review this PR per-line

Colors-111:
Hi @LujhCoconut @Lin-z-w 

Could you please take a look at this PR when you are free? I'd appreciate any feedback.

Thanks!

Colors-111:
> Hi @LujhCoconut @Lin-z-w
> 
> Could you please take a look at this PR when you are free? I'd appreciate any feedback.
> 
> Thanks!

@stmatengss @ykwd

mooncake-store/src/master_service.cpp
```diff
@@ -4422,13 +4439,23 @@ long MasterService::RemoveAll(bool force) {
...
+        ScopedLocalDiskSegmentAccess local_disk_segment_access =
+            segment_manager_.getLocalDiskSegmentAccess();
+        auto& client_local_disk_segment =
+            local_disk_segment_access.getClientLocalDiskSegment();
+        for (auto& [client_id, segment] : client_local_disk_segment) {
+            MutexLocker locker(&segment->offloading_mutex_);
+            segment->pending_remove_all = true;
+        }
+    }
+
```
LujhCoconut: I found a semantic gap here. `long MasterService::RemoveAll(const std::string& tenant_id, bool force)` involves multi-tenancy, but the current implementation seems to only correctly delete the tenant's metadata while incorrectly deleting all tenants' data, which may cause cross-tenant data loss.

```cpp
  // Signal ALL clients to clear their SSD — overlaps with metadata deletion.
  for (auto& [client_id, segment] : client_local_disk_segment) {
      MutexLocker locker(&segment->offloading_mutex_);
      segment->pending_remove_all = true;   // no tenant filtering
  }
```

mooncake-store/src/file_storage.cpp
```diff
@@ -725,6 +731,43 @@ tl::expected<void, ErrorCode> FileStorage::Heartbeat() {
...
+        for (auto it = fs::directory_iterator(storage_dir, ec);
+             it != fs::directory_iterator(); it.increment(ec)) {
+            if (ec) {
+                LOG(WARNING) << "RemoveAll: directory iteration failed: "
+                             << ec.message();
+                break;
+            }
+            const auto& entry = *it;
+            std::error_code remove_ec;
+            fs::remove_all(entry.path(), remove_ec);
```
LujhCoconut: `FileStorage::RemoveAll()` deletes the data file that the backend just recreated (OffsetAllocator only).

The ordering here is reversed:
1. It first calls `storage_backend_->RemoveAll()` (line 740). For `OffsetAllocatorStorageBackend::RemoveAll()` (storage_backend.cpp:3309), this truncates and recreates `kv_cache.data` via O_TRUNC and keeps the `fd` in `data_file_`, ready to accept new writes.
2. It then unconditionally iterates `config_.storage_filepath` and `fs::remove_alls` every entry (lines 750-767). Since `data_file_path_ = storage_path_/kv_cache.data` (storage_backend.cpp:2755,1260) lives in that same directory, the freshly recreated file is deleted again.

Impact: after `RemoveAll()` returns, `data_file_` holds an `fd` to an unlinked inode. Subsequent offload writes "succeed" but land in an invisible deleted file — lost on restart (`ScanMeta` finds nothing), with disk accounting out of sync.

Fix: either wipe the directory before the backend rebuilds, or exclude the backend's data file from the wipe. (Note: `BucketStorageBackend` / `StorageBackend` (FilePerKey) don't recreate files, so for them the directory wipe is merely redundant — not harmful. Only `OffsetAllocatorStorageBackend` is affected.)

mooncake-store/src/storage_backend.cpp
```diff
@@ -3276,6 +3306,53 @@ tl::expected<void, ErrorCode> OffsetAllocatorStorageBackend::ScanMeta(
...
+        shards_[i].map.clear();
+    }
+    total_size_.store(0, std::memory_order_relaxed);
+    total_keys_.store(0, std::memory_order_relaxed);
+
+    // Truncate the data file and rebuild the allocator so the backend
+    // is ready for new writes (same logic as Init's fresh-start path).
+    if (data_file_) {
+        data_file_.reset();
+    }
```
LujhCoconut: Use After Free: `RemoveAll()` destroys `data_file_` while lock-free readers are mid-I/O on it.

`BatchLoad` builds its read plan under per-shard locks, then does disk I/O with no locks held (storage_backend.cpp:3067-3114). It pins each entry's `allocation` (refcounted `shared_ptr`) into the ReadPlan so the offset can't be recycled mid-read (:3104-3107) — but that only protects the offset. The `data_file_` handle itself, a bare `std::unique_ptr<StorageFile>` (storage_backend.h:1200), is dereferenced without synchronization in phase 2 (data_file_->vector_read(...) at :3120/:3143/:3164).

`RemoveAll()` then does `data_file_.reset()` under all shard locks (:3318-3342). The comment claims those locks guard against a concurrent `reset()`, but that only holds RemoveAll-vs-RemoveAll: a reader releases its shard lock at the end of phase 1 (:3109), so RemoveAll acquires the locks uncontended and destroys the `StorageFile` while an in-flight reader is still calling `vector_read` on it. The lock windows never overlap. Result: use-after-free on a destructed `StorageFile` / closed fd, plus a data race on the non-atomic unique_ptr. Reads run on RPC/coroutine threads (FileStorage::BatchGet), `RemoveAll` on the heartbeat/initiator thread (client_service.cpp:2627) — genuinely concurrent. Newly introduced here (pre-PR, `data_file_.reset()` was `Init()`-only, single-threaded).

 Fix: make `data_file_` a `std::shared_ptr<StorageFile>` and copy it into each `ReadPlan` under the shard lock, mirroring the existing `allocation` pinning. A concurrent `reset()` then just rebinds the member while the in-flight reader keeps the old file alive until its I/O completes.

mooncake-store/src/master_service.cpp
Colors-111: Good catch. Reverted the tenant-scoped overload to collect only clients owning that tenant's LOCAL_DISK replicas before signaling. The global `RemoveAll(bool)` keeps the broadcast-to-all optimization since it's a full clear.

mooncake-store/src/file_storage.cpp
Colors-111: Fixed. `FileStorage::RemoveAll` now delegates entirely to `storage_backend_->RemoveAll()` and drops the redundant directory wipe — each backend owns its on-disk layout.

mooncake-store/src/storage_backend.cpp
Colors-111: Fixed. `data_file_` is now `shared_ptr<StorageFile>`, pinned into each `ReadPlan` (BatchLoad) and captured per-batch (BatchStore) so a concurrent `RemoveAll` rebuild can't destroy it mid-I/O.

Colors-111:
Hi @LujhCoconut ,I've addressed all the comments. Feel free to review it again when you're free!

Colors-111:
Hi @LujhCoconut @Lin-z-w , Feel free to review it when you're free!

LujhCoconut:
@Colors-111 Sorry for the long delay — I'll review this PR within the next two days.

mooncake-store/src/master_service.cpp
```diff
@@ -4945,6 +4990,8 @@ auto MasterService::OffloadObjectHeartbeat(const UUID& client_id,
         // MetadataAccessorRW.
         offloading_objects_copy =
             std::move(local_disk_segment_it->second->offloading_objects);
+        // Drain pending_remove_all when offloading is disabled.
+        local_disk_segment_it->second->pending_remove_all = false;
```
LujhCoconut: This line should be removed. When a node hits `KEYS_ULTRA_LIMIT`, `enable_offloading_` is permanently set to `false` and never recovers. From that point on, every `OffloadObjectHeartbeat` enters the `else` branch and unconditionally clears `pending_remove_all` to `false` here. Since `FileStorage::Heartbeat()` calls `OffloadObjectHeartbeat` in STEP 1 *before* `PollRemoveAll` in STEP 2, the flag master just set via `RemoveAll()` is wiped before `PollRemoveAll` can read it — `PollRemoveAll` returns `false` forever, and `FileStorage::RemoveAll()` is never called on the remote node.

`PollRemoveAll` already performs its own check-and-clear on `pending_remove_all` (L5031-5032), which is the sole intended consumer of this flag. `OffloadObjectHeartbeat` has no reason to touch it.

mooncake-store/src/storage_backend.cpp
```diff
@@ -1562,6 +1562,12 @@ tl::expected<void, ErrorCode> StorageBackendAdaptor::ScanMeta(
     return {};
 }
 
+void StorageBackendAdaptor::RemoveAll() {
+    if (storage_backend_) {
+        storage_backend_->RemoveAll();
+    }
+}
+
```
LujhCoconut, line 1640: `StorageBackendAdaptor::RemoveAll()` delegates to the underlying backend but does not reset its own `total_keys` / `total_size` counters (declared at `storage_backend.h:750-752`, guarded by `mutex_`). These are the same counters checked by `IsEnableOffloading()` at L1472-1473:
```cpp
total_keys <= file_storage_config_.total_keys_limit &&
total_size <= file_storage_config_.total_size_limit;
```
After `RemoveAll`, the SSD is empty, but `IsEnableOffloading()` still returns `false` because the adaptor-level counters reflect the pre-cleanup state. This prevents the node from re-entering offload until a process restart (which rescans metadata and rebuilds counters from scratch). Applies only to the `kFilePerKey` backend type — `OffsetAllocatorStorageBackend::RemoveAll()` already correctly resets its own counters.

Suggested fix: acquire `mutex_` and zero both counters at the end:
```cpp
void StorageBackendAdaptor::RemoveAll() {
    if (storage_backend_) {
        storage_backend_->RemoveAll();
    }
    MutexLocker lock(&mutex_);
    total_keys = 0;
    total_size = 0;
}
```

mooncake-store/src/file_storage.cpp
```diff
@@ -859,6 +865,20 @@ tl::expected<void, ErrorCode> FileStorage::Heartbeat() {
...
+    //  - OffsetAllocatorStorageBackend truncates+rebuilds its data file
+    //    (fs::remove_all on the dir here would unlink the freshly rebuilt
+    //    file, leaving data_file_ pointing at a dead inode — see PR #2676).
+    //  - BucketStorageBackend deletes all buckets.
+    //  - StorageBackend (FilePerKey) removes all files under root_dir_.
+    if (storage_backend_) {
+        storage_backend_->RemoveAll();
+    }
+    LOG(INFO) << "FileStorage::RemoveAll: cleared storage backend";
+}
```
LujhCoconut, line 882: Not a blocker.

The master-side `RemoveAll(tenant_id, force)` correctly scopes `pending_remove_all` signaling to clients holding `LOCAL_DISK` replicas for the target tenant (`master_service.cpp:4680-4701`). But the client-side `FileStorage::RemoveAll()` is tenant-unaware — it calls `storage_backend_->RemoveAll()`, which performs a global wipe of the entire storage directory regardless of tenant.

In multi-tenant deployments where a node hosts SSD data for multiple tenants, a tenant-scoped `RemoveAll("tenant_A")` will physically delete `tenant_B`'s files as a side effect, while master still holds valid metadata for `tenant_B`. Subsequent reads will hit `OBJECT_NOT_FOUND` on the storage node. Shared-nothing deployments and the global `RemoveAll(force)` overload are unaffected. This needs either tenant-aware `RemoveAll` in the storage backend layer, or at minimum a documented limitation. It's worth deeper consideration of the multi-tenant scenario and `RemoveAll` usage patterns in a follow-up — but this isn't a blocker for the current PR.

LujhCoconut:
Please solve the merge conflicts. All my concerns are shown above. Thanks for your patience.

mooncake-store/src/master_service.cpp
Colors-111: Fixed. Removed the `pending_remove_all = false`.
`PollRemoveAll` already check-and-clears, so this line was just harmful.

mooncake-store/src/storage_backend.cpp
Colors-111: Fixed. Reset `total_keys/total_size` to 0 in `StorageBackendAdaptor::RemoveAll()` under lock, so `IsEnableOffloading()` reflects the empty disk and offloading can resume without restart.

mooncake-store/src/file_storage.cpp
Colors-111: Backends store files without a tenant dimension, so this is a global wipe — safe for `RemoveAll(force)` and single-tenant/shared-nothing deployments. Per-tenant physical isolation needs backend-level work; added a TODO.

Colors-111:
> Please solve the merge conflicts. All my concerns are shown above. Thanks for your patience.

Thanks for your review, I've resolved the merge conflicts and comment on your concern. Feel free to review it again when you're free!

LujhCoconut (approved):
Thanks for this work. It looks good to me now.

Colors-111:
Hi @ykwd ,all changes have been completed. Please help merge the PR at your convenience.

ykwd (approved):
Thanks for the fix

PR #2678, opened 2026-06-30
[CI] publish master image to Docker Hub via manual workflow

tpiperatgod:
## Description

Adds a manually-triggered (`workflow_dispatch`) GitHub Actions workflow that builds a
multi-architecture (`linux/amd64,linux/arm64`) image from an **already-published**
`mooncake-transfer-engine` wheel and pushes it to `docker.io/kvcacheai/mooncake`,
tagged with the wheel version (and optionally `:latest`).

Two files:
- **`docker/master.Dockerfile`** — a no-GPU image: a stub `libcuda.so.1` satisfies dynamic
  linking, it installs the published wheel, drops the torch EP/PG extensions (`ep_*/pg_*`),
  and runs as uid 65532 under `tini`. Usable to run `mooncake_master` / the HTTP metadata
  server / import checks on nodes without a GPU or driver.
- **`.github/workflows/publish-master-image.yaml`** — `setup-qemu` + `buildx` + Docker Hub
  login + `build-push-action` + an amd64 `import` smoke test.

Inputs: `mooncake_version` (required; also the image tag), `pip_index_url`
(default `https://pypi.org/simple`), `tag_latest` (default `true`).

**Manual-only trigger** is intentional: the image installs a published wheel, so it must run
*after* that version's wheel is on the index. Listening on `v*` tags would race the
release workflow that publishes the wheel in the same tag event.

> Note: this `docker/master.Dockerfile` installs a published wheel and is distinct from a
> source-built master image — same role, different build method.

**Prerequisites for running the workflow:**
- The target `mooncake_version` wheel is already published to `pip_index_url`.
- Repo secret `DOCKERHUB_TOKEN` is set to an organization access token (Read & Write) for
  `kvcacheai/mooncake`. Login username is the org name `kvcacheai` (hardcoded, not a secret).

## Module

- [x] CI/CD
- [x] Other — container image / deployment

## Type of Change

- [x] New feature

## How Has This Been Tested?

Static validation has been done locally; the end-to-end run (multi-arch build, push, smoke)
requires the `DOCKERHUB_TOKEN` secret and is exercised by running the workflow after merge
(a `workflow_dispatch` workflow only appears on the default branch).

**Test commands (static, run locally):**
```bash
# Workflow YAML parses
python3 -c "import yaml; yaml.safe_load(open('.github/workflows/publish-master-image.yaml'))"

# Dockerfile has no host-path COPY (build context-independent; only --from=<stage>)
grep -nE '^\s*COPY\s' docker/master.Dockerfile
```

**Post-merge acceptance (to run via `workflow_dispatch`):**
```bash
gh workflow run publish-master-image.yaml -f mooncake_version=<published-version> -f tag_latest=true
docker buildx imagetools inspect docker.io/kvcacheai/mooncake:<published-version>   # expect amd64 + arm64
# workflow's smoke step prints: ok
```

**Test results:**
- [ ] Unit tests pass — N/A (no unit-testable code; CI workflow + Dockerfile only)
- [x] Manual testing done — workflow YAML parses; Dockerfile build-args match the workflow,
      `file:` path resolves, no context `COPY`. Full build/push/smoke pending the first
      `workflow_dispatch` run (needs the secret).

## Checklist

- [ ] I have formatted my code using `./scripts/code_format.sh` — N/A (Dockerfile + YAML)
- [ ] I have added tests to prove my changes are effective — verification is the workflow's
      own amd64 import smoke step
- [x] For changes >500 LOC: I have filed an RFC issue — N/A (126 LOC)

## AI Assistance Disclosure

Claude Code assisted with the design, the workflow/Dockerfile authoring, and review. The
submitter is responsible for understanding and defending all changes.

### DIFF
--- a/.github/workflows/publish-master-image.yaml
+++ b/.github/workflows/publish-master-image.yaml
@@ -1,0 +1,89 @@
+name: Publish Master Image
+
+# Manual trigger only: this installs an already-published PyPI wheel, so it must be
+# run after that version's wheel is published — avoids racing release.yaml, which builds
+# and publishes the wheel in the same v* tag event.
+on:
+  workflow_dispatch:
+    inputs:
+      mooncake_version:
+        description: "Published mooncake-transfer-engine version (also the image tag), e.g. 0.3.11.post1"
+        required: true
+        type: string
+      pip_index_url:
+        description: "pip index URL"
+        required: false
+        default: "https://pypi.org/simple"
+        type: string
+      tag_latest:
+        description: "Also tag :latest"
+        required: false
+        default: true
+        type: boolean
+
+permissions:
+  contents: read
+
+jobs:
+  publish-master:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+
+      - name: Free up disk space
+        run: |
+          sudo rm -rf /usr/share/dotnet
+          sudo rm -rf /opt/ghc
+          sudo rm -rf /opt/hostedtoolcache/CodeQL
+          sudo rm -rf /usr/local/lib/android
+          df -h
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Log in to Docker Hub
+        uses: docker/login-action@v3
+        with:
+          registry: docker.io
+          # For an org access token (OAT), the login username is the org name
+          # (Docker: docker login --username <ORG_NAME>) — not a secret.
+          username: kvcacheai
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Compute image tags
+        id: tags
+        run: |
+          REPO="docker.io/kvcacheai/mooncake"
+          TAGS="${REPO}:${{ inputs.mooncake_version }}"
+          if [ "${{ inputs.tag_latest }}" = "true" ]; then
+            TAGS="${TAGS}"$'\n'"${REPO}:latest"
+          fi
+          {
+            echo "tags<<EOF"
+            echo "${TAGS}"
+            echo "EOF"
+          } >> "$GITHUB_OUTPUT"
+
+      - name: Build and push (multi-arch)
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          file: docker/master.Dockerfile
+          platforms: linux/amd64,linux/arm64
+          push: true
+          build-args: |
+            MOONCAKE_VERSION=${{ inputs.mooncake_version }}
+            PIP_INDEX_URL=${{ inputs.pip_index_url }}
+          tags: ${{ steps.tags.outputs.tags }}
+
+      - name: Smoke test (amd64 import)
+        run: |
+          docker run --rm --platform linux/amd64 \
+            "docker.io/kvcacheai/mooncake:${{ inputs.mooncake_version }}" \
+            python3 -c "import mooncake.engine, mooncake.store; print('ok')"
--- a/.github/workflows/publish-master-image.yaml
+++ b/.github/workflows/publish-master-image.yaml
@@ -3,6 +3,12 @@ name: Publish Master Image
 # Manual trigger only: this installs an already-published PyPI wheel, so it must be
 # run after that version's wheel is published — avoids racing release.yaml, which builds
 # and publishes the wheel in the same v* tag event.
+#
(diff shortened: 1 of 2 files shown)


tpiperatgod:
Hi @Gregory-Pereira ! I built on your PR here — thanks for the groundwork. The main change on my side is the install method: instead of building mooncake from source, I install it from the published Python wheel, which keeps the image build much lighter. Since this clearly started from your work, I'd like to credit you as a co-author if you're on board.

Icedcoco:
Thanks for adding this. The manual flow makes sense to me, but I’m not fully sure the current publish path has the guardrails we want for an official Docker Hub image.

A few things I’d consider before merging: since `mooncake_version` is interpolated directly into the shell step that computes tags, it may be safer to pass it through `env` and validate it as a version string first. `tag_latest` also defaults to `true`, so a backfill or retry for an older wheel could accidentally move `:latest`; I’m not sure whether that is intended.

I’m also a bit worried that the workflow pushes the multi-arch image before running the smoke test. If the import check fails, the tag has already been published. Maybe we should smoke-test before promoting the final tag, or push a temporary tag first. Relatedly, the workflow publishes both `linux/amd64` and `linux/arm64`, but only smoke-tests amd64; given the glibc/arm64 note in the Dockerfile, it would be good to confirm whether we want an arm64 smoke under QEMU as well.

One smaller question: do we want `pip_index_url` to be arbitrary for the official image publishing workflow, or should this be restricted/confirmed somehow so the image always comes from the expected package index?

tpiperatgod:
> Thanks for adding this. The manual flow makes sense to me, but I’m not fully sure the current publish path has the guardrails we want for an official Docker Hub image.
> 
> A few things I’d consider before merging: since `mooncake_version` is interpolated directly into the shell step that computes tags, it may be safer to pass it through `env` and validate it as a version string first. `tag_latest` also defaults to `true`, so a backfill or retry for an older wheel could accidentally move `:latest`; I’m not sure whether that is intended.
> 
> I’m also a bit worried that the workflow pushes the multi-arch image before running the smoke test. If the import check fails, the tag has already been published. Maybe we should smoke-test before promoting the final tag, or push a temporary tag first. Relatedly, the workflow publishes both `linux/amd64` and `linux/arm64`, but only smoke-tests amd64; given the glibc/arm64 note in the Dockerfile, it would be good to confirm whether we want an arm64 smoke under QEMU as well.
> 
> One smaller question: do we want `pip_index_url` to be arbitrary for the official image publishing workflow, or should this be restricted/confirmed somehow so the image always comes from the expected package index?

Thanks for the review — I've made the corresponding adjustments for the points you raised.

On the smoke-test part: since this is a manually-triggered workflow, I think it's fine to push first and then decide based on the Action's result whether to clean up the image if the smoke test failed.

Icedcoco:
Thanks for addressing the earlier round! I think we're getting really close. Just a couple of lingering thoughts before we merge this in.

My main concern right now is still around the smoke test scope (`publish-master-image.yaml:163-164` and `:171-172`). Since it's currently import-only, it proves the extensions can load, but doesn't quite guarantee the image is actually usable for the entrypoints this PR advertises (like `mooncake_master` and the HTTP metadata server). Since we already have packaging/CLI coverage for these paths in the repo, I'd really love to see the workflow exercise at least one real entrypoint. Smoking something like `mooncake_master --help` or a minimal startup path instead of stopping at module import would give us a lot more confidence.

On a related note, I wanted to double-check the push-before-smoke behavior (`publish-master-image.yaml:148` / `:158-174`). I saw the follow-up comment mentioning this is acceptable because the workflow is manual-only, and I totally get that as an operational tradeoff. I'm still a little uneasy about the failure mode for an official public tag—just because a bad image would become visible before the smoke finishes, and we'd still have to clean it up manually. If you feel that tradeoff makes sense for now, I'm okay with it! Otherwise, maybe a pre-push smoke or an automatic cleanup step later might be worth a look.

One last small thing on the PyPI preflight—is the intent for this image to be explicitly the Python 3.12 flavor of Mooncake? I noticed it hard-requires `cp312` wheels for both amd64 and arm64. If that's by design, it might be worth adding a quick comment in the workflow or Dockerfile just so the next person knows it’s intentional rather than incidental.

tpiperatgod:
Your three suggestions all make sense, and I've made the corresponding adjustments. PTAL when u have time.

Icedcoco:
LGTM.
CI Gate is still red, but the failure is in `build-musa`, and the log looks like an existing MUSA/CUDA build-path issue rather than something introduced by this PR. I don’t see any remaining blocking issues from the diff, so I’m fine with merging this.

ykwd (approved):

PR #2679, opened 2026-06-30
[Bugfix] Remove duplicate replica erase in PutRevoke

fcczzz:
## Description

Remove a duplicate `EraseReplicasWithCacheTotalAccounting` call in `PutRevoke`.

The first call already removes the matching replicas and performs cache-total accounting, local disk usage cleanup, and quota adjustment bookkeeping. The second call uses the same predicate immediately afterward, so it cannot remove additional replicas and only repeats the accounting sync path unnecessarily.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

### DIFF
--- a/mooncake-store/src/master_service.cpp
+++ b/mooncake-store/src/master_service.cpp
@@ -3187,14 +3187,6 @@ auto MasterService::PutRevoke(const UUID& client_id, const std::string& key,
         metadata.reserved_quota_charge_bytes = 0;
     }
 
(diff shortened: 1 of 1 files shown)


LujhCoconut (approved):
Good catch.

Issue #2233, opened 2026-05-26
[RFC]:  L2→L1 promotion background retry for hot LOCAL_DISK-only keys V1.1


Srinivasoo7:
## Motivation

V1
 L2→L1 promotion-on-hit landed in #2071, based on the design discussion in #2031 and the earlier Milestone 13 discussion in #2025. V1 added the core async path: when a `Get` observes a LOCAL_DISK-only key that passes admission gates, the master queues a promotion task and the holder client promotes the object back into MEMORY through the heartbeat path.

However, V1 admission is still tied to Get-time evaluation. If promotion is skipped because DRAM is above watermark, the promotion queue is full, allocation fails, or heartbeat throughput is temporarily constrained, the key only gets another promotion chance on a future Get.

For hot LOCAL_DISK-only keys, this can keep repeated reads on the SSD path longer than necessary even after the original blocking condition clears.

V1.1 proposes a bounded background retry/admission layer on top of the V1 primitives. The Get path remains lightweight and non-blocking. It only records candidate information. A background scheduler later re-evaluates hot LOCAL_DISK-only candidates and enqueues promotion when memory and queue pressure allow.

## Goals

- Preserve the V1 contract:
  - best-effort promotion
  - no blocking on `Get`
  - no source LOCAL_DISK deletion
  - default-off behavior remains unchanged
- Reuse the V1 promotion queue, heartbeat pull, allocation, write, and `NotifyPromotionSuccess` path.
- Retry promotion for hot LOCAL_DISK-only keys previously skipped due to transient conditions.
- Keep all new state bounded by count and TTL.
- Avoid overlap with #2176, which focuses on promotion metrics and `promotion_max_per_heartbeat`.

## Non-Goals

- No synchronous promotion on the Get path.
- No proactive promotion of cold keys.
- No source LOCAL_DISK deletion after successful promotion.
- No replacement of V1 CountMinSketch admission.
- No new client-facing API.
- No changes to #2176 metrics scope except optionally consuming those metrics for validation.

## Proposed Design

Add a master-side bounded candidate set for LOCAL_DISK-only keys that are hot enough to promote but could not be admitted immediately.

When V1 promotion admission rejects a key for a transient reason, record/update a candidate entry:

- key
- approximate access count or sketch score
- last_seen timestamp
- last_reject_reason
- retry_after timestamp
- retry_count

A background scheduler periodically scans a bounded number of candidates and re-checks:

- key still exists
[...]

PR #2680, opened 2026-06-30
[Store] Retry transient promotion candidates

ljb90071-ship-it:
# [Store] Retry transient promotion candidates

Addresses #2233.

## Summary

This change adds a master-side bounded retry candidate set for promotion-on-hit. When a hot LOCAL_DISK-only key passes the frequency gate but is rejected by a transient admission gate, the master records it as a candidate and a background scheduler re-evaluates it later through the existing promotion queue path.

It does not change client-facing RPCs or heartbeat protocol.

## Design

- Add `PromotionCandidate` state under `TenantState`.
- Convert `TryPushPromotionQueue` from `void` to an internal `PromotionQueueResult` so the background retry loop can classify outcomes.
- Record candidates for transient admission failures:
  - DRAM watermark gate
  - global `promotion_queue_limit` cap
  - recoverable promotion queue push failure
- Stop retry for permanent or already-resolved states:
  - object removed
  - MEMORY replica already present
  - no LOCAL_DISK source
  - task already in flight
- Bound retry by candidate count, TTL, max attempts, and exponential backoff.
- Enforce the candidate limit with an atomic global reservation and saturating decrement so reset/cleanup paths cannot underflow the advisory counter.
- Scan retry candidates in bounded metadata-shard batches instead of sweeping all 1024 shards every retry tick.
- Re-enter promotion admission under the snapshot shared lock after releasing shard locks, matching foreground metadata access without shard-lock recursion.
- Clear transient promotion retry state on metadata snapshot reload/reset, including the candidate counter and in-flight soft cap.
- Reuse the existing enqueue path so refcount pinning, holder queue insertion, and `PromotionTask` bookkeeping stay centralized.

## Tests

```bash
cmake --build build-ccf-ninja --target promotion_on_hit_test promotion_retry_benchmark_test -j2
./build-ccf-ninja/mooncake-store/tests/promotion_on_hit_test --gtest_brief=1
./build-ccf-ninja/mooncake-store/tests/promotion_retry_benchmark_test --gtest_brief=1
```

Results:

```text
promotion_on_hit_test: 43/43 PASS
promotion_retry_benchmark_test: PASS
```

Benchmark output:

```json
{"retry_visible":true,"fallback_reads_needed":0,"retry_poll_ms":171,"queue_limit":1,"workload":"two_hot_local_disk_only_keys"}
```

`retry_poll_ms` is the observed local public-heartbeat visibility time for this run; it can vary slightly with shard placement and scheduler timing. The semantic regression target is `fallback_reads_needed=0` after the in-flight slot is released.

### DIFF
--- a/mooncake-store/include/master_service.h
+++ b/mooncake-store/include/master_service.h
@@ -1203,6 +1203,34 @@ class MasterService {
         UUID holder_id;  // owner of source LOCAL_DISK; only Notifier allowed
     };
 
+    enum class PromotionQueueResult {
+        kQueued,
+        kDisabled,
+        kFrequencyRejected,
+        kWatermarkRejected,
+        kNotFound,
+        kAlreadyInFlight,
+        kMemoryReplicaPresent,
+        kQueueCapRejected,
+        kNoLocalDiskSource,
+        kPushFailed,
+    };
+
+    enum class PromotionCandidateReason {
+        kWatermark,
+        kQueueCap,
+        kPushFailed,
+    };
+
+    struct PromotionCandidate {
+        std::chrono::system_clock::time_point first_seen;
+        std::chrono::system_clock::time_point next_retry_time;
+        uint32_t retry_attempts{0};
+        PromotionCandidateReason last_reason{
+            PromotionCandidateReason::kQueueCap};
+        ErrorCode last_error{ErrorCode::OK};
+    };
+
     static constexpr size_t kNumShards = 1024;  // Number of metadata shards
 
     struct TenantState {
@@ -1212,6 +1240,8 @@ class MasterService {
             replication_tasks;
         std::unordered_map<std::string, const OffloadingTask> offloading_tasks;
         std::unordered_map<std::string, PromotionTask> promotion_tasks;
+        std::unordered_map<std::string, PromotionCandidate>
+            promotion_candidates;
 
         std::unordered_map<std::string, std::unordered_set<std::string>>
             group_members;  // group_id → set of keys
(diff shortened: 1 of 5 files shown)


LujhCoconut:
Thanks for this PR. 

However, I feel the benefit may be quite marginal and hard to justify against the added complexity. 

Specifically, the current promotion path is already **Get-driven** and **hotness(threshold)-based**. Hot keys are by definition keys that `get` accessed repeatedly — and each `Get` already triggers `TryPushPromotionQueue`, which essentially serves as a natural retry. 
A key that hasn't reached the frequency threshold is still warm data and hasn't been classified as hot yet, so not promoting it is reasonable. A key that is hot but gets rejected due to transient conditions (watermark, capacity) will simply get another chance on the next `Get`. 

From my reading, the incremental gain here is small, while it introduces additional locking and metadata overhead. 

Happy to discuss further.

LujhCoconut:
I also noticed PR #2690 is tackling the same problem. 

I compared the two and they're fairly similar in approach. PR #2690 has somewhat better engineering — steady_clock, standardized metrics through MasterMetricManager, sliding TTL — but  it's missing the `promotion_in_flight_` cleanup on metadata reload that you handled here. 

Your PR came first, but PR #2690's author @Srinivasoo7 was involved in the RFC #2233 and the related discussions from the start. Based on these two factors, I'm  leaning toward reviewing PR #2690 as the primary one to land.

Sorry for the awkward situation — I know it's never great to hear this after putting in the work. Happy to keep discussing if anything above doesn't sit right. Other performance improvements or follow-ups are more than welcome as well.

PR #2681, opened 2026-06-30
[Store] Refactor SHM UDS FD Passing for Real/Dummy Client

Aionw:
## Description

Refactor the Mooncake Store RealClient/DummyClient shared-memory fd-passing path to use a dedicated one-shot Unix domain socket helper. The new `UdsConnector`, `UdsAcceptor`, and `UdsConnection` centralize abstract-namespace UDS setup, raw payload exchange, and `SCM_RIGHTS` fd passing for the performance-insensitive SHM control path.

This keeps the existing wire protocol unchanged while removing the direct UDS socket handling from `real_client.cpp`, `dummy_client.cpp`, and the old `shm_helper` fd-passing wrappers.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

Added focused UDS transport tests for raw payload exchange, client-to-server fd passing, server-to-client fd passing, accept-loop stop wakeup, and missing-handler start errors.

**Test commands:**
```bash
c++ -std=c++20 -I mooncake-store/include mooncake-store/src/uds_transport.cpp mooncake-store/tests/uds_transport_test.cpp -lgtest -lgtest_main -lglog -pthread -o /tmp/uds_transport_test
/tmp/uds_transport_test
```

**Test results:**
- [x] Unit tests pass

Manual validation:
- Manually compiled `/tmp/uds_transport_test` from the affected source and test files.
- Ran `/tmp/uds_transport_test` unsandboxed; all 5 `UdsTransportTest` cases passed.
- UT

## Checklist

Change size is over 500 LOC mostly due to the new focused unit test and transport helper. No RFC issue was filed because this is a narrow internal refactor that keeps the existing wire protocol and behavior unchanged.

## AI Assistance Disclosure

- [ ] No AI tools were used
AI generated the implementation. Human review covered 100% of the diff.


mooncake-store/src/uds_transport.cpp
```diff
@@ -0,0 +1,328 @@
...
+    msg.msg_iovlen = 1;
+    msg.msg_control = buf;
+    msg.msg_controllen = sizeof(buf);
+
+    while (true) {
+        ssize_t received = recvmsg(fd_, &msg, 0);
+        if (received < 0 && errno == EINTR) continue;
+        if (received <= 0) return -1;
+        break;
+    }
```
Lin-z-w, line 217: `recvFd()` should validate that the payload was received completely before returning the fd. Today any positive `recvmsg()` result with `SCM_RIGHTS` is treated as success, even if `received < data_len` or the message/control data was truncated. That means callers can get a valid fd while the associated metadata struct is only partially initialized.

Please treat `received != data_len`, `MSG_TRUNC`, and `MSG_CTRUNC` as errors. The same completeness check should also be applied on the `sendFd()` side for `sendmsg()`, i.e. `sent` must equal `data_len`.

mooncake-store/src/uds_transport.cpp
```diff
@@ -0,0 +1,328 @@
...
+
+void UdsAcceptor::stop() {
+    if (!running_.exchange(false)) return;
+    wakeAccept();
+    if (thread_.joinable()) thread_.join();
+    if (listen_fd_ >= 0) {
+        ::close(listen_fd_);
+        listen_fd_ = -1;
+    }
+}
```
Lin-z-w, line 310: `stop()` currently only wakes a thread blocked in `accept()` by shutting down the listening fd. If the accept loop has already accepted a client and is blocked inside `handler_` on a read/write or another blocking operation, `stop()` can still block indefinitely in `join()`.

The current `RealClient` handler sets a receive timeout, so the risk is limited for this specific call path, but `UdsAcceptor` itself is a reusable helper and does not enforce that contract. Please either document that handlers must be bounded/non-blocking, or make `UdsAcceptor` track the active client connection and `shutdown()` it from `stop()` as well.

Aionw: Fixed in [7fb0d51](https://github.com/kvcache-ai/Mooncake/pull/2681/commits/7fb0d5198f9775b8976aa80a091372157051a258)

mooncake-store/src/uds_transport.cpp
Aionw: Fixed in [7fb0d51](https://github.com/kvcache-ai/Mooncake/pull/2681/commits/7fb0d5198f9775b8976aa80a091372157051a258)

Lin-z-w (approved):
LGTM.

PR #2682, opened 2026-06-30
[TE] Support rdma+hip multi-protocol segments for single-node disaggregation

Lzy17:
Addresses #2684.

## Summary
This change lets a single node advertise and use **both** `rdma` and `hip` on one local segment, which is the combination AMD prefill/decode (PD) disaggregation needs: the device KV pool moves over the intra-node HIP transport (XGMI) while the host-resident metadata/aux buffers move over RDMA. Previously the transfer engine registered every buffer into all installed transports but published them under a single protocol, so an `rdma+hip` node collapsed to `hip`, the RDMA-registered KV buffers (no `shm_name`) tripped `Corrupted segment descriptor`, and KV transfer died.

It does not change client-facing RPCs, the heartbeat protocol, or the existing single-protocol fast path. Scope is intentionally bounded to `rdma+hip`, mirroring the existing `cxl+tcp` / `cxl+rdma` multi-protocol pattern.

## Design
- `HipTransport::registerLocalMemory`: skip non-device memory gracefully (`return 0`) instead of returning `-1`, so host buffers handed to every transport fall through to RDMA/TCP rather than rolling back the whole batch and tearing down sessions. Clear the sticky HIP error left by the failed pointer query so it cannot poison the caller's subsequent HIP calls.
- `HipTransport::install`: compose with any local segment another transport already published instead of overwriting it, so the node advertises `"rdma,hip"`. Mirrors `RdmaTransport::allocateLocalSegmentID`.
- Tag each HIP-registered `BufferDesc.protocol = "hip"` (RDMA already tags `"rdma"`) so a mixed segment round-trips with per-buffer ownership.
- Generalize the multi-protocol encode/decode allow-list (`encodeSegmentDesc` / `decodeSegmentDesc`) to accept any combination of protocols that have a per-buffer emit/parse branch (`cxl`/`tcp`/`rdma`/`hip`); remove the hardcoded 2-protocol cap and the CXL-must-be-present requirement.
- Add a `hip` per-buffer emit/parse branch (`addr` + `shm_name`).
- `MultiTransport::selectTransport`: for a comma-protocol segment, route each request to the transport owning the buffer that covers the target address; prefer `hip` when an address is registered under multiple protocols (device KV over XGMI). Single-protocol segments keep the existing path unchanged.
- Single-protocol `hip` decode: skip buffers with an empty `shm_name` defensively instead of rejecting the whole segment.

## Tests
Single-node 1P1D (prefill GPU0 / decode GPU1, Qwen2.5-0.5B) on MI355X over the real `rdma+hip` path. Topology discovery finds the bnxt HCA, then installs both transports:
```
Topology discovery complete. Found 1 HCAs.
installTransport, type=rdma
HIP transport installed for intra-node GPU P2P
```
A/B with the patch stashed vs. applied, same servers, same request:
```
curl http://0.0.0.0:8000/v1/chat/completions \
  -d '{"model":"qwen","messages":[{"role":"user","content":"What is the capital of France? Answer in one word."}],"max_tokens":16,"temperature":0}'
```
Results:
- **Without this change:** `hip_transport.cpp Unsupported memory type`, `transfer_metadata.cpp Corrupted segment descriptor ... protocol hip ... shm_name `(empty), then `Failed to get kvcache from prefill instance`.
- **With this change:** completion returns `Paris`; logs show 0 Corrupted-segment, 0 Unsupported-memory, 0 session-not-alive, 0 address-not-found.

### DIFF
--- a/mooncake-transfer-engine/src/multi_transport.cpp
+++ b/mooncake-transfer-engine/src/multi_transport.cpp
@@ -464,7 +464,7 @@ Status MultiTransport::selectTransport(const TransferRequest& entry,
     // by both rdma and hip), pick the highest-performance transport by a fixed
     // priority instead of relying on buffer registration order.
     if (proto.find(',') != std::string::npos) {
-        auto protocol_priority = [](const std::string &p) {
+        auto protocol_priority = [](const std::string& p) {
             if (p == "hip") return 4;
             if (p == "cxl") return 3;
             if (p == "rdma") return 2;
@@ -473,12 +473,13 @@ Status MultiTransport::selectTransport(const TransferRequest& entry,
         };
         std::string chosen;
         int chosen_priority = -1;
-        for (const auto &buffer : target_segment_desc->buffers) {
+        for (const auto& buffer : target_segment_desc->buffers) {
             // CXL buffers locate via offset + cxl_base_addr; all other
             // protocols use the absolute virtual address in buffer.addr.
-            uint64_t start = (buffer.protocol == "cxl")
-                                 ? buffer.offset + target_segment_desc->cxl_base_addr
-                                 : buffer.addr;
+            uint64_t start =
+                (buffer.protocol == "cxl")
+                    ? buffer.offset + target_segment_desc->cxl_base_addr
+                    : buffer.addr;
             if (entry.target_offset >= start &&
                 entry.target_offset < start + buffer.length) {
                 int priority = protocol_priority(buffer.protocol);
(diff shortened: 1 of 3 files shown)


alogfans (approved):
LGTM

alogfans:
Please fix format issues and re-run CIs before merging.

Lzy17:
> Please fix format issues and re-run CIs before merging.

done, thanks!!

PR #2685, opened 2026-06-30
[Store] Make batch_query_keys read-only and return all replica types

smartssw:
## Description

The `/batch_query_keys` HTTP admin endpoint had two problems:

1. **Side effects from reusing a client-facing API.** `HandleBatchQueryKeys` called `BatchGetReplicaList`, which is designed for client reads and performs `GrantLeaseForGroup()`, `promotion_on_hit_`, and cache-hit metric updates. Admin queries therefore prevented keys from being evicted, spuriously promoted disk-only keys to memory, and polluted cache-hit metrics. By contrast `/get_all_keys` uses the side-effect-free `GetAllKeysForAdmin()` path — the two endpoints were inconsistent.

2. **Missing replica info for disk-based keys.** The HTTP handler skipped non-memory replicas (`if (!replica.is_memory_replica()) continue;`), so disk / local_disk / nof keys returned an empty `values` vector and the caller could not inspect their replica locations, even though the metadata was available in the master.

This change adds `BatchGetReplicaListForAdmin` and `GetReplicaListForAdmin` to `MasterService` and `WrappedMasterService`: pure metadata reads that skip lease grants, promotion, and cache-hit counters. `HandleBatchQueryKeys` now uses the admin variant and additionally serializes `disk_values`, `local_disk_values`, and `nof_values` alongside the existing `values` field. `values` remains always-present (empty array for disk-only keys) so existing clients that check for the field or treat it as an array are unaffected; the new fields are optional and only appear when the corresponding replica type is present.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
# Code format check

# Compile the changed files
cd build
cmake --build . --target mooncake_master -j1

# End-to-end manual test against a running mooncake_master
# (write a local_disk key via the store client first)
curl -s "http://<master_host>:<http_port>/batch_query_keys?keys=<local_disk_key>"
```

**Test results:**
- [ ] Unit tests pass

**Code format:** `./scripts/code_format.sh --check` reports all 5 changed files properly formatted.

**Compile verification:** `master_admin_service.cpp` and `rpc_service.cpp` compiled individually with the project's exact CMake flags — no errors from the changed code (only pre-existing third-party warnings from async_simple / yalantinglibs).

**End-to-end manual test:** Wrote a `LOCAL_DISK` replica to the store, then queried it via `/batch_query_keys`:

```bash
$ curl -s "http://127.0.0.1:9003/batch_query_keys?keys=b909bf28318272fd6ce0dc7ceae4c1ee6477878cd662defd8363ff49b5c24bd0_0_k"
```

Response:
```json
{
  "success": true,
  "data": {
    "b909bf28318272fd6ce0dc7ceae4c1ee6477878cd662defd8363ff49b5c24bd0_0_k": {
      "ok": true,
      "error": null,
      "values": [],
      "disk_values": null,
      "local_disk_values": [
        {
          "client_id": "9531234657444698320-2435329600209450907",
          "object_size": 32768,
          "transport_endpoint": "127.0.0.1:44641"
        }
      ],
      "nof_values": null
    }
  }
}
```

The response confirms the expected behavior:
- `values` is `[]` (no memory replica) — unchanged from prior behavior, preserving backward compatibility
- `local_disk_values` is populated with the replica's `client_id`, `object_size`, and `transport_endpoint`
- `disk_values` and `nof_values` are `null` (no replicas of those types)

## Checklist

## AI Assistance Disclosure

- [x] No AI tools were used

### DIFF
--- a/docs/source/http-api-reference/http-service.md
+++ b/docs/source/http-api-reference/http-service.md
@@ -91,7 +91,7 @@ curl "http://localhost:8080/query_key?key=my_object"
 ```
 
 #### `/batch_query_keys`
-Retrieve replica information for multiple keys in a single request, including memory locations and transport endpoints for each key.
+Retrieve replica information for multiple keys in a single request, including memory locations and transport endpoints for each key. The endpoint performs a read-only metadata lookup and does not grant leases, trigger promotion, or update cache-hit metrics.
 
 **Method**: `GET`
 **Parameters**: `keys` (query parameter) - Comma-separated list of object keys to query (format: key1,key2,key3)
@@ -115,6 +115,25 @@ curl "http://localhost:8080/batch_query_keys?keys=key1,key2,key3"
           "transport_endpoint_": "hostname:port",
           "buffer_descriptor": {...}
         }
+      ],
+      "disk_values": [
+        {
+          "file_path": "/path/to/object",
+          "object_size": 4096
+        }
+      ],
+      "local_disk_values": [
+        {
+          "client_id": "12345-67890",
+          "object_size": 4096,
+          "transport_endpoint": "hostname:port"
+        }
+      ],
+      "nof_values": [
+        {
+          "transport_endpoint_": "hostname:port",
+          "buffer_descriptor": {...}
+        }
       ]
     },
     "key2": {
@@ -124,6 +143,8 @@ curl "http://localhost:8080/batch_query_keys?keys=key1,key2,key3"
   }
 }
 ```
+
+The `values` field is always present (empty array when no memory replica exists). The `disk_values`, `local_disk_values`, and `nof_values` fields are optional and only appear when the corresponding replica type is present for the key.
 
 #### `/get_all_keys`
 List all keys currently stored in the distributed system.
--- a/mooncake-store/include/master_service.h
+++ b/mooncake-store/include/master_service.h
@@ -318,11 +318,29 @@ class MasterService {
         -> tl::expected<GetReplicaListResponse, ErrorCode>;
 
     /**
+     * @brief Read-only single-key replica list query for admin use.
+     * Unlike GetReplicaList, this does not grant leases, trigger
+     * promotion, or update cache-hit metrics.
+     */
+    auto GetReplicaListForAdmin(const std::string& key,
+                                const std::string& tenant_id)
+        -> tl::expected<GetReplicaListResponse, ErrorCode>;
+
+    /**
      * @brief Get replica lists for a batch of objects.
      */
     std::vector<tl::expected<GetReplicaListResponse, ErrorCode>>
     BatchGetReplicaList(const std::vector<std::string>& keys,
                         const std::string& tenant_id);
+
+    /**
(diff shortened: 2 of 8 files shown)


ykwd:
Where is this interface currently being used? In particular, is it being used by external systems such as SGLang or vLLM?

ykwd:
@stmatengss  I guess you are familiar with this API. Could you take a look?

smartssw:
> Where is this interface currently being used? In particular, is it being used by external systems such as SGLang or vLLM?

batch_query_keys is an internal admin/diagnostic endpoint. Within the Mooncake repo, its only current HTTP client is the C++ unit tests; it is not yet consumed by SGLang, vLLM, or the Python integration layer.

We have two motivations for this change: 
1. Better debugging — the endpoint now returns replica metadata for disk / local_disk / nof keys, which the previous implementation silently dropped, making it hard to inspect where objects actually live when the disk tier is in use.
2.  Cache-hit-aware scheduling via sglang-router — we plan to have sglang-router pull query results from Mooncake to measure request hit rate and drive cache-hit-aware scheduling. When Mooncake uses the disk tier, the current endpoint cannot report an accurate hit rate because it only counted memory replicas. Making batch_query_keys side-effect-free (no lease grants, no promotion, no cache-hit metric pollution) and returning all replica types are prerequisites for this: the router can then distinguish memory hits from disk/local_disk hits and compute the true cache hit rate for scheduling decisions. 

The change is backward compatible and does not affect existing clients: the request URL and parameters (?keys=key1,key2,...) are unchanged; the existing values field keeps the same shape and remains always-present (an empty array for disk-only keys, exactly as before); the new disk_values / local_disk_values / nof_values fields are optional and only appear when the corresponding replica type is present, so they appear as additions rather than changes. Existing callers that only read values will continue to work without modification.

stmatengss:
Could you add some tests to verify the correctness? Others LGTM.

smartssw:
> Could you add some tests to verify the correctness? Others LGTM.

Tests added in 272478b6, all passing locally:

BatchGetReplicaListForAdminDoesNotPromoteLocalDiskOnlyObject — admin query on a LOCAL_DISK-only key returns the replica but triggers no promotion (vs. the client path which does).
BatchGetReplicaListForAdminDoesNotUpdateCacheHitMetrics — admin query leaves the cache-hit counter unchanged (vs. the client path which bumps it).
BatchQueryKeysReturnsLocalDiskReplicaInfo — HTTP endpoint returns LOCAL_DISK replica info in local_disk_values, with values kept as an empty array for backward compat.
Full suites pass with no regressions (promotion_on_hit_test: 40, master_admin_server_test: 53).

stmatengss (approved):
LGTM

PR #2689, opened 2026-06-30
[Store] Add hidden state type-aware eviction

kanceler:
## Description

This PR adds Hidden State type-aware cache eviction for Mooncake Store. It targets multimodal EPD (Encoder-Prefill-Decode) workloads where Hidden State tensors and KV Cache blocks have different access patterns and recomputation costs.

The implementation is disabled by default and keeps existing KV-only behavior unchanged unless `enable_hidden_type_aware_eviction` is enabled.

Design rationale:

- Hidden State and KV Cache have different reuse granularity. KV Cache can be
  partially reused at prefix/block granularity, while a multimodal Hidden State
  object is reused as a whole object.
- Hidden State and KV Cache have different space/value characteristics.
  Hidden State objects are typically MB-level objects, but their recomputation
  value can vary greatly across images and videos. KV Cache objects are larger
  in aggregate and have more stable reuse value.
- Hidden State and KV Cache have different reuse frequency. KV Cache can be
  refreshed repeatedly during multi-turn conversations, while Hidden State is
  reused less frequently across sessions, recovery, or scheduling events.
- Hidden State and KV Cache evolve differently over time. KV hotness is
  continuously refreshed by conversation turns, while Hidden State may stay
  silent for a long time after a useful access. A unified LRU/lease policy can
  therefore misclassify valuable Hidden State as cold data.

Main changes:

- Add `ObjectDataType::HIDDEN_STATE` and expose it in the Python binding.
- Track Hidden State memory usage via `master_hidden_allocated_bytes`.
- Add a logical Hidden memory budget before global eviction.
- Add Hidden-specific lease TTL and soft-pinned Hidden lease TTL.
- Add unit tests for data type propagation, accounting, budget eviction, fallback behavior, and soft-pin lease behavior.
- Add benchmark report, scripts, and reproducible result archive for review.

The implementation follows a minimal-intrusion design. It reuses the existing
Store lifecycle, replica management, eviction, and cleanup paths, and adds only
type-aware control at the policy layer:

- a logical Hidden budget instead of a physical memory partition;
- separate lease TTL selection for Hidden State and KV Cache;
- lightweight soft pin semantics that extend the lease for high-value Hidden
  objects but still allow normal eviction after expiration;
- fallback to native global eviction behavior when the Hidden budget is not
  active or is configured as disabled.

Benchmark three-run averages show that under the mixed multimodal workload, KV hit rate remains nearly unchanged while Hidden hit rate improves from 26.62% to 46.64%. Video Hidden hit rate improves from 44.44% to 77.78% with soft pin enabled.

This PR is part of a broader submission for the Mooncake KVCache Storage Design and Performance Optimization Contest, Track 4: KVCache separation and cooperative scheduling for Agent and multimodal inference. It focuses specifically on the Mooncake Store eviction-policy component for Hidden State and KV Cache coexistence, rather than the full end-to-end contest deliverable.

Related materials:

- Benchmark report: `benchmarks/eviction_policy_benchmark/EVALUATION.md`
- Benchmark scripts: `benchmarks/eviction_policy_benchmark/`

The complete raw benchmark outputs and supporting experimental materials will be provided as part of the contest report submission.

## Module

- [ ]  Transfer Engine (`mooncake-transfer-engine`)
- [X]  Mooncake Store (`mooncake-store`)
- [ ]  Mooncake EP (`mooncake-ep`)
- [ ]  Mooncake PG (`mooncake-pg`)
- [X]  Integration (`mooncake-integration`)
- [ ]  P2P Store (`mooncake-p2p-store`)
- [ ]  Python Wheel (`mooncake-wheel`)
- [ ]  Common (`mooncake-common`)
- [ ]  Mooncake RL (`mooncake-rl`)
- [ ]  CI/CD
- [X]  Docs
- [X]  Other

Other: benchmark scripts, benchmark report, and result archive.

## Type of Change

- [ ]  Bug fix
- [X]  New feature
- [ ]  Refactor
- [ ]  Breaking change
- [X]  Documentation update
- [X]  Performance improvement
- [ ]  Other

## How Has This Been Tested?

Unit tests:

- `object_data_type_test`: 8/8 passed.
- `master_service_test`: 142/142 passed.

Verified in an official clean upstream clone with only this PR's Mooncake changes applied.

Commands:

```bash
ninja -C build-docker mooncake-store/tests/object_data_type_test mooncake-store/tests/master_service_test
./build-docker/mooncake-store/tests/master_service_test
```

Benchmark summary:

| Case                                                                   | KV hit rate                  | Hidden hit rate       | Saved compute ms (estimated) |
| ---------------------------------------------------------------------- | ---------------------------- | --------------------- | ---------------------------- |
| Baseline best, unified TTL=10s                                         | 83,592.67 / 113,747 = 73.49% | 78.00 / 293 = 26.62%  | 152,198.10                   |
| Policy-Budget, budget=0.08, Hidden/KV TTL=10s                          | 83,412.00 / 113,747 = 73.33% | 138.00 / 293 = 47.10% | 168,635.00                   |
| Policy-Lease, budget=0.08, Hidden TTL=1s, KV TTL=10s                   | 83,404.67 / 113,747 = 73.32% | 138.00 / 293 = 47.10% | 168,624.52                   |
| Policy-Full, budget=0.08, Hidden TTL=10s, Soft Pin TTL=20s, KV TTL=10s | 83,724.33 / 113,747 = 73.61% | 136.67 / 293 = 46.64% | 180,909.52                   |
| Policy-Fallback, budget=1.0, Hidden/KV TTL=10s                         | 83,435.33 / 113,747 = 73.35% | 77.33 / 293 = 26.39%  | 149,991.67                   |

Benchmark parameters:

- Workload fingerprint: `ab69f2dc33841c64`
- `scale=0.0025`
- Store capacity: 64 MiB
- `sessions=240`
- `mixed_write_sessions=260`
- Proxy mode: `session`
- Execution mode: `request_eviction`

Known limitations:

- Hidden recognition depends on writers setting `ReplicateConfig.data_type = HIDDEN_STATE`.
- Hidden budget eviction currently targets object count rather than exact bytes.
- TTL has limited effect in sustained high-pressure capacity-driven eviction.
- Soft pin is weak protection, not permanent eviction exemption.
- `Saved compute ms` is a model-based estimate for relative comparison, not measured end-to-end latency.
- Multi-replica, NoF, offload, and snapshot-restore combinations still need broader follow-up coverage.

## Checklist

- [X]  I have performed a self-review of my own code
- [X]  I have formatted my code using `./scripts/code_format.sh`
- [ ]  I have run `pre-commit run --all-files` and all hooks pass
- [X]  I have updated the documentation (if applicable)
- [X]  I have added tests to prove my changes are effective
- [ ]  For changes >500 LOC: I have filed an RFC issue

## AI Assistance Disclosure

- [ ]  No AI tools were used
- [X]  AI tools were used (specify below)


stmatengss:
That's an awesome idea. We welcome you to merge this PR. BTW, please cleanup this PR (there are so many immediate files)

kanceler:
> That's an awesome idea. We welcome you to merge this PR. BTW, please cleanup this PR (there are so many immediate files)

Thanks. I have cleaned up the unnecessary files, including the raw benchmark result data.

ykwd:
Thanks for this PR. Could you introduce what eviction algorithm or strategy is added in this PR?

kanceler:
> Thanks for this PR. Could you introduce what eviction algorithm or strategy is added in this PR?

This PR adds a Hidden State type-aware policy layer to the MasterService in-memory eviction flow. It does not replace the existing global eviction path.

The main strategy includes:

1. Hidden State gets a logical memory budget. The eviction trigger is still the existing global memory high watermark. When global memory eviction is triggered, the system first checks whether Hidden State usage exceeds the configured logical budget. If it does, it runs a Hidden-only eviction pass first.

2. After the Hidden-only pass, the system continues the original global eviction flow. By default, Hidden State objects are excluded from the later global eviction candidates, so the same type of objects are not handled twice. The objects evicted and the space released in the Hidden-only pass are still counted into the progress of the current `BatchEvict` round.

3. Hidden State and KV Cache can use different lease TTLs. KV keeps the existing KV lease behavior, while Hidden State can use a separate default lease TTL.

4. For high-value Hidden State objects, such as video Hidden State, this PR reuses the existing soft pin mechanism. It does not make soft pin a permanent protection. Instead, when a Hidden State object is soft-pinned, it selects a longer Hidden lease TTL; after the lease expires, the object can still be evicted normally.

The goal of this design is to keep the Store layer model-agnostic and inference-framework-agnostic. Store does not need to understand model-specific recomputation cost or reuse prediction. It only uses object type, logical budget, lease, and soft pin metadata for general cache management. The upper inference framework can use simple rules to decide whether to apply soft pin, for example only soft-pinning video Hidden State. More complex value estimation or reuse prediction can evolve later as an upper-layer policy instead of being coupled into the Store core eviction logic.

The current implementation intentionally keeps the change minimally invasive to the existing eviction path. It is mainly used to validate the necessity and effectiveness of differentiated Hidden/KV cache management. If the community agrees with this direction, the implementation can be further optimized later, for example by reducing the extra metadata scan or by integrating Hidden-only candidate selection more closely with the existing eviction selection logic.

ykwd:
Let me summarize to make sure I understand correctly. Please let me know if I got this right or if I missed anything:

Hidden stats will have a separate memory budget, as well as separate lease and soft pin durations.

kanceler:
> Let me summarize to make sure I understand correctly. Please let me know if I got this right or if I missed anything:
> 
> Hidden stats will have a separate memory budget, as well as separate lease and soft pin durations.

the Hidden State budget is a separate logical budget,can use a separate lease TTL, and soft-pinned Hidden State can use a longer lease TTL.

ykwd:
From the perspective of the contest, I think this is a great PR. However, if we want to merge it into the Mooncake project, it would be better to split it into several smaller PRs. For example:

* One PR to introduce type-aware objects.
* One PR to support configuring different lease and soft pin durations for different object types.
* One PR to support memory quotas for different object types.

For the latter two, we could design a more general mechanism so that it can be reused for other object types in the future, such as model weights.

For the quota part, we should also consider how it interacts with the current multi-tenant functionality. There may be opportunities to reuse some of the existing internal implementation.

kanceler:
> From the perspective of the contest, I think this is a great PR. However, if we want to merge it into the Mooncake project, it would be better to split it into several smaller PRs. For example:
> 
> * One PR to introduce type-aware objects.
> * One PR to support configuring different lease and soft pin durations for different object types.
> * One PR to support memory quotas for different object types.
> 
> For the latter two, we could design a more general mechanism so that it can be reused for other object types in the future, such as model weights.
> 
> For the quota part, we should also consider how it interacts with the current multi-tenant functionality. There may be opportunities to reuse some of the existing internal implementation.

Thanks for the feedback. I agree that splitting this into smaller, more general PRs would make it easier to review and merge.

I am currently organizing the remaining contest-related submissions, including the vLLM-side EPD prototype. In parallel, I will revisit this PR over the next few days and try to decouple it along the directions you suggested: type-aware objects, per-type lease/soft-pin configuration, and per-type memory quotas.

ykwd:
> Thanks for the feedback. I agree that splitting this into smaller, more general PRs would make it easier to review and merge.
> 
> I am currently organizing the remaining contest-related submissions, including the vLLM-side EPD prototype. In parallel, I will revisit this PR over the next few days and try to decouple it along the directions you suggested: type-aware objects, per-type lease/soft-pin configuration, and per-type memory quotas.

That sounds great. Thanks a lot for taking the time to reorganize it!

PR #2071, opened 2026-05-10
[Store] L2->L1 promotion-on-hit

yzhan1:
## Description

Adds the L2→L1 promotion-on-hit feature: when a key whose only complete replica lives on a peer client's local SSD (LOCAL_DISK) is read enough times to clear an admission threshold, the master enqueues a promotion task and the holder client's next FileStorage heartbeat stages a fresh MEMORY replica via Transfer Engine. Symmetric counterpart to offload-on-evict: offload moves cold data DRAM→SSD on eviction, promotion moves rediscovered-warm data SSD→DRAM on access.

The trigger lives inside `MasterService::GetReplicaList`. A CountMinSketch tracks per-key access frequency; gates filter out cases where promotion would be wasteful or unsafe (no LOCAL_DISK source, MEMORY already present, DRAM under high-watermark pressure, queue saturated, or task already in-flight for the same key). When all gates pass, the master records a `PromotionTask` in the per-shard map (refcnt-pinning the source replica) and pushes the key onto the holder's per-LocalDiskSegment `promotion_objects` map. The holder's FileStorage heartbeat thread drains a bounded batch from that map, calls `PromotionAllocStart` to stage a PROCESSING MEMORY replica, reads the bytes from local SSD, RDMA-WRITEs them via `PromotionWrite`, and notifies the master. `NotifyPromotionSuccess` flips the new replica COMPLETE and decrements the source LOCAL_DISK refcnt. A reaper drops expired tasks; orphaned PROCESSING replicas are reaped via the standard discarded-replicas path.

Failure handling matches offload's posture: per-key failures are logged and skipped, and the master-side reaper handles task TTL expiry on its own schedule.

### Key design points addressed in this PR

- **Concurrent-Put disambiguation in NotifyPromotionSuccess.** The PromotionTask stores the `ReplicaID` of the staged MEMORY replica captured during `PromotionAllocStart`. `NotifyPromotionSuccess` commits exactly that replica via `GetReplicaByID`, so a parallel Put creating its own PROCESSING MEMORY replica on the same key cannot be confused with ours.
- **Heartbeat liveness.** Each promotion task does a synchronous SSD read + RDMA write whose wall time scales with object size (64 MB–1 GB tensors are common in KV-cache workloads). `MasterService::PromotionObjectHeartbeat` returns at most `kMaxPerHeartbeat = 1` task per call and leaves the rest queued; the client processes whatever the master returned and never blocks beyond a single task. Leftovers are preserved by construction — no silent drops.
[...]

PR #2690, opened 2026-06-30
[Store] L2→L1 promotion background retry for hot LOCAL_DISK-only keys (V1.1)

Srinivasoo7:
Implements RFC #2233.

## Summary

V1 promotion-on-hit (#2071) is Get-driven: when a hot `LOCAL_DISK`-only key passes the CountMinSketch frequency gate, the master queues an async promotion task. However, transient blockers — DRAM above watermark, promotion queue at cap, recoverable push failure — cause the key to miss its window. It only gets another chance on the next `Get`, which may not arrive promptly for bursty workloads where access patterns have subsided but the key should still be in DRAM.

V1.1 adds a master-side background scheduler that re-evaluates rejected promotion candidates without waiting for another `Get`.

## Design

- **`Get` path stays non-blocking.** `TryPushPromotionQueue` now returns a typed `PromotionQueueResult` instead of `void`. On a transient rejection (watermark, queue cap, recoverable push failure) it records/updates a `PromotionCandidate` under the key's `TenantState`.
- **`PromotionCandidate`** stores: `sketch_score` (frequency at record time), `first_seen`, `last_seen` (refreshed on each rejection — extends effective TTL while the key stays hot), `retry_after` (exponential backoff), `retry_count`, `last_reason`, `last_error`. Uses `steady_clock` for monotonic interval tracking.
- **Bounded candidate set:** global count `kPromotionCandidateLimit` (50 000), per-candidate `kPromotionCandidateMaxRetries` (8) and `kPromotionCandidateTtl` (60 s from `last_seen`).
- **Background scheduler:** `RunPromotionCandidateRetry` runs in the existing eviction thread every 10 ms (early-exit when count == 0). It advances a ring-buffer cursor over `kPromotionRetryShardBatch` (64) shards per tick, collects up to `kPromotionRetryBatchSize` (128) due candidates, then calls `TryPushPromotionQueue(..., record_candidate=false)` to re-enter the V1 admission path.
- **Permanent resolution** (object gone, MEMORY replica appeared, task already in flight) erases the candidate immediately under the shard lock.
- **Metadata reload** calls `ClearCandidatesForReload` from both `MetadataSerializer::Deserialize` and `Reset` to reset all transient candidate state.

## Metrics

All six counters are wired into `MasterMetricManager` and exposed via Prometheus:

| Prometheus counter | Meaning |
|---|---|
| `master_promotion_candidate_recorded_total` | New candidate entries created |
| `master_promotion_candidate_admitted_total` | Candidates successfully promoted by background scheduler |
| `master_promotion_candidate_admission_rejected_total` | Candidates that hit a gate again on retry |
| `master_promotion_candidate_expired_evaluated_total` | Candidates exhausted after retries or TTL — were evaluated at least once |
| `master_promotion_candidate_expired_unevaluated_total` | Candidates that aged out before the scheduler ever reached them (scan budget too small) |
| `master_promotion_candidate_dropped_limit_total` | Candidates dropped at record time because the global count limit was reached |

The `expired_unevaluated` / `expired_evaluated` distinction is implemented by checking `retry_count == 0` at expiry time in the scan loop: a candidate with `retry_count == 0` was never processed by the scheduler before its TTL elapsed.

## Non-goals (unchanged from V1)

- No synchronous promotion on `Get`
- No proactive scan of cold keys — only `Get`-observed candidates
- No `LOCAL_DISK` deletion after promotion
- No new client-facing RPCs or heartbeat protocol changes

## Test plan

Five new `TEST_F` cases added to `promotion_on_hit_test.cpp`:

- [ ] `RetryCandidate_WatermarkRejectionRecordsCandidate` — watermark rejection records a candidate; a second `Get` deduplicates (count stays 1)
- [ ] `RetryCandidate_ExhaustedAfterMaxRetries` — candidate is erased and `expired_evaluated` is incremented after `kPromotionCandidateMaxRetries` scans
- [ ] `RetryCandidate_ObjectRemovedMidRetry` — object removed while candidate is pending; next scan cleans up with no spurious `admitted` increment
- [ ] `RetryCandidate_MultipleKeysTracked` — multiple LOCAL_DISK-only keys tracked independently, `recorded` counter matches key count
- [ ] `RetryCandidate_ClearOnReload` — `ClearCandidatesForReload` resets all candidate state and the global atomic counter to zero

```bash
cmake --build build-ccf-ninja --target promotion_on_hit_test -j4
./build-ccf-ninja/mooncake-store/tests/promotion_on_hit_test --gtest_filter='*RetryCandidate*' --gtest_brief=1
```


Srinivasoo7:
@stmatengss @LujhCoconut 
PTAL 

Thanks

stmatengss:
> @stmatengss @LujhCoconut PTAL
> 
> Thanks

Sure. Thank you for your contribution!

stmatengss:
Please use './scripts/code_format.sh' to format codes.

mooncake-store/tests/promotion_on_hit_test.cpp
```diff
@@ -2227,6 +2227,198 @@ TEST_F(PromotionOnHitTest, MetricsRemoveMidPromotionCountsAsCancelled) {
...
+
+    // Drive retries; each scan increments retry_count until exhausted.
+    // Run more scans than kPromotionCandidateMaxRetries (8) to be sure.
+    for (int i = 0; i <= 10; ++i) {
+        service->RunPromotionCandidateRetryForTesting();
+    }
+
+    EXPECT_EQ(service->CountCandidatesForTesting("default"), 0u)
+        << "Candidate must be erased after max retries";
+    EXPECT_GT(mm.get_promotion_candidate_expired_evaluated() - expired_pre, 0);
```
LujhCoconut, line 2517: One concern on the tests: `RetryCandidate_ExhaustedAfterMaxRetries` drives retries in a tight loop without any delay between iterations. After the first scan sets `retry_after = T + 10ms`, every subsequent call enters microseconds later and hits `retry_after > now`, always skipping. Reaching `retry_count` 8 requires ~2.5s of cumulative backoff, but 11 tight iterations on near-empty shards take maybe tens of ms — this is likely to fail in CI. 

Suggest adding `sleep_for` between iterations or calling `BackoffCandidate` directly to bypass wall-clock time.

mooncake-store/tests/promotion_on_hit_test.cpp
```diff
@@ -2227,6 +2227,198 @@ TEST_F(PromotionOnHitTest, MetricsRemoveMidPromotionCountsAsCancelled) {
...
+    // Remove the object while candidate is pending.
+    auto rm = service->Remove("k_rm", "default", /*force=*/true);
+    ASSERT_TRUE(rm.has_value());
+
+    // Retry scan should find object gone and erase candidate.
+    service->RunPromotionCandidateRetryForTesting();
+    EXPECT_EQ(service->CountCandidatesForTesting("default"), 0u);
+
+    auto& mm = MasterMetricManager::instance();
+    EXPECT_EQ(mm.get_promotion_candidate_admitted(), 0);
```
LujhCoconut: `MasterMetricManager` is a process-level singleton. If any future test (or a reordered one via `--gtest_shuffle`) increments `admitted`, this assertion breaks. Other tests in the same file correctly use baseline subtraction (`const int64_t pre = mm.get_...()`). 
Not broken today (all current tests have watermark=0 so promotion never succeeds), but fragile.

Fix: wrap with a baseline like the other tests do.

For example:
```cpp
  // Before (line 2345-2346)
  auto& mm = MasterMetricManager::instance();
  EXPECT_EQ(mm.get_promotion_candidate_admitted(), 0);

  // After
  auto& mm = MasterMetricManager::instance();
  const int64_t admitted_pre = mm.get_promotion_candidate_admitted();

  // ... test actions (Remove, RunPromotionCandidateRetryForTesting) ...

  EXPECT_EQ(mm.get_promotion_candidate_admitted() - admitted_pre, 0);
```

Srinivasoo7:
Done @LujhCoconut 
PTAL!
Also included the `promotion_in_flight_` as well

Thanks

Srinivasoo7:
@LujhCoconut Checking in to see if you got a chance to review this. 

Thanks

LujhCoconut:
Please address the two correctness issues flagged by gemini-code-assist. Please use './scripts/code_format.sh' to format codes.

mooncake-store/src/master_service.cpp
```diff
@@ -4999,10 +5000,312 @@ tl::expected<void, ErrorCode> MasterService::PushPromotionQueue(
...
+                                << " retry_count=" << c.retry_count;
+                        cit = tenant_state.promotion_candidates.erase(cit);
+                        DecrementCandidateCount();
+                        // retry_count == 0: scheduler never reached this
+                        // candidate before TTL elapsed — scan budget was
+                        // too small. retry_count > 0: scheduler evaluated
+                        // it but gave up after retries or TTL.
+                        if (c.retry_count == 0) {
+                            MasterMetricManager::instance()
+                                .inc_promotion_candidate_expired_unevaluated();
```
LujhCoconut, line 5474: One issue before merging.

Use-after-free: `c.retry_count` is read on L5238 after `erase(cit)` on L5232 destroyed the map node that c (aliased from `cit->second`) refers to — save `retry_count` to a local before the erase.

Suggested change:
```cpp
  const uint32_t saved_retry_count = c.retry_count;
  cit = tenant_state.promotion_candidates.erase(cit);
  DecrementCandidateCount();
  if (saved_retry_count == 0) {
      MasterMetricManager::instance()
          .inc_promotion_candidate_expired_unevaluated();
  } else {
      MasterMetricManager::instance()
          .inc_promotion_candidate_expired_evaluated();
  }
```

LujhCoconut (approved):
LGTM. Could you also update the relevant parameter configurations in `docs/source/deployment/ssd-offload.md`?

docs/source/deployment/ssd-offload.md
```diff
@@ -26,7 +26,9 @@ mkdir -p /nvme/mooncake_offload
 ```bash
 mooncake_master \
     --rpc_port=50051 \
-    --enable_offload=true
+    --enable_offload=true \
+    --offload_on_evict=true \
```
LujhCoconut: Thanks for your patience. You may consider removing `--enable_offload=true` here, as it is not the default configuration. Additionally, due to rapid code changes, could you please resolve the merge conflicts? We will prioritize reviewing this PR for merging.

Srinivasoo7: Hi @LujhCoconut 
Resolved the merge conflicts and made the above requested change in the recent commit. 

Thanks

PR #2569, opened 2026-06-23
Infer gpu device from request source pointers, set correct device for stream creation.

dadadada-147:
Infer gpu device from request source pointers, set correct device for stream creation.

## Description

In a 1P1D Prefill-Decode (PD) separation scenario with long sequences and high concurrency, we observed transmission timeouts in SGLang when using Mooncake Tent’s MNNVL transport for KV cache transfer. The root cause was identified as follows: when creating CUDA streams, Mooncake failed to switch the device context to the GPU corresponding to each TP rank. Consequently, all Mooncake instances across TP ranks initialized their streams on GPU 0. This caused DMA tasks from all ranks to be queued in GPU 0’s KMD (Kernel Mode Driver) queue, preventing the concurrent utilization of the independent copy engines on each GPU. As the number of ranks increased, the queue depth and latency on GPU 0 surged, ultimately triggering the SGLang timeouts.

## Module

- [x] Transfer Engine (`mooncake-transfer-engine`)
- [ ] Mooncake Store (`mooncake-store`)
- [ ] Mooncake EP (`mooncake-ep`)
- [ ] Mooncake PG (`mooncake-pg`)
- [ ] Integration (`mooncake-integration`)
- [ ] P2P Store (`mooncake-p2p-store`)
- [ ] Python Wheel (`mooncake-wheel`)
- [ ] Common (`mooncake-common`)
- [ ] Mooncake RL (`mooncake-rl`)
- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [x] Performance improvement
- [ ] Other

## How Has This Been Tested?

Prefill-Decode (PD) , 1p1d, input_len 8192, output_len 1024, concurrency 64/128

PR #2691, opened 2026-07-01
[TE] Capture GPU device at registration time in MnnvlTransport / NVLinkTransport

### LINKED ISSUE #2620: [TE] Capture GPU device at registration time in MnnvlTransport / NVLinkTransport instead of inferring from request pointers
## Background

PR #2569 fixed a bug where `MnnvlTransport::allocateSubBatch` (and the same bug in `NVLinkTransport`, addressed in follow-up) created CUDA streams without specifying the target GPU, so all streams landed on GPU 0 and the copy engine became a bottleneck in multi-GPU setups.

The fix defers stream creation until `submitTransferTasks`, where the device is inferred by iterating over request source pointers and calling `cudaPointerGetAttributes`:

```cpp
for (auto &req : request_list) {
    device_id = detectDeviceFromPointer(req.source);
    if (device_id >= 0) break;
}
```

This unblocked the immediate performance issue, but the approach has architectural drawbacks worth tracking as a follow-up.

## Concern

The pointer-inference approach relies on implicit assumptions that are **not enforced by the interface**:

1. The first valid CUDA pointer in the batch represents the correct device for the entire batch.
2. `request.source` is always a non-null CUDA device pointer — the `cudaGetDevice()` fallback is the original bug in disguise (silently selects the calling thread's current device, which may not be where the data lives).
3. `cudaPointerGetAttributes` is called on the hot path of every batch submission, adding a per-request CUDA API round-trip just to recover information we already had.

## Proposed approach

`BufferDesc.location` (e.g., `"cuda:3"`) is already the authoritative source of device information and is available during memory registration. `SunriseLinkTransport` follows this pattern — `addMemoryBuffer` parses the location and stores a per-buffer mapping:

```cpp
// mooncake-transfer-engine/tent/src/transport/sunrise_link/sunrise_link_transport.cpp:901
registered_memory_gpu_id_[reinterpret_cast<void*>(desc.addr)] = location.index();
```

(declaration at `tent/include/tent/transport/sunrise_link/sunrise_link_transport.h:123`)

Applying the same pattern to `MnnvlTransport` and `NVLinkTransport` would:

- Eliminate pointer inference and the `cudaPointerGetAttributes` call on every submission
- Remove the implicit "first-pointer wins" assumption — the transport always knows which GPU each registered buffer lives on
- Make the device contract explicit at registration boundaries instead of recovered at use time
- Naturally handle batches that mix buffers from different GPUs (currently silently bound to whichever pointer is inspected first)

## Scope

- [ ] Add a `registered_memory_gpu_id_` (or equivalent) map to `MnnvlTransport`, populated from `BufferDesc.location` in `addMemoryBuffer`
- [ ] Same for `NVLinkTransport`
- [ ] Use the registered device ID directly in `submitTransferTasks` for stream creation; remove `detectDeviceFromPointer` walks
- [ ] Validate behavior when a batch references multiple devices (reject? group? — interface decision)

## References

- PR #2569 — original bug fix using pointer inference
- `SunriseLinkTransport` — existing precedent for registration-time device capture

xiangui33423:
## Summary
  Fixes #2620
  
  This PR moves GPU device ID capture from submission time (`submitTransferTasks`) to registration time (`addMemoryBuffer`), following the `SunriseLinkTransport` pattern as suggested by
  @staryxchen.
  
  ## Changes
  - Add `registered_memory_gpu_id_` map to both `MnnvlTransport` and `NVLinkTransport`
  - Populate device ID from `BufferDesc.location` during `addMemoryBuffer`
  - Use registered device ID in `submitTransferTasks` instead of `cudaPointerGetAttributes`
  - Keep `detectDeviceFromPointer` as fallback for backward compatibility
  - Clean up device mapping in `removeMemoryBuffer`
  
  ## Benefits
  - **Performance**: Eliminates per-request CUDA API overhead
  - **Architecture**: Makes device assignment explicit at registration boundaries
  - **Consistency**: Aligns with `SunriseLinkTransport` implementation

  ## Testing
  - ✅ Passes clang-format-20 checks
  - ✅ Static analysis confirms correctness
  - ⚠️  Testing on CUDA hardware recommended


staryxchen (changes requested):
1. Fix the CPU-buffer fallback so a registered -1 doesn't re-run cudaPointerGetAttributes.
2. Either implement per-device dispatch for multi-GPU batches or update the PR description/issue comment to make clear this PR is single-GPU-batch-only and the multi-GPU work is a follow-up.

staryxchen:
Taking another pass — I think there's a cleaner path.

**Two issues with the current approach:**

1. The tri-state return (`-2`/`-1`/`>=0`) + two-pass loop + `found_registered_gpu` flag all exist to keep `detectDeviceFromPointer` as a fallback. But an unregistered pointer reaching `submitTransferTasks` is a bug — reject it with `InvalidArgument` instead of papering over it. Drop the fallback and all four pieces of complexity go with it.

2. Multi-GPU batches: silently running every request on the first-found GPU's stream is worse than either grouping or rejecting. Per #2620 the intent is per-device grouping.

**Simpler alternative — no new map at all:**

`SegmentDesc.buffers` already indexes locally registered buffers, `BufferDesc.location` already carries the device id (`"cuda:N"`), and `SegmentDesc::findBuffer()` already does the address→buffer lookup (used in `nvlink::relocateSharedMemoryAddress`). So `submitTransferTasks` can just do:

```
for each request:
    buf = local_segment.findBuffer(req.source, req.length)
    if (!buf) return InvalidArgument
    dev = LocationParser(buf->location).index()   // -1 for cpu
    bucket_by_device[dev].push_back(req)
for each (dev, bucket):
    getStreamFromPool(dev) + startTransfer per request
```

- No new map, no new mutex, no fallback path
- Multi-GPU dispatched per-device internally (matches the #2620 guidance)
- No public API change — all inside the transport
- If scan cost matters later, add an addr→BufferDesc index inside `SegmentDesc` — the improvement benefits every transport at once, instead of each keeping its own copy

WDYT?

xiangui33423:
Thanks for the detailed review! You're absolutely right — this is a much cleaner approach.
  
I agree with both points:

  1. Reject unregistered pointers instead of fallback: The tri-state + two-pass logic was indeed over-engineered. If an unregistered pointer reaches submitTransferTasks, that's a contract violation and should fail fast with InvalidArgument.
  2. Per-device bucketing: Silently using the first-found device's stream for mixed-GPU batches is incorrect behavior. Per-device grouping matches the original #2620 intent.

 The SegmentDesc.buffers + findBuffer() approach is much better:
  - Leverages existing infrastructure (findBuffer is already used in relocateSharedMemoryAddress)
  - No new synchronization primitives
  - Multi-GPU dispatch happens internally without API surface changes
  - If we need optimization later, an addr→BufferDesc index in SegmentDesc benefits all transports at once

I'll refactor along these lines:
```
  for each request:
      buf = local_segment.findBuffer(req.source, req.length)
      if (!buf) return InvalidArgument("Unregistered buffer")
      dev = LocationParser(buf->location).index()  // -1 for cpu
      bucket_by_device[dev].push_back(req)

  for each (dev, bucket):
      stream = getStreamFromPool(dev)
      // dispatch bucket on stream
```

 Will push the update shortly.

staryxchen:
Latest revision looks good architecturally — the `findBuffer`/`SegmentDesc` path is clean, `registered_memory_gpu_id_` map is gone, no new mutex, fallback removed. Direction is right.

One concern before merge: the `InvalidArgument` on mixed-device batches is a behavior break vs. main. Both pre-#2569 and post-#2569 code silently ran mixed-device batches on the first-GPU's stream (relying on CUDA P2P) — not optimal, but not a hard failure. Callers may be depending on this today. Best to keep this PR non-breaking:

- Drop the multi-device `return InvalidArgument` branch. Just take the first GPU's device id for stream setup and let subsequent requests from other devices ride on that stream, same as current main. A proper per-device grouping refactor (which needs `SubBatch` structural changes) can be a follow-up.

Minor: CPU-only batch used to fall back to `cudaGetDevice()`; the new code uses `stream_device = 0` unconditionally. On multi-GPU nodes this changes which device the stream lives on. Restore `cudaGetDevice(&stream_device)` for the `batch_device_id < 0` case.

staryxchen:
Nice — architecturally clean. Three small nits, purely dead/redundant state:

1. **`task.cuda_id = device_id`** is written but never read (grepped both files + full tree). The `cuda_id` field on `MnnvlTask` / `NVLinkTask` looks speculative — drop the assignment and the field itself; add it back when a future refactor actually needs it.

2. **`bool device_determined`** duplicates information already carried by `batch_device_id < 0`. Collapse to:
   ```cpp
   if (batch_device_id < 0 && device_id >= 0) batch_device_id = device_id;
   ```

3. **`LocationParser location(buf->location); int device_id = location.index();`** — the local variable is only read once and shadows `buf->location`. Inline it:
   ```cpp
   int device_id = LocationParser(buf->location).index();
   ```

Also worth flagging (not blocking): `findBuffer` is O(n) over all registered buffers, now called once per request in the submit hot path. Fine for small buffer counts; a `SegmentDesc`-level addr index would be a good follow-up if profiling ever shows it.

staryxchen:
Nits from last round all applied cleanly, thanks. Before merge:

- **Update PR description** — it still describes the removed `registered_memory_gpu_id_` map and `detectDeviceFromPointer` fallback, none of which exist in the current code anymore.
- **Stale includes**: `#include <map>` (both headers) and `#include <mutex>` (mnnvl header) are leftovers from the earlier map-based design. Neither is used now.
- **NVLink `removeMemoryBuffer` wildcard branch** is unrelated to the main change — either mention it in the description or split into its own PR.

staryxchen:
Heads-up: this doesn't compile. On line 137 of `nvlink_transport.cpp` (and the mirror in `mnnvl_transport.cpp`):

```cpp
SegmentDesc* local_segment = segment_manager.getLocal();
```

`SegmentManager::getLocal()` returns `SegmentDescRef` (== `std::shared_ptr<SegmentDesc>`), which has no implicit conversion to a raw pointer. Verified locally with `ninja tent_xport_nvlink tent_xport_mnnvl`:

```
error: cannot convert 'mooncake::tent::SegmentDescRef'
    {aka 'std::shared_ptr<mooncake::tent::SegmentDesc>'}
    to 'mooncake::tent::SegmentDesc*' in initialization
```

Before the next push, please:
1. Build locally with CUDA enabled (`ninja tent_xport_nvlink tent_xport_mnnvl`) to make sure it actually compiles — this PR touches only CUDA-gated files, so CI green ≠ compiles.
2. Actually run the modified transports (a minimal register → submit → get status → remove flow on both CPU and GPU buffers) to confirm the new `findBuffer`-based path works end-to-end. Static + type-check alone isn't enough given the change moves logic across boundaries.

xiangui33423:
## CI Build Failure Analysis

### Build Failure Root Cause
The build failure is **not related to changes in this PR**. The failure is caused by a temporary network issue with the sccache service:

```
sccache: error: Server startup failed: cache storage failed to read
dns error: failed to lookup address information: Try again
```

This is a transient infrastructure issue with the Azure Blob Storage cache backend used by sccache. **Re-running the CI should resolve this.**

### Compilation Warnings
The compilation warnings shown in the build log are **pre-existing in the codebase** and not introduced by this PR:

**Files modified in this PR:**
- `mooncake-transfer-engine/include/tent/transport/mnnvl/mnnvl_transport.h`
- `mooncake-transfer-engine/include/tent/transport/nvlink/nvlink_transport.h`
- `mooncake-transfer-engine/tent/src/transport/mnnvl/mnnvl_transport.cpp`
- `mooncake-transfer-engine/tent/src/transport/nvlink/nvlink_transport.cpp`

**Files with warnings (not modified by this PR):**
- `tcp_transport.cpp` - unused `isCudaMemory` function
- `ibgda_device_transport.cpp` - unused `stream` variable
- `rpc_communicator.cpp` - visibility warnings
- `mlx5gda.cpp` - various type conversion warnings

These warnings exist in the main branch and are unrelated to the GPU device capture changes in this PR.

### Recommendation
Please re-trigger the CI build to resolve the sccache network issue.

staryxchen (approved):
LGTM

PR #2696, opened 2026-07-01
[Bugfix] Separate post and poll in RDMA worker pool

### LINKED ISSUE #2693: [Bug]: Uneven multi-GPU transfer performance under heavy data fragmentation
### Bug Report

Internode GDR transfer performance tests reveal significant discrepancies in Mooncake transfer speeds across different GPUs under small message sizes and large batch sizes.

Our test case is <message size 4096, batch size 160000>, performance is shown bellow: 

Image

With a single QP, slow GPUs achieve approximately 0.9 GB/s bandwidth, while fast GPUs reach around 2.7 GB/s.

The fast/slow GPU performance pattern varies drastically, yet it is consistently tied to the receiver side. Pinning the receiver to a specific topology or switching to a different receiver host alters the fast/slow GPU pattern. Modifications on the sender side barely affect this pattern at all.

Whether the receiver is topology-pinned or not changes the fast/slow GPU pattern. However, pinning to an incorrect topology yields identical performance patterns as pinning to the correct affinity topology. We suspect this fast/slow GPU imbalance stems from Mooncake filtering out unrelated network adapters during topology binding.

We notice that this imbalance occurs on different version of Mooncake, including the newest v0.3.11.post1.

### Before submitting...

c-guo16:
Dedicate thread 0 to polling all CQs and remap the remaining workers to cover all posting shards.

## Description

Fixes #2693.

This PR fixes uneven multi-GPU transfer performance caused by asymmetric work distribution in the RDMA worker pool.

Before this change, each worker both posted sends and polled a subset of CQs. Under the default configuration (`MC_WORKERS_PER_CTX=2`, `MC_NUM_CQ_PER_CTX=1`), this effectively meant that worker 0 polled the only CQ while the other worker did not poll any CQ. As a result, worker 0 had less posting capacity than the other workers.

At the same time, slices are sharded by `(target_id * 10007 + device_id) % kShardCount`, and each worker is responsible for a fixed subset of shards. Since `device_id` is derived from the selected peer NIC, the selected receiver-side NIC/topology affects which worker a slice is assigned to. When more slices for a given GPU land on the polling-heavy worker, that GPU becomes slower, which explains why the fast/slow GPU pattern is strongly receiver-dependent and changes with receiver topology binding or receiver host selection.

This change separates the two responsibilities: thread 0 becomes a dedicated poller for all CQs, and the remaining workers handle posting only. Posting workers are also remapped to ensure that all shards remain covered without leaving one worker on a slower post path.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test results:**
- Validated in Mooncake, NIXL + Mooncake, and vLLM-based application scenarios across multiple GPU models.
- After the fix, the previously slower GPUs can achieve performance comparable to the faster GPUs.

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

Used CodeBuddy to help apply the patch to the current tree and draft the PR description. The change was reviewed manually before submission.

### DIFF
--- a/mooncake-transfer-engine/src/transport/rdma_transport/worker_pool.cpp
+++ b/mooncake-transfer-engine/src/transport/rdma_transport/worker_pool.cpp
@@ -65,6 +65,26 @@ static int selectPeerDevice(RdmaTransport::SegmentDesc *peer_segment_desc,
                                        buffer_id, device_id, retry_count);
 }
 
+static bool workerCanPost(int thread_id) {
+    return kTransferWorkerCount == 1 || thread_id != 0;
+}
+
+static bool workerCanPoll(int thread_id) {
+    return kTransferWorkerCount == 1 || thread_id == 0;
+}
+
+static void getPostingShardAssignment(int thread_id, int &post_tid,
+                                      int &post_count) {
+    assert(workerCanPost(thread_id));
+    if (kTransferWorkerCount > 1) {
+        post_tid = thread_id - 1;
+        post_count = kTransferWorkerCount - 1;
+    } else {
+        post_tid = thread_id;
+        post_count = kTransferWorkerCount;
+    }
+}
+
 WorkerPool::WorkerPool(RdmaContext &context, int numa_socket_id)
     : context_(context),
       numa_socket_id_(numa_socket_id),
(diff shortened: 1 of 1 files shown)


alogfans (approved):
LGTM

PR #2697, opened 2026-07-01
[Store] Inline memcpy for small transfers to bypass thread pool

ZhijunLStudio:
## Description

Execute memcpy directly on the calling thread for transfers ≤ 1 MB, bypassing the single-threaded MemcpyWorkerPool and its scheduling overhead (~20–50 µs per transfer).

For small transfers (128 KB–1 MB range), the actual memcpy cost (~10–20 µs) is often lower than the thread-pool scheduling overhead. Direct execution eliminates this overhead entirely.

Also adds `EmptyOperationState` constructor that pre-sets the result field, allowing inline-completed transfers to call `get_result()` without assertion failures.

**Note:** The same-node shortcut (treating same-IP/different-port endpoints as local) was intentionally excluded from this PR. It has a known cross-process segfault risk that needs further analysis.

## Module

## Type of Change

- [x] Performance improvement

## How Has This Been Tested?

**Test commands:**
```bash
# Build Mooncake Store
```

**Test results:**
- [x] Manual testing done — E2E benchmark shows meaningful GET improvement

## Checklist

- [ ] For changes >500 LOC: N/A (81 lines)

## AI Assistance Disclosure

- [ ] No AI tools were used

### DIFF
--- a/mooncake-store/include/transfer_task.h
+++ b/mooncake-store/include/transfer_task.h
@@ -113,7 +113,9 @@ class OperationState {
  */
 class EmptyOperationState : public OperationState {
    public:
-    EmptyOperationState() {
+    explicit EmptyOperationState(
+        TransferStrategy strategy = TransferStrategy::EMPTY)
+        : strategy_(strategy) {
         // Pre-set result so get_result() does not assert - this state is
         // always considered successfully completed (no async work needed).
         result_.emplace(ErrorCode::OK);
@@ -123,9 +125,10 @@ class EmptyOperationState : public OperationState {
 
     void wait_for_completion() override {}
 
-    TransferStrategy get_strategy() const override {
-        return TransferStrategy::EMPTY;
-    }
+    TransferStrategy get_strategy() const override { return strategy_; }
+
+   private:
+    TransferStrategy strategy_;
 };
 
 /**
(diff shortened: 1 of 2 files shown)


yuechen-sys:
Hello @ZhijunLStudio, thanks for your contribution toward Mooncake!

There are several concerns:

### 1. GPU buffer copy

  Before this change, MemcpyWorkerPool checked both src and dst with `gpu_staging::IsDevicePointer()` and used `gpu_staging::CopyAuto()` when either side was device memory. The new <=1MB inline path always calls `std::memcpy()`, which is only valid for host-accessible memory. For small H2D/D2H/D2D local transfers, this can fail with an invalid access or crash instead of returning TRANSFER_FAIL.

### 2. Submission-thread blocking under many small copies

  I am also concerned that this changes the scheduling semantics for workloads with many small memcpy transfers.

  Previously `submitMemcpyOperation()` mostly enqueued work into `MemcpyWorkerPool` and returned a future, so the caller could continue submitting additional transfers. With this change, every <=1MB copy runs synchronously on the submitting thread. That may improve the single-transfer case where the caller immediately waits, but in batch or pipeline workloads it can make the submitter spend all its time copying and cause head-of-line blocking, especially if the submitter is an RPC/client worker thread.

PR #2699, opened 2026-07-01
[Store] Add optimistic lock-free fast-path for tenant quota checks

ZhijunLStudio:
## Description

Add an optimistic shared-lock fast-path to `ComputeTenantQuotaDeficit` that avoids acquiring the shard exclusive mutex in the common case.

Instead of taking the full exclusive lock, the fast path acquires a shared lock via `SharedMutexLocker` and performs atomic reads of `used_bytes`, `reserved_bytes`, and `effective_quota_bytes`. If sufficient quota is available, the function returns immediately. Only when quota is insufficient does it fall back to the exclusive-lock exact check.

**Key implementation details:**
- `TenantQuotaState::used_bytes/reserved_bytes/effective_quota_bytes` are now `std::atomic<uint64_t>` with explicit copy/move constructors
- `TenantQuotaShard::mutex` changed from `std::mutex` to `SharedMutex` for reader-writer locking
- `DrainJob::mutex` likewise changed to `SharedMutex`
- Fast path uses `SharedMutexLocker` with `shared_lock` for safe concurrent map access
- All field accesses updated to use `.load()`, `.store()`, `.fetch_add()`, `.fetch_sub()` with `std::memory_order_relaxed`

Adds a `--disable_quota_fast_path` flag (default false) for debugging and ablation studies.

**Config chain:** `gflags/config file → MasterConfig → MasterServiceSupervisorConfig → WrappedMasterServiceConfig → MasterServiceConfig → MasterService`

## Module

## Type of Change

- [x] Performance improvement

## How Has This Been Tested?

**Test commands:**
```bash
cmake --build builddir --target mooncake_master
cd builddir && ctest -R MasterService

# E2E benchmark with eviction pressure
python3 test_kvcache_e2e.py --prefill-ratio 0.85 --threads 1,2,4,8
```

**E2E results:** CONC_GET t=8: 58,192 ops/s (verified on this branch).

### DIFF
--- a/mooncake-store/include/master_config.h
+++ b/mooncake-store/include/master_config.h
@@ -132,6 +132,9 @@ struct MasterConfig {
     // liveness window. Default 1 is conservative; small-object or RDMA-
     // rich clusters may safely raise it.
     uint32_t promotion_max_per_heartbeat = 1;
+
+    // Ablation study controls (for paper experiments)
+    bool disable_quota_fast_path = false;
 };
 
 class MasterServiceSupervisorConfig {
@@ -212,6 +215,9 @@ class MasterServiceSupervisorConfig {
     uint32_t promotion_queue_limit = 50000;
     uint32_t promotion_max_per_heartbeat = 1;
 
+    // Ablation study controls (for paper experiments)
+    bool disable_quota_fast_path = false;
+
     // Pod identity for K8s label-based routing
     std::string pod_name;
     std::string pod_namespace;
@@ -253,6 +259,7 @@ class MasterServiceSupervisorConfig {
         promotion_admission_threshold = config.promotion_admission_threshold;
         promotion_queue_limit = config.promotion_queue_limit;
         promotion_max_per_heartbeat = config.promotion_max_per_heartbeat;
+        disable_quota_fast_path = config.disable_quota_fast_path;
         rpc_port = static_cast<int>(config.rpc_port);
         rpc_thread_num = static_cast<size_t>(config.rpc_thread_num);
 
@@ -403,6 +410,8 @@ class WrappedMasterServiceConfig {
     uint32_t promotion_admission_threshold = 2;
     uint32_t promotion_queue_limit = 50000;
     uint32_t promotion_max_per_heartbeat = 1;
+    // Ablation study controls (for paper experiments)
+    bool disable_quota_fast_path = false;
     std::string ha_backend_type = "etcd";
     std::string ha_backend_connstring;
     std::string cluster_id = DEFAULT_CLUSTER_ID;
@@ -476,6 +485,7 @@ class WrappedMasterServiceConfig {
         promotion_admission_threshold = config.promotion_admission_threshold;
         promotion_queue_limit = config.promotion_queue_limit;
         promotion_max_per_heartbeat = config.promotion_max_per_heartbeat;
+        disable_quota_fast_path = config.disable_quota_fast_path;
         ha_backend_type = config.ha_backend_type;
         ha_backend_connstring = ResolveConfiguredHABackendConnstring(
             ha_backend_type, config.ha_backend_connstring,
@@ -570,6 +580,7 @@ class WrappedMasterServiceConfig {
         promotion_admission_threshold = config.promotion_admission_threshold;
         promotion_queue_limit = config.promotion_queue_limit;
         promotion_max_per_heartbeat = config.promotion_max_per_heartbeat;
+        disable_quota_fast_path = config.disable_quota_fast_path;
         ha_backend_type = config.ha_backend_type;
         ha_backend_connstring = ResolveConfiguredHABackendConnstring(
             ha_backend_type, config.ha_backend_connstring,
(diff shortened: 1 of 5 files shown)


mooncake-store/src/master_service.cpp
```diff
@@ -1085,6 +1088,29 @@ uint64_t MasterService::ComputeTenantQuotaDeficit(
     const auto normalized_tenant = NormalizeTenantId(tenant_id);
     auto& quota_shard =
         tenant_quota_shards_[getTenantQuotaShardIndex(normalized_tenant)];
+
+    // Fast path: lock-free quota check using atomic reads.
+    // In the common case (sufficient quota), we avoid acquiring the mutex.
+    if (!disable_quota_fast_path_) {
+        auto fast_quota_it = quota_shard.tenants.find(normalized_tenant);
```
Lin-z-w, line 1097: This unlocked lookup is not safe. `quota_shard.tenants` is a `std::map` that is mutated under `quota_shard.mutex` elsewhere, e.g. policy apply/reload can erase entries and `try_emplace` new ones. Calling `find()` here without the same mutex can race with those writers and is undefined behavior. The fast path needs a synchronization scheme that also protects the container lifetime, not just the scalar fields.

mooncake-store/src/master_service.cpp
```diff
@@ -1085,6 +1088,29 @@ uint64_t MasterService::ComputeTenantQuotaDeficit(
...
+        if (fast_quota_it != quota_shard.tenants.end()) {
+            const auto& state = fast_quota_it->second;
+            const uint64_t used =
+                __atomic_load_n(&state.used_bytes, __ATOMIC_RELAXED);
```
Lin-z-w: These `__atomic_load_n` calls are mixed with ordinary non-atomic writes to the same fields. `TenantQuotaState::{used_bytes,reserved_bytes,effective_quota_bytes}` are plain `uint64_t`s, and writers update them normally while holding the shard mutex. Atomic reads do not make concurrent non-atomic writes data-race-free. Either these fields need to become atomics and be updated atomically everywhere, or this read path still needs to hold the mutex.

mooncake-store/include/master_config.h
```diff
@@ -132,6 +132,9 @@ struct MasterConfig {
     // liveness window. Default 1 is conservative; small-object or RDMA-
     // rich clusters may safely raise it.
     uint32_t promotion_max_per_heartbeat = 1;
+
+    // Ablation study controls (for paper experiments)
```
Lin-z-w: Do we need to carry this runtime config in the mainline? The comment says this is for paper ablation experiments, but the field is threaded through the full production config chain. If this is only for benchmarking, I would prefer keeping it out of the runtime config. If it is intended as a production rollback/feature gate, the name/comment should say that and explain why the operational toggle is needed.

Lin-z-w (changes requested):
I think the current fast path is not safe to merge as written. It skips the shard mutex while still reading a mutex-protected `std::map`, and it uses atomic builtins on fields that are otherwise plain non-atomic variables. Actual quota admission still appears to go through the locked `ReserveTenantQuota` path, so this is not necessarily an over-admission bug, but it is undefined behavior under concurrent quota policy updates/accounting changes.

ZhijunLStudio:
@Lin-z-w Thanks for the thorough review! All three concerns have been addressed in the latest commit:

1. **Map access safety** — `TenantQuotaShard::mutex` is now `SharedMutex`. The fast path takes a shared lock, so `std::unordered_map::find()` is properly protected.

2. **Atomic correctness** — `used_bytes`, `reserved_bytes`, `effective_quota_bytes` are now `std::atomic<uint64_t>` with explicit copy/move constructors. All access sites use `.load()`/`.store()`/`.fetch_add()`/`.fetch_sub()` with `std::memory_order_relaxed`.

3. **Config flag** — Kept `disable_quota_fast_path` as a production feature gate / rollback switch (not just for paper experiments).

Lin-z-w:
Thanks for working on this optimization. I think we may want to hold off on merging this PR for now.

I’m planning a broader change to the multi-tenant quota locking model where quota admission/accounting will move away from the current shard-mutex-protected check/update flow and use CAS-based atomic operations instead. If that lands, this PR’s partial fast path in `ComputeTenantQuotaDeficit()` is likely to become unnecessary or need to be rewritten around the new model.

So I would prefer not to add this intermediate fast path plus the `disable_quota_fast_path` config chain right now. It adds production-facing config and synchronization complexity for a path that is likely to be superseded by the upcoming CAS-based quota model.

My suggestion is to pause or close this PR, then revisit the optimization after the quota accounting model is changed. At that point we can benchmark the new CAS path directly and decide whether any additional fast path is still needed.

ZhijunLStudio:
@Lin-z-w Thanks for the thoughtful feedback and for sharing the plan around the CAS-based quota model. That makes sense — if the quota admission/accounting is moving to CAS atomics, this shared-lock fast path would indeed be superseded. I'll close this PR for now and revisit the fast-path idea after the new quota model lands, at which point we can benchmark the CAS path directly and decide if any additional optimization is needed. Appreciate the guidance!

PR #2700, opened 2026-07-01
[Store] Parallelize BatchPutEnd and BatchPutRevoke across metadata shards

ZhijunLStudio:
## Description

Parallelize `BatchPutEnd` and `BatchPutRevoke` by grouping keys by metadata shard index and processing each shard group concurrently via `std::async`.

The original implementation processes all shards sequentially, which becomes a bottleneck under high concurrency (1024 shards, each with lock acquisition). The parallel version dispatches shard groups to independent `std::async` tasks, avoiding serialization across shards.

Adds two config flags:
- `--min_batch_parallel_keys` (default 16): minimum batch size to trigger parallel processing (avoids thread-spawn overhead on tiny batches)
- `--disable_parallel_batches` (default false): fallback to sequential processing for debugging

**E2E ablation:** +17.1% GET throughput — the largest single contributor among all optimizations.

## Module

## Type of Change

- [x] Performance improvement

## How Has This Been Tested?

**Test commands:**
```bash
# Build Mooncake Store
```

**Test results:**
- [x] Manual testing done — E2E ablation shows +17.1% GET throughput

## Checklist

- [ ] For changes >500 LOC: N/A (133 lines)

## AI Assistance Disclosure

- [ ] No AI tools were used

### DIFF
--- a/mooncake-store/include/master_config.h
+++ b/mooncake-store/include/master_config.h
@@ -132,6 +132,10 @@ struct MasterConfig {
     // liveness window. Default 1 is conservative; small-object or RDMA-
     // rich clusters may safely raise it.
     uint32_t promotion_max_per_heartbeat = 1;
+
+    // Ablation study controls (for paper experiments)
+    bool disable_parallel_batches = false;
+    int32_t min_batch_parallel_keys = 16;
 };
 
 class MasterServiceSupervisorConfig {
@@ -212,6 +216,10 @@ class MasterServiceSupervisorConfig {
     uint32_t promotion_queue_limit = 50000;
     uint32_t promotion_max_per_heartbeat = 1;
 
+    // Ablation study controls (for paper experiments)
+    bool disable_parallel_batches = false;
+    int32_t min_batch_parallel_keys = 16;
+
     // Pod identity for K8s label-based routing
     std::string pod_name;
     std::string pod_namespace;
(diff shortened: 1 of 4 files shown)


yokinoshitayoki:
Thanks for the contribution and the benchmark result.

But we would prefer not to merge this parallelization approach for now. The main concern is that adding `std::async` inside `BatchPutEnd` / `BatchPutRevoke` can create many short-lived tasks per request, up to the number of touched metadata shards. Under high concurrency this may introduce thread scheduling overhead, oversubscription, and less predictable tail latency. The added config flags also look more like ablation controls than production-facing behavior.

We agree that `BatchPutEnd` and `BatchPutRevoke` are worth optimizing, but we would rather optimize the batch operations themselves instead of parallelizing the current per-key path. For example, a future direction may be to group keys by metadata shard and process each shard group more efficiently inside the batch implementation, potentially reducing repeated per-key lock acquisition and metadata lookup overhead without spawning per-request async tasks.

The benchmark data is still useful and can help guide that follow-up work.

ZhijunLStudio:
@yokinoshitayoki Thanks for the detailed review. Those are fair concerns — `std::async` per-request spawning tasks up to the number of touched shards does risk thread oversubscription and tail latency under high concurrency, and I agree the config flags lean more toward ablation controls than production knobs. The direction of optimizing the batch internals (reducing per-key lock acquisition / metadata lookup overhead) sounds like the right approach. I'll close this PR. Glad the benchmark data can still be useful for guiding that follow-up work — happy to share more details if needed.

PR #2718, opened 2026-07-02
[Store] Integrate Redis election backend into Master HA framework

skygragon:
## Description

  Integrate Redis as an alternative election backend for the Master HA framework, enabling P2P Store deployments to use Redis
  instead of etcd for leader election.

  Key changes:
  - Make `MasterViewHelper` methods virtual and add `RedisMasterViewHelper` subclass that implements leader election via
  `RedisHelper`
  - Add `CreateMasterViewHelper()` factory to select etcd or redis backend at runtime
  - Add `election_backend` config field (`"etcd"` default, `"redis"` for P2P HA) with Redis-specific flags (`redis_endpoint`,
  `redis_password`, `redis_db_index`, `redis_master_view_ttl_sec`, `redis_heartbeat_interval_sec`)
  - Wire Redis election config through `MasterConfig` → `MasterServiceSupervisorConfig` → gflags in `master.cpp`, with
  validation
  - Merge election test cases into `redis_helper_test`

  ## Module

  - [ ] CI/CD
  - [ ] Docs
  - [ ] Other

  ## Type of Change

  - [ ] Bug fix
  - [x] New feature
  - [ ] Refactor
  - [ ] Breaking change
  - [ ] Documentation update
  - [ ] Performance improvement
  - [ ] Other

  ## How Has This Been Tested?

  **Test commands:**
  ```bash
  # Build with Redis support
  docker exec mooncake-dev bash -c "cd /workspace && rm -rf build && mkdir build && cd build && cmake ..
  -DCMAKE_BUILD_TYPE=Debug && make -j\$(nproc)"

  # Run Redis helper tests (includes election tests)
  docker exec mooncake-dev bash -c "cd /workspace/build && ctest -R redis_helper_test -V"
```

  Test results:
  - [ ] Unit tests pass

  Checklist

  AI Assistance Disclosure

  - [ ] No AI tools were used

  Claude Code was used for code formatting and PR description drafting. Human author is responsible for all design decisions
  and code changes.

### DIFF
--- a/mooncake-store/include/ha_helper.h
+++ b/mooncake-store/include/ha_helper.h
@@ -4,6 +4,7 @@
 #include <csignal>
 #include <glog/logging.h>
 
+#include <memory>
 #include <string>
 #include <thread>
 #include <ylt/coro_rpc/coro_rpc_server.hpp>
@@ -19,16 +20,20 @@ namespace mooncake {
  *        one master can be elected as leader to serve client requests.
  *        Each master view is associated with a unique version id, which
  *        is incremented monotonically each time the master view is changed.
+ *
+ *        This base class implements the etcd-based election. Redis-based
+ *        election is provided by the RedisMasterViewHelper subclass.
  */
 class MasterViewHelper {
    public:
     MasterViewHelper(const MasterViewHelper&) = delete;
     MasterViewHelper& operator=(const MasterViewHelper&) = delete;
     MasterViewHelper();
+    virtual ~MasterViewHelper();
 
     /*
-     * @brief Connect to the etcd cluster. This function should be called at
-     * first
+     * @brief Connect to the etcd cluster. This function should be called first
+     *        when using the etcd election backend.
      * @param etcd_endpoints: The endpoints of the etcd store client.
      *        Multiple endpoints are separated by semicolons.
      * @return: Error code.
@@ -41,15 +46,21 @@ class MasterViewHelper {
      * @param version: Output param, the version of the new master view.
      * @param lease_id: Output param, the lease id of the leader.
      */
-    void ElectLeader(const std::string& master_address, ViewVersionId& version,
-                     EtcdLeaseId& lease_id);
+    virtual void ElectLeader(const std::string& master_address,
+                             ViewVersionId& version, EtcdLeaseId& lease_id);
 
     /*
      * @brief Keep the master to be the leader. This function blocks until the
      * master is no longer the leader.
      * @param lease_id: The lease id of the leader.
      */
-    void KeepLeader(EtcdLeaseId lease_id);
+    virtual void KeepLeader(EtcdLeaseId lease_id);
+
+    /*
+     * @brief Cancel the keep-alive loop (for graceful shutdown).
+     * @param lease_id: The lease id of the leader.
+     */
+    virtual void CancelKeepAlive(EtcdLeaseId lease_id);
 
     /*
      * @brief Get the current master view.
@@ -57,12 +68,22 @@ class MasterViewHelper {
      * @param version: Output param, the version of the master view.
      * @return: Error code.
      */
-    ErrorCode GetMasterView(std::string& master_address,
-                            ViewVersionId& version);
+    virtual ErrorCode GetMasterView(std::string& master_address,
+                                    ViewVersionId& version);
 
    private:
     std::string master_view_key_;
 };
+
+/*
+ * @brief Create the appropriate MasterViewHelper based on election backend
+ *        configuration. Returns an etcd-based helper by default, or a
+ *        Redis-based helper when election_backend is REDIS.
+ * @param config: The supervisor configuration.
+ * @return: A unique_ptr to the created helper, or nullptr on error.
+ */
+std::unique_ptr<MasterViewHelper> CreateMasterViewHelper(
+    const MasterServiceSupervisorConfig& config);
 
 /*
  * @brief A supervisor class for the master service, only used in HA mode.
--- a/mooncake-store/include/ha_helper.h
+++ b/mooncake-store/include/ha_helper.h
@@ -63,6 +63,12 @@ class MasterViewHelper {
     virtual void CancelKeepAlive(EtcdLeaseId lease_id);
 
     /*
+     * @brief Get the leader lease TTL in seconds for the current election
+     *        backend.
+     */
+    virtual int GetLeaderLeaseTTLSeconds() const;
+
+    /*
      * @brief Get the current master view.
      * @param master: Output param, the ip:port address of the master.
      * @param version: Output param, the version of the master view.
--- a/mooncake-store/include/master_config.h
+++ b/mooncake-store/include/master_config.h
@@ -59,6 +59,16 @@ struct MasterConfig {
     uint64_t max_replicas_per_key;
 
     std::string deployment_mode;
+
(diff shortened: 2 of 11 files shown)

### CODE REVIEW

File: mooncake-store/include/ha_helper.h [hunk 1]
```diff
@@ -19,16 +20,20 @@ namespace mooncake {
...
  *        is incremented monotonically each time the master view is changed.
+ *
+ *        This base class implements the etcd-based election. Redis-based
+ *        election is provided by the RedisMasterViewHelper subclass.
  */
 class MasterViewHelper {
    public:
     MasterViewHelper(const MasterViewHelper&) = delete;
     MasterViewHelper& operator=(const MasterViewHelper&) = delete;
     MasterViewHelper();
```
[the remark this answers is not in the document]
skygragon: This is a cosmetic logging issue with no functional impact. RedisMasterViewHelper overrides all virtual methods that use
  master_view_key_, so the key is never actually read in the Redis path. The misleading log line is harmless. Modifying the
  base class constructor signature would touch upstream code and risk merge conflicts, which outweighs the benefit of fixing a
  log message. I'll skip this one.

mooncake-store/src/redis_master_view_helper.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+namespace mooncake {
+
+RedisMasterViewHelper::RedisMasterViewHelper(const std::string& cluster_id,
+                                             const std::string& redis_endpoint,
+                                             const std::string& password,
+                                             int db_index, int ttl_sec,
+                                             int heartbeat_interval_sec)
+    : redis_helper_(cluster_id, redis_endpoint, password, db_index, ttl_sec,
+                    heartbeat_interval_sec) {}
```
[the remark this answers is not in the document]
skygragon: same as above

Icedcoco:
Thanks for the update. I did a first pass and the overall direction looks reasonable, but I need more time to trace the HA failover semantics before I can approve.

One concern from the first pass: `WatchLeaderSubscribe()` can fall into a subscribe-only wait if the extra polling connection cannot be created. The custom Redis channel appears to publish only after a successful election, while key expiry itself does not publish to that custom channel. In that case, if the current leader key expires and no polling thread is running, a contender may keep waiting in `ElectLeader()` instead of retrying election. I think the subscription path should remain an optimization, but correctness should still rely on polling/retry or an explicit vacancy notification mechanism.

Minor test issue as well: the timeout branches in `redis_helper_test.cpp` call `CancelKeepAlive()` and then join the election thread, but `ElectLeader()` is controlled by `cancel_election_`, not the keepalive flag. If the timeout path is hit, the test can hang instead of failing cleanly.

I’ll continue reviewing the config propagation and failover behavior in more detail.

skygragon:
@Icedcoco Thanks for the careful pass and for catching these.

  I pushed a fix that keeps the subscribe path as an optimization only. If WatchLeaderSubscribe() cannot create the backup polling connection, it now unsubscribes and falls back to pure polling instead of waiting only on the custom Redis channel. That keeps key expiry as the correctness path for retrying election.

  I also fixed the timeout cleanup in redis_helper_test.cpp: the timeout branches now cancel the election/watch path explicitly instead of only calling CancelKeepAlive(), so a failed test should exit cleanly rather than hanging on join().

mooncake-store/src/ha_helper.cpp
```diff
@@ -91,6 +96,10 @@ void MasterViewHelper::KeepLeader(EtcdLeaseId lease_id) {
     EtcdHelper::KeepAlive(lease_id);
 }
 
+void MasterViewHelper::CancelKeepAlive(EtcdLeaseId lease_id) {
+    EtcdHelper::CancelKeepAlive(lease_id);
+}
```
wanyue-wy, line 109: `EtcdHelper::CancelKeepAlive` returns an error code. You should either handle the error inside this function, or return it back to the caller.

mooncake-store/src/ha_helper.cpp
```diff
@@ -125,31 +134,38 @@ int MasterServiceSupervisor::Start() {
...
+        LOG(INFO) << "Init leader election helper (backend="
+                  << (config_.election_backend == ElectionBackend::REDIS
+                          ? "redis"
+                          : "etcd")
+                  << ")...";
+
+        auto mv_helper = CreateMasterViewHelper(config_);
+        if (!mv_helper) {
             return -1;
         }
```
wanyue-wy, line 158: you should print error log here

mooncake-store/src/ha_helper.cpp
```diff
@@ -125,31 +134,38 @@ int MasterServiceSupervisor::Start() {
...
                 server.stop();
             });
 
         // To prevent potential split-brain, wait long enough for the old leader
         // to retire.
-        const int waiting_time = ETCD_MASTER_VIEW_LEASE_TTL;
+        const int waiting_time =
+            config_.election_backend == ElectionBackend::REDIS
+                ? config_.redis_master_view_ttl_sec
+                : static_cast<int>(ETCD_MASTER_VIEW_LEASE_TTL);
```
wanyue-wy: Would it be better to move `ETCD_MASTER_VIEW_LEASE_TTL` and `config_.redis_master_view_ttl_sec` into `mv_helper`? We should try to keep external code unaware of the concrete type of `ElectionBackend`.

mooncake-store/src/master.cpp
```diff
@@ -520,10 +588,52 @@ int main(int argc, char* argv[]) {
...
+        }
+        if (master_config.redis_heartbeat_interval_sec >=
+            master_config.redis_master_view_ttl_sec) {
+            LOG(FATAL) << "redis_heartbeat_interval_sec ("
+                       << master_config.redis_heartbeat_interval_sec
+                       << ") must be less than redis_master_view_ttl_sec ("
+                       << master_config.redis_master_view_ttl_sec << ")";
+            return 1;
+        }
+    }
```
wanyue-wy: 1. Only error out when `enable_ha` is true and `election_backend` is invalid — allow users to not use HA ability.
2. Use `enable_ha` as a large if scope, and inside it, perform detailed defensive checks based on whether the backend is `etcd` or `Redis`.

mooncake-store/src/redis_helper.cpp
wanyue-wy: If `polling_ctx` is null, the function returns early at this point, so the `if (polling_ctx)` check at line 356 is unnecessary.

mooncake-store/include/master_config.h
```diff
@@ -59,6 +59,16 @@ struct MasterConfig {
...
+
+    // Redis election backend configuration (used when election_backend ==
+    // "redis")
+    std::string election_backend = "etcd";  // "etcd" (default) or "redis"
+    std::string redis_endpoint;         // Redis endpoint, e.g. "10.0.0.1:6379"
+    std::string redis_password;         // Redis AUTH password (empty = no auth)
+    int redis_db_index = 0;             // Redis DB index
+    int redis_master_view_ttl_sec = 5;  // Leader key TTL in seconds
+    int redis_heartbeat_interval_sec =
+        2;  // KeepLeader renewal interval (TTL/3)
```
wanyue-wy: if we want to keep `redis_heartbeat_interval_sec` to `TTL/3`, why not directly set it to `redis_master_view_ttl_sec/3`

PR #2723, opened 2026-07-02
[CI] Extract reusable wheel build/publish workflows

mgoin:
Factors the duplicated wheel build steps out of `release.yaml`, `release-cuda13.yaml`, `release-non-cuda.yaml`, and `pre-release.yaml` into a shared `_build-wheel.yaml` (and `_publish-wheel.yaml` for the release tail). Release and Pre-Release now call the same build, so a pre-release tag is a faithful dry run with no build-environment drift. Net −630 lines.

As part of this, the Non-CUDA and all Pre-Release x86 builds move into `manylinux2_28` like the CUDA wheels, so their wheels stop requiring `GLIBCXX_3.4.30` and install on RHEL/Rocky/Alma 8+. EFA/MUSA/NPU are left unchanged.

Validated with `actionlint` + `shellcheck` (no new findings); end-to-end verification is a `vX.Y.Z-rc` tag, which triggers the Pre-Release dry run. AI assistance (Claude) was used.

### DIFF
--- a/.github/workflows/_publish-wheel.yaml
+++ b/.github/workflows/_publish-wheel.yaml
@@ -1,0 +1,49 @@
+name: _publish-wheel
+
+# Shared publish tail: collect this run's wheels, attach to the GitHub Release,
+# and upload to PyPI. Used by the Release workflows.
+
+on:
+  workflow_call:
+    inputs:
+      artifact-pattern:
+        type: string
+        required: true
+    secrets:
+      pypi-token:
+        required: false
+
+jobs:
+  publish:
+    runs-on: ubuntu-22.04
+    permissions:
+      contents: write
+      id-token: write
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v4
+
+      - name: Download all wheel artifacts
+        uses: actions/download-artifact@v4
+        with:
+          path: mooncake-wheel/dist-all
+          pattern: ${{ inputs.artifact-pattern }}
+
+      - name: Prepare wheels for release
+        run: |
+          mkdir -p mooncake-wheel/dist-release
+          find mooncake-wheel/dist-all -name "*.whl" -exec cp {} mooncake-wheel/dist-release/ \;
+          echo "Collected wheels for release:"
+          ls -la mooncake-wheel/dist-release/
+
+      - name: Upload wheels to GitHub Release
+        uses: softprops/action-gh-release@v1
+        with:
+          files: mooncake-wheel/dist-release/*.whl
+
+      - name: Publish package to PyPI
+        if: ${{ github.repository == 'kvcache-ai/Mooncake' }}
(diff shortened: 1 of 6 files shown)


.github/workflows/release.yaml
```diff
@@ -5,223 +5,42 @@ on:
...
+      cuda: sbsa-12.8
+      cmake-generator: Ninja
+      torch-cuda-arch-list: '9.0'
+      cmake-args: >-
+        -DUSE_HTTP=ON -DUSE_CUDA=ON -DUSE_MNNVL=ON -DWITH_EP=OFF
+        -DWITH_STORE_RUST=OFF -DENABLE_SCCACHE=ON -DCMAKE_BUILD_TYPE=Release
+      artifact-prefix: mooncake-wheel-arm64
 
   publish-release:
+    if: ${{ !contains(github.ref_name, '-') }}
```
alogfans: So tags like `v1.2.3-rc1` (i.e., with `-rcX` suffix) will not release? This differs from previous code.

.github/workflows/_publish-wheel.yaml
```diff
@@ -0,0 +1,49 @@
...
+    secrets:
+      pypi-token:
+        required: false
+
+jobs:
+  publish:
+    runs-on: ubuntu-22.04
+    permissions:
+      contents: write
+      id-token: write
```
Aionw, line 21: I think the publish permissions should stay on the caller jobs after this refactor. Previously, the inline `publish-release` jobs declared:

permissions:
  contents: write
  id-token: write

After moving the publish steps into `_publish-wheel.yaml`, those permissions are declared only inside the reusable workflow. My understanding is that reusable workflows cannot elevate `GITHUB_TOKEN` permissions beyond what the caller job grants; permissions can only be maintained or reduced through the call chain. If the repository/org default workflow permissions are restricted, the publish job may not be able to upload wheels to the GitHub Release (`contents: write`) or request the OIDC token used by PyPI publishing (`id-token: write`).

Could we add the permissions back to each caller `publish-release` job, e.g. in `release.yaml`, `release-cuda13.yaml`, and `release-non-cuda.yaml`?

permissions:
  contents: write
  id-token: write

The declaration in `_publish-wheel.yaml` can either be kept as documentation of the reusable workflow’s requirements or removed once the callers explicitly grant these permissions.

.github/workflows/release.yaml
mgoin: The `-rc` tags were never published previously either, they were just built and discard, so I removed the wasted work. To lower the number of changes though I went ahead and kept this old behavior in case something relies on it.

.github/workflows/_publish-wheel.yaml
mgoin: This is a good catch and I just added it back to the callers

Aionw (approved):

PR #2726, opened 2026-07-03
Support TLS authentication for Mooncake Master Client and Python/C API connecting to etcd.

gitgaoqian:
## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
```

**Test results:**
- [ ] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used


Icedcoco:
Thanks for working on etcd TLS support. I’ve started reviewing this PR, but the change spans the Go etcd wrapper, Store/HA paths, C/Python APIs, and docs, so I’ll need more time to go through the full behavior carefully.

A few initial issues already look worth addressing before deeper review:

1. `EtcdHelper::SetTLSConfig` / `IsTLSConfigured` appear to be defined both unconditionally near the top of `etcd_helper.cpp` and again in the non-`STORE_USE_ETCD` fallback section, which should break the `STORE_USE_ETCD=OFF` build.
2. The TLS helper fallback paths call the non-TLS helper while holding `etcd_mutex_`, which can deadlock because the non-TLS helper takes the same non-recursive mutex.
3. The documented Python/C API env fallback does not seem to run unless `set_etcd_tls_config()` is called explicitly; `mooncake_store_setup()` and `MooncakeConfig.load_from_env()` do not currently load the `MC_ETCD_*` variables.
4. Please add/describe focused tests for the TLS paths, including `STORE_USE_ETCD=ON/OFF`, command-line/config/env setup, and Python/C API behavior.

I’ll continue reviewing the HA/client interaction and the Go wrapper details after these initial points are addressed.

Icedcoco:
Thanks for adding etcd TLS support and for addressing the earlier review comments around locking, nil-pointer handling, and the environment fallback path. The overall direction looks reasonable.

Before merging, could we please take another pass over the following points?

1. `STORE_USE_ETCD=OFF` build behavior

`mooncake_master` calls `EtcdHelper::SetTLSConfig()` unconditionally in `mooncake-store/src/master.cpp`, while the non-etcd implementation in `mooncake-store/src/etcd_helper.cpp` calls `LOG(FATAL)`. This causes the master to abort immediately in a non-etcd build. The current Ascend CI log reproduces this with:

`Etcd is not enabled in compilation`

Could we either guard the call with `STORE_USE_ETCD` or make the disabled implementation a harmless no-op, and add/verify a corresponding `STORE_USE_ETCD=OFF` startup/build test?

2. TLS configuration errors should not fall back silently

`SetGlobalTLSConfig()` returns without reporting an error when the CA/certificate/key files cannot be loaded or parsed. The caller then continues, and `NewEtcdClient()` may create a client without TLS (or retain a previous global TLS configuration).

For a TLS authentication feature, could we propagate this error to the setup path and fail explicitly instead of silently falling back to plaintext? It would also be useful to add coverage for missing files, invalid CA data, and mismatched client certificate/key files.

3. Python environment-variable fallback

The documentation says that calling `set_etcd_tls_config()` is optional and that `MC_ETCD_CA_FILE`, `MC_ETCD_CERT_FILE`, and `MC_ETCD_KEY_FILE` are used automatically when it is skipped. However, the dictionary-based pybind `setup()` path appears to call `RealClient::setup_internal()` directly, while the environment fallback is implemented in the C API setup path.

Could we align the implementation and documentation here, or clarify the supported path? A focused Python binding test that exercises `setup()` with only the `MC_ETCD_*` variables set would help confirm the intended behavior.

4. Clearing TLS configuration

The C API documentation says that passing null/empty values disables TLS. At the moment, the empty configuration path returns before clearing the Go global TLS configuration and the C++ `EtcdHelper` state. Because these values are process-global, a later store/client could inherit credentials from an earlier instance.

Could we make the disable/reset behavior explicit and add a set-then-clear test?

5. Build-flag coverage

Could we also verify the intended behavior for `USE_ETCD=ON` with `STORE_USE_ETCD=OFF`? The Transfer Engine metadata client is controlled by `USE_ETCD`, but the new CLI TLS setup is guarded by `STORE_USE_ETCD`. If this combination is supported, it appears to need a separate configuration path or test; otherwise, it would be helpful to document the restriction.

Separately, the branch currently has merge conflicts (`mergeable_state=dirty`). Please rebase/update the branch, resolve the conflicts, and rerun the full relevant CI matrix. The current CI Gate is failing, including the non-etcd master startup failure above, and the wheel test jobs should also be checked after syncing with `main`.

The new tests currently cover configuration state and “does not crash” behavior, but do not exercise a real TLS handshake or certificate failure. Adding a focused integration test, or documenting why that is not practical in this PR, would make the TLS behavior easier to validate.

Thanks again for working on this. Once the conflict and CI issues are resolved, I think the remaining points can be reviewed more confidently.

PR #2741, opened 2026-07-03
[TransferEngine] Prefer private-range IPv4 GIDs over link-local IPv6 in auto-selection

### LINKED ISSUE #2729: [Bug]: GID auto-selection incorrectly prefers link-local IPv6 over routable IPv4 when IP is in 10.0.0.0/8
### Bug Report

# GID auto-selection incorrectly prefers link-local IPv6 over routable IPv4 when IP is in 10.0.0.0/8

## Problem Summary

When `MC_GID_INDEX` is not set, Mooncake's automatic GID selection incorrectly chooses a link-local IPv6 GID (`fe80::`) over a valid routable IPv4-mapped RoCEv2 GID, causing connection failures in production RoCE networks that use the 10.0.0.0/8 address space.

## Environment

- **Mooncake version**: main branch (commit 30f01bdb)
- **RDMA NICs**: Mellanox mlx5 (8 physical ports)
- **Driver**: MLNX_OFED_LINUX-24.10-4.1.4.0
- **Kernel**: 5.15.0-119-generic
- **Network**: RoCEv2, physical datacenter fabric using 10.14.x.x addressing

## Network Configuration

Physical RoCE network (MTU 9000, /26 subnets):

```
ens12np0 (mlx5_0): 10.14.1.72/26
ens11np0 (mlx5_1): 10.14.1.8/26
ens13np0 (mlx5_2): 10.14.1.136/26
ens14np0 (mlx5_3): 10.14.1.200/26
ens16np0 (mlx5_4): 10.14.2.72/26
ens15np0 (mlx5_5): 10.14.2.8/26
ens17np0 (mlx5_6): 10.14.2.136/26
ens18np0 (mlx5_9): 10.14.2.200/26
```

Management network (bond0): `10.20.24.8/24`  
Docker (not used for RDMA): `172.17.0.1/16`

## GID Table

**This issue affects all 8 RoCE interfaces** because all their IPv4 addresses are in 10.14.x.x range. Example from mlx5_4:

```bash
$ show_gids | grep mlx5_4
DEV     PORT  INDEX  GID                                       IPv4            VER    IFACE
mlx5_4  1     0      fe80:0000:0000:0000:966d:aeff:fec6:7726                   v1     ens16np0
mlx5_4  1     1      fe80:0000:0000:0000:966d:aeff:fec6:7726                   v2     ens16np0   ← incorrectly selected
mlx5_4  1     2      0000:0000:0000:0000:0000:ffff:0a0e:0248   10.14.2.72      v1     ens16np0
mlx5_4  1     3      0000:0000:0000:0000:0000:ffff:0a0e:0248   10.14.2.72      v2     ens16np0   ← should be selected
```

Verification of network device association:

```bash
$ cat /sys/class/infiniband/mlx5_4/ports/1/gid_attrs/ndevs/1
ens16np0

$ cat /sys/class/infiniband/mlx5_4/ports/1/gid_attrs/ndevs/3
ens16np0
```

**Both GIDs are associated with the same physical network device.** The same incorrect selection happens on mlx5_0 (10.14.1.72), mlx5_1 (10.14.1.8), mlx5_2 (10.14.1.136), mlx5_3 (10.14.1.200), mlx5_5 (10.14.2.8), mlx5_6 (10.14.2.136), and mlx5_9 (10.14.2.200) — **all 8 RoCE interfaces** consistently prefer link-local IPv6 over their routable IPv4 addresses.

## Observed Behavior

When running `transfer_engine_bench` without `MC_GID_INDEX`:

```
I0626 18:33:44.382241 30023 rdma_context.cpp:1091] Find best gid index: 1 on mlx5_5/ (network state: with network device)
I0626 18:33:44.383456 30023 rdma_context.cpp:273] RDMA device: mlx5_5, LID: 0, GID: (GID_Index 1) fe80::...
```

This causes connection failures:

```
E0626 15:11:21.919103 3846938 rdma_endpoint.cpp:472] Invalid argument: received packet mismatch
```

## Root Cause

The auto-selection logic in [`rdma_context.cpp:715-723`](https://github.com/kvcache-ai/Mooncake/blob/main/mooncake-transfer-engine/src/transport/rdma_transport/rdma_context.cpp#L715-L723) hardcodes the entire 10.0.0.0/8 range as "overlay network":

```cpp
static inline bool isOverlayIPv4(const struct in6_addr *addr) {
    ...
    uint8_t octet1 = (ipv4 >> 24) & 0xFF;
    if (octet1 == 10) return true;  // ← treats all 10/8 as overlay
    ...
}
```

This causes:

1. GID index 3 (`10.14.2.72`, routable) → classified as `kNetworkDegraded` (priority 2)
2. GID index 1 (`fe80::`, link-local) → classified as `kNetworkDegraded` (priority 2)
3. When priorities tie, selection falls back to **lowest index** → selects index 1

## Expected Behavior

A routable IPv4-mapped RoCEv2 GID (index 3) should be strongly preferred over a link-local IPv6 GID (index 1), especially when:

- Both are associated with the same physical network device (`ens16np0`)
- The IPv4 address is on a production datacenter fabric (MTU 9000, /26 subnet)
- The interface name clearly indicates a physical NIC (not `docker0`, `cni0`, etc.)

## Comparison with NCCL

NCCL does not have this issue because its GID selection logic:

- Prioritizes RoCEv2 with IPv4-mapped addresses
- Does not assume all 10.0.0.0/8 addresses are overlay networks
- Scans GID table from highest to lowest index

## Workaround

Setting `MC_GID_INDEX=3` manually works correctly:

```bash
export MC_GID_INDEX=3
./transfer_engine_bench --mode=target --auto_discovery --metadata_server=P2PHANDSHAKE
```

Log output confirms correct selection:

```
I0626 15:16:22.581650 1152642 rdma_context.cpp:1110] Using user-specified GID index: 3 on mlx5_4/ (with network device)
I0626 15:16:22.605633 1152642 rdma_context.cpp:273] RDMA device: mlx5_4, LID: 0, GID: (GID_Index 3) 00:00:00:00:00:00:00:00:00:00:ff:ff:0a:0e:02:48
```

And the benchmark runs successfully with full bandwidth.

## Possible Direction

The core issue is that treating the entire 10.0.0.0/8 range as an overlay network is too aggressive — many enterprise datacenters use 10/8 for their production RDMA fabric. It may help to make the overlay classification more precise (e.g. also consider the interface name, since a physical NIC like `ens16np0` is clearly not a container overlay such as `docker0`/`cni0`), and/or to ensure a routable IPv4-mapped RoCEv2 GID is preferred over a link-local IPv6 GID when both map to the same device.

I'm happy to help test any fix. Deferring to the maintainers on the best approach.

## Additional Context

This issue affects production deployments where:

- Datacenter networks commonly use RFC 1918 private addressing (10/8, 172.16/12, 192.168/16)
- RDMA is deployed on physical infrastructure, not containers
- Users expect auto-selection to "just work" like it does with NCCL

The current heuristic optimizes for containerized/cloud environments but breaks bare-metal datacenter deployments.

---

**Impact**: Users must manually set `MC_GID_INDEX` on every deployment, negating the benefit of auto-selection and causing confusion when benchmarks fail with default settings.

### Before submitting...

n-WN:
Fixes #2729.

`isOverlayIPv4()` classifies every RFC 1918 / CGNAT address as "overlay", which drops routable 10.x datacenter-fabric GIDs into the same degraded tier as link-local `fe80::` GIDs; the lowest-gid-index tie-break then deterministically picks the link-local entry (they occupy indices 0/1 on mlx5), which can only ever work same-L2. Every RoCEv2 deployment addressed from 10/8 — like the reporter's, and like SGLang's default docker setups — gets a broken auto-selection.

## Design

Split the degraded tier rather than reorder anything else, shaped around the compatibility concern raised in the issue:

- New tier `k{Network,NoNetwork}PrivateV4`: private-range IPv4-mapped GIDs (10/8, 172.16/12, 100.64/10) rank **strictly below** genuinely routable GIDs (public v4, global v6 — unchanged) and **strictly above** link-local / overlay-named-interface GIDs.
- The interface-name overlay heuristic (`docker*`/`cni*`/`flannel*`/…) remains the strongest demotion signal: a v4-mapped GID on `docker0` still ranks below a private-range GID on a physical NIC.
- **Every pre-existing cross-tier ordering is preserved** — the change is exactly "the former two-way tie inside degraded is now ordered". All 17 existing `rdma_gid_probe_test` cases pass unmodified; deployments pinning `MC_GID_INDEX` are unaffected. For the reported topology this restores the effective pre-4d7c1a19 selection (which the reporter confirmed worked, e.g. 0.3.11.post1).

The policy lives entirely in `rdma_gid_probe.h` (pure functions), so the fix is fully covered by table-driven unit tests — including one that reproduces the exact GID table from the report.

## Validation (multi-GPU Linux host with real mlx5 RoCEv2 hardware on a private RFC1918 fabric, with the same link-local and IPv4-mapped GID-table shape as the report)

Unit: `rdma_gid_probe_test` 22/22 (17 pre-existing unmodified + 5 new).

Live A/B with `transfer_engine_bench` (P2PHANDSHAKE, RDMA):

| build | auto-selected GID on mlx5_0 |
|---|---|
| main (b5ad133) | selects a link-local GID |
| this PR | selects the routable IPv4-mapped private-fabric GID |

End-to-end with the fixed selection across two mlx5 interfaces (read, 64 KB × 128 batch × 4 threads, 5 s): **24.10 GB/s**, no handshake retries.

@pzcyyj would you mind verifying this build on your cross-subnet fabric? On the validation host the mis-selection reproduces on main exactly as you described and this branch selects the private-fabric GID, but yours is the topology where the link-local choice actually breaks connections.

### DIFF
--- a/mooncake-transfer-engine/include/transport/rdma_transport/rdma_gid_probe.h
+++ b/mooncake-transfer-engine/include/transport/rdma_transport/rdma_gid_probe.h
@@ -29,9 +29,18 @@ namespace mooncake {
 enum class AutoGidCandidateClass {
     kNetworkRoutable = 0,
     kNoNetworkRoutable = 1,
-    kNetworkDegraded = 2,
-    kNoNetworkDegraded = 3,
-    kFallbackNonzero = 4,
+    // Private-range (RFC 1918 / CGNAT) IPv4-mapped GIDs. Split out of the
+    // degraded tier (#2729): datacenter RoCE fabrics routinely address NICs
+    // from 10/8, and such a GID is at worst same-subnet-usable — unlike a
+    // link-local fe80:: GID, which can never work across subnets. The
+    // relative order of all pre-existing tiers is unchanged; only the former
+    // tie between private-range IPv4 and link-local/overlay entries within
+    // "degraded" is split.
+    kNetworkPrivateV4 = 2,
+    kNetworkDegraded = 3,
+    kNoNetworkPrivateV4 = 4,
+    kNoNetworkDegraded = 5,
+    kFallbackNonzero = 6,
 };
 
 enum class AutoGidRetryAction {
@@ -72,8 +81,12 @@ inline const char* autoGidCandidateClassToString(
             return "network-routable";
         case AutoGidCandidateClass::kNoNetworkRoutable:
             return "no-network-routable";
+        case AutoGidCandidateClass::kNetworkPrivateV4:
+            return "network-private-v4";
         case AutoGidCandidateClass::kNetworkDegraded:
             return "network-degraded";
+        case AutoGidCandidateClass::kNoNetworkPrivateV4:
+            return "no-network-private-v4";
         case AutoGidCandidateClass::kNoNetworkDegraded:
             return "no-network-degraded";
         case AutoGidCandidateClass::kFallbackNonzero:
@@ -95,20 +108,26 @@ inline std::optional<AutoGidCandidateClass> classifyAutoGidCandidate(
     }
 
     const bool is_roce_v2 = candidate.gid_type == IBV_GID_TYPE_ROCE_V2;
-    const bool is_overlay =
-        is_roce_v2 && (candidate.is_overlay_network ||
-                       (candidate.is_ipv4_mapped && candidate.is_overlay_ipv4));
+    // Interface-name-based overlay detection (docker*/cni*/...) is the
+    // reliable demotion signal; a private address range alone is not — DC
+    // RoCE fabrics commonly use 10/8 (#2729).
+    const bool is_overlay_iface = is_roce_v2 && candidate.is_overlay_network;
+    const bool is_private_v4 = is_roce_v2 && !is_overlay_iface &&
+                               candidate.is_ipv4_mapped &&
+                               candidate.is_overlay_ipv4;
     const bool is_link_local =
         is_roce_v2 && !candidate.is_ipv4_mapped && candidate.is_link_local_ipv6;
-    const bool is_degraded = is_overlay || is_link_local;
+    const bool is_degraded = is_overlay_iface || is_link_local;
 
     if (candidate.has_network_device) {
-        return is_degraded ? AutoGidCandidateClass::kNetworkDegraded
-                           : AutoGidCandidateClass::kNetworkRoutable;
+        if (is_degraded) return AutoGidCandidateClass::kNetworkDegraded;
+        if (is_private_v4) return AutoGidCandidateClass::kNetworkPrivateV4;
+        return AutoGidCandidateClass::kNetworkRoutable;
     }
 
-    return is_degraded ? AutoGidCandidateClass::kNoNetworkDegraded
-                       : AutoGidCandidateClass::kNoNetworkRoutable;
+    if (is_degraded) return AutoGidCandidateClass::kNoNetworkDegraded;
+    if (is_private_v4) return AutoGidCandidateClass::kNoNetworkPrivateV4;
+    return AutoGidCandidateClass::kNoNetworkRoutable;
 }
 
 inline int autoGidCandidateClassPriority(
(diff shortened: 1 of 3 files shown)


stmatengss:
@qoder consider issue (https://github.com/kvcache-ai/Mooncake/issues/2729), review it.

stmatengss (approved):

PR #2746, opened 2026-07-04
[Store] object type eviction policy

kanceler:
## Description

This is split optimization version 4/4 of the original PR: https://github.com/kvcache-ai/Mooncake/pull/2689.

This PR builds on the object type metadata, type-aware eviction scoring, and per-type eviction statistics introduced by the previous three PRs. It adds the ability to configure memory eviction budgets by `ObjectDataType`. The new `ObjectTypeEvictionPolicy` allows each object type to specify a `budget_ratio`; during `BatchEvict`, Mooncake first selects candidates from object types that exceed their configured budget, then passes the remaining eviction target to the existing global adjusted-age eviction flow.

This is the fourth step split out from the original Hidden State type-aware eviction PR. This PR description focuses only on the object-type budget policy added on top of PR1, PR2, and PR3. The `HIDDEN_STATE` type, `ObjectTypeEvictionScorePolicy`, `adjusted_age`, and per-type eviction statistics included in this combined branch are covered by the previous PRs.

Design rationale:

- With only global LRU or global adjusted-age ranking, high-reuse-value object types can be disadvantaged by other types in some workloads, while a single object type can also consume too much memory and hurt the others.
- PR3 already collects per-`ObjectDataType` memory usage and candidate sets during eviction scanning. PR4 uses that information to add a per-type budget pass before falling back to global eviction.
- `budget_ratio` is expressed as a fraction of total memory capacity, making the same policy easier to reuse across different deployment sizes. The default value is `1.0`; when no explicit policy is configured, no object type is actively limited by budget.

Main changes:

- Add `ObjectTypeEvictionPolicy`, containing:
  - `budget_ratio`: the fraction of total memory capacity an object type is allowed to use. Default: `1.0`.
- Add `object_type_eviction_policies` config propagation through `MasterConfig`, `MasterServiceSupervisorConfig`, `WrappedMasterServiceConfig`, `MasterServiceConfig`, and `MasterServiceConfigBuilder`.
- Validate `budget_ratio >= 0.0` in the builder to reject meaningless negative budget policies.
- Expand object type budget policies into `UINT8_MAX + 1` indexed slots during `MasterService` initialization. Unconfigured object types keep the default `1.0` budget ratio.
- Add an object-type budget check before the regular global eviction stage in `BatchEvict`:

```text
type_budget_bytes = budget_ratio * total_mem_capacity
over_budget_ratio = type_used_bytes / total_mem_capacity - budget_ratio
type_evict_ratio = min(evict_ratio_target, over_budget_ratio)
type_evict_num = ceil(per_type_eviction_base[type] * type_evict_ratio)
```

- When an object type's `type_used_bytes` exceeds `type_budget_bytes`, candidates from that type's non-soft-pinned set are evicted first, prioritizing larger `adjusted_age`.
- After the type budget pass finishes, the regular first global eviction stage subtracts the number of objects already evicted from its target count. This avoids expanding a single eviction cycle by independently applying both the budget pass and the global pass.

Compatibility notes:

- The full branch contains the previous three PRs for correctness, but the new runtime behavior added by this PR is limited to the memory eviction policy inside `BatchEvict`: if an object type exceeds its configured budget, Mooncake first performs a type-aware pass over non-soft-pinned candidates from that type.
- This PR does not change the external meaning of `evict_ratio_target` or `evict_ratio_lowerbound` in `BatchEvict`. The function still tries to evict the target number of objects and at least attempts to satisfy the lower bound; this PR only changes internal candidate selection.
- This PR does not change the basic eligibility conditions for entering the eviction candidate set. Objects still need to satisfy the existing lease-expired, non-hard-pinned, and evictable memory replica conditions.
- The type budget pass only uses the non-soft-pinned candidates collected by PR3. Soft-pinned objects are still considered only in the later global second stage, and only when `allow_evict_soft_pinned_objects` is enabled.
- This PR does not change the existing fallback scan semantics of the tenant quota eviction path `EvictTenantMemoryForQuota`. That path is not based on an in-tenant LRU queue; it picks a random starting shard with `RandomIndex(kNumShards)` for the specified tenant and scans existing shard / metadata state to release eligible objects.
- The default `budget_ratio = 1.0` falls back to the original behavior.
- This PR does not change offload, promotion, or NoF eviction paths.
- The type budget selection reuses existing candidate sets and `std::nth_element` threshold selection. It does not introduce a full sort. The extra traversal is limited to `uint8_t` object type slots and candidate indices, so candidate selection/scanning remains in the same complexity class as the existing implementation.

Full eviction policy overview:

- The original LRU ranking value based on `lease_timeout` is adjusted by value-aware scoring:

```text
adjusted_age =
    max(0, rank_reference_time - lease_timeout - eviction_grace)
    * soft_pin_weight / reuse_scale
```

- `BatchEvict` first follows the existing parallel shard scan. It collects candidates that satisfy the basic eviction conditions, while also recording memory usage, evictable object counts, and candidate indices for each `ObjectDataType`.
- For object types exceeding `budget_ratio * total_mem_capacity`, the budget pass first evicts from that type's non-soft-pinned, expired candidates, prioritizing larger `adjusted_age`. The number of objects evicted is bounded by both `evict_ratio_target` and the type's over-budget ratio.
- After the type budget pass, the regular first global stage continues over all non-soft-pinned candidates by `adjusted_age` to fill the remaining target count, preserving the `BatchEvict` best-effort target semantics.
- If the first stage still does not satisfy `evict_ratio_lowerbound`, the existing second-stage behavior is preserved: it first supplements with non-soft-pinned candidates, and then considers soft-pinned candidates only when `allow_evict_soft_pinned_objects` is enabled.
- Offload-on-evict, force-evict, discarded replica release, metadata erase, and NoF eviction paths keep their existing flow. This PR only changes the internal selection order for memory eviction candidates.

Benchmark and experiment results:

The benchmark is designed to approximate realistic multimodal session behavior where KV Cache and Hidden State have different characteristics. It uses a continuous session timeline rather than separate warm / pressure / probe phases. The workload models different average reuse intervals for KV and Hidden State, different object sizes, different session round counts, image counts, image sizes, video ratios, public/private visual reuse, and session rebuilds. Since real cache-hit latency savings are difficult to measure exactly in a store-level benchmark, the report uses a simple modeled saving metric based on recomputation cost and size-based transfer cost.

Full benchmark code, run instructions, and the detailed experiment report are available at:

https://github.com/kanceler/mooncake-object-type-eviction-policy-benchmark

This run aggregates 3 repetitions, each covering 9 representative configurations.

| Case | Policy configuration | Hidden hit | vs baseline | KV hit | vs baseline | Saving Time Rate | Put Fail | Avg Evict |
|---|---|---:|---:|---:|---:|---:|---:|---:|
| baseline default | Default policy, no object-type reuse scale / budget | 111/327=33.94% | +0.00 pp | 495577/512742=96.65% | +0.00 pp | 87.80% | 0 | 5.11% |
| budget only kv80 h100 | `KVCACHE budget_ratio=0.80`, `HIDDEN_STATE budget_ratio=1.0` | 159/327=48.62% | +14.68 pp | 492260/512742=96.01% | -0.64 pp | 88.42% | 0 | 5.09% |
| grace only h20 | `HIDDEN_STATE eviction_grace=20000`, no reuse scale / budget | 222/327=67.89% | +33.95 pp | 491422/512742=95.84% | -0.81 pp | 92.42% | 0 | 5.08% |
| reuse scale only h8 | `HIDDEN_STATE reuse_scale=8.0` | 277/327=84.71% | +50.77 pp | 486987/512742=94.98% | -1.67 pp | 94.38% | 0 | 5.12% |
| reuse scale h8 g10 | `HIDDEN_STATE reuse_scale=8.0`, `HIDDEN_STATE eviction_grace=10000` | 298/327=91.13% | +57.19 pp | 477489/512742=93.12% | -3.52 pp | 93.33% | 0 | 5.09% |
| h8 + kv80 + h100 | `HIDDEN_STATE reuse_scale=8.0`, `KVCACHE budget_ratio=0.80`, `HIDDEN_STATE budget_ratio=1.0` | 281/327=85.93% | +51.99 pp | 483605/512742=94.32% | -2.33 pp | 93.47% | 0 | 5.12% |
| h8 + kv80 + h12 | `HIDDEN_STATE reuse_scale=8.0`, `KVCACHE budget_ratio=0.80`, `HIDDEN_STATE budget_ratio=0.12` | 248/327=75.84% | +41.90 pp | 490691/512742=95.70% | -0.95 pp | 93.47% | 0 | 5.10% |
| h8 + g10 + kv80 + h12 | `HIDDEN_STATE reuse_scale=8.0`, `HIDDEN_STATE eviction_grace=10000`, `KVCACHE budget_ratio=0.80`, `HIDDEN_STATE budget_ratio=0.12` | 247/327=75.54% | +41.60 pp | 490905/512742=95.74% | -0.91 pp | 94.26% | 0 | 5.07% |
| h8 + kv80 + h08 | `HIDDEN_STATE reuse_scale=8.0`, `KVCACHE budget_ratio=0.80`, `HIDDEN_STATE budget_ratio=0.08` | 195/327=59.63% | +25.69 pp | 491814/512742=95.92% | -0.73 pp | 90.94% | 1 | 5.10% |

Experiment summary:

- Type-aware reuse scale, `eviction_grace`, and per-type budget control different dimensions of the eviction policy: reuse scale normalizes adjusted age across object types, `eviction_grace` adds a type-level time bias, and per-type budget provides a capacity-side guardrail so one protected type does not grow without bounds at the expense of others.
- Looking only at Hidden object hit rate can overstate the gain, because Hidden State has a small unique-byte share in this workload. Global byte hit rate is also dominated by KV Cache. Therefore the table includes `Saving Time Rate` to provide an additional view based on modeled recomputation and transfer cost.
- More aggressive Hidden protection can further improve Hidden hit rate, but it also causes a larger KV hit-rate drop. With a budget guardrail, the policy moves closer to a stable tradeoff region where both KV reusable hit and Saving Time Rate remain high.
- Without a budget constraint, `eviction_grace` behaves like a stronger protection knob, which can be useful for highly valuable systematic data types. With `HIDDEN_STATE budget_ratio=0.12`, it does not significantly change object hit rate, but it slightly improves the cost-weighted saving metric. This makes it a fine-grained tuning knob on top of reuse scale and budget, rather than a replacement for either.
- From the perspective of `Saving Time Rate`, the goal is not to favor one cache type unconditionally or maximize one relative hit rate. The policy should give different object types a fair chance to compete under LRU-style pressure, while still allowing type-specific corrections based on data value and reuse characteristics.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Changes

- [ ] Bug fix
- [X] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

This PR mainly adds the object-type budget stage in `BatchEvict` while preserving the default behavior when no budget policy is configured. Recommended coverage includes existing master service tests and object-type policy scenarios:

- With the default `budget_ratio = 1.0`, eviction without explicit object-type budget configuration should preserve the default behavior of the first three PRs.
- For an object type configured with a smaller `budget_ratio`, when that type exceeds its budget, the budget pass should first evict from that type's non-soft-pinned, expired candidates by larger `adjusted_age`.
- Objects evicted by the type budget stage should be subtracted from the target count of the following global first stage.
- Soft-pinned objects should still participate only in the global second stage when `allow_evict_soft_pinned_objects` is enabled.
- Invalid `budget_ratio < 0.0` should be rejected by the builder.

Commands:

```bash
ninja -C build-docker mooncake-store/tests/master_service_test
./build-docker/mooncake-store/tests/master_service_test
ninja -C build-docker mooncake-store/tests/object_data_type_test
```

Additional static check:

```bash
git diff --check origin/main..origin/pr4-object-type-eviction-policy-combined
```

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used
- [X] AI tools were used

### DIFF
--- a/mooncake-store/include/master_config.h
+++ b/mooncake-store/include/master_config.h
@@ -19,6 +19,7 @@ class HttpMetadataServer;
 struct ObjectTypeEvictionScorePolicy {
     double reuse_scale{1.0};
     double soft_pin_weight{1.0};
+    int64_t eviction_grace{0};
 };
 
 inline void ValidateObjectTypeEvictionScorePolicy(
--- a/mooncake-store/include/master_config.h
+++ b/mooncake-store/include/master_config.h
@@ -22,6 +22,10 @@ struct ObjectTypeEvictionScorePolicy {
     int64_t eviction_grace{0};
 };
 
+struct ObjectTypeEvictionPolicy {
+    double budget_ratio{1.0};
+};
+
 inline void ValidateObjectTypeEvictionScorePolicy(
     const ObjectTypeEvictionScorePolicy& policy) {
     if (!std::isfinite(policy.reuse_scale) || policy.reuse_scale <= 0.0) {
@@ -640,6 +644,8 @@ class MasterServiceConfigBuilder {
     uint64_t default_kv_soft_pin_ttl_ = DEFAULT_KV_SOFT_PIN_TTL_MS;
     std::unordered_map<ObjectDataType, ObjectTypeEvictionScorePolicy>
         object_type_eviction_score_policies_;
+    std::unordered_map<ObjectDataType, ObjectTypeEvictionPolicy>
+        object_type_eviction_policies_;
     bool allow_evict_soft_pinned_objects_ =
         DEFAULT_ALLOW_EVICT_SOFT_PINNED_OBJECTS;
     double eviction_ratio_ = DEFAULT_EVICTION_RATIO;
@@ -710,6 +716,15 @@ class MasterServiceConfigBuilder {
         ObjectDataType data_type, ObjectTypeEvictionScorePolicy policy) {
         ValidateObjectTypeEvictionScorePolicy(policy);
         object_type_eviction_score_policies_[data_type] = policy;
+        return *this;
+    }
+
+    MasterServiceConfigBuilder& set_object_type_eviction_policy(
+        ObjectDataType data_type, ObjectTypeEvictionPolicy policy) {
+        if (policy.budget_ratio < 0.0) {
+            throw std::invalid_argument("budget_ratio must be non-negative");
+        }
+        object_type_eviction_policies_[data_type] = policy;
         return *this;
     }
 
@@ -987,6 +1002,8 @@ class MasterServiceConfig {
     uint64_t default_kv_soft_pin_ttl = DEFAULT_KV_SOFT_PIN_TTL_MS;
     std::unordered_map<ObjectDataType, ObjectTypeEvictionScorePolicy>
         object_type_eviction_score_policies;
+    std::unordered_map<ObjectDataType, ObjectTypeEvictionPolicy>
+        object_type_eviction_policies;
     bool allow_evict_soft_pinned_objects =
         DEFAULT_ALLOW_EVICT_SOFT_PINNED_OBJECTS;
     double eviction_ratio = DEFAULT_EVICTION_RATIO;
@@ -1135,6 +1152,7 @@ inline MasterServiceConfig MasterServiceConfigBuilder::build() const {
     config.default_kv_soft_pin_ttl = default_kv_soft_pin_ttl_;
     config.object_type_eviction_score_policies =
         object_type_eviction_score_policies_;
+    config.object_type_eviction_policies = object_type_eviction_policies_;
     config.allow_evict_soft_pinned_objects = allow_evict_soft_pinned_objects_;
     config.eviction_ratio = eviction_ratio_;
     config.eviction_high_watermark_ratio = eviction_high_watermark_ratio_;
--- a/mooncake-store/include/master_config.h
+++ b/mooncake-store/include/master_config.h
@@ -3,6 +3,7 @@
 #include <optional>
 #include <stdexcept>
 #include <string_view>
+#include <unordered_map>
 
 #include <glog/logging.h>
 
@@ -13,6 +14,11 @@ namespace mooncake {
 
 // Forwarded to the HA serve phase via MasterServiceSupervisorConfig.
 class HttpMetadataServer;
+
+struct ObjectTypeEvictionScorePolicy {
+    double reuse_scale{1.0};
+    double soft_pin_weight{1.0};
+};
 
 inline std::string ResolveConfiguredHABackendConnstring(
     std::string_view ha_backend_type, std::string_view ha_backend_connstring,
@@ -617,6 +623,8 @@ class MasterServiceConfigBuilder {
    private:
     uint64_t default_kv_lease_ttl_ = DEFAULT_DEFAULT_KV_LEASE_TTL;
     uint64_t default_kv_soft_pin_ttl_ = DEFAULT_KV_SOFT_PIN_TTL_MS;
+    std::unordered_map<ObjectDataType, ObjectTypeEvictionScorePolicy>
+        object_type_eviction_score_policies_;
     bool allow_evict_soft_pinned_objects_ =
         DEFAULT_ALLOW_EVICT_SOFT_PINNED_OBJECTS;
     double eviction_ratio_ = DEFAULT_EVICTION_RATIO;
@@ -680,6 +688,15 @@ class MasterServiceConfigBuilder {
 
     MasterServiceConfigBuilder& set_default_kv_soft_pin_ttl(uint64_t ttl) {
         default_kv_soft_pin_ttl_ = ttl;
+        return *this;
+    }
+
+    MasterServiceConfigBuilder& set_object_type_eviction_score_policy(
+        ObjectDataType data_type, ObjectTypeEvictionScorePolicy policy) {
+        if (policy.reuse_scale <= 0.0) {
+            throw std::invalid_argument("reuse_scale must be greater than 0");
+        }
+        object_type_eviction_score_policies_[data_type] = policy;
         return *this;
     }
 
@@ -955,6 +972,8 @@ class MasterServiceConfig {
    public:
     uint64_t default_kv_lease_ttl = DEFAULT_DEFAULT_KV_LEASE_TTL;
     uint64_t default_kv_soft_pin_ttl = DEFAULT_KV_SOFT_PIN_TTL_MS;
+    std::unordered_map<ObjectDataType, ObjectTypeEvictionScorePolicy>
+        object_type_eviction_score_policies;
     bool allow_evict_soft_pinned_objects =
         DEFAULT_ALLOW_EVICT_SOFT_PINNED_OBJECTS;
     double eviction_ratio = DEFAULT_EVICTION_RATIO;
@@ -1101,6 +1120,8 @@ inline MasterServiceConfig MasterServiceConfigBuilder::build() const {
     MasterServiceConfig config;
     config.default_kv_lease_ttl = default_kv_lease_ttl_;
     config.default_kv_soft_pin_ttl = default_kv_soft_pin_ttl_;
+    config.object_type_eviction_score_policies =
+        object_type_eviction_score_policies_;
     config.allow_evict_soft_pinned_objects = allow_evict_soft_pinned_objects_;
     config.eviction_ratio = eviction_ratio_;
     config.eviction_high_watermark_ratio = eviction_high_watermark_ratio_;
--- a/mooncake-store/include/master_config.h
+++ b/mooncake-store/include/master_config.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <cmath>
 #include <optional>
 #include <stdexcept>
 #include <string_view>
@@ -19,6 +20,19 @@ struct ObjectTypeEvictionScorePolicy {
     double reuse_scale{1.0};
     double soft_pin_weight{1.0};
 };
+
+inline void ValidateObjectTypeEvictionScorePolicy(
+    const ObjectTypeEvictionScorePolicy& policy) {
+    if (!std::isfinite(policy.reuse_scale) || policy.reuse_scale <= 0.0) {
+        throw std::invalid_argument(
+            "reuse_scale must be finite and greater than 0");
+    }
+    if (!std::isfinite(policy.soft_pin_weight) ||
+        policy.soft_pin_weight < 0.0) {
+        throw std::invalid_argument(
+            "soft_pin_weight must be finite and non-negative");
+    }
+}
 
 inline std::string ResolveConfiguredHABackendConnstring(
     std::string_view ha_backend_type, std::string_view ha_backend_connstring,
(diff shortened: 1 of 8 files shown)


mooncake-store/include/master_config.h
```diff
@@ -14,6 +16,38 @@ namespace mooncake {
...
+
+struct ObjectTypeEvictionPolicy {
+    double budget_ratio{1.0};
+};
+
+inline void ValidateObjectTypeEvictionPolicy(
+    const ObjectTypeEvictionPolicy& policy) {
+    if (!std::isfinite(policy.budget_ratio) || policy.budget_ratio < 0.0 ||
+        policy.budget_ratio > 1.0) {
+        throw std::invalid_argument(
```
Aionw, line 33: Using throw here feels a bit awkward, since it forces every call site to wrap the call in try/catch to avoid crashing the process.

We could instead return a bool to indicate whether validation succeeded. If a boolean alone doesn’t provide enough context about what went wrong, we can simply add a warning-level log here.

mooncake-store/include/master_config.h
```diff
@@ -14,6 +16,38 @@ namespace mooncake {
...
+            "budget_ratio must be finite and between 0.0 and 1.0");
+    }
+}
+
+inline void ValidateObjectTypeEvictionScorePolicy(
+    const ObjectTypeEvictionScorePolicy& policy) {
+    if (!std::isfinite(policy.reuse_scale) || policy.reuse_scale <= 0.0) {
+        throw std::invalid_argument(
```
Aionw, line 41: Same as above.

mooncake-store/include/master_service.h
```diff
@@ -1536,6 +1534,10 @@ class MasterService {
     const uint64_t default_kv_lease_ttl_;     // in milliseconds
     const uint64_t default_kv_soft_pin_ttl_;  // in milliseconds
     const bool allow_evict_soft_pinned_objects_;
+    std::array<double, UINT8_MAX + 1> object_type_reuse_scales_;
```
Aionw, line 1537: Here object type only has 11 elements, we do not need UINT8_MAX here. 

A common pattern is add a static value `static constexpr size_t object_type_count = static_cast<size_t>(ObjectTypeId::HIDDEN_STATE) + 1`, and direct use that value here.

mooncake-store/src/master.cpp
```diff
@@ -427,6 +437,63 @@ void InitMasterConf(const mooncake::DefaultConfig& default_config,
     default_config.GetBool("allow_evict_soft_pinned_objects",
                            &master_config.allow_evict_soft_pinned_objects,
                            FLAGS_allow_evict_soft_pinned_objects);
+    for (uint16_t type_id = 0; type_id <= UINT8_MAX; ++type_id) {
```
Aionw, line 440: Same as above, we have limited data type and do not need to scan from 0 to uint_max

Aionw (commented):
I’m glad to see that we finally have a data-type-based eviction policy.

That said, the current implementation is tightly coupled to HIDDEN_STATE and introduces both the eviction algorithm and data-type-specific eviction management directly into master_service. I’m concerned that as we add more eviction policies in the future, master_service will become increasingly complex.

I’d suggest introducing an abstraction for per-data-type eviction policies first, and then implementing the eviction algorithm from this PR on top of that abstraction. I’d be happy to draft an RFC to describe the design in more detail, and I’d really appreciate any comments or suggestions you may have.

kanceler:
> I’m glad to see that we finally have a data-type-based eviction policy.
> 
> That said, the current implementation is tightly coupled to HIDDEN_STATE and introduces both the eviction algorithm and data-type-specific eviction management directly into master_service. I’m concerned that as we add more eviction policies in the future, master_service will become increasingly complex.
> 
> I’d suggest introducing an abstraction for per-data-type eviction policies first, and then implementing the eviction algorithm from this PR on top of that abstraction. I’d be happy to draft an RFC to describe the design in more detail, and I’d really appreciate any comments or suggestions you may have.

This makes sense to me.

This PR was initially motivated by the CCF contest task around the difference between Hidden State and KV Cache caching granularity and eviction behavior. Because of that, I focused more on validating the data-type-aware eviction behavior for that concrete scenario, and kept the implementation close to the existing BatchEvict / MasterService flow. Looking back, I agree that this makes the policy logic too coupled with HIDDEN_STATE and leaves too much data-type-specific logic inside MasterService.

I also agree that a per-data-type eviction policy abstraction would be a better long-term direction. MasterService should probably keep the common responsibilities, such as metadata scanning, eligibility checks, replica release, and metadata cleanup, while the policy layer handles scoring, per-type accounting, budget enforcement, and victim selection.

If you draft the RFC, I’d be happy to provide comments and summarize the issues I ran into while working on this PR.

If useful, I can also prepare some initial notes based on this PR to help with the RFC discussion.

kanceler:
> I’m glad to see that we finally have a data-type-based eviction policy.
> 
> That said, the current implementation is tightly coupled to HIDDEN_STATE and introduces both the eviction algorithm and data-type-specific eviction management directly into master_service. I’m concerned that as we add more eviction policies in the future, master_service will become increasingly complex.
> 
> I’d suggest introducing an abstraction for per-data-type eviction policies first, and then implementing the eviction algorithm from this PR on top of that abstraction. I’d be happy to draft an RFC to describe the design in more detail, and I’d really appreciate any comments or suggestions you may have.

One thought I had is that eviction should not only focus on how to rank individual objects, but also on how to assign responsibility for a global memory pressure event.

When memory eviction is triggered, MasterService could first decide how much responsibility each data type should take for this eviction round, and then let that data type run its own eviction policy. For example:

1. At the object data type level, MasterService can decide how much responsibility each data type should take for the current high-watermark pressure, similar to the budget_ratio mechanism in this PR. Then each data type can run its own internal eviction policy, with an interface similar to BatchEvict: try to evict type_evict_ratio of eligible objects, and at least make a best effort toward type_evict_ratio_lowerbound.

2. If type-scoped eviction is still not enough, MasterService can fall back to a global eviction pass. However, this global fallback should still use a cross-type fair scoring mechanism rather than raw lease_time; otherwise the fallback may become systematically biased toward some data types.

3. The same issue also exists at the tenant level. In a multi-tenant setup, purely global victim selection can be unfair. For example, tenant A may have a 10GB quota and only use 2GB, while tenant B has a 90GB quota and uses 89GB. If the cluster crosses the high watermark, global eviction may evict tenant A’s old or low-reuse objects first, even though tenant A is not the main contributor to the memory pressure. From tenant A’s perspective, it becomes the main victim without being the main cause.

So maybe the abstraction could support a hierarchy like this:

global memory pressure
  -> assign eviction responsibility to tenants
  -> within each tenant, assign responsibility to object data types
  -> within each type, run the corresponding eviction policy
  -> if scoped eviction cannot satisfy the target, run a fair global fallback

Aionw:
A hierarchical eviction policy abstraction is what we need. However, the current eviction logic in master_service is tightly coupled, which makes changes like this difficult and prone to introducing regressions.

PR #2748, opened 2026-07-04
fix(rdma): rank link-local GID below any network-backed IPv4 in auto-GID

### LINKED ISSUE #2747: Auto GID selection picks a non-routable link-local GID on routed RoCEv2 fabrics using RFC1918 (10.0.0.0/8)
### Summary
On a routed RoCEv2 fabric where the RoCE IPv4 addresses live in RFC1918 space
(e.g. 10.0.0.0/8), automatic GID selection (when no explicit GID index is set)
picks a non-routable link-local GID instead of the routable IPv4 GID. Cross-node
QPs then never reach RTR and every cross-node transfer fails.

### Environment
- Transport: RDMA (RoCEv2), auto-GID (no `MC_GID_INDEX` / `gid_index < 0`)
- Fabric: routed RoCEv2, RoCE NIC IPv4 in 10.102.0.0/16 (i.e. inside 10.0.0.0/8)

### GID table (per rdma0, representative)
| index | type    | GID                       | address      |
|-------|---------|---------------------------|--------------|
| 0     | RoCE v2 | fe80::0690:81ff:fe37:a9e8 | link-local   |
| 1     | RoCE v2 | ::ffff:0a66:0965          | 10.102.9.101 |

### Root cause
`isOverlayIPv4` flags 10.0.0.0/8 (and 172.16/12, 100.64/10) as an "overlay"
address, so the routable IPv4 GID (index 1) is classified as
`kNetworkDegraded`. A network-backed link-local GID (index 0) is also classified
as `kNetworkDegraded`. With both candidates in the same class, the stable
tie-break on ascending `gid_index` selects index 0 — the link-local GID. A
`fe80::` GID cannot route across subnets, so the QP never reaches RTR:

    Failed to modify QP to RTR ... timed out [110]

### Impact
Any routed RoCEv2 deployment using RFC1918 10/8 (etc.) addressing must currently
work around this by pinning `MC_GID_INDEX` manually. A flat-L2 fabric that
exposes a normal (non-overlay) routable IPv4 GID is unaffected.

### Proposed fix
Rank a network-backed link-local GID strictly below every network-backed IPv4
GID (routable and degraded/overlay alike), so link-local is only chosen when no
network-backed IPv4 GID exists. PR to follow.

Lzy17:
## Problem
Fixes #2747. On routed RoCEv2 fabrics using RFC1918 addressing (e.g.
10.0.0.0/8), auto-GID selection picks a non-routable link-local GID over the
routable IPv4 GID, so cross-node QPs never reach RTR and transfers fail.

The routable IPv4 GID is flagged as an overlay by `isOverlayIPv4` (10/8,
172.16/12, 100.64/10) and lands in `kNetworkDegraded`; a network-backed
link-local GID also landed in `kNetworkDegraded`, so the stable tie-break on
ascending `gid_index` chose the link-local GID.

## Fix
Introduce a dedicated `kNetworkLinkLocal` class ranked strictly below every
network-backed IPv4 class, so a link-local GID is chosen only when no
network-backed IPv4 GID exists. Wired into the priority table, the string
helper, and `autoGidStateFromSelection` (still `GID_WITH_NETWORK`).

Only the auto-GID path (`gid_index < 0`) is affected; explicit-index
deployments are unchanged.

## Testing
- Unit: existing `rdma_gid_probe_test.cpp` cases still pass; added an ionic
  regression (network-backed link-local vs network-backed overlay-IPv4 → IPv4
  wins) plus a reversed-order variant.
- HW `transfer_engine_bench`, auto-GID (no explicit index), two fabrics:
  - routed RoCEv2 (10.102.x.x): **before** → selects link-local GID[0], transfer
    fails at RTR; **after** → selects routable GID[1] (10.102.x.x), completes
    (~48 GB/s).
  - flat-L2 (exposes a normal routable IPv4 GID): selection unchanged, no
    regression (~41 GB/s before and after).

### DIFF
--- a/mooncake-transfer-engine/include/transport/rdma_transport/rdma_gid_probe.h
+++ b/mooncake-transfer-engine/include/transport/rdma_transport/rdma_gid_probe.h
@@ -30,8 +30,13 @@ enum class AutoGidCandidateClass {
     kNetworkRoutable = 0,
     kNoNetworkRoutable = 1,
     kNetworkDegraded = 2,
-    kNoNetworkDegraded = 3,
-    kFallbackNonzero = 4,
+    // A network-backed link-local GID is not routable across subnets, so it
+    // must rank strictly below any network-backed IPv4 GID (even a degraded /
+    // RFC1918 one). On routed RoCEv2 fabrics the only usable cross-node GID is
+    // the IPv4 one; picking link-local here breaks the RTR handshake.
+    kNetworkLinkLocal = 3,
+    kNoNetworkDegraded = 4,
+    kFallbackNonzero = 5,
 };
 
 enum class AutoGidRetryAction {
@@ -74,6 +79,8 @@ inline const char* autoGidCandidateClassToString(
             return "no-network-routable";
         case AutoGidCandidateClass::kNetworkDegraded:
             return "network-degraded";
+        case AutoGidCandidateClass::kNetworkLinkLocal:
+            return "network-link-local";
         case AutoGidCandidateClass::kNoNetworkDegraded:
             return "no-network-degraded";
         case AutoGidCandidateClass::kFallbackNonzero:
@@ -103,8 +110,14 @@ inline std::optional<AutoGidCandidateClass> classifyAutoGidCandidate(
     const bool is_degraded = is_overlay || is_link_local;
 
     if (candidate.has_network_device) {
-        return is_degraded ? AutoGidCandidateClass::kNetworkDegraded
-                           : AutoGidCandidateClass::kNetworkRoutable;
+        // A non-routable link-local GID must never outrank a network-backed
+        // IPv4 GID, even one flagged as degraded/overlay (e.g. an RFC1918
+        // 10.0.0.0/8 address that a routed RoCEv2 fabric legitimately uses).
+        if (is_link_local) {
+            return AutoGidCandidateClass::kNetworkLinkLocal;
+        }
+        return is_overlay ? AutoGidCandidateClass::kNetworkDegraded
+                          : AutoGidCandidateClass::kNetworkRoutable;
     }
 
     return is_degraded ? AutoGidCandidateClass::kNoNetworkDegraded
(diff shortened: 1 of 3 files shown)


alogfans:
Looks good to me, but before merging please resolve conflicts and perform CI checks.

mzygQAQ:
i also meet this issues,  i see that another one propose a same issue and pr, and was merged, see https://github.com/kvcache-ai/Mooncake/pull/2741

n-WN:
Heads-up on an overlap (as @mzygQAQ noted): #2741 merged on Jul 5 (`275e84c`) fixes the same root cause from the other direction — instead of demoting link-local into a new bottom tier, it promotes private-range IPv4-mapped GIDs (10/8, 172.16/12, 100.64/10) into a `kNetworkPrivateV4` tier that ranks strictly above link-local/overlay. 

Verified on current main (`f1f5e40`) on an mlx5 RoCEv2 node with the GID-table shape from #2747 (link-local entries at indices 0/1, IPv4-mapped private-range entries at 2/3): auto-GID now selects the v4-mapped RoCEv2 entry (`Find best gid index: 3`), and cross-NIC `transfer_engine_bench` runs at full rate. So the scenario this PR targets should already be resolved on main — @Lzy17 it'd be great if you could confirm on your routed private fabric, since yours is a cross-subnet topology (mine is same-L2, where the mis-selection reproduced but the fe80 choice happened to still connect).

There *is* one residual case where the two designs differ, which may be worth keeping this PR (rebased and reduced) for: **an overlay-named interface's IPv4 GID (e.g. `docker0`) vs. a physical NIC's link-local GID**. #2741 leaves them tied in the degraded tier (lowest index wins); this PR's `kNetworkLinkLocal` ranks *any* network-backed IPv4 — including `docker0` — above link-local. I'd argue the opposite ordering is safer for that corner (a bridge interface's GID can never carry cross-node RDMA, while fe80 at least works same-L2), but if there are fabrics where the docker-named heuristic misfires on legitimate NICs, that's the discussion worth having — ideally with a table-driven case in `rdma_gid_probe_test.cpp` either way.

alogfans:
Could you please confirm that #2741 has solved your problem? Thank you.

Lzy17:
@alogfans Confirmed #2741 fixes it.  Closing this as a duplicate. Thanks!

PR #2759, opened 2026-07-06
[TENT] Per-pool QP allocation with per-pool SL/TC (RFC #2568 step 2)

catyans:
### What

Step 1 (#2640) added the `SelectionPolicy.qp_pool` schema — parsed and carried on `SelectionResult`, but not yet wired to anything. This step makes named QP pools real at the RDMA link layer: each pool gets its own contiguous run of data QPs inside every endpoint, handshaked with **that pool's** Service Level / Traffic Class.

This is the mechanism @alogfans described on the RFC:
> if we expect to configure different traffic_classes for different tasks within the same transfer engine, we need to allocate multiple QPs, each assign different parameters during their handshakes.

### How

- **`EndPointParams::qp_pools`** — optional per-pool layout `{name, num_qp, service_level, traffic_class}`. Empty (the default) = today's single homogeneous run of `qp_mul_factor` data QPs.
- **`computeQpPoolSegments()`** — a pure, free-standing resolver (no RDMA handles) that turns the pool list into concrete `[begin, begin+num_qp)` segments and a total QP count. Unit-tested.
- **`construct()`** — total data-QP count = sum of per-pool `num_qp` when pools are set; `qp_pool_segments_` records each pool's span.
- **`setupOneQP()`** — a QP that belongs to a pool overriding SL/TC handshakes with the pool's values; otherwise it falls back to the global endpoint SL/TC (unchanged default).
- **`rdma_transport`** — parses `transports/rdma/endpoint/qp_pools` from config. Pool SL/TC live in the RDMA config (a link-layer concept); `SelectionPolicy` only *references* a pool by name.

### Wire compatibility

`BootstrapDesc` is unchanged. Both peers derive the same pool layout from the same config, so the flat `qp_num` list still pairs positionally — pool P's i-th QP on one side lines up with pool P's i-th QP on the other. A peer that doesn't configure pools simply runs a single default pool.

### Default path is unchanged

With no `qp_pools` configured, `computeQpPoolSegments` returns `total_qp = qp_mul_factor` and no segments, `poolForQp()` returns `nullptr`, and every QP handshakes with the global SL/TC — byte-for-byte the prior behavior.

### Tests

`tent_qp_pool_layout_test` (4 cases): default flat layout, invalid (non-positive) config, multi-pool contiguous segments with per-pool SL/TC, and full index partitioning (mirrors `poolForQp`). Built and passing locally on an H20 / CX-7 box; the `tent` library also compiles clean with these changes.

Also fixes a latent issue surfaced by the new standalone test: `params.h` used `ibv_mtu` without including `<infiniband/verbs.h>` (it had been compiling only via transitive includes).

### Follow-up

Slice-to-pool **routing** (feeding `SelectionResult.qp_pool` into QP selection so traffic actually lands on its pool's QPs) is intentionally a separate PR, to keep this one reviewable in small steps. This PR builds the pools and gives them their link-layer identity; the next wires transfers onto them.

### DIFF
--- a/mooncake-transfer-engine/tent/include/tent/transport/rdma/params.h
+++ b/mooncake-transfer-engine/tent/include/tent/transport/rdma/params.h
@@ -15,8 +15,12 @@
 #ifndef TENT_PARAMS_H
 #define TENT_PARAMS_H
 
+#include <infiniband/verbs.h>
+
 #include <cstddef>
 #include <cstdint>
+#include <string>
+#include <vector>
 
 namespace mooncake {
 namespace tent {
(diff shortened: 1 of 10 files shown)


mooncake-transfer-engine/tent/include/tent/transport/rdma/params.h
```diff
@@ -58,6 +86,36 @@ struct EndPointParams {
...
+        for (const auto& pool : pools) {
+            QpPoolSegment seg = pool;
+            seg.begin = layout.total_qp;
+            layout.segments.push_back(seg);
+            layout.total_qp += pool.num_qp;
+        }
+    }
+    layout.valid = layout.total_qp > 0;
+    return layout;
+}
```
alogfans, line 127: Check `num_qp > 0`

catyans: Done — added the check. `computeQpPoolSegments` now rejects the whole layout (`valid=false`, so the caller falls back to the default single-pool behavior) if any pool has `num_qp <= 0`, since that would otherwise create an empty/negative `[begin, begin+num_qp)` span and break the router. Added zero- and negative-num_qp unit tests; the layout suite passes 12/12. Pushed.

alogfans (approved):
LGTM

Issue #2758, opened 2026-07-06
[Bug]: TENT cross-node rail attempts every remote device (ignores device_name) and connects to devices with no matching QP → RTR fails, throughput collapses


catyans:
## Bug Report

Follow-up to #2467 (same `RailMonitor` cross-NUMA mapping issue). This report adds:

1. an **errno-level diagnosis** (via a rebuilt bench) showing the visible `modify QP to RTR: EINVAL(22)` is *not* a QP-attribute problem but a **peer-QP-does-not-exist** problem — matching #2467's "maps to the wrong/overlay NIC" root cause;
2. an independent usability gap: **`--device_name` cannot restrict the rail's device set**, so an operator cannot steer around the bad devices;
3. a **five-way perftest control** that cleanly isolates the failure to TENT rather than the fabric.

#2467 already found the `updateBestMapping()` sequential-vs-`direct_rails_` root cause and proposed a fix; the intent here is to add hard evidence on different hardware (4× bonded NIC, dual-NUMA) and flag the `device_name` gap.

### Environment

| Item | Value |
|------|-------|
| Nodes | 2, cross-node RoCEv2 |
| GPU/NIC | 8× H20-3e, ConnectX-7 200G RoCE v2 |
| RDMA devices | `mlx5_0` (head VF, NUMA0), `mlx5_bond_0/1` (NUMA0), `mlx5_bond_2/3` (NUMA1) |
| GID | `MC_GID_INDEX=3` (RoCE v2), explicitly set |
| Backend | `transfer_engine_bench --backend=tent --metadata_server=P2PHANDSHAKE` |
| Commit | reproduced on a June checkout; paths verified identical on `main` @ d4a1f39 |

### What happens

With `--device_name=mlx5_bond_0` on **both** ends, the initiator still tries to connect to the target's `mlx5_bond_2/3` (NUMA1) and the overlay `mlx5_0`, and each fails:

```
Failed to modify QP's state to RTR in endpoint mlx5_bond_2 of 192.168.0.42:<port>,
check peer endpoint reachability, MTU 3, GID fd:03:45:14:00:80:62:42:...:01 and LID 0,
error code 22:Invalid argument
```

[...]

Issue #2516, opened 2026-06-17
[Feature Request]: Topology- and load-aware remote replica selection in SelectBestReplica


catyans:
### Describe your feature request

**Problem**

When the master returns multiple remote MEMORY replicas for the same key, `SelectBestReplica` in `mooncake-store/src/real_client.cpp` (lines 290-326) picks the **first** one it encounters rather than the closest/least-loaded one:

```cpp
// real_client.cpp:293-314
const Replica::Descriptor *first_memory = nullptr;
const Replica::Descriptor *first_nof = nullptr;
for (const auto &r : replicas) {
    if (r.status != ReplicaStatus::COMPLETE) continue;
    if (r.is_memory_replica()) {
        if (local_endpoints.count(...)) return &r;  // local MEMORY — best case
        if (!first_memory) first_memory = &r;       // <-- blindly keeps the FIRST remote one
    }
    ...
}
if (first_memory) return first_memory;              // <-- returned without comparing peers
```

The current selection logic is purely **type + locality** based (local MEMORY > local NOF > first remote MEMORY > first remote NOF > LOCAL_DISK > DISK). Once locality fails, all remote MEMORY replicas are treated as equivalent, so the choice among them is effectively arbitrary (master return order).

This ignores two signals that the transfer layer already tracks per local NIC:

1. **NUMA / PCIe topology distance** — which local NIC reaches a given remote endpoint, and whether that NIC is same-NUMA or cross-NUMA relative to the destination GPU buffer.
2. **Real-time NIC load** — in-flight bytes and EWMA bandwidth per NIC.

On our 8x H20 nodes, cross-NUMA NIC selection shows up to ~2.4x tail-latency inflation versus same-NUMA, so picking an arbitrary remote replica can route a transfer through a congested or cross-NUMA path even when a better peer exists.

**Motivation**

The information needed to make a better choice is **already computed** inside TENT's `DeviceSelector` (`mooncake-transfer-engine/tent/src/transport/rdma/quota.cpp`). Its existing NIC-selection score is:

```cpp
// quota.cpp:129-138
// Score formula: predicted_time × numa_penalty + random_jitter
double predicted_time = (inflight_bytes + slice_bytes) / ewma_bandwidth;
double rank_penalty   = numa_tier_weights[rank];   // [1.0, 5.0, 10.0] = same-NUMA / cross-NUMA / cross-socket
double score          = predicted_time * rank_penalty;
```

[...]

Issue #2467, opened 2026-06-15
Bug: RailMonitor cross-NUMA mapping uses sequential assignment instead of direct_rails_ matching


zhangzhiqiangcs:
## Bug Description

In a dual-NUMA RoCEv2 environment, `RailMonitor::updateBestMapping()` incorrectly maps cross-NUMA devices using sequential assignment (`remote_devices[remote_numa][i % remote_cnt]`) instead of `direct_rails_` same-name matching. This causes initiator's mlx5_5 (NUMA 1) to connect to target's mlx5_0 (NUMA 0, overlay) instead of the expected mlx5_5 (NUMA 1, physical network), resulting in `transport retry counter exceeded`.

## Environment

| Item | Value |
|------|-------|
| Initiator | 10.0.2.114 |
| Target | 10.0.48.39 |
| NIC Count | 9 per node (mlx5_0~8) |
| NUMA Topology | mlx5_0~4 on NUMA 0, mlx5_5~8 on NUMA 1 |
| NIC Model | Mellanox ConnectX-6 (mlx5_1~8), ConnectX-7 (mlx5_0, overlay) |
| Subnet | Same-named NICs across /28 subnets |

## Root Cause

In `rail_monitor.cpp`, `loadDefault()` correctly builds `direct_rails_` with same-name matching (Priority 1):

```cpp
// loadDefault() — correct
direct_rails_[5] = 5;  // mlx5_5 → mlx5_5 ✓
```

However, `updateBestMapping()` only uses `direct_rails_` for same-NUMA mapping, and falls back to sequential assignment for cross-NUMA:

```cpp
// updateBestMapping() — BUG (line 326-329)
if (local_numa == remote_numa)
    remote_nic = direct_rails_[local_nic];     // Same NUMA: same-name matching ✓
else
    remote_nic = remote_devices[remote_numa][i % remote_cnt];  // Cross NUMA: sequential ✗
```

### Concrete Failure Path

1. Initiator's slice memory on NUMA 1 → `source_dev_id=5` (mlx5_5)
2. Target's memory on NUMA 0 → `findBestRemoteDevice(5, remote_numa=0)`
3. Lookup `best_mapping_[0][5]`:
   - `local_numa=1, remote_numa=0` (cross-NUMA)
   - `local_devices[1] = [5,6,7,8]` (mlx5_5~8)
   - `remote_devices[0] = [0,1,2,3,4]` (mlx5_0~4)
[...]

PR #2760, opened 2026-07-06
[TENT] Make RDMA NIC allow/deny list configurable via MC_FILTER_NIC(_EXCLUDE)

catyans:
## What

Add environment variables to configure the RDMA NIC allow/deny list:

- `MC_FILTER_NIC` → `topology/rdma_whitelist` (allow-list; only these NICs are used, takes precedence)
- `MC_FILTER_NIC_EXCLUDE` → `topology/rdma_blacklist` (deny-list)

Values are comma-separated device names, e.g. `MC_FILTER_NIC=mlx5_bond_0,mlx5_bond_1`. Unset = discover all NICs (unchanged default).

## Why

The platform probes already support NIC allow/deny lists via the config keys `topology/rdma_whitelist` / `topology/rdma_blacklist` (see `filterInfiniBandDevices`), but **there was no env var wired to them**, so users had no way to restrict which RDMA NICs the engine discovers and uses.

On multi-NIC / multi-NUMA hosts this matters. The rail selection enumerates every discovered remote device and will attempt QPs to devices that are not reachable from the peer; those fail at `modify QP to RTR` and drag aggregate throughput down. See #2758 (and the related cross-NUMA mapping issue #2467). Giving operators a way to scope the engine to the reachable NICs is a simple, opt-in mitigation that works today, independent of the deeper rail-mapping fix.

## Changes

- `setArrayConfig()`: parse a comma-separated env value into a string array (trims whitespace, drops empty items), mirroring `parseDoubleArray`.
- Wire `MC_FILTER_NIC` / `MC_FILTER_NIC_EXCLUDE` in `loadFromEnv`.
- Unit tests: both mappings, whitespace trimming, and the unset default.
- Env var docs.

## Measured effect

8× H20 / CX-7 200G RoCEv2, 4-bond dual-NUMA, 2 nodes, TENT backend:

| | without filter | `MC_FILTER_NIC=mlx5_bond_0` |
|---|---|---|
| discovered topology | 5 devices (incl. unreachable NUMA1 bonds) | 1 device (mlx5_bond_0) |
| `modify QP to RTR error 22` | repeated on bond_2/3 | none |
| cross-node throughput | 0.14 GB/s | 48 GB/s |

## Notes

Opt-in; default behavior unchanged. Sits on the same TENT device-selection line as #2516 / #2568.

### DIFF
--- a/docs/source/design/transfer-engine/index.md
+++ b/docs/source/design/transfer-engine/index.md
@@ -471,6 +471,8 @@ For advanced users, TransferEngine provides the following advanced runtime optio
 - `MC_WORKERS_PER_CTX` The number of asynchronous worker threads corresponding to each device instance
 - `MC_SLICE_SIZE` The segmentation granularity of user requests in Transfer Engine
 - `MC_RETRY_CNT` The maximum number of retries in Transfer Engine
+- `MC_TE_FILTERS` Restrict which RDMA NICs the engine discovers and uses, as a comma-separated allow-list of device names (e.g. `mlx5_bond_0,mlx5_bond_1`). Only the listed NICs are kept; all others are ignored. Unset (default) discovers all NICs. This is the **same env var and semantics as the legacy Transfer Engine's device whitelist** (see below), so a single variable scopes NICs across both engines. Useful on multi-NIC / multi-NUMA hosts to keep the engine (and its rail selection) off NICs that are not routable to the peer.
+- `MC_TE_FILTERS_EXCLUDE` The deny-list counterpart of `MC_TE_FILTERS`: a comma-separated list of device names to exclude from discovery. Ignored if `MC_TE_FILTERS` is set (allow-list takes precedence). Unset (default) excludes nothing. (New; the legacy engine has an allow-list only.)
 - `MC_AUTO_GID_MAX_RETRIES` The maximum number of automatic local GID reprobe retries during classic RDMA handshake recovery. Default value 2. Set to 0 to disable automatic GID retry.
 - `MC_LOG_LEVEL` This option can be set as `TRACE`/`INFO`/`WARNING`/`ERROR` (see [glog doc](https://github.com/google/glog/blob/master/docs/logging.md)), and more detailed logs will be output during runtime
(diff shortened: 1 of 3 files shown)


docs/source/design/transfer-engine/index.md
```diff
@@ -471,6 +471,8 @@ For advanced users, TransferEngine provides the following advanced runtime optio
 - `MC_WORKERS_PER_CTX` The number of asynchronous worker threads corresponding to each device instance
 - `MC_SLICE_SIZE` The segmentation granularity of user requests in Transfer Engine
 - `MC_RETRY_CNT` The maximum number of retries in Transfer Engine
+- `MC_FILTER_NIC` Restrict which RDMA NICs the engine discovers and uses, as a comma-separated allow-list of device names (e.g. `mlx5_bond_0,mlx5_bond_1`). Only the listed NICs are kept; all others are ignored. Unset (default) discovers all NICs. Useful on multi-NIC / multi-NUMA hosts to keep the engine (and its rail selection) off NICs that are not routable to the peer.
+- `MC_FILTER_NIC_EXCLUDE` The deny-list counterpart of `MC_FILTER_NIC`: a comma-separated list of device names to exclude from discovery. Ignored if `MC_FILTER_NIC` is set (allow-list takes precedence). Unset (default) excludes nothing.
```
alogfans: Does these env vars already existed in Mooncake TE rather than TENT?

catyans: Good catch — yes. The legacy Transfer Engine already exposes a device whitelist via `MC_TE_FILTERS` (`src/topology.cpp` `getIbvDeviceWhitelist()`), but the TENT path had the filter (`filterInfiniBandDevices` reading `topology/rdma_whitelist`) with **no env wired to it**. To avoid a second, divergent name I've switched this PR to reuse the **same `MC_TE_FILTERS`** env (identical comma-separated allow-list semantics), so one variable scopes NICs across both engines. `MC_TE_FILTERS_EXCLUDE` is the only addition — a deny-list the legacy engine doesn't have. Pushed; docs updated to say so.

alogfans (approved):
LGTM

PR #2761, opened 2026-07-06
[TENT] Route slices to their QP pool (RFC #2568 step 3)

catyans:
> **Stacked on #2759** (RFC #2568 step 2). This PR's own change is the last commit, `[TENT] Route slices to their QP pool`. The diff will show step 2's changes too until #2759 merges, after which it narrows to just the routing commit. Please review the final commit here; the pool *allocation* it builds on is #2759.

### What

Step 1 (#2640) added the `SelectionPolicy.qp_pool` schema. Step 2 (#2759) builds a per-pool QP segment inside each endpoint, handshaked with that pool's SL/TC. **This step makes the routing real**: a transfer tagged with a `qp_pool` actually sends on that pool's QPs, giving link-layer isolation between traffic classes end to end.

### How — carry `qp_pool` like `device_mask`, then fold into the pool's segment

The pool name rides the exact same plumbing the existing `device_mask` already uses, so the data flow mirrors code reviewers know:

```
SelectionResult.qp_pool  ->  TaskInfo.qp_pool  ->  SubBatch.qp_pool
   ->  RdmaTask.qp_pool  ->  submitSlices() -> selectQpInPool()
```

- **`selectQpInPool()`** (pure, in `params.h` next to `computeQpPoolSegments`): given the segments, the requested pool, and a worker-lane candidate index, returns the QP index to use. A named, known pool folds the candidate into that pool's `[begin, begin+num_qp)` span; empty / unknown pool / no pools configured => the candidate passes through as `candidate % qp_count` (unchanged spray). Result is always in `[0, qp_count)`.
- **`submitSlices()`**: all slices in a list belong to one task, hence one pool, so it reads `slice->task->qp_pool` once and routes the whole list.

### Default path unchanged

With no `qp_pool` on the matched policy, the name is empty and `selectQpInPool` returns `candidate % qp_count` — byte-for-byte the prior round-robin. No pools configured ⇒ same.

### Tests

`SelectQpInPoolTest` (6 cases): empty pool name sprays across all QPs; no segments configured also sprays; a named pool folds only into its segment; unknown pool falls back to the whole range (never drops a transfer); negative candidate clamps to 0; and the result is always in range for any pool/candidate. Built and all 10 cases (with step 2's 4) pass on an H20 / CX-7 box; the `tent` library compiles clean.

### The line this completes

#2640 (schema) → #2759 (per-pool QPs + SL/TC) → this (routing). Together: a policy can name a `qp_pool`, that pool gets its own QPs with its own Service Level / Traffic Class, and its traffic is isolated onto those QPs — the end-to-end mechanism @alogfans outlined on the RFC.

### DIFF
--- a/mooncake-transfer-engine/tent/include/tent/runtime/transfer_engine_impl.h
+++ b/mooncake-transfer-engine/tent/include/tent/runtime/transfer_engine_impl.h
@@ -55,6 +55,7 @@ struct TaskInfo {
     int xport_priority{0};        // transport priority (for fallback)
     int failover_count{0};        // number of failover attempts
     uint64_t device_mask{~0ULL};  // Device mask for quota allocation
+    std::string qp_pool;          // Named QP pool (RFC #2568 step 3), "" = none
     Request request;
     bool staging{false};
     TransferStatusEnum status{TransferStatusEnum::PENDING};
--- a/mooncake-transfer-engine/tent/include/tent/runtime/transport.h
+++ b/mooncake-transfer-engine/tent/include/tent/runtime/transport.h
@@ -55,6 +55,10 @@ class Transport {
         }
 
         uint64_t device_mask;  // Device mask for transport selection
+        // Named QP pool for this batch's transfers (RFC #2568 step 3). Empty =
+        // no pool (default spray). Carried like device_mask, from the matched
+        // SelectionPolicy down to each RdmaTask.
+        std::string qp_pool;
         BatchID progress_batch_id{0};
         std::function<void(BatchID)> notify_progress;
     };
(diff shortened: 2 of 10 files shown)


alogfans:
One thing to double-check: this keeps the existing per-sub-batch policy granularity. In `commitPreparedSubmit()`, all RDMA requests in the sub-batch inherit the `qp_pool` from the first task, matching the previous `device_mask` behavior.

Could you evaluate whether that is still the intended contract for `qp_pool`, or whether we want to make this more flexible and carry the pool per request / per RdmaTask? The current approach is fine if we explicitly keep “one RDMA sub-batch uses one policy” as the invariant, but it would be good to document that since mixed-policy batches would route later requests through the first task’s pool.

catyans:
Heads-up: it looks like the step-3 changes in this PR are already on `main`. When #2759 was merged, the squashed commit brought in the whole `tent-qp-pool-routing` branch, which at that point already contained the step-3 commit — so `selectQpInPool` (params.h), its call site in endpoint.cpp, and the routing tests are all present on `main` now, byte-for-byte identical. A diff of this branch's tent changes against current `main` is empty, which is also why GitHub now shows it as CONFLICTING.

So there's nothing left to merge here. I'll close this PR unless you'd prefer I keep it open for tracking — happy to do whatever's cleaner on your end. (RFC #2568 is now fully landed: step1 #2640, step2 #2759, step3 folded in via #2759.)

catyans:
Closing — step-3 content is already on main (folded in via #2759's squash-merge), so this PR has no remaining diff. RFC #2568 is fully landed.

Issue #2611, opened 2026-06-25
CI: build-flags job appears to hit runner disk/network instability across recent PRs


VectorPeak:
### Summary

Recent `Build & Test (Linux)` workflow runs appear to be failing for CI infrastructure reasons rather than PR-specific code failures.

In PR #2601, the latest failure is in `build-flags (3.10)` and the log shows the GitHub-hosted runner running out of disk space while linking a store test executable:

```text
[356/424] Linking CXX executable mooncake-store/tests/master_metrics_test
FAILED: [code=1] mooncake-store/tests/master_metrics_test
##[error]No space left on device : '/home/runner/actions-runner/cached/2.335.1/_diag/pages/...log'
```

The PR itself only touches EP buffer validation files:

```text
mooncake-ep/src/ep_py.cpp
mooncake-ep/src/mooncake_ep_buffer.cpp
mooncake-wheel/mooncake/mooncake_ep_buffer.py
```

So this failure does not look directly related to the PR changes.

### Related observations

I also noticed similar CI instability on recent runs from other PRs / main-branch workflow runs. Several recent `Build & Test (Linux)` runs were failing, queued, or stuck in progress around the same period.

The previous run of PR #2601 also failed for infrastructure-looking reasons, including `sccache` / GitHub Actions cache DNS failures and GitHub fetch/network errors, for example:

```text
sccache: error: Server startup failed: cache storage failed to read: Unexpected (temporary) at read => send http request
error sending request ... dns error: failed to lookup address information: Try again
fatal: unable to access 'https://github.com/...': Failed to connect to github.com port 443
```

### Possible cause

The `build-flags` job seems especially heavy. It builds multiple configurations in the same runner, including transfer engine, full project builds, Rust bindings, EP variants, TENT, nvlink allocator, and wheel artifacts.

The run logs also show the root filesystem becoming increasingly full during the job, while `/mnt` still has much more free space available. This suggests the job may need either more aggressive cleanup or to place large build/cache directories under `/mnt`.

### Suggested checks

Could maintainers please take a look at the CI environment for this workflow?

Some possible mitigations:

- Split `build-flags` into smaller jobs or separate matrix stages.
- Clean intermediate build directories before the next large build phase.
- Move large build/cache directories to `/mnt` on GitHub-hosted runners.
- Add `df -h` / `du -sh` diagnostics after each major build phase.
- Consider disabling or isolating `sccache` when the Actions cache backend is unstable.

[...]

PR #2764, opened 2026-07-06
[TENT] Deadline-infeasible drop + degradation hook (RFC #2519 step 3)

catyans:
## What

RFC #2519 **step 3** (degradation layer), on top of step 2 (#2763, EDF ordering) and step 1 (#2618, `deadline_ns` + MLU observability). Owners predicted to miss their deadline are **dropped** from dispatch instead of sent, and a **local-decode signal** is raised so the caller can recompute locally rather than waste bandwidth.

> Depends on #2763 — this branch is stacked on it. Please review #2763 first; the diff here is step-3-only.

## How

- `QueueLimits.mlu_local_threshold` (θ_local, **default 0 = disabled**).
- `setDegradationPolicy(bandwidth_provider, hooks, now_provider)` — dependency-injected, so the admission queue stays decoupled from the device-selection layer and remains unit-testable. `now_provider` defaults to `steady_clock`.
- `pickForDispatch` gains an optional `dropped_owner_ids` out-param. When drop is enabled (θ_local > 0, `deadline_aware`, bandwidth provider set), an owner whose predicted MLU (`= length/bw / (deadline − now)`) reaches θ_local — or whose deadline is already past — is charged out of the outstanding accounting, marked terminal (`CANCELED`), reported in `dropped_owner_ids`, and triggers `on_local_decode_suggested`. Everything else dispatches as before.

**Interface only**: no codec / local-decode body (those live in vLLM/SGLang; TENT only raises the signal), as scoped in the RFC. Strictly additive and fully opt-in — θ_local = 0 (default) is zero behavior change from step 2.

## Tests

5 new unit tests: drop-infeasible / keep-feasible / expired-deadline drop / disabled-when-threshold-zero / no-drop-without-bandwidth-provider — including hook invocation and outstanding-accounting checks. Full `admission_queue_test` suite passes **21/21**.

## Why (measurement)

The MLU sweep in #2519 (H20 / CX-7 RoCEv2, TENT backend) shows the 100 µs deadline tier at **mean MLU 2.63** — well past a θ_local of ~1.5. Under contention those transfers are provably infeasible; step 3 drops them to local-decode instead of spending bandwidth on a transfer that will miss anyway.

## Not in scope

Caller wiring in `TransferEngineImpl` (bridging the real EWMA bandwidth into the provider, and acting on `dropped_owner_ids`) is deliberately separate — this PR is the self-contained, unit-tested policy core. Honest caveat: the drop decision is only as good as the injected bandwidth estimate, and the codec/local-decode implementations are upper-layer concerns.

### DIFF
--- a/mooncake-transfer-engine/tent/include/tent/runtime/admission_queue.h
+++ b/mooncake-transfer-engine/tent/include/tent/runtime/admission_queue.h
@@ -18,6 +18,7 @@
 #include <cstddef>
 #include <cstdint>
 #include <deque>
+#include <functional>
 #include <map>
 #include <utility>
 #include <vector>
@@ -49,6 +50,14 @@ struct QueueLimits {
     // them. This only reorders selection within the existing capacity limits;
     // it does not admit/reject or otherwise change what gets dispatched.
     bool deadline_aware{false};
+    // Opt-in deadline-infeasible drop (RFC #2519 step 3). Local-decode MLU
+    // threshold θ_local. 0 (default) disables drop entirely — behavior is the
+    // step-2 EDF ordering (or FIFO). When > 0 (e.g. 1.5) and a bandwidth
+    // provider is set, an owner whose predicted MLU
+    // (= predicted_transfer_time / remaining_window) reaches this threshold is
+    // dropped instead of dispatched, and on_local_decode_suggested is raised so
+    // the caller can recompute locally. Requires deadline_aware = true.
+    double mlu_local_threshold{0.0};
 };
 
 struct QueueOwnerInput {
@@ -66,6 +75,23 @@ struct QueueSubmit {
     std::vector<QueueOwnerInput> owners;
 };
 
+// RFC #2519 step 3: degradation signal raised when a transfer is predicted to
+// miss its deadline and is dropped from dispatch. The bodies (compression /
+// local recompute) live in the upper layer (vLLM/SGLang); TENT only raises the
+// signal. No hook registered ⇒ the drop still happens but nothing is notified.
+struct DegradationHooks {
+    std::function<void(const Request&)> on_local_decode_suggested;
+};
+
+// Returns the predicted transfer bandwidth in bytes/second, or <= 0 if unknown
+// (in which case the drop decision is skipped). Injected by the owner so the
+// admission queue does not depend on the device-selection layer directly.
+using BandwidthProvider = std::function<double()>;
+
+// Returns "now" as a steady-clock timestamp in nanoseconds, matching the units
+// of Request.deadline_ns. Injectable so tests are deterministic.
+using NowProvider = std::function<uint64_t()>;
+
 // Runtime-private admission model. It is intentionally single-threaded; the
 // eventual TransferEngineImpl integration owns synchronization.
 class LocalTransferAdmissionQueue {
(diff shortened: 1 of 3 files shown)


catyans:
Note on the diff: this branch is stacked on #2763, but since GitHub diffs against `main` (which doesn't have #2763 yet), the file view here shows **both** step 2 and step 3 changes. The step-3-only delta is the `mlu_local_threshold` / `DegradationHooks` / `setDegradationPolicy` / drop logic in `pickForDispatch` (+ the 5 `Step3*` tests). Once #2763 merges I'll rebase and the diff will collapse to step 3 only. Happy to hold this until #2763 is reviewed.

catyans:
Note: #2763 (step 2, which this is stacked on) was updated per @alogfans's review to keep fifo_ EDF-ordered at admission instead of re-sorting on every pickForDispatch. Once #2763 is finalized/merged I'll rebase this PR onto it so the step-3 drop path sits on the ordered-insert version too (the drop logic is unaffected — it still scans the now-ordered queue front-to-back). Keeping this as-is until #2763 settles to avoid churn.

catyans:
Rebased onto `main` now that #2763 (step 2) is merged. As noted earlier, this drops the old per-call `stable_sort` (step 2 landed as ordered-insert at admission, so `pickForDispatch` just consumes the already-EDF-ordered `fifo_` from the front) and keeps the step-3 drop path unchanged — it scans the ordered queue front-to-back, dropping owners predicted to miss their deadline.

Now a single commit on top of main, no conflicts. All 22 admission-queue tests pass locally (the merged step-2 `DeadlineAwareOrdersAcrossSeparateAdmits` + the 4 `Step3*` drop/degradation tests). Still opt-in and inert by default (θ_local = 0).

alogfans (approved):
LGTM

alogfans (changes requested):
Please check CI build

staryxchen:
Hi @catyans, thanks for your contribution. Please delete the sections in the code comments that refer to “RFC xxx step xxx”; I don't think this information should be included in the comments.

catyans:
Thanks @alogfans! The `tent-ci (cuda-off)` failure was an infra flake, not a code issue — the log shows `sccache: Server startup failed ... dns error: failed to lookup address information: Try again` and `Failed to update package lists` during the apt/sccache setup step, before any compilation. Every other job (build-flags, build-docker, all wheels, ascend-test, integration-test) passed, and the admission-queue tests pass locally (22/22, including the merged step-2 `DeadlineAwareOrdersAcrossSeparateAdmits` and the 4 `Step3*` tests).

I've pushed an empty commit to re-trigger CI. Should go green this time.

(Note: the earlier gemini comment about `stable_sort` over stale entries referred to the pre-rebase version — the rebase onto merged #2763 removed that sort entirely; `pickForDispatch` now just consumes the already-EDF-ordered `fifo_` from the front.)

catyans:
CI update, @alogfans: the two **`tent-ci` jobs (cuda-off and cuda-on) now pass** — that's the one that actually builds + runs the admission-queue change, and it's green (matches the 22/22 I get locally). The earlier failure was the apt/sccache network flake I mentioned.

The remaining red is `build-flags (3.10/3.12)`, which is an infra flake, not a compile error: the job ran ~17 min and was killed with orphan `cmake`/`ninja`/`python3.10` processes still alive (runner terminated mid-build), and the log has no `##[error]` compile failure, no `ninja: build stopped`, no disk-space message. This is the same `build-flags` runner instability tracked in #2611. `build-flags` doesn't exercise this change differently from `tent-ci` / the other build jobs, all of which pass.

Happy to empty-commit re-trigger `build-flags` if you'd like a clean board, but the change itself is verified green via tent-ci. Let me know if you'd prefer the re-trigger or are OK merging past the known-flaky job.

alogfans (approved):
LGTM, thanks!

PR #2770, opened 2026-07-07
[Bugfix] Reject empty keys in HTTP metadata server

VectorPeak:
### What Problem This Solves

The Python HTTP metadata server currently treats a missing `key` query parameter as an empty string. As a result, requests such as `PUT /metadata` or `PUT /metadata?key=` can operate on `self.store[""]` instead of being rejected.

This can create accidental empty-string metadata entries when callers omit the query parameter, and later empty-key requests can then appear valid instead of surfacing the original bad request.

### Changes

- Adds metadata key validation in `KVBootstrapServer._handle_metadata()` before dispatching GET, PUT, or DELETE handling.
- Rejects missing, empty, and whitespace-only metadata keys with HTTP 400.
- Strips valid non-empty metadata keys before they are passed to the existing handler path.

### Evidence

- Adds focused handler-level unit coverage for missing keys across GET, PUT, and DELETE.
- Adds coverage for empty and whitespace-only keys.
- Adds coverage for stripping leading and trailing whitespace before metadata operations, plus a valid-key PUT/GET round-trip test to confirm normal metadata storage and retrieval still works.

### Possible call chain / impact

External metadata requests enter through `/metadata`, then `KVBootstrapServer._handle_metadata()` dispatches to `_handle_get()`, `_handle_put()`, or `_handle_delete()`.

This change affects only the Python HTTP metadata server key handling path: missing, empty, and whitespace-only `key` query parameters are rejected, while valid non-empty keys are stripped before continuing through the existing handler path.
## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

Focused unit coverage validates the metadata handler directly:

- missing `key` is rejected for GET, PUT, and DELETE
- empty `key` is rejected
- whitespace-only `key` is rejected
- valid keys with leading/trailing whitespace are stripped before PUT/GET operations
- valid non-empty keys still round-trip through PUT and GET

**Test commands:**
```bash
python -m pytest mooncake-wheel/tests/test_http_metadata_server.py -q
python -m py_compile mooncake-wheel/mooncake/http_metadata_server.py mooncake-wheel/tests/test_http_metadata_server.py
git diff --check
```

**Test results:**
- [x] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

AI assistance was used to identify the empty-key API contract issue, inspect the affected handler path, and draft focused validation coverage. The human submitter is responsible for reviewing and validating the final changes.

### DIFF
--- a/mooncake-wheel/mooncake/http_metadata_server.py
+++ b/mooncake-wheel/mooncake/http_metadata_server.py
@@ -64,7 +64,7 @@ class KVBootstrapServer:
         key = request.query.get('key', '').strip()
         if not key:
             return web.Response(text='metadata key is required', status=400,
-                              content_type='text/plain')
+                              content_type='application/json')
         
         if request.method == 'GET':
             return await self._handle_get(key)
--- a/mooncake-wheel/mooncake/http_metadata_server.py
+++ b/mooncake-wheel/mooncake/http_metadata_server.py
@@ -61,10 +61,10 @@ class KVBootstrapServer:
 
     async def _handle_metadata(self, request: web.Request):
         """Handle metadata requests."""
-        key = request.query.get('key', '')
-        if not key.strip():
+        key = request.query.get('key', '').strip()
+        if not key:
             return web.Response(text='metadata key is required', status=400,
-                              content_type='application/json')
+                              content_type='text/plain')
         
         if request.method == 'GET':
             return await self._handle_get(key)
--- a/mooncake-wheel/mooncake/http_metadata_server.py
+++ b/mooncake-wheel/mooncake/http_metadata_server.py
@@ -62,6 +62,9 @@ class KVBootstrapServer:
     async def _handle_metadata(self, request: web.Request):
         """Handle metadata requests."""
         key = request.query.get('key', '')
+        if not key.strip():
+            return web.Response(text='metadata key is required', status=400,
+                              content_type='application/json')
         
         if request.method == 'GET':
             return await self._handle_get(key)
--- a/mooncake-wheel/tests/test_http_metadata_server.py
+++ b/mooncake-wheel/tests/test_http_metadata_server.py
@@ -1,0 +1,65 @@
+#!/usr/bin/env python3
+import sys
+import unittest
+from pathlib import Path
+
+sys.path.insert(0, str(Path(__file__).resolve().parents[1]))
+
+from mooncake.http_metadata_server import KVBootstrapServer
+
+
+class FakeRequest:
+    def __init__(self, method, key=None, body=b""):
+        self.method = method
+        self.query = {} if key is None else {"key": key}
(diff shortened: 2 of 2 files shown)


mooncake-wheel/mooncake/http_metadata_server.py
```diff
@@ -62,6 +62,9 @@ def _setup_routes(self):
     async def _handle_metadata(self, request: web.Request):
         """Handle metadata requests."""
         key = request.query.get('key', '')
+        if not key.strip():
+            return web.Response(text='metadata key is required', status=400,
+                              content_type='application/json')
```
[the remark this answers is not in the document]
VectorPeak: Thanks for the review. I addressed both points:

- `key` is now stripped before dispatching to GET/PUT/DELETE handlers, so values like `"  valid  "` are stored and looked up as `"valid"`.
- The 400 response now uses `content_type='text/plain'` to match the plain-text error body.

I also added focused tests covering the `text/plain` error response and the whitespace-stripping behavior before metadata operations.

Validation run locally:

```bash
python -m pytest mooncake-wheel/tests/test_http_metadata_server.py -q
python -m py_compile mooncake-wheel/mooncake/http_metadata_server.py mooncake-wheel/tests/test_http_metadata_server.py
git diff --check
```

mooncake-wheel/mooncake/http_metadata_server.py
```diff
@@ -61,7 +61,10 @@ def _setup_routes(self):
 
     async def _handle_metadata(self, request: web.Request):
         """Handle metadata requests."""
-        key = request.query.get('key', '')
+        key = request.query.get('key', '').strip()
+        if not key:
+            return web.Response(text='metadata key is required', status=400,
+                              content_type='text/plain')
```
Aionw: Keep consistency with other http handler, here we should return `application/json`

Aionw (approved):

mooncake-wheel/mooncake/http_metadata_server.py
VectorPeak: Thanks, updated. The empty-key 400 response now keeps `content_type='application/json'` for consistency with the other HTTP handlers, while preserving the previous reviewer-requested key stripping before dispatch.

I also updated the focused tests to assert the consistent content type and re-ran:

```bash
python -m pytest mooncake-wheel/tests/test_http_metadata_server.py -q
python -m py_compile mooncake-wheel/mooncake/http_metadata_server.py mooncake-wheel/tests/test_http_metadata_server.py
git diff --check
```

Issue #2306, opened 2026-06-03
[RFC]: Master-Side LOCAL_DISK Replica Warm Re-Adoption Across Client Restart


yzhan1:
# [RFC]: Master-Side LOCAL_DISK Replica Warm Re-Adoption Across Client Restart

## Summary

When a Mooncake client (e.g., vLLM PD instance using ModeA) restarts, the master expires it after `client_live_ttl_sec` (default 10 s) and **immediately erases** all of its segment registrations and `LOCAL_DISK` replica metadata. The on-disk SSD files survive but become master-side invisible. The new client must rediscover everything via `FileStorage::ScanMeta` — and for the default OffsetAllocator backend, today's `ScanMeta` is in-memory only (`storage_backend.cpp:3124`) so it recovers nothing at all until PR #2215 lands. On the other backends, recovery is slow but possible (see §Motivation). Prefix-cache hit rate drops to ~0% during the gap.

This RFC introduces:

1. A new `LocalDiskReplicaState` enum (`OK → DISCONNECTED → REMOVED`) carried in-place on every LOCAL_DISK replica, with a grace timer instead of immediate erase. A returning client transitions a `DISCONNECTED` replica back to `OK` atomically.
2. A `local_disk_segment_id` bound to an on-SSD marker file at `${storage_path}/.mooncake_local_disk_segment_id` that identifies the storage location (independent of the per-process random `client_id`).

A returning client whose marker file matches a `DISCONNECTED` segment is auto-adopted via the existing `MountLocalDiskSegment` RPC. No `ScanMeta` walk needed.

**Scope** — what this RFC does and does *not* enable:

| Scenario                                                        | Today                                     | With this RFC                                                                                           | Fixed? |
| --------------------------------------------------------------- | ----------------------------------------- | ------------------------------------------------------------------------------------------------------- | ------ |
| A restarts; A's next request after restart                      | misses for the recovery window (see §Motivation) | hits immediately (sub-second)                                                                           | ✅      |
| Master state correctness across A's restart                     | transient zero-replica window             | continuous                                                                                              | ✅      |
| Other client B reading A's keys *while A is dead*               | clean miss                                | **still a clean miss** (A's process is the RDMA endpoint)                                               | ❌      |
| DRAM survives restart                                           | lost                                      | **still lost** (separate RFC track: graceful shutdown)                                                  | ❌      |
[...]

PR #2777, opened 2026-07-07
store: reattach local disk replicas after restart

xzh25:
## Summary

Reattaches LOCAL_DISK replicas when a client restarts with the same on-disk segment marker.

## Changes

- Keep expired LOCAL_DISK replicas hidden during a configurable grace period.
- Persist a storage-bound marker and pass it through `MountLocalDiskSegment`.
- Reattach matching replicas with the restarted client's endpoint and restore SSD accounting.
- Remove stale disconnected copies on overwrite and keep eviction/offload decisions limited to completed disk replicas.
- Keep disconnect state runtime-only; no snapshot schema change or promotion-on-hit changes.

## Validation

- Release build and clang-format check passed.
- `master_service_ssd_test`: 27/27 passed.
- `file_storage_test`, `segment_test`, `serializer_test`, and `master_service_ssd_test_for_snapshot` passed.
- Process repro: 4/4 LOCAL_DISK-only reads recovered after one mount reattachment; first successful read was 5 ms after restart, with no `ScanMeta` recovery in the logs.

Refs #2306

### DIFF
--- a/mooncake-store/include/client_service.h
+++ b/mooncake-store/include/client_service.h
@@ -411,7 +411,9 @@ class Client {
      * @brief Mounts a local disk segment into the master.
      * @param enable_offloading If true, enables offloading (write-to-file).
      */
-    tl::expected<void, ErrorCode> MountLocalDiskSegment(bool enable_offloading);
+    tl::expected<void, ErrorCode> MountLocalDiskSegment(
+        bool enable_offloading, const std::string& local_disk_segment_id = {},
+        const std::string& transport_endpoint = {});
 
     /**
      * @brief Heartbeat call to collect object-level statistics and retrieve the
--- a/mooncake-store/include/master_client.h
+++ b/mooncake-store/include/master_client.h
@@ -420,7 +420,9 @@ class MasterClient {
      * @param enable_offloading If true, enables offloading (write-to-file).
      */
     [[nodiscard]] tl::expected<void, ErrorCode> MountLocalDiskSegment(
-        const UUID& client_id, bool enable_offloading);
+        const UUID& client_id, bool enable_offloading,
+        const std::string& local_disk_segment_id = {},
+        const std::string& transport_endpoint = {});
 
     /**
      * @brief Heartbeat call to collect object-level statistics and retrieve the
--- a/mooncake-store/include/master_config.h
+++ b/mooncake-store/include/master_config.h
@@ -45,6 +45,7 @@ struct MasterConfig {
     double nof_eviction_ratio;
     double nof_eviction_high_watermark_ratio;
     int64_t client_live_ttl_sec;
+    int64_t local_disk_rejoin_grace_sec;
     int64_t nof_heartbeat_interval_sec;
     uint32_t nof_heartbeat_probe_timeout_ms;
     uint32_t nof_heartbeat_failures_threshold;
@@ -164,6 +165,7 @@ class MasterServiceSupervisorConfig {
     RequiredParam<double> nof_eviction_high_watermark_ratio{
         "nof_eviction_high_watermark_ratio"};
     RequiredParam<int64_t> client_live_ttl_sec{"client_live_ttl_sec"};
+    int64_t local_disk_rejoin_grace_sec = 600;
     RequiredParam<int64_t> nof_heartbeat_interval_sec{
         "nof_heartbeat_interval_sec"};
     RequiredParam<uint32_t> nof_heartbeat_probe_timeout_ms{
@@ -266,6 +268,7 @@ class MasterServiceSupervisorConfig {
         nof_eviction_high_watermark_ratio =
             config.nof_eviction_high_watermark_ratio;
         client_live_ttl_sec = config.client_live_ttl_sec;
+        local_disk_rejoin_grace_sec = config.local_disk_rejoin_grace_sec;
         nof_heartbeat_interval_sec = config.nof_heartbeat_interval_sec;
         nof_heartbeat_probe_timeout_ms = config.nof_heartbeat_probe_timeout_ms;
         nof_heartbeat_failures_threshold =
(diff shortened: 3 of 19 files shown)


xzh25:
Thanks. I pushed a follow-up that addresses the actionable items:

- serialize local disk marker creation inside the process
- remove a partially-created marker if write/close fails
- add a disconnected-replica count so reattach skips the full scan when there are no disconnected LOCAL_DISK replicas
- rerun clang-format locally

I also rebuilt locally after the change.

xzh25:
Hi @yzhan1, could you take a quick look at this draft when you have time?

This is intended to address the core #2306 path: retain LOCAL_DISK replicas as disconnected during the grace window, keep them hidden from readers, and reattach them on `MountLocalDiskSegment` using a persistent local disk segment marker.

I am working under a short contest timeline, so even an early direction check would help: whether this approach is worth continuing, or whether you would prefer a different shape before deeper polishing.

xzh25:
Addressed the Codex review edge cases in db7b0c4b:

- enqueue client liveness before local-disk mount/reattach makes replicas visible
- reattach same-marker replacement clients that mount before the old TTL expires
- restore SSD usage accounting when local-disk replicas are reattached
- require a completed LOCAL_DISK replica before offload-on-evict removes memory
- remove disconnected LOCAL_DISK replicas before overwrite paths
- persist local_disk_segment_id/disconnected age in replica snapshots and rebuild the disconnected count on restore

Validated locally:
- code_format.sh --check
- cmake --build build-v100-core -j2
- master_service_ssd_test
- master_service_ssd_test_for_snapshot

xzh25:
Updated PR #2777 with follow-up fixes for the latest review comments:

- run local-disk grace cleanup from the task cleanup thread while disconnected replicas exist
- persist LocalDiskSegment marker/endpoint across snapshot restore, with legacy snapshot compatibility
- release promotion pins/staged replicas when stale local-disk cleanup removes a promotion task
- add a fresh LOCAL_DISK replica on offload completion when existing disk replicas only belong to stale/different clients
- allow forced removes to drop disconnected LOCAL_DISK replicas during the grace window

Local checks passed:
- `./scripts/code_format.sh -b origin/main --check`
- `cmake --build build-v100-core -j2`
- `ctest --test-dir build-v100-unit -R '^(master_service_ssd_test|master_service_ssd_test_for_snapshot|promotion_on_hit_test)$' --output-on-failure`

xzh25:
Addressed the latest review edge cases and rebased this branch onto current main.

Could a maintainer confirm whether this retain-disconnected-and-reattach approach is an acceptable direction for #2306? If the design is not a good fit for Mooncake, I can stop expanding this PR and narrow or close it.

xzh25:
Addressed the latest snapshot-restore and promotion-task comments in `6c14a5d6`. Also fixed the two failing coverage ctest cases locally.

Local checks: format, targeted build, focused tests, and the affected ctest subset all passed.

xzh25:
Rebased onto current main and narrowed back to the core RFC scope. Snapshot-schema and promotion followups were removed; the current PR is one commit. Local build, format, focused tests, and the restart repro pass.

xzh25:
@yzhan1 This is now rebased onto current main and reduced to one core commit. Could you confirm whether the retain-and-reattach direction fits #2306? I will keep further changes limited to maintainer feedback and CI failures.

xzh25:
Pushed a focused follow-up for the three failing CI cases: duplicate LOCAL_DISK completion no longer adds SSD usage, stale-owner cleanup fully cancels promotion tasks, and BatchRemove preserves the existing stale-handle result. The affected tests pass locally; the focused suite is 7/7, and the restart repro reattached 32 replicas with 4/4 reads and no ScanMeta.

LujhCoconut:
Hi @xzh25 — first, apologies for only noticing this PR now; I wish I'd found it while it was active, because the design direction here looks like the right one for #2306 to me.

For anyone else landing here, my reading of the approach:

- A persistent on-disk marker (`.mooncake_local_disk_segment_id`) gives each physical SSD instance a stable identity, independent of both the process UUID and the (possibly ephemeral) RPC port;
- when a client expires, its LOCAL_DISK replicas are retained as DISCONNECTED — hidden from readers and excluded from eviction decisions — for a configurable grace period, instead of being deleted immediately;
- on `MountLocalDiskSegment`, replicas matching the segment id are reattached to the new client incarnation, with the endpoint refreshed and SSD accounting transferred.

Reviewing the current state, a few things would need attention before this could move forward:

- It needs a non-trivial rebase: local-SSD accounting has since moved into `LocalSsdManager`, so the reattach / `AddLocalDiskUsage` parts need reworking onto it.
- Two wire-format changes (the `MountLocalDiskSegment` RPC signature, and the added `YLT_REFL` field on `LocalDiskDescriptor`) will break mixed-version clusters and HA replay during rolling upgrades — they probably need an optional-field or version-tolerance story.
- `update_local_disk_metadata` mutates `object_size`, which unbalances the RAII file-size metric and SSD accounting; making size immutable and rejecting mismatches (as #3182 does) seems safer.
- The reattach path can adopt COMPLETE replicas from a still-live owner presenting the same segment id — without single-writer protection on the SSD directory (e.g., an flock), a misconfigured second client could silently take over a live client's replicas.
- The snapshot serializer doesn't persist the segment id, so after a master restart the existing replicas can no longer be reattached and will age out of the grace window — worth fixing or at least calling out explicitly.
- Smaller: the usage charge in `NotifyOffloadSuccess` can land on a detached segment record (or be skipped) while the replica is still created — same non-atomicity class being discussed in #3182; the marker write has no fsync; and the random id generation should use `include/random.h` per `mooncake-store/AGENTS.md`.

None of these look fundamental — the contract itself (persistent disk identity + retain-and-reattach) still seems sound, and it covers the default-configuration case that endpoint matching can't reach.

So: would you still be interested in continuing this PR — rebasing onto current main and revising along the lines above? Totally understandable if your timeline has moved on; in that case, would you be open to someone else picking it up?

For context: #3052 is the current bug report for this scenario, and #3182 is an active, deliberately narrower fix (ownership transfer on matching transport endpoint — opt-in via a stable port). I've suggested there that it can merge first as the opt-in path, while this design remains the default-on answer under #2306. You're very welcome to join that discussion too.

PR #2781, opened 2026-07-07
[store] Opt-in topology-aware remote replica scoring in SelectBestReplica (#2516)

catyans:
## What

When the master returns multiple **remote MEMORY** replicas for a key,
`SelectBestReplica` (mooncake-store) kept the *first* one it encountered, so the
choice among otherwise-equivalent remote replicas was effectively arbitrary
(master return order). This adds an **opt-in** scoring hook so a better remote
replica can be chosen instead. Addresses #2516.

## How

- **Extract** `SelectBestReplica` + helpers into `mooncake-store/include/replica_selection.h`.
  They previously lived in an anonymous namespace inside `real_client.cpp` and
  were unreachable from tests; moving them to a header makes the pure selection
  logic unit-testable with no behaviour change to callers.
- **Add a `ReplicaScorer` injection point** (`SetRemoteReplicaScorer`) plus a
  built-in protocol-priority scorer (prefer `rdma` over `tcp`). This is the
  layering @LujhCoconut suggested on #2516 ("implementing the ranking logic in
  mooncake-store seems like a better approach"): richer signals — NIC role,
  NUMA distance, live NIC load — live in the transfer engine and can be fed in
  through the scorer, so mooncake-store does not take a dependency on that layer.
- **Disabled by default.** Behaviour is byte-identical to the historical
  "first remote MEMORY" pick unless the operator sets
  `MC_STORE_REPLICA_SCORING=1` or a scorer is injected. Local replicas and the
  non-MEMORY fallback chain (NOF_SSD / LOCAL_DISK / DISK) are unchanged; ties
  keep master return order (strictly-less comparison), so a symmetric cluster
  degrades to today's behaviour exactly.

## Scope / honesty

This is a mechanism PoC, deliberately minimal:
- The built-in scorer only uses info the replica descriptor already carries
  (`protocol_`). The high-value signal from measurements — **which class of NIC
  a replica is reached through** (issue #2516 comment: head-node VF vs tail
  bonded NIC is a +102% latency delta at 1 MB on H20/CX-7) — is *not* computed
  here; it is exactly what the `SetRemoteReplicaScorer` hook is for, to be fed
  from TENT in a follow-up.
- The path only triggers when a key has >1 remote MEMORY replica
  (`replica_num >= 2` with replicas on different segments). With the default
  `replica_num = 1` it is inert.

## Test

New `replica_selection_test` (7 cases): base policy unchanged
(local-wins, scoring-off keeps first remote), opt-in scoring picks lowest score,
built-in rdma>tcp, tie keeps master order, incomplete-replica skip, and
local-still-wins-when-scoring-on. All pass locally (g++ -std=c++20).


catyans:
cc @LujhCoconut — this is the mooncake-store implementation you suggested on #2516 ("implementing the ranking logic in mooncake-store seems like a better approach").

Summary of the approach here:
- `SelectBestReplica` is refactored out of the anonymous namespace in `real_client.cpp` into `replica_selection.h` so the selection logic is unit-testable (7 new tests).
- A `ReplicaScorer` injection point (`SetRemoteReplicaScorer`) lets the transfer engine feed in the topology/load signals you correctly pointed out that `DeviceSelector` can't produce on its own for *remote* replicas — so mooncake-store stays free of any TENT dependency.
- Opt-in and inert by default: without `MC_STORE_REPLICA_SCORING=1` or an injected scorer, behaviour is byte-identical to today's "first remote MEMORY" pick.

Happy to iterate on the hook shape or the built-in scorer's policy. Would appreciate your take on whether the injection boundary is the right seam.

mooncake-store/include/replica_selection.h
```diff
@@ -0,0 +1,147 @@
...
+// racy-benign: an empty function simply falls back to the built-in order.
+inline ReplicaScorer &RemoteReplicaScorer() {
+    static ReplicaScorer scorer;
+    return scorer;
+}
+
+// Inject a topology-/load-aware scorer (e.g. from the transfer engine).
+inline void SetRemoteReplicaScorer(ReplicaScorer scorer) {
+    RemoteReplicaScorer() = std::move(scorer);
+}
```
LujhCoconut, line 68: `RemoteReplicaScorer() = std::move(scorer)` — the `=` assignment here is a non-atomic write to the static `std::function` object. It mutates multiple internal fields (target pointer, vtable pointer, small buffer) across several instructions.

Meanwhile, three read paths touch the same static object concurrently:

- L70: `static_cast<bool>(RemoteReplicaScorer())` — reads internal state
- L79: `const ReplicaScorer &injected = RemoteReplicaScorer()` — binds a reference to the static, does not copy, so injected is an alias for the same object
- L88: `injected ? injected(r) : ... — calls operator()`, which reads the stored function pointer and invokes through it

If L48 (write) and any of the above (read) execute concurrently on the same static object, it's a data race — `operator=` may not have finished updating all fields when operator() reads them, leading to invocation through a torn vtable pointer → SIGSEGV.

catyans: Good catch — you're right that this is a textbook data race: `operator=` mutates multiple internal fields (target pointer + vtable pointer + small-buffer content) non-atomically, so a concurrent `operator()` or `operator bool` reads a torn state → UB/SIGSEGV.

Fixed in the latest push (4f7fab0d):

**Approach**: Wrap the static `ReplicaScorer` with a `std::shared_mutex`. Readers (`GetRemoteReplicaScorer`) take a `shared_lock`, **copy** the scorer out, then invoke the copy outside the lock (so user-supplied scorer code never runs under the lock → no deadlock risk, no concurrency bottleneck on the hot path). Writers (`SetRemoteReplicaScorer`) take a `unique_lock` — expected to be called once at init time.

**Validation**: 8 reader threads × 1 writer thread, 50 000 write iterations, 3.2M concurrent reader calls — zero crash, zero assert failure, on a bare-metal H20 node.

Also added a `ConcurrentSetAndSelectIsRaceFree` gtest that exercises this exact contention pattern so CI catches any future regression.

LujhCoconut (approved):
LGTM. Thanks for this contribution. Reviewing this was enjoyable.

catyans:
Validation update for `7f82f460`:

- Release validation passed on two independent H20 cluster nodes (4/4 relevant test suites on each node).
- Deterministic repeat and concurrent stress validation passed.
- ASan/LSan and TSan runs were clean on the primary validation node.
- `replica_selection.h` reached 100% line coverage in the targeted coverage run.
- Upstream CI is fully green, including x86, arm64, Ascend, MUSA, wheel, format, docs, and integration jobs.

This validates the selector model’s correctness, fallback compatibility, and concurrency safety. It does **not** claim an end-to-end Store hit-rate/TTFT performance gain; that requires the follow-up opt-in signal provider and workload scorecard.

PR #2778, opened 2026-07-07
[Store] Fix S3 list objects pagination

feichai0017:
## Description

Fixes #2735.

This PR fixes S3 prefix listing pagination in `S3Helper::ListObjectsWithPrefix()`:

- Uses `ListObjectsV2` instead of the legacy `ListObjects` API.
- Advances pages with `NextContinuationToken` instead of `NextMarker`.
- Returns an error if S3 reports a truncated response without a continuation token, avoiding an infinite loop.

## Module

- [ ] Transfer Engine (`mooncake-transfer-engine`)
- [x] Mooncake Store (`mooncake-store`)
- [ ] Mooncake EP (`mooncake-ep`)
- [ ] Mooncake PG (`mooncake-pg`)
- [ ] Integration (`mooncake-integration`)
- [ ] P2P Store (`mooncake-p2p-store`)
- [ ] Python Wheel (`mooncake-wheel`)
- [ ] Common (`mooncake-common`)
- [ ] Mooncake RL (`mooncake-rl`)
- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
git diff --check
SKIP=mooncake-code-format pre-commit run --files mooncake-store/src/utils/s3_helper.cpp
```

**Test results:**
- [ ] Unit tests pass
- [ ] Integration tests pass (if applicable)
- [ ] Manual testing done (describe below)

## Checklist

- [x] I have performed a self-review of my own code
- [ ] I have formatted my code using `./scripts/code_format.sh` (blocked locally by missing `clang-format-20`)
- [ ] I have run `pre-commit run --all-files` and all hooks pass
- [x] I have updated the documentation (if applicable; not applicable for this bug fix)
- [ ] I have added tests to prove my changes are effective
- [x] For changes >500 LOC: I have filed an RFC issue (not applicable; this PR changes one source file)

## AI Assistance Disclosure

- [ ] No AI tools were used
- [x] AI tools were used: Codex assisted with issue triage, implementation, and PR description.

PR #2782, opened 2026-07-07
[store] Fix S3 ListObjectsWithPrefix infinite loop on >1000 objects (#2735)

### LINKED ISSUE #2735: [Bug]: S3Helper::ListObjectsWithPrefix will cause dead loop
### Bug Report

The current S3Helper::ListObjectsWithPrefix implementation uses the ListObjects API instead of the ListObjectsV2 API. According to the official documentation for this ListObjects API (https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjects.html), result.GetNextMarker() only returns a value when WithDelimiter is set in the request. Since we clearly don't set it, this will cause an infinite loop when the number of keys in the object storage exceeds 1000.

object_keys=std::vector of length 1299000, but our keys only 1300, beacuse it in the dead loop.
```
#16 0x000063f88cce93da in mooncake::S3Helper::ListObjectsWithPrefix (this=0x63f899078498, prefix="mooncake_master_snapshot/", object_keys=std::vector of length 1299000, capacity 2097152 = {...}) at /workspace/Mooncake/mooncake-store/src/utils/s3_helper.cpp:834
#17 0x000063f88ccc5f2a in mooncake::S3SnapshotObjectStore::ListObjectsWithPrefix (this=this@entry=0x63f897fe1fe0, prefix="mooncake_master_snapshot/", object_keys=std::vector of length 1299000, capacity 2097152 = {...}) at /usr/include/c++/11/bits/unique_ptr.h:173
#18 0x000063f88ccc6e8e in mooncake::ha::backends::embedded::EmbeddedSnapshotCatalogStore::List (this=0x63f8990d0020, limit=0) at /workspace/Mooncake/mooncake-store/src/ha/snapshot/catalog/backends/embedded/embedded_snapshot_catalog_store.cpp:125
#19 0x000063f88cd251d2 in mooncake::MasterService::RestoreState (this=this@entry=0x79c93c3a9020) at /workspace/Mooncake/mooncake-store/src/master_service.cpp:4104
#20 0x000063f88cd26d13 in mooncake::MasterService::MasterService (this=this@entry=0x79c93c3a9020, config=...) at /workspace/Mooncake/mooncake-store/src/master_service.cpp:176
#21 0x000063f88cc0bfbb in mooncake::WrappedMasterService::WrappedMasterService (this=this@entry=0x79c93c3a9020, config=...) at /workspace/Mooncake/mooncake-store/src/rpc_service.cpp:216
#22 0x000063f88ccb02b9 in std::construct_at<mooncake::WrappedMasterService, mooncake::WrappedMasterServiceConfig> (__location=0x79c93c3a9020) at /usr/include/c++/11/bits/stl_construct.h:94
#23 std::allocator_traits<std::allocator<mooncake::WrappedMasterService> >::construct<mooncake::WrappedMasterService, mooncake::WrappedMasterServiceConfig> (__p=0x79c93c3a9020, __a=...) at /usr/include/c++/11/bits/alloc_traits.h:518
#24 std::_Sp_counted_ptr_inplace<mooncake::WrappedMasterService, std::allocator<mooncake::WrappedMasterService>, (__gnu_cxx::_Lock_policy)2>::_Sp_counted_ptr_inplace<mooncake::WrappedMasterServiceConfig> (__a=..., this=0x79c93c3a9010) at /usr/include/c++/11/bits/shared_ptr_base.h:519
```

### Before submitting...

catyans:
## Bug

`S3Helper::ListObjectsWithPrefix` uses the legacy `ListObjects` API and paginates
with `NextMarker`:

```cpp
if (result.GetIsTruncated()) {
    request.WithMarker(result.GetNextMarker());
}
```

Per the [AWS S3 `ListObjects` API](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjects.html),
`NextMarker` is **only populated when the request sets a `Delimiter`**. This code
sets none, so `GetNextMarker()` always returns an empty string. Once the listing
is truncated (> `MaxKeys` = 1000 objects), `WithMarker("")` restarts the listing
from the beginning and the loop never terminates.

Reported in #2735: a bucket with ~1300 objects grew `object_keys` to ~1.3M
entries (the same first 1000 keys re-appended ~1299 times) before the process
was killed — it hangs the snapshot-catalog listing path
(`S3SnapshotObjectStore` → `EmbeddedSnapshotCatalogStore`).

## Fix

Switch to `ListObjectsV2` and paginate with `ContinuationToken`. Per the AWS SDK,
`NextContinuationToken` **is sent whenever `IsTruncated` is true**, regardless of
delimiter, so pagination is correct:

```cpp
auto outcome = s3_client_.ListObjectsV2(request);
...
if (result.GetIsTruncated()) {
    request.WithContinuationToken(result.GetNextContinuationToken());
}
```

`ListObjectsV2` is the API AWS recommends for new development.

## Verification

- API surface confirmed against the AWS SDK for C++ docs
  (`ListObjectsV2Result::GetNextContinuationToken()` / `GetIsTruncated()` /
  `GetContents()`, and `ListObjectsV2Request::WithContinuationToken()`).
- clang-format clean.
- Note: I don't have the AWS SDK in my lab environment, so I could not compile
  the S3 code path locally; the change is a mechanical, doc-verified swap of the
  list API + pagination token, confined to this one function. A maintainer with
  the S3 build enabled can confirm the compile.

Fixes #2735.

### DIFF
--- a/mooncake-store/src/utils/s3_helper.cpp
+++ b/mooncake-store/src/utils/s3_helper.cpp
@@ -6,7 +6,7 @@
 #include <aws/s3/model/DeleteObjectRequest.h>
 #include <aws/core/auth/AWSCredentials.h>
 #include <aws/core/utils/memory/stl/AWSVector.h>
-#include <aws/s3/model/ListObjectsRequest.h>
+#include <aws/s3/model/ListObjectsV2Request.h>
 #include <aws/s3/model/DeleteObjectsRequest.h>
 #include <aws/s3/model/ObjectIdentifier.h>
 #include <aws/s3/model/Delete.h>
@@ -781,7 +781,7 @@ tl::expected<void, std::string> S3Helper::ListObjectsWithPrefix(
     const std::string &prefix, std::vector<std::string> &object_keys) {
     object_keys.clear();
 
-    Aws::S3::Model::ListObjectsRequest request;
+    Aws::S3::Model::ListObjectsV2Request request;
     request.WithBucket(bucket_);
     request.WithPrefix(prefix);
 
@@ -789,12 +789,18 @@ tl::expected<void, std::string> S3Helper::ListObjectsWithPrefix(
     // objects
     request.WithMaxKeys(1000);
 
+    // Use ListObjectsV2 + continuation token for pagination. The legacy
+    // ListObjects API only populates NextMarker when a delimiter is set; with
+    // no delimiter GetNextMarker() returns empty, so WithMarker("") restarts
+    // from the beginning and loops forever once the listing exceeds MaxKeys
+    // (#2735). ContinuationToken is always populated when the result is
+    // truncated, regardless of delimiter.
     bool done = false;
     while (!done) {
-        auto outcome = s3_client_.ListObjects(request);
+        auto outcome = s3_client_.ListObjectsV2(request);
         if (!outcome.IsSuccess()) {
             return tl::make_unexpected(fmt::format(
-                "ListObjects error: {}", outcome.GetError().GetMessage()));
+                "ListObjectsV2 error: {}", outcome.GetError().GetMessage()));
         }
 
         const auto &result = outcome.GetResult();
@@ -806,8 +812,8 @@ tl::expected<void, std::string> S3Helper::ListObjectsWithPrefix(
 
         // Check if there are more objects to fetch
         if (result.GetIsTruncated()) {
-            // Set marker to get next page
-            request.WithMarker(result.GetNextMarker());
+            // Set continuation token to get next page
+            request.WithContinuationToken(result.GetNextContinuationToken());
         } else {
             done = true;
         }


catyans:
On testing: I did **not** add a unit test, and want to be upfront about why.

- `s3_helper.cpp` is compiled only when `find_package(AWSSDK)` succeeds, and `S3Helper` holds a concrete `Aws::S3::S3Client` member (not an injectable interface), so `ListObjectsWithPrefix` can't be exercised without either a live S3 endpoint or refactoring the client to be mockable.
- There is currently no unit test for `S3Helper` in the repo (the `ListObjectsWithPrefix` tests under `tests/ha/snapshot/...` cover the LocalFile / Redis / Embedded backends, not the S3 one).
- The change itself is a mechanical swap of the list API + pagination token, with the API surface verified against the AWS SDK for C++ docs.

If you'd like a regression test, I'm happy to do it — but the clean way would be to make the S3 client injectable (introduce a small interface / pass a client factory) so a fake can simulate a truncated multi-page listing. That's a larger, separate change than this fix; let me know if you'd want it bundled here or as a follow-up.

Icedcoco:
Thanks for the fix and the detailed root-cause write-up.

This overlaps completely with #2778, which fixes the same `ListObjectsWithPrefix` pagination bug by switching to `ListObjectsV2` / `NextContinuationToken`, and also adds a small guard for the unexpected case where S3 reports a truncated response without a continuation token.

Since #2778 was submitted earlier, I would prefer to merge the earlier PR for this duplicate fix. I’m going to close this PR in favor of #2778.

I hope this does not discourage you from contributing. The analysis here is correct and useful, and I’m looking forward to your next PR.

PR #2788, opened 2026-07-08
[TENT] Opt-in per-entry priority promotion (#2528)

### LINKED ISSUE #2528: [TENT] Priority promotion decides on queue head but promotes the whole tier
### Describe the issue

While reading the TENT worker priority-promotion logic
(`mooncake-transfer-engine/tent/src/transport/rdma/workers.cpp`,
`Workers::promoteTimedOutRequests`, around L393-434), I noticed a few
things in the starvation-prevention path that look unintended. Filing to
confirm whether these are by design before proposing a fix.

The relevant block (MEDIUM→HIGH; LOW→MEDIUM is symmetric):

```cpp
std::vector<RdmaSliceList> promoted;
worker.queues[PRIO_MEDIUM].pop(promoted);          // drains the WHOLE queue
if (!promoted.empty()) {
    auto* slice = promoted.front().first;          // only inspects the HEAD
    if (slice && slice->enqueue_ts > 0 &&
        (current_ts - slice->enqueue_ts) >= priority_promotion_timeout_ns_) {
        for (auto& slice_list : promoted)
            worker.queues[PRIO_HIGH].push(slice_list);   // promote ALL
        return;
    }
    for (auto& slice_list : promoted)
        worker.queues[PRIO_MEDIUM].push(slice_list);     // push ALL back
}
```

**1. Promotion decision is made from the head element only, but applied to the whole queue.**
`pop(promoted)` drains the entire MEDIUM queue, then only
`promoted.front()`'s `enqueue_ts` is checked. If the head is timed out,
*every* MEDIUM entry is promoted to HIGH — including ones that were just
enqueued and aren't starving. Conversely, if the head is *not* timed out
but later entries are, none are promoted this round.

**2. Only one level is promoted per call.**
When MEDIUM→HIGH fires, the function `return`s, so the LOW→MEDIUM check is
skipped for that 1ms tick. A LOW entry needs multiple promotion cycles to
climb to HIGH. (Minor, but worth noting alongside #1.)

**3. Drain-then-refill on the non-timeout path.**
When the head is not timed out, the code still `pop`s the entire queue and
`push`es every element back every ~1ms. `pop`/`push` preserve FIFO order
in isolation, so ordering across a single quiet tick is fine — but since
the queue is MPSC (multiple producers can `push` concurrently during the
drain/refill window), interleaving the drained refill with concurrent
producer pushes can change the relative order, and the full drain+refill
is pure overhead on every idle tick.

### Question

Is the **whole-queue promotion driven by the head element** intentional
(e.g. a deliberate coarse-grained "if the oldest is starving, flush the
tier" policy), or is the intended behavior to promote only the entries
that have actually exceeded `priority_promotion_timeout_ns_`?

If per-entry (or at least non-draining) promotion is preferred, I'm happy
to put up a small PR — wanted to confirm the design intent first since
this is starvation-prevention behavior.

(To be clear: I checked `BoundedMPSCQueue::pop(vector)`/`push` and they do
**not** reverse order on their own, so this is *not* a FIFO-reversal bug —
the concern is the head-only decision + whole-queue effect + the
per-tick drain/refill, not order inversion.)

### Environment

- Reading `main` source (TENT transport).
- N/A — code-review observation, not a runtime failure.

catyans:
## What

`Workers::promoteTimedOutRequests` drains a whole priority queue but decides
promotion from the **head entry only**, then promotes **every** entry when the
head has timed out. As reported in #2528 this means:

1. a timed-out head drags freshly enqueued, non-starving entries up with it;
2. timed-out entries behind a fresh head are not promoted;
3. at most one level is promoted per tick.

@alogfans noted on #2528 that the head-only "flush the tier" policy is
intentional (simple + avoids scanning the queue every tick), and is open to
refining it. This PR keeps that policy as the default and adds the per-entry
policy as an **opt-in**, so nothing changes unless an operator turns it on.

## How

- Factor the promotion decision out of `promoteTimedOutRequests` into
  `promotion_policy.h`:
  - `DecidePromotionHeadOnly` — today's behavior (promote all iff head timed out);
  - `DecidePromotionPerEntry` — promote exactly the entries that have themselves
    timed out.
  The decision only depends on each entry's `enqueue_ts`, so it is unit-testable
  without the RDMA stack.
- Add config `transports/rdma/priority_promotion_per_entry` (default `false`):
  - default: byte-for-byte the historical head-only policy, one level per tick;
  - `true`: promote only timed-out entries, and consider both MEDIUM→HIGH and
    LOW→MEDIUM each tick so a starving LOW entry isn't stalled behind an
    unrelated MEDIUM promotion.

## Test

`promotion_policy_test` (5 cases, deterministic, no RDMA/timing):
- head-only promotes fresh entries `{0,1,2}` when only the head timed out, vs
  per-entry `{0}`;
- head-only misses a timed-out tail (`{}`) when the head is fresh, vs per-entry
  `{1,2}`;
- both policies agree on all-timed-out / empty / zero-timestamp (the cases the
  head-only design targets).

All pass locally (g++ -std=c++20). Cost note: per-entry selection is a single
pass over the drained queue (bounded by queue depth, once per ~1ms tick), so it
does not add the per-tick scan cost the head-only shortcut was avoiding beyond
the drain that already happens.

Closes #2528.

### DIFF
--- a/docs/source/design/tent/qos.md
+++ b/docs/source/design/tent/qos.md
@@ -62,6 +62,20 @@ This ensures that:
 - High-priority requests normally never wait behind lower-priority work
 - Low-priority requests eventually get serviced even under continuous high-priority load
 
+**Promotion configuration**:
+
+| Config key | Default | Behavior |
+|---|---|---|
+| `transports/rdma/priority_promotion_timeout_us` | `10000` (10ms) | How long an entry may wait before it is eligible for promotion. |
+| `transports/rdma/priority_promotion_per_entry` | `false` | Selects the promotion policy (see below). |
+
+`priority_promotion_per_entry` controls *which* entries a promotion pass moves up:
+
+- **`false` (default, head-only)**: a pass inspects only the queue head; if the head has timed out, the whole queue is promoted one level. This is the original, lowest-overhead "flush the tier" behavior — coarse, but it never scans the queue.
+- **`true` (per-entry)**: a pass promotes exactly the entries that have themselves timed out, leaving freshly enqueued entries in place, and considers both MEDIUM→HIGH and LOW→MEDIUM in the same tick. This avoids promoting non-starving requests and avoids stalling a starving LOW entry behind an unrelated MEDIUM promotion, at the cost of scanning the drained queue. Behavior is identical to head-only for the all-timed-out / empty cases.
+
+The default is byte-for-byte the historical behavior; set the flag to `true` to opt into finer-grained, fairer promotion.
+
 ### Global Slot Coordination
 
 For multi-process environments, TENT implements global time-sliced coordination using shared memory:
--- a/mooncake-transfer-engine/tent/include/tent/transport/rdma/promotion_policy.h
+++ b/mooncake-transfer-engine/tent/include/tent/transport/rdma/promotion_policy.h
@@ -54,8 +54,11 @@ inline PromotionDecision DecidePromotionHeadOnly(
     PromotionDecision d;
     if (enqueue_ts.empty()) return d;
     const uint64_t head = enqueue_ts.front();
-    const bool head_timed_out =
-        head > 0 && (current_ts - head) >= promotion_timeout_ns;
+    // Guard current_ts >= head before subtracting: both are unsigned, so a
+    // non-monotonic clock / race where current_ts < head would otherwise
+    // underflow and spuriously mark the entry timed out.
+    const bool head_timed_out = head > 0 && current_ts >= head &&
+                                (current_ts - head) >= promotion_timeout_ns;
     if (head_timed_out) {
         d.promote_indices.reserve(enqueue_ts.size());
         for (size_t i = 0; i < enqueue_ts.size(); ++i) {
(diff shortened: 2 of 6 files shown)


alogfans:
Looks good overall — keeping head-only as default while introducing per-entry as opt-in is a safe rollout strategy.

Suggestion: add config docs for transports/rdma/priority_promotion_per_entry (default + behavior difference).

catyans:
Thanks @alogfans! Addressed:

- **Config docs**: added a Promotion configuration section to `docs/design/tent/qos.md` documenting `transports/rdma/priority_promotion_per_entry` (default `false` = head-only "flush the tier", `true` = per-entry) alongside `priority_promotion_timeout_us`, with the behavior difference spelled out.
- **Underflow guard** (from the bot review): both `DecidePromotionHeadOnly` and `DecidePromotionPerEntry` now check `current_ts >= ts` before the unsigned subtraction, so a non-monotonic clock / race (current_ts < enqueue_ts) can't underflow and spuriously promote.

Also rebased onto main (the earlier conflict was a stale S3 commit that's now redundant with the merged #2778). Default remains byte-identical head-only.

alogfans (approved):
LGTM

PR #2790, opened 2026-07-08
[TENT] RailMonitor: prefer same-name device for cross-NUMA rail mapping (#2758/#2467)

catyans:
## Bug

`RailMonitor::updateBestMapping()` maps each local NIC to a remote NIC per
`(local_numa, remote_numa)` pair. For the **same-NUMA** case it uses
`direct_rails_`, which `loadDefault()` builds via **same-name matching**
(`mlx5_5 -> mlx5_5`, Priority 1). But for the **cross-NUMA** case it used
**positional** assignment:

```cpp
remote_nic = remote_devices[remote_numa][i % remote_cnt];  // ignores names
```

On a multi-bond dual-NUMA RoCEv2 fabric where the two nodes disagree on which
NUMA a same-named NIC sits in (e.g. an overlay NIC on a different physical
network), this maps a local NIC to an unrelated remote NIC. The QP then never
reaches RTR:

- #2467 — `transport retry counter exceeded` (initiator mlx5_5/NUMA1 → target
  mlx5_0/NUMA0 overlay instead of the expected same-name mlx5_5).
- #2758 — cross-node QP modify-to-RTR `EINVAL(22)`; errno diagnosis showed the
  peer QP simply does not exist on the wrongly-chosen device.

## Fix

In the cross-NUMA branch, prefer a **same-name** remote device (mirroring
`loadDefault()`'s Priority-1 matching) before falling back to positional
assignment. Same-NUMA behavior is unchanged; positional fallback still applies
when no same-name remote device exists in the target NUMA domain.

## Test

Adds `RailMonitorCrossNumaTest.CrossNumaPrefersSameNameDevice` — an asymmetric
NUMA layout (as in #2467) where positional and same-name matching diverge, and
asserts the same-name device is chosen across NUMA.

**Verification note:** the changed translation unit (`rail_monitor.cpp`)
compiles cleanly in-tree. I could not link/run the full `tent_rail_monitor_test`
binary in my environment due to an unrelated GDS/cuFile build dependency
(`cufile.h` absent); the gtest was therefore not executed locally. A reviewer
with the GDS deps (or CI) can run it.

## Attribution

The root cause and the same-name-vs-positional analysis are from @zhangzhiqiangcs's
#2467; I hit the same issue from #2758. I've offered co-authorship on #2467 —
happy to add attribution, rework, or **close this if @zhangzhiqiangcs is already
preparing a fix**.

Relates to #2467, #2758.

### DIFF
--- a/mooncake-transfer-engine/tent/src/transport/rdma/rail_monitor.cpp
+++ b/mooncake-transfer-engine/tent/src/transport/rdma/rail_monitor.cpp
@@ -323,10 +323,32 @@ void RailMonitor::updateBestMapping() {
             for (size_t i = 0; i < local_cnt; i++) {
                 int local_nic = local_devices[local_numa][i];
                 int remote_nic = -1;
-                if (local_numa == remote_numa)
+                if (local_numa == remote_numa) {
                     remote_nic = direct_rails_[local_nic];
-                else
-                    remote_nic = remote_devices[remote_numa][i % remote_cnt];
+                } else {
+                    // Cross-NUMA: prefer a same-name remote device (e.g.
+                    // mlx5_5 -> mlx5_5) before falling back to positional
+                    // assignment. loadDefault() already builds direct_rails_
+                    // via same-name matching (Priority 1); mirroring it here
+                    // avoids mapping a local NIC to an unrelated remote NIC on
+                    // a different physical/overlay network, which fails QP
+                    // modify-to-RTR with "transport retry counter exceeded" on
+                    // multi-bond dual-NUMA RoCEv2 fabrics (issues #2758/#2467).
+                    auto local_entry = local_->getNicEntry(local_nic);
+                    if (local_entry) {
+                        for (int cand : remote_devices[remote_numa]) {
+                            auto cand_entry = remote_->getNicEntry(cand);
+                            if (cand_entry &&
+                                cand_entry->name == local_entry->name) {
+                                remote_nic = cand;
+                                break;
+                            }
+                        }
+                    }
+                    if (remote_nic < 0)
+                        remote_nic =
+                            remote_devices[remote_numa][i % remote_cnt];
+                }
                 if (!available(local_nic, remote_nic)) {
                     bool found = false;
                     for (int cand : remote_devices[remote_numa]) {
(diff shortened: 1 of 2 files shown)


alogfans (approved):
LGTM

zhangzhiqiangcs:
Thanks for the fix — it covers the asymmetric case, but misses one topology: **symmetric NUMA** (same-named NIC in the *same* NUMA on both nodes).

  The cross-NUMA branch only looks for a same-name device *within `remote_devices[remote_numa]`*. In a symmetric topology the same-name peer lives in a
  different NUMA than the queried domain, so the search misses and falls back to positional — same failure mode as before the fix (mlx5_y → mlx5_x, QP never
  reaches RTR). Our fabric is symmetric (mlx5_5 reports NUMA1 on both nodes).

  **Reproduction** (using this PR's `makeNamedNumaTopology`):

  ```cpp
  TEST(RailMonitorSymmetricNumaTest, CrossNumaSameNameOutsideDomain) {
      auto local  = makeNamedNumaTopology("mlx5_x", 0, "mlx5_y", 1);
      auto remote = makeNamedNumaTopology("mlx5_x", 0, "mlx5_y", 1);  // symmetric
      RailMonitor rail;
      ASSERT_TRUE(rail.load(local.get(), remote.get()).ok());
      // local mlx5_y@NUMA1 → remote NUMA0 domain; same-name mlx5_y is at
      // idx1 (NUMA1), outside it
      EXPECT_EQ(rail.findBestRemoteDevice(/*local_nic=*/1, /*remote_numa=*/0), 1);
  }
  ```
  Returns `0` (mlx5_x) instead of `1` (mlx5_y) → fails.

  **Fix** — fall back to the global same-name map (`direct_rails_`) before positional:

  ```cpp
  // after the in-domain same-name loop, before positional:
  if (remote_nic < 0) {
      auto it = direct_rails_.find(local_nic);
      if (it != direct_rails_.end()) remote_nic = it->second;
  }
  if (remote_nic < 0)
      remote_nic = remote_devices[remote_numa][i % remote_cnt];
  ```

  Asymmetric behavior unchanged (in-domain match still wins); symmetric now resolves correctly.

PR #2793, opened 2026-07-08
[Bugfix][TransferEngine] Fix RDMA handshake thundering herd

dqxlzlz:
### The problem we hit in production

In a large-scale PD-disaggregated deployment, when many nodes start up and connect
at the same time, some QPs get stuck in a permanently degraded state: they report
`transport retry counter exceeded` and their throughput drops to ~0 and never
recovers.

We reproduced this on a real cluster. With 240 concurrent handshakes hitting a
single target (unpatched build):

- Handshake setup time ballooned to **~96 seconds** (normally < 1s).
- **53-71%** of the connections ended up with near-zero throughput.

This confirms the target's handshake RPC server gets overwhelmed by the connection
storm.

### Why this happens (root cause)

The handshake path has **no admission control**: every node fires all of its
handshake RPCs at once. When 40+ nodes x 8 NICs start simultaneously, hundreds of
handshake requests pile up in the target's RPC queue. Meanwhile the RDMA QP has a
fixed hardware retry window (`kTimeout=14` ~67ms x `kRetryCount=7` ~= **470ms**).

The mismatch is fatal: the RPC queue latency (tens of seconds) far exceeds the QP's
470ms retry window. Before the handshake response arrives, the QP's hardware retry
counter is exhausted, the QP transitions to `ERROR`, and it never recovers -
throughput stays at 0.

### What this PR does

A two-layer defense:

1. **Connection Limiter** (per-transport counting semaphore): limits concurrent
   outbound handshakes so a single instance can't flood the target. Configurable via
   the `MC_MAX_CONCURRENT_HANDSHAKES` environment variable (default: 16).
2. **Exponential Backoff**: when a handshake fails, subsequent retries back off
   exponentially (100ms -> 200ms -> ... -> 5s cap) in the worker pool, giving the
   overwhelmed target room to drain its queue and recover.

After the fix, the same 240-concurrent test brings handshake setup time back down
from ~96s to **< 1s**, keeping it well under the 470ms hardware retry window.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

### Unit tests

Added `connection_limiter_test.cpp` (behavioral tests for the semaphore) and
extended `config_test.cpp` (env-var parsing for `MC_MAX_CONCURRENT_HANDSHAKES`).

```bash
cmake -DBUILD_UNIT_TESTS=ON /path/to/build && cmake --build build \
  --target connection_limiter_test config_test
ctest -R "connection_limiter_test|config_test"
```

- `ConnectionLimiter` never exceeds the configured max concurrent slots under
  32-thread contention.
- `acquire()` blocks when full and unblocks on `release()`.
- `MC_MAX_CONCURRENT_HANDSHAKES` parsing: default 16, valid override applied,
  zero/negative/non-numeric values rejected.

All unit tests pass.

### Cluster stress test (thundering herd)

Tested on a 4-node cluster, 8x mlx5_bond NICs per node (32 NICs total), with a
target node under 240 concurrent handshake connections from 3 initiator nodes.

**Test commands:**
```bash
# Target (with fix, limit=4):
MC_MAX_CONCURRENT_HANDSHAKES=4 transfer_engine_bench --mode=target \
  --metadata_server=http://<target>:8765/metadata \
  --local_server_name=<target>:12345 --device_name=mlx5_bond_0

# 3 remote nodes x 80 initiators = 240 concurrent handshakes
```

**Observed results (directly measured):**

| Scenario | Handshake setup time | Near-zero-throughput procs |
|----------|----------------------|----------------------------|
| Before fix (240 concurrent) | ~96 seconds | 53-71% |
| After fix (240 concurrent)  | < 1 second   | N/A (only NIC bandwidth contention) |

Note: the cluster repro directly reproduces the handshake queueing delay
(~96s >> the 470ms QP retry window) and the resulting throughput degradation.
The permanent QP `ERROR` transition is the downstream production symptom
explained in the root-cause section; it was not separately captured as a log
line in this fresh-connection benchmark.

- [x] Unit tests pass
- [x] Manual testing done (4-node cluster, 240 concurrent connections)

## Checklist

- [x] For changes >500 LOC: I have filed an RFC issue (N/A: ~140 LOC)

## AI Assistance Disclosure

An AI coding assistant helped implement the fix and orchestrate the multi-node
stress tests across the 4-node cluster. The human submitter has reviewed every


mooncake-transfer-engine/src/transport/rdma_transport/worker_pool.cpp
```diff
@@ -356,14 +356,28 @@ void WorkerPool::performPostSend(int thread_id) {
...
-            entry.second.clear();
-            continue;
+        if (!endpoint->connected()) {
+            // Check handshake backoff before attempting connection
+            if (isInHandshakeBackoff(entry.first)) {
+                for (auto &slice : entry.second)
+                    failed_slice_list.push_back(slice);
+                entry.second.clear();
+                continue;
+            }
```
alogfans, line 474: The current backoff branch may incorrectly consume the slice retry budget and cause busy-waiting.

mooncake-transfer-engine/src/transport/rdma_transport/rdma_endpoint.cpp
```diff
@@ -395,8 +396,11 @@ int RdmaEndPoint::setupConnectionsByActive() {
 
         // Perform the RPC without holding the lock to avoid deadlock and allow
         // "simultaneous open" handshake handling.
+        // Acquire a handshake slot to limit concurrent connection storms.
+        context_.engine().acquireHandshakeSlot();
         int rc = context_.engine().sendHandshake(peer_server_name, local_desc,
                                                  peer_desc);
+        context_.engine().releaseHandshakeSlot();
```
alogfans: A RAII guard is better.

dqxlzlz: Thanks for the review! Addressed in 2820d1bd, merged latest main in b5cd622c:

    - Backoff slices are redispatched without consuming retry budget (+ short sleep, no busy-wait)
    - Handshake slot now managed by an RAII guard

PR #2795, opened 2026-07-08
[TE] Add metadata refresh polling for segment cache

### LINKED ISSUE #2704: [Bug]: Unhealthy store segment remains available and causes SGLang requests to fail (draft)
## Bug Report

We are using Mooncake Store as the storage backend for SGLang in a Kubernetes deployment with RDMA enabled.

Recently, SGLang failed when accessing Mooncake Store through the Mooncake Store client. After investigation, the failure seemed to be related to only one store service. Other store services looked normal.

The problematic part is that after this store service became abnormal, the affected segment still seemed to be available in Mooncake Store metadata. As a result, requests involving this unhealthy store service continued to fail.

On the problematic store machine, `dmesg` kept printing RDMA-related QP local protection errors. We suspect that this store service or its underlying RDMA transport entered an abnormal state, but Mooncake Store did not exclude the affected segment from later requests.

## Observed Behavior

* We have multiple Mooncake Store services in the cluster.
* Only one store service was affected by this issue.
* Other store services looked normal.
* Requests involving the affected store service kept failing.
* The affected segment still seemed to be available in Mooncake Store metadata.
* The issue cannot be reproduced stably at the moment.
* The problematic store service did not recover automatically.

We also checked the restart history of this store service. It had restarted several times before without causing this issue. For the latest restart, Kubernetes events showed that it was killed under node-level memory pressure, likely because other containers on the same node had high memory usage. We mention this only as context, not as a confirmed root cause.

## Kernel Logs

On the problematic store machine, `dmesg` repeatedly printed logs like the following:

```text
Jun 30 15:01:31 maas-h20-a42h4kds-0001 kernel: mlx5_bond_0/1: QP 118513 error: local protection error (0x3a 0x0 0x95)
Jun 30 15:01:31 maas-h20-a42h4kds-0001 kernel: mlx5_bond_1/1: QP 122226 error: local protection error (0x3a 0x0 0x95)
Jun 30 15:01:31 maas-h20-a42h4kds-0001 kernel: mlx5_bond_1/1: QP 122227 error: local protection error (0x3a 0x0 0x95)
Jun 30 15:01:31 maas-h20-a42h4kds-0001 kernel: mlx5_bond_0/1: QP 118514 error: local protection error (0x3a 0x0 0x95)
```

The same pattern kept being printed continuously in `dmesg`.

## Expected Behavior

When a store service becomes abnormal and can no longer serve requests correctly, Mooncake Store should avoid using the affected store service or segment for later requests.

More specifically, we expect Mooncake Store to:

* detect that the affected store service or segment is no longer usable;
* mark the affected segment as unavailable, or remove it from the available metadata;
* avoid routing later requests to the unhealthy store service;
* expose clear logs or metrics so users can diagnose this state.

## Actual Behavior

The affected store service / segment still seemed to be considered available after it entered an abnormal state.

As a result, requests involving this unhealthy store service continued to fail.

## Request

Could you help check whether Mooncake Store correctly handles this kind of partial store failure?

It would be helpful if Mooncake Store could invalidate or exclude the affected segment when a store service is still present but no longer able to serve RDMA requests correctly.

### Before submitting...

greatwhole:
## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
```

**Test results:**
- [ ] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

Related Issue

Fixes #2704


alogfans:
LGTM. Some advises: 1. Pls fix format issues using scripts/code_format.sh; 2. Some users report that metadata caching refresh may take longer time in production. So it should be considered.

greatwhole:
> LGTM. Some advises: 1. Pls fix format issues using scripts/code_format.sh; 2. Some users report that metadata caching refresh may take longer time in production. So it should be considered.

Fixed the formatting with scripts/code_format.sh. Refreshes run serially, and we now log each refresh duration (duration_ms) to help tune the polling interval in production.

mooncake-transfer-engine/tests/transport_uint_test.cpp
alogfans, line 1: It seems that some CI tests failed:

```
E0711 14:22:08.358328 16305 common.h:372] Socket read failed: Bad file descriptor [9]
[       OK ] TransportTest.ReadInvalidFD (0 ms)
[ RUN      ] TransportTest.PartialRead
[       OK ] TransportTest.PartialRead (0 ms)
[ RUN      ] TransportTest.ReadEmptyFile
W0711 14:22:08.358472 16305 common.h:375] Socket read incompleted: expected 256 bytes, actual 0 bytes
[       OK ] TransportTest.ReadEmptyFile (0 ms)
[ RUN      ] TransportTest.ReadWouldBlockReturnsImmediately
/home/runner/work/Mooncake/Mooncake/mooncake-transfer-engine/tests/transport_uint_test.cpp:200: Failure
  bytesRead
    Which is: 1
  -1
/home/runner/work/Mooncake/Mooncake/mooncake-transfer-engine/tests/transport_uint_test.cpp:202: Failure
Expected: (elapsed) < (std::chrono::milliseconds(100)), actual: 8-byte object <D4-81 ED-0B 00-00 00-00> vs 8-byte object <64-00 00-00 00-00 00-00>
[  FAILED  ] TransportTest.ReadWouldBlockReturnsImmediately (201 ms)
[----------] 9 tests from TransportTest (201 ms total)
```

alogfans:
This PR is LGTM. Is it ready to merge? @greatwhole

greatwhole:
> This PR is LGTM. Is it ready to merge? @greatwhole

Yes, it’s ready to merge. Thanks for the review! @alogfans

alogfans (approved):

PR #2797, opened 2026-07-08
Add fragmentation-aware allocation strategy for Mooncake Store

Lorry1024:
## Summary

This PR adds an opt-in Mooncake Store allocation strategy named `fragmentation_aware` for mixed-size KVCache workloads. The strategy keeps bounded candidate sampling, but ranks sampled segments by whether their largest contiguous free region can satisfy the current request before considering aggregate free ratio.

## Why

In long-running Store pools, aggregate free space can be split into fragmented holes. A segment may report a higher total free ratio while failing a large allocation because no contiguous region is large enough. This creates avoidable allocation attempts and fallback pressure.

## Changes

- Add `AllocationStrategyType::FRAGMENTATION_AWARE`.
- Add `FragmentationAwareAllocationStrategy`.
- Wire `--allocation_strategy=fragmentation_aware`.
- Add deterministic fragmentation and preferred-segment tests.
- Add the strategy to `allocation_strategy_bench`.
- Document design and deployment usage.

## Validation

- `git diff --check`
- `git apply --check` against current upstream
- deterministic fragmentation simulation
- extended ranking and boundary simulation
- topic-aligned Store scalability simulation

## Boundaries

The default allocation strategy remains unchanged. This PR does not redesign SGLang HiCache, RDMA transport, or Mooncake HA. Full upstream CI, RDMA validation, and real SGLang HiCache benchmark should still be run in a production-like environment before claiming end-to-end throughput gains.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -167,6 +167,15 @@ jobs:
         kill $MASTER_PID 2>/dev/null || true
       shell: bash
 
+    - name: Reclaim disk before Go store binding integration tests
+      run: |
+        df -h
+        rm -rf mooncake-store/rust/target
+        rm -rf ~/.cargo/registry ~/.cargo/git
+        rm -rf ~/.cache/sccache || true
+        df -h
+      shell: bash
+
     - name: Run Go store binding integration tests
       run: |
         $GITHUB_WORKSPACE/build/mooncake-store/src/mooncake_master \
--- a/docs/source/deployment/mooncake-store-deployment-guide.md
+++ b/docs/source/deployment/mooncake-store-deployment-guide.md
@@ -461,7 +461,7 @@ mooncake_master \
 
 | Flag | Default | Description |
 |------|---------|-------------|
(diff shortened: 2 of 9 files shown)


yokinoshitayoki:
Thanks for the proposal. One concern is that production workloads usually have only one or two object sizes, so the object-size distribution is much narrower than the mixed-size workload assumed here.

Given that, fragmentation pressure may be limited in practice, so maybe it's not necessary to add and maintain a separate `fragmentation_aware` allocation strategy for now.

yokinoshitayoki (commented):
Thanks for continuing to work on this, but I don’t think we should merge this PR as-is.

My main concern from the previous comment still stands: our production workloads typically use only one or two object sizes, so the mixed-size fragmentation scenario targeted by `fragmentation_aware` does not match our current production workload well.

Also, the CI disk cleanup change seems unrelated to the allocation strategy itself and should not be bundled into this PR.

Given the current evidence, I’d prefer not to add another allocation strategy. If we want to revisit this later, I think we may need to first provide representative production traces or benchmarks showing a concrete fragmentation problem under our workload.

PR #2800, opened 2026-07-08
[Store] Add a paged-KV prefix-cache connector for the Go bindings

Lfan-ke:
## What this adds

`package pagedkv` — a Go connector that lets an inference runtime holding its KV cache in fixed-size token pages publish that KV to Mooncake Store and reuse what a peer already published, so instances that share a prompt prefix do not each recompute it. It builds on the existing Go store bindings in `mooncake-store/go/mooncakestore` and plays the same role for Go that `MooncakeStoreConnector` plays for vLLM.

## Keying

Page identity is a hash chain over fixed-size token pages, seeded with the model name:

```
seed    = H("mooncake-paged-kv" || model)
hash[i] = H(hash[i-1] || tokens[page i])
key[i]  = "<prefix>/<model>/<hex(hash[i])>"
```

Workers therefore agree on keys without coordination; a shared token prefix maps to identical keys, and any earlier divergence changes every later key, so a worker can never load KV that does not match its own tokens. Folding the model name into the seed keeps caches for different models disjoint under a shared store. This follows the hash-based prefix-caching pattern the vLLM `MooncakeStoreConnector` uses to share KV blocks across instances.

## API

```go
// reuse
res, _ := c.LoadCachedPrefix(tokens, pageDst)   // runtime computes only tokens[res.MatchedTokens:]
// publish
n, _ := c.StoreComputedPrefix(tokens, pageSrc)  // skips pages a peer already published
```

Plus `RegisterKVBuffer` / `UnregisterKVBuffer` for the staging region, and a `Role` (`mixed` / `prefill` / `decode`) gating who may publish.

## Scope, stated plainly

This is the store-transfer and keying half only, and it is tied to no particular runtime. Moving bytes between these page buffers and a runtime's own KV tensors stays the runtime's job and is **not** provided here. A runtime can drive this package when it

- keeps KV cache in fixed-size token pages,
- can expose each page as a pointer into memory registered via `RegisterKVBuffer`, and
- derives page identity from stable token ids.

Zero-copy RDMA paths need the full Mooncake build and RDMA hardware; the TCP transport is enough for functional testing.

## Layout

| Path | Purpose |
|:--:|:--:|
| `mooncake-store/go/pagedkv/prefix/` | Pure-Go content-addressed page keying. No CGo, unit-tested standalone. |
| `mooncake-store/go/pagedkv/config.go` | `Role`, `PageLayout`, `Config`. |
| `mooncake-store/go/pagedkv/connector.go` | `Connector` and the load/store/register calls. |
| `mooncake-store/go/examples/pagedkv/` | Runnable end-to-end demo over the TCP transport. |

## Checks

`go build ./pagedkv/...`, `go vet ./pagedkv/... ./examples/pagedkv/`, `go test ./pagedkv/...` and `gofmt -l` are all clean.

## History

This PR previously proposed itself as an Ollama integration. Ollama has since turned out not to own a KV cache at all — `9db4bdba` (#16031) removed its CGO engines and Go-native runner and `7b22ac96` (#17007) deleted its `kvcache` package, so GGML now runs in an out-of-process upstream `llama-server` driven over HTTP. Nothing in this code was ever Ollama-specific, so the package has been renamed and re-scoped to what it actually is, and the documentation page claiming an integration has been dropped.

### DIFF
--- a/docs/source/getting_started/examples/ollama-integration/index.md
+++ b/docs/source/getting_started/examples/ollama-integration/index.md
@@ -1,0 +1,187 @@
+# Mooncake x Ollama Integration
+
+## Overview
+
+[Ollama](https://github.com/ollama/ollama) is a Go runtime for running local LLMs
+on top of a `llama.cpp`/GGML backend. By default each Ollama instance keeps its
+KV cache only in its own process, so a prompt prefix computed on one node cannot
+be reused by another, and Ollama cannot participate in a
+prefill/decode-disaggregated (PD) deployment.
+
+This integration adds an Ollama-side **connector** that offloads and reuses KV
+cache through **Mooncake Store**, following the same hash-based prefix-caching
+pattern already used by the vLLM `MooncakeStoreConnector` and the SGLang HiCache
+L3 backend:
+
+- A **prefill** worker publishes the KV cache it computes for a prompt.
+- A **decode** worker (or any peer) reuses the longest cached prefix from the
+  store instead of recomputing it, cutting redundant prefill and lowering TTFT
+  for workloads with shared prefixes — system prompts, few-shot templates, and
+  multi-turn conversations.
+
+The connector lives in the Go store bindings at
+[`mooncake-store/go/ollama`](https://github.com/kvcache-ai/Mooncake/tree/main/mooncake-store/go/ollama)
+and builds directly on the existing Go `mooncakestore` client, so it reuses the
+same RDMA-capable, zero-copy `put`/`get` path as the other integrations.
+
+| Scenario | Store API used | Role |
+|----------|----------------|------|
+| Publish computed KV (prefill) | `BatchPutFrom` (zero-copy) | `prefill` / `mixed` |
+| Reuse cached prefix (decode) | `BatchExists` + `BatchGetInto` | `decode` / `mixed` |
+
+## Architecture
+
+### Page identity
+
+KV cache is paged: a request's tokens are split into fixed-size pages of
+`PageTokens` tokens (the trailing partial page is not cacheable). Each page is
+addressed by a key that chains SHA-256 over the pages seen so far, seeded with
+the model name:
+
+```
+seed    = H("mooncake-ollama-kv" || model)
+hash[i] = H(hash[i-1] || tokens[page i])
+key[i]  = "mc/ollama/<model>/<hex(hash[i])>"
+```
+
+Two properties fall out of the chain and matter for correctness:
+
+- **Shared prefixes collide on purpose.** Requests that share a token prefix
+  produce identical keys for the shared pages, so any worker can reuse them.
+- **Divergence is total.** Changing any earlier token changes every later key,
+  so a decode worker can never load KV that does not match its own tokens.
+
+Folding the model name into the seed keeps different models disjoint even in a
+shared store, and the `mc/ollama/` namespace keeps Ollama keys from colliding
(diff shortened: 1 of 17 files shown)


Lfan-ke:
To make this connector easy to try during review, I added a small self-contained demo that exercises it end to end on a single machine with **no RDMA**.

- **Live page:** https://lfan-ke.github.io/mooncake/ (may take a few minutes to publish)
- **Source:** https://github.com/Lfan-ke/mooncake/tree/demo/pd-disaggregation/demo — reproduce with `bash demo/run_demo.sh`

It starts a real Mooncake Store (`mooncake_master` + in-process HTTP metadata server over the **TCP transport**), has a prefill worker publish a prompt's KV pages with put-if-absent, then has a second request that shares the prompt's prefix reload the cached pages and **skip prefill** for the shared tokens. The page keys are produced by this PR's real `ollama/prefix` keying; the store round-trip uses `batch_is_exist` / `batch_put_from` / `batch_get_into` on registered buffers.

Reference run (realistic GQA layout, 2048-token prompt with a 1792-token shared prefix):

- **7/8 KV pages (1792 tokens, 87.5%)** served from the store on the second request
- **98 MB of KV reloaded over TCP in ~22 ms** (p50, 20 iterations), ~4.3 GB/s
- loaded bytes **verified byte-exact** against what prefill published, on both this layout and the tiny layout from the example in this PR

Scope, kept explicit in `demo/README.md`: the store and transport are real and the keying is this PR's code, but the orchestration is driven from Python mirroring `connector.go`'s `StoreComputedPrefix` / `LoadCachedPrefix` rather than the literal Go binary (the CGo bindings need the C++ store compiled), the KV bytes are deterministic stand-ins for llama.cpp KV tensors, and end-to-end TTFT needs a running model — the same runtime-side follow-ups already noted in the PR description.

ykwd:
Thanks for the contribution! Could you share how you envision this being used in practice? Since upstream Ollama doesn't currently support PD-disaggregated inference, I am curious about the intended integration path.

Lfan-ke:
@ykwd Thanks for pushing on this — the answer turned out to be worse than "Ollama doesn't support PD yet", so I've re-scoped the PR.

Ollama no longer owns a KV cache at all. `9db4bdba` (2026-05-29, ollama/ollama#16031) removed the CGO engines and the Go-native `ollamarunner`, and `7b22ac96` (2026-07-02, ollama/ollama#17007) deleted the `kvcache` package. GGML models now run in an out-of-process upstream `llama-server` that Ollama drives over HTTP, so there are no KV tensors in Ollama's Go process to wire page buffers to. The runtime-side half this PR pointed at cannot exist as an Ollama change; it would have to be llama.cpp work behind `llama_state_seq_get_data`/`set_data` or the `/slots/{id}?action=save|restore` endpoint — and those are whole-sequence and file-shaped, not paged and prefix-hashed, so they do not map onto this design.

So I've dropped the Ollama framing entirely. Nothing in the code was Ollama-specific — it is tokens plus page pointers plus a role — so the package is now `pagedkv`: a runtime-agnostic connector for any runtime that keeps KV in fixed-size pages in registered memory, playing the same role for Go that `MooncakeStoreConnector` plays for vLLM. I also dropped the docs integration page, since there is no integration to document.

If a runtime-agnostic Go binding isn't something you want in-tree without a first consumer, I'm happy to close this instead — that's a reasonable call and I won't argue it.

PR #2803, opened 2026-07-08
[Bugfix] Skip os.chmod when binary is already readable and executable

### LINKED ISSUE #2802: [Bug]: mooncake_master fails with PermissionError when installed as root and run by non-root user
### Bug Report

After installing the wheel as root (e.g., `sudo pip install mooncake-transfer-engine`), running `mooncake_master` as a non-root user fails immediately:

```
$ mooncake_master --port 50051
Traceback (most recent call last):
  File "/usr/local/python3.12.13/bin/mooncake_master", line 6, in <module>
    sys.exit(main())
             ^^^^^^
  File "/usr/local/python3.12.13/lib/python3.12/site-packages/mooncake/cli.py", line 21, in main
    os.chmod(bin_path, 0o755)
PermissionError: [Errno 1] Operation not permitted: '/usr/local/python3.12.13/lib/python3.12/site-packages/mooncake/mooncake_master'
```

But the binary mooncake_master  already has full execute permissions (`-rwxrwxrwx`), so the chmod is redundant in this case. The same issue affects `mooncake_client` and `transfer_engine_bench`.

### Before submitting...

Csrayz:
## Description

Fix #2802. Fix a PermissionError when running mooncake_master as a non-root user on a wheel installed as root. cli.py unconditionally calls os.chmod(bin_path, 0o755) on every invocation. This fails when:

1. The wheel was installed as root (so the binary is owned by root), but a non-root user runs it.  `os.chmod` requires owner or root privilege. The result is a PermissionError that crashes the CLI before the actual binary is even reached.
2. The file already has full read+execute permissions (-rwxrwxrwx), so the chmod is entirely unnecessary.

This PR adds an os.access pre-check and only runs chmod when the file is missing either read or execute permission.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

manual testing

## Checklist

## AI Assistance Disclosure

- [x] No AI tools were used

### DIFF
--- a/mooncake-wheel/mooncake/cli.py
+++ b/mooncake-wheel/mooncake/cli.py
@@ -4,6 +4,7 @@ Minimal CLI module for mooncake_master.
 """
 
 import os
+import stat
 import sys
 import subprocess
 
@@ -18,8 +19,9 @@ def main():
     bin_path = os.path.join(package_dir, "mooncake_master")
 
     # Make sure the binary is executable
-    if not os.access(bin_path, os.R_OK | os.X_OK):
-        os.chmod(bin_path, 0o755)
+    if not os.access(bin_path, os.X_OK):
+        st = os.stat(bin_path)
+        os.chmod(bin_path, st.st_mode | stat.S_IXUSR | stat.S_IXGRP | stat.S_IXOTH)
 
     # Run the binary with all arguments passed through
     return subprocess.call([bin_path] + sys.argv[1:])
--- a/mooncake-wheel/mooncake/cli.py
+++ b/mooncake-wheel/mooncake/cli.py
@@ -16,10 +16,11 @@ def main():
     # Get the path to the mooncake_master binary
     package_dir = os.path.dirname(os.path.abspath(__file__))
     bin_path = os.path.join(package_dir, "mooncake_master")
-    
+
     # Make sure the binary is executable
-    os.chmod(bin_path, 0o755)
-    
+    if not os.access(bin_path, os.R_OK | os.X_OK):
+        os.chmod(bin_path, 0o755)
+
     # Run the binary with all arguments passed through
     return subprocess.call([bin_path] + sys.argv[1:])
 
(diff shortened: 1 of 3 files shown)


Icedcoco:
This looks like it fixes #2802. One small question: do we need to include `os.R_OK` in the access check here, or would `os.X_OK` be enough? The wrapper only executes the binary via `subprocess.call()`, so read permission does not seem required for the normal path. If the binary is executable but not readable and still root-owned, `os.R_OK | os.X_OK` would still fall through to `chmod` and may hit the same `PermissionError`. What do you think about narrowing this to `os.X_OK`?

Also, the PR description says `Fix #2800`, but I think this should be `Fix #2802`.

Csrayz:
> This looks like it fixes #2802. One small question: do we need to include `os.R_OK` in the access check here, or would `os.X_OK` be enough? The wrapper only executes the binary via `subprocess.call()`, so read permission does not seem required for the normal path. If the binary is executable but not readable and still root-owned, `os.R_OK | os.X_OK` would still fall through to `chmod` and may hit the same `PermissionError`. What do you think about narrowing this to `os.X_OK`?

I agree os.X_OK should be enough for subprocess.call(). I initially kept R_OK just to be conservative, given that the code does os.chmod(binary, 0o755).

I verified locally that a binary with only execute permission can be invoked successfully. I'll update the PR to check only os.X_OK.

> Also, the PR description says `Fix #2800`, but I think this should be `Fix #2802`.

Fixed.

Csrayz:
I also improved the chmod part, it now appends execute bits to the existing mode instead of forcing 0o755. This aligns with the narrowed access check. New commit is pushed.

Icedcoco (approved):
LGTM. This fixes the root-owned wheel install case from #2802 by avoiding an unnecessary `chmod` when the packaged binary is already executable. The updated `os.X_OK` check matches the wrapper’s actual requirement, and preserving the existing mode while adding execute bits is better than forcing `0755`.

Thanks for the update.

Issue #2662, opened 2026-06-29
[RFC]: Add TPU Transport Support via PJRT


Liwink:
## 1. Summary

We propose adding **TPU** as a first-class accelerator target in Mooncake's Transfer Engine (TENT). The implementation introduces a `TpuTransport` class that enables KVCache disaggregation between TPU-based Prefill and Decode nodes, using a **staged host-DRAM pipeline** over existing TENT transports (RDMA / TCP) — without requiring any proprietary TPU SDK at build time.

The key architectural insight is that TPU KVCache transfer decomposes into two independent halves that map cleanly onto TENT's existing capabilities:

| Stage | Mechanism | Owner |
|-------|-----------|-------|
| TPU HBM ↔ Host DRAM | PJRT async copy (`BufferToHostBuffer` / `CopyToHostAsync`) | New `TpuTransport` |
| Host DRAM ↔ Remote Host DRAM | Existing TENT RDMA/TCP transport | Existing TENT infra |

This decomposition avoids flooding the TPU's Inter-Chip Interconnect (ICI) with asynchronous KVCache traffic, preserving ICI bandwidth for synchronous model-parallel collectives.

**Estimated scope:** ~800–1200 LOC (excluding tests).

---

### 2.1 The Problem

Google Cloud TPUs (v4, v5e, v5p, v6e) are a major LLM serving platform — powering production workloads on JAX/MaxText, PyTorch/XLA, and vLLM-TPU. However, **Mooncake currently has zero TPU support**, despite supporting 8+ other accelerator platforms (NVIDIA CUDA, AMD ROCm, Huawei Ascend, Cambricon MLU, Moore Threads MUSA, MetaX MACA, Hygon DCU, Iluvatar CoreX).

This gap prevents the entire Google Cloud TPU ecosystem from benefiting from Mooncake's KVCache disaggregation, hierarchical storage, and P2P distribution capabilities.

### 2.2 Why Not GPUDirect RDMA?

Unlike NVIDIA GPUs, TPUs do not expose third-party NIC access to device HBM:

| Capability | NVIDIA GPU | Google TPU |
|-----------|-----------|-----------|
| Direct NIC→VRAM DMA | ✅ GPUDirect RDMA | ❌ Not available |
| Memory registration API | `cuMemAlloc` + `ibv_reg_mr` | Opaque `PjRtBuffer` |
| Host driver model | Open CUDA toolkit | `libtpu.so` (closed-source runtime) |

[...]

PR #2030, opened 2026-05-04
[TransferEngine] Add TPU integration scaffolding and guards

stmatengss:
## Module

- [ ] Transfer Engine (`mooncake-transfer-engine`)
- [ ] Mooncake Store (`mooncake-store`)
- [ ] Mooncake EP (`mooncake-ep`)
- [ ] Integration (`mooncake-integration`)
- [ ] P2P Store (`mooncake-p2p-store`)
- [ ] Python Wheel (`mooncake-wheel`)
- [ ] PyTorch Backend (`mooncake-pg`)
- [ ] Mooncake RL (`mooncake-rl`)
- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other

## Checklist

- [ ] I have performed a self-review of my own code.
- [ ] I have formatted my own code using `./scripts/code_format.sh` before submitting.
- [ ] I have updated the documentation.
- [ ] I have added tests to prove my changes are effective.

PR #2815, opened 2026-07-09
[Bugfix][TENT] Fix silent TPU data corruption for transfers larger than one staging chunk

Liwink:
## Description

Fixes a silent data-corruption bug in the TENT TPU staging path added by #2733 (RFC #2662).

**Found by running the TPU path on a real TPU VM for the first time** (`v5p-8`, `libtpu 0.0.32`, PJRT C API 0.83). Everything before this had only ever run against the mock adapter.

### The bug

`ProxyManager` stages a transfer in `chunk_size` (4 MiB) pieces and passes `token + chunk_offset` to the platform for every chunk after the first:

```cpp
// mooncake-transfer-engine/tent/src/runtime/proxy_manager.cpp:76
local_stage.source = (uint8_t*)request.source + offset;
```

The adapter ABI only ever specified *base-address* classification. So for `offset > 0`:

1. `mc_tpu_pjrt_is_device_ptr(token + offset)` returns 0,
2. `TpuPlatform::copy` concludes neither side is device memory,
3. it falls through to `CpuPlatform::copy` — a plain `memcpy` of the device token.

On real PJRT that `memcpy` **does not crash**. `PJRT_Buffer_UnsafePointer` returns an address that is host-readable but does not hold the buffer's data. The staging copy produced garbage and reported `COMPLETED`.

Net effect: chunk 0 is correct, every later chunk is silently wrong. **Any TPU transfer larger than 4 MiB was corrupted, with no error raised anywhere** — which is essentially every real KV-cache transfer.

### Why CI never caught it

Two independent reasons, both fixed here:

- `-DUSE_TPU=ON` compiled **zero** TPU code. All TPU sources live under `tent/`, which is gated on `USE_TENT`. The build configured and succeeded while building nothing.
- The mock adapter's "device" pointers were ordinary host memory, so the accidental `memcpy` produced the *right* bytes. The mock structurally could not reproduce the failure.

### Changes

- **`tpu_pjrt_abi.h` / `tpu_pjrt_shim.h`** — specify that classification and copy entrypoints must resolve an interior address to the registered buffer whose range contains it; that a copy may not run past a buffer's end; and that the token must never be dereferenced.
- **`tpu_transport.cpp`** — require exactly one TPU-device side per staging hop, and fail loudly otherwise. Defense in depth: a non-conforming *or absent* adapter can no longer degrade into a silent `memcpy`.
- **`mock_tpu_pjrt_adapter.cpp`** — model the two properties that actually matter on hardware: tokens are opaque (a poisoned read-only mapping; data lives in shadow storage) and addresses may be interior. A copy that bypasses the adapter now yields `0xDD` instead of accidentally passing.
- **`common.cmake`** — `USE_TPU=ON` without `USE_TENT=ON` is now a hard error instead of a silent no-op.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

Note: the `common.cmake` guard is technically build-breaking for anyone passing `-DUSE_TPU=ON` alone — but such a build produced no TPU code, so nothing can regress.

## How Has This Been Tested?

**Test commands:**
```bash
cmake -S . -B build -G Ninja -DUSE_TENT=ON -DUSE_TPU=ON \
  -DWITH_STORE=OFF -DWITH_STORE_RUST=OFF -DBUILD_BENCHMARK=OFF -DUSE_RDMA=ON
cmake --build build -j
ctest --test-dir build/mooncake-transfer-engine/tent
```

**Test results:**

- `100% tests passed, 0 tests failed out of 22`
- New: `tent_tpu_transport_test` (6 cases) drives `TpuTransport`'s staging hop exactly as `ProxyManager` issues it — local stage at an interior device offset, the mirrored remote stage, the READ direction, and the three failure modes (no device side, two device sides, non-local target).
- New: `ClassifiesInteriorDevicePointers` and `CopyFromInteriorOffsetMovesTheRightBytes` in `tent_tpu_pjrt_shim_test`.

**Verified the tests actually catch the bug.** Reverting the fixes (exact-match registry + no transport guard) makes `LocalStageCopiesFromInteriorDeviceOffset` fail exactly as the hardware behaves: status `COMPLETED`, staging buffer filled with `0xDD`. Re-applying the fixes turns it green.

Hardware findings behind this were reproduced with a standalone PJRT probe (`GetPjrtApi` from `libtpu.so` → `BufferFromHostBuffer` → `UnsafePointer` → `memcpy`), confirming the token is readable and holds unrelated bytes.

- [x] Unit tests pass
- [x] Manual testing done (real TPU VM, v5p-8)

Not covered: the cross-node hop. This VM is single-host (gVNIC, no RDMA), so the device↔host staging hop is exercised, not host↔remote-host.

## Checklist

- [ ] I have run `pre-commit run --all-files` and all hooks pass — see note below

On `pre-commit`: every hook passes on the touched files except `cmake-format`, which rewrites large amounts of *unrelated* code in `tent/tests/CMakeLists.txt` and `common.cmake`. Per `AGENTS.md` ("do not include those unrelated edits in the PR") I kept the diff minimal and left those rewrites out; happy to let `pre-commit.ci` autofix, or to land them separately.

## Related work

- Implements the correctness half of the follow-ups in #2662; the merged feature is #2733.
- Does **not** overlap open PR #2030 (`[TransferEngine] Add TPU integration scaffolding and guards`), which targets the *classic* transfer engine (`mooncake-transfer-engine/src`, `include/gpu_vendor/tpu.h`). That PR introduces its own `USE_TPU` option. If it lands, the `USE_TPU ⇒ USE_TENT` guard here needs revisiting, since `USE_TPU` would then also gate classic-TE code. Flagging for whoever merges second.

Separately, hardware benchmarking turned up an architectural issue — the PJRT sub-range copy entrypoints that `Platform::copy` requires run ~20× slower than the whole-buffer ones (0.8 GB/s vs 18.5 GB/s D2H), capping the staging design near ~1 GB/s per chip. That is a design discussion, not a bugfix, so I'll raise it on #2662 rather than here.

## AI Assistance Disclosure

- [ ] No AI tools were used

Claude Code was used to run the TPU-VM investigation (building the PJRT probes that identified the token/interior-pointer behaviour), to implement the fixes, and to draft the tests. I have reviewed every changed line and can defend the change end to end.

### DIFF
--- a/mooncake-common/common.cmake
+++ b/mooncake-common/common.cmake
@@ -207,6 +207,15 @@ if(USE_CUDA)
 endif()
 
 if(USE_TPU)
+  # Every TPU source file lives under mooncake-transfer-engine/tent, which is
+  # only added when USE_TENT is ON. Without this guard -DUSE_TPU=ON configures
+  # and builds cleanly while compiling no TPU code at all.
+  if(NOT USE_TENT)
+    message(
+      FATAL_ERROR
+        "USE_TPU=ON requires USE_TENT=ON: all TPU support lives in TENT. Re-run cmake with -DUSE_TENT=ON."
+    )
+  endif()
   add_compile_definitions(USE_TPU)
   message(STATUS "TPU (PJRT) staging support is enabled")
 endif()
--- a/mooncake-transfer-engine/tent/include/tent/platform/tpu_pjrt_abi.h
+++ b/mooncake-transfer-engine/tent/include/tent/platform/tpu_pjrt_abi.h
@@ -24,6 +24,21 @@
 // the mapping from that token to the underlying PJRT buffer; mc_tpu_pjrt_* copy
 // and classification calls resolve the token through the adapter's own
 // registry.
+//
+// INTERIOR POINTERS (load-bearing): TENT stages transfers through host DRAM in
+// chunks (see ProxyManager, chunk_size defaults to 4 MiB), and hands the
+// adapter `token + chunk_offset` for every chunk after the first. Every
+// entrypoint below therefore takes an address that may point into the MIDDLE of
+// a registered buffer, not only at its base. An adapter whose registry only
+// matches base addresses will report `is_device_ptr(token + off) == 0`, TENT
+// will classify TPU HBM as host memory, and the staging copy silently degrades
+// into a memcpy from a non-data address -- corrupting every transfer larger
+// than one chunk without raising an error. Adapters MUST resolve an address to
+// the registered buffer whose range [base, base + size) contains it.
+//
+// The token is NOT required to be host-dereferenceable, and on real PJRT/TPU it
+// is not: PJRT_Buffer_UnsafePointer returns an internal handle that reads as
+// garbage rather than buffer contents. Never dereference it.
 
 #ifndef TENT_PLATFORM_TPU_PJRT_ABI_H_
 #define TENT_PLATFORM_TPU_PJRT_ABI_H_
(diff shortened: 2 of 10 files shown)


staryxchen (approved):
Thanks for the deep investigation and the thorough write-up. The ABI contract clarification, the transport-level guard in TpuTransport::startTransfer, the mock adapter rewrite with poisoned tokens, and the new test coverage all look excellent — approving those.

PR #2816, opened 2026-07-09
[TENT] Wire live RDMA bandwidth into admission queue degradation policy

catyans:
## Summary

Bridge the existing per-NIC EWMA bandwidth (already tracked by `DeviceSelector` on every RDMA completion) into the admission queue's `BandwidthProvider`, activating the deadline-infeasible drop path (RFC #2519 step 3) with real telemetry instead of a static mock.

- Add `Transport::getEstimatedBandwidth()` virtual method (base returns −1)
- `RdmaTransport` overrides: aggregates EWMA across all local NICs via `DeviceSelector::getAggregateEwmaBandwidth()`
- `TransferEngineImpl::construct()` reads two new config keys and wires `setDegradationPolicy()` when both are set
- Thread-safe: EWMA is `std::atomic<double>` with relaxed ordering; progress worker reads, RDMA workers write

**Opt-in, default-off.** Without explicit config, behavior is byte-identical to today.

### New config keys

| Key | Default | Description |
|-----|---------|-------------|
| `runtime_queue/deadline_aware` | `false` | Enable EDF dispatch ordering |
| `runtime_queue/mlu_local_threshold` | `0.0` | θ_local for deadline-infeasible drop (0 = disabled) |

### Full cmake build (cluster: lingjun-100, gcc-13.3.0, Ubuntu 24.04)

Full in-tree cmake build with `-DUSE_TENT=ON` and all dependencies (ibverbs, jsoncpp, yalantinglibs, gtest, etc.) compiles cleanly to 100%:

```
$ cmake .. -DUSE_TENT=ON && make -j8
...
[100%] Built target tebench
```

### Unit Tests (admission_queue_test, 3 rounds shuffled)

```
$ ./tent/tests/admission_queue_test --gtest_repeat=3 --gtest_shuffle
Repeating all tests (iteration 1) . seed=33810
[  PASSED  ] 23 tests.
Repeating all tests (iteration 2) . seed=33811
[  PASSED  ] 23 tests.
Repeating all tests (iteration 3) . seed=33812
[  PASSED  ] 23 tests.
```

New test `Step3DynamicBandwidthProvider` verifies that when the bandwidth provider returns a changing value (simulating live RDMA telemetry), the drop decision adapts correctly:
- At 1 GB/s: 16-byte transfer with 10ns window → MLU 1.6 ≥ θ 1.5 → **dropped**
- At 10 GB/s: same profile → MLU 0.16 < 1.5 → **dispatched**

### Related TENT tests (3 rounds shuffled each)

```
$ ./tent/tests/tent_runtime_queue_dispatch_test --gtest_repeat=3 --gtest_shuffle
[  PASSED  ] 9 tests. (x3)

$ ./tent/tests/tent_progress_worker_test --gtest_repeat=3 --gtest_shuffle
[  PASSED  ] 6 tests. (x3)
```

### Functional Tests (same environment)

| Test | Result |
|------|--------|
| 2-NIC aggregate bandwidth (2×25 GB/s) | 50 GB/s ✓ |
| EWMA degradation (100 updates at 10 GB/s) | Aggregate drops to 35 GB/s ✓ |
| Concurrent read/write (1M writer updates + reader) | 204,797 reads, zero data races ✓ |
| Empty device list | Returns −1.0 (safe fallback) ✓ |

## Test plan

- [x] Full cmake build with `-DUSE_TENT=ON` (100% targets)
- [x] Unit tests pass (23 × 3 rounds, shuffled)
- [x] Runtime queue dispatch tests pass (9 × 3 rounds)
- [x] Progress worker tests pass (6 × 3 rounds)
- [x] Functional correctness of bandwidth aggregation
- [x] Thread-safety of cross-thread atomic read
- [x] clang-format clean (no violations on new code)
- [ ] CI green


catyans:
The lifetime review is resolved by `5b6a3c4`: the admission queue's bandwidth-provider lambda captures `std::weak_ptr<Transport>`, returns `-1.0` after expiration, and no longer retains `RdmaTransport` past `TransferEngineImpl::deconstruct()`.

Cluster validation (PR head `5b6a3c4`, Lingjun H20/CX-7 nodes 100/102, `mlx5_bond_0`, RoCEv2 GID 3):

- Full targeted build succeeded on both nodes (`admission_queue_test` + `tebench`).
- `admission_queue_test --gtest_repeat=3 --gtest_shuffle` passed on both nodes: 23 x 3 x 2 = 138 successful test executions.
- A remote-only integration harness trained the production `RdmaTransport` EWMA with 200 real 1 MiB RDMA READs, then invoked the production-wired admission queue provider.
- Live aggregate EWMA was `5,000,000,000 B/s` (the configured single-NIC minimum clamp under this synchronous workload).
- For an identical 64 MiB request profile at theta=1.0: a deadline at 0.5x predicted time was dropped, while a deadline at 8x predicted time was dispatched.
- After `TransferEngineImpl` destruction, a weak reference to the real RDMA transport was immediately expired, directly validating that the provider no longer extends transport lifetime.
- The entire live-EWMA -> degradation decision -> destruction sequence passed 3/3 runs.

The experiment-only accessors/harness stayed in the remote clone and are not part of the PR.

mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
```diff
@@ -368,6 +372,30 @@ Status TransferEngineImpl::construct() {
 
     staging_proxy_ = std::make_unique<ProxyManager>(this);
 
+    if (runtime_queue_config_.limits.deadline_aware &&
```
alogfans, line 377: The staging path also supports host-to-host transfer via MNNVL-only fabric. So this patch should not apply in this case.

mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
```diff
@@ -368,6 +372,23 @@ Status TransferEngineImpl::construct() {
...
+        runtime_queue_config_.limits.mlu_local_threshold > 0.0) {
+        auto rdma_xport =
+            transport_list_[static_cast<int>(TransportType::RDMA)];
+        if (rdma_xport) {
+            runtime_queue_->setDegradationPolicy(
+                [rdma_xport]() -> double {
+                    return rdma_xport->getEstimatedBandwidth();
+                },
+                DegradationHooks{}, nullptr);
```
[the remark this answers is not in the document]
catyans: Fixed in 5b6a3c48: the provider captures a `std::weak_ptr<Transport>` and locks it per sample, so clearing `transport_list_` can destroy RDMA workers normally during deconstruct.

mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
catyans: Fixed in ffe705cc. Eligibility is now computed per prepared owner: only a direct RDMA route may use the RDMA EWMA for step-3 MLU drop. MNNVL, TCP, and every staging owner are marked ineligible, so an installed RDMA transport can no longer affect an MNNVL-only host-to-host staging transfer. Added `Step3SkipsNonRdmaOwner` to verify that an otherwise-infeasible non-RDMA owner is dispatched rather than dropped.

catyans:
Follow-up validation for `ffe705cc`: rebuilt the merged admission-queue sources on the backup Linux host and ran all 32 `AdmissionQueueTest` cases, including `Step3SkipsNonRdmaOwner`; 32/32 passed. The PR is mergeable, both review threads are resolved, and the full CI matrix is running.

mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
```diff
@@ -374,6 +374,30 @@ Status TransferEngineImpl::construct() {
 
     staging_proxy_ = std::make_unique<ProxyManager>(this);
```
alogfans, line 375: The staging path also supports host-to-host transfer via MNNVL-only fabric. So this patch should not apply in this case.

mooncake-transfer-engine/tent/include/tent/runtime/admission_queue.h
```diff
@@ -74,6 +74,10 @@ struct QueueOwnerInput {
     std::vector<size_t> derived_task_ids;
     Request request{};
     QueueOwnerKind kind{QueueOwnerKind::User};
+    // True only when the caller has established that this owner's transfer
+    // time is governed by the installed bandwidth provider. This prevents an
+    // RDMA EWMA from being applied to MNNVL/TCP/staging paths.
+    bool degradation_eligible{true};
```
alogfans: Should it default to false instead, to avoid accidentally applying RDMA degradation when a future enqueue path forgets to set it explicitly?

catyans: Done. I changed `degradation_eligible` to default to `false` for both `QueueOwnerInput` and the internal queue owner, so RDMA degradation is now explicitly opt-in for each enqueue path.

I also updated the admission queue tests to mark the RDMA-covered step-3 cases explicitly eligible, and added a regression test (`Step3RequiresExplicitDegradationEligibility`) covering the future-enqueue-path case: with deadline/threshold/provider present but no explicit eligibility, the owner is dispatched instead of being dropped.

Validation on the Lingjun cluster:
- node: `lingjun-100` (`8.130.214.133`)
- container: `codex_mooncake_pr_eval`
- commit: `7e69ee29`
- command: `/root/mooncake-pr2816-build/tent/tests/admission_queue_test`
- result: 33/33 tests passed.

alogfans (approved):
LGTM

PR #2822, opened 2026-07-09
[Store] feat: expose client metrics HTTP config to Python

Lin-z-w:
## Description

Expose the Mooncake Store client-side health and metrics HTTP endpoint through the Python-facing client setup APIs and the packaged store service configuration.

This change migrates the client Python metrics exposure behavior from Mooncake-taas PR5 and includes the relevant PR9 compatibility fix. It adds:

- Python and pybind setup arguments for `enable_client_http_server` and `client_http_port`.
- `MooncakeConfig` JSON/env support via `MOONCAKE_ENABLE_CLIENT_HTTP_SERVER` and `MOONCAKE_CLIENT_HTTP_PORT`.
- Store service propagation of the new config fields into `MooncakeDistributedStore.setup`.
- Client HTTP endpoint tests for `/health`, `/metrics`, and `/metrics/summary`, including disabled-metrics and port-conflict behavior.
- Additional metrics latency buckets for the long-tail ranges.
- Documentation for deployment and observability usage.

The client HTTP endpoint remains disabled by default, so existing Python and SGLang integrations keep their current behavior unless they opt in.

## Module

- [ ] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [x] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
git diff --check

SKIP=mooncake-code-format pre-commit run --files docs/source/deployment/mooncake-store-deployment-guide.md docs/source/getting_started/observability.md mooncake-integration/store/store_py.cpp mooncake-store/include/client_metric.h mooncake-store/include/dummy_client.h mooncake-store/include/pyclient.h mooncake-store/include/real_client.h mooncake-store/include/types.h mooncake-store/src/real_client.cpp mooncake-store/src/real_client_main.cpp mooncake-store/tests/client_metrics_test.cpp mooncake-wheel/mooncake/mooncake_config.py mooncake-wheel/mooncake/mooncake_store_service.py mooncake-wheel/tests/test_mooncake_config.py mooncake-wheel/tests/test_mooncake_store_service_api.py

cmake --build build --target client_metrics_test store
build/mooncake-store/tests/client_metrics_test

PYTHONPATH=/home/lzw/Mooncake/mooncake-wheel python3 -m unittest mooncake-wheel/tests/test_mooncake_config.py mooncake-wheel.tests.test_mooncake_store_service_api.StoreServiceApiTest.test_start_store_service_passes_tenant_id_to_setup mooncake-wheel.tests.test_mooncake_store_service_api.StoreServiceApiTest.test_cli_config_can_override_tenant_id

PATH=/home/lzw/Mooncake/docs/.venv/bin:$PATH make html
```

**Test results:**
- [x] Unit tests pass

Manual validation:

- `client_metrics_test` passed: 12 tests.
- Python unittest passed: 37 tests.
- Sphinx `make html` passed with only sandbox-network intersphinx inventory warnings.
- `mooncake-code-format` was run through the pre-commit hook and formatted the touched C++ files. It was skipped on the final targeted pre-commit/commit because the hook compares against local `origin/main`, which was older than the rebased `upstream/main` and attempted to format unrelated upstream files.

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

AI assistance: OpenAI Codex was used to inspect the codebase, port the changes, resolve rebase conflicts, run validation, and draft this PR description. I reviewed and edited the generated code before submission.

### DIFF
--- a/mooncake-store/src/real_client.cpp
+++ b/mooncake-store/src/real_client.cpp
@@ -626,7 +626,8 @@ tl::expected<void, ErrorCode> RealClient::setup_internal(
     const std::shared_ptr<TransferEngine> &transfer_engine,
     const std::string &ipc_socket_path, int local_rpc_port,
     bool enable_ssd_offload, bool start_offload_rpc_server,
-    const std::string &ssd_offload_path, const std::string &tenant_id) {
+    const std::string &ssd_offload_path, const std::string &tenant_id,
+    bool enable_client_http_server, int client_http_port) {
     this->protocol = protocol;
     this->ipc_socket_path_ = ipc_socket_path;
     const bool should_use_hugepage =
@@ -935,10 +936,20 @@ tl::expected<void, ErrorCode> RealClient::setup_internal(
         }
     }
     client_requester_ = std::make_shared<ClientRequester>();
-    if (FLAGS_enable_http_server) {
-        if (start_http_server() != 0) {
-            LOG(ERROR) << "Failed to start HTTP server on port "
-                       << FLAGS_http_port;
+    const bool should_start_http_server =
+        enable_client_http_server || FLAGS_enable_http_server;
+    const int selected_http_port =
+        enable_client_http_server ? client_http_port : FLAGS_http_port;
+    if (should_start_http_server) {
+        if (selected_http_port <= 0 || selected_http_port > 65535) {
+            LOG(ERROR) << "Invalid client HTTP server port: "
+                       << selected_http_port;
+            return tl::unexpected(ErrorCode::INVALID_PARAMS);
+        }
+        if (start_http_server(selected_http_port) != 0) {
+            LOG(WARNING) << "Failed to start client HTTP server on port "
+                         << selected_http_port
+                         << "; continuing without HTTP endpoints";
         }
     }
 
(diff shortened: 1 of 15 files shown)


mooncake-store/src/real_client.cpp
```diff
@@ -984,6 +996,64 @@ inline std::optional<size_t> get_config_size(const ConfigDict &config,
...
+        value == "disable") {
+        return false;
+    }
+
+    LOG(WARNING) << "Invalid boolean value for config key '" << key
+                 << "': " << it->second << ", using default: " << default_value;
+    return default_value;
+}
+
+inline std::optional<int> get_config_int(const ConfigDict &config,
```
Aionw, line 1034: Since we already have a ConfigDict type, would it make sense to turn it into a proper class? Then these helper functions could become public methods, rather than living in real_client.cpp.

mooncake-store/src/real_client.cpp
```diff
@@ -984,6 +996,64 @@ inline std::optional<size_t> get_config_size(const ConfigDict &config,
...
+                                         int default_value) {
+    auto it = config.find(key);
+    if (it == config.end()) {
+        return default_value;
+    }
+
+    try {
+        std::string value = trim(it->second);
+        size_t pos = 0;
+        int parsed_value = std::stoi(value, &pos);
```
Aionw: Nit: std::from_chars is a more modern alternative and doesn’t throw exceptions.

Here's a some example.
```cpp
auto [p, ec] = std::from_chars(s.data(), s.data() + s.size(), x);

if (ec == std::errc{}) {
}
```

mooncake-store/src/real_client.cpp
Lin-z-w: Thanks for the suggestion. ConfigDict is currently a public alias used by RealClient::setup_internal and existing C++ callers. Turning it into a class would broaden the scope of this PR and could introduce source and ABI compatibility concerns. Since these parsing rules are specific to client setup, I prefer to keep the helpers local in this PR. We can consider introducing a typed ClientConfig separately if we want to formalize the configuration API.

mooncake-store/src/real_client.cpp
Lin-z-w: Good suggestion. I updated get_config_int to use std::from_chars, checking both the error code and that the entire trimmed value was consumed. I also added regression coverage for trailing characters and out-of-range integer values in commit 72a2b0d1.

Aionw (approved):

PR #2823, opened 2026-07-09
Record P2P metadata oplog

skygragon:
## Description

  Record P2P metadata mutations into OpLog so standby metadata can replay failover-visible P2P state changes.

  This change adds OpLog recording for:
  - P2P client register/unregister
  - P2P segment mount/unmount
  - P2P replica add/remove
  - successful `BatchSyncReplica` add/remove operations

  It also wires OpLog configuration through master config, initializes the master-side OpLog manager when enabled, fails fast if the OpLog
  store cannot be initialized, and adds P2P OpLog recording tests.

  ## Module

  - [ ] CI/CD
  - [x] Docs
  - [ ] Other

  ## Type of Change

  - [ ] Bug fix
  - [x] New feature
  - [ ] Refactor
  - [ ] Breaking change
  - [x] Documentation update
  - [ ] Performance improvement
  - [ ] Other

  ## How Has This Been Tested?

  Built and ran the P2P OpLog unit test target in container `mooncake-redis-ha-verify`.

  **Test commands:**
  ```bash
  clang-format-20 -style=file -i \
    mooncake-store/tests/ha/oplog/p2p_record_oplog_test.cpp

  cd /tmp/mooncake-record-oplog-build
  cmake --build . --target p2p_oplog_test -j$(nproc)
  ./mooncake-store/tests/p2p_oplog_test
```
  Test results:

  - [x] Unit tests pass

  p2p_oplog_test: 66 tests passed.

  ## Checklist

  ## AI Assistance Disclosure

  - [ ] No AI tools were used

  AI tools were used to review the change, identify consistency risks, adjust tests, and prepare this PR description. The submitter reviewed
  and validated the final changes.

### DIFF
--- a/mooncake-store/src/p2p_master_service.cpp
+++ b/mooncake-store/src/p2p_master_service.cpp
@@ -19,7 +19,7 @@ P2PMasterService::P2PMasterService(const MasterServiceConfig& config)
 }
 
 ErrorCode P2PMasterService::RecordOplog(OpType type, const std::string& key,
-                                        const std::string& payload, bool sync) {
+                                        const std::string& payload) {
     // TODO: Record remaining failover-visible P2P mutations: client crash
     // cleanup, heartbeat state transitions, replica eviction/rebalance, and
     // task metadata.
@@ -28,21 +28,21 @@ ErrorCode P2PMasterService::RecordOplog(OpType type, const std::string& key,
         return ErrorCode::OK;
     }
 
-    if (sync) {
-        auto result = manager->AppendAndPersist(type, key, payload);
-        if (!result.has_value()) {
-            LOG(ERROR) << "P2PMasterService: failed to persist oplog"
-                       << ", op_type=" << static_cast<int>(type)
-                       << ", key=" << key
-                       << ", error=" << toString(result.error());
-            return result.error();
-        }
-    } else {
-        manager->Append(type, key, payload);
+    auto result = manager->AppendAndPersist(type, key, payload);
+    if (!result.has_value()) {
+        LOG(ERROR) << "P2PMasterService: failed to persist oplog"
+                   << ", op_type=" << static_cast<int>(type) << ", key=" << key
+                   << ", error=" << toString(result.error());
+        return result.error();
     }
     return ErrorCode::OK;
 }
 
+// TODO: Make client/segment lifecycle updates transactional with oplog
+// persistence. These wrappers currently call MasterService first, so an oplog
+// persistence failure can leave Primary memory ahead of Standby. A robust fix
+// needs a separate prepare/log/apply flow or durable retry/outbox mechanism;
+// local rollback is fragile because these calls have nested side effects.
 auto P2PMasterService::RegisterClient(const RegisterClientRequest& req)
     -> tl::expected<RegisterClientResponse, ErrorCode> {
     if (GetClientManager().GetClient(req.client_id)) {
@@ -59,6 +59,9 @@ auto P2PMasterService::RegisterClient(const RegisterClientRequest& req)
 
     auto result = MasterService::RegisterClient(req);
     if (!result.has_value()) {
+        LOG(ERROR) << "RegisterClient(P2P): failed"
+                   << ", client_id=" << req.client_id
+                   << ", error=" << result.error();
         return result;
     }
 
@@ -70,6 +73,8 @@ auto P2PMasterService::RegisterClient(const RegisterClientRequest& req)
     auto err =
         RecordOplog(OpType_REGISTER_CLIENT, "", SerializeP2PPayload(payload));
     if (err != ErrorCode::OK) {
+        LOG(ERROR) << "RegisterClient(P2P): failed to record oplog"
+                   << ", client_id=" << req.client_id << ", error=" << err;
         return tl::make_unexpected(err);
     }
     return result;
@@ -79,6 +84,9 @@ auto P2PMasterService::UnregisterClient(const UnregisterClientRequest& req)
     -> tl::expected<UnregisterClientResponse, ErrorCode> {
     auto result = MasterService::UnregisterClient(req);
     if (!result.has_value()) {
+        LOG(ERROR) << "UnregisterClient(P2P): failed"
+                   << ", client_id=" << req.client_id
+                   << ", error=" << result.error();
         return result;
     }
 
@@ -87,6 +95,8 @@ auto P2PMasterService::UnregisterClient(const UnregisterClientRequest& req)
     auto err =
         RecordOplog(OpType_UNREGISTER_CLIENT, "", SerializeP2PPayload(payload));
     if (err != ErrorCode::OK) {
+        LOG(ERROR) << "UnregisterClient(P2P): failed to record oplog"
+                   << ", client_id=" << req.client_id << ", error=" << err;
         return tl::make_unexpected(err);
     }
     return result;
@@ -97,6 +107,11 @@ auto P2PMasterService::MountSegment(const Segment& segment,
     -> tl::expected<void, ErrorCode> {
     auto result = MasterService::MountSegment(segment, client_id);
     if (!result.has_value()) {
+        LOG(ERROR) << "MountSegment(P2P): failed"
+                   << ", client_id=" << client_id
+                   << ", segment_id=" << segment.id
+                   << ", segment_name=" << segment.name
+                   << ", error=" << result.error();
         return result;
     }
 
@@ -106,6 +121,10 @@ auto P2PMasterService::MountSegment(const Segment& segment,
     auto err =
         RecordOplog(OpType_MOUNT_SEGMENT, "", SerializeP2PPayload(payload));
     if (err != ErrorCode::OK) {
+        LOG(ERROR) << "MountSegment(P2P): failed to record oplog"
+                   << ", client_id=" << client_id
+                   << ", segment_id=" << segment.id
+                   << ", segment_name=" << segment.name << ", error=" << err;
         return tl::make_unexpected(err);
     }
     return {};
@@ -116,6 +135,10 @@ auto P2PMasterService::UnmountSegment(const UUID& segment_id,
     -> tl::expected<void, ErrorCode> {
     auto result = MasterService::UnmountSegment(segment_id, client_id);
     if (!result.has_value()) {
+        LOG(ERROR) << "UnmountSegment(P2P): failed"
+                   << ", client_id=" << client_id
+                   << ", segment_id=" << segment_id
+                   << ", error=" << result.error();
         return result;
     }
 
@@ -125,6 +148,9 @@ auto P2PMasterService::UnmountSegment(const UUID& segment_id,
     auto err =
         RecordOplog(OpType_UNMOUNT_SEGMENT, "", SerializeP2PPayload(payload));
     if (err != ErrorCode::OK) {
+        LOG(ERROR) << "UnmountSegment(P2P): failed to record oplog"
+                   << ", client_id=" << client_id
+                   << ", segment_id=" << segment_id << ", error=" << err;
         return tl::make_unexpected(err);
     }
     return {};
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: mooncake-store/src/p2p_master_service.cpp [hunk 1]
```diff
@@ -17,6 +18,106 @@ P2PMasterService::P2PMasterService(const MasterServiceConfig& config)
...
+    RegisterClientPayload payload;
+    payload.client_id = req.client_id;
+    payload.ip_address = *req.ip_address;
+    payload.rpc_port = *req.rpc_port;
+    payload.segments = req.segments;
+    auto err =
+        RecordOplog(OpType_REGISTER_CLIENT, "", SerializeP2PPayload(payload));
+    if (err != ErrorCode::OK) {
+        return tl::make_unexpected(err);
+    }
```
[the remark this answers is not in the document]
skygragon: This HA path is not a strong-consistency protocol. The P2P metadata OpLog is used as a recovery/catch-up aid, while the current primary's in-memory metadata remains authoritative for online serving. For availability, we intentionally do not terminate the primary on an OpLog persistence failure. The failure is returned to the caller and logged with operation context so the client/control plane can retry, instead of turning a transient OpLog backend failure into a primary outage.

  Rollback is also unsafe for these lifecycle operations because unregister/unmount may cascade into segment and replica cleanup. A future change can introduce a proper prepare/persist/apply flow if we decide to make this path stronger, but that is outside the current availability-oriented design.

mooncake-store/src/p2p_master_service.cpp
```diff
@@ -17,6 +18,118 @@ P2PMasterService::P2PMasterService(const MasterServiceConfig& config)
     client_manager_->Start();
 }
 
+ErrorCode P2PMasterService::RecordOplog(OpType type, const std::string& key,
+                                        const std::string& payload, bool sync) {
```
wanyue-wy: When will the `sync` flag be set to false

mooncake-store/src/p2p_master_service.cpp
```diff
@@ -17,6 +18,118 @@ P2PMasterService::P2PMasterService(const MasterServiceConfig& config)
...
+            return result.error();
+        }
+    } else {
+        manager->Append(type, key, payload);
+    }
+    return ErrorCode::OK;
+}
+
+auto P2PMasterService::RegisterClient(const RegisterClientRequest& req)
+    -> tl::expected<RegisterClientResponse, ErrorCode> {
```
wanyue-wy, line 47: 1. For `RegisterClient`, `UnregisterClient`, `MountSegment`, and `UnmountSegment`, we currently update the in‑memory state first and then record the log. If the log record fails, the function will return an error anyway. Howeverm in this case, we should either roll back the memory state change, or ensure the HA state machine can capture this situation (e.g., by retrying recording log later).

2. To reduce the time cost of routing a unavailable client, `RegisterClient` actually combines both client registration and segment mounting into a single flow – it internally calls the mount logic inside `client_manager::RegisterClient` instead of relying on a separate MountSegment call. As a result, we're missing the mount log entry for this particular scenario.

mooncake-store/src/p2p_master_service.cpp
```diff
@@ -17,6 +18,118 @@ P2PMasterService::P2PMasterService(const MasterServiceConfig& config)
...
+    if (!req.ip_address || !req.rpc_port) {
+        LOG(ERROR) << "RegisterClient(P2P): missing endpoint"
+                   << ", client_id=" << req.client_id;
+        return tl::make_unexpected(ErrorCode::INVALID_PARAMS);
+    }
+
+    auto result = MasterService::RegisterClient(req);
+    if (!result.has_value()) {
+        return result;
+    }
```
wanyue-wy, line 66: we need to print a error log

mooncake-store/src/p2p_master_service.cpp
```diff
@@ -17,6 +18,118 @@ P2PMasterService::P2PMasterService(const MasterServiceConfig& config)
...
+    RegisterClientPayload payload;
+    payload.client_id = req.client_id;
+    payload.ip_address = *req.ip_address;
+    payload.rpc_port = *req.rpc_port;
+    payload.segments = req.segments;
+    auto err =
+        RecordOplog(OpType_REGISTER_CLIENT, "", SerializeP2PPayload(payload));
+    if (err != ErrorCode::OK) {
+        return tl::make_unexpected(err);
+    }
```
wanyue-wy, line 79: we need to print a error log

mooncake-store/src/p2p_master_service.cpp
```diff
@@ -17,6 +18,118 @@ P2PMasterService::P2PMasterService(const MasterServiceConfig& config)
...
+    return result;
+}
+
+auto P2PMasterService::UnregisterClient(const UnregisterClientRequest& req)
+    -> tl::expected<UnregisterClientResponse, ErrorCode> {
+    auto result = MasterService::UnregisterClient(req);
+    if (!result.has_value()) {
+        return result;
+    }
```
wanyue-wy, line 91: we need to print a error log

mooncake-store/src/p2p_master_service.cpp
```diff
@@ -17,6 +18,118 @@ P2PMasterService::P2PMasterService(const MasterServiceConfig& config)
...
+
+    UnregisterClientPayload payload;
+    payload.client_id = req.client_id;
+    auto err =
+        RecordOplog(OpType_UNREGISTER_CLIENT, "", SerializeP2PPayload(payload));
+    if (err != ErrorCode::OK) {
+        return tl::make_unexpected(err);
+    }
```
wanyue-wy, line 101: we need to print a error log

mooncake-store/src/p2p_master_service.cpp
```diff
@@ -17,6 +18,118 @@ P2PMasterService::P2PMasterService(const MasterServiceConfig& config)
...
+    return result;
+}
+
+auto P2PMasterService::MountSegment(const Segment& segment,
+                                    const UUID& client_id)
+    -> tl::expected<void, ErrorCode> {
+    auto result = MasterService::MountSegment(segment, client_id);
+    if (!result.has_value()) {
+        return result;
+    }
```
wanyue-wy, line 116: we need to print a error log

mooncake-store/src/p2p_master_service.cpp
```diff
@@ -17,6 +18,118 @@ P2PMasterService::P2PMasterService(const MasterServiceConfig& config)
...
+
+    MountSegmentPayload payload;
+    payload.client_id = client_id;
+    payload.segment = segment;
+    auto err =
+        RecordOplog(OpType_MOUNT_SEGMENT, "", SerializeP2PPayload(payload));
+    if (err != ErrorCode::OK) {
+        return tl::make_unexpected(err);
+    }
```
wanyue-wy, line 129: we need to print a error log

mooncake-store/src/p2p_master_service.cpp
```diff
@@ -17,6 +18,118 @@ P2PMasterService::P2PMasterService(const MasterServiceConfig& config)
...
+    return {};
+}
+
+auto P2PMasterService::UnmountSegment(const UUID& segment_id,
+                                      const UUID& client_id)
+    -> tl::expected<void, ErrorCode> {
+    auto result = MasterService::UnmountSegment(segment_id, client_id);
+    if (!result.has_value()) {
+        return result;
+    }
```
wanyue-wy, line 143: we need to print a error log

mooncake-store/src/p2p_master_service.cpp
```diff
@@ -17,6 +18,118 @@ P2PMasterService::P2PMasterService(const MasterServiceConfig& config)
...
+
+    UnmountSegmentPayload payload;
+    payload.segment_id = segment_id;
+    payload.client_id = client_id;
+    auto err =
+        RecordOplog(OpType_UNMOUNT_SEGMENT, "", SerializeP2PPayload(payload));
+    if (err != ErrorCode::OK) {
+        return tl::make_unexpected(err);
+    }
```
wanyue-wy, line 155: we need to print a error log

mooncake-store/src/p2p_master_service.cpp
```diff
@@ -268,6 +381,17 @@ tl::expected<void, ErrorCode> P2PMasterService::InnerAddReplica(
...
+        payload.object_key = std::string(key);
+        payload.client_id = client_id;
+        payload.segment_id = segment_id;
+        payload.size = size;
+        ErrorCode record_err =
+            RecordOplog(OpType_ADD_REPLICA, payload.object_key,
+                        SerializeP2PPayload(payload));
+        if (record_err != ErrorCode::OK) {
+            return tl::make_unexpected(record_err);
+        }
```
wanyue-wy, line 424: we need to print a error log

mooncake-store/src/p2p_master_service.cpp
```diff
@@ -276,6 +400,17 @@ tl::expected<void, ErrorCode> P2PMasterService::InnerAddReplica(
...
```
wanyue-wy, line 447: we need to print a error log

mooncake-store/src/p2p_master_service.cpp
```diff
@@ -316,6 +451,16 @@ tl::expected<void, ErrorCode> P2PMasterService::InnerRemoveReplica(
...
+            RemoveReplicaPayload payload;
+            payload.object_key = std::string(key);
+            payload.client_id = client_id;
+            payload.segment_id = segment_id;
+            ErrorCode record_err =
+                RecordOplog(OpType_REMOVE_REPLICA, payload.object_key,
+                            SerializeP2PPayload(payload));
+            if (record_err != ErrorCode::OK) {
+                return tl::make_unexpected(record_err);
+            }
```
wanyue-wy, line 501: we need to print a error log

mooncake-store/src/p2p_master_service.cpp
```diff
@@ -17,6 +18,118 @@ P2PMasterService::P2PMasterService(const MasterServiceConfig& config)
...
+    auto err =
+        RecordOplog(OpType_MOUNT_SEGMENT, "", SerializeP2PPayload(payload));
+    if (err != ErrorCode::OK) {
+        return tl::make_unexpected(err);
+    }
+    return {};
+}
+
+auto P2PMasterService::UnmountSegment(const UUID& segment_id,
+                                      const UUID& client_id)
```
wanyue-wy, line 134: If `UnmountSegment` is called externally, all `replica` data associated with the `segment` should be reclaimed. The unmount logic currently relies on `MasterService::OnSegmentRemoved` to clear that replica information. If we only log the UnmountSegment operation itself, can we be sure that log replay will bring the data back to a consistent state?

mooncake-store/src/p2p_master_service.cpp
skygragon: The `sync` flag was inspired by the lower-level OpLogStore persistence API, but P2P currently has no mutation that should use the best-effort path. All current P2P master-visible mutations need to persist the oplog before updating in-memory state, so I removed the parameter and made `RecordOplog` always use `AppendAndPersist()`.

mooncake-store/src/p2p_master_service.cpp
skygragon: 1.   Agree. This is a real consistency gap. It was planned to be handled in a follow-up PR with a proper prepare/log/apply or retry/outbox design, since local rollback here is fragile due to nested side effects. I added a TODO to make that explicit.
2.   For registration-time segments, I think the current REGISTER_CLIENT payload already covers this: it carries `req.segments`, and the standby applier merges them when applying REGISTER_CLIENT. Standalone MOUNT_SEGMENT is still used for segments mounted after registration. Emitting both would duplicate the same initial segment state in the oplog.

mooncake-store/src/p2p_master_service.cpp
skygragon: Done.  BTW RecordOplog already logs the persistence failure with op_type/key/error, so the outer log is not strictly required. But adding call-site context makes debugging easier, especially for keyless ops like RegisterClient. I added contextual error logs at all RecordOplog failure call sites.

skygragon: Done.  BTW RecordOplog already logs the persistence failure with op_type/key/error, so the outer log is not strictly required. But adding call-site context makes debugging easier, especially for keyless ops like RegisterClient. I added contextual error logs at all RecordOplog failure call sites.

mooncake-store/src/p2p_master_service.cpp
skygragon: Yes. UNMOUNT_SEGMENT replay mirrors the primary-side cascade cleanup: ApplyUnmountSegment calls
  P2PStandbyMetadataStore::RemoveSegment(), and RemoveSegment removes the segment from the client and calls RemoveReplicasBySegmentInternal(segment_id). So replicas associated with the unmounted segment are reclaimed during replay as well.

Issue #2756, opened 2026-07-06
[RFC]: OpLog Batch Record Ordered Writer and Durable Prefix


Icedcoco:
## Introduction

Mooncake Store HA should do more than restart the Master process quickly. After a primary failure and failover, the new Master should preserve as much useful KV cache metadata, segment registry state, and object-location index state as possible.

The existing HA work has two complementary recovery paths:

1. **Snapshot** provides a coarse-grained baseline for restoring complete Master state.
2. **OpLog** provides incremental changes after the snapshot, so a standby can keep following the primary and catch up before promotion.

RFC #1200 proposed an etcd-backed hot-standby mechanism for Master metadata HA. The primary writes metadata mutations as OpLog entries, and the standby reads/watches those entries from etcd and applies them locally. RFC #1650 then introduced an `OpLogStore` / `OpLogChangeNotifier` abstraction to decouple the upper-layer replication logic from the etcd-specific implementation.

PR #2331 is the immediate baseline for this RFC. Although that PR was submitted by @Icedcoco, most of the HA hot-standby and OpLog implementation came from earlier work by @Libotry and @00fish0 in the `dev/snapshot_etcd_storage_integration_testing_fixbug` branch. The work in #2331 was mainly to rebase that legacy implementation onto the current `kvcache-ai/Mooncake:main`, adapt it to the current Mooncake Store architecture, and integrate recently added Store features such as tenant-aware object metadata, segment lifecycle replication, and the updated standby promotion / snapshot handoff paths.

This RFC builds on #2331 and proposes the next step for the OpLog write and replication protocol: **batch records**, an **ordered writer**, and a **durable prefix**. The goal is not to replace the business semantics that #2331 already adapted. Instead, this RFC keeps the current logical `OpLogEntry` semantics and replaces the underlying write and replay protocol to address lock-held durable waits, per-entry backend key amplification, and promotion boundary ambiguity.

Relevant links:

- Roadmap: https://github.com/kvcache-ai/Mooncake/issues/1035
- Hot Standby RFC: https://github.com/kvcache-ai/Mooncake/issues/1200
- OpLog Store Abstraction RFC: https://github.com/kvcache-ai/Mooncake/issues/1650
- Snapshot RFC: https://github.com/kvcache-ai/Mooncake/issues/1150
- etcd wrapper for HA: https://github.com/kvcache-ai/Mooncake/pull/1451
[...]

PR #2331, opened 2026-06-04
[Store] add HA OpLog replication and tenant-aware standby promotion

Icedcoco:
## Description

This PR is based on the HA hot-standby development work originally done by @Libotry and @00fish0 in the `dev/snapshot_etcd_storage_integration_testing_fixbug` branch of https://github.com/Libotry/Mooncake.

Many thanks to @Libotry and @00fish0 for the original implementation and debugging work. My work in this PR is mainly to rebase that development onto the latest `kvcache-ai/Mooncake:main`, adapt it to the current Mooncake Store architecture, and
integrate it with recently added Store features such as tenant-aware object metadata and the updated promotion/snapshot paths.

### Key Changes

- **HA OpLog Configuration**: Added `oplog_store_type`, `oplog_store_root_dir`, and `oplog_poll_interval_ms` to master config, YAML config, CLI flags, and deployment docs.
- **Primary-side OpLog Publishing**: Adapted Store metadata mutation paths to publish durable OpLog entries for object, replica, and segment lifecycle changes.
- **Strong-consistency Guarding**: For HA-enabled metadata mutations, persist the OpLog entry before committing local metadata changes where dropping the update would make Primary and Standby diverge.
- **Segment Lifecycle Replication**: Added `SEGMENT_MOUNT`, `SEGMENT_UNMOUNT`, and `SEGMENT_UPDATE` OpLog entries plus standby-side segment registry replay.
- **Tenant-aware HA Metadata**: Adapted the HA metadata path to support `tenant_id`, tenant-aware standby metadata storage, and backward-compatible OpLog deserialization.
- **Standby Promotion Restore**: Added standby snapshot export and Primary-side restoration so a promoted standby can initialize object metadata, group membership, segment state, and OpLog sequence baseline.
- **Documentation and Tests**: Updated Store HA deployment documentation and added/extended HA tests for OpLog application, standby promotion, tenant isolation, and MasterService HA mutation behavior.

## Module

- [ ] Transfer Engine (`mooncake-transfer-engine`)
- [x] Mooncake Store (`mooncake-store`)
- [ ] Mooncake EP (`mooncake-ep`)
- [ ] Integration (`mooncake-integration`)
- [ ] P2P Store (`mooncake-p2p-store`)
- [ ] Python Wheel (`mooncake-wheel`)
- [ ] PyTorch Backend (`mooncake-pg`)
- [ ] Mooncake RL (`mooncake-rl`)
- [ ] CI/CD
- [ ] Docs
- [ ] Other

[...]

PR #2826, opened 2026-07-09
[Store] implement RFC #2756 batch-record HA OpLog writer

Icedcoco:
## Description

This PR implements the batch-record HA OpLog design described in RFC #2756 and includes the HA OpLog replication, tenant-aware standby metadata, snapshot handoff, and promotion foundation previously proposed in #2331.

#2331 will be closed in favor of this PR. This PR is intended to be reviewed and merged independently as the complete current HA OpLog implementation.

The implementation preserves the existing logical `OpLogEntry` and standby promotion model while replacing the underlying per-entry durable write protocol with batch records, an ordered writer, and a transactionally advanced durable prefix. It also integrates the protocol with current MasterService mutation paths, tenant-aware metadata, segment lifecycle replay, snapshot bootstrap, failure recovery, and promotion catch-up.

### Batch-record protocol

- Add `HaKvBackend` and an etcd txn/CAS implementation.
- Store multiple continuous `OpLogEntry` values in one `OpLogBatchRecord`.
- Advance `{batch_id, last_seq}` in `durable_prefix` atomically with the batch record.
- Validate schema versions, checksums, sequence ranges, tenant IDs, cluster IDs, and canonical batch keys at storage and codec boundaries.
- Require txn/CAS support. Non-transactional backends are not supported by batch-record HA.

### Ordered writer

- Add `OrderedOpLogWriter` with `Reserve()`, `Commit()`, and `Abort()`.
- Assign sequence IDs during `Commit()` so aborted reservations do not create gaps.
- Maintain one backend write in flight and one open waiting batch.
- Stop new admission while the current batch cannot be persisted.
- Retry the blocked batch without allowing later entries to pass it.
- Dispatch durable callbacks in logical sequence order after the durable prefix advances.

### MasterService integration

The batch writer is integrated with `PutEnd`, replica updates, segment lifecycle operations, offload/promotion completion, remove variants, replica clear, eviction, and expired cleanup paths.

Current visibility and resource semantics are:

- Additive operations may become visible on the primary before durability.
- Remove and eviction paths first mark affected replicas as `REMOVED`.
- `REMOVED` replicas are excluded from reads immediately.
- Allocator, quota, local-disk, and metadata resources are finalized from the ordered durable callback.
- Direct legacy OpLog writes are rejected while batch mode is active.

### Standby, cutover, and promotion

- Restore the existing snapshot baseline before OpLog catch-up.
- Use the maximum sequence from actual fixed-width legacy entry keys for cutover instead of trusting `/latest`.
- Run the legacy reader only until the legacy cutover target is reached, then switch to the batch reader.
- Read batch records sequentially from the earliest available batch.
- Require continuous batch IDs and entry sequences and fail closed on corruption or apply errors.
- Catch up to the latest durable prefix before promotion.
- Retry transient standby backend errors with exponential backoff. The standby enters `FAILED` after `batch_oplog_retry_timeout_sec` (180 seconds by default), allowing the supervisor to restart it.
- Retry transient final-catch-up failures within the existing promotion deadline.

### Configuration and compatibility

- `etcd_batch_record` selects the new protocol.
- Explicit `etcd` continues to select the legacy per-entry protocol.
- In etcd-enabled builds, an empty `oplog_store_type` defaults to `etcd_batch_record`.
- Legacy-to-batch cutover is supported without double-writing.
- Batch-to-legacy rollback is not supported.
- Existing snapshot restore remains supported. Batch primaries currently skip snapshot generation; moving snapshot generation to the standby is follow-up work.

### Differences from the original RFC

The implementation intentionally differs from the initial RFC in several places:

- Cutover uses the maximum actual legacy entry sequence rather than `/latest`.
- The writer currently uses two logical batches rather than a separate written/completed batch.
- Batch lookup is sequential; binary lookup by applied sequence is deferred.
- Entries retain the current full-metadata payload instead of introducing a delta schema.
- Callback execution remains ordered and asynchronous.

### Known follow-ups

The following work is intentionally outside this PR:

- Durable allocation records for `PutStart`, `UpsertStart`, `MoveStart`, and `CopyStart`.
- Replayable `put_start_release_timeout_sec` delayed-release state.
- Delta records and possible replica-level data-block lifecycle management.
- Automatic rejection of legacy startup when batch history exists.
- Leadership fencing and forced failover after extended writer failure.
- Callback-result reporting or retry semantics at the RPC boundary.
- Byte-based batch limits, durable-prefix watch, additional metrics, checkpoint lookup, and long-term snapshot/OpLog compaction.

## Module

- [ ] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [x] Documentation update
- [x] Performance improvement
- [ ] Other

## How Has This Been Tested?

The batch-record OpLog implementation adds and extends unit, integration, fault-injection, compatibility, and performance coverage.

Unit tests cover batch-record encoding and validation, durable-prefix storage, transaction handling, ordered writer admission and sequencing, retry recovery, ordered callbacks, legacy cutover, standby replay, promotion catch-up, configuration defaults, and malformed or inconsistent backend state.

MasterService HA tests cover additive and destructive mutation paths, pre-durable visibility, `REMOVED` replica classification, durable resource finalization, quota and allocator accounting, replica clear, remove variants, eviction variants, expired cleanup, segment lifecycle, offload and promotion flows, tenant isolation, snapshot bootstrap, and legacy-write rejection in batch mode.

Integration tests exercise complete primary/standby clusters using real etcd, including normal workloads, different batch sizes, process restart, backend failure and recovery, primary crashes during persistence, remove durability boundaries, standby catch-up, and promotion to the latest durable prefix.

Compatibility tests cover non-HA operation, the repository default configuration, builds without Store etcd support, legacy per-entry OpLog operation, and legacy-to-batch cutover.

Fault-injection tests verify that uncertain or failed backend writes do not advance the durable prefix or execute callbacks prematurely, that later batches cannot pass a blocked batch, and that recovery resumes from the verified durable state.

Performance tests exercise the ordered writer with multiple batch sizes and repeated real-etcd workloads, followed by automated result validation and report generation.

The relevant unit, integration, compatibility, fault-injection, and performance test suites completed successfully.

**Test results:**

- [x] Unit tests pass
- [x] Integration tests pass
- [x] Manual real-etcd and performance testing completed

## Checklist

- [x] For changes >500 LOC: I have filed RFC #2756

## AI Assistance Disclosure

- [ ] No AI tools were used
- [x] AI tools were used

AI tools assisted with implementation, tests, code review, and PR drafting. The submitter is responsible for reviewing and defending the final changes.


mooncake-store/src/master_service.cpp
```diff
@@ -9093,4 +10756,514 @@ void MasterService::HttpMetadataCleanupThreadFunc() {
...
+        // Segment lifecycle was retry-queue based in legacy mode; in
+        // batch-record mode the ordered writer owns retry until durable.
+        auto result =
+            AppendOpLogVisibleBeforeDurable(type, tenant_id, key, payload);
+        if (!result) {
+            LOG(WARNING) << why
+                         << ": segment batch OpLog persist failed for key="
+                         << key << ", type=" << static_cast<int>(type)
+                         << ", err=" << static_cast<int>(result.error());
+        }
```
00fish0: **Avoid losing segment events while the OpLog writer is non-accepting**

An entry that has already been committed to OrderedOpLogWriter is retained and retried after a backend write failure, so the issue is not loss of an already accepted entry. The gap occurs for a subsequent segment mutation while the writer is non-accepting.

After a batch write failure, the writer sets accepting=false while retrying the failed batch (FirstStorageFailureStopsNewReservations verifies this behavior). A concurrent MountSegment() or UnmountSegment() can therefore modify the local segment registry first and then fail at AppendOpLogVisibleBeforeDurable()->Reserve(). PersistSegmentOpForHAOrEnqueue() only logs that error, so the RPC still returns success even though the segment event was never admitted to the OpLog.

This is especially difficult to repair for unmount: after the local segment has already been removed, retrying the same request can take the idempotent SEGMENT_NOT_FOUND success path without emitting the missing SEGMENT_UNMOUNT.

Please reserve the OpLog entry before mutating the local segment state and abort the reservation if the mutation fails, or retain failed segment events in a retryable outbox. Please also add mount/unmount tests that force the writer into the non-accepting state before issuing the segment mutation.

mooncake-store/src/master_service.cpp
00fish0, line 1: **Enforce the OpLog admission contract consistently across all HA-visible mutations**

OrderedOpLogWriter only preserves consistency if every caller follows the same protocol:

1. **reserve** an OpLog slot before making any HA-visible local mutation;
2. abort the reservation if the local mutation fails;
3. **commit** after the mutation succeeds;
4. roll back the local mutation or retain it in a **durable** outbox if the synchronous commit fails; and 
5. never report success for a mutation whose OpLog event was not admitted.

An entry that has already been committed to the writer is retained and retried after an etcd write failure. The gap discussed here is an event that is never admitted, for example when a later Reserve() is rejected while the writer is non-accepting, or when Commit() loses a shutdown race.

The current MasterService call sites do not follow this contract consistently:

- MountSegment() mutates the segment registry before calling PersistSegmentOpForHAOrEnqueue(). The helper returns void and only logs admission errors, so the RPC still succeeds after a failed Reserve().
- ReMountSegment() remounts all segments first and then appends their events one by one through the same best-effort helper. This can produce a partially admitted list, and a retry may return early through the client_already_remounted path without repairing the missing events.
- UnmountSegment() removes the local segment before attempting OpLog admission. If admission fails, retrying the request can return through the idempotent SEGMENT_NOT_FOUND path without generating the missing SEGMENT_UNMOUNT.
- ClientMonitorFunc() commits heartbeat-timeout unmounts without emitting any SEGMENT_UNMOUNT event.
- PutEnd() marks replicas complete, updates quota/accounting and processing state, and publishes the local mutation before calling AppendOpLogVisibleBeforeDurable(). Admission failures are logged and the RPC still returns success.
- CopyEnd() reserves before mutation, but after marking target replicas complete it only logs a synchronous commit failure and continues with quota settlement and replication-task removal.
- NotifyPromotionSuccess() has the same pattern: it reserves first, but after marking the staged replica complete it only logs commit failure and continues finalizing the promotion.
- The staged-removal paths—including PersistStaleHandleCleanupForHA(), BatchReplicaClear(), PutRevoke(), EvictDiskReplica(), MoveEnd(), Remove(), RemoveByRegex(), both RemoveAll() variants, BatchRemove(), BatchEvict(), and NoFBatchEvict()—reserve before calling mark_removed(), but do not restore those statuses if the subsequent synchronous commit fails. Some return an error or skip the item, but the local state has already changed and the durable callback will never run.

Please centralize this protocol rather than requiring every caller to assemble it manually. In particular, PersistSegmentOpForHAOrEnqueue() should not be a void best-effort helper; segment operations should reserve before changing the registry and propagate or durably queue failures. Staged mutations also need rollback on synchronous commit failure, while irreversible cleanup should remain deferred to the durable callback.

00fish0 (commented):
I tested commit e372c2d1 with a 3-Master batch-record OpLog cluster while keeping the original client process and its memory alive across Primary failure. The new change correctly preserves the replica address and size, but 3 blocking issues remain.

### 1. The endpoint blacklist is not cleared after client remount

During promotion, recovered endpoints are added to invalid_replica_endpoints_. However, ReMountSegment() does not validate and remove the remounted endpoint from this set.

As a result:

- The client successfully reconnects and remounts its segment.
- The recovered descriptor still contains the correct address, size and endpoint.
- A single Get continues to return REPLICA_IS_NOT_READY because GetReplicaList() filters the endpoint around master_service.cpp:3135.
- BatchGet succeeds because the batch path does not apply the same invalid-endpoint filtering.

This makes single and batch APIs inconsistent. A successful BatchGet currently does not mean that the replica has been fully revalidated.

Suggested fix:

- During remount, validate the segment identity, generation, base address and capacity.
- After successful validation and allocator recovery, atomically remove the endpoint from invalid_replica_endpoints_.
- Apply consistent replica-validity checks to both single and batch Get paths.

### 2. The allocator does not restore occupied ranges

Although the recovered replica descriptors now preserve their original addresses, the allocator created during remount does not mark those ranges as occupied.

Before failover:
```
old single: 140316178907136, size=19
old batch0: 140316178907156, size=13
old batch1: 140316178907169, size=13
old batch2: 140316178907182, size=13
```
After failover, new writes received:
```
new single: 140316178907136, size=18
new batch0: 140316178907154, size=13
new batch1: 140316178907167, size=13
new batch2: 140316178907180, size=13
```
These allocations overlap the recovered replicas. Before the new writes, the old BatchGet passed. After the new writes, the same BatchGet returned INTERNAL_ERROR due to data verification failure.

This means preserving the descriptor alone is insufficient: new allocations can overwrite data belonging to recovered keys.

Suggested fix:
- Rebuild the allocator’s occupied intervals from recovered replica descriptors.
- Restore real allocation handles associated with the remounted allocator.
- zDo not allow new allocation or mark the endpoint readable until all recovered ranges have been validated and reserved.

The newly added unit tests verify descriptor preservation, but an end-to-end regression test should additionally verify:

1. Old single and batch keys are readable after remount.
2. New allocations do not overlap recovered ranges.
3. Old keys remain readable after new writes.

### 3. Put succeeds even when its PutEnd OpLog entry is not admitted

During the no-fault phase, 16 threads continuously performed Put followed by Get:

- 619,457 successful Puts
- 619,457 successful Gets
- 0 data mismatches
- 67,440 warnings on the Primary:
```
PutEnd: OpLog queue failed, err=-1401
```
Across all three Masters that became Primary:
```
master-0:  67,440
master-1:  61,985
master-2:  76,355
total:    205,780
```
**OrderedOpLogWriter::Reserve() returns TASK_PENDING_LIMIT_EXCEEDED when the number of open waiting slots reaches max_entries_per_batch=64, as shown in ordered_oplog_writer.cpp (line 143).**

**However, PutEnd() only logs a warning and still returns success to the client, as shown in master_service.cpp (line 3963).**

This means a write may be visible on the current Primary and acknowledged to the client without a corresponding PutEnd OpLog entry. Such metadata is not guaranteed to be recovered after failover.

The final inspector verification passed for 93,973 batches and 2,289,997 entries, with no gaps or checksum errors. However, this only proves that the records successfully written to etcd are continuous. It cannot detect acknowledged operations that were never admitted into the OpLog.

The OpLog reservation should be guaranteed before modifying visible metadata, or admission failures should be reliably retried/rolled back and returned to the client. An admission failure must not be silently converted into a successful Put.

00fish0:
> I tested commit [e372c2d](https://github.com/kvcache-ai/Mooncake/commit/e372c2d199bba01ff35a9275f38dc6a079f78014) with a 3-Master batch-record OpLog cluster while keeping the original client process and its memory alive across Primary failure. The new change correctly preserves the replica address and size, but 3 blocking issues remain.

I organized and published a standalone reproducer based directly on the current PR HEAD (e372c2d1):

Test branch: https://github.com/00fish0/Mooncake/tree/test/oplog-concurrent-failover-repro

Test commit: https://github.com/00fish0/Mooncake/commit/58763694

Instructions: https://github.com/00fish0/Mooncake/blob/test/oplog-concurrent-failover-repro/mooncake-store/tests/e2e/oplog_concurrent_failover.md

The reproducer:

- Starts a three-Master cluster using etcd_batch_record.
- Runs 16 concurrent workers, each performing a unique-key Put followed by an immediate single Get.
- Kills the current Primary.
- Waits for a replacement Primary and the same surviving client to reconnect.
- Continues the workload after reconnection.
- Compares single Get with BatchPut/BatchGet.
- Counts PutEnd OpLog admission failures and verifies the final durable OpLog prefix.

gongshanchong:
Hello, this PR is very helpful to me. When do you plan to merge it into the main branch

mooncake-store/src/master_service.cpp
Icedcoco: The different protection levels across these operations are intentional. Mooncake Store currently allows a subset of acknowledged data to be lost during failover, but it must not return corrupted or inconsistent data.

Based on that model, operations that could lead to incorrect reads receive stronger protection, while some HA bookkeeping remains best-effort to avoid blocking foreground requests. This is a deliberate tradeoff among correctness, performance, and availability, so the current design does not require every HA-visible mutation to follow one uniform admission contract.

Icedcoco:
> I tested commit [e372c2d](https://github.com/kvcache-ai/Mooncake/commit/e372c2d199bba01ff35a9275f38dc6a079f78014) with a 3-Master batch-record OpLog cluster while keeping the original client process and its memory alive across Primary failure. The new change correctly preserves the replica address and size, but 3 blocking issues remain.
> ……

Thanks for the detailed reproducer. We agree that the first two issues are real correctness problems: the remounted endpoint must be revalidated and removed from `invalid_replica_endpoints_`, and the allocator must restore the occupied ranges before accepting new allocations. We are working on fixes for both.

The third behavior is consistent with the current availability model. A `Put` may remain readable on the active Primary even if its `PutEnd` entry is not admitted to the OpLog, at the cost of that metadata not being guaranteed to survive a later failover. This intentionally sacrifices some HA-record durability to keep the foreground read/write service available. Therefore, we do not currently consider the third item a blocker for this PR.

Icedcoco:
> Hello, this PR is very helpful to me. When do you plan to merge it into the main branch

Hi @gongshanchong , thanks for your interest. We plan to merge this PR after resolving the currently known issues and completing broader, more comprehensive testing. We expect this to take roughly one week.

mooncake-store/src/master_service.cpp
Icedcoco: You are right that segment events are not strictly protected by the current OpLog admission path. The segment state currently does not have a sufficiently complete lifecycle and transition model to support strict reserve-before-mutate semantics.

For now, segment recovery relies on the promoted Master re-establishing the segment state through its heartbeat and remount communication with the surviving clients. We consider this a known limitation of the current implementation rather than something to address with an outbox in this PR.

00fish0 (commented):
I found another issue in the end-to-end failover stress test. 

I ran 16 concurrent writers, recorded every key/value for which Put() returned OK, killed the primary while writes were in progress, waited for the standby to become primary and clients to remount, stopped all writes, and then read and validated every acknowledged key through the new Master. 

**Of 25,904 acknowledged keys**, the new primary restored 24,545 objects, but the temporary NO_AVAILABLE_HANDLE errors during failover triggered three eviction rounds that evicted 21,937 keys (85.69 MiB), leaving only **2,779 acknowledged keys readable**. 

The configured eviction target was only 5%, but standby metadata does not persist lease information, and RestoreFromStandbySnapshot() reconstructs every object with the same default epoch lease_timeout. 

The eviction code uses the Nth object’s lease_timeout as a cutoff and, after reaching the target count, skips only objects whose timeout is strictly greater than the cutoff. 

Therefore, all restored objects with a timeout equal to the cutoff remain eligible, effectively making 5% a minimum rather than a limit. Repeated triggers during the remount window amplify this and can evict most of the restored dataset. 

I think the restore path should grant fresh leases, or the eviction selection should enforce an exact quota with a deterministic tie-breaker.

### Logs

The relevant log sequence is:

```
15:15:10.677 Restored from standby: 24545 objects, 1 segments, initial_seq_id=24546
15:15:10.678 Master runtime state -> serving, role=leader
15:15:11.451 [EVICT-TRIGGER] memory_ratio=0 high_watermark=0.9 need_mem_eviction=1 eviction_ratio=0.05
15:15:11.819 [EVICT-RESULT] evicted_count=11084, eviction_base=24545, actual_evict_ratio=0.451579, target_evict_ratio=0.05
15:15:11.833 [EVICT-TRIGGER] memory_ratio=0 high_watermark=0.9 need_mem_eviction=1 eviction_ratio=0.05
15:15:11.994 [EVICT-RESULT] evicted_count=5249, eviction_base=13461, actual_evict_ratio=0.389941, target_evict_ratio=0.05
15:15:12.006 [EVICT-TRIGGER] memory_ratio=0.187263 high_watermark=0.9 need_mem_eviction=1 eviction_ratio=0.05
15:15:12.303 [EVICT-RESULT] evicted_count=5604, eviction_base=8230, actual_evict_ratio=0.680923, target_evict_ratio=0.05
```

The final metrics reported:
```
Eviction: Success/Attempts=3/3, AllocFail=12836, keys=21937, size=85.69 MB
```

### How to Reproduce

Start a three-Master HA cluster and keep the storage client process alive throughout the test. Run 16 client threads continuously writing and reading 4 KiB objects, record every key for which Put() returns OK, and kill only the current Primary Master while the workload is running. Allow the same live client to reconnect to the promoted Master and remount its existing memory segment. Once remount completes, stop issuing new writes without terminating the client, and use that client to verify every acknowledged key through the new Master.

Reproduction also appears to be sensitive to the object/value size. In my runs, the issue reproduced with 4 KiB values, while runs using smaller values were much less likely to trigger it. I have not isolated the exact threshold yet, but object size changes allocator pressure, allocation behavior, and the timing of requests during the remount window. Therefore, using 4 KiB objects is currently the most reliable way to reproduce the problem.

ykwd (approved):

00fish0 (approved):

Issue #2282, opened 2026-06-01
[RFC] Unified KVCache and Model Weight Management in Mooncake Store


xianzhiT:
# [RFC] Unified KVCache and Model Weight Management in Mooncake Store

## Summary

This RFC proposes extending Mooncake Store from a KVCache-centric distributed store into a unified inference data resource layer that can manage both KVCache and model weights.

The first milestone is intentionally small: import model weights into Mooncake Store as `ObjectDataType::WEIGHT` objects, hard pin them, and let an inference engine load a model from Mooncake Store during startup.

Longer term, Mooncake can provide a type-aware resource view across KVCache and Weight objects, including quota, eviction priority, pin semantics, tiering, metrics, and scheduler-facing query APIs.

## Motivation

Large model serving consumes two major categories of reusable data:

1. KVCache, which is generated at runtime and reused across requests or instances.
2. Model weights, which are stored as model artifacts before serving and repeatedly loaded during instance startup.

Both data types consume the same physical resources: DRAM, SSD, network bandwidth, and potentially VRAM. If KVCache and model weights are managed by separate systems, each system only sees a partial resource view. This can cause resource fragmentation and local optimization. For example, one system may preserve cold model weights while another system evicts valuable KVCache because it cannot see the full resource picture.

Mooncake Store already has many primitives needed for this direction: object/tensor APIs, RDMA/GDR transfer, replicas, leases, soft/hard pin, SSD offload, copy/move tasks, and `ObjectDataType` values such as `KVCACHE` and `WEIGHT`.

The missing part is not the raw storage path. The missing part is a model-version-level Weight Management Layer:

```text
model_id / revision -> manifest -> shards / tensors -> readiness
-> serving lease -> import / query / delete / unpin
-> quota / eviction / scheduling policy
```

This RFC defines an incremental path to add that layer without building a separate resource management system.

## Goals

1. Treat model weights as first-class inference data in Mooncake Store.
2. Reuse existing Store primitives for object storage, tensor storage, transfer, replica, lease, pin, and offload.
3. Preserve the semantic differences between KVCache and model weights.
4. Start with a minimal end-to-end path: import weights, hard pin them, and load them from Store.
5. Add weight lifecycle management before adding dynamic KVCache/Weight resource competition policies.
6. Expose a unified resource view for schedulers and serving platforms in later phases.

## Non-Goals

The first version does not aim to implement:

- fallback to remote storage on Store miss
- full Weight CRUD management APIs
[...]

PR #2829, opened 2026-07-10
Add file-level model weight cache control plane

xianzhiT:
## Description

Add a file-level weight caching control plane on top of MooncakeStore, so model checkpoints can be imported into the store and served to inference engines (e.g. SGLang) as plain files — without a tensor-level, GPU-coupled save path.

The idea is to treat MooncakeStore as a **generic model cache** that is decoupled from the inference engine: weights are stored file-level (HuggingFace safetensors chunked with a per-model manifest), and engines load files exactly as they would from disk, only the source changes (disk → MooncakeStore). This control plane provides the management side — import / list / inspect / verify / delete — independently of any serving process.

The matching SGLang connector (data plane) is in sgl-project/sglang#30728. Design discussion / RFC: #2282 (Unified KVCache and Model Weight Management); this PR implements Phase 1 (store-backed weight loading) plus the import/manage control plane.

## Module

- [x] Docs

## Type of Change

- [x] New feature

## How Has This Been Tested?

**Test commands:**
```bash
cd mooncake-wheel
python -m pytest tests/test_weight_model_cache.py tests/test_weight_model_cli.py -v
```

**End-to-end (cross-machine, one storage node + one 8×H20 node):**
```bash
# import a checkpoint into MooncakeStore
python -m mooncake.weight_store.cli \
    --master-server-addr <master_ip:50051> \
    --metadata-server <metadata_ip:8080> \
    --protocol rdma --rdma-devices <mlx5_x> \
    --local-hostname <host_ip:port> \
    model import \
    --checkpoint-id Qwen3-235B-A22B \
    --model-id Qwen/Qwen3-235B-A22B \
    --revision main \
    --source /path/to/local/weights

# confirm it landed
python -m mooncake.weight_store.cli ... model list
python -m mooncake.weight_store.cli ... model inspect Qwen3-235B-A22B
python -m mooncake.weight_store.cli ... model verify  Qwen3-235B-A22B
```
Then loaded Qwen3-235B-A22B from the store via the SGLang connector in ~30s
with correct inference output.

**Test results:**
- [x] Unit tests pass
- [x] Manual testing done (cross-machine import + SGLang load + inference, above)

## Checklist

- [x] For changes >500 LOC: I have filed an RFC issue (#2282)

## AI Assistance Disclosure

AI assistance (Claude Code) was used for parts of the implementation, tests, and documentation. All changes have been reviewed and are understood by the human submitter.

### DIFF
--- a/mooncake-wheel/mooncake/weight_store/model.py
+++ b/mooncake-wheel/mooncake/weight_store/model.py
@@ -19,11 +19,15 @@ from .model_keyspace import (
     validate_checkpoint_id,
 )
 
-SCHEMA_VERSION = 1
+SCHEMA_VERSION = 2
 READY = "READY"
 FAILED = "FAILED"
 VALID_STATUSES = {READY, FAILED}
 DEFAULT_FILE_CHUNK_SIZE = 64 * 1024 * 1024
+# Mooncake's native store.remove returns -704 for an object that is already
+# gone (see structured_object_store.MISSING_OBJECT_ERROR). Treat it as success
+# so that re-running a partially-completed delete stays idempotent.
+MISSING_OBJECT_ERROR = -704
 
 WEIGHT_SUFFIXES = {
     ".bin",
@@ -42,6 +46,11 @@ class ModelFileRecord:
     size: int
     sha256: str
     chunks: List[str] = field(default_factory=list)
+    # Authoritative fixed chunk size used at import. Readers mapping file
+    # offsets to chunks MUST use this value, not their own configured chunk
+    # size. 0 means unknown/legacy (pre-schema-2) and a reader MUST refuse
+    # to guess the layout of a multi-chunk file rather than assume a size.
+    chunk_size: int = 0
 
 
 @dataclass(frozen=True)
@@ -76,12 +85,13 @@ class ModelFileManifest:
     @classmethod
     def from_json_bytes(cls, payload: bytes) -> "ModelFileManifest":
         data = json.loads(payload.decode("utf-8"))
-        data["files"] = [
-            ModelFileRecord(chunks=[item["key"]], **item)
-            if "chunks" not in item
-            else ModelFileRecord(**item)
-            for item in data["files"]
-        ]
+        records = []
+        for item in data["files"]:
+            if "chunks" not in item:
+                item = {**item, "chunks": [item["key"]]}
+            # chunk_size defaults to 0 (legacy) when absent from old dicts.
+            records.append(ModelFileRecord(**item))
+        data["files"] = records
         return cls(**data)
 
 
@@ -119,7 +129,19 @@ class ModelFileCacheClient:
         if not source.is_dir():
             raise ValueError(f"source_uri must be a directory: {source_uri}")
         if self._exists(model_manifest_key(checkpoint_id)):
-            raise ValueError(f"model checkpoint already exists: {checkpoint_id}")
+            # A manifest already exists. Only a live (READY) checkpoint blocks a
+            # re-import; a FAILED leftover -- or an unreadable/corrupt manifest
+            # -- is reclaimable, so tear it down idempotently and fall through to
+            # import afresh.
+            try:
+                existing = self._read_manifest(checkpoint_id)
+            except (KeyError, ValueError):
+                existing = None
+            if existing is not None and existing.status == READY:
+                raise ValueError(
+                    f"model checkpoint already exists: {checkpoint_id}"
+                )
+            self.delete_model(checkpoint_id)
 
         files = _list_model_files(source)
         if not files:
@@ -147,12 +169,32 @@ class ModelFileCacheClient:
                         import_started_at,
                     )
                 )
-                size, digest, chunks = self._put_file_chunks(
-                    checkpoint_id=checkpoint_id,
-                    relative_path=rel,
-                    source_path=path,
-                    file_type=file_type,
-                )
+                journal: List[str] = []
+                try:
+                    size, digest, chunks = self._put_file_chunks(
+                        checkpoint_id=checkpoint_id,
+                        relative_path=rel,
+                        source_path=path,
+                        file_type=file_type,
+                        journal=journal,
+                    )
+                except BaseException:
+                    if journal:
+                        # Preserve the partial file's chunk keys so the FAILED
+                        # manifest and delete_model can sweep any survivor whose
+                        # immediate cleanup failed (Finding 3).
+                        records.append(
+                            ModelFileRecord(
+                                path=rel,
+                                type=file_type,
+                                key=key,
+                                size=0,
+                                sha256="",
+                                chunks=list(journal),
+                                chunk_size=self.file_chunk_size,
+                            )
+                        )
+                    raise
                 records.append(
                     ModelFileRecord(
                         path=rel,
@@ -161,6 +203,7 @@ class ModelFileCacheClient:
                         size=size,
                         sha256=digest,
                         chunks=chunks,
+                        chunk_size=self.file_chunk_size,
                     )
                 )
                 imported_bytes += size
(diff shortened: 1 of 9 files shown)


mooncake-wheel/mooncake/weight_store/model_keyspace.py
```diff
@@ -0,0 +1,40 @@
...
+    return f"weight/index/model/{_sha256_text(model_id)}"
+
+
+def _sha256_text(value: str) -> str:
+    if not value:
+        raise ValueError("value must not be empty")
+    return hashlib.sha256(value.encode("utf-8")).hexdigest()
+
+
+def _validate_id(name: str, value: str) -> None:
```
Aionw, line 47: Is `a//b` a valid value here?

xianzhiT: Thanks. a//b should not be accepted as a checkpoint_id. I will tightened validation and added a regression test.

ykwd:
The CI failed for the non-consistent docs, could you take a look? Thanks

xianzhiT:
> The CI failed for the non-consistent docs, could you take a look? Thanks
Thanks for flagging this, i will fix it.

stmatengss:
@qoder Review it

Aionw (commented):
Re-review of the latest head. These four findings are new and intentionally exclude issues already covered by existing review threads.

mooncake-wheel/mooncake/weight_store/model.py
```diff
@@ -0,0 +1,559 @@
...
+                revision=revision,
+                status=READY,
+                source_uri=str(source),
+                created_at=now,
+                updated_at=now,
+                total_size=sum(record.size for record in records),
+                files=sorted(records, key=lambda record: record.path),
+                error=None,
+            )
+            self._write_manifest(manifest)
```
Aionw, line 247: **[P1] Publish `READY` only after verification.** This manifest becomes externally visible before `_add_to_indexes()` and the full chunk verification complete. A direct `mooncake://` consumer can therefore begin loading during a potentially long verification pass and observe data that is later marked `FAILED`. Please keep the checkpoint unpublished/`IMPORTING`, verify from the in-memory records, and make the `READY` manifest the final publication step.

mooncake-wheel/mooncake/weight_store/model.py
```diff
@@ -0,0 +1,559 @@
...
+
+
+@dataclass(frozen=True)
+class ModelFileRecord:
+    path: str
+    type: str
+    key: str
+    size: int
+    sha256: str
+    chunks: List[str] = field(default_factory=list)
```
Aionw, line 48: **[P1] Persist the authoritative chunk layout in the manifest.** The CLI supports a per-import `--file-chunk-size`, but the companion SGLang connector maps file offsets with its own configured/default chunk size. For example, a 32 MiB import read by a default 64 MiB consumer selects the wrong chunk. Please store `chunk_size` or explicit per-chunk offsets/sizes here and require readers to use that metadata.

mooncake-wheel/mooncake/weight_store/model.py
```diff
@@ -0,0 +1,559 @@
...
+                    self._put(chunk_key, chunk, config)
+                    chunks.append(chunk_key)
+                    digest.update(chunk)
+                    size += len(chunk)
+            except BaseException:
+                for chunk_key in chunks:
+                    try:
+                        self._remove(chunk_key, force=True)
+                    except BaseException:
+                        pass
```
Aionw, line 494: **[P2] Preserve keys whose cleanup fails.** Continuing through all removals is useful, but swallowing a removal failure loses the only cleanup record for that partial file: the current file is never appended to `records`, so the `FAILED` manifest and a later `delete_model()` cannot find the surviving chunk. I reproduced an orphan that remains even after deleting the failed checkpoint. Please retain every partial chunk key in a cleanup journal/failed record until removal succeeds.

mooncake-wheel/mooncake/weight_store/model.py
```diff
@@ -0,0 +1,559 @@
...
+                payload_by_path["model.safetensors.index.json"].decode("utf-8")
+            )
+            _validate_weight_map(index.get("weight_map", {}), by_path)
+        return manifest
+
+    def delete_model(self, checkpoint_id: str) -> None:
+        manifest = self._read_manifest(checkpoint_id)
+        for record in manifest.files:
+            self._remove_file_chunks(record)
+        self._remove(model_manifest_key(checkpoint_id), force=True)
```
Aionw, line 335: **[P2] Keep deletion retryable by removing the manifest last.** If manifest removal succeeds and `_remove_from_indexes()` then fails transiently, the catalog keeps a stale ID while a retry cannot recover the model metadata. A recoverable sequence would mark the manifest `DELETING`, remove index membership and chunks idempotently, and delete the manifest only after cleanup succeeds.

xianzhiT: Thanks. The state transition logic for import and delete has been optimized.

mooncake-wheel/mooncake/weight_store/model.py
xianzhiT: Thanks. The state transition logic for import and delete has been optimized.

mooncake-wheel/mooncake/weight_store/model.py
xianzhiT: Fixed. We now persist chunk_size in the manifest and require readers to honor it when resolving file offsets.

mooncake-wheel/mooncake/weight_store/model.py
xianzhiT: Fixed. Partial chunk keys are now preserved in a FAILED record until cleanup succeeds, so delete_model() can reliably reclaim any survivors after a failed import.

mooncake-wheel/mooncake/weight_store/model.py
```diff
@@ -0,0 +1,659 @@
...
+        checkpoint_id: str,
+        model_id: str,
+        revision: str,
+        source_uri: str,
+    ) -> ModelFileManifest:
+        validate_checkpoint_id(checkpoint_id)
+        source = Path(source_uri)
+        if not source.is_dir():
+            raise ValueError(f"source_uri must be a directory: {source_uri}")
+        if self._exists(model_manifest_key(checkpoint_id)):
```
Bo-Vincent, line 131: [P1] Fence concurrent imports for the same checkpoint

This exists check is not an atomic claim. Two importers can both pass it; if A
publishes READY and B later fails, B's exception path can overwrite A's READY
manifest with FAILED for the same `checkpoint_id`.

Please add a same-checkpoint import claim / conditional create, or explicitly
enforce a single-writer contract.

RED test: `test_concurrent_same_checkpoint_failure_does_not_replace_ready_manifest`.

mooncake-wheel/mooncake/weight_store/model.py
```diff
@@ -0,0 +1,659 @@
...
+                f"model checkpoint {checkpoint_id!r} has no readable manifest"
+            )
+        try:
+            return ModelFileManifest.from_json_bytes(payload)
+        except json.JSONDecodeError as exc:
+            raise ValueError(
+                f"model checkpoint {checkpoint_id!r} manifest is not valid JSON"
+            ) from exc
+
+    def _add_to_indexes(self, manifest: ModelFileManifest) -> None:
```
Bo-Vincent, line 392: [P1] Avoid lost updates in the model index

`_add_to_indexes()` uses a read-modify-write JSON set. Two concurrent imports
for different checkpoints can overwrite each other's update, leaving a READY
checkpoint absent from `list_models()`.

Please use an atomic/conditional update or enforce a single-writer control
plane.

RED test: `test_concurrent_imports_do_not_lose_ready_checkpoint_from_global_index`.

PR #2838, opened 2026-07-10
[Store] Speed up HugeTLB population before RDMA registration

zhangzuo21:
## Description

Large HugeTLB-backed Store segments can spend most of startup faulting pages
before RDMA registration. This PR makes parallel population the default for
RDMA Store segments using HugeTLB; no population-mode environment variable is
required.

Allocation remains lazy, then the Store faults one byte per configured
hugepage with up to 16 CPU workers immediately before `MountSegment()`.

Both Store allocation layouts are covered:

- Direct `mmap()` mappings use a generic parallel worker pool.
- NUMA-segmented mappings retain their existing `mbind()` layout. Each worker
  is scheduled on the NUMA node associated with its region before touching
  pages, preserving first-touch placement.

Non-RDMA protocols and non-HugeTLB mappings keep their existing behavior. The
mmap arena also retains its eager `MAP_POPULATE` behavior, and the existing
two-argument `allocate_buffer_mmap_memory()` symbol is preserved.

The transfer engine is unchanged: after Store population, the existing RDMA
pre-touch and automatic parallel MR registration logic still run under their
original conditions.

The same direct-mapping strategy was previously measured on qjh002 with a
500 GiB Store, reserved 2 MiB hugepages, and RDMA device `mlx5_4`:

- `MAP_POPULATE` baseline: about 55.3 seconds during setup
- Deferred parallel population: about 7.1 seconds during setup
- Deferred-mode close: about 1.2 seconds

## Module

- [ ] PyTorch Backend (`mooncake-pg`)
- [ ] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [x] Documentation update
- [ ] Other

## How Has This Been Tested?

On Ubuntu 22.04 / GCC 11.4:

```bash
cmake --build build \
  --target mmap_arena_fallback_test mooncake_store rdma_transport \
           mooncake_master \
  -j16

./build/mooncake-store/tests/mmap_arena_fallback_test
```

All selected targets link successfully and all 7 tests pass, including the
NUMA-region population test. The NUMA path was validated on qjh002, which has
two CPU/NIC NUMA nodes. The host currently has no reserved HugeTLB pages, so the
new NUMA scheduling logic was exercised with an anonymous test mapping; the
500 GiB performance numbers above are from the earlier HugeTLB benchmark.

Also run:

```bash
clang-format --dry-run --Werror <changed C++ files>
git diff --check
```

## Checklist

- [x] I have formatted my own code before submitting.
- [x] I have updated the documentation.

### DIFF
--- a/docs/source/deployment/mooncake-store-deployment-guide.md
+++ b/docs/source/deployment/mooncake-store-deployment-guide.md
@@ -811,7 +811,6 @@ Local hot cache provides a DRAM read cache on top of SSD-resident objects for fa
 |----------|---------|-------------|
 | `MC_STORE_USE_HUGEPAGE` | unset | Set `1` to request HugeTLB-backed `mmap()` |
 | `MC_STORE_HUGEPAGE_SIZE` | `2MB` | Supported: `2MB`, `1GB` |
-| `MC_DISABLE_RDMA_PRE_TOUCH` | `false` | Disable the RDMA transport's register/deregister pre-touch pass. Accepts `1`/`true`/`yes`/`on`; use with `MC_ENABLE_PARALLEL_REG_MR=1` when Store already populated the mapping and parallel MR registration should remain enabled |
 | `MC_MMAP_ARENA_POOL_SIZE` | unset | Pre-allocated arena pool size (e.g., `8gb`). Explicitly set to enable the arena |
 | `MC_DISABLE_MMAP_ARENA` | unset | Disable arena, fall back to per-call `mmap()`. Accepts `1`/`true`/`yes`/`on` (or `0`/`false`/`no`/`off`) |
 
@@ -822,14 +821,6 @@ required:
 ```bash
 export MC_STORE_USE_HUGEPAGE=1
 export MC_STORE_HUGEPAGE_SIZE=2MB
-```
-
-To skip the transfer engine's separate register/deregister pre-touch pass while
-retaining parallel MR registration, optionally set:
-
-```bash
-export MC_DISABLE_RDMA_PRE_TOUCH=1
-export MC_ENABLE_PARALLEL_REG_MR=1
 ```
 
 For direct mappings, workers divide the mapping into page ranges. For
--- a/docs/source/deployment/mooncake-store-deployment-guide.md
+++ b/docs/source/deployment/mooncake-store-deployment-guide.md
@@ -811,18 +811,23 @@ Local hot cache provides a DRAM read cache on top of SSD-resident objects for fa
 |----------|---------|-------------|
 | `MC_STORE_USE_HUGEPAGE` | unset | Set `1` to request HugeTLB-backed `mmap()` |
 | `MC_STORE_HUGEPAGE_SIZE` | `2MB` | Supported: `2MB`, `1GB` |
-| `MC_STORE_HUGEPAGE_POPULATE_MODE` | unset | Set `parallel` to populate RDMA Store HugeTLB mappings with parallel CPU writes immediately before transfer-engine registration. NUMA mappings use node-local workers |
 | `MC_DISABLE_RDMA_PRE_TOUCH` | `false` | Disable the RDMA transport's register/deregister pre-touch pass. Accepts `1`/`true`/`yes`/`on`; use with `MC_ENABLE_PARALLEL_REG_MR=1` when Store already populated the mapping and parallel MR registration should remain enabled |
 | `MC_MMAP_ARENA_POOL_SIZE` | unset | Pre-allocated arena pool size (e.g., `8gb`). Explicitly set to enable the arena |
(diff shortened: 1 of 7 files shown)


alogfans:
CI build error:
```
/usr/bin/ld: /home/runner/work/Mooncake/Mooncake/build/mooncake-transfer-engine/src/libtransfer_engine.a(rdma_transport.cpp.o): in function `mooncake::RdmaTransport::registerLocalMemoryInternal(void*, unsigned long, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool, bool, bool)':
rdma_transport.cpp:(.text+0x6454): undefined reference to `mooncake::Environ::GetBool(char const*, bool)'
```

he-yufeng (approved):
Reviewed as a maintainer: I traced the population path and the register-side interaction rather than just skimming the diff. This is a clean, well-scoped optimization and it holds up.

What I checked:

**Parallel fault-in is thread-safe.** `touch_mmap_pages` partitions pages into disjoint `[begin_page, end_page)` ranges (ceil split, trailing threads get empty ranges), so workers only ever write distinct `page * page_size` addresses. No shared bytes, no race. The `volatile char*` is load-bearing here: without it the compiler is free to drop the dead stores. The final `data[map_size-1] = 0` runs after every join on the main thread, so it doesn't race either.

**The hugepage stride is correct.** `populate_hugetlb_mapping` touches one byte every `hugepage_size`, and since a HugeTLB page faults atomically, one write per 2 MiB page brings in the whole page — equivalent to `MAP_POPULATE`, just parallel. `mmap_map_size(total_size, hugepage_size)` recomputes the same aligned size `allocate_buffer_mmap_memory` mapped, so it touches exactly the mapping and nothing past it.

**The defer / arena / populate interaction is consistent.** The three-arg `allocate_buffer_mmap_memory(..., defer=true)` only drops `MAP_POPULATE` on the direct-hugepage path; if the allocation instead lands in the eagerly-populated arena, `real_client` skips the manual populate via `!is_mmap_arena_allocation(ptr)`. So a mapping is never double-populated and never left un-touched, and NUMA-segmented mappings stay on their `ibv_reg_mr`-driven first-touch path.

**No new SIGBUS risk.** `MAP_HUGETLB` reserves the huge pages at `mmap()` time, so if allocation succeeded the manual first-touch can't fault on an exhausted pool — the same guarantee `MAP_POPULATE` gives.

**ABI is preserved** — the two-arg `allocate_buffer_mmap_memory` still exists and forwards to the new overload with `defer=false`, and the default (non-`rdma`) path is the old behavior unchanged. The `MC_DISABLE_RDMA_PRE_TOUCH` wiring in `rdma_transport.cpp` cleanly lets an operator drop the register-side pre-touch when the Store has already populated, avoiding a duplicate pass.

Tests cover the gating (`should_defer_hugetlb_population` needs both hugepages and `rdma` mode) and the stride (`populate_hugetlb_mapping` zeroes one byte per 2 MiB page plus the tail). The 55 s to 7 s number on a 500 GiB segment is a real win.

Two non-blocking notes: the 16-thread cap is a sensible hardcoded default but might be worth an env knob later for very-high-core hosts; and the trailing `data[map_size-1]` write is redundant given `page_count` already covers the last page.

LGTM.

zhangzuo21:
Pushed follow-up commit 68d01b49: fixed the reported Environ::GetBool linker failure, renamed the opt-in value to MC_STORE_HUGEPAGE_POPULATE_MODE=parallel, and extended deferred population to the default mbind-based NUMA layout with node-local workers. The updated targets link on Ubuntu/GCC 11 and all 8 mmap_arena_fallback_test cases pass on qjh002.

zhangzuo21:
Follow-up 1df7bb0a removes MC_STORE_HUGEPAGE_POPULATE_MODE entirely. Parallel population is now automatic whenever RealClient sets up an RDMA Store segment with HugeTLB; direct and mbind-based NUMA layouts are covered, while non-RDMA/non-HugeTLB paths and the mmap arena remain unchanged. Rebuilt mooncake_store/mooncake_master and reran all 7 relevant tests on qjh002 successfully.

zhangzuo21:
Follow-up b5dedcde removes MC_DISABLE_RDMA_PRE_TOUCH and all related documentation. rdma_transport.cpp now has no diff from the PR base, so the upstream RDMA pre-touch and automatic parallel MR registration behavior remain intact. HugeTLB population is still automatic, parallel, and NUMA-aware. Rebuilt rdma_transport/mooncake_master and reran all 7 relevant tests on qjh002 successfully.

alogfans (approved):

PR #2842, opened 2026-07-10
[TE] Make ThreadLocalStorage per-instance and reclaim per-thread holders

### LINKED ISSUE #2717: [TE] ThreadLocalStorage is per-type (instances alias each other's thread-local state) and never destroys holders
### Describe the bug

While evaluating `ThreadLocalStorage<T>` (tent/include/tent/common/concurrent/thread_local_storage.h) for PR #2714's review, two implementation issues surfaced. Both were verified with minimal programs against the in-tree header on x86_64 Linux with a recent GCC toolchain.

**1. Storage is per-*type*, not per-instance.** `instance_` is a `thread_local static` member of the class template (one slot per `T` per thread), shared by every `ThreadLocalStorage<T>` object:

```cpp
ThreadLocalStorage<Cache> a, b;
a.get().value = 42;
// prints: a.get().value=42 b.get().value=42 same_object=YES (aliased!)
```

Consequence for the existing user: `SegmentManager::tl_remote_cache_` is a member, so **two `SegmentManager` instances in one process share one `RemoteSegmentCache` per thread**. The first manager to call `get()` on a thread wins; the second manager's `getRemoteCached()` then reads/writes the first one's cache map. Since `SegmentID`s are allocated per manager (`next_id_` both start at 1), handle collisions can serve the wrong remote desc. Reachable whenever one process hosts two TENT engines (e.g. tests that spawn server+client in-process).

**2. Holders are never destroyed, so the cleanup half of the class is dead code.** `get()` allocates `new InstanceHolder(this)` into a raw `thread_local` pointer; nothing ever deletes it, so `~InstanceHolder` (which deregisters from `instances_`) never runs and each thread leaks one holder per type. LeakSanitizer on a trivial user:

```
==…==ERROR: LeakSanitizer: detected memory leaks
Direct leak of 16 byte(s) … ThreadLocalStorage<Cache>::get() thread_local_storage.h:39
```

This also means `instances_` only ever grows, and `forEach` visits entries belonging to threads that have long exited. (Had holders *been* destroyed at thread exit, there would instead be a use-after-free when a thread outlives the owning `ThreadLocalStorage` — `~InstanceHolder` dereferences `owner` — so the destructor logic needs rework either way, e.g. a `unique_ptr` thread_local + owner keyed by id/weak ref, or an instance-keyed map slot.)

Happy to send a fix PR if the maintainers agree on a direction (instance-keyed `thread_local` map + owned holders is the minimal one).

### Environment

- Commit 812eb8d (current main at the time of filing); demos run on x86_64 Linux with a recent GCC toolchain.

n-WN:
## Description

Fixes #2717.

`ThreadLocalStorage<T>` kept its thread-local slot as a `thread_local static` member of the class template — one slot per template *instantiation*, shared by every instance in the process — so two instances (e.g. two engines' `SegmentManager::tl_remote_cache_`) aliased each other's per-thread state: whichever instance first called `get()` on a thread owned the slot, and the second instance read/wrote the first one's cache. The slot was also a raw pointer to a heap holder that nothing ever deleted, so the deregistration logic in the holder's destructor was unreachable and every thread leaked one holder per instantiation (`ASAN: 288 bytes in 18 allocations` in the new test binary against the old header).

This reworks the storage to be genuinely per-instance:

- **Per-instance keying**: the per-thread state is a map keyed by a process-monotonic instance id (ids are never reused, so a destroyed-then-reallocated storage can never be served a stale entry), with a one-entry cache in front so the common `get()` stays one `thread_local` access plus an id compare.
- **Owned values**: a thread's `T` values are owned by that thread's map and destroyed at thread exit — no leaked holders.
- **Both teardown orders are safe** via a control block jointly owned by the storage and every thread node: a thread exiting while the owner is alive deregisters from the registry (the path that was unreachable before); an owner destroyed while threads live marks the block dead, and those threads skip the registry at exit.
- **Orphans are swept, not accumulated**: an adversarial-review pass on the first draft found that a long-lived thread would retain one orphaned `T` per destroyed storage until thread exit — under engine create/destroy churn that silently pins one `RemoteSegmentCache` (and every `SegmentDescRef` snapshot inside it) per destroyed engine per thread. The cold path of `get()` (first use of an instance on a thread) now sweeps all of the thread's dead-owner values, so orphan count is bounded by live instances; a churn regression test pins this (1000 create/use/destroy rounds, ≤2 live values throughout).
- **`forEach()`** runs under the registry mutex and visits exactly the values of live registered threads (previously it could visit holders of exited threads, and its cleanup half was dead code). It still synchronizes membership only, not `T`'s contents — unchanged semantics for the existing caller.

The public API (`get()`, `forEach(const std::function<void(T&)>&)`) is unchanged; `SegmentManager` needs no modification.

## Module

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

Multi-core x86_64 Linux test host with a recent GCC toolchain.

**Test commands:**
```bash
cmake -S . -B build-tls-normal -DWITH_STORE=OFF -DWITH_STORE_RUST=OFF \
  -DUSE_TENT=ON -DUSE_CUDA=OFF -DBUILD_UNIT_TESTS=ON
cmake --build build-tls-normal --target thread_local_storage_test -j16
build-tls-normal/mooncake-transfer-engine/tent/tests/thread_local_storage_test
ctest --test-dir build-tls-normal/mooncake-transfer-engine/tent/tests -j1
# Separate Debug builds used -fsanitize=address,undefined and -fsanitize=thread.
```

**Results:**

| build | result |
|---|---|
| new impl, `thread_local_storage_test` | **8/8 × 100** |
| new impl, ASAN/UBSAN/LSAN | **8/8 × 50**, no reports |
| new impl, TSAN | **8/8 × 50**, 0 warnings |
| full TENT suite (includes `SegmentManager` integration via `tl_remote_cache_`) | **23/23** |
| **old impl + same test binary** (control) | **5/7 fail** of the applicable tests (aliasing, thread-exit deregistration, instance resurrection, forEach liveness, churn) + **LSAN: 288 B / 18 allocs leaked** |

`get()` hot path at `-O3`: 0.33–0.41 ns before → 0.66–0.69 ns after (+~0.3 ns; the caller's next step in `getRemoteCached()` is a `clock_gettime` and a map lookup, so this is noise). The orphan sweep runs only on the first-use cold path. The unit test pins a loose ceiling (skipped under sanitizers, where wall-clock asserts flake) to catch lock-on-hot-path-class regressions.

- [x] Unit tests pass
- [x] Manual testing done (sanitizer matrix above)

## Checklist

- [x] Focused pre-commit hooks pass on every touched C/C++ file; the full-file `cmake-format` hook also proposes unrelated reflows in the upstream TENT test CMake file, which are intentionally excluded per `AGENTS.md`

## AI Assistance Disclosure

Claude (Anthropic) was used for implementation and test drafting under my direction; I reviewed every changed line, and the design (per-instance keying, control-block teardown) follows the sketch I posted on #2717 before implementation. All numbers above are from real runs on my hardware.

---

cc @catyans — as discussed on the issue, a careful pass on the teardown paths would be much appreciated; the concurrency stress slice is also still yours if you want to extend it.

### DIFF
--- a/mooncake-transfer-engine/tent/include/tent/common/concurrent/thread_local_storage.h
+++ b/mooncake-transfer-engine/tent/include/tent/common/concurrent/thread_local_storage.h
@@ -15,68 +15,176 @@
 #ifndef TENT_THREAD_LOCAL_STORAGE_H
 #define TENT_THREAD_LOCAL_STORAGE_H
 
+#include <atomic>
+#include <cstdint>
 #include <functional>
+#include <memory>
 #include <mutex>
-#include <thread>
+#include <unordered_map>
 #include <unordered_set>
 
 namespace mooncake {
 namespace tent {
 
+namespace detail {
+// Hoisted out of the class template so ids are unique across ALL
+// ThreadLocalStorage instantiations, not just within one T. The per-thread
+// maps are per-T, so per-T uniqueness would suffice today — global
+// uniqueness removes the aliasing hazard if the per-thread state is ever
+// shared across instantiations.
+inline uint64_t nextThreadLocalStorageId() {
+    static std::atomic<uint64_t> counter{1};
+    return counter.fetch_add(1, std::memory_order_relaxed);
+}
+}  // namespace detail
+
+// Per-instance thread-local storage (#2717).
+//
+// Each (ThreadLocalStorage instance, thread) pair owns a distinct T. The
+// previous implementation kept one `thread_local` slot per template
+// instantiation, so all instances of ThreadLocalStorage<T> in a process
+// aliased each other's per-thread state, and its heap-allocated holders were
+// never destroyed (the deregistration path was unreachable).
+//
+// Lifetime rules:
+// - A thread's T values are destroyed when the thread exits.
+// - Destroying the storage does not destroy other threads' values; they are
+//   orphaned (at most one T per thread per destroyed storage) and reclaimed
+//   at those threads' exit. Instance ids are process-monotonic and never
+//   reused, so an orphaned value can never be served to a new instance that
+//   happens to reuse the same address.
+// - Both teardown orders are safe: a thread exiting while the owner is alive
+//   deregisters its value from the owner's registry; an owner destroyed
+//   while threads are alive marks the jointly-owned control block dead, and
+//   those threads skip deregistration at exit. Orphans are additionally
+//   swept opportunistically: the next first-use get() of ANY instance on
+//   that thread reclaims all of the thread's dead-owner values, so orphan
+//   count stays bounded by live instances even under storage churn.
+//
+// Precondition: callers must keep the storage alive across every get() /
+// forEach() call (the usual member-of-owner pattern satisfies this); the
+// destructor may run concurrently only with other threads' exits, not with
+// their accesses.
+//
+// Concurrency:
+// - get() is lock-free after the first call per (instance, thread): one
+//   thread_local access plus an id compare on the hot path.
+// - forEach() runs under the registry mutex and visits exactly the values of
+//   live registered threads. It synchronizes registry membership only — if
+//   owning threads mutate their T concurrently, the callback observes those
+//   fields with whatever synchronization T itself provides (unchanged from
+//   the previous implementation).
 template <typename T>
 class ThreadLocalStorage {
    public:
     ThreadLocalStorage() = default;
-    ~ThreadLocalStorage() = default;
 
-    // Disable copy/move
+    ~ThreadLocalStorage() {
+        std::lock_guard<std::mutex> lock(control_->mutex);
+        control_->owner_alive.store(false, std::memory_order_release);
+        control_->values.clear();
+    }
+
     ThreadLocalStorage(const ThreadLocalStorage&) = delete;
     ThreadLocalStorage& operator=(const ThreadLocalStorage&) = delete;
 
-    // Access the thread-local instance (lock-free)
+    // Access the calling thread's instance, constructing it on first use.
     T& get() {
-        if (!instance_) {
-            instance_ = new InstanceHolder(this);
+        ThreadState& state = threadState();
+        if (state.cached_id == id_) return *state.cached_value;
+        auto it = state.nodes.find(id_);
+        if (it == state.nodes.end()) {
+            // First use of this instance on this thread — the cold path.
+            // Piggyback a sweep of values whose owners are gone, so a
+            // long-lived thread does not accumulate one orphan per
+            // destroyed storage it ever touched (their T contents would
+            // otherwise stay pinned until thread exit).
+            sweepDeadNodes(state);
+            it = state.nodes.try_emplace(id_).first;
+            ThreadNode& node = it->second;
+            node.control = control_;
+            // The caller holds a reference to *this, so the owner is alive
+            // and registration cannot race ~ThreadLocalStorage's clear().
+            std::lock_guard<std::mutex> lock(control_->mutex);
+            control_->values.insert(&node.value);
         }
-        return instance_->value;
+        state.cached_id = id_;
+        state.cached_value = &it->second.value;
+        return *state.cached_value;
     }
 
-    // Safe iteration over all instances (with locking)
+    // Safe iteration over the values of all live threads that have called
+    // get() on this instance.
     void forEach(const std::function<void(T&)>& fn) {
-        std::lock_guard<std::mutex> lock(global_mutex_);
-        for (auto* inst : instances_) {
-            fn(inst->value);
+        std::lock_guard<std::mutex> lock(control_->mutex);
+        for (T* value : control_->values) {
+            fn(*value);
         }
     }
 
    private:
-    struct InstanceHolder {
-        T value;
-        ThreadLocalStorage<T>* owner;
+    // Shared between the owner and every thread node so that whichever side
+    // is torn down last still has a valid registry (or a dead flag) to look
+    // at.
+    struct ControlBlock {
+        std::mutex mutex;
+        std::unordered_set<T*> values;
+        // Atomic so the orphan sweep can test liveness without taking the
+        // mutex of every node it scans.
+        std::atomic<bool> owner_alive{true};
+    };
 
-        InstanceHolder(ThreadLocalStorage<T>* owner) : owner(owner) {
-            std::lock_guard<std::mutex> lock(owner->global_mutex_);
-            owner->instances_.insert(this);
-        }
+    struct ThreadNode {
+        T value{};
+        std::shared_ptr<ControlBlock> control;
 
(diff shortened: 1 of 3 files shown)


mooncake-transfer-engine/tent/include/tent/common/concurrent/thread_local_storage.h
```diff
@@ -15,70 +15,178 @@
...
+    };
 
-    // Global list of all thread instances
-    std::unordered_set<InstanceHolder*> instances_;
-    std::mutex global_mutex_;
-};
+    static void sweepDeadNodes(ThreadState& state) {
+        for (auto it = state.nodes.begin(); it != state.nodes.end();) {
+            if (!it->second.control->owner_alive.load(
+                    std::memory_order_acquire)) {
```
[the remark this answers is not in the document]
n-WN: Applied in c391e16. For the record, the dereference is unreachable today — `sweepDeadNodes` runs strictly before `try_emplace` on the same thread (thread-local state, no cross-thread visibility), and `control` is assigned immediately after emplace by a nothrow `shared_ptr` copy, so no node with a null control can exist when a sweep observes the map. Took the check anyway: it's free, consistent with `~ThreadNode`'s defensive check, and keeps the invariant local rather than depending on statement ordering in `get()`.

n-WN:
Rebased cleanly onto current `main` (`4188e3ae`; the only conflict was the TENT test CMake list, resolved by retaining the new upstream deadline/promotion tests plus `thread_local_storage_test`) and force-pushed signed commits. Fresh validation on a dedicated Linux test host using the rebased bytes: normal 8/8 × 100, ASAN/UBSAN/LSAN 8/8 × 50, TSAN 8/8 × 50, and the complete CPU/RDMA-capable TENT test directory 23/23. Format and typos checks are already green; the full CI matrix is now running. @catyans the teardown/churn paths are ready for your pass.

he-yufeng (approved):
Reviewed as a maintainer: I read the full header rework and traced the concurrency rather than just skimming the diff. This is a clean fix for a real bug and the design holds up.

What I checked:

**Root cause is right.** `thread_local static InstanceHolder* instance_` is one slot per `ThreadLocalStorage<T>` *instantiation*, so every instance sharing a `T` aliased the same per-thread pointer, and the holder leaked because its deregistration path was unreachable. `InstancesDoNotAlias` plus the LSAN control run against the old header nail both halves.

**The lockless sweep is sound.** `sweepDeadNodes` reads `control->owner_alive` without the mutex, which only works because each instance owns a distinct `ControlBlock` whose `owner_alive` goes true to false exactly once and never resurrects (ids are monotonic and never reused, so a reallocated storage gets a fresh block), and the scanning thread keeps that block alive through the node's `shared_ptr`. The acquire/release pairing with the store under `~ThreadLocalStorage`'s mutex is correct, and racing the sweep against owner teardown converges either way (worst case one extra round before reclaim).

**Both teardown orders are covered.** Thread-exit-first deregisters under the registry mutex; owner-first marks the block dead and clears the registry, and later `~ThreadNode`s see the dead flag and skip it. `ThreadExitDeregistersAndReclaims` and `OwnerDestroyedBeforeThreadExitIsSafe` exercise both.

**No dangling.** `unordered_map` guarantees element-pointer stability across rehash, so the registry's `&node.value` and the one-entry cache stay valid; the sweep resets the cache when it evicts the cached node; and `~ThreadNode` releases the mutex before `T`'s destructor runs, so `forEach` can't observe a value mid-destruction.

The 8 tests map one-to-one onto the invariants (aliasing, both teardown orders, id non-resurrection, forEach liveness, orphan-sweep bound, churn stress), and the ASAN/UBSAN/TSAN matrix with the old-impl control makes the concurrency claims credible rather than just asserted. The self-caught orphan-accumulation fix is a nice touch.

One non-blocking note: `sweepDeadNodes` is O(nodes on this thread) on every first-use cold miss, so a thread that has touched many live instances pays that scan on each new-instance miss. That's fine given it's bounded by live instances and off the hot path; just flagging it in case per-thread instance fan-out ever grows.

LGTM.

mooncake-transfer-engine/tent/include/tent/common/concurrent/thread_local_storage.h
```diff
@@ -15,70 +15,183 @@
...
-    // Safe iteration over all instances (with locking)
+    // Safe iteration over the values of all live threads that have called
+    // get() on this instance.
     void forEach(const std::function<void(T&)>& fn) {
-        std::lock_guard<std::mutex> lock(global_mutex_);
-        for (auto* inst : instances_) {
-            fn(inst->value);
+        std::lock_guard<std::mutex> lock(control_->mutex);
+        for (T* value : control_->values) {
+            fn(*value);
```
alogfans, line 121: Lock is hold during fn() call. Is this in purpose?

n-WN: Yes, intentional — the lock is what makes the `T&` passed to `fn` valid for the duration of the call. A thread exiting deregisters `&node.value` under this same mutex (and `~ThreadNode` only destroys the `T` after releasing it), so holding the lock across `fn` is precisely the guarantee that `fn` never observes a dangling or mid-destruction value. A snapshot-then-call-unlocked variant would lose that: once the lock is released, any thread can exit and destroy the value behind the copied pointer, and making that safe would need per-value refcounting — heavier than warranted here.

Two things worth noting relative to the implementation this replaces:
- The previous `forEach` had the same hold-lock-during-`fn` contract, but on a single **global** mutex — so `fn` blocked (and could deadlock against) every `ThreadLocalStorage<T>` instance of that type. Here the lock is per-instance, so the constraint narrows to: `fn` must not re-enter `forEach` on the *same* instance or trigger a first-use `get()` of it on the same thread.
- I checked the tree: there are currently **no production callers** of `forEach` (the only in-tree instance, `SegmentManager::tl_remote_cache_`, uses `get()` only; `forEach` is exercised by the tests). So the reentrancy constraint binds no existing code today — which is exactly why it is worth documenting at the declaration before a future caller trips on it. I will add that comment.

n-WN:
Pushed two small commits after an adversarial re-review pass: `cce2a389` makes first-use registration exception-safe (if the registry-set insert throws, the already-committed map entry is rolled back before rethrowing — previously a caller that caught the `bad_alloc` and retried `get()` would be handed a value `forEach()` could never see), and `815ddcef`/`cce2a389` complete the `forEach` contract at the declaration: beyond same-instance reentry, it now also spells out that blocking on the exit of a registered thread from inside `fn` (its exit cleanup takes the same mutex) or nesting `forEach` across two instances in opposite orders deadlocks just as deterministically. @alogfans this supersedes the shorter comment I promised in the thread above. Suite still 8/8 on the H100 box.

PR #2845, opened 2026-07-10
[TENT] Add QoS metrics baseline to tebench

catyans:
## Summary

- add a generic `--qos_classes=name:threads:slo_us:weight[:isolated_gbps],...` workload contract to `tebench`
- report per-class SLO attainment, P99 latency, goodput, and isolation leakage
- report aggregate weighted goodput, weighted Jain fairness, and total link utilization
- optionally append schema-versioned JSONL records; unavailable baseline-dependent metrics are emitted as `null`/`N/A`
- retain link capacity and isolated baselines in JSONL so all derived metrics are reproducible
- document metric definitions and matching isolated/mixed-run requirements

## Metric semantics

- SLO attainment is the fraction of completed batches with end-to-end transfer latency at or below the configured class SLO.
- Jain fairness is computed over per-class `throughput / weight`.
- Isolation leakage is `max(0, 1 - mixed_throughput / isolated_throughput)` and is only emitted when the caller supplies a matching isolated baseline.
- Total utilization requires an explicit usable link capacity.

## Validation

- classic and `USE_TENT=ON` `tebench` targets build successfully with GCC 13.3 in a clean Ubuntu 24.04 container.
- `tebench_qos_metrics_test` passes in both builds, covering contract parsing, validation, metric math, JSON validity, and null semantics.
- TENT SHM target/initiator smoke completed about 1 million mixed batches; isolated-to-mixed recomputation matched all six derived metrics exactly.
- classic TCP target/initiator smoke passed and produced valid two-class JSONL.
- cross-host TENT/TCP on `sinian-public-1` -> `sinian-platform` passed. With 64 KiB x batch 4 and four workers, the strict 1 ms foreground SLO attained 17.65%; Jain fairness was 0.8338, maximum isolation leakage 41.21%, and utilization 68.45%. All values were independently recomputed from the JSONL inputs.
- RDMA-enabled build and metric tests pass; both hosts discover five HCAs. Cross-host RDMA transfer is not claimed as passed because the available backup pair reports peer GID/bootstrap mismatch and WQE retry timeouts.
- format, codespell, Sphinx, and path checks pass.

## Dependency

None. The branch is rebased on current `main`; QoS SLOs are reporting thresholds and do not change request scheduling policy.


catyans:
This replaces #2841 after renaming the fork branch from `codex/qos-benchmark-baseline` to `qos-benchmark-metrics` to comply with branch naming policy. The code and commit history are unchanged; prior review discussion and cluster validation are recorded on #2841.

mooncake-transfer-engine/benchmark/utils.h
```diff
@@ -107,6 +113,14 @@ struct XferMetricStats {
...
 
+    double fractionAtOrBelow(double threshold) const {
+        if (samples.empty()) return 0.0;
+        const auto count = std::count_if(
+            samples.begin(), samples.end(),
+            [threshold](double value) { return value <= threshold; });
+        return static_cast<double>(count) / samples.size();
+    }
+
     void add(double value) { samples.push_back(value); }
```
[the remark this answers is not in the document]
catyans: Fixed in f2dd7c8. XferMetricStats now has a bulk vector overload backed by vector::insert, with a focused unit test covering count/min/max/average after bulk insertion.

mooncake-transfer-engine/benchmark/main.cpp
```diff
@@ -78,21 +101,51 @@ int processBatchSizes(BenchRunner& runner, size_t block_size, size_t batch_size,
...
-        mutex.unlock();
+        auto& group_stats = tight ? tight_stats : loose_stats;
+        group_stats.total_duration.add(total_duration);
+        for (auto val : transfer_duration)
+            group_stats.transfer_duration.add(val);
+        if (qos_enabled) {
+            qos_stats[qos_class].total_duration.add(total_duration);
+            for (auto val : transfer_duration)
+                qos_stats[qos_class].transfer_duration.add(val);
+        }
```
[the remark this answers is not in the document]
catyans: Fixed in f2dd7c8. All three transfer-duration accumulation loops now use the bulk add overload while the existing outer mutex still provides synchronization, substantially shortening the locked hot-path work.

catyans:
Follow-up validation for the review fix at `f2dd7c8`:

- built `tebench_qos_metrics_test` with TENT enabled
- `ctest -R tebench_qos_metrics_test --output-on-failure`: 1/1 passed
- clang-format 20 check passed for the changed files

All review threads are resolved. Full GitHub CI is now running/queued.

docs/source/design/tent/tebench.md
```diff
@@ -91,6 +91,56 @@ Each output row corresponds to one benchmark configuration.
...
+an SLO-constrained foreground class and twelve workers to a best-effort
+checkpoint class:
+
+```bash
+./tebench \
+  --target_seg_name=<SEG> \
+  --backend=tent \
+  --start_num_threads=16 \
+  --max_num_threads=16 \
+  --qos_classes=foreground:4:1000:4:12.5,checkpoint:12:0:1:10.0 \
```
alogfans, line 125: An advice: in addition to this format, you can also support json format for better readability.

mooncake-transfer-engine/tent/include/tent/common/qos_metrics.h
alogfans, line 1: As qos_metrics.{h,cpp} are common components, I'd prefer to put them in the tent/src/common directory.

alogfans:
Thanks for the change — the direction on deadline-aware arbitration in `tent/transport/rdma` makes sense.  

My main concern is layering: we want **`tebench` to stay focused on benchmarking/testing**, while **all TENT execution logic lives under `mooncake-transfer-engine/tent/`**.

In the current PR, a substantial part of the deadline/QoS logic is implemented in `benchmark/`, for example:
 - QoS class contract parsing and validation
 - thread-to-class mapping
 - converting SLO/deadline configuration into per-request `deadline_ns`
 - QoS metric semantics and computations (`slo_attainment`, `goodput`, `weighted_goodput`, `jain_fairness`, `isolation_leakage`, `total_utilization`)

Those are not just benchmark presentation details; they define TENT-facing workload semantics and policy, so they would be better placed under `tent/`, with `tebench` only acting as:
 - workload driver / test harness
 - request issuer
 - result printer / optional serializer

Concretely, I’d suggest:
 - move `qos_metrics.{h,cpp}` out of `benchmark/` and into `tent/`
 - keep `tebench` responsible only for wiring inputs to TENT APIs and reporting outputs
 
The RDMA arbitration policy itself is already in the right direction by living under `tent/transport/rdma`. I think the rest of the QoS/deadline semantics should follow the same layering.

docs/source/design/tent/tebench.md
catyans: Implemented in c02a578f. tebench now supports `--qos_classes_json` as a JSON-array alternative to the compact string format, with docs and a focused parser test (`ParsesJsonClassContract`). The old `--qos_classes` format remains supported, and the two flags are mutually exclusive to avoid ambiguous configuration.

Validation on the H20/RDMA cluster container (lingjun-100): built `tebench` and ran `tebench_qos_metrics_test`, 6/6 tests passed.

mooncake-transfer-engine/tent/include/tent/common/qos_metrics.h
catyans: Done in c02a578f. The reusable metrics code now lives under `tent/include/tent/common/qos_metrics.h` and `tent/src/common/qos_metrics.cpp`.

I also split out the benchmark-specific conversion from `XferBenchStats` into `benchmark/qos_metrics_adapter.{h,cpp}`, so the common TENT metrics component no longer depends on benchmark-only types. This keeps the common layer reusable by other TENT components while preserving the existing tebench behavior.

Validation on the H20/RDMA cluster container (lingjun-100): built `tebench` and ran `tebench_qos_metrics_test`, 6/6 tests passed.

staryxchen (approved):
LGTM

Issue #2843, opened 2026-07-10
[TransferEngine] TcpTransport benchmark note: VRAM-backed transfers pay per-chunk staging; original "2x host-path loss on USE_CUDA" was a benchmark-config artifact


n-WN:
### Bug Report / Performance

**TcpTransport throughput drops ~2x when the build enables USE_CUDA, even for pure host-memory transfers** — measured on loopback with `transfer_engine_bench` (write, 4 threads × batch 32, duration 3–4 s, same box, same commit `f1f5e40`, only the build flag differs):

| block size | build without CUDA | build with `-DUSE_CUDA=ON` |
|---|---|---|
| 64 KB | 0.99 GB/s | 0.84–0.88 GB/s |
| 256 KB | 2.10 GB/s | 1.09–1.28 GB/s |
| 1 MB | **3.41 GB/s** | **1.53–1.64 GB/s** |

All buffers are plain DRAM (`numa_alloc`/`malloc`); no GPU memory is involved in the transfer.

### Root cause (from reading `tcp_transport.cpp`)

Under `USE_CUDA` (and the other accelerator macros), both `ClientSession::writeBody`/`readBody` and `ServerSession::writeBody`/`readBody` call `getCudaDeviceId(addr)` → `cudaPointerGetAttributes()` **once per 64 KB chunk** (not once per request), on both ends of the connection. For a 1 MB transfer that is 16 chunks × 2 sides = 32 CUDA runtime calls per request; `cudaPointerGetAttributes` costs on the order of a microsecond and serializes on CUDA runtime internals, which is enough to halve throughput at these block sizes. The result of the lookup cannot change between chunks of one request — the address range is fixed at submit time.

### Suggested fix

Resolve the memory type **once per request** (at header processing / `initiate`) and carry `cuda_device` through the chunk loop, instead of re-querying per chunk. That is a small, behavior-preserving change; a further step could consult the registered-buffer metadata (the engine already knows the location string of registered memory) and skip the CUDA query entirely for buffers registered as `cpu:*`.

Happy to pick this up as a follow-up to my in-flight TCP work if nobody else wants it — the benchmark setup above reproduces it in a couple of minutes on any CUDA-build machine.

### Environment
- main @ `f1f5e40`; x86_64 Linux with a recent GCC/CUDA toolchain; loopback P2PHANDSHAKE metadata.

he-yufeng:
[...]

Issue #2086, opened 2026-05-12
[Bug]: CUDA 13 support clarification and potential data-integrity issue in transfer execution path under concurrent mixed-size descriptors


spencerr221:
## Summary
Requesting clarification on CUDA 13 package support and reporting a potential transfer data-integrity issue observed in the Mooncake execution layer under load.

## Package / Runtime Context
- Python: `3.12`
- OS: `Linux 5.10.134-16.103.al8.x86_64`
- vLLM: `0.20.0` (integration path used for reproducibility)
- Mooncake package in failing startup path: `mooncake-transfer-engine` (startup unavailable in this environment)
- Mooncake package in runtime test path: `mooncake-transfer-engine-cuda13==0.3.10.post2`

## Question 1: CUDA 13 support
Please confirm package support expectations:

1. Is `mooncake-transfer-engine-cuda13` the recommended package for CUDA 13 environments?
2. Is `mooncake-transfer-engine` expected to be unavailable or unsupported on CUDA 13?

Observed behavior:
- With `mooncake-transfer-engine`: startup path is unavailable in this environment.
- After switching to `mooncake-transfer-engine-cuda13==0.3.10.post2`: service startup succeeds.

## Question 2: Potential transfer execution-layer integrity issue
After switching to `mooncake-transfer-engine-cuda13==0.3.10.post2`, startup is successful, but under concurrent PD workload, descriptor-level hash verification indicates non-deterministic destination-side inconsistency.

### Technical observation boundary
In the integration path, descriptor planning is completed before execution:
- Concurrent multi-request descriptor planning: `_build_transfer_params(...)`
- Mixed-size transfer list generation (e.g. `2424832 + 16384`): generated in `lengths` by block/group/coalesce logic in the same planning function

After planning is complete, execution is delegated to Mooncake transfer engine:
- `batch_transfer_sync_write(remote_session, src_ptrs, dst_ptrs, lengths)`

The observed mismatch appears after this handoff boundary.

### Hash-based evidence pattern
For the same `(transfer_id, descriptor_idx)`:
- `src_hash_pre == src_hash_post` (sender-side source remains stable)
- `dst_hash != src_hash_post` (receiver-side destination differs)
- occasionally `dst_hash_round1 != dst_hash_round2` (receiver-side instability across two reads)

Mismatch concentration:
- Predominantly on large descriptors (frequently `length=2424832`)
- Small descriptors (e.g. `16384`) are frequently consistent

[...]

PR #2850, opened 2026-07-10
[TransferEngine] Acknowledged TCP framing: COMPLETED means applied at destination

n-WN:
## Description

Fixes the TCP data-integrity half of #2086 (the CUDA-13 wheel half was answered in the thread).

`TcpTransport` reported a WRITE as `COMPLETED` when its final chunk entered the *initiator's* kernel socket buffer. There is no frame in the v1 protocol by which the receiver confirms — or rejects — anything, so:

- destination memory can still be mutating megabytes after `COMPLETED` — measured **166/400 iterations torn** with the reported ~2.4 MB descriptor size on loopback, worst case the *entire* descriptor undelivered ([repro details](https://github.com/kvcache-ai/Mooncake/issues/2086#issuecomment-4892802494)). Any consumer notified out-of-band (the RDMA-WRITE semantics this same API provides on the rdma path, which disaggregated serving relies on) reads a torn or stale region — matching the reporter's `dst_hash` differing between consecutive reads, only on large descriptors, with TcpTransport silently serving a `protocol=rdma` config;
- a server-side rejection is invisible: a single-chunk WRITE to an unregistered address reported `COMPLETED` — silent data loss;
- errored connections were re-pooled on `is_open()` alone, so a protocol-desynced socket could corrupt subsequent requests;
- `session_mutex_` was locked and unlocked on different threads (UB per [thread.mutex.requirements]) while synchronizing nothing.

## Design: negotiated, wire-compatible acknowledged framing (v2)

- **Negotiation**: servers advertise `tcp_proto_version: 2` in the segment descriptor (old readers ignore the unknown JSON field; decode tolerates absence, defaulting 1). Initiators set the opcode high bit (`0x80`) only against v2-capable descriptors — all four mixed-version pairings behave: old↔old and old-initiator→new-server are byte-identical to today; new-initiator→old-server can only occur through a stale descriptor and fails fast (below).
- **WRITE**: the server sends an 8-byte status frame only after the final chunk has been applied to destination memory; the initiator completes on that frame. The ack read is armed *concurrently* with the body stream, so a rejection (or a legacy peer) aborts a large in-flight WRITE instead of deadlocking on mutually full socket buffers. If that frame is negative or malformed while `asio::async_write` still owns caller memory, the socket closes immediately but `FAILED` is not published until the body handler has quiesced — terminal status remains the source-buffer lifetime boundary.
- **READ**: the server prefixes the payload with a status frame (sent back-to-back — no added RTT), so rejections are signaled rather than inferred from a dropped connection.
- **Magic-tagged frames**: status frames carry `"MCTV"` in the high 32 bits; an initiator that reaches a v1 server through a stale descriptor (v1 streams payload for unknown opcodes) fails on the first frame instead of mis-framing the stream. The 2⁻⁶⁴ data-dependent residual is documented at the definition.
- **Pool hygiene**: `on_complete_` now reports whether the exchange terminated cleanly; only clean exchanges re-pool the socket, everything else is closed and dropped (fixes the desync-reuse hazard for v1 traffic too).
- **`session_mutex_` removed**: sessions are kept alive by `shared_from_this` in every handler; the mutex had no second owner anywhere and its cross-thread unlock was UB.
- **`MC_TCP_PROTO=1`** forces legacy initiator framing (documented operational escape hatch; also lets the tests cover the mixed-version matrix in one process). Zero-length transfers short-circuit to `COMPLETED` client-side, preserving their (accidental) v1 outcome.

## Module

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

Multi-core x86_64 Linux test host with a recent GCC/CUDA toolchain; loopback data path (no NIC/GPU required by the tests).

**Test commands:**
```bash
cmake .. -DBUILD_UNIT_TESTS=ON   # USE_TCP default
make tcp_write_visibility_test && ./mooncake-transfer-engine/tests/tcp_write_visibility_test
```

**Results (new `tcp_write_visibility_test`, connection pool enabled):**

| test | main (v1) | this PR |
|---|---|---|
| `CompletedWriteIsVisibleToSubsequentRead` (400 × 2.4 MB generation-stamped writes + 3 noise writers, destination inspected right after COMPLETED) | **fails: 166/400 torn**, worst case 2,432 KB stale | passes, 0/400 |
| `RejectedWriteMustFail` (unregistered address; pool must not be poisoned after) | fails (reports COMPLETED) | passes |
| `LargeRejectedWriteQuiescesSourceBeforeFailure` (32 MB body, immediate valid negative ack, source `mprotect(PROT_NONE)` at FAILED) | hangs/unsafe terminal boundary | passes |
| `StaleV2DescriptorAgainstLegacyServerQuiescesWriteBeforeFailure` (real POSIX v1 fake server, 32 MB full-duplex deadlock cell, malformed ack, source protection) | hangs once both socket directions fill | passes |
| `V2ReadRoundTripAndRejectedRead` (content integrity; rejected READ = FAILED; pool usable after) | n/a | passes |
| `LegacyInitiatorInteropWithV2Server` (`MC_TCP_PROTO=1`) | n/a | passes |

Throughput A/B, `transfer_engine_bench` TCP loopback (equal build flags, write, 4 threads × batch 32, two runs each):

| block | main GB/s | this PR GB/s |
|---|---|---|
| 16 KB | 0.25 / 0.29 | 0.24 / 0.30 |
| 64 KB | 0.88 / 0.84 | 0.88 / 0.97 |
| 256 KB | 1.28 / 1.15 | 1.34 / 1.30 |
| 1 MB | 1.54 / 1.53 | 1.62 / 1.46 |

Flat within noise — the per-request ack bubble is hidden by connection-level parallelism. (While measuring this we found an unrelated pre-existing ~2× TCP throughput loss on `USE_CUDA` builds — filed as #2843.)

Additional final-byte validation on a dedicated Linux test host: normal suite **6/6 × 5**; the two large early-abort/quiescence cells **20/20 each**; CUDA-enabled ASAN/UBSAN/LSAN **6/6 clean**. A CPU TSan build also passes the two new quiescence cells after suppressing three pre-existing engine-wide races in the custom `RWSpinlock`, polling counters, and slice-cache reuse (raw full-suite TSan reports those unrelated existing races).

- [x] Unit tests pass
- [x] Manual testing done (matrix above)

## Checklist

- [x] Focused pre-commit hooks pass on every touched non-CMake file; `cmake-format` leaves the added target block unchanged but proposes unrelated reflows elsewhere in the upstream test CMake file, which are intentionally excluded per `AGENTS.md`
- [x] I have updated the documentation (if applicable) — `MC_TCP_PROTO` added to the transfer-engine env-var list
- [x] For changes >500 LOC: N/A — the non-test diff is 459 changed lines (329 additions, 130 deletions), below the governance threshold; #2086 contains the design/root-cause discussion

## AI Assistance Disclosure

Claude (Anthropic) was used for implementation, test drafting, and adversarial review under my direction (those passes surfaced the mixed-version large-WRITE deadlock cell, missing pool-path coverage, and the requirement to delay terminal failure until an outstanding body write releases caller memory). I reviewed every changed line; all numbers are from real runs on my hardware, and the failure numbers were measured against unmodified main with the same test binary.

---

cc @stmatengss (assignee of #2086) — the root-cause analysis and reproducer data are in the issue thread; happy to adjust protocol details (e.g. frame layout, or gating v2 behind a config) if you'd prefer a different rollout shape.

### DIFF
--- a/docs/source/design/transfer-engine/index.md
+++ b/docs/source/design/transfer-engine/index.md
@@ -497,6 +497,7 @@ For advanced users, TransferEngine provides the following advanced runtime optio
 - `MC_ENDPOINT_STORE_TYPE` Choose FIFO Endpoint Store (`FIFO`) or Sieve Endpoint Store (`SIEVE`), default is `SIEVE`.
 - `MC_TCP_ENABLE_CONNECTION_POOL` Enable TCP Connection Pool to avoid excessive sockets.
 - `MC_TCP_SLICE_SIZE` The segmentation granularity (in bytes) of TCP transport for splitting large transfers into socket read/write operations. Corresponds to `MC_SLICE_SIZE` for RDMA. Default value 65536 (64KB).
+- `MC_TCP_PROTO` When set to `1`, TCP initiators use the legacy unacknowledged framing even against servers that support acknowledged framing (protocol v2). Under v2 (the default against v2-capable servers), a WRITE completes only after the receiver confirms the payload has been applied to destination memory, and server-side rejections surface as failed transfers instead of silent data loss. Use this variable only as a rollback escape hatch during mixed-version upgrades.
 
 ## C++ API Reference
 
--- a/mooncake-transfer-engine/include/transfer_metadata.h
+++ b/mooncake-transfer-engine/include/transfer_metadata.h
@@ -111,6 +111,10 @@ class TransferMetadata {
         RankInfoDesc rank_info;
 
         int tcp_data_port;
+        // TCP data-plane protocol version advertised by this segment's
+        // server. v2 adds acknowledged WRITE framing and status-prefixed
+        // READ responses (#2086); absent/1 = legacy unacknowledged framing.
+        int tcp_proto_version{1};
 
         // In dual-NIC setups (MC_RDMA_BIND_ADDRESS), the RDMA-reachable
         // address may differ from the TCP-routable segment name.  When
--- a/mooncake-transfer-engine/include/transport/tcp_transport/tcp_transport.h
+++ b/mooncake-transfer-engine/include/transport/tcp_transport/tcp_transport.h
@@ -142,6 +142,10 @@ class TcpTransport : public Transport {
         const std::string &host, uint16_t port);
     void returnConnection(const std::string &host, uint16_t port,
                           std::shared_ptr<asio::ip::tcp::socket> socket);
+    // Close `socket` and drop it from the pool: used for requests that did
+    // not terminate in a well-defined protocol state (#2086).
+    void discardConnection(const std::string &host, uint16_t port,
+                           std::shared_ptr<asio::ip::tcp::socket> socket);
     void cleanupIdleConnections();
 
     static constexpr std::chrono::seconds kConnectionIdleTimeout{60};
--- a/mooncake-transfer-engine/src/transfer_metadata.cpp
+++ b/mooncake-transfer-engine/src/transfer_metadata.cpp
@@ -278,6 +278,7 @@ static int encodeMultiProtocolSegmentDesc(
     segmentJSON["buffers"] = buffersJSON;
     segmentJSON["protocol"] = protocolJSON;
     segmentJSON["tcp_data_port"] = desc.tcp_data_port;
+    segmentJSON["tcp_proto_version"] = desc.tcp_proto_version;
     segmentJSON["timestamp"] = getCurrentDateTime();
 
     return 0;
@@ -319,6 +320,7 @@ int TransferMetadata::encodeSegmentDesc(const SegmentDesc &desc,
     segmentJSON["name"] = desc.name;
     segmentJSON["protocol"] = desc.protocol;
     segmentJSON["tcp_data_port"] = desc.tcp_data_port;
+    segmentJSON["tcp_proto_version"] = desc.tcp_proto_version;
     segmentJSON["timestamp"] = getCurrentDateTime();
     if (!desc.rdma_server_name.empty()) {
         segmentJSON["rdma_server_name"] = desc.rdma_server_name;
@@ -515,6 +517,9 @@ decodeMultiProtocolSegmentDesc(Json::Value &segmentJSON,
     auto desc = std::make_shared<TransferMetadata::SegmentDesc>();
     desc->name = segmentJSON["name"].asString();
     desc->tcp_data_port = segmentJSON["tcp_data_port"].asInt();
+    desc->tcp_proto_version = segmentJSON.isMember("tcp_proto_version")
+                                  ? segmentJSON["tcp_proto_version"].asInt()
+                                  : 1;
     if (segmentJSON.isMember("timestamp"))
         desc->timestamp = segmentJSON["timestamp"].asString();
     if (segmentJSON.isMember("rdma_server_name"))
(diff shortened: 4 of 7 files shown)


n-WN:
Follow-up review/testing is force-pushed in signed commit `5172f13d`. Besides fixing the typos CI failure, this closes the caller-buffer lifetime hole in early WRITE aborts: a negative/malformed ack now closes immediately but does not publish `FAILED` until the outstanding `asio::async_write` handler has quiesced. Two 32 MB regressions cover both a real v2 rejection and stale-v2-descriptor → v1-server full-duplex deadlock, then `mprotect(PROT_NONE)` the source at terminal status. Fresh results on a dedicated Linux test host: normal 6/6 × 5, both abort cells 20/20, CUDA-enabled ASAN/UBSAN/LSAN 6/6, plus focused CPU TSan after suppressing three documented pre-existing engine-wide races. CI has restarted.

n-WN:
The failed `build (3.12)` run was test-fixture-specific, not a transport failure: CI exports `MC_METADATA_SERVER=http://127.0.0.1:8080/metadata`, but the new fixture always reopened its local segment via `getLocalIpAndPort()` (the P2P-handshake name). Under the HTTP backend the descriptor is registered under the requested `server_name`, so all six cases stopped during setup with metadata 404s. The stale-v2 fake peer also listened only on loopback while HTTP RPC metadata advertises the engine-selected LAN address.

Signed follow-up `bf32f831` now selects the registered name by backend and makes the fake legacy peer reachable through either local address. Final-byte validation on a dedicated Linux test host against an **unmodified `HEAD` HTTP metadata server**: normal 6/6 × 3, CUDA ASAN/UBSAN/LSAN 6/6, focused CPU TSan 2/2; P2PHANDSHAKE remains normal 6/6, ASAN 6/6, TSan 2/2. The exact CI-style `ctest -R '^tcp_write_visibility_test$'` passes for both HTTP metadata and P2PHANDSHAKE. New CI run: https://github.com/kvcache-ai/Mooncake/actions/runs/29116164370

alogfans (approved):
LGTM. 

Note: The correctness of the entire concurrent write path relies implicitly on the `io_context` being **single-threaded**. Under that constraint the logic is sound. However, neither the code nor the comments document this assumption. If the executor is ever changed to a `thread_pool` (or the `strand` discipline is not preserved) later, it may be problematic. To harden the two flags:
- Change `write_body_in_flight_` to `std::atomic<bool>`.
- Set `write_body_in_flight_ = true` **before** calling `readWriteAck()` in `writeHeader`, so the flag is always set before the ack read can complete.

n-WN:
Pushed `76696caa` after an adversarial re-review pass over the branch. It closes a hole in the stale-descriptor story: fail-fast relied on the legacy peer's bytes not parsing as a status frame, but a request **shorter than a frame (1–7 bytes)** never yields 8 bytes at all — the v1 server streams `size` bytes of "READ payload" and then keeps the connection open awaiting the next header (for WRITE it is symmetrically stuck parsing our body as a partial header), and TCP slices have no timeout worker, so both sides waited forever. The status-frame reads now carry a deadline that covers only the frame, never payload streaming (READ: frame precedes payload; WRITE: armed only once the body is done). Expiry closes the socket and the pending read's existing handler owns the failure, including source-buffer quiescence. Default 30s, `MC_TCP_STATUS_TIMEOUT_SEC` overrides. New test drives both directions of a 4-byte request against a persistent fake v1 peer and asserts the failure arrives *after* the deadline (an early failure would mean the wrong path failed). Full suite passes 3× in both P2PHANDSHAKE and HTTP-metadata modes.

The same pass also sharpened one **known limitation** worth stating precisely: the magic-tagged frame residual is not merely probabilistic. Through a stale v2 descriptor, a v1 server echoes destination memory, and if the *application data* at the destination happens to begin with a valid OK frame, a short WRITE can report COMPLETED without having been applied. In-band framing cannot fully close this: a nonce would have to ride in the request header, and the header must stay 24 bytes or v1 servers mis-parse it. The complete fix is a connection-level hello (one extra RTT on first dial), which I'd propose as a follow-up rather than growing this PR; the window requires a server downgrade + stale cache + attacker-shaped destination bytes, and expiry-bounded now rather than unbounded.

Issue #2132, opened 2026-05-18
[RFC]: TENT Local Transfer Admission Queue


zbchi:
## Summary

This RFC proposes a local transfer admission queue inside `TransferEngineImpl`. The queue orders logical transfers and applies backpressure before work is handed to `Transport::submitTransferTasks()` or `ProxyManager::submit()`.

This covers the local-runtime part of the Transfer Queue roadmap item in #1883, before adding distributed pressure signals or receiver credits.

It does not replace transport or ProxyManager execution. RDMA slice scheduling, CQ/QP handling, staging state machines, retries, and KVCache placement remain owned by their existing layers.

Phase 1 keeps the normal submit path direct and adds a private queue model plus compatibility tests. Distributed pressure and receiver credits can be added later as inputs to local admission; this RFC does not propose a cluster-wide per-transfer executor.

## Motivation

The current path is optimized for direct submission:

```text
submitTransfer()
  -> merge requests
  -> resolve transport or staging
  -> allocate SubBatch
  -> assign sub_task_id
  -> Transport::submitTransferTasks() or ProxyManager::submit()
```

This is fast when the process is not overloaded. Under heavy load, background prefetch or migration traffic and foreground Store `get` traffic can enter transport queues through the same direct path. By the time foreground work is visible to the runtime, transport-side queues may already contain background work.

Transport execution and ProxyManager staging remain owned by their existing layers. This RFC adds an earlier runtime admission point in `TransferEngineImpl::submitTransfer()`, so accepted work can be capped and latency-sensitive requests can be ordered before handoff.

## Current Code Shape

The queue would be inserted into the current `submitTransfer()` path:

```text
Current direct-submit shape

Application / Store
  |
  | allocateBatch(batch_size)
  v
TransferEngineImpl::submitTransfer(batch_id, requests)
  |
  | merge requests
  | append public TaskInfo entries to Batch::task_list
  | allocate SubBatch state and assign sub_task_id
  |
  +-- direct path --> Transport::submitTransferTasks()
  |                  -> RDMA / TCP / SHM / NVLink / GDS execution queues
  |
  +-- staged path --> ProxyManager::submit()
                     -> shard queue with TaskInfo*
                     -> stage buffers / internal batches / engine callbacks

Status and lifetime side paths

getTransferStatus()
  -> expects valid transport state for non-staging tasks
  -> treats UNSPEC as retry/fallback, not as "queued"

freeBatch()
[...]

PR #2851, opened 2026-07-10
[TENT] Add best-effort RDMA task cancellation

catyans:
## Summary

Add task-level, best-effort cancellation to TENT, with the first transport implementation for RDMA.

- expose cancellation through the C++, C, and Python APIs
- cancel owners that are still waiting in the local admission queue without dispatching them
- let RDMA workers suppress queued, grouped, and retrying slices before `ibv_post_send`
- drain already-posted WRs through the normal CQ path instead of reporting a false cancellation
- preserve partial `transferred_bytes`, prevent failover after cancellation, and make repeated cancellation idempotent
- release RDMA device quota exactly once when an unposted slice is canceled or abandoned

This is a lifecycle follow-up to #2132. It does not change staging cancellation or add cancellation support to non-RDMA transports.

## Semantics

`cancelTransfer(batch_id, task_id)` returning OK means the request was accepted, not that the task is already terminal. Callers continue polling before freeing the batch.

- admission-queued task: terminal `CANCELED`, never dispatched
- worker-owned but unposted RDMA slice: terminal `CANCELED`
- mixed posted/unposted RDMA task: posted WRs drain; final status is `CANCELED`, with completed bytes retained
- fully posted RDMA task: may still finish `COMPLETED`
- merged public tasks: canceling any alias cancels their shared physical task
- staging or direct non-RDMA task: `NotImplemented`

## Validation

- `admission_queue_test`, `tent_runtime_queue_dispatch_test`, and `rdma_cancel_test`: pass
- the three suites above repeated 100 times: pass
- targeted ASan + UBSan cancellation tests: pass
- real mlx5 RDMA integration on a server with eight active bonded RNICs:
  - existing 4 MiB WRITE then READ data-integrity path: pass
  - 1 worker / 1 QP / 1 outstanding WR, 16 tasks: tasks 0-14 completed and task 15 reached `CANCELED`
  - repeated in 10 independent containers: 10/10 pass
- pre-commit: whitespace, EOF, YAML, merge markers, large files, ruff, codespell, clang-format, and cmake-format pass
- clang-format 20 was run in the Linux validation environment and produced no diff

## Compatibility

Default behavior is unchanged unless a caller invokes the new cancellation API. The base `Transport` implementation reports `NotImplemented`, so transports opt in explicitly.

## AI assistance

This change was developed with AI assistance. The implementation, edge-case semantics, tests, sanitizer results, and real-RDMA behavior were reviewed and validated by the submitter.

### DIFF
--- a/docs/source/design/tent/cpp-api.md
+++ b/docs/source/design/tent/cpp-api.md
@@ -75,6 +75,7 @@ For users migrating from Transfer Engine, the following table shows how TE APIs 
 | `allocateBatchID(batch_size)` | `allocateBatch(batch_size)` | Renamed |
 | `freeBatchID(batch_id)` | `freeBatch(batch_id)` | Renamed |
 | `submitTransfer(batch_id, entries)` | `submitTransfer(batch_id, request_list)` | `TransferRequest` → `Request` |
+| *Not available* | `cancelTransfer(batch_id, task_id)` | TENT-only: best-effort cancellation for queued and RDMA tasks |
 | `submitTransferWithNotify(batch_id, entries, notify_msg)` | `submitTransfer(batch_id, request_list, notifi)` | Unified API with optional notification |
 | `getTransferStatus(batch_id, task_id, status)` | `getTransferStatus(batch_id, task_id, status)` | Same |
 | `getBatchTransferStatus(batch_id, status)` | `getTransferStatus(batch_id, status)` | Overloaded; single `TransferStatus` output = overall status |
@@ -273,6 +274,24 @@ Queries the status of transfer requests.
 - `task_id`: Index of the specific task (for single-task query).
 - `status` / `status_list` / `overall_status`: Output parameter(s) for status.
 - Return value: `Status::OK()` on success; otherwise a non-OK status.
+
+#### TransferEngine::cancelTransfer
+
+```cpp
+Status cancelTransfer(BatchID batch_id, size_t task_id);
+```
+
+Requests best-effort cancellation of one public task. A task still waiting in
+the TENT admission queue becomes `CANCELED` without being dispatched. For RDMA,
+workers suppress slices they observe before `ibv_post_send`; work already
+posted to a QP is allowed to drain and may complete successfully. Consequently,
+the API returning `OK` means the cancellation request was accepted, not that
+the task is already terminal. Continue polling `getTransferStatus` before
+calling `freeBatch`.
+
+Cancellation is idempotent. Merged public tasks share one physical transfer,
+so canceling any alias cancels the shared task. Direct cancellation of staging
+or non-RDMA transport work currently returns `Status::NotImplemented`.
 
 #### TransferEngine::freeBatch
 
--- a/mooncake-transfer-engine/tent/include/tent/runtime/admission_queue.h
+++ b/mooncake-transfer-engine/tent/include/tent/runtime/admission_queue.h
@@ -134,6 +134,11 @@ class LocalTransferAdmissionQueue {
 
     Status complete(QueueOwnerId owner_id, TransferStatusEnum terminal_status);
 
+    // Cancel an owner that has not entered the dispatch window. Idempotent for
+    // an owner already canceled; dispatching owners must be canceled through
+    // their selected transport instead.
+    Status cancel(QueueOwnerId owner_id);
+
     Status retireBatch(uint64_t batch_token);
 
     Status resolveOwner(uint64_t batch_token, size_t public_task_id,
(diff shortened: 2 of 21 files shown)


catyans:
Addressed the Gemini cancellation review issues in `a3755091`:

- changed `RdmaTask::success_slices` / `resolved_slices` from volatile ints to atomics and updated status accounting to use atomic fetch/load
- made `Workers::cancel()` best-effort safe after worker shutdown by checking `running_`, `worker_context_`, and `num_workers_`
- added a `slice->task` guard in `cancelUnpostedSlice()`
- replaced C API `CHECK_POINTER(batch_id)` checks with scalar `if (!batch_id) return -1` checks for cancel/status entry points

Validation:

- `clang-format` and `git diff --check` passed
- remote Linux container build with `USE_TENT=ON`, `BUILD_UNIT_TESTS=ON`, `USE_CUDA=OFF`
- built `rdma_cancel_test` and `tent_runtime_queue_dispatch_test`
- `ctest -R "rdma_cancel_test|tent_runtime_queue_dispatch_test" --output-on-failure`: 2/2 passed

catyans:
Merged the latest `upstream/main` in `159712b7` and resolved the only conflict in `tent/tests/CMakeLists.txt` by retaining both `rdma_cancel_test` and the newly added `thread_local_storage_test`.

Post-merge validation on the H20/RDMA cluster:
- clean configure/build with `USE_TENT=ON`, `BUILD_UNIT_TESTS=ON`, `USE_CUDA=OFF`
- focused ctest suite: 5/5 passed (`rdma_cancel_test`, `tent_rdma_transport_test`, `admission_queue_test`, `tent_runtime_queue_dispatch_test`, `thread_local_storage_test`)
- `rdma_cancel_test`: 100 shuffled repeats passed
- `tent_rdma_transport_test`: 20 shuffled repeats passed
- `git diff --check` passed

GitHub CI has been retriggered for the merge commit.

he-yufeng (approved):
Reviewed as a maintainer — traced the cancellation invariants through the worker / quota / slice lifecycle, not just the diff. The design holds and the claimed guarantees check out.

**Idempotency is correct.** `Workers::cancel` gates on `cancel_requested.exchange(true, acq_rel)` and returns `OK` if it was already set, so a repeated `cancelTransfer` on the same (or a merged-alias) task is a no-op. The follow-up wake loop only nudges suspended workers to re-check.

**Quota is released exactly once.** `releaseSliceQuota` is guarded by `slice->quota_charged` and clears it right after `device_selector_->release`, so a second call for the same slice is a no-op. Each slice reaches release through exactly one path — `cancelUnpostedSlice` for an unposted slice, or the CQ completion path for a posted one — never both. The `quota.cpp` change is right too: `DeviceSelector::release` calls `dev.releaseInflight(length)` *before* the `latency <= 0.0` early return, so the cancel path (no latency sample) still returns the inflight charge and only skips the EWMA bandwidth update.

**No `inflight_slices` double-decrement.** `cancelUnpostedSlice` only runs over the to-post lists (the `result` slice-lists and `worker.requests` / its clone); posted slices are moved into `worker.inflight_slice_set` and erased from the to-post list before the CQ loop touches them. So a slice is decremented once — by `cancelUnpostedSlice` if canceled before posting, or by the CQ path (`inflight_slice_set.erase` + `fetch_sub`) once posted. The two sets are disjoint.

**The `word == PENDING` guard in `cancelUnpostedSlice` is sound.** A to-post-list slice always has `word == PENDING`: `rdma_transport.cpp` sets it before the slice enters the worker queues, and any slice that reaches a terminal state via `updateSliceStatus`'s `CAS(&word, PENDING, ...)` leaves the list. So the guard is effectively always true there and `releaseSliceQuota` is always reached (no-op when `quota_charged` is false). If a `cancel()` races a concurrent completion, that CAS fails against the already-changed word, so a `COMPLETED` slice is never clobbered to `CANCELED` — matching the "fully posted may still finish COMPLETED / posted WRs drain" semantics.

One non-blocking note: `cancelUnpostedSlice`'s unconditional `worker.inflight_slices.fetch_sub(1)` + `return true` is correct *because* it's only ever called on to-post-list (unposted) slices, which are disjoint from `inflight_slice_set`. That invariant is what keeps the counter balanced — a future caller passing an already-posted slice would double-decrement. A one-line comment on the function pinning "callers must pass only unposted slices" would keep that from regressing.

The validation coverage (the three suites 100x, ASan/UBSan, and the real 8-RNIC mlx5 run where task 15 reached `CANCELED` while 0-14 completed) lines up with the semantics table. LGTM.

alogfans (approved):

Issue #2630, opened 2026-06-26
[RFC]: NCCL device API in Mooncake TENT


xiaofanl-nvidia:
## Proposal and motivation
Add a TENT NCCL transport backend built on NCCL's device API. 
This proposal is aligned with roadmap here: https://github.com/kvcache-ai/Mooncake/issues/1883 and RFC here: https://github.com/kvcache-ai/Mooncake/issues/2123 by offering a widely adopted solution for device-initiated communication, with easy-to-maintain abstractions and production quality validation. 

### Leverage NCCL's device API
NCCL's device API abstracts hardware topology and features into a clean, cohesive API suite. By adopting it, Mooncake gains access to capabilities such as NVLink SHARP (NVSwitch multicast), GPU-initiated networking (GIN) over multiple network backends (IB, RoCE, EFA, and more), CFT (Compute Fabric Transport), Counted Writes, CuteDSL/cuTiles integration, and others, all without the effort of rebuilding from scratch.

### Fine-grained comm/compute fusion
Supporting device-initiated communication allows per-layer, fine-grained data transfer instead of relying on the CPU to initiate network traffic. Beyond the usual benefits — fewer CPU threads, fewer kernel launches, and fewer staging buffers — it also makes comm/compute fusion possible.

### Build custom collectives
NCCL's own collectives are built on this same device-API surface, so integrating it gives Mooncake the same proven foundation for building high-performance custom collectives. Ready-to-use reference implementations live at https://github.com/NVIDIA/nccl/tree/master/contrib, with minimal examples at https://github.com/NVIDIA/nccl/tree/master/docs/examples/06_device_api.

### Leverage DeepEPv2 / nccl-ep
DeepEPv2 and nccl-ep make heavy use of NCCL's device API, so integrating it lets Mooncake gain the same benefits and build its own EP solutions on the same foundation.

### Leverage the NCCL ecosystem
Taking on NCCL as a dependency lets Mooncake easily use other capabilities from the ecosystem — high-performance host-side collectives, one-sided put/signal APIs, immediate support for new NVIDIA hardware, topology, and CUDA features, and other NCCL extensions such as nccl-ep, nccl-m2n, nccl-ubx, and more.

## Implementation details
Cross-node transfers use one communicator per peer session with multiple GIN contexts for striping. WRITE uses the device-side ncclGin::put with signal/ack, and READ uses the device-side ncclGin::get.
[...]

PR #2852, opened 2026-07-10
[TransferEngine] Add NCCL host and device transport backend

akhillanger:
## Summary

This PR adds two opt-in experimental NCCL backends to Mooncake Transfer Engine:

- A host-driven NCCL RMA transport implementing the existing `Transport` interface.
- A CUDA device transport exposing NCCL LSA and GIN operations to kernels.

## Host RMA transport

`installTransport("nccl")` installs a WRITE-only transport in a `TransferEngine(false)` instance. The transport uses the existing P2P metadata handshake to exchange an opaque NCCL bootstrap payload and registered CUDA-buffer metadata.

Each peer/device session creates a two-rank NCCL communicator, collectively registers corresponding buffers as symmetric NCCL windows, and submits transfers with `ncclPutSignal`. CUDA events report asynchronous completion through the normal Transfer Engine status API.

The host backend follows the classic Transfer Engine single-transport model:

- NCCL must be installed before local buffers are registered.
- NCCL must be the only transport in that engine instance; there is no routing or fallback to another host transport.
- Both peers must register the same number of buffers in the same order with matching sizes.
- Both endpoints must remain available while the first transfer initializes the session and collective windows.
- READ requests are rejected with `Status::NotSupportedTransport`; NCCL 2.30 has no public host-side Get operation.

## Device transport

`getOrCreateNcclTransport()` provides host-side setup for an NCCL device communicator. Applications exchange the generated NCCL unique ID through their existing control plane, initialize every rank collectively, and allocate/register symmetric buffers.

A compact `NcclDeviceContext` is passed by value to CUDA kernels. The device helpers provide:

- Local, LSA, and GIN route discovery.
- LSA peer-pointer access, multimem access, and LSA barriers.
- GIN put, put-with-signal, signal-add, flush, signal-read, and signal-wait operations.
- A lane-selected put call shape compatible with the existing IBGDA device API, allowing consumer-side template selection between backends.

NCCL-specific declarations live in `transport/device/nccl_device_transport.h`; the generic device transport API is unchanged.

## Build

Both backends require CUDA and NCCL 2.30.4 or newer and are disabled by default.

```bash
cmake -S . -B build \
  -DUSE_CUDA=ON \
  -DUSE_NCCL_HOST=ON \
  -DUSE_NCCL_DEVICE=ON

cmake --build build -j
```

The options can be enabled independently:

- `USE_NCCL_HOST` requires the public `nccl.h` host API.
- `USE_NCCL_DEVICE` additionally requires the experimental `nccl_device.h` API.
- `NCCL_ROOT` selects a nonstandard NCCL installation.

Optional manual examples are enabled with:

```bash
-DBUILD_NCCL_HOST_TRANSPORT_EXAMPLE=ON
-DBUILD_NCCL_DEVICE_TRANSPORT_EXAMPLE=ON
```

## Device constraints

- Device initialization and buffer registration are collective and must occur in the same order on every rank.
- Host calls on one device transport require external serialization.
- `initialize()` binds the transport to the current CUDA device; that device must remain current for later host calls, shutdown, and destruction.
- Device code currently supports full-world GIN connectivity; rail-team connectivity is not included.
- NCCL Device API headers used for Mooncake, AOT kernels, or JIT kernels must exactly match the runtime `libnccl`. Device code must be rebuilt or re-JITed after an NCCL upgrade.

## Examples and tests

- `nccl_host_transport_example` uses two GPUs to validate concurrent and bidirectional WRITE transfers and explicit READ rejection.
- `nccl_device_transport_example` uses one process per rank to validate LSA or GIN put/signal completion from a device kernel.
- `transfer_metadata_test` covers ordered NCCL GPU-buffer metadata and opaque handshake-payload round trips.
- `device_backend_api_compatibility_test` instantiates one templated put operation for NCCL and IBGDA contexts.

## Validation

- Host transport, shared Transfer Engine changes, host example, and device host-side implementation compile in syntax-only checks.
- Metadata and routing sources compile with normal and `ENABLE_MULTI_PROTOCOL` configurations.
- Focused CMake probes pass for host-only/device-enabled dependency gating and NCCL versions with a zero patch component.
- CUDA kernel compilation and GPU runtime tests were not run on the current x86-64 runner because its available CUDA/NCCL toolchain and dependencies are AArch64.


mooncake-transfer-engine/src/transport/nccl_transport/nccl_transport.cpp
```diff
@@ -0,0 +1,1181 @@
...
+                    return buffer.addr == reinterpret_cast<uint64_t>(addr);
+                });
+            if (it != local_buffers_.end()) {
+                local_buffers_.erase(it);
+                found = true;
+            }
+        }
+        if (!found) return ERR_ADDRESS_NOT_REGISTERED;
+        return metadata_->removeLocalMemoryBuffer(addr, update_metadata);
+    }
```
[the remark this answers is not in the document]
akhillanger: Addressed in 9f3cdfba. `unregisterMemoryLocked` now performs the protocol-scoped metadata removal before erasing the entry from `local_buffers_`, while the caller holds `buffers_mutex_`. The metadata side stages a cloned descriptor, publishes it first when requested, and commits the local metadata cache only after publication succeeds, so a lookup or publication failure leaves both catalogs unchanged. Batch unregister also prevalidates every address and uses one bulk transaction, making it all-or-nothing. I added coverage for failed atomic bulk removal, successful bulk removal, and same-address protocol-scoped removal.

_Disclosure: This response was AI-generated._

mooncake-transfer-engine/src/transport/nccl_transport/nccl_transport.cpp
```diff
@@ -0,0 +1,1181 @@
...
+        }
+
+        const auto& destination = window->buffers[owner_rank];
+        ncclResult_t result = ncclPutSignal(
+            source, length, ncclUint8, owner_rank, window->handle,
+            dest_addr - destination.addr, 0, 0, 0, comm_, stream_);
+        if (result == ncclSuccess && event) {
+            cuda_result = cudaEventRecord(event, stream_);
+        }
+        if (saved_device >= 0) cudaSetDevice(saved_device);
```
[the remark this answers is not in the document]
akhillanger: Addressed in 9f3cdfba. After a successful `ncclPutSignal`, a `cudaEventRecord` failure now synchronizes `stream_` before returning the error. This drains the already-queued put before the caller can reuse or free its source buffer. The original event-record error is preserved; if the fallback synchronization also fails, that error is appended and the session is marked failed. The normal WRITE and reverse-PUT READ paths passed the two-GPU and forced-network GB200 tests.

_Disclosure: This response was AI-generated._

mooncake-transfer-engine/src/transport/nccl_transport/nccl_transport.cpp
```diff
@@ -0,0 +1,1181 @@
...
+        local_desc.protocol = kHandshakeProtocol;
+        Json::Value request;
+        std::string error;
+        if (!decodeJson(peer_desc.payload, &request, &error)) {
+            local_desc.reply_msg = "Invalid NCCL handshake payload: " + error;
+            return 0;
+        }
+
+        const std::string op = request["op"].asString();
+        Json::Value response;
```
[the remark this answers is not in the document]
akhillanger: Addressed in 9f3cdfba. `decodeJson` now uses strict parsing and rejects any non-object root. `onHandshake` then requires `op` to exist and be a string before calling `asString()`, and the dispatch is protected by a `Json::Exception` boundary so malformed peer input is returned as a handshake rejection rather than escaping the daemon callback. The outer transfer-handshake decoder now also rejects wrong-typed present fields.

_Disclosure: This response was AI-generated._

mooncake-transfer-engine/src/transport/nccl_transport/nccl_transport.cpp
```diff
@@ -0,0 +1,1181 @@
...
+        buffer.addr = item["addr"].asUInt64();
+        buffer.length = item["length"].asUInt64();
+        buffer.device_id = item["device_id"].asInt();
+        if (!buffer.addr || !buffer.length ||
+            buffer.device_id != expected_device) {
+            if (error) *error = "Invalid NCCL buffer catalog entry";
+            return false;
+        }
+        buffers->push_back(buffer);
+    }
```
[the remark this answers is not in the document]
akhillanger: Addressed in 9f3cdfba. `decodeBuffers` now validates that the catalog is an array and that every entry is an object with `UInt64` address/length fields and an integer device ID before conversion. It also rejects zero values, address-range overflow, the wrong device, empty catalogs, and duplicate or overlapping ranges. Decoding uses a temporary vector and only replaces the output after the entire catalog validates, so a malformed later entry cannot leave a partial result.

_Disclosure: This response was AI-generated._

xiaofanl-nvidia:
++ @UNIDY2002 @staryxchen from the RFC https://github.com/kvcache-ai/Mooncake/issues/2630

mooncake-common/common.cmake
```diff
@@ -71,6 +71,8 @@ option(BUILD_EXAMPLES "Build examples" ON)
 option(BUILD_UNIT_TESTS "Build unit tests" ON)
 option(BUILD_BENCHMARK "Build benchmarks" ON)
 option(USE_CUDA "option for enabling gpu features for NVIDIA GPU" OFF)
+option(USE_NCCL_DEVICE "option for enabling the NCCL DeviceTransport backend" OFF)
+option(USE_NCCL_HOST "option for enabling the NCCL host RMA transport" OFF)
```
UNIDY2002, line 75: We have two options here, but I only see `USE_NCCL_DEVICE` in the `build.md` doc.

mooncake-transfer-engine/tests/device_backend_api_compatibility_test.cu
UNIDY2002, line 1: This seems to be a compile-time probe. I'm wondering whether it's necessary, since the regular build process should already reveal whether these features are supported.

mooncake-transfer-engine/tests/transfer_metadata_test.cpp
```diff
@@ -107,6 +107,81 @@ TEST_F(TransferMetadataTest, LocalMemoryBufferTest) {
     ASSERT_EQ(re, 0);
 }
 
+TEST_F(TransferMetadataTest, BulkMemoryRemovalIsAtomic) {
```
UNIDY2002: Why do we need this case? It seems irrelevant to NCCL transport.

mooncake-transfer-engine/src/transfer_metadata.cpp
```diff
@@ -96,6 +98,48 @@ struct TransferHandshakeUtil {
...
+        }
+#endif
+#ifdef USE_UB
+        if (root.isMember("jetty_num")) {
+            if (!root["jetty_num"].isArray()) return ERR_INVALID_ARGUMENT;
+            for (const auto &jetty : root["jetty_num"]) {
+                if (!jetty.isUInt()) return ERR_INVALID_ARGUMENT;
+            }
+        }
+#endif
```
UNIDY2002: I'm afraid these checks are too strict, because sometimes we may launch TE of different versions.

Setting fallback values when the new fields (i.e. `protocol` and `payload`) do not exist should be more acceptable, just like the handling of `local_lid` and `local_gid` below.

mooncake-transfer-engine/src/transfer_metadata.cpp
```diff
@@ -487,6 +548,7 @@ int TransferMetadata::updateSegmentDesc(const std::string &segment_name,
 }
 
 int TransferMetadata::removeSegmentDesc(const std::string &segment_name) {
+    std::lock_guard<std::mutex> txn_guard(local_segment_txn_mutex_);
```
UNIDY2002: Will introducing a `mutex` here have side effects? I'm not sure. cc: @alogfans @staryxchen

mooncake-transfer-engine/src/transfer_metadata.cpp
```diff
@@ -1116,46 +1213,133 @@ int TransferMetadata::removeLocalSegment(const std::string &segment_name) {
...
+        updated = std::make_shared<SegmentDesc>(*it->second);
+        updated->buffers.push_back(buffer_desc);
+        it->second = updated;
+    }
+    if (update_metadata) {
+        return updateSegmentDesc(updated->name, *updated);
     }
-    if (update_metadata) return updateLocalSegmentDesc();
     return 0;
 }
```
UNIDY2002, line 1303: Why do we have to change the impl of `addLocalMemoryBuffer`, `removeLocalMemoryBuffer`, `startHandshakeDaemon`, etc? They seems irrelevant to this PR.

mooncake-transfer-engine/src/transfer_engine_impl.cpp
```diff
@@ -138,10 +138,19 @@ int TransferEngineImpl::init(const std::string& metadata_conn_string,
...
+            // requested and cannot be shared by multiple local processes.
+            // Explicit non-zero ports are honored so independently launched
+            // peers can advertise predictable endpoints.
+            if (!hasExplicitPort(local_server_name) || port == 0) {
+                desc.rpc_port = findAvailableTcpPort(desc.sockfd);
+                if (desc.rpc_port == 0) {
+                    LOG(ERROR)
+                        << "P2P: No valid port found for local TCP service.";
+                    return -1;
+                }
```
UNIDY2002: Would you help me review these lines of codes? @alogfans

UNIDY2002 (commented):
Thank you for your contribution! I have a few comments that should be addressed before we move on.

mooncake-transfer-engine/src/transfer_metadata.cpp
alogfans: `removeSegmentDesc` is less frequently used so I think it doesn't have side effects.

mooncake-transfer-engine/src/transfer_engine_impl.cpp
alogfans: @UNIDY2002 This is okay.

mooncake-transfer-engine/src/transfer_metadata.cpp
```diff
@@ -1261,14 +1449,55 @@ int TransferMetadata::getRpcMetaEntry(const std::string &server_name,
 
 int TransferMetadata::startHandshakeDaemon(
```
alogfans, line 1436: Does startHandshakeDaemon call only once in normal execution? If not, how to update on_receive_handshake correctly? If necessary, it's better to seperate this feature to updateHandshakeHandler().

mooncake-transfer-engine/src/transfer_metadata.cpp
```diff
@@ -1180,6 +1364,10 @@ int TransferMetadata::addRpcMetaEntry(const std::string &server_name,
         if (rc != 0) {
             return rc;
         }
+        {
+            std::lock_guard<std::mutex> lock(handshake_handler_mutex_);
+            handshake_daemon_started_ = true;
```
alogfans: Why need this here? I think it's better to keep it in startHandshakeDaemon () only.

staryxchen (commented):
Thanks for contribution! 

I have a question: The PR description says "A TE read maps to ncclGet", but the actual READ implementation uses requestReversePut — a synchronous handshake to the peer, which then performs the reverse PUT via handleReversePut → putAndWait. But there's no ncclGet call in the code. 

More importantly, this puts a synchronous RPC on the data path. In the RFC #2630 discussion, a cross-node RPC on the data path was called out as one of the core issues with the POC, which motivated the control-plane/data-plane separation principle. Additionally, handleReversePut runs cudaEventSynchronize on the handshake daemon thread, blocking all other peers' handshakes on that thread.

mooncake-common/common.cmake
akhillanger: Addressed. `USE_NCCL_HOST` is now documented alongside `USE_NCCL_DEVICE` in both build-option tables, including its NCCL/CUDA requirements and WRITE-only behavior.

Disclosure: This response was generated with help of AI.

mooncake-transfer-engine/tests/device_backend_api_compatibility_test.cu
akhillanger: I kept this compile probe because the regular library target does not instantiate the header-only device helpers. The CUDA example has two separate opt-in gates—`BUILD_EXAMPLES=ON` and `BUILD_NCCL_DEVICE_TRANSPORT_EXAMPLE=ON`, with the latter disabled by default—so it is not compiled by a normal `USE_NCCL_DEVICE=ON` build. The probe provides CI coverage that instantiates the NCCL and IBGDA APIs from the same consumer template without requiring runtime GPU or NIC hardware. I added comments explaining this distinction.

Disclosure: This response was generated with help of AI.

mooncake-transfer-engine/tests/transfer_metadata_test.cpp
akhillanger: This test directly covers NCCL batch unregistration: one invalid address must leave the entire NCCL metadata catalog unchanged. I renamed it to `NcclBatchUnregisterMetadataRemovalIsAtomic`, changed the test metadata from `rdma` to `nccl`, and added a comment explaining the connection to the NCCL path.

Disclosure: This response was generated with help of AI.

mooncake-transfer-engine/src/transfer_metadata.cpp
akhillanger: Addressed. Missing `protocol` and `payload` fields now explicitly receive empty legacy defaults and are routed to the default handler. Fields that are present with an invalid type are still rejected. I also added a compatibility test that sends an old-schema request with both fields omitted.

Disclosure: This response was generated with help of AI.

mooncake-transfer-engine/src/transfer_metadata.cpp
akhillanger: I retained the mutex after reviewing its scope. It only serializes infrequent local metadata/catalog mutations and is not on the transfer data path. It prevents stale staged state from overwriting a concurrent catalog update or recreating a descriptor after removal. I added a comment documenting that requirement.

Disclosure: This response was generated with help of AI.

mooncake-transfer-engine/src/transfer_metadata.cpp
akhillanger: These generic changes are needed because NCCL shares the metadata catalog and handshake daemon with the existing transports. Protocol-scoped removal prevents NCCL unregistration from deleting another backend's entry at the same address; catalog writers participate in the transaction lock so staged batch operations cannot overwrite concurrent changes; and handshake dispatch allows NCCL to coexist with the existing default handler. I also separated handler replacement from daemon startup to make the lifecycle explicit.

Disclosure: This response was generated with help of AI.

mooncake-transfer-engine/src/transfer_metadata.cpp
akhillanger: Addressed as suggested. Handler replacement is now a separate `updateDefaultHandshakeHandler()` operation. Stable plugin callbacks are installed exactly once using `std::once_flag`, while the default handler can be updated after startup. The existing three-argument start method remains as a source-compatible wrapper, and a test verifies that a handler updated after listener startup receives subsequent requests.

Disclosure: This response was generated with help of AI.

mooncake-transfer-engine/src/transfer_metadata.cpp
akhillanger: Addressed. I removed `handshake_daemon_started_` and no longer modify daemon state from `addRpcMetaEntry`. That method now calls the common start path; callback installation is protected by `std::call_once`, and the handshake plugin remains responsible for its listener-running state.

Disclosure: This response was generated with help of AI.

akhillanger:
@staryxchen thanks for the review:

You are correct. There was a discrepancy in PR description. Also, I removed the reversed-PUT implementation. Currently NCCL does not have a host-side `ncclGet`. The host backend is therefore explicitly WRITE-only for now.

When a submitted task's opcode is READ, `submitTasks()` creates its status-tracking slice, rejects it locally, marks it failed, and returns `NotSupportedTransport` before that request performs metadata lookup, NCCL session lookup/creation, or peer bootstrap. Thus, a READ request cannot initiate or use an NCCL session, although a session created independently by an earlier or concurrent WRITE may already exist. Mixed-protocol routing can fall through to another READ-capable backend. The examples and documentation were updated accordingly, and the PR description has been corrected as well.

mooncake-transfer-engine/src/transfer_metadata.cpp
UNIDY2002: I believe these guards that `return ERR_INVALID_ARGUMENT` are in fact out of scope of this PR. If you do find them necessary, we should open another PR to handle them separately.

mooncake-transfer-engine/include/transfer_metadata.h
```diff
@@ -192,6 +196,11 @@ class TransferMetadata {
                              bool update_metadata);
 
     int removeLocalMemoryBuffer(void *addr, bool update_metadata);
+    int removeLocalMemoryBuffer(void *addr, const std::string &protocol,
+                                bool update_metadata);
+    int removeLocalMemoryBuffers(const std::vector<void *> &addr_list,
+                                 const std::string &protocol,
+                                 bool update_metadata);
```
UNIDY2002: This PR adds a protocol-scoped overload, but only NCCL uses it.

Can we either:
1. split the protocol-scoped removal fix into a separate TE metadata PR and migrate existing multi-protocol transports to it, with tests for rdma+hip/tcp+hip/etc.; or
2. explain why the existing transports are safe with existing address-only removal while NCCL is not?

mooncake-transfer-engine/src/transfer_metadata.cpp
```diff
@@ -1259,33 +1443,112 @@ int TransferMetadata::getRpcMetaEntry(const std::string &server_name,
...
+        handshake_plugin_->registerOnProbeCallBack(
+            [this](const Json::Value &peer, Json::Value &local) -> int {
+                return receivePeerProbe(peer, local);
+            });
+    });
+
+    return handshake_plugin_->startDaemon(listen_port, sockfd);
+}
+
 int TransferMetadata::startHandshakeDaemon(
```
UNIDY2002, line 1436: We do not plan to support multi-transport in TE, so multi-protocol handler dispatching is out of scope of the PR.

I understand your goal of letting NCCL transport co-exist with existing transports like RDMA, but I'm sorry that this modification may add too much complexity, so I would like to request switching back to Mooncake's original implementation.

TENT (the next generation of TE that we are concurrently rolling out in the same repo) supports multi-transport, so I believe your efforts can be applied to the TENT part in the following PRs.

UNIDY2002 (changes requested):
Thank you for your contribution.

My main concern is that this PR is not only adding an NCCL transport. It also introduces generic framework changes: protocol-scoped metadata removal, protocol-dispatched handshake handlers, default-handler replacement, and new local metadata transaction behavior.

Those changes seem motivated by a broader multi-transport coexistence model. That model is not part of the current TE host-transport contract, and I do not think we should expand TE in that direction in this PR. In the current patch, the protocol-scoped remove APIs are only used by NCCL, while existing transports still use the old address-only path, which makes the framework change look partial and NCCL-specific rather than a coherent transport-layer design.

Please keep this PR focused on the NCCL backend within the existing TE model. If the goal is to support richer multi-transport coexistence, protocol-dispatched handshakes, or per-transport registration/removal semantics, that work should be developed under the TENT framework instead. I would be happy to review and help align that direction there, but I do not think these framework changes should be introduced through this NCCL TE transport PR.

alogfans (changes requested):
The PR should be refactored to minimize the change of existing modules.

akhillanger:
Thank you for the comments @UNIDY2002 and @alogfans. I will work on addressing your concerns and get back.

UNIDY2002 (commented):
The code is much cleaner now.

LGTM on the device transport part.

@staryxchen @alogfans When you have a chance, could you please take a look at the host transport part as well? Thanks!

mooncake-transfer-engine/src/transport/nccl_transport/nccl_transport.cpp
```diff
@@ -0,0 +1,1263 @@
...
+                                             remote_buffer.device_id, &error);
+            }
+
+            if (error.empty()) {
+                int saved_device = -1;
+                cudaGetDevice(&saved_device);
+                cudaSetDevice(local_device);
+                cudaEvent_t event = nullptr;
+                cudaError_t cuda_result =
+                    cudaEventCreateWithFlags(&event, cudaEventDisableTiming);
```
staryxchen, line 727: Each WRITE slice allocates a `cudaEvent_t` into `slice->nccl.event`, and it is only destroyed inside `poll()` on success or error. But `TransferTask::~TransferTask()` (transport.h) deallocates slices via the cache without any transport-side cleanup, so a batch destroyed before every POSTED slice has been polled to completion leaks the live CUDA events.

NCCL is the first TE transport that owns a per-slice CUDA handle requiring explicit destruction, so the existing slice lifecycle does not cover this. Could you add a cleanup path — e.g. a transport hook invoked on slice deallocation, or track in-flight events in `Impl` and destroy them in `~Impl()`?

mooncake-transfer-engine/src/transport/nccl_transport/nccl_transport.cpp
```diff
@@ -0,0 +1,1263 @@
...
+                if (error) *error = ncclError(result, "ncclGetUniqueId");
+                return false;
+            }
+            *encoded_id = encodeUniqueId(*unique_id);
+        }
+        bootstrap_ids_.emplace(key, *encoded_id);
+        return true;
+    }
+
+    std::shared_ptr<NcclSession> getOrCreateSession(
```
staryxchen, line 995: Once a session enters `sessions_`, it is only removed in `~Impl`. If `initialize()` fails (`failed_ = true`), `waitReady` returns nullptr for every subsequent transfer to that peer, and the session cannot be recreated because `buffers_frozen_` is already latched — the only recovery is destroying the whole transport.

For an experimental backend this may be acceptable, but could you at least document it in "Device constraints", or evict failed sessions so a later transfer can retry bootstrap?

mooncake-transfer-engine/src/transport/nccl_transport/nccl_transport.cpp
```diff
@@ -0,0 +1,1263 @@
...
+        return -1;
+    }
+    if (attributes.type != cudaMemoryTypeDevice &&
+        attributes.type != cudaMemoryTypeManaged) {
+        return -1;
+    }
+    return attributes.device;
+}
+
+std::string makeSessionKey(const std::string& local_name, int local_device,
```
staryxchen, line 252: `makeSessionKey` orders by `local_name < peer_name`, and `local_rank` is assigned the same way (`local_server_name_ < peer_name ? 0 : 1`). When `local_name == peer_name` (same-node cross-device), both sides take the `else` branch with swapped arguments, producing different keys; and both compute `rank = 1`, so neither creates the unique ID — bootstrap deadlocks.

Same-node transfers should use NVLink, so this is likely out of scope, but the latent bug is there. If you want it correct, compare the `(name, device)` tuple instead of name alone:

```cpp
if (std::tie(local_name, local_device) < std::tie(peer_name, peer_device)) {
    ...
}
```

UNIDY2002:
@akhillanger Hi! Would you help resolve the conflicts with the main branch, so that the CI could be triggered? Thanks.

mooncake-transfer-engine/src/transport/nccl_transport/nccl_transport.cpp
akhillanger: Addressed in 16d1a19f. I added a one-shot cleanup callback to each slice, invoked by ThreadLocalSliceCache::deallocate() before the slice is deleted or cached. NCCL installs a callback that destroys any remaining cudaEvent_t on the correct CUDA device.                                                                     
                                                                                                                        
Submission-error and polling-completion paths now use the same idempotent helper. The event handle and callback are cleared before cleanup or cache reuse, preventing both double destruction and accidental inheritance by another transport. I also added tests covering cleanup through TransferTask::~TransferTask() and callback clearing across slice-cache reuse.

mooncake-transfer-engine/src/transport/nccl_transport/nccl_transport.cpp
akhillanger: Addressed in 16d1a19f using the documentation option. The public NCCL host contract and build constraints now state that the first valid WRITE freezes the registered-buffer catalog, a terminal session failure remains cached for that endpoint/device pair, and subsequent transfers do not retry bootstrap. Recovery requires recreating the NCCL-only TransferEngine on both peers and registering the buffers again.

mooncake-transfer-engine/src/transport/nccl_transport/nccl_transport.cpp
akhillanger: Addressed in 16d1a19f. Session canonicalization now compares the complete (server name, CUDA device) tuple, and the same comparator is used for makeSessionKey() and both outbound and inbound rank-election paths.
At the ordering layer, equal server names on different devices now produce the same canonical key in both directions, complementary ranks, and exactly one rank-0 unique-ID creator. Same-engine transfers remain intentionally rejected and should use the intra-node NVLink/P2P path; this fixes the latent ordering assumption without claiming same-engine NCCL support.

UNIDY2002:
> Device code currently supports full-world GIN connectivity; rail-team connectivity is not included.

Kernels such as DeepEP V2 require rail-team connectivity. Is there any plan or interest to support this feature in the future? @akhillanger

akhillanger:
Good point. DeepEP V2's hybrid path does require rail connectivity: it checks railedGinType, requests NCCL_GIN_CONNECTION_RAIL, and uses ncclTeamTagRail for the inter-node phase. DeepEP also retains a direct mode using full world connectivity, so this requirement applies specifically to its hierarchical hybrid path rather than every V2 kernel.

This PR implements only full world-team GIN connectivity. We can support rail connectivity as a follow-up so this initial backend remains focused.

mooncake-transfer-engine/example/nccl_device_transport_example.cu
```diff
@@ -0,0 +1,338 @@
+// Copyright 2024 KVCache.AI
```
staryxchen:
```suggestion
// Copyright 2026 KVCache.AI
```

mooncake-transfer-engine/include/transport/device/nccl_device.cuh
```diff
@@ -0,0 +1,302 @@
+// Copyright 2024 KVCache.AI
```
staryxchen:
```suggestion
// Copyright 2026 KVCache.AI
```

mooncake-transfer-engine/src/transport/device/nccl_device_transport.cpp
```diff
@@ -0,0 +1,545 @@
+// Copyright 2024 KVCache.AI
```
staryxchen:
```suggestion
// Copyright 2026 KVCache.AI
```

mooncake-transfer-engine/tests/device_backend_api_compatibility_test.cu
```diff
@@ -0,0 +1,80 @@
+// Copyright 2024 KVCache.AI
```
staryxchen:
```suggestion
// Copyright 2026 KVCache.AI
```

staryxchen (commented):
LGTM, only a few nits

akhillanger:
@staryxchen fixed the copyright issues you pointed. Thanks.

staryxchen (approved):
LGTM

UNIDY2002 (approved):
LGTM. Let's wait for the CI results.

UNIDY2002:
All checks have passed. cc: @alogfans

alogfans (approved):
LGTM. Will be merged soon.

ykwd (approved):

PR #2854, opened 2026-07-10
[TransferEngine] Validate batch memory registrations

### LINKED ISSUE #2853: [Bug]: registerLocalMemoryBatch accepts invalid buffer lists
## Bug Report

On current `main` (`4188e3ae`), `TransferEngineImpl::registerLocalMemoryBatch()` accepts overlapping entries within the same batch and also accepts zero-length entries.

`checkOverlap()` only checks `local_memory_regions_`, while the new entries are inserted after the entire validation loop. As a result, entries in the incoming batch are never checked against each other. The batch API also lacks the zero-length check performed by `registerLocalMemory()`.

### Reproduction

```cpp
std::array<char, 256> buffer{};

std::vector<BufferEntry> overlapping = {
    {buffer.data(), 128},
    {buffer.data() + 64, 128},
};
EXPECT_EQ(engine.registerLocalMemoryBatch(overlapping, "cpu:0"),
          ERR_ADDRESS_OVERLAPPED);

std::vector<BufferEntry> zero_length = {
    {buffer.data(), 0},
};
EXPECT_EQ(engine.registerLocalMemoryBatch(zero_length, "cpu:0"),
          ERR_INVALID_ARGUMENT);
```

Both calls currently return `0`.

### Expected behavior

Validate the complete batch before invoking any transport:

- Return `ERR_ADDRESS_OVERLAPPED` when entries overlap each other or an existing registration.
- Return `ERR_INVALID_ARGUMENT` for a zero-length entry, consistent with `registerLocalMemory()`.
- Continue allowing adjacent, non-overlapping entries.

I searched open issues and PRs for `registerLocalMemoryBatch` overlap/zero-length handling and did not find an existing report or fix.

feichai0017:
## Description

Fixes #2853.

`registerLocalMemoryBatch()` previously checked each entry only against existing registrations, so overlapping entries within the same batch and zero-length entries could be accepted.

This PR validates the complete batch before invoking transports:

- Rejects zero-length entries with `ERR_INVALID_ARGUMENT`.
- Rejects intra-batch overlaps with `ERR_ADDRESS_OVERLAPPED`.
- Preserves support for adjacent, non-overlapping entries.

## Module

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

**Test commands:**
```bash
cmake --build build-batch --target transport_uint_test -j 16
ctest --test-dir build-batch -R '^transport_uint_test$' --output-on-failure
```

**Test results:**
- [x] Unit tests pass (11 test cases)

## Checklist

- [x] I have updated the documentation (not applicable for this internal validation fix)
- [x] For changes >500 LOC: I have filed an RFC issue (not applicable; 68 insertions / 2 deletions)

## AI Assistance Disclosure

- [x] AI tools were used: Codex assisted with testing

### DIFF
--- a/mooncake-transfer-engine/src/transfer_engine_impl.cpp
+++ b/mooncake-transfer-engine/src/transfer_engine_impl.cpp
@@ -742,7 +742,32 @@ int TransferEngineImpl::mp_unregisterLocalMemory(
 
 int TransferEngineImpl::registerLocalMemoryBatch(
     const std::vector<BufferEntry>& buffer_list, const std::string& location) {
-    for (auto& buffer : buffer_list) {
+    std::vector<BufferEntry> sorted_buffers = buffer_list;
+    std::sort(sorted_buffers.begin(), sorted_buffers.end(),
+              [](const BufferEntry& lhs, const BufferEntry& rhs) {
+                  return reinterpret_cast<uintptr_t>(lhs.addr) <
+                         reinterpret_cast<uintptr_t>(rhs.addr);
+              });
+
+    for (size_t i = 0; i < sorted_buffers.size(); ++i) {
+        const auto& buffer = sorted_buffers[i];
+        if (buffer.length == 0) {
+            LOG(ERROR)
+                << "Transfer Engine does not support zero length memory region";
+            return ERR_INVALID_ARGUMENT;
+        }
+
+        if (i > 0) {
+            const auto& previous = sorted_buffers[i - 1];
+            auto address = reinterpret_cast<uintptr_t>(buffer.addr);
+            auto previous_address = reinterpret_cast<uintptr_t>(previous.addr);
+            if (address - previous_address < previous.length) {
+                LOG(ERROR) << "Transfer Engine does not support overlapped "
+                              "memory region";
+                return ERR_ADDRESS_OVERLAPPED;
+            }
+        }
+
         if (checkOverlap(buffer.addr, buffer.length)) {
             LOG(ERROR)
                 << "Transfer Engine does not support overlapped memory region";
--- a/mooncake-transfer-engine/tests/transport_uint_test.cpp
+++ b/mooncake-transfer-engine/tests/transport_uint_test.cpp
@@ -17,6 +17,7 @@
 #include <gtest/gtest.h>
 #include <sys/time.h>
 
+#include <array>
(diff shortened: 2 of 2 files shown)


staryxchen (approved):
LGTM

he-yufeng (approved):
Reviewed the batch validation — this is correct, including the non-obvious part.

Checking each entry only against its immediate predecessor after sorting by address is enough to reject *any* intra-batch overlap, not just adjacent ones: if two entries `i < k` overlap, then `i` also overlaps `k-1` (sorted, so `entry[i].addr <= entry[k-1].addr <= entry[k].addr`, and `entry[i].end > entry[k].addr >= entry[k-1].addr`), so an adjacent overlapping pair always exists and is caught on the first hit. `address - previous_address` can't underflow since the list is sorted ascending.

Registration still walks the original `buffer_list` and the check pass stays fully upfront, so ordering and the validate-then-register structure are preserved. Tests cover the overlap, zero-length, and adjacent-but-not-overlapping cases. LGTM.

he-yufeng:
Nice fix. The sibling multi-protocol path `mp_registerLocalMemory` (`transfer_engine_impl.cpp`, under `ENABLE_MULTI_PROTOCOL`) has the same intra-batch gap this PR closes for `registerLocalMemoryBatch`. Its Phase-1 pre-check calls `checkOverlap(buffer.addr, buffer.length)` per buffer, but `checkOverlap` only tests against already-committed regions (`hasOverlapLocked` under `mutex_`), and this batch isn't committed until Phase 5. So two buffers that overlap *each other within the same `buffer_map`* both pass the pre-check and both get registered — exactly the case your sorted-adjacency check now catches on the single-protocol path.

One wrinkle vs. `registerLocalMemoryBatch`: `mp_` is keyed by protocol, and registering the same address under multiple protocols (CXL + TCP/RDMA) looks intentional here, so the overlap check probably wants to be scoped *within each protocol's* buffer list (per `entry.second`) rather than across the whole map. Zero-length is already handled there.

I didn't push a change because the path is behind `ENABLE_MULTI_PROTOCOL` (default OFF, and no CI job enables it), so I couldn't build/test it — flagging rather than shipping something untested onto the gated path. Happy to send a per-protocol version mirroring this PR if you'd like, or you can fold it into a follow-up.

PR #2584, opened 2026-06-23
[Store] Add BatchEvict candidate-selection benchmark

jacklin78911-collab:
## Description

Adds a standalone benchmark that drives the real `MasterService::BatchEvict` path at controlled metadata scale, to make the eviction-cycle cost and lock-hold behavior from #2560 reproducible and reviewable.

It changes no production behavior. The only non-test change is adding `benchmarks::BatchEvictBench` as a friend of `MasterService` (matching the existing test-friend pattern) so the benchmark can populate metadata and take `snapshot_mutex_` from within the benchmark.

The benchmark has two parts:

1. `RealBatchEvictScales` (always on): end-to-end `BatchEvict` wall-clock time vs object count. Default scales are small (10k, 100k) to stay cheap in CI; set `MOONCAKE_EVICT_BENCH_LARGE=1` to also run 1M.
2. `SingleWaiterSnapshotMutexProbe` (opt-in via `MOONCAKE_EVICT_BENCH_LOCK_PROBE=1`): per trial, a single `unique_lock` waiter on `snapshot_mutex_` arrives shortly after a `BatchEvict` cycle begins; the benchmark records how long it waits to acquire, and reports p50/p95/max over N trials.

The workload is synthetic (single tenant, all objects expired, all no-pin, one completed memory replica each). It is meant as a reproducible baseline for the scan/lock-hold cost, not a production-latency claim.

## Module
- [x] Mooncake Store (`mooncake-store`)

## Type of Change
- [x] Other (benchmark / test only)

## How Has This Been Tested?
- `cmake --build build --target batch_evict_bench -j$(nproc)`
- `./build/mooncake-store/benchmarks/batch_evict_bench`
- `MOONCAKE_EVICT_BENCH_LARGE=1 ./build/mooncake-store/benchmarks/batch_evict_bench`
- `MOONCAKE_EVICT_BENCH_LOCK_PROBE=1 MOONCAKE_EVICT_BENCH_LOCK_OBJECTS=10000 MOONCAKE_EVICT_BENCH_LOCK_TRIALS=3 ./build/mooncake-store/benchmarks/batch_evict_bench`
- `clang-format-20` applied.

Relates to #2560.

## Checklist
- [x] I have performed a self-review of my own code.
- [x] I have formatted my own code using clang-format before submitting.
- [x] I have verified the benchmark builds and runs.

Issue #2560, opened 2026-06-22
[Performance]: BatchEvict full-metadata scan dominates eviction-cycle time at 1M objects (and holds snapshot_mutex_ for that duration)


jacklin78911-collab:
### Describe the issue

While reading `MasterService::BatchEvict` (`mooncake-store/src/master_service.cpp`), I wrote a small gtest benchmark that drives the real `BatchEvict` path with benchmark-only timing, to see where its time goes at large metadata scale. Filing the measurements to confirm whether this is worth addressing before proposing anything.

The candidate-selection path traverses all `kNumShards` × tenants × metadata each cycle, builds transient `candidates` vectors of `lease_timeout`, and runs `std::nth_element` to find the threshold — all under `snapshot_mutex_` held as a shared lock for the duration of the scan.

**All numbers are from a synthetic, single-tenant, all-expired, all-no-pin workload** (every object evictable: one completed memory replica, refcnt 0). The first pass reaches the eviction target, so the second pass is not exercised. Real workloads with soft-pinned / unexpired / multi-tenant objects will differ — I'm reporting the structural cost, not a production latency claim.

## Measurements

Commit `ef0312f8`, 128-core host, single-threaded `BatchEvict`, `evict_ratio_target=0.5`, `evict_ratio_lowerbound=0.25`.

### 1. End-to-end runtime is linear in object count

Bare (non-instrumented) `BatchEvict`:

| objects | runtime |
|---------|---------|
| 10k | ~8.7 ms |
| 100k | ~109 ms |
| 1M | ~0.95–1.0 s (median; n=5 range 930 ms–1.2 s, shared instance) |

### 2. The time is in scanning, not sorting or evicting

Per-phase breakdown at 1M (instrumented; ratios are the point, absolute total is higher due to instrumentation overhead):

| phase | share of cycle |
|-------|----------------|
| metadata scan / traversal | ~73–75% |
| └ candidate vector collection | ~28–31% |
| `std::nth_element` (all calls) | ~0.1% |
| actual eviction (`try_evict_group_or_object`) | <0.5% |

To be clear, I checked whether `nth_element` was the cost — it is not (~0.1%). The dominant cost is the full-metadata traversal to collect candidates.

### 3. The scan holds `snapshot_mutex_`, blocking unique-lock waiters

A probe thread issues a single `unique_lock<shared_mutex>` on `snapshot_mutex_` shortly after a cycle begins and records the wait. 30 trials at 1M:

[...]

PR #2855, opened 2026-07-11
[Store] Make batch_evict_bench scale and eviction ratios configurable

jacklin78911-collab:
## Description
Adds three optional gflags CLI flags to the BatchEvict benchmark so the 10M-objects / 5%-eviction workload discussed in the eviction-optimization work can be reproduced directly from the main-branch benchmark:

- `--num_objects`: run a single custom scale when greater than 0; `0` preserves the default scale list
- `--evict_ratio_target`: eviction target ratio (default 0.50)
- `--evict_ratio_lowerbound`: lower-bound ratio (default 0.25)

Switched from environment variables to gflags CLI flags per review.

Defaults are unchanged: without these flags, the benchmark runs the same scales (10k, 100k, +1M with `MOONCAKE_EVICT_BENCH_LARGE=1`) with the same ratios and identical CSV output. Invalid ratio combinations and invalid flag values fail with a non-zero exit. The existing lock-probe environment variables and their parsing are untouched. This is a benchmark-only change; no production code is modified.

Follow-up to the benchmark added in #2584; relates to the eviction-cycle discussion in #2560.

## Module

## Type of Change
- [x] Other (benchmark-only configurability change)

## How Has This Been Tested?

**Test commands:**
```bash
# Default regression
./build/mooncake-store/benchmarks/batch_evict_bench

# Invalid ratio combination: exits non-zero
./build/mooncake-store/benchmarks/batch_evict_bench \
  --evict_ratio_target=0.05 \
  --evict_ratio_lowerbound=0.5

# Invalid non-numeric flag value: exits non-zero
./build/mooncake-store/benchmarks/batch_evict_bench \
  --num_objects=invalid

# Custom workloads
./build/mooncake-store/benchmarks/batch_evict_bench \
  --num_objects=1000000 \
  --evict_ratio_target=0.05 \
  --evict_ratio_lowerbound=0.05

./build/mooncake-store/benchmarks/batch_evict_bench \
  --num_objects=10000000 \
  --evict_ratio_target=0.05 \
  --evict_ratio_lowerbound=0.05

### DIFF
--- a/mooncake-store/benchmarks/batch_evict_bench.cpp
+++ b/mooncake-store/benchmarks/batch_evict_bench.cpp
@@ -1,8 +1,6 @@
 #include "master_service.h"
 
 #include <algorithm>
-#include <cerrno>
-#include <charconv>
 #include <chrono>
 #include <cmath>
 #include <cstdlib>
@@ -11,8 +9,6 @@
 #include <mutex>
 #include <shared_mutex>
 #include <string>
-#include <string_view>
-#include <system_error>
 #include <thread>
 #include <vector>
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: mooncake-store/benchmarks/batch_evict_bench.cpp [hunk 1]
```diff
@@ -20,10 +24,40 @@ namespace mooncake::benchmarks {
 
 class BatchEvictBench {
    public:
+    static bool ConfigureFromEnv() {
+        if (!ReadEvictionRatios(evict_ratio_target_, evict_ratio_lowerbound_)) {
+            return false;
+        }
+
+        if (std::getenv("MOONCAKE_EVICT_BENCH_OBJECTS") != nullptr ||
```
yokinoshitayoki: Could we expose these three new benchmark parameters as gflags command-line options instead of environment variables, for example --num_objects, --evict_ratio_target, and --evict_ratio_lowerbound. This benchmark already uses gflags, and CLI flags are consistent with the other Store benchmarks, discoverable through --help, and avoid unnecessary custom environment-variable parsing.

yokinoshitayoki (approved):
Thanks for your contribution, this PR looks good to me now.

Issue #2856, opened 2026-07-11
[RFC]: TENT QoS Contract Schema and Effective Policy Model


catyans:
## Summary

This RFC proposes a deliberately small TENT QoS Contract M1 at the `tenant_id -> intent -> request` boundary.

M1 defines only the policy fields needed by the next enforcement milestone:

- `priority`
- `max_inflight_bytes`
- `max_inflight_requests`
- `allowed_degraded_actions`

It provides fail-closed configuration parsing, deterministic effective-policy resolution, and resolution-only diagnostics. M1 does **not** enforce those fields at runtime. Runtime priority authorization and inflight policing remain M2 work.

Bandwidth caps, burst budgets, weighted scheduling, deadline profiles, named contracts, and strict runtime lookup are intentionally deferred until their semantics and consumers exist. M1 does not freeze an API for M3-M5 features.

## Motivation

TENT already carries request priority, deadline, intent type, transport-selection hints, and bandwidth telemetry, but it lacks a small policy boundary that answers the first production isolation questions:

- what priority is a tenant/intent authorized to use?
- how many unfinished bytes may it place in the runtime?
- how many small requests may it keep inflight?
- which degraded actions may a future enforcement layer apply?

The policy model should grow in lockstep with runtime enforcement. Defining fields years or milestones before they are consumed creates misleading diagnostics and prematurely freezes semantics.

## Non-goals for M1

M1 does not:

- change runtime admission, dispatch, transport selection, or scheduling;
- guarantee or cap bandwidth;
- implement token buckets, borrowing, WDRR, or receiver-credit allocation;
- define `min_bandwidth_gbps`, `max_bandwidth_gbps`, `burst_bytes`, or `weight`;
- define `deadline_profile` before its runtime semantics exist;
- define named contracts or allow `policy_name` to bypass tenant/intent resolution;
- define strict rejection of unknown runtime tenant/intent values;
- replace Mooncake Store tenant storage quota;
- expose hardware-specific SL/TC details to applications.

## Tenant identity

`tenant_id` is an opaque administrative isolation identifier supplied by the caller.

When Mooncake Store is the caller, it should pass the canonical Store tenant ID value rather than create a second TENT-specific tenant namespace. TENT must not depend on Store implementation headers, and network QoS quota remains distinct from Store capacity quota, but both layers should use the same tenant identity value at their integration boundary.

[...]

PR #2857, opened 2026-07-11
[TENT] Add QoS contract schema resolver

catyans:
## Motivation

TENT carries priority, deadline, and intent metadata, but it lacks a minimal contract boundary for the next runtime-enforcement milestone. This PR implements the narrowed M1 from #2856: schema parsing, deterministic resolution, and resolution-only diagnostics, with no runtime behavior change.

## M1 scope

The schema contains only fields with a near-term M2 consumer:

- `priority`
- `max_inflight_bytes`
- `max_inflight_requests`
- `allowed_degraded_actions`

Resolution is deliberately limited to:

```text
global defaults -> tenant defaults -> tenant + intent override
```

`tenant_id` is an opaque caller-supplied identity. A Store integration should pass its canonical Store tenant ID value rather than create a parallel TENT tenant namespace.

## Changes

- Add `QosContractResolver` for fail-closed parsing and effective-policy resolution.
- Preserve compatibility when `qos` is absent: the resolver stays disabled and preserves requested priority.
- Normalize opaque `tenant_id` and intent keys.
- Validate priority, byte units, request counts, duplicate tenants, unknown keys, and supported degraded actions.
- Preserve explicit empty `allowed_degraded_actions: []` as an override instead of inheritance.
- Add `explainJson()` with `diagnostic_scope: resolution_only` and no planned-enforcement claims.
- Add focused tests for inheritance, fallback, empty identity, explicit-empty override, invalid schema, state reset, and deferred-feature rejection.

## Explicitly deferred

This PR does not define or accept:

- `min_bandwidth_gbps` / `max_bandwidth_gbps`
- `burst_bytes`
- `weight`
- `deadline_profile`
- global `intent_defaults`
- named contracts / QoS `policy_name` override
- runtime `strict_mode`

These fields and features fail closed if configured. They should be introduced only with the milestone that defines and enforces them.

## Runtime non-goals

- No priority authorization or inflight enforcement.
- No admission/dispatch behavior change.
- No bandwidth policing, borrowing, or weighted scheduling.
- No receiver-credit allocation.
- No Store implementation dependency.
- No default behavior change.

## Cluster validation

Validation ran remotely on the Lingjun H20 cluster, not on the local workstation.

| Item | Value |
|---|---|
| Node | `lingjun-100` |
| Container | `codex_mooncake_pr_eval` |
| Commit | `251a029` |
| Compiler / CMake | GCC 13.3.0 / CMake 3.28.3 |
| Release config | `USE_TENT=ON`, `BUILD_UNIT_TESTS=ON`, CUDA/Store disabled |
| Sanitizer config | Debug, ASan+UBSan, leak detection enabled |

No local-to-remote `scp` was used; the cluster cloned the branch directly.

### Release build and adjacent regression

Built these eight targets:

- `tent_qos_contract_test`
- `admission_queue_test`
- `promotion_policy_test`
- `tent_runtime_queue_dispatch_test`
- `tent_transport_hint_test`
- `tent_transport_selector_test`
- `tent_intent_type_test`
- `transfer_engine_config_override_test`

CTest result:

```text
100% tests passed, 0 tests failed out of 8
Total Test time (real) = 1.05 sec
```

### Repeated resolver campaign

```text
tent_qos_contract_test: 1000/1000 passed
0 failures
Total Test time (real) = 1.66 sec
```

### ASan+UBSan

Independent Debug build with:

```text
-fsanitize=address,undefined -fno-omit-frame-pointer
ASAN_OPTIONS=detect_leaks=1:halt_on_error=1
UBSAN_OPTIONS=halt_on_error=1
```

Result:

```text
tent_qos_contract_test: 1/1 passed
0 ASan findings
0 UBSan findings
0 leak findings
```

`git diff --check` also passes.

## Claims boundary

The evidence validates configuration parsing, resolution precedence, fallback, fail-closed rejection, diagnostics, and memory/undefined-behavior safety for M1. This PR makes no bandwidth, fairness, TTFT, or isolation-performance claim because it intentionally changes no runtime enforcement.

Refs #2856

### DIFF
--- a/mooncake-transfer-engine/tent/include/tent/runtime/qos_contract.h
+++ b/mooncake-transfer-engine/tent/include/tent/runtime/qos_contract.h
@@ -31,34 +31,22 @@ namespace tent {
 
 struct QosPolicyFields {
     std::optional<int> priority;
-    std::optional<double> min_bandwidth_gbps;
-    std::optional<double> max_bandwidth_gbps;
-    std::optional<uint32_t> weight;
-    std::optional<uint64_t> burst_bytes;
     std::optional<uint64_t> max_inflight_bytes;
     std::optional<uint64_t> max_inflight_requests;
-    std::optional<std::string> deadline_profile;
     std::optional<std::vector<std::string>> allowed_degraded_actions;
 };
 
 struct QosRequestContext {
-    // Tenant is the administrative isolation domain for QoS resolution
-    // (for example a user, workload group, or serving deployment). When no
-    // explicit policy_name is matched, the resolver applies this tenant's
-    // defaults and then this tenant's intent-specific override. Empty values
+    // Opaque administrative isolation identifier supplied by the caller. A
+    // Store integration should pass its canonical Store tenant ID here rather
+    // than create a second TENT-specific tenant namespace. Empty values
     // normalize to "default".
-    std::string tenant = "default";
+    std::string tenant_id = "default";
     // Intent is the normalized business meaning of the transfer
     // (foreground_get, background_prefetch, checkpoint, ...). When no explicit
-    // policy_name is matched, it first selects global intent defaults and then
     // selects the tenant-local intent override. Empty values normalize to
     // "unspec".
     std::string intent = "unspec";
-    // Optional explicit contract name requested by the caller. If it matches a
-    // named contract, that contract is selected after global defaults and the
-    // tenant/intent lookup is skipped. If it is unknown, compatibility mode
-    // falls back to tenant/intent resolution; strict mode rejects the request.
-    std::optional<std::string> policy_name;
     // Legacy caller priority. It is preserved when no resolved contract sets a
     // priority, and is overridden by the effective QoS contract priority when
     // present.
@@ -69,11 +57,11 @@ struct EffectiveQosPolicy : public QosPolicyFields {
     bool enabled = false;
     bool matched = false;
     // Normalized request identity used for resolution and explain output.
-    std::string tenant = "default";
+    std::string tenant_id = "default";
     std::string intent = "unspec";
     // Name of the contract layer that supplied the most specific match:
-    // compatibility_default, global_default, intent_default.<intent>,
-    // <tenant>.default, <tenant>.<intent>, or an explicit policy_name.
+    // compatibility_default, global_default, <tenant>.default, or
+    // <tenant>.<intent>.
     std::string matched_contract = "compatibility_default";
     int requested_priority = PRIO_HIGH;
     int effective_priority = PRIO_HIGH;
@@ -84,7 +72,6 @@ class QosContractResolver {
     Status loadFromConfig(const Config& config);
 
     bool enabled() const { return enabled_; }
-    bool strictMode() const { return strict_mode_; }
 
     Status resolve(const QosRequestContext& context,
                    EffectiveQosPolicy* out) const;
@@ -113,11 +100,8 @@ class QosContractResolver {
     static void fieldsToJson(json* out, const QosPolicyFields& fields);
 
     bool enabled_{false};
-    bool strict_mode_{false};
     QosPolicyFields global_defaults_;
-    std::unordered_map<std::string, QosPolicyFields> intent_defaults_;
     std::unordered_map<std::string, TenantContract> tenants_;
-    std::unordered_map<std::string, QosPolicyFields> named_contracts_;
 };
 
 }  // namespace tent
--- a/mooncake-transfer-engine/tent/include/tent/runtime/qos_contract.h
+++ b/mooncake-transfer-engine/tent/include/tent/runtime/qos_contract.h
@@ -38,7 +38,7 @@ struct QosPolicyFields {
     std::optional<uint64_t> max_inflight_bytes;
     std::optional<uint64_t> max_inflight_requests;
     std::optional<std::string> deadline_profile;
-    std::vector<std::string> allowed_degraded_actions;
+    std::optional<std::vector<std::string>> allowed_degraded_actions;
 };
 
 struct QosRequestContext {
--- a/mooncake-transfer-engine/tent/include/tent/runtime/qos_contract.h
+++ b/mooncake-transfer-engine/tent/include/tent/runtime/qos_contract.h
@@ -1,0 +1,105 @@
+// Copyright 2026 KVCache.AI
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef TENT_RUNTIME_QOS_CONTRACT_H
+#define TENT_RUNTIME_QOS_CONTRACT_H
+
+#include "tent/common/config.h"
+#include "tent/common/status.h"
+#include "tent/common/types.h"
+#include "tent/thirdparty/nlohmann/json.h"
+
+#include <cstdint>
+#include <optional>
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+namespace mooncake {
+namespace tent {
+
+struct QosPolicyFields {
+    std::optional<int> priority;
+    std::optional<double> min_bandwidth_gbps;
+    std::optional<double> max_bandwidth_gbps;
+    std::optional<uint32_t> weight;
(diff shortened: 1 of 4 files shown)


catyans:
Addressed the Gemini review comments in 41ceecdf:

- Wrapped QoS config loading with fail-closed exception handling and state reset on parse failure.
- Added unknown-key validation for top-level qos, tenant blocks, and policy fields to catch config typos.
- Changed allowed_degraded_actions to optional<vector<string>> so an explicit empty list overrides inherited actions.
- Revalidated min/max bandwidth after policy merging, including named contracts and resolved effective policy.
- Added regression tests for unknown keys, bad version type, optional empty degraded actions, and merged bandwidth conflicts.

Validation rerun on Lingjun cluster node lingjun-099 / container sglang_shouq:

```text
git diff --check
ctest -R "tent_qos_contract_test|admission_queue_test|promotion_policy_test|tent_runtime_queue_dispatch_test|tent_transport_hint_test|tent_transport_selector_test|tent_intent_type_test|transfer_engine_config_override_test" --output-on-failure
100% tests passed, 0 tests failed out of 8
```

catyans:
Updated the review-fix commit to 441fe8bd after replacing the intentionally misspelled unknown-key test fixture with `unexpected_qos_key` so the typos check can pass. Re-ran the same Lingjun cluster validation: 8/8 related tests passed.

mooncake-transfer-engine/tent/include/tent/runtime/qos_contract.h
```diff
@@ -0,0 +1,105 @@
...
+    std::optional<uint64_t> max_inflight_bytes;
+    std::optional<uint64_t> max_inflight_requests;
+    std::optional<std::string> deadline_profile;
+    std::optional<std::vector<std::string>> allowed_degraded_actions;
+};
+
+struct QosRequestContext {
+    std::string tenant = "default";
+    std::string intent = "unspec";
+    std::optional<std::string> policy_name;
```
alogfans: `tenant`, `intent`, and `policy_name` need clearer semantic definitions, especially around how each one influences resolution and final policy selection.

catyans: Done. I clarified the semantic contract in `QosRequestContext` and `EffectiveQosPolicy`:

- `tenant`: administrative isolation domain; used for tenant defaults and tenant+intent overrides when no explicit policy is selected.
- `intent`: normalized business meaning; used for global intent defaults and tenant-local intent overrides.
- `policy_name`: explicit named contract; if it matches, it selects that contract and skips tenant/intent lookup; unknown names fall back in compatibility mode and fail in strict mode.
- `requested_priority`: preserved unless the resolved contract provides a priority.

I also added `ExplicitPolicyNameSkipsTenantIntentResolution` to lock down the explicit-policy precedence behavior.

Cluster validation:
- node: `lingjun-100` (`8.130.214.133`)
- container: `codex_mooncake_pr_eval`
- commit: `bb5324f6`
- command: `/root/mooncake-pr2857-build/tent/tests/tent_qos_contract_test`
- result: 10/10 tests passed.

staryxchen:
Thanks for this clean M1 — schema/diagnostic only, no runtime behavior change, fail-closed validation. The contract-layer direction makes sense to me.

One thing I'm wondering about: the schema scope feels wider than M1. Mapping fields to when they'll actually be enforced:

| Field | Enforced in |
|---|---|
| `priority`, `max_inflight_bytes`, `max_inflight_requests`, `allowed_degraded_actions` | M2 |
| `max_bandwidth_gbps`, `burst_bytes` | M3 |
| `min_bandwidth_gbps` | M3+ (SLO only) |
| `weight` | M5 |
| `deadline_profile` | no semantics yet |

5 of 9 fields won't take effect until M3-M5. The `enforcement: "planned"` block in `explainJson()` is the related risk — an operator reading explain output may believe protections exist that don't yet. Perhaps either drop fields without enforcement in sight (add them in the milestone that enforces them), or at least don't emit `enforcement` keys for unimplemented policies.

A few smaller questions:

- **Named contracts / `policy_name`**: RFC #2856 lists this as open question #7. Is it worth committing to in M1, or could it defer until `tenant + intent` proves insufficient? It's the largest source of complexity in `loadFromConfig`.
- **Three default layers** (`global → intent_default → tenant.default → tenant.intent`): is the `intent_defaults` middle layer carrying its weight, or would `global + tenant.intent` cover real configs?
- **Tenant identity**: `QosRequestContext.tenant` is a free-form string, disconnected from the existing Store tenant model (`tenant_quota_policy_store.cpp`). The RFC mentions sharing tenant identity — aligning these now might avoid two parallel tenant concepts later.
- **Strict mode in `resolve()`**: when `policy_name` hits a named contract, the unknown-tenant check is skipped. So a valid `policy_name` + unknown tenant is accepted in strict mode. Is that intended?

A possibly simpler M1 shape: tenant identity on `Request` (reused from Store), plus `priority` + `max_inflight_*` + `allowed_degraded_actions`, populated from a small config without versioned YAML / `strict_mode` / three default layers / named contracts. The heavier machinery could grow in lockstep with the milestones that consume it. Schema stability probably doesn't require defining M5 fields in M1.

Curious to hear your view, especially on whether `min_bandwidth_gbps` / `weight` / `deadline_profile` need to land before their enforcement milestones.

catyans:
Thanks @staryxchen — I agree with the core concern. Looking at the milestone map this way, “schema stability” is not a sufficient reason to freeze fields whose semantics and enforcement do not exist yet. The current M1 is wider than the first consumer and makes the diagnostic surface look more complete than the runtime really is.

I will revise #2857 toward the smaller shape rather than defend the current one:

1. **Keep only fields with a near-term M2 consumer**: priority, max_inflight_bytes, max_inflight_requests, and allowed_degraded_actions.
2. **Defer min_bandwidth_gbps, max_bandwidth_gbps, burst_bytes, weight, and deadline_profile** to the milestones that define and enforce them. In particular, deadline_profile should not be committed without semantics.
3. **Remove the planned-enforcement output.** explainJson() should describe resolution only and explicitly identify the diagnostic as schema/resolution-only; it should not enumerate protections that are not active.
4. **Defer named contracts and policy_name.** They are the largest source of resolution complexity and there is no evidence yet that tenant + intent is insufficient.
5. **Remove the global intent_defaults layer.** The initial precedence can be reduced to global -> tenant default -> tenant intent, with no explicit-policy bypass.
6. **Drop M1 strict_mode.** Malformed or unknown configuration keys should still fail closed, but unknown runtime tenant/intent values can use the documented fallback chain until there is a real enforcement reason to make lookup strict.

On tenant identity, I also agree that a second unrelated tenant namespace would be a mistake. TENT should not depend on Store implementation headers, but the request boundary can carry one opaque tenant identifier whose value is supplied from the Store tenant identity when Store is the caller. I will align the naming and contract around that shared identifier instead of treating QosRequestContext.tenant as an independent free-form concept. If the shared Request plumbing needs separate agreement, I will keep that as an explicit prerequisite rather than silently inventing a parallel identity.

The current valid-policy_name plus unknown-tenant behavior was intentional explicit-policy precedence, but it is exactly the kind of surprising rule introduced by committing to named contracts too early. Deferring named contracts removes that ambiguity instead of adding another strict-mode special case.

I will update both RFC #2856 and this PR to reflect the narrower M1, adjust the precedence and diagnostics tests, and then rerun the focused cluster validation before requesting another review.

catyans:
Implemented the narrower M1 discussed above in commit 251a029, and updated RFC #2856 plus the PR description to match the actual boundary.

Key changes:

- reduced QosPolicyFields from nine fields to the four with a near-term M2 consumer: priority, max_inflight_bytes, max_inflight_requests, and allowed_degraded_actions;
- deferred bandwidth min/max, burst, weight, and deadline profile; configuring them now fails closed as an unknown key;
- removed global intent_defaults, named contracts, QoS policy_name override, and runtime strict_mode;
- simplified precedence to global -> tenant default -> tenant intent;
- renamed the request identity to opaque tenant_id and documented that Store integrations should pass the canonical Store tenant ID value without introducing a Store header dependency;
- removed the planned-enforcement object from explainJson() and added diagnostic_scope: resolution_only;
- added regression coverage proving deferred fields/features are rejected, explicit empty degraded actions override inheritance, invalid reload clears enabled state, and empty identity normalization remains compatible.

Remote validation on lingjun-100, container codex_mooncake_pr_eval, fresh clone at 251a029:

- Release build: all 8 focused/adjacent targets built;
- focused/adjacent CTest: 8/8 passed;
- resolver repeat campaign: 1000/1000 passed;
- independent ASan+UBSan Debug build: 1/1 passed, with no ASan, UBSan, or leak findings;
- git diff --check: passed.

The updated PR description contains the environment, exact target list, and claims boundary. This remains schema/resolution-only and makes no runtime-enforcement or performance claim.

@staryxchen, this should now reflect the smaller M1 shape you suggested. The remaining tenant Request plumbing and priority/inflight enforcement are explicitly listed as unimplemented M2 work in the RFC.

staryxchen (approved):
Thanks for the work! LGTM

PR #2867, opened 2026-07-12
[TE] Add host control fallback for IBGDA QP setup

UNIDY2002:
## Description

This PR adds an automatic host-backed control-buffer fallback for the IBGDA Device API transport.

Some devices reject DevX `CREATE_QP` when the IBGDA WQ/DBR control memory is backed by GPU memory. The transport now keeps the GPU-backed control buffer as the default fast path, records the DevX `CREATE_QP` failure status, and retries initialization once with host-backed mapped control memory only for `CREATE_QP` `BAD_PARAM` failures.

The fallback is intentionally narrow: UAR creation failures, CQ creation failures, MR registration failures, QP modify failures, and non-`CREATE_QP BAD_PARAM` errors remain hard failures.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
# Format check with clang-format 20.1.8

# Targeted pre-commit on changed files
pre-commit run --files \
  mooncake-transfer-engine/include/transport/device/ibgda/mlx5gda.h \
  mooncake-transfer-engine/src/transport/device/ibgda_device_transport.cpp \
  mooncake-transfer-engine/src/transport/device/mlx5gda.cpp

# H20/qjh007 container build and smoke
cmake --build build-hostctrl --target device_ibgda_transport_smoke_test -j 16
./build-hostctrl/mooncake-transfer-engine/tests/device_ibgda_transport_smoke_test \
  --device=mlx5_1 --num_ranks=1 --num_qps=1 --gpu_id=0

# End-to-end EP test with IBGDA initialized
CUDA_VISIBLE_DEVICES=0,1 MC_TE_FILTERS=mlx5_1 \
  python3 mooncake-wheel/tests/test_mooncake_ep.py
```

**Test results:**
- [ ] Unit tests pass

Manual validation summary:

- `device_ibgda_transport_smoke_test` passed on `mlx5_1` with the default GPU-backed control path.
- The same smoke on a NIC with UAR creation failure did not trigger the fallback, confirming the retry gate is narrow.
- `test_mooncake_ep.py` passed with IBGDA initialized on `mlx5_1`.
- A temporary no-IPC validation run confirmed the IBGDA data path still passes; the temporary test-only Python change was removed and is not part of this PR.
- A temporary forced host-control run measured a small compatibility-path cost, about 3% on the tested H20 setup.

## Checklist

Notes:

- `pre-commit run --all-files` was not run to avoid applying auto-fixes across unrelated files in the shared worktree. Targeted pre-commit hooks passed for all changed files.
- No docs update is required for this runtime compatibility fallback.
- No test file is added because this path requires specific IBGDA hardware behavior; targeted hardware smoke and end-to-end EP validation were used instead.

## AI Assistance Disclosure

- [ ] No AI tools were used

AI assistance was used to draft and organize the implementation, validation steps, and PR text. The human submitter is responsible for reviewing and defending the final changes.

### DIFF
--- a/mooncake-transfer-engine/include/transport/device/ibgda/mlx5gda.h
+++ b/mooncake-transfer-engine/include/transport/device/ibgda/mlx5gda.h
@@ -87,6 +87,16 @@ struct mlx5gda_qp_devctx {
     uint16_t wq_tail;    // last non-completed wqeid
 };
 
+struct mlx5gda_create_qp_failure {
+    bool valid;
+    uint32_t status;
+    uint32_t syndrome;
+    int sys_errno;
+};
+
+void mlx5gda_reset_create_qp_failure();
+mlx5gda_create_qp_failure mlx5gda_last_create_qp_failure();
+
 struct mlx5gda_qp *mlx5gda_create_rc_qp(struct mlx5dv_pd mpd, void *ctrl_buf,
                                         struct mlx5dv_devx_umem *ctrl_buf_umem,
                                         struct memheap *ctrl_buf_heap,
--- a/mooncake-transfer-engine/src/transport/device/ibgda_device_transport.cpp
+++ b/mooncake-transfer-engine/src/transport/device/ibgda_device_transport.cpp
@@ -335,12 +335,14 @@ class IbgdaDeviceTransportImpl : public RdmaTransport {
     int allocateControlBuffer(bool host_backed) {
         ctrl_buf_host_ = host_backed;
         if (ctrl_buf_host_) {
-            int ret = posix_memalign(&ctrl_buf_, 4096, kCtrlBufSize);
+            void* ptr = nullptr;
+            int ret = posix_memalign(&ptr, 4096, kCtrlBufSize);
             if (ret != 0) {
                 LOG(ERROR) << "[EP IBGDA] posix_memalign ctrl_buf failed: "
                            << ret;
                 return -1;
             }
+            ctrl_buf_ = ptr;
             std::memset(ctrl_buf_, 0, kCtrlBufSize);
 
             cudaError_t err = cudaHostRegister(
--- a/mooncake-transfer-engine/src/transport/device/ibgda_device_transport.cpp
+++ b/mooncake-transfer-engine/src/transport/device/ibgda_device_transport.cpp
@@ -25,11 +25,13 @@
 #include <infiniband/mlx5dv.h>
 #include <infiniband/verbs.h>
 
+#include <cstdlib>
 #include <cstring>
 #include <stdexcept>
 
 #include "cuda_alike.h"
 #include "transport/device/ibgda/memheap.h"
+#include "transport/device/ibgda/mlx5_ifc.h"
 #include "transport/device/ibgda/mlx5gda.h"
 #include "topology.h"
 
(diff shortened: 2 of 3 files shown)


he-yufeng (approved):
Reviewed as a maintainer — traced the fallback path end to end (`allocateControlBuffer` / `createQueuePairs` / `retryWithHostControlBuffer` / `freeControlBuffer`) plus the mlx5gda failure-recording it relies on, not just the diff.

**The narrow trigger is correct.** The retry fires only when the DevX `CREATE_QP` obj_create itself returns `BAD_PARAM`:
- `mlx5gda_create_rc_qp` calls `mlx5gda_reset_create_qp_failure()` on entry (`valid=false`), and `print_devx_create_qp_failure` — the only writer that sets `g_last_create_qp_failure.valid=true` — runs exclusively at the `mlx5dv_devx_obj_create(...create_qp...)` failure site. Every other failure in that function (cap query, `cqe_version`, send CQ, UAR, wq/dbr alloc, memset, stream sync) takes `goto fail` without touching the record, so `isCreateQpBadParam()` stays false and those stay hard failures — matching the description exactly.
- `g_last_create_qp_failure` is `thread_local`, and both the write and the `retryWithHostControlBuffer()` read happen synchronously on the same thread inside `createQueuePairs`, so there's no cross-thread staleness.

**Resource handling across the retry is sound.**
- `retryWithHostControlBuffer()` runs `destroyQueuePairs()` before `freeControlBuffer()`, so the already-created QPs are torn down with a still-valid `ctrl_buf_heap_` before `memheap_destroy()` runs — order is right.
- `mr_` is registered on the externally-supplied data pointer in `registerMemory()`, independent of `ctrl_buf_`, so freeing the control buffer on retry can't dangle it. `qp_devctxs_` is a separate device buffer and is fully rewritten from index 0 when `createQueuePairs()` restarts, so no stale devctx survives the recursion.
- The `ctrl_buf_host_` guard terminates the recursion: a second failure with the host-backed buffer returns false → hard `-1`, no infinite retry.

**The GPU/host pointer split is consistent.** `ctrl_buf_dev_` (equal to `ctrl_buf_` for cudaMalloc, the mapped device pointer from `cudaHostGetDevicePointer` for host-backed) is what feeds the GPU-visible wq/cq/dbr in the devctx, while `ctrl_buf_` (host VA) is what goes to `mlx5dv_devx_umem_reg` and to the CPU-memset path in `mlx5gda_create_cq`. `is_host_control_buffer()` keys off `cudaPointerGetAttributes` and clears the probe error with `cudaGetLastError()`, so the CPU-memset vs `cudaMemsetAsync` selection tracks the actual backing.

Two non-blocking nits:
1. In `allocateControlBuffer(host_backed=true)`, the `cudaHostRegister` failure branch frees `ctrl_buf_` but leaves `ctrl_buf_host_ == true`, whereas the `cudaHostGetDevicePointer` branch resets it to false. Harmless today (`ctrl_buf_` is null so `freeControlBuffer()` skips the free), but resetting it in both branches would keep the `ctrl_buf_host_`/`ctrl_buf_` invariant uniform.
2. `createQueuePairs()` calls `cudaStreamSynchronize(stream)` unconditionally (~line 224), even for the host-backed buffer where `mlx5gda_create_cq` deliberately skips the sync. It looks like a no-op there; if so, gating it on `!ctrl_buf_host_` would mirror the mlx5gda side.

LGTM. Keeping the fallback off the fast path and reacting only to the specific `CREATE_QP BAD_PARAM` signature is the right call.

PR #2871, opened 2026-07-13
[Store] Tune Master defaults based on RPC scaling results

Aionw:
## Description

Tune Mooncake Store Master defaults:

- increase the effective default RPC worker count from 4 to 16 while preserving `rpc_thread_num=0` as the compatibility fallback to `max_threads`
- increase the default KV lease TTL from 5 seconds to 10 seconds
- lower both the in-memory and NoF eviction high-watermark code defaults from 0.95 to 0.90

The sample YAML/JSON configuration continues to set the in-memory eviction high watermark explicitly to 1.0, as requested. It does not explicitly set the NoF high watermark, so NoF inherits the 0.90 code default.

## Benchmark Results

These are local single-host TCP measurements and should be treated as comparative results rather than portable performance baselines. jemalloc was preloaded in both the Master and clients.

### End-to-end Store operations

Configuration: 4 KiB value sizes, a batch size of 32, 64 synchronous client lanes, and 5-second measured phases. Values are the median of three runs; all phases completed without misses or failed objects.

| `rpc_thread_num` | `put_batch` KV/s | `get_batch` KV/s | `batch_get_into` KV/s |
| ---: | ---: | ---: | ---: |
| 1 | 30,211 | 127,665 | 44,895 |
| 8 | 31,193 | 127,374 | 46,551 |
| 32 | 29,459 | 130,937 | 47,268 |

End-to-end throughput does not scale linearly with RPC workers because data movement and client-side processing dominate before the Master worker pool does.

### Master control plane

The isolated workload repeatedly called `BatchGetReplicaList` for 32 prefilled keys, with no object payload transfer. It used 64 synchronous lanes and a 5-second measurement window.

| `rpc_thread_num` | Lookups/s | BatchGet RPC/s | P50 latency |
| ---: | ---: | ---: | ---: |
| 1 | 0.784 M | 24.5 K | 2.520 ms |
| 2 | 1.265 M | 39.5 K | 1.601 ms |
| 4 | 2.126 M | 66.4 K | 0.936 ms |
| 8 | 3.244 M | 101.4 K | 0.624 ms |
| 16 | 3.732 M | 116.6 K | 0.519 ms |
| 32 | 3.870 M | 121.0 K | 0.505 ms |

Control-plane throughput improves by about 4.9x from 1 to 32 workers, with most of the gain reached by 16 workers.

### BatchGet / BatchPut / BatchExists = 1:1:1

The direct C++ control-plane client alternated `BatchGetReplicaList`, `BatchPutStart` + `BatchPutEnd`, and `BatchExistKey`. Configuration: 16 client lanes, a batch size of 32, 8 seconds per point, and no Transfer Engine payload movement.

| `rpc_thread_num` | Mixed cycles/s | Relative to 1 thread |
| ---: | ---: | ---: |
| 1 | 4.20 K | 1.00x |
| 2 | 6.42 K | 1.53x |
| 4 | 8.29 K | 1.97x |
| 8 | 10.27 K | 2.44x |
| 16 | 13.26 K | 3.15x |

One mixed cycle contains one high-level operation of each type; BatchPut itself sends two Master RPCs. All points completed without RPC failures or client segfaults after using one pre-connected `MasterClient` shared by the workers.

## Module

## Type of Change

- [x] Performance improvement
- [x] Other: default configuration tuning

## How Has This Been Tested?

**Test commands:**
```bash
git diff --check
python3 -m json.tool mooncake-store/conf/master.json
SKIP=mooncake-code-format pre-commit run --files \
  mooncake-store/include/types.h \
  mooncake-store/src/master.cpp \
  mooncake-store/conf/master.yaml \
  mooncake-store/conf/master.json
```

The repository-wide format hook was not rerun to completion because it rewrites an unrelated baseline file. Its formatting was applied to the touched C++ files; the targeted trailing-whitespace, YAML, conflict, large-file, codespell, and clang-format hooks pass.

**Test results:**
- [x] Manual RPC scaling testing completed; results are included above
- [x] Targeted pre-commit hooks pass
- [ ] Full unit test suite run

## Checklist

- [x] I have formatted the touched C++ code
- [ ] Documentation changes are required
- [ ] I have added automated tests
- [x] This change is under 500 LOC and does not require an RFC

## AI Assistance Disclosure

- [x] AI tools were used

Codex helped run the RPC scaling benchmarks, analyze the results, update the defaults, and prepare this pull request. The human submitter should review and understand every changed line before marking the PR ready.


ykwd:
Shall we also change the default NoF eviction watermark to 0.9?

he-yufeng (approved):
Reviewed for internal consistency — the defaults are all kept in sync, which is the easy thing to get wrong on a change like this:

- `DEFAULT_DEFAULT_KV_LEASE_TTL` 5000 → 10000 is mirrored by the `static_assert(... == 10000)` guard and the `kDefaultKvLeaseTtlFlagValue` string, so the "update the flag value when the constant changes" contract still holds.
- `max_threads` default 4 → 16 is matched at **both** deprecation-warning sites (`FLAGS_max_threads != 16` at the standalone check and the "both set" check), so a user who leaves `max_threads` unset no longer trips the spurious warning, and the `rpc_thread_num > 0 ? rpc_thread_num : max_threads` fallback is unchanged — someone who only set `max_threads=N` still gets N workers.
- In-memory and NoF `*_EVICTION_HIGH_WATERMARK_RATIO` both drop 0.95 → 0.90, evicting a little earlier.

The values themselves are benchmark-driven (thanks for including the numbers) — I didn't re-run the scaling benchmark, so I'm approving the change as internally consistent and safe rather than re-validating the tuning.

One clarifying note, not blocking: since the sample config sets the in-memory high watermark explicitly to `1.0`, the new `0.90` in-memory code default is only reached by deployments that don't ship the sample config — for anyone using it, in-memory high-watermark eviction effectively stays off and only NoF evicts at 0.90. That matches the description, just worth a line in the config comment so operators don't read the `0.90` code default as the effective in-memory behavior.

LGTM.

ykwd:
The following ci failed:
The following tests FAILED:
	 37 - master_service_ssd_test_for_snapshot (Failed)
Errors while running CTest

@LujhCoconut @Aionw @Icedcoco could you take a look? thanks

Aionw:
@ykwd Thanks for flagging this. Raising the default lease TTL from 5s to 10s made `EvictObject` wait exactly as long as the default 10s client TTL, so the client and segment could expire during snapshot restore and the state comparison failed. Fixed in fb198084 by giving this eviction test an explicit 2s lease TTL; the targeted test now passes locally.

PR #2872, opened 2026-07-13
[TE] Fix RDMA transport rail-failure handling and CQ timeout diagnostic

alogfans:
## Description

This PR implements the RDMA transport rail-failure handling and CQ timeout diagnostic fixes.

1. Rail-failure handling now avoids redispatching slices onto paused or failed peer NIC paths, and falls back to another available peer device when possible.
2. CQ timeout diagnostics now track posted slices so timeout logs can identify stalled completions by peer NIC path.

Changes include:

- Skip paused/failed RDMA rails when selecting or redispatching to a peer NIC path.
- Retry on an available alternate peer device when the selected rail is paused.
- Add bounds checks when selecting alternate peer devices to avoid indexing missing rkeys.
- Track posted RDMA slices for timeout diagnostics.
- Remove completed slices from the posted-slice tracking set when CQ completions are polled.
- Log CQ completion timeout diagnostics, including outstanding CQ count, poll gap, poll interval, and stuck transfer groups by peer NIC.
- Adapt posted-slice tracking to the current multi-QP posting logic by tracking the exact posted slice range for each QP batch.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
cmake --build build --target rdma_transport
git diff --check HEAD~3..HEAD
```

**Test results:**
- [x] Unit tests pass
- [x] Manual testing done: `rdma_transport` target builds successfully.

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex was used to cherry-pick the RDMA robustness commits, resolve conflicts against the current branch, adapt the posted-slice tracking API for the current multi-QP post-send path, and run local build validation.


he-yufeng (commented):
Thanks @alogfans — I went through both halves carefully. The rail-failure handling looks correct, and I have one hot-path concern about the diagnostic tracking.

**Rail-failure handling — looks right.**
- Skipping an unavailable rail in `performPostSend` (fail the batch → redispatch) and, in `redispatch`, falling back to an alternate available peer device before `markFailed()` is the right shape.
- The `alt_dev_id >= peer_segment_desc->buffers[buffer_id].rkey.size()` bound is a genuine safety fix — the previous loop could pick an `alt_dev_id` past the end of `rkey[]` and index a missing key. Good catch that `dest_rkey` is reassigned from `rkey[device_id]` after the switch so the alternate path uses the matching key.
- Resetting `slice->ts = 0` on redispatch is consistent with treating the redispatched post as a fresh attempt for the timeout clock.

**Diagnostic tracking — correct, but I'd gate it.**
- The ordering is right: `trackPostedSlices` before `ibv_post_send` (so a fast completion can't be polled before the slice is registered), and on failure you untrack `[first_failed, wr_count)` — i.e. exactly the WRs from `bad_wr` onward, leaving the successfully-posted ones registered. The poll-side `erase` and the stuck-group-by-`peer_nic_path` summary all line up.
- My concern is cost: `posted_slices_mutex_` is taken on **every** `submitPostSend` and on **every** `performPollCq` pass with completions, unconditionally — even though the stuck-group log only fires on a 5s poll stall / 30s outstanding timeout. On a busy device the post and poll threads now contend on this mutex on the critical path, and each poll erases up to `kPollCount` entries under it. For a transport whose throughput depends on the poll loop, that's a real tax to pay for a diagnostic that almost never triggers.
- Suggestion: gate the tracking (both the insert in `submitPostSend` and the erase in `performPollCq`) behind a config flag, defaulting off, the same way `globalConfig().log_rdma_slice_affinity` gates the affinity logging. The rail-failure fix stands on its own and doesn't need the tracking. That keeps the fast path free unless an operator turns diagnostics on.

One small thing to confirm: the stuck-group summary uses `slice->ts` as the post timestamp (`oldest_post_ts`). Worth double-checking `ts` is set at post time on the normal path (not only reset to 0 on redispatch), otherwise `oldest_post_age_ms` could read as 0 for freshly-posted slices.

Happy to approve once the tracking is gated (or if you'd rather keep it always-on, a quick note on the measured poll-loop impact would settle it).

alogfans:
@he-yufeng Agreed. The registry is only needed for the stuck-group breakdown, so paying the mutex/erase cost on every post and poll by default is too expensive for the RDMA fast path. I’ll put the posted-slice tracking and stuck-group summary behind a default-off diagnostic flag. The base timeout diagnostic can remain ungated. Also confirmed `slice->ts` is set on the normal post path before `ibv_post_send`, so `oldest_post_age_ms` uses the post timestamp when tracking is enabled.

he-yufeng (approved):
Thanks for the quick turnaround, @alogfans — the gating is exactly right. With `MC_TRACK_RDMA_POSTED_SLICES` off (the default), all three hot-path touchpoints early-out before taking the mutex:

- `trackPostedSlices` / `untrackPostedSlices` return immediately on `!globalConfig().track_rdma_posted_slices`;
- the poll-loop erase is now under `if (nr_poll > 0 && globalConfig().track_rdma_posted_slices)`.

So the RDMA post/poll fast path pays only a config read + branch by default, and the stuck-group-by-NIC breakdown is available on demand when an operator is diagnosing stalled completions. The env parse follows the existing `parseBoolConfigEnv` / `MC_ENABLE_PARALLEL_REG_MR` pattern and it's documented as off-by-default with the hot-path caveat.

The rail-failure half still looks correct from my earlier read — skip an unavailable rail, fall back to an available alternate peer device (with the `alt_dev_id >= rkey.size()` bound and the matching `dest_rkey` reassignment), else `markFailed`. LGTM.

PR #2877, opened 2026-07-13
[Store] Defer invalid handle cleanup from segment unmount RPC

Aionw:
## Description

`MasterService::UnmountSegment` previously called `ClearInvalidHandles()` synchronously on the RPC path. That cleanup walked every metadata shard and examined every key, so its cost scaled with the total number of keys managed by the master rather than only the keys stored on the segment being unmounted.

This became visible in a stress scenario with four master RPC threads, ten store clients, approximately eight million keys in total, and continuous batch put/get/exists traffic at a 1:1:1 ratio. Unmounting eight store services concurrently caused multiple RPC workers to perform full metadata scans, during which unrelated clients observed connection timeouts. Profiling showed that invalid-handle scanning and metadata destruction dominated the unmount period.

This PR separates immediate replica invalidation from physical metadata removal:

- Segment allocator registrations own a shared `SegmentLifetime`.
- Allocated MEMORY/NoF buffers bind to the exact allocator registration used for allocation and check that lifetime before exposing a descriptor.
- LOCAL_DISK replicas bind to the corresponding client's `LocalDiskSegment` lifetime, including when an existing LOCAL_DISK replica is reported again after remount.
- `PrepareUnmountSegment` invalidates the matching registration immediately, while draining and graceful unmount only remove it from the allocation pool and keep existing replicas readable until final unmount.
- `PrepareUnmountLocalDiskSegment` invalidates the client's local-disk lifetime immediately; pending offload/promotion paths stop using the unavailable segment.
- Get, batch get, exists, batch exists, regex, promotion, replica-list, and admin query paths exclude completed replicas whose bound segment lifetime is unavailable or whose status is `REMOVED`.
- The unmount RPC commits the segment state and schedules the shared `BackgroundWorker`-based `replica_cleanup_worker_` instead of scanning every metadata shard synchronously.
- `SweepUnavailableReplicas()` performs the deferred physical cleanup and, when HA oplog is enabled, persists stale-replica removal before freeing handles, quota, metadata, or local-disk usage.
- `GetAllKeys` hides metadata without an available completed replica by default and the admin endpoint retains `filter_invalid=false` for callers that need the physical metadata view.
- Runtime registration identity distinguishes same-name mounted segments and CXL client registrations that share one underlying allocator.

The existing `AllocatorManager::getAllocators()` interface remains unchanged. The registration/lifetime view is used only by allocation, replica visibility, and segment lifecycle paths that need exact lifetime tracking.

The behavior change is intentionally limited: replicas become invisible immediately after unmount preparation, while their physical metadata is removed asynchronously. Physical metadata counters may therefore be eventually consistent for a short period. If an object still has another available replica, reads continue using only the surviving replica.

Client expiry still performs a synchronous sweep after invalidating expired clients' segment lifetimes; it now uses the same lifetime-driven availability checks as the asynchronous cleanup path.

Snapshot persistence, snapshot wire format, and snapshot restore behavior are intentionally out of scope for this PR and are left unchanged.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [x] Performance improvement
- [ ] Other

## How Has This Been Tested?

The affected targets were built from the current upstream `main`. Targeted tests cover immediate query invisibility after unmount, admin key filtering, HA-durable stale-replica cleanup, LOCAL_DISK remount rebinding, removed-replica classification, and promotion task cleanup after stale handle removal.

**Build command:**

```bash
cmake --build build --target master_service_tenant_quota_test promotion_on_hit_test master_service_ha_test master_service_test master_admin_server_test --parallel 128
```

**Test commands:**

```bash
./build/mooncake-store/tests/master_service_test --gtest_filter=MasterServiceTest.GetAllKeysFiltersUnavailableMetadataByDefault:MasterServiceTest.UnmountSegmentHidesUnavailableReplicas
./build/mooncake-store/tests/master_service_ha_test --gtest_filter=MasterServiceBatchRecordE2ETest.SweepUnavailableReplicasReleasesResourcesAfterDurable:MasterServiceHATest.AddReplicaRebindsExistingLocalDiskReplicaAfterRemount:MasterServiceHATest.GetReplicaListClassifiesRemovedReplicaStates:MasterServiceHATest.BatchGetReplicaListClassifiesRemovedReplicaStates
./build/mooncake-store/tests/promotion_on_hit_test --gtest_filter=PromotionOnHitTest.BatchRemoveStaleHandleErasesPromotionTask
./build/mooncake-store/tests/master_admin_server_test --gtest_filter=MasterAdminServerWithServiceTest.GetAllKeysReturnsKeys:MasterAdminServerWithServiceTest.GetAllKeysExcludesRemovedKey:MasterAdminServerWithServiceTest.GetAllKeysCanIncludeInvalidMetadataWhenRequested:MasterAdminServerWithServiceTest.GetAllKeysRejectsInvalidFilterInvalidParameter
pre-commit run --files mooncake-store/include/allocation_strategy.h mooncake-store/include/allocator.h mooncake-store/include/master_service.h mooncake-store/include/replica.h mooncake-store/src/master_service.cpp mooncake-store/tests/ha/master_service_ha_test.cpp mooncake-store/tests/master_service_test.cpp mooncake-store/tests/promotion_on_hit_test.cpp
```

**Test results:**

- [x] Unit tests pass

Results:

- Build completed for the listed affected targets
- Master service targeted regressions: 2 passed
- HA/local-disk targeted regressions: 4 passed
- Promotion stale-handle regression: 1 passed
- Admin key filtering regressions: 4 passed
- Touched-file pre-commit hooks passed

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

OpenAI Codex assisted with adapting the change to the latest upstream code, reviewing the implementation, running the listed tests, and drafting this pull request description. The submitter reviewed the resulting code and test results.


mooncake-store/include/allocator.h
```diff
@@ -30,6 +30,22 @@ enum class ReplicaType {
...
+   public:
+    SegmentLifetime() : available_(std::make_shared<std::atomic<bool>>(true)) {}
+
+    [[nodiscard]] bool isAvailable() const {
+        return available_->load(std::memory_order_acquire);
+    }
+
+    void setAvailable(bool available) const {
+        available_->store(available, std::memory_order_release);
+    }
```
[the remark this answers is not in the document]
Aionw: Not take for `KISS` reason.

he-yufeng (commented):
Overall this is a solid approach to the unmount-scan cost — separating immediate visibility from physical cleanup via the per-registration `SegmentLifetime` flag is the right shape. Things I checked and that hold up:

- **All descriptor-returning paths are covered and mutually consistent.** `GetReplicaList`/`BatchGetReplicaList`/regex/the `*ForAdmin` variants go through `GetVisibleReplicaDescriptors`; `ExistKey`/`BatchExistKey`/`GetAllKeys` through `HasVisibleCompletedReplica`. I checked these two can't disagree during the async window: `is_completed_and_available()` gates on `has_invalid_mem/nof_handle()` → `!buffer->isAllocatorValid()`, the same predicate `get_available_descriptor()` uses. So Exist won't say "present" while Get says "not ready".
- **Per-registration lifetimes** correctly disambiguate same-name segments and CXL client segments sharing one underlying allocator, since each `addAllocator` mints a fresh `SegmentAllocator`/lifetime.
- `getDescriptorIfAvailable` re-checks `isAllocatorValid()` *after* building the descriptor — good guard against the flag flipping mid-build.
- `PrepareGracefulUnmount`/`SetSegmentStatusByName` passing `invalidate=false` keeps draining replicas readable, matching the design.

One thing I'd like you to confirm before it merges: **snapshot-restored replica buffers don't get the lifetime bound.** `Serializer<AllocatedBuffer>::deserialize` reconnects `allocator_` to the remounted `buf_allocator`, but the buffer's `segment_lifetime_` stays default-constructed (always-available), and the registration created in `SegmentSerializer::Deserialize` (segment.cpp ~944) is a *separate* lifetime. `bindSegmentLifetime` only runs in `SegmentAllocator::allocate()`, so a restored buffer is never bound. After a master restart, a restored replica's unmount-invalidation therefore falls back to `allocator_.expired()` rather than the flag.

For the common allocator-per-segment case this still invalidates at unmount (the allocator is destroyed once the registration and `buf_allocator` are released), so no functional regression there. But it (a) breaks the "immediate flag" invariant for restored buffers, and (b) for a shared underlying allocator that outlives a single unmount — CXL client segments, the case this PR explicitly calls out — the restored replica stays `isAllocatorValid() == true` and remains visible after unmount, i.e. a stale descriptor.

Could you confirm the restore → unmount path (particularly CXL client segments) is covered? The small defensive fix would be to rebind the registration's lifetime onto the restored buffers where the registration is created in `SegmentSerializer::Deserialize`, so the invariant holds regardless of how a buffer came to exist. A regression test that snapshots → restores → unmounts a memory segment and asserts the object is no longer visible would lock it in.

ykwd:
@Aionw @Icedcoco The following CI failed, could you take a look:
The following tests FAILED:
	 40 - master_metrics_test (Failed)
	 63 - master_service_test_for_snapshot (Failed)
Errors while running CTest

Aionw:
CI failure is expected to be fixed in latest commit.

ykwd:
Since this is a critical change to the master service, I also invited @yokinoshitayoki for a double check.

mooncake-store/src/master_service.cpp
```diff
@@ -1977,22 +2050,21 @@ auto MasterService::UnmountSegment(const UUID& segment_id,
...
         ScopedSegmentAccess segment_access =
             segment_manager_.getSegmentAccess();
         auto err = segment_access.CommitUnmountSegment(segment_id, client_id,
                                                        metrics_dec_capacity);
         if (err != ErrorCode::OK) {
+            invalid_handle_cleanup_.Schedule();
             return tl::make_unexpected(err);
         }
     }
+    invalid_handle_cleanup_.Schedule();
```
yokinoshitayoki: Because metadata cleanup is asynchronous, snapshot creation may start after the segment has been removed but before replicas referencing that segment have been deleted from metadata.

`UnmountSegment()` removes the allocator/segment, schedules cleanup, and then releases its shared `snapshot_mutex_` lock. The cleanup worker acquires the shared lock later, so a snapshot thread waiting for the exclusive lock may run first. In that state, metadata still contains replicas whose segment has already been removed and whose allocator has expired. `MetadataSerializer` then tries to serialize those replicas and fails with `SERIALIZE_FAIL`.

I reproduced this deterministically by delaying the cleanup worker and running mount -> put -> unmount -> PersistState:

    Failed to serialize metadata for key 'key_0':
    buffer.allocator_.expired

The failed snapshot is not published, so the running master remains available, but no new recovery point is created. Repeated unmounts can therefore prevent snapshots from advancing and increase recovery data loss if the master fails before a later snapshot succeeds.

Could we ensure snapshots cannot observe this state? For example, the snapshot path could remove invalid handles after acquiring the exclusive snapshot lock and before fork, or the serializer could consistently exclude unavailable replicas/metadata. Please also add a regression test covering snapshot persistence before the async cleanup completes.

mooncake-store/src/segment.cpp
```diff
@@ -949,8 +944,9 @@ tl::expected<void, SerializationError> SegmentSerializer::Deserialize(
             if (mounted_segment.segment.name == name &&
                 mounted_segment.status == SegmentStatus::OK &&
                 mounted_segment.buf_allocator) {
-                segment_manager_->allocator_manager_.addAllocator(
-                    name, mounted_segment.buf_allocator);
+                mounted_segment.allocator_registration =
+                    segment_manager_->allocator_manager_.addAllocator(
+                        name, mounted_segment.buf_allocator);
                 break;
```
yokinoshitayoki, line 1033: The `break` stops registration restoration after the first segment matching a given name. Multiple segment IDs are allowed to share the same name, so all remaining same-name segments are left with a null `allocator_registration`.

During replica deserialization, the buffer looks up its segment by ID and requires that segment's registration to bind the correct lifetime. A replica belonging to one of the segments without a restored registration therefore fails to deserialize with `missing allocator registration`.

Since `mounted_segments_` is an unordered map, it is also nondeterministic which same-name segment receives the registration. Please restore a separate registration for every matching mounted segment instead of stopping after the first one.

mooncake-store/src/master_snapshot_manager.cpp
```diff
@@ -160,11 +160,21 @@ void MasterSnapshotManager::SnapshotThreadFunc() {
             continue;
         }
 
+        // Capture client liveness before taking snapshot_mutex_. This keeps
+        // the documented client_mutex_ -> snapshot_mutex_ lock order while
+        // still allowing stale handles to be removed from the frozen state.
+        auto alive_clients = master_service_->getAliveClientsSnapshot();
```
yokinoshitayoki: alive_clients is a point-in-time snapshot, but it is later used to delete LOCAL_DISK replicas without being consistent with the metadata state being scanned. A client can remount and add a valid completed disk replica after liveness is captured; ClearInvalidHandles() then treats that replica as stale and may delete the entire key from the running master. The asynchronous cleanup worker has the same race because normal RPCs can proceed during its full scan. Please make client liveness and the cleaned metadata state consistent, or separate invalid memory-handle cleanup from client-liveness cleanup, and add a regression test for remount during cleanup.

Aionw:
Scope update: I removed the snapshot-specific fixes from this PR, including the pre-fork unavailable-replica sweep, restored lifetime/registration rebinding, LOCAL_DISK-only client restore tracking, and the related snapshot round-trip tests.

The reason is to keep #2877 focused on asynchronous invalid-handle cleanup and generation-safe live segment lifetime management. Snapshot consistency and restore behavior have separate correctness and compatibility considerations, so they will be addressed in a dedicated follow-up PR instead of expanding the scope here.

The only changes retained under snapshot-related paths are minimal compatibility updates: existing unmount tests now wait for the asynchronous cleanup worker before fixture persistence, and the catalog provider uses the unified replica availability API.

mooncake-store/include/master_service.h
```diff
@@ -216,7 +217,7 @@ class MasterService {
      * @brief Fetch all keys for a single tenant.
      * @return ErrorCode::OK if exists
      */
-    auto GetAllKeys(const std::string& tenant_id)
+    auto GetAllKeys(const std::string& tenant_id, bool filter_invalid = true)
```
yokinoshitayoki: It seems that `filter_invalid` is currently exposed only on `MasterService`. The admin HTTP handler ignores the request, and `GetAllKeysForAdmin()` always uses the default value, so users cannot request `filter_invalid=false`.

yokinoshitayoki:
@Aionw It seems that this branch has conflicts that must be resolved.

mooncake-store/src/master_service.cpp
```diff
@@ -2023,70 +2016,23 @@ void MasterService::GrantLeaseForGroup(const TenantState& tenant_state,
...
-                                ++it;
-                                continue;
-                            }
-                            ++it;
-                            continue;
-                        }
-                    }
-                    it = EraseMetadata(tenant_state, it, tenant_it->first,
-                                       QuotaEraseMode::kFull, &shard);
+                PruneCompletedUnavailableReplicas(it->second, &shard);
```
yokinoshitayoki: This sweep bypasses the HA cleanup path and releases metadata/quota before the corresponding OpLog entry is durable. For client expiry and LOCAL_DISK cleanup, the standby may retain stale replicas.

mooncake-store/src/master_service.cpp
```diff
@@ -2991,13 +2944,14 @@ auto MasterService::GetReplicaList(const std::string& key,
...
+                if (descriptor) {
+                    replica_list.emplace_back(std::move(*descriptor));
+                }
             });
 
         if (replica_list.empty()) {
-            if (metadata.AllReplicas([](const Replica& replica) {
-                    return replica.status() == ReplicaStatus::REMOVED;
-                })) {
+            if (!metadata.HasAvailableReplica()) {
```
yokinoshitayoki: `HasAvailableReplica()` does not consider replica status, so a `REMOVED` replica with an available lifetime is classified as `REPLICA_IS_NOT_READY` instead of `OBJECT_NOT_FOUND`. Could we preserve the removed-replica check here and in the equivalent batch/admin paths?

mooncake-store/src/master_service.cpp
```diff
@@ -3832,6 +3788,17 @@ auto MasterService::AddReplica(const UUID& client_id, const std::string& key,
...
+    {
+        auto local_disk_access = segment_manager_.getLocalDiskSegmentAccess();
+        auto& local_disk_segments =
+            local_disk_access.getClientLocalDiskSegment();
+        auto segment_it = local_disk_segments.find(client_id);
+        if (segment_it == local_disk_segments.end() ||
+            !segment_it->second->lifetime.isAvailable()) {
+            return tl::make_unexpected(ErrorCode::SEGMENT_NOT_FOUND);
+        }
+        replica.bind_local_disk_lifetime(segment_it->second->lifetime);
```
yokinoshitayoki: When an existing LOCAL_DISK replica is found, the replacement path neither stores the incoming replica nor rebinds the existing one. During an expiry/remount race, the valid report can remain attached to the old lifetime and then be removed by the sweep.

yokinoshitayoki:
@Aionw Should we have a look at the failed CI checks?

yokinoshitayoki (approved):
LGTM.

ykwd (approved):

Icedcoco:
Thanks for continuing to iterate on this. I took another pass over the latest head and wanted to check a few cases where I may be missing some intended compatibility constraints.

1. For `LOCAL_DISK`, `AddReplica()` starts with an independent `SegmentLifetime::Available()` and only replaces it when the client has a registered `LocalDiskSegment`. `NotifyOffloadSuccess()` has a similar fallback when the segment lookup fails. Is accepting a replica without a current segment generation intentional?

    My concern is that these independent lifetimes have no owner that can invalidate them after client expiry. Since `SweepUnavailableReplicas()` now relies solely on `replica.is_available()`, such a replica could remain readable and retain metadata/quota indefinitely, even though its endpoint is no longer reachable. Would it be safer to return `SEGMENT_NOT_FOUND` when the generation is absent, and update tests to mount the local-disk segment explicitly?

2. I am also wondering whether the earlier snapshot race is still possible. `UnmountSegment()` invalidates the allocator and schedules cleanup, but I do not see `SweepUnavailableReplicas()` taking `snapshot_mutex_` in the latest version. A snapshot could therefore fork before cleanup finishes, or while the sweep is mutating metadata, and the child may still hit `buffer.allocator_.expired()` during serialization.

    I saw that snapshot behavior is listed as out of scope, but because this PR changes the state that snapshots can observe, do you think we should close that window here? Perhaps unavailable replicas could be excluded consistently during serialization, or invalidation and the snapshot-visible cleanup could be made atomic with respect to fork.

3. How should the cleanup worker recover when the ordered oplog writer is temporarily full? `Reserve()` can return `TASK_PENDING_LIMIT_EXCEEDED` once `max_entries_per_batch` waiting slots are occupied. The sweep currently logs the failure, advances to the next key, and finishes without scheduling another pass.

    With `max_entries_per_batch=1` and a blocked backend, three stale keys appear sufficient to leave one key unprocessed after the only sweep. Would it make sense to reschedule cleanup when capacity becomes available or from the durable callback? A test with more stale keys than the batch capacity might help clarify the intended behavior.

4. One smaller performance question: every `AllocatedBuffer` still constructs an `Available()` lifetime, which allocates a `shared_ptr<atomic>`, and `SegmentAllocator::allocate()` immediately replaces it with the registration lifetime. That appears to add one allocation/free pair per managed buffer. Would a trivial unbound state, treated as available for compatibility, avoid that cost without adding much complexity?

I would appreciate your thoughts on these cases. The overall direction makes sense to me; I mainly want to make sure these failure and compatibility paths are intentional before the PR merges.

Icedcoco:
After cleanup becomes asynchronous, there is a window after `UnmountSegment` returns where different APIs report different states for the same object:

- `Get` cannot find a readable replica.
- `ExistKey` may still return `true` because the replica remains `COMPLETE`.
- `BatchExistKey` may return `false` because it also checks `metadata.IsValid()`.
- `GetAllKeys` may still include the key.

Is this transitional inconsistency intentional? What object state should clients expect after `UnmountSegment` returns but before background cleanup finishes? If invalid replicas should become logically invisible immediately, should these APIs use the same readability predicate?

Separately, as a non-blocking follow-up, the background sweep keeps one snapshot of alive clients for the entire shard scan. A client could remount and restore a `LOCAL_DISK` replica after the snapshot is taken but before its shard is scanned, causing the restored replica to be treated as stale. This requires a narrow race and likely does not need to block this PR, but we may want to track whether cleanup should revalidate client liveness before deletion or be scoped to handles from the unmounted segment.

Aionw:
Problem 1 fixed in https://github.com/kvcache-ai/Mooncake/pull/2877/commits/2c49b4d5b8d2c29bea453cae23aa87a1835ba1e4. Problem 2 out of the scope of this PR.

Icedcoco:
One additional non-blocking case to consider: the query APIs can still report different states while a copy is in progress.

`CopyStart()` keeps the source replica `COMPLETE` and adds a `PROCESSING` target. If the source segment is unmounted before `CopyEnd()` and background cleanup has not run yet, `ObjectMetadata::IsValid()` remains true because the processing target is still valid, while the unavailable source remains `COMPLETE`. In this window, `ExistKey()` and `BatchExistKey()` return true and `GetAllKeys()` includes the key, but `GetReplicaList()` returns `REPLICA_IS_NOT_READY` because no completed readable replica exists.

I reproduced this deterministically with `CopyStart -> pause cleanup -> UnmountSegment`. I do not think this needs to block the current PR, but it would be useful to clarify the intended object visibility semantics for this mixed state. Do you plan to address it here by using a common completed-and-readable predicate, or track it as a follow-up?

Aionw:
I'd prefer to open a separate PR to look into this. The visibility and lifecycle of segments is a mess right now, and it's not suitable to sort all of that out in this PR.

Aionw:
Thanks, this is a real inconsistency in the async-unmount transition, so I addressed this part in this PR after all.

The intended contract is now: physical metadata may remain until the background sweep completes, but public key-query APIs only expose an object when it has at least one **completed and readable** replica. Commit https://github.com/kvcache-ai/Mooncake/pull/2877/commits/d00bc2614b8156b120759444faae46b7358b4da7 adds a shared `HasReadableReplica()` predicate and uses it in `ExistKey()`, `BatchExistKey()`, and `GetAllKeys()`. `ObjectMetadata::IsValid()` remains an internal metadata/task-liveness predicate rather than a public visibility predicate.

I also added deterministic regressions for both `CopyStart` and `MoveStart`: pause cleanup, unmount the only readable source, and verify that `GetReplicaList`, `ExistKey`, `BatchExistKey`, and `GetAllKeys` all hide the object consistently. The full `master_service_test` suite passes (148/148), and the touched-file pre-commit hooks pass.

There are two deeper pre-existing lifecycle cases discovered while checking this scenario, but they are separate from the visibility regression fixed here:
- copy/move tasks and reserved quota can remain after source cleanup: #3401
- `MoveEnd` can leak the source replica refcount if the target disappears: #3402

Icedcoco (approved):

ykwd:
The following tests FAILED:
Errors while running CTest
	 61 - master_metrics_test (Failed)

PR #2878, opened 2026-07-13
[Transfer Engine] Refresh RDMA metadata on HCA and GID change events

alogfans:
## Description

This PR improves RDMA Transfer Engine fault tolerance when a local HCA becomes unavailable or its GID changes at runtime.

Previously, a node could continue publishing metadata that allowed peers to select a failed HCA, or peers could keep rebuilding endpoints with stale GID information after a runtime GID change. This made recovery depend heavily on repeated endpoint failures.

This change updates the local published RDMA metadata when verbs async events indicate local path state changes:

- On HCA/port failure events, mark the local RDMA context inactive and republish the local topology with inactive HCAs disabled.
- On `IBV_EVENT_PORT_ACTIVE`, mark the context active again and republish topology based on all current context states.
- On `IBV_EVENT_GID_CHANGE`, refresh the local GID, update the published `DeviceDesc.gid`, and disconnect old endpoints so future handshakes rebuild with current metadata.
- Keep `devices[]` stable and disable failed HCAs only through topology selection, preserving `devices[]` / `rkey[]` index alignment.
- Use `nicPathServerName()` when constructing alternate peer NIC paths, fixing dual-NIC / `rdma_server_name` path-key mismatches.

**Notes**
This PR does not implement metadata cache TTL (#2795). Peers still need an existing force-refresh path or a future TTL-based refresh change to observe republished metadata automatically after their cache expires.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
```

**Test results:**
- [X] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used
- [X] AI tools were used (specify below) CodeX

### DIFF
--- a/mooncake-transfer-engine/include/transport/rdma_transport/rdma_context.h
+++ b/mooncake-transfer-engine/include/transport/rdma_transport/rdma_context.h
@@ -170,6 +170,12 @@ class RdmaContext {
         const std::vector<AutoGidSelectionIdentity> &tried_selections = {},
         std::string *previous_gid = nullptr, std::string *next_gid = nullptr);
 
+    // Refresh the runtime GID after IBV_EVENT_GID_CHANGE. Auto-GID mode uses
+    // the same candidate filtering/ranking as initial device open; explicit
+    // MC_GID_INDEX keeps the configured index and refreshes only its value.
+    bool refreshCurrentGid(std::string *previous_gid = nullptr,
+                           std::string *next_gid = nullptr);
+
     ibv_context *context() const { return context_; }
 
     RdmaTransport &engine() const { return engine_; }
--- a/mooncake-transfer-engine/include/transport/rdma_transport/rdma_context.h
+++ b/mooncake-transfer-engine/include/transport/rdma_transport/rdma_context.h
@@ -59,6 +59,12 @@ struct GidSelectionSnapshot {
     int gid_index = -1;
 };
 
+enum class GidRefreshResult {
+    UNCHANGED = 0,
+    CHANGED = 1,
+    FAILED = 2,
+};
+
 struct RdmaCq {
     RdmaCq() : native(nullptr), outstanding(0) {}
     ibv_cq *native;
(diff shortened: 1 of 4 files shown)


he-yufeng (commented):
Nice — the auto-GID-vs-explicit-index split reads cleanly, and the async-event handling is correct: the `GID_CHANGE` branch acks and sets `event_acked = true`, and the trailing ack is guarded by `if (!event_acked)`, so there's no double-ack.

One correctness concern on the failure path. `refreshCurrentGid()` returns `false` for two very different situations: a genuine no-op (the resolved GID equals the current one) *and* a hard failure (`ibv_query_port` / `ibv_query_gid` fails, no suitable candidate, null GID, or `refreshLocalDeviceDesc` fails to publish). `refreshPublishedLocalGid()` collapses both into `changed == false`, logs "current GID did not change", and `doProcessContextEvents()` then skips `disconnectAllEndpoints()`. So if a real `IBV_EVENT_GID_CHANGE` fires but we fail to query or publish the new GID, we log it as a benign no-change and leave every endpoint sitting on the old GID — which is exactly the case this event is meant to recover from, since after a GID change the previous GID may no longer be routable. The stale-GID connections would then fail silently until something else tears them down.

Could `refreshCurrentGid` distinguish "no change" from "failed to refresh" — e.g. a tri-state return, or an out-param/error code — so the failure path can log at ERROR and still disconnect (or trigger a retry) instead of being swallowed as "did not change"? Even just returning a distinct value on the publish/query failures would let the caller treat them differently.

Minor: the `MakeNicPath(peer_segment_desc->name → nicPathServerName())` change in `submitPostSend` looks like a necessary companion to the republished topology names rather than part of the GID/HCA event handling — worth calling out in the description, and confirming the rail-failover path is exercised with the new accessor so the alt-path key still matches what's registered.

alogfans:
@he-yufeng Thanks, this is a good catch. I updated the GID refresh path to distinguish the three cases: UNCHANGED, CHANGED or FAILED. The IBV_EVENT_GID_CHANGE handler now only keeps endpoints when the result is explicitly UNCHANGED. For both CHANGED and FAILED, it disconnects all endpoints. The failure case is logged at ERROR, so we no longer swallow a real refresh/publish failure as a benign no-op or leave old-GID endpoints alive. Also agreed on the nicPathServerName() change. I’ll call that out separately in the PR description as a companion fix for alternate rail key construction in dual-NIC / rdma_server_name setups, so the alt-path key matches the registered endpoint/rail state.

yuechen-sys (approved):
I test it in two-node NIC flap experiment. This implementation is effective.

PR #2805, opened 2026-07-08
[Store] Refactor: Extract snapshot orchestration into MasterSnapshotManager

xiangui33423:
# [Store] Refactor snapshot management: Extract MasterSnapshotManager and Repository

## Summary

This PR refactors the snapshot management code in two phases:

**Phase 1:** Extract `MasterSnapshotManager` from `MasterService`  
**Phase 2:** Extract `MasterSnapshotRepository` from `MasterSnapshotManager`

This refactoring separates concerns into three focused components, making the codebase more maintainable, testable, and aligned with single-responsibility principles.

### Before This PR

`MasterService` was responsible for:
- Core service functionality (metadata, segments, tasks)
- Client management
- Memory management
- **Snapshot scheduling and orchestration**
- **Child process lifecycle management**
- **Serialization coordination**
- **Storage operations (upload, cleanup)**
- **Catalog operations (publish, list, delete)**

This made `MasterService` very large (~8000+ lines) and difficult to maintain.

### After This PR

**MasterService:**
- Core service functionality ✓
- Client management ✓
- Memory management ✓
- Snapshot restore/recovery logic ✓

**MasterSnapshotManager:** (NEW)
- Snapshot scheduling ✓
- Child process lifecycle management ✓
- Serialization orchestration ✓

**MasterSnapshotRepository:** (NEW)
- Storage operations (upload, cleanup) ✓
- Catalog operations (publish, list, delete) ✓

#### New Files Created
- `mooncake-store/include/master_snapshot_manager.h` - Manager class declaration
- `mooncake-store/src/master_snapshot_manager.cpp` - Manager implementation

#### Files Modified
- `mooncake-store/include/master_service.h` - Removed snapshot orchestration methods, added manager member
- `mooncake-store/src/master_service.cpp` - Delegates snapshot operations to manager
- `mooncake-store/src/CMakeLists.txt` - Added new source file

#### Extracted Methods (Master Service → Snapshot Manager)
- `SnapshotThreadFunc()` - Snapshot scheduling loop
- `WaitForSnapshotChild()` - Child process monitoring
- `HandleChildTimeout()` - Timeout handling
- `HandleChildExit()` - Exit status handling
- `PersistState()` - Serialization orchestration
- `BuildSnapshotDescriptor()` - Descriptor construction
- `ResolveSnapshotSequenceId()` - Sequence ID resolution
- `UploadSnapshotPayloadFile()` - File upload
- `CleanupOldSnapshot()` - Retention cleanup
[...]

Issue #2688, opened 2026-06-30
[RFC]: [Store] Split snapshot orchestration and serialization logic from MasterService


Aionw:
## Motivation

Snapshot is not part of the core master data path. Conceptually, it is a persistence and recovery capability that observes, captures, and restores master state, and should be modeled as a separate subsystem rather than intrinsic master request-processing logic.

Keeping snapshot orchestration and serialization inside `MasterService` blurs this abstraction boundary. It also makes snapshot format evolution difficult, since part of the snapshot serialization logic is implemented as a private nested component of `MasterService`.

This RFC proposes splitting snapshot orchestration and snapshot state serialization into dedicated components while preserving the current snapshot behavior and runtime semantics.

### 1. Move snapshot orchestration out of `MasterService`

Introduce a dedicated snapshot manager module, for example:

- `ha/snapshot/master_snapshot_manager.h`
- `ha/snapshot/master_snapshot_manager.cpp`

This component will own snapshot-specific orchestration:

- periodic snapshot scheduling
- snapshot ID generation
- snapshot descriptor construction
- child process lifecycle management
- timeout handling
- snapshot payload upload
- catalog publish
- retention cleanup
- snapshot metrics updates

`MasterService` should only provide the state capture/restore hooks needed by the snapshot manager.

### 2. Extract snapshot state serialization from `MasterService`

Move master snapshot state serialization into standalone modules instead of keeping part of the snapshot format inside `MasterService`.

The current snapshot payload is composed of multiple state files:

- `metadata`: object metadata shards, discarded replicas, and replica ID state
- `segments`: mounted memory segments, client-to-segment mappings, allocator ordering, and local disk segment/offloading state
- `task_manager`: persisted client task manager state
- `manifest.txt`: snapshot serializer type, version, and snapshot ID

The refactor should model these as one master snapshot state bundle, with dedicated serializers for each sub-state. `MasterService` should remain responsible for owning live runtime state, while the snapshot serialization layer should own the persisted representation and format compatibility.

### 3. Define explicit snapshot data structures

Introduce explicit snapshot DTOs for the full master snapshot state, such as:

- metadata snapshot state
- segment manager snapshot state
- task manager snapshot state
- snapshot manifest metadata

These structures should represent the persisted snapshot format independently from the internal layout of `MasterService`, `SegmentManager`, and `ClientTaskManager`.

This will make future snapshot format evolution easier and reduce accidental coupling to private master internals.

[...]

PR #2879, opened 2026-07-13
[Store] Extract snapshot restore path into layered architecture（4/5）

xiangui33423:
## Description

Relates to #2688.

**Part 4 of 5** in the series refactoring master snapshot handling out of the
monolithic `MasterService`.

This PR refactors the snapshot **restore path** into a three-phase layered
architecture, building on the codec extraction from PR 3:

- **PR 1** — Extract snapshot orchestration into `MasterSnapshotManager` (#2805, merged)
- **PR 2** — Make `MasterSnapshotManager::Stop()` non-blocking + repository split (merged)
- **PR 3** — Extract the snapshot codec from `MasterService` (under review: `store/extract-snapshot-codec`)
- **PR 4 — this PR** — Refactor snapshot restore into three-phase architecture
- **PR 5** — Clean up obsolete snapshot code left over after the refactor

### What changed

This PR restructures the monolithic `TryRestoreStateFromSnapshot()` method into
a layered, three-phase restore flow with clear separation of concerns:

**Phase 1: Load candidates** — `MasterSnapshotRepository` scans the catalog for
available snapshots to restore from.

**Phase 2: Download + Decode** — `MasterSnapshotRepository` downloads snapshot
payloads from object storage, then `MasterSnapshotCodec` deserializes them into
in-memory structures.

**Phase 3: Apply state** — `MasterService` applies the deserialized state,
cleans up expired metadata, and rebuilds capacity and allocation metrics.

Each layer has a single, well-defined responsibility:
- **Repository** — Storage layer operations (catalog, object store)
- **Codec** — Data format conversion (serialization/deserialization)
- **MasterService** — Business logic (state application, cleanup, metrics)

#### New constant definitions

Added `snapshot_constants.h` to centralize all snapshot-related constants
(file names, format descriptors, directory paths) — eliminating duplication
across multiple files.

#### New repository methods

- `LoadLatestSnapshot()` — Load the latest snapshot descriptor from catalog
- `LoadRestoreCandidates()` — Load list of all restorable snapshot candidates
- `DownloadSnapshotPayloads()` — Download snapshot data from object storage

#### New service method

- `MasterService::ApplySnapshotState()` — Applies deserialized snapshot state,
  cleans up expired metadata, and rebuilds metrics

#### Before (monolithic)

```
RestoreState() 
  └─ TryRestoreStateFromSnapshot() [single 200-line method]
      ├─ Download manifest
      ├─ Validate version
      ├─ Download all payload files
      ├─ Deserialize data
      ├─ Clean up expired metadata
      └─ Rebuild metrics
```

#### After (layered)

```
RestoreState() [coordinator]
  │
  ├─ Phase 1: Repository.LoadRestoreCandidates()
  │   └─ Load list of restorable snapshots from catalog
  │
  └─ For each candidate:
      │
      ├─ Phase 2a: Repository.DownloadSnapshotPayloads()
      │   ├─ Download manifest
      │   ├─ Validate version and protocol
      │   └─ Download all payload files
      │
      ├─ Phase 2b: Codec.Decode()
      │   └─ Deserialize data into memory structures
      │
      └─ Phase 3: MasterService.ApplySnapshotState()
          ├─ Clean up expired metadata
          └─ Rebuild capacity and allocation metrics
```

### Why this matters

**Separation of concerns** — Each layer can be understood, tested, and modified
independently. Storage changes don't affect business logic, and vice versa.

**Testability** — Individual layers can be unit tested with mocked dependencies.
The repository layer can be tested with a fake object store, the codec with
synthetic payloads, and the service layer with pre-deserialized state.

**Extensibility** — Easy to add new storage backends (swap repository
implementation), support new serialization formats (extend codec), or change
restore strategies (modify service orchestration).

**Maintainability** — Clear boundaries make it easier to locate bugs and
understand the restore flow. Each method has a single, well-defined purpose.

### Backward compatibility

- ✅ The on-disk snapshot format is unchanged
- ✅ Existing snapshots can be restored with the new architecture
- ✅ All existing tests continue to pass
- ✅ No breaking changes to public API
- ✅ Retained `TryRestoreStateFromSnapshot()` method signature (though implementation now delegates to the three-phase flow)

### Files

```
 mooncake-store/include/ha/snapshot/snapshot_constants.h  |  25 ++
 mooncake-store/include/master_service.h                  |  14 ++
 mooncake-store/include/master_snapshot_repository.h      |  24 ++
 mooncake-store/src/master_service.cpp                    | 275 +++++++++++++++------
 mooncake-store/src/master_snapshot_manager.cpp           |  57 ++---
 mooncake-store/src/master_snapshot_repository.cpp        | 209 +++++++++++++++-
 6 files changed, 482 insertions(+), 122 deletions(-)
```

**Net effect:** +360 lines, with much clearer structure and separation.

## Module

## Type of Change

- [x] Refactor

## How Has This Been Tested?

The refactoring preserves all existing behavior and is verified by existing test
suites.

**Test commands:**
```bash
# Build the store library
cd mooncake-store
mkdir -p build && cd build
cmake .. -DBUILD_TESTS=ON
make mooncake_store

# Run snapshot-related tests
make master_service_test_for_snapshot
./master_service_test_for_snapshot

make master_snapshot_codec_test
./master_snapshot_codec_test
```

**Test results:**
- [x] Unit tests pass — `master_snapshot_codec_test` (5/5 tests passed)
- [x] Integration tests pass — `MasterServiceSnapshotTest.MountUnmountSegmentWithOffsetAllocator` passed
- [x] Integration tests pass — `MasterServiceSnapshotTest.PutStartEndFlow` passed
- [x] Manual testing done — Verified restore flow with local file backend

**Sample restore log showing three-phase flow:**
```
I0712 16:01:18.655284 master_service.cpp:5900] [Restore] Backend info: LocalFileSnapshotObjectStore...
I0712 16:01:18.665762 master_service.cpp:7868] Restored Replica::next_id_ to 1
I0712 16:01:18.665879 master_service.cpp:6363] [Restore] Total allocated size after restore: 0
I0712 16:01:18.665885 master_service.cpp:6395] [Restore] Total capacity size after restore: 0
I0712 16:01:18.665889 master_service.cpp:5957] [Restore] Successfully restored state from snapshot: 20260712_160118_622
```

## Checklist

- [x] I have added tests to prove my changes are effective (existing tests verify behavior)
- [x] For changes >500 LOC: I have filed an RFC issue (net +360 lines, but part of approved refactoring series #2688)


Icedcoco:
One behavior-preservation concern: the previous `TryRestoreStateFromSnapshot()` wrapped each restore attempt in an exception boundary, while the new `RestoreState()` path calls `Decode()` and `ApplySnapshotState()` directly.

Most serializer failures are returned as `SerializationError`, but a few MessagePack conversions can still throw outside their local `try` blocks. For example, `TaskManagerSerializer::Deserialize()` calls `arr[0].as<std::string>()` before entering its field-conversion `try`.

In that case, the exception can escape `RestoreState()` and prevent fallback to an older healthy snapshot. Could we preserve the previous per-candidate exception boundary, or otherwise ensure `Decode()` converts all exceptions into `SerializationError`? It would also be helpful to add a fallback test using a structurally valid MessagePack payload with an invalid field type.

mooncake-store/src/master_service.cpp
```diff
@@ -5912,68 +5900,68 @@ void MasterService::RestoreState() {
...
+            LOG(WARNING) << "[Restore] Snapshot candidate "
+                         << snapshot.snapshot_id
+                         << " is unusable: failed to download payloads: "
+                         << payloads_result.error().message;
+            continue;
+        }
+
+        // Phase 2b: Decode payloads (codec responsibility)
+        auto decode_result =
+            snapshot_codec_->Decode(this, payloads_result.value());
```
Aionw: **[P1] Preserve the per-candidate exception boundary**

The previous `TryRestoreStateFromSnapshot()` implementation wrapped the entire download/decode/apply sequence in `try/catch`, so an exception from a malformed snapshot caused the service to reset the partially restored state and fall back to an older candidate.

This new path only handles errors returned through `tl::expected`. However, the existing deserializers still contain throwing MessagePack conversions such as `msgpack::object::as<T>()`. A syntactically valid snapshot with an unexpected field type can therefore throw through this loop. Since `RestoreState()` runs from the `MasterService` constructor, that exception aborts service startup instead of trying the remaining snapshots.

Please retain a `try/catch` boundary around each candidate's download, decode, and apply phases; on exception, reset the partially restored state, log the candidate as unusable, and continue to the next candidate.

mooncake-store/include/ha/snapshot/snapshot_constants.h
```diff
@@ -0,0 +1,25 @@
...
+inline constexpr const char* kSnapshotSerializerVersion = "1.0.0";
+
+// Backup directories
+inline constexpr const char* kSnapshotBackupSaveDir =
+    "mooncake_snapshot_save_backup";
+inline constexpr const char* kSnapshotBackupRestoreDir =
+    "mooncake_snapshot_restore_backup";
+
+// List limit
+inline constexpr size_t kUnlimitedSnapshotList = 0;
```
Aionw: **[P2] Include the declaration for `std::size_t`**

This new header declares `size_t` without including `<cstddef>`, so it only compiles when another header happens to provide the type first. Including `snapshot_constants.h` directly currently fails with `‘size_t’ does not name a type`.

Please add `#include <cstddef>` and declare the constant as:

```cpp
inline constexpr std::size_t kUnlimitedSnapshotList = 0;
```

mooncake-store/src/master_service.cpp
```diff
@@ -5912,68 +5900,68 @@ void MasterService::RestoreState() {
...
+        return;
     }
 
     ResetStateAfterFailedRestoreAttempt();
     LOG(ERROR) << "[Restore] Failed to restore from all candidate snapshots "
-               << "(count=" << restore_candidates.size() << "), starting fresh";
+               << "(count=" << candidates_result->size() << "), starting fresh";
 }
 
 bool MasterService::TryRestoreStateFromSnapshot(
```
Aionw: **[P2] Remove the superseded restore implementation**

`RestoreState()` no longer calls `TryRestoreStateFromSnapshot()` after the restore flow was split across the repository, codec, and apply phases. Keeping this private method leaves a second, now-dead implementation of the complete restore path, including manifest validation, payload downloads, decoding, cleanup, and metric rebuilding.

The two implementations have already diverged in exception handling, and future fixes could easily be applied to only one of them. Please remove both the declaration and definition as part of this PR; carrying two restore paths forward is a blocking maintenance risk.

Aionw (changes requested):
Requesting changes for three blocking issues in the restore-path refactor: preserve per-candidate exception fallback, make the new constants header self-contained, and remove the superseded restore implementation to avoid maintaining two divergent paths.

mooncake-store/src/master_service.cpp
Aionw: Duplicate so marked as resolved.

xiangui33423:
> One behavior-preservation concern: the previous `TryRestoreStateFromSnapshot()` wrapped each restore attempt in an exception boundary, while the new `RestoreState()` path calls `Decode()` and `ApplySnapshotState()` directly.
> 
> Most serializer failures are returned as `SerializationError`, but a few MessagePack conversions can still throw outside their local `try` blocks. For example, `TaskManagerSerializer::Deserialize()` calls `arr[0].as<std::string>()` before entering its field-conversion `try`.
> 
> In that case, the exception can escape `RestoreState()` and prevent fallback to an older healthy snapshot. Could we preserve the previous per-candidate exception boundary, or otherwise ensure `Decode()` converts all exceptions into `SerializationError`? It would also be helpful to add a fallback test using a structurally valid MessagePack payload with an invalid field type.

You're right, and thanks for the precise diagnosis. I confirmed the leak: in TaskManagerSerializer::Deserialize(), arr[0].as<std::string>() (task id) and arr[7].as<std::string>() (assigned client) sit outside the field-conversion try block, so a structurally valid array with a wrongly-typed field there throws msgpack::type_error. That propagates out through DecodeTaskManager → Decode → RestoreState(), and since RestoreState() no longer has the per-candidate try/catch that the old TryRestoreStateFromSnapshot() had, the exception aborts restore and skips fallback to a healthy older snapshot.

I went with the 2nd suggestion — making Decode() guarantee it converts all exceptions into SerializationError — rather than reinstating a boundary in RestoreState(). Reasoning: the codec owns deserialization, so it's the right layer to own translating deserialization faults into the expected<> contract its signature already promises. It also covers the same class of unguarded as<>() conversions that exist elsewhere in the segment/metadata serializers, not just this one call site, so RestoreState() stays a clean expected<>-driven loop and the fallback guarantee doesn't depend on every serializer being individually exception-tight.

I deliberately did not change the per-task handling in TaskManagerSerializer — moving those two as<>() calls inside the try would change restore policy (partially restore a corrupt snapshot vs. fall back to a fully-healthy candidate), which is out of scope for this behavior-preservation fix. The Decode() boundary restores the exact pre-refactor behavior: a throwing candidate becomes an unusable candidate, and restore moves on.

Added the fallback test you asked for: MasterSnapshotCodecTest.DecodeWithInvalidTaskFieldTypeReturnsError builds a structurally valid MessagePack task-manager payload whose id field is an int instead of a string, and asserts Decode() does not throw and returns  DESERIALIZE_FAIL. Full codec suite (6 tests) and task_manager_test (11 tests) pass.

  Changes:
  - mooncake-store/src/ha/snapshot/master_snapshot_codec.cpp — wrap the three decode steps in Decode() with a try/catch(...) that returns SerializationError.
  - mooncake-store/tests/ha/snapshot/master_snapshot_codec_test.cpp — new fallback regression test.

mooncake-store/src/master_service.cpp
```diff
@@ -5912,382 +5900,214 @@ void MasterService::RestoreState() {
...
     const auto now = std::chrono::system_clock::now();
-    for (const auto& snapshot : restore_candidates) {
+    for (const auto& snapshot : candidates_result.value()) {
         ResetStateAfterFailedRestoreAttempt();
-        if (TryRestoreStateFromSnapshot(snapshot, now)) {
-            return;
+
+        // Phase 2a: Download payloads (repository responsibility)
+        auto payloads_result =
+            snapshot_repository_->DownloadSnapshotPayloads(snapshot);
```
Aionw: **[P1] Keep the exception boundary at the candidate orchestration level**

Thanks for adding the exception conversion in `MasterSnapshotCodec::Decode()`; that addresses the known MessagePack conversion case. However, it is narrower than the previous protection: `DownloadSnapshotPayloads()` and `ApplySnapshotState()` still execute outside any exception boundary.

For example, the local-file object store uses the throwing `std::filesystem::exists(path)` overload during download, so a path-specific filesystem error can still escape `RestoreState()`. Because restore runs from the `MasterService` constructor, this aborts service startup instead of resetting the partial state and trying the next candidate.

Please wrap the complete per-candidate download/decode/apply sequence in `RestoreState()` with `try/catch`, resetting state and continuing on exception. The codec-level catch can remain as defense in depth, but it does not replace the orchestration-level boundary that guarantees candidate fallback.

Aionw (commented):
One blocking exception-boundary issue remains after the latest update.

Aionw:
LGTM. After clang-format issue fixed this PR can be merged.

Aionw (approved):

PR #2881, opened 2026-07-13
[Store] Add put/get session APIs for ranged multi-buffer transfers

ascend-direct-dev:
## Description

Add put/get session APIs so callers can reserve/query once, transfer object-byte ranges across multiple buffers, then finalize—without re-querying Master on every layer.

**Related RFC: #2887** ([Layerwise KV Cache Session-Based Ranged Transfer for Mooncake Store](https://github.com/kvcache-ai/Mooncake/issues/2887))

### Store session APIs

- Get: `batch_get_session_start` → `batch_get_into_multi_buffer_ranges` → `batch_get_session_end`
- Put: `batch_put_session_start` → `batch_put_from_multi_buffer_ranges` → `batch_put_session_end` / `batch_put_session_revoke`
- Get session caches a `QueryResult` with a single complete memory replica (`FilterQueryResult`); range paths only use the cache (no Master RPC)
- Object-range overflow checks share `is_object_range_overflow` with `execute_ranged_read`
- Put start reuses `Client::StartBatchPut` (same path as `BatchPut`)
- Session ranged transfers go through `Client::BatchTransferReadRanges` / `BatchTransferWriteRanges` → one `SubmitScatter` per layer (all keys/fragments batched)
- RealClient owns session state; transfer is delegated to Client (no TransferFuture leakage)

### Transfer Engine (Ascend Direct batching)

With many keys, assigning a distinct `task_group_id` per scatter fragment caused `MultiTransport` to issue one `submitTransferTaskGroup` per fragment. Ascend Direct does not override that API, so each fragment became one HiXL `TransferSync` with a single `op_desc`, which collapses layerwise bandwidth.

This PR:

- Lets transports opt out of per-group submission via `Transport::requiresTaskGroupSubmission()` (default `true`, TCP behavior unchanged)
- Sets Ascend Direct to `false`, so one scatter submit becomes one `submitTransferTask` with the full task list
- Includes `opcode` in Ascend slice-dispatcher grouping keys so a mixed READ/WRITE batch cannot share one `TransferSync` direction from `slice_list[0]`

## Module

- [ ] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [x] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
# Session unit tests (lease expiry, abnormal usage, ranged put/get)
PROTOCOL=tcp MC_FORCE_TCP=1 ./mooncake-store/tests/pybind_client_test --gtest_filter='*Session*'

# Real TCP transport e2e
bash mooncake-store/tests/e2e/run_session_ranges_tcp_e2e.sh
```

**Test results:**
- [x] Unit tests pass (`RealClientTest.TestPutGetSessionRanges`, `TestPutGetSessionAbnormal`, `TestGetSessionLeaseExpiredDropsSession`)
- [x] Integration tests pass (TCP e2e `session_ranges_tcp_e2e.py` PASSED)
- [x] Manual testing done (layerwise multi-rank put/get example with per-layer timing + verification)

## Checklist

- [x] For changes >500 LOC: I have filed an RFC issue (#2887)

## AI Assistance Disclosure

- [ ] No AI tools were used

Cursor agent assisted implementation and local validation. The human submitter has reviewed every changed line and can defend the change end-to-end.

### DIFF
--- a/mooncake-integration/store/store_py.cpp
+++ b/mooncake-integration/store/store_py.cpp
@@ -3020,6 +3020,111 @@ PYBIND11_MODULE(store, m) {
             "multiple "
             "keys")
         .def(
+            "batch_get_session_start",
+            [](MooncakeStorePyWrapper &self,
+               const std::vector<std::string> &keys) {
+                if (!self.is_client_initialized()) {
+                    LOG(ERROR) << "Client is not initialized";
+                    return std::vector<int>{};
+                }
+                py::gil_scoped_release release;
+                return self.store_->batch_get_session_start(keys);
+            },
+            py::arg("keys"),
+            "Start a get session: query replicas once and cache them")
+        .def(
+            "batch_get_into_multi_buffer_ranges",
+            [](MooncakeStorePyWrapper &self,
+               const std::vector<std::string> &keys,
+               const std::vector<std::vector<uintptr_t>> &all_buffer_ptrs,
+               const std::vector<std::vector<size_t>> &all_sizes,
+               const std::vector<std::vector<size_t>> &all_src_offsets) {
+                if (!self.is_client_initialized()) {
+                    LOG(ERROR) << "Client is not initialized";
+                    return std::vector<int>{};
+                }
+                py::gil_scoped_release release;
+                return self.store_->batch_get_into_multi_buffer_ranges(
+                    keys, CastAddrs2Ptrs(all_buffer_ptrs), all_sizes,
+                    all_src_offsets);
+            },
+            py::arg("keys"), py::arg("all_buffer_ptrs"), py::arg("all_sizes"),
+            py::arg("all_src_offsets"),
+            "Ranged get into multiple buffers using a get session")
+        .def(
+            "batch_get_session_end",
+            [](MooncakeStorePyWrapper &self,
+               const std::vector<std::string> &keys) {
+                if (!self.is_client_initialized()) {
+                    LOG(ERROR) << "Client is not initialized";
+                    return -1;
+                }
+                py::gil_scoped_release release;
+                return self.store_->batch_get_session_end(keys);
+            },
+            py::arg("keys"),
+            "End a get session and drop cached replica metadata")
+        .def(
+            "batch_put_session_start",
+            [](MooncakeStorePyWrapper &self,
+               const std::vector<std::string> &keys,
+               const std::vector<size_t> &sizes,
+               const ReplicateConfig &config = ReplicateConfig{}) {
+                if (!self.is_client_initialized()) {
+                    LOG(ERROR) << "Client is not initialized";
+                    return std::vector<int>{};
+                }
+                py::gil_scoped_release release;
+                return self.store_->batch_put_session_start(keys, sizes,
+                                                            config);
+            },
+            py::arg("keys"), py::arg("sizes"),
+            py::arg("config") = ReplicateConfig{},
+            "Start a put session: reserve objects without transfer")
+        .def(
+            "batch_put_from_multi_buffer_ranges",
+            [](MooncakeStorePyWrapper &self,
+               const std::vector<std::string> &keys,
+               const std::vector<std::vector<uintptr_t>> &all_buffer_ptrs,
+               const std::vector<std::vector<size_t>> &all_sizes,
+               const std::vector<std::vector<size_t>> &all_dst_offsets) {
+                if (!self.is_client_initialized()) {
+                    LOG(ERROR) << "Client is not initialized";
+                    return std::vector<int>{};
+                }
+                py::gil_scoped_release release;
+                return self.store_->batch_put_from_multi_buffer_ranges(
+                    keys, CastAddrs2Ptrs(all_buffer_ptrs), all_sizes,
+                    all_dst_offsets);
+            },
+            py::arg("keys"), py::arg("all_buffer_ptrs"), py::arg("all_sizes"),
+            py::arg("all_dst_offsets"),
+            "Ranged put from multiple buffers using a put session")
+        .def(
+            "batch_put_session_end",
+            [](MooncakeStorePyWrapper &self,
+               const std::vector<std::string> &keys) {
+                if (!self.is_client_initialized()) {
+                    LOG(ERROR) << "Client is not initialized";
+                    return std::vector<int>{};
+                }
+                py::gil_scoped_release release;
+                return self.store_->batch_put_session_end(keys);
+            },
+            py::arg("keys"), "Complete a put session and make objects readable")
+        .def(
+            "batch_put_session_revoke",
+            [](MooncakeStorePyWrapper &self,
+               const std::vector<std::string> &keys) {
+                if (!self.is_client_initialized()) {
+                    LOG(ERROR) << "Client is not initialized";
+                    return std::vector<int>{};
+                }
+                py::gil_scoped_release release;
+                return self.store_->batch_put_session_revoke(keys);
+            },
+            py::arg("keys"), "Revoke an incomplete put session")
+        .def(
             "get_replica_desc",
             [](MooncakeStorePyWrapper &self, const std::string &key) {
                 py::gil_scoped_release release;
--- a/mooncake-store/include/real_client.h
+++ b/mooncake-store/include/real_client.h
@@ -2,6 +2,8 @@
 
 #include <atomic>
 #include <boost/lockfree/queue.hpp>
+#include <chrono>
+#include <condition_variable>
 #include <csignal>
 #include <map>
 #include <memory>
@@ -243,6 +245,33 @@ class RealClient : public PyClient {
         const std::vector<std::vector<void *>> &all_buffers,
         const std::vector<std::vector<size_t>> &all_sizes,
         const ReplicateConfig &config = ReplicateConfig{});
+
+    std::vector<int> batch_get_session_start(
+        const std::vector<std::string> &keys) override;
+
+    std::vector<int> batch_get_into_multi_buffer_ranges(
+        const std::vector<std::string> &keys,
+        const std::vector<std::vector<void *>> &all_buffers,
+        const std::vector<std::vector<size_t>> &all_sizes,
+        const std::vector<std::vector<size_t>> &all_src_offsets) override;
+
+    int batch_get_session_end(const std::vector<std::string> &keys) override;
+
+    std::vector<int> batch_put_session_start(
+        const std::vector<std::string> &keys, const std::vector<size_t> &sizes,
+        const ReplicateConfig &config = ReplicateConfig{}) override;
+
+    std::vector<int> batch_put_from_multi_buffer_ranges(
+        const std::vector<std::string> &keys,
+        const std::vector<std::vector<void *>> &all_buffers,
+        const std::vector<std::vector<size_t>> &all_sizes,
+        const std::vector<std::vector<size_t>> &all_dst_offsets) override;
+
+    std::vector<int> batch_put_session_end(
+        const std::vector<std::string> &keys) override;
+
+    std::vector<int> batch_put_session_revoke(
+        const std::vector<std::string> &keys) override;
 
     int put_parts(const std::string &key,
                   std::vector<std::span<const char>> values,
@@ -886,6 +915,23 @@ class RealClient : public PyClient {
     mutable std::shared_mutex registered_buffer_mutex_;
     std::unordered_map<void *, size_t> registered_buffer_sizes_;
     std::optional<WritableBufferRegion> local_buffer_region_;
+
+    // KV transfer sessions (process-local; not shared with DummyClient).
+    // get_sessions_ stores a FilterQueryResult'd QueryResult (single complete
+    // memory replica + lease); ranges only compare lease locally (no Master).
+    // Put sessions track writable + inflight so end/revoke can seal the
+    // session and wait for outstanding range writes before finalize/free.
+    struct PutSessionEntry {
+        std::vector<Replica::Descriptor> replicas;
+        uint64_t object_size{0};
+        ReplicaWriteMode write_mode{ReplicaWriteMode::SINGLE_REPLICA};
+        bool writable{true};
+        size_t inflight_transfers{0};
+    };
+    mutable std::mutex session_mutex_;
+    std::condition_variable session_cv_;
+    std::unordered_map<std::string, QueryResult> get_sessions_;
+    std::unordered_map<std::string, PutSessionEntry> put_sessions_;
 
     // Dummy VA -> real VA using mapped_shms; last_hit_shm caches locality.
     bool map_dummy_range_in_shm(const MappedShm &shm, uint64_t dummy_addr,
--- a/mooncake-store/src/real_client.cpp
+++ b/mooncake-store/src/real_client.cpp
@@ -261,6 +261,41 @@ inline QueryResult FilterQueryResult(const QueryResult &qr,
         {replica}, qr.lease_timeout,
         include_object_checksum ? qr.object_checksum : std::nullopt);
 }
+
+// Shared object-byte range overflow check (same semantics as
+// execute_ranged_read / session ranged get-put).
+inline bool is_object_range_overflow(size_t offset, size_t size, size_t limit) {
+    return size > limit || offset > limit - size;
+}
+
+inline const Replica::Descriptor *SelectCompleteMemoryReplica(
+    const std::vector<Replica::Descriptor> &replicas,
+    const std::unordered_set<std::string> &local_endpoints) {
+    const Replica::Descriptor *first_memory = nullptr;
+    for (const auto &r : replicas) {
+        if (r.status != ReplicaStatus::COMPLETE || !r.is_memory_replica()) {
+            continue;
+        }
+        if (local_endpoints.count(r.get_memory_descriptor()
+                                      .buffer_descriptor.transport_endpoint_)) {
+            return &r;
+        }
+        if (!first_memory) {
+            first_memory = &r;
+        }
+    }
+    return first_memory;
+}
+
+inline bool HasMemoryReplica(const std::vector<Replica::Descriptor> &replicas) {
+    for (const auto &r : replicas) {
+        if (r.is_memory_replica()) {
+            return true;
+        }
+    }
+    return false;
+}
+
 }  // namespace
 
 PyClient::~PyClient() {}
@@ -3290,7 +3325,8 @@ tl::expected<int64_t, ErrorCode> RealClient::execute_ranged_read(
             return tl::unexpected(ErrorCode::INVALID_PARAMS);
         }
         size = total_size;
-    } else if (size > total_size || src_offset > total_size - size) {
+    } else if (is_object_range_overflow(src_offset, size,
+                                        static_cast<size_t>(total_size))) {
         LOG(ERROR) << "Range overflow: src_offset=" << src_offset
                    << " + size=" << size << " > total=" << total_size;
         return tl::unexpected(ErrorCode::INVALID_PARAMS);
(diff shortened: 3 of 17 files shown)


mooncake-store/src/transfer_task.cpp
```diff
@@ -1288,6 +1307,41 @@ std::optional<TransferFuture> TransferSubmitter::submitRangeRead(
...
+    uint64_t dst_offset) {
+    std::optional<TransferFuture> future;
+
+    if (replica.is_memory_replica()) {
+        auto& mem_desc = replica.get_memory_descriptor();
+        auto& handle = mem_desc.buffer_descriptor;
+
+        size_t slices_size = 0;
+        for (const auto& s : slices) slices_size += s.size;
+        if (dst_offset + slices_size > handle.size_) {
```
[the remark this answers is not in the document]
ascend-direct-dev: Good catch. Fixed in 72403142. Added an explicit overflow guard before the boundary check in both `submitRangeWrite` and `submitRangeRead`:

```cpp
if (dst_offset > std::numeric_limits<uint64_t>::max() - slices_size ||
    dst_offset + slices_size > handle.size_) {
```

Same guard added to `submitRangeRead` (`src_offset`). UT `TestPutGetSessionRanges` still passes.

mooncake-store/src/real_client.cpp
```diff
@@ -4857,7 +4886,396 @@ std::vector<int> RealClient::batch_put_from_multi_buffers(
...
+            (void)client_->master_client_.BatchPutRevoke({keys[i]});
+            results[i] = static_cast<int>(toInt(ErrorCode::INVALID_REPLICA));
+            continue;
+        }
+        put_sessions_[keys[i]] = PutSessionEntry{
+            .replicas = std::move(replicas),
+            .object_size = static_cast<uint64_t>(sizes[i]),
+        };
+    }
+    return results;
```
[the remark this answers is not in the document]
ascend-direct-dev: Agreed, holding `session_mutex_` across a network RPC is a bottleneck. Fixed in 72403142: keys to revoke and hot-cache removal are now collected under the lock, and the single batched `BatchPutRevoke` + `RemoveHotKeys` call runs outside `session_mutex_`:

```cpp
{
    // ... collect hot_keys_to_remove / keys_to_revoke, populate put_sessions_
}
if (client_->hot_cache_ && !hot_keys_to_remove.empty())
    client_->hot_cache_->RemoveHotKeys(hot_keys_to_remove);
if (!keys_to_revoke.empty())
    (void)client_->master_client_.BatchPutRevoke(keys_to_revoke);
```

mooncake-store/src/real_client.cpp
```diff
@@ -4857,7 +4886,396 @@ std::vector<int> RealClient::batch_put_from_multi_buffers(
...
+                    submit_failed = true;
+                    break;
+                }
+                pending_transfers.emplace_back(i, std::move(*future));
+                transferred += static_cast<int64_t>(sizes[j]);
+            }
+            if (!submit_failed) {
+                results[i] = static_cast<int>(transferred);
+            }
+        }
```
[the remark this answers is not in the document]
ascend-direct-dev: Good point. Added a defensive overflow check against the replica buffer size before submitting range reads in 72403142:

```cpp
const size_t replica_limit =
    replica.is_memory_replica()
        ? replica.get_memory_descriptor().buffer_descriptor.size_ : 0;
// ...
if (replica_limit == 0 || src_offsets[j] > replica_limit ||
    sizes[j] > replica_limit - src_offsets[j]) {
    overflow = true; break;
}
```

On overflow the key returns `INVALID_PARAMS` without submitting any transfer. `TestPutGetSessionRanges` still passes.

mooncake-store/src/real_client.cpp
```diff
@@ -4857,7 +4886,396 @@ std::vector<int> RealClient::batch_put_from_multi_buffers(
...
+    std::lock_guard<std::mutex> lock(session_mutex_);
+    for (size_t j = 0; j < end_keys.size(); ++j) {
+        const size_t i = end_indices[j];
+        put_sessions_.erase(end_keys[j]);
+        if (!end_responses[j]) {
+            results[i] = static_cast<int>(toInt(end_responses[j].error()));
+        } else {
+            results[i] = 0;
+        }
+    }
```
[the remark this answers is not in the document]
ascend-direct-dev: Valid concern. Fixed in 72403142 with a size mismatch guard mirroring the existing `Client::BatchPut` finalize path (`process_end_group`): on `end_responses.size() != end_keys.size()` all affected sessions are erased and results set to `RPC_FAIL` without indexing into the response vector.

```cpp
if (end_responses.size() != end_keys.size()) {
    for (...) { put_sessions_.erase(end_keys[j]); results[i] = toInt(ErrorCode::RPC_FAIL); }
    return results;
}
```

mooncake-store/src/real_client.cpp
```diff
@@ -4857,7 +4886,396 @@ std::vector<int> RealClient::batch_put_from_multi_buffers(
...
+    std::lock_guard<std::mutex> lock(session_mutex_);
+    for (size_t j = 0; j < revoke_keys.size(); ++j) {
+        const size_t i = revoke_indices[j];
+        put_sessions_.erase(revoke_keys[j]);
+        if (!revoke_responses[j]) {
+            results[i] = static_cast<int>(toInt(revoke_responses[j].error()));
+        } else {
+            results[i] = 0;
+        }
+    }
```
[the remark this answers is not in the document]
ascend-direct-dev: Same fix applied to `batch_put_revoke` in 72403142: `revoke_responses.size() != revoke_keys.size()` is guarded, sessions erased and results set to `RPC_FAIL` instead of indexing into the response vector. Consistent with the `process_revoke_group` path in `Client::BatchPut`.

mooncake-integration/store/store_py.cpp
```diff
@@ -2924,6 +2924,110 @@ PYBIND11_MODULE(store, m) {
             "Get object data directly into multiple pre-allocated buffers for "
             "multiple "
             "keys")
+        .def(
+            "batch_get_start",
```
ykwd: `batch_get_session_start` might be better, as this name differentiates this API from ordinary `batch_get`. Similar suggestions to other APIs.

zxpdemonio:
Thanks for the RFC and prototype. I think the session-based ranged transfer direction makes sense for layerwise KV cache load/save, especially for avoiding repeated Master queries after `batch_get_start`.

One concern is that the new `batch_get_into_multi_buffer_ranges` path seems to maintain a separate ranged-read implementation from the existing `get_into_ranges` / `get_into_ranges_internal` path. Since `get_into_ranges_internal` already has mature range validation, buffer-capacity checks, overflow handling, result mapping, and optional `QueryResultCache` support, could we consider factoring the common ranged-read machinery and reusing it for the session-based API? The session API can still keep its own lifecycle and cached replica/lease state, but the low-level range preparation and transfer execution should ideally be shared to avoid duplicated logic and behavior drift.

Relatedly, the PR currently introduces `get_sessions_` as a session-level cache of replica and lease information, instead of reusing the existing request-scoped `QueryResultCache`. I agree that a long-lived session needs different lifecycle semantics from a per-request cache, but it would be helpful to clarify the relationship between these two mechanisms. For example, `batch_get_start` could populate session state, while the actual range transfer path reuses the same internal metadata/range execution helpers used by `get_into_ranges`. This would make the new API easier to review and reduce the chance of inconsistent error handling or lease/bounds behavior between the two ranged-read paths.

mooncake-integration/store/store_py.cpp
ascend-direct-dev: Good suggestion, thanks! Agreed — inserting `session` into the names makes the lifecycle APIs clearly distinct from the ordinary `batch_get` / `batch_put` paths.

ascend-direct-dev:
Thanks for the review — agreed on the direction.

**What we reused / aligned:**
1. **Transfer path**: session ranged get/put go through `Client::BatchTransferReadRanges` / `BatchTransferWriteRanges`, which call the existing `SubmitRangeRead` / `SubmitRangeWrite` → `submitRangeRead` / `submitRangeWrite` (same low-level RangeRead/Write as other paths).
2. **`QueryResult` type**: `get_sessions_` now stores a `QueryResult` (same value type as `QueryResultCache`), built at `batch_get_session_start` via existing `FilterQueryResult` so the vector keeps only the selected complete memory replica + `lease_timeout`. Ranges use `QueryResult::IsLeaseExpired`.
3. **Overflow check**: object-byte range overflow is shared with `execute_ranged_read` via `is_object_range_overflow`.

**What we intentionally do *not* call into:**
- **`QueryResultCache` as the session store**: request-scoped cache vs long-lived start→end session; different miss policy (re-query vs fail). Same `QueryResult` type, different map lifecycle — session stays in `get_sessions_`.

mooncake-store/include/real_client.h
```diff
@@ -243,6 +244,33 @@ class RealClient : public PyClient {
         const std::vector<std::vector<size_t>> &all_sizes,
         const ReplicateConfig &config = ReplicateConfig{});
 
+    std::vector<int> batch_get_session_start(
+        const std::vector<std::string> &keys) override;
+
+    std::vector<int> batch_get_into_multi_buffer_ranges(
```
yokinoshitayoki, line 252: Could we use a more concise and explicit name?

ascend-direct-dev: Thanks for the suggestion — happy to rename if we can land on something clearer.

Could you share the names you have in mind?

Context on the current naming:
- `batch_get_session_start` / `batch_get_session_end` follow yangke's earlier suggestion for the session lifecycle APIs.
- `batch_get_into_multi_buffer_ranges` is intentionally aligned with the existing vLLM-facing API style (`batch_get_into_multi_buffers`), with `_ranges` to indicate object-byte ranged transfers.

Open to a more concise/explicit alternative if you have a preferred spelling.

yokinoshitayoki:
Could you provide benchmark results demonstrating the benefit of layer-wise ranged transfers over the existing batch put/get APIs, such as end-to-end latency, throughput, and Master RPC overhead under representative model workloads?

ascend-direct-dev:
Thanks for the question.

We're currently running the benchmarks and will share the numbers (e2e latency / throughput / Master RPC overhead) once the tests finish.

On the motivation of this API — it serves two distinct purposes:

1. **Common KV pooling**: layer-wise vs non-layer-wise transfer needs a fair comparison of the benefit (Master RPC amortization, overlap with compute, etc.). The session ranged APIs make that measurement possible without re-querying Master on every layer.
2. **KV offload**: layer-wise transfer is **required**, not optional. Without it, the full-model HBM KV footprint does not fit; we have to stream layer by layer. Existing whole-object batch put/get cannot cover this path.

So the API is both a performance optimization for the pooling case and a functional necessity for the offload case. We'll follow up with the pooling-side comparison data shortly.

ascend-direct-dev:
> Could you provide benchmark results demonstrating the benefit of layer-wise ranged transfers over the existing batch put/get APIs, such as end-to-end latency, throughput, and Master RPC overhead under representative model workloads?

Thanks for the question. Below are results from our **in-house connector test harness** (simple e2e check, not a full upstream benchmark suite), focused on **Common KV pooling**.

**Setup:** Qwen3-8B, **32K**, hit ≈ **50%** prefix hit. Both runs do layerwise load/save overlapped with compute; the only difference is key layout / whether session ranged APIs amortize Master metadata.

| setup | cold TTFT (s) | hit TTFT (s) | Put items | Get items |
|---|---:|---:|---:|---:|
| **lw_session** | 5.736 | 2.641 | 329 | 109 |
| **lw_nosession** | 6.280 | 2.901 | 11844 | 3924 |

**Why we treat this as mostly a functional issue for Common KV pooling as well**

Before session + ranged put/get, Store only exposed whole-object batch APIs. To transfer KV **layer by layer**, the only workable approach was **one key per (layer, block)** (`#Lxx`-style). That is not just slower Master traffic — it changes the object model (keys × `num_layers`, lifecycle, hit/lookup shape).  
With this PR: **one key per block**, layers packed in one object and addressed by byte offset (`session_start` → `*_ranges` → `session_end`). Put/Get item counts drop by ~`num_layers` here (`329 × 36 ≈ 11844`).

So for pooling, session ranged APIs are first about enabling the natural “one block / multi-layer” layout for layerwise transfer; the TTFT delta in this simple run is secondary. (KV offload still needs the same ranged path as a hard requirement for HBM footprint.)

mooncake-store/src/shm_helper.cpp
```diff
@@ -53,8 +53,7 @@ bool ShmHelper::cleanup() {
         }
         if (shm->base_addr) {
 #if defined(USE_ASCEND_DIRECT)
-            if (globalConfig().ascend_agent_mode &&
-                globalConfig().ascend_use_fabric_mem) {
+            if (globalConfig().ascend_use_fabric_mem) {
```
yokinoshitayoki: Removing ascend_agent_mode changes the behavior of every ShmHelper caller, including the standalone MooncakeHostMemAllocator and SHM hot cache: they now switch from memfd/mmap to Ascend VMM whenever the process-global FabricMem flag is set.
More importantly, allocation and cleanup select the backend using the flag’s current value rather than the segment’s allocation provenance. For example, a segment allocated via memfd before setup_real() enables FabricMem could later be passed to free_memory("ascend") instead of munmap().
Is standalone FabricMem required by the new session API? If so, could we record the allocation backend per ShmSegment, validate/align the VMM allocation size, and add standalone allocation/free tests, including a configuration-change case? Otherwise, this semantic change seems unrelated and should probably be reverted or split into a separate PR.

ascend-direct-dev: Agreed — standalone FabricMem for `MooncakeHostMemAllocator` / decode host allocation was exploratory and is not required by the session API. Reverted the `ascend_agent_mode` removal in `shm_helper.cpp`; will revisit separately if needed.

mooncake-store/src/real_client.cpp
```diff
@@ -5042,7 +5078,423 @@ std::vector<int> RealClient::batch_put_from_multi_buffers(
...
+    auto revoke_responses = client_->BatchPutRevoke(revoke_keys);
+    std::lock_guard<std::mutex> lock(session_mutex_);
+    if (revoke_responses.size() != revoke_keys.size()) {
+        for (size_t j = 0; j < revoke_keys.size(); ++j) {
+            const size_t i = revoke_indices[j];
+            put_sessions_.erase(revoke_keys[j]);
+            results[i] = static_cast<int>(toInt(ErrorCode::RPC_FAIL));
+        }
+        return results;
```
yokinoshitayoki, line 5758: The local put session is erased even when BatchPutEnd returns a response-size mismatch or a per-key error. For an RPC timeout/failure, the Master state may be unknown: the object may still be PROCESSING, or the finalize may already have succeeded. Once the session is erased, the caller can neither retry session_end nor revoke the reservation, potentially leaving a zombie object until Master cleanup.

mooncake-store/src/real_client.cpp
```diff
@@ -5042,7 +5078,423 @@ std::vector<int> RealClient::batch_put_from_multi_buffers(
...
+    }
+
+    // Same RPC FinalizeBatchPut uses after successful transfers.
+    // Session put path does not compute checksums; leave object_checksum unset.
+    std::vector<ObjectMeta> end_metas;
+    end_metas.reserve(end_keys.size());
+    for (const auto& key : end_keys) {
+        end_metas.emplace_back(ObjectMeta{key, std::nullopt});
+    }
+    auto end_responses = client_->BatchPutEnd(end_metas);
```
yokinoshitayoki: A put session may cache a mixed set of MEMORY and NoF replicas when config.nof_replica_num > 0. However, BatchTransferWriteRanges skips every non-memory replica, while this call uses the default ReplicaType::ALL. As a result, Master can mark an unwritten NoF replica as COMPLETE, making stale or uninitialized data readable.

ascend-direct-dev: Good catch — agreed.

`BatchTransferWriteRanges` only writes MEMORY replicas, so `batch_put_session_end` now finalizes with `ReplicaType::MEMORY` and then revokes any leftover NoF reservations with `ReplicaType::NOF_SSD` (same split `FinalizeBatchPut` uses when only memory transfers succeed).

Landed in 324a4ea3.

mooncake-store/src/real_client.cpp
ascend-direct-dev: Agreed — for the session API this differs from one-shot `FinalizeBatchPut`.

`batch_put_session_end` / `batch_put_session_revoke` now keep the local put session on RPC size mismatch and per-key failure so the caller can retry `session_end` or `session_revoke`. Sessions are erased only after a successful finalize/revoke (and revoke treats `OBJECT_NOT_FOUND` as cleaned).

Covered by `RealClientTest.TestPutSessionKeptAfterEndFailure`. Landed in 324a4ea3.

yokinoshitayoki:

mooncake-store/src/real_client.cpp
```diff
@@ -5155,7 +5191,489 @@ std::vector<int> RealClient::batch_put_from_multi_buffers(
...
+                    overflow = true;
+                    break;
+                }
+                entry_slices.emplace_back(Slice{buffers[j], sizes[j]});
+                entry_offsets.push_back(static_cast<uint64_t>(offsets[j]));
+            }
+            if (overflow) {
+                continue;  // results[i] stays INVALID_PARAMS
+            }
+            replicas_per_entry.push_back(it->second.replicas);
```
yokinoshitayoki, line 5520: Could we add per-session state and in-flight transfer tracking here? The replica descriptors are copied while holding the mutex, but the actual transfer runs after the mutex is released. A concurrent end or revoke may therefore finalize or free the replicas while a range write is still in flight, allowing writes to COMPLETE or reclaimed memory. End/revoke should first make the session non-writable and wait for outstanding transfers to finish.

mooncake-store/src/real_client.cpp
```diff
@@ -5155,7 +5191,489 @@ std::vector<int> RealClient::batch_put_from_multi_buffers(
...
+    const std::vector<std::string> &keys) {
+    std::vector<int> results(
+        keys.size(), static_cast<int>(toInt(ErrorCode::INVALID_PARAMS)));
+    if (!client_) {
+        LOG(ERROR) << "Client is not initialized";
+        return results;
+    }
+
+    // Session put only writes memory replicas (BatchTransferWriteRanges skips
+    // NoF). Finalize MEMORY, then revoke any leftover NoF reservations — same
```
yokinoshitayoki: This fallback is only valid for the FLEXIBLE_DUAL_REPLICA mode. For strict configurations such as multiple memory or NoF replicas, the normal put path requires every allocated replica to complete, while this session path never writes NoF and still returns success after revoking it.

mooncake-store/src/real_client.cpp
```diff
@@ -5155,7 +5191,489 @@ std::vector<int> RealClient::batch_put_from_multi_buffers(
...
+        }
+    }
+    if (start_keys.empty()) {
+        return results;
+    }
+
+    // Same first half as Client::BatchPut (StartBatchPut → master
+    // BatchPutStart).
+    auto start_responses =
+        client_->StartBatchPutForSizes(start_keys, start_sizes, config);
```
yokinoshitayoki: start_keys is filtered to exclude keys that already have a session, but config.group_ids is passed unchanged.

mooncake-store/src/real_client.cpp
ascend-direct-dev: Agreed.

Put sessions now track `writable` + `inflight_transfers`. Range writes only proceed while the session is writable and hold an inflight count across the transfer (RAII release). `batch_put_session_end` / `batch_put_session_revoke` seal the session first, wait for outstanding range writes, then finalize/free.

Landed in 6237e4cc.

mooncake-store/src/real_client.cpp
ascend-direct-dev: Agreed — the MEMORY finalize + NoF revoke split is only valid for `FLEXIBLE_DUAL_REPLICA` (`replica_num == 1 && nof_replica_num == 1`).

`batch_put_session_start` now rejects reliable NoF configs (`nof_replica_num > 0` and not flexible). Sessions record `write_mode`; `batch_put_session_end` only applies that fallback in flexible mode, otherwise keeps the session for `session_revoke`.

Covered by `RealClientTest.TestPutSessionStartRejectsReliableNofConfig`. Landed in 6237e4cc.

mooncake-store/src/real_client.cpp
ascend-direct-dev: Good catch — agreed.

When existing put sessions are skipped from `start_keys`, `group_ids` is now filtered to the same `start_indices`. A length mismatch against the original `keys` is rejected up front.

Covered by `RealClientTest.TestPutSessionStartFiltersGroupIds`. Landed in 6237e4cc.

yokinoshitayoki (approved):
LGTM.

ascend-direct-dev:
@zxpdemonio @alogfans Could you please review the Transfer Engine changes related to the session / layerwise path?

**Background:** With many keys, the session / layerwise path was far slower than the existing `batch_get_into_multi_buffers` / `2p` path. The root cause is that ranged scatter assigns a distinct `task_group_id` per fragment, and `MultiTransport` then issues one `submitTransferTaskGroup` call per group. Ascend Direct does not override the group API, so each fragment becomes one `enqueue` → one HiXL `TransferSync` with a single `op_desc`. At large key counts this explodes into too many ADXL calls.

**What we want:**
1. Store session ranged read/write goes through scatter (`BatchTransferReadRanges` / `BatchTransferWriteRanges` → `submitScatter`), so all keys / fragments of a layer are collected into one scatter submit.
2. Ascend Direct should **not** submit by `task_group_id`. With `requiresTaskGroupSubmission() == false`, `MultiTransport` hands the whole batch to one `AscendDirectTransport::submitTransferTask`, and the dispatcher coalesces by `(target_id, opcode)` into a large `op_descs` list for one `TransferSync`.

**Intentionally unchanged:** the default remains `requiresTaskGroupSubmission() == true`, so TCP still uses per-group `submitTransferTaskGroup` / `startTransferSequence` (sequenced fragments on a reused connection).

Main touch points:
- `mooncake-transfer-engine/include/transport/transport.h` (capability query)
- `mooncake-transfer-engine/src/multi_transport.cpp` (merge submits when the transport opts out)
- `mooncake-transfer-engine/include/transport/ascend_transport/ascend_direct_transport/ascend_direct_transport.h` (Ascend Direct opts out of grouping)
- `mooncake-transfer-engine/src/transport/ascend_transport/ascend_direct_transport/slice_dispatcher.cpp` (include opcode in the grouping key to avoid mixed directions)
- Store session ranges → scatter

zxpdemonio:
Thanks for the detailed analysis. The root cause described here is consistent with the current scatter submission path.

We are also working on #3310, which redesigns and applies the task-group batching idea from #1865 to the current scatter API introduced by #3000. Its intended submission model overlaps with the following changes in this PR:

- Transport::requiresTaskGroupSubmission()
- the Ascend Direct override returning false
- the corresponding MultiTransport conditional

In #3310, transports that do not opt into grouped logical tasks still receive the whole transport batch in one submitTransferTask() call, so Ascend Direct can hand all slices to its dispatcher without a separate Ascend-specific submission policy in MultiTransport.

#3310 is still under review and has not landed in main, so I am not suggesting changing this PR base or removing these changes now. I only want to flag the overlap so that, once the direction of #3310 is settled, we can converge on one submission mechanism instead of maintaining two parallel capability models.

The dispatcher grouping change from target_id to (target_id, opcode) is different: it is an independent correctness fix. processSliceList() derives the ADXL transfer direction from the first slice, so READ and WRITE slices targeting the same segment must not be placed in the same dispatcher group. Keeping that part in this PR looks reasonable; a focused mixed-opcode regression test would make the contract explicit.

ascend-direct-dev:
@zxpdemonio Thanks for the heads-up on the overlap with #3310.

Agreed — once #3310 lands, we should converge on a single submission mechanism rather than keep two parallel capability models. We'll wait for #3310 to merge first, then follow up here (or in a small cleanup PR) to drop the Ascend-specific `requiresTaskGroupSubmission()` path and align with the grouped-scatter contract.

alogfans (approved):

ykwd (approved):

PR #2882, opened 2026-07-13
[Store] Add P2P hot standby runtime

skygragon:
## Description

  Add a P2P-specific hot standby runtime for replaying P2P metadata OpLog entries into `P2PStandbyMetadataStore`.

  This PR introduces `P2PHotStandbyService`, which reuses the shared OpLog and standby components (`StandbyStateMachine`, `OpLogStoreFactory`, `OpLogChangeNotifier`, and `OpLogReplicator`) while keeping the orchestration layer P2P-specific. It
  intentionally does not inherit from the centralized `HotStandbyService` because P2P promotion needs `P2PStandbyMetadataStore`, `P2POpLogApplier`, and an export shape covering clients, segments, objects, and replicas.

  ## Module

  - [ ] CI/CD
  - [ ] Docs
  - [ ] Other

  ## Type of Change

  - [ ] Bug fix
  - [x] New feature
  - [ ] Refactor
  - [ ] Breaking change
  - [ ] Documentation update
  - [ ] Performance improvement
  - [ ] Other

  ## How Has This Been Tested?

  **Test commands:**
  ```bash
  ./scripts/code_format.sh --base P2P-Mooncake-Store
  cmake --build build-verify-noetcd --target p2p_oplog_test -j3
  cd build-verify-noetcd && ctest -R "^p2p_oplog_test$" --output-on-failure
```
  Test results:

  - [x] Unit tests pass

  p2p_oplog_test passed in the local verification container. The new p2p_hot_standby_service_test.cpp is included in the p2p_oplog_test target.

  ## Checklist

  ## AI Assistance Disclosure

  - [ ] No AI tools were used

  AI tools were used to help draft code changes, review diffs, and prepare PR text. The human submitter reviewed the changes and is responsible for them.

### DIFF
--- a/mooncake-store/src/ha/oplog/p2p_hot_standby_service.cpp
+++ b/mooncake-store/src/ha/oplog/p2p_hot_standby_service.cpp
@@ -1,0 +1,281 @@
+#include "ha/oplog/p2p_hot_standby_service.h"
+
+#include <glog/logging.h>
+
+#include <thread>
+#include <utility>
+#include <vector>
+
+namespace mooncake {
+
+P2PHotStandbyService::P2PHotStandbyService(P2PHotStandbyConfig config)
+    : config_(std::move(config)) {
+    metadata_store_ = std::make_unique<P2PStandbyMetadataStore>();
+    oplog_applier_ = std::make_unique<P2POpLogApplier>(metadata_store_.get(),
+                                                       config_.cluster_id);
+}
+
+P2PHotStandbyService::~P2PHotStandbyService() { Stop(); }
+
+ErrorCode P2PHotStandbyService::Start(uint64_t baseline_sequence_id) {
+    std::lock_guard<std::mutex> lock(mutex_);
+    if (state_machine_.IsRunning()) {
+        LOG(WARNING) << "P2PHotStandbyService is already running";
+        return ErrorCode::OK;
+    }
+
+    auto start_result = state_machine_.ProcessEvent(StandbyEvent::START);
+    if (!start_result.allowed) {
+        LOG(ERROR) << "P2PHotStandbyService: cannot start: "
+                   << start_result.reason;
+        return ErrorCode::INTERNAL_ERROR;
+    }
+
+    state_machine_.ProcessEvent(StandbyEvent::CONNECTED);
+    metadata_store_->RemoveAllMetadata();
+    oplog_applier_ = std::make_unique<P2POpLogApplier>(metadata_store_.get(),
+                                                       config_.cluster_id);
+    oplog_applier_->Recover(baseline_sequence_id);
+
+    auto err = StartOplogFollowingLocked(baseline_sequence_id);
+    if (err != ErrorCode::OK) {
+        state_machine_.ProcessEvent(StandbyEvent::FATAL_ERROR);
+        return err;
+    }
+
+    state_machine_.ProcessEvent(StandbyEvent::SYNC_COMPLETE);
+    LOG(INFO) << "P2PHotStandbyService started"
+              << ", cluster_id=" << config_.cluster_id
+              << ", baseline_sequence_id=" << baseline_sequence_id;
+    return ErrorCode::OK;
+}
+
+ErrorCode P2PHotStandbyService::StartOplogFollowingLocked(
+    uint64_t baseline_sequence_id) {
+    watcher_oplog_store_ = OpLogStoreFactory::Create(
+        config_.oplog_store_type, config_.cluster_id, OpLogStoreRole::READER,
+        config_.oplog_store_root_dir, config_.oplog_poll_interval_ms);
+    if (!watcher_oplog_store_) {
+        LOG(ERROR) << "P2PHotStandbyService: failed to create reader store"
+                   << ", cluster_id=" << config_.cluster_id;
+        return ErrorCode::INTERNAL_ERROR;
+    }
+
+    oplog_applier_->SetOpLogStore(watcher_oplog_store_.get());
+    oplog_change_notifier_ =
+        watcher_oplog_store_->CreateChangeNotifier(config_.cluster_id);
+    if (!oplog_change_notifier_) {
+        LOG(ERROR)
+            << "P2PHotStandbyService: failed to create OpLogChangeNotifier"
+            << ", cluster_id=" << config_.cluster_id;
+        return ErrorCode::INTERNAL_ERROR;
+    }
+
+    oplog_replicator_ = std::make_unique<OpLogReplicator>(
+        oplog_change_notifier_.get(), oplog_applier_.get());
+    oplog_replicator_->SetStateCallback(
+        [this](StandbyEvent event) { OnWatcherEvent(event); });
+
+    static constexpr int kMaxStartRetries = 3;
+    for (int attempt = 0; attempt < kMaxStartRetries; ++attempt) {
+        if (oplog_replicator_->StartFromSequenceId(baseline_sequence_id)) {
+            return ErrorCode::OK;
+        }
+        if (attempt + 1 < kMaxStartRetries) {
+            std::this_thread::sleep_for(
+                std::chrono::milliseconds(100 * (1 << attempt)));
+        }
+    }
+
+    LOG(ERROR) << "P2PHotStandbyService: failed to start OpLogReplicator";
+    return ErrorCode::INTERNAL_ERROR;
+}
+
+void P2PHotStandbyService::Stop() {
+    std::lock_guard<std::mutex> lock(mutex_);
+
+    if (oplog_replicator_) {
+        oplog_replicator_->Stop();
+        oplog_replicator_.reset();
+    }
+    oplog_change_notifier_.reset();
+    watcher_oplog_store_.reset();
+
+    StandbyState state = state_machine_.GetState();
+    if (state != StandbyState::STOPPED) {
+        auto result = state_machine_.ProcessEvent(StandbyEvent::STOP);
+        if (!result.allowed) {
+            LOG(WARNING) << "P2PHotStandbyService: stop transition rejected: "
+                         << result.reason;
+        }
+    }
+}
+
+ErrorCode P2PHotStandbyService::Promote() {
+    std::lock_guard<std::mutex> lock(mutex_);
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: mooncake-store/src/ha/oplog/p2p_hot_standby_service.cpp [hunk 1]
```diff
@@ -0,0 +1,281 @@
...
+}
+
+void P2PHotStandbyService::OnWatcherEvent(StandbyEvent event) {
+    auto result = state_machine_.ProcessEvent(event);
+    if (!result.allowed) {
+        VLOG(1) << "P2PHotStandbyService: watcher event rejected"
+                << ", event=" << StandbyEventToString(event)
+                << ", reason=" << result.reason;
+    }
+}
```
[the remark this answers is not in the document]
skygragon: Good point. The lower-level OpLogChangeNotifier already handles transient watch reconnects, but P2P still lacks a service-level recovery loop once the state machine enters RECONNECTING/RECOVERING. I added a TODO here and will handle the full recovery orchestration in the follow-up error-handling PR.

mooncake-store/src/ha/oplog/p2p_hot_standby_service.cpp
```diff
@@ -0,0 +1,281 @@
...
+    metadata_store_->RemoveAllMetadata();
+    oplog_applier_ = std::make_unique<P2POpLogApplier>(metadata_store_.get(),
+                                                       config_.cluster_id);
+    oplog_applier_->Recover(baseline_sequence_id);
+
+    auto err = StartOplogFollowingLocked(baseline_sequence_id);
+    if (err != ErrorCode::OK) {
+        state_machine_.ProcessEvent(StandbyEvent::FATAL_ERROR);
+        return err;
+    }
```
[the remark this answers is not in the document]
skygragon: Fixed in 6ca15422

mooncake-store/src/ha/oplog/p2p_hot_standby_service.cpp
```diff
@@ -0,0 +1,281 @@
...
+                  << ", attempted=" << gaps.attempted
+                  << ", fetched=" << gaps.fetched
+                  << ", applied_deletes=" << gaps.applied_deletes;
+    }
+
+    auto err = FinalCatchUpForPromotionLocked(applied_before_catch_up);
+    if (err != ErrorCode::OK) {
+        state_machine_.ProcessEvent(StandbyEvent::PROMOTION_FAILED);
+        return err;
+    }
```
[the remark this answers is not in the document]
skygragon: Fixed in 6ca15422

mooncake-store/src/ha/oplog/p2p_hot_standby_service.cpp
```diff
@@ -0,0 +1,281 @@
...
+    uint64_t expected = oplog_applier_->GetExpectedSequenceId();
+    return expected > 0 ? expected - 1 : 0;
+}
+
+P2PStandbyMetadataStore::ExportedMetadata P2PHotStandbyService::ExportMetadata()
+    const {
+    std::lock_guard<std::mutex> lock(mutex_);
+    return metadata_store_->ExportMetadata();
+}
```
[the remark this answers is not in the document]
skygragon: Fixed in 6ca15422

mooncake-store/src/ha/oplog/p2p_standby_metadata_store.cpp
```diff
@@ -292,10 +307,26 @@ P2PStandbyMetadataStore::ExportMetadata() const {
...
+std::shared_ptr<const P2PStandbyClientInfo> P2PStandbyMetadataStore::GetClient(
     const UUID& client_id) const {
+    std::lock_guard<std::mutex> lock(mutex_);
     auto it = clients_.find(client_id);
-    return it != clients_.end() ? &it->second : nullptr;
+    if (it == clients_.end()) {
+        return nullptr;
+    }
+    return std::make_shared<P2PStandbyClientInfo>(it->second);
+}
```
wanyue-wy, line 318: I notice that the values in `clients_` are not `shared_ptr`, so copying will be triggered here. Should we consider changing `clients_` to store `shared_ptr` instead?

mooncake-store/src/ha/oplog/p2p_hot_standby_service.cpp
```diff
@@ -0,0 +1,289 @@
...
+    oplog_applier_ = std::make_unique<P2POpLogApplier>(metadata_store_.get(),
+                                                       config_.cluster_id);
+    oplog_applier_->Recover(baseline_sequence_id);
+
+    auto err = StartOplogFollowingLocked(baseline_sequence_id);
+    if (err != ErrorCode::OK) {
+        state_machine_.ProcessEvent(StandbyEvent::FATAL_ERROR);
+        ResetOplogFollowingLocked();
+        return err;
+    }
```
wanyue-wy, line 48: we need a error log here

mooncake-store/src/ha/oplog/p2p_hot_standby_service.cpp
```diff
@@ -0,0 +1,289 @@
...
+                  << ", fetched=" << gaps.fetched
+                  << ", applied_deletes=" << gaps.applied_deletes;
+    }
+
+    auto err = FinalCatchUpForPromotionLocked(applied_before_catch_up);
+    if (err != ErrorCode::OK) {
+        state_machine_.ProcessEvent(StandbyEvent::PROMOTION_FAILED);
+        ResetOplogFollowingLocked();
+        return err;
+    }
```
wanyue-wy, line 165: we need a error log here

mooncake-store/src/ha/oplog/p2p_hot_standby_service.cpp
```diff
@@ -0,0 +1,289 @@
...
+
+    static constexpr int kMaxStartRetries = 3;
+    for (int attempt = 0; attempt < kMaxStartRetries; ++attempt) {
+        if (oplog_replicator_->StartFromSequenceId(baseline_sequence_id)) {
+            return ErrorCode::OK;
+        }
+        if (attempt + 1 < kMaxStartRetries) {
+            std::this_thread::sleep_for(
+                std::chrono::milliseconds(100 * (1 << attempt)));
+        }
```
wanyue-wy, line 93: Since `StartOplogFollowingLocked` holds the `mutex_` lock, we might want to evaluate whether the sleep here is reasonable.

mooncake-store/src/ha/oplog/p2p_hot_standby_service.cpp
```diff
@@ -0,0 +1,289 @@
...
+    return ErrorCode::INTERNAL_ERROR;
+}
+
+void P2PHotStandbyService::ResetOplogFollowingLocked() {
+    if (oplog_replicator_) {
+        oplog_replicator_->Stop();
+        oplog_replicator_.reset();
+    }
+    oplog_change_notifier_.reset();
+    watcher_oplog_store_.reset();
```
wanyue-wy, line 109: The `oplog_applier_` holds a raw pointer to `watcher_oplog_store_`. Once `watcher_oplog_store_` is reset, `oplog_applier_` will hold a dangling pointer.

mooncake-store/src/ha/oplog/p2p_hot_standby_service.cpp
```diff
@@ -0,0 +1,289 @@
...
+        read_from_seq = batch.back().sequence_id;
+    }
+
+    if (batch_count >= kMaxCatchUpBatches) {
+        // Do not block promotion solely on the bounded catch-up loop. The
+        // promoted primary continues from the best applied state in this phase.
+        LOG(WARNING) << "P2PHotStandbyService: final catch-up reached batch "
+                        "limit"
+                     << ", max_batches=" << kMaxCatchUpBatches;
+    }
```
wanyue-wy, line 227: From a design perspective, the cap on the number of logs to replay is intentional — it’s meant to prevent a scenario where no new master leader could promote successful. Thus, I think we should evaluate whether we have proper mitigations for two potential risks:
1. For a newly added standby node in the master cluster, can we guarantee that it will not be allowed to participate in elections until it has fully applied the snapshot?
2. With a large client cluster, can followers ensure that their log application rate does not significantly fall behind the log production rate?

If we don’t have solid answers for these, we might run into a situation where a new master promotes online with its actual log recovery position far behind the target, which could cause a substantial number of routing errors and routing lost.

mooncake-store/src/ha/oplog/p2p_hot_standby_service.cpp
skygragon: I think keeping the lock here is intentional: the retry loop owns the partially initialized watcher/notifier/replicator resources, and releasing the mutex between attempts would allow Stop/Promote to reset them concurrently. The backoff is bounded to 300ms total and only runs on the Start control path, so the temporary blocking is acceptable.

mooncake-store/src/ha/oplog/p2p_standby_metadata_store.cpp
skygragon: Good point. `clients_` is mutated by oplog replay (`REGISTER_CLIENT`, `MOUNT_SEGMENT`, `UNMOUNT_SEGMENT`, `UNREGISTER_CLIENT`). Returning an internal `shared_ptr` would only protect lifetime, not concurrent reads/writes after the store mutex is released. We would need extra locking or copy-on-write. So I’d keep this as a snapshot copy.

mooncake-store/src/ha/oplog/p2p_hot_standby_service.cpp
skygragon: Agreed. The bounded final catch-up is intentional to avoid blocking promotion forever, but it should not be the only safeguard for promotion correctness.

  The follow-up plan is to add promotion/election readiness gating: a standby should only be eligible after initial sync/ snapshot completion, its lag is below a configured threshold, and its apply rate/lag metrics show it can keep up with the primary. That should prevent a heavily lagging standby from being elected and relying only on this bounded catch-up loop.

  I added a TODO here to track this HA hardening work.

PR #2884, opened 2026-07-13
[Store] Make decode reconfigure remount make-before-break

SuperMarioYL:
## Description

The `/api/reconfigure` decode branch in `mooncake-wheel/mooncake/mooncake_store_service.py`
was **destroy-first**: it unmounted the currently serving segments *before* mounting the
new path.

```python
# before
if self.mounted_segment_ids:                        # unmount old first
    ret = self.store.unmount_segment(self.mounted_segment_ids)
    ...
    self.mounted_segment_ids.clear()
result = self.store.mount_segment(path, size, ...)  # then mount new
if result["ret"] != 0:
    self.current_mode = "prefill"                   # demoted to prefill on any failure
```

If the new `mount_segment` failed (a transient RDMA/registration hiccup, a bad path, an
out-of-memory device) *after* the old segments were already unmounted, the node was left
with **zero segments** and demoted to `prefill` — a self-inflicted availability outage, even
though the segments it had a moment ago were perfectly healthy.

This PR restructures the decode remount to **create-then-destroy (make-before-break)** for a
distinct target path: mount the new path first, and only retire the previous segments once
the new mount returns `ret == 0`. A failed remount now leaves the previous segments serving
and the node in `decode` mode instead of collapsing capacity to zero.

### Behavior after this change

| Scenario | Old (destroy-first) | New (make-before-break) |
| --- | --- | --- |
| Remount to a **different** path, new mount **fails** | unmount old → mount fails → **demoted to `prefill`, 0 segments** | old segments kept serving, stays in `decode` (HTTP 500 reports the retained state) |
| Remount to a **different** path, new mount **succeeds** | unmount old, then mount new (brief zero-capacity window in between) | mount new, then retire old; if that cleanup fails the new capacity is kept and the stale ids are warned/logged |
| Remount to the **same** path | destroy-first | destroy-first (unchanged, see below) |
| Fresh `prefill` → `decode`, mount fails (no prior segments) | rolled back to `prefill` | rolled back to `prefill` (unchanged) |

### Addressing the same-path collision (reviewer concern)

The old and new decode mounts can target the **same** segment path, and mounting before
unmounting there would hit `SEGMENT_ALREADY_EXISTS`. The inversion is therefore **scoped to
the distinct-path case**: `same_segment` compares the requested `path` against
`last_mount_info["path"]`, and when they match the code stays destroy-first exactly as
before. Only a remount to a *different* path takes the mount-then-unmount ordering, so
make-before-break never double-registers a path.

## Module

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

Added four unit tests in `mooncake-wheel/tests/test_mooncake_store_service_api.py` covering
each ordering path, and confirmed the two new behavioral tests are genuine reproductions
(red on `main`, green on this branch).

**Test commands:**
```bash
cd mooncake-wheel
python3 -m unittest tests.test_mooncake_store_service_api -v
```

**Test results:** `Ran 73 tests` — the four decode-remount tests below pass; the 4 unrelated
failures present are **pre-existing on `main`** (a `FakeStore.unmount_segment` tuple-vs-list
assertion mismatch and a `handle_put` 400/500 assertion) and are untouched by this change.

New / changed tests:
- `test_reconfigure_decode_remount_failure_keeps_previous_segments` — distinct-path remount
  whose mount fails keeps the previous segment ids, stays in `decode`, and unmounts nothing.
  **Red on `main`** (`AssertionError: 'prefill' != 'decode'`), green here.
- `test_reconfigure_decode_remount_success_is_make_before_break` — distinct-path remount that
  succeeds mounts the new segment *while the old one is still serving*, then retires the old
  one, leaving only the new segment active. **Red on `main`** (old segment already gone when
  the new mount runs), green here.
- `test_reconfigure_decode_same_path_remount_falls_back_to_prefill` — same-path remount stays
  destroy-first, so a mount failure still rolls back to `prefill`.
- `test_reconfigure_decode_mount_failure_rolls_back_to_prefill` — a fresh `prefill` → `decode`
  mount that fails (no prior segments) still rolls back to `prefill`.

- [x] Unit tests pass

## Checklist

- [x] I have formatted my code using `./scripts/code_format.sh` (`ruff check` and `ruff format`
      at the repo-pinned `v0.6.9`, plus `codespell`, are clean on the two touched files)
- [x] I have run the relevant `pre-commit` hooks on the changed files (`ruff`, `ruff-format`,
      `codespell` — the C/C++/CMake hooks do not apply to this Python-only change)
- [ ] I have updated the documentation (no user-facing docs affected)
- [ ] For changes >500 LOC: I have filed an RFC issue (diff is +155/−5)

## AI Assistance Disclosure

- [x] No AI tools were used

### DIFF
--- a/mooncake-wheel/mooncake/mooncake_store_service.py
+++ b/mooncake-wheel/mooncake/mooncake_store_service.py
@@ -231,12 +231,26 @@ class MooncakeStoreService:
                     )
 
                 async with self._state_lock:
-                    # If already in decode mode with mounted segments, unmount them first
-                    if self.mounted_segment_ids:
+                    # Make-before-break remount: capture the currently serving
+                    # segments so a failed remount can keep them alive instead of
+                    # dropping all capacity.
+                    previous_segment_ids = list(self.mounted_segment_ids)
+                    previous_path = self.last_mount_info.get("path")
+                    # Remounting the *same* path before unmounting would collide
+                    # (SEGMENT_ALREADY_EXISTS), so the same-path case must stay
+                    # destroy-first; a distinct path can mount-then-destroy.
+                    same_segment = (
+                        bool(previous_segment_ids)
+                        and previous_path is not None
+                        and path == previous_path
+                    )
+
+                    if same_segment:
                         logging.info(
-                            "Reconfigure decode: unmounting previous segments before remount"
+                            "Reconfigure decode: unmounting previous segments "
+                            "before same-path remount"
                         )
-                        ret = self.store.unmount_segment(self.mounted_segment_ids)
+                        ret = self.store.unmount_segment(previous_segment_ids)
                         if ret != 0:
                             return web.Response(
                                 status=500,
@@ -248,11 +262,35 @@ class MooncakeStoreService:
                                 content_type="application/json",
                             )
                         self.mounted_segment_ids.clear()
+                        previous_segment_ids = []
 
                     result = self.store.mount_segment(
                         path, size, offset, protocol, location
                     )
                     if result["ret"] != 0:
+                        if previous_segment_ids:
+                            # New mount failed but the previous segments are still
+                            # healthy; keep serving from them instead of demoting.
+                            logging.warning(
+                                "Reconfigure decode: mount of %s failed (ret=%s); "
+                                "keeping previous decode segments",
+                                path,
+                                result["ret"],
+                            )
+                            return web.Response(
+                                status=500,
+                                text=json.dumps(
+                                    {
+                                        "error": (
+                                            f"Mount failed, ret={result['ret']}; "
+                                            "keeping previous decode segments"
+                                        ),
+                                        "mode": self.current_mode,
+                                    }
+                                ),
+                                content_type="application/json",
+                            )
+                        # Nothing healthy to fall back to; roll back to prefill.
                         self.current_mode = "prefill"
                         self.mounted_segment_ids.clear()
                         self.last_mount_info.clear()
@@ -269,6 +307,19 @@ class MooncakeStoreService:
                             ),
                             content_type="application/json",
                         )
+
+                    # New mount succeeded; retire any previously serving segments.
+                    if previous_segment_ids:
+                        ret = self.store.unmount_segment(previous_segment_ids)
+                        if ret != 0:
+                            # The new segment is already live, so a failed cleanup
+                            # only leaks the old ids; warn but keep the new capacity.
+                            logging.warning(
+                                "Reconfigure decode: new mount succeeded but unmount of "
+                                "previous segments %s failed (ret=%s); leaking them",
+                                previous_segment_ids,
+                                ret,
+                            )
 
                     self.mounted_segment_ids = list(result["segment_ids"])
                     self.current_mode = "decode"
--- a/mooncake-wheel/mooncake/mooncake_store_service.py
+++ b/mooncake-wheel/mooncake/mooncake_store_service.py
@@ -235,43 +235,19 @@ class MooncakeStoreService:
                     # segments so a failed remount can keep them alive instead of
                     # dropping all capacity.
                     previous_segment_ids = list(self.mounted_segment_ids)
-                    previous_path = self.last_mount_info.get("path")
-                    # Same-path remount stays destroy-first. The standard
-                    # allocator wouldn't actually collide here: mountSegment
-                    # mints a fresh UUID per mount and the duplicate check is
-                    # UUID-keyed (segment.cpp:161), so the old "same path ->
-                    # SEGMENT_ALREADY_EXISTS" rationale was wrong for it. The
-                    # real restriction is the NoF path: segment.cpp:1249-1259
-                    # treats the same transport endpoint (te_endpoint) as the
-                    # same namespace even with a new UUID, so a same-endpoint
-                    # make-before-break WOULD collide. Destroy-first is kept to
-                    # stay correct across both allocators; a distinct path can
-                    # mount-then-destroy.
-                    same_segment = (
-                        bool(previous_segment_ids)
-                        and previous_path is not None
-                        and path == previous_path
-                    )
-
-                    if same_segment:
-                        logging.info(
-                            "Reconfigure decode: unmounting previous segments "
-                            "before same-path remount"
-                        )
-                        ret = self.store.unmount_segment(previous_segment_ids)
-                        if ret != 0:
-                            return web.Response(
-                                status=500,
-                                text=json.dumps(
-                                    {
-                                        "error": f"Unmount of previous segments failed, ret={ret}"
-                                    }
-                                ),
-                                content_type="application/json",
-                            )
-                        self.mounted_segment_ids.clear()
-                        previous_segment_ids = []
-
+                    # /api/reconfigure binds through store.mount_segment ->
+                    # MooncakeDistributedStore.mount_segment -> RealClient::mountSegment
+                    # -> Client::MountSegmentAndGetId -> MasterClient::MountSegment,
+                    # the standard allocator path: each mount mints a fresh UUID
+                    # and the duplicate check is UUID-keyed. That path never calls
+                    # MountNoFSegment or enters ScopedNoFSegmentAccess, so the NoF
+                    # te_endpoint dedup restriction does not apply here even in a
+                    # USE_NOF build. A same-path make-before-break therefore cannot
+                    # collide, and the old and new segments can coexist; keeping
+                    # the old segments live across a failed replacement mount
+                    # preserves decode capacity (the bug this PR fixes). Mount the
+                    # new segment first, and only retire the previous segments
+                    # per-id once the new mount succeeds.
                     result = self.store.mount_segment(
                         path, size, offset, protocol, location
                     )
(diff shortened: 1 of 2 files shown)


he-yufeng (commented):
Read the reconfigure path through. This is a real availability fix: the old destroy-first flow meant any failure to mount the new path dropped every currently-serving decode segment and demoted the node to prefill, so a transient mount error took the node fully out of decode. Make-before-break is the right shape, and the tests cover the two cases that matter (fresh prefill->decode mount failure still rolls back, and a different-path remount failure keeps the previous segments and stays in decode).

A few points, none blocking:

1. `same_segment` depends on `previous_path`. It's `bool(previous_segment_ids) and previous_path is not None and path == previous_path`. Is `last_mount_info["path"]` guaranteed to be present whenever `mounted_segment_ids` is non-empty? If a mounted state can ever exist with `previous_path` unset (e.g. a mount that populated `mounted_segment_ids` on a path that didn't flow through the success block that writes `last_mount_info`), then a genuine same-path remount would skip the destroy-first branch, hit `SEGMENT_ALREADY_EXISTS` on `mount_segment`, and fail. It fails safely (previous segments are kept), but a reconfigure that should have succeeded would error out. Worth asserting or documenting that `mounted_segment_ids` non-empty implies `last_mount_info["path"]` set.

2. The leaked-segment path (new mount succeeds, unmount of the old ids fails -> "leaking them") drops the old ids from tracking with only a warning. Is there anything that later reclaims or retries those? Under a backend that repeatedly fails unmount, this accumulates orphaned segments on the node with no handle to clean them up. A counter/metric for leaked segments (or a deferred-retry list) would at least let operators see it happening.

3. Make-before-break inherently double-mounts the same node's capacity for the window between the successful new mount and retiring the old segments. If the master's capacity accounting sees both at once, that brief double-count is probably fine, but worth confirming it doesn't trip any per-node capacity guard mid-reconfigure.

mooncake-wheel/mooncake/mooncake_store_service.py
```diff
@@ -270,6 +308,19 @@ async def handle_reconfigure(self, request):
...
+                            # The new segment is already live, so a failed cleanup
+                            # only leaks the old ids; warn but keep the new capacity.
+                            logging.warning(
+                                "Reconfigure decode: new mount succeeded but unmount of "
+                                "previous segments %s failed (ret=%s); leaking them",
+                                previous_segment_ids,
+                                ret,
+                            )
+
                     self.mounted_segment_ids = list(result["segment_ids"])
```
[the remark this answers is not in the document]
Lin-z-w: Agreed that the failed old segments must remain tracked. One additional detail: RealClient::unmountSegment processes the supplied IDs individually and returns the first error, so ret != 0 may mean that some IDs were successfully unmounted while others were not. Keeping all previous_segment_ids, as suggested here, would therefore retain stale IDs for segments that were already removed.
We should either unmount the old IDs individually, as handle_unmount_shm already does, or change the backend API to return per-ID results. mounted_segment_ids should then contain the new IDs plus only the old IDs whose cleanup actually failed. A partial-cleanup-failure test would be useful as well.

mooncake-wheel/mooncake/mooncake_store_service.py
```diff
@@ -231,12 +231,26 @@ async def handle_reconfigure(self, request):
...
-                    # If already in decode mode with mounted segments, unmount them first
-                    if self.mounted_segment_ids:
+                    # Make-before-break remount: capture the currently serving
+                    # segments so a failed remount can keep them alive instead of
+                    # dropping all capacity.
+                    previous_segment_ids = list(self.mounted_segment_ids)
+                    previous_path = self.last_mount_info.get("path")
+                    # Remounting the *same* path before unmounting would collide
+                    # (SEGMENT_ALREADY_EXISTS), so the same-path case must stay
+                    # destroy-first; a distinct path can mount-then-destroy.
```
Lin-z-w: The comment assumes that mounting the same path while the old segment is active necessarily returns SEGMENT_ALREADY_EXISTS, but that does not match the current backend semantics. RealClient::mountSegment creates a new mmap, Client::MountSegmentAndGetId generates a new UUID, and the master’s duplicate check is based on an existing segment UUID rather than the backing file path. The segment tests also explicitly permit the same name with different IDs.
As a result, same-path remounts should also attempt make-before-break unless a real-backend test demonstrates another concrete restriction. With this branch, a transient same-path mount failure still occurs after the healthy old segments have been removed, causing the same decode-capacity outage this PR is intended to fix.

mooncake-wheel/mooncake/mooncake_store_service.py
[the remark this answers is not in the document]
SuperMarioYL: Good catch — done in b858c982. The retire step now unmounts each previous id individually (mirroring `handle_unmount_shm`) and rebuilds `mounted_segment_ids` as the new ids plus only the ids whose `unmount_segment` actually returned non-zero. So a partial cleanup failure no longer keeps stale ids for segments that were already freed, and it also stops dropping the still-live ones (which addresses @he-yufeng's orphaned-segment concern — the un-freed ids stay tracked and can be retried on a later unmount or a switch back to prefill). Added `test_reconfigure_decode_partial_unmount_failure_keeps_only_failed_ids` for the mixed success/failure case.

mooncake-wheel/mooncake/mooncake_store_service.py
SuperMarioYL: You're right about the standard allocator — `MountSegmentAndGetId` generates a fresh UUID (`client_service.cpp:2817`) and the duplicate check is on the UUID (`segment.cpp:161`), so a same-path remount there would not hit `SEGMENT_ALREADY_EXISTS`. The old comment was wrong about that, and I have corrected it in 50f971a7 to state the real restriction.

The one concrete restriction I found is the NoF path: `segment.cpp:1249-1259` treats the same transport endpoint (`te_endpoint`) as the same namespace and returns `SEGMENT_ALREADY_EXISTS` even for a new UUID, so a same-endpoint make-before-break would collide there. That is why I kept same-path destroy-first — to stay correct across both allocators. Happy to make it standard-allocator-only make-before-break (with destroy-first retained for NoF / same-endpoint) if you would prefer, ideally behind a real-backend NoF test that proves the endpoint-dedup path; I cannot drive that one headless on macOS. What is your call?

Lin-z-w: Thanks for tracing the NoF endpoint-dedup behavior. However, I do not think that restriction applies to this API path.

`/api/reconfigure` calls `MooncakeDistributedStore.mount_segment`, whose binding invokes `RealClient::mountSegment`. That method calls `Client::MountSegmentAndGetId`, constructs a standard `Segment`, and ultimately uses `MasterClient::MountSegment`. It does not call `MountNoFSegment` or enter `ScopedNoFSegmentAccess`, even in a `USE_NOF` build.

NoF registration uses a separate path through `MooncakeDistributedNoFRegister` / `NoFRegisterClient::set_register` and `MasterClient::MountNoFSegment`. Therefore, the NoF `te_endpoint` deduplication does not require this handler to keep same-path remounts destroy-first.

Please use make-before-break for same-path remounts in this handler as well and update the same-path failure test to verify that the previous decode segments remain available when the replacement mount fails. If `/api/reconfigure` is intended to support NoF registration in the future, that integration should be added and tested separately.

SuperMarioYL: Thanks for tracing the binding — that settles it. Done in 53d3f2b6:

- Same-path remounts now go through the same make-before-break path as distinct-path remounts: mount the new segment first, keep serving the previous segments if the new mount fails, and retire the previous segments per-id only once the new mount succeeds. The same-path destroy-first special case is gone (mooncake_store_service.py handle_reconfigure), and the stale NoF comment is corrected to state the real reason — /api/reconfigure binds through MasterClient::MountSegment (fresh-UUID, no NoF te_endpoint dedup).
- The same-path failure test is rewritten to verify your ask: when a same-path replacement mount fails, the previous decode segments remain available — mode stays decode, mounted_segment_ids keeps the old id, nothing is unmounted, and last_mount_info still points at the working path (test_reconfigure_decode_same_path_remount_failure_keeps_previous_segments).
- Left NoF registration out of this handler, as you noted — if /api/reconfigure grows NoF support later it should land + test separately.

SuperMarioYL:
Hi — could a maintainer re-run the failed `test-wheel-ubuntu (ubuntu-22.04, 3.10)` job (run [29573973179](https://github.com/kvcache-ai/Mooncake/actions/runs/29573973179))? I don't have rerun permission on this repo.

The only failing step is `Run tests with ssd`, and the actual error is a master-startup failure, not a test-logic failure:

```
Test that put operations fail when local_buffer_size is set to zero. ...
❌ mooncake_master process failed to start
E... http_metadata_server.cpp:121] Failed to start HTTP metadata server on 0.0.0.0:8080
F... master.cpp:1449] Failed to start HTTP metadata server
*** Check failure stack trace: ***
```

i.e. the C++ `mooncake_master` process couldn't bind port `8080` because a prior test's master hadn't released it — a port-conflict flake in the C++ test harness, in code this PR doesn't touch.

This PR is Python-only (`mooncake_store_service.py` + unit tests in `test_mooncake_store_service_api.py` that use a mocked store, no real master). It doesn't touch `master.cpp` / `http_metadata_server.cpp` / the metadata-server port handling, so it can't be responsible for the 8080 bind conflict. The other three `test-wheel-ubuntu` matrix siblings were cancelled by fail-fast, and `CI Gate` failed only because that required job failed.

Happy to address anything else if a rerun still surfaces a real failure on this diff.

Lin-z-w (approved):
LGTM.

PR #2891, opened 2026-07-14
[TENT] Reuse and release SHM relocation mappings across threads

### LINKED ISSUE #2888: ShmTransport duplicates and leaks relocation mappings across threads
## Problem

`ShmTransport::relocateSharedMemoryAddress` stores newly opened mappings in a function-local thread-local map. The transport member `relocate_map_` is copied into that cache but is never updated when a new mapping is created.

As a result:

- every submission thread can independently open and map the same shared-memory buffer;
- `uninstall()` cannot see those mappings because it inspects only the member map;
- file descriptors and mappings remain allocated;
- thread-local mappings can outlive transport uninstall or replacement.

## Evidence

The current source writes new mappings through a reference to the thread-local map. The member map remains empty after relocation, while `uninstall()` iterates only the member map.

## Expected behavior

- Concurrent threads should reuse one relocation mapping for the same target buffer.
- Mapping ownership should belong to the transport instance.
- File descriptors should close after successful `mmap`.
- `uninstall()` should `munmap` every transport-owned mapping.
- No thread-local ownership should survive transport replacement.

## Impact

Sustained use from multiple submission threads causes redundant mappings and process-resource leaks.

morluto:
Fixes #2888

## Description

`ShmTransport::relocateSharedMemoryAddress` stored newly opened mappings in a function-local thread-local map. The transport member map was copied into that cache but never updated when new mappings were created.

Consequently, each submission thread could independently open and map the same shared-memory buffer, while `uninstall()` could not see or release those mappings.

This change makes the transport member map the single mapping owner:

- lookups use the existing read lock;
- misses acquire the write lock and recheck before creating a mapping;
- both lookup paths use the same address-resolution helper;
- failed metadata, open, and mapping operations do not publish empty cache buckets;
- relocation rejects calls after `uninstall()` while holding the owner lock;
- concurrent threads reuse the mapping published by the first thread;
- file descriptors close immediately after a successful `mmap`;
- `uninstall()` unmaps every published region under the owner lock;
- no relocation ownership remains in thread-local state.

A regression test synchronizes eight threads relocating the same buffer and checks that they receive one shared mapping owned by the transport. It also checks that a failed lookup does not create a target cache entry, `uninstall()` releases the mapping, and subsequent relocation is rejected.

## Module

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

**Validation commands:**

```bash
pre-commit run --files $(git diff --name-only origin/main..HEAD)
git diff --check origin/main..HEAD
```

**Results:**

- [x] All configured pre-commit hooks pass on the touched files
- [x] clang-format 20 passes
- [x] CMake formatting passes
- [x] Regression coverage was added
- [x] Structured branch review reports no actionable findings

## Checklist

- [x] I have run pre-commit on all files touched by this change, and all hooks pass
- [x] Documentation is not applicable because the public API is unchanged
- [x] I have added tests to prove the change is effective
- [x] An RFC is not required because the change is under 500 LOC

## AI Assistance Disclosure

- [x] AI tools were used

Codex assisted with source tracing, mapping-lifetime research, implementation, regression-test design, formatting, and structured review. The submitter is responsible for reviewing and understanding every changed line.

### DIFF
--- a/mooncake-transfer-engine/tent/src/transport/shm/shm_transport.cpp
+++ b/mooncake-transfer-engine/tent/src/transport/shm/shm_transport.cpp
@@ -60,11 +60,11 @@ Status ShmTransport::install(std::string &local_segment_name,
 
 Status ShmTransport::uninstall() {
     if (installed_) {
+        RWSpinlock::WriteGuard guard(relocate_lock_);
         metadata_.reset();
         for (auto &relocate_map : relocate_map_) {
             for (auto &entry : relocate_map.second) {
                 munmap(entry.second.shm_addr, entry.second.length);
-                close(entry.second.shm_fd);
             }
         }
         relocate_map_.clear();
@@ -244,13 +244,26 @@ void *ShmTransport::createSharedMemory(const std::string &path, size_t size) {
 Status ShmTransport::relocateSharedMemoryAddress(uint64_t &dest_addr,
                                                  uint64_t length,
                                                  uint64_t target_id) {
-    thread_local HashMap tl_relocate_map;
-    if (tl_relocate_map.empty()) {
+    {
         RWSpinlock::ReadGuard guard(relocate_lock_);
-        tl_relocate_map = relocate_map_;
+        auto target = relocate_map_.find(target_id);
+        if (target != relocate_map_.end()) {
+            for (auto &entry : target->second) {
+                if (entry.first <= dest_addr &&
+                    dest_addr + length <= entry.first + entry.second.length) {
+                    auto shm_addr = entry.second.shm_addr;
+                    dest_addr = dest_addr - entry.first + ((uint64_t)shm_addr);
+                    return Status::OK();
+                }
+            }
+        }
     }
 
-    auto &relocate_map = tl_relocate_map[target_id];
+    RWSpinlock::WriteGuard guard(relocate_lock_);
+    auto &relocate_map = relocate_map_[target_id];
+
+    // Another thread may have published this mapping while the writer lock was
+    // pending. Recheck before opening and mapping the same shared-memory file.
     for (auto &entry : relocate_map) {
         if (entry.first <= dest_addr &&
             dest_addr + length <= entry.first + entry.second.length) {
@@ -259,8 +272,6 @@ Status ShmTransport::relocateSharedMemoryAddress(uint64_t &dest_addr,
             return Status::OK();
         }
     }
-
-    RWSpinlock::WriteGuard guard(relocate_lock_);
 
     BufferDesc *buffer;
     // Owning reference: `buffer` is used after the lambda returns.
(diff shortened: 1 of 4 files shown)


mooncake-transfer-engine/tent/src/transport/shm/shm_transport.cpp
```diff
@@ -244,13 +244,26 @@ void *ShmTransport::createSharedMemory(const std::string &path, size_t size) {
...
+                    dest_addr = dest_addr - entry.first + ((uint64_t)shm_addr);
+                    return Status::OK();
+                }
+            }
+        }
     }
 
-    auto &relocate_map = tl_relocate_map[target_id];
+    RWSpinlock::WriteGuard guard(relocate_lock_);
+    auto &relocate_map = relocate_map_[target_id];
```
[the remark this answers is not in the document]
morluto: Fixed in e9422945: relocation now checks `metadata_` immediately after acquiring the writer lock and returns `InvalidArgument` after uninstall. The regression test now covers that lifecycle boundary.

mooncake-transfer-engine/tent/src/transport/shm/shm_transport.cpp
```diff
@@ -244,13 +244,30 @@ void *ShmTransport::createSharedMemory(const std::string &path, size_t size) {
...
+        auto target = relocate_map_.find(target_id);
+        if (target != relocate_map_.end()) {
+            for (auto &entry : target->second) {
+                if (entry.first <= dest_addr &&
+                    dest_addr + length <= entry.first + entry.second.length) {
+                    auto shm_addr = entry.second.shm_addr;
+                    dest_addr = dest_addr - entry.first + ((uint64_t)shm_addr);
+                    return Status::OK();
+                }
+            }
```
staryxchen: Suggestion: The read-path lookup and the write-path recheck have byte-identical bodies (same entry.first <= dest_addr && dest_addr + length <= entry.first + entry.second.length check and address arithmetic). The recheck is intentional and the comment explains why, but a small tryResolve(relocate_map, dest_addr, length) -> bool helper would remove the duplication and keep the two paths provably consistent.

mooncake-transfer-engine/tent/src/transport/shm/shm_transport.cpp
```diff
@@ -244,13 +244,30 @@ void *ShmTransport::createSharedMemory(const std::string &path, size_t size) {
...
+        }
+    }
+
+    RWSpinlock::WriteGuard guard(relocate_lock_);
+    if (!metadata_) {
+        return Status::InvalidArgument(
+            "SHM transport is not installed" LOC_MARK);
     }
+    auto &relocate_map = relocate_map_[target_id];
```
staryxchen: This runs before withCachedSegment / shm_open / mmap. If any of those fail (e.g. NeedsRefreshCache from findBuffer), the function returns but leaves an empty inner map for target_id in relocate_map_ until uninstall(). Benign for correctness (the read-path loop over an empty map is a no-op), but it forces every subsequent call for that target_id to fall through to the write lock. Consider deferring the operator[] until after the mapping is confirmed, or eraseing the entry on the error path.

morluto: Fixed in 2edd6e0a: the outer `relocate_map_` entry is now created only after a mapping succeeds. Metadata lookup, unsupported-location, open, and mmap failures leave no target bucket; the regression test asserts this for a failed buffer lookup.

mooncake-transfer-engine/tent/src/transport/shm/shm_transport.cpp
morluto: Fixed in 2edd6e0a: both the read-lock fast path and writer-lock recheck now call the same `tryResolve` helper, so the containment test and address translation have one implementation.

staryxchen (approved):
LGTM

Issue #2706, opened 2026-07-01
[RFC]: Consolidate Master Metadata Routing


Aionw:
### Changes proposed

# RFC: Consolidate Master Metadata Routing

## Summary

`MasterService` should keep the current tenant-isolated metadata layout:

```text
metadata_shards_[shard].tenants[tenant_id].metadata[user_key]
```

This preserves the strong tenant boundary: each tenant owns an independent `TenantState`, and duplicate `user_key` values across tenants are naturally isolated.

The proposed change is not to flatten tenant storage. Instead, it is to consolidate routing and per-object access behind a canonical object identity:

```text
ObjectIdentity{tenant_id, user_key}
  -> MetadataRoute{shard, tenant_state, object_state}
  -> ObjectState{metadata, processing state, tasks, group state}
```

The goal is to get most of the simplicity of a flat metadata model without weakening tenant isolation. Object operations should use one routing/accessor path and then read or mutate object-scoped state from one colocated object entry.

## Motivation

The current lookup path is `tenant_id + user_key -> shard -> TenantState -> metadata[user_key]`. This tenant layer is useful and should remain, but object-scoped state is spread across several tenant-local maps such as `metadata`, `processing_keys`, `replication_tasks`, `offloading_tasks`, `promotion_tasks`, and `group_members`.

That means a single object operation can repeat the same `user_key` lookup across multiple maps after it has already found the shard and tenant. It also makes routing logic easy to duplicate across call sites.

The master already has `ObjectIdentity` and `MetadataAccessorRO/RW`. This RFC proposes making that route/accessor layer the single way to reach object metadata and related state, instead of open-coding the route at individual call sites.

## Goals

- Preserve `MetadataShard::tenants[tenant_id]` as the isolation boundary.
- Use `ObjectIdentity{tenant_id, user_key}` as the canonical input to routing and accessors.
- Consolidate shard selection, tenant lookup, tenant creation, and per-object lookup in one route/accessor path.
- Colocate object-scoped state inside a per-object `ObjectState` so common operations avoid several tenant-local map lookups.
- Keep public APIs compatible: callers may continue passing `tenant_id` and `user_key` separately.
- Preserve single-tenant mode by normalizing requests to the `default` tenant when multi-tenant mode is disabled.
- Keep snapshot compatibility with the existing tenant-nested shape.

## Non-Goals

- Do not flatten `MetadataShard::tenants`.
- Do not remove tenant semantics from Mooncake Store.
[...]

PR #2894, opened 2026-07-14
[Store] Introduce canonical TenantId

Aionw:
## Description

Introduces a canonical `TenantId` value type for Mooncake Store internals.

- Centralizes tenant normalization, validation, default-tenant detection, hashing, and scoped-key encoding.
- Converts raw tenant strings once at public/RPC/admin/task boundaries; strict write callers reject empty or invalid values before invoking `MasterService`.
- Changes tenant-bearing `MasterService` interfaces to accept a valid `TenantId` and uses assertions as a weak boundary-contract check; tenant registration remains a `MasterService` responsibility.
- Preserves existing request semantics: single-tenant mode always resolves to `default`, while multi-tenant read paths continue to map an empty tenant to `default`.
- Migrates MasterService metadata/quota state, MasterClient state, storage, promotion/offload, drain, and KV event internals to `TenantId`.
- Keeps public APIs, RPC/YLT payloads, configuration, admin interfaces, quota policies, and snapshots string-compatible.
- Preserves the existing `tenant + '\0' + local_key` format and shard hashing behavior.
- Avoids redundant tenant string copies on write paths and repeated construction in batch offload reads.

This is the first implementation step for the metadata routing consolidation RFC. It does not change `ObjectState`, metadata layout, routing algorithms, or wire formats.

Related to #2706.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [x] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
cmake --build /tmp/mooncake-tenantid-build --target tenant_id_test master_service_test master_service_tenant_quota_test offload_on_evict_test file_storage_promotion_test task_manager_test promotion_on_hit_test snapshot_child_process_test -j4

/tmp/mooncake-tenantid-build/mooncake-store/tests/master_service_test \
  --gtest_filter='*Tenant*:*tenant*:*Scoped*:*LegacyTaskPayloadDefaultsTenant'

ctest --test-dir /tmp/mooncake-tenantid-build --output-on-failure \
  -R '^(tenant_id_test|master_service_tenant_quota_test|offload_on_evict_test|file_storage_promotion_test|task_manager_test|promotion_on_hit_test|snapshot_child_process_test)$'
```

**Test results:**
- [x] Unit tests pass

Additional validation:
- Tenant-related MasterService tests: 13/13 passed.
- Promotion-on-hit and snapshot child-process round trips passed.
- Touched-file clang-format, codespell, whitespace, EOF, merge-conflict, and large-file checks passed.
- The repository-wide formatter scans unrelated branch divergence and attempted to rewrite unrelated files; those rewrites were excluded from this PR.

## Checklist

- [x] I have updated the documentation (not applicable: no user-facing behavior changed)
- [x] For changes >500 LOC: I have filed an RFC issue (#2706)

## AI Assistance Disclosure

- [ ] No AI tools were used

OpenAI Codex was used to implement the refactor, add and update tests, run validation, review the diff, and prepare this pull request. The human submitter is responsible for reviewing every changed line and understanding the change end-to-end.


he-yufeng (approved):
Focused my review on the compat-critical core, since that's where a tenant-encoding change is riskiest, and it holds up.

**Scoped-key bytes are preserved.** The risk with any tenant-encoding change is that the default tenant's scoped key shifts and existing keys re-hash to a different shard. It doesn't: the old `MakeTenantScopedStorageKey` already normalized empty to `"default"` (via `NormalizeTenantIdRef`) before appending `'\0' + key`, and `TenantId::MakeScopedKey` emits the same bytes. I pulled both encoders into a standalone build and diffed {empty, "default", named} x {key, path, empty}; all pairs are byte-identical, including the `""`/`"default"` collision (both produce `default\0key`). Shard hashing and the on-disk/RPC key layout are unchanged.

**No validation dropped.** The four `IsValidTenantId` ingress checks (`master_admin_service.cpp:234`, `master_service.cpp:683/714/1028`, `tenant_quota_policy_store.cpp`) migrate 1:1 to `TenantId::IsValid()`, with the same rules (reject empty, `_`-prefix, control and DEL). The `IsValid()` calls in the read paths are the pre-existing `ObjectMetadata` replica checks, untouched (upstream had 24, this PR 27, and the three added are exactly the tenant-validation ones).

**The copy-avoidance claim checks out.** Tenant is stored once as a `TenantId` (in `ObjectIdentity`, and as the key type of the metadata/quota `unordered_map<TenantId, ..., TenantIdHash>`), and `getMetadataShardIndex` / `getShardIndex` / `getTenantQuotaShardIndex` all take `const TenantId&`. Normalization now happens once at construction rather than per shard-index call, which is a small win over the old `NormalizeTenantIdRef`.

Two non-blocking thoughts for later steps:

- The constructor normalizes but doesn't validate, so an invalid `TenantId` can exist until one of the four ingress sites rejects it. That matches the old `NormalizeTenantId`/`IsValidTenantId` split so it's not a regression, but a `TryCreate() -> expected<TenantId, ErrorCode>` in a follow-up would let the type carry the "valid" invariant end to end.
- `ParseScopedKey` splits on the first `'\0'`, which stays unambiguous only because `IsValid()` forbids control chars so a well-formed tenant never contains one. Worth a one-line comment noting that dependency, since the parser itself doesn't enforce it.

Clean, well-scoped first cut for #2706: the string boundary is kept and the RFC-relevant state is now typed. LGTM.

mooncake-store/src/rpc_service.cpp
```diff
@@ -11,6 +16,44 @@
...
 namespace mooncake {
+namespace {
+
+TenantId ResolveRequestTenantId(std::string_view raw,
+                                bool enable_multi_tenants) {
+    if (!enable_multi_tenants) {
+        return TenantId::Default();
+    }
+    TenantId tenant_id{std::string(raw)};
+    assert(tenant_id.IsValid());
```
Lin-z-w: `ResolveRequestTenantId()` asserts on an untrusted RPC tenant ID. An invalid value can terminate assertion-enabled builds, while Release builds skip validation entirely. Please validate it at the RPC boundary and return an error instead.

Lin-z-w (approved):
LGTM.

ykwd:
@Aionw Could you resolve the conflict?

PR #2895, opened 2026-07-14
[TE] Enforce QP teardown before MR dereg on shutdown

alogfans:
## Description

This change enforces QP teardown before memory regions are deregistered during RDMA shutdown. Without this ordering, shutdown can deregister MRs while QPs are still alive, leaving outstanding RDMA resources referencing memory that is already being torn down.

Changes included:

- Drain/destroy endpoint QPs before MR deregistration.
- Make RDMA context shutdown explicitly tear down endpoint-store QPs before unregistering memory regions.
- Avoid shutdown ordering races between endpoint/QP lifetime and MR lifetime.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
```

**Test results:**
- [X] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used
- [X] AI tools were used (specify below)

CodeX

### DIFF
--- a/mooncake-transfer-engine/src/transport/rdma_transport/endpoint_store.cpp
+++ b/mooncake-transfer-engine/src/transport/rdma_transport/endpoint_store.cpp
@@ -142,10 +142,24 @@ void FIFOEndpointStore::reclaimEndpoint() {
 size_t FIFOEndpointStore::getSize() { return endpoint_map_.size(); }
 
 int FIFOEndpointStore::destroyQPs() {
-    for (auto &kv : endpoint_map_) {
-        kv.second->destroyQP();
+    RWSpinlock::WriteGuard guard(endpoint_map_lock_);
+    int ret = 0;
+
+    // Always transition QPs to ERR before destroy to flush inflight WRs.
+    for (auto &endpoint : waiting_list_) {
+        endpoint->beginDestroy();
     }
-    return 0;
+    for (auto &kv : endpoint_map_) {
+        kv.second->beginDestroy();
+    }
+
+    for (auto &endpoint : waiting_list_) {
+        if (endpoint->destroyQP()) ret = -1;
+    }
+    for (auto &kv : endpoint_map_) {
+        if (kv.second->destroyQP()) ret = -1;
+    }
+    return ret;
 }
 
 int FIFOEndpointStore::disconnectQPs() {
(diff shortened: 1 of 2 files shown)


he-yufeng (approved):
Read the teardown path end to end; the ordering is right and the pieces line up.

**The fix is the correct RDMA lifecycle order.** `ibv_dereg_mr` on an MR still referenced by a live QP is undefined behavior (and at best fails `EBUSY`), so the QPs have to go first. `deconstruct()` now does disconnect -> drain the waiting list -> `destroyQPs()` -> then the `ibv_dereg_mr` loop, which is the right sequence. The `destroyQPs()` rewrite also does the two-phase teardown properly: `beginDestroy()` on every endpoint first (phase 1, QP to ERR so inflight WRs flush) and only then `destroyQP()`, rather than tearing each QP down individually while its siblings are still live.

**It compiles and is lock-safe** — I checked the parts that aren't in the two-file diff. `beginDestroy()` and `waitingListSize()` are existing methods (`rdma_endpoint.{h,cpp}`, `endpoint_store.h`), so this uses the documented two-phase API rather than introducing an undefined call. And the new `WriteGuard(endpoint_map_lock_)` in `destroyQPs()` doesn't nest: `deconstruct()` calls `disconnectQPs()`, `reclaimEndpoint()`, and `destroyQPs()` sequentially and holds no lock itself, and each of those takes and releases the guard on its own, so there is no double-acquire on the non-recursive spinlock. `beginDestroy()` locks the endpoint's own state, not the store map, so calling it under the map guard is fine as well.

Two small, non-blocking notes:

- FIFO and SIEVE `destroyQPs()` are now near-identical two-phase loops; if a third store ever appears it would be worth a shared helper, but with two it reads fine as is.
- The 10s reclaim timeout then force-destroy is a sensible fallback; a one-line comment on why 10s (relative to the WR drain deadline) would help the next reader, though it is not blocking.

Focused fix for a real shutdown-order hazard, and the invariant it protects (no live QP over a deregistered MR) is exactly the kind that only bites intermittently in the field. LGTM.

PR #2897, opened 2026-07-14
[TE] Use std::atomic to support ARM64 relaxed ordering

alogfans:
## Description

This change replaces several TE/TENT synchronization paths that relied on `volatile`, `__sync_*` builtins, or mixed-width atomic access with `std::atomic` operations.

The main fixes are:
- Convert TE/TENT `RWSpinlock` ticket state to a single `std::atomic<uint64_t>`, so lock acquisition/release no longer mixes 64-bit CAS with 16-bit subfield atomics.
- Convert TE RDMA CQ/QP outstanding counters from `volatile int` plus `__sync_*` to `std::atomic<int>`.
- Convert TE RDMA endpoint/context lifecycle flags (`active_`, `inactive_time_`) to atomics.
- Convert TENT RDMA quota counters (`cqe_now_`, QP depth, inflight slices) to atomics.

This avoids relying on x86-style ordering assumptions and makes the affected hot paths valid under the C++ memory model on ARM64.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
```

**Test results:**
- [X] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

CodeX

### DIFF
--- a/mooncake-transfer-engine/include/common.h
+++ b/mooncake-transfer-engine/include/common.h
@@ -555,18 +555,19 @@ class RWSpinlock {
         if (t.users != t.write) return false;
         ++t.users;
         return ticket.compare_exchange_weak(expected.whole, t.whole,
-                                             std::memory_order_acquire);
+                                            std::memory_order_acquire);
     }
 
     void writeLockAggressive() {
         uint32_t count = 0;
         RWTicket increment;
         increment.users = 1;
-        RWTicket old(ticket.fetch_add(increment.whole,
-                                      std::memory_order_acquire));
+        RWTicket old(
+            ticket.fetch_add(increment.whole, std::memory_order_acquire));
         uint16_t val = old.users;
         RWTicket t;
-        while (val != (t.whole = ticket.load(std::memory_order_acquire), t.write)) {
+        while (val !=
+               (t.whole = ticket.load(std::memory_order_acquire), t.write)) {
             PAUSE();
             if (++count > 1000) std::this_thread::yield();
         }
@@ -594,9 +595,9 @@ class RWSpinlock {
             ++t.read;
             ++t.write;
             new_val = t.whole;
-        } while (!ticket.compare_exchange_weak(
-            expected, new_val, std::memory_order_release,
-            std::memory_order_relaxed));
+        } while (!ticket.compare_exchange_weak(expected, new_val,
+                                               std::memory_order_release,
+                                               std::memory_order_relaxed));
     }
 
     void lockShared() {
@@ -615,7 +616,7 @@ class RWSpinlock {
         ++t.read;
         ++t.users;
         return ticket.compare_exchange_weak(expected.whole, t.whole,
-                                             std::memory_order_acquire);
+                                            std::memory_order_acquire);
     }
 
     void unlockShared() { fetch_add_write(1); }
@@ -630,7 +631,7 @@ class RWSpinlock {
             t.read += delta;
             new_val = t.whole;
         } while (!ticket.compare_exchange_weak(expected, new_val,
-                                                std::memory_order_acquire));
+                                               std::memory_order_acquire));
         return static_cast<uint16_t>(t.read - delta);
     }
 
@@ -643,7 +644,7 @@ class RWSpinlock {
             t.write += delta;
             new_val = t.whole;
         } while (!ticket.compare_exchange_weak(expected, new_val,
-                                                std::memory_order_release));
+                                               std::memory_order_release));
         return static_cast<uint16_t>(t.write - delta);
     }
 
--- a/mooncake-transfer-engine/include/common.h
+++ b/mooncake-transfer-engine/include/common.h
@@ -586,11 +586,17 @@ class RWSpinlock {
     }
 
     void unlock() {
+        uint64_t expected = ticket.load(std::memory_order_relaxed);
+        uint64_t new_val;
         RWTicket t;
-        t.whole = ticket.load(std::memory_order_relaxed);
-        ++t.read;
-        ++t.write;
-        ticket.store(t.whole, std::memory_order_release);
+        do {
+            t.whole = expected;
+            ++t.read;
+            ++t.write;
+            new_val = t.whole;
+        } while (!ticket.compare_exchange_weak(
+            expected, new_val, std::memory_order_release,
+            std::memory_order_relaxed));
     }
 
     void lockShared() {
--- a/mooncake-transfer-engine/include/common.h
+++ b/mooncake-transfer-engine/include/common.h
@@ -528,6 +528,7 @@ static inline bool overlap(const void *a, size_t a_len, const void *b,
 class RWSpinlock {
     union RWTicket {
         constexpr RWTicket() : whole(0) {}
+        constexpr RWTicket(uint64_t v) : whole(v) {}
         uint64_t whole;
         uint32_t readWrite;
         struct {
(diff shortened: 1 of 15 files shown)


staryxchen:
Maybe we can add a job to run UT on arm64 machine

mooncake-transfer-engine/include/common.h
```diff
@@ -587,16 +581,16 @@ class RWSpinlock {
...
 
     void unlock() {
         RWTicket t;
-        t.whole = load_acquire(&ticket.whole);
+        t.whole = ticket.load(std::memory_order_relaxed);
         ++t.read;
         ++t.write;
-        store_release(&ticket.readWrite, t.readWrite);
+        ticket.store(t.whole, std::memory_order_release);
     }
```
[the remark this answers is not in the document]
alogfans: Fixed. We use CAS instead of store in unlock().

mooncake-transfer-engine/tent/include/tent/common/concurrent/rw_spinlock.h
```diff
@@ -85,16 +80,16 @@ class RWSpinlock {
...
 
     void unlock() {
         RWTicket t;
-        t.whole = load_acquire(&ticket.whole);
+        t.whole = ticket.load(std::memory_order_relaxed);
         ++t.read;
         ++t.write;
-        store_release(&ticket.readWrite, t.readWrite);
+        ticket.store(t.whole, std::memory_order_release);
     }
```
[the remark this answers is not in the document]
alogfans: Same as above

mooncake-transfer-engine/tent/include/tent/transport/rdma/cq.h
```diff
@@ -29,6 +31,32 @@ class RdmaContext;
...
+        return *this;
+    }
+    RdmaCQ &operator=(RdmaCQ &&other) noexcept {
+        cq_ = other.cq_;
+        cqe_now_.store(other.cqe_now_.load(std::memory_order_relaxed),
+                       std::memory_order_relaxed);
+        cqe_limit_ = other.cqe_limit_;
+        context_ = other.context_;
+        return *this;
+    }
```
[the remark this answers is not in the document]
alogfans: I updated this to delete copy and move operations for TENT `RdmaCQ`. This matches the current ownership model: TENT stores CQs as `std::vector<RdmaCQ*>`, so the object itself does not need to be copied or moved.

I also tightened TE `RdmaCq` by deleting copy operations. It does not own-destroy `native` directly, but the native CQ is still managed by `RdmaContext`, so copying the wrapper would be misleading and unsafe.

alogfans:
@staryxchen CI in Ascend machines should include this.

he-yufeng (commented):
Focused on the RWSpinlock rewrite and the atomic-counter orderings, since a `volatile`/`__sync_*` to `std::atomic` conversion is easiest to get subtly wrong there. The synchronization-carrying paths check out.

**RWSpinlock:** collapsing the ticket to a single `std::atomic<uint64_t>` is the right call over the old mixed 64-bit-CAS-plus-16-bit-subfield access. Lock acquisition (`tryLock` / `writeLockAggressive` / `tryLockShared`) is `acquire`, and `unlock` plus `unlockShared` / `fetch_add_write` are `release`, so writer and reader critical sections publish correctly.

**One note, non-blocking, on dead code:** `unlockAndLockShared` → `fetch_add_read` now runs the CAS with `memory_order_acquire` only. The old `__sync_fetch_and_add(&ticket.read, 1)` was a full barrier, so this quietly drops the release side of a write-to-read downgrade — a downgrading writer would not publish its stores to readers that join after it. The tent `unlockAndLockShared` looks unreferenced today (the only live caller is the separate `RWSpinlock` in `common.h`), so nothing is broken now, but since the line is being touched it would be worth making that CAS `acq_rel` (or at least success-order `release`) so a future caller does not inherit a silently weakened downgrade. `unlock()` already uses whole-word `release` for the analogous writer-exit path.

**Counters:** the CQ/QP conversions look right. `cqe_now_` reads are `relaxed` (quota hint), state publication (`active_`, `status_ = DESTROYING`) is `release`, and the outstanding-WR `fetch_sub` is `acq_rel`, which is the one that actually has to order against completion processing.

I did not line-by-line the `rdma_context` / `worker_pool` churn, but every synchronization-carrying path I read is sound.

mooncake-transfer-engine/include/common.h
```diff
@@ -608,15 +609,44 @@ class RWSpinlock {
...
+    uint16_t fetch_add_read(uint16_t delta) {
+        uint64_t expected = ticket.load(std::memory_order_relaxed);
+        uint64_t new_val;
+        RWTicket t;
+        do {
+            t.whole = expected;
+            t.read += delta;
+            new_val = t.whole;
+        } while (!ticket.compare_exchange_weak(expected, new_val,
+                                               std::memory_order_acquire));
```
chestnut-Q: fetch_add_read() is a publish operation: The CAS success order should therefore be `memory_order_release`, not `memory_order_acquire`.

mooncake-transfer-engine/tent/include/tent/common/concurrent/rw_spinlock.h
```diff
@@ -106,15 +107,44 @@ class RWSpinlock {
...
+    uint16_t fetch_add_read(uint16_t delta) {
+        uint64_t expected = ticket.load(std::memory_order_relaxed);
+        uint64_t new_val;
+        RWTicket t;
+        do {
+            t.whole = expected;
+            t.read += delta;
+            new_val = t.whole;
+        } while (!ticket.compare_exchange_weak(expected, new_val,
+                                               std::memory_order_acquire));
```
chestnut-Q: Same as `common.h`, it should be `memory_order_release`

mooncake-transfer-engine/include/common.h
```diff
@@ -528,51 +528,46 @@ static inline bool overlap(const void *a, size_t a_len, const void *b,
...
     }
 
     void writeLockAggressive() {
         uint32_t count = 0;
-        uint16_t val = __sync_fetch_and_add(&ticket.users, 1);
-        while (val != load_acquire(&ticket.write)) {
+        RWTicket increment;
+        increment.users = 1;
+        RWTicket old(
+            ticket.fetch_add(increment.whole, std::memory_order_acquire));
```
chestnut-Q: This adds 1 << 32 to the whole 64-bit word, so every time users wraps past 0xFFFF the carry propagates into bits 48–63. Nothing reads those bits today, however, this relies on little-endian targets. We should be careful on big-endian layout.

mooncake-transfer-engine/tent/include/tent/common/concurrent/rw_spinlock.h
```diff
@@ -26,51 +26,46 @@ namespace tent {
...
     }
 
     void writeLockAggressive() {
         uint32_t count = 0;
-        uint16_t val = __sync_fetch_and_add(&ticket.users, 1);
-        while (val != load_acquire(&ticket.write)) {
+        RWTicket increment;
+        increment.users = 1;
+        RWTicket old(
+            ticket.fetch_add(increment.whole, std::memory_order_acquire));
```
chestnut-Q: Same as `common.h`, there is a dangerous carry-over.

chestnut-Q (approved):

PR #2907, opened 2026-07-14
[wheel ]feat: add schema-guided DataProto field encoding

zxpdemonio:
## Motivation

This PR is split out from the larger PR2671 rollout data transfer optimization work.

PR2671 currently mixes multiple review scopes, including schema-guided DataProto encoding, legacy dict/DataProto API changes, codec changes, BufferPool-backed GET, multi-buffer PUT, native copy acceleration, packaging updates, and broader structured object refactoring. To make the work easier to review and validate, this branch extracts the schema-driven DataProto field encoding part as the first standalone PR.

The goal of this PR is to make rollout data transfer less dependent on runtime sampling and codec inference for known non-tensor rollout fields. Different RL frameworks produce rollout data with different field shapes, including typed ragged arrays and dict-like multimodal tensor payloads. Schema hints allow these fields to select the intended structured codec explicitly while preserving the existing inferred path for fields without schema.

This branch is based on `kvcache/main` and is split from PR2671.

## Description

This PR adds schema-guided encoding support for DataProto non-tensor fields in `mooncake-wheel`.

Main changes:

- Add `FieldSchema` as an optional schema hint for DataProto fields.
- Extend `put_dataproto()` and `append_dataproto_fields()` with keyword-only `field_schemas`.
- Keep existing behavior unchanged for fields without schema hints and for `codec="auto"`.
- Add schema section validation for `batch`, `non_tensor_batch`, and `meta_info`.
- Add explicit schema-driven codec dispatch for non-tensor fields.
- Support typed ragged non-tensor fields with explicit dtype hints.
- Add schema-guided `ragged_tensor_dict` support for dict rows containing ragged tensor values.
- Support full, slice, and index reads for `ragged_tensor_dict`.
- Harden `ragged_tensor_dict` manifest/payload validation:
  - reject ambiguous keys
  - reject empty or path-like keys
  - validate expected payload members
  - validate `null_mask` and per-key row counts
- Add focused unit tests for schema-driven DataProto behavior.

This PR intentionally does not include the rest of PR2671, such as BufferPool-backed GET, multi-buffer PUT, native copy extension, packaging changes, or broader structured object refactoring. Those changes should be reviewed as follow-up PRs.

Layering of this PR:

1. External API:
   - Adds optional `field_schemas` to DataProto put/append APIs.

2. Leaf / manifest / schema:
   - Adds `FieldSchema`.
   - Adds schema section validation.
   - Adds schema codec/key validation.

3. Encode:
   - Adds schema-driven non-tensor encode dispatch.
   - Adds typed ragged and ragged tensor dict schema paths.

4. Decode:
   - Adds `ragged_tensor_dict` decode support.
   - Adds manifest/payload consistency checks.

5. Put:
   - Reuses the existing structured object put path.
   - Does not introduce new publish or storage lifecycle logic.

6. Get:
   - Adds full/slice/index read support for schema-guided ragged tensor dict fields.
   - Reuses existing ragged tensor range-read helpers.

Code size:

```text
2 files changed, 493 insertions(+), 7 deletions(-)
```

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [x] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**

```bash
PYTHONPATH=/root/Mooncake-PR2671-schema/mooncake-wheel \
python3 -m py_compile \
  mooncake-wheel/mooncake/structured_object_store.py \
  mooncake-wheel/tests/test_structured_object_store.py
```

```bash
PYTHONPATH=/root/Mooncake-PR2671-schema/mooncake-wheel \
python3 -m pytest --collect-only -q \
  mooncake-wheel/tests/test_structured_object_store.py \
  mooncake-wheel/tests/test_put_get_tensor.py
```

```bash
PYTHONPATH=/root/Mooncake-PR2671-schema/mooncake-wheel \
python3 -m pytest -q \
  mooncake-wheel/tests/test_structured_object_store.py \
  -k "field_schema"
```

```bash
PYTHONPATH=/root/Mooncake-PR2671-schema/mooncake-wheel \
python3 -m pytest -q \
  mooncake-wheel/tests/test_structured_object_store.py \
  -k "dataproto_helper_accepts_legacy_dict_inputs or dataproto_helper_treats_reserved_plain_dict_keys_as_batch_fields or dataproto_helper_rejects_cross_section_field_name_collision or dataproto_ref_handle_roundtrip_materializes_and_cleans_up or dataproto_helper_appends_stage_fields_without_rewriting_existing or put_object_roundtrips_numpy_and_torch_tensor_fields or put_object_roundtrips_wrapped_numpy_and_torch_values or field_schema"
```

**Test results:**

- [x] Unit tests pass

Results:

```text
117 tests collected
```

```text
field_schema: 3 passed, 1 skipped
```

```text
DataProto focused regression set: 8 passed, 3 skipped
```

The skipped tests are due to the local torch/CUDA import issue on the test machine:

```text
ImportError: libc10_cuda.so: undefined symbol: cudaGetDriverEntryPointByVersion
```

This is an environment issue on the test host and is not caused by this PR.

## Checklist

Notes:

- Commit-time pre-commit hooks ran successfully during local commit.
- Full `pre-commit run --all-files` was not run.
- This PR is kept under the >500 LOC RFC threshold
- This PR is split from PR2671 specifically to keep the review scope small.

## AI Assistance Disclosure

- [ ] No AI tools were used

AI tools were used to help inspect the existing PR2671 changes, prepare the split branch, run review checklists, and draft this PR description. I‘m responsible for understanding, validating, and defending all changes.

### DIFF
--- a/mooncake-wheel/mooncake/structured_object_store.py
+++ b/mooncake-wheel/mooncake/structured_object_store.py
@@ -4,10 +4,11 @@ import ctypes
 import io
 import json
 import uuid
+from collections.abc import Iterable
 from concurrent.futures import Future, ThreadPoolExecutor, as_completed
 from contextlib import contextmanager
 from dataclasses import dataclass, field
-from typing import Any, Iterator, Literal, Mapping, Optional, Protocol, Sequence
+from typing import Any, Callable, Iterator, Literal, Mapping, Optional, Protocol, Sequence
 
 import numpy as np
 
@@ -42,6 +43,21 @@ class BundleTransferPolicy:
     max_inflight_put: int = 1
     put_mode: Literal["auto", "batch", "parallel"] = "auto"
     copy_mode: Literal["auto", "zero_copy", "copy"] = "auto"
+
+
+@dataclass(frozen=True)
+class FieldSchema:
+    """Schema hint for DataProto fields.
+
+    ``field_schemas`` maps DataProto field names to hints. Unlisted fields and
+    ``codec="auto"`` keep inference. ``metadata["section"]`` may pin a field to
+    ``"batch"``, ``"non_tensor_batch"``, or ``"meta_info"``. ``ragged_tensor_dict``
+    keys must be a finite iterable of non-empty strings without path separators.
+    """
+
+    codec: str
+    nullable: bool = True
+    metadata: dict[str, Any] = field(default_factory=dict)
 
 
 @dataclass
@@ -471,8 +487,9 @@ class MooncakeBundleTransfer:
         stage: str = "default",
         chunk_bytes: Optional[int] = None,
         policy: Optional[BundleTransferPolicy] = None,
+        field_schemas: Optional[Mapping[str, FieldSchema]] = None,
     ) -> MooncakeDataProtoRef:
-        """Store a DataProto-like object as a stage-level structured object."""
+        """Store DataProto fields, optionally using schema hints for non-tensor fields."""
         return self._put_dataproto_stage(
             None,
             data,
@@ -482,6 +499,7 @@ class MooncakeBundleTransfer:
             chunk_bytes=chunk_bytes,
             policy=policy,
             overwrite=False,
+            field_schemas=field_schemas,
         )
 
     def append_dataproto_fields(
@@ -493,8 +511,9 @@ class MooncakeBundleTransfer:
         overwrite: bool = False,
         chunk_bytes: Optional[int] = None,
         policy: Optional[BundleTransferPolicy] = None,
+        field_schemas: Optional[Mapping[str, FieldSchema]] = None,
     ) -> MooncakeDataProtoRef:
-        """Append fields from a later DataProto stage without rewriting previous stages."""
+        """Append DataProto fields, optionally using schema hints for new fields."""
         ref = _resolve_dataproto_ref(ref)
         return self._put_dataproto_stage(
             ref,
@@ -505,6 +524,7 @@ class MooncakeBundleTransfer:
             chunk_bytes=chunk_bytes,
             policy=policy,
             overwrite=overwrite,
+            field_schemas=field_schemas,
         )
 
     def dataproto_manifest_view(self, ref: DataProtoRefLike) -> dict[str, Any]:
@@ -909,6 +929,16 @@ class MooncakeBundleTransfer:
                 "nulls": read_member_indices("nulls", indices),
             }, metadata
 
+        if codec == "ragged_tensor_dict":
+            return self._read_ragged_tensor_dict_payload_with_reader(
+                payload_members,
+                metadata,
+                lambda: read_member_indices("null_mask", indices),
+                lambda encoded: self._read_structured_non_tensor_payload_indices(
+                    stage_ref, encoded, indices
+                ),
+            )
+
         if codec in {"media_bytes", "bytes_ragged", "utf8_ragged", "msgpack_ragged", "json_ragged"}:
             offsets = read_member_indices(
                 "offsets", [index for row in indices for index in (row, row + 1)]
@@ -1074,6 +1104,16 @@ class MooncakeBundleTransfer:
                 "nulls": read_member("nulls", start, end),
             }, metadata
 
+        if codec == "ragged_tensor_dict":
+            return self._read_ragged_tensor_dict_payload_with_reader(
+                payload_members,
+                metadata,
+                lambda: read_member("null_mask", start, end),
+                lambda encoded: self._read_structured_non_tensor_payload_slice(
+                    stage_ref, encoded, row_slice, total_rows
+                ),
+            )
+
         if codec in {"media_bytes", "bytes_ragged", "utf8_ragged", "msgpack_ragged", "json_ragged"}:
             offsets = read_member("offsets", start, end + 1)
             base = int(offsets[0])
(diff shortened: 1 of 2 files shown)


mooncake-wheel/mooncake/structured_object_store.py
```diff
@@ -2026,11 +2277,138 @@ def _decode_ragged_tensor_values(payload: dict[str, Any], rows: int) -> list[Any
...
+                f"row {row} is {type(item).__name__}"
+            )
+    schema_keys = schema.metadata.get("keys")
+    keys = (
+        None
+        if schema_keys is None
+        else _normalize_ragged_tensor_dict_keys(schema_keys)
+    )
+    non_null = [item for item in values if item is not None]
+    if not non_null:
```
yokinoshitayoki: With declared keys, this early return writes metadata advertising those keys but omits all per-key payloads. For example, values=[None, None] with metadata={"keys": ["image"]} is accepted by put_dataproto(), but get_dataproto() fails because image.data/offsets/shapes/ndims/nulls are missing. Please emit valid empty ragged payloads for every declared key, or make all read paths support this representation, and add full/slice/index tests for an all-null field.

zxpdemonio: Thanks for the catch. Fixed in the latest commit.

For `ragged_tensor_dict` fields with declared schema keys and all-null rows, we now still emit valid empty ragged payloads for every declared key instead of only writing the top-level null mask/metadata. This keeps the manifest consistent with the declared keys and makes `get_dataproto()` work for full, slice, and indexed reads.

I also added a regression test covering the all-null declared-key case for full/slice/index reads.

mooncake-wheel/tests/test_structured_object_store.py
```diff
@@ -1549,6 +1550,158 @@ def test_dataproto_helper_accepts_legacy_dict_inputs() -> None:
...
+                codec="ragged_tensor_dict",
+                metadata={"section": "non_tensor_batch", "keys": ["image"]},
+            )
+        },
+    )
+
+    encoded = ref.encoded_non_tensor["multi_modal_inputs"]
+    assert encoded["codec"] == "ragged_tensor_dict"
+    assert encoded["metadata"]["keys"] == ["image"]
+    assert "image.data" in encoded["payload"]
```
yokinoshitayoki: This regression test fails whenever torch is available. ref.encoded_non_tensor[...] stores payload_members rather than the raw encoded payload, so encoded["payload"] raises KeyError.

mooncake-wheel/mooncake/structured_object_store.py
```diff
@@ -2026,11 +2277,138 @@ def _decode_ragged_tensor_values(payload: dict[str, Any], rows: int) -> list[Any
...
+        all_keys: set[str] = set()
+        for item in non_null:
+            all_keys.update(item.keys())
+        keys = _normalize_ragged_tensor_dict_keys(sorted(all_keys))
+    if _torch is None:
+        raise RuntimeError("torch is required to encode ragged_tensor_dict fields")
+    payload: dict[str, Any] = {"null_mask": null_mask}
+    key_codecs: dict[str, Any] = {}
+    for key in keys:
+        key_values = [None if item is None else item.get(key) for item in values]
```
yokinoshitayoki: When metadata["keys"] is provided, this silently projects every row onto the declared keys. A row such as {"image": t1, "audio": t2} with keys=["image"] round-trips without "audio". It also collapses a present {"image": None} and a missing image key because both decode as an omitted key. A transfer API should not silently change dictionary contents.

mooncake-wheel/mooncake/structured_object_store.py
```diff
@@ -1203,10 +1295,24 @@ def _put_dataproto_stage(
             field_updates[name] = StructuredFieldLocation(stage, member, "batch")
         encoded_updates: dict[str, Any] = {}
         for name, value in non_tensor_batch.items():
-            if _should_encode_non_tensor_field(value):
+            schema = field_schemas.get(name) if field_schemas else None
+            encoded: _EncodedStructuredLeaf | None = None
+            if schema is not None and schema.codec != "auto":
```
yokinoshitayoki: This condition bypasses _encode_with_schema() for codec="auto", so nullable=False is never enforced. FieldSchema(codec="auto", nullable=False) currently accepts None.

mooncake-wheel/mooncake/structured_object_store.py
```diff
@@ -1585,6 +1691,87 @@ def _select_mapping(
...
+        ("meta_info", meta_info),
+    ):
+        for name in fields:
+            schema = field_schemas.get(name)
+            if schema is None:
+                continue
+            declared = _schema_section(name, schema)
+            if declared is None:
+                continue
+            if declared != section:
```
yokinoshitayoki: meta_info is allowed to reuse a field name, but this loop applies the same FieldSchema to every section containing that name. A schema pinned to non_tensor_batch therefore rejects an otherwise valid same-named meta_info entry.

zxpdemonio:
@yokinoshitayoki Hi, I have rebased this PR onto the latest `main` and addressed the previous review comments.

The current CI checks are all green on the latest head. Could you please take another look when you have time? Please let me know if there are any remaining issues before approval.

If there are no other concerns, would it be possible to approve this PR?

Thanks!

mooncake-wheel/tests/test_structured_object_store.py
```diff
@@ -1549,6 +1550,137 @@ def test_dataproto_helper_accepts_legacy_dict_inputs() -> None:
...
+                "text": FieldSchema(
+                    codec="auto",
+                    nullable=False,
+                    metadata={"section": "non_tensor_batch"},
+                )
+            },
+        )
+
+
+def test_dataproto_field_schema_encodes_ragged_tensor_dict() -> None:
```
yokinoshitayoki, line 1693: The slimming commit removes the regression coverage for undeclared ragged-tensor-dict keys and explicit nested None values, even though those checks were added to fix concrete data-loss behavior. It also removes the all-null slice/index assertions that were added for the previous read failure. Should we consider retain compact parameterized coverage for these cases?

zxpdemonio: Thanks, the test cases are added back.

mooncake-wheel/mooncake/structured_object_store.py
```diff
@@ -1585,6 +1675,89 @@ def _select_mapping(
...
+
+def _validate_dataproto_schema_sections(
+    batch: Mapping[str, Any],
+    non_tensor_batch: Mapping[str, Any],
+    meta_info: Mapping[str, Any],
+    field_schemas: Optional[Mapping[str, FieldSchema]],
+) -> None:
+    if not field_schemas:
+        return
+    actual_sections = {
```
yokinoshitayoki: Actual_sections stores only one section per field name, so a same-named meta_info field overwrites non_tensor_batch. A schema declared for non_tensor_batch is therefore rejected as if the field only existed in meta_info.

The inverse is also incorrect: declaring section="meta_info" passes validation, but the schema is still applied to the same-named non_tensor_batch field in _put_dataproto_stage(). Please preserve all section memberships for each name and only apply a schema to the section it declares. The same-name regression test from 3c155697 should also be restored.

zxpdemonio: Thanks, fixed and added test cases.

yokinoshitayoki (approved):
Thanks a lot. This PR looks good to me now.

PR #2909, opened 2026-07-14
[Store] Add fixed soft pin lifecycle

Lin-z-w:
## Description

This PR changes object-level soft pinning from an access-renewed lease into a fixed lifecycle. Previously, reads could extend an object's soft-pin deadline, and expired soft-pin metadata could remain set indefinitely when the object was not accessed or considered for eviction.

The change:

- replaces the C++ and Python `with_soft_pin` boolean with `SoftPinAction::{PRESERVE, ENABLE, DISABLE}`;
- adds an optional request-level `soft_pin_ttl_ms`; an omitted value uses the Master default, zero commits ordinary cache, and values above `max_kv_soft_pin_ttl` are rejected;
- starts the deadline when the first replica becomes readable and prevents reads, repeated `PutEnd`, later replica completion, copy/move, and promotion paths from renewing it;
- keeps a pending action until the write first becomes readable, including partial-revoke handling when another processing replica can still complete;
- indexes committed deadlines in a bounded min-heap so the existing cleanup thread processes only due entries instead of scanning every metadata object;
- handles TTL updates through lazy invalidation and rebuilds the heap when stale entries exceed `max(4096, active registrations * 2)`;
- removes soft-pin evaluation from `Get`, `Exist`, and group read-lease refresh paths while eviction still evaluates the deadline against its current timestamp;
- safely saturates unrepresentable `completion_time + TTL` calculations instead of relying on a startup-time range check;
- rebuilds the derived deadline index only after successful snapshot recovery and clears it between failed snapshot fallback attempts; and
- preserves the existing HA recovery policy: an expired soft deadline is normalized away, after which the ordinary read lease determines whether the object survives restoration.

After a deadline is reached, eviction immediately treats the object as ordinary cache. Stored metadata and `master_soft_pin_key_count` converge within the existing cleanup interval; reads no longer perform cleanup work merely to make the gauge current.

This is intended as object-level lifecycle groundwork for [PR #2835](https://github.com/kvcache-ai/Mooncake/pull/2835). Group retention remains a separate priority source and is not stored in each member's object soft-pin state.

This is a source and behavior breaking change for C++ and Python callers: `with_soft_pin` is removed and default Upsert behavior becomes `PRESERVE`. Master and C++/Python clients are expected to be upgraded together; rolling wire compatibility is not provided. The C/Go/Rust boolean ABI remains available, mapping `true` to `ENABLE` with the default TTL and `false` to `PRESERVE`.

## Module

- [ ] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [x] Breaking change
- [x] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

The complete Master service and snapshot test binaries passed. Focused coverage includes tri-state validation, request TTL and zero-TTL behavior, first-readable activation, Preserve/Disable/re-enable Upsert transitions, deadline-index cleanup, stale-entry invalidation, heap compaction, non-renewal by reads or repeated End calls, gauge balancing, snapshot fallback, and restoration with expired soft-pin and ordinary-lease combinations.

**Test commands:**

```bash
cmake --build build --target master_service_test master_service_test_for_snapshot snapshot_child_process_test -j2

GLOG_minloglevel=2 build/mooncake-store/tests/master_service_test --gtest_brief=1

GLOG_minloglevel=2 build/mooncake-store/tests/master_service_test_for_snapshot --gtest_brief=1

GLOG_minloglevel=2 build/mooncake-store/tests/snapshot_child_process_test --gtest_brief=1

pre-commit run --files mooncake-store/include/master_service.h mooncake-store/src/master_service.cpp mooncake-store/tests/master_service_test.cpp mooncake-store/tests/ha/snapshot/master_service_test_for_snapshot.cpp mooncake-store/tests/ha/snapshot/snapshot_child_process_test.cpp
```

**Test results:**

- [x] Unit tests pass
- [ ] Integration tests pass (not run)

Results:

- `master_service_test`: 151/151 tests passed, 322.35 seconds.
- `master_service_test_for_snapshot`: 69/69 tests passed, 204.93 seconds.
- `snapshot_child_process_test`: 23/23 tests passed, 37.67 seconds.
- Focused soft-pin and recovery runs passed before the complete binaries were run.
- Targeted pre-commit hooks passed for all five files in the final change.

The local build has `USE_NOF=OFF`; the new conditional tests for partial NoF revoke and later NoF completion were added but were not executed locally. `make -C docs html` could not run because `sphinx-build` is not installed in the environment.

## Checklist

An RFC issue draft has been prepared and will be linked here after publication. The related group-retention design remains in PR #2835.

## AI Assistance Disclosure

- [ ] No AI tools were used

OpenAI Codex was used to inspect the existing soft-pin, Upsert, metrics, group lease, and snapshot-recovery paths; implement the API and state-machine changes. I reviewed and edited the generated code before submission.


docs/source/design/mooncake-store.md
yokinoshitayoki: These bullets still describe the previous access-renewed behavior. Could we update these two bullets accordingly?

mooncake-store/include/master_service.h
```diff
@@ -1126,17 +1139,107 @@ class MasterService {
...
+                    if (pending.ttl_ms == 0) {
+                        if (soft_pin_timeout) {
+                            soft_pin_timeout.reset();
+                            --metric_delta;
+                        }
+                    } else {
+                        if (!soft_pin_timeout) {
+                            ++metric_delta;
+                        }
+                        soft_pin_timeout =
```
yokinoshitayoki, line 1326: The startup validation does not make this addition safe for the lifetime of the service. It validates `max_kv_soft_pin_ttl` against `time_point::max() - startup_now`, but the deadline is calculated later, when the first replica becomes readable.

yokinoshitayoki (approved):
LGTM.

ykwd (approved):

Lin-z-w:
Hi @stmatengss , could you please review the Mooncake Wheel changes in this PR, or help find an appropriate reviewer?

mooncake-store/src/master_service.cpp
```diff
@@ -2626,8 +2870,9 @@ void MasterService::RestoreFromStandbySnapshot(
                 std::piecewise_construct, std::forward_as_tuple(user_key),
                 std::forward_as_tuple(
                     standby_meta.client_id, now, standby_meta.size,
-                    std::move(replicas), false, false, standby_meta.data_type,
-                    standby_meta.group_id, tenant_id, user_key));
+                    std::move(replicas), std::nullopt, false,
```
00fish0, line 2865: This PR updates the corresponding restore and deadline-index rebuilding logic. However, the HA standby/promotion path still appears to drop the soft-pin state completely:

1. MetadataPayload and StandbyObjectMetadata do not contain a committed soft-pin deadline;
2. DeserializeStandbyObjectMetadata() uses the snapshot deadline only to decide whether an object is still restorable, but does not retain it;
3. subsequent PUT_END oplog entries do not carry ENABLE, DISABLE, or the committed deadline;
4. RestoreFromStandbySnapshot() constructs every restored object with std::nullopt for the soft-pin deadline.

As a result, a promoted standby restores these objects as ordinary cache, and soft-pin changes made after the standby snapshot cannot be reproduced. Is losing object-level soft pins on HA failover intentional?

00fish0 (commented):
I noticed that full snapshot recovery keeps the soft-pin deadline, while the standby/oplog path currently drops it. Do we actually need to preserve the exact deadline across HA promotion?

Would it be simpler to keep only whether the object was soft-pinned and give it a fresh default_kv_soft_pin_ttl after promotion, similar to how ordinary leases are treated?

This would not preserve request-level TTLs exactly—a failover could extend a nearly expired pin, shorten a longer one, and repeated failovers could effectively renew it—but that may be acceptable if soft pinning is considered best-effort.

We would still need at least an active/inactive bit in the snapshot/oplog; otherwise the standby cannot distinguish soft-pinned objects from ordinary ones.

What HA semantic do we want here: preserve the exact deadline, reset active soft pins to the default TTL on promotion, or simply drop soft-pin state across failover?

mooncake-store/src/master_service.cpp
Lin-z-w: Thanks for catching this. I agree that soft pin should survive HA failover without resetting its lifecycle.

I updated the implementation to preserve the exact absolute committed soft-pin deadline across the snapshot, OpLog, and promotion paths:

- Active snapshot deadlines are stored in standby metadata.
- Authoritative PUT_END records update or explicitly clear the deadline.
- Replica-only PUT_END records preserve the existing standby deadline.
- Promotion restores only deadlines that are still active and registers them in the deadline heap/gauge.
- Already expired deadlines are normalized to ordinary cache entries.

The TTL is not recomputed or extended during failover. I also added coverage for snapshot restore, OpLog apply/preserve/clear behavior, and standby promotion.

Lin-z-w:
> I noticed that full snapshot recovery keeps the soft-pin deadline, while the standby/oplog path currently drops it. Do we actually need to preserve the exact deadline across HA promotion?
> 
> Would it be simpler to keep only whether the object was soft-pinned and give it a fresh default_kv_soft_pin_ttl after promotion, similar to how ordinary leases are treated?
> 
> This would not preserve request-level TTLs exactly—a failover could extend a nearly expired pin, shorten a longer one, and repeated failovers could effectively renew it—but that may be acceptable if soft pinning is considered best-effort.
> 
> We would still need at least an active/inactive bit in the snapshot/oplog; otherwise the standby cannot distinguish soft-pinned objects from ordinary ones.
> 
> What HA semantic do we want here: preserve the exact deadline, reset active soft pins to the default TTL on promotion, or simply drop soft-pin state across failover?

I think preserving the exact deadline best matches the fixed soft-pin lifecycle. Unlike read leases, HA promotion should not renew object-level soft pins. Resetting the TTL could extend or shorten the original request-level TTL, while dropping it could cause premature eviction. Therefore, promotion restores only unexpired deadlines without recomputing or extending them.

Lin-z-w:
> I noticed that full snapshot recovery keeps the soft-pin deadline, while the standby/oplog path currently drops it. Do we actually need to preserve the exact deadline across HA promotion?
> 
> Would it be simpler to keep only whether the object was soft-pinned and give it a fresh default_kv_soft_pin_ttl after promotion, similar to how ordinary leases are treated?
> 
> This would not preserve request-level TTLs exactly—a failover could extend a nearly expired pin, shorten a longer one, and repeated failovers could effectively renew it—but that may be acceptable if soft pinning is considered best-effort.
> 
> We would still need at least an active/inactive bit in the snapshot/oplog; otherwise the standby cannot distinguish soft-pinned objects from ordinary ones.
> 
> What HA semantic do we want here: preserve the exact deadline, reset active soft pins to the default TTL on promotion, or simply drop soft-pin state across failover?

Thanks for pointing this out. After further discussion with the HA owners, I need to correct my previous reply: losing soft-pin state during failover is intentional.

Soft pin is a best-effort eviction hint, not durable state. Persisting its deadline would add unnecessary OpLog overhead and replay complexity. Recovered objects will therefore become ordinary cache; we will not reset the default TTL, as that could incorrectly extend or reactivate soft pinning. Documentation and tests will be updated accordingly.

mooncake-wheel/tests/test_distributed_object_store.py
```diff
@@ -662,16 +663,16 @@ def test_replicate_config_creation_and_properties(self):
...
         
         # Test property assignment
         config.replica_num = 3
-        config.with_soft_pin = True
+        config.soft_pin_action = SoftPinAction.ENABLE
         config.preferred_segment = "node1:12345"
         
         self.assertEqual(config.replica_num, 3)
-        self.assertEqual(config.with_soft_pin, True)
+        self.assertEqual(config.soft_pin_action, SoftPinAction.ENABLE)
```
zxpdemonio, line 723: Could we add one focused wheel regression that actually sends the new soft-pin fields through a Store operation? The current wheel changes only round-trip `soft_pin_action` on `ReplicateConfig`; `soft_pin_ttl_ms` is not referenced, and the replicated cases set `PRESERVE` on newly created objects where it is a no-op. The Master C++ tests do not cover pybind optional conversion or Python-to-client forwarding. Minimal coverage could verify that `ENABLE` with a valid TTL succeeds, while `PRESERVE`/`DISABLE` with a TTL is rejected, plus a default/PRESERVE and DISABLE Upsert transition. A small no-service binding test can separately cover `None -> int -> None` and the negative/uint64-overflow exception boundary.

Lin-z-w: Thanks for the suggestion. Added the focused wheel regressions in `38b819de`:

- A no-service binding test covering `soft_pin_ttl_ms`: `None -> int -> None`, negative values, and `uint64_t` overflow.
- An end-to-end Store test covering `ENABLE` with a valid TTL, rejection of `PRESERVE`/`DISABLE` combined with a TTL, and default `PRESERVE` followed by explicit `DISABLE` through `upsert`.

This exercises the Python → pybind → client → Master forwarding path. Both new tests pass locally.

zxpdemonio (approved):

Icedcoco (approved):

PR #2915, opened 2026-07-14
[Store] feat: implement fsspec-backed tensor file I/O and enhance documentation

stmatengss:
## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
```

**Test results:**
- [ ] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

### DIFF
--- a/mooncake-wheel/mooncake/store_file_io.py
+++ b/mooncake-wheel/mooncake/store_file_io.py
@@ -1,0 +1,333 @@
+from __future__ import annotations
+
+import logging
+import os
+import posixpath
+from importlib import import_module
+from typing import Any
+
+
+LOGGER = logging.getLogger(__name__)
+
+INVALID_PARAMS = -600
+FILE_NOT_FOUND = -1100
+PERSISTENT_FAIL = -1503
+
+
+def _normalize_format(format_name: str, file_name: str) -> str:
+    if format_name is None:
+        format_name = "auto"
+
+    normalized = format_name.lower().replace("-", "_")
+    if normalized == "auto":
+        if str(file_name).lower().endswith(".safetensors"):
+            return "safetensors"
+        return "torch"
+
+    if normalized in {"safetensors", "safe_tensors"}:
+        return "safetensors"
+    if normalized in {"torch", "pt", "pth", "standard", "standard_file"}:
+        return "torch"
+
+    raise ValueError(f"Unsupported file format: {format_name}")
+
+
+def _normalize_filesystem(filesystem: str | None) -> str:
+    if filesystem is None:
+        return "auto"
+
+    normalized = filesystem.lower().replace("-", "_")
+    if normalized in {"local", "localfs", "posix", "file"}:
+        return "file"
+    return normalized
+
+
+def _build_target_url(file_name: os.PathLike[str] | str, filesystem: str | None) -> str:
+    target = os.fspath(file_name)
+    if not target:
+        raise ValueError("file_name must not be empty")
+
+    normalized_fs = _normalize_filesystem(filesystem)
+    if normalized_fs in {"auto", "file"} or "://" in target:
+        return target
+
+    return f"{normalized_fs}://{target}"
+
+
+def _open_fs_target(
+    file_name: os.PathLike[str] | str,
+    filesystem: str | None,
+    storage_options: dict[str, Any] | None,
+):
+    import fsspec
+
+    target_url = _build_target_url(file_name, filesystem)
+    options = dict(storage_options or {})
+    fs, path = fsspec.core.url_to_fs(target_url, **options)
+    return fs, path
+
+
+def _ensure_parent_dir(fs, path: str) -> None:
+    """Create parent directories, tolerating backends that lack makedirs."""
+    parent_dir = posixpath.dirname(path)
+    if parent_dir:
+        try:
+            fs.makedirs(parent_dir, exist_ok=True)
+        except (AttributeError, NotImplementedError):
+            pass
+        except OSError as exc:
+            LOGGER.warning(
+                "Failed to create directory %s: %s", parent_dir, exc
+            )
+
+
+def _write_bytes(
+    file_name: os.PathLike[str] | str,
+    payload: bytes,
+    filesystem: str | None,
+    storage_options: dict[str, Any] | None,
+) -> None:
+    fs, path = _open_fs_target(file_name, filesystem, storage_options)
+    _ensure_parent_dir(fs, path)
+
+    with fs.open(path, "wb") as handle:
+        handle.write(payload)
+
+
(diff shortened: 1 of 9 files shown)


mooncake-wheel/mooncake/store_file_io.py
```diff
@@ -0,0 +1,599 @@
...
+        except ValueError:
+            LOGGER.exception(
+                "Invalid artifact_kind %r while loading tensor from %s",
+                artifact_kind,
+                file_name,
+            )
+            return None
+
+        if kind == "safetensor":
+            return self.load_tensor_from_safetensor(
```
Aionw, line 595: [P1] `load_tensor()` exposes `map_location`, but this safetensors routing branch silently drops it because `load_tensor_from_safetensor()` neither accepts nor forwards the argument. The returned tensor—and the tensor written into Store—therefore stays on the decoder's device instead of the requested one. Please add `map_location` to the wrapper, forward it to `load_tensor_from_file()`, and cover this route with a test.

mooncake-wheel/mooncake/store_file_io.py
```diff
@@ -0,0 +1,599 @@
...
+        return
+
+    if getattr(store_cls, "_mooncake_file_io_patched", False):
+        return
+
+    store_cls.save_tensor_to_file = _save_tensor_to_file
+    store_cls.load_tensor_from_file = _load_tensor_from_file
+    store_cls.save_kv_cache_to_file = _save_tensor_to_file
+    store_cls.load_kv_cache_from_file = _load_tensor_from_file
+    store_cls.save_tensor_to_safetensor = _save_tensor_to_safetensor
```
Aionw, line 670: [P1] These assignments unconditionally replace the existing C++ local safetensors methods on every normal package import. Besides removing the efficient `save_file`/`load_file` local path noted in the other threads, this changes existing multi-entry selection and error semantics; RFC #2868 explicitly calls for preserving the C++ local compatibility path. Please retain and use the native methods for local legacy calls (or avoid overriding existing methods) and route only remote targets through fsspec.

mooncake-wheel/mooncake/store_file_io.py
```diff
@@ -0,0 +1,599 @@
...
+_TORCH_LOAD_SUPPORTS_WEIGHTS_ONLY: bool | None = None
+
+
+def _normalize_format(format_name: str, file_name: str) -> str:
+    if format_name is None:
+        format_name = "auto"
+
+    normalized = format_name.lower().replace("-", "_")
+    if normalized == "auto":
+        if str(file_name).lower().endswith(".safetensors"):
```
Aionw: [P2] `auto` selects torch for every path that does not end in `.safetensors`. For an extensionless remote target such as `s3://bucket/checkpoint`, the default save therefore writes a torch checkpoint, while the default load rejects that same remote torch artifact at lines 346-356. The default API cannot round-trip such paths. Please select torch only for explicit `.pt`/`.pth` suffixes and default ambiguous paths to safetensors (as proposed in RFC #2868), or reject the ambiguity before writing.

mooncake-wheel/mooncake/store_file_io.py
```diff
@@ -0,0 +1,599 @@
...
+    *,
+    file_name: os.PathLike[str] | str | None = None,
+    artifact_kind: str = "tensor",
+    format: str = "auto",
+    filesystem: str = "auto",
+    storage_options: dict[str, Any] | None = None,
+    tensor_name: str | None = None,
+) -> int:
+    """General save API: write to Store, export to file, or both."""
+    if tensor is not None:
```
Aionw, line 527: [P2] This mutates Store before validating `artifact_kind` below. For example, an invalid kind returns `INVALID_PARAMS` after `put_tensor()` has already succeeded, leaving distributed state behind; a corrected retry may then fail because the key already exists. Please perform all side-effect-free routing/format validation before this write, and add an assertion that invalid input does not insert the key.

Aionw (commented):
Thanks for adding the fsspec-backed artifact I/O layer. I found four behavior and compatibility issues that should be addressed before merge. Existing threads already cover the local safetensors full-buffer regressions and the missing `os` import, so I have not duplicated those comments here. I am submitting this as a non-blocking COMMENT review, but recommend fixing the inline findings before approval.

mooncake-wheel/mooncake/store_file_io.py
```diff
@@ -0,0 +1,674 @@
...
+                from safetensors.torch import load as safetensors_load
+                from safetensors.torch import load_file as safetensors_load_file
+            except ImportError:
+                raise ValueError(
+                    "safetensors package is required for the 'safetensors' "
+                    "format. Install with: pip install safetensors"
+                )
+            if _is_remote_target(file_name, filesystem):
+                with fs.open(path, "rb") as handle:
+                    payload = handle.read()
```
Aionw, line 417: [P1] This remote path materializes the entire object in Python memory before decoding it; the save path similarly builds a complete `bytes` payload with `safetensors_save()` before writing. For the model-weight, KV-cache, and RL-checkpoint use cases described by this PR, multi-GB artifacts can therefore require the serialized payload, decoded tensor, and Store copy to coexist and can OOM. Please use a disk-backed temporary/cache file with chunked fsspec transfer and `load_file()` / `save_file()` (or otherwise bound the buffering), and add a large/remote-path regression test.

mooncake-wheel/mooncake/store_file_io.py
```diff
@@ -0,0 +1,674 @@
...
+        and file_name is not None
+        and _can_use_native_safetensor_io(
+            file_name,
+            filesystem,
+            storage_options,
+            tensor_name,
+            map_location=map_location,
+        )
+    ):
+        return native_method(self, key, file_name)
```
Aionw, line 490: [P2] The wrapper advertises `os.PathLike[str]`, but this native branch forwards a `pathlib.Path` unchanged to the C++ binding, whose `file_name` argument accepts only `str`; `load_tensor_from_safetensor(None, Path(...))` therefore raises `TypeError` before entering the native implementation. The save wrapper has the same issue. Please pass `os.fspath(file_name)` to both native methods (preserving `None` for the optional save filename) and cover local `Path` inputs in tests.

Aionw (commented):
I found two remaining issues in the current artifact I/O implementation: remote safetensors I/O has unbounded full-object memory amplification, and the local native compatibility route does not honor the advertised `os.PathLike` input type.

PR #2919, opened 2026-07-14
[Dep] bind wildcard listen address (0.0.0.0/::) without going through getaddrinfo

### LINKED ISSUE #2798: [Bug]:mooncake_master startup failed with "bad address: 0.0.0.0 error: Host not found (authoritative)"

pjdurden:
### Summary
`mooncake_master` binds its RPC and HTTP servers to the numeric wildcard address `0.0.0.0`. The ylt acceptors resolved that through `getaddrinfo()`, which some environments answer with `EAI_NONAME` for numeric literals, so startup died with `bad address: 0.0.0.0 error: Host not found (authoritative)`.

### Fix
Bump `extern/yalantinglibs` to a revision that parses numeric IP literals directly instead of handing them to the resolver.

Fixes #2798

### Submodule bump scope
The `extern/yalantinglibs` bump spans more than the listen fix because the RPC-server dual-stack listen only lands in `#1190` (`7801bc9`), which is the tip of the range; `#1186` alone covers only the HTTP acceptor. Since a submodule pointer cannot cherry-pick, pinning to `#1190` pulls in the intermediate commits, including two `coro_io` changes (`ddffb09`, `bd7bae3`) both confined to `include/ylt/coro_io/client_pool.hpp`.

Those two only affect the client pool's private background idle-reaper and its teardown (a coroutine-frame-leak fix plus a launch/yield refactor). Mooncake's TransferEngine and store RPC clients use only the public pool surface (`client_pools`, `send_request`, `at`) and leave the reaper cadence at defaults, so they do not exercise the changed internals; the net effect is a cleaner pool teardown at shutdown with no change on the request path. The RSA mutual-auth work (`#1173`) in the same span is opt-in and off by default.


he-yufeng (commented):
Thanks for this — the fix layer is the right one (parsing the numeric literal inside the ylt acceptors covers every wildcard-bind caller, not just `mooncake_master`), and the regression test is nicely done: interposing `getaddrinfo` via `RTLD_NEXT` reproduces the `EAI_NONAME`-on-numeric-literal environment on any host, and the null-`real_getaddrinfo` fallback keeps it safe under static linking. The two `async_start` + `hasResult()` assertions are a clean way to prove the listen path no longer routes wildcards through the resolver.

One thing I'd like to nail down before this goes in is the blast radius of the submodule bump. `6a0e067 → 7801bc9` is 16 commits / 49 files / +2974 −515, well beyond the dual-stack listen fix (#1186, #1190) this PR actually needs. It also pulls in unrelated work — the RSA mutual-auth feature (#1173, with new cert fixtures and SSL tests) and, more relevant to us, `coro_io` lifecycle changes: `ddffb09` closes idle coroutines via a cancellation signal to avoid frame leaks, and `bd7bae3` simplifies client-pool cleanup. Those touch the connection-pool path that TransferEngine and the store RPC clients lean on.

Since a submodule bump can't cherry-pick, could you confirm the span was a conscious call — i.e. you've sanity-checked that the intermediate `coro_io` pool/cancellation changes don't shift behavior on our RPC/TE paths — and note that in the PR description? If the intermediate changes are risky to take right now, the alternative is coordinating a ylt tag that isolates the listen fix. Happy to help verify the pool-path side if useful.

mooncake-store/tests/wildcard_listen_test.cpp
```diff
@@ -0,0 +1,80 @@
...
+    ASSERT_FALSE(stopped.hasResult())
+        << "failed to listen on 0.0.0.0: " << server.get_errc().message();
+    EXPECT_EQ(server.port(), port);
+
+    server.stop();
+}
+
+TEST(WildcardListenTest, HttpServerListensWhenResolverRejectsWildcard) {
+    const auto port = static_cast<unsigned short>(mooncake::getFreeTcpPort());
+    coro_http::coro_http_server server(1, port, "0.0.0.0");
```
Aionw: Since 0.0.0.0 already tested, can we change this line into `::` so that we can get ipv6 test

pjdurden: Fixed . Switched the HTTP server test to bind `::` for IPv6 coverage (the RPC test still covers 0.0.0.0). Also guarded getFreeTcpPort with ASSERT_GT before the cast.

pjdurden:
> Thanks for this — the fix layer is the right one (parsing the numeric literal inside the ylt acceptors covers every wildcard-bind caller, not just `mooncake_master`), and the regression test is nicely done: interposing `getaddrinfo` via `RTLD_NEXT` reproduces the `EAI_NONAME`-on-numeric-literal environment on any host, and the null-`real_getaddrinfo` fallback keeps it safe under static linking. The two `async_start` + `hasResult()` assertions are a clean way to prove the listen path no longer routes wildcards through the resolver.
> 
> One thing I'd like to nail down before this goes in is the blast radius of the submodule bump. `6a0e067 → 7801bc9` is 16 commits / 49 files / +2974 −515, well beyond the dual-stack listen fix (#1186, #1190) this PR actually needs. It also pulls in unrelated work — the RSA mutual-auth feature (#1173, with new cert fixtures and SSL tests) and, more relevant to us, `coro_io` lifecycle changes: `ddffb09` closes idle coroutines via a cancellation signal to avoid frame leaks, and `bd7bae3` simplifies client-pool cleanup. Those touch the connection-pool path that TransferEngine and the store RPC clients lean on.
> 
> Since a submodule bump can't cherry-pick, could you confirm the span was a conscious call — i.e. you've sanity-checked that the intermediate `coro_io` pool/cancellation changes don't shift behavior on our RPC/TE paths — and note that in the PR description? If the intermediate changes are risky to take right now, the alternative is coordinating a ylt tag that isolates the listen fix. Happy to help verify the pool-path side if useful.

Fair concern. The span is mostly forced: the test brings up the RPC server on the wildcard, and RPC dual-stack listen only lands in #1190 (the tip), so pinning to it pulls in the intermediate commits since a submodule can’t cherry-pick.

The two coro_io ones you flagged are both confined to client_pool.hpp, so the TE/RPC-client exposure is just the pool-cleanup path. I’ll trace how our clients use client_pool against those commits and note it in the PR description. If you’d rather isolate it fully, I can get an upstream ylt tag backporting #1186+#1190 and pin to that instead.

pjdurden:
Ran the `client_pool` trace on the two flagged `coro_io` commits.

**How Mooncake drives the pool.** The TransferEngine RPC path (`mooncake-transfer-engine/src/transport/rpc_communicator/rpc_communicator.cpp`) and the store clients (`mooncake-store/src/real_client.cpp`, `dummy_client.cpp`) use only the public surface: construct `client_pools<coro_rpc_client>` from a `pool_config`, then `send_request(...)` / `at(addr)`. The `pool_config` fields we set are `client_config.socket_config` (RDMA vs TCP), and on the store path `connect_retry_count` / `reconnect_wait_time` / `host_alive_detect_duration`.

**What the two commits change** (both confined to `include/ylt/coro_io/client_pool.hpp`):
- `ddffb09` adds a `Signal` so the background idle-connection reaper coroutine stops cleanly on pool destruction (Terminate emitted in the dtor; the collector breaks when its sleep is cancelled). It is a coroutine-frame-leak fix on teardown.
- `bd7bae3` drops a redundant `co_await Yield{}` in the clear loop and switches the reaper launch from `directlyStart(..., get_global_executor())` to `sstart(...)`.

Both touch only the private background reaper (driven by `collect_time` / `idle_queue_per_max_clear_count`, which we leave at defaults) and its shutdown. Neither changes acquire/return, connection establishment, the retry/alive-detect config we set, or the RDMA socket path. Net effect for our TE and store RPC clients is a cleaner pool teardown at shutdown, with no change on the request hot path.

Caveat: this is a source and usage-surface analysis of the diffs, not a load test of the pool under churn. If you would still rather isolate the listen fix, the ylt-tag route (backport #1186 + #1190 onto the current base) is on the table.

mooncake-store/tests/wildcard_listen_test.cpp
```diff
@@ -0,0 +1,80 @@
...
+    static auto *real_getaddrinfo =
+        reinterpret_cast<GetAddrInfoFn>(dlsym(RTLD_NEXT, "getaddrinfo"));
+    if (IsWildcardAddress(node)) {
+        return EAI_NONAME;
+    }
+    // If RTLD_NEXT resolution failed (e.g. static linking), fall back to an
+    // error instead of dereferencing a null function pointer.
+    if (real_getaddrinfo == nullptr) {
+        return EAI_SYSTEM;
+    }
```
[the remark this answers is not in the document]
pjdurden: changed to EAI_FAIL for that reason, the fallback in the current head no longer implies errno is set.

Aionw:
Upgrading Yalanting is fine, but the new test is unnecessary here because this behavior is already covered by Yalanting’s own tests rather than Mooncake’s. To keep this PR focused on the Yalanting dependency bump, I’d prefer to drop the test-related changes.

pjdurden:
Makes sense, ylt's own test_acceptor.cpp already covers the wildcard and dual-stack listen. Dropped wildcard_listen_test.cpp and the CMake entry, so the PR is now just the submodule bump. Keeping the pin at 7801bc9 rather than the HTTP-only fix, since master defaults rpc_address to 0.0.0.0 and the RPC-side dual-stack listen only lands there.

Aionw (approved):

PR #2527, opened 2026-06-18
fix(metadata): make rpc_meta republish idempotent in the Python server

he-yufeng:
## What

The Python HTTP metadata server (`mooncake-wheel/mooncake/http_metadata_server.py`) rejected **every** duplicate `rpc_meta` key:

```python
if key.find("rpc_meta") != -1 and key in self.store:
    return web.Response(text='Duplicate rpc_meta key not allowed', status=400, ...)
```

So a node that re-publishes its own unchanged metadata (a normal occurrence on reconnect/retry) gets a `400` instead of a no-op success.

The C++ metadata server already handles this as idempotent — #2311 made it compare the stored value against the request body, returning `200 metadata unchanged` for an identical value and only rejecting a genuinely conflicting one. This PR mirrors that logic in the Python server:

```python
if key.find("rpc_meta") != -1 and key in self.store:
    if self.store[key] == data:
        return web.Response(text='metadata unchanged', status=200, ...)
    return web.Response(text='Duplicate rpc_meta key not allowed', status=400, ...)
```

## Test plan

Added `mooncake-wheel/tests/test_http_metadata_server.py`, which exercises `_handle_put` directly — no compiled engine, server, or network required (it only imports the pure-Python module and mocks the request). It covers the first write, an idempotent re-publish (the regression), a conflicting value (still rejected), and a non-`rpc_meta` overwrite.

```
python -m unittest mooncake-wheel.tests.test_http_metadata_server
# Ran 4 tests — OK
```

The idempotent-republish case fails on `main` (`400 != 200`) and passes with this change. The test is engine-free and could be wired into CI alongside `test_safetensor_functions` if desired.

PR #2924, opened 2026-07-15
[Bugfix] Fail fast when the Python HTTP metadata server cannot bind its port

toffee-desuwa:
## Description

The Python HTTP metadata server (`mooncake-wheel/mooncake/http_metadata_server.py`) started its aiohttp server on a background daemon thread, then unconditionally logged "started" and slept forever without waiting for the bind to succeed. If the port was already in use, the background thread failed the bind (logging the error and setting internal state to `Failed`), but the main thread had already reported success and kept the process alive. A supervisor then sees a live process plus a success log while nothing is listening — a silent failure that misleads health checks and deployment.

This adds a startup handshake: `run()` waits on a `concurrent.futures.Future` that the server thread completes only after `TCPSite.start()` succeeds (ready) or fails (exception). "started" is logged only on success; on bind failure the exception is propagated to `run()` and re-raised, so the CLI exits non-zero instead of masking the failure.

This is a source-inspection find with no existing issue. Note: it touches the same file as #2527 but a different region — this changes the startup path around `run()`/`_run_server`, while #2527 changes `_handle_put` idempotency. The two do not overlap.

The added lines pass `ruff` and `codespell`. The file predates the repo's `ruff-format` hook, so the diff keeps the surrounding style unchanged rather than adding an unrelated whole-file reformat.

## Module

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

**Test commands:**
```bash
python -m pytest mooncake-wheel/tests/test_http_metadata_server_startup.py
```

**Test results:**
- [x] Unit tests pass (3/3)

New regression tests: occupying a loopback port asserts `run()` raises and logs no success message; binding to port `0` asserts normal startup and clean shutdown. A separate test confirms a runtime exception raised during startup reaches `run()` without being swallowed. Verified on Windows 11 (`WinError 10048`); the logic is platform-independent (Linux `Errno 98`).

## Checklist

## AI Assistance Disclosure

The implementation was AI-assisted and independently reviewed across multiple models before submission. The diagnosis and fix were verified against the source and the regression tests were run locally (results above).

### DIFF
--- a/mooncake-wheel/mooncake/http_metadata_server.py
+++ b/mooncake-wheel/mooncake/http_metadata_server.py
@@ -8,6 +8,7 @@ used by Mooncake. It can be used as an alternative to etcd for metadata storage.
 
 import argparse
 import asyncio
+from concurrent.futures import Future
 import logging
 import os
 import signal
@@ -50,8 +51,16 @@ class KVBootstrapServer:
     
     def run(self):
         """Start the server in a background thread."""
-        self.thread = threading.Thread(target=self._run_server, daemon=True)
+        startup: Future[None] = Future()
+        self.thread = threading.Thread(
+            target=self._run_server, args=(startup,), daemon=True
+        )
         self.thread.start()
+        try:
+            startup.result()
+        except BaseException:
+            self.thread.join(timeout=2)
+            raise
         logging.info(f"HTTP Metadata Server started on {self.host}:{self.port}")
         return self.thread
 
@@ -106,8 +115,12 @@ class KVBootstrapServer:
         return web.Response(text='metadata deleted', status=200,
                           content_type='application/json')
                           
-    def _run_server(self):
+    def _run_server(self, startup: Future[None]):
         """Run the server in the current thread."""
+        def notify_started():
+            if not startup.done():
+                startup.set_result(None)
+
         try:
             # Event Loop
             self._loop = asyncio.new_event_loop()
@@ -118,10 +131,19 @@ class KVBootstrapServer:
             
             site = web.TCPSite(self._runner, host=self.host, port=self.port)
             self._loop.run_until_complete(site.start())
+            self._loop.call_soon(notify_started)
             self._loop.run_forever()
-        except Exception as e:
+        except BaseException as e:
+            if not startup.done():
+                startup.set_exception(e)
+            if not isinstance(e, Exception):
+                raise
             logging.error(f"Server error: {str(e)}")
         finally:
+            if not startup.done():
+                startup.set_exception(
+                    RuntimeError("Server exited before startup completed")
+                )
             # Cleanup
             if self._runner is not None:
                 self._loop.run_until_complete(self._runner.cleanup())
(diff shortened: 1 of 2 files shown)


stmatengss:
The Python-based HTTP metadata server is rarely used. Does the built-in C++ metadata server have the same issue?

toffee-desuwa:
Yeah, the C++ one has the same issue. `async_start()` does the bind synchronously and hands back a `Future<error_code>` that's already resolved if it failed, but `HttpMetadataServer::start()` drops that future and sets `running_ = true` / returns `true` regardless, so an in-use port still reports a successful start. The other HTTP callers like `MasterAdminServer::Start` and `RealClient::start_http_server` check `ec.hasResult()` right after `async_start()`; the metadata server just doesn't.

That said, #2345 already seems to be fixing that exact method, so I'd rather not duplicate it — happy to keep this PR to the Python server and leave the C++ side to #2345, unless you'd prefer I fold it in here.

he-yufeng (approved):
Approving the Python side. The `Future`-based handshake is the right fix, and I walked the paths to convince myself it is tight: "started" is logged strictly after `site.start()` returns, a failed bind is re-raised to the caller instead of dying silently on the daemon thread, the `except BaseException` re-raises non-`Exception` types so `KeyboardInterrupt`/`SystemExit` keep their identity, and the `finally` + `if not startup.done()` guard both prevents a `run()` hang and avoids a double `set_exception`. The three tests cover the meaningful cases (bind-failure propagation with no false "started" log, ready server + clean shutdown, and a `BaseException` after readiness not being swallowed).

One non-blocking note: `startup.result()` has no timeout, so the only startup path this does not fail-fast on is a `site.start()` that hangs rather than raises. That is rare and not a regression, but a bounded timeout would close the last gap if you think it is worth it.

On @stmatengss's question about the C++ server and @toffee-desuwa's answer, I confirmed it. `HttpMetadataServer::start()` (`mooncake-store/src/http_metadata_server.cpp:110-118`) calls `server_->async_start()`, drops the returned `Future<error_code>`, then unconditionally sets `running_ = true`, logs "started", and returns `true`. `poll()` then reports `KVPoll::Success` off `running_`, so a failed bind is invisible on the C++ path exactly like it was on the Python one. Worth tracking so the two servers behave the same. Happy to send a follow-up that awaits the `async_start()` future and surfaces the error, or leave it to you if you would rather keep it in this line of work.

he-yufeng:
Opened #2942 for the C++ side we discussed — it mirrors `MasterAdminServer::Start()`'s `async_start().hasResult()` check for the same server type, with a bind-failure regression test.

toffee-desuwa:
Thanks for the careful review, and for taking the C++ side in #2942 — that fully covers @stmatengss's question: this PR stays Python-only, and the C++ path gets the same treatment over there. 
On the timeout note: agreed it's not a regression (a hanging site.start() would hang today too), but happy to add a bounded startup.result(timeout=...) if you'd want it in before merge.

toffee-desuwa:
@stmatengss the C++ side is merged now (#2942) — anything else needed on this one?

PR #2926, opened 2026-07-15
[Store] fix GPU-addressed local copy crashes

zxpdemonio:
## Description

Fix crashes when Mooncake Store local copy paths receive CUDA-addressed source or destination buffers.

The tensor API is one reproducer, but the affected scope is lower than tensor-only APIs. On current upstream `origin/main`, local Store copy paths can still reach host `memcpy` with CUDA pointers, which crashes for both GPU-source writes and GPU-destination reads.

This PR fixes the issue by:

- Forcing the CUDA-like accelerator device object to be linked into the Python module, so CUDA pointer copies can use the accelerator registry instead of falling back to host `memcpy`.
- Making host-staging write paths use D2H-aware gather logic:
  - `put`
  - `put_batch`
  - `put_parts`
  - `upsert`
  - `upsert_parts`
  - `upsert_batch`
- Making local MEMORY `get_into` / `get_into_ranges` avoid direct host `memcpy` into CUDA destinations by using a CPU temporary buffer plus H2D scatter.
- Keeping direct `put_from` / `batch_put_from` paths unchanged; they are covered by the CUDA accelerator registration/linkage fix.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

Verified in a clean checkout by comparing current upstream `origin/main` with `origin/main` plus this single patch.

Manual CUDA smoke cases:

```text
1. GPU-source tensor write
   - Create a CUDA tensor.
   - Write it through the Store tensor API.

2. GPU-destination read
   - Write an object from CPU memory.
   - Read it into a CUDA destination pointer.

3. Direct GPU-source put_from
   - Call put_from() with a CUDA source pointer.

4. Batched direct GPU write/read
   - Batch-write CUDA-backed objects.
   - Batch-read them into CUDA destination buffers.
```

**Test commands:**

```bash
cmake -S . -B build -G Ninja \
  -DCMAKE_BUILD_TYPE=Release \
  -DBUILD_UNIT_TESTS=OFF \
  -DBUILD_BENCHMARK=ON \
  -DBUILD_EXAMPLES=ON \
  -DCUDAToolkit_ROOT=/usr/local/cuda

cmake --build build --target store mooncake_client mooncake_master -j$(nproc)
```

**Test results:**

- [ ] Unit tests pass

Baseline on upstream `origin/main`:

```text
GPU-source tensor write:
Segmentation fault, rc=139

GPU-destination read:
Segmentation fault, rc=139
```

After applying this patch:

```text
GPU-source tensor write:
put ret 0, rc=0

GPU-destination read:
get ret 16777216, rc=0

Direct GPU-source put_from:
put_from ret 0, rc=0

Batched direct GPU write/read:
batch_put_from ret [0, 0]
batch_get_into ret [8388608, 8388608]
rc=0

git diff --check HEAD~1..HEAD:
pass
```

Build notes:

```text
Build passed. Existing third-party / LTO warnings were emitted from pybind11, async_simple, and struct_pack paths.
```

## Checklist

Notes:

- No repository test file is added in this PR; validation was done with manual CUDA smoke cases.
- This PR is below 500 LOC.

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex was used to inspect the affected Store copy paths, prepare the patch, and run manual validation. I am responsible for reviewing, understanding, and maintaining the change.

### DIFF
--- a/mooncake-store/src/device/accelerator_registry.cpp
+++ b/mooncake-store/src/device/accelerator_registry.cpp
@@ -7,6 +7,12 @@
 
 namespace mooncake {
 namespace device {
+
+#if defined(USE_CUDA) || defined(USE_MUSA) || defined(USE_MACA) || \
+    defined(USE_HYGON) || defined(USE_COREX)
+void EnsureCudaLikeAcceleratorDeviceLinked();
+#endif
+
 namespace {
 
 void RegisterStaticAcceleratorDevice(const AcceleratorDevice& device);
@@ -90,6 +96,10 @@ void RegisterStaticAcceleratorDevice(const AcceleratorDevice& device) {
 }  // namespace
 
 const AcceleratorRegistry& GetAcceleratorRegistry() {
+#if defined(USE_CUDA) || defined(USE_MUSA) || defined(USE_MACA) || \
+    defined(USE_HYGON) || defined(USE_COREX)
+    EnsureCudaLikeAcceleratorDeviceLinked();
+#endif
     return MutableRegistry();
 }
 
--- a/mooncake-store/src/device/cuda_like_accelerator_device.cpp
+++ b/mooncake-store/src/device/cuda_like_accelerator_device.cpp
@@ -8,6 +8,9 @@
 
 namespace mooncake {
 namespace device {
+
+void EnsureCudaLikeAcceleratorDeviceLinked() {}
+
 namespace {
 
 void FreeCudaLikePinnedHostBuffer(void* addr) { cudaFreeHost(addr); }
--- a/mooncake-store/src/real_client.cpp
+++ b/mooncake-store/src/real_client.cpp
@@ -279,6 +279,18 @@ inline tl::expected<void, ErrorCode> scatter_host_to_maybe_device(
     return {};
 }
 
+// Gather memory that may be GPU or host into a host destination.
+inline tl::expected<void, ErrorCode> gather_maybe_device_to_host(
+    void *dst, const void *src, size_t size, const std::string &context) {
+    auto runtime_accelerator =
+        device::GetAcceleratorRegistry().RuntimeAccelerators();
+    if (!runtime_accelerator.CopyToHost(dst, src, size)) {
+        LOG(ERROR) << "D2H copy failed: " << context;
+        return tl::unexpected(ErrorCode::TRANSFER_FAIL);
+    }
+    return {};
+}
+
 // SelectBestReplica and the replica-scoring helpers live in
 // replica_selection.h (included above) so they can be unit-tested directly.
 using mooncake::SelectBestReplica;
@@ -1758,7 +1770,7 @@ tl::expected<void, ErrorCode> RealClient::put_internal(
         return tl::unexpected(ErrorCode::INVALID_PARAMS);
     }
     auto &buffer_handle = *alloc_result;
-    auto scatter_result = scatter_host_to_maybe_device(
+    auto scatter_result = gather_maybe_device_to_host(
         buffer_handle.ptr(), value.data(), value.size_bytes(), "put:" + key);
     if (!scatter_result) {
         return tl::unexpected(scatter_result.error());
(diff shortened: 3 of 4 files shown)


he-yufeng (commented):
The direction fix is right, and the reasoning holds up: on a write the source is the caller's buffer (which may be a CUDA pointer) and the destination is the host-side Store buffer, so it has to go through `CopyToHost` (D2H), not `CopyFromHost` (H2D). `scatter_host_to_maybe_device` had the roles inverted — it assumed the Store buffer was the maybe-device side — which is exactly what let a GPU source reach a host `memcpy`. The linker-anchor (`EnsureCudaLikeAcceleratorDeviceLinked`) is a sound way to keep the self-registering TU from being GC'd out of the Python module.

One thing to close before this merges: the same fix is needed on the `upsert` family, which are the write-twins of the `put` paths you changed and carry the identical crash. They still call `scatter_host_to_maybe_device(buffer_handle.ptr(), value.data(), ...)`:

- `upsert_internal` — `real_client.cpp:3792`
- `upsert_parts_internal` — `real_client.cpp:4028`
- `upsert_batch_internal` — `real_client.cpp:4116`

Each has the exact signature of its `put_*` counterpart (host `buffer_handle.ptr()` destination, caller `value.data()` source), so a CUDA-addressed `upsert` value crashes the same way. Routing those three through `gather_maybe_device_to_host` too would make the fix complete for the write side.

For what it's worth, the read paths you left untouched are correctly left as `scatter` — `real_client.cpp:3262`, `:4667`, `:5152` all copy from a host buffer into the caller's (maybe-device) destination, so `CopyFromHost` is the right direction there.

zxpdemonio:
> The direction fix is right, and the reasoning holds up: on a write the source is the caller's buffer (which may be a CUDA pointer) and the destination is the host-side Store buffer, so it has to go through `CopyToHost` (D2H), not `CopyFromHost` (H2D). `scatter_host_to_maybe_device` had the roles inverted — it assumed the Store buffer was the maybe-device side — which is exactly what let a GPU source reach a host `memcpy`. The linker-anchor (`EnsureCudaLikeAcceleratorDeviceLinked`) is a sound way to keep the self-registering TU from being GC'd out of the Python module.
> 
> One thing to close before this merges: the same fix is needed on the `upsert` family, which are the write-twins of the `put` paths you changed and carry the identical crash. They still call `scatter_host_to_maybe_device(buffer_handle.ptr(), value.data(), ...)`:
> 
> * `upsert_internal` — `real_client.cpp:3792`
> * `upsert_parts_internal` — `real_client.cpp:4028`
> * `upsert_batch_internal` — `real_client.cpp:4116`
> 
> Each has the exact signature of its `put_*` counterpart (host `buffer_handle.ptr()` destination, caller `value.data()` source), so a CUDA-addressed `upsert` value crashes the same way. Routing those three through `gather_maybe_device_to_host` too would make the fix complete for the write side.
> 
> For what it's worth, the read paths you left untouched are correctly left as `scatter` — `real_client.cpp:3262`, `:4667`, `:5152` all copy from a host buffer into the caller's (maybe-device) destination, so `CopyFromHost` is the right direction there.

@he-yufeng Thanks, agreed on the scope. The latest revision already routes upsert_internal, upsert_parts_internal, and upsert_batch_internal through gather_maybe_device_to_host(...), matching the put-side fix.

The remaining scatter_host_to_maybe_device calls are read-side copies from host buffers into caller-provided destinations, so those are intentionally left as scatter / CopyFromHost.

I also added a small CUDA regression test covering CUDA source put_tensor, CUDA source upsert_tensor, and get_into into a CUDA destination pointer.

he-yufeng (approved):
Thanks for extending it — `upsert_internal`/`upsert_parts_internal`/`upsert_batch_internal` now route through `gather_maybe_device_to_host` alongside the three `put_*` paths, so the write side is fully covered, and the DISK read paths correctly stay on `scatter_host_to_maybe_device`. Direction fix + linker anchor both look right. LGTM.

PR #2938, opened 2026-07-15
[MUSA] Enable EP/PG extensions by switching to a PyTorch base image

yeahdongcn:
## Description

Builds the Mooncake **EP and PG extensions on MUSA** by moving the MUSA images to a base that ships PyTorch, then turning on `-DWITH_EP=ON`.

`mooncake-ep` and `mooncake-pg` are torch extensions: they compile through `torchada`'s `CUDAExtension` -> `MUSAExtension` mapping and link against the interpreter's torch. The previous base, `mthreads/musa:rc4.3.0-devel-ubuntu22.04-amd64`, carries no torch, so `-DWITH_EP=ON` was simply not possible there — which is why `release-musa.yaml` still pins `BUILD_WITH_EP: "0"` and why the MUSA EP path has never been built in CI.

The new base is `mthreads/pytorch:2.9.1.post1-py3.10-musa5.2.0-devel-ubuntu22.04-amd64`: MUSA 5.2.0 toolkit + `torch 2.9.1.post1+musa5.2.0s5000` + `torch_musa` + `torchada`. It is tagged after `pytorch/pytorch` (`2.13.0-cuda12.6-cudnn9-devel` — torch release first, accelerator token glued) with the flavour/os/arch tail of `mthreads/musa` (`rc4.3.0-devel-ubuntu22.04-amd64`); `TORCH_VERSION` and `MUSA_VERSION` together respell the wheel's own version, `2.9.1.post1+musa5.2.0s5000`.

**`docker/musa.Dockerfile`**
- Composes the base from `TORCH_VERSION` / `PYTHON_VERSION` / `MUSA_VERSION` / `UBUNTU_VERSION`, following the `ARG CUDA_VERSION` / `ARG UBUNTU_VERSION` pattern in `docker/mooncake.Dockerfile`. `BASE_IMAGE` remains overridable to swap the base outright.
- `-DWITH_EP=ON` -> builds `mooncake.ep_<torch>` / `mooncake.pg_<torch>`, staged into `build/ep_pg_staging` and injected into the wheel by `scripts/build_wheel.sh`.
- `EP_TORCH_VERSIONS` is deliberately **empty**, unlike the CUDA image which pins `"2.12.1"`. Any value makes `BuildEpExt.cmake` `pip install torch==<ver>` from `download.pytorch.org`, which would replace the MUSA torch stack with a CUDA build.
- Drops the python bootstrap (`python3`, `python3-dev`, `python3-pip`, `python-is-python3`) from both stages — the base is a PyTorch image and already provides them. `-G Ninja` and `-DPython3_EXECUTABLE` now match `docker/mooncake.Dockerfile`.

**Workflows** — `ci.yml` (`build-musa`) and `release-musa.yaml` move to the same container, so CI builds on the image users get.

### Requires torchada >= 0.1.73

torchada ports each extension's include roots to MUSA **in place**. Through 0.1.72 that also rewrote the *defined name* of a vendor header's own CUDA->MUSA mapping, collapsing

```c
#define cudaEventDisableTiming musaEventDisableTiming
```

in `mooncake-transfer-engine/include/gpu_vendor/musa.h` into `#define musaEventDisableTiming musaEventDisableTiming` — a self-reference that shadows the runtime's real `driver_types.h` value, so the symbol expands to an undeclared identifier. That broke the PG extension (EP was unaffected: its translation units never reach those flags). Fixed upstream in MooreThreads/torchada#92, released as torchada 0.1.73, which the new base image carries. No Mooncake source change was needed.

### For reviewer attention: `release-musa.yaml` gains nothing, and loses wheels

`release-musa.yaml` moves to the new container **and** narrows `python-version` from `['3.9', '3.10', '3.11', '3.12', '3.13']` to `['3.10']`, so the published `mooncake-transfer-engine-musa` wheels would cover **py3.10 only**.

Worth a second look, because that job does not appear to benefit from the new base:

- It does **not** build EP — it keeps `BUILD_WITH_EP: "0"` and passes no `-DWITH_EP=ON` — so it never needs torch.
- It provisions its own interpreter via `actions/setup-python@v5` and then resolves `PYTHON_BIN="$(command -v python${{ matrix.python-version }})"`. That resolves to the **setup-python toolcache** interpreter, which is ahead of the container's `/usr/bin` on `PATH` — so the build runs on a torch-less python regardless of the container.

So the release job would carry a ~20 GB torch image in order to build with a different, torch-less python, while dropping four python versions from the release. The MUSA torch wheels are `cp310`-only, which is what makes a torch-bearing base python-specific — but that constraint only binds the EP-capable image, not this job.

Suggest keeping `release-musa.yaml` on the torch-less `mthreads/musa` base with the full matrix, and moving only `docker/musa.Dockerfile` + `ci.yml` (which does need torch, to build EP). Happy to split it that way — flagging rather than assuming, in case the narrowing is intentional for other reasons.

## Module

- [x] CI/CD
- [x] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [x] Other

## How Has This Been Tested?

Built from a clean tree against the new base and exercised the resulting image. Sources are unmodified upstream — the EP/PG build works with no local patches.

**Test commands:**
```bash
# Build (base resolved from the ARGs, no --build-arg needed)
docker build -f docker/musa.Dockerfile -t mooncake-musa:ep .

# Exercise the extensions the image now ships
docker run --rm mooncake-musa:ep python -c "import mooncake.ep, mooncake.pg"
```

**Test results:**
```
-- [EP] Building with currently-installed PyTorch
-- [EP] Staging ep_2_9_1.cpython-310-x86_64-linux-gnu.so -> /workspace/build/ep_pg_staging
-- [EP] Mooncake EP extension build complete
-- [PG] Staging pg_2_9_1.cpython-310-x86_64-linux-gnu.so -> /workspace/build/ep_pg_staging
-- [PG] Mooncake PG extension build complete
...
PASS import mooncake.ep, mooncake.pg
python 3.10.12 | torchada 0.1.73 | torch.version.musa 50200
```

Also verified: `ep_2_9_1.so` / `pg_2_9_1.so` land in the installed wheel and link against the MUSA torch stack (`libtorch_cpu`, `libmusa_python`, `libmusart`, `libmusa`) plus `engine.so` via `RPATH=$ORIGIN`; `nvlink_allocator.so` is built via `--use-mcc`; `--build-arg` on any of `TORCH_VERSION` / `PYTHON_VERSION` / `MUSA_VERSION` / `UBUNTU_VERSION` / `BASE_IMAGE` re-resolves the base as expected.

- [ ] Unit tests pass

**Not covered:** a functional EP/collective run on real hardware. The build machine has no MTT GPU, so this validates that EP/PG compile, install and import — not that they work end to end on an S5000. That still needs a run on real hardware before anyone relies on MUSA EP.

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

Claude Code assisted with this change: diagnosing why `-DWITH_EP=ON` failed on the old base, root-causing the torchada in-place-porting bug behind the PG compile failure (fixed upstream in MooreThreads/torchada#92 rather than worked around here), drafting the Dockerfile ARG composition, and running the build/import verification quoted above. Every line has been reviewed by the human submitter, who is responsible for defending the change.


yeahdongcn:
The Docker image is still being pushed. I'll mark this PR as `Ready for review` once that's finished.

UNIDY2002 (approved):
Thank you!

PR #2939, opened 2026-07-15
[transport/nvmeof] surface terminal slice failures

### LINKED ISSUE #2916: NVMe-oF can indefinitely defer a known failure while another slice remains active
## Problem

The NVMe-oF direct-batch path waits for every submitted slice to reach a terminal state before reporting a terminal task result.

This is necessary for descriptor safety: a cuFile batch handle cannot be returned to the pool while another operation may still reference its parameters. However, it also means a known terminal failure is reported as `WAITING` or `PENDING` while any sibling slice remains active.

For example:

```text
slice states: [FAILED, WAITING]
task status:  WAITING
```

If the active slice does not terminalize, the caller cannot observe the already-known failure and may poll indefinitely.

This behavior was identified while reviewing #2893. That PR intentionally preserves wait-for-all semantics to avoid recycling an in-flight cuFile descriptor. Returning `FAILED` immediately without changing the cleanup lifecycle would not be safe: existing callers stop polling on a terminal status, while `freeBatchID()` would still reject the unfinished task as busy.

## Evidence

- `aggregateTransferStatus()` replaces a terminal failure with `WAITING` or `PENDING` whenever `is_finished` is false.
- The failure event remains cached and is reported after all slices terminalize; it is deferred rather than lost.
- NVMe-oF callers commonly stop polling when they receive `FAILED` and then call `freeBatchID()`.
- `freeBatchID()` requires every task to be marked finished.
- `CUFileDescPool::freeCUfileDesc()` returns the handle to the pool immediately and therefore requires all I/O to be complete.
- NVIDIA documents `cuFileBatchIOCancel()` as best-effort and does not guarantee cancellation of an ongoing operation.
- The timeout argument to `cuFileBatchIOGetStatus()` bounds that polling call; it is not documented as an eventual-completion guarantee for submitted operations.

## Desired behavior

A batch with a known terminal failure should reach a bounded, observable outcome without recycling a cuFile handle that may still have in-flight operations.

The transport should distinguish between:

1. the transfer outcome being known; and
2. the descriptor being safe to release or reuse.

## Success criteria

- A terminal slice failure is retained independently of later polling order.
- A sibling operation that remains active cannot cause unbounded polling without a defined recovery path.
- Remaining operations are canceled or drained when possible.
- A descriptor is returned to the reusable pool only after it is proven safe.
- A descriptor whose state cannot be safely drained is not reused.
- Callers receive a stable terminal result after cleanup reaches its defined outcome.
- Tests cover:
  - failure followed by successful draining;
  - failure with pending sibling slices;
  - cancellation success;
  - cancellation failure or incomplete cancellation;
  - descriptor reuse only after safe completion.

## Related

- #2890
- #2893
- [NVIDIA cuFile batch API documentation](https://docs.nvidia.com/gpudirect-storage/api-reference-guide/index.html#cufile-batch-api-functional-specification)

zxpdemonio:
## Description

Fixes https://github.com/kvcache-ai/Mooncake/issues/2916.

NVMe-oF direct-batch status aggregation previously waited for every submitted slice to reach a terminal state before reporting a terminal task result. This meant a known failure could be hidden by another active slice, for example `[FAILED, WAITING] -> WAITING`, causing callers to keep polling indefinitely.

This PR separates the known transfer outcome from descriptor reuse safety:

- Preserve terminal slice failures during aggregation even when sibling slices are still `WAITING` or `PENDING`.
- Try to cancel the remaining cuFile batch work once a terminal failure is known.
- If the batch still has active slices after cancellation, mark the descriptor as not reusable so its cuFile handle and IO parameter storage are not returned to the reusable pool.
- Cache the terminal task result so repeated status queries remain stable.
- Keep normal `PENDING/WAITING` behavior when no terminal failure is known.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
cmake -S . -B $mooncake_dir \
  -DCMAKE_BUILD_TYPE=Release \
  -DBUILD_UNIT_TESTS=ON \
  -DBUILD_EXAMPLES=OFF \
  -DBUILD_BENCHMARK=OFF \
  -DWITH_STORE=OFF \
  -DWITH_TE=ON \
  -DWITH_STORE_RUST=OFF \
  -DWITH_EP=OFF \
  -DUSE_CUDA=ON \
  -DUSE_NVMEOF=ON \
  -DUSE_ETCD=OFF

ctest -R nvmeof_status_test --output-on-failure

### DIFF
--- a/mooncake-transfer-engine/include/transport/nvmeof_transport/cufile_desc_pool.h
+++ b/mooncake-transfer-engine/include/transport/nvmeof_transport/cufile_desc_pool.h
@@ -74,6 +74,12 @@ class CUFileDescPool {
     // Get transfer status for a specific slice
     CUfileIOEvents_t getTransferStatus(int idx, int slice_id);
 
+    // Poll cuFile once for the batch and update cached completion events.
+    bool updateBatchStatus(int idx);
+
+    // Get cached transfer status for a specific slice.
+    CUfileIOEvents_t getCachedTransferStatus(int idx, int slice_id);
+
     // Best-effort cancellation for a submitted batch.
     bool cancelBatch(int idx);
 
@@ -92,8 +98,14 @@ class CUFileDescPool {
    private:
     static bool cachePolledEvent(std::vector<CUfileIOEvents_t>& io_events,
                                  const CUfileIOEvents_t& event);
+    static bool isTerminalStatus(CUfileStatus_t status);
+    static CUfileIOEvents_t failedEvent();
+    static void destroyDesc(CUFileBatchDesc* desc);
+    static bool updateBatchStatus(CUFileBatchDesc* desc, int idx);
 
     static const size_t MAX_NR_DESC = 256;  // Max number of descriptors
+    void cleanupQuarantinedDescs();
+
     size_t max_batch_size_;
 
     // Object pool for BatchHandle to avoid frequent cuFileBatchIOSetUp/Destroy
--- a/mooncake-transfer-engine/include/transport/nvmeof_transport/cufile_desc_pool.h
+++ b/mooncake-transfer-engine/include/transport/nvmeof_transport/cufile_desc_pool.h
@@ -47,6 +47,7 @@ struct CUFileBatchDesc {
     // so its output cannot be treated as a positional status snapshot.
     std::vector<CUfileIOEvents_t> io_events;
     std::vector<CUfileIOEvents_t> polled_events;
+    bool reusable = true;
 };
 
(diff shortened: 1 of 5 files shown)


LujhCoconut (approved):
LGTM.

PR #2941, opened 2026-07-15
[Bugfix][TransferEngine] Pause active reconnects to failed peers

Dao007forever:
## Description

After an RDMA endpoint is torn down, pending transfers can immediately try to recreate it. If the peer is unavailable, each active handshake blocks a posting worker for the TCP connection timeout and delays unrelated transfers assigned to that worker.

Add an opt-in, per-context connection pause keyed by peer server address. Deleting an endpoint arms MC_CONN_PAUSE_TTL_MS for that peer, and active connection setup fails fast until the pause expires. Expired entries are removed lazily during lookup and periodically by the monitor worker.

Resolve the deleted peer path inside EndpointStore under its lock using pointer identity. This avoids dereferencing a potentially stale endpoint pointer delivered by an asynchronous QP fatal event.

The feature remains disabled by default with a zero TTL. Add hardware-independent tests for tracker expiry, pruning, concurrency, and configuration parsing.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

Ran in production for ~1 month.
Unit tests

**Test results:**
- [X] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used
- [X] AI tools were used (specify below)

Claude Code.

### DIFF
--- a/mooncake-transfer-engine/include/transport/rdma_transport/connect_pause_tracker.h
+++ b/mooncake-transfer-engine/include/transport/rdma_transport/connect_pause_tracker.h
@@ -1,0 +1,85 @@
+// Copyright 2024 KVCache.AI
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef CONNECT_PAUSE_TRACKER_H
+#define CONNECT_PAUSE_TRACKER_H
+
+#include <cstdint>
+#include <functional>
+#include <mutex>
+#include <string>
+#include <unordered_map>
+
+namespace mooncake {
+
+// Active-connect circuit-breaker state: per peer server name (peer IP), a
+// "pause active reconnection until" timestamp. After an endpoint to a peer is
+// torn down (path failure / QP fatal), the peer is paused so the CQ poller is
+// not blocked re-handshaking a likely-gone peer. The clock is injected so the
+// TTL/expiry/prune logic is unit-testable without RDMA hardware or real sleeps.
+// All methods are thread-safe.
+class ConnectPauseTracker {
+   public:
+    using Clock = std::function<uint64_t()>;  // nanosecond timestamp source
+
+    explicit ConnectPauseTracker(Clock clock) : clock_(std::move(clock)) {}
+
+    // Arm/refresh the pause for `server` until the absolute timestamp
+    // `until_ns`.
+    void pause(const std::string &server, uint64_t until_ns) {
+        std::lock_guard<std::mutex> lock(mu_);
+        until_ns_[server] = until_ns;
+    }
+
+    // Whether `server` is currently paused. Expired entries are deleted on
+    // access (lazy expiry).
+    bool isPaused(const std::string &server) {
+        std::lock_guard<std::mutex> lock(mu_);
+        auto it = until_ns_.find(server);
+        if (it == until_ns_.end()) return false;
+        if (clock_() >= it->second) {
+            until_ns_.erase(it);  // expired -> delete
+            return false;
+        }
+        return true;
+    }
+
+    // Drop all expired entries (so the map doesn't grow for peers that are
+    // never re-checked after their pause lapses). Intended for a periodic tick.
+    void prune() {
+        uint64_t now = clock_();
+        std::lock_guard<std::mutex> lock(mu_);
+        for (auto it = until_ns_.begin(); it != until_ns_.end();) {
+            if (now >= it->second)
+                it = until_ns_.erase(it);
+            else
+                ++it;
+        }
+    }
+
+    // For tests / diagnostics.
+    size_t size() {
+        std::lock_guard<std::mutex> lock(mu_);
+        return until_ns_.size();
+    }
+
+   private:
+    const Clock clock_;
+    std::mutex mu_;
+    std::unordered_map<std::string, uint64_t> until_ns_;
+};
+
+}  // namespace mooncake
+
+#endif  // CONNECT_PAUSE_TRACKER_H
--- a/mooncake-transfer-engine/include/transport/rdma_transport/connect_pause_tracker.h
+++ b/mooncake-transfer-engine/include/transport/rdma_transport/connect_pause_tracker.h
@@ -1,4 +1,4 @@
-// Copyright 2024 KVCache.AI
+// Copyright 2026 KVCache.AI
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -35,11 +35,12 @@ class ConnectPauseTracker {
 
     explicit ConnectPauseTracker(Clock clock) : clock_(std::move(clock)) {}
 
-    // Arm/refresh the pause for `server` until the absolute timestamp
-    // `until_ns`.
-    void pause(const std::string &server, uint64_t until_ns) {
+    // Arm/refresh the pause for `server` for `duration_ns`. The deadline is
+    // derived from the injected clock so production and tests use one time
+    // source.
+    void pauseFor(const std::string &server, uint64_t duration_ns) {
         std::lock_guard<std::mutex> lock(mu_);
-        until_ns_[server] = until_ns;
+        until_ns_[server] = clock_() + duration_ns;
     }
 
     // Whether `server` is currently paused. Expired entries are deleted on
--- a/mooncake-transfer-engine/src/config.cpp
+++ b/mooncake-transfer-engine/src/config.cpp
@@ -14,6 +14,7 @@
 
 #include "config.h"
 
+#include <charconv>
 #include <cstring>
 #include <cstdio>
 #include <cstdlib>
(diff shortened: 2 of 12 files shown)


mooncake-transfer-engine/include/transport/rdma_transport/connect_pause_tracker.h
```diff
@@ -0,0 +1,85 @@
+// Copyright 2024 KVCache.AI
```
staryxchen:
```suggestion
// Copyright 2026 KVCache.AI
```

mooncake-transfer-engine/src/transport/rdma_transport/rdma_context.cpp
```diff
@@ -648,9 +650,41 @@ int RdmaContext::deleteEndpoint(const std::string &peer_nic_path) {
...
+    //
+    // Resolve the peer path *inside* the store, under its lock: the raw pointer
+    // may already be freed (e.g. an IBV_EVENT_QP_FATAL racing endpoint
+    // destruction), so we must not dereference it here. The store only compares
+    // pointer identity and returns the path from the live map key, and we arm
+    // the pause only if the endpoint was actually found. No-op when TTL is 0.
+    std::string deleted_peer_nic_path;
+    int ret = endpoint_store_->deleteEndpointByPtr(endpoint_ptr,
+                                                   &deleted_peer_nic_path);
+    if (!deleted_peer_nic_path.empty()) pauseConnect(deleted_peer_nic_path);
```
staryxchen, line 736: **Pause window is refreshed on every failed attempt — does this match the intended semantics?**

Tracing the call chain, every pause-induced failure re-arms the TTL:

```
setupConnectionsByActive()
  → isConnectPaused() == true → ERR_ENDPOINT     // rdma_endpoint.cpp
  → handlePathFailure(peer, endpoint)             // worker_pool.cpp
    → deleteEndpointByPtr(endpoint)               // worker_pool.cpp
      → pauseConnect(peer)  ← re-arms now+TTL     // (this line)
```

So under sustained traffic to a peer, the pause is refreshed every tick and **never actually expires**. Combined with the rail-pause cycle (`kRailErrorThreshold=5`, `kRailPauseNs=1s`), as long as `conn_pause_ttl_ms` > ~1s (e.g. the typical 5s), the pause stays armed indefinitely while traffic flows.

This is fine — even beneficial — for the k8s rolling-restart case (old podIP:port is permanently dead). But for a same-address restart (host reboot, fixed-IP deployment), the peer is never retried while traffic continues, which seems to contradict the "fails fast until the pause expires" wording in the description and `config.h` comment.

Two questions:
1. Is this self-extending behavior intentional (TTL as a *minimum* idle pause rather than a hard retry deadline)?
2. If so, would it be worth clarifying in the `config.h` comment so operators don't expect a short TTL to mean "retry after N ms"? If not, one option is to have `handlePathFailure`-triggered deletions skip the refresh, so only the original teardown (QP fatal / path failure) arms the pause.

mooncake-transfer-engine/src/transport/rdma_transport/rdma_context.cpp
```diff
@@ -161,6 +161,8 @@ std::string gidBytesToString(const uint8_t *raw) {
 RdmaContext::RdmaContext(RdmaTransport &engine, const std::string &device_name)
     : device_name_(device_name),
       engine_(engine),
+      connect_pause_(
+          [] { return static_cast<uint64_t>(getCurrentTimeInNano()); }),
```
staryxchen, line 165: **Clock source is `CLOCK_REALTIME` — acceptable for a duration-based pause?**

`connect_pause_` is clocked via `getCurrentTimeInNano()` → `clock_gettime(CLOCK_REALTIME)`. For a TTL/duration semantic, `CLOCK_MONOTONIC` is the usual pick — `REALTIME` can jump on NTP step adjustments or leap seconds, and since `until_ns` is stored as an absolute wall-clock deadline, a backward jump extends the pause while a forward jump shortens it.

I see this is the project-wide convention (the rail pause in `worker_pool` and the handshake timeouts use the same clock), so switching just this tracker would be inconsistent. Raising as a heads-up rather than a fix request — is this a known acceptable tradeoff, or worth a TODO for a project-wide monotonic helper?

mooncake-transfer-engine/src/transport/rdma_transport/rdma_endpoint.cpp
```diff
@@ -344,6 +344,16 @@ int RdmaEndPoint::setupConnectionsByActive() {
             return ret;
         }
 
+        // Active-connect circuit-breaker: if we recently tore down an endpoint
+        // to this peer (path failure / QP fatal), skip the handshake for the
+        // configured pause window instead of blocking the posting worker on
+        // an RPC to a likely-gone peer. Fail fast; performPostSend then
+        // redispatches or fails the (not-yet-posted) slices. No-op when
+        // conn_pause_ttl_ms is 0.
+        if (context_.isConnectPaused(peer_nic_path_)) {
```
staryxchen: This check runs inside the endpoint's `RWSpinlock::WriteGuard`, but `isConnectPaused` takes a `std::mutex` and parses the server name — composing a spinlock write-guard with a blocking mutex briefly stalls other waiters on this endpoint. Could the check move outside the write-guard (e.g. read `peer_nic_path_` first, or cache `server_name` on the endpoint since it's stable after init)?

mooncake-transfer-engine/src/transport/rdma_transport/rdma_context.cpp
```diff
@@ -648,9 +650,41 @@ int RdmaContext::deleteEndpoint(const std::string &peer_nic_path) {
...
+    return ret;
 }
 
+void RdmaContext::pauseConnect(const std::string &peer_nic_path) {
+    int ttl_ms = globalConfig().conn_pause_ttl_ms;
+    if (ttl_ms <= 0) return;  // disabled
+    auto server_name = getServerNameFromNicPath(peer_nic_path);
+    if (server_name.empty()) return;
+    uint64_t until =
+        getCurrentTimeInNano() + static_cast<uint64_t>(ttl_ms) * 1000000ull;
```
staryxchen: The `until` deadline is computed with a direct `getCurrentTimeInNano()` call here, bypassing the tracker's injected `Clock` — which somewhat defeats the testability `ConnectPauseTracker::Clock` was designed for. Consider a `pauseFor(server, ttl_ms)` API that computes `until` inside the tracker from its own clock, so the two never diverge.

mooncake-transfer-engine/src/config.cpp
```diff
@@ -393,6 +393,28 @@ void loadGlobalConfig(GlobalConfig& config) {
...
     }
 
+    const char* conn_pause_ttl_env = std::getenv("MC_CONN_PAUSE_TTL_MS");
+    if (conn_pause_ttl_env) {
+        // Robust parse (not atoi): a non-numeric typo must keep the default
+        // rather than silently resolve to 0. 0 is a valid explicit "disable";
+        // negative / out-of-range / garbage are rejected, preserving the
+        // default.
+        try {
+            int val = std::stoi(conn_pause_ttl_env);
```
staryxchen: `std::stoi` parses a numeric prefix, so `"5000s"` silently becomes `5000`, contradicting the "non-numeric typo must keep the default" intent in the comment above. Consider validating the consumed length (`pos == strlen(env)`) or using `std::from_chars` for strict whole-string parsing.

alogfans:
Perhaps this problem may be covered by #2959? @Dao007forever may you have a try?

mooncake-transfer-engine/src/transport/rdma_transport/rdma_context.cpp
Dao007forever: Updated to not re-arm.

Dao007forever:
@alogfans #2959 partially overlaps. It pauses a failed rail when another peer rail is available, but explicitly retries without pausing when no alternative exists. Therefore it does not cover the current case of a dead single-address/pod endpoint case, where each handshake can block a posting worker.

It is complimentary in case a NIC goes bad, where requests can use another NIC. In our case, the whole pod just died.

mooncake-transfer-engine/src/transport/rdma_transport/rdma_context.cpp
Dao007forever: Defer this.

he-yufeng (approved):
This is a clean, well-reasoned fix — thanks for the care that went into it. A few things I checked while reviewing, for the record:

- Keying is consistent between arm and check: `pauseConnect` keys on `getServerNameFromNicPath(deleted_peer_nic_path)` where the path comes from the live store map key, and `isConnectPaused(entry.first)` keys the same way (`entry.first` is a `peer_nic_path`, same value that feeds `isRailAvailable`), so the breaker actually engages rather than arming one key and checking another.
- The disabled default stays cheap: both `pauseConnect` and `isConnectPaused` early-return on `conn_pause_ttl_ms <= 0` before any string work or lock, so the per-slice-group call added to `performPostSend` is a single int compare when the feature is off.
- The failure teardowns in this transport all go through `deleteEndpointByPtr` (worker_pool post/monitor paths and `rdma_transport.cpp` on `retired()`), so arming the pause there covers the paths that matter; the by-name `deleteEndpoint` isn't on a failure path.
- Resolving the path inside the store under the lock and never dereferencing the raw pointer is the right call for the QP-fatal-racing-destruction case, and injecting the clock keeps the tracker unit-testable without hardware. Nice.

One design question, not a blocker: `connect_pause_` lives on `RdmaContext`, so the pause is per local NIC/device. In the motivating k8s-rolling-restart case the peer is gone for every local NIC at once, but each context only learns that after it independently tears an endpoint down — so with N local devices you can still eat up to N handshake timeouts before all of them are paused. That's probably an acceptable simplification (N is small and cross-context shared state would need its own locking), but was a per-context pause the intended granularity, or worth hoisting the tracker to engine scope so one peer failure pauses all local contexts to that address? Happy either way — just flagging the scope.

Approving; the correctness and the hardware-free test coverage for the tracker and the env parse are solid.

staryxchen (approved):
LGTM. Thanks for such a solid, clear, and concise fix.

PR #2942, opened 2026-07-15
[Store] Surface HTTP metadata server bind failures in start()

he-yufeng:
## Description

`HttpMetadataServer::start()` called `server_->async_start()` and dropped its result, then unconditionally set `running_ = true`, logged "started", and returned `true`. `coro_http_server`'s `async_start()` binds synchronously and hands back a future that is already resolved (`hasResult()`) when the bind failed, so a failed bind — e.g. the port is already in use — was reported as a healthy server: `start()` returned `true` and `poll()` reported `KVPoll::Success` off `running_`.

This is the C++ counterpart of the Python bug fixed in #2924. @stmatengss asked there whether the built-in C++ metadata server had the same issue, and @toffee-desuwa confirmed it does (`async_start()` hands back an already-resolved future on failure, but `start()` dropped it).

## Fix

Check the future the same way `MasterAdminServer::Start()` already does for the same `coro_http_server` type: on a failed bind, log and return `false`, leaving `running_` unset so `poll()` reports `KVPoll::Failed`.

## Module

mooncake-store (metadata server)

## Type of change

Bug fix.

## Testing

Added `HttpMetadataServerTest.StartReportsBindFailure`: it starts one server on a free port, then starts a second server on the same (now-taken) port and asserts `start()` returns `false` and `is_running()` is `false`.

Local-verification note: I could not run the full mooncake-store build/test locally — it pulls in Linux-only deps (NUMA/SPDK) that aren't available on my macOS box. The fix is a direct mirror of the proven `MasterAdminServer::Start()` pattern for the same server type, and the test follows the existing tests in this file; both are exercised by CI.

## Checklist

- [x] The change is scoped to the reported bug.
- [x] Follows the existing code style and the sibling `MasterAdminServer::Start()` pattern.
- [x] Adds a regression test.

## AI assistance disclosure

Assisted by Claude Code. I have reviewed every changed line and can defend the change end-to-end.

### DIFF
--- a/mooncake-store/src/http_metadata_server.cpp
+++ b/mooncake-store/src/http_metadata_server.cpp
@@ -112,7 +112,16 @@ bool HttpMetadataServer::start() {
         return true;
     }
 
-    server_->async_start();
+    // async_start() binds synchronously and hands back a future that is already
+    // resolved (hasResult()) when the bind failed; otherwise the server keeps
+    // running. Mirror MasterAdminServer::Start() so a failed bind is surfaced
+    // instead of reporting a healthy server that never came up.
+    auto ec = server_->async_start();
+    if (ec.hasResult()) {
+        LOG(ERROR) << "Failed to start HTTP metadata server on " << host_ << ":"
+                   << port_;
+        return false;
+    }
     running_ = true;
     LOG(INFO) << "HTTP metadata server started on " << host_ << ":" << port_;
     return true;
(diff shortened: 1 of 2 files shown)


Aionw (approved):

Aionw:
The CI failure is triggered by this PR, but it exposes an existing port conflict in the test rather than a problem with the fix itself.

The workflow starts `mooncake_http_metadata_server` on port `8080`. Later, `test_cli.py` launches `mooncake_master` with `--enable_http_metadata_server=true` but does not specify a different HTTP metadata server port, so it also attempts to bind to `8080`.

Before this PR, that bind failure was incorrectly reported as a successful startup. With this fix, the failure is now correctly detected, causing the CLI test to fail.

We should keep the fix and update `test_cli.py` to pass a different port, for example:

```text
--http_metadata_server_port=61353
```

Failing job: https://github.com/kvcache-ai/Mooncake/actions/runs/29448818402/job/87791915941

PR #2945, opened 2026-07-16
[Store] pin host segments with quota

zxpdemonio:
## Description

This PR adds opt-in CUDA pinned-memory registration for Mooncake Store-managed host segment memory.

The pinned memory is limited to Store segment memory only. It does not pin client local buffers, staging buffers, dummy-client SHM buffers, user-provided buffers, file-backed `mountSegment()` mappings, CXL/device segments, or tensor-wrapper temporary buffers.

The feature is disabled by default and controlled by a process-wide quota:

- `MC_STORE_PIN_MEMORY_MAX_BYTES=<bytes>` enables pinned registration up to the configured quota.
- unset, empty, `0`, or invalid values disable the feature.
- `MC_STORE_PIN_MEMORY=0|false|off|no` explicitly disables the feature.

When enabled, Store host segments are registered with `cudaHostRegister()` after allocation and before the segment is mounted. Registration failures fall back to pageable memory so Store startup and allocation remain usable.

This PR is intentionally scoped to Store segment memory. GPU-destination reads still copy through the client local buffer / staging path before the final CPU -> GPU copy, so `get` only sees a smaller improvement. Removing or pinning that final staging path is follow-up work and is not part of this PR.

The tensor wrapper path is also not the primary target of this PR. `put_tensor(cuda)` currently goes through `put_parts`, which still stages through the client local buffer before writing into Store. Optimizing tensor / `put_parts` to avoid that staging copy is also follow-up work.

This PR is split from https://github.com/kvcache-ai/Mooncake/pull/2598.

## Module

- Store

## Type of Change

- Performance improvement
- Documentation update

### Manual performance validation

Manual benchmark on one NVIDIA A10 GPU, local TCP Store setup, 4 GiB Store segment, 2 GiB client local buffer.

Baseline:

- `MC_STORE_PIN_MEMORY_MAX_BYTES=0`

Pinned Store segment:

- `MC_STORE_PIN_MEMORY_MAX_BYTES=4294967296`

The benchmark directly exercised the lower-level Store APIs with CUDA source/destination buffers.

| Direction | Size | Pageable | Pinned Store segment |
| --- | ---: | ---: | ---: |
| put from CUDA buffer | 64 MB | 8.887 GB/s | 24.960 GB/s |
| put from CUDA buffer | 256 MB | 9.034 GB/s | 25.686 GB/s |
| put from CUDA buffer | 512 MB | 9.003 GB/s | 25.823 GB/s |
| get into CUDA buffer | 64 MB | 5.295 GB/s | 5.955 GB/s |
| get into CUDA buffer | 256 MB | 5.232 GB/s | 6.019 GB/s |
| get into CUDA buffer | 512 MB | 5.290 GB/s | 6.044 GB/s |

The main improvement is on GPU-source writes, where the Store segment is the host destination of the device-to-host copy.

Read-side GPU destinations improve less because the current read path still stages through the client local buffer before the final CPU -> GPU copy. Optimizing that staging path is follow-up work.

## Checklist

- [x] I have performed a self-review of my code.
- [x] I have run clang-format / formatting checks.
- [x] I have updated the documentation.
- [ ] I have added or updated committed tests.
- [ ] I have run pre-commit checks.
- [x] The change is scoped to Store segment memory and does not pin unrelated buffers.

## AI Assistance Disclosure

- [x] AI tools were used for implementation, review, benchmarking, and PR description preparation.

### DIFF
--- a/mooncake-store/src/registered_pinned_memory.cpp
+++ b/mooncake-store/src/registered_pinned_memory.cpp
@@ -1,13 +1,12 @@
 #include "registered_pinned_memory.h"
 
-#include <cctype>
-#include <charconv>
 #include <cstdlib>
-#include <optional>
 #include <string>
-#include <system_error>
+#include <string_view>
 
 #include <glog/logging.h>
+
+#include "utils/type_util.h"
 
 #if defined(USE_CUDA)
 #include <cuda_runtime_api.h>
@@ -16,51 +15,31 @@
 namespace mooncake {
 namespace {
 
-std::pair<const char*, const char*> TrimAsciiWhitespace(const char* value) {
-    const char* begin = value;
-    while (std::isspace(static_cast<unsigned char>(*begin))) {
-        ++begin;
-    }
-
-    const char* end = begin;
-    while (*end != '\0') {
-        ++end;
-    }
-    while (end > begin &&
-           std::isspace(static_cast<unsigned char>(*(end - 1)))) {
-        --end;
-    }
-
-    return {begin, end};
-}
-
-std::optional<uint64_t> ParsePinnedMemoryLimit() {
-    const char* value = std::getenv("MC_STORE_PIN_MEMORY_MAX_BYTES");
-    if (!value || value[0] == '\0') return 0;
-
-    auto [number, end] = TrimAsciiWhitespace(value);
-    if (*number == '-') {
-        LOG(WARNING) << "Invalid MC_STORE_PIN_MEMORY_MAX_BYTES='" << value
-                     << "', disabling Store segment pinning";
-        return std::nullopt;
-    }
-
-    uint64_t parsed = 0;
-    auto [ptr, ec] = std::from_chars(number, end, parsed);
-    if (ec != std::errc{} || ptr != end) {
-        LOG(WARNING) << "Invalid MC_STORE_PIN_MEMORY_MAX_BYTES='" << value
-                     << "', disabling Store segment pinning";
-        return std::nullopt;
-    }
-    return parsed;
+std::string_view TrimAsciiWhitespace(std::string_view value) {
+    constexpr std::string_view whitespace = " \t\r\n\f\v";
+    size_t begin = value.find_first_not_of(whitespace);
+    if (begin == std::string_view::npos) return {};
+    size_t end = value.find_last_not_of(whitespace);
+    return value.substr(begin, end - begin + 1);
 }
 
 std::pair<bool, uint64_t> ParsePinnedMemoryConfig() {
-    auto limit = ParsePinnedMemoryLimit();
-    if (!limit.has_value() || *limit == 0) {
+    const char* raw_value = std::getenv("MC_STORE_PIN_MEMORY_MAX_BYTES");
+    if (!raw_value || raw_value[0] == '\0') return {false, 0};
+
+    uint64_t limit = 0;
+    auto value = TrimAsciiWhitespace(raw_value);
+    if (value.empty() || !TypeUtil::ParseUint64(value, limit)) {
+        LOG(WARNING) << "Invalid MC_STORE_PIN_MEMORY_MAX_BYTES='" << raw_value
+                     << "', disabling Store segment pinning";
         return {false, 0};
     }
-    return {true, *limit};
+    return {limit != 0, limit};
+}
+
+void LogPinSkip(const std::string& owner, const char* reason, size_t size) {
+    LOG(WARNING) << "Skip cudaHostRegister for " << owner << ": " << reason
+                 << ", size=" << size;
 }
 
 #if defined(USE_CUDA)
@@ -110,11 +89,8 @@ RegisteredPinnedMemoryManager& RegisteredPinnedMemoryManager::instance() {
 }
 
 RegisteredPinnedMemoryManager::RegisteredPinnedMemoryManager()
-    : RegisteredPinnedMemoryManager(ParsePinnedMemoryConfig()) {}
-
-RegisteredPinnedMemoryManager::RegisteredPinnedMemoryManager(
-    std::pair<bool, uint64_t> config)
-    : RegisteredPinnedMemoryManager(config, DefaultPinOps()) {}
+    : RegisteredPinnedMemoryManager(ParsePinnedMemoryConfig(),
+                                    DefaultPinOps()) {}
 
 RegisteredPinnedMemoryManager::RegisteredPinnedMemoryManager(
     std::pair<bool, uint64_t> config, PinOps pin_ops)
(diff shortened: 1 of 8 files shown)


he-yufeng (commented):
Read through the whole thing — the manager and the real_client wiring. This is careful work and the correctness-sensitive parts hold up well:

- The quota check is overflow-safe (`size > limit_bytes_ || pinned_bytes_ > limit_bytes_ - size` rather than `pinned_bytes_ + size > limit_bytes_`), and the reserve-then-commit split — insert the key with a `nullptr` placeholder under the lock, `cudaHostRegister` outside the lock, then commit the pointer — keeps a single owner per `(addr, size)` since the `emplace` is exclusive. A concurrent `try_pin` for the same key fails at the placeholder insert, so `release`'s window between "mark nullptr" and "erase" can't be hijacked.
- Failure paths refund correctly: a failed `cudaHostRegister` calls `drop_reservation_locked` (placeholder removed, quota returned) and falls back to pageable, and `cudaGetLastError()` clears the sticky error. The env parse is properly guarded (`std::from_chars` + explicit negative check + trim, invalid → disabled) rather than a bare `stoull`.
- Lifetime is right: the region `shared_ptr` is held in `setup_segment_pinned_regions_` for the segment's life and released on teardown, and the pin is dropped if `MountSegment` fails. Scope is correctly limited to Store host segments.

Two things I'd like to resolve before approving, neither a correctness bug:

1. `release()` escalates a non-unloading `cudaHostUnregister` failure to `LOG(FATAL)`. For a long-running Store process, aborting on a pin-bookkeeping unregister failure is a heavy response — the alternative is `LOG(ERROR)` and leak the pin so the service survives a transient CUDA hiccup. The `cudaErrorCudartUnloading` case is already handled gracefully; is aborting on the other errors the intended contract, or would you rather degrade? (The `LOG(FATAL)` in the `try_pin` tracking-mismatch path is fine as an assert — that branch is unreachable given the single-owner invariant.)

2. There are no tests, but the second constructor taking `std::pair<bool, uint64_t> config` is an obvious seam for them. The `cudaHostRegister`/`Unregister` calls are the real barrier to full coverage, but the reservation math — over-quota returns `nullptr`, overlap returns `nullptr`, exact-key double-pin returns `nullptr`, quota is refunded after release — is the part most worth locking down, and it's the part most likely to drift later. Worth a small unit test around the manager's bookkeeping even if the CUDA path stays CI-only.

Design and the careful bookkeeping look good; happy to approve once the `FATAL` question is settled.

zxpdemonio:
> Read through the whole thing — the manager and the real_client wiring. This is careful work and the correctness-sensitive parts hold up well:
> 
> * The quota check is overflow-safe (`size > limit_bytes_ || pinned_bytes_ > limit_bytes_ - size` rather than `pinned_bytes_ + size > limit_bytes_`), and the reserve-then-commit split — insert the key with a `nullptr` placeholder under the lock, `cudaHostRegister` outside the lock, then commit the pointer — keeps a single owner per `(addr, size)` since the `emplace` is exclusive. A concurrent `try_pin` for the same key fails at the placeholder insert, so `release`'s window between "mark nullptr" and "erase" can't be hijacked.
> * Failure paths refund correctly: a failed `cudaHostRegister` calls `drop_reservation_locked` (placeholder removed, quota returned) and falls back to pageable, and `cudaGetLastError()` clears the sticky error. The env parse is properly guarded (`std::from_chars` + explicit negative check + trim, invalid → disabled) rather than a bare `stoull`.
> * Lifetime is right: the region `shared_ptr` is held in `setup_segment_pinned_regions_` for the segment's life and released on teardown, and the pin is dropped if `MountSegment` fails. Scope is correctly limited to Store host segments.
> 
> Two things I'd like to resolve before approving, neither a correctness bug:
> 
> 1. `release()` escalates a non-unloading `cudaHostUnregister` failure to `LOG(FATAL)`. For a long-running Store process, aborting on a pin-bookkeeping unregister failure is a heavy response — the alternative is `LOG(ERROR)` and leak the pin so the service survives a transient CUDA hiccup. The `cudaErrorCudartUnloading` case is already handled gracefully; is aborting on the other errors the intended contract, or would you rather degrade? (The `LOG(FATAL)` in the `try_pin` tracking-mismatch path is fine as an assert — that branch is unreachable given the single-owner invariant.)

@he-yufeng  I've updated this part to be availability-first.
I downgraded non-cudaErrorCudartUnloading cudaHostUnregister failures from LOG(FATAL) to LOG(ERROR). The cleanup is now best-effort: even if cudaHostUnregister fails, the manager still drops the region from its bookkeeping and refunds the pinned quota, so we do not keep stale raw tracking pointers after the Store segment owner is released.
cudaErrorCudartUnloading is still handled as a shutdown-only warning.
I also added a small unit test for this behavior: when the fake unregister backend returns an error, releasing the pinned region no longer aborts, and the same address range can be pinned again afterwards. This verifies that the manager state is cleaned up even on unregister failure.

> 2. There are no tests, but the second constructor taking `std::pair<bool, uint64_t> config` is an obvious seam for them. The `cudaHostRegister`/`Unregister` calls are the real barrier to full coverage, but the reservation math — over-quota returns `nullptr`, overlap returns `nullptr`, exact-key double-pin returns `nullptr`, quota is refunded after release — is the part most worth locking down, and it's the part most likely to drift later. Worth a small unit test around the manager's bookkeeping even if the CUDA path stays CI-only.

Test cases are added.

> 
> Design and the careful bookkeeping look good; happy to approve once the `FATAL` question is settled.

zxpdemonio:
@he-yufeng  Hi, I have rebased this PR onto the latest main and addressed the previous review comments.

The current CI checks are all green on the latest head. Could you please take another look when you have time? Please let me know if there are any remaining issues before approval.

If there are no other concerns, would it be possible to approve this PR?

Thanks!

mooncake-store/src/registered_pinned_memory.cpp
```diff
@@ -0,0 +1,241 @@
...
+                         << region->size_;
+        } else {
+            LOG(ERROR) << "cudaHostUnregister failed, size=" << region->size_
+                       << ", error=" << error_message
+                       << ". Continue with best-effort cleanup.";
+        }
+    }
+
+    std::lock_guard<std::mutex> lock(mutex_);
+    remove_inactive_region_locked(region->addr_, region->size_);
```
yokinoshitayoki, line 233: `cudaHostUnregister()` failure is currently treated as successful cleanup: the manager drops the region and refunds the quota, after which the backing memory may be freed. If unregistration actually failed, CUDA may still consider the range registered, leaving bookkeeping inconsistent and potentially allowing freed/reused memory to remain registered.

zxpdemonio: Thanks for catching this. I updated the failure path so a non-unloading cudaHostUnregister failure is no longer treated as successful cleanup.

Now the manager keeps the range reserved and does not refund the pinned quota on unregister failure. The RealClient free paths explicitly release the pinned region before freeing Store segment memory; if release reports an unregister failure, the backing memory is intentionally leaked instead of being freed/reused while CUDA may still consider it registered.

I also added a unit test to cover the distinction:
- cudaHostRegister failure still refunds the reservation and can be retried.
- cudaHostUnregister failure retains the reservation, so the same range/quota cannot be pinned again.

yokinoshitayoki (approved):
LGTM.

PR #2950, opened 2026-07-16
[CI] Add nightly build/release and test workflow

stmatengss:
## Description

Add a scheduled GitHub Actions workflow (`.github/workflows/nightly.yml`) that provides daily automated builds, releases, and comprehensive testing.

## Module(s) affected
- CI/CD (.github/workflows)

## Type of change
- New feature (non-breaking)

## Schedule
- Runs every night at **00:00 Beijing Time** (16:00 UTC)
- Manual trigger supported via `workflow_dispatch`

### Build (Nightly Wheels)
| Variant | Runner | Python Versions |
|---------|--------|-----------------|
| CUDA 12.8 x86_64 (with EP) | ubuntu-22.04 + manylinux2_28 | 3.10, 3.12 |
| CUDA 12.8 arm64 (SBSA) | ubuntu-22.04-arm | 3.10, 3.12 |
| CUDA 13 x86_64 (with EP) | ubuntu-22.04 + manylinux2_28 | 3.10, 3.12 |

### Publish
- Publishes all built wheels to **TestPyPI**
- Skippable via `skip_publish` workflow input
- Validates wheels with `twine check` before upload

### Nightly Test (Full Suite)
- Builds from source in **Release mode** (no ASAN, no coverage overhead)
- CTest unit tests
- Rust smoke tests and benchmarks (`mooncake-store/rust`)
- Go binding integration tests (`mooncake-store/go`)
- Full Python integration tests via `scripts/run_tests.sh`
- Optional tests enabled: SSD offload eviction, promotion-on-hit

### Failure Notification
- On failure, automatically creates a GitHub issue with label `nightly-failure`

## Required Secrets
- `TESTPYPI_API_TOKEN`: API token for TestPyPI publishing (must be configured in repo settings)

## Checklist
- [x] Workflow YAML is valid
- [x] Reuses existing `_build-wheel.yaml` for consistent builds
- [x] Tests run in Release mode for production-representative results
- [x] Manual trigger supported for ad-hoc nightly runs
- [x] Concurrency control prevents duplicate runs

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -12,7 +12,6 @@ on:
 
 permissions:
   contents: read
-  issues: write
 
 concurrency:
   group: nightly-${{ github.ref_name }}
@@ -207,7 +206,7 @@ jobs:
           sudo bash -x dependencies.sh -y
           mkdir build && cd build
           cmake -G Ninja .. \
-            -DUSE_HTTP=ON -DUSE_CXL=ON -DUSE_UB=ON -DUSE_ETCD=ON \
+            -DUSE_HTTP=ON -DUSE_CXL=ON -DUSE_UB=ON -DUSE_ETCD=ON -DUSE_CUDA=ON \
             -DSTORE_USE_ETCD=ON -DCMAKE_BUILD_TYPE=Release \
             -DBUILD_UNIT_TESTS=ON -DENABLE_SCCACHE=ON
 
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -18,6 +18,16 @@ concurrency:
   group: nightly-${{ github.ref_name }}
   cancel-in-progress: true
 
+# ── Workflow Design Notes ──────────────────────────────────────────────────────
+# 1. version-stamp computes a nightly version string for artifact naming and
+#    traceability (e.g. "0.8.0.dev20250716"). The actual wheel package version
+#    is determined by pyproject.toml inside the reusable _build-wheel.yaml.
+# 2. Build jobs use the reusable _build-wheel.yaml which does its own checkout
+#    and derives VERSION from GITHUB_REF_NAME / pyproject.toml.
+# 3. publish-testpypi uploads all nightly-* artifacts to TestPyPI.
+# 4. nightly-gate aggregates results; notify-failure files an issue on failure.
+# ──────────────────────────────────────────────────────────────────────────────
+
 env:
   SCCACHE_GHA_ENABLED: "true"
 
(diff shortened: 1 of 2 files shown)


stmatengss:
@qoder Review it

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,324 @@
...
+          DEFAULT_KV_LEASE_TTL: "500"
+          TEST_SSD_OFFLOAD_IN_EVICT: "1"
+          TEST_PROMOTION_ON_HIT: "1"
+        run: |
+          export LD_LIBRARY_PATH=${LD_LIBRARY_PATH:-}:/usr/local/lib
+          bash scripts/run_tests.sh
+
+  nightly-gate:
+    if: always()
+    needs: [publish-testpypi, nightly-test]
```
Aionw: Build failures can bypass this gate. Because `publish-testpypi` depends on all three wheel-build jobs, a failure in any build makes the publish job `skipped`. This gate does not depend on the build jobs directly and only rejects a publish result of `failure`, so it can report success and `notify-failure` will not run. Please include the build jobs in `needs` and validate their results, while distinguishing an intentional `skip_publish` from a dependency-induced skip.

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,324 @@
...
+              `**Timestamp**: ${new Date().toISOString()}`,
+              '',
+              'Please investigate the failed jobs in the workflow run linked above.',
+            ].join('\n');
+            await github.rest.issues.create({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              title,
+              body,
+              labels: ['nightly-failure'],
```
Aionw: The repository does not currently have a `nightly-failure` label. Passing a nonexistent label to the create-issue API can cause the notification step itself to fail with a validation error. Please create the label before enabling this workflow, use an existing label, or add a fallback that creates the issue without the label.

Aionw (commented):
Two additional issues in the failure-reporting path:

stmatengss:
Fixed. Waiting for CI pass.

Aionw (commented):
Re-reviewed the latest head (`2b7e2a2`). The following issues are still unresolved:

1. **Nightly version is not applied to the wheel.** `nightly_version` is only logged, while `scripts/build_wheel.sh` continues to build with the static version from `mooncake-wheel/pyproject.toml`. After the first upload, subsequent daily uploads will be skipped by `--skip-existing`, so TestPyPI will not receive updated nightly builds.
2. **Wheel-build failures bypass `nightly-gate`.** If any build job fails, `publish-testpypi` becomes `skipped`. The gate only checks for `publish == failure`, so it can pass and suppress `notify-failure` even though a build failed. The overall Actions run should still be red, but the aggregation and automatic issue notification are incorrect.
3. **The `nightly-failure` label does not currently exist in this repository.** Passing it to `issues.create` will make the notification step fail instead of creating the failure issue.

Please address these before merging. For the version fix, also note that `0.3.11.post1.devYYYYMMDD` sorts before `0.3.11.post1`; a nightly version based on the next target release would be easier for users to upgrade to.

Aionw (approved):

ykwd (approved):

PR #2954, opened 2026-07-16
[Store] Add shared thread-local random helpers

Aionw:
## Description

Add a shared thread-local random facility for Mooncake Store and migrate Store-owned production call sites to use it.

- Provide one lazily seeded `std::mt19937_64` engine per thread.
- Provide unbiased `randomIndex()` and `randomUniform()` helpers, including explicit-engine overloads for deterministic tests.
- Consolidate random selection in allocation strategies, MasterService shard/source selection, and `AutoPortBinder`.
- Add scoped `AGENTS.md` guidance so future Store code extends the shared API instead of creating local random sources.
- Leave `src/cachelib_memory_allocator/` unchanged.

No matching open issue or pull request was found.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [x] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
cmake --build /tmp/mooncake-store-thread-local-random-build \
  --target utils_test allocation_strategy_test mooncake_master -j 16
/tmp/mooncake-store-thread-local-random-build/mooncake-store/tests/utils_test \
  --gtest_filter='RandomTest.*'
ctest --test-dir /tmp/mooncake-store-thread-local-random-build \
  --output-on-failure -R '^allocation_strategy_test$'
```

**Test results:**
- [x] Unit tests pass

The six focused random tests passed, the allocation strategy suite passed, and `mooncake_master` built successfully.

## Checklist

- [x] I have formatted the touched code using the repository formatter

The repository-wide format hook rewrites unrelated baseline files; those rewrites were reverted. All remaining pre-commit hooks passed on the touched files.

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex implemented the shared random facility, migrated call sites, added tests, and ran the validation commands. Before marking this PR ready, the human submitter must review every changed line and be able to defend the change end-to-end.


he-yufeng (approved):
Nice consolidation — reads well and the correctness-sensitive parts are right. A few things I checked:

- The sampling is unbiased because it goes through `std::uniform_int_distribution` rather than `engine() % n`, and `randomIndex`/`randomUniform` reject invalid bounds (`upper_bound == 0`, `lower > upper`) instead of silently wrapping. That also quietly fixes the old call sites: `uniform_int_distribution<size_t>(0, names.size() - 1)` on an empty `names` would have underflowed to `SIZE_MAX` and indexed out of bounds; `randomIndex(names.size())` throws instead (and the callers already guard non-empty upstream, so it isn't reachable in practice).
- Seeding is an improvement over what it replaces: the per-thread engine is a `mt19937_64` seeded from a `seed_seq` of eight `random_device` draws, versus the old `std::mt19937 generator(std::random_device{}())` that fed a single 32-bit value into the engine's state. More entropy into the seed is the right call for a 64-bit engine.
- Thread-local storage keeps it race-free by construction, the explicit-engine overloads make the samplers deterministically testable, and the `AGENTS.md` note plus the `cachelib_memory_allocator` carve-out set the convention clearly.

One forward-looking nit, not a blocker: `randomUniform`'s `static_assert(std::is_integral_v<Integer>)` is looser than what `std::uniform_int_distribution` actually permits — the standard only allows `short/int/long/long long` and their unsigned forms, so instantiating with `char`, `signed/unsigned char`, `bool`, or `int8_t` compiles but is UB. Every current caller uses `int`, so nothing is wrong today, but since `random.h` is now the sanctioned shared entry point, it'd be worth tightening the assert to the permitted set (or documenting the constraint) so a future caller following the AGENTS.md rule can't reach for a narrow type and land in UB.

Approving — correctness, the seeding upgrade, and the test coverage all look good.

ykwd (approved):

Issue #2937, opened 2026-07-15
[Bug]: GPU-less real-client sidecar with standalone_storage=True fails: cudaPointerGetAttributes noise + client_expired after 10s + opaque setup_dummy error


zhaotyer:
### Bug Report

**Environment**
- Mooncake: v0.3.11.post1 (transfer-engine + mooncake-store built from source)
- SGLang: v0.5.14
- Deployment: Kubernetes, single Pod with two containers:
  - `mooncake-client` sidecar running `mooncake_client` (real client), no GPU, RDMA only
  - `sglang-test` running SGLang with `--hicache-storage-backend mooncake` + `standalone_storage=True`, `client_server_address=127.0.0.1:50052`
- Master: separate Deployment (`mooncake-master`), `--enable_offload=true`
- Model: DeepSeek-V3.2-Exp, TP=2

**Topology (why sidecar)**

SGLang is configured with `standalone_storage=True`, which makes its in-process `DummyClient` talk to a standalone `mooncake_client` (real client) over a Unix-domain abstract socket `@mooncake_client_<port>.sock` plus coro_rpc on TCP `<port>`. The abstract socket is network-namespace isolated, so the real client **must** live in the same Pod as SGLang. Putting the real client in its own Pod (the natural DaemonSet shape) makes `setup_dummy` fail with `ECONNREFUSED` on the abstract socket. The sidecar pattern is the documented workaround.

**Symptom 1 — `cudaPointerGetAttributes failed (Error code: 100 - no CUDA-capable device is detected)` on every shm registration**

The sidecar container has no GPU (only the SGLang container requests `nvidia.com/gpu`). When the SGLang dummy client registers its host-memory shm buffer via IPC, the real client calls `RegisterLocalMemory` → `getMemoryLocation` (`mooncake-transfer-engine/src/memory_location.cpp:38-55`), which unconditionally invokes `cudaPointerGetAttributes`. With no CUDA context in the sidecar this returns error 100, logs an `ERROR`-level line, and falls back to `numa_move_pages`. Registration then succeeds (`Mapped new shared memory`), so the error is non-fatal — but it fires on every buffer and reads like a real failure.

This log line is the dominant signal in the sidecar's stderr and makes triage of the *real* problem (Symptom 2) much harder. It would help if either:
[...]

PR #2955, opened 2026-07-16
[TransferEngine] Skip the CUDA pointer probe on GPU-less hosts

he-yufeng:
## Description

`getMemoryLocation()` probes `cudaPointerGetAttributes` for every registered buffer in a CUDA-enabled build. On a GPU-less host — e.g. the RDMA-only `mooncake_client` real-client sidecar in the deployment from #2937 — that call fails with "no CUDA-capable device is detected" and logs an `ERROR` for **every** buffer before falling back to the host/NUMA path. Registration still succeeds, but the per-buffer `ERROR` becomes the dominant line in the sidecar's stderr and buries the real problem (symptom 1 in #2937).

This detects device presence once via `cudaGetDeviceCount` and skips the probe entirely when no device exists, emitting a single `WARNING` instead. Hosts with a GPU keep the exact previous behavior. This addresses only symptom 1 (the diagnostic noise); the `client_expired` liveness behavior (symptoms 2–3) needs the full deployment and is out of scope here.

`cudaGetDeviceCount` is already aliased in every `gpu_vendor/*.h` shim (`musaGetDeviceCount`, `mcGetDeviceCount`, `tangGetDeviceCount`, the MLU inline, …), so the change compiles for all `USE_*` GPU backends, not just `USE_CUDA`.

Addresses #2937.

## Module

## Type of Change

- [x] Bug fix
- [x] Performance improvement

## How Has This Been Tested?

I could not build the full stack on this host (Linux-only deps), so I verified the new control flow with a standalone harness that mirrors the patched branch with fake CUDA-alike calls and a call counter:

```
# GPU-less (the fix path): probe must be skipped, warning emitted once
MOCK_DEVICES=0 -> probe_calls=0 warn_calls=1 last=host
# GPU present (no regression): probe runs on every call as before
MOCK_DEVICES=2 -> probe_calls=5 warn_calls=0 last=host
```

So a GPU-less host no longer calls `cudaPointerGetAttributes` per buffer (0 calls, one `WARNING`) and still classifies buffers via the host/NUMA path, while a host with a device keeps probing every buffer unchanged. Portability was checked by grepping each `gpu_vendor` header for the `cudaGetDeviceCount` alias. Full compile/tests rely on CI.

**Test results:**
- [ ] Unit tests pass

## Checklist

- [ ] I have formatted my code using `./scripts/code_format.sh` — clang-format was not available on this host; I hand-matched the repo `.clang-format` (Google, 80 col) and CI will verify
- [ ] I have added tests to prove my changes are effective — the CUDA-gated path has no repo unit test harness; verified with the standalone harness above
- [ ] For changes >500 LOC: I have filed an RFC issue — not applicable (30 LOC)

## AI Assistance Disclosure

Claude Code helped analyze the issue, draft the change, and build the standalone verification harness above. The change is small and was reviewed line by line: the root cause matches @Icedcoco's analysis on #2937, `cudaGetDeviceCount` portability was checked against each `gpu_vendor` header, and the branch behavior was confirmed with the harness.

### DIFF
--- a/mooncake-transfer-engine/src/memory_location.cpp
+++ b/mooncake-transfer-engine/src/memory_location.cpp
@@ -38,19 +38,37 @@ const std::vector<MemoryLocationEntry> getMemoryLocation(void *start,
 #if defined(USE_CUDA) || defined(USE_MUSA) || defined(USE_HIP) ||  \
     defined(USE_MLU) || defined(USE_MACA) || defined(USE_HYGON) || \
     defined(USE_COREX) || defined(USE_SUNRISE)
-    cudaPointerAttributes attributes;
-    cudaError_t result = cudaPointerGetAttributes(&attributes, start);
-    if (result != cudaSuccess) {
-        LOG(ERROR) << "cudaPointerGetAttributes failed (Error code: " << result
-                   << " - " << cudaGetErrorString(result) << ")" << std::endl;
-        entries.push_back({(uint64_t)start, len, kWildcardLocation});
-        return entries;
-    }
+    // A GPU-less host (e.g. an RDMA-only sidecar) has no device for
+    // cudaPointerGetAttributes to classify, yet a CUDA-enabled build probes
+    // every buffer, each call failing and logging a per-buffer ERROR that
+    // buries real logs. Detect the absence once and skip the probe.
+    static const bool cuda_device_present = [] {
+        int device_count = 0;
+        if (cudaGetDeviceCount(&device_count) != cudaSuccess ||
+            device_count == 0) {
+            LOG(WARNING) << "No CUDA device detected; treating buffers as "
+                            "host memory";
+            return false;
+        }
+        return true;
+    }();
 
-    if (attributes.type == cudaMemoryTypeDevice) {
-        entries.push_back(
-            {(uint64_t)start, len, genGpuNodeName(attributes.device)});
-        return entries;
+    if (cuda_device_present) {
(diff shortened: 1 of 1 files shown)


mooncake-transfer-engine/src/memory_location.cpp
```diff
@@ -38,19 +38,37 @@ const std::vector<MemoryLocationEntry> getMemoryLocation(void *start,
...
+    static const bool cuda_device_present = [] {
+        int device_count = 0;
+        if (cudaGetDeviceCount(&device_count) != cudaSuccess ||
+            device_count == 0) {
+            LOG(WARNING) << "No CUDA device detected; treating buffers as "
+                            "host memory";
+            return false;
+        }
+        return true;
+    }();
```
alogfans: You can extract the lambda function to make coding style simple.

he-yufeng: Done in 42078e4d, the probe now lives in detectCudaDevicePresent() above getMemoryLocation.

he-yufeng:
@alogfans while you are active on the transfer engine: this one has been sitting green for three weeks. It skips the CUDA pointer probe on GPU-less hosts so TransferEngine init no longer fails there (the probe result was only ever used to log, so skipping it changes nothing on GPU hosts). If the shape is fine, a merge or a pointed objection both work.

alogfans (approved):
Sorry for my being late. I'll merge it soon.

Issue #2953, opened 2026-07-16
[RFC]: [Store] Recoverable client liveness states and asynchronous offboarding


Doube-Suzerain:
## Motivation

Mooncake Store currently treats one client heartbeat timeout as a permanent failure. Temporary network jitter, a client pause, or a delayed Master monitor loop can therefore irreversibly unmount recoverable resources and delete their metadata.

We propose adding a recovery window: after the first TTL expires, the Master retains the client's resources but stops routing new requests and work to them; a valid liveness signal makes them serving again. Terminal cleanup starts only if the client remains unavailable for a second TTL.

Resource availability checks are currently distributed across allocation, replica lookup, Copy/Move, task dispatch, drain, offload, and promotion. This proposal also centralizes that decision so the new liveness semantics are applied consistently.

## Changes proposed

- Introduce an **Active → Suspected → Offline** state machine for each client incarnation, with separate active and suspicion TTLs.
- Add a typed `ServingEligibility` seam for all client-owned resource routing, while retaining Suspected resources for possible recovery.
- Separate client liveness from remount readiness and segment lifecycle.
- Express terminal cleanup as a retryable `ClientOffboardingJob`; complete it synchronously first, then move the same job to an asynchronous executor.
- Preserve the existing Ping RPC schema and legacy TTL configuration.
- Rebuild process-local liveness records when restoring Master snapshots; keep NoF on its independent health model.

### Client liveness

Each client incarnation has a `ClientLivenessRecord`:

```text
Active --active TTL--> Suspected --suspicion TTL--> Offline
   ^                       |
   +---- liveness signal --+
```

- **Active** resources may serve new requests.
- **Suspected** resources are retained but cannot receive new allocations, descriptors, or work. A valid liveness signal recovers the same incarnation.
- **Offline** is terminal for that incarnation and starts offboarding.

The suspicion TTL starts when the Master actually commits `Active -> Suspected`, so a delayed monitor does not shorten the recovery window. Ping and successful segment mounts are liveness signals; ordinary Put/Get traffic and task/offload/promotion heartbeats are not.

Ping remains wire-compatible: a ready Active client receives `OK`; an unknown or Offline client receives `NEED_REMOUNT`; Ping may recover a Suspected client before applying the existing remount-readiness check.

Client liveness, remount readiness (`ok_client_`), and segment lifecycle remain independent. In particular, `ok_client_` will no longer determine whether resources may serve or must be retained.

### `ServingEligibility`

[...]

PR #2957, opened 2026-07-16
[Store] Decouple HTTP metadata cleanup from MasterService

Aionw:
## Description

Decouple stale HTTP metadata cleanup from `MasterService`.

`MasterService` now publishes a typed `ClientLeaseExpiredEvent` after the expired client's resources have been reclaimed and all internal locks have been released. A standalone `HttpMetadataCleanupWorker`, owned by `WrappedMasterService`, subscribes to that event and asynchronously removes the client's `ram/` and `rpc_meta/` entries for both co-located and remote HTTP metadata servers.

This keeps HTTP backend configuration, key construction, queueing, and worker lifecycle outside the core master service while preserving the existing best-effort behavior, including no retry or shutdown-drain guarantee.

Overlap notes:
- Distinct from #2256, which revives HTTP metadata cleanup/server functionality. This PR refactors the timeout-cleanup path already present on current `main`; it does not duplicate #2256's HTTP metadata server APIs.
- Related to #2953, but intentionally narrower. It does not implement recoverable liveness states, offboarding jobs, retries, or asynchronous offboarding; the event seam can be reused by that future work.

## Module

- [ ] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Breaking change
- [x] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
cmake -S . -B /tmp/mooncake-79c4-build   -DBUILD_UNIT_TESTS=ON -DWITH_STORE_RUST=OFF -DUSE_HTTP=ON   -DCMAKE_BUILD_TYPE=Debug
cmake --build /tmp/mooncake-79c4-build   --target master_service_test http_metadata_cleanup_worker_test -j8
ctest --test-dir /tmp/mooncake-79c4-build   -R '^master_service_test$' --output-on-failure
/tmp/mooncake-79c4-build/mooncake-store/tests/http_metadata_cleanup_worker_test

cmake -S . -B /tmp/mooncake-79c4-nohttp-build   -DBUILD_UNIT_TESTS=ON -DWITH_STORE_RUST=OFF -DUSE_HTTP=OFF   -DCMAKE_BUILD_TYPE=Debug
cmake --build /tmp/mooncake-79c4-nohttp-build   --target http_metadata_cleanup_worker_test -j8
/tmp/mooncake-79c4-nohttp-build/mooncake-store/tests/http_metadata_cleanup_worker_test
```

**Test results:**
- [x] Unit tests pass
- [ ] Manual testing done
- Full `master_service_test`: passed in 327.68 seconds.
- Worker tests pass with both `USE_HTTP=ON` and `USE_HTTP=OFF`.
- Relevant non-CMake pre-commit hooks and `git diff --check` pass.
- Documentation build was attempted but `sphinx-build` is unavailable in the environment.
- The repository-wide format hooks rewrite unrelated baseline files; those unrelated edits were reverted, and touched C++ files pass the clang-format hook.

## Checklist

- [ ] For changes >500 LOC: I have filed an RFC issue (not applicable; this refactor is below the threshold excluding tests)

## AI Assistance Disclosure

- [ ] No AI tools were used

OpenAI Codex helped design and implement the event boundary and cleanup worker extraction, add tests and documentation, run validation, and draft this PR description. The human submitter must review every changed line and be able to defend the change end-to-end before marking this PR ready for review.


he-yufeng (approved):
Reviewed the decoupling end to end. The direction is good and the tricky parts are handled well.

The `HttpMetadataCleanupWorker` concurrency is solid: `Stop()` flips `running_` and clears the queue under `mutex_`, notifies, and joins, while `ThreadFunc` swaps the queue into a local batch and does the (slow) HTTP removals outside the lock; an in-flight batch finishes and unstarted work is dropped, matching the documented contract. `Enqueue` no-ops once `running_` is false, so a late lease-expiry callback after `Stop()` is safe, and the destructor's `Stop()` is idempotent. Publishing is done the right way too: expirations are buffered into `expiration_events` under the master locks and the callbacks fire afterward, and `Enqueue` only takes the worker's own `mutex_`, so there's no lock-ordering hazard even if that ever changed.

I also chased the teardown lifetime, since the callback captures a raw `worker` pointer: it's correct today because `http_metadata_cleanup_worker_` is declared before `master_service_`, so `~MasterService` runs first and joins `client_monitor_thread_` (the thread that fires `NotifyClientLeaseExpired`) before the worker is destroyed. Good that this is called out in a comment.

One thing worth hardening (agreeing with the bot, but concretely): that safety rests entirely on the declaration order plus the raw-pointer capture, and a future member reorder or a second callback-firing path would reintroduce a use-after-free silently. Two robust options: (a) give `WrappedMasterService` a destructor body that explicitly stops `master_service_`'s client monitor (or clears the lease-expired callbacks) before the members tear down, so ordering no longer matters; or (b) hold the worker as a `shared_ptr` and capture a `weak_ptr` in the callback, so `lock()` keeps it alive for the duration of the call or bails if it's gone. Non-blocking for this PR given the current order is correct and documented, but I'd take one of these as a fast follow-up.

Separately, note the async removal of `rpc_meta/<segment>` was already async before this PR (this just moves it out of `MasterService`), so no regression here; if a restarted client can re-register the same segment name inside the cleanup window, that reuse race predates this change and is a separate item.

LGTM.

Icedcoco:
Thanks for the refactor. Moving HTTP metadata configuration, key construction, queueing, and worker lifecycle out of `MasterService` makes the boundary much clearer. I have two questions about failure handling and asynchronous cleanup ordering before merging:

1. Could we handle `std::thread` construction failures as a best-effort cleanup failure? `HttpMetadataCleanupWorker::Start()` currently restores its internal state and then rethrows, while `WrappedMasterService` does not catch the exception. If thread creation fails because of temporary resource exhaustion, would that cause this optional cleanup feature to prevent the entire Master service from starting? Since `Start()` already returns `bool`, perhaps returning `false` after resetting the state would preserve the previous best-effort behavior.

2. Could we clarify how a stale cleanup event is prevented from deleting metadata published by a newer client incarnation? The event is queued after the segment and snapshot locks are released, and the worker later deletes `ram/<segment_name>` and `rpc_meta/<segment_name>` using only the segment name. If a restarted client republishes the same keys and completes `MountSegment` before the queued cleanup runs, it looks like the old event could delete the new metadata. This window previously existed for the remote asynchronous path, but this PR also makes the co-located path asynchronous. Would it make sense to keep the co-located cleanup ordered with reclamation for now, or make deletion conditional on an incarnation/version before moving it fully asynchronous?

Could you please confirm whether I am missing an ordering or ownership guarantee that rules out either case?

Issue #2914, opened 2026-07-14
[RFC]: EGM-backed DRAM Store tier over Multi-Node NVLink


neverhook:
## Summary

This RFC proposes an **EGM-backed DRAM tier for Mooncake Store** within one
Multi-Node NVLink (MNNVL) scale-up domain.

Extended GPU Memory
([EGM](https://docs.nvidia.com/cuda/cuda-programming-guide/04-special-topics/extended-gpu-memory.html))
allows GPUs to access CPU-attached memory through the NVLink/NVSwitch fabric.
This proposal uses the CUDA VMM path for EGM: Store providers allocate CPU
DRAM with `CU_MEM_LOCATION_TYPE_HOST_NUMA` and export the allocations with
`CU_MEM_HANDLE_TYPE_FABRIC`.

The allocations are mounted as existing `nvlink` Store segments. Consumers
reuse the current Fabric import and mapping path for transfers between GPU HBM
and remote EGM DRAM:

```text
Put: GPU HBM -> CUDA copy over NVLink/NVSwitch -> remote EGM DRAM
Get: remote EGM DRAM -> CUDA copy over NVLink/NVSwitch -> GPU HBM
```

No application-visible CPU staging buffer is required.

## Motivation

GB200/NVL72-class systems provide CPU-attached DRAM that GPUs can access
through Extended GPU Memory (EGM). Mooncake already supports Fabric-handle
publication, remote mapping, and CUDA copies in its `nvlink` transport, but
Store does not currently have a focused way to allocate and publish
`HOST_NUMA` DRAM as Store capacity.

Two workload directions motivate this proposal.

### Higher-bandwidth access to Store DRAM

A Store deployment can access remote DRAM through an external RDMA path.
On GB200/NVL72 systems, EGM provides another path in which GPU traffic to
CPU-attached DRAM is routed through the NVLink/NVSwitch scale-up fabric.

The NVLink fabric offers a substantially higher aggregate interconnect
bandwidth ceiling than a path constrained by individual RNICs. Reusing that
fabric for HBM-to-DRAM transfers may therefore provide higher effective
bandwidth than RDMA-based DRAM access, particularly when capacity is
distributed across multiple GPU-local host NUMA nodes.

This is a performance motivation rather than a guaranteed speedup. Effective
bandwidth depends on host-memory locality, memory-controller bandwidth, CUDA
copy behavior, and contention with other NVLink traffic. RDMA and EGM
[...]

PR #2966, opened 2026-07-16
[1/3] [TransferEngine] Add HOST_NUMA Fabric allocation core

neverhook:
## Summary

This PR provides the minimum Transfer Engine prerequisite for RFC #2914.

It adds CUDA `HOST_NUMA` Fabric allocation and closes the exact-range NVLink
registration lifecycle. It intentionally does **not** add the Store EGM pool or
the complete Provider/Consumer data path.

## Stack

This is **[1/3]** of the RFC implementation stack:

1. #2966 — HOST_NUMA/Fabric Transfer Engine prerequisite (this PR).
2. #3335 — Provider-owned EGM Store pool.
3. #3431 — direct remote-memory reads into GPU destinations.

The fork branches form this exact commit chain. GitHub still displays each
community PR against `main` because a cross-fork PR cannot use another fork
branch as its base.

### HOST_NUMA placement

- Adds `NvlinkHostNumaAllocation` for Fabric-capable CUDA `HOST_NUMA` memory.
- Validates the IPC switch and every visible GPU's VMM, Fabric, and HOST_NUMA
  VMM capabilities independently from NUMA discovery.
- Auto discovery sorts and deduplicates GPU-local NUMA IDs and validates each
  discovered node's granularity.
- Explicit placement validates only the selected node's granularity, so an
  unrelated node cannot block the override.
- Grants access to all visible GPUs in one batched `cuMemSetAccess` call.

### Retryable exact-range registration lifecycle

- Inserts a handle-free registration placeholder before retaining the CUDA
  allocation handle. A duplicate or map-allocation failure therefore cannot
  leave an untracked retained handle.
- If `cuMemRetainAllocationHandle()` fails, returns `ERR_MEMORY`, removes the
  placeholder, and does not publish metadata.
- Export failure, publication rollback, and public unregister share the same
  lock-held cleanup helper. A failed registration can be retried at the same
  address: pending cleanup finishes before a new handle is acquired, while a
  successful registration still rejects duplicates.
- Registration pins protect the allocation's VA throughout registration,
  rollback, and deferred/batch cleanup. `Release()` refuses to unmap a pinned
  range. A range being released remains unavailable for registration until
  address-free succeeds.
- Batch deletion retains handles and pins until metadata publication succeeds;
  partial cleanup retries skip already completed work.
- Metadata cleanup is recorded before external publication completes, so a
  later retry does not repeat a completed delete. `ERR_ADDRESS_NOT_REGISTERED`
  remains idempotent.
- A failed `cuMemRelease()` preserves the registration record and handle
  ownership; a later unregister retries only the remaining release stage.
- Allocation-owned release follows the same rule: ownership is cleared only
  after a successful release, and a failed call can be retried.
- Every rollback checks the release result instead of silently discarding it.
- `Create()` requires an empty output. An error may still return a partial
  owner; the caller must retain it and retry `Release()`. Destruction remains
  best-effort, not a replacement for successful explicit cleanup.

CUDA documents that these APIs may report errors from earlier asynchronous
work and does not explicitly guarantee failure atomicity. This PR deliberately
treats a non-successful release as retaining ownership so cleanup remains
retryable. That policy and its accepted ambiguity are documented in the code.
See the [CUDA Driver API virtual-memory documentation](https://docs.nvidia.com/cuda/archive/12.8.1/cuda-driver-api/group__CUDA__VA.html).

### Lifetime and tests

- Makes `OwnedRangeRegistry` a process-lifetime heap singleton to avoid static
  destruction-order hazards.
- Adds hardware-free scripted coverage for capability/placement failures,
  retain/export/publication failures, metadata retry, idempotent unregister,
  release retry, concurrent register/release, VA reuse, and partial batch cleanup.

### CI integration

- Runs the two fake-driver CTest entries in the existing `unit-tests` build
  with `USE_MNNVL=ON`, reusing the CUDA driver runtime setup.
- Removes the separate NVLink build and its path filter.
- Labels the real GPU entry `nvlink_gpu` and excludes only that entry from the
  CPU runner. Hardware and fake-driver cases use separate `GTEST_FILTER`
  environments, so they do not run twice.
- Adds an optional label exclusion to the shared CTest action. Nightly keeps
  its existing test selection when the input is omitted.

## Scope boundary

This PR does not change Store behavior, Consumer mapping/cache, wire schema,
`TransferMetadata` behavior, RDMA behavior, or Store configuration. Store EGM pooling,
the full Provider/Consumer path, and GB200/NVL72 integration validation belong
to the follow-up Store layer. GB200 validation is therefore not a merge gate
for #2966.

## Review size

- 1 logical commit.
- 12 changed files.
- +2,570/-2 lines.

## Validation

This revision consolidates the CI build and test selection; allocation,
registration, transfer, and cleanup behavior are unchanged.

- [x] clang-format 20 check on PR-changed C/C++ lines.
- [x] `git diff --check` and all applicable touched-file pre-commit hooks.
- [x] CUDA 12.8 ARM64 focused build.
- [x] `nvlink_vmm_unit` (2 tests).
- [x] CUDA-off Transfer Engine build.
- [x] CI shell-script regressions (12 tests, including 5 new CTest-action tests).
- [x] CTest JSON selection: only the real GPU entry is excluded; both
  fake-driver entries remain selected and ran successfully.
- [x] New-head complete `unit-tests` sweep and GitHub Actions CI Gate.

Local CUDA-on/off regression tests passed on the composed stack (#3431).
GitHub Actions passed the complete CTest sweep (226 entries) and CI Gate
for this revision. Both fake-driver entries ran successfully; the hardware-only
entry was excluded on the CPU runner.

Repository-wide `pre-commit run --all-files` is not a merge gate.

Related RFC: #2914

## Module

- [x] CI/CD

## Type of Change

- [x] New feature
- [x] Bug fix

## Checklist

- [x] I have formatted changed C++ lines using clang-format 20.1.8
- [x] For changes >500 LOC: RFC #2914 exists

## AI Assistance Disclosure

- [x] AI tools were used

Codex assisted with the rebase, focused lifecycle fixes and tests, scoped
validation, and PR preparation. The human submitter remains responsible for
reviewing, understanding, and defending every changed line.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -404,6 +404,33 @@ jobs:
         df -h
       shell: bash
 
+    - name: Build and test NVLink HOST_NUMA units
+      if: matrix.python-version == '3.10'
+      run: |
+        cuda_stub=/usr/local/cuda/lib64/stubs
+        export LIBRARY_PATH="$cuda_stub:${LIBRARY_PATH:-}"
+        cmake -G Ninja -S . -B build-nvlink-vmm-tests \
+          -DUSE_HTTP=ON \
+          -DUSE_CUDA=ON \
+          -DUSE_MNNVL=ON \
+          -DWITH_EP=OFF \
+          -DWITH_STORE_RUST=OFF \
+          -DBUILD_UNIT_TESTS=ON \
+          -DBUILD_EXAMPLES=OFF \
+          -DENABLE_SCCACHE=ON \
+          -DCMAKE_BUILD_TYPE=Release \
+          -DCMAKE_EXE_LINKER_FLAGS="-L$cuda_stub"
+        cmake --build build-nvlink-vmm-tests --target \
+          nvlink_host_numa_allocation_test nvlink_transport_test
+        mkdir -p "$RUNNER_TEMP/cuda-driver-stubs"
+        cp -L "$cuda_stub/libcuda.so" \
+          "$RUNNER_TEMP/cuda-driver-stubs/libcuda.so.1"
+        export LD_LIBRARY_PATH="$RUNNER_TEMP/cuda-driver-stubs:$cuda_stub:${LD_LIBRARY_PATH:-}"
+        ctest --test-dir \
+          build-nvlink-vmm-tests/mooncake-transfer-engine/tests \
+          -L nvlink_vmm_unit --output-on-failure
+      shell: bash
+
     - name: Install Rust toolchain
       uses: dtolnay/rust-toolchain@stable
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -315,6 +315,33 @@ jobs:
         df -h
       shell: bash
 
+    - name: Build and test NVLink HOST_NUMA units
+      if: needs.check-paths.outputs.nvlink == 'true'
+      run: |
+        cuda_stub=/usr/local/cuda/lib64/stubs
+        export LIBRARY_PATH="$cuda_stub:${LIBRARY_PATH:-}"
+        cmake -G Ninja -S . -B build-nvlink-vmm-tests \
+          -DUSE_HTTP=ON \
+          -DUSE_CUDA=ON \
+          -DUSE_MNNVL=ON \
+          -DWITH_EP=OFF \
+          -DWITH_STORE_RUST=OFF \
+          -DBUILD_UNIT_TESTS=ON \
+          -DBUILD_EXAMPLES=OFF \
+          -DENABLE_SCCACHE=ON \
+          -DCMAKE_BUILD_TYPE=Release \
+          -DCMAKE_EXE_LINKER_FLAGS="-L$cuda_stub"
+        cmake --build build-nvlink-vmm-tests --target \
+          nvlink_host_numa_allocation_test nvlink_transport_test
+        mkdir -p "$RUNNER_TEMP/cuda-driver-stubs"
+        cp -L "$cuda_stub/libcuda.so" \
+          "$RUNNER_TEMP/cuda-driver-stubs/libcuda.so.1"
+        export LD_LIBRARY_PATH="$RUNNER_TEMP/cuda-driver-stubs:$cuda_stub:${LD_LIBRARY_PATH:-}"
+        ctest --test-dir \
+          build-nvlink-vmm-tests/mooncake-transfer-engine/tests \
+          -L nvlink_vmm_unit --output-on-failure
+      shell: bash
+
     - name: Build transfer engine only
       run: |
         cd mooncake-transfer-engine
@@ -536,6 +563,7 @@ jobs:
       src: ${{ github.event_name == 'workflow_dispatch' && 'true' || steps.filter.outputs.src }}
       tent: ${{ github.event_name == 'workflow_dispatch' && 'true' || steps.filter.outputs.tent }}
       docs: ${{ github.event_name == 'workflow_dispatch' && 'true' || steps.filter.outputs.docs }}
+      nvlink: ${{ github.event_name == 'workflow_dispatch' && 'true' || steps.filter.outputs.nvlink }}
     steps:
       - uses: actions/checkout@v4
         if: github.event_name != 'workflow_dispatch'
@@ -571,6 +599,15 @@ jobs:
               - '.github/workflows/ci.yml'
             docs:
               - 'docs/**'
+              - '.github/workflows/ci.yml'
+            nvlink:
+              - 'mooncake-transfer-engine/**'
+              - 'mooncake-common/**'
+              - 'cmake/**'
+              - 'extern/**'
+              - 'CMakeLists.txt'
+              - 'dependencies.sh'
+              - '.github/actions/**'
               - '.github/workflows/ci.yml'
 
   build-wheel-cu13:
--- a/mooncake-transfer-engine/src/transport/nvlink_transport/nvlink_host_numa_allocation.cpp
+++ b/mooncake-transfer-engine/src/transport/nvlink_transport/nvlink_host_numa_allocation.cpp
@@ -1,0 +1,428 @@
+// Copyright 2024 KVCache.AI
+
+#include "transport/nvlink_transport/nvlink_host_numa_allocation.h"
+
+#if MOONCAKE_NVLINK_HOST_NUMA_ENABLED
+
+#include <algorithm>
+#include <cstdlib>
+#include <limits>
+#include <mutex>
+#include <string>
+#include <unordered_map>
+#include <utility>
+
+#include <glog/logging.h>
+
+namespace mooncake {
+namespace {
+
+Status cudaFailure(const char* stage, CUresult result) {
+    return Status::Memory(std::string(stage) + " failed with CUDA result " +
+                          std::to_string(static_cast<int>(result)));
+}
+
+Status missingDriverFunction(const char* name) {
+    return Status::InvalidArgument(
+        std::string("missing CUDA driver adapter: ") + name);
+}
+
+bool isPowerOfTwo(size_t value) {
+    return value != 0 && (value & (value - 1)) == 0;
+}
+
+CUmemAllocationProp hostNumaProp(int numa_node) {
+    CUmemAllocationProp prop = {};
+    prop.type = CU_MEM_ALLOCATION_TYPE_PINNED;
+    prop.location.type = CU_MEM_LOCATION_TYPE_HOST_NUMA;
+    prop.location.id = numa_node;
+    prop.requestedHandleTypes = CU_MEM_HANDLE_TYPE_FABRIC;
+    return prop;
+}
+
+}  // namespace
+
+struct NvlinkHostNumaAllocation::OwnedRangeRegistry {
+    struct Entry {
+        size_t length;
+        DriverApi api;
+    };
+
+    std::mutex mutex;
+    std::unordered_map<void*, Entry> ranges;
+};
+
+NvlinkHostNumaAllocation::OwnedRangeRegistry&
+NvlinkHostNumaAllocation::OwnedRanges() {
+    static auto* registry = new OwnedRangeRegistry();
+    return *registry;
+}
+
+NvlinkHostNumaAllocation::DriverApi
+NvlinkHostNumaAllocation::ProductionDriverApi() {
+    return DriverApi{
+        cuDeviceGetCount,
+        cuDeviceGet,
+        cuDeviceGetAttribute,
+        cuMemGetAllocationGranularity,
+        cuMemCreate,
+        cuMemAddressReserve,
+        cuMemMap,
+        cuMemSetAccess,
+        cuMemUnmap,
+        cuMemAddressFree,
+        cuMemRelease,
+        cuMemRetainAllocationHandle,
+        cuMemExportToShareableHandle,
+    };
+}
+
+bool NvlinkHostNumaAllocation::RegisterOwnedRange(void* base, size_t length,
+                                                  const DriverApi& api) {
+    if (base == nullptr || length == 0) return false;
+    auto& registry = OwnedRanges();
+    std::lock_guard<std::mutex> lock(registry.mutex);
+    try {
+        return registry.ranges
+            .emplace(base, OwnedRangeRegistry::Entry{length, api})
+            .second;
+    } catch (...) {
+        return false;
+    }
+}
+
+bool NvlinkHostNumaAllocation::UnregisterOwnedRange(void* base, size_t length) {
+    auto& registry = OwnedRanges();
+    std::lock_guard<std::mutex> lock(registry.mutex);
+    auto entry = registry.ranges.find(base);
+    if (entry == registry.ranges.end() || entry->second.length != length)
+        return false;
+    registry.ranges.erase(entry);
+    return true;
+}
+
+bool NvlinkHostNumaAllocation::FindExactOwnedRange(void* base, size_t length,
+                                                   DriverApi* api) {
+    if (api == nullptr) return false;
+    auto& registry = OwnedRanges();
+    std::lock_guard<std::mutex> lock(registry.mutex);
+    auto entry = registry.ranges.find(base);
+    if (entry == registry.ranges.end() || entry->second.length != length)
+        return false;
+    try {
+        *api = entry->second.api;
+        return true;
+    } catch (...) {
+        return false;
+    }
+}
+
+bool NvlinkHostNumaAllocation::OverlapsOwnedRange(void* base, size_t length) {
+    if (base == nullptr || length == 0) return false;
+    const uintptr_t begin = reinterpret_cast<uintptr_t>(base);
+    const uintptr_t end = length > std::numeric_limits<uintptr_t>::max() - begin
+                              ? std::numeric_limits<uintptr_t>::max()
+                              : begin + length;
+
+    auto& registry = OwnedRanges();
+    std::lock_guard<std::mutex> lock(registry.mutex);
+    for (const auto& [owned_base, entry] : registry.ranges) {
+        const uintptr_t owned_begin = reinterpret_cast<uintptr_t>(owned_base);
+        const uintptr_t owned_end = owned_begin + entry.length;
+        if (begin < owned_end && owned_begin < end) return true;
+    }
+    return false;
+}
+
+Status NvlinkHostNumaAllocation::DiscoverHostNumaNodes(
+    std::vector<int>& numa_nodes) {
+    return DiscoverHostNumaNodesWithDriverApi(ProductionDriverApi(),
+                                              numa_nodes);
+}
+
+Status NvlinkHostNumaAllocation::ValidateVisibleGpuFabricSupportWithDriverApi(
+    const DriverApi& api, std::vector<CUdevice>& visible_devices) {
+    visible_devices.clear();
+    if (std::getenv("MC_USE_NVLINK_IPC") != nullptr)
+        return Status::NotSupportedTransport(
+            "MC_USE_NVLINK_IPC disables Fabric allocations");
+    if (!api.device_get_count) return missingDriverFunction("cuDeviceGetCount");
+    if (!api.device_get) return missingDriverFunction("cuDeviceGet");
+    if (!api.device_get_attribute)
+        return missingDriverFunction("cuDeviceGetAttribute");
+
+    int device_count = 0;
+    CUresult result = api.device_get_count(&device_count);
+    if (result != CUDA_SUCCESS) return cudaFailure("cuDeviceGetCount", result);
+    if (device_count <= 0)
+        return Status::DeviceNotFound("no visible CUDA device");
+
+    try {
+        visible_devices.reserve(static_cast<size_t>(device_count));
+    } catch (...) {
+        return Status::Memory("failed to record visible CUDA devices");
+    }
+    for (int ordinal = 0; ordinal < device_count; ++ordinal) {
+        CUdevice device;
+        result = api.device_get(&device, ordinal);
+        if (result != CUDA_SUCCESS) return cudaFailure("cuDeviceGet", result);
+
+        int vmm_supported = 0;
+        result = api.device_get_attribute(
+            &vmm_supported,
+            CU_DEVICE_ATTRIBUTE_VIRTUAL_MEMORY_MANAGEMENT_SUPPORTED, device);
+        if (result != CUDA_SUCCESS)
+            return cudaFailure("cuDeviceGetAttribute(VMM)", result);
+
+        int fabric_supported = 0;
+        result = api.device_get_attribute(
+            &fabric_supported, CU_DEVICE_ATTRIBUTE_HANDLE_TYPE_FABRIC_SUPPORTED,
+            device);
+        if (result != CUDA_SUCCESS)
+            return cudaFailure("cuDeviceGetAttribute(Fabric)", result);
+        if (!vmm_supported || !fabric_supported)
+            return Status::NotSupportedTransport(
+                "a visible CUDA device lacks VMM or Fabric handle support");
+
+#if CUDA_VERSION >= 12090
+        int host_numa_vmm_supported = 0;
+        result = api.device_get_attribute(
+            &host_numa_vmm_supported,
+            CU_DEVICE_ATTRIBUTE_HOST_NUMA_VIRTUAL_MEMORY_MANAGEMENT_SUPPORTED,
+            device);
+        if (result != CUDA_SUCCESS)
+            return cudaFailure("cuDeviceGetAttribute(HOST_NUMA VMM)", result);
+        if (!host_numa_vmm_supported)
+            return Status::NotSupportedTransport(
+                "a visible CUDA device lacks HOST_NUMA VMM support");
+#endif
+
+        visible_devices.push_back(device);
+    }
+    return Status::OK();
+}
+
+Status NvlinkHostNumaAllocation::DiscoverHostNumaNodesWithDriverApi(
+    const DriverApi& api, std::vector<int>& numa_nodes) {
+    numa_nodes.clear();
+    std::vector<CUdevice> visible_devices;
+    Status status =
+        ValidateVisibleGpuFabricSupportWithDriverApi(api, visible_devices);
+    if (!status.ok()) return status;
+
+    for (CUdevice device : visible_devices) {
+        int numa_node = -1;
+        CUresult result = api.device_get_attribute(
+            &numa_node, CU_DEVICE_ATTRIBUTE_HOST_NUMA_ID, device);
+        if (result != CUDA_SUCCESS)
+            return cudaFailure("cuDeviceGetAttribute(HOST_NUMA_ID)", result);
+        if (numa_node < 0)
+            return Status::Numa(
+                "a visible CUDA device has no GPU-local host NUMA node");
+        numa_nodes.push_back(numa_node);
+    }
+
+    std::sort(numa_nodes.begin(), numa_nodes.end());
+    numa_nodes.erase(std::unique(numa_nodes.begin(), numa_nodes.end()),
+                     numa_nodes.end());
+    for (int numa_node : numa_nodes) {
+        size_t ignored = 0;
+        status = GetAllocationGranularityWithDriverApi(numa_node, api, ignored);
+        if (!status.ok()) {
+            numa_nodes.clear();
+            return status;
+        }
+    }
+    return Status::OK();
+}
+
+Status NvlinkHostNumaAllocation::GetAllocationGranularity(int numa_node,
+                                                          size_t& granularity) {
+    return GetAllocationGranularityWithDriverApi(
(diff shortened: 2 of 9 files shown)


he-yufeng (commented):
Thanks for this — the VMM lifecycle handling here is careful work. A few things I want to call out and two points worth a closer look before merge. I focused the review on the new `NvlinkVmmAllocation` lifecycle in `nvlink_vmm_allocation.cpp`.

**What's done well.** The reverse-order `Release()` with per-stage commit and retained ownership on failure is the right shape for CUDA VMM teardown, and pre-allocating the intrusive `CleanupPendingVmmOwnerNode` *before* acquiring any CUDA resource (so recording a rollback can never itself fail under memory pressure) is a nice touch. Moving the owned-range provenance removal to the very start of `Release()`, before the first unmap, so a stuck teardown stops satisfying `IsExactOwnedRange` rather than staying eligible for a fresh remote registration, is exactly the safe direction.

**1. A single stuck teardown can silently wedge the whole allocation path (availability).** `CreateWithDriverApi` opens with:

```cpp
if (!RetryCleanupPendingOwners()) {
    return Status::Memory("a previous VMM factory rollback is still pending cleanup");
}
```

If any quarantined owner keeps failing `Release()` (driver persistently rejecting a teardown stage), every subsequent `Create` fails from here on, process-wide, with no upper bound and no escape valve. The retention itself is defensible, but the failure is invisible to operators: `CleanupPendingOwnerCount()` exists but isn't wired to any metric or log-once, so the symptom presents as "all Fabric VMM allocations suddenly fail" with the root cause buried. Could we at least surface the pending count (a gauge, or a throttled `LOG(ERROR)` naming the retained range) so this degraded state is observable? Whether a bounded-retry-then-give-up policy is acceptable is a product call, but the observability gap seems worth closing regardless.

**2. Rollback re-releases a handle whose first `cuMemRelease` returned an error (worth confirming).** In the success path the handle is released once and the flags cleared:

```cpp
result = api.mem_release(handle);
if (result != CUDA_SUCCESS) {
    status = cudaDriverFailure("cuMemRelease", result);
    return rollback(status);   // -> owner->Release() -> mem_release(handle) again
}
owner->handle_owned_ = false;
```

When that `mem_release` fails, `handle_owned_` is still `true`, so the `rollback` path's `Release()` calls `cuMemRelease` on the same handle a second time. This is only safe if a failing `cuMemRelease` is guaranteed to be atomic (no state change on error), so the retry acts on a still-valid handle rather than double-releasing one that was partially torn down. I believe that's the driver's contract, but given this is the one spot where the same handle can be released twice, it'd be good to either confirm that assumption explicitly or clear `handle_owned_` before entering `rollback` here and lean on the quarantine path if the handle truly leaked. Same reasoning applies to the mid-`Create` failures that reach `rollback` while `mapped_`/`address_reserved_` are set — those are fine because each stage's flag gates its own teardown, but the handle stage is the asymmetric one.

Neither of these blocks the design, which I think is sound. #2 is a small correctness clarification; #1 is the one I'd most like addressed since a wedged allocation path is hard to diagnose in production. Happy to look again once you've had a chance to weigh in.

mooncake-transfer-engine/src/transfer_metadata.cpp
```diff
@@ -1257,27 +1258,49 @@ int TransferMetadata::removeLocalSegment(const std::string &segment_name) {
...
+    std::shared_ptr<SegmentDesc> old_segment_desc;
+    std::shared_ptr<SegmentDesc> new_segment_desc;
     {
         RWSpinlock::WriteGuard guard(segment_lock_);
-        auto new_segment_desc = std::make_shared<SegmentDesc>();
-        auto &segment_desc = segment_id_to_desc_map_[LOCAL_SEGMENT_ID];
+        auto& segment_desc = segment_id_to_desc_map_[LOCAL_SEGMENT_ID];
+        old_segment_desc = segment_desc;
+        new_segment_desc = std::make_shared<SegmentDesc>();
         *new_segment_desc = *segment_desc;
```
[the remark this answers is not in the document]
neverhook: This hunk has been removed from the PR. The current branch no longer modifies transfer_metadata.cpp, and the file is identical to origin/main. A defensive local-segment invariant change would be a separate Metadata correctness change rather than part of the NVLink VMM/Fabric lifecycle scope.

mooncake-transfer-engine/src/transfer_metadata.cpp
```diff
@@ -1257,27 +1258,49 @@ int TransferMetadata::removeLocalSegment(const std::string &segment_name) {
...
```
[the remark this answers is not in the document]
neverhook: This hunk has also been removed from the PR. The current branch leaves transfer_metadata.cpp identical to origin/main; any defensive hardening of the local-segment invariant should be reviewed as a focused Metadata correctness change, outside this lifecycle PR.

mooncake-transfer-engine/src/transport/nvlink_transport/nvlink_memory_lifecycle.cpp
```diff
@@ -0,0 +1,1045 @@
...
+                        result = fabric_driver_api_.mem_set_access(
+                            attempt.address(), entry.length, &access, 1);
+                        if (result != CUDA_SUCCESS) {
+                            LOG(ERROR)
+                                << "NvlinkTransport: cuMemSetAccess failed for "
+                                   "visible device "
+                                << ordinal << ": " << result;
+                            return ERR_MEMORY;
+                        }
+                    }
```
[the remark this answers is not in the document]
neverhook: Fixed in 5e06423c. The consumer lazy-import path now builds all visible-device CUmemAccessDesc entries before importing or reserving any CUDA resource, then applies them with one cuMemSetAccess call. The fake-driver tests verify one batched call containing both visible devices, including failure cleanup and retry paths.

mooncake-transfer-engine/src/transport/nvlink_transport/nvlink_vmm_allocation.cpp
```diff
@@ -0,0 +1,816 @@
...
+        result = api.device_get(&device, ordinal);
+        if (result != CUDA_SUCCESS) {
+            status = cudaDriverFailure("cuDeviceGet", result);
+            return rollback(status);
+        }
+        status = grant_access(CU_MEM_LOCATION_TYPE_DEVICE, device);
+        if (!status.ok()) {
+            return rollback(status);
+        }
+    }
```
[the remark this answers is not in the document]
neverhook: Fixed in 5e06423c. The provider VMM factory now prepares the HOST_NUMA plus all visible-device descriptors before acquiring CUDA ownership and submits the complete batch in one cuMemSetAccess call. Unit tests assert a single call with two DEVICE descriptors or one HOST_NUMA plus two DEVICE descriptors, and cover batched-access rollback.

neverhook:
@he-yufeng Thanks for the detailed lifecycle review. Addressed both points in 5e06423c.

1. Cleanup-pending observability
- The CreateWithDriverApi failure now includes CleanupPendingOwnerCount in the returned status.
- Failed cleanup retries emit a throttled ERROR log with the retained base, length, location type/id, and CUDA error.
- Retryable staged ownership for cuMemUnmap and cuMemAddressFree remains unchanged.

2. cuMemRelease at-most-once ownership
- I did not rely on failure being atomic. Every relevant lifecycle path now consumes its handle-owned marker before invoking cuMemRelease.
- A non-success result is treated as ambiguous and is logged, but the same opaque handle is never retried. This covers factory post-map release, owner Release, consumer import cleanup, cached mapping cleanup, registration failure, and unregistration retry.
- Fake-driver tests model an error that has already consumed the reference and assert exactly one release call.

Also addressed the two batching comments: provider allocation and consumer lazy import now build complete access descriptor arrays before acquiring CUDA ownership, then issue one cuMemSetAccess call. Unrelated whole-header formatting churn was reverted.

Validation for head 5e06423c: clang-format 20 passed, and both ARM64 Python 3.10/3.12 jobs passed Build project plus Build and gate NVLink VMM unit tests in https://github.com/neverhook/Mooncake/actions/runs/29802040245. The broader manually triggered matrix is still finishing, but the two NVLink gates are complete.

neverhook:
Current version summary (`3c850601`):

- Separates visible-GPU Fabric/VMM capability checks from NUMA discovery: explicit placement validates only the selected node's granularity, while all visible GPUs still receive one batched `cuMemSetAccess` grant.
- Closes exact HOST_NUMA registration cleanup: metadata deletion is not repeated, publication failures roll back immediately, `ERR_ADDRESS_NOT_REGISTERED` is idempotent, and failed retained-handle release remains retryable.
- Makes `OwnedRangeRegistry` process-lifetime and adds scripted regression coverage for placement, rollback, publication, metadata, and handle-release failure paths.
- Validation is complete: focused CUDA-on tests pass 2/2, CUDA-off builds 113/113 targets, and exact-head GitHub Actions pass 28/28 checks including CI Gate.

Scope remains the minimum Transfer Engine prerequisite for #2914. Store EGM pool integration, the complete Provider/Consumer path, and GB200/NVL72 end-to-end validation stay in the follow-up Store-only PR and are not merge gates for #2966.

neverhook:
Stack refresh: this PR is now [1/3] and has been rebased onto current `main`. The compile-time capability guard requested in review is consolidated as `MOONCAKE_NVLINK_HOST_NUMA_ENABLED`; runtime Fabric/VMM checks remain unchanged. CUDA 12.8 ARM64 `nvlink_vmm_unit` passes 2/2, the CUDA-off Transfer Engine target builds, and exact-head GitHub Actions are running. #3335 and #3431 are strict child branches of this head.

mooncake-transfer-engine/src/transport/nvlink_transport/nvlink_transport.cpp
```diff
@@ -737,8 +781,130 @@ int NvlinkTransport::registerLocalMemory(void *addr, size_t length,
...
+    const RemoveLocalMemoryBufferOp &remove_buffer,
+    const UpdateLocalSegmentDescOp &update_segment) {
+    CUmemGenericAllocationHandle handle;
+    CUresult result = driver_api.mem_retain_allocation_handle(&handle, addr);
+    if (result != CUDA_SUCCESS) {
+        LOG(WARNING) << "NvlinkTransport: failed to retain the exact HOST_NUMA "
+                        "allocation handle for "
+                     << addr << ": " << result;
+        return 0;
+    }
```
alogfans, line 1114: Why does it return 0?

neverhook:
Review follow-up is now pushed. The lifecycle change rejects retain failures
before publication, uses a placeholder-before-retain record, and preserves
allocation/registration ownership when `cuMemRelease()` fails so a later
cleanup can retry. Focused CUDA-on/off builds pass, including
`nvlink_vmm_unit` 2/2. The release-retry policy and CUDA failure-atomicity
ambiguity are called out explicitly in the updated description.

mooncake-transfer-engine/src/transport/nvlink_transport/nvlink_transport.cpp
neverhook: Fixed. cuMemRetainAllocationHandle failure now removes the placeholder and returns ERR_MEMORY before descriptor export or metadata publication. The focused regression verifies ERR_MEMORY, zero publication calls, and zero retained registrations.

neverhook:
The full three-layer stack has been rebased onto the current `main`. This
prerequisite patch is unchanged by `range-diff`; clang-format, diff checks,
touched-file hooks, the CUDA-on focused build, and `nvlink_vmm_unit` (2/2)
pass. A new exact-head Actions run has been triggered by the force-push.

neverhook:
The full stack has been rebased again onto the current `main`, now including
the upstream standby snapshot compile fix. This prerequisite patch remains
identical by `range-diff`. Static checks and the unmodified CUDA-on focused
build pass, including `nvlink_vmm_unit` 2/2. A new exact-head Actions run has
been triggered.

neverhook:
Rebased the full RFC #2914 stack onto current `main`. Range-diff confirms this [1/3] Transfer Engine layer is semantically unchanged. Local gates passed: clang-format/diff checks, CUDA-on `nvlink_vmm_unit` 2/2, and CUDA-off Transfer Engine build. Exact-head GitHub Actions are now the remaining automated gate.

neverhook:
Stack rebase update:

- Rebased onto the current `main`; the #2966 patch is range-diff equivalent.
- Preserved one logical commit and the existing 8-file scope.
- Local validation: clang-format 20, `git diff --check`, applicable touched-file hooks, CUDA-on `nvlink_vmm_unit` 2/2, and CUDA-off Transfer Engine build.
- #3335 and #3431 were replayed on top in dependency order.

Exact-head GitHub Actions are now the remaining automated gate. GB200/NVL72 integration remains a later Store-stack gate, not a gate for #2966.

neverhook:
Stack maintenance update:

- Rebased the complete `[1/3] -> [2/3] -> [3/3]` stack onto current `main`; this layer remains patch-equivalent by `range-diff`.
- The composed head passed diff/format/pre-commit checks, CUDA-on focused tests (`nvlink_vmm_unit` 2/2, `egm_store_pool_unit` 1/1, direct-read 3/3), and CUDA-off Store/Transfer Engine/Master/Python validation.
- Fresh exact-head GitHub Actions have been triggered and are currently queued.

GB200 was not rerun because this was a patch-equivalent baseline rebase with no functional-path change.

mooncake-transfer-engine/src/transport/nvlink_transport/nvlink_transport.cpp
```diff
@@ -1036,8 +1081,149 @@ int NvlinkTransport::registerLocalMemory(void *addr, size_t length,
...
+                      "allocation handle for "
+                   << addr << ": " << result;
+        const int cleanup_rc = unregisterHostNumaMemoryLocked(
+            addr, false, remove_buffer, update_segment);
+        if (cleanup_rc != 0) {
+            LOG(ERROR) << "NvlinkTransport: failed to release retained "
+                          "HOST_NUMA handle after export failure for "
+                       << addr << ": " << cleanup_rc;
+        }
+        return ERR_MEMORY;
```
staryxchen, line 1132: Setting metadata_removed_locally = true here assumes removeLocalMemoryBuffer can only fail after the local erase (i.e., the error comes from updateLocalSegmentDesc) — that holds today, but it's an undocumented implementation detail of the callee rather than part of its contract. If it ever gains a failure path before the local erase, the retry would skip removal and leave a stale entry in the local segment descriptor. Could we pin this invariant on the callee side, via a comment or a dedicated return code?

.github/workflows/ci.yml
```diff
@@ -258,6 +258,33 @@ jobs:
         df -h
       shell: bash
 
+    - name: Build and test NVLink HOST_NUMA units
```
staryxchen, line 506: This step inherits the broad src filter from build-flags, so any mooncake-*/** change (store, wheel, scripts...) triggers a near-cold rebuild of the transfer engine just to run these two test binaries — the USE_MNNVL=ON USE_CUDA=ON flags don't match any other job, so sccache rarely helps. Since tent/reshard already have narrow filters in check-paths, could this get the same treatment — e.g., a nvlink output covering mooncake-transfer-engine/**, mooncake-common/**, top-level CMake, dependencies.sh, and ci.yml, gating this step via needs.check-paths.outputs.should-run-nvlink?

mooncake-transfer-engine/src/transport/nvlink_transport/nvlink_transport.cpp
neverhook: Addressed with a callee-side contract comment on `TransferMetadata::removeLocalMemoryBuffer()`: the local descriptor entry is removed before optional publication, a publication error does not roll that removal back, and `ERR_ADDRESS_NOT_REGISTERED` means no matching local entry existed. The existing `FailedPublicationRollbackCanBeCompletedByUnregister` regression continues to assert that the failed path removes once and a later cleanup retries only descriptor publication. No return code or public signature was changed.

.github/workflows/ci.yml
neverhook: Addressed with a dedicated `should-run-nvlink` output from `check-paths`. The focused HOST_NUMA step now gates on that output and `workflow_dispatch` forces it on. The core filter covers Transfer Engine/common/CMake/dependency/CI inputs; the stacked Store layer adds `mooncake-store/**` when it extends the same step with `egm_store_pool_test`. This also adapts the step to current `main` by removing the obsolete Python-matrix condition. YAML/static checks and CUDA-on focused tests pass.

neverhook:
Review and stack update:

- Documented the callee-side `removeLocalMemoryBuffer()` contract without changing its signature or behavior.
- Added a dedicated NVLink path output and adapted the focused step to current `main`'s single Python 3.12 `build-flags` job.
- Rebased the full `[1/3] -> [2/3] -> [3/3]` stack onto current `main`. Local CUDA-on/off focused builds and lifecycle tests pass.
- Replied to and resolved both new review threads. Fresh exact-head Actions are queued.

neverhook:
Rebased the stack on current main. The HOST_NUMA and registration patch is unchanged. Focused tests pass; the fresh CI run is still finishing.

neverhook:
Rebased on current main, including the SSD-test fix in #3870. The HOST_NUMA patch is unchanged. Fresh CI has been triggered for this head.

neverhook:
The stack is rebased on current main. I kept the new CI layout and wired the focused NVLink tests into its path filter; the allocation and registration code did not need changes.

staryxchen:
PR#3978 is merged, does this overlap with your changes?

neverhook:
> PR#3978 is merged, does this overlap with your changes?

both add HOST_NUMA/EGM support, but #3978 targets TENT’s MnnvlTransport; #2966 targets the non-TENT NvlinkTransport used by #3335. #3978 does not implement the Store pool or Store direct-read changes.

mooncake-transfer-engine/src/transport/nvlink_transport/nvlink_transport.cpp
```diff
@@ -989,6 +1002,38 @@ int NvlinkTransport::registerLocalMemory(void *addr, size_t length,
...
+            NvlinkHostNumaAllocation::FindExactOwnedRange(addr, length,
+                                                          &exact_range_api);
+        if (!exact_host_numa_range &&
+            NvlinkHostNumaAllocation::OverlapsOwnedRange(addr, length)) {
+            LOG(ERROR) << "NvlinkTransport: HOST_NUMA registration must use "
+                          "the exact owned range: "
+                       << addr << ", length " << length;
+            return ERR_INVALID_ARGUMENT;
+        }
+        if (exact_host_numa_range &&
```
staryxchen: `ERR_ADDRESS_OVERLAPPED` treats a failed registration the same as a published one. If export or publication fails and the rollback `cuMemRelease` also fails, the map entry stays (`handle_owned && !metadata_removed_locally`). A later `register()` hits this check and returns overlap; callers retry register, not unregister. After `Release()` recycles the VA, the leftover key blocks the new allocation, and a later `unregister()` releases the old handle against the new VA. Distinguish "published" from "rollback pending", and let a same-address register finish the pending cleanup.

mooncake-transfer-engine/src/transport/nvlink_transport/nvlink_host_numa_allocation.cpp
```diff
@@ -0,0 +1,428 @@
...
+
+NvlinkHostNumaAllocation::~NvlinkHostNumaAllocation() {
+    Status status = Release();
+    if (!status.ok())
+        LOG(ERROR) << "NvlinkHostNumaAllocation cleanup failed: " << status;
+}
+
+Status NvlinkHostNumaAllocation::Release() {
+    const CUdeviceptr ptr = reinterpret_cast<CUdeviceptr>(base_);
+    if (owned_range_registered_) {
```
staryxchen, line 406: Provenance is removed before unmap, address free, and release. If `cuMemUnmap` fails, both `FindExactOwnedRange` and `OverlapsOwnedRange` return false, so `registerLocalMemory` falls through to the legacy retain/export path with no exact-range check. Keep the range quarantined until teardown finishes.

mooncake-transfer-engine/src/transport/nvlink_transport/nvlink_host_numa_allocation.cpp
```diff
@@ -0,0 +1,428 @@
...
+    auto owner = std::unique_ptr<NvlinkHostNumaAllocation>(
+        new NvlinkHostNumaAllocation());
+    owner->length_ = length;
+    owner->numa_node_ = numa_node;
+    owner->driver_api_ = api;
+
+    auto rollback = [&](Status cause) {
+        Status cleanup = owner->Release();
+        if (!cleanup.ok()) {
+            allocation = std::move(owner);
```
staryxchen, line 357: Incomplete rollback returns an error and a non-null `unique_ptr`. Callers that bail on `!status.ok()` leak the mapped range; the destructor retries `Release()` only once. Keep the owner in a process-lifetime quarantine, or require the caller to `Release()` until success and test that contract.

mooncake-transfer-engine/src/transport/nvlink_transport/nvlink_host_numa_allocation.cpp
```diff
@@ -0,0 +1,428 @@
...
+
+Status NvlinkHostNumaAllocation::Release() {
+    const CUdeviceptr ptr = reinterpret_cast<CUdeviceptr>(base_);
+    if (owned_range_registered_) {
+        if (!UnregisterOwnedRange(base_, length_))
+            return Status::Memory(
+                "HOST_NUMA owned-range provenance removal failed");
+        owned_range_registered_ = false;
+    }
+    if (mapped_) {
```
staryxchen, line 415: `Release()` unmaps the VA even while `NvlinkTransport` still retains the handle and metadata still publishes the range. The retain keeps the physical allocation, not this VA. Refuse `Release()` while the exact range is registered, or unregister first and fail if that cannot finish.

staryxchen (commented):
Overall the approach looks right. Left a few inline comments, please take a look.

Aionw:
The new `nvlink` path filter in `.github/workflows/ci.yml` is wide enough that the focused NVLink step runs on most PRs, which I don't think is the intent.

Replaying the filter over the last 150 first-parent merges on `main` (head `e699484`):

| rule | matches |
| --- | --- |
| `nvlink` filter as written | 78 / 150 (52%) |
| merges actually touching `nvlink_transport/` | 1 / 150 |

The list (`mooncake-transfer-engine/**`, `mooncake-common/**`, `cmake/**`, `extern/**`, `CMakeLists.txt`, `dependencies.sh`, `.github/actions/**`, `ci.yml`) is almost character-for-character the existing `tent` filter, and the step lives in `build-flags`, which is already gated on `src == 'true'`. So the only thing the new output filters out is PRs that touch `src` but not the Transfer Engine; anything touching TE pays for it.

What it pays is a full-repo configure + build (`-S . -B build-nvlink-vmm-tests`, `USE_CUDA=ON -DUSE_MNNVL=ON`) added to a job that already runs its own Transfer Engine build with the opposite flags (`USE_CUDA=OFF -DUSE_MNNVL=OFF`), so sccache reuse between the two is limited.

**Suggestion: drop the dedicated build rather than tune the filter.** The `unit-tests` job already configures `USE_CUDA=ON` + `BUILD_UNIT_TESTS=ON`, and its `Configure CUDA driver runtime` step already resolves the driver library and creates `libcuda.so.1`. Adding `-DUSE_MNNVL=ON` there builds `nvlink_transport/` (two `.cpp` files, no new external deps — `nvlink-allocator` is already built under `USE_CUDA`), and both fake-driver tests then run in the existing ctest sweep. The `nvlink` filter output and the 37-line step in `build-flags` can both go away, and the tests end up where the rest of the unit tests already run.

One prerequisite: `add_test(NAME nvlink_transport_test ...)` registers the whole binary, and the pre-existing `NvlinkTransportTest.WriteAndRead` needs a real GPU (`cudaSetDevice`/`cudaMalloc`, `exit(EXIT_FAILURE)` on error). It would need a label (e.g. `nvlink_gpu`) plus `--label-exclude` on the main ctest run; the `nvlink_vmm_unit` cases run unchanged.

If you would rather keep the dedicated step, narrowing the filter alone also fixes the trigger rate: `mooncake-transfer-engine/src/transport/nvlink_transport/**`, `mooncake-transfer-engine/include/transport/nvlink_transport/**`, `mooncake-transfer-engine/tests/nvlink_*`, `mooncake-transfer-engine/tests/CMakeLists.txt`, `mooncake-common/common.cmake` and `.github/workflows/ci.yml` replays as 10/150 (~7%) over the same window.

mooncake-transfer-engine/src/transport/nvlink_transport/nvlink_host_numa_allocation.cpp
neverhook: Release now keeps a releasing entry until address-free succeeds. Both exact and subrange registration are rejected if unmap or address-free failed; neither can fall through to the legacy path. Address-free and removal of the entry are atomic under the registry lock. The tests also cover reusing the VA while an old owner still needs to retry its handle release.

mooncake-transfer-engine/src/transport/nvlink_transport/nvlink_host_numa_allocation.cpp
neverhook: Added per-range registration pins. Release refuses to unmap while registration or its cleanup is pending, and unregister drops the pin only after metadata cleanup and retained-handle release succeed. Deferred and batch cleanup follow the same rule. Added tests for concurrent register/release, multiple transports, partial batch cleanup, and destructor failure.

mooncake-transfer-engine/src/transport/nvlink_transport/nvlink_host_numa_allocation.cpp
neverhook: I took the explicit-owner contract option. Create now rejects a nonempty output instead of resetting it, and the header documents that an error may still return ownership that the caller must retain and release. Destruction remains best-effort. The allocator and Store tests cover repeated cleanup failures, retained ownership, and retries that skip completed stages.

mooncake-transfer-engine/src/transport/nvlink_transport/nvlink_transport.cpp
neverhook: I separated successful registrations from rollback-pending ones. A same-address retry now finishes the old cleanup before retaining a new handle, while a successful registration still rejects duplicates. A new pin stays held across that handoff so the VA cannot be released between attempts. Added regressions for export, metadata publication, and release failures followed by register retries.

neverhook:
Thanks @Aionw, I went with the shared build. The fake-driver cases now run in `unit-tests`; the separate NVLink build and filter are gone. Only the real-GPU entry is excluded on CPU runners. The full CTest sweep and CI Gate passed across the stack.

staryxchen (approved):
Looks good from TE side

Aionw (approved):
LGTM for ci

Issue #2930, opened 2026-07-15
[Bug] TcpTransport connection pool: unbounded growth with threshold mass-GC; stale entries to restarted peers are never evicted


d3nb:
## Environment

mooncake-transfer-engine-cuda13 0.3.11.post1 (PyPI), used as SGLang PD-disaggregation KV transfer backend (`MC_FORCE_TCP=1`, `MC_TCP_ENABLE_CONNECTION_POOL=true`), same-host prefill(TP4) → decode(TP4), 8×B300, container network.

## Issue 1: pool grows unboundedly, then mass-closes tens of thousands of connections at once

Under sustained load (fixed 10-way concurrency, ~27K-token KV parcels, chunked), `/proc/net/tcp` on the sender side shows a sawtooth with ~8-minute period:

```text
03:15  ESTABLISHED 29024
03:17  ESTABLISHED 38952
03:18  ESTABLISHED 17724   TIME_WAIT 21089   <- ~21K connections closed within one minute
03:24  ESTABLISHED 44419
03:25  ESTABLISHED 45623   <- 83% of the ephemeral port range
03:26  ESTABLISHED 20738   TIME_WAIT 26831   <- ~27K closed at once
```

Two problems: (a) between flushes the pool keeps growing toward the ephemeral-port limit (we measured peaks at 83% of a widened 10240-65000 range — with the default ~28K-port range this would hit `EADDRNOTAVAIL` long before the flush); (b) the flush itself dumps 20K+ sockets into TIME_WAIT within seconds. A bounded pool with per-peer connection reuse would presumably hold this at a few hundred connections (we see ~566 ESTABLISHED steady when idle-ish).

## Issue 2: pooled connections to a dead/restarted peer are never evicted

If one side of the PD pair is restarted alone (its listen ports change), the surviving side's pool keeps the stale entries. Small transfers still succeed (presumably via new connections), but **large multi-slice transfers stall silently and permanently** — the sender reports no error, the receiver times out (`KVPoll.WaitingForInput` timeout after 300s). It also degrades progressively: after one single-sided peer restart we only measured throughput loss at high concurrency; after a second one, all large transfers hung. Restarting both sides together fully restores normal behavior.

## Ask

1. Bound the pool (per-peer cap + LRU reuse) instead of grow-then-mass-GC.
2. Evict/reconnect pool entries on connection failure or peer-identity change, so a peer restart doesn't silently poison large transfers.

[...]

PR #2974, opened 2026-07-17
[Transfer Engine] Schedule TCP transfers on bounded per-peer connection lanes

jacklin78911-collab:
## Summary

Revised implementation for #2930, following review feedback on the earlier
revision of this PR.

The original TCP transport opens a new connection whenever no idle connection
is available for a destination, so socket count follows peak concurrency with
no upper bound. The first revision of this PR bounded sockets with a per-peer
cap plus a FIFO queue of acquisition callbacks; review correctly noted that
this moved the unbounded state from sockets into waiter state rather than
establishing end-to-end boundedness.

This revision changes the ownership model accordingly: **sockets are owned by
fixed per-destination lanes, not by individual Slices.**

```text
bounded pending-admission FIFO
→ bounded Slice work queue
→ fixed per-peer connection lanes
```

A Slice never acquires or owns a socket. It remains in bounded group-owned state
until an idle connected lane pulls it for execution, so backpressure is applied
when transfer work enters the transport rather than during socket acquisition.

## Scope

Implemented:

- a per-destination `PeerConnectionGroup` owning a fixed set of long-lived
  `ConnectionLane`s;
- a bounded work queue and a bounded pending-admission FIFO;
- admission deadlines for Slices waiting to enter the work queue;
- pull-at-idle scheduling, so a slow transfer blocks only its own lane;
- group-level reconnect rounds with a serialized cooldown timer;
- move-only work ownership and exactly-once terminal completion;
- internal failure accounting by reason;
- deterministic shutdown handling for queued, pending-admission, connecting,
  and busy state;
- interoperability coverage for the existing legacy framing path.

Intentionally **not** implemented:

- peer-generation or stale/half-closed socket detection after peer restart;
- payload or transfer-progress deadlines;
- a global bound on the number of peer groups;
- public exposure of transport-internal failure reasons;
- changes to transfer-status counter synchronization.

### Object model

Each destination `ConnectionKey` maps to one `PeerConnectionGroup`.

The group owns, under a single mutex:

- its fixed lane vector;
- the bounded admitted-work queue;
- the bounded pending-admission FIFO;
- admission and reconnect timer state;
- reconnect-round accounting;
- internal failure counters.

Each `ConnectionLane` owns its resolver, socket, session, lifecycle state, and
operation epoch, plus at most one active work item.

The epoch is validated by asynchronous handlers so that late resolve, connect,
timer, session, and shutdown handlers cannot reclaim ownership that has already
moved elsewhere.

### Admission

A new submission is admitted directly to the work queue only when:

- work-queue capacity is available; and
- no older pending admission is waiting.

Otherwise it enters the bounded pending-admission FIFO with a deadline.
Pending admissions are promoted into the work queue in FIFO order whenever
capacity becomes available; newer submissions cannot bypass older pending
ones.

One per-group timer tracks the earliest pending deadline. If the deadline
expires first, the Slice is completed as `FAILED` with the internal
`QUEUE_TIMEOUT` reason.

Only when both the work queue and pending-admission FIFO are at their hard
bounds is a newly submitted Slice rejected with `QUEUE_FULL`. Rejection and
terminal Slice completion occur after releasing the group mutex.

Failure reasons are counted internally through `recordWorkFailure`, including:

- queue full;
- admission timeout;
- runtime unavailable;
- connect failure;
- session failure;
- shutdown.

External callers currently continue to observe a generic `FAILED` transfer
status; the transport-internal enum is not exposed through the public API by
this revision.

### Lane behavior

A disconnected lane attempts to resolve and connect only while admitted work
exists. Connecting lanes do not remove work from the queue.

Once connected, a lane becomes idle. An idle lane pulls the queue head,
transitions to busy, and starts a session for that work item.

After a terminal session result:

- a clean connection returns to the idle state and may serve the next Slice;
- an unusable connection is discarded and the lane returns to disconnected.

Mid-transfer work is never transparently replayed on another lane. A failed
session completes its Slice as failed and leaves retry policy to the caller or
upper layer.

### Ownership and exactly-once completion

```text
submitter-local work item
    → pending-admission FIFO or work queue
    → lane.current
    → local terminal action
    → exactly one Slice terminal call
```

`TcpWorkItem` and `TerminalAction` are move-only. Every ownership transition
moves the work item from exactly one location to the next.

`ClientSession` does not mark the Slice directly. It reports one terminal
`(status, connection_clean)` result to its owning lane. The lane validates its
epoch and claims the current work item under the group mutex; terminal Slice
completion is then performed outside the mutex.

Timeout, promotion, terminal handling, and shutdown compete for ownership
through the same group state, preventing a Slice from being dropped or
completed twice.

### Reconnect rounds and cooldown

Reconnect is demand-driven and rate-limited per destination.

A reconnect round permits:

- at most one concurrent connection probe;
- at most one connection attempt per lane during that round.

If all lane attempts in a round fail without establishing a usable connection,
the admitted work waiting on that round is completed with `CONNECT_FAILED` and
the group enters a one-second cooldown.

A single per-group retry timer owns the transition out of cooldown. Delayed
pumps cannot start a new round while that timer is armed. The timer handler
validates its group state and epoch before beginning another round, and cannot
restart a closing group.

Submissions during cooldown remain subject to the same bounded admission
structures and cannot bypass the cooldown transition.

### Shutdown

Shutdown first stops new admission and marks all groups closing.

Under each group mutex it detaches:

- admitted queued work;
- pending admissions;
- admission and reconnect timer ownership;
- lane-owned state required for cancellation.

Slice completion and cancellation dispatch occur outside the group mutex.

Lane operation epochs and timer epochs are invalidated so late handlers cannot
claim work that shutdown has already detached. Executor-posted cancellation
actions are allowed to run before the existing stop-and-join boundary; this is
a cancellation-dispatch barrier, not a claim that every asynchronous handler
has already quiesced.

After the worker is stopped and joined, remaining Asio ownership is released
while the context is still valid, and deferred busy work is completed exactly
once.

The existing stop/join ordering is preserved. This revision strengthens
admission-, queue-, reconnect-, and shutdown-path ownership without adding a
drain guarantee for transfers already in flight.

## Configuration

Connection-lane scheduling remains controlled by the existing:

```text
MC_TCP_ENABLE_CONNECTION_POOL
```

When enabled, the following settings are available:

| Environment variable | Default | Valid range |
|---|---:|---:|
| `MC_TCP_LANES_PER_PEER` | `4` | `1`–`16` |
| `MC_TCP_MAX_QUEUED_TRANSFERS_PER_PEER` | `1024` | `1`–`65535` |
| `MC_TCP_MAX_PENDING_ADMISSIONS_PER_PEER` | `1024` | `1`–`65535` |
| `MC_TCP_ADMISSION_TIMEOUT_MS` | `1000` | `1`–`600000` |

Invalid values fall back to their defaults.

`MC_TCP_MAX_CONNECTIONS_PER_PEER` from the earlier revision is retained as a
deprecated alias for `MC_TCP_LANES_PER_PEER` when the new variable is not set,
and emits a warning.

These defaults are initial guardrails rather than workload-specific tuning
claims; feedback on them is welcome.

## Compatibility

When connection pooling is disabled, the existing one-shot connection path is
unchanged.

For legacy protocol compatibility, the test suite covers a pooled legacy
framing initiator communicating with the current server for both WRITE and
READ, including transferred-data verification:

```text
legacy-framing initiator → current server: successful WRITE and READ
```

The stale-descriptor tests separately cover the opposite mismatch:

```text
current v2 initiator → legacy server: bounded, safe failure
```

They verify that a stale v2 descriptor does not cause an indefinite hang or an
unsafe successful completion. This revision does not claim successful
cross-version v2-to-v1 operation, nor does it build and launch a binary from an
older Mooncake commit.

## Tests and validation

New behavioral coverage includes:

- per-peer lane, work-queue, and pending-admission bounds;
- immediate `QUEUE_FULL` rejection only after the pending-admission hard bound
  is also reached;
- FIFO promotion from pending admission into the work queue;
- admission timeout completing exactly once;
- both timeout-versus-capacity-release race orderings;
- clean socket reuse by a single lane;
- work remaining queued while a lane connects;
- concurrent admission and shutdown ownership;
- shutdown with queued, pending-admission, connecting, and busy state;
- late connect, retry-timer, and admission-timer handlers;
- group-level reconnect cooldown against an unavailable peer;
- pooled legacy-framing WRITE and READ interoperability;
- safe bounded failure against a legacy server reached through a stale v2
  descriptor.

Local validation:

- `tcp_write_visibility_test`: `24/24` passed;
- ASan/LSan/UBSan: `24/24` passed with no sanitizer or leak diagnostics;
- targeted TSan admission/retry/shutdown/ownership subset: `10/10` passed;
- production `BUILD_UNIT_TESTS=OFF` build: passed, with no TCP test-hook symbols.

TSan was run on targeted ownership and shutdown tests rather than claimed for
the full suite. A pre-existing race exists between
`Transport::Slice::markSuccess()` and `TcpTransport::getTransferStatus()`;
neither access site is modified by this revision, so a globally clean TSan run
is not claimed.

One `tent-ci (cuda-off)` run failed because the GitHub runner exhausted its
disk space (`No space left on device`). A maintainer confirmed that this
infrastructure failure can be ignored for this PR.

## Known limitations

The pending-admission deadline bounds only the time spent waiting to enter the
work queue.

Once a Slice has been promoted into the admitted-work queue, it no longer has
an admission deadline. A connected transfer that makes no progress and
surfaces no I/O error can therefore hold its lane indefinitely and may leave
already-admitted work waiting behind the fixed lanes.

Payload-progress deadlines, stale-connection detection, and peer-generation
recovery are deliberately left to a follow-up phase of #2930.

The queue bounds also limit item counts rather than total queued source bytes.
`queued_bytes` is maintained as internal accounting, including saturation-safe
handling, but it is not currently an admission limit.

## Suggested review order

1. `tcp_transport.h`
   - `TcpWorkItem` and `TerminalAction`;
   - `PeerConnectionGroup`;
   - `ConnectionLane`;
   - timer, queue, and failure-counter state.
2. `tcp_transport.cpp`
   - configuration parsing and group creation;
   - admission, pending FIFO, promotion, and timeout;
   - pull-at-idle lane scheduling;
   - terminal ownership and completion;
   - reconnect rounds and cooldown;
   - shutdown extraction and deferred completion.
3. `tcp_write_visibility_test.cpp`
   - bounds and FIFO behavior;
   - admission timeout races;
   - reconnect cooldown;
   - shutdown and late-handler ownership;
   - legacy protocol interoperability.

### DIFF
--- a/mooncake-transfer-engine/include/transport/tcp_transport/tcp_transport.h
+++ b/mooncake-transfer-engine/include/transport/tcp_transport/tcp_transport.h
@@ -19,19 +19,25 @@
 
 #include <atomic>
 #include <chrono>
+#include <csignal>
 #include <cstddef>
+#include <cstdint>
 #include <deque>
 #include <functional>
 #include <map>
 #include <memory>
 #include <mutex>
+#include <optional>
 #include <string>
+#include <thread>
+#include <type_traits>
 #include <unordered_map>
-#include <unordered_set>
+#include <utility>
 #include <vector>
-#include <csignal>
 
+#include <asio/io_context.hpp>
 #include <asio/ip/tcp.hpp>
+#include <asio/steady_timer.hpp>
 
 #include "transfer_metadata.h"
 #include "transport/transport.h"
@@ -39,25 +45,9 @@
 
 namespace mooncake {
 class TransferMetadata;
+struct ClientSession;
 class TcpContext;
 class TcpTransport;
-
-// Pooled connection for reusing client-side connections
-struct PooledConnection {
-    std::shared_ptr<asio::ip::tcp::socket> socket;
-    std::string host;
-    uint16_t port;
-    std::chrono::steady_clock::time_point last_used;
-    bool in_use;
-
-    PooledConnection(std::shared_ptr<asio::ip::tcp::socket> s,
-                     const std::string &h, uint16_t p)
-        : socket(std::move(s)),
-          host(h),
-          port(p),
-          last_used(std::chrono::steady_clock::now()),
-          in_use(true) {}
-};
 
 class TcpTransport : public Transport {
    public:
@@ -122,10 +112,9 @@ class TcpTransport : public Transport {
     TcpContext *context_;
     std::atomic_bool running_;
     std::thread thread_;
-    bool enable_connection_pool_ =
-        false;  // Use MC_TCP_ENABLE_CONNECTION_POOL=1 to enable connection pool
+    bool enable_connection_pool_ = false;
 
-    // Client-side connection pool
+    // Client-side bounded work queues and fixed connection lanes.
     struct ConnectionKey {
         std::string host;
         uint16_t port;
@@ -142,23 +131,262 @@ class TcpTransport : public Transport {
         }
     };
 
-    std::unordered_map<ConnectionKey,
-                       std::deque<std::shared_ptr<PooledConnection>>,
-                       ConnectionKeyHash>
-        connection_pool_;
-    std::mutex pool_mutex_;
+    enum class WorkFailureReason {
+        QUEUE_FULL,
+        QUEUE_TIMEOUT,
+        RUNTIME_UNAVAILABLE,
+        CONNECT_FAILED,
+        SESSION_FAILED,
+        SHUTDOWN,
+    };
+
+    struct FailureCounters {
+        std::atomic<uint64_t> queue_full{0};
+        std::atomic<uint64_t> queue_timeout{0};
+        std::atomic<uint64_t> connect_failed{0};
+        std::atomic<uint64_t> runtime_unavailable{0};
+        std::atomic<uint64_t> session_failed{0};
+        std::atomic<uint64_t> shutdown{0};
+    };
+
+    struct TcpWorkItem {
+        Slice *slice = nullptr;
+        bool use_v2 = false;
+        std::function<void()> continuation;
+        uint64_t admission_sequence = 0;
+        std::chrono::steady_clock::time_point enqueued_at;
+        std::chrono::steady_clock::time_point admission_deadline;
+
+        TcpWorkItem() = default;
+        TcpWorkItem(Slice *slice_arg, bool use_v2_arg,
+                    std::function<void()> continuation_arg = nullptr)
+            : slice(slice_arg),
+              use_v2(use_v2_arg),
+              continuation(std::move(continuation_arg)) {}
+        TcpWorkItem(TcpWorkItem &&other) noexcept
+            : slice(other.slice),
+              use_v2(other.use_v2),
+              continuation(std::move(other.continuation)),
+              admission_sequence(other.admission_sequence),
+              enqueued_at(other.enqueued_at),
+              admission_deadline(other.admission_deadline) {
+            other.slice = nullptr;
+        }
+        TcpWorkItem &operator=(TcpWorkItem &&) = delete;
+        TcpWorkItem(const TcpWorkItem &) = delete;
+        TcpWorkItem &operator=(const TcpWorkItem &) = delete;
+    };
+
+    struct TerminalAction {
+        TcpWorkItem work;
+        TransferStatusEnum status;
+        bool connection_clean;
+
+        TerminalAction(TcpWorkItem &&work_arg, TransferStatusEnum status_arg,
+                       bool clean_arg)
+            : work(std::move(work_arg)),
+              status(status_arg),
+              connection_clean(clean_arg) {}
+        TerminalAction(TerminalAction &&) noexcept = default;
+        TerminalAction &operator=(TerminalAction &&) = delete;
+        TerminalAction(const TerminalAction &) = delete;
+        TerminalAction &operator=(const TerminalAction &) = delete;
+    };
+
+    static_assert(!std::is_copy_constructible<TcpWorkItem>::value,
+                  "TcpWorkItem must remain move-only");
+    static_assert(!std::is_copy_assignable<TcpWorkItem>::value,
+                  "TcpWorkItem must remain move-only");
+    static_assert(!std::is_copy_constructible<TerminalAction>::value,
+                  "TerminalAction must remain move-only");
+    static_assert(!std::is_copy_assignable<TerminalAction>::value,
+                  "TerminalAction must remain move-only");
+
+    enum class GroupState { OPEN, CLOSING, CLOSED };
+    enum class LaneState {
+        DISCONNECTED,
+        CONNECTING,
+        IDLE,
+        BUSY,
+        COMPLETING,
+        CLOSING,
+        CLOSED,
+    };
+    enum class LaneConnectStage { NONE, RESOLVING, CONNECTING };
+
+    struct ConnectionLaneRuntime {
+        explicit ConnectionLaneRuntime(asio::io_context &io_context)
+            : executor(io_context.get_executor()) {}
+
+        asio::io_context::executor_type executor;
+    };
+
+    struct PeerConnectionGroup;
+
+    struct ConnectionLane {
+        ConnectionLane(size_t lane_id_arg,
+                       const std::shared_ptr<PeerConnectionGroup> &group_arg)
+            : lane_id(lane_id_arg), group(group_arg) {}
+
+        size_t lane_id;
+        std::weak_ptr<PeerConnectionGroup> group;
+        LaneState state = LaneState::DISCONNECTED;
+        LaneConnectStage connect_stage = LaneConnectStage::NONE;
+        uint64_t operation_epoch = 0;
+        uint64_t last_connect_round = 0;
+        std::shared_ptr<asio::ip::tcp::resolver> resolver;
+        std::shared_ptr<asio::ip::tcp::socket> socket;
+        std::shared_ptr<ClientSession> session;
+        std::optional<TcpWorkItem> current;
+    };
+
+    struct PeerConnectionGroup {
+        PeerConnectionGroup(ConnectionKey key_arg,
+                            const asio::io_context::executor_type &executor_arg,
+                            size_t lane_count_arg, size_t queue_capacity_arg,
+                            size_t pending_admission_capacity_arg,
+                            std::chrono::milliseconds admission_timeout_arg,
+                            std::shared_ptr<FailureCounters> counters_arg)
+            : key(std::move(key_arg)),
+              executor(executor_arg),
+              lane_count(lane_count_arg),
+              queue_capacity(queue_capacity_arg),
+              pending_admission_capacity(pending_admission_capacity_arg),
+              admission_timeout(admission_timeout_arg),
+              failure_counters(std::move(counters_arg)) {}
+
+        std::mutex mutex;
+        GroupState state = GroupState::OPEN;
+        ConnectionKey key;
+        asio::io_context::executor_type executor;
+        size_t lane_count;
+        size_t queue_capacity;
+        std::deque<TcpWorkItem> queue;
+        std::deque<TcpWorkItem> pending_admissions;
+        size_t pending_admission_capacity;
+        std::chrono::milliseconds admission_timeout;
+        std::shared_ptr<asio::steady_timer> admission_timer;
+        bool admission_timer_armed = false;
+        uint64_t admission_epoch = 0;
+        uint64_t queued_bytes = 0;
+        bool queued_bytes_saturated = false;
+        uint64_t next_admission_sequence = 1;
+        std::vector<std::shared_ptr<ConnectionLane>> lanes;
+        bool pump_scheduled = false;
+        uint64_t pump_epoch = 0;
+        uint64_t connect_round = 1;
+        size_t connect_attempts_in_round = 0;
+        size_t probes_in_flight = 0;
+        bool connect_round_had_success = false;
+        std::chrono::steady_clock::time_point next_probe_not_before{};
+        std::shared_ptr<asio::steady_timer> retry_timer;
+        uint64_t retry_epoch = 0;
+        uint64_t connect_failure_log_count = 0;
+        std::shared_ptr<FailureCounters> failure_counters;
+    };
+
+    struct ConnectionLaneState {
+        std::mutex mutex;
+        bool shutting_down = false;
+        size_t lanes_per_peer = 4;
+        size_t max_queued_transfers_per_peer = 1024;
+        size_t max_pending_admissions_per_peer = 1024;
+        std::chrono::milliseconds admission_timeout{1000};
+        std::unordered_map<ConnectionKey, std::shared_ptr<PeerConnectionGroup>,
+                           ConnectionKeyHash>
+            groups;
+        std::weak_ptr<ConnectionLaneRuntime> runtime;
+        std::shared_ptr<FailureCounters> failure_counters =
+            std::make_shared<FailureCounters>();
+    };
+
+    std::shared_ptr<ConnectionLaneRuntime> lane_runtime_;
+    std::shared_ptr<ConnectionLaneState> lane_state_;
+
+    // TODO(#2930): The queue item bound does not bound queued source bytes,
+    // and a connected payload without a progress deadline can still stall a
+    // lane indefinitely. Peer-generation recovery is also a separate phase.
 
     std::shared_ptr<asio::ip::tcp::socket> getConnection(
-        const std::string &host, uint16_t port, bool use_pool);
-    void returnConnection(const std::string &host, uint16_t port,
-                          std::shared_ptr<asio::ip::tcp::socket> socket);
-    // Close `socket` and drop it from the pool: used for requests that did
-    // not terminate in a well-defined protocol state (#2086).
-    void discardConnection(const std::string &host, uint16_t port,
-                           std::shared_ptr<asio::ip::tcp::socket> socket);
-    void cleanupIdleConnections();
+        const std::string &host, uint16_t port);
+    void enqueuePooledTransfer(const ConnectionKey &key, TcpWorkItem work);
+    static uint64_t requestGroupPumpLocked(PeerConnectionGroup &group);
+    static void postGroupPump(const std::shared_ptr<PeerConnectionGroup> &group,
+                              uint64_t pump_epoch);
+    static void runGroupPump(const std::shared_ptr<PeerConnectionGroup> &group,
+                             uint64_t pump_epoch);
+    static void startLaneConnect(
+        const std::shared_ptr<PeerConnectionGroup> &group,
+        const std::shared_ptr<ConnectionLane> &lane, uint64_t epoch);
+    static void handleLaneResolved(
+        const std::shared_ptr<PeerConnectionGroup> &group,
+        const std::shared_ptr<ConnectionLane> &lane, uint64_t epoch,
+        asio::error_code ec, asio::ip::tcp::resolver::results_type results);
+    static void handleLaneConnected(
+        const std::shared_ptr<PeerConnectionGroup> &group,
+        const std::shared_ptr<ConnectionLane> &lane, uint64_t epoch,
+        asio::error_code ec);
+    static void handleLaneConnectFailure(
+        const std::shared_ptr<PeerConnectionGroup> &group,
+        const std::shared_ptr<ConnectionLane> &lane, uint64_t epoch,
+        const std::string &error);
+    static bool armRetryTimerLocked(
+        const std::shared_ptr<PeerConnectionGroup> &group);
+    static void handleRetryTimer(
+        const std::shared_ptr<PeerConnectionGroup> &group,
+        const std::shared_ptr<asio::steady_timer> &timer, uint64_t retry_epoch,
+        asio::error_code ec);
+    static size_t expirePendingAdmissionsLocked(
+        PeerConnectionGroup &group, std::chrono::steady_clock::time_point now,
+        std::deque<TcpWorkItem> &expired);
+    static size_t promotePendingAdmissionsLocked(PeerConnectionGroup &group);
+    static void refreshAdmissionTimerLocked(
+        const std::shared_ptr<PeerConnectionGroup> &group,
+        std::deque<TcpWorkItem> &runtime_failed,
+        std::shared_ptr<asio::steady_timer> &timer_to_cancel,
+        bool &timer_armed);
+    static void handleAdmissionTimer(
+        const std::shared_ptr<PeerConnectionGroup> &group,
+        const std::shared_ptr<asio::steady_timer> &timer,
+        uint64_t admission_epoch, asio::error_code ec);
+    static void startLaneSession(
+        const std::shared_ptr<PeerConnectionGroup> &group,
+        const std::shared_ptr<ConnectionLane> &lane, uint64_t epoch);
+    static void handleLaneTerminal(
+        const std::shared_ptr<PeerConnectionGroup> &group,
+        const std::shared_ptr<ConnectionLane> &lane, uint64_t epoch,
+        TransferStatusEnum status, bool connection_clean) noexcept;
+    static void completeTerminalAction(TerminalAction action) noexcept;
+    static void failWorkItem(
+        TcpWorkItem work, WorkFailureReason reason,
+        const std::shared_ptr<FailureCounters> &counters) noexcept;
+    static void failWorkItems(
+        std::deque<TcpWorkItem> work, WorkFailureReason reason,
+        const std::shared_ptr<FailureCounters> &counters) noexcept;
+    static uint64_t recordWorkFailure(
+        WorkFailureReason reason,
+        const std::shared_ptr<FailureCounters> &counters) noexcept;
+    static bool hasUsableLaneLocked(const PeerConnectionGroup &group);
+#ifdef MOONCAKE_TCP_TRANSPORT_TEST_HOOKS
+    static size_t activeSocketCountLocked(const PeerConnectionGroup &group);
+#endif
+    static void beginConnectRoundLocked(PeerConnectionGroup &group);
+    static void enterReconnectCooldownLocked(PeerConnectionGroup &group);
+    static void addQueuedBytesLocked(PeerConnectionGroup &group,
+                                     uint64_t length);
+    static void removeQueuedBytesLocked(PeerConnectionGroup &group,
+                                        uint64_t length);
+    static void clearQueuedBytesLocked(PeerConnectionGroup &group);
+    static void closeSocketNoThrow(
+        const std::shared_ptr<asio::ip::tcp::socket> &socket) noexcept;
+    void shutdownConnectionLanes();
+    void startTransferWithSocket(
+        TcpWorkItem work,
+        std::shared_ptr<asio::ip::tcp::socket> socket) noexcept;
 
-    static constexpr std::chrono::seconds kConnectionIdleTimeout{60};
+#ifdef MOONCAKE_TCP_TRANSPORT_TEST_HOOKS
+    friend bool tcpTransportLaneTypesAreMoveOnlyForTest() noexcept;
+#endif
 };
 }  // namespace mooncake
 
--- a/mooncake-transfer-engine/include/transport/tcp_transport/tcp_transport.h
+++ b/mooncake-transfer-engine/include/transport/tcp_transport/tcp_transport.h
@@ -153,8 +153,6 @@ class TcpTransport : public Transport {
         Slice *slice = nullptr;
         bool use_v2 = false;
         std::function<void()> continuation;
-        uint64_t admission_sequence = 0;
-        std::chrono::steady_clock::time_point enqueued_at;
         std::chrono::steady_clock::time_point admission_deadline;
 
         TcpWorkItem() = default;
@@ -167,8 +165,6 @@ class TcpTransport : public Transport {
             : slice(other.slice),
               use_v2(other.use_v2),
               continuation(std::move(other.continuation)),
-              admission_sequence(other.admission_sequence),
-              enqueued_at(other.enqueued_at),
               admission_deadline(other.admission_deadline) {
             other.slice = nullptr;
         }
@@ -243,13 +239,12 @@ class TcpTransport : public Transport {
     struct PeerConnectionGroup {
         PeerConnectionGroup(ConnectionKey key_arg,
                             const asio::io_context::executor_type &executor_arg,
-                            size_t lane_count_arg, size_t queue_capacity_arg,
+                            size_t queue_capacity_arg,
                             size_t pending_admission_capacity_arg,
                             std::chrono::milliseconds admission_timeout_arg,
                             std::shared_ptr<FailureCounters> counters_arg)
             : key(std::move(key_arg)),
               executor(executor_arg),
-              lane_count(lane_count_arg),
               queue_capacity(queue_capacity_arg),
               pending_admission_capacity(pending_admission_capacity_arg),
               admission_timeout(admission_timeout_arg),
@@ -259,23 +254,19 @@ class TcpTransport : public Transport {
         GroupState state = GroupState::OPEN;
         ConnectionKey key;
         asio::io_context::executor_type executor;
-        size_t lane_count;
         size_t queue_capacity;
         std::deque<TcpWorkItem> queue;
         std::deque<TcpWorkItem> pending_admissions;
         size_t pending_admission_capacity;
         std::chrono::milliseconds admission_timeout;
         std::shared_ptr<asio::steady_timer> admission_timer;
-        bool admission_timer_armed = false;
         uint64_t admission_epoch = 0;
         uint64_t queued_bytes = 0;
         bool queued_bytes_saturated = false;
-        uint64_t next_admission_sequence = 1;
         std::vector<std::shared_ptr<ConnectionLane>> lanes;
         bool pump_scheduled = false;
         uint64_t pump_epoch = 0;
         uint64_t connect_round = 1;
-        size_t connect_attempts_in_round = 0;
         size_t probes_in_flight = 0;
         bool connect_round_had_success = false;
         std::chrono::steady_clock::time_point next_probe_not_before{};
@@ -367,6 +358,9 @@ class TcpTransport : public Transport {
         WorkFailureReason reason,
         const std::shared_ptr<FailureCounters> &counters) noexcept;
     static bool hasUsableLaneLocked(const PeerConnectionGroup &group);
+    static bool hasDisconnectedLaneLocked(const PeerConnectionGroup &group);
+    static bool hasUntriedDisconnectedLaneLocked(
+        const PeerConnectionGroup &group);
 #ifdef MOONCAKE_TCP_TRANSPORT_TEST_HOOKS
     static size_t activeSocketCountLocked(const PeerConnectionGroup &group);
 #endif
--- a/mooncake-transfer-engine/src/transport/tcp_transport/tcp_transport.cpp
+++ b/mooncake-transfer-engine/src/transport/tcp_transport/tcp_transport.cpp
@@ -503,26 +503,69 @@ Transport::Slice* TcpTransport::prepareTransfer(
 
 void TcpTransport::startTransferSequence(std::vector<Slice*> slices) {
     struct Sequence {
+        std::mutex mutex;
         std::vector<Slice*> slices;
         size_t next = 0;
+        bool advancing = false;
+        bool resume_requested = false;
     };
+
     auto sequence = std::make_shared<Sequence>();
     sequence->slices = std::move(slices);
+
     auto advance = std::make_shared<std::function<void()>>();
     std::weak_ptr<std::function<void()>> weak_advance = advance;
+
     *advance = [this, sequence, weak_advance]() {
         auto advance = weak_advance.lock();
-        if (!advance || sequence->next == sequence->slices.size()) return;
-        auto* slice = sequence->slices[sequence->next++];
-        std::function<void()> continuation;
-        if (sequence->next < sequence->slices.size()) {
-            auto executor = context_->io_context.get_executor();
-            continuation = [executor, advance]() mutable {
-                asio::post(executor, [advance]() { (*advance)(); });
-            };
+        if (!advance) return;
+
+        {
+            std::lock_guard<std::mutex> lock(sequence->mutex);
+            if (sequence->next == sequence->slices.size()) return;
+            if (sequence->advancing) {
+                sequence->resume_requested = true;
+                return;
+            }
+            sequence->advancing = true;
         }
-        startTransfer(slice, std::move(continuation), true);
+
+        while (true) {
+            Slice* slice = nullptr;
+            bool has_more = false;
+            {
+                std::lock_guard<std::mutex> lock(sequence->mutex);
+                if (sequence->next == sequence->slices.size()) {
+                    sequence->advancing = false;
+                    return;
+                }
+                slice = sequence->slices[sequence->next++];
+                has_more = sequence->next < sequence->slices.size();
+                sequence->resume_requested = false;
+            }
+
+            std::function<void()> continuation;
+            if (has_more) {
+                continuation = [advance]() { (*advance)(); };
+            }
+
+            // startTransfer may fail synchronously (including during shutdown).
+            // The trampoline above turns a synchronous continuation into
+            // another loop iteration rather than recursive calls. For an
+            // asynchronous completion, this invocation returns and the
+            // continuation becomes the next runner.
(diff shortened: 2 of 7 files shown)


he-yufeng (commented):
Reviewed the connection-pool concurrency design here (I read it against #2930 rather than building the TCP transport locally, so this is a design-level review, not a runtime one).

The core shape is sound. The cap counting `connections.size() + pending_connects.size()` is the right call, since counting only established sockets would let many in-flight connects blow past the cap before any of them finish. FIFO `waiters` with `driveWaitersLocked` gives fair backpressure. The part I looked at hardest is callback re-entrancy, and it holds up: every acquisition callback is invoked outside `state->mutex` (the `reject_for_shutdown` path returns and calls back after the lock scope, and the normal path collects `PoolActions` under the lock and runs them via `enqueuePoolActions` after releasing it), so a callback that re-enters the pool cannot self-deadlock. Rejecting new acquisitions once `shutting_down` and failing queued waiters when the runtime is gone are the right deterministic-shutdown behaviors.

The one property I'd most want the added test to pin down (since it is exactly what an unbounded->bounded pool tends to get subtly wrong): under a shutdown that races in-flight connects, every entry in both `pending_connects` and `waiters` has its callback invoked exactly once, with none leaked (callback dropped) or double-invoked. A stress variant of `tcp_write_visibility_test` that spawns N over-cap acquisitions and calls `shutdownConnectionPool()` mid-connect, asserting the completion count equals N, would lock that in against future refactors.

Nice fix for #2930.

alogfans:
Thanks for working on this issue. Bounding the number of TCP connections is a valuable first step, especially for preventing FD exhaustion and excessive `TIME_WAIT` accumulation under heavy transfers.

One concern is that the current implementation may move part of the resource pressure rather than fully eliminate it. A large transfer can still generate many acquisition callbacks, which then accumulate in `PeerPool::waiters` after the connection limit is reached. This could lead to significant user-space memory consumption and extended task lifetimes, and the resulting backlog may be harder to observe than socket growth.

It may be worth considering whether the connection lifecycle can be modeled differently. Since the number of destination peers is generally bounded, we could maintain a small set of long-lived connection lanes for each peer and let them consume work from a bounded per-peer transfer queue:

`bounded transfer queue -> fixed long-lived connection lanes`

This would apply backpressure when transfer work enters the system, instead of creating one connection-acquisition callback for every slice. It would also make queue depth, waiting time, cancellation, and peer health easier to manage and observe.

Overall, this PR is a useful step toward controlling connection growth. I’d just like us to ensure that the pressure is genuinely bounded rather than shifted from sockets to waiter state.

jacklin78911-collab:
@alogfans 
Thanks — I agree with this assessment.

The current implementation bounds sockets, including in-flight connects, but shifts the remaining unbounded state to `PeerPool::waiters`. Boundedness is therefore not genuinely established, and as you noted, the resulting backlog is harder to observe than socket growth.

A bounded per-peer transfer queue consumed by a small fixed set of long-lived connection lanes looks like the cleaner end-state.

Before changing the branch, could you confirm the preferred scope for this Draft PR?

1. Keep #2974 as a first-step guardrail against FD exhaustion and excessive `TIME_WAIT`, explicitly document the unbounded-waiter limitation, and implement the bounded queue / connection-lane model as a follow-up.

2. Revise this Draft PR around the bounded queue / fixed-lane model before it is considered for merge.

If you prefer the second direction, I will first post a short design note covering the queued work-item granularity, queue-full behavior, transfer failure semantics, reconnect handling, and shutdown invariants before changing the implementation.

I will also carry @he-yufeng's suggested shutdown-race stress test into either design, since the exactly-once completion property remains important.

jacklin78911-collab:
@he-yufeng Thank you for the careful review, especially for validating the re-entrancy reasoning around `PoolActions`.

The exactly-once completion guarantee during a racing shutdown is the key invariant I want to preserve. With the direction discussed above—a bounded per-peer transfer queue consumed by a fixed set of long-lived lanes—I’ll adapt your suggested test to the revised design: submit N tasks beyond the queue/lane capacity, initiate shutdown while some tasks are connecting and others remain queued, and assert that the total completion count is exactly N, with no dropped or duplicate callbacks.

Although the abstraction changes from acquisition callbacks to queued transfer tasks, the invariant remains the same. I’ll include both the deterministic case and a stress variant in the revised test suite.

jacklin78911-collab:
Following the feedback from @alogfans and @he-yufeng, I drafted the lane-based redesign below. This is a design proposal only; no implementation changes have been made yet. Policy choices marked as **pending maintainer confirmation** are intentionally left open.

## Motivation

PR #2974 bounds established and in-progress TCP connections per peer, but excess acquisition callbacks can still accumulate in an unbounded `PeerPool::waiters` queue. This prevents socket growth while shifting pressure into user-space callback state and retained transfer lifetimes.

The proposed replacement is:

```text
bounded per-peer TCP work queue
        ↓
small fixed set of long-lived connection lanes
```

Each lane owns at most one connection and one active transfer. Work remains in the per-peer queue until a connected lane becomes idle, so queued work is never tied to a lane that is still resolving, connecting, or backing off.

The first implementation would address bounded scheduling, connection reuse, and deterministic ownership. It would not claim to solve peer-generation detection, stale connections after peer restart, or connected payloads that make no progress.

## Proposed object model

A shared registry maps each `ConnectionKey` to a `PeerConnectionGroup`.

Each group owns:

* an `OPEN` / `CLOSING` / `CLOSED` state;
* a bounded FIFO queue of `TcpWorkItem`;
* a small fixed vector of `ConnectionLane`;
* group-level pump and reconnect accounting;
* a limit on concurrent connection probes;
* queue-depth, queued-byte, and lane-state metrics;
* an executor obtained from the shared TCP `io_context`.

Each lane owns:

* a `DISCONNECTED`, `CONNECTING`, `BACKOFF`, `IDLE`, `BUSY`, `COMPLETING`, `CLOSING`, or `CLOSED` state;
* an operation epoch for rejecting late handlers;
* resolver, socket, and `ClientSession` objects;
* at most one active `lane.current` WorkItem.

I considered queueing `TransferTask*`, `Slice*`, or a TCP-specific wrapper.

**Pending maintainer confirmation** — I recommend a move-only `TcpWorkItem` wrapping a `Slice*` plus TCP-local scheduling metadata such as protocol selection, admission sequence, and enqueue time. This keeps scheduling metadata out of the transport-wide `Slice` type and remains compatible if one `TransferTask` later contains multiple TCP Slices.

## Task ownership and completion

The design distinguishes:

* a **created** Slice: a `Slice` has been allocated for a `TransferTask`;
* an **API-accepted** task: the existing submission contract accepted the task;
* a **queue-admitted** WorkItem: the WorkItem was successfully inserted into the peer queue.

For queue-admitted work, ownership is:

```text
submitter-local WorkItem
    → group.queue
    → lane.current
    → local TerminalAction
    → one Slice terminal call
```

Every transition is a move. Both `TcpWorkItem` and `TerminalAction` would be explicitly non-copyable.

`ClientSession` would no longer directly call `Slice::markSuccess()` or `Slice::markFailed()`. Instead, it would report one unified result to its owning lane:

```cpp
on_terminal(TransferStatusEnum status, bool connection_clean)
```

The lane would:

1. validate its operation epoch;
2. move `lane.current` into a local `TerminalAction` under the group mutex;
3. change to `COMPLETING`;
4. release the mutex;
5. close or retain the socket according to `connection_clean`;
6. mark the Slice exactly once through the claimed `TerminalAction`;
7. make the lane reusable only after the terminal call finishes;
8. schedule another group pump when appropriate.

No Slice completion, external callback, metric observer, or test hook would run while holding the group mutex.

## Admission and queue-full policy

I considered two queue-full policies:

1. return a synchronous retryable overload `Status`;
2. preserve the existing API acceptance model and immediately fail the created Slice through the existing terminal-status path.

**Pending maintainer confirmation** — I recommend the second policy.

The current API submits a vector and returns one aggregate `Status`. Other transports may already have accepted part of the batch, and a synchronous overload result would not identify which individual entries are safe to retry.

Under the recommended policy:

* the task is API-accepted;
* its Slice has been created;
* its WorkItem is **not** queue-admitted;
* the group retains no WorkItem;
* no rejection handler is posted;
* the submitting thread fails the Slice immediately after releasing the group mutex;
* public transfer status remains `FAILED`; an internal `QUEUE_FULL` reason is available to tests, metrics, and logs.

This avoids recreating an unbounded queue of asynchronous rejection handlers.

The item-count limit bounds scheduler objects, but it does not bound the total source-buffer lifetime represented by queued transfers. Queued bytes should therefore be tracked from the first implementation.

**Pending maintainer confirmation** — the first implementation could enforce the WorkItem-count bound while exposing queued-byte accounting, with an actual byte-cap policy added once its semantics and default are agreed.

## Lane behavior

A `DISCONNECTED` lane attempts to resolve and connect only while queued work exists. It does not remove work from the queue while connecting.

After connection succeeds, the lane becomes `IDLE`. An `IDLE` lane moves the FIFO head into `lane.current` and becomes `BUSY`.

When `ClientSession` reports a terminal result:

* a clean connection may return to `IDLE`;
* a dirty or failed connection is closed and the lane becomes `DISCONNECTED`;
* the current Slice is completed exactly once;
* mid-transfer work is never transparently replayed on another lane.

Reconnect behavior would remain deliberately limited in the first implementation:

* reconnect is demand-driven while queued work exists;
* attempts are bounded within a group-level round;
* attempts are rate-limited at group level;
* only a small number of lanes probe concurrently;
* if the round is exhausted and no lane is usable, remaining queued work is detached and failed rather than left indefinitely stranded.

**Pending maintainer confirmation** — exact retry count, delay/backoff sequence, and maximum concurrent probe count.

Peer-generation detection and full stale-peer recovery remain follow-up work.

## Shutdown model

Shutdown follows one ownership order:

1. stop admission;
2. mark groups `CLOSING`;
3. detach queued WorkItems under their group mutexes;
4. fail detached queued work outside the mutexes;
5. cancel lane operations on the executor when possible;
6. account for posted and asynchronous handlers;
7. wait up to a quiescence deadline;
8. stop the shared `io_context`;
9. join the TCP worker;
10. reset `ClientSession`, socket, resolver, and timer objects while `TcpContext` still exists;
11. complete deferred current WorkItems only after no asynchronous operation can still access their source buffers;
12. destroy group/runtime ownership;
13. finally destroy `TcpContext`.

Late handlers validate operation epochs and cannot claim ownership that has already moved. Async handlers capture group/lane state rather than a raw `TcpTransport*`.

**Pending maintainer confirmation** — exact quiescence deadline. The deadline would bound how long normal asynchronous draining is attempted before fallback. It would not create a hard destructor-time guarantee, because `io_context.stop()` cannot interrupt a handler that is already executing and blocked.

## Configuration proposals

**Pending maintainer confirmation:**

* introduce a small configurable lane count per peer, with a tentative default of 4;
* introduce a configurable per-peer WorkItem queue bound;
* decide whether `MC_TCP_MAX_CONNECTIONS_PER_PEER` should remain temporarily as a deprecated compatibility alias;
* decide whether the first implementation should enforce both an item-count bound and a queued-byte bound, or enforce item count first while tracking queued bytes.

The exact environment-variable names, ranges, and queue defaults remain open.

## Shutdown stress test (as requested by @he-yufeng)

The stress test would use many concurrent submissions with a small lane count and queue capacity while deterministic hooks hold some lanes in `CONNECTING` and others in `BUSY`, then race transport destruction against the workload.

Assertions:

* queue depth never exceeds its bound;
* resolving, connecting, and connected sockets never exceed the lane count;
* under the recommended queue-full policy, overflow tasks are API-accepted but never queue-admitted;
* every created Slice for an API-accepted task becomes terminal exactly once;
* no task remains `WAITING`;
* late handlers cannot claim moved WorkItems;
* no WorkItem, resolver, socket, timer, or `ClientSession` remains attached after shutdown.

The full suite would run normally and under ASan/LSan/UBSan. Targeted TSan tests would cover group ownership, shutdown, late handlers, and terminal transitions. The unrelated baseline Slice-completion/status-polling race would remain documented separately rather than claiming a globally clean TSan run.

## Core invariants

1. Queue size never exceeds its configured bound.
2. Resolving, connecting, and connected sockets never exceed the lane count.
3. Queued work belongs to no lane.
4. A connecting lane owns no WorkItem.
5. Each lane owns at most one current WorkItem.
6. `WorkItem` and `TerminalAction` are non-copyable.
7. Only a claimed `TerminalAction` may terminally complete a Slice.
8. `ClientSession` reports status and connection cleanliness but never marks the Slice.
9. No terminal completion runs under the group mutex.
10. A dirty socket is closed before terminal status becomes visible.
11. Reconnect work is demand-driven, bounded, and rate-limited.
12. Exhausted queued work is failed rather than stranded.
13. No async handler dereferences `TcpTransport` after destruction begins.
14. The item-count limit does not imply a queued-byte bound.

## Decisions to confirm before changing the implementation

1. Whether #2974 should be revised around this lane model before merge.
2. Whether the recommended API-accepted / immediate-Slice-failure queue-full policy matches the intended submission contract.
3. Whether a TCP-specific move-only `TcpWorkItem` around `Slice*` is the preferred queue granularity.

alogfans:
@jacklin78911-collab I prefer option 2.

One key point: we should not model TCP as one socket object per Slice. Socket ownership should be per destination peer: the transport maintains a bounded set of socket fd lanes for each target, and Slices are just work items scheduled onto those lanes.

So the model I’d prefer is:

bounded per-target work queue -> fixed socket fd lanes per target

This gives us a real bound on socket resources and avoids simply moving pressure from sockets to per-Slice waiter objects.

jacklin78911-collab:
@alogfans  Thanks, understood. I’ll revise #2974 around option 2.

The ownership model will be:

```text
per-destination PeerConnectionGroup
    → bounded Slice work queue
    → fixed socket-fd lanes
```

Sockets, resolvers, and reconnect state will belong to the per-destination lanes. A Slice will not acquire or own a socket; it will remain queued until an idle connected lane pulls it for execution.

I’ll remove the acquisition-callback / waiter model from the revised implementation and preserve the exactly-once completion and shutdown invariants in the new queue → lane → completion ownership chain.

alogfans (commented):
Thanks, the latest implementation is much closer to the model I had in mind.

BTW, could we consider an admission-timeout policy instead of failing immediately when the per-target queue is full? I would prefer not to block submitTransfer synchronously. A better model may be: keep a bounded pending-admission queue with a deadline timer. When the work queue has capacity, pending Slices are promoted in FIFO order; if the admission deadline expires first, the Slice is completed as FAILED with an internal QUEUE_TIMEOUT reason.

mooncake-transfer-engine/include/transport/tcp_transport/tcp_transport.h
```diff
@@ -132,23 +120,228 @@ class TcpTransport : public Transport {
...
-                       ConnectionKeyHash>
-        connection_pool_;
-    std::mutex pool_mutex_;
+    enum class WorkFailureReason {
+        QUEUE_FULL,
+        RUNTIME_UNAVAILABLE,
+        CONNECT_FAILED,
+        SESSION_FAILED,
+        SHUTDOWN,
+    };
```
alogfans, line 141: Is this exposable to external users?

alogfans:
Additionally, we recommend to add unit tests for tcp transport to show its reliability, and the interoperationality of previous version.

jacklin78911-collab:
@alogfans Thanks, understood.

For admission, I’ll add a bounded per-target pending-admission FIFO in front
of the bounded work queue. Each pending Slice will carry a deadline; capacity
release promotes pending admissions in FIFO order, while expiry completes the
Slice as FAILED with `QUEUE_TIMEOUT`. Promotion, timeout, and shutdown will be
serialized under the group lock, with terminal completion outside the lock.

Regarding `WorkFailureReason`: it is currently internal, and external callers
only observe a generic FAILED Slice. I agree that the reason should be
observable. Would you prefer exposing it through the public transfer status,
or keeping the API stable and surfacing it through metrics/logs?

I’ll also extend the tests for admission promotion/timeout races, shutdown,
reconnect behavior, and exactly-once completion.

For interoperability, I already have partial legacy-protocol coverage, but I
want to confirm the expected scope: should the test use a legacy peer fixture,
or build and run the TCP transport from a previous Mooncake commit against the
new implementation?

mooncake-transfer-engine/include/transport/tcp_transport/tcp_transport.h
jacklin78911-collab: Currently no. It is internal to `TcpTransport`; external callers only observe
the Slice as `FAILED`.

I agree that `QUEUE_TIMEOUT` should be externally observable. I’ll check the
existing transfer-status API and propose the smallest way to expose it without
leaking the transport-internal enum directly.

jacklin78911-collab:
@alogfans Follow-up: I’ve implemented the admission-timeout revision.

The per-target client path is now:

bounded pending-admission FIFO
→ bounded work queue
→ fixed socket-fd lanes

Pending Slices are promoted in FIFO order when work-queue capacity becomes
available. A single per-group timer tracks the earliest admission deadline;
expiry completes the Slice with the internal `QUEUE_TIMEOUT` reason. The
pending-admission queue also remains bounded, with `QUEUE_FULL` used only when
that hard bound is reached.

Timeout, promotion, and shutdown ownership are serialized under the group
mutex, while terminal Slice completion remains outside the lock. I added
deterministic tests for both orderings:

- timeout wins before capacity release;
- capacity release promotes first, and the cancelled timer handler becomes
  stale without double-completing the Slice.

For interoperability, I added a pooled legacy-framing initiator → current
server test covering both WRITE and READ with data verification. The existing
stale-descriptor tests cover current v2 initiator → legacy server safe-failure
behavior. I have not added a cross-build test using a binary from an older
Mooncake commit in this revision; please let me know if that is the intended
scope.

`WorkFailureReason` remains internal for now, and I left the public transfer
status API unchanged pending your preference on how the reason should be
exposed.

Local validation:
- normal: 24/24
- ASan/LSan/UBSan: 24/24
- targeted TSan admission/retry/shutdown subset: 10/10
- production BUILD_UNIT_TESTS=OFF build passed with no test-hook symbols

The PR remains Draft while the updated CI runs.

jacklin78911-collab:
The `tent-ci (cuda-off)` job failed because the runner ran out of disk space
(`No space left on device`). Could someone with Actions permission please
re-run the failed job?

alogfans:
@jacklin78911-collab You can try to submit one empty commit to trigger CI at another time. BTW it can be ignored for this PR.

jacklin78911-collab:
@alogfans Thanks, understood. Since this was a runner disk-space failure and can be
ignored for this PR, I’ll leave the branch unchanged for now. I’ll trigger a
new CI run with an empty commit later if needed.

jacklin78911-collab:
@alogfans @doujiang24 @chestnut-Q @staryxchen

The branch has been rebased onto current main, and the updated CI is green.

The implementation and PR description now reflect the final bounded-lane and
pending-admission design. The remaining scope question is whether external
per-Slice failure reasons are required in this PR; I proposed keeping the
public result as generic FAILED and handling a transport-agnostic reason API in
a follow-up.

I am also adding a small main-vs-lanes performance characterization to make the
default lane-count tradeoff explicit.

The main correctness areas for review are:

1. admission/promotion/timeout ownership;
2. lane terminal completion versus shutdown;
3. reconnect cooldown and late timer handlers.

mooncake-transfer-engine/src/transport/tcp_transport/tcp_transport.cpp
```diff
@@ -1106,220 +1326,1366 @@ void TcpTransport::worker() {
...
+            const size_t probe_limit =
+                std::min(group->lane_count, kMaxConcurrentLaneProbes);
+            while (!group->queue.empty() &&
+                   group->probes_in_flight < probe_limit) {
+                auto lane_it = std::find_if(
+                    group->lanes.begin(), group->lanes.end(),
+                    [&group](const auto& lane) {
+                        return lane->state == LaneState::DISCONNECTED &&
+                               lane->last_connect_round != group->connect_round;
+                    });
```
alogfans: A failed lane cannot reconnect while another lane remains usable: its last_connect_round still equals the current round, while a new round is started only when no usable lane exists. This can permanently reduce a peer from N lanes to one after transient session failures. Please make the failed lane independently eligible for reconnect and add a two-lane recovery test.

staryxchen:
@jacklin78911-collab Thanks for the contribution! Have you run any e2e test to compare the performance of this PR with the original code?

jacklin78911-collab:
@staryxchen Thanks! I haven’t run a controlled e2e comparison against main yet. I’ll run the same workload on main and this branch and report throughput/latency and socket-count results here.

jacklin78911-collab:
@alogfans Thanks again for the careful review and suggestions.
I did another correctness pass over the async ownership and shutdown paths, and found an additional task-group shutdown corner case. When the active member of a transfer sequence completed after io_context shutdown had started, advancing the remaining sequence members could be posted to an executor that would no longer run, leaving the tail of the sequence pending.
I fixed this by replacing the sequence advancement path with a non-recursive trampoline and added a deterministic regression test covering exactly-once terminal completion of the whole task group during shutdown.
I also ran clang-format 20 on the changed files. All CI checks are green now.
Thanks for taking another look when you have time.

alogfans:
Looks good. Have you ever check both correctness and performance of new tcp transport?

jacklin78911-collab:
@alogfans Thanks for the suggestion. I ran both correctness validation and a performance comparison for the new TCP lane path.

### Correctness

I ran `transfer_engine_validator` with the connection pool enabled:

- `MC_TCP_ENABLE_CONNECTION_POOL=1`
- `MC_TCP_LANES_PER_PEER=4`
- 12 worker threads
- 30s duration

The WRITE/READ data-integrity validation completed successfully:

```text
Test completed: duration 30.07, batch count 4356, throughput 1.22 GB/s
Data validation passed
```

No data mismatches were detected.

### Performance

I also compared the legacy TCP path with the new bounded-lane path using `transfer_engine_bench` on the same single-node loopback setup.

Common configuration:

- 12 submission threads
- Batch size: 128
- Block size: 64 KiB
- Duration: 30s per run
- 3 runs for each configuration

Legacy TCP:

```text
MC_TCP_ENABLE_CONNECTION_POOL=0
```

Bounded-lane TCP:

```text
MC_TCP_ENABLE_CONNECTION_POOL=1
MC_TCP_LANES_PER_PEER=4
```

Results:

| Operation | Legacy TCP | 4-lane TCP |
| --- | --- | --- |
| READ | 0.97 / 1.03 / 1.03 GB/s | 2.39 / 2.26 / 2.33 GB/s |
| WRITE | 1.09 / 1.09 / 1.07 GB/s | 2.64 / 2.64 / 2.67 GB/s |

Median throughput:

- READ: `1.03 -> 2.33 GB/s` (~2.26x)
- WRITE: `1.09 -> 2.64 GB/s` (~2.42x)

This is a single-node loopback sanity/performance regression test rather than a cross-node network benchmark, so I would not generalize the speedup to real network deployments. Under this workload, however, the bounded-lane path did not introduce a throughput regression, and connection reuse substantially improved throughput.

alogfans:
@staryxchen Could you help me review the code again?

mooncake-transfer-engine/src/transport/tcp_transport/tcp_transport_lane_impl.h
```diff
@@ -0,0 +1,1383 @@
...
+                              : LaneState::CLOSING;
+
+            const bool sibling_usable =
+                group->state == GroupState::OPEN && hasUsableLaneLocked(*group);
+            if (sibling_usable) {
+                // A lane-local connect failure must not consume this lane's
+                // eligibility for the rest of a round while another lane
+                // continues serving the peer. Delay the new probe to avoid a
+                // tight loop when the peer temporarily refuses extra lanes.
+                lane->last_connect_round = 0;
```
staryxchen: **Major — lane starvation in reconnect probe selection.** When a lane fails to connect while a sibling is usable, `last_connect_round` is reset to `0` (here and at line 1089) to keep the lane eligible for the current round. But probe selection at line 647-652 uses `find_if` for the first DISCONNECTED lane with `last_connect_round != connect_round`. Since `0 != connect_round`, the reset lane (lowest index) is always picked before any never-tried sibling (which also starts at `0`).

Failure scenario (4 lanes, `kMaxConcurrentLaneProbes=1`): L2 BUSY, L1 probe fails → `last_connect_round=0` → next pump `find_if` returns L1 again, not L3/L4. As long as L1 keeps failing and L2 stays usable, L3/L4 are never probed, capping the pool at 1 lane. Not a premature-failure (a fresh round starts when the usable sibling dies), but workloads needing >1 lane will be persistently limited.

Suggested fix: use a sentinel (e.g. `UINT64_MAX`) to distinguish "retry-eligible this round" from "never tried", or track distinct lanes tried instead of raw `connect_attempts_in_round` for the round-exhaustion check.

mooncake-transfer-engine/src/transport/tcp_transport/tcp_transport_lane_impl.h
```diff
@@ -0,0 +1,1383 @@
...
+        std::deque<TcpWorkItem> failed;
+        std::shared_ptr<asio::steady_timer> admission_timer;
+        [[maybe_unused]] size_t promoted = 0;
+        {
+            std::lock_guard<std::mutex> lock(group->mutex);
+            if (group->pump_scheduled && group->pump_epoch == pump_epoch) {
+                group->pump_scheduled = false;
+                failed.swap(group->queue);
+                clearQueuedBytesLocked(*group);
+                promoted = promotePendingAdmissionsLocked(*group);
```
staryxchen: **Minor — `fail_posted_work` can orphan Slices on allocation failure.** `failed.swap(group->queue)` (line 513) moves the queue into local `failed`, then `promotePendingAdmissionsLocked` (here) moves pending items back so they can also be drained. If `promotePendingAdmissionsLocked` throws (e.g. `deque::emplace_back` `bad_alloc`), the exception escapes the lambda and `failWorkItems(std::move(failed), ...)` at line 540 is never reached. The Slices in `failed` are silently dropped — never `markFailed`, continuations never invoked — and `pending_admissions` is left un-drained.

Low probability but a correctness-impactful orphan path. Suggested fix: wrap the promotion+drain in try/catch, or use a scope guard so `failWorkItems` runs even on throw.

mooncake-transfer-engine/src/transport/tcp_transport/tcp_transport_session_impl.h
```diff
@@ -0,0 +1,822 @@
...
+struct SessionHeader {
+    uint64_t size;
+    uint64_t addr;
+    uint8_t opcode;
+};
+
+#if defined(USE_CUDA) || defined(USE_MUSA) || defined(USE_HIP) ||  \
+    defined(USE_MLU) || defined(USE_MACA) || defined(USE_HYGON) || \
+    defined(USE_COREX)
+static bool isCudaMemory(void* addr) {
```
staryxchen: **Minor — dead code.** `isCudaMemory` has no references anywhere in `session_impl.h`, `tcp_transport.cpp`, or `lane_impl.h`. Appears orphaned when the code was extracted from the old `tcp_transport.cpp`. Consider removing.

mooncake-transfer-engine/src/transport/tcp_transport/tcp_transport.cpp
```diff
@@ -851,21 +250,48 @@ TcpTransport::TcpTransport() : context_(nullptr), running_(false) {
...
     }
-}
 
-TcpTransport::~TcpTransport() {
-    if (running_) {
-        running_ = false;
-        context_->io_context.stop();
-        thread_.join();
-    }
+    if (enable_connection_pool_) {
```
staryxchen: **Minor — lane config silently ignored when pool disabled but task groups used.** Lane config env vars (`MC_TCP_LANES_PER_PEER`, etc.) are parsed inside `if (enable_connection_pool_)`. But `submitTransferTaskGroup` can still route through the lane model via `reuse_connection = true` even when the pool is disabled. In that case the lane model runs with hardcoded defaults (4 lanes, 1024 queue, 1024 pending, 1000ms timeout) and the user's `MC_TCP_LANES_PER_PEER=8` is silently ignored.

Suggested fix: either parse lane config regardless of `enable_connection_pool_`, or document that lane config only applies when `MC_TCP_ENABLE_CONNECTION_POOL=1`.

mooncake-transfer-engine/include/transport/tcp_transport/tcp_transport.h
```diff
@@ -142,23 +131,262 @@ class TcpTransport : public Transport {
...
+        std::atomic<uint64_t> runtime_unavailable{0};
+        std::atomic<uint64_t> session_failed{0};
+        std::atomic<uint64_t> shutdown{0};
+    };
+
+    struct TcpWorkItem {
+        Slice *slice = nullptr;
+        bool use_v2 = false;
+        std::function<void()> continuation;
+        uint64_t admission_sequence = 0;
```
staryxchen: **Nit — dead state `admission_sequence` / `next_admission_sequence`.** `next_admission_sequence` is incremented and `admission_sequence` assigned onto every work item, but neither is ever read. FIFO order is already preserved by `std::deque`. Looks like scaffolding for a priority/dedup mechanism that was never wired up. Consider removing both fields and the assignment line.

staryxchen (commented):
Thanks for the revision — the bounded-lane ownership model and epoch-based stale detection are well designed. Below are focused findings from a code review of the diff.

Verified locally (P2P loopback, `transfer_engine_bench`, threads=4/batch=32/duration=10s, 3 runs each):

| config | READ (GB/s) | WRITE (GB/s) |
|---|---|---|
| PR2974 legacy (POOL=0) | 0.51 / 0.48 / 0.49 | 0.52 / 0.55 / 0.52 |
| PR2974 4-lane (POOL=1) | 1.71 / 1.51 / 1.59 | 1.49 / 1.48 / 1.43 |
| main 4-lane (POOL=1) | 1.54 / 1.42 / 1.65 | 1.80 / 1.45 / 1.42 |

No throughput regression vs main; 4-lane vs legacy ~3x in our low-load setup (directionally consistent with the ~2.3x the author reported at 12 threads).

mooncake-transfer-engine/include/transport/tcp_transport/tcp_transport.h
```diff
@@ -142,23 +131,262 @@ class TcpTransport : public Transport {
...
+        std::chrono::steady_clock::time_point enqueued_at;
```
staryxchen: **Nit — `enqueued_at` read exactly once, could be a local.** Set on every item, read exactly once immediately to compute `admission_deadline = work.enqueued_at + group->admission_timeout`. The local `admission_time` already holds the value. Never read for metrics or latency reporting. Consider dropping the field and computing `admission_deadline = admission_time + group->admission_timeout` directly.

mooncake-transfer-engine/include/transport/tcp_transport/tcp_transport.h
```diff
@@ -142,23 +131,262 @@ class TcpTransport : public Transport {
...
+        size_t queue_capacity;
+        std::deque<TcpWorkItem> queue;
+        std::deque<TcpWorkItem> pending_admissions;
+        size_t pending_admission_capacity;
+        std::chrono::milliseconds admission_timeout;
+        std::shared_ptr<asio::steady_timer> admission_timer;
+        bool admission_timer_armed = false;
+        uint64_t admission_epoch = 0;
+        uint64_t queued_bytes = 0;
+        bool queued_bytes_saturated = false;
```
staryxchen, line 265: **Nit — `queued_bytes` / `queued_bytes_saturated` are observability-only accounting with non-trivial complexity.** These fields never influence an admission decision — the TODO at tcp_transport.h:306-308 explicitly acknowledges queued bytes are not bounded. They feed only test-hook observability. The saturation-recompute loop in `removeQueuedBytesLocked` re-sums the entire queue on every removal after overflow — intricate logic for a counter that gates nothing. First-principles: if bytes don't bound admission, the accounting is not load-bearing state. Consider dropping byte accounting entirely, or replacing the saturating recompute with a simpler "recompute from scratch on removal when saturated" that doesn't recurse through `addQueuedBytesLocked`.

mooncake-transfer-engine/include/transport/tcp_transport/tcp_transport.h
```diff
@@ -142,23 +131,262 @@ class TcpTransport : public Transport {
...
+              queue_capacity(queue_capacity_arg),
+              pending_admission_capacity(pending_admission_capacity_arg),
+              admission_timeout(admission_timeout_arg),
+              failure_counters(std::move(counters_arg)) {}
+
+        std::mutex mutex;
+        GroupState state = GroupState::OPEN;
+        ConnectionKey key;
+        asio::io_context::executor_type executor;
+        size_t lane_count;
```
staryxchen: **Nit — `lane_count` duplicates `lanes.size()`.** Set once at construction from `state->lanes_per_peer` and never mutated. `lanes.size()` is always equal. The code already uses both interchangeably (e.g. `lane_count` at lane_impl.h:644 vs `lanes.size()` at lane_impl.h:614, 668, 903), which is inconsistent. Consider dropping `lane_count` and using `lanes.size()`.

mooncake-transfer-engine/include/transport/tcp_transport/tcp_transport.h
```diff
@@ -142,23 +131,262 @@ class TcpTransport : public Transport {
...
+        size_t queue_capacity;
+        std::deque<TcpWorkItem> queue;
+        std::deque<TcpWorkItem> pending_admissions;
+        size_t pending_admission_capacity;
+        std::chrono::milliseconds admission_timeout;
+        std::shared_ptr<asio::steady_timer> admission_timer;
+        bool admission_timer_armed = false;
```
staryxchen: **Nit — `admission_timer_armed` is redundant with `admission_timer != nullptr`.** Tracing every assignment, `admission_timer_armed` is set true exactly when `admission_timer` becomes non-null and false exactly when it becomes null (under `group->mutex`). The `admission_timer_armed || admission_timer` check at lane_impl.h:166 is defensive against a divergence that cannot occur while `group->state == OPEN`. `handleAdmissionTimer`'s staleness check (`!admission_timer_armed || admission_timer != timer`, lane_impl.h:231) reduces to `admission_timer != timer` for the same reason. Consider dropping `admission_timer_armed` and keying logic on `admission_timer` alone.

mooncake-transfer-engine/src/transport/tcp_transport/tcp_transport_lane_impl.h
```diff
@@ -0,0 +1,1383 @@
...
+        std::lock_guard<std::mutex> lock(group->mutex);
+        if (group->state != GroupState::OPEN ||
+            group->admission_epoch != admission_epoch ||
+            !group->admission_timer_armed || group->admission_timer != timer) {
+            late = true;
+        } else {
+            group->admission_timer.reset();
+            group->admission_timer_armed = false;
+            fired = true;
+            if (ec) {
```
staryxchen, line 242: **Nit — unreachable defensive code.** Line 237 (`if (ec)` in `handleAdmissionTimer`): every path that cancels the admission timer also increments `admission_epoch` under lock, so a cancelled timer's callback always fails the epoch check at line 230 first and never reaches the `ec` branch.

Similarly at line 322-323 (`if (group->state != GroupState::OPEN)` in `handleRetryTimer`): shutdown increments `retry_epoch` atomically with setting state, so the epoch check at line 318 already catches shutdown. This state check is unreachable.

Harmless belt-and-suspenders, but obscures the real state transitions. Consider trimming or adding a comment that they're defensive.

jacklin78911-collab:
@staryxchen Thanks for the detailed review and for running the additional performance comparison.

I pushed a follow-up addressing the findings from this review:

- fixed reconnect probe selection so failed low-index lanes cannot starve never-tried lanes within the same round;
- preserved fresh-round recovery when the last usable session becomes dirty;
- hardened the posted-work and shutdown failure paths so queued/pending Slices are detached and completed without allocation-sensitive ownership gaps;
- parse lane configuration independently of whether the connection pool is enabled;
- removed the unused `isCudaMemory`, `admission_sequence`, `enqueued_at`, `lane_count`, `admission_timer_armed`, and other dead reconnect state;
- kept `queued_bytes` for queue observability/testing, and kept the defensive timer guards with comments clarifying their purpose.

I also added deterministic regression coverage for reconnect starvation and the single-lane dirty-session recovery case.

Local validation after the changes:
- `tcp_write_visibility_test`: 29/29 passed;
- `transfer_engine_validator` with 4 TCP lanes, 12 threads, batch size 128, and 64 KiB blocks: all 12 workers reported `Data validation passed` over a 30s run.

Thanks again for the careful review.

staryxchen (approved):
Good job, thanks for your effort! LGTM.

PR #2978, opened 2026-07-17
[TENT] Fix metrics HTTP server falsely reporting success on port bind failure

anranxia:
## Description

  TentMetrics::initHttpServer() was declared void and dropped the return value of coro_http_server::async_start().
  TentMetrics::initialize() then unconditionally logged "HTTP server listening on <host>:<port>" and returned Status::OK(),
  so a bind failure was silently reported as a successful init — even though the caller in transfer_engine_impl.cpp already
  branches on status.ok(), it never received a failure.

  This is easy to hit under tensor parallelism: every rank on a host defaults to the same metrics port (9100), so only one
  rank can bind while the others (e.g. 7/8 at TP=8) "succeed" in the logs with a dead /metrics endpoint.

## Fix
  - initHttpServer() now returns Status and checks async_start().hasResult() — the same idiom already used in
  mooncake-store/src/rpc_service.cpp and real_client.cpp (a ready future means startup failed).
  - On a busy port it walks the adjacent ports in [http_port, http_port + kHttpPortMaxRetry), so co-located TP ranks each
  grab a distinct port automatically. The port actually bound is written back to config_.http_port and exposed via the new
  TentMetrics::httpPort() getter.
  - If every candidate port is busy, initialize() no longer prints the misleading "listening" line. It degrades gracefully
  to log-only metrics (in-process counters and the periodic summary thread need no port) and logs a WARNING describing the
  unavailable endpoint.
  - HTTP handler registration is factored into registerHttpHandlers() so it can be re-applied to each server instance
  created during port retry.

  Note for operators: with port retry, co-located ranks land on shifting ports (9100, 9101, …). The bound port is logged per
  rank, so Prometheus scrape configs should target the [9100, 9100+N) range accordingly.

## Module

  - [ ] CI/CD
  - [ ] Docs
  - [ ] Other

## Type of Change

  - [x] Bug fix
  - [ ] New feature
  - [ ] Refactor
  - [ ] Breaking change
  - [ ] Documentation update
  - [ ] Performance improvement
  - [ ] Other

## How Has This Been Tested?

Verified on a Linux host (Ubuntu 24.04, gcc 13.3) against the real yalantinglibs/coro_http headers, with
  TENT_METRICS_ENABLED=1 (TENT metrics are OFF by default, so this TU is not in a default build).

**Test commands:**
```bash
  # 1) Added unit test — occupies the configured metrics port, then asserts
  #    initialize() retries onto a free adjacent port instead of claiming success.
  #    Built and run standalone (metrics ON, header-only asio):
  g++ -std=c++20 -O1 -DTENT_METRICS_ENABLED=1 \
      -I mooncake-transfer-engine/tent/include -I /usr/local/include \
      -I extern/yalantinglibs/include -I extern/yalantinglibs/include/ylt/thirdparty \
      mooncake-transfer-engine/tent/tests/metrics_http_server_test.cpp \
      mooncake-transfer-engine/tent/src/metrics/tent_metrics.cpp \
      mooncake-transfer-engine/tent/src/common/status.cpp \
      /usr/lib/x86_64-linux-gnu/libgtest.a /usr/lib/x86_64-linux-gnu/libgtest_main.a \
      -lglog -lpthread -latomic -o /tmp/tent_metrics_http_server_test
  /tmp/tent_metrics_http_server_test

 # 2) The metrics TU also compiles cleanly with codegen at -O2:
  g++ -std=c++20 -c -O2 -DTENT_METRICS_ENABLED=1 \
      -I mooncake-transfer-engine/tent/include -I /usr/local/include \
      -I extern/yalantinglibs/include -I extern/yalantinglibs/include/ylt/thirdparty \
      mooncake-transfer-engine/tent/src/metrics/tent_metrics.cpp -o /tmp/tent_metrics.o
```

**Test results:**
  - [x] Unit tests pass

port-contention scenario reproduced end-to-end against the real coro_http_server: the configured port is skipped, a fallback is bound, and the "listening" log reflects the true port.

## Checklist

## AI Assistance Disclosure

  - [ ] No AI tools were used

The implementation was AI-assisted and independently reviewed across multiple models before submission. The diagnosis and fix were verified against the source and the regression tests were run locally (results above).

### DIFF
--- a/mooncake-transfer-engine/tent/tests/metrics_http_server_test.cpp
+++ b/mooncake-transfer-engine/tent/tests/metrics_http_server_test.cpp
@@ -1,0 +1,103 @@
+// Copyright 2025 KVCache.AI
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <gtest/gtest.h>
+
+#include "tent/metrics/tent_metrics.h"
+
+#if TENT_METRICS_ENABLED
+
+#include <arpa/inet.h>
+#include <netinet/in.h>
+#include <sys/socket.h>
+#include <unistd.h>
+
+namespace mooncake {
+namespace tent {
+namespace {
+
+// RAII helper that binds and listens on an OS-assigned port, thereby keeping
+// that port busy for the duration of the test.
+class PortOccupier {
+   public:
+    PortOccupier() {
+        fd_ = ::socket(AF_INET, SOCK_STREAM, 0);
+        EXPECT_GE(fd_, 0);
+        sockaddr_in addr{};
+        addr.sin_family = AF_INET;
+        addr.sin_addr.s_addr = htonl(INADDR_ANY);
+        addr.sin_port = 0;  // let the kernel pick a free port
+        EXPECT_EQ(::bind(fd_, reinterpret_cast<sockaddr*>(&addr), sizeof(addr)),
+                  0);
+        EXPECT_EQ(::listen(fd_, 1), 0);
+
+        sockaddr_in bound{};
+        socklen_t len = sizeof(bound);
+        EXPECT_EQ(::getsockname(fd_, reinterpret_cast<sockaddr*>(&bound), &len),
+                  0);
+        port_ = ntohs(bound.sin_port);
+    }
+
+    ~PortOccupier() {
+        if (fd_ >= 0) ::close(fd_);
+    }
+
+    uint16_t port() const { return port_; }
+
+   private:
+    int fd_ = -1;
+    uint16_t port_ = 0;
+};
+
+// When the configured metrics port is already in use (e.g. another rank was
+// mistakenly given the same port), initialize() must NOT falsely report a
+// listening endpoint. It degrades to log-only metrics: init still succeeds,
+// but the HTTP server is not bound, so httpPort() reports 0.
+TEST(TentMetricsHttpServer, DegradesToLogOnlyWhenConfiguredPortBusy) {
+    PortOccupier occupier;
+    ASSERT_GT(occupier.port(), 0);
+
+    MetricsConfig config;
+    config.enabled = true;
+    config.http_host = "127.0.0.1";
+    config.http_port = occupier.port();
+    config.report_interval_seconds = 0;  // no periodic reporting thread
+
+    auto& metrics = TentMetrics::instance();
+    Status status = metrics.initialize(config);
+
+    // Metrics subsystem still comes up (counters + log summary work without a
+    // port), but the scrape endpoint is not listening on the busy port.
+    EXPECT_TRUE(status.ok()) << status.ToString();
+    EXPECT_TRUE(metrics.isInitialized());
+    EXPECT_EQ(metrics.httpPort(), 0);
+
+    metrics.shutdown();
+}
+
+}  // namespace
+}  // namespace tent
+}  // namespace mooncake
+
+#else  // !TENT_METRICS_ENABLED
+
+// With metrics compiled out, initialize() is a no-op stub. Assert only that it
+// reports success so the test target still builds and runs in default builds.
+TEST(TentMetricsHttpServer, DisabledAtCompileTime) {
+    mooncake::tent::MetricsConfig config;
+    EXPECT_TRUE(
+        mooncake::tent::TentMetrics::instance().initialize(config).ok());
+}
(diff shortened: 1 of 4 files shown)


staryxchen (commented):
The bug fix itself — returning `Status` from `initHttpServer()` and checking `async_start().hasResult()` — is correct and clearly necessary. Silently reporting success on a bind failure was the real defect.

I'd raise a design concern about the port-retry fallback, though.

Port assignment is a deployment/topology concern, not a metrics-library concern. Auto-fallback makes the bound port depend on startup race order — rank 0 may land on 9100 today and 9103 tomorrow. The PR note acknowledges operators must scrape `[9100, 9100+N)`, but a range scrape can't tell which rank is on which port, so per-rank debugging actually gets harder than with deterministic ports. Explicit per-rank configuration (`base_port + local_rank_id`) is the established pattern in distributed training (NCCL, DDP) and keeps port assignment with the caller that knows the topology.

By Occam's razor, the retry path adds real machinery (64-port scan, `kHttpPortMaxRetry`, atomic `bound_http_port_`, `httpPort()` getter, `registerHttpHandlers()` refactor) for a scenario explicit configuration avoids entirely.

Suggested alternative:
1. Keep the bug fix as-is.
2. Keep the log-only degradation when the configured port is busy — that's the right graceful degradation for an auxiliary endpoint.
3. Drop the port-scan retry; document that co-located ranks should be assigned distinct ports explicitly (e.g. `base_port + local_rank`).

This keeps the fix minimal, the degradation graceful, and leaves port assignment deterministic and in the right layer.

anranxia:
Thanks, this makes sense — agreed that port assignment belongs in the deployment layer, not the metrics library, and that auto-fallback makes the bound port depend on startup race order (bad for per-rank scraping/debugging).

  Updated the PR to your suggested shape:

  - Kept the core fix: `initHttpServer()` returns `Status` and checks `async_start().hasResult()`.
  - Kept graceful **log-only degradation** when the configured port is busy (no false "listening" log).
  - **Dropped the port-scan retry** — `initHttpServer()` now binds only the configured port. Port assignment stays
  deterministic; co-located ranks should be given distinct ports explicitly (e.g. `base_port + local_rank`), which I noted in the code comments.
  - `httpPort()` reports the bound port (`0` when in log-only mode).

  The unit test now asserts that a busy configured port degrades to log-only (`httpPort() == 0`) instead of falsely
  reporting success. Re-verified against the real `coro_http_server`.

mooncake-transfer-engine/tent/tests/metrics_http_server_test.cpp
```diff
@@ -0,0 +1,103 @@
+// Copyright 2025 KVCache.AI
```
staryxchen:
```suggestion
// Copyright 2026 KVCache.AI
```

staryxchen (approved):
Thanks for the fix, LGTM

staryxchen:
@anranxia Could you confirm whether that failed CI job is related to this PR?

anranxia:
> @anranxia Could you confirm whether that failed CI job is related to this PR?

  The CI failure looks flaky and unrelated to this PR.

  The failing job `test-wheel-ubuntu (ubuntu-24.04, 3.12)` fails at the **"Run tests with ssd"** step — a mooncake-store SSD runtime test. That exact step passed on other recent runs of this workflow (e.g. the store-touching PR run `29579700305`), so the step itself isn't broken.

  Meanwhile on this PR:
  - `build (3.12)` passed and the wheel artifact built/downloaded fine,
  - both `tent-ci (cuda-on)` and `tent-ci (cuda-off)` passed,
  - `Check code format` passed.

  This change is C++ confined to `mooncake-transfer-engine/tent/` and is behind the `TENT_METRICS_ENABLED` flag (OFF by default), so it isn't exercised by the wheel build/SSD tests. Could you re-run the failed job? It looks like a flaky SSD test rather than a regression from this PR.

anranxia:
> @anranxia Could you confirm whether that failed CI job is related to this PR?

The failing job `test-wheel-ubuntu (ubuntu-24.04, 3.12)` dies in the "CLI entry point tests" while starting `mooncake_master`:

  E http_metadata_server.cpp:121] Failed to start HTTP metadata server on 0.0.0.0:8080
  F master.cpp:1449] Failed to start HTTP metadata server
  ❌ mooncake_master process failed to start

  Port 8080 was already in use — the log shows another master already `role=leader, state=serving` in the same env, so the entry-point test's second `mooncake_master` couldn't bind 8080. That's a test-env port/cleanup race in mooncake-store, not this change.

staryxchen:
> > @anranxia Could you confirm whether that failed CI job is related to this PR?
> 
> The failing job `test-wheel-ubuntu (ubuntu-24.04, 3.12)` dies in the "CLI entry point tests" while starting `mooncake_master`:
> 
> E http_metadata_server.cpp:121] Failed to start HTTP metadata server on 0.0.0.0:8080 F master.cpp:1449] Failed to start HTTP metadata server ❌ mooncake_master process failed to start
> 
> Port 8080 was already in use — the log shows another master already `role=leader, state=serving` in the same env, so the entry-point test's second `mooncake_master` couldn't bind 8080. That's a test-env port/cleanup race in mooncake-store, not this change.

OK, I will fix this flaky CI job in another PR.

PR #2986, opened 2026-07-17
[Store] Initialize file-per-key storage with eviction disabled

### LINKED ISSUE #2981: [Bug]: File-per-key storage fails to initialize when eviction is disabled
## Bug Report

On current `main` , the file-per-key backend cannot complete `FileStorage::Init()` when `MOONCAKE_OFFLOAD_ENABLE_EVICTION=false`.

The call order is:

1. `storage_backend_->Init()`
2. `FileStorage::IsEnableOffloading()`
3. `storage_backend_->ScanMeta()`

For `StorageBackendAdaptor`, `IsEnableOffloading()` returns `INTERNAL_ERROR` while `meta_scanned_` is false when eviction is disabled. The first metadata scan is later in `FileStorage::Init()`, so initialization returns before it can ever establish the counters required by `IsEnableOffloading()`.

The adaptor unit test already documents that `ScanMeta()` must run before `IsEnableOffloading()` in this configuration, but the production initialization path does the reverse.

## Expected behavior

File-per-key storage should initialize successfully with eviction disabled, reconstruct existing usage before enabling new offloads, and register recovered LOCAL_DISK metadata with the master.

### Before submitting...

- [x] Searched existing issues and open PRs for this initialization failure.

feichai0017:
## Description

Fixes #2981.

- Treat file-per-key offloading as unavailable until metadata has been scanned, and enforce the same state in `BatchOffload()`.
- Refresh the local state after recovery so the first heartbeat publishes the final eviction-disabled state to the master.

## Module

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

**Test commands:**
```bash
pre-commit run --files mooncake-store/src/file_storage.cpp mooncake-store/src/storage_backend.cpp mooncake-store/tests/file_storage_test.cpp mooncake-store/tests/storage_backend_test.cpp
cmake --build build-focused --target storage_backend_test file_storage_test
./build-focused/mooncake-store/tests/storage_backend_test --gtest_brief=1
./build-focused/mooncake-store/tests/file_storage_test --gtest_brief=1
```

**Test results (Ubuntu):**
- [x] Unit tests pass (`storage_backend_test`: 89/89; `file_storage_test`: 24/24)
- [x] Integration tests pass (file-per-key initialization with eviction disabled)
- [ ] Manual testing done

## Checklist

- [ ] I have run `pre-commit run --all-files` and all hooks pass (touched-file hooks pass)

## AI Assistance Disclosure

- [ ] No AI tools were used
- [x] AI tools were used: Codex assisted with investigation, implementation, testing, and review.


he-yufeng (commented):
The direction is right: treating "metadata not scanned yet" as a hard `INTERNAL_ERROR` was too aggressive for a backend that only knows its quota after the initial scan, so mounting conservatively-disabled and refreshing after the scan in `Init()` is a cleaner startup contract. Two things worth a look before merge.

**1. `IsEnableOffloading()` now collapses two distinct states into `false`.** Before, "not scanned yet" surfaced as an error, distinct from a genuine "offloading is disabled". After this change both return `false`, so a caller can no longer tell "not ready yet, retry later" from "the operator turned offloading off". For the two callers here that's fine — `BatchOffload` rejects either way, and `Init()` calls it only after `ScanMeta` has set `meta_scanned_` — but the method is `virtual` with several overrides and is also read on the heartbeat path (`file_storage.cpp:326`, which still logs an error branch). Worth confirming no consumer relied on the error to mean "startup incomplete" rather than "disabled by config", since that signal is now gone.

**2. `KEYS_ULTRA_LIMIT` is a confusing code for "disabled".** When offloading is off, `BatchOffload` returns `ErrorCode::KEYS_ULTRA_LIMIT`, which reads as "the key/size cap was hit" — a capacity condition — rather than "offloading isn't enabled". A caller (or someone reading logs) could easily misdiagnose a disabled backend as a full one. If there's an `OFFLOAD_DISABLED`-style code (or room to add one) it would make the rejection self-explanatory; at minimum a one-line comment at the return site would help.

Neither blocks the fix. #1 is the one I'd most like confirmed, since the lost distinction is easy to depend on implicitly on the heartbeat/startup path.

feichai0017:
@he-yufeng Thanks for flagging both points. 7b3f6463 now reports the pre-scan state as `UNABLE_OFFLOADING`, which `FileStorage::Init()` treats as transient while mounting conservatively disabled. After the scan, `false` remains a quota condition and `BatchOffload()` returns `KEYS_ULTRA_LIMIT`; both states are covered by tests.

feichai0017:
Synced this branch with current `main` and revalidated the exact head on Ubuntu: `file_storage_test` 24/24 and `storage_backend_test` 89/89. The pre-scan state remains distinct as `UNABLE_OFFLOADING`, and the quota-disabled path is covered. Current CI has no failures. @he-yufeng, could you please take another look when convenient?

PR #2989, opened 2026-07-18
[TENT] Fix metrics recording zero-overhead, MLU sentinel, and parallel-vector contract

staryxchen:
## Description

Fixes six design / correctness issues found while reviewing the TENT metrics system (`tent/metrics/`), plus a docs sync. Each change is test-driven.

**Correctness:**
- **Zero-overhead claim was false on the hot path.** `recordTaskCompletionMetrics` called `TentMetrics::instance().record*` directly (not via macros), so when `TENT_METRICS_ENABLED=OFF` every task completion still hit no-op stubs. The whole recording block is now `#if`-gated so the body is empty when disabled, matching the doc claim.
- **MLU sentinel poisoned real data.** An infeasible-at-submit deadline was recorded as `MLU=5.0` into the `tent_deadline_mlu_permille` histogram, indistinguishable from a genuine 5x-over-window sample. Replaced with a dedicated `tent_deadline_infeasible_total` counter.

**Simplicity:**
- **Unused `bytes` parameter.** `recordReadFailed(size_t bytes)` / `recordWriteFailed(size_t bytes)` ignored the argument (failed transfers transfer no bytes). Dropped the parameter and updated all call sites.
- **Parallel-vector ordering contract.** `histograms_` and `histogram_boundaries_` were two vectors whose order had to match by convention; a future contributor updating one without the other would silently produce mislabeled JSON. Collapsed into a single `vector<HistogramEntry>` pairing each histogram with its boundaries.
- **Runtime-configurable buckets removed.** `initialize()` reassigned `histogram_t` members mid-flight when the config supplied `latency_buckets`/`size_buckets` — a code smell that also produced mismatched JSON labels. Buckets are now fixed at compile time (version-controlled), making observability reproducible. (Net **-201 lines**.)

**Docs / build:**
- Synced `metrics.md` (added 6 missing metrics, removed the bucket config, corrected the zero-overhead claim, clarified `*_requests_total` counts success+failure).
- Moved `option(TENT_METRICS_ENABLED)` from the leaf metrics CMakeLists to `tent/CMakeLists.txt` for discoverability.

## Module

- [ ] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Breaking change
- [x] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

Three test layers, all green.

**Test commands:**
```bash
# Configure (metrics enabled + tests + tent)
cmake -B build -S mooncake-transfer-engine \
  -DUSE_TENT=ON -DTENT_METRICS_ENABLED=ON -DBUILD_UNIT_TESTS=ON \
  -Dpybind11_INCLUDE_DIR=<extern/pybind11/include>
cmake --build build -j

# Unit + HTTP integration tests
./build/tent/tests/tent_metrics_recording_test      # 9 tests
./build/tent/tests/metrics_config_loader_test        # 26 tests
./build/tent/tests/tent_metrics_http_server_test     # 1 test
./build/tent/tests/causal_chain_test                 # 4 tests (no regression)

# Disabled-build compile check (verifies the #if gates)
cmake -B build_disabled -S mooncake-transfer-engine \
  -DUSE_TENT=ON -DTENT_METRICS_ENABLED=OFF -DBUILD_UNIT_TESTS=ON \
  -Dpybind11_INCLUDE_DIR=<extern/pybind11/include>
cmake --build build_disabled -j --target tent_metrics_recording_test
./build_disabled/tent/tests/tent_metrics_recording_test
```

**Test results:**
- [x] Unit tests pass — recording (9) + config loader (26) + HTTP server (1) + causal chain (4) = 40 tests.
- [x] Integration tests pass — HTTP endpoints (`/metrics`, `/metrics/json`, `/health`) scraped via `coro_http_client`; Prometheus/JSON output asserts on counters and histogram buckets.
- [x] Disabled-build compiles cleanly and the stub test passes (validates the zero-overhead `#if` gating).

The new test file `metrics_recording_test.cpp` uses delta-based assertions (snapshot before/after) so it is robust to the `TentMetrics` singleton's non-resettable counter state, and drives a real `TransferEngineImpl` with a `FakeTransport` to exercise the actual `recordTaskCompletionMetrics` path.

## Checklist

- [x] I have run `pre-commit` on the files touched by this change and all hooks pass (ran on touched files per AGENTS.md guidance; `--all-files` not run to avoid unrelated edits)
- [ ] For changes >500 LOC: I have filed an RFC issue — *net non-test code change is negative (simplification); the 665-line test file is the bulk of the insertions. Filing left to submitter's judgment.*

## AI Assistance Disclosure

- [ ] No AI tools were used

CodeBuddy (GLM-5.2) was used to implement the fixes, write the tests, and draft this PR. The human submitter has reviewed every changed line and can defend the changes end-to-end.

### DIFF
--- a/mooncake-transfer-engine/tent/include/tent/metrics/config_loader.h
+++ b/mooncake-transfer-engine/tent/include/tent/metrics/config_loader.h
@@ -31,8 +31,6 @@ struct MetricsConfig {
     uint32_t report_interval_seconds = 30;  // 0 means disabled
     bool enable_prometheus = true;
     bool enable_json = true;
-    std::vector<double> latency_buckets;
-    std::vector<double> size_buckets;
 };
 
 // Helper class to load metrics configuration from various sources
@@ -74,10 +72,6 @@ constexpr const char* METRICS_REPORT_INTERVAL =
 constexpr const char* METRICS_ENABLE_PROMETHEUS = "metrics/enable_prometheus";
 constexpr const char* METRICS_ENABLE_JSON = "metrics/enable_json";
 
-// Bucket configurations
-constexpr const char* METRICS_LATENCY_BUCKETS = "metrics/latency_buckets";
-constexpr const char* METRICS_SIZE_BUCKETS = "metrics/size_buckets";
-
 // Environment variable names (with TENT_ prefix)
 constexpr const char* ENV_METRICS_ENABLED = "TENT_METRICS_ENABLED";
 constexpr const char* ENV_METRICS_HTTP_PORT = "TENT_METRICS_HTTP_PORT";
@@ -89,9 +83,6 @@ constexpr const char* ENV_METRICS_REPORT_INTERVAL =
 constexpr const char* ENV_METRICS_ENABLE_PROMETHEUS =
     "TENT_METRICS_ENABLE_PROMETHEUS";
 constexpr const char* ENV_METRICS_ENABLE_JSON = "TENT_METRICS_ENABLE_JSON";
-constexpr const char* ENV_METRICS_LATENCY_BUCKETS =
-    "TENT_METRICS_LATENCY_BUCKETS";
-constexpr const char* ENV_METRICS_SIZE_BUCKETS = "TENT_METRICS_SIZE_BUCKETS";
 }  // namespace config_keys
 
 }  // namespace tent
--- a/mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
+++ b/mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
@@ -2426,33 +2426,36 @@ void TransferEngineImpl::recordTaskCompletionMetrics(
                                                   transport_us);
                     }
                 }
-                // Observability only (RFC #2519): if this transfer carried a
-                // deadline, emit the post-hoc feasibility ratio MLU =
-                // actual_transfer_time / available_window, where the window is
-                // (deadline - submit_time). MLU < 1 met the deadline; >= 1
-                // missed it. This does not drive any admission/scheduling yet.
-                if (task.request.deadline_ns != 0) {
-                    uint64_t start_ns = static_cast<uint64_t>(
-                        std::chrono::duration_cast<std::chrono::nanoseconds>(
-                            start_time.time_since_epoch())
-                            .count());
-                    if (task.request.deadline_ns > start_ns) {
-                        double window_seconds =
-                            (task.request.deadline_ns - start_ns) / 1e9;
-                        TentMetrics::instance().recordDeadlineMLU(
-                            latency_seconds / window_seconds);
-                    } else {
-                        // Deadline already in the past at submit: infeasible.
-                        // Recorded into a dedicated counter so it does not
-                        // pollute the MLU histogram with a sentinel value.
-                        TentMetrics::instance().recordDeadlineInfeasible();
-                    }
-                }
             } else if (new_status == FAILED) {
                 if (task.request.opcode == Request::READ) {
                     TentMetrics::instance().recordReadFailed();
                 } else {
                     TentMetrics::instance().recordWriteFailed();
+                }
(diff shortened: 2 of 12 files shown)


mooncake-transfer-engine/tent/include/tent/metrics/config_loader.h
```diff
@@ -74,10 +72,6 @@ constexpr const char* METRICS_REPORT_INTERVAL =
 constexpr const char* METRICS_ENABLE_PROMETHEUS = "metrics/enable_prometheus";
 constexpr const char* METRICS_ENABLE_JSON = "metrics/enable_json";
 
-// Bucket configurations
-constexpr const char* METRICS_LATENCY_BUCKETS = "metrics/latency_buckets";
-constexpr const char* METRICS_SIZE_BUCKETS = "metrics/size_buckets";
```
alogfans, line 79: Could we explicitly document this as a compatibility change and provide a migration note?

mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
```diff
@@ -2444,22 +2443,23 @@ void TransferEngineImpl::recordTaskCompletionMetrics(
                             latency_seconds / window_seconds);
                     } else {
                         // Deadline already in the past at submit: infeasible.
-                        TentMetrics::instance().recordDeadlineMLU(5.0);
+                        // Recorded into a dedicated counter so it does not
+                        // pollute the MLU histogram with a sentinel value.
+                        TentMetrics::instance().recordDeadlineInfeasible();
```
alogfans: Only invoked when transfer is completed. This may differ from the origin purpose, i.e., Transfers whose deadline was already in the past at submit.

staryxchen: Good catch. Hoisted the deadline-feasibility block out of the COMPLETED branch (7c91f214) so `recordDeadlineInfeasible()` now fires for both COMPLETED and FAILED transfers — the infeasible-at-submit condition is a property of submit time, not outcome. The feasible MLU recording stays gated on COMPLETED since it requires the actual transfer latency. Added `InfeasibleDeadlineRecordedOnFailedTransfer` test (FakeTransport with force_fail) covering the FAILED + past-deadline path.

mooncake-transfer-engine/tent/include/tent/metrics/config_loader.h
staryxchen: Thanks for flagging this. You're right that it's a behavioral change for deployments with non-default custom buckets — I've added a migration note to `metrics.md` (Compatibility section).

On the rationale for removing the knob rather than fixing it in place: the configurable-buckets path had a latent correctness bug — `initialize()` reassigned the histogram members with the configured buckets, but the boundaries vector (used to label the JSON output) kept the compile-time defaults, so `/metrics/json` emitted default boundary keys over custom-bucket counts. Existing custom-bucket users were already getting mislabeled output. Rather than fix the sync (which is awkward now that histograms and boundaries are paired in a struct — custom buckets would need separate lifetime-managed storage), we chose to remove the feature: it's niche (metrics is default OFF, custom buckets even more so), it was buggy, and fixed compile-time buckets make observability reproducible across deployments (same binary → same histogram boundaries → shared dashboards work).

Migration for anyone who did set custom buckets: the config keys (`metrics/latency_buckets`, `metrics/size_buckets`) and env vars (`TENT_METRICS_LATENCY_BUCKETS`, `TENT_METRICS_SIZE_BUCKETS`) are now silently ignored. If custom buckets are genuinely needed, edit `kLatencyBuckets`/`kSizeBuckets` in `tent/metrics/tent_metrics.h` and rebuild.

alogfans (approved):
LGTM

PR #2990, opened 2026-07-18
[Bugfix][Store] Reject invalid metadata client IDs

VectorPeak:
## Description

Reject malformed `client_id` values when `MasterService::MetadataSerializer` deserializes object metadata.

### What Problem This Solves

`DeserializeMetadata()` read the serialized `client_id` string and called `StringToUuid(client_id_str, client_id)`, but ignored the boolean result. If snapshot metadata contained a malformed client ID such as `not-a-uuid`, deserialization continued and could construct `ObjectMetadata` with an invalid UUID value instead of treating that entry as corrupt metadata.

A nearby sibling path already handles this strictly: the catalog-backed snapshot provider checks `StringToUuid()` and returns `DESERIALIZE_FAIL` for invalid client UUID strings. Replica msgpack deserialization also rejects invalid client IDs. This PR brings the master metadata deserializer in line with those paths.

### Changes

- Check the return value from `StringToUuid()` in `DeserializeMetadata()`.
- Return `DESERIALIZE_FAIL` with a targeted error message when the serialized metadata client ID is malformed.
- Add a focused snapshot metadata regression test that builds a payload with `client_id = not-a-uuid` and verifies the object key is not restored into metadata.

The existing shard-level behavior is intentionally preserved: `DeserializeShard()` logs a per-object metadata deserialization failure and continues with the remaining entries.

### Evidence

Before this change, the invalid client ID parse result was ignored:

```cpp
UUID client_id;
StringToUuid(client_id_str, client_id);
```

The new test exercises the restore deserialization entry point with a malformed client ID payload:

```cpp
BuildMetadataPayloadWithClientIdString(
    not-a-uuid, std::string_view(key.data(), key.size()))
```

and asserts that the bad entry is skipped instead of being restored:

```cpp
EXPECT_FALSE(KeyExistsInMetadata(service_.get(), key));
```

### Possible call chain / impact

```text
snapshot metadata bytes
  -> MasterService::MetadataSerializer::Deserialize()
  -> DeserializeShard()
  -> DeserializeMetadata()
  -> StringToUuid(client_id_str, client_id)
```

This affects the master metadata restore/deserialization path for malformed persisted metadata. It does not change serialization output, valid UUID handling, replica parsing, or the shard-level policy that skips individual corrupt object entries.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
git diff --check
cmake -S . -B build-store-test -DBUILD_UNIT_TESTS=ON -DWITH_STORE=ON -DWITH_TE=ON -DWITH_STORE_RUST=OFF -DWITH_EP=OFF -DUSE_CUDA=OFF -DUSE_ETCD=OFF -DSTORE_USE_ETCD=OFF -DSTORE_USE_REDIS=OFF -DCMAKE_BUILD_TYPE=RelWithDebInfo
```

**Test results:**
- [ ] Unit tests pass

`git diff --check` passed. The focused CMake configure was attempted in local WSL, but this machine is missing the `yaml-cpp` development package, so the local build/test target could not be configured here.

## Checklist

`./scripts/code_format.sh` / pre-commit were not run locally because the Windows checkout exposes the shell script with CRLF line endings in WSL and `pre-commit` is not installed in this environment.

## AI Assistance Disclosure

- [ ] No AI tools were used

Used ChatGPT Codex to inspect the deserialization path, implement the small guard and regression test, and draft this PR description. The submitter is responsible for reviewing and defending the change.

### DIFF
--- a/mooncake-store/src/master_service.cpp
+++ b/mooncake-store/src/master_service.cpp
@@ -7852,7 +7852,12 @@ MasterService::MetadataSerializer::DeserializeMetadata(
     // Deserialize client_id string
     std::string client_id_str = array[index++].as<std::string>();
     UUID client_id;
-    StringToUuid(client_id_str, client_id);
+    if (!StringToUuid(client_id_str, client_id)) {
+        return tl::unexpected(SerializationError(
+            ErrorCode::DESERIALIZE_FAIL,
+            fmt::format("deserialize ObjectMetadata invalid client_id UUID: {}",
+                        client_id_str)));
+    }
 
     // Deserialize put_start_time
     uint64_t put_start_time_timestamp = array[index++].as<uint64_t>();
--- a/mooncake-store/tests/ha/snapshot/snapshot_child_process_test.cpp
+++ b/mooncake-store/tests/ha/snapshot/snapshot_child_process_test.cpp
@@ -670,10 +670,9 @@ TEST_F(SnapshotChildProcessTest, DeserializeMetadataSkipsInvalidClientId) {
     CreateDefaultService();
     const std::string key = "invalid_client_id_snapshot_key";
 
-    auto deserialize_result = DeserializeMetadataForTest(
-        BuildMetadataPayloadWithClientIdString("not-a-uuid",
-                                               std::string_view(key.data(),
-                                                                key.size())));
+    auto deserialize_result =
+        DeserializeMetadataForTest(BuildMetadataPayloadWithClientIdString(
+            "not-a-uuid", std::string_view(key.data(), key.size())));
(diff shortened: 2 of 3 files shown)


Icedcoco (approved):
LGTM. This correctly handles StringToUuid() failures during metadata deserialization and adds focused regression coverage.

PR #2991, opened 2026-07-19
[Store] Add recoverable client liveness and asynchronous offboarding

Doube-Suzerain:
## Description

Implements the client-liveness design proposed in RFC #2953.

This PR adds recoverable Store Client liveness and moves terminal Client
resource withdrawal to a dedicated asynchronous offboarding worker. A
temporary heartbeat gap transitions one Client incarnation from `Active` to
`Suspected`, retaining its resources while removing them from serving paths; a
valid liveness signal can recover it. Cleanup begins only after the separate
suspicion TTL expires and the incarnation becomes terminally `Offline`.

The serving decision is applied consistently across allocator registrations,
descriptors, Copy/Move, task and drain routing, LocalDisk/offload, and
promotion. Existing completion, revoke, and failure callbacks may still
converge work while a Client is Suspected. NoF keeps its independent health
model.

Terminal cleanup is represented by a retryable `ClientOffboardingJob` and
executed by a dedicated `ClientOffboardingWorker`. The worker performs
invalid-handle cleanup, memory Segment commit, LocalDisk withdrawal, quota
convergence, and compare-and-erase. In non-oplog mode, scheduled snapshots are
deferred while offboarding is pending; batch-oplog mode orders accepted
metadata cleanup before Segment unmount.

Shutdown does not drain or persist queued offboarding jobs. Master first stops
and joins its process-local snapshot producer while the pending barrier remains
raised, then stops the worker. An in-flight attempt leaves its critical section;
queued and backoff jobs are discarded with a structured summary. This prevents
the exiting process from publishing a new partial snapshot after discarding
work, but it is not a durable handoff protocol: a future leader starts from the
already durable snapshot/oplog prefix.

Compatibility is preserved:

- Ping keeps the existing `OK` / `NEED_REMOUNT` wire behavior.
- `client_live_ttl_sec` and `--client_ttl` remain supported aliases for the
  active TTL.
- HA restore rebuilds process-local liveness records and resource bindings
  without changing the snapshot schema.
- Standby-restored memory remains behind Mooncake's existing invalid-endpoint
  gate until its actual owner successfully ReMounts.
- ReMount readiness remains separate from liveness and Segment lifecycle.
- Multiple Segment IDs may continue sharing the same Client hostname/name.
- The PR has no dependency on #2957 and does not change the existing remote
  HTTP metadata cleanup protocol.

## Module

- [ ] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [x] New feature
- [x] Refactor
- [ ] Breaking change
- [x] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

- The complete configured build, including all benchmark and E2E targets,
  compiles and links successfully.
- After rebasing onto `main@408b831b`, all 21 related CTest targets pass,
  including liveness/config, allocation/region driver, Segment, MasterService,
  SSD/offload, metrics/admin, promotion, HA, and batch-oplog snapshot targets.
- `high_availability_test` passes its enabled supervisor case; two ETCD cases
  are skipped because the current build has no ETCD HA backend.
- Focused regressions cover standby memory before/after actual-owner ReMount,
  same-name reverse unmount, stale-handle cleanup, Copy/Move lock ordering, and
  no-op offload error classification from current `main`.
- `git diff --check` passes.

## Checklist

- [x] For changes >500 LOC: I have filed an RFC issue (#2953)

`clang-format-20` is unavailable in the local validation environment; the
repository format check passed on the previous pushed head and will rerun for
this update.

## AI Assistance Disclosure

- [ ] No AI tools were used

OpenAI Codex helped refine the design against RFC #2953 and review feedback,
implement the state machine, serving gates, asynchronous offboarding worker,
HA/snapshot coordination, tests and documentation, diagnose regression
failures, run validation, and draft this PR description. The human submitter is
responsible for reviewing every changed line and being able to defend the
change end-to-end before marking this PR ready for review.


Icedcoco:
Thanks for putting this together. The overall `Active -> Suspected -> Offline` direction makes sense. I have a few questions about some failure and concurrency paths that I may be missing.

1. In `ClientOffboardingWorker::ThreadFunc`, an incomplete job decrements `pending_jobs_` and is then discarded. `ShouldSkipSnapshot()` preserves only one additional skipped snapshot cycle. If `PrepareUnmountSegment` or `CommitUnmountSegment` fails, is there another path that retries or completes this job?

   My concern is that the `Offline` record remains permanently in the current process, while a later snapshot may persist partially offboarded resources. Since restore rebuilds those resources with fresh `Active` liveness records, could a failover make them serving again? Would it be safer to keep the job retryable and retain the snapshot barrier until cleanup converges, or otherwise persist the incomplete offboarding state?

2. `MountSegment` and `MountLocalDiskSegment` call `ObserveAndRun()` before checking the actual mount result. As written, a Suspected client becomes Active even when the mount later returns an error such as `INVALID_PARAMS` or `INTERNAL_ERROR`. The RFC says a successful mount is a liveness signal. Could the observation/recovery be committed only after the mount returns `OK` or a confirmed idempotent `SEGMENT_ALREADY_EXISTS` result?

3. The HTTP metadata cleanup event contains only `client_id` and segment names. The old liveness record is erased before the event is asynchronously processed, so a new incarnation may remount the same segment name and republish `ram/<segment_name>` or `rpc_meta/<segment_name>` first. Could the queued cleanup then delete the new metadata? Do we have an ordering guarantee that prevents this, or should the deletion carry an incarnation/version check? This may be better addressed in the stacked #2957 PR if that is where the cleanup contract is owned.

Separately, the branch currently has substantial conflicts with `main`, especially in the core Store and snapshot paths. The comparison also shows recent `main` changes as removals; for example, the `ssd_total_capacity_bytes` snapshot support from #2927 is absent from this head. Could you rebase and resolve the stacked #2957 commits and current `main` changes first, while preserving those newer fixes? I think another review after that would be much easier and would avoid discussing behavior that may change during conflict resolution.

Icedcoco:
Thanks for the detailed implementation. I went through the current head once more and found a few edge cases I may be missing:

1. ScopedSegmentAccess::MountSegment now rejects a new segment ID when the segment name already exists. However, clients use local_hostname_ as the name for multiple mounts, and RealClient may split one allocation into several segments. Would this cause the second mount to fail with INVALID_PARAMS? Should duplicate names remain supported, or should the client generate unique names?

2. The asynchronous HTTP metadata cleanup currently queues only segment_name. If the name is released and a new client incarnation mounts the same segment before the worker runs, could the old cleanup delete the new incarnation's metadata? Would it be safer for the cleanup request to carry an incarnation/version token?

3. MountLocalDiskSegment creates a liveness record, while stale-handle cleanup still builds its alive-client set from ok_client_. Since ok_client_ is populated during remount, could a healthy client be treated as stale during the interval before its first remount and have its LOCAL_DISK replicas removed? Should this path use the liveness records directly?

4. ShouldSkipSnapshotForClientOffboarding() skips snapshots whenever any offboarding job is pending, while failed jobs are retried indefinitely. If one cleanup operation remains permanently unsuccessful, could this prevent all future snapshots and leave HA state stale? Should the snapshot barrier be bounded or otherwise expose a fallback policy?

Also, the current Check code format job is still failing. Plz address that along with the points above.

Doube-Suzerain:
@Icedcoco Thanks for taking another look. You were right about the first and third paths. I addressed them in `fb43bae4`:

1. The segment-name rejection was a regression introduced by my idempotency validation. `RealClient` can split one allocation into several Segment IDs while using `local_hostname_` as the name for every mount. I removed only the name-uniqueness rejection and restored Mooncake's original behavior: a retry with the same Segment ID is still accepted only after validating the same owner and segment identity, while different Segment IDs may share a name. I did not change Client naming. The segment and host-index tests now cover mounting and independently unmounting multiple IDs under one name.

2. The metadata race is real for the separately deployed remote HTTP backend. Co-located cleanup is synchronous, but the remote path on current `main` already queues only `segment_name` and later removes the same unversioned `ram/<segment_name>` and `rpc_meta/<segment_name>` keys. This PR keeps that existing cleanup contract; it does not introduce the asynchronous name-only deletion. A complete fix needs an incarnation/version in the metadata protocol and conditional deletion. I would prefer to track that independently rather than add a liveness-local fallback that can still race.

3. Fixed. `ok_client_` remains the ReMount/wire-readiness state only. The stale-handle cleanup paths now derive their retaining-client set from `ClientLivenessRecord::ShouldRetainResources()`, so both `Active` and `Suspected` clients retain their resources. This covers the interval after the first `MountLocalDiskSegment` and before the first ReMount; `Offline` clients are still cleaned up. All stale-cleanup call sites use the same source of truth.

4. A permanently unsuccessful offboarding job intentionally keeps scheduled snapshots behind the barrier. This is fail-closed: the job and terminal `Offline` state are process-local and are not serialized, so allowing a snapshot through after a retry or time limit could persist partially offboarded resources and reconstruct them as `Active` after failover. The worker retries with backoff capped at 30 seconds and exposes queue depth, retries, alerts, and error logs for operator intervention. I do not think a bounded barrier is safe unless the snapshot schema first persists the terminal incarnation and residual work, so I kept the current behavior rather than adding an unsafe fallback.

I also formatted all PR-changed C/C++ lines with clang-format 20. The exact CI format command now passes locally. `mooncake_master`, eight relevant test binaries, and all nine benchmark targets in the current configuration compile and link successfully. The three focused regression tests for the two fixes pass; I have not run the full test suite in this update.

Icedcoco:
Thanks for following up and for addressing the multi-segment name handling and the ShouldRetainResources() path. Those two fixes make sense to me. I have a few follow-up questions about the remaining lifecycle and concurrency semantics:

1. After HA promotion, restored memory replicas are created with a dummy allocator and no ClientLivenessRecord, while AllocatedBuffer::isAvailable() treats a missing record as available. The current test also expects restored data to be readable before the actual owner remounts. Is that the intended serving contract? If so, what prevents an old or unreachable client's restored address from being served before remount? I may be missing an invariant here.

2. The PR description says accepted offboarding jobs are drained during shutdown, but ClientOffboardingWorker::Stop() sets running_ to false and drops queued jobs. Since the terminal liveness state and residual job are process-local, could a leadership change discard cleanup that the fail-closed snapshot policy is meant to protect? Should shutdown wait for accepted jobs to converge, or is there another handoff path?

3. Supporting multiple segment IDs with the same name looks correct, but segment_id_by_name_ still stores only one ID. If the currently indexed segment is unmounted first, CommitUnmountSegment() removes the name mapping even though another segment with the same name remains mounted. Could we add a test for the reverse unmount order and either restore the remaining ID or clarify why name-based lookup is allowed to disappear in that case?

4. I also want to check the lock ordering in the task paths. CreateCopyTask() and CreateMoveTask() hold a metadata accessor while acquiring a ServingGuard, whereas CopyEnd() and MoveEnd() acquire the liveness guard before taking the metadata accessor. Could this form an AB-BA cycle under concurrent completion and task creation? Would it be safer to collect the liveness record, release the metadata accessor, acquire the guard, and then reacquire and validate the metadata state?

The latest CI run has the format and build-related checks passing. The remaining SGLang integration failure only reports that the external T-One job failed or stopped, so it would be helpful to rerun it or provide the underlying T-One failure details before treating it as a code regression.

Doube-Suzerain:
Thanks for another careful pass. All four observations were valid. I addressed
the three code issues and corrected the shutdown description rather than adding
a new handoff protocol:

1. I restored Mooncake's existing `invalid_replica_endpoints_` gate alongside
   the resource-liveness check. A memory replica reconstructed with a dummy
   allocator is therefore not returned by Get/BatchGet, selected as a
   Copy/Move/offload source, or evicted before its real owner successfully
   ReMounts. ReMount then removes the existing endpoint gate and binds the
   restored buffers to the actual owner's liveness record. No fake Active
   record is created from the object's writer ID. The standby-remount tests now
   assert `REPLICA_IS_NOT_READY` before ReMount and normal readability after it.

2. You are right that the PR description was inaccurate: `Stop()` does not
   drain queued/backoff jobs, and there is no persisted job handoff. The actual
   shutdown contract is narrower. Master first stops and joins its local
   snapshot producer while the offboarding barrier is still raised, then stops
   the worker; an in-flight attempt leaves its critical section, while queued
   and backoff jobs are discarded with a summary error. Thus this process
   cannot publish a new partial snapshot after dropping the jobs. A future
   leader starts from the already durable snapshot/oplog prefix, so unpersisted
   cleanup intent can be lost and an older valid snapshot can restore those
   resources. That is the explicit limitation of keeping liveness and residual
   work process-local in this PR. Waiting indefinitely during shutdown would
   not provide a durable handoff, so I corrected the PR description instead of
   adding a drain or fallback protocol.

3. Fixed. When the currently indexed Segment ID is removed,
   `ScopedSegmentAccess` now reindexes the name and owner to a remaining
   Segment with that name, preferring an `OK` Segment. The name disappears only
   after the last matching ID is removed. The regression test unmounts in the
   reverse order and verifies name lookup, ownership, and allocation remain
   available through the surviving ID; snapshot restore also preserves both
   same-name registrations.

4. Fixed with the ordering you suggested. `CreateCopyTask()` and
   `CreateMoveTask()` now perform discovery under metadata, release it, acquire
   the source `ServingGuard`, then reacquire and validate the object, source
   owner/replica, and target availability before submitting the task. A
   deterministic regression test holds Segment and Serving locks to verify
   both paths release the metadata shard before waiting for the guard, covering
   the previous AB-BA cycle with CopyEnd/MoveEnd.

The focused standby and Copy/Move lock-order tests pass. The complete
`master_service_ha_test` binary passes 89/90 tests in this build; its only
failure is the existing promotion E2E case that explicitly requires a build
with `STORE_USE_ETCD`. The new push will also rerun the SGLang integration job;
the previous GitHub job exposed only the external T-One failed/stopped status
and no underlying job log, so I could not attribute that run to this change.

catyans:
Thanks @Icedcoco for the detailed review, and @Doube-Suzerain for addressing the restored-replica visibility, same-name segment indexing, and task lock-order issues. I re-checked those paths on the current head and the fixes look directionally sound.

The PR currently conflicts with `main`, and CI has real failures in `MasterServiceTest.UpsertConflictReplicationTask` and `MasterServiceHATest.SegmentLifecycleWritesBatchRecordOpLogs`, along with broad clang-format differences. Could you rebase, fix these failures, and rerun CI? Happy to do a final pass afterward.

Doube-Suzerain:
@catyans Thanks for the final pass. I rebased the branch onto the current main at 9466cde9 and force-pushed dcc9939f. The PR is now conflict-free.

I addressed the two CI failures without relaxing either production invariant:

1. UpsertConflictReplicationTask was a migrated DSL test still invoking CopyStart as the default actor, which has no liveness record. I updated it, plus the sibling migrated Copy/Move scenarios exposed by the full run, to invoke lifecycle RPCs as the mounted source owner. Wrong-client cases now use another mounted, live client, so they still reach and verify ILLEGAL_CLIENT.
2. SegmentLifecycleWritesBatchRecordOpLogs configured a one-entry fenced writer but issued mount, remount, and unmount before waiting for prior batches to become durable. The test now waits for each batch before issuing the next lifecycle operation. Production OpLog admission and failure behavior are unchanged.
3. I applied clang-format 20.1.8 to PR-changed ranges. The exact changed-lines diff check is clean.

While resolving the earlier Store conflicts, I also preserved current main behavior for replica-ID restoration by continuing to use the private restore constructors; liveness affiliation is layered on top without making those constructors public.

Local validation on the final head:
- complete cmake build succeeds, including Store, HA, snapshot, scenario, benchmark, and Python module targets;
- both originally reported tests pass;
- master_service_test, master_service_ha_test, master_service_scenario_test, master_service_replication_scenario_test, snapshot tests, and all PR-related targets pass in the full run;
- tcp_write_visibility_test passes all 44 cases after running the metadata server in the same network namespace.

The remaining local-only failures are unrelated host capability checks: pinned CUDA host allocation is unavailable, hugepage mmap falls back to regular pages for the 2 MB alignment case, and the real RDMA endpoint re-establish test does not converge on this host fabric. I did not change production code for those. The new push has triggered GitHub CI for the authoritative run.

Icedcoco (approved):
LGTM.

I reviewed the current implementation, especially the HA restore, liveness gating, ReMount, and asynchronous offboarding paths. Restored standby memory remains behind the invalid-endpoint gate until the actual owner successfully ReMounts, so stale restored resources are not immediately exposed through serving paths.

I suggest recording the remaining HA limitation for follow-up: offboarding state is process-local, so a leadership change before cleanup converges may restore stale metadata and related quota/accounting state from the last durable snapshot. This does not block merging in this PR and can be addressed in a subsequent change.

The previously discussed snapshot barrier behavior and versioned HTTP metadata cleanup are left out of this review decision and can be handled separately.

ykwd (approved):

Issue #2873, opened 2026-07-13
[RFC]: Disaggregated Diffusion Model Serving with Mooncake


gangula-karthik:
### Changes proposed

# [RFC] Disaggregated Diffusion Model Serving with Mooncake

## Introduction

This RFC proposes first-class support in Mooncake for **disaggregated serving of diffusion models** (image and video generation), decoupling the pipeline into independently scalable stages — **Condition Encode (E)**, **Denoise (D, the DiT/U-Net transformer)**, and **VAE Decode (V)** — connected by the Mooncake Transfer Engine and Mooncake Store, and orchestrated through a **queue-based multi-stage scheduler**.

This directly targets two roadmap items:
- *(Disaggregated Diffusion Deployment)*: decentralized disaggregated deployment for video generation models, decoupling transformer and decoder stages.
- *(Queue-based Orchestration)*: queue-based scheduling for multi-stage diffusion serving requests.

## Motivation

LLM-style Prefill/Decode (PD) disaggregation exists because prefill is compute-bound and decode is memory-bound. Diffusion pipelines have an analogous — but structurally different — resource asymmetry:

| Stage | Workload character | Typical hardware fit |
|---|---|---|
| Condition encoders (T5/CLIP/LLM captioner, image/video encoders) | Small, bursty, compute-light; highly cacheable (prompts repeat) | Small GPUs / MIG slices / CPU |
| Denoise transformer (DiT) | Compute-bound, iterative (20–100 steps), long-running (seconds–minutes for video); scales with sequence/CFG/Ulysses parallelism | Large multi-GPU groups (TP/SP) |
| VAE decode | Memory-bandwidth- and VRAM-capacity-bound (activations at pixel resolution, especially video); embarrassingly parallel across temporal chunks | Fewer, memory-rich GPUs |

Co-locating these stages on one GPU group causes the same problems PD disaggregation solved for LLMs:

1. **VRAM contention**: for video models (e.g., Wan, HunyuanVideo, Qwen-series generators), VAE decode activation memory competes with DiT weights/activations, forcing tiled decode or smaller batch sizes.
2. **Poor utilization**: text encoders and the VAE are idle >90% of the time while the DiT denoises; the DiT is idle during decode. Independent scaling fixes this (e.g., N denoise groups sharing one decode pool).
3. **Interference and head-of-line blocking**: a 60-second video decode stalls the next request's denoise steps; step-latency jitter breaks SLOs for interactive image generation.
[...]

PR #2992, opened 2026-07-19
[Diffusion] Add DiffusionConnector API surface

gangula-karthik:
## Description

Phase 1.1 of #2873: introduces the `DiffusionConnector` API surface — `StagePayload`,
`StageHandle`, `P2PSegmentRef`, the `DiffusionConnector` constructor, `publish()`/`consume()`
signatures, and the `auto`-mode transport selection rule (`_select_transfer_mode`).

This is API shape only, matching the RFC's proposed sketch — `publish()`/`consume()` raise
`NotImplementedError`. No transfer logic (P2P or Store-relay) is implemented here; that's
tracked as separate follow-on PRs (see table below).

Tracking table for RFC #2873 Phase 1:

| Deliverable | PR |
|---|---|
| DiffusionConnector API surface | this PR |
| P2P whole-tensor transfer | TBD |
| Store-relay whole-tensor transfer | TBD |
| LightX2V reference integration | TBD |

## Module

## Type of Change
- [x] New feature

## How Has This Been Tested?
No transfer logic exists yet to test end-to-end. Verified:
- Module imports cleanly (`from mooncake.diffusion import DiffusionConnector, StagePayload`)
- Constructor/dataclass validation rejects invalid `role`/`transfer_mode`/`kind` values
- `_select_transfer_mode` returns the correct mode for each of the known/capacity combinations
- `publish()`/`consume()` raise `NotImplementedError` as expected

## Checklist
- [x] Self-review done
- [x] Formatted with `./scripts/code_format.sh` (no-op — pure Python, ruff/ruff-format handle formatting)
- [x] `pre-commit run --files mooncake-wheel/mooncake/diffusion_connector.py mooncake-wheel/mooncake/diffusion.py` passes (trailing-whitespace, ruff, ruff-format, codespell); `mooncake-code-format` hook fails locally only because this machine has clang-format 11 not 20 — unrelated, C++-only hook
- [ ] Not >500 LOC, no new RFC needed (RFC #2873 already covers this)

## AI Assistance Disclosure
- [x] AI tools were used (Claude Code) — used to draft the dataclasses/connector class
  against the RFC sketch and codebase style conventions; I reviewed every line and can
  defend the design.

### DIFF
--- a/mooncake-wheel/mooncake/diffusion.py
+++ b/mooncake-wheel/mooncake/diffusion.py
@@ -1,0 +1,8 @@
+from mooncake.diffusion_connector import (
+    DiffusionConnector,
+    P2PSegmentRef,
+    StageHandle,
+    StagePayload,
+)
+
+__all__ = ["DiffusionConnector", "P2PSegmentRef", "StageHandle", "StagePayload"]
--- a/mooncake-wheel/mooncake/diffusion_connector.py
+++ b/mooncake-wheel/mooncake/diffusion_connector.py
@@ -1,0 +1,211 @@
+"""Connector for moving tensors between disaggregated diffusion pipeline stages.
+
+Provides producer/consumer primitives for passing stage outputs (text/image
+embeddings, denoising latents, checkpoint tensors) between the encode,
+denoise, and decode stages of a disaggregated diffusion pipeline, mirroring
+the producer/consumer split of the vLLM ``MooncakeConnector``.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Any, Iterator, Literal, Mapping, Optional
+
+VALID_ROLES = ("encode", "denoise", "decode")
+VALID_TRANSFER_MODES = ("p2p", "store", "auto")
+VALID_PAYLOAD_KINDS = ("embedding", "latent", "checkpoint")
+
+ConnectorRole = Literal[VALID_ROLES]
+TransferMode = Literal[VALID_TRANSFER_MODES]
+PayloadKind = Literal[VALID_PAYLOAD_KINDS]
+
+
+@dataclass(frozen=True)
+class StagePayload:
+    """One stage output a producer wants to make available to a consumer.
+
+    Attributes:
+        request_id: End-to-end request identifier; the same id must be used
+            by every stage that publishes or consumes tensors for a request.
+        kind: What the tensor holds (``embedding``, ``latent``, or
+            ``checkpoint``); consumers use it to route the payload to the
+            right stage input.
+        tensor: The producer's real inference output tensor. It is published
+            in place (registered with the transfer engine, no staging copy),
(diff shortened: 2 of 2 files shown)


stmatengss:
Thanks for setting up the initial API surface! The dataclass layouts look consistent with the RFC. 

However, since this PR only introduces the bare interfaces (`publish` and `consume` throwing `NotImplementedError`), it's a bit difficult to fully evaluate how smoothly this integrates into the actual diffusion serving pipeline. 

Could you clarify or provide a quick sketch on a few integration points before we proceed?
1. **Lifecycle & State Management:** How will `DiffusionConnector` interact with the active serving engine loop? Will it be managed as a persistent worker/sidecar, or instantiated dynamically per pipeline request?
2. **Configuration Routing:** The bot flag noted conditional requirements for `store_config` and `te_config`. In the actual `LightX2V` integration, how are these cluster-level configurations expected to be passed down cleanly to the connector without cluttering the upper serving layers?
3. **P2P Coordinate Resolution:** For the upcoming P2P transfer phase, how will individual stages discover the network endpoints (`P2PSegmentRef`) of their producer/consumer peers during runtime? 

It would be great to get a brief alignment on these design assumptions to ensure the current API shape doesn't require immediate breaking changes when the whole-tensor transfer PRs land.

gangula-karthik:
Thanks for the questions @stmatengss — short answers below.

**1. Lifecycle & state management**

Per-process, long-lived — not a sidecar, not per-request. Each stage process (encode/denoise/decode) constructs one `DiffusionConnector(role=...)` at start-up and keeps it for the process's life, since buffer registration with the Transfer Engine has to happen once, before any request arrives.

What *is* per-request is `StagePayload`/`StageHandle` — those get created on each `publish()`/`consume()` call. Same shape as `MooncakeConnectorWorker` in vLLM: built once, reused across steps.

**2. Configuration routing**

`store_config`/`te_config` are plain dicts assembled once at the deployment layer and passed into the constructor — the connector never fetches config itself.

LightX2V today self-configures per-process (JSON file + env vars for TE, a `disagg_config` sub-dict in the model config for topology). Porting it onto `DiffusionConnector` means the launcher builds `te_config`/`store_config` from those same sources and passes them explicitly — replacing `load_from_env()`, not wrapping it. `disagg_config` (bootstrap/rank/room) stays a separate orchestration concern, not folded into these two configs.

**3. P2P coordinate resolution**

No new discovery service — reuses the existing `P2PHANDSHAKE` metadata path (same one the KV-cache connector uses today).

Flow: producer registers its buffers with the Transfer Engine at start-up → `publish()` returns a `StageHandle` with a `P2PSegmentRef` baked in → handle travels to the consumer via the orchestrator's queue message → `consume(handle, dst)` redeems the segment ref over `P2PHANDSHAKE` against the same metadata service `te_config` already points at.

gangula-karthik:
hi @stmatengss, any update on this ?

PR #2994, opened 2026-07-20
[TENT] Align TE compatibility semantics

alogfans:
## Description

This PR tightens the Transfer Engine compatibility shim when running on the TENT runtime (`MC_USE_TENT=1` / `MC_USE_TEV1=1`).

Main changes:
- Align TENT compatibility error handling with classic TE `Status` / `ERR_*` semantics.
- Add fail-soft guards for uninitialized or failed TENT initialization paths.
- Preserve classic TE memory registration semantics in TENT mode:
  - reject zero-length registrations
  - reject overlapping registrations
  - return `ERR_ADDRESS_NOT_REGISTERED` for unknown unregisters
- Ensure TENT transport registration failures during memory registration are surfaced and rolled back.
- Normalize batch/status behavior:
  - translate TENT allocation failure to `INVALID_BATCH_ID`
  - avoid mutating output transfer status on error
- Improve compatibility for `installTransport`, `mp_*`, C API `discoverTopology`, and C API segment handle width checks.
- Update `transfer_engine_validator` to avoid classic-only metadata access in TENT mode.
- Document TENT/TE compatibility-mode differences and migration guidance.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
```

**Test results:**
- [X] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used
- [X] AI tools were used (specify below) CodeX


mooncake-transfer-engine/src/transfer_engine.cpp
```diff
@@ -310,17 +310,217 @@ std::string TransferEngine::showLinks(bool json) const {
...
+    }
+    if (sockfd >= 0) close(sockfd);
+
+    config.set("metadata_type", "p2p");
+    config.set("rpc_server_hostname", host_name);
+    config.set("rpc_server_port", port);
+    config.set("local_segment_name",
+               maybeWrapIpV6(host_name) + ":" + std::to_string(port));
+    return true;
+}
```
staryxchen, line 565: Classic TE's P2P path keeps desc.sockfd open to hold the port until the RPC server binds (transfer_engine_impl.cpp:141). The PR closes the probe socket immediately, then constructs mooncake::tent::TransferEngine(config) which binds to that port. TOCTOU: another process can grab the port between close and TENT's bind. The window is small but real, and the failure mode (TENT constructor throws or returns unavailable) is hard to diagnose. Suggest: keep the socket open and pass it into TENT, or have TENT bind to port 0 and report the chosen port back.

mooncake-transfer-engine/src/transfer_engine.cpp
```diff
@@ -540,30 +830,93 @@ int TransferEngine::registerLocalMemory(void* addr, size_t length,
...
+                                         length)) {
+                LOG(ERROR)
+                    << "Transfer Engine does not support overlapped memory "
+                       "region";
+                return ERR_ADDRESS_OVERLAPPED;
+            }
+        }
+        auto option =
+            makeTentMemoryOptions(location, remote_accessible, update_metadata);
         auto status = impl_tent_->registerLocalMemory(addr, length, option);
```
staryxchen: registerLocalMemory checks overlap under tent_compat_mutex_, releases the lock, calls impl_tent_->registerLocalMemory, then re-acquires the lock to append. Two threads registering overlapping regions concurrently can both pass the check. Acceptable for the common single-threaded init pattern, but are there any scenarios involving multiple-threaded?

mooncake-transfer-engine/src/transfer_engine.cpp
```diff
@@ -581,28 +941,56 @@ int TransferEngine::registerLocalMemoryBatch(
...
         return impl_->unregisterLocalMemoryBatch(addr_list);
     }
 }
 
 BatchID TransferEngine::allocateBatchID(size_t batch_size) {
     if (use_tent_) {
-        return impl_tent_->allocateBatch(batch_size);
+        if (!impl_tent_) return INVALID_BATCH_ID;
+        auto batch_id = impl_tent_->allocateBatch(batch_size);
+        return batch_id == 0 ? INVALID_BATCH_ID : batch_id;
```
staryxchen, line 1021: This assumes TENT never returns 0 as a valid batch ID. Worth a one-line comment in TENT's allocateBatch documenting this contract so a future change to TENT's ID allocation doesn't silently break the shim.

mooncake-transfer-engine/src/transfer_engine.cpp
staryxchen, line 577: Was return 0 (silent success); now return -1 with a warning. Callers that check rc == 0 for cleanup may now see unexpected failures. The migration table in docs/source/design/tent/cpp-api.md doesn't list removeLocalSegment. Suggest adding a row.

mooncake-transfer-engine/src/transfer_engine.cpp
alogfans: Good catch. This is a real TOCTOU issue in the legacy P2P compatibility path. A robust fix should be handled separately by changing TENT’s RPC startup semantics, preferably true port-0 binding with the selected port reported before publishing the local segment identity.

staryxchen:
@alogfans Code is good to me, but there are a lots of failed CI job.

alogfans:
> @alogfans Code is good to me, but there are a lots of failed CI job.

Identified and fixed.

Primary33 (approved):

PR #2998, opened 2026-07-20
[Store] support tensor APIs with dummy client

zxpdemonio:
## Description

This PR enables tensor APIs for dummy clients when the data path can use the dummy client's local shared-memory buffer and a real client mapped to the same SHM-backed buffer.

Before this change, many tensor APIs rejected dummy clients even when the dummy client already had a local buffer allocator. This blocked the deployment mode where worker processes use dummy clients without Store segments and forward data through a real client process.

Main changes:

- Initialize `client_buffer_allocator_` for dummy clients from the dummy local SHM buffer.
- Allow dummy clients to use tensor write APIs through `put_from` / `batch_put_from` style paths.
- Use accelerator-aware `CopyToHost` in the parallel tensor write helper so CUDA-addressed tensor sources are copied to host correctly.
- Enable dummy-client coverage for tensor `put`, `upsert`, `pub`, `_from`, `_into`, TP, and default-parallelism wrapper APIs where the underlying path is supported.
- Keep full materialized parallel tensor read conservative; dummy clients should use `_into` for the SHM-backed fast path.
- Add dummy+real tensor API tests.

## Module

## Type of Change

- [x] New feature
- [x] Bug fix
- [x] Refactor

## How Has This Been Tested?

**Test commands:**

```bash
# Format / syntax checks
clang-format-20 -style=file --dry-run -Werror \
  mooncake-integration/store/store_py.cpp \
  mooncake-integration/store/store_py_parallel_write.h \
  mooncake-store/src/dummy_client.cpp

python3 -m py_compile mooncake-wheel/tests/test_dummy_client.py
git diff --check

# Build
cmake --build build-dummy-tensor-nocuda \
  --target store mooncake_master mooncake_client -j16

# Dummy + real client tensor tests
cd mooncake-wheel/tests
python3 -m unittest -v test_dummy_client.py

# Multi dummy client smoke test
python3 test_multi_dummy_clients.py --client-id client1
python3 test_multi_dummy_clients.py --client-id client2
```

**Test results:**

- [x] Unit tests pass
- [x] Integration tests pass
- [x] Manual testing done

Manual testing covered:

- dummy client setup with local SHM buffer allocator
- real client mapped to the same dummy SHM buffer
- `put_tensor`
- `get_tensor`
- `get_tensor_into`
- `batch_get_tensor_into` invalid-argument handling
- `put_tensor_from`
- `batch_put_tensor`
- `batch_get_tensor`
- `upsert_tensor`
- `pub_tensor`
- `put_tensor_with_tp`
- `get_tensor_with_tp`
- `batch_put_tensor_with_tp`
- `put_tensor_with_parallelism`
- `batch_put_tensor_with_parallelism`
- `put_tensor_with_parallelism_from`
- `upsert_tensor_with_parallelism`
- `batch_upsert_tensor_with_parallelism_from`

## Checklist

## AI Assistance Disclosure

- [x] AI tools were used

Codex was used to help implement, review, test, and prepare this PR. I reviewed the changes and is responsible for the final patch.

### DIFF
--- a/docs/source/api-reference/python/mooncake-store.md
+++ b/docs/source/api-reference/python/mooncake-store.md
@@ -1142,9 +1142,9 @@ parallelism write wrappers.
 
 The real client owns the SHM buffer allocator. This keeps tensor writes and
 regular object writes from allocating overlapping offsets when they run
-concurrently through the same dummy client.
-Writes that stage through the dummy client's local SHM buffer are serialized
-per dummy client to keep the staged data alive until each write finishes.
+concurrently through the same dummy client. Writes staged through the dummy
+client's local SHM buffer keep their allocation alive until completion through
+the real-side active buffer handle and dummy-side RAII release path.
 
 Full materialized reconstruction reads for writer-sharded or reconstructed
 parallel tensors are still conservative for dummy clients. Use the corresponding
--- a/docs/source/api-reference/python/mooncake-store.md
+++ b/docs/source/api-reference/python/mooncake-store.md
@@ -1134,6 +1134,22 @@ def setup_dummy(self, mem_pool_size: int, local_buffer_size: int, server_address
 store.setup_dummy(1024*1024*256, 1024*1024*64, "localhost:8080")
 ```
 
+Dummy clients do not own Store segments. They use a local shared-memory buffer
+that is mapped by a real client process at `server_address`. Tensor APIs that
+stage through this SHM buffer are supported, including tensor put/get,
+`*_tensor_from`, `*_tensor_into`, tensor upsert/pub, TP wrappers, and unified
+parallelism write wrappers.
+
+The real client owns the SHM buffer allocator. This keeps tensor writes and
+regular object writes from allocating overlapping offsets when they run
+concurrently through the same dummy client.
+Writes that stage through the dummy client's local SHM buffer are serialized
+per dummy client to keep the staged data alive until each write finishes.
+
+Full materialized reconstruction reads for writer-sharded or reconstructed
+parallel tensors are still conservative for dummy clients. Use the corresponding
+`*_into` APIs, or read stored shards directly, when using dummy clients.
+
 ---
 
 #### put()
@@ -1611,7 +1627,8 @@ def batch_get_buffer(self, keys: List[str]) -> List[BufferHandle]
 **Returns:**
 - `List[BufferHandle]`: List of buffer objects, with None for keys not found
 
-**Note:** This function is not supported for dummy client.
+**Note:** This function is supported for dummy clients through the real
+client-owned shared-memory staging buffer.
 
 **Example:**
 ```python
@@ -2045,7 +2062,8 @@ def put_from_with_metadata(self, key: str, buffer_ptr: int, metadata_buffer_ptr:
 **Returns:**
 - `int`: Status code (0 = success, non-zero = error code)
 
-**Note:** This function is not supported for dummy client.
+**Note:** This function is supported for dummy clients through the real
+client-owned shared-memory staging buffer.
 
 **Example:**
 ```python
@@ -2445,7 +2463,8 @@ def upsert_tensor_from(self, key: str, buffer_ptr: int, size: int) -> int
 **Returns:**
 - `int`: Status code (0 = success, non-zero = error code)
 
-**Note:** This function is not supported for dummy client.
+**Note:** This function is supported for dummy clients through the real
+client-owned shared-memory staging buffer.
 
 #### batch_upsert_tensor_from()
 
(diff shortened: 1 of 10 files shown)


mooncake-store/src/dummy_client.cpp
```diff
@@ -497,6 +497,23 @@ int DummyClient::setup_dummy(size_t mem_pool_size, size_t local_buffer_size,
         local_buffer_shm->registered = true;
         local_buffer_shm->is_local = true;
 
+        try {
+            client_buffer_allocator_ =
```
yokinoshitayoki: This creates a `ClientBufferAllocator` in the dummy process over the entire local SHM range. However, when the real client maps this SHM, it already creates its own `context.client_buffer_allocator` over the same underlying range in `RealClient::map_shm_internal()`.

The two allocators keep independent allocation metadata and use only process-local synchronization. As a result, a tensor write using the dummy-side allocator can receive the same offset as a concurrent regular `put`, `put_parts`, or buffer read using the real-side allocator. The two operations then overwrite each other while both still report success.

zxpdemonio: Thanks for catching this. I agree the previous dummy-side allocator over the same local SHM range was unsafe, since the real client also owns an allocator for that mapped range and the two allocators could race/collide.

I updated the design to keep real client as the single allocation owner. The dummy client no longer creates its own ClientBufferAllocator. Tensor staging now asks the real client to allocate from the real-side ShmContext::client_buffer_allocator, and the real client keeps the BufferHandle alive in active_handles until the dummy-side RAII handle releases it via release_buffer_dummy.

I also kept the SHM split/partition approach out of this PR, as suggested.

mooncake-wheel/tests/test_dummy_client.py
```diff
@@ -322,6 +328,163 @@ def test_batch_put_from_operations(self):
...
+                parallel_key,
+                *parallel_batch_keys,
+                parallel_from_key,
+                parallel_upsert_key,
+                parallel_upsert_from_key,
+            ]
+        )
+        cleanup_keys.extend(f"{tp_key}_tp_{rank}" for rank in range(2))
+        cleanup_keys.extend(f"{batch_tp_key}_tp_{rank}" for rank in range(2))
+        tensor = torch.arange(12, dtype=torch.float32).reshape(3, 4)
```
yokinoshitayoki, line 362: Please add a mixed-concurrency regression test that runs regular `put` and `put_tensor` calls against the same dummy client's SHM. This is the workload that exposed the original allocator collision.

yokinoshitayoki (commented):
Maybe we can update the Python API documentation, which still says these APIs are unsupported for dummy clients.

zxpdemonio:
> Maybe we can update the Python API documentation, which still says these APIs are unsupported for dummy clients.

Doc is updated.

mooncake-store/src/dummy_client.cpp
```diff
@@ -733,6 +761,7 @@ uint64_t DummyClient::alloc_from_mem_pool(size_t size) {
 
 int DummyClient::put(const std::string& key, std::span<const char> value,
                      const ReplicateConfig& config) {
+    std::lock_guard<std::mutex> lock(local_buffer_write_mutex_);
```
yokinoshitayoki: The real-owned `ClientBufferAllocator` is already thread-safe, and `active_handles` plus the dummy-side RAII handle keeps each staged allocation alive until `BatchPut` returns. Holding this mutex across every blocking put/upsert RPC therefore limits each dummy client to a single in-flight write, regardless of the real client's configured thread count, causing a potentially severe throughput regression. Please remove this broad serialization and keep the mixed-concurrency regression test. If there is some narrower shared state that still requires protection, lock only that state and document the specific race.

docs/source/api-reference/python/mooncake-store.md
```diff
@@ -1134,6 +1134,22 @@ def setup_dummy(self, mem_pool_size: int, local_buffer_size: int, server_address
...
+Dummy clients do not own Store segments. They use a local shared-memory buffer
+that is mapped by a real client process at `server_address`. Tensor APIs that
+stage through this SHM buffer are supported, including tensor put/get,
+`*_tensor_from`, `*_tensor_into`, tensor upsert/pub, TP wrappers, and unified
+parallelism write wrappers.
+
+The real client owns the SHM buffer allocator. This keeps tensor writes and
+regular object writes from allocating overlapping offsets when they run
+concurrently through the same dummy client.
+Writes that stage through the dummy client's local SHM buffer are serialized
```
yokinoshitayoki: The real-side active handle plus the dummy-side RAII release keeps each staged allocation alive until the write finishes. Please remove or update these lines so the documentation matches the implementation.

zxpdemonio:
Updated the dummy-client documentation in c47e1b8b to match the current implementation: staged dummy SHM writes are kept alive by the real-side active buffer handle and released by the dummy-side RAII path. The old serialization-based wording has been removed.

yokinoshitayoki (approved):
LGTM.

stmatengss (approved):

PR #3000, opened 2026-07-20
Add scatter transfer batching to transfer engine

zxpdemonio:
## Description

The original one-shot stitching capability in `get_into_ranges` belongs in Transfer Engine, where it can serve as a general batched scatter-transfer primitive. The same primitive enables one-shot assembly for weight transformation and KV cache transformation in heterogeneous TP/PP/EP scenarios.

This PR:
- Adds `TransferEngine::transferScatter` as a standalone synchronous API available to both the legacy and TENT backends.
- Converts fragmented ranges into one Transfer Engine batch, with independent local and remote offsets for every fragment.
- Uses caller-provided registered CPU or GPU addresses directly. The production scatter path has no host staging or extra `memcpy`.
- Moves address validation, offset arithmetic, segment lookup, batch submission, completion polling, and per-fragment completion into Transfer Engine.
- Keeps Store responsible for Store-specific work only: metadata lookup, replica selection, lease/error translation, and local-disk replica reads.
- Keeps `Client::TransferScatter` and `TransferSubmitter::transferScatter` as thin forwarding layers, so Transfer Task users can invoke the capability directly.
- Preserves empty range groups in Python results and adds focused fragmented GPU-destination coverage to the existing `transfer_task_test`.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
./scripts/code_format.sh -b fc04cb1d --check
git diff --check

cmake --build build --target transfer_task_test -j16
./build/mooncake-store/tests/transfer_task_test

cmake --build build-cuda --target transfer_task_test -j16
CUDA_VISIBLE_DEVICES=0 ./build-cuda/mooncake-store/tests/transfer_task_test

cmake --build build-tent-cpu --target file_storage_promotion_test -j16
cmake --build build-tent-cuda --target file_storage_promotion_test -j16

python3.12 -m unittest \
  test_distributed_object_store.TestDistributedObjectStoreSingleStore.test_get_into_ranges_operations
```

**Test results:**
- [x] Unit tests pass

`transfer_task_test` passed 6/6 cases in the standard build and 7/7 cases in the CUDA build. The CUDA-only case scatters three fragmented ranges directly into a CUDA allocation through `TransferSubmitter`; its device-to-host copy is used only after the transfer to verify the bytes.

The TENT CUDA-on and CUDA-off `file_storage_promotion_test` targets both built and linked successfully. The focused Python wheel `get_into_ranges` regression test passed, including preservation of a known empty range group.

Manual validation passed CPU-to-CPU, CPU-to-GPU, GPU-to-CPU, and GPU-to-GPU combinations with 6,528 fragmented slices. A single call mixing CPU/GPU sources and CPU/GPU destinations also passed with 1,248 fragmented slices.

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

AI tools were used to help implement, review, and validate the change. The human submitter reviewed the code and test results.


mooncake-transfer-engine/include/transfer_engine_impl.h
```diff
@@ -138,6 +139,123 @@ class TransferEngineImpl {
...
+                }
+
+                TransferRequest request;
+                request.opcode = TransferRequest::READ;
+                request.source = local_address;
+                request.target_id = seg_it->second;
+                request.target_offset =
+                    range.remote_base_offset + remote_offset;
+                request.length = length;
+                requests.emplace_back(request);
```
alogfans: The PR description mentions GPU destination support via host staging and accelerator copy. Could you point out where that staging/copy path is handled, or add a focused test for GPU destination scatter reads?

zxpdemonio: Thanks for pointing this out. The earlier host-staging wording was stale and has been removed from the PR description.

After the final refactor, the implementation is `TransferEngine::transferScatter` in `mooncake-transfer-engine/src/transfer_engine.cpp`. It is defined outside the legacy/TENT conditional so both backends use the same scatter orchestration. For a READ, each `TransferRequest::source` is the caller-provided local destination address plus the fragment offset; that registered address may be CPU or GPU memory. The requests are submitted as one batch, with no host staging or extra production `memcpy`.

The focused `TransferScatterWritesGpuDestinationDirectly` test scatters three fragments directly into a CUDA allocation through `TransferSubmitter`. The device-to-host copy occurs only after completion to verify the result. The CUDA suite passes 7/7 cases. Manual validation also passes CPU-to-CPU, CPU-to-GPU, GPU-to-CPU, and GPU-to-GPU with 6,528 fragments, plus a single mixed CPU/GPU batch with 1,248 fragments.

mooncake-transfer-engine/src/transfer_engine.cpp
```diff
@@ -845,3 +850,142 @@ std::string TransferEngine::showLinks(bool json) const {
...
 }  // namespace mooncake
 #endif
+
+namespace mooncake {
+
+Status TransferEngine::transferScatter(
+    const std::vector<ScatterTransferRange>& ranges) {
+    std::vector<TransferRequest> requests;
+    std::vector<std::pair<size_t, size_t>> request_fragments;
+    std::unordered_map<std::string, SegmentHandle> segment_cache;
```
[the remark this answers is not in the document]
zxpdemonio: Fixed in eabc57cd. transferScatter now owns opened handles in a scoped segment cache and closes every successfully opened handle on all exits. The API is synchronous, so cleanup happens after completion; the legacy backend keeps its existing close behavior, while TENT releases the remote segment reference.

mooncake-store/src/real_client.cpp
```diff
@@ -3480,143 +3409,123 @@ RealClient::get_into_ranges_internal(
...
-                        all_src_offsets[i][j][k] == 0) {
-                        VLOG(1)
-                            << "Object not found for key: " << all_keys[i][j];
-                    }
-                    prepared.results[i][j][k] =
-                        tl::unexpected(metadata_result.error());
+            if (dst_offsets.size() != src_offsets.size() ||
+                dst_offsets.size() != sizes.size()) {
+                continue;
+            }
```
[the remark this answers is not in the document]
zxpdemonio: Fixed in eabc57cd. The later non-memory-only consistency check was removed after moving validation to the common path.

mooncake-transfer-engine/src/transfer_engine.cpp
```diff
@@ -845,3 +850,142 @@ std::string TransferEngine::showLinks(bool json) const {
...
+                complete(range_index, fragment_index, Status::OK());
+                continue;
+            }
+
+            auto [seg_it, inserted] = segment_cache.emplace(
+                range.remote_segment,
+                static_cast<SegmentHandle>(ERR_INVALID_ARGUMENT));
+            if (inserted) seg_it->second = openSegment(range.remote_segment);
+            if (seg_it->second ==
+                static_cast<SegmentHandle>(ERR_INVALID_ARGUMENT)) {
```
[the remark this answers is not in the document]
zxpdemonio: Thanks. ERR_INVALID_ARGUMENT is defined as -1, so the previous check already matched the TENT failure sentinel. I normalized the TENT wrapper to return static_cast<SegmentHandle>(ERR_INVALID_ARGUMENT), making the shared backend contract explicit with one sentinel to check. The TENT transfer_engine target builds successfully with this change.

mooncake-store/src/real_client.cpp
```diff
@@ -3480,143 +3409,123 @@ RealClient::get_into_ranges_internal(
...
-                        .emplace(
-                            all_keys[i][j],
-                            build_ranged_read_metadata_from_query_result(
-                                all_keys[i][j], client_->Query(all_keys[i][j])))
-                        .first;
+        for (size_t j = 0; j < keys.size(); ++j) {
+            const auto &dst_offsets = dst_groups[j];
+            const auto &src_offsets = src_groups[j];
+            const auto &sizes = size_groups[j];
+            auto &range_results = results[i][j];
```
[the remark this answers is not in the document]
zxpdemonio: Fixed in eabc57cd. Fragment-count consistency is now validated before metadata lookup and replica dispatch, so a memory replica cannot install a completion callback over mismatched spans. Invalid entries retain their preinitialized INVALID_PARAMS result. The existing wheel test now covers mismatched destination, source, and size counts.

zxpdemonio:
@alogfans  Hi, the refactor has now been completed, and this PR is ready for another review.

The latest version:

- Moves the scatter transfer capability, including address and offset handling, into `TransferEngine::transferScatter`.
- Keeps the Store layer focused on Store-specific validation, metadata, and replica selection.
- Performs fragmented CPU/GPU transfers directly in a single batch without host staging or additional production `memcpy`.
- Addresses the latest review feedback, including TENT segment-handle cleanup and early fragment-count validation.
- Keeps the total change below 500 added lines.

Validation completed:

- `transfer_task_test`: 7/7 passed with CUDA enabled.
- `get_into_ranges_operations`: passed.
- TENT `transfer_engine` target: built successfully.
- Code format and Git hooks: passed.

All existing review threads have been addressed and resolved. Could you please take another look when convenient? Thanks!

mooncake-transfer-engine/src/transfer_engine.cpp
```diff
@@ -868,3 +852,153 @@ std::string TransferEngine::showLinks(bool json) const {
...
+                terminated[i] = true;
+                --remaining;
+                has_failure = true;
+            }
+        }
+        if (remaining == 0) break;
+        if (std::chrono::steady_clock::now() - start > timeout) {
+            return abort(Status::Socket("scatter transfer timed out"));
+        }
+        std::this_thread::yield();
```
alogfans: The synchronous polling loop currently uses `std::this_thread::yield()` until completion or a 60s timeout. Could we use a small sleep/backoff, or clarify that this follows the existing TransferEngine polling behavior?

zxpdemonio: Addressed in 59b032ec. The synchronous scatter loop now sleeps for 10 microseconds after a full status scan when fragments are still pending. This avoids yield-based tight polling while keeping completion latency low. Verified with the TENT GPU scatter test (1/1) and the legacy transfer_task_test suite (7/7).

Bo-Vincent:
The layering looks sound: scatter execution is pushed down into the Transfer Engine, while Store retains ownership of metadata and lease semantics.
This can directly serve as the native lowering foundation for heterogeneous weight transfer, executing fragmented TransferRegions produced by the planner as batched scatter transfers. I recommend merging it.

mooncake-transfer-engine/src/transfer_engine.cpp
```diff
@@ -868,3 +875,154 @@ std::string TransferEngine::showLinks(bool json) const {
...
+                .source = static_cast<char*>(range.local_buffer) + local_offset,
+                .target_id = seg_it->second,
+                .target_offset = range.remote_base_offset + remote_offset,
+                .length = length,
+            });
+            request_fragments.emplace_back(range_index, fragment_index);
+        }
+    }
+
+    if (requests.empty()) return Status::OK();
```
alogfans: Validation and `openSegment()` failures are reported only through optional fragment callbacks. If every fragment is rejected, `requests` is empty and this function returns `OK`, so callers without callbacks observe a successful operation. Please track pre-submission failures and return a non-OK aggregate status when any fragment fails.

mooncake-transfer-engine/src/transfer_engine.cpp
```diff
@@ -868,3 +875,154 @@ std::string TransferEngine::showLinks(bool json) const {
...
+                         Status::Socket("scatter transfer fragment failed"));
+                terminated[i] = true;
+                --remaining;
+                has_failure = true;
+            }
+        }
+        if (remaining == 0) break;
+        if (std::chrono::steady_clock::now() - start > timeout) {
+            return abort(Status::Socket("scatter transfer timed out"));
+        }
```
alogfans: On timeout or status-query failure, the batch may still contain in-flight tasks. `freeBatchID()` can therefore return `BatchBusy`, but its result is ignored, allowing this synchronous API to return while transfers are still running. Please cancel and drain pending tasks before closing segment handles and freeing the batch, and propagate cleanup failures.

mooncake-transfer-engine/src/transfer_engine.cpp
zxpdemonio: Fixed in 258f10d6. transferScatter now records validation and openSegment failures in the aggregate status independently of the optional fragment callback, so an all-rejected request returns non-OK. I also added a regression assertion for an invalid scatter range with no callback.

mooncake-transfer-engine/src/transfer_engine.cpp
zxpdemonio: Fixed in 258f10d6. Once legacy tasks may have been submitted, transferScatter no longer returns on submit, status-query, or timeout errors until every task reaches a physical terminal state; BatchBusy is retried, and segment handles are closed only after batch cleanup. TENT uses best-effort cancellation followed by terminal-state draining, while an atomic TENT submit rejection frees the unpublished empty batch directly. Callback exceptions and cleanup failures are also aggregated into the returned status.

mooncake-transfer-engine/src/transfer_engine.cpp
```diff
@@ -881,3 +888,221 @@ std::string TransferEngine::showLinks(bool json) const {
...
+                complete(range_index, fragment_index,
+                         abort_requested ? abort_status : Status::OK());
+            } else if (status.s == TransferStatusEnum::WAITING ||
+                       status.s == TransferStatusEnum::PENDING ||
+                       (status.s == TransferStatusEnum::TIMEOUT &&
+                        !use_tent_)) {
+                if (status.s == TransferStatusEnum::TIMEOUT) {
+                    request_abort(Status::Socket("scatter transfer timed out"));
+                }
+                continue;
```
staryxchen, line 1196: On timeout or a stuck fragment, this synchronous API can hang forever.

Two related issues:

1. The earlier 60s wall-clock deadline was removed in the drain/cancel cleanup. The loop now only sleeps 10µs and never exits on its own.
2. For the legacy backend, `TransferStatusEnum::TIMEOUT` is treated as non-terminal (`TIMEOUT && !use_tent_` → `continue`). In `MultiTransport::getTransferStatus`, a slice timeout returns `TIMEOUT` without setting `is_finished`. Legacy has no cancel path either, so `remaining` never decreases and the later `freeBatchID` `BatchBusy` retry also cannot make progress.

TENT already treats `TIMEOUT` as terminal; please make legacy do the same, and restore an overall deadline (or equivalent) so `getTransferStatus` errors / never-terminating tasks cannot block the caller indefinitely.

mooncake-transfer-engine/src/transfer_engine.cpp
```diff
@@ -881,3 +888,221 @@ std::string TransferEngine::showLinks(bool json) const {
...
+            if (!result.ok()) {
+                request_abort(result);
+                continue;
+            }
+
+            const auto [range_index, fragment_index] = request_fragments[i];
+            if (status.s == TransferStatusEnum::COMPLETED) {
+                complete(range_index, fragment_index,
+                         abort_requested ? abort_status : Status::OK());
+            } else if (status.s == TransferStatusEnum::WAITING ||
```
staryxchen, line 1194: When `abort_requested` is set, a fragment that already reached `COMPLETED` is reported via `complete(..., abort_status)` instead of `Status::OK()`.

That collapses partial success into a blanket failure: fragments that finished correctly before (or while) another fragment timed out / failed are marked failed in the per-fragment callback, even though their bytes were transferred.

Please keep per-fragment status faithful to the physical terminal state (`COMPLETED` → OK, failure/cancel/timeout → error), and let `aggregate_status` / the function return value carry the abort. Callers that only care about the batch outcome still see the non-OK aggregate.

mooncake-transfer-engine/src/transfer_engine.cpp
```diff
@@ -881,3 +888,221 @@ std::string TransferEngine::showLinks(bool json) const {
...
+#endif
+    };
+    Status result = submitTransfer(batch_id, requests);
+    if (!result.ok()) {
+        // TENT submit failures occur before tasks are published.
+        if (use_tent_) {
+            fail_pending(result);
+            remember(freeBatchID(batch_id));
+            return segment_cache.close(aggregate_status);
+        }
```
staryxchen: The comment says TENT submit failures occur before tasks are published, so this path frees the batch immediately. That does not match `TransferEngineImpl::submitTransfer`: it almost always returns `OK()`, and transport-level failures are recorded by setting `task.type = UNSPEC` (later surfaced as `FAILED` via `getTransferStatus`), not by failing the submit call.

So this early return is effectively dead for the common failure modes, and the comment is misleading for future maintenance. Please either:

- route TENT through the same drain/cancel path as legacy after any non-OK submit (and drop the “unpublished” claim), or
- document the actual rare cases where TENT submit returns non-OK before tasks are visible, if that contract is intentional and guaranteed.

mooncake-transfer-engine/src/transfer_engine.cpp
zxpdemonio: Thanks for pointing this out.

For this PR, I fixed the immediate root cause of the hang in the current synchronous scatter API:

- Legacy `TIMEOUT` is now treated as a terminal task state in `MultiTransport::getTransferStatus()`.
- A timed-out task is marked finished and the batch records failure, so `freeBatchID()` can make progress instead of staying `BatchBusy`.
- `transferScatter()` now treats `TIMEOUT` as a terminal per-fragment failure instead of continuing to poll it.
- Per-fragment completion is kept faithful to the physical terminal state, so completed fragments are still reported as `OK` even if another fragment fails.

I considered a larger API-level redesign that splits scatter into two phases, for example `submitScatter()` plus an explicit wait/poll policy. That would make the lifecycle model cleaner and avoid putting submission, polling, cancellation, and cleanup all inside one synchronous helper.

I did not include that redesign in this PR because it would expand the public Transfer Engine API and add a new operation/lifecycle object. That is a larger semantic change than the bug fix here, and would likely add another ~100-150 lines plus tests. I would prefer to discuss that direction first and, if maintainers agree, follow up with a separate PR focused on the async/wait-policy API shape.

So this PR keeps the existing synchronous API contract, but fixes the timeout terminal-state handling and the per-fragment result semantics.

staryxchen: Thanks for the fix — treating legacy `TIMEOUT` as a terminal state (and keeping per-fragment `COMPLETED` as OK) addresses the hang root cause for the current synchronous API.

Agree to defer the wall-clock deadline / `submitScatter`+wait-policy redesign to a follow-up PR. Resolving this thread.

mooncake-transfer-engine/src/transfer_engine.cpp
```diff
@@ -881,3 +888,198 @@ std::string TransferEngine::showLinks(bool json) const {
...
+                }
+            }
+        }
+#endif
+    };
+    Status result = submitTransfer(batch_id, requests);
+    if (!result.ok()) {
+        fail_pending(result);
+        remember(freeBatchID(batch_id));
+        return segment_cache.close(aggregate_status);
```
staryxchen: After a non-OK `submitTransfer()`, this path calls `fail_pending()` and a single `freeBatchID()`, then returns. That regresses the earlier drain behavior from the failure-cleanup change.

`selectTransport()` is validated up front, but `submitTransferTask()` can still fail after some transports have already accepted work. In that case `freeBatchID()` may return `BatchBusy`. The previous logic fell through into the poll/drain loop; now the busy status is only `remember()`’d once, segments are closed, and in-flight tasks may still be running.

Please restore drain-on-`BatchBusy` after a submit failure (or mark unfinished tasks terminal before freeing), so segment handles are not closed while the batch is still busy.

mooncake-transfer-engine/src/multi_transport.cpp
```diff
@@ -238,11 +246,17 @@ Status MultiTransport::getTransferStatus(BatchID batch_id, size_t task_id,
...
+        // terminal state for the legacy batch lifecycle; otherwise freeBatchID
+        // can stay BatchBusy forever after a timed-out slice.
         if (status.s == Transport::TransferStatusEnum::WAITING &&
             checkSliceTimeout(task)) {
             status.s = Transport::TransferStatusEnum::TIMEOUT;
         }
+        if (status.s == Transport::TransferStatusEnum::TIMEOUT) {
+            task.is_finished = true;
+            batch_desc.has_failure.store(true, std::memory_order_release);
+        }
```
staryxchen: Marking `TIMEOUT` as terminal (`is_finished = true`) correctly unblocks `freeBatchID()` and fixes the scatter hang. One safety concern remains: the transport is not told to abandon in-flight slices, so a worker may still complete and touch the task/slice after `transferScatter` has freed the batch and closed segment handles.

If the intended contract is “slice timeout means abandon and tear down,” please document that explicitly. Otherwise, consider canceling or draining transport-side work before `freeBatchID()` (this can land with the follow-up wait-policy work if preferred).

mooncake-transfer-engine/src/transfer_engine.cpp
zxpdemonio: Fixed in `1daaa70b`. Scatter submission and waiting are now separated through `ScatterTransferOperation`. After a non-OK legacy `submitTransfer()`, the operation attempts `freeBatchID()`; if the batch is still busy, it retains the batch, requests, callbacks, backend, and segment handles and continues polling/draining instead of calling the callbacks and returning early. Resources are released only after the physical batch can be freed. For TENT, the implementation also distinguishes pre-publication failure from a partially published batch and cancels/drains the latter. The operation destructor waits for completion as a final lifetime guard.

mooncake-transfer-engine/src/multi_transport.cpp
zxpdemonio: Fixed in `1daaa70b`. `MultiTransport` no longer marks a timed-out slice as `is_finished` merely to unblock `freeBatchID()`. A wait timeout is now only a wait-policy result and never means that the physical transfer has been abandoned. The scatter operation keeps polling/draining the legacy transport until it reports a real terminal state; for TENT it requests cancellation where supported and still drains before releasing resources. Batch state, backend references, callbacks, buffers, and segment handles remain owned by `ScatterTransferOperation` until physical completion, so timeout cannot cause an early free or close. This was validated with Legacy/TENT CPU and CUDA builds/tests, including releasing the outer `TransferEngine` before waiting on the operation.

staryxchen (approved):
Good with the TE-side changes in this PR

stmatengss (approved):

PR #3003, opened 2026-07-20
[Store] Add Optional Object Checksum Diagnostics

yokinoshitayoki:
## Description

When corrupted KV cache data or garbled model output is observed, it can be difficult to determine whether the data was modified during the Mooncake Store write, storage, or read path. Mooncake Store previously had no end-to-end object integrity check to help diagnose this type of issue. Related RFC https://github.com/kvcache-ai/Mooncake/issues/3140.

This PR adds optional object-level CRC-64 checksum diagnostics to Mooncake Store, enabled by setting `MOONCAKE_STORE_CHECKSUM=1` on client processes. The client calculates the checksum from the source data before `put`/`upsert`, stores it in Master metadata, and verifies the logical `object_size` bytes returned by `get`. A mismatch returns `CHECKSUM_MISMATCH`, preventing the caller from treating corrupted data as valid.

The feature supports GPU buffers through chunked D2H staging and persists checksum metadata in Master snapshots. Objects without checksum metadata remain readable and skip verification. Range reads are intentionally excluded.

This feature is intended for corruption diagnosis rather than regular production use because it adds full-object scans, GPU staging overhead, and disables the local hot cache while enabled.

Checksum metadata continuity across an HA standby failover is not guaranteed in this version. Standby metadata and operation-log replication do not currently propagate the checksum field and will be addressed separately. Objects without checksum metadata are still readable but are not verified.

## Module

- [ ] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [x] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
  client_integration_test dummy_client_get_buffer_test \
  master_service_test snapshot_child_process_test \
  catalog_backed_snapshot_provider_test -j4

ctest --test-dir build \
  -R '^(object_checksum_client_test|dummy_client_object_checksum_test)$' \
  --output-on-failure

  --gtest_filter='MasterServiceTest.ObjectChecksumIsStoredAndClearedByUpsert'

./build/mooncake-store/tests/snapshot_child_process_test \
  --gtest_filter='SnapshotChildProcessTest.RestorePreservesObjectChecksum'

GTEST_FILTER='*LoadLatestSnapshotIgnoresObjectChecksum*' \
  ./build/mooncake-store/tests/catalog_backed_snapshot_provider_test
```

The D2H checksum path was tested with a NVIDIA A100 GPU.

**D2H test result:**
- Buffer: 8 MiB + 17 bytes CUDA device memory
- CRC matched before corruption
- Modifying the last byte correctly returned `CHECKSUM_MISMATCH`

**Test results:**
- [x] Unit tests pass
- [x] Integration tests pass
- [x] Manual testing done on an NVIDIA A100 GPU

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex assisted with implementation, code review, documentation, and test development.


mooncake-store/src/client_service.cpp
```diff
@@ -251,6 +285,25 @@ FinalizeDecision DetermineFinalizeDecision(
 
 }  // namespace
 
+bool StoreChecksumEnabled() {
+    const char* value = std::getenv("MOONCAKE_STORE_CHECKSUM");
```
Aionw: Move env get operation into `environ.h` file. We prefer manage all of env in one place, please do not spread it everywhere.

Personally, I prefer STORE_CHECKSUM isn't a dynamically changed setting, so maybe we only need to read it once process started.

mooncake-store/src/client_service.cpp
```diff
@@ -40,11 +41,44 @@
...
+                                            : crc << 1;
+        }
+        table[i] = crc;
+    }
+    return table;
+}
+
+constexpr auto kCrc64EcmaTable = MakeCrc64EcmaTable();
+
+class ScopedStoreChecksumBuffer {
```
Aionw: All CRC64 checksum-related code—including type, class, and function declarations and implementations—should not reside in `client_service.cpp`. It would be better to move it into the common module and maintain the implementation there.

mooncake-store/include/store_checksum.h
```diff
@@ -0,0 +1,21 @@
+#pragma once
+
+#include <cstddef>
+#include <cstdint>
+
+namespace mooncake {
+
+bool StoreChecksumEnabled();
+
+class StoreChecksum {
```
Aionw: I think `Checksum` or `CrcChecksum` would be a clearer name here. Store has both verb and noun meanings, which makes the current name somewhat ambiguous.

mooncake-store/include/master_client.h
```diff
@@ -224,7 +224,8 @@ class MasterClient {
      * @return tl::expected<void, ErrorCode> indicating success/failure
      */
     [[nodiscard]] tl::expected<void, ErrorCode> PutEnd(
-        const std::string& key, ReplicaType replica_type);
+        const std::string& key, ReplicaType replica_type,
+        std::optional<uint64_t> store_checksum = std::nullopt);
```
Aionw: I think Checksum or CrcChecksum would be a clearer name here. Store has both verb and noun meanings, which makes the current name somewhat ambiguous.

mooncake-store/include/master_client.h
```diff
@@ -233,7 +234,8 @@ class MasterClient {
      */
     [[nodiscard]] std::vector<tl::expected<void, ErrorCode>> BatchPutEnd(
         const std::vector<std::string>& keys,
-        ReplicaType replica_type = ReplicaType::ALL);
+        ReplicaType replica_type = ReplicaType::ALL,
+        const std::vector<std::optional<uint64_t>>& store_checksums = {});
```
Aionw: Checksum has 1:1 relation with key. To make this semantic more clear and avoid additional length validation, i'd like to change this function decl into

```cpp
struct ObjectMeta {
  std::string key;
  std::optional<uint64_t> checksum;
};

[[nodiscard]] std::vector<tl::expected<void, ErrorCode>> BatchPutEnd(
      const std::vector<ObjectMeta>& metas,
      ReplicaType replica_type = ReplicaType::ALL);
```

mooncake-store/include/master_client.h
yokinoshitayoki: > I think Checksum or CrcChecksum would be a clearer name here. Store has both verb and noun meanings, which makes the current name somewhat ambiguous.我认为 Checksum 或 CrcChecksum 作为名称会更清晰。Store 兼具动词和名词含义，导致当前名称存在一定歧义。

How about object_checksum?

mooncake-store/include/store_checksum.h
yokinoshitayoki: CrcChecksum sounds good.

mooncake-store/include/master_client.h
Aionw: Great. Personally, i prefer using `checksum`.

mooncake-store/include/real_client.h
```diff
@@ -553,11 +553,12 @@ class RealClient : public PyClient {
...
         const std::string &key, void *buffer, size_t dst_offset,
         size_t src_offset, size_t size, const RangedReadMetadata &metadata,
-        bool size_is_buffer_capacity = false);
+        bool size_is_buffer_capacity = false, bool verify_checksum = false);
 
     tl::expected<int64_t, ErrorCode> get_into_range_internal(
         const std::string &key, void *buffer, size_t dst_offset,
-        size_t src_offset, size_t size, bool size_is_buffer_capacity = false);
+        size_t src_offset, size_t size, bool size_is_buffer_capacity = false,
+        bool verify_checksum = false);
```
Aionw: `DummyClient::get_into()` still reaches `get_into_range_internal()` through the shared-memory range helper without passing either flag, so both defaults here are used. In particular, `verify_checksum=false` means the DummyClient full-object API can return corrupted data successfully, while `RealClient::get_into()` explicitly passes `true, true`. Could we remove these default booleans and add a dedicated full-read helper (or at least make the Dummy path pass the full-read semantics explicitly)? A regression test covering corrupted data through DummyClient would also catch this.

mooncake-store/tests/CMakeLists.txt
```diff
@@ -82,6 +82,12 @@ if(USE_NOF)
   add_store_test(nof_heartbeat_test nof_heartbeat_test.cpp)
 endif()
 add_store_test(client_integration_test client_integration_test.cpp)
+find_package(CUDAToolkit QUIET)
+if(CUDAToolkit_FOUND)
+  target_compile_definitions(client_integration_test
+                             PRIVATE MOONCAKE_STORE_TEST_CUDA)
+  target_link_libraries(client_integration_test PRIVATE CUDA::cudart)
```
Aionw: The checksum-dependent client tests all call `GTEST_SKIP()` unless `MOONCAKE_STORE_CHECKSUM=1`, but this CTest registration does not set that environment variable. The normal CI `ctest` invocation therefore skips all five checksum client tests and still exits successfully (I reproduced 0 passed / 5 skipped). Please register a dedicated filtered CTest with `set_tests_properties(... ENVIRONMENT "MOONCAKE_STORE_CHECKSUM=1")`. Also, the test filters in the PR description still use the old `StoreChecksum*` names and currently match zero tests.

mooncake-store/src/ha/snapshot/catalog_backed_snapshot_provider.cpp
```diff
@@ -141,11 +141,12 @@ DeserializeStandbyObjectMetadata(
         //   v3: 9 + replica_count, data_type + hard_pinned or
         //       hard_pinned + group_id
         //   v4: 10 + replica_count, data_type + hard_pinned + group_id
+        //   v5: 11 + replica_count, v4 + object_checksum (ignored here)
```
Aionw, line 144: Ignoring the checksum here means a standby loaded from this snapshot cannot retain it, and the HA oplog payload also has no checksum field. After promotion, checksum-enabled readers see `nullopt` and silently skip validation. The PR description calls this out as a limitation, but it makes integrity coverage disappear exactly across a failover. If preserving it is intentionally out of scope, can we at least expose a strict/require-checksum mode or an observable coverage status instead of treating the missing metadata as normal success?

mooncake-store/src/client_service.cpp
```diff
@@ -1094,15 +1117,96 @@ std::vector<tl::expected<QueryResult, ErrorCode>> Client::BatchQuery(
...
+
+tl::expected<void, ErrorCode> Client::VerifyObjectChecksum(
+    const std::string& object_key, const std::vector<Slice>& slices,
+    size_t object_size, std::optional<uint64_t> expected_checksum) {
+    if (!object_checksum_enabled_) {
+        return {};
+    }
+    if (!expected_checksum.has_value()) {
+        VLOG(1) << "object_checksum_absent key=" << object_key;
+        return {};
```
Aionw, line 1209: This returns the same success result for three materially different cases: checksum disabled, checksum metadata absent, and checksum actually verified. A caller using this diagnostic cannot tell whether a successful read had integrity coverage. Could the API expose an integrity status such as `NotRequested`, `Absent`, or `Verified`, and optionally support a `RequireChecksum` policy that fails when metadata is missing?

mooncake-store/src/master_service.cpp
```diff
@@ -9507,12 +9521,15 @@ MasterService::MetadataSerializer::SerializeMetadata(
...
     // has_soft_pin_timeout, soft_pin_timeout, replicas_count, data_type,
-    // replicas..., hard_pinned, group_id]
+    // replicas..., hard_pinned, group_id, object_checksum?]
 
     size_t array_size = 10;  // client_id, put_start_time, size, lease_timeout,
                              // has_soft_pin_timeout, soft_pin_timeout,
                              // replicas_count, data_type, hard_pinned, group_id
     array_size += metadata.CountReplicas();  // One element per replica
+    if (metadata.object_checksum.has_value()) {
+        ++array_size;
```
Aionw, line 9580: Appending this field is backward-readable by the new parser, but a pre-checksum binary only permits three optional fields and rejects a snapshot containing this fourth one. Consequently, after a checksum-enabled version writes a v5 snapshot, rolling back the master binary can fail to restore it. Please either make the snapshot envelope explicitly versioned/forward-skippable or document and test the required upgrade and rollback procedure.

mooncake-store/src/ha/snapshot/catalog_backed_snapshot_provider.cpp
yokinoshitayoki: > Ignoring the checksum here means a standby loaded from this snapshot cannot retain it, and the HA oplog payload also has no checksum field. After promotion, checksum-enabled readers see `nullopt` and silently skip validation. The PR description calls this out as a limitation, but it makes integrity coverage disappear exactly across a failover. If preserving it is intentionally out of scope, can we at least expose a strict/require-checksum mode or an observable coverage status instead of treating the missing metadata as normal success?

Preserving checksum metadata across standby promotion requires corresponding standby metadata and operation-log changes, which are intentionally out of scope for this diagnostic-only PR and are documented in the PR.

mooncake-store/src/client_service.cpp
yokinoshitayoki: In this PR, missing checksum metadata is intentionally treated as compatible success for existing objects and partially enabled deployments, with the absence recorded at VLOG(1). Exposing a new integrity status or strict policy would broaden the public GET contract, so we prefer to consider a require-checksum mode separately.

mooncake-store/src/CMakeLists.txt
```diff
@@ -305,6 +305,7 @@ target_link_libraries(
          gflags::gflags
          ${EXTRA_LIBS}
          ${SPDK_STATIC_LIBS}
+         mooncake_common
```
Aionw, line 308: This new dependency is propagated correctly to CMake consumers, but the Go binding bypasses CMake and manually lists static libraries in `.github/workflows/ci.yml`. The PR adds direct references from `client_service.o` to `Environ::Get()` and `CrcChecksum::Update()`, while the Go `CGO_LDFLAGS` still omit both `-L$GITHUB_WORKSPACE/build/mooncake-common/src` and `-lmooncake_common`. This matches the current CI failure in `Run Go store binding integration tests`; `libmooncake_store.a` has those two unresolved symbols and `libmooncake_common.a` defines them. Please update the Go link configuration, keeping `-lmooncake_common` after `-lmooncake_store` for static link order. I would not solve this with a global `link_libraries(mooncake_common)`: it would not affect the hand-written Go linker invocation and would unnecessarily spread mooncake_common/yaml/json dependencies to unrelated CMake targets.

mooncake-store/include/master_client.h
```diff
@@ -224,15 +224,16 @@ class MasterClient {
      * @return tl::expected<void, ErrorCode> indicating success/failure
      */
     [[nodiscard]] tl::expected<void, ErrorCode> PutEnd(
-        const std::string& key, ReplicaType replica_type);
+        const std::string& key, ReplicaType replica_type,
```
Aionw: Here we can reuse `ObjectMeta` to make api more consistent. Same for other api changed because adding check_sum argument.

mooncake-store/tests/client_integration_test.cpp
```diff
@@ -52,6 +62,102 @@ UUID ParseClientId(const std::string& client_id_str) {
...
+        client.VerifyObjectChecksum("key", slices, object_size, checksum)
+            .has_value());
+    value[0] ^= 0xFF;
+    auto mismatch =
+        client.VerifyObjectChecksum("key", slices, object_size, checksum);
+    ASSERT_FALSE(mismatch.has_value());
+    EXPECT_EQ(mismatch.error(), ErrorCode::CHECKSUM_MISMATCH);
+}
+
+#ifdef MOONCAKE_STORE_TEST_CUDA
```
Aionw: Unnecessary test. It just copy data from/to cuda, and check crc32 in dram buffer, which already tested in other cases.

Aionw (approved):

docs/source/deployment/mooncake-store-deployment-guide.md
```diff
@@ -979,6 +979,14 @@ Local hot cache provides a DRAM read cache on top of SSD-resident objects for fa
 | `MC_STORE_LOCAL_HOT_CACHE_USE_SHM` | unset | Set `1` to use memfd-backed shared memory |
 | `MC_STORE_LOCAL_HOT_ADMISSION_THRESHOLD` | unset | Minimum CountMinSketch count before a key is admitted to hot cache |
 
+#### Object-Level Checksum Diagnostics
+
+Set `MOONCAKE_STORE_CHECKSUM=1` on a Mooncake Store client process before the client is created to enable object-level CRC-64 checks. The client computes the checksum before `put`/`upsert`, stores it in master metadata, and verifies the logical `object_size` bytes returned by a full-object `get`. For complete diagnostic coverage, enable the switch on every writer and reader client. A client with the switch disabled does not generate or verify checksums; an enabled reader skips verification for objects whose metadata has no checksum.
```
ykwd, line 1023: Could you also update this in our troubleshooting docs and  ai skills?

Icedcoco:
Is `DummyClient::batch_query` expected to expose checksum metadata? It currently constructs `QueryResult` without forwarding `cached_result.value.object_checksum`, unlike the other query paths. Is this intentional, or should the checksum be passed through here as well?

docs/source/deployment/mooncake-store-deployment-guide.md
yokinoshitayoki: > Could you also update this in our troubleshooting docs and ai skills?

Thanks for reviewing, I have updated the docs and ai skills.

yokinoshitayoki:
> Is `DummyClient::batch_query` expected to expose checksum metadata? It currently constructs `QueryResult` without forwarding `cached_result.value.object_checksum`, unlike the other query paths. Is this intentional, or should the checksum be passed through here as well?

Thanks for pointing out. This was unintentional. `DummyClient::batch_query()` now preserves `object_checksum` by reusing `from_cached_query_result_response()`, and a regression test has been added to cover the RPC round trip.

ykwd (approved):

Icedcoco (approved):

PR #3004, opened 2026-07-20
[Store] Make RPC client I/O threads configurable via MC_RPC_CLIENT_IO_THREADS

Aionw:
## Description

Mooncake RPC clients currently allow YLT to initialize its default global
`io_context_pool` from the host CPU count. YLT runs one thread per context, so
large hosts can have a high client-side thread/eventfd baseline even when only a
few RPC connections are active.

This change:

- adds `MC_RPC_CLIENT_IO_THREADS` as a common fallback and component-specific
  `MC_STORE_RPC_CLIENT_IO_THREADS` and `MC_TE_RPC_CLIENT_IO_THREADS` overrides;
- gives Store/Master and Transfer Engine/TENT independent I/O pools so
  data-plane activity cannot starve Store control-plane RPCs;
- resolves the global, Store, and Transfer Engine thread counts once when the
  process-wide `Environ` singleton initializes, then exposes the frozen values
  through member accessors;
- introduces an injectable `EnvironSource`: production uses the OS environment,
  while tests use an in-memory mock without spawning separate processes;
- keeps Store and Transfer Engine pool entry points in their respective modules
  while common owns environment initialization and generic pool construction;
- defaults each pool to `min(16, max(1, hardware_concurrency()))`;
- replaces stale single-target MasterClient and DummyClient pools when their
  endpoint changes, while allowing in-flight requests to retain the old pool;
- applies `RpcCommunicatorConfig::pool_size` to YLT's per-target
  `max_connection` setting, keeps its default at `100` to preserve YLT's
  previously effective behavior and clarifies that it is not an I/O thread
  count;
- fixes static-link ordering for TENT and the Go binding after introducing the
  common pool accessors; and
- documents configuration precedence and `Environ` initialization/restart
  semantics.

Invalid component-specific values and `0` use the common fallback. Invalid
common values and `0` use the default.

This is a runtime configurability and resource-tuning change, not an eventfd
leak fix. The observed eventfds belong to Asio reactors activated by RPC I/O
contexts.

No existing open issue or pull request was found for this specific RPC client
I/O pool configurability change.

## Module

- [x] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] Refactor
- [ ] Breaking change
- [x] Documentation update
- [x] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test coverage:**

- Mock-source common tests for component overrides, fallback,
  initialization-time freezing, and independent pool identity/size.
- Store `rpc_timeout_test` and `dummy_client_get_buffer_test`.
- TENT `transport_uint_test`, plus successful linking of both
  `transport_uint_test` and `rdma_transport_test` in a `USE_TENT=ON` build.
- Standalone Transfer Engine build with TENT, including `transfer_engine` and
  `tent_shared`.
- Pre-commit on every touched file.
- Sphinx HTML build and generated-page inspection.

The documentation build succeeded with only external intersphinx inventory
warnings caused by the environment's proxy.

The local environment does not provide the Go toolchain, so the Go example was
not executed locally. The failing CI link command was updated to include
`mooncake_common` in its static-library rescan group.

## Checklist

- [x] For changes >500 LOC: I have filed an RFC issue (not applicable; this
      change is below the RFC threshold)

Pre-commit was run successfully on every touched file. The all-files mode was
not used to avoid rewriting unrelated files.

## AI Assistance Disclosure

- [ ] No AI tools were used

OpenAI Codex assisted with root-cause analysis, implementation, test execution,
review-feedback updates, and drafting this PR description. The submitter is
responsible for reviewing and defending the final change.


ykwd:
Shall we also change the default value to a more conservative number? For example: min(16, std::thread::hardware_concurrency())

mooncake-store/src/dummy_client.cpp
```diff
@@ -258,10 +253,7 @@ ErrorCode DummyClient::connect(const std::string& server_address) {
...
     MutexLocker lock(&connect_mutex_);
     if (client_addr_param_ != server_address) {
-        // WARNING: The existing client pool cannot be erased. So if there are a
-        // lot of different addresses, there will be resource leak problems.
-        auto client_pool = client_pools_->at(server_address);
-        client_accessor_.SetClientPool(client_pool);
+        client_accessor_.GetOrCreateClientPool(server_address);
         client_addr_param_ = server_address;
     }
     auto pool = client_accessor_.GetClientPool();
```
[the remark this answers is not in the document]
Aionw: Addressed in commit `03a5bb18`. I removed the redundant `GetClientPool()` read here because `invoke_rpc()` acquires and retains the current shared pool itself. Local lightweight validation passed (`clang-format-20`, `git diff --check`, and `bash -n mooncake-store/go/build.sh`); waiting for GitHub Actions to validate the Linux build and tests.

mooncake-store/src/master_client.cpp
```diff
@@ -419,10 +419,7 @@ ErrorCode MasterClient::Connect(const std::string& master_addr) {
...
     MutexLocker lock(&connect_mutex_);
     if (client_addr_param_ != master_addr) {
-        // WARNING: The existing client pool cannot be erased. So if there are a
-        // lot of different addresses, there will be resource leak problems.
-        auto client_pool = client_pools_->at(master_addr);
-        client_accessor_.SetClientPool(client_pool);
+        client_accessor_.GetOrCreateClientPool(master_addr);
         client_addr_param_ = master_addr;
     }
     auto pool = client_accessor_.GetClientPool();
```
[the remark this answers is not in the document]
Aionw: Addressed in commit `03a5bb18`. I removed the redundant `GetClientPool()` read here because `invoke_rpc()` acquires and retains the current shared pool itself. Local lightweight validation passed (`clang-format-20`, `git diff --check`, and `bash -n mooncake-store/go/build.sh`); waiting for GitHub Actions to validate the Linux build and tests.

Icedcoco:
Thanks for this change. I noticed two concrete linking issues in the current CI:

- Both TENT jobs fail because `tent_rpc` references `GetRpcClientIoContextPool()`, but the target does not link `mooncake_common` in the required order.
- The Go integration test uses its own `CGO_LDFLAGS` from `ci.yml`, so the `go/build.sh` update is not applied and `mooncake_common` remains missing there as well.

Both paths fail with `undefined reference to mooncake::GetRpcClientIoContextPool()`. Could you please update these link dependencies and rerun the affected jobs?

Aionw:
Fixed in latest commit, would you mind to take a look @Icedcoco

mooncake-transfer-engine/src/transport/rpc_communicator/rpc_communicator.cpp
```diff
@@ -60,12 +62,20 @@ bool RpcCommunicator::initialize(const RpcCommunicatorConfig& config) {
...
             coro_io::ib_socket_t::config_t{};
     }
+    if (config.pool_size > 0 &&
+        config.pool_size <= std::numeric_limits<uint32_t>::max()) {
+        pool_conf.max_connection = static_cast<uint32_t>(config.pool_size);
+    } else {
+        LOG(WARNING) << "Invalid RPC client per-target pool_size "
+                     << config.pool_size << "; using default "
+                     << pool_conf.max_connection;
+    }
```
alogfans, line 72: This changes the effective semantics of the public pool_size argument from an apparent thread-count setting to max_connection per target. Could we confirm this compatibility change and update the Python API documentation accordingly?

Aionw: Before this change, the configured pool size had no effect, so the number of client I/O threads defaulted to `std::thread::hardware_concurrency()`.

After this change, users can configure the client I/O thread count through environment variables and set the per-endpoint connection pool size using `pool_size`.

alogfans: Thanks, the implementation semantics are clear now. The remaining request is to document this in the Python-facing API: `pool_size` controls the maximum number of cached connections per target, while RPC I/O threads are configured through the environment variables. Please update the corresponding API documentation before resolving this thread.

alogfans:
LGTM for modifications of Transfer Engine.

mooncake-transfer-engine/src/transport/rpc_communicator/rpc_communicator.cpp
Aionw: I double-checked the call paths and found that my previous interpretation was incorrect. The existing `initialize()` documented here is `TransferEngine.initialize()`, which does not go through `RpcInterface` or use pool_size, so the pool_size description cannot be merged into that method.

Although `RpcInterface.initialize_client()` is technically exposed through pybind, it is currently used only by the RPC communicator test and has not been documented as a stable public Python API. Adding it here would unintentionally expand the documented public API surface.

Therefore, I reverted the Python API Reference changes. The pool_size semantics remain documented in the code, while the RPC client I/O environment variables remain covered by the deployment documentation.

ykwd (approved):

alogfans (approved):

PR #3009, opened 2026-07-21
[Store] Fix silent loss of offload keys deferred to the ungrouped pool (#3006)

### LINKED ISSUE #3006: [Bug]: [SSD] Offload keys deferred to the ungrouped pool lose their tasks — NACKed at deferral, silently dropped next cycle, never written to SSD
## Summary

With eager SSD offload enabled, `GroupOffloadingKeysByBucket` defers the **tail partial bucket** of each drain cycle to `ungrouped_offloading_objects_` so it can be written as part of a full bucket next cycle. But the deferral hand-off loses the key's `OffloadTaskItem`:

1. the deferred keys are **NACKed to the master in the same cycle** (they appear in no bucket, so the end-of-`OffloadObjects` sweep reports them with the `data_size = -1` sentinel), which erases their `offloading_tasks` entry and releases the source-replica refcount;
2. next cycle, the backend prepends the deferred keys into the first bucket — but `FileStorage::OffloadObjects` filters every bucket key through `task_by_storage_key`, which is **rebuilt from that cycle's drain only**. The deferred keys have no task, so they are **silently discarded**: never queried, never written, never notified, and (being inside `all_bucket_keys`) not even NACKed again.

Net effect: **every key that lands in the deferred tail is dropped from the offload pipeline with no log and no error.** The object keeps no SSD replica, its refcount pin is gone, and it is eventually memory-evicted — a cache entry lost while the SSD tier has free space. Since deferral fires whenever a drain's key count isn't a multiple of the effective bucket capacity, this is a steady-state leak, not an edge case.

## Code walk (`main` @ `fc04cb1`)

Drain is destructive — the master forgets the queue when the worker heartbeats (`master_service.cpp:4934`):

```cpp
local_disk_segment_it->second->offloading_objects.clear();
```

Deferral carries key+size only, not the task (`storage_backend.cpp:2313`):

```cpp
ungrouped_offloading_objects.emplace(bucket_object.first, bucket_object.second);
VLOG(1) << "Add offloading objects to ungrouped pool. ...";
return {};
```

The same-cycle sweep NACKs them (`file_storage.cpp:642`) — the comment lumps *deferred* keys in with *deliberately skipped* keys (size-limit / already-on-disk), for which NACK mop-up is correct:

```cpp
// Keys skipped by GroupOffloadingKeysByBucket don't appear in any bucket,
// so they never reach BatchOffload or complete_handler.
for (const auto& [storage_key, task] : task_by_storage_key) {
    if (all_bucket_keys.find(storage_key) == all_bucket_keys.end()) {
        failed_tasks.push_back(task);
    }
}
```

And next cycle the task-less deferred keys are dropped without a trace (`file_storage.cpp:516`):

```cpp
for (const auto& storage_key : keys) {
    const auto it = task_by_storage_key.find(storage_key);
    if (it != task_by_storage_key.end()) {
        storage_keys_by_tenant[it->second.tenant_id].push_back(storage_key);
    }
    // else: silently dropped — no store, no notify, no NACK, no log
}
```

## Production evidence

Observed on a 3-node shared pool (eager offload, ~5-object/~236 MB effective buckets, drain cycle ~10 s):

- **~12% of pool exits have no matching disk eviction** (cumulative last-replica exits minus disk-tier evictions) — i.e. that share of objects dies RAM-only while the SSD tier is under its eviction watermark with free space.
- Worker-side `VLOG(1)` attribution over 60 min on one node: `already exists` skip **0**, `size exceeds` skip **0**, D2H/store failures **0**, task-not-found **0** — but `Add offloading objects to ungrouped pool` **209** (≈ one deferral per drain cycle, matching `... have been processed` at 210).
- Master-side NACK count ≈ the RAM-only exit gap (~113–127% cumulatively; NACKs lead because the NACK happens at PUT-side drain, the exit at later eviction).
- Keys sampled from the NACK path and probed against the master's index minutes later: **19/21 absent** (contemporaneous control keys via the identical query: 21/21 present).

## Relationship to #2632 / #2658 — this is #2658's silent regression

Before #2658, these same deferred keys leaked *loudly*: their drained-but-unresolved tasks held a stale refcount pin until the 600 s TTL and produced the `Offloading task expired` warning storm. #2658's NACK sweep now cleans them up promptly — but since the deferral was never distinguished from a deliberate skip, the fix converted a visible leak into a silent one. We measured the same ~9–13% loss share both before and after deploying #2658; only the disposal path changed. (Verified with #2658 included: the NACK sentinel machinery works as designed — the problem is that deferred keys should never have been NACKed.)

Also related but distinct: #2827 / PR #2967 fix a different caller bug in the same function (whole-cycle abort on `OBJECT_ALREADY_EXISTS` skipping the NACK flush). #2967 does not touch the deferral hand-off; the two fixes compose.

Same silent-disposal family as #2997.

## Suggested fix

Either:

- **(A) Carry the task across the deferral.** Have `AllocateOffloadingBuckets` report which keys it deferred (out-param or getter); `FileStorage` keeps a small `deferred_task_by_storage_key_` map, excludes those keys from the NACK sweep, and merges the carried tasks into the next cycle's `task_by_storage_key` so the deferred keys complete normally. Keeps the refcount pin alive across the (one-cycle, ~10 s) deferral, which also guarantees the memory replica is still there to read. Bounded by one partial bucket.
- **(B) Flush the tail partial bucket instead of deferring.** Simpler; removes `ungrouped_offloading_objects_` entirely at a small write-efficiency cost (one sub-full bucket per drain cycle).

Happy to open a PR for (A) if that direction is acceptable.

Environment: observed on `v0.3.12-pre1` + #2658; code path verified byte-identical on `main` @ `fc04cb1`.

ccgibson:
## What

Fixes #3006: keys deferred to `ungrouped_offloading_objects_` by `GroupOffloadingKeysByBucket` lose their offload tasks across the deferral — they are NACKed to the master in the same cycle (freeing the task and the source-replica refcount) and then silently discarded on re-emission the next cycle, because `task_by_storage_key` is rebuilt from that cycle's drain only and the per-bucket filter drops task-less keys with no store, no notify, and no log. Every key landing in a drain cycle's tail partial bucket is lost from the offload pipeline; once memory-evicted, the object is gone even though the SSD tier has space.

Since #2658, the disposal is completely silent. Before #2658 the same population leaked loudly as the 600 s `Offloading task expired` storm — the NACK sweep cleaned up the orphaned tasks promptly but never distinguished *deferred* keys from *deliberately skipped* ones, converting a visible leak into an invisible one.

## How

Carry the task across the deferral (option A from the issue):

- `AllocateOffloadingBuckets` / `GroupOffloadingKeysByBucket` gain an optional out-param reporting the keys deferred to the ungrouped pool this call (`nullptr` default — existing call sites unchanged).
- `FileStorage` keeps a `deferred_task_by_storage_key_` map. Each `OffloadObjects` cycle merges the carried tasks into `task_by_storage_key` (so re-emitted keys are queried, stored, and success-notified normally) and erases the carried keys from `storage_object_sizes` so the backend does not group them a second time on top of its pooled copy.
- The end-of-cycle NACK sweep is extracted into `PartitionUnbucketedTasks` (a pure function, unit-tested directly): deliberately skipped keys (size over bucket limit / already on disk) still NACK; deferred keys are carried instead.

**One subtlety, encoded in the tests:** previously carried keys must be *retained* while un-emitted, not rebuilt from the deferral report alone. The grouping loop is `while (it != offloading_objects.cend())`, so a call whose effective input is empty (e.g. a drain consisting entirely of re-PUTs of carried keys, which the merge removes from the input) leaves the pool untouched — the pooled keys are neither emitted nor re-reported. Rebuilding the carry from `deferred_keys` alone would NACK their tasks while the pool still holds the keys, re-introducing the task-less drop. The first draft of this fix had exactly that bug; the regression test caught it.

Intended side effect: the source-replica refcount pin now survives the deferral, so the memory replica is guaranteed readable when the key is re-emitted (`BatchEvict` skips `refcnt > 0` victims). Worst-case cost is one partial bucket (≤ `bucket_keys_limit − 1` keys) of pinned memory while traffic is idle.

## Production evidence

Observed on a 3-node shared pool (eager offload, ~5-object/~236 MB effective buckets, ~10 s drain cadence) running `v0.3.12-pre1` + #2658:

**Before:** ~12% of pool exits had no corresponding disk-tier eviction — objects leaving the cache without ever reaching SSD — while the SSD tier sat under its eviction watermark. Worker-side `VLOG(1)` attribution over 60 min: `already exists` skips 0, `size exceeds` skips 0, D2H/store failures 0 — but ~209 deferral events/hour, matching the loss rate. 19/21 keys sampled from the NACK path were absent from the master index minutes later (21/21 contemporaneous controls present).

**After (this patch):** in the first 300 s under heavy eviction churn, pool exits and disk evictions advanced in exact lockstep (+8,518 each — gap growth zero). Over a 3.7-hour soak: **+92,350 pool exits vs +92,350 disk evictions — every object reached SSD before leaving** — and the NACK rate fell from ~50/min to ~0.02/min (the residual being genuine deliberate skips).

## Tests

- `GroupOffloadingKeysByBucket_reports_deferred_keys` — 35 keys / limit 10 → 3 buckets + 5-key tail reported as deferred; empty input leaves the pool untouched and reports nothing (the caller must retain); insufficient input re-defers and re-reports all pooled+new keys; sufficient input emits one exact bucket and drains the pool.
- `PartitionUnbucketedTasks_carries_deferred_not_nacked` — deferred key carried with its original task; un-emitted previously carried key retained; neither NACKed; deliberately skipped keys still NACKed.

`file_storage_test`, `offload_on_evict_test`, and `storage_backend_test` all pass on this branch (built from this tree's `docker/mooncake.Dockerfile` builder stage with `-DBUILD_UNIT_TESTS=ON`). Touched files formatted with clang-format 20 per CONTRIBUTING.

## Notes

- Composes with #2967 (different region of `OffloadObjects`; that fix addresses the whole-cycle abort on `OBJECT_ALREADY_EXISTS`).
- Alternative considered: flushing the tail partial bucket instead of deferring (removes `ungrouped_offloading_objects_` entirely at a small write-efficiency cost). Happy to rework in that direction if maintainers prefer.


ccgibson:
Pushed `37e159f` addressing the Codecov gap: the uncovered lines were the `OffloadObjects` integration seam (carry merge, deferral-report plumbing, partition call site, carry assignment), which the unit tests couldn't reach — the fixture constructs `FileStorage` with a null client.

The observation that makes it testable anyway: **a cycle in which every key defers emits no bucket and NACKs nothing, so nothing ever touches the master client.** The new `OffloadObjects_deferred_tail_tasks_carried` drives the integrated path end to end in the null-client harness — including phase 3's retained-carry corner (a drain that is solely a re-PUT of a carried key: empty effective input, pool neither emitted nor re-reported, carry must survive by retention). That is the exact case a rebuild-from-deferral-report simplification would get wrong, so it's now locked by a test at the integration level as well as the unit level.

`file_storage_test`: 26/26 on this branch, built from this tree's builder stage.

he-yufeng (commented):
Traced this end to end against `OffloadObjects` and the grouping path. The mechanism checks out: deferred keys were NACKed at deferral (master releases task + source refcount), then the pool re-emitted them task-less and the per-bucket filter dropped them with no store, no notify, no log. Extracting `PartitionUnbucketedTasks` as a pure function and covering the three pool states in tests (fresh deferral, empty-input retention, re-deferral) is exactly the right shape — the empty-input retention case is the easy one to miss and it's pinned down.

Two small things, neither blocking:

**1. Collision semantics when a deferred key is re-drained before re-emission.** In the merge, `emplace` keeps the freshly drained task but `storage_object_sizes.erase(storage_key)` discards the fresh size, so the pooled copy (older size) wins the grouping. If the same key can be re-PUT while deferred (upsert with a different value size), the bucket would be written with stale size accounting. The carried task holding the source-replica refcount means the old copy is at least still alive, so this is a semantics question rather than a safety hole — but it might be worth either letting the fresh drain's size win or adding a VLOG on the collision so the choice is explicit rather than accidental.

**2. The "should be unreachable" branch.** A deferred key with no known task currently falls through silently in `PartitionUnbucketedTasks`. Since the invariant (deferred ⊆ this cycle's input) is what makes it unreachable, a VLOG(1) or DCHECK there would turn a silent contract drift into an observable one if the caller ever changes.

Nice catch on the ~12% figure in #3006 — that kind of silent loss rate on the SSD tier would have been painful to hunt down in production.

ccgibson:
Thanks for the careful trace — both points were right, and both are addressed in `ee876a3`.

**1. Collision semantics — you caught a genuinely accidental split.** The fresh task won the task map while the pooled stale size won the packing accounting; neither half was chosen on purpose. Fixed by making the fresh drain win *entirely*: a new `BucketStorageBackend::RemoveUngroupedOffloadingObject` drops the pooled copy on collision, so the re-drained key flows through grouping as ordinary new input — current size, fresh task — with a `VLOG(1)` marking the event. This also simplified the merge: non-collision carried keys were never in `storage_object_sizes` to begin with, so the old unconditional `erase()` only ever did work in the collision case — the new shape makes that explicit instead of incidental. Phase 4 added to the integration test: re-PUT of a deferred key with a *different* size, asserting the carried task ends up with the fresh size (`task.size == 2`), pool and carry cardinality unchanged.

**2. The unreachable branch now logs.** I went with `LOG(WARNING)` rather than `VLOG(1)`/`DCHECK`, on the reasoning that contract drift should surface in default-verbosity production logs (where it would actually matter), the branch is unreachable-by-invariant so it cannot spam, and `DCHECK` is invisible in release builds. Happy to switch if the project convention prefers otherwise.

All three suites re-run green on this branch (`file_storage_test` 26/26 incl. the new phase, `offload_on_evict_test` 9/9, `storage_backend_test` 72/72), built from this tree's builder stage.

he-yufeng:
Verified `ee876a3` against both points.

1. The collision is now decided rather than split: on a re-drained carried key the pooled copy is dropped via `RemoveUngroupedOffloadingObject` and the key flows through grouping as new input, so the fresh task and the current size win together. The simplification checks out too — non-collision carried keys are by construction absent from `storage_object_sizes`, so the old unconditional `erase` only ever did work in the collision case; the explicit `find` + `continue` reads much better. Phase 4 pins exactly the invariant that mattered (re-PUT with a different size while deferred, fresh size must supersede).

2. WARNING for the unreachable branch is a better call than my VLOG(1) suggestion — unreachable by invariant means it can't spam, and default-verbosity visibility is the whole point when the contract does drift.

LGTM from my side.

mooncake-store/src/file_storage.cpp
```diff
@@ -440,6 +440,46 @@ tl::expected<FileStorage::BatchGetResult, ErrorCode> FileStorage::BatchGet(
...
+        }
+    }
+    // Retain previously carried keys the backend did not emit this cycle:
+    // their pooled copies are still waiting (the pool only moves on a call
+    // with non-empty input), so their tasks must stay alive.
+    for (const auto& [storage_key, task] : previously_carried) {
+        if (all_bucket_keys.find(storage_key) == all_bucket_keys.end()) {
+            carried_tasks.emplace(storage_key, task);
+        }
+    }
```
LujhCoconut, line 483: Low-probability, but the fix is a one-line guard — might as well close the door while we're here.

When a deferred key is re-drained and the merge loop handles the collision (pool removal + stale task skipped), the second loop in `PartitionUnbucketedTasks` shouldn't retain the stale task from `previously_carried`.

Example: key X deferred at 1GB, re-drained next cycle at 5GB > `bucket_size_limit`.

Merge loop: collision → `RemoveUngroupedOffloadingObject("X")`, stale task skipped. `task_by_storage_key["X"] = task_5GB`. But `deferred_task_by_storage_key_["X"]` still holds task_1GB.
Grouping: 5GB > limit → X skipped entirely. No bucket, no deferred_keys, no pool.
Second loop: `for (const auto& [storage_key, task] : previously_carried)`: X not in `all_bucket_keys` → `carried_tasks.emplace("X", task_1GB)` (zombie).
Third loop: `carried_tasks` already has X → task_5GB never NACKed. Lost + refcount leak.

Fix:
```cpp
        if (all_bucket_keys.find(storage_key) == all_bucket_keys.end() &&
              task_by_storage_key.find(storage_key) == task_by_storage_key.end()) {
```

ccgibson: Confirmed and fixed in `92874c5` — thanks, this is a real leak and the trace is exactly right.

I took a different guard than the one suggested, because the proposed condition turns out to void the retention loop entirely rather than just excluding the collision case.

**Why `task_by_storage_key` can't be the test.** The carry merge inserts every *non-collision* carried task into `task_by_storage_key` (`file_storage.cpp:522`) precisely so the re-emitted keys can be queried and success-notified. So after the merge, `task_by_storage_key` ⊇ every key in `previously_carried` — collision keys via the fresh drain, non-collision keys via that emplace. `task_by_storage_key.find(...) == end()` is therefore always false in the retention loop, and the loop becomes dead code. Applied verbatim, it fails the PR's original regression test:

```
[  FAILED  ] FileStorageTest.PartitionUnbucketedTasks_carries_deferred_not_nacked
[  FAILED  ] FileStorageTest.PartitionUnbucketedTasks_merged_carry_still_retained
```

(the second is a new companion test added for exactly this distinction). It does fix the case you reported, and the integration test still passes — in `OffloadObjects_deferred_tail_tasks_carried` the collision only removes the *pooled* copy, not the entry in `storage_object_sizes`, so the grouping call still reports those keys in `deferred_keys` and the first loop carries them without help from retention. But `pooled_waiting` in the unit test is the case retention exists for, and it gets NACKed while the pool still holds the key — the #3006 drop again.

**The fix.** The property that actually matters is "re-drained this cycle", so I record it where the decision is made — the collision branch of the carry merge inserts into a `redrained_carried_keys` set, which `PartitionUnbucketedTasks` takes as a parameter and skips in the retention loop. Keeping it co-located with the pool removal means the rule and the mutation that creates the hazard can't drift apart.

**Test.** `PartitionUnbucketedTasks_redrained_then_skipped` drives your exact scenario (carried at size 1, re-drained at size 5, over `bucket_size_limit` so `GroupOffloadingKeysByBucket` hits the `++it; continue` at `storage_backend.cpp:2339` and the key lands in neither `buckets_keys` nor `deferred_keys`). Without the exclusion it fails with precisely the two symptoms you predicted — carry non-empty (the zombie) and `failed_tasks.size() == 0` instead of 1 — and it asserts the NACKed task is the *fresh* one (`size == 5`).

This corner can't be driven at integration level: by construction it produces a NACK, and the NACK path dereferences the null `client_` the deferral-carry fixture depends on. Noted in a comment there pointing at the unit test.

One consequence worth stating explicitly: the stale task is now dropped without a NACK, which matches what the ≤-limit collision path already did at `:520`. If the master keys `offloading_tasks` by task ID rather than by key, both paths leak the stale task's refcount until the TTL reaper runs — a pre-existing property of the collision design rather than something this change introduces, but say the word if you'd rather the stale task be NACKed instead of dropped and I'll fold that in.

Also worth noting: with this exclusion, the retention loop has no reachable case left against the current backend — every pooled key is either emitted in a bucket or re-reported in `deferred_keys`, and the collision path is the only thing that removes a key from the pool behind the carry's back. I've left it in as a defensive retention with its invariant documented, in the same spirit as the unreachable branch above it, rather than deleting it.

Suites re-run green on `92874c5`, built from this tree: `file_storage_test` 28/28 (26 + the two new), `offload_on_evict_test` 9/9, `storage_backend_test` 72/72.

LujhCoconut (approved):
LGTM. Thanks for this PR — the `redrained_carried_keys` approach correctly fixes the re-drain collision zombie without breaking the `pooled_waiting` retention path your original `task_by_storage_key` test would have killed. Production evidence and test coverage are solid. Nice.

fcczzz (approved):
Thanks for the fix! I reviewed the current head, and the deferred-task handling looks good to me. No blocking findings.

Are you still planning to move this forward? If so, could you resolve the merge conflicts and rerun CI?

PR #3012, opened 2026-07-21
[Store] Implement asynchronous P2P Redis oplog persistence

skygragon:
## Description

  P2P metadata uses a memory-first, non-transactional consistency model. This change makes Redis OpLog persistence asynchronous, keeping metadata mutations on the request path while dispatching OpLog writes to bounded background workers.

  Key changes:

  - Use a bounded asynchronous Redis OpLog queue with configurable `reject` or `bypass` behavior on overflow.
  - Retry admitted required entries indefinitely and use bounded retries for best-effort `ADD_REPLICA` entries.
  - Make `AddReplica` consistently update memory before recording its OpLog.
  - Allow the Redis progress watermark to advance across explicitly dropped entries.
  - Remove the Redis OpLog index and let readers scan sequence keys while skipping holes.
  - Reuse the common `OpLogApplier` ordering path for P2P entries.
  - Add configuration and tests for queue overflow, retries, holes, and P2P replay behavior.
  - Leave explicit TODOs for later standby out-of-order and gap-timeout handling.

  ## Module

  - [ ] CI/CD
  - [ ] Docs
  - [ ] Other

  ## Type of Change

  - [ ] Bug fix
  - [x] New feature
  - [x] Refactor
  - [ ] Breaking change
  - [ ] Documentation update
  - [x] Performance improvement
  - [ ] Other

  ## How Has This Been Tested?

  The affected targets were formatted and rebuilt in the development container. Direct P2P/Redis OpLog tests and the Redis HA end-to-end chaos test passed.

  `p2p_client_integration_test` passed 19 of 20 cases. The remaining `LargePutGet` failure is an existing intermittent Transfer Engine data mismatch; isolated repetition observed 14 passes and 6 failures out of 20 runs.

  **Test commands:**

  ```bash
  docker exec mooncake-dev cmake --build /workspace/build-redis-oplog -j4 \
    --target redis_oplog_store_test p2p_oplog_test \
    p2p_master_service_test p2p_client_integration_test \
    redis_chaos_test mooncake_master clientctl

  docker exec mooncake-dev bash -lc '
    cd /workspace/build-redis-oplog &&
    ctest -j1 \
      -R "^(redis_oplog_store_test|p2p_oplog_test|p2p_master_service_test|p2p_client_integration_test|redis_chaos_test)$" \
      --output-on-failure
  '
```
  Test results:

  - [x] Unit tests pass
  - [x] Redis HA end-to-end tests pass
  - [ ] Integration tests pass (one existing intermittent LargePutGet failure)
  - [ ] Manual testing done

  ## Checklist

  - [ ] I have formatted my code using ./scripts/code_format.sh

  ## AI Assistance Disclosure

  - [ ] No AI tools were used

  OpenAI Codex assisted with implementation, conflict resolution during rebase, code review, formatting, and test execution. The human submitter reviewed the design and implementation decisions.

### DIFF
--- a/mooncake-store/src/ha/oplog/oplog_applier.cpp
+++ b/mooncake-store/src/ha/oplog/oplog_applier.cpp
@@ -119,25 +119,12 @@ bool OpLogApplier::ApplyOpLogEntry(const OpLogEntry& entry) {
         return false;
     }
 
-    // Apply the operation based on type
-    switch (entry.op_type) {
-        case OpType::PUT_END:
-            ApplyPutEnd(entry);
-            break;
-        case OpType::PUT_REVOKE:
-            ApplyPutRevoke(entry);
-            break;
-        case OpType::REMOVE:
-            ApplyRemove(entry);
-            break;
-        default:
-            if (!ApplyCustomOpLogEntry(entry)) {
-                LOG(ERROR) << "OpLogApplier: unsupported or failed op_type="
-                           << static_cast<int>(entry.op_type)
-                           << ", sequence_id=" << entry.sequence_id
-                           << ", key=" << entry.object_key;
-                return false;
-            }
+    if (!ApplyOpLogEntryInternal(entry)) {
+        LOG(ERROR) << "OpLogApplier: unsupported or failed op_type="
+                   << static_cast<int>(entry.op_type)
+                   << ", sequence_id=" << entry.sequence_id
+                   << ", key=" << entry.object_key;
+        return false;
     }
 
     // Update expected sequence ID
@@ -155,6 +142,22 @@ bool OpLogApplier::ApplyOpLogEntry(const OpLogEntry& entry) {
 }
 
 bool OpLogApplier::ApplyCustomOpLogEntry(const OpLogEntry&) { return false; }
+
+bool OpLogApplier::ApplyOpLogEntryInternal(const OpLogEntry& entry) {
+    switch (entry.op_type) {
+        case OpType::PUT_END:
+            ApplyPutEnd(entry);
+            return true;
+        case OpType::PUT_REVOKE:
+            ApplyPutRevoke(entry);
+            return true;
+        case OpType::REMOVE:
+            ApplyRemove(entry);
+            return true;
+        default:
+            return ApplyCustomOpLogEntry(entry);
+    }
+}
 
 size_t OpLogApplier::ApplyOpLogEntries(const std::vector<OpLogEntry>& entries) {
     size_t applied_count = 0;
@@ -273,23 +276,7 @@ size_t OpLogApplier::ProcessPendingEntries() {
         }
 
         // Apply outside lock.
-        bool applied = true;
-        switch (entry_copy.op_type) {
-            case OpType::PUT_END:
-                ApplyPutEnd(entry_copy);
-                break;
-            case OpType::PUT_REVOKE:
-                ApplyPutRevoke(entry_copy);
-                break;
-            case OpType::REMOVE:
-                ApplyRemove(entry_copy);
-                break;
-            default:
-                applied = ApplyCustomOpLogEntry(entry_copy);
-                break;
-        }
-
-        if (!applied) {
+        if (!ApplyOpLogEntryInternal(entry_copy)) {
             LOG(ERROR) << "OpLogApplier: failed to apply pending entry"
                        << ", sequence_id=" << entry_copy.sequence_id
                        << ", op_type=" << static_cast<int>(entry_copy.op_type);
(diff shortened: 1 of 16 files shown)

### CODE REVIEW

File: mooncake-store/src/ha/oplog/oplog_applier.cpp [hunk 1]
```diff
@@ -276,12 +285,24 @@ size_t OpLogApplier::ProcessPendingEntries() {
...
 
+        if (!applied) {
+            LOG(ERROR) << "OpLogApplier: failed to apply pending entry"
+                       << ", sequence_id=" << entry_copy.sequence_id
+                       << ", op_type=" << static_cast<int>(entry_copy.op_type);
+            std::lock_guard<std::mutex> lock(pending_mutex_);
+            pending_entries_.emplace(entry_copy.sequence_id,
+                                     std::move(entry_copy));
+            break;
+        }
```
wanyue-wy: The `applied` variable feels a bit redundant here — I'd suggest merging this logic directly into the `ApplyCustomOpLogEntry` call instead.

mooncake-store/src/p2p_master_service.cpp
```diff
@@ -30,7 +32,8 @@ ErrorCode P2PMasterService::RecordOplog(OpType type, const std::string& key,
         return ErrorCode::OK;
     }
 
-    auto result = manager->AppendAndPersist(type, key, payload, sync);
```
wanyue-wy, line 33: Why do we have both `sync` and `async_oplog_` controlling whether logging asynchronously? Could we consolidate them, or add some comments and rename the variables to make their purposes clearer?

mooncake-store/src/p2p_master_service.cpp
```diff
@@ -415,20 +416,20 @@ tl::expected<void, ErrorCode> P2PMasterService::InnerAddReplica(
...
                         SerializeP2PPayload(payload),
-                        /*sync=*/!async_add_replica_oplog_);
+                        /*sync=*/!async_oplog_);
         if (record_err != ErrorCode::OK) {
             LOG(ERROR) << "AddReplica(P2P): failed to record oplog"
                        << ", client_id=" << client_id
                        << ", segment_id=" << segment_id
                        << ", error=" << toString(record_err);
             return tl::make_unexpected(record_err);
         }
```
wanyue-wy, line 433: This change could cause an issue: if logging fails, the `master` still adds the route to its in‑memory table and returns an error code to the `client`. To keep local data as consistent as possible with the `master`'s routing state, the `client` would then delete its local copy. That means a logging failure would leave the client and master out of consistency.

mooncake-store/src/p2p_master_service.cpp
```diff
@@ -439,22 +440,22 @@ tl::expected<void, ErrorCode> P2PMasterService::InnerAddReplica(
...
-        auto emplace_it =
```
wanyue-wy, line 453: Same as above

mooncake-store/src/ha/oplog/redis_oplog_store.cpp
```diff
@@ -171,7 +169,7 @@ ErrorCode RedisOpLogStore::EnqueueWrite(const OpLogEntry& entry, bool sync) {
     }
 
     std::shared_ptr<PendingWrite> pending;
-    bool wait_for_completion = sync;
+    const bool wait_for_completion = sync;
     {
```
wanyue-wy: This variable is never modified. Is `sync` sufficient?

mooncake-store/src/ha/oplog/redis_oplog_store.cpp
```diff
@@ -338,52 +375,21 @@ ErrorCode RedisOpLogStore::PersistEntryNoLatest(redisContext* ctx,
     const std::string entry_key = EntryKey(entry.sequence_id);
     const std::string serialized = SerializeOpLogEntry(entry);
     RedisReplyPtr reply((redisReply*)redisCommand(
-        ctx, "SET %b %b NX", entry_key.data(), entry_key.size(),
-        serialized.data(), serialized.size()));
-    if (!reply || reply->type == REDIS_REPLY_ERROR) {
+        ctx, "SET %b %b", entry_key.data(), entry_key.size(), serialized.data(),
+        serialized.size()));
+    if (!reply || reply->type != REDIS_REPLY_STATUS) {
```
wanyue-wy, line 383: Why switch from `SET NX` to `SET`? Could this lead to silent overwrite errors?

skygragon: SET is intentional: a previous primary may leave uncommitted entries beyond latest, and the new primary must be able to overwrite them. We’ll add an atomic guard later to reject conflicting overwrites at or below committed latest.

PR #2959, opened 2026-07-16
[TE] Improve RDMA NIC failover recovery

alogfans:
## Description

This PR improves recovery when a local or peer RNIC becomes unavailable during RDMA transfers, especially under real link down/up scenarios. Refer to #2960

### Changes

- Add local RNIC failure handoff so slices can be retried through another active local RDMA context.
- Publish updated local topology when a context becomes inactive or recovers.
- Delay reactivating a recovered RNIC after `IBV_EVENT_PORT_ACTIVE`.
- Improve passive handshake failure replies so peers receive structured rejection reasons.
- Retry passive setup once with a fresh endpoint when a stale endpoint is retired.
- Rebuild pre-connected QPs instead of reusing partially handshaken QPs.
- Avoid selecting an inactive local context from request-level device preselection.
- Add a real RDMA link failover validation script.
- Extend RDMA reestablish/topology tests for sender and receiver RNIC failover.

## Module

- [X] Transfer Engine (`mooncake-transfer-engine`)
- [ ] Mooncake Store (`mooncake-store`)
- [ ] Mooncake EP (`mooncake-ep`)
- [ ] Mooncake PG (`mooncake-pg`)
- [ ] Integration (`mooncake-integration`)
- [ ] P2P Store (`mooncake-p2p-store`)
- [ ] Python Wheel (`mooncake-wheel`)
- [ ] Common (`mooncake-common`)
- [ ] Mooncake RL (`mooncake-rl`)
- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
# Example: bash scripts/run_ci_test.sh
```

**Test results:**
- [X] Unit tests pass
- [X] Integration tests pass (if applicable)
- [X] Manual testing done (describe below)

- Ran unit tests:

  ```bash
  ./build/mooncake-transfer-engine/tests/topology_test
  ./build/mooncake-transfer-engine/tests/rdma_context_reprobe_test
  ```

- Ran RDMA failover tests with `mlx5_1` through `mlx5_4`:

  ```bash
  MC_METADATA_SERVER=P2PHANDSHAKE \
  MC_TARGET_DEVICE_NAME=mlx5_3,mlx5_4 \
[...]

PR #3016, opened 2026-07-21
[TENT] Port recent TE RDMA reliability updates [WIP]

alogfans:
## Description

This PR ports recent classic Transfer Engine RDMA reliability behavior into TENT and strengthens the compatibility/performance/test baseline.

The branch covers:
- TE compatibility semantics for the TENT-backed TransferEngine path.
- RDMA bad-NIC / bad-link recovery behavior in TENT.
- Auto-GID ranking, reprobe, retry, and metadata refresh logic.
- Active reconnect pause for failed peers.
- Endpoint lifecycle cleanup needed for safe reconnect/failover.
- TENT test coverage aligned with existing TE RDMA unit tests.

## Related TE PRs Covered

This branch intentionally folds in behavior from the following upstream Transfer Engine PRs:

- kvcache-ai/Mooncake#2959 - `[TE] Improve RDMA NIC failover recovery`
- kvcache-ai/Mooncake#2941 - `[Bugfix][TransferEngine] Pause active reconnects to failed peers`
- kvcache-ai/Mooncake#2878 - `[Transfer Engine] Refresh RDMA metadata on HCA and GID change events`
- kvcache-ai/Mooncake#2748 - `fix(rdma): rank link-local GID below any network-backed IPv4 in auto-GID`
- kvcache-ai/Mooncake#2741 - `[TransferEngine] Prefer private-range IPv4 GIDs over link-local IPv6 in auto-selection`
- kvcache-ai/Mooncake#2524 - `[TransferEngine] Fix batch retry RNIC hint advancement`
- kvcache-ai/Mooncake#2269 - `fix(transfer_engine): improve auto gid selection and retry`
- kvcache-ai/Mooncake#2427 - `[TE] Gate eRDMA QP reconstruction behind USE_ERDMA so non-eRDMA reconnect flushes in-flight WRs`

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

```bash
git diff --check
cmake --build build --target tent_rdma_transport_test tent_endpoint_store_test transfer_engine_config_override_test -j8
ctest --test-dir build -R "tent_rdma_transport_test|tent_endpoint_store_test|transfer_engine_config_override_test" --output-on-failure
build/mooncake-transfer-engine/tent/tests/tent_rdma_transport_test --gtest_color=no
```

**Test results:**
- [X] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used
- [X] AI tools were used (specify below) Assisted with CodeX, manual rechecked.


mooncake-transfer-engine/tent/src/transport/rdma/workers.cpp
```diff
@@ -797,6 +819,42 @@ void Workers::monitorThread() {
...
+                    std::string previous_gid;
+                    std::string next_gid;
+                    auto refresh_result =
+                        context->refreshCurrentGid(&previous_gid, &next_gid);
+                    if (refresh_result == GidRefreshResult::CHANGED) {
+                        LOG(WARNING)
+                            << "GID changed for RDMA context "
+                            << context->name() << ": " << previous_gid << " -> "
+                            << next_gid << ", clearing endpoints";
+                        if (context->endpointStore()->clear()) {
```
jfeng18, line 833: Reading `monitorThread()`: on `GidRefreshResult::CHANGED` it calls `endpointStore()->clear()`. The `clear()` implementation documents itself as a terminal shutdown path for `RdmaContext::disable()` — "Context shutdown has already stopped workers, so the normal two-phase waiting-list path cannot drain CQ flush completions. Force synchronous deconstruction here." At this call site workers are still running, and endpoints being deconstructed may hold in-flight WRs.

The poll-side `ep_weak_ptr.lock()` null branch suggests late flush completions were considered, so does the intended semantics here rely on that path? Also noting the classic TE GID_CHANGE handler calls `disconnectAllEndpoints()`, which only `beginDestroy()`s and moves endpoints to the waiting list (async two-phase teardown) rather than synchronously destroying QPs — was switching to synchronous `clear()` here deliberate? And since this thread also dispatches async events for all contexts, a `destroy_qp` that stalls on flush drain would stall event handling too.

mooncake-transfer-engine/tent/src/transport/rdma/workers.cpp
```diff
@@ -797,6 +819,42 @@ void Workers::monitorThread() {
         if (time_since_last_reclaim >= 1000) {  // 1 second = 1000 ms
             for (auto& context : transport_->context_set_) {
                 context->endpointStore()->reclaim();
+                context->pruneConnectPause();
+                if (context->status() == RdmaContext::DEVICE_ENABLED) {
+                    std::string previous_gid;
+                    std::string next_gid;
+                    auto refresh_result =
+                        context->refreshCurrentGid(&previous_gid, &next_gid);
```
jfeng18, line 827: The classic TE refresh (#2878) is event-driven — `IBV_EVENT_GID_CHANGE` is handled in the async-event loop. Here GID changes are detected by a 1 Hz poll in `monitorThread()` (full GID table scan + sysfs ndev reads per NIC per second in auto-GID mode), and `handleContextEvents()` has no GID_CHANGE/LID_CHANGE branches.

Was the polling approach chosen because the TENT event loop can't host the metadata refresh, or is it interim? Also, in IB (non-RoCE) fabrics a LID change is never detected by this loop — `refreshCurrentGid()` only compares GIDs, and `lid_` isn't refreshed, so the published device desc keeps the stale LID. Is IB fabric support in scope for this port?

jfeng18 (commented):
Did a read-through of this port focusing on endpoint lifecycle and the reliability paths. The port fidelity is impressive — `rdma_gid_probe.h` and `connect_pause_tracker.h` are byte-aligned with the classic TE copies (modulo namespace), and the lock structure (endpoint `lock_` → context `gid_reprobe_lock_`/`gid_lock_`; store lock → endpoint lock) checks out with no inversion found.

**One rebase note first** (couldn't anchor it to a line since the code in question is unchanged relative to this branch's base): `accept()` still carries the pre-#3188 behavior on the EP_READY/mismatch path (reject without retiring). #3188 (merged 2026-08-04, fixes #3173) changed that branch on main to `beginDestroyNoLock()` + retire so a peer that comes back with fresh QP numbers can actually reconnect. Reading the code, the recovery chain introduced here (GID refresh → `clear()` → local endpoints re-created with new QP nums → re-bootstrap to peer) lands exactly on that path — it looks like this PR's failover self-healing depends on #3188's retire semantics being preserved through the rebase. Worth double-checking the conflict resolution keeps it, and maybe worth a test where the initiator rebuilds endpoints (new QP nums) after the passive side is already EP_READY.

Two line-level questions below, both about deliberate semantic choices relative to classic TE.

One minor non-blocking note as well: the TENT copy of `rdma_gid_probe.h` drops the #2729 comments explaining the private-V4 vs degraded tier split, and `autoGidCandidateClassPriority` now relies on enum ordering via `static_cast` where the classic copy uses an explicit switch. Since both copies must stay behaviorally in sync, keeping those comments (or eventually sharing one header) would help future tier changes not silently diverge.

PR #3017, opened 2026-07-21
[TENT] Remove dead metrics config flags and wire validateConfig into initialize()

staryxchen:
## Description

The TENT metrics system declared several configuration options that had no
effect at runtime, and shipped validation logic that was never invoked.

**Problem 1 — dead config flags.** `MetricsConfig::enable_prometheus` and
`enable_json` (plus their `TENT_METRICS_ENABLE_PROMETHEUS` / `_JSON` env vars
and `metrics/enable_prometheus` / `metrics/enable_json` JSON keys) were parsed
into the config struct and documented, but `registerHttpHandlers()` registered
all HTTP endpoints unconditionally. The flags were silently ignored, so the
configuration surface lied to operators.

**Problem 2 — never-called validation.** `MetricsConfigLoader::validateConfig()`
existed (checking port != 0, threads != 0) but was never invoked from
`TentMetrics::initialize()`. An invalid config would proceed to
`initHttpServer()` and fail with a confusing error instead of a clear
validation message.

**Fix (Occam: remove rather than implement).** The `enable_prometheus` /
`enable_json` flags provide negligible value — `/metrics` and `/metrics/json`
are both read-only on the same port, and there is no realistic scenario where
a deployment wants one disabled but not the other. Removing them is simpler
than honoring them. `validateConfig` is wired into `initialize()` so invalid
configs fail fast with a clear error; on rejection `initialized_` is reset so
the caller can fix the config and retry.

The previously-removed `latency_buckets` / `size_buckets` keys set the project
precedent of silently ignoring removed config keys, with a "Compatibility"
note in `docs/source/design/tent/metrics.md`. The two flags removed here get
the same treatment.

## Module

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

The change is confined to the config/parsing layer and the `initialize()`
entry point; it does not touch the transfer hot path or any recording logic.
Verification was therefore scoped to the config layer + initialization gate
rather than real-transfer benchmarks.

**Test commands:**
```bash
# Full TENT build with metrics enabled
cmake -DCMAKE_BUILD_TYPE=Debug -DUSE_TENT=ON -DTENT_METRICS_ENABLED=ON \
      -DBUILD_UNIT_TESTS=ON -DUSE_HTTP=ON -DUSE_TCP=ON -DWITH_METRICS=ON \
      -DUSE_CUDA=OFF -DUSE_HIP=OFF -DUSE_ETCD=OFF -DUSE_MNNVL=OFF -DUSE_TPU=OFF \
      -S mooncake-transfer-engine -B mooncake-transfer-engine/build-tent-metrics
cmake --build mooncake-transfer-engine/build-tent-metrics -j384

# Config + HTTP + recording test suites
cd mooncake-transfer-engine/build-tent-metrics/tent/tests
./metrics_config_loader_test      # 25/25
./tent_metrics_http_server_test   #  1/1  (port-busy degradation path)
./tent_metrics_recording_test     # 10/10 (real HTTP client scrape)

# validateConfig wiring smoke (temp binary, deleted after run):
# Constructs MetricsConfig with http_port=0, calls initialize(),
# asserts InvalidArgument return + isInitialized()==false + retry-with-valid succeeds.
```

**Test results:**
- [x] Unit tests pass — 36/36 across the 3 suites above
- [ ] Integration tests pass (if applicable) — N/A: config-layer change, no
      transfer-path impact; `tebench` not run because the change does not touch
      transfer or recording code
- [x] Manual testing done — validateConfig smoke confirms the new gate rejects
      `http_port=0` with `InvalidArgument`, resets `initialized_` for retry,
      and accepts a valid config afterward

Full build completed in ~2 min with no errors or warnings on changed files.

## Checklist

- [x] I have updated the documentation (if applicable) — Compatibility note added to `docs/source/design/tent/metrics.md`; example config + env var docs updated
- [x] I have added tests to prove my changes are effective — existing config loader tests updated; validateConfig smoke covers the new wiring
- [x] For changes >500 LOC: I have filed an RFC issue — N/A (+19/-81)

## AI Assistance Disclosure

An AI coding agent implemented the mechanical edits (field removal, validateConfig
wiring, test/doc updates) under human direction. The human submitter has reviewed
every changed line and can defend the change end-to-end: the removal rationale
(dead flags with no behavioral effect), the `validateConfig` wiring (fail-fast at
`initialize()` entry with retry-friendly state reset), and the test coverage.

### DIFF
--- a/docs/source/design/tent/metrics.md
+++ b/docs/source/design/tent/metrics.md
@@ -349,7 +349,7 @@ The `metrics/latency_buckets` and `metrics/size_buckets` config keys, and the `T
 
 The `metrics/enable_prometheus` and `metrics/enable_json` config keys, and the `TENT_METRICS_ENABLE_PROMETHEUS` / `TENT_METRICS_ENABLE_JSON` environment variables, were removed and are now silently ignored. The `/metrics` and `/metrics/json` HTTP endpoints are now registered unconditionally and cannot be toggled independently — both are read-only and served on the same port, so there was no real scenario where a deployment wanted one disabled.
 
-**Migration:** remove these keys from your `transfer-engine.json` and unset the environment variables. If you need to disable the HTTP endpoint entirely, set `metrics/enabled` to `false` (or `TENT_METRICS_ENABLED=false`) to run in log-only mode.
+**Migration:** remove these keys from your `transfer-engine.json` and unset the environment variables. To fully disable the metrics subsystem (no HTTP server, no periodic summary logging, no recording), set `metrics/enabled` to `false` (or `TENT_METRICS_ENABLED=false`). Note that "log-only mode" — where periodic summaries are still logged but `/metrics` is unavailable — only occurs when metrics are enabled but the HTTP port cannot be bound (e.g. port conflict); it is not triggered by `metrics/enabled=false`.
 
 ### Validation
 
--- a/docs/source/design/tent/metrics.md
+++ b/docs/source/design/tent/metrics.md
@@ -65,9 +65,7 @@ TENT metrics configuration is integrated into the main `transfer-engine.json` co
     "http_port": 9100,
     "http_host": "0.0.0.0",
     "http_server_threads": 2,
-    "report_interval_seconds": 30,
-    "enable_prometheus": true,
-    "enable_json": true
+    "report_interval_seconds": 30
   },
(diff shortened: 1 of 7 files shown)


alogfans (approved):
LGTM

PR #3027, opened 2026-07-21
[CI/Build] Format only changed C++ lines

Aionw:
## Description

The C/C++ pre-commit and CI format paths previously ran overlapping or overly broad whole-file checks:

- the local `code_format.sh` hook ignored pre-commit filenames and formatted every C/C++ file changed relative to `origin/main`
- the mirrored clang-format hook then formatted each staged C/C++ file in full
- pull-request CI checked every line of each changed C/C++ file, so unrelated pre-existing style debt could fail a focused change

This change adds a `--staged` mode and delegates staged range selection and index handling to LLVM's `git-clang-format`, with clang-format 20 supplied explicitly. It removes the duplicate whole-file clang-format hook. A strict guard still rejects any selected staged file that also has unstaged changes, including cases where no formatting rewrite would otherwise be needed.

It also adds a check-only `--changed-lines` mode for CI. The wrapper resolves the merge base explicitly, then passes that concrete commit and `HEAD` to `git-clang-format --diff`. This avoids LLVM 20's broken `--diff_from_common_commit` path, which can feed a triple-dot revision to `git diff-tree` and silently select no lines. Git pathspecs select supported extensions and exclude vendored directories without duplicating Git's changed-file discovery in the shell script. The existing default and `--all` modes retain their whole-file semantics.

## Module

- [x] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Breaking change
- [x] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
bash -n scripts/code_format.sh
pre-commit validate-config .pre-commit-config.yaml
git diff --check
pre-commit run --files .github/workflows/ci.yml CONTRIBUTING.md scripts/code_format.sh
```

Isolated Git fixtures were used to verify that:

- the LLVM 20 `--diff_from_common_commit` command incorrectly returns success with no modified files for an unformatted feature change
- the fixed wrapper checks the same feature commit with an explicit merge base, prints the expected formatting diff, and returns status 1
- the fixed wrapper returns status 0 after the feature change is formatted
- the explicit merge base remains correct when the target branch advances after the feature branch diverges
- `--changed-lines` without `--check` is rejected
- `--staged --check` reads the staged snapshot and prints the expected diff
- a selected staged file with unstaged changes fails immediately with the explicit guard
- staged apply mode formats only selected lines and treats a successful rewrite as success
- automatic discovery, explicit filenames, filenames containing spaces, subdirectory execution, and vendored-directory exclusions preserve the expected behavior

**Test results:**
- [ ] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

OpenAI Codex inspected the existing hooks and CI failure mode, implemented the staged and branch-diff workflows, replaced hand-written hunk and file discovery with LLVM's `git-clang-format` plus Git pathspecs, retained the strict mixed-state guard, fixed LLVM 20 changed-line selection by resolving the merge base explicitly, updated contributor documentation, and ran the validation described above. The human submitter should review every changed line and be prepared to defend the change end-to-end.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -852,7 +852,7 @@ jobs:
           fi
         fi
         echo "Comparing against: ${BASE_REF}"
-        ./scripts/code_format.sh --check --base "${BASE_REF}"
+        ./scripts/code_format.sh --check --changed-lines --base "${BASE_REF}"
       shell: bash
 
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,7 @@
 # Pre-commit hooks configuration for Mooncake
 # Install: pip install -r requirements-dev.txt && pre-commit install
-# Run manually: pre-commit run --all-files
+# Run manually on all files: pre-commit run --all-files
+# Format all C/C++ files explicitly: ./scripts/code_format.sh --all
 # Note: clang-format should already be available (installed via system packages or dependencies.sh)
 # Exclusions: build artifacts, vendored extern code, generated wheels.
 
@@ -26,11 +27,11 @@ repos:
   - repo: local
     hooks:
       - id: mooncake-code-format
-        name: Run Mooncake code format script
-        entry: ./scripts/code_format.sh
+        name: Format staged C/C++ changes
+        entry: ./scripts/code_format.sh --staged
         language: system
-        pass_filenames: false
-        always_run: true
+        files: '\.(c|cc|cpp|cxx|cu|cuh|h|hpp)$'
+        exclude: '^(extern/|build/|.*/build/|FAST25-release/|.*cachelib_memory_allocator/|.*/thirdparty/.*)'
         require_serial: true
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
@@ -48,13 +49,6 @@ repos:
       - id: codespell
         exclude: '^(extern/|FAST25-release/)'
         args: ['--ignore-words-list=te,mooncake,KVCache,cann,hsa']
-
-  - repo: https://github.com/pre-commit/mirrors-clang-format
-    rev: v20.1.8
-    hooks:
-      - id: clang-format
-        files: '\.(c|cc|cpp|cxx|h|hpp)$'
-        exclude: '^(extern/|build/|.*/build/|FAST25-release/|.*/thirdparty/.*)'
 
   - repo: https://github.com/cheshirekow/cmake-format-precommit
     rev: v0.6.13
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -47,10 +47,9 @@ Mooncake uses [pre-commit](https://pre-commit.com/) to enforce consistent format
 | Type | Tool | Purpose |
 |------|------|---------|
 | Generic | trailing-whitespace / end-of-file-fixer | Basic hygiene |
-| Project | `./scripts/code_format.sh` | Enforce Mooncake C/C++ formatting script before commit |
+| Project | `./scripts/code_format.sh --staged` | Format staged C/C++ changes before commit |
 | Python | ruff / ruff-format | Lint + format (includes import sorting) |
 | Spelling | codespell | Catch common typos (ignores domain-specific words) |
-| C/C++ | clang-format | Apply style from the repository's `.clang-format` |
 | CMake | cmake-format | Keep build scripts readable |
 | Meta | check-yaml / check-merge-conflict / check-added-large-files | Prevent bad commits |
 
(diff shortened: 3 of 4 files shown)


Icedcoco:
I have one small concern about the direct `--staged` path.

`get_staged_line_ranges()` derives line ranges from `git diff --cached`, but `clang-format` is then applied to the working-tree file. In the normal `git commit` hook path, pre-commit stashes unstaged changes first, so this should be fine. However, if someone runs `./scripts/code_format.sh --staged --check file.cpp` directly while the same file also has unstaged insertions/deletions before the staged hunk, the cached line numbers can point at the wrong working-tree lines and the script may report `OK` without checking the actual staged change.

Would it make sense to either document `--staged` as intended only for the pre-commit hook path, or add a small guard that fails when a staged file also has unstaged changes?

Icedcoco:
Thanks for updating this. The earlier staged/unstaged line-number concern looks addressed by the strict guard and by delegating the staged snapshot handling to `git-clang-format`.

I did find one remaining issue in the new `--changed-lines` CI path that I think should be fixed before merging.

**Major: the changed-line format check can silently miss formatting errors**

The CI job now runs:

```bash
./scripts/code_format.sh --check --changed-lines --base "${BASE_REF}"
```

This eventually invokes:

```bash
git-clang-format ... \
    --diff \
    --diff_from_common_commit "${BASE_BRANCH}" HEAD \
    -- ...
```

I reproduced this in a minimal Git repository using `git-clang-format-20` and `clang-format-20`. The feature branch added the following unformatted lines:

```cpp
int touched() {
int x=1;return x;
}
```

Running the same argument combination used by the PR:

```bash
git-clang-format-20 \
    --binary clang-format-20 \
    --style file \
    --extensions cpp \
    --diff \
    --diff_from_common_commit main HEAD \
    -- '*.cpp'
```

produced:

```text
no modified files to format
```

and exited with status `0`.

As a control, running the helper without `--diff_from_common_commit`:

```bash
git-clang-format-20 \
    --binary clang-format-20 \
    --style file \
    --extensions cpp \
    --diff main HEAD \
    -- '*.cpp'
```

correctly produced:

```diff
 int touched() {
-int x=1;return x;
+  int x = 1;
+  return x;
 }
```

and exited with status `1`.

The same result is obtained by explicitly calculating the merge base and passing two concrete commits:

```bash
base_commit=$(git merge-base main HEAD)

git-clang-format-20 \
    --binary clang-format-20 \
    --style file \
    --extensions cpp \
    --diff "${base_commit}" HEAD \
    -- '*.cpp'
```

Looking at the LLVM 20 helper implementation explains the behavior: with two commits and `--diff_from_common_commit`, it combines the revisions into a single `main...HEAD` argument and passes that to `git diff-tree`. In this fixture, that command produces no patch, so `git-clang-format` sees no changed lines and returns success.

This means the current CI check can pass even when a line added by the PR is not formatted, which defeats the main purpose of the new `--changed-lines` mode.

I suggest calculating the merge base explicitly in the wrapper and then passing it and `HEAD` as two commits:

```bash
local base_commit
if ! base_commit=$(git -C "${PROJECT_ROOT}" merge-base \
    "${BASE_BRANCH}" HEAD); then
    print_error "Could not determine a merge base for '${BASE_BRANCH}' and HEAD."
    return 1
fi

command+=("${base_commit}" HEAD)
```

The explicit merge base is preferable to simply dropping `--diff_from_common_commit` and using `"${BASE_BRANCH}" HEAD`, because the target branch may have advanced since the feature branch diverged. In that case, comparing directly against the current target tip could include unrelated target-branch changes.

It would also be useful to retain this minimal fixture as a regression check, or at least rerun it against the exact wrapper command before merging. The GitHub `Check code format` job is still pending at the time of this review; even if it turns green, it would not cover this failure mode because the issue is that the checker incorrectly selects no lines.

Icedcoco:
One small, non-blocking question about the filename-less `--staged` path.

I tried staging a formatted change in `test.cpp` while keeping an unrelated unstaged change in `other.cpp`. In that state:

```bash
./scripts/code_format.sh --staged --check
```

fails with:

```text
Cannot process staged lines because a selected file has unstaged changes.
```

However, explicitly selecting the staged file succeeds:

```bash
./scripts/code_format.sh --staged --check test.cpp
```

It looks like the default wildcard pathspec makes the guard inspect every C/C++ file, including files with no staged changes. This should not affect the normal pre-commit path, since pre-commit passes filenames and handles unstaged changes, so I do not consider it blocking.

Do you think the filename-less form should limit the guard to files that actually have staged changes, or would you prefer to document that this form expects a clean C/C++ working tree?

ykwd (approved):

PR #3042, opened 2026-07-22
[store] support TP reshard across split dimensions

zxpdemonio:
## Description

Support heterogeneous TP resharding for model weight reuse when the stored weights and the target runtime use different TP split dimensions or TP sizes.

This allows one stored copy of model weights to be reused by inference or training runtimes with different TP layouts, without regenerating or storing another checkpoint layout.

This change:

- Generalizes the regular tensor formula plan with source and target split dimensions.
- Adds target TP rank and shard count to the formula.
- Expands the formula into source/destination ranges at transfer execution time.
- Transfers weight fragments directly into the target tensor buffer without an intermediate staging memcpy.
- Preserves batched range transfers and the existing query-result cache path.
- Supports TP=1 and plans containing more than 65,536 ranges.
- Refreshes an expired read lease while executing a large range list.
- Prevents manifest-backed weights from falling back to the legacy TP layout.

Only `source_split_dim != target_split_dim` uses the new formula path. Same-split-dimension heterogeneous TP remains on the existing reconstruction path.

The scope is limited to cross-split-dimension TP weight resharding and the lease renewal required by large range lists. It does not introduce a generic reshard planner, Transfer Engine changes, or global caching.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

Coverage includes 2D/3D/4D weights, different source and target split dimensions, heterogeneous TP sizes, all target ranks, TP=1, batched reads, DP/PP/EP metadata combinations, fail-closed invalid target layouts, manifest-versus-legacy routing, and a case producing 66,560 transfer ranges.

**Test commands:**
```bash
cmake --build build --target store -j8
python3 -m py_compile scripts/test_tensor_api.py
python3 scripts/test_tensor_api.py --mode func
git diff --check HEAD^
```

**Test results:**
- [ ] Unit tests pass
- [x] Integration tests pass (52 tests)
- [x] Manual testing done (66,560 ranges with a forced 10 ms lease TTL)

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

OpenAI Codex assisted with implementation, simplification, review, and test execution. I reviewed the final code and am responsible for the change.

### DIFF
--- a/mooncake-integration/store/store_py_internal.h
+++ b/mooncake-integration/store/store_py_internal.h
@@ -90,12 +90,14 @@ struct PlannedQueryResult {
     std::optional<CachedQueryResultResponse> cached_query_result;
 };
 
-struct TensorIntoRegularFullFormulaPlan {
+struct TensorIntoRegularFormulaPlan {
     std::vector<std::string> read_keys;
     std::vector<int64_t> global_shape;
-    int split_dim{0};
+    int source_split_dim{0};
+    int target_split_dim{0};
+    int target_rank{0};
+    int target_shard_count{1};
     size_t element_size{0};
-    size_t data_offset{sizeof(TensorMetadata)};
 };
 
 struct TensorIntoPlan {
@@ -103,8 +105,9 @@ struct TensorIntoPlan {
     uintptr_t registered_buffer_ptr{0};
     size_t registered_buffer_size{0};
     size_t total_length{0};
+    bool allow_generic_fallback{false};
     std::vector<TensorIntoFragment> fragments;
-    std::optional<TensorIntoRegularFullFormulaPlan> regular_full_formula;
+    std::optional<TensorIntoRegularFormulaPlan> regular_formula;
     std::vector<PlannedQueryResult> query_results;
     std::optional<TensorMetadata> materialized_metadata;
 };
--- a/mooncake-integration/store/store_py_parallel_read.h
+++ b/mooncake-integration/store/store_py_parallel_read.h
@@ -1023,7 +1023,7 @@ load_parallelism_full_reconstruction_sources(
     return reconstruction;
 }
 
-bool validate_regular_full_formula_sources(
+bool validate_regular_formula_sources(
     const std::vector<ReconstructedShardSource> &sources,
     const std::vector<int64_t> &global_shape, int split_dim, int32_t dtype,
     int shard_count, const std::string &context) {
@@ -1054,36 +1054,55 @@ bool validate_regular_full_formula_sources(
     return true;
 }
 
-std::optional<TensorIntoPlan> build_parallelism_full_tensor_into_formula_plan(
+std::optional<TensorIntoPlan>
+build_parallelism_regular_tensor_into_formula_plan(
     const std::string &key, uintptr_t buffer_ptr, size_t size,
-    const TensorParallelismSpec &parallelism, const std::string &context) {
+    const TensorParallelismSpec &parallelism, const std::string &context,
+    bool target_shard,
+    std::optional<ParsedParallelismManifest> manifest = std::nullopt) {
     const ParallelAxisSpec *request_tp_axis =
         find_axis_spec_by_kind(parallelism, LayoutAxisKind::TP);
     if (!request_tp_axis) {
         return std::nullopt;
     }
 
-    auto manifest = load_parallelism_manifest(key, context);
     if (!manifest.has_value()) {
+        manifest = load_parallelism_manifest(key, context);
+        if (!manifest.has_value()) {
+            return std::nullopt;
+        }
+    }
+    const auto &resolved_manifest = *manifest;
+    const auto &global_shape = resolved_manifest.global_shape;
+    const int source_split_dim = resolved_manifest.manifest.header.split_dim;
+    const int source_shard_count =
+        resolved_manifest.manifest.header.shard_count;
+    if (source_split_dim < 0 ||
+        source_split_dim >= static_cast<int>(global_shape.size()) ||
+        source_shard_count <= 0) {
         return std::nullopt;
     }
-    const auto &global_shape = manifest->global_shape;
-    const int split_dim = manifest->manifest.header.split_dim;
-    const int shard_count = manifest->manifest.header.shard_count;
-    if (split_dim < 0 || split_dim >= static_cast<int>(global_shape.size()) ||
-        shard_count <= 0) {
+    const int target_split_dim =
+        request_tp_axis->split_dim.value_or(source_split_dim);
+    if (target_split_dim < 0 ||
+        target_split_dim >= static_cast<int>(global_shape.size())) {
         return std::nullopt;
     }
-    if (request_tp_axis->split_dim.has_value() &&
-        request_tp_axis->split_dim.value() != split_dim) {
+    if (target_shard && target_split_dim == source_split_dim) {
         return std::nullopt;
     }
-    if (!is_uniform_shardable_dim(global_shape[split_dim],
-                                  request_tp_axis->size) ||
-        !is_uniform_shardable_dim(global_shape[split_dim], shard_count)) {
+    if (!target_shard && target_split_dim != source_split_dim) {
         return std::nullopt;
     }
-    auto element_size = TensorDtypeElementSize(manifest->manifest.header.dtype);
+    if (!is_uniform_shardable_dim(global_shape[source_split_dim],
+                                  source_shard_count) ||
+        !is_uniform_shardable_dim(
+            global_shape[target_shard ? target_split_dim : source_split_dim],
+            request_tp_axis->size)) {
+        return std::nullopt;
+    }
+    auto element_size =
+        TensorDtypeElementSize(resolved_manifest.manifest.header.dtype);
     if (!element_size.has_value()) {
         return std::nullopt;
     }
@@ -1097,21 +1116,37 @@ std::optional<TensorIntoPlan> build_parallelism_full_tensor_into_formula_plan(
         return std::nullopt;
     }
 
-    size_t tensor_numel = 1;
-    for (auto dim : global_shape) {
-        if (dim < 0) {
-            return std::nullopt;
-        }
-        tensor_numel *= static_cast<size_t>(dim);
+    std::vector<int64_t> target_shape = global_shape;
+    if (target_shard) {
+        target_shape[target_split_dim] /= request_tp_axis->size;
     }
-    const size_t tensor_bytes = tensor_numel * *element_size;
-    const size_t total_length = sizeof(TensorMetadata) + tensor_bytes;
+    std::optional<TensorMetadata> target_metadata;
+    if (target_shard) {
+        target_metadata = build_shard_metadata_from_shapes(
+            resolved_manifest.manifest.header.dtype, global_shape, target_shape,
+            parallelism.axes, 0);
+    } else {
+        target_metadata = BuildTensorMetadata(
+            resolved_manifest.manifest.header.dtype, global_shape, global_shape,
+            TensorLayoutKind::FULL);
+    }
+    if (!target_metadata.has_value()) {
+        return std::nullopt;
+    }
+    auto tensor_bytes = TensorMetadataExpectedDataBytes(*target_metadata);
+    if (!tensor_bytes.has_value() ||
+        *tensor_bytes >
+            std::numeric_limits<size_t>::max() - sizeof(TensorMetadata)) {
+        return std::nullopt;
+    }
+    const size_t total_length = sizeof(TensorMetadata) + *tensor_bytes;
 
     auto region = resolve_writable_buffer_region(buffer_ptr, size, context);
     if (!region.has_value()) {
         return std::nullopt;
     }
-    if (total_length > size || region->offset + total_length > region->size) {
+    if (total_length > size || region->offset > region->size ||
+        total_length > region->size - region->offset) {
         LOG(ERROR) << context << ": buffer too small for reconstructed tensor";
         return std::nullopt;
     }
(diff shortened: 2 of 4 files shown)


scripts/test_tensor_api.py
```diff
@@ -1429,6 +1429,105 @@ def test_27b_unified_parallelism_full_into_formula_all_split_dims_and_hetero_tp(
                     finally:
                         self.assertEqual(self.store.unregister_buffer(buffer_ptr), 0)
 
+    def test_27c_unified_parallelism_cross_split_dim_tp_reshard(self):
+        require_unified_parallelism_api(self)
+        cases = [
+            ((8, 12), 0, 2, 1, 3),
+            ((4, 6, 8), 0, 2, 2, 4),
```
zxpdemonio, line 1441: **blocker (CI):** this case is currently failing on `test-wheel-ubuntu`:

- subTest `case=((65, 1024, 2), 1, 2, 2, 2), target_rank={0,1}`
- repeated `lease_expired_before_data_transfer_completed key=func_cross_dim_tp_5_tp_0`
- assertion: `unexpectedly None : missing tensor for cross split-dim TP reshard`

Range math for this shape:
- source `(65,512,2)`, target `(65,1024,1)`, `contiguous_dim=2`
- **`range_count=33280`, `range_elements=1` → 66560 total 4-byte ranges**

So this is an extremely fragmented transfer, not just "slightly above uint16". Please either:
1. make the transfer path survive this (chunked submits / lease renew), or
2. replace with a >65k-range case that still has larger `range_bytes`, or defer the claim until fixed.

Ref: https://github.com/kvcache-ai/Mooncake/actions/runs/29898119316

mooncake-integration/store/store_py_parallel_read.h
```diff
@@ -1282,6 +1349,81 @@ pybind11::object get_tensor_with_parallelism_shard_full_materialized(
...
+            }
+            dst_offsets[source_rank].push_back(
+                base_dst_offset + sizeof(TensorMetadata) +
+                target_offset * formula.element_size);
+            src_offsets[source_rank].push_back(
+                sizeof(TensorMetadata) + source_offset * formula.element_size);
+            sizes[source_rank].push_back(range_bytes);
+        }
+    }
+    return true;
```
zxpdemonio, line 1445: The expander itself looks directionally correct for cross-dim intersections, but two robustness gaps:

1. **No guard on `target_rank < target_count`**, negative extents, or overflow while multiplying into `range_count` / offsets.
2. **`source_count` comes from `read_keys.size()`** while per-dim sizes divide `global_shape[source_dim]` by that count. If keys and manifest `shard_count` ever diverge, this silently mis-shards.

Also, when `max(source_dim, target_dim)` leaves `intersection_shape[contiguous_dim:] == 1`, you get one-element ranges (exactly the CI-failing case). Worth documenting that this layout is legal but operationally expensive, and/or capping / chunking range submission at the execute layer.

mooncake-integration/store/store_py_parallel_read.h
```diff
@@ -1054,36 +1054,55 @@ bool validate_regular_full_formula_sources(
...
+        request_tp_axis->split_dim.value_or(source_split_dim);
+    if (target_split_dim < 0 ||
+        target_split_dim >= static_cast<int>(global_shape.size())) {
         return std::nullopt;
     }
-    if (request_tp_axis->split_dim.has_value() &&
-        request_tp_axis->split_dim.value() != split_dim) {
+    if (target_shard && target_split_dim == source_split_dim) {
         return std::nullopt;
     }
```
zxpdemonio, line 1093: Same-dim shard requests intentionally bypass the new formula (`nullopt` → old reconstruction). That matches the PR's "cross-split-dimension only" scope, but please make it explicit in the PR description: **same-dim hetero TP is unchanged**, and only `source_split_dim != target_split_dim` uses `append_regular_formula_ranges`.

Also, for cross-dim, if this function returns `nullopt` after `get_tensor_with_parallelism_shard_full_materialized` already sized the buffer with `target_split_dim`, the fallback in `build_parallelism_shard_tensor_into_plan` still builds a plan on `reconstruction->split_dim` (source). Prefer failing closed for cross-dim instead of mixing dims.

mooncake-integration/store/store_py_parallel_read.h
```diff
@@ -1121,25 +1156,29 @@ std::optional<TensorIntoPlan> build_parallelism_full_tensor_into_formula_plan(
     plan.registered_buffer_ptr = reinterpret_cast<uintptr_t>(region->base);
     plan.registered_buffer_size = region->size;
     plan.total_length = total_length;
-    plan.materialized_metadata =
-        BuildTensorMetadata(manifest->manifest.header.dtype, global_shape,
-                            global_shape, TensorLayoutKind::FULL);
-    plan.materialized_metadata->header.data_bytes = tensor_bytes;
+    plan.allow_generic_fallback = !target_shard;
```
zxpdemonio, line 1159: `allow_generic_fallback = !target_shard` means cross-dim shard formula execution failures soft-fail to `None` with no generic retry. That is reasonable given generic reconstruction still errors on split_dim mismatch, but then formula correctness + transfer reliability become hard requirements — another reason the 66k-range lease failure is a merge blocker.

mooncake-integration/store/store_py_parallel_read.h
```diff
@@ -1180,7 +1220,14 @@ std::optional<TensorIntoPlan> build_parallelism_full_tensor_into_plan(
...
+    std::optional<ParsedParallelismManifest> manifest = std::nullopt) {
+    if (auto formula_plan = build_parallelism_regular_tensor_into_formula_plan(
+            key, buffer_ptr, size, parallelism, context, true,
+            std::move(manifest));
+        formula_plan.has_value()) {
+        return formula_plan;
+    }
     auto reconstruction =
         load_parallelism_full_reconstruction_sources(key, parallelism, context);
     if (!reconstruction.has_value()) {
```
zxpdemonio, line 1247: Fallback still shards along `reconstruction->split_dim` (source), using `tp_axis->rank/size` but **ignoring a divergent request `split_dim`**. Safe today only because formula handles cross-dim and same-dim hits this path. If formula build fails for a cross-dim request, this becomes incorrect relative to the buffer sized in `get_tensor_with_parallelism_shard_full_materialized`. Consider returning `nullopt` when request `split_dim != source split_dim` instead of attempting this fallback.

mooncake-integration/store/store_py_parallel_read.h
```diff
@@ -1587,7 +1689,9 @@ pybind11::object get_tensor_with_parallelism(
                 return reconstructed;
             }
         }
-        if (uses_legacy_tp_storage_key(*parallelism)) {
+        if (uses_legacy_tp_storage_key(*parallelism) &&
+            !load_parallelism_manifest(key, "get_tensor_with_parallelism")
+                 .has_value()) {
```
zxpdemonio, line 1715: Good guard: manifest-backed weights no longer fall into legacy TP key reads. Worth a focused regression test (manifest present + legacy-looking parallelism → must not call `resolve_tp_read_key`).

mooncake-integration/store/store_py_internal.h
```diff
@@ -90,21 +90,24 @@ struct PlannedQueryResult {
...
 
 struct TensorIntoPlan {
     uintptr_t user_buffer_ptr{0};
     uintptr_t registered_buffer_ptr{0};
     size_t registered_buffer_size{0};
     size_t total_length{0};
+    bool allow_generic_fallback{false};
     std::vector<TensorIntoFragment> fragments;
-    std::optional<TensorIntoRegularFullFormulaPlan> regular_full_formula;
+    std::optional<TensorIntoRegularFormulaPlan> regular_formula;
```
zxpdemonio, line 110: Rename `regular_full_formula` → `regular_formula` / `TensorIntoRegularFormulaPlan` is clear. Optional: document in a one-line comment that `target_shard_count == 1` means "full gather" (defaults), so readers do not wonder why full plans omit rank fields.

zxpdemonio (commented):
## Review summary

Nice generalization of the regular-tensor formula path to cross-split-dim TP resharding, and the test matrix (2D/3D/4D, multi-axis, batch into) is mostly on point. The core range-expansion idea looks sound for the common cases.

However, **CI is currently red**, and the failure is exactly the ">65536 ranges" scenario this PR claims to support. Please fix that before merge.

> Note: reviewed with `gh` as `zxpdemonio` (PR author). GitHub disallows `REQUEST_CHANGES` on own PRs, so this is a `COMMENT` review; findings are still labeled blocker / should-fix.

### Must fix
1. **`test_27c` fails in CI** on `case=((65, 1024, 2), 1, 2, 2, 2)` for both target ranks. Log shows repeated `lease_expired_before_data_transfer_completed` on `func_cross_dim_tp_5_tp_0`, then `get_tensor_with_parallelism` returns `None`. This case expands to **66,560 ranges of 4 bytes each** (`range_count=33280` per source shard, `range_elements=1`). The PR description claims support for plans with >65,536 ranges, but CI does not pass that claim.
2. Either make large fragmented transfers robust (chunk + lease renew / longer lease / coalesce), or drop/replace this case until the store path can finish it. Right now the feature claim in the description is ahead of the implementation.

### Should fix
3. Cross-dim shard plans set `allow_generic_fallback=false`, and the generic reconstruction path still rejects `split_dim` mismatch. That is consistent with "formula or fail", but `get_tensor_with_parallelism_shard_full_materialized` sizes the buffer using **target** `split_dim` and then may fall through to a source-dim reconstruction plan if formula build fails — layout/size mismatch risk. Prefer failing closed, or sizing/building fallback with the same dim.
4. `append_regular_formula_ranges` should validate `target_rank < target_shard_count`, uniform divisibility, and overflow on `range_count` / offset arithmetic instead of assuming the planner already checked everything.
5. Formula intentionally skips same-dim shard (`target_split_dim == source_split_dim` → `nullopt`). Please call out in the PR body that same-dim hetero TP still uses the old reconstruction path; only cross-dim hits the new formula.
6. Checklist / CI: `test-wheel-ubuntu` and `CI Gate` are failing; format/pre-commit boxes are unchecked. Green CI should be required for this path-heavy change.

### Nits
7. `source_count` is taken from `read_keys.size()` while shape math assumes it equals manifest `shard_count` — assert they match.
8. Consider a unit-level test for `append_regular_formula_ranges` (offset golden values) so algorithm regressions are caught without full store + lease.

scripts/test_tensor_api.py
```diff
@@ -1429,6 +1429,105 @@ def test_27b_unified_parallelism_full_into_formula_all_split_dims_and_hetero_tp(
...
+        for key, (tensor, (source_dim, source_tp), _) in zip(keys, cases):
+            self.assertEqual(
+                put_uniform_full_tensor_with_unified_tp(
+                    self.store, key, tensor, source_tp, source_dim
+                ),
+                0,
+            )
+        targets = [
+            make_read_target("shard", build_tp_parallelism(tp, dim, rank))
+            for _, _, (dim, tp, rank) in cases
```
zxpdemonio, line 1498: Good coverage for batch `*_into` and multi-axis modes. After fixing `test_27c`, consider adding one golden-offset style check (even via a small helper) for a tiny cross-dim case so algorithm bugs are not only caught via full RDMA roundtrips.

mooncake-integration/store/store_py_parallel_read.h
zxpdemonio: Fixed in d6a40e22. Cross-split-dimension requests now fail closed when formula planning fails and can no longer enter the source-dimension generic reconstruction path. The PR description now explicitly states that same-dimension heterogeneous TP continues to use the existing reconstruction path.

mooncake-integration/store/store_py_internal.h
zxpdemonio: Acknowledged. I left this as-is because the defaults are only populated by the full-gather planner and the field names already make the invariant local; adding a comment would duplicate the construction logic.

scripts/test_tensor_api.py
zxpdemonio: I kept the helper private and did not add a testing-only binding. The existing small 2D/3D/4D round trips validate exact output placement, and this update adds a focused fail-closed layout regression. A direct golden-offset test would require exposing implementation detail beyond this PR's scope.

scripts/test_tensor_api.py
zxpdemonio: Fixed in d6a40e22. get_into_ranges_internal now refreshes the query result and retries the current range on LEASE_EXPIRED, without imposing a total range limit. I verified the 66,560-range case with a forced 10 ms lease TTL and reran all 52 functional/type tests.

mooncake-integration/store/store_py_parallel_read.h
zxpdemonio: Addressed the actionable guards: the expander now rejects an invalid target rank/count, negative extents, and non-uniform source/target dimensions. read_keys is built directly from the manifest source_shard_count, and validate_regular_formula_sources verifies the returned source count before the formula is stored, so there is no independent key-count input to reconcile. Overflow remains covered by TensorMetadataExpectedDataBytes during plan construction: range counts, strides, and offsets are sub-products of that already validated tensor extent. Large fragmented execution is handled by lease refresh in get_into_ranges_internal.

mooncake-integration/store/store_py_parallel_read.h
zxpdemonio: Resolved by the lease refresh/retry in get_into_ranges_internal. Cross-dimension formula execution remains formula-or-fail, but large range lists no longer fail merely because the original query lease expires.

mooncake-integration/store/store_py_parallel_read.h
zxpdemonio: Fixed in d6a40e22. The shard planner records whether the request is cross-dimension before attempting the formula and returns nullopt if that formula cannot be built. I added a regression where the target dimension is not divisible by target TP while the source dimension is; the old fallback could return a source-dimension shard, while the new path correctly returns None.

mooncake-integration/store/store_py_parallel_read.h
zxpdemonio: Added a focused regression in test_13_unified_parallelism_tp_matches_wrapper: a legacy-looking shard key is populated with poison data while a manifest exists. The legacy wrapper can read it, but the unified manifest-backed read must still return None.

zxpdemonio:
Vincent will cover this in https://github.com/kvcache-ai/Mooncake/issues/3111

PR #3050, opened 2026-07-22
[TransferEngine] Harden socket handshake frame handling

alogfans:
## Description

Harden the socket handshake frame parser so it rejects invalid-length frames and common non-Mooncake probes before attempting JSON parsing. This avoids listener-side stalls or noisy malformed JSON handling when the handshake port receives incomplete traffic, HTTP probes, TLS probes, or oversized/zero-length frames.

This keeps the existing handshake wire format unchanged for compatibility with versions that have not enabled this patch.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
```

**Test results:**
- [X] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used
- [ ] AI tools were used (specify below) CodeX with manual recheck

### DIFF
--- a/mooncake-transfer-engine/include/common.h
+++ b/mooncake-transfer-engine/include/common.h
@@ -369,9 +369,13 @@ static inline ssize_t readFully(int fd, void *buf, size_t len) {
     size_t nbytes = len;
     while (nbytes && std::chrono::steady_clock::now() < deadline) {
         ssize_t rc = read(fd, pos, nbytes);
-        if (rc < 0 && (errno == EAGAIN || errno == EINTR))
+        if (rc < 0 && errno == EINTR)
             continue;
-        else if (rc < 0) {
+        else if (rc < 0 && (errno == EAGAIN || errno == EWOULDBLOCK)) {
+            LOG(WARNING) << "Socket read timed out: expected " << len
+                         << " bytes, actual " << len - nbytes << " bytes";
+            return len - nbytes;
+        } else if (rc < 0) {
             PLOG(ERROR) << "Socket read failed";
             return rc;
         } else if (rc == 0) {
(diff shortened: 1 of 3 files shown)


mooncake-transfer-engine/include/common.h
```diff
@@ -443,13 +449,27 @@ static inline size_t getHandshakeMaxLength() {
...
     const size_t kMaxLength = getHandshakeMaxLength();
     uint64_t length = 0;
     ssize_t n = readFully(fd, &length, sizeof(length));
     if (n != (ssize_t)sizeof(length)) {
-        LOG(ERROR) << "readString: failed to read length, got: " << n;
+        LOG(WARNING) << "readString: incomplete handshake length, got: " << n;
+        return {type, ""};
+    }
+
+    const char *prefix = reinterpret_cast<const char *>(&length);
```
staryxchen: Could you clarify the intended invariant behind inspecting the in-memory bytes of `length`?

alogfans: We inspect the first 8 wire bytes only to identify and reject HTTP/TLS probes before interpreting them as a Mooncake length prefix.

staryxchen: Thanks for clarifying. One concern is that the current wire format serializes `length` as a native-endian `uint64_t`. On little-endian systems, a valid typed frame with `length == 790` is encoded as `0x16 0x03 0x00 ...`, which matches the TLS heuristic and is rejected before its payload is read. This corresponds to a valid 789-byte payload plus the one-byte request type.

More generally, could we rely on the zero/maximum-length validation and the bounded receive timeout instead? They already reject HTTP/TLS or incomplete probes without adding prefix heuristics that can collide with valid length encodings. If distinguishing protocols is required, a framing-level magic/version field would be unambiguous, though that would require a wire-format change.

alogfans: Good catch, thanks. You're right that inspecting the native-endian length bytes can collide with valid frame lengths. I removed the HTTP/TLS prefix heuristic and now rely on the existing zero/max-length validation plus bounded reads/timeouts for incomplete probes.

I also added a regression test for a valid typed frame with length 790, which is encoded as 0x16 0x03 ... on little-endian systems, to make sure it is accepted.

staryxchen (approved):
LGTM

PR #3055, opened 2026-07-22
[wheel] improve structured non-tensor codecs

zxpdemonio:
## Description

This PR is split out from the larger rollout data transfer optimization PR (#2671).

It improves structured non-tensor codec handling in `mooncake-wheel` for rollout data transfer. The target scenario is not raw transfer bandwidth improvement, but representing real RL rollout non-tensor fields without falling back to pickling the whole object.

The changes focus on the codec layer used by structured rollout objects:

- Improve codec selection and decode handling for structured non-tensor fields.
- Support real rollout field shapes such as typed ragged ndarray rows, ragged tensor dictionaries, msgpack-able lists/dicts, scalar ndarray fields, and nested optional values.
- Preserve DataProto/dict transfer semantics through the existing structured object put/get path.
- Fix ragged tensor codec decode metadata compatibility found during target-scenario validation.

This is intended as the codec foundation for ROLL, Slime, and Miles rollout payloads. Performance-oriented copy hot-path changes are left to a follow-up PR.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
PYTHONPATH=mooncake-wheel python3 -m pytest mooncake-wheel/tests/test_structured_object_store.py -k "ragged or schema or dataproto_helper_object_non_tensor or flat_dict" -q
python3 -m py_compile mooncake-wheel/mooncake/structured_object_store.py
git diff --check
```

Target-scenario validation was also run with a real Miles rollout payload through the unified `put/get(type="dict")` API on a cross-machine RDMA setup.

**Test results:**
- [x] Unit tests pass
- [x] Manual testing done

Result:
`8 passed, 2 skipped, 90 deselected`

Additional checks:
- `py_compile`: passed
- `git diff --check`: passed
- Real Miles unified dict transfer scenario: passed

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex was used to split this patch from the larger rollout transfer optimization branch, validate the target rollout scenarios, squash the branch into one commit, and prepare this PR description.

### DIFF
--- a/mooncake-wheel/mooncake/structured_object_store.py
+++ b/mooncake-wheel/mooncake/structured_object_store.py
@@ -26,6 +26,13 @@ MISSING_OBJECT_ERROR = (
     -704
 )  # Mooncake remove returns -704 for an already-missing object.
 STRUCTURED_FIELD_SPECS_KEY = "__mooncake_structured_fields__"
+_ENCODING_FALLBACK_ERRORS = (
+    TypeError,
+    ValueError,
+    OverflowError,
+    RuntimeError,
+    RecursionError,
+)
 
 
 class BundleStore(Protocol):
@@ -2462,11 +2469,6 @@ def _decode_structured_recursive_field(
         values = _decode_structured_leaf(
             leaf["codec"], leaf_payload, rows, leaf.get("metadata")
         )
-        if leaf["codec"] == "typed_ragged":
-            values = [
-                value.tolist() if isinstance(value, np.ndarray) else value
-                for value in values
-            ]
         missing = payload[leaf_payload_members["missing"]]
         leaf_values[leaf["path"]] = [
             MISSING if bool(missing[row]) else values[row] for row in range(rows)
@@ -2658,8 +2660,10 @@ def _normalize_ragged_tensor_dict_keys(keys: Any) -> list[str]:
     for key in iterable:
         if not isinstance(key, str):
             raise TypeError("ragged_tensor_dict keys must be strings")
-        if "." in key:
-            raise ValueError("ragged_tensor_dict keys must not contain '.'")
+        if not key:
+            raise ValueError("ragged_tensor_dict keys must not be empty")
+        if any(separator in key for separator in (".", "/", "\\")):
+            raise ValueError("ragged_tensor_dict keys must not contain '.', '/', or '\\'")
         if key in seen:
             raise ValueError(f"ragged_tensor_dict keys contain duplicate key {key!r}")
         seen.add(key)
@@ -5309,54 +5313,26 @@ def _encode_msgpack_ragged_values(
 
 def _decode_msgpack_ragged_values(payload: dict[str, Any], rows: int) -> list[Any]:
     data = payload["data"]
+    offsets = payload["offsets"]
     nulls = payload["nulls"]
-    has_nulls = bool(nulls.any()) if rows > 0 else False
+    if len(offsets) != rows + 1:
+        raise ValueError(
+            f"msgpack_ragged offsets length {len(offsets)} does not match rows {rows}"
+        )
+    if len(nulls) != rows:
+        raise ValueError(
+            f"msgpack_ragged nulls length {len(nulls)} does not match rows {rows}"
+        )
     raw_data = bytes(data) if not isinstance(data, bytes) else data
-    if not has_nulls:
-        unpacker = _msgpack.Unpacker(raw=False)
-        unpacker.feed(raw_data)
-        return list(unpacker)
-    # With nulls: stream decode non-null values, insert None at null positions
-    unpacker = _msgpack.Unpacker(raw=False)
-    unpacker.feed(raw_data)
-    non_null_iter = iter(unpacker)
     values = []
-    for is_null in nulls:
+    for row, is_null in enumerate(nulls):
         if bool(is_null):
             values.append(None)
         else:
-            values.append(next(non_null_iter))
+            begin = int(offsets[row])
+            end = int(offsets[row + 1])
+            values.append(_msgpack.unpackb(raw_data[begin:end], raw=False))
     return values
-
-
-
-def _json_default(value: Any) -> Any:
-    if isinstance(value, np.generic):
-        return value.item()
-    if hasattr(value, "__int__"):
-        return int(value)
-    raise TypeError(f"Object of type {type(value).__name__} is not JSON serializable")
-
-
-
-def _parse_torch_dtype(dtype_str: str) -> Any:
-    """Parse torch dtype string (e.g. 'torch.float32') to torch.dtype object."""
-    name = dtype_str.removeprefix("torch.")
-    dtype = getattr(_torch, name, None)
-    if dtype is None or not isinstance(dtype, _torch.dtype):
-        raise ValueError(f"unknown torch dtype: {dtype_str}")
-    return dtype
-
-
-
-def _torch_dtype_to_numpy(dtype_str: str) -> np.dtype:
-    """Convert torch dtype string to numpy dtype for buffer interpretation."""
-    torch_dtype = _parse_torch_dtype(dtype_str)
-    # bfloat16 has no numpy equivalent; use float16 (same element size)
-    if torch_dtype == _torch.bfloat16:
-        return np.dtype(np.float16)
-    return _torch.empty(0, dtype=torch_dtype).numpy().dtype
-
 
 
 class _OwnerBackedList(list):
@@ -5930,6 +5906,14 @@ def _choose_leaf_codec(values: list[Any]) -> _CodecDecision:
     if not nn:
         return _CodecDecision(False, "msgpack_ragged", "all rows are null", "msgpack")
     if _torch is not None and all(isinstance(value, _torch.Tensor) for value in nn):
+        dtypes = {value.dtype for value in nn}
+        if len(dtypes) != 1:
+            return _CodecDecision(
+                False,
+                "ragged_tensor",
+                f"mixed tensor dtype: {sorted(str(dtype) for dtype in dtypes)}",
+                "torch.Tensor",
+            )
         dtype = str(nn[0].dtype)
         return _CodecDecision(True, "ragged_tensor", "all non-null rows are Tensor", "torch.Tensor", {"dtype": dtype})
     if all(_is_media_list(value) for value in nn):
--- a/mooncake-wheel/mooncake/structured_object_store.py
+++ b/mooncake-wheel/mooncake/structured_object_store.py
@@ -815,9 +815,7 @@ class MooncakeBundleTransfer:
                 values = _decode_structured_non_tensor_encoded(
                     encoded, payload, output_rows, metadata
                 )
-            array = np.empty(output_rows, dtype=object)
-            array[:] = values
-            non_tensor_batch[name] = array
+            non_tensor_batch[name] = _object_array_from_decoded_values(values)
         meta_info = _select_mapping(ref.meta_info, meta_info_keys)
         return _build_dataproto_like_result(
             batch, non_tensor_batch, meta_info, data_cls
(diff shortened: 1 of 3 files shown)


mooncake-wheel/mooncake/structured_object_store.py
```diff
@@ -5101,6 +5129,271 @@ def _tensor_payload_parts(value: _TensorPayload) -> tuple[bytes, int, int, Any]:
...
+    for codec in ("msgpack_ragged", "json_ragged"):
+        if codec not in codecs:
+            codecs.append(codec)
+    for codec in codecs:
+        metadata = dict(decision.metadata or {}) if codec == decision.codec else {}
+        codec_values = _normalize_values_for_fallback_codec(values, codec)
+        codec_array = np.asarray(codec_values, dtype=object)
+        try:
+            return _encode_with_schema(path, codec_array, FieldSchema(codec=codec, metadata=metadata))
+        except _ENCODING_FALLBACK_ERRORS as error:
```
zxpdemonio, line 5249: **blocker:** `_ENCODING_FALLBACK_ERRORS` is not defined anywhere in the repo. Once any codec encode fails, Python raises `NameError` while evaluating the `except` type, and the fallback chain stops.

Repro:
```python
_encode_with_fallback('t', np.asarray([object(), object()], dtype=object))
# NameError: name '_ENCODING_FALLBACK_ERRORS' is not defined
```

Please add something like:
```python
_ENCODING_FALLBACK_ERRORS = (TypeError, ValueError, OverflowError, RuntimeError)
```
and a unit test that the first codec failure continues to msgpack/json and finally raises an aggregated `ValueError`.

mooncake-wheel/mooncake/structured_object_store.py
```diff
@@ -2698,68 +2712,80 @@ def _encode_ragged_tensor_dict_values(
...
-        sub_payload, sub_metadata = _encode_ragged_tensor_values(
-            [None if item is None else item.get(key) for item in values]
-        )
-        payload.update(
-            {f"{key}.{name}": value for name, value in sub_payload.items()}
-        )
+        try:
+            key_values = [None if v is None else v.get(key) for v in values]
+            sub_payload, sub_metadata = _encode_ragged_tensor_values(key_values)
+        except _ENCODING_FALLBACK_ERRORS as exc:
```
zxpdemonio, line 2742: Same issue: this references undefined `_ENCODING_FALLBACK_ERRORS`. The intent is to wrap per-key ragged encode failures into a clearer `ValueError`, but it currently fails with `NameError` first.

mooncake-wheel/mooncake/structured_object_store.py
```diff
@@ -5633,13 +5926,44 @@ def _can_json(values: list[Any]) -> _CodecDecision:
...
-            return decision
-    return _CodecDecision(
-        False, "pickle_ragged_fallback", "no optimized codec matched", "python object"
-    )
+    nn = _non_null(values)
+    if not nn:
+        return _CodecDecision(False, "msgpack_ragged", "all rows are null", "msgpack")
+    if _torch is not None and all(isinstance(value, _torch.Tensor) for value in nn):
+        dtype = str(nn[0].dtype)
+        return _CodecDecision(True, "ragged_tensor", "all non-null rows are Tensor", "torch.Tensor", {"dtype": dtype})
```
zxpdemonio, line 5918: `_choose_leaf_codec` now returns `accepted=True` for **any** Tensor rows (including mixed dtype) and only records `nn[0].dtype`, while `_encode_ragged_tensor_values` still `raise ValueError("mixed tensor dtype: ...")`.

Callers will think mixed-dtype is a first-class path. Prefer one of:
1. keep rejecting mixed dtype in choose / return `accepted=False` toward fallback; or
2. actually support mixed dtype in encode (per-dtype storage / explicit cast) and add encode+decode roundtrip tests.

Also note: in `_encode_with_fallback`, `np.asarray(tensors, dtype=object)` often converts `torch.Tensor` to `numpy.ndarray`, so the later choose result can diverge from this unit-test path.

mooncake-wheel/tests/test_put_get_tensor.py
```diff
@@ -286,15 +286,17 @@ def test_tensor(self):
         self.assertTrue(d.accepted)
         self.assertEqual(d.codec, "ragged_tensor")
 
-    def test_tensor_mixed_dtype_rejected(self):
+    def test_tensor_mixed_dtype(self):
         import torch
         d = _choose_leaf_codec([torch.tensor([1], dtype=torch.float32), torch.tensor([1], dtype=torch.int64)])
-        self.assertFalse(d.accepted)
+        self.assertTrue(d.accepted)
+        self.assertEqual(d.codec, "ragged_tensor")
```
zxpdemonio, line 293: After changing this from `rejected` to `accepted`, the test only checks `_choose_leaf_codec`, not `_encode_ragged_tensor_values` / `_encode_with_fallback`. With the current implementation, mixed dtype cannot actually be ragged-encoded successfully. Please add at least one encode (or put/get) assertion, otherwise this gives false confidence.

mooncake-wheel/mooncake/structured_object_store.py
```diff
@@ -5101,6 +5129,271 @@ def _tensor_payload_parts(value: _TensorPayload) -> tuple[bytes, int, int, Any]:
...
+    return values
+
+
+
+def _json_default(value: Any) -> Any:
+    if isinstance(value, np.generic):
+        return value.item()
+    if hasattr(value, "__int__"):
+        return int(value)
+    raise TypeError(f"Object of type {type(value).__name__} is not JSON serializable")
```
zxpdemonio: `_json_default` is defined but never passed to `json.dumps(..., default=_json_default)`. Either wire it into `json_ragged` encode, or delete it so nobody assumes numpy scalars / custom integer types already have a fallback.

mooncake-wheel/mooncake/structured_object_store.py
```diff
@@ -5101,6 +5129,271 @@ def _tensor_payload_parts(value: _TensorPayload) -> tuple[bytes, int, int, Any]:
...
+
+
+
+def _torch_dtype_to_numpy(dtype_str: str) -> np.dtype:
+    """Convert torch dtype string to numpy dtype for buffer interpretation."""
+    torch_dtype = _parse_torch_dtype(dtype_str)
+    # bfloat16 has no numpy equivalent; use float16 (same element size)
+    if torch_dtype == _torch.bfloat16:
+        return np.dtype(np.float16)
+    return _torch.empty(0, dtype=torch_dtype).numpy().dtype
```
zxpdemonio: `_parse_torch_dtype` / `_torch_dtype_to_numpy` are currently dead code; `_decode_ragged_tensor_values(..., metadata=...)` also accepts metadata but never uses it. If this is reserved for a follow-up (restoring dtype from metadata), please delete it in this PR or mark a TODO to keep the review surface smaller.

mooncake-wheel/mooncake/structured_object_store.py
```diff
@@ -2645,17 +2649,21 @@ def _decode_ragged_tensor_values(payload: dict[str, Any], rows: int) -> list[Any
...
+    seen: set[str] = set()
+    for key in iterable:
+        if not isinstance(key, str):
+            raise TypeError("ragged_tensor_dict keys must be strings")
+        if "." in key:
+            raise ValueError("ragged_tensor_dict keys must not contain '.'")
+        if key in seen:
+            raise ValueError(f"ragged_tensor_dict keys contain duplicate key {key!r}")
+        seen.add(key)
+        normalized.append(key)
```
zxpdemonio, line 2670: Compared with the old implementation, empty keys are no longer rejected, nor are `/` / `\\`. Decode still splits payloads with the `f"{key}."` prefix (see `_decode_ragged_tensor_dict_values`), so an empty key becomes a dangerous `.` prefix. Unless there is an explicit compatibility need, please restore empty-key / path-separator validation.

mooncake-wheel/mooncake/structured_object_store.py
```diff
@@ -5101,6 +5129,271 @@ def _tensor_payload_parts(value: _TensorPayload) -> tuple[bytes, int, int, Any]:
...
+
+def _decode_msgpack_ragged_values(payload: dict[str, Any], rows: int) -> list[Any]:
+    data = payload["data"]
+    nulls = payload["nulls"]
+    has_nulls = bool(nulls.any()) if rows > 0 else False
+    raw_data = bytes(data) if not isinstance(data, bytes) else data
+    if not has_nulls:
+        unpacker = _msgpack.Unpacker(raw=False)
+        unpacker.feed(raw_data)
+        return list(unpacker)
```
zxpdemonio: Encode writes `offsets`, but decode never uses them; the no-null path also does not check `len(values) == rows`. Please either:
- slice-decode with offsets (more robust, consistent with other ragged codecs), or
- drop `offsets` so they are not misleading;
and validate row count to avoid silent truncation / over-unpacking.

mooncake-wheel/mooncake/structured_object_store.py
```diff
@@ -2462,6 +2462,11 @@ def _decode_structured_recursive_field(
         values = _decode_structured_leaf(
             leaf["codec"], leaf_payload, rows, leaf.get("metadata")
         )
+        if leaf["codec"] == "typed_ragged":
+            values = [
+                value.tolist() if isinstance(value, np.ndarray) else value
+                for value in values
+            ]
```
zxpdemonio: nit: `_decode_typed_ragged_values` already `.tolist()`s, so this second ndarray→tolist pass is effectively a dead branch. If it exists for old-payload compatibility, add a short comment; otherwise delete it.

mooncake-wheel/mooncake/structured_object_store.py
```diff
@@ -2323,51 +2314,53 @@ def _is_recursive_encoded_non_tensor(encoded: Mapping[str, Any]) -> bool:
...
-    for token in path_tokens[len(root_tokens) :]:
+    for token in _structured_path_tokens(suffix):
+        if isinstance(current, _Missing):
+            return MISSING
         if isinstance(token, str):
-            if current is None:
-                return None
-            if isinstance(current, _Missing) or not isinstance(current, dict):
+            if not isinstance(current, dict) or token not in current:
                 return MISSING
```
zxpdemonio, line 2364: Behavior change: the old path returned `None` when walking through a `None` intermediate, now it returns `MISSING`. For optional nested fields this affects the recursive missing mask. If intentional, please call it out in the PR description or a test; otherwise this may change existing `structured_recursive` semantics.

zxpdemonio (commented):
## Review summary

The direction looks right: moving structured non-tensor fields off pickle fallback toward inferable / schema-driven codecs is a good fit for Miles/ROLL-style ragged payloads. However, this patch has a hard bug that breaks the fallback path, and codec selection is inconsistent with actual encode semantics. Please fix before merging.

### Must fix
1. **`_ENCODING_FALLBACK_ERRORS` is undefined** — the `except` clauses in `_encode_with_fallback` / ragged_tensor_dict encode turn any codec failure into a `NameError` instead of trying the next codec or raising a clear `ValueError`. Reproduced on the PR head with mixed-dtype tensors / `object()` values.
2. **`_choose_leaf_codec` returns `accepted=True` for mixed-dtype Tensors**, but `_encode_ragged_tensor_values` still rejects mixed dtypes. Also, `np.asarray(..., dtype=object)` can turn Tensors into ndarrays, so the fallback path may choose a different codec than the unit tests assert. Current tests only cover choose, not encode.

### Should fix
3. `_json_default` / `_parse_torch_dtype` / `_torch_dtype_to_numpy` are dead code; `json.dumps` never passes `default=_json_default`.
4. `_normalize_ragged_tensor_dict_keys` now allows empty keys and `/` / `\\`. Decode still splits payloads by the `{key}.` prefix — please restore stricter validation, or document why this is safe.
5. `_decode_msgpack_ragged_values` ignores the `offsets` written at encode time and does not check that decoded length == `rows`.
6. Most of the change is codec implementation, but only choose-path assertions in `test_put_get_tensor.py` were updated. Please turn the structured-object / Miles scenarios mentioned in the PR description into regression tests.

he-yufeng:
Heads-up on the overlap with #3054 (mine, restoring zero-copy puts for multi-buffer payloads after #3024): the two PRs touch different layers — this one reworks the codec encode/decode side, #3054 only changes `put_multi_buffer_payload_chunks` in the transport path — so there's no logical conflict, but both edit `structured_object_store.py` and whichever lands second will want a quick rebase.

One invariant worth keeping explicit while you're in the codec layer: ragged encoders should keep emitting `_MultiBufferPayload` (or otherwise route through `_put_multi_buffer_payload`), because that's the only path where zero-copy is reachable again after #3054. If any of the new codec branches hand the transport a plain joined bytes object, the no-concatenation win from #3024 is gone for that branch; if they produce a new payload type, it needs its own zero-copy story.

Also noticed the metadata switched from `data.dtype` to `dtype` in `_encode_typed_ragged_values` — fine if intentional (they should be the same value), just flagging in case it signals the payload shape is changing under the same key.

stmatengss (approved):

PR #3060, opened 2026-07-22
[wheel] add unified release and cleanup APIs

zxpdemonio:
## Description

Add unified structured-object lifecycle APIs to match the existing unified `put(...)` / `get(...)` surface:

- `MooncakeBundleTransfer.release(result, type=...)` dispatches lifecycle handling by payload type.
- `MooncakeBundleTransfer.cleanup(ref, type=...)` dispatches cleanup by payload type.

The type-specific paths are explicit:

- `release_dataproto(result)`
- `release_dict(result)`
- `cleanup_dataproto(ref)`
- `cleanup_dict(ref)`

Today the DataProto and flat-dict implementations share the same private lifecycle mechanics, but the public layering keeps the unified API at the outer layer and keeps DataProto/dict semantics separated. The existing `release_result(result)` helper remains available for compatibility and routes through the DataProto release path.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
PYTHONPATH=mooncake-wheel python3 -m py_compile mooncake-wheel/mooncake/structured_object_store.py mooncake-wheel/tests/test_structured_object_store.py
PYTHONPATH=mooncake-wheel python3 -m pytest -q mooncake-wheel/tests/test_structured_object_store.py -k "buffer_pool or lifecycle_split_helpers_are_available or unified_release_is_available or unified_cleanup_removes_flat_dict_objects or unified_put_rejects_unknown_type or dataproto_helper_cleanup_removes_all_stage_objects"
git diff --check
```

**Test results:**
- [x] Unit tests pass

Result:
`6 passed, 2 skipped, 94 deselected in 0.76s`

Notes:
- The two skipped tests are existing torch/CUDA environment skips in this machine.
- A full `test_structured_object_store.py` run was attempted but did not finish promptly in a non-target slow path, so the focused lifecycle/API tests above are used as the validation signal for this small Python-wheel API patch.
- `./scripts/code_format.sh` formats changed C/C++ files only; this PR changes Python files only. `git diff --check` passed.

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex was used to add the unified lifecycle dispatch APIs, preserve compatibility helpers, run targeted validation, and prepare this PR description.

### DIFF
--- a/mooncake-wheel/mooncake/structured_object_store.py
+++ b/mooncake-wheel/mooncake/structured_object_store.py
@@ -542,12 +542,49 @@ class MooncakeBundleTransfer:
         return self._structured_store.materialize_into(spec, destinations)
 
     @staticmethod
-    def release_result(result: Any) -> None:
+    def release(
+        result: Any, *, type: Literal["dataproto", "dict"] = "dataproto"
+    ) -> None:
         """Release pool-backed buffers in a GET result.
 
-        After get_dataproto / get_dict, ndarray payloads may be backed by
-        the BufferPool. Call this to release those leases deterministically
-        instead of waiting for GC ``__del__``.
+        ``type`` selects the public lifecycle path. DataProto and flat-dict
+        results currently share the same buffer-owner walker, but the split is
+        kept explicit so future type-specific lifecycle rules have a stable
+        dispatch point.
+        """
+        if type == "dataproto":
+            return MooncakeBundleTransfer.release_dataproto(result)
+        if type == "dict":
+            return MooncakeBundleTransfer.release_dict(result)
+        raise ValueError(f"unsupported Mooncake payload type: {type!r}")
+
+    @staticmethod
+    def release_dataproto(result: Any) -> None:
+        """Release pool-backed buffers in a DataProto GET result."""
+        MooncakeBundleTransfer._release_result(result)
+
+    @staticmethod
+    def release_dict(result: Any) -> None:
+        """Release pool-backed buffers in a flat-dict GET result."""
+        MooncakeBundleTransfer._release_result(result)
+
+    @staticmethod
+    def release_result(result: Any) -> None:
+        """Release pool-backed buffers in a DataProto-oriented GET result.
+
+        New code should prefer ``release(..., type=...)``. This compatibility
+        helper keeps the original DataProto-oriented name.
+        """
+        MooncakeBundleTransfer.release_dataproto(result)
+
+    @staticmethod
+    def _release_result(result: Any) -> None:
+        """Release pool-backed buffers in a GET result.
+
+        After get(..., type="dataproto"), get(..., type="dict"),
+        get_dataproto, or get_dict, ndarray payloads may be backed by the
+        BufferPool. Call this to release those leases deterministically instead
+        of waiting for GC ``__del__``.
 
         Works for both flat dicts and nested envelope dicts
         (dataproto: {batch: {...}, non_tensor_batch: {...}, meta_info: {...}}).
@@ -1351,8 +1388,35 @@ class MooncakeBundleTransfer:
                 dict_payload[f"{key}.{name}"] = value
         return dict_payload, metadata
 
+    def cleanup(
+        self, ref: DataProtoRefLike, *, type: Literal["dataproto", "dict"] = "dataproto"
+    ) -> None:
+        """Remove all structured object stages referenced by a structured handle.
+
+        ``type`` selects the public lifecycle path. DataProto and flat-dict
+        refs currently share the same structured-ref cleanup mechanics, but the
+        split is kept explicit so future type-specific lifecycle rules have a
+        stable dispatch point.
+        """
+        if type == "dataproto":
+            return self.cleanup_dataproto(ref)
+        if type == "dict":
+            return self.cleanup_dict(ref)
+        raise ValueError(f"unsupported Mooncake payload type: {type!r}")
+
     def cleanup_dataproto(self, ref: DataProtoRefLike) -> None:
         """Remove all structured object stages referenced by a DataProto handle."""
+        self._cleanup_structured_ref(ref)
+
+    def cleanup_dict(self, ref: DataProtoRefLike) -> None:
+        """Remove all structured object stages referenced by a flat-dict handle.
+
+        Refs returned by ``put(..., type="dict")`` use the same structured
+        handle representation as DataProto refs.
+        """
+        self._cleanup_structured_ref(ref)
+
+    def _cleanup_structured_ref(self, ref: DataProtoRefLike) -> None:
         ref = _resolve_dataproto_ref(ref)
         seen: set[str] = set()
         for stage_ref in ref.stage_refs.values():
--- a/mooncake-wheel/tests/test_structured_object_store.py
+++ b/mooncake-wheel/tests/test_structured_object_store.py
@@ -753,10 +753,17 @@ def test_structured_object_ndarray_read_can_use_buffer_pool() -> None:
     assert store.get_into_ranges_calls == before_range_reads + 1
     assert pool.acquire_sizes == [array.nbytes]
 
-    MooncakeBundleTransfer.release_result(result.objects)
+    MooncakeBundleTransfer.release(result.objects, type="dataproto")
     assert pool.release_count == 1
     MooncakeBundleTransfer.release_result(result.objects)
     assert pool.release_count == 1
+
+    dict_result = transfer.materialize(transfer.read_spec(ref))
+    assert hasattr(dict_result.objects["weights"], "_mooncake_pool_owner")
+    MooncakeBundleTransfer.release(dict_result.objects, type="dict")
+    assert pool.release_count == 2
+    MooncakeBundleTransfer.release_dict(dict_result.objects)
+    assert pool.release_count == 2
 
 
 def test_structured_object_multi_buffer_payload_uses_pool_batch_put() -> None:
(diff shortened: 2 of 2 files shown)


mooncake-wheel/mooncake/structured_object_store.py
```diff
@@ -541,12 +541,38 @@ def materialize_into(
...
+    @staticmethod
+    def release(
+        result: Any, *, type: Literal["dataproto", "dict"] = "dataproto"
+    ) -> None:
+        """Release pool-backed buffers in a GET result."""
+        if type == "dataproto":
+            return MooncakeBundleTransfer.release_dataproto(result)
+        if type == "dict":
+            return MooncakeBundleTransfer.release_dict(result)
+        raise ValueError(f"unsupported Mooncake payload type: {type!r}")
```
zxpdemonio, line 559: Looks good as the outer unified entrypoint.

One clarity ask: today `release_dataproto` and `release_dict` are identical, so `type` does not change behavior. Please state that explicitly in this docstring (e.g. “dispatch is reserved for type-specific lifecycle differences; currently both paths share `_release_result`”), matching the PR description. Otherwise callers may treat a mismatched `type` as a footgun when it is not (yet).

mooncake-wheel/mooncake/structured_object_store.py
```diff
@@ -541,12 +541,38 @@ def materialize_into(
...
+
+    @staticmethod
+    def release_dict(result: Any) -> None:
+        """Release pool-backed buffers in a flat-dict GET result."""
+        MooncakeBundleTransfer._release_result(result)
+
     @staticmethod
     def release_result(result: Any) -> None:
+        """Release pool-backed buffers in a GET result."""
+        MooncakeBundleTransfer.release_dataproto(result)
```
zxpdemonio, line 578: Compatibility shim routing through `release_dataproto` is fine. Optional: mention in the docstring that new code should prefer `release(..., type=...)`, and that this helper remains DataProto-oriented by name even though the underlying walker also accepts flat dicts.

mooncake-wheel/mooncake/structured_object_store.py
```diff
@@ -541,12 +541,38 @@ def materialize_into(
...
+
+    @staticmethod
+    def _release_result(result: Any) -> None:
         """Release pool-backed buffers in a GET result.
 
-        After get_dataproto / get_dict, ndarray payloads may be backed by
-        the BufferPool. Call this to release those leases deterministically
+        After get / get_dataproto, ndarray payloads may be backed by the
+        BufferPool. Call this to release those leases deterministically
         instead of waiting for GC ``__del__``.
```
zxpdemonio: Doc nit: “After get / get_dataproto” should also cover flat-dict gets (`get(..., type="dict")` / `get_dict`), since `_release_result` is now the shared backend for `release_dict` as well.

mooncake-wheel/mooncake/structured_object_store.py
```diff
@@ -1351,8 +1377,25 @@ def _read_ragged_tensor_dict_payload(
...
 
+    def cleanup(
+        self, ref: DataProtoRefLike, *, type: Literal["dataproto", "dict"] = "dataproto"
+    ) -> None:
+        """Remove all structured object stages referenced by a structured handle."""
+        if type == "dataproto":
+            return self.cleanup_dataproto(ref)
+        if type == "dict":
+            return self.cleanup_dict(ref)
+        raise ValueError(f"unsupported Mooncake payload type: {type!r}")
```
zxpdemonio, line 1405: Same note as `release`: both branches share `_cleanup_structured_ref`, so `type` is API symmetry rather than a behavioral switch today. Worth one docstring sentence.

Also, `cleanup*` still sits far from the unified `put`/`get`/`release` block — consider relocating in a follow-up for discoverability (not blocking).

mooncake-wheel/mooncake/structured_object_store.py
```diff
@@ -1351,8 +1377,25 @@ def _read_ragged_tensor_dict_payload(
...
+
     def cleanup_dataproto(self, ref: DataProtoRefLike) -> None:
         """Remove all structured object stages referenced by a DataProto handle."""
+        self._cleanup_structured_ref(ref)
+
+    def cleanup_dict(self, ref: DataProtoRefLike) -> None:
+        """Remove all structured object stages referenced by a flat-dict handle."""
+        self._cleanup_structured_ref(ref)
```
zxpdemonio, line 1417: Using `DataProtoRefLike` for `cleanup_dict` is technically correct (dict puts still return `MooncakeDataProtoRef` via the envelope path), but the name is confusing for dict-only callers. A short docstring note (“refs from `put(..., type='dict')` use the same handle type”) would help.

mooncake-wheel/tests/test_structured_object_store.py
```diff
@@ -1900,12 +1900,36 @@ def test_unified_put_get_roundtrips_flat_dict() -> None:
...
+        {
+            "input_ids": np.arange(6, dtype=np.int64).reshape(3, 2),
+            "tokens": [np.asarray([1, 2]), None, np.asarray([3])],
+            "step": 7,
+        },
+        type="dict",
+    )
+
+    assert store.objects
+    transfer.cleanup(ref, type="dict")
```
zxpdemonio, line 1922: Good coverage that unified `cleanup(..., type="dict")` clears store objects.

Please also add:
1. `transfer.cleanup(ref, type="dataproto")` (or assert the unified wrapper) so both dispatch arms are tested
2. a BufferPool-backed `MooncakeBundleTransfer.release(result, type="dict")` case that increments `pool.release_count` — the current dict release smoke only checks `is None` on a non-pooled dict

mooncake-wheel/tests/test_structured_object_store.py
```diff
@@ -1978,9 +2002,10 @@ def test_ref_aliases_match_dataproto_ref_helpers() -> None:
     assert import_ref is import_dataproto_ref
 
 
-def test_release_result_is_available_for_split_api_compatibility() -> None:
+def test_unified_release_is_available_for_split_api_compatibility() -> None:
     result = {"tokens": [np.asarray([1, 2]), None]}
 
+    assert MooncakeBundleTransfer.release(result, type="dict") is None
```
zxpdemonio, line 2024: This only proves the call is importable/returns `None`. Consider asserting the public split helpers exist too (`release_dataproto`, `release_dict`, `cleanup_dict`) since those are part of the PR’s advertised API surface.

zxpdemonio (commented):
## Review summary

Clean, small API-layering change. Matching `put`/`get` with unified `release(..., type=...)` and `cleanup(..., type=...)`, while keeping the split helpers and `release_result` for compatibility, is the right shape.

I re-ran the focused tests from the PR description: **5 passed, 2 skipped**. No functional blockers in the current identical-backend design.

> Note: posted as `COMMENT` because GitHub disallows `REQUEST_CHANGES` / typically also self-`APPROVE` on own PRs (`zxpdemonio`).

### Should fix / strengthen
1. **`type` is currently ceremonial for both `release` and `cleanup`** — both branches call the same private helper. That matches the PR text, but please call it out in the docstrings so callers do not assume wrong-`type` is unsafe today (or add a cheap runtime guard later if semantics diverge).
2. **Tests under-exercise the new surface**:
   - no BufferPool-backed `release(..., type="dict")` assertion (only dataproto + a no-op smoke for dict)
   - no `cleanup(..., type="dataproto")` coverage (only `cleanup_dataproto` / `cleanup(..., type="dict")`)
   - public split helpers `release_dataproto` / `release_dict` / `cleanup_dict` are not directly asserted

### Nits
3. `_release_result` docstring says “After get / get_dataproto” — please also mention unified `get(..., type="dict")` / `get_dict`.
4. Consider placing `cleanup` / `cleanup_*` near `put`/`get`/`release` for discoverability; they are currently ~800 lines away from the unified put/get block.

zxpdemonio:
Updated in `73fa7dca`:
- clarified in docstrings that `type` is the public lifecycle dispatch point and that DataProto/dict paths currently share the same private mechanics;
- documented the DataProto-oriented compatibility role of `release_result`;
- expanded `_release_result` docs to cover `get(..., type=\"dict\")` / `get_dict`;
- added direct coverage for `release(..., type=\"dict\")` with BufferPool-backed data;
- added unified `cleanup(..., type=\"dataproto\")` coverage;
- asserted the public split helpers are available.
- Validation: `py_compile`, `git diff --check`, and targeted pytest passed: `6 passed, 2 skipped, 94 deselected`.

he-yufeng:
The lifecycle layering reads well: keeping `release_result` as a compat alias over `_release_result` means flat-dict callers keep working even though the intermediate name is now dataproto-oriented, and having `release(..., type=...)` / `cleanup(..., type=...)` as the explicit dispatch point gives the future type-specific rules somewhere stable to land.

One gap worth closing before merge: the new public surface has no test coverage of its own. The only test change rewires an existing call to `release(type="dataproto")`, which exercises the path that already had coverage. Missing:

- `release(..., type="dict")` and `cleanup(..., type="dict")` actually dispatching (today both happen to share the same walker, so a regression that breaks the dict route would pass everything)
- the `ValueError` branch on an unsupported `type` for both `release` and `cleanup`

All three are a few lines each against the existing fake-store harness.

he-yufeng:
All three are in: the dict dispatch on both `release` and `cleanup`, and the unsupported-type `ValueError` branch is pinned with `pytest.raises` on both helpers. Thanks for taking it seriously. Nothing left from my side.

PR #3062, opened 2026-07-22
[Store] Arm the etcd view-change watch once per wait instead of per iteration

### LINKED ISSUE #3059: [Bug]: Thread leak in HA mode — LeaderMonitorThread recreates etcd watch every 30s, leaking Go goroutines
Summary
When using SGLang with PD (prefill-decode) disaggregation, the P-side scheduler uses Mooncake store as the L3 hierarchical KV cache backend (HiCache). When the store client is configured in HA mode (MOONCAKE_MASTER=etcd://...), the LeaderMonitorThreadMain loop in WaitForViewChange cancels and re-arms the etcd prefix watch every 30s timeout. Each re-arm spawns a new Go goroutine (cli.Watch). If the previous goroutine fails to exit within kWatchStopTimeoutMs (5s), PrefixWatchGuard intentionally leaks the watch state to avoid UAF, and the goroutine (and its OS thread) accumulates indefinitely.

Environment
- Use case: SGLang PD disaggregation, P-side scheduler uses Mooncake store as L3 HiCache storage backend
- Mooncake store client (Python mooncake package, store.so), HA mode enabled (master_server_entry starts with etcd://)
- Observed on 8 P-side scheduler processes, each growing ~7 threads/min under request load
- D-side (decode) not affected — does not use store client
- No growth when idle (no requests)
- 
Root Cause
EtcdLeaderCoordinator::WaitForViewChange (etcd_leader_coordinator.cpp:348-443):
1. Each loop iteration creates a PrefixWatchGuard that arms a new watch via EtcdHelper::WatchWithPrefixFromRevision
2. The wrapper (etcd_wrapper.go:968) spawns a Go goroutine running cli.Watch(ctx, p)
3. On the next iteration (after 30s timeout), the guard's destructor calls CancelWatchWithPrefix + WaitWatchWithPrefixStopped (5s timeout)
4. If the goroutine doesn't exit in time, the guard leaks the state (etcd_leader_coordinator.cpp:92-99) and the goroutine survives, holding an OS thread
The etcd watch is a long-lived streaming RPC, but the code tears it down and recreates it every 30s, causing steady goroutine accumulation.

Reproduction
1. Start a mooncake master in HA mode with etcd backend
2. Start SGLang P-side scheduler with HiCache L3 enabled and MOONCAKE_MASTER=etcd://<endpoint>
3. Send requests, observe thread count: ls /proc/<pid>/task | wc -l grows ~7/min per scheduler process
4. Check logs for: "Prefix watch goroutine did not stop in time for prefix ...; leaking ViewChangeWatchState to avoid UAF"
Impact
- P-side scheduler processes eventually hit Too many open files / thread limits
- Requires periodic process restart to mitigate
- Only P-side affected (uses store client for HiCache); D-side unaffected

Suggested Fix

Keep the watch alive across loop iterations instead of canceling and re-arming every 30s. Only re-arm when the watch reports broken:
// Arm watch once, outside the loop
auto* state = new ViewChangeWatchState();
EtcdHelper::WatchWithPrefixFromRevision(..., state, &ViewChangeWatchCallback);
while (true) {
    auto current_view = ReadCurrentView();
    if (view changed) return;
    // Wait for watch event, no cancel/re-arm
    state->cv.wait_for(lock, remaining, ...);
    // Only re-arm if watch broke
    if (state->broken) { rearm(); }
}

he-yufeng:
Fixes #3059.

Every loop iteration in `WaitForViewChange` armed a new watch, and the guard destructor only gave the goroutine 5s to stop. Miss that budget and the guard leaks the state on purpose (UAF avoidance), so each miss is one goroutine plus an OS thread. That's the climb the reporter measured (~7/min per P-side process).

Now one watch lives for the whole wait. Events just flip the flag and the same goroutine keeps delivering; a fresh goroutine only happens when etcd says the watch broke (event_type 2, now tracked on the state). The per-iteration leak is gone, and the cancel/stop at exit is the only one left. Arm-before-read ordering and the fallback poll are untouched.

Compiles clean with the project's normal build flags. I don't have a local etcd rig, so I'm leaning on the CI integration tests for behavior.

### DIFF
--- a/mooncake-store/src/ha/leadership/backends/etcd/etcd_leader_coordinator.cpp
+++ b/mooncake-store/src/ha/leadership/backends/etcd/etcd_leader_coordinator.cpp
@@ -31,6 +31,10 @@ constexpr auto kViewChangeFallbackPollInterval = std::chrono::milliseconds(200);
 // watch context is destroyed. Mirrors the value used by the oplog notifier.
 constexpr int kWatchStopTimeoutMs = 5000;
 
+// event_type delivered by the etcd watch goroutine when the watch itself ends
+// (0 = PUT, 1 = DELETE, 2 = WATCH_BROKEN; see EtcdHelper::WatchCallbackFn).
+constexpr int kWatchEventBroken = 2;
+
 // Shared state between WaitForViewChange and the etcd watch callback. The
 // callback only flips `changed` and wakes the waiter; the waiter decides what
 // the change means by re-reading the view.
@@ -38,6 +42,7 @@ struct ViewChangeWatchState {
     std::mutex mutex;
     std::condition_variable cv;
     bool changed = false;
+    bool broken = false;
 };
 
 // C-style trampoline invoked by the etcd watch goroutine for every event on the
@@ -47,7 +52,7 @@ struct ViewChangeWatchState {
 // CancelWatchWithPrefix + WaitWatchWithPrefixStopped before it is destroyed.
 void ViewChangeWatchCallback(void* context, const char* /*key*/,
                              size_t /*key_size*/, const char* /*value*/,
-                             size_t /*value_size*/, int /*event_type*/,
+                             size_t /*value_size*/, int event_type,
                              int64_t /*mod_revision*/) {
     auto* state = static_cast<ViewChangeWatchState*>(context);
     if (state == nullptr) {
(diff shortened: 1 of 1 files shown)


Icedcoco (approved):
I took another pass over the `WaitForViewChange` path and the Go prefix-watch wrapper. This looks like the right shape for #3059.

The important invariant still holds: the watch is established before the read, and the wrapper waits for the server-side created notification before returning OK, so the read/watch union should not miss a `master_view` update. Keeping one watch alive for the whole wait removes the per-iteration cancel/stop cycle that caused the goroutine/thread buildup. The `WATCH_BROKEN` path also looks reasonable: normal events only trigger a re-read, and only a broken stream re-arms.

I do not see a blocking correctness issue here. The remaining gap is test coverage for the exact steady-state leak/re-arm behavior, but the existing HA watch tests plus the green CI matrix are enough for this fix in my view. LGTM.

PR #3063, opened 2026-07-22
[Store] stop steady-state etcd watch goroutine leak in WaitForViewChange (#3059)

### LINKED ISSUE #3059: [Bug]: Thread leak in HA mode — LeaderMonitorThread recreates etcd watch every 30s, leaking Go goroutines

SuperMarioYL:
## Summary
Fixes #3059 — a steady-state Go goroutine leak in the HA leader-monitor path.

When the store client runs in HA mode (`MOONCAKE_MASTER=etcd://...`),
`LeaderMonitorThreadMain` calls `EtcdLeaderCoordinator::WaitForViewChange`,
which armed a fresh etcd prefix watch **every iteration** of its wait loop.
Under SGLang PD disaggregation load this churns watch goroutines; when
`WaitWatchWithPrefixStopped` did not complete within `kWatchStopTimeoutMs` (5s)
the guard intentionally leaked the C++ state to avoid a use-after-free
(the leak-on-timeout branch in `PrefixWatchGuard::~PrefixWatchGuard`,
`etcd_leader_coordinator.cpp:101-109`) and the goroutine lingered — the
~7-threads/min accumulation reported in #3059.

## Root cause
`mooncake-store/src/ha/leadership/backends/etcd/etcd_leader_coordinator.cpp` —
per-iteration `new ViewChangeWatchState()` + `PrefixWatchGuard` inside the
`while(true)` of `WaitForViewChange`; the watch was armed and torn down on
every loop iteration.

## Fix (call-scoped, no public-signature change)
- Hoist `ViewChangeWatchState` + `PrefixWatchGuard` out of the `while(true)` so
  the watch is allocated **once per `WaitForViewChange` call** and armed **once
  before the loop**.
- Add `watch_healthy` to `ViewChangeWatchState`; extend `ViewChangeWatchCallback`
  to set it `false` on `event_type == 2` (`WATCH_BROKEN`). The Go wrapper already
  delivers `WATCH_BROKEN` (event_type 2 is documented at
  `etcd_oplog_change_notifier.cpp:225`).
- Re-arm **only on `WATCH_BROKEN`**: `CancelWatchWithPrefix` +
  `WaitWatchWithPrefixStopped` + `WatchWithPrefixFromRevision(0)` **and re-read**
  so the union (watch from R_rearm) ∪ (read at R_read ≥ R_rearm) still covers the
  broken window — the same arm-before-read revision algebra preserved by the
  re-arm block (arm the watch at R_rearm, then read at R_read ≥ R_rearm so the
  union covers the gap; see the comment at `etcd_leader_coordinator.cpp:384-390`).
- Drain stale `changed` flag: because the watch is now long-lived, `changed` is no
  longer implicitly reset each iteration. After a loop-top re-read that sees no
  view change, a pending `changed` (e.g. a sibling-key PUT under the prefix) is
  drained and the loop re-reads at a higher revision — this catches any real
  change that arrived during the read and avoids a tight re-read loop, **without**
  dropping an event that arrived mid-read (which a clear-then-wait would do).

The leak-on-timeout UAF policy in `PrefixWatchGuard` (`:88-110`) and the
`!watching` fallback poll are unchanged. No change to the `WaitForViewChange`
public signature; no coordinator-member lifetime state added. This removes the
steady-state per-iteration churn; the residual leak-on-timeout policy is
intentionally unchanged, so a final `WaitWatchWithPrefixStopped` timeout at
call-end can still leak one state+goroutine per call (rare, not steady-state).

## Tests
Adds two tests to `high_availability_test.cpp`:

- `WaitForViewChangeRepeatedStableCallsDoNotBreakChangeDetection` —
  acquires+renews a leader, calls `WaitForViewChange` 6× on a stable leader
  (each times out cleanly with `!changed && timed_out`), then releases
  leadership from a **concurrent thread while the final `WaitForViewChange` is
  blocked** on the long-lived prefix watch (not before it, which the
  synchronous loop-top read would catch regardless of watch-vs-poll). The
  change is therefore delivered as a watch DELETE event, and the test asserts a
  detection latency strictly below `kViewChangeFallbackPollInterval` (200ms): a
  poll-only regression (watch never armed) would be stranded in a 200ms poll
  sleep and could not reliably meet the bound, while the event-driven watch
  delivers in tens of ms. Guards the once-per-call arm/teardown flow reshaped by
  this change.

- `WaitForViewChangeRearmsAndStillDetectsAfterWatchBroken` — covers the
  `WATCH_BROKEN → re-arm + re-read` path (the `if (watch_broken)` block in the
  wait loop) at the coordinator level. While `WaitForViewChange` is blocked on
  the long-lived prefix watch, it resets the etcd store client to force a
  `WATCH_BROKEN` (event_type 2). The Go wrapper's `cancelAllStorePrefixWatches`
  marks the watch broken before cancelling its context, so exactly one
  `WATCH_BROKEN` callback fires — the same mechanism as the existing
  `EtcdStoreClientResetTriggersPrefixWatchBrokenAndReconnect`
  (`high_availability_test.cpp:637`), which proves `WATCH_BROKEN` is delivered
  by the Go wrapper, but driven here through `WaitForViewChange`'s own watch
  rather than a raw `WatchWithPrefixFromRevision` with a null context. After the
  re-arm settle, it releases leadership (deleting the `master_view` key) and
  asserts the re-armed watch still delivers the change before the deadline —
  proving the re-arm path does not drop the watch, deadlock, or crash. The
  coordinator's re-arm reuses the same `ViewChangeWatchState` +
  `WatchWithPrefixFromRevision` primitive as the initial arm, so the only NEW
  logic is the drain + re-arm-on-broken scheduling (covered by the revision-
  algebra comment + the two tests above). The leak itself is a steady-state
  ~7-threads/min observation under load with no deterministic unit repro; a
  coordinator-level etcd-server-restart integration test (needs fixture support
  to stop/restart the external etcd process) is a desirable follow-up.

## Validation
`clang++ -std=c++17 -fsyntax-only` clean against the project headers (third-party
stubs for glog/ylt/gtest on this host). The full `high_availability_test` runs on
the project's Linux CI (this machine cannot build the C++/RDMA/etcd tree).

Closes #3059.

### DIFF
--- a/mooncake-store/src/ha/leadership/backends/etcd/etcd_leader_coordinator.cpp
+++ b/mooncake-store/src/ha/leadership/backends/etcd/etcd_leader_coordinator.cpp
@@ -377,29 +377,32 @@ EtcdLeaderCoordinator::WaitForViewChange(
     // allocate ONCE per WaitForViewChange call. The previous per-iteration
     // arming churned Go watch goroutines; when WaitWatchWithPrefixStopped did
     // not complete in time the guard leaked state (UAF-safe) and a goroutine
-    // lingered -- the ~7-threads/min leak in #3059. Re-arm ONLY on WATCH_BROKEN.
+    // lingered -- the ~7-threads/min leak in #3059. Re-arm ONLY on
+    // WATCH_BROKEN.
     auto* state = new ViewChangeWatchState();
     PrefixWatchGuard guard(master_view_key_, state);
 
-    // Arm the watch ONCE before entering the loop. Arm-before-read: the watch is
-    // established at revision R_watch; the first ReadCurrentView observes
+    // Arm the watch ONCE before entering the loop. Arm-before-read: the watch
+    // is established at revision R_watch; the first ReadCurrentView observes
     // R_read >= R_watch. A change at revision C is caught by C < R_read (the
     // read) or C >= R_watch (the watch), and the ranges overlap, so a change
     // between read and watch cannot be missed. start_revision = 0 avoids a
     // possibly-compacted revision. A long-lived watch keeps this for all later
-    // changes; the gap reopens only on re-arm after WATCH_BROKEN (handled below).
+    // changes; the gap reopens only on re-arm after WATCH_BROKEN (handled
+    // below).
     bool watching = false;
     {
-        const auto remaining = std::chrono::duration_cast<std::chrono::milliseconds>(
-            deadline - std::chrono::steady_clock::now());
+        const auto remaining =
+            std::chrono::duration_cast<std::chrono::milliseconds>(
+                deadline - std::chrono::steady_clock::now());
         if (remaining > kViewChangeFallbackPollInterval) {
             // Defensively clear any lingering watch on this key, then arm a new
             // one. Both calls are no-ops when nothing is registered.
             EtcdHelper::CancelWatchWithPrefix(master_view_key_.c_str(),
                                               master_view_key_.size());
-            EtcdHelper::WaitWatchWithPrefixStopped(
-                master_view_key_.c_str(), master_view_key_.size(),
-                kWatchStopTimeoutMs);
+            EtcdHelper::WaitWatchWithPrefixStopped(master_view_key_.c_str(),
+                                                   master_view_key_.size(),
+                                                   kWatchStopTimeoutMs);
             auto watch_err = EtcdHelper::WatchWithPrefixFromRevision(
                 master_view_key_.c_str(), master_view_key_.size(),
                 /*start_revision=*/0, state, &ViewChangeWatchCallback);
@@ -449,9 +452,9 @@ EtcdLeaderCoordinator::WaitForViewChange(
             if (remaining > kViewChangeFallbackPollInterval) {
                 EtcdHelper::CancelWatchWithPrefix(master_view_key_.c_str(),
                                                   master_view_key_.size());
-                EtcdHelper::WaitWatchWithPrefixStopped(
-                    master_view_key_.c_str(), master_view_key_.size(),
-                    kWatchStopTimeoutMs);
+                EtcdHelper::WaitWatchWithPrefixStopped(master_view_key_.c_str(),
+                                                       master_view_key_.size(),
+                                                       kWatchStopTimeoutMs);
                 auto watch_err = EtcdHelper::WatchWithPrefixFromRevision(
                     master_view_key_.c_str(), master_view_key_.size(),
                     /*start_revision=*/0, state, &ViewChangeWatchCallback);
--- a/mooncake-store/src/ha/leadership/backends/etcd/etcd_leader_coordinator.cpp
+++ b/mooncake-store/src/ha/leadership/backends/etcd/etcd_leader_coordinator.cpp
@@ -33,21 +33,27 @@ constexpr int kWatchStopTimeoutMs = 5000;
 
 // Shared state between WaitForViewChange and the etcd watch callback. The
 // callback only flips `changed` and wakes the waiter; the waiter decides what
-// the change means by re-reading the view.
+// the change means by re-reading the view. `watch_healthy` is flipped to false
+// by the callback on WATCH_BROKEN so the waiter re-arms before blocking again.
+// Both flags live under `mutex`.
 struct ViewChangeWatchState {
     std::mutex mutex;
     std::condition_variable cv;
     bool changed = false;
+    bool watch_healthy = true;
 };
 
 // C-style trampoline invoked by the etcd watch goroutine for every event on the
 // watched prefix (PUT / DELETE / WATCH_BROKEN). Any event is treated as "the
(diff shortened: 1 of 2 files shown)


he-yufeng:
This covers the same ground as #3062, opened about two hours earlier for the same issue: hoist the watch state out of the wait loop, arm once per call, re-arm only when etcd reports WATCH_BROKEN, keep the arm-before-read invariant across re-arms, and keep the short-poll fallback. The #3059 thread has the discussion that led to this shape.

One behavioral difference worth a look whichever base lands: if the re-arm attempt itself fails (RPC error), this version flips `watch_healthy` back to true and only polls from then on, so the watch is never re-armed for the rest of the wait. #3062 retries the arm on later iterations once there is room, so watch service recovers after a transient failure. The gtest addition is a nice gap fill in any case.

Icedcoco:
Thanks for putting this together.

This PR covers the same root cause as #3062, which was opened earlier and already has a smaller, green fix for #3059. I also do not think we should merge this version as-is: it adds extra WATCH_BROKEN/state handling and test changes beyond the minimal leak fix, and the re-arm failure behavior is less desirable because a failed re-arm can leave the call polling instead of retrying the watch setup.

CI is also still red on this PR, including the main build jobs and CI Gate. Given the overlap, later submission time, larger behavioral surface, and failing checks, I think the right path is to proceed with #3062 and keep any useful test coverage ideas for a follow-up if needed.

SuperMarioYL:
Thanks @Icedcoco @he-yufeng for the careful review — agreed, #3062 is the cleaner path: smaller, green, earlier, and the re-arm-retry behavior is better. My version flipping `watch_healthy` back to true on a failed re-arm and then only polling (never re-arming for the rest of the wait) is genuinely worse, so I will defer to #3062.

The one thing worth carrying over is the `TransferMetadataStaleSegmentInvalidationTest` gtest — happy to port it onto #3062 as a follow-up if you would like the coverage; just say the word.

Closing in favor of #3062.

PR #2112, opened 2026-05-15
[Store] feat: introduce per-tenant storage quota in master

dtcccc:
### Overview

This PR adds **per-tenant byte-level storage quota** to the Mooncake Store master, enabling operators to cap memory usage for each `tenant_id` independently. It addresses the quota and isolation aspects of #1836 

Quota is **opt-in** via `--enable_tenant_quota`; with the gate off, all new code paths are strict no-ops and existing single-tenant deployments observe identical behavior.

This PR is built independently against the upstream main branch and does not depend on PR #1886 at build or runtime.

---

### Relationship to PR #1886

This PR and PR #1886 share the same **design philosophy** — treating `tenant_id` as a first-class identity dimension — but are **independently buildable and independently mergeable**:

| Dimension | This PR | PR #1886 |
|---|---|---|
| **Focus** | Quota enforcement & scoped eviction | Identity-native metadata model (`LogicalObjectId`) |
| **Tenant signal** | `ReplicateConfig.tenant_id` (string, default `"default"`) | Same field, same semantics |
| **Key isolation** | Relies on deployment-level key-namespace separation (prefix convention) until #1886 lands | Structural isolation via `LogicalObjectId` composite key |

When both are merged, the combination achieves full per-tenant isolation (structural key separation from #1886 + resource capping from this PR). Until then, each PR independently delivers value: this PR provides quota enforcement and scoped eviction for deployments that already separate tenant key namespaces by convention.

---

#### Two-Phase Quota Accounting (Reserve → Commit / Abort)

PutStart performs a **Reserve** upfront against the tenant's quota headroom (`used + reserved + incoming ≤ max_bytes`). On successful allocation, only the actually-charged bytes (size × memory replica count) are **Committed**; on failure, the reservation is **Aborted**. This prevents concurrent PutStart calls from collectively exceeding quota through a TOCTOU gap.

Every object erase path (Remove, PutRevoke, BatchEvict, BatchReplicaClear, Copy/MoveRevoke, etc.) releases the exact snapshotted charge, ensuring the accounting invariant is maintained throughout the object lifecycle.

#### Per-Tenant Scoped Eviction

When `Reserve` fails, PutStart triggers **`BatchEvictInTenant`** — an eviction pass scoped to the offending tenant only. Other tenants' objects are never disturbed.

- **Frontier index**: Per-tenant `std::set<TenantFrontierEntry>` ordered by `(lease_timeout ASC, size DESC, key ASC)`, co-located with quota state under a single `std::mutex`.
[...]

PR #3071, opened 2026-07-23
[Store] Extract tenant quota table from MasterService

Aionw:
## Description

Refactor the tenant quota in-memory state out of `MasterService` into a dedicated two-layer component:

- `TenantQuotaTable` is a single-threaded core state machine that owns policy state, effective quota calculation, accounting invariants, snapshots, metadata counts, usage rebuilds, and orphan cleanup.
- `ShardedTenantQuotaTable<NumShards>` adds per-tenant sharding and synchronization. `MasterService` explicitly uses 1024 shards.
- `MasterService` keeps quota feature gating, capacity discovery, policy-store persistence, metadata aggregation, error-code mapping, and cross-module transaction orchestration.
- Internal quota state and assignment details are hidden; snapshots use the strong `TenantId` type and convert to strings only at HTTP/metrics boundaries.
- Redundant quota state, shard helpers, recompute locking, and no-op forwarding helpers are removed from `MasterService`.

The refactor preserves existing RPC/Admin formats, configuration, metrics, admission behavior, deterministic proportional quota allocation, policy deletion rollback, and accounting behavior.

This work is covered by the tenant quota design in #2153. It overlaps conceptually with #2112, but differs in scope: #2112 introduces a tenant quota feature and eviction model, while this PR only consolidates the quota behavior already present on current `main`; it does not add a new quota model, API, or configuration.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**

```bash
ctest --test-dir <build-dir> --output-on-failure \
  -R '^(tenant_quota_test|master_service_tenant_quota_test)$'
ctest --test-dir <build-dir> --output-on-failure \
  -R '^(master_admin_server_test|master_metrics_test)$'
pre-commit run --files \
  mooncake-store/include/master_service.h \
  mooncake-store/include/tenant_quota.h \
  mooncake-store/src/master_admin_service.cpp \
  mooncake-store/src/tenant_quota.cpp \
  mooncake-store/tests/master_service_tenant_quota_test.cpp \
  mooncake-store/tests/tenant_quota_test.cpp
git diff --check
```

**Test results:**

- [x] Unit tests pass
- [x] Integration tests pass
- [ ] Manual testing done

Additional regression coverage during development: `master_service_test` passed all 144 tests.

## Checklist

- [x] I have updated the documentation (not applicable: internal refactor with unchanged public behavior)
- [x] For changes >500 LOC: covered by RFC #2153

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex assisted with implementation, test expansion, code review, validation, and PR drafting. The human submitter remains responsible for reviewing and defending the changes end-to-end.


mooncake-store/src/master_service.cpp
```diff
@@ -1158,150 +1010,50 @@ uint64_t MasterService::GetTenantQuotaCapacityBytes() {
...
-        auto it = shard.tenants.find(assignment.tenant_id);
-        if (it == shard.tenants.end()) {
-            continue;
-        }
-        auto& state = it->second;
-        state.effective_quota_bytes = assignment.effective_quota_bytes;
-        RefreshTenantQuotaOverQuota(state);
-    }
+    tenant_quota_table_.RecomputeEffectiveQuotas(
+        GetTenantQuotaAllocatableCapacityBytes());
```
Lin-z-w: Please serialize capacity sampling with quota recomputation. `GetTenantQuotaAllocatableCapacityBytes()` is evaluated before entering `ShardedTenantQuotaTable::RecomputeEffectiveQuotas()`, so the table's `recompute_mutex_` does not protect the capacity read.

For example, thread A can read the old capacity (100 GiB) and pause; after a segment is unmounted, thread B reads and applies the new capacity (50 GiB); thread A can then acquire `recompute_mutex_` and overwrite the result using the stale 100 GiB value. The final effective quotas may therefore exceed the current allocatable capacity until another recomputation occurs.

The previous implementation held `tenant_quota_recompute_mutex_` across both capacity discovery and quota updates. Please preserve that atomicity, either with an outer lock or by evaluating a capacity provider after acquiring `recompute_mutex_`. The same ordering issue also applies to capacity passed into policy application and rollback paths. A regression test should use two different capacities rather than repeatedly recomputing with the same value.

Aionw: Good catch. I restored a Master-level `tenant_quota_recompute_mutex_` and now hold it across both capacity sampling and the corresponding quota-table update. The same coordination covers policy application, policy rollback, metadata usage rebuild, and segment-driven recomputation, while `ShardedTenantQuotaTable` continues to accept a numeric capacity and never calls back into `SegmentManager` under its internal lock. I also added `CapacityIsSampledInsideQuotaRecomputeCoordination`, which blocks the coordination point, changes capacity from 100 to 150, and verifies that recomputation applies 150 rather than the stale value. The related quota/admin/metrics tests and touched-file pre-commit pass. I have left the thread open for re-review.

Lin-z-w (approved):
LGTM.

ykwd (approved):

PR #3072, opened 2026-07-23
[Bugfix][TransferEngine] Invalidate stale remote segments in syncSegmentCache

SuperMarioYL:
## Description

`TransferMetadata::syncSegmentCache` refreshes the local segment-descriptor cache by polling the metadata backend, but it only ever *adds* and *updates* entries — it never *invalidates* them. When a remote peer unmounts its segment (graceful `freeEngine` → `removeSegmentDesc` → backend remove) or crashes and the master expires its key (`ClientMonitorFunc` → `cleanupHttpMetadata` removes `mooncake/ram/<peer>`), every other node's refresh poll fetches the now-missing key, logs `segment <name> is now invalid`, increments `failed_count`, and **leaves the stale `SegmentDesc` in the cache**. The only erase sites (`removeSegmentDesc` at `transfer_metadata.cpp:568-569` and the local-remove path at `:1254-1255`) are LOCAL-only; neither covers a REMOTE segment that vanished from the backend.

The cached, now-dead descriptor keeps getting served: `getSegmentDescByName` returns the cached entry under a read lock with no backend re-check (`transfer_metadata.cpp:1143-1147`), so `MultiTransport::selectTransport` resolves the stale desc and posts RDMA/TCP to a dead endpoint. Transfers to the departed peer fail or time out on **every** subsequent request until the caching node itself restarts.

This change adds conservative invalidation to `syncSegmentCache`'s apply phase: a per-segment consecutive-failure streak; once a cached **remote** segment fails its backend fetch for `kStaleSegmentFailureThreshold` (=2) consecutive sync cycles, it is erased from both `segment_name_to_id_map_` and `segment_id_to_desc_map_` under the write lock (mirroring the erase pattern in `removeSegmentDesc`). A successful fetch resets the streak.

**Why a threshold rather than erasing on the first failure:** `MetadataStoragePlugin::get` returns a single `bool` that collapses "key absent" and "transient error" (etcd blip, curl timeout, 5xx) into `false`. Erasing on the first failure would purge the whole cache during a transient backend blip and cause a transfer outage. A ≥2-consecutive threshold distinguishes a genuinely-removed segment (fails every cycle until erased) from a transient blip (fails once, succeeds next cycle → streak reset → not erased). In-flight transfers that already hold a `shared_ptr` copy of the descriptor are unaffected by a map erase (refcount keeps it alive), so there is no dangle.

## Module

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

Added `TransferMetadataStaleSegmentInvalidationTest` in `mooncake-transfer-engine/tests/transfer_metadata_test.cpp`, using an in-memory fake `MetadataStoragePlugin` (no RDMA/CUDA/hardware):

- `ErasesStaleEntryAfterConsecutiveFailures` — seeds remote segment "B", drops its backend key, asserts "B" is **still cached after the 1st** failed `syncSegmentCache` (streak 1 < threshold, guarding against transient-blip purge) and `getSegmentDescByName("B") == nullptr` **after the 2nd** (stale entry erased). On `main` there is no invalidation branch, so the `== nullptr` assertion fails (red); on this branch it passes (green).
- `TransientFailureIsRetained` — a segment that fails once then succeeds is **not** erased (streak reset).

The test drives the real `syncSegmentCache` / `getSegmentDescByName` code paths via the fake backend; it does not mock the function under fix.

**Test commands:**
```bash
cmake -S mooncake-transfer-engine -B build-te -DCMAKE_BUILD_TYPE=Debug -DBUILD_TRANSFER_ENGINE_TESTS=ON
cmake --build build-te --target transfer_metadata_test -j
ctest --test-dir build-te -R transfer_metadata --output-on-failure
```

**Test results:**
- [x] Unit tests pass — added; verified by design (red-on-`main` / green-on-branch). Authoritative run is Linux CI (the change is authored against `upstream/main` and the metadata unit test needs only gtest + the storage-plugin abstraction, no transport hardware).
- [ ] Integration tests pass (if applicable) — N/A (unit-level fix).
- [ ] Manual testing done — N/A.

## Checklist

- [ ] I have formatted my code using `./scripts/code_format.sh` — not run locally (host lacks the RDMA/CUDA build env + clang-format); additions hand-formatted to the repository's Google style / `.clang-format` to match surrounding code. Will run `pre-commit run --all-files` in CI.
- [ ] I have run `pre-commit run --all-files` and all hooks pass — deferred to CI (see above).
- [ ] I have updated the documentation (if applicable) — N/A.
- [ ] For changes >500 LOC: I have filed an RFC issue — N/A (82 functional LoC).

## AI Assistance Disclosure

- [x] No AI tools were used

### DIFF
--- a/mooncake-transfer-engine/include/transfer_metadata_plugin.h
+++ b/mooncake-transfer-engine/include/transfer_metadata_plugin.h
@@ -26,6 +26,16 @@ struct MetadataStoragePlugin {
     virtual ~MetadataStoragePlugin() {}
 
     virtual bool get(const std::string &key, Json::Value &value) = 0;
+    // Error-aware lookup: same contract as get() but reports whether a
+    // non-found result is an authoritative key absence (kNotFound) or a
+    // transient backend failure (kUnavailable). Plugins that can tell the
+    // two apart (Redis nil reply, HTTP 404 vs 5xx) should override this;
+    // the default delegates to get() and maps false->kNotFound, preserving
+    // the pre-existing behavior for plugins that do not distinguish.
+    virtual GetResult getWithStatus(const std::string &key,
+                                    Json::Value &value) {
+        return get(key, value) ? GetResult::kFound : GetResult::kNotFound;
+    }
     virtual bool set(const std::string &key, const Json::Value &value) = 0;
     virtual bool remove(const std::string &key) = 0;
 };
--- a/mooncake-transfer-engine/include/transfer_metadata_plugin.h
+++ b/mooncake-transfer-engine/include/transfer_metadata_plugin.h
@@ -29,12 +29,15 @@ struct MetadataStoragePlugin {
     // Error-aware lookup: same contract as get() but reports whether a
     // non-found result is an authoritative key absence (kNotFound) or a
     // transient backend failure (kUnavailable). Plugins that can tell the
-    // two apart (Redis nil reply, HTTP 404 vs 5xx) should override this;
-    // the default delegates to get() and maps false->kNotFound, preserving
-    // the pre-existing behavior for plugins that do not distinguish.
+    // two apart (etcd key-absent response, Redis nil reply, HTTP 404 vs 5xx)
+    // must override this; the default delegates to get() and maps any
+    // failure to kUnavailable. Failing closed is deliberate: a plugin that
+    // cannot distinguish absence from an outage must never report kNotFound,
+    // because syncSegmentCache() would then evict a still-live cache entry
+    // on a backend outage.
     virtual GetResult getWithStatus(const std::string &key,
                                     Json::Value &value) {
-        return get(key, value) ? GetResult::kFound : GetResult::kNotFound;
+        return get(key, value) ? GetResult::kFound : GetResult::kUnavailable;
     }
     virtual bool set(const std::string &key, const Json::Value &value) = 0;
     virtual bool remove(const std::string &key) = 0;
--- a/mooncake-transfer-engine/include/transfer_metadata.h
+++ b/mooncake-transfer-engine/include/transfer_metadata.h
@@ -190,6 +190,16 @@ class TransferMetadata {
 
     ~TransferMetadata();
 
+   protected:
+    // Test-only seam: inject a storage plugin directly, bypassing the
+    // conn-string plugin factory (which LOG(FATAL)s without a real
+    // etcd/redis/http backend) and the P2P handshake daemon. Derived
+    // hardware-free unit tests substitute an in-memory
+    // MetadataStoragePlugin to exercise syncSegmentCache without RDMA/CUDA.
+    explicit TransferMetadata(
+        std::shared_ptr<MetadataStoragePlugin> storage_plugin);
+
+   public:
     std::shared_ptr<SegmentDesc> getSegmentDescByName(
         const std::string &segment_name, bool force_update = false);
 
@@ -272,6 +282,11 @@ class TransferMetadata {
     std::unordered_map<uint64_t, std::shared_ptr<SegmentDesc>>
         segment_id_to_desc_map_;
     std::unordered_map<std::string, uint64_t> segment_name_to_id_map_;
+    // Per-REMOTE-segment consecutive fetch-failure streak, guarded by
+    // segment_lock_. Bumped in syncSegmentCache's apply phase when a cached
+    // segment's backend fetch fails; reset to 0 on a successful fetch. Once it
+    // reaches kStaleSegmentFailureThreshold the cached entry is invalidated.
+    std::unordered_map<std::string, int> segment_failure_counts_;
 
     RWSpinlock notify_lock_;
     std::vector<NotifyDesc> notifys;
--- a/mooncake-transfer-engine/include/transfer_metadata.h
+++ b/mooncake-transfer-engine/include/transfer_metadata.h
@@ -40,6 +40,18 @@ namespace mooncake {
 struct MetadataStoragePlugin;
 struct HandShakePlugin;
 
+// Result of a metadata-backend lookup, distinguishing an authoritative
+// key absence (kNotFound — the master removed the segment key on peer
+// unmount/expiry) from a transient backend failure (kUnavailable — curl
+// timeout, etcd blip, connection drop). syncSegmentCache() advances its
+// invalidation streak only for kNotFound so a metadata-service outage
+// cannot purge the live cache.
+enum class GetResult {
+    kFound,
+    kNotFound,
+    kUnavailable,
+};
+
 #define P2PHANDSHAKE "P2PHANDSHAKE"
 
 class TransferMetadata {
@@ -226,8 +238,11 @@ class TransferMetadata {
     int updateSegmentDesc(const std::string &segment_name,
                           const SegmentDesc &desc);
 
+    // Fetch a segment descriptor from the metadata backend. When |status|
+    // is non-null it receives the GetResult so the caller (syncSegmentCache)
+    // can distinguish an authoritative key removal from a transient outage.
     std::shared_ptr<SegmentDesc> getSegmentDesc(
-        const std::string &segment_name);
+        const std::string &segment_name, GetResult *status = nullptr);
 
     SegmentID getSegmentID(const std::string &segment_name);
 
@@ -279,7 +294,8 @@ class TransferMetadata {
 
    private:
     std::shared_ptr<SegmentDesc> getSegmentDescInternal(
-        const std::string &segment_name, bool force_rpc_update);
+        const std::string &segment_name, bool force_rpc_update,
+        GetResult *status = nullptr);
     int getRpcMetaEntryInternal(const std::string &server_name,
                                 RpcMetaDesc &desc, bool force_update);
     int publishSegmentDesc(const std::string &segment_name,
--- a/mooncake-transfer-engine/include/transfer_metadata.h
+++ b/mooncake-transfer-engine/include/transfer_metadata.h
@@ -241,8 +241,8 @@ class TransferMetadata {
     // Fetch a segment descriptor from the metadata backend. When |status|
     // is non-null it receives the GetResult so the caller (syncSegmentCache)
     // can distinguish an authoritative key removal from a transient outage.
-    std::shared_ptr<SegmentDesc> getSegmentDesc(
-        const std::string &segment_name, GetResult *status = nullptr);
+    std::shared_ptr<SegmentDesc> getSegmentDesc(const std::string &segment_name,
+                                                GetResult *status = nullptr);
 
     SegmentID getSegmentID(const std::string &segment_name);
 
--- a/mooncake-transfer-engine/include/transfer_metadata.h
+++ b/mooncake-transfer-engine/include/transfer_metadata.h
@@ -205,6 +205,16 @@ class TransferMetadata {
 
     ~TransferMetadata();
 
+   protected:
+    // Test-only seam: inject a storage plugin directly, bypassing the
+    // conn-string plugin factory (which LOG(FATAL)s without a real
+    // etcd/redis/http backend) and the P2P handshake daemon. Derived
+    // hardware-free unit tests substitute an in-memory
+    // MetadataStoragePlugin to exercise syncSegmentCache without RDMA/CUDA.
+    explicit TransferMetadata(
+        std::shared_ptr<MetadataStoragePlugin> storage_plugin);
+
+   public:
     std::shared_ptr<SegmentDesc> getSegmentDescByName(
         const std::string &segment_name, bool force_update = false);
 
@@ -295,6 +305,11 @@ class TransferMetadata {
     std::unordered_map<uint64_t, std::shared_ptr<SegmentDesc>>
         segment_id_to_desc_map_;
     std::unordered_map<std::string, uint64_t> segment_name_to_id_map_;
+    // Per-REMOTE-segment consecutive fetch-failure streak, guarded by
+    // segment_lock_. Bumped in syncSegmentCache's apply phase when a cached
+    // segment's backend fetch fails; reset to 0 on a successful fetch. Once it
+    // reaches kStaleSegmentFailureThreshold the cached entry is invalidated.
+    std::unordered_map<std::string, int> segment_failure_counts_;
 
     RWSpinlock notify_lock_;
     std::vector<NotifyDesc> notifys;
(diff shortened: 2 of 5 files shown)


he-yufeng:
Read through the whole change. The design holds up well: the two cycle streak is the right answer to get() collapsing "key removed" and "transient backend error" into one false, the success path resetting the streak keeps a flap from accumulating into a purge, LOCAL_SEGMENT_ID is explicitly excluded, and erasing under the write lock while callers hold shared_ptr copies of the descriptor is safe by refcount. The test seam and the streak/reset/local coverage in the new test file match the mechanism.

One scenario worth a thought: a backend wide outage (etcd down, not one segment removed) fails every fetch in the cycle. Each remote name then increments its streak, and after kStaleSegmentFailureThreshold cycles the entire remote cache gets invalidated at once, so nodes lose descriptors they could have kept transferring from until the backend recovered. A cheap guard is to skip the streak increment when the whole fetch set failed in a single cycle (systemic failure, not per segment disappearance), or to document that mass invalidation on backend outage is intended. Not asking to block on it, just flagging since it flips the failure mode from "stale reads" to "everything uncached at once".

SuperMarioYL:
Thanks @he-yufeng for the close read and the positive assessment — appreciate you tracing the refcount/streak/LOCAL_SEGMENT_ID paths.

On the backend-wide outage scenario: you're right that under a systemic failure every remote name increments in lockstep, so after `kStaleSegmentFailureThreshold` cycles the whole remote cache invalidates at once. During the outage those descriptors can't be re-fetched, so we'd burn the warm cache we could otherwise ride out the outage on — strictly worse than keeping (possibly stale but still transferable) descriptors until the backend recovers.

The cheap guard fits well: skip the streak increment when the entire fetch set failed in a single cycle (systemic failure) vs. a per-segment disappearance. Happy to add it here before merge if you'd like, or land it as a follow-up — your call. If you'd rather keep the current fail-safe behavior, I'll document the mass-invalidation-on-outage intent in a comment and leave it.

alogfans (commented):
MetadataStoragePlugin::get() conflates a missing key with backend unavailability and transient errors. Two consecutive failures can therefore invalidate the entire remote cache during a metadata-service outage, making availability strictly worse. Please distinguish NOT_FOUND from UNAVAILABLE and advance the invalidation streak only for an authoritative absence; also add a backend-wide outage regression test. A simple “all fetches failed” guard is not sufficient for single-segment refreshes or when all keys were genuinely removed.

SuperMarioYL:
Thanks @alogfans for the careful review — you've identified the real weakness here. The current `MetadataStoragePlugin::get()` returns a plain `bool` and collapses both cases, which is exactly why the streak heuristic exists as a workaround rather than a proper fix. I'm looking at extending the plugin interface to surface `NOT_FOUND` vs `UNAVAILABLE` so the invalidation can be error-type-aware — advancing the streak only for an authoritative absence, not transient backend errors — plus the backend-wide outage regression test you described. Will follow up with the change once I have it ready.

SuperMarioYL:
Rebased onto current `main` to clear the merge conflict. The branch had fallen ~245 commits behind and the only conflict was in `transfer_metadata_test.cpp`: upstream's new validation/publication/handshake tests and this PR's stale-segment-invalidation tests were both added at the same insertion point, so git couldn't auto-merge them. Resolution kept both test groups intact (no logic changes) — the `transfer_metadata.h`/`.cpp` changes applied cleanly on top. CI is re-running; happy to address anything that comes up.

SuperMarioYL:
Quick CI note: the clang-format gate is green now (the format fix in b9a3a22 landed). The remaining `build-wheel` failures look like runner environment setup rather than this diff — the log errors out during `Failed to update package lists` / `Failed to install Go 1.25.9` / `Failed to build yalantinglibs` before reaching the transfer-engine build. Happy to address anything if a re-run still flags the diff.

catyans:
Following up on @alogfans's review: this is still unresolved on `b9a3a22`. `MetadataStoragePlugin::get()` still returns only `bool`, and `syncSegmentCache()` invalidates a live entry after two consecutive failures. The new transient-failure test injects only one failure, so two backend timeouts still look identical to an authoritative removal and purge the warm cache.

@SuperMarioYL, you mentioned an error-aware plugin result; is that patch still pending? Please add it, plus a consecutive-unavailable/backend-outage regression, before final review.

SuperMarioYL:
Thanks for the follow-up @catyans — the error-aware plugin result is now in f057e787.

**Interface change:** Added `GetResult` enum (`kFound` / `kNotFound` / `kUnavailable`) and a `getWithStatus()` virtual on `MetadataStoragePlugin`. The default delegates to `get()` and maps `false → kNotFound`, so plugins that cannot distinguish (etcd) preserve the current behavior — no regression. `get()` itself is unchanged.

**Overrides where the distinction is clear:**
- **Redis**: nil reply (`resp->str == nullptr`) → `kNotFound`; connection/command error → `kUnavailable`.
- **HTTP**: HTTP 404 → `kNotFound`; curl error / 5xx → `kUnavailable`.

**syncSegmentCache** now advances the invalidation streak **only** for `kNotFound` (authoritative key absence). Transient failures (`kUnavailable`) are counted as `unavailable_count` in the sync log and do **not** touch `segment_failure_counts_`, so a sustained backend outage can no longer purge the live cache.

**New regression test** `BackendOutageDoesNotInvalidateCache`: injects 5 consecutive transient failures (well past the threshold of 2), asserts the cached entry survives every cycle, then verifies it is still present after the outage clears.

Existing tests still hold:
- `ErasesStaleEntryAfterConsecutiveFailures` — `drop()` → `kNotFound` → streak advances → invalidated after 2.
- `TransientFailureIsRetained` — `failNext(1)` → `kUnavailable` → streak untouched → entry survives.

SuperMarioYL:
Rebased onto current `main` to clear the merge conflict. The branch had fallen ~159 commits behind.

The conflict was in `transfer_metadata.cpp`: upstream refactored `getSegmentDesc` into a public wrapper delegating to `getSegmentDescInternal(name, force_rpc_update)`, while this PR's redesign (f057e787) added a `GetResult *status` out-parameter to distinguish NOT_FOUND from UNAVAILABLE.

Resolution: rebound the `GetResult *status` parameter onto `getSegmentDescInternal` — the public `getSegmentDesc(name, status)` now delegates to `getSegmentDescInternal(name, false, status)`, and `syncSegmentCache` calls `getSegmentDescInternal(name, true, &status)`. The status-setting body logic (via `storage_plugin_->getWithStatus()`) is unchanged — only the signature was rebound to match upstream's refactor. No behavioral changes to the redesign.

The `GetResult` enum, `getWithStatus()` virtual, and the kNotFound-vs-kUnavailable branching in `syncSegmentCache` are all preserved.

SuperMarioYL:
Rebased onto latest main — the merge conflict re-appeared from base drift after the earlier resolve (upstream #3837 metadata-versioning landed touching the same `syncSegmentCache`).

The conflict was a single both-sides-additive declaration block: upstream's `std::vector<std::string> updated_names` (for the RPC-meta map invalidation) and our `size_t invalidated_count` (for the stale-remote-segment failure streak) were both added in the same spot — kept both, since each is consumed by its own (auto-merged) logic block. Every other hunk auto-merged cleanly, including the `GetResult`/`kNotFound`/`kUnavailable` distinction and the failure-streak invalidation, so no semantic replay was needed.

Confirmed our distinction is still additive — `GetResult`/`kNotFound`/`kUnavailable` do not appear upstream, so this isn't superseded. Force-pushed a clean rebase (`0ecc2ced`); `mergeable` is back to MERGEABLE.

staryxchen:
This PR and #3072 / #3944 look complementary rather than duplicate: both fix stale remote metadata in Transfer Engine, but from different sides.

- #3072 covers “peer is gone”: `syncSegmentCache` currently never invalidates, so a remote that unmounted or expired stays cached and keeps being used. The poller-side `kNotFound` vs `kUnavailable` distinction is the right way to drop that entry only on an authoritative absence.
- #3944 covers “peer was replaced”: on RDMA failure, mark remote metadata stale, keep the `SegmentID` for in-flight retries, and force the next lookup to reload — instead of waiting for the next poll.

Dumping the two diffs together would leave two invalidation models (erase vs mark-stale). Would you be open to co-working on one cache lifecycle instead? Something like Fresh / Stale / Gone, with the poller driving Gone only on authoritative `kNotFound`, and transport failures driving per-peer Stale (keep the ID, reload on next lookup, rather than invalidating every remote on any RDMA error). Happy to help sync on the design if that’s useful.

mooncake-transfer-engine/include/transfer_metadata_plugin.h
```diff
@@ -26,6 +26,16 @@ struct MetadataStoragePlugin {
...
+    // Error-aware lookup: same contract as get() but reports whether a
+    // non-found result is an authoritative key absence (kNotFound) or a
+    // transient backend failure (kUnavailable). Plugins that can tell the
+    // two apart (Redis nil reply, HTTP 404 vs 5xx) should override this;
+    // the default delegates to get() and maps false->kNotFound, preserving
+    // the pre-existing behavior for plugins that do not distinguish.
+    virtual GetResult getWithStatus(const std::string &key,
+                                    Json::Value &value) {
+        return get(key, value) ? GetResult::kFound : GetResult::kNotFound;
+    }
```
alogfans, line 41: This default mapping looks unsafe. For backends that do not override `getWithStatus()`, a plain `get() == false` may mean either an authoritative missing key or a transient backend failure. Mapping every false result to `kNotFound` allows `syncSegmentCache()` to treat backend unavailability as peer removal and eventually erase a still-valid remote segment cache entry.

This is especially risky because the etcd storage plugin does not appear to override `getWithStatus()` in this PR, so an etcd outage/timeout would still be interpreted as `kNotFound`.

Could we make the default conservative instead, e.g. map `false` to `kUnavailable`, and only return `kNotFound` from plugins that can prove authoritative absence? Alternatively, etcd should implement an error-aware `getWithStatus()` and this PR should include a regression test where consecutive etcd/backend-unavailable failures do not purge the cache.

catyans: I agree with @alogfans here. The Redis and HTTP overrides are a good improvement, but etcd is the default backend and both etcd implementations still inherit `false -> kNotFound`. In those implementations, `false` includes RPC/backend errors as well as an absent key, so two consecutive etcd outages can still evict a live descriptor—the exact availability failure this redesign is meant to remove.

@SuperMarioYL, “preserving the current behavior” isn't a safe compatibility fallback in this case because that behavior is the bug we're trying to fix. I think the base implementation should fail closed as `kUnavailable`, with `kNotFound` returned only by backends that can prove absence; ideally etcd should get an explicit error-aware override too. If the wrapper truly cannot expose absence separately, retaining stale cache entries is safer than deleting them on an ambiguous failure. The new outage regression should also exercise the inherited/default path, not only the fake override.

The overall stale-entry cleanup direction still looks useful, and @staryxchen's Fresh/Stale/Gone convergence with #3944 sounds like a sensible follow-up. I just don't think the primary backend ambiguity is closed on this head yet.

SuperMarioYL: Agreed — "preserving the current behavior" was preserving the bug for any backend that cannot distinguish the two cases. Fixed in d607a10b:

- **Default `getWithStatus()` now fails closed**: an ambiguous `false` maps to `kUnavailable`, never `kNotFound`, so `syncSegmentCache()` cannot evict on a backend it cannot prove absence for.
- **Both etcd implementations now override `getWithStatus()`**: the legacy `SyncClient` path treats an OK range response with an empty payload as `kNotFound` (etcd v3 answers a missing key with an OK response carrying no value), and the wrapper path treats `ret == 0` with no payload as `kNotFound` — matching `EtcdMetaStore`'s existing `InvalidEntry` interpretation of that combination. RPC errors, timeouts, and malformed payloads are `kUnavailable`.
- **New regression `DefaultPluginMappingFailsClosed`** exercises the inherited default path (a plugin that only implements the legacy `bool get()`), asserting a sustained opaque failure streak never invalidates the cached entry. It stays red until the fail-closed flip.

SuperMarioYL:
@staryxchen thanks for laying out the #3944 overlap so clearly — agreed that they're complementary: this PR is the poller side ("peer is gone", only on authoritative `kNotFound`), #3944 is the transport side ("peer was replaced", keep the ID and reload on next lookup).

Happy to co-work on a unified Fresh/Stale/Gone lifecycle as a follow-up. The `kNotFound`/`kUnavailable` distinction here is already the "Gone only on authoritative absence" half of that model, so this PR shouldn't conflict with the convergence — the per-peer Stale-on-RDMA-failure transition is the genuinely new part. Once this lands (or in a dedicated design issue, whichever you prefer) I'd be glad to sync on the cache-state machine.

alogfans:
Does the CI failure relate to this PR?

catyans (approved):
Thanks @SuperMarioYL—this addresses the ambiguity cleanly. LGTM.

The inherited/default path now fails closed as `kUnavailable`, while both etcd implementations only return `kNotFound` for a successful empty lookup. That means an opaque backend failure can no longer advance the eviction streak, and the new `DefaultPluginMappingFailsClosed` regression exercises exactly the path @alogfans and I were concerned about. Nice follow-through here.

On the current CI question: the ROCm 3.12 job reached the build and then exited 137 after the runner received a shutdown signal; the 3.10 sibling was canceled with it. The code-format, CTest, all TENT variants, CUDA/EFA wheels, and both Ubuntu test wheels passed, so this looks like runner infrastructure rather than a failure related to this diff. A rerun should be enough.

I also like the Fresh/Stale/Gone follow-up that @staryxchen proposed for convergence with #3944; keeping that lifecycle unification separate from this authoritative-absence fix seems reasonable.

SuperMarioYL:
@alogfans No — the failure is infra-side, not from this diff.

- `build-wheel-rocm (3.12)` died with `exit code 137` ("The runner has received a shutdown signal") at step 236/258 — the runner was killed mid-build, and there are no compile errors in the log. `(3.10)` was cancelled as a consequence.
- The same pattern shows on recent `main` commits merged today (e.g. 9b9e196d / #3938, 5accaaff / #3921): `CI Gate` red with the rocm builds cancelled, so it's not specific to this branch.
- Everything that actually compiles/exercises this PR's diff is green: `CTest unit tests (Python 3.12)`, `build-wheel` (CUDA), `build-wheel-cu13`, `build-wheel-efa`, `tent-ci`, Ruff and format checks.

A re-run of the rocm jobs should clear it.

alogfans (approved):
LGTM

PR #1562, opened 2026-02-26
[Store]Optimize uring file support for SSD offloading

zhangzuo21:
## Type of Change

* Types
  - [ ] Bug fix
  - [ ✅] New feature
    - [ ] Transfer Engine
    - [ ✅] Mooncake Store
    - [ ] Mooncake EP
    - [ ] Integration
    - [ ] P2P Store
    - [ ] Python Wheel
  - [ ] Breaking change
  - [ ] CI/CD
  - [ ] Documentation update
  - [ ] Other

## Checklist

- [ ✅] I have performed a self-review of my own code.
- [ ✅] I have formatted my own code using `./scripts/code_format.sh` before submitting.
- [ ] I have updated the documentation.
- [ ] I have added tests to prove my changes are effective.

## Summary

This PR contains three independent improvements to the offload (SSD-backed) path in Mooncake Store.

### 1. Per-thread io_uring ring (`uring_file.cpp`)

Previously each `UringFile` instance owned its own `io_uring` ring, which meant every file open/close incurred the overhead of `io_uring_queue_init` / `io_uring_queue_exit` system calls, and concurrent I/O across threads required mutex protection on the shared ring.

This PR replaces the per-file ring with a `thread_local SharedUringRing` singleton. Each thread initializes exactly one ring on first use and the ring lives for the lifetime of that thread. Key benefits:

- **Zero mutex contention**: each ring is exclusively owned by its thread; no locking is needed for I/O operations.
- **Drastically reduced syscall overhead**: ring init/exit no longer happens on every file open/close.
- **Better NVMe queue depth utilization**: within a single thread, multiple SQEs can be batched and submitted together (up to `QUEUE_DEPTH=32`), exposing deeper device parallelism.
- **Fixed-buffer registration**: the registered memory buffer is stored globally (`g_buf`) and lazily registered on each thread-local ring on first use, maintaining zero-copy I/O across all threads.

### 2. Early load buffer release via RPC (`real_client.cpp` — `batch_get_into_offload_object_internal`)

When an object is loaded from SSD and transferred from one client to another over the network, the sender holds a load buffer for that object until a GC lease TTL expires. Previously the buffer was not reclaimed until this timer fired.

This PR adds a fire-and-forget `release_offload_buffer` RPC call sent by the receiver immediately after the data transfer completes. The sender releases the load buffer upon receiving this RPC, reclaiming the memory much earlier than the GC deadline. If the RPC fails (e.g., network issue), the existing GC mechanism acts as a safety backstop.

[...]

PR #3073, opened 2026-07-23
[Store] Fix silent data corruption in io_uring read path (#3066)

### LINKED ISSUE #3066: [Bug]: Silent data corruption risk on Linux 5.15 when using io_uring offload due to broken short-read validation logic
## Environment
- Kernel version: `5.15.0-140-generic` (Ubuntu 22.04 Jammy LTS)
- Feature flag enabled: `MOONCAKE_OFFLOAD_USE_URING=true`
- Storage: Local NVMe disk with O_DIRECT enabled

## Symptom
Non-deterministic data precision / CRC mismatch errors occur randomly during bucket load operations. The issue cannot be reproduced stably, and no explicit IO error code is returned to upper-layer callers, making silent corruption invisible to users.

### Trigger Condition
The corruption appears when io_uring short-read events happen under Linux 5.15:
1. Short reads return successful CQE status without error codes
2. Unfinished trailing buffer regions retain stale dirty memory data
3. Truncated valid business payload leads to mismatched CRC / data inconsistency

## Root Cause Suspect & Code Reference
`mooncake-store/src/uring_file.cpp`, `mooncake-store/src/storage_backend.cpp`,
Line: 
https://github.com/kvcache-ai/Mooncake/blob/592f117822adde8991d4577545ae3a4b0694f870/mooncake-store/src/uring_file.cpp#L274,
https://github.com/kvcache-ai/Mooncake/blob/592f117822adde8991d4577545ae3a4b0694f870/mooncake-store/src/storage_backend.cpp#L1952

There is a confusing assignment logic after invoking `uring_file->read_aligned()`:

```C++
#ifdef USE_URING
          // Try to use read_aligned for O_DIRECT I/O if file is UringFile
          UringFile* uring_file = dynamic_cast<UringFile*>(file.get());
          if (uring_file != nullptr) {
              ... ...
              read_res = uring_file->read_aligned(
                  plan.dest_slice.ptr, aligned_size, aligned_offset);

              if (read_res) {
                  ... ...
                  read_res = plan.dest_slice.size;
              }
          } else
#endif
            ... ...

            if (read_res.value() != plan.dest_slice.size) {
                LOG(ERROR) << "Read size mismatch for key: " << plan.key
                           << ", expected: " << plan.dest_slice.size
                           << ", got: " << read_res.value();
                return tl::make_unexpected(ErrorCode::FILE_READ_FAIL);
            }
        }
    }
```
The variable `read_res` which stores the actual IO byte count returned by io_uring is overwritten by the expected full slice size.
All subsequent short-read validation logic based on `read_res` will always evaluate to `false`:

- The code cannot detect real short reads from the underlying io_uring layer
- No `ErrorCode::FILE_READ_FAIL` will be propagated to upstream
- Silent truncated data is passed to business logic without any alert

## Questions

1. What is the design intention of overwriting `read_res = plan.dest_slice.size` at line 1952? Why override the real IO read result returned by `read_aligned()`?
2. How does Mooncake guarantee data correctness for the io_uring offload path?
3. Are there existing safeguards (buffer cleanup, strict short-read rejection, mandatory retries) to avoid silent stale-buffer data corruption?

### Before submitting...

LujhCoconut:
## Description

Fix silent data corruption in `BucketStorageBackend::BatchLoad` io_uring path when short reads occur. (#3066)

Two bugs:

1. **`read_res` overwritten before validation** (`storage_backend.cpp`): after a successful `read_aligned`, the real IO byte count is replaced with the expected value, making the downstream short-read check dead code.
2. **No stale CQE filtering** (`uring_file.cpp`): `collect()` uses `io_uring_for_each_cqe` to consume all available CQEs regardless of which operation submitted them. Stale CQEs from a previously failed batch can contaminate the current byte count.

**Trigger chain:** 

`BatchLoad` → `read_aligned` → `submit_rw` → `collect()` → `io_uring_for_each_cqe` (consumes all CQEs, no `user_data` filter) → `read_res = plan.dest_slice.size` (overwrites real result) → downstream check dead → short read passes silently → truncated data returned to caller.

**Fix:**

- Add a `min_required` check before pointer adjustment: `read_res.value() >= offset_in_buffer + data_size`. Short reads fail with `FILE_READ_FAIL`.
- Tag each operation with a monotonic `op_id` via `sqe->user_data`, rewrite `collect()` to filter out CQEs with non-matching `user_data`. Uses peek-then-wait to avoid extra syscalls on the common path.

Both fixes are required together: the `user_data` filter prevents stale CQEs from making the byte count coincidentally pass the `min_required` check.

Zero-copy and `AllocateBatch` oversized allocation are unchanged.

Closes #3066.

## Module

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

**Test commands:**
```bash
# Bucket backend with io_uring, full verification
./storage_backend_bench --backend=bucket --test=load --use_uring --verify_rate=1 --num_operations=1000
```

## Checklist

- [ ] For changes >500 LOC: N/A (~70 lines)

## AI Assistance Disclosure

- [x] AI tools were used

Claude Code was used for code analysis and plan review. All changes reviewed and understood by the submitter.

### DIFF
--- a/mooncake-store/src/storage_backend.cpp
+++ b/mooncake-store/src/storage_backend.cpp
@@ -1945,10 +1945,34 @@ tl::expected<void, ErrorCode> BucketStorageBackend::BatchLoad(
                     plan.dest_slice.ptr, aligned_size, aligned_offset);
 
                 if (read_res) {
-                    // Adjust ptr to point to actual data start (no memcpy)
+                    // Verify the aligned read returned enough bytes
+                    // to cover the actual data region.  read_aligned
+                    // reads the full aligned range [aligned_offset,
+                    // aligned_end); the caller-visible data starts at
+                    // offset_in_buffer into that buffer and spans
+                    // plan.dest_slice.size bytes.
+                    size_t min_required =
+                        static_cast<size_t>(offset_in_buffer) +
+                        plan.dest_slice.size;
+                    if (read_res.value() < min_required) {
+                        LOG(ERROR)
+                            << "read_aligned short read for key: " << plan.key
+                            << ", bucket_id=" << plan.bucket_id
+                            << ", expected at least: " << min_required
+                            << " (aligned_size=" << aligned_size
+                            << ", data_size=" << plan.dest_slice.size << ")"
+                            << ", got: " << read_res.value();
+                        return tl::make_unexpected(ErrorCode::FILE_READ_FAIL);
+                    }
+                    // Adjust ptr to point to actual data start
+                    // (zero-copy: no memcpy, buffer was oversized by
+                    // AllocateBatch to accommodate the aligned read)
                     batch_object.at(plan.key).ptr =
                         static_cast<char*>(plan.dest_slice.ptr) +
                         offset_in_buffer;
+                    // Normalize read_res so the common validation
+                    // below passes.  The real short-read check was
+                    // already done above (min_required).
                     read_res = plan.dest_slice.size;
                 }
             } else
--- a/mooncake-store/src/uring_file.cpp
+++ b/mooncake-store/src/uring_file.cpp
@@ -144,6 +144,7 @@ class SharedUringRing {
     tl::expected<size_t, ErrorCode> batch_read(int fd, const ReadDesc* descs,
                                                int cnt) {
         ensure_buf_registered();
+        uint64_t op = ++op_id_;
         size_t total = 0;
         int remaining = cnt;
         int idx = 0;
@@ -162,9 +163,10 @@ class SharedUringRing {
                     io_uring_prep_read_fixed(sqe, fd, d.buf, d.len, d.off, 0);
                 else
                     io_uring_prep_read(sqe, fd, d.buf, d.len, d.off);
+                sqe->user_data = op;
             }
 
-            auto res = collect(batch);
+            auto res = collect(batch, op);
             if (!res) return res;
             total += res.value();
             idx += batch;
@@ -183,9 +185,11 @@ class SharedUringRing {
             LOG(ERROR) << "[SharedUringRing] SQ full (fsync)";
             return tl::make_unexpected(ErrorCode::INTERNAL_ERROR);
         }
+        uint64_t op = ++op_id_;
         io_uring_prep_fsync(sqe, fd, IORING_FSYNC_DATASYNC);
+        sqe->user_data = op;
 
-        auto res = collect(1);
+        auto res = collect(1, op);
         if (!res) return tl::make_unexpected(res.error());
         return {};
     }
@@ -253,8 +257,16 @@ class SharedUringRing {
         return value;
     }
 
-    // Drain exactly @expected CQEs and accumulate bytes.
-    tl::expected<size_t, ErrorCode> collect(int expected) {
+    // Drain exactly @expected CQEs matching @op_id and
+    // accumulate bytes.  Stale CQEs (user_data != op_id, e.g. from
+    // a previous batch_read / submit_rw that hit an error and left
+    // in-flight SQEs) are silently consumed and discarded.
+    //
+    // Uses peek-then-wait: after io_uring_submit_and_wait most CQEs
+    // are already available, so io_uring_peek_cqe() succeeds without
+    // a syscall in the common case.  io_uring_wait_cqe() only kicks
+    // in when the ring is unexpectedly drained (stale CQE storms).
+    tl::expected<size_t, ErrorCode> collect(int expected, uint64_t op_id) {
         int ret = io_uring_submit_and_wait(&ring_, expected);
         if (ret < 0) {
             LOG(ERROR) << "[SharedUringRing] io_uring_submit_and_wait: "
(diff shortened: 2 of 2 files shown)


mooncake-store/src/storage_backend.cpp
```diff
@@ -1945,10 +1945,34 @@ tl::expected<void, ErrorCode> BucketStorageBackend::BatchLoad(
...
+                    // Verify the aligned read returned enough bytes
+                    // to cover the actual data region.  read_aligned
+                    // reads the full aligned range [aligned_offset,
+                    // aligned_end); the caller-visible data starts at
+                    // offset_in_buffer into that buffer and spans
+                    // plan.dest_slice.size bytes.
+                    size_t min_required =
+                        static_cast<size_t>(offset_in_buffer) +
+                        plan.dest_slice.size;
+                    if (read_res.value() < min_required) {
```
fcczzz: `read_res` is the aggregate byte count returned by `submit_rw()`, but `read_aligned()` may split one aligned range into multiple independent SQEs.

Therefore, `read_res >= min_required` does not guarantee that the contiguous buffer range `[0, min_required)` has been populated.

For example, with `aligned_size = 8192`, `offset_in_buffer = 100`, and `data_size = 4096`, the read is split into two 4 KiB SQEs. If the first SQE returns 100 bytes and the second returns 4096 bytes, the aggregate result is 4196, so this check passes. However, most of the caller-visible range `[100, 4196)` is still uninitialized/stale.

Could we retain per-SQE metadata (requested length and buffer offset) and validate each completion, or compute the actual covered intervals and verify that they cover the full caller-visible data range?

LujhCoconut: Partially addressed: the defensive check is now in; the per-SQE metadata approach is rejected for current state.

Change the validation to require the FULL aligned range instead of the caller-visible minimum:

```cpp
read_res.value() != aligned_size  -> FILE_READ_FAIL
```

The fix rests on the 4096-alignment invariant chain, which makes this full-length check both sound and free of false rejections:

1. Bucket files are written by `WriteBucket()` zero-padded up to `align_up(total_size, kDirectIOAlignment)` (4096), so the file size is always a multiple of 4096.
2. The read range `[aligned_offset, aligned_end)` is 4096-aligned on both ends, and `aligned_end <= align_up(total_size) = file_size`, so EOF can never fall inside the read range.
3. `submit_rw()` chunks are all multiples of 4096 (`MIN_CHUNK = 4096`, power-of-two sizing), and no CQE can return more than requested — hence aggregate == requested total iff every SQE read its full chunk.

Given (2), any short read inside the range is genuinely anomalous, so rejecting it is the correct behavior. Given (3), this one-line check is equivalent to per-SQE validation: the "hole + compensation" scenario you described (an early SQE short read masked by tail padding) can no longer pass, because any short SQE makes the aggregate strictly less than `aligned_size`.

Per-SQE metadata tracking would buy nothing extra here while touching the shared hot I/O path (`submit_rw` / `collect`) and adding interface complexity. The aggregate-only semantics of `collect()` are still worth hardening separately (per-CQE length validation inside `submit_rw`, which would also cover `submit_vector` / `batch_read`), but that belongs in its own PR.

fcczzz: Thanks for the clarification. I agree that requiring `read_res == aligned_size` addresses my original aggregate-compensation concern, provided that the complete aligned range is always expected to be readable.
However, after tracing the current write path, that padding invariant does not appear to hold. `WriteBucket()` calls `BucketStorageBackend::OpenFile(..., FileMode::Write)`, while `OpenFile()` only returns a UringFile for `FileMode::Read`. Therefore, the `dynamic_cast<UringFile*>` in `WriteBucket()` is always null for this path, and writes take the `PosixFile::vector_write()` fallback, which writes exactly bucket_metadata->data_size without 4K padding.
This behavior predates this PR and appears intentional since #1562, which changed bucket writes back to buffered I/O specifically to avoid alignment padding. When uring is enabled, the current behavior is therefore an unpadded POSIX write followed by an aligned O_DIRECT/io_uring read.
For an object near EOF, the aligned read may legitimately return less than aligned_size while still covering the complete object payload. The new equality check would reject that valid read.

mooncake-store/src/uring_file.cpp
```diff
@@ -263,19 +275,34 @@ class SharedUringRing {
...
+        while (processed < expected) {
+            struct io_uring_cqe* cqe;
+            int wait_ret = io_uring_peek_cqe(&ring_, &cqe);
+            if (wait_ret == -EAGAIN) {
+                wait_ret = io_uring_wait_cqe(&ring_, &cqe);
+            }
+            if (wait_ret < 0) {
+                LOG(ERROR) << "[SharedUringRing] CQE wait error: "
+                           << strerror(-wait_ret);
+                return tl::make_unexpected(ErrorCode::INTERNAL_ERROR);
```
fcczzz, line 289: Returning here does not guarantee that the current operation has stopped. Its SQEs may already have been submitted, and a wait error such as -EINTR does not cancel them. The caller may then release or reuse the destination buffer while those reads are still in flight, allowing a late completion to overwrite memory belonging to a later operation.
Filtering the stale CQE on the next call only discards the notification; it does not undo the I/O's buffer write.

mooncake-store/src/uring_file.cpp
```diff
@@ -253,8 +257,16 @@ class SharedUringRing {
...
+    // accumulate bytes.  Stale CQEs (user_data != op_id, e.g. from
+    // a previous batch_read / submit_rw that hit an error and left
+    // in-flight SQEs) are silently consumed and discarded.
+    //
+    // Uses peek-then-wait: after io_uring_submit_and_wait most CQEs
+    // are already available, so io_uring_peek_cqe() succeeds without
+    // a syscall in the common case.  io_uring_wait_cqe() only kicks
+    // in when the ring is unexpectedly drained (stale CQE storms).
+    tl::expected<size_t, ErrorCode> collect(int expected, uint64_t op_id) {
         int ret = io_uring_submit_and_wait(&ring_, expected);
```
fcczzz, line 270: io_uring_submit_and_wait() returns the number of SQEs actually submitted, not a boolean success value. Treating every non-negative return as a full submission assumes that all expected SQEs belonging to the current op_id were submitted.
If fewer current SQEs are submitted—particularly when stale completions satisfy part of wait_nr—the loop below still waits for expected matching CQEs and can block forever waiting for completions that will never exist. If old unsubmitted SQEs are still present in the SQ, even ret == expected would not prove that all of them belong to the current operation.

LujhCoconut: Thanks. I agree with this point, but it's out of scope for this PR. Handling in-flight SQEs after a wait error requires more careful consideration — it should be addressed in a follow-up fix.

mooncake-store/src/uring_file.cpp
LujhCoconut: I agree — if a wait error occurs after submission, the in-flight SQEs are not cancelled, and a late CQE could theoretically write into a buffer that has already been released. That said, this is an existing issue (the old code had the same gap), the probability is extremely low in practice, and a proper fix (e.g., IORING_OP_ASYNC_CANCEL on the error path) would be invasive enough to warrant a separate design discussion. I'd prefer to track this as a known limitation and address it in a follow-up.

fcczzz (approved):

ykwd (approved):

PR #3077, opened 2026-07-23
Fix:  allocation_strategy not taking effect when enable_ha is enabled on master

gitgaoqian:
…visor config

When enable_ha=true, the --allocation_strategy flag was silently ignored and the allocation strategy always fell back to RANDOM.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
```

**Test results:**
- [ ] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

### DIFF
--- a/mooncake-store/include/master_config.h
+++ b/mooncake-store/include/master_config.h
@@ -187,6 +187,8 @@ class MasterServiceSupervisorConfig {
     std::string root_fs_dir = DEFAULT_ROOT_FS_DIR;
     int64_t global_file_segment_size = DEFAULT_GLOBAL_FILE_SEGMENT_SIZE;
     BufferAllocatorType memory_allocator = BufferAllocatorType::OFFSET;
+    AllocationStrategyType allocation_strategy_type =
+        AllocationStrategyType::RANDOM;
     uint64_t put_start_discard_timeout_sec = DEFAULT_PUT_START_DISCARD_TIMEOUT;
     uint64_t put_start_release_timeout_sec = DEFAULT_PUT_START_RELEASE_TIMEOUT;
     bool enable_disk_eviction = true;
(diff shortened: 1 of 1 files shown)


he-yufeng:
Traced the full chain before commenting: the supervisor config now parses and stores the enum, this PR adds the missing forward into WrappedMasterServiceConfig, and from there MasterServiceConfig picks it up (master_config.h:1160) and MasterService instantiates the strategy from it. That closes the HA gap end to end.

Coverage matches the existing non-HA block one for one: all five AllocationStrategyType values (types.h:448) are handled with the same warning-on-unknown fallback.

One nit, non-blocking: the string to enum conversion now lives in two places in this header (the supervisor config and MasterServiceConfig at ~line 561). A small shared helper would keep them from drifting the day a sixth strategy value shows up.

fcczzz (approved):
Confirmed that the issue exists in the HA path: allocation_strategy was not propagated through MasterServiceSupervisorConfig, so non-default strategies silently fell back to RANDOM.
This PR closes the configuration chain correctly, and the behavior is consistent with the non-HA path. The duplicated string-to-enum conversion is a minor maintainability concern; extracting a shared parser would be a nice follow-up, but I don’t consider it blocking.
A regression test covering the HA propagation path would also be helpful.
LGTM.

gitgaoqian:
@fcczzz Thanks for the review! I'll track the shared parser refactoring in a follow-up ticket and add the regression test for the HA path shortly.

@XucSh @YiXR @stmatengss @ykwd Could you please take a look when available? All checks have passed. Thanks!

ykwd (approved):
Thanks for the great work!

PR #3074, opened 2026-07-23
[TENT] Add mixed traffic workloads to tebench

catyans:
## Description

Add a benchmark-only mixed traffic contract to `tebench` so latency-sensitive foreground requests and bulk KV movement can run concurrently with different block sizes, batch sizes, TENT intents, deadlines, and QoS reporting thresholds.

This extends the QoS metrics baseline from #2845 and provides a concrete workload fixture for the validation matrix discussed in #2866. Existing block/batch sweeps and `--qos_classes` behavior remain unchanged.

Key changes:
- add `--workload_classes_json` with per-class transfer shape, intent, optional deadline, and QoS fields
- attach the selected class intent/deadline to each TENT request
- calculate class throughput from actual transferred bytes for heterogeneous request sizes
- preserve classic backend support for mixed sizes when intents/deadlines are unspecified
- add parser, validation, metric tests, and user documentation

## Module

- [x] Transfer Engine (`mooncake-transfer-engine`)
- [ ] Mooncake Store (`mooncake-store`)
- [ ] Mooncake EP (`mooncake-ep`)
- [ ] Mooncake PG (`mooncake-pg`)
- [ ] Integration (`mooncake-integration`)
- [ ] P2P Store (`mooncake-p2p-store`)
- [ ] Python Wheel (`mooncake-wheel`)
- [ ] Common (`mooncake-common`)
- [ ] Mooncake RL (`mooncake-rl`)
- [ ] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [x] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
cmake -S . -B build-codex-mixed -G Ninja \
  -DUSE_TENT=ON -DBUILD_BENCHMARK=ON -DBUILD_UNIT_TESTS=ON \
  -DBUILD_EXAMPLES=OFF -DUSE_CUDA=OFF -DUSE_HIP=OFF \
  -DCMAKE_BUILD_TYPE=Release
cmake --build build-codex-mixed --target tebench_qos_metrics_test tebench -j 16
ctest --test-dir build-codex-mixed -R tebench_qos_metrics_test --output-on-failure
```

**Remote H20 smoke:**
- Alibaba Cloud Linux host `lingjun-101`, Ubuntu 24.04 build container, GCC 13.3
- `USE_TENT=ON` build passed
[...]

Issue #2946, opened 2026-07-16
[RFC]: Store-to-TENT QoS Integration Boundary for Foreground Transfers and Background Offload


catyans:
## Summary

This RFC discusses how Mooncake Store traffic should participate in TENT QoS so that latency-sensitive transfers are protected from background offload traffic.

A representative serving workload has two priority classes:

- **HIGH**: foreground point-to-point KV handoff, such as Prefill-to-Decode (P2D), and foreground remote-prefix fetch.
- **LOW**: background writes to DRAM, local SSD, remote memory, or remote storage.

The main contention is on NIC TX and accelerator PCIe outbound bandwidth. A background Store write should not delay a foreground KV handoff that gates time-to-first-token. This RFC compares three integration boundaries:

1. **Option A — route all storage data movement through TENT.**
2. **Option B — use layered QoS and unified admission while retaining existing executors.**
3. **Option C — route only the network portion of remote Store traffic through TENT.**

The proposed rollout is **C first, then B if PCIe-side interference remains measurable**. Option A should remain a longer-term choice only if TENT is intentionally expanded into a general data-movement runtime.

This RFC is an integration proposal. It does not redefine the intent schema, QoS contract, budget scheduler, execution plan, or local admission queue proposed in #2865, #2856, #2866, #2863, and #2132. Instead, it asks how Mooncake Store should consume those mechanisms and where the ownership boundary should be.

## Motivation

Consider the following traffic classes:

| Traffic | Typical resources | Priority | Reason |
| --- | --- | --- | --- |
| Foreground P2D KV handoff | accelerator PCIe OUT + NIC TX | HIGH | decode cannot start until the KV state arrives |
| Foreground remote-prefix fetch | NIC RX + accelerator PCIe IN, depending on placement | HIGH | a hit avoids repeated prefill computation |
| DRAM offload | accelerator PCIe OUT + memory bandwidth | LOW | background cache management |
| Local SSD write | accelerator PCIe OUT + memory bandwidth + NVMe | LOW | background persistence/eviction |
| Remote Store write | accelerator PCIe OUT, depending on GDR/staging, + NIC TX | LOW | background persistence/write-through |

Foreground prefix fetch often uses the opposite direction from background writes, so it should remain concurrent when the resources do not conflict. The scheduler must therefore reason about resources and direction; a single global HIGH/LOW gate would unnecessarily serialize independent RX/IN and TX/OUT work.

The required behavior is cooperative preemption:

```text
[...]

Issue #2865, opened 2026-07-12
[RFC]: TENT Transfer Intent API v1 and Cross-Language Compatibility


catyans:
## Summary

This RFC proposes a stable TENT Transfer Intent API v1 and explicit semantic parity across the C++, C, pybind, and high-level Python interfaces.

TENT already has the core intent fields in its internal C++ `Request`:

- `priority`
- `deadline_ns`
- `policy_name`
- `transport_hint`
- `intent_type`

However, these fields were added incrementally and are not exposed consistently:

- C++ `Request` contains all fields.
- pybind exposes the fields added by #2810 and earlier work.
- the C `tent_request_t` currently contains `priority` and `transport_hint`, but not `deadline_ns`, `policy_name`, or `intent_type`.
- the high-level declarative Python API is being developed separately in #2541.
- upper-layer connectors do not yet share one versioned contract for defaults, unknown values, clock domain, authorization, and compatibility.

The RFC defines what an application is allowed to request and how TENT interprets that request. It does not define how TENT constructs an Execution Plan (#2863), how QoS is enforced (#2856), or how operators inspect the result (#2864).

```text
Application / Store / SGLang / vLLM
                 |
                 v
        Transfer Intent API v1
                 |
       validation + normalization
                 |
       effective policy resolution
                 |
                 v
         Execution Plan (#2863)
```

The first implementation milestone should add a versioned C request ABI, shared validation/normalization, and cross-language compatibility tests. It must not change default runtime scheduling behavior.

## Motivation

The TENT roadmap in #1058 calls for a clean declarative API where users express what data should move and the runtime selects the path. The existing fields are a strong foundation, but an API is not stable merely because fields exist in a C++ struct.

### 1. Cross-language behavior is incomplete

The current C API converts only the fields present in `tent_request_t`. A C caller cannot express the same intent as a C++ or pybind caller. This makes connector behavior depend on the language binding rather than on the request semantics.

### 2. The existing C struct is not safely extensible as an array ABI

`tent_submit()` accepts an array of fixed-size `tent_request_t`. Appending fields changes the array stride expected by a newly compiled library. A new library reading a request array produced by an old binary can step beyond the old element boundary. Zero-initialization alone does not solve binary layout/stride compatibility.

The current struct must remain supported. A new versioned request entry and submission API are required instead of silently extending the old array ABI.

### 3. Zero-initialization does not provide a safe priority default

[...]

PR #3079, opened 2026-07-23
[Store][TENT] Add transfer intent and path observability

catyans:
## Summary

This is the behavior-preserving Phase 0 slice of #2946. It makes Store transfer intent and the actually executed data path observable before any Store-to-TENT scheduling policy is enabled.

- adds a bounded `StoreTransferIntent` context with foreground get, background prefetch, migration, weight loading, and unspecified values;
- normalizes existing READ calls to `foreground_get` while preserving explicit caller intent;
- records Store bytes by `(intent, path)` for Transfer Engine, local memcpy, NVMe-oF, file, and unknown paths;
- records `mooncake_store_unmanaged_transfer_bytes{intent=...}` for bytes outside Transfer Engine control;
- keeps every existing call source-compatible through default arguments;
- does not alter transfer selection, priority, QP choice, admission, or bandwidth behavior.

The bounded vocabulary follows the connector direction in #2865 and intentionally excludes tenant/trace IDs from metric labels.

## Why first

Without path attribution, a future Store QoS result could incorrectly claim protection for local memcpy, file, NVMe-oF, or opaque staging traffic. These metrics establish the enforcement boundary needed by #2946 before enabling LOW/HIGH policy.

## Remote validation

Lingjun H20 `lingjun-101`, fresh branch clone, GCC 13.3, Release-like CMake build, `USE_CUDA=OFF`:

- `transfer_task_test`: **7/7 passed**
- `client_metrics_test`: **15/15 passed**
- total targeted tests: **22/22 passed**

The build ran in a task-owned one-shot dependency container; no existing containers, images, volumes, or networks were changed.

Related: #2946, #2865, #3074.

### DIFF
--- a/mooncake-store/include/client_metric.h
+++ b/mooncake-store/include/client_metric.h
@@ -103,11 +103,22 @@ Result execute_timed_operation(Operation&& operation, SuccessFn&& success_fn,
 enum class TransferOperationKind { kRead, kWrite };
 
 struct TransferMetric {
+    std::array<std::string, 2> transfer_path_label_names = {"intent", "path"};
+    std::array<std::string, 1> transfer_intent_label_names = {"intent"};
+
     TransferMetric(std::map<std::string, std::string> labels = {})
         : total_read_bytes("mooncake_transfer_read_bytes", "Total bytes read",
                            labels),
           total_write_bytes("mooncake_transfer_write_bytes",
                             "Total bytes written", labels),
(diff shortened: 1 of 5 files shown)


catyans:
Marking ready for review: the behavior-preserving observability slice is covered by 22/22 targeted tests on the H20 smoke node, format/docs and the completed architecture builds are green, and the remaining build jobs are still queued/running.

PR #3080, opened 2026-07-23
[Store] Fix mem storage display showing 16777216 TB when no segment i…

leonzzhu:
## Description

When no segment has been mounted, the master admin metrics log shows:

```
Mem Storage: 16777216.00 TB / 0 B
```

This happens because `ylt::metric::gauge_t` defaults to `-1` (int64_t) when
the gauge has never been updated. Passing this negative value through
`byte_size_to_string(uint64_t)` causes an implicit `int64_t → uint64_t`
conversion where `-1` becomes `0xFFFFFFFFFFFFFFFF` = `16777216.00 TB`.

**Fix:**

1. **`get_summary_string()`**: clamp all gauge capacity/allocated values to
   non-negative before formatting, so a never-set gauge displays as `0`
   instead of garbage.

2. **`byte_size_to_string()`**: add explicit `bytes == 0` early return to
   avoid the `>= TB` branch rendering zero as `"0.00 TB"`.

After fix:
```
Mem Storage: 0 B / 0 B
```

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
# Verified by starting a fresh master with no segments mounted,
# checking admin metrics log output before and after the fix.
#
# Before:
#   Mem Storage: 16777216.00 TB / 0 B
#
# After:
#   Mem Storage: 0 B / 0 B
```

**Test results:**
- [ ] Unit tests pass
- [x] Manual testing done — verified on a fresh master deployment with no segments mounted

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

Codebuddy was used to help trace the root cause and draft the fix.
The human submitter reviewed and tested all changes on real hardware.

### DIFF
--- a/mooncake-store/src/serialize/serializer.cpp
+++ b/mooncake-store/src/serialize/serializer.cpp
@@ -1000,6 +1000,17 @@ auto Serializer<OffsetBufferAllocator>::deserialize(const msgpack::object &obj)
         allocator->offset_allocator_ = offset_allocator_result.value();
         allocator->cur_size_ = cur_size;
 
+        // The snapshot restores cur_size_ directly from persisted data
+        // without going through the live allocate()/adoptImportedBuffer()
+        // paths, so no inc_allocated_mem_size() was paired with it. The
+        // allocator destructor still calls dec_allocated_mem_size(cur_size_)
+        // to undo its contribution to the global metric; without a matching
+        // inc the gauge would go negative and wrap to ~16M TB when formatted
+        // as uint64. Pair it here so the accounting stays symmetric and the
+        // gauge ends at 0 after this (often throwaway) allocator is destroyed.
+        MasterMetricManager::instance().inc_allocated_mem_size(
+            segment_name, static_cast<int64_t>(cur_size));
+
         return allocator;
     } catch (const std::exception &e) {
         return tl::unexpected(SerializationError(


LujhCoconut (approved):
LGTM.

mooncake-store/include/utils.h
```diff
@@ -150,6 +150,10 @@ std::string expected_to_str(const tl::expected<T, ErrorCode>& expected) {
 
     std::ostringstream oss;
     oss << std::fixed << std::setprecision(2);
+    if (bytes == 0) {
+        oss << "0 B";
+        return oss.str();
+    }
```
Lin-z-w: This special case appears redundant. The existing < 1 KB branch already formats zero as "0 B", and UtilsTest.ByteSizeToString already verifies that behavior. Unless this is needed for a separate semantic reason, it can be omitted.

Lin-z-w (commented):
The signed-to-unsigned conversion explains the huge displayed value, but the patch currently masks the negative metric rather than fixing why the accounting became negative. Please address the reconstruction/decrement mismatch and add a regression test.

Lin-z-w:
I also smoke-tested the pre-fix master from a fresh root directory with snapshot restore and HA disabled, without mounting any segments. Both periodic logs and /metrics/summary reported Mem Storage: 0 B / 0 B, while /metrics reported master_allocated_bytes 0 and master_total_capacity_bytes 0. Therefore, the issue is not reproducible from an untouched gauge or a normal empty master. Please provide the actual reproduction sequence and fix the path that drives the metric negative instead of clamping the output.

leonzzhu:
> I also smoke-tested the pre-fix master from a fresh root directory with snapshot restore and HA disabled, without mounting any segments. Both periodic logs and /metrics/summary reported Mem Storage: 0 B / 0 B, while /metrics reported master_allocated_bytes 0 and master_total_capacity_bytes 0. Therefore, the issue is not reproducible from an untouched gauge or a normal empty master. Please provide the actual reproduction sequence and fix the path that drives the metric negative instead of clamping the output.

Sorry for the late reply.

My scenario is HA mode, and the bug shows up when I start up a standby node.
I added some debug prints and managed to reproduce it. Here are the key
lines from the standby startup log (IPs redacted):

I0730 11:27:51.905532    42 master.cpp:950] Starting C++ HTTP metadata server on 0.0.0.0:8080
I0730 11:27:51.907295    42 master.cpp:959] C++ HTTP metadata server started successfully
I0730 11:27:52.910202    42 rpc_service.cpp:299] Master admin server started on port 9003
I0730 11:27:52.910238    42 master_service_supervisor.cpp:124] Master runtime state -> starting, role=standby
I0730 11:27:52.911669    55 rpc_service.cpp:290] Master Admin Metrics: role=standby, state=starting, service_ready=false, master={Mem Storage: 0 B / 0 B ...}
I0730 11:27:53.381217    42 standby_controller.cpp:151] [StandbyCtrl] StartStandby called, observed_leader=nullopt, enable_snapshot_restore=1
I0730 11:27:53.383056    42 oplog_applier.cpp:169] OpLogApplier: recovered from sequence_id=0, expected_sequence_id set to=1
I0730 11:27:53.512856    42 allocator.cpp:199] initializing_offset_buffer_allocator segment_name=<segment-ip>:<port> base_address=0x... size=60397977600
.....
W0730 11:27:53.557933    42 master_metric_manager.cpp:563] [DEBUG-MEM-METRIC] dec_allocated_mem_size segment=<segment-ip>:<port> val=12288 val_as_u64=12288 gauge_before=0 would_underflow=YES negative=no per_segment_before=0
I0730 11:27:53.559439    42 hot_standby_service.cpp:289] Loaded snapshot baseline: snapshot_id=20260730_112401_759, snapshot_seq_id=0, keys=0
    ...
I0730 11:28:02.911878 rpc_service.cpp:290] Master Admin Metrics: role=standby, state=standby, service_ready=false, master={Mem Storage: 16777216.00 TB / 0 B ...}

This debug print is emitted right when the snapshot has finished loading:
the value is subtracted inside the destructor of the throwaway allocator
used to read the snapshot. As the log shows, the gauge was at 0 before the
dec, and we dec by 12288, so it goes to -12288. That negative int64 is then
passed into byte_size_to_string(uint64_t), which implicitly converts it to
uint64 and wraps it to 16777216.00 TB.

Root cause: Serializer<OffsetBufferAllocator>::deserialize() restores
cur_size_ straight from the persisted snapshot without going through the
live allocate()/adoptImportedBuffer() paths, so no inc_allocated_mem_size()
is paired with it. The (throwaway) allocator is destroyed right after the
snapshot is loaded, and its destructor calls dec_allocated_mem_size(cur_size_),
driving the global gauge negative. The live restore path
(RestoreOffsetBufferAllocator) goes through allocate() which already inc()s,
so this does not double-count.

Fix: pair inc_allocated_mem_size(segment, cur_size_) where cur_size_ is
restored in deserialize(), so the inc/dec stays symmetric and the gauge ends
at 0 after the temporary allocator is destroyed. I dropped the earlier
display-only clamp in get_summary_string() since it only masked the symptom.

Would appreciate your review. Thanks!

Lin-z-w (approved):
Thanks for the detailed reproduction. I verified the HA standby snapshot path and ran an A/B check with cur_size_=12288: the base commit leaves the metric at -12288, while the latest commit correctly returns it to 0. The fix looks correct to me.

ykwd (approved):

PR #3090, opened 2026-07-24
[TE] Make RDMA dedicated poller optional and guard post/poll handoff

alogfans:
## Description

This PR addresses a correctness risk introduced by #2696, which separated post-send workers from the CQ polling worker. In that mode, worker 0 polls CQEs while workers 1..N post sends. A fast CQE can be processed by the poller while the posting worker is still inside `submitPostSend()` and has not finished local bookkeeping, leaving shared slice state exposed to a race.

This change restores the historical conservative worker model by default:

```text
default:
  every RDMA worker posts its shard and polls CQs
```

The separated mode is now opt-in:

```text
MC_RDMA_DEDICATED_POLLER=1:
  worker 0 polls only
  workers 1..N post only
```

For the opt-in dedicated-poller mode, this PR also keeps a lightweight striped lock around the same-peer post/completion handoff so that completion handling cannot race with post-send local ownership cleanup for the same peer.

**Changes**

- Add `GlobalConfig::rdma_dedicated_poller`, default `false`.
- Add env config `MC_RDMA_DEDICATED_POLLER`.
- Restore default historical worker scheduling: all RDMA workers can post and poll.
- Preserve the previous separated worker mode behind the new opt-in flag.
- Replace per-peer mutex map with fixed striped locks to avoid unbounded growth and hot-path map locking.
- Add a hardware-free unit test for the striped post/poll lock behavior.

**Why**

The previous worker split changed an implicit ordering guarantee: post-send and CQ polling for a slice could now happen on different threads. Although a CQE is only produced after a WR is posted, the poll worker could still process the completion before the post worker finished local queue cleanup and ownership handoff. The default mode now avoids that risk by restoring the original scheduling model, while keeping the separated mode available for controlled validation.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
```

**Test results:**
- [X] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used
- [X] AI tools were used (specify below) CodeX with manual check


mooncake-transfer-engine/src/transport/rdma_transport/worker_pool.cpp
```diff
@@ -67,17 +67,19 @@ static int selectPeerDevice(RdmaTransport::SegmentDesc *peer_segment_desc,
...
 
 static bool workerCanPost(int thread_id) {
-    return kTransferWorkerCount == 1 || thread_id != 0;
+    return kTransferWorkerCount == 1 || !globalConfig().rdma_dedicated_poller ||
+           thread_id != 0;
 }
 
 static bool workerCanPoll(int thread_id) {
-    return kTransferWorkerCount == 1 || thread_id == 0;
+    return kTransferWorkerCount == 1 || !globalConfig().rdma_dedicated_poller ||
```
staryxchen, line 75: **Default mode races on `ibv_poll_cq`**

With `rdma_dedicated_poller=false`, every worker can poll, but `performPollCq` still walks **all** CQs (`for (cq_index = 0; ...)` → `context_.poll` / `ibv_poll_cq`) without restoring the pre-#2696 thread sharding:

```cpp
// historical (pre-#2696)
for (int cq_index = thread_id; cq_index < context_.cqCount();
     cq_index += kTransferWorkerCount)
```

#2696 removed that sharding because only worker 0 polled. Re-enabling multi-worker polling without restoring it means, under defaults (`MC_WORKERS_PER_CTX=2`, `MC_NUM_CQ_PER_CTX=1`), multiple threads call `ibv_poll_cq` on the same CQ concurrently — undefined for verbs, and can double-process or drop CQEs.

This also does not match the PR claim of restoring the historical model (historical = post + **sharded** poll).

**Suggestion:** keep dedicated poller as the default, or when `rdma_dedicated_poller=false` restore CQ sharding in `performPollCq` / `hasOutstandingCq`.

mooncake-transfer-engine/src/transport/rdma_transport/worker_pool.cpp
```diff
@@ -561,6 +578,8 @@ void WorkerPool::performPollCq(int thread_id) {
         for (int i = 0; i < nr_poll; ++i) {
             Transport::Slice *slice = (Transport::Slice *)wc[i].wr_id;
             assert(slice);
+            std::lock_guard<std::mutex> lock(
+                postPollMutexFor(slice->peer_nic_path));
```
staryxchen: **Lock held across heavy completion work**

This stripe lock is held for the entire completion handling path below (including `markRailFailed`, `deleteEndpointByPtr`, and error logging), not just the post/completion ownership handoff. Same-stripe `submitPostSend` is then blocked behind endpoint teardown / logging.

**Suggestion:** keep the critical section tight around the slice/local-ownership handoff that must not race with post-send, then do failure handling after unlocking.

mooncake-transfer-engine/src/config.cpp
```diff
@@ -259,6 +259,14 @@ void loadGlobalConfig(GlobalConfig& config) {
                 << "Ignore value from environment variable MC_WORKERS_PER_CTX";
     }
 
+    const char* rdma_dedicated_poller_env =
+        std::getenv("MC_RDMA_DEDICATED_POLLER");
+    if (rdma_dedicated_poller_env) {
+        parseBoolConfigEnv(rdma_dedicated_poller_env,
+                           "MC_RDMA_DEDICATED_POLLER",
```
staryxchen, line 266: **Docs / ops gap**

`MC_RDMA_DEDICATED_POLLER` changes default RDMA worker scheduling (and interacts with the #2696 dedicated-poller perf fix), but it is only wired into `loadGlobalConfig` / `dumpGlobalConfig` — there is no user-facing documentation of the flag, its default, or when operators should flip it.

**Suggestion:** document the env var (behavior, default, relationship to #2696 / `MC_WORKERS_PER_CTX`), and call out the chosen default explicitly in release notes / TE config docs.

staryxchen (commented):
Inline comments on three review findings: concurrent CQ polling in default mode, lock scope on the completion path, and missing docs/ops coverage for the new env flag.

mooncake-transfer-engine/src/transport/rdma_transport/worker_pool.cpp
alogfans: Agreed. I kept this PR focused on restoring the historical CQ sharding semantics so the default mode does not concurrently poll the same CQ. The completion-path lock scope is a separate behavior/risk change, so I left it unchanged here and can address it in a follow-up PR.

mooncake-transfer-engine/src/config.cpp
alogfans: Fixed.

mooncake-transfer-engine/src/transport/rdma_transport/worker_pool.cpp
alogfans: Fixed.

staryxchen (approved):
LGTM

PR #3096, opened 2026-07-24
[CI/Build] Add ARM64 Non-CUDA release wheels

Aionw:
## Description

Add ARM64/aarch64 release support for the `mooncake-transfer-engine-non-cuda` Python wheels.

- Add production and pre-release ARM64 Non-CUDA jobs through the existing `_build-wheel.yaml` reusable workflow.
- Use `ubuntu-22.04-arm`, `pytorch/manylinuxaarch64-builder:cuda12.8`, Ninja, and the `NON_CUDA_BUILD` variant.
- Preserve HTTP, transfer-engine etcd, and Store etcd support while keeping `USE_CUDA=OFF` and `WITH_EP=OFF`.
- Disable only the ARM Rust Store bindings with `WITH_STORE_RUST=OFF` because of the known ARM link-ordering issue; fixing Rust bindings is outside this PR.
- Make production publishing wait for both architectures and use a Non-CUDA-specific artifact pattern.
- Skip both production Non-CUDA build jobs for pre-release tags, which are handled by `pre-release.yaml`.
- Make pre-release validation require the new ARM64 artifacts and raise the expected wheel minimum from 20 to 24.

No open issue or PR specifically implementing ARM64 Non-CUDA wheels was found before opening this PR. #2858 is indirectly related because it reports pip falling back to broken legacy sdists when no compatible wheel is available; this change expands ARM wheel coverage but does not fix or remove those stale sdists. The manylinux ARM job follows the architecture/container pattern established by #2968.

## Module

- [x] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

The changed workflows were parsed as YAML, pre-commit passed on both touched files, and the final diff passed Git whitespace validation. A real ARM wheel build was not run locally because this checkout does not provide an ARM runner; the new pre-release job is the runtime validation path before a production tag.

**Test commands:**
```bash
python3 -c 'import pathlib, yaml; paths=[pathlib.Path(".github/workflows/release-non-cuda.yaml"), pathlib.Path(".github/workflows/pre-release.yaml")]; [yaml.safe_load(path.read_text()) for path in paths]'
pre-commit run --files .github/workflows/release-non-cuda.yaml .github/workflows/pre-release.yaml
git diff --check
```

**Test results:**
- [ ] Unit tests pass (not applicable; workflow-only change)
- [ ] Integration tests pass (requires an ARM pre-release workflow run)
- [ ] Manual testing done (no local ARM build was available)

## Checklist

- [ ] I have performed a self-review of my own code (human submitter review required before marking ready)
- [x] I have formatted my code using `./scripts/code_format.sh` (run by pre-commit)
- [ ] I have run `pre-commit run --all-files` and all hooks pass (scoped touched-file run passed)
- [ ] I have updated the documentation (not applicable; no existing support statement became false)
- [ ] I have added tests to prove my changes are effective (not applicable; validation requires an ARM workflow run)
- [ ] For changes >500 LOC: I have filed an RFC issue (not applicable; 41 insertions / 5 deletions)

## AI Assistance Disclosure

- [ ] No AI tools were used

OpenAI Codex inspected repository history and policies, implemented the workflow and guard changes, ran validation, and prepared this draft PR. The human submitter must review every changed line and be able to defend the change end-to-end before marking the PR ready.

### DIFF
--- a/.github/workflows/pre-release.yaml
+++ b/.github/workflows/pre-release.yaml
@@ -41,6 +41,20 @@ jobs:
         -DSTORE_USE_ETCD=ON -DENABLE_SCCACHE=ON -DCMAKE_BUILD_TYPE=Release
       artifact-prefix: mooncake-wheel-non-cuda-pre-release
 
+  build-non-cuda-arm64:
+    uses: ./.github/workflows/_build-wheel.yaml
+    with:
+      runner: ubuntu-22.04-arm
+      container: pytorch/manylinuxaarch64-builder:cuda12.8
+      cuda: container
+      cmake-generator: Ninja
+      variant-flag: NON_CUDA_BUILD
+      cmake-args: >-
+        -DBUILD_UNIT_TESTS=OFF -DUSE_HTTP=ON -DUSE_ETCD=ON -DUSE_CUDA=OFF -DWITH_EP=OFF
+        -DSTORE_USE_ETCD=ON -DWITH_STORE_RUST=OFF -DENABLE_SCCACHE=ON
+        -DCMAKE_BUILD_TYPE=Release
+      artifact-prefix: mooncake-wheel-non-cuda-arm64-pre-release
+
   build-cuda13:
     uses: ./.github/workflows/_build-wheel.yaml
     with:
@@ -85,7 +99,13 @@ jobs:
 
   validate-release:
     name: Validate release artifacts
-    needs: [build-cuda, build-non-cuda, build-cuda13, build-cuda-arm64, build-cuda13-arm64]
+    needs:
+      - build-cuda
+      - build-non-cuda
+      - build-non-cuda-arm64
+      - build-cuda13
+      - build-cuda-arm64
+      - build-cuda13-arm64
     runs-on: ubuntu-22.04
     permissions:
       contents: read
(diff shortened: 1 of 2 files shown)


ykwd (approved):

Icedcoco (dismissed):
Thanks for adding the ARM64 Non-CUDA wheel path. I noticed two things and wanted to get your thoughts:

1. Both new ARM64 jobs pass `build-with-ep: '0'`, but #3095 removed that input from `_build-wheel.yaml` on the current `main`. Could this cause reusable-workflow validation to fail after the PR is merged? Would it make sense to rebase and remove these two entries?

2. `release-non-cuda.yaml` matches all `v*` tags, while only `publish-release` excludes pre-release tags. As a result, an RC/pre-release tag would build the same four ARM64 Non-CUDA wheels both here and in `pre-release.yaml`. Is this intentional? If not, should the production build jobs use the same pre-release guard as `release.yaml`?

Icedcoco (approved):
LGTM. The rebased workflow changes look good to me, and the earlier concerns have been addressed.

PR #3101, opened 2026-07-24
[Bugfix][Store] Update bucket timestamp after offload to avoid immedi…

mjwtom:
…ate eviction (#3097)

BatchOffload now updates the bucket's last_access_ns_ with the current time when committing metadata under LRU eviction policy. This prevents freshly offloaded buckets from being evicted immediately due to having a zero timestamp.

## Module

- [x] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

配置
export MOONCAKE_OFFLOAD_FILE_STORAGE_PATH=xxx
export MOONCAKE_OFFLOAD_STORAGE_BACKEND_DESCRIPTOR=bucket_storage_backend
export MOONCAKE_OFFLOAD_BUCKET_EVICTION_POLICY=lru

**Test commands:**
```bash
```

**Test results:**
- [x] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

GLM-5

### DIFF
--- a/mooncake-store/src/storage_backend.cpp
+++ b/mooncake-store/src/storage_backend.cpp
@@ -1739,6 +1739,7 @@ tl::expected<int64_t, ErrorCode> BucketStorageBackend::BatchOffload(
         ReleasePreparedWrite(pending);
         return tl::make_unexpected(write_bucket_result.error());
     }
+    VLOG(1) << "Written bucket with id: " << bucket_id;
     // Save a copy of bucket->keys before std::move(bucket) into buckets_
     // consumes the shared_ptr. Needed for complete_handler and rollback.
     const auto bucket_keys = bucket->keys;
@@ -1772,8 +1773,16 @@ tl::expected<int64_t, ErrorCode> BucketStorageBackend::BatchOffload(
                 CHECK(inserted)
                     << "Reserved key became duplicated: " << bucket_keys[i];
             }
+            auto ts = 0LL;
+            // Update LRU timestamp for in case of eviction.
+            if (bucket_backend_config_.eviction_policy ==
+                BucketEvictionPolicy::LRU) {
+                ts =
+                    std::chrono::steady_clock::now().time_since_epoch().count();
+                bucket->last_access_ns_.store(ts, std::memory_order_relaxed);
+            }
             buckets_.emplace(bucket_id, std::move(bucket));
-            lru_index_.emplace(0LL, bucket_id);
+            lru_index_.emplace(ts, bucket_id);
         }
         if (duplicate_found) {
             LOG(ERROR) << "Reserved key became duplicated before commit, "
@@ -2676,7 +2685,9 @@ void BucketStorageBackend::RollbackCommittedBucket(
 
         // Remove bucket metadata
         total_size_ -= bucket_meta->meta_size;
-        lru_index_.erase({0LL, bucket_id});
+        lru_index_.erase(
+            {bucket_meta->last_access_ns_.load(std::memory_order_relaxed),
+             bucket_id});
         buckets_.erase(bucket_it);
     }
 
@@ -3072,6 +3083,8 @@ tl::expected<void, ErrorCode> BucketStorageBackend::FinalizeEviction(
         if (bucket_cleanup_failed) {
             cleanup_failed_count++;
         }
+        VLOG(1) << "Evicted bucket with id: " << bucket_id
+                << ", policy: " << bucket_backend_config_.eviction_policy;
     }
     if (!pending.buckets.empty()) {
         LOG(INFO) << "[Evict] finalized: attempted=" << pending.buckets.size()
--- a/mooncake-store/include/storage_backend.h
+++ b/mooncake-store/include/storage_backend.h
@@ -186,6 +186,20 @@ enum class BucketEvictionPolicy {
     LRU,   // Evict least recently read bucket first
 };
 
+inline std::ostream& operator<<(std::ostream& os,
+                                const BucketEvictionPolicy& policy) {
+    switch (policy) {
+        case BucketEvictionPolicy::NONE:
+            return os << "none";
+        case BucketEvictionPolicy::FIFO:
+            return os << "fifo";
+        case BucketEvictionPolicy::LRU:
+            return os << "lru";
+        default:
+            return os << "unknown";
+    }
+}
+
 struct BucketBackendConfig {
     int64_t bucket_size_limit =
         256 * kMB;  // Max total size of a single bucket (256 MB)
@@ -1085,6 +1099,12 @@ class BucketStorageBackend : public StorageBackendInterface {
      */
     tl::expected<std::shared_ptr<StorageFile>, ErrorCode> GetFileInstance()
         const;
+
+    // Test-only: number of entries in the LRU eviction index.
+    size_t GetLruIndexSizeForTest() const {
+        SharedMutexLocker lock(&mutex_, shared_lock);
+        return lru_index_.size();
+    }
 
    private:
     // Alignment helper functions for O_DIRECT I/O
--- a/mooncake-store/tests/storage_backend_test.cpp
+++ b/mooncake-store/tests/storage_backend_test.cpp
@@ -4645,4 +4645,114 @@ TEST_F(StorageBackendTest,
               std::make_optional(value_b));
 }
 
+TEST_F(StorageBackendTest,
+       BucketStorageBackend_OffloadSetsBucketAccessTimestamp) {
+    // Regression test for "update bucket ts after offload": BatchOffload must
+    // stamp the new bucket's last_access_ns_ with the current time when
+    // committing metadata. The timestamp itself is private, so we verify it
+    // through LRU eviction behavior:
+    //   - Bucket A is offloaded and then read, so its timestamp is t1 > 0.
+    //   - Bucket B is offloaded afterwards; with the fix its timestamp is
+    //     t2 > t1, without the fix it stays 0.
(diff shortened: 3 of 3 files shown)


mjwtom:
test-wheel-ubuntu失败的原因，不是代码问题，是磁盘满了
ERROR: Could not install packages due to an OSError: [Errno 28] No space left on device

mooncake-store/src/storage_backend.cpp
LujhCoconut, line 2691: This `lru_index_.erase({0LL, bucket_id})` no longer matches after this PR — BatchOffload now inserts {ts, bucket_id} with a real timestamp. When complete_handler fails and RollbackCommittedBucket runs, this erase silently misses, leaving a stale entry in `lru_index_`. DeleteBucket at line 3167 already does this correctly with `bucket_metadata->last_access_ns_.load(...)`. Please apply the same pattern here.

```cpp
lru_index_.erase({bucket_meta->last_access_ns_.load(std::memory_order_relaxed),
                 bucket_id});
```

he-yufeng (commented):
Read through the diff and the surrounding eviction machinery. The fix itself looks right to me:

- The clock matches what BatchLoad stamps with (steady_clock, line ~1826), so fresh offload timestamps order correctly against read refreshes.
- The std::set key is {ts, bucket_id}, so two buckets created in the same tick cannot collide.
- SelectEvictionCandidate already repairs stale index entries lazily, so the one-directional update here (write ts at insert, never touch the index again) is safe.
- The test fails without the fix (B would sit at 0 and get evicted first) and passes with it, which is exactly the discrimination we want.

One inconsistency worth fixing while you are here: DeleteBucket still erases by the old invariant, `lru_index_.erase({0LL, bucket_id})` (~line 2617). After this change the live entry is keyed by the real timestamp, so that erase silently no-ops and the dead entry rides in the index until SelectEvictionCandidate discards it on the next scan. Harmless today because the lazy repair tolerates it, but the codebase already has the correct pattern a few hundred lines down (~line 3097): load last_access_ns_ from the bucket and erase by that. Same fix here keeps the index tight.

A related gap this does not need to cover, noting so it does not get lost: buckets restored from metadata files on restart also enter lru_index_ with 0 (the `if (success) lru_index_.emplace(0LL, bucket_id)` in the load path), and last_access_ns_ is not persisted in the metadata file. So after a master restart every restored bucket is ts=0 and the first eviction rounds pick victims essentially by insertion order, regardless of how hot they were before. Same zero-timestamp shape as what you fixed here, separate change though. Happy to take that one myself if you would rather keep this PR focused.

Leaving this as a comment rather than an approval since eviction ordering touches quota/data retention; would like another pair of eyes on it before it lands.

mjwtom:
> Read through the diff and the surrounding eviction machinery. The fix itself looks right to me:
> 
> * The clock matches what BatchLoad stamps with (steady_clock, line ~1826), so fresh offload timestamps order correctly against read refreshes.
> * The std::set key is {ts, bucket_id}, so two buckets created in the same tick cannot collide.
> * SelectEvictionCandidate already repairs stale index entries lazily, so the one-directional update here (write ts at insert, never touch the index again) is safe.
> * The test fails without the fix (B would sit at 0 and get evicted first) and passes with it, which is exactly the discrimination we want.
> 
> One inconsistency worth fixing while you are here: DeleteBucket still erases by the old invariant, `lru_index_.erase({0LL, bucket_id})` (~line 2617). After this change the live entry is keyed by the real timestamp, so that erase silently no-ops and the dead entry rides in the index until SelectEvictionCandidate discards it on the next scan. Harmless today because the lazy repair tolerates it, but the codebase already has the correct pattern a few hundred lines down (~line 3097): load last_access_ns_ from the bucket and erase by that. Same fix here keeps the index tight.
> 
> A related gap this does not need to cover, noting so it does not get lost: buckets restored from metadata files on restart also enter lru_index_ with 0 (the `if (success) lru_index_.emplace(0LL, bucket_id)` in the load path), and last_access_ns_ is not persisted in the metadata file. So after a master restart every restored bucket is ts=0 and the first eviction rounds pick victims essentially by insertion order, regardless of how hot they were before. Same zero-timestamp shape as what you fixed here, separate change though. Happy to take that one myself if you would rather keep this PR focused.
> 
> Leaving this as a comment rather than an approval since eviction ordering touches quota/data retention; would like another pair of eyes on it before it lands.

From my perspective, initializing timestamps to zero during post-restore recovery barely causes any practical issues. This behavior essentially degenerates into a native FIFO eviction scheme already. Since bucket IDs are assigned in an incremental manner, eviction will be performed strictly following the bucket ID sequence after service restarts. In other words, the system runs on FIFO eviction for a period right after restart, which I consider an acceptable tradeoff.

Besides, frequent metadata flushing to disk would introduce considerable I/O overhead. I previously assumed that hardcoding zero as the timestamp value was an intentional design decision rather than an oversight.

Metadata updates are recommended to be implemented via the write-and-rename mechanism, even though it incurs a slight performance penalty.

mooncake-store/src/storage_backend.cpp
mjwtom: Think you for remiding. Fixed

LujhCoconut:
> > Read through the diff and the surrounding eviction machinery. The fix itself looks right to me:
> > 
> > * The clock matches what BatchLoad stamps with (steady_clock, line ~1826), so fresh offload timestamps order correctly against read refreshes.
> > * The std::set key is {ts, bucket_id}, so two buckets created in the same tick cannot collide.
> > * SelectEvictionCandidate already repairs stale index entries lazily, so the one-directional update here (write ts at insert, never touch the index again) is safe.
> > * The test fails without the fix (B would sit at 0 and get evicted first) and passes with it, which is exactly the discrimination we want.
> > 
> > One inconsistency worth fixing while you are here: DeleteBucket still erases by the old invariant, `lru_index_.erase({0LL, bucket_id})` (~line 2617). After this change the live entry is keyed by the real timestamp, so that erase silently no-ops and the dead entry rides in the index until SelectEvictionCandidate discards it on the next scan. Harmless today because the lazy repair tolerates it, but the codebase already has the correct pattern a few hundred lines down (~line 3097): load last_access_ns_ from the bucket and erase by that. Same fix here keeps the index tight.
> > A related gap this does not need to cover, noting so it does not get lost: buckets restored from metadata files on restart also enter lru_index_ with 0 (the `if (success) lru_index_.emplace(0LL, bucket_id)` in the load path), and last_access_ns_ is not persisted in the metadata file. So after a master restart every restored bucket is ts=0 and the first eviction rounds pick victims essentially by insertion order, regardless of how hot they were before. Same zero-timestamp shape as what you fixed here, separate change though. Happy to take that one myself if you would rather keep this PR focused.
> > Leaving this as a comment rather than an approval since eviction ordering touches quota/data retention; would like another pair of eyes on it before it lands.
> 
> From my perspective, initializing timestamps to zero during post-restore recovery barely causes any practical issues. This behavior essentially degenerates into a native FIFO eviction scheme already. Since bucket IDs are assigned in an incremental manner, eviction will be performed strictly following the bucket ID sequence after service restarts. In other words, the system runs on FIFO eviction for a period right after restart, which I consider an acceptable tradeoff.
> 
> Besides, frequent metadata flushing to disk would introduce considerable I/O overhead. I previously assumed that hardcoding zero as the timestamp value was an intentional design decision rather than an oversight.
> 
> Metadata updates are recommended to be implemented via the write-and-rename mechanism, even though it incurs a slight performance penalty.

I agree with this reasoning. The ts=0 in the Init restore path is intentional — after restart we have no access history, degenerating to FIFO is a safe and reasonable default, and the I/O cost of persisting timestamps isn't worth it. The real value of this PR is in distinguishing the two scenarios: ts=0 in Init means "no information available," but ts=0 in BatchOffload meant "mark a freshly created bucket as the oldest," which was the actual bug. Now that it's fixed, the intent in both paths is explicit in the code.

From my perspective, there are no blocking issues — this is good to merge.

LujhCoconut (approved):
LGTM.

On the reviewers' concern about losing all LRU access history on restart: what a restart actually loses is the access *recency* information, not all ordering information. Bucket IDs are assigned monotonically, so creation order naturally survives — after restore, buckets fall back to being ordered by bucket_id, i.e. plain FIFO by creation time, which is a reasonable degradation that mitigates the issue until each restored bucket gets its first read and a real timestamp.

In any case, this is a pre-existing behavior of the restore path, not something introduced or worsened by this PR, so it shouldn't block merging. The post-restart eviction ordering can be discussed and addressed in a follow-up if needed.

mjwtom:
请问，是需要进一步review吗？按照我的理解，继续等待review就可以～

LujhCoconut:
@mjwtom 抱歉，消息太多，漏掉了这条。看起来我approve了，但还没merge，明天上班的时候再看一眼，没问题的话我就让maintainer merge一下。

mjwtom:
> @mjwtom 抱歉，消息太多，漏掉了这条。看起来我approve了，但还没merge，明天上班的时候再看一眼，没问题的话我就让maintainer merge一下。

收到，麻烦啦～

ykwd (approved):

Morpheus799:
> @mjwtom 抱歉，消息太多，漏掉了这条。看起来我approve了，但还没merge，明天上班的时候再看一眼，没问题的话我就让maintainer merge一下。

请问能merge了吗😢pr修掉的这个bug会显著增加sglang侧issues/33289这个死锁崩溃的发生概率，希望能快点merge🙏谢谢

PR #3109, opened 2026-07-24
[TENT] Fix failover attempt metrics attribution

anranxia:
### Problem

TENT exposes request-level metrics for a logical transfer, but a logical transfer may contain more than one physical transport attempt. During failover, the task's transport type is changed before the request eventually completes, so the existing metrics only see the terminal transport and final logical outcome.

For example, if an RDMA read fails and is transparently retried over TCP, the recovered request is currently reported as one successful TCP request. The failed RDMA attempt is invisible: it contributes neither a failure nor its own latency. As a result, per-transport failure rates can look healthier than the actual transport behavior, and the logical transport-stage latency cannot be used to understand the cost of each physical attempt.

This change separates the two metric lifecycles:

- **Logical request metrics** continue to describe the user-visible transfer and are recorded once when the request reaches its terminal state.
- **Physical attempt metrics** describe every invocation of a concrete transport, including failed attempts that are later recovered by failover.

### Implementation

- Capture `initial_type` when the task is created so queue-wait and dispatch latency remain attributed to the transport that originally received the request.
- Track the active physical attempt with `attempt_post_time` and `attempt_active`, starting the timer immediately before transport submission.
- Finish and record the failed attempt before failover overwrites `task.type`; the replacement submission then starts a new attempt with the fallback transport.
- Treat synchronous submission errors as failed physical attempts as well, so they cannot bypass the counters or latency histogram.
- Attribute the legacy transport-stage latency to the final physical attempt while preserving the existing logical request counters and final-transport attribution.
- Apply the same attempt lifecycle to both READ and WRITE operations through shared helpers.

### New metrics

| Metric | Meaning |
| --- | --- |
| `tent_transport_attempts_total{transport,operation}` | Number of physical transport attempts submitted |
| `tent_transport_attempt_failures_total{transport,operation}` | Number of physical attempts that ended in failure, including synchronous submission failures |
| `tent_transport_attempt_latency_us{transport,operation}` | Submission-to-completion latency of each physical attempt |

The labels are limited to `transport` and `operation`, so the additional cardinality is bounded by the number of registered transports and the READ/WRITE operation set.

For a recovered RDMA → TCP request, the expected accounting is:

- one successful logical request attributed to TCP, preserving the existing request-level view;
- one RDMA attempt and one RDMA attempt failure, with the failed RDMA latency recorded;
- one TCP attempt with its latency recorded and no TCP attempt failure.

This preserves compatibility for existing dashboards while making transport reliability and failover cost observable without double-counting logical requests. The new metrics remain guarded by the existing `TENT_METRICS_ENABLED` build option. Documentation includes metric semantics, PromQL examples, and cardinality guidance.

## Module

- [ ] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [x] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**

```bash
cmake -S . -B build-tent-proxy -G Ninja \
  -DUSE_TENT=ON -DUSE_HTTP=ON -DUSE_CUDA=OFF \
  -DBUILD_UNIT_TESTS=ON -DBUILD_EXAMPLES=ON \
  -DENABLE_DEBUG_SYMBOLS=OFF -DTENT_METRICS_ENABLED=ON
cmake --build build-tent-proxy --parallel 64
ctest --test-dir build-tent-proxy/mooncake-transfer-engine/tent/tests -j 32 --output-on-failure
./build-tent-proxy/mooncake-transfer-engine/tent/tests/tent_metrics_recording_test \
  --gtest_filter=MetricsRecordingTest.RecoveredFailoverRecordsDistinctAttempts \
  --gtest_repeat=10
pre-commit run --files docs/source/design/tent/metrics.md \
  mooncake-transfer-engine/tent/include/tent/metrics/tent_metrics.h \
  mooncake-transfer-engine/tent/include/tent/runtime/transfer_engine_impl.h \
  mooncake-transfer-engine/tent/src/metrics/tent_metrics.cpp \
  mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp \
  mooncake-transfer-engine/tent/tests/metrics_recording_test.cpp
make -C docs html SPHINXBUILD="python3 -m sphinx"
```

**Test results:**

- [x] Unit tests pass

Validation was run after rebasing onto upstream `main` at `a2ad2806`:

- full CMake/Ninja build passed;
- TENT CTest passed 32/32;
- recovered-failover regression passed 10/10 repeated runs;
- clang-format 20.1.8 dry-run and pre-commit on all six touched files passed;
- Sphinx HTML build passed, and the three new metrics were verified in the rendered page.

An additional top-level CTest run passed 128/131 test executables. The three failures are unrelated to this change: `tcp_transport_test` and `transfer_metadata_test` require a metadata server, while `hot_standby_service_test` has a non-ETCD expectation mismatch. The modified TENT suite remains 32/32.

## Checklist

- [ ] For changes >500 LOC: I have filed an RFC issue (not applicable; production changes are below the threshold excluding tests)

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex was used to analyze the existing metrics/failover paths, implement and review the code/tests/docs, and run the validation commands.

### DIFF
--- a/mooncake-transfer-engine/tent/src/metrics/tent_metrics.cpp
+++ b/mooncake-transfer-engine/tent/src/metrics/tent_metrics.cpp
@@ -31,6 +31,12 @@ TentMetrics::~TentMetrics() { shutdown(); }
 
 #if TENT_METRICS_ENABLED
 
+namespace {
+const char* operationName(Request::OpCode operation) {
+    return operation == Request::READ ? "read" : "write";
+}
+}  // namespace
+
 Status TentMetrics::initialize(const MetricsConfig& config) {
     // Validate configuration before touching initialized_. An invalid config
     // (e.g. port 0, zero HTTP threads) would otherwise cause confusing
@@ -201,18 +207,29 @@ void TentMetrics::shutdown() {
 
 void TentMetrics::registerMetrics() {
     // Register all counters as base metric_t* pointers so that counters with
-    // different label arities (N=1 per-transport, N=2 failover from→to) share
-    // one vector for Prometheus serialize().
+    // different label arities (N=1 per-transport, N=2 failover from→to and
+    // transport-attempt operation labels) share one vector for Prometheus
+    // serialize().
     counters_ = {
-        &read_bytes_total_,    &write_bytes_total_,
-        &read_requests_total_, &write_requests_total_,
-        &read_failures_total_, &write_failures_total_,
-        &failover_total_,      &deadline_infeasible_total_,
+        &read_bytes_total_,
+        &write_bytes_total_,
+        &read_requests_total_,
+        &write_requests_total_,
+        &read_failures_total_,
+        &write_failures_total_,
+        &failover_total_,
+        &transport_attempts_total_,
+        &transport_attempt_failures_total_,
+        &deadline_infeasible_total_,
     };
 
-    // Register all histograms - add new histograms here. Each entry pairs the
-    // histogram with its compile-time bucket boundaries; the struct keeps the
-    // two in sync so getJsonMetrics() cannot mislabel buckets.
+    // Register the N=1 per-transport histograms with their bucket boundaries
+    // and parallel sum counters. Each entry keeps the histogram, its
+    // compile-time boundaries, and its sum counter in sync so both the
+    // Prometheus and JSON serializers cannot mislabel buckets or drop _sum.
+    // The N=2 transport_attempt_latency_ histogram is serialized separately
+    // (it can't share this N=1-typed vector); see getPrometheusMetrics /
+    // getJsonMetrics.
     histograms_ = {
         {&read_latency_, &kLatencyBuckets, &read_latency_sum_},
         {&write_latency_, &kLatencyBuckets, &write_latency_sum_},
@@ -325,6 +342,32 @@ void TentMetrics::recordTransportFailover(TransportType from,
                                                    transportTypeName(to)});
 }
 
+void TentMetrics::recordTransportAttemptStarted(TransportType tp,
+                                                Request::OpCode operation) {
+    if (!initialized_ || !runtime_enabled_.load(std::memory_order_relaxed))
+        return;
+    transport_attempts_total_.inc(std::array<std::string, 2>{
+        transportTypeName(tp), operationName(operation)});
+}
+
+void TentMetrics::recordTransportAttemptFinished(TransportType tp,
+                                                 Request::OpCode operation,
+                                                 TransferStatusEnum status,
+                                                 double latency_us) {
+    if (!initialized_ || !runtime_enabled_.load(std::memory_order_relaxed))
+        return;
+    auto label = std::array<std::string, 2>{transportTypeName(tp),
+                                            operationName(operation)};
+    if (status == FAILED) {
+        transport_attempt_failures_total_.inc(label);
+    }
+    if (latency_us >= 0.0) {
+        auto latency_val = static_cast<int64_t>(latency_us);
+        transport_attempt_latency_.observe(label, latency_val);
+        transport_attempt_latency_sum_.inc(label, latency_val);
+    }
+}
+
 std::string TentMetrics::getPrometheusMetrics() {
     if (!initialized_) return "";
 
@@ -354,6 +397,11 @@ std::string TentMetrics::getPrometheusMetrics() {
             serializeHistogramPrometheus(entry.h, *entry.boundaries, *entry.sum,
                                          result);
         }
+        // N=2 transport-attempt latency histogram (labels: transport,
+        // operation) goes through the same helper, instantiated for N=2.
+        serializeHistogramPrometheus(&transport_attempt_latency_,
+                                     kLatencyBuckets,
+                                     transport_attempt_latency_sum_, result);
 
         return result;
     } catch (const std::exception& e) {
@@ -373,12 +421,43 @@ int64_t sumCounterValues(CounterPtr counter) {
     }
     return total;
 }
+
+template <uint8_t N>
+void serializeHistogramToJson(
+    nlohmann::json& root,
+    ylt::metric::basic_dynamic_histogram<int64_t, N>* hist,
+    const std::vector<double>& boundaries,
+    ylt::metric::basic_dynamic_counter<int64_t, N>* sum) {
+    auto bucket_counts = hist->get_bucket_counts();
+    int64_t total_count = 0;
+    int64_t total_sum = 0;
+    nlohmann::json buckets_obj;
+    for (size_t i = 0; i < bucket_counts.size(); ++i) {
+        int64_t bucket_total = sumCounterValues(bucket_counts[i]);
+        total_count += bucket_total;
+        if (i < boundaries.size()) {
+            buckets_obj[std::to_string(static_cast<int64_t>(boundaries[i]))] =
+                bucket_total;
+        }
+    }
+    // ylt keeps the histogram's sum_ private, so read the parallel sum counter
+    // maintained alongside each observe() call (matches the Prometheus path).
+    for (auto& e : sum->copy()) {
+        total_sum += e->value.load();
+    }
+    nlohmann::json hist_obj;
+    hist_obj["count"] = total_count;
+    hist_obj["sum"] = total_sum;
+    hist_obj["buckets"] = buckets_obj;
+    root[hist->str_name()] = hist_obj;
+}
 }  // namespace
 
+template <uint8_t N>
 void TentMetrics::serializeHistogramPrometheus(
-    ylt::metric::basic_dynamic_histogram<int64_t, 1>* hist,
+    ylt::metric::basic_dynamic_histogram<int64_t, N>* hist,
     const std::vector<double>& boundaries,
-    ylt::metric::basic_dynamic_counter<int64_t, 1>& sum,
+    ylt::metric::basic_dynamic_counter<int64_t, N>& sum,
     std::string& out) const {
     // Walk the same data the JSON path uses (get_bucket_counts() + copy()),
     // so the two endpoints cannot drift. Unlike ylt's serialize() this never
@@ -386,8 +465,32 @@ void TentMetrics::serializeHistogramPrometheus(
     // output string (taking the # HELP / # TYPE header with it) whenever
     // every label combo has sum_==0, which is reachable in production when
     // sub-microsecond latencies truncate to 0 under int64_t observation.
+    //
+    // Templated over label arity N: the per-transport histograms are N=1 and
+    // the transport-attempt latency histogram is N=2. For N=1 the emitted
+    // text is byte-identical to the original single-label implementation.
     auto bucket_counts = hist->get_bucket_counts();
     if (bucket_counts.empty()) return;
+
+    const auto& label_names = hist->labels_name();
+
+    // A unique key per label tuple, used to dedup combos and look up the sum.
+    auto combo_key = [](const std::array<std::string, N>& lv) {
+        std::string k;
+        for (uint8_t i = 0; i < N; ++i) {
+            k.append(lv[i]);
+            k.push_back('\x1f');  // separator that cannot appear in a label
+        }
+        return k;
+    };
+    // Render `name0="v0",name1="v1"` for a label tuple (no surrounding braces).
+    auto append_labels = [&](std::string& dst,
+                             const std::array<std::string, N>& lv) {
+        for (uint8_t i = 0; i < N; ++i) {
+            if (i) dst.append(",");
+            dst.append(label_names[i]).append("=\"").append(lv[i]).append("\"");
+        }
+    };
 
     // Build the union of label combos across ALL buckets. ylt's observe()
     // increments exactly one bucket per observation (the bucket containing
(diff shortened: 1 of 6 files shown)


anranxia:
Hi @ShangmingCai, I noticed that the [Build & Test (Linux) workflow run](https://github.com/kvcache-ai/Mooncake/actions/runs/30107901893) for this PR was canceled while the TENT jobs were still running. The lightweight checks passed, and `CI Gate` appears to be failing because the matrix jobs were canceled rather than due to a test failure.

Was the cancellation intentional, or should the workflow be re-run? If there is any concern with the PR or CI setup that I should address, please let me know. Thanks!

mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
```diff
@@ -1499,8 +1499,16 @@ Status TransferEngineImpl::commitPreparedSubmit(
 
         if (owner.staging) {
             task.staging = true;
-            staging_proxy_->submit(&task, (BatchID)batch, owner.staging_params);
-            task.post_time = std::chrono::steady_clock::now();
+            startTransportAttempt(task, task.type,
+                                  std::chrono::steady_clock::now());
+            auto status = staging_proxy_->submit(&task, (BatchID)batch,
+                                                 owner.staging_params);
+            if (!status.ok()) {
```
staryxchen, line 1509: Suggested follow-up: the four synchronous submit-failure branches added here (staging in \`commitPreparedSubmit\`, non-staging in \`commitPreparedSubmit\`, both \`dispatchQueuedOwner\` paths, and \`resubmitTransferTask\`) are not covered by tests — \`FakeTransport::submitTransferTasks\` always returns \`Status::OK()\`, so \`RecoveredFailoverRecordsDistinctAttempts\` only exercises poll-observed failure. This staging branch is the most worth covering since it newly clears \`task.staging\` and resets \`task.type = UNSPEC\` on failure (a behavior change beyond metrics). Extending \`FakeTransport\` with a \`force_submit_fail\` mode and asserting one failed attempt under the correct \`transport\`/\`operation\` label would close the gap.

mooncake-transfer-engine/tent/src/metrics/tent_metrics.cpp
```diff
@@ -336,6 +378,11 @@ std::string TentMetrics::getPrometheusMetrics() {
             entry.h->serialize(tmp);
             result += tmp;
         }
+        for (const auto& entry : attempt_histograms_) {
```
staryxchen: Minor: this loop duplicates the \`histograms_\` loop just above. \`counters_\` already stores \`metric_t*\` base pointers so mixed arities share one loop — the same could be done for Prometheus serialization here (the typed \`HistogramEntry\`/\`AttemptHistogramEntry\` vectors are only needed for JSON aggregation). Not blocking, but worth unifying so the pattern does not replicate with the next histogram arity.

mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
```diff
@@ -2388,23 +2418,28 @@ void TransferEngineImpl::recordTaskCompletionMetrics(
...
                                 task.post_time - task.dispatch_time)
                                 .count();
+                        TENT_RECORD_STAGE_LATENCY(TentMetrics::Stage::Dispatch,
+                                                  initial_type, dispatch_us);
+                    }
+                    if (task.attempt_post_time.time_since_epoch().count() > 0) {
                         double transport_us =
                             std::chrono::duration<double, std::micro>(
-                                end_time - task.post_time)
+                                end_time - task.attempt_post_time)
```
staryxchen: Question on backward compatibility: this rewrite changes the semantics of two *existing* stage metrics, not just adds new ones:

- \`tent_stage_transport_us\` — previously \`end_time - task.post_time\` (first post -> completion, including failover); now \`end_time - task.attempt_post_time\` (final attempt only). For a recovered RDMA->TCP request, the RDMA + failover interval is dropped from the sample.
- \`tent_stage_dispatch_us\` label — previously \`task.type\` (final transport); now \`initial_type\` (first transport). A recovered RDMA->TCP request dispatch latency now lands under \`rdma\` instead of \`tcp\`.

The PR description says "preserves compatibility for existing dashboards," but dashboards alerting on these two series will see materially different values for any request that experienced failover.

Could this be made truly additive instead — keep \`tent_stage_transport_us\` / \`tent_stage_dispatch_us\` on their original semantics (\`post_time\`, \`task.type\`), and rely solely on the new \`tent_transport_attempt_latency_us\` histogram for the per-attempt view? The new histogram already captures every physical attempt latency, so rewriting the legacy stage metrics does not seem strictly necessary. If the legacy semantics must change for a reason I am missing, could that be called out explicitly in the PR description so dashboard owners can migrate?

mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
anranxia: Added a force_submit_fail mode to FakeTransport (returns non-ok from submitTransferTasks without enqueuing a task) plus two tests: SynchronousSubmitFailureRecordsFailedAttempt covers the commitPreparedSubmit direct-path branch, and FailoverResubmitSyncFailureRecordsBothAttempts covers the resubmitTransferTask branch (RDMA async-fail → failover → TCP sync-fail). They assert the failed attempt is counted and that recordTaskCompletionMetrics does not double-count an attempt that was already closed synchronously. The two dispatchQueuedOwner branches go through the same startTransportAttempt/finishTransportAttempt helpers and require the runtime queue, which this metrics harness doesn't enable, so they're covered by the shared helper path rather than a dedicated queue test.

mooncake-transfer-engine/tent/src/metrics/tent_metrics.cpp
anranxia: Done. Histograms are now held as base ylt::metric::metric_t* pointers (the same type-erasure counters_ already uses), so histograms of any label arity serialize in a single loop. Removed the now-unused per-entry boundaries field; getJsonMetrics() keeps aggregating the typed members directly.

mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
anranxia: Good catch — reverted this to a purely additive change. tent_stage_queue_wait_us / tent_stage_dispatch_us / tent_stage_transport_us keep their original semantics (attributed to the final transport, with the transport stage spanning the whole request), so existing dashboards and alerts are unaffected. All per-attempt / per-transport reliability now lives exclusively in the new additive tent_transport_attempt_total / tent_transport_attempt_failures_total / tent_transport_attempt_latency_us metrics. Updated metrics.md and the failover test to match, and dropped the no-longer-needed initial_type field.

mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
```diff
@@ -1499,8 +1499,16 @@ Status TransferEngineImpl::commitPreparedSubmit(
 
         if (owner.staging) {
             task.staging = true;
-            staging_proxy_->submit(&task, (BatchID)batch, owner.staging_params);
-            task.post_time = std::chrono::steady_clock::now();
+            startTransportAttempt(task, task.type,
+                                  std::chrono::steady_clock::now());
```
alogfans: staging_proxy_->submit() is an orchestration step rather than a concrete transport attempt; ProxyManager may later issue internal transport submissions that are counted separately. Recording an outer attempt here can double-count and misattribute latency. Please either count only actual Transport::submitTransferTasks() calls or expose staging as a separate metric, and add coverage for staged submission failure/failover. It would also be safer to store attempt_type explicitly instead of reading the mutable task.type when finishing an attempt.

anranxia: Double-counting / staging attribution: You're right that staging_proxy_->submit() is orchestration, not a transport attempt — ProxyManager chunks the transfer and the per-chunk submitStagingTransfer() calls recurse through commitPreparedSubmit/dispatchQueuedOwner and are counted there as the real Transport::submitTransferTasks() attempts. I removed the start/finishTransportAttempt calls from both staging branches so only concrete transport submissions are counted; the staging branch now just stamps post_time for stage decomposition. No more double-count or outer-span latency misattribution.

attempt_type: Done — added TaskInfo::attempt_type, captured in startTransportAttempt, and finishTransportAttempt now labels with it instead of the mutable task.type, so an attempt is attributed to the transport that ran it even after failover overwrites task.type.

Staged-failure coverage: The metrics unit harness drives everything through LOCAL_SEGMENT_ID, which short-circuits findStagingPolicy, and there's no seam to inject a staging policy/proxy without a remote-segment + buffer-topology integration fixture. Since the staging branches no longer touch attempt metrics, there's no attempt-metric behavior left to assert on them; the per-transport failover attribution that attempt_type guarantees is covered by FailoverResubmitSyncFailureRecordsBothAttempts. Happy to add a dedicated staging integration test (or a small test hook) in a follow-up if you'd prefer explicit coverage.

anranxia: The tent-ci (cuda-off-metrics-on) failure is a transient CI infra flake, not a code issue: the build stopped at 
mooncake-common/src/asio_impl.cpp because sccache couldn't reach the GitHub Actions Cache backend — sccache: error: cache storage failed to read ... dns error: failed to lookup address information: Try again while connecting to *.blob.core.windows.net. No compile error occurred (the later rdma_transport.cpp line is just the last in-flight job before ninja stopped). 
Could you also re-run the failed job? Thanks!

alogfans: @anranxia You can submit an empty commit to retrigger ci.

alogfans (approved):
LGTM

anranxia:
@ShangmingCai The only remaining gate is a /docs CODEOWNERS approval — this PR touches docs/source/design/tent/metrics.md. alogfans has approved the TENT code and all CI is green. Could you (or another /docs owner) approve the docs change? Thanks!

anranxia:
Rebased onto latest main to resolve the conflicts with #3126 (the Prometheus histogram silent-drop / real _sum fix), which landed after this PR was opened and touched the same metrics files.

  How the conflict was resolved (@staryxchen, since #3126 is yours — please sanity-check this):

  - Kept #3126 as-is in intent: the HistogramEntry{h, boundaries, sum} struct, the 8 parallel *_sum_ counters incremented alongside each observe(), the silent-drop-safe bucket-walking Prometheus serializer, and the new "sum" field in the JSON endpoint are all preserved.
  - Generalized #3126's serializeHistogramPrometheus() and the JSON histogram serializer from single-label to template
  <uint8_t N>, so the existing N=1 per-transport histograms and this PR's new N=2 tent_transport_attempt_latency_us (labels: transport, operation) share one implementation instead of duplicating the bucket-walk logic. For N=1 the emitted
  Prometheus text is byte-identical to #3126, so its regression tests (PrometheusExposesHistogramWhenAllSamplesAreZero, ...ForTcpTransport, JsonAndPrometheusAgreeOnHistogramCount) are unaffected.
  - The N=2 attempt-latency histogram gets its own parallel transport_attempt_latency_sum_ counter, following the exact same pattern #3126 established for the N=1 histograms.

  Net effect: #3126's fix and this PR's per-attempt/failover metrics now coexist; no behavior change to any metric that
  existed before either PR. All conflicts were in tent_metrics.{h,cpp}; the test file merged cleanly (both #3126's and this
  PR's tests are present). CI is re-running on the rebased commit.

anranxia:
The rebase surfaced one cross-test coupling: PrometheusExposesZeroValuedHistogramForTcpTransport asserted an absolute tcp stage_queue_wait +Inf count of 3, but this PR's RecoveredFailoverRecordsDistinctAttempts completes a real RDMA→TCP recovery that records tcp stage latency earlier in the same (singleton, non-reset) process. Made that assertion delta-based (snapshot the +Inf cumulative, assert it advances by exactly 3) — same convention as  JsonAndPrometheusAgreeOnHistogramCount. The delta still validates the serializer's cumulative correctness, so it's not 
masking anything.

staryxchen:
> Rebased onto latest main to resolve the conflicts with #3126 (the Prometheus histogram silent-drop / real _sum fix), which landed after this PR was opened and touched the same metrics files.
> 
> How the conflict was resolved (@staryxchen, since #3126 is yours — please sanity-check this):
> 
> * Kept [[tent] Fix Prometheus histogram silent-drop and clean stale config keys #3126](https://github.com/kvcache-ai/Mooncake/pull/3126) as-is in intent: the HistogramEntry{h, boundaries, sum} struct, the 8 parallel *_sum_ counters incremented alongside each observe(), the silent-drop-safe bucket-walking Prometheus serializer, and the new "sum" field in the JSON endpoint are all preserved.
> * Generalized [[tent] Fix Prometheus histogram silent-drop and clean stale config keys #3126](https://github.com/kvcache-ai/Mooncake/pull/3126)'s serializeHistogramPrometheus() and the JSON histogram serializer from single-label to template
>   <uint8_t N>, so the existing N=1 per-transport histograms and this PR's new N=2 tent_transport_attempt_latency_us (labels: transport, operation) share one implementation instead of duplicating the bucket-walk logic. For N=1 the emitted
>   Prometheus text is byte-identical to [[tent] Fix Prometheus histogram silent-drop and clean stale config keys #3126](https://github.com/kvcache-ai/Mooncake/pull/3126), so its regression tests (PrometheusExposesHistogramWhenAllSamplesAreZero, ...ForTcpTransport, JsonAndPrometheusAgreeOnHistogramCount) are unaffected.
> * The N=2 attempt-latency histogram gets its own parallel transport_attempt_latency_sum_ counter, following the exact same pattern [[tent] Fix Prometheus histogram silent-drop and clean stale config keys #3126](https://github.com/kvcache-ai/Mooncake/pull/3126) established for the N=1 histograms.
> 
> Net effect: #3126's fix and this PR's per-attempt/failover metrics now coexist; no behavior change to any metric that existed before either PR. All conflicts were in tent_metrics.{h,cpp}; the test file merged cleanly (both #3126's and this PR's tests are present). CI is re-running on the rebased commit.

LGTM, no concerns from my side.

anranxia:
Only the /docs CODEOWNERS approval is left before this can merge — the PR documents the new metrics in 
docs/source/design/tent/metrics.md. Code is approved and CI is green. @ShangmingCai / @stmatengss / @ykwd, 
could one of you approve the docs update? Thanks!

anranxia:
@alogfans all green and mergeable now — could you hit merge when you get a chance? Thanks!

PR #3112, opened 2026-07-24
[Wheel] Add native fast-copy PUT hot path

zxpdemonio:
## Description

This PR splits the native fast-copy PUT hot-path optimization out of the previous large rollout transfer optimization branch.

RL rollout transfer has several hot paths made of many small ndarray rows and large object payloads. The previous source-buffer PUT path could rely on temporary source registration and Python-level copy loops, which is expensive for Miles/Slime-style rollout data and makes source-buffer lifetime semantics harder to reason about.

This change:
- adds a small native `_fast_copy` extension for scatter-copying ndarray lists into a destination buffer while releasing the GIL;
- pins ndarray references and validates the destination byte count before the GIL is released, so the native copy path cannot outlive borrowed Python objects or overflow the BufferPool lease;
- introduces `_DirectCopyPayload` so typed-ragged ndarray rows can be copied directly into BufferPool leases at PUT time;
- removes the temporary source-buffer registration PUT path and routes non-tensor structured PUTs through BufferPool staging plus `batch_put_from`;
- keeps tensor-object zero-copy semantics explicit and intentionally rejects source-buffer zero-copy requests for typed-ragged/direct-copy payloads;
- keeps this PR independent from the typed-ragged layout PR (#3113). If both PRs are merged, #3113 should land first and this PR can be rebased on top to resolve the mechanical overlap in `structured_object_store.py`;
- keeps the rollout-transfer tuned defaults used by the optimization branch: 64 MB bundle chunks and parallel PUT threshold aligned with the chunk size. This default change is deliberate for structured/bundle rollout PUTs and remains overridable by caller policy/configuration.

The native extension build uses NumPy headers through the wheel build environment; `pyproject.toml` includes NumPy in the build-system requirements, and the NPU `--no-isolation` wheel build path now installs NumPy before invoking the build backend.

Performance validation from rollout transfer benchmark targets:

| Scenario | Baseline behavior | Optimized behavior |
| --- | ---: | ---: |
| Miles rollout dict, 4 GiB | PUT 14.184s on the slow path before DirectCopy; latest main baseline GET did not finish within 3 minutes | PUT 6.460s, GET 3.942s |
| Slime rollout capture, 17.2 GB | transfer remains on the BufferPool/RDMA hot path | PUT 1.984s, GET 1.709s |

The Miles numbers are for the native fast-copy/BufferPool PUT hot path. The Slime numbers validate that the source-register cleanup and tuned chunk defaults keep the large rollout-object path on the expected RDMA hot path.

## Module

- [ ] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [x] Documentation update
- [x] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
cd mooncake-wheel
python setup.py build_ext --inplace
python - <<'PY'
from mooncake._fast_copy import concat_arrays_into
print(concat_arrays_into)
PY
python -m unittest -v tests.test_fast_copy
python -m pytest tests/test_structured_object_store.py -q --tb=short
python -m py_compile mooncake/structured_object_store.py tests/test_structured_object_store.py tests/test_fast_copy.py
git diff --check
cd ../docs
make html SPHINXOPTS=-W
```

**Test results:**
- [x] Unit tests pass
- [x] Manual testing done

Result:
```text
structured-object validation: 76 passed, 25 skipped
late-chunk rollback regression: passed
fast-copy boundary tests: 4 passed
native extension build/import: passed
Sphinx docs strict build: passed
code_format: passed
py_compile: passed
git diff --check: passed
```

Additional manual validation:
```text
Miles rollout dict cross-machine benchmark:
PUT avg: 6460ms
GET avg: 3941.9ms

Slime rollout capture cross-machine benchmark:
17.2 GB PUT: 1984ms
17.2 GB GET: 1708.6ms
```

## Checklist

Note: this focused split is slightly above 500 additions after retaining reviewer-requested safety/regression coverage for the native extension boundary. Python formatting/pre-commit were not used as the validation signal for this focused split to avoid unrelated repository-wide formatting churn; native extension build/import, targeted unit tests, `py_compile`, and `git diff --check` were used.

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex was used to split this focused PR out of the previous large rollout transfer optimization branch, review the diff, run validation, and prepare the PR description. The human submitter is responsible for understanding and defending all changes.

### DIFF
--- a/docs/source/api-reference/python/dataproto-structured-object-transfer.md
+++ b/docs/source/api-reference/python/dataproto-structured-object-transfer.md
@@ -122,6 +122,8 @@ Numeric numpy arrays are stored as structured ndarray members. Contiguous arrays
 
 `non_tensor_batch` object arrays are structured-encoded according to their contents. Numeric scalar object arrays, ragged tensors, bytes, strings, JSON-like values, and selected media payloads have explicit codecs. These fields are serialized by design and should not be treated as zero-copy tensor data.
 
+For typed-ragged ndarray rows, Mooncake copies the rows directly into BufferPool staging memory with the native fast-copy extension and writes each chunk through `batch_put_from`. If the extension or BufferPool path is unavailable, the implementation transparently falls back to the regular copy path. This optimization does not change the stored format or require caller configuration.
+
 ## Materializing into caller buffers
 
 `destinations` can reuse caller-provided buffers:
@@ -162,6 +164,8 @@ transfer.put_structured_object(payload, policy=policy)
 ```
 
 `copy_mode="zero_copy"` requires tensor payloads to be provided as `tensor_object_buffer`; plain torch tensors are rejected because they do not expose a registered tensor-object buffer.
+
+Typed-ragged and other non-tensor structured payloads do not support `copy_mode="zero_copy"`. Use `auto` to enable BufferPool staging and the native fast-copy path when available, or `copy` to force the regular `store.put` path.
 
 ## Cleanup
 
--- a/docs/source/api-reference/python/dataproto-structured-object-transfer.md
+++ b/docs/source/api-reference/python/dataproto-structured-object-transfer.md
@@ -118,7 +118,7 @@ Tensor fields are stored through the best available Mooncake path:
 3. If a tensor-native path is unavailable, Mooncake falls back to a serialized tensor payload.
 4. Scalar tensors use a correctness fallback until the native tensor codec preserves zero-dimensional shape.
 
-Numeric numpy arrays are stored as structured ndarray members. Contiguous arrays can be passed through without copying; non-contiguous arrays are made contiguous before storage. Row slices are materialized through structured range reads when the backend supports them.
+Numeric numpy arrays are stored as structured ndarray members. Non-tensor ndarray PUTs are staged and copied before being written to Mooncake; non-contiguous arrays are made contiguous as part of that staging path. Row slices are materialized through structured range reads when the backend supports them, and `destinations` can still be used on GET to materialize selected fields into caller-provided buffers.
 
 `non_tensor_batch` object arrays are structured-encoded according to their contents. Numeric scalar object arrays, ragged tensors, bytes, strings, JSON-like values, and selected media payloads have explicit codecs. These fields are serialized by design and should not be treated as zero-copy tensor data.
 
--- a/mooncake-wheel/mooncake/_fast_copy.c
+++ b/mooncake-wheel/mooncake/_fast_copy.c
@@ -1,0 +1,194 @@
+#define PY_SSIZE_T_CLEAN
+#include <Python.h>
+#include <stdint.h>
+#include <stdlib.h>
+#include <string.h>
+#include <pthread.h>
+
+#define NPY_NO_DEPRECATED_API NPY_1_7_API_VERSION
+#include <numpy/arrayobject.h>
+
+typedef struct {
+    void **src_ptrs;
+    size_t *src_sizes;
+    int count;
+    char *dst;
+    size_t offset;
+    size_t bytes_copied;
+} ThreadWork;
+
+static void *copy_thread_func(void *arg) {
+    ThreadWork *w = (ThreadWork *)arg;
+    char *d = w->dst + w->offset;
+    size_t off = 0;
+    for (int i = 0; i < w->count; i++) {
+        if (w->src_sizes[i] > 0) {
+            memcpy(d + off, w->src_ptrs[i], w->src_sizes[i]);
+            off += w->src_sizes[i];
+        }
+    }
+    w->bytes_copied = off;
+    return NULL;
+}
+
+static PyObject *concat_arrays_into(PyObject *self, PyObject *args) {
+    PyObject *list_obj;
+    unsigned long long dest_ptr_val;
+    unsigned long long dest_size_val;
+    Py_ssize_t start = 0;
+    Py_ssize_t count = -1;
+    int nthreads = 1;
+    ThreadWork *works = NULL;
+    pthread_t *threads = NULL;
+    char *thread_started = NULL;
+
+    if (!PyArg_ParseTuple(args, "O!KK|nni", &PyList_Type, &list_obj,
+                          &dest_ptr_val, &dest_size_val, &start, &count,
+                          &nthreads))
+        return NULL;
+
+    Py_ssize_t list_len = PyList_GET_SIZE(list_obj);
+    if (start < 0) start = 0;
+    if (start > list_len) start = list_len;
+    if (count < 0 || start + count > list_len) count = list_len - start;
+    if (count == 0) return PyLong_FromSize_t(0);
+    if (nthreads < 1) nthreads = 1;
+    if (nthreads > (int)count) nthreads = (int)count;
+
+    size_t dest_size = (size_t)dest_size_val;
+    void **ptrs = (void **)malloc(count * sizeof(void *));
+    size_t *sizes = (size_t *)malloc(count * sizeof(size_t));
+    PyObject **items = (PyObject **)calloc(count, sizeof(PyObject *));
+    if (!ptrs || !sizes || !items) {
+        free(ptrs);
+        free(sizes);
+        free(items);
+        return PyErr_NoMemory();
+    }
+
+    Py_ssize_t held = 0;
+    size_t total_input_bytes = 0;
+    for (Py_ssize_t i = 0; i < count; i++) {
+        PyObject *item = PyList_GET_ITEM(list_obj, start + i);
+        if (!PyArray_Check(item)) {
+            PyErr_Format(PyExc_TypeError, "arrays[%zd] is not an ndarray",
+                         start + i);
+            goto fail;
+        }
+        Py_INCREF(item);
+        items[held++] = item;
+        PyArrayObject *arr = (PyArrayObject *)item;
+        if (!PyArray_IS_C_CONTIGUOUS(arr)) {
+            PyErr_Format(PyExc_ValueError, "arrays[%zd] is not C-contiguous",
+                         start + i);
+            goto fail;
+        }
+        size_t nbytes = (size_t)PyArray_NBYTES(arr);
+        if (nbytes > SIZE_MAX - total_input_bytes) {
+            PyErr_SetString(PyExc_OverflowError, "array byte sizes overflow");
+            goto fail;
+        }
+        ptrs[i] = PyArray_DATA(arr);
+        sizes[i] = nbytes;
+        total_input_bytes += nbytes;
+    }
+    if (total_input_bytes > dest_size) {
+        PyErr_Format(PyExc_ValueError,
+                     "destination buffer too small: need %zu bytes, got %zu",
+                     total_input_bytes, dest_size);
+        goto fail;
+    }
+
+    /* Partition work across threads. */
+    works = (ThreadWork *)calloc(nthreads, sizeof(ThreadWork));
+    threads = (pthread_t *)malloc(nthreads * sizeof(pthread_t));
+    thread_started = (char *)calloc(nthreads, sizeof(char));
+    if (!works || !threads || !thread_started) {
+        PyErr_NoMemory();
+        goto fail;
+    }
+
+    int n = (int)count;
+    int per_t = (n + nthreads - 1) / nthreads;
+    size_t offset = 0;
+    int actual_threads = 0;
+    for (int t = 0; t < nthreads; t++) {
+        int s = t * per_t;
+        int c = per_t;
+        if (s + c > n) c = n - s;
+        if (c <= 0) break;
+
+        works[t].src_ptrs = ptrs + s;
+        works[t].src_sizes = sizes + s;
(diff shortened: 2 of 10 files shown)


mooncake-wheel/mooncake/_fast_copy.c
```diff
@@ -0,0 +1,175 @@
...
+        PyArrayObject *arr = (PyArrayObject *)item;
+        if (!PyArray_IS_C_CONTIGUOUS(arr)) {
+            free(ptrs);
+            free(sizes);
+            PyErr_Format(PyExc_ValueError, "arrays[%zd] is not C-contiguous",
+                         start + i);
+            return NULL;
+        }
+        ptrs[i] = PyArray_DATA(arr);
+        sizes[i] = (size_t)PyArray_NBYTES(arr);
```
zxpdemonio: **blocker:** after collecting `PyArray_DATA` / `NBYTES` from borrowed list items, the GIL is released below and pthreads read those pointers. Please hold strong refs (`Py_INCREF` the list or each array) until after `Py_END_ALLOW_THREADS`, and consider rejecting non-writeable / overlapping views if that matters for correctness.

Also please add a `dest_size` (or expected total) argument and fail if `sum(sizes) > dest_size` — today a Python-side size mismatch is unchecked `memcpy` into pool memory.

zxpdemonio (commented):
## Review summary

Solid split of the native fast-copy PUT hot path out of #2671. `_DirectCopyPayload` + BufferPool staging + `batch_put_from`, with soft fallback when `_fast_copy` is unavailable, is a clear improvement for Miles/Slime-style many-small-ndarray PUTs. Rejecting non-tensor `zero_copy` source-buffer puts is also the right semantic cleanup.

I do have a few **must-fix / should-fix** items before merge, especially around the C extension safety contract and the global default chunk-size change.

> Note: posted as `COMMENT` because GitHub disallows `REQUEST_CHANGES` on own PRs (`zxpdemonio`).

### Must fix
1. **`concat_arrays_into` does not validate destination capacity.** It takes a raw `dest_ptr` and `memcpy`s `sum(NBYTES)` bytes with no `dest_size` argument. Please pass the lease size and check `copied == expected` / `copied <= dest_size` before returning, otherwise a mismatch is silent memory corruption of RDMA-registered pool memory.
2. **GIL release without owning refs to the source arrays.** Pointers are taken via borrowed `PyList_GET_ITEM` refs, then `Py_BEGIN_ALLOW_THREADS` runs. Please `Py_INCREF` the list (or each array) across the copy window so another thread cannot drop the last ref while pthreads are still reading `PyArray_DATA`.

### Should fix
3. **`DEFAULT_BUNDLE_CHUNK_BYTES` 512 MiB → 64 MiB** is a global behavior change for all structured/bundle puts, not just rollout. Call this out loudly in the PR description / changelog, and consider keeping the old default with an explicit rollout-oriented override if other workloads relied on 512 MiB chunks.
4. **Heavy overlap with #3113** on `structured_object_store.py` (`_encode_typed_ragged_values`, transport put paths). This PR still uses `torch.cat` for `ragged_tensor`, while #3113 rewrites that path. Please document intended merge order and rebase one onto the other before landing both.
5. **`setup.py` imports `numpy` at module import time** to build the extension. Ensure wheel CI / docs state that `numpy` must be present at build time (pyproject already adds it — good — but call out failure mode if missing).

### Nits
6. C formatting around `Py_BEGIN_ALLOW_THREADS if (actual_threads == 1)` is hard to read; put the `if` on the next line.
7. `_put_chunk_batch(..., nthreads=1)` never uses intra-chunk `nthreads > 1` (parallelism is only across chunk groups). Either wire it or drop the parameter to avoid implying unused tuning.

mooncake-wheel/mooncake/_fast_copy.c
```diff
@@ -0,0 +1,175 @@
...
+        works[t].offset = offset;
+
+        size_t tb = 0;
+        for (int j = s; j < s + c; j++) tb += sizes[j];
+        offset += tb;
+        actual_threads++;
+    }
+
+    /* Copy with GIL released. */
+    Py_BEGIN_ALLOW_THREADS if (actual_threads == 1) {
```
zxpdemonio: Style/nit: put the `if (actual_threads == 1)` on its own line after `Py_BEGIN_ALLOW_THREADS` for readability. Functionally OK.

If `pthread_create` fails you fall back to inline `copy_thread_func` while other threads may already be running — that is fine, but worth a short comment so future edits do not assume all `actual_threads` were started.

mooncake-wheel/mooncake/structured_object_store.py
```diff
@@ -19,8 +19,14 @@
...
-DEFAULT_BUNDLE_CHUNK_BYTES = 512 * 1024**2
-AUTO_PARALLEL_MIN_BYTES = 4 * 1024**3
+# -- C fast-path: concat_arrays_into(list[ndarray], dest_ptr) ----------------
+try:
+    from mooncake._fast_copy import concat_arrays_into as _concat_arrays_into
+except Exception:  # pragma: no cover
+    _concat_arrays_into = None
+
+DEFAULT_BUNDLE_CHUNK_BYTES = 64 * 1024**2
+AUTO_PARALLEL_MIN_BYTES = DEFAULT_BUNDLE_CHUNK_BYTES
```
zxpdemonio, line 29: This default change (512 MiB → 64 MiB) and tying `AUTO_PARALLEL_MIN_BYTES` to it will affect every caller that relied on the previous default, not only rollout transfer. Please highlight as a deliberate product default change in the PR body, or gate the 64 MiB default behind an explicit rollout helper / config so unrelated workloads keep prior chunking behavior.

mooncake-wheel/mooncake/structured_object_store.py
```diff
@@ -3729,7 +3756,134 @@ def _put_multi_buffer_payload(
...
+    def _put_direct_copy_payload(
+        self,
+        key: str,
+        value: _DirectCopyPayload,
+        chunk_bytes: int,
+        transfer_policy: BundleTransferPolicy,
+        config: Any = None,
+    ) -> tuple[dict[str, Any], list[str]]:
+        if transfer_policy.copy_mode == "zero_copy":
+            raise RuntimeError("zero-copy put requires tensor-object buffers")
```
zxpdemonio, line 3765: Good explicit rejection of `zero_copy` for direct-copy / non-tensor payloads. Please keep a regression test that names this exact error (already present for typed-ragged — good) and mention in the PR that source-buffer zero-copy for multi-buffer/typed-ragged puts is intentionally removed (tests for the old path were deleted).

mooncake-wheel/mooncake/structured_object_store.py
```diff
@@ -3729,7 +3756,134 @@ def _put_multi_buffer_payload(
...
+        num_chunks = len(chunk_batches)
+        chunk_keys = [
+            key if num_chunks == 1 else f"{key}/chunk/{idx}"
+            for idx in range(num_chunks)
+        ]
+        def _put_chunk_batch(keys, batches, nthreads=1):
+            for ck, (start, count, size) in zip(keys, batches):
+                lease = pool.acquire(size)
+                try:
+                    _concat_arrays_into(arrays, lease.ptr, start, count, nthreads)
```
zxpdemonio: Please assert `copied == size` (return value of `concat_arrays_into`) before `batch_put_from`. Today a short/long copy would still publish `size` bytes from the lease.

Also: when `_concat_arrays_into` is somehow None you never enter `_DirectCopyPayload` (guarded at encode time) — good. The concatenate fallbacks above for missing pool / oversize row are fine; consider a unit test that forces the oversize-row fallback (`ab > chunk_bytes`).

mooncake-wheel/setup.py
```diff
@@ -162,10 +162,26 @@ def finalize_options(self):
         self.plat_name = get_platform()
 
 
+
+# ---------------------------------------------------------------------------
+# C extensions
+# ---------------------------------------------------------------------------
+import numpy as np
+
+_fast_copy_ext = Extension(
```
zxpdemonio, line 171: Extension wiring looks fine (`-O2`, pthread, numpy headers). Soft-import fallback in Python is good for environments without a built `.so`.

One build note: `import numpy as np` at setup import time means `pip install` without numpy already present can fail earlier/more confusingly — ensure packaging docs/CI always install numpy before building the wheel.

yokinoshitayoki:
Should we have a look at the failed PRs?

zxpdemonio:
> Should we have a look at the failed PRs?

The failure has been resolved

mooncake-wheel/setup.py
```diff
@@ -162,10 +162,26 @@ def finalize_options(self):
         self.plat_name = get_platform()
 
 
+
+# ---------------------------------------------------------------------------
+# C extensions
+# ---------------------------------------------------------------------------
+import numpy as np
```
yokinoshitayoki: `setup.py` now imports NumPy at build time, but the NPU build uses `--no-isolation` and does not install NumPy. This fails in a clean environment with `ModuleNotFoundError`. Please install NumPy in the NPU build path or use an isolated build.

mooncake-wheel/tests/test_structured_object_store.py
```diff
@@ -3004,12 +2898,29 @@ def test_dataproto_helper_typed_ragged_uses_multi_buffer_put() -> None:
...
     assert ref.encoded_non_tensor["tokens"]["codec"] == "typed_ragged"
-    assert result["non_tensor_batch"]["tokens"].tolist() == [[1, 2], None, [3, 4, 5]]
+    tokens = result["non_tensor_batch"]["tokens"].tolist()
+    assert tokens == [[1, 2], None, [3, 4, 5]]
     assert rows[0].nbytes + rows[2].nbytes in pool.acquire_sizes
 
 
-def test_dataproto_helper_typed_ragged_zero_copy_put() -> None:
-    store, transfer = make_transfer()
+def test_typed_ragged_uses_direct_copy_payload_when_fast_copy_available() -> None:
```
yokinoshitayoki, line 2906: The native fast-copy test is skipped when the extension is not built, while the remaining tests can pass through the fallback path. Please add direct boundary tests for `concat_arrays_into` and run them against the built wheel in CI.

zxpdemonio:
Addressed this review in 8d05db5d. The NPU no-isolation wheel build path now installs NumPy before running the build backend, and CI installation testing now forces import of mooncake._fast_copy and runs direct concat_arrays_into boundary tests against the installed wheel. Local validation: native extension build passed, tests.test_fast_copy passed 4 tests, focused structured-object fast-copy tests passed 4 tests, bash -n passed, py_compile passed, code_format passed, and git diff --check passed.

mooncake-wheel/mooncake/structured_object_store.py
```diff
@@ -3729,7 +3750,138 @@ def _put_multi_buffer_payload(
...
+                        raise RuntimeError(
+                            f"native fast-copy wrote {copied} bytes, expected {size}"
+                        )
+                    results = _batch_put_from_with_optional_config(
+                        batch_put_from, [ck], [lease.ptr], [size], config
+                    )
+                    transport._check_batch_put_results(results, [ck], "batch_put_from")
+                except Exception:
+                    _cleanup_keys(self._store, [ck], strict=False)
+                    raise
```
yokinoshitayoki: On the sequential path, this handler removes only the currently failing key. If a later batch_put_from call fails, earlier chunks remain in the store because _put_direct_copy_payload raises before returning them to the outer written_keys list.

yokinoshitayoki:
And we may need to update the docs to align with the changes.

mooncake-wheel/mooncake/structured_object_store.py
zxpdemonio: Fixed in 050044ab. Cleanup now belongs to the whole direct-copy payload: sequential and parallel failures remove every candidate chunk key, while each worker only releases its lease. In the parallel path, executor shutdown completes before cleanup, so no running worker can recreate a removed chunk. Added a regression test that fails the second of three sequential batch_put_from calls and verifies that the store is empty and both acquired leases are released.

zxpdemonio:
> And we may need to update the docs to align with the changes.

Thanks and doc updated.

mooncake-wheel/setup.py
yokinoshitayoki: Please move the NumPy import to the top-level import section.

docs/source/api-reference/python/dataproto-structured-object-transfer.md
yokinoshitayoki: This zero-copy statement is now stale: non-tensor ndarray writes are staged and copied. Please update the documentation to match the new copy policy.

zxpdemonio:
@yokinoshitayoki  Thanks for the review. I have addressed the latest comments and CI is now green after rerunning the transient failed job.
Could you please take another look when you have time? Please let me know if there are any remaining concerns.

yokinoshitayoki (approved):

stmatengss (approved):

PR #3118, opened 2026-07-25
[Store] Selectively materialize BatchEvict candidates after timestamp filtering

jacklin78911-collab:
## Description

`MasterService::BatchEvict` currently scans the metadata and constructs a full
`Candidate`, including tenant and key identity, for every eligible object before
selecting the oldest requested subset.

This change keeps the metadata census O(N), but, for low eviction ratios,
selectively materializes full candidates only for the requested eviction
frontier plus a bounded reserve.

The implementation includes:

- a lightweight lease-timestamp census;
- selective candidate materialization for the target plus reserve;
- execution-time eligibility revalidation;
- reserve candidates for entries that fail revalidation;
- a refill scan when the reserve is exhausted;
- a high-ratio bypass that preserves the existing full-materialization path
  when most eligible candidates would be needed anyway.

Existing eviction semantics are preserved, including oldest-first selection,
soft-pin fallback, grouped eviction, replica constraints, target/lower-bound
handling, and offload behavior.

RFC: #3124  
Related performance investigation: #2560.

### Measured on this PR head

Earlier performance figures I had came from a separate research branch. These were measured on the code in this PR (head `22f7b6d0`) against its merge-base baseline (`6041a609`), so they describe what this diff actually delivers.

**Protocol.** Both versions built from unmodified checkouts with identical compiler, flags and optimization level; the benchmark harness is byte-identical on both commits (`mooncake-store/benchmarks/batch_evict_bench.cpp`, blob `dd44a22c`). Estimator, sample count and reporting rule were fixed before execution: 8 pairs per cell in ABBA order, 3 runs per arm with the arm value taken as the minimum, median paired difference with a bootstrap 95% CI, and a concurrent A/A null cell to calibrate the measurement floor. 197/197 processes accepted; no retries, replacements or discarded samples. Pinned CPUs, one process at a time, zero cgroup throttling observed.

**Results — 1M objects**

| Eviction ratio | Baseline | This PR | Median | Conservative bound (CI) |
|---|---|---|---|---|
| 1% | 288.1 ms | 88.3 ms | −69.9% | **−61.5%** |
| 10% | 529.7 ms | 285.1 ms | −47.1% | **−36.0%** |
| 30% (prebypass) | 898.9 ms | 895.0 ms | −0.04% | no difference detected |

The paired differences are one-sided heavy-tailed — the measurement noise on this host is kernel-side and can only lengthen a run, and the baseline arm is several times longer so it absorbs more of it. The CI lower bound is therefore the defensible figure, not the point estimate.

**Resolution.** The concurrent A/A null cell (both arms the baseline binary, 1% ratio) gives a measurement floor of ±9.4%. The 30% cell, where this PR takes the existing full-materialization path, resolves to ±15.2% at its longer runtimes; the honest statement there is that no regression larger than that was detected, and the point estimate is −0.04%.

Raw samples, per-run kernel counters and the analysis scripts are archived; happy to attach anything reviewers want to see.

### Deterministic mechanism measurements

Follow-up to the timing comment above. These are counter-based rather than timing-based: five repetitions per configuration produced bit-identical values for every count below, so they carry no measurement uncertainty. Instrumentation was counter-only on throwaway branches of both commits, with per-worker counters merged after the scan joins; eviction accounting matched the uninstrumented runs in every configuration.

**Identity materialization**

| Objects | Ratio | Baseline candidates | This PR | Avoided |
|---|---|---|---|---|
| 1M | 1% | 1,000,000 | 11,024 | 98.90% |
| 1M | 10% | 1,000,000 | 110,000 | 89.00% |
| 1M | 30% | 1,000,000 | 1,000,000 | 0% (prebypass) |
| 5M | 1% | 5,000,000 | 55,000 | 98.90% |
| 5M | 10% | 5,000,000 | 550,000 | 89.00% |

Selective candidate counts are exactly K + reserve in every cell; no timestamp-tie inflation occurred in this workload.

**Shard accessor acquisitions** — this addresses the natural question of whether the two-pass design increases lock traffic. With `kNumShards` = 1,024 the decomposition is exact:

```
baseline              = kNumShards + K
selective (this PR)   = kNumShards (census) + kNumShards (frontier) + K
prebypass (high ratio) = kNumShards + K          identical to baseline
```

The second pass therefore costs a fixed 1,024 additional acquisitions, independent of object count: 9.3% more at 1M/1%, and 0.2% more at 5M/10%. At high ratios the counts are identical, so prebypass takes the existing path rather than an approximation of it.

**Shard accessor hold time**, summed across workers and counting all phases including the frontier pass and execution: 30.3% of baseline at 1M/1%, 34.2% at 1M/10%, 20.3% at 5M/1%, 25.3% at 5M/10%. These are timings rather than counts and vary between runs; per-phase medians with ranges are in the archived report. At 30% the two versions are within run-to-run variation of each other, consistent with the identical acquisition counts.

**Memory, both directions.** The selective path avoids constructing identities but introduces a timestamp vector sized by the eligible population.

| | Identity bytes avoided | Candidate storage avoided (80 B each) | Timestamp vector peak | Net |
|---|---|---|---|---|
| 1M / 1% | 34.5 MB | 79.1 MB | 16.4 MB | +97.2 MB |
| 1M / 10% | 31.1 MB | 71.2 MB | 16.4 MB | +85.9 MB |
| 5M / 1% | 176.9 MB | 395.6 MB | 107.1 MB | +465.4 MB |
| 5M / 10% | 159.2 MB | 356.0 MB | 107.1 MB | +408.1 MB |

The timestamp vector costs 14–21% of what it avoids. Its peak is roughly twice the final vector because per-worker locals and the merged vector coexist. Allocator metadata and per-allocation overhead for the avoided `std::string` instances are not counted, so the saving column is conservative. Replacing the timestamp vector with a bounded histogram or buffer over the census would remove that cost entirely, and I'm happy to follow up with that if reviewers think it's worth doing.

**What this workload does not cover.** Every object was eligible for eviction — no pinned, unexpired, or replica-blocked objects — so behavior at lower eligible fractions is not characterized here. Lease timestamps were strictly increasing, so the timestamp-tie path that widens the frontier was never exercised. Refill never triggered, since the benchmark has no concurrent churn; the refill path's necessity is covered by the unit tests rather than by these measurements.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [x] Performance improvement
- [ ] Other

## How Has This Been Tested?

Added deterministic tests that invoke `BatchEvict` directly and cover:

- exact oldest-first eviction with distinct lease timestamps;
- exact target/lower-bound behavior;
- soft-pin fallback;
- whole-group eviction and group safety;
- the high-ratio full-materialization path;
- execution-stage failures absorbed by the reserve;
- reserve exhaustion followed by refill.

The reserve and refill tests use a grouped object with an active-lease member
to produce deterministic execution-time revalidation failures without test
hooks, sleeps, or background-thread timing.

**Test commands:**

```bash
cmake --build build-pr \
  --target batch_evict_test master_service_test offload_on_evict_test -j12

ctest --test-dir build-pr --output-on-failure \
  -R '^(batch_evict_test|master_service_test|offload_on_evict_test)$'
````

**Test results:**

* [x] Unit tests pass
* [ ] Integration tests pass (if applicable)
* [ ] Manual testing done (described below)

Results:

* `batch_evict_test`: 8/8 passed
* `master_service_test`: 144/144 passed
* `offload_on_evict_test`: 9/9 passed
* Total: 161/161 passed

## Checklist

* [x] I have performed a self-review of my own code
* [x] I have formatted my code using `./scripts/code_format.sh`
* [ ] I have run `pre-commit run --all-files` and all hooks pass
* [ ] I have updated the documentation (if applicable)
* [x] I have added tests to prove my changes are effective
* [x] For changes >500 LOC: I have filed an RFC issue

## AI Assistance Disclosure

* [ ] No AI tools were used
* [x] AI tools were used (specify below)

ChatGPT and Claude assisted with implementation, code review, test
construction, experiment orchestration, and debugging. Problem selection,
design constraints, experimental protocol, regression diagnosis, claim
boundaries, and final interpretation were determined by the author.

The author reviewed, tested, understands, and takes responsibility for all
submitted content.


zhangzuo21:
LGTM for me, please resolve conflicts, thanks!

jacklin78911-collab:
@zhangzuo21 Thanks for the review! Conflicts with `main` are resolved and the
branch is up to date. Rebuilt and re-ran the suites — `batch_evict_test`,
`master_service_test` and `offload_on_evict_test` all pass. The merge only
resolves conflicts; the algorithm is unchanged, so the measurements posted
above still describe this code.

ykwd:
CI failed
======================================================================
FAIL: test_human_readable_sizes (__main__.TestConfigDictSetup.test_human_readable_sizes)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/runner/work/Mooncake/Mooncake/mooncake-wheel/tests/test_distributed_object_store.py", line 65, in test_human_readable_sizes
    self.assertEqual(retcode, 0)
AssertionError: -1 != 0

jacklin78911-collab:
@ykwd Thank you for flagging this, and apologies for the slow reply — I've just had surgery and am in recovery, so I may be slower than usual for a little while.

Looking at the trace: the failure is `test_human_readable_sizes` in `mooncake-wheel/tests/test_distributed_object_store.py`, which exercises the Python-side config setup path, while this PR only touches the BatchEvict candidate-materialization path in `master_service.cpp`. I don't currently see a connection between the two, but I'd rather verify that than assume it — I'll check whether the same test is failing on `main` and on other recent PRs, and report back here.

If it does turn out to be related to this change, I'll fix it as soon as I'm able. Thanks again for your patience.

ykwd:
@jacklin78911-collab  No worries at all. Please take care of yourself and focus on your recovery first. 🙏

For this CI failure, @zhangzuo21 will help take a look and investigate. Thanks for checking on this despite your situation, and hope you have a smooth recovery!

jacklin78911-collab:
@ykwd Thank you, that's very kind — and thanks to @zhangzuo21 for picking this up.

I'm back and recovering well. Happy to help debug if it turns out to be related to this change, and otherwise just let me know if anything on the PR needs adjusting from my side.

jacklin78911-collab:
The failed job exited during checkout before any source was fetched: direct
GitHub checkout failed, and the ghfast.top fallback timed out after ~130s.
No build or test step ran. Re-running the failed job.

ykwd:
The CI failed again. This time it failed on SSD tests. @LujhCoconut is going to help take a look

LujhCoconut:
This is a CI race condition unrelated to this PR, and is addressed by PR #3212 （not merged yet）.

ykwd (approved):

PR #3110, opened 2026-07-24
[CI/Build] Fix CUDA stub release smoke test

Aionw:
## Description

Follow-up to #3106 for `release/v0.3.12`.

The installed-wheel smoke executes `mooncake_master --version`, but CUDA builder containers contain the link-time `libcuda.so` stub rather than the driver-provided `libcuda.so.1`. CUDA 12 and CUDA 13 release jobs therefore failed at this step on both x86_64 and aarch64.

This change:

- locates the toolkit stub in `lib64/stubs` or `targets/*/lib/stubs`;
- exposes it as `libcuda.so.1` only inside the temporary smoke environment;
- skips the alias for `NON_CUDA_BUILD`, preserving detection of accidental CUDA dependencies;
- continues executing the installed master binary, covering the original pre-`main()` ELF regression.

The alias is created after wheel repair and outside the artifact path. Smoke runs before artifact upload, and `publish-release` requires every build job to succeed, so a smoke failure blocks PyPI publication.

Failed runs: [CUDA 12](https://github.com/kvcache-ai/Mooncake/actions/runs/30108545321), [CUDA 13](https://github.com/kvcache-ai/Mooncake/actions/runs/30108545213).

EFA, MUSA, and Ascend use separate release workflows and remain outside this focused hotfix.

## Module

- [ ] Transfer Engine (`mooncake-transfer-engine`)
- [ ] Mooncake Store (`mooncake-store`)
- [ ] Mooncake EP (`mooncake-ep`)
- [ ] Mooncake PG (`mooncake-pg`)
- [ ] Integration (`mooncake-integration`)
- [ ] P2P Store (`mooncake-p2p-store`)
- [x] Python Wheel (`mooncake-wheel`)
- [ ] Common (`mooncake-common`)
- [ ] Mooncake RL (`mooncake-rl`)
- [x] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
# Build an executable with a CUDA driver NEEDED entry.
echo '#include <stdio.h>\nint main(void) { puts("entered main"); return 0; }' \
  | gcc -x c - -o /tmp/mooncake-cuda-loader-smoke \
      -L/usr/local/cuda-13.0/targets/x86_64-linux/lib/stubs \
[...]

PR #3119, opened 2026-07-25
[CI/Build] Remove CUDA dependency from mooncake_master

Aionw:
## Description

`mooncake-transfer-engine-non-cuda==0.3.12.post1` gives three packaged ELF files a `libcudart.so.12` dependency across CPython 3.10–3.13:

- `mooncake/mooncake_master`
- `mooncake/mooncake_client`
- `mooncake/store.so`

Only the master dependency is accidental. `mooncake_client` and `store.so` contain real CUDA staging references (`cudaMemcpy`, `cudaPointerGetAttributes`, and related APIs), preserving the behavior introduced by #1892 for GPU pointers that can reach Store even when global `USE_CUDA=OFF`. In contrast, `mooncake_master` has no undefined CUDA symbols: it inherits `CUDA::cudart` only because the static `mooncake_store` target propagates optional runtime dependencies to every consumer.

This focused fix:

- keeps Store's toolkit auto-detection and CUDA staging behavior unchanged;
- links the metadata-only `mooncake_master` with `--as-needed`, removing unused cudart from its `DT_NEEDED` entries;
- checks the installed master binary for `libcuda.so*`/`libcudart.so*` during regular non-CUDA wheel smoke tests;
- adds CPython 3.10 and 3.12 `NON_CUDA_BUILD` wheels to the nightly workflow, so the assertion runs nightly and during non-CUDA releases without adding a full wheel build to every PR.

This is narrower than #1839, which proposes general dynamic CUDA library loading. It also complements #3110/#3114: those changes handle the driver stub for CUDA wheel smoke tests, while this PR prevents the non-CUDA master from acquiring an unused CUDA runtime dependency.

## Module

- [x] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
# Configure the exact edge case: USE_CUDA=OFF while a CUDA toolkit is visible.
cmake -S . -B /tmp/mooncake-master-asneeded-build -G Ninja \
  -DCUDAToolkit_ROOT=/usr/local/cuda-12.2 \
  -DBUILD_UNIT_TESTS=OFF -DBUILD_BENCHMARK=OFF \
  -DUSE_HTTP=ON -DUSE_ETCD=OFF -DUSE_CUDA=OFF -DWITH_EP=OFF \
  -DSTORE_USE_ETCD=OFF -DENABLE_SCCACHE=OFF -DWITH_STORE_RUST=OFF \
  -DCMAKE_BUILD_TYPE=Release
cmake --build /tmp/mooncake-master-asneeded-build \
  --target mooncake_master mooncake_client store --parallel 128

# Execute master and inspect all three affected targets.
LD_LIBRARY_PATH=/tmp/mooncake-master-asneeded-build/mooncake-common:/usr/local/lib \
  /tmp/mooncake-master-asneeded-build/mooncake-store/src/mooncake_master --version
readelf -d /tmp/mooncake-master-asneeded-build/mooncake-store/src/mooncake_master
readelf -d /tmp/mooncake-master-asneeded-build/mooncake-store/src/mooncake_client
readelf -d /tmp/mooncake-master-asneeded-build/mooncake-integration/store.cpython-310-x86_64-linux-gnu.so

UV_CACHE_DIR=/tmp/mooncake-uv-cache uvx --from pytest --with pyyaml \
  pytest mooncake-wheel/tests/test_release_wheel_tags.py -q
pre-commit run --files \
  .github/workflows/_build-wheel.yaml .github/workflows/nightly.yml \
  mooncake-store/src/CMakeLists.txt
git diff --check
```

**Test results:**
- [ ] Unit tests pass

The local configuration reported `mooncake_store: CUDAToolkit detected, enabling D2H staging` despite `USE_CUDA=OFF`, reproducing the release-builder condition. All three targets built successfully. `mooncake_master --version` ran without CUDA on its library path, and `readelf` showed:

- `mooncake_master`: no CUDA dependency
- `mooncake_client`: `libcudart.so.12` retained
- `store.so`: `libcudart.so.12` retained

The targeted workflow test for `nightly.yml:build-non-cuda-x86` passed, and all targeted pre-commit hooks passed. The full workflow configuration suite currently has one unrelated failure from the newly added bare-runner `nightly.yml:build-cuda128-arm64` job on `main`.

## Checklist

Documentation and an RFC are not applicable to this focused release/build fix.

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex inspected the `v0.3.12.post1` artifacts, distinguished the master's unused transitive dependency from Store's intentional CUDA staging symbols, implemented the scoped linker and CI changes, and ran the validations listed above. The human submitter is responsible for reviewing and defending every changed line.

### DIFF
--- a/.github/workflows/_build-wheel.yaml
+++ b/.github/workflows/_build-wheel.yaml
@@ -197,6 +197,20 @@ jobs:
           "$smoke_venv/bin/python" -m pip install --no-deps \
             mooncake-wheel/dist-py${{ steps.pytag.outputs.tag }}/*.whl
 
+          if [ "${VARIANT_FLAG:-}" = "NON_CUDA_BUILD" ]; then
+            site_packages=$("$smoke_venv/bin/python" -c \
+              'import sysconfig; print(sysconfig.get_paths()["purelib"])')
+            while IFS= read -r -d '' binary; do
+              file "$binary" | grep -q ELF || continue
+              if readelf -d "$binary" | grep -Eq \
+                'Shared library: \[(libcuda|libcudart)\.so'; then
+                echo "Non-CUDA wheel binary depends on CUDA: $binary"
+                readelf -d "$binary" | grep 'Shared library:'
+                exit 1
+              fi
+            done < <(find "$site_packages" -type f -print0)
+          fi
+
           export LD_LIBRARY_PATH="/usr/local/lib:${LD_LIBRARY_PATH:-}"
           for dir in /usr/local/cuda/lib64 \
                      /usr/local/cuda/lib64/stubs \
--- a/.github/workflows/_build-wheel.yaml
+++ b/.github/workflows/_build-wheel.yaml
@@ -200,15 +200,13 @@ jobs:
           if [ "${VARIANT_FLAG:-}" = "NON_CUDA_BUILD" ]; then
             site_packages=$("$smoke_venv/bin/python" -c \
               'import sysconfig; print(sysconfig.get_paths()["purelib"])')
-            while IFS= read -r -d '' binary; do
-              file "$binary" | grep -q ELF || continue
-              if readelf -d "$binary" | grep -Eq \
-                'Shared library: \[(libcuda|libcudart)\.so'; then
-                echo "Non-CUDA wheel binary depends on CUDA: $binary"
-                readelf -d "$binary" | grep 'Shared library:'
-                exit 1
-              fi
-            done < <(find "$site_packages" -type f -print0)
+            master="$site_packages/mooncake/mooncake_master"
+            if readelf -d "$master" | grep -Eq \
+              'Shared library: \[(libcuda|libcudart)\.so'; then
+              echo "Non-CUDA mooncake_master depends on CUDA"
+              readelf -d "$master" | grep 'Shared library:'
+              exit 1
+            fi
           fi
 
           export LD_LIBRARY_PATH="/usr/local/lib:${LD_LIBRARY_PATH:-}"
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -102,8 +102,23 @@ jobs:
       artifact-prefix: nightly-cuda13-x86
       version-override: ${{ needs.version-stamp.outputs.nightly_version }}
 
+  build-non-cuda-x86:
+    needs: version-stamp
+    uses: ./.github/workflows/_build-wheel.yaml
+    with:
+      runner: ubuntu-22.04
+      container: pytorch/manylinux2_28-builder:cuda12.8
+      python-versions: '["3.10", "3.12"]'
+      cuda: container
+      variant-flag: NON_CUDA_BUILD
+      cmake-args: >-
+        -DBUILD_UNIT_TESTS=OFF -DUSE_HTTP=ON -DUSE_ETCD=ON -DUSE_CUDA=OFF -DWITH_EP=OFF
+        -DSTORE_USE_ETCD=ON -DENABLE_SCCACHE=ON -DCMAKE_BUILD_TYPE=Release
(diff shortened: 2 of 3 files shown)


Icedcoco:
Thanks for the focused fix. I have two questions before this lands:

First, could you clarify the intended behavior for Store CUDA staging when `USE_CUDA=OFF` but `CUDAToolkit` is available? #1892 deliberately enabled staging independently of `USE_CUDA`, since GPU pointers could still reach `PutToLocalFile` through configurations such as `nvidia-peermem`. With the new condition, it looks like the CUDA device would no longer be registered, so such pointers could fall through to the host-pointer path and be dereferenced by `std::string::append`. Is that configuration no longer intended to be supported? If it should remain supported, would it be safer to disable CUDA discovery only for `NON_CUDA_BUILD` packaging, for example with `CMAKE_DISABLE_FIND_PACKAGE_CUDAToolkit`, rather than changing the shared CMake behavior?

Second, do we want the new ELF assertion to be exercised by PR CI? The check only runs when `VARIANT_FLAG=NON_CUDA_BUILD`, while the non-CUDA wheel job currently visible in this PR uses `EFA_NON_CUDA_BUILD` through a separate workflow. Unless I missed another job, the new assertion therefore does not seem to run in this PR. Would it make sense to run one regular `NON_CUDA_BUILD` Python version in PR CI, or is this intentionally meant to remain a release-only guard?

Aionw:
Thanks — both points were correct, and I’ve updated the PR in 3bf8ec6f after rebasing onto the latest main.

For the first point, the Store behavior from #1892 is preserved. I restored CUDAToolkit auto-detection independently of the global `USE_CUDA` option, so CUDA staging remains available when GPU pointers arrive through configurations such as `nvidia-peermem`.

I also inspected the v0.3.12.post1 wheel: `mooncake_client` and `store.so` contain real CUDA references, while `mooncake_master` had a `libcudart` `DT_NEEDED` entry but no undefined CUDA symbols. The fix is now scoped to the metadata-only master: it links with `--as-needed`, which removes only the unused CUDA runtime dependency.

I reproduced the relevant configuration locally with `USE_CUDA=OFF` and CUDAToolkit visible. CMake still reported CUDA staging enabled; after building all three targets, `readelf` showed:

- `mooncake_master`: no CUDA dependency
- `mooncake_client`: `libcudart.so.12` retained
- `store.so`: `libcudart.so.12` retained

For the second point, after rebasing I found the new nightly wheel workflow from #2950. I moved the regular `NON_CUDA_BUILD` coverage there instead of adding another full wheel build to every PR. The nightly job builds CPython 3.10 and 3.12 non-CUDA wheels, publishes them through the existing TestPyPI path, runs the master ELF assertion, and participates in `nightly-gate`. The same assertion also remains active in the release non-CUDA workflow. This is therefore intentionally a nightly/release packaging guard rather than a PR gate.

The targeted workflow test for the new nightly non-CUDA job passes, and pre-commit passes on all changed files. (The full workflow test currently has one unrelated failure from the newly added bare-runner `nightly.yml:build-cuda128-arm64` job.) Thanks for catching the scope issue.

Icedcoco (approved):
LGTM from my side. The previous concerns have been addressed: Store’s CUDA staging behavior is preserved, and the fix is now correctly scoped to `mooncake_master`.

The remaining caveat is that the non-CUDA package as a whole is not yet CUDA-runtime-free, since `mooncake_client` and `store.so` still depend on `libcudart`. That is outside the scope of this focused fix and can be tracked separately.

ykwd (approved):

PR #3125, opened 2026-07-26
[Bugfix] Fix double free of RDMA slice on device-selection failure

SongOf:
## Description

  `RdmaTransport::submitTransferTask()`'s `!found_device` error path manually
  deallocated every slice already queued in the function-local
  `slices_to_post` accumulator. Those slices are unconditionally recorded in
  their owning `TransferTask::slice_list` as soon as they are allocated, and
  `~TransferTask()` already returns every slice in `slice_list` to
  `ThreadLocalSliceCache` exactly once. Deallocating them a second time here
  double-frees them into the cache, so a later `allocate()` can hand the same
  `Slice*` out to two unrelated transfers, eventually crashing with glibc's
  "double free detected in tcache 2".

  Fix: drop the manual deallocation in the error path and let the caller's
  `TransferTask` destructor own the single free, as it already does for every
  other slice.

  Also extracts the pre-existing `RdmaTransportTestPeer` /
  `RdmaContextTestPeer` friend-accessor test helpers (previously duplicated
  inside `rdma_context_reprobe_test.cpp`) into a shared
  `rdma_test_peers.h`, reused by the new test file.

  ## Module
  
  - [ ] CI/CD
  - [ ] Docs
  - [ ] Other
  
  ## Type of Change

  - [x] Bug fix
  - [ ] New feature
  - [ ] Refactor
  - [ ] Breaking change
  - [ ] Documentation update
  - [ ] Performance improvement
  - [ ] Other

  ## How Has This Been Tested?

  Built and run inside a Linux (Ubuntu 22.04) Docker container with the
  project's `dependencies.sh`, since this macOS sandbox lacks libibverbs and
  other Linux-only APIs the transfer engine depends on.

  **Test commands:**
  ```bash
  cd build
  ninja rdma_transport_submit_task_test
  ./mooncake-transfer-engine/tests/rdma_transport_submit_task_test \
      --gtest_filter='*FoundDeviceFailureFreesSliceTwice*'

  Test results:
  - [x] Unit tests pass

  The new SubmitTransferTaskTest.FoundDeviceFailureFreesSliceTwice test
  drives the real, unmodified submitTransferTask() entry point (no RDMA
  hardware needed, via a bare/unconstructed RdmaContext) and proves the
  double free purely by pointer comparison: after one task hits the
  !found_device path and is destroyed, a second task's two independent
  slice allocations must come back as two distinct Slice*. Verified this
  test FAILS on the pre-fix code (reliably reproducing glibc's "free():
  double free detected in tcache 2" crash) and PASSES cleanly after the fix.
  Also re-ran the existing rdma_context_reprobe_test after extracting the
  shared test-peer helpers to confirm no regression.

  Checklist

  - [ ] For changes >500 LOC: I have filed an RFC issue (not applicable, ~10
  LOC in the actual source fix)

  AI Assistance Disclosure
  
  - [ ] No AI tools were used

  Used Claude Code to write and iteratively verify the reproducing test
  in a Linux container, and implement the fix. All changes reviewed line by
  line before commit.

### DIFF
--- a/mooncake-transfer-engine/src/transport/rdma_transport/rdma_transport.cpp
+++ b/mooncake-transfer-engine/src/transport/rdma_transport/rdma_transport.cpp
@@ -651,8 +651,14 @@ Status RdmaTransport::submitTransferTask(
             }
             if (!found_device) {
                 auto source_addr = slice->source_addr;
-                for (auto &entry : slices_to_post)
-                    for (auto s : entry.second) getSliceCache().deallocate(s);
+                // Do not deallocate slices already queued in slices_to_post
+                // here: every slice, including those, is already recorded in
+                // its owning TransferTask::slice_list (unconditionally,
+                // right after allocation above), and ~TransferTask() returns
+                // everything in slice_list to the cache exactly once.
+                // Deallocating them again here double-frees them into
+                // ThreadLocalSliceCache, letting a later allocate() hand out
+                // the same Slice* to two unrelated transfers.
                 LOG(ERROR)
                     << "Memory region not registered by any active device(s): "
                     << source_addr;
--- a/mooncake-transfer-engine/tests/CMakeLists.txt
+++ b/mooncake-transfer-engine/tests/CMakeLists.txt
@@ -293,6 +293,13 @@ target_link_libraries(rdma_context_reprobe_test PUBLIC transfer_engine gtest
                                                        gtest_main)
 add_test(NAME rdma_context_reprobe_test COMMAND rdma_context_reprobe_test)
 
+add_executable(rdma_transport_submit_task_test
+               ${WORKSPACE}/rdma_transport_submit_task_test.cpp)
+target_link_libraries(rdma_transport_submit_task_test PUBLIC transfer_engine
+                                                             gtest gtest_main)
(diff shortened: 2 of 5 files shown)


alogfans (approved):
LGTM. But a warning like 'leak some slices' may be printed.

SongOf:
> LGTM. But a warning like 'leak some slices' may be printed.

@alogfans 
Confirmed, this is real but it's a separate, pre-existing issue from the double-free this commit fixes, not something this change introduces or worsens.

  Mechanism: slices already queued in slices_to_post when a later slice hits !found_device (or the Device ... is not active branch) have already had their owning task's slice_count incremented, but since we return before the
  end-of-function flush, submitPostSend() is never called for them. No RDMA completion will ever arrive, so success_slice_count + failed_slice_count never reaches slice_count, getTransferStatus() reports WAITING forever, and
  freeBatchID() returns BatchBusy indefinitely — the batch (and all its slices) stays around until process exit.

  This existed before this fix too (the old code's manual deallocate() loop freed the slices' memory but didn't fix the completion-accounting problem, and even double-freed them on top). I'll track it as a follow-up rather than fold it into this PR, since it's a distinct root cause (missing flush-before-return) from the double-free this commit addresses.

PR #2958, opened 2026-07-16
[TransferEngine] Add standard Prometheus metrics to Classic TE(1/2)

xiangui33423:
## Description

Adds standard Prometheus-style metrics to the **Classic Transfer Engine**,
closing the observability gap with TENT. This is the **first, intentionally
small PR** for #2713 — it does **not** unify the Classic TE and TENT metrics
implementations or refactor the architecture; that work is deferred to PR2.

Before this change, the Classic TE `WITH_METRICS` path was **log-only**: it
accumulated a bytes counter and a latency histogram and periodically printed a
throughput/latency line (gated on `MC_TE_METRIC`). There were no
standard-named counters, no inflight gauge, no failure/request counters, and no
HTTP export.

**What this PR adds**

A new `TransferEngineMetrics` singleton
(`mooncake-transfer-engine/include/transfer_engine_metrics.h` +
`src/transfer_engine_metrics.cpp`) that mirrors TENT's metric naming and
behavior:

| Metric | Type | Meaning |
|--------|------|---------|
| `transfer_requests_total` | Counter | Total transfer tasks submitted |
| `transfer_bytes_total` | Counter | Total bytes transferred (successful tasks) |
| `transfer_failures_total` | Counter | Tasks that failed / canceled / timed out |
| `transfer_latency_seconds` | Histogram | Completion latency (successful tasks) |
| `transfer_size_bytes` | Histogram | Transfer request size distribution |
| `inflight_transfers` | Gauge | Transfers submitted but not yet terminal |

**Design highlights**

- **Exactly-once recording** reuses the existing `TransferEngineImpl` anchors:
  `recordSubmitted()` fires once per task when `start_time` is first set at
  submit; the terminal branch of `getTransferStatus()` records exactly one
  `recordCompleted()`/`recordFailed()` and immediately resets `start_time`, so
  repeated status polling never double-counts. FAILED/CANCELED/TIMEOUT all map
  to a failure and clear the inflight slot.
- **HTTP export is opt-in.** A coro_http server exposes the same endpoints as
  TENT (`/metrics`, `/metrics/summary`, `/metrics/json`, `/health`) and only
  starts when `MC_TE_METRIC_HTTP_PORT` is set, so default behavior is
  unchanged. `transfer_engine` already links `yalantinglibs`, so no new
  dependency is introduced.
- **Existing logging is preserved unchanged** — the new metrics are purely
  additive.
- Gated by the existing `WITH_METRICS` compile flag (fully compiled out
[...]

PR #3126, opened 2026-07-26
[tent] Fix Prometheus histogram silent-drop and clean stale config keys

staryxchen:
## Description

`getPrometheusMetrics()` used ylt's `basic_dynamic_histogram::serialize()`, which silently drops the entire metric (including the `# HELP` / `# TYPE` header it already wrote) whenever every label combo has `sum_==0`. This condition is reachable in production: `recordStageLatency()` casts `double latency_us` to `int64_t`, so sub-microsecond latencies (loopback TCP or SHM `queue_wait`) truncate to 0 and `sum_` stays at 0.

A `tebench` TCP run surfaced this: `tent_stage_queue_wait_us` reported `count=4846` in `/metrics/json` but was **completely absent** from `/metrics`. The JSON endpoint's custom bucket-walk was unaffected, so the two endpoints silently drifted.

**Fix:**
- Add `serializeHistogramPrometheus()` that walks `get_bucket_counts()` + `copy()` directly — the same data source the JSON endpoint already uses. A histogram with ≥1 observed sample now always appears in Prometheus output.
- This also unifies the two endpoints on one bucket-walk code path, eliminating the structural drift risk.
- ylt's `sum_` is private with no public accessor, so `_sum` is emitted as `0` (matches JSON, which also omits sum; the true sum is already lost at record time for sub-µs latencies).

**Also:** removed the deprecated `metrics/latency_buckets` and `metrics/size_buckets` keys from the sample `transfer-engine.json`. `docs/source/design/tent/metrics.md` already declares them removed and silently ignored, but the sample config still carried them, misleading users who copy it.

No related issue found; not overlapping with open PRs #2958 (Classic TE metrics) or #3109 (failover attribution).

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Unit/regression tests** (built with `-DTENT_METRICS_ENABLED=ON -DUSE_TENT=ON`, run 3× stable):
- `MetricsHttpTest.PrometheusExposesHistogramWhenAllSamplesAreZero` — reproduces the bug condition (sub-µs latencies → `int64_t` 0 → `sum_==0`) with `transport=unspec`; asserts the metric appears in `/metrics` with count=3.
- `MetricsHttpTest.PrometheusExposesZeroValuedHistogramForTcpTransport` — same, with `transport=tcp` (the tebench scenario that surfaced the bug).
- `MetricsHttpTest.JsonAndPrometheusAgreeOnHistogramCount` — delta-based assertion that both endpoints agree on `_count`.
- `MetricsConfigLoaderTest.StaleBucketKeysAreSilentlyIgnored` — stale keys do not influence `MetricsConfig`.

Suite results: `tent_metrics_recording_test` 14/14 PASS, `metrics_config_loader_test` 26/26 PASS.

**Test commands:**
```bash
cmake --build build-tent-metrics --target tent_metrics tent_metrics_recording_test metrics_config_loader_test
./build-tent-metrics/tent/tests/tent_metrics_recording_test
./build-tent-metrics/tent/tests/metrics_config_loader_test
```

**Manual / end-to-end testing done:**

`tebench` over TCP (`--xport_type=tcp`, 1MB block, 4 threads, write, DRAM) — scraped `/metrics` and `/metrics/json` mid-flight. Before fix: `stage_queue_wait_us` was 0 lines in Prometheus despite JSON `count=4846`. After fix: 15 lines in Prometheus, `count` and cumulative `+Inf` bucket match JSON (within scrape-time delta).

Same flow also verified over SHM (`--xport_type=shm`) — `transport="shm"` label correct, `queue_wait` appears in Prometheus, JSON/Prometheus consistent.

## Checklist

- [x] I have updated the documentation (if applicable) — docs already declared the keys removed; only the sample config was out of sync, now fixed
- [ ] For changes >500 LOC: I have filed an RFC issue — N/A (5 files, +288/-19)

## AI Assistance Disclosure

- [ ] No AI tools were used
- [x] AI tools were used

AI tools were used to: explore the codebase and ylt histogram internals, implement `serializeHistogramPrometheus`, write the regression tests, and run `tebench` end-to-end verification. The human submitter reviewed every changed line and re-ran the tebench verification flow independently to confirm the metric appears in Prometheus output.


mooncake-transfer-engine/tent/tests/metrics_config_loader_test.cpp
```diff
@@ -188,6 +188,45 @@ TEST(MetricsConfigLoaderTest, LoadWithDefaultsEnvOverridesDefault) {
...
+    // carrying both real keys and the removed stale bucket keys.
+    std::string json_content = R"({
+        "metrics": {
+            "enabled": true,
+            "http_port": 7777,
+            "http_host": "0.0.0.0",
+            "http_server_threads": 2,
+            "report_interval_seconds": 15,
+            "latency_buckets": [0.001, 0.002, 0.005],
+            "size_buckets": [1024, 4096, 16384]
```
alogfans, line 210: As latency_buckets and size_buckets seem no longer be used (according to the modification of default config), you can remove them safely.

staryxchen:
Good catch on the stale `latency_buckets` / `size_buckets` keys.

These two keys are actually the **input** of this test (`StaleBucketKeysAreSilentlyIgnored`), not dead code — they are deliberately kept in the test JSON to verify the loader silently ignores them while still honoring the real keys alongside. Removing them would collapse this regression guard into a plain "real keys load" test.

That said, the original comment overclaimed a "compile-error guard" that C++ cannot actually provide. I have:

1. **Kept** `latency_buckets` / `size_buckets` in the test JSON (they are the test's input).
2. **Corrected the comment** — dropped the inaccurate "future regression will turn into a compile error" claim, and instead states plainly that the guard is behavioral (loader ignores unknown keys; real keys still honored). If a future change re-introduces a buckets field on `MetricsConfig`, the test should be extended with an explicit assertion on the loaded value.

Pushed as `899330ce`. Tests still pass (26/26 in `metrics_config_loader_test`).

mooncake-transfer-engine/tent/src/metrics/tent_metrics.cpp
```diff
@@ -357,6 +363,104 @@ int64_t sumCounterValues(CounterPtr counter) {
...
+            out.append(std::to_string(cumulative)).append("\n");
+        }
+
+        // _sum: ylt keeps sum_ private with no public accessor, so the real
+        // per-label sum cannot be read here. Emit 0 so the metric type shape
+        // matches Prometheus expectations (_bucket + _sum + _count). This
+        // matches the JSON endpoint, which also omits sum. Note the true sum
+        // is already lost at record time for sub-microsecond latencies
+        // (observe() truncates to int64_t). A future ylt change exposing
+        // sum_ would let us populate this accurately.
```
alogfans: Emitting `_sum` as `0` reports incorrect data for non-zero observations and breaks queries such as `_sum / _count`.

staryxchen: You are right — `_sum` emitted as `0` was incorrect for any metric with non-zero observations (read_size, write_size, latency_us, etc.). The original comment claiming "true sum is already lost at record time" only held for sub-microsecond latencies cast to `int64_t`, not for the general case. Fixed in `17101d8a`.

ylt's `basic_dynamic_histogram::sum_` is private with no public accessor, so `sum_->copy()` (which ylt uses internally for serialization) is not reachable from outside. To recover the real sum without patching the extern dependency, I added a parallel `basic_dynamic_counter<int64_t, 1>` per histogram (`read_latency_sum_`, `write_size_sum_`, etc.), incremented alongside every `observe()` call in `recordReadCompleted` / `recordWriteCompleted` / `recordDeadlineMLU` / `recordStageLatency`. `serializeHistogramPrometheus` reads it back via `copy()` and emits the true per-label sum; the JSON endpoint was also missing `sum` and now emits it.

Verified with `tebench --op_type=mix --xport_type=shm` (4 threads, 1MB blocks):

| metric | `_sum` | `_count` | `_sum/_count` |
|---|---|---|---|
| `tent_read_size_bytes` | 186133774336 | 177615 | 1048069 ≈ 1MB block |
| `tent_write_size_bytes` | 90485817344 | 86398 | 1047308 ≈ 1MB block |
| `tent_read_latency_us` | 9235021 | 177615 | 52.0 us |
| `tent_stage_transport_us` | 387965 | 264017 | 1.5 us |
| `tent_stage_queue_wait_us` | 0 | 264015 | 0 (sub-us truncation) |

`_sum / _count` now produces correct averages. `stage_queue_wait_us_sum=0` is the known sub-microsecond cast truncation (same behavior as ylt's internal `sum_`, not a regression).

alogfans (approved):
LGTM

PR #3127, opened 2026-07-26
[tent] Support mixed DRAM+VRAM segments and multi-transport allocation in tebench

staryxchen:
## Description

tebench could only allocate buffers of a single `seg_type` (DRAM or VRAM) and only enable a single transport per run (`--xport_type`). This prevented exercising mixed-transport workloads where the engine's `transport_selector` naturally picks SHM for DRAM and NVLink for VRAM within one process.

This change adds `--seg_type_mix=dram,vram` so a single tebench process can register a mixed DRAM+VRAM segment and drive concurrent traffic over both memory types, letting the engine select transports based on buffer location (real production behavior, not an artificial `transport_hint` override).

It also fixes a latent bug in `allocateBuffers` where the `shm_path` set by `ShmTransport::allocateLocalMemory` was silently dropped before `registerLocalMemory`, causing SHM to install but never register its buffers — silently falling back to TCP.

### Changes

**1. `allocateBuffers()` (`benchmark/tent_backend.cpp`)**
- Use the `MemoryOptions&` overload of `allocateLocalMemory` so transports like SHM (which set `options.shm_path` inside `allocateLocalMemory`) have those fields carried into `registerLocalMemory`. The previous no-options overload silently dropped `shm_path`, causing SHM to install but never register its buffers — silently falling back to TCP.
- Reset `options.type` to `UNSPEC` before `registerLocalMemory` so the buffer registers to ALL enabled transports (via `getSupportedTransports(UNSPEC)`), not just the one that allocated it. This is what makes multi-transport work: a DRAM buffer ends up registered to both SHM (has `shm_path`) and TCP (does not need it), so requests resolve regardless of which transport the engine picks.
- Loop over `seg_type_mix` to allocate both DRAM and VRAM buffers into one `pinned_buffer_list_`, tracking seg_type per buffer in a new `pinned_buffer_seg_type_` vector.

**2. `allocateLocalMemory(MemoryOptions&)` (`tent/src/runtime/transfer_engine_impl.cpp`)**
- When `options.type` is `UNSPEC`, mirror the `Location`-based overload's preference: CPU memory picks SHM > RDMA > TCP; non-CPU (cuda/rocm) picks NVLink > MNNVL > RDMA > TCP. Previously this UNSPEC branch skipped SHM entirely and non-CPU skipped NVLink, which meant a caller passing `MemoryOptions{type=UNSPEC}` (e.g. tebench without `--xport_type`) would allocate via TCP even when SHM was enabled and preferred, and VRAM buffers could not be allocated via NVLink.

**3. New `--seg_type_mix` flag (`benchmark/utils.cpp` / `utils.h`)**
- Comma-separated list of segment types, e.g. `"dram,vram"`. When set, target registers buffers of each listed type in one segment, and initiator threads round-robin across them so a single tebench process drives traffic over multiple memory types (and thus multiple transports — SHM for DRAM, NVLink for VRAM) concurrently. Empty falls back to `--seg_type` (single type, existing behavior). Transports are enabled via `MC_TENT_CONF` config file.

**4. `getLocalBufferBase` / `getTargetBufferBase` (`benchmark/tent_backend.h`)**
- In mixed mode, select buffer by the thread's assigned seg_type (round-robin by `thread_id`), then index within the filtered subset. Target buffers are filtered by location prefix (`"cpu:"` vs `"cuda:"`).

**5. Documentation (`docs/source/design/tent/tebench.md`)**
- Added `--seg_type_mix` flag description to Section 5.4.
- Expanded `--xport_type` in Section 5.6 with `MC_TENT_CONF` multi-transport note.
- Added new Section 5.8 documenting the mixed DRAM+VRAM + multi-transport scenario with usage examples.

No related issue found; not overlapping with open PRs.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Unit test regression** (built with `-DTENT_METRICS_ENABLED=ON -DUSE_TENT=ON -DUSE_CUDA=ON`):
- Existing `tent_metrics_recording_test` and `metrics_config_loader_test` still pass (14/14 + 26/26).

**Manual / end-to-end testing done:**

1. **Mixed DRAM+VRAM + SHM+NVLink** (`--seg_type_mix=dram,vram --local_gpu_id=-1`, 8 threads, 30s): `/metrics` shows both `transport="shm"` (216 GB / 206k reqs) and `transport="nvlink"` (320 GB / 306k reqs). Engine selects transport by buffer location — NVLink for VRAM, SHM for DRAM. 0 failures, 0 failovers. `stage_queue_wait_us` correctly appears for both transports.

2. **Single-seg_type regression — TCP** (`--xport_type=tcp --seg_type=DRAM`, no `--seg_type_mix`): only `transport="tcp"` label, 6.28 GB / 5990 reqs, 0 failures. Behavior unchanged from before.

3. **Single-seg_type regression — SHM** (`--xport_type=shm --seg_type=DRAM`, no `--seg_type_mix`): only `transport="shm"` label, 311 GB / 296k reqs, 0 failures. Behavior unchanged from before.

**Test commands:**
```bash
cmake -S . -B build -DUSE_TENT=ON -DTENT_METRICS_ENABLED=ON -DUSE_CUDA=ON
cmake --build build --target tebench
# Mixed scenario (requires GPU + NVLink hardware):
MC_TENT_CONF=/path/to/tent_shm_nvlink.json ./build/.../tebench --backend=tent \
  --metadata_type=p2p --seg_name=tgt --seg_type_mix=dram,vram --local_gpu_id=-1
# Single-seg_type regression:
./build/.../tebench --backend=tent --xport_type=tcp --seg_type=DRAM
./build/.../tebench --backend=tent --xport_type=shm --seg_type=DRAM
```

## Checklist

- [x] I have updated the documentation (if applicable) — added `--seg_type_mix` to `docs/source/design/tent/tebench.md` Section 5.4, expanded `--xport_type` in 5.6 with `MC_TENT_CONF` note, added new Section 5.8 with mixed DRAM+VRAM + multi-transport usage examples
- [x] I have added tests to prove my changes are effective — N/A, this is a tebench tooling change; existing engine unit tests still pass, and end-to-end verification is documented above
- [x] For changes >500 LOC: I have filed an RFC issue — N/A (6 files, +288/-56)

## AI Assistance Disclosure

- [ ] No AI tools were used
- [x] AI tools were used

AI tools were used to: explore the codebase and transport selection logic, implement the `allocateBuffers` mixed-mode and `--seg_type_mix` flag, fix the `allocateLocalMemory` UNSPEC branch, update the tebench documentation, and run `tebench` end-to-end verification (mixed DRAM+VRAM, single-seg_type regression, NVLink vs SHM performance analysis). The human submitter reviewed every changed line and re-ran the tebench verification flow independently.

### DIFF
--- a/mooncake-transfer-engine/benchmark/tent_backend.cpp
+++ b/mooncake-transfer-engine/benchmark/tent_backend.cpp
@@ -114,20 +114,16 @@ static IntentType getIntentType(const std::string& intent_type) {
     return it->second;
 }
 
-int TENTBenchRunner::allocateBuffers() {
-    const auto total_buffer_size = XferBenchConfig::total_buffer_size;
-    const auto& seg_type = XferBenchConfig::seg_type;
-    const auto& xport_type = XferBenchConfig::xport_type;
-
-    // Resolve device prefix, start index, and buffer count per seg_type
-    std::string device_prefix;
-    int start_idx = 0, num_buffers = 0;
-
-    if (seg_type == "DRAM") {
+// Resolve device prefix, start index, and buffer count for a single seg_type.
+// Returns 0 on success, -1 on failure.
+static int resolveSegTypeParams(const std::string& seg_type,
+                                std::string& device_prefix, int& start_idx,
+                                int& num_buffers) {
+    if (seg_type == "DRAM" || seg_type == "dram") {
         device_prefix = "cpu";
         num_buffers = numa_num_configured_nodes();
 #if defined(USE_CUDA) || defined(USE_SUNRISE)
-    } else if (seg_type == "VRAM") {
+    } else if (seg_type == "VRAM" || seg_type == "vram") {
         device_prefix = "cuda";
         int gpu_count = 0;
         auto err = cudaGetDeviceCount(&gpu_count);
@@ -143,7 +139,7 @@ int TENTBenchRunner::allocateBuffers() {
                 << gpu_count << ")";
         }
 #elif defined(USE_HIP)
-    } else if (seg_type == "VRAM") {
+    } else if (seg_type == "VRAM" || seg_type == "vram") {
         device_prefix = "rocm";
         int gpu_count = 0;
         hipGetDeviceCount(&gpu_count);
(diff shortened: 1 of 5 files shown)


mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
alogfans, line 1: We believe this change to the core allocation logic is unnecessary for the benchmark feature and should be reverted.

mooncake-transfer-engine/benchmark/tent_backend.h
```diff
@@ -59,15 +59,59 @@ class TENTBenchRunner : public BenchRunner {
...
+        if (seg_type_mix_.empty()) {
+            // Single seg_type mode: original behavior.
+            return info_.buffers[thread_id % info_.buffers.size()].base +
+                   block_size * batch_size * (thread_id / info_.buffers.size());
+        }
+        // Mixed mode: pick this thread's seg_type, then index into the
+        // subset of target segment buffers whose location prefix matches.
+        const std::string& seg_type =
+            seg_type_mix_[thread_id % seg_type_mix_.size()];
+        std::string prefix = (seg_type == "vram") ? "cuda" : "cpu";
```
alogfans: The VRAM location is hard-coded to the `cuda:` prefix. This breaks HIP and other GPU backends that use prefixes such as `rocm:` or `supa:`. Please use the common location parser to determine whether a buffer is host or device memory instead of matching a vendor-specific prefix.

mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
staryxchen: > We believe this change to the core allocation logic is unnecessary for the benchmark feature and should be reverted.

Agreed. The engine's allocateLocalMemory(MemoryOptions&) UNSPEC branch is back to its original RDMA > TCP preference, untouched.

mooncake-transfer-engine/benchmark/tent_backend.h
staryxchen: > The VRAM location is hard-coded to the `cuda:` prefix. This breaks HIP and other GPU backends that use prefixes such as `rocm:` or `supa:`. Please use the common location parser to determine whether a buffer is host or device memory instead of matching a vendor-specific prefix.

Good catch. getTargetBufferBase now uses LocationParser(loc).type() to classify each target buffer as host ("cpu") vs device (anything else: cuda, rocm, supa, …)

mooncake-transfer-engine/benchmark/tent_backend.cpp
```diff
@@ -161,48 +157,118 @@ int TENTBenchRunner::allocateBuffers() {
...
+                // register (existing behavior).
+                options.type = getTransportType(xport_type);
+            } else {
+                // Multi-transport mode (transports enabled via MC_TENT_CONF).
+                // Pick the allocate transport based on seg_type so the right
+                // transport-specific allocator runs (e.g. SHM sets shm_path,
+                // NVLink sets up GPU memory handle). registerLocalMemory
+                // below then resets options.type to UNSPEC so the buffer
+                // registers to ALL enabled transports, not just this one.
+                options.type = (seg_type == "vram") ? NVLINK : SHM;
```
alogfans: This also forces SHM/NVLINK when `--seg_type_mix` is not enabled, changing the existing single-segment fallback behavior. Recommend preserve the default value, i.e., remove this line.

staryxchen: Good catch — fixed in `25779c38`. The forced `options.type = (seg_type == "vram") ? NVLINK : SHM` is now gated on `!seg_type_mix_.empty()`, so it only applies when `--seg_type_mix` is explicitly set. The single-seg_type fallback (`--seg_type` without `--seg_type_mix`, no `--xport_type`) now leaves `options.type` as `UNSPEC`, restoring the original engine-default transport selection behavior. The matching `options.type = UNSPEC` reset before `registerLocalMemory` is gated the same way.

Verified:
- Single-seg_type fallback (`--seg_type=DRAM`, no `--xport_type`, no `--seg_type_mix`): `/metrics` shows `transport="tcp"` (engine default), no SHM forced. Original behavior restored.
- Mixed mode (`--seg_type_mix=dram,vram` + `MC_TENT_CONF` enabling shm+nvlink): both `transport="shm"` (140 GB) and `transport="nvlink"` (203 GB) still appear with non-zero counts. Multi-transport allocation intact.

alogfans (approved):
LGTM

PR #3128, opened 2026-07-27
[Store] Add configurable persist mode for bucket data

### LINKED ISSUE #3089: [Bug]: BucketStorageBackend never fsyncs bucket data before committing metadata (write-ordering guarantee is unreachable)
### Summary

`BucketStorageBackend::WriteBucket` is written to flush bucket data to stable storage *before* committing the bucket metadata, and a comment states this is a deliberate "write-ordering durability guarantee." That flush never runs: it lives inside a `dynamic_cast<UringFile*>` branch that the bucket **write** path can never enter, so on the actual (buffered `PosixFile`) path no `fsync`/`fdatasync` happens at all before metadata is committed.

The `offset_allocator` backend does perform this sync; only the bucket backend is missing it. This looks like an unintended regression rather than a design choice (see below).

### Where the guarantee is, and why it doesn't run

`WriteBucket` (`mooncake-store/src/storage_backend.cpp`, current `main`):

```cpp
// storage_backend.cpp:2546-2549
// Flush bucket data to stable storage before writing metadata.
// This prevents a crash from leaving valid metadata pointing at
// incomplete data (write-ordering durability guarantee).
auto sync_result = uring_file->datasync();   // inside `if (uring_file != nullptr)`
```

`uring_file` is `dynamic_cast<UringFile*>(file.get())`, where `file` was opened with `FileMode::Write` (`storage_backend.cpp:2476`). But `BucketStorageBackend::OpenFile` only returns a `UringFile` for reads:

```cpp
// storage_backend.cpp:3345 and :3357
if (file_storage_config_.use_uring && mode == FileMode::Read) { ... }
```

So for a `FileMode::Write` open the `dynamic_cast` is always null, the whole branch (aligned write **and** the `datasync()`) is skipped, and control falls to the buffered `PosixFile` path — which writes and then goes straight to `StoreBucketMetadata` with no sync in between.

### Confirmation

- With `MOONCAKE_OFFLOAD_USE_URING=true`, offloaded `.bucket` files are **not** 4096-aligned (e.g. `size=400044`), i.e. the O_DIRECT/aligned write branch that contains the sync did not execute.
- `strace -e fdatasync` on an offload run against `main` shows **no** fdatasync on the `.bucket` files.

### Why this looks like a regression, not a decision

The sync was introduced reachable in #1500, when the file-open helper returned a `UringFile` for both read and write. #1562 ("Optimize uring file support for SSD offloading") then implemented "use O_DIRECT for reads only" **asymmetrically**:

- `StorageBackend::create_file` (file-per-key) kept `if (use_uring_)` and passed `use_direct_io = (mode == Read)`, so writes still get a (buffered) `UringFile`.
- `BucketStorageBackend::OpenFile` instead added `&& mode == FileMode::Read` to the return condition, so bucket writes get a `PosixFile` — which orphaned the `datasync()` in `WriteBucket`.

Same intent, two implementations; the bucket one dropped the sync as a side effect. The `offset_allocator` backend still syncs its data file (`data_file_->datasync()` at `storage_backend.cpp:3484` and `:5039`), so the durability intent is consistent across the codebase except here.

### Impact

If the process/host crashes between the bucket data write and the metadata commit reaching disk, recovery can find committed metadata pointing at bucket data that never reached stable storage — the exact hazard the comment describes. This is the KV-cache offload path, so the failure mode is silent: a later read returns short/zero-filled data rather than an error.

### Proposed fix

Rather than a per-bucket `fdatasync` (which restores correctness but is costly), I'd suggest **amortizing one flush barrier across an offload batch**:

1. write each bucket's data file and kick off writeback with `sync_file_range(fd, .., SYNC_FILE_RANGE_WRITE)` (cheap, non-blocking);
2. issue a single flush barrier for the batch;
3. then commit the metadata for those buckets.

This keeps the ordering guarantee (all bucket data is on stable storage before any of the corresponding metadata is committed) while paying the device flush once per batch instead of once per bucket.

The reason this matters: the cost here is the device flush barrier, and it is a fixed per-call cost, not proportional to data. Measured on a single NVMe SSD (Samsung 990 PRO, ext4, barrier on, write-back cache), for a 2 MB write:

| operation | time |
|---|---|
| write to page cache only | 0.13 ms |
| `sync_file_range(WRITE)` (start writeback, no barrier) | 0.15 ms |
| `sync_file_range(WAIT)` (writeback done, no device flush) | 0.52 ms |
| `fdatasync` (writeback + device flush barrier) | 5.13 ms |

The ~4.6 ms is the flush barrier and stays constant at 16 MB too. Amortizing it across a batch is 6–12× cheaper than syncing per bucket:

| batch | per-bucket fdatasync | writeback + one barrier |
|---|---|---|
| 16 × 2 MB | 5.98 ms/bucket | 0.93 ms/bucket (6.4×) |
| 64 × 2 MB | 5.31 ms/bucket | 0.45 ms/bucket (11.8×) |

Note that `sync_file_range` / `O_DIRECT` on their own do **not** provide power-loss durability (data can sit in the drive's volatile cache) — a flush barrier is still required; this approach just pays it once per batch.

I'm happy to send a PR along these lines if this sounds right.

*(Found while working on the store offload path; environment: local build, Ubuntu 24.04, gcc 13.3, single NVMe SSD.)*

Daejun:
## Description

Fixes #3089.

`BucketStorageBackend::WriteBucket` carries a comment saying it flushes bucket data to stable storage before committing the bucket metadata, and calls `datasync()` to do it. That call never runs: it sits inside a `dynamic_cast<UringFile*>` branch, but `OpenFile()` only returns a `UringFile` for `FileMode::Read`, so bucket writes always take the buffered `PosixFile` path. Both `.bucket` and `.meta` are left dirty and which one reaches the device first is decided by kernel writeback, so a crash can leave a committed `.meta` pointing at bucket data that was never persisted. Recovery walks `.meta` files, so those keys come back and read as corrupt.

The first version of this PR made the flush unconditional. Following review (@LujhCoconut, @he-yufeng in the thread below), it does not: the cost is real and what it buys depends on the deployment, so the level is now an explicit setting.

**`MOONCAKE_OFFLOAD_BUCKET_PERSIST_MODE` on `BucketBackendConfig`**, applying to the bucket **data** file:

| Level | After the data write | Per 62.5 MiB bucket (p50) | Offload throughput (mean) | What it changes |
|---|---|---|---|---|
| `disabled` (default) | nothing | 10.55 ms | 5693 MiB/s | nothing — today's behavior |
| `relaxed` | `sync_file_range(WAIT_BEFORE\|WRITE\|WAIT_AFTER)` | 20.15 ms | 3006 MiB/s | the data is written back and waited for before `.meta` is written, instead of both files racing on writeback scheduling |
| `strict` | `fdatasync()` | 25.56 ms | 2398 MiB/s | the same, and makes the bucket data itself durable, subject to the device honoring cache flushes |

Two limits are stated explicitly in the docs and in the header, because they bound what any level can promise:

- **`relaxed` buys ordering, not durability.** `sync_file_range()` does not write file metadata, and its man page is explicit that for anything other than overwrites of already-instantiated blocks — which a freshly created bucket file is not — the data is not guaranteed to be there after a crash. The ordering property then rests on the filesystem committing metadata in order (ext4, xfs). It also adds nothing for a process crash: a dying process leaves the page cache intact.
- **No level makes an offloaded object durable, because `StoreBucketMetadata()` does not flush `.meta`.** A crash in the window right after an offload loses that bucket at every level — recovery finds no `.meta` and drops the orphaned data file. The setting removes the *corrupt* outcome, not the *lost* one. That is what #3089 reports; making offload a durable operation is a larger change (it needs `.meta` and the directory entry too) and is not attempted here.

`disabled` is the default, so no existing deployment sees its offload throughput move. The mode is deliberately separate from `OffsetPersistMode`: a node may want different levels on the two backends, and changing one must not silently move the other's durability behavior.

**Also in this change:**
- The flush moves out of the dead `UringFile`-only branch into `PersistBucketData()`, so it runs on the path bucket writes actually take. A failed flush aborts the offload and drops the just-written data file via `CleanupOrphanedBucket()`; the adjacent metadata-write-failure path is switched to the same helper so both abort paths clean up identically.
- `StorageFile::writeback_wait()` for the `relaxed` level, implemented once on the file descriptor since `sync_file_range()` is the same operation for every implementation. `WAIT_BEFORE` is included because without it the kernel runs the writeback in `WB_SYNC_NONE` mode, which skips a folio already under writeback rather than waiting for it. It costs nothing measurable here, where `WriteBucket` is the only writer of a freshly truncated file.
- `storage_backend_bench --bucket_persist_mode`, so the table above can be reproduced with the tool this repo already ships for this backend.
- One trailing newline at the end of `posix_file.cpp`, which `end-of-file-fixer` requires once the file is touched.

Not in this change: the recovery-side check that `BucketMetadata::data_size` matches the actual `.bucket` size. It is worth doing regardless of the level chosen, but it belongs to `Init()` rather than the write path, so it will be a separate PR.

## Module

## Type of Change

- [x] Bug fix
- [x] New feature

## How Has This Been Tested?

Five new cases in `storage_backend_test`, each checked by making the corresponding mutation to the production code and confirming the case fails:

1. `BucketOffloadFailsWhenDataPersistFails` — at both `relaxed` and `strict`, an injected flush failure makes the offload fail, leaves the key uncommitted and no orphan `.bucket`/`.meta` behind; a retry without the injection succeeds and the value reads back unchanged.
2. `BucketDataFlushPrecedesMetadataCommit` — the flush is issued *before* the metadata commit. The outcome of a single offload cannot show this (both orders fail the same way and `CleanupOrphanedBucket` removes both files either way), so the test blocks the metadata write with a directory at the `.meta` path and checks that the one-shot fault was already consumed, which is only true if the flush ran first.
3. `BucketPersistModeIssuesTheAdvertisedSyncCall` — arming only one of the two sync kinds separates the levels, so a swapped or downgraded mapping (`strict` silently doing `relaxed`'s work) fails.
4. `BucketOffloadIgnoresPersistFailureWhenDisabled` — at the default the flush is never issued, so the injected failure is not reached and the offload succeeds. This pins the setting as additive.
5. `BucketBackendConfigParsesPersistMode` — the three values, case-insensitivity, the fallback to the default on an unrecognized value, and unset.

The fault is injected in `StorageFile`, not in the backend, so case 1 is a real regression test for the flush being on the write path: if `WriteBucket` does not issue a flush, the injection is never consumed and the offload wrongly succeeds. The injection is process-global, so the fixture clears it in `TearDown()` and the env-var case restores the previous value on scope exit.

**Test commands:**
```bash
cmake --build build --target storage_backend_test -j$(nproc)
ctest -R 'storage_backend_test|file_storage_test|file_storage_promotion_test|posix_file_test|file_util_test|offset_allocator_test|ssd_metrics_test|offload_on_evict_test|promotion_on_hit_test|nvme_kv_storage_backend_test|client_storage_backend_test'

# cost table above, per level
./build/mooncake-store/benchmarks/storage_backend_bench \
  --backend=bucket --test=offload --value_size=131072 --batch_size=500 \
  --num_operations=60 --warmup_operations=6 --verify=false \
  --storage_path=<directory on the target device> \
  --bucket_persist_mode=disabled   # then relaxed, then strict
```

`--storage_path` matters: its default is under `/tmp`, which is often a different device.

**Test results:**
- [x] Unit tests pass — `storage_backend_test` 95/95; the SSD-tier ctest set 11/11
- [x] Manual testing done — cost table measured on a Samsung 990 PRO NVMe, dedicated ext4 partition, buffered I/O, single-threaded, 5 repetitions of 60 offloads, device otherwise idle. Runs are separated by `sync`, since `disabled` leaves the whole run's data dirty and the kernel would otherwise write it back while the next level is being timed.

## Checklist

- [x] I have formatted my code using `./scripts/code_format.sh` (clang-format 20.1.8)
- [x] I have updated the documentation (if applicable) — `docs/source/deployment/ssd/ssd-offload.md`
- [ ] For changes >500 LOC: N/A (~250 LOC excluding tests and docs)

## AI Assistance Disclosure

Claude Code investigated the code paths, ran the measurements, and drafted the change, tests and documentation. The human submitter reviewed every changed line and can defend the change end to end.

### DIFF
--- a/docs/source/deployment/ssd/ssd-offload.md
+++ b/docs/source/deployment/ssd/ssd-offload.md
@@ -161,6 +161,47 @@ Applies when `MOONCAKE_OFFLOAD_STORAGE_BACKEND_DESCRIPTOR=bucket_storage_backend
 | `MOONCAKE_OFFLOAD_BUCKET_EVICTION_POLICY` | `fifo` | Eviction policy: `none` / `fifo` / `lru` |
 | `MOONCAKE_OFFLOAD_BUCKET_MAX_PHYSICAL_BYTES` | `0` (disabled) | Hard cap on **real on-disk** bytes (`du`-equivalent) under this backend's `ssd_offload_path`. `0` disables it. Its scope depends on the deployment — see below. |
 | `MOONCAKE_OFFLOAD_BUCKET_DISK_SCAN_CACHE_MS` | `500` | How long the directory-scan result is cached before re-scanning, to bound the cost of the physical-usage check. `<= 0` scans on every check. |
+| `MOONCAKE_OFFLOAD_BUCKET_PERSIST_MODE` | `disabled` | How bucket data is flushed before its metadata is committed: `disabled` / `relaxed` / `strict`. Costs offload throughput; see [Bucket data durability](#bucket-data-durability) for what each level does and does not guarantee |
+
+(bucket-data-durability)=
+#### Bucket data durability
+
+A bucket is written as two files: `<id>.bucket` holds the data and `<id>.meta` holds the metadata that restart recovery walks. Neither is flushed by default, so which of the two reaches the device first is decided by kernel writeback. If a crash lands after `.meta` is on the device but while part of `<id>.bucket` is still in the page cache, recovery finds metadata pointing at data that was never persisted, and those keys read back as corrupt.
+
+`MOONCAKE_OFFLOAD_BUCKET_PERSIST_MODE` chooses what the offload path does about that window. It applies to the **data** file only:
+
+| Value | After the data write | What it changes |
+|---|---|---|
+| `disabled` (default) | nothing | Nothing. Existing behavior |
+| `relaxed` | `sync_file_range(WAIT_BEFORE\|WRITE\|WAIT_AFTER)` on the data file | The bucket data is written back and waited for before `.meta` is written, instead of both files racing on writeback scheduling. On a filesystem whose metadata journal commits in order (ext4, xfs), that is enough to stop `.meta` from becoming durable ahead of the data |
+| `strict` | `fdatasync()` on the data file | The same, and additionally makes the bucket data itself durable: `fdatasync()` writes the file metadata needed to read the data back and flushes the device cache |
+
+Two limits are worth being explicit about, because they bound what any level can promise:
+
+- **`sync_file_range()` is not a data-integrity operation.** It does not write out file metadata; `sync_file_range(2)` warns that unless the application is strictly overwriting already-instantiated blocks — which this path is not, since each bucket is a freshly created file — there is no guarantee the data is available after a crash. `relaxed` buys ordering, not durability. It also adds nothing for a process crash: a dying process leaves the page cache intact, and the kernel still writes both files out.
+- **No level makes an offloaded object durable, because `.meta` is never flushed.** After `strict` returns, the bucket data is on the device but `.meta` is still only in the page cache. A crash in that window loses `.meta`, so restart recovery never sees the bucket and deletes the orphaned `.bucket` — the object is lost, exactly as it would be at `disabled`. What the setting removes is the *corrupt* outcome (metadata durable, data not), not the *lost* one. `strict`'s power-loss guarantee also assumes the device honors cache flushes and the filesystem is not mounted with barriers disabled, and it does not cover the directory entry.
+
+Both levels above `disabled` cost write throughput on the offload path. The cost is dominated by writing the bucket's own data out, so it grows with bucket size; `strict` adds a further roughly constant amount for what `fdatasync()` does beyond that (filesystem journal commit and device cache flush). Measured at the 62.5 MiB bucket that `MOONCAKE_OFFLOAD_BUCKET_KEYS_LIMIT=500` produces at 128 KiB values, on one Samsung 990 PRO NVMe with a dedicated ext4 partition, single-threaded, 5 repetitions of 60 offloads:
+
+| Value | Per bucket (p50) | Offload throughput (mean) |
+|---|---|---|
+| `disabled` | 10.55 ms | 5693 MiB/s |
+| `relaxed` | 20.15 ms | 3006 MiB/s |
+| `strict` | 25.56 ms | 2398 MiB/s |
+
+```bash
+./build/mooncake-store/benchmarks/storage_backend_bench \
+  --backend=bucket --test=offload --value_size=131072 --batch_size=500 \
+  --num_operations=60 --warmup_operations=6 --verify=false \
+  --storage_path=<directory on the target device> \
+  --bucket_persist_mode=disabled   # then relaxed, then strict
+```
+
+`--storage_path` matters: its default is under `/tmp`, which is often a different device, or a tmpfs. The two columns are not two views of the same statistic — the per-bucket figure is the p50 latency and the throughput is derived from mean latency — so they do not divide into each other. The ratios travel better than the absolute numbers; measure on the target device before choosing.
+
+`disabled` does not avoid the write, it defers it: the bucket data stays dirty in the page cache until kernel writeback picks it up, so its cost lands on whatever else is using the device at that moment rather than on the offload call. Its 5693 MiB/s is the rate of filling the page cache, not a device rate. For the same reason, benchmarking these levels against each other requires a `sync` between runs, or `disabled`'s deferred writeback is charged to whichever level runs next.
+
+This setting is independent of `MOONCAKE_OFFSET_PERSIST_MODE`, which controls the offset-allocator backend.
 
 ### File-per-key backend settings
 
@@ -306,6 +347,7 @@ mooncake_client \
 
 - `MOONCAKE_OFFLOAD_FILE_STORAGE_PATH` must be an absolute path to an existing, writable directory. Symbolic links and paths containing `..` are rejected.
 - On real client restart, `bucket_storage_backend` and `file_per_key_storage_backend` scan existing SSD metadata and report it to the master, so previously offloaded objects remain accessible. `offset_allocator_storage_backend` does not support restart recovery.
+- What `bucket_storage_backend` recovers after an unclean shutdown depends on `MOONCAKE_OFFLOAD_BUCKET_PERSIST_MODE`; see [Bucket data durability](#bucket-data-durability).
 - Eviction only notifies the master and deletes local files; objects replicated on other nodes are unaffected.
 - Each machine requires its own real client process. In multi-node deployments, ensure `--host` and `--port` are correctly set so nodes can reach each other.
 
--- a/mooncake-store/benchmarks/storage_backend_bench.cpp
+++ b/mooncake-store/benchmarks/storage_backend_bench.cpp
@@ -136,6 +136,9 @@ DEFINE_string(cache_mode, "buffered",
               "Cache mode: buffered (page cache), direct (O_DIRECT), "
               "fadvise_dontneed (evict after use), drop_cache_external (print "
               "instructions)");
+DEFINE_string(bucket_persist_mode, "disabled",
+              "Bucket backend only: flush level applied to bucket data before "
+              "its metadata is committed (disabled/relaxed/strict)");
 DEFINE_bool(flush_between_ops, false,
             "Call fdatasync after each write (NOT IMPLEMENTED - requires "
             "backend Flush API)");
@@ -200,6 +203,15 @@ enum class CacheMode {
     FADVISE_DONTNEED,    // Buffered + evict after use
     DROP_CACHE_EXTERNAL  // Print instructions for external cache drop
 };
+
+mooncake::BucketPersistMode StringToBucketPersistMode(const std::string& str) {
+    if (str == "disabled") return mooncake::BucketPersistMode::kDisabled;
+    if (str == "relaxed") return mooncake::BucketPersistMode::kRelaxed;
+    if (str == "strict") return mooncake::BucketPersistMode::kStrict;
+    LOG(WARNING) << "Unknown bucket persist mode: " << str
+                 << ", using disabled";
+    return mooncake::BucketPersistMode::kDisabled;
+}
 
 CacheMode StringToCacheMode(const std::string& str) {
     if (str == "buffered") return CacheMode::BUFFERED;
@@ -819,6 +831,8 @@ std::shared_ptr<mooncake::StorageBackendInterface> CreateBackend(
             mooncake::BucketBackendConfig bucket_config;
             bucket_config.bucket_size_limit = 256 * MB;
             bucket_config.bucket_keys_limit = 500;
+            bucket_config.persist_mode =
+                StringToBucketPersistMode(FLAGS_bucket_persist_mode);
             return std::make_shared<mooncake::BucketStorageBackend>(
                 config, bucket_config);
         }
--- a/mooncake-store/include/file_interface.h
+++ b/mooncake-store/include/file_interface.h
@@ -67,6 +67,48 @@ class StorageFile {
      */
     virtual tl::expected<void, ErrorCode> datasync() { return {}; }
 
+    // ---- Test-only fault injection ----
+    // Which sync call a test wants to fail. kAny matches both.
+    enum class SyncKind { kNone = 0, kWritebackWait, kDatasync, kAny };
+
+    // Makes the next matching sync call on any StorageFile report failure
+    // without issuing the syscall, so a test can verify that a caller which is
+    // supposed to flush actually issues that specific call and propagates the
+    // error. One-shot: consumed by the first matching call. Process-global
+    // because storage backends create their StorageFile instances internally,
+    // so a test has no handle to set it on. Pass kNone to disarm.
+    static void SetSyncFailureForTest(SyncKind kind) {
+        sync_failure_for_test_.store(static_cast<int>(kind),
+                                     std::memory_order_relaxed);
+    }
+
+    // Whether an injected failure is still waiting to be consumed. A test uses
+    // this to tell "the call was never issued" from "the call was issued and
+    // failed", which is how the ordering of two flushes can be observed.
+    static bool IsSyncFailureArmedForTest() {
+        return sync_failure_for_test_.load(std::memory_order_relaxed) !=
+               static_cast<int>(SyncKind::kNone);
+    }
+
+    /**
+     * @brief Writes this file's dirty pages back and waits for them, without
+     * flushing file metadata or the device's volatile write cache.
+     *
+     * sync_file_range(WAIT_BEFORE|WRITE|WAIT_AFTER) over the whole file.
+     * Cheaper than datasync() because it neither writes out the inode nor
+     * issues a device cache flush -- which is also why it is not a
+     * data-integrity operation: after it returns, the data pages have been
+     * written back, but for a newly created file the metadata describing where
+     * they live need not be on the device yet. Use it to order this file's
+     * data against a later write to another file, not to make it durable; use
+     * datasync() for that. A file accessed only through O_DIRECT has no
+     * page-cache pages to write back, so this does nothing for it.
+     *
+     * Operates on the file descriptor, so the base implementation is correct
+     * for every StorageFile.
+     */
+    virtual tl::expected<void, ErrorCode> writeback_wait();
+
     // Prevent destructor from unlinking the arena file on write failure.
     void SetDeleteOnWriteFail(bool v) { delete_on_write_fail_ = v; }
 
@@ -155,6 +197,21 @@ class StorageFile {
     ErrorCode get_error_code() { return error_code_; }
 
    protected:
+    // Returns and clears the injected sync failure if it matches `kind`. Call
+    // from the sync implementations before doing any real work.
+    static bool ConsumeSyncFailureForTest(SyncKind kind) {
+        int armed = sync_failure_for_test_.load(std::memory_order_relaxed);
+        if (armed != static_cast<int>(SyncKind::kAny) &&
+            armed != static_cast<int>(kind)) {
+            return false;
+        }
+        return sync_failure_for_test_.compare_exchange_strong(
+            armed, static_cast<int>(SyncKind::kNone),
+            std::memory_order_relaxed);
+    }
+
+    static std::atomic<int> sync_failure_for_test_;
+
     bool delete_on_write_fail_ = true;
     std::string filename_;
     int fd_;
--- a/mooncake-store/include/storage_backend.h
+++ b/mooncake-store/include/storage_backend.h
@@ -1148,6 +1148,11 @@ class BucketStorageBackend : public StorageBackendInterface {
     std::unordered_map<std::string, int64_t> GUARDED_BY(offloading_mutex_)
         ungrouped_offloading_objects_;
 
+    // Test-only fault injection: when set, the next WriteBucket data datasync()
+    // is treated as failed. Consumed (reset to false) on read. See
+    // SetDatasyncFailureForTest().
+    std::atomic<bool> test_datasync_failure_{false};
+
     // File handle cache for UringFile to avoid repeated open/close overhead
     mutable Mutex file_cache_mutex_;
     mutable std::unordered_map<std::string, std::shared_ptr<StorageFile>>
@@ -1159,6 +1164,15 @@ class BucketStorageBackend : public StorageBackendInterface {
 
     // Clear file cache (called on destruction or when needed)
     void ClearFileCache();
+
+   public:
+    // ---- Test accessors ----
+    // Force the next bucket-data datasync() in WriteBucket to be treated as
+    // failed, so tests can verify that a failed durability flush aborts the
+    // offload and does not commit metadata. One-shot (consumed on use).
+    void SetDatasyncFailureForTest() {
+        test_datasync_failure_.store(true, std::memory_order_relaxed);
+    }
 };
 
 class OffsetAllocatorStorageBackend : public StorageBackendInterface {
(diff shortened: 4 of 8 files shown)


he-yufeng (commented):
Verified the premise against the current tree: OpenFile only hands out UringFile for FileMode::Read, so every bucket write lands on the buffered PosixFile path, and the old dynamic_cast branch made the durability comment a lie for exactly the writes it was meant to protect. Both PosixFile and UringFile override datasync() (file_interface.h:169/230), so lifting the call out of the branch enforces the ordering on both paths with no behavior change on the uring side.

Two things I like in the shape of this: the one-shot fault injector is consumed on read so a leaked flag cannot poison a later test, and the failure path now funnels into the same CleanupOrphanedBucket the duplicate path uses instead of a hand-rolled fs::remove, one less cleanup variant to keep honest.

One question before this lands: datasync on every WriteBucket means one fdatasync per bucket on the offload hot path. On slow SSDs that is usually still noise next to the data write itself, but if anyone has burst-offload latency numbers worth posting, it would settle the cost side rather than leaving it to intuition.

Comment only, not an approval: durability ordering on the data path should get a second maintainer's eyes before merge.

Daejun:
@he-yufeng Numbers for the cost question, at three bucket sizes including the one the offload path actually produces.

`FileStorage::OffloadObjects` groups keys through `GroupOffloadingKeysByBucket`, which fills a bucket up to `bucket_keys_limit` (default 500) or `bucket_size_limit` (default 256 MB) and defers the remainder to the next heartbeat, then calls `BatchOffload` once per bucket. `BatchOffload` takes a fresh `bucket_id` and calls `WriteBucket` once, so bucket size = `batch_size × value_size` and each timed op pays exactly one `fdatasync`. At 128 KB values a bucket written in production holds 500 keys ≈ 62 MB.

**Setup**

- Samsung 990 PRO 1TB NVMe, dedicated ext4 partition (`rw,relatime`), volatile write cache in write-back mode, `none` I/O scheduler. Ubuntu 24.04.4, kernel 7.0.0, gcc 13.3.0, `RelWithDebInfo`.
- `storage_backend_bench --backend=bucket --test=offload`: single-threaded, sequential `BatchOffload`, `--verify=false`, default `--cache_mode=buffered`, `--storage_path` on that NVMe partition.
- Baselines differ between runs. The 2 MB and 16 MB rows compare `main @ a1ca71f4` against that same commit plus this change alone. The 62 MB row compares `e13780da`, this PR's parent, against the PR head: my local `main` had fallen 18 commits behind and `storage_backend.cpp` changed in between. `storage_backend_bench.cpp` is unchanged across that range, so the flags are comparable.
- Between the 2/16 MB runs and the 62 MB run the branch was rebased onto `e13780da`. On the success path the two versions differ only by one relaxed atomic exchange for the test hook; the later commits add the failure-path `CleanupOrphanedBucket` and that hook.
- Each run writes 1–4 GB against a ~25 GB dirty limit (`dirty_ratio=20`, 125 GB RAM), so the baseline arm's data stays in page cache for the timed region.
- For the 62 MB run, `iostat -x -d 5` was recorded for the whole run; no third-party device activity appears in it.

**Results** (mean, `value_size=131072`)

| bucket | ops × reps | baseline | with `datasync()` | delta |
|---|---|---|---|---|
| 2 MB (`batch=16`) | 500 × 5 | 5860 MB/s, P50 0.32 ms | 367 MB/s, P50 5.46 ms | −93.7%, +5.14 ms |
| 16 MB (`batch=128`) | 256 × 3 | 4752 MB/s, P50 3.73 ms | 1535 MB/s, P50 10.22 ms | −67.7%, +6.49 ms |
| 62 MB (`batch=500`, production) | 60 × 3 | 5511 MB/s, P50 10.74 ms | 2439 MB/s, P50 25.03 ms | −55.7%, +14.29 ms |

Per-run figures (MB/s / P50 ms):

```
2 MB bucket (500 ops, 5 reps)       16 MB bucket (256 ops, 3 reps)
rep  baseline        fixed          rep  baseline        fixed
1    5412.68 / 0.31  374.60 / 5.30  1    2970.79 / 5.87  1537.31 / 10.14
2    5954.87 / 0.32  372.10 / 5.34  2    5671.46 / 2.67  1488.80 / 10.64
3    5974.55 / 0.32  372.72 / 5.33  3    5613.67 / 2.66  1577.84 /  9.87
4    6058.64 / 0.31  368.74 / 5.34
5    5901.81 / 0.32  345.34 / 5.97

62 MB bucket (60 ops, 3 reps)
rep  baseline         fixed
1    5119.49 / 11.15  2434.19 / 24.97
2    5702.68 / 10.55  2442.54 / 25.02
3    5710.77 / 10.52  2440.79 / 25.11
```

Daejun:
@LujhCoconut You offered on #3089 to review a PR for this — this is it, and it takes the per-bucket `file->datasync()` approach you described there.

One measurement bears on the cost estimate in that comment. `GroupOffloadingKeysByBucket` fills a bucket to `bucket_keys_limit` (default 500) before `OffloadObjects` calls `BatchOffload` on it, so at 128 KB values a bucket written in production is ≈62 MB. At that size the sync adds **+14.29 ms** per bucket, not ~5 ms: the device barrier is around the +5.14 ms measured at a 2 MB bucket, and the remainder is writeback of the bucket's own data, which grows with the bucket. Numbers for 2 MB / 16 MB / 62 MB buckets are in the comment above.

So the barrier is roughly a third of the added cost at production bucket size, which is the part batch amortization could remove.

LujhCoconut:
@Daejun  Thanks for the fix. After going through the benchmark data posted here and thinking carefully about the actual failure model, I'm not sure per-bucket `fdatasync` earns its keep here.

Cost, by this PR's own numbers: +14.29 ms per production-size (~62 MB) bucket, P50 10.74 → 25.03 ms. Offload being on the async path makes the added write latency itself tolerable, but the flush barrier also interferes with latency-sensitive reads on the same SSD. Benefit: covers a window that only exists on host crash/power loss (a process crash keeps the page cache), affecting at most the single bucket mid-write at crash time. That reads as cost > benefit to me.

A cheaper option: at recovery, simply distrust the bucket with the highest `bucket_id`. This works because at most one heartbeat is in flight at a time and `OffloadObjects` writes buckets one by one, so at crash time only the newest bucket can be mid-write. Bucket IDs are monotonic and recovery re-seeds from `max_bucket_id`. Zero hot-path cost; worst case we discard one bucket of cache.

Caveat: it's a heuristic — the dirty-page window can span the last few seconds of completed buckets too. If a strict guarantee is wanted later, a CRC in `.meta` verified at recovery (like the offset_allocator record CRC) would do it with no hot-path I/O — though it's not free either: verification means reading every bucket back on restart, so recovery time becomes O(cache size), and it needs a `.meta` format change.

Happy to discuss further.

Daejun:
@LujhCoconut The cost objection is right, and decomposing it makes the case against `fdatasync` here sharper than the single number did.

Measured on the same host: one production-size bucket (500 keys × 128 KB = 62.5 MB), buffered, fresh file per bucket, one `pwritev` with an iovec per key — what `WriteBucket` does — timing the sync call separately from the write. p50 of 30 reps, device otherwise idle:

| sync level after the data write | sync cost | total per bucket | throughput |
|---|---|---|---|
| none — current behaviour | — | 9.26 ms | 6749 MB/s |
| `sync_file_range(WRITE)` | 1.62 ms | 11.33 ms | 5516 MB/s |
| `sync_file_range(WRITE\|WAIT_AFTER)` | 9.54 ms | 19.23 ms | 3250 MB/s |
| `fdatasync` — this PR | 14.64 ms | 24.21 ms | 2582 MB/s |

`pwritev` itself is 9.26–9.70 ms in all four, so the differences are the sync call alone.

So `fdatasync`'s 14.64 ms is 9.54 ms of pushing 62.5 MB to the device plus a 5.10 ms device barrier. The barrier is roughly constant — it computes to ~4.6 ms at the 2 MB bucket measured in #3089 — and this standalone figure agrees to within 2.5% with the +14.29 ms measured through `BatchOffload` above.

Two things follow.

**`WAIT_AFTER` is where the ordering window actually closes.** It still pays the 9.54 ms writeback, so it is 65% of `fdatasync`, but unlike `sync_file_range(WRITE)` it confirms the bucket data has left page cache before `.meta` is written at all — that is a guarantee, not a narrowed window. What it gives up is the device barrier: the data may still be in the drive's volatile cache, so it does not cover power loss.

**`sync_file_range(WRITE)` costs 1.62 ms, 11% of `fdatasync`.** What it buys is not durability — it is that the bucket data stops depending on when the kernel happens to flush it. Today `.bucket` and `.meta` are both left dirty, and which one reaches the device first is decided entirely by writeback-thread scheduling and `dirty_expire_centisecs` (30 s on this host). Kicking the data explicitly removes the data side of that dependency: we initiate its writeback ourselves, and only the metadata is left to the flusher. The residual case is the flusher completing the few-KB `.meta` before the in-flight 62.5 MB `.bucket` has drained, so this narrows the window rather than closing it.

On distrusting the highest `bucket_id` at recovery — I do not think the vulnerable set is the newest bucket. `BucketStorageBackend::Init` walks only `BUCKET_METADATA_FILE_SUFFIX` files, and the orphan pass after that loop deletes `BUCKET_DATA_FILE_SUFFIX` files with no matching `.meta`, so a bucket whose `.meta` never reached disk is invisible to recovery and harmless. The dangerous state is `.meta` durable with `.bucket` incomplete, which needs the flusher to have finished the small `.meta` while the large `.bucket` was still draining — that happens to buckets the flusher has already started on, which are the older ones. The newest bucket is the one most likely to have neither file persisted, i.e. the safe case. With a 30 s dirty window against a 10 s heartbeat the candidate set spans roughly three heartbeats of buckets, not one.

There may also be a cheaper recovery-side check than CRC. `BucketMetadata` already carries `data_size`, and `Init` verifies only that the `.bucket` file *exists* — the `fs::status` / `is_regular_file` check — not that its size matches. Comparing the two is one `stat` per bucket — O(bucket count) rather than O(cache size), and no `.meta` format change. It would not catch same-length corruption, which is what a CRC is for, but it does catch truncation.

My conclusion from all of this is that the sync level should be an explicit deployment choice rather than something this PR decides on everyone's behalf. The levels are not interchangeable — they buy different failure coverage at different write-throughput cost — and which one is right depends on whether the host has power-loss protection and on how much a warm cache after restart is worth against offload throughput. That belongs to whoever operates the node.

Concretely: a sync-level setting on `BucketBackendConfig` resolved from the environment, mirroring `OffsetAllocatorBackendConfig::persist_mode` / `OffsetPersistMode`, with four positions:

| level | action | cost per bucket | what it covers |
|---|---|---|---|
| `none` | today's behaviour | — | nothing |
| `writeback` | `sync_file_range(WRITE)` | 1.62 ms | data writeback stops waiting on flusher scheduling; narrows the window |
| `wait` | `sync_file_range(WRITE\|WAIT_AFTER)` | 9.54 ms | data confirmed out of page cache before `.meta` is written — closes the window against page-cache loss |
| `durable` | `fdatasync` | 14.64 ms | the same, and survives power loss |

Each step buys something nameable, so I would rather expose all four than pick for everyone. `writeback` and `wait` differ qualitatively, not just in price: the first makes the intended ordering deliberate, the second makes it guaranteed short of power loss. `none` as the default keeps the change purely additive, so no existing deployment sees its offload throughput move.

LujhCoconut:
@Daejun 

Thanks for the exceptionally detailed analysis — the writeback/barrier decomposition makes the cost structure much clearer, and you're right about my "distrust the newest bucket" suggestion: recovery keys off `.meta` files and the orphan pass handles the rest, so the newest bucket is the safe case, while the vulnerable set is the older buckets the flusher has already started on (roughly a dirty-window's worth, not one). I stand corrected.

I'm on board with making the sync level an explicit deployment choice, with `none` as the default so existing deployments see no behavior change. One small suggestion: do we need the `writeback` level? It costs 1.62 ms but buys no verifiable guarantee — it narrows the window without closing it, so there's nothing an operator can reason about or a test can assert. Dropping it leaves three levels with genuinely distinct semantics:

- `none` — no guarantee (today's behavior, default)
- `wait` — `sync_file_range(WRITE|WAIT_AFTER)`, survives page-cache loss
- `durable` — `fdatasync`, survives power loss

These map cleanly onto the `OffsetPersistMode` enum already defined for the offset allocator backend (`kDisabled` / `kRelaxed` / `kStrict`, mooncake-store/include/storage_backend.h) — possibly renamed to something backend-neutral like `PersistMode`. Reusing it would keep the config vocabulary consistent across backends and avoid introducing a new set of knobs.

The recovery-side size check (`data_size` vs the actual `.bucket` size) also seems worth doing regardless of which level is chosen — one stat per bucket, no format change.

Daejun:
@LujhCoconut Agreed on dropping `writeback`. The strongest postcondition a test could assert is that the host started flushing the dirty pages — not that any of them reached the device — and how much it narrows the window is not a fixed quantity — it moves with flusher scheduling, the dirty thresholds the host's memory size implies, and whatever else is queued on the device, so a figure measured here would not carry to anyone else's deployment.

On reusing `OffsetPersistMode` I have a different view. An operator may well want different levels on the two backends — a node running the offset allocator at `strict` while the bucket backend stays at `none`, or the reverse — and a single shared setting makes that impossible. It also means changing it for one backend silently moves the other's durability behaviour on an existing deployment. So: its own mode on `BucketBackendConfig`, read from its own variable, with three levels — `none` (default), `relaxed` = `sync_file_range(WRITE|WAIT_AFTER)`, `strict` = `fdatasync`.

On the recovery-side size check: agreed. It is independent of the level chosen and this PR does not touch recovery, so I will take it in a separate PR rather than widen this one.

Daejun:
@LujhCoconut @he-yufeng Rewritten to the three-level setting we converged on, and force-pushed. The title and description now describe the setting rather than the unconditional flush.

`writeback` is dropped, as agreed. The mode is separate from `OffsetPersistMode` for the reason in my earlier comment: a node may want different levels on the two backends, and changing one should not move the other.

Measured through the backend at the bucket size the offload path produces (500 keys x 128 KiB = 62.5 MiB), Samsung 990 PRO on a dedicated ext4 partition, single-threaded, 5 repetitions of 60 offloads, device otherwise idle:

| level | per bucket (p50) | offload throughput (mean) |
|---|---|---|
| `disabled` | 10.55 ms | 5693 MiB/s |
| `relaxed` | 20.15 ms | 3006 MiB/s |
| `strict` | 25.56 ms | 2398 MiB/s |

The two columns are not two views of one statistic — the latency is a p50 and the throughput is derived from mean latency — so they do not divide into each other. `--bucket_persist_mode` is part of this PR so the table can be reproduced; the full invocation is in the PR description, and `--storage_path` matters because its default is under `/tmp`, which is usually a different device.

**Two limits I want to be explicit about, because one of them bears on the judgment we made on 30 July.**

`relaxed` buys ordering, not durability. `sync_file_range()` does not write out file metadata, and its man page is explicit that unless the caller is strictly overwriting already-instantiated blocks — which a freshly created bucket file is not — the data is not guaranteed to be there after a crash. What it provides is that the bucket data is written back and waited for before `.meta` is written; whether `.meta` can still become durable ahead of the data then rests on the filesystem committing metadata in order (ext4, xfs). It also adds nothing against a process crash, since a dying process leaves the page cache intact — that was wrong in my earlier description. Relatedly the flags are `WAIT_BEFORE|WRITE|WAIT_AFTER`: without `WAIT_BEFORE` the kernel runs the writeback in `WB_SYNC_NONE` mode, which skips a folio already under writeback rather than waiting for it. It costs nothing measurable here, where `WriteBucket` is the only writer of a freshly truncated file.

**No level makes an offloaded object durable, because `StoreBucketMetadata()` does not flush `.meta`.** After `strict` returns, the bucket data is on the device and `.meta` is not; a crash in that window loses `.meta`, `Init()` never sees the bucket, and the orphan pass deletes the durable `.bucket`. The object is gone, exactly as it would be at `disabled`. So `strict` does not buy crash survival for the offloaded object, which I think is how we were both reading it on 30 July. What the setting removes is the corrupt outcome — durable metadata pointing at data that was not persisted — which is what #3089 reports, and I think that is still worth having as an opt-in. The docs and the header now say this rather than leaving it to be inferred.

If you would rather `strict` also mean the object survives, that needs `.meta` and the directory entry too — two more barriers on top of the numbers above, and probably an atomic `.meta` write. I would rather do that as a follow-up than widen this PR, but it is your call.

One thing I withdraw from the earlier cost discussion: I described `fdatasync`'s extra ~5 ms over `relaxed` as the device barrier. `iostat` across the run puts flush service time at 1.2–2.6 ms, so most of that residual is elsewhere, most likely the filesystem journal commit. I have not measured the split, so the PR no longer asserts one.

**On the tests**, since that is the part I would want a reviewer to check. The fault is injected in `StorageFile`, not in the backend: if `WriteBucket` does not issue a flush at all — the shape of the original bug — the injection is never consumed and the offload succeeds, failing the test. My first version injected inside the backend and was vacuous; removing the flush call left it green. The ordering itself also needed its own case, because both orders fail a single offload identically and `CleanupOrphanedBucket` removes both files either way; the test now blocks the metadata write with a directory at the `.meta` path and checks the one-shot fault was already consumed, which holds only if the flush ran first. A third case arms one sync kind at a time, so a swapped or downgraded level mapping fails. Each case was checked by making the mutation it targets.

Two measurement notes for anyone reproducing this: the levels have to be separated by a `sync`, since `disabled` leaves the whole run's data dirty and the kernel writes it back while the next level is being timed — that inflated `relaxed` to 52.7 ms in my first attempt against 20.1 ms once isolated. And `disabled`'s 5693 MiB/s is the rate of filling the page cache, not a device rate: it does not avoid the write, it defers it onto whatever else is using the device at that moment.

The recovery-side `data_size` vs actual `.bucket` size check is not in here, as said — separate PR.

LujhCoconut:
This is now an exceptionally high-quality PR; reviewing it was a genuine pleasure. The documented latency and throughput measurements across the three modes will help users make an informed choice. I don't see any further blockers.
  
Since several PRs were just merged, there may be some merge conflicts to resolve on your end.

Thanks for your great work and patience.

PR #3129, opened 2026-07-27
[TENT] Validate sub-batch type in getTransferStatus

jfeng18:
## Summary

`getTransferStatus` is a public Transport API, but five transports dereferenced the result of `dynamic_cast<XxxSubBatch*>(batch)` without a null check: TCP, SHM, MNNVL, NVLink, and io_uring. SunriseLink's `getTransferStatus`, RDMA's `cancelTransferTask`, and (since #3280) GDS already validate — this PR applies the same guard to the remaining five for a uniform error surface.

The internal poll path is type-safe (`pollTaskStatus` indexes transport and sub-batch by the same `task.type`), so this is API hardening/consistency, not a reachable crash on current main.

Also adds a mismatch regression test (`TcpTransportSubBatchTest.RejectsForeignSubBatchInGetTransferStatus`, per review): hands an ShmSubBatch to TcpTransport (and the reverse) asserting InvalidArgument, plus a matching-type case. CPU-only, no install() required.

> Rebased on latest main: the GDS hunk from the original version was dropped — #3280 added an equivalent check upstream. `rdma_transport.cpp` remains out of scope (being reworked in #3016).

## Test plan

- [x] Regression test discriminates: reverting the guards makes it dereference nullptr
- [x] CI (all transports; tent-ci cuda-on/off)

### DIFF
--- a/mooncake-transfer-engine/tent/src/transport/gds/gds_transport.cpp
+++ b/mooncake-transfer-engine/tent/src/transport/gds/gds_transport.cpp
@@ -315,6 +315,9 @@ Status GdsTransport::submitTransferTasks(
 Status GdsTransport::getTransferStatus(SubBatchRef batch, int task_id,
                                        TransferStatus& status) {
     auto gds_batch = dynamic_cast<GdsSubBatch*>(batch);
+    if (!gds_batch) {
+        return Status::InvalidArgument("Invalid GDS sub-batch" LOC_MARK);
+    }
     unsigned num_tasks = gds_batch->io_param_ranges.size();
     if (task_id < 0 || task_id >= (int)num_tasks)
         return Status::InvalidArgument("Invalid task ID");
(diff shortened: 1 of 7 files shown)


he-yufeng (commented):
Spot-checked the claim against the current tree and it holds: TCP has three dynamic_cast sites, and freeSubBatch plus submitTransferTasks already bail out with InvalidArgument while getTransferStatus dereferenced the result bare. The same asymmetric guard pattern is what this PR removes in all six transports, and the wording matches the precedent set in SunriseLink's getTransferStatus and RDMA's cancelTransferTask, so the error surface stays uniform instead of inventing a new one.

One suggestion, non-blocking: a mismatch regression is cheap to write here. Allocate a ShmSubBatch, hand it to TcpTransport::getTransferStatus, and assert InvalidArgument instead of UB. Without it, a future transport added by copy-paste can quietly reintroduce the same bare cast, since the compiler has no way to flag this class of mistake.

Comment only, not an approval.

jfeng18:
Thanks for the spot-check and the suggestion @he-yufeng — added in 876cadcd: `TcpTransportSubBatchTest.RejectsForeignSubBatchInGetTransferStatus` hands an ShmSubBatch to TcpTransport (and the reverse) asserting InvalidArgument, plus a matching-type case to confirm the guard discriminates rather than rejects everything. TCP/SHM only, so it runs on CPU-only CI; neither allocateSubBatch nor the type check needs install(), so default-constructed transports suffice.

jfeng18:
@staryxchen @dtcccc @00fish0 — re-verified this against latest main before pinging: the five guards are still the only dynamic_cast null-checks in tent's getTransferStatus implementations (GDS got its own via #3280, RDMA stays excluded via #3016), the mismatch regression test still passes on all tent-ci suites, and the branch rebases cleanly with CI 27/27. Would appreciate a look when one of you has a cycle. Thanks!

PR #3133, opened 2026-07-27
[wheel] preserve grouped writes for structured objects

zxpdemonio:
### Motivation

Structured object transfer may fan out one logical object into multiple physical Mooncake Store keys, including metadata, payload chunks, multi-buffer chunks, and the final manifest. When callers pass a `ReplicateConfig` with `group_ids`, Mooncake Store batch APIs require the number of group IDs to match the physical key count.

Before this change, structured object writes did not consistently preserve group semantics across all internal write paths, especially multi-buffer writes and manifest updates during append. This could make grouped placement/replication semantics incomplete for structured rollout-data transfer.

### Description

This PR adds grouped write support for structured objects.

- Expands one logical `config.group_ids` value across the physical keys written by structured object batch paths.
- Preserves `config` through multi-buffer PUT paths.
- Ensures structured object manifest writes, including append-generated manifests, are written with the same group semantics.
- Allows manifest updates to keep cleanup-only object IDs in `buffer_object_ids`, so append/merge manifests remain valid even when old stage keys are only cleanup targets.
- Supports real `mooncake.store.ReplicateConfig`, including the pybind object case where `copy.copy()` is not available.
- Rejects ambiguous structured object configs with multiple logical group IDs, since one structured object should map to one logical group.

This is a focused Python wheel change for structured object transfer semantics.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
git diff --check
python3 -m py_compile mooncake-wheel/mooncake/structured_object_store.py mooncake-wheel/tests/test_structured_object_store.py

cd mooncake-wheel
PYTHONPATH=. python3 -m pytest tests/test_structured_object_store.py -q -k "group or replicate_config or appends_stage_fields_without_rewriting_existing or overwrite_replaces_encoded_metadata or cleanup_removes_all_stage_objects or typed_ragged_uses_multi_buffer_put"
PYTHONPATH=. python3 -m pytest tests/test_structured_object_store.py -q -k "not real_store"
```

**Test results:**
- [x] Unit tests pass
- [x] Manual testing done

Result:
```text
12 passed, 101 deselected in 2.14s
88 passed, 15 skipped, 10 deselected, 14 warnings in 12.54s
```

Additional checks:
```text
git diff --check: passed
py_compile: passed
```

Note:
```text
The full test_structured_object_store.py run includes real-store setup tests that can hang in this test environment. The focused regression suite and the non-real-store structured object suite passed. The skipped tests are caused by the local torch/CUDA import environment.
```

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex was used to implement the focused structured object group semantics patch, run targeted validation, perform self-review, and prepare this PR description.

### DIFF
--- a/mooncake-wheel/mooncake/structured_object_store.py
+++ b/mooncake-wheel/mooncake/structured_object_store.py
@@ -356,7 +356,7 @@ def import_dataproto_ref(handle: Mapping[str, Any]) -> MooncakeDataProtoRef:
     if not is_dataproto_ref_handle(handle):
         raise ValueError("not a Mooncake DataProto ref handle")
     version = int(handle.get("version", -1))
-    if version != DATAPROTO_REF_HANDLE_VERSION:
+    if version not in (DATAPROTO_REF_HANDLE_VERSION, 2):
         raise ValueError(
             f"unsupported DataProto ref handle version: {handle.get('version')!r}"
         )
@@ -400,8 +400,9 @@ def import_dataproto_ref(handle: Mapping[str, Any]) -> MooncakeDataProtoRef:
             stage=stage, member=member, section=section
         )
     storage_group_id = handle.get(STORAGE_GROUP_ID)
-    if storage_group_id is not None and (
-        not isinstance(storage_group_id, str) or not storage_group_id
+    if (version == 2 and storage_group_id is None) or (
+        storage_group_id is not None
+        and (not isinstance(storage_group_id, str) or not storage_group_id)
     ):
         raise ValueError("DataProto ref handle has invalid storage_group_id")
     return MooncakeDataProtoRef(
--- a/mooncake-wheel/mooncake/structured_object_store.py
+++ b/mooncake-wheel/mooncake/structured_object_store.py
@@ -8,6 +8,7 @@ import uuid
 from concurrent.futures import Future, ThreadPoolExecutor, as_completed
 from contextlib import contextmanager
 from dataclasses import dataclass, field
+from types import SimpleNamespace
 from typing import Any, Callable, Iterator, Literal, Mapping, Optional, Protocol, Sequence
 
 import numpy as np
@@ -128,6 +129,7 @@ class MooncakeDataProtoRef:
     partition: str = "default"
     global_indexes: list[int] | None = None
     encoded_non_tensor: dict[str, Any] = field(default_factory=dict)
+    _storage_group_id: str | None = field(default=None, repr=False)
 
 
 @dataclass(frozen=True)
@@ -306,7 +308,8 @@ class _DataProtoRowSelection:
 
 
 DATAPROTO_REF_HANDLE_TYPE = "mooncake_dataproto_ref"
-DATAPROTO_REF_HANDLE_VERSION = 1
+DATAPROTO_REF_HANDLE_VERSION = 2
+STORAGE_GROUP_ID = "storage_group_id"
 DataProtoRefLike = MooncakeDataProtoRef | Mapping[str, Any]
 
 
@@ -320,7 +323,7 @@ def export_dataproto_ref(ref: MooncakeDataProtoRef) -> dict[str, Any]:
         raise TypeError(f"expected MooncakeDataProtoRef, got {type(ref).__name__}")
     handle = {
         "type": DATAPROTO_REF_HANDLE_TYPE,
-        "version": DATAPROTO_REF_HANDLE_VERSION,
+        "version": DATAPROTO_REF_HANDLE_VERSION if ref._storage_group_id is not None else 1,
         "kind": "bundle_stages",
         "batch_size": int(ref.batch_size),
         "namespace": ref.namespace,
@@ -341,6 +344,8 @@ def export_dataproto_ref(ref: MooncakeDataProtoRef) -> dict[str, Any]:
         "meta_info": _json_safe_value(ref.meta_info),
         "encoded_non_tensor": _json_safe_value(ref.encoded_non_tensor),
     }
+    if ref._storage_group_id is not None:
+        handle[STORAGE_GROUP_ID] = ref._storage_group_id
     json.dumps(handle, ensure_ascii=False)
     return handle
 
@@ -349,7 +354,8 @@ def import_dataproto_ref(handle: Mapping[str, Any]) -> MooncakeDataProtoRef:
     """Import a JSON-safe DataProto transport handle into a lazy ref."""
     if not is_dataproto_ref_handle(handle):
         raise ValueError("not a Mooncake DataProto ref handle")
-    if int(handle.get("version", -1)) != DATAPROTO_REF_HANDLE_VERSION:
+    version = int(handle.get("version", -1))
+    if version not in (1, DATAPROTO_REF_HANDLE_VERSION):
         raise ValueError(
             f"unsupported DataProto ref handle version: {handle.get('version')!r}"
         )
@@ -392,6 +398,11 @@ def import_dataproto_ref(handle: Mapping[str, Any]) -> MooncakeDataProtoRef:
         field_index[name] = StructuredFieldLocation(
             stage=stage, member=member, section=section
         )
+    storage_group_id = handle.get(STORAGE_GROUP_ID) if version == 2 else None
+    if version == 2 and (
+        not isinstance(storage_group_id, str) or not storage_group_id
+    ):
+        raise ValueError("DataProto ref handle requires storage_group_id")
     return MooncakeDataProtoRef(
         batch_size=int(handle["batch_size"]),
         stage_refs=stage_refs,
@@ -403,6 +414,7 @@ def import_dataproto_ref(handle: Mapping[str, Any]) -> MooncakeDataProtoRef:
         encoded_non_tensor=dict(
             _require_mapping(handle.get("encoded_non_tensor", {}), "encoded_non_tensor")
         ),
+        _storage_group_id=storage_group_id,
     )
 
 
@@ -479,6 +491,35 @@ class MooncakeBundleTransfer:
             default_chunk_bytes=self.default_chunk_bytes,
         )
         self._structured_store = _StructuredObjectLayer(self._bundle_store)
+
+    def _new_storage_group_id(self, config: Any) -> str:
+        raw_group_ids = getattr(config, "group_ids", None)
+        if raw_group_ids is None:
+            group_id = None
+        elif isinstance(raw_group_ids, str):
+            group_id = raw_group_ids
+        else:
+            try:
+                group_ids = list(raw_group_ids)
+            except TypeError as error:
+                raise ValueError(
+                    "structured object store config.group_ids must contain strings"
+                ) from error
+            if len(group_ids) > 1:
+                raise ValueError(
+                    "structured object store config.group_ids must contain exactly one "
+                    "logical group id; it is expanded across internal Mooncake keys"
+                )
+            group_id = group_ids[0] if group_ids else None
+        if group_id is None:
+            return f"structured-{uuid.uuid4().hex}"
+        if not isinstance(group_id, str):
+            raise ValueError(
+                "structured object store config.group_ids must contain strings"
+            )
+        if not group_id:
+            return f"structured-{uuid.uuid4().hex}"
+        return group_id
 
     def put_bundle(
         self,
@@ -1599,7 +1640,9 @@ class MooncakeBundleTransfer:
                     partition=ref.partition,
                     global_indexes=ref.global_indexes,
                     encoded_non_tensor=dict(ref.encoded_non_tensor),
+                    _storage_group_id=ref._storage_group_id,
                 )
+            group_id = self._new_storage_group_id(config)
             return MooncakeDataProtoRef(
                 batch_size=batch_size,
                 stage_refs={},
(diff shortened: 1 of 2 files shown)


zxpdemonio (commented):
## Review summary

Focused and useful change: structured-object PUTs now keep a single logical `config.group_ids` consistent with Mooncake Store’s physical-key batch contract, including multi-buffer paths and append-generated manifests. The pybind `ReplicateConfig` copy fallback and the “exactly one logical group id” rejection are the right product rules.

I re-ran the focused group/cleanup tests: **7 passed**. No functional blockers found.

> Note: posted as `COMMENT` because GitHub disallows `REQUEST_CHANGES` on own PRs (`zxpdemonio`).

### Should fix / clarify
1. **Two concerns in one PR** — commit 2 (`allow cleanup-only object ids in bundle manifests`) is required for append+groups (validation), but it is a separate semantic change to `buffer_object_ids`. Please call that out in the PR description so reviewers do not miss it.
2. **`_copy_write_config` via `dir()`/`setattr` is brittle** — works for current `ReplicateConfig` (covered by test), but can break on unexpected pybind properties. Prefer copying a known field allowlist (`replica_num`, `with_soft_pin`, `group_ids`, …) or document the supported config surface.
3. **Expansion granularity differs by path** — contiguous chunk `batch` expands `group_ids` to `len(keys)` in one `batch_put_from`; multi-buffer `batch_put_buffer_groups_from` still submits **one key per call** (so `group_ids` length stays 1). Both satisfy the store API, but please note this in a comment so future “must be expanded to full physical key count in one RPC” assumptions are not wrong.

### Nits
4. Consider rejecting empty `group_ids` with a more specific message (`must contain exactly one` already covers `[]`, but an explicit empty-list message would help).
5. CI on this PR currently shows most jobs as `skipping` — worth confirming path filters / gate still run the wheel tests before merge.

mooncake-wheel/mooncake/structured_object_store.py
```diff
@@ -5538,15 +5565,71 @@ def _call_write_with_optional_config(fn: Any, *args: Any, config: Any = None) ->
...
+        group_ids = list(group_ids)
+    if len(group_ids) != 1:
+        raise ValueError(
+            "structured object store config.group_ids must contain exactly one "
+            "logical group id; it is expanded across internal Mooncake keys"
+        )
+    if key_count == 1 and not normalize_to_list:
+        return config
+    grouped_config = _copy_write_config(config)
+    grouped_config.group_ids = [group_ids[0]] * key_count
```
zxpdemonio, line 6135: Looks correct: require exactly one logical group id, expand to `key_count`, and avoid mutating the caller’s config (copy first; early-return for already-list length-1).

Please also handle / document:
- empty `group_ids` (`[]`) — currently hits the same generic error; an explicit “empty group_ids” message would be clearer
- whether pre-expanded `group_ids` with `len == key_count` should ever be accepted (today always rejected if `len != 1`, which matches the PR text)

mooncake-wheel/mooncake/structured_object_store.py
```diff
@@ -5538,15 +5565,71 @@ def _call_write_with_optional_config(fn: Any, *args: Any, config: Any = None) ->
...
+            continue
+        if callable(value):
+            continue
+        try:
+            setattr(copied, name, value)
+        except Exception as error:
+            raise TypeError(
+                f"structured object store config field {name!r} is not writable"
+            ) from error
+    return copied
```
zxpdemonio, line 6167: The `copy.copy` → default-construct + `dir()`/`setattr` fallback is understandable for pybind `ReplicateConfig`.

**Should-fix:** prefer an allowlisted field copy (`replica_num`, `with_soft_pin`, `preferred_segment`, `group_ids`, …) instead of reflecting all public attributes. Reflective copy can pick up unexpected read-only/computed fields on future pybind changes and fail noisily, or shallow-share mutable nested state.

The real-store `ReplicateConfig` roundtrip test is good coverage — keep it.

mooncake-wheel/mooncake/structured_object_store.py
```diff
@@ -5564,7 +5647,11 @@ def _batch_put_from_with_optional_config(
     config: Any = None,
 ) -> Sequence[int]:
     return _call_write_with_optional_config(
-        batch_put_from, list(keys), list(ptrs), list(sizes), config=config
+        batch_put_from,
+        list(keys),
+        list(ptrs),
+        list(sizes),
+        config=_config_for_grouped_keys(config, len(keys)),
```
zxpdemonio, line 6218: Good central chokepoint for multi-key `batch_put_from`.

Note: `batch_put_buffer_groups_from` still calls this with a **single** key per iteration, so expansion there is a no-op beyond string→list normalization. That is API-correct, but different from contiguous chunk batch puts that expand to N in one RPC. A one-line comment near `batch_put_buffer_groups_from` would prevent future “why isn’t group_ids length N here?” confusion.

mooncake-wheel/mooncake/structured_object_store.py
```diff
@@ -3447,23 +3449,24 @@ def put_bundle_manifest(
...
                 "buffers": dict(buffers),
-                "buffer_object_ids": sorted(
-                    {
-                        str(payload_spec["key"])
-                        .removeprefix(f"{self._key_prefix}/")
-                        .split("/buffer/", 1)[0]
-                        for payload_spec in buffers.values()
-                    }
-                ),
+                "buffer_object_ids": sorted(item for item in buffer_object_ids if item),
```
zxpdemonio, line 3642: This second-commit change is important: `buffer_object_ids` now includes object ids derived from `cleanup_keys`, which unblocks meta-only / append manifests where old-stage keys must remain valid cleanup targets under `_validate_manifest`.

Please mention this explicitly in the PR description (it is easy to read as “groups only”). The new `test_bundle_manifest_update_allows_meta_only_cleanup_keys` is the right regression.

mooncake-wheel/mooncake/structured_object_store.py
```diff
@@ -3751,6 +3757,16 @@ def _policy(
...
         return result
 
+    def _object_id_from_bundle_key(self, key: str) -> str | None:
+        prefix = f"{self._key_prefix}/"
+        if not key.startswith(prefix):
+            return None
+        suffix = key[len(prefix) :]
+        for marker in ("/manifest", "/meta", "/buffer/"):
+            if marker in suffix:
+                return suffix.split(marker, 1)[0]
```
zxpdemonio: Helper looks fine for current key shapes (`…/<object_id>/{manifest,meta,buffer/…}`).

Nit: `marker in suffix` is substring-based; an object_id that itself contained `/meta` would truncate early. Unlikely with UUID-style ids, but `rsplit`/`partition` on known suffixes would be stricter if ids ever become free-form.

mooncake-wheel/tests/test_structured_object_store.py
```diff
@@ -434,6 +439,183 @@ def write_manifest(
...
 
+def _seen_group_ids(configs: list[object]) -> list[list[str]]:
+    return [
+        list(config.group_ids)
+        for config in configs
+        if config is not None and hasattr(config, "group_ids")
+    ]
+
+
+def test_structured_object_expands_store_replicate_config() -> None:
```
zxpdemonio, line 504: Good coverage of real `mooncake.store.ReplicateConfig` (no `copy.copy`), multi-chunk expansion, multi-buffer, append+cleanup, and ambiguous group rejection.

Optional extras:
- `group_ids` passed as a plain string (normalization path)
- parallel `put_mode` path still preserves the single logical group id
- assert caller `config.group_ids` identity/length unchanged after a multi-key expand (you already check the ReplicateConfig case)

zxpdemonio:
@stmatengss Could you please help review this PR when you have time? It adds grouped-write semantics for structured objects, including append manifest updates, without changing the Catalog lifecycle boundary. All CI checks are passing. Thanks.

zxpdemonio:
@stmatengss @yokinoshitayoki Could you please help review this PR when you have time? It preserves grouped writes and append manifest updates while keeping the Catalog lifecycle boundary unchanged. The current CI gate is green. Thanks.

yokinoshitayoki:
This direction looks good. Applying group semantics to structured objects is useful. As a follow-up, upper layers will also need to generate and pass the same group_id to Mooncake.

mooncake-wheel/mooncake/structured_object_store.py
```diff
@@ -6001,15 +6024,89 @@ def _call_write_with_optional_config(fn: Any, *args: Any, config: Any = None) ->
...
+        return config
+    group_ids = getattr(config, "group_ids", None)
+    if group_ids is None:
+        return config
+    normalize_to_list = isinstance(group_ids, str)
+    if normalize_to_list:
+        group_ids = [group_ids]
+    else:
+        group_ids = list(group_ids)
+    if not group_ids:
```
yokinoshitayoki, line 6123: group_ids=[] is returned unchanged here, and the new test treats it as ungrouped. However, the underlying ReplicateConfig uses std::optional<std::vector<std::string>>, so an empty list means the optional is set with length zero. Mooncake Store requires group_ids.size() to match the number of keys, causing a single-key write to fail with INVALID_PARAMS.
Please copy the config and normalize an empty list to group_ids=None. The test should also use a Store stub that enforces the real group_ids length contract, since the current InMemoryStore accepts this invalid configuration.

zxpdemonio:
@yokinoshitayoki Thanks for the feedback. We addressed this in `cbd12fd3`: upper layers no longer need to generate or pass a `group_id`.

The implementation keeps the storage group entirely inside the structured-object layer:

- An initial `dict` or DataProto put generates a private `structured-<uuid>` group ID when none is provided.
- The ID is stored on `MooncakeDataProtoRef` and serialized in the opaque v2 DataProto handle, so it survives cross-instance handoff.
- Subsequent append calls import the handle, inherit the same ID, and apply it to all newly written Store keys. A conflicting explicit group config is rejected before any write.
- Legacy v1 handles remain ungrouped and retain the existing caller-config behavior.

This does not require a root object, Catalog entry, Store lookup, or hard pin. It also does not change append snapshot semantics: concurrent appends from the same base still return separate refs and are not implicitly merged.

The focused group/append tests and the full structured-object test file pass (`103 passed`, with the existing torch/CUDA-dependent cases skipped on the test host).

mooncake-wheel/mooncake/structured_object_store.py
```diff
@@ -479,6 +492,35 @@ def __init__(
...
+                raise ValueError(
+                    "structured object store config.group_ids must contain strings"
+                ) from error
+            if len(group_ids) > 1:
+                raise ValueError(
+                    "structured object store config.group_ids must contain exactly one "
+                    "logical group id; it is expanded across internal Mooncake keys"
+                )
+            group_id = group_ids[0] if group_ids else None
+        if group_id is None:
```
yokinoshitayoki, line 516: My remaining concern is the explicit ungrouped case: according to the ReplicateConfig contract, group_ids=[""] means that the key should remain ungrouped, but _new_storage_group_id() currently replaces it with a generated group ID, so callers cannot opt out of automatic grouping. Could we preserve [""] as ungrouped?  
Also, legacy compatibility appears one-way: new readers accept v1 handles, but old readers will reject the v2 handles now produced by default.

zxpdemonio: Thanks, addressed in `33782964`.

- `group_ids=[""]` (and the equivalent empty string form) now remains explicitly ungrouped. `_new_storage_group_id()` returns `None`, the caller config is passed through unchanged, and the exported handle omits `storage_group_id`. `None` / `[]` still mean no logical ID was specified, so the structured-object layer generates one automatically.
- Exported handles remain version 1. `storage_group_id` is an optional top-level v1 extension; the previous v1 reader ignores unknown fields, while the new reader imports the field and preserves it across cross-instance append. I verified this with the pre-change reader implementation: it successfully imported and materialized a newly exported grouped handle.

The mixed-version boundary is that an old reader can read and clean up the new v1 handle, but if that old reader appends or re-exports it, it drops the unknown group field and cannot preserve automatic grouping. Fully preserving group semantics through an old intermediate writer would require a persisted catalog/lookup mechanism, which would substantially expand this PR.

zxpdemonio: One additional transition safeguard is included in `6a73dd8d`: export remains v1 for old-reader compatibility, while the new importer still accepts the v2 handles produced by the earlier Draft revision. v2 continues to require a valid `storage_group_id`.

zxpdemonio: Clarification on the scope boundary: the Catalog work is already tracked separately. #3345 provides the DataProto metadata catalog, and #3374 adds fragment/read-buffer lifecycle management. #3133 should remain limited to Store grouped-write semantics and carrying the storage group in structured-object refs/handles; it should not add a Catalog or lookup layer. The old-intermediate-writer case is therefore a cross-version integration boundary for the catalog-managed path, not additional scope for this PR.

zxpdemonio:
@yokinoshitayoki To clarify the responsibility boundary:

- #3133 owns structured-object storage grouping. It generates the group ID inside the structured-object layer, applies it to the object’s physical Store writes, and persists it in DataProto handles so append can reuse it. Upper layers do not need to provide a group ID.
- #3374 owns lifecycle management for Catalog-managed DataProto fragments: publication, reader pin/release, retirement, and physical cleanup. It consumes structured-object handles but does not generate or define their storage group IDs.

yokinoshitayoki (approved):
Thanks a lot for your clarification, this PR looks good to me now.

stmatengss (approved):

PR #3135, opened 2026-07-27
[Store] Make re-offload of persisted keys an idempotent skip in BatchOffload

### LINKED ISSUE #2827: [Bug]: [SSD] DSV4 enable SSD offload，stress testing, repeat offloading the same batch key, lead to OBJECT_ALREADY_EXISTS/persist failed/INVALID_KEY
### Bug Report

# 测试场景：
DeepSeekV4模型，PD分离，4机，1P1D，P和D都是双机。

## P节点的1台机器的vllm启动脚本。
```shell
unset ftp_proxy
unset https_proxy
unset http_proxy
rm -rf ~/ascend/log
rm -rf /opt/XXX/Test/vLLM/mooncake-ssd/logs/*
#!/bin/bash
TIMESTAMP=$(date +"%Y%m%d%H%M")

#export HCCL_INTRA_PCIE_ENABLE=1
#export HCCL_INTRA_ROCE_ENABLE=0

MASTER_IP_ADDRESS="x.xxx.xx.xx"
IP_ADDRESS="x.xxx.xx.xx"
SERVICE_PORT=10000
LOG_FILE="/opt/XXX/Test/vLLM/mooncake-ssd/logs/deepseekv4_start_prefill_${TIMESTAMP}.log"

NETWORK_INTERFACE="eth0"
export ASCEND_RT_VISIBLE_DEVICES=0,1,2,3,4,5,6,7
PREFILL_DATA_PARALLEL_RANK=0

PREFILL_DATA_PARALLEL_SIZE=2
PREFILL_TENSOR_PARALLEL_SIZE=8

MODEL_PATH="/opt/Dsv4-flash-bf16"
MODEL_NAME="DeepSeek-V4-Flash"

# # mooncake
export MOONCAKE_CONFIG_PATH="/opt/XXX/Test/vLLM/mooncake-ssd/configs/mooncake.json"
export LD_LIBRARY_PATH=/usr/local/lib:/usr/local/lib64:${LD_LIBRARY_PATH}
export PYTHONHASHSEED=0
export HCCL_INTRA_ROCE_ENABLE=1
export ASCEND_CONNECT_TIMEOUT=10000
export ASCEND_TRANSFER_TIMEOUT=10000

# # jemalloc
export LD_PRELOAD=/usr/lib64/libjemalloc.so.2:$LD_PRELOAD
# # AIV
export HCCL_OP_EXPANSION_MODE="AIV"
export TASK_QUEUE_ENABLE=1
export VLLM_RPC_TIMEOUT=3600000
export VLLM_EXECUTE_MODEL_TIMEOUT_SECONDS=30000
export HCCL_EXEC_TIMEOUT=1200
export HCCL_CONNECT_TIMEOUT=1200

export HCCL_IF_IP=$IP_ADDRESS
export GLOO_SOCKET_IFNAME=$NETWORK_INTERFACE
export TP_SOCKET_IFNAME=$NETWORK_INTERFACE
export HCCL_SOCKET_IFNAME=$NETWORK_INTERFACE
export OMP_PROC_BIND=false
export OMP_NUM_THREADS=10
export PYTORCH_NPU_ALLOC_CONF=expandable_segments:True
export HCCL_BUFFSIZE=1024

#export VLLM_VERSION=0.20.2
export VLLM_ASCEND_APPLY_DSV4_PATCH=1
export VLLM_ASCEND_ENABLE_FLASHCOMM1=1
#export USE_MULTI_GROUPS_KV_CACHE=1
#export USE_MULTI_BLOCK_POOL=1

vllm serve "$MODEL_PATH" \
--host "$IP_ADDRESS" \
--port "$SERVICE_PORT" \
--data-parallel-size "$PREFILL_DATA_PARALLEL_SIZE" \
--data-parallel-rank "$PREFILL_DATA_PARALLEL_RANK" \
--data-parallel-address "$MASTER_IP_ADDRESS" \
--data-parallel-rpc-port 12321 \
--tensor-parallel-size "$PREFILL_TENSOR_PARALLEL_SIZE" \
--enable-expert-parallel \
--served-model-name "$MODEL_NAME" \
--seed 1024 \
--max-model-len 204800 \
--max-num-batched-tokens 4096 \
--max-num-seqs 16 \
--block-size 128 \
--enforce-eager \
--async-scheduling \
--no-disable-hybrid-kv-cache-manager \
--no-enable-prefix-caching \
--trust-remote-code \
--gpu_memory_utilization 0.95 \
--speculative-config '{"num_speculative_tokens": 1,"method": "mtp","enforce_eager": true}' \
--safetensors-load-strategy 'prefetch' \
--model-loader-extra-config='{"enable_multithread_load": "true", "num_threads": 128}' \
--tokenizer-mode deepseek_v4 \
--tool-call-parser deepseek_v4 \
--enable-auto-tool-choice \
--reasoning-parser deepseek_v4 \
--additional-config '{"enable_cpu_binding": true, "enable_shared_expert_dp": true}' \
--kv-transfer-config \
    '{"kv_connector": "MultiConnector",
    "kv_role": "kv_producer",
    "kv_connector_extra_config": {
        "connectors": [{
	    "kv_connector": "MooncakeHybridConnector",
	    "kv_role": "kv_producer",
	    "kv_port": "30000",
	    "kv_connector_extra_config": {
                "prefill": {
                        "dp_size": 2,
                        "tp_size": 8
                },
                "decode": {
                        "dp_size": 2,
                        "tp_size": 8
                }
	    }
        },
	{
	    "kv_connector": "AscendStoreConnector",
	    "kv_role": "kv_producer",
	    "kv_connector_extra_config": {
                "backend": "mooncake",
		"lookup_rpc_port": "0"
            }
	}]
    }
    }' > "$LOG_FILE" 2>&1 &
```
## mooncake_master的启动命令。
```shell
export LD_LIBRARY_PATH=/usr/local/lib:/usr/local/lib64:${LD_LIBRARY_PATH}

mooncake_master \
  --rpc_port=39486 \
  --eviction_high_watermark_ratio=0.7 \
  --eviction_ratio=0.2 \
  --default_kv_lease_ttl=11000 \
  --enable_offload=true \
  --offload_on_evict=true
```

# P节点的vllm日志
```text
W20260708 12:39:38.314945  7519 storage_backend.cpp:1353] Duplicate key detected in BatchOffload: Dsv4-flash-bf16@pcp0@dcp0@head_or_tp_rank:0@pp_rank:0@group:0@cache_role:kv@cache_family:c4@cb74f08f554769a92285d5d1a278df180c9d1763a35ca22a68da22ddf5fbec7f, bucket_id=7304518881347. Returning OBJECT_ALREADY_EXISTS.
...
E20260708 12:48:28.416213  7519 file_storage.cpp:731] Key not found: Dsv4-flash-bf16@pcp0@dcp0@head_or_tp_rank:0@pp_rank:0@group:0@cache_role:kv@cache_family:c4@797000e494ea5584f4407dbdd0f8087369dc029a48ccecbc37b63875e55dc7a2
E20260708 12:48:28.416296  7519 file_storage.cpp:382] BatchQuerySlices failed with error: INVALID_KEY
E20260708 12:48:30.313634  7528 file_storage.cpp:731] Key not found: Dsv4-flash-bf16@pcp0@dcp0@head_or_tp_rank:0@pp_rank:0@group:0@cache_role:kv@cache_family:c4@8d1310d6592a04f35eb7eb83c04b6f407b857e81bd428f18339d54580aef6d9a
E20260708 12:48:30.313707  7528 file_storage.cpp:382] BatchQuerySlices failed with error: INVALID_KEY
E20260708 12:48:30.458350  7511 file_storage.cpp:731] Key not found: Dsv4-flash-bf16@pcp0@dcp0@head_or_tp_rank:0@pp_rank:0@group:0@cache_role:kv@cache_family:c4@6b288f4977b946d91003ddcca6332024e5f82185632c7f1c16f600edde81a2fc
E20260708 12:48:30.458447  7511 file_storage.cpp:382] BatchQuerySlices failed with error: INVALID_KEY
E20260708 12:48:30.642908  7526 file_storage.cpp:731] Key not found: Dsv4-flash-bf16@pcp0@dcp0@head_or_tp_rank:0@pp_rank:0@group:0@cache_role:kv@cache_family:c4@dceb210cdce4c83faf29c754daf4918a4ab549fd0ab270dff3f2305735bbf059
E20260708 12:48:30.643005  7526 file_storage.cpp:382] BatchQuerySlices failed with error: INVALID_KEY
E20260708 12:48:30.651410  7513 file_storage.cpp:731] Key not found: Dsv4-flash-bf16@pcp0@dcp0@head_or_tp_rank:0@pp_rank:0@group:5@cache_role:kv@cache_family:c1@b748b1174ac20e00b2329dbdaf9d84f54fc91145820592d0085169055936f49b
E20260708 12:48:30.651494  7513 file_storage.cpp:382] BatchQuerySlices failed with error: INVALID_KEY
E20260708 12:48:30.946017  7517 file_storage.cpp:731] Key not found: Dsv4-flash-bf16@pcp0@dcp0@head_or_tp_rank:0@pp_rank:0@group:5@cache_role:kv@cache_family:c1@b2a33ffe860be7487d2f3abe1dd8062ad2f01d474a9378cda4507368b1786913
E20260708 12:48:30.946106  7517 file_storage.cpp:382] BatchQuerySlices failed with error: INVALID_KEY
...
E20260708 12:57:39.994860 11520 storage_backend.cpp:1424] Key not found: Dsv4-flash-bf16@pcp0@dcp0@head_or_tp_rank:0@pp_rank:0@group:0@cache_role:kv@cache_family:c4@16e8aff8f59dd4f5396cff3500306159171b8eb1745e182f9f3f7ccd94827d35
E20260708 12:57:39.995028 11521 storage_backend.cpp:1424] Key not found: Dsv4-flash-bf16@pcp0@dcp0@head_or_tp_rank:0@pp_rank:0@group:0@cache_role:kv@cache_family:c4@16e8aff8f59dd4f5396cff3500306159171b8eb1745e182f9f3f7ccd94827d35
E20260708 12:57:39.995077 11521 file_storage.cpp:690] Batch load object failed,err_code = INVALID_KEY
E20260708 12:57:39.995029 11520 file_storage.cpp:690] Batch load object failed,err_code = INVALID_KEY
E20260708 12:57:39.995108 11520 file_storage.cpp:311] Batch load object failed,err_code = INVALID_KEY
E20260708 12:57:39.995096 11521 file_storage.cpp:311] Batch load object failed,err_code = INVALID_KEY
E20260708 12:57:39.995134 11520 real_client.cpp:5434] Batch get offload object failed,err_code = INVALID_KEY
E20260708 12:57:39.995139 11521 real_client.cpp:5434] Batch get offload object failed,err_code = INVALID_KEY
...

```

单条请求测试（50K输入，1K输出）时，mooncake + 池化 + ssd没有问题。
32并发，50K输入，1K输出时，会报错，报错见：deepseekv4_start_prefill_XX.log

# 分析
## 问题定位结论

32 并发下失败的核心链路是 **Mooncake SSD Offload 并发写冲突 → 整批 persist 失败 → Master 与本地 SSD 元数据不一致 → D 侧读 KV 时 `INVALID_KEY`**。单请求不触发，是因为没有并发重复 offload。

---

### 1. 错误时间线（日志）

| 时间 | 现象 | 含义 |
|------|------|------|
| **12:39:38** | 首次 `Duplicate key detected in BatchOffload` → `OBJECT_ALREADY_EXISTS` | 多线程/多请求同时 offload 同一 KV block |
| **12:48 ~ 13:10+** | 大量 `Key not found` → `BatchQuerySlices/BatchLoad INVALID_KEY` | SSD 本地索引找不到 block |
| **12:57:39** | `Batch get offload object failed, err_code = INVALID_KEY`（约 4.7 万次） | D 节点经 RPC 从 P 节点 SSD 拉 KV 失败 |
| **14:19+** | 32 并发 POST 开始，`missing_count=N` 持续出现 | KV 传输时 block 缺失，需补存 |

---

#### 根因 A：并发 offload 同一 KV key（主因）

日志典型片段：

```
W ... Duplicate key detected in BatchOffload: Dsv4-flash-bf16@pcp0@dcp0@head_or_tp_rank:0@...@{block_hash}
    bucket_id=7304518881347. Returning OBJECT_ALREADY_EXISTS.
I ... Cleaned up orphaned bucket data file: /opt/mooncake_offload/rank_8/7304518881347.bucket
E ... Failed to store objects with error: OBJECT_ALREADY_EXISTS
E ... Failed to persist objects with error: OBJECT_ALREADY_EXISTS
```

触发条件（32 并发 + 池化 + prefix caching）：
- 多请求共享 prefix block → **相同 block hash**
- Mooncake 池化后多请求同时触发 offload
- 本地 `object_bucket_map_` 检测到 key 已存在，返回 `OBJECT_ALREADY_EXISTS`

本地代码路径（`storage_backend.cpp`）：

```1344:1353:Mooncake/mooncake-store/src/storage_backend.cpp
        // Pre-check for duplicates before modifying any state
        for (const auto& key : committed_keys) {
            if (object_bucket_map_.find(key) != object_bucket_map_.end()) {
                LOG(WARNING)
                    << "Duplicate key detected in BatchOffload: " << key
                    << ", bucket_id=" << bucket_id
                    << ". Returning OBJECT_ALREADY_EXISTS.";
                lock.unlock();
                CleanupOrphanedBucket(bucket_id);
                return tl::make_unexpected(ErrorCode::OBJECT_ALREADY_EXISTS);
```

#### 根因 B：`file_storage.cpp` 未将 `OBJECT_ALREADY_EXISTS` 视为幂等成功

`PutStart` 路径会把 `OBJECT_ALREADY_EXISTS` 当成功：

```1553:1555:Mooncake/mooncake-store/src/client_service.cpp
        if (err == ErrorCode::OBJECT_ALREADY_EXISTS) {
            VLOG(1) << "object_already_exists key=" << key;
            return {};
```

但 **Offload persist 路径**遇到该错误会直接失败整批：

```563:577:Mooncake/mooncake-store/src/file_storage.cpp
        if (!offload_res) {
            LOG(ERROR) << "Failed to store objects with error: "
                       << offload_res.error();
            ...
            } else {
                return tl::make_unexpected(offload_res.error());
            }
        }
```

后果：
1. 一个 bucket 里只要有一个 duplicate key，**整批 keys 都 persist 失败**
2. 部分 key 可能已在 Master 注册 LOCAL_DISK replica，但本地 SSD 实际未写入
3. D 侧 `BatchGet` → `Key not found` → **`INVALID_KEY`**

#### 根因 C：配置加剧并发冲突

| 配置项 | 当前值 | 影响 |
|--------|--------|------|
| `enable_prefix_caching` | 日志中为 **True**（`run_pd-p.sh` 写了 `--no-enable-prefix-caching`，但实际运行未生效） | 多请求共享 block，duplicate offload 概率高 |
| `default_kv_lease_ttl` | **11000ms（11 秒）** | 50K prefill 远超 11s，lease 过期可能导致 Master 侧 key 失效 |
| `max_num_seqs` | 16 | 32 并发下仍有足够并行度触发 race |
| `offload_on_evict=true` | 开启 | 高并发下 eviction + 重复 offload 叠加 |

---

### 3. 错误传播链

```mermaid
sequenceDiagram
    participant P as P节点 vLLM
    participant MC as Mooncake Client
    participant SSD as 本地 SSD Backend
    participant M as mooncake_master
    participant D as D节点

    P->>MC: 多请求并发 offload 同一 block
    MC->>SSD: BatchOffload (WriteBucket)
    SSD-->>SSD: 检测到 Duplicate key
    SSD-->>MC: OBJECT_ALREADY_EXISTS
    MC-->>MC: file_storage 整批 persist 失败
    Note over M,SSD: Master 可能有 replica 记录，SSD 无数据
    D->>MC: BatchGet offload object
    MC->>SSD: BatchLoad
    SSD-->>D: INVALID_KEY (Key not found)
```

---

### 4. 非主因（可忽略）

- `libjemalloc.so.2` LD_PRELOAD 警告：环境缺库，不影响功能
- Traceback `ValueError: ... is not a multimodal model`：个别请求带了 image 内容，与 Mooncake KV 无关
- `reasoning token IDs` 警告：DSV4 parser 配置问题，非本次根因

---

### 5. 修复建议（按优先级）

**P0 — Mooncake 代码修复**

1. 在 `file_storage.cpp` 的 `OffloadObjects` 中，对 `OBJECT_ALREADY_EXISTS` **按幂等成功处理**（与 `PutStart` 一致），不要 fail 整批。
2. 或在 `storage_backend.cpp` 的 `BatchOffload` 中，duplicate key 时 **跳过该 key、继续提交其余 key**，而不是 return error。
3. 同一 bucket 内 duplicate 时，只对 duplicate key 跳过，不影响 batch 内其他 key。

**P1 — 配置调整**

```bash
# start_mooncake_master.sh - lease 应覆盖完整 prefill + transfer 时间
--default_kv_lease_ttl=600000   # 建议 10 分钟，按实际 prefill 时长调整

# run_pd-p.sh - 确认 prefix caching 关闭（与脚本意图一致）
--no-enable-prefix-caching

# 若 32 并发仍冲突，可先压测验证
--max-num-seqs 32
```

**P2 — 验证步骤**

1. 修复后重跑 32 并发，确认不再出现 `Failed to persist objects with error: OBJECT_ALREADY_EXISTS`
2. 确认 `Batch get offload object failed, INVALID_KEY` 归零
3. 确认 D 侧 `missing_count` 不再持续增长
4. 检查 `/opt/mooncake_offload/rank_*/` 下 bucket 文件与 Master 元数据一致

---

### 6. 总结

| 项目 | 结论 |
|------|------|
| **直接报错** | `OBJECT_ALREADY_EXISTS` → `INVALID_KEY` |
| **根本原因** | 32 并发下同一 KV block 被重复 offload，`file_storage` 未做幂等处理，整批 persist 失败 |
| **为何单请求正常** | 无并发 race，duplicate 不触发 |
| **修复方向** | Offload 路径对 `OBJECT_ALREADY_EXISTS` 幂等 + 增大 lease TTL + 确认关闭 prefix caching |

### Before submitting...

he-yufeng:
## Description

Fixes #2827.

Under stress or retry the same keys get offloaded again, and today that fails the whole batch: `PrepareEviction` rejects any write key already visible in `object_bucket_map_` or `pending_write_keys_`, and the commit phase repeats the same all-or-nothing check. One duplicate key therefore kills every fresh key in the same bucket, and the error propagates out of `FileStorage` as a persist failure. `GroupOffloadingKeysByBucket` already skips persisted keys under `offloading_mutex_`, but its `IsExist` check and the commit under `mutex_` sit on different locks with `BuildBucket`/`WriteBucket` disk I/O in between, so a racing duplicate still reaches this path. This was the residual window discussed in the issue with @LujhCoconut: grouping is the fast path, this change is the correctness net under `mutex_`.

The change makes re-offload behave like Put, an idempotent no-op:

- `PrepareEviction` moves duplicates into a new `skipped_keys` list instead of rejecting the batch, and only registers the genuinely new keys in `pending_write_keys_`.
- The commit phase skips those keys (plus anything committed since, as defense in depth) and inserts only the rest into `object_bucket_map_`. The complete handler and the rollback path only ever see newly committed keys, so Master is never notified twice for the same key.
- A batch where every key is already persisted succeeds without invoking the complete handler; the redundant bucket file is cleaned up exactly as the old rejection path did.
- A key that is pending eviction still fails the batch. Skipping there could drop the only replica once eviction lands, so loud failure stays the right behavior.

This intentionally changes the duplicate contract from whole-batch atomic rejection to skip-and-commit, per the semantics aligned in the issue. The three Phase 0 duplicate tests and the reentrant pending-write test are updated to the new contract; the partial-batch test now also asserts that the handler only receives the newly committed key and that the new key reads back its own value.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

Built and ran in a Linux container (ubuntu:22.04 base with the store deps, macOS cannot build the store natively):

```bash
cmake --build build-linux --target storage_backend_test master_service_test
./build-linux/mooncake-store/tests/storage_backend_test
./build-linux/mooncake-store/tests/master_service_test
```

**Test results:**
- [x] Unit tests pass

`storage_backend_test`: 82/82 pass, including the four updated duplicate tests (`DuplicateKeyIdempotentSkip`, `DuplicateKeyCleanupOrphanedFiles`, `DuplicateBatchPartialSkip`, `PendingWriteSkipsReentrantDuplicate`). `master_service_test`: 144/144 pass (the diff does not touch master code, run as a regression guard).

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

Kimi Code (AI coding agent) helped draft the implementation and the test updates. I reviewed the full diff, traced the concurrent paths (pending write, pending eviction, rollback) by hand, and ran the builds and tests above myself in the container.


LujhCoconut:
The direction is right — an idempotent skip for duplicate re-offload (Put semantics) is exactly what #2827 asks for, keeping the loud failure for pending-eviction keys is prudent, and the tests match the new contract. 

After tracing the code paths, though, I believe this revision has two blocking issues, one race, and one reporting-loop gap that should be fixed before merge. 

**Root cause** in one sentence: `PrepareEviction` snapshots key state under `mutex_` (`skipped_keys` / `write_keys` / `write_size`), but that snapshot is then followed — outside the lock — by the eviction RPC + `FinalizeEviction` + `WriteBucket`; the commit phase never revalidates `skipped_keys` against the current `object_bucket_map_`, and quota accounting (`write_size` / `total_size_`) is booked for the full bucket rather than the actually-committed set. Concretely:

- **Blocking 1 — An all-duplicate batch can trigger real, unnecessary evictions**

  - With all keys skipped, `result.write_keys` is empty, yet `pending_write_size_` still grows by the full bucket size (`storage_backend.cpp:2826-2827`); the quota loop (`2880-2889`) and the disk check (`2857`, keyed on the input `write_keys`) evict real buckets for phantom demand, and `FinalizeEviction` (`1728`) deletes them *before* commit discovers there is nothing to commit (`1784-1789`). All-duplicate batches need no race: the carry-over path (`2378-2388`) re-emits pooled keys without an `IsExist` recheck. Same root cause, second symptom: under exhaustion the guard at `2939` (keyed on the input `write_keys`) returns `FILE_WRITE_FAIL` for an all-duplicate batch, and `OffloadObjects` retry-loops (rollback itself is correct — no data loss).

- **Blocking 2 — `total_size_` leaks permanently, including across restarts** 
  - Commit credits the full `bucket->data_size` (`1792`), eviction only debits keys whose map entry points at that bucket (`2908-2919`), so skipped keys' bytes are never reclaimed — A(100B) in bucket 1, A+B re-offloaded with A skipped into bucket 2: after both buckets are evicted, `total_size_` retains +100. `Init` re-credits the full serialized size (`2139-2140`), so the leak survives restarts.

- *skipped keys are never revalidated at commit*
  - he unconditional `continue` (`1771-1772`) makes the recheck at `1774` dead code for exactly the keys that need it; if the old bucket is evicted between the snapshot and commit (the window spans the eviction RPC + `FinalizeEviction` + `WriteBucket`, all outside the lock), the old copy is deleted and the new bytes get no index entry — the key becomes unreadable.

- **skipped keys fall out of the Master reporting loop**
  - They are in `all_bucket_keys` (`file_storage.cpp:509-512`), so they never reach the -1 sentinel report (`641-647`), and they are not in `committed_keys` either — the Master-side `offloading_task` and the source replica's refcount linger until the TTL reaper fires (`master_service.cpp:7330-7349`, ~10 min default).

The fixes are all local (~20-30 lines) — specifics and code are in the inline comments below.

mooncake-store/src/storage_backend.cpp
```diff
@@ -1767,36 +1757,51 @@ tl::expected<int64_t, ErrorCode> BucketStorageBackend::BatchOffload(
...
+        object_bucket_map_.reserve(object_bucket_map_.size() +
+                                   committed_indices.size());
+        for (size_t i : committed_indices) {
+            auto [it, inserted] =
+                object_bucket_map_.insert({bucket_keys[i], metadatas[i]});
+            CHECK(inserted)
+                << "Reserved key became duplicated: " << bucket_keys[i];
+            committed_keys.push_back(bucket_keys[i]);
+            committed_metadatas.push_back(metadatas[i]);
+        }
```
LujhCoconut, line 1865: suggestion
```cpp
        int64_t committed_data_size = bucket->meta_size;
        object_bucket_map_.reserve(object_bucket_map_.size() +
                                   committed_indices.size());
        for (size_t i : committed_indices) {
            auto [it, inserted] =
                object_bucket_map_.insert({bucket_keys[i], metadatas[i]});
            CHECK(inserted)
                << "Reserved key became duplicated: " << bucket_keys[i];
            committed_data_size +=
                metadatas[i].data_size + metadatas[i].key_size;
            committed_keys.push_back(bucket_keys[i]);
            committed_metadatas.push_back(metadatas[i]);
        }
        total_size_ += committed_data_size;
```

mooncake-store/src/storage_backend.cpp
```diff
@@ -1767,36 +1757,51 @@ tl::expected<int64_t, ErrorCode> BucketStorageBackend::BatchOffload(
...
-                CHECK(inserted)
-                    << "Reserved key became duplicated: " << bucket_keys[i];
+            if (object_bucket_map_.find(bucket_keys[i]) ==
+                object_bucket_map_.end()) {
+                committed_indices.push_back(i);
+            } else {
+                VLOG(1) << "Key already committed by a concurrent offload, "
+                           "skipping: "
+                        << bucket_keys[i];
             }
```
LujhCoconut, line 1844: Skipped keys are trusted blindly here, but the snapshot can be stale by commit time: between `PrepareEviction` and this lock (eviction RPC + `FinalizeEviction` + `WriteBucket` in between), a skipped key's old bucket can be evicted — old copy deleted, new bytes in this file with no index entry. Revalidate against the current map:

suggestion
```cpp
        for (size_t i = 0; i < bucket_keys.size(); ++i) {
            if (skipped_keys.find(bucket_keys[i]) != skipped_keys.end()) {
                // May have been evicted since the snapshot: skip only if
                // still present, otherwise it must be committed.
                if (object_bucket_map_.find(bucket_keys[i]) !=
                    object_bucket_map_.end()) {
                    continue;
                }
            }
            if (object_bucket_map_.find(bucket_keys[i]) ==
                object_bucket_map_.end()) {
                committed_indices.push_back(i);
            } else {
                VLOG(1) << "Key already committed by a concurrent offload, "
                           "skipping: "
                        << bucket_keys[i];
            }
        }
```

mooncake-store/src/storage_backend.cpp
```diff
@@ -2803,16 +2808,25 @@ BucketStorageBackend::PrepareEviction(
...
                 return tl::make_unexpected(ErrorCode::OBJECT_ALREADY_EXISTS);
             }
+            result.write_keys.push_back(key);
         }
-        result.write_keys = write_keys;
         result.write_size = required_size;
         pending_write_size_ += required_size;
-        pending_write_keys_.insert(write_keys.begin(), write_keys.end());
+        pending_write_keys_.insert(result.write_keys.begin(),
+                                   result.write_keys.end());
```
LujhCoconut, line 2999: With all keys skipped, `result.write_keys` is empty, but `pending_write_size_` still grows by the full bucket size — the quota loop and disk check (`storage_backend.cpp:2857`) evict real buckets for phantom demand, and under exhaustion `storage_backend.cpp:2939` returns `FILE_WRITE_FAIL` for what should be a no-op. Book only real writes (and test `result.write_keys` instead of the input `write_keys` at `storage_backend.cpp:2857`/`2939`):


suggestion
```cpp
        if (!result.write_keys.empty()) {
            result.write_size = required_size;
            pending_write_size_ += required_size;
            pending_write_keys_.insert(result.write_keys.begin(),
                                       result.write_keys.end());
        }
```

mooncake-store/src/storage_backend.cpp
LujhCoconut:
```cpp
            uint64_t req_sz = (!result.write_keys.empty() && required_size > 0)
                                  ? static_cast<uint64_t>(required_size)
                                  : 0;
```

LujhCoconut: suggestion 

```cpp
    if (!result.write_keys.empty() && quota_exceeded) {
        RestorePreparedEvictionLocked(std::move(result));
        return tl::make_unexpected(ErrorCode::FILE_WRITE_FAIL);
    }
```

LujhCoconut, line 3128: A bucket's `keys` list is all-inclusive — `BuildBucket` filters nothing, so keys skipped as duplicates at commit time are in it too. The second loop at `2923-2926` unconditionally reports every key in the list to the Master via `eviction_handler`. When the skipped key's authoritative copy lives in an older bucket that is still alive (read-hot vs. the colder new bucket — a common case), the Master drops a live LOCAL_DISK replica record with no self-healing path. The old all-or-nothing rejection kept the keys list and the index in one-to-one correspondence, so the unconditional collection used to be safe; the PR's partial-commit semantics break that assumption.

Fix: fold the collection into the matched branch of the first loop — only keys actually erased from the index (bucket_id match) are reported. Size accounting is untouched; only the notification surface changes.

```cpp
        int64_t evicted_size = evict_meta->meta_size;
        // Remove all keys belonging to this bucket from the object map, and
        // report ONLY those to the master: a key skipped as a duplicate at
        // commit time keeps pointing at its authoritative bucket, so
        // reporting it here would make the master drop a live replica.
        for (const auto& key : evict_meta->keys) {
            auto obj_it = object_bucket_map_.find(key);
            if (obj_it != object_bucket_map_.end() &&
                obj_it->second.bucket_id == evict_id) {
                const int64_t object_size =
                    obj_it->second.data_size + obj_it->second.key_size;
                total_size_ -= object_size;
                evicted_size += object_size;
                object_bucket_map_.erase(obj_it);
                pending_eviction_keys_.insert(key);
                result.keys.push_back(key);
            }
        }
        total_size_ -= evict_meta->meta_size;
        result.evicted_size += evicted_size;
```

**Mirrored rollback** — `storage_backend.cpp:2964-2979`：
```cpp
    // Mirror the eviction loop: only keys actually erased from the index may
    // be restored. A skipped duplicate still points at its authoritative
    // bucket and must be neither re-pointed nor double-charged.
    const std::unordered_set<std::string> evicted_keys(pending.keys.begin(),
                                                       pending.keys.end());
    for (const auto& key : pending.keys) {
        pending_eviction_keys_.erase(key);
    }
    for (auto& [bucket_id, bucket_meta] : pending.buckets) {
        if (!bucket_meta || buckets_.find(bucket_id) != buckets_.end()) {
            continue;
        }

        for (size_t i = 0; i < bucket_meta->keys.size(); ++i) {
            const auto& key = bucket_meta->keys[i];
            if (evicted_keys.find(key) == evicted_keys.end()) {
                continue;
            }
            const auto& object_meta = bucket_meta->metadatas[i];
            object_bucket_map_[key] = StorageObjectMetadata{
                bucket_id, object_meta.offset, object_meta.key_size,
                object_meta.data_size, ""};
            total_size_ += object_meta.data_size + object_meta.key_size;
        }
```
> Why this must ship together: on eviction rollback (`RestorePreparedEvictionLocked`, reached when `eviction_handler` fails or the `2939` guard fires), the current code re-inserts and re-charges every key in `bucket_meta->keys`. After Fix F, `pending.keys` is exactly the set of keys that were actually erased — filtering the restore by it keeps rollback strictly symmetric with eviction. Without it, rollback would re-point a skipped key at this bucket (`object_bucket_map_[key] = ...` overwrites) and double-charge `total_size_`.

LujhCoconut, line 3186: Please refer to the comment mentioned above.

he-yufeng:
All three addressed in b1d75081:

**Main point (live replica)**: verified the path by hand before touching it. A skipped duplicate is physically in the new bucket's keys list (`BuildBucket` filters nothing), but its index entry still points at the older bucket, and the unconditional collection loop was written back when the keys list and the index were 1:1 by construction. The collection is now folded into the matched branch of the index-cleanup loop: only keys actually erased (`bucket_id == evict_id`) are reported to the master and marked pending-eviction. Size accounting (`total_size_`, `evicted_size`, `pending_eviction_size_`) is untouched, and `CommitPreparedEviction`/`RestorePreparedEvictionLocked` stay symmetric since they re-derive state from `pending.buckets`, not the notification list.

One second-order effect worth noting for the record: with the gate, an evicted redundant bucket frees its full file on disk but only its index-matched keys count against `total_size_`, so a deep watermark eviction may pick one more bucket than before to satisfy the same deficit. That is the correct behavior, just visible in tests with heavy pressure.

**req_sz**: now keys on `result.write_keys`, so a fully-duplicate batch demands no free disk space for the no-op write.

**quota gate**: same treatment, `result.write_keys` instead of the raw batch, so the all-skipped batch succeeds instead of failing FILE_WRITE_FAIL on quota.

Regression test added (`BucketEvictionDoesNotReportSkippedDuplicateKeys`): batch2 writes a bucket physically carrying the skipped duplicate; the older authoritative bucket is read-hot and survives the watermark eviction; the skipped key is not reported (only the genuinely evicted keyB is) and still resolves from the older bucket afterwards. `storage_backend_test` 83/83 pass in the Linux container; `master_service_test` green except the timing-sensitive expiry/soft-pin tests, which fail intermittently on pristine main as well (both pass on isolated rerun).

he-yufeng:
Format gate is green now too: the three blocking points from your last pass were addressed in b1d75081, and the clang-format violation in the new test is fixed in 9591c157 (the TEST_F line join plus the lambda continuation, nothing semantic). Ready for another look when you have a moment.

he-yufeng:
Resolved the conflict with current main: upstream added the LRU timestamp write inside the commit path, and this branch now computes ts and stores last_access_ns_ on the committed-indices path the same way, so the idempotent skip and the LRU index update compose. Green locally against the merged tree (configure only goes as far as the missing yalantinglibs prefix here, so behavior coverage stays with CI).

he-yufeng:
The red on the latest run is a repo-wide builder outage, not this change: main's own Build & Test (Linux) run from today fails at the same steps (apt package updates and submodule sync die before any compilation), and the same failure shows on unrelated PR branches. The format gate is green since 4777b8df. Should recover on its own once the image/network issue clears.

he-yufeng:
Rebased onto current main. Two overlaps with #3467, both resolved by union: the commit path keeps this PR's committed_indices early-exit and splice in the new LRU timestamp assignment, and the eviction guard now reads result.write_keys with the phys_exceeded condition upstream added. No behavior change intended on either side.

he-yufeng:
For reviewers reading the red lanes: the two rocm builds did not fail on the code. Both runners took a shutdown signal at exit 137 mid-build ([203/222] and the matching step on 3.12), which is runner-side. The previous red from August was my stale base against the master_service changes on main; this branch is now merged with current main and the CUDA wheel builds are the lanes that matter. A re-run of the rocm job whenever someone with permissions has a moment would clear it.

he-yufeng:
Rebased onto current main (0a3596cd). The branch had drifted 12 days behind, including the BatchOffload phase rework (eviction now notifies master before file deletion), so the four commits were replayed one by one instead of text-merging:

- PrepareEviction now splits write keys into skipped (already persisted or in flight: idempotent no-op) vs fresh, and still fails loudly for pending-eviction keys, where a skip could lose the only replica.
- The commit-phase dedup net and the all-skipped cleanup (orphaned bucket file removed, nothing notified) are unchanged in shape.
- The red CI on the old head was a self-hosted ROCm runner dying mid-build (exit 137, shutdown signal), not a compile error.

Verification on the rebased head: storage_backend_test 93/93 green in a Linux container (Ubuntu 22.04), including the skipped-duplicate eviction tests; pre-commit clean on the three touched files. The #2827 mechanism is unchanged — PrepareEviction on current main still rejects the whole batch when any write key is already visible, so this is still live.

he-yufeng:
@LujhCoconut coming back to your July threads on the old head. After the rebase onto the BatchOffload phase rework, the two regions you flagged no longer exist: the prepared-eviction restore flow (`RestorePreparedEvictionLocked`, the `quota_exceeded` branch) was removed upstream, and the bucket-key reporting loop was replaced.

The semantics behind your second comment are now handled directly: 3f79d682 makes eviction reporting go through the index-erased set only, so a key skipped as a duplicate at commit time is never reported to the Master, and no-op writes are gated on the same path.

Could you take another look when you get a chance?

LujhCoconut:
The idempotent-skip direction is still right for #2827, and the eviction-reporting fold-in plus `result.write_keys` gating look good on the new head. One severity note up front: under the content-addressed same-key-same-value assumption the store already relies on, none of the items below can serve a wrong value or corrupt committed data; the worst case is a cache entry turning into a clean miss. What they affect is accounting accuracy, master notification, and the success contract; the fixes are small and were mostly already written in July (inline below).

One correction on your Sept 13 note first: the reporting-loop part is indeed handled on this head (2947-2959), but the prepared-eviction restore flow was not removed upstream. `RestorePreparedEvictionLocked` and the quota guard are still there, both on this head (2998-3029, 2979-2982) and on current main (2922-2953, 2903-2906). And re-reading the head against the July threads, the remaining suggestions (commit-time credit, revalidation, booking gate, mirrored restore) aren't in the tree either. Details below (line numbers: head's `storage_backend.cpp`):

1. **Skipped keys never reach the Master in either direction.** They're in `all_bucket_keys` (`file_storage.cpp:382`) so the sweep doesn't NACK them, and they're not in `committed_keys` so the complete handler doesn't report them either. Their offloading tasks and **source-replica (DRAM) refcounts then linger until the TTL reaper**. Since duplicates are the common case under the retry-storm conditions of #2827, source release is systematically delayed exactly when memory is tightest. This needs a cross-layer change (surface the skipped set from `BatchOffload`, NACK with the -1 sentinel), so a tracked follow-up is fine, but it should top that list.

2. **`total_size_` credited for the full bucket at commit** (1684) vs matched-only debits in eviction (2947-2959) and rollback (2637-2644, via `committed_keys` at 1723); `Init` re-credits the full serialized size (2076-2086), so the leak survives restarts. The rate scales with duplicates that reach the skip path (the grouping `IsExist` catches most), but #2827 proves that rate is non-zero under stress, and the leak is strictly one-way. Fix:

```cpp
        int64_t committed_data_size = bucket->meta_size;
        object_bucket_map_.reserve(object_bucket_map_.size() +
                                   committed_indices.size());
        for (size_t i : committed_indices) {
            auto [it, inserted] =
                object_bucket_map_.insert({bucket_keys[i], metadatas[i]});
            CHECK(inserted)
                << "Reserved key became duplicated: " << bucket_keys[i];
            committed_data_size +=
                metadatas[i].data_size + metadatas[i].key_size;
            committed_keys.push_back(bucket_keys[i]);
            committed_metadatas.push_back(metadatas[i]);
        }
        total_size_ += committed_data_size;
```

3. **All-duplicate batches still book full size into `pending_write_size_`** (2829-2832). `req_sz` (2860) and the tail guard (2979) are gated, the booking isn't, so a no-op write can evict real buckets under quota pressure. The phantom is window-limited, but a full cache makes that window fire often. Fix:

```cpp
        if (!result.write_keys.empty()) {
            result.write_size = required_size;
            pending_write_size_ += required_size;
            pending_write_keys_.insert(result.write_keys.begin(),
                                       result.write_keys.end());
        }
```

4. **Skipped keys trusted blindly at commit** (1662-1674): a silent contract violation plus a cache-entry loss, not data loss. The snapshot is followed by RPC + `FinalizeEviction` + `WriteBucket` outside `mutex_`; skipped keys aren't in `pending_write_keys_` (2831-2832) and `SelectEvictionCandidate` has no key guard, so the old bucket can be evicted mid-window (the carry-over path 2320-2330 re-emits pooled keys without an `IsExist` recheck) while the batch returns success. Two cushions keep the harm small: (a) per item 1 the source DRAM replica is retained, so reads should still resolve to the source replica until the TTL reaper; (b) on restart `Init` re-`emplace`s all keys (2078-2086) and the `.meta` carries skipped keys, so the key re-enters the index pointing at intact bytes. What remains is a silent success that monitoring can't see, where the old code failed loudly. I won't block on this, but it's cheap:

```cpp
            if (skipped_keys.find(bucket_keys[i]) != skipped_keys.end() &&
                object_bucket_map_.find(bucket_keys[i]) !=
                    object_bucket_map_.end()) {
                continue;  // Still persisted elsewhere: idempotent skip.
            }
            // Otherwise fall through: evicted since the snapshot, commit it.
```

5. **`RestorePreparedEvictionLocked` doesn't mirror matched-only prepare** (3012-3019): it restores every key unconditionally, re-pointing skipped duplicates away from their authoritative bucket and double-charging `total_size_`. This is the flow covered by the Sept 13 note above. Low-frequency error path, small fix: restore only keys in `pending.keys` (keep `+= bucket_meta->meta_size` unconditional).

Small: `PendingEviction::keys` header comment ("All keys in evicted buckets") no longer matches the semantics; the `low_watermark_ratio` 0.25→0.50 change in `BucketWatermarkEvictionUsesHandlerAndKeepsNewest` passes at 0.25 too and drops multi-bucket coverage, so maybe revert unless there's a reason I'm missing.

To sum up: items 2, 3, and 5 are the July fixes re-applied (~15 lines in total), item 4 is encouraged but I won't block on it, and item 1 can be tracked as a follow-up. Happy to take another pass once those land.

he-yufeng:
@LujhCoconut Thanks for the careful pass, and for the correction first: you are right that `RestorePreparedEvictionLocked` and the quota guard were never removed — they are still on this head (2998-3029, 2979-2982). My Sept 13 note misread the tree there; sorry for the noise. Items 2, 3, 4, 5 and the two smalls are all landed in 75810f1c, in the exact shape you sketched, each verified line by line against the head before applying:

1. **`total_size_` commit credit** now sums only committed keys' data+key plus the shared `meta_size`, so commit and eviction/rollback/Init debit and credit the same matched set; the one-way leak is gone (and no longer needs Init-side compensation to stay balanced).
2. **`pending_write_size_` booking** is gated on `write_keys` being non-empty, so an all-duplicate batch no longer books phantom bytes that can evict real buckets.
3. **Skipped-trust at commit** falls through when the key is no longer in `object_bucket_map_` (evicted since the snapshot) and commits it, instead of returning silent success over a lost entry; the still-persisted case keeps the idempotent skip.
4. **`RestorePreparedEvictionLocked`** restores only keys in `pending.keys` (matched-only mirror of prepare) while keeping the unconditional `meta_size` credit, so skipped duplicates stay pointed at their authoritative bucket and `total_size_` is not double-charged.
5. Smalls: `PendingEviction::keys` header comment now describes the matched set; the watermark test keeps its multi-bucket coverage at 0.25 (the 0.50 bump had no reason).

Verification: pre-commit (clang-format included) is clean. The store suite runs in CI; I have no local C++ build on this host, so if CI surfaces anything I will fix forward.

On item 1 (skipped keys never reaching the Master, source-replica refcounts lingering until the TTL reaper): agreed it is cross-layer and should top the follow-up list. I will file it as a tracked issue with your `BatchOffload` skipped-set + NACK -1 sketch as the starting shape and link it back here.

he-yufeng:
Filed #4097 for item 1 with the cross-layer shape (skipped set surfaced from BatchOffload + NACK with the -1 sentinel) and a link back here.

LujhCoconut:
Thanks for the quick turnaround — I read 75810f1c line by line and the four fixes plus the two smalls match the suggestions exactly, no concerns from my side on the content. It looks like the commit hasn't been pushed yet, though: the PR head and your branch are both still at 0a3596cd. Once it's pushed and the store suite is green in CI, I'll do the final pass.

he-yufeng:
@LujhCoconut Pushed now — the head is 75810f1c, byte-identical to what you reviewed (the commit had only sat on my local rebased branch, which is on me). CI is running; store suite result will be here once the gate lets it through.

he-yufeng:
@LujhCoconut CI turned up one fallout of the accounting fix you'd asked for, and it's the test's premise, not the code. `BucketEvictionDoesNotReportSkippedDuplicateKeys` was written against the old full-bucket crediting: two batches of 6K booked 24K+meta, which cleared the 0.50 watermark and let the eviction fire. Your matched/matched accounting books only committed keys (12K+meta), so the same watermark never crosses and the eviction returned empty.

b69a8c8c re-tunes the fixture's ratio to 0.30 and documents why, so the test exercises the reporting assertions again (keyA skipped, keyB reported) on the accounting you specified. Full storage suite is 93/93 locally after the change. The remaining red on the last run was `ShmTransportTest.RelocateCopySurvivesConcurrentCap`, which passes locally three times in a row on this tree and doesn't intersect the diff, and the wheel jobs, which died with exit 137 (runner OOM) before reaching the build.

LujhCoconut (approved):
LGTM. Thanks for this contribution.

LujhCoconut:
I think this PR is ready to merge @ykwd. The CI failure is unrelated to this PR and has been fixed in PR #4131 .

ykwd (approved):

PR #3141, opened 2026-07-27
[Bugfix][Store] Fix recovery-related eviction behavior

### LINKED ISSUE #3139: [Bug][Store]: Recovery-related false eviction triggers and stale resource state
### Bug Report

**Description**

During HA recovery and segment remount, several conditions can cause unnecessary eviction or incorrect memory availability:

- `PutStart` may trigger memory eviction when too few memory segments are mounted and eviction cannot help the allocation.
- Restored memory replicas may become readable without receiving a new lease, causing recently recovered objects to be evicted prematurely.
- Temporary address gaps created during offset allocator recovery may remain reserved instead of being returned to available capacity.
- `BatchEvict` may continue processing objects after an OpLog reservation failure, even though subsequent reservations are likely to fail for the same backpressure condition.

Expected behavior:

- Trigger memory eviction only when it can recover capacity for the failed allocation.
- Grant a new lease when remount changes an object from having no readable memory replica to having one.
- Release recovery-only address gaps after allocator restoration.
- Stop the current eviction cycle on OpLog reservation failure and retry in a later cycle.

### Before submitting...

Icedcoco:
## Description

Fixes #3139.

This PR addresses several recovery-related allocation and eviction issues:

- Release temporary address gaps after restoring an offset allocator so they return to available capacity and are not included in allocation metrics.
- Avoid arming memory eviction when an allocation failure cannot be resolved by eviction, such as when too few memory segments are mounted.
- Grant a new lease when segment remount changes an object from having no readable memory replica to having one.
- Stop the current `BatchEvict` cycle when OpLog reservation fails, retain the eviction trigger, and retry in a later cycle.
- Distinguish OpLog submission outcomes from durability using explicit submission result states.

Deferred eviction policy changes, including stricter eviction limits and replica-level evictability rules, are outside the scope of this PR.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**

```bash
cmake --build build --target buffer_allocator_test master_service_ha_test -j32

./build/mooncake-store/tests/buffer_allocator_test

  ./build/mooncake-store/tests/master_service_ha_test \
  --gtest_filter='MasterServiceHATest.PutStartWithoutWritableSegmentsDoesNotArmMemoryEviction:MasterServiceHATest.PutStartWithTooFewSegmentsDoesNotArmMemoryEviction:MasterServiceHATest.PutStartWithReclaimableUsageArmsMemoryEviction:MasterServiceHATest.RemountMakesRestoredMemoryReplicaReady:MasterServiceHATest.FailedRemountKeepsReplicaInvalidAndCanBeRetried:MasterServiceHATest.BatchEvictStopsAfterFirstOpLogReservationFailure'

pre-commit run --files \
  mooncake-store/include/allocator.h \
  mooncake-store/src/allocator.cpp \
  mooncake-store/tests/buffer_allocator_test.cpp \
  mooncake-store/tests/ha/master_service_ha_test.cpp
```

**Test results:**

- [x] Unit tests pass
- [ ] Manual testing done

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex was used for code and log analysis, implementation assistance, and test execution. This PR has also been reviewed by Codex and other AI models.


mooncake-store/include/allocator.h
```diff
@@ -301,9 +301,6 @@ class OffsetBufferAllocator
     // offset allocator implementation
     std::shared_ptr<offset_allocator::OffsetAllocator> offset_allocator_;
 
-    // Keeps address gaps occupied after descriptor-based reconstruction.
-    std::vector<std::unique_ptr<AllocatedBuffer>> restored_gap_buffers_;
-
     friend class Serializer<OffsetBufferAllocator>;
     friend struct RestoredOffsetBufferAllocator;
     friend std::optional<struct RestoredOffsetBufferAllocator>
```
00fish0: nit: Since `allocator->restored_gap_buffers_ = std::move(gaps);` is deleted, should we delete this?

mooncake-store/src/master_service.cpp
```diff
@@ -1123,6 +1138,16 @@ auto MasterService::ReMountSegment(const std::vector<Segment>& segments,
...
                 standby_allocator_keepalive_.erase(restore.segment.name);
             }
+            for (const auto& [metadata, was_readable] : restored_objects) {
+                if (!was_readable &&
+                    metadata->HasReplica([this](const Replica& replica) {
+                        return replica.is_memory_replica() &&
+                               IsReplicaReadable(replica);
+                    })) {
+                    metadata->GrantLease(default_kv_lease_ttl_,
+                                         default_kv_soft_pin_ttl_);
```
00fish0: nit: It looks like this only renews the hard lease. 

Snapshot-restored metadata is created with soft pin disabled. Should the comment/implementation be updated to clarify this?

mooncake-store/src/master_service.cpp
00fish0, line 1: There still seems to be an eviction window between snapshot restore and segment remount: restored replicas have an expired lease but are not readable yet, so they may be evicted before remount can recover them. Should recovering replicas be excluded from eviction until remount succeeds or recovery times out?

```
T0: Snapshot restore
    Replica is COMPLETE, but the allocator/buffer is not restored.
    Its lease is already expired.

T1: Eviction scan
    The replica appears COMPLETE with an expired lease,
    so it may be marked REMOVED.

T2: Client remount
    The buffer should be restored and a new lease granted,
    but the replica may already have been evicted.
```

mooncake-store/include/allocator.h
00fish0, line 1: Follow-up idea: could we reuse the recovery primitives from #2932 and add a descriptor-based fixed-address import path for OffsetAllocator? This would rebuild the final live allocations directly, avoiding the first-fit filler approach and its temporary accounting churn. 

I’d still keep the live descriptors as the source of truth, since a serialized allocator snapshot may become stale after OpLog replay.

mooncake-store/src/master_service.cpp
Icedcoco: Fixed in d4018f19. RestoreFromStandbySnapshot now grants metadata containing an unreadable memory replica one existing hard-lease interval. This closes the restore-to-remount eviction window without excluding the replica indefinitely: a successful remount renews the lease, while a failed or missing remount lets the lease expire and makes the replica evictable again. The added tests cover immediate BatchEvict protection, eviction after timeout, and the remount path.

mooncake-store/src/master_service.cpp
Icedcoco: Yes, RestoreFromStandbySnapshot creates the metadata with soft pin disabled, so this path only renews the hard lease. That behavior predates this PR and is not introduced by the recovery eviction changes, so I am keeping it out of this PR and recording the soft-pin semantics as follow-up work.

mooncake-store/include/allocator.h
Icedcoco: Removed the redundant friend function declaration in d4018f19. The RestoredOffsetBufferAllocator struct remains a friend, which is sufficient for the required private access.

mooncake-store/include/allocator.h
Icedcoco: For this new-node recovery path, we still need to create an allocator; the recovery primitives from #2932 do not cover descriptor-based allocator construction here. Reusing them through a fixed-address import path is a reasonable follow-up, with the replayed live descriptors remaining the source of truth, but I am keeping that refactor out of this PR.

Icedcoco:
I pushed `c2596af` to correct the recovery/eviction boundary introduced in this PR.

Previously, `RestoreFromStandbySnapshot()` granted a temporary normal lease to objects whose memory replicas had not been remounted. Once that lease expired, ordinary eviction could remove those replicas even though the segment had never become readable. For a large recovery, the TTL could be consumed before the segment was serviceable.

This commit changes the protection mechanism from an expiring object lease to replica eligibility:

- standby restore no longer grants a pre-remount lease;
- a memory replica is ordinarily evictable only when it is complete, unreferenced, and passes `IsReplicaReadable()`;
- the same predicate is used by `BatchEvict`, tenant quota eviction, offload source selection, and OpLog remaining-descriptor prediction/final mutation;
- the regression test now verifies that an unremounted replica remains protected indefinitely, successful remount grants a fresh lease, and normal eviction resumes after that lease expires.

This keeps the normal read path unchanged and ensures that recovery time does not consume the client-visible lease window.

The change aligns PR #3141 with the Phase 1 recovery invariant described in RFC #3321. It is intentionally scoped and does not implement the remaining RFC work, such as a separate recovery TTL, unconditional remount refresh, pin-state persistence, or replacing the current remount collection with a vector.

mooncake-store/src/master_service.cpp
```diff
@@ -1004,6 +1004,9 @@ auto MasterService::ReMountSegment(const std::vector<Segment>& segments,
 
             bool ambiguous_endpoint = false;
             bool unsupported_cxl = false;
+            // Track pre-remount readability so the first successful recovery
+            // starts a fresh lease without extending already-readable objects.
+            std::unordered_map<ObjectMetadata*, bool> restored_objects;
```
Aionw, line 1009: Could these pointers become invalid after the individual shard lock is released? `ReMountSegment()` keeps the global snapshot lock, but not every metadata mutation path follows that lock—for example, `NoFBatchEvict()` directly acquires shard locks and may move replicas or erase metadata. Please either make all such paths participate in `snapshot_mutex_`, or store stable object/replica identities and look them up again while holding the shard lock.

mooncake-store/src/master_service.cpp
```diff
@@ -3170,6 +3194,11 @@ bool MasterService::IsReplicaReadable(const Replica& replica) const {
     return !endpoint || !invalid_replica_endpoints_.contains(*endpoint);
 }
 
+bool MasterService::IsMemoryReplicaEvictable(const Replica& replica) const {
+    return replica.is_memory_replica() && replica.is_completed() &&
```
Aionw, line 3198: This makes eviction eligibility depend on derived runtime readability, which further overloads `COMPLETE`: the data may be complete while the replica is still unavailable because its segment has not been remounted. Would it be cleaner to model availability explicitly in metadata, for example as `RECOVERING` versus `AVAILABLE`, so reads and eviction use the same state machine? The design should also define expiry or cleanup for replicas that never remount; otherwise they remain non-evictable indefinitely.

mooncake-store/src/master_service.cpp
```diff
@@ -3657,7 +3700,9 @@ auto MasterService::AllocateAndInsertMetadata(
             }
             if (write_mode != ReplicaWriteMode::FLEXIBLE_DUAL_REPLICA) {
                 MasterMetricManager::instance().inc_put_start_alloc_failures();
-                need_mem_eviction_ = true;
+                if (memory_eviction_may_help) {
```
Aionw, line 3703: Do we need `memory_eviction_may_help` for this bug fix? Segment count plus `allocator->size() > 0` does not establish that eviction can make the allocation succeed: the request may exceed segment capacity, or all allocated objects may be pinned or leased. For the narrow bug fix, it would be simpler to suppress eviction only when there are fewer segments than the requested replica count. A complete solution should have the allocation strategy return a structured failure reason.

mooncake-store/src/master_service.cpp
```diff
@@ -8916,6 +8981,7 @@ void MasterService::BatchEvict(double evict_ratio_target,
                         deferred_replicas,
                         /*allow_soft_pinned=*/false);
 
+                    stop_eviction = evict_result.stop_cycle;
```
Aionw, line 8984: `stop_cycle` is currently set for every reservation failure, while the final handling assumes the failure is transient. Queue exhaustion (`TASK_PENDING_LIMIT_EXCEEDED`) is transient, but invalid configuration or a missing/stopped writer may not recover on retry. Could we preserve the actual error and distinguish transient backpressure from permanent failures? The reservation is admission to the in-memory ordered writer, which later batches records into an etcd transaction, so stopping on queue exhaustion is reasonable; treating every error the same is not.

mooncake-store/src/master_service.cpp
```diff
@@ -9107,7 +9183,19 @@ void MasterService::BatchEvict(double evict_ratio_target,
         }
     }
 
-    if (evicted_count > 0 || released_discarded_cnt > 0) {
+    if (stop_eviction) {
```
Aionw, line 9186: The global and memory-specific metrics are intentionally different, but this branch duplicates the same success/failure accounting used by the following branches. Could we determine `need_mem_eviction_` and whether the cycle made progress first, then record the total and memory metrics in one common block? That would make the only meaningful distinction here—retry versus completion—clearer.

Aionw (commented):
This PR combines four independently reviewable fixes that have limited implementation coupling: allocator recovery gaps, `PutStart` eviction triggering, replica remount/lease handling, and OpLog backpressure. Could we split them into focused PRs? That would make correctness review, rollback, and regression diagnosis significantly easier.

PR #2383, opened 2026-06-09
[Store] recover etcd client after leader hang (SIGSTOP)

Icedcoco:
## Description

Issue: kvcache-ai/Mooncake#2253

When the current etcd leader is suspended with SIGSTOP (kill -19), the TCP connection remains half-open. Mooncake master detects the lease keepalive failure and stops serving, but recovery attempts reuse the same stale storeClient, causing GrantLease/CreateWithLease to timeout with 'context deadline exceeded'.

Changes:

- Go wrapper: configure gRPC keepalive (10s/3s/PermitWithoutStream) to detect half-open connections faster; add synchronized getStoreClient() accessor; add EtcdStoreResetClientWrapper that cancels all active keepalive/watch contexts, creates a fresh clientv3.Client, and closes the old one.

- Go wrapper (review fix): cancelAllStorePrefixWatches() no longer deletes map entries early; let goroutine defer clean up to avoid race with new watch registration. Watch goroutine now sends WATCH_BROKEN on all ctx.Done() paths, allowing C++ watchers to detect reset and reconnect.

- C++ EtcdHelper: add ResetEtcdStoreClient() API.

- C++ HA coordinator: trigger ResetConnection() after GrantLease or CreateWithLease failures in TryAcquireLeadership, and after ETCD_OPERATION_ERROR in the keepalive thread.

- C++ HA coordinator (review fix): on CreateWithLease non-transaction failure, reset the etcd client FIRST before attempting RevokeLease. This ensures the stale client is always replaced even if RevokeLease also fails on the same broken connection.

- Tests: add gtest cases verifying basic operations, keepalive termination, and prefix watch broken/reconnect after reset. Add e2e bash script for 3-node etcd SIGSTOP and SIGKILL regression testing.

## Module

- [ ] Transfer Engine (`mooncake-transfer-engine`)
- [x] Mooncake Store (`mooncake-store`)
- [ ] Mooncake EP (`mooncake-ep`)
- [ ] Integration (`mooncake-integration`)
- [ ] P2P Store (`mooncake-p2p-store`)
- [ ] Python Wheel (`mooncake-wheel`)
- [ ] PyTorch Backend (`mooncake-pg`)
- [ ] Mooncake RL (`mooncake-rl`)
- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other

[...]

PR #3144, opened 2026-07-27
[Bugfix][Common] Avoid etcd keepalive pings without active streams

### LINKED ISSUE #3142: [Bug]: HA etcd clients repeatedly receive too_many_pings GOAWAY since v0.3.12
### Summary

In prefill/decode disaggregated deployments using etcd as the HA backend, Mooncake clients repeatedly log:

```text
ERROR: [transport] Client received GoAway with error code ENHANCE_YOUR_CALM and debug data equal to ASCII "too_many_pings".
```

The issue is reproducible with Mooncake `v0.3.12` and later. It was not observed with `v0.3.11.post1`. The apparent regression point is #2383.

The errors are usually more frequent shortly after startup and become less frequent over time as gRPC-Go increases the keepalive interval after each GOAWAY.

### Environment

- Mooncake: `v0.3.12` and later
- Deployment: prefill/decode disaggregation
- HA backend: etcd
- etcd client: `go.etcd.io/etcd/client/v3 v3.5.21`
- gRPC-Go: `google.golang.org/grpc v1.79.3`
- Reproduced with etcd `v3.3.25`
- Multiple etcd endpoints make the issue easier to reproduce

### Expected Behavior

Idle etcd connections should not be repeatedly closed and re-established during normal HA operation.

### Actual Behavior

etcd periodically closes idle gRPC connections with:

```text
ENHANCE_YOUR_CALM: too_many_pings
```

gRPC-Go then reconnects the affected SubChannel automatically. Mooncake continues running, but the connection churn produces error logs and unnecessary control-plane activity.

Setting the following etcd option does not resolve the issue:

```bash
--grpc-keepalive-min-time=1s
```

### Reproduction

1. Start etcd with its normal client endpoint configuration. The issue is still reproducible if `--grpc-keepalive-min-time=1s` is added.
2. Start Mooncake master in HA mode with etcd as the HA backend.
3. Start a Mooncake client using the HA backend address.
4. Leave the deployment running with little or no control-plane activity.
5. Observe repeated `too_many_pings` errors in the client logs.

To confirm the affected endpoint, enable gRPC-Go logging before starting the client:

```bash
GRPC_GO_LOG_SEVERITY_LEVEL=info \
GRPC_GO_LOG_VERBOSITY_LEVEL=2 \
<mooncake client startup command>
```

Relevant logs can be filtered with:

```bash
rg 'original dial target|Subchannel picks|too_many_pings|received goaway|Connectivity change'
```

The resulting sequence shows that the affected channel targets an etcd endpoint:

```text
Channel created with target etcd-endpoints://.../<etcd-ip>:2379
Subchannel picks address "<etcd-ip>:2379"
Client received GoAway ... "too_many_pings"
SubChannel: READY -> IDLE
Closing: "received goaway and there are no active streams"
SubChannel: IDLE -> CONNECTING -> READY
```

### Root Cause

PR #2383 added transport-level gRPC keepalive to the Store etcd client to detect half-open connections after an etcd leader is suspended with `SIGSTOP`:

```go
return clientv3.Config{
    Endpoints:            validEndpoints,
    DialTimeout:          5 * time.Second,
    DialKeepAliveTime:    10 * time.Second,
    DialKeepAliveTimeout: 3 * time.Second,
    PermitWithoutStream:  true,
}
```

The problematic option is:

```go
PermitWithoutStream: true
```

It instructs gRPC-Go to send HTTP/2 PING frames even when the underlying transport has no active RPC streams.

etcd configures its gRPC enforcement policy with:

```go
keepalive.EnforcementPolicy{
    MinTime:             configuredMinTime,
    PermitWithoutStream: false,
}
```

When a transport has an active RPC stream, etcd validates client PING intervals against `--grpc-keepalive-min-time`.

When a transport has no active streams, etcd does not use that configured minimum. Because `PermitWithoutStream` is false on the server, repeated idle PINGs are treated as policy violations. After enough violations, etcd sends:

```text
GOAWAY
error code: ENHANCE_YOUR_CALM
debug data: too_many_pings
```

This also explains why lowering `--grpc-keepalive-min-time` does not fix the problem.

The issue is more visible with multiple endpoints because clientv3 uses `round_robin` and creates a SubChannel for each endpoint. A watch or lease stream may be active on one transport while the remaining transports are idle but still send PINGs.

After receiving `too_many_pings`, gRPC-Go closes the affected transport, reconnects it, and increases the keepalive interval for future connections. This explains why the errors become less frequent over time.

### Impact

No process crashes, data corruption, or direct interruption of active data-path operations have been observed.

The GOAWAY normally targets a transport with no active RPC streams, so gRPC-Go can reconnect it automatically. However, the issue still causes:

- Repeated TCP/gRPC connection establishment
- Additional client and etcd CPU/network overhead
- Noisy error logs
- Potential connection storms in large multi-rank deployments
- Possible transient latency for new etcd operations during reconnection
- Possible delay in HA control-plane operations if reconnection overlaps with an etcd or master failover

The connection carrying an active leadership lease keepalive or master-view watch is normally not subject to the idle-stream enforcement path.

### Proposed Fix

Remove `PermitWithoutStream`, or explicitly set it to `false`:

```go
return clientv3.Config{
    Endpoints:            validEndpoints,
    DialTimeout:          5 * time.Second,
    DialKeepAliveTime:    storeDialKeepAliveTime,
    DialKeepAliveTimeout: storeDialKeepAliveTimeout,
    PermitWithoutStream:  false,
}
```

`DialKeepAliveTime` and `DialKeepAliveTimeout` should remain unchanged.

This should preserve the purpose of #2383. Leadership lease keepalive and watch operations use active RPC streams, so transport keepalive can still detect a half-open connection when the etcd leader is suspended. Only PINGs on completely idle transports are disabled.

### Validation

The fix should be verified with:

1. An idle HA client connected to multiple etcd endpoints, confirming that no `too_many_pings` GOAWAY is received.
2. The #2383 etcd leader `SIGSTOP`/`SIGKILL` regression test, confirming that the master can still recover.
3. Master-view watch recovery after an etcd connection interruption.
4. Leadership lease keepalive failure and client reset/reconnection.
5. A comparison of SubChannel reconnection counts before and after the change.

### Before submitting...

- [x] Searched for related issues and pull requests
- [x] Confirmed that `v0.3.12` contains the apparent regression
- [x] Reproduced the issue with a minimal client
- [x] Verified the cause with a `PermitWithoutStream=true/false` comparison

### Before submitting...

Icedcoco:
## Description

Fixes #3142.

Since #2383, the Store etcd client has configured gRPC keepalive with `PermitWithoutStream: true`. This allows idle transports to send HTTP/2 PING frames even when they have no active RPC streams.

etcd does not permit keepalive PINGs without active streams. After repeated idle PINGs, it closes the transport with:

```text
ENHANCE_YOUR_CALM: too_many_pings
```

This PR removes `PermitWithoutStream: true`, restoring its default value of `false`.

`DialKeepAliveTime` and `DialKeepAliveTimeout` remain unchanged. Connections with active lease or watch streams therefore retain the transport keepalive behavior introduced by #2383, while completely idle transports no longer send PINGs that violate the etcd server policy.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

The issue and fix were verified locally with an A/B test using:

- etcd `v3.3.25`
- `--grpc-keepalive-min-time=1s`
- gRPC-Go diagnostic logging:
  `GRPC_GO_LOG_SEVERITY_LEVEL=info`
  and `GRPC_GO_LOG_VERBOSITY_LEVEL=2`
- The actual `newStoreClientConfig` configuration
- One initial etcd `Get`, followed by 60 seconds with no RPC activity, and one final `Get`

Before the change, the client reported
`PermitWithoutStream=true`. After approximately 30 seconds of inactivity, the following error was reproduced:

```text
Client received GoAway with error code ENHANCE_YOUR_CALM and debug data equal to ASCII "too_many_pings".
```

The affected SubChannel transitioned from `READY` to `IDLE` and then reconnected. The final etcd `Get` succeeded after reconnection.

After the change, the same test reported `PermitWithoutStream=false`. During the full 60-second idle period:

- No `GOAWAY` was received
- No `too_many_pings` error was logged
- The SubChannel remained `READY`
- The final etcd `Get` succeeded

This before/after comparison changed only `PermitWithoutStream`, confirming that the change resolves the issue reported in #3142. The temporary reproduction test was used for local validation and is not included in this PR because it requires a real etcd server and a time-based 60-second wait.

**Test commands:**

```bash
GRPC_GO_LOG_SEVERITY_LEVEL=info \
GRPC_GO_LOG_VERBOSITY_LEVEL=2 \
go test -run '^TestIdleStoreClientKeepalive$' -count=1 -v -timeout=90s

pre-commit run --files mooncake-common/etcd/etcd_wrapper.go
git diff --check
```

**Test results:**

- [ ] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex assisted with root-cause analysis, the local before/after reproduction, and drafting this PR description. The submitted change has been reviewed by the human contributor.

### DIFF
--- a/mooncake-common/etcd/etcd_wrapper.go
+++ b/mooncake-common/etcd/etcd_wrapper.go
@@ -97,7 +97,6 @@ func newStoreClientConfig(validEndpoints []string) clientv3.Config {
 		DialTimeout:          5 * time.Second,
 		DialKeepAliveTime:    storeDialKeepAliveTime,
 		DialKeepAliveTimeout: storeDialKeepAliveTimeout,
-		PermitWithoutStream:  true,
 	}
 }
 


he-yufeng:
Verified the failure semantics this has to keep intact: the #2383 SIGSTOP scenario still gets its half-open detection, because the lease KeepAlive call is a long-lived bidi stream on the same transport, so DialKeepAliveTime pings keep flowing whenever a lease or watch is active. The flag being removed only governed fully idle transports, and there it was working against #2383's own goal: etcd answers idle pings with too_many_pings and closes the transport, which kills the warm connection the keepalive was meant to preserve.

The two sibling clients in this file (the KV client and the snapshot client) never set keepalive params at all and have been fine, so even after this change the store client remains the most defensive of the three.

The change itself looks right to me. Leaving the formal approve to a second pair of eyes since it touches failure-detection timing on the metadata path.

yokinoshitayoki (commented):
Thanks a lot for your contribution. The issue addressed by this PR is real, and removing `PermitWithoutStream: true` looks like the correct fix for the repeated `too_many_pings` GOAWAYs. Perhaps we could run the #2383 leader `SIGSTOP`/`SIGKILL` E2E test and report the recovery behavior, so we can understand whether disabling keepalive on fully idle transports has any measurable impact on the failure-detection path introduced there.

Icedcoco:
> Thanks a lot for your contribution. The issue addressed by this PR is real, and removing `PermitWithoutStream: true` looks like the correct fix for the repeated `too_many_pings` GOAWAYs. Perhaps we could run the #2383 leader `SIGSTOP`/`SIGKILL` E2E test and report the recovery behavior, so we can understand whether disabling keepalive on fully idle transports has any measurable impact on the failure-detection path introduced there.

I ran the original #2383 `etcd_leader_hang_e2e.sh` against a local three-node etcd cluster, and also did an A/B timing comparison using the same checkout and test environment:

| Store client config | `SIGSTOP` to `master_view` recovery | `SIGKILL` to `master_view` recovery | Result |
| --- | ---: | ---: | --- |
| Before this PR (`PermitWithoutStream=true`) | ~18 s | ~12 s | Passed |
| This PR (`PermitWithoutStream=false`) | ~16 s | ~12 s | Passed |

Both runs completed the full `SIGSTOP -> recover -> SIGKILL -> recover` cycle and exited successfully with `etcd leader hang e2e passed`.

The timing was collected with one-second Bash trace timestamps. It includes the script's endpoint health checks and one-second `master_view` polling, so I would not interpret the two-second difference in the `SIGSTOP` result as an improvement. The relevant result is that this run showed no measurable recovery regression after disabling keepalive on fully idle transports.

This is also consistent with the expected path: the leadership lease uses an active long-lived `KeepAlive` stream, so `DialKeepAliveTime=10s` and `DialKeepAliveTimeout=3s` still apply there. Removing `PermitWithoutStream` only changes transports with no active RPC streams.

I passed `MASTER_BIN` explicitly when running the script because its current default `ROOT_DIR` calculation resolves one directory above this checkout.

he-yufeng:
This is exactly the evidence the change needed, thanks for running it. SIGSTOP at ~16s vs ~18s and SIGKILL at 12s vs 12s is within script noise, and the mechanism explains why: the lease keepalive rides its own long-lived stream, so it keeps the 10s/3s detection either way. My static read of the same conclusion is now backed by your E2E, and the too_many_pings pressure goes away. Looks good to land from where I sit; leaving the formal approve to a second maintainer as usual on the metadata path.

yokinoshitayoki (approved):
Thanks a lot for your contribution. This PR looks good to me.

PR #3145, opened 2026-07-27
[TENT] Harden ShmTransport backing-file and mapping lifecycle

alexps9:
## Description

This PR hardens the TENT `ShmTransport` shared-memory lifecycle used for same-host DRAM transfers.

Currentyly several edge cases can turn into hard crashes (`SIGBUS`) or silent use of stale peer mappings:

1. **Consumer CXL open used `O_CREAT`.** If the peer backing file was missing, the consumer created a 0-byte file and `mmap`'d the expected length. Access beyond EOF then SIGBUS'd instead of returning a clean error. The POSIX `shm_open` consumer path already omitted `O_CREAT`; this aligns the CXL path with that behavior.
2. **Producer create lacked `O_EXCL`.** `shm_open`/`open` + `ftruncate` could reopen an existing object and truncate it while another process still held a mapping. Create now uses `O_CREAT|O_EXCL`, retries with a new pid-qualified name on `EEXIST`, and cleans up on truncate/mmap failure.
3. **`relocate_map_` never dropped stale mmaps.** Entries were only cleared on `uninstall()`. After peer unregister / buffer removal, a later `NeedsRefreshCache` miss now `munmap`s and erases that segment's cached mappings, limiting unbounded growth and stale use.
4. **Sample config cleanup.** Remove unused `transports/shm/async_memcpy_threshold` from `tent/config/transfer-engine.json` (never read by `ShmTransport`). Clarify in `transport_loader.cpp` that SHM remains opt-in (`enable` defaults to `false`).

This is a correctness fix for the existing TENT SHM transport, guided by [Ray Plasma](https://github.com/ray-project/plasma)’s shared-memory object lifecycle model.

**Files:** `shm_transport.{h,cpp}`, `shm_transport_test.cpp`, `transport_loader.cpp`, `transfer-engine.json` (~300 LOC).

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**

```bash
# Build TENT + unit test
cmake .. -DUSE_TENT=ON -DBUILD_UNIT_TESTS=ON
cmake --build . -j --target tent_shm_transport_test tebench

# Unit tests (covers B1/B2/B3 + existing concurrency test)
./mooncake-transfer-engine/tent/tests/tent_shm_transport_test

# Same-host microbench: start target, then initiator
./tebench --backend=tent --seg_type=DRAM --xport_type=shm \
  --total_buffer_size=67108864
./tebench --backend=tent --seg_type=DRAM --xport_type=shm \
  --target_seg_name=<SEG_FROM_TARGET> --op_type=read --duration=3 \
  --start_block_size=4096 --max_block_size=1048576 \
  --start_batch_size=1 --max_batch_size=1 \
  --start_num_threads=1 --max_num_threads=1
# Compare with --xport_type=tcp
```

**New / updated unit coverage:**

- `CxlConsumerDoesNotCreateMissingFile` — missing CXL backing file fails; no file is created
- `CreateSharedMemoryDoesNotTruncateExisting` — `O_EXCL` preserves existing content; allocate retries with a new name
- `DropsStaleMappingsOnNeedsRefreshCache` — `NeedsRefreshCache` drops segment mmap cache
- `DropSegmentMappingsReleasesMmap` — explicit invalidation unmaps pages

**Test results:**

- [x] Unit tests pass (`tent_shm_transport_test`: 5/5)

Same-host DRAM, 1 thread, batch=1, duration=3s (illustrative):

| Transport | 1 MiB BW | 1 MiB P99 Tx |
|-----------|----------|--------------|
| SHM (after fix) | ~44.7 GB/s | ~30 μs |
| TCP (after fix) | ~0.6 GB/s | ~3.2 ms |

SHM remains far faster than same-host TCP. Post-fix SHM is within normal run-to-run variance of the pre-fix baseline (~46 GB/s); no structural regression observed.

**Scope note:** Validation was same-host DRAM (SHM vs TCP). CXL consumer behavior is covered by unit test; no CXL hardware was available.

## Checklist

- [x] I have run `pre-commit` on the touched files and hooks pass
- [x] I have updated the documentation (if applicable) — N/A (bug fix; sample config only)
- [x] For changes >500 LOC: I have filed an RFC issue — N/A (~300 LOC)

## AI Assistance Disclosure

- [ ] No AI tools were used

Claudecode (Opus 5) was used to help locate the SHM lifecycle issues. Grok 4.5 was used to draft and refine this PR description. 
The human submitter reviewed and can defend every changed line.

### DIFF
--- a/mooncake-transfer-engine/tent/src/transport/shm/shm_transport.cpp
+++ b/mooncake-transfer-engine/tent/src/transport/shm/shm_transport.cpp
@@ -280,18 +280,6 @@ bool ShmTransport::tryResolve(const RelocateMap &relocate_map,
     return false;
 }
 
-void ShmTransport::dropSegmentMappings(SegmentID target_id) {
-    RWSpinlock::WriteGuard guard(relocate_lock_);
-    auto target = relocate_map_.find(target_id);
-    if (target == relocate_map_.end()) {
-        return;
-    }
-    for (auto &entry : target->second) {
-        munmap(entry.second.shm_addr, entry.second.length);
-    }
-    relocate_map_.erase(target);
-}
-
 Status ShmTransport::relocateSharedMemoryAddress(uint64_t &dest_addr,
                                                  uint64_t length,
                                                  uint64_t target_id) {
@@ -319,30 +307,19 @@ Status ShmTransport::relocateSharedMemoryAddress(uint64_t &dest_addr,
     // Owning reference: `buffer` is used after the lambda returns.
     SegmentDescRef pin;
     auto &segment_manager = metadata_->segmentManager();
-    Status lookup_status = segment_manager.withCachedSegment(
+    // Do not munmap relocate_map_ entries on NeedsRefreshCache: memcpy runs
+    // after this lock is released, and there is no transfer-level refcount /
+    // quiesce that proves no reader still holds a resolved address. POSIX
+    // shm_unlink already keeps the object alive for existing mappings; keep
+    // those mappings until uninstall().
+    CHECK_STATUS(segment_manager.withCachedSegment(
         target_id, pin, [&](SegmentDesc *segment) {
             buffer = segment->findBuffer(dest_addr, length);
             if (!buffer || buffer->shm_path.empty())
                 return Status::NeedsRefreshCache(
                     "Requested address is not in registered buffer" LOC_MARK);
             return Status::OK();
-        });
-    if (lookup_status.IsNeedsRefreshCache()) {
-        // Peer may have unregistered/freed the buffer. Drop stale mmap entries
-        // so we neither grow relocate_map_ unboundedly nor read freed memory.
-        // Unlock is not needed: we already hold the write lock; erase inline.
-        auto stale = relocate_map_.find(target_id);
-        if (stale != relocate_map_.end()) {
-            for (auto &entry : stale->second) {
-                munmap(entry.second.shm_addr, entry.second.length);
-            }
-            relocate_map_.erase(stale);
-        }
-        return lookup_status;
-    }
-    if (!lookup_status.ok()) {
-        return lookup_status;
-    }
+        }));
 
     void *shm_addr = nullptr;
     bool mapping_found = false;
--- a/mooncake-transfer-engine/tent/src/transport/shm/shm_transport.cpp
+++ b/mooncake-transfer-engine/tent/src/transport/shm/shm_transport.cpp
@@ -19,6 +19,7 @@
 #include <fcntl.h>
 #include <glog/logging.h>
 #include <sys/mman.h>
+#include <sys/stat.h>
 #include <unistd.h>
 
 #include <algorithm>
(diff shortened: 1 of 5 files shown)


he-yufeng:
One lifecycle question on the munmap paths before this lands. relocateSharedMemoryAddress hands the mapped address to the caller under the write lock, but the actual memcpy runs after the guard is released, so nothing serializes a copy against the drop. A thread can resolve an address from relocate_map_, start copying, and a concurrent NeedsRefreshCache (or a future dropSegmentMappings call) munmaps that same address mid-copy: SIGBUS on the reader. POSIX shm unlink keeps the object alive for existing mappings, so simply holding the mapping is the safer default until there is a transfer-level guarantee (refcount or quiesce on the segment) that no reader is in flight. Does that guarantee exist somewhere I missed? If not, the inline drop in relocate and the public dropSegmentMappings should probably gate on it.

Related: dropSegmentMappings is public but has no caller in this PR, and its "safe to call when the peer has unregistered" contract is exactly the condition the engine cannot currently prove per call. Either wire it to the unregister hook with that guarantee or keep it private for now.

The O_EXCL create plus no-O_CREAT on the consumer side is independently correct (a consumer creating a 0-byte object and then faulting on it was a real trap) and could ship on its own if the munmap part needs more work.

alexps9:
> One lifecycle question on the munmap paths before this lands. relocateSharedMemoryAddress hands the mapped address to the caller under the write lock, but the actual memcpy runs after the guard is released, so nothing serializes a copy against the drop. A thread can resolve an address from relocate_map_, start copying, and a concurrent NeedsRefreshCache (or a future dropSegmentMappings call) munmaps that same address mid-copy: SIGBUS on the reader. POSIX shm unlink keeps the object alive for existing mappings, so simply holding the mapping is the safer default until there is a transfer-level guarantee (refcount or quiesce on the segment) that no reader is in flight. Does that guarantee exist somewhere I missed? If not, the inline drop in relocate and the public dropSegmentMappings should probably gate on it.
> 
> Related: dropSegmentMappings is public but has no caller in this PR, and its "safe to call when the peer has unregistered" contract is exactly the condition the engine cannot currently prove per call. Either wire it to the unregister hook with that guarantee or keep it private for now.
> 
> The O_EXCL create plus no-O_CREAT on the consumer side is independently correct (a consumer creating a 0-byte object and then faulting on it was a real trap) and could ship on its own if the munmap part needs more work.

Thanks for your constructive advice. There is currently no transfer-level refcount or segment-quiescence guarantee. `relocateSharedMemoryAddress()` resolves and returns the mapped address while holding the relocate lock, but the subsequent `memcpy` runs after that lock is released. As a result, `NeedsRefreshCache`—or a public drop helper—could `munmap` the address while a reader is still in flight, causing a `SIGBUS`.

Since `shm_unlink` already preserves the object for existing mappings, retaining those mappings until `uninstall()` is the safer default.

I have removed both the inline `munmap` and the public `dropSegmentMappings` helper in `e251f89`. The PR now contains only the independent create and open fixes:
- `O_CREAT | O_EXCL` on the producer 
- no `O_CREAT` on the consumer

I also add a regression test verifying that `NeedsRefreshCache` retains existing mappings.

Eager reclamation can be added later once we have an explicit transfer-level lifetime guarantee, such as reference counting or segment quiescence.

mooncake-transfer-engine/tent/src/transport/shm/shm_transport.cpp
alogfans, line 374: Removing `O_CREAT` prevents creating a new zero-length file, but an existing stale or truncated backing file can still be shorter than `buffer->length`. `mmap()` may succeed and later raise `SIGBUS` on access beyond EOF. Please use `fstat()` to verify that the backing file is at least `buffer->length` before mapping it.

he-yufeng:
e251f89 matches exactly what I had in mind: munmap paths gone, mappings retained until uninstall, and the regression test pins the NeedsRefreshCache case so a future eager-reclamation attempt has to break it loudly. The remaining surface (O_EXCL create with pid in the name, no O_CREAT on the consumer, cleanup on ftruncate/mmap failure) is the part that was independently correct from the start. This shape looks good to me; leaving the formal approve to the TENT owners.

mooncake-transfer-engine/tent/src/transport/shm/shm_transport.cpp
alexps9: Fixed in 68ea2732.

staryxchen (approved):
LGTM

PR #3151, opened 2026-07-28
[Bugfix][TE] Respect remote_accessible in RDMA verbs MR registration

RuiqingFeng:
## Description

`RdmaTransport::registerLocalMemoryInternal()` previously discarded the `remote_accessible` argument (`(void)remote_accessible;`) and unconditionally granted remote read/write permissions (`IBV_ACCESS_REMOTE_READ | IBV_ACCESS_REMOTE_WRITE`). This violated the API contract and exposed local-only memory to remote peers.

This PR fixes the issue by:
1. Setting `IBV_ACCESS_REMOTE_*` permissions only when `remote_accessible` is true. When `false`, registers with `IBV_ACCESS_LOCAL_WRITE` only and leaves the `rkey` vector empty.
2. Filtering out RDMA buffers with empty `rkey` vectors during metadata encoding so local-only memory is not advertised across the network.
3. Adding bounds checks in `selectPeerDevice()` for `buffer_id` and `device_id` against the `rkey` array size to prevent out-of-bounds indexing on local-only memory.

## Module

- [ ] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [x] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

Added unit tests in `rdma_context_reprobe_test.cpp` and `transfer_metadata_test.cpp`:
1. `RdmaMemoryRegistrationPolicyTest.LocalOnlyBufferHasNoPublishedRkey`: Verifies that registering local memory with `remote_accessible=false` keeps its `rkey` empty.
2. `TransferMetadataPublicationTest.OmitsBufferWithoutRkey`: Verifies that buffers with empty `rkey`s are excluded when encoding P2P segment descriptions for network transfer.

**Test commands:**
```bash
cmake --build build-ci-local --target rdma_context_reprobe_test transfer_metadata_test
./build-ci-local/mooncake-transfer-engine/tests/rdma_context_reprobe_test --gtest_filter='*RegistrationPolicy*'
./build-ci-local/mooncake-transfer-engine/tests/transfer_metadata_test --gtest_filter='TransferMetadataPublicationTest.*'

### DIFF
--- a/mooncake-transfer-engine/src/transport/rdma_transport/rdma_transport.cpp
+++ b/mooncake-transfer-engine/src/transport/rdma_transport/rdma_transport.cpp
@@ -214,7 +214,6 @@ int RdmaTransport::registerLocalMemoryInternal(void *addr, size_t length,
                                                bool remote_accessible,
                                                bool update_metadata,
                                                bool force_sequential) {
-    BufferDesc buffer_desc;
     int access_rights = IBV_ACCESS_LOCAL_WRITE;
     if (remote_accessible)
         access_rights |= IBV_ACCESS_REMOTE_WRITE | IBV_ACCESS_REMOTE_READ;
--- a/mooncake-transfer-engine/src/transport/rdma_transport/rdma_transport.cpp
+++ b/mooncake-transfer-engine/src/transport/rdma_transport/rdma_transport.cpp
@@ -214,12 +214,10 @@ int RdmaTransport::registerLocalMemoryInternal(void *addr, size_t length,
                                                bool remote_accessible,
                                                bool update_metadata,
                                                bool force_sequential) {
-    (void)remote_accessible;
-    const int kBaseAccessRights = IBV_ACCESS_LOCAL_WRITE |
-                                  IBV_ACCESS_REMOTE_WRITE |
-                                  IBV_ACCESS_REMOTE_READ;
-
-    int access_rights = kBaseAccessRights;
+    BufferDesc buffer_desc;
+    int access_rights = IBV_ACCESS_LOCAL_WRITE;
+    if (remote_accessible)
+        access_rights |= IBV_ACCESS_REMOTE_WRITE | IBV_ACCESS_REMOTE_READ;
     if (MCIbRelaxedOrderingEnabled) {
         access_rights |= IBV_ACCESS_RELAXED_ORDERING;
     }
@@ -416,7 +414,8 @@ int RdmaTransport::registerLocalMemoryInternal(void *addr, size_t length,
         BufferDesc buffer_desc;
         for (auto &context : context_list_) {
             buffer_desc.lkey.push_back(context->lkey(chunk_addr));
-            buffer_desc.rkey.push_back(context->rkey(chunk_addr));
+            if (remote_accessible)
+                buffer_desc.rkey.push_back(context->rkey(chunk_addr));
         }
         buffer_desc.name = resolved_name;
         buffer_desc.addr = (uint64_t)chunk_addr;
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: mooncake-transfer-engine/src/transport/rdma_transport/rdma_transport.cpp [hunk 1]
```diff
@@ -214,12 +214,10 @@ int RdmaTransport::registerLocalMemoryInternal(void *addr, size_t length,
                                                bool remote_accessible,
                                                bool update_metadata,
                                                bool force_sequential) {
-    (void)remote_accessible;
-    const int kBaseAccessRights = IBV_ACCESS_LOCAL_WRITE |
-                                  IBV_ACCESS_REMOTE_WRITE |
-                                  IBV_ACCESS_REMOTE_READ;
-
-    int access_rights = kBaseAccessRights;
+    BufferDesc buffer_desc;
```
staryxchen: this is no longer needed.

staryxchen (commented):
1. Protocol coverage — Metadata filtering only skips empty rkey when protocol == "rdma", while barex / efa / cxi share the same encode path but are not filtered. EFA/CXI also still ignore remote_accessible. Is this PR intentionally RDMA-only?
2. PR title — The title looks concatenated (...registrationFix/rdma...); worth cleaning up before merge.

RuiqingFeng:
> 1. Protocol coverage — Metadata filtering only skips empty rkey when protocol == "rdma", while barex / efa / cxi share the same encode path but are not filtered. EFA/CXI also still ignore remote_accessible. Is this PR intentionally RDMA-only?
> 2. PR title — The title looks concatenated (...registrationFix/rdma...); worth cleaning up before merge.

Thanks for the review!

Protocol coverage——Yes, this PR is intentionally scoped to RDMA verbs MR registration, as it addresses an immediate requirement in Broadcom RDMA NIC environment. I will follow up with EFA / CXI / barex support in a separate PR.
PR title——Updated. Cleaned up the title to [Bugfix][TE] Respect remote_accessible for RDMA verbs MR registration.

mooncake-transfer-engine/src/transport/rdma_transport/rdma_transport.cpp
RuiqingFeng: Good catch! Removed the unused outer-scope BufferDesc buffer_desc; declaration. Thanks!

staryxchen (approved):
LGTM
@alogfans pls double chck

alogfans (commented):
Overall LGTM. However, when the peer side rejects this memory region, it should print to log clearly about this to avoid slient bugs.

RuiqingFeng:
> Overall LGTM. However, when the peer side rejects this memory region, it should print to log clearly about this to avoid slient bugs.

Added an explicit error log `[RDMA] No rkey for MR access: seg=... addr=... len=...` in `selectPeerDevice()` when a peer memory region lacks remote access. Thanks!

alogfans (commented):
Instead of omitting local-only RDMA buffers from the segment metadata, could we keep their address ranges registered and explicitly mark them, while leaving the `rkey` unavailable? This allows the peer to reject such regions deterministically and distinguish them from missing or stale metadata, reducing the risk of accidental fallback or misuse.

RuiqingFeng:
> Instead of omitting local-only RDMA buffers from the segment metadata, could we keep their address ranges registered and explicitly mark them, while leaving the `rkey` unavailable? This allows the peer to reject such regions deterministically and distinguish them from missing or stale metadata, reducing the risk of accidental fallback or misuse.

Agreed! Updated to preserve local-only buffers in metadata with an empty rkey list ([]). Peers can now match the address range and fail with error `No rkey for MR access` without unnecessary reloads. Good idea, thanks!

alogfans (approved):

PR #3152, opened 2026-07-28
[Store] Fix ABBA deadlock between ApplySnapshotState and snapshot_mutex_

rockuw:
ApplySnapshotState (the post-restore cleanup in the snapshot restore path) held the exclusive segment_mutex_ (via getSegmentAccess) while calling UnmountSegment() and Ping(). UnmountSegment acquires snapshot_mutex_ (shared) and Ping acquires client_mutex_. The documented lock order in master_service.h is client/snapshot BEFORE segment, so this is an inversion (segment -> snapshot/client).

Run concurrently with a snapshot->segment path such as GracefulUnmountSegment (snapshot UNIQUE -> segment UNIQUE), the two threads wedge in an ABBA:
  - ApplySnapshotState holds segment(U) and wants snapshot(S) via UnmountSegment
  - GracefulUnmountSegment holds snapshot(U) and wants segment(U) With the writer-preference of std::shared_mutex this is reachable even against shared snapshot readers. This is the same deadlock family as commit 299e95e7 (GracefulUnmountScheduler <-> snapshot_mutex_).

Gather the unready/all segment lists under the segment lock and release it before calling UnmountSegment()/Ping(), matching the pattern UnmountSegment itself uses (release the segment lock before the long-running ClearInvalidHandles).

Add tests/master_service_deadlock_test.cpp, a fork-isolated white-box test that reproduces the ABBA deterministically (the inverted order deadlocks; the canonical order completes).

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
```

**Test results:**
- [X] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

### DIFF
--- a/mooncake-store/include/master_service.h
+++ b/mooncake-store/include/master_service.h
@@ -78,6 +78,10 @@ class SnapshotChildProcessTest;
 class PromotionOnHitTest;
 class MasterServiceTenantQuotaTest;
 class MasterServiceHATest;
+// Friended so the lock-order deadlock reproduction test can drive the real
+// private snapshot_mutex_ / segment_manager_ locks in the exact orders used
+// by GracefulUnmountSegment and ApplySnapshotState.
+class MasterServiceDeadlockTest;
 }  // namespace test
 namespace benchmarks {
 class BatchEvictBench;
@@ -107,6 +111,7 @@ class MasterService {
     friend class test::PromotionOnHitTest;
     friend class benchmarks::BatchEvictBench;
     friend class test::MasterServiceTenantQuotaTest;
+    friend class test::MasterServiceDeadlockTest;  // lock-order ABBA repro
     friend class MasterSnapshotManager;    // Allow access to internal state for
                                            // snapshot
     friend class ha::MasterSnapshotCodec;  // Allow codec to access private
--- a/mooncake-store/src/master_service.cpp
+++ b/mooncake-store/src/master_service.cpp
@@ -7625,20 +7625,32 @@ tl::expected<void, SerializationError> MasterService::ApplySnapshotState(
                 segment_name);
         }
 
-        ScopedSegmentAccess segment_access =
-            segment_manager_.getSegmentAccess();
+        // Gather segment data under the segment lock, then RELEASE it before
+        // calling UnmountSegment()/Ping() below. UnmountSegment acquires
+        // snapshot_mutex_ (shared) and Ping acquires client_mutex_; the
+        // documented lock order (see master_service.h) is client/snapshot
+        // BEFORE segment. Holding the segment lock here while calling them
+        // inverts that order (segment -> snapshot/client) and ABBA-deadlocks
+        // against snapshot->segment paths such as GracefulUnmountSegment.
+        // Reproducer: tests/master_service_deadlock_test.cpp.
         std::vector<std::pair<Segment, UUID>> unready_segments;
-        if (segment_access.GetUnreadySegments(unready_segments) ==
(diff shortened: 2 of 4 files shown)


Icedcoco:
Thanks for working on this. I have a few questions about the diagnosis and test coverage:

1. The new test constructs the two lock orders directly and does not call `ApplySnapshotState()`. It looks like both tests would still pass if the production fix were reverted. Could we instead exercise the real restore path with a non-`OK` segment, so the test fails before this change and passes afterward? Also, should `FinishesWithin()` verify the child’s exit status rather than treating any exit as successful completion?

2. Could the immediate production failure actually be a same-thread recursive lock rather than the described two-thread ABBA? `ApplySnapshotState()` already holds `segment_mutex_` when it calls `UnmountSegment()`, which later tries to acquire `segment_mutex_` again. With any unready segment, that path appears able to block by itself. Does that match your understanding of the observed hang?

3. I am not sure the `GracefulUnmountSegment()` ABBA described in the PR is currently reachable. `ApplySnapshotState()` is only called from `RestoreState()` during `MasterService` construction, before the RPC service and background workers are started. Is there another path that can invoke `GracefulUnmountSegment()` concurrently during restore, or would it be more accurate to describe it as a theoretical lock-order inversion rather than the production trigger?

4. The patch snapshots `all_segments` before unmounting the unready segments, then later calls `Ping()` and adds to `total_size` using that old list. This means successfully removed segments are still pinged and included in the reported restored capacity. Would it be better to read `all_segments` again after unmounting, with the capacity metric reset moved accordingly?

rockuw:
> Thanks for working on this. I have a few questions about the diagnosis and test coverage:
> 
> 1. The new test constructs the two lock orders directly and does not call `ApplySnapshotState()`. It looks like both tests would still pass if the production fix were reverted. Could we instead exercise the real restore path with a non-`OK` segment, so the test fails before this change and passes afterward? Also, should `FinishesWithin()` verify the child’s exit status rather than treating any exit as successful completion?
> 2. Could the immediate production failure actually be a same-thread recursive lock rather than the described two-thread ABBA? `ApplySnapshotState()` already holds `segment_mutex_` when it calls `UnmountSegment()`, which later tries to acquire `segment_mutex_` again. With any unready segment, that path appears able to block by itself. Does that match your understanding of the observed hang?
> 3. I am not sure the `GracefulUnmountSegment()` ABBA described in the PR is currently reachable. `ApplySnapshotState()` is only called from `RestoreState()` during `MasterService` construction, before the RPC service and background workers are started. Is there another path that can invoke `GracefulUnmountSegment()` concurrently during restore, or would it be more accurate to describe it as a theoretical lock-order inversion rather than the production trigger?
> 4. The patch snapshots `all_segments` before unmounting the unready segments, then later calls `Ping()` and adds to `total_size` using that old list. This means successfully removed segments are still pinged and included in the reported restored capacity. Would it be better to read `all_segments` again after unmounting, with the capacity metric reset moved accordingly?

多谢Review！你说得对，这个确实不是线上hang的根因，线上的问题是我们的内部版本在添加offload queue size metrics时引入的一个死锁，当前主干没有这个问题。

ApplySnapshotState是restore阶段调用的，确实不会和GracefulUnmountSegment竞争，这个pr可以关闭了。

PR #3154, opened 2026-07-28
[Store] Fix segment capacity metric leak on master service teardown

liuzijing2014:
## Problem

`MasterMetricManager` is a process-lifetime singleton, but segment capacity is only decremented in `CommitUnmountSegment`. In HA mode the supervisor constructs a fresh `WrappedMasterService` for every leadership term and destroys the old one on demotion. If the service is destroyed while segments are still mounted, the records die with the instance and their `total_mem_capacity` contribution is never released.

The only path that unmounts segments of silent clients is the client-expiry sweep, which needs ~`client_live_ttl_sec` to fire, while a clean demotion tears the service down in milliseconds — so a demotion typically strands the full mounted capacity in the singleton. After the next promotion, clients remount into the fresh instance and the same capacity is counted again: each leadership change inflates reported capacity by one full fleet of segments (2x, 3x, ...).

We hit this in production during an etcd latency incident that caused repeated leadership swaps: both masters ended up reporting exactly 2x the real capacity. `global_mem_used_ratio` plateaued at ~0.47 and could never reach `eviction_high_watermark_ratio` (0.9), so proactive eviction stopped entirely and writes ran into continuous `NO_AVAILABLE_HANDLE` allocation failures at the real memory ceiling, with reactive eviction as the only relief.

## Fix

Release the remaining mounted segments' capacity contribution in `~SegmentManager()`. Any segment still in `mounted_segments_` at destruction has by definition not gone through `CommitUnmountSegment`, so the decrement is exact and cannot double-count. CXL segments are skipped, mirroring the mount/unmount paths.

## Testing

Added `MasterMetricsTest.ServiceTeardownReleasesSegmentCapacity`: mounts a segment, destroys the service without unmounting (as a demotion does), and asserts the gauges return to baseline. `master_metrics_test` (12), `segment_test` (14) and `master_service_test` (145) all pass locally.

### DIFF
--- a/mooncake-store/include/segment.h
+++ b/mooncake-store/include/segment.h
@@ -452,6 +452,14 @@ class SegmentManager {
         : memory_allocator_(memory_allocator), enable_cxl_(enable_cxl) {}
 
     /**
+     * @brief Destructor. Releases the capacity metric contribution of
+     *        segments that are still mounted, since MasterMetricManager
+     *        outlives MasterService instances (e.g. across HA leadership
+     *        changes).
+     */
+    ~SegmentManager();
+
+    /**
      * @brief Get RAII-style access to segment management operations
      * @return ScopedSegmentAccess object that holds the lock
      */
--- a/mooncake-store/src/segment.cpp
+++ b/mooncake-store/src/segment.cpp
@@ -1557,6 +1557,24 @@ void NoFSegmentManager::GetMountedSegmentsSnapshot(
     }
 }
 
+SegmentManager::~SegmentManager() {
+    // Segments that are still mounted here never went through
+    // CommitUnmountSegment, so their contribution to the capacity metrics
+    // has not been released. MasterMetricManager outlives MasterService
+    // instances (a new one is constructed per HA leadership term), so
+    // release it here to keep the gauges consistent with the segments
+    // that are actually mounted.
+    for (const auto& [segment_id, mounted_segment] : mounted_segments_) {
+        const auto& segment = mounted_segment.segment;
+        if (segment.protocol == "cxl") {
+            // CXL mounts do not contribute to total_mem_capacity.
+            continue;
+        }
+        MasterMetricManager::instance().dec_total_mem_capacity(segment.name,
+                                                               segment.size);
+    }
+}
+
 void SegmentManager::initializeCxlAllocator(const std::string& cxl_path,
                                             const size_t cxl_size) {
     LOG(INFO) << "Init CXL global allocator.";
--- a/mooncake-store/tests/master_metrics_test.cpp
+++ b/mooncake-store/tests/master_metrics_test.cpp
@@ -270,6 +270,36 @@ TEST_F(MasterMetricsTest, BasicRequestTest) {
     ASSERT_DOUBLE_EQ(metrics.get_segment_mem_used_ratio("xxxxxx_segment"), 0.0);
 }
 
+TEST_F(MasterMetricsTest, ServiceTeardownReleasesSegmentCapacity) {
+    auto& metrics = MasterMetricManager::instance();
+    const int64_t capacity_before = metrics.get_total_mem_capacity();
+
+    Segment segment;
+    segment.id = generate_uuid();
+    segment.name = "teardown_test_segment";
+    segment.base = 0x300000000;
+    segment.size = 1024 * 1024 * 16;
+    UUID client_id = generate_uuid();
+
+    {
+        WrappedMasterServiceConfig service_config;
+        service_config.default_kv_lease_ttl = 100;
+        service_config.enable_metric_reporting = false;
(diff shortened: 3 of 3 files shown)


Aionw (approved):
Solid PR. LGTM

Icedcoco:
Thanks for putting this together. The diagnosis around `MasterMetricManager` outliving each serving `MasterService` makes sense to me.

I may be missing something, but I think there are two lifecycle cases worth checking before we put the cleanup in `SegmentManager`'s destructor.

The first one looks important for this PR. `CatalogBackedSnapshotProvider` creates a temporary `SegmentManager` and populates its `mounted_segments_` through `SegmentSerializer::Deserialize` (`mooncake-store/src/ha/snapshot/catalog_backed_snapshot_provider.cpp:472`). That deserialization path inserts the segment records directly and does not call `inc_total_mem_capacity`; the increment currently only happens through `ScopedSegmentAccess::MountSegment` (`mooncake-store/src/segment.cpp:217`).

With the proposed destructor, destroying this temporary snapshot reader would still decrement every deserialized segment. That could make `master_total_capacity_bytes` lower than the capacity actually accounted by the serving service, or negative while the pod is in standby. After promotion, a client remount could then bring the gauge back only to zero rather than to the real capacity.

Would it make sense to keep this cleanup at the `MasterService` lifecycle boundary, or otherwise track whether a `SegmentManager` instance actually contributed these metrics? A focused test that deserializes a snapshot into a temporary `SegmentManager`, destroys it, and verifies that the capacity gauge is unchanged would help cover this path.

There is also a related CXL scope question. The per-client CXL mount path does not increment `total_mem_capacity`, so skipping those mounted segment entries is reasonable. However, `initializeCxlAllocator()` increments the same gauge once per `MasterService` instance (`mooncake-store/src/segment.cpp:1560`), and a new `MasterService` is constructed for each HA leadership term. I could not find a configuration check rejecting `enable_ha` together with `enable_cxl`, although standby remount of restored CXL segments is currently rejected.

If HA+CXL is intentionally unsupported, documenting or enforcing that restriction may be enough for this PR. Otherwise, should the teardown also release the single global `cxl_size` contribution? This seems separate from the snapshot issue above, so I think it could also be handled as an explicitly scoped follow-up.

stmatengss (approved):

liuzijing2014:
> Thanks for putting this together. The diagnosis around `MasterMetricManager` outliving each serving `MasterService` makes sense to me.
> 
> I may be missing something, but I think there are two lifecycle cases worth checking before we put the cleanup in `SegmentManager`'s destructor.
> 
> The first one looks important for this PR. `CatalogBackedSnapshotProvider` creates a temporary `SegmentManager` and populates its `mounted_segments_` through `SegmentSerializer::Deserialize` (`mooncake-store/src/ha/snapshot/catalog_backed_snapshot_provider.cpp:472`). That deserialization path inserts the segment records directly and does not call `inc_total_mem_capacity`; the increment currently only happens through `ScopedSegmentAccess::MountSegment` (`mooncake-store/src/segment.cpp:217`).
> 
> With the proposed destructor, destroying this temporary snapshot reader would still decrement every deserialized segment. That could make `master_total_capacity_bytes` lower than the capacity actually accounted by the serving service, or negative while the pod is in standby. After promotion, a client remount could then bring the gauge back only to zero rather than to the real capacity.
> 
> Would it make sense to keep this cleanup at the `MasterService` lifecycle boundary, or otherwise track whether a `SegmentManager` instance actually contributed these metrics? A focused test that deserializes a snapshot into a temporary `SegmentManager`, destroys it, and verifies that the capacity gauge is unchanged would help cover this path.
> 
> There is also a related CXL scope question. The per-client CXL mount path does not increment `total_mem_capacity`, so skipping those mounted segment entries is reasonable. However, `initializeCxlAllocator()` increments the same gauge once per `MasterService` instance (`mooncake-store/src/segment.cpp:1560`), and a new `MasterService` is constructed for each HA leadership term. I could not find a configuration check rejecting `enable_ha` together with `enable_cxl`, although standby remount of restored CXL segments is currently rejected.
> 
> If HA+CXL is intentionally unsupported, documenting or enforcing that restriction may be enough for this PR. Otherwise, should the teardown also release the single global `cxl_size` contribution? This seems separate from the snapshot issue above, so I think it could also be handled as an explicitly scoped follow-up.

Thanks for the feedback. I think the lifecycle around snapshot restore makes sense. Let me double check

Dao007forever:
@Icedcoco does https://github.com/kvcache-ai/Mooncake/pull/3169 address your concerns?

liuzijing2014:
@stmatengss @Icedcoco @Aionw mind taking a look of this follow up PR: https://github.com/kvcache-ai/Mooncake/pull/3168, which would fix the HA + snapshot restore path as well.

PR #3159, opened 2026-07-28
[store] Avoid staging copy for same node tensor put

zxpdemonio:
### Motivation

The Python tensor write path used to copy tensor payloads through a client-side staging buffer before the data reached the Mooncake Store segment.

For local same-node tensor writes, this staging copy is unnecessary when the target Store segment is local to the writer process. It is especially expensive for GPU tensor -> Store CPU segment writes, where the old path first copied the GPU tensor into a client local/staging buffer and then copied that buffer into the Store segment.

This PR removes the extra staging copy from the local tensor write path.

This PR is split from the broader local tensor/offload optimization work. The scope here is intentionally limited to same-node tensor write staging removal.

### Scenario

Before this PR:

```text
tensor payload
  -> client local/staging buffer
  -> Store segment
```

After this PR, when the write can safely use the local same-node path:

```text
tensor payload
  -> Store segment
```

### Changes

- Add a direct multi-buffer write path for tensor payloads.
- Let local same-node tensor writes pass tensor payload buffers directly into the Store write path.
- Extend the transfer task local-copy path to consume caller-provided source buffers for local writes.
- Route put/upsert tensor writes through the shared multi-buffer writer to avoid duplicate tensor staging logic.
- Route single `put_tensor` and `upsert_tensor` through the same no-staging multi-buffer path as batch tensor writes.

#### Local write staging removal

Environment:

```text
Mode: local same-node tensor put
Protocol: rdma
RDMA device: erdma_0
Sizes: 64 MB, 256 MB, 512 MB
Warmup / iterations: 2 / 7
Baseline: f224dcaf
This PR: c378659e
Benchmark harness: external local script, not included in this PR
```

The benchmark log confirms `MC_STORE_MEMCPY` was auto-disabled under RDMA, so the baseline uses the old staging/TransferEngine path. The after case uses the same public tensor APIs and removes the local staging copy when the destination is the local Store segment.

Local CPU tensor -> Store CPU segment put:

| Size | Before | After | Improvement |
| --- | --- | --- | --- |
| 64 MB | 5.106 GB/s / 13.142 ms | 13.924 GB/s / 4.820 ms | +173% |
| 256 MB | 5.160 GB/s / 52.027 ms | 13.341 GB/s / 20.121 ms | +159% |
| 512 MB | 5.459 GB/s / 98.348 ms | 13.340 GB/s / 40.245 ms | +144% |

Local GPU tensor -> Store CPU segment put:

| Size | Before | After | Improvement |
| --- | --- | --- | --- |
| 64 MB | 2.515 GB/s / 26.689 ms | 7.595 GB/s / 8.836 ms | +202% |
| 256 MB | 2.586 GB/s / 103.820 ms | 7.752 GB/s / 34.626 ms | +200% |
| 512 MB | 2.752 GB/s / 195.109 ms | 7.776 GB/s / 69.038 ms | +183% |

The main performance gain comes from local same-node tensor writes where the extra staging copy is removed.

#### Pinned Store segment memory local CUDA write

Environment:

```text
Mode: local same-node CUDA tensor write into pinned Store CPU segment
Protocol: tcp
MC_STORE_MEMCPY: 1
Pinned Store segment memory: enabled
MC_STORE_PIN_MEMORY_MAX_BYTES: 4294967296
Store segment size: 4096 MB
Client local buffer size: 1024 MB
Sizes: 64 MB, 256 MB, 512 MB
Warmup / iterations: 3 / 8
Baseline: 6c721eea
This PR: 00230d8f
Benchmark harness: external local one-shot script, not included in this PR
```

This benchmark uses pinned Store segment memory and local memcpy. It validates that single tensor writes now reach the same no-staging multi-buffer path as batch tensor writes.

Local CUDA tensor -> pinned Store CPU segment `put_tensor`:

| Size | Base GB/s | PR GB/s | Speedup | Base median | PR median |
| ---: | ---: | ---: | ---: | ---: | ---: |
| 64 MB | 3.363 | 24.923 | 7.41x | 19.955 ms | 2.693 ms |
| 256 MB | 3.333 | 25.824 | 7.75x | 80.532 ms | 10.395 ms |
| 512 MB | 3.345 | 25.921 | 7.75x | 160.483 ms | 20.712 ms |

Local CUDA tensor -> pinned Store CPU segment `upsert_tensor`:

| Size | Base GB/s | PR GB/s | Speedup | Base median | PR median |
| ---: | ---: | ---: | ---: | ---: | ---: |
| 64 MB | 3.379 | 24.702 | 7.31x | 19.858 ms | 2.717 ms |
| 256 MB | 3.334 | 25.712 | 7.71x | 80.520 ms | 10.440 ms |
| 512 MB | 3.346 | 25.923 | 7.75x | 160.430 ms | 20.710 ms |

## Module

- [ ] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Breaking change
- [x] Documentation update
- [x] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**

```bash
./scripts/code_format.sh --check -b origin/main

cmake --build build-tensor-staging-nocuda-notest -j8 --target store

cmake -S . -B build-tensor-staging-cuda \
  -DUSE_CUDA=ON \
  -DWITH_STORE_RUST=OFF \
  -DWITH_EP=OFF \
  -DBUILD_UNIT_TESTS=OFF \
  -DBUILD_TESTS=OFF \
  -DCMAKE_BUILD_TYPE=Release

cmake --build build-tensor-staging-cuda -j8 --target store mooncake_master
```

Manual validation was run with local benchmark/smoke harnesses that are not included in this PR:

- local CPU tensor -> Store CPU segment put
- local GPU tensor -> Store CPU segment put
- CPU tensor write smoke for `put_tensor`, `upsert_tensor`, `batch_put_tensor`, and `batch_upsert_tensor`
- CUDA duplicate `put_tensor` smoke to confirm the existing successful return behavior for an existing key
- CUDA tensor local write benchmark with pinned Store segment memory and local memcpy enabled

**Test results:**

- [x] Format check passes
- [x] Build passes without CUDA enabled
- [x] Build passes with CUDA enabled
- [x] Manual CPU tensor write smoke passes
- [x] Manual CUDA duplicate `put_tensor` smoke passes
- [x] Manual performance validation done

Manual test results:

- Local CPU tensor put improved from 5.1-5.5 GB/s to 13.3-13.9 GB/s.
- Local GPU tensor put improved from 2.5-2.8 GB/s to 7.6-7.8 GB/s.
- With pinned Store segment memory, local CUDA `put_tensor` improved from 3.33-3.36 GB/s to 24.92-25.92 GB/s.
- With pinned Store segment memory, local CUDA `upsert_tensor` improved from 3.33-3.38 GB/s to 24.70-25.92 GB/s.

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used
- [x] AI tools were used (Codex was used to help with implementation, benchmarking, and review)

### DIFF
--- a/mooncake-integration/store/store_py.cpp
+++ b/mooncake-integration/store/store_py.cpp
@@ -1045,6 +1045,10 @@ class MooncakeStorePyWrapper {
         const std::vector<std::string> &keys,
         const std::vector<PyTensorInfo> &infos,
         const ReplicateConfig &config = ReplicateConfig{}) {
+        if (auto cuda_ipc_results = try_dummy_cuda_ipc_batch_upsert_tensor_impl(
+                keys, infos, config)) {
+            return *cuda_ipc_results;
+        }
         return batch_write_tensor_impl(
             keys, infos, config, "upsert",
             [this](const std::vector<std::string> &write_keys,
--- a/mooncake-integration/store/store_py.cpp
+++ b/mooncake-integration/store/store_py.cpp
@@ -164,12 +164,6 @@ py::tuple tensor_shape_tuple(const TensorMetadata &metadata) {
     return py::cast(shape_vec);
 }
 
-void append_tensor_payload_span(std::vector<std::span<const char>> &values,
-                                uintptr_t data_ptr, size_t tensor_size) {
-    if (tensor_size == 0) return;
-    values.emplace_back(reinterpret_cast<const char *>(data_ptr), tensor_size);
-}
-
 pybind11::object buffer_to_tensor(BufferHandle *buffer_handle, char *usr_buffer,
                                   int64_t data_length,
                                   bool own_user_buffer = false) {
@@ -918,24 +912,7 @@ class MooncakeStorePyWrapper {
                         const ReplicateConfig &config) {
         // Validation & Metadata extraction (GIL Held)
         auto info = extract_tensor_info(tensor, key);
-        if (!info.valid()) return to_py_ret(ErrorCode::INVALID_PARAMS);
-        if (use_dummy_client_) {
-            return put_tensor_info_impl(key, info, config);
-        }
-
-        // Prepare spans
-        std::vector<std::span<const char>> values;
-        values.emplace_back(reinterpret_cast<const char *>(&info.metadata),
-                            sizeof(TensorMetadata));
-        append_tensor_payload_span(values, info.data_ptr, info.tensor_size);
-
-        // Store (GIL Released)
-        py::gil_scoped_release release_gil;
-        int ret = store_->put_parts(key, values, config);
-        if (ret != 0)
-            LOG(ERROR) << "put_parts failed for key " << key << " with code "
-                       << ret;
-        return ret;
+        return put_tensor_info_impl(key, info, config);
     }
 
     int put_tensor(const std::string &key, pybind11::object tensor) {
@@ -1287,17 +1264,15 @@ class MooncakeStorePyWrapper {
                                    : results[0];
         }
 
-        std::vector<std::span<const char>> values;
-        values.emplace_back(reinterpret_cast<const char *>(&info.metadata),
-                            info.metadata.header.data_offset);
-        append_tensor_payload_span(values, info.data_ptr, info.tensor_size);
-
-        py::gil_scoped_release release_gil;
-        int ret = store_->put_parts(key, values, config);
-        if (ret != 0)
-            LOG(ERROR) << "put_parts failed for key " << key << " with code "
-                       << ret;
-        return ret;
+        auto results = batch_put_tensor_infos_impl(
+            std::vector<std::string>{key}, std::vector<PyTensorInfo>{info},
+            config);
+        if (results.size() != 1) {
+            LOG(ERROR) << "put returned unexpected result count for key "
+                       << key;
+            return to_py_ret(ErrorCode::RPC_FAIL);
+        }
+        return results[0];
     }
 
     int put_manifest_impl(const std::string &key,
@@ -1494,38 +1469,21 @@ class MooncakeStorePyWrapper {
                                    : results[0];
         }
 
-        std::vector<std::span<const char>> values;
-        values.emplace_back(reinterpret_cast<const char *>(&info.metadata),
-                            info.metadata.header.data_offset);
-        append_tensor_payload_span(values, info.data_ptr, info.tensor_size);
-
-        py::gil_scoped_release release_gil;
-        int ret = store_->upsert_parts(key, values, config);
-        if (ret != 0)
-            LOG(ERROR) << "upsert_parts failed for key " << key << " with code "
-                       << ret;
-        return ret;
+        auto results = batch_upsert_tensor_infos_impl(
+            std::vector<std::string>{key}, std::vector<PyTensorInfo>{info},
+            config);
+        if (results.size() != 1) {
+            LOG(ERROR) << "upsert returned unexpected result count for key "
+                       << key;
+            return to_py_ret(ErrorCode::RPC_FAIL);
+        }
+        return results[0];
     }
 
     int upsert_tensor_impl(const std::string &key, pybind11::object tensor,
                            const ReplicateConfig &config) {
         auto info = extract_tensor_info(tensor, key);
-        if (!info.valid()) return to_py_ret(ErrorCode::INVALID_PARAMS);
-        if (use_dummy_client_) {
-            return upsert_tensor_info_impl(key, info, config);
-        }
-
-        std::vector<std::span<const char>> values;
-        values.emplace_back(reinterpret_cast<const char *>(&info.metadata),
-                            sizeof(TensorMetadata));
-        append_tensor_payload_span(values, info.data_ptr, info.tensor_size);
-
-        py::gil_scoped_release release_gil;
-        int ret = store_->upsert_parts(key, values, config);
-        if (ret != 0)
-            LOG(ERROR) << "upsert_parts failed for key " << key << " with code "
-                       << ret;
-        return ret;
+        return upsert_tensor_info_impl(key, info, config);
     }
 
     int upsert_tensor(const std::string &key, pybind11::object tensor) {
@@ -1548,9 +1506,14 @@ class MooncakeStorePyWrapper {
                 return to_py_ret(ErrorCode::INVALID_PARAMS);
             }
 
+            std::vector<std::string> tp_keys;
+            tp_keys.reserve(tp_size);
             for (int rank = 0; rank < tp_size; ++rank) {
-                int ret = upsert_tensor_info_impl(get_tp_key_name(key, rank),
-                                                  (*infos)[rank], config);
+                tp_keys.push_back(get_tp_key_name(key, rank));
+            }
+            auto results =
+                batch_upsert_tensor_infos_impl(tp_keys, *infos, config);
+            for (int ret : results) {
                 if (ret != 0) return ret;
             }
             return 0;
@@ -1608,21 +1571,6 @@ class MooncakeStorePyWrapper {
         py::gil_scoped_release release_gil;
         return store_->batch_upsert_from(keys, buffers, sizes,
                                          ReplicateConfig{});
-    }
-
-    std::vector<int> batch_upsert_tensor_infos_impl(
-        const std::vector<std::string> &keys,
-        const std::vector<PyTensorInfo> &infos,
-        const ReplicateConfig &config = ReplicateConfig{}) {
-        return batch_write_tensor_impl(
-            keys, infos, config, "upsert",
-            [this](const std::vector<std::string> &write_keys,
-                   const std::vector<void *> &buffer_ptrs,
-                   const std::vector<size_t> &buffer_sizes,
-                   const ReplicateConfig &write_config) {
-                return store_->batch_upsert_from(write_keys, buffer_ptrs,
-                                                 buffer_sizes, write_config);
-            });
     }
 
     std::vector<int> batch_upsert_tensor_impl(
--- a/mooncake-integration/store/store_py.cpp
+++ b/mooncake-integration/store/store_py.cpp
@@ -1038,7 +1038,7 @@ class MooncakeStorePyWrapper {
                    const std::vector<std::vector<size_t>> &sizes,
                    const ReplicateConfig &write_config) {
                 return real_client_->batch_put_from_multi_buffers(
-                    write_keys, buffers, sizes, write_config);
+                    write_keys, buffers, sizes, write_config, true);
             });
     }
 
@@ -1064,7 +1064,7 @@ class MooncakeStorePyWrapper {
                    const std::vector<std::vector<size_t>> &sizes,
                    const ReplicateConfig &write_config) {
                 return real_client_->batch_upsert_from_multi_buffers(
-                    write_keys, buffers, sizes, write_config);
+                    write_keys, buffers, sizes, write_config, true);
             });
     }
 
--- a/mooncake-integration/store/store_py.cpp
+++ b/mooncake-integration/store/store_py.cpp
@@ -1058,7 +1058,7 @@ class MooncakeStorePyWrapper {
                    const std::vector<std::vector<void *>> &buffers,
                    const std::vector<std::vector<size_t>> &sizes,
                    const ReplicateConfig &write_config) {
-                return real_client_->batch_upsert_from_multi_buffers(
+                return store_->batch_upsert_from_multi_buffers(
                     write_keys, buffers, sizes, write_config);
             });
     }
--- a/mooncake-integration/store/store_py.cpp
+++ b/mooncake-integration/store/store_py.cpp
@@ -912,6 +912,7 @@ class MooncakeStorePyWrapper {
                         const ReplicateConfig &config) {
         // Validation & Metadata extraction (GIL Held)
         auto info = extract_tensor_info(tensor, key);
+        if (!info.valid()) return to_py_ret(ErrorCode::INVALID_PARAMS);
         return put_tensor_info_impl(key, info, config);
     }
 
@@ -1062,7 +1063,7 @@ class MooncakeStorePyWrapper {
                    const std::vector<std::vector<void *>> &buffers,
                    const std::vector<std::vector<size_t>> &sizes,
                    const ReplicateConfig &write_config) {
-                return store_->batch_upsert_from_multi_buffers(
+                return real_client_->batch_upsert_from_multi_buffers(
                     write_keys, buffers, sizes, write_config);
             });
     }
@@ -1071,6 +1072,11 @@ class MooncakeStorePyWrapper {
         const std::vector<std::string> &keys,
         const pybind11::list &tensors_list,
         const ReplicateConfig &config = ReplicateConfig{}) {
+        if (pybind11::len(tensors_list) != keys.size()) {
+            LOG(ERROR) << "batch_put_tensor keys and tensors size mismatch";
+            return std::vector<int>(keys.size(),
+                                    to_py_ret(ErrorCode::INVALID_PARAMS));
+        }
         std::vector<PyTensorInfo> infos(keys.size());
         for (size_t i = 0; i < keys.size(); ++i) {
             infos[i] = extract_tensor_info(tensors_list[i], keys[i]);
@@ -1487,6 +1493,7 @@ class MooncakeStorePyWrapper {
     int upsert_tensor_impl(const std::string &key, pybind11::object tensor,
                            const ReplicateConfig &config) {
         auto info = extract_tensor_info(tensor, key);
+        if (!info.valid()) return to_py_ret(ErrorCode::INVALID_PARAMS);
         return upsert_tensor_info_impl(key, info, config);
     }
 
@@ -1581,6 +1588,11 @@ class MooncakeStorePyWrapper {
         const std::vector<std::string> &keys,
         const pybind11::list &tensors_list,
         const ReplicateConfig &config = ReplicateConfig{}) {
+        if (pybind11::len(tensors_list) != keys.size()) {
+            LOG(ERROR) << "batch_upsert_tensor keys and tensors size mismatch";
+            return std::vector<int>(keys.size(),
+                                    to_py_ret(ErrorCode::INVALID_PARAMS));
+        }
         std::vector<PyTensorInfo> infos(keys.size());
         for (size_t i = 0; i < keys.size(); ++i) {
             infos[i] = extract_tensor_info(tensors_list[i], keys[i]);
--- a/mooncake-integration/store/store_py.cpp
+++ b/mooncake-integration/store/store_py.cpp
@@ -1049,11 +1049,40 @@ class MooncakeStorePyWrapper {
         return batch_write_tensor_impl(
             keys, infos, config, "put",
             [this](const std::vector<std::string> &write_keys,
-                   const std::vector<void *> &buffer_ptrs,
-                   const std::vector<size_t> &buffer_sizes,
+                   const std::vector<void *> &buffers,
+                   const std::vector<size_t> &sizes,
                    const ReplicateConfig &write_config) {
-                return store_->batch_put_from(write_keys, buffer_ptrs,
-                                              buffer_sizes, write_config);
+                return store_->batch_put_from(write_keys, buffers, sizes,
+                                              write_config);
+            },
+            [this](const std::vector<std::string> &write_keys,
+                   const std::vector<std::vector<void *>> &buffers,
+                   const std::vector<std::vector<size_t>> &sizes,
+                   const ReplicateConfig &write_config) {
+                return real_client_->batch_put_from_multi_buffers(
+                    write_keys, buffers, sizes, write_config);
+            });
+    }
+
+    std::vector<int> batch_upsert_tensor_infos_impl(
+        const std::vector<std::string> &keys,
+        const std::vector<PyTensorInfo> &infos,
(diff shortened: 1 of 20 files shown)


mooncake-integration/store/store_py.cpp
```diff
@@ -752,25 +746,30 @@ class MooncakeStorePyWrapper {
...
+        if (write_config.prefer_alloc_in_same_node) {
+            return write_config;
+        }
+        if (write_config.replica_num == 1 &&
+            write_config.nof_replica_num == 0 &&
+            write_config.preferred_segment.empty() &&
+            !write_config.group_ids.has_value() &&
+            write_config.preferred_segments.empty() &&
+            write_config.preferred_nof_segments.empty()) {
+            write_config.prefer_alloc_in_same_node = true;
```
zxpdemonio: **Blocker:** this auto-enable ignores an explicit `prefer_alloc_in_same_node=False`.

For a default-looking config (`replica_num==1`, empty preferred segments, no group ids), any caller who sets `prefer_alloc_in_same_node=False` still gets forced onto the no-staging / same-node path.

Please honor an explicit `False` (or use a tri-state / separate "auto" policy). As written, there is no clean opt-out, and default tensor put/upsert lose the staging fallback described in the PR body when local allocation is unavailable.

mooncake-integration/store/store_py_parallel_write.h
```diff
@@ -14,27 +14,92 @@ int write_manifest_impl(const std::string &key,
...
-std::vector<int> batch_write_tensor_impl(const std::vector<std::string> &keys,
-                                         const std::vector<PyTensorInfo> &infos,
-                                         const ReplicateConfig &config,
-                                         const char *operation_name,
-                                         BatchWriteFromFn &&batch_write_from) {
+inline void append_tensor_write_buffers(const PyTensorInfo &info,
+                                        std::vector<void *> &buffers,
+                                        std::vector<size_t> &sizes) {
+    buffers.push_back(const_cast<TensorMetadata *>(&info.metadata));
+    sizes.push_back(sizeof(TensorMetadata));
```
zxpdemonio: **Nit / future-proofing:** the staging path sizes/copies with `info.metadata.header.data_offset`, but the no-staging path hard-codes `sizeof(TensorMetadata)`.

Please use `info.metadata.header.data_offset` here (and ideally assert it matches the first buffer length) so payload placement stays consistent if metadata padding / offset ever diverges.

mooncake-store/include/pyclient.h
```diff
@@ -314,12 +314,19 @@ class PyClient {
...
         const std::vector<std::string> &keys,
         const std::vector<void *> &buffers, const std::vector<size_t> &sizes,
         const ReplicateConfig &config = ReplicateConfig{}) = 0;
 
+    virtual std::vector<int> batch_upsert_from_multi_buffers(
+        const std::vector<std::string> &keys,
+        const std::vector<std::vector<void *>> &,
+        const std::vector<std::vector<size_t>> &,
+        const ReplicateConfig & = ReplicateConfig{}) {
+        return std::vector<int>(keys.size(), -toInt(ErrorCode::INVALID_PARAMS));
```
zxpdemonio: **Non-blocking:** `batch_upsert_from_multi_buffers` defaults to `INVALID_PARAMS`, while `DummyClient` already implements `batch_put_from_multi_buffers`.

Tensor Python APIs reject dummy clients today, so this may not bite immediately, but if upsert multi-buffer is now part of the public write surface, please add the dummy/RPC helper for parity with put.

mooncake-wheel/tests/test_put_get_tensor.py
```diff
@@ -128,6 +128,33 @@ def test_put_get_tensor(self):
         self.store.remove(key_bool)
         self.store.remove(key_rand)
 
+    def test_batch_pub_tensor_same_node(self):
+        import torch
+
+        config = ReplicateConfig()
+        config.prefer_alloc_in_same_node = True
```
zxpdemonio: **Non-blocking:** good negative test for mismatched preferred segment, but this only covers explicit `prefer_alloc_in_same_node=True`.

Given `MakeDefaultLocalTensorWriteConfig` now changes default tensor write behavior, please add cases for:
1. default config (no explicit prefer flag) round-trip,
2. explicit `prefer_alloc_in_same_node=False` opt-out,
3. upsert / CUDA on the no-staging path.

mooncake-store/src/transfer_task.cpp
```diff
@@ -1035,16 +1035,81 @@ std::optional<TransferFuture> TransferSubmitter::submit(
...
     std::vector<std::vector<Slice>>& all_slices,
-    TransferRequest::OpCode op_code) {
+    TransferRequest::OpCode op_code, bool require_same_process) {
+    if (replicas.size() != all_slices.size()) {
+        LOG(ERROR) << "Mismatched replicas and slice batches: replicas="
+                   << replicas.size() << ", slices=" << all_slices.size();
+        return std::nullopt;
+    }
+
+    if (require_same_process) {
```
zxpdemonio: **Note:** `require_same_process` always uses the memcpy worker and bypasses `MC_STORE_MEMCPY` / TransferEngine selection. That is appropriate for this optimization, but it should be documented because operators who set `MC_STORE_MEMCPY=0` under RDMA will still hit local memcpy on this path.

zxpdemonio (commented):
## Verdict

Useful same-process tensor write optimization with solid measured gains, but **one behavioral bug should be fixed before merge**, and the auto-enable of `prefer_alloc_in_same_node` needs clearer fallback semantics.

## Blockers

1. **`MakeDefaultLocalTensorWriteConfig` cannot honor an explicit `prefer_alloc_in_same_node=False`.**
   For an otherwise-default `ReplicateConfig`, the helper always flips the flag to `true`, even when the caller set it to `false`. There is currently no clean opt-out short of setting a dummy preferred segment / replica policy. That makes the new no-staging path mandatory for common tensor writes and contradicts the PR claim that unsupported paths keep the staging fallback.

2. **Default tensor put/upsert semantics change without a remote/staging fallback.**
   Default `put_tensor` / `upsert_tensor` / batch variants now force same-node allocation + direct buffers. If the local segment cannot allocate, the write fails hard (`NO_AVAILABLE_HANDLE` / same-process submit failure) instead of falling back to the old staging + TransferEngine path. Please either:
   - keep auto-prefer but add an automatic staging fallback when local alloc/submit fails, or
   - make auto-prefer opt-in (or respect explicit `False`) and document the capacity/behavior change prominently.

## Non-blocking

3. **`append_tensor_write_buffers` uses `sizeof(TensorMetadata)` while the staging path uses `metadata.header.data_offset`.** Today builders set them equal, but the no-staging path should use `data_offset` for layout consistency and future-proofing.

4. **`batch_upsert_from_multi_buffers` has no `DummyClient` implementation** (base default returns `INVALID_PARAMS`). Tensor APIs already reject dummy clients, so this is mostly API asymmetry vs `batch_put_from_multi_buffers`, but worth closing if upsert multi-buffer is now a public surface.

5. **Test coverage is thin relative to the behavior change.**
   - Added `test_batch_pub_tensor_same_node` only covers explicit `prefer_alloc_in_same_node=True`.
   - Missing coverage for: default config auto-prefer, explicit `False` opt-out, upsert no-staging, and CUDA tensor correctness under the new default path (beyond the small get equality assert).

6. **`require_same_process` intentionally bypasses `MC_STORE_MEMCPY=0`.** That matches the benchmark story, but please call it out in docs so operators who disabled memcpy are not surprised.

## What looks good

- Constraining no-staging to the current client segment via `normalize_same_process_tensor_write_config`.
- Forcing memcpy through `submit_batch(..., require_same_process=true)` with local-endpoint checks.
- Enabling upsert on the prefer-same-node path and fixing `FinalizeBatchUpsert` to use allocated-replica transfer success instead of `pending_transfers` (which are only tracked on merged ops).
- Sharing put/upsert through `batch_*_from_multi_buffers` / `BatchWriteFromMultiBuffers`.
- Accelerator-aware memcpy worker already handles GPU to CPU for the direct path.

## CI

Checks were still pending/skipped at review time; please confirm format + the listed tensor tests remain green on the latest head.

mooncake-integration/store/store_py.cpp
zxpdemonio: Thanks, agreed. I removed `MakeDefaultLocalTensorWriteConfig` instead of adding an opt-out. The tensor path now preserves the caller's `ReplicateConfig`; after `PutStart` / `UpsertStart`, `Client` checks the allocated replicas and uses direct tensor slices only when all memory replicas are local to this process. Otherwise it lazily builds the old staging slices. So default allocation semantics are unchanged and Python no longer forces a preferred segment.

mooncake-integration/store/store_py_parallel_write.h
zxpdemonio: Done. The direct tensor buffer list now uses `info.metadata.header.data_offset` for the metadata slice, matching the staging path and the serialized tensor layout.

mooncake-store/include/pyclient.h
zxpdemonio: Good point. I left dummy-client `batch_upsert_from_multi_buffers` out of this PR intentionally: tensor Python APIs still reject dummy clients, and adding dummy upsert RPC parity would expand the public dummy-client surface beyond this staging-elimination change. I can split that into a follow-up if we want to expose upsert multi-buffer through dummy clients.

mooncake-wheel/tests/test_put_get_tensor.py
zxpdemonio: Updated. Since the Python-layer auto-enable was removed, there is no explicit opt-out behavior to test. The test now covers default-config tensor batch write plus default tensor upsert, and the existing CUDA test still covers CUDA put/upsert. I removed the mismatched preferred-segment negative case because tensor writes no longer validate or override allocator placement in Python; they follow allocator fallback.

mooncake-store/src/transfer_task.cpp
zxpdemonio: Done by removing the `require_same_process` batch path. Local write bypass now happens in `TransferSubmitter::submit()` only after the allocated memory replica is known to be local to the current process, so there is no separate `require_same_process` API/documentation behavior anymore.

zxpdemonio (commented):
## Re-review (head `a242517`)

Thanks for narrowing the change and addressing the earlier review points:

- Removed `MakeDefaultLocalTensorWriteConfig` / auto-flip of `prefer_alloc_in_same_node`
- Metadata slice now uses `header.data_offset`
- put/upsert share `batch_*_from_multi_buffers`
- Added default-config + CUDA batch coverage

## Blocker

**Default `put_tensor` / `upsert_tensor` no longer stage into the registered client buffer, and there is no staging fallback.**

`batch_write_tensor_impl` now passes:

1. stack `TensorMetadata` (`&info.metadata`)
2. raw `tensor.data_ptr`

directly into `BatchPut` / `BatchUpsert`.

That is correct **only** when `TransferSubmitter` selects `LOCAL_MEMCPY` (same-process dest + `MC_STORE_MEMCPY` enabled). Under RDMA, `MC_STORE_MEMCPY` defaults to **off**, so writes go through TransferEngine. RDMA then requires the source address to be in a registered MR (`RdmaTransport`: `"Memory region not registered by any active device(s)"`). The old staging path allocated from `client_buffer_allocator_`, which is registered at setup; the new path is not.

So with the current head:

| Path | Likely result |
| --- | --- |
| Local TCP + auto memcpy | Works (matches wheel tests) |
| Local RDMA + default `MC_STORE_MEMCPY` | Likely fails / cannot reproduce the claimed RDMA gains without forcing memcpy |
| Remote / non-local replica | Likely fails for default tensor put/upsert |

This also contradicts the PR description (“Unsupported or non-local paths still keep the existing staging fallback”).

Please either:

1. **Restore a staging fallback** when the write cannot use local memcpy (or when TE registration lookup would fail), or
2. **Gate the no-staging path** behind an explicit same-process / memcpy-capable condition and keep the old registered staging path otherwise, and
3. **Update the PR body** to match the real semantics (and note that the published RDMA numbers likely need `MC_STORE_MEMCPY=1` or an equivalent force-memcpy path that is no longer in this diff).

## Non-blocking

- `DummyClient::batch_upsert_from_multi_buffers` still absent; fine if intentionally out of scope, but worth a one-line comment near the RealClient API.
- Tests still only cover the local same-process happy path. A negative/regression test for `PROTOCOL=rdma` + unset `MC_STORE_MEMCPY` (or a unit-level assert that unregistered sources are rejected/fallback) would lock the intended behavior.
- Checklist marks Docs updated, but this diff has no doc files; either add a short note under tensor put semantics or uncheck.

## Verdict

Direction is good after the narrow, but **I would not merge until the RDMA / non-local registered-buffer regression is fixed or explicitly scoped with fallback + accurate docs.**

yokinoshitayoki:
@zxpdemonio It seems that this PR has conflicts that must be resolved.

zxpdemonio:
> @zxpdemonio It seems that this PR has conflicts that must be resolved.

@yokinoshitayoki  The rebase conflict has been resolved.

The conflict was caused by the CUDA IPC support that recently landed upstream. After rebasing on top of it, we wired the dummy tensor write path to use the CUDA IPC handoff as well.

With this change, same-node dummy-client GPU tensor writes can avoid the dummy/client local staging buffer: the dummy client exports the CUDA tensor via CUDA IPC, and the real client maps that payload and writes it directly into the local Store segment. I also added CUDA coverage for dummy upsert_tensor and verified the path with client metrics showing batch_upsert_from_cuda_ipc.

mooncake-store/src/real_client.cpp
```diff
@@ -104,6 +104,76 @@ bool ReleasePinnedRegionsForFree(
...
+using BatchWriteMethod = std::vector<tl::expected<void, ErrorCode>> (Client::*)(
+    const std::vector<ObjectKey> &, std::vector<std::vector<Slice>> &,
+    const ReplicateConfig &);
+
+ReplicateConfig PreferLocalWriteSegment(const std::shared_ptr<Client> &client,
+                                        const ReplicateConfig &config) {
+    ReplicateConfig write_config = config;
+    if (client && write_config.preferred_segment.empty() &&
+        write_config.preferred_segments.empty()) {
+        write_config.preferred_segment = client->GetSegmentEndpoint();
```
yokinoshitayoki: Could we reuse or extend the host-aware local_first allocation machinery introduced in #2638 instead of introducing a second locality policy in RealClient?

zxpdemonio: Yes. I removed the RealClient-side PreferLocalWriteSegment policy. prefer_alloc_in_same_node now activates the existing master-side host-aware local-first allocation machinery from #2638, so it reuses the client host ID, host-ordered segment selection, explicit preferred-segment precedence, and normal remote fallback. RealClient now passes the caller config through unchanged. I also added a request-level placement test with the master global strategy left at its default.

yokinoshitayoki:

zxpdemonio:
Thanks for the reminder. I checked the failed jobs. They were caused by EPEL mirror metadata download errors (404s and timeouts), not by the PR code. I reran the failed jobs, and the full workflow is now green: https://github.com/kvcache-ai/Mooncake/actions/runs/31068240815. No source changes were needed.

zxpdemonio:
Update summary for the latest head (`a85ec52`):

- Kept the default tensor API allocation semantics unchanged. The no-staging write path is opt-in through `prefer_alloc_in_same_node`; allocation now reuses the existing host-aware local-first allocator and retains its ordered remote fallback.
- Removed the client-side segment override. After allocation, direct copy is used only when the selected Store segment is eligible for the same-process local-memcpy path.
- Routed single and batch `put_tensor` / `upsert_tensor` writes through the shared multi-buffer implementation, avoiding duplicated staging logic.
- Completed DummyClient CUDA IPC support for tensor upsert, so dummy and real clients expose the same put/upsert behavior.
- Added coverage for host-aware placement and CUDA tensor write paths, and updated the API documentation to match the final behavior.

The current Linux workflow, including format, builds, wheel tests, and `CI Gate`, is green: https://github.com/kvcache-ai/Mooncake/actions/runs/31068240815

yokinoshitayoki:
@zxpdemonio Seems this branch has conflicts that must be resolved, could you take a look?

zxpdemonio:
Supplemental DummyClient tensor-path validation

The existing performance tables in the PR description use a real client. This is an additional end-to-end measurement for the tensor + DummyClient path.

Environment:

- PR head: `a85ec526`
- DummyClient process -> standalone real client on the same node
- CUDA IPC source handoff
- pinned Store segment: 4 GiB
- `MC_STORE_MEMCPY=1`
- protocol: TCP
- client local buffer: 64 MiB
- tensor sizes: 64/256/512 MiB
- warmup/iterations: 3/8
- ready CUDA tensors are explicitly synchronized before timing; asynchronous current-stream readiness is tracked separately by #3303
- external one-shot harness, not included in this PR

| API | Size | Throughput | Median latency | Readback |
| --- | ---: | ---: | ---: | --- |
| `put_tensor(cuda)` | 64 MiB | 22.730 GB/s | 2.952 ms | passed |
| `put_tensor(cuda)` | 256 MiB | 25.267 GB/s | 10.624 ms | passed |
| `put_tensor(cuda)` | 512 MiB | 25.735 GB/s | 20.862 ms | passed |
| `upsert_tensor(cuda)` | 64 MiB | 22.757 GB/s | 2.949 ms | passed |
| `upsert_tensor(cuda)` | 256 MiB | 25.292 GB/s | 10.614 ms | passed |
| `upsert_tensor(cuda)` | 512 MiB | 25.741 GB/s | 20.857 ms | passed |

Each write was followed outside the timed region by a DummyClient CUDA readback and payload comparison. The real-client log also confirmed that the 4 GiB Store segment was successfully registered as pinned memory. The 256 and 512 MiB tensors both exceed the 64 MiB client local buffer and still complete, confirming that the payload does not depend on a full-tensor client staging buffer.

zxpdemonio:
Rebased this PR onto the latest main, including the changes from #3197, and reran the same-node dummy-client CUDA tensor write benchmark.

Configuration: MC_STORE_MEMCPY=1, 4 GiB pinned Store segment, 64 MiB client local buffer, CUDA device source. Every write was read back and verified.

| Size | put_tensor | upsert_tensor |
|---:|---:|---:|
| 64 MiB | 22.63 GB/s | 22.83 GB/s |
| 256 MiB | 25.30 GB/s | 25.32 GB/s |
| 512 MiB | 25.79 GB/s | 25.80 GB/s |

The 256 MiB and 512 MiB tensors are larger than the 64 MiB client local buffer and still sustain about 25 GB/s, confirming that the same-node CUDA IPC/local-memcpy path remains no-staging after the rebase.

zxpdemonio:
@yokinoshitayoki This is the last remaining PR for the same-node tensor put optimization. It has been rebased onto the latest main, the dummy-client CUDA put/upsert performance and readback have been revalidated, and the full CI is green. Could you please take another look when you have time? Thanks!

yokinoshitayoki:
@zxpdemonio It seem that this branch has conflicts that must be resolved. Could you take a look first?

mooncake-integration/store/store_py_parallel_write.h
```diff
@@ -97,73 +146,109 @@ std::vector<int> batch_write_tensor_impl(const std::vector<std::string> &keys,
...
+                    MakeIndexedConfig(config, original_indices);
+                std::vector<int> op_results = staged_batch_write(
+                    valid_keys, buffer_ptrs, buffer_sizes, write_config);
+                apply_tensor_batch_results(results, original_indices,
+                                           op_results, operation_name);
+            }
+            return results;
+        };
+
+        if (use_dummy_client_ || !config.prefer_alloc_in_same_node)
```
yokinoshitayoki: The tensor path skips staging based only on the requested prefer_alloc_in_same_node flag, before allocation returns the actual replica. preferred_segment is best-effort and can fall back to a remote segment when the local segment is missing or full. In that case raw tensor pointers are still sent without the registered staging buffer. Please gate direct buffers on the allocated replica being locally accessible and otherwise retry through run_staged_write.

After resolving this issue and the conflict with the main, I think we can merge this PR.

zxpdemonio:
@yokinoshitayoki Rebased onto the latest main (360044c5) and resolved the conflicts. The resolution preserves the upstream DFS write/finalization semantics while keeping the shared same-node put/upsert path from this PR.

I also reran format, the focused regression tests, and the dummy-client CUDA put/upsert benchmark. Build and tests passed; 256/512 MiB writes remain at about 25.0/25.6 GB/s with readback verification. The branch is mergeable now, and a fresh CI run has started.

ykwd (approved):
Thanks for the great work. We can merge this after all the CI passed

mooncake-integration/store/store_py_parallel_write.h
zxpdemonio: Fixed in f4186527. Tensor writes now preserve the caller allocation policy and wait for StartBatchPut/StartBatchUpsert to return the actual replicas. Direct tensor/CUDA IPC slices are used only when every allocated target is locally addressable by this process and MC_STORE_MEMCPY is enabled; otherwise the payload is lazily copied into the registered client buffer and submitted in the same transaction. This covers both default and host-preferred put/upsert paths without retrying allocation or changing placement semantics. Added regression coverage for local direct writes, non-local staging, put/upsert, and MC_STORE_MEMCPY on/off.

yokinoshitayoki (approved):
Thanks a lot for your contribution, this PR looks good to me now.

stmatengss:
lgtm thx

stmatengss (approved):

Issue #3052, opened 2026-07-22
[Bug]: client 重启后，  ssd 上的数据全部失效


sun-baoshan-sunshine:
### Bug Report
mooncake version： v0.3.11.post1

**1. Background**
When a client with SSD offload enabled crashes unexpectedly and is subsequently restarted in a short timeframe, the SSD data previously held by the client becomes stale/invalid, rendering it inaccessible.
**2. Analysis**
Under normal circumstances, upon restart, the client is expected to scan the data residing on the SSD and report the metadata to the Master via ScanMeta. However, if the new client's registration/reporting occurs within the TTL (Time-To-Live) window of the previous client instance — meaning the old client has not yet been evicted by the Master — the new client's registration attempt is effectively a no-op and is rejected. The root cause of this issue resides in the logic of auto MasterService::AddReplica(const UUID& client_id, const std::string& key, Replica& replica).
```c++
auto MasterService::AddReplica(const UUID& client_id, const std::string& key,
                               Replica& replica)
    -> tl::expected<void, ErrorCode> {
    std::shared_lock<std::shared_mutex> shared_lock(snapshot_mutex_);
    MetadataAccessorRW accessor(this, key);
    if (!accessor.Exists()) {
        accessor.Create(
            client_id,
            replica.get_descriptor().get_local_disk_descriptor().object_size,
            std::vector<Replica>{}, false);
    }
    auto& metadata = accessor.Get();
    if (replica.type() != ReplicaType::LOCAL_DISK) {
        LOG(ERROR) << "Invalid replica type: " << replica.type()
                   << ". Expected ReplicaType::LOCAL_DISK.";
        return tl::make_unexpected(ErrorCode::INVALID_PARAMS);
    }

    if (!metadata.HasReplica(&Replica::fn_is_local_disk_replica)) {
        std::vector<Replica> replicas;
        replicas.emplace_back(std::move(replica));
        metadata.AddReplicas(std::move(replicas));
        return {};
    }

    metadata.VisitReplicas(
        [client_id](const Replica& rep) {
            return rep.type() == ReplicaType::LOCAL_DISK &&
                   rep.get_descriptor().get_local_disk_descriptor().client_id ==
                       client_id;
        },
        [&replica](Replica& rep) {
            rep.get_descriptor()
                .get_local_disk_descriptor()
                .transport_endpoint = replica.get_descriptor()
                                          .get_local_disk_descriptor()
                                          .transport_endpoint;
[...]

PR #3160, opened 2026-07-28
[Store] Fix EraseMetadata leaking stale entries in offloading_objects

huniu20:
## Summary

`MasterService::EraseMetadata` clears `tenant_state.offloading_tasks` but leaves the mirror entry in `LocalDiskSegment::offloading_objects` behind. When a `Remove` / `BatchRemove(force=true)` (or `BatchEvict`, `RemoveByRegex`, lease-expiry cleanup) interposes between `PushOffloadingQueue` and the next `OffloadObjectHeartbeat`, the heartbeat drains a task-less key back to the client. The client writes a bucket and calls `NotifyOffloadSuccess` — resurrecting a `LOCAL_DISK`-only metadata entry, or leaking an orphan bucket on disk.

This is the same "phantom metadata persists in steady state" pattern @LujhCoconut described in #2827, but on the per-key eraser side.

## Relation to open work

- **#2676** (@Colors-111) — same class of orphan, scoped to `RemoveAll`; the per-key paths (`Remove` / `BatchRemove` / `BatchEvict` / lease-expiry) all still leak. This PR closes that gap.
- **#2827** (@he-yufeng, @LujhCoconut) — this PR removes one concrete producer of `OBJECT_ALREADY_EXISTS` on re-offload of a just-removed key; the worker-side dedup path discussed there is complementary.
- **#3009** (@bestdo77) — orthogonal, targets the worker-side ungrouped pool. Both fixes are needed independently.

## Fix

11 lines in `EraseMetadata`: after `offloading_tasks.erase(...)`, erase the same key from every client's `offloading_objects` map. Iteration is bounded by mounted local-disk segments (typ. 1..64) and is a hash-map no-op on segments that don't hold the key.

## Regression test

`OffloadOnEvictTest.BatchRemoveDropsOffloadingObjectsMirror`:
- Without fix: heartbeat drains **3** stale entries after `BatchRemove`.
- With fix: **0**.

## Testing

- `offload_on_evict_test`: 10/10 ✅ (9 pre-existing + 1 new regression)
- `master_service_test`: 145/145 ✅
- `master_service_ssd_test`: 18/18 ✅
- `clang-format-20`: clean

+46 lines total. No API changes.

### DIFF
--- a/mooncake-store/src/master_service.cpp
+++ b/mooncake-store/src/master_service.cpp
@@ -1888,6 +1888,17 @@ MasterService::EraseMetadata(
             source->dec_refcnt();
         }
         tenant_state.offloading_tasks.erase(offload_it);
+
+        // Mirror entry in local_disk_segment.offloading_objects must be
+        // dropped too, otherwise the next OffloadObjectHeartbeat drains a
+        // task-less key back to the client and produces an orphan bucket.
+        const std::string scoped_key = tenant_id.MakeScopedKey(key);
+        ScopedLocalDiskSegmentAccess ssd_access =
+            segment_manager_.getLocalDiskSegmentAccess();
+        for (auto& [_, segment] : ssd_access.getClientLocalDiskSegment()) {
+            MutexLocker locker(&segment->offloading_mutex_);
+            segment->offloading_objects.erase(scoped_key);
+        }
     }
     tenant_state.processing_keys.erase(key);
     tenant_state.replication_tasks.erase(key);
--- a/mooncake-store/tests/offload_on_evict_test.cpp
+++ b/mooncake-store/tests/offload_on_evict_test.cpp
@@ -389,6 +389,41 @@ TEST_F(OffloadOnEvictTest, ComboD_EvictionWorks) {
     service->RemoveAll();
 }
 
+// Regression: EraseMetadata must drop the mirror entry in
+// LocalDiskSegment::offloading_objects. Otherwise BatchRemove leaves a
+// task-less key in the offload queue which the next
+// OffloadObjectHeartbeat drains back to the client, producing an
+// orphan bucket on SSD.
+TEST_F(OffloadOnEvictTest, BatchRemoveDropsOffloadingObjectsMirror) {
+    MasterServiceConfig config;
+    config.enable_offload = true;
+    config.default_kv_lease_ttl = 0;  // no lease so BatchRemove succeeds
+    auto service = std::make_unique<MasterService>(config);
+
+    constexpr size_t seg_size = 1024 * 1024 * 16;
+    auto ctx =
+        PrepareSegment(*service, "test_segment", kDefaultSegmentBase, seg_size);
+    auto mount_ld = service->MountLocalDiskSegment(ctx.client_id, true);
+    ASSERT_TRUE(mount_ld.has_value());
+
+    const std::vector<std::string> keys = {"key_r1", "key_r2", "key_r3"};
+    for (const auto& k : keys) {
+        PutObject(*service, ctx.client_id, k);
+    }
+
+    // Remove before heartbeat drains the offload queue.
+    auto rm = service->BatchRemove(keys, TenantId::Default(), /*force=*/true);
+    for (const auto& r : rm) {
+        EXPECT_TRUE(r.has_value());
+    }
+
+    auto queued = DrainOffloadQueue(*service, ctx.client_id);
+    EXPECT_TRUE(queued.empty())
+        << "OffloadObjectHeartbeat returned " << queued.size()
+        << " stale entries after BatchRemove; EraseMetadata failed to clean "
+           "offloading_objects.";
+}
+
 }  // namespace mooncake::test
 
 int main(int argc, char** argv) {


he-yufeng:
Checked the two invariants this has to keep, both hold:

1. Lock order. The documented order is shard lock -> offloading_mutex_ (the comment in OffloadObjectHeartbeat releases the mutex before taking shard locks for exactly this). EraseMetadata always runs under a MetadataShardAccessorRW, so the new segment-access + offloading_mutex_ nesting keeps the direction. RemoveAll is the only place that grabs the segment access before shards and it does so in a separate, already-released phase, so no reverse nesting exists for this to collide with.
2. Key form. The mirror is inserted under tenant_id.MakeScopedKey(user_key) at PushOffloadingQueue and the snapshot deserializer preserves scoped keys, so erasing by MakeScopedKey(key) hits the right entry.

One thing to sanity check before merge: EraseMetadata sits on the eviction path, and the cleanup now walks every local disk segment under the segment access lock for each erased key that had an offload in flight. It is gated behind offload_it != end, so plain evictions skip it, but with many mounted disk segments an eviction burst of offloaded keys pays O(segments) mutex acquisitions per key. Probably fine at realistic segment counts, just worth confirming that is the intended tradeoff since OffloadTaskItem does not record the target segment (scanning is the only correct option without extending the task).

huniu20:
Thanks @he-yufeng — confirming the O(segments) pass is intended:

1. **Gated on `offload_it != end`**, so plain evictions skip it entirely; only keys mid-offload pay the cost.
2. **Bounded fan-out** — mounted disk clients are 1..64 in realistic deployments.
3. **Alternative rejected on purpose**: adding a `target_client_id` to `OffloadTaskItem` would enlarge the hot-path record and introduce a new cross-map invariant — precisely the bug class this PR fixes. Happy to revisit if profiling on a many-segment cluster shows it matters.

> **CI note**: `Build Docker Image` hit the 62-min GitHub job timeout on `docker build` (no code paths this PR touches); `CI Gate` cascades from that. All other 23 checks are green.

LujhCoconut:
Discussion & Suggestion (Non-blocking)
* consider moving the `offloading_objects` cleanup out of the `offload_it != end` branch, so the mirror erase is unconditional.

The gate decides whether to clean the mirror based on the *other* map's state — but the two maps being able to diverge is exactly the bug class this PR fixes. When they're in sync, an unconditional erase is a hash-map no-op; when they're out of sync (mirror present, task already gone), the gated version skips the one entry that actually needs cleaning. Today no code path produces that divergence, so this is defense-in-depth rather than a live bug — but the unconditional version is strictly more convergent and makes `EraseMetadata` self-healing against any future path (e.g. the parallel work in #2676 / #2827) that forgets to dual-write again.

The extra cost is bounded — one shared segment access plus a no-op `offloading_mutex_` acquisition per mounted disk segment (typ. 1..64) per erased key. If the per-key overhead on plain evictions is a concern, a middle ground would be keeping the gate but adding a debug assertion that no mirror entry exists when the task doesn't.

Happy to discuss further. Non-blocking.

LujhCoconut (approved):
Just ignore the above comment. After thinking it through carefully, the current gated implementation (cleaning the mirror only inside the `offload_it != end` branch) is reasonable as-is.  

Because:

1. **The divergence scenario does not exist today**: no code path produces a "mirror present, task gone" state, so an unconditional erase would have nothing real to fix — it only defends against a hypothetical future bug (YAGNI).
2. **The gate is a valuable fast-path**: plain evictions (the vast majority of key deletions, with no offload in flight) pay zero overhead under the gated version; an unconditional erase would make every key deletion take the global `segment_mutex_` and iterate all disk segments acquiring locks — real hot-path cost in large BatchEvict bursts.
3. **The cost-benefit does not hold**: trading a currently-true fact (the two maps do not diverge) for fixed hot-path overhead is the right engineering call; "logically stricter" is not the same as "better engineering".


So LGTM.

huniu20:
Thanks for the second look, @LujhCoconut — agreed. YAGNI + keeping the eviction hot-path allocation-free is the right trade-off here. Sticking with the gated version.

fcczzz (approved):
The fix in this PR looks correct for the queued-but-not-yet-dispatched case, so I don’t think this should block the merge. 

One related ordering I couldn’t rule out is: after OffloadObjectHeartbeat has drained the entry from offloading_objects and the worker has read the source object, Remove or BatchRemove(force=true) erases the metadata and offloading_tasks, followed by a late NotifyOffloadSuccess. Would the notification then fall through to AddReplica and potentially recreate metadata for the removed key?

My understanding is that #2818 provides rollback in the bucket backend if the notification is rejected, while simply rejecting notifications without an active task could interfere with the restart re-registration path in #3052, which uses the same RPC. Am I missing an existing guard that prevents this ordering? If not, would it be worth tracking separately, perhaps with a task ID/generation or an explicit distinction between normal completion and re-registration?

huniu20:
Thanks @fcczzz — verified: `AddReplica` at master_service.cpp:3843–3848 calls `accessor.Create()` on a missing key, so a late `NotifyOffloadSuccess` after `EraseMetadata` does resurrect the metadata (source refcount is already released by then). Out of scope for this PR; this overlaps with #2827 / #2818 / #3052 and is probably best discussed there.

ykwd (approved):

PR #3161, opened 2026-07-28
[TransferEngine] Refactor RDMA worker ownership to avoid endpoint UAF

alogfans:
## Description

This PR addresses a correctness issue reported after #2696. The previous RDMA polling/worker changes exposed an endpoint lifetime race when multiple worker threads are enabled: different workers could concurrently operate on the same RDMA endpoint lifecycle, including post-send, completion failure handling, retry, redispatch, and endpoint deletion. That could lead to use-after-free risks when one worker was still using an endpoint while another worker retired or deleted it.

This PR refactors RDMA worker dispatch around a stable owner model:

```text
peer_nic_path -> owner worker -> owner CQ -> endpoint lifecycle
```

Main changes:
- Replace the shared RDMA slice shard queues with per-worker owner queues.
- Route submitted and redispatched slices to the stable owner worker for their `peer_nic_path`.
- Bind endpoint creation to the owner worker’s CQ.
- Make each worker poll only its owned CQ and process completions locally.
- Keep endpoint completion handling, retry, redispatch, and endpoint deletion on the owner worker path.
- Ensure RDMA contexts have at least one CQ per worker for worker-owned endpoint polling.
- Change endpoint reclaim timeout behavior: if an endpoint still has outstanding WRs after the timeout, keep the retired endpoint alive in the waiting list instead of force-freeing it. This trades a bounded retired-resource leak for avoiding stale in-flight reference UAF.
- Add endpoint store stress coverage for stale raw pointer lookup/delete APIs to ensure they compare pointer identity without dereferencing stale pointers.

## Module

- [ ] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Breaking change
- [x] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**

```bash
git diff --check
cmake --build build --target endpoint_store_test rdma_endpoint_state_test
build/mooncake-transfer-engine/tests/endpoint_store_test --gtest_color=no
build/mooncake-transfer-engine/tests/rdma_endpoint_state_test --gtest_color=no
```

**Test results:**
- [x] Unit tests pass
- [ ] Integration tests pass (not run; RDMA integration requires suitable RDMA devices/environment)
- [ ] Manual testing done

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used
- [x] AI tools were used

AI assistance was used to analyze the RDMA worker concurrency model, draft the owner-based refactor, and prepare this PR description. The human submitter is responsible for reviewing and validating every changed line.

### DIFF
--- a/mooncake-transfer-engine/src/transport/rdma_transport/rdma_context.cpp
+++ b/mooncake-transfer-engine/src/transport/rdma_transport/rdma_context.cpp
@@ -167,7 +167,6 @@ RdmaContext::RdmaContext(RdmaTransport &engine, const std::string &device_name)
           [] { return static_cast<uint64_t>(getCurrentTimeInNano()); }),
       next_comp_channel_index_(0),
       next_comp_vector_index_(0),
-      next_cq_list_index_(0),
       worker_pool_(nullptr),
       active_(true) {
     static std::once_flag g_once_flag;
@@ -194,8 +193,21 @@ int RdmaContext::construct(size_t num_cq_list, size_t num_comp_channels,
         return ERR_INVALID_ARGUMENT;
     }
 
-    // Create endpoint store based on configuration
     auto &config = globalConfig();
+    if (config.workers_per_ctx <= 0) {
+        LOG(ERROR) << "Invalid workers_per_ctx=" << config.workers_per_ctx
+                   << " for device " << device_name_;
+        return ERR_INVALID_ARGUMENT;
+    }
+    const auto worker_count = static_cast<size_t>(config.workers_per_ctx);
+    if (num_cq_list < worker_count) {
+        LOG(INFO) << "Increasing RDMA CQ count for " << device_name_
+                  << " from " << num_cq_list << " to " << worker_count
+                  << " to keep each transfer worker on a dedicated CQ";
+        num_cq_list = worker_count;
+    }
+
+    // Create endpoint store based on configuration
     switch (config.endpoint_store_type) {
         case EndpointStoreType::FIFO:
             endpoint_store_ =
@@ -725,21 +737,8 @@ RdmaContext::findMemoryRegionContaining(uintptr_t addr) const {
 
 std::shared_ptr<RdmaEndPoint> RdmaContext::endpoint(
     const std::string &peer_nic_path) {
-    if (cq_list_.empty()) {
-        LOG(ERROR) << "No CQ available for endpoint on " << deviceName();
-        return nullptr;
-    }
-
-    const auto &config = globalConfig();
-    if (config.workers_per_ctx <= 0) {
-        LOG(ERROR) << "Invalid workers_per_ctx=" << config.workers_per_ctx
-                   << " for endpoint on " << deviceName();
-        return nullptr;
-    }
-    int owner_thread =
-        static_cast<int>(std::hash<std::string>{}(peer_nic_path) %
-                         static_cast<size_t>(config.workers_per_ctx));
-    int cq_index = owner_thread % static_cast<int>(cq_list_.size());
+    int cq_index = cqIndexForPeer(peer_nic_path);
+    if (cq_index < 0) return nullptr;
     return endpoint(peer_nic_path, cq_index);
 }
 
@@ -765,6 +764,7 @@ std::shared_ptr<RdmaEndPoint> RdmaContext::endpoint(
         return endpoint;
     }
 
+    endpoint_store_->reclaimEndpoint();
     endpoint =
         endpoint_store_->insertEndpoint(peer_nic_path, this, cq(cq_index));
     endpoint_store_->reclaimEndpoint();
@@ -849,15 +849,33 @@ int RdmaContext::gidIndex() const {
     return gid_index_;
 }
 
-ibv_cq *RdmaContext::cq() {
-    int index = (next_cq_list_index_++) % cq_list_.size();
-    return cq_list_[index].native;
-}
-
 ibv_cq *RdmaContext::cq(int cq_index) {
     if (cq_list_.empty()) return nullptr;
-    if (cq_index < 0) return cq();
+    if (cq_index < 0) return nullptr;
     return cq_list_[static_cast<size_t>(cq_index) % cq_list_.size()].native;
+}
+
+int RdmaContext::postingThreadForPeer(
+    const std::string &peer_nic_path) const {
+    const auto &config = globalConfig();
+    if (config.workers_per_ctx <= 0) {
+        LOG(ERROR) << "Invalid workers_per_ctx=" << config.workers_per_ctx
+                   << " for endpoint on " << deviceName();
+        return -1;
+    }
+    return static_cast<int>(std::hash<std::string>{}(peer_nic_path) %
+                            static_cast<size_t>(config.workers_per_ctx));
+}
+
+int RdmaContext::cqIndexForPostingThread(int thread_id) const {
+    const int cq_count = cqCount();
+    if (cq_count <= 0 || thread_id < 0) return -1;
+    if (thread_id >= 0 && thread_id < cq_count) return thread_id;
+    return thread_id % cq_count;
+}
+
+int RdmaContext::cqIndexForPeer(const std::string &peer_nic_path) const {
+    return cqIndexForPostingThread(postingThreadForPeer(peer_nic_path));
 }
 
 ibv_comp_channel *RdmaContext::compChannel() {
--- a/mooncake-transfer-engine/src/transport/rdma_transport/rdma_context.cpp
+++ b/mooncake-transfer-engine/src/transport/rdma_transport/rdma_context.cpp
@@ -27,6 +27,7 @@
 #include <cassert>
 #include <exception>
 #include <fstream>
+#include <functional>
 #include <memory>
 #include <thread>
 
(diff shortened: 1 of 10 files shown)


UNIDY2002:
@alogfans Please merge/rebase with the latest main to include the hot-fix commit a6b4db4cfa371a3fc7f189a19977c102e14c4dfc

mooncake-transfer-engine/src/transport/rdma_transport/rdma_context.cpp
```diff
@@ -727,7 +753,8 @@ std::shared_ptr<RdmaEndPoint> RdmaContext::endpoint(
         return endpoint;
     }
 
-    endpoint = endpoint_store_->insertEndpoint(peer_nic_path, this);
+    endpoint =
+        endpoint_store_->insertEndpoint(peer_nic_path, this, cq(cq_index));
```
Primary33, line 769: This correctly binds the endpoint’s initial QPs to the owner worker’s CQ, but I think the reconstruction path can break that ownership. RdmaEndPoint::reconstruct() still calls context_.cq(), which selects a CQ in round-robin order instead of reusing the CQ assigned here.
resetConnection() can call reconstruct() after a handshake failure or timeout. The rebuilt QPs may then send completions to another worker’s CQ, which can trigger the ownership assertion in performPollCq(). If num_cq_per_ctx is greater than workers_per_ctx, it may even select a CQ that no worker polls, leaving the completions stuck.

alogfans: Good catch. The initial endpoint creation now binds QPs to the owner worker’s CQ, so `reconstruct()` must preserve that CQ instead of calling `context_.cq()` again. I updated `RdmaEndPoint` to remember the CQ passed to `construct()` and reuse it during reconstruction, keeping completions on the same worker-owned CQ after reset/retry.

mooncake-transfer-engine/src/transport/rdma_transport/worker_pool.cpp
```diff
@@ -564,16 +537,57 @@ void WorkerPool::performPollCq(int thread_id) {
...
+        else
+            qp_depth_set[slice->rdma.qp_depth] = 1;
+        wc_list.push_back(wc[i]);
+    }
+    if (nr_poll)
+        context_.cqOutstandingCount(cq_index)->fetch_sub(
+            nr_poll, std::memory_order_acq_rel);
+
+    for (auto &entry : qp_depth_set)
+        entry.first->fetch_sub(entry.second, std::memory_order_acq_rel);
```
Primary33, line 574: I think this ordering reopens an endpoint lifetime race. `finishDestroy()` uses the per-QP WR depth as the reclaim gate, but we decrement all of those counters before `processCompletions()` has finished using `slice->rdma.endpoint`.

If the endpoint was already retired, or an earlier WC in this batch retires it, the monitor thread can now reclaim and free it while the rest of the batch is still being processed. If the allocator then reuses the same address for a new endpoint, a later error WC may pass the stale pointer to `deleteEndpointByPtr()`. Since that API only compares pointer values, it could match and retire the new endpoint instead—an ABA issue rather than a direct stale-pointer dereference.

The new stale-pointer test does not cover this case because it does not allocate another endpoint after reclaim, so none of the stale addresses can be reused.

Primary33 (commented):
One concern about the completion handling order:

mooncake-transfer-engine/src/transport/rdma_transport/worker_pool.cpp
alogfans: I added a lightweight completion-batch pin on RdmaEndPoint:
- WorkerPool::performPollCq() now pins each unique endpoint in the polled WC batch before decrementing WR depth.
- RdmaEndPoint::finishDestroy() now refuses reclaim while completion refs are nonzero.
- Pins are released only after the whole WC batch has gone through processCompletions().
- Added EndpointStoreTest.ReclaimLeavesCompletionPinnedEntries to lock the reclaim gate behavior down.

mooncake-transfer-engine/src/transport/rdma_transport/rdma_endpoint.cpp
```diff
@@ -294,22 +298,29 @@ bool RdmaEndPoint::finishDestroy() {
...
+        // destruction while stale slice references may still exist.
         bool has_outstanding = false;
         for (size_t i = 0; i < qp_list_.size(); ++i) {
             if (wr_depth_list_[i].load(std::memory_order_relaxed) != 0) {
                 has_outstanding = true;
                 break;
             }
         }
-        if (has_outstanding) {
+        if (has_outstanding ||
```
staryxchen, line 316: The reclaim-timeout path now leaks retired endpoints permanently, and nothing bounds the waiting list — the limit check at endpoint_store.cpp:57 logs a warning but then continues creation. Under a NIC where ibv_modify_qp-to-ERR persistently fails, QPs accumulate until the device can't create new endpoints. Since this replaces  force-free (UAF risk) with an intentional leak, could we (a) correct the "bounded retired-resource leak" wording in the description, and (b) add a real bound or escalation, e.g. disable the context after N leaked endpoints, or re-log periodically instead of once?

mooncake-transfer-engine/src/transport/rdma_transport/rdma_context.cpp
```diff
@@ -712,6 +713,31 @@ RdmaContext::findMemoryRegionContaining(uintptr_t addr) const {
...
+
+    const auto &config = globalConfig();
+    if (config.workers_per_ctx <= 0) {
+        LOG(ERROR) << "Invalid workers_per_ctx=" << config.workers_per_ctx
+                   << " for endpoint on " << deviceName();
+        return nullptr;
+    }
+    int owner_thread =
+        static_cast<int>(std::hash<std::string>{}(peer_nic_path) %
+                         static_cast<size_t>(config.workers_per_ctx));
```
staryxchen: The owner→CQ mapping is now encoded in three places: RdmaContext::endpoint() (one-arg, rdma_context.cpp:726-730), postingThreadForPeer(), and  cqIndexForPostingThread(). They agree today, but a future edit to any one of them silently breaks the invariant that each CQ is polled by exactly its owner — in release builds the only symptom is transfers hanging (the assert at worker_pool.cpp:564 is compiled out). Could this be centralized into a single helper on RdmaContext (e.g. cqIndexForPeer(peer_nic_path)), with the worker pool delegating to it?

mooncake-transfer-engine/src/transport/rdma_transport/rdma_context.cpp
```diff
@@ -815,6 +842,12 @@ ibv_cq *RdmaContext::cq() {
     return cq_list_[index].native;
 }
```
staryxchen: dead code now

mooncake-transfer-engine/src/transport/rdma_transport/worker_pool.cpp
```diff
@@ -564,16 +538,66 @@ void WorkerPool::performPollCq(int thread_id) {
         }
     }
 
-    // Slices this loop drove to a terminal state, successes and
+    const static size_t kPollCount = 64;
+    std::unordered_map<std::atomic<int> *, int> qp_depth_set;
+    std::unordered_set<RdmaEndPoint *> pinned_endpoints;
+    std::vector<ibv_wc> wc_list;
```
staryxchen: Minor perf nit on the completion hot path: `wc_list` (worker_pool.cpp:544) is a fresh heap vector whose only purpose is to hold a copy of the stack array `wc[kPollCount]` that `context_.poll()` just filled — and it is built via per-entry `push_back` (line 570), so a full 64-entry poll incurs ~10 reallocations on every poll iteration, and the vector is then passed once to `processCompletions()` (line 581) which only reads it.

Since the completions are already contiguous in `wc`, the copy adds nothing. Would it be simpler to change `processCompletions` to take `(const ibv_wc *wc, int count)` (or a `std::span`) and drop `wc_list` entirely? If keeping the vector is preferred for API reasons, at least `wc_list.reserve(kPollCount)` would avoid the growth cascade.

mooncake-transfer-engine/src/transport/rdma_transport/rdma_endpoint.cpp
alogfans: Thanks, agreed. I corrected the wording so it no longer describes this as a “bounded retired-resource leak”.

I also reconsidered the fail-closed option here. Returning nullptr once the retired waiting list reaches max_size_ would bound resource growth, but it also turns a degraded NIC/resource condition into immediate endpoint creation failure, which is a fairly high availability risk. I’m leaving the existing “log and continue creation” behavior unchanged in this PR.

This PR keeps the safer UAF-avoidance behavior: timed-out retired endpoints stay alive rather than being force-freed. The remaining risk is now described explicitly as intentional retention under persistent reclaim failure, not as a bounded leak. A stronger escalation policy, such as disabling the context after N retained endpoints, seems worth doing as a separate follow-up with a clearer failure-mode discussion.

mooncake-transfer-engine/src/transport/rdma_transport/worker_pool.cpp
alogfans: Done. I changed processCompletions() to take a raw completion array plus count.

Issue #3157, opened 2026-07-28
[RFC]: Lock-Free Tenant Quota Charge/Release with Unified Charged Bytes


Lin-z-w:
### Summary

This RFC proposes replacing Mooncake Store's per-tenant `used_bytes` and `reserved_bytes` quota-accounting state with one atomic `charged_bytes` counter.

Memory is charged when Put, Copy, Move, or promotion admission starts. A successful operation keeps the charge, while a failed, revoked, partially completed, or expired operation refunds the unused portion. Completion no longer moves bytes from a global reserved counter to a global used counter.

The tenant quota control plane remains lock-based. Policy persistence, effective-quota recomputation, tenant registration, snapshots, and administrative mutation continue to use the existing control-plane locks. Only the high-frequency charge and release operations become independent of the quota-table shard mutex.

In this RFC, "lock-free" describes only the tenant quota charge/release primitive. A complete Put, Copy, Move, or promotion still uses its existing metadata, allocator, and task locks. Move is included because it shares the same `ReplicationTask` quota-settlement path as Copy; this RFC does not otherwise broaden the lock-free scope.

The proposed design:

- uses one compare-and-swap loop for quota admission;
- keeps operation-local `pending_charge_bytes` for exact rollback;
- stores a stable quota-account handle in each metadata `TenantState`;
- blocks new admissions through a bit in the same atomic charge state;
- preserves strict policy-deletion semantics, including zero-memory objects;
- rebuilds charged usage from authoritative metadata after restore;
- replaces the old quota API and metrics with a single charged-byte model; and
- does not provide a dual implementation, compatibility aliases, or downgrade guarantees.

### Motivation

The current quota state distinguishes:

```text
used_bytes      completed memory-replica charge
reserved_bytes  admitted but not yet completed memory-replica charge
```

Admission checks:

```text
used_bytes + reserved_bytes + incoming_bytes <= effective_quota_bytes
```

`Reserve`, `Commit`, `CommitAdditional`, `Abort`, `Release`, and `ReleasePartial` all acquire the tenant quota shard mutex. Although the quota table is sharded, every operation for the same tenant maps to the same shard and therefore serializes on one mutex. A large tenant writing keys distributed across many metadata shards can still be bottlenecked by its single quota mutex.

[...]

PR #3162, opened 2026-07-28
[Store] Make tenant quota charge and release lock-free

Lin-z-w:
## Description

Replace the tenant quota data-plane state machine based on `used_bytes`, `reserved_bytes`, and quota-shard locking with a unified atomic `charged_bytes` account.

The quota control plane remains lock-based. Policy persistence, effective-quota recomputation, tenant registration, snapshots, administrative mutation, and cross-shard operations continue to use the existing locks. Only the high-frequency charge and release paths used by Put, Copy, Move, and promotion bypass the quota-table shard mutex.

The main changes are:

- add `TenantQuotaAccount`, with a stable handle and a CAS-based `TryCharge`/`Release` interface;
- encode the admission gate and charged-byte count in one atomic state, with a policy sequence protecting charges across policy updates;
- keep the non-thread-safe `TenantQuotaTable` as the control-plane table stored within each shard and `ShardedTenantQuotaTable` as the lock-owning cross-shard layer;
- add an independent `TenantQuotaLedger` for per-object pending, committed, and replacement charge ownership;
- bind stable quota handles to metadata-shard tenant state so normal charge/release operations do not perform registry lookup;
- settle Put, Copy, Move, Upsert, and promotion lifecycle changes through the object ledger, while task-local pending charges remain explicit until commit;
- centralize surviving-object settlement after revoke, expiry, and durable finalization so pending and replacement ownership cannot be stranded;
- detect underflow, duplicate release/transfer, `actual > accounted`, rebuild mismatch, and overflow without silently truncating accounting state;
- validate policies against the atomic accounting range `1..2^63-1` and validate complete policy/rebuild inputs before mutating any shard;
- rebuild `charged_bytes` from authoritative metadata after restore;
- replace the old quota metrics and admin fields with `charged_bytes` and `admission_closed`; and
- update the deployment documentation for the breaking quota API and metric changes.

For every live object, the local invariant is:

```text
quota_ledger.pending_bytes
  + quota_ledger.committed_bytes
  + quota_ledger.replaced_bytes
  = bytes in the tenant's charged_bytes still owned by that object
```

This is an intentional breaking quota API and observability change. It does not provide compatibility aliases for `used_bytes`, `reserved_bytes`, `committed_count`, or `metadata_object_count`, and it does not support mixed-version Masters or rolling downgrade.

Related RFC: [#3157: Lock-Free Tenant Quota Charge/Release with Unified Charged Bytes](https://github.com/kvcache-ai/Mooncake/issues/3157).

## Module

- [ ] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [x] Breaking change
- [x] Documentation update
- [x] Performance improvement
- [ ] Other

### Correctness and regression tests

**Test commands:**

\`\`\`bash
  tenant_quota_test \
  tenant_quota_ledger_test \
  master_service_tenant_quota_test \
  promotion_on_hit_test \
  master_service_ha_test \
  mooncake_master \
  -j16

build/mooncake-store/tests/tenant_quota_test
build/mooncake-store/tests/tenant_quota_ledger_test
build/mooncake-store/tests/master_service_tenant_quota_test
build/mooncake-store/tests/promotion_on_hit_test

build/mooncake-store/tests/master_service_ha_test \
  --gtest_filter='MasterServiceBatchRecordE2ETest.RemoveDurableCallbackReleasesResources:MasterServiceBatchRecordE2ETest.ClearInvalidHandlesReleasesResourcesAfterDurable:MasterServiceBatchRecordE2ETest.UpsertStartStaleCleanupReleasesResourcesAfterDurable'

git diff --check \
  f224dcaf18807013b80bc1a1dd67f09dabd80201...bdf5232188625a516cf3d4fb2723f9718ea8143b
\`\`\`

**Test results on current head \`bdf52321\`:**

- \`tenant_quota_test\`: 32/32 passed, including concurrent admission, stable handles, policy updates, all-or-nothing policy/rebuild validation, overflow, and rebuild coverage.
- \`tenant_quota_ledger_test\`: 11/11 passed, covering full and partial settlement, pending refund, additional settlement, partial committed release, replacement transfer/release/rollback, duplicate operation detection, mismatch rollback, and rebuild consistency.
- \`master_service_tenant_quota_test\`: 29/29 passed, covering Put, Upsert, Copy, Move, policy deletion/reload, stable handle binding, partial revoke/expiry settlement, zero-charge admission, capacity recomputation, and orphan cleanup.
- \`promotion_on_hit_test\`: 48/48 passed, including promotion charge, settlement, failure refund, expiry refund, and final release.
- Three targeted HA durable-finalization quota tests passed (3/3).
- A clean \`RelWithDebInfo\` build of the current PR head produced \`mooncake_master\` successfully.
- \`git diff --check f224dcaf...bdf52321\` passed.

### Tenant quota hot-path microbenchmark

The standalone harness isolates the quota state transition itself. It compares the previous mutex-protected \`Reserve + Commit + Release\` lifecycle with the unified \`TryCharge + local settlement + Release\` CAS lifecycle. It models the current \`TenantQuotaAccount\` algorithm, performs no tenant-map lookup in either implementation, and is not included in this PR. The final rebuild-type cleanup in \`bdf52321\` does not change the modeled charge/release algorithm.

Environment and method:

- two AMD EPYC 7742 sockets, 128 physical cores / 256 hardware threads;
- GCC 11.4.0;
- optimized build with \`-O3 -DNDEBUG\`;
- workers pinned to physical CPUs;
- one hot tenant and no simulated non-quota work;
- five alternating mutex/CAS rounds, reporting the median;
- 10 million measured successful lifecycles per implementation per round;
- 10,000 warmup lifecycles per worker; and
- no quota rejection, accounting error, or leaked final charge.

Results rerun on July 30, 2026:

| Threads | Mutex M lifecycle/s | CAS M lifecycle/s | Speedup |
|---:|---:|---:|---:|
| 1 | 25.553 | 81.399 | 3.19x |
| 4 | 9.347 | 22.723 | 2.43x |
| 16 | 5.101 | 17.813 | 3.49x |
| 32 | 5.572 | 17.971 | 3.23x |
| 64 | 5.928 | 17.987 | 3.03x |

An additional 32-thread run measured 30 million successful lifecycles with \`/usr/bin/time -v\`:

| Metric | Mutex | CAS |
|---|---:|---:|
| Benchmark time | 4.42 s | 1.67 s |
| Process wall time | 4.53 s | 1.71 s |
| Lifecycle throughput | 6.785 M/s | 17.926 M/s |
| Voluntary context switches | 431,818 | 18 |
| System CPU time | 136.17 s | 0.18 s |

The CAS run averaged 1.253 retries per lifecycle, completed with zero accounting errors, and left the final charge at zero.

### Master control-plane end-to-end A/B

The end-to-end comparison used two separately built Masters:

- mutex baseline: \`f224dcaf\`, the direct parent of the PR, using \`used_bytes\`/\`reserved_bytes\` and the quota-table shard mutex;
- final CAS candidate: \`bdf52321\`, the current PR head, using stable handles, \`charged_bytes\`, the object ledger, and lock-free quota charge/release.

Both Master binaries used GCC 11.4.0 and the same \`RelWithDebInfo\` configuration (\`-O2 -g -DNDEBUG\`). Each measurement used a fresh Master process with:

- real \`MasterClient -> TCP RPC -> PutStart/BatchPutStart -> allocator/metadata/quota -> PutEnd/BatchPutEnd -> RPC response\` calls;
- the same benchmark client binary for both Masters;
- 64 Master RPC threads;
- eight clients with four worker threads each;
- one registered \`default\` tenant;
- eight 1 GiB memory segments;
- sufficient quota, no eviction, and one-byte values to avoid quota exhaustion;
- four seconds per run; and
- three interleaved runs per implementation, reporting the median.

This measures the complete Master control-plane request path affected by quota. It does not include payload transfer through the Transfer Engine.

| Workload | Mutex runs, successful objects/s | Final CAS runs, successful objects/s | Median result |
|---|---|---|---:|
| Put | 104.6k / 107.4k / 107.1k | 112.9k / 110.7k / 112.3k | +4.84% |
| BatchPut, batch size 32 | 202.6k / 210.0k / 189.6k | 502.5k / 566.4k / 606.0k | 2.80x (+180%) |

A quota-disabled BatchPut control used the same 32 client threads:

| Implementation | Quota disabled | Quota enabled median |
|---|---:|---:|
| Mutex baseline | 609.2k objects/s | 202.6k objects/s |
| Final CAS candidate | 596.0k objects/s | 566.4k objects/s |

The quota-disabled controls differ by 2.16%. The mutex quota implementation reduced BatchPut throughput by 66.7%, while the final CAS result was within 4.96% of its quota-disabled control. All benchmark client operations succeeded, and the Master logs contained no quota rejection, accounting mismatch, runtime error, or fatal error.

The existing \`master_bench\` performs an invalid initial remount against a fresh Master and aborts before a multi-second workload completes. The A/B test therefore used the same temporary benchmark client binary for both Masters with that remount attempt disabled; the per-second Ping remained enabled. This workaround does not change Put or BatchPut request execution, and no benchmark source or build-system change is included in this PR.

## Test Results

- [x] Unit tests pass
- [x] Integration tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex helped implement and review the quota-accounting refactor, add the ledger and regression tests, build a local standalone benchmark, run the correctness and performance validation, and draft this PR description. I reviewed and edited the generated code before submission.


mooncake-store/src/master_service.cpp
```diff
@@ -1640,6 +1627,35 @@ size_t MasterService::EraseReplicasWithCacheTotalAccounting(
...
+        return tl::make_unexpected(ErrorCode::INTERNAL_ERROR);
+    }
+    if (metadata.HasReplica([](const Replica& replica) {
+            return replica.is_memory_replica() && replica.is_processing();
+        })) {
+        return {};
+    }
+
+    auto account = GetBoundTenantQuotaHandle(tenant_state);
+    auto settle_result = metadata.quota_ledger.SettlePrimaryWrite(
```
Aionw, line 1672: **[P1] Keep task-owned promotion bytes out of primary settlement.** `PromotionAllocStart` can leave a task-owned PROCESSING MEMORY replica on this object, while `UpsertStart` does not reject an existing promotion task. A same-size `UpsertEnd(..., MEMORY)` then marks that staged replica COMPLETE, so `CompletedMemoryQuotaCharge(metadata)` returns the object size even though this ledger's primary `pending + committed` bytes are still zero. `SettlePrimaryWrite` therefore reports `kAccountingMismatch`; afterward `NotifyPromotionSuccess` sees the replica is no longer PROCESSING, refunds the promotion task's pending charge, and leaves a COMPLETE MEMORY replica with `charged_bytes == 0`. I reproduced this on `bdf52321`: `UpsertEnd` returned `INTERNAL_ERROR`, `NotifyPromotionSuccess` failed, and the 1024-byte MEMORY replica remained while the tenant charge dropped to zero. Please either reject `UpsertStart` while a promotion task exists, or distinguish task-owned replicas during primary settlement, and add this interleaving as a regression test.

Aionw (commented):
Found one reproducible quota-accounting issue in the interaction between promotion and same-size Upsert.

mooncake-store/src/master_service.cpp
Lin-z-w: Thanks for the detailed report. I reproduced the sequence and confirmed the accounting mismatch.

The underlying Promotion/Upsert lifecycle overlap predates this PR. Under the previous reserved-byte accounting, the same sequence could silently leave a COMPLETE MEMORY replica uncharged; the new ledger exposes it as an accounting mismatch.

For this PR, I will take the first approach you suggested: add an explicit `promotion_tasks` gate in `UpsertStart`. I will also add a focused regression test verifying that a same-size Upsert is rejected while promotion is in flight, that the promotion can still complete normally, and that `charged_bytes` remains correct.

The broader lifecycle issues—such as validating a successful primary Start in `PutEnd/UpsertEnd` and consistently cleaning up promotion state when stale-handle removal affects its source—will be tracked in a separate bug and addressed in a follow-up PR.

Thanks again for catching this interaction.

mooncake-store/src/master_service.cpp
yokinoshitayoki, line 2934: `RestoreFromStandbySnapshot()` creates `ObjectMetadata` after the constructor's `RebuildTenantQuotaUsageFromMetadata()` has already run against an empty metadata table. The restored object's `quota_ledger` therefore remains zero, and the tenant's `charged_bytes` does not include any restored COMPLETE MEMORY replicas. The supervisor then registers the RPC service without another rebuild.
After promotion, this allows new allocations to exceed the tenant quota. It also makes normal operations on restored objects fail: for example, a same-size Upsert reaches `SettlePrimaryWrite()` with `actual_committed_bytes > pending + committed` and returns `INTERNAL_ERROR`; Copy and eviction can similarly hit accounting mismatches.
The previous implementation also undercounted this standby-restore path, but this PR explicitly guarantees rebuilding accounting after restore, and the new per-object ledger turns the omission into additional hard failures. Please rebuild both object ledgers and global tenant usage after all standby metadata has been restored and before `RegisterRpcService()`, and add a strict multi-tenant + COMPLETE MEMORY standby-promotion regression test.

mooncake-store/src/master_service.cpp
```diff
@@ -4402,10 +4462,19 @@ auto MasterService::UpsertStart(const UUID& client_id, const std::string& key,
                     merged_config.with_soft_pin || metadata.IsSoftPinned();
 
                 const std::string existing_group_id = metadata.group_id;
-                const uint64_t old_quota_charge =
-                    metadata.committed_quota_charge_bytes != 0
-                        ? metadata.committed_quota_charge_bytes
-                        : CompletedMemoryQuotaCharge(metadata);
+                TenantQuotaLedger replacement_charge;
+                auto* quota_account = GetBoundTenantQuotaHandle(tenant_state);
+                if (enable_multi_tenants_) {
```
yokinoshitayoki: A valid disk-only or NoF-only object owns zero tenant-quota bytes. This commonly happens after the object's MEMORY replicas have been offloaded or evicted. In that state, `metadata.quota_ledger.TotalChargedBytes()` is zero.

This branch nevertheless calls `TransferReplacementCharge()` unconditionally when multi-tenancy is enabled. That method rejects `transfer_bytes == 0` as `kAccountingMismatch`, so every size-changing Upsert on such an object returns `INTERNAL_ERROR` before allocation begins.

Please skip the replacement transfer, rollback, and second transfer when the old ledger owns zero bytes, rather than treating a valid zero-charge object as corrupted. A regression test can create a disk-only object with `charged_bytes == 0`, perform a size-changing Upsert to a MEMORY replica, and verify that the operation succeeds and charges only the new replica.

mooncake-store/src/master_service.cpp
Lin-z-w: Thanks for catching this. Fixed in `fc055f65`.

`RestoreFromStandbySnapshot()` now calls `RebuildTenantQuotaUsageFromMetadata()` after all standby metadata has been restored. This rebuilds both each object's local quota ledger and the global tenant `charged_bytes` before the supervisor registers the RPC service.

mooncake-store/src/master_service.cpp
Lin-z-w: Thanks, fixed in `fc055f65`.

`UpsertStart()` now enters the replacement-charge lifecycle only when the old object ledger owns nonzero bytes. The initial transfer, allocation-failure rollback, and post-allocation transfer are all guarded by the same condition, so valid disk-only or NoF-only objects with zero charge no longer produce an accounting mismatch. The existing nonzero replacement-charge path is unchanged.

yokinoshitayoki (approved):

ykwd (approved):

Icedcoco:
I found one HA/quota edge case that looks reproducible on `fc055f65`. Could we take another look at this before merging?

In `FinalizeRemovedReplicasAfterDurable` (`mooncake-store/src/master_service.cpp:1716`), the release amount is calculated from the aggregate ledger state:

```cpp
CommittedBytes() - CompletedMemoryQuotaCharge(metadata)
```

A same-size `UpsertStart` can run after `MoveEnd` returns but before the move's durable callback. It marks the target replica as `PROCESSING`, so the callback may see `CommittedBytes() == 2S` and no completed memory charge, then release `2S` even though only the old source replica (`S`) is being finalized.

I verified this with a focused reproducer: the tenant quota snapshot changed from `2048` to `0`, and the following `UpsertEnd` returned `INTERNAL_ERROR` from the primary-write settlement.

Could we derive the release from the replicas actually finalized by this callback, rather than from the whole ledger balance? It would also be useful to add a regression test covering:

`MoveEnd -> same-key same-size UpsertStart -> durable finalize -> UpsertEnd`

The expected result should be a successful `UpsertEnd` and a final quota charge of `S`.

yokinoshitayoki:
@Lin-z-w Seems this branch has conflicts that must be resolved, should we take a look?

Lin-z-w:
> I found one HA/quota edge case that looks reproducible on `fc055f65`. Could we take another look at this before merging?
> 
> In `FinalizeRemovedReplicasAfterDurable` (`mooncake-store/src/master_service.cpp:1716`), the release amount is calculated from the aggregate ledger state:
> 
> ```c++
> CommittedBytes() - CompletedMemoryQuotaCharge(metadata)
> ```
> 
> A same-size `UpsertStart` can run after `MoveEnd` returns but before the move's durable callback. It marks the target replica as `PROCESSING`, so the callback may see `CommittedBytes() == 2S` and no completed memory charge, then release `2S` even though only the old source replica (`S`) is being finalized.
> 
> I verified this with a focused reproducer: the tenant quota snapshot changed from `2048` to `0`, and the following `UpsertEnd` returned `INTERNAL_ERROR` from the primary-write settlement.
> 
> Could we derive the release from the replicas actually finalized by this callback, rather than from the whole ledger balance? It would also be useful to add a regression test covering:
> 
> `MoveEnd -> same-key same-size UpsertStart -> durable finalize -> UpsertEnd`
> 
> The expected result should be a successful `UpsertEnd` and a final quota charge of `S`.

Thanks for the detailed reproducer. Fixed in `295c5ebd` and included in the latest PR head `6551b1b5`.

`FinalizeRemovedReplicasAfterDurable()` now derives the release amount from the MEMORY replicas actually removed by that callback:

```cpp
metadata.size * erased_memory_replicas
```

It no longer uses the aggregate `CommittedBytes() - CompletedMemoryQuotaCharge(metadata)` difference. If the calculated release exceeds the ledger’s committed bytes, the callback explicitly logs an accounting mismatch without modifying the ledger.

I also added `MoveFinalizeDuringSameSizeUpsertReleasesOnlyRemovedReplicaQuota`, covering the exact sequence:

`MoveEnd -> same-size UpsertStart -> durable finalize -> UpsertEnd`

The test verifies that the charge changes from `2S` to `S` after durable finalization, `UpsertEnd` succeeds, and the final charge remains `S`.

Icedcoco (approved):

Icedcoco:
Looks like the Check code format step is failing on CI. Please fix it when you get a chance.

Lin-z-w:
> Looks like the Check code format step is failing on CI. Please fix it when you get a chance.

fixed in [785e557](https://github.com/kvcache-ai/Mooncake/pull/3162/commits/785e55755c0788c6edbfeb398356e32a6b894751).

yokinoshitayoki:
@Lin-z-w Hi, I moved the new multi-tenant deployment and quota API updates into the dedicated Multi-Tenant Deployment guide. This aligns [PR 3378](https://github.com/kvcache-ai/Mooncake/pull/3378) with the documentation structure recently merged into main and resolves the resulting conflict. Could you double check the update?

Lin-z-w:
> @Lin-z-w Hi, I moved the new multi-tenant deployment and quota API updates into the dedicated Multi-Tenant Deployment guide. This aligns [PR 3378](https://github.com/kvcache-ai/Mooncake/pull/3378) with the documentation structure recently merged into main and resolves the resulting conflict. Could you double check the update?

Thanks! I double-checked the update. The relocated tenant quota documentation is consistent with PR #3162, and the deployment guide links to the new guide correctly. Looks good to me.

PR #3148, opened 2026-07-28
[CI/Build] Fix nightly workflow and offload platform checks

Aionw:
## Description

Fix the Nightly Build & Test workflow startup failure and move expensive cross-platform, packaging, and coverage checks out of per-PR CI.

The first scheduled nightly run failed before creating any jobs because `nightly.yml` passed the removed `build-with-ep` input to `_build-wheel.yaml`. This change removes those stale inputs.

To reduce PR CI cost and latency, the following checks now run nightly:

- MUSA compilation
- CUDA 13 ARM64 wheel builds for Python 3.10 and 3.12
- Docker image build
- Ascend hardware tests
- Coverage-instrumented Debug + ASAN build, CTest, lcov report generation, and Codecov upload

The nightly gate includes every migrated job. Core PR feedback remains in place, including Debug + ASAN tests, CUDA 13 x86 wheels required by the external integration test, EFA builds, and TENT builds. PR builds no longer install coverage tools, compile with `--coverage`, or spend time generating duplicate reports for both Python matrix entries.

No existing open PR or issue was found covering this change.

## Module

- [x] CI/CD

## Type of Change

- [x] Bug fix
- [x] Performance improvement

## How Has This Been Tested?

**Test commands:**
```bash
actionlint .github/workflows/ci.yml .github/workflows/nightly.yml
pre-commit run --files .github/workflows/ci.yml .github/workflows/nightly.yml
git diff --check
```

**Test results:**
- [x] GitHub Actions workflow syntax and reusable-workflow inputs pass actionlint
- [x] All pre-commit hooks applicable to the changed files pass
- [x] YAML and whitespace checks pass
- [ ] Runtime CI jobs pass (requires GitHub-hosted and self-hosted runners)

## Checklist

- [x] I have performed a self-review of my own code
- [x] I have formatted my code using `./scripts/code_format.sh`
- [ ] I have run `pre-commit run --all-files` and all hooks pass
- [x] I have updated the documentation (not applicable; workflow-only change)
- [ ] I have added tests to prove my changes are effective (workflow static validation only)
- [ ] For changes >500 LOC: I have filed an RFC issue (not applicable)

## AI Assistance Disclosure

- [x] AI tools were used (specified below)

Codex assisted with diagnosing the startup failure, analyzing recent PR CI job durations, refactoring the workflows, and running static validation. This PR is intentionally opened as a draft so the human submitter can review every changed line before marking it ready.

PR #3163, opened 2026-07-28
[CI/Build] Deduplicate EFA and store CI workflows

Aionw:
## Description

Deduplicate the GitHub Actions implementation for AWS EFA wheel builds and the shared Mooncake Store binding tests while preserving the existing workflow behavior.

This change:

- adds `_build-efa-wheel.yaml` as the single implementation for CI and release EFA wheel builds
- makes `ci_efa.yml`, `release-efa.yaml`, and `release-efa-non-cuda.yaml` thin callers
- reuses the existing `_publish-wheel.yaml` for both EFA release workflows
- centralizes the repeated job-level `run-ci` conditions in `ci.yml` with YAML anchors
- moves the duplicated Rust smoke/benchmark and Go binding integration blocks into scripts under `scripts/ci/`
- preserves the CI/release Python matrices, CMake flags, CUDA installation, libfabric exclusion check, artifact names and patterns, permissions, secrets, and job dependencies

### Relationship to #3148

#3148 also changes `ci.yml` and `nightly.yml`, but has a different scope: it fixes the undefined nightly `build-with-ep` inputs and moves expensive platform and coverage jobs out of per-PR CI. This PR intentionally does neither. It is limited to the accepted workflow deduplication and script extraction, so any overlap is file-level rather than a duplicate implementation goal.

## Module

- [x] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
pre-commit run --files \
  .github/workflows/_build-efa-wheel.yaml \
  .github/workflows/ci.yml \
  .github/workflows/ci_efa.yml \
  .github/workflows/nightly.yml \
  .github/workflows/release-efa.yaml \
  .github/workflows/release-efa-non-cuda.yaml \
  scripts/ci/run_store_go_integration.sh \
  scripts/ci/run_store_rust_smoke.sh
bash -n scripts/ci/run_store_rust_smoke.sh scripts/ci/run_store_go_integration.sh
git diff --check
```

A static parity audit also compared the pre- and post-refactor triggers, conditions, matrices, workflow-call inputs, CMake/CUDA flags, permissions, artifact names and patterns, publish secrets, and job dependencies. A repository-wide reusable-workflow contract check found only the two known pre-existing nightly `build-with-ep` mismatches.

**Test results:**
- [ ] Unit tests pass
- [ ] Integration tests pass (requires GitHub-hosted CUDA/EFA jobs)
- [x] Manual/static validation completed as described above
- [x] All touched-file pre-commit hooks pass
- [x] All 27 workflow files parse as YAML
- [x] `bash -n` and `git diff --check` pass

`actionlint` and `shellcheck` were not available locally. Full CUDA/EFA builds were not run.

## Checklist

- [ ] I have run `pre-commit run --all-files` and all hooks pass (touched files pass)
- [x] I have updated the documentation (not applicable; workflow-only refactor)
- [ ] I have added tests to prove my changes are effective (static parity validation only)

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex analyzed the workflow duplication, implemented the reusable workflow and scripts, and ran the static validation. This PR is intentionally opened as a draft so the human submitter can review every changed line and confirm they can defend the change end-to-end before marking it ready.

### DIFF
--- a/.github/workflows/_build-efa-wheel.yaml
+++ b/.github/workflows/_build-efa-wheel.yaml
@@ -1,0 +1,203 @@
+name: _build-efa-wheel
+
+# Shared AWS EFA wheel build used by pull-request CI and release workflows.
+# EFA hardware is not required to build: the distro libfabric is used for
+# compilation, then auditwheel leaves libfabric/libefa to the system runtime.
+
+on:
+  workflow_call:
+    inputs:
+      variant:
+        # cuda | non-cuda
+        type: string
+        required: true
+      use-cuda:
+        type: boolean
+        required: true
+      python-versions:
+        # JSON array consumed by the called workflow's Python matrix.
+        type: string
+        required: true
+      build-profile:
+        # ci uses Ninja and CI diagnostics; release preserves the release make build.
+        type: string
+        required: true
+      cmake-args:
+        type: string
+        required: true
+      variant-flag:
+        # build_wheel.sh package variant, e.g. EFA_BUILD.
+        type: string
+        required: true
+      torch-cuda-arch-list:
+        type: string
+        default: ''
+      artifact-prefix:
+        type: string
+        required: true
+
+env:
+  SCCACHE_GHA_ENABLED: "true"
+
+jobs:
+  build:
+    runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        python-version: ${{ fromJSON(inputs.python-versions) }}
+    env:
+      BUILD_PROFILE: ${{ inputs.build-profile }}
+      CMAKE_ARGS: ${{ inputs.cmake-args }}
+      EFA_VARIANT: ${{ inputs.variant }}
+      USE_CUDA: ${{ inputs.use-cuda }}
+      VARIANT_FLAG: ${{ inputs.variant-flag }}
+
+    steps:
+      - name: Validate build inputs
+        run: |
+          case "$BUILD_PROFILE" in
+            ci|release) ;;
+            *) echo "::error::Unknown EFA build profile: $BUILD_PROFILE"; exit 1 ;;
+          esac
+
+          case "$EFA_VARIANT:$USE_CUDA:$VARIANT_FLAG" in
+            cuda:true:EFA_BUILD|non-cuda:false:EFA_NON_CUDA_BUILD) ;;
+            *) echo "::error::Inconsistent EFA variant inputs"; exit 1 ;;
+          esac
+        shell: bash
+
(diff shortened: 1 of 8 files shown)


whn09 (commented):
Reviewed this from the EFA angle — LGTM. I diffed the base and head workflow files item by item and the EFA paths are preserved exactly:

- cmake args identical for all three callers (ci_efa.yml, release-efa.yaml, release-efa-non-cuda.yaml), including the CUDA stubs linker flag and TORCH_CUDA_ARCH_LIST
- make -j3 for release vs Ninja for CI correctly split via build-profile
- persist-credentials semantics preserved (release only)
- the "libfabric/libefa must NOT be bundled" check is carried over verbatim — that one matters a lot for us, since the EFA wheel must resolve to the system /opt/amazon/efa/lib at runtime
- artifact naming and the PyPI publish path unchanged

Two non-blocking notes:

1. The description says full CUDA/EFA builds weren't run. Since the EFA variants are the ones with the most moving parts (CUDA stubs + auditwheel exclusion), it'd be good to trigger both once before merge.
2. The CI variants now occupy two runners instead of one job, so dependencies.sh + the CUDA toolkit install run twice. Fine by me — the parallelism is probably worth it — just flagging the cost.

One request: could you add an EFA cu13 variant while you're in here? Right now _build-efa-wheel.yaml hardcodes cuda: '12.8.1', and there is no release-efa-cuda13.yaml counterpart to release-cuda13.yaml, so EFA users on CUDA 13 have to pull in a cu12 runtime. Real EFA hosts are already moving there — the p5/p5en instances I validate on run torch 2.12.1+cu130 with driver 595.71.05.

Your refactor actually makes this cheap: add a cuda-version input (plus maybe a cuda-toolkit-action one, since ci_cu13.yml pins Jimver/cuda-toolkit@v0.2.29 while everything else is on v0.2.24), then a thin release-efa-cuda13.yaml caller with variant-flag: EFA_BUILD and a cu130 artifact prefix. Before this PR that would have meant duplicating a 162-line workflow. Happy to test the resulting wheel on real EFA hardware if you'd rather I take it as a follow-up PR.

Aionw:
EFA cu13 will be landed in a separate PR after this PR merged.

ykwd (approved):

PR #3165, opened 2026-07-28
[Store] Validate durable oplog prefix at startup

Icedcoco:
## Description

Validate the durable oplog prefix when `OpLogBatchStorage` starts, so inconsistent or corrupted metadata is rejected before recovery proceeds.

This PR:

- requires `batch_id` and `last_seq` to be either both zero or both nonzero;
- requires an empty batch namespace when the durable prefix is `{0, 0}`;
- verifies that a nonzero prefix references an existing, decodable terminal batch;
- verifies that the terminal batch ID and `last_seq` match the durable prefix;
- applies the same validation after losing the prefix initialization CAS race;
- propagates backend `Get` and `Range` errors without converting them into corruption errors;
- rejects batch records whose payload ID does not match the requested storage key.

The validation is intentionally limited to the durable-prefix boundary. It does not scan historical batches or repair inconsistent metadata.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

Added unit tests covering valid prefixes, inconsistent zero fields, missing or corrupted terminal batches, ID and sequence mismatches, CAS races, and backend error propagation.

**Test commands:**

```bash
  ./build/mooncake-store/tests/oplog_batch_storage_test

  ctest --test-dir build \
  -R '^oplog_batch_storage_test$' \
  --output-on-failure

pre-commit run --files \
  mooncake-store/include/ha/oplog/oplog_batch_storage.h \
  mooncake-store/src/ha/oplog/oplog_batch_storage.cpp \
  mooncake-store/tests/ha/oplog/oplog_batch_storage_test.cpp
```

**Test results:**

- [x] Unit tests pass (34/34)

## Checklist

- [ ] I have updated the documentation (not applicable; no user-facing behavior changed)

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex assisted with the design, implementation, test coverage, and verification. The human submitter reviewed the final changes and is responsible for the submitted code.

### DIFF
--- a/mooncake-store/include/ha/oplog/oplog_batch_storage.h
+++ b/mooncake-store/include/ha/oplog/oplog_batch_storage.h
@@ -24,6 +24,7 @@ class OpLogBatchStorage {
    private:
     bool IsValidClusterId() const;
     ErrorCode RejectLegacyLayout() const;
+    ErrorCode ValidateDurablePrefixAtStartup(const DurablePrefix& prefix);
 
     std::string cluster_id_;
     HaKvBackend& backend_;
--- a/mooncake-store/src/ha/oplog/oplog_batch_storage.cpp
+++ b/mooncake-store/src/ha/oplog/oplog_batch_storage.cpp
@@ -209,6 +209,11 @@ ErrorCode OpLogBatchStorage::ReadBatch(uint64_t batch_id,
         LOG(ERROR) << "Failed to decode OpLog batch record: " << reason;
         return ErrorCode::INTERNAL_ERROR;
     }
+    if (batch.batch_id != batch_id) {
+        LOG(ERROR) << "OpLog batch id does not match key: requested="
+                   << batch_id << ", payload=" << batch.batch_id;
+        return ErrorCode::INTERNAL_ERROR;
+    }
     return ErrorCode::OK;
 }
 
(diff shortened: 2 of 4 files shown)


he-yufeng:
Walked the invariant chain before commenting:

- The zero/nonzero XOR matches the writer: `WriteBatchAndAdvancePrefix` only advances both fields together under the expected-prefix CAS, and the retry read validates `batch_id` + `last_seq` against the stored batch, so a torn prefix really is corruption.
- The zero-prefix namespace check is the right corruption signal, and limit 1 is all it needs.
- The terminal-batch lookup is safe today because nothing in `ha/` deletes batch records yet, so the prefix cannot outlive its batch.

One forward-looking note: bounded retention (#3167) will introduce batch pruning. Once a pruned terminal batch is a legitimate state, this validator turns that state into a startup failure. Might be worth leaving a marker here (or in the retention design) that the missing-terminal case becomes warn-and-continue rather than INTERNAL_ERROR once pruning exists, so the two changes do not collide later.

Icedcoco:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

Icedcoco:
> One forward-looking note: bounded retention (#3167) will introduce batch pruning. Once a pruned terminal batch is a legitimate state, this validator turns that state into a startup failure. Might be worth leaving a marker here (or in the retention design) that the missing-terminal case becomes warn-and-continue rather than INTERNAL_ERROR once pruning exists, so the two changes do not collide later.

Hi @he-yufeng , thanks for the review! #3167 won't prune terminal batches because retention will keep two snapshots along with all batch oplogs since the older snapshot. So this validator remains safe.

00fish0 (approved):
Nice work!

PR #3168, opened 2026-07-28
[Store] Scope capacity metric release to the serving MasterService

liuzijing2014:
Follow-up to #3154, addressing the snapshot-reader lifecycle issue raised in https://github.com/kvcache-ai/Mooncake/pull/3154#issuecomment-5101860445.

## Problem

#3154 released mounted segments' `total_mem_capacity` contribution in `~SegmentManager`. However, `CatalogBackedSnapshotProvider` creates a temporary `SegmentManager` and populates `mounted_segments_` via `SegmentSerializer::Deserialize`, which inserts records directly without going through `MountSegment` — those records never incremented the gauge. Destroying the temporary reader therefore decrements capacity that was never added, leaving the gauge below the capacity actually accounted by the serving service (or negative on a standby with snapshot restore enabled).

## Fix

Move the release from `~SegmentManager` to the end of `~MasterService`, next to the existing `standby_accounted_memory_bytes_` cleanup, exposed as `SegmentManager::releaseCapacityMetrics()`. Only the serving instance releases its contribution; standalone and deserialized managers leave the gauges untouched.

The CXL `initializeCxlAllocator()` contribution discussed in the same comment is intentionally left for a separate follow-up.

## Testing

Added `MasterMetricsTest.SnapshotReaderTeardownKeepsCapacityIntact`: serializes a manager with an accounted mount, deserializes into a temporary `SegmentManager`, destroys it, and asserts the gauges are unchanged. The existing `ServiceTeardownReleasesSegmentCapacity` still covers the leadership-change leak. `master_metrics_test`, `segment_test`, `master_service_test`, and `serializer_test` pass locally.

### DIFF
--- a/mooncake-store/include/segment.h
+++ b/mooncake-store/include/segment.h
@@ -452,12 +452,16 @@ class SegmentManager {
         : memory_allocator_(memory_allocator), enable_cxl_(enable_cxl) {}
 
     /**
-     * @brief Destructor. Releases the capacity metric contribution of
-     *        segments that are still mounted, since MasterMetricManager
-     *        outlives MasterService instances (e.g. across HA leadership
-     *        changes).
+     * @brief Releases the capacity metric contribution of segments that
+     *        are still mounted. Intended to be called when the owning
+     *        MasterService is torn down: MasterMetricManager outlives
+     *        MasterService instances (e.g. across HA leadership changes).
+     *        This is deliberately not done in the destructor, because other
+     *        SegmentManager instances (such as the temporary snapshot
+     *        readers) hold deserialized records that never contributed to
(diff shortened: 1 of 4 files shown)


Icedcoco (approved):
LGTM. This addresses the key issue from #3154 by scoping capacity metric cleanup to the serving `MasterService`, while ensuring that temporary snapshot-deserialized `SegmentManager` instances do not release capacity they never accounted for.

The added snapshot reader regression test covers the lifecycle concern we discussed. CXL capacity accounting is still outside the scope of this PR and should be handled separately, particularly for HA teardown and snapshot restore.

ykwd (approved):

PR #3169, opened 2026-07-28
[Store] Fix segment capacity metric ownership

Dao007forever:
## Description

Follow-up to #3154. The process-scoped master metrics outlive individual `MasterService` terms, but snapshot-deserialized `SegmentManager` instances do not own the capacity contributions represented by their restored segment metadata. Unconditionally reconciling every stored segment during destruction can therefore decrement capacity owned by another serving instance.

This change:

- records runtime ownership for each mounted segment capacity contribution;
- releases only owned contributions from `SegmentManager` destruction, after `MasterService` has joined its workers;
- explicitly assigns ownership when restored segments are activated by a serving master;
- tracks the shared global CXL allocator contribution separately from per-client CXL mounts; and
- preserves regular capacity accounting for `protocol == "cxl"` mounts when CXL mode is disabled.

This keeps memory-capacity gauges correct across HA leadership terms, constructor unwinding, normal unmounts, and snapshot-only `SegmentManager` lifecycles.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

Added regression coverage for service teardown, standalone `SegmentManager` destruction, snapshot-deserialized non-ownership, and CXL-protocol accounting when CXL mode is disabled.

**Test commands:**
```bash
clang-format --dry-run --Werror --lines=<changed-lines> <changed-file>
git diff --check
```

**Test results:**
- [ ] Unit tests pass — not run because CMake and an existing test binary are unavailable in the local environment
- [x] Manual testing done: changed C++ ranges pass clang-format validation and the committed diff passes whitespace validation

## Checklist

- [x] For changes >500 LOC: not applicable (under 500 LOC)

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex assisted with analyzing the metric ownership lifecycle, implementing the changes, adding tests, and reviewing the resulting diff. The human submitter remains responsible for reviewing and understanding every changed line before marking this PR ready for review.

### DIFF
--- a/mooncake-store/include/segment.h
+++ b/mooncake-store/include/segment.h
@@ -3,6 +3,7 @@
 #include <boost/functional/hash.hpp>
 #include <chrono>
 #include <map>
+#include <optional>
 #include <ostream>
 #include <set>
 #include <shared_mutex>
@@ -55,6 +56,10 @@ struct MountedSegment {
     Segment segment;
     SegmentStatus status;
     std::shared_ptr<BufferAllocatorBase> buf_allocator;
+    // Runtime-only ownership marker. Snapshot deserialization reconstructs
+    // segment state without registering capacity metrics, so this field is
+    // deliberately not serialized.
+    bool capacity_metric_accounted{false};
 };
 
 struct MountedNoFSegment {
@@ -152,6 +157,11 @@ class ScopedSegmentAccess {
                                      const UUID& client_id) const;
 
     bool GetSegment(const UUID& segment_id, Segment& segment) const;
+
+    /**
+     * @brief Register the capacity metric for an already-restored segment.
+     */
+    ErrorCode AccountSegmentCapacityMetric(const UUID& segment_id);
 
     struct AllocatorReplacement {
         UUID segment_id;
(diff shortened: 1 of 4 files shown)


Icedcoco:
Thanks for taking the time to investigate the broader ownership model and for covering several lifecycle cases. The per-segment ownership marker does address the snapshot reader problem identified in #3154.

However, I think there is still an issue in the enabled-CXL snapshot restore path. `initializeCxlAllocator()` initially adds the global `cxl_size` contribution and records it in `cxl_capacity_metric_contribution_`. Later, `ApplySnapshotState()` calls `reset_total_mem_capacity()`, but `AccountSegmentCapacityMetricLocked()` deliberately skips CXL segments when `enable_cxl_` is true. As a result, the global CXL contribution is not restored after the reset. When the `SegmentManager` is eventually destroyed, `ReleaseMetricContributions()` still subtracts the recorded `cxl_size`, which can make the global capacity negative. The current CXL test only covers `protocol == "cxl"` while CXL mode is disabled, so it does not exercise this path.

In parallel, #3168 was completed first and contains the narrower fix needed for the original HA capacity leak, including the snapshot reader regression coverage. Since the two PRs are alternative changes based on the same #3154 commit and modify the same ownership boundary, we will proceed with #3168 rather than merge both.

I’m sorry that we cannot merge this PR in its current form after the work you put into it. The enabled-CXL accounting path still deserves a dedicated follow-up, ideally covering initialization, snapshot reset and restore, and service teardown in one regression test. Contributions there, or in other areas of Mooncake, would be very welcome.

Dao007forever:
@Icedcoco I think in the normal CXL flow, snapshot restoration and metric reset happen before the global CXL allocator is initialized:

```
RestoreState() calls ApplySnapshotState() -> initializeCxlAllocator() -> ~SegmentManager()
```

The allocator then contributes cxl_size once, so teardown can safely subtract it once.

I might be wrong since I'm not familiar with CXL.

PR #3170, opened 2026-07-28
[CI/Build] Align Asio with yalantinglibs target

CanYangGetYang:
## What changed

- make `mooncake_common` publicly depend on `yalantinglibs::yalantinglibs`
- make `asio_shared` consume Asio through the same yalantinglibs target
- remove the independent system Asio lookup

## Why

`mooncake-common/include/default_config.h` publicly includes
`<ylt/easylog.hpp>`, but `mooncake_common` did not propagate the corresponding
CMake target. Builds could therefore succeed only because yalantinglibs happened
to be installed in a compiler-default include path.

There was also an ABI consistency issue: Transfer Engine targets already consume
Asio headers transitively through yalantinglibs, while `asio_shared` independently
selected a system Asio installation. If those header versions differ,
`libtransfer_engine.so` can request Asio symbols that the generated `libasio.so`
does not export. Using one CMake target for both sides keeps their compile-time
Asio ABI aligned.

## Validation

- configured Mooncake from the current `main` branch
- built `asio_shared`, `mooncake_common`, `transfer_engine`, and `engine`
- verified that the Asio symbol required by `libtransfer_engine.so` is exported
  by the generated `libasio.so`
- imported `TransferEngine` from the generated Python extension
- ran whitespace and diff checks on the changed file

### DIFF
--- a/mooncake-common/src/CMakeLists.txt
+++ b/mooncake-common/src/CMakeLists.txt
@@ -1,27 +1,5 @@
 find_package(yaml-cpp REQUIRED)
 
-find_package(asio QUIET)
-
-if(asio_FOUND)
-    message(STATUS "Found ASIO via find_package")
-    set(ASIO_INCLUDE_DIR ${asio_INCLUDE_DIR})
-else()
-    find_path(ASIO_INCLUDE_DIR
-        NAMES asio.hpp
-        PATHS
-            /usr/local/include
-            /usr/include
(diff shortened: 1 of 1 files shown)


CanYangGetYang:
Anyone can review it?

stmatengss (approved):
LGTM. Please double-check. @ykwd

Aionw (approved):

PR #3174, opened 2026-07-29
[TransferEngine] Allow idempotent duplicate memory registration

### LINKED ISSUE #3153: [Bug]: V0. 3.12.post1 Transfer Engine does not support overlapped memory region
### Bug Report

transfer_engine_impl.cpp:787] Transfer Engine does not support overlapped memory region

When I upgraded to version V0.3.12.post1, this error appeared

### Before submitting...

alogfans:
## Description

Fixes #3153.

Mooncake 0.3.12 added engine-level memory-region reservation to reject overlapping registrations, including in-flight duplicates. That also made already-completed byte-identical registrations fail with `ERR_ADDRESS_OVERLAPPED`, which breaks callers that register the same buffer through repeated init/reload/reconnect paths.

This PR allows completed duplicate registrations of the exact same region to be idempotent:

- `registerLocalMemory(addr, length, location, remote_accessible)` returns success when the same completed region is registered again.
- `registerLocalMemoryBatch(...)` returns success when the same completed batch regions are registered again.
- Duplicate completed registrations bump an engine-level ref count.
- Unregister only calls the transport when the final reference is released.
- True overlaps and in-flight duplicate registrations remain rejected.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
cmake --build build --target transport_uint_test -j4
build/mooncake-transfer-engine/tests/transport_uint_test --gtest_color=no
```

**Test results:**
- [x] Unit tests pass

`transport_uint_test` passes: 20/20 tests.

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex assisted with implementation, test updates, and PR description preparation. The human submitter is responsible for reviewing and defending all changed lines.

### DIFF
--- a/mooncake-store/tests/pybind_client_test.cpp
+++ b/mooncake-store/tests/pybind_client_test.cpp
@@ -1250,14 +1250,13 @@ TEST_F(RealClientTest, ErrBufferRegistrationErrors) {
         EXPECT_EQ(py_client_->register_buffer(buf, sizeof(buf)), 0)
             << "First registration should succeed";
 
-        // Second registration of the same memory
-        {
-            GLogMuter muter;
-            EXPECT_NE(py_client_->register_buffer(buf, sizeof(buf)), 0)
-                << "Duplicate registration of the same buffer should fail.";
-        }
+        EXPECT_EQ(py_client_->register_buffer(buf, sizeof(buf)), 0)
+            << "Duplicate registration of the same buffer should succeed "
+               "and increment the registration refcount.";
 
         // Cleanup
+        EXPECT_EQ(py_client_->unregister_buffer(buf), 0)
+            << "First unregistration should decrement the refcount";
         EXPECT_EQ(py_client_->unregister_buffer(buf), 0)
             << "Unregistration should succeed";
     }
--- a/mooncake-transfer-engine/include/transfer_engine_impl.h
+++ b/mooncake-transfer-engine/include/transfer_engine_impl.h
@@ -401,6 +401,7 @@ class TransferEngineImpl {
         uint64_t length;
         std::string location;
         bool remote_accessible;
+        uint64_t ref_count = 1;
     };
 
     using MemoryRegionMap = std::map<uintptr_t, MemoryRegion>;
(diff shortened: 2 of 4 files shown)


he-yufeng:
This is the contract I was hoping for when I triaged #3153, and the shape holds up on a close read:

- `tryBumpLocalMemoryRegionRefs` only short-circuits on a full exact match (addr + length + location + remote_accessible), and falls back to the reserve path otherwise, so true overlaps and mismatched re-registrations keep the rejection behavior from #2869/#2870. remote_accessible being part of the match is the right conservative call.
- The refcount is symmetric: unregister decrements and only reaches the transports at zero, so register x2 / unregister x1 leaves the region live, and unregister of an unknown addr still falls through to the old error path. Both sides take the same unique_lock, so bump and release serialize.
- Batch is atomic as before: a mixed [exact-dup, fresh] batch fails the bump and hits the reserve path, which rejects the whole batch on the dup. That keeps registerBatch all-or-nothing, but it does mean a partially duplicate batch still fails entirely instead of succeeding on the fresh entries. Worth a line in the docs or the PR description so callers know the idempotency is per-batch-atomic, not per-entry.

Nice tests covering the refcount roundtrip. Leaving the formal approve for a second pair of eyes since this is the registration hot path.

he-yufeng:
Read through the refcount mechanics against the registration/unregistration paths. The strictness on the bump check (length + location + remote_accessible all must match, in-flight overlap falls through to the reserve path) looks right, and the ref counting itself is symmetric under the same lock.

One gap worth considering: on the idempotent bump path the call returns before any transport runs, so `update_metadata=true` on a duplicate registration silently does nothing. The pre-reservation behavior re-published metadata on every registration, and the reconnect/reload flows this PR is targeting are exactly the ones that may need the metadata re-published (e.g. after a metadata server restart, a client that re-registers the same buffer to re-announce it now gets a silent no-op). Either the bump path should still run the metadata update when asked, or the PR description should say duplicates are a pure no-op so callers do not rely on re-registration as a refresh.

Smaller boundary note, not a blocker: a batch mixing an already-registered region with a new one still fails wholesale, because the bump check requires every region to match before any is bumped. That matches the pre-existing all-or-nothing behavior, but it means init paths that register incrementally-larger batches still need to dedupe client-side.

Issue #3167, opened 2026-07-28
[RFC]: Standby-Generated Snapshots and Bounded OpLog Retention


Icedcoco:
> Implementation note: discussion after RFC submission changed several delivery details without changing the RFC's primary goal. The implementation keeps the legacy and batch OpLog snapshot systems permanently separate, uses `enable_oplog` as the new-mode switch, writes a 300-million-object state in fixed object-count chunks, and does not remove the legacy reader/writer. The [delivery roadmap](../../../plans/2026-07-22-oplog-ha-pr-delivery-roadmap-zh.md) and PR-N01 through PR-N13 specs are authoritative for implementation order and current progress; the RFC body below remains the submitted design context.

## Summary

This RFC moves HA snapshot generation away from the serving primary and onto standby masters. A standby captures a complete applied batch boundary, writes an immutable snapshot to shared object storage, and publishes it through a fenced etcd transaction. Recovery first tries two independently addressable snapshot pointers and replays the OpLog suffix. If both snapshots are unusable, recovery may fall back to the complete OpLog only when every batch from the initial durable boundary is still present and valid.

Once two snapshots are independently verified, Mooncake may garbage-collect older snapshot objects, publish a monotonic compaction floor, and delete OpLog batches covered by the older protected snapshot. The protocol always publishes the floor before deleting OpLog data. Snapshot GC failures are reported and retried but do not block safe OpLog pruning.

The feature is disabled by default. The first implementation supports etcd only. It does not add a primary fallback writer, an automatic write backpressure controller, or support for Redis-backed snapshot coordination.

## Motivation

The existing primary snapshot path forks the serving process while holding a global snapshot barrier. That design has three long-term problems:

1. Snapshot work competes with serving-primary resources and complicates the primary mutation lock graph.
2. Snapshot descriptors do not carry a complete batch cursor, so a reader cannot unambiguously start suffix replay or prove which batches are safe to remove.
3. OpLog retention is unbounded. Deleting keys without a reader-visible floor would turn legitimate pruning into an indistinguishable recovery gap.

Standbys already maintain replicated object metadata and segment state. They are the natural place to create recovery checkpoints without involving the serving primary.

## Goals

- Generate snapshots only from standby state.
- Capture a state that ends at a complete durable batch boundary.
- Keep the serving primary independent of capture, encoding, and upload work.
- Publish immutable snapshots through a fenced, monotonic etcd protocol.
- Give recovery two independently addressable snapshot candidates.
[...]

PR #3178, opened 2026-07-29
[Store] Add batch OpLog snapshot metadata protocol

Icedcoco:
## Description

Part of #3167.

This PR rewrites **N01** as the standalone control-plane protocol for batch OpLog snapshots. It deliberately does not extend the legacy `SnapshotDescriptor` or snapshot catalog.

The new protocol defines:

- `BatchOpLogSnapshotDescriptor`, including the complete batch/sequence cursor, producer view, manifest size, and manifest CRC32C;
- `BatchOpLogSnapshotManifest`, including the segments artifact and ordered object-chunk descriptors;
- the fixed `standby-oplog-materialized/v1` format identifier;
- builders for the maintenance/latest/fallback/compaction-floor etcd keys and all immutable object-store artifact keys.

Descriptor and manifest values use compact JsonCpp encoding. Decoding rejects duplicate keys, missing required fields, incorrect JSON types, integer overflow, unsupported schema/format values, inconsistent cursors and snapshot IDs, empty artifact keys, and non-contiguous chunk indices. Unknown fields in the current schema are ignored so optional metadata can be added later.

### Relationship to RFC #3167

RFC #3167 moves snapshot generation from the serving primary to standby masters and uses a complete batch cursor for suffix replay and safe OpLog etention.

This PR provides the isolated metadata contract shared by the later writer, publisher, and restore PRs. It supersedes the original version of PR #3178, which modified the legacy descriptor/catalog. The legacy reader and writer remain unchanged and continue to use their existing descriptor bytes; the new batch OpLog snapshot path has no runtime caller in N01.

This PR intentionally does **not** implement capture, binary artifact encoding, object-store upload, maintenance leases, pointer publication, restore, runtime configuration, snapshot GC, or OpLog pruning. Those behaviors remain in the subsequent N-series PRs tracked by #3167.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

The focused tests cover compact JSON round trips, current-schema extension fields, malformed/duplicate/missing fields, type and integer bounds, unknown versions, cursor/ID consistency, empty-cluster cursors, chunk ordering and counts, cluster key validation, and snapshot-ID path injection.

Legacy catalog and existing batch OpLog codec tests are included as regression coverage.

**Test commands:**

```bash
  batch_oplog_snapshot_types_test \
  embedded_snapshot_catalog_store_test \
  oplog_batch_codec_test \
  -j32

LSAN_OPTIONS=detect_leaks=0 ctest --test-dir build --output-on-failure \
  -R '^(batch_oplog_snapshot_types_test|embedded_snapshot_catalog_store_test|oplog_batch_codec_test)$'

pre-commit run --files \
  mooncake-store/include/ha/snapshot/batch_oplog_snapshot_types.h \
  mooncake-store/src/ha/snapshot/batch_oplog_snapshot_types.cpp \
  mooncake-store/tests/ha/snapshot/batch_oplog_snapshot_types_test.cpp \
  mooncake-store/src/CMakeLists.txt \
  mooncake-store/tests/CMakeLists.txt
```

**Test results:**

- [x] 3/3 focused and regression test targets passed
- [x] Pre-commit passed on the files changed by this PR
- [x] Integration tests are not applicable because N01 has no runtime caller

## Checklist

- [x] I have formatted the changed files using the repository hooks
- [x] I have run pre-commit on all files changed by this PR
- [x] Documentation is not required because this PR has no user-facing behavior
- [x] I have added tests covering the new protocol and legacy isolation
- [x] RFC #3167 tracks the complete architectural change

## AI Assistance Disclosure

- [ ] No AI tools were used
- [x] AI tools were used

AI tools assisted with implementation, test design, and organizing the PR description. The human submitter will review all changed lines before marking the PR ready and is responsible for the implementation.


Icedcoco:
Schema compatibility is backward-readable but not forward-readable. All snapshot readers must be upgraded before schema-v1 publication is enabled.

Aionw:
> Schema compatibility is backward-readable but not forward-readable. All snapshot readers must be upgraded before schema-v1 publication is enabled.

Would it be worth designing a new snapshot schema using a backward-compatible serialization protocol such as Protobuf, and replacing the hand-written codec to improve robustness and maintainability?

mooncake-store/include/ha/snapshot/catalog/snapshot_catalog_store.h
```diff
@@ -110,31 +110,67 @@ inline bool ParseDecimal(std::string_view text, Integer& value) {
     return result.ec == std::errc() && result.ptr == end;
 }
 
+inline bool ValidateSnapshotDescriptor(const SnapshotDescriptor& descriptor) {
```
Aionw: Since v0 and v1 already have different layouts and invariants, could we make ValidateSnapshotDescriptor, SerializeSnapshotDescriptor, and DeserializeSnapshotDescriptor thin dispatchers over version-specific codecs?
We are likely to introduce more descriptor versions over time. If parsing, field-count checks, validation, and serialization for every version remain interleaved in the shared functions, the compatibility rules will become increasingly difficult to reason about and test.
I suggest keeping SnapshotDescriptor as the normalized in-memory model, while grouping the wire-format logic under version-specific codecs such as V0Codec and V1Codec. The public functions should only determine the version, reject unsupported versions, and dispatch.
SerializeSnapshotDescriptor() should also validate internally and preferably return tl::expected, so callers cannot accidentally serialize an unsupported descriptor or silently discard version-specific fields.

Icedcoco:
This PR has been rewritten following the design discussion. It no longer changes the legacy SnapshotDescriptor or catalog codec. N01 now defines a separate JsonCpp descriptor/manifest protocol for the batch OpLog snapshot path; current-schema readers ignore additional optional fields, and legacy readers never receive the new bytes. The new path has no runtime caller in this PR, so rollout ordering between legacy and batch OpLog snapshot readers is not involved at this stage.

mooncake-store/include/ha/snapshot/batch_oplog/metadata.h
```diff
@@ -0,0 +1,86 @@
...
+    std::string snapshot_format{kBatchOpLogSnapshotFormat};
+    std::string snapshot_id;
+    uint64_t last_included_seq{0};
+    uint64_t last_included_batch_id{0};
+    ViewVersionId producer_view_version{0};
+    BatchOpLogSnapshotObjectDescriptor segments;
+    std::vector<BatchOpLogSnapshotChunkDescriptor> object_chunks;
+};
+
+std::string EncodeBatchOpLogSnapshotDescriptor(
```
Aionw, line 56: Nit: would it be better to make this function a static member function of `BatchOpLogSnapshotDescriptor`, so its name can be shortter as `Encode`. Same comment for other function like this (include Decode).

mooncake-store/include/ha/snapshot/batch_oplog_snapshot_types.h
```diff
@@ -0,0 +1,86 @@
...
+    uint64_t last_included_seq{0};
+    uint64_t last_included_batch_id{0};
+    ViewVersionId producer_view_version{0};
+    BatchOpLogSnapshotObjectDescriptor segments;
+    std::vector<BatchOpLogSnapshotChunkDescriptor> object_chunks;
+};
+
+std::string EncodeBatchOpLogSnapshotDescriptor(
+    const BatchOpLogSnapshotDescriptor& descriptor);
+bool DecodeBatchOpLogSnapshotDescriptor(
```
Aionw: We already have type `tl::expected<>`, so return value here can be `tl::expected<xxxDescriptor, std::string>` string here to represent error_code.

mooncake-store/src/ha/snapshot/batch_oplog_snapshot_types.cpp
```diff
@@ -0,0 +1,427 @@
...
+    }
+    if (metadata.producer_view_version < 0) {
+        SetReason(reason, "producer_view_version must be non-negative");
+        return false;
+    }
+    return true;
+}
+
+template <typename SnapshotMetadata>
+void EncodeIdentity(const SnapshotMetadata& metadata, Json::Value* root) {
```
Aionw: This function assume root is not nullptr, so use Json::Value& to enforce that is better.

mooncake-store/src/ha/snapshot/batch_oplog_snapshot_types.cpp
```diff
@@ -0,0 +1,427 @@
...
+    root["chunk_index"] = static_cast<Json::UInt64>(descriptor.chunk_index);
+    root["key"] = descriptor.key;
+    root["object_count"] = static_cast<Json::UInt64>(descriptor.object_count);
+    root["stored_size"] = static_cast<Json::UInt64>(descriptor.stored_size);
+    root["crc32c"] = static_cast<Json::UInt>(descriptor.crc32c);
+    return root;
+}
+
+bool DecodeChunkDescriptor(const Json::Value& root, uint64_t expected_index,
+                           BatchOpLogSnapshotChunkDescriptor* descriptor,
```
Aionw: Same as above, use ref here is better

mooncake-store/include/ha/snapshot/batch_oplog/metadata.h
Icedcoco: I kept these as free functions to match the existing codec-style APIs in this module and keep the metadata structs as passive data objects. The longer names also make descriptor vs. manifest call sites explicit.

mooncake-store/include/ha/snapshot/batch_oplog_snapshot_types.h
Icedcoco: Updated in ea86a006. Both public Decode functions now return tl::expected<T, std::string>, so the decoded value and error message are carried by one return value.

mooncake-store/src/ha/snapshot/batch_oplog_snapshot_types.cpp
Icedcoco: Updated in ea86a006. EncodeIdentity now takes Json::Value&; I also changed the other required internal output parameters to references for consistency.

mooncake-store/src/ha/snapshot/batch_oplog_snapshot_types.cpp
Icedcoco: Updated in ea86a006. DecodeChunkDescriptor now takes the required descriptor output by reference, as do the related internal decode helpers.

Aionw (approved):
LGTM

00fish0 (approved):
nit: 

Since this introduces a new batch OpLog snapshot protocol that is intentionally separate from the legacy master snapshot/catalog path, should we also reflect that separation in the directory structure?

Right now the new file sits directly under `ha/snapshot/` as `batch_oplog_snapshot_types.*`. That is clear enough for this PR, but the PR description suggests later N-series changes will add writer/publisher/restore/GC/pruning pieces. Once those land, it may become harder to distinguish the legacy master-state snapshot path from the new OpLog-batch checkpoint path if both live flatly under `ha/snapshot/`.

Would it make sense to start with something like:

`ha/snapshot/batch_oplog/types.h`
`ha/snapshot/batch_oplog/types.cpp`

Then future files could naturally live under the same namespace/directory, while the existing `catalog/`, `master_snapshot_codec`, and catalog-backed provider code remain visibly scoped to the legacy snapshot path.

Icedcoco:
@00fish0 Addressed in `89ab2b96e`. The batch OpLog snapshot metadata and its test now live under `ha/snapshot/batch_oplog/`, with CMake and include paths updated accordingly. I used `metadata.h/.cpp` rather than `types.h/.cpp` because the header depends on the top-level `types.h`; reusing that basename would make the quoted include resolve to itself. The focused test still passes.

00fish0:
> @00fish0 Addressed in `89ab2b96e`. The batch OpLog snapshot metadata and its test now live under `ha/snapshot/batch_oplog/`, with CMake and include paths updated accordingly. I used `metadata.h/.cpp` rather than `types.h/.cpp` because the header depends on the top-level `types.h`; reusing that basename would make the quoted include resolve to itself. The focused test still passes.

@Icedcoco  LGTM, thanks!

ykwd (approved):

PR #1538, opened 2026-02-11
[TE][Store] refactor: centralize environment variable handling across components

stmatengss:
## Description

Centralizes all `getenv()` calls across mooncake-transfer-engine and mooncake-store into the `Environ` singleton in `mooncake-common/environ.h`. Eliminates scattered env var parsing, consolidates defaults, and fixes several semantic bugs discovered during migration.

Key changes:

- **Environ API extensions**: Added `GetRedisUsername()`, `GetNvlinkFabricMemEnabled()` (tri-state: returns true only when `MC_USE_NVLINK_IPC` is explicitly `"0"`), and ~30 other getters covering `MC_MAX_SGE`, `MC_MAX_WR`, `MC_MAX_INLINE`, `MC_MTU`, `MC_WORKERS_PER_CTX`, `MC_SLICE_SIZE`, `MC_RETRY_CNT`, etc.
- **GID index fix**: Default changed from `3` → `-1` (auto-selection), `NCCL_IB_GID_INDEX` fallback now handled inside Environ instead of config.cpp
- **HIP transport fix**: `supportFabricMem()` had inverted logic — was disabling fabric-mem when IPC was disabled. Fixed: `if (hip_ipc) return false;`
- **NVLink transport fix**: Switched from `GetUseNvlinkIpc()` (bool, can't distinguish unset from `"0"`) to `GetNvlinkFabricMemEnabled()` to preserve opt-in semantics
- **CXL test hardening**: Initialize raw pointers to `nullptr`, add `cxl_available_` skip guard to `EvictOperation` test
- **Build wiring**: `mooncake_common` linked into standalone TE builds, Python extension builds, P2P Store builds, Rust builds, and CI Go test CGO_LDFLAGS (including `-lmlx5`, `-lxxhash`)
- **Rust link fix**: Added `libmooncake_common.a` to `mooncake-store/rust/build.rs` link list to resolve `undefined symbol: mooncake::Environ::Get()` at Rust build time
- **config.cpp log dir check**: Replaced `opendir()` with `std::filesystem::is_directory()` (from main)

## Module

- [x] Transfer Engine (`mooncake-transfer-engine`)
- [x] Mooncake Store (`mooncake-store`)
- [x] Mooncake EP (`mooncake-ep`)
- [x] Integration (`mooncake-integration`)
- [x] P2P Store (`mooncake-p2p-store`)
- [ ] Python Wheel (`mooncake-wheel`)
- [x] PyTorch Backend (`mooncake-pg`)
[...]

PR #3179, opened 2026-07-29
[Store] Consolidate string parsing primitives

Aionw:
## Description

Consolidate duplicated string parsing and manipulation in Mooncake Store without introducing a generic `*_utils` dumping ground.

- Add focused common headers for ASCII string handling, canonical boolean parsing, and strict integer parsing.
- Make `Environ` and Store call sites use those primitives directly, preserving caller-provided defaults for invalid environment values.
- Replace local trim/case/split implementations with the focused common APIs or Boost.StringAlgo where appropriate.
- Remove the obsolete `TypeUtil` implementation and other compatibility wrappers instead of forwarding legacy interfaces.
- Declare Mooncake Store's direct `mooncake_common` and JsonCpp dependencies explicitly.

This intentionally removes internal helper APIs such as `TypeUtil`, `string_to_bool`, `GetEnvOr`, and `splitString`; configuration behavior remains covered by focused tests.

Related work: #1538 centralizes environment-variable ownership across Transfer Engine and Store. This PR is narrower and complementary: it owns lexical parsing primitives and removes Store-local duplicates, without migrating Transfer Engine configuration or adding broad `Environ` getters.

### Configuration compatibility

The migrated parsers intentionally reject ambiguous values instead of silently applying a partial conversion:

- At former `GetEnvOr<bool>` call sites, non-canonical numeric strings such as `2` previously became `true`; they are now invalid and use the caller-provided default with a warning.
- At former `GetEnvOr` integer call sites, strings with trailing non-whitespace characters such as `1048576MB` previously parsed as their numeric prefix; they are now invalid and use the caller-provided default with a warning.

Environment integer getters continue to accept surrounding ASCII whitespace and a single leading `+`. This deliberate tightening surfaces typoed configurations instead of applying a value that differs from the user's intent.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [x] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
cmake -S . -B build-string-parsers \
  -DCMAKE_BUILD_TYPE=Release \
  -DENABLE_DEBUG_SYMBOLS=OFF \
  -DUSE_CUDA=OFF \
  -DWITH_STORE_RUST=OFF \
  -DBUILD_BENCHMARK=OFF \
  -DBUILD_EXAMPLES=OFF \
  -DBUILD_UNIT_TESTS=ON
cmake --build build-string-parsers --parallel 128

ctest --test-dir build-string-parsers --output-on-failure \
  -R '^(default_config_test|environ_test|ascii_string_test|bool_parser_test|integer_parser_test|rpc_client_io_threads_test|registered_pinned_memory_test|utils_test|client_metrics_test|embedded_snapshot_catalog_store_test|storage_backend_test|file_storage_test)$'

cmake --build build-string-parsers-redis --target mooncake_store --parallel 128
pre-commit run --files <touched-files>
```

**Test results:**
- [x] Unit tests pass (12 focused targets)
- Full default build passed after rebasing onto current `main`.
- Redis-enabled Store target passed before the final rebase.
- NoF configuration passed; NoF compilation could not be completed on this host because the system SPDK header `spdk/env.h` is unavailable.

## Checklist

- [ ] I have performed a human self-review of every changed line
- [x] I have run pre-commit on every touched file and all applicable hooks pass
- [x] User-visible parsing changes are documented above; no separate documentation update is needed
- [x] RFC is not required for this focused refactor

## AI Assistance Disclosure

- [ ] No AI tools were used

OpenAI Codex implemented the refactor, resolved the rebase conflicts, and ran the builds, focused tests, and formatting checks. Human review is required before merge.


he-yufeng:
Spent some time in this one since it rewires several parsing sites I have touched before. Overall the consolidation looks careful, and a few of the semantic upgrades are real fixes. Posting notes instead of a rubber stamp.

What I verified as strict improvements: the old `GetEnvOr<bool>` went through `std::stoll`, so `MOONCAKE_OFFLOAD_ENABLE_EVICTION=false` threw inside stoll and silently fell back to the default (eviction stayed on). `Environ::GetBool` actually honors false now. Same class as the `enable_ssd_offload` string-bool fix from a while back. The strict `from_chars` parser also keeps the uint32 range and negative-input guards, and whitespace/plus handling matches what stoll accepted.

Two user-visible behavior shifts worth calling out in the PR description so the changelog is honest. Bool sites: a numeric value like `2` used to parse as true via stoll, now it falls back to the default with a warning. Integer sites: stoll tolerated trailing garbage (`1048576 ` with junk suffixes), the strict parser rejects those and uses the default. Both are the right call, but someone with a sloppy-but-working env file will see different behavior after upgrade.

Sites I checked that stay behavior-identical: `MC_MS_FILTERS` splitting (token_compress_off matches the old manual loop including the trailing empty entry, and the ASCII trim is equivalent to the old isspace trim for these inputs), the CRC toggle accepting the same false tokens plus a compatible superset, and `get_auto_discover` keeps its malformed-value tolerance.

Aionw:
Thanks for the careful review and for checking the migrated call sites.

I agree that strict parsing is the right behavior here: accepting `2` as `true` or silently consuming only the numeric prefix before a non-whitespace suffix can make a typoed configuration take effect differently from what the user intended.

I’ve updated the PR description to call out both compatibility changes explicitly. Environment integer parsing still accepts surrounding ASCII whitespace and a single leading `+`.

Thanks also for confirming the unchanged behavior of `MC_MS_FILTERS`, the CRC toggle, and `get_auto_discover`.

yokinoshitayoki (approved):
LGTM.

ykwd (approved):

Icedcoco (approved):

PR #3182, opened 2026-07-29
[Bugfix][Store] Reassign LOCAL_DISK replica after client restart (#3052)

### LINKED ISSUE #3052: [Bug]: client 重启后，  ssd 上的数据全部失效

ShiroKSH:
## Description

Part of #3052: a restarted client can reclaim its LOCAL_DISK replica when it reports the same non-empty transport endpoint. The owner UUID changes in the live Replica state, old SSD usage is released, and the reporting client is charged exactly once.

The recovery contract is deliberately narrow: a transport endpoint must be stable across client UUID rotations and unique to one physical local-disk instance.

## Recovery model

- No completed LOCAL_DISK replica: append a new COMPLETE replica.
- Same UUID: refresh the endpoint without changing SSD or file accounting.
- Different UUID plus the same non-empty endpoint: transfer ownership; release the old owner's SSD usage and charge the reporting client.
- Different or empty endpoint: leave the existing replica unchanged, preserving deduplication.
- Different reported size for a matching replica: reject with INVALID_PARAMS without changing metadata or accounting.
- REMOVED replica pending durable cleanup: append a new COMPLETE replica. The durable callback removes only the original replica ID.
- Oplog rejection: make no metadata or SSD-accounting change.

This remains narrower than #2777: no persistent marker or grace-state changes.

## Module

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

**Checks run:**

- `git diff --check`
- `PATH=/tmp/mooncake-clang-format-20/clang_format/data/bin:$PATH ./scripts/code_format.sh --check --base upstream/main`
- [GitHub Build & Test run 33247568340](https://github.com/kvcache-ai/Mooncake/actions/runs/33247568340): full build/package matrix and Ubuntu 22.04/24.04 wheel install/runtime tests passed (23 jobs passed, 2 skipped).

**Coverage added:**

- Non-HA restart transfer: real owner mutation, size-mismatch rejection, SSD and global file accounting, and stale-client cleanup.
- Existing different-endpoint deduplication plus empty-endpoint compatibility.
- HA durable payload contains the new owner; REMOVED-to-COMPLETE recovery survives the old durable cleanup; rejected oplog writes do not create metadata.
- Usage transitions serialize with client unregistration, and standby-restored usage is rebuilt exactly once when the owner remounts.

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used


Icedcoco:
Thanks for working on this issue. The endpoint-based recovery direction may be reasonable, but after reviewing the current implementation and testing it locally, I think we should pause the implementation and first complete the state-transition design.

I confirmed the following issues:

1. `Replica::get_descriptor()` returns a `Descriptor` by value. In `AddReplica`, references such as `reported_local_disk` and `existing_local_disk` are therefore bound to subobjects of temporary descriptors and become dangling. The subsequent reads are undefined behavior.

2. Updating `existing.get_descriptor().get_local_disk_descriptor()` does not update the underlying `Replica`. It only modifies a temporary descriptor. As a result, the real `client_id`, endpoint, and object size are not transferred.

3. The new test does not currently compile because it calls the private `MasterService::ClearInvalidHandles()` method directly from a `TEST_F`-generated subclass. The fixture is a friend, but the generated subclass is not.

4. After adding only a fixture access helper locally so that the test could run, the new test still failed: the old client retained 128 bytes, the restarted client remained at 0, the replica still belonged to the old UUID, and cleanup removed the object. In other words, the reported issue remains reproducible with the current implementation.

There is also an unresolved HA transition. `replacing_existing` considers every `LOCAL_DISK` replica regardless of status, while oplog serialization skips non-`COMPLETE` replicas. A replica already marked `REMOVED` but not yet finalized can therefore be treated as replaceable without becoming a new `COMPLETE` replica. The intended behavior for this durable window needs to be defined explicitly.

Rather than fixing these failures one by one, could you first write down the intended model for `AddReplica`? At minimum, please define the behavior for:

- no existing `LOCAL_DISK` replica;
- the same UUID reporting again;
- a different UUID reporting the same non-empty endpoint;
- a different UUID reporting a different endpoint;
- an empty endpoint;
- an existing `REMOVED` replica awaiting durable cleanup;
- a reported size different from the existing size;
- an oplog append failure.

Please also identify the invariants that must hold after each transition:

- which UUID owns the metadata;
- whether the replica is visible and `COMPLETE`;
- the old and new clients' `ssd_used_bytes`;
- the global allocated-file accounting;
- the descriptor persisted through the oplog;
- whether a previously scheduled cleanup can still remove the recovered replica.

The endpoint discriminator also needs an explicit contract: is `transport_endpoint` guaranteed to be stable across restart and unique to one physical storage instance? If either property is not guaranteed, endpoint equality alone is not sufficient recovery identity.

Once that model is coherent, please implement the transition against the real `Replica` state rather than descriptor snapshots, keep metadata/accounting/oplog changes consistent, and add focused non-HA and HA tests for the cases above. The existing different-endpoint deduplication test should remain as a compatibility check.

Finally, please run the actual test target and formatting checks before the next review. The current PR is also failing `Check code format`, and the remaining build jobs were skipped because of that failure.

ShiroKSH:
Rebased onto current `main` and addressed the transition model in the implementation:

- a new LOCAL_DISK report creates one COMPLETE replica;
- the same UUID is idempotent, while a size mismatch is rejected;
- a different UUID may take ownership only when it reports the same non-empty endpoint; accounting is released from the old client and the actual Replica is updated in place;
- a different or empty endpoint does not take ownership;
- REMOVED replicas are not eligible for takeover, so a report during durable removal creates a new COMPLETE replica;
- the post-transition oplog payload is written before metadata mutation, so an append failure leaves metadata untouched.

The focused non-HA and HA tests cover endpoint/size handling, ownership transfer, stale-client cleanup, durable-remove overlap, and oplog failure. I also synchronized the in-memory HA backend used by the asynchronous writer and reader. Fresh Linux CI is running on the rebased head.

Icedcoco:
Thanks for the follow-up. The overall transition model looks reasonable, but I found three points that would be useful to clarify before merging.

1. Could we revisit the stability assumption for `transport_endpoint`? `AddReplica` now uses a matching non-empty endpoint as the recovery identity. However, the default `real_client` configuration starts the offload RPC server with port `0`, so the OS assigns a new port after each restart. `local_rpc_addr` is then rebuilt from that auto-allocated port. In this default setup, the endpoint may change across restarts, causing `AddReplica` to return `false` and allowing the stale-cleanup path to remove the old replica. Would it make sense to use a stable, configurable endpoint or a separate persistent local-disk identity, and add a test covering the default startup configuration?

2. Could we also guard the accounting transition when the reporting client has no mounted `LocalDiskSegment`? `NotifyOffloadSuccess` currently treats the segment as optional. `AddReplica` can therefore transfer ownership and release the old client's `ssd_used_bytes`, while the caller skips charging the new client because `local_disk_segment` is null. This leaves the metadata owner updated but the SSD accounting undercharged. Should we require the segment to be mounted before applying the transfer, or make the ownership and accounting update one checked operation? A focused test for a missing or concurrently removed segment would help pin down the intended behavior.

3. Could we route all `LOCAL_DISK` state transitions through the same implementation? When an `offloading_task` exists, `NotifyOffloadSuccess` handles the object in the inline branch and does not call `AddReplica`. That branch still writes through `get_descriptor()`, which returns a copy, so the endpoint and size updates do not modify the real `Replica`. It also bypasses the new endpoint/size validation, ownership accounting, and HA oplog handling. Is this branch still required as a separate path? If so, could we align it with the same transition rules and add a test for the task-present case, including HA behavior?

One CI note: `Check code format`, docs, and wheel builds are passing. The failing `build (3.12)` and `build-flags (3.10)` jobs stop while compiling the untouched `mooncake-store/tests/master_scenario.cpp`, which still references `TenantQuotaSnapshot::used_bytes` and `reserved_bytes`; those fields were replaced by `charged_bytes` on the updated `main` branch. This looks like a base-branch mismatch rather than an error in this PR. The corresponding fix is in upstream commit `0e4483cd`:
https://github.com/kvcache-ai/Mooncake/commit/0e4483cd837893e78ea0809c5d8e51410e331f8a

Could you rebase or merge the current `main`, resolve the resulting conflicts in the Store and HA files, and rerun the full build matrix? Since the failing jobs stopped before the relevant test targets were executed, the new recovery tests still need a green CI run for validation.

mooncake-store/src/real_client.cpp
```diff
@@ -1040,9 +1046,11 @@ tl::expected<void, ErrorCode> RealClient::setup_internal(
     }
     if (enable_ssd_offload && start_offload_rpc_server) {
         // Start RPC server for offload operations (batch_get / release_buffer).
-        // Use port 0 to let the OS auto-allocate an available port.
+        // A LOCAL_DISK replica uses this endpoint as its restart identity, so
+        // it must use the configured stable port rather than an OS-assigned one.
         offload_rpc_server_ =
-            std::make_unique<coro_rpc::coro_rpc_server>(1, 0, "0.0.0.0");
+            std::make_unique<coro_rpc::coro_rpc_server>(1, local_rpc_port,
+                                                         "0.0.0.0");
```
LujhCoconut: This is a behavioral regression for embedded clients. Before this change the offload RPC server bound port 0 (OS-assigned), so multiple clients on one host never collided. Now it binds the fixed `local_rpc_port` (default 50052), and a second client on the same host — another embedded client, or a standalone `mooncake_client` whose `--port` also defaults to 50052 — will hit EADDRINUSE here, `get_errc()` becomes non-zero, and setup fails with INTERNAL_ERROR. Since the default is shared and there is no fallback, existing single-host multi-client deployments break on upgrade without an explicit per-client port.

ShiroKSH:
Addressed the embedded-client regression in `82911997` and merged current `upstream/main` in `7aa4df29`.

- Embedded SSD offload now defaults `local_rpc_port` to `0`, so its offload RPC server uses OS-assigned auto-bind again and multiple embedded clients on one host do not collide.
- A positive `local_rpc_port` remains the opt-in stable endpoint for LOCAL_DISK restart recovery.
- The merge conflict in `master_service.cpp` keeps the shared `AddReplicaImpl` transition path and adapts ownership accounting to `LocalSsdManager`.

Checks run:
- `git diff --check upstream/main...HEAD`
- `PATH=<clang-format-20-venv>/Scripts:$PATH ./scripts/code_format.sh --check --base upstream/main`
- `python -m py_compile mooncake-wheel/mooncake/mooncake_config.py mooncake-wheel/mooncake/mooncake_store_service.py`
- `python -m pytest mooncake-wheel/tests/test_mooncake_config.py mooncake-wheel/tests/test_mooncake_store_service_api.py::StoreServiceApiTest::test_start_store_service_passes_tenant_id_to_setup`

Icedcoco:
Thanks for the follow-up. The HA transition looks much more consistent now. I still have two questions around `LocalSsdManager` accounting:

1. Could we check what happens after an actual standby promotion? `PromotionContext` carries the object metadata, and `RestoreFromStandbySnapshot()` recreates the `LOCAL_DISK` replica with the new owner, but I could not find a path that restores or rebuilds `LocalSsdManager::used_bytes`. When the client mounts again, its usage starts at zero; a subsequent report with the same UUID and endpoint is treated as idempotent, so `NotifyOffloadSuccess()` does not charge it again. Am I missing another reconciliation step here? If not, would it make sense to rebuild the usage from the restored `COMPLETE LOCAL_DISK` descriptors when the client remounts, and add a standby apply → promotion → remount test?

2. Could we also make the mounted-client check and the accounting update part of one checked transition? `NotifyOffloadSuccess()` currently calls `GetUsage(client_id)` before changing the owner, then calls `AdjustUsedBytes()` afterward. Since these are separate manager operations, the client could be unregistered in between. In that case, the PUT_END and metadata may already contain the new owner, and the old owner’s usage may have been released, while charging the new owner silently fails because the return value is ignored. Would holding the client lifecycle access across the transition, or exposing a single checked accounting operation from `LocalSsdManager`, be a reasonable way to close this window? A focused concurrent-unregister test would also help clarify the intended behavior.

LujhCoconut:
I've thought about this one quite a bit before commenting, and want to start by saying the direction still looks right to me: endpoint-as-discriminator is what @he-yufeng proposed in #3052 and what we agreed on there, and the transition model has come out really clean after these rounds.

One thing I keep coming back to — and I want to be careful here, because it may read as contradicting my earlier review comment on this file. It isn't: binding port 0 by default is the right behavior, the EADDRINUSE regression was real, and your fix is correct. But that very fix makes the endpoint ephemeral by default, and @he-yufeng's original caveat in #3052 was that endpoint matching only works "when `local_rpc_addr_` is stable across restarts". With port 0, an embedded client gets a fresh port after every restart: after a UUID rotation the report never matches (`master_service.cpp:4745`), the stale replica ends up deleted rather than transferred, and since `ScanMeta` only runs once at startup, nothing re-registers the data. I suppose both of my comments are really the same point: **the port can't be the disk's identity** — pinned ports collide on one host, auto-bound ports change on restart. So recovery via endpoint matching is inherently opt-in (a stable `local_rpc_port`, or standalone `mooncake_client` on its main `--port` — that part is nice), and the default embedded path keeps the original #3052 behavior.

While thinking this through, I also came across #2777, which I honestly hadn't noticed before — and it caught my eye precisely because it dissolves the tension above: a persistent on-disk marker gives the disk a stable identity *without touching the port at all*, so recovery works by default and same-host co-location is unaffected. It seems to have stalled waiting for a direction check, and would need a rebase plus some hardening — but it feels like the natural long-term answer here.

So, gently: would it make sense to soften `Fixes #3052` to something like "Part of #3052" (the docs already describe the opt-in honestly), merge this as the opt-in path, and then revisit #2777's approach under #2306 for default-on recovery? Happy to be talked out of it — I mainly don't want #3052 to auto-close while the default scenario is still open, and it'd be a shame for either PR to stall.

ShiroKSH:
Addressed the remaining accounting cases in `2c2c74d2` and merged current `upstream/main`.

- `MountLocalDiskSegment` now reconciles `used_bytes` from restored COMPLETE LOCAL_DISK descriptors for both a new registration and a registration already restored from snapshot state. The reconciliation runs under the exclusive snapshot transition, so concurrent replica writes cannot be missed or double-counted.
- `AddReplicaImpl` now runs the metadata/oplog transition and the new owner's usage delta through one `LocalSsdManager::ApplyUsageTransition` lifecycle access. Concurrent unregister waits for an admitted transition; a transition that has not acquired the client is rejected before metadata changes.
- Added a standby restore → existing-client remount → idempotent report regression test and a deterministic lifecycle test that proves unregister waits for the checked usage transition.
- Updated the PR description from `Fixes #3052` to `Part of #3052` to keep the default embedded-port recovery gap open.

Checks:

- `git diff --check upstream/main...HEAD`
- `./scripts/code_format.sh --check --base upstream/main` with clang-format 20.1.8
- Fresh GitHub build matrix is running on `2c2c74d2`; code-format, spelling, path, manifest, and release-gate checks are already green.

mooncake-store/src/real_client.cpp
ShiroKSH: Fixed in `82911997`: embedded SSD offload uses port `0` by default again, preserving OS auto-bind for multiple same-host clients. A positive `local_rpc_port` remains the explicit opt-in for a stable recovery endpoint.

ShiroKSH:
Fresh CI on `1517c150` is green: [run 33247568340](https://github.com/kvcache-ai/Mooncake/actions/runs/33247568340) completed with 23 successful jobs, 2 skipped jobs, and no failures. This includes the full Linux build/package matrix plus Ubuntu 22.04/24.04 wheel install and runtime coverage.

LujhCoconut:
Could you rebase onto current main before this merges? Heads-up that this is more than a textual conflict: since your last push, main landed the recoverable-client-liveness work (#2991, Sept 9). It added a `ClientLivenessRecord` affiliation to `LocalDiskReplicaData` (`replica.h:217-224`), retaining-guard wrappers in `AddReplica` (`master_service.cpp:5328-5340`) and `NotifyOffloadSuccess` (`:8311-8323`), and rewrote `MountLocalDiskSegment` around the liveness machinery (`:8121-8164`). Two integration points follow.

First, the ownership transfer in `AddReplicaImpl` updates `client_id` via `update_local_disk_owner`, but the replica still carries the *old* owner's liveness record. The stale sweeps classify by `client_id` (`has_stale_local_disk_client`), so they will correctly keep the transferred replica — the breakage is on the read path instead: `getDescriptorIfAvailable` requires `client_liveness && record->IsServing()`, and once the old owner's record ages past ACTIVE, `GetReplicaList` filters the replica out and, when it is the only copy, reads fail with `REPLICA_IS_NOT_READY`, while the new owner keeps being charged `used_bytes`. The transferred replica becomes a permanent zombie — #3052 in a new shape. The fix is to rebind the replica's liveness record to the new owner during the transfer, i.e. `bindClientLiveness` with the reporting client's record on top of the `client_id` update; note your `AddReplicaImpl` also needs to nest into main's guard+bind structure rather than bypass it (main still keeps `HasMountedLocalDiskSegment` and uses it in `AddReplicaForRetainedClient`).

Second, the `used_bytes` rebuild scan needs re-fitting into the new `MountLocalDiskSegment` structure, which looks nothing like the version you patched.

On tests: as written they cannot catch this — the sweeps classify by the already-updated `client_id`, and the old owner's record stays ACTIVE for the whole test, so the `IsServing()` gate never trips. A test that actually pins it would age the old owner's record out of ACTIVE (or construct it OFFLINE) after the transfer, then assert `GetReplicaList` still returns the replica; without the rebind that fails deterministically.

One smaller thing to fold into the same pass, this one introduced on your own head: `RegisterClientRpcService` now registers the offload handlers only when `enable_offload && start_offload_rpc_server`, while main registers them unconditionally — so a standalone client with offload enabled but the flag off advertises an endpoint with no handler behind it, and offload reads silently break after upgrade. Either keep the unconditional registration or fail loudly at setup.

The transition model itself still looks right to me, and once the rebase lands with the liveness rebind I'm happy to take another pass.

PR #3186, opened 2026-07-29
[Bugfix] Fix stranded slices and misreported task status on RDMA submit failure

SongOf:
## Description
  
  When `RdmaTransport::submitTransferTask()` aborts partway on device-selection
  failure (`!found_device`) or an inactive device:

  1. **Permanently stuck batch.** Slices already queued in `slices_to_post` were
     neither posted nor failed, but their `slice_count` was already incremented.
     Nothing ever advances `success + failed` for an unposted slice, so
     `getTransferStatus()` reports `WAITING` forever and `freeBatchID()` returns
     `BatchBusy` indefinitely — the whole batch leaks until process exit.
  2. **Never-attempted tasks misreported as COMPLETED.** Tasks after the failing
     one are never processed, so their `slice_count` stays 0, which trivially
     satisfies `success + failed == slice_count`.

  Fix — the abort path now leaves every task with a well-defined terminal state:

  - The slice that failed device selection is counted into `slice_count` and
    failed via `Slice::markFailed()`; counting happens before any `markFailed()`
    call, since `check_batch_completion()` compares `completed_slice_count`
    against `slice_count` and would otherwise complete the task prematurely.
  - The failing task's own queued slices are fast-failed the same way rather
    than posted (the caller retries the whole task anyway); other tasks' queued
    slices are unrelated and flushed normally. Every slice in `slice_list` is
    thus counted and terminal, so `failed_slice_count >= 1` guarantees `FAILED`.
  - Never-reached sibling tasks get a new `TransferTask::submit_failed` flag,
    consulted by both `getTransferStatus()` overloads. Written release / read
    acquire: plain `volatile` gives no cross-core ordering on ARM/POWER.
  - Under `USE_EVENT_DRIVEN_COMPLETION`, siblings have no slice to drive
    `check_batch_completion()`, so the abort path replicates its task-finished
    tail for them (`has_failure`, `finished_task_count`, and for the batch's
    last task the mutex-guarded `is_finished` publish + `notify_all`). Waiters
    are now woken by the abort instead of hitting the 60s timeout.
  - `RdmaContext::submitPostSend()` and its destructor are made `virtual` as a
    test seam so the flush path is exercisable without RDMA hardware. One
    virtual call per flushed batch; no production behavior change.

  ## Module
  
  - [ ] CI/CD
  - [ ] Docs
  - [ ] Other

  ## Type of Change

  - [x] Bug fix
  - [ ] New feature
  - [ ] Refactor
  - [ ] Breaking change
  - [ ] Documentation update
  - [ ] Performance improvement
  - [ ] Other

  ## How Has This Been Tested?

  Six new tests in `rdma_transport_submit_task_test.cpp` (the existing
  double-free regression test still passes). A `FakeCompletionContext`
  overrides the new virtual `submitPostSend()` to complete slices via
  `markSuccess()`, so no RDMA hardware is needed.

  Default build (5 tests): queued-slice fast-fail with both slices counted and
  failed; first-slice failure (accounting starts from `slice_count == 0`);
  never-reached sibling reported `FAILED` not `COMPLETED`; and a live
  three-task call — success / partial failure / never reached — checked per
  task through both `getTransferStatus()` overloads.

  `USE_EVENT_DRIVEN_COMPLETION` build (7 tests): two more with a real thread
  blocked on `completion_cv`, asserting it is woken by the abort itself with
  `finished_task_count` complete — one for the mixed batch (each task accounted
  for by a different mechanism), one for the single first-slice failure. Both
  cap the wait at 10s so a missing notification fails instead of hanging;
  before the fix they fail exactly that way.

  **Test commands:**
  ```bash
  # Linux container (Ubuntu 22.04, aarch64), no RDMA hardware required
  cmake -S . -B build -DBUILD_UNIT_TESTS=ON
  cmake --build build --target rdma_transport_submit_task_test -j4
  ./build/mooncake-transfer-engine/tests/rdma_transport_submit_task_test

  
  Test results:
  - [x] Unit tests pass — 5/5 default, 7/7 with USE_EVENT_DRIVEN_COMPLETION;
  full ctest (26 tests) matches the pre-change baseline exactly (the 3
  cmake --build build -j4 && (cd build && ctest --output-on-failure)
  
  Test results:
  - [x] Unit tests pass — 5/5 default, 7/7 with USE_EVENT_DRIVEN_COMPLETION;
  full ctest (26 tests) matches the pre-change baseline exactly (the 3
  failures are pre-existing environment-only issues in the container:
  tcp_transport_test / transfer_metadata_test need an etcd metadata
  server, memory_location_test needs NUMA privileges).
  
  Checklist
  
  AI Assistance Disclosure
  
  - [ ] No AI tools were used

### DIFF
--- a/mooncake-transfer-engine/src/transport/rdma_transport/rdma_transport.cpp
+++ b/mooncake-transfer-engine/src/transport/rdma_transport/rdma_transport.cpp
@@ -744,10 +744,93 @@ Status RdmaTransport::submitTransfer(
     return submitTransferTask(task_list);
 }
 
+void RdmaTransport::failQueuedSlicesForTask(
+    std::unordered_map<std::shared_ptr<RdmaContext>, std::vector<Slice *>>
+        &slices_to_post,
+    TransferTask *task) {
+    for (auto &entry : slices_to_post) {
+        auto &pending = entry.second;
+        std::vector<Slice *> keep;
+        keep.reserve(pending.size());
+        for (auto *s : pending) {
+            if (s->task == task)
+                s->markFailed();
+            else
+                keep.push_back(s);
+        }
+        pending.swap(keep);
+    }
+}
+
 Status RdmaTransport::submitTransferTask(
     const std::vector<TransferTask *> &task_list) {
     std::unordered_map<std::shared_ptr<RdmaContext>, std::vector<Slice *>>
         slices_to_post;
+
+    // Aborts submission of `task` (at `task_index` in task_list) after
+    // `failed_slice` -- already in task.slice_list but not yet counted --
+    // failed device selection. task's own slices still in slices_to_post
+    // never touched hardware and the whole task must be retried, so fail
+    // them in place instead of posting them; other tasks' queued slices
+    // are flushed normally. failed_slice itself is counted into
+    // slice_count and failed like any other slice, so every slice in
+    // slice_list is counted and terminal: failed_slice_count >= 1
+    // guarantees a FAILED report, and under USE_EVENT_DRIVEN_COMPLETION
+    // its markFailed() drives the task's is_finished /
+    // finished_task_count bookkeeping once all counted slices are
+    // terminal. Every not-yet-reached task after `task` is marked
+    // submit_failed so getTransferStatus() reports FAILED instead of
+    // misreading its slice_count == 0 as COMPLETED; `task` gets the flag
+    // too, as a uniform "submission aborted" marker.
+    //
+    // The sibling tasks have no slice passing through markFailed(), so
+    // under USE_EVENT_DRIVEN_COMPLETION the loop below replicates
+    // check_batch_completion()'s task-finished tail for them: account each
+    // as finished and, for the batch's last, publish completion and notify
+    // exactly the way the slice path does.
+    auto failTaskSubmission = [&](TransferTask &task, size_t task_index,
+                                  Slice *failed_slice) {
+        // Count failed_slice before any markFailed() below:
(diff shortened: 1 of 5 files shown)


mooncake-transfer-engine/src/transport/rdma_transport/rdma_transport.cpp
```diff
@@ -744,10 +744,62 @@ Status RdmaTransport::submitTransfer(
...
+        // getTransferStatus(): plain volatile gives no cross-core ordering
+        // on ARM/POWER, so submit_failed could otherwise become visible
+        // before markFailed()'s failed_slice_count increments above.
+        __atomic_store_n(&task.submit_failed, true, __ATOMIC_RELEASE);
+        for (size_t j = task_index + 1; j < task_list.size(); ++j) {
+            __atomic_store_n(&task_list[j]->submit_failed, true,
+                             __ATOMIC_RELEASE);
+            __atomic_store_n(&task_list[j]->is_finished, true,
+                             __ATOMIC_RELEASE);
+        }
```
alogfans, line 831: Under `USE_EVENT_DRIVEN_COMPLETION`, these untouched sibling tasks are only marked as `is_finished`; they never update `BatchDesc::finished_task_count` or notify `completion_cv`. As a result, an event-driven waiter may remain blocked indefinitely. Please update the batch-level completion accounting and notify the waiter when the final task finishes.

SongOf: Fixed and added necessary unit tests. @alogfans

Issue #3111, opened 2026-07-24
[RFC]:  Manifest-driven heterogeneous model weight resharding and storage


Bo-Vincent:
### Summary

This RFC proposes manifest-driven heterogeneous model-weight resharding and storage for Mooncake. A source and target may use different `TP`, `PP`, `EP`, and `DP` layouts while sharing one model-semantic manifest, N-D logical-box planner, and backend-neutral transfer-region contract.

The proposed contract supports three paths:

*   live runtime GPU-to-GPU (G2G) transfer;
    
*   runtime-to-Store upload; and
    
*   Store-to-runtime load.
    

Live G2G and Store-to-runtime use a `LogicalTransferPlan`. Runtime-to-Store uses a separate `WeightUploadPlan`, while sharing tensor identity, DP replica selection, coverage, and fragment semantics with the load path.

The framework runtime remains the source of truth for model semantics and live GPU addresses. Mooncake consumes explicit placement and runtime-binding data, computes N-D overlaps in global logical-tensor coordinates, and lowers the result into Transfer Engine (TE), Store, or future executor operations.

The first version only accepts byte-compatible layouts. A mismatch in dtype, shape, item size, or `layout_fingerprint` fails closed.

### Motivation

RL rollouts, elastic scale-out, and inference replacement can require a conversion such as:

```text
TP4 / PP2 / EP8 / DP2  ->  TP8 / PP4 / EP2 / DP4

```

The existing one-dimensional TP overlap model cannot express all of the following together:

*   source and target sharding different tensor dimensions;
    
*   pipeline-parallel layer or tensor ownership changes;
    
*   expert-parallel and tensor-parallel changes in the same tensor; and
    
*   a common load contract for live GPU and persistent Store sources.
    

Mooncake already provides TE, Store, Group, leases, memory registration, and ranged transfer. What is missing is a stable contract between framework-owned model semantics and Mooncake-owned physical transfer and storage operations.

### Goals and non-goals

Goals:

*   Support joint TP, PP, EP, and DP changes, including different source and target shard dimensions.
    
*   Persist source-native TP/PP/EP fragments as an immutable weight revision that can serve future target layouts.
    
*   Avoid model-wide, tensor-wide, or all-expert all-gathers.
    
*   Avoid a model-sized temporary GPU or CPU buffer.
    
*   Bind one logical load plan to either live-runtime or Store fragments; keep a dedicated upload plan for the write path.
    
[...]

PR #3187, opened 2026-07-29
[Reshard] Add resource and model-weight manifest contracts

Bo-Vincent:
## Description

Related RFC: #3111

This PR introduces the manifest contract layer for `mooncake-reshard`. It defines framework-neutral resource identity and the first model-weight specialization without coupling Mooncake Core to SGLang, vLLM, PyTorch, model-specific parameter names, Mooncake Store, or a transfer backend.

The central contract is that one `WeightPlacementManifest` describes one complete, address-free global logical placement of a model-weight generation. Framework-owned adapters on each worker export local `WeightPlacementPart` values; a collection barrier assembles and validates the exact part set before a stable placement ID and digest are produced. Live GPU addresses remain in per-participant `WeightRuntimeBindingManifest` values and are validated immediately before use.

Later PRs add the backend-neutral N-D planner, Store integration, and Transfer Engine lowering under the same `mooncake-reshard` namespace.

#### 1. Resource-neutral contracts

- `ResourceManifest`: resource identity and kind;
- `PlacementManifest`: stable, address-free placement identity and digest;
- `RuntimeBindingManifest`: ephemeral physical state fenced by placement identity, runtime instance, generation, and lease;
- `ResourceKind.MODEL_WEIGHT`, with `KV_CACHE` reserved for a future specialization;
- shared typed placement-to-binding identity validation.

Model revision, tensor geometry, layout, parallel ownership, and weight generation remain weight-specific.

#### 2. Complete model-weight placement

`WeightPlacementManifest` contains:

- `ParallelTopology` with TP, PP, EP, and DP sizes plus an explicit participant-to-rank mapping;
- one `WeightPlacementPart` for every selected participant;
- canonical tensor descriptors and N-D logical fragments;
- model revision, weight generation, and placement-set identity;
- a placement ID and digest derived only after global validation succeeds.

`ParallelTopology.world_size` is the selected participant count, not the product of TP/PP/EP/DP sizes. Parallel axes may share workers, and a placement may select one complete DP replica while preserving the runtime's declared `dp_size`.

Global validation rejects missing or duplicate participants, mixed topology or generation identity, rank mismatches, conflicting tensor descriptors, overlapping logical boxes, and incomplete tensor coverage. PP owners are validated independently, so fragments from different PP owners cannot be combined to manufacture coverage.

#### 3. Per-participant runtime binding

Each live participant exports one `WeightRuntimeBindingManifest` that:

- identifies its `participant_id`;
- attests the complete global placement ID and digest;
- binds exactly that participant's logical fragments to worker, endpoint, device, address, and allocation owner;
- carries runtime instance, generation, and lease fences.

Empty participants require no physical binding. Any participant referenced by a later execution plan must provide a valid binding. GPU addresses and owner objects never enter the serializable placement.

#### 4. Framework-neutral tensor semantics

The weight contract records:

- stable tensor identity, global shape, dtype, and item size;
- layout fingerprint plus optional layer and expert identity;
- N-D logical boxes using `global_offset` and `local_shape`;
- TP, PP, EP, and DP owner coordinates;
- canonical `shard_dims` with explicit split, replication, and ownership axis types;
- contiguous runtime views with address-range and alias validation.

Mooncake does not infer layer, expert, partition, or layout semantics from parameter names. Framework-owned adapters must normalize those facts into the canonical typed contracts before entering Mooncake Core; framework objects are never inspected by the core.

#### 5. Packaging and documentation

The tracked source lives under `mooncake-reshard/python/mooncake/reshard`. The combined Mooncake wheel stages this package during the build and removes the staging copy on exit. The design document describes global placement assembly, identity, binding, and lifecycle boundaries.

### Why Split Placement, Parts, and Binding?

- `WeightPlacementPart` is the framework-local contribution produced independently by each selected worker.
- `WeightPlacementManifest` is the complete global logical fact used by planning and control-plane identity.
- `WeightRuntimeBindingManifest` is per-participant live physical state.

This lets logical planning start from complete source and target placements without waiting for every target GPU address. Physical binding remains late, generation-fenced, and lease-fenced.

### Non-Goals

This PR does not:

- generate heterogeneous reshard regions;
- execute Transfer Engine operations;
- persist weights in Mooncake Store;
- implement a KVCache manifest or KV resharding;
- infer model semantics from tensor names;
- define discovery, activation, rollback, or control-plane policy;
- integrate M2N, scatter batching, or another optimized executor.

Planner, Store, TE, framework-adapter, and KV-specialization work follow as separate changes.

## Module

- [x] CI/CD
- [x] Docs
- [ ] Other

Feature ownership remains under `mooncake-reshard`; wheel and CI changes only publish and test that module.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] Refactor
- [ ] Breaking change
- [x] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**

```bash
python -m pytest -q mooncake-reshard/tests

python -m ruff format --check mooncake-reshard
python -m ruff check mooncake-reshard
python -m compileall -q mooncake-reshard/python
bash -n scripts/build_wheel.sh
bash -n scripts/test_installation.sh
git diff --check
```

**Test results:**

- [x] Unit tests pass: `186 passed`
- [ ] Integration tests pass (not applicable to this contract-only PR)
- [x] Manual review completed for package staging, installed import checks, and commit boundaries

Coverage includes public API identity, canonical typed boundaries, strict JSON decoding, topology and participant validation, complete global placement assembly, PP/DP coverage rules, split/replicated/ownership axis geometry, placement serialization and identity, N-D geometry, aliases, per-participant binding, physical address ranges, and lease/generation fencing.

## Commit Structure

1. Define resource and model-weight manifest contracts.
2. Cover manifest assembly and runtime-binding behavior.
3. Stage the reshard package in the Mooncake wheel.
4. Document placement and binding contracts.
5. Register module ownership and labels.

## Checklist

- [ ] I have formatted my code using `./scripts/code_format.sh` (not applicable; no C/C++ changed)
- [x] I have run the relevant Ruff checks

## AI Assistance Disclosure

- [ ] No AI tools were used
- [x] AI tools were used

AI assistance was used to inspect repository conventions, implement and organize the contracts and tests, review the change, and run validation. The submitter has reviewed the result and is responsible for understanding and maintaining it.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -12,6 +12,7 @@
 /mooncake-integration/transfer_engine @ShangmingCai @alogfans
 /mooncake-integration/store @ykwd @stmatengss
 /mooncake-pg @UNIDY2002 @ympcMark @yuechen-sys
+/mooncake-reshard @ShangmingCai @stmatengss
 /mooncake-store @ykwd @stmatengss @XucSh @YiXR
 /mooncake-store/*/ha/ @Libotry @YiXR @00fish0
 /mooncake-transfer-engine @alogfans @doujiang24 @chestnut-Q @staryxchen
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -12,7 +12,7 @@
 /mooncake-integration/transfer_engine @ShangmingCai @alogfans
 /mooncake-integration/store @ykwd @stmatengss @zxpdemonio
 /mooncake-pg @UNIDY2002 @ympcMark @yuechen-sys
-/mooncake-reshard @ShangmingCai @stmatengss
+/mooncake-reshard @ShangmingCai @stmatengss @Bo-Vincent @zxpdemonio 
 /mooncake-store @ykwd @stmatengss @XucSh @YiXR
 /mooncake-store/*/ha/ @Libotry @YiXR @00fish0 @Icedcoco
 /mooncake-transfer-engine @alogfans @doujiang24 @chestnut-Q @staryxchen
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -12,6 +12,7 @@
 /mooncake-integration/transfer_engine @ShangmingCai @alogfans
 /mooncake-integration/store @ykwd @stmatengss @zxpdemonio
 /mooncake-pg @UNIDY2002 @ympcMark @yuechen-sys
+/mooncake-reshard @ShangmingCai @stmatengss
 /mooncake-store @ykwd @stmatengss @XucSh @YiXR
 /mooncake-store/*/ha/ @Libotry @YiXR @00fish0 @Icedcoco
 /mooncake-transfer-engine @alogfans @doujiang24 @chestnut-Q @staryxchen
--- a/mooncake-reshard/python/mooncake/reshard/weight/binding.py
+++ b/mooncake-reshard/python/mooncake/reshard/weight/binding.py
@@ -1,0 +1,142 @@
+"""Validation between logical weight placement and runtime locations."""
+
+from __future__ import annotations
+
+from ..contracts import RuntimeInstanceId, validate_resource_binding_identity
+from .placement import WeightPlacementManifest
+from .runtime import WeightRuntimeBindingManifest
+from .types import canonical_strides_bytes, require_manifest_items
+from .validation import _validate_runtime_binding_address_ranges
+
+
+def validate_runtime_binding(
+    placement: WeightPlacementManifest,
+    binding: WeightRuntimeBindingManifest,
+) -> None:
+    """Validate an ephemeral binding against one exact logical placement."""
+
+    if not isinstance(placement, WeightPlacementManifest):
+        raise ValueError("placement must be a WeightPlacementManifest")  # noqa: TRY004
+    if not isinstance(binding, WeightRuntimeBindingManifest):
+        raise ValueError("binding must be a WeightRuntimeBindingManifest")  # noqa: TRY004
+    validate_resource_binding_identity(placement, binding)
+    if placement.revision != binding.revision:
+        raise ValueError("placement and runtime binding revision differ")
+
+    try:
+        placement_part = next(
+            part
+            for part in placement.parts
+            if part.participant_id == binding.participant_id
+        )
+    except StopIteration as error:
+        raise ValueError(
+            f"unknown runtime binding participant: {binding.participant_id}"
+        ) from error
+
+    placement_by_id = {
+        fragment.placement_fragment_id: fragment
+        for fragment in placement_part.fragments
+    }
+    binding_by_id = {
+        fragment.placement_fragment_id: fragment for fragment in binding.fragments
+    }
+    tensor_by_id = {tensor.tensor_id: tensor for tensor in placement.tensors}
+    unknown = sorted(binding_by_id.keys() - placement_by_id.keys())
+    if unknown:
+        raise ValueError(f"unknown placement fragment in runtime binding: {unknown[0]}")
+    missing = sorted(placement_by_id.keys() - binding_by_id.keys())
+    if missing:
+        raise ValueError(f"missing placement fragment in runtime binding: {missing[0]}")
+
+    for placement_fragment in placement_part.fragments:
+        runtime_fragment = binding_by_id[placement_fragment.placement_fragment_id]
+        if runtime_fragment.nbytes != placement_fragment.nbytes:
+            raise ValueError(
+                "runtime binding byte size does not match placement: "
+                f"{placement_fragment.placement_fragment_id}"
+            )
+        if runtime_fragment.local_shape != placement_fragment.local_shape:
+            raise ValueError(
+                "runtime binding local_shape does not match placement: "
+                f"{placement_fragment.placement_fragment_id}"
+            )
+        tensor = tensor_by_id[placement_fragment.tensor_id]
+        if runtime_fragment.itemsize != tensor.itemsize:
+            raise ValueError(
+                "runtime binding itemsize does not match placement: "
+                f"{placement_fragment.placement_fragment_id}"
+            )
+        expected_strides_bytes = canonical_strides_bytes(
+            placement_fragment.local_shape,
+            tensor.itemsize,
+        )
+        if runtime_fragment.strides_bytes != expected_strides_bytes:
+            raise ValueError(
+                "runtime binding stride does not describe a canonical "
+                "contiguous view: "
+                f"{placement_fragment.placement_fragment_id}"
+            )
+
+    _validate_runtime_binding_address_ranges(
+        instance_id=binding.instance_id,
+        tensors=placement.tensors,
+        placements=placement_part.fragments,
+        bindings=binding.fragments,
+    )
+
+
+def _validate_runtime_binding_subset(
+    placement: WeightPlacementManifest,
+    bindings: object,
+) -> tuple[WeightRuntimeBindingManifest, ...]:
+    """Validate one or more participants and their shared address spaces."""
+
+    items = require_manifest_items(
+        bindings,
+        "runtime bindings",
+        WeightRuntimeBindingManifest,
+    )
+    participant_ids = [item.participant_id for item in items]
+    if len(participant_ids) != len(set(participant_ids)):
+        raise ValueError("duplicate runtime binding participant")
+    for binding in items:
+        validate_runtime_binding(placement, binding)
+
+    part_by_participant = {part.participant_id: part for part in placement.parts}
+    by_instance: dict[RuntimeInstanceId, list[WeightRuntimeBindingManifest]] = {}
+    for binding in items:
+        by_instance.setdefault(binding.instance_id, []).append(binding)
+    for instance_id, instance_bindings in by_instance.items():
+        _validate_runtime_binding_address_ranges(
+            instance_id=instance_id,
+            tensors=placement.tensors,
+            placements=tuple(
+                fragment
+                for binding in instance_bindings
+                for fragment in part_by_participant[binding.participant_id].fragments
+            ),
+            bindings=tuple(
+                fragment
+                for binding in instance_bindings
+                for fragment in binding.fragments
+            ),
+        )
+    return items
+
+
+def validate_runtime_bindings(
+    placement: WeightPlacementManifest,
+    bindings: object,
+) -> None:
+    """Validate the exact runtime binding set for a complete placement."""
+
+    items = _validate_runtime_binding_subset(placement, bindings)
+    expected = {part.participant_id for part in placement.parts if part.fragments}
+    actual = {item.participant_id for item in items}
+    missing = sorted(expected - actual)
+    if missing:
+        raise ValueError(f"missing runtime binding participant: {missing[0]}")
+    unknown = sorted(actual - expected)
+    if unknown:
+        raise ValueError(f"unknown runtime binding participant: {unknown[0]}")
--- a/mooncake-reshard/python/mooncake/reshard/weight/binding.py
+++ b/mooncake-reshard/python/mooncake/reshard/weight/binding.py
@@ -1,0 +1,136 @@
+"""Validation between logical weight placement and runtime locations."""
+
+from __future__ import annotations
+
+from ..contracts import validate_resource_binding_identity
+from .placement import WeightPlacementManifest
+from .runtime import WeightRuntimeBindingManifest
+from .types import _canonical_stride, _require_sequence
+from .validation import _validate_runtime_binding_address_ranges
+
+
+def validate_runtime_binding(
+    placement: WeightPlacementManifest,
+    binding: WeightRuntimeBindingManifest,
+) -> None:
+    """Validate an ephemeral binding against one exact logical placement."""
+
+    validate_resource_binding_identity(placement, binding)
+    if placement.revision != binding.revision:
+        raise ValueError("placement and runtime binding revision differ")
+
+    try:
+        placement_part = next(
+            part
+            for part in placement.parts
+            if part.participant_id == binding.participant_id
+        )
+    except StopIteration as error:
+        raise ValueError(
+            f"unknown runtime binding participant: {binding.participant_id}"
+        ) from error
+
+    placement_by_id = {
+        fragment.placement_fragment_id: fragment
+        for fragment in placement_part.fragments
+    }
+    binding_by_id = {
+        fragment.placement_fragment_id: fragment for fragment in binding.fragments
+    }
+    tensor_by_id = {tensor.tensor_id: tensor for tensor in placement.tensors}
+    unknown = sorted(binding_by_id.keys() - placement_by_id.keys())
+    if unknown:
+        raise ValueError(f"unknown placement fragment in runtime binding: {unknown[0]}")
+    missing = sorted(placement_by_id.keys() - binding_by_id.keys())
+    if missing:
+        raise ValueError(f"missing placement fragment in runtime binding: {missing[0]}")
+
+    for placement_fragment in placement_part.fragments:
+        runtime_fragment = binding_by_id[placement_fragment.placement_fragment_id]
+        if runtime_fragment.nbytes != placement_fragment.nbytes:
+            raise ValueError(
+                "runtime binding byte size does not match placement: "
+                f"{placement_fragment.placement_fragment_id}"
+            )
+        if runtime_fragment.local_shape != placement_fragment.local_shape:
+            raise ValueError(
+                "runtime binding local_shape does not match placement: "
+                f"{placement_fragment.placement_fragment_id}"
+            )
+        tensor = tensor_by_id[placement_fragment.tensor_id]
+        if runtime_fragment.itemsize != tensor.itemsize:
+            raise ValueError(
+                "runtime binding itemsize does not match placement: "
+                f"{placement_fragment.placement_fragment_id}"
+            )
+        expected_strides_bytes = tuple(
+            stride * tensor.itemsize
+            for stride in _canonical_stride(placement_fragment.local_shape)
+        )
+        if runtime_fragment.strides_bytes != expected_strides_bytes:
+            raise ValueError(
+                "runtime binding stride does not describe a canonical "
+                "contiguous view: "
+                f"{placement_fragment.placement_fragment_id}"
+            )
+
+    _validate_runtime_binding_address_ranges(
+        instance_id=binding.instance_id,
+        tensors=placement.tensors,
+        placements=placement_part.fragments,
+        bindings=binding.fragments,
+    )
+
+
+def _validate_runtime_binding_subset(
+    placement: WeightPlacementManifest,
+    bindings: tuple[WeightRuntimeBindingManifest, ...],
+) -> tuple[WeightRuntimeBindingManifest, ...]:
+    """Validate one or more participants and their shared address spaces."""
+
+    items = tuple(_require_sequence(bindings, "runtime bindings"))
+    if not all(isinstance(item, WeightRuntimeBindingManifest) for item in items):
+        raise ValueError("runtime bindings must contain WeightRuntimeBindingManifest")
+    participant_ids = [item.participant_id for item in items]
+    if len(participant_ids) != len(set(participant_ids)):
+        raise ValueError("duplicate runtime binding participant")
+    for binding in items:
+        validate_runtime_binding(placement, binding)
+
+    part_by_participant = {part.participant_id: part for part in placement.parts}
+    by_instance: dict[str, list[WeightRuntimeBindingManifest]] = {}
+    for binding in items:
+        by_instance.setdefault(binding.instance_id, []).append(binding)
+    for instance_id, instance_bindings in by_instance.items():
+        _validate_runtime_binding_address_ranges(
+            instance_id=instance_id,
+            tensors=placement.tensors,
+            placements=tuple(
+                fragment
+                for binding in instance_bindings
+                for fragment in part_by_participant[binding.participant_id].fragments
+            ),
+            bindings=tuple(
+                fragment
+                for binding in instance_bindings
+                for fragment in binding.fragments
+            ),
+        )
+    return items
+
+
+def validate_runtime_bindings(
+    placement: WeightPlacementManifest,
+    bindings: tuple[WeightRuntimeBindingManifest, ...],
+) -> None:
+    """Validate the exact runtime binding set for a complete placement."""
+
+    items = _validate_runtime_binding_subset(placement, bindings)
+    expected = {part.participant_id for part in placement.parts if part.fragments}
+    actual = {item.participant_id for item in items}
+    missing = sorted(expected - actual)
+    if missing:
+        raise ValueError(f"missing runtime binding participant: {missing[0]}")
+    unknown = sorted(actual - expected)
+    if unknown:
+        raise ValueError(f"unknown runtime binding participant: {unknown[0]}")
--- a/mooncake-reshard/python/mooncake/reshard/weight/runtime.py
+++ b/mooncake-reshard/python/mooncake/reshard/weight/runtime.py
@@ -1,0 +1,80 @@
+"""Canonical ephemeral runtime binding contracts."""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+from ..contracts import (
+    LeaseId,
+    ParticipantId,
+    PlacementId,
+    ResourceId,
+    ResourceKind,
+    RevisionId,
+    RuntimeInstanceId,
+)
+from .types import (
+    RuntimeBindingFragment,
+    _require_nonempty_string,
+    _require_u64,
+    require_manifest_items,
+    require_sha256_digest,
+)
+
+
+@dataclass(frozen=True)
+class WeightRuntimeBindingManifest:
+    """Ephemeral physical locations and lifetime fence for one placement."""
+
+    resource_id: ResourceId
+    placement_id: PlacementId
+    placement_digest: str
+    instance_id: RuntimeInstanceId
+    generation: int
+    lease_id: LeaseId
+    revision: RevisionId
+    participant_id: ParticipantId
+    fragments: tuple[RuntimeBindingFragment, ...]
+
+    @property
+    def resource_kind(self) -> ResourceKind:
+        """Identify this binding as model weight data."""
+
+        return ResourceKind.MODEL_WEIGHT
+
+    def __post_init__(self) -> None:
+        for name in (
+            "resource_id",
+            "placement_id",
+            "instance_id",
+            "lease_id",
+            "revision",
+            "participant_id",
+        ):
+            _require_nonempty_string(getattr(self, name), name)
+        require_sha256_digest(self.placement_digest, "placement_digest")
+        _require_u64(self.generation, "generation")
+        fragments = require_manifest_items(
+            self.fragments,
+            "WeightRuntimeBindingManifest fragments",
+            RuntimeBindingFragment,
+        )
+        object.__setattr__(
+            self,
+            "fragments",
+            tuple(
+                sorted(
+                    fragments,
+                    key=lambda item: item.placement_fragment_id,
+                )
+            ),
+        )
+        placement_ids = [item.placement_fragment_id for item in self.fragments]
+        if len(placement_ids) != len(set(placement_ids)):
+            raise ValueError("duplicate placement fragment in runtime binding")
+        fragment_ids = [item.fragment_id for item in self.fragments]
+        if len(fragment_ids) != len(set(fragment_ids)):
+            raise ValueError("duplicate runtime fragment_id in runtime binding")
+
+
+__all__ = ["WeightRuntimeBindingManifest"]
--- a/mooncake-reshard/python/mooncake/reshard/weight/runtime.py
+++ b/mooncake-reshard/python/mooncake/reshard/weight/runtime.py
@@ -1,0 +1,232 @@
+"""Runtime inventory adapters and ephemeral binding contracts."""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Any, Callable, Optional
+
+from ..contracts import (
+    ResourceKind,
+    RuntimeBindingManifest,
+)
+from .types import (
+    RuntimeBindingFragment,
+    _read_field,
+    _read_optional_field,
+    _read_weight_resource_id,
+    _require_integer,
+    _require_integer_tuple,
+    _require_manifest_items,
+    _require_nonempty_string,
+    _require_sequence,
+    _require_u64,
+)
+
+
+@dataclass(frozen=True)
+class WeightRuntimeBindingManifest(RuntimeBindingManifest):
+    """Ephemeral physical locations and lifetime fence for one placement."""
+
+    revision: str
+    participant_id: str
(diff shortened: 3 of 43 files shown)


zxpdemonio (commented):
## Review — `[Weight] Add runtime weight manifest contracts`

I reviewed the contract module, docs, wheel staging, and tests against RFC #3111. Locally: `139 passed` with `PYTHONPATH=mooncake-model-weight/python`.

### Verdict

Solid first-phase contract: placement / binding split is clear, validation is fail-closed in the right places, and the PR stays within the stated non-goals. **CI is currently red** (spell check), and there are a few contract/docs nits worth fixing before merge. No deep semantic blockers beyond that if the deferred planner checks are intentional.

### Blocker / merge gate

1. **Spell Check fails on `nd_*` identifiers**  
   Typos treats `nd` as `and` in `test_model_weight_manifest.py` (`nd_descriptor`, `nd_placement`, `nd_runtime`). CI Gate is red because of this.  
   Fix by renaming (`n_dim_*` / `multi_dim_*`) or adding an allowlist entry in `.typos.toml`.

### Contract review (what looks good)

- Clear split: `PlacementManifest` (serializable, address-free) vs `RuntimeBindingManifest` (ephemeral, generation/lease-fenced).
- Canonicalization of `partition_dim` / `shard_dims`, stable `placement_id` + digest, and bind-time digest checks.
- Geometry checks: bounds, non-shard axis integrity, nbytes = `prod(local_shape) * itemsize`.
- Same-rank N-D box overlap rejection (sweep + AABB) and runtime address-range overlap with explicit alias opt-in.
- Contiguity / canonical stride / `address_semantics="view"` rules are documented and tested.
- u64 address-range overflow checks; strict JSON schema for placement round-trip.
- Wheel staging + EXIT cleanup; `pkgutil.extend_path` enables `mooncake.model_weight` as a split package.
- Scope matches RFC phase-1 (no planner / TE / Store).

### Issues / questions

2. **README local test recipe is wrong** (`mooncake-model-weight/README.md`)  
   It suggests:
   ```bash
   PYTHONPATH=mooncake-wheel python -m pytest ...
   ```
   That fails with `No module named 'mooncake.model_weight'` unless the package has been staged into the wheel. The working recipes are:
   ```bash
   PYTHONPATH=mooncake-model-weight/python python -m pytest -q mooncake-model-weight/tests
   # or, with extend_path:
   PYTHONPATH=mooncake-wheel:mooncake-model-weight/python ...
   ```

3. **Cross-rank logical overlap is intentionally unchecked**  
   Same `tensor_id` with overlapping boxes on different `ParallelRank` values is accepted (verified). That is fine if coverage/conflict belongs to the planner, but please call it out explicitly in `docs/source/design/model-weight-manifest.md` so adapters do not assume the contract enforces a global partition.

4. **Incomplete shard coverage is also accepted**  
   A TP shard that only covers part of `global_shape` validates today. Same ask: document as planner-owned, so this is not mistaken for a full placement invariant.

5. **Strict `type(x) is int` will reject NumPy integers**  
   Framework inventories often carry `numpy.int64` extents. Torch `Size` is fine (Python `int`), but raw NumPy metadata will raise. Consider `operator.index()` / `numbers.Integral` (still rejecting `bool`), or document that adapters must coerce.

6. **`address` rejects `0`** (`minimum=1`)  
   Reasonable as a null sentinel for GPU views; worth one sentence in the design doc so CPU/testing harnesses do not treat it as accidental.

7. **Empty placement is allowed**  
   `PlacementManifest(model_id=..., revision=..., tensors=(), fragments=())` succeeds and gets a digest. Harmless, but if empty manifests should be invalid at the control-plane boundary, reject them here or document “empty is a valid sentinel”.

8. **Generated `placement_fragment_id` is content-hashed, but explicit IDs become part of `placement_id`**  
   Adapters that sometimes omit and sometimes supply IDs for the same geometry will see identity churn. Document “IDs must be stable across restarts” for framework exporters.

### Non-blocking nits

- `RuntimeManifest` can exist without `lease_id`; only projection to binding requires it — good, already documented.
- `build_wheel.sh` installs a second EXIT trap that replaces the first; safe today because the later trap also cleans staging, but a single composed cleanup would be clearer.
- No public schema version on serialized placement beyond the internal `"schema": "weight-placement"` used for `placement_id`. Fine for v0; consider surfacing it in JSON before Store persistence lands.
- PR checklist marks docs updated (good); spell-check + README fix should land with this PR.

### Test coverage

Strong for construction, projection, bind, aliases, overflow, and type strictness. Missing (optional follow-ups):

- README/`PYTHONPATH` smoke as documented
- explicit “cross-rank overlap allowed” / “incomplete coverage allowed” tests if those are permanent contract guarantees
- installed-wheel `import mooncake.model_weight` assertion in `test_installation.sh` (tests help, but a one-liner import check is cheaper signal)

### Summary

Approve directionally after fixing the typos CI failure and correcting the README import path. The remaining items are mostly “make deferred invariants explicit” so the next planner PR does not rediscover them.

Aionw:
`manifest.py` is already approximately 1,300 lines and covers several distinct responsibilities:

- public data types and fragments;
- placement serialization and identity generation;
- runtime inventory adapters;
- geometry, overlap, address, and alias validation;
- placement/runtime projection and binding.

The corresponding test file is also approximately 1,700 lines. Since future planner, Store, and framework-integration work will likely extend this module, it may be worth splitting it along these existing responsibility boundaries to keep the implementation easier to review and maintain.

Bo-Vincent:
> `manifest.py` is already approximately 1,300 lines and covers several distinct responsibilities:
> 
> * public data types and fragments;
> * placement serialization and identity generation;
> * runtime inventory adapters;
> * geometry, overlap, address, and alias validation;
> * placement/runtime projection and binding.
> 
> The corresponding test file is also approximately 1,700 lines. Since future planner, Store, and framework-integration work will likely extend this module, it may be worth splitting it along these existing responsibility boundaries to keep the implementation easier to review and maintain.

Thanks, this was a very helpful suggestion. I have split both the implementation and tests along the proposed responsibility boundaries, while keeping the public import surface unchanged.

stmatengss:
@Bo-Vincent We should consider the manifest of both weights and kvcache in current designs.

zxpdemonio:
> @Bo-Vincent We should consider the manifest of both weights and kvcache in current designs.

Agree

Bo-Vincent:
> @Bo-Vincent We should consider the manifest of both weights and kvcache in current designs.

Thanks. I have updated the PR so the manifest layer is resource-neutral rather than weight-only.

`mooncake.reuse.contracts` now defines the shared `ResourceManifest`, `PlacementManifest`, and `RuntimeBindingManifest` contracts. The common layer only owns resource identity, placement attestation, runtime instance, generation, and lease semantics.

Model-weight-specific tensor geometry, layout, fragments, and parallel ownership remain in `mooncake.reuse.weight` through `WeightPlacementManifest` and `WeightRuntimeBindingManifest`.

`ResourceKind.KV_CACHE` is reserved as the extension point for a future KVCache specialization. This PR does not implement KVCache placement or resharding yet, and it deliberately avoids forcing weight-specific tensor semantics into the future KVCache contract.

This gives model weights and KVCache a shared identity and lifecycle model while allowing each resource type to define its own placement and binding semantics.

Aionw:
Thanks for restructuring the package. I still have a concern about the new `mooncake-reuse` naming.

Based on the current contents, this module owns resource manifest contracts and resource-specific placement/binding manifests. It does not own reuse policies, lookup, deduplication, orchestration, or execution. Therefore, `reuse` seems to describe how these manifests may eventually be used rather than the concrete responsibility of the module, and the name may become an overly broad catch-all later.

Would it be clearer to name the module after the abstraction it actually owns, for example:

- `mooncake-manifest` / `mooncake.manifest`
- `mooncake.manifest.weight`
- `mooncake.manifest.kv_cache` when a concrete KVCache specialization is introduced

mooncake-reshard/python/mooncake/reshard/weight/types.py
```diff
@@ -0,0 +1,294 @@
...
+
+_MAX_U64 = (1 << 64) - 1
+
+
+@dataclass(frozen=True)
+class ParallelRank:
+    """Framework-provided owner coordinates used only for routing.
+
+    Logical sharding is defined by ``global_offset``, ``local_shape``, and
+    ``shard_dims``. This coordinate is not a replacement for the topology and
```
zxpdemonio, line 30: [P1] Please make the parallel topology part of the placement contract. There is currently no companion topology manifest in this PR or RFC implementation, so deferring it to planner inputs leaves serialized placement unable to describe its own layout. For example, ParallelRank(dp=0, ...) cannot distinguish an absent DP axis from DP size=2/rank=0, and max(rank)+1 is not recoverable because cross-rank coverage is intentionally incomplete. For N-D sharding, shard_dims also does not say which TP/EP axis owns each dimension. Therefore this cannot losslessly carry the existing TensorParallelism kind/rank/size/split_dim contract or validate/synthesize source and target layouts from the manifest. Please add placement-level topology/axis specs (at least kind, size, and split-dimension semantics, plus PP/EP ownership metadata as needed) and include them in placement identity/digest, instead of relying on undocumented external state.

mooncake-reshard/tests/weight_manifest/test_validation.py
```diff
@@ -0,0 +1,74 @@
...
+    ],
+)
+def test_physical_contract_rejects_values_outside_u64_abi(factory) -> None:
+    with pytest.raises(ValueError, match="64-bit"):
+        factory()
+
+
+def test_physical_contract_rejects_unrepresentable_exclusive_end() -> None:
+    with pytest.raises(ValueError, match="64-bit"):
+        binding_fragment(address=2**64 - 4, nbytes=4)
```
zxpdemonio, line 81: [P2] The core N-D geometry validator still lacks direct regression coverage. The current validation tests exercise scalar types and u64 address bounds, but not same-owner N-D overlap, adjacent boxes, logical bounds, splitting a non-shard dimension, nbytes mismatch, permitted cross-rank overlap, or intentionally incomplete coverage. These are the main invariants implemented by validation.py, especially the sweep-based overlap check, so please add focused cases that would fail if those branches regress.

mooncake-reshard/python/mooncake/reshard/weight/types.py
```diff
@@ -0,0 +1,294 @@
...
+        if self.partition_dim is None:
+            return ()
+        return (self.partition_dim,)
+
+
+@dataclass(frozen=True)
+class PlacementFragment:
+    """An address-free logical tensor box assigned to one parallel rank."""
+
+    placement_fragment_id: str
```
zxpdemonio: [P3] placement_fragment_id participates in placement_id and the binding fence, but every API requires each framework adapter to invent it and this module provides no canonical generator. The same logical fragment can therefore get a different placement identity after an adapter change or process restart even when tensor_id, rank, offset, and shape are unchanged, forcing unnecessary replanning and invalidating otherwise compatible bindings. Please provide one canonical helper or derive the ID from logical fragment content when omitted, while retaining explicit IDs only when the caller intentionally needs that identity.

mooncake-reshard/python/mooncake/reshard/weight/runtime.py
```diff
@@ -0,0 +1,149 @@
...
+        )
+
+
+def _runtime_binding_fragment_from_record(
+    record: Any,
+    owner_resolver: Optional[Callable[[Any], Any]],
+    address_semantics: Optional[str],
+    expected_generation: int,
+) -> RuntimeBindingFragment:
+    is_contiguous = _read_field(record, "is_contiguous")
```
zxpdemonio: [P1] A non-contiguous view can be accepted as contiguous here. The contract trusts the inventory-provided is_contiguous flag and drops stride/offset evidence before validate_runtime_binding; _canonical_stride is never used. I reproduced local_shape=(4,4) with stride=(1,4) and is_contiguous=True, and the binding still validates. That means the contract cannot prove that [address, address+nbytes) is the canonical contiguous view that the placement describes. Please preserve the necessary stride and normalized offset/address evidence in the binding and validate it against the placement local_shape before transfer, rather than treating the inventory boolean as proof.

mooncake-reshard/python/mooncake/reshard/weight/validation.py
```diff
@@ -0,0 +1,234 @@
...
+    left_binding: RuntimeBindingFragment,
+    right_placement: PlacementFragment,
+    right_binding: RuntimeBindingFragment,
+    tensors: Mapping[str, TensorDescriptor],
+) -> bool:
+    return (
+        left_binding.address == right_binding.address
+        and left_binding.nbytes == right_binding.nbytes
+        and len(left_placement.aliases) >= 2
+        and left_placement.aliases == right_placement.aliases
```
zxpdemonio, line 526: [P1] This exemption can authorize unrelated tensors to share an address. It only requires both fragments to carry the same non-empty aliases tuple; it never checks that either tensor_id belongs to that declared alias set. I reproduced a.weight and b.weight with aliases=(unrelated-x, unrelated-y) at the same address, and validate_runtime_binding accepts them. A malformed adapter can therefore hide an accidental overlap and silently bind two weights to the same bytes. Please validate both tensor IDs as members of the alias group, or model alias groups explicitly at manifest level and reference a validated group ID here.

Bo-Vincent:
> Thanks for restructuring the package. I still have a concern about the new `mooncake-reuse` naming.
> 
> Based on the current contents, this module owns resource manifest contracts and resource-specific placement/binding manifests. It does not own reuse policies, lookup, deduplication, orchestration, or execution. Therefore, `reuse` seems to describe how these manifests may eventually be used rather than the concrete responsibility of the module, and the name may become an overly broad catch-all later.
> 
> Would it be clearer to name the module after the abstraction it actually owns, for example:
> 
> * `mooncake-manifest` / `mooncake.manifest`
> * `mooncake.manifest.weight`
> * `mooncake.manifest.kv_cache` when a concrete KVCache specialization is introduced

Thanks, this concern makes sense. I renamed the stable module boundary to
`mooncake-reshard` / `mooncake.reshard`.

`manifest` describes this first PR, but it would become too narrow once the
backend-neutral N-D planner and lowering adapters land. The new namespace is
intended to remain stable across the phased series:

- `mooncake.reshard.contracts`
- `mooncake.reshard.weight`
- future `mooncake.reshard.kv_cache`, once concrete KV reshard semantics exist

This module does not own discovery, lookup, deduplication policy,
orchestration, activation, or the physical Transfer Engine and Store
implementations.

mooncake-reshard/python/mooncake/reshard/weight/types.py
Bo-Vincent: Thanks, agreed. The current head now makes the parallel topology part of the placement contract.

`ParallelTopology` records the TP/PP/EP/DP sizes together with the explicit participant-to-rank mapping. Each `TensorDescriptor` also carries `parallel_axes`, which maps the framework-provided axis kind to its logical split dimension; PP and DP remain ownership-only axes, while EP sharding is constrained to the leading expert dimension.

Both the topology and the per-tensor axis semantics are serialized and included in the canonical placement ID and digest. Global assembly also validates participant completeness, rank bounds, topology identity, and consistency between `parallel_axes` and `shard_dims`.

The corresponding topology, serialization, identity, and validation cases are covered by the current manifest test suite.

mooncake-reshard/tests/weight_manifest/test_validation.py
Bo-Vincent: Thanks. I added direct regression coverage for the N-D geometry invariants.

The tests now cover same-owner N-D overlap rejection, adjacent boxes, logical bounds, splitting a non-shard dimension, byte-size mismatch, permitted cross-rank overlap, and incomplete coverage semantics.

A local `WeightPlacementPart` may be incomplete because it is only one participant's contribution, while the assembled global `WeightPlacementManifest` rejects incomplete tensor coverage. The current manifest suite passes with 162 tests.

mooncake-reshard/python/mooncake/reshard/weight/types.py
Bo-Vincent: Thanks. `placement_fragment_id` is now optional and is generated canonically when omitted.

The generated ID is a SHA-256 digest over the logical fragment content: tensor ID, global offset, local shape, byte size, parallel rank, and normalized alias set. This keeps the identity stable across process restarts and inventory ordering changes.

Explicit IDs remain supported for adapters that intentionally own a stable external fragment identity.

mooncake-reshard/python/mooncake/reshard/weight/runtime.py
Bo-Vincent: Thanks for the concrete reproduction. The binding path no longer treats `is_contiguous=True` as sufficient evidence.

`RuntimeBindingFragment` now preserves item size, local shape, byte strides, storage base address, storage size, and normalized storage offset. Inventory conversion requires stride evidence and cross-checks element and byte representations when both are present.

Binding validation then verifies the exact logical shape, item size, canonical contiguous strides, `address == storage_address + storage_offset_bytes`, and allocation bounds. A regression test reproduces `stride=(1, 4)` with `is_contiguous=True` and confirms that it is rejected.

mooncake-reshard/python/mooncake/reshard/weight/validation.py
Bo-Vincent: Thanks. Alias authorization is now validated against the actual tensor identities.

A declared alias group must contain at least two tensor IDs, and every fragment carrying the group must contain its own `tensor_id` in that group. The runtime overlap exemption additionally requires the same normalized alias group, exact physical range identity, and compatible tensor descriptors.

The reported case using `aliases=("unrelated-x", "unrelated-y")` for `a.weight` and `b.weight` is now rejected and covered by a regression test.

mooncake-reshard/python/mooncake/reshard/weight/validation.py
```diff
@@ -0,0 +1,347 @@
...
+                    f"{tensor.tensor_id}: dp={dp_rank}"
+                )
+
+
+def _fragments_exactly_cover_tensor(
+    tensor: TensorDescriptor,
+    fragments: Sequence[PlacementFragment],
+) -> bool:
+    boxes = tuple(
+        dict.fromkeys(
```
zxpdemonio: [P1] The topology is now serialized, but its semantics are not yet validated against the fragment layout. This deduplicates identical boxes across ranks before coverage, while _parallel_tensor_owner only separates PP and ownership-only EP. Consequently, a tensor declared as TP split_dim=0 can still be accepted when tp=0 and tp=1 both hold the complete tensor (the current cross-rank-overlap test explicitly permits this), or when only one TP rank owns the full tensor. The manifest is self-describing syntactically but can contain a layout that contradicts parallel_axes, leaving the planner unable to distinguish sharding from replication. Please derive coverage groups from the axis semantics: ownership/replica axes should form independent covers, while fragments across split axes should form one non-overlapping complete cover, with required split-rank participation validated.

mooncake-reshard/python/mooncake/reshard/weight/binding.py
```diff
@@ -0,0 +1,136 @@
...
+        if runtime_fragment.itemsize != tensor.itemsize:
+            raise ValueError(
+                "runtime binding itemsize does not match placement: "
+                f"{placement_fragment.placement_fragment_id}"
+            )
+        expected_strides_bytes = tuple(
+            stride * tensor.itemsize
+            for stride in _canonical_stride(placement_fragment.local_shape)
+        )
+        if runtime_fragment.strides_bytes != expected_strides_bytes:
```
zxpdemonio, line 74: [P2, non-blocking] Exact equality with the canonical stride rejects valid contiguous framework views when a dimension has extent 1. For example, PyTorch reports as_strided(base, (1, 4), (100, 1)).is_contiguous() as true because the first stride is irrelevant, but this check rejects it. That may force adapters to materialize an unnecessary contiguous copy for singleton-shaped weights or scales. Please validate row-major contiguity only on dimensions whose extent is greater than 1 and normalize singleton-dimension strides. This can be a follow-up because it fails closed rather than transferring incorrect data.

zxpdemonio (approved):
LGTM.

Aionw:
Make framework adaptation explicit and keep the reshard core canonical and typed

The current implementation is described as framework-agnostic, but the public boundary achieves this through `Any`, reflective `_read_field()` access, optional compatibility fields, and runtime duck typing.

I do not think this eliminates framework adaptation. Each inference framework still has to translate its tensor identity, parallel ranks, sharding semantics, stride units, storage offsets, and allocation lifetime into the implicit field schema expected by these methods. The adaptation is already present; it is merely hidden, unchecked, and partially implemented inside the shared core.

I think a small explicit adapter per inference framework is an acceptable and worthwhile cost:

```text
SGLang / vLLM / PyTorch objects
              ↓
     framework-owned adapter
              ↓
canonical WeightPlacementPart /
WeightRuntimeBindingManifest
              ↓
       Mooncake reshard core
```

Core remains framework-independent because it neither imports nor inspects framework objects. The adapters are intentionally framework-specific because they are the only layer that can correctly interpret framework semantics.

#### Why this separation is valuable

1. **Stable placement identity across frameworks and process restarts**

   Placement IDs and digests are computed from manifest contents. Allowing multiple equivalent representations such as `partition_dim/shard_dims`, `model_id/resource_id`, or element/byte offsets makes canonical identity dependent on how an adapter happened to express the same placement.

   Normalizing once in the framework adapter ensures that logically identical placements produce the same canonical content and therefore the same placement identity. This matters for plan reuse, cache lookup, binding attestation, and recovery after restart.

2. **Earlier detection of unsafe memory-layout errors**

   Confusing element strides with byte strides, or storage offsets with byte offsets, can produce an incorrect physical address or transfer range. A typed adapter with one byte-based canonical representation makes unit conversion explicit and testable at the framework boundary.

3. **Planner semantics become trustworthy**

   The planner needs to distinguish split, replication, and ownership. `split_dim: Optional[int]` does not express those states and currently allows axis metadata that contradicts fragment geometry.

   Explicit `SplitAxis`, `ReplicatedAxis`, and `OwnershipAxis` variants give each state a concrete validation rule. The planner can then consume a validated placement instead of reinterpreting ambiguous metadata.

4. **Framework API changes remain local**

   PyTorch, SGLang, and vLLM do not expose identical tensor, topology, or allocation APIs, and those APIs will evolve independently. With adapters, a framework upgrade changes only its adapter and conformance tests; the manifest model and planner remain unchanged.

5. **Adding a framework becomes a bounded task**

   A new framework should implement two explicit operations:

   ```python
   def export_placement_part(...) -> WeightPlacementPart: ...
   def export_runtime_binding(...) -> WeightRuntimeBindingManifest: ...
   ```

   Once those outputs pass the shared contract tests, the framework can use the same planner and executor without adding more compatibility branches to the core.

6. **The shared implementation becomes smaller**

   Removing speculative input compatibility eliminates reflective readers, duplicated normalization, conflict checks between aliases, and many mapping-versus-attribute tests. Tests can focus on actual reshard invariants: coverage, topology, aliasing, identity, and address safety.

#### Concrete changes requested

1. Remove generic inventory parsing from the core:

   - `WeightPlacementPart.from_runtime_inventory(inventory: Any)`
   - `WeightRuntimeBindingManifest.from_runtime_inventory(inventory: Any)`
   - `_read_field()`
   - `_read_optional_field()`
   - `_read_weight_resource_id()`
   - related mapping/attribute duck-typing helpers

   Core constructors should accept only canonical typed values. Concrete adapters can be introduced with their corresponding framework integration PRs.

2. Keep only one canonical representation:

   - `resource_id`, not `resource_id/model_id`
   - `shard_dims`, not `partition_dim/shard_dims`
   - `strides_bytes`, not `stride/strides_bytes`
   - `storage_offset_bytes`, not `storage_offset/byte_offset`

3. Replace `TensorParallelAxis(kind, split_dim: Optional[int])` with explicit variants:

   ```python
   SplitAxis(kind, dim)
   ReplicatedAxis(kind)
   OwnershipAxis(kind)
   ```

   Validation should then enforce:

   - split-axis ranks collectively partition the declared dimension;
   - replicated-axis ranks each provide an independent complete cover;
   - ownership-axis ranks provide independent covers only for declared owners.

4. Use structural interfaces such as `Protocol` for resource-neutral behavior instead of inherited dataclasses. Identity validation only reads manifest properties, so dataclass inheritance is unnecessary and currently causes the custom constructor and repeated `object.__setattr__` initialization.

5. If mapping or JSON inventory input is required, handle it in a separate `serde` or `wire` decoder:

   ```text
   framework object ── typed adapter ──┐
                                       ├── canonical domain model
   JSON/RPC payload ── schema decoder ─┘
   ```

#### Suggested acceptance criteria

- no `Any` or reflective framework-field access in the public core contract;
- each domain value has one canonical representation and unit;
- framework-specific normalization lives in explicit adapters;
- split, replication, and ownership are distinct types;
- the split-axis geometry counterexample is rejected;
- strict type checking covers the domain and adapter interfaces;
- adding another framework does not require changes to the core manifest parser.

In short:

> Core should be framework-independent; adapters should be explicitly framework-specific. The small per-framework adapter cost buys stable placement identity, trustworthy planner semantics, earlier detection of memory-layout errors, and isolation from framework API changes.

Aionw:
Thanks for the latest update. The framework/core boundary is substantially better now: the reflective inventory adapters are gone, wire decoding is isolated in `serde.py`, the core has one canonical representation for each value, and the split-axis counterexample from the previous review is now rejected.

I found two remaining follow-ups before I would consider the type-safety work complete:

### 1. Static type safety is not enforced yet

The new domain model is better typed at runtime, but there is still no mypy/pyright configuration or CI check, and several annotations erase the distinctions we want the type system to enforce:

- `SplitAxis.kind`, `ReplicatedAxis.kind`, and `OwnershipAxis.kind` are still plain `str`, rather than narrowed `Literal` types;
- resource, placement, participant, placement-fragment, and runtime-fragment IDs are all plain `str`, so exchanging them remains statically valid;
- `_axis_from_wire() -> Any`, `_axis_to_wire(axis: Any)`, and `_require_manifest_items() -> tuple[Any, ...]` discard type information after decoding/validation;
- `WeightPlacementManifest.placement_id` is annotated as `str`, but its custom constructor temporarily assigns `Optional[str]` through `object.__setattr__`, bypassing the declared type.

Please add a strict type-checking job for this module and tighten these boundaries, for example with `Literal` axis kinds, typed IDs (`NewType` or small value types), a generic `_require_manifest_items[T]`, and `ParallelAxis` as the serde return/input type. The acceptance test should be that invalid axis kinds and exchanged ID categories fail static checking, rather than only being discovered by a later runtime validator.

The `Protocol` conversion removed the need for inherited dataclass fields, so it would also be good to finish that change by replacing the `init=False`/temporary-invalid-state constructor with a normal construction path or a factory that builds a fully valid immutable value.

### 2. Alias groups are not validated against the global tensor set

The current head still accepts this placement and binding:

```python
tensors = ("a", "b")
aliases = ("a", "b", "ghost")  # "ghost" is not present in the manifest
# a and b are then bound to the same runtime address
```

Both fragments contain their own tensor IDs, so the pairwise runtime overlap check accepts the alias. However, the alias declaration is not self-contained: it references a tensor that does not exist in the placement.

Please validate alias groups during global placement assembly, before runtime binding:

```python
set(fragment.aliases) <= set(tensor_by_id)
```

and ensure that every fragment participating in the same alias group declares a consistent group. This keeps placement identity and alias authorization based only on tensors actually present in the manifest, instead of allowing partially malformed adapter output to survive until binding time.

Bo-Vincent:
> Thanks for the latest update. The framework/core boundary is substantially better now: the reflective inventory adapters are gone, wire decoding is isolated in `serde.py`, the core has one canonical representation for each value, and the split-axis counterexample from the previous review is now rejected.
> 
> I found two remaining follow-ups before I would consider the type-safety work complete:
> 
> ### 1. Static type safety is not enforced yet
> The new domain model is better typed at runtime, but there is still no mypy/pyright configuration or CI check, and several annotations erase the distinctions we want the type system to enforce:
> 
> * `SplitAxis.kind`, `ReplicatedAxis.kind`, and `OwnershipAxis.kind` are still plain `str`, rather than narrowed `Literal` types;
> * resource, placement, participant, placement-fragment, and runtime-fragment IDs are all plain `str`, so exchanging them remains statically valid;
> * `_axis_from_wire() -> Any`, `_axis_to_wire(axis: Any)`, and `_require_manifest_items() -> tuple[Any, ...]` discard type information after decoding/validation;
> * `WeightPlacementManifest.placement_id` is annotated as `str`, but its custom constructor temporarily assigns `Optional[str]` through `object.__setattr__`, bypassing the declared type.
> 
> Please add a strict type-checking job for this module and tighten these boundaries, for example with `Literal` axis kinds, typed IDs (`NewType` or small value types), a generic `_require_manifest_items[T]`, and `ParallelAxis` as the serde return/input type. The acceptance test should be that invalid axis kinds and exchanged ID categories fail static checking, rather than only being discovered by a later runtime validator.
> 
> The `Protocol` conversion removed the need for inherited dataclass fields, so it would also be good to finish that change by replacing the `init=False`/temporary-invalid-state constructor with a normal construction path or a factory that builds a fully valid immutable value.
> 
> ### 2. Alias groups are not validated against the global tensor set
> The current head still accepts this placement and binding:
> 
> ```python
> tensors = ("a", "b")
> aliases = ("a", "b", "ghost")  # "ghost" is not present in the manifest
> # a and b are then bound to the same runtime address
> ```
> 
> Both fragments contain their own tensor IDs, so the pairwise runtime overlap check accepts the alias. However, the alias declaration is not self-contained: it references a tensor that does not exist in the placement.
> 
> Please validate alias groups during global placement assembly, before runtime binding:
> 
> ```python
> set(fragment.aliases) <= set(tensor_by_id)
> ```
> 
> and ensure that every fragment participating in the same alias group declares a consistent group. This keeps placement identity and alias authorization based only on tensors actually present in the manifest, instead of allowing partially malformed adapter output to survive until binding time.

Thanks for the detailed review. Both requested follow-ups are addressed in the current head.

### Static type safety

- Parallel axis kinds now use `Literal` types.
- Resource, placement, participant, tensor, fragment, runtime-instance, lease, revision, and placement-set IDs are distinct `NewType` categories.
- Serde preserves the typed domain model instead of returning `Any`.
- The canonical core no longer accepts reflective framework inventory parsing or duck-typed runtime objects.
- A strict Pyright check and negative type fixtures were added. Invalid axis kinds and exchanged ID categories are now rejected statically.

### Alias validation

Alias groups are now validated during complete placement assembly.

The placement rejects unknown alias members, fragments missing their own tensor ID from an alias group, inconsistent alias-group declarations, and shared runtime ranges unless the tensors are in the same validated compatible alias group.

Thanks again for the suggestions. The framework adapter boundary is now explicit: adapters normalize framework runtime state into typed manifests, while the Mooncake reshard core only consumes the canonical contracts.

Aionw (approved):
LGTM

.github/CODEOWNERS
```diff
@@ -12,6 +12,7 @@
 /mooncake-integration/transfer_engine @ShangmingCai @alogfans
 /mooncake-integration/store @ykwd @stmatengss @zxpdemonio
 /mooncake-pg @UNIDY2002 @ympcMark @yuechen-sys
+/mooncake-reshard @ShangmingCai @stmatengss
```
stmatengss:
```suggestion
/mooncake-reshard @ShangmingCai @stmatengss @Bo-Vincent @zxpdemonio 
```

stmatengss (approved):

PR #3189, opened 2026-07-30
[Bugfix] Check free HugeTLB capacity

cupkk:
## Description

The HiCache HugeTLB sizing helper currently reads `nr_hugepages`, which is the configured pool size, when auto-detecting available capacity. If another process already holds HugeTLB pages, the preflight can overestimate capacity and report that a launch fits even though the required pages are not free.

Read `free_hugepages` instead, and add a regression test with different configured and free page counts. The sysfs root is kept as an internal constant so the behavior can be tested without touching host sysfs.

## Module

- [ ] CI/CD
- [ ] Docs
- [x] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
python -m unittest scripts/test_hicache_hugepage_requirements.py
python -m compileall -q scripts/check_hicache_hugepage_requirements.py scripts/test_hicache_hugepage_requirements.py
git diff --check
```

**Test results:**
- [x] Unit tests pass (7 focused tests)
- [ ] Integration tests pass (not run; no HugeTLB/CUDA/RDMA environment required for this helper change)
- [x] Manual testing done: reproduced that the previous implementation selected `nr_hugepages`; the regression fixture verifies `7` free pages are returned when `11` pages are configured.

## Checklist

- [ ] I have updated the documentation (not applicable; CLI behavior and options are unchanged)
- [ ] For changes >500 LOC: I have filed an RFC issue (not applicable; 25 insertions and 3 deletions)

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex assisted with repository inspection, the focused implementation, and test drafting. I reviewed every changed line and verified the behavior and remote diff.

### DIFF
--- a/scripts/check_hicache_hugepage_requirements.py
+++ b/scripts/check_hicache_hugepage_requirements.py
@@ -26,6 +26,8 @@ _UNIT_MAP = {
     "tib": 1024**4,
 }
 
+_HUGEPAGE_SYSFS_ROOT = Path("/sys/kernel/mm/hugepages")
+
 
 @dataclass(frozen=True)
 class BudgetSummary:
@@ -66,12 +68,16 @@ def read_available_hugetlb(page_size_bytes: int) -> int | None:
         return None
 
     page_size_kib = page_size_bytes // 1024
-    path = Path(f"/sys/kernel/mm/hugepages/hugepages-{page_size_kib}kB/nr_hugepages")
+    path = (
+        _HUGEPAGE_SYSFS_ROOT
+        / f"hugepages-{page_size_kib}kB"
+        / "free_hugepages"
+    )
     try:
-        nr_hugepages = int(path.read_text(encoding="utf-8").strip())
+        free_hugepages = int(path.read_text(encoding="utf-8").strip())
     except (FileNotFoundError, PermissionError, ValueError, OSError):
         return None
-    return nr_hugepages * page_size_bytes
+    return free_hugepages * page_size_bytes
 
 
 def evaluate_budget(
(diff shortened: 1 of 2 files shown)


scripts/check_hicache_hugepage_requirements.py
```diff
@@ -66,12 +68,16 @@ def read_available_hugetlb(page_size_bytes: int) -> int | None:
...
+        / f"hugepages-{page_size_kib}kB"
+        / "free_hugepages"
+    )
     try:
-        nr_hugepages = int(path.read_text(encoding="utf-8").strip())
+        free_hugepages = int(path.read_text(encoding="utf-8").strip())
     except (FileNotFoundError, PermissionError, ValueError, OSError):
         return None
-    return nr_hugepages * page_size_bytes
+    return free_hugepages * page_size_bytes
```
Aionw, line 80: **[P2] Exclude HugeTLB pages already promised to reservations**

`free_hugepages` includes pages that have been reserved by existing HugeTLB mappings but have not been faulted in yet. Those pages are not generally available to this new launch; the kernel describes the generally available count as `free_huge_pages - resv_huge_pages` ([reservation docs](https://docs.kernel.org/mm/hugetlbfs_reserv.html)). For example, another process can reserve most of the pool with `mmap(MAP_HUGETLB)` while leaving `free_hugepages` unchanged, so this helper still reports that the launch fits and it can later fail as those reservations are faulted. Please read the sibling `resv_hugepages` value, subtract it (clamped at zero), and add a regression case with a nonzero reservation count.

Aionw (changes requested):
I found one correctness issue in the capacity calculation. The focused unit tests and `compileall` pass. I also ran pre-commit on the two changed files; `ruff-format` reformatted both, so please commit those formatting changes after addressing the functional issue.

PR #3193, opened 2026-07-30
[Bugfix][Store] Support JsonCpp header layouts across Linux distributions

Icedcoco:
## Description

Fix the Mooncake Store build failure seen in the [x86 nightly manylinux job](https://github.com/kvcache-ai/Mooncake/actions/runs/30471153615/job/90644300155).

The build installs `jsoncpp-devel` and CMake successfully finds `JsonCpp::JsonCpp`, but AlmaLinux exposes the header as `<json/json.h>`. Several recently added OpLog files hard-coded the Ubuntu layout, `<jsoncpp/json/json.h>`, causing compilation to fail.

Use the repository's existing `__has_include` compatibility pattern to support both header layouts in the OpLog inspector, auditor, benchmark, and auditor test.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**

```bash
cmake --build build  --target oplog_batch_inspector oplog_batch_bench oplog_batch_auditor_test  -j32

env LSAN_OPTIONS=detect_leaks=0  ./build/mooncake-store/tests/oplog_batch_auditor_test

env LSAN_OPTIONS=detect_leaks=0  ./build/mooncake-store/tools/oplog_batch_inspector --version

env LSAN_OPTIONS=detect_leaks=0  ./build/mooncake-store/benchmarks/oplog_batch_bench --version

pre-commit run --files \
  mooncake-store/benchmarks/oplog_batch_bench.cpp \
  mooncake-store/tests/ha/oplog/oplog_batch_auditor_test.cpp \
  mooncake-store/tools/oplog_batch_auditor.cpp \
  mooncake-store/tools/oplog_batch_inspector.cpp
```

**Test results:**

- [x] Unit tests pass: 8/8 `OpLogBatchAuditorTest` cases
- [x] Manual testing done: all three affected targets build and both binaries load

## Checklist

- [ ] I have updated the documentation

## AI Assistance Disclosure

- [x] No AI tools were used
- [ ] AI tools were used

### DIFF
--- a/mooncake-store/benchmarks/oplog_batch_bench.cpp
+++ b/mooncake-store/benchmarks/oplog_batch_bench.cpp
@@ -1,6 +1,10 @@
 #include <gflags/gflags.h>
 #include <glog/logging.h>
-#include <jsoncpp/json/json.h>
+#if __has_include(<jsoncpp/json/json.h>)
+#include <jsoncpp/json/json.h>  // Ubuntu
+#else
+#include <json/json.h>  // CentOS
+#endif
 
 #include <algorithm>
 #include <atomic>
--- a/mooncake-store/tests/ha/oplog/oplog_batch_auditor_test.cpp
+++ b/mooncake-store/tests/ha/oplog/oplog_batch_auditor_test.cpp
@@ -1,6 +1,10 @@
 #include <gtest/gtest.h>
 
-#include <jsoncpp/json/json.h>
+#if __has_include(<jsoncpp/json/json.h>)
+#include <jsoncpp/json/json.h>  // Ubuntu
+#else
+#include <json/json.h>  // CentOS
+#endif
 
 #include <algorithm>
 #include <sstream>
--- a/mooncake-store/tools/oplog_batch_auditor.cpp
+++ b/mooncake-store/tools/oplog_batch_auditor.cpp
@@ -6,7 +6,11 @@
 #include <map>
 #include <string_view>
 
-#include <jsoncpp/json/json.h>
+#if __has_include(<jsoncpp/json/json.h>)
+#include <jsoncpp/json/json.h>  // Ubuntu
+#else
+#include <json/json.h>  // CentOS
+#endif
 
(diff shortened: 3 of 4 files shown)


Aionw:
Instead of relying on prebuilt binaries, why don't we fetch the source and compile it during the build process? This would:
1. Make the build more portable across different OS distributions, without requiring us to maintain separate compatibility handling for each environment.
2. Avoid version-specific adaptation work caused by differences between older and newer releases.
3. By releasing static linked target, end-user do not need to install those dependencies.

Icedcoco:
> Instead of relying on prebuilt binaries, why don't we fetch the source and compile it during the build process? This would:
> 
> 1. Make the build more portable across different OS distributions, without requiring us to maintain separate compatibility handling for each environment.
> 2. Avoid version-specific adaptation work caused by differences between older and newer releases.
> 3. By releasing static linked target, end-user do not need to install those dependencies.

Building a pinned JsonCpp version from source is a reasonable follow-up. It would give us a consistent header layout and version across distributions, and static linkage could remove the native runtime dependency on `libjsoncpp.so`.

I would keep this PR scoped to restoring the existing builds. Migrating JsonCpp affects more than these four Store files: `FindJsonCpp.cmake` currently discovers the system package, `mooncake_common` links `jsoncpp` directly, and the Go and Rust bindings explicitly link `-ljsoncpp`. We also need to account for standalone Transfer Engine builds and native/wheel packaging.

I suggest handling that migration separately: introduce a pinned source target, decide whether to retain a system-package option for offline and downstream builds, update all consumers, and then remove the compatibility includes after validating the top-level build, standalone Transfer Engine, bindings, and wheel builds.

Aionw (approved):

ykwd (approved):

00fish0 (approved):

PR #3197, opened 2026-07-30
[Store] Avoid staging for same-process GPU reads

zxpdemonio:
## Description

Avoid the extra CPU staging buffer for same-process GPU read destinations in the Store get path.

Before this change, `get_into(cuda_ptr)` always allocated a client-side CPU staging buffer when the destination pointer was a CUDA pointer, then copied the result from CPU staging to GPU. This is necessary for remote transfers, but it is unnecessary for same-process local memcpy: the transfer layer can copy directly from the local Store segment into the caller's CUDA destination.

This PR exposes the existing transfer-layer local memcpy decision through `Client::CanUseLocalMemcpy()` and uses it in `RealClient::execute_ranged_read()` to skip staging only when all of the following are true:

- the selected replica is a memory replica
- `MC_STORE_MEMCPY` allows local memcpy
- the replica endpoint is the same process endpoint
- the destination is a CUDA/device pointer
- the local hot cache is disabled

It also routes dummy-client CUDA IPC tensor reads through the same ranged get path instead of the scatter path. The dummy client still exports the destination CUDA tensor with CUDA IPC, and the real client opens the IPC handle; after that, the read uses `get_into_range_internal()`, so same-process memory replicas can use the local memcpy no-staging path and remote/disabled-local-memcpy cases keep the conservative staging behavior. This avoids requiring the CUDA IPC destination pointer to be registered with TransferEngine for the same-node local memcpy case.

For `get_into_ranges()`, same-process GPU memory reads now use the same ranged get path as well. CPU range reads and remote range reads keep the existing scatter path.

Remote transfers, disk replicas, non-memory replicas, hot-cache-enabled full-object reads, and `MC_STORE_MEMCPY=0` keep the existing staging behavior.

Base branch: latest `origin/main` (`a6b4db4cfa371a3fc7f189a19977c102e14c4dfc`)

Branch: `cruz/get_no_staging`

Related RFC: [#3229 [RFC] RL Training Weight Offload and Restore Paths for Mooncake Store](https://github.com/kvcache-ai/Mooncake/issues/3229)

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [x] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**

```bash
cmake -S . -B build-get-no-staging-nocuda -DBUILD_UNIT_TESTS=OFF -DBUILD_EXAMPLES=OFF -DBUILD_BENCHMARK=OFF -DUSE_CUDA=OFF
cmake --build build-get-no-staging-nocuda -j8 --target store

cmake -S . -B build-get-no-staging-cuda -DBUILD_UNIT_TESTS=OFF -DBUILD_EXAMPLES=OFF -DBUILD_BENCHMARK=OFF -DUSE_CUDA=ON
cmake --build build-get-no-staging-cuda -j8 --target store mooncake_master

cmake -S . -B build-get-no-staging-tests -DBUILD_UNIT_TESTS=ON -DBUILD_EXAMPLES=OFF -DBUILD_BENCHMARK=OFF -DUSE_CUDA=OFF
cmake --build build-get-no-staging-tests -j8 --target transfer_task_test
./build-get-no-staging-tests/mooncake-store/tests/transfer_task_test --gtest_filter=TransferTaskTest.IsSameProcessEndpoint:TransferTaskTest.CanUseLocalMemcpyRequiresSameProcessEndpoint:TransferTaskTest.CanUseLocalMemcpyHonorsMemcpyEnv

./scripts/code_format.sh --check -b origin/main
git diff --check
```

**Test results:**

- [x] Unit tests pass

Same-node `get_into(cuda_ptr)` benchmark, Store segment pinned memory enabled, `MC_STORE_MEMCPY=1`, `protocol=tcp`, `CUDA_VISIBLE_DEVICES=0`.

| Size | main GB/s | PR GB/s | Speedup | main median | PR median |
| ---: | ---: | ---: | ---: | ---: | ---: |
| 64 MB | 5.784 | 23.558 | 4.07x | 11.603 ms | 2.849 ms |
| 256 MB | 5.659 | 24.053 | 4.25x | 47.434 ms | 11.160 ms |
| 512 MB | 5.709 | 24.122 | 4.23x | 94.044 ms | 22.256 ms |

Additional path validation:

- `MC_STORE_MEMCPY=1 + get_into(cuda_ptr)`: passed, direct same-process local memcpy path
- `MC_STORE_MEMCPY=0 + get_into(cuda_ptr)`: passed, existing staging path preserved
- `MC_STORE_MEMCPY=1 + get_into(cpu_ptr)`: passed, CPU destination path preserved
- dummy-client CUDA IPC read helper: build and code-path validation confirm it now reuses `get_into_range_internal()` instead of `SubmitScatter`, so same-node local memcpy does not require registering the IPC-opened CUDA destination pointer with TransferEngine
- `get_into_ranges()` memory reads: code-path validation confirms same-process GPU destination buffers now reuse ranged get/local memcpy; CPU and remote range reads keep the existing scatter path

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

AI assistance was used to inspect the existing Store/transfer paths, prepare the small code change, run validation, and draft this PR description. I am responsible for reviewing and defending the final change.

### DIFF
--- a/mooncake-store/src/real_client.cpp
+++ b/mooncake-store/src/real_client.cpp
@@ -3639,15 +3639,48 @@ RealClient::get_into_ranges_internal(
             if (metadata.replica.is_memory_replica()) {
                 if (client_->CanUseLocalMemcpy(metadata.replica) &&
                     runtime_accelerator.FindDeviceForPointer(buffers[i])) {
+                    // Planning cache entries may be close to expiry. Renew and
+                    // reselect the replica before copying into device memory.
+                    auto refresh_result = resolve_ranged_read_metadata(keys[j]);
+                    if (!refresh_result) {
+                        std::fill(range_results.begin(), range_results.end(),
+                                  tl::unexpected(refresh_result.error()));
+                        continue;
+                    }
+                    std::optional<RangedReadMetadata> refreshed_metadata;
+                    refreshed_metadata.emplace(std::move(*refresh_result));
+                    auto lease_refresh_at =
+                        [](const RangedReadMetadata &value) {
+                            const auto now = std::chrono::steady_clock::now();
+                            return now +
+                                   (value.query_result.lease_timeout - now) / 2;
+                        };
+                    auto refresh_at = lease_refresh_at(*refreshed_metadata);
                     for (size_t k = 0; k < range_results.size(); ++k) {
+                        // Renew halfway through the remaining lease in long
+                        // batches.
                         const size_t dst_offset = dst_offsets[k];
                         if (dst_offset > capacities[i] ||
                             sizes[k] > capacities[i] - dst_offset) {
                             continue;
                         }
+                        if (std::chrono::steady_clock::now() >= refresh_at) {
+                            auto next_refresh_result =
+                                resolve_ranged_read_metadata(keys[j]);
+                            if (!next_refresh_result) {
+                                std::fill(range_results.begin() + k,
+                                          range_results.end(),
+                                          tl::unexpected(
+                                              next_refresh_result.error()));
+                                break;
+                            }
+                            refreshed_metadata.emplace(
+                                std::move(*next_refresh_result));
+                            refresh_at = lease_refresh_at(*refreshed_metadata);
+                        }
                         range_results[k] = execute_ranged_read(
                             keys[j], buffers[i], dst_offset, src_offsets[k],
-                            sizes[k], metadata, false, false);
+                            sizes[k], *refreshed_metadata, false, false);
                     }
                     continue;
                 }
--- a/mooncake-store/src/real_client.cpp
+++ b/mooncake-store/src/real_client.cpp
@@ -3595,6 +3595,8 @@ RealClient::get_into_ranges_internal(
             .first->second;
     };
 
+    auto runtime_accelerator =
+        device::GetAcceleratorRegistry().RuntimeAccelerators();
     struct ScatterLease {
         std::chrono::steady_clock::time_point expires_at;
         std::optional<ErrorCode> error;
@@ -3633,6 +3635,20 @@ RealClient::get_into_ranges_internal(
 
             const auto &metadata = metadata_result.value();
             if (metadata.replica.is_memory_replica()) {
+                if (client_->CanUseLocalMemcpy(metadata.replica) &&
+                    runtime_accelerator.FindDeviceForPointer(buffers[i])) {
+                    for (size_t k = 0; k < range_results.size(); ++k) {
+                        const size_t dst_offset = dst_offsets[k];
+                        if (dst_offset > capacities[i] ||
+                            sizes[k] > capacities[i] - dst_offset) {
+                            continue;
+                        }
+                        range_results[k] = execute_ranged_read(
+                            keys[j], buffers[i], dst_offset, src_offsets[k],
+                            sizes[k], metadata, false, false);
+                    }
+                    continue;
+                }
                 const auto &handle =
                     metadata.replica.get_memory_descriptor().buffer_descriptor;
                 auto [lease_it, inserted] = scatter_leases.try_emplace(keys[j]);
--- a/mooncake-store/src/real_client.cpp
+++ b/mooncake-store/src/real_client.cpp
@@ -4584,58 +4584,22 @@ RealClient::batch_get_into_cuda_ipc_dummy_helper(
 
     std::vector<tl::expected<int64_t, ErrorCode>> results(
         requests.size(), tl::unexpected(ErrorCode::INVALID_PARAMS));
-    std::vector<device::CudaIpcBufferMapping> mappings;
-    std::vector<void *> buffers;
-    std::vector<std::vector<std::string>> all_keys;
-    std::vector<std::vector<std::vector<size_t>>> all_dst_offsets;
-    std::vector<std::vector<std::vector<size_t>>> all_src_offsets;
-    std::vector<std::vector<std::vector<size_t>>> all_sizes;
-    std::vector<size_t> buffer_capacities;
-    std::vector<size_t> original_indices;
-    mappings.reserve(requests.size());
-    buffers.reserve(requests.size());
-    all_keys.reserve(requests.size());
-    all_dst_offsets.reserve(requests.size());
-    all_src_offsets.reserve(requests.size());
-    all_sizes.reserve(requests.size());
-    buffer_capacities.reserve(requests.size());
-    original_indices.reserve(requests.size());
 
     for (size_t i = 0; i < requests.size(); ++i) {
         const auto &request = requests[i];
+        if (request.size == 0) {
+            results[i] = 0;
+            continue;
+        }
         auto mapping = device::CudaIpcBufferMapping::Open(request.destination);
         if (!mapping) {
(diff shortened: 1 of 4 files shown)


mooncake-store/src/real_client.cpp
```diff
@@ -3354,7 +3354,8 @@ tl::expected<int64_t, ErrorCode> RealClient::execute_ranged_read(
         auto runtime_accelerator =
             device::GetAcceleratorRegistry().RuntimeAccelerators();
         void *dst = static_cast<char *>(buffer) + dst_offset;
-        if (runtime_accelerator.FindDeviceForPointer(dst)) {
+        if (runtime_accelerator.FindDeviceForPointer(dst) &&
```
yokinoshitayoki, line 3357: Skipping staging here passes the GPU destination slices directly to Client::Get. When the shared-memory hot cache is enabled and the key reaches the admission threshold, ProcessSlicesAsync forwards those slices to LocalHotCacheHandler::SubmitPutTask, which copies slice.ptr using plain std::memcpy. Since slice.ptr is a device pointer, this can cause an invalid memory access or crash. Please either retain staging when hot-cache promotion is possible or make the hot-cache copy path accelerator-aware.

mooncake-store/src/real_client.cpp
```diff
@@ -3476,7 +3477,8 @@ tl::expected<int64_t, ErrorCode> RealClient::execute_ranged_read(
     auto runtime_accelerator =
         device::GetAcceleratorRegistry().RuntimeAccelerators();
     void *dst = static_cast<char *>(buffer) + dst_offset;
-    if (runtime_accelerator.FindDeviceForPointer(dst)) {
+    if (runtime_accelerator.FindDeviceForPointer(dst) &&
```
yokinoshitayoki, line 3497: CanUseLocalMemcpy checks the replica selected by SelectBestReplica, but the ranged Client::Get calls below still receive the unfiltered query_result and independently select the first complete replica. Because the master does not guarantee replica ordering, a remote replica may appear before the selected same-process replica. In that case, this code skips staging based on the local replica but actually passes the GPU pointer to the remote transfer path. Please filter the QueryResult to the selected replica, as the full-object path already does, and use it in both ranged-memory branches.

zxpdemonio: Fixed. The ranged memory path now filters QueryResult to the replica selected by SelectBestReplica and passes that filtered result to both the staging and no-staging Client::Get branches. The locality decision and the actual transfer therefore use the same replica even when the master returns a remote replica first.

mooncake-store/src/real_client.cpp
zxpdemonio: Fixed. Full-object GPU reads now retain staging whenever the local hot cache is enabled, so ProcessSlicesAsync only receives host-accessible slices and LocalHotCacheHandler never sees a CUDA pointer through this path. The no-staging path remains enabled when the selected replica is same-process, local memcpy is enabled, and hot cache is disabled.

yokinoshitayoki:

zxpdemonio:
Thanks for the reminder. I checked the failed jobs. The initial failures were caused by EPEL mirror metadata download errors (404s and timeouts), not by the PR code. During the rerun, one GitHub-hosted runner also lost communication while building the CUDA 13 / Python 3.10 wheel, so I reran that failed job once more. The full workflow is now green: https://github.com/kvcache-ai/Mooncake/actions/runs/31068241065. No source changes were needed.

mooncake-store/src/real_client.cpp
```diff
@@ -3631,6 +3637,20 @@ RealClient::get_into_ranges_internal(
...
+                        const size_t dst_offset = dst_offsets[k];
+                        if (dst_offset > capacities[i] ||
+                            sizes[k] > capacities[i] - dst_offset) {
+                            continue;
+                        }
+                        range_results[k] = execute_ranged_read(
+                            keys[j], buffers[i], dst_offset, src_offsets[k],
+                            sizes[k], metadata, false, false);
+                    }
+                    continue;
```
yokinoshitayoki, line 3700: This branch executes local GPU fragments using the QueryResult cached during planning, then continues before the key is added to scatter_leases. A cached query can be valid but close to expiry, so the transfer may cross the lease boundary; Client::Get checks expiry only after copying. Please refresh/re-resolve the selected metadata before these reads and preserve periodic refresh for long batches.

zxpdemonio: Fixed in 71cc442b. The local GPU range path now bypasses the planning cache and re-resolves the metadata before its first read. For long batches, it refreshes and reselects the replica halfway through the remaining lease; the refreshed metadata is passed to execute_ranged_read, so an updated non-local selection still uses the existing staging path. I verified that real_client.cpp builds in both CUDA and non-CUDA configurations, and the full pybind_client_test suite passes (51/51).

zxpdemonio:
Update summary for the latest head (`73b7864`):

- Added the no-staging read path only for same-process memory replicas when local memcpy is enabled; remote, cross-process, and non-memory replicas keep the existing transfer path.
- Routed CUDA IPC tensor reads and local GPU range reads through the shared ranged-get implementation instead of adding a separate tensor-only transfer stack.
- Hardened CUDA read readiness and buffer lifetime handling, and refreshed the selected replica lease before the direct read completes.
- Reused the canonical local-memcpy eligibility check and removed the duplicate helper/test setup introduced by the initial implementation.
- Added tests for the `MC_STORE_MEMCPY` gate, same-process endpoint matching, ranged reads, and DummyClient CUDA tensor reads.

The current Linux workflow, including format, builds, wheel tests, and `CI Gate`, is green: https://github.com/kvcache-ai/Mooncake/actions/runs/31353420025

yokinoshitayoki (approved):
LGTM.

zxpdemonio:
Supplemental DummyClient tensor-path validation

The existing performance table in the PR description uses a real client. This is an additional end-to-end measurement for the tensor + DummyClient path.

Environment:

- PR head: `73b78648`
- DummyClient process -> standalone real client on the same node
- CUDA IPC destination handoff
- pinned Store segment: 4 GiB
- `MC_STORE_MEMCPY=1`
- protocol: TCP
- client local buffer: 64 MiB
- tensor sizes: 64/256/512 MiB
- warmup/iterations: 3/8
- external one-shot harness, not included in this PR

| API | Size | Throughput | Median latency | Payload check |
| --- | ---: | ---: | ---: | --- |
| `get_tensor_into_cuda` | 64 MiB | 20.107 GB/s | 3.338 ms | passed |
| `get_tensor_into_cuda` | 256 MiB | 22.875 GB/s | 11.735 ms | passed |
| `get_tensor_into_cuda` | 512 MiB | 23.424 GB/s | 22.919 ms | passed |

The destination CUDA tensor was synchronized and checked after each measured series. The real-client log also confirmed that the 4 GiB Store segment was successfully registered as pinned memory. The 256 and 512 MiB objects both exceed the 64 MiB client local buffer and still complete, confirming that the payload does not depend on a full-object client staging buffer.

stmatengss (approved):

PR #3198, opened 2026-07-30
[Bugfix][TE] Split RDMA slices at MR boundaries and handle submit errors

RuiqingFeng:
## Summary
Fixes two critical RDMA transfer bugs in `RdmaTransport`:

1. **MR Boundary Straddling**:
   - **Problem**: When a transfer request straddles two adjacent MRs, a fixed-size slice could span across the MR boundary. Since a single Verbs WR cannot span multiple MRs, this led to `AddressNotRegistered` or transfer failures.
   - **Fix**: Dynamically adjust `slice_length` at MR boundaries so that no single slice crosses into a neighboring MR.

2. **Submit Error Handling**:
   - **Problem**: When `submitTransferTask` encounters a failure mid-submission (e.g. `!found_device`), previously queued slices and unstarted tasks were left unhandled. As a result, the transfer batch hung in `WAITING` state permanently, causing `freeBatchID` to fail with `BatchBusy`.
   - **Fix**: Mark all unposted slices and remaining unstarted tasks as `FAILED` upon submission errors, ensuring the batch completes and resources are safely reclaimed.

## Module
- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

### Test Cases & Covered Issues:
1. **MR Boundary Straddling**:
   - `RDMALargeMrTest.WriteStraddlesChunkBoundary`: Triggered by target buffer straddling across two 64MB MR chunks.
   - `RDMALargeMrTest.WriteWithSourceStraddlingChunkBoundary`: Triggered by source buffer straddling across two 64MB MR chunks.

2. **Submit Error Handling**:
   - `SubmitTransferTaskTest.PartialSubmitFailsBatch`: Triggered when slice 0 succeeds but slice 1 fails device selection mid-submission, verifying batch status transitions to `FAILED` instead of hanging in `WAITING`.
   - `SubmitTransferTaskTest.PartialSubmitFailsAllTasks`: Triggered when task 0 fails mid-submission, verifying subsequent unstarted tasks in the batch are marked `FAILED`.

### Execution:
Verified test failure on pre-fix commits and 100% pass on post-fix:

```bash
MC_MAX_MR_SIZE=67108864 ./mooncake-transfer-engine/tests/rdma_large_mr_test
./mooncake-transfer-engine/tests/rdma_transport_submit_task_test

### DIFF
--- a/mooncake-transfer-engine/src/transport/rdma_transport/rdma_transport.cpp
+++ b/mooncake-transfer-engine/src/transport/rdma_transport/rdma_transport.cpp
@@ -54,6 +54,78 @@ static std::string resolveBufferLocation(
     return location;
 }
 
+// Calculate remaining bytes in buffer for 'address'.
+// Caches 'last_buffer_idx' across sequential slices to avoid rescanning
+// buffers.
+static uint64_t bytesUntilBufferEnd(
+    const TransferMetadata::SegmentDesc *segment_desc, uint64_t address,
+    bool require_remote_key, size_t *last_buffer_idx = nullptr) {
+    if (!segment_desc || segment_desc->buffers.empty()) return 0;
+
+    auto is_valid_buffer = [&](const TransferMetadata::BufferDesc &buffer) {
+#ifdef ENABLE_MULTI_PROTOCOL
+        if (!buffer.protocol.empty() && buffer.protocol != "rdma") return false;
+#endif
+        if (require_remote_key ? buffer.rkey.empty() : buffer.lkey.empty())
+            return false;
+        if (address < buffer.addr || address - buffer.addr >= buffer.length)
+            return false;
+        return true;
+    };
+
+    if (last_buffer_idx && *last_buffer_idx < segment_desc->buffers.size()) {
+        const auto &buffer = segment_desc->buffers[*last_buffer_idx];
+        if (is_valid_buffer(buffer)) {
+            return buffer.length - (address - buffer.addr);
+        }
+    }
+
+    uint64_t max_remaining = 0;
+    for (size_t i = 0; i < segment_desc->buffers.size(); ++i) {
+        const auto &buffer = segment_desc->buffers[i];
+        if (is_valid_buffer(buffer)) {
+            uint64_t remaining = buffer.length - (address - buffer.addr);
+            if (remaining > max_remaining) {
+                max_remaining = remaining;
+                if (last_buffer_idx) *last_buffer_idx = i;
+            }
+        }
+    }
+    return max_remaining;
+}
+
+// Calculate slice length capped at MR boundary to avoid multi-MR WRs.
+// Caches source and target buffer indices across slices for a single request.
+struct SliceLengthCalculator {
+    const Transport::TransferRequest &request;
+    size_t block_size;
+    size_t fragment_size;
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: mooncake-transfer-engine/src/transport/rdma_transport/rdma_transport.cpp [hunk 1]
```diff
@@ -54,6 +54,46 @@ static std::string resolveBufferLocation(
     return location;
 }
 
+static uint64_t bytesUntilBufferEnd(
+    const TransferMetadata::SegmentDesc *segment_desc, uint64_t address,
+    bool require_remote_key) {
+    if (!segment_desc) return 0;
+
+    uint64_t max_remaining = 0;
+    for (const auto &buffer : segment_desc->buffers) {
```
staryxchen: Suggestion (non-blocking): this helper is called twice per slice (source + target), and each call linearly scans the whole buffer list, so submit cost is O(slices x buffers) per request — noticeable with a small MC_SLICE_SIZE or segments with many buffers. Since offsets advance monotonically within a request, keeping the last matched buffer per side and checking it before rescanning would make this amortized O(1) with only a few extra lines.

mooncake-transfer-engine/src/transport/rdma_transport/rdma_transport.cpp
```diff
@@ -746,20 +786,58 @@ Status RdmaTransport::submitTransferTask(
...
     const int kMaxRetryCount = globalConfig().retry_cnt;
     const size_t kFragmentSize = globalConfig().fragment_limit;
     const size_t kSubmitWatermark =
         globalConfig().max_wr * globalConfig().num_qp_per_ep;
+    auto fail_unposted_slices = [&]() {
+        for (auto &entry : slices_to_post)
+            for (auto *slice : entry.second) slice->markFailed();
+        slices_to_post.clear();
+    };
+    auto fail_unstarted_tasks = [&](size_t first_task_index) {
```
staryxchen, line 840: Suggestion (non-blocking): a short comment here would help the next reader — the fabricated zero-length slice is subtle. It exists so the existing `success + failed == slice_count` terminal check (and event-driven `check_batch_completion`) can drive an unstarted task to FAILED without special-casing the status code, and `~TransferTask` reclaims the slice via `slice_list`. Worth saying in one or two lines.

RuiqingFeng: Thanks! Added a explanatory comment above `fail_unstarted_tasks to` clarify the dummy zero-length slice pattern and its state transition.

mooncake-transfer-engine/src/transport/rdma_transport/rdma_transport.cpp
RuiqingFeng: Great catch! Updated. Encapsulated slice length calculation into `SliceLengthCalculator` and cached the last matched source/target buffer index per request to achieve amortized $O(1)$ buffer lookup. Thanks!

staryxchen (approved):
Thanks for the fix! LGTM

PR #3199, opened 2026-07-30
[Store] Add multi-level parallel read via shared ThreadPool

NUABO:
## Description
- Add MC_BATCH_LOAD_THREADS env var for bucket-level parallelism. When > 1 and multiple buckets, dispatch each bucket to a thread.
- Add MC_BUCKET_READ_THREADS env var for intra-bucket POSIX read parallelism. When > 1, split read plans into chunks and dispatch to threads sharing the same fd (pread is thread-safe).

  MC_BATCH_LOAD_THREADS  — bucket-level pool (independent files)
  MC_BUCKET_READ_THREADS — intra-bucket pool (POSIX pread chunking)
     • Both 0 (default):  original sequential path, zero change.
     • Only batch > 1:    buckets dispatched to bucket pool; URing uses
                       batch_read() (one io_uring submission, queue
                       depth 32), POSIX stays sequential per-bucket.
     • Only bucket > 1:   buckets sequential, POSIX plans split across
                       intra-bucket pool.  URing uses batch_read().
     • Both > 1:          bucket pool dispatches buckets, within each
                       POSIX plans further split across intra-bucket
                       pool — true 2-level concurrency.
     • URing batch_read() is enabled when either env var > 0

- Extract UringBatchReadBucket (URing alignment + batch_read + zero-copy) and IntraBucketReadChunked (POSIX chunk-dispatch) to eliminate duplication

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
```

**Test results:**
- [ ] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

### DIFF
--- a/mooncake-store/src/storage_backend.cpp
+++ b/mooncake-store/src/storage_backend.cpp
@@ -19,14 +19,13 @@
 #include <vector>
 #include <algorithm>
 #include <future>
-
-#include "thread_pool.h"
 #include <chrono>
 #include <unordered_set>
 
 #include <ylt/struct_pb.hpp>
 
 #include "mutex.h"
+#include "thread_pool.h"
 #include "utils.h"
 #include "crc32c.h"
 #include "ascii_string.h"
@@ -1680,7 +1679,9 @@ BucketStorageBackend::BucketStorageBackend(
     const BucketBackendConfig& bucket_backend_config_)
     : StorageBackendInterface(file_storage_config_),
       storage_path_(file_storage_config_.storage_filepath),
-      bucket_backend_config_(bucket_backend_config_) {
+      bucket_backend_config_(bucket_backend_config_),
+      batch_threads_(EnvBatchLoadThreads()),
+      bucket_threads_(EnvBucketReadThreads()) {
     // Allocate aligned buffer for O_DIRECT I/O operations
     void* buf = nullptr;
     int ret = posix_memalign(&buf, kDirectIOAlignment, kAlignedBufferSize);
@@ -1695,6 +1696,15 @@ BucketStorageBackend::BucketStorageBackend(
             buf, [](void* p) { free(p); });
         LOG(INFO) << "BucketStorageBackend: Allocated " << kAlignedBufferSize
                   << " bytes aligned buffer at " << buf;
+    }
+
+    if (batch_threads_ > 1) {
+        batch_load_pool_ =
+            std::make_unique<ThreadPool>(static_cast<size_t>(batch_threads_));
+    }
+    if (bucket_threads_ > 1) {
+        intra_bucket_pool_ =
+            std::make_unique<ThreadPool>(static_cast<size_t>(bucket_threads_));
     }
 }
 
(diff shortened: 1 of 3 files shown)


LujhCoconut:
Thanks for the contribution. Before diving into the design, **two things need to be addressed first**.

**1. This PR does not compile.** `BucketStorageBackend(const FileStorageConfig&, const BucketBackendConfig&)` ends up with two identical out-of-line definitions in `storage_backend.cpp` (the pre-existing one and the one added by this PR), which is a redefinition error. Note the new one also drops the initialization of `storage_path_` and the `aligned_io_buffer_` allocation — the thread-count/pool setup should be merged into the existing constructor instead. 

_**A ready-for-review PR should at least build locally — please make sure the code compiles and passes basic tests (e.g. a `BatchLoad` smoke test with the new env vars enabled) before submitting, and also run the format checks (`./scripts/code_format.sh`) before (re)submitting.**_


**2. The intra-bucket path is incompatible with io_uring + O_DIRECT.** With `use_uring` enabled, bucket files are opened with `O_DIRECT`, but the intra-bucket chunked path issues `vector_read()` at `plan.offset + plan.key_size`, which is generally not 4096-aligned, so every read fails with `EINVAL`. And even setting correctness aside, I don't think this path is necessary on the uring backend: N threads each issuing blocking per-key reads only reach aggregate queue depth N at N times the syscall/thread cost, whereas a single `batch_read()` submission reaches queue depth 32 from one thread — threads are the wrong tool for parallelism *within* one file when io_uring is available.

The design would hold up much better as: **within a bucket, use `batch_read()` on the uring path and fall back to threaded `pread` only for the non-uring (buffered POSIX) backend**; bucket-level threading across independent files can stay as-is. This combination is worth thinking through carefully — happy to discuss before you invest more time in the current direction.

NUABO:
> Thanks for the contribution. Before diving into the design, **two things need to be addressed first**.
> 
> **1. This PR does not compile.** `BucketStorageBackend(const FileStorageConfig&, const BucketBackendConfig&)` ends up with two identical out-of-line definitions in `storage_backend.cpp` (the pre-existing one and the one added by this PR), which is a redefinition error. Note the new one also drops the initialization of `storage_path_` and the `aligned_io_buffer_` allocation — the thread-count/pool setup should be merged into the existing constructor instead.
> 
> _**A ready-for-review PR should at least build locally — please make sure the code compiles and passes basic tests (e.g. a `BatchLoad` smoke test with the new env vars enabled) before submitting, and also run the format checks (`./scripts/code_format.sh`) before (re)submitting.**_
> 
> **2. The intra-bucket path is incompatible with io_uring + O_DIRECT.** With `use_uring` enabled, bucket files are opened with `O_DIRECT`, but the intra-bucket chunked path issues `vector_read()` at `plan.offset + plan.key_size`, which is generally not 4096-aligned, so every read fails with `EINVAL`. And even setting correctness aside, I don't think this path is necessary on the uring backend: N threads each issuing blocking per-key reads only reach aggregate queue depth N at N times the syscall/thread cost, whereas a single `batch_read()` submission reaches queue depth 32 from one thread — threads are the wrong tool for parallelism _within_ one file when io_uring is available.
> 
> The design would hold up much better as: **within a bucket, use `batch_read()` on the uring path and fall back to threaded `pread` only for the non-uring (buffered POSIX) backend**; bucket-level threading across independent files can stay as-is. This combination is worth thinking through carefully — happy to discuss before you invest more time in the current direction.

hi @LujhCoconut thanks for thorough review, both points were spot on.

  Compilation and constructor: The duplicate constructor has been removed. 

  Intra-bucket path and io_uring: You were right — threading within a bucket made no sense for the URing backend. The revised design follows your suggestion:

  - URing files within a bucket: always use batch_read() — a single io_uring submission with queue depth up to 32, regardless of thread-pool settings. No intra-bucket threads touch URing files.
  - POSIX (buffered) files within a bucket: optional intra-bucket parallelism via a dedicated ThreadPool, gated on MC_BUCKET_READ_THREADS > 1. Plans are split into chunks and dispatched; pread is thread-safe
  on the same fd.
  - Bucket-level parallelism: unchanged — a separate ThreadPool gated on MC_BATCH_LOAD_THREADS > 1 dispatches independent bucket files across threads.

  The two pools are independent, so when both env vars are set bucket-level dispatch and intra-bucket POSIX chunking compose into true 2-level concurrency. URing files are never dispatched to either pool for
  intra-bucket reads — they always go through batch_read().

  When both env vars are 0 (the default), the code falls through to the original sequential path with zero change.

  Happy to iterate further if anything still looks off.

mooncake-store/src/storage_backend.cpp
```diff
@@ -1858,19 +1875,129 @@ tl::expected<void, ErrorCode> BucketStorageBackend::BatchQuery(
...
+                   << " error=" << result.error();
+        return tl::make_unexpected(result.error());
+    }
+
+    // Guard against silent short reads (community #3066/#3073): the
+    // underlying collect() only flags cqe->res < 0, so a short read
+    // returns fewer bytes without surfacing an error.  Verify the
+    // aggregated byte count covers every key's actual data region
+    // (offset_in_buffer + data_size) before exposing the pointers.
+    if (result.value() < required_bytes) {
```
fcczzz, line 1928: Comparing only the aggregate byte count cannot prove that every descriptor returned enough data. For example, descriptor A may require only 512 useful bytes but return its full 4096-byte aligned range, while descriptor B requires 2048 bytes but returns 0 at EOF. The aggregate check still passes (`4096 >= 512 + 2048`), and the incomplete pointer for B is exposed below.

`io_uring` provides an individual `cqe->res` for each request, but `batch_read()` currently sums those values and discards the per-descriptor results. Please preserve/map each completion length—e.g. by encoding the descriptor index in `user_data`—and verify every result against `buf_offsets[i] + plans[i].dest_slice.size` before adjusting any pointers. A regression test using a truncated bucket file would cover this path.

fcczzz (changes requested):
Thanks for addressing the previous feedback. I found one remaining correctness issue in the io_uring batch short-read validation; details are in the inline comment. Once per-descriptor completion lengths are validated and covered by a regression test, I’m happy to take another look.

PR #3201, opened 2026-07-30
[Store] Add producer view to durable oplog prefix

Icedcoco:
> [!IMPORTANT]
> This PR is stacked on #3165 and must not be merged before #3165.
> After #3165 is merged, this branch will be rebased onto the updated `main`.

## Description

This PR adds a backward-compatible producer view field to the durable oplog prefix.

- Add `producer_view_version` to `DurablePrefix` while keeping schema version 1.
- Preserve byte-identical legacy encoding when the producer view is zero.
- Validate producer view values during decoding and leave the output unchanged on decode failure.
- Carry the producer view across atomic batch and durable-prefix commits.
- Include the producer view in compare-failure idempotency checks.
- Support exact-value CAS for valid non-canonical prefixes, such as an explicitly encoded zero producer view.

This PR only introduces the format and storage semantics needed by later HA work. It does not claim producer views, fence stale writers, or switch the production writer.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**

```bash
  oplog_batch_codec_test \
  oplog_batch_storage_test \
  oplog_batch_standby_reader_test \
  ordered_oplog_writer_test -j32

LSAN_OPTIONS=detect_leaks=0 ./build/mooncake-store/tests/oplog_batch_codec_test
LSAN_OPTIONS=detect_leaks=0 ./build/mooncake-store/tests/oplog_batch_storage_test
LSAN_OPTIONS=detect_leaks=0 ./build/mooncake-store/tests/oplog_batch_standby_reader_test
LSAN_OPTIONS=detect_leaks=0 ./build/mooncake-store/tests/ordered_oplog_writer_test
```

**Test results:**

- `oplog_batch_codec_test`: 34/34 passed
- `oplog_batch_storage_test`: 38/38 passed
- `oplog_batch_standby_reader_test`: 20/20 passed
- `ordered_oplog_writer_test`: 28/28 passed
- Pre-commit hooks passed for all files changed by this PR

- [x] Unit tests pass
- [ ] Manual testing done

## Checklist

- [x] I have performed a self-review of every changed line

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex was used to refine the design, implement the code and tests, investigate review findings, and run verification. The human submitter is responsible for reviewing and defending every changed line.

### DIFF
--- a/mooncake-store/include/ha/oplog/oplog_batch_types.h
+++ b/mooncake-store/include/ha/oplog/oplog_batch_types.h
@@ -17,7 +17,7 @@ struct DurablePrefix {
     uint64_t last_seq{0};
     // Leadership view that produced this durable boundary. Zero means that
     // the prefix has no producer-view metadata (legacy or not yet assigned).
-    uint64_t producer_view_version{0};
+    ViewVersionId producer_view_version{0};
 
     friend bool operator==(const DurablePrefix&,
                            const DurablePrefix&) = default;
--- a/mooncake-store/include/ha/oplog/oplog_batch_types.h
+++ b/mooncake-store/include/ha/oplog/oplog_batch_types.h
@@ -15,6 +15,7 @@ static constexpr int kOpLogBatchIdWidth = 20;
 struct DurablePrefix {
     uint64_t batch_id{0};
     uint64_t last_seq{0};
+    uint64_t producer_view_version{0};
 };
 
 struct BatchRecordRange {
--- a/mooncake-store/include/ha/oplog/oplog_batch_types.h
+++ b/mooncake-store/include/ha/oplog/oplog_batch_types.h
@@ -15,7 +15,12 @@ static constexpr int kOpLogBatchIdWidth = 20;
 struct DurablePrefix {
     uint64_t batch_id{0};
     uint64_t last_seq{0};
+    // Leadership view that produced this durable boundary. Zero means that
+    // the prefix has no producer-view metadata (legacy or not yet assigned).
     uint64_t producer_view_version{0};
+
+    friend bool operator==(const DurablePrefix&,
+                           const DurablePrefix&) = default;
 };
 
 struct BatchRecordRange {
--- a/mooncake-store/src/ha/oplog/oplog_batch_storage.cpp
+++ b/mooncake-store/src/ha/oplog/oplog_batch_storage.cpp
@@ -219,11 +219,27 @@ ErrorCode OpLogBatchStorage::WriteBatchAndAdvancePrefix(
          .expected_value = EncodeDurablePrefix(expected_prefix)});
     txn.puts.push_back({.key = BuildBatchRecordKey(cluster_id_, batch.batch_id),
                         .value = encoded_batch});
-    txn.puts.push_back(
-        {.key = durable_key,
-         .value = EncodeDurablePrefix(
-             {.batch_id = batch.batch_id, .last_seq = batch.last_seq})});
+    txn.puts.push_back({.key = durable_key,
+                        .value = EncodeDurablePrefix(
+                            {.batch_id = batch.batch_id,
+                             .last_seq = batch.last_seq,
+                             .producer_view_version =
+                                 expected_prefix.producer_view_version})});
     ErrorCode err = backend_.Txn(txn);
+    if (err == ErrorCode::ETCD_TRANSACTION_FAIL) {
+        std::string raw_prefix;
+        DurablePrefix decoded_prefix;
+        if (backend_.Get(durable_key, raw_prefix) == ErrorCode::OK &&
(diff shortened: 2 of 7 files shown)


mooncake-store/include/ha/oplog/oplog_batch_types.h
```diff
@@ -15,6 +15,7 @@ static constexpr int kOpLogBatchIdWidth = 20;
 struct DurablePrefix {
     uint64_t batch_id{0};
     uint64_t last_seq{0};
+    uint64_t producer_view_version{0};
```
Aionw: Will it be a problem to use optional<uint64_t> to express empty semantic here?

mooncake-store/src/ha/oplog/oplog_batch_storage.cpp
```diff
@@ -104,7 +104,11 @@ ErrorCode OpLogBatchStorage::InitDurablePrefix(DurablePrefix& prefix) {
         return ErrorCode::OK;
     }
     if (err == ErrorCode::ETCD_TRANSACTION_FAIL) {
-        return ReadDurablePrefix(prefix);
+        err = ReadDurablePrefix(prefix);
```
Aionw: Nit: can be simplified using if assign statement.

mooncake-store/src/ha/oplog/oplog_batch_storage.cpp
```diff
@@ -170,19 +219,37 @@ ErrorCode OpLogBatchStorage::WriteBatchAndAdvancePrefix(
...
+                             .producer_view_version =
+                                 expected_prefix.producer_view_version})});
     ErrorCode err = backend_.Txn(txn);
+    if (err == ErrorCode::ETCD_TRANSACTION_FAIL) {
+        std::string raw_prefix;
+        DurablePrefix decoded_prefix;
+        if (backend_.Get(durable_key, raw_prefix) == ErrorCode::OK &&
+            raw_prefix != txn.compares[0].expected_value &&
+            DecodeDurablePrefix(raw_prefix, &decoded_prefix) &&
+            decoded_prefix.batch_id == expected_prefix.batch_id &&
```
Aionw: Better to give `struct DurablePrefix` a compare operator to simplify this.

Icedcoco: Done. DurablePrefix now has a default equality operator, and the decoded-prefix and idempotency checks use it instead of comparing each field separately.

mooncake-store/include/ha/oplog/oplog_batch_types.h
Icedcoco: Thanks for asking. producer_view_version is the leadership view/epoch of the primary that produced this durable boundary. Zero intentionally means no producer-view metadata, covering legacy prefixes where the JSON field is absent and prefixes before a view is assigned. Encoding zero omits the field to preserve the legacy representation, so absent and explicit zero are deliberately the same semantic value here; std::optional<uint64_t> would add a distinction the format does not currently need. I added a field comment to make this explicit.

mooncake-store/src/ha/oplog/oplog_batch_storage.cpp
Icedcoco: Done. This now uses the assignment in the condition and returns the read error directly.

mooncake-store/include/ha/oplog/oplog_batch_types.h
00fish0: Nit: since this field represents a leadership view, should it use `ViewVersionId` (or validate against `ViewVersionId`'s max) rather than accepting all `uint64_t` values?

Other HA producer/view fields are `ViewVersionId` (`int64_t` today). Accepting `UINT64_MAX` makes values outside the existing view domain part of the durable format, which may be awkward once this is used for fencing/view comparisons.

Not blocking for this PR, but aligning the persisted value domain now seems cleaner.

00fish0 (approved):
Thanks for the great work!

mooncake-store/include/ha/oplog/oplog_batch_types.h
Icedcoco: Good point. I changed `DurablePrefix::producer_view_version` to `ViewVersionId` in d3e4cd114 and aligned the JSON codec with the signed view domain. Decoding now rejects negative values and values above `std::numeric_limits<ViewVersionId>::max()`, while preserving the existing zero/absent legacy semantics. The codec tests cover the maximum valid value, `-1`, and the first out-of-range value.

PR #3205, opened 2026-07-30
feat(rdma): enable PCI relaxed ordering by default

1998zxn:
## Description

Change the default value of `MC_IB_PCI_RELAXED_ORDERING` from `0` (disabled)
to `1` (enabled if supported by hardware).

Without relaxed ordering, bursty RDMA traffic can saturate switch buffers and trigger PFC (Priority Flow Control) pause frames. PFC, while designed to prevent packet loss in RoCE networks, introduces deeper problems: pause frames propagate upstream causing Head-of-Line Blocking, and in extreme cases can lead to PFC deadlock where traffic across an entire switch domain stalls. This significantly increases the operational burden of managing the entire cluster.

Tested across multiple real-world scenarios involving large-scale KV cache transfers with concurrent multi-node RDMA clients. With RO enabled:
- No functional regressions (no transfer failures or data errors)
- Significant performance improvement observed
- Network congestion and PFC trigger frequency observably reduced

Changing the default also ensures users who are unaware of this option do not silently miss out on the performance benefit.

The option remains fully overridable via MC_IB_PCI_RELAXED_ORDERING=0.

Update env-var descriptions and documentation accordingly.

## Module

- [*] Transfer Engine (`mooncake-transfer-engine`)
- [*] Docs

## Type of Change

- [*] Performance improvement

## Checklist

- [*] I have performed a self-review of my own code
- [*] I have formatted my code using `./scripts/code_format.sh`
- [*] I have added tests to prove my changes are effective


stmatengss:
@alogfans

alogfans (approved):

Issue #3200, opened 2026-07-30
[RFC]: Decouple Mooncake PG from PyTorch


caozhanhao:
## Summary

This RFC proposes decoupling the Mooncake PG into two distinct layers:

* **PG Core (`libmooncake_pg.so`)**: A separate C++ shared library built alongside the Transfer Engine via the top-level CMake project. It contains zero dependencies on PyTorch.
* **PG Torch**: A lightweight PyTorch extension that maintains the existing `c10d` and Python interfaces by acting as an adapter on top of PG Core.

PG Core will expose a preliminary C API (`mooncake_pg.h`) for native C++ tests, sanitizers, micro-benchmarks, and non-PyTorch runtime integrations. The initial API is experimental and does not guarantee source or ABI compatibility.

## Motivation

Currently, Mooncake PG is compiled as a PyTorch extension, whereas the Transfer Engine is compiled directly by the top-level CMake build system. Because these two components are built under different build contexts, shared header-only dependencies can be selected or configured inconsistently across the build boundary. This leads to incompatible template instantiations and subtle ABI mismatches at runtime (such as the YLT header mismatch addressed in #3164).

Furthermore, the existing build architecture tightly couples the entire PG implementation to the PyTorch C++ ABI. Consequently:
1. **ABI Fragmentation**: Every supported PyTorch C++ ABI version requires a complete, dedicated rebuild of the entire PG library.
2. **Heavyweight Testing Overhead**: Native sanitizer runs and standalone tests must link against PyTorch, while the existing test workflow also runs through Python.

By migrating the core communication logic into the top-level CMake build, PG Core and the Transfer Engine will share a single, unified dependency graph. PyTorch ABI-sensitive code is strictly isolated to the thin `PG Torch` adapter layer.

## Architecture & Design

```text
+-----------------------+
|  Native Application   |
+-----------+-----------+
            |
            | (mooncake_pg.h)
            v
+-----------------------+              +-----------------------+
|        PG Core        | -----------> |    Transfer Engine    |
| (`libmooncake_pg.so`) |              +-----------------------+
+-----------------------+
            ^
            |
            | (C API Boundary)
+-----------+-----------+
|       PG Torch        |
+-----------+-----------+
            |
            v
+-----------------------+
|   torch.distributed   |
+-----------------------+
```

### Layer Boundary (`mooncake_pg.h`)

[...]

PR #3206, opened 2026-07-30
[PG] Decouple PG from PyTorch

caozhanhao:
## Description

Implements #3200.

### Motivation

Mooncake PG was previously compiled as part of a PyTorch extension, while the Transfer Engine was built by the top-level CMake project. The separate build contexts could select or configure shared header-only dependencies differently, causing ABI mismatches such as #3164.

The previous layout also coupled the entire PG implementation to the PyTorch C++ ABI, requiring a complete PG rebuild for every supported PyTorch version and making native testing dependent on PyTorch.

### Architecture

This PR separates Mooncake PG into a framework-neutral core and a thin PyTorch adapter:

```text
torch.distributed
        |
        v
PG Torch adapter (`pg_<torch-version>.so`)
        |
        | experimental C API (`mooncake_pg.h`)
        v
PG Core (`libmooncake_pg.so`)
        |
        v
Transfer Engine
```

After this change:

* `libmooncake_pg.so` is built once by the top-level CMake project and shares the Transfer Engine dependency graph.
* PG Core owns the control plane, communicators, collective and P2P workers, fault handling, completion state, and GPU runtime integration without depending on PyTorch.
* Each PyTorch-specific extension contains only the `c10d`/Tensor adapter and links to `libmooncake_pg.so`.
* PG Core is exposed through opaque context, communicator, completion, and stream handles in `mooncake_pg.h`.
* Existing Python and `torch.distributed` entry points remain available.

For the full design rationale, alternatives, and C API inventory, see #3200.

#### Result-based error propagation

PG Core can no longer use `TORCH_CHECK` or allow C++ exceptions to cross the ABI boundary. Fallible operations across the communicator, control plane, RPC layer, and Transfer Engine integration now return:

```cpp
PGResult<T>  // ylt::expected<T, PGError>
```

`PGError` contains a common `PGErrorCode` and diagnostic message. The following helpers standardize propagation and validation:

* `PG_TRY(expression)` returns immediately when an operation fails.
* `PG_TRY(lhs, expression)` additionally unwraps the successful value into `lhs`.
* `PG_TRY_TE(expression)` converts Transfer Engine return codes into `TransferEngineError`.
* `PG_TRY_CUDA(expression)` converts CUDA runtime failures into `SystemError`.
* `PG_VALIDATE_ARG` and `PG_VALIDATE_STATE` report caller and state errors consistently.
* `PG_ASSERT` represents internal bugs rather than expected runtime failures.

At the C boundary, `PGErrorCode` is translated to `mooncakePgResult_t`, with a thread-local diagnostic available through `mooncakePgGetLastError()`. Assertion failures, allocation failures, and other unexpected C++ exceptions are caught and converted instead of escaping across a potentially incompatible C++ ABI.

PG Torch converts non-success C results back into `TORCH_CHECK` failures, preserving the existing Python-facing exception behavior.

#### Explicit ownership and lifetime rules

PG Core no longer owns or retains `at::Tensor` objects. The C API therefore makes resource ownership explicit:

* Send and receive buffers are caller-owned and must remain valid until the corresponding operation completes.
* `failedRanksHint` is a caller-owned output buffer.
* The optional active-ranks mirror is caller-owned.
* An injected Transfer Engine is non-owning and must outlive the PG context.

PG Torch retains user tensors, failed-rank tensors, packed temporary buffers, completion handles, and CUDA events for the required lifetime.

`MooncakeWorkTracker` retains these resources when a user releases a `c10d::Work` before the underlying operation completes. Resources referenced by a captured CUDA graph are retained until backend shutdown because the graph may be replayed after the original `Work` object has been released.

#### Framework-neutral completion semantics

PyTorch futures and events are replaced at the core boundary with explicit completion contracts:

* CPU collectives and CPU/GPU P2P operations return an opaque `mooncakePgCompletion_t`.
* GPU collectives are ordered on the caller-provided stream and do not return a host completion handle.
* GPU `Work::wait()` continues to establish stream ordering without synchronously blocking the host.
* GPU collective `waitUntilTasksSubmitted` fencing now occurs inside PG Core before the C API call returns; this was previously performed by the Torch `Work` object.
* CPU copy-back callbacks execute after the corresponding completion becomes ready.

#### Tensor-list adaptation

The C API accepts contiguous raw buffers, while several `c10d` operations accept lists of tensors.

PG Torch now packs list inputs into contiguous temporary buffers before calling PG Core and copies results back after completion. The adapter retains both temporary and destination tensors so CPU copy-back remains asynchronous and CUDA copy-back remains correctly ordered on the caller’s stream.

### Compatibility notes

* No intentional changes are made to the existing Python or `torch.distributed` APIs.
* The new C API is preliminary and does not yet guarantee source or ABI compatibility.
* Native C API callers must follow the documented buffer, stream, completion, and Transfer Engine lifetime requirements.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

- All tests under `mooncake-pg/tests` passed.
- recovery and scale-up e2e tests are also passed

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex was used to assist in designing and implementing this PR. All code and architectural changes have been thoroughly reviewed and verified by me in person.


mooncake-pg/src/mooncake_communicator.cpp
```diff
@@ -0,0 +1,1650 @@
...
+
+    // The data-plane teardown has finished. Remove the group from the local
+    // Agent and notify the Coordinator that this rank has left it.
+    if (isValidGroup()) {
+        PG_TRY(agent_.unregisterGroup(meta_->group_id));
+    }
+    return {};
+}
+
+PGResult<std::string> MooncakeCommunicator::getPreferredHca(
```
UNIDY2002: This API was internally used by Mooncake EP. Now that Mooncake EP no longer needs this, and as far as I know, there is no external usage, probably we can discard this API (either now or in the future).

UNIDY2002 (commented):
Please pay attention to the CI failures:
https://github.com/kvcache-ai/Mooncake/actions/runs/30874424008/job/91906939986?pr=3206

mooncake-pg/src/mooncake_communicator.cpp
caozhanhao: Removed in fa8fd06931407a732d8f256d059e5cd552d56c14

caozhanhao:
> Please pay attention to the CI failures: https://github.com/kvcache-ai/Mooncake/actions/runs/30874424008/job/91906939986?pr=3206

Fixed in 26f279b52d5c7577d409e60b7f10c9caeb777ecb

`libmooncake_pg.so` now statically links Transfer Engine, which brings in dependencies such as yaml-cpp. Because it was added to the wheel after auditwheel, its dependency chain was never repaired. The fix moves the CUDA kernels into a separate `libmooncake_pg_device.so`, allowing the host-only `libmooncake_pg.so` to go through auditwheel while keeping the CUDA fatbins untouched.

An alternative would be to follow #2883 and resolve the Transfer Engine symbols through `engine.so`. We chose the host/device split here because PG now has its own native C API, and keeping its dependencies within the native layer gives it a cleaner dependency boundary.

UNIDY2002 (approved):
Tested on H20 / CUDA 13 environments with both Torch 2.11 and Torch 2.13 builds.

- Built and installed the CUDA 13 wheels, and verified the versioned PG/EP adapters load correctly.
- Passed native PG C API coverage and CUDA PG collectives, P2P, initialization/reinitialization, inference topology, and elastic recovery tests.
- Passed SGLang Mooncake Elastic EP healthy-serving and CUDA-graph-enabled recovery validation on the Torch 2.11 runtime.
- Ran a fixed-seed four-GPU SGLang performance comparison against the base commit; median output throughput remained within the 5% acceptance threshold.

LGTM.

PR #3209, opened 2026-07-30
[TransferEngine] Reject peer segment descriptors with more keys than devices

eharris128:
## Description

A peer segment descriptor carries `devices[]`, `buffers[].rkey[]` and `priority_matrix` as three independent fields, and `decodeSegmentDesc()` never checks them against each other. The rdma, efa and cxi transports pick a device with `Topology::selectDevice()`, whose HCA list is built from the distinct HCA names in that `priority_matrix`, and use the resulting `device_id` to index both `buffers[buffer_id].rkey` and `devices[]`. Its range is therefore decoupled from `devices.size()`.

A descriptor with a short `devices` array, a longer `rkey` vector and many distinct HCAs drives `device_id` past the end of `devices[]`. The rkey bound (in `WorkerPool::selectPeerDevice` for rdma, inside `selectDevice` itself for efa and cxi) still passes, and the subsequent `devices[device_id].name` accesses read a `DeviceDesc` out of bounds and copy a wild `std::string`. In the RDMA transport those are `worker_pool.cpp:216` and `:256`, running in a transfer worker thread with no enclosing try/catch, so the resulting SIGSEGV is an uncaught process crash.

This change:

- Requires `rkey.size() == devices.size()` for a non-empty rkey vector in `decodeSegmentDesc()`, on both the single-protocol and multi-protocol paths. A publisher emits exactly one key per device (one per context in `context_list_`), so this extends the existing rkey/lkey consistency check, rejects the malformed peer once at decode time instead of per-slice, and covers all three transports. barex is excluded because it hands the whole rkey vector to the peer rather than indexing it by `device_id`, and its key count comes from the mempool MR set rather than the device list. Buffers with an empty rkey (registered local-only) stay valid.
- Keeps a `device_id < devices.size()` bound as defense in depth on the value actually used to index the array, in all three submit paths: `WorkerPool::selectPeerDevice()` for rdma, and the per-slice resolve loops in `efa_context.cpp` and `cxi_context.cpp`.
- Adds two tests over the real decode path: a peer publishing 64 keys for 1 device is rejected, and a well-formed 4-key/4-device descriptor still resolves.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

Two new unit tests in `transfer_metadata_test.cpp` exercise the real P2P-handshake decode path:

- `TransferMetadataValidationTest.RejectsMoreKeysThanDevices` — a peer descriptor with 64 keys and 1 device is rejected at decode time (`getSegmentID` returns -1).
- `TransferMetadataValidationTest.AcceptsOneKeyPerDevice` — the legitimate multi-NIC shape (4 devices, 4 keys) still decodes, with device and key counts preserved.

**Test commands:**
```bash
cmake --build build --target transfer_metadata_test
./build/mooncake-transfer-engine/tests/transfer_metadata_test
```

**Test results:**
- [ ] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

Claude Code assisted with reviewing the change and drafting this PR description. The human submitter is responsible for understanding and defending all changes.

### DIFF
--- a/mooncake-transfer-engine/src/transfer_metadata.cpp
+++ b/mooncake-transfer-engine/src/transfer_metadata.cpp
@@ -672,15 +672,20 @@ decodeMultiProtocolSegmentDesc(Json::Value &segmentJSON,
                 buffer.lkey.push_back(
                     static_cast<decltype(buffer.lkey)::value_type>(
                         lkeyJSON.asUInt64()));
+            // The same topology-derived device_id indexes both rkey and
+            // devices, and a publisher emits exactly one key per device, so a
+            // key vector longer than the device list lets that index run past
+            // the end of devices[]. See the comment in decodeSegmentDesc.
             if (buffer.name.empty() || !buffer.addr || !buffer.length ||
                 (!buffer.rkey.empty() &&
-                 buffer.rkey.size() != buffer.lkey.size())) {
+                 (buffer.rkey.size() != buffer.lkey.size() ||
+                  buffer.rkey.size() != desc->devices.size()))) {
                 LOG(WARNING)
                     << "Corrupted segment descriptor, name " << segment_name
                     << " buffer_protocol " << buffer_protocol << ", "
                     << buffer.name << ", " << buffer.addr << ", "
                     << buffer.length << ", " << buffer.rkey.size() << ", "
-                    << buffer.lkey.size();
+                    << buffer.lkey.size() << ", " << desc->devices.size();
                 return nullptr;
             }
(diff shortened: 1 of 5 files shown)


alogfans (approved):
LGTM

PR #3212, opened 2026-07-31
[CI] synchronize metadata server handoff

zhangzuo21:
## Summary

Fix a flaky CI failure caused by a race during the metadata server handoff in `scripts/run_tests.sh`.

## Root cause

The workflow started `mooncake_http_metadata_server` in a separate GitHub Actions step.

Later, `run_tests.sh` used `killall` to terminate that process and immediately started `mooncake_master`, which provides its own HTTP metadata server on the same port (`8080`).

`killall` only sends `SIGTERM` and does not wait for the process to release the port. If the standalone server was still shutting down, `mooncake_master` failed to bind to port `8080` and exited. This caused `store.setup()` to fail after its retries, resulting in:

```text
AssertionError: -1 != 0
## Module

- [✅] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [✅] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?



**Test commands:**
```bash
```

**Test results:**
- [ ] Unit tests pass

## Checklist

- [✅] I have performed a self-review of my own code

## AI Assistance Disclosure

- [ ] No AI tools were used


Aionw (approved):

staryxchen (commented):
Please make sure there is no CI job rely on `mooncake_http_metadata_server` to run

zhangzuo21:
> Please make sure there is no CI job rely on `mooncake_http_metadata_server` to run

Thanks for catching this. run_tests.sh stops its standalone metadata server before the later workflow steps run, so port 8080 would otherwise no longer be served. I updated the subsequent Start Mooncake Master step to pass --enable_http_metadata_server=true, ensuring the Python API/performance checks can continue using http://127.0.0.1:8080/metadata.

Icedcoco:
Hi, @zhangzuo21 . One thing I wanted to double-check is how this interacts with the nightly workflow. `nightly.yml` starts `bootstrap_server.py` on port 8080 and later invokes `scripts/run_tests.sh`, while this change makes `run_tests.sh` start another metadata server unconditionally.

Since `metadata_port_is_listening` checks whether anything is listening on 8080, the existing nightly server could satisfy the readiness check even if the newly spawned process failed to bind. During cleanup, the original nightly server would still hold the port, so `stop_metadata_server` would eventually return an error under `set -e`.

I realize the nightly path already has a related ownership mismatch before this PR, so I am not sure whether it is intended to be addressed here. Would it make sense to make the server ownership explicit, for example by reusing a caller-provided server/PID or stopping the nightly server before `run_tests.sh` starts its own? Or is there some cleanup behavior in the workflow that I may have missed?

zhangzuo21:
> Hi, @zhangzuo21 . One thing I wanted to double-check is how this interacts with the nightly workflow. `nightly.yml` starts `bootstrap_server.py` on port 8080 and later invokes `scripts/run_tests.sh`, while this change makes `run_tests.sh` start another metadata server unconditionally.
> 
> Since `metadata_port_is_listening` checks whether anything is listening on 8080, the existing nightly server could satisfy the readiness check even if the newly spawned process failed to bind. During cleanup, the original nightly server would still hold the port, so `stop_metadata_server` would eventually return an error under `set -e`.
> 
> I realize the nightly path already has a related ownership mismatch before this PR, so I am not sure whether it is intended to be addressed here. Would it make sense to make the server ownership explicit, for example by reusing a caller-provided server/PID or stopping the nightly server before `run_tests.sh` starts its own? Or is there some cleanup behavior in the workflow that I may have missed?

Thanks for pointing this out. You are right—there was no cleanup step between the nightly metadata-server startup and run_tests.sh.
I updated run_tests.sh to support two explicit ownership modes:
managed (default): the script starts and stops its own metadata server.
external: the script reuses a caller-managed server, verifies that port 8080 is available, and does not stop that server.
The nightly workflow now selects external mode. In that mode, the final mooncake_master is also started with its embedded HTTP metadata server disabled, so it continues using the existing server without attempting to bind port 8080.
This avoids the duplicate bind while preserving ownership of the nightly server with the workflow that started it.

Icedcoco:
Thanks for addressing the metadata-server ownership issue. I reviewed the latest changes and haven’t found any other issues.

The current Spell Check failure is unrelated to this PR: the `mis-set` typo was introduced in #3296 and has already been fixed in #3309. Could you merge the latest `main` into this branch and rerun CI? Once the checks are green, this should be good to merge from my side. @zhangzuo21

Icedcoco (approved):

PR #3214, opened 2026-07-31
[Store][TE] fabric_mem best-effort alloc via percentile ladder

ascend-direct-dev:
## Description

When Ascend Store mounts a fabric_mem segment (`ascend` / `ubshmem` + `ascend_use_fabric_mem`), allocation is now best-effort instead of hard-failing at the configured size.

- Try sizes on a percentile ladder: **100% → 90% → … → 50%** of the target, each attempt **1G-aligned**.
- On success with a smaller size, mount the **actual** allocated size and log a WARNING (`Fabric mem best-effort: target=..., actual=...`).
- Fail only if even 50% of the target cannot be allocated.
- Intermediate probe failures are quiet (no ERROR spam during retries).
- Exact-size fabric alloc keeps original semantics: prefer `adxl::MallocMem` when the weak symbol is present; **adxl failure does not fall back to direct VMM** (direct VMM only when the symbol is absent).
- No new config / env knobs; best-effort applies whenever the fabric_mem path is used.
- Non-Ascend builds: the best-effort call site is guarded by `USE_ASCEND_DIRECT` / `USE_UBSHMEM` so GPU/CPU-only CI stays clean.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
# Unit tests (Ascend / ACL mock path)
ctest -R ascend_direct_transport_test --output-on-failure

# Or focused gtest filter
./mooncake-transfer-engine/tests/ascend_direct_transport_test \
  --gtest_filter='FabricMemBestEffortAllocTest.*'
```

**Test results:**
- [x] Unit tests pass (`FabricMemBestEffortAllocTest`: percentile ladder / below 50% / full target)
- [x] Integration tests pass (if applicable) — CI Build & Test (Linux) green, including `ascend-test`, tent, arm64, Docker

Manual (optional on A3):
- Configure `global_segment_size` larger than available contiguous fabric mem → Store starts with WARNING and mounted capacity < configured
- With enough memory → 100% path succeeds (no shrink)
- With available fabric mem < 50% of configured → startup fails clearly

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

Cursor was used to implement the best-effort allocation ladder, wire Store mount to the actual size, add ACL-mock unit tests, and fix non-Ascend CI compile guards.

### DIFF
--- a/mooncake-transfer-engine/src/transport/ascend_transport/ascend_allocator.cpp
+++ b/mooncake-transfer-engine/src/transport/ascend_transport/ascend_allocator.cpp
@@ -4,91 +4,21 @@
 #include <vector>
 
 #include <glog/logging.h>
+#include "ascend_allocator.h"
 #include "config.h"
 #include "acl/acl.h"
 #include "transport/ascend_transport/ascend_direct_transport/adxl_compat.h"
 
 namespace mooncake {
 namespace {
-constexpr size_t kFabricMemPageSize = 1024 * 1024 * 1024;  // 1G
-#ifdef ASCEND_SUPPORT_FABRIC_MEM
-struct AllocRecord {
-    aclrtDrvMemHandle handle;
-    bool is_direct_alloc;
-};
-std::mutex g_vmm_alloc_mutex;
-std::unordered_map<void *, AllocRecord> g_vmm_alloc_records;
+constexpr size_t kFabricMemPageSize = 1024ULL * 1024 * 1024;  // 1G
+constexpr int kBestEffortStartPercent = 100;
+constexpr int kBestEffortMinPercent = 50;
+constexpr int kBestEffortPercentStep = 10;
 
-int allocate_physical_memory(size_t total_size, aclrtDrvMemHandle &handle) {
-    int32_t user_dev_id;
-    auto ret = aclrtGetDevice(&user_dev_id);
-    if (ret != ACL_ERROR_NONE) {
-        LOG(ERROR) << "Failed to get device: " << ret;
-        return -1;
-    }
-    int32_t physical_dev_id;
-    ret = aclrtGetPhyDevIdByLogicDevId(user_dev_id, &physical_dev_id);
-    if (ret != ACL_ERROR_NONE) {
-        LOG(ERROR) << "Failed to get physical dev id: " << ret;
-        return -1;
-    }
-    aclrtPhysicalMemProp prop = {};
-    prop.handleType = ACL_MEM_HANDLE_TYPE_NONE;
-    prop.allocationType = ACL_MEM_ALLOCATION_TYPE_PINNED;
-    prop.memAttr = ACL_MEM_P2P_HUGE1G;
-    prop.location.type = ACL_MEM_LOCATION_TYPE_HOST_NUMA;
-    // Only 0 2 4 6 is available for fabric mem, map 4 device to one numa.
-    const int32_t kDevicesPerChip = 4;
-    const int32_t kNumaNodeStep = 2;
-    prop.location.id = (physical_dev_id / kDevicesPerChip) * kNumaNodeStep;
-    prop.reserve = 0;
-    LOG(INFO) << "Malloc host memory for numa:" << prop.location.id;
-    ret = aclrtMallocPhysical(&handle, total_size, &prop, 0);
-    if (ret != ACL_ERROR_NONE) {
-        LOG(INFO) << "Malloc host memory for numa:" << prop.location.id
-                  << " failed, try common allocate instead.";
-        prop.location.type = ACL_MEM_LOCATION_TYPE_HOST;
-        prop.location.id = 0;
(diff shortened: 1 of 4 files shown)


mooncake-transfer-engine/src/transport/ascend_transport/ascend_allocator.cpp
```diff
@@ -178,6 +216,60 @@ void *ascend_allocate_memory(size_t total_size, const std::string &protocol) {
...
+        void *ptr = allocate_fabric_exact(sz, /*quiet=*/true);
+        if (ptr != nullptr) {
+            *actual_size = sz;
+            if (sz < target_size) {
+                LOG(WARNING) << "Fabric mem best-effort: target=" << target_size
+                             << ", actual=" << sz << " (" << pct << "%)";
+            }
+            return ptr;
+        }
+    }
```
alogfans, line 264: The 1 GiB round-down can produce an allocation below the documented 50% minimum. For example, a 2.1 GiB target may accept 1 GiB; please calculate the minimum aligned size explicitly and skip candidates below it.

mooncake-transfer-engine/src/transport/ascend_transport/ascend_allocator.cpp
```diff
@@ -19,17 +43,22 @@ struct AllocRecord {
...
 std::unordered_map<void *, AllocRecord> g_vmm_alloc_records;
 
-int allocate_physical_memory(size_t total_size, aclrtDrvMemHandle &handle) {
+int allocate_physical_memory(size_t total_size, aclrtDrvMemHandle &handle,
+                             bool quiet) {
     int32_t user_dev_id;
     auto ret = aclrtGetDevice(&user_dev_id);
     if (ret != ACL_ERROR_NONE) {
-        LOG(ERROR) << "Failed to get device: " << ret;
+        if (!quiet) {
```
Aionw, line 51: Should quiet also gate `ERROR` level logging?

mooncake-transfer-engine/src/transport/ascend_transport/ascend_allocator.cpp
```diff
@@ -178,6 +216,60 @@ void *ascend_allocate_memory(size_t total_size, const std::string &protocol) {
...
+    if (!globalConfig().ascend_use_fabric_mem) {
+        void *ptr = ascend_allocate_memory(target_size, protocol);
+        if (ptr) {
+            *actual_size = target_size;
+        }
+        return ptr;
+    }
+
+#ifdef ASCEND_SUPPORT_FABRIC_MEM
+    (void)protocol;
```
Aionw, line 237: Why just remove `protocol` argument?

mooncake-transfer-engine/src/transport/ascend_transport/ascend_allocator.cpp
ascend-direct-dev: Good catch — fixed in 1c3e80b5.

We now compute an unaligned floor `min_size = target * 50%` and skip any 1G-aligned candidate with `sz < min_size`, so e.g. a 2.1 GiB target no longer accepts 1 GiB (~47.6%). Added `AlignDownBelowMinPercentIsRejected` to cover this.

mooncake-transfer-engine/src/transport/ascend_transport/ascend_allocator.cpp
ascend-direct-dev: Yes — intentionally. `quiet=true` is only used for intermediate percentile probes, so we suppress both INFO and ERROR there to avoid retry spam. The final failure still logs ERROR in `ascend_allocate_memory_best_effort` after the ladder is exhausted; exact-size `ascend_allocate_memory` keeps the original non-quiet ERROR path.

mooncake-transfer-engine/src/transport/ascend_transport/ascend_allocator.cpp
ascend-direct-dev: Kept for API symmetry with `ascend_allocate_memory` and for the non-fabric fallback above (`ascend_allocate_memory(target_size, protocol)`).

On the fabric_mem path we only call `allocate_fabric_exact(sz)`, which does not branch on protocol (`ascend` / `ubshmem` both use the same fabric allocator once `ascend_use_fabric_mem` is set), so `(void)protocol` is just to silence the unused-parameter warning.

alogfans (approved):
LGTM in TE

ascend-direct-dev:
@ykwd please review

Aionw (approved):
LGTM in store side.

ykwd (approved):

PR #3219, opened 2026-07-31
[TE] perf(efa): optionally register device memory only on topology-local NICs

### LINKED ISSUE #3217: [TE][EFA] GPU buffers are registered on all NICs, costing 7x vs per-GPU rail selection
## Summary

`EfaTransport` registers every single-chunk buffer on **all** EFA NICs. On p5.48xlarge (32 NICs, 8 GPUs) that is 8x more `fi_mr_reg` calls than NIXL's libfabric plugin issues for the same GPU buffer, and it costs **7.0x** in registration time. This is the largest single term in Mooncake's EFA registration cost — larger than either #3210 (concurrency cap, 2.6x) or #3216 (ascending sort, 3.1x worst case).

Filing this for discussion rather than as a PR, because reducing the NIC set is a **throughput trade-off**, not a free win, and the right policy needs bandwidth numbers that I have not measured yet.

## Current behavior

`EfaTransport::registerLocalMemoryInternal()` (`mooncake-transfer-engine/src/transport/efa_transport/efa_transport.cpp`):

```cpp
if (chunks.size() <= 1) {
    // Single chunk: all NICs
    for (size_t n = 0; n < num_nics; ++n) {
        nic_assignments[0].push_back(n);
    }
}
```

Every KV-cache buffer takes this branch, so each one is registered on all 32 domains. And since `use_parallel_reg` is 0 for device memory — the NIC-parallel path requires pre-touch, which is skipped for VRAM because a CPU store into a `cudaMalloc` pointer segfaults — those 32 registrations happen **serially** per buffer.

## What NIXL does

`nixlLibfabricRailManager::selectRailsForMemory()` (`src/utils/libfabric/libfabric_rail_manager.cpp:735`) branches on memory type:

- `VRAM_SEG`: queries the buffer's PCI bus ID via `cudaQueryAddr()`, then returns **only that GPU's topology-local rails** (`getEfaDevicesForPci()`, built from hwloc). On p5.48xlarge that is 4 of 32.
- `DRAM_SEG`: all rails.

So NIXL's fan-out for GPU memory is `NICs / GPUs`, and it falls back to all rails only when the PCI mapping lookup fails.

## Measurement

P5-1 (p5.48xlarge, 32 EFA NICs, 8x H100), 48 x 391 MB GPU buffers, one buffer per `batch_register_memory()` call, strictly serial. Only the engine's device list varies — no code change, the 4th argument of the Python `initialize()` is the device filter:

| NICs per buffer | total | 1st reg | 48th reg | avg |
|---|---|---|---|---|
| 32 (Mooncake today) | **123.7 s** | 128 ms | 5014 ms | 2578 ms |
| 4 (NIXL's VRAM choice) | **17.6 s** | 18 ms | 595 ms | 367 ms |
| 1 | 4.4 s | 6 ms | 150 ms | 91 ms |

7.0x from 32 → 4, and near-linear in the NIC count (123.7 / 4.4 = 28x for 32x the NICs).

It also *multiplies* a second effect. Device-memory registration on EFA costs roughly `k x (device bytes already registered on that domain)`, k ~= 260 ms/GiB (measured in #3216) — and that is **per libfabric domain**, so 32 domains pay it 32 times. At 4 NICs the same curve is still present (18 ms first vs 595 ms last) but every term is 8x cheaper. The fan-out and the accumulation are not independent problems; the fan-out amplifies the accumulation.

## Why this is not a straightforward fix

Registering a VRAM buffer on 4 of 32 NICs means only those 4 NICs can serve a transfer touching that buffer. That is a deliberate exchange of registration time for potential per-buffer bandwidth, and whether it is a good trade depends on:

- whether the transfer path already prefers topology-local NICs, in which case the other 28 registrations were mostly unused and this is close to free;
- how transfers striped across a buffer set are distributed across NICs at the deployed process count — this must be measured with all ranks running, not single-process;
- what happens on failover when a local NIC is down and the buffer is not registered anywhere else.

There is also a middle option: overlapping rather than disjoint windows (see the `sliding window MR` idea), giving each buffer more than `NICs/GPUs` coverage without going to all 32.

## Questions for maintainers

1. **Is registering GPU memory on all NICs intentional, or is it the incidental result of the `chunks.size() <= 1` branch?** My assumption had been that operators would narrow the device list by hand when they wanted this, but NIXL makes topology-local selection the default for VRAM, which suggests the opposite default is defensible.
2. **Should the selection be topology-derived (hwloc/PCI, like NIXL) or should it reuse the existing `Topology` machinery?** Mooncake already has `Topology::selectDevice()` with location-string affinity; the information needed may already be present.
3. **Disjoint (`NICs/GPUs`) or overlapping windows?** Disjoint matches NIXL and is simplest; overlapping trades some registration time back for redundancy and per-buffer bandwidth.
4. **Opt-in or default?** Given the throughput implications I would expect an initial landing behind a flag, with the default flipped once bandwidth numbers exist.

I am happy to do the bandwidth measurements and write the PR — mainly looking for agreement on the policy before implementing, since the choice among 2–4 above determines the shape of the change.

## Related

- #3210 — bound the batch MR registration fan-out (`MC_MAX_CONCURRENT_REG_MR`), 2.6x
- #3216 — register a batch smallest-buffer-first, 3.1x worst case; documents the per-domain accumulation this issue multiplies

Environment: p5.48xlarge (32 EFA NICs, 8x H100, 192 cores), `USE_EFA=ON USE_CUDA=ON`, libfabric with the `efa` provider.

whn09:
## Description

> **Bandwidth measured; ready for review.** Both halves are now on hardware: the registration-time win and the cross-machine `transfer_engine_bench` comparison showing `local` costs no throughput at the `--gpu_id=-1` shape. See "Cross-machine bandwidth" below.

> **#3210 is merged**, so this now applies directly to `main`; the diff is 227 lines. Closes #3217. (Originally stacked on #3216 as well; #3216 has since been closed — see "What happened to #3216" below.)

`EfaTransport::registerLocalMemoryInternal()` registers every single-chunk buffer on **every** NIC. That looks incidental to the `chunks.size() <= 1` branch rather than deliberate — the multi-chunk paths already partition NICs (disjoint or round-robin), only the common single-chunk case does not.

NIXL's libfabric plugin does the opposite: `selectRailsForMemory()` → `getEfaDevicesForPci()` registers a `VRAM_SEG` buffer only on the rails topologically local to that GPU. On p5.48xlarge that is 4 NICs versus 32.

### Why the fan-out costs more than 8x

Registering **device** memory on EFA costs roughly `k x (device bytes already registered on that libfabric domain)`, `k ~= 260 ms/GiB` on p5.48xlarge, flat across 0-19 GiB and nearly independent of the buffer's own size. The cost is paid **once per domain**, so the NIC count multiplies the accumulation rather than adding to it.

Measured on P5-1 (p5.48xlarge, 32 EFA NICs, 8x H100), 48 x 391 MB GPU buffers registered strictly serially — one buffer per `batch_register_memory()` call:

| NIC set | total | 1st reg | 48th reg | mean |
|---|---|---|---|---|
| `all` (32 NICs), GPU 0 | **116.1 s** | 122 ms | 4702 ms | 2418 ms |
| `local` (4 NICs), GPU 0 | **14.9 s** | 18 ms | 602 ms | 310 ms |
| `local` (4 NICs), GPU 5 | 15.1 s | 33 ms | 596 ms | 314 ms |
| `all` (32 NICs), GPU 5 | 116.9 s | 162 ms | 4766 ms | 2435 ms |

**7.8x.** The GPU 0 / GPU 5 rows are the check that this is really per-GPU rail selection and not "always GPU 0's four NICs": both GPUs land at the same time (14.9 vs 15.1 s), which they would not if one of them were using a remote rail group.

### The same thing at the deployed process count

The table above is single-process and strictly serial, which isolates the mechanism but is not how anything runs. Replaying a Kimi-K3 KV registration the way SGLang issues it — 8 ranks, one `TransferEngine` and one GPU per rank, 182 GPU buffers / 14.3 GiB each, ranks released from a barrier so all 8 hit the NICs together — in the buffer order SGLang actually passes, at `MC_MAX_CONCURRENT_REG_MR=16`, slowest rank:

| NIC set | wall | CPU | cores busy | peak threads |
|---|---|---|---|---|
| `all` (32) | 35.0 s | 2981 CPU-s | 85 | 168 |
| `local` (4) | **5.2 s** | **394 CPU-s** | 77 | 168 |

6.7x wall, 7.6x CPU.

### This does not replace `MC_MAX_CONCURRENT_REG_MR` (#3210)

Worth stating explicitly, because it is the natural assumption and it is wrong: `local` reduces the *work per registration*, not the *number of registrations in flight*. The thread count in `runBoundedParallel()` is the buffer count, which has nothing to do with the NIC set.

| sel | cap | wall | CPU | cores busy | peak threads |
|---|---|---|---|---|---|
| `all` | unset | 105.2 s | 19630 CPU-s | 187 | 1416 |
| `all` | 16 | 35.0 s | 2981 CPU-s | 85 | 168 |
| `local` | unset | 12.8 s | 2330 CPU-s | **182** | **1289** |
| `local` | 16 | 5.2 s | 394 CPU-s | 77 | 168 |

Unbounded, `local` still saturates the box: 1289 threads and 182 of 192 cores busy. On a real serving node those cores are needed for inference. The cap is what bounds concurrency; `local` is what bounds the work. Both are needed, and `local`'s wall-clock win from the cap (2.1-2.3x) is the same as `all`'s (2.3x), i.e. the two are orthogonal.

### What happened to #3216

#3216 sorted each batch smallest-first, on the theory that with `cost(i) = k x prior_bytes(i)` the total `k x sum over pairs (i<j) of size(i)` is minimized by non-decreasing sizes. Measuring it across this matrix showed no effect anywhere:

| sel | cap | descending | SGLang's order | ascending | spread |
|---|---|---|---|---|---|
| `local` | 1 | 23731 ms | 23705 ms | 23720 ms | **0.11%** |
| `all` | 1 | 180627 ms | 180864 ms | 180229 ms | **0.35%** |
| `local` | 2 | 13022 ms | 13059 ms | 12986 ms | 0.56% |
| `all` | 16 | 34025 ms | 33698 ms | 33718 ms | 0.9% |
| `local` | 16 | 5251 ms | 5428 ms | 5090 ms | 6.3% |
| `all` | unset | 78220 ms | 91396 ms | 103753 ms | *inverted* |

`cap=1` is the strictly serial case the model assumes, and the largest NIC count is where the effect should be biggest — `all`/`cap=1` shows 0.35%. Unbounded, the ordering is *reversed* from the prediction.

The reason the original 2.6x was wrong: that measurement called `batch_register_memory()` once **per buffer**, so `registrationOrder()` only ever saw a one-element list and sorted nothing. The 2.6x came from reordering the 48 calls in the harness, not from the code under test. With a real 182-buffer batch, `runBoundedParallel()` hands out indices via `next.fetch_add(1)` to `workers` threads, so `workers` registrations are always in flight; a sort fixes the *dispatch* order while the model needs *completion* to be serial. #3216 is closed.

### No new topology code

`Topology::discover()` already computes precisely what NIXL derives from hwloc. For a `cuda:N` entry, `preferred_hca` holds the HCAs at minimum PCI distance from that GPU within its NUMA node — on p5.48xlarge exactly the 4 EFA devices sharing the GPU's PCIe root complex. `buildLocalNicMap()` only inverts that mapping into `context_list_` indices. No PCI walk, no hwloc dependency, no new discovery path.

(`avail_hca` is not usable for this: it is "every HCA that is not preferred", i.e. the other 28.)

Partial registration needed no new plumbing either. `lkey`/`rkey` are already pushed as `0` for NICs a buffer was not registered on, and `EfaTransport::selectDevice()` already skips a NIC whose `rkey` is 0 and retries the next one.

### Host memory is deliberately excluded

`local` applies to device memory only, for three separate reasons:

1. **No accumulation to amplify.** Host registration is flat in prior bytes (#3216: 696 ms for the 1st of 48 x 391 MB, ~580 ms for the 48th), so narrowing the NIC set saves proportionally, not super-proportionally.
2. **It is ~4x cheaper to start with**, so the absolute win is small.
3. **A host buffer has no single owning device.** `cpu:N`'s `preferred_hca` is a whole NUMA node's NICs — measured as 16 of 32 on P5-1, i.e. half the machine. That is not a rail group, so "local" would not mean the same thing.

### Why opt-in rather than the default

It trades bandwidth for registration time: fewer NICs can serve a transfer touching that buffer, so a single-buffer transfer is capped at its rail group's bandwidth instead of the node's aggregate. With `--gpu_id=-1` (one buffer per GPU, the shape `transfer_engine_bench` and real KV caches use) all 32 NICs are still covered collectively, and each GPU's traffic stays on its own rails. That configuration measures as no loss at all (table below). The cap is real for a job that puts its whole working set behind one GPU's buffers, which is why this is opt-in rather than the default.

Secondly, `RdmaTransport::registerLocalMemoryInternal()` registers on every context unconditionally, in both its serial and `use_parallel_reg` paths, with no NIC-subset logic at all (`location` only enters metadata, for the *peer's* `selectDevice()` path selection). Defaulting `local` on would put the EFA transport out of step with the rest of the codebase; leaving it opt-in keeps the default behavior identical to today's for every transport.

I would not treat `all` as the permanent default. The measured win is large (registration ~4.5x, CPU ~7.6x) and the multi-GPU shape that real deployments use shows no bandwidth cost, so the conservative default here is mainly about not changing behavior under reviewers' feet in the same PR that introduces the mechanism. A follow-up could flip it, ideally as something smarter than an unconditional `local` — e.g. selecting per batch based on whether the buffers already span every rail, so the single-GPU case keeps the full node's bandwidth. Happy to take direction from reviewers on whether that belongs here instead.

## Module

- [ ] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [x] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
# New unit test: buildLocalNicMap() inverts preferred_hca correctly.
# Pure map inversion, needs no EFA hardware.
GTEST_FILTER='EFALocalNicMapTest.*' \
  ./build/mooncake-transfer-engine/tests/efa_transport_test

# Env-var parsing, including that a typo must not silently pick a policy.
GTEST_FILTER='EfaNicSelectionEnvTest.*' \
  ./build/mooncake-transfer-engine/tests/config_test

# Registration timing on hardware (p5.48xlarge), 48 x 391 MB GPU buffers,
# one buffer per batch_register_memory() call, MC_EFA_NIC_SELECTION=all|local,
# repeated on GPU 0 and GPU 5 to check per-GPU rail selection.
#
# 8-rank Kimi-K3 replay across {all,local} x {desc,SGLang,asc} x {unset,16,4,2,1},
# with /usr/bin/time -v and a thread/load sampler for the CPU columns.
```

**Test results:**
- [x] Unit tests pass

```
[  PASSED  ] 1 test.    # EFALocalNicMapTest.InvertsPreferredHca
[  PASSED  ] 14 tests.  # EfaNicSelectionEnvTest.* + MaxConcurrentRegMrEnvTest.*
```

`EfaNicSelectionEnvTest` covers the default (`all`), `local`, an explicit `all` overwriting a preset field, case-insensitivity, and that an unknown value or empty string leaves the field untouched rather than silently selecting a policy — registering buffers on the wrong NIC set is a correctness-adjacent surprise, not a tuning knob.

Manual testing is every table above, on P5-1 (p5.48xlarge, 32 EFA NICs, 8x H100): the serial per-GPU sweep, the 8-rank replay with CPU accounting, the cap x NIC-set matrix, and the ordering matrix that closed #3216.

### Cross-machine bandwidth: `all` vs `local`

Two p5.48xlarge in the same subnet (`172.31.0.0/20`), P5-1 -> P5-2, all 8 GPUs per side:

```bash
# target (P5-2)
MC_EFA_NIC_SELECTION=all|local ./transfer_engine_bench \
  --mode=target --protocol=efa --metadata_server=P2PHANDSHAKE \
  --local_server_name=172.31.0.239 --gpu_id=-1 --use_vram=true \
  --buffer_size=2147483648

# initiator (P5-1); segment port read from `ss -ltnp`, not from --local_server_name
MC_EFA_NIC_SELECTION=all|local ./transfer_engine_bench \
  --mode=initiator --protocol=efa --metadata_server=P2PHANDSHAKE \
  --local_server_name=172.31.15.67 --segment_id=172.31.0.239:<port> \
  --gpu_id=-1 --use_vram=true --buffer_size=2147483648 \
  --block_size=1048576 --operation=write --duration=10 \
  --threads=16 --batch_size=128 --report_unit=GB
```

| `MC_EFA_NIC_SELECTION` | write (GB/s) | read (GB/s) |
|---|---|---|
| `all` | 363.95, 373.68, 373.35, 373.32 | 370.53 |
| `local` | 363.96, 373.14, 363.81, 373.94, 373.38 | 373.88 |

**No throughput cost.** Best-of: 373.94 (`local`) vs 373.68 (`all`) GB/s for writes, a 0.07% difference. Both settings bounce between the same two clusters, ~364 and ~373 GB/s, and both clusters appear under both settings — so the spread is run-to-run variance, not an effect of the NIC set. 373.9 GB/s is 93% of the instance's 400 GB/s (3200 Gbps) line rate.

This is the expected result rather than a surprise: the NIC that actually carries a slice is chosen by `selectDevice()` from the *peer's* topology affinity for the destination buffer, which already prefers that buffer's own rails. `local` narrows "which NICs hold an MR for this buffer" down to the ones the data path was going to pick anyway; the MRs on the other 28 NICs were never used by these transfers. The target's registration wall time for the same 8 x 2 GiB, measured from the same pair of runs, was 21.77 s on `all` versus 4.85 s on `local`.

**What is not tested yet, and why:**

- **p6-b300 (16 NICs).** No capacity has been available. The mechanism is provider-side and should carry over; the magnitude will scale with the fan-out ratio (16/4 rather than 32/4).

## Checklist

Formatted with clang-format 20.1.8 (`--changed-lines` against `main`), matching the CI gate. Documented `MC_EFA_NIC_SELECTION` in `docs/source/design/transfer-engine/index.md` alongside the other EFA env vars. 226 LOC, so no RFC.

Checked for overlap: no open PR touches `registerLocalMemoryInternal()`. Rebased onto `main` at `eaa6c79d`, which already contains #3210 (it touches `registerLocalMemoryBatch()` in the same file); unit tests re-run green after the rebase.

## AI Assistance Disclosure

- [ ] No AI tools were used

Claude Code was used to run the registration-timing and CPU measurements, to trace how `Topology::discover()` and NIXL's `selectRailsForMemory()` arrive at the same rail set, to design the matrix that disproved #3216, and to draft the code, tests, and this description. Every line has been reviewed by me and I can defend the change end-to-end.


mooncake-transfer-engine/src/transport/efa_transport/efa_transport.cpp
```diff
@@ -314,8 +314,37 @@ int EfaTransport::registerLocalMemoryInternal(void* addr, size_t length,
...
-    if (chunks.size() <= 1) {
+    if (chunks.size() <= 1 && local_nics) {
+        // Single chunk, topology-local NIC selection requested for this device.
+        nic_assignments[0] = *local_nics;
+        if (globalConfig().trace) {
+            LOG(INFO) << "EFA local NIC selection: " << addr << " ("
+                      << resolved_name << ") on " << local_nics->size() << "/"
+                      << num_nics << " NICs";
+        }
+    } else if (chunks.size() <= 1) {
```
alogfans, line 347: Nonblocking comment: when chunks.size() > 1 we will continue registering all NICs, it's better to extend it by later PRs.

alogfans (approved):
LGTM

PR #3222, opened 2026-07-31
[Store] Add Fast-Fail for Impossible OffsetAllocator Requests

bitborne:
## Description

Cache a conservative upper-bound hint for the largest allocatable free region in `OffsetAllocator` and use it to reject impossible requests before acquiring the allocator mutex.

The main changes are:

- Initialize the atomic hint from allocator state during construction and deserialization.
- Check the requested size against the hint before entering the mutex-protected allocation path.
- Keep successful allocations off the hint-update path. Allocation can only reduce the largest free region, so the previous value remains a safe upper bound.
- Tighten the hint after a request passes the hint but fails in the mutex-protected allocator.
- Enable free-side hint maintenance only after a locked allocation failure has tightened the hint. While maintenance is active, have the internal allocator return the region produced by deallocation and neighbor merging so the hint can be raised incrementally without scanning allocator bins through `storageReport()` after every free. Maintenance is disabled again after the allocator becomes fully free.
- Collect the detailed failure metrics only when `VLOG(1)` is enabled, avoiding another `storageReport()` call in the locked failure path by default.
- Keep `OffsetBufferAllocator::getLargestFreeRegion()` authoritative for segment selection; the conservative atomic value remains an internal allocation filter only.
- Reconstruct both the hint and its maintenance state from authoritative allocator metadata after deserialization, while keeping the serialized format unchanged.
- Add a reusable multithreaded benchmark covering direct `OffsetAllocator` calls and the same-segment Put allocation stage, including capacity pressure, deterministic fragmentation, configurable mixed success/failure ratios, and a successful allocate/free control.

The mutex-protected allocator remains authoritative. A stale larger hint can only allow a request to fall through to the existing locked path; it cannot make an invalid allocation succeed. The first locked failure tightens the hint for later requests. `memory_order_relaxed` is sufficient because the atomic value is only a filtering hint and does not publish or protect allocator metadata.

This change is complementary to #2445 and #2797, which use the largest free region to rank segments but do not modify `OffsetAllocator` internals. It also avoids replacing the current mutex with a shared mutex, as discussed in the closed PR #2695.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [x] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**

```bash
cmake --build build \
  --target offset_allocator_test buffer_allocator_test \
           offset_allocator_concurrency_bench \
  -j8

./build/mooncake-store/tests/offset_allocator_test
./build/mooncake-store/tests/offset_allocator_test \
  --gtest_filter=OffsetAllocatorTest.LargestFreeRegionHintTightensOnFailureAndRaisesOnFree \
  --v=1
./build/mooncake-store/tests/buffer_allocator_test

for ratio in 0.05 0.30 0.60; do
  taskset -c 0-15 \
    ./build/mooncake-store/benchmarks/offset_allocator_concurrency_bench \
    --threads=16 \
    --iterations=300000 \
    --warmup_iterations=10000 \
    --layer=offset_allocator \
    --scenario=mixed \
    --mixed_failure_ratio="${ratio}"
done

taskset -c 0-15 \
  ./build/mooncake-store/benchmarks/offset_allocator_concurrency_bench \
  --threads=16 --iterations=300000 --warmup_iterations=10000 \
  --layer=offset_allocator --scenario=success

taskset -c 0-15 \
  ./build/mooncake-store/benchmarks/offset_allocator_concurrency_bench \
  --threads=16 --iterations=300000 --warmup_iterations=10000 \
  --layer=offset_allocator --scenario=capacity
```

The A/B benchmark used `RelWithDebInfo` builds of the PR base and candidate on the same host, pinned to CPUs 0-15. The execution order was rotated between rounds to reduce order and system-load bias. Throughput and per-operation latency were measured in separate phases, with 4.8 million operations in each phase per round. The successful and 5%-failure results are medians of five rounds; the other results are medians of three rounds.

The high-water workloads use a 1 GiB allocator at 93.75% actual utilization, with a 64 MiB largest free region. Successful operations allocate and immediately release 4 KiB. Expected-failure operations request 128 MiB and therefore cannot fit. For example, a 30% mixed workload contains approximately 30 expected failures and 70 successful allocate/free operations per 100 attempts.

The implementation evolved in three measured steps. Results from the earlier steps are retained below because they motivated the later changes; measurements from different steps were collected in separate runs and should be compared only within each table.

First, the lazy upper-bound policy was compared with the initial eager policy that refreshed the atomic value after every successful allocation. The binaries used the same benchmark and differed only in the hint-update policy.

| Scenario | Eager-refresh QPS | Lazy-hint QPS | QPS change | Eager p99 | Lazy p99 |
|---|---:|---:|---:|---:|---:|
| Successful allocate/free | 838.54 K | 920.17 K | +9.7% | 82,548 ns | 82,908 ns |
| 5% expected failures | 790.06 K | 960.61 K | +21.6% | 81,826 ns | 84,151 ns |
| 30% expected failures | 1.13 M | 1.16 M | +2.7% | 80,053 ns | 78,540 ns |
| 100% expected failures | 1.95 B | 2.33 B | +19.2% | 80 ns | 70 ns |

This established the lazy-hint policy: successful allocations do not update the atomic hint, and the first request that passes the conservative hint but fails under the mutex tightens it for subsequent requests.

Second, the initial lazy-hint implementation was compared with the upstream allocator. This exposed the remaining cost of refreshing the hint through `storageReport()` on every free, especially in success-heavy workloads.

| Scenario | Upstream QPS | Initial lazy-hint QPS | QPS change | Upstream p99 | Initial lazy-hint p99 |
|---|---:|---:|---:|---:|---:|
| Successful allocate/free | 1.03 M | 932.93 K | -9.1% | 82,626 ns | 82,515 ns |
| 5% expected failures | 1.01 M | 880.45 K | -13.1% | 78,939 ns | 85,181 ns |
| 30% expected failures | 1.15 M | 1.18 M | +3.1% | 69,932 ns | 79,429 ns |
| 100% expected failures | 10.73 M | 2.34 B | +217x | 25,238 ns | 71 ns |

The final implementation therefore avoids full-bin scans on the normal free path. Hint maintenance is activated only after a locked failure tightens the hint, and subsequent frees raise it from the newly merged region returned by the allocator. The following final comparison used rotated A/B execution order; the successful and 5%-failure results are medians of five rounds, while the other results are medians of three rounds.

| Scenario | Baseline QPS | Candidate QPS | QPS change | Baseline p99 | Candidate p99 |
|---|---:|---:|---:|---:|---:|
| Successful allocate/free | 902.52 K | 906.39 K | +0.4% | 82,437 ns | 87,667 ns |
| 5% expected failures | 925.96 K | 861.95 K | -6.9% | 81,195 ns | 85,522 ns |
| 30% expected failures | 1.04 M | 1.14 M | +9.2% | 74,381 ns | 82,447 ns |
| 60% expected failures | 1.52 M | 2.06 M | +35.1% | 62,198 ns | 75,563 ns |
| 100% expected failures | 13.34 M | 1.93 B | +145x | 29,927 ns | 80 ns |

The incremental free-side update removes the earlier failure-free throughput regression while preserving the impossible-request fast path. In this local run, the 5%-failure workload still shows a 6.9% throughput cost, while the candidate becomes throughput-positive between 5% and 30% expected failures. Mixed-workload p99 remains higher because its tail is dominated by the successful operations that still enter the mutex; the 100%-failure workload, which directly measures the intended fast path, reduces p99 by 99.7%. Relative QPS improvement is not expected to scale linearly with failure ratio because a locked failed allocation is cheaper than a successful allocate/free pair even in the baseline.

An independent rerun by a reviewer on the final lazy-hint/free-maintenance design showed the same overall trend, with smaller success-path overhead and positive throughput at 5% and 30% expected failures:

| Scenario | Reviewer upstream QPS | Reviewer candidate QPS | QPS change | Reviewer upstream p99 | Reviewer candidate p99 |
|---|---:|---:|---:|---:|---:|
| Successful allocate/free | 850.63 K | 824.05 K | -3.1% | 85,701 ns | 81,844 ns |
| 5% expected failures | 800.10 K | 884.97 K | +10.6% | 83,948 ns | 82,806 ns |
| 30% expected failures | 942.50 K | 1.19 M | +26.7% | 78,397 ns | 80,061 ns |
| 100% expected failures | 13.10 M | 2.31 B | +176x | 29,856 ns | 70 ns |

The 5% result is sensitive to host and run variance, so the PR does not claim a universal crossover point. Both final comparisons show that success-path overhead is small, while the benefit grows substantially as impossible-request frequency increases.

The benchmark also covers the same-segment Put allocation stage and deterministic fragmentation. All benchmark runs completed with zero unexpected results in both the throughput and latency phases.

User-space lock profiling was performed on an earlier fast-fail build using a `pthread_mutex_lock` uprobe and the `syscalls:sys_enter_futex` tracepoint. Across 3,232,000 measured and warmup allocation attempts, `pthread_mutex_lock` calls decreased from 3,232,057 to 57 and futex syscalls decreased from 1,197,395 to 22. The final policy retains the same repeated-request fast path after the first mutex-protected failure tightens the hint. QPS measured while perf probes were active was not used because probe overhead affects the two implementations asymmetrically.

Callgrind was also run on a focused stale-hint allocation-failure test repeated 5,000 times. Gating failure-only metrics behind `VLOG_IS_ON(1)` reduced the instruction count from 339,772,630 to 336,407,164, a 0.99% reduction, while the same test passes with verbose logging both disabled and enabled.

**Test results:**

- [x] Unit tests pass
- [x] Manual testing done (focused 16-thread A/B benchmark described above)

The complete `OffsetAllocatorTest` suite passed 51/51 tests, and `buffer_allocator_test` passed 18/18 tests. The tests cover hint tightening after failure, incremental updates after free and neighbor merging, metadata-capacity exhaustion, the verbose failure path, exact segment-selection queries, serialization/deserialization and restored-handle release, SmallFloat bin boundaries, multiplier values 1 through 4, and mutex-free rejection after the hint has been tightened.

## Checklist

- [x] I have run pre-commit on all modified files and all applicable hooks pass
- [x] For changes >500 LOC: N/A; the production implementation is below 500 LOC excluding tests and benchmark code and is not a major architectural change

## AI Assistance Disclosure

- [ ] No AI tools were used

Tool: OpenAI Codex

Used for code navigation, implementation, unit tests, benchmark design and automation, and review of implementation details and benchmark results. The submitter reviewed every changed line and is responsible for understanding and defending the changes.

### DIFF
--- a/mooncake-store/include/offset_allocator/offset_allocator.h
+++ b/mooncake-store/include/offset_allocator/offset_allocator.h
@@ -171,7 +171,10 @@ class OffsetAllocator : public std::enable_shared_from_this<OffsetAllocator> {
     // the request cannot be represented by this allocator.
     [[nodiscard]] uint64_t normalizedAllocationSize(size_t size) const;
 
-    // Returns a lock-free snapshot of the largest allocatable free region.
+    // Returns a mutex-free atomic upper-bound hint for the largest allocatable
+    // free region. The hint may be larger than the current value after a
+    // successful allocation and is tightened on a locked allocation failure
+    // or after free.
     [[nodiscard]] uint64_t getLargestFreeRegion() const noexcept {
         return m_largest_free_region.load(std::memory_order_relaxed);
     }
--- a/mooncake-store/include/offset_allocator/offset_allocator.h
+++ b/mooncake-store/include/offset_allocator/offset_allocator.h
@@ -173,8 +173,8 @@ class OffsetAllocator : public std::enable_shared_from_this<OffsetAllocator> {
 
     // Returns a mutex-free atomic upper-bound hint for the largest allocatable
     // free region. The hint may be larger than the current value after a
-    // successful allocation and is tightened on a locked allocation failure
-    // or after free.
+    // successful allocation, is tightened on a locked allocation failure, and
+    // is raised when free creates a larger region.
     [[nodiscard]] uint64_t getLargestFreeRegion() const noexcept {
         return m_largest_free_region.load(std::memory_order_relaxed);
     }
@@ -223,6 +223,7 @@ class OffsetAllocator : public std::enable_shared_from_this<OffsetAllocator> {
     uint64_t m_allocated_size GUARDED_BY(m_mutex) = 0;
     uint64_t m_allocated_num GUARDED_BY(m_mutex) = 0;
     std::atomic<uint64_t> m_largest_free_region{0};
+    bool m_largest_free_region_tightened GUARDED_BY(m_mutex) = false;
 
     // Private constructor - use create() factory method instead
     OffsetAllocator(uint64_t base, size_t size, uint32 init_capacity,
@@ -251,7 +252,9 @@ class __Allocator {
     void reset();
 
     OffsetAllocation allocate(uint32 size);
-    void free(OffsetAllocation allocation);
+    // Returns the size of the newly merged free region, or zero if allocator
+    // metadata capacity still prevents another allocation.
+    uint32 free(OffsetAllocation allocation);
 
     uint32 allocationSize(OffsetAllocation allocation) const;
     OffsetAllocStorageReport storageReport() const;
--- a/mooncake-store/include/offset_allocator/offset_allocator.h
+++ b/mooncake-store/include/offset_allocator/offset_allocator.h
@@ -2,6 +2,7 @@
 // (C) Sebastian Aaltonen 2023
 // MIT License (see file: LICENSE)
 
+#include <atomic>
 #include <memory>
 #include <optional>
 #include <vector>
@@ -170,6 +171,11 @@ class OffsetAllocator : public std::enable_shared_from_this<OffsetAllocator> {
     // the request cannot be represented by this allocator.
     [[nodiscard]] uint64_t normalizedAllocationSize(size_t size) const;
 
+    // Returns a lock-free snapshot of the largest allocatable free region.
+    [[nodiscard]] uint64_t getLargestFreeRegion() const noexcept {
+        return m_largest_free_region.load(std::memory_order_relaxed);
+    }
+
     // Get storage report (thread-safe)
     [[nodiscard]]
     OffsetAllocStorageReport storageReport() const;
@@ -200,6 +206,8 @@ class OffsetAllocator : public std::enable_shared_from_this<OffsetAllocator> {
     [[nodiscard]]
     OffsetAllocatorMetrics get_metrics_internal() const REQUIRES(m_mutex);
 
+    void refreshLargestFreeRegion() REQUIRES(m_mutex);
+
     std::unique_ptr<__Allocator> m_allocator GUARDED_BY(m_mutex);
     uint64_t m_base;
     // The real offset and size of the allocated memory need to be multiplied by
@@ -211,6 +219,7 @@ class OffsetAllocator : public std::enable_shared_from_this<OffsetAllocator> {
     // Lightweight metrics maintained during allocation/deallocation
     uint64_t m_allocated_size GUARDED_BY(m_mutex) = 0;
     uint64_t m_allocated_num GUARDED_BY(m_mutex) = 0;
+    std::atomic<uint64_t> m_largest_free_region{0};
 
     // Private constructor - use create() factory method instead
     OffsetAllocator(uint64_t base, size_t size, uint32 init_capacity,
@@ -330,6 +339,9 @@ OffsetAllocator::OffsetAllocator(T& serializer) {
                 "Deserializing OffsetAllocator failed: corrupt header");
         }
         m_allocator = std::make_unique<__Allocator>(serializer);
+        m_largest_free_region.store(
+            m_allocator->storageReport().largestFreeRegion << m_multiplier_bits,
+            std::memory_order_relaxed);
     } catch (const std::exception& e) {
         LOG(ERROR) << "Deserializing OffsetAllocator failed, error="
                    << e.what();
--- a/mooncake-store/include/offset_allocator/offset_allocator.h
+++ b/mooncake-store/include/offset_allocator/offset_allocator.h
@@ -345,9 +345,14 @@ OffsetAllocator::OffsetAllocator(T& serializer) {
                 "Deserializing OffsetAllocator failed: corrupt header");
         }
         m_allocator = std::make_unique<__Allocator>(serializer);
-        m_largest_free_region.store(
-            m_allocator->storageReport().largestFreeRegion << m_multiplier_bits,
-            std::memory_order_relaxed);
+        const uint64_t largest_free_region =
+            m_allocator->storageReport().largestFreeRegion << m_multiplier_bits;
+        m_largest_free_region.store(largest_free_region,
+                                    std::memory_order_relaxed);
+        const uint64_t allocator_capacity =
+            static_cast<uint64_t>(m_allocator->m_size) << m_multiplier_bits;
+        m_largest_free_region_tightened =
+            largest_free_region < allocator_capacity;
     } catch (const std::exception& e) {
         LOG(ERROR) << "Deserializing OffsetAllocator failed, error="
                    << e.what();
--- a/mooncake-store/src/offset_allocator.cpp
+++ b/mooncake-store/src/offset_allocator.cpp
@@ -602,6 +602,10 @@ std::optional<OffsetAllocationHandle> OffsetAllocator::allocate(size_t size) {
 
     OffsetAllocation allocation = m_allocator->allocate(fake_size);
     if (allocation.isNoSpace()) {
+        // A request can pass the conservative hint but still fail because
+        // the allocator state changed concurrently. Tighten the hint after
+        // observing the authoritative state under the mutex.
+        refreshLargestFreeRegion();
         // Log metrics to help understand why allocation failed
         // Note: We're already holding m_mutex, so use internal method
         OffsetAllocatorMetrics metrics = get_metrics_internal();
@@ -609,8 +613,6 @@ std::optional<OffsetAllocationHandle> OffsetAllocator::allocate(size_t size) {
                 << ", fake_size=" << fake_size << ", " << metrics;
         return std::nullopt;
     }
-
-    refreshLargestFreeRegion();
 
     // Update lightweight metrics
     m_allocated_size += size;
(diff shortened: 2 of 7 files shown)


ykwd:
Thanks a lot for working on this! This is a very thoughtful and carefully executed optimization. The lazy-hint design is particularly nice, as it avoids updating the hint on every successful allocation while still preserving the fast-fail behavior for impossible requests.

The performance evaluation is also exceptionally thorough. The profiling makes the trade-offs very clear and provides strong evidence for the design.

One surprising result is the overhead in success-heavy workloads. Compared with upstream, the successful allocate/free case is still about 9%~13% slower, which seems fairly significant for adding a hint load before acquiring the mutex. It may be worth investigating where the additional cost is coming from.

mooncake-store/src/offset_allocator.cpp
```diff
@@ -669,11 +688,20 @@ OffsetAllocatorMetrics OffsetAllocator::get_metrics() const {
...
+                    : 0;
+    m_largest_free_region.store(largest_free_region, std::memory_order_relaxed);
+}
+
 void OffsetAllocator::freeAllocation(const OffsetAllocation& allocation,
                                      uint64_t size) {
     MutexLocker lock(&m_mutex);
     if (m_allocator) {
         m_allocator->free(allocation);
+        refreshLargestFreeRegion();
```
Aionw: I benchmarked this PR locally and reproduced the same performance regression described in the PR.

After isolating the individual changes, I found that the unconditional `refreshLargestFreeRegion()` in the free path is the main source of the regression. Although `storageReport()` itself is relatively lightweight, calling it on every free extends the mutex-protected critical section and significantly amplifies lock contention under concurrent, failure-free workloads.

To address this, I changed `m_allocator->free()` to return the size of the resulting merged free region. `freeAllocation()` can then use that value to incrementally update `m_largest_free_region`, avoiding an additional `storageReport()` call in the hot free path.

This remains safe because freeing an allocation can only introduce one new candidate for the largest free region: the region produced by merging the freed allocation with its adjacent free regions. The cached value is only a conservative upper bound used for fast-fail decisions, while an allocation failure still refreshes it from the authoritative allocator state.

I also gated the failure-path metrics collection behind `VLOG_IS_ON(1)`, avoiding a second `storageReport()` call when verbose logging is disabled.

In my local measurements, these changes removed most of the failure-free workload regression while preserving the fast-fail improvement for impossible allocation requests. Callgrind showed that the instruction overhead relative to the base revision dropped from approximately 4.8% to 0.9%.

mooncake-store/include/offset_allocator/offset_allocator.h
```diff
@@ -330,6 +345,9 @@ OffsetAllocator::OffsetAllocator(T& serializer) {
                 "Deserializing OffsetAllocator failed: corrupt header");
         }
         m_allocator = std::make_unique<__Allocator>(serializer);
+        m_largest_free_region.store(
```
Aionw: **[P1] Restore hint-maintenance state after deserialization**

A deserialized allocator may already contain live allocations, but this path restores only `m_largest_free_region` and leaves `m_largest_free_region_tightened` at its default value of `false`. When a restored allocation is subsequently freed, `freeAllocation()` skips the upward hint update, so the cached hint can become smaller than the authoritative largest free region and incorrectly reject a valid allocation before acquiring the mutex.

I reproduced this with a 1,024-byte allocator containing one 512-byte allocation: after serialization/deserialization and releasing the restored allocation, `storageReport().largestFreeRegion` is 1,024 while `getLargestFreeRegion()` remains 512, and `allocate(1024)` incorrectly returns no space.

Please initialize the hint-maintenance state for restored allocators in both this constructor and the preconstructed-allocator path used by MsgPack deserialization. A regression test covering partial allocation → serialize/deserialize → release restored handle → allocate the full capacity would catch this.

Aionw (commented):
Found one correctness issue in the restored allocator hint state; details are included inline.

Aionw:
I benchmark latest commit locally, performance looks acceptable. But there is a correctness issue need to be fixed.

| Scenario | Upstream QPS | Latest QPS | QPS change | Upstream p99 | Latest p99 |
|---|---:|---:|---:|---:|---:|
| Successful allocate/free | 850.63 K | 824.05 K | -3.1% | 85,701 ns | 81,844 ns |
| 5% expected failures | 800.10 K | 884.97 K | +10.6% | 83,948 ns | 82,806 ns |
| 30% expected failures | 942.50 K | 1.19 M | +26.7% | 78,397 ns | 80,061 ns |
| 100% expected failures | 13.10 M | 2.31 B | +176x | 29,856 ns | 70 ns |

mooncake-store/src/offset_allocator.cpp
bitborne: Implemented in `ee686ef0`.

`__Allocator::free()` now returns the size of the resulting merged free region, and `freeAllocation()` uses it to raise the hint incrementally without calling `storageReport()` on every free. Failure metrics are also collected only when `VLOG_IS_ON(1)` is enabled.

I additionally kept free-time atomic maintenance inactive until a mutex-protected allocation failure has tightened the hint, moved request normalization before the mutex, and moved handle construction outside the critical section.

The final concurrent benchmark no longer reproduces the original 9%–13% success-heavy regression. The five-round result was approximately throughput-neutral in the successful control and improved QPS by 10.3% at 5% expected failures.

mooncake-store/include/offset_allocator/offset_allocator.h
bitborne: Fixed in `52e20117`.

The hint-maintenance state is now derived from the authoritative allocator state in both the regular deserialization constructor and the preconstructed-allocator recovery path. A restored fragmented allocator therefore continues to raise the hint when restored allocations are released, without adding the transient state to the serialized format.

The regression test follows the reported scenario directly: a 1,024-byte allocator contains one 512-byte allocation, is serialized and deserialized, releases the restored handle, and then successfully allocates the full 1,024-byte capacity.

The complete `OffsetAllocatorTest` suite passes 51/51 tests, and `BufferAllocatorTest` passes 18/18 tests.

bitborne:
Thanks for the careful review and independent benchmark results! The restore correctness issue has been fixed in 52e20117 with a dedicated regression test. I also updated the PR description to document the full optimization process and retain both the earlier and final performance results. This was very helpful.

Aionw (approved):
LGTM

ykwd (approved):

PR #3226, opened 2026-08-01
[Store] Add parallel bucket writes for SSD offload

NUABO:
## Description
Add optional thread pool that dispatches per-bucket BatchOffload calls concurrently within a single heartbeat tick.  Each bucket's D2H staging, eviction, WriteBucket (disk I/O), and master RPC run independently — buckets write to different files so the heavy disk writes are truly parallel.

- MC_OFFLOAD_WRITE_THREADS (default 0, disabled)
    • 0:  original sequential per-bucket loop, zero change.
    • >1: thread pool of that size; when multiple buckets are
           returned by AllocateOffloadingBuckets, each bucket is
           dispatched to a worker.  The pool is created once in
           FileStorage::Init() and reused across heartbeat ticks.

- Parallel-path error handling mirrors the sequential path:
    • KEYS_ULTRA_LIMIT or other non-INVALID_READ errors are
      captured and propagated after all futures complete.
    • INVALID_READ errors add the affected keys to failed_tasks
      (mutex-protected) so the master can release refcounts.
    • Keys skipped by GroupOffloadingKeysByBucket are also
      reported as failed.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
```

**Test results:**
- [ ] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

### DIFF
--- a/mooncake-store/src/file_storage.cpp
+++ b/mooncake-store/src/file_storage.cpp
@@ -702,7 +702,8 @@ tl::expected<void, ErrorCode> FileStorage::OffloadObjects(
     // left waiting on the TTL reaper.
     std::optional<ErrorCode> abort_error;
 
-    // Pre-populate all_bucket_keys (needed by both sequential and parallel paths).
+    // Pre-populate all_bucket_keys (needed by both sequential and parallel
+    // paths).
     for (const auto& keys : buckets_keys) {
         for (const auto& k : keys) all_bucket_keys.insert(k);
     }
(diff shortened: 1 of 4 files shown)


mooncake-store/src/file_storage.cpp
LujhCoconut: **This issue is introduced by the PR.** The original sequential path only has one thread executing `OffloadObjects` — no concurrent accessors, no UAF possible.

The new code introduces a dangerous reference-capture + barrier pattern at two locations:

**Capture site (:534-538):** `&task_by_storage_key`, `&failed_tasks`, `&failed_mutex`, `&fatal_error`, `&fatal_mutex`, `&complete_handler` — all references to `OffloadObjects` stack locals.

**Barrier site (:679):**
```cpp
for (auto& f : futures) f.get();
```

**Real trigger path:** `PinnedBufferPool::Buffer::Pageable` → `std::make_unique<char[]>(size)` (staging buffer up to 256MB, `pinned_buffer_pool.h:43`, called by `AllocNew` on pool miss). Under memory pressure with multiple workers, this throws `std::bad_alloc`. `packaged_task` stores it, `future.get()` rethrows at the barrier. `OffloadObjects` unwinds, destroying all captured stack variables. Meanwhile, other workers are still running in the ThreadPool with live references to those destroyed variables → UAF.

Unlike `.at()` calls (which have `find` guards), this OOM path is genuinely reachable under concurrent pinned-memory pressure.

**Note:** Even with the fix below, the exception will still escape `OffloadObjects` → `Heartbeat()` → heartbeat thread lambda (no try-catch at any level, `file_storage.cpp:394-403`) → `std::terminate`. The fix eliminates the UAF (undefined behavior), but the process still dies. To truly survive, a try-catch is also needed at the `Heartbeat()` call site.

**Fix — drain-then-rethrow at the barrier:**

```cpp
std::exception_ptr first_exception;
for (auto& f : futures) {
    try {
        f.get();
    } catch (...) {
        if (!first_exception) first_exception = std::current_exception();
    }
}
if (first_exception) std::rethrow_exception(first_exception);
```

And to survive the exception, wrap `Heartbeat()` in a try-catch in the heartbeat thread loop.

LujhCoconut:
The biggest maintainability issue: the parallel path's worker lambda body (first occurrence of the two-line anchor above) and the sequential path's `for`-loop body (second occurrence) are line-by-line copies — ~120 lines of effectively identical logic.

Only three semantic differences exist across the entire duplicated region (see table above). Outside it, the parallel path only adds: (a) lambda capture / `packaged_task` wrapping before the duplicate, (b) `future`/`enqueue` dispatch after. Inside, the rest is character-identical.

Any future offload logic change must be applied in both places — easy to miss one and introduce behavioral divergence between the two paths.

Suggestion: extract a private `ProcessOneBucket()` method (signature above). The parallel path: dispatch → call helper with `failed_mutex != nullptr` → aggregate errors under lock. The sequential path: `for` loop → call helper with `failed_mutex == nullptr` → aggregate directly.

mooncake-store/src/file_storage.cpp
```diff
@@ -508,8 +516,199 @@ tl::expected<void, ErrorCode> FileStorage::OffloadObjects(
...
+                        }
+                    }
+                    if (obj_success) {
+                        host_batch_object[obj_key] = std::move(host_slices);
+                    }
+                }
+
+                auto offload_start = std::chrono::steady_clock::now();
+                auto bucket_complete_handler =
+                    [this, offload_start, &complete_handler](
```
LujhCoconut: The parallel path captures `complete_handler` by reference at two levels (`&complete_handler` in both the worker lambda and the nested `bucket_complete_handler`), while the sequential path captures it by value (no `&`). Both are safe today since the barrier ensures `OffloadObjects`'s stack outlives all workers, but the inconsistency is a minor code smell.

This is **not** a security fix — value-capturing `complete_handler` alone does not eliminate the dangling-reference risk from other captures like `&task_by_storage_key`. The real protection is Comment 1's drain-then-rethrow. This is simply a consistency cleanup: change both capture sites in the parallel path to value-capture, matching the sequential path.

mooncake-store/src/file_storage.cpp
LujhCoconut: The nested `bucket_complete_handler` in the parallel path also captures `complete_handler` by reference, while the identical lambda in the sequential path captures it by value. Remove the `&` to match. Consistency cleanup only — handle together with  R620.

Issue #3229, opened 2026-08-02
[RFC] RL Training Weight Offload and Restore Paths for Mooncake Store


zxpdemonio:
# [RFC]: RL Training Weight Offload and Restore Paths for Mooncake Store

## Changes proposed

This RFC proposes a staged design for using Mooncake Store as a high-performance offload/restore backend for GPU-resident model weights during RL training.

The primary scenario is an iterative RL training loop:

1. The trainer or rollout worker owns updated model weights in GPU HBM.
2. The worker offloads those GPU tensors into Mooncake Store when the weights need to leave HBM.
3. Mooncake keeps the weight object in Store segment memory, with hard-pin/upsert semantics provided by the weight-storage work.
4. A later trainer or rollout step restores the Store object directly into GPU tensors.
5. In the common production deployment, the Python process is a DummyClient and the resource-owning process is a standalone RealClient on the same node.

The goal is to make this same-node GPU <-> Store memory path bandwidth-optimal while preserving the existing cross-node RDMA behavior.

This RFC is a follow-up to Mooncake issue #1621, "[Call for Contribution] Enhance Model Weight Storage for Mooncake Store". That RFC identifies model weights as a first-class Mooncake Store workload and calls out the semantic/API work needed for that workload, especially hard pin support, native upsert, and documentation/examples for RL and model-management systems.

This proposal focuses on the next layer: the high-performance tensor data path for weight offload and restore. It is also complementary to the SGLang/LMSYS P2P weight-transfer work, which uses Mooncake TransferEngine for cluster-wide weight distribution. In short:

- Issue #1621 defines the Store semantics needed by model weights.
- The P2P weight-transfer path optimizes cross-node weight distribution.
- This RFC optimizes Store-backed same-node GPU <-> CPU memory movement for RL weight offload/restore.

Together they form a complete in-memory weight lifecycle: fast distribution across workers, first-class Store object semantics, and fast local offload/restore when weights must be retained outside HBM.

## Motivation

Mooncake Store already has the right high-level building blocks:

- RealClient-owned Store segments for distributed memory storage.
- DummyClient + RealClient deployment mode for applications that should not own Store segments directly.
- Tensor APIs for Python callers.
- Pinned Store segment memory support for faster GPU <-> CPU DMA.
- TransferEngine/RDMA for cross-node transfer.

[...]

PR #3234, opened 2026-08-02
[Store] support CUDA IPC for dummy client buffers

zxpdemonio:
## Description

This PR adds a same-node CUDA IPC handoff path for dummy-client CUDA buffers.
It is part of the RL training weight offload/restore path discussed in RFC #3229.

When a dummy client receives CUDA-backed tensor or object buffers, it exports a CUDA IPC memory handle to the real client instead of routing the payload through dummy shared-memory staging. The real client opens the handle for the duration of the operation and reuses the existing Store read/write helpers with the mapped CUDA pointer.

Covered paths:
- tensor write: `put_tensor`, `batch_put_tensor`, and tensor parallel write helpers when used by a dummy client
- tensor read into CUDA tensors: `get_tensor_into_cuda`, `batch_get_tensor_into_cuda`
- generic object pointer APIs: `put_from`, `batch_put_from`, `get_into`, `batch_get_into` with CUDA pointers

Current Mooncake capability check:
- Checked `kvcache/main` at `eaa6c79d`.
- Store / integration / wheel do not have CUDA IPC handoff support for dummy-client tensor/object buffers (`CudaIpcBuffer`, `get_tensor_into_cuda`, `batch_get_tensor_into_cuda`, `put_from_cuda_ipc`, and `batch_put_from_cuda_ipc` all have 0 matches on those paths).
- Transfer Engine already has CUDA IPC usage in transport implementations such as NVLink/intranode transports, but that is a transport-layer memory registration/peer-mapping facility. It does not provide the Store dummy-client API path added here, where a dummy process hands a CUDA tensor/object buffer to the real client without SHM staging.

Compatibility notes:
- Existing Python APIs keep their signatures and return conventions.
- Existing `PyClient` base virtual methods are unchanged; CUDA IPC helpers are scoped to `DummyClient` internals.
- Non-CUDA buffers keep the existing dummy SHM paths.
- Existing `put_tensor` behavior for regular and non-contiguous tensors is preserved; the CUDA IPC path is used only when the dummy client can export the payload as a CUDA IPC buffer, otherwise it falls back to the original path.

Safety details:
- CUDA IPC handles carry allocation-base offset and payload size for PyTorch suballocations.
- The real client validates the imported range against the CUDA allocation before using the mapped pointer.
- CUDA IPC mappings are RAII-managed and stay alive until the underlying Store operation returns.
- Non-CUDA buffers keep the existing dummy SHM paths.
- The CUDA Driver API symbol used for allocation-range lookup is resolved at runtime, so regular Store build/link targets do not need a direct `libcuda` link dependency.

Related RFC: https://github.com/kvcache-ai/Mooncake/issues/3229

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [x] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
./scripts/code_format.sh
cmake --build build-cuda-ipc-standalone -j32 --target mooncake_client store

# Manual CUDA dummy-client integration test with a fresh master + real client:
# - put_tensor(cuda) + get_tensor_into_cuda
# - put_from(cuda_ptr) + get_into(cuda_ptr)
# - batch_put_from(cuda_ptrs) + batch_get_into(cuda_ptrs)
```

**Test results:**
- [ ] Unit tests pass

Compatibility validation result:
```text
setup_dummy 0
bytes_get 49152 True
tensor_get (128, 8) torch.float32 True
noncontig_get (128, 8) True
cuda_object_get_into 4194304 True
compatibility and cuda ipc verify ok
```

Manual CUDA dummy-client validation result:
```text
setup_dummy 0
tensor_get_into_cuda 4194304 True
put_from_cuda 0 get_size 4194304
get_into_cuda 4194304 True
batch_put_from_cuda [0, 0]
batch_get_into_cuda [1048576, 1048576] [True, True]
cuda ipc pr verify ok
```

Performance validation:
- Environment: same-node dummy client + real client, `MC_STORE_MEMCPY=1`, `MC_STORE_PIN_MEMORY_MAX_BYTES=4294967296`, torch `2.9.1+cu128`.
- Workload: CUDA `float16` tensors, `put_tensor` / `batch_put_tensor`, median of 8 measured iterations after 2 warmups.
- Baseline: current dummy-client path without this CUDA IPC handoff.

| API | Batch size | Size | Baseline GB/s | PR GB/s | Speedup | Baseline median | PR median |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| `put_tensor(cuda)` | 1 | 64 MiB | 3.079 | 21.213 | 6.89x | 20.302 ms | 2.946 ms |
| `put_tensor(cuda)` | 1 | 256 MiB | 3.072 | 23.601 | 7.68x | 81.379 ms | 10.593 ms |
| `put_tensor(cuda)` | 1 | 512 MiB | 3.086 | 24.022 | 7.78x | 162.015 ms | 20.815 ms |
| `batch_put_tensor(cuda)` | 2 | 64 MiB | 3.049 | 21.972 | 7.21x | 41.002 ms | 5.689 ms |
| `batch_put_tensor(cuda)` | 2 | 256 MiB | 3.083 | 23.786 | 7.72x | 162.181 ms | 21.021 ms |
| `batch_put_tensor(cuda)` | 2 | 512 MiB | 3.097 | 24.124 | 7.79x | 322.905 ms | 41.452 ms |

## Checklist

RFC issue: https://github.com/kvcache-ai/Mooncake/issues/3229

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex was used to help implement, review, and validate the change. The submitter is responsible for the final code and PR.

### DIFF
--- a/mooncake-integration/store/store_py.cpp
+++ b/mooncake-integration/store/store_py.cpp
@@ -5,6 +5,7 @@
 #include <functional>
 #include <limits>
 #include <numeric>
+#include <optional>
 #include <unordered_map>
 #include <unordered_set>
 
@@ -15,6 +16,7 @@
 #include "memory_alloc.h"
 #include "ssd_register_client.h"
 #include "device/accelerator_registry.h"
+#include "device/cuda_ipc_buffer.h"
 
 #include <cstdlib>  // for atexit
 #include <memory>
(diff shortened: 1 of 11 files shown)


Aionw (commented):
Could we avoid the parallel-vector API altogether instead of adding size checks at every layer?

The current CUDA IPC path passes logically related fields as separate vectors:

- `keys`
- `buffers` / `payloads`
- `metadata_buffers`
- `metadata_sizes`
- `src_offsets`
- `sizes`

This makes invalid states representable and requires every caller and RPC handler to manually verify that all vectors have the same length. It also creates a risk that a future change pushes an element into one vector but not the others.

I suggest representing each operation as a single request item:

```cpp
struct CudaIpcWriteRequest {
    std::string key;
    std::optional<ShmBufferRef> metadata;
    CudaIpcBufferHandle payload;
};

struct CudaIpcReadRequest {
    std::string key;
    CudaIpcBufferHandle destination;
    size_t source_offset;
    size_t size;
};
```

The RPC methods would then accept `std::vector<CudaIpcWriteRequest>` or `std::vector<CudaIpcReadRequest>`. This removes the cross-vector length checks and makes it impossible to associate a key with the wrong handle, offset, or size.

The new Python API can follow the same model:

```python
store.batch_get_tensor_into_cuda([
    ("key-1", tensor1),
    ("key-2", tensor2),
])
```

The binding can convert each `(key, tensor)` pair directly into one internal request item. Per-item validation, such as CUDA handle validity and allocation bounds, should remain, but repeated vector-length validation would no longer be necessary.

For existing Python APIs that still accept parallel lists, we can validate once at the Python boundary and immediately normalize them into a request-item vector. All subsequent C++ and RPC layers should operate on the structured representation.

zxpdemonio:
> Could we avoid the parallel-vector API altogether instead of adding size checks at every layer?
> 
> The current CUDA IPC path passes logically related fields as separate vectors:
> 
> * `keys`
> * `buffers` / `payloads`
> * `metadata_buffers`
> * `metadata_sizes`
> * `src_offsets`
> * `sizes`
> 
> This makes invalid states representable and requires every caller and RPC handler to manually verify that all vectors have the same length. It also creates a risk that a future change pushes an element into one vector but not the others.
> 
> I suggest representing each operation as a single request item:
> 
> ```c++
> struct CudaIpcWriteRequest {
>     std::string key;
>     std::optional<ShmBufferRef> metadata;
>     CudaIpcBufferHandle payload;
> };
> 
> struct CudaIpcReadRequest {
>     std::string key;
>     CudaIpcBufferHandle destination;
>     size_t source_offset;
>     size_t size;
> };
> ```
> 
> The RPC methods would then accept `std::vector<CudaIpcWriteRequest>` or `std::vector<CudaIpcReadRequest>`. This removes the cross-vector length checks and makes it impossible to associate a key with the wrong handle, offset, or size.
> 
> The new Python API can follow the same model:
> 
> ```python
> store.batch_get_tensor_into_cuda([
>     ("key-1", tensor1),
>     ("key-2", tensor2),
> ])
> ```
> 
> The binding can convert each `(key, tensor)` pair directly into one internal request item. Per-item validation, such as CUDA handle validity and allocation bounds, should remain, but repeated vector-length validation would no longer be necessary.
> 
> For existing Python APIs that still accept parallel lists, we can validate once at the Python boundary and immediately normalize them into a request-item vector. All subsequent C++ and RPC layers should operate on the structured representation.

@Aionw Thanks for the suggestion. I agree that the parallel-vector form made the CUDA IPC path too easy to misuse.

I updated the implementation to normalize the inputs into structured request items at the Python/DummyClient boundary:
- `CudaIpcWriteRequest`
- `CudaIpcReadRequest`
- `CudaIpcShmBufferRef`

The DummyClient -> RealClient RPC helpers now take `std::vector<...Request>` instead of separate `keys` / `buffers` / `metadata` / `sizes` vectors. Existing Python APIs keep the same list-style interface for compatibility, but the C++/RPC layer no longer carries the invalid intermediate states you pointed out.

I also verified the updated path with dummy + real client CUDA IPC coverage for:
- `put_tensor(cuda)` + `get_tensor_into_cuda`
- `put_from(cuda_ptr)` + `get_into(cuda_ptr)`
- `batch_put_from(cuda_ptrs)` + `batch_get_into(cuda_ptrs)`

Build, format, and `git diff --check` all pass.

stmatengss (approved):

PR #3236, opened 2026-08-02
[Bugfix] Drain RDMA async event queue on each epoll wakeup

SongOf:
Description
  
  The device async fd is registered edge-triggered (RdmaContext::joinNonblockingPollList), but WorkerPool::doProcessContextEvents() consumed only one event per wakeup — ibv_get_async_event() returns exactly one record per read() (kernel: ib_uverbs_event_read() dequeues a single list entry and returns). Anything still queued therefore got no further EPOLLIN and waited for an unrelated later event to release it, leaving the handler permanently N events behind.

  Bursts are routine, not exceptional: IBV_EVENT_COMM_EST fires once per connection, so batch connection setup alone builds a backlog. Port and device errors queued behind it — including the IBV_EVENT_PORT_ERR that drives               set_active(false) + refreshPublishedLocalTopology() — are delayed by the whole backlog, and stall entirely if no further event arrives.

  Fix: read until the fd reports EAGAIN. EINTR retries instead of aborting the drain (unreachable on this non-blocking fd, but aborting there would reproduce the same stranding). Other errno values still abort and propagate, so the loop cannot spin on a broken fd. A wakeup with nothing pending now returns success instead of ERR_CONTEXT.

  Adds rdma_async_event_drain_test, which drives the real doProcessContextEvents() against a scripted event queue installed by wrapping ibv_get_async_event / ibv_ack_async_event. No RDMA device required, so it runs on ordinary CI runners.

  No existing issue or open PR covers this. #3161 ([TransferEngine] Refactor RDMA worker ownership, WIP) touches the same file but addresses endpoint ownership/UAF.

  Module

  - [x] Transfer Engine (mooncake-transfer-engine)
  - [ ] Mooncake Store (mooncake-store)
  - [ ] Mooncake EP (mooncake-ep)
  - [ ] Mooncake PG (mooncake-pg)
  - [ ] Integration (mooncake-integration)
  - [ ] P2P Store (mooncake-p2p-store)
  - [ ] Python Wheel (mooncake-wheel)
  - [ ] Common (mooncake-common)
  - [ ] Mooncake RL (mooncake-rl)
  - [ ] CI/CD
  - [ ] Docs
  - [ ] Other
  
  Type of Change

  - [x] Bug fix
  - [ ] New feature
  - [ ] Refactor
  - [ ] Breaking change
  - [ ] Documentation update
  - [ ] Performance improvement
  - [ ] Other

  How Has This Been Tested?

[...]

PR #3243, opened 2026-08-03
[Bugfix][TE] Fix dma-buf offsets for chunked GPU MRs

RuiqingFeng:
## Description

Large GPU buffers are split into multiple memory regions (MRs) when exceeding `max_mr_size`. Previously, every chunk reused the dma-buf offset exported for the base buffer address. As a result, later chunks registered or mapped the wrong GPU pages.

This change fixes the issue by adding each chunk's relative displacement (`chunk_addr - addr`) to its `chunk_dmabuf_exp.offset` in both serial and parallel registration paths inside `RdmaTransport::registerLocalMemoryInternal`. It also adds a GPU loopback regression test (`RDMAGpuDmabufChunkTest.LaterChunkUsesItsOwnDmabufOffset`) that transfers data targeting a later chunk.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
MC_MAX_MR_SIZE=67108864 \
WITH_NVIDIA_PEERMEM=0 \
./build/mooncake-transfer-engine/tests/rdma_large_mr_test \
  --gtest_filter=RDMAGpuDmabufChunkTest.LaterChunkUsesItsOwnDmabufOffset

### DIFF
--- a/mooncake-transfer-engine/src/transport/rdma_transport/rdma_transport.cpp
+++ b/mooncake-transfer-engine/src/transport/rdma_transport/rdma_transport.cpp
@@ -394,6 +394,11 @@ int RdmaTransport::registerLocalMemoryInternal(void *addr, size_t length,
     for (size_t ci = 0; ci < chunks.size(); ++ci) {
         void *chunk_addr = chunks[ci].first;
         size_t chunk_len = chunks[ci].second;
+        const uint64_t chunk_offset = reinterpret_cast<uintptr_t>(chunk_addr) -
+                                      reinterpret_cast<uintptr_t>(addr);
+        DmabufExport chunk_dmabuf_exp = dmabuf_exp;
+        if (chunk_dmabuf_exp.method == DmabufExport::Method::kDmabufReg)
+            chunk_dmabuf_exp.offset += chunk_offset;
 
         if (do_pre_touch) {
             // Parallel pre-touch the memory to speed up registration.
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by staryxchen — APPROVED** [review]
Good catch! LGTM.

File: mooncake-transfer-engine/src/transport/rdma_transport/rdma_transport.cpp [hunk 1]
```diff
@@ -429,10 +434,10 @@ int RdmaTransport::registerLocalMemoryInternal(void *addr, size_t length,
             const int ar = access_rights;  // Local copy for lambda capture
 
             for (size_t i = 0; i < context_list_.size(); ++i) {
-                reg_threads.emplace_back([this, &ret_codes, &dmabuf_exp, i,
-                                          chunk_addr, chunk_len, ar]() {
+                reg_threads.emplace_back([this, &ret_codes, &chunk_dmabuf_exp,
```
alogfans: chunk_dmabuf_exp is captured by ref. Is it right?

RuiqingFeng: Good catch! Updated chunk_dmabuf_exp to capture-by-value in the latest commit. Thanks!

alogfans:
Could you please recheck the reason not passing CI? It seems some code cannot work well.

RuiqingFeng:
> * Build & Test (Linux) / build (3.10) (pull_request)

This test (rdma_async_event_drain_test) is a brand-new regression test recently introduced in PR #3236 (012af11c). Since it invokes bindToSocket() during SetUp() for the first time, it triggers libnuma.so's internal static allocation (numa_node_to_cpus), which was not suppressed in PR #3236, causing LSan to exit with status 8 on process teardown

RuiqingFeng:
> Could you please recheck the reason not passing CI? It seems some code cannot work well.

Hi， @alogfans, the CI failure on _rdma_async_event_drain_test_ (introduced in PR #3236) was caused by a LeakSanitizer false positive on **libnuma.so**'s internal static topology cache when calling `bindToSocket()`, which overrode the process exit code to `8` even though all test assertions passed (`[ PASSED ] 5 tests.`). 
I have added an LSan suppression (`leak:libnuma.so`) specifically in rdma_async_event_drain_test.cpp to ignore this system library allocation without touching production code. The commit has been pushed and all CI checks should now pass smoothly. Thanks!

PR #3252, opened 2026-08-03
[TransferEngine] Share one host KV segment across a TP group

ascend-direct-dev:
## Motivation

MLA keeps the same KV cache in every rank of a tensor-parallel group. When those ranks offload KV to host memory during decode, each one allocates its own copy, so host usage grows with the TP size while every copy holds identical bytes.

This PR adds a *shared segment*: the owner rank allocates the pages once through the platform's virtual memory manager, and every other rank maps the very same physical memory. Host usage for offloaded MLA KV drops to 1/TP.

## Design

C++ shares one contiguous host span. Tensor layout (which offset holds which layer's K/V) is computed in Python and folded into the segment name so peers that disagree still fail the fingerprint check.

**Two-phase creation.** The handle exchange belongs to the caller's process group:

1. `SharedSegment::Create` reserves address space (owner also allocates and exports) and returns a fixed-size blob.
2. The caller `all_gather`s the blob over its own group.
3. `segment->Complete(blobs)` maps the owner's pages and verifies that every rank declared the same thing.

Doing it this way avoids calling back into Python collectives from C++ under the GIL. Non-owner blobs are the same length as the owner's, so a plain fixed-size all-gather carries them.

**Addresses are not forced to match.** Ranks agree on the byte layout, not on virtual addresses, so each rank reserves `size` rather than `size * world_size`.

**Backends.** Compiled in per platform and both optional:

| Platform | Mechanism | Handle |
| --- | --- | --- |
| Ascend | ACL VMM via `adxl::AdxlEngine` | `aclrtMemFabricHandle` |
| NVIDIA | CUDA VMM (`cuMemCreate`/`cuMemMap`) | `CUmemFabricHandle`, CUDA 12.4+ |

`SharedSegment::Supported()` reports whether the running system can actually share memory. Builds without either backend return `kNotImplemented` rather than failing to link.

## Python API

```python
from mooncake.shared_segment import create_shared_segment

seg = create_shared_segment(
    "vllm_sparse_kv",
    blocks={
        "k": dict(count=num_layers, shape=k_shape, dtype=torch.bfloat16),
        "v": dict(count=num_layers, shape=v_shape, dtype=torch.bfloat16),
    },
    world_size=tp_size,
    rank_id=tp_rank,
    tp_group=tp_group,
)
k_caches_cpu = seg.tensors("k")  # zero-copy views onto the shared pages
```

## Test plan

- [x] `shared_segment_test`: fingerprint, blob encode/decode, and the full two-phase protocol against a fake backend.
- [x] Ascend build and pybind smoke test on a CANN toolchain.
- [x] CUDA backend compiles clean under `-Wall -Wextra` against the CUDA driver headers.
- [ ] Multi-rank run on real Ascend hardware, paired with HiXL https://gitcode.com/cann/hixl/merge_requests/851

### DIFF
--- a/mooncake-integration/shared_segment.py
+++ b/mooncake-integration/shared_segment.py
@@ -1,0 +1,246 @@
+"""Process-shared host memory for one-writer/many-reader KV offload.
+
+MLA keeps the same KV in every rank of a TP group, so offloading it per rank
+multiplies host memory by the TP size for no benefit. A shared segment allocates
+the pages once, in ``owner_rank``, and maps them into every rank of the group.
+
+Virtual addresses are deliberately not forced to match across ranks: peers agree
+on the byte layout, so the reserved address space equals the segment size rather
+than size x world_size. Layout (which offset holds which tensor) is computed
+here; C++ only shares the raw span.
+
+    from mooncake.shared_segment import create_shared_segment
+
+    seg = create_shared_segment(
+        "vllm_sparse_kv",
+        blocks={
+            "k": dict(count=num_layers, shape=k_shape, dtype=torch.bfloat16),
+            "v": dict(count=num_layers, shape=v_shape, dtype=torch.bfloat16),
+        },
+        world_size=tp_size,
+        rank_id=tp_rank,
+        tp_group=tp_group,
+        mmap=True,
+    )
+    k_caches_cpu = seg.tensors("k")
+"""
+
+from __future__ import annotations
+
+import ctypes
+from dataclasses import dataclass
+from typing import Any, Dict, List, Mapping, Optional, Sequence, Tuple
+
+import torch
+import torch.distributed as dist
+
+from .engine import SharedSegment as _CppSharedSegment
+
+__all__ = [
+    "SharedSegment",
+    "SharedSegmentError",
+    "create_shared_segment",
+    "shared_segment_supported",
+]
+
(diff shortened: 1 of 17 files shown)


mooncake-integration/shared_segment.py
```diff
@@ -0,0 +1,242 @@
...
+        raise SharedSegmentError(
+            "This mooncake build has no mmap shared-segment backend"
+            if mmap
+            else "This mooncake build has no VMM backend for shared segments"
+        )
+    # Rank order decides the byte layout, so it must not depend on dict
+    # iteration order differing between ranks.
+    names = list(block_order) if block_order is not None else sorted(blocks)
+    if set(names) != set(blocks):
+        raise SharedSegmentError("block_order must list exactly the declared blocks")
```
alogfans: shared_segment_supported() or _CppSharedSegment.create() may fail before all_gather. If the owner fails to allocate memory or one peer fails to reserve its address range, the remaining ranks will block at line 238.

All ranks should exchange their local creation status before proceeding. If any rank fails, every rank should release its local resources and return the same failure. Import results should also be synchronized after Complete().

ascend-direct-dev:
Superseded by https://github.com/kvcache-ai/Mooncake/pull/3285 (same branch tip `221e6ec9`; this PR could not be reopened after a force-push).

PR #3244, opened 2026-08-03
[Store] Harden Put/Upsert completion against promotion lifecycle

Lin-z-w:
## Description

Fixes #3203.

`PutEnd` / `UpsertEnd` previously trusted only `metadata.client_id` and could complete any matching staged replica. Promotion also stages a `PROCESSING` replica under the object's metadata, but does not register a primary write in `processing_keys`. An invalid or out-of-order primary End could therefore complete a promotion-owned replica. If that replica later became stale, cleanup could remove it while leaving the associated `PromotionTask`, quota reservation, source refcount, in-flight count, and queued client work behind.

This change:

- requires an active primary Start marker before `PutEnd`, `UpsertEnd`, or `PutRevoke`, and only finalizes/revokes matching `PROCESSING` replicas;
- makes primary writes and promotion mutually exclusive at Upsert admission, promotion admission/retry, and promotion allocation;
- cancels promotion state when stale-handle cleanup removes the task's allocated replica, including the durable HA finalization path;
- captures removed replica IDs during the existing cleanup pass, avoiding an additional replica scan.

The validation reuses existing transient `processing_keys` and `promotion_tasks` state. It adds no persistent object metadata, RPC fields, snapshot fields, or oplog fields.

PR #3162 includes only the `UpsertStart` promotion-task gate as part of broader quota work. This standalone follow-up is based on current `main` and completes the End ownership and stale-cleanup fixes without including the quota changes.

## Module

- [ ] Transfer Engine (`mooncake-transfer-engine`)
- [x] Mooncake Store (`mooncake-store`)
- [ ] Mooncake EP (`mooncake-ep`)
- [ ] Mooncake PG (`mooncake-pg`)
- [ ] Integration (`mooncake-integration`)
- [ ] P2P Store (`mooncake-p2p-store`)
- [ ] Python Wheel (`mooncake-wheel`)
- [ ] Common (`mooncake-common`)
- [ ] Mooncake RL (`mooncake-rl`)
- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
cmake --build /tmp/mooncake-issue-3203-build --target promotion_on_hit_test master_service_test -j 16
/tmp/mooncake-issue-3203-build/mooncake-store/tests/promotion_on_hit_test
[...]

PR #3253, opened 2026-08-03
[Store] Fix: preempt in-progress offload task on UpsertStart 

huniu20:
## Description

`UpsertStart` currently rejects with `OBJECT_HAS_REPLICATION_TASK` when the
target key has an entry in `offloading_tasks`. Callers have no way to recover:
the offload will not drain until the store worker finishes writing to SSD, and
short-retry clients (e.g. `MooncakeStoreClient.batch_upsert_from`, 3 tries at
1s intervals) give up long before that under load, hanging on the key.

Preempt the offload instead. The new upsert data supersedes what the task
queued, so keeping the offload alive would only write stale bytes. Cancel it
inline the same way `EraseMetadata` does: `dec_refcnt` on the source replica,
erase `offloading_tasks[key]`, and drop the mirror `offloading_objects` entry
on the local disk segment. This mirrors the existing in-progress `Put`
preemption a few lines below. `Copy`/`Move` and promotion continue to reject
because they have external observers on the source; offload has none.

**Test results:**
- New `UpsertPreemptsInProgressOffload` / `BatchUpsertPreemptsInProgressOffload`
  fail on `main` with `-708` (`OBJECT_HAS_REPLICATION_TASK`) and pass after
  the fix.
- Full `offload_on_evict_test` (12 tests) and full `master_service_test`
  (147 tests) pass with no regressions. 

## Related Work

This is the offloading-task counterpart to the `UpsertStart` admission-gate
work in #3244 and #3162 (@Lin-z-w), which addresses the promotion-task gate.
The three gates (`replication_tasks`, `offloading_tasks`, `promotion_tasks`)
are independent; this PR only changes the `offloading_tasks` case and keeps
the other two unchanged.

### DIFF
--- a/mooncake-store/src/file_storage.cpp
+++ b/mooncake-store/src/file_storage.cpp
@@ -573,19 +573,18 @@ tl::expected<void, ErrorCode> FileStorage::OffloadObjects(
                 client_->ValidateOffloadGenerations(validate_tasks);
             if (!validation) {
                 // Master-side gate still applies in NotifyOffloadSuccess.
-                LOG(WARNING)
-                    << "[OFFLOAD] ValidateOffloadGenerations failed, "
-                       "skipping pre-SSD-IO check: "
-                    << validation.error();
+                LOG(WARNING) << "[OFFLOAD] ValidateOffloadGenerations failed, "
+                                "skipping pre-SSD-IO check: "
+                             << validation.error();
             } else {
                 const auto& results = validation.value();
                 for (size_t i = 0; i < validate_keys.size(); ++i) {
                     if (i >= results.size() || !results[i]) {
                         const auto& obj_key = validate_keys[i];
                         const auto& task = task_by_storage_key.at(obj_key);
-                        LOG(INFO) << "[OFFLOAD] drop stale key, key="
-                                  << task.key
-                                  << ", generation=" << task.generation;
+                        LOG(INFO)
+                            << "[OFFLOAD] drop stale key, key=" << task.key
+                            << ", generation=" << task.generation;
                         failed_tasks.push_back(task);
                         batch_object.erase(obj_key);
                     }
--- a/mooncake-store/src/file_storage.cpp
+++ b/mooncake-store/src/file_storage.cpp
@@ -504,12 +504,8 @@ tl::expected<void, ErrorCode> FileStorage::OffloadObjects(
             return result.error();
         }
         // Master returns a per-task acceptance vector: 1 = accepted, 0 =
-        // rejected as stale (post-SSD-IO generation mismatch on the NACK,
-        // success, or orphan-fallback path). Locally roll back on-disk
-        // state only for the rejected keys, leaving accepted siblings in
-        // the same bucket intact — otherwise a partial rejection would
-        // leave orphan bucket entries whose replicas Master no longer
-        // references, which ScanMeta could later re-register after restart.
+        // rejected as stale. Roll back rejected keys only, so accepted
+        // siblings in the same bucket are not clobbered.
         const auto& accepted = result.value();
         if (accepted.size() == keys.size()) {
             std::vector<std::string> rejected_keys;
@@ -588,13 +584,9 @@ tl::expected<void, ErrorCode> FileStorage::OffloadObjects(
             continue;
         }
 
-        // Pre-SSD-IO generation check. UpsertStart may have preempted an
-        // offload after its mirror was drained: the master marker is gone
-        // and OffloadTaskItem.generation no longer matches. Drop stale
-        // keys locally: master already released source refcount and erased
-        // the marker in UpsertStart, so a NACK from us would race the new
-        // generation's marker (see the NACK generation-guard in
-        // MasterService::NotifyOffloadSuccess for defence-in-depth).
+        // Pre-SSD-IO generation check: skip keys UpsertStart has already
+        // preempted so we don't emit a NACK that would race the new
+        // generation's marker. NotifyOffloadSuccess re-checks defensively.
         {
             std::vector<OffloadTaskItem> validate_tasks;
             std::vector<std::string> validate_keys;
--- a/mooncake-store/src/file_storage.cpp
+++ b/mooncake-store/src/file_storage.cpp
@@ -555,6 +555,47 @@ tl::expected<void, ErrorCode> FileStorage::OffloadObjects(
             continue;
         }
 
+        // Pre-SSD-IO generation check. UpsertStart may have preempted an
+        // offload after its mirror was drained: the master marker is gone
+        // and OffloadTaskItem.generation no longer matches. Route stale
+        // keys through failed_tasks so the NACK flush releases the source
+        // refcount and no SSD bucket is written.
+        {
+            std::vector<OffloadTaskItem> validate_tasks;
+            std::vector<std::string> validate_keys;
+            validate_tasks.reserve(batch_object.size());
+            validate_keys.reserve(batch_object.size());
+            for (const auto& [obj_key, _] : batch_object) {
+                validate_tasks.push_back(task_by_storage_key.at(obj_key));
+                validate_keys.push_back(obj_key);
+            }
+            auto validation =
+                client_->ValidateOffloadGenerations(validate_tasks);
+            if (!validation) {
+                // Master-side gate still applies in NotifyOffloadSuccess.
+                LOG(WARNING)
+                    << "[OFFLOAD] ValidateOffloadGenerations failed, "
+                       "skipping pre-SSD-IO check: "
+                    << validation.error();
+            } else {
+                const auto& results = validation.value();
+                for (size_t i = 0; i < validate_keys.size(); ++i) {
+                    if (i >= results.size() || !results[i]) {
+                        const auto& obj_key = validate_keys[i];
+                        const auto& task = task_by_storage_key.at(obj_key);
+                        LOG(INFO) << "[OFFLOAD] drop stale key, key="
+                                  << task.key
+                                  << ", generation=" << task.generation;
+                        failed_tasks.push_back(task);
+                        batch_object.erase(obj_key);
+                    }
+                }
+            }
+            if (batch_object.empty()) {
+                continue;
+            }
+        }
+
         // D2H staging: replace device slices with host memory slices
         // so that storage_backend (ConcatSlicesToString / BuildBucket /
         // WriteBucket) always receives host pointers.
--- a/mooncake-store/src/file_storage.cpp
+++ b/mooncake-store/src/file_storage.cpp
@@ -503,6 +503,39 @@ tl::expected<void, ErrorCode> FileStorage::OffloadObjects(
                        << result.error() << " keys count: " << keys.size();
             return result.error();
         }
+        // Master returns a per-task acceptance vector: 1 = accepted, 0 =
+        // rejected as stale (post-SSD-IO generation mismatch on the NACK,
+        // success, or orphan-fallback path). Locally roll back on-disk
+        // state only for the rejected keys, leaving accepted siblings in
+        // the same bucket intact — otherwise a partial rejection would
+        // leave orphan bucket entries whose replicas Master no longer
+        // references, which ScanMeta could later re-register after restart.
+        const auto& accepted = result.value();
+        if (accepted.size() == keys.size()) {
+            std::vector<std::string> rejected_keys;
+            rejected_keys.reserve(keys.size());
+            for (size_t i = 0; i < keys.size(); ++i) {
+                if (!accepted[i]) {
+                    rejected_keys.push_back(keys[i]);
+                }
+            }
+            if (!rejected_keys.empty()) {
+                LOG(INFO) << "[OFFLOAD] master rejected "
+                          << rejected_keys.size() << " of " << keys.size()
+                          << " keys as stale; running partial rollback";
+                if (auto bucket_backend =
+                        std::dynamic_pointer_cast<BucketStorageBackend>(
+                            storage_backend_)) {
+                    bucket_backend->PartialRollbackKeys(rejected_keys);
+                }
+            }
+        } else {
+            // Shape mismatch is a protocol violation; log loudly but treat
+            // the batch as accepted so we don't clobber committed state.
+            LOG(ERROR) << "[OFFLOAD] NotifyOffloadSuccess returned "
+                       << accepted.size() << " flags for " << keys.size()
+                       << " keys; skipping partial-rollback bookkeeping";
+        }
         return ErrorCode::OK;
     };
 
@@ -557,9 +590,11 @@ tl::expected<void, ErrorCode> FileStorage::OffloadObjects(
 
         // Pre-SSD-IO generation check. UpsertStart may have preempted an
         // offload after its mirror was drained: the master marker is gone
-        // and OffloadTaskItem.generation no longer matches. Route stale
-        // keys through failed_tasks so the NACK flush releases the source
-        // refcount and no SSD bucket is written.
+        // and OffloadTaskItem.generation no longer matches. Drop stale
+        // keys locally: master already released source refcount and erased
+        // the marker in UpsertStart, so a NACK from us would race the new
+        // generation's marker (see the NACK generation-guard in
+        // MasterService::NotifyOffloadSuccess for defence-in-depth).
         {
             std::vector<OffloadTaskItem> validate_tasks;
             std::vector<std::string> validate_keys;
@@ -583,9 +618,8 @@ tl::expected<void, ErrorCode> FileStorage::OffloadObjects(
                         const auto& obj_key = validate_keys[i];
                         const auto& task = task_by_storage_key.at(obj_key);
                         LOG(INFO)
-                            << "[OFFLOAD] drop stale key, key=" << task.key
-                            << ", generation=" << task.generation;
-                        failed_tasks.push_back(task);
+                            << "[OFFLOAD] drop stale key locally, key="
+                            << task.key << ", generation=" << task.generation;
                         batch_object.erase(obj_key);
                     }
                 }
--- a/mooncake-store/src/master_service.cpp
+++ b/mooncake-store/src/master_service.cpp
@@ -4280,24 +4280,28 @@ auto MasterService::PutEnd(const UUID& client_id, const ObjectMeta& object_meta,
             return replica.is_dfs_replica() && replica.is_processing();
         })) {
         auto& tenant_state = accessor.GetTenantState();
-        bool task_created = false;
+        // One marker covers every mirror pushed below, so the mirrors are
+        // collected before the marker is recorded.
+        std::optional<ReplicaID> source_id;
+        std::vector<UUID> mirror_clients;
         metadata.VisitReplicas(
             [](const Replica& replica) {
                 return replica.is_completed() && replica.is_memory_replica();
             },
-            [this, &object_id, &tenant_state, &task_created](Replica& replica) {
-                auto result = PushOffloadingQueue(object_id, replica);
-                if (result) {
-                    if (!task_created) {
-                        replica.inc_refcnt();
-                        tenant_state.offloading_tasks.emplace(
-                            object_id.user_key,
-                            OffloadingTask{replica.id(),
-                                           std::chrono::system_clock::now()});
-                        task_created = true;
-                    }
+            [this, &object_id, &source_id, &mirror_clients](Replica& replica) {
+                auto result =
+                    PushOffloadingQueue(object_id, replica, &mirror_clients);
+                if (result && !source_id.has_value()) {
+                    replica.inc_refcnt();
+                    source_id = replica.id();
                 }
             });
+        if (source_id.has_value()) {
+            tenant_state.offloading_tasks.emplace(
+                object_id.user_key,
+                OffloadingTask{*source_id, std::chrono::system_clock::now(),
+                               std::move(mirror_clients)});
+        }
     }
 
     // If the object is completed, remove it from the processing set.
@@ -4767,9 +4771,12 @@ auto MasterService::UpsertStart(const UUID& client_id, const std::string& key,
                         ErrorCode::OBJECT_HAS_REPLICATION_TASK);
                 }
 
-                // Reject if an offload-to-disk task is in progress (same
-                // reason).
-                if (tenant_state.offloading_tasks.count(key) > 0) {
+                // Cancel a still-queued offload so the upsert can take the key
+                // over. Once a store worker owns the task it is reading the
+                // source buffer for its SSD write, so the upsert waits for
+                // NotifyOffloadSuccess to clear the marker instead.
+                if (!CancelQueuedOffloadTask(tenant_state, metadata,
+                                             object_id)) {
                     LOG(INFO) << "key=" << key
                               << ", error=object_has_offloading_task";
                     return tl::make_unexpected(
@@ -7051,7 +7058,8 @@ auto MasterService::NotifyOffloadSuccess(
 }
 
 tl::expected<void, ErrorCode> MasterService::PushOffloadingQueue(
-    const ObjectIdentity& object_id, Replica& replica) {
+    const ObjectIdentity& object_id, Replica& replica,
+    std::vector<UUID>* mirror_clients) {
     const auto& segment_names = replica.get_segment_names();
     if (segment_names.empty()) {
         return {};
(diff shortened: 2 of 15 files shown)


mooncake-store/src/master_service.cpp
```diff
@@ -4265,13 +4265,33 @@ auto MasterService::UpsertStart(const UUID& client_id, const std::string& key,
...
-                              << ", error=object_has_offloading_task";
-                    return tl::make_unexpected(
-                        ErrorCode::OBJECT_HAS_REPLICATION_TASK);
+                              << ", action=upsert_preempts_offloading_task";
+                    auto source =
+                        metadata.GetReplicaByID(offload_it->second.source_id);
+                    if (source != nullptr) {
+                        source->dec_refcnt();
+                    }
+                    tenant_state.offloading_tasks.erase(offload_it);
```
Lin-z-w: Do not preempt an offload that has already been handed to the worker.

`OffloadObjectHeartbeat()` returns the queued task and clears it from `offloading_objects` before `FileStorage` reads the source bytes or writes them to SSD. The corresponding `offloading_tasks[key]` entry remains until `NotifyOffloadSuccess()`.

Therefore, if the heartbeat wins the race, this block does not actually cancel the offload: removing the mirror is a no-op, while `dec_refcnt()` and erasing the task marker allow the Upsert to proceed even though the worker is still using the old task. For a same-size Upsert, the source buffer is reused in place, so the worker may persist stale or partially overwritten data. When its late `NotifyOffloadSuccess()` arrives, the task marker is gone and the completion falls through to `AddReplica()`, potentially attaching that stale/corrupted SSD replica to the new object generation.

The new tests only cover cancellation before the heartbeat drains the queue. Please distinguish queued tasks from in-flight tasks and reject/cancel them safely, or carry a generation/task ID through `OffloadTaskItem` and validate it in `NotifyOffloadSuccess()`. A regression test should cover:

`Put V1 -> heartbeat drains task -> Upsert V2 -> V1 NotifyOffloadSuccess -> old SSD replica must not be added to V2`.

mooncake-store/src/master_service.cpp
```diff
@@ -4265,13 +4265,46 @@ auto MasterService::UpsertStart(const UUID& client_id, const std::string& key,
...
+                            segment_manager_.getLocalDiskSegmentAccess();
+                        for (auto& [_, segment] :
+                             ssd_access.getClientLocalDiskSegment()) {
+                            MutexLocker locker(&segment->offloading_mutex_);
+                            if (segment->offloading_objects.erase(
+                                    scoped_key) > 0) {
+                                queued = true;
+                            }
+                        }
+                    }
```
Lin-z-w: Do not treat “any queued mirror” as proof that no offload is in flight.

`PutEnd()` calls `PushOffloadingQueue()` for every completed MEMORY replica, and `PushOffloadingQueue()` may insert the same scoped key into multiple `LocalDiskSegment::offloading_objects` maps. A single `offloading_tasks[key]` entry can therefore correspond to multiple mirrors.

If worker A has already drained its mirror while worker B's mirror is still queued, this loop erases B's entry and sets `queued = true`. The code then decrements the source refcount, erases the task marker, and allows the Upsert even though worker A is still reading/writing the old object. Its late `NotifyOffloadSuccess()` can still attach stale or corrupted SSD data to the new generation. A heartbeat can also drain a later segment while this loop is traversing the segments.

Please require proof that all destinations are still queued and cancel them atomically, or explicitly track per-destination/in-flight state (or a task generation) instead of inferring the whole task state from whether any mirror was erased. A regression test should queue the key on two local-disk segments, drain only one, and verify that `UpsertStart()` is rejected.

mooncake-store/tests/offload_on_evict_test.cpp
```diff
@@ -424,6 +424,128 @@ TEST_F(OffloadOnEvictTest, BatchRemoveDropsOffloadingObjectsMirror) {
...
+
+    constexpr size_t seg_size = 1024 * 1024 * 16;
+    auto ctx =
+        PrepareSegment(*service, "test_segment", kDefaultSegmentBase, seg_size);
+    auto mount_ld = service->MountLocalDiskSegment(ctx.client_id, true);
+    ASSERT_TRUE(mount_ld.has_value());
+
+    const std::string key = "upsert_over_offload";
+
+    // ComboA: PutEnd populates offloading_tasks[key] immediately.
```
yokinoshitayoki: Could we also add regression coverage with `config.offload_on_evict = true`?

mooncake-store/src/master_service.cpp
```diff
@@ -4265,13 +4265,46 @@ auto MasterService::UpsertStart(const UUID& client_id, const std::string& key,
...
+                    bool queued = false;
+                    {
+                        ScopedLocalDiskSegmentAccess ssd_access =
+                            segment_manager_.getLocalDiskSegmentAccess();
+                        for (auto& [_, segment] :
+                             ssd_access.getClientLocalDiskSegment()) {
+                            MutexLocker locker(&segment->offloading_mutex_);
+                            if (segment->offloading_objects.erase(
+                                    scoped_key) > 0) {
+                                queued = true;
```
yokinoshitayoki: Can this check account for the same key being queued on more than one `LocalDiskSegment`? `PutEnd` calls `PushOffloadingQueue` for every completed MEMORY replica, so one worker may already have drained its mirror while another segment still contains the same scoped key. In that mixed state, this loop erases the remaining mirror and sets `queued = true`, causing the master task to be preempted even though the first worker is already IN-FLIGHT. That recreates the stale completion/refcount race this change is intended to avoid.

Could we track the state per offload destination/task, or otherwise require proof that no copy has already been drained before allowing preemption? A regression test with two MEMORY replicas where only one worker heartbeat has drained the task should cover this case.

huniu20:
**Update: PR reworked to address the concerns from @Lin-z-w and @yokinoshitayoki.**

**Concerns addressed:**
- In-flight offload preempted unsafely, letting the worker persist stale/partially-overwritten data and a late `NotifyOffloadSuccess` attach a stale SSD replica to the new generation.
- "Any mirror still queued" is not proof the task is not in flight: one key can have mirrors on multiple `LocalDiskSegment`s, and a heartbeat may have already drained one of them.
- `offload_on_evict = true` path was uncovered.

**Fix:**
- `OffloadTaskItem` carries a monotonic `generation`. `UpsertStart` preempts on the `offloading_tasks[key]` marker directly — queued or in-flight — erases mirrors across **every** LocalDisk segment atomically, and moves the old memory replicas into `discarded_replicas_` so the worker keeps reading a soon-to-be-freed buffer while the upsert allocates a fresh one (no in-place reuse).
- `ValidateOffloadGenerations` gates SSD IO before the bucket is written; `NotifyOffloadSuccess` drops any completion whose generation no longer matches (defence in depth).

**Regression tests:**
- `UpsertPreemptsInFlightOffloadAndDropsStaleNotify` — `Put V1 → drain → Upsert V2 → V1 late notify` must not add a stale SSD replica.
- `UpsertPreemptsOffloadAcrossMultipleSegments` — key queued on two LocalDisk segments, only one drained; `UpsertStart` still preempts safely. The assertion on the second segment's mirror runs immediately after `UpsertStart` returns (before the new `PutEnd`), because `PutEnd`'s own offload push may legitimately re-populate either segment (allocator choice) and is orthogonal to preempt cleanup.
- `UpsertPreemptsOffloadWithOffloadOnEvict` — covers the `offload_on_evict = true` push path.

yokinoshitayoki:
@huniu20 It seems that we have some conflicts need to fix. And we may need to have a look at the failed CI tests.

mooncake-store/src/file_storage.cpp
```diff
@@ -555,6 +555,46 @@ tl::expected<void, ErrorCode> FileStorage::OffloadObjects(
...
+            } else {
+                const auto& results = validation.value();
+                for (size_t i = 0; i < validate_keys.size(); ++i) {
+                    if (i >= results.size() || !results[i]) {
+                        const auto& obj_key = validate_keys[i];
+                        const auto& task = task_by_storage_key.at(obj_key);
+                        LOG(INFO)
+                            << "[OFFLOAD] drop stale key, key=" << task.key
+                            << ", generation=" << task.generation;
+                        failed_tasks.push_back(task);
```
fcczzz: This sends a task that `ValidateOffloadGenerations()` has already identified as stale through the generic NACK path. If a new offload for the same key is queued before the old worker reports failure, `NotifyOffloadSuccess()` handles `data_size < 0` before its generation guard and looks up `offloading_tasks` only by key. The stale NACK can therefore decrement the newer task's source refcount and erase its marker.
Please either drop validation-rejected stale tasks locally, or make NACK cleanup conditional on task.generation == task_it->second.generation. Please also add a regression test where an old NACK arrives after a newer offload has been queued.

mooncake-store/src/master_service.cpp
```diff
@@ -6174,6 +6213,18 @@ auto MasterService::NotifyOffloadSuccess(
...
+                    // generation no longer matches. Sentinel 0 keeps the
+                    // orphan-fallback path for pre-generation payloads.
+                    if (task.generation != 0 &&
+                        task.generation != task_it->second.generation) {
+                        LOG(INFO)
+                            << "key=" << request_object_id.user_key
+                            << ", action=drop_stale_offload_notify"
+                            << ", task_gen=" << task.generation
+                            << ", current_gen=" << task_it->second.generation;
+                        continue;
```
fcczzz: This generation check happens after `BatchOffload()` has already committed the object locally. `ValidateOffloadGenerations()` does not close this race: `UpsertStart` can cancel the task after validation but before the SSD commit.
In that ordering, this `continue` rejects the replica metadata but the RPC still returns success. `BucketStorageBackend` only calls `RollbackCommittedBucket()` when the completion handler returns an error, so the stale object remains in its local index and on disk. A later offload of the newer value can then fail with `OBJECT_ALREADY_EXISTS`, and `ScanMeta` may re-register the stale data after restart.
Please make the per-task rejection observable to the worker so the corresponding local commit is rolled back or deleted, without rolling back accepted objects from the same batch.

fcczzz (changes requested):

mooncake-store/src/master_service.cpp
huniu20: Thanks. Both addressed:

**1. NACK gen guard (file_storage.cpp:588)** — validation-rejected keys are now
dropped locally instead of round-tripping through `failed_tasks`, and the NACK
branch of `NotifyOffloadSuccess` gained the same generation guard as the
success branch. A stale NACK can no longer touch a newer generation's marker
or refcnt.

**2. Observable per-task rejection (master_service.cpp:6226)** —
`NotifyOffloadSuccess` now returns `tl::expected<std::vector<uint8_t>, ErrorCode>`
aligned with `tasks` (1 = accepted, 0 = rejected). The worker's
`complete_handler` uses it to call the new
`BucketStorageBackend::PartialRollbackKeys()`, which evicts only the rejected
keys from `object_bucket_map_` and reclaims the bucket file if partial
rollback drains it — accepted siblings in the same batch are left intact.

Regression tests: `StaleNackDoesNotClobberNewerOffloadTask`,
`NotifyOffloadSuccessMixedBatchReportsPerTaskAcceptance`, and a tightened
`UpsertPreemptsInFlightOffloadAndDropsStaleNotify` (`notify->front() == 0`).

Also folded in the follow-up snapshot codec fix that broke MixedMemoryAndLocalDiskRoundTrip after the rebase.

PTAL.

mooncake-store/src/storage_backend.cpp
```diff
@@ -2788,6 +2788,77 @@ void BucketStorageBackend::RollbackCommittedBucket(
...
+        // No accepted sibling left. Reuse the bucket-scope rollback so the
+        // on-disk file is reclaimed instead of turning into an orphan.
+        LOG(INFO) << "PartialRollbackKeys: bucket " << bucket_id
+                  << " drained by partial rollback; running full rollback";
+        RollbackCommittedBucket(bucket_id, /*keys=*/{});
+        return;
+    }
+    LOG(INFO) << "PartialRollbackKeys: evicted " << keys_to_evict.size()
+              << " keys from bucket " << bucket_id
+              << " (accepted siblings retained)";
```
fcczzz: `PartialRollbackKeys()` only removes rejected keys from `object_bucket_map_`. If the bucket still contains an accepted sibling, `bucket_meta->keys` / `metadatas` and the on-disk `.meta` / `.bucket` files remain unchanged. As a result, `BucketScan()` / `ScanMeta()` can still report the rejected key, and `Init()` will restore it after restart. This also decrements `total_size_` even though the bytes remain on disk.
Please persist the removal—e.g. rewrite the bucket metadata, rebuild the bucket with only accepted keys, or introduce a durable tombstone—and add an end-to-end `BucketStorageBackend` regression test that commits two keys in one bucket, rejects one, then re-initializes and runs `ScanMeta()`, verifying that the rejected key stays absent while the accepted key remains readable.

huniu20: Thanks, fixed in the latest push.

`PartialRollbackKeys` now also compacts `bucket_meta->keys/metadatas` and rewrites the bucket `.meta`, so rejected keys cannot be resurrected by `Init()` / `ScanMeta()` / `ReRegisterOffloadedObjects()` after restart.

I also fixed the related `total_size_` accounting issue: partial rollback no longer subtracts payload bytes that still physically remain in the `.bucket` file; it only re-accounts the `.meta` size delta. If rewriting `.meta` fails, we now escalate to full bucket rollback to avoid leaving stale keys recoverable on disk.

Added regression coverage for:
- partial rollback persisting across restart
- all-rejected partial rollback falling back to full bucket rollback

Validated:
- `storage_backend_test`: 91/91 passed
- `offload_on_evict_test`: 18/18 passed

PTAL.

yokinoshitayoki (changes requested):
This PR has expanded significantly beyond the original change. Could you provide representative workload data demonstrating that preempting already in-flight offloads is necessary and that the benefit justifies the added protocol complexity? Otherwise maybe it's better to switch to the more conservative queued-only behavior. Without such evidence, the current scope is difficult to justify.

huniu20:
Thanks for the careful review. You're right that this PR grew beyond the original QUEUED-only fix, and it's fair to ask for justification.

**Where the added complexity came from**

For full context: the QUEUED-only version was the original scope. The IN-FLIGHT extension (generation guard, per-task acceptance vector, `PartialRollbackKeys`, snapshot codec bump) was added, then a subsequent review round surfaced two follow-up correctness issues in that IN-FLIGHT path — bucket `.meta` needing to be rewritten so rejected keys can't be resurrected by `Init()` / `ScanMeta()` on restart, and `total_size_` needing to stay in physical-bucket terms once `.bucket` and `.meta` diverge after partial rollback. Both of those fixes are only meaningful because we do IN-FLIGHT preemption; under QUEUED-only, no partial rejection can occur and neither issue exists.

So the complexity you're seeing is real, but it's the compounding correctness surface of doing IN-FLIGHT properly, not gold-plating.

**Why IN-FLIGHT preemption exists in this scope**

With `offload_on_evict=true`, once eviction pushes a key into `offloading_objects`, the DRAM replica is refcnt-protected until `NotifyOffloadSuccess` returns. A concurrent `UpsertStart` on that key hits one of three windows:

1. **QUEUED** — task in master queue, worker hasn't pulled it yet.
2. **IN-FLIGHT (drain)** — worker has pulled the task and is doing the SSD write.
3. **AWAITING-ACK** — SSD write done, `NotifyOffloadSuccess` RPC in flight.

Rejecting the upsert (`OBJECT_HAS_REPLICATION_TASK`) is only safe for window 1. For windows 2 and 3 the client would have to poll until the worker's next heartbeat cycle completes — with `heartbeat_interval_seconds` defaulting to 10s and often tuned to 1–2s in practice, that translates into substantial tail latency on the hot `Upsert` path. In our internal training workload (rollout re-issuing keys every few hundred ms while offload is draining) that window is hit routinely, not exceptionally.

That said, I agree that's a narrative argument, not a measurement. I don't have a reproducer packaged that would give you numbers you can verify independently.

**Proposal**

I'm happy to split this into two PRs:

- **PR A (this one, re-scoped)**: QUEUED-only preemption — the original `3417aa1c` change plus its clang-format follow-up. No `generation` field, no acceptance vector, no `PartialRollbackKeys`, no snapshot codec bump. Small, self-contained, no new protocol surface. The two follow-up correctness fixes above become moot because their precondition (partial rejection) can no longer occur.
- **PR B (follow-up)**: IN-FLIGHT preemption via generation guard + per-task acceptance vector + `PartialRollbackKeys` + snapshot persistence, with workload numbers or a synthetic reproducer so the necessity argument is empirical.

If that split works for you, I'll force-push PR A on this branch and open PR B separately once I have the evidence. @yokinoshitayoki cc @fcczzz

yokinoshitayoki:
> Thanks for the careful review. You're right that this PR grew beyond the original QUEUED-only fix, and it's fair to ask for justification.感谢您细致的审查。您说得对，这个 PR 的改动范围确实超出了最初仅处理 QUEUED 状态的修复，要求提供合理性说明是合理的。
> 
> **Where the added complexity came from新增复杂性的来源**
> 
> For full context: the QUEUED-only version was the original scope. The IN-FLIGHT extension (generation guard, per-task acceptance vector, `PartialRollbackKeys`, snapshot codec bump) was added, then a subsequent review round surfaced two follow-up correctness issues in that IN-FLIGHT path — bucket `.meta` needing to be rewritten so rejected keys can't be resurrected by `Init()` / `ScanMeta()` on restart, and `total_size_` needing to stay in physical-bucket terms once `.bucket` and `.meta` diverge after partial rollback. Both of those fixes are only meaningful because we do IN-FLIGHT preemption; under QUEUED-only, no partial rejection can occur and neither issue exists.完整背景如下：最初的范围确实只是处理 QUEUED 状态的版本。后来增加了 IN-FLIGHT 扩展（包括生成保护、每个任务的接受向量、 `PartialRollbackKeys` 、快照编解码器版本升级），随后的一轮审查又发现了该 IN-FLIGHT 路径中的两个后续正确性问题——桶 `.meta` 需要重写，以防止被拒绝的键在重启时被 `Init()` / `ScanMeta()` 恢复；以及一旦 `.bucket` 和 `.meta` 在部分回滚后出现分歧， `total_size_` 需要保持在物理桶的范畴内。这两项修复之所以有意义，完全是因为我们实现了 IN-FLIGHT 抢占；在仅处理 QUEUED 状态的方案下，不会发生部分拒绝，这两个问题也就不存在。
> 
> So the complexity you're seeing is real, but it's the compounding correctness surface of doing IN-FLIGHT properly, not gold-plating.因此，您看到的复杂性是真实存在的，但这是为了正确实现 IN-FLIGHT 功能而累积的正确性保障，并非过度设计。
> 
> **Why IN-FLIGHT preemption exists in this scope为什么需要在此范围内进行飞行中抢占**
> 
> With `offload_on_evict=true`, once eviction pushes a key into `offloading_objects`, the DRAM replica is refcnt-protected until `NotifyOffloadSuccess` returns. A concurrent `UpsertStart` on that key hits one of three windows:通过 `offload_on_evict=true` ，一旦驱逐将键推入 `offloading_objects` ，DRAM 副本就会受到引用计数保护，直到 `NotifyOffloadSuccess` 返回。对该键的并发 `UpsertStart` 会命中以下三个窗口之一：
> 
> 1. **QUEUED** — task in master queue, worker hasn't pulled it yet.排队中 — 任务在主队列中，工作线程尚未拉取。
> 2. **IN-FLIGHT (drain)** — worker has pulled the task and is doing the SSD write.飞行中（排空） — 工作线程已拉取任务，正在执行 SSD 写入。
> 3. **AWAITING-ACK** — SSD write done, `NotifyOffloadSuccess` RPC in flight.等待确认 — SSD 写入已完成， `NotifyOffloadSuccess` RPC 正在传输中。
> 
> Rejecting the upsert (`OBJECT_HAS_REPLICATION_TASK`) is only safe for window 1. For windows 2 and 3 the client would have to poll until the worker's next heartbeat cycle completes — with `heartbeat_interval_seconds` defaulting to 10s and often tuned to 1–2s in practice, that translates into substantial tail latency on the hot `Upsert` path. In our internal training workload (rollout re-issuing keys every few hundred ms while offload is draining) that window is hit routinely, not exceptionally.拒绝该 upsert（ `OBJECT_HAS_REPLICATION_TASK` ）操作仅在窗口 1 中是安全的。对于窗口 2 和 3，客户端必须轮询直到工作节点的下一个心跳周期完成 — 而 `heartbeat_interval_seconds` 默认值为 10 秒，实际调优后通常为 1 到 2 秒，这会在热路径 `Upsert` 上转化为显著的尾部延迟。在我们的内部训练工作负载中（在卸载进行时，每几百毫秒重新下发键值），该窗口是常规命中而非异常情况。
> 
> That said, I agree that's a narrative argument, not a measurement. I don't have a reproducer packaged that would give you numbers you can verify independently.话虽如此，我同意这是一个叙述性论证，而非基于测量数据。我手头没有可复现的测试包，无法提供你能够独立验证的具体数值。
> 
> **Proposal  建议方案**
> 
> I'm happy to split this into two PRs:我很乐意将其拆分为两个 PR：
> 
> * **PR A (this one, re-scoped)**: QUEUED-only preemption — the original `3417aa1c` change plus its clang-format follow-up. No `generation` field, no acceptance vector, no `PartialRollbackKeys`, no snapshot codec bump. Small, self-contained, no new protocol surface. The two follow-up correctness fixes above become moot because their precondition (partial rejection) can no longer occur.PR A（当前这个，重新划定范围）：仅限 QUEUED 状态的抢占——原始的 `3417aa1c` 变更及其 clang-format 后续修复。不包含 `generation` 字段、验收向量、 `PartialRollbackKeys` ，也不涉及快照编解码器版本升级。改动小、自包含、不引入新的协议接口。上述两个后续正确性修复将不再相关，因为它们的前提条件（部分拒绝）已不可能发生。
> * **PR B (follow-up)**: IN-FLIGHT preemption via generation guard + per-task acceptance vector + `PartialRollbackKeys` + snapshot persistence, with workload numbers or a synthetic reproducer so the necessity argument is empirical.PR B（后续 PR）：通过生成守卫 + 每任务验收向量 + `PartialRollbackKeys` + 快照持久化实现 IN-FLIGHT 状态的抢占，并附带工作负载数据或合成复现用例，使必要性论证建立在实证基础上。
> 
> If that split works for you, I'll force-push PR A on this branch and open PR B separately once I have the evidence. @yokinoshitayoki cc 如果这个拆分方案对你来说可行，我会在这个分支上强制推送 PR A，并在拿到证据后单独发起 PR B。抄送@fcczzz

Thanks, this split works for me. Please re-scope this PR to queued-only and move IN-FLIGHT preemption to a follow-up backed by workload data or a reproducer. Please ping me when it is ready for review.

huniu20:
Rebased and rescoped to QUEUED-only per @yokinoshitayoki's suggestion. Summary of changes since the last review round:

QUEUED-only preemption. UpsertStart cancels an offload task only when every mirror is still queued; once any store worker owns the task the upsert waits on NotifyOffloadSuccess as before.

Multi-mirror correctness (@Lin-z-w). OffloadingTask::mirror_clients records each destination at push time; CancelQueuedOffloadTask reclaims all of them under offloading_mutex_ and rolls back if any is missing.

Regression coverage. 5 new tests: UpsertPreemptsQueuedOffload, BatchUpsertPreemptsQueuedOffload, UpsertRejectedWhenOffloadInFlight, UpsertRejectedWhenOneOfTwoMirrorsDrained, UpsertPreemptsQueuedOffloadWithOffloadOnEvict.

IN-FLIGHT extension deferred. The generation guard acceptance vector PartialRollbackKeys / snapshot codec bump that @fcczzz reviewed are dropped from this PR and will return in a follow-up with workload data.

Diff is 3 files, +365/-22. PTAL @Lin-z-w @yokinoshitayoki @fcczzz.

fcczzz (commented):
Thanks for narrowing the scope. The updated approach looks sound to me: tracking every mirror client and cancelling only while all mirrors remain queued provides a clear distinction between QUEUED and IN-FLIGHT offloads. Restoring any mirrors already reclaimed when cancellation cannot proceed also seems to preserve the original queue state correctly.

It looks like the branch currently conflicts with the latest main around MasterService::EvictTenantMemoryForQuota(). When convenient, could you please rebase? After that, I think this should be in good shape for approval.

yokinoshitayoki:
@huniu20 Could we have a look at the failed CI test?

huniu20:
The CI failure is an `ubuntu-22.04` flake in the `snapshot_test` family,
unrelated to this PR. Same code (`5f2a05bf` ≡ `e5425e4d`) produced
different verdicts on runs 2 and 3 — `build (3.12)` failed on run 2 but
passed on run 3, `build (3.10)` passed on run 2 but failed on run 3 —
while the failing tests (`MixedMemoryAndLocalDiskRoundTrip`,
`UnmountSegmentImmediateCleanup`) both run under configs that never call
`UpsertStart` and never populate `offloading_tasks`. Locally the failing
test passes 5/5. Can we retry the job? @yokinoshitayoki

yokinoshitayoki:
retried.

yokinoshitayoki (approved):
Thanks a lot for your efforts, I think this PR looks good to me now, we can merge this PR after the CI tests passed.

ykwd (approved):

fcczzz (approved):
LGTM

PR #3254, opened 2026-08-03
[Store] Fix processing_keys double-erase UAF in MetadataAccessorRW

rockuw:
## Description

Fix a use-after-free (double erase) in the `MasterService::MetadataAccessorRW`
constructor's invalid-handle cleanup that segfaults `mooncake_master`
deterministically (fault address `0x0`, inside
`std::unordered_set<std::string>::erase(const_iterator)`).

**Root cause.** When an object loses all valid replicas, the constructor's
cleanup does:

```cpp
const bool had_processing =
    processing_it_ != tenant_state_->processing_keys.end();
this->Erase();  // -> EraseMetadata() already does processing_keys.erase(key)
if (tenant_state_ != nullptr && had_processing) {
    this->EraseFromProcessing();  // erases the SAME node via a dangling iterator
}
```

`Erase()` -> `EraseMetadata()` already removes the key from `processing_keys`
by key (freeing the node). The subsequent `EraseFromProcessing()` then erases
the *same* node again through the now-dangling `processing_it_` iterator.
libstdc++ re-reads the cached hash from the freed node, walks the bucket chain
looking for it by address, runs off the chain end and dereferences nullptr
-> SIGSEGV.

**Trigger observed in production.** A snapshot restore resurrected dead
("ghost") clients and their mounted segments. When the restore grace period
expired, `ClientMonitorFunc` expired the ghosts and `PrepareUnmountSegment`
destroyed their segment allocators, invalidating every replica handle on those
segments. Any RPC hitting a key whose replicas all lived on ghost segments and
which was still in `processing_keys` (an interrupted put) executed the buggy
cleanup before `ClearInvalidHandles` could sweep the key via the safe path.
The crash repeats every restart while such state persists in snapshots.

Note this cleanup path only runs when `!(enable_ha_ && enable_oplog_)`; the HA
+ oplog configuration takes a different path, but the default configuration
hits it.

**Fix.** Drop the redundant `EraseFromProcessing()` call; `EraseMetadata()`
already handles `processing_keys` (and `replication_tasks` / promotion tasks)
unconditionally. The three remaining `EraseFromProcessing()` call sites are
safe: they are either mutually exclusive with `Erase()` (`else if` after an
`Exists()`/validity check) or order the iterator erase before `Erase()`, in
which case the by-key erase inside `EraseMetadata()` is a no-op.

**Test.** Adds `master_service_processing_key_double_erase_test`, a
forked-child reproduction that drives the real production path through public
APIs plus one friend hook:

1. `MountSegment` (a "ghost" client mounts a segment);
2. `PutStart` without `PutEnd` (key stays in `processing_keys` with an
   incomplete replica on that segment), plus a second completed key on the
   same metadata shard to keep the tenant non-empty;
3. `PrepareUnmountSegment` directly (`UnmountSegment` would run
   `ClearInvalidHandles` and erase the crafted object via the safe path
   first), destroying the allocator so the replica handle becomes invalid;
4. `PutEnd` -> `MetadataAccessorRW` constructor cleanup.

On buggy code the forked child dies with SIGSEGV and the test FAILs; after the
fix the child exits cleanly (`PutEnd` reports `OBJECT_NOT_FOUND`) and the test
passes. So the test fails before this change and passes after it.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

The new regression test forks a child that runs the incident scenario; on the
unfixed code the child SIGSEGVs inside `processing_keys.erase()` exactly as in
production, and the test fails. With the fix, the child exits 0 and the test
passes.

**Test commands:**
```bash
cd build && cmake .. && make master_service_processing_key_double_erase_test -j
./build/mooncake-store/tests/master_service_processing_key_double_erase_test
# regression guard on the surrounding areas:
make master_service_test master_service_ssd_test -j
./build/mooncake-store/tests/master_service_ssd_test
```

**Test results:**
- [x] Unit tests pass

Reproduced the original crash in production (periodic SIGSEGV at the same
instruction after snapshot restore), confirmed the root cause with the repro
test, and verified the fixed master passes the store unit tests.

## Checklist

      (pre-commit is not installable in my environment; ran the applicable
      hooks manually instead: code_format.sh, trailing-whitespace,
      end-of-file-fixer, no merge-conflict markers)

## AI Assistance Disclosure

- [ ] No AI tools were used

Claude Code assisted with the crash triage, root-cause analysis, the fix and
the regression test. The submitter has reviewed every changed line and can
defend the change end-to-end.

### DIFF
--- a/mooncake-store/include/master_service.h
+++ b/mooncake-store/include/master_service.h
@@ -82,6 +82,10 @@ class MasterServiceTenantQuotaTest;
 // relying on segment pressure plus the background eviction thread.
 class BatchEvictTest;
 class MasterServiceHATest;
+// Friended so the processing_keys double-erase reproduction test can
+// invalidate a segment allocator via PrepareUnmountSegment WITHOUT the
+// ClearInvalidHandles sweep that MasterService::UnmountSegment performs.
+class MasterServiceProcessingKeyDoubleEraseTest;
 }  // namespace test
 namespace benchmarks {
 class BatchEvictBench;
@@ -112,6 +116,8 @@ class MasterService {
     friend class benchmarks::BatchEvictBench;
     friend class test::MasterServiceTenantQuotaTest;
     friend class test::BatchEvictTest;
+    // double-erase processing_keys UAF repro (2026-08-03 prod segfault)
+    friend class test::MasterServiceProcessingKeyDoubleEraseTest;
     friend class MasterSnapshotManager;    // Allow access to internal state for
                                            // snapshot
     friend class ha::MasterSnapshotCodec;  // Allow codec to access private
@@ -1739,12 +1745,12 @@ class MasterService {
                 }
                 // If no valid replicas remain, delete the whole object.
                 if (!it_->second.IsValid()) {
-                    const bool had_processing =
-                        processing_it_ != tenant_state_->processing_keys.end();
+                    // NOTE: Erase() -> EraseMetadata() already removes the key
+                    // from processing_keys (by key), so calling
+                    // EraseFromProcessing() here would re-erase the same node
+                    // via the now-dangling processing_it_ iterator
+                    // (use-after-free, prod segfault 2026-08-03).
                     this->Erase();
-                    if (tenant_state_ != nullptr && had_processing) {
-                        this->EraseFromProcessing();
-                    }
                     if (tenant_state_ != nullptr) {
                         service_->ErasePromotionTaskIfPresent(
                             *tenant_state_, object_id_.user_key,
--- a/mooncake-store/tests/CMakeLists.txt
+++ b/mooncake-store/tests/CMakeLists.txt
@@ -68,6 +68,8 @@ endif()
 add_store_test(master_service_test master_service_test.cpp)
 add_store_test(batch_evict_test batch_evict_test.cpp)
 add_store_test(master_service_config_test master_service_config_test.cpp)
+add_store_test(master_service_processing_key_double_erase_test
+               master_service_processing_key_double_erase_test.cpp)
 add_store_test(master_service_tenant_quota_test
                master_service_tenant_quota_test.cpp)
 add_store_test(batch_remove_test batch_remove_test.cpp)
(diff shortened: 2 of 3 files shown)


he-yufeng:
Verified the root cause link by link before saying anything: the accessor's `Erase()` reaches `EraseMetadata`, which does `processing_keys.erase(key)` (by key) and frees the node `processing_it_` points into; the follow-up `EraseFromProcessing()` then calls `erase(processing_it_)` on the dangling iterator. The fix removes exactly that redundant call, which is the minimal correct change.

I also checked the other `EraseFromProcessing` sites for the same ordering hazard and they are all safe: the `!IsValid()` / `else if (InProcessing())` branch around line 1745 only calls it when no `Erase()` happened, the completion-driven call around 3822 has no preceding `Erase()`, and the one near 4046 erases the processing entry before any metadata erase. No siblings to chase.

Not approving on a first read through a metadata-concurrency path, but the repro test covers the deterministic trigger well and the mechanism above is why the crash was repeatable rather than racy.

rockuw:
Rebased onto latest `main` (14d39fff) to resolve the conflict introduced by #3118 (the new `BatchEvictTest` friend declaration landed next to this PR's; both are kept).

Re-verified on the new base: `master_service_processing_key_double_erase_test` passes, and `master_service_test` (147 tests) and `master_service_ssd_test` (18 tests) all pass. The rebase touched only the friend-declaration block; the fix and the repro test are byte-identical to the version whose forked child SIGSEGVs when the fix is reverted.

rockuw:
Fixed the CI build failure: CI links an older glog that lacks `google::IsGoogleLoggingInitialized()` (added in glog 0.6), so the new test failed to compile there while building fine locally. Switched the test to the same pattern as the other store tests (`InitGoogleLogging` in `SetUp()` / `ShutdownGoogleLogging` in `TearDown()`, e.g. `buffer_allocator_test.cpp`), which works on both old and new glog. No functional change; test re-verified locally.

Aionw (approved):
Thanks for investigating this issue.LGTM

rockuw:
@ykwd @stmatengss pls take a look.

ykwd (approved):
Nice catch! And thanks for the fix

Issue #3098, opened 2026-07-24
[RFC][Store] Scenario DSL and deterministic concurrency runner for MasterService tests


Aionw:
## Summary

> [!IMPORTANT]
> This RFC and PR #3100 are an experimental prototype. The PR is provided primarily as a concrete, runnable artifact for design and code review; it is not a commitment to merge this implementation or to adopt the current DSL, checkpoint API, scheduler, file organization, or migration boundary as the final design. Any part of the approach may change substantially—or the proposal may be rejected—based on review feedback.

Introduce a typed, embedded C++ scenario DSL for `MasterService` tests and a deterministic concurrency runner based on test-only semantic checkpoints.

The goal is to move business-behavior coverage away from implementation-coupled fixtures while keeping stress, backend, protocol-compatibility, and low-level fault-injection tests as focused tests.

## Motivation

The existing `MasterService` test suites have accumulated several problems:

- large monolithic files make reviews expensive;
- setup, client UUID management, tenant policy files, and lifecycle cleanup are repeated;
- many tests describe implementation steps instead of business scenarios;
- concurrency tests rely on timing, sleeps, or mutex contention and are difficult to reproduce in CI;
- failures do not preserve the observed schedule or a reusable event trace.

A scenario DSL provides a stable vocabulary for public `MasterService` behavior. Semantic checkpoints provide deterministic control at the small number of concurrency boundaries where ordering is part of the contract.

## Goals

- Express `MasterService` business scenarios through its public C++ interfaces.
- Make success the default and require expected failures to be explicit.
- Support tenant, object, replica, quota, segment, copy/move, upsert, task, drain, lease, eviction, and offload scenarios.
- Distinguish order-independent `Parallel(...)` execution from explicit `Interleave(...)` choreography.
- Reproduce stale-writer and policy-deletion races without sleeps or scheduler luck.
- Capture a versioned JSON schedule and event trace on failure and replay it strictly.
- Compile all checkpoint infrastructure out of non-unit-test builds.
- Split the former monolithic suites by behavior area.

## Non-goals

- Random schedule exploration or state-space enumeration.
- A linearizability checker.
- Instruction-level race reproduction between semantic checkpoints.
- RPC or cross-process schedule replay in the first version.
- Replacing stress tests, allocator/backend tests, serialization compatibility tests, or performance tests with the DSL.
- Proving strict same-client/same-key ABA safety without an operation generation in the API.

## Proposed DSL

[...]

PR #3166, opened 2026-07-28
[Store] Add exponential backoff for ordered oplog writer retries

Icedcoco:
## Description

Replace the fixed 1 ms retry interval in `OrderedOpLogWriter` with per-batch exponential backoff.

The retry delay starts at 1 ms, doubles after each failed write, and is capped at 1 second. A successful write resets the delay for the next batch. `Stop()` can interrupt an active backoff immediately through the existing condition variable.

This change preserves the current retryable error set, batch ordering, durable-prefix handling, and callback behavior. Error classification, retry deadlines, and terminal-state handling are intentionally left out of scope.

## Module

- [ ] Transfer Engine (`mooncake-transfer-engine`)
- [x] Mooncake Store (`mooncake-store`)
- [ ] Mooncake EP (`mooncake-ep`)
- [ ] Mooncake PG (`mooncake-pg`)
- [ ] Integration (`mooncake-integration`)
- [ ] P2P Store (`mooncake-p2p-store`)
- [ ] Python Wheel (`mooncake-wheel`)
- [ ] Common (`mooncake-common`)
- [ ] Mooncake RL (`mooncake-rl`)
- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [x] Performance improvement
- [ ] Other

## How Has This Been Tested?

Added unit tests covering:

- Exponential growth and the 1-second backoff cap
- Backoff reset after a successful batch
- Immediate interruption of a long backoff by `Stop()`
- Existing retry, ordering, and callback behavior

**Test commands:**

```bash
cmake --build build --target ordered_oplog_writer_test -j32
LSAN_OPTIONS=detect_leaks=0 ./build/mooncake-store/tests/ordered_oplog_writer_test
pre-commit run --files \
  mooncake-store/src/ha/oplog/ordered_oplog_writer.cpp \
  mooncake-store/tests/ha/oplog/ordered_oplog_writer_test.cpp
```

**Test results:**

- [x] Unit tests pass (`31/31`)
- [ ] Integration tests pass (not applicable)
- [ ] Manual testing done (not applicable)

## Checklist

- [x] I have performed a self-review of my own code
- [x] I have formatted my code using `./scripts/code_format.sh`
- [x] I have run `pre-commit run --all-files` and all hooks pass
- [ ] I have updated the documentation (not applicable; no user-facing API change)
[...]

PR #3256, opened 2026-08-03
[Store] Add minimal MasterService scenario DSL

Aionw:
## Description

Add the smallest useful typed C++ embedded scenario DSL for `MasterService` tests and prove it end to end by migrating `MasterServiceTest.PutStartEndFlow`.

The test-only foundation provides:

- declarative `MemoryNode` setup with lazy `MasterService` construction;
- typed `PutStart` and `PutEnd` actions with named actors;
- deterministic name-derived client UUIDs;
- success-by-default validation and explicit `ExpectError(ErrorCode)`;
- `Object(...).IsReadable()` verification;
- RAII ownership of service and scenario resources.

The old direct-fixture body is removed, so the behavior is covered once through the DSL. Public production APIs are unchanged.

Related RFC: #3098

This is the first small split from experimental draft #3100, not a duplicate of its broad implementation. It deliberately excludes concurrency checkpoints, scheduling, replay artifacts, tenants/quotas, additional operations, RPC, snapshots, and broad test migration.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
cmake -S . -B build -G Ninja -DBUILD_UNIT_TESTS=ON -DWITH_STORE_RUST=OFF
cmake --build build --target master_scenario_test master_service_test --parallel 128
GLOG_minloglevel=3 ./build/mooncake-store/tests/master_scenario_test --gtest_color=no
GLOG_minloglevel=3 ./build/mooncake-store/tests/master_service_test --gtest_color=no
pre-commit run --files mooncake-store/tests/CMakeLists.txt mooncake-store/tests/master_service_test.cpp mooncake-store/tests/master_scenario.cpp mooncake-store/tests/master_scenario.h mooncake-store/tests/master_scenario_test.cpp
git diff --check
```

**Test results:**
- [x] Unit tests pass
- [x] Manual testing done (focused scenario and full relevant legacy suite)
- `master_scenario_test`: 1/1 passed.
- `master_service_test`: 146/146 passed.
- All touched-file pre-commit hooks passed.

A separate production/non-unit build is unaffected because all implementation and build wiring is under `mooncake-store/tests`, which is included only when `BUILD_UNIT_TESTS` is enabled.

## Checklist

- [x] I have updated the documentation (not applicable: test-only behavior; RFC #3098 documents the experiment)
- [x] For changes >500 LOC: I have filed an RFC issue (not applicable to this 368-line diff; related RFC #3098 exists)

Touched-file pre-commit validation was used per repository guidance; `--all-files` was not run to avoid unrelated rewrites.

This PR is intentionally a draft. The human submitter must review every changed line and be able to defend the implementation before marking it ready.

## AI Assistance Disclosure

- [ ] No AI tools were used

OpenAI Codex helped reduce RFC #3098 and draft #3100 to this focused typed DSL slice, implement the test-only code, migrate the single scenario, and run validation. The human submitter remains responsible for reviewing and defending every changed line.

### DIFF
--- a/mooncake-store/tests/CMakeLists.txt
+++ b/mooncake-store/tests/CMakeLists.txt
@@ -69,6 +69,8 @@ add_store_test(master_service_test master_service_test.cpp)
 add_store_test(batch_evict_test batch_evict_test.cpp)
 add_store_test(master_scenario_test master_scenario.cpp
                master_scenario_test.cpp)
+add_store_test(master_service_scenario_test master_scenario.cpp
+               master_service_scenario_test.cpp)
 add_store_test(master_service_config_test master_service_config_test.cpp)
 add_store_test(master_service_processing_key_double_erase_test
                master_service_processing_key_double_erase_test.cpp)
--- a/mooncake-store/tests/CMakeLists.txt
+++ b/mooncake-store/tests/CMakeLists.txt
@@ -67,6 +67,8 @@ if(ENABLE_KV_EVENTS)
 endif()
 add_store_test(master_service_test master_service_test.cpp)
 add_store_test(batch_evict_test batch_evict_test.cpp)
+add_store_test(master_scenario_test master_scenario.cpp
+               master_scenario_test.cpp)
 add_store_test(master_service_config_test master_service_config_test.cpp)
 add_store_test(master_service_processing_key_double_erase_test
                master_service_processing_key_double_erase_test.cpp)
--- a/mooncake-store/tests/master_scenario.cpp
+++ b/mooncake-store/tests/master_scenario.cpp
@@ -1,0 +1,184 @@
+#include "master_scenario.h"
+
+#include <gtest/gtest.h>
+
+#include <algorithm>
+#include <utility>
+
+#include "types.h"
+
+namespace mooncake::test {
+namespace {
+
+constexpr uint64_t kFnvOffset = 14695981039346656037ULL;
+constexpr uint64_t kFnvPrime = 1099511628211ULL;
+
+uint64_t StableHash(std::string_view kind, std::string_view name,
+                    uint64_t seed) {
+    uint64_t hash = seed;
+    for (const char value : kind) {
+        hash = (hash ^ static_cast<unsigned char>(value)) * kFnvPrime;
+    }
+    for (const char value : name) {
+        hash = (hash ^ static_cast<unsigned char>(value)) * kFnvPrime;
+    }
+    return hash;
+}
+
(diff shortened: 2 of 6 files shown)


Icedcoco:
Thanks for keeping this initial slice focused. I had two questions about the assertion semantics before we build more scenarios on top of it.

First, how should combined expectations behave? For example, `PutStart(...).ExpectError(...).ExpectReplicas(1)` returns as soon as the expected error is observed, so the replica expectation is not checked. Similarly, `Object(...).IsNotReady().HasReplicas(1)` returns after checking `IsNotReady()` and silently skips `HasReplicas()`.

Would it make sense to treat these combinations as invalid scenario definitions and report a test failure? A small runtime validation may be enough here.

Second, would it be useful to add a few negative contract tests for the DSL itself? The migrated scenario covers valid success and expected-error paths, but it does not verify that the DSL emits a gtest failure for an unexpected success/error, a wrong error code, or a replica count/status mismatch. Since this code becomes the assertion layer for future scenarios, perhaps a few focused `EXPECT_NONFATAL_FAILURE` cases would help guard against expectations being accepted silently. What do you think would be the minimum useful coverage here?

Separately, the current CI failure looks very likely to be the same pre-existing LSan issue discussed in #3166. In both cases, all gtest cases in the affected binary passed, and process teardown then reported a direct 8192-byte leak from `numa_node_to_cpus` in `libnuma`. The affected binary is different here (`rdma_async_event_drain_test` rather than `mmap_arena_fallback_test`), so I cannot confirm that the trigger is identical, but the failure signature is the same and does not appear related to this PR. Does that match your reading?

Aionw:
Using C++20 concepts, invalid expectation combinations are now rejected at compile time. For example, an error expectation cannot be combined with replica count/status expectations, and `IsNotReady()` cannot be combined with replica assertions.

I’ve also added focused contract tests for the scenario DSL itself. These tests verify that unexpected successes or errors, incorrect error codes, and replica count/status mismatches are surfaced as gtest failures rather than being silently accepted.

Icedcoco:
LGTM pending CI

ykwd:
Could you take a look at the failed CI? Thanks
The following tests FAILED:
	 30 - graceful_shutdown_test (Failed)

Aionw:
After rebased on latest main branch, seems all CI test has passed.

ykwd (approved):

PR #3257, opened 2026-08-03
[Store] Fix no-operation update of LOCAL_DISK replica endpoint/size o…

liangxu2000:
### Problem

**In one line:** the code that updates an existing LOCAL_DISK replica's `endpoint`/`object_size` writes through `rep.get_descriptor()`, which returns a **by-value copy**, so the write lands on a throwaway temporary and the real record is never changed — the update is a no-operation. Two sites are affected (`master_service.cpp`): the `AddReplica` update branch and the `NotifyOffloadSuccess` update branch.

```cpp
[&replica](Replica& rep) {
    rep.get_descriptor()                       // get_descriptor() returns a temporary COPY by value
        .get_local_disk_descriptor()
        .transport_endpoint = ...;             // written into the copy -> discarded at end of statement
    rep.get_descriptor().get_local_disk_descriptor().object_size = ...;   // same, no effect
});
```

`get_descriptor()` (`replica.h`) builds a fresh `Descriptor` copy each call; the replica's real data lives in the private `data_` member, which is never touched.

**Is the update always ineffective?** Yes — whenever this "update" branch runs, nothing is written. But it only becomes a *visible* failure under a specific condition:

| Condition | Does the lost update matter? | Outcome |
|---|---|---|
| Re-report carries the **same** endpoint (common case) | No — stored value was already correct | No observable difference (why the bug hid so long) |
| Re-report carries a **different** endpoint (node restart, new RPC address) | **Yes** — record keeps the stale value | Stale endpoint → reads dial a dead address → **read failure** |

**HA note:** in HA mode the oplog-persistence branch persists the **new** endpoint/size while the in-memory metadata keeps the **old** value — the two diverge, and behavior can drift if state is later restored from oplog/snapshot.

### Solution

1. **Add a real mutator** `Replica::update_local_disk_location(transport_endpoint, object_size)` (`replica.h`) that mutates the internal `LocalDiskReplicaData` directly via `std::get_if<LocalDiskReplicaData>(&data_)` (does nothing for non-local_disk replicas). It also keeps the `allocated_file_size` metric in sync: when `object_size` changes it does `dec_allocated_file_size(old) + inc_allocated_file_size(new)`, mirroring the accounting already done in the replica constructor/destructor.
2. **Replace both no-operation update sites** with a call to this mutator:
   - `AddReplica` update branch (`master_service.cpp`);
   - `NotifyOffloadSuccess` update branch (`master_service.cpp`).

### Key mechanism: `Replica::update_local_disk_location`

The real data lives in the private `data_` member, unreachable from the call sites; the fix provides an in-class entry point that mutates it directly:

```cpp
// replica.h
void update_local_disk_location(std::string transport_endpoint,
                                uint64_t object_size) {
    if (auto* disk_data = std::get_if<LocalDiskReplicaData>(&data_)) {
        if (disk_data->object_size != object_size) {
            MasterMetricManager::instance().dec_allocated_file_size(
                disk_data->object_size);
            MasterMetricManager::instance().inc_allocated_file_size(object_size);
        }
        disk_data->transport_endpoint = std::move(transport_endpoint);
        disk_data->object_size = object_size;
    }
}
```

Call sites now do:

```cpp
const auto& desc = replica.get_descriptor().get_local_disk_descriptor();
rep.update_local_disk_location(desc.transport_endpoint, desc.object_size);
```

### Testing summary

WSL Ubuntu Debug build (gcc, -O0): full `master_service_test` **147/147 passed**, including the new regression test below; zero regressions.

---

### Scenario: `NotifyOffloadSuccessRefreshesEndpointForSameClient`

- **Purpose**: the same owning client re-reports an already-offloaded key with a new transport endpoint; the existing LOCAL_DISK replica's endpoint must be refreshed in place (this exercises the previously no-operation update branch).
- **Setup**: one `MasterService`; via `NotifyOffloadSuccess`, the same `client_id` reports key `K` first with `endpoint_old`, then again with `endpoint_new` (the old record still exists, so the second call takes the update branch); read back with `GetReplicaList`.
- **Result**: exactly **1** LOCAL_DISK replica remains, with endpoint **`endpoint_new`**.

**Before / after, reproduced on the build tree by reverting the fix:**

Before the fix (old `get_descriptor()`-copy write) — the test FAILS, the endpoint read back is still `endpoint_old`:

```
[ RUN      ] MasterServiceTest.NotifyOffloadSuccessRefreshesEndpointForSameClient
master_service_test.cpp:7720: Failure
  "endpoint_new"
    Which is: "endpoint_old"
[  FAILED  ] MasterServiceTest.NotifyOffloadSuccessRefreshesEndpointForSameClient
```

After the fix (`update_local_disk_location`) — the test PASSES:

```
[ RUN      ] MasterServiceTest.NotifyOffloadSuccessRefreshesEndpointForSameClient
[       OK ] MasterServiceTest.NotifyOffloadSuccessRefreshesEndpointForSameClient (1021 ms)
[  PASSED  ] 1 test.
```

The test fails on the unpatched code and passes after the fix, so it genuinely guards the defect rather than being a tautological green.

---

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
cmake -S . -B build -GNinja -DCMAKE_BUILD_TYPE=Debug -DBUILD_UNIT_TESTS=ON
ninja -C build master_service_test
```

**Test results:**
- [x] Unit tests pass (master_service_test **147/147 PASSED**, incl. 1 new regression test)
- [ ] Integration tests pass (if applicable) (end-to-end integration tests not run; change is master-side metadata only)

## Checklist

- [ ] For changes >500 LOC: I have filed an RFC issue (this PR is ~82 LOC, below the RFC threshold)

## AI Assistance Disclosure

- [ ] No AI tools were used

AI coding assistants were used to help implement the feature, write the tests, diagnose test behavior, and review the change. The human submitter has reviewed the changes and is responsible for defending them end-to-end.

### DIFF
--- a/mooncake-store/include/replica.h
+++ b/mooncake-store/include/replica.h
@@ -422,6 +422,30 @@ class Replica {
         return std::nullopt;
     }
 
+    /**
+     * @brief Update a LOCAL_DISK replica's transport endpoint and object size
+     * in place, keeping the allocated_file_size metric in sync when the size
+     * changes. No-op for non-local_disk replicas.
+     *
+     * Used when the same owning client re-reports an offloaded object (e.g.
+     * re-registration after a restart, where the transport endpoint changes).
+     * Mutates the internal replica data directly: get_descriptor() returns a
+     * by-value copy, so writing through it does not modify the replica.
+     */
+    void update_local_disk_location(std::string transport_endpoint,
+                                    uint64_t object_size) {
+        if (auto* disk_data = std::get_if<LocalDiskReplicaData>(&data_)) {
+            if (disk_data->object_size != object_size) {
+                MasterMetricManager::instance().dec_allocated_file_size(
+                    disk_data->object_size);
+                MasterMetricManager::instance().inc_allocated_file_size(
+                    object_size);
+            }
+            disk_data->transport_endpoint = std::move(transport_endpoint);
+            disk_data->object_size = object_size;
+        }
+    }
+
     [[nodiscard]] size_t get_memory_buffer_size() const {
         if (is_memory_replica()) {
             const auto& mem_data = std::get<MemoryReplicaData>(data_);
--- a/mooncake-store/src/master_service.cpp
+++ b/mooncake-store/src/master_service.cpp
@@ -3962,15 +3962,10 @@ auto MasterService::AddReplica(const UUID& client_id, const std::string& key,
                        client_id;
         },
         [&replica](Replica& rep) {
-            rep.get_descriptor()
-                .get_local_disk_descriptor()
-                .transport_endpoint = replica.get_descriptor()
-                                          .get_local_disk_descriptor()
-                                          .transport_endpoint;
-            rep.get_descriptor().get_local_disk_descriptor().object_size =
-                replica.get_descriptor()
-                    .get_local_disk_descriptor()
-                    .object_size;
+            const auto desc =
+                replica.get_descriptor().get_local_disk_descriptor();
+            rep.update_local_disk_location(desc.transport_endpoint,
+                                           desc.object_size);
         });
     return false;
 }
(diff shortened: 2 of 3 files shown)


fcczzz:
Thanks for working on this — fixing the ineffective update through get_descriptor() looks useful on its own.
I may be missing some context, but this seems related to #3052 and overlaps with #3182 (and possibly the broader recovery work in #2777). Could you clarify the intended scope of this PR?

The regression test here uses the same client_id reporting a changed endpoint, whereas the restart path described in #3052 appears to involve a new client UUID reporting before the old UUID expires. If this PR is intended to address the full client-restart recovery problem, would it make sense to also cover the changed-UUID and stale-client-cleanup path, and clarify how ownership, SSD accounting, and HA persistence should behave?

If the intended scope is only to fix the same-client descriptor-update no-op, that also seems worth stating explicitly, together with how this change is expected to compose with the related PRs.

PR #3259, opened 2026-08-03
[TENT] Fix double-free in HttpMetaStore under concurrent workers: a s…

KubrickLiu:
## Description

`HttpMetaStore` shared a single `CURL*` easy handle across all worker threads.
libcurl documents that an easy handle must not be used from multiple threads
at the same time. Under concurrent `get()`/`set()` from worker threads (e.g.
16+ workers in the transfer benchmark), `curl_easy_reset()` double-freed
handle-owned strings that libcurl had `strdup`'d internally, corrupting the
glibc heap:

### Fix
- Allocate a per-request `CURL*` via a small RAII wrapper (`ScopedCurl`)
  instead of sharing one handle across threads — the pattern libcurl
  recommends for the easy interface.
- Guard `curl_global_init()` with `std::call_once` for a race-free lazy init,
  and report a failure from `curl_global_init()` instead of ignoring it.
- Use `curl_easy_escape(nullptr, ...)` for URL encoding (the handle argument
  is unused in modern libcurl and passing NULL is thread-safe).

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## Testing

- Reproduced the double-free under an AddressSanitizer build with concurrent
  metadata workers; the ASan report on `HttpMetaStore::get` no longer fires
  after the fix.
- Verified transfers over RDMA with the HTTP metadata backend across two
  machines; sustained runs complete without heap corruption or crashes.

### Environment
- NVIDIA Blackwell RTX PRO, driver 580.65.06, CUDA 13.0
- Ubuntu, glibc 2.35, system libcurl

### DIFF
--- a/mooncake-transfer-engine/tent/include/tent/metastore/http.h
+++ b/mooncake-transfer-engine/tent/include/tent/metastore/http.h
@@ -45,8 +45,8 @@ class HttpMetaStore : public MetaStore {
         return size * nmemb;
     }
 
-    std::string encodeUrl(const std::string &key) {
-        char *newkey = curl_easy_escape(nullptr, key.c_str(), key.size());
+    std::string encodeUrl(CURL *curl, const std::string &key) {
+        char *newkey = curl_easy_escape(curl, key.c_str(), key.size());
         std::string encodedKey(newkey ? newkey : "");
         std::string url = endpoint_ + "?key=" + encodedKey;
         if (newkey) curl_free(newkey);
--- a/mooncake-transfer-engine/tent/include/tent/metastore/http.h
+++ b/mooncake-transfer-engine/tent/include/tent/metastore/http.h
@@ -46,16 +46,15 @@ class HttpMetaStore : public MetaStore {
     }
 
     std::string encodeUrl(const std::string &key) {
-        char *newkey = curl_easy_escape(client_, key.c_str(), key.size());
-        std::string encodedKey(newkey);
+        char *newkey = curl_easy_escape(nullptr, key.c_str(), key.size());
+        std::string encodedKey(newkey ? newkey : "");
         std::string url = endpoint_ + "?key=" + encodedKey;
-        curl_free(newkey);
+        if (newkey) curl_free(newkey);
         return url;
     }
 
    private:
     std::atomic<bool> connected_;
-    CURL *client_;
     std::string endpoint_;
 };
 }  // namespace tent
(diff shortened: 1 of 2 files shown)


he-yufeng:
The shape is the right one for libcurl threading: easy handles are per-request state, so a fresh ScopedCurl per call removes the shared-handle reset race instead of guarding it, and call_once on the global init matches the docs. Two things worth a look before merge:

`curl_easy_escape(nullptr, ...)` only accepts a null handle on libcurl 7.82.0 and later; older versions dereference it. I could not find a floor declared in the build, so if any target platform ships an older libcurl this needs a temp handle (or a version check) instead of nullptr.

The per-request handle also drops connection reuse: every get/set now pays a fresh TCP connect, which the shared handle amortized. A thread_local easy handle would keep both thread safety and reuse if the benchmark numbers matter here. Not a blocker for a correctness fix, but worth knowing whether the transfer benchmark regresses before this becomes the steady state.

KubrickLiu:
Thanks for the review!

> curl_easy_escape(nullptr, ...) only accepts a null handle on libcurl 7.82.0+

I don't pin a minimum libcurl version, so I switched to passing
the per-request handle we already create (instead of nullptr). It's now correct
on any libcurl version.

> The per-request handle also drops connection reuse ...

Agreed. This PR keeps per-request handles to land the correctness fix first,
but I'll follow up shortly with a thread_local easy handle so we keep thread
safety and get connection reuse back. Thanks!～～

@he-yufeng

alogfans:
Overall looks good. Fix format issue and reperform CI checks.

KubrickLiu:
The build failure is a pre-existing break on main, unrelated to this PR.

It was introduced by #3234 (commit 22c8fa81), which added a call to
`RealClient::get_into_ranges_internal` at `real_client.cpp:4624` passing 9
arguments, while the function is declared/defined with 7
(`real_client.h:579`, `real_client.cpp:3551`) — hence "candidate expects 7
arguments, 9 provided". Reproduces on a clean `origin/main` (113acf75).

This PR only touches `tent/src/metastore/http.cpp` and its header, so it's
unaffected functionally. A rebase should go green once #3234's call site is
fixed. cc @Cruz-Zhao (author of #3234).

zxpdemonio:
Thanks for tracing this. The follow-up fix, #3294, has now been merged into `main` as `a6b4db4c`.

The regression came from a semantic merge conflict between #3000 and #3234: #3000 reduced `get_into_ranges_internal` from nine arguments to seven, while the concurrently added CUDA IPC call site from #3234 retained the old nine-argument form. Git completed the textual merge without flagging that incompatibility.

Rebasing this PR onto the latest `main` should pick up the fix and unblock the build.

KubrickLiu:
Thanks  @alogfans  ~ Fixed the format issue and re-ran CI, all green now. PTAL.

KubrickLiu:
yufeng @he-yufeng, could you take a look and see if this change makes sense, Thanks~

alogfans (approved):
LGTM

PR #3261, opened 2026-08-03
[TENT] cuda_probe: zero-init cudaPointerAttributes and warn on libcud…

KubrickLiu:
## Description

Follow-up to the review on the cuda_probe stack-smashing report, split out
per request.

This does NOT paper over the crash with a local buffer pad. Instead it targets
the ABI root cause:
- Zero-initialize `cudaPointerAttributes attributes{}` (docs require reserved
  fields to be zero).
- On first probe, compare the build-time `CUDART_VERSION` against the running
  `cudaRuntimeGetVersion()`; if the major versions differ, log a clear,
  actionable error pointing at the toolchain mismatch.

Rationale: `cudaPointerGetAttributes` takes no size argument and writes as many
bytes as the *running* cudart's struct. A binary built against older CUDA
headers (e.g. CUDA 12, no `reserved[8]`) but run against CUDA 13 cudart writes
past the stack object and smashes the canary. The durable fix is to build
against a matching CUDA toolkit; this change makes that failure diagnosable
instead of a mysterious "stack smashing detected".

## Module

## Type of Change
- [x] Bug fix

### DIFF
--- a/mooncake-transfer-engine/tent/src/platform/cuda/cuda_probe.cpp
+++ b/mooncake-transfer-engine/tent/src/platform/cuda/cuda_probe.cpp
@@ -290,7 +290,7 @@ bool cudaProbeUsable() {
 }  // namespace
 
 MemoryType CudaPlatform::getMemoryType(void* addr) {
-    if (!cudaProbeUsable()) return MTYPE_CPU;
+    if (!cudaProbeUsable()) return MTYPE_UNKNOWN;
     cudaPointerAttributes attributes{};
     cudaError_t result = cudaPointerGetAttributes(&attributes, addr);
     if (result != cudaSuccess) {
--- a/mooncake-transfer-engine/tent/src/platform/cuda/cuda_probe.cpp
+++ b/mooncake-transfer-engine/tent/src/platform/cuda/cuda_probe.cpp
@@ -36,7 +36,6 @@
 #include <unordered_set>
 
 #include <algorithm>
-#include <mutex>
 
 namespace mooncake {
 namespace tent {
@@ -264,23 +263,34 @@ Status CudaPlatform::probe(std::vector<Topology::NicEntry>& nic_list,
 }
 
 namespace {
-void checkCudartAbiOnce() {
-    static std::once_flag once;
-    std::call_once(once, [] {
+bool cudaProbeUsable() {
+    static const bool usable = [] {
+        int device_count = 0;
+        if (cudaGetDeviceCount(&device_count) != cudaSuccess ||
+            device_count == 0) {
+            LOG(WARNING) << "No CUDA device detected; treating buffers as "
+                            "host memory";
+            return false;
+        }
         int runtime_version = 0;
-        if (cudaRuntimeGetVersion(&runtime_version) != cudaSuccess) return;
-        if (runtime_version / 1000 != CUDART_VERSION / 1000) {
+        // Major version only: struct layout changes across CUDA majors.
+        if (cudaRuntimeGetVersion(&runtime_version) == cudaSuccess &&
+            runtime_version / 1000 != CUDART_VERSION / 1000) {
             LOG(ERROR) << "CUDA ABI mismatch: built against CUDART "
                        << CUDART_VERSION << ", loaded libcudart is "
                        << runtime_version
-                       << "; rebuild against a matching CUDA toolkit.";
+                       << "; skipping pointer probe, rebuild against a "
+                          "matching CUDA toolkit.";
+            return false;
         }
-    });
+        return true;
+    }();
+    return usable;
 }
 }  // namespace
 
 MemoryType CudaPlatform::getMemoryType(void* addr) {
-    checkCudartAbiOnce();
+    if (!cudaProbeUsable()) return MTYPE_CPU;
     cudaPointerAttributes attributes{};
     cudaError_t result = cudaPointerGetAttributes(&attributes, addr);
     if (result != cudaSuccess) {
@@ -313,7 +323,10 @@ const std::vector<RangeLocation> CudaPlatform::getLocation(void* start,
     const static size_t kPageSize = 4096;
     std::vector<RangeLocation> entries;
 
-    checkCudartAbiOnce();
+    if (!cudaProbeUsable()) {
+        entries.push_back({(uint64_t)start, len, kWildcardLocation});
+        return entries;
+    }
     cudaPointerAttributes attributes{};
     cudaError_t result = cudaPointerGetAttributes(&attributes, start);
     if (result != cudaSuccess) {
--- a/mooncake-transfer-engine/tent/src/platform/cuda/cuda_probe.cpp
+++ b/mooncake-transfer-engine/tent/src/platform/cuda/cuda_probe.cpp
@@ -36,6 +36,7 @@
 #include <unordered_set>
 
 #include <algorithm>
+#include <mutex>
 
 namespace mooncake {
 namespace tent {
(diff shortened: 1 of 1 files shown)


mooncake-transfer-engine/tent/src/platform/cuda/cuda_probe.cpp
```diff
@@ -262,10 +263,26 @@ Status CudaPlatform::probe(std::vector<Topology::NicEntry>& nic_list,
     return Status::OK();
 }
 
+namespace {
+void checkCudartAbiOnce() {
```
KubrickLiu: 不在依赖 buf 字节做兼容，而且考虑准确报错，起码当编译版本和运行版本存在偏差时候，准确报出版本情况

staryxchen:
Thanks for splitting this out. One concern: this is diagnosis, not a fix.

  The zero-init + version check don't prevent the crash.

  - attributes{} only zeroes fields the compiler knows about at build time. If the running libcudart writes a larger struct (the ABI-mismatch case this PR assumes), it still writes past the stack object. Zero-init doesn't bound the runtime write.
  - checkCudartAbiOnce() only LOG(ERROR)s on mismatch — no early return, no abort. Code then proceeds straight into cudaPointerGetAttributes(), the very call  that smashes the stack. We get a log line, then still crash.

  The root-cause hypothesis is unverified.

  #3202's repro environment was "CUDA 13.0" with no mention of build/runtime mismatch. Mooncake's CI builds against CUDA 13 (see #if CUDART_VERSION >= 13000 guards in nvlink_transport.cpp, mooncake_ep_elastic_compiled.cuh). A CUDA-13 build against a CUDA-13 runtime passes this check silently — 13 == 13, no log — and the crash still reproduces. If that's what happened, #3202's padding was the actual defense, and this PR replaces it with a diagnostic that never fires.

  Could you share the actual values from the crashing environment?

  - Build-time CUDART_VERSION (macro value)
  - Runtime cudaRuntimeGetVersion() return value

  If they differ, the hypothesis holds — but please add an early-return so we don't fall through into the crashing call.

he-yufeng:
Read through both hunks. This is the right shape for the stack-smashing report: zero-init only helps the direction where the loaded libcudart writes fewer fields than the build-time struct (uninitialized reads of type/memoryType), and the one-time version check is what actually guards the reported direction (a newer runtime writing a larger struct past the local). Good that the description says this out loud instead of claiming the init fixes the crash.

Checked the file: these are the only two cudaPointerGetAttributes call sites, so coverage is complete.

Two small things:

1. The `/ 1000` major comparison is the right granularity for current CUDA versioning (12.8 -> 12080), and this struct has only changed layout across majors historically. Worth a one-line comment stating the encoding assumption, so nobody later "simplifies" it into a raw equality that would false-positive on every minor bump.

2. Merge-order heads-up: my own #2955 (skip the probe on GPU-less hosts) touches `getMemoryType` a few lines above this hunk, so whichever lands second needs a trivial rebase. Not an argument against either change, just flagging it for whoever merges first.

No objection from me. Leaving the approve to someone who can exercise the CUDA path on hardware.

KubrickLiu:
Thanks!

- No git conflict with #2955 — that changes `getMemoryLocation()` in the legacy
  `src/memory_location.cpp`, this touches `cuda_probe.cpp` (tent). I mirrored
  your device-presence guard and added the same no-GPU skip to the tent path.
- Kept the `/1000` major-only comparison and noted why in a comment.

@he-yufeng

KubrickLiu:
Thanks for pushing on this — both points were right. 

I went back and measured the failing setup:
- Built against CUDART 12080 (the build host's default `/usr/local/cuda` → 12.8),
  where `sizeof(cudaPointerAttributes)` is 24.
- Run against libcudart 13 (`cudaRuntimeGetVersion` = 13000), where the struct
  is 88 bytes (CUDA 13 added `long reserved[8]`).

So the runtime wrote 88 bytes into a 24-byte stack object — the ABI mismatch you
suspected. And you're right that a 13-built + 13-run binary wouldn't hit this and
the check stays silent there, which is the intended behavior.

On the early-return: agreed, logging alone still fell into the crashing call.
I reworked it into a guard that returns before probing when the major version
differs (or no GPU is present), so we no longer enter `cudaPointerGetAttributes`
in the mismatched case. The durable fix is still pinning the build to a matching
toolkit — this just makes the failure diagnosable and safe instead of a silent
smash.

@staryxchen

mooncake-transfer-engine/tent/src/platform/cuda/cuda_probe.cpp
```diff
@@ -262,10 +262,37 @@ Status CudaPlatform::probe(std::vector<Topology::NicEntry>& nic_list,
...
+    }();
+    return usable;
+}
+}  // namespace
+
 MemoryType CudaPlatform::getMemoryType(void* addr) {
-    cudaPointerAttributes attributes;
-    cudaError_t result;
-    result = cudaPointerGetAttributes(&attributes, addr);
+    if (!cudaProbeUsable()) return MTYPE_CPU;
```
staryxchen: When cudaProbeUsable() fails on CUDA major-version ABI mismatch, getMemoryType still returns MTYPE_CPU even though GPUs remain usable, so device memory can be misrouted through host paths and risk corruption/crashes. I think it's better to return MTYPE_UNKNOWN when CUDA major ABI mismatch.

mooncake-transfer-engine/tent/src/platform/cuda/cuda_probe.cpp
```diff
@@ -296,10 +323,12 @@ const std::vector<RangeLocation> CudaPlatform::getLocation(void* start,
     const static size_t kPageSize = 4096;
     std::vector<RangeLocation> entries;
 
-    cudaPointerAttributes attributes;
-    cudaError_t result;
-
-    result = cudaPointerGetAttributes(&attributes, start);
+    if (!cudaProbeUsable()) {
```
alogfans: Could you remember the result in the first time to avoid duplicated execution?

mooncake-transfer-engine/tent/src/platform/cuda/cuda_probe.cpp
KubrickLiu: really appreciate the careful review. You're
right, MTYPE_UNKNOWN is the more appropriate choice here: returning MTYPE_CPU on
an ABI mismatch could misroute device memory through the host path even though
the GPU is still usable. I've updated getMemoryType() to return MTYPE_UNKNOWN in
that case. Thanks ～～ 
`if (!cudaProbeUsable()) return MTYPE_UNKNOWN;`

mooncake-transfer-engine/tent/src/platform/cuda/cuda_probe.cpp
KubrickLiu: Sure, it's already cached via a function-local `static`:

```cpp
bool cudaProbeUsable() {
    static const bool usable = [] {
        // cudaGetDeviceCount / cudaRuntimeGetVersion run once here
        ...
    }();
    return usable;
}
```

The lambda runs only on the first call; later calls just read the cached bool.

mooncake-transfer-engine/tent/src/platform/cuda/cuda_probe.cpp
```diff
@@ -262,10 +262,37 @@ Status CudaPlatform::probe(std::vector<Topology::NicEntry>& nic_list,
...
+    }();
+    return usable;
+}
+}  // namespace
+
 MemoryType CudaPlatform::getMemoryType(void* addr) {
-    cudaPointerAttributes attributes;
-    cudaError_t result;
-    result = cudaPointerGetAttributes(&attributes, addr);
+    if (!cudaProbeUsable()) return MTYPE_UNKNOWN;
```
staryxchen: Here still cannot distinguish between “no-GPU” and “ABI mismatch.” It would be more appropriate for the former to return MTYPE_CPU.

KubrickLiu: I've split the two cases:
﻿
No GPU: getMemoryType now returns MTYPE_CPU, and getLocation skips the pointer probe and falls through to host NUMA analysis (concrete cpu:<node> instead of wildcard) — a GPU-less host can only hold host memory.
ABI mismatch: still MTYPE_UNKNOWN / wildcard, since the probe is genuinely unreliable there.
The probe is now two helpers, cudaDevicePresent() and cudaAbiMatches(), so the cases are distinguished at every call site. 


This also aligns with how #2955 handles GPU-less hosts (probe skipped, treated as host memory). PTAL.

staryxchen (approved):
LGTM

KubrickLiu:
Gentle ping — CI is green and this has one approval (staryxchen). Would appreciate a code-owner review to move this forward. Thanks ~~ @alogfans

PR #3263, opened 2026-08-03
[Bugfix][TE] Keep context_list_ index-aligned when an RNIC fails to init

SongOf:
Description
  
  Topology::resolve() gives each HCA an id equal to its position in getHcaList(), and disableDevice() deliberately keeps the surviving ids there rather than re-running resolve() — its comment notes that compacting would make the lkey/rkey and context arrays point at the wrong RNIC.

  initializeRdmaResources() broke that contract from the other side: it appended to context_list_ only for devices whose construct() succeeded, so one failed RNIC compacted the array while the topology kept the sparse ids. A device_id from selectDevice() then either named a different NIC (silently defeating the computed NUMA/PCIe affinity) or ran past the end of context_list_ — an unchecked out-of-bounds read, since the request-level lookup in submitTransferTask() has no bounds check and the per-slice one is only an assert. Reachable on any multi-NIC host where a NIC fails construct(): port not IBV_PORT_ACTIVE for the global MC_IB_PORT, no usable GID, or a failed DMA-BUF check.

  Fix: push a fresh, never-constructed RdmaContext as a placeholder instead of dropping the slot. It owns no ibv_context/PD, reports itself inactive, and no-ops registration, so the existing active() checks already handle it; the partially built context is released so a failed RNIC does not pin an open uverbs fd. Five entry points are guarded against a null pd_ (both registerMemoryRegion() overloads, preTouchMemory(), lkey(), rkey()) because ibv_reg_mr(nullptr, ...) segfaults. No hot-path changes — disableDevice() already keeps disabled ids out of selectDevice().

  Follow-up (separate PR): TENT has the same defect — context_set_ and BufferDesc::lkey are both compacted while dev_id stays a sparse topology NicID, with a wider trigger (any type != NIC_RDMA entry) and no disableDevice() equivalent.

  Module
  
  - [ ] CI/CD
  - [ ] Docs
  - [ ] Other
  
  Type of Change

  - [x] Bug fix
  - [ ] New feature
  - [ ] Refactor
  - [ ] Breaking change
  - [ ] Documentation update
  - [ ] Performance improvement
  - [ ] Other

  How Has This Been Tested?

  Two tests added to rdma_context_reprobe_test.cpp, reusing its libibverbs symbol interposition so no RDMA hardware is needed:

  - ContextListKeepsOneSlotPerHcaWhenInitFails — drives initializeRdmaResources() and asserts slot count, per-slot device name, and inactive placeholders. Fails before this change with contexts.size()=0 vs hca_list.size()=3.
  - DisabledDeviceKeepsRemainingHcaIndexesStable — pins the disableDevice()/selectDevice() contract the fix relies on.

  Test commands:
  cmake -S . -B build -DBUILD_UNIT_TESTS=ON -DWITH_TE=ON && cmake --build build -j"$(nproc)"
  ./build/mooncake-transfer-engine/tests/rdma_context_reprobe_test
  cd build && ctest --output-on-failure

  Test results:
  - rdma_context_reprobe_test: 7/7 pass (5 pre-existing + 2 new).
  - ctest: 28/31. The 3 failures — tcp_transport_test, transfer_metadata_test (Unable to find metadata storage plugin etcd), memory_location_test (mbind: Operation not permitted) — are environment-only and reproduce identically
  on the unpatched baseline, verified by stashing the change and re-running.
  - ./scripts/code_format.sh --check: all four files OK (clang-format 20.1.8). pre-commit on the touched files: all hooks pass, no rewrites.

  Coverage gap: the tests cover the "all devices fail" path. The mixed case is not covered end-to-end (making construct() succeed without hardware needs faked comp channels, epoll fds, CQs and a live WorkerPool); the two tests instead pin both halves of the invariant — context_list_.size() == getHcaList().size(), and every selectDevice() id falls inside that range.

  - [x] Unit tests pass
  - [ ] Integration tests pass (if applicable) — not run; needs RDMA hardware + metadata server

  Manual verification = the baseline comparison above: new test confirmed failing before the fix and passing after, with the 3 environment failures identical either way.

  Checklist

  - [ ] For changes >500 LOC: I have filed an RFC issue — n/a, 95 lines

Unchecked on purpose: self-review is the submitter's own attestation; --all-files was not run (only pre-commit run --files on the four touched files, all passing) because AGENTS.md warns it may rewrite unrelated files.

  AI Assistance Disclosure

  - [ ] No AI tools were used

### DIFF
--- a/mooncake-transfer-engine/src/transport/rdma_transport/rdma_context.cpp
+++ b/mooncake-transfer-engine/src/transport/rdma_transport/rdma_context.cpp
@@ -606,6 +606,8 @@ int RdmaContext::registerMemoryRegionInternal(void *addr, size_t length,
 
 int RdmaContext::registerMemoryRegion(void *addr, size_t length, int access,
                                       const DmabufExport &exp) {
+    // Placeholder context for a failed RNIC: no PD to register against.
+    if (!pd_) return 0;
     MemoryRegionMeta mrMeta;
     int ret = registerMemoryRegionInternal(addr, length, access, exp, mrMeta);
     if (ret != 0) {
@@ -617,6 +619,7 @@ int RdmaContext::registerMemoryRegion(void *addr, size_t length, int access,
 }
 
 int RdmaContext::registerMemoryRegion(void *addr, size_t length, int access) {
+    if (!pd_) return 0;  // placeholder context: skip the dma_buf export too
     // Single-NIC convenience path: export, register, and close the fd here.
     // The shared-fd benefit only matters when a buffer is registered against
     // multiple NICs (see RdmaTransport::registerLocalMemoryInternal).
@@ -645,6 +648,7 @@ int RdmaContext::unregisterMemoryRegion(void *addr) {
 }
 
 int RdmaContext::preTouchMemory(void *addr, size_t length) {
+    if (!pd_) return 0;  // placeholder context
     DmabufExport exp;
     int ret = exportDmabuf(addr, exp);
     if (ret != 0) {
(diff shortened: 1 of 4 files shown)


staryxchen (approved):
LGTM

Issue #14, opened 2024-12-03
Questions about thread pool and shard


CharlesJhonson:
Thanks for sharing the paper and code of this project! I have a question about the code:
What is the purpose of setting shard when there is already a worker_pool? What is the basis for setting the number to 8?

alogfans:
The shard mechanism is designed to evenly distribute the user's RDMA requests among the different threads of the worker pool. In addition, if context_.cqCount() is greater than one, the shard mechanism also mitigates concurrent contention for some RDMA resources (QP and doorbell), which has more significant performance impact in small I/Os.

Setting the number to 8 is based on our experience.

CharlesJhonson:
Thank you for your answer.
 If we want to avoid competition between QPs, shouldn't we consider assigning a CQ to each QP?

alogfans:
Yes, so Mooncake supports allocating multiple CQs per device, which are assigned to QPs in a round-robin manner.

CharlesJhonson:
Ok , thank you very much!

PR #3264, opened 2026-08-03
[Bugfix][TENT] Keep context_set_ and buffer keys NicID-indexed

SongOf:
Description

  A device id in TENT is a Topology NicID — DeviceSelector::devices_ is keyed by it, MemEntry::device_list holds them, and both feed slice->source_dev_id / target_dev_id. Two arrays that this id subscripts were built compacted instead:

  - install() appended to context_set_ only for entries that were NIC_RDMA and whose construct() succeeded.
  - addBufferInternal() push_back'd into BufferDesc::lkey/rkey only for slots holding a context.

  One skipped NIC shifts every higher id: context_set_[dev_id] names a different RNIC, or runs past the end — an out-of-bounds read on a vector<shared_ptr> immediately followed by ->status(). The shift also applies remotely, since rkey is published in the segment descriptor and the peer subscripts it with a NicID from our device_list.

  Unlike the classic TE, TENT has no disableDevice() equivalent — the "Disable RDMA device" warning only skips a push_back, so a failed device stays in the topology and is still handed out. The non-RDMA branch needs no failure at all: a single NIC_TCP or type-less entry in the topology JSON shifts the indexes.

  Fix: give every NicID a slot. Skipped NICs hold an inert RdmaContext (DEVICE_UNINIT, or DEVICE_DISABLED after a failed construct), which consumers already reject via status(), so the hot path gains no checks. lkey/rkey are sized to context_list_ and filled by index, leaving a zero key in the gaps that no context backs. Three sites unsafe against an inert context are guarded: RdmaContext::cq() (dereferenced a null params_), Workers::asyncPollCq() (cq->poll() unchecked), and the monitorThread() reclaim tick (endpointStore()->reclaim() — an inert context never built a store, and this only fires on the first 1 Hz heartbeat, so it crashes well after a clean startup). getEndpoint() took context_set_[0], which may now be inert; it takes the first enabled context instead.

  The install loop and reclaim tick are extracted into initializeContexts() and reclaimEndpoints() so tests can drive them without a metadata service or worker threads.

  Wire format is unchanged in a healthy cluster — with all NICs up the sparse and compacted vectors are identical — and a longer rkey vector fixes rather than breaks an unpatched peer's lookup.

  Counterpart to the classic-TE fix ([Bugfix][TE] Keep context_list_ index-aligned when an RNIC fails to init); same defect class, different mechanism.

  Module

  - [ ] CI/CD
  - [ ] Docs
  - [ ] Other
  
  Type of Change

  - [x] Bug fix
  - [ ] New feature
  - [ ] Refactor
  - [ ] Breaking change
  - [ ] Documentation update
  - [ ] Performance improvement
  - [ ] Other

  How Has This Been Tested?

  Three tests in tent/tests/rdma_transport_test.cpp, two of them verified against the unpatched code:

  - ContextSetKeepsOneSlotPerNonRdmaNic — the NIC_RDMA branch, which never touches libibverbs, so it runs anywhere. Reverting the push_back ordering makes it fail.
  - ReclaimTickSkipsInertContexts — asserts endpointStore() is null on every inert slot, then runs the tick. Removing the guard makes the binary exit 139 (SIGSEGV).
  - ContextSetKeepsOneSlotWhenConstructFails — the construct()-failure branch. Self-skips without RDMA hardware: IbvLoader dlcloses libibverbs when no device is present while keeping its symbol table, so construct() cannot be driven safely in that state.

  Test commands:
  cmake --build build-tent -j2
  ./build-tent/mooncake-transfer-engine/tent/tests/tent_rdma_transport_test \
        --gtest_filter='RdmaNicIndexAlignmentTest.*'
  cd build-tent && ctest --output-on-failure

  Test results:
  - tent_rdma_transport_test: 2 pass, 1 skipped (no RDMA hardware).
  - ctest: 60/63. tcp_transport_test and transfer_metadata_test abort on a missing etcd metadata plugin; memory_location_test on mbind EPERM. All three fail identically without this change.
  - ./scripts/code_format.sh --check: 7 files OK (clang-format 20.1.8). pre-commit on the touched files: all hooks pass, no rewrites.
  - [x] Unit tests pass
  - [ ] Integration tests pass (if applicable) — not run; needs RDMA hardware

  Manual verification = the two revert experiments above (test fails / segfaults without the fix, passes with it).

  Checklist

  - [ ] For changes >500 LOC: I have filed an RFC issue — n/a, 188 lines

Unchecked on purpose: self-review is the submitter's own attestation; --all-files was not run (only pre-commit run --files on the 7 touched files, all passing) because AGENTS.md warns it may rewrite unrelated files.
  
  Known gap, not addressed here: workers.cpp subscripts a peer's rkey vector without a bounds check, so an unpatched peer can still drive an out-of-bounds read during a rolling upgrade. Pre-existing; needs its own hardening, in the spirit of #3209 on the classic TE.

  AI Assistance Disclosure
  
  - [ ] No AI tools were used

### DIFF
--- a/mooncake-transfer-engine/tent/include/tent/transport/rdma/rdma_transport.h
+++ b/mooncake-transfer-engine/tent/include/tent/transport/rdma/rdma_transport.h
@@ -56,6 +56,7 @@ struct RdmaSubBatch : public Transport::SubBatch {
 class RdmaTransport : public Transport {
     friend class Workers;
     friend class RdmaEndPoint;
+    friend class RdmaTransportTestPeer;
 
    public:
     RdmaTransport();
@@ -122,6 +123,11 @@ class RdmaTransport : public Transport {
     Status setupLocalSegment();
 
     std::shared_ptr<Config> config() const { return conf_; }
+
+   private:
+    // Builds context_set_ with one slot per NicID; returns how many RNICs
+    // came up. Remaining slots hold inert contexts.
+    size_t initializeContexts();
 
    private:
     bool installed_;
--- a/mooncake-transfer-engine/tent/include/tent/transport/rdma/workers.h
+++ b/mooncake-transfer-engine/tent/include/tent/transport/rdma/workers.h
@@ -38,6 +38,8 @@ class RdmaTransport;
 class DeviceSelector;
 
 class Workers {
+    friend class RdmaTransportTestPeer;
+
    public:
     static constexpr size_t kCapacity = 1024 * 8;
     using BoundedSliceQueue = BoundedMPSCQueue<RdmaSliceList, kCapacity>;
@@ -74,6 +76,11 @@ class Workers {
     void releaseSliceQuota(RdmaSlice* slice, double latency = 0.0);
 
     void monitorThread();
+
+    // 1 Hz heartbeat from monitorThread(): drains every context's retiring
+    // endpoints so reclaim is not gated on new insertions, which stall under
+    // failure load.
+    void reclaimEndpoints();
 
     int handleContextEvents(std::shared_ptr<RdmaContext>& context);
 
--- a/mooncake-transfer-engine/tent/src/transport/rdma/buffers.cpp
+++ b/mooncake-transfer-engine/tent/src/transport/rdma/buffers.cpp
@@ -141,6 +141,11 @@ Status LocalBufferManager::addBufferInternal(BufferDesc& desc,
                 context->registerMemReg((void*)desc.addr, desc.length, access);
         }
     }
+    // NicID-keyed like context_list_, not compacted: slice dispatch subscripts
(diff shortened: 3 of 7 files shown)


alogfans:
```
14/117 Test  #14: rdma_async_event_drain_test ..................***Failed    0.59 sec
```

Is this CI error related to this PR?

SongOf:
> ```
> 14/117 Test  #14: rdma_async_event_drain_test ..................***Failed    0.59 sec
> ```
> 
> Is this CI error related to this PR?

No. It looks like rdma_async_event_drain_test is intermittently failing, and I'm going to try to fix this test. @alogfans

alogfans (approved):
LGTM

PR #3266, opened 2026-08-04
test: poll for the async eviction in PutStartExpiringTest

### LINKED ISSUE #3245: [Store][Test] PutStartExpiringTest can race with asynchronous eviction under coverage
## Summary

`MasterServiceTest.PutStartExpiringTest` failed in PR #3201 because the test retries `PutStart()` after a fixed 50 ms sleep, while the memory eviction that releases expired/discarded replicas runs asynchronously.

In the failed CI run, the eviction cycle took approximately 59 ms. The retry therefore happened before eviction completed and before capacity became available.

This appears to be a pre-existing timing-sensitive test rather than a functional regression introduced by PR #3201.

## Observed failure

Related PR: https://github.com/kvcache-ai/Mooncake/pull/3201

Failed CI job:

https://github.com/kvcache-ai/Mooncake/actions/runs/30544794003/job/90903216242

Environment:

- GitHub Actions Ubuntu 22.04 runner
- Python 3.12 build job
- Coverage instrumentation enabled through `--coverage`
- 109 CTest targets executed
- 108 passed
- Only `master_service_test` failed
- Inside that binary, 145 of 146 tests passed

Failure:

```text
[ RUN      ] MasterServiceTest.PutStartExpiringTest

/home/runner/work/Mooncake/Mooncake/mooncake-store/tests/master_service_test.cpp:5472: Failure
Value of: put_start_result.has_value()
  Actual: false
Expected: true

unknown file: Failure
C++ exception with description "Bad expected access" thrown in the test body.
```

The `Bad expected access` exception is secondary. The test uses a non-fatal
`EXPECT_TRUE()` and then immediately calls `.value()` on the failed result.

## Test scenario

`PutStartExpiringTest` configures:

- `put_start_discard_timeout_sec = 3`
- `put_start_release_timeout_sec = 5`
- Three memory segments of 16 MiB each
- Object size of 6 MiB
- Three replicas per object

After the replicas of `key_2` become eligible for discard/release, the test:

1. Calls `PutStart(key_2)` and expects `NO_AVAILABLE_HANDLE`.
2. This failed allocation triggers asynchronous memory eviction.
3. Sleeps for a fixed 50 ms.
4. Calls `PutStart(key_2)` again and expects success.

Relevant test code is around:

`mooncake-store/tests/master_service_test.cpp:5460-5473`

## CI timing evidence

The failed job contains the following sequence:

```text
15:49:21.504084 [EVICT-TRIGGER] memory_ratio=0.75 ... need_mem_eviction=1
15:49:21.563361 [EVICT-RESULT] evicted_count=0 ...
15:49:21.563400 [EVICT-DONE] BatchEvict execution completed.
```

The eviction cycle took approximately 59.3 ms from trigger to completion.
This already exceeds the test's fixed 50 ms wait.

The assertion failed before the `EVICT-RESULT` and `EVICT-DONE` messages were
emitted, which is consistent with the second `PutStart()` running before
discarded replicas had been released.

The test does not print the error returned by the second `PutStart()`.
`NO_AVAILABLE_HANDLE` is the most likely result based on the allocation path,
but this should be confirmed by improving the assertion diagnostics.

## Relevant implementation path

When allocation cannot find enough memory replicas,
`AllocateAndInsertMetadata()`:

1. Sets `need_mem_eviction_ = true`.
2. Returns `ErrorCode::NO_AVAILABLE_HANDLE`.

The background eviction thread checks this flag asynchronously. It sleeps for
up to 10 ms between checks and then calls `BatchEvict()`.

`BatchEvict()` scans all 1024 metadata shards, using up to 16 worker threads,
joins those workers, and only later calls
`ReleaseExpiredDiscardedReplicas(now)`.

There is no synchronization between the test and completion of this eviction
cycle. The test therefore assumes that thread wake-up, shard scanning, and
discarded-replica release will all finish within 50 ms.

Coverage instrumentation and CI load make that assumption unreliable.

The `evicted_count=0` log line does not show whether discarded replicas were
released. That count is separate from the discarded-replica release count, so
it should not be interpreted as evidence that no memory was released.

## Relationship to PR #3201

PR #3201 changes only the oplog batch storage, codec, and snapshot tests:

- `oplog_batch_storage.h/.cpp`
- `oplog_batch_types.h`
- `oplog_batch_codec.cpp`
- Oplog batch storage/codec tests
- `snapshot_child_process_test.cpp`

It does not modify:

- `master_service.cpp`
- `master_service.h`
- `master_service_test.cpp`
- The allocation or eviction implementation

Other relevant tests, including `master_service_ha_test` and all modified
oplog batch tests, passed in the same CI job.

Therefore, the available evidence indicates that PR #3201 exposed an existing
timing-sensitive test rather than introducing the failure directly.

## Local result

The test passed five consecutive iterations in a local non-coverage build:

```bash
LSAN_OPTIONS=detect_leaks=0 \
  --gtest_filter=MasterServiceTest.PutStartExpiringTest \
  --gtest_repeat=5 \
  --gtest_break_on_failure=0
```

Local eviction cycles completed in approximately 6-14 ms, so the fixed 50 ms
sleep happened to be sufficient. This difference further supports a
load-sensitive race rather than a deterministic functional failure.

## Suggested fix

Replace the fixed 50 ms sleep with a bounded condition-based retry:

- Reuse the existing `MasterServiceTest::WaitUntil()` helper.
- Retry `PutStart()` until it succeeds or the timeout expires.
- Treat `NO_AVAILABLE_HANDLE` as the expected transient result.
- Fail immediately on any unexpected error.
- Preserve the successful result instead of calling `PutStart()` again.
- Use `ASSERT_TRUE()` before accessing `.value()`.
- Include the returned error code in failure diagnostics.

Increasing the sleep to another fixed value would only reduce the probability
of failure and would keep the test timing-dependent.

A production change to reorder work inside `BatchEvict()` may be worth
evaluating separately, but it is not required to make this test deterministic.

## Acceptance criteria

- `PutStartExpiringTest` waits for observable eviction completion instead of a
  fixed delay.
- A failed retry reports the actual returned error.
- The test never calls `.value()` after a non-fatal failed assertion.
- The test passes repeatedly with coverage instrumentation and under CI load.
- No production eviction behavior is changed solely to accommodate the test.

### Before submitting...

he-yufeng:
Fixes #3245.

The retry after the eviction-triggering `PutStart` slept a fixed 50 ms, and the CI failure shows the asynchronous release taking about 59 ms on a loaded runner, so `put_start_result.has_value()` was false and `.value()` threw. The block now polls `PutStart` every 10 ms with a 5 s deadline: it converges the moment the eviction finishes instead of betting on a duration that varies with runner load.

Verification: the repo's existing build directory points at a container path and cannot regenerate here, so a full CTest run was not possible locally. The poll shape was compiled standalone with `-Wall -Wextra` against a stubbed service and converges. CI runs the real suite.

### DIFF
--- a/mooncake-store/tests/master_service_test.cpp
+++ b/mooncake-store/tests/master_service_test.cpp
@@ -5512,13 +5512,17 @@ TEST_F(MasterServiceTest, PutStartExpiringTest) {
     EXPECT_FALSE(put_start_result.has_value());
     EXPECT_EQ(put_start_result.error(), ErrorCode::NO_AVAILABLE_HANDLE);
 
-    // Wait a moment for the eviction to complete.
-    std::this_thread::sleep_for(std::chrono::milliseconds(50));
-
-    // Put key_2 again, should success because the previous one has been
-    // discarded and released.
-    put_start_result = service_->PutStart(client_id, key_2, TenantId::Default(),
-                                          slice_length, config);
+    // The eviction runs asynchronously, so poll for it instead of sleeping a
+    // fixed 50 ms that a loaded CI runner can overrun (it took ~59 ms in the
+    // failure from #3245).
+    const auto eviction_deadline =
+        std::chrono::steady_clock::now() + std::chrono::seconds(5);
+    do {
+        std::this_thread::sleep_for(std::chrono::milliseconds(10));
+        put_start_result = service_->PutStart(
+            client_id, key_2, TenantId::Default(), slice_length, config);
+    } while (!put_start_result.has_value() &&
+             std::chrono::steady_clock::now() < eviction_deadline);
     EXPECT_TRUE(put_start_result.has_value());
     replica_list = put_start_result.value();
     EXPECT_EQ(replica_list.size(), kReplicaCnt);
--- a/mooncake-store/tests/master_service_test.cpp
+++ b/mooncake-store/tests/master_service_test.cpp
@@ -5506,7 +5506,13 @@ TEST_F(MasterServiceTest, PutStartExpiringTest) {
     }
 
     // Put key_2 again, should fail because eviction has not been triggered. And
-    // this PutStart should trigger the eviction.
+    // this PutStart should trigger the eviction. Only BatchEvict moves the
+    // eviction attempt counter, so take the baseline before the trigger: the
+    // eviction thread polls every 10 ms, and sampling after the failing
+    // PutStart could race a completed BatchEvict and wait for a second one
+    // that never comes.
+    const int64_t eviction_attempts_before =
+        MasterMetricManager::instance().get_mem_eviction_attempts();
     put_start_result = service_->PutStart(client_id, key_2, TenantId::Default(),
                                           slice_length, config);
     EXPECT_FALSE(put_start_result.has_value());
@@ -5517,17 +5523,16 @@ TEST_F(MasterServiceTest, PutStartExpiringTest) {
     // to put_start_release_timeout_sec cannot tell that path apart from the
     // periodic DiscardExpiredProcessingReplicas fallback, which releases the
     // same replicas on the same 5 s scale and would pass the test without
-    // exercising the immediate eviction. Only BatchEvict moves the eviction
-    // attempt counter, so wait on that and then retry once.
-    const int64_t eviction_attempts_before =
-        MasterMetricManager::instance().get_mem_eviction_attempts();
+    // exercising the immediate eviction, and the periodic path never touches
+    // the attempt counter.
     WaitUntil([&] {
         return MasterMetricManager::instance().get_mem_eviction_attempts() >
                eviction_attempts_before;
     });
     put_start_result = service_->PutStart(client_id, key_2, TenantId::Default(),
                                           slice_length, config);
-    ASSERT_TRUE(put_start_result.has_value());
+    ASSERT_TRUE(put_start_result.has_value())
+        << toString(put_start_result.error());
     replica_list = put_start_result.value();
     EXPECT_EQ(replica_list.size(), kReplicaCnt);
     for (size_t i = 0; i < kReplicaCnt; i++) {
--- a/mooncake-store/tests/master_service_test.cpp
+++ b/mooncake-store/tests/master_service_test.cpp
@@ -5512,18 +5512,22 @@ TEST_F(MasterServiceTest, PutStartExpiringTest) {
     EXPECT_FALSE(put_start_result.has_value());
     EXPECT_EQ(put_start_result.error(), ErrorCode::NO_AVAILABLE_HANDLE);
 
-    // The eviction runs asynchronously, so poll for it instead of sleeping a
-    // fixed 50 ms that a loaded CI runner can overrun (it took ~59 ms in the
-    // failure from #3245).
-    const auto eviction_deadline =
-        std::chrono::steady_clock::now() + std::chrono::seconds(5);
-    do {
-        std::this_thread::sleep_for(std::chrono::milliseconds(10));
-        put_start_result = service_->PutStart(
-            client_id, key_2, TenantId::Default(), slice_length, config);
-    } while (!put_start_result.has_value() &&
-             std::chrono::steady_clock::now() < eviction_deadline);
-    EXPECT_TRUE(put_start_result.has_value());
+    // The failed PutStart above sets need_mem_eviction_, and the eviction
+    // thread answers with an asynchronous BatchEvict. Polling PutStart for up
+    // to put_start_release_timeout_sec cannot tell that path apart from the
+    // periodic DiscardExpiredProcessingReplicas fallback, which releases the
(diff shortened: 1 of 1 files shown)


Icedcoco:
Hi, @he-yufeng . Thanks for addressing this flake. Polling instead of relying on a fixed 50 ms sleep makes sense. I had two questions about the current approach:

1. The polling deadline is 5 seconds, which is also the configured `put_start_release_timeout_sec`. Could the periodic cleanup path in `EvictionThreadFunc()` release the discarded replicas within that window even if the `need_mem_eviction_`-triggered path did not work? If so, the test might still pass without exercising the immediate eviction behavior it intends to verify. There is already `MasterServiceTest::WaitUntil()`, as well as the `get_mem_eviction_attempts()` metric. Would it make sense to record the attempt count, wait for it to increase, and then retry `PutStart()` once? Or do you see a reason to prefer polling `PutStart()` directly here?

2. If the polling still times out, `EXPECT_TRUE()` is non-fatal, so the following `.value()` would still throw `Bad expected access`, which could hide the original error. The loop also retries errors other than `NO_AVAILABLE_HANDLE`. Would you consider using `ASSERT_TRUE()` before accessing the value and failing early with the actual error when it is not the expected transient `NO_AVAILABLE_HANDLE`?

Curious to hear your thoughts.

he-yufeng:
@Icedcoco both points were right, thanks for the careful read. Reworked in cc1b091a:

1. You caught the real hole in my first version: a 5 s PutStart poll sits exactly on `put_start_release_timeout_sec`, and the periodic `DiscardExpiredProcessingReplicas` + `ReleaseExpiredDiscardedReplicas` path releases the same replicas on that scale, so the test could pass without the `need_mem_eviction_` path ever running. I took your suggestion almost verbatim: record `get_mem_eviction_attempts()`, fail `PutStart` to set `need_mem_eviction_`, `WaitUntil` the counter moves, retry once. Only `BatchEvict` increments that counter (success and fail alike, `master_service.cpp:8742-8753`), while the periodic fallback never touches it, and the 4 s `WaitUntil` window sits well under the 5 s fallback, so the test now fails if the immediate eviction path is broken instead of silently passing.

2. Done. The loop is gone, so there is no blanket retry of unrelated errors anymore, and the single retry is `ASSERT_TRUE` before `.value()`, so a genuine failure reports its own error instead of dying on bad expected access.

One judgment call to flag: if the first `BatchEvict` runs but releases nothing (the `inc_mem_eviction_fail` branch also moves the counter), the single retry fails the test. That matches the original single-retry design and this scenario is built so the discarded replicas are the eviction candidates, so I preferred determinism over another retry layer. Happy to revisit if you have seen that path flap in practice.

On verification: no local build tree here (the repo build dirs point at container paths), so the metric-gated shape is validated by a standalone g++ -Wall -Wextra -Werror syntax stub plus the source-level chain above; relying on CI for the real suite, same as before.

Icedcoco:
Thanks for reworking this. Waiting on the eviction attempt counter looks much better and avoids confusing the triggered eviction with the periodic cleanup path.

I noticed two smaller details and wanted to get your thoughts:

1. `eviction_attempts_before` is currently captured after the `PutStart()` that sets `need_mem_eviction_`. Could the eviction thread complete `BatchEvict()` and increment the counter between those two operations? If that happens, the test would record the incremented value as its baseline and wait for another eviction attempt that may never occur. Would it be safer to capture the counter immediately before the eviction-triggering `PutStart()`?

2. Changing the retry check to `ASSERT_TRUE()` prevents the secondary `Bad expected access`, which addresses the original failure mode. However, if `PutStart()` still fails, the assertion does not report the returned `ErrorCode`. Would you consider including `put_start_result.error()` in the assertion message so that a future failure shows the actual cause directly?

Does that match your reading of the synchronization here?

he-yufeng:
@Icedcoco both correct again, thanks. Fixed in 73b20126:

1. The baseline is now taken immediately before the eviction-triggering `PutStart()`. You called the race exactly right: the eviction thread polls every 10 ms, so a completed `BatchEvict` between the failing `PutStart` and the counter read would land in the baseline and the wait would then expect a second eviction that never comes. In this scenario nothing else can move the counter in between (the only writer is `BatchEvict`, and nothing has set `need_mem_eviction_` yet), so before-trigger is the correct sample point.

2. Done, `ASSERT_TRUE(...has_value()) << toString(...error())`, matching the convention already used elsewhere in this file.

Icedcoco:
Looks like the issue I raised earlier has been resolved. We can proceed once CI passes.

he-yufeng:
The one red job is infra, not the test: `build-wheel-efa` dies while downloading the Go toolchain (`proxy.golang.org` returned 403 Forbidden), before any compilation starts. The change only touches `master_service_test.cpp`, and the suite itself is unaffected. Should go green on a re-run.

ykwd (approved):
Thanks for this great work!

PR #3248, opened 2026-08-03
[Bugfix] Use GPU_PREFIX for HCA peer affinity under USE_HIP

staryxchen:
## Description

`Topology::precomputeResolvedHcaPeerAffinity` previously filtered topology entries with a hardcoded `"cuda:"` prefix. Under `USE_HIP`, discovered GPU locations are `hip:N`, so HCA peer affinity (`MC_ENABLE_HCA_PEER_AFFINITY` / `MC_NIC_PEER_AFFINITY`) never resolved for GPU buffers and RDMA fell back to unrestricted preferred-HCA selection.

This change matches entries with `GPU_PREFIX` (works for both `cuda:` and `hip:`), updates `hip_transport_test` to register with `GPU_PREFIX`, and adds a unit test that asserts affinity pins the vendor GPU location.

## Module

- [x] Transfer Engine (`mooncake-transfer-engine`)
- [ ] Mooncake Store (`mooncake-store`)
- [ ] Mooncake EP (`mooncake-ep`)
- [ ] Mooncake PG (`mooncake-pg`)
- [ ] Integration (`mooncake-integration`)
- [ ] P2P Store (`mooncake-p2p-store`)
- [ ] Python Wheel (`mooncake-wheel`)
- [ ] Common (`mooncake-common`)
- [ ] Mooncake RL (`mooncake-rl`)
- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
# Unit test (USE_HIP build)
./topology_test --gtest_filter='*HcaPeerAffinity*'
./topology_test   # full suite: 13/13 passed

# Manual affinity probe (before/after)
env MC_ENABLE_HCA_PEER_AFFINITY=1 MC_NIC_PEER_AFFINITY='L=P0' \
  <affinity_probe>
# After fix: selectDeviceByLocalHca(hip:0, L) always returns P0
```

**Test results:**
- [x] Unit tests pass
- [ ] Integration tests pass (if applicable)
- [x] Manual testing done (describe below)

On ROCm host: with the fix, `hip:0` selections under `MC_NIC_PEER_AFFINITY=L=P0` are 200/200 on `P0` (pre-fix: ~50/50 `P0`/`P1`).

## Checklist

- [x] I have performed a self-review of my own code
[...]

PR #3270, opened 2026-08-04
[TE] Replace hardcoded cuda: location keys with GPU_PREFIX

staryxchen:
## Description

Follow-up to #3248: audit remaining hardcoded `"cuda:"` Mooncake location / topology keys in Transfer Engine examples, tests, and IBGDA NIC selection.

These paths build under multi-vendor configs (`USE_HIP` / `USE_MUSA`) but still embedded `"cuda:0"` (or `"cuda:" + id`) while registration / discovery already use `GPU_PREFIX` (`hip:` / `musa:`). That can make NIC priority matrices miss the active GPU key.

This PR switches those call sites to `GPU_PREFIX`. Intentionally left unchanged:
- `GPU_PREFIX` definitions themselves
- TENT CUDA / Sunrise platform probes & plugins (vendor-native `cuda:`)
- Sunrise parser unit tests that assert `"cuda:"` parsing
- Docs and `torch.device("cuda:N")` strings

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
cmake -G Ninja .. -DUSE_HIP=ON -DUSE_CUDA=OFF -DBUILD_UNIT_TESTS=ON -DBUILD_EXAMPLES=ON ...
cmake --build . -j"$(nproc)" --target \
  transfer_metadata_test topology_test rdma_transport_test2 \
  transfer_engine_bench transfer_engine_validator hip_transport_test

MC_METADATA_SERVER=P2PHANDSHAKE \
  ./mooncake-transfer-engine/tests/transfer_metadata_test \
  --gtest_filter='TransferMetadataTest.NcclMetadataAndHandshakePayloadRoundTrip'

./mooncake-transfer-engine/tests/topology_test
```

**Test results:**
- [x] Unit tests pass

- Build (USE_HIP): related tests/examples compiled successfully
- `topology_test`: 13/13 passed
- `NcclMetadataAndHandshakePayloadRoundTrip` (GPU_PREFIX buffer names): passed
- Full `transfer_metadata_test`: 13 passed / 1 failed (`RpcMetaEntryTest` listen EBADF — unrelated env)
- `hip_transport_test`: failed on P2P listen/connect port mismatch (pre-existing; registration succeeded)

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

Cursor Agent assisted with repo-wide `"cuda:"` audit, `GPU_PREFIX` replacements, and local USE_HIP validation. Human submitter reviewed all changed lines.

### DIFF
--- a/mooncake-transfer-engine/example/nccl_host_transport_example.cpp
+++ b/mooncake-transfer-engine/example/nccl_host_transport_example.cpp
@@ -25,6 +25,7 @@
 #include <vector>
 
 #include "common.h"
+#include "cuda_alike.h"
 #include "transfer_engine.h"
 
 namespace {
@@ -179,8 +180,8 @@ int main(int argc, char** argv) {
                 0 ||
             !engines[rank]->installTransport("nccl", nullptr) ||
             engines[rank]->registerLocalMemory(
-                buffers[rank], kBufferBytes, "cuda:" + std::to_string(rank)) !=
-                0) {
+                buffers[rank], kBufferBytes,
+                GPU_PREFIX + std::to_string(rank)) != 0) {
             LOG(ERROR) << "Failed to initialize NCCL TE rank " << rank;
             return 1;
         }
--- a/mooncake-transfer-engine/example/transfer_engine_bench.cpp
+++ b/mooncake-transfer-engine/example/transfer_engine_bench.cpp
@@ -475,8 +475,8 @@ std::string loadNicPriorityMatrix() {
            " \"cpu:1\": [[" +
            device_names +
            "], []], "
-           " \"cuda:0\": [[" +
-           device_names +
+           " \"" +
+           GPU_PREFIX + "0\": [[" + device_names +
            "], []], "
            " \"musa:0\": [[" +
            device_names +
--- a/mooncake-transfer-engine/example/transfer_engine_bench_with_notify.cpp
+++ b/mooncake-transfer-engine/example/transfer_engine_bench_with_notify.cpp
@@ -295,8 +295,8 @@ std::string loadNicPriorityMatrix() {
            " \"cpu:1\": [[" +
            device_names +
            "], []], "
-           " \"cuda:0\": [[" +
-           device_names +
+           " \"" +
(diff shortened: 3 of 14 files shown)


he-yufeng:
Went through the whole sweep. The shape is right and the exclusions are the right ones (vendor-native probes, the Sunrise parser assertions, docs).

Verified scope coverage per file, since a missing include would only show up in a vendor build:

- The 8 files that gain `#include "cuda_alike.h"` get GPU_PREFIX from the default-build branch (`"cuda:"`) and the per-vendor branch under HIP/MUSA/etc., since cuda_alike.h routes to `gpu_vendor/*.h` by flag.
- The bench/validator files and `ibgda_device_transport.cpp` already include cuda_alike.h on main.
- Both Sunrise test files include `gpu_vendor/sunrise.h` directly, which defines GPU_PREFIX itself, so they are covered too.

Two observations, neither blocking:

1. `ibgda_device_transport.cpp` only builds in the CUDA path (the MACA stub replaces it otherwise), so the change there is a no-op in practice. Keeping it in the sweep is still the right call for grep-hygiene, just noting it does not change any vendor build.
2. The Sunrise test edits are also no-ops today, since `gpu_vendor/sunrise.h` defines GPU_PREFIX as `"cuda:"`. The real wins are the example/bench files, which do build under USE_HIP / USE_MUSA and would otherwise register as `hip:` / `musa:` while the location keys said `cuda:0`.

One thing worth a second pair of eyes in CI: `transfer_metadata_test.cpp` asserts on metadata keys, so under a vendor build the expectations now move with GPU_PREFIX, which is the intent, but it is the one file where a mismatch would fail loudly rather than silently.

Not approving myself since I cannot exercise any of the vendor builds locally; the CUDA-side correctness follows from GPU_PREFIX being the identical string there.

PR #3273, opened 2026-08-04
[Store] Fix flaky standby catch-up test

Icedcoco:
## Description

Fix the flaky `PromotionCatchUpTest.CatchesUpPrefixThatAppearsBeforePromotion` failure reported by PR #3222 CI.

The test used an unsynchronized fake `HaKvBackend` while the replication thread and the test thread accessed it concurrently. It also published the durable prefix and batch record through two independent `Put` calls. The replication thread could observe the updated prefix before the corresponding batch record, receive `ETCD_KEY_NOT_EXIST (-1001)`, and transition the standby service to `FAILED`.

The same failure was reproduced on the base branch, so this was an existing test race rather than a regression introduced by PR #3222.

This change is test-only:

- Protect fake backend state and injected errors with a mutex.
- Implement compare-and-put transaction semantics in the fake backend.
- Publish the batch record and updated durable prefix atomically with `WriteBatchAndAdvancePrefix`.
- Keep the production missing-batch error handling unchanged.

## Module

- [x] Mooncake Store

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

**Test commands:**

```bash
cmake --build build --target hot_standby_service_test -j32

  ./build/mooncake-store/tests/hot_standby_service_test \
  --gtest_filter=PromotionCatchUpTest.CatchesUpPrefixThatAppearsBeforePromotion \
  --gtest_repeat=100 \
  --gtest_brief=1

  ./build/mooncake-store/tests/hot_standby_service_test \
  --gtest_brief=1

pre-commit run --files \
  mooncake-store/tests/ha/standby/hot_standby_service_test.cpp
```

**Test results:**

- The flaky test passed all 100 repeated iterations.
- The full test binary passed 28 tests, with 15 tests skipped by configuration.
- All targeted pre-commit hooks passed.
- `LSAN_OPTIONS=detect_leaks=0` was used because LeakSanitizer exits with a ptrace-related runtime error in the local environment.

## Checklist

- [x] I have formatted the changed code.
- [x] I have run `pre-commit run --all-files` (targeted pre-commit hooks passed for the changed file).
- [x] Documentation update is not applicable because this is a test-only change.
- [x] I have added test coverage for the flaky scenario.
- [x] This change is below 500 LOC and does not require an RFC.

## AI Assistance Disclosure

- [ ] No AI tools were used.
- [x] AI tools were used.

Codex assisted with CI failure analysis, race reproduction, implementation, and verification. The human submitter reviewed every changed line and is responsible for the final change.

### DIFF
--- a/mooncake-store/tests/ha/standby/hot_standby_service_test.cpp
+++ b/mooncake-store/tests/ha/standby/hot_standby_service_test.cpp
@@ -6,6 +6,7 @@
 #include <chrono>
 #include <map>
 #include <memory>
+#include <mutex>
 #include <string>
 #include <string_view>
 #include <thread>
@@ -558,6 +559,7 @@ std::string MakeValidPayload(uint64_t client_id_first = 1,
 class FakeHaKvBackend : public HaKvBackend {
    public:
     ErrorCode Get(std::string_view key, std::string& value) override {
+        std::lock_guard<std::mutex> lock(mutex_);
         if (next_get_error_ != ErrorCode::OK) {
             auto error = next_get_error_;
             next_get_error_ = ErrorCode::OK;
@@ -574,11 +576,13 @@ class FakeHaKvBackend : public HaKvBackend {
         return ErrorCode::OK;
     }
     ErrorCode Put(std::string_view key, std::string_view value) override {
+        std::lock_guard<std::mutex> lock(mutex_);
         values_[std::string(key)] = std::string(value);
         return ErrorCode::OK;
     }
     ErrorCode Range(std::string_view begin_key, std::string_view end_key,
                     size_t limit, std::vector<KvPair>& kvs) override {
+        std::lock_guard<std::mutex> lock(mutex_);
         if (range_error_ != ErrorCode::OK) {
             return range_error_;
         }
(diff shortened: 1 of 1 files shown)


he-yufeng:
Nice catch, and the three layers of the fix line up correctly. I read through both the test change and the fake backend:

1. The mutex on `FakeHaKvBackend` was needed independent of this test: the replication thread runs `Range`/`Get` concurrently with the test thread's `Put`, and the setters are now locked too, which is the part people usually miss.
2. `WriteBatchAndAdvancePrefix` really does use a transaction (`kValueEquals` on the prefix key, then both puts), so the fake's stub `Txn` that returned OK without applying anything was silently not exercising the production path. The new fake covers both compare kinds the production code uses today (`kKeyNotExists` in the init path, `kValueEquals` here).
3. The test moving from two independent `Put` calls to the atomic `WriteBatchAndAdvancePrefix` is the actual race fix: the replication thread can no longer observe prefix-advanced-but-batch-missing, which is the state that produced ETCD_KEY_NOT_EXIST and the FAILED transition.

One nit for a follow-up, not a blocker: the fake's `Txn` treats every compare kind other than `kKeyNotExists` as value equality, so if `oplog_batch_storage` ever adds a version- or existence-based compare the fake would silently pass on wrong semantics. A `LOG(FATAL)` or explicit switch on unknown kinds would make that loud instead of invisible.

Aionw (approved):
LGTM

ykwd (approved):

00fish0 (approved):
Good catch!

PR #3276, opened 2026-08-04
[CI/Build] Make CMake target dependencies explicit

Aionw:
## Description

Make Mooncake's existing CMake target-to-target dependency graph explicit without changing dependency acquisition, versions, source layout, public outputs, or runtime behavior.

This change:

- replaces global or incidental include/link state with canonical CMake target edges for the dependencies already used by each target;
- groups each target's sources, usage requirements, compile settings, optional dependencies, and install rules together;
- separates Store sources into shared, master, and client object groups that map to the existing code boundaries;
- links `mooncake_master` through a reduced Store master library, so only Store client code depends on `transfer_engine`;
- keeps system integrations target-scoped for glog, gflags, JsonCpp, yaml-cpp, standalone Asio, yalantinglibs, xxHash, zstd, hiredis, liburing, libzmq, and pybind11; and
- removes fallback/compatibility discovery logic where an existing canonical target is available.

The previous graph relied on directory-wide include paths, bare library names, and the complete `mooncake_store` archive to provide accidental transitive requirements. Making direct edges explicit improves graph accuracy and prepares the build for a later, separate dependency-acquisition migration. This PR does not introduce FetchContent/ExternalProject or change any dependency revision.

No overlapping open issue or PR was found for this focused build-graph cleanup.

## Dependency Graph Change

These diagrams show CMake build/link relationships rather than runtime call
flow. Target names are exact; slash-separated dependency/consumer nodes are
visual groupings, not invented aggregate targets. The text in parentheses
identifies the source boundary assigned to an object target. For readability,
test targets and backend-specific transport targets are omitted, and canonical
third-party targets are grouped by consumer.

Red lines are relationships or implicit dependency mechanisms removed by this
PR. Green lines are newly explicit CMake target edges. Gray lines are retained
relationships. A green line does not mean a new code dependency: it means an
existing source-level dependency is now represented directly in CMake.

### Before

```mermaid
flowchart LR
  master["mooncake_master<br/>master.cpp"]
  client["mooncake_client<br/>real_client_main.cpp"]
  store["mooncake_store<br/>all shared + master + client .cpp files"]
  te["transfer_engine"]
  deps["third-party libraries<br/>bare names, manual include variables,<br/>and partial canonical targets"]
  global["directory-wide include/link state<br/>(not a target)"]
  te_consumers["base / tcp_transport / rpc_communicator /<br/>transport object targets"]

  master -->|"full Store archive"| store
  client --> store
  client --> te
  store -->|"PRIVATE"| te
  store -. "mixed implicit requirements" .-> deps
  global -. "directory scope" .-> store
  global -. "incidental includes" .-> te_consumers

  linkStyle 0,4,5,6 stroke:#d73a49,stroke-width:3px
  linkStyle 1,2,3 stroke:#8c959f,stroke-width:2px
```

The old `mooncake_master -> mooncake_store -> transfer_engine` path existed
because the single Store archive contained both master and client code. The
linker's `--as-needed` option could discard unused runtime libraries, but it did
not make the CMake graph match the source boundary.

### After

```mermaid
flowchart LR
  master["mooncake_master<br/>master.cpp"]
  client["mooncake_client<br/>real_client_main.cpp"]
  store_master["mooncake_store_master<br/>shared + master sources only"]
  store["mooncake_store<br/>shared + master + client sources"]
  te["transfer_engine"]

  master -. "removed" .-> store
  master -->|"new reduced Store path"| store_master
  client -->|"retained"| store
  client -->|"retained direct edge"| te
  store -->|"client implementation edge"| te

  classDef focus stroke:#0969da,stroke-width:3px
  classDef support stroke:#8c959f,stroke-width:1.5px
  class master,client,te focus
  class store_master,store support

  linkStyle 0 stroke:#d73a49,stroke-width:3px
  linkStyle 1 stroke:#2da44e,stroke-width:3px
  linkStyle 2,3,4 stroke:#8c959f,stroke-width:2px
```

The red dotted edge is shown only to identify the removed relationship; it is
not part of the resulting graph. `mooncake_master` now terminates at
`mooncake_store_master`, which contains no client sources and has no path to
`transfer_engine`. `mooncake_client` retains both its complete Store library
and direct `transfer_engine` edges.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [x] Other: build-system dependency graph cleanup

## How Has This Been Tested?

Rebased onto `origin/main` at `919c9893` before the final validation.

**Test commands:**

```bash
cmake -S . -B build-pr-check \
  -DCMAKE_BUILD_TYPE=Release \
  -DCMAKE_EXPORT_COMPILE_COMMANDS=ON \
  -DCMAKE_DISABLE_FIND_PACKAGE_CUDAToolkit=TRUE \
  -DUSE_CUDA=OFF -DUSE_HIP=OFF -DUSE_TCP=ON \
  -DBUILD_UNIT_TESTS=ON -DBUILD_EXAMPLES=OFF -DBUILD_BENCHMARK=OFF
cmake --build build-pr-check --parallel 128
ctest --test-dir build-pr-check --output-on-failure --parallel 128

ctest --test-dir build-pr-check --output-on-failure --parallel 13 \
  -R '^(tcp_transport_test|transfer_metadata_test|utils_test|client_tcp_local_memcpy_test|storage_backend_test|client_storage_backend_test|file_storage_test|task_integration_test|dummy_client_get_buffer_test|dummy_client_object_checksum_test|uds_transport_test|hot_standby_service_test|master_service_ha_test)$'

cmake -S . -B build-pr-optional \
  -DCMAKE_BUILD_TYPE=Release \
  -DCMAKE_DISABLE_FIND_PACKAGE_CUDAToolkit=TRUE \
  -DUSE_CUDA=OFF -DUSE_HIP=OFF -DUSE_TCP=ON \
  -DSTORE_USE_ETCD=OFF -DSTORE_USE_REDIS=ON \
  -DENABLE_KV_EVENTS=OFF \
  -DBUILD_UNIT_TESTS=OFF -DBUILD_EXAMPLES=OFF -DBUILD_BENCHMARK=OFF
cmake --build build-pr-optional --target mooncake_master --parallel 128

env PYTHONPATH=build-pr-check/mooncake-integration \
  python3 -c 'import engine, store; print(engine.__file__); print(store.__file__)'

git diff --check
git diff --name-only -z | xargs -0 env SKIP=cmake-format,trailing-whitespace pre-commit run --files
git diff --name-only -z | xargs -0 pre-commit run trailing-whitespace --files
```

`compile_commands.json` and final link lines were also audited. Direct glog consumers including `base/status.cpp` and `tcp_transport.cpp` receive glog usage requirements from their targets. `mooncake_master` does not link `transfer_engine`; `mooncake_client` does.

**Test results:**

- [ ] Unit tests pass
- [x] Manual testing done

- The complete Linux CPU/TCP build passed, including Store master/client, Transfer Engine, Python extensions, and every test executable.
- 100/113 CTest targets passed in the restricted sandbox. Rerunning the failures with network-interface and Unix-socket permissions, plus rerunning the file test without cross-test directory races, brought the result to 109/113.
- The four remaining CTest targets require configuration/services not provided by this build: `tcp_transport_test` and `transfer_metadata_test` require a working metadata server; `hot_standby_service_test` and `master_service_ha_test` contain cases requiring `STORE_USE_ETCD`.
- `oplog_test_failpoint_test`, whose direct include edge was added after rebasing the new upstream target, passed.
- The Redis-enabled `mooncake_master` build passed.
- liburing was detected and exercised by the normal client build.
- ZeroMQ could not be exercised because this host does not provide `zmq.h`; configuration failed at the expected required dependency check when `ENABLE_KV_EVENTS=ON`.
- Both built Python extensions imported successfully.
- All executed pre-commit hooks passed. `cmake-format` was skipped because it rewrites unrelated pre-existing formatting throughout already-modified CMake files; those mechanical rewrites were intentionally kept out of this PR.
- `git diff --check` passed.

## Checklist

- [ ] I have formatted my code using `./scripts/code_format.sh` (no C/C++ source files changed; see the CMake formatting note above)
- [ ] I have run `pre-commit run --all-files` and all hooks pass (touched-file hooks were run; see the CMake formatting note above)
- [x] I have updated the documentation (not applicable: no user-facing behavior changed)
- [ ] I have added tests to prove my changes are effective (the existing build and test graph exercises the change)
- [ ] For changes >500 LOC: I have filed an RFC issue (this PR is limited to expressing the existing build graph; no runtime or dependency-acquisition architecture changes)

## AI Assistance Disclosure

- [ ] No AI tools were used

OpenAI Codex was used to inspect the source/include graph, implement the CMake target edges, reorganize target definitions, and run validation. The human submitter is responsible for reviewing and understanding every changed line before merge.


whn09 (approved):
Thanks for the cleanup — the glog change is correct: both efa_context.h and efa_endpoint.h include <glog/logging.h> and are public headers, so PUBLIC is right.

One thing to fix though: `gflags::gflags` shouldn't be here at all. efa_transport uses zero gflags symbols — no DEFINE_*, no DECLARE_*, no FLAGS_* in either the headers or the .cpp files. Runtime config goes through `globalConfig()` (config.h), which itself doesn't pull in gflags.
  
The only trace is a stale `#include <gflags/gflags.h>` at efa_context.h:18 that no longer uses anything from it. Adding a PUBLIC link edge here would make the dependency graph less accurate, not more.

Suggest:
target_link_libraries(efa_transport PUBLIC glog::glog PRIVATE fabric)

Happy to drop the stale include in a follow-up so the header matches.

Icedcoco:
Thanks for the CMake cleanup — the shared/master/client object split makes the dependency graph much easier to follow.

One thing I'm concerned about is: in `mooncake-store/src/CMakeLists.txt`, `mooncake_store_client_objects` doesn't link `xxHash::xxhash`, even though `storage/distributed/distributed_storage_backend.cpp` is a client source and uses `XXH64` (plus `#include <xxhash.h>`). The master objects and the final `mooncake_store` target both get the edge, so it looks like the client group may have just been missed.

It builds fine on Ubuntu since `xxhash.h` happens to be in `/usr/include`, but the old code explicitly searched `/usr/local/include` and applied the found directory to the whole store library. For setups where xxhash is installed outside the default include path, the client object compile would lose that `-I` and fail. Adding `xxHash::xxhash` to the `mooncake_store_client_objects` link list would close the gap — did you intentionally rely on the system include path here, or would adding that edge sound reasonable to you?

Aionw:
Fxied in https://github.com/kvcache-ai/Mooncake/pull/3276/commits/a41cdf1dc56e13cd5b329299c85201e2d47d2615

Icedcoco:
I was checking a few less-common build paths and wanted to discuss three edge cases before deciding whether they are expected.

1. `WITH_STORE_C_SHARED`

With the current `PUBLIC transfer_engine` dependency in `mooncake-store/src/CMakeLists.txt`, I can reproduce multiple-definition errors when building `WITH_STORE_C_SHARED=ON`. The link ends up receiving `tcp_transport.cpp.o` through the propagated `transport` interface sources, while the same objects are also included through the whole-archived `libmooncake_store.a`.

Was making `transfer_engine` public necessary for consumers to see its headers? Would keeping the link dependency `PRIVATE` and exposing the required include path separately be a better fit for this target, or is there another intended way to keep `mooncake_store_c` self-contained?

2. Imported canonical targets

The new fallback targets such as `zstd::libzstd`, `xxHash::xxhash`, `liburing::liburing`, and `hiredis::hiredis` are created unconditionally. When Mooncake is added as a subdirectory of a parent project that already provides one of these canonical targets, CMake fails with a duplicate-target error.

Would it make sense to guard the fallback definitions with `if(NOT TARGET ...)`, similar to the existing hiredis handling in the TENT metastore? I noticed the same pattern also appears in the standalone Transfer Engine and TENT io_uring setup.

3. Remaining bare `glog` link

In `mooncake-transfer-engine/tent/tests/CMakeLists.txt`, `tent_metrics_http_server_test` still links the bare `glog` name, while the surrounding targets use `glog::glog`. Since `tent_metrics`/`tent_common` already provide the canonical dependency transitively, was the bare entry intentional? If not, could we either remove it or switch it for consistency?

alogfans (approved):

ykwd:
Total Test time (real) = 470.59 sec
Errors while running CTest

The following tests FAILED:
	107 - mmap_arena_fallback_test (Failed)

Icedcoco:
I have two questions after another pass over the latest head:

1. With `USE_3FS=ON`, `dfs_global_allocator.cpp` is part of `MOONCAKE_STORE_MASTER_SOURCES` and constructs `Hf3fsAdapter`, while `hf3fs_adapter.cpp` and `${HF3FS_SOURCES}` are only added to `MOONCAKE_STORE_CLIENT_SOURCES`. Since `mooncake_store_master` contains only the shared and master objects, wouldn’t this leave `mooncake_master` with unresolved `Hf3fsAdapter` symbols? I could only get as far as configuration locally because I do not have the 3FS SDK installed, but the target graph seems to miss those implementations. Would moving the HF3FS sources into `MOONCAKE_STORE_SHARED_SOURCES` make more sense, since both master and client use them? It may also be worth adding a `USE_3FS=ON` build check for `mooncake_master`.

2. The top-level and standalone projects still declare CMake 3.16 as the minimum version, but the new dependency providers use `find_path(... REQUIRED)` and `find_library(... REQUIRED)`, whose `REQUIRED` option was added in CMake 3.18. Is CMake 3.16 still intended to be supported? If so, should these use explicit `NOTFOUND` checks instead? Otherwise, should the minimum required version be raised to 3.18?

Icedcoco:
The 3FS source placement and the dependency provider checks look correct now. I noticed one remaining CMake 3.16 compatibility case in `mooncake-common/src/CMakeLists.txt`: `find_path(ASIO_INCLUDE_DIR asio.hpp REQUIRED)` still uses the `REQUIRED` option introduced in CMake 3.18. Could this one be converted to the same explicit `NOTFOUND` check as the other providers? After that, I don’t see any remaining blocking issues.

UNIDY2002 (approved):

ykwd (approved):

UNIDY2002:
(medium confidence) This PR seems to be breaking `ascend-test`.
https://github.com/kvcache-ai/Mooncake/actions/runs/32442336097/job/96655367711
@Aionw PTAL, great thanks!

Aionw:
I'll take a look.

staryxchen:
`ascend-test / build-and-test` is failing at Configure CMake with:

```text
CMake Error at cmake/Dependencies.cmake:21 (message):
  xxHash development files not found
```

Job: https://github.com/kvcache-ai/Mooncake/actions/runs/33393280998/job/99491727725

This seems related to `mooncake_provide_xxhash()` treating missing xxHash as fatal. The Ascend image (`mooncake-hixl-ci:v5`) may not have the xxHash development files. Could we confirm whether the image (or `ci_ascend.yml`) needs to install them so CMake can get past configure?

PR #3278, opened 2026-08-04
fix(transfer-engine): block SIGTERM/SIGINT for the shutdown watcher

### LINKED ISSUE #3271: [Bug]: TE Graceful shutdown may deadlock when SIGTERM or SIGINT is delivered to the watcher thread
## Description

While stress-testing graceful_shutdown_test after fixing a separate child-readiness race, I found a lower-probability issue in the graceful shutdown implementation.

Mooncake uses a self-pipe pattern instead of cleaning up the TransferEngine directly inside the signal handler:

```
Signal-receiving thread              Watcher thread
          |                                |
          | signal handler                 | blocking read(pipe)
          | write signal to pipe --------->|
          | pause()                        | cleanupEngines()
                                           | _Exit(128 + signal)
```

This design is generally appropriate because complex cleanup is not async-signal-safe. Operations such as locking mutexes, releasing memory, or destroying threads may deadlock or corrupt state when executed directly inside a signal handler.

However, process-directed signals such as SIGTERM and SIGINT may be delivered to any thread that does not block them. The watcher thread currently does not block these signals.

If the signal is delivered to a normal thread, shutdown works as expected:

```
Normal thread receives the signal
    -> handler writes to the pipe
    -> watcher reads from the pipe
    -> watcher cleans up the engines
    -> process exits
```

**If the signal is delivered to the watcher thread itself, the process can deadlock:**

```
Watcher receives the signal
    -> watcher enters the signal handler
    -> handler writes the signal to the pipe
    -> handler waits in pause()
    -> no thread remains to read the pipe
    -> cleanup never runs
    -> process never exits
```

In other words, the only pipe reader becomes blocked inside the signal handler after writing a message to itself.

## Steps to reproduce

Running multiple test processes concurrently increases the reproduction probability:
```
seq 1 400 | xargs -P 8 -I ID \
  build/mooncake-transfer-engine/tests/graceful_shutdown_test \
  --gtest_brief=1 \
  --gtest_shuffle \
  --gtest_random_seed=ID
```

The child did not exit within the five-second timeout and had to be terminated with SIGKILL. Example failure:
```
[ RUN      ] GracefulShutdownTest.SigintTriggersCleanExit
graceful_shutdown_test.cpp: Failure
child did not exit before timeout

Value of: WIFEXITED(status)
  Actual: false
Expected: true

[  FAILED  ] GracefulShutdownTest.SigintTriggersCleanExit
```

## Expected behavior

The watcher thread should remain available to read the self-pipe and perform cleanup regardless of which process-directed signal is received.

After receiving SIGTERM or SIGINT, the process should clean up registered engines and exit with 128 + signal, for example:

- 143 for SIGTERM
- 130 for SIGINT

## Actual behavior

If the signal is delivered to the watcher thread, the watcher enters the signal handler, writes to its own pipe, and waits indefinitely. No thread remains to consume the pipe notification, so cleanup and process exit never occur.

## Possible fix

Ensure that the watcher thread cannot receive SIGTERM or SIGINT.
One possible approach is:
1. Temporarily block SIGTERM and SIGINT in the thread creating the watcher.
2. Create the watcher so that it inherits the blocked signal mask.
3. Restore the creating thread’s original signal mask.
4. Keep the signals blocked in the watcher for its entire lifetime.

For example, this can be implemented using pthread_sigmask() around watcher thread creation.

All failure paths should restore the caller’s original signal mask and close any newly created pipe file descriptors.

## Impact

This can cause a process to hang indefinitely during graceful shutdown. In containerized or orchestrated environments, the process may eventually be force-killed after its termination grace period, preventing normal Transfer Engine cleanup.

The issue is timing-dependent and may be difficult to reproduce in a single test run, but it was observed during repeated stress testing.

### Before submitting...

he-yufeng:
Fixes #3271.

The graceful-shutdown self-pipe can deadlock when a process-directed signal is delivered to the watcher thread itself. `shutdownSignalHandler` writes one byte into the pipe and then waits in `pause()` forever; if it ran on the watcher thread, the only pipe reader is the thread now suspended inside the handler, so `cleanupEngines()` never runs and the process hangs until something SIGKILLs it. The fork guard (`g_handlers_pid != getpid()`) covers children, not this case.

The fix blocks SIGTERM and SIGINT around `std::thread(signalWatcher)` creation in `startSignalWatcherLocked()`, so the watcher inherits the blocked mask for its whole lifetime and the signal always lands on some other thread. The creator's original mask is restored afterwards on every path, including the thread-creation failure path, since the caller may have its own reasons to block these signals and `installGracefulShutdownHandlers()` runs per engine. If `pthread_sigmask` itself fails, install proceeds with today's exposure instead of refusing to install handlers.

@00fish0 analyzed this in the issue and sketched the same approach; I checked with them first and picked it up after no reply.

## Verification

No local build tree here (the repo build dirs point at container paths), so the change is validated two ways and CI runs the real suite:

- Standalone g++ -Wall -Wextra -Werror build of the extracted `startSignalWatcherLocked()` shape: mask restore verified (the calling thread is not left with SIGTERM blocked), and the pipe protocol still drives a 128+SIGTERM exit.
- A second standalone run proves the watcher thread really inherits the blocked mask: the watcher checks `sigismember` for both signals on entry and would exit with a distinctive code if either were unblocked; it exits 143 instead.

The existing graceful_shutdown tests keep their meaning: every scenario still has at least one unblocked thread, so the signal always has somewhere to land.

Signed-off-by: Yufeng He <40085740+he-yufeng@users.noreply.github.com>

### DIFF
--- a/mooncake-transfer-engine/src/graceful_shutdown.cpp
+++ b/mooncake-transfer-engine/src/graceful_shutdown.cpp
@@ -15,6 +15,7 @@
 #include "graceful_shutdown.h"
 
 #include <errno.h>
+#include <pthread.h>
 #include <signal.h>
 #include <stdlib.h>
 #include <unistd.h>
@@ -103,9 +104,30 @@ bool startSignalWatcherLocked(pid_t current_pid) {
 
     if (pipe(g_signal_pipe) != 0) return false;
 
+    // The watcher must never take SIGTERM/SIGINT itself: the handler writes
+    // to the pipe and then pauses forever, so if it ran on the watcher
+    // thread no reader would be left and cleanup would never run. Block both
+    // signals around thread creation so the watcher inherits the mask, and
+    // restore the caller's original mask on every path (it may have had its
+    // own reasons to block these signals).
+    sigset_t watcher_block_set;
+    sigset_t saved_mask;
+    sigemptyset(&watcher_block_set);
+    sigaddset(&watcher_block_set, SIGTERM);
+    sigaddset(&watcher_block_set, SIGINT);
+    const bool mask_blocked =
+        pthread_sigmask(SIG_BLOCK, &watcher_block_set, &saved_mask) == 0;
+
+    bool watcher_started = false;
     try {
         std::thread(signalWatcher).detach();
+        watcher_started = true;
     } catch (...) {
+    }
+
+    if (mask_blocked) pthread_sigmask(SIG_SETMASK, &saved_mask, nullptr);
+
+    if (!watcher_started) {
         close(g_signal_pipe[0]);
         close(g_signal_pipe[1]);
         g_signal_pipe[0] = -1;
--- a/mooncake-transfer-engine/src/graceful_shutdown.cpp
+++ b/mooncake-transfer-engine/src/graceful_shutdown.cpp
@@ -123,7 +123,8 @@ bool startSignalWatcherLocked(pid_t current_pid) {
         std::thread(signalWatcher).detach();
         watcher_started = true;
     } catch (const std::exception& e) {
-        LOG(WARNING) << "action=start_signal_watcher_failed, error=" << e.what();
+        LOG(WARNING) << "action=start_signal_watcher_failed, error="
+                     << e.what();
     } catch (...) {
     }
 
--- a/mooncake-transfer-engine/src/graceful_shutdown.cpp
+++ b/mooncake-transfer-engine/src/graceful_shutdown.cpp
@@ -122,6 +122,8 @@ bool startSignalWatcherLocked(pid_t current_pid) {
     try {
         std::thread(signalWatcher).detach();
         watcher_started = true;
+    } catch (const std::exception& e) {
+        LOG(WARNING) << "action=start_signal_watcher_failed, error=" << e.what();
     } catch (...) {
     }
 


mooncake-transfer-engine/src/graceful_shutdown.cpp
```diff
@@ -103,9 +104,30 @@ bool startSignalWatcherLocked(pid_t current_pid) {
...
+    sigaddset(&watcher_block_set, SIGINT);
+    const bool mask_blocked =
+        pthread_sigmask(SIG_BLOCK, &watcher_block_set, &saved_mask) == 0;
+
+    bool watcher_started = false;
     try {
         std::thread(signalWatcher).detach();
+        watcher_started = true;
     } catch (...) {
+    }
```
alogfans, line 129: Is it safe to print something in the catch block?

he-yufeng:
Yes, safe here: this catch sits in `installGracefulShutdownHandlers()`, the normal install path, not in the signal handler, so the async-signal-safety rules do not apply. Added a warning with the exception's `what()` in ce7c004c, and the file already sees glog through `memory_location.h`. The silent fallback (install returns false, no handlers) now at least leaves a trace when thread creation fails, which is exactly the case worth diagnosing.

he-yufeng:
The `tent-ci (cuda-off-metrics-on)` failure is `RuntimeQueueDispatch.DispatchesOnlyOneWindowOnSubmit`, a TENT timing test. This change only touches `mooncake-transfer-engine/src/graceful_shutdown.cpp`, which is outside the TENT tree, so it cannot affect that test; the failure reads as a flake. The other red job is the CI gate downstream of it. Main currently has unrelated failures of its own (test-sglang-integration).

alogfans (approved):
LGTM

PR #3279, opened 2026-08-04
[Store] Add fatal crash stack traces

Aionw:
## Description

Add a small, centralized crash handler for the standalone Mooncake Store processes.

- Install glog's failure signal handler during `mooncake_master` and `mooncake_client` startup, before normal service work begins.
- Cover `SIGSEGV`, `SIGABRT` (including standard C/C++ `assert()`), `SIGBUS`, `SIGFPE`, and `SIGILL`.
- Write fatal-signal diagnostics directly to stderr, independent of `FLAGS_log_dir` and `MC_LOG_DIR`.
- Preserve the existing `SIGTERM` disposition because glog 0.4.0 also treats `SIGTERM` as a failure signal. `SIGINT` is not modified.
- Re-raise the original fatal signal after reporting, preserving signal-based termination and the host's core-dump policy.
- Keep Python extensions and general Transfer Engine/Store library use free of automatic process-global signal-handler changes.

No existing open issue or PR was found for this change.

### stderr output

A native crash produces a best-effort failure header and native stack trace similar to:

```text
*** Aborted at <unix-time> (unix time) ***
PC: @ <address> <symbol-or-unknown>
*** SIGSEGV (...) received by PID <pid> (TID <tid>); stack trace: ***
    @ <address> <frame>
    @ <address> <frame>
    @ <address> main
    @ <address> __libc_start_main
    @ <address> _start
```

A standard assertion first emits the libc assertion message, followed by the `SIGABRT` trace:

```text
mooncake_client: <file>:<line>: Assertion `<expression>` failed.
*** Aborted at <unix-time> (unix time) ***
PC: @ <address> <symbol-or-unknown>
*** SIGABRT (...) received by PID <pid> (TID <tid>); stack trace: ***
    @ <address> raise
    @ <address> abort
    @ <address> __assert_fail
    @ <address> <Mooncake frame>
```

Stack walking and symbolization are best effort under severe memory or stack corruption. After printing, the process still terminates from the original signal (for example, shell status 139 for `SIGSEGV` or 134 for `SIGABRT`), and a core dump is produced when enabled by the OS/container configuration.

## Module

## Type of Change

- [x] New feature

## How Has This Been Tested?

**Test commands:**
```bash
cmake -S . -B build-crash-handler-services \
  -DBUILD_UNIT_TESTS=OFF \
  -DBUILD_EXAMPLES=OFF \
  -DBUILD_BENCHMARK=OFF \
  -DWITH_STORE_RUST=OFF \
cmake --build build-crash-handler-services \
  --target mooncake_master mooncake_client \
  --parallel 128
pre-commit run --files \
  mooncake-common/include/crash_handler.h \
  mooncake-common/src/crash_handler.cpp \
  mooncake-common/src/CMakeLists.txt \
  mooncake-store/src/master.cpp \
  mooncake-store/src/real_client_main.cpp
git diff --check
```

**Test results:**
- [x] `mooncake_master` and `mooncake_client` build successfully.
- [x] Manual glog 0.4.0 signal validation confirmed stderr stack output and original-signal exit statuses.
- [x] Pre-commit passes on every changed file.
- [x] `git diff --check` passes.

## Checklist

- [x] I have formatted my code using the repository formatting hook
- [x] Documentation is not required; the behavior and output are documented in this PR
- [x] An RFC issue is not required because this change is under 500 LOC

## AI Assistance Disclosure

- [x] AI tools were used

Codex helped inspect signal ownership, implement the centralized wrapper, run builds and formatting checks, and draft this PR description. The human submitter reviewed the resulting scope and explicitly limited automatic installation to the standalone master and client service processes.

### DIFF
--- a/mooncake-common/include/crash_handler.h
+++ b/mooncake-common/include/crash_handler.h
@@ -1,0 +1,26 @@
+// Copyright 2026 Mooncake Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+namespace mooncake {
+
+// Installs best-effort stderr stack traces for fatal crash signals. The
+// installation is process-global and idempotent. It deliberately leaves
+// SIGINT and SIGTERM unchanged so host applications retain shutdown ownership.
+// After reporting, the original fatal signal is re-raised by glog's handler,
+// preserving normal signal termination and operating-system core-dump policy.
+void InstallCrashHandler();
+
+}  // namespace mooncake
--- a/mooncake-common/src/CMakeLists.txt
+++ b/mooncake-common/src/CMakeLists.txt
@@ -1,7 +1,8 @@
 find_package(yaml-cpp REQUIRED)
 
-set(MOONCAKE_COMMON_SOURCES crc_checksum.cpp default_config.cpp environ.cpp
-                            rpc_client_io_context.cpp)
+set(MOONCAKE_COMMON_SOURCES
+    crash_handler.cpp crc_checksum.cpp default_config.cpp environ.cpp
+    rpc_client_io_context.cpp)
 
 add_library(asio_shared SHARED asio_impl.cpp)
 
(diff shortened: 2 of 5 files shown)


Icedcoco:
Thanks for the PR — the crash handler is a nice, small addition, and I checked the SIGTERM handling against glog 0.4.0's actual behavior, so the overall approach looks solid to me. I did have a few smaller things I'd love your thoughts on, rather than blocking comments.

1. In `InstallCrashHandler()`, the `sigaction` calls are currently silent on failure: if saving the SIGTERM disposition fails, we return without installing anything at all, and if restoring it fails, glog's SIGTERM handler stays in place. Both are very unlikely on Linux, but the failure modes are exactly "no crash output" and "SIGTERM no longer means shutdown," so would it be worth logging those errors (or otherwise handling them) so the behavior doesn't fail silently?

2. There's a small window between `google::InstallFailureSignalHandler()` and the SIGTERM restore where a SIGTERM would be treated as a crash signal. In `mooncake_master` this is technically possible during startup, though the process still exits with the same signal — just with an extra stack trace — and in `mooncake_client` it's masked by `ResourceTracker`. I suspect it's negligible in practice; do you agree, or would it be worth documenting the window in a comment?

3. I'm curious about the choice to link `glog::glog` PUBLIC on `mooncake_common`. Since the crash handler is only used by the two Store binaries, this makes glog a hard dependency for every `mooncake_common` consumer — common-only builds would fail at configure time without it, even though all current CI builds install it. Putting the handler in `mooncake-store` would keep common free of that dependency, though centralizing it in common also has its appeal. Was that a deliberate tradeoff?

Overall I don't see any blocking issues in the code itself, so happy to approve once we've aligned on these.

PR #3281, opened 2026-08-04
[TENT] RDMA: skip NICs that cannot GPUDirect-DMA to a GPU, and fix fallback device rotation

anranxia:
## Description

On strict-topology GPU clusters — where, per `nvidia-smi topo -m`, only each GPU's PCIe-closest ("PIX") NIC can perform GPUDirect RDMA to that GPU — the TENT transfer engine sprays RDMA slices across all NICs. Slices landing on a NIC that cannot P2P-DMA to the target GPU fail; the engine then exhausts per-slice retries and the transfer aborts (`Slice ... failed: retry count exceeded` → `No more transports available after rdma failed`, ending in an uncaught `std::out_of_range: unordered_map::at` and rc=134).

`ibv_reg_mr` is control-plane only: it succeeds on every NIC regardless of whether the PCIe P2P path to the GPU actually works, so registration cannot detect this. Classic TE avoids the problem by pinning a single tier-1 NIC, but that also gives up the multi-rail aggregation that is TENT's main advantage on permissive fabrics.

This PR makes TENT **learn NIC↔GPU GPUDirect reachability at runtime and converge onto the working rail(s)** instead of crashing, while leaving multi-rail behavior on permissive fabrics completely intact (the mechanism stays inert until something actually fails).

### What changed

1. **`GdrReachability` — a process-level reachability tracker** (new `gdr_reachability.{h,cpp}`). Thread-safe singleton keyed on `(nic, gpu)` locally and `(machine_id, nic, gpu)` remotely. It reuses `RailMonitor`'s algorithm: an error threshold within a sliding window, then exponential cooldown (capped at 300s) with automatic re-admission on the next success. A fast-path `hasAnyExclusion()` atomic keeps the whole thing zero-cost until the first real failure is recorded.

2. **Runtime learning in the completion path** (`workers.cpp` `asyncPollCq`). A protection/access error on a GPU buffer means the chosen NIC cannot P2P-DMA to that GPU. The local side (source NIC → source GPU) surfaces as `IBV_WC_LOC_PROT_ERR`; the remote side (target NIC → target GPU) as `IBV_WC_REM_ACCESS_ERR` or, on the strict
   fabrics we tested, `IBV_WC_REM_OP_ERR`. These are recorded; a subsequent success clears the entry (self-heal). Device selection then skips unreachable rails in `DeviceSelector::allocate` (mask) and `selectFallbackDevice` (per-pair check).

3. **Fallback device-selection fix** (`workers.cpp` `selectFallbackDevice`). The retry path used `idx = slice->retry_count` as a forward-only cursor. Because `getDeviceByFlatIndex` walks the per-GPU **priority-ranked** NIC list, flat index 0 is the `(source PIX NIC, target PIX NIC)` pair — the ideal GPUDirect-capable path. The old cursor deprioritized index 0 as retries grew, so on a strict fabric the only working same-machine loopback pair `(PIX, PIX)` at index 0 was never revisited once `retry_count > 0`, exhausting the retry budget and aborting the transfer. The fallback now **scans from index 0 and takes the first reachable pair** (rail-down and GDR-excluded pairs are still skipped), so it converges on the best available pair immediately. Fallback only runs on failure and does not touch the retry-0 multi-rail distribution, so healthy/permissive fabrics are unaffected.

## Module

## Type of Change

- [x] Bug fix
- [x] Performance improvement

## How Has This Been Tested?

Built with `-DUSE_TENT=ON -DUSE_CUDA=ON -DWITH_METRICS=ON` and exercised with `tebench` on two strict-topology pods (each 8×H20, 8× `mlx5_bond_*` RoCE NICs; only each GPU's PIX NIC can GPUDirect). Runtime learning only — default error threshold, no extra flags.

**Test commands:**
```bash
# target (one node)
ulimit -l unlimited
./tebench --backend tent --seg_type VRAM --local_gpu_id 0 \
          --metadata_type p2p --rpc_server_port 18600

# initiator (same node for loopback; peer node for cross-node)
ulimit -l unlimited
./tebench --target_seg_name=<target-ip>:18600 --backend tent --seg_type VRAM \
          --op_type read --local_gpu_id 0 --target_gpu_id 0 \
          --start_block_size 1048576 --max_block_size 8388608 \
          --start_batch_size 8 --max_batch_size 8 \
          --start_num_threads 8 --max_num_threads 8 --metadata_type p2p
```

**Test results** (GB/s at 1/2/4/8 MiB, batch 8, 8 threads):

| Scenario                          | 1M   | 2M   | 4M   | 8M   | retry-exceeded | result |
|-----------------------------------|------|------|------|------|----------------|--------|
| same-machine loopback GPU0→GPU0   | 33.9 | 52.0 | 55.2 | 54.9 | 0              | rc=0   |
| cross-node GPU0→GPU0              | 24.1 | 42.3 | 48.4 | 49.0 | 0              | rc=0   |

Before this change, both scenarios aborted with rc=134 (`retry count exceeded` → `No more transports available` → `std::out_of_range`). Cross-node throughput is well above classic TE's single-rail baseline (~24.5 GB/s) on the same fabric. On a separate permissive pod (all NICs can GDR), the tracker stays inert — no exclusions, multi-rail throughput
unchanged.

- [ ] Unit tests pass

## Checklist

- [x] For changes >500 LOC: I have filed an RFC issue — N/A (<400 LOC)

## AI Assistance Disclosure

AI assistance (Claude Code) was used to help draft the `GdrReachability` component and the completion/fallback wiring. All changes were reviewed line-by-line and validated end-to-end on real strict-topology hardware by the submitter.

### DIFF
--- a/mooncake-transfer-engine/tent/src/transport/rdma/workers.cpp
+++ b/mooncake-transfer-engine/tent/src/transport/rdma/workers.cpp
@@ -14,6 +14,8 @@
 
 #include "tent/transport/rdma/workers.h"
 
+#include "tent/transport/rdma/gdr_reachability.h"
+
 #include <sys/epoll.h>
 
 #include <algorithm>
@@ -62,6 +64,7 @@ Workers::Workers(RdmaTransport* transport)
     device_selector_ = std::make_unique<DeviceSelector>();
     device_selector_->loadTopology(transport_->local_topology_);
     auto& conf = transport_->conf_;
+    GdrReachability::instance().configure(conf.get());
 
     // RailMonitor consumes JSON text, while the public configuration is a file
     // path. Load it once here instead of reopening the file for every worker
@@ -667,6 +670,27 @@ void Workers::asyncPollCq() {
                               << ", local_nic: " << context->name()
                               << "): " << ibv_wc_status_str(wc[i].status);
                 }
+                // GPUDirect reachability learning: a protection/access error
+                // on a GPU buffer means the chosen NIC cannot P2P-DMA to that
+                // GPU (ibv_reg_mr succeeded but the PCIe path is unusable).
+                // Record it so selection avoids that NIC and converges onto a
+                // reachable rail instead of exhausting retries. The local side
+                // (source NIC -> source GPU) surfaces as LOC_PROT; the remote
+                // side (target NIC -> target GPU) as REM_ACCESS or, for a
+                // remote GDR-read failure, REM_OP (observed on strict fabrics).
+                bool local_gdr_err = (wc[i].status == IBV_WC_LOC_PROT_ERR);
+                bool remote_gdr_err = (wc[i].status == IBV_WC_REM_ACCESS_ERR ||
+                                       wc[i].status == IBV_WC_REM_OP_ERR);
+                if (local_gdr_err && slice->source_gpu_ordinal >= 0 &&
+                    slice->source_nic_name) {
+                    GdrReachability::instance().reportLocalFailure(
+                        slice->source_nic_name, slice->source_gpu_ordinal);
+                } else if (remote_gdr_err && slice->target_gpu_ordinal >= 0 &&
+                           slice->target_nic_name && slice->target_machine_id) {
+                    GdrReachability::instance().reportRemoteFailure(
+                        *slice->target_machine_id, slice->target_nic_name,
+                        slice->target_gpu_ordinal);
+                }
                 slice->retry_count++;
                 if (slice->retry_count >=
                     transport_->params_->workers.max_retry_count) {
@@ -686,6 +710,22 @@ void Workers::asyncPollCq() {
                 }
             } else {
                 num_slices += ep->acknowledge(slice, COMPLETED);
+                // A successful GPU transfer re-admits any learned GDR
+                // unreachability for the (GPU, NIC) pair(s) it used, so a
+                // transient exclusion (or a recovered path) heals. Skipped
+                // entirely until something has actually been excluded.
+                if (GdrReachability::hasAnyExclusion()) {
+                    auto& gdr = GdrReachability::instance();
+                    if (slice->source_gpu_ordinal >= 0 &&
+                        slice->source_nic_name)
+                        gdr.reportLocalSuccess(slice->source_nic_name,
+                                               slice->source_gpu_ordinal);
+                    if (slice->target_gpu_ordinal >= 0 &&
+                        slice->target_nic_name && slice->target_machine_id)
+                        gdr.reportRemoteSuccess(*slice->target_machine_id,
+                                                slice->target_nic_name,
+                                                slice->target_gpu_ordinal);
+                }
                 // A successful transfer proves this rail is healthy; clear
                 // any accumulated error count so a previously-cooled-down
                 // rail can be used again without waiting for the full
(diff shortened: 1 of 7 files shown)


mooncake-transfer-engine/tent/src/transport/rdma/workers.cpp
```diff
@@ -667,6 +670,27 @@ void Workers::asyncPollCq() {
...
+                if (local_gdr_err && slice->source_gpu_ordinal >= 0 &&
+                    slice->source_nic_name) {
+                    GdrReachability::instance().reportLocalFailure(
+                        slice->source_nic_name, slice->source_gpu_ordinal);
+                } else if (remote_gdr_err && slice->target_gpu_ordinal >= 0 &&
+                           slice->target_nic_name && slice->target_machine_id) {
+                    GdrReachability::instance().reportRemoteFailure(
+                        *slice->target_machine_id, slice->target_nic_name,
+                        slice->target_gpu_ordinal);
+                }
```
alogfans, line 693: Is it possible to check reachability before ibv_post_send? Of course these should be preserved.

mooncake-transfer-engine/tent/src/transport/rdma/workers.cpp
```diff
@@ -969,11 +1018,25 @@ Status Workers::selectFallbackDevice(RouteHint& source, RouteHint& target,
         dst_total += target.topo_entry->device_list[trank].size();
```
alogfans, line 1053: Fallback now scans from combination 0 on every retry. Since RailMonitor requires three errors by default before excluding a path, non-GDR failures may repeatedly retry the same path until the retry budget is exhausted. Please rotate through all combinations with wraparound, for example (previous_index + offset) % total_combos, while retaining the GDR filtering.

mooncake-transfer-engine/tent/src/transport/rdma/workers.cpp
anranxia: Good point — done in 73d2471f. `selectOptimalDevice` (the retry-0 primary path) now consults `GdrReachability` before posting: if the chosen (source NIC, target NIC) pair is already known-unreachable, it routes through `selectFallbackDevice` to re-select a reachable pair, reusing the existing `rail.available()` redirect. So both selection paths now filter before `ibv_post_send`. The reactive learning in `asyncPollCq` is preserved as-is (it's what seeds the exclusions in the first place). I also factored the shared check into a `gdrPairExcluded()` helper so the primary and fallback paths don't duplicate the logic.

mooncake-transfer-engine/tent/src/transport/rdma/workers.cpp
anranxia: Agreed, and fixed in 73d2471f. The fallback now rotates with wraparound as you suggested, resuming just past the pair the slice tried last:

      start = (slice->last_fallback_idx + 1) % total_combos;
      for (k = 0; k < total_combos; ++k) { idx = (start + k) % total_combos; ... }

  `last_fallback_idx` is a per-slice field seeded to -1 (reset alongside retry_count when the slice is acquired), so the *first* fallback starts at flat index 0 — the (PIX, PIX) pair — and only *subsequent* retries advance. This keeps your non-GDR case from burning the whole budget on one path (it rotates through all combinations before RailMonitor's threshold kicks in), while still letting a strict-topology transfer converge onto the one GPUDirect-capable pair on the first fallback. GDR-excluded and rail-down pairs are skipped throughout, and the  chosen index is written back so the next retry continues the rotation.

anranxia:
Rebased onto latest main to resolve the conflict. The only clash was in workers.cpp with #3000 (which moved lkey/rkey assignment into generatePostPath); I adopted that refactor and kept only my last_fallback_idx write in selectFallbackDevice. No functional changes since the last review.

anranxia:
Fixed the code-format CI failure in 90ab4ca0 — reformatted workers.cpp with clang-format 20.1.8 to match CI (whitespace only, no functional change).

anranxia:
Rebased onto latest main (a6b4db4c). The earlier build failures were unrelated to this PR — the real error was in mooncake-store/real_client.cpp (which this PR doesn't touch), from the test-merge hitting an in-flight main state; the rest were fail-fast cancellations. Should be clear now that main is green again.

anranxia:
@alogfans Both review comments are addressed (proactive reachability check + wraparound fallback rotation), rebased onto latest main, and all CI checks are green. Could you approve and merge when convenient? Thanks!

alogfans (approved):
LGTM

PR #3283, opened 2026-08-04
[CI/Build] Simplify wheel workflows

Aionw:
## Description

Simplify the standard wheel workflows by moving supported build-profile details into `_build-wheel.yaml` and having callers select only a package variant and architecture.

This change:

- centralizes runner, container, CMake, CUDA architecture, EP, and allocator settings for CUDA 12.8, CUDA 13, and non-CUDA wheels;
- replaces repeated nightly and pre-release build jobs with reusable-workflow matrices;
- replaces duplicated x86/ARM release jobs with matrices;
- calls the shared CUDA 13 wheel workflow directly from `ci.yml` and removes the one-layer `ci_cu13.yml` wrapper; and
- keeps artifact names, Python version coverage, release publication dependencies, and nightly gating intact.

The workflow set is reduced by 166 net lines, leaving each caller focused on orchestration rather than build implementation details. There is no user-facing runtime behavior change.

### Related work

Open PR #3269 also touches the CUDA wheel configuration to enable NCCL transports. This PR does not implement that feature: it centralizes the configuration location. If #3269 lands first, its flags need to be carried into the centralized CUDA x86 profiles in `_build-wheel.yaml`; if this PR lands first, #3269 can update that single location instead of both CI workflow files.

No matching open issue or PR was found for the workflow-deduplication scope itself. Although the mechanical diff exceeds 500 changed lines, this is an internal deletion-heavy refactor rather than a new architecture, so no RFC issue was opened.

## Module

- [x] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

The workflow graph and profile mappings were reviewed against the previous jobs, including runners, container images, CMake flags, artifact prefixes, Python matrices, and downstream `needs` relationships.

**Test commands:**
```bash
SKIP=codespell pre-commit run --from-ref origin/main --to-ref HEAD
python3 <workflow YAML duplicate-key validation script>
```

**Test results:**
- [ ] Unit tests pass
- [x] Manual testing done: YAML parsing, duplicate-key validation, changed-file pre-commit hooks, and diff checks passed

`codespell` was skipped because it reports the existing Rust term `crate` in `ci.yml` as `create`; all other applicable pre-commit hooks passed. A live GitHub Actions run is still required to exercise the reusable-workflow matrices.

## Checklist

- [ ] I have formatted my code using `./scripts/code_format.sh` (not applicable; no C/C++ changes)
- [ ] I have run `pre-commit run --all-files` and all hooks pass (changed-file hooks passed except the documented `crate` codespell false positive)
- [ ] I have updated the documentation (not applicable; no user-facing behavior)
- [ ] I have added tests to prove my changes are effective (workflow static validation only)
- [ ] For changes >500 LOC: I have filed an RFC issue (internal deletion-heavy refactor; no new architecture)

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex inspected the existing workflows, implemented the deduplication, ran static validation, and reviewed the resulting diff. The PR remains draft pending human review of every changed line and live GitHub Actions validation.


Icedcoco:
Thanks for the refactor — having all the wheel profiles in one place makes the matrix much easier to follow. I went through the six variant/architecture combinations and they line up with the old jobs, and CI looks good too.

One thing I wanted to ask about:

The build step now uses `cmake --build . --parallel 128` instead of the old `-j"$(nproc)"`. Was that intentional? On the standard GitHub runners (2–4 vCPUs) that looks like it could oversubscribe CPU and memory, especially for cold nightly/release builds. CI passing suggests it's fine in practice, but I'm curious if there was a reason behind the fixed number — e.g., `nproc` misbehaving inside the manylinux containers — or whether keeping `--parallel "$(nproc)"` would be safer.

Aionw:
`make -j128` fixed in https://github.com/kvcache-ai/Mooncake/pull/3283/commits/8f17ef6a2aa1d0927265df0f3cecff91e06e02f6. Now it's `nproc`.

Icedcoco:
LGTM pending CI

ykwd (approved):

PR #3285, opened 2026-08-04
[TransferEngine] Share one host KV segment across a TP group

ascend-direct-dev:
## Motivation

MLA keeps the same KV cache in every rank of a tensor-parallel group. When those ranks offload KV to host memory during decode, each one allocates its own copy, so host usage grows with the TP size while every copy holds identical bytes.

This PR adds a *shared segment*: the owner rank allocates the pages once through the platform's virtual memory manager, and every other rank maps the very same physical memory. Host usage for offloaded MLA KV drops to 1/TP.

## Design

C++ shares one contiguous host span. Tensor layout (which offset holds which layer's K/V) is computed in Python and folded into the segment name so peers that disagree still fail the fingerprint check.

**Two-phase creation.** The handle exchange belongs to the caller's process group:

1. `SharedSegment::Create` reserves address space (owner also allocates and exports) and returns a fixed-size blob.
2. The caller `all_gather`s the blob over its own group.
3. `segment->Complete(blobs)` maps the owner's pages and verifies that every rank declared the same thing.

Doing it this way avoids calling back into Python collectives from C++ under the GIL. Non-owner blobs are the same length as the owner's, so a plain fixed-size all-gather carries them.

**Addresses are not forced to match.** Ranks agree on the byte layout, not on virtual addresses, so each rank reserves `size` rather than `size * world_size`.

**Backends.** Compiled in per platform and both optional:

| Platform | Mechanism | Handle |
| --- | --- | --- |
| Ascend | ACL VMM via `adxl::AdxlEngine` | `aclrtMemFabricHandle` |
| NVIDIA | CUDA VMM (`cuMemCreate`/`cuMemMap`) | `CUmemFabricHandle`, CUDA 12.4+ |

`SharedSegment::Supported()` reports whether the running system can actually share memory. Builds without either backend return `kNotImplemented` rather than failing to link.

## Python API

```python
from mooncake.shared_segment import create_shared_segment

seg = create_shared_segment(
    "vllm_sparse_kv",
    blocks={
        "k": dict(count=num_layers, shape=k_shape, dtype=torch.bfloat16),
        "v": dict(count=num_layers, shape=v_shape, dtype=torch.bfloat16),
    },
    world_size=tp_size,
    rank_id=tp_rank,
    tp_group=tp_group,
)
k_caches_cpu = seg.tensors("k")  # zero-copy views onto the shared pages
```

## Test plan

- [x] `shared_segment_test`: fingerprint, blob encode/decode, and the full two-phase protocol against a fake backend.
- [x] Ascend build and pybind smoke test on a CANN toolchain.
- [x] CUDA backend compiles clean under `-Wall -Wextra` against the CUDA driver headers.
- [x] Multi-rank run on real Ascend hardware, paired with HiXL https://gitcode.com/cann/hixl/merge_requests/851


ascend-direct-dev:
[test_create_shared_segment_d2rh.py](https://github.com/user-attachments/files/30889545/test_create_shared_segment_d2rh.py)

mooncake-transfer-engine/src/shared_segment/shared_segment_internal.h
```diff
@@ -0,0 +1,127 @@
...
+// for HostRegister.
+std::unique_ptr<SharedSegmentBackend> CreateMmapSharedSegmentBackend();
+
+#ifdef USE_ASCEND_DIRECT
+std::unique_ptr<SharedSegmentBackend> CreateAscendSharedSegmentBackend();
+#endif
+
+#ifdef USE_CUDA
+std::unique_ptr<SharedSegmentBackend> CreateCudaSharedSegmentBackend();
+#endif
```
alogfans: Factory method should be better, e.g., `std::unique_ptr<SharedSegmentBackend> CreateSharedSegmentBackend(std::string &impl)`

mooncake-transfer-engine/include/transport/ascend_transport/ascend_direct_transport/adxl_compat.h
```diff
@@ -184,6 +188,17 @@ class ASCEND_FUNC_VISIBILITY AdxlEngine {
...
                                                   void** ptr);
 
+    /**
+     * @brief 将MallocMem申请的内存导出为fabric share handle.
+     * ACL export 每个分配最多一次：首次导出并缓存，之后返回缓存 handle.
+     * @param [in] addr MallocMem返回的虚拟内存ptr
+     * @param [out] handle 导出的fabric share handle
+     * @return 成功:SUCCESS, 地址非MallocMem申请或已释放:PARAM_INVALID,
+     * 失败:其它.
+     */
```
alogfans, line 201: Formalize the code comments as possible.

alogfans:
Feel free to mark it ready for review after performing coding style changes.

mooncake-transfer-engine/include/transport/ascend_transport/ascend_direct_transport/adxl_compat.h
ascend-direct-dev: Updated in b1e17e99. The ExportToShareableHandle Doxygen comment is now formalized in English, including the one-time ACL export and cache behavior, parameter descriptions, and return semantics.

mooncake-transfer-engine/src/shared_segment/shared_segment_internal.h
ascend-direct-dev: Updated in b1e17e99. The internal header now exposes a single CreateSharedSegmentBackend(const std::string& impl) factory. Backend-specific creators are kept as implementation details in shared_segment.cpp, and an unknown-implementation regression test was added.

mooncake-transfer-engine/include/transport/ascend_transport/ascend_direct_transport/adxl_compat.h
```diff
@@ -184,6 +188,21 @@ class ASCEND_FUNC_VISIBILITY AdxlEngine {
...
+     * @param [out] handle Exported fabric shareable handle.
+     * @return SUCCESS on success, PARAM_INVALID if addr was not returned by
+     * MallocMem or has already been freed, and another error status otherwise.
+     */
+    __attribute__((weak)) static Status ExportToShareableHandle(
+        void* addr, ShareableHandle& handle);
+
     /**
      * @brief 释放MallocMem申请的内存内存
      * @param [in] ptr 释放的虚拟内存ptr
```
alogfans: Translate as well.

mooncake-integration/shared_segment.py
```diff
@@ -0,0 +1,319 @@
...
+    try:
+        if world_size == 1:
+            blobs = [blob]
+        elif tp_group is None:
+            raise SharedSegmentError("tp_group is required when world_size > 1")
+        else:
+            blobs = _all_gather_blob(blob, world_size, tp_group)
+        segment.complete(blobs)
+    except RuntimeError as exc:
+        raise SharedSegmentError(str(exc)) from exc
```
alogfans: Once a rank raises an error, the whole cluster cannot continue or shutdown gracefully.

mooncake-integration/shared_segment.py
```diff
@@ -0,0 +1,319 @@
...
+        )
+
+    # Sorted names keep the byte layout identical across ranks regardless of
+    # dict iteration order.
+    names = sorted(blocks)
+    specs = {
+        block_name: _parse_block(block_name, blocks[block_name])
+        for block_name in names
+    }
+    offsets, stride, total = _build_layout(specs, names)
```
alogfans, line 101: Should precesion (bf16/fp16) and/or layout be considered?

mooncake-transfer-engine/src/CMakeLists.txt
```diff
@@ -11,6 +11,11 @@ if(USE_HIP)
 endif()
 
 add_library(transfer_engine ${ENGINE_SOURCES} $<TARGET_OBJECTS:transport>)
+# Private headers live next to the .cpp sources. CUDA builds compile from the
+# source tree so quoted includes work; USE_HIP hipifies into the build tree and
+# needs this path explicitly.
+target_include_directories(transfer_engine
+                           PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/shared_segment)
```
alogfans: This requires CUDA 12.4 or above.

mooncake-transfer-engine/include/transport/ascend_transport/ascend_direct_transport/adxl_compat.h
ascend-direct-dev: Updated in d4bac404030aea6c03120266efc2168748dab025. All remaining Chinese Doxygen comments in adxl_compat.h have been translated to English.

mooncake-integration/shared_segment.py
ascend-direct-dev: Updated in d4bac404030aea6c03120266efc2168748dab025. Shared-segment creation and completion now perform rank-wide error consensus with all_gather_object, so every rank observes the same failure and exits collectively instead of one rank failing alone.

mooncake-integration/shared_segment.py
ascend-direct-dev: Updated in d4bac404030aea6c03120266efc2168748dab025. Layout tags now include block name, count, shape, dtype, and byte size; the declaration fingerprint therefore rejects dtype/precision and layout mismatches across ranks.

mooncake-transfer-engine/src/CMakeLists.txt
ascend-direct-dev: Updated in d4bac404030aea6c03120266efc2168748dab025. The CUDA shared-segment backend has been removed for now, including the CUDA mmap HostRegister path. This feature currently provides POSIX mmap and Ascend VMM/HostRegister support; CUDA can be added separately if needed.

alogfans (approved):
LGTM. Thanks!

PR #3286, opened 2026-08-04
[Build] conductor: default RelWithDebInfo for standalone builds

Chase-Rong:
## Description

Standalone conductor builds (`mooncake-conductor/CMakeLists.txt`, `NOT GLOBAL_CONFIG`) never included `mooncake-common/common.cmake`, whose fallback sets `CMAKE_BUILD_TYPE=RelWithDebInfo`. As a result, a standalone conductor build got **no optimization flags at all (-O0)** while every top-level component built with -O2.

This was the dominant factor in conductor `/query` latency: the vLLM v1 hash chain costs **2311 ns/block at -O0 vs 344 ns/block at -O2 (6.7×)** — an optimization-level gap, not an API gap. End-to-end, a 1M-token 100%-hit `/query` drops from 185.9 ms to ~28 ms with this fix alone (cold 1M: 18.7 → 4.4 ms).

Debug info (`-g`) is preserved. The source tree contains no `assert()`/`DCHECK`, so `NDEBUG` changes nothing. When built via the top-level project (`GLOBAL_CONFIG`), the new fallback never runs and behavior is unchanged.

The SHA-256 API swap that was originally in this PR has been split out to a follow-up PR per review feedback.

## Module

- [x] Other — `mooncake-conductor`

## Type of Change

- [x] Performance improvement
- [x] Other: build fix (default build type for standalone conductor)

## How Has This Been Tested?

**Test commands:**
```bash
cmake --build mooncake-conductor/build -j2 && ./mooncake-conductor/build/tests/conductor_test
pre-commit run --files mooncake-conductor/CMakeLists.txt
```

**Test results:**
- [x] Unit tests pass — 161/161
- [x] Integration tests pass — vLLM ×2 + mooncake_master + conductor topology: tiered-hit routing, stickiness, zero rejected KV events
- [x] Manual testing done — `/query` latency bench cold vs 100%-hot (numbers above); verified generated flags are `-O2 -g -DNDEBUG`

## Checklist

- [x] I have run `pre-commit run --all-files` and all hooks pass (on touched files)
- [x] I have updated the documentation (if applicable) — N/A
- [x] I have added tests to prove my changes are effective — N/A (build-system only; covered by existing tests)
- [x] For changes >500 LOC: I have filed an RFC issue — N/A (6 LOC)

## AI Assistance Disclosure

Claude Code: root-cause analysis (-O0 build identified as the dominant cost via micro-benchmark), implementation, measurement. The human submitter has reviewed every changed line.

### DIFF
--- a/mooncake-conductor/CMakeLists.txt
+++ b/mooncake-conductor/CMakeLists.txt
@@ -6,6 +6,10 @@ if(NOT GLOBAL_CONFIG)
   set(CMAKE_CXX_STANDARD_REQUIRED ON)
   include(../mooncake-common/FindJsonCpp.cmake)
   include(../mooncake-common/FindGLOG.cmake)
+  # Match the top-level common.cmake policy: keep debuginfo by default.
+  if(NOT CMAKE_BUILD_TYPE)
+    set(CMAKE_BUILD_TYPE "RelWithDebInfo")
+  endif()
 endif()
 
 find_package(cppzmq REQUIRED)


he-yufeng:
Two very different changes bundled here, and I would score them separately.

The `RelWithDebInfo` default is clearly right. I checked the standalone path: `mooncake-conductor/CMakeLists.txt` under `NOT GLOBAL_CONFIG` never included `common.cmake`, so it really did build at -O0 while the top-level default is RelWithDebInfo. That is also the dominant factor in your 10x number (2311 vs 344 ns/block is an optimization-level gap, not an API gap).

The API swap I am less sure about, for two reasons:

1. The comparison is against the wrong baseline. The code you are deleting already had `Sha256Reuse`, which amortizes EVP setup over the whole chain with a caller-owned context. The fair benchmark for the swap is EVP-with-reuse vs SHA256_*, not EVP-fresh-per-block vs SHA256_*, and the -O2 numbers in the PR are presumably the former. If reuse-vs-low-level is within noise, the swap buys nothing.
2. `SHA256_Init/Update/Final` are deprecated since OpenSSL 3.0 and disappear entirely in `no-deprecated` builds and under FIPS providers, where the EVP layer is the only sanctioned path. `OPENSSL_SUPPRESS_DEPRECATED` hides the warning but does not bring the symbols back on those platforms, so this takes a portability hit that RelWithDebInfo does not need.

Suggestion: land the CMake fix on its own (it is most of the win), and only do the API swap if a reuse-vs-low-level benchmark on the same -O2 build still shows a meaningful gap. If it does, note the FIPS/no-deprecated constraint in the PR so CI can cover or gate it.

Chase-Rong:
> Two very different changes bundled here, and I would score them separately.
> 
> The `RelWithDebInfo` default is clearly right. I checked the standalone path: `mooncake-conductor/CMakeLists.txt` under `NOT GLOBAL_CONFIG` never included `common.cmake`, so it really did build at -O0 while the top-level default is RelWithDebInfo. That is also the dominant factor in your 10x number (2311 vs 344 ns/block is an optimization-level gap, not an API gap).
> 
> The API swap I am less sure about, for two reasons:
> 
> 1. The comparison is against the wrong baseline. The code you are deleting already had `Sha256Reuse`, which amortizes EVP setup over the whole chain with a caller-owned context. The fair benchmark for the swap is EVP-with-reuse vs SHA256__, not EVP-fresh-per-block vs SHA256__, and the -O2 numbers in the PR are presumably the former. If reuse-vs-low-level is within noise, the swap buys nothing.
> 2. `SHA256_Init/Update/Final` are deprecated since OpenSSL 3.0 and disappear entirely in `no-deprecated` builds and under FIPS providers, where the EVP layer is the only sanctioned path. `OPENSSL_SUPPRESS_DEPRECATED` hides the warning but does not bring the symbols back on those platforms, so this takes a portability hit that RelWithDebInfo does not need.
> 
> Suggestion: land the CMake fix on its own (it is most of the win), and only do the API swap if a reuse-vs-low-level benchmark on the same -O2 build still shows a meaningful gap. If it does, note the FIPS/no-deprecated constraint in the PR so CI can cover or gate it.

Thanks for your suggestions.  The PR is now split.

1. CMake fix on its own. This PR is now only the RelWithDebInfo default. The API swap moved to a follow-up PR so this one can land clean.

2. Reuse-vs-low-level on the same -O2 build  to be clear about the baseline: the benchmark was never fresh-ctx-per-block. The "EVP reused" case keeps one EVP_MD_CTX for the whole chain and does reset + init + update + final per block, which is exactly what Sha256Reuse did. Numbers on the -O2 build (best of 5, 200K blocks,  ~100-byte inputs):

- EVP, reused ctx (= old code): 197 ns/block
- EVP, fresh ctx each block: 205 ns/block
- low-level SHA256_Init/Update/Final: 68 ns/block

The part I found interesting: reuse barely b heap allocation Sha256Reuse was saving is only ~4% of what EVP charges per call. The rest is the per-block reset+init — provider-side reinit and the indirect dispatch — and you can't amortize that away

I also re-ran this end-to-end after the split to put a real number on it rather than extrapolating: 1M-token query, 100% hit, same machine — 28.6 ms with just this PR, 19.2 ms with the swap on top. So it's roughly a third off again,not noise.

3. FIPS / no-deprecated. You're right on the mechanics — I checked openssl/sha.h and the declarations really are behind OPENSSL_NO_DEPRECATED_3_0, so OPENSSL_SUPPRESS_DEPRECATED only hides the warning, it doesn't keep those builds working. I asked kimi about how to handle this cleanly, and it suggested an approach that looks right to me:
- At configure time, a CMake compile probe (ries to build a tiny snippet callingSHA256_Init. If it compiles, we use the low-o-deprecated builds), we automatically fallback to the current EVP-reuse code — no plat the user.- FIPS is a runtime policy, not something a  there'd also be an explicit CMake option toforce the EVP path for FIPS-conscious deploy- CI can then build both configurations (one_DEPRECATED_3_0 defined) so the fallback pathgets compiled and tested instead of silentlyThat goes a bit beyond noting the constraintr breaks a build — worst case on thoseplatforms is the current EVP performance. Happy to adjust if you'd rather keep it to documentation + CI gating only.

PR #3287, opened 2026-08-04
[TransferEngine] Add NCCL runtime loading support

UNIDY2002:
## Description

Adds Python selection and validation for NCCL host RMA transport, and dynamically resolves NCCL symbols for both host RMA and DeviceTransport. Mooncake imports without loading NCCL; selecting either NCCL transport reports an unavailable capability when the process library is absent, incomplete, or incompatible. DeviceTransport requires an exact build-header/runtime NCCL version match and documents using `LD_PRELOAD` to select that runtime before process start.

## Module

- [ ] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [x] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
PYTHONPATH=mooncake-wheel python3 -m unittest mooncake-wheel/tests/test_mooncake_config.py mooncake-wheel/tests/test_nccl_host_transport.py
cmake -DUSE_NCCL_DEVICE=ON -DUSE_NCCL_HOST=ON ...
```

**Test results:**
- [x] Unit tests pass

The Python configuration suite passes 35 tests; the NCCL host integration test skips without a CUDA-enabled Mooncake wheel and PyTorch. On two H20 GPUs with CUDA 13.0, Torch 2.11.0, and NCCL 2.30.7, the combined host/device build succeeded, `engine.so` had no `libnccl.so.2` dependency, the host RMA write/read-rejection test passed, and the two-process DeviceTransport example passed. With Torch's older NCCL 2.28.9 loaded first, `import engine` succeeded and DeviceTransport reported the missing `ncclCommQueryProperties` capability instead of failing module loading. Selecting NCCL 2.30.7 before startup made the DeviceTransport probe available.

## Checklist

`pre-commit` is not installed in the local environment; CI will run the repository hooks.

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex assisted with implementation, focused test coverage, and validation documentation. The submitter reviewed the runtime-loading contract and validation results.


UNIDY2002:
@xiaofanl-nvidia We are considering shipping the NCCL host and DeviceTransport support as part of Mooncake's default distribution.

A practical issue is that many user environments load an older NCCL. If Mooncake directly links newer NCCL symbols, importing `mooncake.engine` can fail immediately with an undefined-symbol error before the user even selects an NCCL transport.

This draft PR changes the host-side NCCL calls to runtime loading. Normal Mooncake import no longer requires NCCL symbols; selecting an NCCL transport probes the process `libnccl.so.2` and reports the capability as unavailable when the library is missing, incomplete, or incompatible. DeviceTransport still requires an exact build-header/runtime NCCL version match, and users can select the matching runtime with `LD_PRELOAD` before process startup.

The tradeoff seems real:

- Pros: Mooncake can ship the NCCL support without turning an older user NCCL into a package-wide import regression. Users with a compatible runtime can use the feature immediately.
- Cons: this adds substantial complexity: dynamic symbol tables, runtime capability/error handling, and a more nuanced version contract. It does not make an older NCCL compatible with the newer APIs; it only turns incompatibility into an explicit transport-availability failure.

Would you recommend adopting this dynamic-loading approach now, or waiting to enable/distribute the NCCL integration until we can reasonably assume end-user environments have sufficiently new NCCL versions?

he-yufeng:
Read through the runtime-loading shape. Moving the version check and symbol resolution behind dlopen so mooncake imports cleanly without NCCL present, and reporting the capability on select instead of at import time, is the right shape. Worth noting the exact build==runtime check is not new here, it is the pre-existing DeviceTransport rule now being exercised at select time, which is the better place for it.

Two things I would want answered before merging:

1. Handle semantics in mixed processes. `dlopen("libnccl.so.2", RTLD_LOCAL)` on an already-loaded library returns the existing handle, so if the process loaded a different NCCL first (PyTorch is the common case), the runtime check compares against THAT version, not the one the wheel would have preferred. The LD_PRELOAD guidance covers it, but the error message should probably say "already loaded by another component" when the handle pre-existed, or users will chase a mismatch that is really a load-order problem.

2. Symbol-table completeness. The new table must cover every NCCL entry point used across host RMA and DeviceTransport (send/recv, comm init/destroy, error strings, stream ops). A missing one will not fail at load, it will fail lazily at first use, which is much harder to diagnose in production. A static assert or a one-time resolution that enumerates all required symbols up front and reports the first missing one would make the incomplete-library case explicit instead of latent.

PR #3289, opened 2026-08-04
[Bugfix][TE] Serialize libnuma's lazy cpumask cache fill in bindToSocket

SongOf:
Description

  bindToSocket() calls numa_node_to_cpus(), whose per-node cpumask cache libnuma builds lazily and without any locking. Upstream says so directly in libnuma.c:

  /* This would be better with some locking, but I don't want to make libnuma
     dependent on pthreads right now. The races are relatively harmless. */

  Worker pools bind every thread they spawn at startup, so several threads hit that cold-start path at once. Each allocates the cache and every allocation but the last is orphaned. LeakSanitizer reports the orphan and fails the whole test binary even when every assertion passed — which is what broke rdma_async_event_drain_test in CI:

  [  PASSED  ] 5 tests.
  ==16537==ERROR: LeakSanitizer: detected memory leaks
  Direct leak of 8192 byte(s) ... in numa_node_to_cpus (libnuma.so.1+0x69b9)

  This guards the one lazily-initialized call. Everything else bindToSocket() touches reads state set by libnuma's __attribute__((constructor)) numa_init(), which runs single-threaded at load time, so the lock is deliberately scoped to numa_node_to_cpus() alone.

  The mutex lives in an inline function rather than a static local inside bindToSocket(): that function is static inline, so a static local would be per-translation-unit and worker_pool.cpp, ub_context.cpp and transfer_task.cpp would each lock a different mutex.

  tent carries an independent copy of bindToSocket() with the same defect, fixed identically.

  Relationship to #3243. That PR already made CI green by suppressing the report via __lsan_default_suppressions() returning "leak:libnuma.so" — a fair call for the immediate failure, since the leak is bounded and one-shot.
  This change is complementary rather than a duplicate: it removes the race instead of hiding the symptom. Two practical differences: the suppression symbol is defined in rdma_async_event_drain_test.cpp and so only links into that one binary, leaving any other ASAN test that starts a WorkerPool unprotected; and concurrent unsynchronized writes to libnuma's globals are a data race, not only a leak. The suppression can stay — the two are independent.

  Module
  
  - [ ] CI/CD
  - [ ] Docs
  - [ ] Other

  Type of Change

  - [x] Bug fix
  - [ ] New feature
  - [ ] Refactor
  - [ ] Breaking change
  - [ ] Documentation update
  - [ ] Performance improvement
  - [ ] Other
  
  How Has This Been Tested?

  Test commands:
  cmake -G Ninja .. -DENABLE_ASAN=ON -DCMAKE_BUILD_TYPE=Debug -DBUILD_UNIT_TESTS=ON
  cmake --build . --target common_test
  ./mooncake-transfer-engine/tests/common_test --gtest_filter='BindToSocketTest.*'

  Test results:

  Verified under ASAN in an ubuntu:22.04 container, driving 32 threads through bindToSocket() behind a barrier:

  │                 Subject                                              │      Result       │
  │ common.h, before the fix                                │ leaked 10/10 runs │
  │ common.h, after the fix                                   │ clean 12/12 runs  │
  │ tent/.../os.h, after the fix                                  │ clean 12/12 runs  │
  │ Minimal libnuma repro (no Mooncake code) │ leaked 8/8 runs   │

  The minimal repro isolates the defect to libnuma itself — the leak stack contains no Mooncake frames.

  - [ ] Unit tests pass

  Not verified locally, please watch CI. A full CMake build of the test suite was not completed on my machine (Apple Silicon; dependencies.sh kept failing on ports.ubuntu.com 502s). The runs above compile the real headers under ASAN but do not link the full transfer_engine. Note also that nodemask_sz varies by host, so the leaked byte count on arm64 is not the 8192 seen on CI's x86_64 — same mechanism, different size.

  Known gaps. The tent fix ships without a test: the ASAN job does not set USE_TENT and the tent-ci job does not set ENABLE_ASAN, so a regression test there could never fail. Enabling ASAN on tent-ci looks feasible, but tent has never run under a sanitizer and may surface pre-existing leaks, so that belongs in its own PR. Separately, the two mutexes are distinct objects (mooncake:: and mooncake::tent::), leaving a theoretical window if a legacy pool and a tent pool cold-start simultaneously in one process.

  Checklist
  
  AI Assistance Disclosure

  - [ ] No AI tools were used

### DIFF
--- a/mooncake-transfer-engine/include/common.h
+++ b/mooncake-transfer-engine/include/common.h
@@ -32,6 +32,7 @@
 #include <ctime>
 #include <iomanip>
 #include <iostream>
+#include <mutex>
 #include <optional>
 #include <sstream>
 #include <string_view>
@@ -76,6 +77,16 @@ static inline std::string getHostname() {
     return hostname;
 }
 
+// libnuma fills the cache numa_node_to_cpus() reads lazily and without locking,
+// so concurrent first callers each allocate it and all but one are orphaned --
+// a leak LeakSanitizer fails the build on. Worker pools bind every thread at
+// startup, so they hit that window. An inline function, not a static local:
+// bindToSocket() has internal linkage, so a static local would be per-TU.
+inline std::mutex &numaNodeCpuCacheMutex() {
+    static std::mutex mutex;
+    return mutex;
+}
+
 static inline int bindToSocket(int socket_id) {
     if (unlikely(numa_available() < 0)) {
         LOG(WARNING) << "The platform does not support NUMA";
@@ -86,7 +97,10 @@ static inline int bindToSocket(int socket_id) {
     if (socket_id < 0 || socket_id >= numa_num_configured_nodes())
         socket_id = 0;
     struct bitmask *cpu_list = numa_allocate_cpumask();
-    numa_node_to_cpus(socket_id, cpu_list);
+    {
+        std::lock_guard<std::mutex> guard(numaNodeCpuCacheMutex());
+        numa_node_to_cpus(socket_id, cpu_list);
+    }
     int nr_possible_cpus = numa_num_possible_cpus();
     int nr_cpus = 0;
     for (int cpu = 0; cpu < nr_possible_cpus; ++cpu) {
--- a/mooncake-transfer-engine/tent/include/tent/common/utils/os.h
+++ b/mooncake-transfer-engine/tent/include/tent/common/utils/os.h
@@ -27,6 +27,7 @@
 #include <ctime>
 #include <iomanip>
 #include <iostream>
+#include <mutex>
 #include <sstream>
 #include <thread>
 
@@ -41,6 +42,16 @@
 
 namespace mooncake {
 namespace tent {
+// libnuma fills the cache numa_node_to_cpus() reads lazily and without locking,
+// so concurrent first callers each allocate it and all but one are orphaned --
+// a leak LeakSanitizer reports. Workers bind every thread at startup, so they
+// hit that window. An inline function, not a static local: bindToSocket() has
+// internal linkage, so a static local would be per-TU.
+inline std::mutex &numaNodeCpuCacheMutex() {
+    static std::mutex mutex;
+    return mutex;
+}
+
 static inline int bindToSocket(int socket_id) {
     if (numa_available() < 0) {
         LOG(WARNING) << "The platform does not support NUMA";
@@ -51,7 +62,10 @@ static inline int bindToSocket(int socket_id) {
     if (socket_id < 0 || socket_id >= numa_num_configured_nodes())
         socket_id = 0;
     struct bitmask *cpu_list = numa_allocate_cpumask();
-    numa_node_to_cpus(socket_id, cpu_list);
+    {
+        std::lock_guard<std::mutex> guard(numaNodeCpuCacheMutex());
+        numa_node_to_cpus(socket_id, cpu_list);
+    }
     int nr_possible_cpus = numa_num_possible_cpus();
     int nr_cpus = 0;
     for (int cpu = 0; cpu < nr_possible_cpus; ++cpu) {
--- a/mooncake-transfer-engine/tests/common_test.cpp
+++ b/mooncake-transfer-engine/tests/common_test.cpp
@@ -5,6 +5,7 @@
 #include <optional>
 #include <string>
 #include <thread>
+#include <vector>
 
 #include "common.h"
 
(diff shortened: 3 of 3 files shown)


he-yufeng:
The diagnosis and the fix shape are right. The inline-function-static-local is the correct way to get one process-wide mutex from an internal-linkage header, and scoping the lock to just the `numa_node_to_cpus` call keeps it cheap. The 32-thread barrier test is honest about only proving it under ASAN.

One coverage question worth an answer before merge: `common.h` and `tent/common/utils/os.h` now each carry their own copy of the mutex. If a process links both (TE and TENT code paths in one binary), the two subsystems serialize against themselves but not against each other, and the same libnuma cache fill can still race across the two mutexes. If that combination never happens in practice, a one-line note saying so closes the gap; otherwise the mutex wants to live in a shared header both include.

Minor: `numa_allocate_cpumask()` sits outside the lock, which is fine since it is per-thread state; just confirming that was deliberate and not an oversight in the second file too.

SongOf:
> The diagnosis and the fix shape are right. The inline-function-static-local is the correct way to get one process-wide mutex from an internal-linkage header, and scoping the lock to just the `numa_node_to_cpus` call keeps it cheap. The 32-thread barrier test is honest about only proving it under ASAN.
> 
> One coverage question worth an answer before merge: `common.h` and `tent/common/utils/os.h` now each carry their own copy of the mutex. If a process links both (TE and TENT code paths in one binary), the two subsystems serialize against themselves but not against each other, and the same libnuma cache fill can still race across the two mutexes. If that combination never happens in practice, a one-line note saying so closes the gap; otherwise the mutex wants to live in a shared header both include.
> 
> Minor: `numa_allocate_cpumask()` sits outside the lock, which is fine since it is per-thread state; just confirming that was deliberate and not an oversight in the second file too.

Thanks — both points answered below.

  Cross-subsystem coverage. Verified: TE and TENT are selected at runtime, not composed. TransferEngine's constructor picks one stack per instance from a process-wide env var (transfer_engine.cpp:376):

  if (getenv("MC_USE_TENT") || getenv("MC_USE_TEV1")) {
      use_tent_ = true;
  }
  if (!use_tent_) {
      impl_ = std::make_shared<TransferEngineImpl>(auto_discover);
  }

  Since every instance in a process reads the same environment, one stack runs and the other's worker pools are never started — so the two mutexes never contend for the same libnuma cache fill in practice. 

  numa_allocate_cpumask() outside the lock — deliberate, in both files. I checked it against the libnuma 2.0.14 source rather than assuming:

  numa_allocate_cpumask()
  {
        int ncpus = numa_num_possible_cpus();
        return numa_bitmask_alloc(ncpus);
  }

  int numa_num_possible_cpus(void) { return cpumask_sz; }

  cpumask_sz is set by set_sizes() under __attribute__((constructor)) numa_init(), which runs single-threaded at load time, and numa_bitmask_alloc just allocates a fresh bitmask. Same reasoning applies to the other calls in bindToSocket() — numa_available, numa_num_configured_nodes, numa_num_possible_cpus and numa_all_cpus_ptr all read constructor-initialized state. numa_node_to_cpus() is the only lazily-initialized path, which is why the lock is scoped to it alone.

he-yufeng:
The premise checks out independently: `tests/rdma_async_event_drain_test.cpp:46` already documents that the cache `numa_node_to_cpus()` allocates is intentionally retained until exit, so the lazy fill is real and the only question has been who serializes it. Scope also looks complete: these two `bindToSocket` copies are the only first-party `numa_node_to_cpus` call sites (`numa_run_on_node` / `numa_set_preferred` in store code operate on the calling thread's nodemask, not that shared cache), and the inline-accessor choice is correct, since a static local inside the internal-linkage `bindToSocket` would have been one mutex per translation unit.

One thing I would like clarified before this merges: the two new `numaNodeCpuCacheMutex()` accessors live in different namespaces, so a process that links both the legacy engine and TENT gets two distinct mutexes guarding the same process-global libnuma cache. If anything ever binds threads through both variants concurrently (migration-period binaries embedding both engines come to mind), the race survives between the two locks. Is that a shape we support? If yes, a single shared lock would be safer; an arguably simpler alternative is warming the cache once during engine init, single-threaded, before worker pools spawn, which removes the race at the source and keeps the hot path lock-free.

No objection to the approach itself. Also noting the test only guards under ASAN/LSAN, which the comment already states plainly.

he-yufeng:
Apologies for the repeat of the coexistence question in my second comment, it was written before I saw your answer. The runtime selection point settles it: one stack per process means the two mutexes never contend for the same cache fill, and the libnuma constructor-initialized state explains why the lock only needs to wrap `numa_node_to_cpus`. Nothing left open on my side.

alogfans (approved):
LGTM

PR #3296, opened 2026-08-05
[TE] EFA: derive WR/CQ pacing depths from the libfabric provider

whn09:
## Description

On EFA, the two counters that pace submission — `max_wr_depth_` (per-endpoint WR credit) and the CQ-entry limit used by `submitSlicesOnPeer()` — never handshake with the queues they are supposed to pace. They are derived from `GlobalConfig` alone, while the actual queue depths come from the libfabric provider.

Concretely, `buildSharedEndpoint()` never sets `hints_->tx_attr->size`, so libfabric picks the transmit queue depth itself. On EFA that depth is a *per-device* attribute — `efa_prov_info.c:244` returns `rounddown_power_of_two(device->efa_attr.max_sq_wr)`, measured as **4096 on p5.48xlarge and 2048 on p6-b300.48xlarge**. Meanwhile `max_wr_depth_` was initialized from the compiled-in `max_wr = 256` default. The counter and the queue were simply two unrelated numbers:

- When the counter is *below* the real depth (the default case: 256 vs 4096), submission stalls at 1/16 of the queue the provider actually handed us.
- When it is *above* (e.g. an operator sets `MC_MAX_WR=8192` on b300, where the queue is 2048), we hand out credit `fi_write()` must refuse with `FI_EAGAIN`.

Since no compiled-in default can be correct on every instance type, this makes the EFA transport adopt what the provider gave it:

- **WR depth** — read back `fi_info_->tx_attr->size` after `fi_getinfo()` and use it as `max_wr_depth_`. hints *cannot* request a deeper queue (a hint of 16384 makes `fi_getinfo` fail outright with `No data available`), so read-back-and-clamp is the only way to stay consistent with the provider.
- **CQ depth** — `fi_cq_open()` writes the size it chose back into `cq_attr.size`; read that instead of recomputing the provider's formula. The formula (`efa_domain.c:146`) is `MAX(rx_attr->size + tx_attr->size, efa_env.cq_size)`, and that last term is operator-settable via `FI_EFA_CQ_SIZE` (default 8192) — so a local copy of the formula would be wrong on any host where it has been set, and would silently drift if a future provider changes the rule.
- **`MC_MAX_WR`** — a new `max_wr_from_env` flag distinguishes "operator asked for this" from "nobody set it". When unset, the provider's depth wins. When set, the operator's value is honored but still clamped to what the hardware can accept, with a warning. The RDMA transport is untouched: it passes `max_wr` to `ibv_create_qp()`, so there the number already *is* the queue depth and verbs fails loudly on a mismatch (`rdma_endpoint.cpp:140`).

Operators who want a genuinely deeper queue should set **`FI_EFA_TX_SIZE`** (`efa_prov_info.c:664`), which moves the hardware queue; `MC_MAX_WR` can only move the counter. Verified: `FI_EFA_TX_SIZE=16384` → `tx=16384, rx=8192, CQ=24576`, and the transport now tracks that automatically. The EFA doc is updated accordingly.

**Note for reviewers:** the same defect exists in the CXI transport (`cxi_context.cpp:185` and `:731` derive the identical counters from `GlobalConfig`). It is deliberately left out of this PR — I have no Slingshot hardware to validate a fix on, and an untested change there would be worse than the status quo.

## Module

- [ ] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [x] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

Hardware: 2x **p5.48xlarge** (8x H100, 32 EFA NICs each, EFA installer 1.47.0, us-east-2), built with `-DUSE_EFA=ON -DUSE_CUDA=ON`. Patched and baseline trees differ by exactly this one commit (verified with `diff -rq`: only the 4 code files differ), and both binaries were built from identical toolchains on both hosts (matching sha256 across hosts).

**Test commands:**

```bash
# Unit tests for the new MC_MAX_WR provenance flag
./mooncake-transfer-engine/tests/config_test

# GPU-to-GPU regression check, 2 nodes, 32 NICs
#   target (P5-2):
transfer_engine_bench --mode=target --protocol=efa \
    --metadata_server=P2PHANDSHAKE --use_vram=true --gpu_id=0 \
    --buffer_size=4294967296
#   initiator (P5-1):
transfer_engine_bench --mode=initiator --protocol=efa \
    --metadata_server=P2PHANDSHAKE --use_vram=true --gpu_id=0 \
    --buffer_size=4294967296 \
    --segment_id=<target> --operation=write \
    --block_size=1048576 --threads=32 --batch_size=128 --duration=10
```

**Provider-depth adoption and clamping** (p5, provider tx queue = 4096):

| `MC_MAX_WR` | resulting `max_wr` | note |
| --- | --- | --- |
| unset | 4096 | adopts the provider depth (was 256) |
| 512 | 512 | operator value honored, no warning |
| 65535 | 4096 | clamped, one warning emitted |

`max_cqe` is reported as 12288 on all 32 NICs, matching what `fi_cq_open()` hands back for this device (`rx 8192 + tx 4096`).

**No throughput regression**, 1 MiB blocks / 32 threads / batch 128, GPU-to-GPU:

| variant | write (GB/s) | read (GB/s) |
| --- | --- | --- |
| baseline | 48.59 | — |
| this PR | 48.52, 48.60, 48.61, 48.63 | 48.59, 48.80, 48.80 |

Zero failed slices and zero `FI_EAGAIN` in every run. ~48.7 GB/s is the line rate for this link, so both variants saturate it; the deeper WR credit neither helps nor hurts once the wire is the bottleneck. A 16 KiB small-block sweep was also run, but a co-tenant workload started sharing the same NICs partway through, so I do not consider those numbers clean enough to report — they showed no regression, but with more run-to-run spread on the baseline than on this branch.

**Test results:**

- [x] Unit tests pass
- [ ] Integration tests pass (if applicable) — no suite covers EFA; the two-node benchmark above is the integration check

Five new cases in `config_test.cpp` cover `MC_MAX_WR` provenance: unset, valid override, rejected value, non-numeric, and out-of-range — asserting that `max_wr_from_env` is set only when the operator supplied a value the parser accepted.

## Checklist

- [x] I have run pre-commit on the files this PR touches and all hooks pass
- [ ] For changes >500 LOC: I have filed an RFC issue — n/a, ~193 added lines

`code_format.sh --check --changed-lines` exits 0 with clang-format 20.1.0 (the version the script pins). `pre-commit run --from-ref origin/main --to-ref HEAD` passes every applicable hook, including codespell over both touched docs. I ran it scoped to this branch's commits rather than `--all-files`, per AGENTS.md: broader hooks rewrite unrelated files and those edits do not belong in this PR.

Rebased onto current `main` to resolve conflicts with #3219, which landed after this branch was cut. Both conflicts were adjacent-insertion only, with no semantic overlap: `index.md` adds a second entry to the same env-var list, and `config_test.cpp` appends a second fixture at the end of the file. Both sides' content is retained.

## AI Assistance Disclosure

- [ ] No AI tools were used

Claude Code assisted with reading the libfabric/EFA provider sources to pin down where the queue depths actually come from, with drafting the change and its tests, and with running the two-node benchmarks. I have reviewed every changed line and can defend the change end-to-end.


mooncake-transfer-engine/src/transport/efa_transport/efa_context.cpp
```diff
@@ -222,6 +251,71 @@ int EfaContext::buildSharedEndpoint(size_t max_wr, size_t max_inline) {
...
         return ERR_CONTEXT;
     }
+
+    // wr_depth_ paces fi_write/fi_read against the provider's transmit queue,
+    // so its ceiling must be the queue the provider actually gave us --
+    // fi_info_->tx_attr->size -- and not a number chosen independently of the
+    // hardware.  Unlike verbs, where max_wr is passed to ibv_create_qp() and a
+    // too-large value fails loudly at QP creation, nothing here ever told
+    // libfabric how deep we wanted the queue: fi_endpoint() below takes its
+    // depth from fi_info_, so the counter and the queue were free to disagree.
```
alogfans: Brief describe the design idea as possible.

alogfans (approved):
Overall LGTM. The following comment is advised.

UNIDY2002:
Hi @whn09, #3309 resolves a typo failure. Would you have a look?

Issue #2039, opened 2026-05-05
[Bug]: Massive performance degradation when a replica fails in multi-node SGLang HiCache with Mooncake backend


riZZZhik:
Hello!
We are constantly facing issues with multi-node SGLang HiCache deployments using the Mooncake backend.
Different root causes lead to the same behavior: massive performance degradation across all replicas.

### Replica tries to allocate previously used metadata port

We start up to 14 replicas with 8 clients each.

If any replica tries to use a port that was allocated earlier but is now free, or a port that is currently allocated, the following happens:

1. Mooncake on the client that failed to allocate the requested port correctly selects another port.
2. However, all other clients start performing put / get operations against the unacquired port on the host that attempted to acquire it, constantly producing logs like:

```
E0428 09:59:43.129539 12445 transfer_metadata.cpp:1154] Failed to find location of localhost:13292
E0428 09:59:43.129846 13043 transfer_metadata_plugin.cpp:290] GET http://commodity-llm-qwen35-397b-a17b-fp8-mooncake-58vblr.api-tinkoff-ai.mlc.region.local:8080/metadata?key=mooncake%2Frpc_meta%2Flocalhost%3A13292 http=404 body: metadata not found
E0428 09:59:43.130234 11886 worker_pool.cpp:244] Worker: Cannot make connection for endpoint: localhost:13292@mlx5_0, mark it inactive
```

We mitigated this behavior by limiting the port range in https://github.com/kvcache-ai/Mooncake/pull/2008.
We also noticed that this issue can sometimes resolve itself after approximately 15 minutes.

### Short-term RDMA failure

In our cluster metrics, we observe a momentary increase in TX Discards up to 30k and Buffer Overrun up to 5 on one of the IB connections.

At the same time, all SGLang clients produce several errors like this with fixed peer_nic:
```
E0505 16:29:11.891577 12891 worker_pool.cpp:296] Worker: Process failed for slice (opcode: 0, source_addr: 0x7f204ae9c000, length: 65536, dest_addr: 0x7f6bad588000, local_nic: mlx5_9, peer_nic: localhost:14278@mlx5_5, dest_rkey: 2099456, retry_cnt: 0): transport retry counter exceeded

E0505 16:29:11.910105 12891 transfer_metadata_plugin.cpp:961] SocketHandShakePlugin: connect()10.178.54.150:16031: Connection refused [111]

[...]

PR #3316, opened 2026-08-06
[Store] Give LOCAL_DISK segments an unmount half, and let the offload RPC honour connect timeouts

Juhyun-Kim-Memphis:
## Description

`LOCAL_DISK` is the only segment kind with no unmount half. It leaves the master
exclusively by client expiry, so a store that is shutting down stays advertised
as the owner of every key it offloaded for up to one `client_ttl` after it stops
pinging, and a read that picks up that replica reaches a peer that can no longer
serve it.

The asymmetry is easy to see by listing what each segment kind exposes:

| Segment | Mount | Unmount | Graceful unmount | Callable from the client |
| -- | -- | -- | -- | -- |
| Memory | ✅ | `UnmountSegment` | `GracefulUnmountSegment` | ✅ |
| NVMe-oF (`NoF`) — also a disk | ✅ | `UnmountNoFSegment` | — | ✅ |
| `LOCAL_DISK` | ✅ every layer | master-internal function only | ❌ | ❌ |

`MountLocalDiskSegment` runs through `client_service` → `master_client.h` →
`rpc_service.h` → `master_service.h` → `segment.h`. `UnmountLocalDiskSegment`
exists in `segment.h` and at exactly one call site — the client-expiry branch of
`ClientMonitorFunc` — with no registered handler. Since `UnmountNoFSegment` is a
registered RPC and NVMe-oF is also a disk, this does not look like a property of
disks; it looks like a missing half.

`Client::~Client()` inherits the same gap: it stops the ping thread first, then
unmounts the memory segments tracked in `mounted_segments_`, and a `LOCAL_DISK`
segment is never put in that map. #2077 hit the same thing from the remount side
and worked around it there.

### How long a read blocks, and why it is not configurable

`ClientRequester::ClientRequester()` hardcodes `connect_retry_count = 3` and
`reconnect_wait_time = 1000 ms` and leaves the connect timeout at coro_rpc's
built-in 30 s. Unlike `MasterClient`, it reads no environment override, so
3 × 30 s + 1 s ≈ 91 s and nothing in the configuration can shorten it. A
sequential reader measured 91,002 ms twice, in two independent sessions, on a
two-node RDMA rig.

The caller's own deadline always fires first — a KV-cache read has a couple of
seconds, since a miss just falls back to prefill recompute — so the cost is not
latency. It is that the worker stays parked in the RPC: every concurrent read
issued into the same window parks another one, which turns a routine scale-in
into a capacity loss rather than a latency blip.

### The change

Two parts, covering two different paths. They are not alternatives.

**1. Give `LOCAL_DISK` the unmount half its mount half already has.**
`MasterService::UnmountLocalDiskSegment(client_id)` is exposed through the same
layers (`rpc_service` → `master_client` → `client_service`). `FileStorage` gains
`DrainLocalDiskSegment(grace_period_ms)`, reachable from Python as
`unmount_local_disk_segment(grace_period_seconds)` and over the store's REST API
as `POST /api/unmount_local_disk`, so a shutdown hook can call it. This closes
the window on the **planned** path — a scale-in, a rolling replacement.

Three things were less obvious than they look, and are why this is not a
three-line patch:

- **The grace period is load-bearing.** A memory replica is served by the NIC
  without the store process; a disk replica is read and pushed *by* that
  process. So the process has to outlive the deregistration long enough for the
  reads the master already handed out. Deregister-then-wait is also the
  race-free order: reads that already have the replica list complete against a
  live store, and reads that have not asked yet get a clean miss.
- **The heartbeat undoes a naive unmount.** `FileStorage::Heartbeat` re-mounts
  the segment and re-registers every object whenever `OffloadObjectHeartbeat`
  answers `SEGMENT_NOT_FOUND`. On the default 10 s interval, an unmount without
  a latch is reverted within ten seconds. The drain therefore sets `draining_`
  *before* the RPC, and the heartbeat returns early while it is set. The client
  TTL is renewed by `Ping` on a different thread, so the client stays alive for
  its memory segments.
- **The sweep has to be scoped by owner, not by liveness.** The replica cleanup
  classifies a `LOCAL_DISK` replica as stale when its owner is missing from the
  set it is handed, and a client only enters `ok_client_` by remounting. Passing
  the alive snapshot alone therefore drops the disk replicas of any peer that
  has not remounted yet — one store's deregistration becomes everyone's. The set
  passed here is the alive snapshot plus every current disk-segment owner, minus
  the caller. There is a regression test for exactly this
  (`UnmountLocalDiskSegmentDropsOnlyItsOwnReplicas`).

Object metadata is treated exactly as on expiry, including erasing a key whose
last replica was that disk. That is safe because `AddReplica` recreates a
missing object (`accessor.Create`), so a store that comes back re-adopts its own
files through `MountLocalDiskSegment` + `NotifyOffloadSuccess` — the recovery
#2077 added. Measured both ways below: 256/256 keys came back after a deliberate
drain as well as after a crash-expiry, and an earlier session saw 1152 keys
restored in 4 ms on the expiry path.

**2. Let the offload RPC honour the timeout variables the master client already
honours.** Both pools now go through one helper
(`detail::ApplyRpcTimeoutEnvOverrides`), so `MC_RPC_CONNECT_TIMEOUT_MS` and
`MC_RPC_TIMEOUT_MS` apply to the store→store path too and the two cannot drift.
**Defaults are unchanged when the variables are unset.** This covers the
**unplanned** path — a crash, an OOM kill, a lost node — where nobody gets to
unmount anything. `ClientRequester::invoke_rpc` also maps
`coro_rpc::errc::timed_out` to `ErrorCode::RPC_TIMEOUT`, as `master_client.cpp`
already does; without it a fired timeout was indistinguishable from any other
RPC failure.

Related: #2953 (the client liveness state machine RFC — its goal of separating
segment lifecycle from liveness is the same direction as part 1, and this is the
narrow version of it), #2039 (one dead peer propagating to all replicas), #2077.

## Module

- [x] Docs

## Type of Change

- [x] Bug fix
- [x] New feature

## How Has This Been Tested?

Unit tests, plus a two-node RDMA rig running the built artifacts.

**Test commands:**
```bash
cmake .. -G Ninja -DCMAKE_BUILD_TYPE=Release -DWITH_STORE_RUST=OFF \
  -DWITH_P2P_STORE=OFF -DWITH_EP=OFF -DBUILD_UNIT_TESTS=ON -DUSE_ETCD=OFF
ninja master_service_ssd_test rpc_timeout_test segment_test offload_on_evict_test
./mooncake-store/tests/master_service_ssd_test
./mooncake-store/tests/rpc_timeout_test
./mooncake-store/tests/segment_test
./mooncake-store/tests/offload_on_evict_test
```

**Test results:** 49 tests across the four suites, all passing — 23
`master_service_ssd_test` (5 new), 3 `rpc_timeout_test` (1 new), 14
`segment_test`, 9 `offload_on_evict_test`.

New tests:

| Test | What it pins |
| -- | -- |
| `UnmountLocalDiskSegmentRequiresOffloadEnabled` | `UNABLE_OFFLOAD` when offload is off, mirroring mount |
| `UnmountLocalDiskSegmentIsIdempotent` | repeats and unknown clients succeed |
| `UnmountLocalDiskSegmentStopsOffloadWork` | the heartbeat then answers `SEGMENT_NOT_FOUND` |
| `UnmountLocalDiskSegmentDropsOnlyItsOwnReplicas` | a peer that has not remounted keeps its disk replicas |
| `UnmountLocalDiskSegmentKeepsReAdoptionWorking` | a disk-only key is erased, and re-registration recreates it |
| `TimeoutEnvOverridesAreOptIn` | unset leaves the built-in default; set is applied; both pools share it |

- [x] Unit tests pass

### Rig measurement

Two nodes, RDMA, the store owning the disk and the reader on the other node,
`--client_ttl=10`, 256 disk-resident keys (confirmed by the offload-RPC counter
rising), 8 concurrent readers. Each arm stamps its own wall clock immediately
before the action; `.metadata.deletionTimestamp` is SIGTERM plus
`terminationGracePeriodSeconds`, not the request time, and anchoring on it is how
an earlier session mistook an ordinary shutdown for a spontaneous departure.

| Arm | What | Result |
| -- | -- | -- |
| Planned scale-in, through the drain | `POST /api/unmount_local_disk` in preStop, grace 20 s | **0 reads blocked**, worst read after the request **40.3 ms / 41.9 ms** over two runs |
| Crash, no override | store process `kill -9` | **every one of the 8 workers parked on one read** — 4 × 33.2 s, 4 × 62.0 s |
| Crash, `MC_RPC_CONNECT_TIMEOUT_MS=1000` | same kill | worst read **4,196 ms** — 3 retries × 1 s plus the waits |
| Correctness, all arms | payload derived from the key, checked byte for byte | **654,123 reads, 0 wrong answers, 0 client exceptions** |
| Re-adoption | store returns to the disk it left | **256/256 keys back**, after both the crash-expiry path and a deliberate drain |

The mechanism is visible in the master's own gauges. On the planned scale-in the
two diverge: `file_cache_nums_` goes 1136 → 0 at **T0+1.7 s**, when the drain's
RPC returns, while `master_active_clients` only drops at **T0+31.6 s** when the
TTL expires. So the disk tier is deregistered while the store can still serve,
and nothing is handed a dead owner for the following 30 seconds. On the crash the
two fall together at T0+11.1 s, TTL expiry being the only path — that is the
window part 2 bounds.

One honest gap: the uncapped crash blocked for 33.2 s and 62.0 s rather than the
full 91 s, i.e. one and two connect attempts rather than three. A sequential
reader reproduced 91,002 ms to the millisecond in an earlier session, so 91 s
stands as the ceiling; the lower figures here look like eight workers contending
on one client pool, which I did not chase further. Either way it is tens of
seconds, and the override brings it to 4.2 s.

A note for anyone reproducing this: `kubectl delete --force --grace-period=0`
does **not** give you a crash. The kubelet still runs preStop — the disk replicas
came out 1.03 s after the request while the client stayed alive for another 10 s —
so it measures a truncated drain. Kill the process instead.

## Checklist

On the last two: the diff is ~520 lines including tests and docs (~230 lines of
non-test source). #2953 is the open RFC this follows, and I would rather fold
this into that discussion than open a competing one — happy to file a separate
RFC if you would prefer. `pre-commit` was not run end to end because the build
host had no `clang-format`; the C++ follows the existing 4-space / 80-column
style of the files it touches, and I will fix up anything the hooks flag.

## AI Assistance Disclosure

Written with Claude Code. It traced the segment-lifecycle and replica-cleanup
paths, wrote the patch and the tests, and drove the rig. The two design
corrections that mattered — the heartbeat re-mount race, and scoping the replica
sweep by owner rather than by liveness — came out of that reading and from a
test failure, and both are covered by tests above. I have reviewed every line
and am responsible for it.

### DIFF
--- a/mooncake-store/src/file_storage.cpp
+++ b/mooncake-store/src/file_storage.cpp
@@ -790,6 +790,16 @@ tl::expected<void, ErrorCode> FileStorage::Heartbeat() {
         return tl::make_unexpected(ErrorCode::INVALID_PARAMS);
     }
 
+    // A drain deregistered the disk tier on purpose. Skipping the whole tick
+    // matters for the SEGMENT_NOT_FOUND branch below, which would re-mount the
+    // segment and re-register every object; there is also no point taking new
+    // offload work for a store that is leaving. The client TTL is renewed by
+    // Ping on the storage heartbeat thread, not here, so the client stays
+    // alive for its memory segments.
+    if (draining_.load()) {
+        return {};
+    }
+
     // Join previous rescan if completed
     if (rescan_future_.valid() && rescan_future_.wait_for(std::chrono::seconds(
                                       0)) == std::future_status::ready) {
@@ -929,6 +939,33 @@ void FileStorage::RemoveAll() {
         storage_backend_->RemoveAll();
     }
     LOG(INFO) << "FileStorage::RemoveAll: cleared storage backend";
+}
+
+tl::expected<void, ErrorCode> FileStorage::DrainLocalDiskSegment(
+    uint64_t grace_period_ms) {
+    if (client_ == nullptr) {
+        LOG(ERROR) << "client is nullptr";
+        return tl::make_unexpected(ErrorCode::INVALID_PARAMS);
+    }
+
+    // Latch before the RPC, so a heartbeat tick that overlaps the
+    // deregistration cannot re-mount the segment behind it.
+    draining_.store(true);
+
+    auto result = client_->UnmountLocalDiskSegment();
+    if (!result) {
+        LOG(ERROR) << "action=drain_local_disk_segment, error="
+                   << result.error();
+        draining_.store(false);
+        return result;
+    }
+
+    LOG(INFO) << "action=drain_local_disk_segment, grace_period_ms="
+              << grace_period_ms;
+    if (grace_period_ms > 0) {
+        std::this_thread::sleep_for(std::chrono::milliseconds(grace_period_ms));
+    }
+    return {};
 }
 
 tl::expected<void, ErrorCode> FileStorage::ProcessPromotionTasks() {
--- a/mooncake-store/src/file_storage.cpp
+++ b/mooncake-store/src/file_storage.cpp
@@ -879,6 +879,16 @@ tl::expected<void, ErrorCode> FileStorage::Heartbeat() {
         return tl::make_unexpected(ErrorCode::INVALID_PARAMS);
     }
 
+    // A drain deregistered the disk tier on purpose. Skipping the whole tick
+    // matters for the SEGMENT_NOT_FOUND branch below, which would re-mount the
+    // segment and re-register every object; there is also no point taking new
+    // offload work for a store that is leaving. The client TTL is renewed by
+    // Ping on the storage heartbeat thread, not here, so the client stays
+    // alive for its memory segments.
+    if (draining_.load()) {
+        return {};
+    }
+
     // Join previous rescan if completed
     if (rescan_future_.valid() && rescan_future_.wait_for(std::chrono::seconds(
                                       0)) == std::future_status::ready) {
@@ -905,6 +915,14 @@ tl::expected<void, ErrorCode> FileStorage::Heartbeat() {
     // === STEP 1: Send heartbeat and get offloading decisions ===
     {
         MutexLocker locker(&offloading_mutex_);
+        // Re-checked under the lock: the entry check above is only a fast
+        // path. A drain (which latches and deregisters under this same
+        // mutex) may have completed while this tick was parked on it; going
+        // on would hit SEGMENT_NOT_FOUND below and re-mount the segment the
+        // drain just deregistered.
+        if (draining_.load()) {
+            return {};
+        }
         auto fetch_offload_tasks = [&]() -> tl::expected<void, ErrorCode> {
             return client_->OffloadObjectHeartbeat(enable_offloading_,
                                                    offloading_objects);
@@ -1020,6 +1038,45 @@ void FileStorage::RemoveAll() {
         storage_backend_->RemoveAll();
     }
     LOG(INFO) << "FileStorage::RemoveAll: cleared storage backend";
+}
+
+tl::expected<void, ErrorCode> FileStorage::DrainLocalDiskSegment(
+    uint64_t grace_period_ms) {
+    if (client_ == nullptr) {
+        LOG(ERROR) << "client is nullptr";
+        return tl::make_unexpected(ErrorCode::INVALID_PARAMS);
+    }
+
+    // Latch and deregister under offloading_mutex_, the lock Heartbeat holds
+    // across its master RPCs and the SEGMENT_NOT_FOUND re-mount. The latch
+    // alone is not enough: a heartbeat that read draining_ == false before
+    // this store() and was then scheduled out would, on resume, see
+    // SEGMENT_NOT_FOUND from the deregistration below and re-mount the
+    // segment. Under the mutex, such a tick is either still parked on the
+    // lock (it re-checks draining_ once it gets it and skips) or it already
+    // finished its RPCs before we latched (anything it mounted or registered
+    // is deregistered again right here, and the master refuses registrations
+    // that arrive after this call returns -- see
+    // MasterService::NotifyOffloadSuccess).
+    {
+        MutexLocker locker(&offloading_mutex_);
+        draining_.store(true);
+
+        auto result = client_->UnmountLocalDiskSegment();
+        if (!result) {
+            LOG(ERROR) << "action=drain_local_disk_segment, error="
+                       << result.error();
+            draining_.store(false);
+            return result;
+        }
+    }
+
+    LOG(INFO) << "action=drain_local_disk_segment, grace_period_ms="
+              << grace_period_ms;
+    if (grace_period_ms > 0) {
+        std::this_thread::sleep_for(std::chrono::milliseconds(grace_period_ms));
+    }
+    return {};
 }
 
 tl::expected<void, ErrorCode> FileStorage::ProcessPromotionTasks() {
@@ -1379,6 +1436,17 @@ tl::expected<void, ErrorCode> FileStorage::ReRegisterOffloadedObjects() {
             [this, &total_keys, &total_batches, &total_failures](
                 const std::vector<std::string>& keys,
                 std::vector<StorageObjectMetadata>& metadatas) {
+                // A rescan can be mid-flight when a drain deregisters the
+                // disk tier; every batch it would register from here on
+                // re-advertises the departing store. The master refuses such
+                // registrations anyway (the segment entry is gone), so this
+                // stops the scan on the first batch instead of on a refused
+                // RPC.
+                if (draining_.load()) {
+                    LOG(INFO) << "ReRegisterOffloadedObjects: aborting scan, "
+                              << "the disk tier is draining";
+                    return ErrorCode::SEGMENT_NOT_FOUND;
+                }
                 total_batches++;
                 total_keys += keys.size();
                 for (auto& metadata : metadatas) {
--- a/mooncake-store/src/master_service.cpp
+++ b/mooncake-store/src/master_service.cpp
@@ -5985,6 +5985,60 @@ auto MasterService::MountLocalDiskSegment(const UUID& client_id,
     return {};
 }
 
+auto MasterService::UnmountLocalDiskSegment(const UUID& client_id)
+    -> tl::expected<void, ErrorCode> {
+    if (!enable_offload_) {
+        LOG(ERROR) << "The offload functionality is not enabled";
+        return tl::make_unexpected(ErrorCode::UNABLE_OFFLOAD);
+    }
+
+    {
+        ScopedLocalDiskSegmentAccess ssd_access =
+            segment_manager_.getLocalDiskSegmentAccess();
+        const auto& client_segments = ssd_access.getClientLocalDiskSegment();
+        if (client_segments.find(client_id) == client_segments.end()) {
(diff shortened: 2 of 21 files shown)


mooncake-store/src/file_storage.cpp
```diff
@@ -931,6 +941,33 @@ void FileStorage::RemoveAll() {
...
+tl::expected<void, ErrorCode> FileStorage::DrainLocalDiskSegment(
+    uint64_t grace_period_ms) {
+    if (client_ == nullptr) {
+        LOG(ERROR) << "client is nullptr";
+        return tl::make_unexpected(ErrorCode::INVALID_PARAMS);
+    }
+
+    // Latch before the RPC, so a heartbeat tick that overlaps the
+    // deregistration cannot re-mount the segment behind it.
+    draining_.store(true);
```
fcczzz: Setting this atomic before the RPC only prevents heartbeats that observe the flag after this point. A heartbeat can read `draining_ == false` at the start of `Heartbeat()`, pause, and then resume after `UnmountLocalDiskSegment()` has removed the segment. It will receive `SEGMENT_NOT_FOUND`, remount the segment, and start re-registering its metadata. The grace period can then end with the master advertising the exiting store again, recreating the stale-owner window this change is intended to close.

mooncake-store/src/master_service.cpp
```diff
@@ -5985,6 +5985,60 @@ auto MasterService::MountLocalDiskSegment(const UUID& client_id,
...
+    auto staying_clients = getAliveClientsSnapshot();
+    {
+        ScopedLocalDiskSegmentAccess ssd_access =
+            segment_manager_.getLocalDiskSegmentAccess();
+        for (const auto& owner : ssd_access.getClientLocalDiskSegment()) {
+            staying_clients.insert(owner.first);
+        }
+    }
+    staying_clients.erase(client_id);
+    ClearInvalidHandles(staying_clients);
```
fcczzz: `staying_clients` is only a point-in-time snapshot. Another store can mount its `LOCAL_DISK` segment after the owners are collected above and register replicas through `NotifyOffloadSuccess()` before this sweep reaches their metadata shard. That owner is absent from the old snapshot, so `ClearInvalidHandles()` classifies its completed disk replicas as stale and may erase the key when that was its last replica. If the store's rescan already returned successfully, it will not retry the lost registration.

Since this RPC intends to remove replicas owned by exactly `client_id`, please use an owner-targeted cleanup, or otherwise establish a consistency boundary that excludes concurrent mount/register operations across owner collection and the sweep.

fcczzz (changes requested):
Thanks for the thorough investigation and validation. 

I found two blocking concurrency holes in the planned-drain path: an already-running heartbeat can remount after deregistration, and the master can sweep replicas registered by another store after the non-atomic owner snapshot was taken. Both paths violate the drain's core ownership guarantee, so I think they need to be addressed, with deterministic interleaving tests, before this is safe to merge.

Juhyun-Kim-Memphis:
@fcczzz 
Both are real — thanks for catching them. Fixed in cf28cbd3.

The first was a check-then-act gap: the drain now performs its latch and deregistration under the same lock the heartbeat holds across the master call that reports `SEGMENT_NOT_FOUND` and the re-mount that follows it, and the heartbeat re-checks after acquiring it. A tick can no longer resume past the deregistration and undo it; an in-flight rescan stops the same way.

For the second I switched the sweep to target only the deregistering owner, as you suggested — that removes the snapshot and the misclassification along with it. Registrations racing the deregistration itself are handled by making the segment removal and a registration's admission check mutually exclusive, so a registration either lands before the removal and gets swept, or is refused.

Added three tests that pin the interleavings deterministically instead of relying on the scheduler.

fcczzz (commented):
Thanks, both blocking races are addressed. I don’t see another code blocker now, but the PR currently conflicts with main; please rebase/resolve the conflicts and ping me for a final check before approval.

ykwd:
Hi, the check code format CI failed

Juhyun-Kim-Memphis:
@fcczzz Rebased onto `e949ce24` and the conflicts are resolved — ready for your final check.

#3427 landed in between and moved LOCAL_DISK client management into `LocalSsdManager`, so this is not a mechanical rebase: `MasterService::UnmountLocalDiskSegment` now drops the registration through `local_ssd_manager_.UnregisterClient()` under the exclusive `snapshot_mutex_`, and `NotifyOffloadSuccess`'s admission check reads the same registry instead of the old segment map. The two races you found are unchanged in substance — the drain still latches offloading off and deregisters under `offloading_mutex_`, and the master-side sweep is still owner-targeted (`ClearLocalDiskHandlesOwnedBy`) rather than a liveness-complement.

`Check code format` is green again (thanks @ykwd for flagging it).

Beyond CI, I compiled this exact tree and ran the four suites that cover these paths — `file_storage_test` (26), `master_service_ssd_test` (25), `rpc_timeout_test` (3), `promotion_on_hit_test` (52) — 106 cases, all passing, on ubuntu 24.04 with `USE_ETCD=OFF USE_CUDA=OFF`.

ykwd (approved):

mooncake-wheel/mooncake/mooncake_store_service.py
```diff
@@ -614,6 +620,45 @@ async def handle_unmount(self, request):
...
+        Meant for a preStop hook. The master stops naming this store as the
+        owner of its offloaded keys, so a reader gets a clean miss instead of a
+        peer that is about to disappear; the call then holds for
+        grace_period_seconds so offload reads already in flight finish here.
+        Runs off the event loop because that wait is seconds long.
+        """
+        try:
+            try:
+                data = await request.json()
+            except Exception:
```
zxpdemonio: Blocking: please do not treat every JSON parsing failure as an empty request. A malformed preStop payload (for example, an intended `{"grace_period_seconds": 30}` with a syntax error) reaches the real unmount with the default grace period of 0 and returns 200, so the process may exit without the requested grace period. Validate the body before invoking the store: return 400 for malformed JSON, and require `grace_period_seconds` to be a non-bool, non-negative integer within a checked upper bound. Please also add handler tests for malformed JSON, invalid types/negative values, the default and positive-value paths, and verify invalid input never calls `unmount_local_disk_segment`.

Juhyun-Kim-Memphis: @zxpdemonio  Good catch — fixed in 9fb66cad. Malformed JSON and a non-object body now get 400 before touching the store, and `grace_period_seconds` must be a non-bool, non-negative integer no greater than 3600s. Added handler tests for each rejection path plus the default/positive-value paths, asserting `unmount_local_disk_segment` is never called on invalid input.

zxpdemonio (approved):
LGTM.

Juhyun-Kim-Memphis:
@fcczzz  Rebased onto current main (9fb66cad), both races from your review still hold under the tests. Could you take a final look?

fcczzz (approved):
LGTM.

PR #3319, opened 2026-08-06
[Store] Support pinned SSD-to-GPU restore

zxpdemonio:
## Description

This PR adds a capacity-bounded pinned-host restore arena for same-process SSD-to-GPU reads. It is part of the local tensor offload/restore path discussed in #3229.

The optimized path is selected only when all of the following are true:

- the current real client owns the `LOCAL_DISK` replica;
- the destination slices are CUDA device memory;
- local memcpy is enabled and the full process endpoint matches;
- io_uring is disabled; and
- `MC_STORE_PINNED_RESTORE_ARENA_SIZE_BYTES` is greater than zero.

The data path changes from:

```text
SSD -> pageable FileStorage ClientBuffer -> local H2D copy -> GPU
```

to:

```text
SSD -> capacity-bounded pinned restore arena -> local H2D copy -> GPU
```

The implementation:

- preallocates pinned host memory at FileStorage initialization instead of pinning memory per request;
- reuses `ClientBufferAllocator`, `PinnedBufferPool`, and the existing local memcpy worker;
- keeps local restore allocations alive with an RAII owner through H2D completion without publishing them in the remote batch/GC map;
- preserves the original remote Transfer Engine and lease paths;
- supports tensor ranged reads by loading the complete object required by the existing storage-backend contract, then applying the source offset before H2D, without another FileStorage staging allocation;
- falls back to the normal FileStorage arena when the pinned restore arena is exhausted or pinned allocation is unavailable; and
- retains the existing four-argument `Client::BatchGetOffloadObject` symbol while adding an explicit internal source-access mode.

The default arena size is zero, so existing deployments are unchanged. CPU destinations, remote owners, and io_uring continue to use their previous paths. The file-per-key backend may still use its own temporary pageable buffer internally; removing that backend-specific copy and increasing local H2D concurrency are outside this PR.

Related RFC: #3229

## Module

- [ ] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [x] Documentation update
- [x] Performance improvement
- [ ] Other

## How Has This Been Tested?

The test build used `USE_CUDA=OFF`, `USE_HTTP=ON`, and `USE_ETCD=OFF`. CUDA Toolkit was auto-detected for the focused pinned-host-to-GPU tests.

**Test commands:**

```bash
cmake --build build-ssd-pinned-tests --target file_storage_test transfer_task_test pybind_client_test -j4
./build-ssd-pinned-tests/mooncake-store/tests/file_storage_test --gtest_filter=FileStorageTest.BatchGetUsesPinnedArenaAndFallsBackWhenFull --gtest_color=no
./build-ssd-pinned-tests/mooncake-store/tests/transfer_task_test --gtest_filter=TransferTaskTest.BatchGetOffloadObjectHonorsLocalMemcpySetting:TransferTaskTest.BatchGetOffloadObjectCopiesPinnedHostToGpu --gtest_color=no
./build-ssd-pinned-tests/mooncake-store/tests/pybind_client_test --gtest_filter=RealClientTest.PinnedSsdRestoreReadsNonTailRangeIntoGpu:RealClientTest.DoubleTearDownIsIdempotent:RealClientTest.EmptyBatchOperations:RealClientTest.BasicPutGetOperations --gtest_color=no
pre-commit run --files mooncake-store/src/real_client.cpp mooncake-store/tests/CMakeLists.txt mooncake-store/tests/pybind_client_test.cpp
```

**Test results:**

- [x] Unit tests pass
- [x] Integration tests pass

- The CUDA end-to-end regression offloads a 64 KiB object, removes its DRAM replica, restores a non-tail 8 KiB + 16 KiB range into a registered GPU buffer, and verifies the returned length, offload-read count, and data. Before the fix it reproduced `expected: 65536, got: 24576` / `INVALID_PARAMS`; after the fix it passes.
- Focused `FileStorageTest`: 1/1 passed.
- Focused `TransferTaskTest`: 2/2 passed.
- Focused `RealClientTest`: 4/4 passed.
- Repository code format and all applicable pre-commit hooks passed.

### Same-node end-to-end SSD-to-GPU performance

The manual A/B test used the bucket backend with a 1 GiB bucket-size limit, `MC_STORE_MEMCPY=1`, io_uring disabled, warm page cache, one warmup, and five timed restores. The pinned run used a 768 MiB pinned restore arena. Before every measurement the object had only a `LOCAL_DISK` replica. Every restore verified the offload-read count and sampled GPU data.

The numbers below are **end-to-end restore latency from SSD (with a warm page cache) to GPU**, not standalone pinned-host-to-GPU copy bandwidth. The baseline is the current working two-stage path with pinned restore arena disabled: SSD to a registered pageable CPU destination, followed by an explicit H2D copy. The PR path reads SSD data into the preallocated pinned restore arena and then performs a local H2D copy into the registered GPU destination.

| Size | Pageable SSD-to-GPU median | Pageable end-to-end throughput | Pinned SSD-to-GPU median | Pinned end-to-end throughput | Speedup |
|---:|---:|---:|---:|---:|---:|
| 64 MiB | 40.70 ms | 1.65 GB/s | 13.69 ms | 4.90 GB/s | 2.97x |
| 256 MiB | 159.16 ms | 1.69 GB/s | 52.21 ms | 5.14 GB/s | 3.05x |
| 512 MiB | 359.26 ms | 1.49 GB/s | 103.99 ms | 5.16 GB/s | 3.45x |

For reference, the standalone copy stage from pinned host memory to the same NVIDIA A10 GPU reaches the expected PCIe bandwidth:

| Size | Pinned host-to-GPU median | H2D throughput |
|---:|---:|---:|
| 64 MiB | 2.80 ms | 23.98 GB/s |
| 256 MiB | 11.13 ms | 24.12 GB/s |
| 512 MiB | 22.23 ms | 24.15 GB/s |

Stage profiling shows that the end-to-end result is limited by the current single `preadv`/`FileStorage::BatchLoad` stage (about 5.3 GB/s from the warm page cache into either pageable or pinned host memory), rather than by pinned H2D. Parallelizing and pipelining large file reads is a separate storage-path optimization.

## Checklist

- [ ] I have run `pre-commit run --all-files` and all hooks pass (the repository-wide `origin/main` baseline has unrelated existing failures; all hooks applicable to this PR update pass)
- [x] For changes >500 LOC: related RFC #3229 is filed

## AI Assistance Disclosure

- [ ] No AI tools were used

OpenAI Codex assisted with implementation, code simplification, and multi-axis review. The changes were manually reviewed and tested.


mooncake-store/src/real_client.cpp
```diff
@@ -3431,12 +3431,28 @@ tl::expected<int64_t, ErrorCode> RealClient::execute_ranged_read(
     };
 
     if (replica.is_local_disk_replica()) {
+        const auto &endpoint =
+            replica.get_local_disk_descriptor().transport_endpoint;
+        void *dst = static_cast<char *>(buffer) + dst_offset;
+        std::unordered_map<std::string, std::vector<Slice>> objects{
+            {key, {{dst, size}}}};
+        if (can_use_pinned_restore(endpoint, objects)) {
+            const uint64_t restore_size = src_offset + size;
```
fcczzz: `restore_size` is passed to `FileStorage` as the destination slice size, but the current `StorageBackend::BatchLoad` implementations require it to equal the full object size. A non-tail range where `src_offset + size < object_size` will therefore fail with `INVALID_PARAMS`.

Could we pass `total_size` here or make the backend explicitly range-aware, and add an end-to-end non-tail range test?

zxpdemonio: Good catch. The existing storage backends require the destination slice size to match the full object, so the pinned ranged-read path now passes total_size to FileStorage and applies source_offset only after the full object is loaded. I also added a CUDA end-to-end non-tail range regression (8 KiB + 16 KiB < 64 KiB) that waits for a LOCAL_DISK replica, removes the DRAM replica, restores into a registered GPU buffer, and verifies the result and data. It reproduced the 65536-vs-24576 INVALID_PARAMS failure before the fix and passes afterward.

LujhCoconut (approved):
LGTM.

stmatengss (approved):

PR #3324, opened 2026-08-07
[TE] Add RDMA ctrl frame codec and sender credit ledger

zhtshr:
## Description

First slice of the RDMA two-sided control-plane work (split from the previous monolithic PR for reviewability).

Roadmap: [#3377](https://github.com/kvcache-ai/Mooncake/issues/3377).

Per review feedback, this work lands as a **separate transport** (`rdma_twosided`) rather than folding into classic `rdma`. This PR only adds the shared control-plane primitives under that tree:

- Typed `CtrlFrame` encode/decode (`transport/rdma_twosided/ctrl_frame.{h,cpp}`)
- Sender-side credit ledger (`transport/rdma_twosided/sender_credit.{h,cpp}`)
- Unit tests: `ctrl_frame_test`, `sender_credit_test`
- CMake object library `rdma_twosided` wired into `transport` / `transfer_engine`

Review follow-ups in this update:
- Epoch fencing on `tryReserve` / `rollbackReservation` / `available`
- Stricter ctrl-frame decode (version/type/flags/trailing; transactional payloads)

No CtrlChannel / MsgChannel / transfer path wiring yet, and classic `rdma` is unchanged functionally. Follow-up PRs will stack notify QP, two-sided data path, multi-rail, PD install selection (`rdma` **or** `rdma_twosided`), docs, and transport-level benchmarks (see RFC).

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
ninja ctrl_frame_test sender_credit_test
./mooncake-transfer-engine/tests/ctrl_frame_test
./mooncake-transfer-engine/tests/sender_credit_test
```

**Test results:**
- [x] Unit tests pass (ctrl_frame_test / sender_credit_test; no RDMA device required)

## Checklist

- [x] I have run `pre-commit` on touched files and hooks pass

RFC roadmap: https://github.com/kvcache-ai/Mooncake/issues/3377

## AI Assistance Disclosure

- [ ] No AI tools were used

Cursor agent assisted implementing and splitting the two-sided control-plane work. Human submitter reviewed the PR1 scope (ctrl frame + credit ledger under `rdma_twosided`).


alogfans:
The core motivation for adding a two-sided RDMA path is still not clear enough to me. This PR introduces a fairly large amount of new machinery—control-plane credit, acknowledgements, bounce buffers, dedicated metadata, and accompanying documentation—but I do not yet see a sufficiently strong explanation of what concrete problem this solves that the existing one-sided path or current transport model cannot. I think the PR should first clarify the real workload or business scenario behind this change, what limitation or bottleneck exists in the current design, and whether the expected benefit is compatibility, flow control, stability, tail latency, throughput, or resource efficiency. Without a clear production-driven use case or benchmark evidence, this adds significant long-term maintenance and testing cost for the project.

My second concern is that the implementation feels too intrusive to the current Transfer Engine architecture. From the description, this is not an isolated feature but a new communication mode whose control plane, credit model, and buffer-registration semantics are being woven into the existing TE path. That increases architectural complexity and makes the mainline logic harder to reason about, test, and evolve. If two-sided support is truly needed long term, I would strongly recommend maintaining it as a dedicated transport instead of folding so much specialized logic into the current transport flow. That would keep the abstraction boundaries cleaner, reduce coupling between one-sided and two-sided behavior, and make rollout, debugging, and future iteration much safer.

zhtshr:
> The core motivation for adding a two-sided RDMA path is still not clear enough to me. This PR introduces a fairly large amount of new machinery—control-plane credit, acknowledgements, bounce buffers, dedicated metadata, and accompanying documentation—but I do not yet see a sufficiently strong explanation of what concrete problem this solves that the existing one-sided path or current transport model cannot. I think the PR should first clarify the real workload or business scenario behind this change, what limitation or bottleneck exists in the current design, and whether the expected benefit is compatibility, flow control, stability, tail latency, throughput, or resource efficiency. Without a clear production-driven use case or benchmark evidence, this adds significant long-term maintenance and testing cost for the project.
> 
> My second concern is that the implementation feels too intrusive to the current Transfer Engine architecture. From the description, this is not an isolated feature but a new communication mode whose control plane, credit model, and buffer-registration semantics are being woven into the existing TE path. That increases architectural complexity and makes the mainline logic harder to reason about, test, and evolve. If two-sided support is truly needed long term, I would strongly recommend maintaining it as a dedicated transport instead of folding so much specialized logic into the current transport flow. That would keep the abstraction boundaries cleaner, reduce coupling between one-sided and two-sided behavior, and make rollout, debugging, and future iteration much safer.

Thanks for the review — agreeing on both the motivation clarity and the architecture concern.

## Motivation

We are not trying to replace the one-sided zero-copy path for peak bandwidth. The two-sided path targets a different trade-off for workloads such as PD:

1. **Simpler application buffer lifetime** — avoid repeated `registerLocalMemory` / `unregisterLocalMemory` around transfers and the constraint that buffers must stay registered until there is no inflight work.
2. **More flexible TE-owned networking buffers** — bytes move via TE bounce/RQ resources whose capacity can expand/shrink under load (credit + WAITING), instead of tying every business allocation to an MR registration.
3. **Decouple remote memory ownership** — SEND/RECV + local placement so the sender need not hold a remote rkey to the peer’s application buffers; each side manages its own placement memory.
4. **Acceptable latency for the target use case** — two-sided adds bounce/copy vs one-sided zero-copy, but for our PD-oriented scenarios the latency is acceptable, so we prefer lifetime/ownership simplicity over max GB/s.

One-sided registered WRITE/READ remains the right choice when stable MRs and peak bandwidth matter. `rdma_twosided` is an **opt-in alternate transport** (PD installs either `rdma` or `rdma_twosided`, not both).

## Architecture / intrusion

We agree folding this into the existing classic `rdma` transport is too invasive. Going forward we will implement two-sided as a **separate transport** (working name `rdma_twosided`):

- Keep the current `rdma` one-sided path largely untouched.
- Put ctrl/credit/msg/bounce under the new transport tree (copy shared RDMA plumbing where needed for clarity, rather than deeply sharing hot-path code).
- Reuse the existing TE handshake/metadata protocol (extend capability fields as needed), not a new discovery stack.

This PR (#3324) stays limited to the ctrl-frame codec and sender credit ledger (+ unit tests) as building blocks for that transport. We’ll relocate them under the `rdma_twosided` directory as we land the transport skeleton so they are not framed as shared classic-`rdma` core.

Happy to adjust naming/details if you have a preferred transport name or layout.

staryxchen:
> > The core motivation for adding a two-sided RDMA path is still not clear enough to me. This PR introduces a fairly large amount of new machinery—control-plane credit, acknowledgements, bounce buffers, dedicated metadata, and accompanying documentation—but I do not yet see a sufficiently strong explanation of what concrete problem this solves that the existing one-sided path or current transport model cannot. I think the PR should first clarify the real workload or business scenario behind this change, what limitation or bottleneck exists in the current design, and whether the expected benefit is compatibility, flow control, stability, tail latency, throughput, or resource efficiency. Without a clear production-driven use case or benchmark evidence, this adds significant long-term maintenance and testing cost for the project.
> > My second concern is that the implementation feels too intrusive to the current Transfer Engine architecture. From the description, this is not an isolated feature but a new communication mode whose control plane, credit model, and buffer-registration semantics are being woven into the existing TE path. That increases architectural complexity and makes the mainline logic harder to reason about, test, and evolve. If two-sided support is truly needed long term, I would strongly recommend maintaining it as a dedicated transport instead of folding so much specialized logic into the current transport flow. That would keep the abstraction boundaries cleaner, reduce coupling between one-sided and two-sided behavior, and make rollout, debugging, and future iteration much safer.
> 
> Thanks for the review — agreeing on both the motivation clarity and the architecture concern.
> 
> ## Motivation
> We are not trying to replace the one-sided zero-copy path for peak bandwidth. The two-sided path targets a different trade-off for workloads such as PD:
> 
> 1. **Simpler application buffer lifetime** — avoid repeated `registerLocalMemory` / `unregisterLocalMemory` around transfers and the constraint that buffers must stay registered until there is no inflight work.
> 2. **More flexible TE-owned networking buffers** — bytes move via TE bounce/RQ resources whose capacity can expand/shrink under load (credit + WAITING), instead of tying every business allocation to an MR registration.
> 3. **Decouple remote memory ownership** — SEND/RECV + local placement so the sender need not hold a remote rkey to the peer’s application buffers; each side manages its own placement memory.
> 4. **Acceptable latency for the target use case** — two-sided adds bounce/copy vs one-sided zero-copy, but for our PD-oriented scenarios the latency is acceptable, so we prefer lifetime/ownership simplicity over max GB/s.
> 
> One-sided registered WRITE/READ remains the right choice when stable MRs and peak bandwidth matter. `rdma_twosided` is an **opt-in alternate transport** (PD installs either `rdma` or `rdma_twosided`, not both).
> 
> ## Architecture / intrusion
> We agree folding this into the existing classic `rdma` transport is too invasive. Going forward we will implement two-sided as a **separate transport** (working name `rdma_twosided`):
> 
> * Keep the current `rdma` one-sided path largely untouched.
> * Put ctrl/credit/msg/bounce under the new transport tree (copy shared RDMA plumbing where needed for clarity, rather than deeply sharing hot-path code).
> * Reuse the existing TE handshake/metadata protocol (extend capability fields as needed), not a new discovery stack.
> 
> This PR (#3324) stays limited to the ctrl-frame codec and sender credit ledger (+ unit tests) as building blocks for that transport. We’ll relocate them under the `rdma_twosided` directory as we land the transport skeleton so they are not framed as shared classic-`rdma` core.
> 
> Happy to adjust naming/details if you have a preferred transport name or layout.

Have you considered implementing the same functionality using libfabric? After a quick review of the code, I think using an abstraction layer like libfabric would make the architecture more clean. In fact, the repository already contains some transports based on libfabric (EFA is a good example), so we can reuse the same underlying logic.

zhtshr:
> > > The core motivation for adding a two-sided RDMA path is still not clear enough to me. This PR introduces a fairly large amount of new machinery—control-plane credit, acknowledgements, bounce buffers, dedicated metadata, and accompanying documentation—but I do not yet see a sufficiently strong explanation of what concrete problem this solves that the existing one-sided path or current transport model cannot. I think the PR should first clarify the real workload or business scenario behind this change, what limitation or bottleneck exists in the current design, and whether the expected benefit is compatibility, flow control, stability, tail latency, throughput, or resource efficiency. Without a clear production-driven use case or benchmark evidence, this adds significant long-term maintenance and testing cost for the project.
> > > My second concern is that the implementation feels too intrusive to the current Transfer Engine architecture. From the description, this is not an isolated feature but a new communication mode whose control plane, credit model, and buffer-registration semantics are being woven into the existing TE path. That increases architectural complexity and makes the mainline logic harder to reason about, test, and evolve. If two-sided support is truly needed long term, I would strongly recommend maintaining it as a dedicated transport instead of folding so much specialized logic into the current transport flow. That would keep the abstraction boundaries cleaner, reduce coupling between one-sided and two-sided behavior, and make rollout, debugging, and future iteration much safer.
> > 
> > 
> > Thanks for the review — agreeing on both the motivation clarity and the architecture concern.
> > ## Motivation
> > We are not trying to replace the one-sided zero-copy path for peak bandwidth. The two-sided path targets a different trade-off for workloads such as PD:
> > 
> > 1. **Simpler application buffer lifetime** — avoid repeated `registerLocalMemory` / `unregisterLocalMemory` around transfers and the constraint that buffers must stay registered until there is no inflight work.
> > 2. **More flexible TE-owned networking buffers** — bytes move via TE bounce/RQ resources whose capacity can expand/shrink under load (credit + WAITING), instead of tying every business allocation to an MR registration.
> > 3. **Decouple remote memory ownership** — SEND/RECV + local placement so the sender need not hold a remote rkey to the peer’s application buffers; each side manages its own placement memory.
> > 4. **Acceptable latency for the target use case** — two-sided adds bounce/copy vs one-sided zero-copy, but for our PD-oriented scenarios the latency is acceptable, so we prefer lifetime/ownership simplicity over max GB/s.
> > 
> > One-sided registered WRITE/READ remains the right choice when stable MRs and peak bandwidth matter. `rdma_twosided` is an **opt-in alternate transport** (PD installs either `rdma` or `rdma_twosided`, not both).
> > ## Architecture / intrusion
> > We agree folding this into the existing classic `rdma` transport is too invasive. Going forward we will implement two-sided as a **separate transport** (working name `rdma_twosided`):
> > 
> > * Keep the current `rdma` one-sided path largely untouched.
> > * Put ctrl/credit/msg/bounce under the new transport tree (copy shared RDMA plumbing where needed for clarity, rather than deeply sharing hot-path code).
> > * Reuse the existing TE handshake/metadata protocol (extend capability fields as needed), not a new discovery stack.
> > 
> > This PR (#3324) stays limited to the ctrl-frame codec and sender credit ledger (+ unit tests) as building blocks for that transport. We’ll relocate them under the `rdma_twosided` directory as we land the transport skeleton so they are not framed as shared classic-`rdma` core.
> > Happy to adjust naming/details if you have a preferred transport name or layout.
> 
> Have you considered implementing the same functionality using libfabric? After a quick review of the code, I think using an abstraction layer like libfabric would make the architecture more clean. In fact, the repository already contains some transports based on libfabric (EFA is a good example), so we can reuse the same underlying logic.

Thanks for the suggestion. We considered libfabric — Mooncake already uses it for EFA/CXI.

For `rdma_twosided` we are staying on ibverbs. This path targets classic verbs devices (RoCE/IB/eRDMA) and should stay aligned with the existing `rdma` stack. The new work is mostly TE protocol (credit, bounce, control frames), which libfabric would not remove; it would mainly add another API layer.

Architecture-wise, the main cleanup is keeping two-sided as a **separate** transport rather than folding it into classic `rdma`. If we later need one two-sided protocol across EFA and verbs, revisiting OFI as a shared backend would make sense as a follow-up.

mooncake-transfer-engine/src/transport/rdma_twosided/sender_credit.cpp
```diff
@@ -0,0 +1,187 @@
...
+    if (it == entries_.end() || !it->second.has_update)
+        return ERR_TOO_MANY_REQUESTS;
+    auto &e = it->second;
+    for (size_t i = 0; i < kResourceCount; ++i) {
+        if (e.consumed[i] > e.grants[i] || n[i] > e.grants[i] - e.consumed[i])
+            return ERR_TOO_MANY_REQUESTS;
+    }
+    for (size_t i = 0; i < kResourceCount; ++i) e.consumed[i] += n[i];
+    return 0;
+}
```
alogfans, line 134: Epoch fencing is incomplete. `activate`, `applyGrant`, and `deactivate` validate the epoch, but `tryReserve`, `rollbackReservation`, and `available` only use `(peer, session)`. After a session ID is reactivated with a newer epoch, stale work from the previous epoch can consume or roll back credits belonging to the new epoch. Please pass and validate the epoch in every session operation, or include it in the key.

mooncake-transfer-engine/src/transport/rdma_twosided/ctrl_frame.cpp
```diff
@@ -0,0 +1,202 @@
...
+        readPod(ptr, end, frame.flags) || readPod(ptr, end, frame.session) ||
+        readPod(ptr, end, frame.epoch) || readPod(ptr, end, frame.seq) ||
+        readPod(ptr, end, frame.ack_seq) || readPod(ptr, end, payload_len))
+        return ERR_INVALID_ARGUMENT;
+    if (static_cast<size_t>(end - ptr) < payload_len)
+        return ERR_INVALID_ARGUMENT;
+    frame.type = static_cast<CtrlFrameType>(type);
+    frame.payload.assign(ptr, ptr + payload_len);
+    return 0;
+}
```
alogfans, line 250: Decoder validation should be tightened. The current codec does not reject unsupported versions, unknown frame types/flags, or trailing bytes. Payload decoders also accept trailing data, mutate output objects before full success, and reserve based on an untrusted count before checking that the payload is large enough. Please preflight the exact encoded size, parse transactionally, and define the version/type compatibility policy explicitly.

alogfans:
It's necessary to extend benchmarks/validators to show that proposed two-sided rdma transport works correctly and effectively.

alogfans (approved):
LGTM. However further PRs should be proposed to make this feature ready for work.

PR #3325, opened 2026-08-07
[Bugfix] clean up partially initialized TENT RDMA contexts

runzhech:
## Description

During `RdmaContext` initialization, the context remained in the `DEVICE_DISABLED` state while RDMA resources were being allocated. If initialization failed after partially creating resources, the error path called `disable()` for rollback. However, `disable()` returned immediately for a `DEVICE_DISABLED` context, leaving resources such as the device context, protection domain, epoll file descriptor, completion channels, and completion queues unreleased.

Because the state machine had no explicit initialization state, it could not distinguish a fully disabled context from a partially initialized one. Cleanup therefore relied on an inaccurate logical state instead of the resources actually owned by the context, potentially causing resource leaks and inconsistent lifecycle behavior after initialization failures.

To fix it, I've made the following changes:

- add an explicit `DEVICE_INITIALIZING` state
- roll back partially initialized RDMA resources based on actual handles
- make context cleanup idempotent and independent of disabled state
- prevent inbound connections and MR operations during initialization
- clean up residual resources during context destruction

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test results:**
- [ ] Unit tests pass

The test I've performed manually is described below:

Mock or wrap the `ibv_*` calls used by `RdmaContext::enable()` and inject failures at each initialization stage, including PD, epoll, completion channel, CQ, notification CQ, and port query creation.

For each failure, verify that:

initialization returns an error;
context status returns to `DEVICE_DISABLED`;
all previously created resources are released exactly once;
all handles, pointers, and containers are reset.
Call `disable()` again to verify idempotent cleanup, then remove the failure and call `enable()` to verify successful retry.

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

VSCode Copilot assisted with reading and editing the source files. I have reviewed every changed line and can defend the change end-to-end.

### DIFF
--- a/mooncake-transfer-engine/tent/include/tent/transport/rdma/context.h
+++ b/mooncake-transfer-engine/tent/include/tent/transport/rdma/context.h
@@ -24,6 +24,7 @@
 #include <cstdint>
 #include <list>
 #include <memory>
+#include <mutex>
 #include <string>
 #include <thread>
 #include <unordered_map>
@@ -123,9 +124,9 @@ class RdmaContext {
    private:
     int openDevice(const std::string &device_name, uint8_t port);
 
-    // Release every resource currently owned by this context in reverse
-    // construction order. This is intentionally state-independent so it can
-    // clean up a partially completed enable() and can safely be retried.
+    // Release every resource currently owned by this context. This is
+    // intentionally state-independent so it can clean up a partially completed
+    // enable() and can safely be retried.
     void cleanupResources();
 
    private:
--- a/mooncake-transfer-engine/tent/include/tent/transport/rdma/context.h
+++ b/mooncake-transfer-engine/tent/include/tent/transport/rdma/context.h
@@ -66,6 +66,7 @@ class RdmaContext {
     enum DeviceStatus {
         DEVICE_UNINIT,
         DEVICE_DISABLED,
+        DEVICE_INITIALIZING,
         DEVICE_ENABLED,
         DEVICE_PAUSED
     };
@@ -122,6 +123,13 @@ class RdmaContext {
 
    private:
     int openDevice(const std::string &device_name, uint8_t port);
+
+    // Release every resource currently owned by this context in reverse
+    // construction order. This is intentionally state-independent so it can
+    // roll back a partially completed enable() and can safely be retried.
+    void cleanupResources();
+
+    void rollbackEnable();
 
    private:
     // initialized during ctor, will never be changed during the context's
--- a/mooncake-transfer-engine/tent/include/tent/transport/rdma/context.h
+++ b/mooncake-transfer-engine/tent/include/tent/transport/rdma/context.h
@@ -66,7 +66,6 @@ class RdmaContext {
     enum DeviceStatus {
         DEVICE_UNINIT,
         DEVICE_DISABLED,
-        DEVICE_INITIALIZING,
         DEVICE_ENABLED,
         DEVICE_PAUSED
     };
@@ -126,10 +125,8 @@ class RdmaContext {
 
     // Release every resource currently owned by this context in reverse
     // construction order. This is intentionally state-independent so it can
-    // roll back a partially completed enable() and can safely be retried.
+    // clean up a partially completed enable() and can safely be retried.
     void cleanupResources();
-
-    void rollbackEnable();
 
    private:
     // initialized during ctor, will never be changed during the context's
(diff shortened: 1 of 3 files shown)


mooncake-transfer-engine/tent/src/transport/rdma/context.cpp
```diff
@@ -439,16 +453,34 @@ int RdmaContext::enable() {
 }
 
 int RdmaContext::disable() {
-    if (status_ == DEVICE_UNINIT || status_ == DEVICE_DISABLED) {
+    if (status_ == DEVICE_INITIALIZING) {
+        LOG(ERROR) << "RDMA context " << name()
+                   << " cannot be disabled while it is initializing";
+        return -1;
+    }
```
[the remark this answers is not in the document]
runzhech: `disable()` is not designed as a concurrent cancellation path for `enable()`. All `enable()` failures already call `rollbackEnable()` to clean up partial resources.
Supporting cancellation requires explicit synchronization, so rejecting `disable()` during `DEVICE_INITIALIZING` is safer under the current contract.

staryxchen (commented):
Thanks for the contribution. The core leak fix is correct, but I think the state-machine extension is more than the current call graph requires.

cleanupResources() (nullity-based, state-independent cleanup) is the real fix for the PD / epoll / comp_channel / CQ / device_context leaks — that part belongs in disable() and a simpler disable() early-return on resource-nullity would suffice. The std::make_unique<RdmaCQ>() change is also correct and necessary: a CQ
  that fails construct() never enters cq_list_, so disable() can't reclaim it — this must be fixed at the allocation site.

What I'd question is DEVICE_INITIALIZING + rollbackEnable(). In the TENT call graph, enable() is only called from construct() (single-threaded install()), and disable() only from ~RdmaContext() and enable()'s own failure path — both same-thread, externally serialized by the transport lifecycle. There's no concurrent
  caller the new state coordinates against today. Once disable()'s early-return is resource-based, enable() could keep calling disable() directly; rollbackEnable() is needed only because disable() now rejects DEVICE_INITIALIZING.

The new state + CAS are defensible as future-proofing, but for the present code they add machinery without addressing a current race. I'd lean toward trimming them unless there's a concurrent caller I'm missing.

runzhech:
@staryxchen As you noted, there's no concurrent caller today that would need the CAS/state machinery. I've made the following changes:
- Removed DEVICE_INITIALIZING and rollbackEnable().
- Made disable() resource-driven: early-return on empty resources, do actual cleanup via cleanupResources().
- enable() now calls disable() directly on failure.

Please take another look.

mooncake-transfer-engine/tent/src/transport/rdma/context.cpp
```diff
@@ -439,16 +440,23 @@ int RdmaContext::enable() {
...
         return -1;
     }
 
+    cleanupResources();
+    status_ = DEVICE_DISABLED;
+    return 0;
+}
+
+void RdmaContext::cleanupResources() {
     for (auto& entry : mr_set_) {
```
staryxchen: mr_set_ is iterated and cleared here without holding mr_set_mutex_, while every other access site takes the lock. In the current call graph disable() runs during teardown (same thread, after workers are stopped), so this isn't a live race today — but the broken locking discipline could bite if teardown ordering ever changes. Consider holding mr_set_mutex_ around the iterate-and-clear to keep the invariant, or add a comment documenting the relies-on-teardown-ordering assumption

mooncake-transfer-engine/tent/include/tent/transport/rdma/context.h
```diff
@@ -123,6 +123,11 @@ class RdmaContext {
    private:
     int openDevice(const std::string &device_name, uint8_t port);
 
+    // Release every resource currently owned by this context in reverse
```
staryxchen: the comment says "reverse construction order", but the actual cleanup isn't strictly reverse — cq_list_ is cleaned before notify_cq_ (constructed after it), and event_fd_ before comp_channel_ (also constructed after). Functionally the order is fine (no interdependency requiring strict reverse), but the comment misleads future readers. It's better to drop "reverse construction order" from the comment.

runzhech:
@staryxchen 
The latest commit made the following 2 changes:
Held mr_set_mutex_ around the mr_set_ iterate-and-clear in cleanupResources() to match the locking discipline of the other access sites.
Removed "reverse construction order" from the cleanupResources() comment, since the actual cleanup order is not strictly reverse.
Thanks for catching those.

staryxchen (approved):
Thanks for the contribution! LGTM

Issue #3124, opened 2026-07-26
[RFC][Store] Selective candidate materialization for BatchEvict


jacklin78911-collab:
### Changes proposed

A draft implementation is available in #3118. This RFC documents the design, correctness boundaries, alternatives, and complexity trade-offs for review.

## Background

`MasterService::BatchEvict` materializes a full candidate — including tenant identity and key — for every eviction-eligible object, and only then selects the oldest requested subset.

When the requested eviction target is a small fraction of the eligible population, most of that identity construction is discarded without being used. For example, with 1M eligible objects and a 1% target, roughly 1M full candidates are constructed in order to evict roughly 10K objects. The cost of preparing candidates is therefore governed by the eligible population rather than by the amount of work actually performed.

Because `BatchEvict` holds a `snapshot_mutex_` shared lock for its duration, this preparation cost is also exported to control-plane operations that need the unique lock.

Related performance investigation: #2560.

## Proposal

Keep the existing O(N) metadata census, but materialize full candidate identity only for:

- the requested eviction frontier;
- a bounded reserve absorbing execution-time revalidation failures;
- additional candidates found through refill if that reserve is exhausted.

For high eviction ratios the existing single-pass full-materialization path is retained, since most eligible candidates would be needed anyway and the two-phase approach would otherwise add a census without reducing materialization. The ratio at which the implementation switches is not a tuned constant: it is the analytic point at which the reserve would already exceed the frontier limit, so it follows from the reserve-slack and frontier-limit parameters rather than being chosen independently.

## Correctness

The change affects *when* full candidate identity is constructed, not eviction eligibility or execution semantics. The eligibility conditions used by the census are identical to those used when materializing, so the selected set matches what the census counted. Preserved unchanged:

- oldest-first candidate selection;
- execution-time lease revalidation;
- soft-pin fallback;
- whole-group eviction semantics;
- replica constraints;
- target and lower-bound behavior;
- offload behavior.

Because metadata is read twice on the selective path, an object may stop being eligible between the census and materialization. Two guards cover this: if the frontier holds fewer objects than the target requires, the implementation falls back to full candidate collection so the eviction count is still derived from the requested target; and if the reserve is exhausted during execution, refill continues past the cutoff until the target is met.

## Relation to prior work

[...]

PR #3327, opened 2026-08-07
[Store] Add deterministic MasterService eviction scenarios

Aionw:
## Description

Move the deterministic `BatchEvict` correctness coverage into a dedicated
`master_service_evict_scenario_test` target and express it through the shared
`MasterScenario` DSL.

This change adds collection-level scenario declarations for building and
checking object ranges, so eviction cases describe object state, eviction, and
postconditions without test-local setup/assertion loops. It also consolidates
the existing BatchEvict OpLog tests and adds independently passing interaction
coverage for:

- tenant-scoped REMOVE OpLog records;
- OpLog reservation failure leaving the candidate readable;
- global eviction preserving tenant isolation and quota accounting;
- tenant admission evicting only that tenant's expired object.

The focused algorithm scenarios continue to cover oldest-first selection,
target/lower-bound behavior, soft-pin fallback, whole-group safety, high-ratio
eviction, reserve consumption, and refill after reserve exhaustion.

This PR changes test organization and test infrastructure only; it does not
change MasterService production behavior.

Related RFC: #3124. The selective candidate implementation from #3118 is
already merged; this PR separates and extends its deterministic correctness
coverage rather than duplicating that implementation.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [x] Other: test coverage

## How Has This Been Tested?

**Test commands:**

```bash
cmake --build build \
  --target master_scenario_test \
           master_service_scenario_test \
           master_service_evict_scenario_test \
  --parallel 128

  -R '^(master_scenario_test|master_service_scenario_test|master_service_evict_scenario_test)$'
```

The final 12-test Evict scenario set was also run directly against a clean
`main` production build: 12/12 passed.

**Test results:**

- [x] Unit tests pass
- [ ] Manual testing done

## Checklist

- [x] I have formatted my code using the staged C/C++ formatting hook
- [x] Documentation is not applicable because production behavior is unchanged
- [x] I have added tests to prove the covered behavior is effective
- [x] Related RFC exists for the underlying BatchEvict work: #3124

Staged pre-commit hooks pass. The repository's whole-file CMake formatter also
rewrites unrelated historical lines in `tests/CMakeLists.txt`; that unrelated
formatting churn is intentionally excluded from this PR.

## AI Assistance Disclosure

- [ ] No AI tools were used
- [x] AI tools were used

Codex assisted with test migration, the scenario DSL extension, interaction
test construction, formatting, and test execution. The human submitter reviewed
and approved the submitted diff and remains responsible for the change.


Icedcoco:
I went through the migration and had three questions/comments:

1. It looks like `BatchEvictReleasesMemoryAfterDurable` was removed without equivalent coverage. That test verifies that memory and quota are released only after the OpLog becomes durable. Could we retain it or port it to the new suite?

2. What semantics do we want `DoesNotExist()` to represent? It currently checks whether `GetReplicaList()` returns `OBJECT_NOT_FOUND`, which can mean either that the metadata has been deleted or that all replicas are temporarily marked `REMOVED`. Would `IsUnreadable()` be clearer here, with a separate assertion for durable cleanup, or do you prefer the current client-visible interpretation?

3. How do you see the intended boundary of `MasterScenario`? It now has friend access to `MasterService`, directly updates lease metadata, and inspects `ordered_oplog_writer_`. RFC #3098 suggested keeping private fault injection in focused tests. Would it be better to keep these eviction-specific white-box controls in the dedicated fixture, or is expanding the shared DSL intentional here?

Aionw:
Addressed the first point in commit `393cc339`.

I restored the durable eviction reclamation scenario in `master_service_evict_scenario_test`. It now verifies that:

- the evicted object is client-visible as missing while the OpLog transaction is blocked;
- tenant quota cannot be reused before the remove record becomes durable;
- after durability, quota is reclaimed and a replacement object can be admitted;
- the memory node is limited to one object, so admitting that replacement also verifies that the evicted allocation was released.

The post-durability quota assertion is eventual because the batch can become visible in storage before `OrderedOpLogWriter` runs the durable-finalize callback on its separate callback thread.

Aionw:
For the remaining two points:

2. The current client-visible interpretation is intentional. `MasterService` exposes a replica-oriented read API, and `GetReplicaList()` deliberately returns `OBJECT_NOT_FOUND` both when metadata is absent and when every replica is `REMOVED`. `DoesNotExist()` therefore means “not found through the public API”; it is not asserting that the internal metadata entry has already been physically erased. `IsUnreadable()` would be weaker because it would also conflate this state with `REPLICA_IS_NOT_READY`.

3. The shared DSL’s white-box access is intentional but narrowly scoped to test setup and synchronization. Objects are still created through the public API; `ExpireAt()` only adjusts lease/pin timestamps to construct otherwise impractical deterministic eviction preconditions, using the existing shard accessor and object lock. Likewise, inspecting `ordered_oplog_writer_->IsAccepting()` is a synchronization point for fault injection. Eviction behavior and correctness are asserted through public `MasterService` APIs rather than by inspecting or rewriting the resulting internal state. That is the boundary I intend for `MasterScenario`: white-box controls may arrange deterministic preconditions, but behavioral outcomes should remain black-box assertions.

Icedcoco:
LGTM pending CI

ykwd:
@Aionw The code format check failed

ykwd (approved):

Icedcoco (approved):

PR #3333, opened 2026-08-07
[Store] Fix race in HA durable-finalization tests (#3332)

### LINKED ISSUE #3332: [Store][Test] Wait for durable-finalize callbacks in HA tests
Several HA tests wait for an OpLog batch to become visible in the backend and then immediately assert that resources have been reclaimed.

This is not a sufficient synchronization condition. The batch transaction and the durable-finalize callback are separate steps.

### Observed failures

- PR #3201: `MasterServiceHATest.RemoveAllFinalizesAfterDurable` failed with `TENANT_QUOTA_EXCEEDED` when checking the post-finalize `PutStart`.
- PR #3327: `MasterServiceHATest.RemoveHidesBeforeDurableAndReleasesAfterFinalize` failed with the same error.
- The failures occurred in the Python 3.10 build while the Python 3.12 build passed, which is consistent with a timing-dependent test race.

### Root cause

The affected tests use this pattern:

```cpp
backend->AllowTxn();
ReadBatchEventually(storage, batch_id, batch);
auto after_finalize = service.PutStart(...);
EXPECT_TRUE(after_finalize.has_value());
```

`ReadBatchEventually()` only confirms that the batch record is readable from the HA backend. After the backend transaction succeeds, `OrderedOpLogWriter` updates its durable state and enqueues the finalize callback on a separate callback thread. The callback may run after `ReadBatchEventually()` returns.

As a result, the test can observe the batch record while the quota, memory, or disk resource is still held by the removed object. The immediate `PutStart()` then fails with `TENANT_QUOTA_EXCEEDED` even though the OpLog transaction has already completed.

### Scope

Please audit the durable-finalization tests in `mooncake-store/tests/ha/master_service_ha_test.cpp`, including:

- `RemoveDurableCallbackReleasesResources`
- `RemoveHidesBeforeDurableAndReleasesAfterFinalize`
- `BatchRemoveFinalizesEachObjectAfterDurable`
- `RemoveAllFinalizesAfterDurable`
- `BatchReplicaClearAllReleasesAfterDurable`
- `BatchReplicaClearSegmentReleasesAfterDurable`
- `NoFBatchEvictReleasesNoFSpaceAfterDurable`

Other tests that check local disk, memory, or quota state immediately after `ReadBatchEventually()` should be reviewed as well.

### Suggested direction

Keep `ReadBatchEventually()` for asserting that the batch record was persisted, but use a separate bounded wait for the actual finalize side effect:

- poll tenant quota until the expected value is reached;
- poll memory or local-disk usage until it is reclaimed; or
- add a focused test-only synchronization helper if no public observable state is available.

The tests should never assume that backend batch visibility implies that the durable-finalize callback has already completed.

### Acceptance criteria

- The affected tests wait for the resource-reclamation condition rather than only backend batch visibility.
- The tests remain deterministic under different thread scheduling.
- Re-running the HA test target repeatedly does not produce intermittent `TENANT_QUOTA_EXCEEDED` failures.

Icedcoco:
## Description

Fixes #3332.

Several HA tests assumed that an OpLog batch being readable from the backend meant that the durable-finalize callback had already completed. Since persistence and callback execution occur asynchronously, the tests could attempt to allocate resources too early and intermittently fail with `TENANT_QUOTA_EXCEEDED`.

This change:

- Adds a bounded `PutStartEventually` helper for quota/resource reclamation checks.
- Updates remove, batch remove, remove-all, replica-clear, batch-evict, and NoF batch-evict tests to wait for successful allocation after durable finalization.
- Polls local-disk usage until reclaimed in the disk eviction test.

## Module

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

**Test commands:**

```bash
SKIP=cmake-format pre-commit run --files mooncake-store/tests/ha/master_service_ha_test.cpp
```

**Test results:**

- [x] Targeted pre-commit checks pass.

## Checklist

- [x] I have formatted my code using the project hooks
- [x] I have run `pre-commit run --all-files`
- [x] I have added test synchronization to cover the reported race
- [x] No documentation update is required

## AI Assistance Disclosure

- [x] AI tools were used.

AI assistance was used to inspect issue #3332, audit the affected HA tests, implement the minimal synchronization changes, and run formatting and pre-commit checks. The human submitter has reviewed the changes and is responsible for understanding and defending them.

### DIFF
--- a/mooncake-store/include/master_service.h
+++ b/mooncake-store/include/master_service.h
@@ -130,6 +130,9 @@ class MasterService {
         std::function<bool(const std::string&, uint32_t, std::string*)>;
     using DurableFinalizeCallback =
         std::function<void(const OpLogEntry& durable_entry)>;
+    using BatchOpLogWriterFactory =
+        std::function<std::unique_ptr<OrderedOpLogWriter>(
+            OrderedOpLogWriterConfig, OrderedOpLogWriter::WriteBatchFn)>;
 
     MasterService();
     MasterService(const MasterServiceConfig& config);
@@ -152,6 +155,7 @@ class MasterService {
 
     ErrorCode SetBatchOpLogBackendForTesting(
         std::shared_ptr<HaKvBackend> backend);
+    void SetBatchOpLogWriterFactoryForTesting(BatchOpLogWriterFactory factory);
 
     /**
      * @brief Test-only wrapper around BatchEvict / NoFBatchEvict so that
@@ -2288,6 +2292,7 @@ class MasterService {
     std::shared_ptr<HaKvBackend> batch_oplog_kv_backend_;
     std::unique_ptr<OpLogBatchStorage> batch_oplog_storage_;
     std::unique_ptr<OrderedOpLogWriter> ordered_oplog_writer_;
+    BatchOpLogWriterFactory batch_oplog_writer_factory_;
 
     // OpLog publishing helpers
     std::string SerializeMetadataForOpLog(const ObjectMetadata& metadata) const;
--- a/mooncake-store/include/ha/oplog/ordered_oplog_writer.h
+++ b/mooncake-store/include/ha/oplog/ordered_oplog_writer.h
@@ -54,18 +54,17 @@ class OrderedOpLogWriter {
 
     OrderedOpLogWriter(OrderedOpLogWriterConfig config,
                        WriteBatchFn write_batch);
-    ~OrderedOpLogWriter();
+    virtual ~OrderedOpLogWriter();
 
     tl::expected<Reservation, ErrorCode> Reserve();
-    tl::expected<PendingHandle, ErrorCode> Commit(Reservation&& reservation,
-                                                  OpLogEntry entry,
-                                                  DurableCallback callback);
+    virtual tl::expected<PendingHandle, ErrorCode> Commit(
+        Reservation&& reservation, OpLogEntry entry, DurableCallback callback);
     void Abort(Reservation&& reservation);
 
     bool IsAccepting() const;
     ErrorCode LastError() const;
     void Start();
-    void Stop();
+    virtual void Stop();
 
    private:
     struct Impl;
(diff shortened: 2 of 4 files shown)


Aionw:
Could we avoid fixing this with `PutStartEventually()`? It turns every `PutStart()` failure into a retry and uses a mutating business operation as the synchronization primitive, which may hide unrelated failures.

The test needs deterministic control over two distinct stages:

1. The OpLog batch has been persisted.
2. Its durable-finalize callback has completed.

I suggest making the OpLog writer an injectable dependency of `MasterService` via a writer interface/factory. Tests could then use a gated writer, or a decorator around `OrderedOpLogWriter`, that exposes separate `PersistThrough(sequence_id)` and `RunCallbacksThrough(sequence_id)` controls.

This would let the tests assert resource ownership before callback execution and reclamation immediately after it, without polling.

The callback control should remain a property of the test writer rather than being exposed through `OrderedOpLogWriterConfig`, since the config should contain only production batching and runtime configuration.

mooncake-store/include/master_service.h
```diff
@@ -2288,6 +2292,7 @@ class MasterService {
     std::shared_ptr<HaKvBackend> batch_oplog_kv_backend_;
     std::unique_ptr<OpLogBatchStorage> batch_oplog_storage_;
     std::unique_ptr<OrderedOpLogWriter> ordered_oplog_writer_;
+    BatchOpLogWriterFactory batch_oplog_writer_factory_;
```
Aionw, line 2498: Can we ensure this factory is not nullptr? That says, explicitly initialize it at MasterService's constructor, not just adding a new member variable only for testing.

Icedcoco: Good point. I’ve updated the constructor to initialize `batch_oplog_writer_factory_` with the production `OrderedOpLogWriter` factory. `InitializeBatchOpLogWriter()` now always uses the factory, and the test-only setter rejects an empty factory. The relevant HA tests and pre-commit checks pass. Fixed in `901fe1200`.

Aionw (approved):
LGTM

ykwd (approved):

00fish0 (approved):

ykwd:
@Icedcoco  Could you resolve the conflicts before merging? thanks

Icedcoco:
> @Icedcoco Could you resolve the conflicts before merging? thanks

@ykwd Conflicts resolved! Ready for merge.

PR #2336, opened 2026-06-05
[Store][DO NOT Merge Now](fix) avoid MR resource leak and destructor hang on master failure

LujhCoconut:
## Description

This PR fixes RDMA Memory Region (MR) resource leaks in `mooncake-store` that occur when the `master service` exits or becomes unreachable before the `store service` gracefully unmounts its segments.

### Root Cause (source code verified)

Three independent code paths leak MRs:

1. **`Client::~Client()` performs synchronous RPC to master during destruction.**
   When master is dead, each `UnmountSegment()` call hits `UnmountSegmentImpl()`, which returns early on RPC failure without calling `transfer_engine_->unregisterLocalMemory()`. All mounted segments' MRs are leaked.
   - File: `client_service.cpp:355-363` (destructor loop)
   - File: `client_service.cpp:2598-2603` (early return on RPC failure)

2. **`UnmountSegmentImpl` bails out on RPC failure without local cleanup.**
   If `master_client_.UnmountSegment()` fails, the function returns immediately, skipping `transfer_engine_->unregisterLocalMemory()`.
   - File: `client_service.cpp:2598-2603`

3. **Mount path lacks rollback on RPC failure.**
   In `MountSegmentAndGetId`, `registerLocalMemory()` succeeds first, but if the subsequent `MountSegment()` RPC fails, the TE-side MR is never unregistered.
   - File: `client_service.cpp:2670-2696`

### Why RPC is skippable in destructor

`MasterService::ClientMonitorFunc` (`master_service.cpp:5658`) handles dead clients via TTL expiry:
- Store sends periodic `Ping` heartbeats
- If store exits, heartbeats stop
- After `client_live_ttl_sec_`, master marks client as expired
- `ClientMonitorFunc` auto-runs `PrepareUnmountSegment → ClearInvalidHandles → CommitUnmountSegment`

Therefore, the destructor's `UnmountSegment` RPC is best-effort optimization, not a correctness dependency.

### What this PR fixes

| Problem | Location | Fix |
|---------|----------|-----|
| Destructor MR leak when master is dead | `Client::~Client()` | Skip master RPC, call `unregisterLocalMemory()` directly |
| `UnmountSegmentImpl` skips TE cleanup on RPC failure | `Client::UnmountSegmentImpl()` | Continue with `unregisterLocalMemory()` even if master notification fails |
| Mount path MR leak on RPC failure | `Client::MountSegmentAndGetId()` | Rollback `registerLocalMemory()` if `MountSegment()` fails |

### Changes

[...]

PR #3335, opened 2026-08-07
[2/3] [Store] Add EGM-backed Store pool over NVLink

neverhook:
## Summary

This is the Store-only follow-up for RFC #2914. It adds a default-off,
Provider-owned EGM pool while keeping the Consumer API, Master schema, and
placement behavior unchanged.

## Stack

This is **[2/3]** of the RFC implementation stack:

1. #2966 — HOST_NUMA/Fabric Transfer Engine prerequisite.
2. #3335 — Provider-owned EGM Store pool (this PR).
3. #3431 — direct remote-memory reads into GPU destinations.

The head branch is a strict child of #2966. GitHub shows both commits against
`main` because a community PR cannot use a branch from the submitter's fork as
its base. After #2966 merges, this branch will be rebased onto `main` and the
review diff will collapse to the single Store commit.

### Minimal EGM Store pool

- Adds `enable_egm_store_pool=false|true` and
  `egm_numa_nodes=auto|<comma-separated NUMA IDs>`.
- Requires `protocol=nvlink`, `global_segment_size>0`, and
  `local_buffer_size=0` when enabled.
- Discovers or explicitly selects HOST_NUMA nodes; computes common Store/CUDA
  alignment; rounds capacity down; distributes it across nodes; and splits
  allocations by aligned `max_mr_size`.
- Mounts the resulting allocations as ordinary `nvlink` Store segments.
- After the final local endpoint is selected, configures a call-scoped
  Transfer Engine with discovery disabled and only `nvlink` installed, then
  injects it through the existing `Client::Create()` parameter. Each auto-port
  attempt uses a fresh engine and installation failures do not fall back to
  RDMA, TCP, or topology discovery.

### Retryable ownership and publication lifecycle

- Completes all allocations before registration or Master publication.
- Ordinary Store mounts and EGM mounts share one private caller-supplied UUID
  transaction. It inserts the local record before Transfer Engine registration
  and Master publication, so either failure can use the same rollback path.
- Retains allocation ownership across ambiguous RPC outcomes and error returns
  carrying a partial owner.
- Cleans up in reverse order and does not release allocations before
  publication cleanup completes.
- Retains incomplete records so later `close()` calls retry only unfinished
  stages.
- Clears the Python native client only after teardown succeeds.

### Python setup guard

- Both `setup()` overloads and `setup_dummy()` reject a new setup while a
  native client still exists.
- This includes an initially failed setup: callers must run `close()` first so
  any cleanup-pending state is preserved and retried.
- A successful `close()` clears the object and permits setup again.

## Provider and Consumer configuration

The EGM Provider configures and injects a private NVLink-only Transfer
Engine after selecting its final endpoint:

```python
provider.setup({
    "local_hostname": "10.192.8.81:12345",
    "metadata_server": "http://10.192.8.81:8079/metadata",
    "master_server_addr": "10.192.8.81:50051",
    "protocol": "nvlink",
    "global_segment_size": str(64 * 1024**3),
    "local_buffer_size": "0",
    "rdma_devices": "",
    "enable_egm_store_pool": "true",
    "egm_numa_nodes": "auto",
})
```

A normal Consumer does not receive the Provider-side injected engine. It must
enable Transfer Engine discovery, and on hosts with an RDMA HCA it must force
MNNVL selection:

```bash
export MC_MS_AUTO_DISC=1
export MC_FORCE_MNNVL=1
```

```python
consumer.setup({
    "local_hostname": "10.192.8.58:12400",
    "metadata_server": "http://10.192.8.81:8079/metadata",
    "master_server_addr": "10.192.8.81:50051",
    "protocol": "nvlink",
    "global_segment_size": "0",
    "local_buffer_size": "0",
    "rdma_devices": "",
    "enable_egm_store_pool": "false",
})
```

`MC_MS_AUTO_DISC=1` enables discovery; `MC_FORCE_MNNVL=1` makes discovery
choose cross-node NVLink instead of RDMA when HCAs are present. With discovery
unset or `0`, an ordinary `protocol=nvlink` Consumer can still reach the
existing `unsupported_protocol` path. This PR documents that existing behavior
without changing generic transport initialization.

## Scope decisions

- HOST_NUMA-backed local workspace support is deliberately not included;
  `local_buffer_size=0` remains the Provider-pool contract.
- Metrics, Consumer rewrites, RDMA/TENT changes, generic metadata transactions,
  and the GB200 harness remain outside the production diff.
- The cross-node GPU-read staging fix is the focused [3/3] layer in #3431.

The separate GB200 Provider/Consumer harness and results are not part of this
production branch.

## Review size

Store increment relative to #2966:

- 1 logical Store commit.
- 16 changed files.
- +1,659/-44 lines.

Temporary stacked GitHub view against current `main`:

- 2 logical commits.
- 28 changed files.
- +4,229/-46 lines.

## Validation

This rebase keeps the Store implementation unchanged. The EGM test now runs
in the ordinary CTest sweep with the fake-driver NVLink tests inherited from
#2966; the separate build and Store-specific path filter are removed.

- [x] clang-format 20 check on PR-changed C/C++ lines.
- [x] `git diff --check` and PR-scoped hooks, with the baseline CMake exception below.
- [x] CUDA 12.8 ARM64 focused build.
- [x] `nvlink_vmm_unit` (2 tests).
- [x] `egm_store_pool_unit` in CUDA-on and CUDA-off builds.
- [x] CUDA-off Transfer Engine, Store, Master and Python binding build.
- [x] CPU/TCP tensor snapshot regressions (3 tests, no skips) and probe API exports.
- [x] Python setup/close lifecycle tests (2 tests).
- [x] New-head complete `unit-tests` sweep and GitHub Actions CI Gate.

Local CUDA-on/off regression tests passed on the composed stack (#3431).
GitHub Actions passed the complete CTest sweep (227 entries) and CI Gate
for this revision. Both fake-driver entries ran successfully; the hardware-only
entry was excluded on the CPU runner.

The CMake hook also flags an unchanged upstream line for
`dynamic_replication_lease_table_test`; that unrelated formatting change is
not included. Other applicable hooks pass. Repository-wide
`pre-commit run --all-files` is not a merge gate.

Related RFC: [#2914](https://github.com/kvcache-ai/Mooncake/issues/2914)

## Module

- [x] CI/CD
- [x] Docs

## Type of Change

- [x] New feature
- [x] Documentation update
- [x] Bug fix

## Checklist

- [x] I have formatted changed C++ code using clang-format 20.1.8
- [x] I have added focused tests to prove the changes are effective
- [x] For changes >500 LOC: RFC #2914 exists

## AI Assistance Disclosure

- [x] AI tools were used

Codex assisted with the semantic rebuild on #2966, scoped lifecycle changes,
focused tests, validation, and PR preparation. The human submitter remains
responsible for reviewing, understanding, and defending every changed line.


neverhook:
Current Store head is `a94583d4`, with #2966 (`3c850601`) as its only prerequisite. The production increment is default-off Provider EGM capacity, aligned NUMA/chunk planning, ordinary nvlink segment publication, and retryable reverse cleanup; `local_buffer_size=0` remains intentional. Local CUDA-on/off focused tests, Python binding, clang-format 20, diff/YAML checks, and strict Sphinx build pass. The separate GB200 branch records HBM→EGM Put and EGM→HBM Get latency/bandwidth plus correctness and teardown evidence; results are pending hardware execution.

bhaktatejas922:
Very interested in this on GB300 NVL72. We run PD-disaggregated serving there with MNNVL/IMEX and fabric-handle KV transfer already in production, and a store pool reachable over NVLink instead of RDMA would be a major win for cross-engine prefix restore at long context. Happy to run the provider/consumer validation flow on real GB300 hardware and report results here if useful.

mooncake-transfer-engine/include/transport/nvlink_transport/nvlink_host_numa_allocation.h
```diff
@@ -0,0 +1,145 @@
...
+    int numa_node_ = -1;
+    bool mapped_ = false;
+    bool address_reserved_ = false;
+    bool handle_owned_ = false;
+    bool handle_release_failed_ = false;
+    uint64_t allocation_handle_ = 0;
+#if defined(USE_MNNVL) && defined(USE_CUDA) && CUDA_VERSION >= 12030
+    bool owned_range_registered_ = false;
+    DriverApi driver_api_;
+#endif
```
alogfans: As this feature requires both MNNVL and specified CUDA driver version, we can use macro alias to reflect this and show it in compile time and/or runtime.

neverhook:
> Very interested in this on GB300 NVL72. We run PD-disaggregated serving there with MNNVL/IMEX and fabric-handle KV transfer already in production, and a store pool reachable over NVLink instead of RDMA would be a major win for cross-engine prefix restore at long context. Happy to run the provider/consumer validation flow on real GB300 hardware and report results here if useful.

Thanks a lot for offering to help with the validation — this would be extremely valuable.

A real GB300 NVL72 environment with MNNVL/IMEX already in production is especially interesting for this PR. The current validation branch was prepared around the Provider/Consumer flow on GB200/NVL72, so having the same path exercised independently on GB300 would give us very useful additional coverage.

Please feel free to run the validation flow and report the results here. Correctness and teardown behavior are the most important parts, and any transfer latency/bandwidth numbers would also be very helpful.

Really appreciate you volunteering the hardware and time for this.

neverhook:
Scope update: the remote GPU-read no-staging change has been removed from this Store pool commit and submitted independently as #3431. This PR is again Store-only (`74c61458`, +1,104/-28 relative to #2966); the separate GB200 validation branch will compose both changes for EGM-to-HBM testing.

neverhook:
## GB200/NVL72 validation result: PASS

This run provides hardware evidence for the composed EGM Store stack proposed in RFC #2914. It does not validate the current #3335 head in isolation.

The run used two GB200 nodes in the same NVL72 NVLink domain. The separate validation branch combined the HOST_NUMA prerequisite from #2966, the Store pool from #3335, and #3431, which supplies the direct remote-EGM-to-HBM read behavior required by the `local_buffer_size=0` Get test. The benchmark harness itself remains outside the production PRs.

### Hardware acceptance result

| Validation area | Result |
|---|---|
| HBM → remote EGM Put correctness | PASS |
| Remote EGM → HBM Get correctness | PASS |
| Single-GPU and four-GPU concurrent transfers | PASS |
| Sustained transfers after excluding first-use initialization | PASS |
| Store Put/Get with Consumer `local_buffer_size=0` | PASS |
| Provider metadata removal and complete teardown | PASS |
| Fabric health and C2C error checks | PASS; no new errors |

The test covered 208 Put/Get pairs across four GPUs and 128 MiB, 512 MiB, 1 GiB, and 2 GiB logical payloads. This count includes 16 first-use samples, 32 warmup samples, and 160 steady-state samples. Device-side pattern generation and verification were used so large-payload measurements did not include host copies.

### MNNVL bandwidth

`EGM → HBM` means a Consumer GPU reading Provider EGM. `HBM → EGM` means a Consumer GPU writing Provider EGM.

| Direction | Single-GPU sustained p50 | Four-GPU aggregate p50 |
|---|---:|---:|
| EGM → HBM | 211.77–211.90 GB/s | 390.62 GB/s |
| HBM → EGM | 193.66–193.88 GB/s | 336.98 GB/s |

The scan selected 4 GiB with one stream for both directions. Increasing the payload from 2 GiB to 4 GiB changed throughput by only 0.025% for EGM → HBM and 0.014% for HBM → EGM, so the matrix reached a clear plateau. Sustained-run variation was well below the 5% stability limit.

For context, `nvidia-smi c2c -s` reported a per-GPU C2C capacity sum of 223.56 GB/s on these nodes. The measured single-GPU results are approximately 94.8% and 86.7% of that runtime-reported value; these ratios are not a general NVLink-utilization metric. The separate 450 GB/s architectural reference uses a different reporting basis and is recorded only for comparison, not as a pass threshold.

### Store path

At the largest tested logical payload per GPU:

- Put aggregate p50: 191.10 GiB/s
- Get aggregate p50: 175.92 GiB/s
- Put/Get maxima: 211.08/211.82 GiB/s

The Provider publishes approximately 150 MiB Store segments, so logical payloads larger than one segment were tested as batches of Store objects no larger than 128 MiB each. These numbers measure aggregate Store throughput; they do not imply that one Store key spans multiple segments.

### Route and lifecycle evidence

Raw MNNVL, Store, and Provider-endpoint route checks passed. Direct C2C byte counters were not available, so route qualification used `C2C_ROUTE_INFERRED`: NVLink Tx/Rx deltas, C2C capability/error counters, Fabric health, and expected transfer direction.

- Fabric remained healthy and every GPU reported zero new C2C errors.
- Every Consumer completed its cleanup.
- Node A unpublished all Provider metadata and reported no remaining EGM chunk records.
- The Store Provider, Master, and supplementary RDMA target exited successfully.
- Node A received the successful Node B completion for the same validation session and configuration.

### Supplementary results

The following results are useful context but are not part of the current Store Put/Get hardware gate.

#### GPU load/store latency

Dependent GPU-SM accesses were measured without setup, mapping, or kernel-launch time:

- Remote EGM dependent load p50: 2.38–2.45 μs/op
- Remote EGM store plus system-fence completion p50: 2.35–2.36 μs/op
- Local EGM dependent load p50: 0.66–0.69 μs/op
- EGM-to-EGM load/store plus system fence: 3.05–3.36 μs/op

These latency measurements are forward-looking evidence for the sparse-access/DSA motivation in the RFC. The production change in #3335 still exposes ordinary Store Put/Get, not a public remote-pointer API.

#### EGM H2H and host-DRAM RDMA

- EGM-to-EGM CE/SM bandwidth and route checks passed.
- The host-DRAM RDMA benchmark was collected only as a DRAM/RNIC comparison and is not part of the MNNVL acceptance decision.

### Verdict

**The GB200/NVL72 hardware gate for the composed stack passes.**

This result closes the hardware item for review purposes. #3335 still needs its current base conflict resolved, a rebase after #2966, and CI on the resulting Store-only head before it can be considered merge-ready.

mooncake-transfer-engine/include/transport/nvlink_transport/nvlink_host_numa_allocation.h
neverhook: Addressed in [1/3] #2966 while rebasing the stack. The repeated build guard is now represented by `MOONCAKE_NVLINK_HOST_NUMA_ENABLED` and is used consistently across the allocator, transport, and focused tests. The alias covers `USE_MNNVL`, `USE_CUDA`, and CUDA Toolkit 12.3+ at compile time. Runtime driver/Fabric/VMM support is still validated through the existing CUDA device-attribute checks, since `CUDA_VERSION` describes the build headers rather than the installed driver's actual capabilities. CUDA 12.8 ARM64 focused tests and the CUDA-off build both pass.

neverhook:
Stack refresh: this PR is now [2/3] and its head is a strict child of the current #2966 head; #3431 is stacked directly on top. The Store-only increment remains one commit, 14 files, +1,103/-28. Current-stack CUDA-on/off focused builds pass, including `nvlink_vmm_unit` 2/2, `egm_store_pool_unit` 1/1, and direct-read 3/3. The earlier GB200 report remains the composed hardware result, while exact-head GitHub Actions for the rebased stack are now running.

mooncake-transfer-engine/src/transport/nvlink_transport/nvlink_transport.cpp
```diff
@@ -737,8 +781,130 @@ int NvlinkTransport::registerLocalMemory(void *addr, size_t length,
...
+    const RemoveLocalMemoryBufferOp &remove_buffer,
+    const UpdateLocalSegmentDescOp &update_segment) {
+    CUmemGenericAllocationHandle handle;
+    CUresult result = driver_api.mem_retain_allocation_handle(&handle, addr);
+    if (result != CUDA_SUCCESS) {
+        LOG(WARNING) << "NvlinkTransport: failed to retain the exact HOST_NUMA "
+                        "allocation handle for "
+                     << addr << ": " << result;
+        return 0;
+    }
```
alogfans, line 1114: Error number should be reported in this case.

neverhook:
Review follow-up is now pushed. The Python wrapper rejects `setup()`, its
overload, and `setup_dummy()` while a native client is active or cleanup-pending;
retry is allowed only after successful `close()`. Provider/Consumer NVLink
configuration is now documented, including `MC_MS_AUTO_DISC=1` and
`MC_FORCE_MNNVL=1` for ordinary Consumers. CUDA-on/off focused tests pass, and
the new Python setup lifecycle test passes 2/2.

mooncake-transfer-engine/src/transport/nvlink_transport/nvlink_transport.cpp
neverhook: Fixed in the inherited 1/3 layer. cuMemRetainAllocationHandle failure now returns ERR_MEMORY before descriptor export or metadata publication, and the regression verifies no publication and no retained registration.

neverhook:
The full stack has been rebased onto the current `main`. This Store layer keeps
the same scope and adds only the required compatibility update for the new
`CheckRegisterMemoryParams(..., protocol)` signature. CUDA-on/off EGM pool
(1/1), direct-read (3/3), and Python setup lifecycle (2/2) focused tests pass.

The current `main` independently fails the same ARM64 Store build because two
HA sources reference missing `last_sequence_id` and `producer_view_version`
fields. That baseline failure was reproduced on a detached plain-`main`
checkout and was not absorbed into this PR; Store focused tests were run in a
detached validation checkout with only those two compile fields shimmed. A new
exact-head Actions run has been triggered.

neverhook:
The full stack has been rebased onto the current `main`, including #3515. This
Store layer is unchanged by `range-diff`. The prior HA baseline compile blocker
is gone: unmodified CUDA-on/off builds now pass directly, with EGM pool 1/1,
direct-read 3/3, and Python setup lifecycle 2/2. A new exact-head Actions run
has been triggered.

mooncake-store/src/egm_store_pool.cpp
```diff
@@ -0,0 +1,344 @@
...
+                value.end());
+    return value;
+}
+
+template <typename T>
+EgmStorePoolResult<T> Error(std::string message) {
+    return tl::make_unexpected(std::move(message));
+}
+
+EgmStorePoolResult<bool> ParseBool(const std::string& value) {
```
Aionw: Could we remove these feature-local parsing helpers and reuse the existing common implementations? `mooncake-common` already provides `TryParseBool` (with `BoolTokenSet::kTrueFalse` if we want to keep this narrow token set), `TrimAsciiWhitespace`, and `TryParseInteger<int>`; `real_client.cpp` already uses the same helpers for Store configuration. Keeping separate implementations here duplicates the full-consumption/overflow logic and gives EGM booleans different whitespace behavior from the rest of the config path. `ParseNodes` can keep its comma-list handling while delegating token trimming and integer parsing to the common helpers.

Aionw (commented):
Please reuse the existing common parsing primitives for the EGM configuration.

neverhook:
Rebased [2/3] onto the new #2966 head and adapted the Store source to the current split CMake targets (`egm_store_pool.cpp` remains client-only). The configuration parsing review is addressed with `TryParseBool(..., BoolTokenSet::kTrueFalse)`, `TrimAsciiWhitespace`, and `TryParseInteger<int>` while preserving the narrow `true/false/1/0` contract. Local CUDA-on/off EGM tests, Store/TE/master/Python builds, and Python setup lifecycle 2/2 passed. Exact-head GitHub Actions are running.

mooncake-store/src/egm_store_pool.cpp
neverhook: Addressed in the rebased Store commit. The feature-local trim/bool/from_chars helpers are removed: `enable_egm_store_pool` now uses `TryParseBool` with `BoolTokenSet::kTrueFalse`, and NUMA parsing uses `TrimAsciiWhitespace` plus `TryParseInteger<int>`. This preserves case-insensitive `true/false/1/0`, rejects broader tokens such as `yes/on/enable`, preserves explicit-list sorting/deduplication and invalid-integer rejection, and adds focused coverage for whitespace, case, empty/negative/leading-plus/partial/overflow inputs.

mooncake-store/src/client_service.cpp
```diff
@@ -1028,6 +1052,18 @@ std::optional<std::shared_ptr<Client>> Client::Create(
...
+                          std::move(labels), tenant_id, false);
+}
+
+std::optional<std::shared_ptr<Client>> Client::CreateInternal(
+    const std::string& local_hostname, const std::string& metadata_connstring,
+    const std::string& protocol, const std::optional<std::string>& device_names,
+    const std::string& master_server_entry,
+    const std::shared_ptr<TransferEngine>& transfer_engine,
+    std::map<std::string, std::string> labels, const std::string& tenant_id,
+    bool force_manual_nvlink) {
```
Aionw: Can we avoid threading EGM-specific policy through generic `Client` construction? `force_manual_nvlink` is carried through a new private `CreateInternal()` and `InitTransferEngine()` solely to disable discovery and manually install the NVLink transport, and it also requires making `RealClient` a friend. Please represent this as a generic Transfer Engine initialization option/override, or configure and inject the existing `TransferEngine` through the already-supported parameter. Manual NVLink selection should be transport initialization behavior rather than an EGM-specific boolean propagated through `Client`.

mooncake-store/src/client_service.cpp
```diff
@@ -3554,6 +3591,59 @@ tl::expected<UUID, ErrorCode> Client::MountSegmentAndGetId(
     return segment_id;
 }
 
+tl::expected<void, ErrorCode> Client::MountEgmStorePoolSegment(
```
Aionw: Could this be implemented as a generic caller-supplied-ID mount transaction instead of an EGM-specific copy of `MountSegmentAndGetId()`? The new path duplicates validation, overlap detection, `Segment` construction, TE registration, and Master publication, while changing only the UUID source and transaction ordering. That ordering is actually safer: the ordinary path registers with the TE and then returns directly if `MasterClient::MountSegment()` fails, before inserting anything into `mounted_segments_`, so the registration has no rollback record. Please factor a shared rollback-safe `MountSegmentWithId()`/internal primitive: the ordinary path can generate the UUID and EGM can supply one, with both using the same transaction and cleanup state.

Aionw (commented):
The EGM ownership model makes sense, but these Client changes introduce feature-specific escape hatches where generic initialization and mount primitives would be simpler and safer.

neverhook:
Stack rebase and review update:

- Rebased onto the current #2966 head and inherited the latest `main` Store target split/Ascend linkage changes.
- Removed the EGM-specific `force_manual_nvlink`/`CreateInternal` path. The EGM Provider now configures a call-scoped NVLink-only Transfer Engine after endpoint selection and injects it through the existing `Client::Create()` parameter.
- Ordinary Store and EGM mounts now share one private caller-supplied UUID transaction, including the same rollback record and idempotent cleanup path.
- Preserved one logical Store commit.
- Local validation passed: CUDA-on VMM 2/2 and EGM 1/1; CUDA-off Transfer Engine/Store/Master/Python binding builds, EGM 1/1, and Python setup lifecycle 2/2.

Exact-head GitHub Actions are now the remaining automated gate. The previous GB200 run remains historical supporting evidence and was not rerun for this internal review refactor.

mooncake-store/src/client_service.cpp
neverhook: Addressed. The EGM-specific `force_manual_nvlink`, `Client::CreateInternal()`, and `RealClient` friendship are removed. After the concrete local endpoint is selected, RealClient uses a private call-scoped factory to create a fresh TransferEngine, reject TENT, disable discovery, initialize metadata/RPC, and install only `nvlink`; the configured instance is then passed through the existing `Client::Create(..., transfer_engine, ...)` parameter. Ordinary setup and caller-supplied engines bypass the factory, and install/init failures do not fall back to another transport. Focused tests cover order, endpoint selection, bypass behavior, and fail-closed errors.

mooncake-store/src/client_service.cpp
neverhook: Addressed. Ordinary Store mounts now generate a UUID and delegate to the same private `MountSegmentWithId()` transaction used by EGM through an internal access helper. The shared path performs validation/overlap checks, inserts the local record before TE registration and Master publication, and compensates failures through the existing lock-held unmount helper. Cleanup failures retain the record for retry; missing Master segments and `ERR_ADDRESS_NOT_REGISTERED` remain idempotent. EGM teardown uses an internal idempotent cleanup-by-ID helper while the public missing-ID behavior is unchanged. Scripted tests cover both failure stages, original-error propagation, retained-record retry, duplicate/overlap, and idempotent cleanup.

neverhook:
Stack maintenance update:

- Rebased the complete `[1/3] -> [2/3] -> [3/3]` stack onto current `main`; this layer remains patch-equivalent by `range-diff`.
- The composed head passed diff/format/pre-commit checks, CUDA-on focused tests (`nvlink_vmm_unit` 2/2, `egm_store_pool_unit` 1/1, direct-read 3/3), and CUDA-off Store/Transfer Engine/Master/Python validation.
- Fresh exact-head GitHub Actions have been triggered and are currently queued.

GB200 was not rerun because this was a patch-equivalent baseline rebase with no functional-path change.

neverhook:
Stack update:

- Rebased [2/3] onto the reviewed #2966 head and current `main`.
- Extended the dedicated NVLink path filter with `mooncake-store/**` for the EGM unit target.
- Adapted the existing factory retry log to current `ClientAutoPortConfig`; no setup or retry behavior changed.
- CUDA-on VMM 2/2, EGM 1/1, direct-read 3/3; CUDA-off Store/TE/Master/Python builds and lifecycle 2/2 pass. Fresh exact-head Actions are queued.

neverhook:
Rebased onto #2966 and resolved the CMake conflict by keeping main’s new Store config sources alongside `egm_store_pool.cpp`. CUDA-on/off and Python setup/close tests pass. No EGM behavior changes in this rebase.

neverhook:
Rebased onto the updated #2966. No changes to the EGM pool patch; the SSD-test fix in #3870 now comes from the base. Fresh CI has been triggered.

neverhook:
Rebased this on the refreshed prerequisite and kept main’s new Store sources and shared allocation header. The EGM code did not need any further changes.

neverhook:
Rebased onto main now that #2966 is merged, so this is back to a Store-only diff. Cleanup-pending mounts are no longer eligible for heartbeat remount, and an overlap rejection won't unregister an existing buffer. I also serialized close/retry and changed final-failure retention to keep only unsafe allocations, so the Client can shut down. The added failure-path tests pass in CUDA-on and CUDA-off builds; fresh CI has been triggered.

Aionw (commented):
The EGM ownership model still makes sense. However, most of the Store-side complexity in this revision is there so that each step can be injected in unit tests; production does not need it. Two facts already in the tree make most of it unnecessary:

1. `NvlinkTransport` recognizes HOST_NUMA ranges by exact address (`FindExactOwnedRange`), so an EGM chunk can be mounted through the ordinary `Client::MountSegment(ptr, size, "nvlink")`.
2. The existing global-segment path already has a teardown order: `~Client` unmounts every entry in `mounted_segments_`, and RealClient frees the backing memory afterwards. `NvlinkHostNumaAllocation` is already an RAII owner whose destructor calls `Release()`.

Suggested shape:
- Add an `nvlink` case to `Client::InitTransferEngine`, next to ascend/ubshmem/sunrise_link, and drop the manual TE factory.
- Allocate EGM chunks as one more branch of the existing allocation loop in `setup_internal`. Hold them in `std::vector<std::unique_ptr<NvlinkHostNumaAllocation>>`, mount them through the same `MountSegment` call, and release them after `client_.reset()`.
- On release failure, leak and log, as `setup_segment_memory_must_leak_` already does.

With that, the following can be deleted: `client_internal.h`, the `SegmentMountOperations` seams, the `EgmStorePoolHooks`/allocation interface, the retryable Setup/Rollback/Teardown state machine, the cleanup-pending setup guard, and most of `egm_store_pool_test.cpp`. The Store part should shrink substantially. Details inline.

This partly revisits my earlier comments. Injecting a pre-built TE and sharing a caller-supplied-ID mount transaction ended up needing more seams than they save; a transport case in `InitTransferEngine` plus the existing `MountSegment` is simpler.

mooncake-store/src/real_client.cpp
```diff
@@ -534,6 +536,73 @@ inline bool HasMemoryReplica(const std::vector<Replica::Descriptor> &replicas) {
 
 }  // namespace
 
+tl::expected<std::shared_ptr<TransferEngine>, ErrorCode>
+RealClient::CreateManualNvlinkTransferEngine(
```
Aionw, line 540: This factory chain exists only because `Client::InitTransferEngine` has no `nvlink` case when auto-discovery is off: it falls through to `unsupported_protocol`. The chain is `CreateManualNvlinkTransferEngine`, `TransferEngineSetupOperations`, `ResolveTransferEngineForSetup`, `TransferEngineFactory`, the `setup_internal_impl` split, and the extra branch in the auto-port retry loop.

Could we add `nvlink` to the existing `installTransport(protocol)` branch next to ascend/ubshmem/sunrise_link instead? RealClient could then keep calling `Client::Create(..., transfer_engine = nullptr, ...)`, and the TENT restriction becomes a single `isUsingTent()` check on the EGM path.

mooncake-store/include/real_client.h
```diff
@@ -1227,6 +1232,47 @@ class RealClient : public PyClient {
         const std::string &object_key, ErrorCode error,
         std::vector<int> &results);
 
+    friend class EgmStorePoolTestPeer;
```
Aionw, line 1235: `friend class EgmStorePoolTestPeer` and `TransferEngineSetupOperations` are test-only seams in a public header; `TransferEngineSetupOperations` holds four `std::function`s that production always leaves null. Both go away with the `InitTransferEngine` change suggested in `real_client.cpp`. After that, the header only needs a member that owns the EGM allocations.

mooncake-store/src/client_internal.h
```diff
@@ -0,0 +1,44 @@
...
+
+#pragma once
+
+#include <functional>
+
+#include "client_service.h"
+
+namespace mooncake::internal {
+
+struct SegmentMountOperations {
```
Aionw, line 11: Could this file be dropped? `SegmentMountOperations`, `ClientAccess`, the friend declaration, and the `operations = nullptr` parameters on `Client`'s private mount/unmount helpers all exist so unit tests can replace register/mount/unmount.

From Store's point of view an EGM chunk is an ordinary nvlink segment, because the transport identifies HOST_NUMA ranges by exact address. So `Client::MountSegment()` is sufficient, and no caller-supplied UUID is needed.

mooncake-store/src/client_service.cpp
```diff
@@ -3796,15 +3810,29 @@ tl::expected<void, ErrorCode> Client::UnmountSegment(const void* buffer,
...
     const void* buffer, size_t size, const std::string& protocol,
     const std::string& location) {
+    UUID segment_id = generate_uuid();
+    auto result =
+        MountSegmentWithId(segment_id, buffer, size, protocol, location);
+    if (!result) return tl::unexpected(result.error());
+    return segment_id;
+}
+
+tl::expected<void, ErrorCode> Client::MountSegmentWithId(
```
Aionw, line 3820: Revisiting my earlier comment here: once EGM mounts through the ordinary `MountSegment()` (see `client_internal.h`), this feature no longer needs a caller-supplied ID. That also means it no longer needs the staged transaction built around it (`cleanup_pending_segments_`, `check_overlap`, TENT special cases, remount filtering).

mooncake-store/src/client_service.cpp
```diff
@@ -3841,20 +3861,103 @@ tl::expected<UUID, ErrorCode> Client::MountSegmentAndGetId(
...
+                           << " cleanup_error=" << cleanup.error();
+            }
             return tl::unexpected(err);
         }
+        cleanup_pending_segments_.erase(segment_id);
+    }
 
-        segment_id = segment.id;
-        mounted_segments_[segment_id] = segment;
+    if (operations == nullptr) {
```
Aionw, line 3937: `if (operations == nullptr) EnsureStorageControlPlaneStarted();` makes production behavior depend on whether a test fake is installed. It goes away once `operations` is removed.

mooncake-store/src/egm_store_pool.h
```diff
@@ -0,0 +1,122 @@
...
+EgmStorePoolResult<void> ValidateEgmStorePoolOptions(
+    const EgmStorePoolOptions& options, const std::string& protocol,
+    size_t global_segment_size, size_t local_buffer_size);
+
+EgmStorePoolResult<EgmStorePoolPlan> PlanEgmStorePool(
+    size_t requested_bytes,
+    const std::vector<std::pair<int, size_t>>& node_granularities,
+    size_t max_mr_size, size_t store_alignment = 0);
+
+class EgmStorePoolAllocation {
```
Aionw, line 59: `EgmStorePoolAllocation`, the `NvlinkAllocation` adapter, `EgmStorePoolAllocationAttempt`, `EgmStorePoolHooks`, and `MakeNvlinkHostNumaHooks` have a single production implementation; they exist so that `FakeAllocation` can be plugged in.

`NvlinkHostNumaAllocation` is already the RAII owner: its destructor calls `Release()`, and `Release()` does not repeat completed stages. Holding `std::vector<std::unique_ptr<NvlinkHostNumaAllocation>>` directly in RealClient should be enough.

mooncake-store/src/egm_store_pool.cpp
```diff
@@ -0,0 +1,336 @@
...
+        }
+        return attempt;
+    };
+    return hooks;
+}
+
+EgmStorePool::EgmStorePool(EgmStorePoolHooks hooks)
+    : hooks_(std::move(hooks)) {}
+
+EgmStorePoolResult<void> EgmStorePool::Setup(const EgmStorePoolOptions& options,
```
Aionw, line 222: Could this Setup/Rollback/Teardown transaction reuse the existing global-segment loop in `RealClient::setup_internal` instead? That path has no staged rollback: `client_.reset()` makes `~Client` unmount every mounted segment, and the backing memory is freed afterwards.

EGM fits the same order as one more branch of the allocation if/else chain:
1. Split the capacity across the NUMA nodes.
2. Chunk each node's share with `GetNextSegmentSize`. If nvlink must be split by `max_mr_size`, add it to `GetTransportRegistrationLimit`.
3. Create each chunk with `NvlinkHostNumaAllocation::Create(node, size, granularity)`.
4. Mount it through the same `MountSegment` call.

`PlanEgmStorePool`, `CheckedLcm`, and `EgmStorePoolPlan` then reduce to aligning each chunk to the granularity.

mooncake-store/src/egm_store_pool.h
```diff
@@ -0,0 +1,122 @@
...
+#include <vector>
+
+#include <ylt/util/tl/expected.hpp>
+
+#include "types.h"
+
+namespace mooncake {
+
+template <typename T>
+using EgmStorePoolResult = tl::expected<T, std::string>;
```
Aionw, line 20: Nit: the rest of Store returns `ErrorCode` and logs at the failure site. Here the error is a `std::string` that RealClient then collapses into `INTERNAL_ERROR`, so the text is only ever logged. Could we follow the Store convention?

mooncake-store/src/egm_store_pool.cpp
```diff
@@ -0,0 +1,336 @@
...
+        return options;
+    }
+    auto parsed = ParseNodes(nodes->second);
+    if (!parsed) return Error<EgmStorePoolOptions>(parsed.error());
+    options.auto_nodes = false;
+    options.nodes = std::move(*parsed);
+    return options;
+}
+
+EgmStorePoolResult<void> ValidateEgmStorePoolOptions(
```
Aionw, line 112: `ValidateEgmStorePoolOptions` runs once in `setup_internal(ConfigDict)` and again inside `EgmStorePool::Setup`. That second call is the only reason `Setup` takes `protocol` and `local_buffer_size`. Validating once at the config boundary should be enough.

mooncake-store/src/real_client.cpp
```diff
@@ -1414,11 +1584,36 @@ tl::expected<void, ErrorCode> RealClient::setup_internal(
...
+    if (egm_options->enabled) {
+        transfer_engine_factory = [metadata_server](
+                                      const std::string &endpoint) {
+            return CreateManualNvlinkTransferEngine(endpoint, metadata_server);
+        };
+    }
+
+    auto setup = setup_internal_impl(
+        local_hostname, metadata_server,
+        egm_options->enabled ? 0 : global_segment_size, local_buffer_size,
```
Aionw, line 1611: Running the whole setup with `global_segment_size = 0` and mounting EGM afterwards changes the ordering relative to the normal path: the IPC and HTTP servers are already started before any capacity is published. Also, if the EGM mount fails, a live `client_` and those servers are left behind, and the cleanup-pending guard then has to handle that.

Allocating EGM inside the normal segment loop avoids both problems. The EGM options could be stored as a member, like `use_hugepage_`, so the long `setup_internal` signature does not need to change.

mooncake-store/src/real_client.cpp
```diff
@@ -1335,8 +1445,68 @@ inline std::optional<int> get_config_int(const ConfigDict &config,
...
+
+    auto pool = std::make_unique<EgmStorePool>(
+        MakeNvlinkHostNumaHooks(std::move(hooks)));
+    auto setup = pool->Setup(options, protocol, global_segment_size, 0,
+                             static_cast<size_t>(globalConfig().max_mr_size));
+    if (!setup) {
+        // The underlying Client is already initialized even if the pool
+        // rolled back completely. Retain the pool marker so another setup is
+        // rejected until close() tears down that Client.
+        egm_store_pool_ = std::move(pool);
```
Aionw, line 1483: The whole cleanup-pending lifecycle follows from letting Teardown fail partway and making it retryable:
- the pool is kept as a marker after a failed setup (here);
- re-setup is rejected (L1505);
- the destructor calls `RetainAllocationsForProcessLifetime`;
- `close()` and `ensure_can_setup` change in `store_py.cpp`;
- `test_store_setup_lifecycle.py` is added to cover it.

The existing segment path handles a release failure by leaking and logging (`setup_segment_memory_must_leak_`). Could EGM do the same? Then none of the above is needed.

mooncake-store/tests/egm_store_pool_test.cpp
```diff
@@ -0,0 +1,1147 @@
...
+    std::vector<std::string> events;
+    AutoDiscoverConfig auto_discover;
+    std::string metadata_server;
+    std::string local_endpoint;
+    std::string hostname;
+    std::string installed_protocol;
+    uint16_t port = 0;
+};
+
+class EgmStorePoolTestPeer {
```
Aionw, line 35: Most of these 30 cases test the seams themselves: the factory call order, the three branches of `ResolveTransferEngine`, the `ScriptedMountOperations` event sequences, and the `FakeOperations` hook orchestration.

Once the seams are removed, I'd keep the config-parsing tests (plus the chunk math, if it stays a pure function) and leave allocation, mount, and release to an integration test on MNNVL hardware. Staged `Release()` retry is already covered by `nvlink_host_numa_allocation_test` in [1/3].

mooncake-store/tests/CMakeLists.txt
```diff
@@ -47,6 +47,10 @@ add_store_test(segment_pool_test segment_pool_test.cpp)
 add_store_test(runtime_accelerator_test runtime_accelerator_test.cpp)
 add_store_test(batch_read_fanout_test batch_read_fanout_test.cpp)
 add_store_test(registered_pinned_memory_test registered_pinned_memory_test.cpp)
+add_store_test(egm_store_pool_test egm_store_pool_test.cpp)
+target_include_directories(egm_store_pool_test
```
Aionw, line 51: Needing `../src` on the include path shows that the test depends on internal headers (`client_internal.h`, `egm_store_pool.h`). After the simplification, this line and the `egm_store_pool_unit` label can go.

mooncake-store/src/real_client.cpp
neverhook: Agreed. NVLink now uses the normal manual-install branch, and the factory chain and setup overload are gone. The discovery environment override still works; before allocating EGM, RealClient checks that the actual engine is non-TENT and has NVLink installed. This also covers the automatic-port path.

mooncake-store/src/client_internal.h
neverhook: Removed. EGM now keeps native HOST_NUMA owners and calls ordinary MountSegment, with no ClientAccess, caller-supplied UUID or scripted mount operations. The shared mount/heartbeat code is back to main's implementation. Generic failed-mount cleanup remains separate; #2336 already covers that area.

mooncake-store/src/egm_store_pool.cpp
neverhook: The EGM branch now lives in the existing allocation loop and uses GetNextSegmentSize on aligned per-node shares. The Pool/Plan objects and LCM helper are gone. I kept max_mr_size local to EGM: adding NVLink to GetTransportRegistrationLimit would also change ordinary NVLink registration and dynamic-mount behavior, which this pool doesn't need. Leaving this thread open for your confirmation of that narrower choice.

mooncake-store/src/egm_store_pool.h
neverhook: Done. The small config/capacity helpers return ErrorCode and log where they fail. Config parsing and its cross-field validation run once at the dictionary boundary; there is no second pool-level validation. The accepted boolean and NUMA tokens are unchanged.

mooncake-store/src/real_client.cpp
neverhook: EGM allocation and mounting now finish in the normal global-segment loop, before IPC, HTTP or offload services start. The parsed NUMA option is consumed once at the positional setup entry, without extending that signature. Errors and exceptions from an EGM setup run terminal cleanup before returning or rethrowing.

mooncake-store/src/real_client.cpp
neverhook: The retry machinery and global setup guard are gone. Close is now terminal: it reports a cleanup error, clears both Python native references and continues closing other resources. Repeated setup semantics are tracked separately in #4294.

I kept an explicit EGM-only unmount before releasing each owner. Client destruction logs unmount failures but cannot report them back, so relying on destruction alone could free memory after an unconfirmed Master cleanup. Failed or ambiguous mounts and failed unmounts retain their allocation until process exit; CUDA release failures do the same. FileStorage workers are stopped first, since they hold a Client reference and may access those segments. There is no retry queue or background recovery.

Leaving this open for your review of those local safety checks.

mooncake-store/tests/egm_store_pool_test.cpp
neverhook: Replaced the 30 orchestration cases with 13 config/capacity cases. The existing real Client tests now also check discovery overrides and EGM rejection of a non-NVLink backend, for both explicit and automatic ports. The special include path and EGM CTest label are gone. Allocation/mount/release hardware testing remains separate; no old GB200 result is being counted for this implementation.

neverhook:
@Aionw, I've reworked this around the existing Store path. The earlier version had grown a separate lifecycle around EGM, with interfaces used mainly by the tests. I agree that was too much machinery for adding another kind of segment backing.

NVLink now uses normal transport initialization. EGM allocation and mounting run in the existing segment loop, before the IPC/HTTP servers start, and RealClient owns the native HOST_NUMA allocations directly. The factory, allocation adapter, hooks, caller-supplied UUID transaction and retryable-close state machine are gone.

I kept the `max_mr_size` cap local to EGM. Moving it into the generic NVLink path would also change ordinary registration and dynamic mounts, which this pool doesn't need.

Close still checks each EGM unmount result: `Client` destruction only logs failures, so it cannot tell the allocation owner whether Master cleanup succeeded. FileStorage workers stop first because they can still access those segments. A failed or uncertain mount/unmount retains the allocation until process exit; a CUDA release failure retains whatever resources remain. Close reports the error and finishes without keeping a client for another cleanup attempt. This trades reclamation after an error for a simpler lifecycle. It still requires callers to quiesce in-flight requests before close and does not promise immediate remote cleanup.

Generic mount/MR rollback remains outside this PR; related work is in the separate, still-open #2336. The Python repeated-setup guard is also removed, with that API decision tracked in #4294. Tests now focus on config/capacity rules and real Client initialization.

CI is green, including the full CTest sweep and docs build. This revision has not been rerun on GB200; the earlier hardware results aren't validation of the new cleanup behavior.

Could you take another look, particularly at the EGM-only cap and explicit unmount checks? I've left those two inline discussions open for your confirmation.

mooncake-store/src/real_client.cpp
```diff
@@ -851,6 +854,7 @@ tl::expected<void, ErrorCode> RealClient::setup_internal(
     bool enable_ssd_offload, bool start_offload_rpc_server,
     const std::string &ssd_offload_path, const std::string &tenant_id,
     bool enable_client_http_server, int client_http_port) {
+    auto egm_numa_nodes = std::exchange(pending_egm_numa_nodes_, std::nullopt);
```
Aionw: Whether EGM is enabled and which NUMA nodes it uses should be controlled explicitly by the configuration. `ParseEgmStorePoolConfig` already produces an explicit `enabled` flag and node list. Here, though, that config is collapsed into a `std::optional<std::vector<int>>` member:

- "EGM is enabled" becomes "the member has a value".
- "Auto-discover nodes" becomes "the vector is empty", and discovery only runs deep inside the segment loop.

Neither decision is visible in `setup_internal`'s signature or at its call site, so the EGM path is driven by whatever happens to be in the member rather than by the configuration.

Could we pass the parsed `EgmStorePoolConfig` into `setup_internal` explicitly and branch on `config.enabled`? Node selection could also be an explicit mode in the config, for example an `auto_numa_nodes` flag, instead of an empty-vector sentinel. That way, how the nodes are initialized follows directly from what the user configured.

mooncake-store/src/real_client.cpp
```diff
@@ -1147,9 +1225,28 @@ tl::expected<void, ErrorCode> RealClient::setup_internal(
 
             auto pinned_region =
                 TryPinStoreSegment(ptr, mapped_size, this->protocol, "setup");
-            auto mount_result =
-                client_->MountSegment(ptr, mount_size, protocol, seg_location);
+            tl::expected<void, ErrorCode> mount_result;
+            try {
```
Aionw: Is this try/catch here because only the EGM path needs to handle a throw? For non-EGM segments it just rethrows. The EGM handling in the catch also looks redundant to me:

- `MountSegmentAndGetId` inserts into `mounted_segments_` only after the Master mount succeeds. If the mount fails or throws before that point, the Client has no record for this range.
- On failure, the ConfigDict path calls `tearDownAll_internal()`. There, `UnmountSegment(base, len)` fails because the record is missing, and teardown already responds by releasing the owner so the memory is retained until the process exits. That is the same outcome as this catch block.
- The one case where the outcomes differ is when the record was already inserted and `EnsureStorageControlPlaneStarted()` then throws. Teardown could unmount and free that memory safely, but this catch leaks it.

Could we drop this try/catch, along with the EGM-specific release/pop in the `!mount_result` branch below, and rely on teardown's existing rule that a failed unmount retains the allocation? A short comment in teardown noting that failed mounts have no Client record would make the leak path explicit. The `catch (...) { tearDownAll_internal(); throw; }` in the ConfigDict overload (L1535) could go too, keeping only the `!result` teardown.

mooncake-store/src/real_client.cpp
neverhook: The parsed config is now passed directly into a private `setup_internal` overload; positional setup passes the default-disabled config. Branches use `config.enabled`, and `auto_numa_nodes` selects discovery explicitly. The pending member and empty-vector mode are gone, without changing the public setup signatures.

The config tests check the default, auto and explicit modes. The real Client tests also check that each mode rejects a non-NVLink backend before allocation, with either explicit or automatic ports, and that disabled EGM ignores the node setting.

mooncake-store/src/real_client.cpp
neverhook: Removed the inner catch and the early `release()/pop_back()` on mount failure. The owner stays in the cleanup list, so teardown can reclaim it when a Client record exists. A missing record or failed unmount still retains the allocation; there's a short comment at that point.

I kept the outer ConfigDict catch. `FileStorage` construction can throw after EGM segments have been mounted, while the Python wrapper still owns the native client. That catch runs the same terminal teardown and rethrows; without it, those resources stay live until the caller closes or destroys the wrapper. This keeps exception cleanup immediate without adding retry state. Does that boundary look reasonable to you?

PR #3342, opened 2026-08-09
[Bugfix] use BufferPool for zcopy benchmark

### LINKED ISSUE #3329: [Bug]: store_kv_bench --io-api zcopy always fails on non-NOF builds: "direct hugepage alloc failed for zcopy pool"
### Bug Report

**Environment**
- Mooncake commit: `a2966b6a` 
- Build: default CMake (`USE_NOF=OFF`), store wheel built without SPDK/NoF
- Transport: TCP
- Store API: `--io-api zcopy`

**Summary**
Running `mooncake-store/benchmarks/store_kv_bench.py` with `--io-api zcopy`
always fails during session setup on a memory-only (non-NOF) build:

```
RuntimeError: direct hugepage alloc failed for zcopy pool: size=16384
```

The benchmark's zcopy path is hard-coupled to the NOF build option, even
though the zero-copy store APIs themselves (`put_from` / `get_into` /
`batch_put_from` / `batch_get_into`) have nothing to do with NOF and work
fine on memory-only clusters.

**Steps to reproduce**
Memory-only cluster (1+0), no `--nof-replica-num`:

```bash
python3 mooncake-store/benchmarks/store_kv_bench.py \
  --scenario mixed_rw \
  --prepare-mode auto \
  --io-api zcopy \
  --local-hostname <host>:50071 \
  --metadata-server <metadata-url> \
  --master-server <master:50051> \
  --protocol tcp \
  --nr-objects 64 --write-objects 4096 --batch-size 4 --runtime 10 \
  --rwmixread 70 --key-prefix mixed --key-size 20 --value-size 4096 \
  --memory-replica-num 1 --verify --pattern 0x5a
```

Note the failure is unrelated to replica count — it triggers before any I/O.

**Actual behavior**
```
File "store_kv_bench.py", line 488, in __init__
    self.zcopy_pool = ZcopyBufferPool(...)
File "store_kv_bench.py", line 396, in __init__
    raise RuntimeError(
RuntimeError: direct hugepage alloc failed for zcopy pool: size=16384
```

**Expected behavior**
`--io-api zcopy` should work on non-NOF (memory-only) builds. The zero-copy
APIs are orthogonal to NOF.

**Root cause**
`ZcopyBufferPool` in `mooncake-store/benchmarks/store_kv_bench.py` allocates
its registered buffer pool via `get_alloc_func_addr()` /
`get_free_func_addr()` (exported in `mooncake-integration/store/store_py.cpp`),
which point to `hugepage_memory_alloc` / `hugepage_memory_free` in
`mooncake-store/src/memory_alloc.cpp`:

```cpp
void *hugepage_memory_alloc(size_t size) {
#ifndef USE_NOF
    return nullptr;   // non-NOF builds always return null
#else
    return mooncake::SpdkWrapper::GetInstance().Alloc(size, 0x1000, -1);
#endif
}
```

`hugepage_memory_alloc` only has a real implementation when the store is built
with `USE_NOF=ON` (default `OFF`); otherwise it always returns `nullptr`, so
the zcopy pool construction always fails. This helper was originally added for
the SPDK/NoF worker pool (#2172) and later reused by the benchmark's zcopy
path (#2273), creating an unintended coupling between `--io-api zcopy` and the
NOF build option.

### Before submitting...

030611:
## Description

Fixes #3329.

`store_kv_bench --io-api zcopy` used the NoF/SPDK-only hugepage allocation helpers. In `USE_NOF=OFF` builds those helpers return `nullptr`, so the benchmark failed before any Store I/O.

This change:
- allocates the benchmark zcopy region with the native `BufferPool` / `BufferLease` API; this prefers the setup-time local buffer and preserves the native short-lived registered overflow fallback when local allocation is temporarily unavailable;
- keeps the lease alive for the benchmark lifetime and releases it before closing the pool;
- treats `--local-buffer-size` as a benchmark sizing policy and fails early when `value_size * slots` exceeds it;
- adds CPU-only regression coverage for the non-NoF module surface, pointer I/O, cleanup, capacity guard, and acquire failure.

Full TCP + `USE_NOF=OFF` Store integration was not run locally because the built Linux Mooncake Python extension and services are unavailable.

## Module

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

**Test commands:**
```bash
python -m unittest mooncake-store/benchmarks/test_store_kv_bench.py -v
python -m py_compile mooncake-store/benchmarks/store_kv_bench.py mooncake-store/benchmarks/test_store_kv_bench.py
uvx --from ruff==0.6.9 ruff check mooncake-store/benchmarks/store_kv_bench.py mooncake-store/benchmarks/test_store_kv_bench.py
uvx --from ruff==0.6.9 ruff format --check mooncake-store/benchmarks/store_kv_bench.py mooncake-store/benchmarks/test_store_kv_bench.py
uvx --from codespell==2.2.6 codespell mooncake-store/benchmarks/store_kv_bench.py mooncake-store/benchmarks/test_store_kv_bench.py
git diff --check
```

**Test results:**
- [x] Unit tests pass (8/8)
- [ ] Integration tests pass (not run; see limitation above)
- [ ] Manual testing done

## Checklist

- [ ] I have formatted my code using `./scripts/code_format.sh` (Python-only change; pinned Ruff format check passed)
- [ ] I have run `pre-commit run --all-files` and all hooks pass (hook bootstrap timed out; the relevant pinned Ruff, formatting, codespell, and file-hygiene checks passed directly)
- [x] For changes >500 LOC: I have filed an RFC issue (not applicable; 111 additions / 59 deletions)

## AI Assistance Disclosure

- [ ] No AI tools were used

AI assistance was used for code review, test design, and drafting. The submitted changes and tests were reviewed locally, and the commands/results above were verified.


zxpdemonio:
The BufferPool migration itself looks correct. One semantic clarification: checking total_size against local_buffer_size and passing max_bytes=local_buffer_size does not make the allocation local-buffer-only. The native BufferPool first tries the setup-time local buffer, but it can still fall back to a short-lived registered overflow buffer when local allocation is temporarily unavailable. I think keeping the existing BufferPool overflow behavior is reasonable here. Please update the PR description to say that the benchmark prefers the local buffer and may use the native overflow fallback, and remove the claim that this capacity guard avoids depending on overflow registration. The guard can remain as a benchmark sizing policy.

030611:
Thanks for the clarification. I updated the description to state that BufferPool prefers the setup-time local buffer but may use its native short-lived registered overflow fallback, and that the capacity check is only a benchmark sizing policy. No code change was needed.

fcczzz (approved):
LGTM, thanks for the fix!
Non-blocking: could BufferPool be resolved only for zcopy mode? Accessing it at import time may break plain mode with an older or custom store module that does not expose it.

030611:
Good point. Updated in `c2cb754` so `BufferPool` is resolved only when `ZcopyBufferPool` is constructed; importing or running plain mode no longer requires the store module to expose it. I also added an import regression test using a store module without `BufferPool` (`9 passed`).

ykwd (approved):

PR #3348, opened 2026-08-10
[Store] Support configurable fileread worker pool size

NUABO:
## Description
Make the number of FilereadWorkerPool worker threads configurable via the MC_FILEREAD_WORKERS environment variable, matching the existing MC_NOF_WORKERS pattern. Falls back to the previous default (10) when the variable is unset, empty, or invalid, so behavior is unchanged by default.

Increasing the worker pool size allows more file read operations to be processed concurrently, which can significantly improve throughput in I/O-intensive workloads, especially on systems with fast NVMe storage devices. In scenarios with high request concurrency, a larger thread pool reduces queueing delays and better utilizes the underlying hardware parallelism.

However, the optimal value depends on the workload and system characteristics. Users are advised to tune this parameter based on their specific deployment environment, monitoring thread utilization and system load to avoid excessive context switching or resource contention that might degrade performance.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
```

**Test results:**
- [ ] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

### DIFF
--- a/mooncake-store/src/transfer_task.cpp
+++ b/mooncake-store/src/transfer_task.cpp
@@ -19,6 +19,26 @@
 #include "spdk/spdk_wrapper.h"
 #endif
 
+static int GetPositiveEnvOrDefault(const char* name, int default_value) {
+    const char* raw_value = std::getenv(name);
+    if (!raw_value || raw_value[0] == '\0') {
+        return default_value;
+    }
+
+    errno = 0;
+    char* end_ptr = nullptr;
+    long parsed = std::strtol(raw_value, &end_ptr, 10);
+    if (errno != 0 || end_ptr == raw_value ||
+        (end_ptr != nullptr && *end_ptr != '\0') || parsed <= 0 ||
+        parsed > std::numeric_limits<int>::max()) {
+        LOG(WARNING) << "Invalid value for " << name << ": " << raw_value
+                     << ", using default " << default_value;
+        return default_value;
+    }
+
+    return static_cast<int>(parsed);
+}
+
 #ifdef USE_NOF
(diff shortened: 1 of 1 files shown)


Aionw (approved):
LGTM

fcczzz (changes requested):
Thanks for the change! One build-configuration issue I noticed: `GetPositiveEnvOrDefault()` is currently defined inside `#ifdef USE_NOF`, while `GetFilereadWorkerCount()` calls it unconditionally. It looks like this causes the non-NoF build to fail at `transfer_task.cpp:186`, as shown in the current CI run.  
Would you mind moving the generic environment helper outside the `USE_NOF` guard and rerunning the non-NoF build?

NUABO:
> Thanks for the change! One build-configuration issue I noticed: `GetPositiveEnvOrDefault()` is currently defined inside `#ifdef USE_NOF`, while `GetFilereadWorkerCount()` calls it unconditionally. It looks like this causes the non-NoF build to fail at `transfer_task.cpp:186`, as shown in the current CI run. Would you mind moving the generic environment helper outside the `USE_NOF` guard and rerunning the non-NoF build?

Thanks for your review. I also noticed this issue and will update it as soon as possible.

fcczzz (approved):
Thanks for the quick update! LGTM!

ykwd (approved):

PR #3351, opened 2026-08-10
[TransferEngine] Fix graceful shutdown test hangs in USE_ETCD builds

### LINKED ISSUE #3349: [Bug]: graceful_shutdown_test can hang after fork in USE_ETCD CI builds
## Summary

`GracefulShutdownTest.SigtermTriggersCleanExit` intermittently hangs in the Linux CI build and is killed by the test after a five-second timeout.

This failure is not specific to the PR on which it appears. The same signature has occurred on `main` and on multiple unrelated PRs.

The current evidence points to a bug in the test process model:

1. With `USE_ETCD=ON`, `graceful_shutdown_test` loads the Go c-shared etcd wrapper, whose runtime starts background threads before the tests run.
2. The test calls `fork()` from this already-multithreaded process.
3. The child then constructs `TransferEngine`, takes C++ locks, allocates memory, initializes logging/runtime state, and creates a `std::thread`.
4. These operations are not safe between `fork()` and `exec()` in the child of a multithreaded process.

The fixed 100 ms sleep used by the parent is a separate synchronization bug that makes the failure more likely under CI load.

## Observed failure

Example job:

https://github.com/kvcache-ai/Mooncake/actions/runs/31331251297/job/93289807744?pr=3344

```text
[ RUN      ] GracefulShutdownTest.SigtermTriggersCleanExit
WARNING: Logging before InitGoogleLogging() is written to STDERR
I0809 19:41:01.889427 17248 transfer_engine_impl.cpp:990] Metrics reporting is disabled (set MC_TE_METRIC=1 to enable)
/home/runner/work/Mooncake/Mooncake/mooncake-transfer-engine/tests/graceful_shutdown_test.cpp:37: Failure
Failed
child did not exit before timeout

/home/runner/work/Mooncake/Mooncake/mooncake-transfer-engine/tests/graceful_shutdown_test.cpp:58: Failure
Value of: (((status) & 0x7f) == 0)
  Actual: false
Expected: true
Child did not exit normally (signaled: true)

[  FAILED  ] GracefulShutdownTest.SigtermTriggersCleanExit (5111 ms)
```

The `WIFEXITED(status)` failure is secondary. After waiting for five seconds, `waitChildWithTimeout()` sends `SIGKILL` to the child, so the resulting status is necessarily signaled rather than normally exited.

## Expected behavior

The test child should:

1. Finish constructing the requested `TransferEngine` state.
2. Install the graceful shutdown handlers.
3. Report that it is ready.
4. Receive SIGTERM or SIGINT.
5. Exit normally with status `128 + signo`.

The test must also terminate within a bounded time when child setup fails.

## Affected CI runs

I found the same failure signature in at least six independent workflow runs:

- https://github.com/kvcache-ai/Mooncake/actions/runs/31331251297
- https://github.com/kvcache-ai/Mooncake/actions/runs/31263153332
- https://github.com/kvcache-ai/Mooncake/actions/runs/31169844813
- https://github.com/kvcache-ai/Mooncake/actions/runs/31159834899
- https://github.com/kvcache-ai/Mooncake/actions/runs/31086084818
- https://github.com/kvcache-ai/Mooncake/actions/runs/31082176523

These runs include both `main` and unrelated PRs. One occurrence is in a Python 3.10 job, so this does not appear to be specific to Python 3.12.

The example PR, #3344, only removes unused declarations from:

```text
mooncake-store/include/master_metric_manager.h
```

It does not modify Transfer Engine, graceful shutdown, signal handling, or this test.

## Relevant CI configuration

The main Linux CI config enables the Go etcd wrapper and ASan:

```bash
cmake -G Ninja .. \
  -DUSE_HTTP=ON \
  -DUSE_CXL=ON \
  -DUSE_UB=ON \
  -DUSE_ETCD=ON \
  -DSTORE_USE_ETCD=ON \
  -DENABLE_ASAN=ON \
  -DENABLE_SCCACHE=ON \
  -DCMAKE_BUILD_TYPE=Debug \
  -DENABLE_DEBUG_SYMBOLS=OFF
```

The test is then run as part of:

```bash
MC_METADATA_SERVER=http://127.0.0.1:8080/metadata \
DEFAULT_KV_LEASE_TTL=500 \
ctest -j --output-on-failure
```

## Local reproduction

Using a build matching the important CI options, especially `USE_ETCD=ON` and `ENABLE_ASAN=ON`, I stress-ran the failing test concurrently:

```bash
seq 1 500 | xargs -P32 -I{} \
  timeout 10s \
  ./build/mooncake-transfer-engine/tests/graceful_shutdown_test \
  --gtest_filter=GracefulShutdownTest.SigtermTriggersCleanExit
```

In one captured run, 8 out of 500 invocations reproduced the same approximately 5.1-second timeout.

With a minimal build using `USE_ETCD=OFF`, 1,000 concurrent invocations completed without this timeout.

The exact rate is scheduling-dependent, but enabling the Go c-shared etcd wrapper makes the original CI failure reproducible.

### 1. The test forks from a multithreaded process

The test currently does the following:

```cpp
pid_t pid = fork();

if (pid == 0) {
    auto engine = std::make_unique<TransferEngine>(false);
    engine->enableGracefulShutdown();
    pause();
    _exit(99);
}
```

With `USE_ETCD=ON`, `transfer_engine` links:

```text
mooncake-common/etcd/libetcd_wrapper.so
```

This library is built with:

```text
go build -buildmode=c-shared
```

Tracing the CI-equivalent test binary shows that the Go runtime has created roughly 60 threads before GTest reaches `fork()`.

After `fork()`, only the calling thread exists in the child. The child nevertheless performs substantial non-async-signal-safe work, including:

- constructing `TransferEngineImpl`;
- allocating `shared_ptr` and shutdown token state;
- acquiring `g_registry_mutex` and `g_install_mutex`;
- interacting with logging and sanitizer runtime state;
- creating and detaching the signal watcher `std::thread`.

POSIX requires the child of a multithreaded process to execute only async-signal-safe operations until a successful `exec()`:

https://pubs.opengroup.org/onlinepubs/9799919799/functions/fork.html

The test therefore relies on undefined/unsupported post-fork behavior.

The Go wrapper is the trigger that makes this easy to reproduce, but this should not be treated as an etcd or Go runtime bug. Any background thread holding runtime, allocator, logging, or library state at the time of `fork()` can expose the same test defect.

The exact runtime-internal lock or loop where every failed child stalls has not been reduced to one symbol. However, tracing confirms that failed children remain running or stuck before completing their ready state, and the test already violates the required process contract independently of that internal detail.

### 2. A fixed sleep is used as readiness synchronization

The parent currently does:

```cpp
usleep(100000);
kill(pid, SIGTERM);
```

The 100 ms delay does not prove that the child has completed either:

```cpp
std::make_unique<TransferEngine>(false);
```

or:

```cpp
engine->enableGracefulShutdown();
```

Under ASan, Go runtime initialization, and concurrent CI load, the parent can send SIGTERM while the child is still initializing.

In a simpler single-threaded build, this race can appear as an immediate signal termination before the Mooncake handler is installed. In the `USE_ETCD=ON` build, it can interact with the invalid post-fork runtime state and produce the observed hang.

### #3271 and #3278

Issue #3271 and PR #3278 addressed a different production bug:

- the shutdown watcher thread could receive SIGTERM/SIGINT itself;
- the signal handler would write to the pipe and then pause on the only reader thread;
- no thread remained to consume the pipe byte and perform cleanup.

PR #3278 fixed that by blocking SIGTERM/SIGINT while creating the watcher so the watcher inherits the blocked mask:

https://github.com/kvcache-ai/Mooncake/pull/3278

The failing CI commits already contain that fix. The current failure therefore does not indicate that #3278 is missing or ineffective for its intended case.

### #3274

PR #3274 replaces the fixed 100 ms sleep with a readiness pipe:

https://github.com/kvcache-ai/Mooncake/pull/3274

That addresses the readiness race, but it does not address the multithreaded `fork()` problem. The child still constructs `TransferEngine` and starts the watcher before `exec()`.

I applied the #3274 patch to the CI-equivalent build and stress-ran it. In one 500-run batch, six children hung before sending the readiness byte.

The current #3274 parent also uses an unbounded blocking `read()`. If the child never reaches the readiness notification, the parent can wait indefinitely instead of failing after five seconds.

For these reasons, #3274 is not sufficient by itself.

## Proposed fix

Run the signal target in a fresh process image instead of constructing `TransferEngine` directly in the post-fork child.

A minimal implementation can keep everything in the existing test source:

1. Give `graceful_shutdown_test` a custom `main()` instead of linking `gtest_main`.
2. Add an internal helper mode, for example:

   ```text
   graceful_shutdown_test --shutdown-child single <ready-fd>
   graceful_shutdown_test --shutdown-child destroyed <ready-fd>
   graceful_shutdown_test --shutdown-child multiple <ready-fd>
   ```

3. From the GTest parent, create a readiness pipe and use `posix_spawn()` to execute `/proc/self/exe` in helper mode.
4. In helper mode, after the new process image has initialized normally:
   - construct the requested engine state;
   - call `enableGracefulShutdown()`;
   - write one readiness byte;
   - wait for SIGTERM or SIGINT.
5. In the parent, use `poll()` with a finite timeout to wait for readiness.
6. After readiness, send the signal and wait for the expected exit status.
7. On every error or timeout path, close the pipe, kill the child if needed, and reap it with `waitpid()`.

The process flow becomes:

```text
multithreaded GTest parent
        |
        | posix_spawn("/proc/self/exe", helper mode)
        v
fresh process image and runtime initialization
        |
        | construct TransferEngine
        | enableGracefulShutdown()
        | write readiness byte
        v
receive SIGTERM/SIGINT
        |
        v
cleanup and _Exit(128 + signo)
```

The following existing cases can share this helper:

- `SigtermTriggersCleanExit`
- `SigintTriggersCleanExit`
- `EngineDestroyedBeforeSignal`
- `MultipleEngines`

`IdempotentEnable` can remain an in-process test.

`ForkAfterInstallDoesNotHangChildSignal` should remain a direct `fork()` test because it specifically verifies inherited handler behavior. In that test, the engine and handler are installed before `fork()`, and the child only needs to call async-signal-safe operations such as `pause()` and `_exit()`. It does not need a fixed sleep because the handler state already exists when `fork()` returns.

The expected implementation scope is limited to:

```text
mooncake-transfer-engine/tests/graceful_shutdown_test.cpp
mooncake-transfer-engine/tests/CMakeLists.txt
```

No production graceful shutdown change should be required.

## Approaches that would only hide the issue

The following changes may reduce the failure rate but do not fix the invalid process model:

- increasing the 100 ms sleep;
- increasing the five-second timeout;
- marking the test `RUN_SERIAL`;
- disabling ASan for this test;
- disabling `USE_ETCD` for this test;
- adding only an unbounded readiness pipe;
- attempting to reset all affected library state with `pthread_atfork()`.

In particular, `pthread_atfork()` cannot generally repair inherited Go, sanitizer, allocator, logging, and arbitrary C++ library state.

## Acceptance criteria

- The test passes with the normal Linux CI configuration, including `USE_ETCD=ON` and `ENABLE_ASAN=ON`.
- At least 2,000 invocations under 32-way concurrency complete without a timeout.
- The same test passes with `USE_ETCD=OFF`.
- SIGTERM produces normal exit status 143.
- SIGINT produces normal exit status 130.
- Failure before readiness is reported within a bounded timeout.
- Failure after signal delivery is reported within a bounded timeout.
- Every failure path reaps the child and leaves no orphaned test process.
- Existing coverage for idempotent enable, destroyed engines, multiple engines, and fork-after-install behavior is retained.

## Impact

The confirmed impact is flaky CI and unrelated PRs being blocked by a Transfer Engine test failure.

This investigation does not show a failure in the normal graceful shutdown path of an application that installs the handlers without first entering an unsafe post-fork child state.

Applications that fork from a multithreaded process and then construct or reinitialize Mooncake components in the child without `exec()` would have the same general process-safety problem, but that is separate from the normal supported shutdown flow.

### Before submitting...

Icedcoco:
## Description

Fixes #3349.

`graceful_shutdown_test` could intermittently hang in CI when built with `USE_ETCD=ON`.

The test binary loads the Go c-shared etcd runtime, which starts background threads before GTest runs. The affected tests then called `fork()` and constructed `TransferEngine` in the child process. This performed memory allocation, mutex operations, logging initialization, and `std::thread` creation between `fork()` and `exec()`, which is unsafe when the parent process is already multithreaded.

The parent also used a fixed 100 ms sleep before sending SIGTERM or SIGINT. This did not guarantee that the child had installed its graceful shutdown handlers.

This change:

- starts the signal target with `posix_spawn()` and self-executes the test binary in an internal child mode;
- constructs `TransferEngine` only after the child has entered a fresh process image;
- replaces the fixed sleep with a readiness pipe;
- applies bounded timeouts to both readiness and child exit;
- kills and reaps the child on every relevant failure path;
- handles `EINTR` while waiting, reading, and writing;
- shares the helper across the single-engine, destroyed-engine, multiple-engine, SIGTERM, and SIGINT cases;
- keeps `ForkAfterInstallDoesNotHangChildSignal` as a direct `fork()` test because it specifically validates inherited handler behavior, while ensuring its child only calls async-signal-safe operations.

No production graceful shutdown code is changed. This is separate from the shutdown watcher signal-mask issue fixed by #3278.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

The test was built and run with both the CI-equivalent `USE_ETCD=ON + ENABLE_ASAN=ON` configuration and a minimal `USE_ETCD=OFF` configuration.

**Test commands:**

```bash
cmake --build build --target graceful_shutdown_test -j32

LSAN_OPTIONS=detect_leaks=0 \
ctest --test-dir build \
  -R '^graceful_shutdown_test$' \
  --output-on-failure
```

The previously failing SIGTERM case was also stress-tested with 32 concurrent workers:

```bash
seq 1 2000 | xargs -P32 -n1 sh -c \
  'LSAN_OPTIONS=detect_leaks=0 \
   ./build/mooncake-transfer-engine/tests/graceful_shutdown_test \
   --gtest_filter=GracefulShutdownTest.SigtermTriggersCleanExit \
   >/dev/null 2>&1'
```

The complete test binary was stress-tested 500 times with 32 concurrent workers.

**Test results:**

- [x] Unit tests pass
- [x] Manual testing done
- CI-equivalent `USE_ETCD=ON + ASan`: passed
- `USE_ETCD=OFF`: passed
- SIGTERM stress test, 2,000 runs at 32-way concurrency: 0 failures
- Complete suite stress test, 500 runs at 32-way concurrency: 0 failures
- No orphaned `graceful_shutdown_test` processes remained

## Checklist

- [x] I have updated the documentation (not applicable; test-only change)
- [x] I have added or updated tests to prove my changes are effective

Targeted pre-commit hooks passed. The full-file `cmake-format` hook was skipped because it rewrites unrelated pre-existing sections of `mooncake-transfer-engine/tests/CMakeLists.txt`; those unrelated changes were not included.

## AI Assistance Disclosure

- [ ] No AI tools were used

OpenAI Codex was used to investigate the CI failure pattern, implement the test process-isolation change, run local validation, and prepare this PR description. The human submitter is responsible for reviewing and defending the final change.

### DIFF
--- a/mooncake-transfer-engine/tests/graceful_shutdown_test.cpp
+++ b/mooncake-transfer-engine/tests/graceful_shutdown_test.cpp
@@ -13,72 +13,212 @@
 // limitations under the License.
 
 #include <gtest/gtest.h>
+#include <poll.h>
 #include <signal.h>
+#include <spawn.h>
 #include <sys/wait.h>
 #include <unistd.h>
 
+#include <cerrno>
+#include <chrono>
+#include <climits>
+#include <cstdio>
+#include <cstdlib>
+#include <cstring>
 #include <memory>
 
 #include "transfer_engine.h"
+
+extern char** environ;
 
 using namespace mooncake;
 
 namespace {
 
-void waitChildWithTimeout(pid_t pid, int* status) {
-    for (int i = 0; i < 50; ++i) {
+constexpr char kChildFlag[] = "--graceful-shutdown-child";
+constexpr int kChildTimeoutMs = 5000;
+
+enum class ChildMode { kSingle, kDestroyed, kMultiple };
+
+const char* childModeName(ChildMode mode) {
+    switch (mode) {
+        case ChildMode::kSingle:
+            return "single";
+        case ChildMode::kDestroyed:
+            return "destroyed";
+        case ChildMode::kMultiple:
+            return "multiple";
+    }
+    return "unknown";
+}
+
+bool parseChildMode(const char* value, ChildMode* mode) {
+    if (strcmp(value, "single") == 0) {
+        *mode = ChildMode::kSingle;
+    } else if (strcmp(value, "destroyed") == 0) {
+        *mode = ChildMode::kDestroyed;
+    } else if (strcmp(value, "multiple") == 0) {
+        *mode = ChildMode::kMultiple;
+    } else {
+        return false;
+    }
+    return true;
+}
+
+int runShutdownChild(ChildMode mode, int ready_fd) {
+    std::unique_ptr<TransferEngine> engine1;
+    std::unique_ptr<TransferEngine> engine2;
+
+    if (mode == ChildMode::kDestroyed) {
+        auto engine = std::make_unique<TransferEngine>(false);
+        engine->enableGracefulShutdown();
+    } else {
+        engine1 = std::make_unique<TransferEngine>(false);
+        engine1->enableGracefulShutdown();
+        if (mode == ChildMode::kMultiple) {
+            engine2 = std::make_unique<TransferEngine>(false);
+            engine2->enableGracefulShutdown();
+        }
+    }
+
+    const char ready = '1';
+    ssize_t bytes_written;
+    do {
+        bytes_written = write(ready_fd, &ready, sizeof(ready));
+    } while (bytes_written < 0 && errno == EINTR);
+    if (bytes_written != static_cast<ssize_t>(sizeof(ready))) return 111;
+    close(ready_fd);
+    for (;;) pause();
+}
+
+void killAndReap(pid_t pid) {
+    if (kill(pid, SIGKILL) != 0 && errno != ESRCH) {
+        ADD_FAILURE() << "kill(SIGKILL) failed: " << strerror(errno);
+    }
+
+    int status = 0;
+    while (waitpid(pid, &status, 0) < 0 && errno == EINTR) {
+    }
+}
+
+bool waitChildWithTimeout(pid_t pid, int* status) {
+    auto deadline = std::chrono::steady_clock::now() +
+                    std::chrono::milliseconds(kChildTimeoutMs);
+    for (;;) {
         pid_t ret = waitpid(pid, status, WNOHANG);
-        ASSERT_NE(ret, -1) << "waitpid() failed";
-        if (ret == pid) return;
+        if (ret == pid) return true;
+        if (ret < 0) {
+            if (errno == EINTR) continue;
+            int error = errno;
+            if (error != ECHILD) killAndReap(pid);
+            ADD_FAILURE() << "waitpid() failed: " << strerror(error);
+            return false;
+        }
+        if (std::chrono::steady_clock::now() >= deadline) break;
         usleep(100000);
     }
-    kill(pid, SIGKILL);
-    waitpid(pid, status, 0);
-    FAIL() << "child did not exit before timeout";
+
+    killAndReap(pid);
+    ADD_FAILURE() << "child did not exit before timeout";
+    return false;
+}
+
+bool waitForChildReady(int fd) {
+    pollfd pfd{fd, POLLIN, 0};
+    auto deadline = std::chrono::steady_clock::now() +
+                    std::chrono::milliseconds(kChildTimeoutMs);
+    int ret;
+    for (;;) {
+        auto remaining = std::chrono::duration_cast<std::chrono::milliseconds>(
+                             deadline - std::chrono::steady_clock::now())
+                             .count();
+        if (remaining <= 0) {
+            ret = 0;
+            break;
+        }
+        ret = poll(&pfd, 1, static_cast<int>(remaining));
+        if (ret >= 0 || errno != EINTR) break;
+    }
+
+    if (ret == 0) {
+        ADD_FAILURE() << "child did not report readiness before timeout";
+        return false;
+    }
+    if (ret < 0) {
+        ADD_FAILURE() << "poll() failed: " << strerror(errno);
+        return false;
+    }
+
+    char ready = 0;
+    ssize_t bytes_read;
+    do {
+        bytes_read = read(fd, &ready, sizeof(ready));
+    } while (bytes_read < 0 && errno == EINTR);
+    if (bytes_read != static_cast<ssize_t>(sizeof(ready)) || ready != '1') {
+        ADD_FAILURE() << "child exited before reporting readiness";
+        return false;
+    }
+    return true;
+}
+
+pid_t spawnReadyChild(ChildMode mode) {
+    int ready_pipe[2];
+    if (pipe(ready_pipe) != 0) {
+        ADD_FAILURE() << "pipe() failed: " << strerror(errno);
+        return -1;
+    }
+
+    char ready_fd[32];
+    snprintf(ready_fd, sizeof(ready_fd), "%d", ready_pipe[1]);
+    char* child_argv[] = {
+        const_cast<char*>("/proc/self/exe"), const_cast<char*>(kChildFlag),
+        const_cast<char*>(childModeName(mode)), ready_fd, nullptr};
+
+    pid_t pid = -1;
+    int spawn_error = posix_spawn(&pid, "/proc/self/exe", nullptr, nullptr,
+                                  child_argv, environ);
+    close(ready_pipe[1]);
+    if (spawn_error != 0) {
+        close(ready_pipe[0]);
+        ADD_FAILURE() << "posix_spawn() failed: " << strerror(spawn_error);
+        return -1;
+    }
+
+    bool ready = waitForChildReady(ready_pipe[0]);
+    close(ready_pipe[0]);
+    if (!ready) {
+        killAndReap(pid);
+        return -1;
+    }
+    return pid;
+}
+
+void expectGracefulExit(ChildMode mode, int signo) {
+    pid_t pid = spawnReadyChild(mode);
+    ASSERT_GT(pid, 0);
+
+    if (kill(pid, signo) != 0) {
+        int error = errno;
+        killAndReap(pid);
+        FAIL() << "kill(" << signo << ") failed: " << strerror(error);
+    }
+
+    int status = 0;
+    ASSERT_TRUE(waitChildWithTimeout(pid, &status));
+    ASSERT_TRUE(WIFEXITED(status))
+        << "Child did not exit normally (signaled: " << WIFSIGNALED(status)
+        << ")";
+    EXPECT_EQ(WEXITSTATUS(status), 128 + signo);
 }
 
 }  // namespace
 
 TEST(GracefulShutdownTest, SigtermTriggersCleanExit) {
-    pid_t pid = fork();
-    ASSERT_NE(pid, -1) << "fork() failed";
-
-    if (pid == 0) {
-        auto engine = std::make_unique<TransferEngine>(false);
-        engine->enableGracefulShutdown();
-        pause();
-        _exit(99);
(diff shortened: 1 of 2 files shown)


mooncake-transfer-engine/tests/graceful_shutdown_test.cpp
```diff
@@ -13,72 +13,212 @@
...
+    char ready = 0;
+    ssize_t bytes_read;
+    do {
+        bytes_read = read(fd, &ready, sizeof(ready));
+    } while (bytes_read < 0 && errno == EINTR);
+    if (bytes_read != static_cast<ssize_t>(sizeof(ready)) || ready != '1') {
+        ADD_FAILURE() << "child exited before reporting readiness";
+        return false;
+    }
+    return true;
```
alogfans, line 177: In waitForChildReady(), could we split EOF / read error / unexpected marker into distinct failure messages? That would make CI failures easier to diagnose.

mooncake-transfer-engine/tests/graceful_shutdown_test.cpp
```diff
@@ -13,72 +13,212 @@
...
+}
+
+void killAndReap(pid_t pid) {
+    if (kill(pid, SIGKILL) != 0 && errno != ESRCH) {
+        ADD_FAILURE() << "kill(SIGKILL) failed: " << strerror(errno);
+    }
+
+    int status = 0;
+    while (waitpid(pid, &status, 0) < 0 && errno == EINTR) {
+    }
```
alogfans: In killAndReap(), should we check/report non-EINTR waitpid() failures as well? Right now cleanup silently ignores them.

alogfans (approved):
Conditionally Approve. You should aware of the following comments.

mooncake-transfer-engine/tests/graceful_shutdown_test.cpp
Icedcoco: Good point. I split the readiness failures into three cases: EOF before readiness, a read() error with strerror(errno), and an unexpected marker with its numeric value. This should make future CI failures unambiguous. Addressed in 9591cc58e.

mooncake-transfer-engine/tests/graceful_shutdown_test.cpp
Icedcoco: Yes. killAndReap() now retries waitpid() only for EINTR and reports any other failure instead of silently returning. Addressed in 9591cc58e.

PR #3355, opened 2026-08-10
[Store] Dual rdma with coroutine te submit

G0-KU:
## Description

Offload TransferEngine **wait** (`WaitAllTransferBatches` / poll + sleep) from coroutine/RPC workers onto a dedicated `te_poll` thread pool, so multi-key P2P coroutines can yield during TE completion instead of occupying business threads.

This is a **store-side bridge**, not true TE async completion callbacks. Submit stays on the caller; only wait is offloaded when `te_async_poll_worker_num > 0` (default `32`; `0` keeps the previous sync-wait behavior).

### What this adds

- **DataManager TE wait bridge**
  - New `te_poll_executor_` (`AsyncMemcpyExecutor`) sized by `te_async_poll_worker_num`
  - `WaitAllTransferBatchesAsync`: when the pool is enabled, runs only `WaitAllTransferBatches` on poll workers and returns a `Future`; when `te_async_poll_worker_num == 0`, wait stays synchronous on the caller (legacy behavior)
  - `TransferDataAsync`: validate + submit on the caller, then returns that wait `Future` (submit is not offloaded)

- **Forward (accessor TE)**
  - Remote forward write/read already use 3-phase peer RPCs (`PreWrite`/`Pin` → TE → `WriteCommit`/`UnPin`)
  - TE wait is now `co_await TransferDataAsync` / `TeTransferFn` Future instead of blocking the coroutine worker
  - After wait, `AwaitTeExpectedFuture` resumes on `CurrentExecutor`, or `coro_io::get_global_executor()` if none is bound—so Commit/Revoke/UnPin/metrics do not run on `te_poll` threads
  - Forward write path catches poll shutdown/`setException` so the outer write Promise always completes

- **Local TE Put/Get**
  - `BuildDataCopierViaTe`: offloads wait via `FutureHandle`, keeping `TeSubmitResult` (handle / temp buffer) alive until wait finishes
  - `PutViaTe`: PreWrite + submit on caller; only wait goes to the pool; DRAM staging copy + `WriteCommit` / revoke run after wait on the caller / resume executor (`PutViaTeTaskHandle`), not inside the poll task

- **Reverse (owner TE)**
  - `ClientRpcService::{Read,Write}RemoteData` become `Lazy` handlers
  - Delegate to `ReadRemoteDataAsync` / `WriteRemoteDataAsync`: submit on the RPC path, `co_await` wait offload, then copy/commit (write) or return (read)
  - Sync `DataManager::{Read,Write}RemoteData` remain as `syncAwait` wrappers for existing unit tests

- **Config / wiring**
  - `te_async_poll_worker_num` on `P2PClientConfig` / `LocalTransferConfig`, exposed via client config builder (incl. dict), `real_client_main` gflag, and Python `MooncakeStore` ctor
  - Default `32`; set `0` to disable the poll pool

### InflightTracker change (needed for reverse transfer)

Reverse Lazy handlers hold `peer_tracker_` Guard across `co_await`. The old Guard kept a `std::shared_lock`, which must unlock on the **same OS thread** that locked—unsafe if the coroutine resumes on a `te_poll` (or other) thread.

**Change:** drain by atomic `inflight_` + CV instead of r/w lock.

- `Close()` rejects new `Enter()`
- `Wait()` blocks until `inflight_ == 0` (does **not** by itself reject new enters—always **`Close()` then `Wait()`**)
- Guard may be destroyed on any thread

`peer_tracker_` and `local_inflight_tracker_` share this implementation.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [x] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Unit**
- `inflight_tracker_test` — admit/close/drain; **cross-thread Guard destroy**; `Close`+`Wait`
- `data_manager_test` — poll executor create/disable; `TransferDataAsync` / `WaitAllTransferBatchesAsync`; FutureHandle; shutdown cancel surfaces as exception
- `client_config_builder_test` — `te_async_poll_worker_num` default / args / dict
- `client_rpc_service_test` — Lazy `Read/WriteRemoteData` via `syncAwait`

**Integration**
- `p2p_client_integration_test`
  - `TeAsyncPollForwardRemoteBatchPutAndGet` (`te_async={0,4}`)
  - `TeAsyncPollReverseRemoteBatchPutAndGet` (`te_async={0,4}`)

**Test commands:**
```bash
# adjust binary/ctest names to local build
ctest -R 'inflight_tracker_test|data_manager_test|client_config_builder_test|client_rpc_service_test|p2p_client_integration_test' --output-on-failure
```

**Test results:**
- [x] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

Cursor agent assisted implementation and PR text; author reviewed and owns the change.


mooncake-store/include/te_wait_future.h
```diff
@@ -0,0 +1,40 @@
...
+    try {
+        co_return co_await std::move(fut).via(ex);
+    } catch (const std::exception& e) {
+        LOG(ERROR) << "TE wait future exception: " << e.what();
+        co_return tl::unexpected(ErrorCode::INTERNAL_ERROR);
+    } catch (...) {
+        LOG(ERROR) << "TE wait future unknown exception";
+        co_return tl::unexpected(ErrorCode::INTERNAL_ERROR);
+    }
+}
```
wanyue-wy: 1. If this is intended to be a general‑purpose coroutine component, the naming and logging should not contain TE‑specific information, and it should be placed in `task_handle.h`.

2. If `data_manager` removes `te_poll_executor_` and `WaitAllTransferBatchesAsync` uses an independent executor for coroutine logic, would `AwaitTeExpectedFuture` still be necessary?

mooncake-store/src/data_manager.cpp
```diff
@@ -473,16 +496,100 @@ tl::expected<std::unique_ptr<TaskHandle<void>>, ErrorCode> DataManager::Put(
...
+    PutViaTeTaskHandle(
+        DataManager* dm, DataManager::KeyCtx kctx, UUID write_operation_id,
+        DataManager::TeSubmitResult ctx,
+        async_simple::Future<tl::expected<void, ErrorCode>> wait_future)
+        : dm_(dm),
+          kctx_(std::move(kctx)),
+          write_operation_id_(write_operation_id),
+          ctx_(std::move(ctx)),
+          wait_future_(std::move(wait_future)) {}
+
```
wanyue-wy: 1. It is recommended to abstract `PutViaTeTaskHandle` into a more general‑purpose handle.

2. If `PutViaTeTaskHandle` is still retained, `GetExpectedFuture` should be made a private member function of it.

mooncake-store/src/data_manager.cpp
```diff
@@ -1508,6 +1650,45 @@ tl::expected<void, ErrorCode> DataManager::TransferData(
...
+        return MakeReadyExpectedFuture(TransferData(
+            local_transfer_base, total_size, peer_buffers, opcode));
+    }
+    auto validate_result = ValidateRemoteBuffers(peer_buffers);
+    if (!validate_result) {
+        LOG(ERROR) << "TransferDataAsync: ValidateRemoteBuffers failed: "
+                   << toString(validate_result.error());
+        return MakeReadyExpectedFuture(
+            tl::make_unexpected(validate_result.error()));
+    }
```
wanyue-wy: Extract a common submit function to reduce duplication between `TransferData` and `TransferDataAsync`

mooncake-store/src/data_manager.cpp
```diff
@@ -1674,6 +1855,18 @@ tl::expected<void, ErrorCode> DataManager::WaitAllTransferBatches(
...
+DataManager::WaitAllTransferBatchesAsync(
+    std::vector<std::tuple<Transport::BatchID, size_t, std::string>> batches) {
+    if (!te_poll_executor_) {
+        return MakeReadyExpectedFuture(WaitAllTransferBatches(batches));
+    }
+    return te_poll_executor_->SubmitSingleTask<tl::expected<void, ErrorCode>>(
+        [this, batches = std::move(batches)]() mutable {
+            return WaitAllTransferBatches(batches);
+        });
+}
```
wanyue-wy, line 1975: The current implementation causes each thread in the `te_poll_executor_` pool to block on `WaitAllTransferBatches`. I suggest removing `te_poll_executor_` and adopting a "polling‑coroutine + independent‑executor model", where each wait becomes a suspended coroutine that repeatedly polls `transfer_engine_->getTransferStatus`. Then, it would be better to create a dedicated executor for these coroutines (e.g., using coro_io::io_context_pool(pool_size)).

mooncake-store/src/p2p_client_service.cpp
```diff
@@ -1012,6 +1015,7 @@ P2PClientService::CreatePutHandlesFromRoute(
         RunWriteWithRetry(std::move(promise), std::move(task.first_task),
                           std::move(task.first_op),
                           std::move(task.retry_op_list), keys[i], sizes[i])
+            .via(coro_io::get_global_executor())
             .start([](auto&&) {});
```
wanyue-wy, line 1033: It would be better for the `via` parts in `P2PClientService` to use an independent executor rather than `get_global_executor`, so that RPC is not affected. (you could also consider reusing the executor from `data_manager`). The same applies elsewhere.

mooncake-store/include/te_wait_future.h
G0-KU: Yes — still needed (now AwaitExpectedFuture). Wait runs on te_wait_pool_, so the Future completes there. Without hop-back, copy/commit/UnPin would resume on that pool. We don’t run wait on the caller’s executor: many callers are the RPC pool, and TE polling would crowd PreWrite/PinKey/Heartbeat off those threads. Hop-back is the cost of keeping wait off RPC.

mooncake-store/src/data_manager.cpp
G0-KU: te_poll_executor_ is gone. Wait is now a coroutine on a dedicated coro_io::io_context_pool (te_wait_pool_): poll getTransferStatus, co_await sleep_for to yield.

PR #3369, opened 2026-08-10
[TENT] reclaim rdma endpoints outside store lock

zxpdemonio:
## Description

This PR fixes a RDMA endpoint teardown deadlock risk in the TENT endpoint store.

`FIFOEndpointStore::reclaim()` and `SIEVEEndpointStore::reclaim()` previously called `RdmaEndPoint::finishDestroy()` while holding `endpoint_map_lock_`. `finishDestroy()` may enter blocking verbs teardown paths, while async event handling may need to look up endpoint ownership through the same store lock. Holding the store lock across endpoint destruction can therefore block teardown progress.

The fix changes reclaim to:

1. Take a strong-reference snapshot of `waiting_list_` under the store lock.
2. Release the store lock before calling `finishDestroy()`.
3. Re-acquire the store lock only to erase successfully destroyed endpoints.

This keeps retiring endpoints discoverable in `waiting_list_` while avoiding a Store lock / Endpoint teardown lock cycle.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
git diff --check
```

**Test results:**
- [ ] Unit tests pass

`git diff --check` passed.

Full unit test execution was not completed in the current environment because the local Mooncake checkout is missing required build dependencies/submodules (`pybind11`, Python development headers, and TENT standalone CMake dependency wiring for `yalantinglibs`).

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex was used to inspect the downstream TENT fix, compare it with upstream Mooncake, and prepare this minimal upstream patch. The submitter reviewed the final diff.

### DIFF
--- a/mooncake-transfer-engine/tent/src/transport/rdma/endpoint_store.cpp
+++ b/mooncake-transfer-engine/tent/src/transport/rdma/endpoint_store.cpp
@@ -122,6 +122,9 @@ void FIFOEndpointStore::evictOne() {
 void FIFOEndpointStore::reclaim() {
     std::vector<std::shared_ptr<RdmaEndPoint>> candidates;
     {
+        // Only protect the waiting_list_ iteration. finishDestroy() may wait
+        // for CQ drain and take the endpoint lock, so it must run after this
+        // store lock has been released.
         RWSpinlock::ReadGuard guard(endpoint_map_lock_);
         candidates.reserve(waiting_list_.size());
         for (auto& endpoint : waiting_list_) candidates.push_back(endpoint);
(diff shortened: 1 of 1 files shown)


alogfans:
Is it better to remove lock in `finishDestroy`?

zxpdemonio:
Good point. The expensive part is already outside the store lock: reclaim() now only takes a short read lock to snapshot waiting_list_, releases it, and then calls finishDestroy() on the copied shared_ptrs. I kept that snapshot lock because waiting_list_ is an unordered_set shared with remove()/evictOne()/getOrInsert(); iterating it without any lock would introduce a data race if another thread moves an endpoint into or out of the waiting list.

I added a comment to make this lock boundary explicit and pushed it in 38dc614b. Verified with tent_endpoint_store_test (8 tests passed).

alogfans (approved):
LGTM

PR #3370, opened 2026-08-10
[TENT] drain control callbacks on unregister

zxpdemonio:
## Description

This PR makes TENT `ControlService` callback unregister/replacement bounded and lifetime-safe.

Before this change, `setBootstrapRdmaCallback()` and `setNotifyCallback()` directly replaced the callback function. If a transport unregisters a callback while an RPC worker is still executing the old callback, shutdown may proceed and destroy transport-owned state while the in-flight callback still references it.

The fix adds:

1. Per-callback mutex and condition variable.
2. In-flight counters for BootstrapRdma and Notify callbacks.
3. Bounded drain when replacing/unregistering callbacks.
4. Self-unregister protection to avoid callback-triggered deadlock.
5. Local callback copies in RPC handlers so the callback remains valid during invocation.
6. Explicit `rpc_server_.reset()` in `ControlService::~ControlService()` while callback synchronization state is still alive.

If callback drain times out, the service logs an error and continues replacing the callback to keep shutdown bounded.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
git diff --check
```

**Test results:**
- [ ] Unit tests pass

`git diff --check` passed.

Full build/unit test execution was not completed in the current environment because the local Mooncake checkout is missing required build dependencies/submodules (`pybind11`, Python development headers, `numa.h`, and TENT standalone CMake dependency wiring for `yalantinglibs`).

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex was used to inspect the downstream TENT fix, compare it with upstream Mooncake, and prepare this minimal upstream patch. The submitter reviewed the final diff.

### DIFF
--- a/mooncake-transfer-engine/tent/include/tent/runtime/control_plane.h
+++ b/mooncake-transfer-engine/tent/include/tent/runtime/control_plane.h
@@ -19,9 +19,12 @@
 #include <netdb.h>
 
 #include <atomic>
+#include <chrono>
+#include <condition_variable>
 #include <cstdint>
 #include <functional>
 #include <memory>
+#include <mutex>
 #include <string>
 #include <thread>
 #include <unordered_map>
@@ -121,13 +124,9 @@ class ControlService {
 
     SegmentManager& segmentManager() { return *manager_.get(); }
 
-    void setBootstrapRdmaCallback(const OnReceiveBootstrap& callback) {
-        bootstrap_callback_ = callback;
-    }
+    void setBootstrapRdmaCallback(const OnReceiveBootstrap& callback);
 
-    void setNotifyCallback(const OnNotify& callback) {
-        notify_callback_ = callback;
-    }
+    void setNotifyCallback(const OnNotify& callback);
 
     Status start(uint16_t& port, bool ipv6_ = false);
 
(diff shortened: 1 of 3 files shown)


alogfans:
In rdma_transport.cpp, RdmaTransport::uninstall() never calls setBootstrapRdmaCallback(nullptr). It destroys workers, contexts, and other transport-owned state while ControlService may still retain a callback bound to the destroyed RdmaTransport. Please unregister and drain the bootstrap callback before destroying any state used by onSetupRdmaConnections().

alogfans:
Also, we recommend to use an RAII invocation guard for the in-flight counter and thread-local marker. Manual restoration works for the current ordinary exception paths, but future throwing operations could leave the counter permanently elevated.

zxpdemonio:
Addressed both points in 456d2791

- RdmaTransport::uninstall() now unregisters the BootstrapRdma callback via setBootstrapRdmaCallback(nullptr) before tearing down workers, contexts, buffer managers, or metadata-owned state. I kept it outside the installed_ branch so partially-installed transports are covered as well.
- onBootstrapRdma() and onNotify() now use an RAII invocation guard to restore the thread-local marker and decrement the in-flight counter on all exits from the invocation scope, including future throwing paths.
 
Verification: built the TENT target with BUILD_UNIT_TESTS=OFF / USE_TENT=ON successfully. I also tried configuring the unit-test build, but the environment failed while downloading googletest from GitHub, before compiling this change.

alogfans (approved):
LGTM

PR #3380, opened 2026-08-11
[TENT] Make ProxyManager staging cleanup safe and asynchronous

RuiqingFeng:
## Description

Fixes concurrency, resource-lifetime, and shutdown-safety issues in the TENT `ProxyManager` staging path by stabilizing `TaskInfo` storage, using atomic staging status, enforcing batch capacity limits, scoping staging buffer caches to individual tasks, and returning recoverable allocation errors.
Deferred cleanup retains internal batch references and keeps staging buffers pinned when `progressBatch()` fails or remains `PENDING`, preventing premature slab reuse and use-after-free.
Additionally, remote staging has been refactored from thread-blocking "fake async" to true event-driven asynchronous execution. By introducing `ControlClient::delegateAsync()` built on top of `CoroRpcAgent::callAsync()`, remote completion is now tracked non-blockingly via `RemoteStageOperation` handles, removing the need for `ThreadPool` thread dispatching, `std::future` polling, or test-specific friend declarations (`ProxyManagerTestPeer`).
Shutdown stops staging workers and drains retained local/cross-stage batches to a confirmed terminal state before releasing staging memory, while unresolved remote stage buffers remain pinned for DMA safety.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [x] Performance improvement
- [ ] Other

## How Has This Been Tested?

Added comprehensive unit test coverage in `mooncake-transfer-engine/tent/tests/progress_worker_test.cpp`:
- `ProxyManager.ConcurrentFirstAllocationIsBounded`
- `ProxyManager.ReclaimsStageBuffersBetweenWorkerTasks`
- `ProxyManager.DrainsInflightBatchesAfterProgressError`
- `ProxyManager.CleansUpDeferredStageBuffersWhenPendingBatchCompletes`
- `ProxyManager.RetainsBatchToPreventSlabReuseLeak`
- `ProxyManager.DeferredCleanupPollErrorKeepsStageBuffersPinned`
- `ProxyManager.RemoteCleanupProbeDoesNotBlockOnPendingOperation`
- `ProxyManager.DestructorWaitsForDeferredBatchBeforeFreeingStageMemory`
- `TransferEngineImpl.DirectSubmitRejectsBatchCapacityOverflow`

**Test commands:**
```bash
cmake -S . -B build-tent \
  -DUSE_TENT=ON \
  -DUSE_CUDA=OFF \
  -DWITH_STORE=OFF \
  -DWITH_STORE_RUST=OFF \
  -DBUILD_UNIT_TESTS=ON

cmake --build build-tent --target tent_progress_worker_test -j

./build-tent/mooncake-transfer-engine/tent/tests/tent_progress_worker_test
```

**End-to-End Data Integrity Verification:**
```bash
./build-tent/mooncake-transfer-engine/benchmark/tebench --backend=tent --xport_type=shm --check_consistency=true --op_type=mix --duration=2
```

**Test results:**
- [x] Unit tests pass

Ran all unit tests in `tent_progress_worker_test` and verified end-to-end data integrity via `tebench --check_consistency=true`; all tests passed.

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

### DIFF
--- a/mooncake-transfer-engine/tent/src/runtime/proxy_manager.cpp
+++ b/mooncake-transfer-engine/tent/src/runtime/proxy_manager.cpp
@@ -14,6 +14,7 @@
 
 #include "tent/runtime/proxy_manager.h"
 #include "tent/runtime/transfer_engine_impl.h"
+#include "remote_stage_operation.h"
 #include <algorithm>
 #include <cstring>
 #include <sstream>
@@ -32,13 +33,115 @@ ProxyManager::ProxyManager(TransferEngineImpl* impl, size_t chunk_size,
 
 ProxyManager::~ProxyManager() { deconstruct(); }
 
+void ProxyManager::flushPendingCleanups(bool unpin_remote_buffers) {
+    if (!has_pending_cleanups_.load(std::memory_order_relaxed)) return;
+    std::vector<Batch*> batches_to_release;
+    std::vector<uint64_t> local_addrs_to_unpin;
+    std::vector<std::pair<std::string, uint64_t>> remote_addrs_to_unpin;
+    {
+        std::lock_guard<std::mutex> lk(pending_cleanups_mu_);
+        if (pending_cleanups_.empty()) {
+            has_pending_cleanups_.store(false, std::memory_order_relaxed);
+            return;
+        }
+        for (auto it = pending_cleanups_.begin();
+             it != pending_cleanups_.end();) {
+            for (auto batch_it = it->batches.begin();
+                 batch_it != it->batches.end();) {
+                TransferStatus status;
+                auto poll_status = impl_->progressBatch(*batch_it, status);
+                if (!poll_status.ok()) {
+                    LOG_EVERY_N(WARNING, 100)
+                        << "Failed to poll deferred staging batch: "
+                        << poll_status;
+                    ++batch_it;
+                    continue;
+                }
+                if (status.s == PENDING) {
+                    ++batch_it;
+                    continue;
+                }
+                batches_to_release.push_back((Batch*)*batch_it);
+                batch_it = it->batches.erase(batch_it);
+            }
+
+            internal::pollRemoteOperations(it->remote_operations);
+            if (it->batches.empty()) {
+                local_addrs_to_unpin.insert(local_addrs_to_unpin.end(),
+                                            it->local_addrs.begin(),
+                                            it->local_addrs.end());
+                it->local_addrs.clear();
+                if (it->remote_operations.empty() && unpin_remote_buffers) {
+                    remote_addrs_to_unpin.insert(remote_addrs_to_unpin.end(),
+                                                 it->remote_addrs.begin(),
+                                                 it->remote_addrs.end());
+                    it->remote_addrs.clear();
+                }
+            }
+
+            if (it->local_addrs.empty() && it->remote_addrs.empty() &&
+                it->batches.empty() && it->remote_operations.empty()) {
+                it = pending_cleanups_.erase(it);
+                continue;
+            }
+            ++it;
+        }
+        if (pending_cleanups_.empty()) {
+            has_pending_cleanups_.store(false, std::memory_order_relaxed);
+        }
+    }
+
+    for (auto batch : batches_to_release) impl_->releaseBatch(batch);
+    for (auto addr : local_addrs_to_unpin) unpinStageBuffer(addr);
+    for (auto& [server, addr] : remote_addrs_to_unpin) {
+        ControlClient::unpinStageBuffer(server, addr);
+    }
+}
+
 Status ProxyManager::deconstruct() {
-    running_ = false;
+    if (!running_.exchange(false, std::memory_order_acq_rel)) {
+        return Status::OK();
+    }
     for (size_t i = 0; i < kShards; ++i) {
         shards_[i].cv.notify_all();
-        shards_[i].thread.join();
+        if (shards_[i].thread.joinable()) shards_[i].thread.join();
     }
-    // The workers are joined above, but Pin/Unpin arrive on the RPC thread.
+
+    auto has_pending_batches = [this]() {
+        std::lock_guard<std::mutex> lk(pending_cleanups_mu_);
+        return std::any_of(
+            pending_cleanups_.begin(), pending_cleanups_.end(),
+            [](const PendingBufferCleanup& p) { return !p.batches.empty(); });
+    };
+
+    flushPendingCleanups(false);
+    while (has_pending_batches()) {
+        std::this_thread::sleep_for(std::chrono::milliseconds(1));
+        flushPendingCleanups(false);
+    }
+
+    flushPendingCleanups(false);
+    std::vector<PendingBufferCleanup> remaining;
+    {
+        std::lock_guard<std::mutex> lk(pending_cleanups_mu_);
+        remaining = std::move(pending_cleanups_);
+        pending_cleanups_.clear();
+        has_pending_cleanups_.store(false, std::memory_order_relaxed);
+    }
+    for (auto& pending : remaining) {
+        if (!pending.remote_operations.empty()) {
+            LOG(WARNING)
+                << "Leaving " << pending.remote_addrs.size()
+                << " remote stage buffers pinned because "
+                << pending.remote_operations.size()
+                << " remote staging requests have not reached a confirmed "
+                   "terminal state";
+            continue;
+        }
+        for (auto& [server, addr] : pending.remote_addrs) {
+            ControlClient::unpinStageBuffer(server, addr);
+        }
+    }
     std::unique_lock<std::shared_mutex> guard(stage_buffers_mutex_);
     for (auto entry : stage_buffers_) {
         impl_->unregisterLocalMemory(entry.second.chunks);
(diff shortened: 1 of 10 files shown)


mooncake-transfer-engine/tent/src/runtime/proxy_manager.cpp
```diff
@@ -318,6 +329,44 @@ Status ProxyManager::transferEventLoop(StagingTask& task,
...
+            if (xfer_status.s == PENDING) continue;
+            auto free_status = impl_->freeBatch(batch);
+            if (!free_status.ok()) {
+                LOG(WARNING)
+                    << "Failed to free drained staging batch: " << free_status;
+            }
+            return true;
+        }
+    };
+    auto cleanup_inflight = [&]() {
```
alogfans, line 474: Could this function run indefinitely? This may block the dedicated workers.

mooncake-transfer-engine/tent/src/runtime/proxy_manager.cpp
```diff
@@ -155,7 +156,9 @@ Status ProxyManager::submit(TaskInfo* task, BatchID batch,
 
 Status ProxyManager::getStatus(TaskInfo* task, TransferStatus& task_status) {
     if (!task || !task->staging) return Status::InvalidArgument("Invalid task");
-    task_status.s = task->staging_status;
+    TransferStatusEnum staging_status;
+    __atomic_load(&task->staging_status, &staging_status, __ATOMIC_ACQUIRE);
```
alogfans: Recommend to use std::atomic for task->staging_status

mooncake-transfer-engine/tent/src/runtime/proxy_manager.cpp
```diff
@@ -239,17 +241,26 @@ void ProxyManager::runner(size_t id) {
...
-        auto status = transferEventLoop(task, &cache);
+        StageBufferCache cache(*this);
+        bool buffers_safe_to_release;
+        auto status = transferEventLoop(task, &cache, buffers_safe_to_release);
+        if (buffers_safe_to_release) {
+            cache.reset();
+        } else {
+            LOG(ERROR) << "Staging cleanup could not drain all in-flight "
+                          "operations; keeping stage buffers pinned";
+        }
```
alogfans: Is it possible to put `cache` in a deferred-cleanup collection?

alogfans:
Finally you are advised to check the correctness of staged transfer using `tebench --consistency_check`.

mooncake-transfer-engine/tent/src/runtime/proxy_manager.cpp
RuiqingFeng: Good catch! Removed the `while (true)` loop so `drain_batch()` now performs non-blocking polling and immediately delegates the batch to `impl_->freeBatch()` (which triggers lazy freeing in the background).
If the batch is still `PENDING`, it returns `false` to keep stage buffers safely pinned without blocking the worker thread. Thanks！

mooncake-transfer-engine/tent/src/runtime/proxy_manager.cpp
RuiqingFeng: Updated! Changed `staging_status` from `volatile TransferStatusEnum` with GCC `__atomic` builtins to `std::atomic<TransferStatusEnum>` in `TaskInfo`.

mooncake-transfer-engine/tent/src/runtime/proxy_manager.cpp
```diff
@@ -32,13 +32,73 @@ ProxyManager::ProxyManager(TransferEngineImpl* impl, size_t chunk_size,
...
+        for (auto it = pending_cleanups_.begin();
+             it != pending_cleanups_.end();) {
+            bool all_done = true;
+            for (auto batch : it->batches) {
+                TransferStatus status;
+                if (impl_->progressBatch(batch, status).ok() &&
+                    status.s == PENDING) {
+                    all_done = false;
+                    break;
+                }
```
alogfans: progressBatch() errors are currently treated as completion because only an OK result with PENDING keeps all_done false. A polling error does not guarantee that the transport has stopped accessing the staging buffers.

mooncake-transfer-engine/tent/src/runtime/proxy_manager.cpp
```diff
@@ -318,6 +404,49 @@ Status ProxyManager::transferEventLoop(StagingTask& task,
...
+            } else {
+                all_drained = false;
+                if (retain_status.ok() && bptr) {
+                    undrained_batches.push_back(chunk.batch);
+                }
+            }
+        }
+        for (auto& future : remote_futures) {
+            if (!future.valid()) continue;
+            auto status = future.get();
```
alogfans: The local batch cleanup is now non-blocking, but future.get() may still block indefinitely if a remote staging request stalls. This can block the shard worker and also prevent deconstruct() from completing.

mooncake-transfer-engine/tent/src/runtime/proxy_manager.cpp
```diff
@@ -32,13 +32,73 @@ ProxyManager::ProxyManager(TransferEngineImpl* impl, size_t chunk_size,
...
+
+    for (auto& pending : ready_to_clean) {
+        for (auto batch : pending.batches) {
+            impl_->releaseBatch((Batch*)batch);
+        }
+        for (auto addr : pending.local_addrs) unpinStageBuffer(addr);
+        for (auto& [server, addr] : pending.remote_addrs) {
+            ControlClient::unpinStageBuffer(server, addr);
+        }
+    }
```
alogfans: Entries remaining after flushPendingCleanups() have not been confirmed complete, but deconstruct() unconditionally releases their batch references and unpins the backing buffers. At this point the transport workers have not yet been stopped, so an in-flight DMA operation may still access this memory. Please cancel or drain these transfers to a confirmed terminal state, or quiesce the transports before releasing the staging buffers.

mooncake-transfer-engine/tent/src/runtime/proxy_manager.cpp
RuiqingFeng: Updated in the latest revision. A `progressBatch()` error is now treated as unconfirmed completion, so the retained batch reference and staging buffers remain deferred until a successful poll reports a terminal state. Thanks!

mooncake-transfer-engine/tent/src/runtime/proxy_manager.cpp
RuiqingFeng: Updated. The std::future and ThreadPool wrapper have been removed. Remote Delegate requests now use `ControlClient::delegateAsync()`, with completion tracked by RemoteStageOperation. Polling is non-blocking, and unresolved operations are deferred while their remote staging buffers remain pinned. Thanks!

mooncake-transfer-engine/tent/src/runtime/proxy_manager.cpp
RuiqingFeng: Updated. `deconstruct()` now stops and joins the staging workers, then drains retained local and cross-stage batches to a confirmed terminal state before releasing their references or local staging memory. Remote buffers associated with unresolved Delegate operations are not unpinned. Thanks!

staryxchen (approved):
LGTM. Please rebase onto main and rerun the tests to verify everything still passes on the latest base.

staryxchen:
Hi @RuiqingFeng, I noticed there are two unrelated commits slipped into this PR. Please keep this PR focused on a single topic — staging memory races and ProxyManager cleanup. Mixing unrelated fixes into one PR makes review harder and complicates bisect/blame later on, since a regression introduced by either fix would be indistinguishable from a staging-path bug when reverting. Could you move both commits to a separate standalone PR? Each fix looks reasonable in isolation and deserves its own PR so it can be reviewed and landed independently. We've already recorded issue #3459 for one such runtime queue dispatch test race, so that PR can reference it directly.

RuiqingFeng:
> Hi @RuiqingFeng, I noticed there are two unrelated commits slipped into this PR. Please keep this PR focused on a single topic — staging memory races and ProxyManager cleanup. Mixing unrelated fixes into one PR makes review harder and complicates bisect/blame later on, since a regression introduced by either fix would be indistinguishable from a staging-path bug when reverting. Could you move both commits to a separate standalone PR? Each fix looks reasonable in isolation and deserves its own PR so it can be reviewed and landed independently. We've already recorded issue #3459 for one such runtime queue dispatch test race, so that PR can reference it directly.

Hi @staryxchen, thanks for the guidance! That completely makes sense.
I have rebased this PR to keep it strictly focused on ProxyManager staging memory cleanup. The two unrelated fixes have been moved into standalone PR #3460  (with the runtime queue test race fix referencing #3459 ).
Thanks again for the review!

staryxchen:
> > Hi @RuiqingFeng, I noticed there are two unrelated commits slipped into this PR. Please keep this PR focused on a single topic — staging memory races and ProxyManager cleanup. Mixing unrelated fixes into one PR makes review harder and complicates bisect/blame later on, since a regression introduced by either fix would be indistinguishable from a staging-path bug when reverting. Could you move both commits to a separate standalone PR? Each fix looks reasonable in isolation and deserves its own PR so it can be reviewed and landed independently. We've already recorded issue #3459 for one such runtime queue dispatch test race, so that PR can reference it directly.
> 
> Hi @staryxchen, thanks for the guidance! That completely makes sense. I have rebased this PR to keep it strictly focused on ProxyManager staging memory cleanup. The two unrelated fixes have been moved into standalone PR #3460 (with the runtime queue test race fix referencing #3459 ). Thanks again for the review!

Thanks for the contribution! Merged.

PR #3204, opened 2026-07-30
[Store] Add terminal state for ordered oplog writer failures

Icedcoco:
> [!IMPORTANT]
> This PR is stacked on #3166 and must not be merged before #3166.
> After #3166 is merged, this branch will be rebased onto the updated `main`.

## Description

Depends on #3166. This branch is stacked on top of W01 and currently includes its retry backoff changes.

This PR adds a terminal failure contract to `OrderedOpLogWriter`:

- Exposes a read-only terminal state and a one-shot callback containing the error, terminal reason, durable prefix, and occurrence time.
- Treats transaction conflicts as fencing failures and other non-retryable errors as immediate terminal failures.
- Adds an optional per-batch retry deadline for retryable storage errors.
- Keeps the normal exponential backoff unchanged and checks the deadline immediately before starting the next retry transaction.
- Rejects new reservations after terminal failure and returns the latched error to valid outstanding reservations.
- Preserves the first terminal cause across later errors and `Stop()`.

`retry_timeout` defaults to zero, so the existing indefinite retry behavior remains unchanged unless a timeout is explicitly configured.

This PR only establishes the writer-level contract. Supervisor fail-stop handling, leadership release, and producer-view fencing remain follow-up work.

## Module

- [ ] Transfer Engine (`mooncake-transfer-engine`)
- [x] Mooncake Store (`mooncake-store`)
- [ ] Mooncake EP (`mooncake-ep`)
- [ ] Mooncake PG (`mooncake-pg`)
- [ ] Integration (`mooncake-integration`)
- [ ] P2P Store (`mooncake-p2p-store`)
- [ ] Python Wheel (`mooncake-wheel`)
- [ ] Common (`mooncake-common`)
- [ ] Mooncake RL (`mooncake-rl`)
- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**

```bash
cmake --build build --target ordered_oplog_writer_test -j32
LSAN_OPTIONS=detect_leaks=0 \
  ./build/mooncake-store/tests/ordered_oplog_writer_test

pre-commit run --files \
  mooncake-store/include/ha/oplog/ordered_oplog_writer.h \
  mooncake-store/src/ha/oplog/ordered_oplog_writer.cpp \
[...]

PR #3384, opened 2026-08-11
[Store] Fence batch oplog writers by producer view

Icedcoco:
> [!IMPORTANT]
> This PR is stacked on #3204 and must not be merged before #3204.
> After #3204 is merged, this branch will be rebased onto the updated `main`.

## Description

Depends on #3204. This branch currently includes its `OrderedOpLogWriter` terminal-state changes.

This PR adds the storage-level producer-view fencing contract for batch OpLog:

- Stores the current producer view in an independent `/oplog/<cluster>/producer_view` key, separate from the durable batch/sequence prefix.
- Adds `OpLogBatchStorage::ClaimProducerView()` to create or atomically advance that key with compare-and-swap.
- Treats claiming the current view as an idempotent success and rejects lower views with `ETCD_TRANSACTION_FAIL`.
- Makes claims independent of concurrent durable-prefix advancement, so batch and sequence progress cannot interfere with leadership claims.
- Adds `ValidateProducerView()` for the final pre-serving leadership check used by the follow-up integration.
- Adds a fenced `WriteBatchAndAdvancePrefix()` overload that compares the producer-view key in the same transaction that writes the batch record and advances the durable prefix.
- Resolves ambiguous transaction errors by rereading the affected state after a possible commit, covering initialization, view claims, and batch writes without weakening fencing.
- Keeps the existing unfenced overload as the S03 staging path; F01 will bind the claimed view into the production writer lifecycle.
- Uses a strict canonical positive decimal encoding for producer views and fails closed on malformed persisted values.
- Removes producer-view metadata from `DurablePrefix`; batch records and snapshot metadata retain their existing producer-view fields.

Production leadership wiring remains follow-up work in F01. The future supervisor will claim the ACQUIRED session view before promotion and validate leadership again immediately before serving.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**

```bash
  oplog_batch_storage_test \
  oplog_batch_codec_test \
  ordered_oplog_writer_test \
  oplog_batch_standby_reader_test -j32

LSAN_OPTIONS=detect_leaks=0 ./build/mooncake-store/tests/oplog_batch_storage_test
LSAN_OPTIONS=detect_leaks=0 ./build/mooncake-store/tests/oplog_batch_codec_test
LSAN_OPTIONS=detect_leaks=0 ./build/mooncake-store/tests/ordered_oplog_writer_test
LSAN_OPTIONS=detect_leaks=0 ./build/mooncake-store/tests/oplog_batch_standby_reader_test

pre-commit run --files <all changed files>
```

**Test results:**

- [x] Unit tests pass
- [ ] Manual testing done

Results:

- `oplog_batch_storage_test`: 50/50 passed
- `oplog_batch_codec_test`: 32/32 passed
- `ordered_oplog_writer_test`: 35/35 passed
- `oplog_batch_standby_reader_test`: 20/20 passed
- Pre-commit: all hooks passed

Coverage includes claim creation, same-view idempotency, stale and concurrent claim rejection, durable-prefix independence, ambiguous committed-transaction recovery, strict value parsing, backend-error propagation, atomic fenced writes, missing/corrupt view handling, and stale-writer fencing.

## Checklist

- [x] I have formatted the changed C++ files
- [x] I have run pre-commit on all changed files and all hooks pass
- [x] Documentation is not required for this staged internal storage contract
- [x] An RFC is not required because this change is under 500 LOC

## AI Assistance Disclosure

- [ ] No AI tools were used
- [x] AI tools were used

AI tools assisted with design review, implementation, test development, and code review. The submitter remains responsible for reviewing and defending the final changes.

### DIFF
--- a/mooncake-store/src/ha/oplog/oplog_batch_storage.cpp
+++ b/mooncake-store/src/ha/oplog/oplog_batch_storage.cpp
@@ -50,6 +50,23 @@ bool TryParseBatchIdFromKey(const std::string& key, uint64_t& batch_id) {
            result.ptr == suffix.data() + suffix.size();
 }
 
+bool TryParseProducerView(std::string_view value,
+                          ViewVersionId& producer_view_version) {
+    ViewVersionId parsed = 0;
+    const auto result =
+        std::from_chars(value.data(), value.data() + value.size(), parsed);
+    if (result.ec != std::errc() || result.ptr != value.data() + value.size() ||
+        parsed <= 0 || std::to_string(parsed) != value) {
+        return false;
+    }
+    producer_view_version = parsed;
+    return true;
+}
+
+bool IsAmbiguousTxnError(ErrorCode error) {
+    return error == ErrorCode::ETCD_OPERATION_ERROR;
+}
+
 }  // namespace
 
 OpLogBatchStorage::OpLogBatchStorage(std::string cluster_id,
@@ -98,18 +115,24 @@ ErrorCode OpLogBatchStorage::InitDurablePrefix(DurablePrefix& prefix) {
                             .expected_value = ""});
     txn.puts.push_back(
         {.key = durable_key, .value = EncodeDurablePrefix(initial)});
-    err = backend_.Txn(txn);
-    if (err == ErrorCode::OK) {
+    const ErrorCode txn_err = backend_.Txn(txn);
+    if (txn_err == ErrorCode::OK) {
         prefix = initial;
         return ErrorCode::OK;
     }
-    if (err == ErrorCode::ETCD_TRANSACTION_FAIL) {
-        if ((err = ReadDurablePrefix(prefix)) != ErrorCode::OK) {
-            return err;
-        }
-        return ValidateDurablePrefixAtStartup(prefix);
+    if (txn_err != ErrorCode::ETCD_TRANSACTION_FAIL &&
+        !IsAmbiguousTxnError(txn_err)) {
+        return txn_err;
     }
-    return err;
+
+    err = ReadDurablePrefix(prefix);
+    if (err != ErrorCode::OK) {
+        return IsAmbiguousTxnError(txn_err) &&
+                       err == ErrorCode::ETCD_KEY_NOT_EXIST
+                   ? txn_err
+                   : err;
+    }
+    return ValidateDurablePrefixAtStartup(prefix);
 }
 
 ErrorCode OpLogBatchStorage::ReadDurablePrefix(DurablePrefix& prefix) {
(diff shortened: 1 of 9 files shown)


mooncake-store/src/ha/oplog/oplog_batch_storage.cpp
```diff
@@ -130,6 +130,55 @@ ErrorCode OpLogBatchStorage::ReadDurablePrefix(DurablePrefix& prefix) {
...
 
+ErrorCode OpLogBatchStorage::ClaimProducerView(
+    ViewVersionId producer_view_version, DurablePrefix& prefix) {
+    if (!IsValidClusterId() || producer_view_version <= 0 ||
+        !backend_.SupportsTxn()) {
+        return ErrorCode::INVALID_PARAMS;
+    }
+
+    const std::string key = BuildDurablePrefixKey(cluster_id_);
+    while (true) {
```
Aionw, line 179: The CAS itself prevents a lost update, but this loop has no backoff, deadline, or cancellation. A live writer changes this same key on every batch, so repeated compare failures can make leader startup spin indefinitely while hammering etcd. There is also an ambiguous-outcome case: `backend_.Txn()` maps a commit RPC error to an operation error, but the server may already have committed the new view; returning immediately then reports a failed claim even though the old writer has been fenced. Please use a bounded retry policy with backoff/cancellation and read the prefix back after retryable operation errors (`view == requested` => success, `view > requested` => fenced, lower view => retry). A separate producer-view key, compared by batch transactions, would avoid the claim-versus-prefix-update contention entirely.

Aionw (commented):
One concern around the producer-view claim transaction and its retry/error semantics.

Icedcoco:
Update: this PR now uses a revised S03 design. The producer view is stored in an independent `/oplog/<cluster>/producer_view` key instead of `DurablePrefix`, so view claims are no longer affected by OpLog batch or sequence advancement. Claims are monotonic and idempotent, while fenced batch writes compare both the producer view and durable prefix in the same transaction.

The branch was also cleaned up into a single S03-only commit, removing the previous W03 and merge-history commits.

mooncake-store/src/ha/oplog/oplog_batch_storage.cpp
Aionw: Thanks for separating the producer view from the durable prefix. That addresses the claim-versus-batch-update contention and removes the main reason this loop could spin under normal write traffic.

The ambiguous transaction outcome is still unresolved, though. `backend_.Txn()` returns immediately for every non-compare error, but an etcd `Commit()` RPC can fail at the client after the server has already persisted the requested view. In that case this method reports a failed claim even though the old writer has been fenced.

Please read the producer-view key back after a retryable operation error and classify the result: requested view => success, higher view => fenced, lower/missing view => retry or return the original error according to a bounded policy. It would also be useful to add a fake-backend test that applies the transaction and then returns `ETCD_OPERATION_ERROR`; the current propagation test only models an error before the write.

Aionw (approved):
LGTM

00fish0 (approved):

Issue #3158, opened 2026-07-28
[RFC]: Decouple Mooncake Store business logic from metrics


Aionw:
## Summary

Mooncake Store currently uses observability gauges as inputs to several business decisions. This reverses the intended dependency direction:

```text
current:   business mutation -> metric bookkeeping -> metric read -> business decision
proposed:  business mutation -> authoritative domain state -> business decision
                                                   \-> metrics projection
```

Metrics should be a lossy, replaceable projection of runtime state. Disabling, resetting, restoring, or changing the metrics implementation must not change eviction, promotion, or replica-placement behavior.

This RFC proposes to move all business reads to authoritative internal state, keep metrics write-only from business code, and derive storage gauges from domain snapshots.

## Motivation

Depending on metrics for correctness is counterintuitive and fragile:

- metric gauges duplicate state already owned by allocators and segment managers;
- every allocate/deallocate, mount/unmount, restore, CXL, and NoF path must update both copies correctly;
- the global ratio reads allocated and capacity gauges separately, so it is not a coherent snapshot;
- reset/rebuild logic can temporarily make a healthy, full cluster appear empty;
- a missed increment/decrement changes eviction and promotion behavior rather than only degrading observability;
- singleton metric state leaks across tests and encourages tests to use telemetry as a business-state oracle;
- future metrics backends, compile-time switches, sampling, or asynchronous export become correctness-sensitive.

The rule should be:

> Business logic may emit metrics, but must not read metrics. Metrics may read/project business state, but must never be the source of truth.

## Current coupling audit

The audit covered production code in Mooncake Store, Transfer Engine, TENT, PG, and EP. Tests, benchmarks, admin summaries, and RPCs whose explicit purpose is to expose metrics are classified separately.

### True control-flow dependencies

| Business behavior | Current input | Effect |
| --- | --- | --- |
| DRAM eviction trigger and eviction target | `MasterMetricManager::get_global_mem_used_ratio()` in `MasterService::EvictionThreadFunc` | Decides whether to evict and computes `BatchEvict` target/lower bound. Runs every 10 ms. |
| NoF eviction trigger and eviction target | `MasterMetricManager::get_global_nof_used_ratio()` in `MasterService::EvictionThreadFunc` | Decides whether to evict and computes `NoFBatchEvict` target/lower bound. |
[...]

PR #3385, opened 2026-08-11
[Store] Decouple business decisions from storage metrics

Hubert-Zhu:
## Description

This refactor makes storage metrics a projection of authoritative Store domain state instead of an input to business decisions:

- Add DRAM/NoF usage snapshots derived from mounted allocators (with shared CXL allocator deduplication).
- Keep O(1) aggregate DRAM/NoF reads for eviction and promotion via a simple domain usage tracker (metric-gauge style atomics), while retaining O(n) per-segment snapshots for observability.
- Refresh Prometheus DRAM/NoF gauges from tiered snapshots on admin metric reads and periodic reports; clear stale HA leader projections on standby.
- Remove business-oriented DRAM/NoF ratio APIs from `MasterMetricManager`.
- Cover metric reset, snapshot restore/reset, allocator replacement, standby/HA transition, shared CXL, NoF, lock isolation, and Prometheus projection behavior.

Implementation is split into reviewable slices:

1. Decouple DRAM eviction and promotion decisions from metrics.
2. Project storage metrics from domain snapshots, with API cleanup and resilience tests.
3. Make business-path domain usage reads O(1) with a lightweight tracker.
4. Harden allocator lifetime, CXL publication, and HA metric handoff.

## Module

- [ ] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Breaking change
- [x] Documentation update
- [x] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
./scripts/code_format.sh -b origin/main
pre-commit run --files <touched files>
# unit coverage added/updated in:
#   segment_test, master_metrics_test, promotion_on_hit_test
```

**Test results:**
- [ ] Unit tests pass

Local formatting/pre-commit were run on touched files during preparation. Full compile/unit validation is expected from Linux PR CI.

## Checklist

- [x] For changes >500 LOC: I have filed an RFC issue (#3158)

## AI Assistance Disclosure

- [ ] No AI tools were used

Cursor and Codex


Hubert-Zhu:
I have not full self-review all the test cases, but the feature code is self-reviewed line by line. This is my first big PR in this repo. Please forgive me if I miss any edge cases.

Aionw:
@Hubert-Zhu While reviewing the new aggregate usage tracker lifecycle, I found a concurrency issue around explicit detachment.

`UnmountSegment()`, `Remove()`, and `BatchEvict()` all take `snapshot_mutex_` in shared mode, so they can run concurrently. During unmount, `PrepareUnmountSegment()` holds `segment_mutex_` and calls `allocator->DetachUsageTracker()`, which subtracts the allocator's current contribution and resets `usage_tracker_`. At the same time, an `AllocatedBuffer` destroyed by Remove/eviction can successfully `weak_ptr::lock()` the same allocator and enter `deallocate() -> RecordDeallocation()` without taking `segment_mutex_`.

That creates two problems:

1. `usage_tracker_` is read and reset concurrently, which is a data race on the shared_ptr object.
2. Depending on the interleaving, detach can subtract the allocator's full usage while deallocation also subtracts its delta, potentially underflowing the aggregate `size_t`. Since eviction and promotion now read this tracker directly, the corruption can persist and affect business decisions.

I do not think we should put a mutex/shared_mutex on every allocate/deallocate just to protect this rare lifecycle transition. A zero-hot-path-overhead design would be:

- Attach exactly once before an allocator is published; keep its registration immutable while the allocator is live.
- Remove the public/general-purpose `DetachUsageTracker()` path.
- On normal unmount, remove the allocator from `AllocatorManager` and release the mounted shared_ptr, but do not explicitly detach.
- An in-flight deallocation already holds a local allocator shared_ptr, so the allocator cannot be destroyed until that operation finishes. Let the allocator destructor RAII-detach the registration when the last shared_ptr is released.
- Model the per-allocator attachment as an allocator-owned `std::unique_ptr<StorageUsageRegistration>`; the registration can retain the shared aggregate `StorageUsageTracker`.
- Handle allocator replacement/CXL replacement as private, explicitly quiescent ownership-transfer operations: release the old allocator before publishing/attaching the new one, or transfer the registration internally. They should not require a generally callable detach API.

The aggregate may retain an unmounted allocator for the short period needed for already-running operations to drain, which seems compatible with the documented best-effort topology semantics and avoids adding synchronization to the allocation hot path.

It would also be valuable to add a forced-interleaving or TSAN test covering concurrent `AllocatedBuffer` destruction and `PrepareUnmountSegment()`.

Hubert-Zhu:
Make sense. I will fix it.

mooncake-store/src/master_metric_manager.cpp
```diff
@@ -761,13 +743,42 @@ int64_t MasterMetricManager::get_total_nof_capacity() {
...
+        for (const auto& [segment_name, usage] : tier.segments) {
+            current_segments.insert(segment_name);
+            allocated_by_segment.update({segment_name},
+                                        static_cast<int64_t>(usage.used_bytes));
+            capacity_by_segment.update(
+                {segment_name}, static_cast<int64_t>(usage.capacity_bytes));
+        }
+        for (const auto& segment_name : projected_segments) {
+            if (!current_segments.contains(segment_name)) {
+                allocated_by_segment.update({segment_name}, 0);
```
Aionw: [P2] Remove labels that disappear from the authoritative snapshot

`dynamic_gauge::update(..., 0)` keeps or recreates the label. Immediately afterward, `projected_segments` is replaced with `current_segments`, so the next projection no longer knows about the missing segment and can never remove that zero-valued series. Repeated mount/unmount cycles therefore accumulate stale Prometheus label sets. Please call `remove_label_value({{"segment", segment_name}})` on both per-segment gauges for entries absent from the new snapshot, and add a test asserting that serialized metrics no longer contain the removed segment.

mooncake-store/src/allocator.cpp
```diff
@@ -11,6 +11,34 @@
...
+
+void BufferAllocatorBase::RecordAllocation(size_t bytes) noexcept {
+    cur_size_.fetch_add(bytes, std::memory_order_relaxed);
+    if (usage_registration_) {
+        usage_registration_->AddUsedBytes(bytes);
+    }
+}
+
+void BufferAllocatorBase::RecordDeallocation(size_t bytes) noexcept {
+    if (auto hook = record_deallocation_hook_.load(std::memory_order_acquire)) {
```
Aionw: [P2] Keep test coordination out of the deallocation hot path

This testing hook adds an acquire atomic load and branch to every production deallocation, exposes a global mutable test API on `BufferAllocatorBase`, and makes the test rely on a process-global function pointer and static state. The race can be forced without touching production code: define a test-only `BlockingAllocator` or subclass `OffsetBufferAllocator`, override the virtual `deallocate()`, signal an entered semaphore, wait on a resume semaphore, and then delegate to the real deallocation or `RecordDeallocation`. That proves the local `shared_ptr` keeps the allocator and registration alive while unmount drops its reference, with zero production-path overhead. Please remove the setter, static atomic, and hook branch.

Aionw (approved):
LGTM

ykwd (approved):

PR #3390, opened 2026-08-11
[Bugfix][TE] Install shutdown handlers before unblocking signals

SongOf:
Description
  startSignalWatcherLocked() restored the caller's signal mask before installGracefulShutdownHandlers() had called sigaction(), leaving a window where SIGTERM/SIGINT was deliverable while the disposition was still SIG_DFL. A signal arriving in that window kills the process outright and the graceful shutdown never runs — the containerized-startup case the feature exists for. Creating the watcher thread is the slowest step on that path, so the exposure is widest on a cold process.

   pthread_sigmask(SIG_BLOCK, {TERM,INT})
   std::thread(signalWatcher).detach()     <-- slowest step
   pthread_sigmask(SIG_SETMASK, saved)     <-- unblocked here; a signal that arrived above is delivered NOW
   sigaction(SIGTERM/SIGINT)               <-- ...but the handler lands here

  This hands the saved mask back to the caller and restores it only after sigaction() has installed the handlers, on every path including the thread-creation failure path. A signal arriving mid-install now stays pending and reaches the real handler. The exposed window shrinks to the pipe() call preceding the mask block.

  The mask itself was introduced by #3278; this only widens its scope to cover the whole install. Independent of the test-side fork() hangs fixed in #3351.

  Module

  - [ ] CI/CD
  - [ ] Docs
  - [ ] Other

  Type of Change

  - [x] Bug fix
  - [ ] New feature
  - [ ] Refactor
  - [ ] Breaking change
  - [ ] Documentation update
  - [ ] Performance improvement
  - [ ] Other

  How Has This Been Tested?

  graceful_shutdown_test passes (6/6) and 25 rounds x 8 concurrent workers ran without a failure. The suite is unaffected by this change: it signals only after the child reports readiness, so it never enters the install window.

  To measure the window itself, the install sequence was extracted into a standalone harness mirroring graceful_shutdown.cpp, with a knob to widen each phase. Counting runs where the child died by the default disposition instead of exiting 128 + signo, before vs. after this change:

  - masked window widened to 150 ms, 200 runs: 200 killed, then 0
  - post-unblock window widened to 150 ms, 200 runs: 200 killed, then 0
  - natural timing, kill delay swept 50 us - 2 ms, 576 runs: 52 killed, then 18

  The residual 18 are signals delivered before enableGracefulShutdown() is entered at all, which no ordering inside the function can affect.

  Test commands:
  cmake --build build --target graceful_shutdown_test -j"$(nproc)"
  ASAN_OPTIONS=detect_leaks=0 ctest --test-dir build \
    -R '^graceful_shutdown_test$' --output-on-failure

  Test results:
  - [x] Unit tests pass

  Checklist

  No test is added: the race needs the signal to land inside a window that is microseconds wide and not externally observable, which cannot be pinned deterministically without a test-only seam in the install path. The numbers
  above come from the standalone harness; happy to add it under tests/ as a stress target if reviewers want it in-tree.

  AI Assistance Disclosure

  - [ ] No AI tools were used

### DIFF
--- a/mooncake-transfer-engine/src/graceful_shutdown.cpp
+++ b/mooncake-transfer-engine/src/graceful_shutdown.cpp
@@ -95,7 +95,10 @@ void signalWatcher() {
     _Exit(1);
 }
 
-bool startSignalWatcherLocked(pid_t current_pid) {
+bool startSignalWatcherLocked(pid_t current_pid, sigset_t* saved_mask,
+                              bool* mask_blocked) {
+    *mask_blocked = false;
+
     if (g_signal_pipe[0] >= 0) close(g_signal_pipe[0]);
     if (g_signal_pipe[1] >= 0) close(g_signal_pipe[1]);
     g_signal_pipe[0] = -1;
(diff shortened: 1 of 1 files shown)


alogfans (approved):
LGTM

PR #3403, opened 2026-08-12
[CI/Build] Prevent fixed-port collisions in nightly tests

Icedcoco:
## Description

This PR fixes two independent process/port reuse issues found in the nightly test workflow.

### 1. `mooncake_client` cannot bind to port 50052

The original failure occurred in the upstream nightly run:

https://github.com/kvcache-ai/Mooncake/actions/runs/31511779866/job/93849394041

The standalone `mooncake_client` failed to start on `0.0.0.0:50052`, causing the dummy-client tests to fail.

This was not caused by another `mooncake_client` still listening on the port. An earlier outbound connection could be assigned source port 50052 from Linux's ephemeral port range. The resulting socket remained in `TIME_WAIT` on the runner's network interface and prevented a subsequent wildcard bind to `0.0.0.0:50052`.

The fix reserves port 50052 through `net.ipv4.ip_local_reserved_ports` before running the tests. Existing reserved-port settings are preserved. This only prevents automatic ephemeral allocation; `mooncake_client` can still explicitly bind to port 50052.

The same workaround already exists in the regular CI workflow, but was missing from the nightly workflow.

### 2. A stale master process keeps ports 50051/9003 occupied

After fixing the 50052 collision, the first validation run exposed a second issue when consecutive tests restarted `mooncake_master` with different configurations:

https://github.com/Icedcoco/Mooncake/actions/runs/31578274830/job/94055282133

The Python console entry point started the packaged binary with `subprocess.call()`. Therefore, the PID recorded by `scripts/run_tests.sh` was the Python wrapper PID rather than the actual server PID.

Terminating the wrapper did not reliably terminate its child process. The old master could remain alive on ports 50051 and 9003, causing the next master to report:

```text
Failed to start master admin server on 0.0.0.0:9003: Address already in use
```

The new process would then exit, while subsequent clients could connect to the stale master running with the previous test's configuration.

The fix changes the `mooncake_master` and `mooncake_client` console entry points to use `os.execv()`. The Python wrapper is replaced by the packaged binary, so the PID recorded by the test script remains the PID of the actual service. Signals and exit status now apply directly to that service.

`transfer_engine_bench` is unchanged because it is not managed as a long-running background service by this test sequence.

## Module

- [x] CI/CD

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

The TIME_WAIT behavior was reproduced with an isolated network namespace and a restricted ephemeral range. Reserving 50052 made automatic allocation skip that port while explicit binding to 50052 continued to work.

A fork nightly workflow was also run after both fixes:

https://github.com/Icedcoco/Mooncake/actions/runs/31583091793/job/94070540664

Confirmed results:

- All 121 CTest tests passed.
- The Rust smoke test passed.
- The Go integration tests passed.
- The Python wheel was built and installed successfully.
- The dummy-client tests passed and `mooncake_client` bound to port 50052.
- The SSD-offload master was terminated successfully.
- The promotion-on-hit master started on ports 50051/9003 with the expected configuration.
- No `kill: No such process`, `Address already in use`, or master-admin bind failure occurred.

Targeted pre-commit checks also passed:

```bash
SKIP=cmake-format pre-commit run --files \
  .github/workflows/nightly.yml \
  mooncake-wheel/mooncake/cli.py \
  mooncake-wheel/mooncake/cli_client.py
```

**Test results:**

- [x] Unit tests pass
- [ ] Integration tests pass
- [x] Manual testing done

## Remaining Failure / Follow-up

The validation workflow still finished with a failure in:

```text
TestPromotionOnHit.test_promotion_after_repeated_hits
```

This happened after both port/process issues were cleared. The new master started successfully with `enable_offload=1`, `offload_on_evict=1`, and promotion-on-hit enabled.

The metrics at failure time showed:

```text
Promotion: in_flight=1, admitted=7, completed=2,
failed=0, cancelled=4, expired=0
```

The test selected a `LOCAL_DISK`-only key, triggered promotion reads, waited a fixed 25 seconds, and then expected that key to have a `MEMORY` replica. The selected key had not completed promotion within that window.

The current evidence does not distinguish between:

- a timing assumption in the test, where a fixed 25-second wait is insufficient for multiple queued promotions; or
- a product-side scheduling/cancellation issue that leaves the selected promotion stalled or cancelled.

This failure is separate from the 50052 and 9003 issues and is intentionally not addressed in this PR. A follow-up should reproduce the promotion test independently, track the selected key's task lifecycle, and then either replace the fixed delay with condition-based polling or fix the underlying promotion scheduling behavior.

## Checklist

- [x] Documentation is not applicable
- [x] An RFC is not required because this change is under 500 LOC

## AI Assistance Disclosure

- [ ] No AI tools were used
- [x] AI tools were used

Codex assisted with CI log analysis, root-cause reproduction, implementation, and validation. The human submitter has reviewed and is responsible for the final changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -252,6 +252,7 @@ jobs:
         source test_env/bin/activate
         mkdir -p /tmp/mooncake_storage
         mooncake_master \
+          --default_kv_lease_ttl=500 \
           --eviction_high_watermark_ratio=0.95 \
           --cluster_id=ci_test_cluster \
           --port 50051 \
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -286,6 +286,9 @@ jobs:
 
       - name: Run CTest unit tests
         run: |
+          # Keep ephemeral TIME_WAIT sockets off mooncake_client's RPC port.
+          reserved_ports=$(sysctl -n net.ipv4.ip_local_reserved_ports)
+          sudo sysctl -w "net.ipv4.ip_local_reserved_ports=${reserved_ports:+$reserved_ports,}50052"
           cd build
           export LD_LIBRARY_PATH=${LD_LIBRARY_PATH:-}:/usr/local/lib
           MC_METADATA_SERVER=http://127.0.0.1:8080/metadata \
--- a/mooncake-wheel/mooncake/cli.py
+++ b/mooncake-wheel/mooncake/cli.py
@@ -6,7 +6,6 @@ Minimal CLI module for mooncake_master.
 import os
 import stat
 import sys
-import subprocess
 
 
 def main():
@@ -23,8 +22,8 @@ def main():
         st = os.stat(bin_path)
         os.chmod(bin_path, st.st_mode | stat.S_IXUSR | stat.S_IXGRP | stat.S_IXOTH)
 
-    # Run the binary with all arguments passed through
-    return subprocess.call([bin_path] + sys.argv[1:])
+    # Preserve the CLI process ID so callers can reliably stop the server.
+    os.execv(bin_path, [bin_path] + sys.argv[1:])
 
 
 if __name__ == "__main__":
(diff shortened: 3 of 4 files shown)


Aionw (approved):

zxpdemonio:
LGTM.

zxpdemonio (approved):

PR #3404, opened 2026-08-12
[Bugfix][TENT] Keep a delegated transfer off the RPC event loop

SongOf:
Description

  CoroRpcAgent runs the server with a single io_context (kRpcThreads == 1), and every connection's coroutine is scheduled on it. A non-coroutine handler is invoked inline from that coroutine, so it never suspends and pins the thread for its whole duration — for every peer's connection, not just its own.

  ControlService::onDelegate does exactly that: it runs an entire transfer to completion via transferSync(). While a delegated transfer is in flight this node answers no RPC at all, BootstrapRdma included.

  Fix: make process() a coroutine and add a per-handler offload flag to registerFunction(). An offloaded handler runs via co_await coro_io::post() on ylt's blocking executor, which suspends the connection coroutine and returns the io_context thread to the event loop. The response is still delivered on the same RPC, so callers are unaffected. Only Delegate is marked; the other ten handlers stay inline.

  Two things for reviewers to weigh:
  - Requests within a connection are no longer ordered against an offloaded handler. Synchronous callers are unaffected; pipelining ones would see it.
  - coro_io::post uses the global blocking executor, whose queue is unbounded, so concurrent Delegates consume executor threads. The single-thread model had implicit backpressure; an explicit bound is left to a follow-up.

  Raising kRpcThreads was rejected as the alternative: it would make all eleven handlers concurrent at once, turning the unlocked func_map_ read into a real data race and exposing every other single-RPC-thread assumption.

  Module

  - [ ] CI/CD
  - [ ] Docs
  - [ ] Other
  
  Type of Change

  - [x] Bug fix
  - [ ] New feature
  - [ ] Refactor
  - [ ] Breaking change
  - [ ] Documentation update
  - [ ] Performance improvement
  - [ ] Other

  How Has This Been Tested?
  
  Test commands:
  ctest -R "tent_delegate_rpc_test|tent_rpc_handler_isolation_test" --output-on-failure
  ctest --output-on-failure

  Test results:
  - [x] Unit tests pass

  Two new tests, neither needing an RDMA device:

  - tent_rpc_handler_isolation_test (3 cases) — with an 800 ms handler in flight, an unrelated RPC on another connection must return within 400 ms. Switching that handler back to inline makes it fail, so it tests what it claims to.
  - tent_delegate_rpc_test (3 cases) — Delegate had no coverage at all. Drives ControlClient::delegate → onDelegate → transferSync against a real engine on loopback with a FakeTransport: happy path, error propagation back over the RPC, and 8 concurrent delegates (only reachable with this change). Stubbing onDelegate to a no-op fails all three.

  Full TENT suite: 66 tests, 63 pass. The 3 failures (tcp_transport_test, transfer_metadata_test, memory_location_test) are pre-existing container-environment issues — no etcd metadata server, and no CAP_SYS_NICE for mbind —
  unrelated to this change.

  Checklist
  
  codespell flagged a pre-existing Re-use typo in a comment in control_plane.cpp (from 94ac60df) because this PR touches that file; corrected so the hook passes.

  AI Assistance Disclosure

  - [ ] No AI tools were used

### DIFF
--- a/mooncake-transfer-engine/tent/include/tent/rpc/rpc.h
+++ b/mooncake-transfer-engine/tent/include/tent/rpc/rpc.h
@@ -63,7 +63,18 @@ class CoroRpcAgent {
    public:
     using Function = std::function<void(const std::string_view & /* request */,
                                         std::string & /* response */)>;
-    Status registerFunction(int func_id, const Function &func);
+
+    // Handlers run on the single io_context (kRpcThreads), so a blocking one
+    // stalls every connection, not just its own.
+    //
+    // offload=false (default): inline, no thread hop. Right for anything that
+    // returns promptly.
+    // offload=true: runs on ylt's blocking executor while the connection
+    // coroutine suspends, keeping the io_context free. Callers see the same
+    // request/response, but ordering within a connection is no longer
+    // guaranteed.
+    Status registerFunction(int func_id, const Function &func,
+                            bool offload = false);
 
     Status start(uint16_t &port, bool ipv6 = false);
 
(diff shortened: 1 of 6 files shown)


alogfans (approved):
LGTM

PR #3405, opened 2026-08-12
TENT: link tent_runtime against mooncake_common

xiaodouzi666:
`libtent_shared.so` uses symbols from `mooncake_common` (e.g.
`mooncake::Environ::Get`) but does not declare the dependency, so with the
default `--as-needed` link policy the DT_NEEDED entry is absent and consumers
that link only `-ltent_shared` fail at load time with undefined symbols.

One-line CMake fix: link `tent_runtime` PUBLIC against `mooncake_common`.

Verified on Ubuntu 22.04 / gcc 11.4: `readelf -d libtent_shared.so` now lists
`libmooncake_common.so` in DT_NEEDED and `ldd -r` reports zero unresolved
symbols; an external consumer (the NIXL Mooncake plugin) links and loads
cleanly against `-ltent_shared` alone.


alogfans (approved):
LGTM

PR #3407, opened 2026-08-12
[Bugfix][TENT] Pace the transfer poll loops and stop leaking their batches

SongOf:
Description

  Three loops poll progressBatch() with nothing between iterations: transferSync(), waitTransferCompletion(), and ProxyManager's transferEventLoop(), which pushes an in-flight chunk straight back onto its queue and picks it up again immediately.

  progressBatch() goes to getBatchStatus(), which takes the engine's progress_mutex_ on every call and runs refillDispatchWindow() inside it. So the spin does not merely burn a core — it repeatedly acquires the global progress lock that every other thread's submit, poll and reclaim path also needs, and by releasing and immediately re-acquiring it lets the spinner cut ahead of threads that are actually making progress. All three loops run on the staging worker threads, so several can spin at once.

  Add nextPollDelay(poll_count) and pause on it between polls. Zero for the first 128 polls — the caller yields instead, which also gives a thread waiting on progress_mutex_ a chance at it — so a transfer that completes quickly keeps its current latency. After that the delay grows exponentially from 1 µs, capped at 200 µs, which bounds how long a completed transfer can go unnoticed. The shift is clamped so a long-running loop cannot overflow it back to zero. On a 20 ms transfer this turns roughly ten thousand polls into a couple of hundred.

  transferEventLoop() is paced by whole sweeps rather than per iteration: it is a work queue, and delaying each pop would slow a queue that is actually draining. It backs off only after a full pass left every chunk in the state it started in. The switch itself is unchanged.

  Separately, transferSync() and waitTransferCompletion() leave through CHECK_STATUS early returns — on submitTransfer() and on progressBatch() failure — that skipped freeBatch() and leaked the Batch slab along with whatever SubBatches it had acquired. Free it from a scope guard instead. The guard mirrors the existing BatchRef but calls freeBatch() rather than releaseBatch(); the success paths keep their current behavior, including which one propagates the free error and which one ignores it.

  Module
  
  - [ ] CI/CD
  - [ ] Docs
  - [ ] Other

  Type of Change

  - [x] Bug fix
  - [ ] New feature
  - [ ] Refactor
  - [ ] Breaking change
  - [ ] Documentation update
  - [ ] Performance improvement
  - [ ] Other

  How Has This Been Tested?

  Test commands:
  ctest -R "tent_poll_backoff_test|tent_engine_failover_e2e_test" --output-on-failure
  ctest --output-on-failure

  Test results:
  - [x] Unit tests pass

  nextPollDelay is a free function taking only the poll count, so tent_poll_backoff_test pins its shape down without standing up an engine or timing a sleep: hot phase, first sleeps, monotonicity, the cap, and saturation at
  UINT64_MAX. Replacing it with a constant zero fails four of the five cases.

  EngineFailoverE2E.TransferSyncFreesBatchWhenSubmitFails covers the leak: a request over an unregistered buffer makes submitTransfer() fail, and the alive-batch count must come back to where it started. Disabling the guard
  fails it. This exercises the submit-failure exit; the poll-failure exit leaves through the same guard but is not separately covered, as forcing it would need a new fault mode in FakeTransport.

  The loops themselves are already covered by existing tests — EngineFailoverE2E.TransferSyncUsesProgressBatchWhenPollDisabled and WaitTransferCompletionUsesProgressBatchWhenPollDisabled drive both against a real engine; both
  still pass.

  aliveBatchCountForTest() is the only production code here that exists for the test, alongside the existing swapTransportForTest() hook on the same class.

  Full TENT suite: 64 tests, 61 pass. The 3 failures (tcp_transport_test, transfer_metadata_test, memory_location_test) are pre-existing container-environment issues — no etcd metadata server, and no CAP_SYS_NICE for mbind —
  unrelated to this change.

  Checklist
  
  AI Assistance Disclosure

  - [ ] No AI tools were used

### DIFF
--- a/mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
+++ b/mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
@@ -16,6 +16,7 @@
 #include "tent/runtime/control_plane.h"
 
 #include <algorithm>
+#include <chrono>
 #include <cstdlib>
 #include <fstream>
 #include <limits>
@@ -23,6 +24,7 @@
 #include <optional>
 #include <random>
 #include <stdexcept>
+#include <thread>
 
 #include "tent/common/config.h"
 #include "tent/common/status.h"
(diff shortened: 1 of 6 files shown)


mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
```diff
@@ -2354,37 +2356,90 @@ void TransferEngineImpl::notifyRuntimeQueueReady() {
...
         CHECK_STATUS(progressBatch(batch_id, xfer_status));
         if (xfer_status.s != PENDING) {
-            freeBatch(batch_id);
+            (void)batch_guard.reset();
             return xfer_status.s == COMPLETED
                        ? Status::OK()
                        : Status::InternalError(
                              "Transfer failed: " +
                              std::to_string((int)xfer_status.s));
         }
```
alogfans, line 2425: waitTransferCompletion() intentionally ignores the return value of batch_guard.reset(), while transferSync() propagates it. Since that difference is subtle, could we add a short comment here to make it clear this is preserving the existing behavior rather than an accidental omission?

alogfans (approved):
Overall LGTM

mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
SongOf: Good catch — added a comment at that line. Confirmed it's intentional: before this change waitTransferCompletion() called bare freeBatch(batch_id); (return value discarded) while transferSync() used CHECK_STATUS(freeBatch(batch_id)). The (void) just makes that pre-existing difference explicit — callers of this path get the transfer outcome, not the cleanup outcome. Pushed in the latest revision.

staryxchen (approved):
LGTM

PR #3410, opened 2026-08-12
[Bugfix][TENT] Lock ProxyManager's stage buffer map

SongOf:
Description
  
  ProxyManager::stage_buffers_ is a plain unordered_map with no synchronisation, reached from the kShards staging worker threads via StageBufferCache::allocateLocal → pinStageBuffer, and from the RPC thread via the Pin/Unpin handlers.

  The chunk bitmaps inside each entry are atomic_flags and are fine. The map holding them is not: allocateStageBuffers inserts with operator[] while pinStageBuffer and unpinStageBuffer call find() or walk the whole map. An insert that rehashes under another thread's traversal is undefined behaviour, not a benign stale read. Triggering it needs two different locations pinned for the first time concurrently — the steady state only reads existing entries — which is presumably why it has not surfaced.

  This predates the recent RPC threading work: each shard's runner() builds its own StageBufferCache and calls pinStageBuffer directly, so the single RPC thread never serialised it.

  Guard the map with a shared_mutex. Reads dominate and only touch atomics inside an entry, so pinStageBuffer holds it shared across both the lookup and the bitmap scan, and unpinStageBuffer across its walk — workers still scan concurrently and the bitmap stays lock-free. Exclusive is taken only to insert or erase, at most once per location.

  allocateStageBuffers holds the exclusive lock across the allocation and registration. Those are slow but run once per location, and racing there instead would mean registering the same hundreds of megabytes twice and discarding one. shared_mutex cannot upgrade in place, so pinStageBuffer drops its shared lock before calling it and re-acquires afterwards; allocateStageBuffers re-checks under its own exclusive lock, which makes losing that race harmless.

  Lock ordering is one-way: callers hold no engine lock on entry (lockStageBuffer/unlockStageBuffer forward directly), and the registration path never calls back into ProxyManager.

  Module

  - [ ] CI/CD
  - [ ] Docs
  - [ ] Other

  Type of Change

  - [x] Bug fix
  - [ ] New feature
  - [ ] Refactor
  - [ ] Breaking change
  - [ ] Documentation update
  - [ ] Performance improvement
  - [ ] Other

  How Has This Been Tested?

  Test commands:
  ctest -R tent_stage_buffer_concurrency_test --output-on-failure
  ctest --output-on-failure

  Test results:
  - [x] Unit tests pass

  Adds tent_stage_buffer_concurrency_test, which drives a real ProxyManager from several threads: concurrent first pins of distinct locations (the insert-versus-find shape), concurrent pins of one location (each caller must get its own chunk), and pins overlapping unpins (which walks the map). It uses 4 KiB × 8 buffers rather than the 4 MiB × 64 default to stay cheap, and needs no RDMA device.

  A data race is not proven by a passing run, and this was not verified under ThreadSanitizer. The build container's seccomp profile blocks the personality(ADDR_NO_RANDOMIZE) call libtsan requires at startup, so the TSan run could not be completed. The tests establish functional correctness; the race argument rests on review of the five access sites (deconstruct, allocateStageBuffers, freeStageBuffers, pinStageBuffer, unpinStageBuffer). A TSan job in CI would close this properly.

  Other TENT tests are unaffected; the usual 3 container-environment failures (tcp_transport_test, transfer_metadata_test, memory_location_test — no etcd metadata server, no CAP_SYS_NICE for mbind) are unrelated to this change.

  Checklist

  AI Assistance Disclosure

  - [ ] No AI tools were used

### DIFF
--- a/mooncake-transfer-engine/tent/include/tent/runtime/proxy_manager.h
+++ b/mooncake-transfer-engine/tent/include/tent/runtime/proxy_manager.h
@@ -15,6 +15,8 @@
 #ifndef PROXY_MANAGER_H_
 #define PROXY_MANAGER_H_
 
+#include <shared_mutex>
+
 #include "tent/common/types.h"
 #include "tent/common/status.h"
 #include "tent/common/concurrent/thread_pool.h"
@@ -100,6 +102,11 @@ class ProxyManager {
     const size_t chunk_size_;
     const size_t chunk_count_;
     TransferEngineImpl* impl_;
+    // Guards the map, not the chunk bitmaps in it: those are atomic_flags and
+    // stay lock-free under a shared lock. Exclusive only to insert or erase an
+    // entry, at most once per location. Reached by the kShards staging workers
+    // via StageBufferCache and by the RPC thread via Pin/Unpin.
+    mutable std::shared_mutex stage_buffers_mutex_;
     std::unordered_map<std::string, StageBuffers> stage_buffers_;
     std::atomic<bool> running_;
     struct WorkerShard {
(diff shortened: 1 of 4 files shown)


alogfans (approved):
LGTM

PR #3411, opened 2026-08-12
[Bugfix][Store] Fix MasterScenario tenant quota assertions after lock-free rework

SongOf:
Description

  main currently does not compile. #3327 added the MasterScenario tenant quota DSL against TenantQuotaSnapshot::used_bytes / reserved_bytes; #3162 landed afterwards and collapsed both fields into a single charged_bytes when it made charge/release lock-free. Neither PR's CI observed the other.

  master_scenario.cpp:415: error: 'const struct mooncake::TenantQuotaSnapshot' has no member named 'used_bytes'; did you mean 'charged_bytes'?

  Quota is now charged once at PutStart and held until release, so a snapshot cannot distinguish reserved from committed bytes. This collapses the DSL's Uses()/Reserves() into a single Charges(). Every call site previously set exactly one of the two to a non-zero value, so the assertions keep their original strength; the pre/post PutEnd pair now verifies that completing an object takes no second charge.

  Tests only — no production code changed.

  Module

  - [x] Mooncake Store (mooncake-store)

  Type of Change

  - [x] Bug fix

  How Has This Been Tested?

  Test commands:
  cmake --build build --target master_service_evict_scenario_test
  ctest --test-dir build -R master_service_evict_scenario_test

  Test results:
  - [ ] Unit tests pass — not run locally; mooncake-store requires Linux headers and the local Docker daemon was unresponsive. Relying on CI, which is exactly what this PR unblocks.
  - [x] clang-format --dry-run -Werror clean on all three files
  - [x] pre-commit run --files <the 3 files> — all hooks pass

  Checklist

  - [x] I have added tests to prove my changes are effective (repairs existing tests)
  - [ ] For changes >500 LOC (not applicable, ~30 LOC)

  AI Assistance Disclosure

### DIFF
--- a/mooncake-store/tests/master_scenario.cpp
+++ b/mooncake-store/tests/master_scenario.cpp
@@ -411,10 +411,8 @@ MasterScenario& MasterScenario::Then(TenantQuotaSpec tenant_quota) {
     }
 
     const auto matches = [&tenant_quota](const TenantQuotaSnapshot& value) {
-        return (!tenant_quota.used_bytes.has_value() ||
-                value.used_bytes == *tenant_quota.used_bytes) &&
-               (!tenant_quota.reserved_bytes.has_value() ||
-                value.reserved_bytes == *tenant_quota.reserved_bytes);
+        return !tenant_quota.charged_bytes.has_value() ||
+               value.charged_bytes == *tenant_quota.charged_bytes;
     };
     const auto deadline =
         std::chrono::steady_clock::now() + tenant_quota.eventual_timeout;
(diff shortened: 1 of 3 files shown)


Icedcoco (approved):

ykwd:
Thanks for the catch. Could you resolve the conflict?

SongOf:
> Thanks for the catch. Could you resolve the conflict?
Thanks for your review. The upstream commit 6cc13f0e (PR #2683 "[Store] Add DFS replica support with POSIX backend") already contains exactly the same fix. I'll close this PR.

PR #3422, opened 2026-08-13
[Bugfix][Store] Handle BatchEvict OpLog failures

Icedcoco:
## Description

Part of #3139.

`BatchEvict()` previously represented OpLog reservation and submission with a boolean result. This caused several problems:

- the original `ErrorCode` was lost;
- the scan continued after a reservation failure;
- stopped, unavailable, invalid, and terminal writer failures could trigger repeated eviction attempts as if they were transient backpressure;
- replicas were marked `REMOVED` before `Commit()`, but a rejected commit has no durable callback to finalize or restore them;
- eviction success/failure metrics were updated through duplicated branches.

This PR:

- preserves the exact reservation or submission `ErrorCode`;
- stops the current eviction scan after the first OpLog failure;
- automatically retries only `TASK_PENDING_LIMIT_EXCEEDED`;
- does not arm a busy retry for non-transient writer failures;
- rolls replicas changed by a rejected commit back from `REMOVED` to `COMPLETE`;
- records eviction and memory-eviction metrics through one common outcome block.

The rollback is intentionally narrow: it only permits `REMOVED -> COMPLETE` for replicas changed by the failed submission attempt.

## Relationship to #3141

PR #3141 combines several independent recovery and eviction fixes, including allocator recovery, `PutStart` eviction triggering, segment remount/recovery behavior, and `BatchEvict` OpLog failure handling.

This PR is the focused replacement for only the `BatchEvict` OpLog reservation/submission failure-handling portion of #3141. It does not replace or include #3141's allocator, `PutStart`, remount, lease, or recovery-eligibility changes.

Once this PR lands, the overlapping `BatchEvict` OpLog changes can be removed from #3141 while its other recovery fixes remain independent.

The tests reuse the gated writer and callback controls merged in #3333. This PR does not duplicate those helpers. It also does not implement the ordered-writer terminal-state machinery from #3204; terminal and other non-queue errors are handled through their existing `ErrorCode`.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

The regression coverage verifies:

- queue-full reservation failure stops after the first failure and retains retry;
- unavailable writer failure stops without scheduling retry;
- rejected commit restores affected replicas to `COMPLETE`;
- persisted removal records are observable before durable-finalize callbacks;
- replicas are released immediately after gated callbacks run;
- failed reservation ownership releases its batch slot through RAII.

**Test commands:**

```bash
cmake --build build --target master_service_ha_test -j32

env TMPDIR=/tmp/mooncake-pr3141-fix-batch-eviction-oplog-failure \
  LSAN_OPTIONS=detect_leaks=0 \
  GLOG_minloglevel=1 \
  ./build/mooncake-store/tests/master_service_ha_test \
  --gtest_brief=1

env TMPDIR=/tmp/mooncake-pr3141-fix-batch-eviction-oplog-failure \
  LSAN_OPTIONS=detect_leaks=0 \
  GLOG_minloglevel=1 \
  ./build/mooncake-store/tests/master_service_ha_test \
  --gtest_filter='MasterServiceHATest.BatchEvictStopsAfterFirstOpLogReservationFailure:MasterServiceHATest.BatchEvictStopsWithoutRetryWhenOpLogWriterUnavailable:MasterServiceHATest.BatchEvictCommitFailureRestoresRemovedReplicas:MasterServiceBatchRecordE2ETest.RemoveDurableCallbackReleasesResources:MasterServiceHATest.RemoveHidesBeforeDurableAndReleasesAfterFinalize:MasterServiceHATest.BatchRemoveFinalizesEachObjectAfterDurable:MasterServiceHATest.RemoveAllFinalizesAfterDurable:MasterServiceHATest.BatchReplicaClearAllReleasesAfterDurable:MasterServiceHATest.BatchReplicaClearSegmentReleasesAfterDurable:MasterServiceHATest.EvictDiskReplicaReleasesLocalDiskAfterDurable' \
  --gtest_repeat=100 \
  --gtest_break_on_failure \
  --gtest_brief=1

pre-commit run --files \
  mooncake-store/include/replica.h \
  mooncake-store/tests/ha/master_service_ha_test.cpp

git diff --check HEAD^ HEAD
```

**Test results:**

- [x] Unit tests pass: full `master_service_ha_test` suite, 70/70
- [x] Focused failure and gated-callback tests pass: 1000/1000 across 100 iterations
- [x] Post-format targeted regression tests pass: 3/3

## Checklist

- [x] I have formatted my code using the repository formatting hook
- [x] Documentation is not required for this internal failure-handling change
- [x] An RFC is not required because the production change is below 500 LOC

## AI Assistance Disclosure

- [ ] No AI tools were used
- [x] AI tools were used

Codex assisted with code-path analysis, implementation, test development, and test execution. The human submitter will review every changed line before submission and remains responsible for understanding and defending the change.

### DIFF
--- a/mooncake-store/src/master_service.cpp
+++ b/mooncake-store/src/master_service.cpp
@@ -10086,16 +10086,14 @@ void MasterService::BatchEvict(double evict_ratio_target,
         return 0;
     };
 
-    // HA strong-consistency: persist the post-eviction state BEFORE the
-    // helper mutates `metadata`. Returns true on success (or when HA is
-    // disabled). On false, the caller must NOT call try_evict_or_offload
-    // and must NOT erase the metadata entry — local state must stay in
-    // sync with what was published.
+    // HA strong-consistency: persist the post-eviction state before the
+    // helper mutates metadata. Keep the removed IDs local until submission
+    // succeeds so a consumed reservation cannot strand REMOVED replicas.
     auto persist_evict_oplog_or_skip =
         [&, this](const TenantId& tenant_id, const std::string& key,
-                  ObjectMetadata& metadata) -> bool {
-        if (!enable_oplog_ || !ordered_oplog_writer_) {
-            return true;
+                  ObjectMetadata& metadata) -> tl::expected<void, ErrorCode> {
+        if (!enable_oplog_) {
+            return {};
         }
 
         // Predict the descriptor list after evict_replicas() runs:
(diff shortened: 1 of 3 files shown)


mooncake-store/src/master_service.cpp
```diff
@@ -10022,16 +10022,29 @@ void MasterService::BatchEvict(double evict_ratio_target,
         return 0;
     };
 
-    // HA strong-consistency: persist the post-eviction state BEFORE the
-    // helper mutates `metadata`. Returns true on success (or when HA is
-    // disabled). On false, the caller must NOT call try_evict_or_offload
-    // and must NOT erase the metadata entry — local state must stay in
-    // sync with what was published.
+    struct EvictOpLogSubmissionResult {
```
Aionw: Do we need four states here? All callers only inspect failed() and error, while the reservation/submission distinction is already handled by logging inside this helper. Could this return tl::expected<void, ErrorCode> instead, treating a disabled OpLog as success? That would also remove the submission_failed adapter below.

Icedcoco: Agreed. The four-state result was carried over from the earlier #3141 design, where reservation and submission outcomes were exposed so the caller could decide whether replicas had already been marked `REMOVED` and needed rollback.

That distinction is now fully owned by this helper: a reservation failure leaves metadata unchanged, while a submission failure rolls the affected replicas back before returning. Callers therefore only need success/failure plus the exact `ErrorCode`.

I changed the helper to return `tl::expected<void, ErrorCode>` and removed the `submission_failed` adapter. Disabled OpLog and a successful submission return success; both failure paths return their original error. Updated in 23b9fecd.

mooncake-store/src/master_service.cpp
```diff
@@ -10040,76 +10038,61 @@ void MasterService::BatchEvict(double evict_ratio_target,
...
+                         << ", err=" << static_cast<int>(reservation.error())
+                         << ", stopping eviction scan";
+            return tl::make_unexpected(reservation.error());
         }
-
+        std::vector<ReplicaID> removed_ids;
+        metadata.VisitReplicas(is_evictable_memory_replica,
+                               [&removed_ids](Replica& replica) {
+                                   removed_ids.push_back(replica.id());
+                                   replica.mark_removed();
```
Aionw, line 10117: Could we also cover the asynchronous terminal-failure path here? The rollback below only runs when Commit rejects synchronously. If Commit accepts this entry and the writer later hits a fencing or other non-retryable error while persisting the batch, the durable callback is dropped when the writer exits, while these replicas remain REMOVED. MasterService currently constructs OrderedOpLogWriter without a terminal callback or fail-stop hook, so the active service can keep the object unreadable and its memory unreleased indefinitely. Please either fail-stop the service on writer terminal state or retain enough rollback state to restore non-durable evictions, and add a test that lets BatchEvict commit before making the backend terminal.

Icedcoco: Agreed. This asynchronous terminal-failure path is already considered in the design. The intended recovery is service-level fail-stop, rather than locally rolling back entries after `Commit()` has accepted them, because the durability outcome can be ambiguous.

This PR intentionally covers the synchronous `Commit()` rejection path only: it restores replicas marked by the rejected attempt and preserves the exact error. The writer terminal-state propagation and supervisor/service fail-stop handling will be implemented in follow-up development. We will keep the PR scope and description explicit about this boundary.

Aionw:
Overall LGTM. But we need to rerun ci first and create an issue to track 
>The writer terminal-state propagation and supervisor/service fail-stop handling will be implemented in follow-up development

Aionw (approved):

ykwd (approved):

Issue #3322, opened 2026-08-07
[RFC]: MPComm Transport for TENT — an RDMA-native multi-rail transport backend


c-guo16:
### Summary

This RFC proposes adding **MPComm** as a TENT transport backend, registered as `TransportType::MPCOMM`.

MPComm (Memory Pool Communication) is an RDMA-native data-plane library for large-scale
heterogeneous memory pools, open-sourced by the Tencent Astral Network Team under Apache-2.0:
<https://github.com/Tencent/UCL-MPComm>. It drives multiple RDMA NICs concurrently with two-level
load balancing (across NICs, and across QPs within a NIC) and NUMA-aware worker placement, exposing
one-sided `put`/`get` primitives.

Data path:

```
TENT Request (WRITE/READ)
  -> MpcommTransport::submitTransferTasks()
  -> MPComm putAsync() / getAsync()        (slicing, NIC/QP selection, workers are internal to MPComm)
  -> libmpcomm -> libibverbs
```

Validated scope is **host DRAM and GPU device memory over RoCE/IB**, in both directions and in
mixed combinations. GPU memory is registered through `nvidia-peermem`, the same mechanism
`RdmaTransport` depends on for GPU-Direct. The implementation is complete and validated; this RFC
exists because the change exceeds the 500-LOC threshold in `CONTRIBUTING.md`
("For major architectural changes (>500 LOC excluding tests), we would expect a GitHub issue (RFC)").

Related roadmap: #1058 (Transfer Engine NEXT).

### Motivation

The obvious question is *why a second RDMA transport when `RdmaTransport` already exists*. Three
reasons:

**1. Measured throughput advantage over the existing TENT RDMA path.**
Numbers published in the MPComm repository (2 nodes, single NUMA, 4x CX-7, dual-plane 2x200G bond
per NIC, batch=4, 1GB block):

| Path | Platform | MPComm PUT/GET | Baseline PUT/GET | Ratio |
|---|---|---|---|---|
| Zero-copy | AMD Turin / RTX PRO 5000 | 195.7 / 195.7 GB/s | Mooncake TENT 142.2 / 142.3 | ~1.38x |
| Zero-copy | AMD Genoa / NVIDIA H20 | 165.7 / 152.3 GB/s | Mooncake TENT 122.0 / 111.7 | ~1.36x |
[...]

PR #3423, opened 2026-08-13
MPComm Transport for TENT — an RDMA-native multi-rail transport backend

c-guo16:
## Description

Adds a transport backend for the TENT runtime built on **UCL-MPComm** (Unified Communication
Library - Memory Pool Communication), an RDMA-native, multi-rail data plane from the Tencent Astral
Network Team, published at [Tencent/UCL-MPComm](https://github.com/Tencent/UCL-MPComm) under
Apache-2.0. The transport is registered under the name `mpcomm`, and the library is shortened to
MPComm in the rest of this description and in the code.

Design discussed in RFC #3322, including the open questions and the follow-up review items; this PR
is the implementation of what was agreed there.

**Why another RDMA transport.** UCL-MPComm keeps slicing, NIC selection, QP selection and its worker
threads inside the library, and drives multiple rails from one transfer. On the same hardware its own
benchmarks report **~195.7 GB/s** against **~142.2 GB/s** for TENT's RDMA transport (≈1.38x, AMD Turin + 4x
CX-7; ≈1.36x on H20) — numbers from the UCL-MPComm repository, reproducible with the commands in the
doc added here. In this integration `tebench` reaches >390 GB/s aggregate across two nodes.

### Structure

Everything that reaches MPComm goes through a single injectable boundary, so the TENT-side logic
builds and tests without the provider. This follows `UrmaAdapter` and `TpuPjrtShim`.

| Unit | Responsibility | Needs `mpcomm.h` |
|---|---|:---:|
| `mpcomm_adapter.{h,cpp}` | The provider boundary. A thin pass-through — MPComm owns its own scheduling, so there is none to model. Compiles to an unavailable stub without the provider | **yes, only here** |
| `mpcomm_peer_registry.{h,cpp}` | Peer cache and endpoint attribute parsing | no |
| `mpcomm_task_mapping.{h,cpp}` | Request and completion mapping | no |
| `mpcomm_transport.{h,cpp}` | What needs the TENT runtime: SegmentID resolution, capabilities, batch driving | no |

### Notable behaviour

- **Peers are keyed by MPComm host id**, not `SegmentID`. MPComm keys connections by host id and
  cannot close one, and a second `connect()` replaces its connection record wholesale — dropping the
  remote keys it carries and leaking the previous queue pairs. Reopening a segment must therefore
  reuse the existing connection.
- **Connection and remote keys are tracked separately.** A failed key query leaves the peer in
  `CONNECTED_NO_KEYS` and the next request retries the query alone. Keys are refetched when the
  peer's published buffer set grows, compared by coverage rather than equality — descriptors are
  cached per thread and expire independently, so an equality test makes threads alternately
  invalidate each other and issue a blocking query each time.
- **GPU capabilities are gated on `nvidia_peermem`**, the same probe `RdmaTransport` uses: MPComm
  registers device memory through the ordinary `ibv_reg_mr` path and has no dma-buf fallback. A host
  without the module is excluded during selection rather than failing at registration.
- **`MPCOMM` is appended as wire value 12**, leaving `UB` at 11. Assertions cover the C++ enum, the C
  API macro (checked through the conversion function rather than by repeating the constant) and the
  Python binding (`static_assert`).
- **Security.** MPComm's handshake is unauthenticated plain TCP and exchanges buffer addresses and
  rkeys, which is everything an RDMA operation needs. The doc states that this port belongs at the
  same trust level as the fabric itself.

### Known limitations

Recorded rather than worked around; none blocks the DRAM or VRAM data path.

1. The first transfer to a peer performs the MPComm handshake **inline on the submitting thread**.
   `RdmaTransport` connects lazily on the transfer path too, but its handshake runs over an RPC with
   a timeout, whereas MPComm's is a bare TCP exchange with none. Adding one requires a change in the
   library.
2. **A restarted peer is not recovered automatically.** With no way to close a connection, dropping
   and rebuilding one would leak queue pairs and blank the keys concurrent transfers are using, so
   the transport keeps the cached peer instead.
3. **The listener binds `INADDR_ANY` and this is not configurable** (MPComm's own behaviour). The
   advertised endpoint follows `rpc_server_hostname`; a setting for advertising an address separate
   from the RPC one is not implemented.
4. No cancellation, bandwidth estimation or NIC load statistics, and the `qp_pool` and
   progress-notification hooks are not consumed. These stay at the base-class defaults, so they
   report as unsupported rather than being assumed by policy code.
5. **CI does not cover the data path** — that needs RDMA devices and libmpcomm. See testing below for
   how the boundary is covered instead.
6. The MPComm ABI is pinned at configure time only; there is no run-time version check, since the
   library exposes no version query.

### Out of scope

- The legacy Transfer Engine transport path. This is TENT-only, so `MOONCAKE_PROTOCOL=mpcomm` and
  `transfer_engine_bench --protocol=mpcomm` are intentionally unsupported.
- CXL, file and other non-RDMA segment types.
- Changes to UCL-MPComm itself.

### Review focus

- `mpcomm_peer_registry.cpp` — the only non-trivial concurrency here: claim-then-connect ownership,
  the wait loop, and the RAII cleanup that has to leave the entry in a state later callers can make
  progress from.
- `mpcomm_adapter.h` — whether the boundary is drawn where you would want it for a provider that
  brings its own scheduling.
- `common.cmake` — the MPComm version check, and the scoped `CMP0144` policy: `find_package(mpcomm)`
  consults an upper-case `<PACKAGENAME>_ROOT`, which for this package is exactly the `MPCOMM_ROOT`
  the build already uses, so CMake 3.27+ warns unless the policy is set explicitly.

## Module

- [ ] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

Three layers: a hardware-free suite that runs anywhere, a hardware suite, and a two-node benchmark.

**Test commands:**
```bash
# 1. Boundary tests - no RDMA device, no libmpcomm, built in every configuration
cmake .. -DUSE_TENT=ON -DBUILD_UNIT_TESTS=ON
make -j tent_mpcomm_boundary_test
ctest -R tent_mpcomm_boundary_test --output-on-failure
#   => 26 tests from 4 suites, 0.10 s, all passed

# 2. Real data path - requires RDMA devices and libmpcomm
cmake .. -DUSE_TENT=ON -DUSE_MPCOMM=ON -DUSE_CUDA=ON -DBUILD_UNIT_TESTS=ON \
         -DMPCOMM_ROOT=<prefix>
make -j tent_mpcomm_transport_test
ctest -R tent_mpcomm_transport_test --output-on-failure
#   => passed in 1.61 s (forks a target, WRITE then READ, verifies the payload)

# 3. Two nodes, RoCE, multi-NIC, dual NUMA
./tebench --backend=tent --xport_type=mpcomm --tent_transport_hint=mpcomm \
          --metadata_type=p2p --seg_type=DRAM ...
#   => >390 GB/s aggregate; also run with --seg_type=VRAM on one and both sides,
#      and with --check_consistency=true
```

The boundary suite drives the transport against an injected `MpcommAdapter` and covers endpoint
publication and parsing, single-flight connection under concurrency, key-query retry over an existing
connection, refreshing keys after a peer registers memory, the WRITE/READ mapping, short-transfer and
error handling, releasing each handle exactly once, and teardown. It deliberately does not model
MPComm's own scheduling.

**Test results:**
- [x] Unit tests pass

Manual: two-node runs over DRAM and VRAM segments (both sides VRAM, and one side only), payload
verification enabled, plus a single-node RDMA loopback test.

## Checklist

- [x] For changes >500 LOC: I have filed an RFC issue — #3322

## AI Assistance Disclosure

- [ ] No AI tools were used

An AI coding assistant was used for drafting code, documentation and tests, and for reviewing the
result. Every change has been read, built and tested on real hardware by the submitter, who is
responsible for defending all of it.

### DIFF
--- a/mooncake-transfer-engine/tent/include/tent/transfer_engine.h
+++ b/mooncake-transfer-engine/tent/include/tent/transfer_engine.h
@@ -111,6 +111,7 @@ typedef struct tent_notifi_info tent_notifi_info;
 #define TRANSPORT_SUNRISE_LINK (9)
 #define TRANSPORT_TPU (10)
 #define TRANSPORT_UB (11)
+#define TRANSPORT_MPCOMM (12)
 
 struct tent_memory_options {
     char location[64];
--- a/docs/source/design/transfer-engine/index.md
+++ b/docs/source/design/transfer-engine/index.md
@@ -543,6 +543,14 @@ kunpeng_ub_transport
 sunrise_link_transport
 :::
 
+## MPComm Transport Component
+
+::::{toctree}
+:maxdepth: 1
+
+mpcomm_transport
+::::
+
 ## Benchmark and Tuning Guide
 
 :::{toctree}
--- a/docs/source/design/transfer-engine/mpcomm_transport.md
+++ b/docs/source/design/transfer-engine/mpcomm_transport.md
@@ -1,0 +1,558 @@
+# MPComm Transport
+
+## Overview
+
+UCL-MPComm (Unified Communication Library - Memory Pool Communication) is a high-performance RDMA
+communication library for heterogeneous memory pooling, developed by the Tencent Astral Network
+Team. In Mooncake it is integrated as a transport backend of the TENT transfer framework and
+registered as `MPCOMM`; the library is shortened to MPComm throughout this document.
+
+- Upstream repository: <https://github.com/Tencent/UCL-MPComm>
+
+MPComm drives multiple RDMA NICs concurrently and performs two-level load balancing (across NICs
+and across QPs within a NIC), with NUMA-aware worker placement. It exposes one-sided
+`put`/`get` primitives that the transport maps onto TENT's `WRITE`/`READ` requests.
+
+At runtime, `TransferEngineImpl` loads MPComm Transport when `USE_MPCOMM` is enabled at build time
+and `transports/mpcomm/enable=true` is set in the configuration.
+
+> **TENT only.** This transport is implemented for the TENT runtime
+> (`mooncake-transfer-engine/tent/`) and is not available through the legacy Transfer Engine
+> transport path. Builds must therefore enable `USE_TENT` together with `USE_MPCOMM`, and
+> benchmarks must run with `--backend=tent`.
+
+### Code Structure
+
+The transport is split so that everything reaching MPComm goes through one interface. That keeps the
+provider out of most translation units, which is what makes the TENT-side logic buildable and
+testable without libmpcomm - the same arrangement `TpuPjrtShim` and `UrmaAdapter` use for their
+providers.
+
+| Unit | Responsibility | Needs `mpcomm.h` |
+|------|----------------|:----------------:|
+| `mpcomm_adapter.{h,cpp}` | The MPComm boundary. A thin pass-through: MPComm owns its slicing, NIC/QP selection and worker threads, so there is no scheduling to model here. Compiles to an unavailable stub when `USE_MPCOMM` is off | **yes** (only here) |
+| `mpcomm_peer_registry.{h,cpp}` | Peer cache and endpoint attribute parsing: which peers are connected, whose keys are current, and who may talk to a peer at any moment | no |
+| `mpcomm_task_mapping.{h,cpp}` | Request and completion mapping: WRITE/READ onto put/get, provider outcome onto `TransferStatus`, handle release | no |
+| `mpcomm_transport.{h,cpp}` | What needs the TENT runtime: resolving a `SegmentID` to a peer, reading the endpoint it advertises, driving batches, publishing capabilities | no |
+
+`MpcommTransport` also accepts an adapter through a second constructor, which is how tests
+substitute the provider.
+
+### Transfer Pipeline
+
+1. **Init**: `MPComm::init()` is called with the local segment name as the MPComm host id, the
+   RDMA device list derived from the TENT `Topology`, and a TCP port used for MPComm's own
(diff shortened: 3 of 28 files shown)


mooncake-transfer-engine/tent/include/tent/transfer_engine.h
```diff
@@ -111,6 +111,7 @@ typedef struct tent_notifi_info tent_notifi_info;
 #define TRANSPORT_SUNRISE_LINK (9)
 #define TRANSPORT_TPU (10)
 #define TRANSPORT_UB (11)
+#define TRANSPORT_MPCOMM (12)
```
alogfans, line 114: Is this conflict with the current code base?

c-guo16: No — 12 is free on current main. Checked against 1a49975d:

* transfer_engine.h stops at #define TRANSPORT_UB (11)
* types.h has ..., TPU, UB, kNumTransportTypes, so UB is 11 and the sentinel is currently 12
* git merge-tree HEAD upstream/main reports no conflicts

This PR keeps UB at 11, appends MPCOMM as 12, and the sentinel moves 12 → 13. That last part means your existing WireValuesRemainStableWithUbAppended (which asserts kNumTransportTypes == 12) needed updating — done in the same commit.

alogfans (approved):

Issue #3397, opened 2026-08-12
[RFC]: Decouple NVMe-oF Replicas from SPDK behind an Initiator Interface


LujhCoconut:
# RFC: Decouple NVMe-oF Replicas from SPDK behind an Initiator Interface

- Related:
  - Issue [#3131](https://github.com/kvcache-ai/Mooncake/issues/3131) — NoF writes fail: user buffers never registered with SPDK's RDMA translation table
  - PR [#3251](https://github.com/kvcache-ai/Mooncake/pull/3251) — block-unaligned I/O for NoF replicas (SGL-based); still open as of 2026-08-12
  - RFC [#3360](https://github.com/kvcache-ai/Mooncake/issues/3360) — Refactor Mooncake Store segment management around SegmentPool, esp. sub-issue [#3364](https://github.com/kvcache-ai/Mooncake/issues/3364) (Integrate NoF mounted regions into SegmentPool). Orthogonal and independent of this RFC; see §4.
  - Companion docs: `nof-call-chains.md` (verified call chains + implicit contracts C1–C16), `spdk-nof-review.md` (review findings M1–M10 / E1–E4), `nof-spdk-impl-plan.md` (detailed implementation plan), `nof-refactor-diagrams.md` (full mermaid diagram set)

---

## 1. Summary

Today "NoF replica" and "SPDK" are the same thing in Mooncake Store: every component on the NoF data and control plane calls SPDK C APIs through a global singleton (`SpdkWrapper`), and SPDK types (`spdk_nvme_qpair`, `spdk_nvme_ns`, `spdk_nvme_cmd_cb`) leak into public headers consumed by the client, master and benchmarks.

This RFC proposes introducing a small, stable abstraction — an `NVMeoFInitiator` interface plus a separate `DmaBufferAllocator` interface — moving all SPDK code into a single `SpdkInitiator` implementation behind those interfaces, and injecting instances instead of reaching for a global singleton. No wire protocol, replica-type, or segment-management changes are involved.

This is a data-plane refactor. A separate community RFC ([#3360](https://github.com/kvcache-ai/Mooncake/issues/3360)) covers control-plane ownership of NoF mounted regions in the master; the two are orthogonal (§4).

---

## 2. Problem Statement

[...]

PR #3424, opened 2026-08-13
[Store] Decouple NoF data path from SPDK behind neutral interfaces

### LINKED ISSUE #3131: [Bug]: NoF 写入失败：`spdk_rdma_get_translation: No translation for ptr` 是否是我配置的不对导致的这个问题？
### 问题描述

当 Mooncake Store 客户端使用 `ReplicateConfig{replica_num=0, nof_replica_num=1}`（仅 NoF 副本模式）写入对象时，Master 成功在 NVMe-oF SSD 池上分配空间，SPDK NVMe-oF controller 也成功 attach。但实际数据传输失败，因为 SPDK 的 RDMA 传输层找不到用户 buffer 的内存翻译记录。

### 环境信息

- Mooncake 版本：v0.3.12（tag `v0.3.12`，commit `a1ca71f4`）
- 同时验证了 `v0.3.12-pre1-185-ga1ca71f4`（截至 2026-07-27 的最新 main 分支）——**代码完全相同，未修复**
- SPDK：v23.01.1
- RDMA 网卡：ConnectX-6（mlx5_0），RoCE v2
- NVMe-oF target：Samsung PM1753 over RDMA（192.168.3.110:4420-4423）
- LMCache：v0.5.3.dev30（MP 模式，`lmcache server` + `mooncake_store` L2 adapter）

### 复现步骤

1. 启动 Mooncake master + metadata server，注册 NVMe-oF SSD 池（4 × Samsung PM1753 namespace）
2. 启动 Mooncake Store 客户端，配置 `protocol: "rdma"`、`global_segment_size: "0"`、`local_buffer_size: 536870912`
3. 调用 `batch_put_from(key, buffer, size, ReplicateConfig{replica_num=0, nof_replica_num=1})`，其中 `buffer` 指向已通过 `client->register_buffer()` 预注册的 DRAM 内存

### 期望行为

对象数据通过 SPDK RDMA 写入 NVMe-oF SSD 池。

### 实际行为

Master 日志显示分配成功：
```
master_service.cpp:3088] key=..., config=ReplicateConfig: { replica_num: 0, nof_replica_num: 1, ... }, action=put_start_begin
allocator.cpp:261] allocation_succeeded size=33554432 segment=traddr:192.168.3.110 trsvcid:4422 subnqn:nqn.1994-11.com.samsung:nvme:PM1753:2.5-inch:S8FFNG0Y901187 trtype:RDMA
```

SPDK NVMe-oF controller attach 成功：
```
spdk_wrapper.cpp:297] Attached to 192.168.3.110 nqn.1994-11.com.samsung:nvme:PM1753:2.5-inch:S8FFNG0Y901187
```

但数据传输失败：
```
common.c:216: spdk_rdma_get_translation: *ERROR*: No translation for ptr 0x7f1644000000, size 131072
nvme_rdma.c:1461: nvme_rdma_get_memory_translation: *ERROR*: RDMA memory translation failed, rc -22
nvme_rdma.c:2316: nvme_rdma_qpair_submit_request: *ERROR*: nvme_rdma_req_init() failed
transfer_task.cpp:511] work 0, seg 0x7f1398016730 submit io fail
client_service.cpp:2449] Operation for key ... failed: TRANSFER_FAIL (Transfer 0 failed)
```

`ptr 0x7f1644000000` 是用户提供的 buffer（LMCache L1 DRAM，128KB chunk = `MC_NOF_SUBMIT_CHUNK_BYTES`）。该 buffer 已通过 Mooncake Transfer Engine 的 `client->register_buffer()` → `ibv_reg_mr()` 注册，但**未注册到 SPDK NVMe-oF RDMA 的内存映射**。

### 根因分析

`transfer_task.cpp` 的 NoF 写入路径直接将用户 buffer 指针传给 SPDK，没有先注册到 SPDK NVMe-oF RDMA 传输层的内存映射：

```cpp
// transfer_task.cpp:491-503（SpdkNofWorkerPool::workerThread）
void* submit_ptr = reinterpret_cast<void*>(
    reinterpret_cast<char*>(task->ptr) + lba_off * block_size);

int ret = SpdkWrapper::GetInstance().SubmitRequest(
    task->seg_handle, submit_ptr, submit_lba, submit_lba_count, ...);
```

```cpp
// spdk_wrapper.cpp:387（SpdkWrapper::SubmitRequest）
return spdk_nvme_ns_cmd_write(ns, qpair, ptr, lba, lba_count, cb_fn, cb_ctx, 0);
```

SPDK 的 NVMe-oF RDMA 传输层维护着自己独立的 RDMA 内存映射（`nvme_rdma_qpair::rkey_map`），与 Mooncake Transfer Engine 的 RDMA 内存注册**相互独立**。当 `spdk_nvme_ns_cmd_write()` 被调用时，SPDK 内部通过 `nvme_rdma_get_memory_translation()` 在自己的内存映射中查找 buffer。由于用户 buffer 从未注册到 SPDK 的 RDMA context，查找失败返回 `-22 (EINVAL)`。

### Before submitting...

LujhCoconut:
## Description

Related RFC #3397

Decouple Mooncake Store's NoF (NVMe-oF) data path from SPDK behind two neutral interfaces, and fix #3131 along the way.

**Before**: all NoF traffic funneled into the `SpdkWrapper` Meyers singleton (env init + qpair cache + I/O + DMA alloc + probe in one class). SPDK types leaked through its header, and 38 SPDK/DPDK static libs propagated as PUBLIC link requirements to client/master/bench.

**After**:

- Two role interfaces under `include/nof/`: `NVMeoFInitiator` (byte-oriented I/O behind an opaque `NofSegmentHandle`) and `DmaBufferAllocator`.
- SPDK implementations live in the new `mooncake_nof_spdk` static lib (pimpl — SPDK types confined to one `.cpp`, SPDK/DPDK libs linked `PRIVATE`).
- Instances come from `CreateNofRuntime()` — the only `#ifdef USE_NOF` site naming SPDK types — and are injected via `shared_ptr`. The SPDK env initializes lazily on first use and is owned by a refcounted guard.
- `register_buffer` now also registers user buffers with the SPDK RDMA translation table (process-global registry, per-2MB-page refcount, TE rollback on failure) — **fixes #3131**. Requires SPDK ≥ 22.05; a CMake probe warns on older versions.

No wire-protocol / RPC / Python ABI changes. Non-NoF builds keep the exact original behavior (allocator injected as `nullptr` → original protocol → VRAM → `aligned_alloc` order).

```mermaid
flowchart TB
  subgraph B["Before — everything funnels into one singleton"]
    TS1["TransferSubmitter / worker pool"] -->|GetInstance ×6| SW["SpdkWrapper (Meyers singleton):<br/>env init · qpair cache · IO · DMA alloc · probe"]
    RC1["RealClient::setup"] -->|eager InitializeEnv| SW
    CB1["ClientBufferAllocator / utils"] -->|use_spdk_dma flag| SW
    MS1["Master heartbeat probe"] --> SW
    SW -.->|SPDK types in header,<br/>38 SPDK/DPDK libs PUBLIC| L1["leaks into client / master / bench"]
  end
  subgraph A["After — interface split + dependency injection"]
    RC2["RealClient::setup"] -->|CreateNofRuntime:<br/>only USE_NOF ifdef| F["NofRuntime factory"]
    F --> NI["NVMeoFInitiator «interface»"]
    F --> DA["DmaBufferAllocator «interface»"]
    NI -.pimpl.- SI["SpdkInitiator::Impl — SPDK confined<br/>to one .cpp, libs PRIVATE"]
    DA -.-> SA["SpdkDmaAllocator /<br/>SystemDmaAllocator"]
    TS2["TransferSubmitter / NofWorkerPool"] -->|injected shared_ptr| NI
    CB2["ClientBufferAllocator"] -->|injected; nullptr when NoF absent| DA
    MS2["Master heartbeat probe"] --> NI
  end
```

Lifetime and registration fixes riding along:

```mermaid
flowchart LR
  subgraph LIB["mooncake_nof_spdk — the only SPDK-aware code"]
    G["SpdkEnvGuard:<br/>lazy, refcounted,<br/>init/fini serialized"]
    R["NofPageRegistry: process-global,<br/>per-2MB-page refcount,<br/>-EBUSY → external page"]
    Q["ctrlr/ns/qpair cache:<br/>same endpoint → same qpair,<br/>connect outside mutex"]
    G --- R
    G --- Q
  end
  RB["RealClient::register_buffer"] -->|1. TE register OK| R
  R -->|2. RegisterMemory fail → roll back TE| RB
  W["NofWorkerPool"] -->|SubmitIO bytes /<br/>PollCompletion| Q
```

Also fixed in passing: `io_count` was a raw pointer to a worker-thread stack variable (now `shared_ptr<atomic<int64_t>>`); a failed `SubmitIO` leaked its sub-task pool slot (now returned); the probe context recycle race; `OpenSegment` blocked a global mutex during NVMe-oF connect.

### Upgrade notes for existing NoF users

No breaking configuration or API changes: build flags (`-DUSE_NOF=ON`), existing env vars (`MC_NOF_*`, `MC_NVME_*`, `MC_STORE_NUMA_SOCKET_ID`), transport strings, and the Python ABI are all unchanged. One optional env var is added. Behavioral differences to be aware of:

- **New optional `MC_NOF_BACKEND`** (`spdk` by default, case-insensitive): selects the NoF backend pair. Set it to `none` to disable NoF at runtime in a `USE_NOF` build — one binary can be deployed on nodes without a working SPDK/hugepage environment. Unrecognized values fall back to `spdk` with a warning. Note the master process honors it too (with `none`, heartbeat probes fail and NoF segments are unmounted after the threshold).
- **SPDK env init is now lazy.** `RealClient::setup` used to initialize the SPDK env eagerly and fail when hugepages were missing — even for clients that never touch NoF. Setup now always succeeds; a broken SPDK env surfaces as an error on the *first NoF operation* instead. Keep configuring hugepages before NoF traffic, as the deployment guide says.
- **`register_buffer` registers with SPDK automatically** — no extra call or parameter. The buffer must be block-size aligned (existing hard gate) and must live in hugepage-backed memory (2MB hugetlb). The pointer itself need not be 2MB-aligned — a non-aligned buffer is registered by its containing 2MB pages (logging a `RegisterMemory` warning); but 4KB-backed malloc memory fails inside SPDK's `spdk_mem_register` with paddr-misalignment `-EINVAL`.
- **The #3131 fix needs SPDK ≥ 22.05** (`spdk_mem_register`). Older SPDK still builds (with a CMake warning) and behaves exactly as before — i.e. NoF-over-RDMA writes from unregistered user buffers keep failing with "No translation".

### Use cases

**1. Pure-NoF write to the SSD pool** (the #3131 scenario — now works over RDMA):

```python
from mooncake.store import MooncakeDistributedStore, ReplicateConfig

store = MooncakeDistributedStore()
store.setup(...)  # protocol="rdma", local_buffer_size > 0 — see deployment guide

# Now also registered with SPDK's RDMA translation table (automatic):
store.register_buffer(buf_ptr, buf_size)

config = ReplicateConfig()
config.replica_num = 0      # no DRAM replica
config.nof_replica_num = 1  # NoF SSD replica only
store.batch_put_from(keys, buffer_ptrs, sizes, config)
```

Requirements: `USE_NOF=ON` build against SPDK ≥ 22.05; the buffer must be block-size aligned (existing hard gate) and must be hugepage-backed (2MB hugetlb). The pointer itself need not be 2MB-aligned — `register_buffer` normalizes it to its containing 2MB pages, logging a `RegisterMemory` warning — but 4KB-backed malloc memory fails inside SPDK's `spdk_mem_register` with paddr-misalignment `-EINVAL`.

**2. One binary, deployed everywhere** — nodes without hugepages/SPDK no longer need a separate build:

```bash
export MC_NOF_BACKEND=none  # NoF disabled at runtime; the rest works as usual
```

**3. Standard SSD-pool deployment** (master + store service + `spdk_tgt_create` + `mooncake_ssd_register`): unchanged — see `docs/source/deployment/ssd/nvmf-ssd-deployment-guide.md`.

**4. Custom/test initiator injection** (C++): any `NVMeoFInitiator` implementation can be injected — directly into `NofWorkerPool`, or via the new tail parameter of `Client::Create(...)`. The alignment gate, QoS, and worker loop run unchanged against it; `nvmeof_initiator_test` exercises the full pool this way without SPDK.

### What if the backend is not SPDK later?

That swap is the point of the split. A non-SPDK backend implements `NVMeoFInitiator` (and optionally `DmaBufferAllocator`) and plugs into `CreateNofRuntime()` — the single place that names SPDK types. Callers (`NofWorkerPool`, `RealClient`, master probe) do not change. Backend differences are advertised through `NofCapabilities`: e.g. a backend without an RDMA translation table returns `requires_memory_registration = false` and its `RegisterMemory` is a no-op `OK`. The new test suite already runs the worker pool against a non-SPDK `MockInitiator`, proving the seam.

Backend selection exists from day one as one env var choosing a matched Initiator+Allocator pair: `MC_NOF_BACKEND=spdk` (default) or `none` (runtime-disable NoF in a `USE_NOF` build). Adding a backend is one enum value + one factory branch + one env value — deliberately no plugin machinery before a second backend exists.

## Module

- [x] Docs

## Type of Change

- [x] Bug fix (#3131)
- [x] Refactor

## How Has This Been Tested?

**Test commands:**
```bash
cmake --build build -j            # USE_NOF=OFF: 0 errors, 0 warnings
./build/mooncake-store/tests/nvmeof_initiator_test   # new mock-initiator suite
./build/mooncake-store/tests/transfer_task_test
./build/mooncake-store/tests/client_buffer_test
```

**Test results:**
- [x] Unit tests pass (nvmeof_initiator_test 4/4, transfer_task_test 11/11, client_buffer_test 16/16)

⚠️ `USE_NOF=ON` build and runtime NoF validation still pending — requires a machine with SPDK + an NVMe-oF target. Will run before marking ready (incl. non-2MB-aligned hugepage-slice registration and NoF-over-RDMA pure-NoF write from #3131's repro).

## Checklist

- [x] I have run pre-commit on the files changed in this PR and all hooks pass *(local pre-commit 2.17 too old for the repo config; ran the equivalent hooks manually: clang-format via `code_format.sh --staged`, whitespace/EOF checks)*
- [ ] For changes >500 LOC: I have filed an RFC issue *(design doc exists; will file/link the RFC issue before undrafting)*


mooncake-store/src/transfer_task.cpp
```diff
@@ -1402,33 +1412,39 @@ std::optional<TransferFuture> TransferSubmitter::submitSpdkNofOperation(
...
+    // dma_alignment 的消费点(评审 #7):无 SGL 时 ptr 仍须块对齐
+    // (RFC §5.3 的现状前置条件);#3251 的 SGL 落地后 supports_sgl=true,
+    // ptr 对齐自动放宽到 DWORD,接口不变。
+    const auto caps = nof_initiator_->GetCapabilities();
+    const uint64_t ptr_alignment =
+        caps.supports_sgl ? caps.dma_alignment : block_size;
+    if (block_size == kInvalidBlockSize ||
         handle.buffer_address_ % block_size != 0 || size % block_size != 0 ||
-        reinterpret_cast<std::uintptr_t>(ptr) % block_size != 0) {
+        reinterpret_cast<std::uintptr_t>(ptr) % ptr_alignment != 0) {
```
Enigmo-x: 这里buffer ptr不需要按block_size对齐，统一dma_alignment对齐就行了应该

mooncake-store/src/nof/spdk_initiator.cpp
```diff
@@ -0,0 +1,899 @@
...
+
+    ErrorCode Register(void* ptr, size_t size) {
+        constexpr uintptr_t kHugepageSize = 2ULL << 20;
+        const uintptr_t begin = reinterpret_cast<uintptr_t>(ptr);
+        const uintptr_t first_page = begin & ~(kHugepageSize - 1);
+        const uintptr_t end_page =
+            (begin + size + kHugepageSize - 1) & ~(kHugepageSize - 1);
+
+        std::lock_guard<std::mutex> lock(mutex_);
+        if (registered_sizes_.count(ptr) != 0) {
```
Enigmo-x: 这里只判断ptr一直，后一次的size如果超过前一次，会出现后一次大范围不会补全注册？

fcczzz:
I have not finished reviewing the overall design yet. For now, I only helped run a few NoF experiments in Docker. The following are the test results and the issues I encountered while setting up the environment.

### Test environment

- SPDK v23.01.1
- 2 MiB hugepages
- SPDK `Malloc0`: 64 MiB with a 4 KiB block size

## 1. NoF worker-pool smoke test

I ran `nof_worker_pool_bench` with 4 KiB writes, iodepth 8, a 32 MiB range, and a 10-second runtime.

```text
completed_ops=635921
failed_ops=0
bw=248.41 MiB/s
iops=63592.10
fail_iops=0.00
clat_mean=124.14 us
clat_p99=135.00 us
benchmark_rc=0
```

The SPDK initiator successfully connected to the RDMA target, and the workload completed without I/O failures.

## 2. User-buffer registration test

I ran a focused test for the `#3131` path using a pure NoF configuration:

```text
replica_num=0
nof_replica_num=1
```

The test used two adjacent 4 KiB buffers inside the same 2 MiB hugepage. Both addresses were intentionally not 2 MiB aligned.

The test performed the following operations:

1. Register both buffers.
2. Write from the source buffer with `put_from`.
3. Unregister the source buffer.
4. Read into the destination buffer with `get_into`.
5. Compare the returned data byte-for-byte.
6. Unregister the destination buffer and close the client.

I did not observe the previous `No translation for ptr` failure. This experiment covered non-aligned registration ranges and the case where two registered buffers share the same 2 MiB page.

## 3. Store-level end-to-end test

I also ran the repository's `store_kv_bench.py` as a more end-to-end functional test.

The test used:

- the public Python Store API;
- metadata server and master;
- a pure NoF configuration: `memory=0, nof=1`;
- the zcopy `put_from/get_into` path;
- four concurrent lanes;
- batches of eight objects.

It wrote 1,024 4 KiB objects and then read back and verified all 1,024 objects.

```text
summary.ok=true
write.successful_kvs=1024
read.successful_kvs=1024
overall.successful_kvs=2048
failed_kvs=0
failed_requests=0
misses=0
verify_failures=0
journal_records=0
```

This exercised the following path:

```text
Python Store API
→ RealClient
→ metadata server
→ master NoF allocation
→ NoF worker pool
→ SPDK initiator
→ RDMA target
→ readback verification
```

## Issues encountered

### Missing yalantinglibs dependency

A clean `USE_NOF=ON` build failed with:

```text
fatal error: ylt/struct_json/json_reader.h: No such file or directory
```

`spdk_initiator.cpp` indirectly includes `types.h`, which requires yalantinglibs. However, the new `mooncake_nof_spdk` target does not currently declare a dependency on `yalantinglibs::yalantinglibs`.

I temporarily injected the yalantinglibs include path globally to continue testing. It looks like the target should declare the dependency directly:

```cmake
target_link_libraries(
  mooncake_nof_spdk
  PRIVATE
    ${SPDK_STATIC_LIBS}
    glog::glog
    yalantinglibs::yalantinglibs)
```

### IOVA mode inside the container

The initiator initially failed with:

```text
EAL: FATAL: Cannot use IOVA as 'PA' since physical addresses are not available
```

For these experiments, I used the following local test-only change:

```cpp
opts.iova_mode = "va";
```

The previous `spdk_wrapper.cpp` implementation appears to have the same default behavior, so this looks like an existing container-compatibility limitation rather than a regression introduced by this PR.

## Scope and limitations

These are experimental results only, not a complete design review or approval of the PR.

The tests used an SPDK memory bdev and a same-host RDMA connection. They did not cover:

- physical NVMe devices;
- cross-host RDMA traffic;
- multiple NoF endpoints or worker scaling;
- reconnect, target restart, or error recovery;
- long-running stability;
- all code paths and design decisions changed by this PR.

The end-to-end run was short and included cold-start costs, so its throughput and latency numbers should be treated as functional evidence rather than representative performance measurements.

mooncake-store/src/nof/spdk_initiator.cpp
```diff
@@ -0,0 +1,899 @@
...
+                std::lock_guard<std::mutex> lock(probe_ctx->error_mutex);
+                *error_reason = probe_ctx->error_reason.empty()
+                                    ? "completion_error"
+                                    : probe_ctx->error_reason;
+            }
+        }
+
+        // Ride-along fix (former probe-ctx recycle race): recycle only AFTER
+        // the caller has read all results, never inside the callback.
+        RecycleProbeRequestContext(probe_ctx);
```
fcczzz: `completion_timeout` only stops the caller from waiting; it does not cancel or drain the submitted NVMe command. The command still retains `&probe_ctx->adaptor`, but this unconditional recycle allows the LIFO pool to reset the same context for the next probe. When the old completion is processed by a later poll, it can write `done/success` into the new probe and cause the heartbeat to consume the previous request's result. Please keep timed-out contexts out of the pool until their original callback runs, or explicitly cancel/drain the command, and add a delayed-completion regression test.

mooncake-store/src/real_client.cpp
```diff
@@ -3224,6 +3217,23 @@ tl::expected<void, ErrorCode> RealClient::register_buffer_internal(
     if (!result) {
         return result;
     }
+    // #3131: SPDK 的 RDMA 传输维护独立的 translation table,TE 注册对它
+    // 不可见。TE 注册成功后,向 initiator 注册;失败则回滚 TE 注册,
+    // 不留半注册状态。
+    if (nof_runtime_.initiator) {
+        auto rc = nof_runtime_.initiator->RegisterMemory(buffer, size);
```
fcczzz, line 3417: This new SPDK registration has no teardown counterpart. `tearDownAll_internal()` never walks `registered_buffer_sizes_` and calls `UnregisterMemory()`, and destroying the initiator does not remove these entries because they live in the process-global `NofPageRegistry`. Closing a client without manually unregistering every buffer therefore leaves orphan page registrations; if the memory is later freed or remapped, another client may reuse stale SPDK translation state. Please drain outstanding NoF I/O and release this client's registrations during teardown, and add a close-without-unregister/reopen regression test.

mooncake-store/src/nof/spdk_initiator.cpp
fcczzz: This makes registration idempotent globally by pointer rather than per owner. `CreateNofRuntime()` creates a distinct `SpdkInitiator` for each client, so two clients can register the same buffer; the second call adds no reference, and the first client's `UnregisterMemory()` then removes the mapping while the second client still relies on it. A repeated registration of the same base with a larger size is also silently treated as fully registered. Please track per-initiator ownership separately from the process-global per-page refcounts.

fcczzz (commented):
The abstraction direction looks reasonable, and the new SPDK experiments provide useful evidence for the normal #3131 path. I still see lifecycle gaps around timeout handling and memory registration: a timed-out probe can recycle an in-flight callback context, and client teardown does not release its process-global SPDK registrations. The global registry also needs ownership accounting when multiple initiators register the same address. I think these should be addressed before merge.

mooncake-store/src/nof/spdk_initiator.cpp
```diff
@@ -0,0 +1,828 @@
...
+        opts.name = "mooncake";
+        // Containers without physical-address access (restricted
+        // /proc/self/pagemap) cannot run IOVA-as-PA; MC_NOF_IOVA_MODE lets
+        // operators opt into "va" there. When unset, SPDK keeps its
+        // PA-preferred default (matching baseline SpdkWrapper).
+        const char* iova_mode = std::getenv("MC_NOF_IOVA_MODE");
+        if (iova_mode && *iova_mode != '\0') {
+            opts.iova_mode = iova_mode;
+        }
+        ok_ = (spdk_env_init(&opts) == 0);
```
fcczzz, line 144: `SpdkEnvGuard` does not appear to be safely reacquirable after its last reference is released. The deleter calls `spdk_env_fini()`, while a subsequent constructor calls `spdk_env_init(&opts)` again. In SPDK v23.01.1, reinitialization within the same process requires a NULL argument; passing non-NULL options returns -EINVAL.
One sequence that exposes this is: create a client and initialize SPDK through the DMA allocator or a NoF operation; destroy the client without constructing the page-registry environment pin; then create another client in the same process. The second runtime’s first Acquire() will fail, leaving NoF—and potentially setup-time staging-buffer allocation—unavailable.
Could we either keep the SPDK environment alive until process exit, or explicitly track the reinitialization state and call `spdk_env_init(nullptr)` after `spdk_env_fini()`? The concurrent ordering between the last release and a new `Acquire()` would also need to be serialized.
It would also be helpful to add a close/reopen regression test that does not call `RegisterMemory`, since registration currently pins the environment and could mask this failure.

LujhCoconut: fixed

PR #3436, opened 2026-08-14
[TransferEngine] Bind buffer device for EFA CUDA loopback

### LINKED ISSUE #3434: [Bug]: EFA tryLoopbackCopy creates unintended CUDA contexts on GPU 0
### Bug Report

Mooncake `v0.3.12.post1` appears to create unintended CUDA contexts on GPU0 when using the EFA same-process loopback path.

#### Environment

- 8× NVIDIA B200
- One SGLang process per GPU
- CUDA 13
- Mooncake built with `USE_EFA=ON` and `USE_CUDA=ON`
- AWS EFA installer 1.48.0

#### Observed behavior

Initially, each process appears only on its assigned GPU. During Mooncake initialization, processes assigned to GPUs 1–7 also acquire approximately 616 MiB each on GPU0:

```text
GPU0  DP0  normal model allocation
GPU0  DP1  616 MiB
GPU0  DP2  616 MiB
...
GPU0  DP7  616 MiB
```

This consumes up to approximately 4.2 GiB on GPU0. No equivalent extra contexts appear on the other GPUs.

All primary OOMs then occur on DP0/GPU0. For example:

```text
CUDA out of memory. Tried to allocate 12.01 GiB.
GPU 0 free: 9.81 GiB.
DP1–DP7: 616 MiB each on GPU0.
```

Without those additional contexts, approximately 14 GiB would have been free, enough for the requested allocation. Mooncake subsequently reports:

```text
EFA loopback cudaMemcpy failed: out of memory
```

The active `engine.so` was verified to come from `mooncake-transfer-engine 0.3.12.post1`.

#### Suspected cause

`EfaContext::tryLoopbackCopy()` calls:

```cpp
cudaMemcpy(dst, src, len, cudaMemcpyDefault);
```

This runs on a Mooncake background thread without explicitly selecting the CUDA device associated with the buffers. Because CUDA device selection is thread-local, the call appears to initialize a context on the default device, GPU0.

This loopback path was added after `v0.3.11.post1`, where we did not observe the problem.

#### Suggested fix

Before the loopback copy, determine the buffer’s CUDA device and bind that device on the calling thread. Host-to-host copies could use ordinary `memcpy`.

The TCP transport previously fixed the same GPU0-context behavior:

- https://github.com/kvcache-ai/Mooncake/pull/2307

Related changes:

- EFA loopback implementation: https://github.com/kvcache-ai/Mooncake/pull/2298
- EFA registration-thread context fix: https://github.com/kvcache-ai/Mooncake/pull/3177

I have not yet tested a patched build, but the wrong-device contexts, GPU0-only OOMs, and memory amounts are consistently reproducible.

### Before submitting...

yejun-yun:
## Description

EFA same-process loopback can run on a submitting thread with no current CUDA context. Calling `cudaMemcpy` directly in that case can initialize a context on GPU 0 even when the buffers belong to another GPU. Fixes #3434. 

This PR:

- determines the CUDA device from the destination or source buffer;
- binds that device’s primary context before copying;
- uses `memcpy` for host-only loopback copies;
- adds a multi-GPU regression test that submits from a fresh thread and verifies GPU 1 is selected instead of GPU 0.

## Module

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

```bash
./scripts/code_format.sh --staged
pre-commit run --files \
  mooncake-transfer-engine/src/transport/efa_transport/efa_context.cpp \
  mooncake-transfer-engine/tests/efa_gpu_loopback_test.cpp
```

Formatting and applicable pre-commit hooks pass. The EFA GPU regression test was not run locally because it requires Linux, EFA hardware, and at least two CUDA GPUs.

## Checklist

## AI Assistance Disclosure

- [x] AI tools were used

Cursor assisted with drafting the implementation and regression test. I reviewed the resulting changes.

### DIFF
--- a/mooncake-transfer-engine/src/transport/efa_transport/efa_context.cpp
+++ b/mooncake-transfer-engine/src/transport/efa_transport/efa_context.cpp
@@ -413,9 +413,11 @@ int EfaContext::deconstruct() {
 static void logCudaFailure(const char* call, CUresult ret, int device_ordinal) {
     const char* err = nullptr;
     cuGetErrorString(ret, &err);
-    LOG(WARNING) << "EFA: " << call << " failed for CUDA device "
-                 << device_ordinal << ": " << (err ? err : "unknown") << " ("
-                 << ret
+    LOG(WARNING) << "EFA: " << call << " failed"
+                 << (device_ordinal >= 0
+                         ? " for CUDA device " + std::to_string(device_ordinal)
+                         : "")
+                 << ": " << (err ? err : "unknown") << " (" << ret
                  << "); GPU memory registration or a GPU-aware copy may fail"
                     " with a bare \"Operation not supported\"";
 }
@@ -438,10 +440,13 @@ static void logCudaFailure(const char* call, CUresult ret, int device_ordinal) {
 //
 // cuDevicePrimaryCtxRetain() returns the same primary context the CUDA runtime
 // uses, so this attaches to the process's existing context rather than creating
-// another.  The retain is intentionally not released: the primary context
-// outlives every registration, and dropping the last reference here would tear
-// down the context the rest of the process is using.
-static bool bindCudaContextIfNeeded(int device_ordinal) {
+// another.  Registration keeps its retain for the process lifetime.  Scoped
+// callers request `retained_primary` and release that reference after restoring
+// the previous context.
+static bool bindCudaContextIfNeeded(int device_ordinal,
+                                    bool* retained_primary = nullptr) {
+    if (retained_primary) *retained_primary = false;
+
     CUdevice want;
     CUresult ret = cuDeviceGet(&want, device_ordinal);
     if (ret != CUDA_SUCCESS) {
@@ -456,13 +461,15 @@ static bool bindCudaContextIfNeeded(int device_ordinal) {
         logCudaFailure("cuCtxGetCurrent", ret, device_ordinal);
         return false;
     }
+    bool same_device = false;
     if (cur != nullptr) {
         ret = cuCtxGetDevice(&cur_dev);
         if (ret != CUDA_SUCCESS) {
             logCudaFailure("cuCtxGetDevice", ret, device_ordinal);
             return false;
         }
-        if (cur_dev == want) return true;
+        same_device = cur_dev == want;
+        if (same_device && !retained_primary) return true;
     }
 
     CUcontext primary = nullptr;
@@ -471,6 +478,9 @@ static bool bindCudaContextIfNeeded(int device_ordinal) {
         logCudaFailure("cuDevicePrimaryCtxRetain", ret, device_ordinal);
         return false;
     }
+    if (retained_primary) *retained_primary = true;
+    if (same_device && cur == primary) return true;
+
     ret = cuCtxSetCurrent(primary);
     if (ret != CUDA_SUCCESS) {
         logCudaFailure("cuCtxSetCurrent", ret, device_ordinal);
@@ -832,15 +842,27 @@ bool EfaContext::tryLoopbackCopy(Transport::Slice* slice) {
     // backends do not expose the cuda* symbols).  CUDA host-only copies also
     // use memcpy; copies involving device memory use cudaMemcpyDefault after
     // binding the destination device, or the source device for D2H copies.
-    // The caller's context is restored afterward, and D2D copies synchronize
+    // The caller's context is restored afterward, and CUDA copies synchronize
     // before the slice is reported complete.
 #if defined(USE_CUDA)
+    CUcontext previous = nullptr;
+    CUresult context_rc = cuCtxGetCurrent(&previous);
+    if (context_rc != CUDA_SUCCESS) {
+        logCudaFailure("cuCtxGetCurrent", context_rc, -1);
+        slice->markFailed();
+        return true;
+    }
+
     cudaError_t dst_status;
     cudaError_t src_status;
     int dst_device = getCudaDeviceForPointer(dst, dst_status);
     int src_device = getCudaDeviceForPointer(src, src_status);
     if (dst_status != cudaSuccess || src_status != cudaSuccess) {
         cudaError_t rc = dst_status != cudaSuccess ? dst_status : src_status;
+        context_rc = cuCtxSetCurrent(previous);
+        if (context_rc != CUDA_SUCCESS) {
+            logCudaFailure("cuCtxSetCurrent (restore)", context_rc, -1);
+        }
         LOG(ERROR) << "EFA loopback cudaPointerGetAttributes failed: "
                    << cudaGetErrorString(rc) << " (dst=" << dst
                    << ", src=" << src << ", len=" << len << ")";
@@ -851,27 +873,32 @@ bool EfaContext::tryLoopbackCopy(Transport::Slice* slice) {
     int device = dst_device >= 0 ? dst_device : src_device;
     if (device < 0) {
         memcpy(dst, src, len);
-    } else {
-        CUcontext previous = nullptr;
-        CUresult context_rc = cuCtxGetCurrent(&previous);
+        context_rc = cuCtxSetCurrent(previous);
         if (context_rc != CUDA_SUCCESS) {
-            logCudaFailure("cuCtxGetCurrent", context_rc, device);
+            logCudaFailure("cuCtxSetCurrent (restore)", context_rc, -1);
             slice->markFailed();
             return true;
         }
-
-        bool bound = bindCudaContextIfNeeded(device);
+    } else {
+        bool retained_primary = false;
+        bool bound = bindCudaContextIfNeeded(device, &retained_primary);
         cudaError_t rc = cudaSuccess;
         if (bound) {
             rc = cudaMemcpy(dst, src, len, cudaMemcpyDefault);
-            if (rc == cudaSuccess && dst_device >= 0 && src_device >= 0) {
-                rc = cudaStreamSynchronize(0);
-            }
+            if (rc == cudaSuccess) rc = cudaStreamSynchronize(0);
         }
 
         context_rc = cuCtxSetCurrent(previous);
         if (context_rc != CUDA_SUCCESS) {
             logCudaFailure("cuCtxSetCurrent (restore)", context_rc, device);
+        }
+
+        CUresult release_rc = CUDA_SUCCESS;
+        if (retained_primary) {
+            release_rc = cuDevicePrimaryCtxRelease(device);
+            if (release_rc != CUDA_SUCCESS) {
+                logCudaFailure("cuDevicePrimaryCtxRelease", release_rc, device);
+            }
         }
 
         if (!bound) {
@@ -886,6 +913,10 @@ bool EfaContext::tryLoopbackCopy(Transport::Slice* slice) {
             return true;
         }
         if (context_rc != CUDA_SUCCESS) {
+            slice->markFailed();
+            return true;
+        }
+        if (release_rc != CUDA_SUCCESS) {
             slice->markFailed();
             return true;
         }
--- a/mooncake-transfer-engine/src/transport/efa_transport/efa_context.cpp
+++ b/mooncake-transfer-engine/src/transport/efa_transport/efa_context.cpp
@@ -422,67 +422,8 @@ static void logCudaFailure(const char* call, CUresult ret, int device_ordinal) {
                     " with a bare \"Operation not supported\"";
 }
 
-// Make `device_ordinal`'s primary context current on the calling thread, unless
-// a context for that same device already is.
-//
-// Why a context is needed at all: fi_mr_regattr() on FI_HMEM_CUDA memory
-// reaches libfabric's cuda_get_dmabuf_fd(), which calls the driver API
-// cuMemGetHandleForAddressRange() and does no context management of its own.
-// The export also runs against the CURRENT context, so a context belonging to
-// another device is no better than none -- both end up as a bare "Operation not
-// supported" from the provider.
-//
-// Who arrives here without the right context: registerLocalMemoryBatch() runs
-// one std::async(std::launch::async) per buffer and registerLocalMemory() can
-// fan out one std::thread per NIC, so a registering thread may have touched no
-// CUDA API at all; and since std::async may reuse threads, one thread can
-// register buffers on several devices in turn.
-//
-// cuDevicePrimaryCtxRetain() returns the same primary context the CUDA runtime
-// uses, so this attaches to the process's existing context rather than creating
-// another.  The retain is intentionally not released: the primary context
-// outlives every registration, and dropping the last reference here would tear
-// down the context the rest of the process is using.
-static bool bindCudaContextIfNeeded(int device_ordinal) {
-    CUdevice want;
-    CUresult ret = cuDeviceGet(&want, device_ordinal);
-    if (ret != CUDA_SUCCESS) {
-        logCudaFailure("cuDeviceGet", ret, device_ordinal);
-        return false;
-    }
-
-    CUcontext cur = nullptr;
-    CUdevice cur_dev;
-    ret = cuCtxGetCurrent(&cur);
-    if (ret != CUDA_SUCCESS) {
-        logCudaFailure("cuCtxGetCurrent", ret, device_ordinal);
-        return false;
-    }
-    if (cur != nullptr) {
-        ret = cuCtxGetDevice(&cur_dev);
-        if (ret != CUDA_SUCCESS) {
-            logCudaFailure("cuCtxGetDevice", ret, device_ordinal);
-            return false;
-        }
-        if (cur_dev == want) return true;
-    }
-
-    CUcontext primary = nullptr;
-    ret = cuDevicePrimaryCtxRetain(&primary, want);
-    if (ret != CUDA_SUCCESS) {
-        logCudaFailure("cuDevicePrimaryCtxRetain", ret, device_ordinal);
-        return false;
-    }
-    ret = cuCtxSetCurrent(primary);
-    if (ret != CUDA_SUCCESS) {
-        logCudaFailure("cuCtxSetCurrent", ret, device_ordinal);
-        return false;
-    }
-    return true;
-}
-
-// Loopback runs synchronously on the caller's thread. Temporarily use the
-// buffer's primary context, then restore the caller and balance the retain.
+// RAII guard: temporarily make a device's primary context current, then restore
+// the caller's original context and release the retain on destruction.
 class ScopedCudaContext {
    public:
     ScopedCudaContext() {
@@ -618,7 +559,9 @@ int EfaContext::registerMemoryRegionInternal(void* addr, size_t length,
     int ret;
     if (iface != FI_HMEM_SYSTEM) {
 #if defined(USE_CUDA)
-        if (!bindCudaContextIfNeeded(device_ordinal)) return ERR_CONTEXT;
+        ScopedCudaContext reg_context;
+        if (!reg_context.valid() || !reg_context.bind(device_ordinal))
+            return ERR_CONTEXT;
 #endif
         // GPU memory: use fi_mr_regattr with explicit iface and device
         struct iovec iov = {.iov_base = addr, .iov_len = length};
@@ -917,12 +860,6 @@ bool EfaContext::tryLoopbackCopy(Transport::Slice* slice) {
     // The caller's context is restored afterward, and CUDA copies synchronize
     // before the slice is reported complete.
 #if defined(USE_CUDA)
-    ScopedCudaContext context;
-    if (!context.valid()) {
-        slice->markFailed();
-        return true;
-    }
-
     cudaError_t dst_status;
     cudaError_t src_status;
     int dst_device = getCudaDeviceForPointer(dst, dst_status);
@@ -939,27 +876,20 @@ bool EfaContext::tryLoopbackCopy(Transport::Slice* slice) {
     int device = dst_device >= 0 ? dst_device : src_device;
     if (device < 0) {
         memcpy(dst, src, len);
-        if (!context.restore()) {
-            slice->markFailed();
-            return true;
-        }
     } else {
-        if (!context.bind(device)) {
+        ScopedCudaContext context;
+        if (!context.valid() || !context.bind(device)) {
             slice->markFailed();
             return true;
         }
 
         cudaError_t rc = cudaMemcpy(dst, src, len, cudaMemcpyDefault);
         if (rc == cudaSuccess) rc = cudaStreamSynchronize(0);
-        bool restored = context.restore();
+        context.restore();
         if (rc != cudaSuccess) {
             LOG(ERROR) << "EFA loopback CUDA copy failed: "
                        << cudaGetErrorString(rc) << " (dst=" << dst
                        << ", src=" << src << ", len=" << len << ")";
-            slice->markFailed();
-            return true;
-        }
-        if (!restored) {
             slice->markFailed();
             return true;
         }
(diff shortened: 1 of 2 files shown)


mooncake-transfer-engine/src/transport/efa_transport/efa_context.cpp
```diff
@@ -806,16 +812,24 @@ bool EfaContext::tryLoopbackCopy(Transport::Slice* slice) {
...
+    } else {
+        bindCudaContextIfNeeded(device);
+        auto rc = cudaMemcpy(dst, src, len, cudaMemcpyDefault);
+        if (rc != cudaSuccess) {
+            LOG(ERROR) << "EFA loopback cudaMemcpy failed: "
+                       << cudaGetErrorString(rc) << " (dst=" << dst
+                       << ", src=" << src << ", len=" << len << ")";
+            slice->markFailed();
+            return true;
+        }
```
whn09: **Blocking: this rebinds the caller's CUDA device and never restores it.**

`tryLoopbackCopy()` does not run on an EFA worker thread. The only callers of `submitPostSend()` are `EfaTransport::submitTransfer()` / `submitTransferTask()`, i.e. **the application's own submitting thread**; `workerThreadFunc()` only polls CQs. (The comment added at line 432 says the opposite — see my note there.)

Since `bindCudaContextIfNeeded()` only ever calls `cuCtxSetCurrent()` and never restores, the caller's thread comes back bound to the buffer's device. I reproduced this through the real `TransferEngine` API on 8x B300 — caller thread on GPU 0, engine + registered buffer on GPU 1, one loopback WRITE:

```
REVIEW: caller cudaGetDevice() before submit = 0
REVIEW: caller cudaGetDevice() after  submit = 1; a fresh cudaMalloc landed on device 1
```

For a vLLM/SGLang-style caller that pins each worker thread to its own device, every subsequent `cudaMalloc` and kernel launch on that thread now lands on the wrong GPU. That is a worse failure than the one being fixed (which only wasted a primary context on GPU 0).

Suggested fix — scope the bind to the copy, and synchronize the D2D case:

```cpp
// submitTransfer() runs on the CALLER's thread, so binding the buffer's
// device here must be undone: leaving that context current changes the
// device the CUDA runtime reports for every later cudaMalloc / kernel
// launch on the caller's thread.
CUcontext prev = nullptr;
bool have_prev = cuCtxGetCurrent(&prev) == CUDA_SUCCESS;
bindCudaContextIfNeeded(device);
auto rc = cudaMemcpy(dst, src, len, cudaMemcpyDefault);
// cudaMemcpy does NOT synchronize the host for device-to-device transfers,
// so the copy may still be in flight when markSuccess() publishes the slice.
if (rc == cudaSuccess) rc = cudaStreamSynchronize(0);
if (have_prev) cuCtxSetCurrent(prev);
if (rc != cudaSuccess) { /* ...existing error handling... */ }
```

On the D2D half of that: CUDA documents "for transfers from device memory to device memory, no host-side synchronization is performed", and it shows up clearly at scale on B300 — a 1 GiB D2D `cudaMemcpy` returns in 26 us while the copy itself needs another 328 us:

```
D2D 1GiB rep0: cudaMemcpy returned in 0.026 ms, +sync 0.328 ms  => ASYNC
D2D 1GiB rep1: cudaMemcpy returned in 0.010 ms, +sync 0.328 ms  => ASYNC
```

GPU-to-GPU KV-cache loopback is exactly this path, and the new test cannot catch it because it never verifies the copied bytes.

I built and ran this patch on the B300 box: the caller's device is preserved (`0 -> 0`, fresh `cudaMalloc` lands on 0) and all three pre-existing loopback tests stay green (`GpuLoopbackWrite`, `GpuLoopbackWriteThenRead`, `GpuLoopbackMultiWrite`). Note it makes the new test fail at line 234 — see my comment there, that assertion encodes the leak.

FWIW the same missing restore pre-exists in the MR-registration fan-out (`efa_transport.cpp`, `worker();  // the caller is a worker too`), so the caller thread is affected there too. Not introduced by this PR, but the same guard would cover it if you want to fix both.

mooncake-transfer-engine/src/transport/efa_transport/efa_context.cpp
```diff
@@ -434,8 +429,9 @@ static void logCudaFailure(const char* call, CUresult ret, int device_ordinal) {
 // Who arrives here without the right context: registerLocalMemoryBatch() runs
 // one std::async(std::launch::async) per buffer and registerLocalMemory() can
 // fan out one std::thread per NIC, so a registering thread may have touched no
-// CUDA API at all; and since std::async may reuse threads, one thread can
-// register buffers on several devices in turn.
+// CUDA API at all.  Loopback copies also run on EFA worker threads.  Since
```
whn09: This sentence is not accurate, and it's load-bearing for the bug in this hunk: loopback copies run on the **caller's** thread, not on an EFA worker thread. `submitPostSend()` is reached only from `EfaTransport::submitTransfer()` / `submitTransferTask()`; `workerThreadFunc()` does CQ polling only and never touches `tryLoopbackCopy()`.

That distinction is why the un-restored `cuCtxSetCurrent()` matters — on a private worker thread leaking the context would be harmless, on the application's thread it is not. Suggest something like:

```
// CUDA API at all.  Loopback copies run on the CALLER's thread
// (submitTransfer -> submitPostSend -> tryLoopbackCopy), so any context
// bound there must be restored before returning to the caller.
```

Also, the second half ("Since these threads may be reused for buffers on different devices, bind by device rather than merely checking whether any context is current") describes behaviour `bindCudaContextIfNeeded()` already had before this PR — its body is unchanged here. Fine to keep, but it reads as if this PR changed it.

mooncake-transfer-engine/src/transport/efa_transport/efa_context.cpp
```diff
@@ -408,17 +408,12 @@ int EfaContext::deconstruct() {
...
-// helper exists to remove -- so name the driver call that actually failed.
 static void logCudaFailure(const char* call, CUresult ret, int device_ordinal) {
     const char* err = nullptr;
     cuGetErrorString(ret, &err);
     LOG(WARNING) << "EFA: " << call << " failed for CUDA device "
                  << device_ordinal << ": " << (err ? err : "unknown") << " ("
-                 << ret
-                 << "); GPU memory registration may fail with a bare"
-                    " \"Operation not supported\"";
+                 << ret << "); CUDA operation may fail";
```
whn09: Please keep the original wording here. `"CUDA operation may fail"` carries no information, whereas the sentence it replaced named the exact symptom this helper exists to explain: `fi_mr_regattr()` surfacing a bare `"Operation not supported"` from the provider with no attribution.

That opaque errno is genuinely the hardest EFA GPU failure to diagnose — this log line is the only breadcrumb pointing from it to the real driver failure, and the deleted comment is what tells the next reader why the helper exists at all. The helper is now shared with the loopback path, so if the wording needs to be more general, something like `"; GPU memory registration or a GPU-aware copy may fail with a bare \"Operation not supported\""` keeps the attribution.

mooncake-transfer-engine/src/transport/efa_transport/efa_context.cpp
```diff
@@ -466,6 +462,17 @@ static void bindCudaContextIfNeeded(int device_ordinal) {
...
+static int getCudaDeviceForPointer(const void* ptr) {
+    cudaPointerAttributes attributes;
+    cudaError_t ret = cudaPointerGetAttributes(&attributes, ptr);
+    if (ret != cudaSuccess) {
+        cudaGetLastError();
+        return -1;
+    }
+    if (attributes.type == cudaMemoryTypeDevice) return attributes.device;
+    return -1;
+}
```
whn09: Verified this helper is safe to call before binding — worth recording since it's the obvious worry with the new ordering. On a thread with no current context, `cudaPointerGetAttributes()` does **not** trigger lazy context creation; measured on B300:

```
[fresh-thread/before]       dev0_primary_active=0  cur_ctx=(nil)  cur_ctx_dev=-1
getCudaDeviceForPointer(gpu1_ptr) = 1
[fresh-thread/after-probe]  dev0_primary_active=0  cur_ctx=(nil)  cur_ctx_dev=-1
```

Two small things:

1. **Managed memory changes behaviour.** `cudaMemoryTypeManaged` falls through to `return -1`, so a managed buffer now takes the plain `memcpy` branch where it previously went through `cudaMemcpyDefault`. It happens to work under UVM via page migration, but it is a silent semantic change. Either handle `cudaMemoryTypeManaged` explicitly or note in a comment that the host `memcpy` is intentional for it.

2. **`cudaGetLastError()` runs on the caller's thread** (see the loopback comment), so it can swallow a pending error the application has not observed yet. Low risk in practice — on CUDA 13 a plain `malloc()` pointer returns `cudaSuccess` with `type == cudaMemoryTypeUnregistered`, so this branch is rarely taken:

```
host malloc ptr: rc=no error type=0 (0=unregistered,...) -> getCudaDeviceForPointer=-1
```

Still, clearing another component's error from a shared thread is worth a comment at minimum.

mooncake-transfer-engine/src/transport/efa_transport/efa_context.cpp
```diff
@@ -806,16 +812,24 @@ bool EfaContext::tryLoopbackCopy(Transport::Slice* slice) {
     // including non-EFA GPU backends such as MUSA/MLU/MACA, which never run
     // on AWS EFA hardware — registers loopback buffers as host memory, so a
     // plain memcpy is both correct and the only portable option (those
-    // backends do not expose the cuda* symbols).  *MemcpyDefault picks
-    // H2H/H2D/D2H/D2D from the pointer attributes.
+    // backends do not expose the cuda* symbols).  CUDA host-only copies also
+    // use memcpy; copies involving device memory use cudaMemcpyDefault after
+    // binding the destination device, or the source device for D2H copies.
```
whn09, line 859: Worth mentioning in the PR description: this is not just a correctness change, the host-only path gets substantially faster. Measured on B300 for a 4 KB host-to-host loopback slice:

```
host-only 4KB loopback slice, OLD (cudaMemcpyDefault) : 6277 ns
host-only 4KB loopback slice, NEW (2x probe + memcpy) : 1001 ns
```

~6x, and that already includes both `cudaPointerGetAttributes()` probes. Host-memory loopback is the common Mooncake CPU KV-cache case, so this is the most broadly felt effect of the PR and currently goes unclaimed.

mooncake-transfer-engine/src/transport/efa_transport/efa_context.cpp
```diff
@@ -806,16 +812,24 @@ bool EfaContext::tryLoopbackCopy(Transport::Slice* slice) {
...
+        if (rc != cudaSuccess) {
+            LOG(ERROR) << "EFA loopback cudaMemcpy failed: "
+                       << cudaGetErrorString(rc) << " (dst=" << dst
+                       << ", src=" << src << ", len=" << len << ")";
+            slice->markFailed();
+            return true;
+        }
     }
 #elif defined(USE_HIP)
     auto rc = hipMemcpy(dst, src, len, hipMemcpyDefault);
```
whn09, line 900: Nit: the HIP branch has the same shape as the pre-fix CUDA branch — `hipMemcpy` with no device binding — so if the reasoning in this PR holds, it holds for HIP too. EFA + HIP is not a real deployment combination, so I would not add the machinery; a one-line comment saying why HIP deliberately skips the bind would keep the asymmetry from looking accidental.

mooncake-transfer-engine/tests/efa_gpu_loopback_test.cpp
```diff
@@ -173,6 +178,74 @@ class EFAGpuLoopbackTest : public ::testing::Test {
...
+            device_result = cuCtxGetDevice(&worker_device);
+        }
+    });
+    submitter.join();
+
+    ASSERT_TRUE(started_without_context);
+    ASSERT_TRUE(transfer_ok);
+    ASSERT_EQ(context_result, CUDA_SUCCESS);
+    ASSERT_EQ(device_result, CUDA_SUCCESS);
+    EXPECT_EQ(worker_device, 1);
```
whn09: These two assertions encode the context leak as the expected behaviour: they require that a context is *still current* on the thread after the copy returns, and that it is device 1's. Once `tryLoopbackCopy()` restores the caller's previous context (see my comment on `efa_context.cpp`), `cuCtxGetCurrent()` returns `nullptr` on this fresh thread, `device_result` keeps its `CUDA_ERROR_UNKNOWN` initializer, and the test fails right here:

```
/home/ubuntu/Mooncake/mooncake-transfer-engine/tests/efa_gpu_loopback_test.cpp:234: Failure
[  FAILED  ] EFAGpuLoopbackTest.LoopbackUsesBufferDevice
```

The assertion that actually pins the bug is the `device_zero_active_after == 0` check below — that one is exactly right, please keep it. I'd suggest dropping lines 233-235 and replacing them with the complementary property: that the submitting thread's device is unchanged. Something like this, which I ran on B300 and which fails on this PR as-is and passes with the restore:

```cpp
ASSERT_EQ(cudaSetDevice(0), cudaSuccess);
int before = -1, after = -1;
ASSERT_EQ(cudaGetDevice(&before), cudaSuccess);
// ... submit the loopback transfer to the GPU-1 buffer ...
ASSERT_EQ(cudaGetDevice(&after), cudaSuccess);
EXPECT_EQ(after, before) << "loopback copy rebound the caller's device";
void *probe = nullptr;
ASSERT_EQ(cudaMalloc(&probe, 1 << 20), cudaSuccess);
cudaPointerAttributes attr;
ASSERT_EQ(cudaPointerGetAttributes(&attr, probe), cudaSuccess);
EXPECT_EQ(attr.device, before) << "subsequent allocations land on the wrong GPU";
```

Two smaller test issues while you're here: `context_result` is reused for the before- and after-`cuCtxGetCurrent()` calls, so a failure in the first is silently overwritten by the second; and the `ASSERT_*` failures here return before `destroyEngine(setup)`, leaking `dev_buf`.

mooncake-transfer-engine/tests/efa_gpu_loopback_test.cpp
```diff
@@ -173,6 +178,74 @@ class EFAGpuLoopbackTest : public ::testing::Test {
...
+    auto setup = createEngine(1);
+    if (!setup.ok) GTEST_SKIP() << "EFA/CUDA setup unavailable";
+
+    auto segment_desc =
+        setup.engine->getMetadata()->getSegmentDescByID(setup.segment_id);
+    ASSERT_NE(segment_desc, nullptr);
+    uint64_t remote_base = (uint64_t)segment_desc->buffers[0].addr;
+
+    const size_t kDataLength = 4096;
+    ASSERT_EQ(cudaMemset(setup.dev_buf, 0xAB, kDataLength), cudaSuccess);
```
whn09: The buffer is filled with `0xAB` but the destination is never read back, so this test verifies only the context side effects, not that the copy landed. For a change whose whole point is "bind the *right* device before copying", the strongest available assertion is that the bytes arrived — and it is also the only thing that would catch the D2D asynchrony I flagged in `efa_context.cpp` (a 1 GiB D2D `cudaMemcpy` returns 328 us before the data is there).

Suggest a `cudaMemcpy` D2H of the destination range plus a byte compare after the transfer completes, as `GpuLoopbackWriteThenRead` already does.

whn09 (changes requested):
Reviewed as EFA transport code owner. **Verified on real hardware**: 2x p6-b300 (8x B300 SXM6 each), EFA installer 1.50.0, CUDA 13, built with `-DUSE_EFA=ON -DUSE_CUDA=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo` at PR head `2de5964`. The new test **passes** there:

```
[       OK ] EFAGpuLoopbackTest.LoopbackUsesBufferDevice (6076 ms)
```

The bug you're fixing is real — I measured it. On a thread with no current context, a bare `cudaMemcpy` does activate GPU 0's primary context and make it current:

```
[fresh-thread/before]           dev0_primary_active=0  cur_ctx=(nil)      cur_ctx_dev=-1
[fresh-thread/after-bare-memcpy] dev0_primary_active=1  cur_ctx=0x...570  cur_ctx_dev=0
```

And your probe-then-bind ordering is safe: `cudaPointerGetAttributes` does **not** create a context, so `getCudaDeviceForPointer()` cannot itself light up GPU 0.

Two things need to change before this can merge, both detailed inline:

1. **(blocking)** The bind is never undone, and `tryLoopbackCopy()` runs on the **caller's** thread — so a loopback transfer silently rebinds the *application's* thread to another GPU. Measured through the real `TransferEngine` API.
2. **(strongly recommended, pre-existing)** `cudaMemcpy` is not host-synchronous for D2D, so `markSuccess()` can publish a GPU→GPU slice before the data lands.

One thing worth adding to the PR description, since it's a real win you're not claiming: replacing `cudaMemcpy` with `memcpy` on the host-only path makes host loopback **~6x faster** (4 KB slice: 6.28 us -> 1.00 us, including both pointer probes). The GPU path costs only +112 ns/slice for the probe + bind fast path, so there is no perf objection here.

Measured per-call costs on B300, for reference:

| call | ns/call |
|---|---|
| `cudaPointerGetAttributes` (device ptr) | 85 |
| `cudaPointerGetAttributes` (host ptr) | 482 |
| `bindCudaContextIfNeeded` fast path | 27 |

Happy to hand over the patch and the two extra regression tests I wrote while verifying this.

yejun-yun:
Thanks for the review and testing. I pushed an update that restores the caller’s CUDA context after the loopback copy, waits for GPU copies to finish before reporting success, and improves the regression tests. I left the pre-existing registration context-lifetime behavior unchanged.

whn09:
Re-reviewed at head `3bdc0a91` and validated on **2× p5.48xlarge** (8×H100 80GB, 32 EFA NICs, libfabric 2.4.0amzn5.0 / EFA installer 1.49.0, CUDA 13.0.88, driver 595.91.07). Both this head and the merge base `11fe0780` were built with `-DUSE_EFA=ON -DUSE_CUDA=ON`, and I checked each build really takes the GPU registration path (`fi_mr_regattr` with `FI_HMEM_CUDA`, not `fi_mr_reg`).

Thanks — all three of my earlier blocking comments are addressed, and the measurements say the fix does a bit more than the description claims. One small change left, plus one thing I'd like fixed here or in an immediate follow-up.

## Verified on hardware

| check | this PR | merge base `11fe0780` |
|---|---|---|
| `efa_gpu_loopback_test` | **5/5 pass** | — |
| this PR's test source × base library | — | `LoopbackPreservesEmptyCallerContext` **fails** (both assertions) |
| `efa_transport_test` (11 host-buffer tests) | 11/11 | 11/11 |
| cross-device loopback (dst on GPU 0, src on GPU 1) | pass | pass |
| cross-node `transfer_engine_bench` write, 128 KB blocks, 48 threads, batch 128, `--gpu_id=-1` | 141.0 / 145.8 / 149.8 / 156.0 GB/s | 145.4 / 144.7 / 151.0 / 140.3 GB/s |
| MR registration span for 8 GPUs | 10.17 / 10.18 / 10.50 s | 10.18 / 10.20 / 10.22 s |

No `fi_mr_regattr` failure and no `ERR_CONTEXT` in any run, so changing `bindCudaContextIfNeeded()` from best-effort to a hard failure in `registerMemoryRegionInternal` does not regress normal GPU MR registration — bandwidth differences are inside EFA's usual run-to-run spread.

Two things worth folding into the PR description:

* **The host path change is a correctness fix, not only a ~6× speed-up.** A pure host↔host loopback submitted from a thread that never touched CUDA leaks a context on the base library (`cuCtxGetCurrent`: `nullptr` → `0x564c…`), because `cudaMemcpy(cudaMemcpyDefault)` lazily creates and binds device 0's primary context on that thread. With this PR the thread stays at `nullptr`. So host buffers were affected too, not just `FI_HMEM_CUDA` ones.
* `LoopbackPreservesEmptyCallerContext` is the test that actually catches the reported bug — on the base library it trips both `loopback copy changed the caller's CUDA context` and `loopback copy activated CUDA device 0`. `LoopbackPreservesCallerDevice`, by contrast, **passes** against the base library: the old code never called `cuCtxSetCurrent`, so it can only regress the new bind/restore logic. Fine to keep, just don't count it as a regression test for the original issue.

## One change requested: don't fail a slice whose bytes already landed

`efa_context.cpp:942-945` and `:962-965` call `slice->markFailed()` when only `context.restore()` fails. The copy has already completed and been synchronized at that point, `markFailed()` is terminal on EFA (no slice redispatch, unlike the RDMA transport), and failing the transfer cannot repair the thread's context anyway.

Confirmed by fault injection (forcing `restore()` to report failure while the copy itself succeeds): the destination byte comparison **passes** and the transfer is still reported **FAILED**. `restore()` already logs the driver error, so:

```cpp
    int device = dst_device >= 0 ? dst_device : src_device;
    if (device < 0) {
        memcpy(dst, src, len);
        if (!context.restore()) {
            LOG(ERROR) << "EFA loopback host copy completed but restoring the "
                          "caller's CUDA context failed";
        }
    } else {
        if (!context.bind(device)) {
            slice->markFailed();
            return true;
        }

        cudaError_t rc = cudaMemcpy(dst, src, len, cudaMemcpyDefault);
        if (rc == cudaSuccess) rc = cudaStreamSynchronize(0);
        if (!context.restore()) {
            LOG(ERROR) << "EFA loopback CUDA copy: restoring the caller's CUDA "
                          "context failed";
        }
        if (rc != cudaSuccess) {
            LOG(ERROR) << "EFA loopback CUDA copy failed: "
                       << cudaGetErrorString(rc) << " (dst=" << dst
                       << ", src=" << src << ", len=" << len << ")";
            slice->markFailed();
            return true;
        }
    }
```

## Please also fix (here or right after): the registration path still hijacks the caller's GPU

Same class of bug as the one this PR fixes, at the other call site of `bindCudaContextIfNeeded()` (`efa_context.cpp:621`), and it reproduces on **both** this head and the base. A thread does `cudaSetDevice(1)`, confirms its own `cudaMalloc` lands on GPU 1, then registers a buffer that lives on GPU 0:

```
registerLocalMemory(buf_on_GPU0, "cuda:0") -> rc=0
cuCtxGetCurrent        0x…552a0 -> 0x…68cb0
cudaGetDevice                 1 -> 0
next cudaMalloc landed on GPU  1 -> 0        <-- silently on the wrong GPU
```

The registration path retains the buffer device's primary context, makes it current and deliberately never restores it, so every subsequent CUDA call on that application thread moves to the wrong device. It bites any process that registers buffers for a device other than the calling thread's — multi-GPU-per-process setups, mooncake-store, `--gpu_id=-1` — and it is exactly the failure mode `ScopedCudaContext` was introduced for. Wrapping that call site in the new RAII helper is ~3 lines. (The "the retain is intentionally not released" comment above `bindCudaContextIfNeeded` should then be narrowed, since `ScopedCudaContext::restore()` does release.)

## Nits (non-blocking)

* `ScopedCudaContext` is constructed before the pointer probes, so the host-only path makes CUDA driver calls it does not need, and `restore()` there is a no-op with an extra failure mode. Probing first and only constructing/binding when a device pointer is involved would make the host path completely CUDA-free. Note this is **not** a regression: I checked that a CUDA-enabled build cannot reach that path without a GPU anyway — with `CUDA_VISIBLE_DEVICES=""` all 32 NICs fail `fi_getinfo` ("No data available") and `installTransport("efa")` refuses to come up. Also worth recording: after a single `cudaPointerGetAttributes()` (which registration always performs) `cuCtxGetCurrent()` returns `CUDA_SUCCESS` with a null context and leaves device 0's primary context inactive, so the ctor's early-out is unreachable in practice.
* No test covers a cross-device loopback (dst on GPU A, src on GPU B), which is the `device = dst_device >= 0 ? dst_device : src_device` branch plus the `cudaMemcpyDefault`/UVA P2P copy. I ran one and it works correctly on both builds; a case in `efa_gpu_loopback_test` would lock that in.

Unrelated to this PR, spotted while running the bench: `transfer_engine_bench` SIGSEGVs (exit 139) when `--segment_id` carries an empty port — `Illegal port number in . Use default port 12001` → `Unable to get target segment ID` → crash. I can file that separately.

yejun-yun:
Fixed the issues you mentioned.

whn09 (approved):
LGTM

alogfans (approved):

Issue #3377, opened 2026-08-11
[RFC]: Classic TE RDMA Two-Sided Transport (`rdma_twosided`)


zhtshr:
## Summary

Add an **opt-in** classic Transfer Engine transport `rdma_twosided` for
PD-oriented workloads that prefer simpler buffer lifetime and TE-managed
bounce memory over peak one-sided bandwidth.

This is **not** a replacement for classic `rdma` (ibverbs WRITE/READ).
A process installs **either** `rdma` **or** `rdma_twosided`, not both.

Work is split into **four** reviewable PRs (large functional modules). Each PR
that introduces a runnable network path ships correctness tests plus a minimal
benchmark/smoke metric. End-to-end effectiveness vs one-sided is validated in
the final PR.

Related: #3324 (PR1, merged), #3440 (PR2, merged), #3613 (PR3, in review).

## Motivation

Classic one-sided RDMA requires applications to keep buffers registered for
the lifetime of inflight work and exposes remote rkeys to peers. For some PD
paths this is awkward:

1. **Buffer lifetime** — avoid repeated register/unregister around transfers.
2. **TE-owned networking buffers** — capacity can expand/shrink under credit.
3. **Decouple remote memory ownership** — SEND/RECV + local placement; no peer
   app rkey for the default path.
4. **Acceptable latency** — bounce/copy cost is OK for the target PD scenarios;
   peak GB/s remains on classic `rdma`.

## Non-goals

- Replacing classic `rdma` as the default high-bandwidth path.
- Implementing this behind libfabric for the first version (EFA/CXI already use
  OFI for their fabrics; this path targets verbs/RoCE/eRDMA and stays on
  ibverbs to align with classic `rdma`).
- Changing the existing TE handshake/discovery stack (only extend capability
  fields as needed).

## Architecture

```text
TransferEngine / MultiTransport
  ├─ install("rdma")           → classic one-sided (unchanged)
  └─ install("rdma_twosided")  → new transport (mutually exclusive)

rdma_twosided
  ├─ CtrlChannel (per-peer notify QP)  — typed CtrlFrames, credit, notify
  ├─ MsgChannel  (per-peer/rail msg QP) — SEND/RECV + bounce pool
  └─ TE managed buffers + credit admission (WAITING when short)
```

### PR1 — Protocol primitives (#3324, merged)
**Deliverable:** `CtrlFrame` codec + `SenderCreditLedger` under
`transport/rdma_twosided/`, unit tests only.

[...]

PR #3440, opened 2026-08-14
[TE] Add rdma_twosided control-plane notify channel

zhtshr:
## Description

Second slice of the classic TE two-sided work ([RFC #3377](https://github.com/kvcache-ai/Mooncake/issues/3377)). Follows merged PR1 (#3324).

Adds an **opt-in** `rdma_twosided` transport whose control plane is a per-peer `CtrlChannel` (RC QP SEND/RECV). Classic `rdma` stays one-sided; a process installs **either** `rdma` **or** `rdma_twosided`.

This PR:
- Adds `RdmaTwoSidedTransport` (`MC_USE_RDMA_TWOSIDED=1`) and wires MultiTransport mutual exclusion
- Implements `CtrlChannel` + handshake fields (`notify_qp_num` / `notify_rq_depth` / `ctrl_channel`)
- Routes `sendNotify*` over RDMA `NOTIFY_COMPAT` frames, with OOB fallback (`MC_RDMA_NOTIFY_OOB_FALLBACK`)
- Adds `rdma_notify_test` (correctness + latency smoke; self-skips without an RDMA device)

No MsgChannel / bounce / managed-buffer data path yet (PR3). TE `submitTransfer` two-sided path remains PR4.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
ninja ctrl_frame_test sender_credit_test rdma_notify_test
./mooncake-transfer-engine/tests/ctrl_frame_test
./mooncake-transfer-engine/tests/sender_credit_test
MC_TE_FILTERS=erdma_0 MC_TEST_DEVICE_NAME=erdma_0 MC_MTU=1024 \
  ./mooncake-transfer-engine/tests/rdma_notify_test
```

**Test results:**
- [x] Unit tests pass (`ctrl_frame_test` / `sender_credit_test`)
- [x] `rdma_notify_test` 4/4 on `erdma_0` (single / bidi / latency smoke ~10k ops/s / burst 128)

Notify latency smoke on eRDMA: 2000 notifies in ~191 ms, ~10.5k ops/s, avg ~95 µs.

## Checklist

- [x] I have run `pre-commit` on touched files and hooks pass

RFC: https://github.com/kvcache-ai/Mooncake/issues/3377

## AI Assistance Disclosure

- [ ] No AI tools were used

Cursor agent assisted implementing the `rdma_twosided` control-plane channel. Human submitter reviewed the PR2 scope (CtrlChannel + notify path only).

### DIFF
--- a/mooncake-transfer-engine/src/transport/rdma_twosided/ctrl_channel.cpp
+++ b/mooncake-transfer-engine/src/transport/rdma_twosided/ctrl_channel.cpp
@@ -65,7 +65,7 @@ CtrlChannel::CtrlChannel(RdmaTwoSidedTransport &transport, RdmaContext &context,
     max_pending_sends_ = cfg.rdma_notify_max_pending_sends;
 }
 
-CtrlChannel::~CtrlChannel() { destroyResources(); }
+CtrlChannel::~CtrlChannel() { disconnect(); }
 
 int CtrlChannel::construct() {
     std::lock_guard<std::mutex> lock(resource_mutex_);
@@ -151,13 +151,7 @@ int CtrlChannel::createResources() {
     return 0;
 }
 
-void CtrlChannel::destroyResources() {
-    connected_.store(false, std::memory_order_release);
-    {
-        std::lock_guard<std::mutex> lock(send_mutex_);
-        pending_sends_ = 0;
-        send_cv_.notify_all();
-    }
+void CtrlChannel::releaseIbResources() {
     if (qp_) {
         ibv_destroy_qp(qp_);
         qp_ = nullptr;
@@ -176,6 +170,25 @@ void CtrlChannel::destroyResources() {
     recv_mrs_.clear();
     recv_buffers_.clear();
     send_buffer_.clear();
+}
+
+void CtrlChannel::destroyResources() {
+    // Called from createResources() while resource_mutex_ is already held.
+    // Do not acquire send_mutex_ here (lock-order inversion with
+    // sendCtrlFrame).
+    connected_.store(false, std::memory_order_release);
+    releaseIbResources();
+}
+
+void CtrlChannel::disconnect() {
+    connected_.store(false, std::memory_order_release);
+    {
+        std::lock_guard<std::mutex> lock(send_mutex_);
+        pending_sends_ = 0;
+        send_cv_.notify_all();
+    }
+    std::lock_guard<std::mutex> lock(resource_mutex_);
+    releaseIbResources();
 }
 
 int CtrlChannel::postRecv(size_t idx) {
@@ -570,11 +583,6 @@ int CtrlChannel::pollSendCompletions(int max_entries) {
 
 int CtrlChannel::pollCompletions(int max_entries) {
     return drainCompletions(max_entries, /*log_poll_error=*/true);
-}
-
-void CtrlChannel::disconnect() {
-    std::lock_guard<std::mutex> lock(resource_mutex_);
-    destroyResources();
 }
 
 }  // namespace mooncake
--- a/mooncake-transfer-engine/src/transport/rdma_twosided/ctrl_channel.cpp
+++ b/mooncake-transfer-engine/src/transport/rdma_twosided/ctrl_channel.cpp
@@ -471,6 +471,9 @@ int CtrlChannel::sendCtrlFrame(const CtrlFrame &frame_in) {
         }
         pending_sends_++;
     }
+    // Queued, not delivered: the SEND completion (handled in drainCompletions)
+    // decides actual success. A failed WC drops connected_, which routes the
+    // next notify to a fresh channel or to OOB fallback.
     return 0;
 }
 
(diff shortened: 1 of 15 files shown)


mooncake-transfer-engine/src/transport/rdma_twosided/ctrl_plane.cpp
```diff
@@ -0,0 +1,179 @@
...
+    // Serialize active connect per peer to avoid duplicate handshakes.
+    auto channel =
+        std::make_shared<CtrlChannel>(*this, *contexts[0], peer_server_name);
+    // Release lock during handshake I/O; re-check afterwards.
+    lock.unlock();
+    if (channel->connectActive()) {
+        return nullptr;
+    }
+    lock.lock();
+    auto again = ctrl_channels_.find(peer_server_name);
```
alogfans: Does multiple callers can still create channels concurrently?

mooncake-transfer-engine/src/transport/rdma_twosided/ctrl_channel.cpp
```diff
@@ -0,0 +1,580 @@
...
+}
+
+int CtrlChannel::pollCompletions(int max_entries) {
+    return drainCompletions(max_entries, /*log_poll_error=*/true);
+}
+
+void CtrlChannel::disconnect() {
+    std::lock_guard<std::mutex> lock(resource_mutex_);
+    destroyResources();
+}
```
alogfans: This introduces a lock-order inversion: sendCtrlFrame() holds send_mutex_ before acquiring resource_mutex_, while disconnect() holds resource_mutex_ and then calls destroyResources(), which acquires send_mutex_. Concurrent send and channel replacement can therefore deadlock. Please use a consistent lock order or split shutdown into non-nested state-transition and resource-destruction phases.

mooncake-transfer-engine/src/transport/rdma_twosided/ctrl_plane.cpp
zhtshr: Yes. The old re-check after handshake did not serialize in-flight connects.

We now insert a placeholder into ctrl_channels_ before the handshake. Other callers wait on ctrl_cv_ until the entry is connected, replaced, or erased, instead of starting another connectActive().

mooncake-transfer-engine/src/transport/rdma_twosided/ctrl_channel.cpp
zhtshr: Agreed. sendCtrlFrame() took send_mutex_ then resource_mutex_, while disconnect() nested them the other way.

disconnect() no longer nests the locks: it wakes send waiters, drops send_mutex_, then tears down IB resources under resource_mutex_ only.

mooncake-transfer-engine/src/transport/rdma_twosided/ctrl_plane.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+    std::unique_lock<std::mutex> lock(ctrl_mutex_);
+    while (true) {
+        auto it = ctrl_channels_.find(peer_server_name);
+        if (it != ctrl_channels_.end() && it->second) {
+            if (it->second->connected()) return it->second;
+            // Active or passive connect is already in flight for this peer.
+            ctrl_cv_.wait(lock);
+            continue;
+        }
```
alogfans, line 181: How to avoid ctrl_cv_ wait indefinitely?

mooncake-transfer-engine/src/transport/rdma_twosided/ctrl_channel.cpp
```diff
@@ -0,0 +1,588 @@
...
+                       << peer_server_name_ << ": " << strerror(std::abs(ret))
+                       << " [" << ret << "]";
+            return ERR_ENDPOINT;
+        }
+        pending_sends_++;
+    }
+    return 0;
+}
+
+int CtrlChannel::sendNotify(const NotifyDesc &notify) {
```
alogfans, line 480: This should be called after receiving WCs instead of submitting WRs.

mooncake-transfer-engine/src/transport/rdma_twosided/ctrl_plane.cpp
zhtshr: The wait could block forever because a single connected_ bool couldn't distinguish "connecting" from "dead": a channel killed by an error WC kept its map entry, so the next caller mistook it for an in-flight connect and parked on ctrl_cv_, which disconnect()/error-WC never notify.

Fixed: a refcounted ctrl_connecting_ marks genuine in-flight connects (refcounted since an active and a passive connect to the same peer can overlap), so an entry that is !connected() and unmarked is dead and gets reclaimed by the next caller. Every wait is now bounded by wait_until (MC_RDMA_NOTIFY_CONNECT_TIMEOUT_MS, default 10s), returning nullptr so the caller falls back to OOB. ConnectScope + UnlockGuard guarantee the placeholder is retracted and waiters woken on all paths, including exceptions. Added a regression test that hung before the fix.

mooncake-transfer-engine/src/transport/rdma_twosided/ctrl_channel.cpp
zhtshr: Right, success shouldn't be concluded at ibv_post_send. The channel stays fire-and-forget, so return 0 now explicitly means queued, not delivered — delivery is settled on the SEND WC.

What that changes in practice: an error WC drops connected_, and together with the dead-entry reclaim above, the next sendNotify no longer blocks on the stale channel — it returns an error and falls back to OOB notify. Previously that path hung in ensureCtrlChannel, so the OOB fallback was unreachable. Also settled the in-flight pending_sends_ accounting on the error-WC path, which was never decremented.

alogfans (approved):
LGTM, thanks!

PR #3441, opened 2026-08-14
[Reshard] Add N-D logical weight transfer planner

Bo-Vincent:
## Description

This PR adds the address-free logical-planning phase for heterogeneous
model-weight resharding in the new `mooncake-reshard` module.

It consumes complete typed source and target placements, or a committed logical
`WeightManifest` Store snapshot as the source, and produces a backend-neutral
`LogicalTransferPlan`. The plan contains compact N-D `TransferRegion` values:
logical overlap offsets and shapes, source/target base byte offsets,
`inner_bytes`, outer loop counts, and source/target strides.

The planner supports different source and target TP / PP / EP / DP layouts
without embedding framework-specific tensor-name parsing or model-layout
inference in Mooncake core:

- TP and EP use N-D logical-box split and merge, including different source
  and target shard dimensions.
- PP routes through explicit, framework-provided tensor or layer ownership;
  it does not infer stages from parameter names or layer-count formulas.
- EP is represented by a logical expert coordinate, so independent expert
  allocations remain independent fragments rather than being packed or
  all-gathered.
- DP selects a complete source replica or follows explicit DP ownership
  without changing tensor geometry.

The planner validates canonical resource identity, revision, weight
generation, tensor descriptors, ownership, logical coverage, aliases, and
planning limits. It keeps N-D regions compact and fails closed when a later
segment expansion would exceed its explicit limit.

This PR intentionally does not add runtime GPU binding, address or allocation
validation, lease attestation, Mooncake Store persistence/lifecycle, Transfer
Engine lowering/execution, model-specific packing or quantization conversion,
or framework runtime-object parsing. Those belong to later phases.

Related RFC: #3111

## Module

- [ ] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [x] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
  python -m pytest -q mooncake-reshard/tests

pyright --project mooncake-reshard/pyrightconfig.json

```

**Test results:**
- [x] Unit tests pass: `284 passed`
- [ ] Integration tests pass (if applicable): not applicable; this PR produces
  an address-free logical plan and adds no runtime Store or Transfer Engine
  execution path.

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex was used to inspect the existing manifest and planner boundaries, draft
tests and documentation, and run local validation. The submitter reviewed the
design, implementation, and results.

### DIFF
--- a/mooncake-reshard/python/mooncake/reshard/_typing.py
+++ b/mooncake-reshard/python/mooncake/reshard/_typing.py
@@ -1,0 +1,10 @@
+"""Typing compatibility definitions for the supported Python runtimes."""
+
+import sys
+
+if sys.version_info >= (3, 10):
+    from typing import TypeAlias
+else:
+    from typing_extensions import TypeAlias
+
+__all__ = ["TypeAlias"]
--- a/mooncake-reshard/python/mooncake/reshard/geometry.py
+++ b/mooncake-reshard/python/mooncake/reshard/geometry.py
@@ -1,0 +1,265 @@
+"""Resource-neutral N-D logical-box helpers for reshard planners."""
+
+from __future__ import annotations
+
+import heapq
+from array import array
+from math import prod
+from typing import Iterable, Protocol, Sequence
+
+from ._compat import _strict_zip
+
+
+_MAX_HIGH_DIMENSIONAL_PAIRWISE_COMPARISONS = 1_000_000
+
+
+class LogicalBox(Protocol):
+    """An address-free logical box supplied by a resource-specific adapter."""
+
+    @property
+    def global_offset(self) -> tuple[int, ...]: ...
+
+    @property
+    def local_shape(self) -> tuple[int, ...]: ...
+
+
+class OverlapRegion(Protocol):
+    """The logical overlap portion of a resource-specific transfer region."""
+
+    @property
+    def overlap_offset(self) -> tuple[int, ...]: ...
+
+    @property
+    def overlap_shape(self) -> tuple[int, ...]: ...
+
+
+def box_contains(
+    outer_offset: tuple[int, ...],
+    outer_shape: tuple[int, ...],
+    inner_offset: tuple[int, ...],
+    inner_shape: tuple[int, ...],
+) -> bool:
+    """Return whether one N-D logical box is wholly inside another."""
+
+    return all(
+        outer_begin <= inner_begin
+        and inner_begin + inner_extent <= outer_begin + outer_extent
+        for outer_begin, outer_extent, inner_begin, inner_extent in _strict_zip(
+            outer_offset,
+            outer_shape,
+            inner_offset,
+            inner_shape,
+        )
+    )
+
+
+def boxes_overlap(
+    boxes: Sequence[tuple[tuple[int, ...], tuple[int, ...]]],
+) -> bool:
+    """Return whether any two N-D boxes overlap.
+
+    One-dimensional intervals and two-dimensional rectangles use bounded sweep
+    algorithms. Higher-dimensional boxes retain full intersection semantics but
+    fail closed after an explicit pairwise-comparison budget rather than making
+    unbounded validation work part of the public manifest contract.
+    """
+
+    if len(boxes) < 2:
+        return False
+    ndim = len(boxes[0][0])
+    if ndim == 1:
+        return _one_dimensional_boxes_overlap(boxes)
+    if ndim == 2:
+        return _two_dimensional_boxes_overlap(boxes)
+    return _high_dimensional_boxes_overlap(boxes)
+
+
+def _one_dimensional_boxes_overlap(
+    boxes: Sequence[tuple[tuple[int, ...], tuple[int, ...]]],
+) -> bool:
+    ordered = sorted((offset[0], offset[0] + shape[0]) for offset, shape in boxes)
+    previous_end = ordered[0][1]
+    for begin, end in ordered[1:]:
+        if begin < previous_end:
+            return True
+        previous_end = max(previous_end, end)
+    return False
+
+
+class _RangeMaxTree:
+    """Coordinate-compressed range-add/range-maximum tree for 2-D sweeps."""
+
+    def __init__(self, size: int) -> None:
+        if size <= 0:
+            raise ValueError("range tree size must be positive")
+        capacity = size * 4
+        self._size = size
+        self._maximum = array("i", [0]) * capacity
+        self._lazy = array("i", [0]) * capacity
+
+    def add(self, left: int, right: int, value: int) -> None:
+        self._add(1, 0, self._size - 1, left, right, value)
+
+    def maximum(self, left: int, right: int) -> int:
+        return self._maximum_in(1, 0, self._size - 1, left, right)
+
+    def _add(
+        self,
+        node: int,
+        begin: int,
+        end: int,
+        left: int,
+        right: int,
+        value: int,
+    ) -> None:
+        if left <= begin and end <= right:
+            self._maximum[node] += value
+            self._lazy[node] += value
+            return
+        middle = (begin + end) // 2
+        if left <= middle:
+            self._add(node * 2, begin, middle, left, right, value)
+        if right > middle:
+            self._add(node * 2 + 1, middle + 1, end, left, right, value)
+        self._maximum[node] = self._lazy[node] + max(
+            self._maximum[node * 2], self._maximum[node * 2 + 1]
+        )
+
+    def _maximum_in(
+        self,
+        node: int,
+        begin: int,
+        end: int,
+        left: int,
+        right: int,
+    ) -> int:
+        if left <= begin and end <= right:
+            return self._maximum[node]
+        middle = (begin + end) // 2
+        result = 0
+        if left <= middle:
+            result = self._maximum_in(node * 2, begin, middle, left, right)
+        if right > middle:
+            result = max(
+                result,
+                self._maximum_in(node * 2 + 1, middle + 1, end, left, right),
+            )
+        return self._lazy[node] + result
+
+
+def _two_dimensional_boxes_overlap(
+    boxes: Sequence[tuple[tuple[int, ...], tuple[int, ...]]],
+) -> bool:
+    coordinates = sorted(
+        {
+            coordinate
+            for offset, shape in boxes
+            for coordinate in (offset[1], offset[1] + shape[1])
+        }
+    )
+    coordinate_index = {
+        coordinate: index for index, coordinate in enumerate(coordinates)
+    }
+    tree = _RangeMaxTree(len(coordinates) - 1)
+    events: list[tuple[int, int, int, int]] = []
+    for offset, shape in boxes:
+        x_begin = offset[0]
+        x_end = offset[0] + shape[0]
+        y_begin = coordinate_index[offset[1]]
+        y_end = coordinate_index[offset[1] + shape[1]] - 1
+        events.append((x_end, 0, y_begin, y_end))
+        events.append((x_begin, 1, y_begin, y_end))
+
+    for _, is_start, y_begin, y_end in sorted(events):
+        if not is_start:
+            tree.add(y_begin, y_end, -1)
+            continue
+        if tree.maximum(y_begin, y_end) > 0:
+            return True
+        tree.add(y_begin, y_end, 1)
+    return False
+
+
+def _high_dimensional_boxes_overlap(
+    boxes: Sequence[tuple[tuple[int, ...], tuple[int, ...]]],
+) -> bool:
+    sweep_dim = max(
+        range(len(boxes[0][0])),
+        key=lambda dim: len(
+            {(offset[dim], offset[dim] + shape[dim]) for offset, shape in boxes}
+        ),
+    )
+    ordered = sorted(
+        enumerate(boxes),
+        key=lambda item: item[1][0][sweep_dim],
+    )
+    active: dict[int, tuple[tuple[int, ...], tuple[int, ...]]] = {}
+    expirations: list[tuple[int, int]] = []
+    comparisons = 0
+    for index, (offset, shape) in ordered:
+        begin = offset[sweep_dim]
+        while expirations and expirations[0][0] <= begin:
+            _, expired_index = heapq.heappop(expirations)
+            active.pop(expired_index, None)
+        for candidate_offset, candidate_shape in active.values():
+            if comparisons >= _MAX_HIGH_DIMENSIONAL_PAIRWISE_COMPARISONS:
+                raise ValueError(
+                    "high-dimensional overlap budget exceeded before validation"
+                )
+            comparisons += 1
+            if all(
+                left_begin < right_begin + right_extent
+                and right_begin < left_begin + left_extent
+                for left_begin, left_extent, right_begin, right_extent in _strict_zip(
+                    candidate_offset,
+                    candidate_shape,
+                    offset,
+                    shape,
+                )
+            ):
+                return True
+        active[index] = (offset, shape)
+        heapq.heappush(expirations, (offset[sweep_dim] + shape[sweep_dim], index))
+    return False
+
+
+def boxes_exactly_cover(
+    container_offset: tuple[int, ...],
+    container_shape: tuple[int, ...],
+    boxes: Sequence[tuple[tuple[int, ...], tuple[int, ...]]],
+) -> bool:
+    """Require in-bounds, non-overlapping boxes with exact logical volume."""
+
+    if not boxes:
+        return False
+    if any(
+        not box_contains(container_offset, container_shape, offset, shape)
+        for offset, shape in boxes
+    ):
+        return False
+    if sum(prod(shape) for _, shape in boxes) != prod(container_shape):
+        return False
+    return not boxes_overlap(boxes)
+
+
+def regions_exactly_cover(
+    target: LogicalBox,
+    regions: Iterable[OverlapRegion],
+) -> bool:
+    """Require resource-neutral N-D regions to completely cover one target box."""
+
+    return boxes_exactly_cover(
+        target.global_offset,
+        target.local_shape,
+        tuple((region.overlap_offset, region.overlap_shape) for region in regions),
+    )
+
+
+__all__ = [
+    "LogicalBox",
+    "OverlapRegion",
+    "box_contains",
+    "boxes_exactly_cover",
+    "boxes_overlap",
+    "regions_exactly_cover",
+]
--- a/mooncake-reshard/python/mooncake/reshard/weight/_planner/api.py
+++ b/mooncake-reshard/python/mooncake/reshard/weight/_planner/api.py
@@ -1,0 +1,164 @@
+from __future__ import annotations
+
+from typing import Optional
+
+from ...contracts import ParticipantId
+from ..manifest import WeightPlacementManifest
+from ..storage_manifest import WeightManifest
+from .contracts import LogicalTransferPlan, PlanningLimits
+from .core import (
+    _collect_placements,
+    _logical_transfer_plan,
+    _plan_transfer,
+)
+
+
+def plan_placement_transfer(
+    source_placement: WeightPlacementManifest,
+    target_placement: WeightPlacementManifest,
+    *,
+    planning_limits: Optional[PlanningLimits] = None,
+) -> LogicalTransferPlan:
+    """Plan a reshard between two complete address-free placements."""
+
+    source_tensors, source_fragments = _collect_placements(
+        (source_placement,),
+        "source",
+    )
+    target_tensors, target_fragments = _collect_placements(
+        (target_placement,),
+        "target",
+    )
+    source = source_placement
+    target = target_placement
+    if source.resource_id != target.resource_id:
+        raise ValueError("source and target resource_id differ")
+    if source.revision != target.revision:
+        raise ValueError("source and target revision differ")
+    if source.weight_generation != target.weight_generation:
+        raise ValueError("source and target weight_generation differ")
+    transfer = _plan_transfer(
+        source.resource_id,
+        source.revision,
+        source_tensors,
+        source_fragments,
+        target_tensors,
+        target_fragments,
+        planning_limits=planning_limits,
+    )
+    return _logical_transfer_plan(
+        transfer,
+        source_tensors=source_tensors,
+        target_tensors=target_tensors,
+        source_placement=source_placement,
+        source_manifest=None,
+        target_placement=target_placement,
+    )
+
+
+def plan_placement_transfer_to_local_target(
+    source_placement: WeightPlacementManifest,
+    target_placement: WeightPlacementManifest,
+    target_participant_id: Optional[ParticipantId] = None,
+    *,
+    planning_limits: Optional[PlanningLimits] = None,
+) -> LogicalTransferPlan:
+    """Plan one target executor using address-free source and target layouts."""
+
+    source_tensors, source_fragments = _collect_placements(
+        (source_placement,),
+        "source",
+    )
+    if target_participant_id is None:
+        if target_placement.topology.world_size != 1:
+            raise ValueError(
+                "target_participant_id is required for a multi-participant target"
+            )
+        target_participant_id = target_placement.parts[0].participant_id
+    try:
+        target_part = next(
+            part
+            for part in target_placement.parts
+            if part.participant_id == target_participant_id
+        )
+    except StopIteration as error:
+        raise ValueError(
+            f"unknown target participant: {target_participant_id}"
+        ) from error
+    target_tensors = {tensor.tensor_id: tensor for tensor in target_part.tensors}
+    target_fragments = list(target_part.fragments)
+    source = source_placement
+    if source.resource_id != target_placement.resource_id:
+        raise ValueError("source and target resource_id differ")
+    if source.revision != target_placement.revision:
+        raise ValueError("source and target revision differ")
+    if source.weight_generation != target_placement.weight_generation:
+        raise ValueError("source and target weight_generation differ")
+    transfer = _plan_transfer(
+        source.resource_id,
+        source.revision,
+        source_tensors,
+        source_fragments,
+        target_tensors,
+        target_fragments,
+        local_target=True,
+        planning_limits=planning_limits,
+    )
+    result = _logical_transfer_plan(
+        transfer,
+        source_tensors=source_tensors,
+        target_tensors=target_tensors,
+        source_placement=source_placement,
+        source_manifest=None,
+        target_placement=target_placement,
+        target_participant_ids=frozenset({target_participant_id}),
+    )
+    if len(result.target_executors) != 1:
+        raise ValueError("local target placement must describe exactly one executor")
+    return result
+
+
+def plan_stored_transfer_to_target_placement(
+    source_manifest: WeightManifest,
+    target_placement: WeightPlacementManifest,
+    *,
+    planning_limits: Optional[PlanningLimits] = None,
+) -> LogicalTransferPlan:
+    """Plan a Store-backed transfer into address-free target placements."""
+
+    target_tensors, target_fragments = _collect_placements(
+        (target_placement,),
+        "target",
+    )
+    target = target_placement
+    if source_manifest.resource_id != target.resource_id:
+        raise ValueError("source and target resource_id differ")
+    if source_manifest.revision != target.revision:
+        raise ValueError("source and target revision differ")
+    if source_manifest.weight_generation != target.weight_generation:
+        raise ValueError("source and target weight_generation differ")
+    source_tensors = {tensor.tensor_id: tensor for tensor in source_manifest.tensors}
+    transfer = _plan_transfer(
+        source_manifest.resource_id,
+        source_manifest.revision,
+        source_tensors,
+        source_manifest.fragments,
+        target_tensors,
+        target_fragments,
+        planning_limits=planning_limits,
+    )
+    return _logical_transfer_plan(
+        transfer,
+        source_tensors=source_tensors,
+        target_tensors=target_tensors,
+        source_placement=None,
+        source_manifest=source_manifest,
+        target_placement=target_placement,
+    )
+
+
+__all__ = [
+    "plan_placement_transfer",
+    "plan_placement_transfer_to_local_target",
+    "plan_stored_transfer_to_target_placement",
+]
--- a/mooncake-reshard/python/mooncake/reshard/weight/_planner/contracts.py
+++ b/mooncake-reshard/python/mooncake/reshard/weight/_planner/contracts.py
@@ -1,0 +1,674 @@
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from math import prod
+from types import MappingProxyType
+from typing import (
+    ClassVar,
+    Generic,
+    Iterable,
+    Mapping,
+    Optional,
+    Sequence,
+    TypeVar,
+    cast,
+)
+
+from ..._compat import _strict_zip
+from ..._typing import TypeAlias
+from ...contracts import (
+    ParticipantId,
+    PlacementFragmentId,
+    PlacementId,
+    ResourceId,
+    RevisionId,
+    TensorId,
+)
+from ...geometry import box_contains as _box_contains
+from ..manifest import (
+    ParallelRank,
+    PlacementFragment,
+    TensorDescriptor,
+    WeightPlacementManifest,
+)
+from ..storage_manifest import StoredFragment
+from .geometry import (
+    _derive_region_geometry,
+    _fragment_itemsize,
+    _validate_outer_strides,
+)
+from .fragments import (
+    GeometryFragment,
+    LogicalSourceFragment,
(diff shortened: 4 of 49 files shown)


mooncake-reshard/python/mooncake/reshard/weight/_planner/ownership.py
```diff
@@ -0,0 +1,194 @@
...
+        raise ValueError("target manifests have no fragments")
+    fragments_by_dp_and_tensor: dict[
+        int, dict[TensorId, list[LogicalTargetFragment]]
+    ] = {}
+    for fragment in target_fragments:
+        fragments_by_dp_and_tensor.setdefault(fragment.rank.dp, {}).setdefault(
+            fragment.tensor_id, []
+        ).append(fragment)
+    dp_ranks = sorted(fragments_by_dp_and_tensor)
+    for dp_rank in dp_ranks:
```
zxpdemonio: **[P1] DP ownership is being treated as full replication here.** The public manifest accepts `OwnershipAxis(kind="dp")`, and the existing `TensorParallelism` contract uses DP as an ownership/scope axis. A valid placement with tensor A owned by DP0 and tensor B owned by DP1 passes `WeightPlacementManifest` construction, but even an identity plan is rejected here with `target tensor is not fully covered: b: dp=0`. Please branch on the tensor's declared DP semantics: require complete per-rank coverage only for `ReplicatedAxis("dp")`; for `OwnershipAxis("dp")`, validate and route by the declared owner. Otherwise this cannot losslessly replace the existing parallel-tensor planner.

mooncake-reshard/python/mooncake/reshard/weight/_planner/core.py
```diff
@@ -0,0 +1,489 @@
...
+
+        if not _boxes_exactly_cover(
+            target.global_offset,
+            target.local_shape,
+            tuple((offset, shape) for offset, shape, _ in overlaps),
+        ):
+            raise ValueError(
+                f"target fragment is not fully covered: {target.fragment_id}"
+            )
+        operations.extend(
```
zxpdemonio: **[P1] The number of materialized operations has no upper bound.** Orthogonal fragmentations produce a Cartesian product: N source stripes on dim 0 and N target stripes on dim 1 create N^2 `TransferRegion` objects. On this head I measured 32 -> 1,024, 64 -> 4,096, and 128 -> 16,384 operations. A valid large manifest can therefore exhaust memory before any downstream batching applies. Please add a total planning/operation budget and fail closed before materialization, or expose a streaming/batched planning form. This is also called out as required by RFC #3111.

mooncake-reshard/python/mooncake/reshard/weight/_planner/contracts.py
```diff
@@ -0,0 +1,888 @@
...
+                for count, stride in _strict_zip(
+                    self.outer_loop_counts, self.target_strides
+                )
+            )
+            + self.inner_bytes
+        )
+        if target_end > self.target.nbytes:
+            raise ValueError("transfer region exceeds target fragment")
+
+    def iter_segments(self) -> Iterable[tuple[int, int, int]]:
```
zxpdemonio: **[P1] This iterator is lazy, but it is not bounded.** A single valid region can have `prod(outer_loop_counts)` segments; the current cross-dimension test already creates 64 regions with 8,192 segments each. An executor that consumes this iterator normally, or wraps it in `list()`, can spend unbounded time/memory on a valid manifest. Please make the lowering API accept an explicit segment/byte budget or return bounded batches, and fail before exceeding it. The design document should not call this "bounded lazy lowering" until the bound is enforced.

mooncake-reshard/python/mooncake/reshard/weight/_planner/contracts.py
```diff
@@ -0,0 +1,888 @@
...
+                (
+                    operation.target.instance_id,
+                    operation.target.worker_id,
+                    operation.target.device,
+                    operation.target.binding.storage_address,
+                    operation.target.binding.storage_nbytes,
+                )
+            )
+
+    for source in source_allocations:
```
zxpdemonio: **[P1] This overlap check is O(number_of_source_allocations * number_of_target_allocations).** It still compares every pair when source and target have different `instance_id` values, which is the common G2G case where overlap is impossible. With tens of thousands of weight fragments this becomes hundreds of millions or billions of comparisons during every `TransferPlan` construction. Please group allocations by `(instance_id, worker_id, device)` first and use a sorted interval sweep within matching address spaces.

mooncake-reshard/python/mooncake/reshard/weight/_planner/fragments.py
```diff
@@ -0,0 +1,238 @@
...
+class BoundWeightFragment:
+    """A placement fragment bound to one live runtime allocation and lease."""
+
+    placement: PlacementFragment
+    binding: RuntimeBindingFragment
+    instance_id: RuntimeInstanceId
+    runtime_lease_id: LeaseId
+    lease_generation: int
+    owner: object | None = field(default=None, compare=False, repr=False)
+    attestation: RuntimeBindingAttestation | None = field(
```
zxpdemonio: **[P1] An executable plan can currently retain no allocation/registration owner at all.** `owner=None` is accepted (and is the default used by the planner tests), while `lease_id` and `generation` are only snapshots and are not revalidated against a live lease service here. If a local source or target allocation is freed or unregistered after binding and before submission, the plan still carries a stale address. Please require either a concrete lifetime token for local bindings or an executor-side lease acquisition/revalidation that is atomic with submission; without one of those, the executable contract does not provide the lifetime guarantee described by the RFC.

mooncake-reshard/python/mooncake/reshard/weight/_planner/core.py
```diff
@@ -0,0 +1,489 @@
...
+        operations=tuple(operations),
+    )
+
+
+def _build_pipeline_routes(
+    operations: Sequence[LogicalTransferOperation | LiveTransferOperation],
+) -> tuple[PipelineRouteGroup, ...]:
+    indices_by_route: dict[tuple[int, int], list[int]] = {}
+    for index, operation in enumerate(operations):
+        source_pp = operation.source.rank.pp
```
zxpdemonio: **[P2] This loses the existing PP `stage_id` identity by routing only on `rank.pp`.** `TensorParallelism` stores `stage_id` independently from PP rank and includes it in the shard identity, while the new descriptor has no corresponding field. If the intended compatibility contract is `stage_id == pp_rank`, the adapter must enforce and test that invariant; otherwise virtual/interleaved PP stages on the same PP rank collapse here. Please settle this mapping before replacing the current Store planner.

mooncake-reshard/python/mooncake/reshard/weight/_planner/contracts.py
```diff
@@ -0,0 +1,1010 @@
...
+            )
+        _validate_operation_budget(self.operations, self.limits)
+        for operation in self.operations:
+            _validate_bound_operation(operation)
+        if not all(
+            isinstance(participant, BoundParticipantPlan)
+            for participant in (*self.source_participants, *self.target_participants)
+        ):
+            raise ValueError("bound transfer plan has invalid participant metadata")
+        _validate_binding_provenance(
```
Aionw: **[P1] Revalidate target coverage in the public bound-plan contract.** These checks establish that each retained operation is attested and that participant metadata agrees with those operations, but they never establish that the retained operations completely cover the selected target placement. On this head I built a normal TP2 → TP1 plan with two operations / 64 bytes, then constructed a `BoundTransferPlan` from just the first 32-byte operation with correctly narrowed source/target participants and pipeline route. Construction succeeded, and `resolve_bound_participant_plan()` succeeded on both sides. An empty plan is accepted as well. Because this is an exported public contract, a caller can bypass the binder's coverage check while still presenting internally valid provenance. Please carry enough target-placement/selection identity to rerun exact fragment coverage here, or make direct construction unavailable so only the validated binder can create this type.

mooncake-reshard/python/mooncake/reshard/weight/_planner/validation.py
```diff
@@ -0,0 +1,555 @@
...
+        )
+        if identity in seen:
+            continue
+        seen.add(identity)
+        result.append(operation)
+
+    by_target: dict[CopyIdentity, BoundTransferOperation] = {}
+    deduplicated: list[BoundTransferOperation] = []
+    for operation in result:
+        identity = _target_copy_identity(operation)
```
Aionw: **[P1] Do not deduplicate across distinct runtime-binding lifetimes.** `_target_copy_identity()` omits the target participant, placement, `runtime_lease_id`, and attestation. I created two valid PP alias targets with the same instance/worker/device/address/generation but distinct lease IDs and non-null owners. This lookup merged them, and the resulting plan retained only one operation, one target participant, and one participant's lease/attestation; the other owner, lease, and completion route disappeared. Either restrict physical-copy deduplication to the same attested binding/lifetime scope, or preserve every aliased target participant's lease/owner/completion provenance separately even when only one physical write is needed.

mooncake-reshard/python/mooncake/reshard/weight/_planner/contracts.py
```diff
@@ -0,0 +1,1010 @@
...
+            binding.participant_id,
+            placement.digest,
+            binding.lease_id,
+            fragment.worker_id,
+            fragment.rank,
+        )
+        expected_indices.setdefault(key, []).append(index)
+        lease = RuntimeLeaseSnapshot.from_fragment(fragment)
+        leases = expected_leases.setdefault(key, [])
+        if lease not in leases:
```
Aionw: **[P2] Avoid quadratic lease-snapshot deduplication.** `leases` is a list, so `lease not in leases` performs a linear dataclass comparison for every operation. With `F` distinct fragments under one participant/worker key this performs roughly `F²/2` comparisons during every bound-plan construction, which is material for model workers with thousands of weight fragments. Please use a dictionary keyed by `fragment_id` (and reject a conflicting snapshot for an existing ID) so this remains linear before the final sort.

Aionw (changes requested):
Requesting changes for two additional correctness/lifetime gaps reproduced on head `f20c4e6`, plus one remaining quadratic validation path. The first reproduction reduced a valid TP2→TP1 plan from 64 to 32 bytes while keeping participant and route metadata internally consistent; construction and both participant-resolution calls still succeeded. The second used two PP alias targets with distinct leases and non-null owners sharing one accepted physical address; binding retained only one target participant. Details are inline.

mooncake-reshard/python/mooncake/reshard/weight/_planner/ownership.py
Bo-Vincent: Fixed the DP ownership handling.

`ReplicatedAxis(kind="dp")` still permits selecting one complete source replica.
`OwnershipAxis(kind="dp")` now uses the declared tensor owner instead and no
longer requires every tensor to appear on every DP rank.

The coverage and source-selection logic were updated in the canonical planner,
with regression coverage in
`test_dp_ownership_routes_each_tensor_without_requiring_a_full_dp_replica`.

mooncake-reshard/python/mooncake/reshard/weight/_planner/core.py
Bo-Vincent: Added a plan-level materialization budget.

`PlanningLimits.max_transfer_regions` now bounds the total number of
`TransferRegion` objects that logical planning may construct. The planner
counts the planned overlap regions and rejects the plan before unbounded
materialization when the configured budget would be exceeded.

This is fail-closed and is covered by regression tests for cross-dimension
Cartesian overlap growth.

mooncake-reshard/python/mooncake/reshard/weight/_planner/contracts.py
Bo-Vincent: Added bounded lowering for N-D regions.

`iter_segments()` now requires an explicit `max_segments` argument, and
`PlanningLimits.max_lowered_segments` bounds the total flattened segment count
for both logical and bound plans. Calls that would exceed the declared budget
fail before unbounded expansion.

The design documentation now describes this as budgeted lowering rather than
unconditionally bounded lazy lowering.

mooncake-reshard/python/mooncake/reshard/weight/_planner/contracts.py
Bo-Vincent: Reworked cross-side allocation validation to avoid the previous O(S * T)
comparison pattern.

The validator now groups allocations by address space, deduplicates identical
storage intervals, sorts each side, and uses a two-pointer interval sweep.
It preserves the same fail-closed behavior for in-place source/target storage
overlap.

`test_cross_side_allocation_validation_uses_a_linear_interval_sweep` verifies
the comparison count remains linear in the number of distinct intervals.

mooncake-reshard/python/mooncake/reshard/weight/_planner/fragments.py
```diff
@@ -0,0 +1,238 @@
...
+        return self.binding.storage_offset_bytes
+
+
+# A logical plan uses address-free placement fragments. An executable live plan
+# may only use fragments that have an address and a lease-generation fence.
+LogicalSourceFragment: TypeAlias = PlacementFragment
+LogicalTargetFragment: TypeAlias = PlacementFragment
+LiveSourceFragment: TypeAlias = BoundWeightFragment
+LiveTargetFragment: TypeAlias = BoundWeightFragment
+GeometryFragment: TypeAlias = PlacementFragment | BoundWeightFragment
```
Bo-Vincent: Updated the reshard package to keep its public runtime imports compatible with
Python 3.9.

The production package no longer relies on runtime PEP 604 aliases for this
surface, both Pyright configurations target Python 3.9, and a dedicated
Python 3.9 import gate is included in the repository checks.

Validation includes a real Python 3.9 import, full Python 3.9 test execution,
strict positive Pyright checks, and expected-failure negative type fixtures.

mooncake-reshard/python/mooncake/reshard/weight/_planner/fragments.py
Bo-Vincent: Implemented Route A in the current head.

`bind_logical_transfer_plan()` now returns `BoundTransferPlan`, not an
executable transfer request. The bound plan is an attested
placement-and-runtime snapshot: it has no `execute` or `submit` API, and it
does not claim to retain allocation lifetime.

The framework `owner` reference remains opaque and is explicitly not treated
as a lifetime token. A later TE executor must acquire real source/target
allocation guards, revalidate the selected bindings before submission, and
release the guards after completion.

Coverage was added in
`test_bound_plan_is_not_an_executable_allocation_lifetime_token`.

mooncake-reshard/python/mooncake/reshard/weight/_planner/core.py
Bo-Vincent: Preserved explicit PP stage semantics in placement metadata.

`PlacementFragment` now carries optional `pipeline_stage_id`. Pipeline routing
uses `(source_pp, source_pipeline_stage_id, target_pp,
target_pipeline_stage_id)` rather than only `(source_pp, target_pp)`. For a
simple pipeline, an omitted stage remains compatible and defaults to `rank.pp`.

The field participates in canonical placement identity when present,
round-trips through serde, and is covered by a virtual-stage routing
regression test.

mooncake-reshard/python/mooncake/reshard/weight/_planner/contracts.py
Bo-Vincent: Thanks, this is a valid gap.

`BoundTransferPlan` now requires the complete `target_placement` and verifies that its resource identity, revision, and weight generation match the plan. Its public constructor also revalidates complete, non-overlapping coverage for every selected target participant against the target placement fragments.

This closes both reproductions: an empty bound plan is rejected, and a reconstructed TP2 -> TP1 plan with one 32-byte overlap removed now fails with `bound target fragment is not fully covered`.

The validation remains participant-selective, so `plan_placement_transfer_to_local_target(...)` still validates only the target participant intentionally selected by that API, while requiring full coverage for that participant.

mooncake-reshard/python/mooncake/reshard/weight/_planner/validation.py
Bo-Vincent: Thanks, this is a valid lifetime/provenance issue.

The binding stage no longer collapses distinct logical target operations during alias handling, so target participant, lease, and owner provenance are retained rather than silently discarded.

Exact physical target-range overlap is now accepted only when both source and target declare the same complete alias group and the targets belong to the same attested runtime binding and lease scope. Equal addresses from different target runtime bindings or lease identities are rejected fail-closed as `conflicting target physical range` rather than being merged.

A regression test covers the reported PP-alias shape with shared instance, worker, device, and address but distinct target lease IDs and non-null owners.

mooncake-reshard/python/mooncake/reshard/weight/_planner/contracts.py
Bo-Vincent: Bound participant provenance now collects `RuntimeLeaseSnapshot` values in a dictionary keyed by `RuntimeFragmentId`, replacing the previous `if lease not in leases` list-membership path.

The implementation also rejects a repeated runtime fragment ID when its snapshot differs from the previously observed one, before the final deterministic sort.

A regression test uses 32 independently bound fragments and instruments snapshot equality. The previous implementation performed 1056 equality checks for that case; the new collection remains under a linear bound.

Aionw:
One overall maintainability concern: this PR adds about 10.4k lines, of which roughly 6.6k are tests.

There are, however, several opportunities to reduce the implementation and test surface:

1. **Remove the `CopyRange` compatibility path unless an executor explicitly requires it.** The planner itself emits `TransferRegion`; `CopyRange` is currently exercised mainly by the binding compatibility branch and tests. Standardizing on `TransferRegion` would remove the operation union, legacy byte-range coverage path, binding branch, and their associated tests.
2. **Do not accept derived metadata as independent constructor input.** In particular, `pipeline_routes` is deterministically derived from `operations`. Making it a computed property (and evaluating the same approach for participant operation indices) would reduce duplicate Logical/Bound validation and eliminate metadata-forgery states.
3. **Share the Logical/Bound validation skeletons and box primitives.** The two target-coverage validators are largely the same apart from how they project a placement fragment and validate attestations. Likewise, box containment/overlap/coverage helpers are duplicated across modules.
4. **Consider property-based tests for the geometry matrix.** Hypothesis is a good fit for rank/shape/partition generation, exact coverage, byte-equivalence against a small oracle, permutation invariance, and bind-time address expansion. This could replace a meaningful portion of the hand-written Cartesian test matrix while retaining targeted regression cases. Since this PR explicitly tests Python 3.9, this would need an appropriate test-only version constraint (Hypothesis 6.142 dropped Python 3.9 support).
5. **Consider splitting the change by layer**, for example logical contracts/planning first and runtime binding/attestation second.

My recommendation is therefore to simplify the data model and derived-state handling first, optionally use Hypothesis as a test-only dependency, and avoid adding a large runtime framework for the planner itself.

mooncake-reshard/python/mooncake/reshard/weight/_planner/api.py
```diff
@@ -0,0 +1,161 @@
...
+        source_manifest.fragments,
+        target_tensors,
+        target_fragments,
+        planning_limits=planning_limits,
+    )
+    return _logical_transfer_plan(
+        transfer,
+        source_tensors=source_tensors,
+        target_tensors=target_tensors,
+        source_placement=None,
```
Aionw, line 154: **[P1] Preserve the stored manifest identity when a plan can cross a process or trust boundary.** This is unlikely in the immediate in-process happy path because the operation is produced from an already validated `WeightManifest`. The failure becomes realistic when a coordinator caches, serializes, or sends a `LogicalTransferPlan` to another worker: a buggy or untrusted reconstruction can replace a `StoredFragment.object_key` (or object range) with another object having the same tensor geometry. This call records `source_placement=None` and the plan retains no source manifest/group snapshot or digest, so `_validate_logical_fragment_snapshots()` skips source membership validation and `bind_logical_transfer_plan()` accepts the substituted object. I reproduced this by changing the planned key to `another-tenant/private-object`; binding still succeeded. Please retain an immutable source-manifest identity and verify every stored fragment against it before binding. If stored plans are guaranteed to remain process-local and trusted instead, that restriction should be documented and enforced at the API boundary.

mooncake-reshard/python/mooncake/reshard/_typing.py
```diff
@@ -0,0 +1,5 @@
+"""Typing compatibility definitions for the supported Python runtimes."""
+
+from typing_extensions import TypeAlias
```
Aionw: **[P2] Declare `typing_extensions` as a direct runtime dependency.** A normal installation is unlikely to fail today because the current `aiohttp` dependency happens to install `typing_extensions` transitively. The failure is deterministic in a minimal/vendor installation that installs only declared direct requirements, or if `aiohttp` changes its dependency metadata: importing the public reshard package reaches this module and raises `ModuleNotFoundError`. This is especially relevant to the NPU/MUSA wheel variants, which rewrite the Python floor to 3.9. Please either declare `typing_extensions` in the wheel metadata, or import `TypeAlias` from the standard library where available and add an explicit Python 3.9 conditional dependency.

Aionw (commented):
Two follow-up findings from the latest head. Both are low-frequency in the current happy path, but the concrete boundary scenarios below are worth closing.

mooncake-reshard/python/mooncake/reshard/weight/_planner/ownership.py
```diff
@@ -0,0 +1,192 @@
...
+    supported. ``OwnershipAxis("dp")`` means different DP ranks can own
+    different logical parameters; selecting or expanding those parameters
+    requires framework-provided ownership semantics, not rank arithmetic in
+    the canonical planner.
+    """
+
+    unsupported = sorted(
+        tensor.tensor_id for tensor in tensors if has_dp_ownership(tensor)
+    )
+    if unsupported:
```
zxpdemonio: **[P1] The documented DP-ownership path is unconditionally disabled here.** The design document says `OwnershipAxis(kind="dp")` routes each tensor through its declared owner, and the earlier review response said that behavior was fixed. However, every public planning path calls this function for both source and target descriptors, and the current regression test now expects even an identity ownership placement to fail.

This also blocks a lossless compatibility adapter for the existing `TensorParallelism` contract, where DP can be an ownership/scope axis rather than a complete replica. The fragment inventory already identifies each tensor's source and target DP owner, so please either restore owner-based selection/routing, or explicitly narrow the PR/RFC compatibility claim and remove the unsupported manifest/documentation semantics.

mooncake-reshard/python/mooncake/reshard/weight/_planner/contracts.py
```diff
@@ -0,0 +1,1274 @@
...
+)
+
+
+RuntimeTensorOwner = tuple[tuple[str, int], ...]
+_SourceFragmentT = TypeVar("_SourceFragmentT", bound=GeometryFragment)
+_TargetFragmentT = TypeVar("_TargetFragmentT", bound=GeometryFragment)
+
+
+@dataclass(frozen=True)
+class RuntimeLeaseSnapshot:
```
Aionw: **[P2] Remove the orphaned runtime-binding subsystem from this logical-only PR.** The stable exports now expose only `LogicalTransferPlan` and the three address-free planning functions, and the new design document explicitly defers runtime binding to a later phase. However, this package still adds `RuntimeLeaseSnapshot`, `ExecutorTransferPlan`, `TransferPlan`, runtime attestation, bound-fragment handling, executor resolution, and physical validation, with no public binding entry point and almost no end-to-end behavioral coverage. That leaves a large private subsystem whose contracts can drift before the phase that is supposed to own it. Please either remove the runtime binding/attestation/executor branch from this PR, or restore a supported public entry point and its full tests if it is intentionally in scope.

mooncake-reshard/python/mooncake/reshard/weight/_planner/api.py
```diff
@@ -0,0 +1,164 @@
...
+        source_manifest=None,
+        target_placement=target_placement,
+        target_participant_ids=frozenset({target_participant_id}),
+    )
+    if len(result.target_executors) != 1:
+        raise ValueError("local target placement must describe exactly one executor")
+    return result
+
+
+def plan_stored_transfer_to_target_placement(
```
Aionw, line 121: **[P2] Add direct behavioral coverage for the public Store-backed planner.** `plan_stored_transfer_to_target_placement()` is newly exported, but the current tests only exercise `WeightManifest` construction/serialization; no test calls this planning path. The source-manifest snapshot issue fixed in this update could therefore regress while the suite remains green. Please add at least one successful Store-manifest-to-N-D-placement plan and negative reconstruction cases that replace a selected stored fragment or mismatch the manifest identity/digest.

Aionw (commented):
Two remaining scope and coverage issues on the current logical-only head.

mooncake-reshard/python/mooncake/reshard/weight/_planner/contracts.py
```diff
@@ -0,0 +1,732 @@
...
+                    "logical transfer plan exceeds max_total_lowered_segments"
+                )
+        if not all(
+            isinstance(executor, PlacementExecutorPlan)
+            for executor in (*self.source_executors, *self.target_executors)
+        ):
+            raise ValueError(
+                "logical transfer plan has invalid canonical executor metadata"
+            )
+        self.validate_source_manifest_snapshot()
```
Aionw, line 523: **[P2] Keep placement-source provenance self-consistent at the canonical plan boundary.** A future executor can and should revalidate runtime bindings, but there is no `LogicalTransferPlan` executor or TE lowering path in this PR today, and this public/reconstructible contract already validates Store sources and target placement snapshots. The placement-source side remains forgeable: I replaced one operation's source rank from `pp=0` to nonexistent `pp=99` while retaining its fragment ID, and reconstruction succeeded; `source_executors` still reported `pp=0` while `pipeline_routes` contained `source_pp=99`. Reconstruction also accepts empty `source_executors`, `source_tensors`, or `target_tensors`. Please validate every placement source fragment and source executor projection against `source_placement`, and validate the tensor catalogs against their provenance, or make these values derived `init=False` fields. A later executor can repeat this as defense in depth, but should not be the first component to discover an internally contradictory canonical plan.

Aionw (commented):
Two remaining logical-contract and scope issues on the current head.

mooncake-reshard/python/mooncake/reshard/contracts/adapter.py
```diff
@@ -0,0 +1,72 @@
...
+
+from .manifest import (
+    PlacementManifest,
+    ResourceKind,
+    RuntimeBindingManifest,
+    StoredResourceManifest,
+)
+
+
+class ResourceAdapter(ABC):
```
Aionw: **[P2] Defer this unused public adapter registry until a concrete consumer exists.** The PR is now explicitly scoped to logical model-weight planning, but it also exports `ResourceAdapter`, `ResourceAdapterRegistry`, and `StoredResourceManifest` as new public contracts. There are no registry instances, adapter implementations, callers, or tests in the repository; the API is therefore fixing an unvalidated shape for a future KV/runtime phase. Please remove this registry/protocol surface from this PR and introduce it with the first real second adapter, when its required methods and type boundaries can be derived from an end-to-end use case.

mooncake-reshard/python/mooncake/reshard/weight/_planner/contracts.py
Bo-Vincent: `LogicalTransferPlan` now revalidates placement-source provenance at its public construction boundary. Each source fragment must exactly match `source_placement`; source executor participant/rank/fragment projections, source tensor catalog, and target tensor catalog are also checked against their placement provenance.

Regression coverage now rejects the reported forged `pp=99` source fragment, empty `source_executors`, empty `source_tensors`, and empty `target_tensors`. Store-backed source snapshot validation remains unchanged and continues to reject substituted stored fragments.

mooncake-reshard/python/mooncake/reshard/contracts/adapter.py
Bo-Vincent: Removed the unused `ResourceAdapter` and `ResourceAdapterRegistry` public surface. The current logical-planning PR no longer defines a registry shape without a concrete consumer.

The resource-neutral structural contracts remain intentionally limited to `ResourceKind`, `ResourceManifest`, `PlacementManifest`, `RuntimeBindingManifest`, and `StoredResourceManifest`. `KV_CACHE` remains a reserved resource kind only; this PR does not add a KVCache adapter, registry, or implementation.

Aionw (approved):
LGTM

mooncake-reshard/python/mooncake/reshard/weight/_planner/contracts.py
```diff
@@ -0,0 +1,743 @@
...
+        """Fail closed if a stored source no longer matches its plan snapshot."""
+
+        if self.source_manifest is None:
+            if self.source_manifest_identity is not None:
+                raise ValueError("logical plan has unexpected source manifest identity")
+            return
+        source_manifest = validate_weight_manifest_snapshot(self.source_manifest)
+        if self.source_manifest_identity != source_manifest.manifest_identity:
+            raise ValueError("logical plan source manifest identity differs")
+        if self.source_tensors != source_manifest.tensors:
```
zxpdemonio: **[P1] Canonicalize the Store tensor catalog before comparing snapshots.** `WeightManifest` preserves the caller-provided tensor order, while `_logical_transfer_plan()` sorts `source_tensors` by `tensor_id` before this order-sensitive tuple comparison. A valid two-tensor manifest constructed as `(b, a)` against the equivalent canonical target `(a, b)` therefore reaches this line and fails with `ValueError: logical plan source tensor catalog differs`. This makes Store-backed resharding depend on collection/serialization order even though the catalog contents are identical. Please compare canonicalized catalogs (or normalize the manifest order at a clearly versioned boundary) and add a reversed-order multi-tensor Store planner test; the current direct test contains only one tensor and cannot catch this.

mooncake-reshard/python/mooncake/reshard/geometry.py
```diff
@@ -0,0 +1,131 @@
...
+
+def boxes_overlap(
+    boxes: Sequence[tuple[tuple[int, ...], tuple[int, ...]]],
+) -> bool:
+    """Return whether any two N-D boxes overlap."""
+
+    if len(boxes) < 2:
+        return False
+    ndim = len(boxes[0][0])
+    sweep_dim = max(
```
zxpdemonio, line 186: **[P1] Replace the active-list scan with a bounded-complexity box-intersection algorithm.** Choosing the sweep dimension by the number of distinct intervals does not bound the active set: a valid 2-D exact cover where each row is split at a different x-coordinate keeps O(N) mutually non-overlapping boxes active and makes the scan O(N^2). On the current implementation, 500/1,000/2,000/4,000 boxes took about 0.16/0.62/2.48/9.91 seconds for one `boxes_exactly_cover()` call. This validation is repeated by manifest construction, planner coverage, and plan reconstruction, while `PlanningLimits` permits up to one million regions, so a valid fragmented layout can consume unbounded CPU long before the region limit protects the caller. Please use an O(N log N) 1-D/2-D sweep with a coordinate-compressed segment tree and a bounded general N-D intersection algorithm (or fail closed above an explicit high-dimensional cutoff), and add the adversarial row-split layout as a regression case.

mooncake-reshard/python/mooncake/reshard/weight/_planner/contracts.py
Bo-Vincent: `WeightManifest` continues to preserve its exported tensor order and immutable identity. `LogicalTransferPlan` now canonicalizes the Store tensor catalog by `tensor_id` only when comparing the snapshot against the plan.

Added a two-tensor reversed-order Store source regression test. A logically equivalent `(b, a)` catalog now plans successfully against the canonical `(a, b)` target, while manifest identity and selected stored-fragment validation remain fail-closed.

mooncake-reshard/python/mooncake/reshard/geometry.py
Bo-Vincent: Coverage validation now uses an ordered interval scan for 1-D inputs and a coordinate-compressed sweep with range-max queries for 2-D inputs. Both avoid the unbounded active-list scan.

For 3-D and higher boxes, exact intersection is retained under an explicit pairwise-comparison budget; exceeding that budget fails closed.

Added a 4096-row adversarial 2-D exact-cover regression, a high-dimensional budget regression, and a Hypothesis oracle test against brute-force overlap detection.

mooncake-reshard/tests/model_weight_nd_planner/__init__.py
zxpdemonio, line 1: No Chinese comment

zxpdemonio (approved):
LGTM.

stmatengss (approved):

PR #3445, opened 2026-08-14
[CI/Build] Preserve CTest failure diagnostics

Aionw:
## Description

PR CTest failures currently surface only through the combined GitHub Actions log. The check annotation reports the CTest exit code, while the failing assertion or sanitizer output can be buried among parallel test output, and no test diagnostics are preserved as artifacts.

This change improves the TENT CTest entry point in the daily PR workflow:

- emit JUnit XML from the TENT CTest run;
- add a failure-only job summary with failed CTest targets, normalized GoogleTest cases, and bounded assertion/sanitizer/crash context;
- upload JUnit, `LastTest.log`, and `LastTestsFailed.log` as per-matrix artifacts for 14 days;
- keep summary generation and artifact upload together in one reusable local composite action;
- keep raw diagnostic upload independent from summary rendering through `if: ${{ always() }}`.

The summary parser recognizes GoogleTest failures, native assertions, ASan/LSan/TSan/UBSan output, runtime errors, and common crash messages. The existing test execution and pass/fail behavior are unchanged.

## Module

- [x] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [x] Other — CI diagnostics

## How Has This Been Tested?

**Test commands:**

```bash
.github/actions/ctest-diagnostics/test.sh
bash -n \
  .github/actions/ctest-diagnostics/summarize.sh \
  .github/actions/ctest-diagnostics/test.sh
pre-commit run --files \
  .github/actions/ctest-diagnostics/action.yml \
  .github/actions/ctest-diagnostics/summarize.sh \
  .github/actions/ctest-diagnostics/test.sh \
  .github/actions/ctest-diagnostics/testdata/last-tests-failed.txt \
  .github/actions/ctest-diagnostics/testdata/last-test.txt \
  .github/actions/ctest-diagnostics/testdata/expected-summary.md \
  .github/workflows/ci.yml
git diff --check
```

The fixture/golden test covers:

- duplicate timed/untimed GoogleTest failure lines and the aggregate failure line;
- a native `Assertion 'ready' failed.` message;
- the standard `WARNING: ThreadSanitizer` form;
- exact rendered Markdown for failed targets, normalized cases, and failure context.

The Ubuntu 22.04 TENT job installs CMake/CTest `3.22.1-1ubuntu1.22.04.2`, which satisfies the CTest 3.21+ requirement for `--output-junit` ([TENT build log](https://github.com/kvcache-ai/Mooncake/actions/runs/31956333465/job/95187517732)).

**Test results:**

- [ ] Unit tests pass — not run; this changes only workflow diagnostics
- [ ] Integration tests pass — not applicable
- [x] Fixture/golden test passes
- [x] Focused pre-commit hooks pass
- [x] Shell syntax and diff checks pass

## Checklist

- [x] I have formatted the touched YAML/shell files through the configured pre-commit hooks
- [ ] I have run `pre-commit run --all-files` — not required; PR-scoped hooks passed
- [ ] I have updated the documentation — not applicable
- [x] I have added tests to prove my changes are effective — local fixture/golden test included
- [ ] For changes >500 LOC: I have filed an RFC issue — not applicable

## AI Assistance Disclosure

- [ ] No AI tools were used
- [x] AI tools were used

OpenAI Codex inspected recent GitHub Actions failures, implemented the reusable diagnostics action and workflow changes, addressed review feedback, and ran the focused validation. The human submitter reviewed the changes before they were pushed and should be able to defend them end to end.


stmatengss (commented):
Verdict: Approve. Well-designed.

The composite action follows GH Actions best practices: proper input declarations with sensible defaults, if-no-files-found: warn preventing job failure if a diagnostic file is missing. The if: always() && steps.ctest.outcome == 'failure' guard is the correct pattern. summarize.sh is robust: set -euo pipefail, GITHUB_STEP_SUMMARY validated with :?, all grep calls guarded with || true, and the failure-context regex matches gtest failures plus ASan/LSan/TSan/UBSan, runtime errors, segfaults, and terminate called.

Low: --output-junit requires CTest 3.21+. Worth verifying the CI image's CMake version.

ykwd (approved):

Icedcoco:
Thanks for putting this together. I walked through the failure path and had a few questions before this lands:

1. Could we broaden the failure-context matching in `summarize.sh`? I tried a minimal CTest failure containing a native assertion such as `Assertion 'ready' failed.`. The message was present in `LastTest.log`, but the generated summary only showed the failed CTest target. The current pattern also appears to miss the usual `WARNING: ThreadSanitizer` form. Is the intended scope limited to GoogleTest and ASan, or should these cases be included as well?

2. Could we add a small fixture-based test for the summary script and provide an end-to-end run of the final revision? The current `build` and `tent-ci` checks were skipped, while the previous run was cancelled. The test commands in the PR description also still reference the old `scripts/ci/summarize_ctest_failure.sh` path. Since the composite action only runs after a CTest failure, a normal green CI run would not exercise its main behavior.

3. Would it make sense to add `if: ${{ always() }}` to the artifact upload step inside the composite action? As written, if summary generation fails, the upload step will inherit the default success condition and be skipped. Since preserving the raw logs is more important than rendering the summary, it seems useful to keep those two failure paths independent.

4. Could we normalize the `Failed test cases` output before displaying it? A normal GoogleTest failure produces a timed failure line, an aggregate line such as `1 test, listed below:`, and an untimed copy of the same test name. The current `grep | sort -u` includes all three, so one failed case is presented as three entries. Would extracting the final failed-test block, or filtering the aggregate line and removing the timing suffix before deduplication, give a cleaner result?

The overall direction makes sense; these are mainly about making sure the diagnostic path covers the failures it claims to handle and remains reliable when its own summary step has trouble.

Aionw:
@Icedcoco Fixed in https://github.com/kvcache-ai/Mooncake/pull/3445/commits/1ac7a7e9a0de9a72abc8d05856c1d5f9f0db6813

Icedcoco (approved):
LGTM from my side. The latest revision addresses the previous review points: native assertions and TSan output are covered, failed GoogleTest cases are normalized, and artifact upload is now independent of summary rendering through `if: ${{ always() }}`. The fixture/golden test also passes locally.

I only have two small, non-blocking follow-ups:

1. The failure-context matcher still does not match standalone `unknown file: Failure` or `Subprocess aborted` lines. These are not a concern for the current scope, but adding them would make the “GoogleTest failures” and “common crash messages” coverage more complete.

2. `.github/actions/ctest-diagnostics/test.sh` is currently a manual fixture test and is not invoked by CI. It would be useful to wire it into an existing lightweight check in a follow-up so parser regressions are caught automatically.

Neither point changes my review decision. Assuming the remaining CI jobs complete successfully, I’m happy to approve this PR.

ykwd:
There is conflicts that must be resolved

PR #3326, opened 2026-08-07
[Store] Add bounded standby snapshot capture

Icedcoco:
> [!IMPORTANT]
> This PR implements N02 and depends on both #3178 (N01) and #3201 (S02). It must not be merged before those two PRs. After both dependencies are merged, this branch will be rebased onto the updated `main`.

## Description

Part of #3167.

This PR adds the bounded, complete-batch capture foundation for standby-generated batch OpLog snapshots.

The implementation:

- extracts `StandbyMetadataStore` into a standalone concrete type while retaining the existing tenant-to-object map and single-mutex design;
- removes the per-object `last_sequence_id`, with snapshot consistency represented by one global complete-batch cursor instead;
- adds an insert-only restore path that rejects duplicate `(tenant_id, key)` entries;
- adds a move-only traversal cursor and bounded object copying through `CopyNextSnapshotChunk()`;
- records the last fully applied durable prefix, including its batch ID, sequence ID, and producer view;
- processes capture requests only after `OpLogBatchStandbyReader::PollOnce()` completes;
- freezes OpLog apply while capturing the complete durable-prefix cursor, producer view, segment registry, and object chunks;
- releases the metadata-store mutex between chunks without resuming apply, preserving iterator validity without allocating a full object vector;
- coordinates promotion and shutdown with active captures so they wait only for the current bounded copy before cancelling the capture and resuming final catch-up;
- rejects inconsistent or incomplete capture boundaries without rolling back the reader cursor.

### Dependency order

This PR consumes:

- the batch snapshot protocol introduced by #3178 (N01);
- the durable-prefix producer view introduced by #3201 (S02).

The required merge order is therefore:

1. #3178 and #3201
2. this PR (N02)

### Scope

This PR only provides the capture lifecycle and bounded traversal primitives.

It intentionally does **not** implement artifact encoding, object-store upload, snapshot publication, scheduling, restore from the new format, garbage collection, or OpLog pruning. Those remain in subsequent N-series PRs tracked by #3167.

The legacy promotion snapshot path remains unchanged.

## Module

- [ ] Transfer Engine (`mooncake-transfer-engine`)
- [x] Mooncake Store (`mooncake-store`)
- [ ] Mooncake EP (`mooncake-ep`)
- [ ] Mooncake PG (`mooncake-pg`)
- [ ] Integration (`mooncake-integration`)
- [ ] P2P Store (`mooncake-p2p-store`)
- [ ] Python Wheel (`mooncake-wheel`)
[...]

PR #3447, opened 2026-08-14
[Store] Add chunked standby snapshot artifact writer

Icedcoco:
## Description

This PR implements N03 of the standby snapshot roadmap, building on the bounded
complete-batch capture introduced in #3326.

It materializes a frozen standby state into the
`standby-oplog-materialized/v1` artifact set:

- `segments.bin`
- chunked `objects/<index>.bin` files
- `manifest.json`
- `descriptor.json`

The writer processes one object chunk at a time to bound peak memory. It copies
a chunk from the standby store, releases the store lock, encodes and uploads the
chunk, then releases its buffers before continuing. Capture ends immediately
after all segments and object chunks have been materialized; manifest and
descriptor validation no longer hold or reference standby state.

This PR also:

- adds struct_pack codecs for segments and object chunks;
- validates chunk index and object count during decoding;
- records and verifies stored size and CRC32C for every artifact;
- adds object inspection support for local and S3 backends;
- uses backend CRC32C when available and full readback otherwise;
- rejects pre-existing candidate prefixes;
- performs best-effort candidate cleanup after upload or verification failure;
- keeps batch-oplog-specific code under `ha/snapshot/batch_oplog/`.

This PR intentionally does not acquire the maintenance lease, publish etcd
pointers, schedule snapshots, implement restore, or add compression. Those
remain in subsequent roadmap PRs.

## Module

## Type of Change

- [x] New feature

## How Has This Been Tested?

Added focused unit coverage for:

- segments and object-chunk round trips;
- truncated artifacts and chunk index/count mismatches;
- empty and multi-chunk snapshots;
- pre-existing candidate rejection;
- upload failure cleanup;
- backend checksum mismatch;
- checksum-unavailable readback and readback failure;
- local object size inspection.

Manual S3-compatible smoke testing also covered:

- compilation against AWS SDK for C++ `1.11.836`;
- 1 KiB single-PUT upload;
- 101 MiB multipart upload;
- checksum-enabled HEAD and stored-size inspection;
- range readback with byte-for-byte and local CRC32C verification;
- prefix listing, deletion, and unfinished multipart cleanup.

The complete smoke test passed with checksum calculation and validation set to
`when_required`. Moto did not expose server-side CRC32C, so the fallback
readback path was exercised; the positive server-CRC32C path still requires
validation against Amazon S3 or another backend that returns
`x-amz-checksum-crc32c`.

## Checklist

- [x] Documentation is not required for this implementation-only slice
- [x] RFC #3167 covers the architecture for this change

## AI Assistance Disclosure

- [x] AI tools were used

Codex assisted with implementation analysis, test development, isolated S3
smoke validation, and PR description drafting. The submitter is responsible
for reviewing and understanding every changed line.

### DIFF
--- a/mooncake-store/src/ha/snapshot/batch_oplog/codec.cpp
+++ b/mooncake-store/src/ha/snapshot/batch_oplog/codec.cpp
@@ -1,6 +1,7 @@
 #include "ha/snapshot/batch_oplog/codec.h"
 
 #include <limits>
+#include <utility>
 
 #include <ylt/struct_pack.hpp>
 
@@ -21,8 +22,7 @@ tl::expected<T, std::string> Decode(const std::vector<uint8_t>& encoded,
 
 template <typename T>
 std::vector<uint8_t> Encode(const T& value) {
-    auto encoded = struct_pack::serialize(value);
-    return {encoded.begin(), encoded.end()};
+    return struct_pack::serialize<std::vector<uint8_t>>(value);
 }
 
 }  // namespace
@@ -39,8 +39,9 @@ DecodeBatchOpLogSnapshotSegments(const std::vector<uint8_t>& encoded) {
 }
 
 std::vector<uint8_t> EncodeBatchOpLogSnapshotObjectChunk(
-    uint64_t chunk_index, const std::vector<StandbyObjectEntry>& objects) {
-    return Encode(BatchOpLogSnapshotObjectChunk{chunk_index, objects});
+    uint64_t chunk_index, std::vector<StandbyObjectEntry> objects) {
+    return Encode(
+        BatchOpLogSnapshotObjectChunk{chunk_index, std::move(objects)});
 }
 
 tl::expected<BatchOpLogSnapshotObjectChunk, std::string>
--- a/mooncake-store/src/ha/snapshot/batch_oplog/codec.cpp
+++ b/mooncake-store/src/ha/snapshot/batch_oplog/codec.cpp
@@ -1,0 +1,65 @@
+#include "ha/snapshot/batch_oplog/codec.h"
+
+#include <limits>
+
+#include <ylt/struct_pack.hpp>
+
+namespace mooncake {
+
+namespace {
+
+template <typename T>
+tl::expected<T, std::string> Decode(const std::vector<uint8_t>& encoded,
+                                    const char* artifact) {
+    T decoded;
+    if (struct_pack::deserialize_to(decoded, encoded) !=
+        struct_pack::errc::ok) {
+        return tl::make_unexpected(std::string("Invalid ") + artifact);
+    }
+    return decoded;
+}
+
+template <typename T>
+std::vector<uint8_t> Encode(const T& value) {
+    auto encoded = struct_pack::serialize(value);
+    return {encoded.begin(), encoded.end()};
+}
+
+}  // namespace
+
+std::vector<uint8_t> EncodeBatchOpLogSnapshotSegments(
+    const std::vector<StandbySegmentInfo>& segments) {
+    return Encode(segments);
+}
+
+tl::expected<std::vector<StandbySegmentInfo>, std::string>
+DecodeBatchOpLogSnapshotSegments(const std::vector<uint8_t>& encoded) {
+    return Decode<std::vector<StandbySegmentInfo>>(encoded,
+                                                   "snapshot segments");
+}
+
+std::vector<uint8_t> EncodeBatchOpLogSnapshotObjectChunk(
+    uint64_t chunk_index, const std::vector<StandbyObjectEntry>& objects) {
+    return Encode(BatchOpLogSnapshotObjectChunk{chunk_index, objects});
+}
+
+tl::expected<BatchOpLogSnapshotObjectChunk, std::string>
+DecodeBatchOpLogSnapshotObjectChunk(const std::vector<uint8_t>& encoded,
+                                    uint64_t expected_chunk_index,
+                                    uint64_t expected_object_count) {
+    auto decoded =
+        Decode<BatchOpLogSnapshotObjectChunk>(encoded, "snapshot object chunk");
+    if (!decoded) {
+        return decoded;
+    }
+    if (decoded->chunk_index != expected_chunk_index) {
+        return tl::make_unexpected("Snapshot object chunk index mismatch");
+    }
+    if (expected_object_count > std::numeric_limits<size_t>::max() ||
+        decoded->objects.size() != expected_object_count) {
+        return tl::make_unexpected("Snapshot object chunk count mismatch");
+    }
+    return decoded;
+}
+
+}  // namespace mooncake
--- a/mooncake-store/src/ha/snapshot/batch_oplog/writer.cpp
+++ b/mooncake-store/src/ha/snapshot/batch_oplog/writer.cpp
@@ -1,6 +1,5 @@
 #include "ha/snapshot/batch_oplog/writer.h"
 
-#include <string_view>
 #include <utility>
 #include <vector>
 
@@ -15,10 +14,6 @@
 namespace mooncake {
 
 namespace {
-
-std::vector<uint8_t> ToBytes(std::string_view value) {
-    return {value.begin(), value.end()};
-}
 
 tl::expected<void, std::string> VerifyObject(SnapshotObjectStore& object_store,
                                              const std::string& key,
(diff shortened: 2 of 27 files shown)


mooncake-store/src/ha/snapshot/batch_oplog/codec.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+namespace {
+
+template <typename T>
+tl::expected<T, std::string> Decode(const std::vector<uint8_t>& encoded,
+                                    const char* artifact) {
+    T decoded;
+    if (struct_pack::deserialize_to(decoded, encoded) !=
+        struct_pack::errc::ok) {
+        return tl::make_unexpected(std::string("Invalid ") + artifact);
```
Aionw, line 18: Can we add truncation here? Suppose we decode a 50G artifact, this will blow up.

mooncake-store/src/ha/snapshot/batch_oplog/writer.cpp
```diff
@@ -0,0 +1,220 @@
...
+#include "ha/snapshot/batch_oplog/codec.h"
+#include "ha/snapshot/batch_oplog/metadata.h"
+#include "ha/snapshot/object/snapshot_object_store.h"
+#include "hot_standby_service.h"
+
+namespace mooncake {
+
+namespace {
+
+std::vector<uint8_t> ToBytes(std::string_view value) {
```
Aionw: Here you can just use `value.data()` and `value.size()` for the bytes. No need to wrap it in an extra `std::vector<uint8_t>` layer, as that adds an extra copy and hurts performance.

mooncake-store/src/utils/s3_helper.cpp
```diff
@@ -191,6 +192,42 @@ tl::expected<void, std::string> S3Helper::UploadString(
...
+    }
+    const auto content_length = outcome.GetResult().GetContentLength();
+    if (content_length < 0) {
+        return tl::make_unexpected(
+            fmt::format("Invalid content length: {}", content_length));
+    }
+    stored_size = static_cast<uint64_t>(content_length);
+    crc32c.reset();
+
+    const std::string encoded = outcome.GetResult().GetChecksumCRC32C();
```
Aionw, line 237: `ChecksumCRC32C` is not necessarily a full-object checksum for multipart uploads; it may be `COMPOSITE`. In that case, comparing it with `Crc32cValue()` over the complete artifact will reject a valid upload. Please check that `GetChecksumType()` is `FULL_OBJECT` before exposing this checksum. Otherwise, leave `crc32c` unset so verification falls back to a complete readback. The upload path should also explicitly request CRC32C with the full-object checksum type if we want this fast path to work reliably.

mooncake-store/src/ha/snapshot/batch_oplog/codec.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+tl::expected<std::vector<StandbySegmentInfo>, std::string>
+DecodeBatchOpLogSnapshotSegments(const std::vector<uint8_t>& encoded) {
+    return Decode<std::vector<StandbySegmentInfo>>(encoded,
+                                                   "snapshot segments");
+}
+
+std::vector<uint8_t> EncodeBatchOpLogSnapshotObjectChunk(
+    uint64_t chunk_index, const std::vector<StandbyObjectEntry>& objects) {
+    return Encode(BatchOpLogSnapshotObjectChunk{chunk_index, objects});
```
Aionw: This copies the entire object chunk because `objects` is passed by const reference and then stored by value in `BatchOpLogSnapshotObjectChunk`. During serialization, the original vector, this copy, and the encoded output can all coexist, significantly increasing peak memory for large chunks. Could the encoder take ownership of the vector and move it into the wrapper, or serialize without creating a second object vector? Also, `Encode()` currently copies the `struct_pack` output again when converting it to `std::vector<uint8_t>`.

mooncake-store/src/ha/snapshot/batch_oplog/codec.cpp
Icedcoco: Good point. Snapshot artifacts will be bounded by `snapshot_chunk_object_count`; N08 wires the production configuration (default 1,000,000 objects per chunk), so normal writer output is split instead of producing one unbounded object artifact. This PR is writer-only, so I did not add an arbitrary codec-local byte cap here. In the N05 restore path, we will also consider size admission before download and truncation/size validation after download, before decode, so an oversized or corrupted artifact can be rejected at the correct trust boundary.

mooncake-store/src/ha/snapshot/batch_oplog/writer.cpp
Icedcoco: Fixed in `f2c8c129`. Manifest and descriptor CRC32C are now computed directly from `std::string::data()`/`size()`, and both artifacts use the existing `UploadString()` path. The temporary full-size `std::vector<uint8_t>` copies were removed.

mooncake-store/src/utils/s3_helper.cpp
Icedcoco: Fixed in `f2c8c129`. Single-part uploads submit a full-object CRC32C. Multipart uploads explicitly request `CRC32C` with `ChecksumType::FULL_OBJECT` and provide the full-object checksum and object size on completion. `InspectObject()` now exposes CRC32C only when `HeadObject` reports `FULL_OBJECT`; `COMPOSITE` or missing checksum types leave it unset and the writer falls back to complete readback.

mooncake-store/src/ha/snapshot/batch_oplog/codec.cpp
Icedcoco: Fixed in `f2c8c129`. The object-chunk encoder now takes ownership of the object vector, and the writer moves the chunk into it after saving `object_count`. `struct_pack` also serializes directly into `std::vector<uint8_t>`, removing both extra full-size copies.

Icedcoco:
Additional compatibility fix in `f2c8c129`: legacy snapshot bootstrap now assigns the decoded `hard_pinned` value to `StandbyObjectMetadata` instead of discarding it. The tests cover both legacy provider decoding and legacy snapshot bootstrap through standby promotion/export, preserving `hard_pinned=true`.

Validation passed: pre-commit on all touched files plus `batch_oplog_snapshot_codec_test`, `batch_oplog_snapshot_writer_test`, `hot_standby_snapshot_bootstrap_test`, and `catalog_backed_snapshot_provider_test`. The other review items are addressed in their inline threads.

mooncake-store/src/utils/s3_helper.cpp
```diff
@@ -444,6 +502,9 @@ tl::expected<void, std::string> S3Helper::UploadBufferMultipart(
     Aws::S3::Model::CreateMultipartUploadRequest create_request;
     create_request.SetBucket(bucket_.c_str());
     create_request.SetKey(key.c_str());
+    create_request.SetChecksumAlgorithm(
```
Aionw, line 505: [P1] Keep multipart checksum algorithms consistent

CreateMultipartUpload pins this multipart upload to CRC32C, but each UploadPartRequest leaves its checksum algorithm unset. With the default requestChecksumCalculation=WHEN_SUPPORTED, the AWS SDK for C++ automatically selects CRC64NVME, while S3 requires every part algorithm to match the algorithm selected by CreateMultipartUpload. As a result, uploads at or above the 100 MiB multipart threshold can fail under the default configuration. The smoke test uses when_required, which suppresses this path. Please explicitly select CRC32C, or provide ChecksumCRC32C, for every part and cover the default checksum configuration in the test.

Icedcoco: Fixed in `70334f5f`. Each multipart part now computes CRC32C once before the retry loop and sets it with `UploadPartRequest::SetChecksumCRC32C()` on every attempt. In AWS SDK for C++ 1.11.836, this setter also selects `ChecksumAlgorithm::CRC32C`, so the part requests stay consistent with `CreateMultipartUpload` under both the default `WHEN_SUPPORTED` mode and `WHEN_REQUIRED`.

I also rechecked the default-path mechanism against the 1.11.836 generated request/interceptor sources: an unset part algorithm otherwise resolves to `crc64nvme` under `WHEN_SUPPORTED`. Pre-commit and the four focused snapshot tests pass. I attempted to rebuild the isolated default-mode S3 smoke, but the required `aws-crt-cpp` submodule fetch failed with an empty reply and then timed out; the temporary checkout was removed, so I am not claiming an AWS-backed local smoke result for this follow-up.

Aionw (approved):
LGTM

ykwd (approved):

PR #3449, opened 2026-08-14
[EP] Add NCCL Device API backend for ElasticBuffer

akhillanger:
## Description

This PR adds an opt-in NCCL Device API backend for Mooncake EP's `ElasticBuffer` dispatch and combine path.

`ElasticBuffer` keeps `transport="auto"` as its Python default. In a build compiled with NCCL Device API support, all ranks use a group-wide consensus to select NCCL for a supported topology; otherwise they select the existing IPC + IBGDA backend. Users can also request `transport="nccl"` or `transport="ibgda"` explicitly. `MOONCAKE_EP_TRANSPORT=ibgda` provides a rollout override, and `buffer.transport` reports the backend selected by the group.

The NCCL backend:

- Keeps Python responsible for tensors while the native `_ep` interface receives raw device pointers and CUDA stream handles; native EP headers and sources do not depend on Torch or ATen.
- Compiles the elastic kernels against `NcclOps` or `IbgdaOps` traits, so the device hot path has no runtime transport branch.
- Uses NCCL LSA symmetric-memory operations within an LSA team and rail GIN operations between LSA teams.
- Provides scale-up, scale-out, and world team mappings, indexed signals, barriers, and native NCCL put/get operations through the Mooncake Transfer Engine Device API.
- Creates a dedicated NCCL communicator and symmetric allocation for each NCCL `ElasticBuffer`.
- Preserves the existing IPC + IBGDA implementation and workspace layout when that backend is selected.

NCCL support is disabled by default at build time. Enable it with `-DUSE_NCCL_DEVICE=ON` and provide NCCL through `NCCL_ROOT` or the normal CMake search path. The resulting `_ep` extension links directly to `libnccl`.

## Supported Configuration

- NCCL 2.30.4 or newer.
- NCCL Device API headers must exactly match the runtime `libnccl`; rebuild the extension after changing NCCL versions.
- CUDA EP builds on SM100-family GPUs.
- Precompiled BF16 and FP8 kernels for hidden size 4096, 256 experts, top-k 8, at most 128 tokens per rank, and 24 communication SMs.
- Fixed communicator membership for the lifetime of an NCCL `ElasticBuffer`; membership changes require creating a new buffer.
- Coordinated `destroy()` by all ranks before tearing down the distributed process group.

The precompiled NCCL topology matrix is:

- One contiguous LSA team containing 2 or 8 GPUs (`1x2`, `1x8`).
- Two contiguous LSA teams containing 4 or 8 GPUs each (`2x4`, `2x8`), using hybrid mode and rail GIN for cross-team communication.
- Four contiguous LSA teams containing 4 GPUs each (`4x4`), using hybrid mode and rail GIN for cross-team communication.

Other topology shapes are not selected by `transport="auto"` in this implementation. Explicit NCCL construction also validates the native topology and fails with an error when it is unsupported.

## Module

- [ ] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] Refactor
- [ ] Breaking change
- [x] Documentation update
- [x] Performance improvement
- [ ] Other

## How Has This Been Tested?

The implementation was built with CUDA 13.1 on GB300 systems against matching NCCL 2.30.4, 2.30.7, and 2.31.2 header/runtime pairs.

Representative build and test commands:

```bash
cmake -S . -B build -DWITH_EP=ON -DUSE_CUDA=ON \
  -DUSE_NCCL_DEVICE=ON -DNCCL_ROOT=/path/to/nccl
cmake --build build --target device_backend_api_compatibility_test _ep

torchrun --nproc-per-node=2 mooncake-ep/tests/test_elastic_buffer.py \
  --transport nccl --route alltoall --num-tokens 32 --max-tokens 128 \
  --hidden 4096 --num-experts 256 --num-topk 8 --num-sms 24

torchrun --nproc-per-node=2 mooncake-ep/benchmarks/elastic_buffer_perf.py \
  --transport nccl --route alltoall --num-tokens 128 --max-tokens 128 \
  --hidden 4096 --num-experts 256 --num-topk 8 --num-sms 24 \
  --warmup 20 --iters 200 --reuse-handle --no-check-correctness
```

### Correctness

Validation covered:

- CPU-synchronized, asynchronous, and cached-handle dispatch/combine.
- Local, cross-LSA, and all-to-all routes.
- Expanded expert layout.
- Group-wide automatic transport selection, explicit transport choices, builds without NCCL support, unsupported-topology fallback, and the IBGDA rollout override.
- Existing IPC/P2P behavior with `transport="ibgda"`.
- The NCCL and IBGDA Device API compile-compatibility target.

| Topology | NCCL routes | Result |
|---|---|---|
| 1 node, 2 GPUs | local, all-to-all | PASS |
| 2 nodes, 4 GPUs/node | local, cross-LSA, all-to-all | PASS |
| 4 nodes, 4 GPUs/node | local, cross-LSA, all-to-all | PASS |

| NCCL version | Native build | 2-GPU dispatch/combine |
|---|---|---|
| 2.30.4 | PASS | PASS |
| 2.30.7 | PASS | PASS |
| 2.31.2 | PASS | PASS |

### Performance

Configuration: BF16, 128 tokens per rank, hidden size 4096, top-k 8, 256 experts, 24 communication SMs, 20 warm-up iterations, 200 measured iterations, and five alternating-order repetitions. The tables report medians.

#### Mooncake NCCL versus upstream DeepEP v2

| Topology | Implementation | Dispatch critical (ms) | Combine critical (ms) | E2E critical (ms) | Wall/iter (ms) |
|---|---|---:|---:|---:|---:|
| 1 node, 2 GPUs | Mooncake NCCL | 0.067 | 0.065 | 0.130 | 0.259 |
| 1 node, 2 GPUs | DeepEP v2 NCCL GIN | 0.082 | 0.075 | 0.156 | 0.292 |
| 2 nodes, 4 GPUs/node | Mooncake NCCL rail | 0.119 | 0.104 | 0.220 | 0.335 |
| 2 nodes, 4 GPUs/node | DeepEP v2 NCCL GIN | 0.120 | 0.104 | 0.219 | 0.336 |
| 4 nodes, 4 GPUs/node | Mooncake NCCL rail | 0.211 | 0.159 | 0.337 | 0.447 |
| 4 nodes, 4 GPUs/node | DeepEP v2 NCCL GIN | 0.216 | 0.157 | 0.334 | 0.446 |

#### Mooncake NCCL versus the existing single-node backend

GPU-memory registration was unavailable on the test nodes, so the existing backend used its CUDA IPC/P2P path.

| Actual data path | Dispatch critical (ms) | Combine critical (ms) | E2E critical (ms) | Wall/iter (ms) |
|---|---:|---:|---:|---:|
| IPC/P2P | 0.061 | 0.060 | 0.121 | 0.250 |
| NCCL Device API | 0.061 | 0.062 | 0.122 | 0.258 |

A valid cross-node IBGDA measurement is not included because legacy `ibv_reg_mr` GPU registration failed without `nvidia_peermem` on the test nodes.

Additional validation included Python syntax compilation, shell syntax checks, CUDA profiler inspection, formatting checks, relevant non-rewriting pre-commit hooks, and `git diff --check`.

## Known Limitations

- NCCL is an opt-in build feature and the NCCL-enabled `_ep` extension has a direct runtime dependency on a matching `libnccl`.
- Only the topology and workload specializations listed above have precompiled NCCL kernels.
- `update_ep_member()` is unsupported for an NCCL buffer because NCCL communicator membership is fixed.
- Destruction is collective; applications must call `destroy()` on every rank before process-group teardown.
- Automatic fallback occurs during group transport selection. A failure after native NCCL initialization starts is reported to the caller rather than switching a live buffer to IBGDA.

## Checklist

- [ ] I have performed a self-review of every changed line
- [x] I have formatted the changed code using the repository formatting tools
- [x] I have run the applicable non-rewriting pre-commit hooks on touched files
- [x] I have updated the documentation where applicable
- [x] I have added tests to prove the changes are effective

The human submitter must review and understand every changed line before marking the PR ready.

## AI Assistance Disclosure

- [ ] No AI tools were used
- [x] AI tools were used as specified below

This implementation and PR description were generated with help of AI (OpenAI Codex). The human submitter is responsible for reviewing and defending the change end-to-end.


UNIDY2002:
@akhillanger Thank you for your valuable contribution!

One thing to note is that we recently refactored Mooncake EP, and its C++ implementation no longer depends on Torch APIs (it now only accepts raw pointers). Would you mind rebasing your changes onto our latest `main` branch?

Sorry for the potential inconvenience. Since the ElasticBuffer APIs have not yet stabilized, if you encounter any APIs that do not fit your use case, please feel free to reach out to us. You are also welcome to modify them directly, as long as the changes do not introduce native Torch dependencies.

akhillanger:
@UNIDY2002 Thanks for the heads-up. I rebased this PR onto the latest `main`, including the framework-neutral EP refactor, and adapted the NCCL integration to the current raw-pointer native API.

The Python wrapper still owns and allocates tensors, but it now passes raw device pointers and CUDA stream handles across the native boundary. The EP C++/CUDA headers and sources add no `torch::`, ATen, or `torch/extension` references, and the rebuilt `_ep` shared object has no Torch/ATen dynamic-library dependency. I also retained the new native CMake `_ep` target rather than restoring the removed `setup.py`/Torch extension build.

Validation after the rebase:

- native `_ep` CUDA/NCCL build: passed
- NCCL/IBGDA Device API compatibility target: passed
- 2-GPU NCCL dispatch/combine correctness smoke: passed
- 2-GPU explicit IBGDA/P2P regression smoke: passed
- Sphinx HTML, changed-lines clang-format, Ruff, codespell, and non-rewriting pre-commit hooks: passed

The PR description has also been updated to reflect the current native API and build commands. While validating on GB300, I added an optional native-CMake `MOONCAKE_EP_CUDA_ARCH` setting for family-specific NVCC targets such as `sm_100f`.

This response was generated with help of AI.

UNIDY2002:
@akhillanger Thank you for your work. BTW, a known issue of our native Torch decoupling refactor may be a slightly increased wall clock time usage. But good news is, in real-world deployment, such Python-layer overhead will disappear during CUDA graph replays. So we do not need to worry too much about the wall clock increase.

I will start a review soon.

mooncake-pg/torch/BuildPgExt.cmake
```diff
@@ -58,13 +62,30 @@ else()
...
+endif()
+
 # ---------------------------------------------------------------------------
 # 2. Build the PG Python extension.
 # ---------------------------------------------------------------------------
 if("${EP_TORCH_VERSIONS}" STREQUAL "")
   message(STATUS "[PG] Building with currently-installed PyTorch")
   execute_process(
-    COMMAND ${Python3_EXECUTABLE} setup.py build_ext --build-lib .
+    COMMAND ${Python3_EXECUTABLE} setup.py build_ext --build-lib . --force
```
UNIDY2002: Do we really need the --force here?

UNIDY2002: Do we really need the `--force` here?

UNIDY2002 (commented):
During validation we tested several `nvidia-nccl-cu13` versions against this PR's NCCL Device API backend.
- 2.30.4: build fails — `NCCL_GIN_RESOURCE_SHARING_THREAD` is referenced in `nccl_device.cuh` but absent from 2.30.4 headers. The documented `>= 2.30.4` requirement does not hold.
- 2.31.2: build fails — `sizeof(ncclDevComm_t)` is `248` in 2.31.2, triggering the `static_assert` in `nccl_device_transport.cpp:500` (`248 == 240`).
- 2.30.7: builds cleanly. Runtime not validated.
The hardcoded `kNativeCommBytes = 240` makes the working version range quite narrow. Worth considering how to handle NCCL version compatibility here.

mooncake-transfer-engine/include/transport/device/nccl_device_transport.h
```diff
@@ -91,15 +112,25 @@ class NcclBufferRegistration {
 // communicators and windows remain behind opaque pointers.
 class NcclDeviceContext {
    public:
+    // ncclDevComm is embedded so kernels read its hot fields from parameter
+    // memory instead of chasing a global-memory pointer. The exact-size check
+    // in nccl_device.cuh intentionally forces an update/rebuild if NCCL changes
+    // this version-specific layout.
+    static constexpr size_t kNativeCommBytes = 240;
+    static constexpr size_t kNativeCommAlignment = 8;
```
alogfans: This seems rely on a specified version of NCCL.

mooncake-wheel/mooncake/mooncake_elastic_buffer.py
```diff
@@ -166,6 +229,27 @@ def __init__(
...
+            has_nccl_device_support is not None and has_nccl_device_support()
+        )
+        self.transport = _select_transport(
+            self.requested_transport,
+            nccl_available,
+            self.num_ranks,
+            self.num_rdma_ranks,
+            self.num_nvlink_ranks,
+            self.allow_hybrid_mode,
+        )
```
alogfans: How to ensure that all nodes in the same cluster choose the same transport, or at least check it in advance?

akhillanger:
@UNIDY2002 Thanks for catching this. I fixed both compatibility issues:

- Removed the `NCCL_GIN_RESOURCE_SHARING_THREAD` mapping. Mooncake EP uses only GPU- and CTA-level sharing, so the thread value was unused and unnecessarily excluded NCCL 2.30.4.
- Replaced the exact `sizeof(ncclDevComm_t) == 240` requirement with a checked 256-byte inline capacity. NCCL 2.30.x uses 240 bytes, while 2.31.2 uses 248 bytes. 256 is the smallest natural round capacity above 248 and increases the context by only 16 bytes relative to the old implementation. We still copy only the actual `sizeof(ncclDevComm_t)`. Host- and device-side `static_assert(sizeof(ncclDevComm_t) <= 256)` checks ensure a future larger layout fails safely at compile time rather than overflowing the inline buffer. This preserves the inline fast path, avoiding a global-memory pointer chase in device code.

I validated matching header/runtime installations of `nvidia-nccl-cu13` 2.30.4, 2.30.7, and 2.31.2. For each version, `device_transport`, `device_backend_api_compatibility_test`, `mooncake_ep_device`, and `_ep` built successfully, and the two-GPU NCCL ElasticBuffer dispatch/combine correctness test passed.

This response was generated with help of AI

mooncake-transfer-engine/include/transport/device/nccl_device_transport.h
akhillanger: You are right that the original sizeof(ncclDevComm_t) == 240 check effectively restricted the implementation to NCCL releases with that exact layout. In response to @UNIDY2002's comment above, we changed this to a 256-byte inline capacity with sizeof(ncclDevComm_t) <= 256, while copying only the actual structure size. This supports the 240-byte layout in NCCL 2.30.x and the 248-byte layout in NCCL 2.31.2 while preserving inline device access. A future layout larger than the capacity will fail safely at compile time. 

The NCCL Device API remains version-specific, so the extension must still be compiled against matching NCCL headers and runtime; the change removes the unnecessary exact-layout restriction rather than claiming cross-version binary ABI compatibility. We validated matching NCCL 2.30.4, 2.30.7, and 2.31.2 installations.

mooncake-wheel/mooncake/mooncake_elastic_buffer.py
akhillanger: Good point. I addressed this in `b728ade7` by adding a process-group consensus step before either transport starts native bootstrap.

Each rank now exchanges its normalized requested mode and an NCCL-readiness bit derived from local NCCL Device API availability and topology support. The group applies the following rules:

- If normalized requests differ across ranks, every rank raises the same configuration error.
- In `auto` mode, NCCL is selected only when every rank is NCCL-ready; otherwise every rank selects IBGDA.
- With explicit `nccl`, every rank raises before bootstrap if any rank is not NCCL-ready.
- With explicit `ibgda`, every rank selects IBGDA.

This consensus finishes before NCCL unique-ID exchange or native transport construction, so ranks cannot enter different transport-specific bootstrap paths. I added focused tests for all-ready NCCL selection, group-wide IBGDA fallback, inconsistent requests, explicit-NCCL failure, and the collective wiring. The 23 focused tests and a real two-process Gloo smoke test pass.

This response was generated with help of AI

mooncake-pg/torch/BuildPgExt.cmake
akhillanger: That is right. --force is not needed when building against the currently installed PyTorch. Removed it.

docs/source/getting_started/build.md
```diff
@@ -197,7 +197,7 @@ The following options can be passed to `cmake ..`.
 | Option | Default | Description |
 | --- | --- | --- |
 | `-DUSE_CUDA=ON/OFF` | `OFF` | Enable GPU memory support, including GPUDirect RDMA, NVMe-oF, and GPU-aware TCP transport. Required when transferring GPU memory, even when using TCP. |
-| `-DUSE_NCCL_DEVICE=ON/OFF` | `OFF` | Enable the experimental NCCL DeviceTransport backend. Requires CUDA and NCCL 2.30.4+; AOT- and JIT-compiled NCCL device code must use headers that exactly match the runtime `libnccl`. |
+| `-DUSE_NCCL_DEVICE=ON/OFF` | `ON` for fresh CUDA EP builds; otherwise `OFF` | Enable the NCCL DeviceTransport backend. CUDA EP configuration falls back to `OFF` when NCCL 2.30.4+ with `nccl_device.h` is not found; passing `ON` explicitly makes the dependency required. AOT- and JIT-compiled NCCL device code must use headers that exactly match the runtime `libnccl`. |
```
UNIDY2002: We recommend keeping `USE_NCCL_DEVICE=OFF` by default for now. CUDA 13 CI wheels are built against NCCL 2.30.7 and introduce direct `libnccl.so.2` dependencies into the EP extensions; importing `mooncake.ep` fails in environments with older NCCL versions such as 2.28.9 due to missing symbols like `ncclCommQueryProperties`, even when NCCL transport is not selected. This default can be revisited once the newer NCCL symbols are loaded dynamically rather than required at import time.

mooncake-pg/torch/BuildPgExt.cmake
UNIDY2002, line 1: Could you clarify why the NCCL-related environment propagation is needed in `BuildPgExt.cmake`? At this stage, the NCCL configuration has already been applied when top-level CMake builds `transfer_engine` and `libmooncake_pg.so`; `setup.py` only builds the version-specific Torch binding against that existing core library. `mooncake-pg/torch/setup.py` also does not consume these NCCL environment variables, so they appear unused in the current build path.

docs/source/getting_started/build.md
akhillanger: Ok. Sounds good. I changed USE_NCCL_DEVICE back to OFF by default. NCCL DeviceTransport is now enabled only with -DUSE_NCCL_DEVICE=ON, which makes NCCL 2.30.4+ a required build dependency.

I also documented that NCCL-enabled EP extensions currently link directly to libnccl, so importing mooncake.ep requires a compatible runtime even when the NCCL transport is not selected. The build-time error guidance now points to the correct CMake option as well. We can revisit the default after NCCL symbols are loaded dynamically.

mooncake-pg/torch/BuildPgExt.cmake
akhillanger: You are right. This environment propagation was left over from the previous extension build flow and is not used by the current PG Torch binding. The native PG libraries are already configured and linked with NCCL by top-level CMake, while setup.py only builds the version-specific binding against libmooncake_pg.so.

I removed the unused NCCL variables and environment-propagation block from BuildPgExt.cmake.

This response was generated with help of AI.

alogfans (approved):
LGTM in principle. Please resolve code conflicts.

akhillanger:
Thanks @alogfans. Addressed the conflicts. Branch is rebased now.

mooncake-ep/include/elastic/mooncake_ep_elastic_hybrid_combine_official.cuh
```diff
@@ -743,26 +753,55 @@ __global__ void __launch_bounds__(kNumThreads, 1)
...
+                    channel_idx, scaleout_rank_idx);
+            if constexpr (Ops::kIsNccl) {
+                gin.template red_add_rel<transport::ScaleoutTeam>(
+                    remote_tail, expected_signal, lane_idx);
+            } else {
+                gin.template publish_tail<transport::ScaleoutTeam>(
+                    remote_tail, channel_idx, expected_signal, expected_signal,
+                    lane_idx);
+            }
+        }
```
yuechen-sys, line 777: There could be unnecessary to use independent NCCL implementation.

yuechen-sys (approved):
Thanks for your contribution to the Mooncake EP project! We find the current implementation clear, efficient, and ready to merge. We look forward to collaborating with you further! For furture work, We will explore elastic EP based on NCCL gin.

PR #3450, opened 2026-08-15
[Store][TE] Split Ascend agent-mode store pool across n engines

ascend-direct-dev:
## Description

In Ascend standalone (`ascend_agent_mode`), split a large store pool (for example 480G) into `n_device` master segments, each allocated and registered on a distinct ACL context / ADXL engine — the same 1:1 mapping as an embedded 16×30G layout.

Transfers resolve the dest buffer's `device_id` (serialized in Ascend buffer JSON, same field NCCL already uses) to `rank_info.endpoints[device_id]`. Same-host RoCE `+1` engine offset is removed. `batch_put`/`batch_get` that share one TE `target_id` across chunks are grouped by dest engine before execute.

**Compatibility:** mixed-version peers that omit `device_id` decode it as `-1` and fall back to `endpoints.front()` (engine 0). Both sides should upgrade together, or traffic piles onto device 0.

GPU/RDMA paths are unchanged: RDMA/TCP metadata JSON, RDMA transport, and non-`USE_ASCEND_DIRECT` store allocation are not modified.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [x] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
source /Users/pkgs/Ascend/cann/bin/setenv.bash
cd /Users/youxiao/codes/Mooncake/build
cmake -DUSE_ASCEND_DIRECT=ON ..
make mooncake_store transfer_engine ascend_transport
make ascend_direct_transport_test
./mooncake-transfer-engine/tests/ascend_direct_transport_test
```

**Test results:**
- [x] Unit tests pass (`ascend_direct_transport_test`: 80 tests PASSED)

**A3 platform, 16p_standalone:**
- Without splitting the store pool into multiple segments: average bandwidth **8 GB/s**
- With one segment per NPU (this PR): average bandwidth **30 GB/s**

clang-format-20 applied to the nine changed files. Host `pre-commit` could not fetch hook repos (GitHub 443 timeout); CI pre-commit job is the source of truth.

## Checklist

Production LOC is under the RFC threshold (test file accounts for most of the diff). No RFC.

The pre-commit checkbox: files were clang-format-20'd; full hook env was not installable from this host. Please treat the CI pre-commit job as authoritative.

## AI Assistance Disclosure

- [ ] No AI tools were used

Cursor Grok 4.6 assisted with implementation, unit tests, and this PR. Every changed line was reviewed by the submitter.

### DIFF
--- a/mooncake-store/src/real_client.cpp
+++ b/mooncake-store/src/real_client.cpp
@@ -85,6 +85,25 @@ bool IsHostStoreSegmentProtocol(const std::string &protocol) {
            protocol == "efa" || protocol == "cxi" || protocol == "rpc_only";
 }
 
+#ifdef USE_ASCEND_DIRECT
+// Split standalone store capacity across NPUs: cap each mount at total/n.
+size_t AgentModeStoreChunkCap(size_t total_size) {
+    const uint32_t n = ContextManager::getInstance().getDeviceCount();
+    if (n <= 1 || total_size == 0) {
+        return 0;
+    }
+    return total_size / n;
+}
+
+bool RestoreAgentModeDeviceZero() {
+    if (!ContextManager::getInstance().setCurrentContext(0)) {
+        LOG(ERROR) << "Failed to restore current context for device 0";
+        return false;
+    }
+    return true;
+}
+#endif
+
 std::shared_ptr<RegisteredPinnedRegion> TryPinStoreSegment(
     void *ptr, size_t size, const std::string &protocol,
     const char *segment_owner) {
@@ -914,8 +933,22 @@ tl::expected<void, ErrorCode> RealClient::setup_internal(
         uint64_t total_glbseg_size = global_segment_size;  // For logging
         uint64_t current_glbseg_size = 0;                  // For logging
 
-        const auto split_limit = GetTransportRegistrationLimit(protocol);
+        auto split_limit = GetTransportRegistrationLimit(protocol);
         const size_t alignment = facebook::cachelib::Slab::kSize;
+#ifdef USE_ASCEND_DIRECT
+        if (protocol == "ascend" && globalConfig().ascend_agent_mode) {
+            const size_t cap = AgentModeStoreChunkCap(global_segment_size);
+            if (cap > 0) {
+                if (!split_limit.has_value() || cap < *split_limit) {
+                    split_limit = cap;
+                }
+                LOG(INFO) << "Agent-mode store chunk cap: " << cap
+                          << " bytes across "
+                          << ContextManager::getInstance().getDeviceCount()
+                          << " device(s)";
+            }
+        }
+#endif
 
         // For RDMA, auto-discover NUMA nodes with NICs and distribute
         // global_segment across them for full NIC utilization.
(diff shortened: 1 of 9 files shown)


Aionw (approved):
LGTM for store side.

mooncake-transfer-engine/src/transport/ascend_transport/ascend_allocator.cpp
```diff
@@ -16,6 +17,50 @@ constexpr int kBestEffortStartPercent = 100;
...
+    }
+    const uint32_t n = mgr.getDeviceCount();
+    if (n == 0) {
+        return true;
+    }
+    uint32_t idx = 0;
+    {
+        std::lock_guard<std::mutex> lock(g_agent_alloc_mutex);
+        idx = g_agent_alloc_next % n;
+    }
```
alogfans: Why use locking instead of atomic counter?

mooncake-transfer-engine/src/transport/ascend_transport/ascend_direct_transport/transfer_executor_base.cpp
```diff
@@ -44,6 +43,40 @@ void markSlicesFailed(const std::vector<Transport::Slice*>& slice_list) {
...
+        if (buf.device_id < 0) {
+            break;
+        }
+        const auto idx = static_cast<size_t>(buf.device_id);
+        if (idx < endpoints.size()) {
+            return endpoints[idx];
+        }
+        break;
+    }
+    return endpoints.front();
```
alogfans, line 71: If metadata is broken, this fallback choice will be the first device. Advise to make the decision visible to users.

mooncake-store/src/real_client.cpp
```diff
@@ -69,6 +69,29 @@ bool IsHostStoreSegmentProtocol(const std::string &protocol) {
...
+    }
+    const size_t per_device = total_size / n;
+    if (per_device == 0) {
+        return 0;
+    }
+    return std::min(max_mr_size, per_device);
+}
+
+bool RestoreAgentModeDeviceZero() {
+    if (!ContextManager::getInstance().setCurrentContext(0)) {
```
alogfans, line 99: Should recover to the previous context before entring Mooncake's logic.

mooncake-transfer-engine/src/transport/ascend_transport/ascend_allocator.cpp
ascend-direct-dev: Good point — the mutex only guarded a counter peek/increment. Switched to `std::atomic<uint32_t>`: bind still peeks the current slot, and we `fetch_add(1)` only after a successful alloc so failed best-effort probes do not skip devices.

mooncake-transfer-engine/src/transport/ascend_transport/ascend_direct_transport/transfer_executor_base.cpp
ascend-direct-dev: Agreed. The fallback to `endpoints.front()` now logs a WARNING with the reason: missing `device_id`, `device_id` out of range, or dest address matching no buffer.

mooncake-store/src/real_client.cpp
ascend-direct-dev: This path is standalone store setup (`ascend_agent_mode`). `setup_internal` already pins ACL context to device 0 before any mount/alloc, and there is no caller ACL work before entering Mooncake. After the per-device alloc rotation we restore device 0 to match that entry state, rather than a pre-Mooncake context that does not exist here.

alogfans (approved):
LGTM in TE

ykwd:
The CI checks are all failing. It looks like we need to merge the latest `main`.

ykwd (approved):

PR #3454, opened 2026-08-15
[Feature] Add per-file storage mode to storage_benchmark_v1

rayeaster:
## Description

This PR tries to add optional argument for storage benchmark script: `file-mode` which support default `single` (current implementation) and new `per-file` which enable more flexible testing against backend storage. For example, Lustre (parallel file system) could utilize multiple OST write and local PCC read cache to improve tail IO latency

## Module

- [ ] CI/CD
- [ ] Docs
- [X ] Other

## Type of Change

- [ ] Bug fix
- [X ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

ran `benchmark.py with` both `file-mode`: `single` and `per-file`

## Checklist

- [X ] I have performed a self-review of my own code

## AI Assistance Disclosure

- [ ] No AI tools were used
- [ X] AI tools were used (specify below)

use augment code and GLM5.2 for the feature

### DIFF
--- a/benchmarks/storage_benchmark_v1/storage/disk.py
+++ b/benchmarks/storage_benchmark_v1/storage/disk.py
@@ -197,6 +197,9 @@ class DiskHashTable(Storage):
         try:
             fd, base, owned = self._open_page(page_id, write=False)
             os.pread(fd, length, base + offset_in_page)
+            if owned:
+                close_fd, fd = fd, None
+                os.close(close_fd)
 
             latency = (time.perf_counter() - start) * 1000.0
             self.stats['read_count'] += 1
@@ -249,14 +252,18 @@ class DiskHashTable(Storage):
                 os.fsync(fd)
                 self.stats['sync_count'] += 1
                 self._pending_syncs = 0
-                latency = (time.perf_counter() - start) * 1000.0
             elif self.fsync_mode == 'batch':
                 self._pending_syncs += 1
                 if self._pending_syncs >= self.fsync_batch_size:
                     os.fsync(fd)
                     self.stats['sync_count'] += 1
                     self._pending_syncs = 0
-                latency = (write_done - start) * 1000.0
+            if owned:
+                close_fd, fd = fd, None
+                os.close(close_fd)
+
+            if self.fsync_mode == 'always' or owned:
+                latency = (time.perf_counter() - start) * 1000.0
             else:
                 latency = (write_done - start) * 1000.0
 
--- a/benchmarks/storage_benchmark_v1/storage/disk.py
+++ b/benchmarks/storage_benchmark_v1/storage/disk.py
@@ -15,8 +15,9 @@ from .interface import Storage
 def calc_percentiles(data):
     """Calculate latency percentiles"""
     if not data:
-        return {'avg_ms': 0, 'p50_ms': 0, 'p95_ms': 0, 'p99_ms': 0}
+        return {"avg_ms": 0, "p50_ms": 0, "p95_ms": 0, "p99_ms": 0}
     import statistics
+
     sorted_data = sorted(data)
 
     def get_percentile(p):
@@ -29,10 +30,10 @@ def calc_percentiles(data):
         return sorted_data[lower] * (1.0 - weight) + sorted_data[upper] * weight
 
     return {
-        'avg_ms': statistics.mean(data),
-        'p50_ms': get_percentile(50),
-        'p95_ms': get_percentile(95),
-        'p99_ms': get_percentile(99),
+        "avg_ms": statistics.mean(data),
+        "p50_ms": get_percentile(50),
+        "p95_ms": get_percentile(95),
+        "p99_ms": get_percentile(99),
     }
 
 
@@ -51,10 +52,15 @@ class DiskHashTable(Storage):
     This allows simulating large traces with limited storage.
     """
 
-    def __init__(self, storage_dir: str, page_size: int,
-                 max_pages: int = 100000,
-                 file_mode: str = 'single',
-                 fsync_mode: str = 'batch', fsync_batch_size: int = 100):
+    def __init__(
+        self,
+        storage_dir: str,
+        page_size: int,
+        max_pages: int = 100000,
+        file_mode: str = "single",
+        fsync_mode: str = "batch",
+        fsync_batch_size: int = 100,
+    ):
         """Initialize disk hash table
 
         Args:
@@ -78,22 +84,22 @@ class DiskHashTable(Storage):
         self.fsync_batch_size = fsync_batch_size
         self.storage_dir.mkdir(parents=True, exist_ok=True)
         self.storage_file = self.storage_dir / "data.bin"
-        if self.file_mode == 'single':
+        if self.file_mode == "single":
             self._allocate_file()
         self.fd = None
         self._buffer = os.urandom(page_size)
         self.stats = {
-            'read_count': 0,
-            'write_count': 0,
-            'read_bytes': 0,
-            'write_bytes': 0,
-            'read_latencies_ms': [],
-            'write_latencies_ms': [],
-            'read_time_s': 0.0,
-            'write_time_s': 0.0,
-            'sync_count': 0,
-            'hit': 0,
-            'miss': 0,
+            "read_count": 0,
+            "write_count": 0,
+            "read_bytes": 0,
+            "write_bytes": 0,
+            "read_latencies_ms": [],
+            "write_latencies_ms": [],
+            "read_time_s": 0.0,
+            "write_time_s": 0.0,
+            "sync_count": 0,
+            "hit": 0,
+            "miss": 0,
         }
         self._pending_syncs = 0
         self._written_pages: set = set()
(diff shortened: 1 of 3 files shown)


stmatengss (commented):
Verdict: Approve. Correct implementation.

FD lifecycle management is handled correctly. The new _open_page() returns (fd, base_offset, owned): single mode reuses the persistent data.bin fd (owned=False), per-file mode opens a new file per operation (owned=True). Both read() and write() have finally blocks that close the fd only when owned=True, preventing fd leaks in per-file mode while leaving the persistent fd alone in single mode.

Medium: Per-file write mode uses O_WRONLY | O_CREAT | O_TRUNC, meaning every write truncates the file. This differs from single-mode behavior where pwrite at an offset preserves other data. If a future user expects incremental writes within a page, this would silently overwrite. The docstring documents it, but the truncation should be called out more explicitly in the CLI help text.

Low: DiskHashTable.__init__ never validates file_mode against known values — passing file_mode='invalid' silently falls through to single mode.

rayeaster:
> Verdict: Approve. Correct implementation.
> 
> FD lifecycle management is handled correctly. The new _open_page() returns (fd, base_offset, owned): single mode reuses the persistent data.bin fd (owned=False), per-file mode opens a new file per operation (owned=True). Both read() and write() have finally blocks that close the fd only when owned=True, preventing fd leaks in per-file mode while leaving the persistent fd alone in single mode.
> 
> Medium: Per-file write mode uses O_WRONLY | O_CREAT | O_TRUNC, meaning every write truncates the file. This differs from single-mode behavior where pwrite at an offset preserves other data. If a future user expects incremental writes within a page, this would silently overwrite. The docstring documents it, but the truncation should be called out more explicitly in the CLI help text.
> 
> Low: DiskHashTable.**init** never validates file_mode against known values — passing file_mode='invalid' silently falls through to single mode.

Thanks @stmatengss for the review. Documentation updated for the new argument `file-mode` as Medium suggests.

LujhCoconut:
Thanks for this — clean, well-scoped benchmark tooling addition. The per-file mode is fully backward compatible (all new params have defaults), the fd ownership/`finally`-close lifecycle is correct, and `choices=['single', 'per-file']` gives nice CLI validation.

One thing I'd fix before/around merge: the fsync semantics silently change in per-file mode, which can make benchmark numbers misleading (see the inline note on `write()`). With the default `--fsync-mode none` both modes are comparable, but `--fsync-mode end` currently performs zero fsyncs in per-file mode, and `--fsync-mode batch` only fsyncs the file at each batch boundary. If per-file + end/batch is a mode you intend people to measure durability with, this needs a code fix or at least an explicit doc caveat.

Two minor suggestions in the inline comments below (O_TRUNC in the docstrings, and stale `page_*.bin` cleanup). Neither blocks the feature.

benchmarks/storage_benchmark_v1/storage/disk.py
LujhCoconut: In per-file mode the fsync semantics differ from single mode in a way that could skew measurements, and it isn't documented:

- `--fsync-mode end`: `write()` never fsyncs ('end' falls through to the `else` branch), and `close()` only fsyncs `self.fd` — but `self.fd` is never set in per-file mode (it's only assigned in `_get_fd()`, which the single branch of `_open_page` reaches). So `end` = zero fsyncs.

- `--fsync-mode batch`: `_pending_syncs` is a cross-file counter, and `os.fsync(fd)` runs on the current per-op fd only, so just one file per batch is ever synced; the trailing partial batch is never flushed because `close()` can't reach it either.

The default `none` is unaffected (comparable), but if per-file + end/batch is a supported combination, consider fsyncing each per-op fd before closing it (or at least documenting that per-file end/batch are weaker than single). Otherwise the durability numbers between the two modes aren't apples-to-apples.

benchmarks/storage_benchmark_v1/storage/disk.py
```diff
@@ -147,6 +152,20 @@ def _map_page_id(self, page_id: int) -> int:
...
+
+        single: reuse the persistent data.bin fd, base = physical * page_size.
+        per-file: open page_<physical>.bin for this op, base = 0, owned = True
+                  (caller must close fd when done).
+        """
+        physical_page_id = self._map_page_id(page_id)
+        if self.file_mode == 'per-file':
+            path = self.storage_dir / f"page_{physical_page_id}.bin"
+            flags = os.O_WRONLY | os.O_CREAT | os.O_TRUNC if write else os.O_RDONLY
+            return os.open(str(path), flags, 0o644), 0, True
```
LujhCoconut: Minor: `O_TRUNC` means a partial-page write (`offset_in_page > 0` or `length < page_size`) truncates the whole file and produces a sparse file, silently discarding the rest of the page. That's fine for the current whole-page MLA layout and it's documented in the README/CLI help, but the code docstrings (`DiskHashTable.__init__` and `_open_page`) don't mention it. A future layout that emits partial access would hit this silently. Worth a one-line docstring note.

benchmarks/storage_benchmark_v1/storage/disk.py
```diff
@@ -53,25 +53,30 @@ class DiskHashTable(Storage):
...
             fsync_batch_size: Writes between fsync
         """
         self.storage_dir = Path(storage_dir)
         self.page_size = page_size
         self.max_pages = max_pages
         self.max_page_id = max_pages - 1
+        self.file_mode = file_mode
         self.fsync_mode = fsync_mode
         self.fsync_batch_size = fsync_batch_size
         self.storage_dir.mkdir(parents=True, exist_ok=True)
```
LujhCoconut, line 85: Minor: there's no cleanup of `page_*.bin` (or `thread_N/` dirs) between runs, so repeated benchmark runs accumulate stale files. Harmless within a single run (`_written_pages` starts empty and writes use `O_TRUNC`), but a re-run with a smaller `--max-pages` can leave orphaned page files. A `page_*.bin` sweep on init in per-file mode (or a README note) would be a nice touch.

rayeaster:
@LujhCoconut I agree with all your suggestion by adding more documentation for mentioned `fsync with end/batch` behavior along with `partial page write` and `orphaned stale test files/directories`:

- `fsync end/batch` is more dedicated for single-file mode IMO, so I think documentation of weaker expectation should suffice instead of forcing same semantic as single with `per-file`
- `orphaned stale test files` deletion might be helpful for repeated run but I think it only affect if smaller page chosen later, as you said. I would prefer keep the cleanup logic out of benchmark itself and let test runner to manually decide how to do with previously orphaned files/directories if any.

Thanks for your insightful review comments. Really appreciate

benchmarks/storage_benchmark_v1/storage/disk.py
```diff
@@ -244,6 +270,12 @@ def write(self, page_id: int, offset_in_page: int = 0, length: int = None) -> fl
         except OSError as e:
             print(f"Write error (page_id={page_id}, offset_in_page={offset_in_page}, length={length}): {e}")
             return 0.0
+        finally:
+            if owned and fd is not None:
+                try:
+                    os.close(fd)
```
fcczzz, line 310: Per-file `close()` is currently outside both the timing and success boundaries: `latency`, `_written_pages`, and the write stats are committed before the owned fd is closed, while any close error is swallowed. Because Python runs `finally` before returning, request wall time/QPS include the close, but the returned read/write latency (and therefore Request Storage I/O Latency and per-op bandwidth) does not. On a filesystem where close blocks or reports a deferred write error, this can understate per-file tail latency and still mark the write as present. Could we close the owned fd before recording latency/stats and propagate close failures, keeping `finally` only for best-effort cleanup?

rayeaster: @fcczzz thanks for sharing the review comment. PR is updated regarding to your suggestion by `close` before stats calculation for `per-file` mode. Please let me know if anything is missing, really appreciate

fcczzz:
Thanks — the normal path is fixed: I verified that per-file read/write latency now includes `close()`.

The close-error path is still incomplete, though. `DiskHashTable.write()` catches the close `OSError` and returns `0.0`, after which `StorageBenchmark.process_request()` unconditionally increments `write_pages`. In a fault-injection run, the caller reported one successful write while the storage layer reported zero writes and did not publish the page; retrying the same request then produced `write_pages=2` but `storage.write_count=1`.

Could we propagate the error instead of converting it to a valid latency value, or return an explicit success/error result that the caller checks before updating its counters?

rayeaster:
> Thanks — the normal path is fixed: I verified that per-file read/write latency now includes `close()`.
> 
> The close-error path is still incomplete, though. `DiskHashTable.write()` catches the close `OSError` and returns `0.0`, after which `StorageBenchmark.process_request()` unconditionally increments `write_pages`. In a fault-injection run, the caller reported one successful write while the storage layer reported zero writes and did not publish the page; retrying the same request then produced `write_pages=2` but `storage.write_count=1`.
> 
> Could we propagate the error instead of converting it to a valid latency value, or return an explicit success/error result that the caller checks before updating its counters?

@fcczzz PR updated wrt your suggestion by replacing `0.0` with `None` for failed (caught error) IO to avoid miscalculation of metrics. Would you mind to take another look? Thank you

fcczzz:
Thanks — the original close-error accounting issue is fixed now.
The current head also has a failing Ruff check; those need to be cleaned up before the CI gate can pass.

rayeaster:
> Thanks — the original close-error accounting issue is fixed now. The current head also has a failing Ruff check; those need to be cleaned up before the CI gate can pass.

@fcczzz PR updated with format cleanup, thanks

fcczzz (approved):
LGTM.

ykwd (approved):

PR #3456, opened 2026-08-16
[Bugfix][Store] Fix hybrid histogram serialization

### LINKED ISSUE #3416: [Bug]: `hybrid_histogram` serialization emits duplicate buckets, causing Prometheus to drop samples during scrape
### Environment
- Component: mooncake-store client metrics (`hybrid_histogram_1t`)
- Related file: `mooncake-store/include/hybrid_metric.h`

### Symptom
Prometheus reports the following error while scraping `/metrics` — the same metric appears multiple times with different values at the same timestamp, so samples are dropped:

```
component="scrape manager" scrape_pool=<pool> target=http://<ip>:<port>/metrics msg="Error on ingesting samples with different value but same timestamp" num_dropped=1540
```

(scrape pool name, IP and port are redacted)

### Description
When scraping the client's `/metrics`, histogram metrics (e.g. `mooncake_client_rpc_latency`) have their
per-`rpc_name` bucket series emitted **multiple times**, and the number of duplicates depends on the label's
position in the list: earlier labels are duplicated more. Counter metrics (`mooncake_client_rpc_count`) are not affected.

The duplication not only bloats the response body, but also makes the same series appear several times within a
single scrape timestamp. Since other threads keep calling `observe` concurrently during serialization, these
duplicate lines can carry inconsistent values, triggering Prometheus's `different value but same timestamp`
error and dropping 1540 samples in a single scrape.

Concrete behavior (using `mooncake_client_rpc_latency` with 11 rpc_names as an example):

| rpc_name | occurrences |
|---|---|
| FetchTasks | 11 |
| GetStorageConfig | 10 |
| … | … |
| last rpc_name | 1 |

### Reproduction
1. Enable client metrics reporting (`MC_STORE_CLIENT_METRIC`, enabled by default).
2. Have the client invoke several different RPCs (e.g. FetchTasks / Ping / GetStorageConfig).
3. Scrape the `/metrics` endpoint and observe the duplicated `*_latency_bucket` series.
4. On the Prometheus side, observe `Error on ingesting samples with different value but same timestamp` and a steadily increasing `num_dropped`.

### Evidence
Counting `FetchTasks` buckets with value `9789` (the 13 buckets with `le >= 7000`, including `+Inf`):

```
143 = 13 × 11
```

where 13 is the number of `le`-buckets ≥ 7000 in the FetchTasks histogram, and 11 is the total number of
rpc_names — i.e. the full FetchTasks histogram was emitted 11 times, matching its position at index 0.

### Root Cause
In `basic_hybrid_histogram::serialize()`:

```cpp
void serialize(std::string& str) override {
    auto value_map = sum_->copy();
    ...
    serialize_head(str);

    std::string value_str;          // ← declared outside the loop, never reset
    auto bucket_counts = get_bucket_counts();
    for (auto& e : value_map) {      // one iteration per label (rpc_name)
        ...
        for (size_t i = 0; i < bucket_counts.size(); i++) {
            value_str.append(name_).append("_bucket{...");  // keeps appending to the same string
            ...
        }
        str.append(value_str);       // re-emits ALL accumulated buckets from previous labels
        ...
    }
}
```

`value_str` is declared outside the loop and only ever `append`ed, never cleared, so each newly processed label
re-emits the buckets of every previously processed label, producing O(n²) duplication.

Affected metrics (all use `hybrid_histogram_1t`):
- `MasterClientMetric::rpc_latency` → `mooncake_client_rpc_latency`
- `TransferOperationMetric::read_op_latency_us` → `mooncake_transfer_read_operation_latency`
- `TransferOperationMetric::write_op_latency_us` → `mooncake_transfer_write_operation_latency`

### Before submitting...

JimmyWang0417:
## Description

Closes #3416.

`basic_hybrid_histogram::serialize()` reused its bucket scratch string across label values, so earlier bucket series were emitted repeatedly. It also called `str.clear()` when all stored sums were zero, which erased metrics that earlier serializers had already appended to the shared response buffer.

This change clears the scratch string for each label and rolls back only the bytes appended by the current histogram when it emits no samples.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
g++ -std=c++20 -I. -Iextern/yalantinglibs/include \
  ../mooncake-w1-minimal-test.cpp -lgtest -lgtest_main -pthread \
  -o ../mooncake-w1-minimal-test
../mooncake-w1-minimal-test --gtest_color=no
git clang-format --diff origin/main -- \
  mooncake-store/include/hybrid_metric.h \
  mooncake-store/tests/client_metrics_test.cpp
git diff origin/main...HEAD --check
```

**Test results:**
- [ ] Unit tests pass
- [x] Manual testing done: the focused CPU-only gtest harness failed both cases on `dc549aec` (duplicate count `6` instead of `3`; existing buffer became empty) and passed both cases after the fix.

The registered `client_metrics_test` CMake target was not built locally because it unconditionally links Transfer Engine, cachelib, etcd wrapper, and `ibverbs`; this host does not have the ibverbs headers or library. The added repository tests therefore rely on upstream CI for the full target build. The repository's format script also requires clang-format 20, while this host has clang-format 22.1.8; `git clang-format --diff` with the available version reported no changes. `pre-commit` is not installed locally.

## Checklist

- [ ] I have updated the documentation (not applicable; no user-facing contract changed)
- [ ] For changes >500 LOC: I have filed an RFC issue (not applicable; 46 insertions and 2 deletions)

## AI Assistance Disclosure

- [ ] No AI tools were used

AI tools assisted with investigation and drafting. The submitter reviewed every changed line and validation result.

### DIFF
--- a/mooncake-store/include/hybrid_metric.h
+++ b/mooncake-store/include/hybrid_metric.h
@@ -289,8 +289,10 @@ class basic_hybrid_histogram : public dynamic_metric {
             return;
         }
 
+        const auto initial_size = str.size();
         serialize_head(str);
 
+        bool emitted_any = false;
         std::string value_str;
         auto bucket_counts = get_bucket_counts();
         for (auto& e : value_map) {
@@ -300,6 +302,7 @@ class basic_hybrid_histogram : public dynamic_metric {
                 continue;
             }
 
+            value_str.clear();
             value_type count = 0;
             for (size_t i = 0; i < bucket_counts.size(); i++) {
                 auto counter = bucket_counts[i];
@@ -324,6 +327,7 @@ class basic_hybrid_histogram : public dynamic_metric {
             }
 
             str.append(value_str);
+            emitted_any = true;
 
             std::string labels_str;
             build_label_string_with_static(labels_str, sum_->labels_name(),
@@ -341,8 +345,8 @@ class basic_hybrid_histogram : public dynamic_metric {
             str.append(std::to_string(count));
             str.append("\n");
         }
-        if (value_str.empty()) {
-            str.clear();
+        if (!emitted_any) {
+            str.resize(initial_size);
         }
     }
 
--- a/mooncake-store/tests/client_metrics_test.cpp
+++ b/mooncake-store/tests/client_metrics_test.cpp
@@ -40,6 +40,16 @@ int GetTestPort(std::unordered_set<int>& used_ports) {
     return -1;
 }
 
+size_t CountOccurrences(const std::string& text, const std::string& needle) {
+    size_t count = 0;
+    size_t position = 0;
+    while ((position = text.find(needle, position)) != std::string::npos) {
+        ++count;
+        position += needle.size();
+    }
+    return count;
+}
+
 class ScopedEnv {
    public:
     explicit ScopedEnv(const char* name) : name_(name) {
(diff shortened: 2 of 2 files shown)


he-yufeng (approved):
Verified both failure modes against main's header with a standalone build, then against this branch:

- main, two observed labels: beta's bucket block comes out with 6 lines because `value_str` still carries alpha's 3; the per-iteration `value_str.clear()` fixes exactly that (beta=3 after).
- main, all-zero series: `str.clear()` would nuke the head and any pre-existing buffer content; `str.resize(initial_size)` preserves both (checked with a pre-filled buffer, output unchanged).
- Walked the mixed path too (one zero-sum label skipped, one hot): skip leaves `value_str` cleared for the next iteration, so no residue; standalone run confirms zero-label=0 lines, hot=2 bucket lines.

The `emitted_any` flag is doing real work here: resize-on-empty is only safe because the head was just appended and nothing followed it. One note for the record, not a blocker: the new tests cover duplication and all-zero, and the mixed zero/hot case is the one remaining shape, currently only covered by reading the loop. A third case in the same test file would nail it, but the loop makes it mechanical.

CI is red repo-wide right now (wheel builders fail on patchelf/auditwheel before tests run), so none of this ran in CI; the above is from compiling the header standalone. Approving on that basis.

JimmyWang0417:
Thank you for the thorough validation and approval. I appreciate you checking both original failure modes and the mixed zero/hot path independently.

I also ran a focused mixed-path harness against the PR's base header and this branch. That shape passes on both, so it is useful coverage but not a regression case for either bug fixed here. Since you marked it non-blocking, I plan to keep this PR focused on the two RED-to-GREEN regressions already included rather than change the approved head.

Thanks again for the careful review.

Aionw (approved):
LGTM

ykwd (approved):

PR #3457, opened 2026-08-16
[Bugfix][Store] Preserve RequiredParam names when copied

JimmyWang0417:
## Description

`RequiredParam<T>` defined copy operations that copied only `value_`. Copy
assignment therefore retained the destination's old parameter name, while copy
construction left `name_` uninitialized. After clearing a copied value,
`Get()` could report the wrong parameter or read an uninitialized pointer while
building the error message.

This change defaults both copy operations so the optional value and parameter
name are preserved together. Focused tests cover copy construction and copy
assignment through the public `Get()` interface.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
g++ -std=c++20 -Imooncake-store/include \
  mooncake-store/tests/config_helper_test.cpp \
  -lgtest -lgtest_main -pthread -o /tmp/config_helper_test
/tmp/config_helper_test --gtest_color=no
git clang-format --diff origin/main -- \
  mooncake-store/include/config_helper.h \
  mooncake-store/tests/config_helper_test.cpp
git diff origin/main...HEAD --check
```

**Test results:**
- [ ] Unit tests pass
- [x] Manual testing done: the exact repository test source failed both cases
  against `dc549aec` (both errors retained `old_name`) and passed both cases
  against the fixed header.

The registered `config_helper_test` CMake target was not built locally because
`add_store_test` unconditionally links Transfer Engine, cachelib, the etcd
wrapper, and `ibverbs`; those Store dependencies are unavailable on this host.
The standalone CPU-only build above compiles and runs the exact added test file,
while the full registered target relies on upstream CI. The repository format
script requires clang-format 20, while this host has clang-format 22.1.8;
`git clang-format --diff` with the available version reported no changes.
`pre-commit` is not installed locally.

## Checklist

- [ ] I have updated the documentation (not applicable; no user-facing contract changed)
- [ ] For changes >500 LOC: I have filed an RFC issue (not applicable; 53 insertions and 9 deletions)

## AI Assistance Disclosure

- [ ] No AI tools were used

AI tools assisted with investigation and drafting. The submitter reviewed every
changed line and validation result.

### DIFF
--- a/mooncake-store/include/config_helper.h
+++ b/mooncake-store/include/config_helper.h
@@ -20,14 +20,8 @@ class RequiredParam {
     RequiredParam() : name_(nullptr) {}
     RequiredParam(const char* name) : name_(name) {}
 
-    // Add copy constructor
-    RequiredParam(const RequiredParam& other) { value_ = other.value_; }
-
-    // Add copy assignment operator
-    RequiredParam& operator=(const RequiredParam& other) {
-        value_ = other.value_;
-        return *this;
-    }
+    RequiredParam(const RequiredParam&) = default;
+    RequiredParam& operator=(const RequiredParam&) = default;
 
     /**
      * @brief Assignment operator to set the value
--- a/mooncake-store/tests/CMakeLists.txt
+++ b/mooncake-store/tests/CMakeLists.txt
@@ -73,6 +73,7 @@ add_store_test(master_scenario_test master_scenario.cpp
                master_scenario_test.cpp)
 add_store_test(master_service_scenario_test master_scenario.cpp
                master_service_scenario_test.cpp)
+add_store_test(config_helper_test config_helper_test.cpp)
 add_store_test(master_service_config_test master_service_config_test.cpp)
 add_store_test(master_service_processing_key_double_erase_test
                master_service_processing_key_double_erase_test.cpp)
--- a/mooncake-store/tests/config_helper_test.cpp
+++ b/mooncake-store/tests/config_helper_test.cpp
@@ -1,0 +1,49 @@
+#include <gtest/gtest.h>
+
+#include <stdexcept>
+#include <string>
+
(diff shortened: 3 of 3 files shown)


stmatengss (commented):
Verdict: Approve. Real bug, correct fix.

The hand-written copy constructor and copy assignment only copied value_, omitting name_. After a copy, name_ was a garbage pointer (default constructor sets it to nullptr), so calling Get() on an unset copy would throw std::runtime_error with an invalid what() string — undefined behavior via dereference of a bad pointer. The copy assignment also kept the destination's old name, so error messages would report the wrong parameter name. Replacing both with = default correctly copies all members.

Medium: Neither regression test covers copying an UNSET source — the exact crash path. The = default fix handles it correctly, but there is no test guarding against a future regression. Recommend adding a CopyUnsetPreservesNullName test.

JimmyWang0417:
Thank you for the positive verdict and the careful review. This is one of my first contributions here, so I'm genuinely excited to receive such encouraging feedback!

Great catch on the test gap. I've added `CopyUnsetPreservesNullName` to cover both copy construction and copy assignment from a never-set source. The focused test fails against the pre-fix implementation with corrupted or stale parameter names and passes with the defaulted copy operations. I also tightened the helper to assert the full diagnostic message rather than a substring.

Thanks again for taking the time to review this so thoroughly.

Aionw (approved):
LGTM.

Nit: UT here seems not that necessary, because those test cases just testing cpp copy ctr which already tested by cpp itself.

stmatengss (approved):

PR #3462, opened 2026-08-16
feat(store): MOONCAKE_OFFLOAD_MAX_PHYSICAL_BYTES for bucket eviction

### LINKED ISSUE #3426: [Bug]: Disk eviction relies on fs::space(), which ignores cgroup/emptyDir quotas — pod gets OOM-evicted
### Bug Report

Summary

BucketStorageBackend (and StorageBackend file-per-key) has two eviction guards: a logical quota (total_size_ > max_total_size) and a physical guard added in #2419 (fs::space(storage_path_).available < required + 256MB). In a containerized deployment (K8s emptyDir with sizeLimit, or any cgroup/quota-limited mount), std::filesystem::space() returns the host filesystem capacity, not the enforced quota. The physical guard therefore never fires, and because the logical total_size_ counter systematically undercounts real on-disk usage, physical disk usage silently exceeds the intended limit until the kubelet out-of-band evicts the whole pod (exit 137).

Environment

- SGLang HiCache L3 → Mooncake, 2× TP8 instances sharing the backend, write_through, DSPARK.
- Backend: bucket_storage_backend, MOONCAKE_OFFLOAD_BUCKET_EVICTION_POLICY=lru, MOONCAKE_OFFLOAD_BUCKET_MAX_TOTAL_SIZE=200GiB.
- K8s pod, offload path on an emptyDir volume with sizeLimit: 300Gi. Underlying fs is the node overlay (reports 878G total / 556G avail).

Observed

Under a real workload (input 50k–500k tokens, output 1k–2k), physical /nvme usage climbed past the 200GiB logical limit to 292G (256k tier) → 301G (500k tier), at which point kubelet evicted both pods: Evicted: Usage of EmptyDir volume "nvme" exceeds the limit "300Gi". Logical total_size_ stayed at ~200GiB the whole time; eviction was running normally by the logical quota.

Root cause (verified in source)

storage_backend.cpp PrepareEviction():
auto space_info = fs::space(storage_path_, ec);       // returns HOST overlay: 556G avail
uint64_t actual_available = space_info.available;
initial_disk_full = actual_available < req_sz + kMinFreeSpace;   // always false under emptyDir
- The physical guard (disk_still_full, from #2419) never triggers because fs::space() sees the 878G host fs, not the 300Gi emptyDir quota that kubelet actually enforces out-of-band.
- The logical guard works, but total_size_ only sums data_size + meta_size; per the code's own comment "actual disk usage >= data_size + meta_size" (256MB bucket granularity + block alignment + deferred space reclaim). So a 200GiB logical limit corresponds to ~290–300G physical.
- Net effect: in containers the backend effectively has only a logical limit, and that logical limit does not bound physical usage.

Impact

Any containerized deployment with a disk quota (K8s emptyDir sizeLimit, cgroup v2 io/space lover-run its quota and be killed out-of-band, regardless of MAX_TOTAL_SIZE. Users must guess
a heavily discounted logical limit (we had to drop to 150GiB to keep physical under 300Gi) w.

Proposed fix

Add an explicit, container-aware physical byte cap that does not rely on fs::space() auto-deFLOAD_MAX_PHYSICAL_BYTES. When set, drive eviction off min(fs::space().available,max_physical_bytes − current_physical_usage), tracking actual on-disk bytes (statx/allocated size) rather than logical data_size + meta_size. This lets operators declare the real quota when fs::space() cannot
discover it.

For reference: offset_allocator_storage_backend is immune because it pre-allocates a fixed-sis bounded by the file size, independent of fs::space() accuracy). The bucket/file-per-key
backends need an equivalent hard physical bound for containerized use.

Notes

Related: #2419 introduced the fs::space()-based physical check for the bare-metal shared-disk case; this report is about that same mechanism being defeated by container quotas.

### Before submitting...

dalegraydesk:
## Summary
- Add `MOONCAKE_OFFLOAD_MAX_PHYSICAL_BYTES` for bucket backends.
- When set, `PrepareEviction` treats available space as `min(fs::space().available, max_physical_bytes - logical_used)` so container quotas that `fs::space()` cannot see still trigger eviction.

## Test plan
- [ ] Set `MOONCAKE_OFFLOAD_MAX_PHYSICAL_BYTES` below host free space and confirm eviction fires before kubelet emptyDir eviction
- [ ] Unset env: behavior matches previous `fs::space()`-only guard

Fixes #3426

### DIFF
--- a/mooncake-store/include/storage_backend.h
+++ b/mooncake-store/include/storage_backend.h
@@ -216,8 +216,8 @@ struct BucketBackendConfig {
 
     // Optional hard physical cap for container quotas where fs::space() reports
     // the host filesystem instead of the volume sizeLimit / cgroup quota.
-    // When > 0, eviction also treats (max_physical_bytes - total_size_) as
-    // available space. 0 = disabled (rely on fs::space only).
+    // When > 0, eviction also caps available space using logical used bytes
+    // (total + pending eviction/write). 0 = disabled (rely on fs::space only).
     int64_t max_physical_bytes = 0;
 
     bool Validate() const;
--- a/mooncake-store/include/storage_backend.h
+++ b/mooncake-store/include/storage_backend.h
@@ -214,6 +214,12 @@ struct BucketBackendConfig {
     int64_t max_total_size = 0;  // 0 = unlimited; evict when total_size_
                                  // exceeds this threshold (bytes)
 
+    // Optional hard physical cap for container quotas where fs::space() reports
+    // the host filesystem instead of the volume sizeLimit / cgroup quota.
+    // When > 0, eviction also treats (max_physical_bytes - total_size_) as
+    // available space. 0 = disabled (rely on fs::space only).
+    int64_t max_physical_bytes = 0;
+
     bool Validate() const;
 
     static BucketBackendConfig FromEnvironment();
--- a/mooncake-store/src/storage_backend.cpp
+++ b/mooncake-store/src/storage_backend.cpp
@@ -2906,6 +2906,10 @@ BucketStorageBackend::PrepareEviction(
         // Container quotas (emptyDir sizeLimit / cgroup) are invisible to
         // fs::space(); when configured, also bound available by physical cap.
         if (bucket_backend_config_.max_physical_bytes > 0) {
+            // In-flight evictions count bytes in both total_size_ and
+            // pending_eviction_size_ until removal completes, so `used`
+            // double-counts briefly. That biases toward early eviction
+            // (safer when a kubelet sizeLimit is looming).
             const int64_t used = total_size_ + pending_eviction_size_ +
                                  pending_write_size_;
             const int64_t remaining =
--- a/mooncake-store/src/storage_backend.cpp
+++ b/mooncake-store/src/storage_backend.cpp
@@ -105,6 +105,9 @@ BucketBackendConfig BucketBackendConfig::FromEnvironment() {
                           Environ::GetInt64("MOONCAKE_BUCKET_MAX_TOTAL_SIZE",
                                             config.max_total_size));
 
+    config.max_physical_bytes = Environ::GetInt64(
+        "MOONCAKE_OFFLOAD_MAX_PHYSICAL_BYTES", config.max_physical_bytes);
+
     const auto policy_str = Environ::GetString(
         "MOONCAKE_OFFLOAD_BUCKET_EVICTION_POLICY",
         Environ::GetString("MOONCAKE_BUCKET_EVICTION_POLICY", "fifo"));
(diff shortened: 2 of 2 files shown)


he-yufeng:
Read this against the accounting paths in storage_backend.cpp. One question on the `used` sum: between PrepareEviction marking a bucket (pending_eviction_size_ += evicted_size, ~2981) and the final removal (total_size_ -= object_size and pending_eviction_size_ -= evicted_size, ~2957/3006), those bytes are counted in both total_size_ and pending_eviction_size_. With `used = total_size_ + pending_eviction_size_ + pending_write_size_`, in-flight evictions double-count, so the cap triggers earlier than configured precisely while eviction is running. The bias is in the safe direction (over-evict beats under-evict when a kubelet sizeLimit is looming), so this may be intentional, but a one-line comment saying so would stop the next reader from "fixing" it.

Naming nit, take or leave: the sibling cap is MOONCAKE_BUCKET_MAX_TOTAL_SIZE while this one is MOONCAKE_OFFLOAD_MAX_PHYSICAL_BYTES; both land on BucketBackendConfig. Either prefix works, but matching the existing BUCKET one would read more naturally in the same env block.

Not approving since this is a quota path and the runtime behavior needs a container-quota environment to exercise; the test plan items are unchecked, so I assume those runs are still pending.

Morpheus799:
Reporter of #3426 here. I built this the other way — measuring **actual on-disk bytes**
instead of the logical `total_size_` — and ran it live on the same setup. Sharing the
results, because they show why a `used = total_size_ + pending_eviction_size_ +
pending_write_size_` basis won't hold the disk in the real deployment.

**The `used` sum is logical, so it inherits two problems that leave physical disk unbounded:**

**1. Per-instance scope on a shared directory (the dominant one).**
SGLang HiCache runs one Mooncake store — hence one `BucketStorageBackend` — *per TP rank*,
and passes a single `ssd_offload_path` to all of them (no per-rank subdir). So N backends
(N=8 for TP8) share ONE offload directory, yet each enforces the cap on *its own*
`total_size_`. Confirmed with a one-line diagnostic on official v0.3.12.post1: 8 instances,
all reading the same real directory usage but each `total_size_` only ~1/N of it. Under a
sustained 500k-token run the shared disk climbed to **231 GB with `[Evict] triggered` = 0**
(no instance's `total_size_` ever approached the 200 GiB cap) and kubelet OOM-evicted the
pods. With `used = total_size_ + …`, `max_physical_bytes - used` stays large on every
instance, so this PR would behave the same in that deployment.

**2. `total_size_` under-counts real disk (~1.5×).** It's decremented per object while
256 MB bucket files linger until fully drained, so even single-instance a 200 GiB `used`
corresponds to ~300 GB physical.

**What actually holds the disk:** measure the real directory — `Σ stat.st_blocks*512 over
storage_path_` (== `du`, == kubelet's emptyDir accounting), cached briefly. Because every
co-located backend scans the *same* directory, they observe the same true shared usage and
bound it together. Live result with this variant, same 200 GiB cap, same 500k load: disk
**held at ~200 GiB (199.7–200.1), pods survived, eviction fired 160+ times per instance** —
and it fired on the *physical* measurement (physical ≈ 200 GiB = cap) while each instance's
`total_size_` was only ~100 GiB. A logical-`used`-based cap would have seen `used` ≈ 100 GiB
≪ 200 GiB and not evicted.

This is exactly @LujhCoconut's point 2 ("track actual on-disk bytes, not logical
data_size+meta_size"). @he-yufeng's double-count observation is real, but it's *within* the
logical sum; fixing it doesn't change that `used` is per-instance and ~1.5× low, so it still
can't bound a shared physical disk.

I have a working implementation of the st_blocks-scan variant (enforced in PrepareEviction,
evicting until projected post-eviction usage + the incoming write fits under the cap, cache
invalidated after file deletion). I will soon open another PR to show how it works.

dalegraydesk:
Thanks for the detailed review on the accounting model.

- Merged latest `main` into the branch.
- Documented that `used = total_size_ + pending_eviction_size_ + pending_write_size_` briefly double-counts in-flight evictions (intentionally conservative for quota pressure).

Acknowledged that per-TP-instance scope on a shared offload directory is a separate limitation; this PR stays scoped to the logical cap hook from #3426.

PR #3464, opened 2026-08-16
TENT: deliver self-targeted notifications in-process

xiaodouzi666:
## Description

The TENT data plane short-circuits transfers whose target is `LOCAL_SEGMENT_ID` to the local path, but notifications have no such branch: `sendNotification()` hands every target to the first transport that supports notifications. Over RDMA that reaches the local pseudo-endpoint, which never performs the bootstrap handshake and so never owns a notification QP — the send fails (`Notification QP not connected`, `rdma_transport.cpp` → `endpoint.cpp`) and a receiver polling for the notification waits forever.

We hit this end to end through the NIXL Mooncake plugin (see the RFC below): the first intra-agent transfer carrying a notification hangs its receiver until timeout. It appears to have gone unnoticed because the upstream intra-agent DRAM unit test is commented out and the VRAM variant only compiles with CUDA on a GPU machine.

Fix: deliver self-targeted notifications in-process — queue them in the engine on send, drain the queue in `receiveNotification()` after polling the transport. This mirrors the data plane's local short-circuit, works for every transport combination, and needs no wire at all (including builds where no transport supports notifications, where a queued self-notification is now delivered rather than rejected).

A follow-up commit in this branch fixes something the review turned up: `maybeFireSubmitHooks()` restarts its target loop from the beginning on every poll, so once self-targeted sends always succeed, a batch that also notifies an unreachable peer re-queues the local copy on each poll. Each target is now dropped from the hook as it takes delivery.

RFC: ai-dynamo/nixl#2083 — this is the engine-side fix for the issue recorded there as the intra-agent notification hang.
Related: #3405, #3406 (both touch TENT; no conflicts with this branch).

## Module

## Type of Change

- [ ] Bug fix

## How Has This Been Tested?

**Test commands:**
```bash
cmake -G Ninja -DUSE_TENT=ON -DUSE_HTTP=ON -DBUILD_UNIT_TESTS=ON -DBUILD_EXAMPLES=ON ..
ninja && ctest --output-on-failure   # tent suite: 32/32 passed
```

**Test results:**
- [x] Unit tests pass

The new `local_notification_test` covers in-order exactly-once delivery of self-targeted notifications; resolution of the engine's own advertised name and of the empty name to `LOCAL_SEGMENT_ID` (in p2p mode the advertised name is the normalized `ip:port`, not the configured `local_segment_name`); and delivery with every transport disabled. It needs no NIC and no metadata server, so it runs in the CPU-only CI legs.

End to end: the NIXL Mooncake-plugin intra-agent VRAM test that used to hang (`Checking notification flow:` never returning, engine logging `Notification QP not connected`) now completes both its P-Thr=OFF and P-Thr=ON passes with zero QP errors. Reproduced on two H20 machines with RDMA; both runs used a byte-identical test binary and differed only in the engine shared library, so the change in behaviour is attributable to this patch.

## Checklist

- [x] For changes >500 LOC: I have filed an RFC issue (413 LOC here; an RFC covering this work exists at ai-dynamo/nixl#2083)

On the pre-commit box: every hook passes on the changed files except `cmake-format`, which rewrites 57 lines of pre-existing content in `tent/tests/CMakeLists.txt` — `main` is already not clean under the pinned v0.6.13. Reformatting it here would bury the change in unrelated churn, so I left it alone. The CI format gate (clang-format 20, changed lines only) passes.

## AI Assistance Disclosure

- [ ] No AI tools were used

AI-assisted development tools were used during diagnosis and drafting. I have reviewed every changed line and can defend the change end to end.

### DIFF
--- a/mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
+++ b/mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
@@ -2231,15 +2231,22 @@ Status TransferEngineImpl::maybeFireSubmitHooks(Batch* batch, bool check) {
             }
         }
         if (!all_completed) continue;
-        Status last = Status::OK();
-        for (auto target_id : hook.targets) {
-            last = sendNotification(target_id, hook.notifi);
-            if (!last.ok()) {
-                LOG(WARNING) << "sendNotification failed: " << last.ToString();
+        // Drop each target as it takes delivery. The hook is only marked
+        // fired once every target has, so this loop reruns from a later poll
+        // while any of them still fails, and the ones that already succeeded
+        // must not be notified again. Self-targeted sends made that
+        // reachable: they now always succeed, so a peer that stays
+        // unreachable used to re-queue the local notification on every poll.
+        for (auto it = hook.targets.begin(); it != hook.targets.end();) {
+            auto status = sendNotification(*it, hook.notifi);
+            if (!status.ok()) {
+                LOG(WARNING)
+                    << "sendNotification failed: " << status.ToString();
                 break;
             }
+            it = hook.targets.erase(it);
         }
-        if (last.ok()) hook.fired = true;
+        if (hook.targets.empty()) hook.fired = true;
     }
     return Status::OK();
 }
--- a/mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
+++ b/mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
@@ -2231,20 +2231,24 @@ Status TransferEngineImpl::maybeFireSubmitHooks(Batch* batch, bool check) {
             }
         }
         if (!all_completed) continue;
-        // Drop each target as it takes delivery. The hook is only marked
-        // fired once every target has, so this loop reruns from a later poll
-        // while any of them still fails, and the ones that already succeeded
-        // must not be notified again. Self-targeted sends made that
-        // reachable: they now always succeed, so a peer that stays
-        // unreachable used to re-queue the local notification on every poll.
+        // Drop each target as it takes delivery and carry the rest to a later
+        // poll. The hook is only marked fired once every target has, so the
+        // ones that already succeeded must not be notified again.
+        //
+        // Every remaining target is attempted on every pass, including the
+        // ones after a failure. targets is an unordered_set, so stopping at
+        // the first failure would make delivery to the others depend on
+        // iteration order: an unreachable peer visited first would hold back
+        // a self-targeted notification that cannot fail, and its receiver
+        // would block until that unrelated peer came back.
         for (auto it = hook.targets.begin(); it != hook.targets.end();) {
             auto status = sendNotification(*it, hook.notifi);
-            if (!status.ok()) {
-                LOG(WARNING)
-                    << "sendNotification failed: " << status.ToString();
-                break;
+            if (status.ok()) {
+                it = hook.targets.erase(it);
+                continue;
             }
-            it = hook.targets.erase(it);
+            LOG(WARNING) << "sendNotification failed: " << status.ToString();
+            ++it;
         }
         if (hook.targets.empty()) hook.fired = true;
     }
--- a/mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
+++ b/mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
@@ -2456,6 +2456,10 @@ Status TransferEngineImpl::probePeerAliveByID(SegmentID target_id) {
 
 Status TransferEngineImpl::receiveNotification(
     std::vector<Notification>& notifi_list) {
+    // Each poll reports what this poll delivered. Callers reuse one vector
+    // across polls, and appending to leftovers would redeliver what the
+    // caller already handled.
+    notifi_list.clear();
     Status status = Status::OK();
     bool has_transport = false;
     for (size_t type = 0; type < kSupportedTransportTypes; ++type) {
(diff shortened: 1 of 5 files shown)


mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
```diff
@@ -2146,12 +2157,31 @@ Status TransferEngineImpl::probePeerAliveByID(SegmentID target_id) {
...
 Status TransferEngineImpl::receiveNotification(
     std::vector<Notification>& notifi_list) {
+    Status status = Status::OK();
+    bool has_transport = false;
     for (size_t type = 0; type < kSupportedTransportTypes; ++type) {
         auto& transport = transport_list_[type];
         if (!transport || !transport->supportNotification()) continue;
-        return transport->receiveNotification(notifi_list);
+        has_transport = true;
+        status = transport->receiveNotification(notifi_list);
```
staryxchen, line 2480: Here reuses one vector (poll, process, poll again without clear()) will keep previously delivered local notifications and append new ones. Suggest: notifi_list.clear() at the start of TransferEngineImpl::receiveNotification

mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
```diff
@@ -2146,12 +2157,31 @@ Status TransferEngineImpl::probePeerAliveByID(SegmentID target_id) {
...
+    // Append self-targeted notifications queued by sendNotification(). They
+    // are deliverable even when no transport supports notifications at all.
+    {
+        std::lock_guard<std::mutex> lk(local_notifi_mutex_);
+        if (!local_notifi_list_.empty()) {
+            notifi_list.insert(
+                notifi_list.end(),
+                std::make_move_iterator(local_notifi_list_.begin()),
+                std::make_move_iterator(local_notifi_list_.end()));
+            local_notifi_list_.clear();
```
staryxchen, line 2493: Suggest: if any local notifications were appended, return OK (or skip the drain when the transport receive failed).

mooncake-transfer-engine/tent/tests/local_notification_test.cpp
staryxchen, line 1: local_notification_test covers the queue primitive, not the hang. The NIXL intra-agent failure was submitTransfer(..., notifi) → maybeFireSubmitHooks() → sendNotification(LOCAL_SEGMENT_ID). A local-only submit + poll + receiveNotification() would pin that path without a NIC. Also please disable transports/sunrise_link in makeConfig so WorksWithoutAnyNotificationTransport is actually transport-free, and poll twice with the same output vector (no clear()) so we do not silently depend on the caller allocating a fresh list each time.

mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
xiaodouzi666: Done in 239103a — `receiveNotification()` now clears the output list on entry.

It mattered more than it looked: both delivery tests now poll twice through the same vector, and they fail without the clear. Thanks for catching it.

mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
xiaodouzi666: Done in 239103a. A poll that delivered something returns OK, and the local queue is drained regardless of the transport poll outcome — in-process delivery does not depend on it. Notifications that landed in the list have already been taken off their queue, so the caller has to see them.

One caveat I would rather flag than leave for you to find: the reasoning above assumes a later poll comes back empty and surfaces the error then. Under a steady trickle of local notifications that poll may never come, so a permanently broken transport can stay masked. Dropping notifications we have already consumed looked like the worse trade, but if you would rather have a transport error reported unconditionally I will change it.

mooncake-transfer-engine/tent/tests/local_notification_test.cpp
xiaodouzi666: All three done in 239103a.

`TransferBoundSelfNotificationIsDelivered` goes through `submitTransfer(..., notifi)` and polls to COMPLETED, so it exercises `maybeFireSubmitHooks() -> sendNotification(LOCAL_SEGMENT_ID)` rather than the queue primitive.

One thing was needed to make it a regression test rather than a round trip: with a real transport installed the engine's own RPC server answers on loopback, so a self-addressed notification reaches itself over the wire even without the fix. The test swaps in a stand-in transport that advertises notification support and fails to send, which is what the RDMA local pseudo-endpoint does. Checked both ways — with the local short-circuit removed it fails, with it the suite passes.

`makeConfig` now disables all eleven transports the loader can populate, `sunrise_link` included, and both delivery tests poll twice through the same output vector.

Following your third point into the hook path turned up something else: `maybeFireSubmitHooks()` restarts its target loop from the beginning on every poll, so once self-targeted sends started succeeding, a batch that also notified an unreachable peer re-queued the local copy on each poll. That is 409521f, with a two-engine test that sees seven deliveries without it and one with it.

catyans (changes requested):
The per-target erase fixes repeated delivery only when a successful target is visited before the failing one. `targets` is an `unordered_set`, and this loop stops at the first send failure. If an unreachable remote peer happens to be visited before `LOCAL_SEGMENT_ID`, the local notification is never attempted and a self-targeted receiver remains blocked until that unrelated peer recovers. That leaves the liveness of the feature dependent on unordered iteration order.

Please continue past failed targets, erasing each successful target and retaining only the failures for the next poll. A regression test should keep the remote peer unavailable and assert that the local target is delivered promptly—and exactly once—before remote recovery.

xiaodouzi666:
You're right, and it's worse than a missed case. Fixed in dd9f670, the loop now erases each target that takes delivery and keeps the rest for the next pass, so every remaining target is attempted every time.

catyans (approved):
Re-reviewed `dd9f670e1a`. The loop now attempts every remaining target on each pass, erases each successful target immediately for exactly-once delivery, and retains only failures for retry. The two-unreachable-peer regression makes the liveness assertion independent of `unordered_set` iteration order and confirms that the self-targeted notification is delivered before remote recovery. The original issue is resolved. Approved.

The current `tent-ci (ub-mock)` failure is in the unrelated `RuntimeQueueDispatch.ProgressWorkerRefillsWindowFromTransportNotify` test; the notification-specific change and tests pass. Please rerun or resolve that CI leg before merge.

alogfans (approved):
LGTM

PR #3467, opened 2026-08-17
[Store] Bucket: add MAX_PHYSICAL_BYTES cap on real shared-disk usage

### LINKED ISSUE #3426: [Bug]: Disk eviction relies on fs::space(), which ignores cgroup/emptyDir quotas — pod gets OOM-evicted

Morpheus799:
## Summary

Adds an optional hard cap on the bucket backend's **actual on-disk** usage,
`MOONCAKE_OFFLOAD_BUCKET_MAX_PHYSICAL_BYTES`, enforced by measuring the real
allocation of the offload directory (`Σ stat.st_blocks*512`, the same basis as
`du` and kubelet's emptyDir accounting) rather than the in-memory `total_size_`.
Disabled by default (`0`) — no behavior change and zero overhead when unset.

Fixes #3426.

## Motivation

In the common SGLang HiCache + TP deployment, none of the existing guards bound
the physical offload directory, and the pod gets OOM-evicted despite
`MAX_TOTAL_SIZE` being set:

1. **`max_total_size` is enforced per `BucketStorageBackend` instance.** HiCache
   creates one Mooncake store — hence one bucket backend — **per TP rank**, and
   all of them share a single offload directory (one `ssd_offload_path`, no
   per-rank subdir). Each instance caps only its own `total_size_`, i.e. ~1/N of
   the shared disk, so no instance ever reaches the cap and eviction never fires.
2. **`total_size_` under-counts real disk (~1.5×)** even single-instance: it is
   decremented per object, while a 256 MB bucket file is only removed once *all*
   its objects are gone (cf. #3220/#3221).
3. **`fs::space()` is quota-blind in containers** — it reports the host
   filesystem, not the cgroup / emptyDir `sizeLimit`, so the existing physical
   guard never triggers.

## What this PR does

- `ActualDiskBytesUsedLocked()` sums real file allocation (`stat.st_blocks*512`)
  over `storage_path_`, cached briefly (see `DISK_SCAN_CACHE_MS`).
- `PrepareEviction` snapshots this once, then evicts buckets until the projected
  post-eviction usage **plus the incoming write** fits under the cap; if it still
  cannot fit after the eviction budget, it rejects the write with
  `FILE_WRITE_FAIL` (handled as an offload miss upstream) instead of overrunning.
- The cache is invalidated after `FinalizeEviction` deletes files.

Because every co-located backend scans the **same** directory, they all observe
the same true shared usage and bound the shared disk together — which the
per-instance `total_size_` cannot. Measuring real `st_blocks` also removes the
~1.5× undercount and does not depend on `fs::space()`.

## Configuration

| Env var | Default | Description |
|---|---|---|
| `MOONCAKE_OFFLOAD_BUCKET_MAX_PHYSICAL_BYTES` | `0` (off) | Hard cap on actual on-disk bytes of the offload dir. Set to the real volume quota minus headroom. |
| `MOONCAKE_OFFLOAD_BUCKET_DISK_SCAN_CACHE_MS` | `500` | How long the directory-scan resulter but a staler bound; smaller = tighter but scans more often; `<=0` = scan every check. |

Requires `MOONCAKE_OFFLOAD_BUCKET_EVICTION_POLICY != none` (eviction is a no-op
otherwise). `offset_allocator` is unaffected — its fixed preallocated file
already bounds disk.

## Validation

Official v0.3.12.post1 baseline, 2×H100 TP8 (DeepSeek-V4-Flash), HiCache→Mooncake,
`/nvme` emptyDir `sizeLimit=300Gi`, 500k-token workload, cap = 200 GiB. Confirmed
the per-instance behavior with a one-line diagnostic log (8 backends per pod, all
reading the same real dir usage but each `total_size_` only ~1/N):

- **Without this cap:** physical climbed to **231 GB with `[Evict] triggered` = 0**
  (no instance's `total_size_` ever approached 200 GiB) → kubelet OOM-evicted.
- **With this cap:** disk **held at ~200 GiB (199.7–200.1), pods survived,
  eviction fired 160+× per instance** — driven by the *physical* measurement
  (physical ≈ 200 GiB = cap) while each instance's `total_size_` was only ~100 GiB
  (a logical-based cap would not have fired).

## Notes / limitations

- Each instance can only evict its own buckets, so eviction distribution across
  co-located instances is not globally optimal, but the physical bound holds. A
  node-level shared physical accounting is left as future work.
- Compare with #3462, which drives the same cap off logical `total_size_`
  (per-instance and ~1.5× low) rather than real on-disk bytes, so it does not bound
  the shared disk in the TP deployment above.


Morpheus799:
I'm now testing this pr on the latest main branch to make sure it works. Down below is the impact on performance.

**Disabled (default `MAX_PHYSICAL_BYTES=0`): zero overhead.** No scan runs and the
`> 0` guard short-circuits every added branch, so existing deployments are unaffected.

**Enabled:** the only real cost is the directory scan in `ActualDiskBytesUsedLocked()`
(`stat` each file, sum `st_blocks`). For the bucket backend this is a modest file count
(256 MB buckets → a few hundred to ~1.4k files at 200–350 GiB), ~1–2 ms per scan on a
warm inode cache. It is:
- **snapshotted once per `PrepareEviction`**, not per evicted bucket, and
- **cached for `DISK_SCAN_CACHE_MS` (default 500 ms)**, so it runs at most ~2×/s
  regardless of offload rate (tunable; `<=0` disables caching for a tighter bound).

The scan happens under the backend's metadata lock, i.e. a ~1–2 ms hold at most ~2×/s —
negligible next to prefill/decode latency, and adjustable via `DISK_SCAN_CACHE_MS` (or
moved off the lock in a follow-up if needed). **The read path (`BatchLoad`) is untouched.**

The extra eviction once near the cap is the intended function (bounding the shared disk),
not overhead — it reuses the existing FIFO/LRU eviction; the alternative is the disk
overrunning its quota and the pod being OOM-evicted.

Empirically, a ~2 h continuous 500k-token run with the cap at 200 GiB held the disk at
199.6–199.9 GiB with ~1.4k evictions per instance and both servers healthy throughout —
no measurable serving degradation.

Morpheus799:
Just ran a small over-cap test using wheel built on the latest main branch. Disk dropped from ~22.7GiB straight to 10260MB (~10GiB = the cap) via 32 evictions, so this pr did work. @LujhCoconut have a look?

LujhCoconut:
Thanks for this PR! Could you also update the related docs — mainly adding the two new env vars to the settings table in                         
`docs/source/deployment/ssd/ssd-offload.md` and a brief note on when to prefer this cap over `MAX_TOTAL_SIZE` (shared-dir / emptyDir deployments)?

Morpheus799:
> Thanks for this PR! Could you also update the related docs — mainly adding the two new env vars to the settings table in `docs/source/deployment/ssd/ssd-offload.md` and a brief note on when to prefer this cap over `MAX_TOTAL_SIZE` (shared-dir / emptyDir deployments)?

finished

Morpheus799:
Waiting for some reviewers……

LujhCoconut (approved):
LGTM. Thanks for your contribution.

Morpheus799:
出了点问题，让我研究下

Morpheus799:
感觉目前这个方案不太灵活，逻辑模式就只管每实例单独的目录，physcap就只管共享目录，够用倒是够用

Morpheus799:
I want to flag a semantic ambiguity in MAX_PHYSICAL_BYTES: because ActualDiskBytesUsedLocked() scans storage_path_ (Σ st_blocks*512, matching du/kubelet), the cap's meaning changes with the directory layout:

- Shared directory (all TP ranks → one dir): every rank scans everything, so the value caps the total across all ranks (the whole volume).
- Separate dirs / disks (per-rank subdir or distinct mounts): each rank scans only its own files, so the same value caps each rank individually.

So one config value has two meanings, and the user must know their layout to set it right (on a shared 300 GiB volume it's the total; with per-rank subdirs on that same volume it must be set to total / N, or it over-commits by N×).

For context, the logical counter total_size_ doesn't resolve this either: at runtime it only reflects the instance's own writes/evictions, but it under-counts real disk (~1.5×: no block rounding, and 256 MB bucket files linger until fully evicted while it decrements per object) — and on restart in a shared dir, Init() scans the whole directory and folds all ranks' files into this instance's total_size_ (a recovery-time double-count， not reliable). So it isn't a clean per-rank physical measure to fall back on.

The scope question: if we want a cap on the sum of all ranks' real usage that holds under any layout (including ranks on separate disks), a per-directory scan can't do it — no process can readdir across disks. That needs each rank to report its physical usage and the master to aggregate. The master has most of the plumbing (LocalDiskSegment.ssd_used_bytes, OffloadObjectHeartbeat, the PushOffloadingQueue gate), but ssd_used_bytes is logical today, there's no live physical-usage uplink, and no master-driven eviction — a non-trivial master-side change.

Your call:
- (A) Extend to a master-aggregated global cap now, or
- (B) Keep this PR client-side, I would tighten the docs to state precisely what the cap means per layout and how to set it later. Merge as-is, and track the aggregated cap as a follow-up.

Happy either way — just want to align before adding master-side changes here.

@LujhCoconut please have a look

Morpheus799:
总之就是不动master侧的情况下也只能这样了，或者也可以通过维护文件列表的方式计算出每个rank的真实占用，取代逻辑估算值。但要是想把它们加起来然后统一设置个cap的话，应该就得动master侧。

LujhCoconut:
我觉得先按现在来，bugfix首先解决，有一些优化的话可以follow-up。现在的状态我觉得没啥问题。挺好的了。

Morpheus799:
> 我觉得先按现在来，bugfix首先解决，有一些优化的话可以follow-up。现在的状态我觉得没啥问题。挺好的了。

行，那反正现在是没问题了，docs我也补了你可以看下，可以merge了

LujhCoconut:
晚点CI过了，maintainer会来merge的，赞👍

ykwd (approved):

ykwd:
Thanks for the great work!

PR #3468, opened 2026-08-17
feat: Add strict_local_numa to hard-exclude cross-NUMA RDMA rails

XiaokunDing:
## Background

On dual-socket / multi-NUMA machines, selecting an RDMA NIC from a remote NUMA node for GPU buffer transfers can introduce unnecessary cross-socket traffic and unstable performance. In some deployments, it may also increase the chance of RDMA failures such as `LOC_PROT` or QP transition errors.

The existing `numa_penalties` setting only provides a soft preference. A remote NUMA NIC can still be selected when its score is better due to EWMA bandwidth, inflight bytes, or probe / round-robin sampling. This is not strict enough for environments where GPU traffic must stay on the local NUMA node.

## Change

This MR adds a `strict_local_numa` switch for RDMA device selection.

When enabled, the RDMA scheduler excludes the cross-NUMA tier from the normal candidate list, so weighted selection, probe mode, and round-robin selection will not actively pick a remote NUMA NIC. If no same-NUMA candidate is available, the existing fallback still retries all tiers to avoid making the node unusable.

The CUDA topology discovery path is also adjusted so GPU Tier-0 devices are limited to same-NUMA NICs. PCIe distance is now used only as a secondary ranking within the same NUMA node.

## Example Configuration

```json
{
  "topology": {
    "rdma_whitelist": [
      "mlx5_bond_1",
      "mlx5_bond_4",
      "mlx5_bond_3",
      "mlx5_bond_2",
      "mlx5_bond_7",
      "mlx5_bond_6",
      "mlx5_bond_8",
      "mlx5_bond_5"
    ]
  },
  "transports": {
    "rdma": {
      "enable": true,
      "numa_penalties": [1.0, 5.0, 100000.0],
      "strict_local_numa": true
    }
  }
}
```

## Compatibility

`strict_local_numa` is disabled by default, so existing soft NUMA penalty behavior is preserved unless explicitly enabled.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
```

**Test results:**
- [ ] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

### DIFF
--- a/mooncake-transfer-engine/tent/src/common/config.cpp
+++ b/mooncake-transfer-engine/tent/src/common/config.cpp
@@ -85,6 +85,16 @@ static inline void setConfig(Config& config, const std::string& env_key,
     if (val) config.setFromString(config_key, std::string(val));
 }
 
+// Bool env vars: setFromString() stores "1" as int, so get(..., false) misses
+// it.
+static inline void setBoolConfig(Config& config, const std::string& env_key,
+                                 const std::string& config_key) {
+    const char* val = std::getenv(env_key.c_str());
+    if (!val) return;
+    config.set(config_key,
+               ConfigHelper::parseBool(val, config.get(config_key, false)));
+}
+
 // Like setConfig, but parses the env value as a comma-separated list and
 // stores it as a string array. Empty/whitespace-only items are dropped so a
 // trailing comma or spaces around names are tolerated (e.g. "mlx5_0, mlx5_1").
@@ -167,8 +177,8 @@ Status ConfigHelper::loadFromEnv(Config& config) {
               "transports/rdma/disable_gpu_direct_rdma");
     setConfig(config, "MC_LOG_RDMA_SLICE_AFFINITY",
               "transports/rdma/log_slice_affinity");
-    setConfig(config, "MC_STRICT_LOCAL_NUMA",
-              "transports/rdma/strict_local_numa");
+    setBoolConfig(config, "MC_STRICT_LOCAL_NUMA",
+                  "transports/rdma/strict_local_numa");
     // Restrict which RDMA NICs the engine discovers/uses (comma-separated
     // device names). MC_TE_FILTERS is an allow-list — same name and semantics
     // as the legacy Transfer Engine's device whitelist, so a single env works
--- a/mooncake-transfer-engine/tent/src/common/config.cpp
+++ b/mooncake-transfer-engine/tent/src/common/config.cpp
@@ -167,6 +167,8 @@ Status ConfigHelper::loadFromEnv(Config& config) {
               "transports/rdma/disable_gpu_direct_rdma");
     setConfig(config, "MC_LOG_RDMA_SLICE_AFFINITY",
               "transports/rdma/log_slice_affinity");
+    setConfig(config, "MC_STRICT_LOCAL_NUMA",
+              "transports/rdma/strict_local_numa");
     // Restrict which RDMA NICs the engine discovers/uses (comma-separated
     // device names). MC_TE_FILTERS is an allow-list — same name and semantics
     // as the legacy Transfer Engine's device whitelist, so a single env works
--- a/mooncake-transfer-engine/tent/src/platform/cuda/cuda_probe.cpp
+++ b/mooncake-transfer-engine/tent/src/platform/cuda/cuda_probe.cpp
@@ -193,39 +193,29 @@ static void discoverCudaTopology(std::vector<Topology::NicEntry>& nic_list,
         }
         for (char* ch = pci_bus_id; (*ch = to_lower(*ch)); ch++);
         int numa_node = getNumaNodeFromPciDevice(pci_bus_id);
+        int min_distance = INT_MAX;
+        std::unordered_map<int, std::vector<int>> distance_map;
+        for (const auto& device : nic_list) {
+            int dist = getPciDistance(device.pci_bus_id.c_str(), pci_bus_id);
+            distance_map[dist].push_back(&device - &nic_list[0]);
+            min_distance = std::min(min_distance, dist);
+        }
 
         Topology::MemEntry entry;
         entry.name = "cuda:" + std::to_string(i);
         entry.numa_node = numa_node;
         entry.pci_bus_id = pci_bus_id;
         entry.type = Topology::MEM_CUDA;
-
-        // Tier 0 is same-NUMA only; PCIe distance is secondary. Distance can
-        // collapse (ties / realpath failure) and would otherwise pick a remote
-        // NIC. Cross-NUMA NICs are never promoted here: when no same-NUMA NIC
-        // exists rank 0 stays empty and all NICs fall into the cross-NUMA rank,
-        // so strict_local_numa selection can hard-exclude them.
-        int min_distance = INT_MAX;
-        std::unordered_map<int, std::vector<int>> distance_map;
-        int nic_id = 0;
-        for (const auto& device : nic_list) {
-            if (numa_node >= 0 && device.numa_node == numa_node) {
-                int dist =
-                    getPciDistance(device.pci_bus_id.c_str(), pci_bus_id);
-                distance_map[dist].push_back(nic_id);
-                min_distance = std::min(min_distance, dist);
-            }
-            nic_id++;
+        if (distance_map.count(0)) {
+            // Prefer NICs with distance 0 (e.g. same PCIe switch/RC)
+            entry.device_list[0] = std::move(distance_map[0]);
+        } else if (distance_map.count(min_distance)) {
+            // No exact match — fall back to NICs with closest PCIe distance
+            entry.device_list[0] = std::move(distance_map[min_distance]);
         }
-
-        if (!distance_map.empty()) {
-            int pick = distance_map.count(0) ? 0 : min_distance;
-            entry.device_list[0] = std::move(distance_map[pick]);
-        }
-
         std::unordered_set<int> preferred_set;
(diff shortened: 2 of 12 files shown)


alogfans (commented):
`strict_local_numa` still does not provide **hard exclusion** semantics in the current implementation.

 There are three remaining paths where a cross-NUMA NIC can still be selected:

 1. **Baseline mode bypasses the flag entirely**  
    In `quota.cpp`, the `!smart_selection_enabled_` path iterates ranks until it finds the first non-empty one, but it does not apply any `strict_local_numa` filtering. If all local ranks are empty, it will still pick the cross-NUMA tier.

 2. **Smart-mode fallback re-adds cross-NUMA devices**  
    In `buildCandidates()`, the first pass skips the last rank when `strict_local_numa` is enabled, but if the filtered candidate set becomes empty, the fallback path rebuilds candidates from **all** ranks. That means remote NUMA NICs are still reintroduced under strict mode.

 3. **CUDA topology can hide cross-NUMA NICs inside rank 0**  
    In `cuda_probe.cpp`, when no same-NUMA NIC exists, the fallback path places the closest remote NICs into `device_list[0]`. After that, the selector can no longer distinguish them from true local-NUMA devices, so rank-based filtering is no longer sufficient.

I think the intended strict-mode semantics should be:

- when `strict_local_numa == true`, cross-NUMA NICs must never enter the eligible set;
- if no same-NUMA NIC exists for the location, return `DeviceNotFound` instead of falling back to a remote NIC;
- baseline mode, smart selection, and fallback should all share the same rank-filtering rule.

In other words, this change currently implements a **strong preference** against cross-NUMA NICs, but not a true **hard exclusion** yet.

XiaokunDing:
> `strict_local_numa` still does not provide **hard exclusion** semantics in the current implementation.
> 
> There are three remaining paths where a cross-NUMA NIC can still be selected:
> 
> 1. **Baseline mode bypasses the flag entirely**
>    In `quota.cpp`, the `!smart_selection_enabled_` path iterates ranks until it finds the first non-empty one, but it does not apply any `strict_local_numa` filtering. If all local ranks are empty, it will still pick the cross-NUMA tier.
> 2. **Smart-mode fallback re-adds cross-NUMA devices**
>    In `buildCandidates()`, the first pass skips the last rank when `strict_local_numa` is enabled, but if the filtered candidate set becomes empty, the fallback path rebuilds candidates from **all** ranks. That means remote NUMA NICs are still reintroduced under strict mode.
> 3. **CUDA topology can hide cross-NUMA NICs inside rank 0**
>    In `cuda_probe.cpp`, when no same-NUMA NIC exists, the fallback path places the closest remote NICs into `device_list[0]`. After that, the selector can no longer distinguish them from true local-NUMA devices, so rank-based filtering is no longer sufficient.
> 
> I think the intended strict-mode semantics should be:
> 
> * when `strict_local_numa == true`, cross-NUMA NICs must never enter the eligible set;
> * if no same-NUMA NIC exists for the location, return `DeviceNotFound` instead of falling back to a remote NIC;
> * baseline mode, smart selection, and fallback should all share the same rank-filtering rule.
> 
> In other words, this change currently implements a **strong preference** against cross-NUMA NICs, but not a true **hard exclusion** yet.

Thanks for the careful review — you're right that the original change was only a strong preference, not a hard exclusion. I've reworked it so strict_local_numa now enforces true hard-exclusion semantics across all three paths.

mooncake-transfer-engine/tent/src/platform/cuda/cuda_probe.cpp
```diff
@@ -193,29 +193,39 @@ static void discoverCudaTopology(std::vector<Topology::NicEntry>& nic_list,
...
+        int min_distance = INT_MAX;
+        std::unordered_map<int, std::vector<int>> distance_map;
+        int nic_id = 0;
+        for (const auto& device : nic_list) {
+            if (numa_node >= 0 && device.numa_node == numa_node) {
+                int dist =
+                    getPciDistance(device.pci_bus_id.c_str(), pci_bus_id);
+                distance_map[dist].push_back(nic_id);
+                min_distance = std::min(min_distance, dist);
+            }
```
staryxchen: Hard exclusion currently treats “NUMA/location unknown” the same as “cross-NUMA”, so allocations that should still pick a NIC fail with `DeviceNotFound`.

`isNumaRankEligible()` drops the last rank whenever the flag is on. That is only correct when rank 2 means “known remote”. Two discovery paths put **every** NIC in that rank when information is missing:

1. **`numa_node == -1`** here: rank 0/1 are filled only when GPU and NIC NUMA are known and equal. If sysfs returns `-1` (VMs, some GPUs, often `mlx5_bond_*`), all NICs land in rank 2.
2. **Wildcard `*`** from `insertFallbackMemEntry()`: every NIC is placed in `device_list[2]`. Workers also fall back to `*` when location cannot be resolved.

Please apply hard exclusion only when **both** the memory location and the NIC have a known NUMA id and they differ. If either side is unknown, keep the old soft-penalty path. A test for `numa_node == -1` and one for allocate against `*` would cover this.

mooncake-transfer-engine/tent/src/transport/rdma/workers.cpp
```diff
@@ -111,6 +111,9 @@ Workers::Workers(RdmaTransport* transport)
         }
     }
 
+    params.strict_local_numa =
+        conf->get("transports/rdma/strict_local_numa", false);
```
staryxchen, line 115: `strict_local_numa` is not applied on the worker retry path, so a cross-NUMA NIC can still be posted after the first attempt fails.

mooncake-transfer-engine/tent/src/platform/cuda/cuda_probe.cpp
```diff
@@ -193,29 +193,39 @@ static void discoverCudaTopology(std::vector<Topology::NicEntry>& nic_list,
...
-        } else if (distance_map.count(min_distance)) {
-            // No exact match — fall back to NICs with closest PCIe distance
-            entry.device_list[0] = std::move(distance_map[min_distance]);
+
+        // Tier 0 is same-NUMA only; PCIe distance is secondary. Distance can
+        // collapse (ties / realpath failure) and would otherwise pick a remote
+        // NIC. Cross-NUMA NICs are never promoted here: when no same-NUMA NIC
+        // exists rank 0 stays empty and all NICs fall into the cross-NUMA rank,
+        // so strict_local_numa selection can hard-exclude them.
+        int min_distance = INT_MAX;
```
staryxchen: The CUDA rank-0 rewrite is not gated on `strict_local_numa`. A remote NIC that previously sat in rank 0 (closest PCIe / collapsed distance) now always sits in rank 2 and takes the default 10× `numa_penalties` hit even with the flag off. That may be desirable, but it contradicts the compatibility note that soft-penalty behavior is unchanged unless the flag is enabled. Please either gate the topology change on the flag, or document it as an intentional default ranking change.

mooncake-transfer-engine/tent/src/common/config.cpp
```diff
@@ -167,6 +167,8 @@ Status ConfigHelper::loadFromEnv(Config& config) {
               "transports/rdma/disable_gpu_direct_rdma");
     setConfig(config, "MC_LOG_RDMA_SLICE_AFFINITY",
               "transports/rdma/log_slice_affinity");
+    setConfig(config, "MC_STRICT_LOCAL_NUMA",
+              "transports/rdma/strict_local_numa");
```
staryxchen: `MC_STRICT_LOCAL_NUMA=1` will not turn the feature on. `setFromString` stores `"1"` as an integer; `get(..., false)` then fails the JSON bool conversion and returns the default `false`. Only `"true"`/`"false"` work. Please accept `1`/`0`, or document that the env var must be `true`.

staryxchen (commented):
Hard exclusion is incomplete in a few places outside `DeviceSelector`. Inline comments cover unknown NUMA/wildcard, the worker retry path, the unconditional CUDA rank-0 change, and the env-var bool parsing.

One item that cannot be attached to this diff: ROCm / MemoryProber still promote the closest-PCIe NIC into rank 0 even when it is remote (`rocm_probe.cpp`, `memory_prober.cpp`). On those paths a cross-NUMA NIC never reaches rank 2, so `strict_local_numa` does not exclude it. Please apply the same same-NUMA-only rank-0 rule used in `cuda_probe.cpp`, or the flag is CUDA-only.

catyans:
Thanks—the source-side selector, unknown-NUMA handling, and env parsing concerns are addressed. One target-side escape remains in the production worker path.

`selectOptimalDevice()` selects from the first non-empty target rank and only *prefers* a same-NUMA NIC within that rank; if none exists, it keeps the cross-NUMA choice instead of scanning later ranks or failing. `selectFallbackDevice()` filters cross-NUMA `sdev`, but never filters `tdev`. A remote topology with a cross-NUMA NIC in rank 0 and a local NIC in rank 1 can therefore still post on the remote target rail, both initially and on retry, while `strict_local_numa=true`.

The new tests cover `DeviceSelector`, not these `Workers` paths. Could the same topology-aware eligibility rule be applied to every target candidate, with a worker-path regression test proving that initial selection and retry return `DeviceNotFound` rather than choosing a known cross-NUMA target?

staryxchen (approved):
LGTM

PR #3476, opened 2026-08-17
Fix data copy while not on same device

XiaokunDing:
## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
```

**Test results:**
- [ ] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

### DIFF
--- a/mooncake-transfer-engine/tent/src/runtime/control_plane.cpp
+++ b/mooncake-transfer-engine/tent/src/runtime/control_plane.cpp
@@ -427,8 +427,8 @@ void ControlService::onSendData(const std::string_view& request,
     }
 
     if (local_desc->findBuffer(peer_mem_addr, length)) {
-        auto status = Platform::getLoader().copy((void*)peer_mem_addr, &desc[1],
-                                                 length);
+        auto status =
+            Platform::getLoader().copy((void*)peer_mem_addr, &desc[1], length);
         if (!status.ok()) {
             // A non-empty response is interpreted as an RPC error by the
             // client (see ControlClient::sendData). Without this the sender's
--- a/mooncake-transfer-engine/tent/src/runtime/control_plane.cpp
+++ b/mooncake-transfer-engine/tent/src/runtime/control_plane.cpp
@@ -467,8 +467,8 @@ void ControlService::onRecvData(const std::string_view& request,
                             length);
         } else {
             response.resize(length);
-            auto status = loader.copy(response.data(), (void*)peer_mem_addr,
-                                      length);
+            auto status =
+                loader.copy(response.data(), (void*)peer_mem_addr, length);
             if (!status.ok()) {
                 // Clear the payload so the client sees a length mismatch (see
                 // ControlClient::recvData) rather than copying partial/garbage
--- a/mooncake-transfer-engine/tent/src/runtime/control_plane.cpp
+++ b/mooncake-transfer-engine/tent/src/runtime/control_plane.cpp
@@ -84,8 +84,11 @@ Status ControlClient::sendData(const std::string& server_addr,
         // and the extra copy in call().
         request.append(reinterpret_cast<const char*>(local_mem_addr), length);
     } else {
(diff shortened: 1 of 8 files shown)


alogfans (approved):
LGTM. Will merge after fixing format problems.

XiaokunDing:
> LGTM. Will merge after fixing format problems.

fixed

alogfans:
@copilot resolve the merge conflicts in this pull request

XiaokunDing:
> @copilot resolve the merge conflicts in this pull request

done

mooncake-transfer-engine/tent/src/runtime/control_plane.cpp
```diff
@@ -106,8 +106,7 @@ Status ControlClient::recvData(const std::string& server_addr,
     if (response.size() != length)
         return Status::RpcServiceError(
             "RecvData failed: target address not in registered buffer");
-    Platform::getLoader().copy(local_mem_addr, response.data(), length);
-    return Status::OK();
+    return Platform::getLoader().copy(local_mem_addr, response.data(), length);
```
staryxchen, line 112: The new `onRecvData` path reports copy failures by replacing the payload with an error string, but `recvData` still treats any size mismatch as

```
"RecvData failed: target address not in registered buffer"
```

and drops the server's message. After a CUDA copy failure the sender would retry/fail looking like an unregistered buffer, which is the opposite of what this change is trying to make diagnosable.

The RecvData protocol (success iff `response.size() == length`) also means that if the error string happens to have the same length as the request, the client will copy that text into `local_mem_addr` and return OK.

Could we fix both in this PR?

1. Client: if `size != length`, surface `response` (fall back to a generic message only when it is empty), same idea as `sendData`.
2. Server: if the error string length equals `length`, pad it (e.g. append a byte) so it cannot be interpreted as a successful payload.

Happy to take this in a follow-up if you prefer to keep this PR strictly to the stream-device fix — but it is a one-line client change plus a small pad on the error path we are already adding.

mooncake-transfer-engine/tent/src/runtime/control_plane.cpp
```diff
@@ -428,7 +427,15 @@ void ControlService::onSendData(const std::string_view& request,
     }
 
     if (local_desc->findBuffer(peer_mem_addr, length)) {
-        Platform::getLoader().copy((void*)peer_mem_addr, &desc[1], length);
+        auto status =
+            Platform::getLoader().copy((void*)peer_mem_addr, &desc[1], length);
+        if (!status.ok()) {
+            // A non-empty response is interpreted as an RPC error by the
+            // client (see ControlClient::sendData). Without this the sender's
+            // transfer would be reported COMPLETED even though the destination
```
staryxchen, line 438: The new handling here is right: fail the RPC when the destination copy fails so the sender is not told COMPLETED after a failed write.

The symmetric hole is still open on the client. `ControlClient::sendData` ignores the local GPU→request copy:

```
loader.copy(request.data() + sizeof(desc), local_mem_addr, length);
```

`std::string::resize` zero-fills the payload. If that copy fails we send zeros, the peer's `onSendData` copy (host → registered GPU) typically succeeds, and an empty response makes this COMPLETED — the same class of bug this comment describes.

Could you `CHECK_STATUS` that copy (or equivalent) in this PR as well? The device-stream fix should make the common cuda:R case succeed; this is only so a residual copy error cannot be reported as success.

staryxchen (commented):
Two leftover copy-error holes in the TCP control-plane path — not regressions from the stream-device fix, but they sit next to the error handling this PR already adds. Would you mind folding both into this PR?

staryxchen (approved):
LGTM, will merge after CI pass

Issue #3465, opened 2026-08-17
[Bug]: Silent offload failure: keys exist on offload SSD but return empty value after eviction offload


waliwali777:
### Describe the bug
We observe silent offload failure against local SSD offload backend. Keys are visible via `/get_all_keys`, but corresponding values return empty when fetched.

We are testing with recent Mooncake commit: https://github.com/kvcache-ai/Mooncake/commit/67aaf6de74f35c7c5cfd072e8643b0a80eb5999a (2026‑08‑12).
The same workload with June‑end Mooncake code works perfectly without empty‑value issue. This regression appears after we upgraded to newer codebase.

### Master launch command:

```C++
mooncake_master \
--rpc_port=8001 \
--rpc_thread_num=8 \
--metrics_port=8007 \
--default_kv_lease_ttl=5000 \
--allow_evict_soft_pinned_objects=true \
--eviction_ratio=0.35 \
--eviction_high_watermark_ratio=0.7 \
--cluster_id=mooncake_cluster \
--allocation_strategy=free_ratio_first \
--quota_bytes=3298534883328 \
--enable_ha=false \
--enable_offload=true \
--offload_on_evict=true \
--offload_force_evict=false \
--enable_disk_eviction=true \
--promotion_on_hit=false \
--put_start_discard_timeout_sec=300 \
--put_start_release_timeout_sec=600

```
### Real‑Client Environment Variables & Settings
``` C++
MC_IB_PCI_RELAXED_ORDERING                         = 1
MOONCAKE_LOCAL_BUFFER_SIZE                         = 6442450944
MOONCAKE_OFFLOAD_BUCKET_EVICTION_POLICY            = lru
MOONCAKE_OFFLOAD_BUCKET_KEYS_LIMIT                 = 128
MOONCAKE_OFFLOAD_BUCKET_MAX_TOTAL_SIZE             = 3113851289600
MOONCAKE_OFFLOAD_BUCKET_SIZE_LIMIT_BYTES           = 268435456
MOONCAKE_OFFLOAD_CLIENT_BUFFER_GC_INTERVAL_SECONDS = 2
MOONCAKE_OFFLOAD_FILE_STORAGE_PATH                 = /ssd_file/mooncake_cache/
MOONCAKE_OFFLOAD_HEARTBEAT_INTERVAL_SECONDS        = 1
MOONCAKE_OFFLOAD_LOCAL_BUFFER_SIZE_BYTES           = 21474836480
[...]

PR #3479, opened 2026-08-17
fix(store): deduplicate SSD carryover keys

### LINKED ISSUE #3478: [Bug]: SSD carryover keys can be emitted into multiple offload buckets
### Describe the bug

The bucket SSD backend can schedule the same key into more than one offload bucket when a key carried over in `ungrouped_offloading_objects_` is returned by the master again on the next heartbeat.

`GroupOffloadingKeysByBucket()` prepends the carryover keys, then iterates the new heartbeat batch without deduplicating keys already present in the carryover set. Depending on unordered-map iteration order and bucket limits, the same key can be emitted in two buckets. The first bucket persists successfully; the later bucket fails with `OBJECT_ALREADY_EXISTS`, causing the whole later bucket to be NACKed.

This creates repeated silent per-bucket offload failures under sustained memory eviction and can leave unrelated keys in the rejected bucket waiting for another offload attempt.

### Reproduction

Tested on current `main` commit `411c61c216e42c49b2a98a5f69d692d78a0ce981` with:

- bucket storage backend
- `offload_on_evict=true`
- 2.5 GiB memory segment
- 10 GiB of 1 MiB values
- 8 concurrent writers
- `MOONCAKE_OFFLOAD_BUCKET_KEYS_LIMIT=128`
- io_uring and RDMA enabled

The workload consistently logged repeated failures such as:

```text
Duplicate offload reservation: key=..., stored=1, evicting=0, writing=0
Failed to store objects with error: OBJECT_ALREADY_EXISTS
```

A scaled run produced 118 duplicate reservations. The report configuration from #3465 produced 330 `OBJECT_ALREADY_EXISTS` failures during a 100 GiB run.

### Expected behavior

A key carried over from the previous heartbeat should appear at most once in the next set of generated buckets. Bucket grouping should count unique keys toward both key-count and size limits.

### Root cause

`GroupOffloadingKeysByBucket()` does not deduplicate the current heartbeat input against `ungrouped_offloading_objects_`. Existing unit tests also repeated the same input and expected the duplicated carryover entries to fill an additional bucket, unintentionally encoding the incorrect behavior.

### Additional context

Related to #3465 because it occurs on the same local SSD eviction/offload path and creates silent offload failures. This issue tracks the independently reproducible duplicate-bucketing bug; it does not claim to fully explain the empty-value symptom in #3465.

982945902:
## Summary

- deduplicate keys carried over in `ungrouped_offloading_objects_` from keys returned in the next heartbeat
- fill buckets according to the number of unique keys rather than loop iterations
- update the existing grouping tests so new keys, rather than duplicate carryover entries, fill subsequent buckets
- add a focused regression test for a carryover key repeated in the next heartbeat batch

## Root cause

`GroupOffloadingKeysByBucket()` prepended carryover keys and then iterated the new heartbeat input without excluding keys already present in that carryover set. The same key could therefore be emitted into more than one bucket. Once the first bucket committed, a later bucket failed atomically with `OBJECT_ALREADY_EXISTS` and all keys in that bucket were NACKed.

## Validation

- `FileStorageTest.GroupOffloadingKeysByBucket*`: 5/5 passed
- pre-fix scaled RDMA workload: 118 duplicate offload reservations
- post-fix scaled RDMA workload: zero `OBJECT_ALREADY_EXISTS` or persistence errors
- post-fix workload: 10,240 × 1 MiB values, 8 concurrent writers, every value matched on readback
- `git diff --check`

Fixes #3478

Related to #3465. This PR fixes the independently reproducible duplicate-bucketing failure; it does not claim to fully resolve the empty-value symptom reported there.

### DIFF
--- a/mooncake-store/src/storage_backend.cpp
+++ b/mooncake-store/src/storage_backend.cpp
@@ -2411,6 +2411,11 @@ tl::expected<void, ErrorCode> BucketStorageBackend::GroupOffloadingKeysByBucket(
     MutexLocker offloading_locker(&offloading_mutex_);
     auto& ungrouped_offloading_objects = ungrouped_offloading_objects_;
     auto it = offloading_objects.cbegin();
+    std::unordered_set<std::string> carryover_keys;
+    carryover_keys.reserve(ungrouped_offloading_objects.size());
+    for (const auto& [key, _] : ungrouped_offloading_objects) {
+        carryover_keys.insert(key);
+    }
     int64_t residue_count = static_cast<int64_t>(
         offloading_objects.size() + ungrouped_offloading_objects.size());
     int64_t total_count = residue_count;
@@ -2437,8 +2442,8 @@ tl::expected<void, ErrorCode> BucketStorageBackend::GroupOffloadingKeysByBucket(
             ungrouped_offloading_objects.clear();
         }
 
-        for (int64_t i = static_cast<int64_t>(bucket_keys.size());
-             i < bucket_backend_config_.bucket_keys_limit; ++i) {
+        while (static_cast<int64_t>(bucket_keys.size()) <
+               bucket_backend_config_.bucket_keys_limit) {
             if (it == offloading_objects.cend()) {
                 for (const auto& bucket_object : bucket_objects) {
                     ungrouped_offloading_objects.emplace(bucket_object.first,
@@ -2448,6 +2453,11 @@ tl::expected<void, ErrorCode> BucketStorageBackend::GroupOffloadingKeysByBucket(
                         << "Total ungrouped count: "
                         << ungrouped_offloading_objects.size();
                 return {};
+            }
+
+            if (carryover_keys.find(it->first) != carryover_keys.end()) {
+                ++it;
+                continue;
             }
 
             if (it->second > bucket_backend_config_.bucket_size_limit) {
--- a/mooncake-store/src/storage_backend.cpp
+++ b/mooncake-store/src/storage_backend.cpp
@@ -2409,15 +2409,15 @@ tl::expected<void, ErrorCode> BucketStorageBackend::GroupOffloadingKeysByBucket(
     const std::unordered_map<std::string, int64_t>& offloading_objects,
     std::vector<std::vector<std::string>>& buckets_keys) {
     MutexLocker offloading_locker(&offloading_mutex_);
-    auto& ungrouped_offloading_objects = ungrouped_offloading_objects_;
-    auto it = offloading_objects.cbegin();
-    std::unordered_set<std::string> carryover_keys;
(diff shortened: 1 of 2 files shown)


982945902:
Additional validation after the no-copy carryover refactor (`49149be0`):

- `FileStorageTest.GroupOffloadingKeysByBucket*`: **5/5 passed**
- full `file_storage_test`: **24/25 passed**; the only failure was `HeartbeatRunsDiskWatermarkEvictionWithoutOffloadWork`, caused by the local container network setup rather than this change:
  - with `--network host`, the in-process HTTP metadata server could not bind its randomly selected port
  - with an isolated Docker network, the HTTP server started, but the test's hard-coded `127.0.0.1` transfer endpoint was rejected because no valid LAN address was available
- final RDMA + io_uring SSD regression on the exact PR head:
  - 2.5 GiB memory segment
  - 10,240 × 1 MiB values (10 GiB total)
  - 8 concurrent writers
  - put: 24.90 s, zero failures
  - get/verify: 211.70 s, zero failures
  - every value matched
  - zero `OBJECT_ALREADY_EXISTS`, offload persistence, empty-value, short-read, or mismatch errors
- `git diff --check` passed

LujhCoconut (approved):
LGTM. 

The fix targets the root cause in #3478 and the regression test is effective. One nit (non-blocking): `residue_count` / `total_count over`-count carryover input duplicates, so `residue_count` never hits 0 — log-only (VLOG at line 2503), no functional impact.

he-yufeng:
Read the whole flow against main's version. The dedup is the right shape: carryover is loaded into the first bucket once, the fill loop skips heartbeat entries already carried, and the residue restore on an underfilled bucket still lands everything back in the member map. The regression test covers the repeat-in-next-heartbeat case directly, and updating the keys-limit test to use fresh keys instead of duplicate carryovers is the honest version of that fixture.

Two observations, neither blocking:

- `residue_count`/`total_count` still count a carried key twice when it also appears in the new heartbeat. They only feed VLOG lines, so the skew is cosmetic, but a note there would save the next reader a recount.
- This overlaps my open #3135 (re-offload of a persisted key as an idempotent skip in BatchOffload). They compose well: this PR stops the duplicate from reaching BatchOffload, and #3135 keeps the one that still slips through (crash between grouping and offload) from failing the batch. Worth landing both.

I did not run the suite here (no store build on this machine), so leaving the formal review to someone who has.

ykwd:
Code format check failed

982945902:
Fixed in `0e490c0c`. I applied the clang-format 20 changes reported by CI in `storage_backend.cpp` and `file_storage_test.cpp`, then reran the same changed-lines check locally with clang-format 20.1.8; it passes.

ykwd (approved):

PR #1677, opened 2026-03-17
[Store] Make segment splitting transport-aware and use balanced partitioning

00fish0:
## Description

Previously, segment splitting was unconditionally applied using `globalConfig().max_mr_size` regardless of transport protocol. This caused two issues:

  1. **Unnecessary splitting for non-RDMA**: TCP and other transports have no hardware MR size limit, but segments were still split at 1TB boundaries, creating multiple allocators under the same segment name. This adds overhead in `allocateSingle()` (random selection + fallback loop instead of direct call) and increases memory allocator fragmentation.

  2. **Unbalanced segment sizes cause skewed load under `RandomAllocationStrategy`**: The greedy approach (`min(remaining, max_mr_size)`) created unevenly sized segments (e.g., 1TB + 0.4TB for a 1.4TB region). Since `RandomAllocationStrategy` picks segments with equal probability regardless of capacity, the smaller segment receives the same request rate but fills up much faster — once full, all allocation pressure falls onto the remaining segment, effectively wasting the parallelism that splitting was meant to provide.

  While `FreeRatioFirstAllocationStrategy` can compensate by sorting on free ratio, `RandomAllocationStrategy` is the default and fastest path, so fixing segment sizes at the source benefits all strategies.

  The existing code already acknowledged part of this with the comment:

```
// Tcp is not limited by max_mr_size, but we ignore it for now.
```

## Changes

- Skip segment splitting for non-RDMA transports (tcp, ascend, etc.) to avoid unnecessary allocator fragmentation. The max_mr_size limit is an RDMA hardware constraint (ibv_reg_mr()), not applicable to other transports.
- For RDMA, use balanced equal-sized partitioning instead of greedy splitting. For example, 1.4TB is now split into 2×0.7TB instead of 1TB + 0.4TB, ensuring more even allocation across allocators.
- Make CheckRegisterMemoryParams protocol-aware so it only enforces max_mr_size validation for RDMA.

## Module

- [ ] Transfer Engine (`mooncake-transfer-engine`)
- [x] Mooncake Store (`mooncake-store`)
- [ ] Mooncake EP (`mooncake-ep`)
- [ ] Integration (`mooncake-integration`)
- [ ] P2P Store (`mooncake-p2p-store`)
- [ ] Python Wheel (`mooncake-wheel`)
- [ ] PyTorch Backend (`mooncake-pg`)
- [ ] Mooncake RL (`mooncake-rl`)
- [ ] CI/CD
- [ ] Docs
- [ ] Other

[...]

PR #3486, opened 2026-08-17
[Store]: Split segments by transport limit

yokinoshitayoki:
## Motivation

Segment splitting was previously applied unconditionally using `globalConfig().max_mr_size`, even for transports such as TCP that do not have an MR size limit. The existing greedy strategy could also produce uneven segments (for example, `1 TB + 0.4 TB`), causing allocation skew with the default `RandomAllocationStrategy`.

Thanks to #1677 for identifying the transport-awareness and allocation-balance issues and for providing the original direction for this improvement. This PR implements the same goal with a smaller change rebased on the latest `main`.

## Changes
- Apply `max_mr_size` only to `efa` and `cxi`, and apply `max_seg_size` to `ub`. RDMA relies on Transfer Engine's internal MR chunking, while TCP and other unconstrained transports use a single Store segment.
- Use best-effort balanced partitioning for constrained transports while preserving page/slab alignment and the minimum required segment count.
- Cover all three Store segment creation paths: setup, file-backed mount, and allocate-and-mount.
- Make memory registration validation protocol-aware and document the updated behavior.

## Module

- [ ] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [x] Documentation update
- [x] Performance improvement
- [ ] Other

## How Has This Been Tested?

- Built with CUDA 13.1 and tested on an NVIDIA A100-SXM4-40GB (`sm80`).
- Passed all 61 `pybind_client_test` tests.
- Verified TCP and RDMA remain a single Store segment when their size exceeds `max_mr_size`.
- Verified a constrained 6-slab region is split evenly into `3 + 3` slabs.
- Verified real RDMA registration and cleanup on `ibp12s0` with `MC_MAX_MR_SIZE=16777216` (16 MiB).
- Built the Sphinx HTML documentation successfully.

**Test results:**
- [x] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex helps me to write code and run tests, but it's reviewed by human.

### DIFF
--- a/docs/source/api-reference/http/http-service.md
+++ b/docs/source/api-reference/http/http-service.md
@@ -280,9 +280,10 @@ mc_store_rest_server --config /path/to/mooncake_config.json --port 8080
 ```
 
 ### `/api/mount_shm`
-Mount a named shared memory object as one or more Mooncake store segments. If
-the requested size exceeds the maximum registration size, the service may split
-the region and return multiple segment ids.
+Mount a named shared memory object as one or more Mooncake store segments.
+Protocols with a registration-size limit split oversized regions and return
+multiple segment ids. Protocols without such a limit, such as TCP, use a single
+segment regardless of `max_mr_size`.
 
 **Method**: `POST`
 **Content-Type**: `application/json`
@@ -366,8 +367,9 @@ curl -X POST http://localhost:8080/api/unmount_shm \
 
 ### `/api/mount`
 Allocate memory inside the store process and mount it as one or more Mooncake
-store segments. If the requested size exceeds the maximum registration size,
-the service may split it and return multiple segment ids. The response includes
+store segments. Protocols with a registration-size limit split oversized
+requests and return multiple segment ids. Protocols without such a limit, such
+as TCP, use a single segment regardless of `max_mr_size`. The response includes
 the actual allocated size after alignment.
 
 **Method**: `POST`
--- a/docs/source/api-reference/http/http-service.md
+++ b/docs/source/api-reference/http/http-service.md
@@ -282,8 +282,8 @@ mc_store_rest_server --config /path/to/mooncake_config.json --port 8080
 ### `/api/mount_shm`
 Mount a named shared memory object as one or more Mooncake store segments.
 Protocols with a registration-size limit split oversized regions and return
-multiple segment ids. Protocols without such a limit, such as TCP, use a single
-segment regardless of `max_mr_size`.
+multiple segment ids. Protocols without such a Store-level limit, such as TCP
+and RDMA, use a single segment regardless of `max_mr_size`.
 
 **Method**: `POST`
 **Content-Type**: `application/json`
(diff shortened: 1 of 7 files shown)


staryxchen:
Thanks for the cleanup. One design note: Transfer Engine already auto-chunks large RDMA registrations (#2644). `RdmaTransport::registerLocalMemory` splits buffers larger than `max_mr_size` into multiple MRs / BufferDescs, so a single Store segment no longer has to stay `<= max_mr_size` for `ibv_reg_mr` correctness.

That means the Store-level split is no longer a workaround for RDMA registration. TE never sees the large buffer on the mount path today, because `CheckRegisterMemoryParams` still rejects `length > max_mr_size` before `registerLocalMemory`.

Skipping the split for TCP is still useful. For RDMA, if we keep splitting, it should be framed as a Store allocator/load-balancing choice (`RandomAllocationStrategy` is per-segment, not capacity-weighted), not as a hardware MR limit. If the goal is only to honor `max_mr_size`, we could relax that check and let TE handle the chunks.

`ub` is different: Kunpeng still shrinks to `max_seg_size` instead of auto-chunking, so a Store-side split is still required there.

ykwd:
@staryxchen Thanks for the clarification! One question about letting TE handle the splitting: suppose a Store segment is larger than the hardware MR limit, so TE splits it into multiple MRs / BufferDescs internally. If a slice allocated by Store happens to cross one of those TE chunk boundaries, would calling the TE read/write APIs on that single slice still work correctly?

staryxchen:
@ykwd Yes. On RDMA, a single TE read/write that straddles a chunk seam is split internally before posting.

`RdmaTransport::submitTransferTask` caps each slice at the remaining bytes of both the source and target `BufferDesc` (`SliceLengthCalculator`), so no WR spans two MRs. Each sub-slice then gets the lkey/rkey of the chunk that actually contains it. The caller still submits one `TransferRequest`.

This is covered by `RDMALargeMrTest.WriteStraddlesChunkBoundary` (destination crosses the seam) and `WriteWithSourceStraddlingChunkBoundary` (source crosses the seam).

Store slices are much smaller than a typical `max_mr_size`, but they can still land on a chunk boundary. That does not need a Store-side split; TE will split the WRs.

One caveat: this submit-time split is RDMA-specific. EFA/CXI also chunk at registration, but they still post one slice per request, so a cross-chunk transfer would fail `selectDevice` there. For RDMA, the Store mount path can leave large segments intact and let TE handle both registration and the seam.

yokinoshitayoki:
@staryxchen Thanks for the clarification. For this PR, I’ll remove RDMA from GetTransportRegistrationLimit() while keeping EFA/CXI/UB unchanged. Longer term, Maybe EFA/CXI should support cross-BufferDesc slicing like RDMA, after which Store would only need to split for UB.

staryxchen:
@yokinoshitayoki That plan sounds good: drop RDMA from `GetTransportRegistrationLimit()` in this PR, and keep the Store-side split for EFA/CXI/UB.

I do not have EFA or CXI hardware to verify submit-time cross-`BufferDesc` slicing the way RDMA does, so I will not try to add that here. I will open a follow-up issue to track adding RDMA-style cross-`BufferDesc` slicing for EFA/CXI; after that, Store should only need to split for UB.

staryxchen:
Opened #3500 to track RDMA-parity large-buffer support for EFA/CXI/UB, with sub-issues #3501 (EFA), #3502 (CXI), and #3503 (UB).

yokinoshitayoki:
@staryxchen Thanks a lot. I have updated the code, could you have a look again?

staryxchen (approved):
Looks good — this matches what we discussed. A nit: adding a dynamic-mount test that RDMA does not split is better.

Aionw (approved):
LGTM

yokinoshitayoki:
@staryxchen Thanks for your suggestion, I've added a dynamic-mount test.

ykwd (approved):
LGTM

PR #3488, opened 2026-08-17
perf(store): batch io_uring bucket reads

### LINKED ISSUE #3487: perf(store): batch io_uring reads in BucketStorageBackend
## Problem

`BucketStorageBackend::BatchLoad()` groups requested objects by bucket, but the io_uring path then calls `read_aligned()` once per key. Each call submits its SQEs and immediately waits for completion before the next key is processed.

This keeps cross-key queue depth at one even though `UringFile::batch_read()` already exists for submitting up to 32 independent reads together. The existing batch API only returns aggregate bytes, so it cannot safely identify per-key CQE failures or short reads.

## Proposed behavior

- Submit the aligned reads for one bucket through `UringFile::batch_read()`.
- Associate every CQE with its own descriptor, independent of completion order.
- Drain all CQEs in the submitted batch before returning an error.
- Report `completed`, `bytes_read`, and `error` per descriptor.
- Reject a bucket load if any descriptor fails or does not cover the requested value range.
- Preserve the synchronous `BatchLoad()` API; only device-side execution becomes batched.

## SSD benchmark

Development-machine SSD, ext4, O_DIRECT, registered buffers, one thread, queue depth 32. Sequential and batch modes use the same file, buffers, random offsets, and alternate execution order.

| Request size | Sequential | Batched | Speedup |
| --- | ---: | ---: | ---: |
| 4 KiB | 14,433 IOPS / 56.38 MiB/s | 198,277 IOPS / 774.52 MiB/s | 13.74x |
| 64 KiB | 7,586 IOPS / 474.16 MiB/s | 107,861 IOPS / 6,741.33 MiB/s | 14.22x |
| 1 MiB | 3,377 IOPS / 3,376.94 MiB/s | 18,695 IOPS / 18,694.75 MiB/s | 5.54x |

The benchmark is intended for relative comparison on the same machine; absolute throughput depends on the SSD/controller/filesystem.

## Validation needed

- More than 32 requests to cover multiple ring submissions.
- Per-request short-read reporting.
- CQE error draining followed by a successful operation on the same thread-local ring.
- Bucket-level O_DIRECT integration with aligned destination buffers.
- Build compatibility with and without liburing.

982945902:
## Summary

- batch aligned reads for all requested objects in the same bucket instead of submitting and waiting once per key
- return per-request `completed`, `bytes_read`, and `error` state from `UringFile::batch_read()`
- tag CQEs with their descriptor index so out-of-order completions are mapped correctly
- drain every CQE in a submitted batch before returning a per-request failure
- preserve bucket-level short-read validation and the synchronous `BatchLoad()` API
- add a reproducible sequential-vs-batched O_DIRECT benchmark

Fixes #3487

## Error handling

`batch_read()` resets all descriptors before submission. Each submitted CQE marks exactly one descriptor completed and records either its byte count or `FILE_READ_FAIL`. If one CQE fails, the implementation still collects the rest of that submitted batch before returning an error. Descriptors in later, unsubmitted batches remain `completed=false`.

`BucketStorageBackend::BatchLoad()` rejects:

- submission/CQE failures
- descriptors that did not complete
- short reads that do not cover `offset_in_buffer + requested_value_size`

## Tests

With liburing enabled:

- `posix_file_test`: 10/10 passed
  - more than queue depth (40 requests)
  - per-request short read
  - CQE failures drained before reusing the same thread-local ring
- `storage_backend_test`: 92/92 passed
  - 40 objects loaded from one bucket through O_DIRECT/io_uring
  - truncated bucket file rejected as `FILE_READ_FAIL`

Without liburing:

- configured and built `uring_batch_read_bench`, `posix_file_test`, and `storage_backend_test`

## SSD benchmark

Development-machine SSD, ext4, O_DIRECT, registered buffers, one thread, queue depth 32. Sequential and batch modes use the same file, buffers, random offsets, and alternate execution order.

| Request size | Sequential | Batched | Speedup |
| --- | ---: | ---: | ---: |
| 4 KiB | 14,433 IOPS / 56.38 MiB/s | 198,277 IOPS / 774.52 MiB/s | 13.74x |
| 64 KiB | 7,586 IOPS / 474.16 MiB/s | 107,861 IOPS / 6,741.33 MiB/s | 14.22x |
| 1 MiB | 3,377 IOPS / 3,376.94 MiB/s | 18,695 IOPS / 18,694.75 MiB/s | 5.54x |

Absolute throughput depends on the SSD/controller/filesystem; the comparison is intended to measure the relative cost of per-key submit-and-wait versus batched submission on the same machine.


he-yufeng:
Read the tagging scheme end to end. The mapping is sound today: batches are capped at QUEUE_DEPTH (32), so `i + 1` always fits the low 8 bits, single-op CQEs carry a zero low byte and can never collide with a batched index, and collect_batch's mask/dispatch/advance loop covers out-of-order completion, per-desc errors, and invalid indexes.

One implicit invariant worth pinning while it's fresh: the scheme silently breaks if QUEUE_DEPTH ever goes past 255, because `i + 1` would then overflow BATCH_INDEX_MASK into the op tag bits and attribute a completion to the next operation. A `static_assert(QUEUE_DEPTH <= (1u << BATCH_INDEX_BITS))` next to next_operation_tag() keeps that from ever shipping. Not a blocker.

I can't build or run io_uring here (macOS), so this is a read-only review; leaving the verdict to someone with a Linux box.

LujhCoconut:
LGTM.

Two latent `main` bugs fixed as a side effect, worth noting:
1. `collect()` returned on a negative `io_uring_submit_and_wait()` without draining, so prepped-but-unsubmitted SQEs would get submitted by the next operation — writes would fire at the wrong time. The `drain_submitted()` + `reset_ring()` path closes this.
2. `read`/`write`/`vector_read`/`vector_write` didn't check `initialized_`, so a failed `io_uring_queue_init()` left callers operating on an uninitialized ring.

Optional, non-blocking:
- `batch_read` lacks the alignment check that `read_aligned` has, so an unaligned caller buffer reports `FILE_READ_FAIL` instead of the clearer `FILE_INVALID_BUFFER`.
- The `ReadDesc` cast is guarded only by `sizeof` equality; `len`/`off` are both 8 bytes, so swapping them wouldn't trip the assert. Prefer one shared definition or `offsetof` assertions.

982945902:
Thanks for the review! I addressed both optional suggestions in commit `93476a82`:

- `batch_read()` now validates the buffer address, length, and offset for O_DIRECT alignment before submitting any SQEs, and returns `FILE_INVALID_BUFFER` for invalid descriptors.
- `SharedUringRing` now directly aliases `UringFile::ReadDesc`, removing the duplicate definition, `sizeof` assertion, and `reinterpret_cast`.

I also added coverage for misaligned buffer, length, and offset cases. The full `posix_file_test` suite passes locally (14/14); CI has been triggered for the updated commit.

LujhCoconut (approved):
LGTM. Thanks for the contribution.

ykwd (approved):

ykwd:
Let's wait the CI tests

PR #3490, opened 2026-08-17
[TENT] Resolve transport policy device masks once, not per request

SongOf:
Description

  TransportSelector::select() re-resolved every SelectionPolicy device name into a NIC mask on each call. Topology::getNicId is a linear scan comparing strings, so every request paid policy_devices × nic_count comparisons, and every failover retry paid it again. Both inputs are fixed once setTopology() runs — the topology is loaded before it and TENT's Topology has no runtime mutator — so resolve there and let select() read a cached mask.

  Resolving per request also fired its diagnostics per request: one mistyped device name emitted a WARNING per transfer. The messages were not actionable either:

  - "RDMA device not found or ID >= 64" conflated a typo with a NIC past the 64-bit mask width; now reported separately.
  - When no name resolved, the mask fell back to ~0ULL, silently turning a device-isolation policy into no isolation. Fallback kept (a typo must not stop transfers), now a single ERROR naming the policy and every unresolved device.
  - setTopology(nullptr) left masks resolved against a topology no longer held. Resolution is now unconditional.

  Also fixes a use-after-move: the "Loaded transport policy" line read policy.name / policy.transports.size() after push_back(std::move(policy)), so transports_count always logged 0.

  Adds a warning when a policy sets service_level/traffic_class. Link-layer QoS belongs to the QP a transfer lands on, so it is configured per pool under transports/rdma/endpoint/qp_pools and applied in setupOneQP(); the SelectionPolicy fields are a superseded duplicate that nothing reads — operators setting them got neither the QoS nor any indication of it. Three stale comments corrected accordingly.

  No behavior change for correctly configured policies.

  Module

  - [ ] Reshard (mooncake-reshard)
  - [ ] CI/CD
  - [ ] Docs
  - [ ] Other

  Type of Change

  - [x] Bug fix
  - [ ] New feature
  - [ ] Refactor
  - [ ] Breaking change
  - [ ] Documentation update
  - [x] Performance improvement
  - [ ] Other

  How Has This Been Tested?

  7 new cases in transport_selector_test.cpp covering mask resolution, stability across repeated select() calls, re-resolution on a replaced topology, fail-open when nothing resolves, partial resolution, NICs past the mask width, and the no-topology case. DeviceMaskIsStableAcrossSelects is the regression guard for per-request resolution.

  Test commands:
  cmake --build build-tent --target transport_selector_test
  ctest -R transport_selector_test --output-on-failure

  Test results:
  - [ ] Unit tests pass

  Checklist

  AI Assistance Disclosure

  - [ ] No AI tools were used

### DIFF
--- a/mooncake-transfer-engine/tent/include/tent/runtime/transport_selector.h
+++ b/mooncake-transfer-engine/tent/include/tent/runtime/transport_selector.h
@@ -120,17 +120,28 @@ struct SelectionPolicy {
     // Transport preference list (evaluated in order)
     std::vector<TransportType> transports;
 
-    // Per-policy link-layer QoS. nullopt = fall back to the global RdmaParams
-    // value. InfiniBand Service Level (0-15) and Traffic Class / DSCP (0-255).
+    // Superseded, and read by nothing: link-layer QoS is a property of the QP
+    // a transfer lands on, so it is defined per pool in
+    // transports/rdma/endpoint/qp_pools and applied at QP setup
+    // (endpoint.cpp, setupOneQP). Still parsed so a config that sets them here
+    // is reported rather than silently ignored.
     std::optional<int> service_level;
     std::optional<int> traffic_class;
-    // Named QP pool this policy's traffic should land on; parsed and stored for
-    // now, routing to be wired later. Unset = the current single "data QP".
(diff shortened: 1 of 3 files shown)


staryxchen (approved):
Good Job! LGTM.

PR #3492, opened 2026-08-18
[Bugfix][Store] Retry HA stale-handle cleanup oplog persist on transient backpressure

zchuango:
## Description

In HA mode (`enable_ha_ && enable_oplog_`), `ClearInvalidHandles` persists
stale-handle cleanup through the oplog writer before removing metadata.
Today a single transient failure makes the persist call give up
immediately:

- `TASK_PENDING_LIMIT_EXCEEDED` — the batch writer's slots are momentarily
  full while it seals the current batch; capacity frees up within
  milliseconds.
- `UNAVAILABLE_IN_CURRENT_STATUS` — the writer is briefly rejecting writes
  while its `write_batch` retries against the KV backend.

When that happens, the stale handles / last-replica metadata survive until
the next cleanup sweep, so invalid client handles linger longer than
necessary and (for the last-replica path) removed objects stay visible in
the metadata map despite the client being gone.

This PR wraps both oplog persist calls in `ClearInvalidHandles` with a
bounded exponential-backoff retry (`RetryOplogPersist`):

- `TASK_PENDING_LIMIT_EXCEEDED`: retry up to 10 attempts, backoff
  `min(2^attempt, 16ms)` — the writer frees slots imminently, waiting pays
  off.
- `UNAVAILABLE_IN_CURRENT_STATUS`: retry up to 5 attempts with the same
  backoff — recovery depends on the KV backend, so bail out earlier
  instead of spinning on a persistent outage.
- Any other error (INVALID_PARAMS etc.): no retry, returned immediately.

**Control flow is unchanged**: if the persist still fails after retries,
the key is skipped exactly as before and cleanup is retried on the next
sweep. No fallback semantics are added or removed.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
# Not run locally: the submitter's host is macOS, and this project can
# only be built inside a Linux container. Intended command:
ctest --test-dir build -R master_service
```

**Test results:**
- [ ] Unit tests pass — not verified locally; relying on this PR's CI run

The change is a pure retry wrapper around existing calls: the success path
is byte-identical to before, and the failure path after exhausting retries
returns the same error the caller already handles (skip key, retry next
sweep). Formatting verified with `clang-format` 20.1.8 per
`.clang-format` (clean, no violations on touched lines).

## Checklist

      (clang-format 20.1.8, no violations on touched lines)
- [ ] I have run pre-commit on the files changed in this PR and all hooks
      pass — toolchain unavailable on the submitter's macOS host; relying
      on CI
- [ ] I have updated the documentation (if applicable) — not applicable,
      no behavior contract change
- [ ] I have added tests to prove my changes are effective — retry timing
      depends on the oplog writer's internal backpressure state; open to
      adding a unit test if reviewers suggest a good injection point
- [ ] For changes >500 LOC: I have filed an RFC issue — not applicable
      (+59/-4 LOC)

## AI Assistance Disclosure

- [ ] No AI tools were used

Ported and adapted with AI assistance (Proma Agent): locating the upstream
call sites, adapting the retry helper to current `main` (the original
commits target a fork whose `ClearInvalidHandles` control flow has
diverged), and formatting checks. The human submitter has reviewed every

### DIFF
--- a/mooncake-store/src/master_service.cpp
+++ b/mooncake-store/src/master_service.cpp
@@ -1981,6 +1981,59 @@ tl::expected<void, ErrorCode> MasterService::PersistStaleHandleCleanupForHA(
     return {};
 }
 
+namespace {
+
+constexpr int kOplogRetryMaxAttempts = 10;
+constexpr int kOplogRetryMaxAttemptsUnavailable = 5;
+constexpr int kOplogRetryMaxDelayMs = 16;
+
+// RetryOplogPersist retries an oplog persist operation on transient
+// backpressure with bounded exponential backoff, so that a momentary
+// hiccup does not defer stale-handle cleanup to the next sweep.
+// Persistent or non-backpressure failures keep the existing behavior:
+// the caller skips the key and retries on the next cleanup round.
+template <typename F>
+auto RetryOplogPersist(F&& persist_fn) -> decltype(std::declval<F>()()) {
+    for (int attempt = 0; attempt <= kOplogRetryMaxAttempts; ++attempt) {
+        auto result = persist_fn();
+        if (result) {
+            return result;
+        }
+        const ErrorCode err = result.error();
+
+        if (err == ErrorCode::TASK_PENDING_LIMIT_EXCEEDED) {
+            // Slots full: writer is sealing the current batch and will
+            // free capacity imminently. Worth waiting.
+            if (attempt == kOplogRetryMaxAttempts) {
+                return result;
+            }
+            std::this_thread::sleep_for(std::chrono::milliseconds(
+                std::min(1 << attempt, kOplogRetryMaxDelayMs)));
+            continue;
+        }
+
+        if (err == ErrorCode::UNAVAILABLE_IN_CURRENT_STATUS) {
+            // Writer not accepting: write_batch is retrying against the
+            // KV backend. Recovery depends on the backend; bail out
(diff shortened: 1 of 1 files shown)


he-yufeng:
Read the helper against its call sites and the thread it runs on. The shape is right: ClearInvalidHandles runs on replica_cleanup_worker_, so the bounded backoff sleeps a background sweeper, not a request path; the asymmetric caps (11 attempts for slot-full backpressure vs 6 for writer-not-accepting) match how each state recovers; non-backpressure errors keep the old give-up-and-resweep behavior.

One gap worth closing before merge: RetryOplogPersist is a template over an arbitrary callable, so it is unit-testable with a fake persist_fn and no store at all, but the PR has no test for it. A three-case test (fail N times with TASK_PENDING_LIMIT_EXCEEDED then succeed; UNAVAILABLE_IN_CURRENT_STATUS bails after the 6th call; INVALID_PARAMS returns on the first failure) would pin the two different caps so a later "simplify" doesn't silently merge them.

Not approving: this is the HA oplog persistence path, and I did not compile it here (no store build on this machine).

PR #3494, opened 2026-08-18
[Bugfix][TENT] Unpin remote stage buffers after late completion

RuiqingFeng:
## Description
When a remote Delegate request outlived `ProxyManager`, destruction discarded the pending cleanup record. A later terminal callback updated only the detached operation, leaving the remote stage buffer pinned permanently.
This change transfers cleanup ownership to unresolved remote operations during `ProxyManager` destruction. A terminal callback then schedules one asynchronous Unpin RPC. Remote buffers not owned by a pending operation are still released immediately.
The first commit adds a failing loopback regression test. The second commit contains only the cleanup fix.

## Module
- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?
The regression test delays a real loopback Delegate RPC, destroys `ProxyManager`, releases the RPC, and verifies exactly one Unpin followed by a successful repin of the same buffer.
**Test commands:**
```bash
cmake --build build-tent --target tent_progress_worker_test
build-tent/mooncake-transfer-engine/tent/tests/tent_progress_worker_test \
  --gtest_filter='ProxyManager.LateRemoteCompletionReleasesAbandonedStageBuffer'
```
**Test results:**
- [x] Unit tests pass
  - Verified regression test `ProxyManager.LateRemoteCompletionReleasesAbandonedStageBuffer` fails before fix and passes after fix.

## Checklist

## AI Assistance Disclosure
- [ ] No AI tools were used

### DIFF
--- a/mooncake-transfer-engine/tent/src/runtime/remote_stage_operation.cpp
+++ b/mooncake-transfer-engine/tent/src/runtime/remote_stage_operation.cpp
@@ -32,17 +32,17 @@ RemoteStageOperation::RemoteStageOperation(std::string server_addr,
       remote_buffer_(remote_buffer),
       deferred_cleanup_(std::move(deferred_cleanup)) {}
 
-void RemoteStageOperation::complete(Status status) {
+void RemoteStageOperation::complete(Status status, bool confirmed) {
     DeferredCleanup cleanup;
     {
         std::lock_guard<std::mutex> lock(mutex_);
-        result_ = std::move(status);
+        result_ = RemoteStageResult{std::move(status), confirmed};
         cleanup = takeCleanupIfReadyLocked();
     }
     runDeferredCleanup(std::move(cleanup));
 }
 
-std::optional<Status> RemoteStageOperation::tryTakeResult() {
+std::optional<RemoteStageResult> RemoteStageOperation::tryTakeResult() {
     std::lock_guard<std::mutex> lock(mutex_);
     if (abandoned_ || !result_) return std::nullopt;
     auto result = std::move(result_);
@@ -67,7 +67,8 @@ bool RemoteStageOperation::ownsRemoteBuffer(const std::string& server_addr,
 
 RemoteStageOperation::DeferredCleanup
 RemoteStageOperation::takeCleanupIfReadyLocked() {
-    if (!abandoned_ || !result_ || cleanup_started_ || !deferred_cleanup_) {
+    if (!abandoned_ || !result_ || !result_->confirmed || cleanup_started_ ||
+        !deferred_cleanup_) {
         return {};
     }
     cleanup_started_ = true;
@@ -86,9 +87,15 @@ bool pollRemoteOperations(
             ++it;
             continue;
         }
-        if (!result->ok()) {
+        if (!result->confirmed) {
+            LOG(WARNING) << "Remote staging completion is unconfirmed: "
+                         << result->status;
+            ++it;
+            continue;
+        }
+        if (!result->status.ok()) {
             LOG(WARNING) << "Failed to drain remote staging request: "
-                         << *result;
+                         << result->status;
         }
         it = remote_operations.erase(it);
     }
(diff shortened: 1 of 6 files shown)


he-yufeng:
Read the full ownership transfer. The shape is right: `takeCleanupIfReadyLocked` fires exactly once (abandoned + result + not started), the cleanup runs outside the mutex, and buffers with no pending owner still get the immediate release. The failing-test-first split made the leak easy to verify.

One thing I could not settle from the diff: the deferred unpin goes through `tl_rpc_agent.callAsync` at ProxyManager destruction time. If the RPC agent tears down before or during ProxyManager::deconstruct (static destruction order at process exit, or a crash path), does callAsync fail cleanly into runDeferredCleanup's warning, or can it throw/crash on a dead channel? If the latter, the deferred path needs to check agent liveness before calling in. The in-codebase answer might be obvious to you; asking because everything else about the transfer reads correct.

RuiqingFeng:
> Read the full ownership transfer. The shape is right: `takeCleanupIfReadyLocked` fires exactly once (abandoned + result + not started), the cleanup runs outside the mutex, and buffers with no pending owner still get the immediate release. The failing-test-first split made the leak easy to verify.
> 
> One thing I could not settle from the diff: the deferred unpin goes through `tl_rpc_agent.callAsync` at ProxyManager destruction time. If the RPC agent tears down before or during ProxyManager::deconstruct (static destruction order at process exit, or a crash path), does callAsync fail cleanly into runDeferredCleanup's warning, or can it throw/crash on a dead channel? If the latter, the deferred path needs to check agent liveness before calling in. The in-codebase answer might be obvious to you; asking because everything else about the transfer reads correct.

Hi @he-yufeng, really appreciate you reviewing the full ownership transfer. During normal `ProxyManager::deconstruct()` , `tl_rpc_agent` is still alive. Synchronous exceptions from `callAsync()` are caught by `runDeferredCleanup()`, while asynchronous connection or call errors are logged via the Unpin callback.
The process-level TLS/RPC runtime destruction scenario you pointed out is indeed not covered. Will keep the current PR focused on late-completion cleanup and handle the runtime lifecycle and shutdown order in a follow-up PR.

mooncake-transfer-engine/tent/src/runtime/remote_stage_operation.cpp
```diff
@@ -1,26 +1,79 @@
...
-    std::lock_guard<std::mutex> lock(mutex_);
-    result_ = std::move(status);
+    DeferredCleanup cleanup;
+    {
+        std::lock_guard<std::mutex> lock(mutex_);
+        result_ = std::move(status);
+        cleanup = takeCleanupIfReadyLocked();
+    }
+    runDeferredCleanup(std::move(cleanup));
 }
```
alogfans: An RPC error does not prove that the remote Delegate handler has stopped. For example, the connection may fail while the offloaded server handler is still accessing the stage buffer. Treating every callback result as terminal can therefore unpin and reuse the buffer prematurely. Please distinguish a confirmed server response from an unknown transport outcome, and defer cleanup until remote completion is explicitly confirmed.

RuiqingFeng: Thanks for catching this!  Updated the fix to ensure `Unpin` is deferred until remote completion is explicitly confirmed:

- `confirmed = true`: Set when an explicit server response is received (whether success or server-side error), confirming that the remote handler has stopped accessing the buffer.
- `confirmed = false`: Set on transport/network failures (e.g., timeout, connection drop). The task fails, but the remote buffer remains pinned to prevent premature reuse.

Also added a regression test for unconfirmed remote completion and verified all tests pass. Thank you!

alogfans (approved):
LGTM

PR #3497, opened 2026-08-18
[Store] Gate HA serving on successful standby restore

Icedcoco:
## Description

Gate HA serving on successful standby restoration and release leadership on all promotion-phase failures.

An elected candidate is allowed to serve only after:

1. standby final catch-up and promotion/export succeed;
2. the complete `PromotionContext`, including a valid empty context, is restored and validated;
3. leadership is revalidated immediately before serving.

If promotion or restoration fails, the candidate remains unavailable, its service delegate is not exposed, and leadership is released. A leadership-release failure does not make the candidate serviceable.

The previous behavior logged restoration errors and continued serving from empty or partially restored metadata. This was a correctness bug rather than a supported availability fallback: the in-memory state could disagree with the durable OpLog and make later recovery unsafe. This PR intentionally changes that observable failure behavior to fail closed.

Mooncake does not automatically discard recovery data after an error. The deployment guide documents the explicit recovery path for deployments that accept cache loss: stop all processes using the old `cluster_id` and start a new empty cluster with a previously unused `cluster_id`.

This branch is rebased onto current `main` and includes #3566. The combined HA-without-OpLog path is covered end to end in the real supervisor:

`CreateStandbyController()` -> `NoopStandbyController` -> empty `PromotionContext` -> restore gate -> `service_ready=true`.

The test starts a real `MasterServiceSupervisor` in a child process and observes `/health` and `/get_all_keys`; production code has no test-only injection path or public testing API.

## Module

- [x] Docs

## Type of Change

- [x] Bug fix
- [x] Breaking change
- [x] Documentation update

## How Has This Been Tested?

**Build and test commands:**

```bash
cmake -S . -B build-r02-latest \
  -DSTORE_USE_ETCD=ON \

cmake --build build-r02-latest \
  --target high_availability_test \
           master_service_ha_test \
           hot_standby_snapshot_bootstrap_test \
  -j32

LSAN_OPTIONS=detect_leaks=0 \
  build-r02-latest/mooncake-store/tests/high_availability_test \
  --etcd_endpoints=127.0.0.1:22379

LSAN_OPTIONS=detect_leaks=0 \
  build-r02-latest/mooncake-store/tests/hot_standby_snapshot_bootstrap_test

LSAN_OPTIONS=detect_leaks=0 \
  build-r02-latest/mooncake-store/tests/master_service_ha_test \
  --gtest_filter='MasterServiceHATest.RestoreRejectsDescriptorSizeMismatch:MasterServiceHATest.RestoreRejectsOverlappingMemoryDescriptors:MasterServiceHATest.RestoreRejectsDfsMode'

pre-commit run --files \
  docs/source/design/store/mooncake-store.md \
  docs/source/deployment/mooncake-store-deployment-guide.md \
  mooncake-store/src/ha/leadership/master_service_supervisor.cpp \
  mooncake-store/tests/ha/leadership/high_availability_test.cpp
```

**Test results:**

- [x] Fresh current-`main` build succeeds.
- [x] Full `high_availability_test` passes (14/14) against isolated local etcd.
- [x] Full `hot_standby_snapshot_bootstrap_test` passes (7/7).
- [x] Focused restore rejection tests pass (3/3).
- [x] Pre-commit checks pass.

## Checklist

- [x] I have formatted the changed C++ code using pre-commit.
- [x] I have run pre-commit on the files changed in this PR and all hooks pass.
- [x] I have updated the HA design and deployment documentation.

## AI Assistance Disclosure

- [x] AI tools were used.

AI assistance was used for code inspection, implementation, test construction, documentation, rebase validation, and verification. The human submitter must review and understand every changed line.

### DIFF
--- a/mooncake-store/src/ha/leadership/master_service_supervisor.cpp
+++ b/mooncake-store/src/ha/leadership/master_service_supervisor.cpp
@@ -215,17 +215,13 @@ void EnterStandbyMode(MasterAdminServer& admin_server,
     SetRuntimeState(admin_server, standby_controller.GetStandbyRuntimeState());
 }
 
-int RunSupervisorLoop(
-    const HABackendSpec& spec, const MasterServiceSupervisorConfig& config,
-    MasterAdminServer& admin_server,
-    std::unique_ptr<LeaderCoordinator> coordinator_override = nullptr,
-    std::unique_ptr<StandbyController> standby_controller = nullptr) {
+int RunSupervisorLoop(const HABackendSpec& spec,
+                      const MasterServiceSupervisorConfig& config,
+                      MasterAdminServer& admin_server) {
     auto label_reconciler = MakeLeaderLabelReconciler(config);
     label_reconciler.SetLeader(false);
     SetRuntimeState(admin_server, MasterRuntimeState::kStarting);
-    if (!standby_controller) {
-        standby_controller = CreateStandbyController(spec, config);
-    }
+    auto standby_controller = CreateStandbyController(spec, config);
     std::atomic<bool> accept_standby_runtime_updates{false};
     standby_controller->SetStandbyRuntimeStateCallback(
         [&](MasterRuntimeState state) {
@@ -240,10 +236,7 @@ int RunSupervisorLoop(
                      accept_standby_runtime_updates, std::nullopt);
 
     while (true) {
-        auto coordinator = coordinator_override
-                               ? decltype(CreateLeaderCoordinator(spec))(
-                                     std::move(coordinator_override))
-                               : CreateLeaderCoordinator(spec);
+        auto coordinator = CreateLeaderCoordinator(spec);
         if (!coordinator) {
             if (HandleSupervisorError("create leader coordinator",
                                       coordinator.error(), spec.type)) {
@@ -525,15 +518,6 @@ int RunSupervisorLoop(
 }
 
 }  // namespace
-
-int RunSupervisorLoopForTesting(
-    const HABackendSpec& spec, const MasterServiceSupervisorConfig& config,
-    MasterAdminServer& admin_server,
-    std::unique_ptr<LeaderCoordinator> coordinator,
-    std::unique_ptr<StandbyController> standby_controller) {
-    return RunSupervisorLoop(spec, config, admin_server, std::move(coordinator),
-                             std::move(standby_controller));
-}
 
 MasterServiceSupervisor::MasterServiceSupervisor(
     const MasterServiceSupervisorConfig& config)
--- a/mooncake-store/src/ha/leadership/master_service_supervisor.cpp
+++ b/mooncake-store/src/ha/leadership/master_service_supervisor.cpp
@@ -390,18 +390,30 @@ int RunSupervisorLoop(const HABackendSpec& spec,
             wrapped_config, config.http_metadata_server,
             config.http_metadata_remote_url);
 
-        // Restore from standby if we have context
-        if (promotion_ctx->applied_seq_id > 0 ||
-            !promotion_ctx->objects.empty() ||
-            !promotion_ctx->segments.empty()) {
-            auto restore_result = wrapped_master_service->RestoreFromStandby(
-                promotion_ctx->objects, promotion_ctx->applied_seq_id,
-                promotion_ctx->segments);
-            if (!restore_result) {
-                LOG(ERROR) << "Standby restore failed: "
-                           << toString(restore_result.error());
+        // Restore is the serving gate: do not register or expose a candidate
+        // service until the complete promotion context has been applied.
+        SetRuntimeState(admin_server, MasterRuntimeState::kRecovering);
+        auto restore_result = wrapped_master_service->RestoreFromStandby(
+            promotion_ctx->objects, promotion_ctx->applied_seq_id,
+            promotion_ctx->segments);
+        if (!restore_result) {
+            LOG(ERROR) << "Standby restore failed: "
+                       << toString(restore_result.error());
+            wrapped_master_service.reset();
+            DeactivateServingState(admin_server, label_reconciler);
+            EnterStandbyMode(admin_server, *standby_controller,
+                             accept_standby_runtime_updates,
+                             leadership_session->view);
+            SetRuntimeState(admin_server, MasterRuntimeState::kRecovering);
+            if (HandleLeadershipPhaseError(
+                    "standby restore failure", "restore standby state",
+                    leader_coordinator, *leadership_session,
+                    restore_result.error(), spec.type)) {
+                return -1;
             }
+            continue;
         }
+        SetRuntimeState(admin_server, MasterRuntimeState::kLeaderWarmup);
 
         mooncake::RegisterRpcService(server, *wrapped_master_service);
 
--- a/mooncake-store/src/ha/leadership/master_service_supervisor.cpp
+++ b/mooncake-store/src/ha/leadership/master_service_supervisor.cpp
@@ -215,13 +215,17 @@ void EnterStandbyMode(MasterAdminServer& admin_server,
     SetRuntimeState(admin_server, standby_controller.GetStandbyRuntimeState());
 }
 
-int RunSupervisorLoop(const HABackendSpec& spec,
-                      const MasterServiceSupervisorConfig& config,
-                      MasterAdminServer& admin_server) {
+int RunSupervisorLoop(
+    const HABackendSpec& spec, const MasterServiceSupervisorConfig& config,
+    MasterAdminServer& admin_server,
+    std::unique_ptr<LeaderCoordinator> coordinator_override = nullptr,
+    std::unique_ptr<StandbyController> standby_controller = nullptr) {
     auto label_reconciler = MakeLeaderLabelReconciler(config);
     label_reconciler.SetLeader(false);
     SetRuntimeState(admin_server, MasterRuntimeState::kStarting);
-    auto standby_controller = CreateStandbyController(spec, config);
+    if (!standby_controller) {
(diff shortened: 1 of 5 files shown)


mooncake-store/src/ha/leadership/master_service_supervisor.cpp
```diff
@@ -390,18 +399,30 @@ int RunSupervisorLoop(const HABackendSpec& spec,
...
-        if (promotion_ctx->applied_seq_id > 0 ||
-            !promotion_ctx->objects.empty() ||
-            !promotion_ctx->segments.empty()) {
-            auto restore_result = wrapped_master_service->RestoreFromStandby(
-                promotion_ctx->objects, promotion_ctx->applied_seq_id,
-                promotion_ctx->segments);
-            if (!restore_result) {
-                LOG(ERROR) << "Standby restore failed: "
-                           << toString(restore_result.error());
+        // Restore is the serving gate: do not register or expose a candidate
```
Aionw, line 395: I don't think a successful standby restore should be a hard prerequisite for serving in this case.

Snapshot or OpLog corruption is shared recovery state. If every candidate reads the same corrupted state, each candidate will fail restoration and release leadership, potentially leaving the cluster permanently without a live master. For a KV-cache workload, rebuilding metadata from an empty state is preferable to cluster-wide unavailability.

Could we distinguish between:

- transient backend errors or leadership loss: fail closed and retry/release leadership;
- confirmed unrecoverable recovery-data corruption: retain valid leadership, atomically advance to a new HA generation, cold-start with empty metadata/OpLog state, and let standbys follow the new generation.

Simply ignoring the restore error is not sufficient because the current OpLog keys reuse `/oplog/<cluster>/...`; the corrupted durable prefix or batches would continue poisoning the new primary and future standbys. The cold-start path needs to fence off the old generation.

Also, calling `RestoreFromStandby()` for an empty promotion context introduces unnecessary coupling: some storage modes, such as DFS, reject restoration even when the context is empty.

Until there is a fenced cold-rebuild path, making restore success a serving gate seems to trade recoverable cache loss for indefinite cluster unavailability.

Icedcoco: Thanks for raising this. I agree that shared recovery-state corruption can leave every candidate unable to serve, and that an operator may prefer an explicit cold rebuild for a KV-cache workload.

However, I do not think R02 should implicitly fall back to empty state. The current OpLog namespace has no recovery generation: batches, `durable_prefix`, and `producer_view` all remain under `/oplog/<cluster>/...`. Serving from empty state after a restore failure would therefore leave the new primary inconsistent with durable history and would continue poisoning future standbys. The current error codes also cannot safely distinguish confirmed corruption from configuration, implementation, or transient backend errors.

I will keep R02 fail-closed as designed. A fenced cold-rebuild path should be handled separately, with an explicit operator decision, a new generation or cluster namespace, and fencing of old writers and standbys. Mooncake should not decide by itself that recovery data may be discarded.


Regarding the empty promotion context: the unconditional restore is intentional. An empty context means that no metadata was exported; it does not prove that the configured storage mode supports HA restoration. Calling `RestoreFromStandby()` also acts as the capability gate. DFS allocator state is currently not restorable, and DFS is already documented and checked as incompatible with snapshot, OpLog, and standby recovery. The noop controller also does not return a successful empty promotion context. Skipping restore for an empty context would therefore allow an unsupported recovery mode to bypass this gate.

So I propose keeping the current R02 behavior and treating operator-controlled, generation-fenced cold rebuild as a separate design and PR.

Aionw: Thanks for the clarification. I think the core concern is compatibility with the existing HA semantics, rather than whether R02 itself should implement an automatic cold rebuild.

Before R02, leadership acquisition and state restoration were not coupled as hard serving prerequisites: an empty promotion context skipped restoration, and a restoration failure was logged without preventing the elected master from serving. R02 changes that operational contract by making successful restoration mandatory.

As a result, the same configuration and recovery failure that previously produced a live master with rebuilt/empty cache state can now leave the cluster indefinitely without any serving master. Deferring a generation-fenced cold-rebuild mechanism to a separate PR does not address this compatibility break, because R02 removes the existing availability behavior before a replacement recovery path exists.

If this is an intentional semantic change, I think it should be treated explicitly as a breaking HA design change, with corresponding configuration validation, documentation, and an operational migration/recovery path. Otherwise, R02 should preserve the previous serving semantics until the replacement cold-rebuild mechanism is available.

In other words, the question is not whether Mooncake should silently discard recovery data. The question is whether this PR should change an existing fail-open HA behavior into an unrecoverable fail-closed state without first providing an alternative.

Icedcoco: Thanks, I agree that R02 changes the observable HA failure behavior, and I have now made that explicit. However, I do not consider the previous fail-open behavior a compatibility contract that should be preserved.

Before R02, the supervisor could expose a newly elected master even when `RestoreFromStandby()` had failed. In that case, the serving state was not proven to match the durable HA history. I consider this a correctness bug rather than an intentional “lose cache but remain available” recovery policy.

I pushed commit `571ebc53` to clarify and cover this:

- The PR is now explicitly marked as both a bug fix and a breaking behavioral change.
- The architecture and deployment documentation state that election alone is insufficient for serving. Final catch-up, promotion/export, complete context restoration, and leadership revalidation must all succeed.
- The documentation now describes the operational recovery path. Operators should first repair temporary backend, configuration, or snapshot-access failures. If the recovery history is confirmed unusable and cache loss is acceptable, they must stop all processes using the old `cluster_id` and start a new empty cluster with a previously unused `cluster_id`.
- Mooncake does not automatically delete or ignore recovery data.
- A new supervisor test verifies that a valid empty `PromotionContext` passes through restoration and reaches the post-restore leadership-monitor phase.

I did not add new configuration validation for restoration corruption because this is runtime recovery state and cannot be determined safely at startup. Existing unsupported combinations, including DFS with snapshot/OpLog recovery, are already rejected. Adding a fail-open switch in R02 would reintroduce the same unsafe behavior.

So the intended R02 contract is now explicit: restoration failure remains fail-closed, while deliberate recovery with data loss is an operator decision performed through a new cluster namespace. A generation-fenced managed reset can be designed separately if we want to automate the mechanics without letting Mooncake decide by itself to discard recovery state.

Aionw (changes requested):
Thanks for making the fail-closed behavior explicit, documenting the breaking change, and adding an operator-controlled recovery procedure. With #3566 merged, I no longer consider the leadership-only HA path or the fail-closed policy itself blocking here.

I still think the following changes are needed before this PR is ready:

1. **Update the branch onto current `main` and rerun CI.**

   #3566 was merged after this PR's latest commit and after its current checks ran. Although GitHub reports the PR as mergeable, the combined result has not been validated by this PR's CI.

2. **Cover the real combined HA-without-OpLog path.**

   `EmptyPromotionContextPassesRestoreGate` injects a `FakeStandbyController` that directly returns an empty context. It verifies only the restore half of the path. The observable regression boundary is:

   `CreateStandbyController()` → `NoopStandbyController` from #3566 → empty `PromotionContext` → restore gate → serving.

   Please add coverage for that real path after updating onto `main`, so the two PRs are validated together rather than through separate implementation-level tests.

3. **Do not add test-only branches to the production supervisor loop.**

   The optional `coordinator_override` and `standby_controller` parameters change production control flow solely to inject test doubles, and `RunSupervisorLoopForTesting` exposes that path in a public production header. This makes the production loop carry two construction paths and expands its API for tests. Please test through an existing behavior boundary, or factor a production abstraction that is used unconditionally by both production and tests, without a test-only public entry point or conditional production path.

The recovery-policy discussion is otherwise addressed by the updated documentation and the explicit new-`cluster_id` recovery procedure.

mooncake-store/src/ha/leadership/master_service_supervisor.cpp
Icedcoco: Thanks, I addressed all three requested changes.

1. The branch is rebased onto current main at 7da466d36, including #3566, and force-pushed with lease. New PR CI is now running from head 27c546f25.

2. I replaced the injected empty-context test with real combined-path coverage. HighAvailabilityTest.HaWithoutOplogRestoresEmptyContextAndServes starts a real MasterServiceSupervisor in a child process against real etcd with enable_oplog=false. The observed path is:

   CreateStandbyController() -> NoopStandbyController -> empty PromotionContext -> RestoreFromStandby() -> ha_state=serving, service_ready=true.

   The test also verifies that /get_all_keys is available after the transition.

3. I removed RunSupervisorLoopForTesting, the public declarations added for it, the optional coordinator_override / injected standby_controller parameters, and the related fake-controller tests. Production now always constructs its coordinator and standby controller through the normal factories; there is only one production control path.

Local verification on a fresh current-main build:

- full high_availability_test: 14/14 passed against isolated local etcd;
- full hot_standby_snapshot_bootstrap_test: 7/7 passed;
- restore descriptor mismatch, overlap, and DFS rejection tests: 3/3 passed;
- pre-commit: passed.

The fail-closed policy and documented new-cluster_id recovery procedure are otherwise unchanged.

Aionw (approved):
Re-reviewed at 27c546f25 after #3566 merged. The previous blockers are addressed: the branch is rebased onto current main with CI rerun, the integration test exercises the real etcd HA path from NoopStandbyController through an empty PromotionContext and RestoreFromStandby to serving, and the test-only supervisor injection paths have been removed. The fail-closed behavior and operator-controlled new-cluster-id recovery procedure are also documented clearly. I found no remaining blocking issues in the updated diff. Approving; merge should still wait for the required CI checks, which are currently in progress.

ykwd (approved):

PR #3498, opened 2026-08-18
[Bugfix][TENT] Accept legacy control RPC IDs during rolling upgrades

RuiqingFeng:
## Description

Adding Probe before Delegate changed the implicit control RPC IDs:
| Wire ID | Legacy request | Current request |
| --- | --- | --- |
| `6` | Delegate object | Probe empty payload |
| `7` | Pin string | Delegate object |
| `8` | Unpin number | Pin string |

An older client therefore reaches the wrong handler after the server is upgraded. Pinning the current enum values prevents another accidental shift, and the server now uses the request shape to route these three legacy IDs to the corresponding current handlers.
The first commit adds a failing loopback regression test. The second commit contains only the compatibility fix. This supports server-first rolling upgrades; servers must be upgraded before clients.

## Module
- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?
The regression test sends legacy Delegate, Pin, and Unpin wire IDs to a current loopback server and verifies that each request reaches the intended handler.
**Test commands:**
```bash
cmake --build build-tent --target tent_rpc_handler_isolation_test
build-tent/mooncake-transfer-engine/tent/tests/tent_rpc_handler_isolation_test \
  --gtest_filter='ControlRpcCompatibility.LegacyIdsReachExpectedHandlers'
```
**Test results:**
- [x] Unit tests pass
  - Verified regression test `ControlRpcCompatibility.LegacyIdsReachExpectedHandlers` passes.

## Checklist

## AI Assistance Disclosure
- [ ] No AI tools were used

### DIFF
--- a/mooncake-transfer-engine/tent/include/tent/rpc/rpc.h
+++ b/mooncake-transfer-engine/tent/include/tent/rpc/rpc.h
@@ -37,16 +37,16 @@ namespace tent {
 
 enum RpcFuncID {
     GetSegmentDesc = 1,
-    BootstrapRdma,
-    SendData,
-    RecvData,
-    Notify,
-    Probe,
-    Delegate,
-    Pin,
-    Unpin,
-    SubscribeSegmentUpdate,
-    NotifySegmentUpdated,
+    BootstrapRdma = 2,
+    SendData = 3,
+    RecvData = 4,
+    Notify = 5,
+    Delegate = 6,
+    Pin = 7,
+    Unpin = 8,
+    Probe = 9,
+    SubscribeSegmentUpdate = 10,
+    NotifySegmentUpdated = 11,
 };
 
 class ClientPool;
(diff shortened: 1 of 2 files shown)


alogfans:
[P1] This only supports old clients talking to new servers, not the reverse direction.
In rpc.cpp, the new server guesses legacy IDs from the payload shape, but during a rolling upgrade a new client talking to an old server will still break: old servers will interpret Delegate=7 as legacy Pin=7, etc.
The better fix is to preserve all existing wire IDs for existing RPCs and assign Probe a new unused ID, instead of inferring protocol version from the JSON payload’s first character.

RuiqingFeng:
> [P1] This only supports old clients talking to new servers, not the reverse direction. In rpc.cpp, the new server guesses legacy IDs from the payload shape, but during a rolling upgrade a new client talking to an old server will still break: old servers will interpret Delegate=7 as legacy Pin=7, etc. The better fix is to preserve all existing wire IDs for existing RPCs and assign Probe a new unused ID, instead of inferring protocol version from the JSON payload’s first character.

Thanks for the feedback! I have updated the PR accordingly:

- Preserved the historical wire IDs for all existing control RPCs (GetSegmentDesc through Unpin as 1..8, and segment update notifications as 10..11).
- Assigned Probe a distinct unused ID (9), ensuring bidirectional compatibility during rolling cluster upgrades.
- Removed all runtime payload inspection/version guessing in rpc.cpp.
- Added compile-time static_assert and unit test coverage to ensure wire IDs remain stable.

Issue #3501, opened 2026-08-18
[TE][EFA]: Split transfers that straddle BufferDesc chunk boundaries


staryxchen:
Parent: #3500

## Problem

`EfaTransport::registerLocalMemory` already auto-splits buffers larger than `max_mr_size` / PTE limit into multiple `BufferDesc`s, similar to RDMA (#2644).

`EfaTransport::submitTransferTask` still posts **one slice per request**. `selectDevice` requires the entire `request.length` to fit in a single `BufferDesc`. A Store slice (or any TE read/write) that crosses a chunk seam fails with `ERR_ADDRESS_NOT_REGISTERED`.

Until this is fixed, Store must keep splitting EFA segments in `GetTransportRegistrationLimit()`.

## Proposed work

Mirror RDMA `SliceLengthCalculator`: cap each posted slice at the remaining bytes of both the source and target `BufferDesc`, then look up lkey/rkey per sub-slice.

## Test plan

Needs EFA hardware. Suggested coverage, matching `rdma_large_mr_test.cpp`:

- WRITE whose destination straddles a chunk seam
- WRITE whose source straddles a chunk seam
- WRITE whose target is past the first chunk (no truncation)

I do not have EFA hardware to implement or verify this.

staryxchen:
Hi @whn09 @yejun-yun Both of you are quite experienced with EFA, could you help take this feature?

whn09:
> Hi [@whn09](https://github.com/whn09) [@yejun-yun](https://github.com/yejun-yun) Both of you are quite experienced with EFA, could you help take this feature?

@staryxchen Can you have a look at https://github.com/kvcache-ai/Mooncake/pull/3505 ?

Issue #3500, opened 2026-08-18
[TE]: Track RDMA-parity large-buffer support for EFA/CXI/UB


staryxchen:
## Background

RDMA Transfer Engine already handles buffers larger than the device MR limit end-to-end:

1. Registration auto-chunks into multiple MRs / `BufferDesc`s (#2644).
2. Submit-time slicing caps each WR at the source and target `BufferDesc` remainder (`SliceLengthCalculator` in `RdmaTransport::submitTransferTask`), so a single TE read/write that straddles a chunk seam still works.

Because of that, Mooncake Store no longer needs to split RDMA segments at `max_mr_size`. #3486 is dropping `rdma` from `GetTransportRegistrationLimit()` and leaving the Store-side split in place for EFA, CXI, and UB.

## Goal

Bring EFA, CXI, and UB to the same TE behavior as RDMA, so Store only needs a registration-size split for transports that still cannot register or transfer across chunk seams.

After all three sub-issues land, `GetTransportRegistrationLimit()` should only apply to transports that still lack this support (today that is expected to be none, or UB until its registration shrink is fixed).

## Why this is split into sub-issues

I do not have EFA or CXI hardware to implement and verify the submit-time path, and UB is a different gap (it still shrinks oversized registrations instead of chunking). Tracking them separately keeps owners and test requirements clear.

## Sub-issues

- [x] #3501 EFA: submit-time cross-`BufferDesc` slicing (registration chunking already exists)
- [ ] #3502 CXI: submit-time cross-`BufferDesc` slicing (registration chunking already exists)
- [ ] #3503 UB: registration auto-chunking **and** submit-time cross-`BufferDesc` slicing

## References

- #2644 RDMA registration auto-chunk
- #3486 Store protocol-aware segment split
- `RDMALargeMrTest.WriteStraddlesChunkBoundary` / `WriteWithSourceStraddlingChunkBoundary`

PR #3505, opened 2026-08-18
[TransferEngine] Split EFA transfers that straddle BufferDesc chunk boundaries

whn09:
## Description

Implements #3501 (child of #3500).

`EfaTransport::registerLocalMemoryInternal()` already splits a buffer larger than
`min(max_mr_size, per-NIC PTE budget)` into several chunks and publishes one
`BufferDesc` per chunk, but submission did not follow:

- `submitTransferTask()` posted exactly **one slice per request**, and
- `selectDevice()` only accepts a `BufferDesc` that contains the **whole**
  `request.length`.

So any read/write crossing a chunk seam failed with `ERR_ADDRESS_NOT_REGISTERED`
on the source side, or `Cannot select device for dest_addr` from
`EfaContext::submitPostSend()` on the target side — which is why
mooncake-store still has to split EFA segments in
`GetTransportRegistrationLimit()`.

**Fix (as prescribed in the issue):** mirror the RDMA transport — cap each
posted slice at the remaining bytes of *both* the source and the target
`BufferDesc`, and look up the lkey per sub-slice (the target rkey is already
resolved per slice inside `submitPostSend()`, so capping to one peer chunk fixes
that side automatically).

Notes on the deliberate differences from RDMA's `SliceLengthCalculator`:

- **No cut at `globalConfig().slice_size`.** EFA posts one WR per slice and SRD
  handles multi-MB messages natively (`MC_SLICE_SIZE` is a no-op on this
  transport), so a request that stays inside one chunk on both sides still
  becomes exactly one slice — behaviour and performance are unchanged for
  everything that works today.
- Chunks whose key vector is all zeros (a NIC the chunk was not registered on
  under the disjoint per-NIC partition) are skipped, so capping a slice never
  turns one miss into several.
- Slices already queued for other contexts are marked failed before returning an
  error, instead of being leaked.
- Drive-by: an out-of-range or inactive `advise_retry_cnt` device on the request
  no longer returns `InvalidArgument`; it falls back to per-slice device
  selection.

Follow-up (not in this PR): once this and the sibling issues land, EFA's
`GetTransportRegistrationLimit()` in mooncake-store can be relaxed.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

Hardware: 2x `p5.48xlarge` (8x H100, 32x EFA NIC, libfabric 2.4.0amzn5.0,
CUDA 13.0.88). Both trees built with
`-DUSE_EFA=ON -DUSE_CUDA=ON -DBUILD_UNIT_TESTS=ON`. Every result below is
base-vs-PR on the same box, where "base" is `main` **carrying the same new test
file**, so the tests are shown red before green.

**Test commands:**
```bash
# 1. New regression test (needs EFA hardware; MC_MAX_MR_SIZE forces the split
#    at 64 MiB so a 256 MiB buffer becomes 4 chunks)
MC_MAX_MR_SIZE=67108864 ./mooncake-transfer-engine/tests/efa_large_mr_test

# 2. Existing EFA suite, default config
./mooncake-transfer-engine/tests/efa_transport_test

# 3. Cross-node wire path: block_size is deliberately not a divisor of
#    MC_MAX_MR_SIZE, so the request spanning byte 67108864 straddles the seam
#    on BOTH sides.  The target runs the UNPATCHED binary -- the fix is
#    initiator-side only and wire-compatible.
#  target (node B):
MC_MAX_MR_SIZE=67108864 ./transfer_engine_bench --mode=target \
  --metadata_server=P2PHANDSHAKE --local_server_name=<B>:12345 --protocol=efa \
  --gpu_id=-1 --block_size=1000000 --batch_size=32 --threads=4 \
  --buffer_size=128000000
#  initiator (node A):
MC_MAX_MR_SIZE=67108864 ./transfer_engine_bench --mode=initiator \
  --metadata_server=P2PHANDSHAKE --local_server_name=<A>:12346 \
  --segment_id=<B>:<listen_port> --protocol=efa --operation=write|read \
  --gpu_id=-1 --block_size=1000000 --batch_size=128 --threads=1 \
  --buffer_size=128000000 --duration=5
```

**Test results:**

1. `efa_large_mr_test` (4 cases: baseline past the first chunk, target
   straddling, source straddling, READ straddling on both sides)

| build | result |
|---|---|
| base (`main` + this test) | **1 passed, 3 FAILED** — `Cannot select device for dest_addr 0x…80001` (target side) and `Memory region not registered by any active EFA device(s)` (source side) |
| this PR | **4/4 passed** (8.6 s) |

2. `efa_transport_test`: **11/11 passed** on both builds (28.4 s vs 28.5 s) — no
   regression in the single-chunk path.

3. Cross-node, real `fi_write`/`fi_read` (not the same-process memcpy loopback):

| build | write | read |
|---|---|---|
| base | **aborted** at the straddling request (`AddressNotRegistered`) | **aborted** likewise |
| this PR | completes, 1651 batches in 5.00 s = 42.24 GB/s (≈3.03 ms per 128 MB batch) | 1473 batches in 5.00 s = 37.70 GB/s |

   (Single submission thread with 1 MB blocks, so these are throughput sanity
   numbers, not a peak-bandwidth measurement.)

4. Byte-exactness on the wire (extra local harness, not part of this PR — the
   in-repo gtest is single-process, where `tryLoopbackCopy()` short-circuits to
   `memcpy` and never touches an lkey/rkey; this complements it): a 4 MiB
   straddling WRITE + READ round-trip between the two nodes compares
   byte-for-byte, and a subsequent inner 1 MiB straddling WRITE lands exactly at
   the intended address with both guard regions around it bit-identical. Passes
   on this PR; the base build aborts before the first transfer.

- [x] Unit tests pass

`efa_large_mr_test` is built but intentionally **not** registered with
`add_test()`, exactly like the existing `rdma_large_mr_test`, because it needs
EFA hardware and `MC_MAX_MR_SIZE` in the environment before process start.

## Checklist

`pre-commit run --files <the 3 files>` passes (clang-format 20.1.8). The
`cmake-format` hook additionally wants to reflow three pre-existing lines in
`mooncake-transfer-engine/tests/CMakeLists.txt` that this PR does not touch
(`shared_segment_test`'s `target_include_directories` and a comment above
`worker_pool_rail_state_test`); per `AGENTS.md` those unrelated edits were left
out.

## AI Assistance Disclosure

- [ ] No AI tools were used

Claude Code was used to implement the slicing helpers, write the new test, and
run the base-vs-PR hardware validation above. The submitter has reviewed every
changed line.

### DIFF
--- a/mooncake-transfer-engine/src/transport/efa_transport/efa_transport.cpp
+++ b/mooncake-transfer-engine/src/transport/efa_transport/efa_transport.cpp
@@ -939,13 +939,102 @@ Status EfaTransport::submitTransfer(
     return submitTransferTask(task_list);
 }
 
+// Bytes from `address` to the end of the registered chunk that covers it, or 0
+// if no chunk does.  `last_buffer_idx` caches the previous hit so that slicing
+// one request across a seam does not rescan the buffer list per slice.
+//
+// registerLocalMemoryInternal() splits a buffer larger than
+// min(max_mr_size, per-NIC PTE budget) into several chunks and publishes one
+// BufferDesc per chunk, so what the caller registered as a single contiguous
+// range can be several MRs here -- on either side of the transfer.
+static uint64_t efaBytesUntilChunkEnd(
+    const TransferMetadata::SegmentDesc* segment_desc, uint64_t address,
+    bool require_remote_key, size_t& last_buffer_idx) {
+    if (!segment_desc || segment_desc->buffers.empty()) return 0;
+
+    auto is_valid_buffer = [&](const TransferMetadata::BufferDesc& buffer) {
+        const auto& keys = require_remote_key ? buffer.rkey : buffer.lkey;
+        if (keys.empty()) return false;
+        if (address < buffer.addr || address - buffer.addr >= buffer.length)
+            return false;
+        // A chunk under the disjoint per-NIC partition carries a zero key for
+        // every NIC it was not registered on (see the nic_assignments fan-out
+        // in registerLocalMemoryInternal).  All-zero means no NIC can serve it
+        // and selectDevice() will reject it, so capping a slice at such a chunk
+        // would only turn one miss into several.
+        for (auto key : keys)
+            if (key != 0) return true;
+        return false;
+    };
+
+    if (last_buffer_idx < segment_desc->buffers.size()) {
+        const auto& buffer = segment_desc->buffers[last_buffer_idx];
+        if (is_valid_buffer(buffer)) {
+            return buffer.length - (address - buffer.addr);
+        }
+    }
+
+    uint64_t max_remaining = 0;
+    for (size_t i = 0; i < segment_desc->buffers.size(); ++i) {
+        const auto& buffer = segment_desc->buffers[i];
+        if (is_valid_buffer(buffer)) {
+            uint64_t remaining = buffer.length - (address - buffer.addr);
+            if (remaining > max_remaining) {
+                max_remaining = remaining;
+                last_buffer_idx = i;
+            }
+        }
+    }
+    return max_remaining;
+}
+
+// Slice length for a request, capped at the nearest registration boundary.
+//
+// Unlike RDMA this deliberately does not cut at globalConfig().slice_size: EFA
+// posts one WR per slice and SRD handles multi-MB messages natively, so a
+// request that stays inside one chunk on both sides still becomes exactly one
+// slice, as before.  The one thing that does force a split is an MR boundary --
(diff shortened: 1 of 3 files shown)


staryxchen (approved):
Thank you for responding so quickly. LGTM.

alogfans (approved):

PR #3508, opened 2026-08-18
fix(transfer-engine): reclaim timed-out batches safely

voidxb:
## Summary

This PR fixes transfer-engine resource leaks around timed-out or failed transfer batches and Barex submission/registration failure paths.

Changes include:

- Reclaim timed-out transfer batches safely through deferred cleanup.
- Free Python wrapper batch IDs on submit failure, timeout, and status-check timeout.
- Mark Barex slices as failed when they are prepared but cannot be submitted.
- Roll back Barex memory registration on partial registration or metadata update failures.
- Harden Barex resource cleanup for partially initialized transports.

## Timeout Cleanup Tradeoff

A transfer `TIMEOUT` only means polling observed that a slice exceeded its timeout. It does not guarantee that the underlying transport operation has stopped or that transport callbacks will no longer touch the batch/task/slice memory.

Freeing `BatchDesc` immediately on timeout can therefore introduce use-after-free if a late callback later calls `markSuccess()` or `markFailed()`.

This PR uses a deferred cleanup worker as a conservative short-term fix:

- The worker is started lazily only when `freeBatchID()` sees a busy timed-out/failed batch.
- Normal completed transfers do not create the worker.
- The worker retries reclamation off the hot path until callbacks make the batch terminal.
- During shutdown, it performs a final retry and exits even if unresolved batches remain busy, logging a warning instead of blocking teardown indefinitely.

This trades a small amount of exceptional-path complexity and one lazily-created thread for avoiding both immediate UAF risk and permanent leaks for recoverable late-completion cases.

A longer-term alternative would be to make slice completion strictly idempotent and allow timeout paths to force pending slices into a terminal failed state, but that requires a broader transport callback semantics change.


mooncake-transfer-engine/src/multi_transport.cpp
```diff
@@ -102,14 +112,59 @@ MultiTransport::BatchID MultiTransport::allocateBatchID(size_t batch_size) {
...
+        // TIMEOUT does not guarantee that the transport callback has returned.
+        // Keep the batch alive until every slice reports completion or failure.
+        if (status.s != Transport::TransferStatusEnum::COMPLETED &&
+            status.s != Transport::TransferStatusEnum::FAILED) {
             return Status::BatchBusy(
                 "BatchID cannot be freed until all tasks are done");
         }
     }
+
     delete &batch_desc;
```
alogfans, line 178: [P1] A terminal task state does not guarantee that the completion callback has returned. In the event-driven path, `task.is_finished` is set before the callback finishes accessing `batch_desc` (`transport.h:248-258`), so this cleanup thread may delete the batch concurrently and still cause a use-after-free. Please introduce explicit callback quiescence or reference ownership and reclaim the batch only after the final callback has fully exited.

mooncake-transfer-engine/src/multi_transport.cpp
```diff
@@ -102,14 +112,59 @@ MultiTransport::BatchID MultiTransport::allocateBatchID(size_t batch_size) {
...
+        // stopped touching the batch, so immediate deletion can cause UAF; the
+        // worker retries reclamation off the hot path until callbacks make the
+        // batch terminal, while normal completed transfers never create it.
+        deferred_cleanup_thread_ =
+            std::thread(&MultiTransport::deferredCleanupLoop, this);
+    }
+    deferred_cleanup_cv_.notify_one();
+    LOG(WARNING) << "Batch " << batch_id
+                 << " is still busy; cleanup has been deferred";
+    return Status::OK();
```
alogfans: `freeBatchID()` now returns success even when the batch is still active and only scheduled for deferred cleanup. This changes the existing `BatchBusy` contract and may allow callers to release or reuse transfer buffers while the underlying DMA is still accessing them. Please preserve the busy result or introduce a separate deferred-release API with explicit ownership semantics.

voidxb: done, please review again

mooncake-transfer-engine/src/multi_transport.cpp
voidxb: done, please review again

mooncake-transfer-engine/src/multi_transport.cpp
```diff
@@ -102,14 +112,69 @@ MultiTransport::BatchID MultiTransport::allocateBatchID(size_t batch_size) {
...
+
+    deferred_cleanup_batches_.insert(batch_id);
+    if (!deferred_cleanup_thread_.joinable()) {
+        // Start the cleanup worker lazily after a caller observes BatchBusy.
+        // Deferred cleanup is only a safety net for timed-out or failed batches;
+        // freeBatchID still preserves the public contract that BatchBusy means
+        // transfer buffers may remain owned by in-flight transport work.
+        deferred_cleanup_thread_ =
+            std::thread(&MultiTransport::deferredCleanupLoop, this);
+    }
```
alogfans, line 133: Once a batch is enqueued for deferred cleanup, we need to explicitly signal to the caller that the batch handle is no longer valid and must not be used for any further operations.

voidxb: I fixed , please review it again

PR #3509, opened 2026-08-18
[Store] Clean replication resources when source disappears

### LINKED ISSUE #3401: [Bug]: Replication tasks retain targets and reserved quota after source cleanup
## Bug Report

While reviewing #2877, I found a pre-existing replication-task lifecycle gap. This is not caused by making segment cleanup asynchronous, but async cleanup makes the retained state easier to observe.

When an in-flight Copy/Move source becomes unavailable, `CleanupStaleHandles()` removes the invalid COMPLETE source replica. It only cancels promotion tasks for removed replica IDs, so a replication task whose `source_id` was removed is left behind together with its PROCESSING target replicas and reserved tenant quota. Those resources remain until the client calls `CopyEnd/MoveEnd/CopyRevoke/MoveRevoke` or `DiscardExpiredProcessingReplicas()` reaches the task timeout.

### Reproduction

A deterministic unit test can use the existing `MasterServiceTest` fixture:

1. Create a non-HA `MasterService` with a tenant quota large enough for exactly the source and one temporary target.
2. Mount `source_segment` and `target_segment`.
3. `PutStart` + `PutEnd` a key on `source_segment`.
4. Call either:
   - `CopyStart(client, key, ..., "source_segment", {"target_segment"})`, or
   - `MoveStart(client, key, ..., "source_segment", "target_segment")`.
5. Unmount `source_segment` and wait until invalid-handle cleanup has removed the source replica.
6. Do not call the corresponding End/Revoke API.
7. Inspect the key's `TenantState`, or attempt another allocation under the strict quota.

Observed state after step 5:

- the source replica is absent;
- the target replica is still PROCESSING;
- `tenant_state.replication_tasks` still contains the key;
- `reserved_quota_charge_bytes` is still held;
- physical metadata remains until the replication-task timeout.

A compact assertion sketch:

```cpp
ASSERT_TRUE(service.CopyStart(client, key, tenant, src, {dst}).has_value());
ASSERT_TRUE(service.UnmountSegment(src_id, src_owner).has_value());
WaitUntil([&] { return source_replica_is_gone(); });

EXPECT_TRUE(tenant_state.replication_tasks.contains(key));
EXPECT_TRUE(target_replica_is_processing());
EXPECT_GT(tenant_state.replication_tasks.at(key)
              .reserved_quota_charge_bytes,
          0);
```

### Expected behavior

Removing a replica that is the source of a Copy/Move task should invalidate that task promptly, discard its PROCESSING targets, release its reserved quota, and balance the source reference if it still exists.

### Relevant code

- `MasterService::CleanupStaleHandles()`
- `MasterService::CancelPromotionTaskForRemovedReplicas()`
- `MasterService::DiscardExpiredProcessingReplicas()`

This should be handled separately from #2877 so that the asynchronous-unmount PR can remain focused on RPC latency and immediate logical invisibility.

982945902:
## Description

Fixes #3401.

When the COMPLETE source replica of an in-flight Copy/Move task disappears during stale-handle cleanup, the master now:

- removes the task's PROCESSING target replicas;
- releases the task's reserved tenant quota;
- clears dynamic replication state;
- keeps a lightweight `source_removed` task tombstone so a concurrent or delayed `CopyEnd`/`MoveEnd` still returns `REPLICA_IS_GONE` and `CopyRevoke`/`MoveRevoke` keeps its existing success semantics;
- performs the same cleanup only after the stale-replica OpLog entry becomes durable in HA mode.

The tombstone is removed by the next End/Revoke request or by the existing replication-task timeout reaper.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
master_service_tenant_quota_test \
  --gtest_filter='MasterServiceTenantQuotaTest.CopySourceCleanupReleasesTargetAndReservedQuota:MasterServiceTenantQuotaTest.MoveSourceCleanupReleasesTargetAndReservedQuota:MasterServiceTenantQuotaTest.CopySourceCleanupWaitsForOpLogDurability'

master_service_test \
  --gtest_filter='MasterServiceTest.CopyEnd:MasterServiceTest.CopyRevoke:MasterServiceTest.MoveEnd:MasterServiceTest.MoveRevoke:MasterServiceTest.CopyInProgressDoesNotKeepUnmountedSourceVisible:MasterServiceTest.MoveInProgressDoesNotKeepUnmountedSourceVisible'

git diff --check
```

The regression tests were also compiled against unmodified `origin/main`; all three failed because quota remained charged until the replication timeout. With this change, all three pass.

**Test results:**
- [x] Unit tests pass
- [x] Integration tests pass (HA OpLog durability path uses a blocking transactional test backend)

## Checklist

- [x] I have updated the documentation (not applicable: internal lifecycle fix)
- [x] For changes >500 LOC: not applicable

`./scripts/code_format.sh` requires clang-format 20, while the current development host provides clang-format 17. The touched files were formatted with the available formatter and `git diff --check` passes.

## AI Assistance Disclosure

- [ ] No AI tools were used

OpenAI Codex assisted with the lifecycle analysis, implementation, HA durability review, and regression testing. The submitter reviewed the final change and test results.

### DIFF
--- a/mooncake-store/include/master_service.h
+++ b/mooncake-store/include/master_service.h
@@ -1535,7 +1535,7 @@ class MasterService {
         uint64_t pending_quota_charge_bytes{0};
         UUID dynamic_replication_lease_id{};
         uint64_t dynamic_replication_version_epoch{0};
-        bool durable_cleanup_pending{false};
+        uint64_t durable_cleanup_generation{0};
         bool source_removed{false};
     };
 
@@ -1882,7 +1882,8 @@ class MasterService {
         MetadataShardAccessorRW* shard);
     void FinalizeRemovedReplicasAfterDurable(
         const OpLogEntry& durable_entry,
-        const std::vector<ReplicaID>& replica_ids, QuotaEraseMode quota_mode);
+        const std::vector<ReplicaID>& replica_ids, QuotaEraseMode quota_mode,
+        uint64_t replication_cleanup_generation = 0);
     void FinalizeMetadataEraseAfterDurable(const OpLogEntry& durable_entry,
                                            QuotaEraseMode quota_mode);
     void FinalizeExpiredProcessingReplicasAfterDurable(
@@ -1893,7 +1894,8 @@ class MasterService {
         const std::vector<ReplicaID>& target_ids,
         const UUID& dynamic_replication_lease_id,
         uint64_t dynamic_replication_version_epoch,
-        const std::chrono::system_clock::time_point& ttl);
+        const std::chrono::system_clock::time_point& ttl,
+        uint64_t cleanup_generation);
     struct StaleHandleCleanupPlan {
         std::vector<ReplicaID> removed_ids;
         std::vector<Replica::Descriptor> remaining;
@@ -2074,8 +2076,8 @@ class MasterService {
         NO_THREAD_SAFETY_ANALYSIS;
     void CancelReplicationTaskForRemovedSource(
         TenantState& tenant_state, ObjectMetadata& metadata,
-        const std::vector<ReplicaID>& removed_replica_ids)
-        NO_THREAD_SAFETY_ANALYSIS;
+        const std::vector<ReplicaID>& removed_replica_ids,
+        uint64_t cleanup_generation = 0) NO_THREAD_SAFETY_ANALYSIS;
 
     // Lease related members
     const uint64_t default_kv_lease_ttl_;     // in milliseconds
@@ -2432,6 +2434,7 @@ class MasterService {
     void ClientMonitorFunc();
     std::thread client_monitor_thread_;
     std::atomic<bool> client_monitor_running_{false};
+    std::atomic<uint64_t> next_replication_cleanup_generation_{1};
     static constexpr uint64_t kClientMonitorSleepMs =
         1000;  // 1000 ms sleep between client monitor checks
     // boost lockfree queue requires trivial assignment operator
--- a/mooncake-store/include/master_service.h
+++ b/mooncake-store/include/master_service.h
@@ -1536,6 +1536,7 @@ class MasterService {
         UUID dynamic_replication_lease_id{};
         uint64_t dynamic_replication_version_epoch{0};
         bool durable_cleanup_pending{false};
+        bool source_removed{false};
     };
 
     struct OffloadingTask {
@@ -2071,6 +2072,10 @@ class MasterService {
         TenantState& tenant_state, ObjectMetadata& metadata,
         const std::vector<ReplicaID>& removed_replica_ids)
         NO_THREAD_SAFETY_ANALYSIS;
+    void CancelReplicationTaskForRemovedSource(
+        TenantState& tenant_state, ObjectMetadata& metadata,
+        const std::vector<ReplicaID>& removed_replica_ids)
+        NO_THREAD_SAFETY_ANALYSIS;
 
     // Lease related members
     const uint64_t default_kv_lease_ttl_;     // in milliseconds
@@ -2172,8 +2177,14 @@ class MasterService {
                             << ", bytes=" << before_charge - after_charge;
                     }
                 }
-                // If no valid replicas remain, delete the whole object.
-                if (!it_->second.IsValid()) {
+                // If no valid replicas remain, delete the whole object unless
+                // a cancelled replication task still preserves End/Revoke
+                // protocol semantics.
+                const bool preserve_cancelled_replication =
+                    replication_task_it_ !=
+                        tenant_state_->replication_tasks.end() &&
+                    replication_task_it_->second.source_removed;
+                if (!it_->second.IsValid() && !preserve_cancelled_replication) {
                     // NOTE: Erase() -> EraseMetadata() already removes the key
                     // from processing_keys (by key), so calling
                     // EraseFromProcessing() here would re-erase the same node
--- a/mooncake-store/include/master_service.h
+++ b/mooncake-store/include/master_service.h
@@ -1907,8 +1907,8 @@ class MasterService {
         const std::function<bool(const Replica&)>& is_stale) const;
     tl::expected<void, ErrorCode> PersistStaleHandleCleanupForHA(
         const std::string& why, const TenantId& tenant_id,
-        const std::string& key, ObjectMetadata& metadata,
-        const StaleHandleCleanupPlan& plan);
+        const std::string& key, TenantState& tenant_state,
+        ObjectMetadata& metadata, const StaleHandleCleanupPlan& plan);
     void RebuildGroupRoutingIndex();
     void GrantLeaseForGroup(const TenantState& tenant_state,
                             const std::string& key,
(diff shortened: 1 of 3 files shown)


he-yufeng (commented):
Read through the tombstone flow and one structural point is worth recording. Today every object erasure goes through EraseMetadata, which also removes the key's replication task (master_service.cpp:2131), so object-gone with task-present was unreachable before this change. The preservation branches in FinalizeRemovedReplicasAfterDurable / ClearInvalidHandles / CleanupStaleHandles keep that invariant from the other side, and the delayed End/Revoke branch then lands on Erase() -> EraseMetadata -> task erase, which stays idempotent against the earlier cancel since pending_quota_charge_bytes is already zeroed. That chain is the load-bearing part and it holds.

One coverage gap: the delayed protocol callbacks are asserted for CopyEnd and MoveRevoke, but the symmetric delayed MoveEnd (expects REPLICA_IS_GONE) and CopyRevoke (expects success) pair is not. The code paths are mirrors, but these branches are exactly what this PR adds, so a parametrized test over the four combinations would close the loop.

Icedcoco:
Thanks for the detailed fix. I had two questions while tracing the cleanup paths and may be missing an ordering guarantee here.

First, in `PersistStaleHandleCleanupForHA()` (`mooncake-store/src/master_service.cpp:2028`), the stale source is marked removed and the cleanup OpLog is queued, but the replication task does not appear to be marked as `durable_cleanup_pending` on this path. Could `CopyEnd`/`CopyRevoke` or `MoveEnd`/`MoveRevoke` still arrive before the durable-finalize callback and clean up the task locally? If so, would it make sense to gate those callbacks (and the timeout reaper) until the OpLog is durable, and add a test with the blocking backend covering a callback that arrives during that window?

Second, in the `MetadataAccessorRW` cleanup path (`mooncake-store/include/master_service.h:2108`), invalid memory replicas are removed and `CancelPromotionTaskForRemovedReplicas()` is called, but the corresponding replication-task cleanup helper is not called there. Could an intervening RW request remove the source before `ClearInvalidHandles()` runs, leaving the replication task and target replicas without a source ID for `CancelReplicationTaskForRemovedSource()` to match? Would it be safer to invoke the new helper in this path as well, with a test covering that ordering?

982945902:
@Icedcoco Thanks for calling out both ordering gaps. I pushed follow-up commit `28ebf2c6` addressing them:

1. `PersistStaleHandleCleanupForHA()` now marks a matching replication task as `durable_cleanup_pending` before queueing the cleanup OpLog, and clears the flag if queueing fails. This keeps Copy/Move End/Revoke callbacks and the timeout cleanup path gated until durable finalization.
2. `MetadataAccessorRW` now calls `CancelReplicationTaskForRemovedSource()` immediately after removing invalid memory replicas, alongside promotion-task cancellation.

Added focused regression coverage for:
- Copy callbacks arriving while the cleanup transaction is blocked;
- Move callbacks arriving while the cleanup transaction is blocked;
- an intervening `MetadataAccessorRW` cleanup removing the replication source before the normal stale-handle sweep.

Validation:
- built `master_service_tenant_quota_test` successfully;
- ran the three focused tests above: **3/3 passed**;
- `./scripts/code_format.sh --staged` with clang-format 20: no changes required.

Icedcoco:
Thanks for the follow-up. The two original gaps look addressed now. I took another pass through the cleanup lifecycle and had two additional questions. I may be missing an existing guarantee here.

First, `CancelReplicationTaskForRemovedSource()` removes the PROCESSING targets through `EraseReplicasWithCacheTotalAccounting()`, so their `AllocatedBuffer`s appear to be released immediately. Since `CopyStart`/`MoveStart` has already returned those target descriptors to the client, could there still be an in-flight RDMA write when the source disappears? If the allocator reuses that memory for another object, a late write could potentially overwrite the new object. Would it be safer to retire these targets through `discarded_replicas_` until the replication release deadline, similar to the timeout cleanup path?

Second, could `durable_cleanup_pending` be shared by two different cleanup operations for the same task? For example, the timeout reaper may already have queued an OpLog and set the flag, after which stale-source cleanup queues another entry. The earlier callback could then observe the same flag and finalize the task before the stale-source cleanup entry is durable. There also seems to be a smaller failure window where `Reserve()` succeeds but `Commit()` fails: the source has already been marked `REMOVED`, while the failure path only clears the flag, which may allow a delayed End/Revoke to remove the local state without a durable cleanup entry. Is there another ordering guarantee that prevents these cases? If not, would it make sense to associate the pending state with a specific cleanup generation/sequence, and restore the source state when the commit is rejected?

982945902:
@Icedcoco Thanks — these three cases do overlap, so I handled them as one cleanup-transaction protocol in follow-up commit `1dd770c2`.

The updated rules are:

1. **One cleanup transaction per replication task.** `durable_cleanup_pending` is replaced by a nonzero cleanup generation. Timeout cleanup and stale-source cleanup both refuse/defer when a generation is already owned by another cleanup.
2. **Callbacks are generation-bound.** Both durable-finalize paths capture their generation and revalidate it together with the task identity before changing metadata. An older callback therefore cannot finalize or clear a newer cleanup.
3. **Commit before local mutation.** Stale-source cleanup now reserves and commits the OpLog before marking replicas `REMOVED`. If `Commit()` rejects after a successful `Reserve()`, only the matching generation is cleared; the source remains `COMPLETE`, so no replica-status rollback is necessary.
4. **Targets are retired, not immediately freed.** When the replication source disappears, PROCESSING targets are popped from metadata and moved to `discarded_replicas_` with TTL `task.start_time + put_start_release_timeout_sec_`. Their `AllocatedBuffer`s remain owned until the replication release deadline, preventing allocator reuse while a late RDMA write may still arrive.

The effective state flow is:

```mermaid
stateDiagram-v2
    [*] --> Active
    Active --> Reserved: reserve slot + assign generation\n(no Replica mutation)
    Reserved --> Active: Reserve/Commit rejected\nclear matching generation
    Reserved --> Pending: Commit accepted\nthen mark stale source REMOVED
    Pending --> Pending: competing cleanup sees generation != 0\nand defers
    Pending --> Pending: stale callback generation mismatch\nignored
    Pending --> SourceGone: matching OpLog durable\nremove source; retire targets until deadline\nclear generation
    SourceGone --> Completed: delayed End/Revoke finishes task
    Completed --> [*]
```

Added regression coverage for:

- target buffers remaining in `discarded_replicas_` until the release deadline;
- stale-source cleanup overlapping the timeout reaper, including a mismatched/older durable callback;
- `Reserve()` succeeding while `Commit()` rejects, verifying the source stays `COMPLETE`, no `REMOVED` replica is left behind, and the generation is cleared.

Validation:

- `master_service_tenant_quota_test`: **39/39 passed**;
- focused source-cleanup/generation/commit-failure subset: **9/9 passed**;
- staged clang-format 20 check: clean.

PR #3354, opened 2026-08-10
[Store] Return explicit standby restore errors

Icedcoco:
## Description

This PR hardens the standby-promotion restore path by changing `MasterService::RestoreFromStandbySnapshot()` and its wrapper from `void` to `tl::expected<void, ErrorCode>`.

Restore data is now validated and constructed in temporary state before it is published. The restore rejects:

- invalid or ambiguous memory segment metadata
- unknown memory replica endpoints
- descriptor size mismatches
- overlapping descriptors or capacity overflow
- invalid tenants or duplicate objects
- object cursors newer than the snapshot cursor

If validation fails, the existing metadata, standby segment state, allocator state, and memory accounting remain unchanged.

The supervisor explicitly checks the restore result and logs failures. This PR intentionally preserves the existing serving behavior; fail-closed serving is left to the R02 follow-up. Replica ID preservation and validation remain scoped to I00.

## Module

- [ ] Transfer Engine (`mooncake-transfer-engine`)
- [x] Mooncake Store (`mooncake-store`)
- [ ] Mooncake EP (`mooncake-ep`)
- [ ] Mooncake PG (`mooncake-pg`)
- [ ] Integration (`mooncake-integration`)
- [ ] P2P Store (`mooncake-p2p-store`)
- [ ] Python Wheel (`mooncake-wheel`)
- [ ] Common (`mooncake-common`)
- [ ] Mooncake RL (`mooncake-rl`)
- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

The HA tests cover successful restore, explicit validation failures, and failure atomicity. The tenant-scoped restore test verifies the normal non-HA-facing restore behavior.

**Test commands:**

```bash
env LSAN_OPTIONS=detect_leaks=0 \
  ctest --test-dir build -R '^master_service_ha_test$' --output-on-failure

env LSAN_OPTIONS=detect_leaks=0 \
  ./build/mooncake-store/tests/master_service_test \
  --gtest_filter='MasterServiceTest.StandbySnapshotRestorePreservesTenantScopedKeys'
```

**Test results:**

- [x] Unit tests pass
- [ ] Integration tests pass (not applicable)
- [ ] Manual testing done (not applicable)

[...]

PR #3511, opened 2026-08-18
[TENT] Fix DeviceSelector inflight quota accounting

anranxia:
## Description

TENT's `DeviceSelector` picks a NIC by scoring each device's per-NIC `inflight_bytes`, but charge and release were asymmetric, so the counter that drives scheduling drifted away from reality. Root causes here:

- **Release keyed on the wrong device.** `releaseSliceQuota` released against the current routing NIC (`source_dev_id`), but a rail fallback rewrites `source_dev_id` *after* the charge is made — leaking the original NIC's charge and driving the fallback NIC's unsigned counter into underflow.
- **Retries never re-charged.** The retry path released the quota then resubmitted without re-charging, so retried traffic ran invisible to the scheduler and its (cumulative) latency polluted EWMA.
- **Estimate vs. actual byte mismatch.** The allocator charges a per-slice estimate (`ceil(total/num_slices)`) while release used `slice->length`, leaving residue on some NICs and underflow on others.
- **Baseline mode underflow.** With smart scheduling off, allocation never adds inflight yet completion still subtracted it → unsigned underflow.
- **EWMA polluted by non-success.** Failed/timed-out completions updated EWMA, using the cumulative since-enqueue latency instead of the current NIC attempt.
- **Software-timeout leak.** The terminal timeout path never released the charge.

## Fix

- **Track the charged device and bytes on the slice** (`charged_dev_id` / `charged_bytes`), separately from the routing device, so release always unwinds exactly what was charged, on the device it was charged against.
- **Reconcile the charge in `generatePostPath`** onto the final routing NIC at the slice's real length — covering fallback migration (release the stale charge, re-charge the new NIC) and retry (re-enter the inflight view).
- **Gate all inflight mutation on smart mode** (`chargeDevice` / `release`), so baseline round-robin never touches the counter and cannot underflow.
- **Feed EWMA only from successful completions**, using the current NIC attempt's latency instead of the cumulative since-enqueue time.
- **Release the exact charged bytes on software timeout**, so terminal slices no longer leak their charge.

**Note:** the initial `allocate` still charges the per-slice *estimate* rather than the exact length, on purpose. Charging at selection time keeps the scheduler's per-NIC inflight view fresh the moment a request is submitted — which is exactly what the multi-path distribution decision reads — before the slice ever reaches a worker; deferring the charge to the worker would leave queued-but-not-yet-posted slices invisible to the scheduler and let concurrent requests pile onto the same NIC. The reconcile in `generatePostPath` then converts that estimate to the exact length on the final routing NIC, so the estimate only lives for the brief submit→post window while steady-state (in-flight, posted) inflight stays exact.

## Module

## Type of Change

- [x] Bug fix
- [x] Performance improvement

## How Has This Been Tested?

Built with `-DUSE_TENT=ON` and the existing TENT unit tests pass.

**Test results:**
- [x] Unit tests pass

## Checklist

## AI Assistance Disclosure

Claude Code assisted with root-cause analysis, implementing the fix, and verification. The human submitter has reviewed every changed line and can defend the change end-to-end.


he-yufeng (commented):
Walked the accounting chain end to end and it closes: allocate records exactly what it charged per slice, generatePostPath reconciles onto the final routing NIC before every (re)submit, and releaseSliceQuota now unwinds exactly charged_bytes on charged_dev_id once, with the software-timeout path finally covered. The baseline-mode guards on chargeDevice/releaseInflight remove the underflow-by-design hole too.

One seam worth a sentence: chargeSliceQuota does release-then-charge on a stale charge, so inflight briefly dips below reality between the two calls. Since inflight is a scoring signal rather than an admission gate this only perturbs scheduling for a moment, and the reverse order would transiently double-count during fallback, which is arguably worse. Just confirming that bias is the intended one; a comment in chargeSliceQuota saying so would save the next reader the derivation.

anranxia:
> Walked the accounting chain end to end and it closes: allocate records exactly what it charged per slice, generatePostPath reconciles onto the final routing NIC before every (re)submit, and releaseSliceQuota now unwinds exactly charged_bytes on charged_dev_id once, with the software-timeout path finally covered. The baseline-mode guards on chargeDevice/releaseInflight remove the underflow-by-design hole too.
> 
> One seam worth a sentence: chargeSliceQuota does release-then-charge on a stale charge, so inflight briefly dips below reality between the two calls. Since inflight is a scoring signal rather than an admission gate this only perturbs scheduling for a moment, and the reverse order would transiently double-count during fallback, which is arguably worse. Just confirming that bias is the intended one; a comment in chargeSliceQuota saying so would save the next reader the derivation.

  Thanks for walking the whole chain — and yes, that's exactly the intended bias.

  Confirmed: chargeSliceQuota releases the stale charge before charging the new one on purpose. Since inflight is a scoring signal, not an admission gate, the brief dip between the two calls only nudges a single selection, whereas charging-first would transiently double-count during a fallback migration and over-penalize that NIC — the strictly worse direction. I've added a comment in chargeSliceQuota spelling this out so the next reader doesn't have to re-derive it.

  Also rebased onto latest main: the earlier red CI was an unrelated mooncake-store build break (master_service.cpp
  referencing StandbyObjectMetadata::last_sequence_id) inherited from the older base, not from this change — it's gone now that the branch sits on current main.

anranxia:
Heads-up on the red CI — it's not from this PR.

  All the failing jobs (build-wheel*, tent-ci, ...) die at the same spot in a different module:

  mooncake-store/src/master_service.cpp:3046: error: 'const struct StandbyObjectMetadata' has no member named
  'last_sequence_id'

  This is a pre-existing build break on main from a merge-order race: #3354 added a per-object  entry.metadata.last_sequence_id cursor check in RestoreFromStandbySnapshot, and #3326 then removed that field from StandbyObjectMetadata (and its populator ToStandbyMetadata) for bounded snapshot capture. Both were green individually; together they leave mooncake_store uncompilable. build-flags (which builds only mooncake-transfer-engine, no store) passes, and Check code format passes — this PR's TENT changes are unaffected.

  I've opened #3521 to fix the main break. Once it lands I'll rebase this branch onto main and CI here should go green. This PR itself needs no changes for the failures.

anranxia:
@he-yufeng review comments addressed and rebased onto latest main (CI green except the two runner-timeout jobs) — could you help re-run those and merge when green? Thanks!

anranxia:
@he-yufeng CI is all green now — could you approve & merge when you get a chance? Thanks!

catyans (approved):
LGTM. Re-reviewed the current head: each slice now records the exact device and byte charge, reconciles it onto the final routing NIC before every post/retry, and releases it exactly once across success, failure, cancellation, fallback, and software-timeout paths. Baseline mode no longer mutates inflight accounting, and EWMA only learns from successful current-attempt latency. CI is green.

suggestion (non-blocking): add focused quota-accounting tests for estimate-to-exact reconciliation, fallback migration, retry recharge, and baseline mode so these invariants remain pinned independently of hardware tests.

anranxia:
> LGTM. Re-reviewed the current head: each slice now records the exact device and byte charge, reconciles it onto the final routing NIC before every post/retry, and releases it exactly once across success, failure, cancellation, fallback, and software-timeout paths. Baseline mode no longer mutates inflight accounting, and EWMA only learns from successful current-attempt latency. CI is green.
> 
> suggestion (non-blocking): add focused quota-accounting tests for estimate-to-exact reconciliation, fallback migration, retry recharge, and baseline mode so these invariants remain pinned independently of hardware tests.

@catyans thanks for the review! Added the quota-accounting unit tests you suggested (`tent_quota_accounting_test`), covering estimate→exact reconciliation, fallback migration, retry recharge, and baseline mode, plus EWMA gating and the charge-failure path — 10/10 passing.

To exercise the real reconcile logic hardware-independently, I made `Workers::chargeSliceQuota` / `releaseSliceQuota` `static` (taking `DeviceSelector*`) and added a `friend` test peer. They stay `private` members of `Workers`.

mooncake-transfer-engine/tent/tests/quota_accounting_test.cpp
staryxchen, line 1: The new tests lock in charge/release math through the friend accessor, but not the production call sites that actually broke (retry recharge in generatePostPath, software-timeout release, CQ EWMA gating). A slice that goes through generatePostPath / timeout / failed WC would still catch a regression if someone later removes those calls.

mooncake-transfer-engine/tent/tests/quota_accounting_test.cpp
```diff
@@ -0,0 +1,310 @@
+// Copyright 2025 KVCache.AI
```
staryxchen:
```suggestion
// Copyright 2026 KVCache.AI
```

mooncake-transfer-engine/tent/tests/quota_accounting_test.cpp
anranxia: Good point — the unit test pins the reconcile math via the friend accessor but not the three call sites (retry recharge in generatePostPath, software-timeout release, CQ EWMA gating). Those all run deep inside Workers and need a live RdmaTransport + real RDMA devices to drive, which is exactly why the helpers were extracted as static (so at least the math is hardware-independent). The call sites themselves are one-line delegations and are exercised by the 2-node RoCE e2e (tebench) and tent_rdma_transport_test on hardware CI. Happy to add a hardware-gated integration assertion alongside the RDMA transport tests if you'd prefer that over relying on e2e — let me know.

staryxchen (approved):
LGTM

mooncake-transfer-engine/tent/tests/quota_accounting_test.cpp
staryxchen: > Good point — the unit test pins the reconcile math via the friend accessor but not the three call sites (retry recharge in generatePostPath, software-timeout release, CQ EWMA gating). Those all run deep inside Workers and need a live RdmaTransport + real RDMA devices to drive, which is exactly why the helpers were extracted as static (so at least the math is hardware-independent). The call sites themselves are one-line delegations and are exercised by the 2-node RoCE e2e (tebench) and tent_rdma_transport_test on hardware CI. Happy to add a hardware-gated integration assertion alongside the RDMA transport tests if you'd prefer that over relying on e2e — let me know.

Got it. Will merge after CI pass

PR #3514, opened 2026-08-18
[Bugfix][TENT] Keep lazyFreeBatch sweeping past a batch it cannot reclaim

SongOf:
## Description

  `lazyFreeBatch()` returned on the first `getTransferStatus` / `retireQueueForBatch` error, so a batch that could not be reclaimed stranded every batch queued behind it in the insertion-ordered `freelist`; a permanent error (e.g. a transport that keeps failing the poll) leaked those `Batch` slabs and SubBatches forever. Record the first error, skip that batch, finish the sweep, then return it.

  ## Module

  ## Type of Change

  - [x] Bug fix

  ## How Has This Been Tested?

  Added `BatchLifecycle.FreeListSweepSkipsAStuckBatchInsteadOfStopping` (FakeTransport with a permanently failing poll on one batch); it fails on main and passes with the fix.

  **Test commands:**
  ```bash
  cmake --build build-tent --target tent_progress_worker_test
  ctest --test-dir build-tent/mooncake-transfer-engine/tent/tests --output-on-failure

  Test results:
  - [x] Unit tests pass (TENT ctest 38/38)

  Checklist

  AI Assistance Disclosure

  - [ ] No AI tools were used

### DIFF
--- a/mooncake-transfer-engine/tent/include/tent/metrics/tent_metrics.h
+++ b/mooncake-transfer-engine/tent/include/tent/metrics/tent_metrics.h
@@ -112,6 +112,12 @@ class TentMetrics {
     // distinguishable from genuine MLU samples in the histogram above.
     void recordDeadlineInfeasible(TransportType tp);
 
+    // Record a batch abandoned by lazyFreeBatch after repeated failed reclaim
+    // attempts. Such a batch stays on the freelist without further retries and
+    // is only reclaimed at engine teardown, so a nonzero value means resources
+    // are parked until shutdown and the last reclaim error is in the log.
+    void recordBatchQuarantined();
+
     enum class Stage {
         QueueWait,
         Dispatch,
@@ -224,6 +230,12 @@ class TentMetrics {
         "tent_deadline_infeasible_total",
         "Transfers whose deadline was already in the past at submit",
         kTransportLabel};
+    // Label-less: quarantine is a batch-lifecycle event, not a per-transport
+    // one (a batch may hold SubBatches on several transports).
+    ylt::metric::counter_t quarantined_batches_total_{
+        "tent_quarantined_batches_total",
+        "Batches abandoned by lazyFreeBatch after repeated failed reclaim "
+        "attempts; reclaimed only at engine teardown"};
 
     // Histograms - paired with their bucket boundaries in a single vector so
     // the two cannot drift out of sync (ylt histogram doesn't expose its
@@ -457,6 +469,14 @@ class ScopedLatencyRecorder {
         }                                                                      \
     } while (0)
 
+#define TENT_RECORD_BATCH_QUARANTINED()                   \
+    do {                                                  \
+        if (::mooncake::tent::TentMetrics::isEnabled()) { \
+            ::mooncake::tent::TentMetrics::instance()     \
+                .recordBatchQuarantined();                \
+        }                                                 \
+    } while (0)
+
 #define TENT_SCOPED_READ_LATENCY(tp, bytes)                               \
     ::mooncake::tent::ScopedLatencyRecorder _tent_latency_recorder_(      \
         ::mooncake::tent::ScopedLatencyRecorder::OperationType::Read, tp, \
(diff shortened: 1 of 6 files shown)


he-yufeng:
The sweep fix itself reads right: skipping a stranded batch instead of returning early is the correct direction for an insertion-ordered freelist, and the PoisonedPollTransport test pins the head-of-line blocking case.

One hole in the other half of the change though: all three lazyFreeBatch call sites in progress_worker.cpp discard the return value ((void) casts), so the first_error this PR now carefully reports is never observed by anyone. A poisoned batch will be re-polled on every pass forever with zero log lines. Either log a rate-limited warning at the call site on non-OK, or note in the PR why the returned status is meant for future callers.

SongOf:
> The sweep fix itself reads right: skipping a stranded batch instead of returning early is the correct direction for an insertion-ordered freelist, and the PoisonedPollTransport test pins the head-of-line blocking case.
> 
> One hole in the other half of the change though: all three lazyFreeBatch call sites in progress_worker.cpp discard the return value ((void) casts), so the first_error this PR now carefully reports is never observed by anyone. A poisoned batch will be re-polled on every pass forever with zero log lines. Either log a rate-limited warning at the call site on non-OK, or note in the PR why the returned status is meant for future callers.

@he-yufeng Good catch. Rather than logging only at the worker's call sites, the skip is now logged inside lazyFreeBatch itself (LOG_EVERY_N(WARNING, 100), with the batch pointer and status) — freeBatch's first-call path also drops the return value, so a sync-polling caller without the worker would otherwise stay just as silent. The three (void) sites in progress_worker.cpp now carry a comment saying the status is intentionally dropped there and why. Verified the line fires once in the poisoned-batch test.

alogfans (approved):
Looks good to merge overall — this fixes the core head-of-line blocking issue in `lazyFreeBatch()`, and the added test covers the key regression scenario. One follow-up worth considering is better handling/observability for batches that remain permanently unreclaimable, so they don’t sit on the freelist retrying and warning forever.

SongOf:
> Looks good to merge overall — this fixes the core head-of-line blocking issue in `lazyFreeBatch()`, and the added test covers the key regression scenario. One follow-up worth considering is better handling/observability for batches that remain permanently unreclaimable, so they don’t sit on the freelist retrying and warning forever.

Agreed, this suggestion is worth following up on.

SongOf:
> Looks good to merge overall — this fixes the core head-of-line blocking issue in `lazyFreeBatch()`, and the added test covers the key regression scenario. One follow-up worth considering is better handling/observability for batches that remain permanently unreclaimable, so they don’t sit on the freelist retrying and warning forever.

@alogfans Followed up in https://github.com/kvcache-ai/Mooncake/commit/c3d9d662287cdd643971e6228fc83f6f267940f2, in two parts:

Handling — quarantine, not force-free. After kMaxReclaimAttempts (4096) consecutive failed reclaim attempts, the batch is marked abandoned: the sweep stops polling and warning about it, and a single final LOG(ERROR) records the last error and that engine teardown will reclaim it (deconstruct() already frees the freelist unconditionally, so the quarantined batch has a guaranteed exit). The failure counter only advances on error passes and resets when a poll recovers to PENDING, so an in-flight batch can never be quarantined — only a permanently failing poll or queue retire reaches the threshold. Force-freeing after a timeout was deliberately rejected: a failing poll is exactly the case where we cannot prove no DMA is in flight.

Observability — a dedicated counter. Quarantine events increment a new tent_quarantined_batches_total counter, exposed on /metrics (Prometheus), /metrics/json, and the periodic summary line. It is label-less on purpose — quarantine is a batch-lifecycle event, and a batch may hold SubBatches on several transports. Nonzero means resources are parked until shutdown and the root cause is in the log; worth alerting on.

Covered by BatchLifecycle.PermanentlyStuckBatchIsQuarantinedNotRetriedForever (drives 4096 real sweep passes, asserts polling and warnings stop, later batches still reclaim, and teardown frees the quarantined one) and MetricsRecordingTest.QuarantinedBatchIncrementsCounter (asserts the counter increments exactly once on both endpoints). Verified on both TENT_METRICS_ENABLED=ON and OFF builds — 38/38 TENT ctest each.

PR #3516, opened 2026-08-18
[Store] Restore HA snapshot compatibility after producer-view split

bitborne:
## Description

This fixes the HA snapshot consumers that were left on the pre-S03 data layout after the producer-view fencing change was merged.

- Reads the producer view from the independent `/oplog/<cluster>/producer_view` key instead of `DurablePrefix`.
- Keeps the producer view paired with the durable prefix when that prefix becomes fully applied, so later view changes cannot relabel an older snapshot boundary.
- Preserves zero producer-view metadata when the independent key is absent, which keeps prefixes written before producer-view fencing readable.
- Fails closed on malformed producer-view values and propagates backend errors through the existing standby poll disposition.
- Removes the obsolete per-object `last_sequence_id` restore check and its test because N02 now uses one global applied OpLog cursor.
- Updates the standby reader and snapshot-capture tests to use the S03 key layout.

This is a compatibility/build fix for the merged HA changes in #3326 and #3384. It does not add a new snapshot protocol or change the producer-view fencing contract. RFC #3167 remains the design reference.

The open N03 artifact-writer work in #3447 is separate: it consumes the capture object, while this patch repairs the capture object's existing producer-view source and the stale restore consumer before that work is rebased.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**

```bash
cmake -S . -B /tmp/mooncake-ha-build -G Ninja \
  -DBUILD_UNIT_TESTS=ON \
  -DBUILD_EXAMPLES=OFF \
  -DBUILD_BENCHMARK=OFF \
  -DWITH_TE=OFF \
  -DWITH_STORE_RUST=OFF \

cmake --build /tmp/mooncake-ha-build \
  --target mooncake_store \
  oplog_batch_standby_reader_test \
  hot_standby_service_test \
  master_service_ha_test -j32

LSAN_OPTIONS=detect_leaks=0 \
  /tmp/mooncake-ha-build/mooncake-store/tests/oplog_batch_standby_reader_test

clang-format-20 --dry-run --Werror \
  mooncake-store/include/ha/oplog/oplog_batch_standby_reader.h \
  mooncake-store/include/ha/oplog/oplog_batch_storage.h \
  mooncake-store/src/ha/oplog/oplog_batch_standby_reader.cpp \
  mooncake-store/src/hot_standby_service.cpp \
  mooncake-store/tests/ha/master_service_ha_test.cpp \
  mooncake-store/tests/ha/oplog/oplog_batch_standby_reader_test.cpp \
  mooncake-store/tests/ha/standby/hot_standby_service_test.cpp

pre-commit run --files <all changed files>
```

**Test results:**

- [x] `mooncake_store` and the three HA test targets compile successfully with `-j32`.
- [x] `oplog_batch_standby_reader_test`: 23/23 passed, including detached producer-view lookup, malformed values, missing-key compatibility, and applied-prefix/view pairing.
- [x] `master_service_ha_test`: the 9 restore-focused tests pass.
- [x] clang-format check passed.
- [x] Pre-commit 4.6.0 passed on all changed files.

## Checklist

- [x] I have formatted my code using clang-format 20 and the repository's C/C++ pre-commit hook
- [x] I have updated the documentation (not applicable: this restores internal HA build/runtime compatibility)

## AI Assistance Disclosure

- [ ] No AI tools were used
- [x] AI tools were used

Codex helped review the HA data-layout change, implement the compatibility fix, update tests, and run verification.

### DIFF
--- a/mooncake-store/include/ha/oplog/oplog_batch_standby_reader.h
+++ b/mooncake-store/include/ha/oplog/oplog_batch_standby_reader.h
@@ -26,6 +26,8 @@ struct OpLogBatchStandbyPollResult {
     bool durable_prefix_present{false};
     size_t applied_entries{0};
     DurablePrefix durable_prefix{};
+    // Producer view is stored independently from the durable prefix.
+    ViewVersionId producer_view_version{0};
 };
 
 class OpLogBatchStandbyReader {
@@ -37,6 +39,9 @@ class OpLogBatchStandbyReader {
     std::optional<DurablePrefix> GetLastAppliedDurablePrefix() const {
         return last_applied_durable_prefix_;
     }
+    std::optional<ViewVersionId> GetLastAppliedProducerViewVersion() const {
+        return last_applied_producer_view_version_;
+    }
 
    private:
     OpLogBatchStorage storage_;
@@ -44,6 +49,7 @@ class OpLogBatchStandbyReader {
     bool batch_format_seen_{false};
     std::optional<DurablePrefix> last_observed_prefix_;
     std::optional<DurablePrefix> last_applied_durable_prefix_;
+    std::optional<ViewVersionId> last_applied_producer_view_version_;
     std::optional<uint64_t> last_scanned_batch_last_seq_;
     uint64_t last_applied_batch_id_{0};
 };
--- a/mooncake-store/include/ha/oplog/oplog_batch_storage.h
+++ b/mooncake-store/include/ha/oplog/oplog_batch_storage.h
@@ -16,6 +16,7 @@ class OpLogBatchStorage {
     ErrorCode InitDurablePrefix(DurablePrefix& prefix);
     ErrorCode ClaimProducerView(ViewVersionId producer_view_version);
     ErrorCode ValidateProducerView(ViewVersionId producer_view_version) const;
+    ErrorCode ReadProducerView(ViewVersionId& producer_view_version) const;
     ErrorCode ReadDurablePrefix(DurablePrefix& prefix);
     ErrorCode WriteBatchAndAdvancePrefix(const OpLogBatchRecord& batch,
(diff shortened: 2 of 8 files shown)


he-yufeng (commented):
Verified the pairing invariant: last_applied_durable_prefix_ and last_applied_producer_view_version_ are now written together at exactly the two PollOnce sites (bootstrap observe, completed catch-up), so the hot_standby_service consistent check can never see one without the other. The snapshot correctly stamps the last APPLIED view rather than the currently observed one, which is what stops a mid-fence view change from relabeling an older boundary. The legacy path (missing producer_view key treated as view 0, malformed values failing closed through the poll error) reads right.

One note for the record: the bootstrap branch now also requires !last_applied_durable_prefix_, so a re-observed initial prefix can no longer overwrite the recorded bootstrap boundary. That is the right direction; just worth knowing it changes behavior on a repeat initial observe.

Icedcoco:
Thanks for the detailed analysis and test coverage. After checking the downstream snapshot, replay, and fencing paths, `producer_view_version` does not need to identify the view that originally wrote the captured durable prefix. A snapshot taken in view 8 at a complete prefix written in view 7 is still valid: replay is determined by the batch/sequence cursor, while writer fencing is enforced separately by the producer-view transaction.

The compilation regression and detached producer-view lookup were already fixed in the earlier #3515, which has now been merged. Since the additional prefix/view caching is not required by the current snapshot contract, this PR is superseded by #3515 and can be closed.

PR #3520, opened 2026-08-19
[TENT] Route RDMA slices by QP pool

alogfans:
## Description

This PR fixes RDMA QP-pool routing so slices that request different `qp_pool`s are posted to the correct pool-specific QPs and therefore use the intended link-layer QoS parameters.

The change does three things:

1. Resolves per-QP link-layer QoS from the owning QP pool, so different pools can use different `service_level` / `traffic_class` values.
2. Routes slices to the worker that owns the QP selected for their requested pool.
3. Splits mixed-pool batches before `submitSlices()`, so a batch containing multiple `qp_pool`s does not inherit the first slice's pool.

For the traditional/default case where no QP pools are configured, the worker keeps the legacy single-submit path. The only added work on that path is checking whether QP-pool routing is enabled.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
git diff --check

cmake --build build --target tent_xport_rdma tent_qp_pool_layout_test tent_rdma_transport_test -j$(nproc)

ctest --test-dir build -R 'tent_qp_pool_layout_test|tent_rdma_transport_test' --output-on-failure
```

**Test results:**
- [x] Unit tests pass

`tent_qp_pool_layout_test` covers:
- two QP pools with different `service_level` / `traffic_class`
- per-field fallback to endpoint defaults
- routing a named pool to its owning worker
- grouping mixed-pool slices before submit

`tent_rdma_transport_test` passed on a machine with mlx5 RDMA devices.

Note: `pre-commit run --files ...` could not be run because `pre-commit` is not installed in the current environment.

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

AI tools were used to help isolate the patch, structure the implementation, add focused tests, and run local verification. The human submitter should review and understand every changed line before submitting.

### DIFF
--- a/mooncake-transfer-engine/tent/src/transport/rdma/workers.cpp
+++ b/mooncake-transfer-engine/tent/src/transport/rdma/workers.cpp
@@ -26,10 +26,12 @@
 #include <sstream>
 #include <string>
 #include <thread>
+#include <unordered_set>
 
 #include "tent/transport/rdma/bw_arbitration.h"
 #include "tent/transport/rdma/endpoint_store.h"
 #include "tent/transport/rdma/promotion_policy.h"
+#include "tent/transport/rdma/qp_pool_routing.h"
 #include "tent/transport/rdma/shared_quota.h"
 #include "tent/common/utils/ip.h"
 #include "tent/common/utils/string_builder.h"
@@ -379,6 +381,7 @@ void Workers::submitFromTick(WorkerContext& worker, RdmaSlice* slice) {
     if (slice && slice->task) {
         priority = slice->priority;
     }
+    bool wake_worker = false;
     // This lane's queue and counter are about to account for the slice, so
     // move the count here before it is visible on the queue. With a shared
     // queue pair the lane re-queueing the slice is the one that polled the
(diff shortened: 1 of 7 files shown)


catyans:
The QP-pool `service_level` and `traffic_class` values are accepted as unrestricted integers and then narrowed to `uint8_t` in `resolveQpLinkLayerQos()`. Unlike the selection-policy parser, this path does not enforce SL `0..15` or TC `0..255`. A value such as `traffic_class: 256` therefore silently becomes `0`, while an out-of-range SL may make QP setup fail instead of rejecting the configuration clearly.

Could we validate these fields while loading `qp_pools` (keeping `-1` as the fallback sentinel) and add boundary/out-of-range tests?

alogfans:
@catyans 

Good catch, fixed. `qp_pools` loading now validates `service_level` as `-1` or `0..15` and `traffic_class` as `-1` or `0..255`, returning `InvalidArgument` for invalid config before any `uint8_t` narrowing.

I also added boundary/out-of-range tests for the allowed sentinel and range limits. The targeted RDMA tests pass.

mooncake-transfer-engine/tent/src/transport/rdma/workers.cpp
```diff
@@ -393,15 +396,21 @@ void Workers::submitFromTick(WorkerContext& worker, RdmaSlice* slice) {
...
     // The worker must never block on its own queue (issue #3637): a full
-    // queue parks the slice in requeue_overflow and the next tick retries.
-    // Either way it stays counted as inflight, which keeps the worker from
-    // suspending while a parked flush is pending.
-    if (!worker.queues[priority].try_push(slice_list)) {
+    // local queue parks the slice in requeue_overflow and the next tick
+    // retries. A different worker is reached through the MPSC producer path
+    // and must not touch that worker's local overflow vector.
+    if (laneIndex(worker) != tl_wid) {
+        worker.queues[priority].push(slice_list);
```
staryxchen, line 406: Cross-worker QP-pool redirect can wedge the worker threads. `submitTransferTasks` still round-robins slices across workers, and `asyncPostSend` then forwards each slice to the worker that polls the selected QP. That forward goes through `queues[priority].push()`, which spins until a slot is free, and it enqueues one slot per slice.

The same-worker path uses `try_push` and parks in `requeue_overflow` (issue #3637). This path cannot touch another worker's overflow vector, so a full queue spins inside `asyncPostSend`. Each priority queue holds 8192 lists. A popped list can carry every slice of a batch, and those slices are then re-queued individually, so one tick can try to write far more than 8192 slots.

On the layout this PR tests (`kv` = QPs `[0,4)`, `ctrl` = `[4,6)`, 6 workers) the forwards form a cycle: worker 0 sends `ctrl` slices to worker 4, and worker 4 sends `kv` slices to worker 0. Both workers have already popped their queues for this tick. Once each side has pushed 8192 slices, both spin in `push` and neither returns to `pop`. `asyncPollCq` does not run, so completions for those workers' CQs sit unpolled and the QP `wr_depth` never drains. Slices still in the queue are not in `inflight_slice_set`, so `expireTimedOutSlices` will not fail them. `Workers::stop()` only observes `running_` at the top of the worker loop, so `join` hangs too.

Below a full queue this is a completion stall for the whole redirect burst. At capacity, with this cycle, transfers hang.

Could the redirect stay non-blocking? `try_push` toward the owner, and on a full queue park the slices where that worker can drain them later (one `RdmaSliceList` per owner, or a concurrent overflow), instead of spinning in `push`.

PR #3521, opened 2026-08-19
[Bugfix][Store] Drop stale per-object last_sequence_id snapshot cursor check

anranxia:
## Description

`main` currently fails to compile `mooncake_store` due to a merge-order race between two already-merged PRs:

- **#3354 ([Store] Return explicit standby restore errors)** added a per-object cursor check in `MasterService::RestoreFromStandbySnapshot`: `if (entry.metadata.last_sequence_id > initial_oplog_sequence_id) { ... return INVALID_VERSION; }`.
- **#3326 ([Store] Add bounded standby snapshot capture)** then removed `last_sequence_id` from `StandbyObjectMetadata` (and stopped populating it in `MetadataPayload::ToStandbyMetadata`) as part of bounded snapshot capture.

Each was green on its own branch, but together they leave `master_service.cpp` referencing a field that no longer exists:

```
mooncake-store/src/master_service.cpp:3046: error: 'const struct mooncake::StandbyObjectMetadata' has no member named 'last_sequence_id'
```

This blocks every full-repo build job (build-wheel, tent-ci, ...) on unrelated PRs.

## Fix
remove the now-unsupportable per-object cursor check and its obsolete test `RestoreRejectsObjectCursorBeyondSnapshot`. Re-adding the field is not the right fix — `ToStandbyMetadata` no longer populates it, so it would be a dead always-zero value; and bounded snapshot capture (#3326) already bounds captured objects to the snapshot cursor, so the per-object comparison is redundant by construction. `initial_oplog_sequence_id` is still used elsewhere in the function, so its signature is unchanged.

## Module

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

Restores `mooncake_store` compilation on `main` (the build break above is gone). Grepped the tree to confirm no other references to the removed `StandbyObjectMetadata::last_sequence_id` remain, and the obsolete test that depended on it was removed.

**Test results:**
- [ ] Unit tests pass
- [x] Manual testing done (build-break fix; CI full-repo build validates)

## Checklist

## AI Assistance Disclosure

Claude Code helped bisect the merge-order race between #3354 and #3326 and prepare the minimal fix. The human submitter has reviewed every changed line and can defend the change end-to-end.

### DIFF
--- a/mooncake-store/src/master_service.cpp
+++ b/mooncake-store/src/master_service.cpp
@@ -3043,15 +3043,6 @@ tl::expected<void, ErrorCode> MasterService::RestoreFromStandbySnapshot(
                 << tenant_id.value() << ", key=" << user_key;
             return tl::make_unexpected(ErrorCode::INVALID_PARAMS);
         }
-        if (entry.metadata.last_sequence_id > initial_oplog_sequence_id) {
-            LOG(ERROR) << "RestoreFromStandbySnapshot: object cursor "
-                       << entry.metadata.last_sequence_id
-                       << " exceeds snapshot cursor "
-                       << initial_oplog_sequence_id
-                       << ", tenant=" << tenant_id.value()
-                       << ", key=" << user_key;
-            return tl::make_unexpected(ErrorCode::INVALID_VERSION);
-        }
 
         const auto shard_idx = entry.metadata.group_id.empty()
                                    ? getShardIndex(tenant_id, user_key)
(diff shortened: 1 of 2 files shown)


he-yufeng:
Verified on current main (12928dde): master_service.cpp:3046 still references entry.metadata.last_sequence_id, which #3326 removed from StandbyObjectMetadata, so the tree does not compile right now. This PR is the minimal un-break: delete the stale check and its test, both of which only exist for the removed field.

For the record, this overlaps #3515 (broader, also touches the producer-view storage/reader side) and #3516 (consumer-side pairing). Landing order that keeps each reviewable: this one first to un-break main, then #3516's consumer pairing, then whatever remains of #3515 after a rebase.

Icedcoco:
Thanks for investigating this and preparing the minimal fix. The earlier #3515 has now been merged and includes the same removal of the stale `last_sequence_id` check and its obsolete test, along with the related producer-view compatibility fix. This PR is therefore fully superseded by #3515 and can be closed.

PR #3526, opened 2026-08-19
[TENT] Bound the pending-batch drain wait in ProxyManager shutdown

jfeng18:
## Description

`ProxyManager::deconstruct()` (introduced in #3380) waits for in-flight staging batches to reach a terminal state with no upper bound:

```cpp
while (has_pending_batches()) {
    std::this_thread::sleep_for(std::chrono::milliseconds(1));
    flushPendingCleanups(false);
}
```

On the normal path the transport timeout drives every batch to a terminal state and this loop exits quickly. But if a peer dies and the transport timeout never fires for a batch, the loop spins forever and process shutdown hangs. This PR bounds the drain wait with a 30s deadline (`std::chrono::steady_clock`); once the deadline passes we log a warning and fall through to the remaining-path handling that #3380 already ships (leave unresolved cleanups pinned with a warning and skip releasing them). This is liveness hardening, not a fix for a reproducible bug: when the transport timeout works, the loop exits far before the deadline, so there is no behavior change on the normal path.

### Remaining-path guard extension

The keep-pinned guard in the remaining path only checked `!pending.remote_operations.empty()`. With the original unbounded loop that was sufficient: the loop guaranteed `pending.batches` was empty by the time the remaining path ran, so the batches-nonempty case was unreachable. The new timeout makes that case reachable, and without a guard the remaining path would unpin remote stage buffers while a non-terminal cross-stage batch targeting them could still be in flight against a live peer — the slab-reuse-under-DMA class that #3380 was fixing. The guard is therefore extended to `!pending.remote_operations.empty() || !pending.batches.empty()`. This is a no-op for every previously reachable path (batches were always empty there) and only covers the state newly reachable after the timeout.

## Module

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [x] Other: liveness hardening

## How Has This Been Tested?

- `clang-format --dry-run --Werror` clean on the touched file.
- CI. The bounded-wait path is a shutdown corner case (batch stuck non-terminal past the transport timeout); regular CI exercises the normal `deconstruct()` exit, which is unchanged — the loop still drains batches to terminal state and the deadline never fires when the transport timeout works.

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings

## AI Assistance Disclosure

This change was developed with AI assistance; every changed line has been reviewed by the submitter.

### DIFF
--- a/mooncake-transfer-engine/tent/src/runtime/proxy_manager.cpp
+++ b/mooncake-transfer-engine/tent/src/runtime/proxy_manager.cpp
@@ -115,7 +115,16 @@ Status ProxyManager::deconstruct() {
     };
 
     flushPendingCleanups(false);
+    constexpr auto kShutdownDrainTimeout = std::chrono::seconds(30);
+    const auto drain_deadline =
+        std::chrono::steady_clock::now() + kShutdownDrainTimeout;
     while (has_pending_batches()) {
+        if (std::chrono::steady_clock::now() >= drain_deadline) {
+            LOG(WARNING) << "Timed out waiting for in-flight staging batches "
+                            "to reach a terminal state during shutdown; "
+                            "leaving the remaining cleanups pinned";
+            break;
+        }
         std::this_thread::sleep_for(std::chrono::milliseconds(1));
         flushPendingCleanups(false);
     }
@@ -129,6 +138,20 @@ Status ProxyManager::deconstruct() {
         has_pending_cleanups_.store(false, std::memory_order_relaxed);
     }
     for (auto& pending : remaining) {
+        if (!pending.batches.empty()) {
+            // The drain deadline above makes non-terminal batches reachable
+            // here; their transport workers may still complete and touch the
+            // remote stage buffers, so keep everything pinned (the deferred
+            // unpin path below assumes the batches already drained).
+            LOG(WARNING) << "Leaving " << pending.remote_addrs.size()
+                         << " remote stage buffers pinned because "
+                         << pending.remote_operations.size()
+                         << " remote staging requests and "
+                         << pending.batches.size()
+                         << " staging batches have not reached a confirmed "
+                            "terminal state";
+            continue;
+        }
         size_t deferred_remote_buffers = 0;
         for (auto& operation : pending.remote_operations) {
             if (!operation) continue;
--- a/mooncake-transfer-engine/tent/src/runtime/proxy_manager.cpp
+++ b/mooncake-transfer-engine/tent/src/runtime/proxy_manager.cpp
@@ -24,7 +24,11 @@ namespace mooncake {
 namespace tent {
 ProxyManager::ProxyManager(TransferEngineImpl* impl, size_t chunk_size,
                            size_t chunk_count)
-    : chunk_size_(chunk_size), chunk_count_(chunk_count), impl_(impl) {
+    : chunk_size_(chunk_size),
+      chunk_count_(chunk_count),
+      impl_(impl),
+      shutdown_drain_timeout_(std::chrono::milliseconds(
+          impl->conf_->get("staging/shutdown_drain_timeout_ms", 30000L))) {
     running_ = true;
     for (size_t i = 0; i < kShards; ++i) {
         shards_[i].thread = std::thread(&ProxyManager::runner, this, i);
@@ -115,9 +119,8 @@ Status ProxyManager::deconstruct() {
     };
 
     flushPendingCleanups(false);
-    constexpr auto kShutdownDrainTimeout = std::chrono::seconds(30);
     const auto drain_deadline =
-        std::chrono::steady_clock::now() + kShutdownDrainTimeout;
+        std::chrono::steady_clock::now() + shutdown_drain_timeout_;
     while (has_pending_batches()) {
         if (std::chrono::steady_clock::now() >= drain_deadline) {
             LOG(WARNING) << "Timed out waiting for in-flight staging batches "
@@ -131,6 +134,8 @@ Status ProxyManager::deconstruct() {
 
     flushPendingCleanups(false);
     std::vector<PendingBufferCleanup> remaining;
+    std::vector<BatchID> deferred_batches;
+    std::vector<uint64_t> deferred_local_addrs;
     {
         std::lock_guard<std::mutex> lk(pending_cleanups_mu_);
         remaining = std::move(pending_cleanups_);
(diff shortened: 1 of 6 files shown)


he-yufeng (commented):
Read this against the current deconstruct() flow. The hang risk is real: the wait loop only exits once every batch reaches a terminal state, and a dead peer can keep a batch from ever getting there. The deadline break is the right shape.

The part that matters more to me is the second hunk. On the timeout path a pending entry can carry in-flight batches with an empty remote_operations list, and the old condition let exactly those fall into the unpin loop, dropping remote pins for a stage operation that never confirmed. Gating on batches non-empty closes that hole, and on the normal path (has_pending_batches false) every entry has empty batches, so nothing changes there.

One thing I walked through so the next reader does not have to: local stage buffers are still unregistered and freed after a timeout break, but with the shard threads joined and the peer presumed gone, a late RDMA write lands on a deregistered MR and fails at the NIC rather than touching freed memory. That is the safe direction for this failure.

Leaving this as a comment rather than an approval since it is a shutdown path with shared state; would like a second pair of eyes from someone with a Linux build before it lands.

jfeng18:
Thanks for the thorough walkthrough @he-yufeng — especially for independently tracing the empty-remote_operations-with-in-flight-batches case through the old condition; that's exactly the hole the second hunk closes, and it's good to have the local-buffer direction (NIC-level failure on a deregistered MR, not freed-memory corruption) written down for future readers.

For the second pair of eyes: @doujiang24 @alogfans — would either of you have bandwidth for a look? The change is confined to `ProxyManager::deconstruct()` (shutdown path, +17/-7).

mooncake-transfer-engine/tent/src/runtime/proxy_manager.cpp
```diff
@@ -115,7 +115,16 @@ Status ProxyManager::deconstruct() {
...
+    constexpr auto kShutdownDrainTimeout = std::chrono::seconds(30);
+    const auto drain_deadline =
+        std::chrono::steady_clock::now() + kShutdownDrainTimeout;
     while (has_pending_batches()) {
+        if (std::chrono::steady_clock::now() >= drain_deadline) {
+            LOG(WARNING) << "Timed out waiting for in-flight staging batches "
+                            "to reach a terminal state during shutdown; "
+                            "leaving the remaining cleanups pinned";
+            break;
+        }
```
alogfans, line 130: [P1] Reaching the shutdown deadline does not guarantee that the underlying DMA or RPC operation has stopped. Although this path keeps the remote stage buffers pinned, `ProxyManager::deconstruct()` still unregisters and frees the local stage buffers, and `TransferEngineImpl::deconstruct()` subsequently frees all Batch/SubBatch/Slice objects. A late completion may therefore access freed memory or batch state.

Please cancel and quiesce the transports before continuing teardown, or transfer ownership of the unfinished batches and both local and remote buffers to a cleanup object that remains alive until transport shutdown is confirmed. The timeout path should also have a regression test.

jfeng18: Confirmed, and thanks for catching this — traced the exact teardown order to verify:

`TransferEngineImpl::deconstruct()` destroys the ProxyManager first, then frees every Batch/SubBatch/Slice back to the Slab (L487-505), and only **after that** resets the transports (L508) — which is where RDMA `Workers` get joined. So after a timeout break, `Workers::asyncPollCq()` can still write status into a slice that has already been returned to the Slab: a host-side UAF, exactly as you said. (he-yufeng's point about the DMA direction also holds — `ibv_dereg_mr` invalidates the lkey so the NIC fails the WQE instead of writing host memory — the two observations are about different layers.)

**Proposed fix (ownership transfer)**: on timeout, `deconstruct()` stops freeing the local stage buffers and instead moves the undrained `pending_cleanups_` plus the affected `stage_buffers_` entries into a deferred-teardown holder owned by `TransferEngineImpl`, which releases them only **after** `transport.reset()` — i.e. after Workers are joined and the CQ fully drained, so every slice has reached a terminal state before any backing memory or batch object is freed. Remote pins stay held as they do today.

I looked at cancel+quiesce first, but `cancelTransfer` currently returns NotImplemented for staging transfers, and cancel is best-effort anyway (posted WQEs still complete through the CQ) — so ownership transfer is the direction that doesn't depend on new cancellation plumbing. It also matches existing precedent (mooncake-pg's `resource_abandoned_` drain-fail path, the etcd coordinator's intentional leak on timeout). Once staging cancellation is wired up, it can be added to shorten the drain wait.

**Regression test**: a FakeTransport (per `fault_proxy_test.cpp`) whose batch never leaves PENDING, asserting deconstruct() returns within the bound and ASAN-clean teardown.

Does this direction work for you? If so I'll update the PR.

jfeng18: Implemented in b185ca2a:

- **Ownership transfer**: on deadline expiry, `deconstruct()` hands the undrained batches plus the stage-buffer arenas their `local_addrs` actually hit (address-range matched; untouched arenas are freed as before) to `TransferEngineImpl::adoptDeferredStageTeardown()`, which detaches the batches from `batch_set_.active` / `freelist` / `alive_batches_` under `progress_mutex_` so the deconstruct sweep no longer returns their SubBatch/Slice to the Slab early.
- **Release point**: the deferred holder is only touched after `transport.reset()` joins the workers and drains the CQs. One caveat surfaced during implementation: at that point a *clean* reclaim is impossible — `freeSubBatch()` is a virtual on the just-destroyed transports, and `freeLocalMemory()/unregisterLocalMemory()` route through raw `Transport*` captured in `allocated_memory_`. So the bitmap (plain host memory, staging workers already joined) is deleted, and the chunks + batch shells are intentionally abandoned to process exit, kept reachable from a static registry so LSan stays quiet — same pattern as mooncake-pg's `resource_abandoned_` drain-fail path. MR deregistration itself is already covered by the `local_segment_tracker_` sweep while transports are alive.
- **Regression test**: `tent_proxy_shutdown_timeout_test` with a never-completing fake transport; the timeout is now `staging/shutdown_drain_timeout_ms` (default 30000, unchanged behavior) so the test runs with 200ms instead of waiting 30s. Asserts bounded destruction, both batches adopted into the deferred holder, and zero `freeLocalMemory` calls for the touched arena (discriminates against the old early-free path).

Normal-path behavior is unchanged (empty deferred holder short-circuits). PTAL when you get a chance.

alogfans (approved):
LGTM. Before merging you should fix conflicts and pass all CI checks.

jfeng18:
Rebased onto latest main (ca94b6b3) to resolve a semantic overlap with #3494, which rewrote the same remaining-path unpin logic. Flagging the one intentional change so you don't have to diff it blind:

#3494 assumes `pending.batches` is empty when control reaches the remaining path (upstream still has the unbounded drain loop). This PR's deadline makes `batches` non-empty reachable, so I split the remaining path by state instead of merging the two guards:

- **`batches` non-empty** (only reachable via timeout): our conservative branch — no abandon, no unpin of any remote pin; collect batches + local_addrs into the deferred-teardown holder. This preserves the "don't touch remote pins on timeout" contract we discussed.
- **`batches` empty, `operations` non-empty**: #3494 fully preserved (abandonForCleanup + async unpin on late completion — exactly the case it fixes).
- **both empty**: #3494's immediate-unpin path, unchanged.

Net effect: the guard narrowed from `!remote_operations.empty() || !batches.empty()` to just `!pending.batches.empty()`, because #3494's ownership-transfer mechanism now handles the `remote_operations` case better than the blanket keep-pinned did (no leak). The two fixes are orthogonal — #3494 covers batches-empty + late completion, this covers batches-non-empty + timeout. PTAL to confirm the split reads right on your end.

Issue #3496, opened 2026-08-18
[Bug]: HotStandbyService tests do not exercise claimed etcd behaviors


Aionw:
## Bug Report

Several `HotStandbyService` tests are named after etcd-backed standby behaviors but do not execute those production paths.

With `STORE_USE_ETCD=ON`, the affected tests unconditionally call `GTEST_SKIP()` because they require external etcd infrastructure. With the default `STORE_USE_ETCD=OFF`, `HotStandbyService::Start()` returns `INTERNAL_ERROR` at the compile-time backend guard before endpoint validation, connection, synchronization, replication, or verification logic can run. The non-etcd branches therefore either assert the same unconditional error or only call `Stop()`/`SUCCEED()`.

Affected intended behaviors include:

- etcd-backed Start and invalid endpoint handling
- transition from etcd-backed Start to WATCHING
- connection and synchronization failures
- sync status after synchronization
- warm-start variants
- replication-loop metrics and disconnect handling
- verification-loop enabled/disabled behavior

These scenarios are valuable, but the current test implementations can give a false impression that they are covered.

## Expected coverage

- Use deterministic fake/test seams where they can drive the real named production behavior.
- Use an etcd-enabled integration fixture for behavior that necessarily depends on etcd.
- Assert observable state, status, metrics, and failure results rather than only checking for no crash.
- Keep non-etcd feature-gate coverage accurately named and separate from etcd behavior coverage.
- Avoid adding public production APIs solely to expose private implementation details.

PR #3495 adds deterministic snapshot-only coverage for the already-running behavior and preserves the remaining etcd-dependent test cases as explicit skipped coverage gaps linked to this issue, rather than allowing them to pass without exercising the claimed behavior.

## Acceptance criteria

- Each retained scenario executes the corresponding production path or is replaced by equivalent behavior coverage.
- Tests run deterministically in an appropriate unit or integration configuration.
- The tests no longer unconditionally skip in their intended build configuration.
- Assertions distinguish endpoint, connection, synchronization, replication, and verification failures.

PR #3527, opened 2026-08-19
fix(store): reject cross-shard duplicates during standby restore

### LINKED ISSUE #3398: [Bugfix] Standby restore misses duplicate objects routed to another shard
## Description

`MasterService::RestoreFromStandbySnapshot()` can insert duplicate metadata for the same `(tenant_id, key)` when an existing object and the incoming standby object have different `group_id` values.

The restore path computes the target shard exclusively from the incoming metadata:

```cpp
const auto shard_idx = entry.metadata.group_id.empty()
                           ? getShardIndex(tenant_id, user_key)
                           : getShardIndex(entry.metadata.group_id);
```

The existing-object check then searches only that target shard. If the existing object is routed by its key or by another group, it can live in a different shard and the check misses it.

This behavior predates PR #3354. That PR adds explicit duplicate validation, but the new validation does not cover this cross-shard case.

## Reproduction

1. Restore an ungrouped object such as `(default, existing)`.
2. Restore the same tenant/key again with `group_id = "group"`, choosing a group that hashes to a different shard.
3. Expect the second restore to return `OBJECT_ALREADY_EXISTS`.
4. The second restore currently succeeds and inserts another metadata entry.

A minimal test follows this shape:

```cpp
auto existing = MakeStandbyObject("existing", endpoint);
ASSERT_TRUE(service.RestoreFromStandbySnapshot(
    {existing}, 7, {MakeStandbyMemorySegment(endpoint)}));

auto duplicate = MakeStandbyObject("existing", endpoint);
duplicate.metadata.group_id = "group";
auto result = service.RestoreFromStandbySnapshot(
    {duplicate}, 7, {MakeStandbyMemorySegment(endpoint)});

EXPECT_FALSE(result.has_value());
EXPECT_EQ(result.error(), ErrorCode::OBJECT_ALREADY_EXISTS);
```

The expectation fails because `result.has_value()` is true.

## Impact

Two metadata entries may exist for one logical object identity. The group-routing index points to only one shard, potentially leaving the other entry unreachable through normal key lookup while it still contributes metadata, replicas, and metrics.

Although promotion normally constructs a fresh `MasterService`, this violates the restore API's duplicate-rejection and failure-atomicity expectations and affects repeated or non-empty-state restores.

## Suggested fix

Resolve the existing object's shard from the current routing index independently of the incoming `group_id`, and reject the restore if the identity already exists. Add a regression test covering ungrouped-to-grouped and grouped-to-different-group restores.

## Context

Found while reviewing PR #3354. This issue is separate from the segment name/transport-endpoint alias model.

982945902:
## Summary

- reject standby snapshot entries when the same `(tenant, key)` already exists under the current metadata routing
- perform the duplicate check before restoring metrics, segment state, or routing metadata
- add regression coverage for ungrouped-to-grouped and grouped-to-different-group cross-shard duplicates

## Root cause

`RestoreFromStandbySnapshot()` checked only the shard selected by the incoming snapshot entry's `group_id`. If an existing object was routed by its key or by another group, restore could insert duplicate metadata into a different shard.

## Tests

- `MasterServiceHATest.RestoreRejectsUngroupedObjectDuplicatedIntoAnotherShard`
- `MasterServiceHATest.RestoreRejectsGroupedObjectDuplicatedIntoAnotherGroupShard`
- all `MasterServiceHATest.Restore*` tests: 11 passed
- full `master_service_ha_test`: 83 passed, 1 unrelated failure (`MasterServiceBatchRecordE2ETest.PromotionCatchesUpToDurablePrefix`) because the local build used `STORE_USE_ETCD=OFF`; tracked by #3496
- `git diff --check`

Fixes #3398

### DIFF
--- a/mooncake-store/src/master_service.cpp
+++ b/mooncake-store/src/master_service.cpp
@@ -3043,6 +3043,19 @@ tl::expected<void, ErrorCode> MasterService::RestoreFromStandbySnapshot(
                 << tenant_id.value() << ", key=" << user_key;
             return tl::make_unexpected(ErrorCode::INVALID_PARAMS);
         }
+        const size_t existing_shard_idx =
+            getMetadataShardIndex(tenant_id, user_key);
+        {
+            MetadataShardAccessorRO existing_shard(this, existing_shard_idx);
+            auto existing_tenant = existing_shard->tenants.find(tenant_id);
+            if (existing_tenant != existing_shard->tenants.end() &&
+                existing_tenant->second.metadata.contains(user_key)) {
+                LOG(ERROR)
+                    << "RestoreFromStandbySnapshot: object already exists, "
+                    << "tenant=" << tenant_id.value() << ", key=" << user_key;
+                return tl::make_unexpected(ErrorCode::OBJECT_ALREADY_EXISTS);
+            }
+        }
         const auto shard_idx = entry.metadata.group_id.empty()
                                    ? getShardIndex(tenant_id, user_key)
                                    : getShardIndex(entry.metadata.group_id);
--- a/mooncake-store/tests/ha/master_service_ha_test.cpp
+++ b/mooncake-store/tests/ha/master_service_ha_test.cpp
@@ -508,6 +508,32 @@ class MasterServiceHATest : public ::testing::Test {
         return segment;
     }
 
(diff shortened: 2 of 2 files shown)


Icedcoco (approved):
LGTM

ykwd (approved):

PR #3532, opened 2026-08-19
[TENT] Canonicalize AMD GPU location prefix to hip: (keep rocm: as parse alias)

### LINKED ISSUE #3386: [RFC]: [TENT] Align AMD GPU location prefix with TE: `rocm:` vs `hip:`?
## Background

On the same AMD/ROCm host we observed:

| Backend | Discover location keys |
|---------|------------------------|
| Classic TE (`USE_HIP=ON`) | `hip:0` … `hip:7` |
| TENT (ROCm platform) | `rocm:0` … `rocm:7` |

Classic TE uses compile-time `GPU_PREFIX` from vendor headers:

```cpp
// mooncake-transfer-engine/include/gpu_vendor/hip.h
const static std::string GPU_PREFIX = "hip:";
```

TENT ROCm probe / plugin hardcode `rocm:`:

```cpp
// mooncake-transfer-engine/tent/src/platform/rocm/rocm_probe.cpp
entry.name = "rocm:" + std::to_string(i);
entry.type = Topology::MEM_ROCM;
```

`Topology::memTypeFromLocation` today maps `"rocm"` → `MEM_ROCM`, but does **not** recognize `"hip"`.

## Why this matters

1. **Custom NIC priority matrix cannot be shared as-is** between TE and TENT (`MC_CUSTOM_TOPO_JSON` / `topology/priority_matrix` keys must match discover names).
2. **App location strings** (`registerLocalMemory` / `allocateLocalMemory(..., "hip:0")`) break or lose affinity when switching `MC_USE_TENT=1`.
3. **Ops / dump tools** (`transfer_engine_topology_dump --backend te|tent`) look like different machines even when hardware is identical.
4. Docs and examples mostly show `cuda:N`; AMD path is already under-documented, and the dual prefix makes it worse.

## Proposed options

**Option A — Unify on `hip:` (preferred for discussion)**  
TENT ROCm platform emits `hip:N`, treat as AMD GPU location (keep or rename `MEM_ROCM` internally). Aligns with TE `USE_HIP` and HIP API naming.

**Option B — Keep `rocm:` as canonical, accept `hip:` as alias**  
Discover still prints `rocm:`; parsing / matrix / register accept both. TE→TENT migration easier for matrix files that already use `hip:`.

**Option C — Status quo + document**  
Document that TE=`hip:`, TENT=`rocm:`; users rewrite matrix / location when switching backends.

**Option D — Configurable prefix**  
e.g. conf / env override. Most flexible, highest complexity and footguns.

## Questions for reviewers

1. Should external location strings match TE’s vendor `GPU_PREFIX` (`hip:`) for AMD?
2. Is `rocm:` intentional (vendor stack) or historical from the ROCm platform plugin?
3. If we change to `hip:`, do we need a deprecation window for `rocm:` in conf / matrix / public APIs?
4. Should `memTypeFromLocation` accept `hip` → `MEM_ROCM` even before a full rename?

## Non-goals (for this issue)

- Unifying CUDA (`cuda:`) naming across TE/TENT (already aligned).
- Changing Ascend / other vendor prefixes.
- Redesigning the full topology schema.

## Related

- Recent TENT custom NIC matrix work (shared matrix format with TE).
- Local dump comparison: TE `hip:*` vs TENT `rocm:*` on the same host.

staryxchen:
## Description

Implements **Option A** from the RFC discussion in #3386, where
@amd-arozanov endorsed unifying on `hip:`:

> I like the option A. We already use USE_HIP in CMakeFiles for TE and TENT
> and "HIP" naming was initially introduced in TE before TENT. So, renaming
> rocm->hip sounds good to me.
>
> The only question is if we want to continue support "ROCM" as alias

This PR also answers that open question: **yes** — `rocm:` is kept as a
parse-only alias, so existing TENT configs, custom NIC priority matrices
(`MC_CUSTOM_TOPO_JSON`), and app location strings keep working unchanged.
Only discovery output changes (`rocm:N` → `hip:N`), aligning with classic
TE's `GPU_PREFIX` (`include/gpu_vendor/hip.h`).

**Context** (from #3386): on the same AMD/ROCm host, classic TE discovers
`hip:0..7` while TENT discovers `rocm:0..7`, so (1) custom NIC matrices
cannot be shared between backends, (2) app location strings like
`registerLocalMemory(..., "hip:0")` lose affinity under `MC_USE_TENT=1`,
(3) topology dumps look like different machines on identical hardware.

Resolutions for the RFC's reviewer questions:

| RFC question | Resolution |
|---|---|
| Match TE's vendor `GPU_PREFIX` (`hip:`)? | Yes — emit and canonical form is now `hip:` |
| Was `rocm:` intentional or historical? | Historical (HIP naming predates TENT, per review) |
| Deprecation window for `rocm:`? | Not needed — accepted indefinitely as parse-only alias; only emitted names change |
| Accept `hip` in `memTypeFromLocation`? | Yes — both `hip` and `rocm` map to `MEM_ROCM` |

What changed:

- Discovery and dumps now emit `hip:N` (ROCm probe, plugin `query_location`,
  `MemoryProber::probeDeviceMemory`).
- All parse paths accept both prefixes: allocators, plugin dispatch
  (`class_name` matching), `getTypeEnum`, `matchesMemoryPattern`,
  `memTypeFromLocation`.
- Shared helpers live in a new self-contained header
  `tent/runtime/amd_location.h` (`kAmdGpuLocationType`, alias constant,
  `isAmdGpuLocationType`), consumed by both the runtime and the out-of-tree
  rocm device plugin — single source of truth for the prefix.
- Out-of-tree plugins that still register `class_name = "rocm"` continue to
  dispatch correctly (the prober matches either name on both sides).
- Bonus fix: the CPU platform used to silently allocate DRAM on socket 0 for
  `rocm:N` locations; it now rejects all GPU locations with a clear error.
- Docs (`docs/source/api-reference/cpp/tent.md`) and the `tebench` TENT
  backend updated; `MemoryProber::listPlugins` now reports `hip`.

Compatibility notes (user-visible):

- Configs/matrices/apps using `rocm:N` keep working (parse-only alias).
- Anything that string-matches discovery output must switch to `hip:N`.

Closes #3386.

## Module

## Type of Change

- [x] Refactor
- [ ] Breaking change *(parse paths accept both prefixes; only consumers that
      string-match emitted names are affected — flagged in case maintainers
      view that as breaking)*

## How Has This Been Tested?

**Test commands:**
```bash
cmake -G Ninja -B build -DUSE_TENT=ON -DUSE_HTTP=ON -DBUILD_UNIT_TESTS=ON \
      -DBUILD_EXAMPLES=OFF -DUSE_CUDA=OFF -DWITH_STORE=OFF -DCMAKE_BUILD_TYPE=Release
ninja -C build mooncake-transfer-engine/tent/tests/all
ctest --test-dir build/mooncake-transfer-engine/tent/tests -j --output-on-failure
```

**Test results:**
- [x] Unit tests pass — 38/38 host-side tent tests, including the new
      `TopologyPriorityMatrixTest.AmdGpuLocationUsesHipPrefixAndRocmAlias`
      (covers `memTypeFromLocation("hip:0"/"rocm:3")` → `MEM_ROCM`,
      `makeAmdGpuLocation`, mixed `hip:`/`rocm:` matrix keys) and the
      44-case `TransportSelectorTest` (MTYPE_ROCM pattern matching).
- [ ] **Device-side tests NOT run** — `tent_rocm_platform_test` (legacy
      `rocm:0` allocate → `getLocation` returns `hip:0`, probe names, D2D/H2D
      copies) requires an AMD GPU + ROCm toolchain, unavailable on the dev
      host used for this PR. Note `ci_rocm.yml` only builds the wheel and
      does not run it — reviewer with AMD hardware input welcome.
- [x] `rocm_plugin.cpp` syntax-verified against the new shared header with a
      stubbed HIP header (`g++ -fsyntax-only`); runtime compile of the plugin
      itself happens under `USE_HIP=ON` in the ROCm wheel CI.

## Checklist

- [ ] For changes >500 LOC: I have filed a RFC issue — *(149 insertions; RFC #3386 filed anyway)*

## AI Assistance Disclosure

CodeBuddy (GLM) assisted with implementation review (first-principles /
minimal-change check), the `amd_location.h` deduplication refactor, running
the host-side test suite, and drafting this PR. All changed lines have been
reviewed by the human submitter, who can defend the change end-to-end.

### DIFF
--- a/mooncake-transfer-engine/tent/src/platform/cpu_allocator.cpp
+++ b/mooncake-transfer-engine/tent/src/platform/cpu_allocator.cpp
@@ -22,9 +22,9 @@ namespace mooncake {
 namespace tent {
 Status CpuPlatform::allocate(void** pptr, size_t size, MemoryOptions& options) {
     LocationParser location(options.location);
-    if (location.type() == "cuda") {
+    if (location.type() == "cuda" || isAmdGpuLocationType(location.type())) {
         return Status::NotImplemented(
-            "CUDA not supported in cpu platform" LOC_MARK);
+            "GPU locations are not supported in cpu platform" LOC_MARK);
     }
     int socket_id = 0;
     if (location.type() == "cpu") socket_id = location.index();
--- a/mooncake-transfer-engine/tent/tests/topology_priority_matrix_test.cpp
+++ b/mooncake-transfer-engine/tent/tests/topology_priority_matrix_test.cpp
@@ -109,6 +109,80 @@ TEST(TopologyPriorityMatrixTest, RejectsMalformedMatrixEntries) {
         topology.parsePriorityMatrix(R"({"cpu:0": [[1], ["mlx5_1"]]})").ok());
 }
 
+TEST(TopologyPriorityMatrixTest, AmdGpuLocationUsesHipPrefixAndRocmAlias) {
+    EXPECT_EQ(memTypeFromLocation("hip:0"), Topology::MEM_ROCM);
+    EXPECT_EQ(memTypeFromLocation("rocm:3"), Topology::MEM_ROCM);
+    EXPECT_EQ(makeAmdGpuLocation(2), "hip:2");
+    EXPECT_TRUE(isAmdGpuLocationType("hip"));
+    EXPECT_TRUE(isAmdGpuLocationType("rocm"));
+    EXPECT_FALSE(isAmdGpuLocationType("cuda"));
+    EXPECT_EQ(canonicalizeLocation("rocm:1"), "hip:1");
+    EXPECT_EQ(canonicalizeLocation("hip:1"), "hip:1");
+    EXPECT_EQ(canonicalizeLocation("cpu:0"), "cpu:0");
+
+    Topology topology;
+    ASSERT_TRUE(topology
+                    .parsePriorityMatrix(R"json({
+          "hip:0": [["mlx5_0"], ["mlx5_1"]],
+          "rocm:1": [["mlx5_1"], ["mlx5_0"]]
+        })json")
+                    .ok());
+
+    const auto* hip0 = topology.getMemEntry("hip:0");
+    ASSERT_NE(hip0, nullptr);
+    EXPECT_EQ(hip0->type, Topology::MEM_ROCM);
+
+    // The legacy "rocm:1" key is stored under its canonical name, and both
+    // the canonical and the legacy name resolve to the same entry.
+    const auto* rocm1 = topology.getMemEntry("rocm:1");
+    ASSERT_NE(rocm1, nullptr);
+    EXPECT_EQ(rocm1->name, "hip:1");
+    EXPECT_EQ(rocm1->type, Topology::MEM_ROCM);
+    EXPECT_EQ(topology.getMemEntry("hip:1"), rocm1);
+}
+
+// Regression test: a matrix configured with only the legacy "rocm:N" key
+// must be reachable via the canonical runtime name, so the configured NIC
+// priorities are not silently ignored by exact-name lookups.
+TEST(TopologyPriorityMatrixTest,
+     LegacyRocmMatrixKeyResolvesViaCanonicalHipName) {
+    Topology topology;
+    ASSERT_TRUE(topology
+                    .parsePriorityMatrix(R"json({
+          "rocm:0": [["mlx5_0"], ["mlx5_1"]]
+        })json")
+                    .ok());
+
+    // Stored under the canonical name only.
+    const auto* hip0 = topology.getMemEntry("hip:0");
+    ASSERT_NE(hip0, nullptr);
+    EXPECT_EQ(hip0->name, "hip:0");
+    EXPECT_EQ(hip0->type, Topology::MEM_ROCM);
+
+    // Exact-id lookups with the canonical name resolve to the entry and
+    // carry the configured per-GPU NIC priorities.
+    const auto mem_id = topology.getMemId("hip:0");
+    EXPECT_GE(mem_id, 0);
+    const auto* by_id = topology.getMemEntry(mem_id);
+    ASSERT_NE(by_id, nullptr);
+    ASSERT_EQ(by_id->device_list[0].size(), 1u);
+    EXPECT_EQ(topology.getNicName(by_id->device_list[0][0]), "mlx5_0");
+    ASSERT_EQ(by_id->device_list[1].size(), 1u);
+    EXPECT_EQ(topology.getNicName(by_id->device_list[1][0]), "mlx5_1");
+}
+
+TEST(TopologyPriorityMatrixTest, RejectsConflictingHipAndRocmKeys) {
+    Topology topology;
+    // "rocm:0" and "hip:0" refer to the same device; specifying both is a
+    // config error rather than a silent last-wins.
+    EXPECT_FALSE(topology
+                     .parsePriorityMatrix(R"json({
+          "hip:0": [["mlx5_0"], ["mlx5_1"]],
+          "rocm:0": [["mlx5_1"], ["mlx5_0"]]
+        })json")
+                     .ok());
+}
+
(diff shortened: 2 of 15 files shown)


amd-arozanov (approved):
LGTM. Thanks!

mooncake-transfer-engine/tent/src/platform/cpu_allocator.cpp
```diff
@@ -22,9 +22,9 @@ namespace mooncake {
 namespace tent {
 Status CpuPlatform::allocate(void** pptr, size_t size, MemoryOptions& options) {
     LocationParser location(options.location);
-    if (location.type() == "cuda") {
+    if (location.type() == "cuda" || isAmdGpuLocationType(location.type())) {
         return Status::NotImplemented(
-            "CUDA not supported in cpu platform" LOC_MARK);
+            "GPU locations are not supported in cpu platform" LOC_MARK);
     }
```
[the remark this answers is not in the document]
staryxchen: Good catch — this is pre-existing behavior: before this PR, `CpuPlatform::allocate` only rejected `cuda:` locations, and other device location types (`gpu:`, `ascend:`, `tpu:`) already fell through to socket-0 DRAM allocation. This PR extends the rejection to AMD (`hip:`/`rocm:`) only because that is in scope of the prefix alignment. Rejecting *all* non-CPU location types would be a behavior change beyond this refactor (and potentially breaking for users who unknowingly rely on the fallback), so we deliberately left it out. Happy to address it in a separate follow-up change if maintainers prefer.

mooncake-transfer-engine/tent/tests/topology_priority_matrix_test.cpp
```diff
@@ -109,6 +109,31 @@ TEST(TopologyPriorityMatrixTest, RejectsMalformedMatrixEntries) {
...
+    EXPECT_TRUE(isAmdGpuLocationType("rocm"));
+    EXPECT_FALSE(isAmdGpuLocationType("cuda"));
+
+    Topology topology;
+    ASSERT_TRUE(topology
+                    .parsePriorityMatrix(R"json({
+          "hip:0": [["mlx5_0"], ["mlx5_1"]],
+          "rocm:1": [["mlx5_1"], ["mlx5_0"]]
+        })json")
+                    .ok());
```
alogfans, line 129: `parsePriorityMatrix()` still stores the original JSON key in `MemEntry::name`. As a result, a legacy entry such as `rocm:0` is classified as `MEM_ROCM`, but remains named `rocm:0`.

After this PR, discovery and pointer probing emit `hip:0`. The RDMA selection path performs an exact `getMemId(location)` lookup and falls back to the wildcard entry when it misses, so `hip:0` will not match a matrix entry stored as `rocm:0`. The configured per-GPU NIC priorities may therefore be silently ignored, contrary to the stated compatibility guarantee.

The current test only verifies that `getMemEntry("rocm:1")` exists; it does not exercise lookup using the canonical runtime name.

Please canonicalize AMD location names (`rocm:N` → `hip:N`) at the topology storage/query boundary, define how duplicate `hip:N`/`rocm:N` keys are handled, and add a regression test that parses a matrix containing only `rocm:0` and successfully retrieves the same NIC priority list through `hip:0`.

staryxchen: Fixed in b4e6bf76: legacy `rocm:N` matrix keys are now canonicalized to `hip:N` at parse time via `canonicalizeLocation()` (topology.h), and `getMemEntry`/`getMemId` canonicalize query names as well, so lookups succeed from either side. Duplicate `hip:N` + `rocm:N` keys for the same device are rejected with `MalformedJson` (explicit error rather than silent last-wins). Added regression tests `LegacyRocmMatrixKeyResolvesViaCanonicalHipName` (matrix with only `rocm:0`, NIC priority list retrieved via `hip:0`) and `RejectsConflictingHipAndRocmKeys`.

alogfans (approved):
LGTM

PR #3533, opened 2026-08-19
[Store] Skip full metadata scan in ReMountSegment when unneeded (#3517)  

### LINKED ISSUE #3517: [Bug]: Master put RPC_TIMEOUT for minutes when scaling up a P node
### Describe the bug

Scaling up (adding) a P node causes the store master to time out `put` operations for several minutes. The failing operations report `RPC_TIMEOUT` with "Master failed to start put operation":

```text
E0819 00:28:12.025913 10022 client_service.cpp:3075] Operation for key model@tp_rank:2@pcp0@dcp0@pp_rank:0@group:0@783fcef30ea685aa5a838410635fe3ca41f8d58cb426b2a3f0a66718e95ef1d8 failed: RPC_TIMEOUT (Master failed to start put operation)
E0819 00:28:12.025913 10022 client_service.cpp:3075] Operation for key model@tp_rank:2@pcp0@dcp0@pp_rank:0@group:0@2cb6144a37d5a13e2c3300698274c608a69052b1ead3ef37cb43bd6544b1aa1c failed: RPC_TIMEOUT (Master failed to start put operation)
```

Root cause
The may be introduced by ff217b8c ("[Store] implement RFC #2756 batch-record HA OpLog writer", PR #2826).

Before that commit, MasterService::ReMountSegment held snapshot_mutex_ as a shared lock and delegated to ScopedSegmentAccess::ReMountSegment, which simply mounted each segment (O(#segments)). After the commit, in order to recover allocators from the standby snapshot, ReMountSegment now:

takes exclusive locks on both snapshot_mutex_ and client_mutex_;
scans the entire metadata space — for (size_t shard_index = 0; shard_index < kNumShards; ++shard_index) over every tenant, every key, and every replica — to find replicas pointing at the segments being remounted.
Every PutStart / BatchPutStart requires shared locks on the same two mutexes, so while a single ReMountSegment scans, all puts are blocked. The client's coro_rpc request then exceeds its default 30s timeout and surfaces as RPC_TIMEOUT.

commit: 12928dde40
master command: 
```
mooncake_master --enable_ha=true --ha_backend_type=redis --cluster_id=Test --rpc_address=xxx --rpc_port=50051 --metrics_port=9003 --max_threads=4 --rpc_thread_num=4 --default_kv_lease_ttl=5000 --client_ttl=10 --eviction_ratio=0.05 --eviction_high_watermark_ratio=0.95 --logtostderr=true --log_dir=/mooncake/logs --v=1 --ha_backend_connstring=redis://xxx:6379
```

client command:
```
mooncake_client --master_server_address=redis://xxxx:6379 --host=xxxx --device_names=mlx5_24,mlx5_28,mlx5_29,mlx5_33 --port=50064 --protocol=rdma --metadata_server=redis://xxxx:6379 --global_segment_size=256GB --enable_offload=false --enable_http_server=true --http_port=9300 --threads=16
```

### Before submitting...

KubrickLiu:
## Description

ReMountSegment unconditionally scans all kNumShards metadata shards under
exclusive client_mutex_/snapshot_mutex_ locks to reconcile
DummyBufferAllocator-backed placeholder replicas left over from standby
snapshot restoration. This scan is only meaningful when the segment being
remounted still has unreconciled replicas tracked in
standby_accounted_memory_bytes_ (i.e. it was restored via
RestoreFromStandby after a leader promotion and hasn't been reclaimed by a
real client remount yet).

For segments that never went through standby restoration — most notably a
newly mounted segment during scale-up, whose first ReMountSegment call is
triggered by Ping() returning NEED_REMOUNT right after MountSegment() — the
scan still runs anyway and walks every tenant/key/replica across the full
metadata store while holding locks that PutStart/BatchPutStart also need,
causing them to stall for minutes under load.

This PR guards the scan with `any_standby_kept_alive`, short-circuiting the
loop whenever none of the segments being remounted has an entry in
`standby_accounted_memory_bytes_`. Behavior for the necessary case (post
leader-promotion reconciliation) is unchanged.

Fixes #3517

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [x] Performance improvement
- [ ] Other

### DIFF
--- a/mooncake-store/src/master_service.cpp
+++ b/mooncake-store/src/master_service.cpp
@@ -1105,7 +1105,13 @@ auto MasterService::ReMountSegment(const std::vector<Segment>& segments,
         bool ambiguous_endpoint = false;
         bool unsupported_cxl = false;
         std::unordered_set<ObjectMetadata*> affected_objects;
-        for (size_t shard_index = 0; shard_index < kNumShards; ++shard_index) {
+        bool any_standby_kept_alive = std::any_of(
+            segments.begin(), segments.end(), [this](const Segment& segment) {
+                return standby_accounted_memory_bytes_.contains(segment.name);
+            });
+        for (size_t shard_index = 0;
+             any_standby_kept_alive && shard_index < kNumShards;
+             ++shard_index) {
             MetadataShardAccessorRW shard(this, shard_index);
             for (auto& [tenant_id, tenant] : shard->tenants) {
                 (void)tenant_id;


he-yufeng (commented):
Verified the gate against the two places `standby_accounted_memory_bytes_` is written, and the skip is sound:

- The map is only ever populated by the wholesale reassignment in the standby snapshot restore (same pass that creates the DummyBufferAllocator placeholder replicas and fills `standby_allocator_keepalive_`). So a placeholder for segment X can only exist when X is in the map.
- Entries are erased only at the end of the success path, after `replace_memory_buffer` has converted that segment's placeholders into real buffers. Every `fail_remount` returns before the erase, so an erased name means no placeholders remain for it.

With those two directions closed, "no incoming segment in the map" really does imply "no placeholder can match a restore", and skipping the shard scan cannot leave a placeholder unreconciled. A partial two-batch remount also behaves: the batch carrying the kept-alive segment still scans.

Commenting rather than approving only because this is the HA restore path and I have no local build here; the invariant above is what I checked, and it holds.

KubrickLiu:
Thanks yufeng @he-yufeng, appreciate you verifying the invariant ~

@Icedcoco @lxy-alexander could you also take a look when you have a chance ~~

Icedcoco:
LGTM pending CI

lxy-alexander:
Sorry for the late reply @KubrickLiu . The client still calls ReMountSegment() when Ping() returns NEED_REMOUNT. However, this RPC is also used for newly added P nodes, so calling it does not necessarily mean that standby recovery is required. I think the implementation should distinguish a fresh segment registration from an actual remount after Master failover. Maybe we need re-define the client status later. @Icedcoco @he-yufeng

KubrickLiu:
Agreed @lxy-alexander, that's the real root cause — conflating fresh registration with post-failover remount under one status. I'd suggest keeping this PR scoped to the #3517 fix for now, and tracking the ClientStatus redesign as a follow-up.

Aionw:
> Sorry for the late reply @KubrickLiu . The client still calls ReMountSegment() when Ping() returns NEED_REMOUNT. However, this RPC is also used for newly added P nodes, so calling it does not necessarily mean that standby recovery is required. I think the implementation should distinguish a fresh segment registration from an actual remount after Master failover. Maybe we need re-define the client status later. @Icedcoco @he-yufeng

Would you mind opening a new issue to track this, @KubrickLiu @lxy-alexander?

From what I've seen so far, it would be much cheaper and architecturally cleaner to send segment status updates (e.g., `stale`, `remount`, `new`) instead.

ykwd (approved):

PR #3535, opened 2026-08-19
[TENT] Make RPC server thread count configurable via MC_TENT_RPC_THREADS

gogongxt:
## Description

The TENT TCP data path routes every transfer through `ControlService::onSendData`/`onRecvData`, which do a full-payload blocking memory copy inline in the handler (a GPU copy is a synchronous `cudaMemcpy`). `CoroRpcAgent` runs the server on a single io_context thread (`kRpcThreads == 1`), so the receiver serializes both the socket I/O and the copies of concurrent bulk transfers. On a 2x100G node pair without RDMA this caps the TCP transport at ~20-30 Gbps regardless of sender concurrency.

#3404 chose per-handler offload over raising `kRpcThreads`, citing the unlocked `func_map_` read. That concern no longer applies: `func_map_` is mutex-protected since #3404 and registration completes before the server starts. Offload alone also cannot fix this — we measured it: with `offload=true` on SendData/RecvData the handler copies run on the blocking executor, but the single io_context thread still serializes all socket reads, plateauing at **~34 Gbps** with exactly one receiver thread pegged at 100% CPU. Raising the io_context threads reaches 5x that on the same rig.

This PR makes the thread count configurable via a new `rpc_server_threads` config key, following the `rpc_server_port` precedent:

- `ConfigHelper::loadFromEnv` maps `MC_TENT_RPC_THREADS` → `rpc_server_threads`, so the knob works from the env var or from `MC_TENT_CONF` (inline JSON or file).
- `getRpcServerThreadsFromConfig` (modeled on `getRpcServerPortFromConfig`) validates the value: full-string integer parsing, bounded to [1, 1024].
- `CoroRpcAgent::start(port, ipv6, threads)` takes the count as a parameter; default 1 everywhere — no behavior change unless opted in. The effective value (>1) is logged at startup.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [x] Performance improvement
- [ ] Other

## How Has This Been Tested?

Two nodes (2x100G each), TENT TCP transport, DRAM and CUDA buffers, 16 MB requests, `max_concurrent_tasks=64`. iperf3 on the same pair: 52 Gbps single-stream, 188 Gbps with 8 streams.

| variant | DRAM->DRAM, conc=64 | CUDA->CUDA, conc=64 |
|---|---|---|
| main (1 thread) | 20.5 Gbps | 23.0 Gbps |
| SendData/RecvData `offload=true` (measured as an alternative) | ~32 Gbps | — |
| this PR, `rpc_server_threads=8` | 109.1 Gbps | 106.6 Gbps |

Data correctness: WRITE+READ round-trip byte-compare against a GPU target (8x16MB, 32x1MB, 4x64MB) passes; the TCP data-path round-trip test added in the stacked PR covers the CPU path on loopback.

**Test commands:**
```bash
# receiver: engine target with MC_TENT_RPC_THREADS=8, MC_USE_TENT=1
# sender: batch transfer of non-contiguous 16MB slices, sweep concurrency
```

**Test results:**
- [ ] Unit tests pass

## Checklist

- [ ] I have formatted my code using `./scripts/code_format.sh` (clang-format-20 unavailable in this environment; changed lines hand-aligned to the repo `.clang-format` style)
- [x] I have added tests to prove my changes are effective (manual benchmarks above; automated coverage in the stacked PR)

## AI Assistance Disclosure

- [ ] No AI tools were used

Claude Code helped implement the change, run the cross-node benchmarks and prepare this PR; the submitter reviewed every changed line and can defend the change end-to-end.

### DIFF
--- a/mooncake-transfer-engine/tent/include/tent/rpc/rpc.h
+++ b/mooncake-transfer-engine/tent/include/tent/rpc/rpc.h
@@ -64,9 +64,9 @@ class CoroRpcAgent {
     using Function = std::function<void(const std::string_view & /* request */,
                                         std::string & /* response */)>;
 
-    // Handlers run on the server io_context thread pool (rpcServerThreads(),
-    // 1 by default), so a blocking one stalls every connection sharing its
-    // thread, not just its own.
+    // Handlers run on the server io_context thread pool (the threads
+    // argument of start(), 1 by default), so a blocking one stalls every
+    // connection sharing its thread, not just its own.
     //
     // offload=false (default): inline, no thread hop. Right for anything that
     // returns promptly.
@@ -77,7 +77,11 @@ class CoroRpcAgent {
     Status registerFunction(int func_id, const Function &func,
                             bool offload = false);
 
-    Status start(uint16_t &port, bool ipv6 = false);
+    // threads: number of io_context worker threads (default 1, the
+    // historical behavior). The TCP data-path handlers do full-payload
+    // blocking copies inline, so a single thread caps TCP throughput;
+    // sourced from the rpc_server_threads config key.
+    Status start(uint16_t &port, bool ipv6 = false, size_t threads = 1);
 
     Status stop();
 
@@ -111,10 +115,6 @@ class CoroRpcAgent {
     std::unordered_map<int, Handler> func_map_;
 
     std::atomic<bool> running_{false};
-    // RPC server worker threads, overridable via MC_TENT_RPC_THREADS
-    // (default 1). The TCP data-path handlers do full-payload blocking
-    // copies inline, so a single thread caps TCP throughput.
-    static size_t rpcServerThreads();
 };
 
 }  // namespace tent
--- a/mooncake-transfer-engine/tent/include/tent/rpc/rpc.h
+++ b/mooncake-transfer-engine/tent/include/tent/rpc/rpc.h
@@ -64,8 +64,9 @@ class CoroRpcAgent {
     using Function = std::function<void(const std::string_view & /* request */,
                                         std::string & /* response */)>;
 
-    // Handlers run on the single io_context (kRpcThreads), so a blocking one
-    // stalls every connection, not just its own.
+    // Handlers run on the server io_context thread pool (rpcServerThreads(),
+    // 1 by default), so a blocking one stalls every connection sharing its
+    // thread, not just its own.
     //
     // offload=false (default): inline, no thread hop. Right for anything that
     // returns promptly.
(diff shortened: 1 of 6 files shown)


staryxchen:
The change itself works and is well-guarded (magic-static init, validation with warnings). Two design points before this lands:

**1. Please wire this through the tent config system rather than a direct `getenv`.**

Tent already has a unified mechanism for exactly this kind of tunable: `MC_TENT_CONF` (file or inline JSON) plus ~23 env→config-key mappings in `ConfigHelper::loadFromEnv` (config.cpp), e.g. `MC_WORKERS_PER_CTX` → `transports/rdma/...`. A direct `std::getenv` read creates a second, parallel configuration path — the knob can't be set via the config file and bypasses the config schema entirely. Please add a config key (e.g. `rpc/server_threads`) and map `MC_TENT_RPC_THREADS` through `setConfig` like the existing tunables, keeping the default of 1.

**2. Why more io_context threads rather than `offload=true` for the data-path handlers?**

The `offload` flag exists precisely for the "blocking handler stalls the io_context" problem — `Delegate` already uses it, with a comment describing that failure mode. Registering `SendData`/`RecvData` with `offload=true` would need no new knob at all. I can see a reason to prefer your approach — offloaded handlers share the global block executor with `Delegate`'s whole-transfer handlers, so data-path copies could contend with long-running transfers there — but that trade-off isn't stated in the PR. Could you add a line to the description explaining why thread-count scaling beats offloading here (a benchmark comparison would be even better)?

Minor: `std::stoi` accepts trailing garbage (`MC_TENT_RPC_THREADS=8x` → 8 silently), and there's no upper bound on the value. Neither is blocking.

gogongxt:
Thanks for the review! Both points addressed in 6f381e49.

**1. Config system** — `MC_TENT_RPC_THREADS` now maps to a top-level `rpc_server_threads` key via `ConfigHelper::loadFromEnv`, following the `rpc_server_port` precedent: validation lives in `TransferEngineImpl::construct()` (`getRpcServerThreadsFromConfig`, modeled on `getRpcServerPortFromConfig`), which rejects non-integer values and bounds the value to [1, 1024] — this also fixes the `stoi` trailing-garbage issue you noted, since the value now goes through full-string parsing. The knob is settable from `MC_TENT_CONF` (JSON or file) or the env var, and `CoroRpcAgent::start()` takes the thread count as a parameter instead of reading `getenv` itself.

**2. Threads vs `offload=true`** — we measured exactly that comparison before choosing this approach. Registering `SendData`/`RecvData` with `offload=true` (handlers on the blocking executor, socket I/O still on the single io_context thread) plateaus at **~34 Gbps** with one receiver thread pegged at 100% CPU — the handler copies are parallel but the socket reads stay serialized. Raising the io_context threads to 8 reaches **109 Gbps** on the same rig (20.5 Gbps baseline, 16MB DRAM writes, conc=64; iperf3 on the pair: 52/188 Gbps single/8-stream). So offloading alone can't fix this; the socket I/O of concurrent bulk transfers needs multiple io_context threads. I've added this to the PR description.

The contention concern you raised (data-path copies competing with `Delegate` handlers on the shared blocking executor) is a second, independent reason to prefer io_context threads for this path.

staryxchen (approved):
Thanks for detailed analysis and thorough experiments. LGTM, will merge after CI passed

PR #3536, opened 2026-08-19
[TENT] Reduce copies in the TCP data path (sendData/onRecvData)

gogongxt:
## Description

The TENT TCP data path makes redundant full-payload passes over every request:

- `ControlClient::sendData` (CPU source): `resize()` zero-fills the whole request buffer, `Platform::copy` fills it, then `call()` copies the entire string again into `callCoroutine`'s by-value parameter — three full passes, two of them pure waste.
- `ControlService::onRecvData` (CPU source): `resize()` zero-fills the response before the copy.
- `callCoroutine` (client READ path): `get_resp_attachment()` copies the response out, then the internal buffer is discarded — `release_resp_attachment()` already returns it by value, so the copy is unnecessary.

Changes:

- Add `CoroRpcAgent::callOwned()`, which moves the request into the coroutine instead of copying it.
- `callCoroutine` takes the response via `std::string response = lease->release_resp_attachment();` — one more full pass removed per READ.
- `sendData` (CPU source): `reserve` + `append` builds the RPC attachment in a single copy and moves it into the call. GPU sources keep the `resize` + D2H staging copy.
- `onRecvData` (CPU source): `assign()` is a single copy with no zero-fill. GPU targets unchanged.

Also adds `tent_tcp_datapath_roundtrip_test` (modeled on `rdma_transport_test`'s `WriteThenReadAcrossProcesses`): a forked server serves a registered DRAM buffer over loopback, the client WRITEs 8 non-contiguous patterned slices and READs them back, full `memcmp`. Two ctest cases cover `rpc_server_threads=1` and `=4`, locking in the combination with #3535 on any machine without special hardware.

Stacked on top of #3535 (the RPC-threads change); the gains are only visible once the receiver is not pinned to a single thread.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [x] Performance improvement
- [ ] Other

## How Has This Been Tested?

Same two-node rig as #3535 (TENT TCP, 16 MB requests unless noted, `rpc_server_threads=8`):

| WRITE, conc=64 | threads only (base PR) | + this PR |
|---|---|---|
| DRAM->DRAM | 109.1 Gbps | 125.1 Gbps |
| CUDA->CUDA | 106.6 Gbps | 120.7 Gbps |
| DRAM, 64 MB requests | 30.3 Gbps | 48.1 Gbps |
| CUDA, 64 MB requests | 32.0 Gbps | 53.1 Gbps |

Automated: `ctest -R tcp_datapath` — both round-trip cases pass (single-threaded and multi-threaded RPC server).

Data correctness: WRITE+READ round-trip byte-compare against a GPU target (8x16MB, 32x1MB, 4x64MB) all pass.

**Test commands:**
```bash
ctest --test-dir build -R tent_tcp_datapath_roundtrip_test --output-on-failure
# plus the same cross-node bench as the base PR, before/after this patch
```

**Test results:**
- [x] Unit tests pass

## Checklist

- [ ] I have formatted my code using `./scripts/code_format.sh` (clang-format-20 unavailable in this environment; changed lines hand-aligned to the repo `.clang-format` style)

## AI Assistance Disclosure

- [ ] No AI tools were used

Claude Code helped implement the change, run the cross-node benchmarks and prepare this PR; the submitter reviewed every changed line and can defend the change end-to-end.

### DIFF
--- a/mooncake-transfer-engine/tent/include/tent/rpc/rpc.h
+++ b/mooncake-transfer-engine/tent/include/tent/rpc/rpc.h
@@ -64,9 +64,9 @@ class CoroRpcAgent {
     using Function = std::function<void(const std::string_view & /* request */,
                                         std::string & /* response */)>;
 
-    // Handlers run on the server io_context thread pool (rpcServerThreads(),
-    // 1 by default), so a blocking one stalls every connection sharing its
-    // thread, not just its own.
+    // Handlers run on the server io_context thread pool (the threads
+    // argument of start(), 1 by default), so a blocking one stalls every
+    // connection sharing its thread, not just its own.
     //
     // offload=false (default): inline, no thread hop. Right for anything that
     // returns promptly.
@@ -77,7 +77,11 @@ class CoroRpcAgent {
     Status registerFunction(int func_id, const Function &func,
                             bool offload = false);
 
-    Status start(uint16_t &port, bool ipv6 = false);
+    // threads: number of io_context worker threads (default 1, the
+    // historical behavior). The TCP data-path handlers do full-payload
+    // blocking copies inline, so a single thread caps TCP throughput;
+    // sourced from the rpc_server_threads config key.
+    Status start(uint16_t &port, bool ipv6 = false, size_t threads = 1);
 
     Status stop();
 
@@ -111,10 +115,6 @@ class CoroRpcAgent {
     std::unordered_map<int, Handler> func_map_;
 
     std::atomic<bool> running_{false};
-    // RPC server worker threads, overridable via MC_TENT_RPC_THREADS
-    // (default 1). The TCP data-path handlers do full-payload blocking
-    // copies inline, so a single thread caps TCP throughput.
-    static size_t rpcServerThreads();
 };
 
 }  // namespace tent
--- a/mooncake-transfer-engine/tent/include/tent/rpc/rpc.h
+++ b/mooncake-transfer-engine/tent/include/tent/rpc/rpc.h
@@ -88,6 +88,11 @@ class CoroRpcAgent {
     Status call(const std::string &server_addr, int func_id,
                 const std::string_view &request, std::string &response);
 
+    // Same as call(), but moves the request into the coroutine instead of
+    // copying it, for large payloads.
+    Status callOwned(const std::string &server_addr, int func_id,
+                     std::string request, std::string &response);
+
     using AsyncCallback = std::function<void(Status, std::string)>;
     void callAsync(const std::string &server_addr, int func_id,
                    const std::string &request, AsyncCallback callback);
--- a/mooncake-transfer-engine/tent/include/tent/rpc/rpc.h
+++ b/mooncake-transfer-engine/tent/include/tent/rpc/rpc.h
@@ -64,8 +64,9 @@ class CoroRpcAgent {
     using Function = std::function<void(const std::string_view & /* request */,
                                         std::string & /* response */)>;
 
-    // Handlers run on the single io_context (kRpcThreads), so a blocking one
-    // stalls every connection, not just its own.
+    // Handlers run on the server io_context thread pool (rpcServerThreads(),
+    // 1 by default), so a blocking one stalls every connection sharing its
+    // thread, not just its own.
     //
     // offload=false (default): inline, no thread hop. Right for anything that
     // returns promptly.
@@ -110,7 +111,10 @@ class CoroRpcAgent {
     std::unordered_map<int, Handler> func_map_;
 
     std::atomic<bool> running_{false};
-    constexpr static size_t kRpcThreads = 1;
+    // RPC server worker threads, overridable via MC_TENT_RPC_THREADS
+    // (default 1). The TCP data-path handlers do full-payload blocking
+    // copies inline, so a single thread caps TCP throughput.
+    static size_t rpcServerThreads();
 };
 
 }  // namespace tent
--- a/mooncake-transfer-engine/tent/include/tent/runtime/control_plane.h
+++ b/mooncake-transfer-engine/tent/include/tent/runtime/control_plane.h
@@ -133,7 +133,7 @@ class ControlService {
 
     void setNotifyCallback(const OnNotify& callback);
 
-    Status start(uint16_t& port, bool ipv6_ = false);
+    Status start(uint16_t& port, bool ipv6_ = false, size_t threads = 1);
 
    private:
     void onGetSegmentDesc(const std::string_view& request,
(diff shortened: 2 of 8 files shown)


staryxchen:
Thanks for the PR — the copy analysis checks out. I verified that `call()` constructs the by-value coroutine parameter from the `string_view`, so `callOwned` does remove a full-payload pass, and the `resize()` zero-fills are real. The CPU/GPU branching is also safe (`MTYPE_UNKNOWN` falls back to the old path). Two suggestions before this lands:

**1. One more full-payload copy on the READ path (client side)**

In `callCoroutine` (rpc.cpp:252-253, just above the `callOwned` addition):

```cpp
std::string response{lease->get_resp_attachment()};
lease->release_resp_attachment();
```

This copies the attachment out and then discards the moved internal buffer. Since `release_resp_attachment()` returns the buffer by value, the copy can be dropped entirely:

```cpp
std::string response = lease->release_resp_attachment();
```

One more full pass removed per READ, in the same spirit as this PR. (This copy exists both before and after this change, so the benchmark numbers are unaffected — it's an incremental gain, not a correction.) I checked that this call path never sets an external response buffer, so the move is safe here.

**2. A TCP data-path round-trip test**

Correctness verification for this change is currently manual-only. `tcp_transport_test.cpp` covers `TcpParams`/`TcpTask` semantics but not payload integrity, so no automated test guards the TCP data path end to end. Since `rdma_transport_test.cpp` already has a `WriteThenReadAcrossProcesses` template (fork a server process, fill a pattern, WRITE+READ back, full `memcmp`), a TCP counterpart would be straightforward to model on it and runs on any loopback without special hardware. It would also lock in the `MC_TENT_RPC_THREADS` + copy-reduction combination, not just this PR.

gogongxt:
Thanks for the review! Both suggestions implemented in the updated branch:

**1.** `callCoroutine` now takes the response via `std::string response = lease->release_resp_attachment();` — no copy on the READ path. (Rebased on top of the config-system rework of #3535.)

**2.** Added `tent_tcp_datapath_roundtrip_test` modeled on `WriteThenReadAcrossProcesses`: a forked server serves a registered DRAM buffer over loopback, the client WRITEs 8 non-contiguous 4MB patterned slices and READs them back into a disjoint region, full `memcmp`. Two ctest cases covering `rpc_server_threads=1` and `=4`, so the MC_TENT_RPC_THREADS + copy-reduction combination is now guarded by CI-runnable tests on any machine. Both pass:

```
1/2 tent_tcp_datapath_roundtrip_test_single_threaded ...... Passed (3.9s)
2/2 tent_tcp_datapath_roundtrip_test_multi_threaded_rpc ... Passed (4.3s)
```

One implementation note: the two cases run as separate ctest invocations (via `--gtest_filter`) because each test forks its server child, which must happen before the test process initializes any engine state — running both forks in one process hit stale forked statics in the second test.

staryxchen (approved):
LGTM, will merge after #3535 merged.

BTW, could you update the latest performance for both WRITE and READ?

alogfans (approved):
LGTM. However I'm not sure why tent-ci check fails.

staryxchen:
@gogongxt @alogfans — apologies, the CI failure is in the `release_resp_attachment()` change at rpc.cpp:230, which came from my suggestion, and my safety check was incomplete. Here is the root cause and a fix that keeps the optimization.

**Root cause: the internal response buffer is padded for small attachments.**

On the client receive path, the coro_rpc client resizes the internal attachment buffer to `max(attach_length, sizeof(std::string))` (coro_rpc_client.hpp ~1604) using `struct_pack::detail::resize`, which deliberately skips zero-fill (`resize_and_overwrite`, struct_pack/util.h:130). The socket read then fills only the first `attach_length` bytes. For a 4-byte `"slow"` response the internal state is:

```
internal std::string: [ "slow" | 28 bytes uninitialized garbage ]   size = 32
view (get_resp_attachment()): (ptr -> buffer, len = 4)             zero-copy metadata
```

The real attachment length is only recorded in the view; the string's own `size()` is padded. The three ways to take the attachment differ exactly there:

```cpp
std::string s{lease->get_resp_attachment()};           // copy: uses view len 4 -> correct, O(n) copy
size_t len = lease->get_resp_attachment().size();      // metadata only: free
std::string s = lease->release_resp_attachment();      // move: O(1), but takes size=32 with it
```

The old code was correct precisely because constructing from the view copies exactly `attach_length` bytes. The change to `release_resp_attachment()` moved the whole internal string out — ownership transfers at zero cost, but for any response with `0 < attach_length < sizeof(std::string)` the result carries an uninitialized tail and the wrong length. That matches the CI signature: `tent_rpc_handler_isolation_test` fails on all three test-running jobs with `response = "slow\xA1\x7F\0\0..."` (32 bytes) vs `"slow"`. It also explains why the new round-trip test still passes: 4 MB attachments get no padding, and `SendData`'s empty success response goes through the `clear()` path. Credit to the existing isolation test — this is exactly the small-response case that big-payload benchmarks never exercise.

**Fix — take the length from the view, the data from the move:**

```cpp
const size_t len = lease->get_resp_attachment().size();
std::string response = lease->release_resp_attachment();
response.resize(len);
```

- `len` is read from the view before the move — the one authoritative source of the attachment length.
- `resize(len)` shrinking a string is O(1): it just sets the length field (and one terminator); it never reallocates or reads the tail. The garbage bytes become logically invisible.

Result per response size: large attachments (the READ data path) are a pure move — `len == size`, resize is a no-op, zero copies; small control responses get correctly trimmed; empty responses are untouched. The optimization survives, the semantics are restored.

gogongxt:
Good catch, and thank you for the thorough root-cause analysis — much appreciated.

Fixed in f5407a9c, exactly along your suggested lines: the length is taken from the attachment view before the move, and the moved string is trimmed with a shrinking `resize(len)` (O(1), sets the length field only). Large attachments keep the pure-move path (`len == size`, resize is a no-op).

Verified locally on the same tree that failed in CI:

```
tent_rpc_handler_isolation_test ....................... Passed   <- was failing in CI
tent_tcp_datapath_roundtrip_test_single_threaded ...... Passed
tent_tcp_datapath_roundtrip_test_multi_threaded_rpc ... Passed
```

Agreed on the credit to the isolation test — this is exactly the small-response case the bulk benchmarks never exercise. Worth noting the new round-trip test wouldn't have caught it either (4MB attachments get no padding, and SendData's empty response goes through `clear()`), which is a good argument for keeping the isolation test's small-payload semantics in any future data-path changes. CI is re-running on the pushed commit.

PR #3537, opened 2026-08-19
[CI/Build] Header only YLT (yalantinglibs)

ur4t:
## Description

Skip installing YLT and suppress warnings.

## Module

- [x] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [x] Documentation update
- [ ] Performance improvement
- [x] Other

## How Has This Been Tested?

**Test commands:**
```bash
```

**Test results:**
- [ ] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [x] No AI tools were used


staryxchen (commented):
What is the purpose of this change?

CMakeLists.txt
```diff
@@ -37,6 +37,11 @@ execute_process(
 )
 string(STRIP ${PYTHON_SYS_PATH} PYTHON_SYS_PATH)
 
+include_directories(${CMAKE_CURRENT_SOURCE_DIR}/extern/yalantinglibs/include)
```
Aionw: These include paths are only added when configuring from the repository root. The independently supported `cmake -S mooncake-transfer-engine ...` path includes `mooncake-common/common.cmake` directly and never evaluates this block. Since `dependencies.sh` now initializes the YLT submodule but no longer installs it, a clean machine has `extern/yalantinglibs` checked out but no include path to it; `rpc_communicator` then fails with `fatal error: ylt/coro_rpc/coro_rpc_client.hpp: No such file or directory`. Please expose the vendored YLT headers through a shared interface target/include setup used by both the root and standalone Transfer Engine builds. (A preinstalled `/usr/local/include/ylt` can mask this regression.)

Aionw (commented):
Standalone Transfer Engine build regression; see the inline comment.

ur4t:
> What is the purpose of this change?

@staryxchen To further unify build process and dependency management within Cmake.

UNIDY2002:
Please look at this failing job: https://github.com/kvcache-ai/Mooncake/actions/runs/32253375464/job/96288646441?pr=3537
It seems that building directly within `mooncake-transfer-engine/` is failing, although I'm not sure why do we need this step.
This step was introduced in https://github.com/kvcache-ai/Mooncake/pull/323. Any comments? @alogfans

Edit: also, please resolve the conflict with the latest main. Thanks!

alogfans:
@UNIDY2002 This seems that `ylt/easylog.hpp` not found (which bundled in yalantinglib)

mooncake-common/FindYLT.cmake
```diff
@@ -0,0 +1,37 @@
...
+    OFF
+    CACHE BOOL "Install YLT independent standalone")
+
+set(YLT_VERSION 0.5.7)
+FetchContent_Declare(
+  yalantinglibs
+  URL https://github.com/alibaba/yalantinglibs/archive/refs/tags/${YLT_VERSION}.tar.gz
+  URL_HASH
+    SHA256=1c1057289e5488f90dd326fd2bb9d3173bad11eb5b06bc0a8bf0fa80857e1cfa
+  FIND_PACKAGE_ARGS NAMES yalantinglibs
```
Aionw: `FIND_PACKAGE_ARGS` is only supported by newer CMake versions, while this project still declares CMake 3.16 support. With CMake 3.22, it is parsed as part of `URL_HASH`, producing `SHA256=...;FIND_PACKAGE_ARGS;NAMES;yalantinglibs` and failing during configure. The current ROCm CI reproduces exactly this failure: https://github.com/kvcache-ai/Mooncake/actions/runs/32346057439/job/96358272117. Please avoid this keyword unless the minimum CMake version is raised; an explicit `find_package(... QUIET)` / target check works on the currently supported versions.

mooncake-common/common.cmake
```diff
@@ -666,5 +666,7 @@ if(NOT TARGET gflags::gflags)
     endif()
   endforeach()
 endif()
+
+include(${CMAKE_CURRENT_LIST_DIR}/FindYLT.cmake)
```
Aionw, line 678: Even on CMake 3.31, a clean machine still fails after FetchContent populates YLT: the following `find_package(yalantinglibs CONFIG REQUIRED)` searches for an *installed* `yalantinglibsConfig.cmake`, which FetchContent does not install. I reproduced this with the exact YLT 0.5.7 source while hiding `/usr/local` from package discovery. A preinstalled system YLT masks the problem. One compatible structure is:

```cmake
find_package(yalantinglibs CONFIG QUIET)
if(NOT TARGET yalantinglibs::yalantinglibs)
  FetchContent_Declare(
    yalantinglibs
    URL ...
    URL_HASH SHA256=...
  )
  FetchContent_MakeAvailable(yalantinglibs)
endif()
```

Then this caller should only include `FindYLT.cmake` and remove the subsequent required `find_package`. If the goal is a fully pinned dependency, omit the initial `find_package` as well and always use FetchContent.

Aionw (commented):
The standalone path is now wired through the shared CMake setup, but two clean-build blockers remain in the FetchContent integration. See the inline comments for reproductions and a compatible fix.

staryxchen (commented):
Should this submodule be removed as well?
https://github.com/kvcache-ai/Mooncake/blob/cdf8b1ea04d6350bcb00d122c4d288f5da71d423/.gitmodules#L5

mooncake-common/common.cmake
ur4t: > If the goal is a fully pinned dependency, omit the initial `find_package` as well and always use FetchContent.

Yes, updated.

ur4t:
> Should this submodule be removed as well?
> 
> https://github.com/kvcache-ai/Mooncake/blob/cdf8b1ea04d6350bcb00d122c4d288f5da71d423/.gitmodules#L5

@staryxchen Yes, updated.

mooncake-common/FindYLT.cmake
```diff
@@ -0,0 +1,36 @@
...
+    OFF
+    CACHE BOOL "Install YLT independent standalone")
+
+set(YLT_VERSION 0.5.7)
+FetchContent_Declare(
+  yalantinglibs
+  URL https://github.com/alibaba/yalantinglibs/archive/refs/tags/${YLT_VERSION}.tar.gz
+  URL_HASH
+    SHA256=1c1057289e5488f90dd326fd2bb9d3173bad11eb5b06bc0a8bf0fa80857e1cfa
+)
```
staryxchen, line 16: If fetch content fails due to network issues, please consider using an alternative method (eg. using a proxy). Ref:
https://github.com/kvcache-ai/Mooncake/blob/cdf8b1ea04d6350bcb00d122c4d288f5da71d423/.github/workflows/ci_ascend.yml#L52

ur4t: Updated, support Github mirrors via reading env var `ASCEND_GITHUB_MIRROR_URLS`.

staryxchen (approved):
It's fine with me—please have someone else double-check.

UNIDY2002:
The current main references yalantinglibs at https://github.com/alibaba/yalantinglibs/commit/7801bc9ad9021781f15217552214e325a1cf7373, a version newer than 0.5.7. Now we seems to be pinning it backwards. Is it intentional?

ur4t:
> The current main references yalantinglibs at [alibaba/yalantinglibs@7801bc9](https://github.com/alibaba/yalantinglibs/commit/7801bc9ad9021781f15217552214e325a1cf7373), a version newer than 0.5.7. Now we seems to be pinning it backwards. Is it intentional?

No, I'll catch up main later.

mooncake-store/tests/host_port_fix_test.cpp
```diff
@@ -1,3 +1,5 @@
+#include <csignal>
```
caozhanhao: The `<csignal>` includes appear to be workarounds for yalantinglibs v0.5.7. This was fixed upstream in [yalantinglibs#1152](https://github.com/alibaba/yalantinglibs/pull/1152). Once yalantinglibs is updated to match main, we should be able to remove these includes.

Aionw (approved):
LGTM

caozhanhao (approved):
LGTM

ykwd (approved):
LGTM

ascend-direct-dev (approved):
LGTM

PR #2035, opened 2026-05-05
[TransferEngine] Use allocation base addr for dmabuf-based mem registration

ivanium:
## Description

`cuMemGetHandleForAddressRange` requires the exact allocation
boundaries. The current dmabuf path of `registerMemoryRegionInternal`
passes the user-provided `addr` together with
`CU_POINTER_ATTRIBUTE_RANGE_SIZE`, which is wrong whenever `addr`
sits at an offset inside a larger allocation — for example, PyTorch's
caching allocator packs many tensors into a single `cudaMalloc`
block, so the `addr` handed to `registerMemory` is rarely the
allocation base.

In that case the dmabuf-fd lookup either fails outright or maps a
region that does not actually correspond to the user's buffer, and
`ibv_reg_dmabuf_mr` is then called with a hard-coded
`offset = 0 /* offset */`, which makes the resulting MR point at the
wrong bytes even when the lookup happens to succeed.

This PR uses `cuMemGetAddressRange` to obtain the true allocation
base and size, passes those to `cuMemGetHandleForAddressRange`, and
computes `dmabuf_offset = addr - allocBase` for `ibv_reg_dmabuf_mr`.

### Relationship to PR #2019

PR #2019 has also addressed this issue with `cuPointerGetAttribute(CU_POINTER_ATTRIBUTE_RANGE_START_ADDR)`. This is functionally the same as this PR, but I personally think `cuMemGetAddressRange` is more reliable for physical memory related usages such as RDMA.
Also #2019 bundles some other changes to MACA.

Happy to defer to whichever lands first.

## Module

- [x] Transfer Engine (`mooncake-transfer-engine`)
- [ ] Mooncake Store (`mooncake-store`)
- [ ] Mooncake EP (`mooncake-ep`)
- [ ] Integration (`mooncake-integration`)
- [ ] P2P Store (`mooncake-p2p-store`)
- [ ] Python Wheel (`mooncake-wheel`)
- [ ] PyTorch Backend (`mooncake-pg`)
- [ ] Mooncake RL (`mooncake-rl`)
- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other

## How Has This Been Tested?

Tested with https://github.com/vllm-project/vllm/pull/40900

- Built on a system with `WITH_NVIDIA_PEERMEM=OFF`, `USE_CUDA=ON`.
[...]

PR #2019, opened 2026-04-30
[TransferEngine][Integration] feat: add MACA/MetaX GPU support and fix RDMA dmabuf registration

JoeZhang-0x000:
## Summary

- **MACA Compatibility**: Fix MACA GPU vendor header macros — correct `CUdeviceptr` alias, add missing CUDA memory type / pointer attribute / dmabuf handle macros, and implement an inline `cuGetErrorString` wrapper function
- **RDMA dmabuf registration fix**: Use allocation base address (`CU_POINTER_ATTRIBUTE_RANGE_START_ADDR`) when calling `cuMemGetHandleForAddressRange`, and compute proper offset for `ibv_reg_dmabuf_mr`. This fixes GPU memory registration on MACA (and potentially other DMA-BUF capable devices) where the passed pointer may not be the allocation start
- **Connector remote_request_id**: Add `remote_request_id` support to `MooncakeConnectorV1` so KV cache transfer between prefiller and decoder can correctly map remote request IDs to local ones

## Related Issues

- #1975 — `[RFC]: Mooncake EP Multi-Vendor GPU Adaptation Design`
- #1965 — `[Feature Request]: MACA Support for Mooncake EP/PG`

### `mooncake-transfer-engine/include/gpu_vendor/maca.h`
- Fix `CUdeviceptr` → `mcDeviceptr_t`
- Add `CU_MEMORYTYPE_HOST/DEVICE`, `CU_POINTER_ATTRIBUTE_*`, `CU_MEM_RANGE_HANDLE_TYPE_DMA_BUF_FD`, `CU_DEVICE_ATTRIBUTE_DMA_BUF_SUPPORTED` macros
- Implement `cuGetErrorString()` as inline wrapper around `mcGetErrorString()`

### `mooncake-transfer-engine/src/transport/rdma_transport/rdma_context.cpp`
- Add `USE_MACA` to dmabuf-based memory registration `#if` guard
- Use `CU_POINTER_ATTRIBUTE_RANGE_START_ADDR` to get allocation base before calling `cuMemGetHandleForAddressRange`
- Compute correct `offset = addr - allocBase` for `ibv_reg_dmabuf_mr`
- Improve error log messages with base address and size info

### `mooncake-wheel/mooncake/mooncake_connector_v1.py`
- Add `remote_request_id` field to `RecvReqMeta`
- Pass `remote_request_id` through scheduler → metadata → worker
[...]

PR #3538, opened 2026-08-19
[Bugfix][TransferEngine] Export the requested range for dma-buf registration under CUDA VMM (fixes #2511)

### LINKED ISSUE #2511: [Bug]: When enable expandable_segments, mooncake register memory failed
### Bug Report

When I export PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True

it will cause registration failed, like this issue mentioned: https://github.com/kvcache-ai/Mooncake/issues/351#issuecomment-2888725354

Will Mooncake support it?

### Before submitting...

hope1262946533:
## Problem

With `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True`, PyTorch allocates the KV cache through the CUDA virtual memory management API (`cuMemCreate` + `cuMemMap`). `RdmaContext::exportDmabuf()` derives the dma-buf export range from `cuMemGetAddressRange()`, which for VMM memory reports **only the single mapping containing the pointer** — one expandable-segment growth chunk, 20 MiB by default (`c10/core/AllocatorConfig.h`, `large_segment_size_{20971520}`) — not the surrounding contiguous VA reservation.

The registration then asks for the whole buffer, so `offset + length` exceeds the exported dma-buf and `ibv_reg_dmabuf_mr()` fails with `EINVAL` for every buffer larger than one chunk. Captured from a real run, with the fd size read via `lseek(fd, 0, SEEK_END)`:

```
offset=8214528  length=1916853120  iova=0x12397fd5800  fd=240
dmabuf_size=20971520   offset+length=1925067648   -> FAIL errno=22 (Invalid argument)
```

Because the caller in SGLang does not abort on registration failure, the server starts up healthy and the problem only surfaces as `Failed to send kv chunk` on the first KV transfer. See #2511 for the original report and the analysis comment there.

## Fix

`exportDmabuf()` now takes the length the caller is going to register. When the range reported by `cuMemGetAddressRange()` cannot cover `[addr, addr + length)`, the dma-buf is exported for that range instead, with a page-aligned base:

```cpp
if (exportOffset + length > allocSize) {
    const size_t page = (size_t)sysconf(_SC_PAGESIZE);
    uintptr_t aligned = (uintptr_t)addr & ~(uintptr_t)(page - 1);
    if (aligned < (uintptr_t)allocBase) aligned = (uintptr_t)allocBase;
    exportBase = (CUdeviceptr)aligned;
    exportOffset = (uintptr_t)addr - aligned;
    exportSize = (exportOffset + length + page - 1) & ~(size_t)(page - 1);
}
```

`cuMemGetHandleForAddressRange()` accepts a range spanning several mappings as long as they are contiguously mapped, which is exactly the expandable-segment layout. This was verified with a standalone probe that maps 8 separate `cuMemCreate` handles into one reservation and registers the whole range successfully.

The failure log in `registerMemoryRegionInternal()` now also prints `length` and `dmabuf_offset`. The absence of those two numbers is the reason #2511 stayed undiagnosed for two months.

## Why this is low risk

1. **The new branch only runs where the current code always fails.** `exportOffset + length > allocSize` means the existing export cannot possibly satisfy the kernel's `offset + length <= size` check. For `cudaMalloc`-backed memory the reported allocation always covers the registration, so the behaviour introduced by #2035 and #2019 for caching-allocator sub-allocations is untouched.
2. **The page-aligned base preserves the kernel's page-offset rule.** `ibv_reg_dmabuf_mr()` requires `offset % page_size == iova % page_size`; aligning the base down to a page boundary makes `exportOffset % page == (uintptr_t)addr % page` by construction. `sysconf(_SC_PAGESIZE)` is used rather than a hardcoded 4096 so 64 KiB-page kernels are handled too.
3. **Chunked registration is unaffected.** `RdmaTransport::registerLocalMemoryInternal()` still adds `chunk_offset` to `out.offset` for each chunk, and the `length` passed to `exportDmabuf()` is the whole-buffer length, so the single exported dma-buf still covers every chunk.
4. **Shared-fd lifetime is unchanged.** Still one export per buffer, imported by every NIC, closed once by `closeDmabufExport()` (#2523).
5. **The `aligned < allocBase` clamp** guards against aligning below the allocation base, reachable only if `allocBase` itself is not page-aligned.

## Testing

Verified on SGLang PD disaggregation with prefill and decode on separate 8-GPU hosts, DeepSeek-V4-Pro, ConnectX-8 RoCE (`mlx5_bond_0..7`), torch 2.11.0+cu130, `WITH_NVIDIA_PEERMEM=0`, `MC_GID_INDEX=3`, `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True` at the default 20 MiB granularity. Everything except the export range is identical across rows:

| mooncake | export range | failed registrations per side | first request |
| --- | --- | --- | --- |
| 0.3.11.post1 | as shipped | 1944 (243 buffers × 8 ranks), all `EINVAL` | HTTP 500, `Failed to send kv chunk` |
| 0.3.12.post1 | as shipped | fails likewise, `EINVAL` on the same buffers | HTTP 500 |
| 0.3.11.post1 | corrected as in this PR | **0** | correct output, KV transferred correctly |

The 0.3.12.post1 row logs fewer error lines (16 per side) than 0.3.11.post1, but the two counts are not comparable: registration aborts and rolls back on the first failing chunk, so a differing number of attempts is reached. What matters is that both versions fail with `EINVAL` on the same large buffers and both return HTTP 500.

`max_total_num_tokens` in the fixed configuration stays within 0.26% of the `expandable_segments`-disabled baseline, so the fix costs no KV capacity.

**How this was validated.** Rebuilding the wheel on that cluster was not practical, so the behaviour was validated by interposing `ibv_reg_dmabuf_mr()` with an `LD_PRELOAD` library implementing the same range computation. The arithmetic is identical: for the failing 1.79 GiB buffer the interposing library logged `new offset = 2048, exported size = 1916858368`, and the code in this PR produces exactly those two numbers for the same inputs. That said, **this patch itself has not been compiled or run in-tree** — please let CI build it, and a maintainer with a VMM-capable setup may want to repeat the end-to-end check.

## Not covered

1. Non-CUDA backends (`USE_HIP_DMABUF`, MLU, MACA, SUPA) are untouched. If their address-range queries have the same VMM caveat, they need the same treatment.
2. Functional verification only; no load or long-running test was performed.
3. 64 KiB-page hosts were not tested. `sysconf(_SC_PAGESIZE)` is used on the assumption that it is the relevant granularity there (cf. sgl-project/sglang#23971).
4. Each GPU was pinned to a single NIC in this setup, so the multi-NIC shared-fd path was not exercised.

### DIFF
--- a/mooncake-transfer-engine/include/transport/rdma_transport/rdma_context.h
+++ b/mooncake-transfer-engine/include/transport/rdma_transport/rdma_context.h
@@ -131,14 +131,18 @@ class RdmaContext {
     int registerMemoryRegion(void *addr, size_t length, int access,
                              const DmabufExport &exp);
 
-    // Exports a single dma_buf fd for the allocation backing addr. GPU device
+    // Exports a single dma_buf fd covering [addr, addr + length). GPU device
     // memory yields kDmabufReg with a live fd; host memory and the
     // nvidia-peermem path yield kHostReg with no fd. Any fd placed in out.fd
     // MUST be closed by the caller (via closeDmabufExport) AFTER every
     // registerMemoryRegion() call consuming it has returned — each successful
     // registration takes its own reference, so closing earlier would invalidate
     // the fd for the remaining NICs.
-    static int exportDmabuf(void *addr, DmabufExport &out);
+    // `length` is the length of the whole buffer the caller is going to
+    // register (chunked registrations derive their own offset from out.offset),
+    // and is used to guarantee the exported dma_buf really covers that range —
+    // see the VMM note in exportDmabuf().
+    static int exportDmabuf(void *addr, size_t length, DmabufExport &out);
 
     // Closes the fd held by a DmabufExport, if any. Idempotent.
     static void closeDmabufExport(DmabufExport &exp);
--- a/mooncake-transfer-engine/src/transport/rdma_transport/rdma_context.cpp
+++ b/mooncake-transfer-engine/src/transport/rdma_transport/rdma_context.cpp
@@ -477,14 +477,14 @@ int RdmaContext::exportDmabuf(void *addr, size_t length, DmabufExport &out) {
             exportOffset = (uintptr_t)addr - aligned;
             exportSize =
                 (exportOffset + length + page - 1) & ~(size_t)(page - 1);
-            LOG(INFO) << "dma_buf: reported allocation for " << (uintptr_t)addr
-                      << " (base=" << (uintptr_t)allocBase
-                      << " size=" << allocSize << ") does not cover length "
-                      << length << "; exporting the requested range instead"
-                      << " (base=" << (uintptr_t)exportBase
-                      << " size=" << exportSize
-                      << "). Expected for CUDA VMM allocations such as PyTorch"
-                         " expandable_segments.";
+            VLOG(1) << "dma_buf: reported allocation for " << (uintptr_t)addr
+                    << " (base=" << (uintptr_t)allocBase
+                    << " size=" << allocSize << ") does not cover length "
+                    << length << "; exporting the requested range instead"
+                    << " (base=" << (uintptr_t)exportBase
+                    << " size=" << exportSize
+                    << "). Expected for CUDA VMM allocations such as PyTorch"
+                       " expandable_segments.";
         }
 
         // flags must be 0: the PCIE-BAR1 mapping flag is rejected (error 801)
--- a/mooncake-transfer-engine/src/transport/rdma_transport/rdma_context.cpp
+++ b/mooncake-transfer-engine/src/transport/rdma_transport/rdma_context.cpp
@@ -389,9 +389,10 @@ int RdmaContext::deconstruct() {
     return 0;
 }
 
-int RdmaContext::exportDmabuf(void *addr, DmabufExport &out) {
+int RdmaContext::exportDmabuf(void *addr, size_t length, DmabufExport &out) {
     out = DmabufExport{};
     (void)addr;  // unused on the host-only (#else) build
+    (void)length;
 #if defined(USE_MLU) || defined(USE_MACA) || defined(USE_CUDA) || \
     defined(USE_SUPA)
     // Decide host vs GPU without assuming the presence of nvidia-peermem. Host
@@ -446,17 +447,57 @@ int RdmaContext::exportDmabuf(void *addr, DmabufExport &out) {
         }
 
         int dmabuf_fd;
+        // cuMemGetAddressRange() only reports the mapping that contains addr.
+        // For memory allocated through the CUDA virtual memory management API
+        // (cuMemCreate + cuMemMap), as used by PyTorch's
+        // PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True, that mapping is a
+        // single growth chunk — 20 MiB by default, see c10's
+        // large_segment_size / large_segment_size_mb — even though the
+        // surrounding VA reservation is contiguous and orders of magnitude
+        // larger. Exporting only that chunk makes offset + length exceed the
+        // resulting dma_buf, and ibv_reg_dmabuf_mr() below then fails with
+        // EINVAL for every buffer larger than one chunk (Mooncake#2511).
+        //
+        // When the reported allocation does not cover the range the caller is
+        // about to register, export exactly [addr, addr + length) instead.
+        // cuMemGetHandleForAddressRange() accepts a range spanning several
+        // mappings as long as they are contiguously mapped, which is precisely
+        // the expandable-segment layout. The export base is page-aligned so
+        // that offset % page_size == iova % page_size, which
+        // ibv_reg_dmabuf_mr() requires.
+        CUdeviceptr exportBase = allocBase;
+        size_t exportSize = allocSize;
+        uint64_t exportOffset = (uintptr_t)addr - (uintptr_t)allocBase;
+        if (exportOffset + length > allocSize) {
+            const size_t page = (size_t)sysconf(_SC_PAGESIZE);
+            uintptr_t aligned = (uintptr_t)addr & ~(uintptr_t)(page - 1);
+            if (aligned < (uintptr_t)allocBase)
+                aligned = (uintptr_t)allocBase;
+            exportBase = (CUdeviceptr)aligned;
+            exportOffset = (uintptr_t)addr - aligned;
+            exportSize =
+                (exportOffset + length + page - 1) & ~(size_t)(page - 1);
+            LOG(INFO) << "dma_buf: reported allocation for " << (uintptr_t)addr
+                      << " (base=" << (uintptr_t)allocBase
+                      << " size=" << allocSize << ") does not cover length "
+                      << length << "; exporting the requested range instead"
+                      << " (base=" << (uintptr_t)exportBase
+                      << " size=" << exportSize
+                      << "). Expected for CUDA VMM allocations such as PyTorch"
+                         " expandable_segments.";
+        }
+
         // flags must be 0: the PCIE-BAR1 mapping flag is rejected (error 801)
         // on some GPU/driver combinations (e.g. B200).
         result = cuMemGetHandleForAddressRange(
-            &dmabuf_fd, allocBase, allocSize,
+            &dmabuf_fd, exportBase, exportSize,
             CU_MEM_RANGE_HANDLE_TYPE_DMA_BUF_FD, 0);
         if (result != CUDA_SUCCESS) {
             const char *errStr;
             cuGetErrorString(result, &errStr);
             LOG(ERROR) << "Failed to retrieve dmabuf for " << (uintptr_t)addr
-                       << " base=" << (uintptr_t)allocBase
-                       << " size=" << allocSize << " cuda error=" << errStr;
+                       << " base=" << (uintptr_t)exportBase
+                       << " size=" << exportSize << " cuda error=" << errStr;
 #if defined(USE_CUDA) || defined(USE_SUPA)
             cuDevicePrimaryCtxRelease(cuDev);
 #endif
@@ -464,7 +505,7 @@ int RdmaContext::exportDmabuf(void *addr, DmabufExport &out) {
         }
         out.method = DmabufExport::Method::kDmabufReg;
         out.fd = dmabuf_fd;
-        out.offset = (uintptr_t)addr - (uintptr_t)allocBase;
+        out.offset = exportOffset;
 #if defined(USE_CUDA) || defined(USE_SUPA)
         cuDevicePrimaryCtxRelease(cuDev);
 #endif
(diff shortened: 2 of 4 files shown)


alogfans (approved):
LGTM

staryxchen (commented):
Please fix format issue

enternal111:
i merged it into 0.3.12.post1, but it still failed,why

enternal111:
i merged it into 0.3.12.post1, but it still failed, why?
E20260821 10:22:52.325659 20837 rdma_context.cpp:629] Failed to register memory 0xb64fc96000 length 271056896 dmabuf_offset 0: Bad address [14]

hope1262946533:
> Please fix format issue

Now it maybe work, thanks.

hope1262946533:
> i merged it into 0.3.12.post1, but it still failed, why? E20260821 10:22:52.325659 20837 rdma_context.cpp:629] Failed to register memory 0xb64fc96000 length 271056896 dmabuf_offset 0: Bad address [14]

Try use below env variables:
PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True,large_segment_size_mb:4096
WITH_NVIDIA_PEERMEM=0 ...

hope1262946533:
@enternal111 
Try use below env variables:
PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True,large_segment_size_mb:4096
WITH_NVIDIA_PEERMEM=0 ...

hope1262946533:
@staryxchen 
Now it maybe works, thanks.

enternal111:
> > i merged it into 0.3.12.post1, but it still failed, why? E20260821 10:22:52.325659 20837 rdma_context.cpp:629] Failed to register memory 0xb64fc96000 length 271056896 dmabuf_offset 0: Bad address [14]
> 
> Try use below env variables: PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True,large_segment_size_mb:4096 WITH_NVIDIA_PEERMEM=0 ...



> @enternal111 Try use below env variables: PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True,large_segment_size_mb:4096 WITH_NVIDIA_PEERMEM=0 ...

my kernel is low than 0.5.12, so this pr won’t help when WITH_NVIDIA_PEERMEM=1, right? 
the log may cause misunderstanding by the way

hope1262946533:
> > > i merged it into 0.3.12.post1, but it still failed, why? E20260821 10:22:52.325659 20837 rdma_context.cpp:629] Failed to register memory 0xb64fc96000 length 271056896 dmabuf_offset 0: Bad address [14]
> > 
> > 
> > Try use below env variables: PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True,large_segment_size_mb:4096 WITH_NVIDIA_PEERMEM=0 ...
> 
> > @enternal111 Try use below env variables: PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True,large_segment_size_mb:4096 WITH_NVIDIA_PEERMEM=0 ...
> 
> my kernel is low than 0.5.12, so this pr won’t help when WITH_NVIDIA_PEERMEM=1, right? the log may cause misunderstanding by the way

Setting WITH_NVIDIA_PEERMEM=0 instructs Mooncake to use the VMM-based method for memory registration, utilizing ibv_reg_dmabuf_mr instead of ibv_reg_mr_iova2.

enternal111:
> > > > i merged it into 0.3.12.post1, but it still failed, why? E20260821 10:22:52.325659 20837 rdma_context.cpp:629] Failed to register memory 0xb64fc96000 length 271056896 dmabuf_offset 0: Bad address [14]
> > > 
> > > 
> > > Try use below env variables: PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True,large_segment_size_mb:4096 WITH_NVIDIA_PEERMEM=0 ...
> > 
> > 
> > > @enternal111 Try use below env variables: PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True,large_segment_size_mb:4096 WITH_NVIDIA_PEERMEM=0 ...
> > 
> > 
> > my kernel is low than 0.5.12, so this pr won’t help when WITH_NVIDIA_PEERMEM=1, right? the log may cause misunderstanding by the way
> 
> Setting WITH_NVIDIA_PEERMEM=0 instructs Mooncake to use the VMM-based method for memory registration, utilizing ibv_reg_dmabuf_mr instead of ibv_reg_mr_iova2.

failed when i set WITH_NVIDIA_PEERMEM=0
[error.log](https://github.com/user-attachments/files/31310841/error.log)

hope1262946533:
> > > > > i merged it into 0.3.12.post1, but it still failed, why? E20260821 10:22:52.325659 20837 rdma_context.cpp:629] Failed to register memory 0xb64fc96000 length 271056896 dmabuf_offset 0: Bad address [14]
> > > > 
> > > > 
> > > > Try use below env variables: PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True,large_segment_size_mb:4096 WITH_NVIDIA_PEERMEM=0 ...
> > > 
> > > 
> > > > @enternal111 Try use below env variables: PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True,large_segment_size_mb:4096 WITH_NVIDIA_PEERMEM=0 ...
> > > 
> > > 
> > > my kernel is low than 0.5.12, so this pr won’t help when WITH_NVIDIA_PEERMEM=1, right? the log may cause misunderstanding by the way
> > 
> > 
> > Setting WITH_NVIDIA_PEERMEM=0 instructs Mooncake to use the VMM-based method for memory registration, utilizing ibv_reg_dmabuf_mr instead of ibv_reg_mr_iova2.
> 
> failed when i set WITH_NVIDIA_PEERMEM=0 [error.log](https://github.com/user-attachments/files/31310841/error.log)

I supply some answers of AI for your problem:

---

**Is this a Mooncake issue?**

No, this is **not a Mooncake bug**. Mooncake is correctly reporting that it cannot initialize because the underlying RDMA/InfiniBand environment is broken or misconfigured.

---

**What is the real problem?**

The real problem is a **failed RDMA (Remote Direct Memory Access) environment configuration**, specifically for GPU-Direct RDMA. Mooncake is the victim here, not the cause.

The log shows three critical failures:

1. **Missing `nvidia-peermem` Kernel Module**  
   `E20260821 12:30:19.433391 rdma_context.cpp:1343] DMA BUF supported required for GPU RDMA without nvidia-peermem`  
   The system requires the `nvidia-peermem` driver to enable direct GPU-to-RDMA communication (GPUDirect RDMA). This driver is either not installed or not loaded.

2. **RDMA Device Initialization Failed**  
   `E20260821 12:30:19.435154 rdma_context.cpp:209] Failed to open device mlx5_3 on port  with GID 3`  
   The Mellanox ConnectX NICs (`mlx5_0` to `mlx5_3`) cannot be opened, likely because the RoCEv2 GID table is not configured correctly (GID index 3 is missing or invalid).

3. **No Available RNIC**  
   `E20260821 12:30:19.448818 rdma_transport.cpp:827] RdmaTransport: No available RNIC`  
   After failing to open all four NICs, the RDMA transport has no devices left to use.

Additionally, earlier warnings like `get_ip failed` and `get_local_ip_by_nic failed` indicate that the network interface (IP `172.26.155.79`) itself may not be fully up or accessible.

---

**How to fix it (checklist)**

1. **Check and load `nvidia-peermem`**
   ```bash
   lsmod | grep nvidia_peermem
   sudo modprobe nvidia_peermem
   ```
   If this fails, reinstall or update the NVIDIA driver to a version that includes `nvidia-peermem`.

2. **Verify RDMA NICs and RoCE GID table**
   ```bash
   ibv_devinfo -v | grep GID
   ```
   Ensure that a valid RoCEv2 GID exists (usually at index 3). If not, assign a valid IP address to the RDMA interface and restart the RDMA service.

3. **Check the network interface status**
   Confirm that IP `172.26.155.79` is correctly assigned and the interface is up.

4. **Optionally, bypass RDMA for testing**
   Set this environment variable to force TCP/IP fallback:
   ```bash
   export NCCL_IB_DISABLE=1
   ```

5. **Verify basic RDMA functionality**
   Run a simple test:
   ```bash
   ibv_devinfo
   ib_write_bw
   ```

enternal111:
> ```shell
> lsmod | grep nvidia_peermem
> ```

nvidia_peermem and rdma works for sure, cause it is normal when not set WITH_NVIDIA_PEERMEM, and it will set to 1. my question is: will this pr help when WITH_NVIDIA_PEERMEM=1, cause it will fail request with Bad address [14].

hope1262946533:
> nvidia_peermem and rdma works for sure, cause it is normal when not set WITH_NVIDIA_PEERMEM, and it will set to 1. my question is: will this pr help when WITH_NVIDIA_PEERMEM=1, cause it will fail request with Bad address [14].

**As per the current Mooncake design, expandable_segments: True must be combined with WITH_NVIDIA_PEERMEM=0,** so that VMM addresses are handled through the dma-buf path — which is exactly the approach this PR takes. **What this PR fixes is the errno 22 that still occurs under precisely this combination (expandable_segments: True + WITH_NVIDIA_PEERMEM=0)**.

With WITH_NVIDIA_PEERMEM=1 (the default), GPU memory goes through the nvidia-peermem path, which cannot pin CUDA VMM memory — that is why requests fail with Bad address [14]. That path is intentionally not modified by this PR.

About your startup failure with WITH_NVIDIA_PEERMEM=0: it happens before any registration. In that mode Mooncake requires CU_DEVICE_ATTRIBUTE_DMA_BUF_SUPPORTED = 1 for every GPU-RNIC pair; on your hosts the attribute returns 0, so all RNICs are disabled. Please check it with cuDeviceGetAttribute() — if it is 0, the GPU/driver must be upgraded to a dma-buf-capable combination. It is a prerequisite, unrelated to this PR.

catyans:
The two range-validation issues remain: `exportOffset + length` and the alignment round-up can overflow, and `sysconf(_SC_PAGESIZE)` is used without checking for failure. Please validate the page size and use checked/subtraction-based range arithmetic before calling `cuMemGetHandleForAddressRange`; this can be covered with a GPU-independent helper test.

alogfans:
@staryxchen Do you think it is ok to merge?

staryxchen:
> @staryxchen Do you think it is ok to merge?

I'll take a close look at it and run some tests today.

staryxchen (approved):
I did a basic local check of this change. I could confirm the VMM root cause on my machine. But I could not run the dma-buf / `ibv_reg_dmabuf_mr` path end-to-end: the environment reports `CU_DEVICE_ATTRIBUTE_DMA_BUF_SUPPORTED=0` and `cuMemGetHandleForAddressRange()` returns "operation not supported" (driver 535.161.07). 

From code review the change looks correct and appropriately scoped, LGTM.

PR #3543, opened 2026-08-20
[TransferEngine] Fix TCP endpoint refresh correctness

jacklin78911-collab:
## Description

This PR fixes two TCP endpoint-refresh correctness gaps in Transfer Engine.

First, after `syncSegmentCache()` successfully refreshes a logical segment, the corresponding cached RPC host entry is invalidated so the next `getRpcMetaEntry()` lookup reloads the authoritative host instead of returning stale metadata.

Second, TCP lane pooling now tracks the current `ConnectionKey` for each logical peer. When a logical peer moves to a different endpoint, the old endpoint group stops accepting new work for that peer, retains ownership of already accepted work, and retires only after queued, pending, and busy work has completed.

Idle sockets, timers, resolvers, and late callbacks are cleaned up using the existing group state and `operation_epoch` rules.

The retirement check also preserves a shared endpoint when another logical peer still maps to the same `{host, port}`.

Related to #2930.

## Module

- Transfer Engine (`mooncake-transfer-engine`)

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

Deterministic coverage includes:

- refreshed metadata routes new work to the new TCP endpoint;
- stale RPC host cache is invalidated after explicit metadata refresh;
- an old endpoint group retires after a logical peer moves to a new endpoint;
- one logical peer moving away from a shared endpoint does not retire the group while another peer still uses it;
- sequential A → B → C endpoint changes retire stale groups correctly;
- endpoint retirement racing with busy completion preserves exactly-once ownership;
- endpoint retirement racing with shutdown preserves accepted work and cleanup semantics.

Final local validation:

- Normal `tcp_write_visibility_test`: 36/36 passed.
- Targeted ASan/LSan/UBSan: 7/7 passed with leak detection enabled and no sanitizer report.
- `clang-format --dry-run --Werror`: passed.
- `git diff --check upstream/main..HEAD`: passed.

Targeted tests were also exercised under TSan instrumentation during development. Test assertions passed, but TSan still reported existing project baseline races in `Transport::Slice` status access.

No new TSan report pointed to endpoint-retirement ownership, `retiring_groups`, group mutex state, or `operation_epoch`.

This is not claimed as a clean TSan pass.

`pre-commit` was unavailable in the current local environment, so it is not claimed as passed.

## Scope and limitations

This PR does not solve:

- same host/port G1 → G2 generation detection;
- concurrent stale-descriptor publication without metadata generation;
- a global peer-group bound;
- byte-based queue bounds;
- the existing transfer-status synchronization race;
- multi-rail behavior;
- TENT TCP behavior.

## Checklist

- [x] The changed C++ files pass `clang-format --dry-run --Werror`.
- [x] I have added tests to prove the changes are effective.
- [x] Documentation is not required for this scoped correctness fix.
- [ ] `pre-commit` passed locally — unavailable in the current environment.

## AI Assistance Disclosure

- [x] AI tools were used.

Codex assisted with implementation review, deterministic test design, local validation, and preparation of the PR description. The human submitter remains responsible for reviewing and understanding every changed line.

### DIFF
--- a/mooncake-transfer-engine/include/transfer_metadata.h
+++ b/mooncake-transfer-engine/include/transfer_metadata.h
@@ -121,7 +121,12 @@ class TransferMetadata {
         // this is for ascend
         RankInfoDesc rank_info;
 
-        int tcp_data_port;
+        // TCP data-plane endpoint. Keeping the routable host beside the port
+        // makes one SegmentDesc an immutable endpoint snapshot for clients.
+        // Older descriptors may omit tcp_data_host; getSegmentDesc() fills it
+        // once from legacy RPC metadata before caching the descriptor.
+        std::string tcp_data_host;
+        int tcp_data_port{0};
         // TCP data-plane protocol version advertised by this segment's
         // server. v2 adds acknowledged WRITE framing and status-prefixed
         // READ responses (#2086); absent/1 = legacy unacknowledged framing.
@@ -261,6 +266,10 @@ class TransferMetadata {
     void dumpMetadataContentUnlocked();
 
    private:
+    std::shared_ptr<SegmentDesc> getSegmentDescInternal(
+        const std::string &segment_name, bool force_rpc_update);
+    int getRpcMetaEntryInternal(const std::string &server_name,
+                                RpcMetaDesc &desc, bool force_update);
     int encodeSegmentDesc(const SegmentDesc &desc, Json::Value &segmentJSON);
     std::shared_ptr<TransferMetadata::SegmentDesc> decodeSegmentDesc(
         Json::Value &segmentJSON, const std::string &segment_name);
--- a/mooncake-transfer-engine/include/transport/tcp_transport/tcp_transport.h
+++ b/mooncake-transfer-engine/include/transport/tcp_transport/tcp_transport.h
@@ -217,6 +217,7 @@ class TcpTransport : public Transport {
         asio::io_context::executor_type executor;
     };
 
+    struct ConnectionLaneState;
     struct PeerConnectionGroup;
 
     struct ConnectionLane {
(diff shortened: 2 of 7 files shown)


alogfans (commented):
The RPC host and TCP port are read from two independently updated caches. If `syncSegmentCache()` runs concurrently with `startTransfer()`, this path may combine an old `SegmentDesc::tcp_data_port` with a newly refreshed RPC host. Please expose the complete TCP endpoint through a single immutable metadata snapshot, or otherwise make the host and port refresh/read atomic.

jacklin78911-collab:
@alogfans Thanks for catching this. Fixed in `d04e9dfb`.

`SegmentDesc` now carries `tcp_data_host` together with `tcp_data_port`, so the TCP endpoint is represented by a single metadata snapshot. `startTransfer()` constructs the endpoint exclusively from this descriptor snapshot and no longer combines `tcp_data_port` with a separately refreshed RPC host.

For backward compatibility, descriptors without `tcp_data_host` resolve the legacy RPC host once and populate the missing field before the descriptor is cached.

Therefore, a concurrent metadata refresh can only switch from the complete old `(host, port)` endpoint to the complete new endpoint, rather than producing a mixed pair.

I also added `ConcurrentEndpointRefreshNeverCombinesHostAndPortSnapshots`, which deterministically overlaps `startTransfer()` with metadata refresh and verifies that requests observe either the complete old endpoint or the complete new endpoint, never a mixed host/port pair.

alogfans (approved):
LGTM. Will merge after fixing conflicts.

jacklin78911-collab:
@alogfans Thank you.The conflicts are resolved.

PR #3545, opened 2026-08-20
[Store] Fix promotion-on-hit delivery starvation and retry loss

LujhCoconut:
## Description

Fixes the nightly `test_promotion_after_repeated_hits` flake ([run 32276229813](https://github.com/kvcache-ai/Mooncake/actions/runs/32276229813/job/96162038864)).

**Problem.** The test overflows a 32MB segment, waits for offload to turn some keys LOCAL_DISK-only, reads one `cold_key` to clear `promotion_admission_threshold=1`, then asserts a MEMORY replica appears. It flakes because the promotion path races three independent clocks (10ms eviction / 10s client heartbeat / 25s test sleep):

1. A read-only `batch_get_replica_desc` (the test's own phase-2 introspection) admits *every* LOCAL_DISK-only key, flooding the promotion mailbox.
2. The mailbox is an `unordered_map` and each heartbeat delivers only `promotion_max_per_heartbeat=1`, hash-arbitrary — so the target key is starved.
3. A watermark-rejected candidate gets 8 retries (~2.3s) before permanent deletion, while the condition it waits on (offload draining `used_ratio`) is 10–30s scale.
4. `NotifyPromotionFailure` (and `PromotionAllocStart` failing with `NO_AVAILABLE_HANDLE`) drops the task without re-recording a candidate — every delivery gets exactly one chance.

**Fix.**

- `local_ssd/manager`: deliver promotions in **recency order** — a monotonic `seq` is refreshed on enqueue, duplicate enqueue, and a new `TouchPromotion` (called when a read hits an already in-flight key). `TakePromotions` ranks by `seq` desc, so the most-recently-read key is delivered first regardless of the hash order of the map.
- `master_service`: re-record a retry candidate in `NotifyPromotionFailure` for transient execution failures (using `CountMinSketch::count()`, read-only, so the executor-side failure does not pollute the frequency signal). The re-record is bounded by a per-chain `execution_failures` counter propagated candidate → task at admission (`kMaxPromotionExecutionFailures = 3`): without it a persistently-failing key (e.g. a broken SSD file that still has a LOCAL_DISK replica) would re-record → re-admit → fail → re-record *forever*, because admission consumes the candidate, so every re-record is a fresh insert and the retry budget / TTL never apply. On give-up only the self-sustaining cycle stops — a genuine read can still re-admit the key with a fresh count. Give-ups are observable via the new `master_promotion_execution_gave_up_total` counter.
- Retry budget rescaled to the heartbeat timescale: `kPromotionCandidateMaxRetries` 8→64, max backoff 1s→5s, TTL 60s→300s, so candidates survive long enough for their gate to actually clear.
- `test_promotion_on_hit.py`: replace fixed `time.sleep(25)` with condition polling (`wait_until`); each poll is a single-key `batch_get_replica_desc`, which re-fires the promotion gate — the wait becomes self-healing instead of a bet on the phase.
- `scripts/run_tests.sh`: pin the clock phase with `--promotion_max_per_heartbeat=16` and `MOONCAKE_OFFLOAD_HEARTBEAT_INTERVAL_SECONDS=2`.

## Module

- [x] CI/CD

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

**Test commands:**

```bash
# unit tests (mailbox recency, failure re-record, retry budget)
cmake --build build --target local_ssd_test promotion_on_hit_test && \
    ctest --test-dir build -R "local_ssd|promotion_on_hit" --output-on-failure

# nightly e2e
TEST_PROMOTION_ON_HIT=1 bash scripts/run_tests.sh
```

**Test results:**

- [x] Unit tests pass — added: mailbox recency delivery / cross-tick order / client-scoped touch, `NotifyFailureRecordsCandidateAndRetries`, `NotifyFailureGivesUpAfterMaxExecutionFailures`, retry-budget exhaustion follows `kPromotionCandidateMaxRetries`.
- [x] Integration tests pass — e2e `test_promotion_on_hit.py` **5/5** against a branch-built `mooncake_master` (fresh master + fresh storage dir per run, CI-identical flags, ~40s/run; cold keys varied across runs — poh_1/2/4/11 — all promoted; histograms consistently showed ~5 LOCAL_DISK-only keys whose admission burst drains via `promotion_max_per_heartbeat=16`). Note for local repro: the wheel's `store.so` must be built from the same commit as the master — an older installed wheel fails `MountSegment` with `invalid rpc arg` (RPC struct mismatch), which is an environment issue, not a test failure.

## Checklist

## AI Assistance Disclosure

- [x] AI tools were used — assisted analysis and code review of the mechanism (root-cause tracing of the three-clock race and the mailbox/retry/failure semantics). The human submitter reviewed every changed line.

### DIFF
--- a/mooncake-store/include/local_ssd/manager.h
+++ b/mooncake-store/include/local_ssd/manager.h
@@ -29,6 +29,9 @@ class LocalSsdTaskMailbox {
     ErrorCode EnqueuePromotion(PromotionTaskItem task);
     std::vector<PromotionTaskItem> TakePromotions(size_t max_items);
     bool RemovePromotion(const TenantId& tenant_id, std::string_view key);
+    // Re-mark a queued promotion as most-recently-touched. No-op (returns
+    // false) when the key is not queued, e.g. already taken by a heartbeat.
+    bool TouchPromotion(const TenantId& tenant_id, std::string_view key);
 
     void RequestRemoveAll();
     bool ConsumeRemoveAll();
@@ -38,8 +41,17 @@ class LocalSsdTaskMailbox {
     bool enable_offloading_ GUARDED_BY(mutex_);
     std::unordered_map<std::string, OffloadTaskItem> pending_offloads_
         GUARDED_BY(mutex_);
-    std::unordered_map<std::string, PromotionTaskItem> pending_promotions_
+    // Dedup map + recency order. seq is monotone; enqueue, a duplicate
+    // enqueue, and TouchPromotion all re-mark the entry with a fresh seq, so
+    // TakePromotions can deliver the most-recently-touched key first instead
+    // of an arbitrary (hash-order) one.
+    struct PromotionEntry {
+        PromotionTaskItem task;
+        uint64_t seq;
+    };
+    std::unordered_map<std::string, PromotionEntry> pending_promotions_
         GUARDED_BY(mutex_);
+    uint64_t promotion_seq_ GUARDED_BY(mutex_){0};
     bool pending_remove_all_ GUARDED_BY(mutex_){false};
 
     friend class LocalSsdManager;
@@ -83,6 +95,8 @@ class LocalSsdManager {
         const UUID& client_id, size_t max_items);
     bool RemovePromotion(const UUID& client_id, const TenantId& tenant_id,
                          std::string_view key);
+    bool TouchPromotion(const UUID& client_id, const TenantId& tenant_id,
+                        std::string_view key);
 
     void RequestRemoveAll();
     void RequestRemoveAll(const std::vector<UUID>& clients);
--- a/mooncake-store/include/master_metric_manager.h
+++ b/mooncake-store/include/master_metric_manager.h
@@ -311,6 +311,7 @@ class MasterMetricManager {
     void inc_promotion_completed_bytes(int64_t bytes);
     void inc_promotion_expired(int64_t val = 1);
     void inc_promotion_failed(int64_t val = 1);
+    void inc_promotion_execution_gave_up(int64_t val = 1);
     void inc_promotion_cancelled(int64_t val = 1);
     void inc_promotion_rejected_frequency(int64_t val = 1);
     void inc_promotion_rejected_watermark(int64_t val = 1);
@@ -336,6 +337,7 @@ class MasterMetricManager {
     int64_t get_promotion_completed_bytes();
     int64_t get_promotion_expired();
     int64_t get_promotion_failed();
+    int64_t get_promotion_execution_gave_up();
     int64_t get_promotion_cancelled();
     int64_t get_promotion_rejected_frequency();
     int64_t get_promotion_rejected_watermark();
@@ -695,6 +697,7 @@ class MasterMetricManager {
     ylt::metric::counter_t promotion_completed_bytes_;
     ylt::metric::counter_t promotion_expired_;
     ylt::metric::counter_t promotion_failed_;
+    ylt::metric::counter_t promotion_execution_gave_up_;
     ylt::metric::counter_t promotion_cancelled_;
     ylt::metric::counter_t promotion_rejected_frequency_;
     ylt::metric::counter_t promotion_rejected_watermark_;
--- a/mooncake-store/include/master_service.h
+++ b/mooncake-store/include/master_service.h
@@ -1530,6 +1530,7 @@ class MasterService {
         kWatermark,
         kQueueCap,
         kPushFailed,
+        kExecutionFailed,
     };
 
     struct PromotionCandidate {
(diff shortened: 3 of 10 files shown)


he-yufeng (commented):
The retry-budget sizing reads right (the gates clear on the 10s heartbeat scale, so 2.3s was never enough), and the count()-rather-than-increment() choice on the sketch is exactly correct: an executor-side failure is not demand signal, and incrementing there would pollute the admission gate.

One thing worth a second look: the re-record goes through RecordOrUpdateCandidate, whose update path resets retry_count to 0 and refreshes last_seen (the TTL anchor). For a key that keeps failing on execution (say SSD throttling that lasts an hour), each NotifyPromotionFailure re-records with a fresh 64-retry budget and a fresh 5-minute TTL, so the documented bounds never actually exhaust. The permanent-condition erases you listed stop the loop only when the condition is detectably permanent; a persistent-but-transient-looking failure spins. If that matters, the kExecutionFailed re-record wants a variant that keeps the original first_seen and increments retry_count instead of resetting the budget.

LujhCoconut:
> The retry-budget sizing reads right (the gates clear on the 10s heartbeat scale, so 2.3s was never enough), and the count()-rather-than-increment() choice on the sketch is exactly correct: an executor-side failure is not demand signal, and incrementing there would pollute the admission gate.
> 
> One thing worth a second look: the re-record goes through RecordOrUpdateCandidate, whose update path resets retry_count to 0 and refreshes last_seen (the TTL anchor). For a key that keeps failing on execution (say SSD throttling that lasts an hour), each NotifyPromotionFailure re-records with a fresh 64-retry budget and a fresh 5-minute TTL, so the documented bounds never actually exhaust. The permanent-condition erases you listed stop the loop only when the condition is detectably permanent; a persistent-but-transient-looking failure spins. If that matters, the kExecutionFailed re-record wants a variant that keeps the original first_seen and increments retry_count instead of resetting the budget.

Thanks for the careful read — the diagnosis is correct, and it's a real hole. But the fix you suggest targets the wrong branch of the record path, so it wouldn't actually close it. I've closed it a different way, which is already pushed as a follow-up commit.

**The diagnosis holds.** Confirmed: a key whose promotion fails at execution time does not consume the 64-retry budget or the 300s TTL. The reason is that admission *consumes* the candidate — `TryPushPromotionQueue` calls `EraseCandidate` immediately before emplacing the `PromotionTask` — so the next `NotifyPromotionFailure` re-record is always a **fresh insert**, with a fresh `retry_count = 0` and a fresh `last_seen` anchor. The 64-retry / TTL bounds only ever burn when the *gate* rejects the key (watermark / queue-cap), where the candidate stays in the map and the `BackoffCandidate` path actually increments the counters. The execution-failure cycle never touches that path.

**But the suggested fix lands on the `update` branch, which this cycle never reaches.** The reset behavior you flagged (`retry_count → 0`, `last_seen` refresh) lives in `RecordOrUpdateCandidate`'s update path — the branch taken only when the candidate *still exists*, i.e. the gate rejected the key and it's parked in the candidate map. An execution-failure cycle is the opposite shape: admission *succeeds*, the candidate is erased, and the re-record is a fresh insert. Keeping the original `first_seen` and incrementing `retry_count` on the update path changes nothing for this loop, because that branch is never entered.

**The fix that actually bounds it: propagate the failure count across the admission boundary.** A per-chain `execution_failures` counter is copied candidate → task at admission, and `NotifyPromotionFailure` re-records with `execution_failures + 1` until `kMaxPromotionExecutionFailures` (= 3), after which it stops re-recording. Because the counter rides the task (which survives admission) rather than the candidate (which is consumed by admission), the bound can't be reset by the erase. On give-up only the self-sustaining cycle stops — a genuine read still re-admits the key with a fresh count of 0. Give-ups are observable via the new `master_promotion_execution_gave_up_total` counter.

A side benefit of bounding it by a small task-side counter rather than by retry-count arithmetic: a poison key can occupy the recency-ordered delivery queue for at most a few heartbeat ticks (the initial admission + 3 re-records) before it stops re-queueing itself, so it can't monopolize delivery the way an unbounded self-sustaining cycle would.

Happy to point you at the exact commit if useful.

he-yufeng:
You're right, and the admission-consumes-the-candidate point is what I missed: the failure cycle never enters the update branch at all, since TryPushPromotionQueue erases the candidate when it emplaces the task. A counter riding the task survives exactly the erase that resets the candidate, so the bound actually holds, and a genuine read still re-admits fresh. The shape is correct. Thanks for closing it properly.

fcczzz (approved):
Thanks for the fix! The retry handling and recency-ordered mailbox changes look sound, and the focused tests pass.

One non-blocking follow-up: the existing-task TouchPromotion path currently comes after the watermark check. If a task is queued while usage is below the watermark and the ratio rises before a subsequent read, that read returns at the watermark gate without refreshing the task’s mailbox recency. The task will still eventually drain, so this does not appear to block the current flake fix, but the “recent read moves an in-flight key forward” behavior does not hold under high memory pressure.

It may be worth handling the existing-task touch before applying the watermark gate to new admissions, or covering this ordering in a follow-up test.

Otherwise, LGTM.

ykwd (approved):

stmatengss (approved):

PR #3547, opened 2026-08-20
[transfer-engine] Ship intra-node NVLink transport in x86_64 CUDA wheels

ShangmingCai:
## Why

`IntraNodeNvlinkTransport` is gated behind `USE_INTRA_NVLINK`, which is `OFF` by default (`mooncake-common/common.cmake`) and is **not** set by any release, pre-release, or CI build profile — its only `=ON` occurrence in the tree is `docs/source/getting_started/build.md`. So no published wheel of any variant contains the transport, and `MC_INTRANODE_NVLINK` cannot do anything for wheel users.

Verified on the published `mooncake-transfer-engine-cuda13==0.3.12.post1` wheels (`mooncake/engine.so`):

| symbol | x86_64 | aarch64 |
|---|---|---|
| `RdmaTransport` / `TcpTransport` | present | present |
| `NvlinkTransport` (MNNVL) | absent | present |
| `IntraNodeNvlink` | **absent** | **absent** |

Downstream this matters for single-node PD disaggregation: SGLang documents `SGLANG_MOONCAKE_CUSTOM_MEM_POOL=INTRA_NODE_NVLINK` + `MC_INTRANODE_NVLINK=true` for A100/H20/H100 (`docs/advanced_features/pd_disaggregation.mdx`), but on a prebuilt wheel that configuration silently does nothing and the transfer falls back to RDMA loopback through the local HCA.

## What this changes

**1. Build the x86_64 CUDA wheels with the transport** — `-DUSE_INTRA_NVLINK=ON` added to the `cuda:x86_64` and `cuda13:x86_64` profiles in `_build-wheel.yaml`. The transport is opt-in at runtime (`MC_INTRANODE_NVLINK`), so this only makes it reachable. arm64 is left alone; those profiles already carry `USE_MNNVL=ON` and can be a follow-up if wanted.

**2. Keep the RDMA/TCP fallback correct when only `USE_INTRA_NVLINK` is set.** This is required by (1), not cosmetic. `USE_INTRA_NVLINK` moves a build into the `#elif defined(USE_MNNVL) || defined(USE_INTRA_NVLINK) || defined(USE_MUSA)` selection block, whose no-HCA path requests the *cross-node* protocol:

```cpp
} else if (force_gpu_p2p || local_topology_->getHcaList().empty()) {
    installTransport("nvlink", nullptr);   // only constructible #ifdef USE_MNNVL
```

`MultiTransport` only constructs `"nvlink"` under `#ifdef USE_MNNVL`, so on an intra-only build a host with no RDMA HCA would get `nullptr` → `Failed to install NVLink transport` → `return -1`, i.e. engine init fails where it previously fell back to TCP. That would break every no-HCA x86_64 wheel user (TCP-only deployments, dev boxes, CI). The patch gates the cross-node branch on the protocol actually being compiled, preserves `MC_FORCE_HCA` semantics from the non-NVLink path, and adds the explicit TCP fallback.

Behavior is unchanged for every currently shipped configuration:

| build | condition | before | after |
|---|---|---|---|
| MNNVL / MUSA (arm64 wheels) | any | unchanged | unchanged — the new branches are unreachable when the GPU-P2P protocol is compiled |
| x86_64 CUDA | HCA present | RDMA (`autoDiscoverTransport()` → `"rdma"`; barex/efa are not compiled in this variant) | RDMA |
| x86_64 CUDA | no HCA | TCP | TCP |
| x86_64 CUDA | no HCA + `MC_FORCE_HCA` | RDMA | RDMA |
| x86_64 CUDA | `MC_FORCE_MNNVL` | ignored | ignored + warning that MNNVL is not compiled |
| x86_64 CUDA | `MC_INTRANODE_NVLINK` | ignored | **intra-node NVLink** |

**3. Docs** — note in `supported-protocols.md` that the prebuilt x86_64 CUDA wheels now satisfy the `USE_INTRA_NVLINK` requirement, plus the runtime env var.

## Testing

Not compiled locally — I don't have a CUDA host for this. The change is intended to be validated by this PR's own CI: `ci.yml`'s `build-wheel` and `build-wheel-cu13` jobs call `_build-wheel.yaml`, so both x86_64 CUDA profiles compile with the new flag here (may need the `run-ci` label). Runtime verification of `nvlink_intra` on NVLink hardware, and whether it should also be enabled for arm64, are worth a maintainer's call.

---

### DIFF
--- a/.github/workflows/_build-wheel.yaml
+++ b/.github/workflows/_build-wheel.yaml
@@ -64,7 +64,7 @@ jobs:
               arch_list='8.0;9.0'
               ep_versions='2.11.0;2.12.0;2.12.1;2.13.0'
               build_nvlink=true
-              cmake_args='-DBUILD_UNIT_TESTS=OFF -DUSE_HTTP=ON -DUSE_ETCD=ON -DUSE_CUDA=ON -DWITH_EP=ON -DSTORE_USE_ETCD=ON -DENABLE_SCCACHE=ON -DCMAKE_BUILD_TYPE=Release'
+              cmake_args='-DBUILD_UNIT_TESTS=OFF -DUSE_HTTP=ON -DUSE_ETCD=ON -DUSE_CUDA=ON -DUSE_INTRA_NVLINK=ON -DWITH_EP=ON -DSTORE_USE_ETCD=ON -DENABLE_SCCACHE=ON -DCMAKE_BUILD_TYPE=Release'
               ;;
             cuda13:x86_64)
               variant_flag=CU13_BUILD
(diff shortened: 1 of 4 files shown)


Aionw (approved):
LGTM for workflow side.

alogfans (approved):
LGTM

PR #1622, opened 2026-03-06
Fix NVLink IPC offset corruption for sub-allocated GPU tensors

ishandhanani:
## Problem

`registerLocalMemory()` in `IntraNodeNvlinkTransport` stores the caller-provided pointer as the buffer base address. When frameworks like PyTorch sub-allocate tensors within larger `cudaMalloc` segments (via caching allocators), this pointer is not the `cudaMalloc` base. Since `cudaIpcGetMemHandle` always returns a handle for the entire `cudaMalloc` allocation, `relocateSharedMemoryAddress()` computes the wrong offset on the target side -- the sub-allocation offset within the segment is lost.

This corrupts all small tensors (<1MB) that share a `cudaMalloc` segment. Large tensors (>1MB) are unaffected because PyTorch gives them their own segment.

## Fix

Use `cuMemGetAddressRange()` to resolve the true `cudaMalloc` base address and allocation size before registration. Register at segment granularity and skip duplicates when multiple tensors share the same segment.

## Testing

Reproduced with a two-process test (seed on GPU 0, target on GPU 1) transferring tensors of varying sizes (64B, 16KB, 128KB, 512KB, 4MB). Before fix: 4/6 tensors corrupted. After fix: 6/6 correct.

Also validated end-to-end with SGLang remote instance weight loading (Qwen3-235B-A22B-FP8, TP=4, NVLink transport).

### Note

Claude code helped me find this bug while working on sharding weights for transfer in https://github.com/sgl-project/sglang/pull/19983. But I reviewed the fix and have tried it and it works

PR #3548, opened 2026-08-20
[Bugfix][TransferEngine] Register HIP IPC memory at allocation granularity

### LINKED ISSUE #3546: [Bug]: HIP IPC transfers silently return the wrong bytes for sub-allocated device buffers
## Environment

- 8× AMD Instinct MI308X (gfx942), ROCm 7.2.0
- PyTorch 2.9.1 (`torch.version.hip == 7.2.26015-fc0010cf6a`), Python 3.10.9
- Mooncake `main` @ `f90ae69`, built with `-DUSE_HIP=ON -DCMAKE_HIP_ARCHITECTURES=gfx942`
- `protocol="hip"`, `P2PHANDSHAKE`, single node, intra-node GPU↔GPU

## What happens

When several device buffers that live inside **one** allocation are registered
individually, a peer reading them back gets the wrong data. Every
`transfer_sync_read` returns `0`, the byte counts match, nothing is logged — only
the contents are wrong. Each buffer comes back holding the *first* buffer's bytes.

We ran into this in a disaggregated RL weight-transfer path that registers each
FSDP2 parameter shard separately so inference workers can read straight into
their own parameter memory. It first looked like partial corruption
(e.g. `3141269/4194304 elements differ`), because the parameters that happened to
land at an allocation base were correct and the rest were not.

## Steps to reproduce

Two processes are required. A loopback transfer inside a single process never
opens an IPC handle, so this cannot be observed that way — which is also why
`mooncake-wheel/tests/test_transfer_on_hip.py` passes on an affected build.

```python
import multiprocessing as mp
import torch

N = 1 << 20  # bytes per buffer


def source(meta, done):
    torch.cuda.set_device(0)
    pool = torch.empty(4 * N, dtype=torch.uint8, device="cuda:0")
    views = [pool[i * N : (i + 1) * N] for i in range(4)]
    for i, v in enumerate(views):
        v.fill_(i + 1)  # buffer i holds the byte i+1
    torch.cuda.synchronize()

    e = TransferEngine()
    assert e.initialize("127.0.0.1", "P2PHANDSHAKE", "hip", "") == 0
    for v in views:
        assert e.register_memory(v.data_ptr(), N) == 0
    meta.put((f"127.0.0.1:{e.get_rpc_port()}", [v.data_ptr() for v in views]))
    done.get(timeout=120)


def sink(meta, done):
    torch.cuda.set_device(0)
    segment, addresses = meta.get(timeout=120)

    e = TransferEngine()
    assert e.initialize("127.0.0.1", "P2PHANDSHAKE", "hip", "") == 0
    pool = torch.zeros(4 * N, dtype=torch.uint8, device="cuda:0")
    views = [pool[i * N : (i + 1) * N] for i in range(4)]
    for v in views:
        assert e.register_memory(v.data_ptr(), N) == 0
    codes = [
        e.transfer_sync_read(segment, v.data_ptr(), a, N)
        for v, a in zip(views, addresses)
    ]
    torch.cuda.synchronize()
    print("return codes:", codes)
    print("expected [1, 2, 3, 4], got", [int(v[0]) for v in views])
    done.put(True)


if __name__ == "__main__":
    ctx = mp.get_context("spawn")
    meta, done = ctx.Queue(), ctx.Queue()
    procs = [ctx.Process(target=f, args=(meta, done)) for f in (source, sink)]
    for p in procs:
        p.start()
    for p in procs:
        p.join()
```

Observed:

```
return codes: [0, 0, 0, 0]
expected [1, 2, 3, 4], got [1, 1, 1, 1]
```

Expected: `got [1, 2, 3, 4]`.

The four views above make the sub-allocation explicit, but nothing here depends
on that: four separately allocated small tensors reproduce it just as well,
because PyTorch's caching allocator places them in one segment anyway. Measured
with `hipMemGetAddressRange` on four separately allocated 2 MiB tensors:

```
buf0  offset=0        at allocation base
buf1  offset=2097152  NOT at base   (all four inside one 20 MiB allocation)
buf2  offset=4194304  NOT at base
buf3  offset=6291456  NOT at base
```

`batch_register_memory` behaves the same, since `registerLocalMemoryBatch` loops
over `registerLocalMemory`.

## What we have narrowed it down to

`hipIpcGetMemHandle` always exports the whole allocation, but
`HipTransport::registerLocalMemory` records the caller's sub-range address as the
buffer base (`hip_transport.cpp:688,696`):

```cpp
hipIpcGetMemHandle(&handle, addr);   // handle covers the whole allocation
desc.addr = (uint64_t)addr;          // but we claim the buffer starts here
desc.length = length;
```

The reader relocates with (`hip_transport.cpp:772`):

```cpp
dest_addr = dest_addr - entry.addr + (uint64_t)shm_addr;
```

`shm_addr` maps the whole allocation while `entry.addr` is the sub-range, so every
sub-range collapses onto the allocation base. That matches the observed output
exactly: buffers 1–3 return buffer 0's bytes, and buffer 0 is correct because its
offset happens to be zero.

This looks like the same family as #1622, #1831 and #2035, where the CUDA paths
were changed to register the allocation base instead of the requested range. It
seems distinct from #2684 / #2752, which are about multi-protocol `rdma+hip`
transport selection and segment-descriptor failures rather than silently wrong
payloads.

### Before submitting...

PACTHEMAN123:
## Description

Fixes #3546.

`hipIpcGetMemHandle` always exports the entire `hipMalloc` allocation, but the
HIP IPC registration path recorded the caller's sub-range address as the buffer
base. The reader's relocation (`dest_addr - entry.addr + shm_addr`) therefore
collapsed every sub-range onto the allocation base, and each buffer silently
returned the first buffer's bytes with `rc=0`. Framework caching allocators
sub-allocate constantly, so any caller registering individual tensors out of one
allocation is affected — including `batch_register_memory`, which loops over the
same function.

The CUDA IPC paths had exactly this bug and it was fixed by registering the
allocation base instead of the requested range — #1622 for
`intranode_nvlink_transport` and #1831 for TENT. The HIP port was missed: its only
`hipMemGetAddressRange` calls are in the fabric-memory branch and in the free
logic, never in the IPC branch. This applies the same fix there: resolve the real
base with `hipMemGetAddressRange`, skip allocations already exported, take the IPC
handle for the base, record the whole allocation in `BufferDesc`, and key
unregistration off that base as well, matching
`IntraNodeNvlinkTransport::registerLocalMemory` and `unregisterLocalMemory`.

## Module

- [x] Python Wheel (`mooncake-wheel`) — regression test only
- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

8× AMD Instinct MI308X (gfx942), ROCm 7.2.0, PyTorch 2.9.1
(`torch.version.hip == 7.2.26015-fc0010cf6a`), Python 3.10.9, built from `main` @
`f90ae69` with `-DUSE_HIP=ON -DCMAKE_HIP_ARCHITECTURES=gfx942`.

The reproducer in #3546 goes from `got [1, 1, 1, 1]` to `got [1, 2, 3, 4]`, and
flips back when the pre-fix `engine.so` is swapped in again.

**Test commands:**

```bash
cd mooncake-wheel/tests
python -m unittest -v test_transfer_sub_allocation_on_hip   # new
python -m unittest -v test_transfer_on_hip                  # existing, unaffected
```

**Test results:**

The new test is the reproducer from #3546 turned into a regression test. Against
the pre-fix build:

```
AssertionError: Lists differ: [1, 1, 1, 1] != [1, 2, 3, 4]
FAILED (failures=1)
```

Against the fixed build: `OK`. `test_transfer_on_hip.py` is `OK` on both, which is
why the new test is a separate two-process file: a loopback transfer never opens
an IPC handle, so it cannot observe the relocation.

- [x] Unit tests pass

Note: the new test is hardware-gated with `unittest.SkipTest` and is not wired
into `scripts/run_tests.sh`, matching `test_transfer_on_hip.py` and
`test_transfer_on_cuda.py`, which are not listed there either. Let me know if you
would rather it lived somewhere CI can reach.

## Checklist

- [x] For changes >500 LOC: I have filed an RFC issue — n/a, 48 lines excluding tests

`pre-commit run --files` over the three changed files passes every hook, including
the clang-format 20 C/C++ hook. I did not run `--all-files`, since it rewrites
files unrelated to this change.

## AI Assistance Disclosure

- [ ] No AI tools were used

Qoder was used for the investigation, the patch and the test.

### DIFF
--- a/mooncake-transfer-engine/src/transport/hip_transport/hip_transport.cpp
+++ b/mooncake-transfer-engine/src/transport/hip_transport/hip_transport.cpp
@@ -697,8 +697,10 @@ int HipTransport::registerLocalMemory(void* addr, size_t length,
             return -1;
         }
 
-        // Skip if this hipMalloc block is already registered
-        if (registered_base_addrs_.count((uint64_t)base_ptr)) {
+        // Count aliases so the registration outlives every sub-range but one
+        auto it = registered_base_refs_.find((uint64_t)base_ptr);
+        if (it != registered_base_refs_.end()) {
+            it->second++;
             return 0;
         }
 
@@ -721,7 +723,7 @@ int HipTransport::registerLocalMemory(void* addr, size_t length,
 #endif
         int rc = metadata_->addLocalMemoryBuffer(desc, true);
         if (rc == 0) {
-            registered_base_addrs_.insert((uint64_t)base_ptr);
+            registered_base_refs_[(uint64_t)base_ptr] = 1;
         }
         return rc;
     }
@@ -788,8 +790,11 @@ int HipTransport::unregisterLocalMemory(void* addr, bool update_metadata) {
                          << "); memory may already be freed, using the "
                          << "provided address";
         }
+        // Keep the registration alive until the last sub-range releases it
         std::lock_guard<std::mutex> lock(register_mutex_);
-        registered_base_addrs_.erase((uint64_t)key_ptr);
+        auto it = registered_base_refs_.find((uint64_t)key_ptr);
+        if (it != registered_base_refs_.end() && --it->second > 0) return 0;
+        registered_base_refs_.erase((uint64_t)key_ptr);
     }
     return metadata_->removeLocalMemoryBuffer(key_ptr, update_metadata);
 }
--- a/mooncake-transfer-engine/src/transport/hip_transport/hip_transport.cpp
+++ b/mooncake-transfer-engine/src/transport/hip_transport/hip_transport.cpp
@@ -697,8 +697,13 @@ int HipTransport::registerLocalMemory(void* addr, size_t length,
             return -1;
         }
 
-        // Count aliases so the registration outlives every sub-range but one
-        auto it = registered_base_refs_.find((uint64_t)base_ptr);
+        // Re-registering the same address is idempotent; a new sub-range of
+        // an already exported allocation just takes another reference.
+        uint64_t base = (uint64_t)base_ptr;
+        if (!registered_addr_to_base_.emplace((uint64_t)addr, base).second) {
+            return 0;
+        }
+        auto it = registered_base_refs_.find(base);
         if (it != registered_base_refs_.end()) {
             it->second++;
             return 0;
@@ -708,6 +713,7 @@ int HipTransport::registerLocalMemory(void* addr, size_t length,
         hipIpcMemHandle_t handle;
         if (!checkHip(hipIpcGetMemHandle(&handle, base_ptr),
                       "HipTransport: hipIpcGetMemHandle failed")) {
+            registered_addr_to_base_.erase((uint64_t)addr);
             return -1;
         }
 
@@ -723,7 +729,9 @@ int HipTransport::registerLocalMemory(void* addr, size_t length,
 #endif
         int rc = metadata_->addLocalMemoryBuffer(desc, true);
         if (rc == 0) {
-            registered_base_refs_[(uint64_t)base_ptr] = 1;
+            registered_base_refs_[base] = 1;
+        } else {
+            registered_addr_to_base_.erase((uint64_t)addr);
         }
         return rc;
     }
@@ -775,28 +783,33 @@ int HipTransport::registerLocalMemory(void* addr, size_t length,
 }
 
 int HipTransport::unregisterLocalMemory(void* addr, bool update_metadata) {
-    void* key_ptr = addr;
     if (!use_fabric_mem_) {
-        void* base_ptr = nullptr;
-        size_t alloc_size = 0;
-        hipError_t result = hipMemGetAddressRange(
-            (hipDeviceptr_t*)&base_ptr, &alloc_size, (hipDeviceptr_t)addr);
-        if (result == hipSuccess) {
-            key_ptr = base_ptr;
-        } else {
-            (void)hipGetLastError();
-            LOG(WARNING) << "HipTransport: hipMemGetAddressRange failed for "
-                         << addr << " during unregister (error " << result
-                         << "); memory may already be freed, using the "
-                         << "provided address";
-        }
-        // Keep the registration alive until the last sub-range releases it
+        // Hold register_mutex_ through the metadata removal, mirroring the
+        // register path, so a concurrent register of another sub-range cannot
+        // republish the buffer in between.
         std::lock_guard<std::mutex> lock(register_mutex_);
-        auto it = registered_base_refs_.find((uint64_t)key_ptr);
-        if (it != registered_base_refs_.end() && --it->second > 0) return 0;
-        registered_base_refs_.erase((uint64_t)key_ptr);
+        auto alias_it = registered_addr_to_base_.find((uint64_t)addr);
+        if (alias_it != registered_addr_to_base_.end()) {
+            uint64_t base = alias_it->second;
+            registered_addr_to_base_.erase(alias_it);
+            // Both maps are updated in pairs under this mutex, so the base
+            // always has a refcount entry here. Only the last alias removes
+            // the shared registration.
+            if (--registered_base_refs_[base] > 0) return 0;
+            registered_base_refs_.erase(base);
+            return metadata_->removeLocalMemoryBuffer((void*)base,
+                                                      update_metadata);
+        }
+        // Not registered here (host memory owned by another transport, or
+        // already unregistered). The caller's address may still equal the
+        // base of a registration whose aliases are live (slab-level cleanup
+        // passing the allocation base); that registration is only removable
+        // through its aliases above.
+        if (registered_base_refs_.count((uint64_t)addr)) {
+            return ERR_ADDRESS_NOT_REGISTERED;
+        }
     }
-    return metadata_->removeLocalMemoryBuffer(key_ptr, update_metadata);
+    return metadata_->removeLocalMemoryBuffer(addr, update_metadata);
 }
 
 int HipTransport::relocateSharedMemoryAddress(uint64_t& dest_addr,
--- a/mooncake-transfer-engine/src/transport/hip_transport/hip_transport.cpp
+++ b/mooncake-transfer-engine/src/transport/hip_transport/hip_transport.cpp
@@ -683,9 +683,28 @@ int HipTransport::registerLocalMemory(void* addr, size_t length,
             return 0;
         }
 
+        // Resolve the true hipMalloc base address. Framework caching
+        // allocators (PyTorch, etc.) sub-allocate tensors within larger
+        // hipMalloc segments. hipIpcGetMemHandle always returns a handle for
+        // the entire segment, so we must register at segment granularity or
+        // relocateSharedMemoryAddress() maps every sub-range back onto the
+        // segment base and the peer silently reads the wrong bytes.
+        void* base_ptr = nullptr;
+        size_t alloc_size = 0;
+        if (!checkHip(hipMemGetAddressRange((hipDeviceptr_t*)&base_ptr,
+                                            &alloc_size, (hipDeviceptr_t)addr),
+                      "HipTransport: hipMemGetAddressRange failed")) {
+            return -1;
+        }
+
+        // Skip if this hipMalloc block is already registered
+        if (registered_base_addrs_.count((uint64_t)base_ptr)) {
+            return 0;
+        }
+
         // Get IPC handle
         hipIpcMemHandle_t handle;
-        if (!checkHip(hipIpcGetMemHandle(&handle, addr),
+        if (!checkHip(hipIpcGetMemHandle(&handle, base_ptr),
                       "HipTransport: hipIpcGetMemHandle failed")) {
             return -1;
         }
@@ -693,14 +712,18 @@ int HipTransport::registerLocalMemory(void* addr, size_t length,
         // Register buffer with metadata
         (void)remote_accessible;
         BufferDesc desc;
-        desc.addr = (uint64_t)addr;
-        desc.length = length;
+        desc.addr = (uint64_t)base_ptr;
+        desc.length = alloc_size;
         desc.name = location;
         desc.shm_name = serializeBinaryData(&handle, sizeof(hipIpcMemHandle_t));
 #ifdef ENABLE_MULTI_PROTOCOL
         desc.protocol = "hip";
 #endif
-        return metadata_->addLocalMemoryBuffer(desc, true);
+        int rc = metadata_->addLocalMemoryBuffer(desc, true);
+        if (rc == 0) {
+            registered_base_addrs_.insert((uint64_t)base_ptr);
+        }
+        return rc;
     }
 
     // Fabric memory registration
@@ -750,7 +773,25 @@ int HipTransport::registerLocalMemory(void* addr, size_t length,
 }
 
 int HipTransport::unregisterLocalMemory(void* addr, bool update_metadata) {
-    return metadata_->removeLocalMemoryBuffer(addr, update_metadata);
+    void* key_ptr = addr;
+    if (!use_fabric_mem_) {
+        void* base_ptr = nullptr;
+        size_t alloc_size = 0;
+        hipError_t result = hipMemGetAddressRange(
+            (hipDeviceptr_t*)&base_ptr, &alloc_size, (hipDeviceptr_t)addr);
+        if (result == hipSuccess) {
+            key_ptr = base_ptr;
+        } else {
+            (void)hipGetLastError();
+            LOG(WARNING) << "HipTransport: hipMemGetAddressRange failed for "
+                         << addr << " during unregister (error " << result
+                         << "); memory may already be freed, using the "
+                         << "provided address";
+        }
+        std::lock_guard<std::mutex> lock(register_mutex_);
+        registered_base_addrs_.erase((uint64_t)key_ptr);
+    }
+    return metadata_->removeLocalMemoryBuffer(key_ptr, update_metadata);
 }
 
 int HipTransport::relocateSharedMemoryAddress(uint64_t& dest_addr,
--- a/mooncake-transfer-engine/src/transport/hip_transport/hip_transport.cpp
+++ b/mooncake-transfer-engine/src/transport/hip_transport/hip_transport.cpp
@@ -652,6 +652,22 @@ Status HipTransport::submitTransferTask(
     return Status::OK();
 }
 
+namespace {
+
+// Matches the descs this transport published into the shared local buffer
+// list. Under multi-protocol the protocol field is authoritative; without it,
+// the serialized IPC/fabric handle is the discriminator (only GPU-IPC
+// transports fill shm_name, and they never coexist with HIP in one build).
+bool isOwnDesc(const TransferMetadata::BufferDesc& desc) {
+#ifdef ENABLE_MULTI_PROTOCOL
+    return desc.protocol == "hip";
+#else
+    return !desc.shm_name.empty();
+#endif
+}
+
+}  // namespace
+
 int HipTransport::registerLocalMemory(void* addr, size_t length,
                                       const std::string& location,
                                       bool remote_accessible,
@@ -680,6 +696,7 @@ int HipTransport::registerLocalMemory(void* addr, size_t length,
                 LOG(INFO) << "HipTransport: skipping non-device memory " << addr
                           << ", leaving it to other transports";
             }
+            skipped_addrs_.insert((uint64_t)addr);
             return 0;
         }
 
@@ -731,6 +748,10 @@ int HipTransport::registerLocalMemory(void* addr, size_t length,
         if (rc == 0) {
             registered_base_refs_[base] = 1;
         } else {
+            // addLocalMemoryBuffer appends the desc before publishing it;
+            // drop the leftover so a failed registration leaves nothing
+            // behind for a later rollback or unregister to miss.
+            metadata_->removeLocalMemoryBuffer((void*)base, isOwnDesc, false);
             registered_addr_to_base_.erase((uint64_t)addr);
         }
         return rc;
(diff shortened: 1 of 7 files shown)


amd-arozanov (approved):
LGTM, thanks! Please fix the comment from copilot

mooncake-transfer-engine/src/transport/hip_transport/hip_transport.cpp
```diff
@@ -750,7 +775,28 @@ int HipTransport::registerLocalMemory(void* addr, size_t length,
...
+                         << "); memory may already be freed, using the "
+                         << "provided address";
+        }
+        // Keep the registration alive until the last sub-range releases it
+        std::lock_guard<std::mutex> lock(register_mutex_);
+        auto it = registered_base_refs_.find((uint64_t)key_ptr);
+        if (it != registered_base_refs_.end() && --it->second > 0) return 0;
+        registered_base_refs_.erase((uint64_t)key_ptr);
+    }
+    return metadata_->removeLocalMemoryBuffer(key_ptr, update_metadata);
```
staryxchen: `registerLocalMemory` holds `register_mutex_` across `addLocalMemoryBuffer`, but this path drops the lock before `removeLocalMemoryBuffer`. After collapsing aliases onto one metadata key (`base`), that window is a real race:
1. refs == 1 (last alias of a caching-allocator slab)
2. Thread A: `--refs` → 0, erase, unlock
3. Thread B: register another sub-range of the same `hipMalloc`; the map is empty, so it publishes a new `BufferDesc` and sets refs = 1
4. Thread A: `removeLocalMemoryBuffer(base)` deletes B's entry
B then reports success while the peer's relocate fails with "address not found". This did not happen before the change, when metadata was keyed by the caller's pointer: A's remove could not delete B's `BufferDesc`.

Please keep `register_mutex_` held through `removeLocalMemoryBuffer`, matching the register path. Lock order stays `register_mutex_` then metadata `segment_lock_`, so it should not invert with register.

This is the same allocator pattern the PR is fixing, just on two threads (one tensor unregistering while another sub-range of the same slab registers).

mooncake-transfer-engine/src/transport/hip_transport/hip_transport.cpp
```diff
@@ -750,7 +775,28 @@ int HipTransport::registerLocalMemory(void* addr, size_t length,
...
+            (void)hipGetLastError();
+            LOG(WARNING) << "HipTransport: hipMemGetAddressRange failed for "
+                         << addr << " during unregister (error " << result
+                         << "); memory may already be freed, using the "
+                         << "provided address";
+        }
+        // Keep the registration alive until the last sub-range releases it
+        std::lock_guard<std::mutex> lock(register_mutex_);
+        auto it = registered_base_refs_.find((uint64_t)key_ptr);
+        if (it != registered_base_refs_.end() && --it->second > 0) return 0;
```
staryxchen: The refcount is the number of `hipMemGetAddressRange` hits on a base, not the set of pointers that actually went through `registerLocalMemory`. `TransferEngineImpl::unregisterLocalMemory` always calls the transport first and only then silently erases its own map, so HIP is the source of truth for the extra decrement.
That makes several pairings corrupt refs:
- Double `unregister(view0)` while view1..3 are still live: `GetAddressRange(view0)` still succeeds, refs is decremented twice, and a later legitimate `unregister(view3)` returns `ERR_ADDRESS_NOT_REGISTERED`. That is the batch-unregister failure from the previous review, just via a different unpaired call.
- `unregister` of a sibling that was never registered (wrong pointer, slab-level cleanup) decrements someone else's ticket and can drop the shared `BufferDesc` while the engine still thinks the original view is registered.
- If `hipMemGetAddressRange` fails because the memory was already freed, `key_ptr` stays the sub-range address. Metadata is now keyed by `base`, so both the map update and `removeLocalMemoryBuffer` miss; the "using the provided address" fallback does not match the new key. Pre-fix, metadata *was* keyed by the caller pointer, so unregister-after-free still worked.

Please key aliases by the registered `addr` (or `addr → base`), and only decrement when that pointer is a known alias. Return `ERR_ADDRESS_NOT_REGISTERED` otherwise. `hipMemGetAddressRange` should identify the slab on register; unregister identity should be the pointer that was registered, so it still works after free.

mooncake-wheel/tests/test_transfer_sub_allocation_on_hip.py
staryxchen, line 1: The two-process spawn test is the right way to observe HIP IPC relocation; a loopback transfer never opens a handle. A few gaps mean the second commit's refcount can regress as a hang instead of a failure, and the batch path from #3546 is never exercised.

- Only the sink unregisters. `_source` holds the four aliases until the process exits, so source-side last-alias teardown (the path that actually removes the `BufferDesc`) is untested.
- Unregister runs in the child before `result.put`. If `unregister_memory` returns non-zero, the parent hits `queue.Empty` after 120s (`no result: source=… sink=…`) instead of seeing the unregister error. Put the payload first, or send unregister status on the same queue.
- Child failures use `assert`. Those become the same timeout, not a unittest error. Prefer putting an error string on `result` (or using `unittest` in the parent after joining).
- The issue is specifically `batch_register_memory` / `batch_unregister_memory` looping the same function. A sequential `register_memory` loop does not catch `unregisterLocalMemoryBatch` returning `ERR_ADDRESS_NOT_REGISTERED` on the 2nd..Nth alias.

Would also help to unregister every sub-range on the source before `done` completes, and to keep the tensors alive across that unregister (so this is not unregister-after-free). Not blocking on the data-path fix, but without this the refcount lifetime is only incidentally covered.

stmatengss:
@Lzy17 This PR is related to AMD. PTAL. Thx.

Lzy17:
Thanks for taking this on, @PACTHEMAN123 

  We built all three revisions and tested on 8× MI355X (gfx950), ROCm 7.2.0, PyTorch
  2.9.1+rocm7.2.0: `test_transfer_sub_allocation_on_hip` fails on base (`[1,1,1,1]`)
  and passes on the head, `test_transfer_on_hip` shows no regression, and the PD
  warmup KV transfer completes under SGLang prefill/decode disaggregation.

  We believe one issue is left, and refcounting can't close it. `TcpTransport` publishes one
  `BufferDesc` per sub-range; `HipTransport` now publishes one per *allocation*. When
  a sub-range starts at the allocation base — the first tensor in a caching-allocator
  segment — both entries share an address, and `removeLocalMemoryBuffer` matches on
  `addr` only with no protocol filter, so the transports delete each other's entries.

  Register `v0..v3` in one segment, then unregister the base-address view twice:

  | call | base | head |
  | --- | --- | --- |
  | `unregister v0` | 0 | 0 |
  | `unregister v0` (repeat) | -3 | -3 |
  | `unregister v1` / `v2` | 0 / 0 | 0 / 0 |
  | `unregister v3` | **0** | **-3** |
  
`v3` was registered and never unregistered. Clean (0/0/0/0) with no repeated call,
  so this needs something to touch the base address twice — slab-level cleanup, or a
  retry. This is the symptom @staryxchen predicted for commit 2, but keying by the
  registered address should have fixed it: tracing 16ebc541 on its own, `v3` should
  return 0. So the remaining cause looks like it's in the metadata layer rather than
  the refcount. `IntraNodeNvlinkTransport` looks the same on the CUDA side;
  

 Everything other than the above LGTM! Happy to share the probe scripts, and we can validate a revision quickly.

alogfans (approved):
LGTM

PACTHEMAN123:
> Thanks for taking this on, @PACTHEMAN123
> 
> We built all three revisions and tested on 8× MI355X (gfx950), ROCm 7.2.0, PyTorch 2.9.1+rocm7.2.0: `test_transfer_sub_allocation_on_hip` fails on base (`[1,1,1,1]`) and passes on the head, `test_transfer_on_hip` shows no regression, and the PD warmup KV transfer completes under SGLang prefill/decode disaggregation.
> 
> We believe one issue is left, and refcounting can't close it. `TcpTransport` publishes one `BufferDesc` per sub-range; `HipTransport` now publishes one per _allocation_. When a sub-range starts at the allocation base — the first tensor in a caching-allocator segment — both entries share an address, and `removeLocalMemoryBuffer` matches on `addr` only with no protocol filter, so the transports delete each other's entries.
> 
> Register `v0..v3` in one segment, then unregister the base-address view twice:
> 
> call	base	head
> `unregister v0`	0	0
> `unregister v0` (repeat)	-3	-3
> `unregister v1` / `v2`	0 / 0	0 / 0
> `unregister v3`	**0**	**-3**
> `v3` was registered and never unregistered. Clean (0/0/0/0) with no repeated call, so this needs something to touch the base address twice — slab-level cleanup, or a retry. This is the symptom @staryxchen predicted for commit 2, but keying by the registered address should have fixed it: tracing [16ebc54](https://github.com/kvcache-ai/Mooncake/commit/16ebc541f9662d299debf83d50f75e3d1f5a0934) on its own, `v3` should return 0. So the remaining cause looks like it's in the metadata layer rather than the refcount. `IntraNodeNvlinkTransport` looks the same on the CUDA side;
> 
> Everything other than the above LGTM! Happy to share the probe scripts, and we can validate a revision quickly.

Thanks for the validation and analysis — I reproduced the same table on MI308X and agree the root cause is that all transports share one buffer list in the local segment, where removeLocalMemoryBuffer matches by addr only, so a same-address entry from another transport can be deleted by mistake.

Two directions to fix:
A. Ownership (general): add an unconditional owner field on `BufferDesc` and match addr && owner on remove which immunizes every transport pair (incl. nvlink/maca), but touches ~12 transports; better as a follow-up PR.
B. Predicate removal (small, this PR): add a removeLocalMemoryBuffer(addr, pred, update_metadata) overload where a transport removes only entries it recognizes as its own; HIP's descs are the only ones with a non-empty shm_name (serialized IPC handle), so that's an exact discriminator; touches only transfer_metadata.{h,cpp} + hip_transport.cpp.
 
I'm inclined to go with B here and A as a follow-up — @Lzy17 does that work for you?

Lzy17:
@PACTHEMAN123 

Yes — B is the right scope for this PR, and A is the better end state.

  4fc9c99f fixes what we reported. On 8× MI355X (gfx950), ROCm 7.2.0, PyTorch
  2.9.1+rocm7.2.0 we measure your sequence as 0 / -3 / 0 / 0 / **0**, and all five
  tests in `test_transfer_on_hip` + `test_transfer_sub_allocation_on_hip` pass.
  Thanks for generalizing it past the sequence we posted — folding the fabric path
  into the same refcount and rejecting a base address whose aliases are still live
  both go beyond what we reported.

  One thing we should have caught in our first comment: B makes ownership opt-in per
  transport, and only TCP and HIP opted in. `RdmaTransport` still calls the match-all
  overload at `rdma_transport.cpp:374`, `:567` and `:619`. `transport_map_` is a
  `std::map`, so `listTransports()` yields hip before rdma — HIP's base-keyed desc is
  appended first, and RDMA's unfiltered removal takes it.

  Same allocation, `v0` at the base, all four registered, then unregistered in
  ascending order — no repeated or unpaired call anywhere:

  | build | transports | v0 | v1 | v2 | v3 | re-register v3 |
  | --- | --- | --- | --- | --- | --- | --- |
  | 16ebc541 | rdma+hip | 0 | 0 | 0 | 0 | 0 |
  | 4fc9c99f | tcp+hip | 0 | 0 | 0 | 0 | 0 |
  | **4fc9c99f** | **rdma+hip** | 0 | 0 | 0 | **-3** | **-7** |

  `unregister(v0)` has HIP only decrement its refcount, so RDMA's match-all removal
  erases HIP's desc instead of its own; the last alias then finds nothing, and
  `TransferEngineImpl::unregisterLocalMemory` returns on `first_error` before
  `eraseMemoryRegionLocked`, so the VA can no longer be re-registered
  (`ERR_ADDRESS_OVERLAPPED`). Unregistering in descending order is clean, which is the
  ordering dependency. Your own new `test_base_address_view_double_unregister` also
  fails on 4fc9c99f in an rdma+hip engine and passes in tcp+hip.

  This matters for us specifically because SGLang's ROCm image builds
  `-DUSE_HIP=ON -DENABLE_MULTI_PROTOCOL=ON` and registers the device KV pool under
  both transports.

  We verified a fix: give RDMA the same predicate, reusing the ownership test its
  submit path already applies at `:1171` (`protocol.empty() || protocol == "rdma"`,
  so legacy untagged descs stay covered). 21 insertions in one file. With it, all
  three orderings are clean in rdma+hip, re-registration succeeds, tcp+hip is
  unchanged, and all five tests pass in both configurations. Happy to open it as a
  commit against your branch, or paste the diff here — whichever you prefer.

PACTHEMAN123:
> @PACTHEMAN123
> 
> Yes — B is the right scope for this PR, and A is the better end state.
> 
> [4fc9c99](https://github.com/kvcache-ai/Mooncake/commit/4fc9c99f28e11054a042a40ae4b71927f7f3e157) fixes what we reported. On 8× MI355X (gfx950), ROCm 7.2.0, PyTorch 2.9.1+rocm7.2.0 we measure your sequence as 0 / -3 / 0 / 0 / **0**, and all five tests in `test_transfer_on_hip` + `test_transfer_sub_allocation_on_hip` pass. Thanks for generalizing it past the sequence we posted — folding the fabric path into the same refcount and rejecting a base address whose aliases are still live both go beyond what we reported.
> 
> One thing we should have caught in our first comment: B makes ownership opt-in per transport, and only TCP and HIP opted in. `RdmaTransport` still calls the match-all overload at `rdma_transport.cpp:374`, `:567` and `:619`. `transport_map_` is a `std::map`, so `listTransports()` yields hip before rdma — HIP's base-keyed desc is appended first, and RDMA's unfiltered removal takes it.
> 
> Same allocation, `v0` at the base, all four registered, then unregistered in ascending order — no repeated or unpaired call anywhere:
> 
> build	transports	v0	v1	v2	v3	re-register v3
> [16ebc54](https://github.com/kvcache-ai/Mooncake/commit/16ebc541f9662d299debf83d50f75e3d1f5a0934)	rdma+hip	0	0	0	0	0
> [4fc9c99](https://github.com/kvcache-ai/Mooncake/commit/4fc9c99f28e11054a042a40ae4b71927f7f3e157)	tcp+hip	0	0	0	0	0
> **[4fc9c99](https://github.com/kvcache-ai/Mooncake/commit/4fc9c99f28e11054a042a40ae4b71927f7f3e157)**	**rdma+hip**	0	0	0	**-3**	**-7**
> `unregister(v0)` has HIP only decrement its refcount, so RDMA's match-all removal erases HIP's desc instead of its own; the last alias then finds nothing, and `TransferEngineImpl::unregisterLocalMemory` returns on `first_error` before `eraseMemoryRegionLocked`, so the VA can no longer be re-registered (`ERR_ADDRESS_OVERLAPPED`). Unregistering in descending order is clean, which is the ordering dependency. Your own new `test_base_address_view_double_unregister` also fails on [4fc9c99](https://github.com/kvcache-ai/Mooncake/commit/4fc9c99f28e11054a042a40ae4b71927f7f3e157) in an rdma+hip engine and passes in tcp+hip.
> 
> This matters for us specifically because SGLang's ROCm image builds `-DUSE_HIP=ON -DENABLE_MULTI_PROTOCOL=ON` and registers the device KV pool under both transports.
> 
> We verified a fix: give RDMA the same predicate, reusing the ownership test its submit path already applies at `:1171` (`protocol.empty() || protocol == "rdma"`, so legacy untagged descs stay covered). 21 insertions in one file. With it, all three orderings are clean in rdma+hip, re-registration succeeds, tcp+hip is unchanged, and all five tests pass in both configurations. Happy to open it as a commit against your branch, or paste the diff here — whichever you prefer.

@Lzy17  Thanks for the follow-up and for validating the fix on MI355X — glad the fabric refcount and the live-base guard held up beyond the original report. The rdma+hip gap is real, but I don't have an RDMA testbed to reproduce or validate it (my MI308X box has no RDMA NICs), so please go ahead and open the commit against this branch, thx :)

Lzy17:
Opened PACTHEMAN123/Mooncake#1 against your branch with the RDMA opt-in — three call sites switched to the ownership predicate, reusing the same test your submit path already applies. Validated on 8× MI355X: rdma+hip is clean in all orderings and re-registration succeeds, tcp+hip is unchanged, and all five HIP tests pass in both configs (details in the PR). `maintainer_can_modify` is on, so feel free to tweak. Thanks!

PACTHEMAN123:
> Opened [PACTHEMAN123#1](https://github.com/PACTHEMAN123/Mooncake/pull/1) against your branch with the RDMA opt-in — three call sites switched to the ownership predicate, reusing the same test your submit path already applies. Validated on 8× MI355X: rdma+hip is clean in all orderings and re-registration succeeds, tcp+hip is unchanged, and all five HIP tests pass in both configs (details in the PR). `maintainer_can_modify` is on, so feel free to tweak. Thanks!

@Lzy17  Merged your commit into the branch — it compiles clean on my side and the tcp+hip tests plus the original probe sequence are unchanged. Thx！

Lzy17 (approved):
LGTM! Please make sure CI is clean. Thanks!

PACTHEMAN123:
The failed `master_service_ssd_test_for_snapshot` has no overlap with this PR's diff (transfer engine only), so this looks flaky or environment-dependent rather than related to these changes. Could someone with access re-run the failed jobs?

catyans (approved):
LGTM. The current head registers HIP IPC at the allocation base, reference-counts sub-range aliases, and scopes HIP/TCP/RDMA metadata removal to each transport's own descriptor. The cross-process payload and alias-lifecycle coverage, MI355X validation, and full CI all pass.

alogfans:
cc @ShangmingCai @stmatengss

PACTHEMAN123:
Hi @alogfans @ShangmingCai @stmatengss, just following up on this PR. The current head is mergeable, all CI checks are green, and it has received approvals. Are there any remaining concerns, or could a maintainer help merge it? I can update the branch against the latest `main` if required. Thanks!

PR #3552, opened 2026-08-20
[Store] Support graceful segment unmount on standalone client shutdown

### LINKED ISSUE #3551: [Bug]: standalone `mooncake_client` does not gracefully drain mounted segments on `SIGTERM`
### Bug Report

# Bug Report

## Description

We observed two related shutdown problems with standalone `mooncake_client`:

1. When the shutdown cleanup runs, `Client::~Client()` immediately calls `UnmountSegment()` for mounted segments. There is no configurable serving grace period before the segments are removed.

   Under high concurrency, peer clients may already have obtained segment or replica information. Removing the segment immediately can cause in-flight or subsequent `get`/`put` operations using that information to fail.

2. In some processes, `kill -TERM <pid>` does not enter the cleanup path at all. The process exits immediately and becomes a zombie, without logging:

   ```text
   Received signal 15, cleaning up resources
   ```

   The standalone signal path uses `pthread_sigmask()` plus a dedicated `sigwait()` thread. Signal masks are per-thread, and we observed several threads with `SIGTERM` unblocked:

   ```text
   main thread:  SigBlk=0000000000004203
   other thread: SigBlk=0000000000000000
   ```

   If process-directed `SIGTERM` is delivered to an unblocked thread, the default action can terminate the process before the `sigwait()` thread performs cleanup.

The Python console-script launcher is not the cause: it uses `execv()` to replace itself with the native `mooncake_client` ELF, so the resulting process is standalone.

## Steps to reproduce

1. Start standalone `mooncake_client` and mount at least one segment.
2. Send `SIGTERM` to the exact live PID:

   ```bash
   kill -TERM <pid>
   ```

3. Observe the process state and client logs. Other clients may output some error about "Put/Load"

## Actual behavior

Depending on signal delivery:

- cleanup runs but mounted segments are unmounted immediately, without a serving grace period; or
- cleanup is bypassed and the process terminates immediately.

Both behaviors can cause transient failures on peer clients that are still using previously obtained segment information.

## Expected behavior

On `SIGTERM`, standalone `mooncake_client` should reliably enter cleanup, stop accepting new allocations for its segments, keep existing data readable for a configurable grace period, and only then unmount the segments and exit.

## Proposal

- Add a configurable graceful-unmount period for standalone client shutdown.
- During shutdown, request graceful unmount for every mounted segment and keep the process and registered memory alive until the grace period completes or a bounded timeout is reached.
- Make signal handling robust when some process threads do not inherit the blocked signal mask, while keeping cleanup work outside the async signal-handler context.
- Preserve the current immediate-unmount behavior when the grace period is disabled.

### Before submitting...

justadogistaken:
Standalone mooncake_client unmounted its segments immediately on exit, so peers holding segment info could fail subsequent get/put. Add an opt-in grace period via MC_GRACEFUL_UNMOUNT_SECONDS: on teardown, request graceful unmount for all mounted segments and keep the process and its memory registrations alive until they are released or a bounded timeout expires.

Also install a process-wide sigaction handler for SIGINT/SIGTERM/SIGHUP that forwards to the existing sigwait thread. pthread_sigmask only affects the calling thread, so a termination signal delivered to a thread that never inherited the block mask took the default action and killed the process before any cleanup ran.

MC_GRACEFUL_UNMOUNT_SECONDS=0 (default) keeps the previous behavior.

Try to fix #3551 

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
export MC_GRACEFUL_UNMOUNT_SECONDS=30
start mooncake_client 
pkill -TERM -f mooncake_client
```

1. See if there still put/load errors exist on others.
2. We can see the graceful shutdown log.

```
grep -E "Graceful unmount|Received signal" mooncake_client.log | tail -20
I0820 16:35:56.381542 82704 real_client.cpp:581] Received signal 15, cleaning up resources
I0820 16:35:56.384151 82704 client_service.cpp:2908] Graceful unmount requested for 1/1 segments, grace_period_ms=30000, now waiting for completion (max 75s)
I0820 16:36:37.605084 82704 client_service.cpp:2921] Graceful unmount of all segments completed
I0820 16:41:47.074059 96751 real_client.cpp:581] Received signal 15, cleaning up resources
I0820 16:41:47.076062 96751 client_service.cpp:2908] Graceful unmount requested for 1/1 segments, grace_period_ms=30000, now waiting for completion (max 75s)
I0820 16:42:29.476811 96751 client_service.cpp:2921] Graceful unmount of all segments completed
```

**Test results:**
- [ ] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

### DIFF
--- a/mooncake-integration/store/store_py.cpp
+++ b/mooncake-integration/store/store_py.cpp
@@ -2,6 +2,7 @@
 #include <pybind11/stl.h>
 #include <numa.h>
 
+#include <chrono>
 #include <functional>
 #include <limits>
 #include <numeric>
@@ -2379,13 +2380,42 @@ PYBIND11_MODULE(store, m) {
             py::arg("keys"),
             "Check if multiple objects exist. Returns list of results: 1 if "
             "exists, 0 if not exists, -1 if error")
-        .def("close",
-             [](MooncakeStorePyWrapper &self) {
-                 if (!self.store_) return 0;
-                 int rc = self.store_->tearDownAll();
-                 self.store_.reset();
-                 return rc;
-             })
+        .def(
+            "close",
+            [](MooncakeStorePyWrapper &self, uint64_t grace_period_seconds) {
+                if (!self.store_) return 0;
+                CloseOptions options;
+                options.grace_period =
+                    std::chrono::seconds(grace_period_seconds);
+                // timeout is left at zero on purpose: the upper bound depends
+                // on the client's own confirmation schedule, so the client is
+                // the only party able to derive a value that can actually
+                // observe completion.
+                // Only RealClient owns segments, so it is the only one with a
+                // graceful phase; other clients tear down immediately.
+                auto real_client =
+                    std::dynamic_pointer_cast<RealClient>(self.store_);
+                int rc = 0;
+                {
+                    // Close() may wait for the grace period to elapse, so the
+                    // interpreter must not be blocked while it runs.
+                    py::gil_scoped_release release;
+                    if (real_client) {
+                        rc = static_cast<int>(
+                            to_py_ret(real_client->Close(options)));
+                    } else {
+                        rc = self.store_->tearDownAll();
+                    }
+                }
+                self.store_.reset();
+                return rc;
+            },
+            py::arg("grace_period_seconds") = 0,
+            "Close the store. With grace_period_seconds > 0 the master is "
+            "asked to stop allocating on this client's segments while they "
+            "stay readable, and the call waits for them to be released, so "
+            "peers holding segment information can finish their reads. Zero "
+            "(the default) tears down immediately.")
         .def("health_check", &MooncakeStorePyWrapper::health_check,
              "Health check for store connectivity. "
              "Returns 0 if healthy, 1 if not initialized/closed, "
--- a/mooncake-store/src/client_service.cpp
+++ b/mooncake-store/src/client_service.cpp
@@ -517,6 +517,7 @@ Client::~Client() {
         std::lock_guard<std::mutex> lock(mounted_segments_mutex_);
         mounted_segments_.clear();
         gracefully_unmounting_segments_.clear();
+        gracefully_unmounting_done_cv_.notify_all();
     }
 
     for (auto& entry : graceful_cleanup_callbacks) {
@@ -3587,6 +3588,57 @@ tl::expected<void, ErrorCode> Client::UnmountSegmentById(
     return {};
 }
 
+size_t Client::GracefulUnmountAll(uint64_t grace_period_ms) {
+    // Snapshot the ids first: UnmountSegmentById takes mounted_segments_mutex_
+    // itself, so we must not hold it across the calls.
+    std::vector<UUID> segment_ids;
+    {
+        std::lock_guard<std::mutex> lock(mounted_segments_mutex_);
+        segment_ids.reserve(mounted_segments_.size());
+        for (const auto& [id, segment] : mounted_segments_) {
+            segment_ids.push_back(id);
+        }
+    }
+
+    size_t graceful_ok = 0;
+    for (const auto& id : segment_ids) {
+        auto result = UnmountSegmentById(id, grace_period_ms);
+        if (result.has_value()) {
+            graceful_ok++;
+        } else {
+            // Leave the failed segment mounted; ~Client() will unmount it
+            // immediately as the fallback path.
+            LOG(WARNING) << "Graceful unmount failed for segment "
+                         << UuidToString(id)
+                         << ", error=" << toString(result.error())
+                         << "; it will fall back to immediate unmount";
+        }
+    }
+    if (!segment_ids.empty()) {
+        LOG(INFO) << "Graceful unmount requested for " << graceful_ok << "/"
+                  << segment_ids.size()
+                  << " segments, grace_period_ms=" << grace_period_ms;
+    }
+    return graceful_ok;
+}
+
+bool Client::WaitForGracefulUnmountAll(
+    std::chrono::steady_clock::time_point deadline) {
+    std::unique_lock<std::mutex> lock(mounted_segments_mutex_);
+    const bool drained = gracefully_unmounting_done_cv_.wait_until(
+        lock, deadline,
+        [this]() { return gracefully_unmounting_segments_.empty(); });
+    if (drained) {
+        LOG(INFO) << "Graceful unmount of all segments completed";
+    } else {
+        LOG(WARNING) << "Graceful unmount deadline reached with "
+                     << gracefully_unmounting_segments_.size()
+                     << " segments still pending; leftovers are handled by "
+                        "~Client() and master-side client expiry";
+    }
+    return drained;
+}
+
 bool Client::WaitForGracefulUnmountDelay(std::chrono::milliseconds delay) {
     std::unique_lock<std::mutex> lock(graceful_unmount_timer_mutex_);
     return graceful_unmount_timer_cv_.wait_for(
@@ -3643,6 +3695,7 @@ void Client::OnGracefulUnmountTimer(const UUID& segment_id, int retry_left) {
                         << rc;
                 }
                 gracefully_unmounting_segments_.erase(it);
+                gracefully_unmounting_done_cv_.notify_all();
             }
             auto callback_it =
                 graceful_unmount_cleanup_callbacks_.find(segment_id);
--- a/mooncake-store/src/client_service.cpp
+++ b/mooncake-store/src/client_service.cpp
@@ -417,6 +417,9 @@ Client::Client(const std::string& local_hostname,
       metadata_connstring_(metadata_connstring),
       protocol_(protocol),
       object_checksum_enabled_(Environ::Get().GetStoreChecksumEnabled()),
+      graceful_unmount_scheduler_([this](const GracefulUnmountCheck& check) {
+          this->OnGracefulUnmountTimer(check);
+      }),
       pinned_buffer_pool_(std::make_unique<PinnedBufferPool>()),
       write_thread_pool_(2),
       task_thread_pool_(4) {
@@ -459,11 +462,10 @@ Client::~Client() {
         leader_monitor_thread_.join();
     }
 
-    {
-        std::lock_guard<std::mutex> lock(graceful_unmount_timer_mutex_);
-        graceful_unmount_timer_stopping_ = true;
-    }
-    graceful_unmount_timer_cv_.notify_all();
+    // Join the confirmation timer thread before tearing down segment state:
+    // its callback touches mounted_segments_ and the transfer engine, and the
+    // teardown below must be the single owner of whatever is left pending.
+    graceful_unmount_scheduler_.Stop();
 
     // Stop queued timer/task callbacks before tearing down segment state.
     task_running_ = false;
@@ -3639,25 +3641,20 @@ bool Client::WaitForGracefulUnmountAll(
     return drained;
 }
 
-bool Client::WaitForGracefulUnmountDelay(std::chrono::milliseconds delay) {
-    std::unique_lock<std::mutex> lock(graceful_unmount_timer_mutex_);
-    return graceful_unmount_timer_cv_.wait_for(
-        lock, delay, [this]() { return graceful_unmount_timer_stopping_; });
-}
-
 void Client::StartGracefulUnmountTimer(const UUID& segment_id,
                                        uint64_t grace_period_ms) {
-    auto delay =
-        std::chrono::milliseconds(grace_period_ms) + std::chrono::seconds(10);
-    task_thread_pool_.enqueue([this, segment_id, delay]() {
-        if (this->WaitForGracefulUnmountDelay(delay)) {
-            return;
-        }
-        this->OnGracefulUnmountTimer(segment_id, /*retry_left=*/3);
-    });
+    // One scheduler entry rather than one blocked worker, so N segments get N
+    // concurrent deadlines instead of being serialized by the pool width.
+    graceful_unmount_scheduler_.Schedule(
+        GracefulUnmountCheck{segment_id, kGracefulUnmountConfirmRetries},
+        std::chrono::steady_clock::now() +
+            std::chrono::milliseconds(grace_period_ms) +
+            kGracefulUnmountConfirmInterval);
 }
 
-void Client::OnGracefulUnmountTimer(const UUID& segment_id, int retry_left) {
+void Client::OnGracefulUnmountTimer(const GracefulUnmountCheck& check) {
+    const UUID& segment_id = check.segment_id;
+
     // Query master to confirm the segment has been removed
     {
         std::lock_guard<std::mutex> lock(mounted_segments_mutex_);
@@ -3710,18 +3707,10 @@ void Client::OnGracefulUnmountTimer(const UUID& segment_id, int retry_left) {
         return;
     }
 
-    if (retry_left > 0) {
-        try {
-            task_thread_pool_.enqueue([this, segment_id, retry_left]() {
-                if (this->WaitForGracefulUnmountDelay(
-                        std::chrono::seconds(10))) {
-                    return;
-                }
-                this->OnGracefulUnmountTimer(segment_id, retry_left - 1);
-            });
-        } catch (const std::runtime_error& e) {
-            VLOG(1) << "Skip graceful unmount retry enqueue: " << e.what();
-        }
+    if (check.retry_left > 0) {
+        graceful_unmount_scheduler_.Schedule(
+            GracefulUnmountCheck{segment_id, check.retry_left - 1},
+            std::chrono::steady_clock::now() + kGracefulUnmountConfirmInterval);
     } else {
         LOG(WARNING) << "Graceful unmount cleanup timeout for segment "
                      << UuidToString(segment_id);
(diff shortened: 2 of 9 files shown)


mooncake-store/src/client_service.cpp
```diff
@@ -3587,6 +3587,63 @@ tl::expected<void, ErrorCode> Client::UnmountSegmentById(
     return {};
 }
 
+void Client::GracefulUnmountAllAndWait(uint64_t grace_period_ms,
```
Aionw: This API combines initiating graceful unmounts with the teardown wait policy, and the wait itself polls once per second. Please keep the client-side mechanism event-driven: notify a condition variable when `OnGracefulUnmountTimer` erases a segment, or use the existing cleanup callbacks, then wait until the caller-provided deadline. The teardown caller should own the timeout policy. Polling here adds up to one second of avoidable shutdown latency and makes completion harder to test deterministically.

mooncake-store/src/real_client.cpp
```diff
@@ -587,6 +589,43 @@ ResourceTracker &ResourceTracker::getInstance() {
...
+//   thread blocks these signals, but pthread_sigmask only affects the calling
+//   thread; threads created before ResourceTracker, or threads whose mask is
+//   reset by a third-party runtime, may still have them unblocked. If a
+//   termination signal lands on such a thread while no handler is installed,
+//   the default action terminates the process immediately and the sigwait
+//   thread never wakes up -> no cleanup, no graceful unmount, no log.
+//   A sigaction handler is process-scoped (covers ALL threads regardless of
+//   their mask), so it catches the signal wherever it lands and hands control
+//   to the sigwait thread, which runs the real (non-async-signal-safe) cleanup.
+extern "C" void forward_term_signal_to_sigwait_thread(int sig) {
```
Aionw: The project already builds and links standalone Asio. Can this use `asio::signal_set` instead of maintaining a custom `sigaction -> pthread_kill -> sigwait` bridge? If Asio is unsuitable because of the existing signal-mask model, please document that constraint and cover it with tests. The current handler also loads `std::atomic<pthread_t>` without guaranteeing that the atomic is lock-free in signal context, and a single shared signal slot makes multiple deliveries difficult to reason about safely.

mooncake-store/src/real_client.cpp
```diff
@@ -711,6 +763,26 @@ void ResourceTracker::startSignalThread() {
...
         ready_future.wait();  // Ensure thread is ready
+
+        // In addition to the sigwait thread, install a process-wide sigaction
+        // handler for the termination signals. sigaction is process-scoped and
+        // therefore effective on EVERY thread irrespective of its per-thread
+        // signal mask -- this closes the gap where a process-directed signal is
+        // delivered to a thread that never inherited our block mask, which
+        // would otherwise trigger the default action and kill the process
+        // before the sigwait thread ever runs.
+        struct sigaction fa{};
```
Aionw: This is not protected from concurrent handler execution as the comment claims: `sa_mask` augments the mask of the thread currently running the handler, so another unblocked thread can still enter the handler and overwrite `g_forwarded_signal`. The `sigaction` return values are also ignored, and these registrations replace any existing process handlers without restoring them. Since this feature is specifically for the standalone client, process signal ownership should live in `real_client_main.cpp`, not in the shared `ResourceTracker` used by library consumers.

mooncake-store/src/real_client.cpp
```diff
@@ -1354,6 +1426,34 @@ tl::expected<void, ErrorCode> RealClient::tearDownAll_internal() {
         return {};
     }
 
+    // Graceful unmount before tearing anything down: the Client object (and
+    // its task_thread_pool_, which drives the graceful-unmount timers) must
+    // stay alive while we wait, so this has to run before client_.reset().
+    // MC_GRACEFUL_UNMOUNT_SECONDS=0 (default) keeps the legacy behavior of
+    // immediate unmount in ~Client().
+    {
+        const char *grace_env = std::getenv("MC_GRACEFUL_UNMOUNT_SECONDS");
```
Aionw: This changes a much wider surface than standalone signal shutdown. `tearDownAll_internal()` is also reached from the pybind `close()`, `RealClient` destruction, the C API, and `atexit`. When this environment variable is present, the Python binding currently holds the GIL while this code may block for `grace + 45s`, potentially freezing the interpreter during `close()` or object destruction. Please parse and validate this setting once in `real_client_main.cpp` and pass an explicit graceful-shutdown option only from the standalone signal path; ordinary `tearDownAll()` should keep deterministic immediate semantics. If Python later exposes graceful close, it should be explicit and release the GIL while waiting.

Aionw:
Thanks for submitting the PR. Overall, it's fine for the client to support graceful shutdown on the C++ side, but we also need to consider the impact on the Python wrapper, because graceful shutdown is currently also implemented in the Python wrapper.

justadogistaken:
@Aionw 

Thanks for the detailed review.

Signal handling. Dropped the custom sigaction → pthread_kill → sigwait bridge (you were right that sa_mask doesn't prevent concurrent handler entry, and the shared ResourceTracker shouldn't claim process-wide signal state). Now using asio::signal_set in real_client_main.cpp.

Now a standalone-only gflag (--graceful_unmount_seconds) parsed once in real_client_main.cpp, not an environment variable. So the pybind close(), the C API, destruction and atexit are unaffected and never block while holding the GIL.

Python wrapper. The wrapper exposes graceful unmount as an explicitly orchestrated HTTP operation that forwards grace_period_seconds into this same C++ mechanism, and its own shutdown path (shutdown_event → service.stop() → close()) still unmounts immediately. This change targets the complementary case: a standalone client terminated by a signal with no orchestrator involved. The Python surface is unchanged.

Tests. Added integration coverage for event-driven completion and for the caller-supplied deadline.

mooncake-store/src/real_client_main.cpp
```diff
@@ -103,6 +120,34 @@ void RegisterClientRpcService(coro_rpc::coro_rpc_server &server,
...
 
+namespace {
+
+// Runs the graceful-unmount sequence for the standalone client.
+//
+// The standalone binary owns the termination-signal policy: it decides the
+// grace period and the deadline, while Client only exposes the mechanism
+// (initiate + await). RealClient::tearDownAll() keeps its immediate,
+// deterministic semantics and is left to the regular teardown below.
+void GracefulUnmountBeforeExit(RealClient &client) {
```
Aionw: ### Suggested direction

`real_client_main.cpp` should not orchestrate shutdown by reaching through `RealClient::client_`. Please put the complete client shutdown lifecycle behind one explicit API, for example:

```cpp
struct CloseOptions {
    std::chrono::milliseconds grace_period{0};
    std::chrono::milliseconds timeout{0};
};

tl::expected<void, ErrorCode> RealClient::Close(
    const CloseOptions &options);
```

#### Responsibilities

- **Standalone entry point:** owns `asio::signal_set`, parses gflags, stops new ingress as appropriate, and calls `Close(options)`.
- **Python wrapper:** owns its existing asyncio signal handling, forwards the same options through pybind, and releases the GIL while `Close()` waits.
- **`RealClient::Close()`:** atomically enters the closing state, requests graceful unmount, waits for **all** pending segments (both pre-existing and newly requested), applies the timeout fallback, and performs teardown exactly once.
- **C API / destructor:** calls the same API with immediate/default options.
- **`ResourceTracker`:** remains a resource registry and `atexit` backstop; it should not own process signals.

```mermaid
flowchart LR
    subgraph Frontends["Process-facing frontends"]
        S["Standalone main<br/>asio::signal_set<br/>gflags + stop ingress"]
        P["Python wrapper<br/>asyncio signal handler<br/>release GIL while waiting"]
        C["C API / destructor<br/>immediate defaults"]
        RT["ResourceTracker<br/>registry + atexit only<br/>no sigwait"]
    end

    S -->|"CloseOptions"| R["RealClient::Close(options)"]
    P -->|"CloseOptions"| R
    C -->|"CloseOptions{}"| R
    RT -->|"CloseOptions{}"| R

    R --> A["Atomically enter Closing"]
    A --> U["Request graceful unmount<br/>for mounted segments"]
    U --> W["Wait for every pending segment<br/>pre-existing + newly requested"]
    W --> D{"Completed before timeout?"}
    D -->|"Yes"| T["Teardown exactly once"]
    D -->|"No"| F["Timeout fallback"]
    F --> T
```

#### Signal ownership constraint

When standalone Asio owns `SIGINT` / `SIGTERM` / `SIGHUP`, the existing `ResourceTracker::sigwait` consumer for those signals must be disabled or removed. POSIX leaves concurrent `sigaction` / `sigwait` use for the same signal unspecified, so the current claim that Asio always wins is not valid.

This structure gives every frontend one shutdown implementation without letting the shared library claim process-wide signal state.

justadogistaken:
@Aionw Thanks for the detailed direction. I've done modifications based on your suggestions.
RealClient::Close(const CloseOptions &) now matches your suggested signature(milliseconds fields, tl::expected return) and is the single shutdown implementation. The PyClient::Close virtual was removed: only RealClient owns segments, so it is the only one with a graceful phase; other clients keep using tearDownAll(), which is exactly Close(CloseOptions{}). The standalone entry now only owns asio::signal_set + gflags; GracefulUnmountBeforeExit and the client_ reach-through are gone.

Frontends always leave timeout at zero and let the client derive it — as you noted, the bound depends on the client's own confirmation schedule, so a caller-supplied value can't reliably observe completion. Inside Client, initiate/await stay event-driven (CV notified in OnGracefulUnmountTimer, no polling), and ResourceTracker::DisableSignalHandling() runs before the sigwait thread can start, so asio is the single signal owner.

mooncake-store/src/real_client_main.cpp
```diff
@@ -144,8 +163,39 @@ int main(int argc, char *argv[]) {
...
+    asio::signal_set signals(signal_io, SIGINT, SIGTERM, SIGHUP);
+    signals.async_wait([&](const asio::error_code &ec, int sig) {
+        if (ec) {
+            return;
+        }
+        LOG(INFO) << "Received signal " << sig << ", shutting down";
+        CloseOptions options;
+        options.grace_period =
+            std::chrono::seconds(FLAGS_graceful_unmount_seconds);
+        auto closed = client_inst->Close(options);
```
Aionw, line 197: [P1] Stop and drain ingress before entering the graceful wait

Close() may wait for the grace period plus the confirmation budget, but server.stop() is only called after Close() returns. During that window the RPC server can still accept work; a newly mounted segment is outside the GracefulUnmountAll() snapshot, and an in-flight handler can race the final client_.reset(). Please stop accepting new RPCs and coordinate/drain in-flight handlers before starting Close(), and make the Closing state reject new operations.

mooncake-store/src/real_client.cpp
```diff
@@ -1347,13 +1368,36 @@ int RealClient::initAll(const std::string &protocol_,
...
+    // registrations that keep remote reads working, so it has to stay alive for
+    // the whole wait, i.e. this must run before client_.reset() below.
+    if (options.grace_period > std::chrono::milliseconds::zero() && client_) {
+        client_->GracefulUnmountAll(options.grace_period.count());
+        // Waits for the entire pending set, so segments that a caller had
+        // already started unmounting with a grace period are covered as well.
+        // On expiry we simply continue: ~Client() releases whatever is left and
+        // the master still applies its own scheduler and client expiry.
+        const auto timeout =
+            options.timeout > std::chrono::milliseconds::zero()
```
Aionw, line 1411: [P1] Make the derived timeout valid for bulk close

The underlying timer-pool design predates this PR, so replacing the whole scheduler can be handled separately. However, this new timeout assumes all per-segment timers progress concurrently. task_thread_pool_ has four workers, and every timer occupies one worker while sleeping for grace_period + 10s, so segments after the first four start in later batches and can miss grace_period + 45s even when the master behaves normally. Please make the bulk-close path correct for more than four segments, for example with one aggregate timer/query task or another scoped solution, and add a multi-segment test.

mooncake-integration/store/store_py.cpp
```diff
@@ -2379,13 +2380,42 @@ PYBIND11_MODULE(store, m) {
...
+                        rc = static_cast<int>(
+                            to_py_ret(real_client->Close(options)));
+                    } else {
+                        rc = self.store_->tearDownAll();
+                    }
+                }
+                self.store_.reset();
+                return rc;
+            },
+            py::arg("grace_period_seconds") = 0,
```
Aionw, line 2413: [P2] Wire the grace option into the Python signal shutdown path

The pybind API now accepts close(grace_period_seconds) and correctly releases the GIL, but the existing asyncio signal path ends in MooncakeStoreService.stop(), which still calls close() without an argument. Therefore Python signal shutdown always selects immediate teardown. Please expose the grace setting through the Python service configuration or CLI and forward it from stop(), so the process-facing Python wrapper owns the signal policy while reusing RealClient::Close().

mooncake-store/tests/client_integration_test.cpp
```diff
@@ -2047,6 +2048,86 @@ TEST_F(ClientIntegrationTest, MountSegmentAndGetIdAndUnmountSegmentById) {
...
+    const auto start = std::chrono::steady_clock::now();
+    EXPECT_FALSE(client->WaitForGracefulUnmountAll(start))
+        << "an expired deadline must not report completion";
+    EXPECT_LT(std::chrono::steady_clock::now() - start,
+              std::chrono::seconds(5));
+
+    // No mounted segments remain, so this is a no-op that must not block.
+    EXPECT_EQ(client->GracefulUnmountAll(grace_period_ms), 0u);
+
+    free(test_buffer);
```
Aionw, line 2172: [P2] Unregister the MR before freeing its backing memory

This test deliberately expires the wait while the segment is still in gracefully_unmounting_segments_, so its transfer-engine memory registration is still active here. Freeing test_buffer before destroying the client leaves that registration pointing at freed storage, and the destructor only unregisters it later. Please reset or destroy client first, then free the buffer, or otherwise complete the pending unmount before releasing the backing allocation.

mooncake-store/src/real_client.cpp
```diff
@@ -1347,13 +1368,36 @@ int RealClient::initAll(const std::string &protocol_,
...
     if (!closed_.compare_exchange_strong(expected, true,
                                          std::memory_order_acq_rel)) {
         return {};
     }
 
+    // Graceful phase, before anything is torn down. The Client object owns the
+    // thread pool that drives the graceful-unmount timers and holds the memory
+    // registrations that keep remote reads working, so it has to stay alive for
+    // the whole wait, i.e. this must run before client_.reset() below.
+    if (options.grace_period > std::chrono::milliseconds::zero() && client_) {
```
Aionw, line 1404: [P2] Reject negative durations instead of treating them as defaults

CloseOptions uses signed chrono durations. These greater-than-zero checks silently reinterpret a negative grace_period as immediate shutdown and a negative timeout as the auto-derived timeout. The Python and gflag frontends are unsigned, but direct C++ callers can construct either value. If this public API validates its options, please reject negative fields with INVALID_PARAMS before setting closed_; reserve exactly zero for the documented immediate or derived-default semantics.

justadogistaken:
@Aionw Thanks for the detailed review. All five comments are addressed now. Details and the two places where I deviated from the literal suggestion are below.

**[P1] Stop and drain ingress before entering the graceful wait**
[P1] Stop and drain ingress before entering the graceful wait

server.stop() now runs before Close(). 

Two follow-on changes in the same entry point:

Removed signals.cancel(). Since the handler now calls server.stop() first, start() returns while Close() is still running, so the old cancel() ran concurrently with the handler — and asio's signal_set is explicitly not safe for concurrent access. It was also redundant: ~signal_set cancels any still-pending wait. signal_io.stop() stays (safe from any thread, doesn't interrupt a running handler), and the join() is what actually waits for Close().
Added IsClosing(), gating allocateAndMountSegment, map_shm_internal_with_device, ascend_shm_internal and ascend_ipc_shm_internal. server.stop() alone isn't enough: the IPC channel is only torn down by stop_ipc_server() inside Close(), after the wait, so it stays reachable for the whole grace period and could still mount a segment past the GracefulUnmountAll() snapshot.

**[P1] Make the derived timeout valid for bulk close**
Now on DeadlineScheduler — the primitive the master already uses for its side of this same unmount (graceful_unmount_scheduler_). One timer thread with a deadline heap, so every segment's schedule progresses concurrently and the derived bound holds for any number of segments.

Added GracefulUnmountAllHandlesMoreSegmentsThanWorkers: six segments, i.e. more than the pool width, asserting completion within that bound.

**[P2] Reject negative durations instead of treating them as defaults**

Validated before the closed_ CAS, so a rejected call leaves the client usable — covered by a test that runs a valid Close() afterwards. Zero keeps the immediate/derived semantics; CloseOptions documents the requirement.

**[P2] Wire the grace option into the Python signal shutdown path**

MooncakeStoreService takes graceful_unmount_seconds (rejects negatives), exposes --graceful-unmount-seconds, and stop() forwards it to close().

**[P2] Unregister the MR before freeing its backing memory**

client.reset() before free() in both tests, with a comment on the deadline one explaining why the order matters there.

justadogistaken:
@Aionw Help take a look.

PR #3554, opened 2026-08-20
[TransferEngine] Identify TCP peers by incarnation ID

jacklin78911-collab:
## Description

This PR adds a TCP peer incarnation ID to distinguish different TCP server lifetimes that reuse the same host and port.

Previously, TCP connection groups were identified only by host and port. If a peer process restarted while keeping the same endpoint, the transport could not distinguish the old process instance from the new one and might reuse stale connection groups.

This change:
- adds an optional tcp_instance_id to SegmentDesc;
- generates one stable incarnation ID per TcpTransport lifetime;
- includes the ID in TCP ConnectionKey;
- reuses the existing endpoint refresh and connection-group retirement logic.

Legacy descriptors without tcp_instance_id remain compatible.

## Testing

Validated with:

- Same host/port peer restart regression test;
- TCP incarnation lifecycle tests;
- Endpoint refresh and retirement tests;
- Full tcp_write_visibility_test;
- ASan/LSan tcp_write_visibility_test.

Results:

- tcp_write_visibility_test: 39/39 passed
- ASan/LSan: 39/39 passed
- clang-format --dry-run --Werror: passed
- git diff --check: passed

### DIFF
--- a/mooncake-transfer-engine/src/transfer_metadata.cpp
+++ b/mooncake-transfer-engine/src/transfer_metadata.cpp
@@ -380,6 +380,9 @@ static int encodeMultiProtocolSegmentDesc(
     }
     segmentJSON["tcp_data_port"] = desc.tcp_data_port;
     segmentJSON["tcp_proto_version"] = desc.tcp_proto_version;
+    if (!desc.tcp_instance_id.empty()) {
+        segmentJSON["tcp_instance_id"] = desc.tcp_instance_id;
+    }
     segmentJSON["timestamp"] = getCurrentDateTime();
 
     return 0;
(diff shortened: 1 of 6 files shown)


catyans (approved):
LGTM. The incarnation ID is published and decoded compatibly, participates in metadata cache equality, and extends the pooled TCP connection key so a same-endpoint peer restart creates a fresh group while the old group drains. The restart, lifecycle, legacy-descriptor, and shared-endpoint cases are covered, and CI is green.

mooncake-transfer-engine/src/transfer_metadata.cpp
```diff
@@ -644,6 +650,10 @@ decodeMultiProtocolSegmentDesc(Json::Value &segmentJSON,
     desc->tcp_proto_version = segmentJSON.isMember("tcp_proto_version")
                                   ? segmentJSON["tcp_proto_version"].asInt()
                                   : 1;
+    if (segmentJSON.isMember("tcp_instance_id") &&
+        segmentJSON["tcp_instance_id"].isString()) {
+        desc->tcp_instance_id = segmentJSON["tcp_instance_id"].asString();
+    }
```
alogfans: Could we validate the length and format of tcp_instance_id  before storing it?

jacklin78911-collab: Thanks! Added validation before storing the ID: non-empty values must be exactly 32 lowercase hex characters. Missing or empty IDs remain supported for legacy peers. Added regression coverage for both decoding paths.

catyans (approved):
Thanks @alogfans for catching the validation gap, and @jacklin78911-collab for tightening it up. LGTM on the current head.

The decoder now rejects malformed non-empty incarnation IDs before either the single- or multi-protocol path stores them, while missing/empty IDs preserve legacy compatibility. The focused tests cover length, alphabet, JSON type, and both protocol shapes. CI is green.

PR #3559, opened 2026-08-20
[Bugfix][TransferEngine] Prevent peer failures from latching RDMA contexts inactive

Dao007forever:
## Description

A single dead or restarting peer can currently produce 32 consecutive all-rails-unavailable submit batches and disable a healthy local RNIC context. Because an inactive context posts no more work, CQ success cannot clear the counter; if the local port never went down, `IBV_EVENT_PORT_ACTIVE` never arrives and the context remains unavailable until process restart.

This PR keeps submit-side failures peer scoped and limits context-level health evidence to signals that directly identify the local RNIC:

- `submitPostSend()` no longer charges the context breaker when all peer rails are unavailable. Existing `RailState` and active-connect pauses continue to isolate failed/restarting peers.
- Direct local WC failures reported by `isLocalWcFailure()` retain the existing 32-consecutive-failure threshold. A successful completion resets the count.
- A local-WC trip deactivates the context and records a bounded reactivation deadline (default: 5000 ms). The monitor checks it on every loop, reactivates the context half-open, and resets the count when the deadline expires. Recovery timing remains subject to the monitor's existing 100 ms event wait.
- Fatal async events clear the breaker deadline so TTL recovery cannot resurrect an event-disabled context. `IBV_EVENT_PORT_ACTIVE` schedules delayed GID recovery only for an inactive context; redundant events on an active context preserve its local-WC streak. The GID deadline stays armed throughout the probe and retries, preventing breaker recovery from bypassing it.
- A small `WorkerPool` mutex serializes local-WC trips with fatal/GID recovery, preventing late CQEs from rearming TTL recovery after a fatal event.

Runtime setting:

- `MC_CONTEXT_PAUSE_TTL_MS` (default `5000`, range `1-600000`). Zero is rejected because it would restore the permanent latch.

No overlapping open/merged PR or open issue was found. The implementation is under 500 non-test lines, so the RFC threshold does not apply.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

Built in the repository Ubuntu 22.04 development container with CUDA, Store, examples, and benchmarks disabled. The following focused CTests pass:

```text
config_test
worker_pool_rail_state_test
rdma_endpoint_reestablish_test
rdma_async_event_drain_test
```

All 15 cases in `worker_pool_rail_state_test` pass. They cover 32 actual all-rails-unavailable submit batches leaving the context active with a zero failure count, local-WC threshold tripping, success reset, exact-deadline half-open recovery, redundant `PORT_ACTIVE` events, late completion failures during GID probing and retry, and fatal-event cancellation. The RDMA hardware cases in `rdma_endpoint_reestablish_test` skip because the container has no RDMA devices.

The PR-scoped pre-commit hooks pass, including clang-format 20, codespell, and cmake-format.

**Test results:**

- [x] Focused unit tests pass
- [x] Manual testing done

## Checklist

- [ ] I have updated the documentation (not applicable; internal recovery logic)
- [x] For changes >500 LOC: not applicable (under 500 non-test lines)

The human submitter must review every changed line before merge. Hosted Linux matrix results are pending for the updated commit.

## AI Assistance Disclosure

- [ ] No AI tools were used

OpenAI Codex implemented the reviewer-requested simplification, ran the focused Linux tests and pre-commit checks, and updated this description. The human submitter remains responsible for reviewing and defending the change end to end.


mooncake-transfer-engine/src/transport/rdma_transport/worker_pool.cpp
```diff
@@ -334,6 +342,49 @@ int WorkerPool::submitPreparedPostSend(
...
+    if (rec.tripped_now) {
+        LOG(WARNING) << "Context breaker tripped: context "
+                     << context_.deviceName() << " pausing after " << rec.streak
+                     << " consecutive " << failure_kind << " spanning "
+                     << rec.distinct_peers << " peer servers ["
+                     << rec.peer_sample << "], pause_ttl_ms="
+                     << globalConfig().context_pause_ttl_ms;
+    }
+    return rec.tripped_now;
+}
```
alogfans: `submitPostSend()`’s all-rails-unavailable path and `handleLocalFailure()`’s local WC-failure path currently share the same breaker streak via `health_tracker_`. Those two signals have different evidentiary strength: submit-side all-rails failures can still be dominated by a remote peer/path problem, while a local WC failure is direct evidence about the local RNIC/QP/MR. With the current shared streak, a sequence like 31 single-peer submit-side failures followed by 1 local WC failure can immediately trip the context breaker. That effectively lets weaker submit-side evidence “preload” the stronger local-fault path and partially undermines the `context_failure_min_peers` safeguard. At minimum, the streak should reset when the failure kind changes; more robustly, submit-side and local-completion evidence should maintain separate health state / streaks.

Dao007forever: Good point. Let me think about it.

alogfans:
Please rebase from main and perform ci check again.

catyans:
issue: `maybeActivateRecoveredContext()` can still reactivate a context after a newer fatal event. Once the monitor CASes `recovery_activate_after_ns_` to zero, a concurrent `DEVICE_FATAL`/`PORT_ERR` can clear the timer and deactivate the context, but the already-running recovery probe can then finish and call `onRecoveredContextActivate()`. The tracker mutex keeps each flag/tracker transition internally consistent, but it does not establish event ordering or invalidate that stale probe.

Please gate recovery with a generation/token (or serialize the probe with fatal-event handling) and add a regression for `PORT_ACTIVE -> recovery probe starts -> fatal event -> probe completes`; the final state must remain inactive until a subsequent valid recovery event.

Dao007forever:
I traced the production call paths and don’t believe this interleaving is possible. `maybeActivateRecoveredContext()` and `doProcessContextEvents()` both run synchronously on the single monitorWorker thread; `doProcessContextEvents()` is the only production caller of `handleContextEvent()`. If a fatal event arrives during the recovery probe, it remains queued until the probe returns, after which the same monitor thread processes it and leaves the context inactive.

mooncake-transfer-engine/src/transport/rdma_transport/worker_pool.cpp
Dao007forever: Addressed by separating submit-side and local-completion evidence into independent streaks under one coordinated breaker. Added regression coverage for the 31-submit + 1-local sequence and the reverse case.

Now either type needs to reach 32 to trip the breaker.

catyans (approved):
Thanks for clarifying the monitor-thread serialization; I agree the previously described interleaving is not reachable in production. LGTM. The current head also separates submit-side and local-completion streaks and keeps breaker/active transitions coordinated. The remaining ub-mock failure is in the unrelated `RuntimeQueueDispatch.RejectsCancellationForUnsupportedTransport` test; please rerun or resolve it before merge.

Dao007forever:
`ImportError: Mooncake PG was not built against torch==2.14.0.` => could we fix this?

staryxchen:
The incident diagnosis looks right: a single dead/restarting peer was treated as local RNIC failure, and `set_active(false)` had no recovery path unless `IBV_EVENT_PORT_ACTIVE` arrived.

I am concerned the solution is too heavy to maintain. `ContextHealthTracker` is a state machine for composing three behaviors (keep the submit-side sensor, add TTL half-open, coexist with the GID probe / fatal events) and then keeping `tripped_` in sync with `context_.active()` via lock-held callbacks. Dual evidence channels, a 64-peer set, an injected clock, and rollback knobs that re-enable the old failure modes all follow from that. That is a lot of mechanism relative to `ConnectPauseTracker`, and later recovery changes have to reason about all of it.

The problem only needs two facts: consecutive local-failure evidence, and whether this inactive period is allowed to self-recover. `active_` and `recovery_activate_after_ns_` already cover “in service” and event-driven recovery; a breaker pause is a second deadline, not a second object. I would:

- not charge the context breaker from submit-side all-rails-unavailable (that is usually one peer; `RailState` and #2941 already isolate peers; local evidence stays async events + `isLocalWcFailure()`)
- on local-WC trip, `set_active(false)` and store `reactivate_at`; fatal events clear it so TTL cannot resurrect an event-disabled context; `PORT_ACTIVE` keeps the GID probe and also clears it
- on the monitor tick, if `reactivate_at` is due and no GID probe is armed, `set_active(true)` and reset the counter
- if a silent-NIC heuristic is still needed, snapshot `rail_states_` instead of accumulating a submit streak

That is `failure_count` + `reactivate_at` plus fields that already exist. Are you open to collapsing it this way, or is there a failure mode that still needs the submit-side trip and the extracted machine?

Dao007forever:
> if a silent-NIC heuristic is still needed, snapshot rail_states_ instead of accumulating a submit streak

I think this part is equally complex?

staryxchen:
@Dao007forever Fair point — that snapshot was only a fallback if we still wanted a silent-NIC heuristic. I don’t think it is required. If we don’t charge the context breaker from submit-side all-rails-unavailable, neither the streak nor a `rail_states_` scan is needed for the incident.

The rest of the proposal is the part I care about: local evidence only (async events + `isLocalWcFailure()`), and a `reactivate_at` deadline instead of `ContextHealthTracker`. Are you aligned with dropping the submit-side trip and the extracted state machine, or is there a failure mode that still needs them?

Dao007forever:
@staryxchen Got it. I've updated the PR. Thanks for the suggestion!

mooncake-transfer-engine/src/transport/rdma_transport/worker_pool.cpp
```diff
@@ -1295,9 +1354,8 @@ void WorkerPool::maybeActivateRecoveredContext() {
                   << " GID changed during recovery, disconnected all endpoints";
     }
 
-    context_.set_active(true);
+    resetContextBreaker(true);
```
staryxchen, line 1357: `refreshPublishedLocalGid()` failure still calls `set_active(false)` without `context_state_lock_`, and it does not clear `breaker_reactivate_after_ns_`. The CAS already set `recovery_activate_after_ns_` to 0 before the GID I/O, so a poller can trip and arm TTL during that window; a later successful probe then `resetContextBreaker(true)` and overrides it. Please go through `resetContextBreaker(false)` here, and do not leave event-owned recovery with the GID timer cleared while local-WC trips are still allowed.

mooncake-transfer-engine/tests/worker_pool_rail_state_test.cpp
```diff
@@ -202,4 +246,98 @@ TEST_F(WorkerPoolRailStateTest, RailsArePausedIndependently) {
...
+// context breaker. This is the incident shape: one restarting peer must not
+// remove a healthy local RNIC from service.
+TEST_F(WorkerPoolRailStateTest, PeerRailFailuresDoNotDeactivateContext) {
+    WorkerPoolTestPeer::setContextActive(*worker_pool_, true);
+    for (int i = 0; i < WorkerPoolTestPeer::contextFailureThreshold(); ++i)
+        failRail(kPeerA, /*immediate_pause=*/true);
+
+    EXPECT_TRUE(WorkerPoolTestPeer::contextActive(*worker_pool_));
+    EXPECT_EQ(WorkerPoolTestPeer::contextFailureCount(*worker_pool_), 0);
+}
```
staryxchen, line 259: This does not cover the 7/15 incident. `markRailFailed()` never charged the context breaker; the old trip was `submitPostSend()` treating all-rails-unavailable as local RNIC failure. A test that only pauses rails would pass on main. Please assert that an all-rails-unavailable submit leaves `context_failure_count_` at 0 and the context active.

mooncake-transfer-engine/src/transport/rdma_transport/worker_pool.cpp
```diff
@@ -1245,8 +1302,10 @@ bool WorkerPool::handleContextEvent(ibv_event_type event_type,
     } else if (event_type == IBV_EVENT_PORT_ACTIVE) {
         // PORT_ACTIVE only means the link started coming back. Real mlx5/RoCE
         // data path can still reject RTR for a while after link-up, so delay
-        // publishing this local RNIC back to metadata. Injected tests follow
-        // the same path.
+        // publishing this local RNIC back to metadata. Clear any local-WC
+        // breaker deadline so its shorter TTL cannot bypass this recovery
+        // delay; successful recovery below performs the active transition.
+        clearContextBreaker();
```
staryxchen, line 1308: `clearContextBreaker()` zeros the local-WC streak even when the context is still active. A noisy `PORT_ACTIVE` (no prior `PORT_ERR`) wipes a 31-failure count; a trip during the 30s GID wait is then undone by `resetContextBreaker(true)`. If that is intentional, please say so next to the call. Otherwise this path should only cancel an armed deadline, not a live streak.

mooncake-transfer-engine/src/transport/rdma_transport/worker_pool.cpp
```diff
@@ -1397,6 +1455,9 @@ void WorkerPool::monitorWorker() {
             // Drop expired active-connect pause entries so the map doesn't grow
             // for peers that are never re-attempted after their pause lapses.
             context_.pruneConnectPause();
+            // Half-open recovery for the context breaker: reactivate the
+            // context once its pause TTL has elapsed.
+            maybeReactivateContext();
```
staryxchen, line 1460: TTL is sampled once per second here, but `MC_CONTEXT_PAUSE_TTL_MS` accepts 1. Default 5000 is fine; the minimum is misleading. Either raise the floor to match the tick, or check the deadline on every monitor loop, same as `maybeActivateRecoveredContext()`.

mooncake-transfer-engine/src/transport/rdma_transport/worker_pool.cpp
```diff
@@ -456,6 +446,70 @@ int WorkerPool::submitPreparedPostSend(
...
+    breaker_reactivate_after_ns_ = 0;
+    context_failure_count_.store(0, std::memory_order_relaxed);
+    context_.set_active(context_active);
+}
+
+void WorkerPool::clearContextBreaker() {
+    std::lock_guard<std::mutex> lock(context_state_lock_);
+    breaker_reactivate_after_ns_ = 0;
+    context_failure_count_.store(0, std::memory_order_relaxed);
+}
```
staryxchen, line 511: These two only differ on whether they call `set_active()`. That is one function with a `bool` (or `std::optional<bool>`) for the active flag. Splitting them makes the fatal / `PORT_ACTIVE` / GID-success paths look like different state machines when they are the same deadline + count reset.

staryxchen (commented):
Follow-up on the simplified breaker: a few implementation notes on locking, tests, and the remaining helpers.

staryxchen (approved):
LGTM

Dao007forever:
@staryxchen @alogfans I can't see the merge button, could you help me merge it?

Issue #2971, opened 2026-07-17
[Bug]: ETCD OpLog is enabled implicitly for etcd HA but lacks retention and complete metadata replication


no-teet-on:
### Description

  Before the commit that introduced the MC_ENABLE_HA_METADATA_REPLICATION feature gate, selecting etcd as the HA backend automatically enabled OpLog following:

  capabilities.has_oplog_following =
      spec.type == HABackendType::ETCD;

  As a result, every deployment using etcd for leader election also entered the etcd OpLog replication path, with no way to use etcd only for leadership coordination.

  ### Current problems

  1. Leader election is tightly coupled with metadata replication

     Using ha_backend_type=etcd implicitly enables OpLog following. Users cannot independently enable etcd leader election while disabling the experimental metadata replication path.

  2. OpLog entries do not have a TTL

     Entries under the following prefix are written using regular etcd Put or BatchCreate operations without an attached lease:

     /oplog/<cluster_id>/<sequence_id>

     The following metadata keys also have no TTL:

     /oplog/<cluster_id>/latest
     /oplog/<cluster_id>/snapshot/<snapshot_id>/sequence_id

  3. There is no automatic OpLog cleanup

     EtcdOpLogStore::CleanupOpLogBefore() can delete entries below a sequence boundary, but it is not called by any production path. It is currently only exercised by tests.

     Therefore, once production OpLog writes are enabled, the number of keys and the etcd database size can grow indefinitely.

  4. The primary write path is not fully integrated

     The production request path does not appear to:
      - Create an etcd OpLog store in writer mode.
      - Attach it through OpLogManager::SetOpLogStore().
      - Append production metadata mutations to the OpLog.

     Consequently, a standby may start an OpLog watcher even though the active primary is not publishing the required records.

  5. Promotion does not hand replicated metadata to the serving master

     HotStandbyService maintains an internal metadata store and provides ExportMetadataSnapshot(). However, the promotion path creates a new WrappedMasterService without transferring the replicated
     standby metadata or its last applied sequence ID.

     This means that successfully applying OpLog records in the standby does not necessarily make that state available to the newly promoted serving master.

  ### Impact

  - etcd HA deployments unexpectedly depend on an incomplete replication path.
  - Enabling OpLog may give users a false expectation of lossless metadata failover.
  - Stale or empty metadata may be served after failover.
  - OpLog keys may accumulate indefinitely and eventually consume the etcd storage quota.
[...]

PR #3566, opened 2026-08-21
[Bugfix][Store] Allow HA promotion without OpLog

Icedcoco:
## Description

Allow HA mode to enter the serving state when OpLog and snapshot recovery are both disabled.

When `enable_ha=true` and `enable_oplog=false`, the HA supervisor uses `NoopStandbyController`. After acquiring leadership, the supervisor calls `PromoteStandbyAndExport()`, but the no-op implementation returned `UNAVAILABLE_IN_CURRENT_STATUS`. As a result, the elected leader could never enter the serving state.

The no-op controller now returns an empty `PromotionContext`, which represents that there is no standby metadata to restore. A regression test covers the HA-without-OpLog promotion path.

This addresses the HA-only startup regression reported in #2971.

This change is separate from #3497: that PR handles actual promotion and restore failures, while this PR fixes the expected-success path when no standby recovery capability is enabled.

## Module

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

**Test commands:**

```bash
cmake --build build --target hot_standby_snapshot_bootstrap_test -j32

GTEST_FILTER=StandbyControllerTest.HaWithoutOplogPromotesWithEmptyContext \
  build/mooncake-store/tests/hot_standby_snapshot_bootstrap_test

build/mooncake-store/tests/hot_standby_snapshot_bootstrap_test

pre-commit run --files \
  mooncake-store/src/ha/standby_controller.cpp \
  mooncake-store/tests/ha/standby/hot_standby_snapshot_bootstrap_test.cpp
```

Manual HA startup was also verified against a local etcd instance with:

```bash
mooncake_master \
  --enable_ha=true \
  --enable_oplog=false \
  --ha_backend_type=etcd \
  --etcd_endpoints=127.0.0.1:23791
```

The master successfully acquired leadership and `/health` reported:

```json
{
  "status": "ok",
  "role": "leader",
  "ha_state": "serving",
  "service_ready": true
}
```

**Test results:**

- [x] Focused regression test passes
- [x] Full `hot_standby_snapshot_bootstrap_test` passes (7/7)
- [x] Manual HA startup without OpLog passes
- [x] Pre-commit checks pass

## Checklist

- [x] I have formatted my code using pre-commit
- [x] Documentation is not required for this narrowly scoped bug fix
- [x] An RFC is not required for this 21-line bug fix

## AI Assistance Disclosure

- [x] AI tools were used

Codex assisted with root-cause analysis, implementation, test construction, and local verification. The submitter is responsible for reviewing and understanding every changed line.

### DIFF
--- a/mooncake-store/src/ha/standby_controller.cpp
+++ b/mooncake-store/src/ha/standby_controller.cpp
@@ -83,7 +83,7 @@ class NoopStandbyController final : public StandbyController {
 
     tl::expected<PromotionContext, ErrorCode> PromoteStandbyAndExport()
         override {
-        return tl::unexpected(ErrorCode::UNAVAILABLE_IN_CURRENT_STATUS);
+        return PromotionContext{};
     }
 
     void UpdateObservedLeader(const std::optional<MasterView>&) override {}
--- a/mooncake-store/tests/ha/standby/hot_standby_snapshot_bootstrap_test.cpp
+++ b/mooncake-store/tests/ha/standby/hot_standby_snapshot_bootstrap_test.cpp
@@ -186,6 +186,25 @@ TEST(StandbyControllerTest, OplogEnablementControlsReaderController) {
     EXPECT_EQ(ErrorCode::OK, unsupported->PromoteStandby());
 }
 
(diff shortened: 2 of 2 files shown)


Aionw (approved):
LGTM

PR #3567, opened 2026-08-21
[Wheel] Support DataProto matrix partial reads

zxpdemonio:
## Description

Add partial matrix reads for dense two-dimensional DataProto `batch` fields.

The public `batch_slices` mapping selects a column slice per field and composes
with the existing `rows` selection. A single M x N implementation covers row
batches, column batches, and general submatrices without separate code paths.

The implementation:

- plans byte ranges from matrix shape, dtype, selected rows, and column slice;
- coalesces consecutive full-width rows;
- uses one BufferPool lease and one `get_into_ranges` call across selected
  fields, DataProto stages, and payload chunks;
- keeps NumPy results writable and rewrites native tensor metadata for the
  sliced shape;
- commits caller destinations only after every selected field and the final
  DataProto result have been constructed successfully;
- falls back to regular materialization when ranged reads are unavailable or
  the request exceeds the range or physical-fragment limits.

The fast path guarantees one batched ranged call. The fallback is intentionally
whole-request and does not guarantee a single backend call; this keeps the
destination commit semantics and the hot path small. Logical ranges are capped
per field, while physical fragments are capped at 4096 across the entire
batched request before BufferPool allocation.

The append write path is unchanged. Dense matrix fields added through
`append_dataproto_fields()` are resolved from their existing stage refs and can
participate in the same batched partial read.

## Module

- [ ] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [x] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
cd mooncake-wheel
python3 -m pytest -q tests/test_partial_read.py
python3 -m pytest -q tests/test_structured_object_store.py \
  -k "dataproto_matrix or empty_matrix_read"
python3 -m pytest -q \
  tests/test_structured_object_store.py::test_dataproto_later_field_failure_does_not_commit_matrix_destination \
  tests/test_structured_object_store.py::test_dataproto_result_failure_does_not_commit_matrix_destination \
  tests/test_structured_object_store.py::test_dataproto_copied_result_keeps_matrix_destination
python3 -m pytest -q \
  tests/test_structured_object_store.py::test_dataproto_helper_appends_stage_fields_without_rewriting_existing

python3 -m py_compile \
  mooncake/_partial_read.py \
  mooncake/structured_object_store.py \
  tests/test_partial_read.py \
  tests/test_structured_object_store.py
```

**Test results:**
- [x] Unit tests pass: planner 15 passed; matrix path 7 passed; destination
  atomicity 3 passed; append regression 1 passed
- [ ] Integration tests pass (not run in this environment)
- [ ] Manual testing done

The broader structured-object suite completed with 90 passed and 28 skipped.
Three native fast-copy PUT tests fail because the installed
`_concat_arrays_into` extension returns zero bytes. The same three tests fail
identically on current `main` at `266a966b`.

Ruff check and format validation pass for both new files. `py_compile` and
`git diff --check` pass for all changed Python files. Full-file `ruff-format`
was not applied to the existing legacy files because it rewrites hundreds of
unrelated lines.

## Checklist

- [x] For changes >500 LOC: I have filed an RFC issue (not applicable: 401
  production additions, 417 including documentation)

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex was used to help implement, simplify, review, and validate the change.
The human submitter is responsible for reviewing and defending the final code.


zxpdemonio:
@stmatengss @yokinoshitayoki Could you please help review this PR? In particular, feedback on the DataProto matrix partial-read semantics and compatibility would be appreciated. Thanks!

mooncake-wheel/mooncake/structured_object_store.py
```diff
@@ -906,9 +968,60 @@ def _get_dataproto(
...
                 )
         meta_info = _select_mapping(ref.meta_info, meta_info_keys)
-        return _build_dataproto_like_result(
+        staged_matrices = {read.name: batch[read.name] for read in matrix_reads}
+        for read in matrix_reads:
+            if read.destination is not None:
+                batch[read.name] = read.destination
+        result = _build_dataproto_like_result(
             batch, non_tensor_batch, meta_info, data_cls
         )
```
yokinoshitayoki, line 1079: The constructor receives destinations before they contain the fetched data. If data_cls.from_dict() copies or converts a matrix, the returned object retains stale values even though the destination is populated afterward. Construct the result from initialized staging values, then commit destinations and rebind references after successful construction.

mooncake-wheel/mooncake/structured_object_store.py
```diff
@@ -3223,6 +3336,78 @@ def materialize_into(
...
+            [read.decode(data) for read, data in zip(reads, staged)]
+            if staged is not None
+            else [self._materialize_matrix_fallback(read) for read in reads]
+        )
+        return dict(zip((read.name for read in reads), values)), reads
+
+    def _materialize_matrix_fallback(self, read: _MatrixRead) -> Any:
+        value = self._read_structured_member(
+            read.name, read.payload_spec, read.field_spec, None, None
+        )
```
yokinoshitayoki, line 3546: After the nonblocking staging allocation fails, this call passes destination=None and re-enters read_payload_range_into_pool_array(), which calls pool.acquire() with blocking enabled by default. GET can wait indefinitely while the caller holds existing leases, even with a supplied NumPy destination. Use heap staging or consistently nonblocking allocation in the fallback.

mooncake-wheel/mooncake/structured_object_store.py
zxpdemonio: Addressed in cb9ec47b93d73b6019acbe8d2c2eaf489ef18a8f. The result is now constructed from staging values first. All DataProto.batch destination bindings are preflighted before any destination is written, and the constructed result is rebound after the destination commit. Added regression coverage for copied, immutable, and rejecting batch mappings.

mooncake-wheel/mooncake/structured_object_store.py
zxpdemonio: Addressed in cb9ec47b93d73b6019acbe8d2c2eaf489ef18a8f. The matrix fallback path now disables BufferPool usage, so it cannot re-enter a blocking pool.acquire(). Added a regression test that forces the fallback with MAX_MATRIX_RANGES=1 and verifies that no BufferPool allocation is attempted.

yokinoshitayoki:
@zxpdemonio I tried to rerun these two checks but still failed. Could you take a look?
[Build & Test (Linux) / Check Python with Ruff (pull_request)](https://github.com/kvcache-ai/Mooncake/actions/runs/34221069144/job/102371360300?pr=3567)
[Build & Test (Linux) / CI Gate (pull_request)](https://github.com/kvcache-ai/Mooncake/actions/runs/34221069144/job/102371938517?pr=3567)

Aionw (approved):

ykwd (approved):

yokinoshitayoki (approved):

stmatengss (approved):

PR #3570, opened 2026-08-21
[TENT] fix: synchronize with caller's per-thread stream before issuing NVLink/MNNVL batched copy

TTThanos:
## Description

TENT's NVLinkTransport and MnnvlTransport submit cudaMemcpyBatchAsync on an internal pooled stream with **no synchronization** against the caller's cudaStreamPerThread. When the caller queues producer work (e.g. a KV compute kernel) on its per-thread stream and then immediately submits a transfer, the copy can race with—and read the source buffer before—that work has executed, silently moving stale data.

The legacy IntraNodeNvlinkTransport prevents this by recording an event on cudaStreamPerThread and cudaEventSynchronize-ing on it before issuing the copy. Port the same contract into TENT:

  * Add CallerSyncEventPool — a thread-local, per-device cache of cudaEventDisableTiming events (lazy create, destroy at thread exit with device restore).
  
  * Add syncWithCallerStream() — records the event on cudaStreamPerThread and CPU-blocks until it completes, guaranteeing all producer work queued on the caller's stream is visible before the transport's internal stream submits the batched copy.
  
  * Call syncWithCallerStream() at the top of submitTransferTasks() in both NVLinkTransport and MnnvlTransport, right after the batch capacity check and before any buffer lookup or copy submission.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

A new test tent_nvlink_stream_sync_race_test exercises the full NVLinkTransport API path (IPC export → p2p metadata → openSegment → cudaIpcOpenMemHandle → batched copy → status poll). It uses a double-fork architecture so the gtest process never touches CUDA, and detects the race via byte-pattern markers (0x11 stale / 0x77 fresh / 0xEE read-back sentinel): the producer floods cudaStreamPerThread with queued work, then immediately submits a WRITE; if the transport copies before the producer executes, the destination holds stale bytes.

**Test commands:**
```bash
```

**Test results:**
- [ ] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used
Code and test were co-authored with an AI coding assistant GLM5.2

### DIFF
--- a/mooncake-transfer-engine/tent/src/transport/mnnvl/mnnvl_transport.cpp
+++ b/mooncake-transfer-engine/tent/src/transport/mnnvl/mnnvl_transport.cpp
@@ -23,6 +23,7 @@
 #include <cstdint>
 #include <iomanip>
 #include <memory>
+#include <unordered_map>
 
 #include <cuda.h>
 #include <cuda_runtime.h>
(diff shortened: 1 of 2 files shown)


alogfans (approved):
LGTM

Issue #3452, opened 2026-08-15
[Bug]: Memory leak: mooncake‑master OOM restart under long‑context PD‑separation workload


Lichunyan3:
### Environment

- SGLang version: `0.5.15.post1+pr #27010`
- Mooncake‑master version: `0.3.12.post1`
- Hardware/Network: H100, PD‑separate architecture, RDMA enabled, serving GLM‑5.2-W4AFP8 model

#### mooncake:
mooncake_master \
  -enable_ha=true \
  -etcd_endpoints="http://10.214.81.67:32379" \
  -rpc_address="10.214.81.67" \
  -rpc_port=50063 \
  -max_threads=64 \
  -metrics_port=9003 \
  -eviction_high_watermark_ratio=0.9 \
  -cluster_id="$MC_METADATA_CLUSTER_ID"

#### p-master config:
GLOO_SOCKET_IFNAME=bond1.1081 \
NCCL_SOCKET_IFNAME=bond1.1081 \
NCCL_IB_DISABLE=0 \
NCCL_P2P_DISABLE=0 \
NCCL_IB_TC=106 \
NCCL_IB_HCA=mlx5_2,mlx5_3,mlx5_4,mlx5_5,mlx5_7,mlx5_8,mlx5_9,mlx5_10 \
python3 -m sglang.launch_server \
  --model /models/GLM-5.2-W4AFP8 \
  --served-model-name glm-5.2 \
  --tp 8 \
  --pp-size 2 \
  --reasoning-parser glm45 \
  --tool-call-parser glm47 \
  --mem-fraction-static 0.85 \
  --host 0.0.0.0 \
  --port 8000 \
  --dist-init-addr 10.214.81.67:5000 \
  --nnodes 2 \
  --node-rank 0 \
  --trust-remote-code \
  --quantization w4afp8 \
  --disable-shared-experts-fusion \
  --attention-backend dsa \
  --max-running-requests 16 \
  --cuda-graph-max-bs 16 \
  --chunked-prefill-size 12288 \
  --enable-prefill-cp \
  --cp-strategy interleave \
  --watchdog-timeout 3000 \
  --disaggregation-ib-device mlx5_2,mlx5_3,mlx5_4,mlx5_5,mlx5_7,mlx5_8,mlx5_9,mlx5_10 \
  --disaggregation-mode prefill \
  --kv-cache-dtype fp8_e4m3 \
  --page-size 64 \
  --enable-hierarchical-cache \
[...]

PR #3576, opened 2026-08-21
[Store] Shrink metadata maps after eviction cycles

jfeng18:
## Description

Contributes to #3452 (mooncake-master production OOM / RSS audit).

`std::unordered_map::erase` never reduces the bucket array. A metadata shard that once held millions of keys therefore keeps its high-water bucket memory (8 bytes/bucket) forever — `master_service.cpp` previously contained no `rehash()`/`shrink_to_fit()` call at all. After a 25M-object peak, even if eviction removes 90% of the entries, the bucket arrays stay at peak size.

This PR adds a conditional shrink after each `BatchEvict` cycle:

- **Condition**: `bucket_count() > kShrinkMinBucketCount (1024) && size() < bucket_count() / 4`. The bucket floor avoids rehash churn on small maps; the quarter threshold plus the 2x headroom (`rehash(size() * 2)`) keeps a freshly shrunk map from immediately growing again.
- **Scope**: `TenantState::metadata` (the dominant per-object map), only for shards that actually evicted in the current cycle. `BatchEvict` is already driven periodically, so no extra throttling is needed.
- **Lock & iterator safety**: the shrink runs while holding the shard lock via `MetadataShardAccessorRW`, in a dedicated pass after eviction finishes. The pass iterates `shard->tenants` — not the map being rehashed — so no live iterator is invalidated.

The shrink logic is a small reusable helper (`ShrinkBucketsIfSparse`) in `master_service.h`.

## Module

- [x] Mooncake Store

## Type of Change

- [x] Performance optimization

## Testing

- `ShrinkBucketsIfSparseThresholds`: unit test for the helper — small maps untouched, no shrink at exactly a quarter full, shrink once below.
- `BatchEvictShrinksSparseMetadataMaps`: discriminating service-level test — 2048 same-shard objects inserted (plus one hard-pinned survivor so the tenant deterministically outlives full eviction), then `BatchEvict`; asserts the shard's metadata `bucket_count` drops below half its high-water mark. Reverting the fix makes this assertion fail because the bucket array never shrinks.
- `clang-format --dry-run --Werror` clean on all touched files. Local machine cannot build mooncake-store (dependency set too large); compilation is verified by CI.

## AI Assistance Disclosure

AI-assisted (code drafted with an AI agent); the change has been human-reviewed line by line.

### DIFF
--- a/mooncake-store/include/master_service.h
+++ b/mooncake-store/include/master_service.h
@@ -105,6 +105,25 @@ namespace benchmarks {
 class BatchEvictBench;
 }  // namespace benchmarks
 
+// std::unordered_map/set never shrink their bucket array on erase, so a
+// container that once held millions of entries keeps its high-water bucket
+// memory (8 bytes per bucket) forever. ShrinkBucketsIfSparse rehashes a
+// container down to roughly twice its live size once the bucket array is
+// both large enough to matter and less than a quarter full. The bucket
+// floor avoids rehash churn on small containers; the 2x headroom keeps a
+// freshly shrunk container from growing again right away.
+// Rehashing invalidates iterators: callers must hold the lock guarding the
+// container and must not be iterating it.
+inline constexpr size_t kShrinkMinBucketCount = 1024;
+
+template <typename UnorderedContainer>
+void ShrinkBucketsIfSparse(UnorderedContainer& container) {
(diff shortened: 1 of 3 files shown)


Aionw (approved):
LGTM

ykwd (approved):

PR #3391, opened 2026-08-11
[TransferEngine] Support multi-target tebench runs

zxpdemonio:
## Description

This PR implements the first code step from RFC #3371: generalizing `tebench` from a single-target benchmark driver to an endpoint-level multi-target benchmark driver.

The change keeps topology orchestration outside `tebench`. `tebench` only provides the primitive to open multiple target segments and route requests to a selected target id. Launchers can then express 1-to-N fan-out, M-to-1 incast, M-to-N, and all-to-all by starting peers with different comma-separated `--target_seg_name` values.

Related RFC: https://github.com/kvcache-ai/Mooncake/issues/3371

Main changes:

- Extend the existing `--target_seg_name` option to accept comma-separated target segment names while preserving single-target behavior.
- Carry `target_id` through `BenchRunner::runSingleTransfer()` and backend implementations so each transfer is submitted to the selected remote segment.
- Support multi-target execution for both classic TransferEngine and TENT backends.
- Add `write_seed` and `read_verify` operations for multi-process shared-target verification flows.
- Add overflow-safe range helpers for address layout validation.
- Document the boundary: `tebench` provides benchmark knobs and metrics; external launchers own node orchestration and topology construction.

Example usage:

1. 1-to-N fan-out: one initiator sends traffic to two targets.

```bash
# Target A
./tebench --backend=tent --metadata_type=p2p \
  --rpc_server_port=41081 \
  --seg_name=target_a \
  --total_buffer_size=33554432 \
  --xport_type=rdma --tent_transport_hint=rdma

# Target B
./tebench --backend=tent --metadata_type=p2p \
  --rpc_server_port=41082 \
  --seg_name=target_b \
  --total_buffer_size=33554432 \
  --xport_type=rdma --tent_transport_hint=rdma

# Initiator
./tebench --backend=tent --metadata_type=p2p \
  --rpc_server_port=41083 \
  --seg_name=initiator_0 \
  --target_seg_name=10.0.0.11:41081,10.0.0.12:41082 \
  --total_buffer_size=33554432 \
  --xport_type=rdma --tent_transport_hint=rdma \
  --start_block_size=4096 --max_block_size=4096 \
  --start_batch_size=1 --max_batch_size=1 \
  --start_num_threads=4 --max_num_threads=4 \
  --duration=10 --check_consistency=true
```

2. M-to-1 incast: start one target, then start multiple initiators with the same target list.

```bash
# Target
./tebench --backend=tent --metadata_type=p2p \
[...]

PR #3578, opened 2026-08-21
[CI] Serialize mmap_arena_fallback_test to curb coverage flake

jfeng18:
## Description

Mitigates a recurring CI flake in `mmap_arena_fallback_test` by running it serially (`RUN_SERIAL TRUE`). One-line CMake change, no test code touched.

### Symptom

- Under the coverage build with `ctest --parallel $(nproc)`, the test intermittently fails at the **suite level**: every gtest case prints `[ OK ]`, yet ctest reports the test as `Failed` (non-zero process exit).
- A plain re-run of the job passes. This has broken CI for several unrelated PRs in the past. Since #3482 the test runs in nightly, so the flake now produces recurring nightly red noise instead.

### Root cause hypothesis (suspected, ~70% confidence)

- CI builds with `CXXFLAGS=--coverage` and runs `ctest --parallel $(nproc)`, so 100+ test binaries share the same set of `.gcda` files.
- libgcov serializes `.gcda` merge writes at `atexit` using `fcntl` file locks. On GitHub Actions (overlayfs), this locking is unreliable under high concurrency; a failed merge can rewrite the process exit code.
- That matches the observed signature exactly: all cases pass, the process still exits non-zero, and a re-run (less contention) succeeds.
- `mmap_arena_fallback_test` exercises a wide slice of `utils.cpp` / `mmap_arena.cpp`, giving it unusually heavy `.gcda` write traffic at exit — a natural repeat offender.

No existing issue or open PR covers this; nobody appears to be working on it.

### Why `RUN_SERIAL`

- One line, zero risk to test semantics or coverage data collection — it only removes the concurrency window during this binary's gcov flush.
- If the flake ever recurs even when serialized, the escalation path is per-test `GCOV_PREFIX` isolation (separate `.gcda` directories per test), which is a larger change and not justified yet.

### Impact

- Only a marginal increase in total ctest wall time for coverage/nightly runs; no functional change.

## Module

- [x] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [x] Other: CI flake mitigation

## How Has This Been Tested?

CMake-only change (test property). Verified `cmake-format --check` (v0.6.13, same as the pre-commit hook) passes on the modified file. Since the flake is intermittent and CI-environment-specific, effectiveness will be observed via nightly runs after merge.

**Test commands:**
```bash
cmake-format --check mooncake-store/tests/CMakeLists.txt
```

**Test results:**
- [ ] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

AI assistance was used for the flake root-cause investigation and drafting this change; the submitter reviewed and stands behind the analysis and the one-line mitigation.

### DIFF
--- a/mooncake-store/tests/CMakeLists.txt
+++ b/mooncake-store/tests/CMakeLists.txt
@@ -196,6 +196,12 @@ add_store_test(health_check_test health_check_test.cpp)
 add_store_test(http_metadata_server_test http_metadata_server_test.cpp)
 add_store_test(mmap_arena_test mmap_arena_test.cpp)
 add_store_test(mmap_arena_fallback_test mmap_arena_fallback_test.cpp)
+# mmap_arena_fallback_test exercises a large slice of utils.cpp/mmap_arena.cpp,
+# so under --coverage builds it produces heavy .gcda merge traffic at exit.
+# Running it in parallel with the rest of the suite intermittently trips
+# libgcov's file locking on CI: every gtest case passes yet the process exits
+# non-zero. Serialize it as mitigation; see the PR for the full analysis.
+set_tests_properties(mmap_arena_fallback_test PROPERTIES RUN_SERIAL TRUE)
 add_store_test(object_data_type_test object_data_type_test.cpp)
 add_subdirectory(e2e)
 


Aionw:
It seems https://github.com/kvcache-ai/Mooncake/pull/3391 fixed this already.

jfeng18:
Thanks for looking! Verified against #3391's diff — it takes a different angle on this: the LSan suppression for libnuma (the weak-symbol visibility path, which was one of the two candidate mechanisms in the PR description) is now hardened there, which plausibly explains the flakes you've seen stop.

That said, the `.gcda` parallel-merge race (the other candidate, from coverage builds with `ctest --parallel`) is orthogonal to LSan and not addressed by #3391 — the test still runs in parallel under `--coverage`. The failure signature we hit on PR CI (every gtest case green, process exits non-zero, rerun passes) matched that mechanism, and that environment still exists in nightly.

I'll close this since the observed flake has an upstream explanation and the one-line mitigation here wasn't load-bearing. If the signature recurs on nightly coverage runs, the RUN_SERIAL (or GCOV_PREFIX isolation) route documented here is the ready-made fix — happy to resend it then.

PR #3581, opened 2026-08-21
[Bugfix][TENT] Refresh SHM mappings across reallocations

982945902:
## Description

TENT's SHM relocation cache keyed mappings only by remote segment ID and virtual address. If a peer reallocated a shared-memory buffer at the same virtual address but published a new `shm_path`, the consumer silently reused the previous mapping and read stale data.

This change makes the cached mapping identity include the published `shm_path` and buffer length. When the identity changes, SHM maps the new backing object and retires the previous mapping until transport shutdown, preserving the existing safety guarantee for copies that may still hold an older resolved address.

## Module

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

**Test commands:**
```bash
cmake --build /workspaces/build --target tent_shm_transport_test -j8
/workspaces/build/tests/tent_shm_transport_test
```

**Test results:**
- [x] Unit tests pass
- [x] Manual testing done: reproduced same-address reallocation returning generation A before the fix; the regression test now reads generation B while confirming generation A remains mapped.

## Checklist

`pre-commit` is not installed in the available development image; `git diff --check` and the PR-scoped C++ formatting script pass.

## AI Assistance Disclosure

Codex helped identify and reproduce the stale SHM mapping, implement the generation-aware cache behavior, and add the regression test. The human submitter should review and understand every changed line before merge.


alogfans (approved):
LGTM. However, delaying munmap could increase memory usage over time, especially if the same address is reallocated frequently.

catyans (approved):
LGTM. Reallocation now refreshes mappings by base/length/`shm_path` while keeping old mappings valid for concurrent copies. The regression covers a new SHM object at the same remote address; the remaining CI failure is an unrelated sccache download 504.

alogfans:
@copilot resolve the merge conflicts in this pull request

982945902:
Resolved the merge conflicts by merging the latest `main` into this branch.

The conflict was limited to `shm_transport_test.cpp`; I kept both the SHM reallocation regression test from this PR and the newer all-or-nothing submission regression test from `main`.

Validation:
- `tent_shm_transport_test` passed
- `clang-format --dry-run --Werror` passed for the three files changed by this PR

The PR is now reported as mergeable, and CI has been re-triggered.

alogfans:
Does CI failure related to this pr? You can rebase code and try again.

catyans:
I took a look at the three red jobs @alogfans mentioned. They fail in `MasterServiceSSDSnapshotTest.EvictObject`, `RuntimeQueueDispatch.RejectsCancellationForUnsupportedTransport`, and the SSD offload concurrent-stress test (948 read failures). None of those paths exercises the SHM mapping/reallocation code changed here.

The current head already merged main on Sep 3, and the author reported the focused `tent_shm_transport_test` passing after resolving that conflict. So these look unrelated to the PR itself, although another rebase/rerun is still the practical way to clear the gate and may pick up fixes for those newer flakes.

PR #3582, opened 2026-08-21
[Bugfix][TENT] Preserve ownership after failed memory free

982945902:
## Description

`TransferEngineImpl::freeLocalMemory()` removed an allocation from `allocated_memory_` even when the owning transport returned an error. The memory could remain allocated while the engine permanently lost the ownership record, so callers could not retry the release.

This change commits the ownership removal only after the transport reports a successful free. A fail-once transport regression test verifies that the first error preserves the allocation, the second call reaches the transport and succeeds, and a later call is correctly rejected after ownership has been released.

## Module

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

**Test commands:**
```bash
cmake --build /workspaces/build --target \
  tent_local_memory_lifecycle_test tent_runtime_queue_dispatch_test -j8
/workspaces/build/tests/tent_local_memory_lifecycle_test
/workspaces/build/tests/tent_runtime_queue_dispatch_test
```

**Test results:**
- [x] Unit tests pass
- [x] Manual testing done: before the fix, a fail-once transport was called only once and the retry returned `Address region not registered`; the regression test now observes two transport free calls and a successful retry.

## Checklist

`pre-commit` is not installed in the available development image; `git diff --check` and the PR-scoped C++ formatting script pass.

## AI Assistance Disclosure

Codex helped identify and reproduce the ownership-loss path, implement the success-only ownership commit, and add the fail-once regression test. The human submitter should review and understand every changed line before merge.

### DIFF
--- a/mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
+++ b/mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
@@ -690,8 +690,9 @@ Status TransferEngineImpl::freeLocalMemory(void* addr) {
          ++it) {
         if (it->addr == addr) {
             auto status = it->transport->freeLocalMemory(addr, it->size);
+            if (!status.ok()) return status;
             allocated_memory_.erase(it);
-            return status;
+            return Status::OK();
         }
     }
     return Status::InvalidArgument("Address region not registered" LOC_MARK);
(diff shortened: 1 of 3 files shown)


mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
```diff
@@ -690,8 +690,9 @@ Status TransferEngineImpl::freeLocalMemory(void* addr) {
          ++it) {
         if (it->addr == addr) {
             auto status = it->transport->freeLocalMemory(addr, it->size);
+            if (!status.ok()) return status;
```
alogfans: Advise to prints warning.

982945902: Added a warning before returning the transport error. It includes the address, allocation size, and original status while preserving the ownership entry for a later retry. Fixed in ce87244d.

alogfans (approved):
LGTM

PR #3587, opened 2026-08-21
[Store] Split utility code by function

Aionw:
## Description

Split the monolithic Mooncake Store utility surface into focused modules:

- move result formatting, byte-size parsing, networking, and timestamp helpers into dedicated headers/sources
- isolate mmap/arena/HugeTLB/NUMA buffer handling from protocol-backed client buffer allocation
- move key-to-path resolution under `FileUtil`
- split the former utility test binary into byte-size, network, and random test targets

This is a responsibility and dependency-boundary refactor; it does not intentionally change user-facing behavior.

No overlapping open issue or PR was found. An RFC issue was not filed because this only reorganizes existing implementations and APIs without introducing a new architecture or behavior.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
pre-commit run
/tmp/byte_size_test-657e
/tmp/random_test-657e
/tmp/network_test-657e --gtest_filter=-UtilsTest.GetInterfaceIPv4AddressLoopback:UtilsTest.GetInterfaceIPv4AddressMissingInterface
/tmp/network_test-657e '--gtest_filter=UtilsTest.GetInterfaceIPv4Address*'
/tmp/mmap_arena_fallback_test-657e
```

All affected Mooncake Store production and test translation units were also recompiled using the existing CPU build's compile commands.

**Test results:**
- [x] Unit tests pass
- [ ] Integration tests pass (not run as a full suite)
- [x] Manual testing done: verified the split targets compile and link against the existing CPU build

## Checklist

- [x] I have updated the documentation (not applicable; no user-facing behavior changes)
- [ ] For changes >500 LOC: I have filed an RFC issue (not applicable; this PR only splits existing code by responsibility)

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex helped reorganize the existing utility and allocation code, update include dependencies and tests, and run validation. The human submitter reviewed and approved every changed line and is responsible for the change end to end.


Icedcoco (commented):
Thanks for the cleanup. I have two questions about the dependency split:

1. I noticed that `mooncake-store/include/utils.h` and the old `utils/*.h` paths are removed without forwarding headers. `ResolvePathFromKey` also moves from a free function to `FileUtil::ResolvePathFromKey`, while `time_gen()` becomes inline.

    My assumption is that these utilities are internal implementation details rather than part of the supported C++ API, especially since they are not installed or documented in the public API reference. Is that the intended compatibility boundary? If so, this seems reasonable, though it may be worth clarifying what “APIs unchanged” means in the PR description. If out-of-tree users are expected to include these headers directly, would keeping thin forwarding headers or compatibility wrappers be worthwhile?

2. I also noticed that `#include "common/network.h"` is placed at the end of `health_check_test.cpp`, `file_storage_test.cpp`, and `dummy_client_get_buffer_test.cpp`, after the calls to `getFreeTcpPort()`.

    These tests currently get the declaration transitively through `real_client.h`, so the trailing includes do not appear to affect compilation. Were they intended to make the dependency explicit? If so, should they be moved to the normal include block at the top? Otherwise, they may simply be removable.

Aionw:
Thanks for catching this. The `common/network.h` includes were intended to make the dependency explicit, but they were accidentally appended at the end of the files. I moved them into the normal include blocks in all three tests in 48d44a3d. The affected test targets compile successfully.

Icedcoco (approved):
LGTM

ykwd (approved):

PR #2180, opened 2026-05-22
[Store] Add opt-in grouped object routing semantics

CAICAIIs:
## Description

Refs #2127.

This PR adds opt-in grouped object semantics to Mooncake Store through `ReplicateConfig::group_ids`.

Grouped objects are routed by `hash(group_id)` while existing object-key APIs remain compatible. When grouping is enabled, the master tracks a lightweight object-to-group routing index, refreshes the current group members when any member is accessed, and expands eviction from a grouped candidate to currently safe-to-evict group members.

## Changes

- Add optional `ReplicateConfig::group_ids`.
- Validate `group_ids` only when explicitly provided.
- Treat empty string entries as ungrouped, allowing mixed grouped / ungrouped batch writes.
- Route grouped object metadata by `group_id` while preserving key-level lookup, remove, put, and upsert behavior.
- Keep group membership immutable while an object exists.
- Refresh lease protection for current group members when one member is accessed.
- Expand eviction from a grouped candidate to current group members while reusing existing object-level safety checks.
- Rebuild grouped routing state after snapshot restore.
- Preserve compatibility with legacy snapshots that do not contain `group_id`.
- Expose `group_ids` through the Python binding and preserve key-to-group ordering in batch/tensor paths.

## Scope

This PR intentionally keeps the first implementation lightweight and best-effort.

It does not add group completeness tracking, `group_size`, `group_index`, `RemoveGroup`, atomic group visibility, SGLang integration, or benchmark tooling. Those can be handled as follow-up work after the core Store semantics are reviewed.

## Module

- [ ] Transfer Engine (`mooncake-transfer-engine`)
- [x] Mooncake Store (`mooncake-store`)
- [ ] Mooncake EP (`mooncake-ep`)
- [x] Integration (`mooncake-integration`)
- [ ] P2P Store (`mooncake-p2p-store`)
- [x] Python Wheel (`mooncake-wheel`)
- [ ] PyTorch Backend (`mooncake-pg`)
- [ ] Mooncake RL (`mooncake-rl`)
- [ ] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [x] Documentation update
- [ ] Other

## How Has This Been Tested?

Ran on a Linux test host:

```bash
cmake --build build-noetcd --target master_service_test snapshot_child_process_test -j 8
./build-noetcd/mooncake-store/tests/master_service_test
[...]

Issue #2127, opened 2026-05-18
[RFC]: Lightweight Best-Effort Group Semantics for Mooncake Store Objects


yokinoshitayoki:
## Background

This is a follow-up design proposal to Issue #1887, which describes a common problem in Mooncake Store: several physical objects may represent one logical cache unit.

Examples include:

- In KV cache, the K tensor and V tensor may be stored as separate objects.
- In SGLang PP/TP deployments, the KV cache for the same token span may be split across ranks or layers.
- In DSA-like designs, one logical cache unit may include both KV objects and indexer objects.

This RFC refers to these related objects as belonging to the same group.

Today, Mooncake Store manages lifecycle at object granularity. Objects in the same logical group can be uploaded, accessed, leased, and evicted independently. As discussed in Issue #1887, this means an originally complete group may later become partial after object-level eviction removes only some of its members.

For upper layers such as KV cache systems, a partial group is often not useful: if only K remains without V, or only one TP rank keeps a token span, the logical cache entry cannot be safely reused.

## Motivation

Issue #1887 lays out the broader design space for group semantics, including:

- `group_id`
- `group_size`
- `group_index`
- group-aware existence checks
- group-level eviction
- sharding strategy
- metadata design
- `remove` semantics
- SSD offloading compatibility
- fault tolerance
- relation to prefix-aware eviction

This RFC focuses on a smaller implementation subset within that design space. The intent is to address the common eviction-fragmentation case first, while leaving stronger completeness semantics and prefix-aware policy decisions to follow-up discussion.

To understand the practical impact, we ran end-to-end experiments with real SGLang HiCache + Mooncake L3 traffic. The goal was to answer a question:

> Under real long-running HiCache traffic, does current Mooncake object-level eviction create a meaningful amount of eviction-fragmented partial logical groups?

The result is:

- Yes, object-level eviction does create partial logical groups.
- However, the bad-cache live-byte ratio drops to around 1% once L3 capacity reaches 64GB and above.

Based on the measured impact and the implementation complexity of stronger completeness semantics, this RFC proposes a lightweight, best-effort initial implementation.

The goal is not to provide transactional group semantics. The goal is to reduce the common case where eviction leaves useless partial groups behind, while keeping the design simple and backward compatible.

### Workload

The experiments used a full SGLang HiCache + Mooncake L3 path:

- SGLang server with HiCache enabled.
- Mooncake Store as the L3 backend.
[...]

PR #2508, opened 2026-06-16
[Store] Group BatchGetReplicaList metadata lookup by shard

bitborne:
## Description

This PR is a follow-up to #2405.

#2405 reduced Mooncake Store lookup-side metadata shard lock contention by avoiding fully per-key metadata lookup behavior. `BatchGetReplicaList` still used the old path: the wrapped RPC batch handler iterated over keys and called `GetReplicaList` once per key.

This PR adds a batch-aware `MasterService::BatchGetReplicaList` implementation and routes the wrapped RPC batch handler to it directly. The new path groups keys by metadata shard before accessing metadata, instead of issuing one service lookup per key.

The new path:

1. Groups batch keys by metadata shard.
   - Bucket input keys by the shard selected from tenant-aware key routing.
   - Respect grouped-object routing through `object_group_ids_`.
   - Acquire each touched shard accessor once per batch.
   - Preserve the original response order.

2. Preserves existing `GetReplicaList` behavior.
   - Keep not-found and not-ready replica semantics unchanged.
   - Keep tenant isolation unchanged.
   - Keep lease granting behavior unchanged.
   - Keep existing get metrics behavior unchanged.
   - Keep promotion-on-hit behavior unchanged.

3. Follows the shard traversal direction used by #2405.
   - Start from a randomized shard.
   - Traverse shards in reverse order from that start point.
   - Skip untouched shards.

This keeps public API behavior, request ordering, tenant semantics, lease behavior, metrics, and promotion policy unchanged. No user-facing API or configuration behavior is changed.

This PR also adds `mooncake-store/benchmarks/batch_get_replica_bench`, a benchmark for measuring `MasterClient::BatchGetReplicaList` against a running master. The benchmark supports fixed-request mode for fair baseline/current comparisons and optional CSV output for external result aggregation.

## Module

- [ ] Transfer Engine (`mooncake-transfer-engine`)
- [x] Mooncake Store (`mooncake-store`)
- [ ] Mooncake EP (`mooncake-ep`)
- [ ] Integration (`mooncake-integration`)
- [ ] P2P Store (`mooncake-p2p-store`)
- [ ] Python Wheel (`mooncake-wheel`)
- [ ] PyTorch Backend (`mooncake-pg`)
- [ ] Mooncake RL (`mooncake-rl`)
- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Breaking change
[...]

PR #3589, opened 2026-08-22
[Store] Decouple object routing from grouped-object semantics

CAICAIIs:
## Description

Refs #2127, #2180.

This PR reworks grouped-object semantics so that object routing is decoupled from groups. Objects are always routed by `hash(tenant, key)`; `group_id` becomes an annotation that no longer determines which metadata shard an object lands in. Group membership moves to a separate `group_domain_` domain (sharded by `hash(tenant, group_id)`) that stores the member key list and the group's shared `Lease`.

It fixes two problems in the current (#2180) implementation:

- **Routing coupled to groups**: every key-level operation consulted the `object_group_ids_` index (an extra lookup under a lock) to locate an object.
- **Partial-group eviction fragments groups**: because a member's route was tied to its group, eviction could eject members one at a time and leave half-ejected groups — the bad-cache fragmentation that #2127 reports. Decoupling routing and enforcing all-or-none group eviction removes this fragmentation.

## Changes

- Route object metadata purely by `hash(tenant, key)`; remove the `object_group_ids_` routing index, `group_routing_mutex_`, and `GetGroupRoute`.
- Add `group_domain_` / `GroupState` holding the member key list and the group's shared `Lease`.
- Grouped members share one `Lease` (single-source TTL, extended monotonically on read) instead of each having a per-object lease.
- Read path (`exist` / `get` / batch) refreshes only the object's own lease — identical to the ungrouped path.
- Write/remove paths register/unregister members in `group_domain_`; empty groups are dropped.
- Snapshot/standby restore rebuilds `group_domain_` from object metadata, preserving the group's lease deadline.
- Remove `GrantLeaseForGroup` and `RebuildGroupRoutingIndex`.

Grouped objects are evicted all-or-none at the group level (either the whole group is protected or it is a candidate), based on the shared group TTL. Object routing is decoupled from groups, so members live in different metadata shards: group eviction partitions members by shard and acquires every member shard lock in ascending order without holding the trigger shard lock, re-looking-up and re-validating each member under its own lock (lease, pin, evictable replica). Group protection is derived from member leases at eviction time: any unexpired member skips the whole group.

## Scope

`ReplicateConfig::group_ids` is unchanged, so SGLang needs no change and behavior stays best-effort / backward compatible. No long-lived group metrics are added.

Not included (left as follow-ups, matching the lightweight scope of #2180): NoF / SSD-offload group awareness, group completeness tracking (`group_size` / `group_index`), `RemoveGroup`, and atomic group visibility.

## Module

- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [x] Refactor
- [ ] Performance improvement
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other

## How Has This Been Tested?

Unit tests on a Linux host (`cmake -G Ninja -DWITH_STORE_RUST=OFF -DCMAKE_BUILD_TYPE=Debug`):

```bash
./mooncake-store/tests/master_service_test
./mooncake-store/tests/master_service_group_test
./mooncake-store/tests/master_service_evict_scenario_test
./mooncake-store/tests/master_service_scenario_test
```

Results:
- `master_service_test`: 110 passed.
- `master_service_group_test`: 13 passed.
- `master_service_evict_scenario_test`: 13 passed.
- `master_service_scenario_test`: 32 passed.
- `clang-format-20 --dry-run --Werror` clean on all changed files.

Coverage added:
- Group register/cleanup and rebuild-from-metadata (lease-deadline-preserving restore).
- Route-decoupling invariant: routing is a pure function of `hash(tenant, key)`, identical for grouped and ungrouped objects.
- Read path refreshes only the read member's own lease (no whole-group lease refresh).
- Grouped eviction (all-or-none) with leased / pinned / unsafe / safe members.

End-to-end SGLang HiCache + Mooncake L3 (Qwen3-8B, TP=2, standalone store service, ~80 eviction cycles):

- **Read path, no regression**: TTFT p90 0.16s with group semantics off vs 0.16s with it on; throughput ~22 req/s in both.
- **Group eviction is all-or-none**: 0 members were skipped by safety checks during group eviction (no half-set residue), while groups with a still-valid member lease were correctly skipped (group protection).

## Checklist

- [x] I have formatted my code using `clang-format-20` (via the CI build box)
- [ ] For changes >500 LOC: I have filed an RFC issue (see #2127)

## AI Assistance Disclosure

Implementation and tests were drafted with AI assistance and reviewed by a human; the experiment design follows the #2127 RFC methodology.


Icedcoco:
Thanks for putting this refactor together. The separation between object routing and group membership is clear, and the CI checks are green. Before merging, could we clarify a few semantic points?

1. Snapshot compatibility: In #2180, grouped objects were stored on `hash(group_id)`, while this PR restores each snapshot shard at its recorded index and then only rebuilds `group_shards_`. Could an old snapshot therefore leave grouped metadata on the old shard while lookups now use `hash(tenant, key)`, making those objects unreachable after an upgrade? If old snapshots are expected to remain compatible, could we add a migration/rerouting step and a regression test? Otherwise, should this upgrade limitation be documented?

2. Cross-shard locking: During grouped eviction, the trigger metadata shard remains locked while `GroupHasActiveLease` reads other metadata shards, and the eviction then acquires write locks for those shards. Since quota eviction can be triggered concurrently by different writers, could this create an AB/BA lock-ordering deadlock? Would it be safer to release the trigger-shard lock before the cross-shard scan, or serialize grouped eviction with a consistent lock order?

3. Eviction semantics: The PR description says grouped eviction is “all-or-none”, but the implementation skips pinned, leased, or otherwise unsafe members and continues evicting the remaining members. `GroupedEvictionSkipsUnsafeMembersAndEvictsSafePeers` also appears to test this best-effort behavior. Is the intended contract still the RFC’s best-effort eviction of currently safe members? If so, could we align the PR description and comments with that behavior? If all-or-none is the intended guarantee, should we preflight every member before mutating any of them?

4. Documentation and lease behavior: The read path now refreshes only the accessed object’s lease, while the current documentation still says that reading one member refreshes the group lease. The documentation also still describes grouped objects as being routed by `hash(group_id)`. Could we update these sections and clarify whether the lease change is an intentional compatibility-impacting behavior change?

Clarifying these points would make the upgrade behavior and group semantics much easier to review.

mooncake-store/include/master_service.h
```diff
@@ -1688,6 +1684,27 @@ class MasterService {
...
 
+    // Group domain: sharded by hash(tenant, group_id), storing only the
+    // member key list (no group-level lease). Object routing is decoupled from
+    // groups — objects are always routed by hash(tenant, key); group membership
+    // is a lifecycle constraint consulted at eviction time. Group protection is
+    // derived from member object leases at eviction, keeping the read path free
+    // of any group-table lookup.
+    struct GroupState {
+        TenantId tenant_id;
+        std::string group_id;
```
Aionw: Remove duplicate group_id if possible.

mooncake-store/include/master_service.h
```diff
@@ -1688,6 +1684,27 @@ class MasterService {
...
+        std::unordered_set<std::string> member_keys;
+
+        bool Empty() const { return member_keys.empty(); }
+    };
+
+    struct GroupShard {
+        mutable SharedMutex mutex;
+        // key: tenant_id.MakeScopedKey(group_id)
+        std::unordered_map<std::string, GroupState> groups GUARDED_BY(mutex);
```
Aionw, line 1705: Maybe we do not need shard here.

mooncake-store/include/master_service.h
```diff
@@ -1688,6 +1684,27 @@ class MasterService {
...
+    // Group domain: sharded by hash(tenant, group_id), storing only the
+    // member key list (no group-level lease). Object routing is decoupled from
+    // groups — objects are always routed by hash(tenant, key); group membership
+    // is a lifecycle constraint consulted at eviction time. Group protection is
+    // derived from member object leases at eviction, keeping the read path free
+    // of any group-table lookup.
+    struct GroupState {
+        TenantId tenant_id;
+        std::string group_id;
+        std::unordered_set<std::string> member_keys;
```
Aionw, line 1695: Maybe `std::unordered_map<std::string,  std::shared_ptr<Lease>>` is better here, so that when we do evict, we do not need to recheck if some key is group member.

That said, put will not block by evict since we do not hold group lock.

mooncake-store/include/master_service.h
```diff
@@ -1865,17 +1913,30 @@ class MasterService {
...
                                const std::string& key,
                                const std::string& group_id);
+    // Reads the member keys registered for `group_id`; empty if unregistered.
+    std::vector<std::string> GetGroupMemberKeys(
+        const TenantId& tenant_id, const std::string& group_id) const;
+    // True when any member still holds a valid lease at `now`, so the group was
+    // recently used and must be skipped this eviction round. Members live in
+    // their own metadata shards (hash(tenant, key)); the trigger key's
+    // already-locked tenant state is reused instead of re-locking its shard.
+    bool GroupHasActiveLease(const TenantId& tenant_id,
```
Aionw: Would be better to move this as group member function, not master service member function?

mooncake-store/include/master_service.h
```diff
@@ -1101,8 +1162,14 @@ class MasterService {
...
         // the Clock's epoch (i.e., time_since_epoch() is zero).
+        // Authoritative lease for ungrouped objects only. Grouped objects do
+        // not maintain it; use the group-aware lease methods instead of this
+        // field.
         mutable std::chrono::system_clock::time_point lease_timeout
-            GUARDED_BY(lock);  // hard lease
+            GUARDED_BY(lock);  // hard lease (ungrouped)
+        // Shared group TTL for grouped objects (the authoritative lease); null
+        // if ungrouped. Not persisted; rebuilt from snapshots.
+        mutable std::shared_ptr<GroupLease> group_lease_ GUARDED_BY(lock);
```
Aionw: Single lease, we can using shared_ptr<Lease> here to simplify our code.

mooncake-store/include/master_service.h
```diff
@@ -1321,30 +1388,62 @@ class MasterService {
         // lifetime is intentionally independent from reads.
         void GrantReadLease(const uint64_t ttl) const {
```
Aionw: If we use atomic<shared_ptr>, here grant can acquire without lock. Besideds, input ttl can be typed, that siad, using `std::chrono::milliseconds` or something like that.

mooncake-store/include/master_service.h
```diff
@@ -1688,6 +1783,24 @@ class MasterService {
     };
     std::array<MetadataShard, kNumShards> metadata_shards_;
 
+    // Group domain: all groups in one shard. Routing stays hash(tenant, key);
+    // group state is just member keys + one shared lease, consulted at eviction
+    // (all-or-none) — the read path never touches this table.
+    struct GroupState {
+        std::unordered_set<std::string> member_keys;
```
Aionw, line 1695: No need to account group members.

mooncake-store/include/master_service.h
```diff
@@ -1865,17 +2007,25 @@ class MasterService {
...
+    void UnregisterGroupMember(const TenantId& tenant_id,
                                const std::string& key,
                                const std::string& group_id);
+    // Reads the member keys registered for `group_id`; empty if unregistered.
+    std::vector<std::string> GetGroupMemberKeys(
+        const TenantId& tenant_id, const std::string& group_id) const;
+    // Returns the group's shared GroupLease, or nullptr when the group is not
+    // registered. Eviction consults this single lease instead of scanning every
+    // member's own lease.
+    std::shared_ptr<GroupLease> GetGroupLease(
```
Aionw: `GetGroupLease()` function is enough to handle all of group related features, other functions can be removed

mooncake-store/src/master_service.cpp
```diff
@@ -2218,29 +2206,96 @@ void MasterService::ReleaseLocalDiskUsage(
...
-                tenant_state.group_members[metadata.group_id].insert(key);
-                rebuilt_group_ids[tenant_id.MakeScopedKey(key)] =
-                    metadata.group_id;
-                groups_needing_refresh.insert(
-                    tenant_id.MakeScopedKey(metadata.group_id));
+                // lease_ is mutable, so it can be wired through a const ref;
+                // RegisterGroupMember returns the group's shared lease.
+                auto lease =
+                    RegisterGroupMember(tenant_id, key, metadata.group_id);
+                metadata.lease_ = lease;
```
Aionw, line 2333: [P1] `DeserializeShard` has already restored each object's persisted lease deadline, but this assignment replaces every grouped object with a newly-created `Lease` whose deadline is zero. `RestoreState()` then calls `ApplySnapshotState()`, whose cleanup erases every object for which `IsLeaseExpired(now)` is true, so all grouped objects in a normal snapshot are dropped immediately (including current-format snapshots). Please build the shared lease from the maximum restored member deadline before assigning it, and cover the full cleanup-enabled restore path.

mooncake-store/include/lease.h
```diff
@@ -0,0 +1,78 @@
...
+   public:
+    // Unless the deadline is already over half a TTL out, extend it (max) to
+    // now + ttl. The early-out keeps hot-group reads lock-free.
+    void GrantReadLease(std::chrono::milliseconds ttl) {
+        const auto ttl_ns =
+            std::chrono::duration_cast<std::chrono::nanoseconds>(ttl).count();
+        const int64_t now = NowNs();
+        const int64_t half_ttl_ns = ttl_ns / 2;
+        int64_t cur = deadline_ns_.load(std::memory_order_relaxed);
+        if (cur > now + half_ttl_ns) {
```
Aionw: [P1] Every successful read RPC still advertises the full `default_kv_lease_ttl_`, and the client turns that into `start_time + lease_ttl_ms`. With this early return, a lease that has 750 ms remaining stays at 750 ms after `GrantReadLease(1000 ms)`, while the caller is told it owns a 1000 ms lease. The server can therefore evict before the client-side lease expires. Please always monotonic-max to `now + ttl`, or return the actual remaining TTL.

mooncake-store/src/master_service.cpp
```diff
@@ -10191,59 +10150,84 @@ void MasterService::BatchEvict(double evict_ratio_target,
...
-                try_evict_or_offload(tenant_id, member_key, member_metadata,
-                                     tenant_state, deferred_replicas);
-            result.freed_bytes += freed;
-            if (freed > 0) {
-                result.evicted_objects++;
-                if (!enable_oplog_) {
-                    PublishKvRemovedAfterEvict(member_key, freed, "cpu",
-                                               member_metadata, tenant_id);
-                }
+            MetadataShardAccessorRW other(this, shard_idx);
```
Aionw: [P1] The caller already holds the trigger shard's write lock when it enters this helper. Iterating `members_by_shard` in sorted order does not establish a global order because that pre-held trigger shard can have any index. Concurrent group evictions starting from members on shards A and B (for example, BatchEvict racing quota eviction, or two quota evictions) can therefore deadlock A→B/B→A; the quota path around line 9821 has the same pattern. Please release the trigger shard before grouped eviction, then acquire all member shard locks in canonical order and re-lookup/revalidate the objects.

docs/source/design/store/mooncake-store.md
```diff
@@ -453,13 +453,14 @@ Mooncake Store remains an object-oriented KV cache: objects are still put, queri
...
+- Memory eviction first consults the group's single shared TTL: if the group was read recently it is skipped as a whole; otherwise the group's current members are evicted all-or-none, applying the existing per-object safety checks (hard pins, soft pins when soft-pin eviction is disabled, incomplete writes, busy replicas, unavailable replica states) on a best-effort basis. The number of evicted objects is bounded by the requested eviction ratio.
+- Object removal APIs, copy/move tasks, and NoF eviction keep their existing object-level semantics. Group membership metadata is cleaned up when objects are removed.
+- On snapshot/standby restore, group state is rebuilt from object metadata (fresh group TTLs). Snapshots produced by a router that placed grouped objects on `hash(group_id)` shards are not re-routed to `hash(tenant, key)` after an upgrade and should be regenerated; current-format snapshots restore correctly.
```
Aionw: [P2] This says legacy grouped snapshots are not re-routed, but `MetadataSerializer::Deserialize()` now calls `ReRouteRestoredObjectsByKey()` and the new migration test explicitly verifies that behavior. Please document the automatic migration instead; otherwise operators will be told to regenerate snapshots unnecessarily. (The grouped-lease restore issue above still needs fixing for those migrated objects to survive cleanup.)

Aionw (changes requested):
Requesting changes for three correctness blockers in grouped lease restore and cross-shard eviction, plus the inline documentation issue. One additional build regression cannot be attached inline because the file is unchanged:

- [P2] `BUILD_BENCHMARK=ON` no longer compiles: `mooncake-store/benchmarks/batch_evict_bench.cpp:182` still assigns `metadata.lease_timeout`, which was removed in this PR. A syntax-only build against this head fails with `ObjectMetadata has no member named lease_timeout`; please update the benchmark to the new lease API.

CAICAIIs:
## Performance check: with/without group, before/after this PR (via `master_synthetic_bench`)

I benchmarked this PR against upstream `main` using the store metadata-plane benchmark from
**[#3147](https://github.com/kvcache-ai/Mooncake/pull/3147) `master_synthetic_bench`** (extended with
`--group_size` grouping + per-request `p50/p99/p999` latency reporting). The same bench binary drives both
masters; only the master binary differs. Both are `Debug` builds with the same compiler, so there is no
build-type/optimization confound.

### Benchmark setup
- Plane: client → `mooncake_master` metadata RPC (PutStart/End, Get, Exist).
- Master: `eviction_high_watermark_ratio=0.8`, `eviction_ratio=0.1`, `default_kv_lease_ttl=300`.
- Workload: 2 segments × 4 GiB (8 GiB total), `value_size=1 MiB`, `replica_num=1`, `placement_mode=preferred`,
  `arrival_model=fixed`. With `--group_size=5`, every 5 put keys share one `group_id` (members hash to several
  shards → exercises the all-or-none cross-shard group eviction path).
- Load profiles (per-segment QPS): medium = Exist 15 / Put 15 / Get 15; high = Exist 30 / Put 45 / Get 30.
- Reported cells are means of 2 runs (60 s each), plus a 120 s confirmation run below.

### Main result (group ON, before vs after)

| run | load | p50(ms) | p99(ms) | p999(ms) | evict_keys | rpc_fail |
|---|---|---:|---:|---:|---:|---:|
| upstream/main | medium | 1.720 | 9.421 | 10.956 | 29000 | 1468 |
| upstream/main | high | 1.400 | 8.414 | 10.786 | 99782 | 3374 |
| this PR | medium | 1.278 | 4.218 | 5.278 | 29180 | 1468 |
| this PR | high | 1.162 | 4.122 | 5.475 | 99880 | 3376 |

Delta (this PR vs upstream/main): medium **p99 −55.2% / p999 −51.8%**, high **p99 −51.0% / p999 −49.2%**; p50 −17% to −26%.
Eviction volume (+0.1% to +0.6%) and `rpc_failure_events` (unchanged) are essentially identical, so the latency
gap is not an artifact of different failure/eviction handling.

### group OFF control (120 s, medium load) — isolating whether the win is group-specific

To rule out a benchmark artifact I reran with `--group_size=0` (no grouping) on both masters:

| run | group | p50(ms) | p99(ms) | p999(ms) | rpc_fail | evict_keys |
|---|---|---:|---:|---:|---:|---:|
| upstream/main | ON | 1.385 | 8.704 | 10.461 | 3265 | 64860 |
| this PR | ON | 1.098 | 3.738 | 5.066 | 3270 | 65060 |
| upstream/main | OFF | 1.209 | 5.494 | 8.316 | 3266 | 65219 |
| this PR | OFF | 1.097 | 3.423 | 4.227 | 3268 | 65169 |

With **group OFF** there is still a clear gap (p99 5.49 → 3.42 ms, p999 8.32 → 4.23 ms), so the improvement is
**not purely group-eviction-specific**. This matches the PR's own description: the old design ("routing coupled to
groups") took a global `group_routing_mutex_` + `object_group_ids_` lookup **on every key in every Get/Exist/Put
batch** (verified in upstream `main`, e.g. inside `BatchGetReplicaList`). This PR removes that entirely — routing
becomes pure `hash(tenant, key)`, so each metadata op is a direct shard-map lookup. The group-ON case shows the
largest gap because group eviction additionally contended on that same path.

### Root cause of the per-op lookup

The per-op group-routing lookup was introduced by **PR #2180** ("Add opt-in grouped object routing semantics",
`06ae8dd0`), later extended by #2232 / #2405 / #2508 / #2685. upstream `main` still carries that design, so it is
a fair "before" for this comparison.

mooncake-store/src/master_service.cpp
```diff
@@ -11571,7 +11648,12 @@ MasterService::MetadataSerializer::Deserialize(
 
         const msgpack::object& shard_obj = shard_oh.get();
 
-        // Get shard reference and deserialize
+        // Objects are restored to the shard index recorded in the snapshot (not
```
Aionw: [P3] This comment is now stale: `MetadataSerializer::Deserialize()` calls `ReRouteRestoredObjectsByKey()` below, and the updated documentation states that legacy `hash(group_id)` snapshots are migrated automatically. Please update or remove this block; as written, it incorrectly tells operators that these snapshots become unreachable and must be regenerated.

Aionw (approved):
LGTM

ykwd (approved):

Issue #3395, opened 2026-08-12
[Bug]: 0.3.12.post1 Mass client expiry triggers a full-table scan under shard write locks; entire RPC layer freezes until scan completes


lizipao:
### Bug Report

With mooncake-store master `v0.3.12.post1`, when a client node hosting many objects shuts down, at `client_ttl` seconds after its last ping the master expires all of its clients in one sweep:

```
I0811 19:41:04.965971 master_service.cpp:7061] client_id=5063333486259842164-7836975202054854303, action=client_expired
I0811 19:41:04.966004 master_service.cpp:7061] client_id=9605647728395089542-15849331516245319852, action=client_expired
I0811 19:41:04.966007 master_service.cpp:7061] client_id=14793435909040123795-1308281298534625169, action=client_expired
```

One second before, the master was fully healthy (`role=leader, state=serving, ... Ping=34.20/34.20` in admin metrics). Right after the burst, the RPC layer stops answering: requests time out, Pings are no longer processed, CPU utilization **drops** (one core works; the rest wait). The admin metrics thread keeps printing normally throughout. The process is eventually killed by our platform supervisor as unresponsive.

This is **reproducible on demand**: every shutdown of a cache-heavy client node triggers it. With CP (context parallelism) enabled in our backend it happens essentially 100% of the time; without CP, less often (our key count is several times smaller then). On `0.3.11.post1` we operated for months without a fatal incident of this kind

### Before submitting...

- [x] Ensure you searched for relevant issues and read the [documentation]

ykwd:
Thanks for reporting this issue. It is fixed in this pr: https://github.com/kvcache-ai/Mooncake/pull/2877

ykwd:
One slightly puzzling aspect is that the root cause may be that, during unmount, the master blocks the threads used to serve RPCs, including the threads handling ping requests. As a result, after `client_ttl`, the master may assume that none of the clients have sent a ping and mark all of them as offline.

In version 0.3.11, the default number of threads was 4, while in 0.3.12 the default thread count was increased. However, this issue did not occur in 0.3.11 and only appeared in 0.3.12. That said, I believe we had already increased the number of master threads in our 0.3.12 deployment, so theoretically, a single client unmount should not have been able to cause this issue.

@Aionw Could you take a look at this? Thanks

XFDG:
Closing as already-fixed on `main`.

[...]

PR #3593, opened 2026-08-22
[Store] Sweep stale handles in bounded write-lock batches

pjdurden:
# Issue #3395: mass client expiry freezes the RPC layer

## 1. Root cause

`MasterService::ClearStaleHandles()`
(`mooncake-store/src/master_service.cpp`) is the shard walk behind both
stale-handle sweeps:

* `ClearInvalidHandles()` / `ClearInvalidHandles(alive_clients)` — run by the
  replica cleanup worker, by `UnmountSegment` / `UnmountNoFSegment`, and, most
  importantly here, by the expiry branch of `ClientMonitorFunc()`.
* `ClearLocalDiskHandlesOwnedBy(owner)` — run by `UnmountLocalDiskSegment`.

The walk was structured as:

```cpp
for (size_t i = 0; i < kNumShards; i++) {
    MetadataShardAccessorRW shard(this, i);   // exclusive shard lock
    for (every tenant in the shard)
        for (every key of that tenant)
            ... classify, and clean if stale ...
}
```

The exclusive lock was taken unconditionally for every one of the 1024 shards
and held for the *entire* walk of that shard, whether or not the shard had
anything to clean. Every RPC that touches metadata (`Get`, `Put`, `Remove`,
`ExistKey`, ...) needs the same shard lock, so each of them blocked until the
sweep had finished walking its shard, and the whole table was walked with no
window in between.

That is tolerable on a small table. It is not tolerable in the reported
scenario:

* When a cache-heavy node goes away, `ClientMonitorFunc` expires all of its
  clients in one `client_ttl` tick (with Context Parallelism there are many
  clients per node, which is why the reporter sees it ~100% of the time with CP
  and rarely without). The expiry branch then calls
  `ClearInvalidHandles(alive_clients)`, which sweeps the *whole* metadata
  table.
* Every memory replica of the departed segments now reports
  `has_invalid_mem_handle()`, and with LOCAL_DISK offload every replica of the
  departed owners matches `has_stale_local_disk_client()`, so the sweep does
  not just classify — it also erases, publishes quota releases and (in HA mode)
  appends an OpLog record per key, all still under the shard's write lock.
* The sweep runs on the client monitor thread itself, so while it is running no
  `Ping` is drained from `client_ping_queue_` and no further expiry is
  processed. That is what makes it look like "one core busy, everything else
  blocked, ping stopped": the sweep is single threaded, and the RPC threads are
  queued behind the shard locks it holds.

So the freeze duration scales with the number of objects in the master, not
with the number of expired clients, and it is paid in one uninterruptible
block.

## 2. The fix and why

`ClearStaleHandles()` now walks each shard in two passes:

1. **Pick** — under the shard's *shared* lock, collect the tenant/key pairs
   that either own a replica matching `is_stale` or are already invalid.
   Readers are unaffected by a shared lock, and a shard with no candidate never
   takes the exclusive lock at all. The predicate used here
   (`metadata.HasReplica(is_stale) || !metadata.IsValid()`) is exactly the pair
   of conditions the old loop branched on, so nothing new is selected or
   skipped. Tenants that are already fully empty are collected separately, so
   the sweep still garbage-collects them the way it used to.
2. **Clean** — take the exclusive lock for at most
   `kStaleHandleCleanupBatchSize` (64) picked keys at a time, then release it.
   The lock is dropped and retaken between batches, so RPCs on that shard get a
   turn instead of waiting for the whole shard.

Because the lock is released between the two passes and between batches, pass 2
looks each key up again and re-runs `BuildStaleHandleCleanupPlan()` under the
exclusive lock before touching anything; a key that disappeared or stopped
being stale in the meantime is simply skipped. The HA/OpLog branch, the quota
accounting, the `disk_object_count` accounting and the empty-tenant erase are
all unchanged — only the locking shape changed.

Worst-case total work is the same; what changed is that it is no longer one
uninterruptible exclusive hold per shard. The peak exclusive hold drops from
"one whole shard" to "64 keys", and the common case (nothing stale in the
shard) drops to a shared-lock scan.

## 3. Files changed

* `mooncake-store/src/master_service.cpp` — rewrote
  `MasterService::ClearStaleHandles()` as pick-under-shared-lock plus
  clean-in-bounded-write-lock-batches; added the
  `kStaleHandleCleanupBatchSize` constant to the file's anonymous namespace.
* `mooncake-store/include/master_service.h` — extended the `ClearStaleHandles`
  declaration comment to describe the new locking contract.
* `mooncake-store/tests/master_service_test.cpp` — added
  `MasterServiceTest.ClearInvalidHandlesSweepsShardAcrossLockBatches`, plus the
  `ClearInvalidHandlesForTest` / `MetadataShardIndexForTest` fixture helpers it
  needs.

## 4. Risk and uncertainty

* **Widened race window, bounded.** A liveness-complement sweep
  (`ClearInvalidHandles`) classifies by absence from a point-in-time
  `alive_clients` set, so a client that re-registers mid-sweep can have a fresh
  replica swept. That hazard already existed for every shard the sweep had not
  reached yet (it is called out in the `ClearLocalDiskHandlesOwnedBy` comment
  in `master_service.h`, and is why the owner-targeted sweep exists). This
  change widens it within a shard by the time between pass 1 and the batch that
  covers a key. It does not add a new class of misclassification, and the
  owner-targeted sweep used by `UnmountLocalDiskSegment` is immune to it by
  construction.
* **`UnmountSegment` ordering.** `UnmountSegment` sweeps between
  `PrepareUnmountSegment` and `CommitUnmountSegment`. The replicas it must
  catch are already invalid before the sweep starts (the allocator is dropped
  in Prepare), so both passes see them. Objects whose replicas go stale after
  pass 1 were already missed by the old code once the shard had been passed;
  they are picked up by the periodic `replica_cleanup_worker_` sweep and by the
  lazy cleanup in the `MetadataAccessorRW` constructor.
* **Memory.** Pass 1 materialises the candidate list for one shard (tenant id +
  key string per candidate). It is bounded by the shard's key count — the same
  order of memory as one shard's key set, and only for candidates.
* **Batch size** (64) is a fixed constant, not a flag. It is a latency/overhead
  tradeoff; a smaller value yields shorter stalls at the cost of more
  lock round trips.
* This is the sweep half of the problem. It does not change the fact that the
  sweep runs inline on the client monitor thread, so a very large sweep still
  delays that thread's next tick; it just no longer holds shard locks across
  the whole table while doing it.

## 5. How this was verified

* Reviewed every caller of `ClearStaleHandles` (`ClearInvalidHandles` x2,
  `ClearLocalDiskHandlesOwnedBy`) and confirmed the new code reproduces the old
  branch-for-branch behaviour: stale-plan branch, HA/OpLog branch, invalid-key
  erase branch, and empty-tenant erase.
* Confirmed that reading `ObjectMetadata::HasReplica` / `IsValid` under a
  *shared* shard lock is already an established pattern in this file
  (`GetKeyCount`, `ExistKey` via `MetadataAccessorRO`): replica vectors are
  only mutated under the exclusive shard lock.
* Added a regression test,
  `MasterServiceTest.ClearInvalidHandlesSweepsShardAcrossLockBatches`, that
  forces 100 keys per segment onto a *single* metadata shard (the batch bound
  is per shard, so keys spread over 1024 shards would never cross a batch
  boundary), unmounts one of two segments, runs the sweep, and asserts via
  `GetKeyCount()` that exactly the departed segment's keys were physically
  erased while the live segment's keys still resolve to a memory replica.
  `GetKeyCount()` is the meaningful assertion here: `GetReplicaList` already
  hides replicas of an unmounted segment before any sweep runs, so only the
  physical key count distinguishes "swept" from "hidden".
* `clang-format` (16.0.0, the repo's pre-commit formatter) was run on the three
  touched files; it reported no changes outside the edited regions.
* **Not verified by a build or a test run.** This checkout is a shallow clone
  with empty `extern/` submodules and none of the required dependencies
  (boost, glog, gtest, yalantinglibs) present, so `mooncake-store` cannot be
  configured or compiled here. The new test and the refactor have not been
  executed. They need `mooncake_store_test` (or
  `scripts/run_ci_test.sh`) to be run on a machine with the toolchain before
  merge.

### DIFF
--- a/mooncake-store/include/master_service.h
+++ b/mooncake-store/include/master_service.h
@@ -1064,7 +1064,12 @@ class MasterService {
     // misclassify a concurrent mount, whatever the interleaving.
     void ClearLocalDiskHandlesOwnedBy(const UUID& owner);
     // Shard walk shared by the two sweeps above; removes completed replicas
-    // matching is_stale, erasing a key when no valid replica remains.
+    // matching is_stale, erasing a key when no valid replica remains. Each
+    // shard is first scanned under its shared lock to pick the keys that
+    // match, and only those are cleaned, in bounded batches under the write
+    // lock: a mass client expiry marks handles stale table-wide, and walking
+    // a whole shard while holding it exclusively blocked every RPC for that
+    // shard until the sweep moved on.
     tl::expected<void, ErrorCode> ClearStaleHandles(
         const std::function<bool(const Replica&)>& is_stale);
     bool ProcessClientOffboardingJob(ClientOffboardingJob& job);
--- a/mooncake-store/src/master_service.cpp
+++ b/mooncake-store/src/master_service.cpp
@@ -70,6 +70,13 @@ namespace {
 
 constexpr int kMaxTenantQuotaEvictionRetries = 2;
 
+// Upper bound on the number of keys MasterService::ClearStaleHandles cleans
+// up while holding one metadata shard's write lock. A mass client expiry
+// marks handles stale all over the table, and without a cap the sweep held
+// each shard exclusively for as long as it took to walk that whole shard,
+// so every RPC that touched those keys stalled behind the sweep.
+constexpr size_t kStaleHandleCleanupBatchSize = 64;
+
 // Per-cycle offload cap as a fraction of `offloading_queue_limit_`. Used only
 // when offload-on-evict mode is active. Defers memory eviction for at most
 // this fraction of the queue limit per BatchEvict cycle; beyond that, eviction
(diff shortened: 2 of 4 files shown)


catyans (approved):
This is a nice improvement. Holding the shard write lock across a full stale-handle sweep was always going to hurt once the metadata table grew, and the two-pass approach keeps the critical section much more predictable.

I specifically checked the gap between the shared-lock scan and each write batch. The code does not trust the earlier snapshot: it looks the key up again and rebuilds the cleanup plan under the write lock, while still walking every key selected by the scan. The HA path also keeps the existing durable-finalize behavior. This looks sound to me.

LGTM. The full CTest suite and the other wheel jobs passed. The remaining Ubuntu 22.04 failure is the SSD concurrent-stress test under capacity pressure and does not appear to exercise this cleanup path, but it would still be good to rerun it before merging.

Aionw (approved):
LGTM

ykwd (approved):

PR #3598, opened 2026-08-23
[Bugfix][TENT] Fix RailMonitor topology use-after-free

### LINKED ISSUE #3595: [Bug] TENT RDMA: SIGSEGV in RailMonitor::updateBestMapping() after peer restart (use-after-free of remote Topology)
# [Bug] TENT RDMA: SIGSEGV in RailMonitor::updateBestMapping() after peer restart (use-after-free of remote Topology)

### Describe the bug

In a PD-disaggregated vLLM serving setup (1 Prefill + 1 Decode) using the Mooncake
transfer engine (TENT RDMA transport) for KV cache transfer, the transfer engine on
the **prefill** node crashed with a **segmentation fault** inside
`RailMonitor::updateBestMapping()` shortly after the peer (decode-side) RDMA
endpoints became unreachable. The crash killed the vLLM worker process hosting the
engine, which then cascaded into the whole prefill pipeline hanging (remaining
pipeline-parallel ranks died with `Connection reset by peer`, engine core actor
wedged, requests stuck forever).

Timeline from the engine stderr log (2026-08-22):

1. `19:11:42` — repeated connection failures to the peer endpoints begin:
   ```
   E ... workers.cpp:295] Unable to connect endpoint 192.168.0.243:15510@mlx5_bond_3:
       InvalidArgument: Missing peer GID in bootstrap
       Raised at .../tent/src/transport/rdma/endpoint.cpp:398
   ```
   (peer process on the same host, ports 15510/16645, appears to have restarted)
2. `19:12:47` — retries exhausted, thousands of:
   ```
   W ... workers.cpp:362] Slice 0x... failed: retry count exceeded
   ```
   plus `fallback device selection for slice ...` messages
3. `19:12:59` — **segfault**:
   ```
   !!!!!!! Segfault encountered !!!!!!!
     File ".../libc_sigaction.c", line 0, in 0x000075562a04832f
     File ".../tent/src/transport/rdma/rail_monitor.cpp", line 300, in mooncake::tent::RailMonitor::updateBestMapping()
     File ".../tent/src/transport/rdma/rail_monitor.cpp", line 79, in mooncake::tent::RailMonitor::markFailed(int, int)
     File ".../tent/src/transport/rdma/workers.cpp", line 306, in mooncake::tent::Workers::disableEndpoint(mooncake::tent::RdmaSlice*)
     File ".../tent/src/transport/rdma/workers.cpp", line 364, in mooncake::tent::Workers::asyncPostSend()
     File ".../tent/src/transport/rdma/workers.cpp", line 570, in mooncake::tent::Workers::workerThread(int)
     ...
   ```

### Root cause analysis

`RailMonitor` keeps the peer topology as a **raw pointer**:

```cpp
// tent/include/tent/transport/rdma/rail_monitor.h
const Topology *local_{nullptr};
const Topology *remote_{nullptr};
```

That pointer originates from the segment metadata cache:

```cpp
// tent/src/transport/rdma/workers.cpp, Workers::getRouteHint
hint.topo = &std::get<MemorySegmentDesc>(hint.segment->detail).topology;
```

i.e. it points **inside a cached `SegmentDesc`**, whose lifetime is owned by the
segment manager cache.

The only place that re-validates / reloads this pointer is
`Workers::selectOptimalDevice`:

```cpp
// workers.cpp:720
if (!rail.ready() || target.topo != rail.remote())
    rail.load(source.topo, target.topo, ...);
```

But the failure path — `disableEndpoint()` → `markFailed()` → `updateBestMapping()`
(`workers.cpp:304-306`, `rail_monitor.cpp:79`) — dereferences the cached `remote_`
**without any validity check**:

```cpp
// rail_monitor.cpp:285,293-300
const int remote_nic_count = (int)remote_->getNicCount();
...
auto remote_entry = remote_->getNicEntry(remote_nic);
if (!remote_entry || remote_entry->type != Topology::NIC_RDMA)   // line 300: SIGSEGV here
```

When the peer restarts, the segment cache entry is refreshed/replaced and the old
`Topology` object is freed. The next slice failure then calls
`updateBestMapping()` on a **dangling `remote_`** → use-after-free → SIGSEGV in an
RDMA worker thread, taking down the entire process.

Notes:
- `Topology::getNicEntry()` itself is bounds-checked (`topology.cpp:170`, returns
  `nullptr` for out-of-range ids), and line 300 checks `!remote_entry` first — so
  the crash cannot come from an out-of-range NIC id. The only way line 300 faults
  is if `remote_` itself points to a destroyed object.
- Thread-safety is not the issue here: per `rail_monitor.h` each RDMA worker thread
  owns its own `RailMonitor`.
- There is also an ABA variant of this bug: if the segment cache frees the old
  `SegmentDesc` and a new one happens to be allocated at the same address, the
  `target.topo != rail.remote()` comparison in `selectOptimalDevice` stays false
  and the stale topology content is silently kept.

### Suggested fixes

1. Make `RailMonitor` hold `std::shared_ptr<const Topology>` instead of a raw
   pointer (the topology is already `shared_ptr`-managed elsewhere, e.g.
   `buffers.h:119`). This removes the UAF class entirely.
2. Short-term: re-validate the cached remote topology on the failure path too —
   e.g. in `Workers::disableEndpoint()` compare the current segment topology
   pointer against `rail.remote()` and reload / skip `markFailed` when it has
   changed.

### To Reproduce

1. PD-disaggregated setup, TENT RDMA transport, peer B actively serving RDMA
   endpoints.
2. Restart the peer process (or otherwise make its endpoints disappear — "Missing
   peer GID in bootstrap" appears on side A).
3. Keep sending transfer requests from side A so slices fail with
   `retry count exceeded` and `disableEndpoint()` is invoked.
4. Side A segfaults in `RailMonitor::updateBestMapping()` once the error threshold
   is reached (`markFailed` → `updateBestMapping`) after the segment cache has
   dropped the old peer descriptor.

### Environment

- **Mooncake version**: v0.3.11.post1 (commit `e9c61075720039bcfc5fffd19f847608402be3d0`,
  built from source). Local tree carries diagnostic-logging-only patches to
  `endpoint.cpp` / `workers.cpp`; `rail_monitor.cpp` / `rail_monitor.h` /
  `topology.cpp` are **unmodified upstream code**.
- **GPU**: 8 × NVIDIA H800 275GB (SXM), driver 580.126.20
- **CUDA**: 13.0 (nvcc V13.0.88)
- **OS**: Ubuntu 24.04.3 LTS, kernel 6.8.0-106-generic
- **RDMA**: Mellanox mlx5 devices exposed as `mlx5_bond_0..7` (RoCE, bonded),
  rdma-core 50.0-2ubuntu0.2 (`libibverbs1`, `ibverbs-providers`)
- **Deployment**: Kubernetes (KubeRay), hostNetwork pods; vLLM (v1, multiproc
  executor, PP>1) with Mooncake as the KV transfer engine
- **Config**: defaults for `rail_error_threshold` (3), `rail_error_window_secs`
  (10), `rail_cooldown_secs` (30)

### Additional context

The segfault is fatal to the hosting vLLM worker process; in our case all 4
pipeline-parallel prefill workers exited (one from the segfault, the rest from
gloo `Connection reset by peer` during `isend_tensor_dict`) while the engine core
actor stayed alive but wedged — the pod looked "Running" but prefill served no
requests for ~6 hours until manual intervention. A crash in a background RDMA
worker thread therefore has service-wide blast radius; beyond fixing the UAF, it
may be worth considering installing a fatal-signal handler that escalates to a
controlled engine restart instead of silently wedging.

lxy-alexander:
## Description

Fixes #3595.

`RailMonitor` previously stored raw pointers to the local and remote
`Topology` objects:

```cpp
const Topology *local_{nullptr};
const Topology *remote_{nullptr};
```

These objects belong to cached Transfer Engine segment metadata. When the peer
restarted and its metadata was refreshed, the old metadata could be released
while `RailMonitor` still referenced its topology. The RDMA failure path could
then dereference a dangling pointer in `updateBestMapping()` and cause a
`SIGSEGV`.

This PR changes `RailMonitor` to retain shared ownership of both topologies:

```cpp
std::shared_ptr<const Topology> local_;
std::shared_ptr<const Topology> remote_;
```

The worker creates aliasing `shared_ptr` references whose ownership is tied to
the corresponding pinned segment metadata:

```cpp
rail.load(std::shared_ptr<const Topology>(source.pin, source.topo),
          std::shared_ptr<const Topology>(target.pin, target.topo),
          rail_topo_json_, transport_->conf_.get());
```

This keeps the topology objects alive for as long as `RailMonitor` may access
them, including during RDMA failure handling. The reload condition also checks
both the local and remote topology pointers before reusing the existing rail
mapping.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

A targeted lifecycle regression test,
`RailMonitorLifetimeTest.KeepsTopologiesAliveForFailureUpdates`, was added to
`mooncake-transfer-engine/tent/tests/rail_monitor_test.cpp`.

The test reproduces the lifetime condition behind the reported failure by:

1. Creating local and remote `Topology` objects.
2. Loading them into a `RailMonitor`.
3. Releasing all external `shared_ptr` references to both topologies.
4. Using `weak_ptr` references to verify that `RailMonitor` keeps the
   topologies alive.
5. Calling `markFailed()` three times to reach the default error threshold and
   exercise the failure-driven `updateBestMapping()` path.
6. Verifying that the failed rail becomes unavailable.
7. Destroying `RailMonitor` and verifying that both topology objects are then
   released.

The test covers the exact `RailMonitor` topology-lifetime invariant and the
`markFailed()` to `updateBestMapping()` path involved in the reported
use-after-free. It does not reproduce a complete multi-node peer restart or
exercise the full RDMA endpoint and metadata-cache refresh flow; those require
an integration environment with RDMA-capable peers.

**Test commands:**

```bash
cmake -S . -B build \
  -DUSE_TENT=ON \
  -DBUILD_UNIT_TESTS=ON \
  -DWITH_STORE=OFF \
  -DWITH_STORE_RUST=OFF

cmake --build build \
  --target tent_rail_monitor_test \
  -j4

./build/mooncake-transfer-engine/tent/tests/tent_rail_monitor_test
```

**Test results:**

```text
[==========] 8 tests from 4 test suites ran. (1500 ms total)
[  PASSED  ] 8 tests.
```

- [x] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

AI tools assisted with root-cause analysis, implementation, regression-test preparation, and verification. The human submitter reviewed and understood every changed line.

### DIFF
--- a/mooncake-transfer-engine/tent/src/transport/rdma/workers.cpp
+++ b/mooncake-transfer-engine/tent/src/transport/rdma/workers.cpp
@@ -934,8 +934,9 @@ Status Workers::selectOptimalDevice(RouteHint& source, RouteHint& target,
 
     auto& rail = getOrCreateRail(worker.rails, target.segment->machine_id);
     if (!rail.ready() || target.topo != rail.remote())
-        rail.load(source.topo, target.topo, rail_topo_json_,
-                  transport_->conf_.get());
+        rail.load(std::shared_ptr<const Topology>(source.pin, source.topo),
+                  std::shared_ptr<const Topology>(target.pin, target.topo),
+                  rail_topo_json_, transport_->conf_.get());
     if (slice->target_dev_id < 0) {
         int mapped_dev_id = rail.findBestRemoteDevice(
             slice->source_dev_id, target.topo_entry->numa_node);
--- a/mooncake-transfer-engine/tent/tests/rail_monitor_test.cpp
+++ b/mooncake-transfer-engine/tent/tests/rail_monitor_test.cpp
@@ -18,8 +18,8 @@
 #include <thread>
 
 #include "tent/common/config.h"
+#include "tent/runtime/segment.h"
 #include "tent/transport/rdma/rail_monitor.h"
-#include "tent/runtime/topology.h"
 
 namespace mooncake {
 namespace tent {
@@ -56,6 +56,31 @@ static std::shared_ptr<Topology> makeSingleNicTopology(const std::string& nic,
         ADD_FAILURE() << "Topology::parse failed: " << status.ToString();
     }
     return topo;
+}
+
+static SegmentDescRef makeSingleNicSegment(const std::string& nic,
+                                           int numa_node = 0) {
+    auto desc = std::make_shared<SegmentDesc>();
+    desc->type = SegmentType::Memory;
+    auto& topology = std::get<MemorySegmentDesc>(desc->detail).topology;
+    auto json_str = R"({
+        "nics": [{"name": ")" +
+                    nic + R"(", "type": 0, "numa_node": )" +
+                    std::to_string(numa_node) + R"(}],
+        "mems": [{
+            "name": "cuda0",
+            "type": 1,
+            "numa_node": )" +
+                    std::to_string(numa_node) +
+                    R"(,
+            "device_list": {"rank0": [0]}
+        }]
+    })";
+    auto status = topology.parse(json_str);
+    if (!status.ok()) {
+        ADD_FAILURE() << "Topology::parse failed: " << status.ToString();
+    }
+    return desc;
 }
 
 static std::shared_ptr<Topology> makeTwoNicTopology(const std::string& first,
@@ -97,7 +122,7 @@ TEST(RailMonitorConfigTest, CustomJsonOverridesAutomaticPeerMapping) {
     })";
 
     RailMonitor rail;
-    ASSERT_TRUE(rail.load(local.get(), remote.get(), rail_json, nullptr).ok());
+    ASSERT_TRUE(rail.load(local, remote, rail_json, nullptr).ok());
     EXPECT_EQ(rail.findBestRemoteDevice(/*local_nic=*/0, /*remote_numa=*/0), 1);
     EXPECT_EQ(rail.findBestRemoteDevice(/*local_nic=*/1, /*remote_numa=*/0), 0);
     EXPECT_TRUE(rail.available(/*local_nic=*/0, /*remote_nic=*/1));
@@ -155,7 +180,7 @@ TEST(RailMonitorCrossNumaTest, CrossNumaPrefersSameNameDevice) {
     // remote NUMA-0 domain: mlx5_y (idx0); remote NUMA-1 domain: mlx5_x (idx1).
     auto remote = makeNamedNumaTopology("mlx5_y", 0, "mlx5_x", 1);
     RailMonitor rail;
-    ASSERT_TRUE(rail.load(local.get(), remote.get()).ok());
+    ASSERT_TRUE(rail.load(local, remote).ok());
     ASSERT_TRUE(rail.ready());
 
     // local mlx5_y (idx1, NUMA 1) reaching the remote NUMA-0 domain: the only
@@ -175,7 +200,7 @@ TEST(RailMonitorRecoverTest, RecoverResetsErrorCount) {
     auto local = makeSingleNicTopology("mlx5_0");
     auto remote = makeSingleNicTopology("mlx5_1");
     RailMonitor rail;
-    ASSERT_TRUE(rail.load(local.get(), remote.get()).ok());
+    ASSERT_TRUE(rail.load(local, remote).ok());
     ASSERT_TRUE(rail.ready());
 
     // Initially available
@@ -202,7 +227,7 @@ TEST(RailMonitorRecoverTest, RecoverUnpausesPausedRail) {
     auto local = makeSingleNicTopology("mlx5_0");
     auto remote = makeSingleNicTopology("mlx5_1");
     RailMonitor rail;
-    ASSERT_TRUE(rail.load(local.get(), remote.get()).ok());
+    ASSERT_TRUE(rail.load(local, remote).ok());
 
     // Drive error_count to the default threshold (3) to trigger pause
     for (int i = 0; i < 3; ++i) rail.markFailed(0, 0);
@@ -224,10 +249,39 @@ TEST(RailMonitorRecoverTest, RecoverUnknownPairIsNoop) {
     auto local = makeSingleNicTopology("mlx5_0");
     auto remote = makeSingleNicTopology("mlx5_1");
     RailMonitor rail;
-    ASSERT_TRUE(rail.load(local.get(), remote.get()).ok());
+    ASSERT_TRUE(rail.load(local, remote).ok());
 
     // NIC IDs 5 and 7 are not in the topology — must not crash
     EXPECT_NO_FATAL_FAILURE(rail.markRecovered(5, 7));
+}
+
+TEST(RailMonitorLifetimeTest, KeepsSegmentSnapshotsAliveForFailureUpdates) {
+    auto local = makeSingleNicSegment("mlx5_0");
+    auto remote = makeSingleNicSegment("mlx5_1");
+    std::weak_ptr<SegmentDesc> weak_local = local;
+    std::weak_ptr<SegmentDesc> weak_remote = remote;
+    auto* local_topology = &local->getMemory().topology;
+    auto* remote_topology = &remote->getMemory().topology;
+
+    {
+        RailMonitor rail;
+        ASSERT_TRUE(
+            rail.load(std::shared_ptr<const Topology>(local, local_topology),
+                      std::shared_ptr<const Topology>(remote, remote_topology))
+                .ok());
+        local.reset();
+        remote.reset();
+
+        EXPECT_FALSE(weak_local.expired());
+        EXPECT_FALSE(weak_remote.expired());
+        EXPECT_NO_FATAL_FAILURE({
+            for (int i = 0; i < 3; ++i) rail.markFailed(0, 0);
+        });
+        EXPECT_FALSE(rail.available(0, 0));
+    }
+
+    EXPECT_TRUE(weak_local.expired());
+    EXPECT_TRUE(weak_remote.expired());
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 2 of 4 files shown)


mooncake-transfer-engine/tent/src/transport/rdma/workers.cpp
```diff
@@ -933,9 +933,11 @@ Status Workers::selectOptimalDevice(RouteHint& source, RouteHint& target,
             "No device could access the slice memory region" LOC_MARK);
 
     auto& rail = getOrCreateRail(worker.rails, target.segment->machine_id);
-    if (!rail.ready() || target.topo != rail.remote())
-        rail.load(source.topo, target.topo, rail_topo_json_,
-                  transport_->conf_.get());
+    if (!rail.ready() || source.topo != rail.local() ||
```
staryxchen: `source.topo != rail.local()` compares snapshot identity, not NIC topology content. Local `SegmentDesc` is copy-on-write, so any `updateLocal()` (including a duplicate buffer registration that only bumps `ref_count`) publishes a new snapshot whose `Topology` member has a new address even when the NIC set is unchanged.

`RailMonitor::load()` then rebuilds mappings and clears `rail_states_` / `direct_rails_`. A rail that was paused or in cooldown can become selectable again on the next `selectOptimalDevice` after an unrelated local buffer register.

Holding the aliasing `shared_ptr` is enough to keep the old local topology alive for `markFailed()` → `updateBestMapping()`. Reloading on every local pointer change is not required to fix the UAF.

Suggestion: refresh the stored pins when only the snapshot identity changed, and call `load()` only when the NIC topology actually changed (name / count / NUMA).

mooncake-transfer-engine/tent/tests/rail_monitor_test.cpp
```diff
@@ -224,12 +224,36 @@ TEST(RailMonitorRecoverTest, RecoverUnknownPairIsNoop) {
...
+        EXPECT_FALSE(weak_remote.expired());
+        EXPECT_NO_FATAL_FAILURE({
+            for (int i = 0; i < 3; ++i) rail.markFailed(0, 0);
+        });
+        EXPECT_FALSE(rail.available(0, 0));
+    }
+
+    EXPECT_TRUE(weak_local.expired());
+    EXPECT_TRUE(weak_remote.expired());
+}
```
staryxchen: This test shows `RailMonitor` keeps a `shared_ptr<Topology>` alive, but that is not the production ownership model.

`selectOptimalDevice` uses the aliasing constructor:

```cpp
std::shared_ptr<const Topology>(source.pin, source.topo)
```

`Topology` is a member of the cached `SegmentDesc`, not a separately owned object. The reported SIGSEGV is: the segment cache drops the old snapshot while `RailMonitor` still uses Topology inside it.

This test would still pass if the call site used `std::shared_ptr<const Topology>(source.topo)`, which compiles and would delete an embedded `Topology`. It also does not cover an empty `pin` (the aliasing constructor then stores the raw pointer with `use_count() == 0` and does not keep the object alive).

Please add a regression that:

1. Puts `Topology` inside a `SegmentDesc` / `SegmentDescRef`
2. Loads via `shared_ptr<const Topology>(pin, &desc->topology)`
3. Drops the external `SegmentDescRef`
4. Calls `markFailed()` through the threshold so `updateBestMapping()` runs

mooncake-transfer-engine/tent/src/transport/rdma/workers.cpp
lxy-alexander: Thank you @staryxchen  for your review, this is a good catch. 

I will remove the newly added `source.topo != rail.local()` reload condition
and retain the existing remote snapshot check in this UAF fix to avoid changing
the established remote-refresh behavior.

mooncake-transfer-engine/tent/tests/rail_monitor_test.cpp
lxy-alexander: Thank you @staryxchen for the clarification and review. I rewrote the regression test to mirror the actual RouteHint ownership relationship base on your review:
- Each Topology is embedded in a SegmentDesc.
- A SegmentDescRef owns each segment snapshot.
- RailMonitor::load() receives aliasing shared_ptr instances:
```
std::shared_ptr<const Topology>(local, local_topology)
std::shared_ptr<const Topology>(remote, remote_topology)
```
- The external SegmentDescRef instances are then reset.
- The test verifies that both segment snapshots remain alive while RailMonitor exists.
- It calls markFailed() three times to reach the default threshold and force updateBestMapping() to access the -embedded topologies.
- Finally, it verifies that the snapshots are released after RailMonitor is destroyed.

Before the fix:
```
rail_monitor_test.cpp:275: Failure
Value of: weak_local.expired()
  Actual: true
Expected: false
RailMonitor must retain the local segment snapshot

rail_monitor_test.cpp:277: Failure
Value of: weak_remote.expired()
  Actual: true
Expected: false
RailMonitor must retain the remote segment snapshot

[==========] 8 tests from 4 test suites ran. (1500 ms total)
[  PASSED  ] 7 tests.
[  FAILED  ] 1 test.
```
After the fix:
```
[==========] 8 tests from 4 test suites ran. (1500 ms total)
[  PASSED  ] 8 tests.
```
I have not added a separate empty-pin test yet. Both source.pin and target.pin are populated by successful calls to:
```
withCachedSegment(segment_id, hint.pin, ...)
```
The local path assigns the current local snapshot, while the remote path obtains an owning reference through getRemoteCached(). Therefore, an empty pin should not be reachable when selectOptimalDevice() is entered successfully. I do not know if I understand correctly. Thank you again.

staryxchen (approved):
Thanks, LGTM.

PR #3599, opened 2026-08-23
[Store] Add exist_key_grant_lease flag to opt out of leasing on existence probes

duhaode520:
## Description

`ExistKey` / `BatchExistKey` currently grant a read lease on **every hit** (`master_service.cpp:2780` and `:2851`). A read lease pins the object so it cannot be evicted until the lease TTL expires (default 10s), and `BatchEvict` skips any leased object outright:

```cpp
// master_service.cpp:10250
if (!it->second.IsLeaseExpired(now) || !has_evictable) continue;
```

This is fine when existence checks track actual reads one-to-one. **But for read-heavy workloads the probe rate and coverage vastly exceed the read rate, and leasing on probes starves eviction.** The concrete trigger is vLLM's KV-cache-offload connector (`MooncakeStoreConnector`, in the vLLM tree at `vllm/distributed/kv_transfer/kv_connector/v1/mooncake/store/`):

- On every request, the scheduler calls `client.lookup(...)` (`connector.py` → `scheduler.py`) to decide how much prefix it can reuse.
- `lookup` expands **all** prefix-candidate block hashes — further multiplied across each rank namespace — into one big `candidate_keys` list and issues a single `store.batch_is_exist(candidate_keys)` (`worker.py:1944`), which lands as `BatchExistKey` on the master.
- It then reads back **only the longest matching prefix** (`worker.py:1964-1990`). Every other hit — probed but not read this request — still got a fresh 10s lease.

With a small memory tier under steady-state eviction, the same hot blocks are re-probed faster than TTL/2, so their leases never expire. Nearly the entire working set stays pinned, `BatchEvict` finds no evictable candidates, and new `Put`s fail with `NO_AVAILABLE_HANDLE`. This was observed in production as **Eviction Success/Attempts = 39/1123** with puts rejected (`-200`).

**Fix:** add a master config flag `exist_key_grant_lease` (default `true`, i.e. upstream behavior unchanged), wired through the existing gflag/config path exactly like `promotion_on_hit`. When set to `false`, `ExistKey`/`BatchExistKey` no longer grant leases — intended for read-heavy + frequent-eviction deployments (e.g. vLLM KV offload on a small memory tier).

**Why this is safe:** actual reads are unaffected — `GetReplicaList` grants its own lease (`master_service.cpp:3678` / `3865`) to protect the read-in-flight window. The only behavioral change when disabled is that a probe→read race may evict an entry in between, which for the offload connector is just a normal cache miss (the tokens are recomputed locally). Correctness is preserved either way. This is an opt-in, backward-compatible flag with no `USE_PHOENIX` dependency.

## Module

## Type of Change

- [x] Bug fix
- [x] Performance improvement

## How Has This Been Tested?

Added three unit tests in `mooncake-store/tests/master_service_test.cpp`:

- `ExistKeyGrantsLeaseByDefaultBlockingEviction` — default (`true`): an `ExistKey` hit leases the object, so it survives `RunBatchEvictForTesting(1.0, 1.0)`.
- `ExistKeyDoesNotGrantLeaseWhenDisabled` — `false`: the probed object is evicted under full pressure; a subsequently re-put object read via `GetReplicaList` still survives (confirming real reads remain protected).
- `BatchExistKeyDoesNotGrantLeaseWhenDisabled` — `false`: same for the batch path.

- [x] Unit tests pass

## Checklist

- [x] I have formatted my code using `./scripts/code_format.sh` (verified with clang-format 20.1.0)
- [x] For changes >500 LOC: N/A (~130 LOC)

## AI Assistance Disclosure

> AI assistance was used to trace the root cause across the Mooncake and vLLM codebases, draft the config wiring (following the existing `promotion_on_hit` pattern), and write the tests. The human submitter has reviewed every changed line and can defend the change end-to-end.

### DIFF
--- a/docs/source/deployment/mooncake-store-deployment-guide.md
+++ b/docs/source/deployment/mooncake-store-deployment-guide.md
@@ -598,6 +598,7 @@ mooncake_master \
 | `--eviction_ratio` | `0.05` | Fraction evicted at high watermark |
 | `--eviction_high_watermark_ratio` | `0.90` | Usage ratio triggering eviction |
 | `--client_ttl` | `10` s | Seconds before a silent client is considered disconnected |
+| `--exist_key_grant_lease` | `true` | Grant a read lease on `ExistKey`/`BatchExistKey` hits. Set `false` if these leases block eviction |
 
 ### Tenant Quota
 
--- a/docs/source/deployment/ssd/ssd-offload.md
+++ b/docs/source/deployment/ssd/ssd-offload.md
@@ -122,6 +122,8 @@ These flags control when the master asks clients to persist objects to SSD, whet
 
 Start with `--enable_offload=true` for eager SSD persistence. Add `--offload_on_evict=true` when SSD writes should happen only under memory pressure. For SSD-heavy workloads where offload-on-evict is dropping too many objects, raise both `--offloading_queue_limit` and `--offload_cap_ratio`; for example, `--offloading_queue_limit=500000 --offload_cap_ratio=0.8` allows up to `400000` objects to be queued in one eviction cycle.
 
+With `--offload_on_evict=true`, eviction is what moves objects to SSD, and read leases block eviction. If existence checks are used as speculative scans, the leases they grant can keep DRAM full and stall SSD writes; set `--exist_key_grant_lease=false` on the master in that case.
+
 ---
 
 ## SSD Offload Configuration
--- a/mooncake-store/include/master_config.h
+++ b/mooncake-store/include/master_config.h
@@ -141,6 +141,13 @@ struct MasterConfig {
     // rich clusters may safely raise it.
     uint32_t promotion_max_per_heartbeat = 1;
 
+    // Grant a read lease on ExistKey/BatchExistKey hits (upstream default).
+    // Set false for read-heavy workloads (e.g. vLLM KV-offload lookup) where
+    // high-rate existence probes would otherwise pin the working set and
+    // starve eviction. Actual reads stay protected: GetReplicaList grants its
+    // own lease.
+    bool exist_key_grant_lease = true;
+
     // Dynamic MEMORY replica fanout for hot read-only objects.
     // Kept intentionally small: mode + frequency window + max replicas.
     std::string dynamic_replication_mode = "off";
@@ -249,6 +256,7 @@ class MasterServiceSupervisorConfig {
     uint32_t promotion_admission_threshold = 2;
     uint32_t promotion_queue_limit = 50000;
     uint32_t promotion_max_per_heartbeat = 1;
+    bool exist_key_grant_lease = true;
     std::string dynamic_replication_mode = "off";
     uint32_t dynamic_replication_heat_window_seconds = 10;
     double dynamic_replication_admission_qps_threshold = 0.8;
@@ -309,6 +317,7 @@ class MasterServiceSupervisorConfig {
         promotion_admission_threshold = config.promotion_admission_threshold;
         promotion_queue_limit = config.promotion_queue_limit;
         promotion_max_per_heartbeat = config.promotion_max_per_heartbeat;
+        exist_key_grant_lease = config.exist_key_grant_lease;
         dynamic_replication_mode = config.dynamic_replication_mode;
         dynamic_replication_heat_window_seconds =
             config.dynamic_replication_heat_window_seconds;
@@ -505,6 +514,7 @@ class WrappedMasterServiceConfig {
     uint32_t promotion_admission_threshold = 2;
     uint32_t promotion_queue_limit = 50000;
     uint32_t promotion_max_per_heartbeat = 1;
+    bool exist_key_grant_lease = true;
     std::string dynamic_replication_mode = "off";
     uint32_t dynamic_replication_heat_window_seconds = 10;
     double dynamic_replication_admission_qps_threshold = 0.8;
@@ -599,6 +609,7 @@ class WrappedMasterServiceConfig {
         promotion_admission_threshold = config.promotion_admission_threshold;
         promotion_queue_limit = config.promotion_queue_limit;
         promotion_max_per_heartbeat = config.promotion_max_per_heartbeat;
+        exist_key_grant_lease = config.exist_key_grant_lease;
         dynamic_replication_mode = config.dynamic_replication_mode;
         dynamic_replication_heat_window_seconds =
             config.dynamic_replication_heat_window_seconds;
@@ -723,6 +734,7 @@ class WrappedMasterServiceConfig {
         promotion_admission_threshold = config.promotion_admission_threshold;
         promotion_queue_limit = config.promotion_queue_limit;
         promotion_max_per_heartbeat = config.promotion_max_per_heartbeat;
+        exist_key_grant_lease = config.exist_key_grant_lease;
         dynamic_replication_mode = config.dynamic_replication_mode;
         dynamic_replication_heat_window_seconds =
             config.dynamic_replication_heat_window_seconds;
@@ -796,6 +808,7 @@ class MasterServiceConfigBuilder {
     uint64_t max_kv_soft_pin_ttl_ = DEFAULT_MAX_KV_SOFT_PIN_TTL_MS;
     bool allow_evict_soft_pinned_objects_ =
         DEFAULT_ALLOW_EVICT_SOFT_PINNED_OBJECTS;
+    bool exist_key_grant_lease_ = true;
     double eviction_ratio_ = DEFAULT_EVICTION_RATIO;
     double eviction_high_watermark_ratio_ =
         DEFAULT_EVICTION_HIGH_WATERMARK_RATIO;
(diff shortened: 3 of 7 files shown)


ykwd:
Thanks for raising this issue.

Regarding the statement, “Every other hit — probed but not read this request — still got a fresh 10s lease,” could you clarify what “other hit” refers to here?

Does it mean a prefix block that already exists in HBM but is not read in this request, or does it mean a suffix block that does not exist in HBM and therefore cannot be reused because there is a gap earlier in the prefix?

If it is the former, I think the lease should probably be preserved, since the suffix inherently depends on the prefix being available. If it is the latter, then after this prefill completes, that gap should be filled anyway.

duhaode520:
> Thanks for raising this issue.
> 
> Regarding the statement, “Every other hit — probed but not read this request — still got a fresh 10s lease,” could you clarify what “other hit” refers to here?
> 
> Does it mean a prefix block that already exists in HBM but is not read in this request, or does it mean a suffix block that does not exist in HBM and therefore cannot be reused because there is a gap earlier in the prefix?
> 
> If it is the former, I think the lease should probably be preserved, since the suffix inherently depends on the prefix being available. If it is the latter, then after this prefill completes, that gap should be filled anyway.

@ykwd Thanks for reviewing— that statement in the description was inaccurate.

With vLLM's `MooncakeConnector`, a suffix behind a gap cannot be a cache hit, so that case does not really apply here. For an existing prefix that is probed but not read in this request, I agree that preserving the lease is the right default behavior.

Our deployment is a slightly different case: a relatively small DRAM tier backed by a large SSD tier with `offload_on_evict`, where DRAM eviction is actually the demotion path to SSD. Under high concurrency, frequent existence probes continuously refresh leases, `BatchEvict` skips those objects, and eventually new `Put`s fail with `NO_AVAILABLE_HANDLE` 

So I think a better scope is to disable probe leases only while `need_mem_eviction_` is true. That preserves the current behavior when there is enough memory, while avoiding probe-induced eviction starvation under actual memory pressure.

What do you think? Happy to rework the PR that way.

ykwd:
I understand the background of the issue now, and I agree that this is indeed something we should address. I’m not yet sure what the best solution would be, though. Let us loop in the SSD collaborators as well to get their input.

fcczzz (commented):
Thanks for working on this issue. After looking into the scenario, I think there are two separate problems here.

The immediate problem addressed by this PR is that speculative `ExistKey` calls, such as prefix scans, may create leases for many objects that are never actually read, which then prevents eviction.

My concern with the current solution is that, when `exist_key_grant_lease` is disabled, the lease semantics of `ExistKey` depend on the current memory pressure: it may grant a lease when there is sufficient memory, but stop granting one under eviction pressure. This makes the contract difficult for callers to reason about.

Would it be clearer to keep the existing `ExistKey` semantics unchanged and introduce an explicit non-leasing probe API, for example `ProbeKey` / `BatchProbeKey`? Its contract would be a point-in-time existence check: the object may be evicted before a subsequent `Get`. Callers such as the vLLM prefix scan could use this API, while actual reads would still acquire a lease through `GetReplicaList`.

Separately, this case also exposes a broader admission issue: `Put` currently needs to be admitted into MEMORY before it can be offloaded to SSD. If MEMORY is occupied by genuinely hot or leased objects, removing speculative leases still cannot make the remaining SSD capacity available for new writes. Supporting a memory-preferred, SSD-fallback admission path may need a separate RFC. I don’t think that larger issue needs to be solved in this PR, but it would be useful to clarify that it is outside the scope of this fix.

duhaode520:
> Thanks for working on this issue. After looking into the scenario, I think there are two separate problems here.
> 
> The immediate problem addressed by this PR is that speculative `ExistKey` calls, such as prefix scans, may create leases for many objects that are never actually read, which then prevents eviction.
> 
> My concern with the current solution is that, when `exist_key_grant_lease` is disabled, the lease semantics of `ExistKey` depend on the current memory pressure: it may grant a lease when there is sufficient memory, but stop granting one under eviction pressure. This makes the contract difficult for callers to reason about.
> 
> Would it be clearer to keep the existing `ExistKey` semantics unchanged and introduce an explicit non-leasing probe API, for example `ProbeKey` / `BatchProbeKey`? Its contract would be a point-in-time existence check: the object may be evicted before a subsequent `Get`. Callers such as the vLLM prefix scan could use this API, while actual reads would still acquire a lease through `GetReplicaList`.
> 
> Separately, this case also exposes a broader admission issue: `Put` currently needs to be admitted into MEMORY before it can be offloaded to SSD. If MEMORY is occupied by genuinely hot or leased objects, removing speculative leases still cannot make the remaining SSD capacity available for new writes. Supporting a memory-preferred, SSD-fallback admission path may need a separate RFC. I don’t think that larger issue needs to be solved in this PR, but it would be useful to clarify that it is outside the scope of this fix.

@fcczzz Thanks for the review. I agree that `ProbeKey` / `BatchProbeKey` gives a cleaner API contract, and we can follow that up in a separate PR.

I see it as a different direction from `exist_key_grant_lease`: `ProbeKey` requires upstream callers to adopt a new RPC, while this option is mainly for existing deployments or third-party connectors that cannot be changed immediately.

`exist_key_grant_lease` stays enabled by default, so disabling it is an explicit operational choice and changes the `ExistKey` lease behavior intentionally.

I also agree that making the behavior depend on memory pressure may be confusing. If so, I’m fine reverting to the simpler first version: when `exist_key_grant_lease=false`, `ExistKey` never grants or refreshes leases.

Also, I agree that the MEMORY-to-SSD admission issue is broader and can stay out of scope for this PR.

fcczzz:
> > Thanks for working on this issue. After looking into the scenario, I think there are two separate problems here.
> > The immediate problem addressed by this PR is that speculative `ExistKey` calls, such as prefix scans, may create leases for many objects that are never actually read, which then prevents eviction.
> > My concern with the current solution is that, when `exist_key_grant_lease` is disabled, the lease semantics of `ExistKey` depend on the current memory pressure: it may grant a lease when there is sufficient memory, but stop granting one under eviction pressure. This makes the contract difficult for callers to reason about.
> > Would it be clearer to keep the existing `ExistKey` semantics unchanged and introduce an explicit non-leasing probe API, for example `ProbeKey` / `BatchProbeKey`? Its contract would be a point-in-time existence check: the object may be evicted before a subsequent `Get`. Callers such as the vLLM prefix scan could use this API, while actual reads would still acquire a lease through `GetReplicaList`.
> > Separately, this case also exposes a broader admission issue: `Put` currently needs to be admitted into MEMORY before it can be offloaded to SSD. If MEMORY is occupied by genuinely hot or leased objects, removing speculative leases still cannot make the remaining SSD capacity available for new writes. Supporting a memory-preferred, SSD-fallback admission path may need a separate RFC. I don’t think that larger issue needs to be solved in this PR, but it would be useful to clarify that it is outside the scope of this fix.
> 
> @fcczzz Thanks for the review. I agree that `ProbeKey` / `BatchProbeKey` gives a cleaner API contract, and we can follow that up in a separate PR.
> 
> I see it as a different direction from `exist_key_grant_lease`: `ProbeKey` requires upstream callers to adopt a new RPC, while this option is mainly for existing deployments or third-party connectors that cannot be changed immediately.
> 
> `exist_key_grant_lease` stays enabled by default, so disabling it is an explicit operational choice and changes the `ExistKey` lease behavior intentionally.
> 
> I also agree that making the behavior depend on memory pressure may be confusing. If so, I’m fine reverting to the simpler first version: when `exist_key_grant_lease=false`, `ExistKey` never grants or refreshes leases.
> 
> Also, I agree that the MEMORY-to-SSD admission issue is broader and can stay out of scope for this PR.

Thanks, this distinction makes sense to me. I agree that the config can serve as a compatibility escape hatch for existing deployments or third-party connectors that cannot adopt a new RPC immediately, while `ProbeKey` / `BatchProbeKey` would be the cleaner long-term API.

Between the current pressure-dependent behavior and the original implementation, I would prefer the original, simpler behavior: when `exist_key_grant_lease=false`, `ExistKey` should consistently avoid granting or refreshing leases. That gives the option a stable contract, and operators enabling it explicitly accept the probe-to-read eviction race.

It would also be helpful to document that this is a global Master setting and therefore affects every client using that Master, especially in shared deployments.

Could we create or link a follow-up issue for `ProbeKey` / `BatchProbeKey` so that the caller-controlled API does not get lost? I agree that the MEMORY-to-SSD admission policy should be discussed separately in an RFC.

duhaode520:
@fcczzz Thanks for the suggestions. I’ve reverted the implementation to the simpler behavior: when exist_key_grant_lease=false, ExistKey does not grant or refresh leases, regardless of memory pressure. I also added documentation to clarify the semantics and intended use of this option.

I’ve also added an RFC for the ProbeKey / BatchProbeKey direction, which I can follow up on separately.

If there are no further concerns, I think this PR should be ready to merge.

fcczzz:
Thanks for the update. The consistent behavior when `exist_key_grant_lease=false` addresses my main concern, and #3769 looks like a good longer-term direction.

Could you please rebase onto `main`? The PR currently has conflicts in the `ExistKey` implementation and tests. After the rebase and CI pass, this looks reasonable to me as a compatibility escape hatch.

duhaode520:
Thanks @fcczzz for reviewing and suggestions. All conflicts have been resolved. 
@ykwd When you have a chance, could you please review the latest patch and check if it looks good?

fcczzz (approved):
LGTM

ykwd:
Seems like this is duplicate with #3801

PR #3600, opened 2026-08-23
[WIP][CI] Enable TENT runtime in T-One e2e as a parallel job

staryxchen:
## Description

Two changes: (1) enable the TENT (Transfer Engine NEXT) runtime in the T-One e2e pipeline as the first step towards TENT productionization, **without reducing classic-engine coverage**, and (2) a source fix required to build TENT wheels on older glibc, discovered by the first `use-tent` CI build.

**1. T-One e2e dual-job enablement** — `e2e-ci.yml` now runs two T-One jobs in parallel: the existing classic job (unchanged) and a new TENT job that activates the runtime via `MC_USE_TENT=1`.

- `_build-wheel.yaml` gains a `use-tent` input (default `false`); `ci.yml`'s `build-wheel-cu13` (the artifact fed to T-One) opts in. One dual-engine wheel serves both jobs: without `MC_USE_TENT` the facade runs classic, with it TENT activates. **Release wheels are unaffected** (only the PR CI cu13 build opts in).
- The TENT job is selected via `USE_TENT=true` in `env_info` (same mechanism as `ARTIFACT_ID`) with a distinct T-One job name (`mooncake-ci-tent-<SHA>`); `e2e-gate` aggregates both.
- `tone_tests/scripts/common.sh` injects container-level `MC_USE_TENT=1` when `USE_TENT=true` and **fails fast** if the installed engine lacks TENT — preventing a silent classic fallback that would make the job pass without testing TENT.

**Explicitly noted behavior change for the classic job**: its wheel is now *compiled* with TENT enabled (previously without). Runtime behavior is unchanged — the classic facade is the default when `MC_USE_TENT` is unset — but the binary carries dormant TENT code.

**2. TENT wheel build fix for older glibc** — the first `use-tent` build on the manylinux2_28 runner (RHEL8-era glibc) failed because `mnnvl_transport.cpp` uses `SYS_pidfd_open`/`SYS_pidfd_getfd` without including `<sys/syscall.h>` or defining fallbacks. The fix adds the explicit include plus the arch-independent asm-generic syscall numbers (434/438) as fallbacks. Dev machines with newer glibc never see this — a portability bug the new CI coverage caught on its first run.

Validation performed before this PR (2×8×H20, sglang 0.5.17 PD
disaggregation, single-node and cross-node):

**GSM8K accuracy (200 questions, 5-shot completion, PD vs no-PD baseline):**

| Model | PD topology (TENT) | PD score | Baseline (no PD) |
|---|---|---|---|
| Qwen3-8B | single-node, P=TP2 / D=TP1 | 0.950 | 0.950 |
| Qwen3-8B | single-node, P=TP1 / D=TP2 | 0.950 | 0.950 |
| DeepSeek-V4-Flash | cross-node, TP4 / TP4 | 0.970 | 0.965 |

**Cross-node performance, classic vs TENT (Qwen3-8B, 24.5k-token transfers concurrent burst):**

| TP (P/D) | serial large, classic / TENT | burst p50/p99, classic / TENT | burst tput, classic / TENT |
|---|---|---|---|
| 1/1 | 3.91 / 3.92 s | 0.79/0.95 · 0.73/0.96 s | 5,180 / 5,410 tok/s |
| 2/2 | 2.09 / 2.12 s | 0.73/0.80 · 0.73/0.98 s | 11,958 / 10,771 tok/s |
| 4/4 | 1.14 / 1.15 s | 0.46/0.54 · 0.48/0.53 s | 18,420 / 17,983 tok/s |

DeepSeek-V4-Flash cross-node TP4/TP4: 1.96 / 2.00 s serial large, 4,330 /
4,228 tok/s burst (classic / TENT). All differences within run variance.

**Dynamic transport selection (TENT's core capability, no app-side
changes):**

| Scenario | Route chosen | Evidence |
|---|---|---|
| intra-node GPU↔GPU | NVLink | 99.999% of KV bytes (12h soak) |
| cross-node | RDMA | 100% of bytes (13.85 GB benchmark window) |

**Stability**: 12h single-node soak — 172,584/172,584 requests (100.000%),
22 TB KV transferred via TENT, 0 transfer failures / 0 failovers, flat GPU
memory. TENT also builds clean on current main including the
scikit-build-core build changes (full rebuild 55s, engine init smoke rc=0).

No existing issue/PR covers this work (searched MC_USE_TENT / T-One / tone_tests).

## Module

- [x] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
# Local: full TENT rebuild on updated main
cmake -G Ninja .. -DCMAKE_BUILD_TYPE=Release -DUSE_TENT=ON -DUSE_HTTP=ON \
  -DUSE_ETCD=OFF -DUSE_CUDA=ON -DUSE_INTRA_NVLINK=ON -DWITH_STORE=OFF \
  -DWITH_EP=OFF -DTENT_METRICS_ENABLED=ON && ninja

# Local: TENT activation smoke (expect TENT runtime log lines, rc=0)
MC_USE_TENT=1 python -c "
e = TransferEngine()
print(e.initialize('127.0.0.1', 'P2PHANDSHAKE', 'rdma', ''))"

# Shell/YAML syntax of changed files
bash -n scripts/tone_tests/scripts/common.sh
python -c "import yaml; [yaml.safe_load(open(f'.github/workflows/{f}')) for f in
  ['_build-wheel.yaml','ci.yml','e2e-ci.yml','integration-test.yml']]"
```

**Test results:**
- [ ] Unit tests pass
- [x] Integration tests pass (12h TENT PD soak + GSM8K accuracy parity +
      cross-node classic/TENT benchmark parity; see Description)
- [x] The glibc fix was validated by the CI itself: the same
      `build-wheel-cu13 / build (3.12)` job that failed on manylinux2_28
      before the fix is the verification target of the fix.

E2E verification on T-One itself is expected from the first labeled PR run
(`run-e2e-ci`): the TENT job log should show `TENT runtime verified.` and
TENT startup lines in container sglang logs. The classic-leg T-One job on
this PR (run via the base-branch workflow, i.e. TENT-compiled wheel with
classic runtime) is also part of the verification: it exercises exactly the
post-merge classic-job behavior.

## Checklist

Note on pre-commit: `pre-commit run --files` passes for the four workflow
files; for `scripts/tone_tests/scripts/common.sh` the `trailing-whitespace`
hook rewrites ~99 *pre-existing* lines (the file predates the hook's
enforcement; CI does not enforce it on shell files). Per the lean-PR policy
those unrelated whitespace-only edits are intentionally left out of this PR.

## AI Assistance Disclosure

- [ ] No AI tools were used

Tencent CodeBuddy (GLM) assisted with: workflow/shell patch authoring, local
TENT build and cross-machine validation (functional/soak/accuracy/benchmark),
and PR description drafting. The human submitter has reviewed and can defend
every changed line.


Aionw (changes requested):
`USE_TENT=true` is only present in the T-One controller shell. In the double-machine path, `run_test.sh` copies `run/.shrc` to the remote node and then SSHes there to source `.shrc` followed by `common.sh`; however, the generated `.shrc` does not export `USE_TENT`. As a result, the remote node defaults it to `false` in `common.sh`, so its container is launched without `MC_USE_TENT=1`.

This makes the 1P1D, EPD, and vLLM double-machine cases run with a TENT/classic mix or silently lose remote-side TENT coverage. Please persist/export `USE_TENT` in the generated `.shrc`, and include it in `is_base_env_prepared`'s reuse check so a cached environment cannot retain the wrong runtime selection.

Also, the current E2E run (32634455804) was created for `ee07ab8d` and uses the base-branch `pull_request_target` workflow; its job list contains no `integration-test-tent`. Please rerun the dual job against the current head after fixing the propagation issue.

staryxchen:
Refresh PR head after rebase onto main + notify-mr + GDR debug (f1790dba).

catyans:
The previous propagation concern is addressed on the current head: `USE_TENT` is exported in `.shrc`, copied to the remote node, and included in the cached-environment identity, so both nodes use the same runtime selection. This looks good; I would wait for the dual T-One run and remaining CI before final approval.

PR #3602, opened 2026-08-23
[Bugfix][Store] Graceful shutdown for standalone mooncake_client on SIGTERM

### LINKED ISSUE #3551: [Bug]: standalone `mooncake_client` does not gracefully drain mounted segments on `SIGTERM`

XFDG:
## Affected code paths
- mooncake-store/src/real_client.cpp
  - ResourceTracker::ResourceTracker()  (~line 590): constructor
  - startSignalThread()                (~line 663): sigwait thread entry
  - cleanupAllResources()              (~line 631): tear-down path

## Root cause
Two problems in standalone mooncake_client signal handling:

1. SIGTERM bypasses cleanup (zombie process):
   pthread_sigmask was only called inside the signal thread's lambda, so
   only that thread had SIGTERM blocked. Other threads (coroutine
   threads, RPC threads, etc.) did NOT inherit the mask and could receive
   the signal directly, terminating the process before cleanup ran.

2. Immediate unmount with no grace period:
   cleanupAllResources() called tearDownAll() immediately, unmounting
   all segments with grace_period_seconds=0. Peers that had already
   fetched segment info would get errors on subsequent get/put calls.

## The fix
1. Move SIGTERM/SIGINT/SIGHUP blocking into ResourceTracker's constructor,
   which runs in the main thread before any application threads spawn. All
   subsequently created threads inherit the mask, so process-directed
   signals are only delivered to the sigwait thread.
2. Read the MC_STORE_TERM_GRACE_SECONDS env var and sleep() for that many
   seconds before calling tearDownAll(). Default is 0 (immediate unmount,
   preserving existing behaviour when the variable is not set).

## How it has been tested
- Binary strings verified in compiled .so:
  - "MC_STORE_TERM_GRACE_SECONDS" present
  - "SIGTERM grace period: waiting " present
  - "Received signal" present

Fixes kvcache-ai/Mooncake#3551

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
```

**Test results:**
- [x] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

### DIFF
--- a/mooncake-store/src/real_client.cpp
+++ b/mooncake-store/src/real_client.cpp
@@ -589,12 +589,27 @@ ResourceTracker &ResourceTracker::getInstance() {
 
 ResourceTracker::ResourceTracker() {
     // In embedded environments (e.g. Python), the host runtime owns signal
-    // handling.  Blocking SIGINT with pthread_sigmask (done inside
-    // startSignalThread) would prevent Python from raising KeyboardInterrupt,
-    // causing the process to hang on Ctrl-C.  Detect Python at runtime via
-    // dlsym so we don't need to include <Python.h> or change any public API.
+    // handling.  Blocking SIGINT with pthread_sigmask would prevent Python
+    // from raising KeyboardInterrupt, causing the process to hang on Ctrl-C.
+    // Detect Python at runtime via dlsym so we don't need to include
+    // <Python.h> or change any public API.
     if (!dlsym(RTLD_DEFAULT, "Py_IsInitialized")) {
         // Standalone C/C++ process – install our own signal handling.
+
+        // Block SIGTERM/SIGINT/SIGHUP in the main thread BEFORE any
+        // application threads are spawned.  All subsequently created threads
+        // (ylt coroutine threads, RPC threads, etc.) inherit this mask, so
+        // process-directed signals are only delivered to the sigwait thread.
+        // Previously this masking was done inside the signal thread lambda,
+        // which only masked the signal thread itself — leaving other threads
+        // able to receive SIGTERM directly and terminate before cleanup runs.
+        sigset_t block_set;
+        sigemptyset(&block_set);
+        sigaddset(&block_set, SIGINT);
+        sigaddset(&block_set, SIGTERM);
+        sigaddset(&block_set, SIGHUP);
+        pthread_sigmask(SIG_BLOCK, &block_set, nullptr);
+
         startSignalThread();
     }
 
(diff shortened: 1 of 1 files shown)


ykwd:
Seems like there is a PR doing similar things: https://github.com/kvcache-ai/Mooncake/pull/3552

PR #3604, opened 2026-08-23
[Bugfix][TE] Evict stale endpoints on RDMA port recovery

### LINKED ISSUE #3596: [tent] RDMA QPs are never rebuilt after a port flap: post-recovery transfers run ~5x slower than during the outage, with persistent 20s timeouts
# [tent] RDMA QPs are never rebuilt after a port flap: post-recovery transfers run ~5x slower than during the outage, with persistent 20s timeouts

## Summary

In the TENT transfer engine, `IBV_EVENT_PORT_ERR` only pauses the device
(`RdmaContext::pause()`) and `IBV_EVENT_PORT_ACTIVE` only resumes it
(`RdmaContext::resume()`). QPs that entered an error state while the port was
down are **never destroyed or rebuilt**. After the port comes back, slice
selection immediately includes the flapped rail again, so every striped batch
lands some slices on a dead QP. The net effect is counter-intuitive and severe:
**transfers after link recovery are ~5x slower than during the actual outage**,
and batches keep hitting the 20s sync-transfer timeout indefinitely (we
observed this still ongoing 25+ minutes after the link recovered, until
restart).

## Environment

- `mooncake-transfer-engine-cuda-peermem0 == 0.3.11.post1` (TENT, `MC_USE_TENT=1`)
- 8x RoCE rails, Linux bonding 802.3ad (LACP) over Mellanox mlx5 LAG devices
  (`mlx5_bond_1..8`, GID index 3, RoCEv2)
- vLLM P/D disaggregation: KV blocks striped across all 8 rails via
  `batch_transfer_sync_write`

## Timeline of a single link-flap test (all times UTC, 2026-08-23)

| Time | Event |
|---|---|
| 01:57:58 | `ip link set <bond> down` (admin-down; LACP slaves keep carrier, so no driver/verbs event yet — data path unaffected) |
| 02:04:27 | Carrier finally lost (switch-side LACP timeout) → driver logs `Link DOWN` → TENT logs `Received context async event port error` / `Action: mlx5_bond_3 down` within **milliseconds** of the kernel event |
| 02:04:31+ | RailMonitor kicks in, per-slice `fallback device selection` around the dead rail. Zero failed transfers, zero WC errors, zero retries |
| 02:17:56 | `ip link set <bond> up` |
| 02:17:57 | `async event port active` / `Action: mlx5_bond_3 up` (~1s after admin-up). Fallback logging stops immediately — the rail rejoins slice selection |
| 02:18 → still ongoing | `Sending to <peer> failed (ret=-1) after ~20s` warnings from the vLLM connector, ~2–20 per minute, **never stopping** |

## Measured transfer latency per period (`batchTransferSync completed`)

| Period | n | avg | max |
|---|---|---|---|
| Before flap | 204 | **0.27s** | 1.7s |
| Admin-down silent window (carrier still up) | 144 | 0.28s | 0.9s |
| Real degradation (7 rails, fallback active) | 209 | 1.03s | 5.6s |
| **After link recovery** | 53 | **5.48s** | 11.2s |

Post-recovery is 5x worse than the 7-rail degraded period — running on "8
rails" is much slower than correctly running on 7, because ~1/8 of slices in
every batch are placed on the flapped rail's poisoned QP and stall until
timeout.

Additional symptoms after "recovery":

- `submit blocked` diagnostics with `pending` in the thousands that never
  drains (`wr_count=0`) on multiple endpoints
- No WC errors, no CQ errors, no retries logged anywhere — WRs are posted
  into QPs that simply never complete
- NIC hardware error counters (`local_ack_timeout_err`, `req_cqe_error`,
  `packet_seq_err`) completely static
- Upstream effect: KV push from prefill times out at 20s, prefill GPUs idle
  at 0–10%, requests fail with TTFT timeout at ~200/min

## Root cause

`mooncake-transfer-engine/tent/src/transport/rdma/workers.cpp`,
`Workers::handleContextEvents` (~line 799):

```cpp
} else if (event.event_type == IBV_EVENT_DEVICE_FATAL ||
           event.event_type == IBV_EVENT_PORT_ERR) {
    context->pause();                       // only flips a status flag
    LOG(WARNING) << "Action: " << context->name() << " down";
} else if (event.event_type == IBV_EVENT_PORT_ACTIVE) {
    context->resume();                      // only flips the flag back
    LOG(WARNING) << "Action: " << context->name() << " up";
}
```

`RdmaContext::pause()/resume()` (`context.cpp` ~line 494) are plain
`DeviceStatus` CAS operations gating device selection. Nothing in the
recovery path:

- walks the endpoint store to find QPs that transitioned to `IBV_QPS_ERR`
  (or whose peer did) while the port was down,
- tears those endpoints/QPs down,
- or rebuilds them on `PORT_ACTIVE`.

Only `IBV_EVENT_QP_FATAL`/`WQ_FATAL` evict an endpoint from the store, and
verbs does not necessarily raise per-QP fatal events in this scenario (we saw
none), so the broken endpoints stay in the store and keep being selected.

## Expected behavior

On `IBV_EVENT_PORT_ACTIVE` (and/or on detecting QP-level errors), TENT should
evict and asynchronously rebuild the affected endpoints/QPs before re-enabling
the device for slice selection — or keep the device paused until its endpoints
have been re-established. Fallback during the outage already works well
(1.03s avg vs 0.27s baseline); the missing piece is strictly the rejoin path.

## Reproduction

1. Two TENT peers with N>1 RoCE rails, steady `batch_transfer_sync_write`
   traffic striped across all rails.
2. Down one rail at the physical/link layer (e.g. shut the switch port, or
   `ip link set <slave> down` so carrier is actually lost — note that
   admin-down of a LACP bond master alone does **not** drop carrier or
   generate verbs events).
3. Wait for `Action: <dev> down`, observe fallback working (no failures).
4. Restore the link, wait for `Action: <dev> up`.
5. Observe: batch latencies jump well above the degraded-period baseline and
   sync transfers start timing out permanently.

## Workaround

Restart one side of the transfer engine (forces full QP re-establishment).
Blacklisting the flapped device post-recovery would likely also work but
requires a restart to apply.

## Note on detection asymmetry (FYI, separate concern)

`PORT_ERR` only reached verbs when carrier was actually lost (in our LACP
setup, ~6.5 min after the bond admin-down), while `PORT_ACTIVE` arrived ~1s
after admin-up. TENT reacted within milliseconds in both directions — the
detection delay lives in the kernel/driver/bonding layer, not in TENT. Worth
documenting for people designing link-flap tests: admin-down of a LACP bond
master does not test the RDMA failure path.

XFDG:
## Affected code paths
- mooncake-transfer-engine/tent/include/tent/transport/rdma/context.h
  - RdmaContext::resume(): now calls evictEndpoints()
  - new RdmaContext::evictEndpoints() accessor
- mooncake-transfer-engine/tent/src/transport/rdma/context.cpp
  - resume(): flips status then evicts all endpoints
  - evictEndpoints(): delegates to endpoint_store_->evictAll()
- mooncake-transfer-engine/tent/include/tent/transport/rdma/endpoint_store.h
  - EndpointStore: new pure virtual evictAll()
- mooncake-transfer-engine/tent/src/transport/rdma/endpoint_store.cpp
  - FIFOEndpointStore::evictAll()
  - SIEVEEndpointStore::evictAll()

## Root cause
When an RDMA port goes down and comes back up, the
IBV_EVENT_PORT_ACTIVE path only called context->resume(), which flips a
DeviceStatus flag. QPs that entered IBV_QPS_ERR while the link was down
were never destroyed or rebuilt. The endpoint store kept returning these
stale endpoints, so work requests were posted into dead QPs and never
completed, causing transfers to stall.

Counter-intuitive effect: post-recovery transfers run ~5x SLOWER than
during the outage (5.48s vs 1.03s in the reporter's 7-rail setup), with
persistent 20s timeouts.

## The fix
Add EndpointStore::evictAll() to both FIFOEndpointStore and
SIEVEEndpointStore. evictAll() moves every entry from the active map to
the waiting list and calls beginDestroy() on each, transitioning QPs to
ERR state so the hardware flushes inflight WRs to the CQ.

RdmaContext::resume() now calls evictEndpoints() after successfully
flipping status back to DEVICE_ENABLED. The next getOrInsert() call for
each peer creates a fresh endpoint with a new QP, restoring full
throughput.

## How it has been tested
- git diff reviewed: 4 files, 60 insertions, 1 deletion.
- getOrInsert() already handles the transition: a looked-up endpoint in
  EP_DESTROYING/EP_DESTROYED is removed and a fresh one is created.

Fixes kvcache-ai/Mooncake#3596

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
```

**Test results:**
- [x] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

### DIFF
--- a/mooncake-transfer-engine/tent/src/transport/rdma/endpoint_store.cpp
+++ b/mooncake-transfer-engine/tent/src/transport/rdma/endpoint_store.cpp
@@ -119,6 +119,24 @@ void FIFOEndpointStore::evictOne() {
     LOG(INFO) << victim << " evicted from FIFOEndpointStore";
 }
 
+void FIFOEndpointStore::evictAll() {
+    std::vector<std::shared_ptr<RdmaEndPoint>> to_evict;
+    {
+        RWSpinlock::WriteGuard guard(endpoint_map_lock_);
+        to_evict.reserve(endpoint_map_.size());
+        for (auto& entry : endpoint_map_) to_evict.push_back(entry.second);
+        endpoint_map_.clear();
+        fifo_list_.clear();
+        fifo_map_.clear();
+    }
+    for (auto& ep : to_evict) {
+        waiting_list_.insert(ep);
+        ep->beginDestroy();
+    }
+    LOG(INFO) << "FIFOEndpointStore: evicted all " << to_evict.size()
+              << " endpoints on port recovery";
+}
+
 void FIFOEndpointStore::reclaim() {
     std::vector<std::shared_ptr<RdmaEndPoint>> candidates;
     {
@@ -298,6 +316,26 @@ void SIEVEEndpointStore::evictOne() {
     endpoint_map_.erase(victim);
     LOG(INFO) << "Endpoint " << victim << " has been evicted";
     return;
+}
+
+void SIEVEEndpointStore::evictAll() {
+    std::vector<std::shared_ptr<RdmaEndPoint>> to_evict;
+    {
+        RWSpinlock::WriteGuard guard(endpoint_map_lock_);
+        to_evict.reserve(endpoint_map_.size());
+        for (auto& entry : endpoint_map_) to_evict.push_back(entry.second.first);
+        endpoint_map_.clear();
+        fifo_list_.clear();
+        fifo_map_.clear();
+        hand_ = std::nullopt;
+    }
+    for (auto& ep : to_evict) {
+        waiting_list_.insert(ep);
+        waiting_list_len_.fetch_add(1, std::memory_order_relaxed);
+        ep->beginDestroy();
+    }
+    LOG(INFO) << "SIEVEEndpointStore: evicted all " << to_evict.size()
+              << " endpoints on port recovery";
 }
 
 void SIEVEEndpointStore::reclaim() {
--- a/mooncake-transfer-engine/tent/src/transport/rdma/endpoint_store.cpp
+++ b/mooncake-transfer-engine/tent/src/transport/rdma/endpoint_store.cpp
@@ -323,7 +323,9 @@ void SIEVEEndpointStore::evictAll() {
     {
         RWSpinlock::WriteGuard guard(endpoint_map_lock_);
         to_evict.reserve(endpoint_map_.size());
-        for (auto& entry : endpoint_map_) to_evict.push_back(entry.second.first);
+        for (auto& entry : endpoint_map_) {
+            to_evict.push_back(entry.second.first);
+        }
         endpoint_map_.clear();
         fifo_list_.clear();
         fifo_map_.clear();
--- a/mooncake-transfer-engine/tent/src/transport/rdma/endpoint_store.cpp
+++ b/mooncake-transfer-engine/tent/src/transport/rdma/endpoint_store.cpp
@@ -124,13 +124,20 @@ void FIFOEndpointStore::evictAll() {
     {
         RWSpinlock::WriteGuard guard(endpoint_map_lock_);
         to_evict.reserve(endpoint_map_.size());
-        for (auto& entry : endpoint_map_) to_evict.push_back(entry.second);
+        for (auto& entry : endpoint_map_) {
+            to_evict.push_back(entry.second);
+            // Publish into waiting_list_ under the same lock: reclaim()
+            // iterates waiting_list_ under endpoint_map_lock_, so inserting
+            // here avoids an unsynchronized read/write on the unordered_set.
+            waiting_list_.insert(entry.second);
+        }
         endpoint_map_.clear();
         fifo_list_.clear();
         fifo_map_.clear();
     }
+    // beginDestroy() may call back into reclaim() (which takes
+    // endpoint_map_lock_), so it must run after the lock is released.
     for (auto& ep : to_evict) {
-        waiting_list_.insert(ep);
         ep->beginDestroy();
     }
     LOG(INFO) << "FIFOEndpointStore: evicted all " << to_evict.size()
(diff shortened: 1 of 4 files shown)


mooncake-transfer-engine/tent/src/transport/rdma/endpoint_store.cpp
```diff
@@ -119,6 +119,24 @@ void FIFOEndpointStore::evictOne() {
...
+        to_evict.reserve(endpoint_map_.size());
+        for (auto& entry : endpoint_map_) to_evict.push_back(entry.second);
+        endpoint_map_.clear();
+        fifo_list_.clear();
+        fifo_map_.clear();
+    }
+    for (auto& ep : to_evict) {
+        waiting_list_.insert(ep);
+        ep->beginDestroy();
+    }
```
alogfans, line 142: waiting_list_ is normally accessed under endpoint_map_lock_, and reclaim() can iterate it concurrently with port recovery. evictAll() releases that lock before inserting the evicted endpoints, introducing an unsynchronized read/write on the unordered_set. Please publish the endpoints into waiting_list_ while holding the write lock; beginDestroy() can still run afterwards if calling endpoint methods under the store lock must be avoided.

XFDG: Fixed in 60dcdb31. `evictAll()` now inserts the evicted endpoints into `waiting_list_` while holding the `endpoint_map_lock_` write lock, before clearing the map. The `beginDestroy()` calls still run after the lock is released, since they may call back into `reclaim()` which re-acquires the lock. This eliminates the unsynchronized read/write on the `unordered_set`. Both `FIFOEndpointStore` and `SIEVEEndpointStore` were fixed.

alogfans (approved):
LGTM

PR #3606, opened 2026-08-23
[Store] Avoid direct-I/O padding for POSIX restores

zupengwang:
## Description

`FileStorage::AllocateBatch` currently reserves
`align_up(data_size, 4096) + 8192` bytes and aligns every destination pointer,
even when `use_uring=false` (the default). That extra space is required by the
io_uring/O_DIRECT bucket-read path, which rounds unaligned file ranges out to
full pages, but POSIX reads write exactly `slice.size` bytes.

This change:

- allocates only the object payload (or one byte for an empty object) for POSIX
  restores;
- preserves the existing alignment and tail space when `use_uring=true`;
- makes the pinned-arena fallback test size the arena relative to its payload,
  instead of relying on the old O_DIRECT padding to exhaust it; and
- adds a regression test showing that sixteen 4 KiB objects fit in a 64 KiB
  POSIX restore buffer.

No public API or configuration changes are involved.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [x] Performance improvement
- [ ] Other

## Resource impact

The table uses the repository's `OffsetAllocator::normalizedAllocationSize`
and a 64 MiB arena. It includes allocator size-class rounding, not just the
requested-byte formula.

| Object size | Before / object | After / object | Objects before | Objects after |
| ---: | ---: | ---: | ---: | ---: |
| 4 KiB | 12 KiB | 4 KiB | 5,461 | 16,384 |
| 128 KiB | 144 KiB | 128 KiB | 455 | 512 |
| 1 MiB | 1,152 KiB | 1,024 KiB | 56 | 64 |

The 1 MiB case gains 14.3% effective restore-buffer capacity because the extra
8 KiB pushed the old request into the next allocator size class. The 4 KiB
case gains approximately 3x capacity.

## How Has This Been Tested?

Base commit: `3b5a594133958449a1feaf1c9e0fc1b553e34a28`

The new regression test was first run against the base implementation and
failed with `allocate_result == false` (`BUFFER_OVERFLOW`). It passes after the
change.

**POSIX build and full related suite:**

```bash
cmake --build build-posix --target file_storage_test -j8
LD_LIBRARY_PATH="$PWD/.local/lib${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}" \
  build-posix/mooncake-store/tests/file_storage_test --gtest_color=no
```

Result: 28/28 tests passed.

**io_uring/O_DIRECT compatibility build:**

Built with liburing 2.15 (`io_uring: Enabled for mooncake_store`), then ran:

```bash
cmake --build build-uring --target file_storage_test -j8
MOONCAKE_OFFLOAD_USE_URING=true \
LD_LIBRARY_PATH="$PWD/.local/lib${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}" \
  build-uring/mooncake-store/tests/file_storage_test \
  --gtest_filter=FileStorageTest.BatchLoadRecordsSsdMetrics:FileStorageTest.AllocateBatchAvoidsDirectIoPaddingForPosixReads \
  --gtest_color=no
```

Result: 2/2 tests passed. The run logged `O_DIRECT mode enabled` for the bucket
files and restored 100 non-page-aligned objects totaling 21,000 bytes.

**Formatting:**

```bash
PATH="$PWD/.format-venv/bin:$PATH" \
  ./scripts/code_format.sh --staged --check
```

Result: clang-format 20.1.8 reported no changes.

The host could not register the 1.25 GiB client buffer as an io_uring fixed
buffer (`EFAULT`) and used the implementation's documented regular-io_uring
fallback. O_DIRECT itself was active and the data-integrity checks passed. No
GPU or RDMA hardware was required for these storage-path tests.

- [x] Unit tests pass
- [x] Integration testing performed for POSIX and O_DIRECT bucket reads
- [x] Manual resource-capacity measurement performed

## Checklist

- [ ] I have run the complete pre-commit hook set
- [x] Documentation is not applicable (no user-facing API or configuration change)
- [ ] For changes >500 LOC: I have filed an RFC issue (not applicable; 53 insertions)

## AI Assistance Disclosure

- [ ] No AI tools were used
- [x] AI tools were used

AI assistance was used for code-path inspection and initial test drafting. The
diff was reviewed, formatted, and validated with the commands above.

### DIFF
--- a/mooncake-store/src/file_storage.cpp
+++ b/mooncake-store/src/file_storage.cpp
@@ -1,5 +1,6 @@
 #include "file_storage.h"
 
+#include <algorithm>
 #include <memory>
 #include <optional>
 #include <utility>
@@ -1101,13 +1102,14 @@ FileStorage::AllocateBatch(const std::vector<std::string>& keys,
             return tl::make_unexpected(ErrorCode::INVALID_PARAMS);
         }
 
-        // Allocate oversized buffer for O_DIRECT alignment:
-        //   +4096 for aligning the ptr to 4096 boundary
-        //   +4096 for aligned read tail padding (actual_offset may not be
-        //   aligned)
         size_t data_size = static_cast<size_t>(sizes[i]);
-        size_t alloc_size =
-            align_up(data_size, kDirectIOAlignment) + 2 * kDirectIOAlignment;
+        size_t alloc_size = std::max<size_t>(data_size, 1);
+        if (config_.use_uring) {
+            // O_DIRECT reads need room to align the pointer and to round an
+            // unaligned file range out to full pages.
+            alloc_size = align_up(data_size, kDirectIOAlignment) +
+                         2 * kDirectIOAlignment;
+        }
 
         auto alloc_result = allocator.allocate(alloc_size);
         if (!alloc_result && !gc_triggered &&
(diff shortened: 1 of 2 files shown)


fcczzz (commented):
The allocation split itself looks correct: POSIX reads only need the payload, while the existing alignment and tail space remain sufficient for O_DIRECT. However, the PR is now 12 commits behind main, and rebasing reproducibly conflicts in file_storage.cpp after the config extraction. Please rebase, keep only the new <algorithm> include, and rerun the POSIX and io_uring tests. The updated fixture on main already clears the io_uring environment variables, which also resolves the outstanding pinned-arena test concern.

zupengwang:
Thanks — I rebased onto current main at `9bf84fee` and force-pushed `2b2dc65a`.

The `file_storage.cpp` conflict was resolved by taking main's config extraction and restoring only `#include <algorithm>` plus the allocation split; none of the removed config-parsing includes or code were retained.

Validation on `2b2dc65a`:

- Default/POSIX `file_storage_test`: 18/18 passed.
- io_uring/O_DIRECT `StorageBackendTest.BucketBatchLoadUsesUringAcrossQueueDepth` and `StorageBackendTest.BucketBatchLoadRejectsShortRead`: 2/2 passed, with logs confirming `O_DIRECT mode enabled`.
- `clang-format 20 --dry-run --Werror` and `git diff --check`: passed.

One earlier full-suite run hit a transient test-port collision (`Address already in use`) in `HeartbeatAfterDrainDoesNotRemount`. I reran that test three times (3/3 passed), then reran the full suite (18/18 passed).

fcczzz (approved):
LGTM.

ykwd (approved):

PR #3609, opened 2026-08-24
fix(ep,pg): prevent cudaErrorIllegalAddress during CUDA graph capture with TBO

UNIDY2002:
## Description

Fixes CUDA graph capture failures, illegal memory access errors, and a TBO-specific startup deadlock when running SGLang with Two-Batch Overlap (TBO), Mooncake elastic EP, Mooncake PG, and CUDA graphs enabled.

TBO runs EP dispatch/combine on a secondary stream concurrent with TP attention on the capture stream. The regression introduced in `90c5ce09` exposed several independent ordering and lifetime problems across the EP and PG paths. This PR keeps the original capture-safety fixes and completes the graph execution path so the captured graph records real EP communication and preserves per-dispatch metadata.

## Module

## Type of Change

- [x] Bug fix

### Fix 1 — Captured EP work routed to `retire()` instead of `retainUntilShutdown()` (`mooncake-pg`)

`MooncakeWorkCuda::is_captured_` was set using `at::cuda::currentStreamCaptureStatus()`, which queries the thread-default stream. In TBO, the EP collective stream may be capturing while the thread-default stream is not, so captured EP works were incorrectly routed to `retire()`. Subsequent `event->query()` calls on these works during capture triggered `cudaErrorIllegalAddress`.

**Fix:** Pass `is_captured` explicitly from `launchCollective()` in `mooncake_backend.cpp`, computed as `cudaStreamIsCapturing(stream)` OR `currentStreamCaptureStatus() != None`. Works created during capture on any stream are correctly retained until shutdown. The `capture_depth_` counter ensures `evictCompleted()` skips event queries while captured work is live, without external caller involvement.

### Fix 2 — `recv_hook` lambda invoking launcher during capture (`mooncake-ep`)

`dispatch()` and `combine()` return a `recv_hook` lambda when `return_recv_hook=True`. This lambda calls `launcher(LOW_LATENCY_RECV_PHASE)`, which spin-waits on `cudaHostAllocMapped` signal buffers. TBO's execute path can invoke this hook on the capture stream, causing `cudaErrorIllegalAddress`.

**Fix:** Add `if (graph_capture) return;` at the top of both `recv_hook` lambdas. The deferred host hook is therefore inactive during capture and remains available in eager execution.

### Fix 3 — Record the actual EP graph work

The previous capture workaround skipped the EP communication kernel during graph capture. That allowed startup to complete but produced an empty EP graph and invalid serving results.

**Fix:** During capture, `dispatch()` and `combine()` launch one combined `LOW_LATENCY_SEND_PHASE | LOW_LATENCY_RECV_PHASE` kernel on the capturing compute stream. The graph now contains the real EP operation while host-side peer waits and deferred receive hooks remain outside capture.

### Fix 4 — Preserve TBO dispatch metadata across interleaved logical dispatchers

TBO uses multiple logical dispatchers that share one native `MooncakeEpBuffer`. A process-wide buffer flip alone does not pair a later `combine()` with the dispatch metadata that produced it. Interleaved subbatches can consequently observe another dispatcher's top-k indices.

**Fix:** Assign each dispatch metadata pointer a stable private top-k snapshot slot in the native workspace. `dispatch()` copies `topk_idx` to that slot on the launch stream, and `combine()` resolves the matching slot from the metadata pointer. The native communication-buffer flip remains unchanged, preserving the validated eager communication protocol.

### Fix 5 — Avoid the eager PG enqueue handshake cycle

The PG worker submission path can otherwise create a cross-stream cycle during TBO warmup: the enqueue stream waits for the issue stream while the issue stream waits for completion work that depends on enqueue progress.

**Fix:** Detect capture on both the issue and enqueue streams. During eager execution, synchronize the issue stream before submission, wait for enqueue submission to be observed, and synchronize the issue stream after the completion event is recorded. During capture, retain graph-safe event ordering and avoid host waits for worker submission.

## API Compatibility

No external Mooncake API was changed. Python bindings, EP `dispatch()`/`combine()` signatures, and public capture interfaces remain unchanged. The changes are internal to the native EP and Torch PG implementations.

## How Has This Been Tested?

SGLang DSv3 TP4, TBO enabled (`--enable-two-batch-overlap`), Mooncake elastic EP backend (`--elastic-ep-backend mooncake`), Mooncake MoE A2A, `--deepep-mode low_latency`, `--cuda-graph-max-bs 128`, H20 8×GPU single node, Torch 2.11.0+cu130, SGLang `0ffed946`.

**Test command:**

```bash
python3 -m sglang.launch_server \
  --model-path <dsv3-model> \
  --tp 4 \
  --elastic-ep-backend mooncake \
  --moe-a2a-backend mooncake \
  --deepep-mode low_latency \
  --enable-two-batch-overlap \
  --disable-custom-all-reduce \
  --cuda-graph-max-bs 128
```

**Results:**

- [x] All 4 ranks complete 16/16 CUDA graph captures in approximately 169 seconds without hanging or hitting `cudaErrorIllegalAddress`
- [x] Server reaches "The server is fired up and ready to roll!"
- [x] GSM8K completes all 200 examples with accuracy `0.650`, invalid `0.000`, and `Ran 1 test ... OK`
- [x] The empty-graph candidate is rejected by semantic validation: it reached readiness but scored `0.005`

## Checklist

- [x] I have formatted my code using `clang-format`

## AI Assistance Disclosure

- [x] AI tools were used (Claude assisted with root-cause analysis and implementation; the human submitter has reviewed all changed lines and can defend each change end-to-end)

### DIFF
--- a/mooncake-ep/src/mooncake_ep_buffer.cpp
+++ b/mooncake-ep/src/mooncake_ep_buffer.cpp
@@ -56,9 +56,8 @@ bool stream_is_capturing(cudaStream_t stream) {
 constexpr int kTopkShadowMaxTokens = 1024;
 constexpr int kTopkShadowMaxTopk = 32;
 constexpr int kTopkShadowSlots = 8;
-constexpr size_t kTopkShadowBytes =
-    static_cast<size_t>(kTopkShadowMaxTokens) * kTopkShadowMaxTopk *
-    sizeof(int64_t);
+constexpr size_t kTopkShadowBytes = static_cast<size_t>(kTopkShadowMaxTokens) *
+                                    kTopkShadowMaxTopk * sizeof(int64_t);
 
 int64_t* topk_shadow(void* workspace, int shadow_slot, int num_experts) {
     auto* base = reinterpret_cast<uint8_t*>(workspace) +
--- a/mooncake-ep/src/mooncake_ep_buffer.cpp
+++ b/mooncake-ep/src/mooncake_ep_buffer.cpp
@@ -1,7 +1,6 @@
 #include <mooncake_ep_buffer.h>
 #include <glog/logging.h>
 #include <algorithm>
-#include <atomic>
 #include <cstdlib>
 #include <sstream>
 #include <transfer_engine.h>
@@ -44,10 +43,6 @@ cudaStream_t create_comm_stream() {
     return stream;
 }
 
-// CUDA Graph capture is initiated by PyTorch on the caller's stream, while
-// the decoupled EP implementation normally launches communication on its own
-// stream.  A raw CUDA capture query keeps this boundary Torch-free and lets
-// the legacy path stay on the capturing stream for the duration of capture.
 bool stream_is_capturing(cudaStream_t stream) {
     cudaStreamCaptureStatus status = cudaStreamCaptureStatusNone;
     auto error = cudaStreamIsCapturing(stream, &status);
@@ -56,6 +51,19 @@ bool stream_is_capturing(cudaStream_t stream) {
         return false;
     }
     return status != cudaStreamCaptureStatusNone;
+}
+
+constexpr int kTopkShadowMaxTokens = 1024;
+constexpr int kTopkShadowMaxTopk = 32;
+constexpr int kTopkShadowSlots = 8;
+constexpr size_t kTopkShadowBytes =
+    static_cast<size_t>(kTopkShadowMaxTokens) * kTopkShadowMaxTopk *
+    sizeof(int64_t);
+
+int64_t* topk_shadow(void* workspace, int shadow_slot, int num_experts) {
+    auto* base = reinterpret_cast<uint8_t*>(workspace) +
+                 2 * static_cast<size_t>(num_experts) * sizeof(int);
+    return reinterpret_cast<int64_t*>(base + shadow_slot * kTopkShadowBytes);
 }
 
 }  // namespace
@@ -223,6 +231,10 @@ MooncakeEpBuffer::dispatch(
     auto num_scales = hidden / 128;
     int num_local_experts = num_experts / num_ranks;
 
+    auto compute_stream_raw =
+        reinterpret_cast<cudaStream_t>(compute_stream_ptr);
+    const bool graph_capture = stream_is_capturing(compute_stream_raw);
+
     // Buffer control
     BufferPair layout(gdr_buffer, num_max_dispatch_tokens_per_rank, hidden,
                       num_ranks, num_experts);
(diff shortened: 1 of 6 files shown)


UNIDY2002:
Done by codex. Will review in person shortly.

mooncake-ep/src/mooncake_ep_buffer.cpp
```diff
@@ -300,17 +342,21 @@ MooncakeEpBuffer::dispatch(
             num_ranks, use_fp8, workspace, launch_stream, timeout_ticks, phases,
             active_qps_per_rank);
     };
-    if (return_recv_hook) {
-        launcher(LOW_LATENCY_SEND_PHASE);
-        mark_send_done();
+    if (graph_capture) {
+        launcher(LOW_LATENCY_SEND_PHASE | LOW_LATENCY_RECV_PHASE);
```
UNIDY2002, line 350: Sadly, this PR just a workaround, because of this line. Although it fixes the TBO issue, it clears the TBO semantics as well :(

UNIDY2002: Fixed in 3167cdeaacd42816158507f3ae19c8c9744a7afd

UNIDY2002:
Validation summary:

- On H20 with Torch 2.13 and the latest tested SGLang, Mooncake EP + Mooncake PG + TBO + CUDA Graph completed distributed initialization, CUDA Graph capture, readiness, semantic serving, and repeated graph replay with the deferred-RECV implementation (commit 3167cdea). Capture completed in 167.72 s, readiness followed in 5.02 s, the deterministic smoke request returned Paris, and replay batches 4/8/16 completed at 158.15/298.85/574.87 output tokens/s.

- The broader PR matrix also passed the no-TBO graph control, TBO eager control, TBO graph replay, DeepEP EP + Mooncake PG, and Mooncake EP + NCCL PG controls. The original PyPI 0.3.12.post1 baseline remained stuck during Mooncake PG initialization and did not reach graph capture, so no baseline performance comparison is claimed.

- The final deferred-RECV path records separate SEND and RECV launches during graph capture, preserving the TBO interval between them. Public Python/native APIs are unchanged, and the CUDA/MUSA platform boundary is preserved.

Remaining risks for follow-up:

- The PG worker now synchronizes issue_stream before and after eager task submission. This can prevent a TBO cross-stream/task-slot dependency cycle, but the post-submit synchronization makes eager PG operations host-synchronous and can reduce overlap. This is a real performance/async-semantics risk, but it does not block this PR because the target graph startup and serving regression is fixed. We should address it in a follow-up with per-slot/event ordering.

- The EP top-k shadow storage has fixed limits of 1024 tokens, top-k 32, and 8 pointer-indexed slots. The copy path lacks a token-count bound check, and slot wraparound or a missing dispatch/combine pairing can produce workspace overwrite or stale top-k metadata.

- capture_depth_ is tied to captured Work object lifetime rather than an explicit CUDA Graph capture lifetime, so early Work destruction could allow event eviction during an active capture.

- Unconditional cudaGetLastError() calls in the PG backend/tracker can clear and mask the original CUDA failure, reducing diagnosability.

These items are recorded as follow-up work; they are not blockers for the validated PR scope.

Disclaimer: This comment was generated by AI.

ShangmingCai (approved):
LGTM

PR #3612, opened 2026-08-24
[TransferEngine] Link gcov runtime for Rust coverage builds

alogfans:
## Description

Fix Transfer Engine Rust smoke test linking when the underlying C++ libraries are built with coverage instrumentation.

Coverage builds compile C/C++ objects with `--coverage`, which introduces references to gcov runtime symbols such as `__gcov_init`, `__gcov_exit`, and `__gcov_merge_add`. When Cargo links the `transfer_engine_rust` test binary directly against those archives, the gcov runtime was not included, causing `minimal_smoke` to fail at link time.

This change teaches `mooncake-transfer-engine/rust/build.rs` to locate the compiler-provided `libgcov` runtime and link it when available.

## Module

- [x] CI/CD

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

**Test commands:**
```bash
rustfmt mooncake-transfer-engine/rust/build.rs

cd mooncake-transfer-engine/rust
cargo test --test minimal_smoke --verbose -- --nocapture
```

**Test results:**

Manual verification confirmed that the Rust build script now emits:

```text
cargo:rustc-link-search=native=/usr/lib/gcc/x86_64-linux-gnu/11
cargo:rustc-link-lib=gcov
```

Full local test execution is still blocked on the local environment missing `libclang.so`, which is required by bindgen and unrelated to the gcov link failure.

## Checklist

## AI Assistance Disclosure

Codex was used to analyze the Rust/C++ link failure, identify the missing gcov runtime, and prepare the build script change.

### DIFF
--- a/mooncake-transfer-engine/rust/build.rs
+++ b/mooncake-transfer-engine/rust/build.rs
@@ -55,6 +55,14 @@ fn has_library(search_dirs: &[PathBuf], name: &str) -> bool {
             })
 }
 
+fn has_static_library(search_dirs: &[PathBuf], name: &str) -> bool {
+    let file_name = format!("lib{name}.a");
+    search_dirs.iter().any(|dir| dir.join(&file_name).exists())
+        || ["/usr/lib/x86_64-linux-gnu", "/usr/lib64", "/usr/local/lib"]
+            .into_iter()
+            .any(|dir| Path::new(dir).join(&file_name).exists())
+}
+
 fn is_tent_archive(lib: &str) -> bool {
     lib == "tent"
         || lib.starts_with("tent_")
@@ -343,13 +351,16 @@ fn main() {
     }
 
     // Coverage-instrumented C++ archives reference __gcov_* symbols. Cargo
-    // links the Rust test binary directly, so add the compiler's gcov runtime
-    // when it is available.
-    if emit_compiler_runtime_search("libgcov.a")
-        || emit_compiler_runtime_search("libgcov.so")
-        || has_library(&search_dirs, "gcov")
-    {
-        println!("cargo:rustc-link-lib=gcov");
+    // links the Rust test binary directly, so add GCC's static gcov runtime
+    // when it is available. Keep this last: libgcov.a must follow the
+    // instrumented archives on the link line. Non-coverage builds have no
+    // __gcov_* references, so the static archive contributes no objects.
+    if emit_compiler_runtime_search("libgcov.a") || has_static_library(&search_dirs, "gcov") {
+        if link_tent {
+            println!("cargo:rustc-link-arg=-lgcov");
+        } else {
+            println!("cargo:rustc-link-lib=gcov");
+        }
     }
 
     let include_dir = env::var("MOONCAKE_TE_INCLUDE_DIR").unwrap_or_else(|_| {
--- a/mooncake-transfer-engine/rust/build.rs
+++ b/mooncake-transfer-engine/rust/build.rs
@@ -125,6 +125,18 @@ fn compiler_print_file_name(file_name: &str) -> Option<PathBuf> {
     None
 }
 
+fn emit_compiler_runtime_search(file_name: &str) -> bool {
+    let Some(path) = compiler_print_file_name(file_name) else {
+        return false;
+    };
+    let Some(parent) = path.parent() else {
+        return false;
+    };
+
+    println!("cargo:rustc-link-search=native={}", parent.display());
+    true
+}
+
 fn main() {
     println!("cargo:rerun-if-env-changed=MOONCAKE_BUILD_DIR");
     println!("cargo:rerun-if-env-changed=MOONCAKE_TE_LIB_DIR");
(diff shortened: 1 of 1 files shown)


ShangmingCai (commented):
Thanks for tracking this down — I reproduced the failure and the fix looks right. A few notes, one of which is a latent gap rather than a blocker.

## Confirmed the failure

Latest nightly `nightly-coverage` ([run 32650599759](https://github.com/kvcache-ai/Mooncake/actions/runs/32650599759)) fails exactly as described:

```
rust-lld: error: undefined symbol: __gcov_init
  >>> referenced by status.cpp
  >>>               status.cpp.o:(_sub_I_00100_0) in archive .../libtransfer_engine_rust-*.rlib
```

Worth noting for anyone else looking at this: the **store** Rust smoke test passes in that same job, because `mooncake-store/rust/build.rs` (lines ~339-340 and ~385-386) already carries this identical gcov logic. So this PR is porting an approach that CI already proves works, which I think is the strongest argument for taking it as-is.

## Why it works, and why the placement is load-bearing

In the non-TENT path, `transfer_engine`/`base` go out as `rustc-link-lib=static=`, so rustc bundles those instrumented objects *into* `libtransfer_engine_rust-*.rlib`, which sits early on the link line. Because `cargo:rustc-link-lib=gcov` is the last `rustc-link-lib` emitted, `-lgcov` lands after the rlib group — the correct left-to-right archive order.

Suggest making that explicit in the comment so a future refactor doesn't hoist the block: *"Keep this last — libgcov.a must follow the instrumented archives on the link line."*

(For the record, the `-Bdynamic`/static-archive concern is a non-issue: both lld and GNU ld fall back from `libgcov.so` to `libgcov.a`, and gcc's own `--coverage` spec passes a bare `-lgcov`.)

## TENT + coverage will still fail to link

This is the one substantive gap. When `link_tent` is true, the instrumented archives are emitted as `cargo:rustc-link-arg=-Wl,--start-group -ltransfer_engine … -Wl,--end-group`. rustc emits `-C link-arg` values from `add_user_defined_link_args`, which runs *after* `add_local_native_libraries` in `linker_with_args` (`compiler/rustc_codegen_ssa/src/back/link.rs`; the comment there reads "Passed after compiler-generated options to support manual overriding when necessary"). So on the real command line `-lgcov` **precedes** the group, `libgcov.a` is scanned and discarded, and `__gcov_*` stays undefined.

Nothing hits this today — `tent-ci` builds without `--coverage`, and `nightly-coverage` builds without `USE_TENT` — so it isn't blocking. But it's a trap for whoever adds coverage to `tent-ci`. Either emit gcov as a link-arg after `--end-group` in the tent branch, or just record the limitation in the comment.

## Smaller things

1. **The `libgcov.so` probe is effectively dead, and its `-L` gets no rpath.** GCC ships only `libgcov.a`. Note the divergence from the store version: store pushes the directory into `search_dirs` *before* `emit_link_searches`, so it also picks up `-Wl,-rpath`; here it is printed directly, after `emit_link_searches` has already run. A shared libgcov would therefore link but fail to load. Either drop the `.so` probe or route it through `search_dirs` like store does.

2. **Clang coverage builds are a silent no-op.** `cc -print-file-name=libgcov.a` returns a nonexistent path under clang (whose runtime is `libclang_rt.profile-*.a`), so `compiler_print_file_name` correctly returns `None` and nothing is emitted. Harmless, and CI is gcc-on-ubuntu-22.04, so fine as-is.

   If you'd rather solve the TENT ordering, the clang case, and the "when is this even needed" question with one mechanism: gate on `--coverage` appearing in `CFLAGS`/`CXXFLAGS`/`LDFLAGS` (the coverage job already exports all three) and emit `cargo:rustc-link-arg=--coverage`. gcc's `LINK_COMMAND_SPEC` places `-lgcov` right after `%o`, i.e. after *every* user-supplied arg including the tent group, and clang maps `--coverage` to its own profile runtime.

3. **Reuse:** the new helper is exactly the inline block a few lines above it — the `MOONCAKE_LINK_ASAN` branch can collapse to `emit_compiler_runtime_search("libasan.so");`.

4. **Worth one clause on why this is ungated,** given the file documents the opposite policy for ASan directly above. Unlike libasan, `libgcov.a` is a static archive with no load-order constraint, so with no `__gcov_*` references no members are extracted and non-coverage builds are unaffected. Saying so preempts the obvious "why isn't this behind a flag like ASan?" review question.

## One process note

Since local testing was blocked on the missing `libclang.so`, this is currently verified only at the "build.rs emits the right directives" level, with no successful link. A `workflow_dispatch` run of `nightly.yml` on this branch would confirm the smoke test actually goes green before merge.

---
<sub>Review assisted by Claude Code; findings above were verified against the failing CI logs and rustc/lld sources.</sub>

stmatengss (approved):

staryxchen (approved):
LGTM

alogfans:
Thanks, agreed. I updated the gcov block to document the ordering requirement explicitly: it must stay last because libgcov.a has to follow the instrumented archives. I also added the ungated-linking rationale: non-coverage builds have no __gcov_* references, so the static archive contributes no objects.

I dropped the libgcov.so probe and now only look for the static libgcov.a path. For TENT builds, the gcov dependency is emitted as `cargo:rustc-link-arg=-lgcov`, so it follows the TENT archive group rather than being ordered before it by rustc's native-library handling.

Verified locally with:
- cargo fmt --check
- cargo check --lib
- env CFLAGS=--coverage CXXFLAGS=--coverage LDFLAGS=--coverage cargo test --lib --no-run

I still agree that a workflow_dispatch run of nightly.yml is the right final confirmation for the CI smoke path.

PR #3613, opened 2026-08-24
[TE] Add RDMA two-sided data-plane channel

zhtshr:
## Description

Data-plane channel for `rdma_twosided` (RFC #3377, PR3 of 4).

Builds on the control plane merged in #3440. Adds:

- **MsgChannel** – per-peer SEND/RECV QP for data frames; multi-rail capable.
- **BouncePool** – MR-registered bounce buffer with dynamic expand/shrink (low-watermark trigger / idle-shrink).
- **Credit admission** – wires SenderCreditLedger into SESSION_OPEN / CREDIT_GRANT / DATA_ACK ctrl frames; tasks exceeding the window enter WAITING and redispatch on grant.
- **Config** – ~15 `MC_RDMA_MSG_*` / `MC_RDMA_CREDIT_*` env knobs.
- **Handshake** – extends capability fields (`msg_qp_num`, `msg_rq_psn`, `supports_two_sided_msg`).

End-to-end TE integration (managed-buffer APIs, `submitTransfer` routing, correctness + perf tests) follows in PR4.

Closes no issue (tracked under #3377).

## Module

## Type of Change

- [x] New feature

## How Has This Been Tested?

**Test commands:**
```bash
# PR3 compiles independently (no TE-layer dependency)
ninja transfer_engine   # OK

# PR2 regression (full suite, eRDMA loopback)
MC_TE_FILTERS=erdma_0 MC_TEST_DEVICE_NAME=erdma_0 MC_MTU=1024 \
  ./build-tent/mooncake-transfer-engine/tests/rdma_notify_test
  ./build-tent/mooncake-transfer-engine/tests/ctrl_frame_test
  ./build-tent/mooncake-transfer-engine/tests/sender_credit_test
```

**Test results:**
- [x] Unit tests pass (rdma_notify 7/7, ctrl_frame 6/6, sender_credit 5/5)

Functional validation (MsgChannel SEND/RECV, bounce expand, credit sustainability) requires PR4 TE integration; verified locally with PR4 applied (6/6 rdma_twosided_test).

## Checklist

## AI Assistance Disclosure

AI pair-programming assistant used for code review feedback analysis, concurrency fix design (PR2 review comments), and PR workflow automation. All code reviewed and validated by the human submitter.

### DIFF
--- a/mooncake-transfer-engine/src/transport/rdma_transport/rdma_context.cpp
+++ b/mooncake-transfer-engine/src/transport/rdma_transport/rdma_context.cpp
@@ -1567,32 +1567,48 @@ int RdmaContext::poll(int num_entries, ibv_wc *wc, int cq_index) {
     return nr_poll;
 }
 
-void RdmaContext::registerMsgChannel(MsgChannel *channel) {
+void RdmaContext::registerMsgChannel(std::shared_ptr<MsgChannel> channel) {
     if (!channel) return;
     std::lock_guard<std::mutex> lock(msg_channels_mutex_);
-    for (auto *existing : msg_channels_) {
+    for (auto it = msg_channels_.begin(); it != msg_channels_.end();) {
+        auto existing = it->lock();
+        if (!existing) {
+            it = msg_channels_.erase(it);
+            continue;
+        }
         if (existing == channel) return;
+        ++it;
     }
-    msg_channels_.push_back(channel);
+    msg_channels_.push_back(std::move(channel));
 }
 
 void RdmaContext::unregisterMsgChannel(MsgChannel *channel) {
     if (!channel) return;
     std::lock_guard<std::mutex> lock(msg_channels_mutex_);
+    // Called from ~MsgChannel, where the weak_ptr no longer locks, so match on
+    // identity and drop expired entries as we go.
     msg_channels_.erase(
-        std::remove(msg_channels_.begin(), msg_channels_.end(), channel),
+        std::remove_if(msg_channels_.begin(), msg_channels_.end(),
+                       [channel](const std::weak_ptr<MsgChannel> &entry) {
+                           auto held = entry.lock();
+                           return !held || held.get() == channel;
+                       }),
         msg_channels_.end());
 }
 
 int RdmaContext::pollMsgChannels(int max_per_channel) {
-    std::vector<MsgChannel *> snapshot;
+    // Promote under the lock so each polled channel stays alive for the poll
+    // even if the transport replaces the rail concurrently.
+    std::vector<std::shared_ptr<MsgChannel>> snapshot;
     {
         std::lock_guard<std::mutex> lock(msg_channels_mutex_);
-        snapshot = msg_channels_;
+        snapshot.reserve(msg_channels_.size());
+        for (auto &entry : msg_channels_) {
+            if (auto channel = entry.lock()) snapshot.push_back(channel);
+        }
     }
     int processed = 0;
-    for (auto *channel : snapshot) {
-        if (!channel) continue;
+    for (auto &channel : snapshot) {
         int n = channel->pollCompletions(max_per_channel);
         if (n > 0) processed += n;
     }
@@ -1601,7 +1617,10 @@ int RdmaContext::pollMsgChannels(int max_per_channel) {
 
 bool RdmaContext::hasMsgChannels() const {
     std::lock_guard<std::mutex> lock(msg_channels_mutex_);
-    return !msg_channels_.empty();
+    for (auto &entry : msg_channels_) {
+        if (!entry.expired()) return true;
+    }
+    return false;
 }
 
 int RdmaContext::submitPostSend(
--- a/mooncake-transfer-engine/src/transport/rdma_transport/rdma_context.cpp
+++ b/mooncake-transfer-engine/src/transport/rdma_transport/rdma_context.cpp
@@ -48,6 +48,7 @@
 #include "transport/rdma_transport/rdma_endpoint.h"
 #include "transport/rdma_transport/rdma_transport.h"
 #include "transport/rdma_transport/worker_pool.h"
+#include "transport/rdma_twosided/msg_channel.h"
 #include "transport/transport.h"
 
 namespace mooncake {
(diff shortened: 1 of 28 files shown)


mooncake-transfer-engine/src/transport/rdma_transport/rdma_context.cpp
```diff
@@ -1468,6 +1469,43 @@ int RdmaContext::poll(int num_entries, ibv_wc *wc, int cq_index) {
...
+        snapshot = msg_channels_;
+    }
+    int processed = 0;
+    for (auto *channel : snapshot) {
+        if (!channel) continue;
+        int n = channel->pollCompletions(max_per_channel);
+        if (n > 0) processed += n;
+    }
+    return processed;
+}
```
alogfans, line 1530: pollMsgChannels() takes a snapshot of raw MsgChannel* pointers and releases msg_channels_mutex_ before dereferencing them. A concurrent rail replacement or transport teardown can unregister and destroy a channel while it is still present in this snapshot. Will this happen?

mooncake-transfer-engine/src/transport/rdma_twosided/two_sided.cpp
```diff
@@ -0,0 +1,684 @@
...
+                    sender_credit_.rollbackReservation(
+                        peer, session, epoch,
+                        {{CreditResource::BounceSlots, slots_needed},
+                         {CreditResource::BounceBytes, length}});
+                    std::lock_guard<std::mutex> lock(twosided_mutex_);
+                    auto it = twosided_tasks_.find(task_id);
+                    if (it != twosided_tasks_.end()) {
+                        it->second.credit_reserved = false;
+                    }
+                }
```
alogfans: This failure can occur after earlier chunks have already been posted. Rolling back the full reservation and redispatching from offset zero duplicates those chunks, and the receiver's cumulative byte accounting can then complete the task before all unique data has arrived. Please store the next unsent offset and resume from there, or make the receive path idempotent. This is required for the basic path because ordinary send-queue backpressure can trigger it.

alogfans:
BTW, please add a standalone rdma benchmark program or unit tests to show that the proposed two-sided rdma transport can serve heavy workloads.

stmatengss:
@zhtshr Please resolve all comments.

mooncake-transfer-engine/src/transport/rdma_transport/rdma_context.cpp
zhtshr: Yes. A rail replacement or teardown can drop the last shared_ptr while
pollMsgChannels is still dereferencing the raw pointer it took from the
snapshot.

RdmaContext now stores weak_ptr and pollMsgChannels locks each entry before
use, skipping expired ones, so the channel stays alive across the poll.

mooncake-transfer-engine/src/transport/rdma_twosided/two_sided.cpp
zhtshr: Agreed. The task state now stores the next unsent offset and redispatch
resumes from there, so ERR_TOO_MANY_REQUESTS is a pause with the
reservation held rather than an abort.

The receive path also counts completed chunks against the chunk total
instead of cumulative bytes, so duplicate arrival can't complete a task
early.

zhtshr:
Added a standalone benchmark in mooncake-transfer-engine/example/rdma_twosided_bench.cpp:
configurable threads, queue depth, block size, opcode and duration, with every payload
verified against a per-request stamp. 16 threads x depth 4 x 64 KiB over 30s gives
2.85M batches, 94.9k IOPS, 6.22 GB/s, no failures or stamp mismatches. Two caveats:
it's a single-host loopback over one NIC, so it stresses the software path rather
than fabric bandwidth; and throughput drops as in-flight chunks grow (~6-8 GB/s at
64, ~0.15 at 4096) once the bounce pool saturates before the link does.

This series also fixes several races found while exercising it: concurrent
MsgChannel rail setup is now serialised per "peer|nicPath" with the connect-scope
pattern from #3440, so a handshake that throws cannot strand the in-flight key and
leave later builders waiting out the full deadline; the receiver-side per-task byte
ledger, which was growing unboundedly and could complete a task early on duplicate
arrival, now counts completed chunks; and MsgChannel poll now releases every
in-flight bounce slot and settles pending_sends_ on an error WC, mirroring the
ctrl-channel fix.

catyans (changes requested):
I found two data-correctness blockers in the current two-sided path:

1. The receiver ACK ledger is keyed only by `task_id`, while every sending transport initializes `next_task_id_` to 1. Two peers can therefore send the same task ID to one receiver. Their bytes/chunks are merged in `recv_acked_bytes_`, so an interleaved multi-chunk transfer can make one peer receive a cumulative ACK for bytes it has not delivered and complete early, while the other task can remain pending after the shared entry is erased. A peer restart can collide with an incomplete ledger entry in the same way. Please key receiver state by at least `(peer/session, task_id)` and add a two-sender interleaving/restart regression test.

2. `slice_seq` is only 8 bits, but dispatch casts the unbounded chunk index to `uint8_t`, and READ response placement reconstructs the destination offset solely as `slice_seq * max_payload`. With the default 64 KiB payload, a READ larger than 256 chunks (about 16 MiB) wraps to offset zero, overwrites earlier data, and can still be marked successful because completion counts bytes. Please carry a sufficiently wide byte offset/chunk index on the wire—or reject oversized READs—and add a verified READ test crossing the 256-chunk boundary.

The current CI and benchmark are green, but the benchmark defaults to one sender and 64 KiB requests, so it does not exercise either case. I would request changes until these are covered.

zhtshr:
All three red checks on af7575e6 are infrastructure failures, not code failures.

af7575e6 is a content-free amend of aeaafdd3 (git diff between them is empty), yet build-wheel-cu13 (3.10) passed on aeaafdd3 and failed here — same bytes, different result. Its Build project step is still in_progress with all later steps pending and no log uploaded: a lost runner.

build-wheel-rocm (3.10) has zero error:` lines in its log. It reaches[192/228] Linking CXX executable .../mooncake_master, goes silent for 18 minutes, then exits 137 (SIGKILL, "The runner has received a shutdown signal") — a link-stage OOM.CI Gateis just the cascade of those two.

Everything substantive is green, including tent-ci on all four variants.

I only have pull on this repo so I cannot re-run jobs (I already force-pushed an amend to get a fresh run). Could a maintainer re-run the three failed jobs?

catyans:
The original ACK-key and `slice_seq` issues are addressed, but the new deferred `READ_RESP` path has a lost-wakeup window. `sendOrQueueReadResp()` first observes `ERR_TOO_MANY_REQUESTS` and only then locks `resp_mutex_` to append to `pending_resps_`. Between those operations, the last outstanding SEND can complete, release a slot, and call `drainPendingReadResps()` while the queue is still empty. The response is then enqueued after the drain returns; if there are no more SEND completions and no pool expansion, nothing triggers another drain, so the requester waits for that slice indefinitely.

Please make enqueueing and the drain trigger race-free—for example, enqueue and then immediately participate in a serialized drain, or use a single-drainer/armed flag protocol that rechecks the queue after slot release. A deterministic regression test can pause between the initial pool-full result and enqueue, complete the last send in that window, then assert that the queued `READ_RESP` is still posted without requiring another completion or expansion.

catyans (approved):
Thanks for sticking with this one—the current head resolves the issues that were blocking me. LGTM.

The receive ledger is now scoped by `(peer, session, task_id)`, `slice_seq` is wide enough for large transfers, and partial sends resume from the next unsent offset instead of replaying the prefix. I also rechecked the deferred `READ_RESP` path: enqueue now participates in the single-drainer protocol itself, while a concurrent slot release arms a recheck before the drainer exits. The deterministic tests cover both sides of that lost-wakeup window.

The weak-to-shared promotion in `RdmaContext` and the CQ/resource lock also address the channel-lifetime concerns that @alogfans and Copilot raised. Add in the standalone verified-load benchmark and the fully green CI, and this feels ready. Really appreciate the thorough follow-through across several subtle races.

alogfans (approved):
LGTM. This feature should be opt-in in the current stage

alogfans:
@stmatengss Do you think it's okay to merge two-sided rdma xport for now?

PR #3614, opened 2026-08-24
[Bugfix][Store] Fix storage_backend_test build break (#3465); replication task cleanup (#3401)

### LINKED ISSUE #3401: [Bug]: Replication tasks retain targets and reserved quota after source cleanup

XFDG:
## Affected code paths

- mooncake-store/include/master_service.h
  - new method declaration: CancelReplicationTaskForRemovedSource()
- mooncake-store/src/master_service.cpp
  - CancelReplicationTaskForRemovedSource() implementation
  - CleanupStaleHandles(): calls CancelReplicationTaskForRemovedSource() after CancelPromotionTaskForRemovedReplicas()

## Root cause

When an in-flight Copy/Move replication source replica becomes unavailable (e.g. its segment is unmounted), CleanupStaleHandles() removes the invalid source replica but only cancels promotion tasks for removed replica IDs. The replication task itself is left behind together with its PROCESSING target replicas and reserved tenant quota.

The source refcnt is never decremented, the PROCESSING targets are never discarded, and the reserved quota is never released — all until the client calls CopyEnd/MoveRevoke or the task times out.

## The fix

Add CancelReplicationTaskForRemovedSource(), called from CleanupStaleHandles() after the replicas are removed. When the removed set includes the task source_id, it:
1. Decrements the source refcnt (symmetric to CopyStart increment).
2. Pops and frees the PROCESSING target replicas.
3. Releases the reserved tenant quota.
4. Clears dynamic replication state if applicable.
5. Erases the replication task.

This mirrors the cleanup done by DiscardExpiredProcessingReplicas and EraseMetadata for a completed replication task.

## Impact

- Replication tasks are now properly cleaned up when their source replica becomes unavailable.
- Reserved tenant quota is released immediately instead of leaking until task timeout.
- Source refcnt is properly decremented, preventing memory leaks.

## How it has been tested

- Compiled mooncake-store successfully on Beijing H200 (CUDA 12.8) with no warnings.
- All CleanupStaleHandles call paths (unmount, client-expiry, LOCAL_DISK sweep) now trigger the new cleanup.

Fixes kvcache-ai/Mooncake#3401

### DIFF
--- a/mooncake-store/include/master_service.h
+++ b/mooncake-store/include/master_service.h
@@ -2072,6 +2072,20 @@ class MasterService {
         const std::vector<ReplicaID>& removed_replica_ids)
         NO_THREAD_SAFETY_ANALYSIS;
 
+    // When a replica that is the source_id of an in-flight Copy/Move
+    // replication task is removed (e.g. its segment unmounted), the task is
+    // stranded: its PROCESSING targets never complete, the reserved tenant
+    // quota is never released, and the source refcnt is never decremented.
+    // This cancels such a task promptly: discard the PROCESSING targets,
+    // release reserved quota, decrement the source refcnt, and erase the
+    // task. Mirrors the cleanup done by DiscardExpiredProcessingReplicas /
+    // EraseMetadata for a completed replication task. No-op when there is no
+    // replication task, or its source_id is not among the removed replicas.
+    void CancelReplicationTaskForRemovedSource(
+        TenantState& tenant_state, ObjectMetadata& metadata,
+        const std::vector<ReplicaID>& removed_replica_ids)
+        NO_THREAD_SAFETY_ANALYSIS;
+
     // Lease related members
     const uint64_t default_kv_lease_ttl_;     // in milliseconds
     const uint64_t default_kv_soft_pin_ttl_;  // in milliseconds
--- a/mooncake-store/src/master_service.cpp
+++ b/mooncake-store/src/master_service.cpp
@@ -6964,7 +6964,7 @@ void MasterService::CancelReplicationTaskForRemovedSource(
                          r.id()) != task.replica_ids.end();
     };
     auto removed_targets =
-        metadata.PopReplicasWithCacheTotalAccounting(target_pred);
+        PopReplicasWithCacheTotalAccounting(metadata, target_pred);
     std::vector<ReplicaID> erased_ids;
     erased_ids.reserve(removed_targets.size());
     for (const auto& r : removed_targets) erased_ids.push_back(r.id());
--- a/mooncake-store/src/master_service.cpp
+++ b/mooncake-store/src/master_service.cpp
@@ -6937,6 +6937,56 @@ void MasterService::CancelPromotionTaskForRemovedReplicas(
                                        metadata.user_key);
 }
 
+void MasterService::CancelReplicationTaskForRemovedSource(
+    TenantState& tenant_state, ObjectMetadata& metadata,
+    const std::vector<ReplicaID>& removed_replica_ids) {
+    if (removed_replica_ids.empty()) return;
+
+    auto task_it = tenant_state.replication_tasks.find(metadata.user_key);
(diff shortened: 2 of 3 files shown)


Aionw:
Thanks for PR. But this is duplicated with https://github.com/kvcache-ai/Mooncake/pull/3509

XFDG:
## Closing as duplicate — superseded by #3509

Thanks @Aionw for flagging the duplication.

After reviewing #3509 in full, its #3401 fix is strictly more complete than the one bundled here:

- #3509 keeps a lightweight **source_removed tombstone** so a concurrent or delayed CopyEnd/MoveEnd/CopyRevoke/MoveRevoke still returns the correct REPLICA_IS_GONE semantics. This PR (#3614) erases the task outright, which would break a delayed End/Revoke arriving after stale-handle cleanup.
- #3509 gates local cleanup on OpLog durability in HA mode (commit-before-mutate, generation-bound cleanup transaction).
- #3509 retires the PROCESSING target buffers to discarded_replicas_ until the release deadline instead of freeing them immediately, avoiding allocator reuse while a late RDMA write may still land.

Additionally, the other half of this PR (the #3465 storage_backend_test build break) is already resolved on main: both PopReplicasWithCacheTotalAccounting overloads are now non-static members and all call sites are member calls, so the qualified-static compile error no longer reproduces.

For both halves, #3509 is the correct, complete home. Closing #3614 in favor of it. CC @Icedcoco.

PR #3615, opened 2026-08-24
[Bugfix][Store] Properly enable O_DIRECT for OffsetAllocator vector I/O

### LINKED ISSUE #3611: [Bug]: Store's OffsetAllocatorStorageBackend: use_direct_io=true is misleading dead code
## Description

`OffsetAllocatorStorageBackend` constructs its `UringFile` with `use_direct_io = true`, and the log line `[UringFile] O_DIRECT mode enabled for <file>` is printed. However, **O_DIRECT is never actually enabled on the file descriptor**, and the I/O hot path (`vector_read`/`vector_write`) contains **no alignment handling whatsoever**. This makes `use_direct_io=true` deceptive dead code in this backend.

There are two distinct problems:

### Problem 1: `open()` does not set `O_DIRECT`

In `mooncake-store/src/storage_backend.cpp`, `OffsetAllocatorStorageBackend::Init()` opens the data file with:

```cpp
// storage_backend.cpp:3912-3913
int flags = O_CLOEXEC | O_RDWR | O_CREAT | O_TRUNC;   // no O_DIRECT
int raw_fd = open(data_file_path_.c_str(), flags, 0644);
```

then passes `use_direct_io = true` to `UringFile`:

```cpp
// storage_backend.cpp:3934-3935
data_file_ = std::make_shared<UringFile>(
    data_file_path_, fd_guard.release(), 32, true);
```

But the `UringFile` constructor (`mooncake-store/src/uring_file.cpp:414-428`) **only stores the flag and prints the log** — it does not call `fcntl(fd, F_SETFL, flags | O_DIRECT)`:

```cpp
// uring_file.cpp:578-590
UringFile::UringFile(const std::string& filename, int fd,
                     unsigned /*queue_depth*/, bool use_direct_io)
    : StorageFile(filename, fd), use_direct_io_(use_direct_io) {
    ...
    if (use_direct_io_) {
        LOG(INFO) << "[UringFile] O_DIRECT mode enabled for " << filename;  // misleading!
    }
}
```

As a result, all I/O on this fd goes through the **page cache (buffered I/O)**, not direct I/O. The log message is misleading — it suggests O_DIRECT is active when it is not.

For comparison, `BucketStorageBackend::OpenFile` (`storage_backend.cpp:2604-2617`) does this correctly by **both** adding `O_DIRECT` to the `open()` flags **and** passing `use_direct_io = true`:

```cpp
// storage_backend.cpp:3570-3590 & 932-963  (BucketStorageBackend — correct)
if (file_storage_config_.use_uring && mode == FileMode::Read) {
    flags |= O_DIRECT;                    // <-- fd gets O_DIRECT
}
int fd = open(path.c_str(), flags | access_mode, 0644);
...
if (file_storage_config_.use_uring && mode == FileMode::Read) {
    return std::make_unique<UringFile>(path, fd, 32, true);
}
```

### Problem 2: `vector_read`/`vector_write` ignore `use_direct_io_` entirely

The `use_direct_io_` flag only affects the scalar methods `read()` / `write()` / `read_aligned()` / `write_aligned()` (which allocate 4096-aligned bounce buffers and validate alignment).

However, `vector_read` and `vector_write` (`uring_file.cpp:623-671`) **do not check `use_direct_io_` at all** — no alignment validation, no bounce buffer:

```cpp
// uring_file.cpp:791-805
tl::expected<size_t, ErrorCode> UringFile::vector_write(const iovec* iov,
                                                        int iovcnt,
                                                        off_t offset) {
    if (fd_ < 0) return make_error<size_t>(ErrorCode::FILE_NOT_FOUND);
    // no use_direct_io_ check, no alignment handling
    auto res = SharedUringRing::instance().vector_write(fd_, iov, iovcnt, offset);
    ...
}

// uring_file.cpp:807-839  (same pattern for vector_read)
```

Critically, **`OffsetAllocatorStorageBackend` exclusively uses the vector I/O path**:

- `BatchOffload` calls `data_file->vector_write(iovs.data(), iovs.size(), offset)` (`storage_backend.cpp:5178`)
- `BatchLoad` calls `plan.data_file->vector_read(...)` **three times per key** (`storage_backend.cpp:5418`, `5442`, `5463`)

The scalar methods that *would* respect `use_direct_io_` are never called by this backend.

Furthermore, the `iovec`s constructed in `BatchOffload` (`storage_backend.cpp:~5150-5175`) are **completely unaligned**:

- `&header.key_len` / `&header.value_len` — 4-byte stack variables, not 4096-aligned
- `key.data()` — `std::string` internal buffer, not guaranteed aligned
- `slice.ptr` — caller-provided, not guaranteed aligned
- `offset` — from `OffsetAllocator::allocate()`, arbitrary offset, not 4096-aligned
- `iov_len` values — 4, key length, value length, etc., not 4096-aligned

## Expected behavior

Either:
- O_DIRECT should be genuinely enabled (via `open()` flags or `fcntl(F_SETFL)`) **and** the vector I/O path should enforce 4096 alignment on buffers, lengths, and offsets; **or**
- the `use_direct_io = true` argument and the misleading "O_DIRECT mode enabled" log should be removed/changed to reflect that this backend uses buffered I/O.

## Actual behavior

- `use_direct_io = true` is passed to `UringFile`, the log prints "O_DIRECT mode enabled", but the fd is opened without `O_DIRECT` — all I/O is buffered (page cache).
- The vector I/O hot path has no alignment handling, so the `use_direct_io_` flag is completely bypassed.
- If someone were to "fix" Problem 1 by adding `O_DIRECT` to `open()` **without also fixing** Problem 2, `vector_read`/`vector_write` would immediately fail with `EINVAL` due to unaligned buffers/offsets/lengths.

## Environment

- Component: `mooncake-store`
- Relevant files:
  - `mooncake-store/src/storage_backend.cpp` (`OffsetAllocatorStorageBackend::Init`, `BatchOffload`, `BatchLoad`)
  - `mooncake-store/src/uring_file.cpp` (`UringFile` constructor, `vector_read`, `vector_write`)
  - `mooncake-store/include/file_interface.h` (`UringFile` declaration, `use_direct_io` parameter)

## Checklist

- [x] I have reviewed the changed code end-to-end and can defend the change.
- [x] AI assistance disclosure: This issue was drafted with AI assistance; the findings have been verified against the source code.

### Before submitting...

huangxiaobai-ydmy:
## Description

Fixes #3611.

`OffsetAllocatorStorageBackend` passed `use_direct_io=true` to `UringFile` and logged "O_DIRECT mode enabled", but the data file was opened without `O_DIRECT`, and `BatchOffload`/`BatchLoad` use unaligned `vector_write`/`vector_read` iovecs that bypassed direct-I/O alignment handling.

This PR properly enables O_DIRECT end-to-end:

1. Set `O_DIRECT` on the fd when opening OffsetAllocator data files (`Init`, `TryRecoverFromMetadata`, `RemoveAll`) when uring is enabled.
2. Add bounce-buffer alignment handling (with RMW on writes) in `UringFile::vector_read` and `vector_write` so unaligned iovecs work under `use_direct_io=true`.
3. Add `UringVectorIoHandlesUnalignedDirectIo` in `posix_file_test.cpp` to cover unaligned vector I/O under O_DIRECT, including neighbor-data preservation across RMW writes.

Follow-up to maintainer review: the earlier approach of setting `use_direct_io=false` was incorrect; this implements the expected O_DIRECT behavior instead.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

Added unit test `UringVectorIoHandlesUnalignedDirectIo` for unaligned vector read/write with `use_direct_io=true`.

Full build/tests not run locally (Windows environment without Mooncake build toolchain); please rely on CI.

**Test commands:**
```bash
# CI: Build & Test (Linux)
# Local (Linux): ctest --test-dir build -R PosixFileTest.UringVectorIoHandlesUnalignedDirectIo
```

**Test results:**
- [ ] Unit tests pass
- [x] Manual testing done (code review; unit test added)

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

Cursor AI assisted with issue analysis, O_DIRECT implementation, unit test, clang-format fixes, and PR description drafting. The submitter reviewed the changes.

### DIFF
--- a/mooncake-store/src/uring_file.cpp
+++ b/mooncake-store/src/uring_file.cpp
@@ -858,8 +858,13 @@ tl::expected<size_t, ErrorCode> UringFile::vector_write(const iovec* iov,
             auto read_res = SharedUringRing::instance().read(
                 fd_, bounce, aligned_len, aligned_off);
             if (!read_res) {
-                std::memset(bounce, 0, aligned_len);
-            } else if (read_res.value() < aligned_len) {
+                // A real read error must not proceed with a zero-filled
+                // bounce buffer — that would silently corrupt neighbors.
+                free_aligned_buffer(bounce);
+                return make_error<size_t>(ErrorCode::FILE_READ_FAIL);
+            }
+            if (read_res.value() < aligned_len) {
+                // EOF / short read: only zero the unread suffix.
                 std::memset(static_cast<char*>(bounce) + read_res.value(), 0,
                             aligned_len - read_res.value());
             }
--- a/mooncake-store/src/uring_file.cpp
+++ b/mooncake-store/src/uring_file.cpp
@@ -788,13 +788,95 @@ tl::expected<void, ErrorCode> UringFile::batch_read(ReadDesc* descs, int cnt) {
 // vector_write / vector_read
 // ---------------------------------------------------------------------------
 
+namespace {
+
+size_t IovecTotalLen(const iovec* iov, int iovcnt) {
+    size_t total = 0;
+    for (int i = 0; i < iovcnt; ++i) total += iov[i].iov_len;
+    return total;
+}
+
+bool IovecRegionDirectReady(const iovec* iov, int iovcnt, off_t offset,
+                            size_t alignment) {
+    if (offset % static_cast<off_t>(alignment) != 0) return false;
+    for (int i = 0; i < iovcnt; ++i) {
+        if (reinterpret_cast<uintptr_t>(iov[i].iov_base) % alignment != 0) {
+            return false;
+        }
+        if (iov[i].iov_len % alignment != 0) return false;
+    }
+    return true;
+}
+
+void CopyFromIovec(char* dst, const iovec* iov, int iovcnt) {
+    for (int i = 0; i < iovcnt; ++i) {
+        std::memcpy(dst, iov[i].iov_base, iov[i].iov_len);
+        dst += iov[i].iov_len;
+    }
+}
+
+void CopyToIovec(char* src, iovec* iov, int iovcnt) {
+    for (int i = 0; i < iovcnt; ++i) {
+        std::memcpy(iov[i].iov_base, src, iov[i].iov_len);
+        src += iov[i].iov_len;
+    }
+}
+
+}  // namespace
+
 tl::expected<size_t, ErrorCode> UringFile::vector_write(const iovec* iov,
                                                         int iovcnt,
                                                         off_t offset) {
     if (fd_ < 0) return make_error<size_t>(ErrorCode::FILE_NOT_FOUND);
+    if (!iov || iovcnt <= 0)
+        return make_error<size_t>(ErrorCode::FILE_INVALID_BUFFER);
+
+    const size_t total = IovecTotalLen(iov, iovcnt);
+    if (total == 0) return make_error<size_t>(ErrorCode::FILE_INVALID_BUFFER);
+
     auto start = std::chrono::steady_clock::now();
-    auto res =
-        SharedUringRing::instance().vector_write(fd_, iov, iovcnt, offset);
+    tl::expected<size_t, ErrorCode> res;
+
+    if (!use_direct_io_ ||
+        IovecRegionDirectReady(iov, iovcnt, offset, ALIGNMENT_)) {
+        res = SharedUringRing::instance().vector_write(fd_, iov, iovcnt,
+                                                       offset);
+    } else {
+        const off_t aligned_off = offset & ~static_cast<off_t>(ALIGNMENT_ - 1);
+        const size_t head = static_cast<size_t>(offset - aligned_off);
+        const size_t span = head + total;
+        const size_t aligned_len =
+            ((span + ALIGNMENT_ - 1) / ALIGNMENT_) * ALIGNMENT_;
+
+        void* bounce = alloc_aligned_buffer(aligned_len);
+        if (!bounce) return make_error<size_t>(ErrorCode::FILE_WRITE_FAIL);
+
+        // Preserve bytes outside [offset, offset+total) within the aligned
+        // region; otherwise padding would clobber neighboring data.
+        const bool needs_rmw = (head != 0) || (aligned_len != total);
+        if (needs_rmw) {
+            auto read_res = SharedUringRing::instance().read(
+                fd_, bounce, aligned_len, aligned_off);
+            if (!read_res) {
+                std::memset(bounce, 0, aligned_len);
+            } else if (read_res.value() < aligned_len) {
+                std::memset(static_cast<char*>(bounce) + read_res.value(), 0,
+                            aligned_len - read_res.value());
+            }
+        }
+        CopyFromIovec(static_cast<char*>(bounce) + head, iov, iovcnt);
+
+        res = SharedUringRing::instance().write(fd_, bounce, aligned_len,
+                                                aligned_off);
+        free_aligned_buffer(bounce);
+
+        if (res && res.value() >= span) {
+            res = total;
+        } else if (res) {
+            res = make_error<size_t>(ErrorCode::FILE_WRITE_FAIL);
+        }
+    }
+
     auto us = std::chrono::duration_cast<std::chrono::microseconds>(
                   std::chrono::steady_clock::now() - start)
                   .count();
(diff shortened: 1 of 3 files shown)


Aionw:
Thanks for PR.

This isn't a proper fix here. What we expect is proper handle `O_DIRECT` flag instead of change the default value to match log.

huangxiaobai-ydmy:
Thanks for the review. I've updated the PR to properly handle O_DIRECT instead of changing the default to false:
1.Set O_DIRECT on the fd when opening OffsetAllocator data files (Init / recovery / RemoveAll).
2.Add bounce-buffer alignment handling in UringFile::vector_read and vector_write so unaligned iovecs (header + key + slices) 3.work correctly under O_DIRECT.
Added a unit test covering unaligned vector I/O with use_direct_io=true.

huangxiaobai-ydmy:
Hi @Aionw, thanks again for the earlier feedback.

I've updated the PR to properly handle O_DIRECT as you suggested (instead of changing the default to false):
1.Set O_DIRECT on the fd when opening OffsetAllocator data files.
2.Add bounce-buffer alignment handling in UringFile::vector_read / vector_write for unaligned iovecs.
3.Add a unit test (UringVectorIoHandlesUnalignedDirectIo) for the unaligned vector I/O path.
CI is green on the latest commit (30f5702), including CI Gate and Check code format.
Whenever you have a moment, could you take another look? Happy to address any further comments. Thanks!

catyans:
Thanks for following up on @Aionw's point — enabling `O_DIRECT` on the OffsetAllocator file and handling unaligned vector I/O with a bounce buffer feels like the right fix. The neighbor-preservation test is helpful too.

I think there is one failure case we should tighten before relying on this path: in `UringFile::vector_write()`, if the preservation read for an unaligned read-modify-write fails, we zero the entire aligned bounce buffer and then continue with the write. For an existing block, that can turn a read error into silent corruption of neighboring record bytes outside the requested range. A genuine EOF/short read can safely zero just the unread suffix, but an actual read error should abort the write rather than manufacture zero-filled contents.

Could we return `FILE_READ_FAIL` from that branch and add a focused regression test showing that a failed preservation read performs no write and leaves the adjacent data intact? That would complement the current successful-RMW test nicely and make this behavior much safer to reason about. Thanks!

huangxiaobai-ydmy:
Thanks for the careful review — agreed.

Updated `UringFile::vector_write()` so a failed preservation read now returns `FILE_READ_FAIL` and does **not** write (no zero-filled bounce buffer). Short/EOF reads still only zero the unread suffix.

Also added `UringVectorWriteAbortsWhenPreservationReadFails`: seed neighboring bytes, reopen the file `O_WRONLY|O_DIRECT` so the RMW read fails, assert `FILE_READ_FAIL`, then verify the neighbor data is unchanged.

catyans (approved):
Thanks, this addresses exactly the failure path I was worried about. LGTM.

Returning `FILE_READ_FAIL` before issuing any write preserves the neighboring bytes, while a genuine short/EOF read still zero-fills only the unread suffix. The `O_WRONLY | O_DIRECT` regression is a neat way to force the preservation read to fail, and it verifies that the existing data stays untouched. Nice, careful follow-up—and it keeps the original `O_DIRECT` direction intact.

huangxiaobai-ydmy:
Hi @ykwd @stmatengss @XucSh @YiXR — gentle ping when you have a moment.

CI is green and @catyans has already approved. This PR only touches the OffsetAllocator O_DIRECT / unaligned vector I/O path. If the direction looks good, would you mind taking a look for a CODEOWNER review when convenient? No rush, and happy to address any further feedback. Thanks!

mooncake-store/src/uring_file.cpp
```diff
@@ -788,13 +788,100 @@ tl::expected<void, ErrorCode> UringFile::batch_read(ReadDesc* descs, int cnt) {
...
 
+namespace {
+
+size_t IovecTotalLen(const iovec* iov, int iovcnt) {
+    size_t total = 0;
+    for (int i = 0; i < iovcnt; ++i) total += iov[i].iov_len;
+    return total;
+}
+
+bool IovecRegionDirectReady(const iovec* iov, int iovcnt, off_t offset,
```
Aionw: Nit. This function can be named IsIovecRegionAligned

mooncake-store/src/uring_file.cpp
```diff
@@ -788,13 +788,100 @@ tl::expected<void, ErrorCode> UringFile::batch_read(ReadDesc* descs, int cnt) {
...
+
 tl::expected<size_t, ErrorCode> UringFile::vector_write(const iovec* iov,
                                                         int iovcnt,
                                                         off_t offset) {
     if (fd_ < 0) return make_error<size_t>(ErrorCode::FILE_NOT_FOUND);
+    if (!iov || iovcnt <= 0)
+        return make_error<size_t>(ErrorCode::FILE_INVALID_BUFFER);
+
+    const size_t total = IovecTotalLen(iov, iovcnt);
+    if (total == 0) return make_error<size_t>(ErrorCode::FILE_INVALID_BUFFER);
```
Aionw: If total is zero, which means write nothing, maybe here return no error is ok.

mooncake-store/src/uring_file.cpp
```diff
@@ -788,13 +788,100 @@ tl::expected<void, ErrorCode> UringFile::batch_read(ReadDesc* descs, int cnt) {
...
+        res =
+            SharedUringRing::instance().vector_write(fd_, iov, iovcnt, offset);
+    } else {
+        const off_t aligned_off = offset & ~static_cast<off_t>(ALIGNMENT_ - 1);
+        const size_t head = static_cast<size_t>(offset - aligned_off);
+        const size_t span = head + total;
+        const size_t aligned_len =
+            ((span + ALIGNMENT_ - 1) / ALIGNMENT_) * ALIGNMENT_;
+
+        void* bounce = alloc_aligned_buffer(aligned_len);
```
Aionw: `bounce` value always need free here, could we use std::unique_ptr to manage it automatically?

mooncake-store/src/uring_file.cpp
```diff
@@ -808,13 +895,40 @@ tl::expected<size_t, ErrorCode> UringFile::vector_read(const iovec* iov,
                                                        int iovcnt,
                                                        off_t offset) {
     if (fd_ < 0) return make_error<size_t>(ErrorCode::FILE_NOT_FOUND);
+    if (!iov || iovcnt <= 0)
+        return make_error<size_t>(ErrorCode::FILE_INVALID_BUFFER);
+
+    const size_t expected_bytes = IovecTotalLen(iov, iovcnt);
+    if (expected_bytes == 0)
```
Aionw: I think read empty buffer is ok here, no need to return error.

huangxiaobai-ydmy:
Thanks @Aionw — addressed the four comments:

1. Renamed `IovecRegionDirectReady` → `IsIovecRegionAligned`
2. Empty `vector_write` (`total == 0`) now returns success (`0`)
3. Bounce buffers are owned by `std::unique_ptr` with an aligned-buffer deleter
4. Empty `vector_read` (`expected_bytes == 0`) now returns success (`0`)

Happy to adjust further if needed.

Aionw (approved):
LGTM

fcczzz (approved):
Thanks for the update — the preservation-read error handling and regression test look good.

One non-blocking suggestion: since this implementation rounds direct I/O up to 4 KiB, could we require `capacity_` to be a multiple of 4 KiB when uring is enabled? Otherwise, a write reaching the end can extend the file beyond the configured capacity, causing recovery's exact file-size check to reject it.

The default capacity is already aligned, so validating this constraint would be a simple way to cover the edge case.

huangxiaobai-ydmy:
Thanks for the suggestion — agreed.

Added an Init-time check so that when uring/O_DIRECT is enabled, `capacity_` must be a multiple of 4 KiB; otherwise Init returns `INVALID_PARAMS`. Also added `OffsetAllocatorStorageBackend_RejectsUnalignedCapacityWithUring` to cover the rejection path.

ykwd (approved):

PR #3616, opened 2026-08-24
[Bugfix] Restore vLLM KV event replay after sequence gaps

Chase-Rong:
## Description

Fix Conductor KV event replay for vLLM publishers.

The previous implementation sent an incomplete DEALER request and treated replay as a single response frame. This was incompatible with the vLLM ROUTER multipart protocol, so missing events were not actually recovered.

  This change:

  - Implements the vLLM multipart replay request/response protocol.
  - Drains replay streams through the END_SEQ marker before dispatching events.
  - Replays missing events before the live event that exposes the gap.
  - Preserves failed replay ranges with their upper bounds to prevent duplicate
    dispatch during later retries or reconnects.
  - Recreates the DEALER socket after replay timeout or malformed responses.
  - Restricts replay support to vLLM subscriptions.
  - Updates English and Chinese Conductor documentation.
  - Adds regression coverage for live retry, reconnect retry, timeout handling,
    and duplicate prevention.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

- Manual C++ build of the ZMQ client test target.
  - ZMQ tests: 18/18 passed.
  - clang-format dry-run passed.
  - `git diff --check` passed.

**Test results:**
- [ ] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used
  AI tools were used during implementation and review. The human submitter must
  review and understand every changed line before submitting the PR.


catyans:
Thanks for taking this on — fixing the actual vLLM ROUTER/DEALER framing and draining through `END_SEQ` is a meaningful improvement, and the timeout/socket-reset coverage is useful.

There is one retry-ordering case that worries me. When the replay for a gap times out, `ProcessMessage()` still dispatches the live event that exposed the gap and advances `last_seq_`. On the next live message, the pending range is retried and its older events are dispatched before that new live message, but after the previous boundary event. The timeout test therefore permits an observed order like `10, 15, 11, 12, 13, 14, 16...20` — it checks that every sequence appears once, but not that delivery remains monotonic.

That can do more than leave the index temporarily incomplete: `EventHandler` applies Store/Remove mutations immediately and does not reject an event just because its transport sequence is lower than one already applied. A replayed old Remove can therefore undo a newer Store (or vice versa) after recovery appears to succeed.

Could we keep dispatch monotonic, for example by buffering the live boundary until its missing range is complete, or by adding sequence-aware stale-event suppression at the mutation boundary? A regression that records handler order and uses conflicting Store/Remove events would capture the actual state guarantee better than count-only assertions.

One housekeeping note: `dev/conductor` has moved forward and this head now conflicts with it, so it will need a rebase as well. Happy to take another look after the update.

Issue #3617, opened 2026-08-24
[Feature Request]: real_client and http service support version


mjwtom:
### Describe your feature request

Master already supports --version to show the version information.
We like real_client to support it and also able to get version from http service & metrics for deployment.

### Before submitting a new issue...

- [ ] Make sure you already searched for relevant issues and read the [documentation](https://kvcache-ai.github.io/Mooncake/)

rishabhsinha17:
Taking this one.

PR #3619, opened 2026-08-24
[Store] Add /version HTTP endpoint to master and real client (#3617)

mjwtom:
Expose build version information over HTTP so operators can verify which binary a running process is using without shell access to the host.

- Master admin server: new `GET /version` handler, registered alongside `/health` so it stays available in standby.
- Real client HTTP server: same `/version` payload on the client HTTP port; it does not touch `client_`, so it is unaffected by `MC_STORE_CLIENT_METRIC` and by client initialization state.
- `mooncake_client`: set the gflags version string so `--version` works, and log the display version at startup.

Both endpoints return `version` (used for RPC handshake compatibility) and `display_version` (release plus short git hash).

Docs: document the endpoint in the master admin endpoint table, the client endpoint table, the HTTP service reference, and the three client HTTP server switches that gate it.

Tests: add `/version` coverage for the real client HTTP server, and for the master admin server in both serving and standby states.

## Module

- [x] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [x] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
```

./mooncake_client --version
mooncake_client version 0.3.12.post1 (git: f9e8311f)

**Test results:**
- [x] Unit tests pass

curl xxx:port/version
{"version":"2.0.0","display_version":"0.3.12.post1 (git: f9e8311f)"}

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

### DIFF
--- a/docs/source/api-reference/http/http-service.md
+++ b/docs/source/api-reference/http/http-service.md
@@ -260,6 +260,29 @@ Basic health check endpoint for service availability verification.
 curl http://localhost:8080/health
 ```
 
+#### `/version`
+Report the master version. Always available, including while the master is in
+standby.
+
+**Method**: `GET`
+**Content-Type**: `application/json; charset=utf-8`
+**Response**: JSON object with:
+- `version` (string): Store version used for RPC handshake compatibility
+- `display_version` (string): Human-readable release plus short git hash
+
+**Example**:
+```bash
+curl http://localhost:8080/version
+```
+
+```json
+{"version":"2.0.0","display_version":"0.3.12.post1 (git: f9e8311f)"}
+```
+
+Real clients expose the same `/version` payload on their own client HTTP port
+when `enable_client_http_server` is on. See
+[Client Metrics Endpoint](../../getting_started/observability.md#client-metrics-endpoint).
+
 ## Store REST API Endpoints
 
 The following endpoints are served by the Python store REST service, which wraps
--- a/docs/source/api-reference/python/mooncake-store.md
+++ b/docs/source/api-reference/python/mooncake-store.md
@@ -1150,7 +1150,7 @@ the positional overload.
   data is stored under `MOONCAKE_DFS_ROOT_DIR`, but this separate directory is
   still validated during FileStorage initialization.
 - `tenant_id` (str): Tenant namespace for object keys. Defaults to `"default"`.
-- `enable_client_http_server` (bool): Enable the client-local `/health`, `/metrics`, and `/metrics/summary` HTTP endpoints. Defaults to `False`.
+- `enable_client_http_server` (bool): Enable the client-local `/health`, `/metrics`, `/metrics/summary`, and `/version` HTTP endpoints. Defaults to `False`.
 - `client_http_port` (int): Port for the client-local HTTP endpoints. Defaults to `9300`.
 
 **Store segment pinned memory:** CUDA-enabled builds can register Store-managed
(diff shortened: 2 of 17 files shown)


Icedcoco (approved):
LGTM. The PR cleanly adds `/version` to the master admin and real client HTTP servers, exposes build information through Prometheus metrics, and enables `mooncake_client --version`. The implementation is backward-compatible, focused, and covered by tests.

Non-blocking suggestion: add a `/version` assertion to the `MC_STORE_CLIENT_METRIC=0` test to confirm it remains available independently of client metrics. The documented version values are illustrative examples.

mjwtom:
> LGTM. The PR cleanly adds `/version` to the master admin and real client HTTP servers, exposes build information through Prometheus metrics, and enables `mooncake_client --version`. The implementation is backward-compatible, focused, and covered by tests.
> 
> Non-blocking suggestion: add a `/version` assertion to the `MC_STORE_CLIENT_METRIC=0` test to confirm it remains available independently of client metrics. The documented version values are illustrative examples.

Thank you for the advice. Patch updated.

rishabhsinha17:
I had #3674 open for the client half of this before spotting this PR, closed it in favor of this one. Two small notes from implementing the same surface independently:

- `ClientMetric::summary_metrics()`: a version line there would let `/metrics/summary` identify the build without a second request. That was the one piece of #3674 not present here. Happy to send it as a follow-up on top of this PR if useful.
- `RealClient::start_http_server` builds the `/version` JSON by string concatenation while the master side goes through `HttpVersionResponse` with `YLT_REFL`. Both values are compile-time constants so it works today; reusing the reflected struct on the client would keep the two payloads from drifting.

LGTM otherwise.

mjwtom:
> I had #3674 open for the client half of this before spotting this PR, closed it in favor of this one. Two small notes from implementing the same surface independently:
> 
> * `ClientMetric::summary_metrics()`: a version line there would let `/metrics/summary` identify the build without a second request. That was the one piece of [[Store] Add version support to real_client and its HTTP service #3674](https://github.com/kvcache-ai/Mooncake/pull/3674) not present here. Happy to send it as a follow-up on top of this PR if useful.
> * `RealClient::start_http_server` builds the `/version` JSON by string concatenation while the master side goes through `HttpVersionResponse` with `YLT_REFL`. Both values are compile-time constants so it works today; reusing the reflected struct on the client would keep the two payloads from drifting.
> 
> LGTM otherwise.

Good idea for ClientMetric::summary_metrics() and updated。

For Note 2 
Both `HttpVersionResponse` and `WriteJsonResponse` are file‑static private definitions inside `master_admin_service.cpp` (lines 486‑490). To share them, we would have to lift these definitions into public headers, which makes `real_client.cpp` depend on HTTP response types from the master admin server. Introducing this cross‑module coupling for a two‑field payload consisting entirely of compile‑time constants is not worthwhile.

More importantly, consistency considerations point the other way. All payloads in the client HTTP server are built via string concatenation; `/health` at `real_client.cpp:1871` follows exactly the same pattern. Converting only `/version` to use the reflected struct would make it the sole exception in this file.


thx

mjwtom (commented):
继续求review

ykwd (approved):

PR #3634, opened 2026-08-24
[Reshard] Bind logical transfer plans to runtime snapshots

Bo-Vincent:
## Description

This PR adds the late runtime-binding and physical-attestation layer for
heterogeneous model-weight resharding.

The preceding logical-planning layer produces an address-free
`LogicalTransferPlan`: it identifies the required source and target logical
regions, but deliberately has no GPU address, allocation, endpoint, lease, or
runtime-instance information. This PR consumes that plan together with typed
`WeightRuntimeBindingManifest` inputs and returns an attested `TransferPlan`.
The resulting plan contains selected runtime fragments and executor
projections for a later backend to consume.

Binding is fail-closed. It verifies source and target placement identity and
digest, revision and weight generation, participant selection, runtime
fragment geometry, device/address/allocation bounds, lease metadata, target
coverage, alias scope, and executor snapshots. It also rejects forged or
reconstructed plans whose physical evidence no longer matches the planned
placement.

The returned `TransferPlan` is a validated runtime snapshot, not an execution
API. 

Related RFC: #3111

## Module

- [ ] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [x] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**

```bash

bash scripts/check_reshard_types.sh

./scripts/code_format.sh --check --base origin/main

(cd docs && python3 -m sphinx -W -b html source build/html)

```

**Test results:**

- [x] Unit tests pass
  - `358 passed, 2 subtests passed in 5.22s`
- [x] Manual testing done

- Pyright positive suite: `0 errors`.
- Negative type fixtures reject invalid ID and parallel-axis usage as expected.
- Scoped pre-commit passed for all 24 changed files.
- Sphinx `-W` build passed for all 95 documentation sources.
- `git diff --check` passed.

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex was used for implementation assistance and verification support. The
human submitter is responsible for the final changes.


mooncake-reshard/python/mooncake/reshard/weight/_planner/bound_contracts.py
```diff
@@ -0,0 +1,530 @@
...
+        expected_key: ExecutorKey = (
+            binding.instance_id,
+            placement.placement_id,
+            binding.participant_id,
+            placement.digest,
+            binding.lease_id,
+            executor.worker_id,
+            executor.rank,
+        )
+        if key != expected_key:
```
zxpdemonio, line 381: Executor provenance is currently keyed only by instance/placement/participant/digest/lease/worker/rank, so it does not prove the physical snapshot of each operation fragment. Replacing an operation's runtime fragment address (or endpoint/device/nbytes/generation) while keeping those routing fields unchanged could pass validation. Please make operations and executors reference the same canonical binding map/digest and compare fragment_id, address, nbytes, worker, endpoint, device, generation, and backing allocation per fragment; add a negative regression where any field tampering fails closed.

mooncake-reshard/python/mooncake/reshard/weight/_planner/bound_contracts.py
```diff
@@ -0,0 +1,530 @@
...
+            raise ValueError("transfer plan mixes stored and live source fragments")
+        if executors:
+            raise ValueError("stored source must not have live executor provenance")
+        return
+    live_fragments = tuple(
+        fragment for fragment in fragments if isinstance(fragment, BoundWeightFragment)
+    )
+    if len(live_fragments) != len(fragments):
+        raise ValueError(f"transfer plan {side} fragment is not runtime-bound")
+    if not executors:
```
zxpdemonio, line 316: The live-source path returns when executors == (), unlike the target path, which requires an executor projection. This allows a TransferPlan containing BoundWeightFragment sources to be constructed after source executor provenance is dropped. Please require executor provenance for live sources; only an all-stored source may omit it, with a regression test.

mooncake-reshard/python/mooncake/reshard/weight/_planner/bound_contracts.py
```diff
@@ -0,0 +1,530 @@
...
+                raise ValueError("transfer plan exceeds max_segments_per_region")
+            total_lowered_segments += operation.segment_count
+            if total_lowered_segments > self.planning_limits.max_total_lowered_segments:
+                raise ValueError("transfer plan exceeds max_total_lowered_segments")
+        has_stored_source = any(
+            isinstance(operation.source, StoredFragment)
+            for operation in self.operations
+        )
+        if has_stored_source:
+            if self.source_manifest_identity is None:
```
zxpdemonio, line 509: TransferPlan currently stores only source_manifest_identity for stored sources, not the actual fragments selected by operations (or per-fragment digests). Replacing object_key/object_offset/nbytes after construction or deserialization could therefore pass identity validation and read the wrong Store object. Please retain selected stored snapshots (or canonical fragment digests), revalidate object key, offset, size, and geometry on post_init/pickle restore, and add a forged-fragment regression test.

mooncake-reshard/python/mooncake/reshard/weight/_planner/bound_contracts.py
```diff
@@ -0,0 +1,530 @@
...
+    PlanningLimits,
+    PipelineRouteGroup,
+    _is_canonical_operation,
+    _OperationViews,
+)
+from .fragments import BoundWeightFragment
+
+
+@dataclass(frozen=True)
+class RuntimeLeaseSnapshot:
```
zxpdemonio: RuntimeLeaseSnapshot currently stores only view-level address/nbytes and lease_generation, omitting storage_address/storage_nbytes/storage_offset_bytes. Resolve/restore cannot independently prove that an alias or subview still belongs to the original allocation. Please include allocation metadata (or a binding digest covering view, allocation, stride, and generation) and add negative tests for allocation replacement and out-of-bounds alias/subview.

mooncake-reshard/python/mooncake/reshard/weight/_planner/binding.py
```diff
@@ -0,0 +1,581 @@
...
+    source_binding_by_id = _validated_bindings_by_placement_id(
+        (
+            (logical_plan.source_placement,)
+            if logical_plan.source_placement is not None
+            else ()
+        ),
+        source_bindings,
+        "source",
+        source_required,
+        retain_unselected=True,
```
zxpdemonio: Using retain_unselected=True for source carries participants not referenced by the logical plan into the bound source projection (for example, source-dp1 remains when only source-dp0 is needed). Validate the complete input set if desired, but derive output source_executors only from referenced fragments to avoid extra lease snapshots, serialization, and scheduling overhead; add an assertion that unselected participants are absent.

mooncake-reshard/python/mooncake/reshard/weight/_planner/binding.py
```diff
@@ -0,0 +1,581 @@
...
+                        attestation=attestation,
+                    )
+                )
+    result.sort(
+        key=lambda item: (item.rank.dp, item.rank.pp, item.rank.ep, item.rank.tp)
+    )
+    return tuple(result)
+
+
+def bind_logical_transfer_plan(
```
zxpdemonio, line 427: Please provide reproducible performance data before we assess the overhead: report logical planning, runtime binding, TransferPlan construction, pickle serialization/restore latency, and peak memory, including fragment/operation counts, DP replicas, and TP/EP/PP sizes. Also quantify physical-validation cost for many segments, projection growth with retain_unselected=True, and behavior when PlanningLimits differs from max_segment_checks=1,000,000.

docs/source/design/model-weight-reshard-planner.md
```diff
@@ -72,15 +77,20 @@ For 3-D and higher logical boxes, exact intersection remains supported under an
...
+or `submit` operation and does not retain allocation lifetime. A later Transfer
+Engine executor must acquire real allocation guards and revalidate its bindings
+atomically with submission.
 
 Transfer Engine lowering, DMA submission, Store persistence/lifecycle, and
-framework activation remain outside this logical planner. Framework adapters
-own model semantics and conversion into canonical manifests; Mooncake core does
-not infer those semantics from framework objects or parameter names.
+framework activation remain outside this phase. Framework adapters own model
+semantics and conversion into canonical manifests; Mooncake core does not infer
```
zxpdemonio, line 106: Please make the with_parallelism boundary an explicit contract. This PR need not implement a Store adapter, but a future adapter must define Store key/metadata to canonical WeightManifest/WeightPlacementManifest mapping and reject snapshots that cannot be represented losslessly rather than guessing. In particular, reshard supports SplitAxis(kind="ep", dim=0), while Store validate_parallelism_spec() currently rejects EP split_dim and requires expert_id. Document whether Store encoding will be extended, support restricted, or an authoritative manifest supplied by the adapter; do not rely on mode="full" reconstruction.

zxpdemonio (changes requested):
Request changes: Most of the original comments are now implemented: live-source executor provenance, selected stored snapshots, backing-allocation metadata, participant projection, the with_parallelism boundary, and the core benchmark data. The remaining items are: (1) **Blocker:** the public TransferPlan constructor still accepts a coordinated forgery of both operations and source_stored_fragment_snapshots without checking the committed manifest content_sha256; please retain/validate the authoritative WeightManifest (or make this constructor internal) and add a regression; (2) **P2:** validate_runtime_bindings() accepts a RuntimeFragmentId reused across participants although the binding path rejects it; align the public contract and add a regression; (3) **P2:** a live TransferPlan retains process-local owner objects through RuntimeBindingAttestation, contradicting the documented allocation-lifetime boundary; please clarify or fix the ownership semantics; (4) add parameterized provenance tampering regressions for every attested field; and (5) extend the benchmark with retain_unselected projection inflation and PlanningLimits mismatch scenarios, or explicitly narrow the benchmark requirement.

mooncake-reshard/python/mooncake/reshard/weight/_planner/bound_contracts.py
Bo-Vincent: Fixed in d1321a8f. RuntimeLeaseSnapshot now captures the view stride plus backing-allocation address, size, and offset. During TransferPlan construction and restore, every operation-side BoundWeightFragment is compared directly with the corresponding executor lease snapshot rebuilt from the same attested binding. The check is symmetric for source and target. Regression tests now reject a forged operation address/allocation and a forged executor allocation snapshot.

mooncake-reshard/python/mooncake/reshard/weight/_planner/bound_contracts.py
Bo-Vincent: Fixed in d1321a8f. A live source now requires executor provenance. The only allowed empty source executor projection is an all-StoredFragment source. A public TransferPlan regression removes the source executor projection from a live plan and verifies that construction fails closed.

mooncake-reshard/python/mooncake/reshard/weight/_planner/bound_contracts.py
Bo-Vincent: Fixed in d1321a8f. A Store-backed TransferPlan now retains the canonical tuple of selected StoredFragment snapshots in addition to source_manifest_identity. __post_init__ and pickle restore require every Store operation source to match that selected snapshot exactly, including object key, offset, size, and logical geometry. A pickle-restore regression that forges object_key now fails closed.

mooncake-reshard/python/mooncake/reshard/weight/_planner/binding.py
Bo-Vincent: Fixed in d1321a8f. Full input bindings are still validated, but executor projections are now derived only from the runtime fragment IDs referenced by bound operations. resolve_executor_plans applies the same selection when comparing a current binding. The DP regression now verifies that an unselected source DP1 binding is absent from TransferPlan.source_executors.

mooncake-reshard/python/mooncake/reshard/weight/_planner/binding.py
Bo-Vincent: Fixed in f4ae5fc6. The PR now includes an opt-in reproducible benchmark:

  python mooncake-reshard/benchmarks/runtime_binding.py

It measures logical planning, runtime binding, TransferPlan revalidation, pickle dump/restore, binding peak memory, and a strided physical-validation workload. On H20, with 31 median samples for TP4/PP1/EP1/DP2 to TP8/PP2/EP2/DP1, 8 source fragments, 8 target fragments, 8 operations, 4 selected source executors, and 8 target executors, the results were: logical planning 0.762 ms, binding 1.375 ms, TransferPlan revalidation 0.393 ms, pickle dump 0.148 ms, pickle restore 0.968 ms, binding peak 32,016 bytes, and 54.701 ms for 128 regions / 16,384 physical segments.

The benchmark is explicitly CPU-side contract cost only; it does not measure Store, TE, G2G, or serving E2E throughput. The public bind and TransferPlan reconstruction paths now both use PlanningLimits.max_total_lowered_segments, rather than a separate fixed 1,000,000 segment budget.

docs/source/design/model-weight-reshard-planner.md
Bo-Vincent: Documented in d1321a8f. This phase does not accept Store keys or with_parallelism metadata as planner input. A future Store adapter must losslessly construct a complete WeightManifest or WeightPlacementManifest, including tensor descriptors, fragment geometry/object ranges, and TP/PP/EP/DP semantics; otherwise it must reject rather than infer from keys, names, ranks, or mode="full".

The document also records the current Store limitation: validate_parallelism_spec supports TP split_dim, but EP requires expert_id and rejects split_dim. Therefore Store-backed SplitAxis(kind="ep", dim=0) remains fail-closed until Store encoding is extended or the adapter supplies an additional authoritative canonical manifest.

mooncake-reshard/python/mooncake/reshard/weight/_planner/bound_contracts.py
Bo-Vincent: `RuntimeLeaseSnapshot` currently attests only the view, so an
alias/subview cannot prove that it still belongs to the original allocation.

I will add `storage_address`, `storage_nbytes`, and
`storage_offset_bytes` to the runtime binding evidence and include them in the
binding snapshot identity. Binding will validate that the view remains within
the declared allocation and reject allocation base/size/offset/generation
changes.

I will also add negative tests for allocation replacement and out-of-bounds
alias/subview cases.

mooncake-reshard/python/mooncake/reshard/weight/_planner/bound_contracts.py
```diff
@@ -0,0 +1,625 @@
...
+                    "stored transfer plan requires a source manifest identity"
+                )
+            if (
+                self.source_manifest_identity.resource_id != self.resource_id
+                or self.source_manifest_identity.revision != self.revision
+                or self.source_manifest_identity.weight_generation
+                != self.weight_generation
+            ):
+                raise ValueError("transfer plan source manifest identity differs")
+            _validate_stored_source_fragment_provenance(
```
zxpdemonio: **Blocker:** `TransferPlan` can still be forged by coordinating both `operations` and `source_stored_fragment_snapshots`. The constructor only checks that these two tuples agree and that resource/revision/generation match; it never ties them back to the committed manifest's `content_sha256`. I reproduced this by changing `object_key` in both values to an arbitrary payload key and calling construction; it succeeds. Please retain and validate the authoritative `WeightManifest` (or an equivalent canonical full-manifest commitment), or make this constructor internal, and add a regression that mutates both copies together.

mooncake-reshard/tests/model_weight_planner/test_bound_executor_snapshots.py
```diff
@@ -0,0 +1,966 @@
...
+    )
+    replacement_binding = replace(
+        fragment.attestation.binding,
+        fragments=(replacement_binding_fragment,),
+    )
+    replacement_attestation = RuntimeBindingAttestation(
+        fragment.attestation.placement,
+        replacement_binding,
+    )
+    forged_fragment = replace(
```
zxpdemonio: The implementation now compares the full runtime snapshot, but the regression coverage is still narrower than the contract requested in the original review. Current tests cover selected address/allocation changes plus a few lease/device/worker cases, but do not parameterize operation/executor tampering for `fragment_id`, `nbytes`, `endpoint`, `strides_bytes`, `lease_generation`, and every backing-range field. Please add a table-driven negative regression proving that each attested field fails closed on both operation and executor sides.

mooncake-reshard/benchmarks/runtime_binding.py
```diff
@@ -0,0 +1,343 @@
...
+    )
+    physical_regions = _physical_regions(
+        transfer_plan,
+        rows=args.physical_rows,
+        count=args.physical_regions,
+    )
+    physical_ms, _ = _median_ms(
+        lambda: _validate_target_physical_ranges(
+            physical_regions,
+            max_segment_checks=sum(region.segment_count for region in physical_regions),
```
zxpdemonio, line 309: The benchmark now reports the main contract costs and uses the configured segment budget, but it still does not exercise the two edge cases requested in the original review: projection inflation when unselected participants are retained, and behavior across differing `PlanningLimits`/segment-check budgets. Please add those scenarios (or explicitly document that they are out of scope); a CI threshold is not required, but the comparison should be reproducible.

mooncake-reshard/python/mooncake/reshard/weight/_planner/attestation.py
```diff
@@ -0,0 +1,103 @@
...
+    WeightPlacementManifest,
+    WeightRuntimeBindingManifest,
+)
+
+
+@dataclass(frozen=True)
+class RuntimeBindingAttestation:
+    """One validated binding of a placement participant to live storage."""
+
+    placement: WeightPlacementManifest
```
zxpdemonio, line 87: **P2:** `RuntimeBindingAttestation` retains the complete `WeightRuntimeBindingManifest`, including each `RuntimeBindingFragment.owner`. Because `owner` is a process-local allocation reference, a live `TransferPlan` strongly retains those owner objects through every attestation, while the design docs say a bound plan does not retain allocation lifetime. Please make the intended ownership contract explicit; if the plan must not hold allocation guards, detach owner from the attestation evidence while keeping the physical fields needed for validation.

zxpdemonio:
**P2:** `validate_runtime_bindings()` currently checks duplicate fragment IDs only within each participant binding. It accepts the same `RuntimeFragmentId` in two different participants, while `bind_logical_transfer_plan()` rejects that same input as a duplicate global runtime ID. This leaves the public validator and the binding boundary with inconsistent identity semantics. Please enforce global uniqueness in the public validator (or explicitly define participant-local ID scope) and add a cross-participant duplicate-ID regression.

mooncake-reshard/python/mooncake/reshard/weight/_planner/bound_contracts.py
Bo-Vincent: Fixed. Store plans now derive selected fragments from the committed WeightManifest. Coordinated operation/cache tampering is rejected on restore.

mooncake-reshard/tests/model_weight_planner/test_bound_executor_snapshots.py
Bo-Vincent: Added table-driven negative tests for every attested runtime field on both operation and executor projections.

mooncake-reshard/benchmarks/runtime_binding.py
Bo-Vincent: Added selected-vs-complete projection measurements and exact/off-by-one PlanningLimits checks.

mooncake-reshard/python/mooncake/reshard/weight/_planner/attestation.py
Bo-Vincent: Attestations now use owner-free RuntimeBindingEvidence. Restore rejects forged owners, and a regression verifies the plan does not retain allocation owners.

Bo-Vincent:
> **P2:** `validate_runtime_bindings()` currently checks duplicate fragment IDs only within each participant binding. It accepts the same `RuntimeFragmentId` in two different participants, while `bind_logical_transfer_plan()` rejects that same input as a duplicate global runtime ID. This leaves the public validator and the binding boundary with inconsistent identity semantics. Please enforce global uniqueness in the public validator (or explicitly define participant-local ID scope) and add a cross-participant duplicate-ID regression.

Added global runtime fragment_id uniqueness validation across participants, with a regression test.

zxpdemonio (approved):
LGTM.

stmatengss (approved):

PR #3638, opened 2026-08-24
fix(store): avoid P2P oplog sequence reuse after promotion

skygragon:
## Description

  Fix P2P HA promotion sequence initialization to avoid reusing OpLog sequence IDs when the promoted standby has not applied up to the latest persisted OpLog sequence.

  After promotion, the new primary now initializes its next sequence from `max(latest_applied_sequence_id, latest_oplog_sequence_id)`, so subsequent writes continue from `latest + 1` instead
  of potentially overwriting existing OpLog entries.

  ## Module

  - [ ] CI/CD
  - [ ] Docs
  - [ ] Other

  ## Type of Change

  - [x] Bug fix
  - [ ] New feature
  - [ ] Refactor
  - [ ] Breaking change
  - [ ] Documentation update
  - [ ] Performance improvement
  - [ ] Other

  ## How Has This Been Tested?

  Validated in local Docker container `mooncake-dev-p2p-promote-latest-sequence`.

  **Test commands:**
  ```bash
  clang-format-20 -style=file -i \
    mooncake-store/include/ha/oplog/p2p_hot_standby_service.h \
    mooncake-store/src/ha_helper.cpp \
    mooncake-store/tests/ha/oplog/p2p_hot_standby_service_test.cpp

  git diff --check

  cmake -S . -B build-codex-p2p -G Ninja \
    -DCMAKE_BUILD_TYPE=Debug \
    -DWITH_STORE=ON \
    -DWITH_TE=ON \
    -DWITH_P2P_STORE=OFF \
    -DBUILD_UNIT_TESTS=ON \
    -DBUILD_EXAMPLES=OFF \
    -DUSE_ETCD=OFF \
    -DSTORE_USE_ETCD=OFF \
    -DUSE_REDIS=OFF \
    -DSTORE_USE_REDIS=OFF \
    -DUSE_HTTP=OFF \
    -DUSE_CXL=OFF \
    -DUSE_CUDA=OFF \
    -DENABLE_ASAN=OFF

  cmake --build build-codex-p2p --target p2p_oplog_test -j2

  cd build-codex-p2p
  ./mooncake-store/tests/p2p_oplog_test \
    --gtest_filter=P2PHotStandbyServiceTest.RestorePromotedMetadataCanContinueFromStoreLatestSequence
```
  Test results:

  - [x] Unit tests pass

  ## Checklist

  ## AI Assistance Disclosure

  - [ ] No AI tools were used

  AI tools were used to inspect the code path, implement the fix, format the touched files, and run local build/test validation. The human submitter is responsible for reviewing and defending
  all changes.

### DIFF
--- a/mooncake-store/include/ha/oplog/p2p_hot_standby_service.h
+++ b/mooncake-store/include/ha/oplog/p2p_hot_standby_service.h
@@ -84,6 +84,7 @@ class P2PHotStandbyService {
     P2PStandbySyncStatus GetSyncStatus() const;
     bool IsReadyForPromotion() const;
     uint64_t GetLatestAppliedSequenceId() const;
+    ErrorCode GetLatestOpLogSequenceId(uint64_t& sequence_id) const;
 
     P2PStandbyMetadataStore::ExportedMetadata ExportMetadata() const;
     bool WaitForAppliedSequence(
@@ -115,7 +116,6 @@ class P2PHotStandbyService {
         uint64_t& baseline_sequence_id,
         const std::vector<std::string>& discovered_endpoints = {});
     ErrorCode ResyncFromSnapshotLocked();
-    ErrorCode GetLatestOpLogSequenceId(uint64_t& sequence_id) const;
     bool WaitForAppliedSequenceLocked(
         uint64_t sequence_id,
         std::chrono::milliseconds timeout = std::chrono::seconds(30),
--- a/mooncake-store/src/ha_helper.cpp
+++ b/mooncake-store/src/ha_helper.cpp
@@ -9,6 +9,7 @@
 #include "redis_master_view_helper.h"
 #endif
 
+#include <algorithm>
 #include <limits>
 #include <optional>
 #include <iomanip>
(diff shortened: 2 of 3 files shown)


he-yufeng (commented):
Read through the promotion path and this diff. The failure mode is real: readiness is gated on the applier's view, and the on-disk latest sequence can sit past that view, so numbering new writes from applied + 1 can collide with OpLog entries that are already persisted. Continuing from max(applied, latest) is the right minimal guard. Failing closed when the OpLog read fails also looks correct here, since this block only runs when p2p_standby exists; the no-OpLog HA promotion path from #3566 never reaches it.

One thing I would like spelled out: the exported metadata still reflects the applied horizon. If entries in (applied, latest] turn out to be real rather than a bare counter gap, their effects never reach the new primary while the numbering jumps over them. Is the intended invariant that IsReadyForPromotion already guarantees that range is empty, with the max() purely as insurance for the residual race? A short comment near the max() would save the next reader a trip through the state machine.

I did not run the new test locally; from reading, it exercises exactly the numbering continuation described above.

PR #3640, opened 2026-08-24
[Store] Add fenced snapshot publication with maintenance lease

Icedcoco:
## Description

Add the N04 control-plane primitives for publishing batch OpLog standby snapshots.

This PR introduces:

- a fixed 30-second snapshot maintenance lease backed by `EtcdHelper`;
- keepalive readiness checks and active lease revocation on release;
- an owner token that is unique to each maintenance attempt;
- atomic publication of the `latest` and `fallback` snapshot pointers;
- CAS fencing on the maintenance lock and the previously observed pointer bytes;
- monotonic publication based on `last_included_batch_id`;
- validation that a candidate belongs to the currently held maintenance lease.

The first publication creates only `latest`. Subsequent publications atomically perform:

```text
fallback = old latest
latest = candidate
```

The pointer values are the original descriptor JSON bytes returned by the N03 artifact writer. `producer_view_version` remains diagnostic and does not participate in candidate selection.

This PR does not add snapshot scheduling, restoration, OpLog batch deletion, or post-publication artifact cleanup.

The batch OpLog-specific publisher implementation and tests are kept under `ha/snapshot/batch_oplog/` and do not modify the legacy snapshot catalog/provider path.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

Coverage includes:

- first publication;
- atomic `latest`/`fallback` rotation;
- equal or stale batch rejection;
- maintenance owner fencing;
- candidate/lease owner mismatch;
- `latest` and `fallback` CAS races;
- publication from an older producer view when the cursor advances;
- corrupt fallback rejection;
- release and lost-lease behavior;
- real-etcd lease competition and revoke lifecycle when etcd is available.

**Test commands:**

```bash
GOCACHE=/tmp/mooncake-n04-gocache cmake --build build-n04 -j32

ctest --test-dir build-n04 --output-on-failure \
  -R 'batch_oplog_snapshot_(types|codec|writer)_test'

build-n04/mooncake-store/tests/batch_oplog_snapshot_publisher_test \
  --gtest_filter='-BatchOpLogSnapshotPublisherTest.RealEtcdLeaseAndPublisherLifecycle'

pre-commit run --files \
  mooncake-store/include/ha/snapshot/snapshot_maintenance_lease.h \
  mooncake-store/include/ha/snapshot/batch_oplog_snapshot_publisher.h \
  mooncake-store/include/ha/snapshot/batch_oplog/batch_oplog_snapshot_publisher.h \
  mooncake-store/src/ha/snapshot/snapshot_maintenance_lease.cpp \
  mooncake-store/src/ha/snapshot/batch_oplog/batch_oplog_snapshot_publisher.cpp \
  mooncake-store/tests/ha/snapshot/batch_oplog/publisher_test.cpp \
  mooncake-store/src/CMakeLists.txt \
  mooncake-store/tests/CMakeLists.txt
```

**Test results:**

- [x] Full `BUILD_UNIT_TESTS=ON`, `STORE_USE_ETCD=ON` build completed
- [x] Snapshot metadata, codec, and writer tests passed
- [x] All 8 fake-backend publisher tests passed
- [x] Pre-commit hooks passed
- [ ] Real-etcd lifecycle test passed locally

The real-etcd lifecycle test is included, but the local sandbox prevented connections to `127.0.0.1:2379`, so that case was skipped locally.

## Checklist

- [x] Documentation is not required for this internal protocol primitive
- [x] An existing standby snapshot and bounded OpLog retention RFC covers this work

## AI Assistance Disclosure

- [ ] No AI tools were used
- [x] AI tools were used

Codex assisted with implementation, focused test coverage, build troubleshooting, and preparation of this PR description. The submitter will review and validate all changed lines before submission.

### DIFF
--- a/mooncake-common/etcd/etcd_wrapper.go
+++ b/mooncake-common/etcd/etcd_wrapper.go
@@ -35,6 +35,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -42,6 +43,7 @@ import (
 
 	rpctypes "go.etcd.io/etcd/api/v3/v3rpc/rpctypes"
 	clientv3 "go.etcd.io/etcd/client/v3"
+	"go.etcd.io/etcd/client/v3/concurrency"
 )
 
 // prefixWatchInfo stores cancel function and callback context for a prefix watch
@@ -54,6 +56,44 @@ type prefixWatchInfo struct {
 	broken bool
 	// brokenNotified prevents duplicate WATCH_BROKEN callbacks from exit races.
 	brokenNotified bool
+}
+
+type maintenanceSession struct {
+	session *concurrency.Session
+	cancel  context.CancelFunc
+}
+
+func (s *maintenanceSession) close() error {
+	err := s.session.Close()
+	s.cancel()
+	return err
+}
+
+var startMaintenanceSession = func(ctx context.Context, cli *clientv3.Client,
+	ttl int) (*concurrency.Session, error) {
+	return concurrency.NewSession(cli, concurrency.WithTTL(ttl), concurrency.WithContext(ctx))
+}
+
+func newMaintenanceSession(cli *clientv3.Client, ttl int,
+	startupTimeout time.Duration) (*maintenanceSession, error) {
+	ctx, cancel := context.WithCancel(context.Background())
+	timer := time.AfterFunc(startupTimeout, cancel)
+	session, err := startMaintenanceSession(ctx, cli, ttl)
+	timedOut := !timer.Stop()
+	if err != nil {
+		cancel()
+		return nil, err
+	}
+	if session == nil {
+		cancel()
+		return nil, errors.New("maintenance session creation returned nil")
+	}
+	if timedOut {
+		_ = session.Close()
+		cancel()
+		return nil, context.DeadlineExceeded
+	}
+	return &maintenanceSession{session: session, cancel: cancel}, nil
 }
 
 // Use different etcd client so they are not affected by each other,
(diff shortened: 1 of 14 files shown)


Aionw:
Architectural suggestion, also addressing a lease cleanup race:

The current lifecycle is split across C++ and Go: grant a lease, start a C++ keepalive thread, wait for the Go-side cancellation entry, cancel it, and then join the thread. If `WaitKeepAliveReady` times out before the worker registers its cancellation entry, `CancelKeepAlive` returns `ETCD_OPERATION_ERROR`; `StopKeepAlive` masks that error and immediately joins. The worker can then register after the cancellation attempt and keep the lease alive indefinitely, so `Acquire`/`Release` can hang forever.

Since this repository already uses `go.etcd.io/etcd/client/v3` v3.5.21, I suggest keeping the lease lifecycle entirely in Go with `concurrency.Session`:

1. Create a session with `concurrency.NewSession(client, concurrency.WithTTL(30), concurrency.WithContext(ctx))`.
2. Preserve the current fail-fast semantics by acquiring the fixed maintenance key with a transaction: compare `CreateRevision(lockKey) == 0`, then put the owner token with `WithLease(session.Lease())`.
3. Return an opaque session handle, the lease ID, and the lock's create revision to C++.
4. Use `session.Done()` only as an advisory local liveness signal and `session.Close()` for release/revoke.

This removes the C++ keepalive thread, the custom readiness polling, and the cross-language cancellation registry. I would keep the fixed-key CAS rather than introduce `concurrency.Mutex` unless waiting/fairness is required; `Mutex.TryLock` is another valid option but changes the key layout.

For the publication fence, compare `CreateRevision(maintenanceKey)` with the revision returned when the lock was created (and optionally compare `LeaseValue` as well). That identifies the exact lock incarnation and is stronger than comparing only its value. Similarly, comparing the observed `ModRevision` of `latest` and `fallback` would avoid value-level ABA while keeping the final etcd transaction authoritative.

A deterministic test should delay keepalive/session startup and exercise cancellation during acquisition, since the current real-etcd lifecycle test does not cover the readiness-timeout path.

Aionw (approved):
LGTM. But we need to fix conflict.

ykwd (approved):

PR #3642, opened 2026-08-24
[Store] Add batch OpLog snapshot bootstrap

Icedcoco:
## Description

This PR implements PR-N05 of the standby snapshot and bounded OpLog retention RFC (#3167).

It adds a batch-OpLog-specific snapshot bootstrap path that:

- restores candidates in the fixed order: `latest`, `fallback`, then proven complete OpLog history;
- verifies pointer and object-store `descriptor.json` bytes are identical;
- validates manifest, segments, and object chunks using size, CRC32C, snapshot identity, cursor, chunk index, and object count;
- rejects duplicate objects, zero or duplicate per-object `ReplicaID` values, invalid tenant IDs, and inconsistent cursors;
- distinguishes invalid snapshot candidates from infrastructure failures, so backend timeouts and connection failures do not silently fall back;
- restores and replays the suffix in temporary standby state, installing it into `HotStandbyService` only after the full restore succeeds;
- seeds `OpLogBatchStandbyReader` with the snapshot batch cursor and requires batch 1 / sequence 1 when proving complete OpLog history.

The new provider and tests are isolated under `ha/snapshot/batch_oplog/`. The legacy snapshot provider, catalog fallback, scheduler, pruning, and production configuration are unchanged.

## Module

## Type of Change

- [x] New feature

## How Has This Been Tested?

**Test commands:**

```bash
cmake -S . -B /tmp/mooncake-n05-build \
  -DCMAKE_BUILD_TYPE=Debug \
  -DBUILD_UNIT_TESTS=ON \
  -DSTORE_USE_ETCD=ON \
  -DUSE_ETCD=OFF \
  -DUSE_HTTP=ON \
  -DUSE_UB=OFF \
  -DUSE_CUDA=OFF

cmake --build /tmp/mooncake-n05-build -j32

ctest --test-dir /tmp/mooncake-n05-build --output-on-failure \
  -R 'batch_oplog_snapshot_provider_test|oplog_batch_standby_reader_test|batch_oplog_snapshot_types_test|batch_oplog_snapshot_codec_test|batch_oplog_snapshot_writer_test|hot_standby_snapshot_bootstrap_test'
```

**Test results:**

- [x] Unit tests pass
- [ ] Manual testing done

The complete build finished successfully. All six N05 and adjacent snapshot /
standby test targets passed.

A full `ctest -j32` run completed with 107/141 targets passing. The remaining 34 failures require capabilities unavailable in the sandbox, mainly TCP/UDS socket binding, runtime service stacks, or in-source test-data paths. None of the N05-related targets failed.

## Checklist

- [x] Documentation is not applicable because production configuration and
      user-facing wiring are outside PR-N05
- [x] For changes >500 LOC: RFC #3167 covers this design

## AI Assistance Disclosure

- [ ] No AI tools were used

OpenAI Codex assisted with implementation, code inspection, test development, build validation, and drafting this PR description. The human submitter is responsible for reviewing and understanding every changed line before merge.

### DIFF
--- a/mooncake-store/include/ha/oplog/oplog_batch_standby_reader.h
+++ b/mooncake-store/include/ha/oplog/oplog_batch_standby_reader.h
@@ -34,6 +34,9 @@ class OpLogBatchStandbyReader {
                             OpLogApplier& applier);
 
     OpLogBatchStandbyPollResult PollOnce(size_t max_batches = 1024);
+    // Seed the reader after a materialized snapshot. The next poll starts at
+    // cursor.batch_id + 1 and validates the complete suffix.
+    ErrorCode SetBaselineCursor(const DurablePrefix& cursor);
     ErrorCode ReadProducerView(ViewVersionId& producer_view_version) const {
         return storage_.ReadProducerView(producer_view_version);
     }
@@ -49,6 +52,7 @@ class OpLogBatchStandbyReader {
     std::optional<DurablePrefix> last_applied_durable_prefix_;
     std::optional<uint64_t> last_scanned_batch_last_seq_;
     uint64_t last_applied_batch_id_{0};
+    bool require_complete_history_{false};
 };
 
 }  // namespace mooncake
--- a/mooncake-store/include/ha/snapshot/batch_oplog/batch_oplog_snapshot_provider.h
+++ b/mooncake-store/include/ha/snapshot/batch_oplog/batch_oplog_snapshot_provider.h
@@ -1,0 +1,43 @@
+#pragma once
+
+#include <cstdint>
+#include <string>
+
+#include <ylt/util/tl/expected.hpp>
+
+#include "ha/standby_metadata_store.h"
+#include "types.h"
+
+namespace mooncake {
+
+class HaKvBackend;
+class OpLogApplier;
+class SnapshotObjectStore;
+
+struct BatchOpLogSnapshotRestoreResult {
+    uint64_t last_included_seq{0};
+    uint64_t last_included_batch_id{0};
+    ViewVersionId producer_view_version{0};
+    ReplicaID max_replica_id{0};
+};
+
+// Restores the new batch-OpLog snapshot format into caller-owned temporary
+// state. Legacy catalog snapshots intentionally do not use this provider.
+class BatchOpLogSnapshotProvider final {
+   public:
+    BatchOpLogSnapshotProvider(std::string cluster_id, HaKvBackend& backend,
+                               SnapshotObjectStore& object_store,
+                               std::string snapshot_root);
+
+    tl::expected<BatchOpLogSnapshotRestoreResult, ErrorCode> RestoreBaseline(
+        StandbyMetadataStore& metadata, StandbySegmentRegistry& registry,
+        OpLogApplier* applier = nullptr);
+
+   private:
+    std::string cluster_id_;
+    HaKvBackend& backend_;
+    SnapshotObjectStore& object_store_;
+    std::string snapshot_root_;
+};
+
+}  // namespace mooncake
--- a/mooncake-store/include/ha/snapshot/batch_oplog/batch_oplog_snapshot_provider.h
+++ b/mooncake-store/include/ha/snapshot/batch_oplog/batch_oplog_snapshot_provider.h
@@ -17,6 +17,8 @@ class SnapshotObjectStore;
 struct BatchOpLogSnapshotRestoreResult {
     uint64_t last_included_seq{0};
     uint64_t last_included_batch_id{0};
+    uint64_t last_applied_seq{0};
+    uint64_t last_applied_batch_id{0};
     ViewVersionId producer_view_version{0};
     ReplicaID max_replica_id{0};
 };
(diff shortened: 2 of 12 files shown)


Aionw (changes requested):
Requesting changes for one fail-closed recovery issue.

**[P1] Complete-OpLog fallback must not initialize a missing durable prefix.**

`RestoreCompleteOpLog()` calls `OpLogBatchStorage::InitDurablePrefix()`. That API creates and persists `{batch_id: 0, last_seq: 0}` when both the prefix and batch records are absent. Consequently, after `latest` and `fallback` are unusable, a namespace whose recovery authority was deleted or corrupted is accepted as a legitimate empty cluster, and recovery mutates etcd while making that decision. This contradicts the RFC and the comment in `ReplaySuffix()` that a missing prefix cannot prove complete history, even for an empty cluster.

Please use a read-only validated-prefix path here. A missing durable prefix should fail with `INCOMPLETE_OPLOG_CATCH_UP`, while an existing zero prefix should still be accepted only after proving that no batch records exist. Please also add a regression test that starts with no prefix and verifies both that restore fails and that no prefix is written.

**Design suggestion:** could batch-OpLog bootstrap be the default behavior whenever batch OpLog mode is enabled, rather than requiring `SetBatchOpLogSnapshotProvider()`? The setter currently has no call sites, and injecting a concrete `final` provider is not a useful substitution seam. I would prefer automatically selecting `latest -> fallback -> complete OpLog` in the enabled mode, while injecting only the backend/object-store dependencies and keeping legacy catalog restore explicit.

Icedcoco:
Thanks for calling this out. I traced the empty-cluster HA startup path, and there is a constraint that makes the read-only change insufficient by itself.

On the first startup of an empty HA cluster, the supervisor starts the node as a standby before leadership is acquired. At that point `latest`, `fallback`, `durable_prefix`, and the batch namespace are all absent. The primary only calls `InitDurablePrefix()` after the standby has successfully started, acquired leadership, and been promoted. If N05 changes to `ReadDurablePrefix()` and rejects a missing key, the initial standby cannot start, so it cannot be promoted and the primary never gets a chance to create `{0,0}`.

There are therefore two states with the same observable contents in the current authority namespace:

```text
new empty cluster: pointers absent, prefix absent, batches absent
complete authority loss: pointers absent, prefix absent, batches absent
```

No read-only validation of those keys can distinguish them. The current `InitDurablePrefix()` path still fails closed when any batch record exists; the ambiguity is limited to the case where the prefix and all batch records are absent together.

A strict fail-closed solution would require an additional initialization protocol, such as an explicit provisioning step or a cluster-generation marker stored in an independent authority. A marker in the same namespace only reduces the probability because it can disappear with the prefix and batches. Adding that protocol would also need to change the supervisor/N08 startup contract, rather than only the N05 restore helper.

Given this constraint, we are considering keeping the current behavior as an explicit availability tradeoff: treat a completely empty OpLog authority as a new empty cluster, accepting the small risk that simultaneous loss of the prefix and every batch is interpreted as empty. Would you consider that acceptable for N05, or do you want this series to introduce the explicit initialization/independent-marker protocol before pointer bootstrap can merge?

Aionw (changes requested):
Follow-up after the discussion: I agree that the all-absent state must be accepted for first HA bootstrap under the current supervisor protocol, and I am not requesting default enablement as part of this PR.

[P1] Do not classify N-series snapshot control keys as legacy layout

RestoreCompleteOpLog() calls OpLogBatchStorage::InitDurablePrefix(), which first calls RejectLegacyLayout(). RejectLegacyLayout() rejects any key in /oplog/<cluster>/snapshot/, but the new latest and fallback pointer keys live in that same range. Therefore, if latest and fallback are present but both fail validation, recovery reaches the complete-OpLog fallback and then fails on the pointer keys before it can read the durable prefix or replay any batches. This makes the advertised latest -> fallback -> complete OpLog recovery order unavailable in an important corruption case.

Please make legacy-layout detection allow the known N-series control keys, or use a recovery-side validation path that does not reject them. Please also add a regression test where latest and fallback are unusable while the durable prefix and complete batch history are valid; RestoreBaseline() should recover by replaying the complete OpLog. An additional all-absent namespace test would document and preserve the intentional empty-cluster availability tradeoff.

Icedcoco:
Addressed this follow-up.

`RejectLegacyLayout()` now allows the four known N-series snapshot control keys: `latest`, `fallback`, `maintenance`, and `compaction_floor`. Unknown keys under `/oplog/<cluster>/snapshot/` are still treated as legacy sidecars and fail closed.

This keeps the intended recovery order working when both `latest` and `fallback` are present but invalid: the provider can now proceed to complete batch OpLog replay. Added coverage for the allowed control-key set, legacy-key rejection, the intentional all-absent empty-cluster bootstrap behavior, and complete replay after both pointers are invalid.

Validation: the affected storage/provider tests and the adjacent snapshot/standby test set pass; pre-commit passes. The all-absent namespace behavior remains the previously discussed availability tradeoff required by the current first-HA-bootstrap protocol.

Aionw (approved):
LGTM

ykwd (approved):

PR #3643, opened 2026-08-24
[Store] Break down batch get latency into metadata and data phases

justadogistaken:
## Description

`batch_get_into_multi_buffers` always performs two distinct phases internally, with no branch that can skip either one:

```
Step 1: client_->BatchQuery(keys)   → RPC to master, resolve replica addresses
Step 2: client_->BatchGet(...)      → RDMA read / disk read of the actual data
```

Today both phases are only visible as a single mixed `mooncake_transfer_batch_get_latency` value. Any caller that wraps the call in its own timer — for example a vLLM KV connector computing `load_get_avg_ms` and `throughput_mib_s` — therefore reports **metadata RPC time plus data transfer time combined**, and a slow load cannot be attributed to the master round-trip or to the data path.

This is not a hypothetical concern. On a deployment where the master sits across a high-RTT link, the reported end-to-end throughput was ~7-8 GB/s while `ib_write_bw` on the same hosts reached 196 Gbps, with no way to explain the gap from the exported metrics. After splitting the phases, the breakdown was:

| Component | Latency | Share |
|---|---|---|
| `BatchQuery` (master RPC) | ~45 ms | **67%** |
| `BatchGet` (data transfer) | ~22 ms | 33% |
| Total | ~67 ms | |

With ~600 MB moved per call, the data phase alone works out to ~27 GB/s (~220 Gbps), i.e. the data path was already close to line rate and the bottleneck was entirely the master RTT. The mixed metric made the deployment look like an RDMA problem when it was a master-placement problem.

Worth noting for context: `MasterService::BatchGetReplicaList` is a pure in-memory shard lookup under a shared lock (Redis is only a persistence backend and is not on the query path), so the metadata cost here is network RTT, not master-side work. Also, the existing `lookup_exists` metric uses a different RPC (`BatchExistKey`) at a different point in the request lifecycle, so it cannot be used to infer or subtract the metadata cost inside `load_get`.

### Changes

Instrument `batch_get_into_multi_buffers_internal` to time both phases and report them separately.

**New client metric histograms** (exported through the existing `/metrics` and `/metrics/summary` client endpoints, so no new plumbing is required to consume them):

| Metric | Description |
|---|---|
| `mooncake_transfer_batch_get_metadata_latency` | Time spent in `BatchQuery` (master RPC) |
| `mooncake_transfer_batch_get_data_latency` | Time spent in the data transfer (RDMA / disk read) |

The data phase timer covers both `BatchGet` paths — memory/NOF replicas and the LOCAL_DISK / DISK / DFS read paths — so the breakdown stays meaningful regardless of where the replica lives. It is deliberately named `data` rather than `rdma` for the same reason: the phase is not RDMA-only.

**New accessor for per-call attribution.** Cumulative histograms answer "is the cluster healthy", but a connector reporting per-request latency needs the breakdown of the call it just made. `RealClient::getLastBatchGetTimings()` returns the breakdown of the most recent call under a lock, exposed to Python as:

```python
store.batch_get_into_multi_buffers(keys, all_buffer_ptrs, all_sizes)
timings = store.get_last_batch_get_timings()
# {'metadata_us': 45200, 'data_us': 21800, 'total_us': 67000}
```

This lets a downstream KV-cache connector report metadata and data latency as two independent metrics instead of one mixed `load_get`, which is what turned the opaque 7-8 GB/s number above into an actionable "move the master closer" conclusion.

Docs: added a "Batch Get Latency Breakdown" section to `docs/source/getting_started/observability.md` covering both metrics and the Python accessor.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
client_metrics_test --gtest_filter='*BatchGetBreakdown*'
  [  PASSED  ] 1 test.

pybind_client_test --gtest_filter='*TestBatchPutAndGetMultiBuffers*'
  [  PASSED  ] 1 test.  (3036 ms, including in-proc master start&stop)

client_metrics_test
  [  PASSED  ] 17 tests.

```

```
(Worker_TP2 pid=4890) INFO 08-24-2026-20:52:48.224 [worker.py:1069] MooncakeStore transfer complete: req_id=p0-c8834ca47827-t2 kv_role=kv_both direction=load_from_store_pool status=ok blocks=39 block_ids=[22, 23, 24, 3073, 52, 1017, 1016, 1015, 586, 2362, 3248, 3249, 3250, 3251, 2281, 2282, 658, 659, 1276, 1277, 1278, 2897, 2896, 2895, 2894, 1530, 1529, 1528, 415, 416, 417, 418, 419, 420, 421, 1791, 1792, 2348, 2347] failed_blocks=0 failed_block_ids=[] requested_bytes=63897600 loaded_bytes=63897600 duration_ms=23.381 throughput_mib_s=2606.263 token_len=3968 batch_index=1/1 metadata_ms=2.538 transfer_ms=20.763
(Worker_TP3 pid=4891) INFO 08-24-2026-20:52:48.227 [worker.py:1069] MooncakeStore transfer complete: req_id=p0-c8834ca47827-t2 kv_role=kv_both direction=load_from_store_pool status=ok blocks=39 block_ids=[23, 24, 3073, 52, 1017, 1016, 1015, 586, 2362, 3248, 3249, 3250, 3251, 2281, 2282, 658, 659, 1276, 1277, 1278, 2897, 2896, 2895, 2894, 1530, 1529, 1528, 415, 416, 417, 418, 419, 420, 421, 1791, 1792, 2348, 2347, 22] failed_blocks=0 failed_block_ids=[] requested_bytes=63897600 loaded_bytes=63897600 duration_ms=25.578 throughput_mib_s=2382.401 token_len=3968 batch_index=1/1 metadata_ms=2.096 transfer_ms=23.393
```

**Test results:**
- [ ] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

### DIFF
--- a/docs/source/getting_started/observability.md
+++ b/docs/source/getting_started/observability.md
@@ -208,3 +208,26 @@ curl http://<client-host>:9300/metrics/summary
 Set `MC_STORE_CLIENT_METRIC=0` to disable client metric collection. If the
 client HTTP server remains enabled while metrics are disabled, `/metrics` and
 `/metrics/summary` return HTTP 503 with `metrics not available`.
+
+### Batch Get Latency Breakdown
+
+A batch get spends its time in two distinct phases: querying object metadata
+from the master, and transferring the data over RDMA or reading it back from
+disk. Both phases are exported separately so that a slow load can be attributed
+without correlating several mixed metrics.
+
+| Metric | Description |
+|--------|-------------|
+| `mooncake_transfer_batch_get_latency` | End-to-end batch get latency |
+| `mooncake_transfer_batch_get_metadata_latency` | Time spent in the metadata query (master RPC) |
+| `mooncake_transfer_batch_get_data_latency` | Time spent in the data transfer (RDMA / disk read) |
+
+The same breakdown for the most recent call is also available from Python,
+which lets a KV-cache connector report per-request metadata and data latency
+separately instead of a single mixed load metric:
+
+```python
+store.batch_get_into_multi_buffers(keys, all_buffer_ptrs, all_sizes)
+timings = store.get_last_batch_get_timings()
+# {'metadata_us': 412, 'data_us': 1893, 'total_us': 2340}
+```
--- a/docs/source/getting_started/observability.md
+++ b/docs/source/getting_started/observability.md
@@ -221,13 +221,3 @@ without correlating several mixed metrics.
 | `mooncake_transfer_batch_get_latency` | End-to-end batch get latency |
 | `mooncake_transfer_batch_get_metadata_latency` | Time spent in the metadata query (master RPC) |
 | `mooncake_transfer_batch_get_data_latency` | Time spent in the data transfer (RDMA / disk read) |
-
-The same breakdown for the most recent call is also available from Python,
-which lets a KV-cache connector report per-request metadata and data latency
-separately instead of a single mixed load metric:
-
-```python
-store.batch_get_into_multi_buffers(keys, all_buffer_ptrs, all_sizes)
-timings = store.get_last_batch_get_timings()
-# {'metadata_us': 412, 'data_us': 1893, 'total_us': 2340}
-```
--- a/mooncake-integration/store/store_py.cpp
+++ b/mooncake-integration/store/store_py.cpp
@@ -3027,29 +3027,6 @@ PYBIND11_MODULE(store, m) {
             "multiple "
             "keys")
         .def(
-            "get_last_batch_get_timings",
-            [](MooncakeStorePyWrapper &self) -> py::dict {
-                py::dict result;
-                result["metadata_us"] = 0;
-                result["data_us"] = 0;
-                result["total_us"] = 0;
-
-                auto real_client = self.get_real_client();
-                if (!real_client) {
-                    LOG(ERROR)
-                        << "get_last_batch_get_timings requires RealClient";
-                    return result;
-                }
-                const auto timings = real_client->getLastBatchGetTimings();
-                result["metadata_us"] = timings.metadata_us;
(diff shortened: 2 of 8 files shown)


Aionw (commented):
Thanks for the PR.

We indeed lack the ability to break down a `batch_get`/`batch_put` call and understand where the time is spent. However, I don't think exposing a public API for retrieving the timing breakdown is the right abstraction.

A more common approach would be to use distributed tracing, such as OpenTelemetry. That said, introducing tracing into the hot path may have a non-trivial performance impact, so we need to be careful about how it is implemented.

The current implementation already demonstrates this trade-off: in order to expose the timing of the most recent batch operation, every `batch_get` call needs to update shared timing state and acquire a mutex. I don't think this is an acceptable trade-off for a performance-sensitive path, especially for an observability feature.

I would prefer that we explore a tracing-based approach, ideally one that can be disabled or sampled with near-zero overhead when tracing is not enabled.

justadogistaken:
@Aionw Thanks for your feedback. agreed on both points.
**On the API abstraction**: Agreed that a public getter for per-call retrieval isn't the right way. I'm not pursuing a tracing-based per-call API in this PR either — that needs its own design (sampling, hot-path cost, etc.); happy to open a separate issue to track it.

**On the mutex/shared state**: Removed last_batch_get_timings_ / its mutex / RecordBatchGetTimings() along with it.

PR #1974, opened 2026-04-25
[TransferEngine] Bandwidth-aware NIC selection + DMA-BUF and IPv6 binding fixes

Nishikant-K-P:
## Description

Two-commit series. Replaces the legacy `getPciDistance` GPU↔NIC selection
heuristic with NCCL-style PCI ancestor-chain walking and bandwidth-aware
ranking, plus 23 hardware-free unit tests for the new logic. Each commit
is self-contained and bisectable.

> **Note:** This PR was originally a four-commit series. Per reviewer
> request to split, the IPv6 RPC binding fix moved to #2014 and the
> DMA-BUF CUDA device index fix moved to #2015. This PR is now scoped to
> the topology rewrite + tests, with all reviewer feedback applied.

1. **`[TE]` Replace `getPciDistance` with NCCL-style PCI topology and
   bandwidth-aware NIC selection.**
   The old metric was a string-prefix-of-sysfs heuristic that broke on
   GB200/GB300 (GPUs and NICs on separate PCI domains share only
   `/sys/devices/pci000` as common prefix → flat artifact distances → only
   1 NIC per NUMA selected as preferred). Replaced with NCCL-style PCI
   ancestor-chain walking + path-type classification (PIX / PXB / PHB /
   NODE / SYS) and a 3-level NIC ranking (path type → hops → PCIe bandwidth
   from `current_link_speed * current_link_width`). The bandwidth tiebreaker
   auto-filters lower-bandwidth front-end NICs to `avail_hca` without
   requiring `MC_CUSTOM_TOPO_JSON` or a manual blacklist. Pure-logic helpers
   (`PciPathType`, `InfinibandDevice`, `normalizePciBusId`,
   `classifyGpuNicPath`, `selectPreferredHcas`) moved to
   `include/pci_topology.h` for testability and shared use between
   `discoverCudaTopology` and the unit tests.

2. **`[TE]` Add 23 PCI topology unit tests.**
   Hardware-free tests in `pci_topology_test.cpp` covering
   `normalizePciBusId` (5), `classifyGpuNicPath` (10 — all 5 path types
   plus edge cases), NIC selection ranking via the production
   `selectPreferredHcas` helper (6), and fallback paths (2).

### Reviewer feedback addressed

- Selection ranking extracted into `selectPreferredHcas()` in
  `pci_topology.h`; tests now exercise production code instead of a
  duplicate copy.
- HCA ancestor chains pre-computed once before the GPU loop in
[...]

PR #3644, opened 2026-08-24
[TENT] Derive per-NIC bandwidth from the negotiated port speed

SongOf:
Description

  DeviceSelector stamped every RDMA device with a constant 400 Gbps and nothing ever overwrote it, so the "theoretical bandwidth" was 50 GB/s for every NIC. That value seeds the EWMA the smart selector scores with and bounds it to [0.1x, 10x] = [5, 500] GB/s regardless of the real link. A 25G NIC (~3.1 GB/s measured) sits below that floor, so the clamp pins its EWMA at 5 GB/s and it never learns its real rate; on hosts mixing link speeds the selector keeps steering slices at the slow NIC. The same number feeds getAggregateEwmaBandwidth(), used by the admission queue's deadline-drop predictor. The default/min/max_bandwidth_gbps config knobs were read into SchedulingParams but never consulted.

  - RdmaContext::openDevice keeps active_speed/active_width from the ibv_query_port it already issues; new linkSpeedGbps(). XDR (encoding 256) overflows the legacy uint8_t field, so CMake probes for ibv_port_attr::active_speed_ex (HAVE_IBV_ACTIVE_SPEED_EX) and prefers it when present.
  - New DeviceSelector::setDeviceBandwidth(): validates the speed against the configured [min, max], falls back to default_bandwidth_gbps with a warning when unknown (0) or out of range, and re-seeds the EWMA. theoreticalBandwidth() now reads SchedulingParams, so the config knobs take effect. The kDefaultBwGbps/kMinBwGbps/kMaxBwGbps constants are dropped; SchedulingParams is the single source and Workers uses its defaults as the config fallback.
  - Workers seeds each device from its context status: a context that never constructed (`DEVICE_UNINIT`/`DISABLED`) is marked **unavailable** and gets no bandwidth at all; a port down at open (`DEVICE_PAUSED`) is seeded but unavailable until `PORT_ACTIVE`; `DEVICE_ENABLED` is seeded from the negotiated speed. `DeviceSelector::setDeviceAvailable()` excludes a device from candidate construction (both passes, both modes) and from the aggregate bandwidth the admission queue's deadline predictor reads; all-unavailable yields -1, which switches the drop predictor off. `handleContextEvents` flips availability on `PORT_ERR`/`DEVICE_FATAL`/`PORT_ACTIVE`. The configured default is applied only to a usable NIC whose speed could not be read. (review: failed contexts were left at 400G in the aggregate and stayed selection candidates)
  - The link speed is re-read on `IBV_EVENT_PORT_ACTIVE` and, behind a CMake probe (`HAVE_IBV_EVENT_DEVICE_SPEED_CHANGE`, rdma-core >= 62), on `IBV_EVENT_DEVICE_SPEED_CHANGE`; `RdmaContext::refreshPortAttributes()` re-issues `ibv_query_port`, and `Workers::refreshLinkSpeed()` re-seeds through `setDeviceBandwidth()` only when the speed actually changed, before the device becomes selectable again. (review: a link returning at a lower speed kept the old seed and clamp)
  - The speed/width decoding table is lifted from show_links.cpp into mooncake-common/include/ib_link_speed.h so classic TE and TENT share one copy. The classic RdmaContext now also prefers `ibv_port_attr::active_speed_ex` (same `HAVE_IBV_ACTIVE_SPEED_EX` probe, in `src/CMakeLists.txt` for the classic tree), so show_links reports 800 Gbps for XDR 4x instead of 0. (review)
  - Port events are matched against the port the context opened: a device's async fd delivers `PORT_ERR`/`PORT_ACTIVE` for every port of that device, so on a multi-port HCA a flap on an unused port no longer pauses the context or drops it from selection. The decision half of `handleContextEvents` is now `applyContextEvent()`, so a synthesized event can drive it in tests. (review)
  - `getNicLoadStats()` reports only NICs able to carry traffic, the same rule as the aggregate; the C struct is unchanged (C ABI, consumed by the Rust binding via FFI). (review)

  No public API or wire format changes. Not related to #1974 (classic TE GPU↔NIC PCI ranking).

  Module

  - [ ] Reshard (mooncake-reshard)
  - [x] Common (mooncake-common)
  - [ ] CI/CD
  - [x] Docs
  - [ ] Other

  Type of Change

  - [x] Bug fix
  - [ ] New feature
  - [ ] Refactor
  - [ ] Breaking change
  - [x] Documentation update
  - [ ] Performance improvement
  - [ ] Other

  How Has This Been Tested?

  Test commands:
  cd build
  cmake --build . --target ib_link_speed_test tent_device_selector_test tent_rdma_transport_test tent_qos_contract_test bw_arbitration_test rdma_context_reprobe_test
  ctest -R 'ib_link_speed_test|tent_device_selector_test|tent_rdma_transport_test|tent_qos_contract_test|bw_arbitration_test|rdma_context_reprobe_test' --output-on-failure

  Test results:
  - [x] Unit tests pass — 100% tests passed, 0 tests failed out of 6 (Ubuntu 22.04, rdma-core 39, no RNIC)

  New tests, both written first and observed failing on the unpatched tree:
  - ib_link_speed_test — SDR…XDR × 1x/2x/4x/8x/12x encoding table; unknown encodings report 0.
  - tent_device_selector_test — link speed seeds the EWMA; unknown / out-of-range speed falls back to the configured default; unknown device is rejected; a 25G link learns 3.1 GB/s instead of being pinned at 5 GB/s; the clamp ceiling and floor scale with the link speed; re-seeding after a 400G→25G change releases the stale clamp; unavailable devices leave the aggregate and are never allocated (smart and round-robin), all-unavailable yields DeviceNotFound and a negative aggregate, and the priority fallback offers only available devices.
  - tent_rdma_transport_test — a context whose construct() failed is unavailable to the selector after Workers construction and contributes nothing to the aggregate (hardware-free via RdmaTransportTestPeer); refreshPortAttributes() on an inert context is rejected and leaves the speed at 0; synthesized `PORT_ERR`/`PORT_ACTIVE` for the opened port flip availability and the aggregate, events for another port are ignored in both directions, `DEVICE_FATAL` applies regardless of port, `CQ_ERR` leaves availability alone.
  - tent_device_selector_test — unavailable devices leave getNicLoadStats() and return when restored.
  - rdma_context_reprobe_test (classic, interposed ibv_query_port) — the negotiated speed/width are recorded (HDR 4x → 200 Gbps); with `active_speed_ex` = 256 and `active_speed` = 0, activeSpeed() is 256 and the shared table yields 800 Gbps. The XDR case is skipped where the header lacks the field (CI's rdma-core 39); it was run RED→GREEN against a verbs.h copy with the field added.

  Manual: the HAVE_IBV_ACTIVE_SPEED_EX branch was compile-checked against a copy of verbs.h with the field added, and the CMake probe confirmed to report yes/no against headers with/without it; the HAVE_IBV_EVENT_DEVICE_SPEED_CHANGE branch was syntax-checked with the enumerator injected (CI's rdma-core 39 predates it). Not covered without hardware: openDevice populating the fields from a real port, and the PORT_ERR/PORT_ACTIVE/DEVICE_SPEED_CHANGE events reaching handleContextEvents (the last also needs rdma-core >= 62 + Linux 7.0 + mlx5).

  Checklist
  - [ ] For changes >500 LOC: I have filed an RFC issue (N/A, ~370 LOC incl. tests)

  AI Assistance Disclosure

  - [ ] No AI tools were used

### DIFF
--- a/mooncake-common/include/ib_link_speed.h
+++ b/mooncake-common/include/ib_link_speed.h
@@ -1,0 +1,75 @@
+// Copyright 2025 KVCache.AI
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef MOONCAKE_IB_LINK_SPEED_H_
+#define MOONCAKE_IB_LINK_SPEED_H_
+
+namespace mooncake {
+
+// Per-lane signalling rate for an ibv_port_attr::active_speed encoding
+// (see ibv_query_port(3)). Returns 0 for an encoding that is not known.
+inline double ibLaneSpeedGbps(int active_speed) {
+    switch (active_speed) {
+        case 1:
+            return 2.5;  // SDR
+        case 2:
+            return 5.0;  // DDR
+        case 4:
+            return 10.0;  // QDR
+        case 8:
+            return 10.0;  // FDR10
+        case 16:
+            return 14.0;  // FDR
+        case 32:
+            return 25.0;  // EDR
+        case 64:
+            return 50.0;  // HDR
+        case 128:
+            return 100.0;  // NDR
+        case 256:
+            return 200.0;  // XDR
+        default:
+            return 0.0;
+    }
+}
+
+// Lane count for an ibv_port_attr::active_width encoding. Returns 0 for an
+// encoding that is not known.
+inline int ibLinkWidthLanes(int active_width) {
+    switch (active_width) {
+        case 1:
+            return 1;
+        case 2:
+            return 4;
+        case 4:
+            return 8;
+        case 8:
+            return 12;
+        case 16:
+            return 2;
+        default:
+            return 0;
+    }
+}
+
+// Link speed in Gbps from the raw ibv_port_attr encodings, or 0 when either
+// encoding is unknown so the caller can fall back explicitly rather than
+// scheduling against a guessed rate.
+inline double ibLinkSpeedGbps(int active_speed, int active_width) {
+    return ibLaneSpeedGbps(active_speed) * ibLinkWidthLanes(active_width);
+}
+
+}  // namespace mooncake
+
+#endif  // MOONCAKE_IB_LINK_SPEED_H_
--- a/mooncake-transfer-engine/src/show_links.cpp
+++ b/mooncake-transfer-engine/src/show_links.cpp
@@ -16,7 +16,8 @@
 
 #include <json/json.h>
 
-#include <algorithm>
+#include "ib_link_speed.h"
+
 #include <sstream>
 
 #include "transfer_engine_impl.h"
@@ -64,64 +65,9 @@ static bool hasTransportDetails(const NicDiagInfo& nic) {
     return nic.gid_index >= 0;
 }
 
+// Truncated to whole Gbps so the diagnostic output keeps its integer shape.
 static int computeSpeedGbps(int active_speed, int active_width) {
-    if (active_width <= 0) return 0;
-
-    int lane_gbps = 0;
-    switch (active_speed) {
-        case 1:
-            lane_gbps = 2;
-            break;
-        case 2:
-            lane_gbps = 5;
-            break;
-        case 4:
-            lane_gbps = 10;
-            break;
-        case 8:
-            lane_gbps = 10;
-            break;
-        case 16:
-            lane_gbps = 14;
-            break;
-        case 32:
-            lane_gbps = 25;
-            break;
-        case 64:
-            lane_gbps = 50;
-            break;
-        case 128:
-            lane_gbps = 100;
-            break;
-        case 256:
-            lane_gbps = 200;
-            break;
-        default:
-            lane_gbps = std::max(0, active_speed);
-            break;
-    }
-    int lanes = 1;
-    switch (active_width) {
-        case 1:
-            lanes = 1;
-            break;
-        case 2:
-            lanes = 4;
-            break;
-        case 4:
-            lanes = 8;
-            break;
-        case 8:
-            lanes = 12;
-            break;
-        case 16:
-            lanes = 2;
-            break;
-        default:
-            lanes = 1;
-            break;
-    }
-    return lane_gbps * lanes;
+    return static_cast<int>(ibLinkSpeedGbps(active_speed, active_width));
 }
 
 std::string buildShowLinksJson(TransferEngineImpl* impl) {
--- a/mooncake-transfer-engine/tent/src/transport/rdma/quota.cpp
+++ b/mooncake-transfer-engine/tent/src/transport/rdma/quota.cpp
@@ -391,6 +391,10 @@ Status DeviceSelector::getNicLoadStats(std::vector<NicLoadStats>& stats) const {
     // devices_ is populated during topology load and remains stable while
     // transfers update the per-device atomic counters below.
     for (const auto& [dev_id, dev] : devices_) {
+        // Same rule as the aggregate: a NIC that cannot carry traffic has
+        // no meaningful load or bandwidth to report, and with zero inflight
+        // it would score as the best NIC of the lot.
+        if (!dev.available.load(std::memory_order_relaxed)) continue;
         std::string device_name = local_topology_->getNicName(dev_id);
         if (device_name.empty()) device_name = std::to_string(dev_id);
         stats.push_back(NicLoadStats{
--- a/mooncake-transfer-engine/tent/src/transport/rdma/quota.cpp
+++ b/mooncake-transfer-engine/tent/src/transport/rdma/quota.cpp
@@ -18,6 +18,8 @@
 #include "tent/common/utils/random.h"
 #include "tent/common/utils/os.h"
 
+#include <glog/logging.h>
+
 #include <algorithm>
 #include <iostream>
 #include <iomanip>
@@ -31,13 +33,39 @@ Status DeviceSelector::loadTopology(std::shared_ptr<Topology>& local_topology) {
         if (!entry || entry->type != Topology::NIC_RDMA) continue;
         DeviceInfo& info = devices_[dev_id];
         info.dev_id = dev_id;
-        info.bw_gbps = kDefaultBwGbps;
+        info.bw_gbps = 0.0;  // unknown until setDeviceBandwidth
         info.numa_id = entry->numa_node;
-        info.ewma_bandwidth_bps.store(info.getTheoreticalBandwidth(),
+        info.ewma_bandwidth_bps.store(theoreticalBandwidth(info),
                                       std::memory_order_relaxed);
     }
     // Initialize device base priorities after all devices are loaded
     fillDevicePriorities();
+    return Status::OK();
+}
+
+double DeviceSelector::theoreticalBandwidth(const DeviceInfo& dev) const {
+    const auto& p = sched_params_;
+    double gbps = dev.bw_gbps;
+    if (gbps < p.min_bandwidth_gbps || gbps > p.max_bandwidth_gbps)
+        gbps = p.default_bandwidth_gbps;
+    return gbps * 1e9 / 8.0;
+}
+
+Status DeviceSelector::setDeviceBandwidth(int dev_id, double gbps) {
+    auto it = devices_.find(dev_id);
+    if (it == devices_.end())
+        return Status::InvalidArgument("device not found");
+    auto& dev = it->second;
+    const auto& p = sched_params_;
+    if (gbps < p.min_bandwidth_gbps || gbps > p.max_bandwidth_gbps) {
+        LOG(WARNING) << "Device " << local_topology_->getNicName(dev_id)
+                     << " link speed " << gbps << " Gbps is "
+                     << (gbps <= 0.0 ? "unknown" : "outside the valid range")
+                     << ", assuming " << p.default_bandwidth_gbps << " Gbps";
+    }
+    dev.bw_gbps = gbps;
+    dev.ewma_bandwidth_bps.store(theoreticalBandwidth(dev),
+                                 std::memory_order_relaxed);
     return Status::OK();
 }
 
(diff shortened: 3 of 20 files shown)


he-yufeng (commented):
Read the full diff and checked the encodings against the kernel enums (ib_port_speed / ib_port_width in include/rdma/ib_verbs.h): both tables match, including the easy-to-miss IB_WIDTH_2X = 16, and handling the XDR overflow of the u8 active_speed via active_speed_ex is the right call. The premise also checks out on current main: bw_gbps is stamped kDefaultBwGbps in loadTopology and nothing ever overwrites it, so the EWMA clamp really is [5, 500] GB/s for every NIC today.

Two things I verified beyond the tables: setDeviceBandwidth re-seeds the EWMA, not just the stored rate, and the inert-slot path keeps the configured default without double-warning. Consolidating the show_links table also removes its guessy fallbacks (max(0, active_speed), lanes = 1).

I did not build or run the new test locally (no RDMA stack on this machine), so this stays a comment, but from a read-through this is a good fix.

mooncake-common/include/ib_link_speed.h
```diff
@@ -0,0 +1,75 @@
...
+            return 25.0;  // EDR
+        case 64:
+            return 50.0;  // HDR
+        case 128:
+            return 100.0;  // NDR
+        case 256:
+            return 200.0;  // XDR
+        default:
+            return 0.0;
+    }
```
alogfans: This is not enough: we have 400 or even 800 Gbps RNICs.

mooncake-transfer-engine/tent/src/transport/rdma/workers.cpp
alogfans, line 1: `active_speed_` and `active_width_` are sampled only once in `openDevice()`. The existing recovery path resumes the context on `IBV_EVENT_PORT_ACTIVE`, but it does not re-query the port or update `DeviceSelector`; `IBV_EVENT_DEVICE_SPEED_CHANGE` is not handled either.

If a link comes back at a different negotiated speed—for example, 400G falling back to 100G—the EWMA seed and clamp will remain based on the old speed. In particular, a large stale EWMA floor can reproduce the exact slow-link selection problem this PR is intended to fix.

Could we refresh the port attributes and call `setDeviceBandwidth()` on `PORT_ACTIVE` and `DEVICE_SPEED_CHANGE`?

mooncake-transfer-engine/tent/src/transport/rdma/workers.cpp
```diff
@@ -179,14 +179,27 @@ Workers::Workers(RdmaTransport* transport)
...
 
+    // Seed each device's bandwidth from the speed its port negotiated.
+    // context_set_ is indexed by NicID, the same id the selector uses; an
+    // inert slot (NIC skipped at init) has no port to read, so leave it on
+    // the configured default without logging a second warning for it.
+    for (size_t dev_id = 0; dev_id < transport_->context_set_.size();
+         ++dev_id) {
+        const auto& context = transport_->context_set_[dev_id];
+        if (!context || context->status() == RdmaContext::DEVICE_UNINIT)
+            continue;
```
alogfans: An inert context is different from an enabled NIC whose speed is unknown. `loadTopology()` has already inserted every RDMA NIC into `devices_`, and `getAggregateEwmaBandwidth()` sums every entry. Skipping an uninitialized context here therefore leaves that NIC's existing bandwidth seed in the aggregate even though it cannot carry traffic.

This can overestimate the capacity used by the admission/deadline predictor; depending on the device mask, the NIC may also remain a scheduling candidate. Please explicitly mark failed or skipped contexts unavailable and exclude them from candidate construction and aggregate bandwidth calculations. The configured default should apply only to usable NICs whose speed cannot be determined.

SongOf: @alogfans Thanks for your review. Confirmed and fixed. `loadTopology()` filters on `NIC_RDMA` only, so a NIC whose `construct()` failed stayed in `devices_` at the 400G default; the seed loop's "skip inert" left it there, and `buildCandidates` had no status check — a dead NIC with zero inflight and a 50 GB/s seed actually scored *best* and got picked until `max_retry_count`. The aggregate over-counted by the same amount, and a port down at open (`DEVICE_PAUSED`) had the same shape.

Fixed in 89cfb6a0: `DeviceSelector` gains a per-device `available` flag (atomic, runtime-settable). Unavailable devices are excluded from candidate construction (both passes — the priority fallback never relaxes availability — and the round-robin path) and from the aggregate; all-unavailable yields -1, which switches the admission drop off. `Workers` seeds per context status: `DEVICE_UNINIT`/`DISABLED` → unavailable with no bandwidth at all, `DEVICE_PAUSED` → seeded but unavailable until `PORT_ACTIVE`, `ENABLED` → seeded and available; `handleContextEvents` now receives the NicID and flips the flag on `PORT_ERR`/`DEVICE_FATAL`/`PORT_ACTIVE`. The configured default is thereby only ever applied to a usable NIC whose speed could not be read. `getNicLoadStats` keeps listing unavailable NICs so operators can see them.

Tests (each observed failing first): exclusion from allocation in smart and round-robin mode, from the aggregate, all-unavailable, the priority-filter fallback, and — hardware-free via `RdmaTransportTestPeer` — a failed context being unavailable to the selector after `Workers` construction with a negative aggregate.

mooncake-transfer-engine/tent/src/transport/rdma/workers.cpp
SongOf: @alogfans  Agreed, and the second half of the suggestion is stronger than I first read it: `IBV_EVENT_DEVICE_SPEED_CHANGE` (rdma-core ≥ 62 / Linux 7.0, currently raised by mlx5) is a device-level event that fires when a port's effective speed changes *without* a link flap — e.g. a VF over LAG losing a PF — which `PORT_ACTIVE` cannot observe.

Fixed in 0f271d5d: `RdmaContext::refreshPortAttributes()` re-issues `ibv_query_port` on the open device (the speed/width decode is shared with `openDevice()` via `recordPortSpeed()`). `Workers::refreshLinkSpeed()` calls it and re-seeds through `setDeviceBandwidth()` only when `linkSpeedGbps()` actually changed, so a link returning at the same speed keeps what it learned. `handleContextEvents` calls it on `PORT_ACTIVE` — before the device is marked available again, so no worker scores it on the old rate — and on `DEVICE_SPEED_CHANGE` behind a CMake probe (`HAVE_IBV_EVENT_DEVICE_SPEED_CHANGE`, same pattern as `HAVE_IBV_ACTIVE_SPEED_EX`); it is an enumerator, so older rdma-core compiles without that branch. `active_speed_`/`active_width_` and `bw_gbps` become atomic since the monitor thread now writes them under worker reads.

The selector re-seed is unit-tested (EWMA learned at 400G, then `setDeviceBandwidth(25)` resets it and a 3.1 GB/s link converges instead of sitting on the stale 5 GB/s floor), and `refreshPortAttributes()` on an inert context is rejected. The speed-change branch was syntax-checked with the enumerator injected (CI's rdma-core 39 predates it); exercising the events themselves needs an RNIC. `ibv_query_port_speed()` for LAG-effective speed can follow separately, resolved as an optional symbol through `IbvLoader`.

mooncake-transfer-engine/tent/src/transport/rdma/quota.cpp
staryxchen, line 389: getNicLoadStats() still lists unavailable devices with their meaningless 400G seed, inconsistent with the aggregate the admission queue reads.

mooncake-transfer-engine/src/show_links.cpp
```diff
@@ -64,64 +65,9 @@ static bool hasTransportDetails(const NicDiagInfo& nic) {
     return nic.gid_index >= 0;
 }
 
+// Truncated to whole Gbps so the diagnostic output keeps its integer shape.
```
staryxchen, line 68: the legacy (non-TENT) path stores uint8_t active_speed, so the XDR encoding (256) wraps to 0. Will  show_links report 0 Gbps on XDR hardware even though the new shared table supports it?

mooncake-transfer-engine/tent/src/transport/rdma/workers.cpp
```diff
@@ -815,15 +847,44 @@ int Workers::handleContextEvents(std::shared_ptr<RdmaContext>& context) {
...
+        // Out of selection and out of the aggregate until the port returns;
+        // its EWMA cannot learn anything while no traffic flows.
+        device_selector_->setDeviceAvailable(dev_id, false);
         LOG(WARNING) << "Action: " << context->name() << " down";
     } else if (event.event_type == IBV_EVENT_PORT_ACTIVE) {
         context->resume();
+        // The link may have renegotiated while down: re-seed before the
+        // device becomes selectable so no worker scores it on the old rate.
+        refreshLinkSpeed(dev_id, *context);
+        device_selector_->setDeviceAvailable(dev_id, true);
```
staryxchen: PORT_ERR/PORT_ACTIVE handlers ignore event.element.port. Async events are device-scoped, so on a dual-port HCA used as two rails, a link flap on one port wrongly marks the other rail's device unavailable and drops it from the aggregate bandwidth. The pause()/resume() half is pre-existing, but the PR extends it to scheduling — cheap fix is to compare event.element.port against context->portNum().

SongOf: Confirmed — the event's `element.port_num` was never compared, so on a single-device multi-port HCA a flap on a port this process does not use paused the context (pre-existing) and, since 89cfb6a0, also dropped it from selection and the aggregate. Fixed in 8d3bae68: `PORT_ERR`/`PORT_ACTIVE` are matched against `portNum()` and ignored (INFO log) for any other port; `DEVICE_FATAL` and `DEVICE_SPEED_CHANGE` stay device-scoped. On mlx5 with one device per port this is a no-op. The decision is split out as `applyContextEvent()` so synthesized events drive it hardware-free: flips for the opened port, ignored for another port in both directions, `DEVICE_FATAL` regardless of port, `CQ_ERR` untouched — each observed failing first.

mooncake-transfer-engine/tent/src/transport/rdma/quota.cpp
SongOf: Agreed, and my "so operators can see them" rationale did not hold: the listed numbers (zero inflight, the never-corrected 400G seed) describe nothing, and since this snapshot is the load signal for #2516's replica scoring (#2996), a dead NIC would have looked like the emptiest, fastest one — the mis-selection 89cfb6a0 removed from slice placement, one layer up. Fixed in 2a801581: `getNicLoadStats()` skips unavailable devices, the same rule as the aggregate; the contract is documented on `NicLoadStats` and the C struct. I did not add an availability field because `tent_nic_load_stat`/`nic_load_stat` are C ABI and consumed through FFI by the Rust binding. Test: two NICs → two entries, one marked unavailable → the other only, restored → both.

mooncake-transfer-engine/src/show_links.cpp
SongOf: Yes — it did before this PR too (the old table's `case 256` was equally unreachable through the uint8_t field), but lifting the table to a shared header while only wiring `active_speed_ex` into TENT left the classic half claiming XDR support it could not deliver. Fixed in e50c92e8: the classic `RdmaContext` prefers `active_speed_ex` under the same `HAVE_IBV_ACTIVE_SPEED_EX` probe (added to `src/CMakeLists.txt` for the classic tree; TENT keeps its own, the cache variable is shared), so `show_links` reports 800 Gbps for XDR 4x. Tested through the classic interposed `ibv_query_port` in `rdma_context_reprobe_test`: HDR 4x → 200 Gbps on any host; the XDR case asserts `activeSpeed() == 256` and 800 Gbps, is skipped where the header lacks the field (CI's rdma-core 39), and was run RED→GREEN here against a verbs.h copy with the field added.

staryxchen (approved):
LGTM

PR #3646, opened 2026-08-24
[PG] drop support for torch 2.11.0

UNIDY2002:


he-yufeng:
What motivates dropping 2.11.0 from the matrix? Removing it here ends wheel builds for anyone pinned to torch 2.11, so the PR needs a reason (build breakage, EOL policy, wheel size). If it is just cleanup, I would keep it until there is a support-window decision.

UNIDY2002:
Okay, let's not drop it for now. I'm coming up with a JIT strategy.

PR #3647, opened 2026-08-25
[TENT] Preserve per-request RDMA selection policy

waizuichougou:
## Description

`Transport::SubBatch` carries one RDMA `device_mask` and one `qp_pool`, but a
single public submission can contain multiple requests that all select RDMA
with different policy results. `commitPreparedSubmit()` previously grouped
only by transport and copied the first task's policy to the whole RDMA
SubBatch. Later requests could therefore use the wrong NIC mask or QP pool.

This is reachable through the native TENT APIs: `tent::Request` has per-request
`priority`, `policy_name`, and `intent_type`
(`tent/include/tent/common/types.h:132-152`), the Python binding exposes them
(`tent/src/python/pybind.cpp:346-382`), and selection copies them into a
per-request `SelectionContext`
(`tent/src/runtime/transfer_engine_impl.cpp:1134-1139`) before matching the
policy (`tent/src/runtime/transport_selector.cpp:390-449`). A single native
TENT batch can therefore legitimately resolve different RDMA masks or QP pools.

This change:

- keeps requests with different selection or scheduling attributes separate
  during request merging;
- groups RDMA physical owners by resolved `(device_mask, qp_pool)` before each
  transport submission;
- assigns physical SubBatch task IDs from the actual size before each policy
  group is appended, so a synchronous failure cannot shift later groups, and
  preserves IDs for merged public aliases;
- copies the selected device mask into each `RdmaTask`, then uses it for lazy
  allocation and fallback RNIC selection while preserving request priority;
- refreshes RDMA policy metadata on queued dispatch and transport failover;
- keeps the existing three-argument `DeviceSelector::allocate()` overload and
  does not change the public `SubBatch` API.

The SubBatch API remains scalar by design. A per-request mask vector is not
needed because every `submitTransferTasks()` call is homogeneous, and the
effective mask is copied into each transport task before the SubBatch is reused
for another policy group.

This is a follow-up to the cached device-mask resolution in
[#3490](https://github.com/kvcache-ai/Mooncake/pull/3490), and is complementary
to the GPUDirect reachability and fallback handling in
[#3281](https://github.com/kvcache-ai/Mooncake/pull/3281).

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**

```bash
cmake -S . -B build-tent \
  -DUSE_TENT=ON -DUSE_CUDA=OFF \
  -DWITH_STORE=OFF -DWITH_STORE_RUST=OFF \
  -DBUILD_UNIT_TESTS=ON -DCMAKE_BUILD_TYPE=Release

cmake --build build-tent --target \
  tent_rdma_transport_test \
  tent_transport_hint_test \
  request_merge_test \
  tent_metrics_recording_test \
  tent_engine_failover_e2e_test \
  tent_runtime_queue_dispatch_test -j
ctest --test-dir build-tent/mooncake-transfer-engine/tent/tests \
  --output-on-failure \
  -R '^(tent_rdma_transport_test|tent_transport_hint_test|request_merge_test|tent_metrics_recording_test|tent_engine_failover_e2e_test|tent_runtime_queue_dispatch_test)$'
```

**Test results:**

- [x] Unit tests pass

The six focused and adjacent regression test suites above passed. They cover
mixed-policy batching, synchronous submit failure, request merging, transport
attempt metrics, failover, and runtime-queue dispatch.

All applicable pre-commit hooks passed, including C++ formatting and codespell.

## Checklist

- [x] The C/C++ formatting pre-commit hook passed
- [x] Documentation is unchanged because this is an internal bug fix
- [x] An RFC is not required because this is a small bug fix

## AI Assistance Disclosure

- [ ] No AI tools were used

OpenAI Codex assisted with the implementation, testing, and writing of this
PR.

### DIFF
--- a/mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
+++ b/mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
@@ -1276,13 +1276,12 @@ MergeResult mergeRequests(const std::vector<Request>& requests,
         if (last.req.transport_hint != curr.req.transport_hint) {
             return false;
         }
-        // These fields affect transport selection or scheduling. Merging
+        // These fields affect transport selection. Merging
         // requests with different values would silently apply the first
         // request's policy to the whole combined transfer.
         if (last.req.priority != curr.req.priority ||
             last.req.policy_name != curr.req.policy_name ||
-            last.req.intent_type != curr.req.intent_type ||
-            last.req.deadline_ns != curr.req.deadline_ns) {
+            last.req.intent_type != curr.req.intent_type) {
             return false;
         }
         if (!last.boundary.source_key || !curr.boundary.source_key ||
@@ -1568,10 +1567,6 @@ Status TransferEngineImpl::commitPreparedSubmit(
             "batch public task capacity exceeded" LOC_MARK);
     }
 
-    std::vector<size_t> task_id_list[kSupportedTransportTypes];
-    // task_id_list also contains derived public tasks created by request
-    // merging. Keep a parallel list of physical owners so requests can be
-    // grouped with the policy that will be copied into their transport task.
     std::vector<size_t> physical_task_id_list[kSupportedTransportTypes];
     std::unordered_map<size_t, TaskInfo> merged_task_id_map;
 
@@ -1591,7 +1586,6 @@ Status TransferEngineImpl::commitPreparedSubmit(
             task = merged_task_id_map[merged_task_id];
             task.derived = true;
             if (task.type != UNSPEC) {
-                task_id_list[task.type].push_back(task_id);
                 auto owner_it =
                     owner_task_id_by_merged_task.find(merged_task_id);
                 if (owner_it != owner_task_id_by_merged_task.end()) {
@@ -1653,7 +1647,6 @@ Status TransferEngineImpl::commitPreparedSubmit(
 
         task.sub_task_id = -1;
         task.derived = false;
-        task_id_list[task.type].push_back(task_id);
         physical_task_id_list[task.type].push_back(task_id);
         owner_task_id_by_merged_task[merged_task_id] = task_id;
         public_tasks_by_physical_owner[task_id].push_back(task_id);
--- a/mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
+++ b/mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
@@ -25,6 +25,7 @@
 #include <random>
 #include <stdexcept>
 #include <thread>
+#include <unordered_set>
 
 #include "tent/common/config.h"
 #include "tent/common/status.h"
@@ -1275,6 +1276,15 @@ MergeResult mergeRequests(const std::vector<Request>& requests,
         if (last.req.transport_hint != curr.req.transport_hint) {
             return false;
         }
+        // These fields affect transport selection or scheduling. Merging
+        // requests with different values would silently apply the first
+        // request's policy to the whole combined transfer.
+        if (last.req.priority != curr.req.priority ||
+            last.req.policy_name != curr.req.policy_name ||
+            last.req.intent_type != curr.req.intent_type ||
+            last.req.deadline_ns != curr.req.deadline_ns) {
+            return false;
+        }
         if (!last.boundary.source_key || !curr.boundary.source_key ||
             !last.boundary.target_key || !curr.boundary.target_key) {
             return false;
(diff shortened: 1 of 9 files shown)


he-yufeng (commented):
Read the rework through the submit path. The invariants hold as far as I traced: mergeRequests now refuses to merge across priority/policy_name/intent_type/deadline differences, the RDMA branch groups physical owners by (device_mask, qp_pool) so each SubBatch submit is homogeneous, the SubBatch scalars are re-set per group before submitTransferTasks, and RdmaTask copies them out before return, so sharing one SubBatch across groups is safe. The sub_task_id assignment resolving from the current SubBatch size per group also closes the gap a failed earlier group would otherwise leave, and resubmitTransferTask got the same treatment.

One thing I did not do: build and run the RDMA tests locally (no RDMA stack here), so this stays a comment rather than an approval. From a read-through the fix is sound.

mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
```diff
@@ -1575,7 +1590,15 @@ Status TransferEngineImpl::commitPreparedSubmit(
         if (merged_task_id_map.count(merged_task_id)) {
             task = merged_task_id_map[merged_task_id];
             task.derived = true;
-            if (task.type != UNSPEC) task_id_list[task.type].push_back(task_id);
+            if (task.type != UNSPEC) {
+                task_id_list[task.type].push_back(task_id);
```
staryxchen: task_id_list  is written and never read.

mooncake-transfer-engine/tent/tests/transport_hint_test.cpp
```diff
@@ -346,6 +357,134 @@ TEST(TransportHint, MixedHintsInBatchSplitAcrossTransports) {
...
+    ASSERT_TRUE(engine.registerLocalMemory(buf.data(), kBufLen).ok());
+
+    Request request_a0 = makeLocalWriteRequest(buf.data(), kRequestLength);
+    request_a0.policy_name = "rdma-a";
+    Request request_b =
+        makeLocalWriteRequest(buf.data() + kRequestLength, kRequestLength);
+    request_b.policy_name = "rdma-b";
+    Request request_a1 =
+        makeLocalWriteRequest(buf.data() + 2 * kRequestLength, kRequestLength);
+    request_a1.policy_name = "rdma-a";
```
staryxchen: All three requests use same length, so getTransferStatus would still pass if public IDs were mapped to the wrong SubBatch slots.

mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
```diff
@@ -1275,6 +1276,15 @@ MergeResult mergeRequests(const std::vector<Request>& requests,
...
         }
+        // These fields affect transport selection or scheduling. Merging
+        // requests with different values would silently apply the first
+        // request's policy to the whole combined transfer.
+        if (last.req.priority != curr.req.priority ||
+            last.req.policy_name != curr.req.policy_name ||
+            last.req.intent_type != curr.req.intent_type ||
+            last.req.deadline_ns != curr.req.deadline_ns) {
+            return false;
+        }
```
staryxchen: These fields affect transport selection, so adjacent requests must not be fused across priority / policy_name / intent_type. deadline_ns does not belong in that set.

Request::deadline_ns is documented as observability-only (RFC #2519): it does not drive admission, routing, or QP/NIC selection. A merged request already keeps the first request's metadata, so different deadlines only change the MLU numerator/denominator for that combined transfer. Splitting on this field will drop merge opportunities and create extra RDMA submits without fixing the bug this PR is about.

Please drop deadline_ns from can_merge (and from the new merge tests) unless the intent is specifically to preserve per-request MLU, in which case call that out—it is a metrics choice, not a selection-policy fix.

mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
waizuichougou: Thanks for catching this. I removed the unused task_id_list and both remaining writes to it.

mooncake-transfer-engine/tent/tests/transport_hint_test.cpp
waizuichougou: I changed the A/B/A requests to use distinct lengths (1024, 2048, and 3072) and now assert transferred_bytes for every public task ID. This makes the test fail if the B and A1 public IDs are mapped to the wrong SubBatch slots.

mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
waizuichougou: I removed deadline_ns from can_merge and from the new merge tests, and updated the related comments and test names.

waizuichougou:
Thank you @staryxchen for the careful and constructive review. I'm new to contributing to open-source projects, so I really appreciate the concrete feedback. It helped me tighten both the implementation and the tests.

I've addressed all three comments in commit 6b4844f9. Thank you again!

staryxchen (approved):
LGTM, will merge after CI pass

PR #3648, opened 2026-08-25
[TENT] Remove global lock serialization on per-batch hot paths

staryxchen:
## Description

**Problem.** TENT throughput does not scale with caller threads. In `tebench` (4KiB/1 write), the op rate is flat at ~140k ops/s from 4 to 16 threads while per-op latency grows linearly with thread count (55us at 8 threads, 113us at 16) — the classic M/D/1 signature of a single serialized server. Under the same workload the classic engine reaches 474k ops/s at 8 threads with flat latency. The engine effectively degrades to one core no matter how many threads submit, poll, or free batches.

**Root cause.** `TransferEngineImpl::progress_mutex_` serializes every per-batch operation: `submitTransfer`, `getTransferStatus`, `freeBatch`. Sharding that mutex alone is not sufficient — the batch registry (`alive_batches_` / active set) sits behind a single global mutex taken on *every poll*, which reintroduces the same serialization (measured at ~5.9k cycles per poll at 8 threads with rdtsc section counters).

**Solution.** Give each batch a complete lock domain: 64 `BatchShard`s keyed by BatchID, each owning both its recursive mutex *and* its slice of the batch registry.

- Poll / free / retain lock only the batch's shard. Membership checks become O(1) set lookups under the already-held shard lock — no global lock remains on the hot path.
- Lock order is `progress_mutex_ -> shard`, never reversed. With `enable_runtime_queue=true`, hot paths stay on `progress_mutex_` (queue state is engine-global) and take the shard only for membership ops.
- `freeBatch` fast path (runtime queue off, no refs, tasks terminal) frees inline under the shard and holds it through deallocation, so a concurrent poller can never observe a dangling `Batch*`. A poll error defers the batch to the lazy sweep (retry / quarantine) instead of failing the call, preserving the semantics of the recent reclaim rework on main.

**Results.** All numbers below were measured the same day, on the same machine pair, with the same `tebench` parameters and `TENT_METRICS_ENABLED=OFF` at build time; "before" is `main` without this change, "classic" is the classic engine from the same revision.

Local, RDMA loopback (ops/s):

| Threads | 1 | 2 | 4 | 8 | 16 |
|---|---|---|---|---|---|
| before | 112k | 142k | 161k | 145k | 142k |
| **after** | 112k | 170k | 329k | **464k** | **673k** |
| classic | 104k | 117k | 280k | 474k | 975k |

Cross-machine RDMA (ops/s):

| Threads | 1 | 2 | 4 | 8 | 16 |
|---|---|---|---|---|---|
| before | 62k | 87k | 118k | 135k | 138k |
| **after** | 61k | 94k | 190k | **322k** | **512k** |
| classic | 65k | 120k | 242k | 470k | 852k |

Single-thread latency is unchanged (9.0us local / 16.2us cross). At 8 threads the fix delivers **3.2x local / 2.4x cross-machine** and reaches classic parity locally (464k vs 474k); 8-thread latency drops from 55us to 17.3us. 256KiB transfers sustain 52.6 GB/s at 8 threads. A further gap to classic remains at 16 threads and cross-machine (60–70%); that residual is in transport-level per-op overhead and the completion model, not the engine lock domain, and is left for follow-up work.

## Module

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [x] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
cmake -B build-tent -DCMAKE_BUILD_TYPE=Release -DUSE_TENT=ON -DTENT_METRICS_ENABLED=OFF -DWITH_STORE=OFF
cmake --build build-tent --target tent_progress_worker_test tebench
./build-tent/mooncake-transfer-engine/tent/tests/tent_progress_worker_test
```

**Test results:**
- [x] Unit tests pass — 22/22 in `tent_progress_worker_test`, including 3 new concurrency tests: `ConcurrentIndependentBatchPolls` (8 threads poll+free independent batches), `PollRacesWithFreeSameBatch` (poll vs free of one batch returns OK or InvalidArgument, never crashes), `RuntimeQueuePollAndFreeDoesNotDeadlock`
- [x] Manual testing done (describe below) — full `tebench` matrix on the same machine pair: before/after/classic × local/cross-machine × 1–16 threads × 4KiB ops, plus 256KiB ops; methodology and numbers in Description

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

CodeBuddy (GLM) assisted with root-cause profiling (rdtsc section counters), implementing the sharded registry, resolving rebase conflicts against the reclaim rework on main, and running the benchmark matrix. Every changed line has been reviewed and is understood by the human submitter.

### DIFF
--- a/mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
+++ b/mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
@@ -553,11 +553,17 @@ Status TransferEngineImpl::deconstruct() {
             }
             Slab<Batch>::Get().deallocate(batch);
         };
-        for (auto& batch : batch_set_.active) release_batch(batch);
-        for (auto& batch : batch_set_.freelist) release_batch(batch);
-        batch_set_.active.clear();
-        batch_set_.freelist.clear();
-        alive_batches_.clear();
+        // Registry state is sharded; collect members under each shard.
+        // Shutdown is quiesced (progress_mutex_ held, no pollers expected),
+        // so a brief per-shard lock per batch is sufficient.
+        for (auto& shard : batch_shards_) {
+            std::lock_guard<std::recursive_mutex> shard_lk(shard.mtx);
+            for (auto* batch : shard.active_batches) release_batch(batch);
+            shard.active_batches.clear();
+            shard.alive_batches.clear();
+        }
+        for (auto* batch : batch_freelist_) release_batch(batch);
+        batch_freelist_.clear();
     }
 
     // Now safe to destroy transports (workers join here)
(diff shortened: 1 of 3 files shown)


mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
```diff
@@ -1000,12 +1080,20 @@ Status TransferEngineImpl::retainBatch(BatchID batch_id, Batch*& batch) {
...
+                "Batch runtime ref underflow" LOC_MARK);
+        }
+        --batch->runtime_refs;
+        deferred_free = (batch->runtime_refs == 0 && batch->free_requested);
     }
-    --batch->runtime_refs;
-    if (batch->runtime_refs == 0 && batch->free_requested) {
+    if (deferred_free) {
         CHECK_STATUS(lazyFreeBatch());
     }
```
alogfans, line 1098: batch_freelist_ is documented as being guarded by progress_mutex_, but this path calls lazyFreeBatch() after releasing the shard lock without acquiring progress_mutex_. Concurrent free/release/progress-worker activity can therefore iterate or erase the freelist at the same time. Please acquire progress_mutex_ before invoking the sweep here, while preserving the documented progress_mutex_ -> shard lock order.

staryxchen: > batch_freelist_ is documented as being guarded by progress_mutex_, but this path calls lazyFreeBatch() after releasing the shard lock without acquiring progress_mutex_. Concurrent free/release/progress-worker activity can therefore iterate or erase the freelist at the same time. Please acquire progress_mutex_ before invoking the sweep here, while preserving the documented progress_mutex_ -> shard lock order.

Thanks for the review. `lazyFreeBatch()` already holds `progress_mutex_` on entry, so the freelist is not accessed unlocked. The shard is dropped first so the sweep can take `progress_mutex_ -> shard` without inverting the order. An extra lock here would be redundant.

alogfans (approved):
LGTM. Thanks!

PR #2080, opened 2026-05-11
[Store] Add DSA-like workload allocation strategy and update allocator benchmarks

yokinoshitayoki:
## Description

This PR extends the DSA-like benchmark coverage for Mooncake Store by updating both the OffsetAllocator benchmark and the AllocationStrategy benchmark, and adds corresponding performance result documentation.

The benchmark workload is based on paired KV cache and indexer allocation using production-like object sizes:

* KV cache object size: 3,274,752 B(3.12 MB)
* Indexer object size: 658,432 B(643 KB)

The workload models DSA-style allocation behavior where each round samples `N` uniformly from `[1, 128]`, then issues `N` KV allocations followed by `N` indexer allocations. On allocation failure, the benchmark randomly evicts 5% of all live objects and retries. Utilization is sampled at eviction-trigger time, immediately before eviction, so the reported `UtilRatio_*` values represent the peak fill level reached before the allocator is forced to evict.

Changes included in this PR:

* Update `allocator_bench.cpp` DSA workload:
  * Use a 1 TB pool.
  * Use 5% random eviction on allocation failure.
  * Sample utilization only at eviction-trigger time.
  * Report the number of eviction-trigger samples.
* Update `allocation_strategy_bench.cpp` DSA workload:
  * Use 1 TB per segment.
  * Test segment counts `{1, 2, 4, 8, 16}`.
  * Compare KV-only (non-DSA baseline) and paired KV+Indexer (DSA workload).
  * Compare `RandomAllocationStrategy` and `FreeRatioFirstAllocationStrategy`.
  * Use the same 5% mixed live-object eviction model as `allocator_bench.cpp`.
* Update benchmark result documentation:
  * Refresh OffsetAllocator DSA result in `allocator-benchmark-result.md`.
  * Add full AllocationStrategy DSA result matrix in `allocation-strategy-benchmark-result.md`.

## Module

- [ ] Transfer Engine (`mooncake-transfer-engine`)
- [x] Mooncake Store (`mooncake-store`)
- [ ] Mooncake EP (`mooncake-ep`)
- [ ] Integration (`mooncake-integration`)
- [ ] P2P Store (`mooncake-p2p-store`)
- [ ] Python Wheel (`mooncake-wheel`)
- [ ] PyTorch Backend (`mooncake-pg`)
- [ ] Mooncake RL (`mooncake-rl`)
- [ ] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
[...]

PR #3651, opened 2026-08-25
[Store] Add size-class-aware segment placement

982945902:
## Description

Add an opt-in `size_class_aware` Segment allocation strategy for Mooncake Store.

For `ReplicaType::MEMORY` and `ReplicaType::NOF_SSD` backed by `OffsetBufferAllocator`, the strategy groups live allocations into logarithmic requested-size classes and ranks each candidate with:

```text
score = size_class_free_ratio_weight * free_ratio
      + size_class_matching_share_weight * matching_share
```

The weights are configurable through `--size_class_free_ratio_weight` and `--size_class_matching_share_weight`. Their defaults are `1.0` and `0.15`; invalid, non-finite, or negative values are rejected. Setting the matching-share weight to zero reproduces free-ratio-first candidate ranking under the same sampling seed.

The existing preferred/excluded Segment handling, multi-replica placement, and allocation-failure fallback semantics are preserved. Allocators without size-profile support, including CacheLib, fall back to weighted free-ratio ordering. Size-class profiles are rebuilt naturally while restoring `OffsetBufferAllocator`; this change introduces no wire-format or snapshot-format changes.

This is complementary to #2797 rather than a duplicate: this PR chooses a preferred Segment cohort from the live requested-size distribution, while #2797 ranks Segments using largest-free-region/fragmentation information. This PR does not change `OffsetAllocator` internals.

## Module

- [ ] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [x] Documentation update
- [x] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
# Release targets
cmake --build /build/release --target \
  allocation_strategy_test buffer_allocator_test \
  master_service_config_test allocation_strategy_bench mooncake_master

./allocation_strategy_test
./buffer_allocator_test
./master_service_config_test

# PR-scoped checks
pre-commit run --files <changed files>
```

**Test results:**
- [x] Unit tests pass
- [x] Manual testing done (allocation-strategy benchmark)

- `allocation_strategy_test`: 99/99 passed
- `buffer_allocator_test`: 20/20 passed
- `master_service_config_test`: 8/8 passed
- `master_service_test_for_snapshot`: 70/70 passed, including size-class profile restore and post-restore deallocation
- Release builds passed for the allocator tests, benchmark, and `mooncake_master`
- Pre-commit passed on all changed files

### Weight selection benchmark

The benchmark now accepts an explicit random seed and resets the candidate-sampling engine before each strategy run. With matching-share weight `0`, `size_class_aware` exactly matches `free_ratio_first` fragmentation in every tested configuration, removing sampling noise from the comparison.

A sweep of matching-share weights from `0` through `1.0` used a fixed free-ratio weight of `1.0`, seed `42`, 1 GiB per Segment, 80% prefill, and 50,000 measured allocations. The default `0.15` was selected as a moderate affinity: larger values increasingly regressed DSA-style workloads, while `0.15` reached the target mixed-size/skewed case without materially changing the highlighted DSA guardrails.

Three-run Release benchmark; throughput is the median, fragmentation is deterministic for the fixed seed:

| Pattern | Segments | Skewed | Fragmentation improvement | Throughput change |
|---|---:|:---:|---:|---:|
| `kv_mixed` | 10 | no | +0.96% | -6.00% |
| `kv_mixed` | 100 | no | +8.01% | -7.70% |
| `kv_mixed` | 100 | yes | **+30.76%** | **+7.34%** |
| `dsa_pair` | 10 | no | +1.50% | -14.85% |
| `dsa_pair` | 100 | no | +1.50% | +3.71% |
| `dsa_pair` | 100 | yes | -0.12% | +10.07% |

Across the complete sweep matrix at weight `0.15`, average fragmentation improvement was `+1.36%`, the worst configuration was `-4.92%`, and four configurations regressed by more than 2%. The benefit is workload- and topology-dependent, so the strategy remains opt-in and both weights are exposed for deployment-specific tuning. The benchmark exercises the shared `OffsetAllocator` placement path with memory replicas; NoF SSD support is covered by allocation, deallocation, restore, and strategy-selection unit tests.

## Checklist

- [x] I have formatted my code using the repository formatting hooks
- [x] For changes >500 LOC: not applicable (production change is below 500 LOC)

## AI Assistance Disclosure

- [ ] No AI tools were used

OpenAI Codex assisted with design analysis, implementation, tests, documentation, conflict resolution, and validation. The human submitter reviewed the resulting change and is responsible for understanding and defending it end-to-end.


Aionw (commented):
Is there any benchmark result about this PR?

982945902:
Thanks for the reminder. I have updated the PR description with three-run Release benchmark results, including the exact workload parameters, throughput, and fragmentation numbers for both `kv_mixed` and `dsa_pair`. I also adjusted the strategy to use a bounded size-class affinity on top of free-ratio ranking after the initial hard-cohort design showed regressions in some configurations. Could you please take a look and let me know whether the benchmark coverage and results meet your expectations?

Aionw (commented):
HA restoration currently misses size-class accounting.

live_bytes_by_size_class_ is updated only in allocate/deallocate, while snapshot deserialization restores offset_allocator_ and cur_size_ directly. Consequently, restored allocations start with zero class counters, and freeing one performs an unsigned fetch_sub from zero, which can permanently skew matching_share.

For the fix, I do not think a size-class-specific restore hook belongs in BufferAllocatorBase: profiling is not a universal allocator capability, and a default no-op would hide future omissions. Serializer<AllocatedBuffer>::deserialize already resolves OffsetBufferAllocator to restore OffsetAllocationHandle. Please retain that concrete pointer and, after successfully constructing the restored AllocatedBuffer, call a private OffsetBufferAllocator::RestoreAllocationProfile(size). This introduces no additional RTTI in the hot path and keeps Offset-specific restoration at the existing type boundary.

Please also add an HA snapshot round-trip test that checks the profile before and after restore, then releases the restored allocation and verifies that the counters return to zero.

982945902:
@Aionw Addressed in `ef966bfd`.

- Kept the restore hook private to `OffsetBufferAllocator` and granted access only to `Serializer<AllocatedBuffer>`.
- Retained the concrete `OffsetBufferAllocator` pointer while restoring the offset handle, then restored the requested-size profile only after the `AllocatedBuffer` was successfully constructed.
- Did not call `RecordAllocation`, because `cur_size_` is already restored by `Serializer<OffsetBufferAllocator>`.
- Added an HA snapshot round-trip regression test covering two size classes, profile equality after restore, and counters returning to zero after deleting the restored allocations.

Validation: `allocation_strategy_test` 99/99, `buffer_allocator_test` 20/20, and `master_service_test_for_snapshot` 70/70 passed in Release mode; pre-commit passed on all touched files.

ykwd:
Thanks for the PR! I appreciate the motivation to improve mixed-attention workloads, since they do indeed involve objects of different sizes. One question: how is “fragmentation improvement” defined in this benchmark?

ykwd:
Related work: https://github.com/kvcache-ai/Mooncake/pull/2080

982945902:
Thanks for pointing this out.

The benchmark defines fragmentation per Segment as:

```text
1 - largest_free_region / total_free_space
```

The cluster-level value is weighted by each Segment's free bytes, equivalently:

```text
sum(total_free_space_i - largest_free_region_i)
------------------------------------------------
              sum(total_free_space_i)
```

`Frag_avg` is the arithmetic mean of this cluster-level value sampled every `--convergence_sample_interval` allocation attempts. The improvement reported in the PR is calculated relative to `free_ratio_first`:

```text
(free_ratio_first Frag_avg - size_class_aware Frag_avg)
-------------------------------------------------------
              free_ratio_first Frag_avg
```

For example, the reported 30.76% case is `(0.0972 - 0.0673) / 0.0972`.

Regarding #2080: the current `dsa_pair` pattern reuses the production-like KV and indexer object sizes introduced there (`3,274,752 B` and `658,432 B`), but it is not an exact reproduction of #2080's DSA workload. #2080 issues paired bursts of N KV allocations followed by N indexer allocations and samples utilization at eviction triggers. This PR's size-class churn workload randomly interleaves the two sizes at a 50/50 ratio, prefills the cluster, and periodically samples fragmentation. It is intended as a size-distribution guardrail for the placement strategy rather than a replacement for the existing DSA benchmark.

I will clarify this distinction in the benchmark naming/documentation so it does not imply full equivalence with #2080. We can also add `size_class_aware` to the existing `--workload=dsa` matrix if a direct comparison under the exact #2080 workload is preferred.

ykwd:
Thanks for the detailed explanation. In practical LLM workloads, there are only a small number of object sizes that need to be stored, so fragmentation per segment is not a major concern. The key factors are the achievable memory utilization and allocation speed. (Reference: #2080 )

982945902:
Thanks for the clarification. I reran the complete benchmark matrix using the #2080 DSA methodology.

First, a correction to my earlier benchmark description: these measurements exercise the virtual-address `OffsetAllocator` and placement path. The latency/throughput values are allocator-control-path microbenchmark results, not physical DRAM writes or end-to-end Mooncake Store throughput.

The complete matrix contains 144 results:

- `Random`, `FreeRatioFirst`, and `SizeClassAware`
- KV-only and paired KV+Indexer workloads
- homogeneous and skewed Segment capacities
- 1/2/3 replicas
- 1/2/4/8/16 Segments
- 1 TiB base Segment capacity
- 1x-capacity warmup followed by 3x-capacity measurement
- 5% fail-triggered random eviction

### Currently pushed weighted strategy

Relative to `FreeRatioFirst`, averaged across the 48 matching configurations:

- achievable utilization: **-0.0720 percentage points**
- allocation microbenchmark throughput: **-1.47%**
- allocation P99: **+5.43%**
- eviction rounds: **-0.25 rounds**

For the paired KV+Indexer workload specifically:

- homogeneous capacities: utilization **-0.0724 percentage points**, throughput **-7.29%**, P99 **+16.95%**
- skewed capacities: utilization **-0.2155 percentage points**, throughput **+4.15%**, P99 **-4.42%**

The worst utilization regression was **-0.8578 percentage points**.

### Unpushed refinement

I also evaluated a local refinement that removes the configurable weighted score, uses the actual `OffsetAllocator` bins, and restricts affinity to near-tied free-ratio cohorts. Across the same complete matrix, relative to `FreeRatioFirst`:

- achievable utilization: **+0.0086 percentage points**
- allocation microbenchmark throughput: **-4.84%**
- allocation P99: **+4.53%**
- eviction rounds: **-0.10 rounds**

Its best case improved utilization by **0.1105 percentage points**, but the average benefit remained negligible and did not justify the additional placement complexity and overhead.

Therefore, the complete #2080-aligned results do not support the performance-improvement claim of this PR. The earlier fragmentation table is useful only as an allocator diagnostic and does not translate into improved achievable utilization for the production-like DSA workload. I will not push the additional refinement or continue using the previous fragmentation/throughput numbers as evidence of practical performance improvement.

982945902:
Follow-up: I split the allocator size-class precision experiment into a separate PR: #3787. It increases the SmallFloat mantissa from 3 to 4 bits and includes backward-compatible snapshot migration. In two alternating, CPU-pinned runs of the complete 96-case DSA matrix, achievable utilization improved by +3.8409 pp for KV-only and +3.9326 pp for paired KV+Indexer, while aggregate throughput was effectively neutral (+0.24% and +0.10%). This is independent of the segment-selection policy in this PR.

yokinoshitayoki (commented):
Thank you for your contribution and the detailed benchmark, but based on the latest benchmark results, maybe this change is not suitable for merging into the main branch.

982945902:
Agreed. The complete #2080-aligned benchmark shows that size-class-aware Segment placement adds complexity without a meaningful utilization benefit, so I am closing this PR.

I have moved the investigation to a simpler, independent approach in #3787: increasing the OffsetAllocator SmallFloat mantissa from 3 to 4 bits. That directly reduces allocation-size rounding rather than changing Segment selection. In two alternating, CPU-pinned runs of the complete 96-case DSA matrix, it improved achievable utilization by +3.8409 percentage points for KV-only and +3.9326 percentage points for paired KV+Indexer, with effectively neutral aggregate allocation throughput (+0.24% and +0.10%). It also includes raw and MsgPack backward-compatible migration tests, plus an old-3-bit-writer to new-4-bit-reader cross-version validation.

Thanks for the review and for pointing me toward the practical utilization and allocation-speed metrics.

PR #3652, opened 2026-08-25
[Bugfix][Store] BucketIdGenerator: add per-client entropy to fresh-start seed to prevent bucket file collision (#3528)

### LINKED ISSUE #3528: [Bug&RFC] Bucket LOCAL_DISK: clients sharing a `storage_path` collide on bucket ids → L3 cache broken
### Bug Report

**Summary.** When multiple clients share one `storage_path` (e.g. SGLang embeds one Mooncake client per TP rank on the same `ssd_offload_path`), the bucket `LOCAL_DISK` backend has no per-client namespacing and no stable identity. This causes three coupled problems: silent collision, ambiguous recovery ownership, and per-client accounting that can neither be attributed nor survive restart. Existing issues cover pieces; none covers the **"N clients, one `storage_path`"** case.

**Environment.** Mooncake bucket `LOCAL_DISK` backend (`main`). Occurs whenever multiple clients share one `storage_path` — e.g. SGLang embeds one Mooncake client per TP rank on the same `ssd_offload_path` (sgl-project/sglang#31926).

**Three coupled problems** (`mooncake-store/src/storage_backend.cpp` unless noted):
1. **Silent collision.** `BucketIdGenerator` seeds `(time_gen() << 12) | 0` (whole-second) then only `fetch_add`s; filename `<storage_path>/<bucket_id>.bucket` has no client tag; opened `O_CREAT|O_TRUNC`, no `O_EXCL`. Same-second construction → identical ids → silent overwrite. The backend struct holds only `storage_path_` — no `client_id`/pid to disambiguate.
2. **Ambiguous recovery.** `Init()` scans the whole dir and adopts every `.meta`, keyed by numeric `bucket_id`; `MasterService::AddReplica` (`master_service.cpp`) keeps one LOCAL_DISK replica per key with no `client_id` check (see #3052). Shared dir → every client claims every file → arbitrary first-claim → dangling refs when the "owner" evicts.
3. **No stable, attributable identity.** `client_id` is a random UUID minted per `Client` ctor (`client_service.cpp`, boost random), not stable across restart, and never assigned by the master. So a client cannot be recognized as the same logical owner after a restart, and its on-disk data cannot be attributed to it — which is what makes per-client, layout-independent accounting impossible today.

**Related.** #3052 (client-restart / `client_id` validation), #2306 + #2777 (marker re-adoption — marker is per-directory), #3030 (replica-id validation), #3220 + #3221 (per-object delete + GC), #2721 (SSD-capacity report), SGLang sgl-project/sglang#31926.

**The gap.** All of the above assume **one client per `storage_path`**. "N clients, one directory" is unsupported, and even the identity primitives that exist (RFC #2306's marker) are per-directory.

**Proposal** (we prefer B; A is a stopgap):
- **(A) Stopgap** — enforce one-dir-per-client (via SGLang #31926), document the contract, and fail fast on a shared directory. Leaves accounting bound to the directory layout.
- **(B) Preferred — directory-independent, Mooncake-side.** Embed a durable `owner_id` in each bucket's `.meta` (not the filename or directory). Per-client accounting and `Init()` recovery then filter by `owner_id`, so both work for **any** layout — shared dir, per-client dirs, or multiple disks. Master re-adopts by matching the in-metadata `owner_id` (not #2306's per-directory marker). Per-client bytes are reported to the master (cf. #2721); a node/disk cap is enforced by master-side aggregation, not a shared-dir scan.
  - Open design point: the durable anchor by which a restarted process recovers its own `owner_id` — a master-side registry keyed by a stable client handle, or a minimal persisted id decoupled from the offload directory.

**Ask.** Decide the ownership model for a shared `storage_path`. We advocate (B) so per-client identity and accounting become layout-independent; (A)/#31926 is a stopgap. This should be coordinated with the warm re-adoption work (#2306). Willing to contribute the PR once the direction is agreed.

### Before submitting...

XFDG:
## Root cause

When multiple clients share one `storage_path` (e.g. SGLang embeds one Mooncake client per TP rank on the same `ssd_offload_path`), `BucketIdGenerator`'s fresh-start seed was

    (time_gen() << TIMESTAMP_SHIFT) | SEQUENCE_ID_SHIFT

which carries **no client entropy**: the low 12 sequence bits were always 0 and the timestamp has whole-second granularity. Every client that constructed its generator within the same second therefore seeded the **identical** value and produced the **identical** id sequence. Their bucket files (`<storage_path>/<bucket_id>.bucket`) then collided under `O_CREAT|O_TRUNC` (no `O_EXCL`): a later client silently truncated an earlier client's bucket, leaving keys registered (visible via `get_all_keys` / `IsExist`) but reads returning empty bytes.

Reporter (Morpheus799) confirmed the precondition is deterministic on DeepSeek-V3/TP8: all 8 ranks in a pod seed the identical bucket id (observed `7320020267008` x8) because `time_gen()` is whole-second and the TP ranks initialize within the same second. The failure surfaces as `Read size mismatch ... expected: N, got: 0` before any eviction runs — i.e. the collision itself, not an eviction/read race. Requests still succeed because HiCache treats the failed read as a miss and recomputes, so the corruption is silent at the API level but the L3 cache is unusable under this configuration.

## The fix

Add `GenerateBucketIdEntropySeed()`, which folds a per-process unique nonce (`random_device` + `getpid()` + ASLR stack address) plus a monotonic counter into the low 12 sequence bits. Two clients starting in the same second now get different sequences and their bucket files never overlap. The high timestamp bits are preserved, so monotonicity and the "last used bucket id" recovery order are unchanged. Collision probability is ~1/2^12 and non-deterministic, unlike the previous always-zero low bits.

This is the minimal, self-contained fix at the source. The issue's preferred option (B) — a durable, layout-independent `owner_id` baked into each bucket's `.meta` — is explicitly an open design point needing master/SGLang coordination; the entropy fix removes the deterministic corruption regardless of which ownership model is later chosen, and needs no protocol changes.

## How it has been tested

- Core logic verified standalone via `g++-13` (`entropy_test.cc`): high timestamp bits preserved (monotonicity), low 12 bits non-zero, two generators diverge, explicit-start path unaffected, monotonic within a generator. ALL PASS.
- Existing test `InitializeWithInvalidStart_UseTimestampFallback` still passes (entropy only adds to low bits, so `gen.CurrentId() >= (time << 12) | 0`).
- Added `FreshStartSeedsDivergeAcrossClients` regression test.

Fixes kvcache-ai/Mooncake#3528

### DIFF
--- a/mooncake-store/src/storage_backend.cpp
+++ b/mooncake-store/src/storage_backend.cpp
@@ -20,6 +20,7 @@
 #include <algorithm>
 #include <chrono>
 #include <unordered_set>
+#include <random>
 
 #include <ylt/struct_pb.hpp>
 
@@ -1689,7 +1690,8 @@ void StorageBackendAdaptor::RemoveAll() {
 BucketIdGenerator::BucketIdGenerator(int64_t start) {
     if (start <= 0) {
         auto cur_time_stamp = time_gen();
-        current_id_ = (cur_time_stamp << TIMESTAMP_SHIFT) | SEQUENCE_ID_SHIFT;
+        current_id_ = (cur_time_stamp << TIMESTAMP_SHIFT) |
+                      GenerateBucketIdEntropySeed();
     } else {
         current_id_ = start;
     }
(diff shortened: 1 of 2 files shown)


XFDG:
## Duplicate of #2614 — please close in favor of that one

Heads up @zhaoye — this PR (#3652) and #2614 (@leonzzhu) both fix the same #3528 bucket_id collision in the same files (`storage_backend.cpp` + `storage_backend_test.cpp`), with the same root cause (whole-second `time_gen()` → same-second clients → `O_TRUNC` collision).

#2614 is the more complete approach:
- **19-bit process salt in high bits** (mine only perturbs the low 12 bits).
- **Explicit restart/ScanMeta path** that bumps time until `id > max_on_disk` — #3652 does not handle restart re-collision.
- **Salt logging** for per-process attribution.
- Open since 2026-06-25, under active review.

Please close #3652 and continue review feedback on #2614. CC @Morpheus7993.

PR #3654, opened 2026-08-25
[Build] Unify toolchain

ur4t:
## Description

Use `pip` to manage toolchains including `cmake`, `ninja` and even `go`, and update `cmake` usage patterns.

## Module

- [x] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [x] Documentation update
- [ ] Performance improvement
- [x] Other

## How Has This Been Tested?

**Test commands:**
```bash
```

**Test results:**
- [ ] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [x] No AI tools were used


dependencies.sh
```diff
@@ -258,90 +254,28 @@ if [ "$OS" = "ubuntu" ] || [ "$OS" = "debian" ]; then
...
-    if ! grep -q "export GOPROXY=" ~/.bashrc; then
-        echo 'export GOPROXY=https://goproxy.cn,https://goproxy.io,direct' >> ~/.bashrc
-        echo -e "${YELLOW}GOPROXY added to ~/.bashrc for future sessions${NC}"
-    fi
-elif [ -n "$GOPROXY" ]; then
-    echo -e "${GREEN}GOPROXY already set to: ${GOPROXY}${NC}"
+# Add toolchain to PATH if not already there
+if ! grep -q "export PATH=\$PATH:${MOONCAKE_TOOLCHAIN}/bin" ~/.bashrc; then
+    echo -e "${YELLOW}Adding toolchain to your PATH in ~/.bashrc${NC}"
+    echo 'export PATH=$PATH:${MOONCAKE_TOOLCHAIN}/bin' >> ~/.bashrc
```
Aionw: [P1] This single-quoted string writes `${MOONCAKE_TOOLCHAIN}` literally into `.bashrc`, but that variable only exists in the installer subprocess. It is normally undefined when a new shell reads the file, so this effectively adds only `/bin` to PATH. In addition, running `bash dependencies.sh` cannot modify the caller's PATH, so the nightly, Ascend, and EFA callers either keep using the image-provided tools or fail to find CMake/Ninja on a clean image. Please persist the literal `/opt/mooncake-toolchain/bin` path and explicitly update the current PATH or `GITHUB_PATH` at each workflow caller.

dependencies.sh
```diff
@@ -258,90 +254,28 @@ if [ "$OS" = "ubuntu" ] || [ "$OS" = "debian" ]; then
     print_success "ldd found: $(ldd --version 2>&1 | head -1)"
 fi
 
-print_section "Installing Go $GOVER"
+print_section "Installing Toolchain: CMake, Ninja and Go"
 
-USED_CN_MIRROR=false
+python3 -m venv ${MOONCAKE_TOOLCHAIN}
+${MOONCAKE_TOOLCHAIN}/bin/python3 -m pip install --upgrade pip cmake ninja
+${MOONCAKE_TOOLCHAIN}/bin/python3 -m pip install go-bin==1.27.0
```
Aionw: [P2] `dependencies.sh` does not enable `set -e`, and none of these new venv/pip commands call `check_success`. If venv creation, PyPI access, or wheel installation fails, the script still prints `Toolchain installed successfully` and may exit with status 0, allowing the build to silently use an older host toolchain. Please check each command's return status or make this section fail immediately.

scripts/ascend/dependencies_ascend.sh
```diff
@@ -1,154 +0,0 @@
...
-# Unless required by applicable law or agreed to in writing, software
-# distributed under the License is distributed on an "AS IS" BASIS,
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
-
-# If git clone fails, you can place dependencies and Mooncake source code in a directory for compilation and installation.
-
-# ASCEND TRANSPORT is scheduled for deprecation, please use ASCEND DIRECT TRANSPORT on ASCEND platform.
-# Use dependencies_ascend_installation.sh to install dependencies instead.
```
Aionw, line 19: [P2] After this script is deleted, `docs/source/design/transfer-engine/ascend_transport.md` lines 33-36 and `docs/source/zh_archive/ascend_transport.md` lines 24-25 still instruct users to run it. Although this commit updates other build documentation, these two sections still contain a command that no longer exists. Please replace or remove the corresponding one-step build instructions.

Aionw (changes requested):
Three issues still need to be addressed before merging: the toolchain directory is not reliably added to subsequent shell/workflow PATHs; toolchain installation failures are reported as success; and the deleted Ascend one-step script is still referenced by both the English and Chinese documentation. The devcontainer PATH typo was fixed in the latest commit, so I did not submit a duplicate comment for it.

docs/source/getting_started/build.md
```diff
@@ -22,11 +22,9 @@ sudo bash dependencies.sh
...
 
 ```bash
-mkdir build
-cd build
-cmake ..
-make -j
-sudo make install
+cmake -B build -G Ninja
+cmake --build build
+sudo cmake --install install
```
Aionw: [P1] The project is configured and built in `build`, but this command asks CMake to install from a different `install` directory, so the primary build guide fails at the installation step. Please use `cmake --install build`. The same typo appears in the other three examples in this file, the Kunpeng UMDK instructions, and several archived Chinese build/store examples, so all of those occurrences need the same correction.

docs/source/design/transfer-engine/efa_transport.md
```diff
@@ -63,29 +63,23 @@ To build from source instead (for development, an unreleased revision, or a cust
...
-mkdir build && cd build
-
-cmake .. \
+cmake -B build -G Ninja \
     -DUSE_EFA=ON \
     -DUSE_CUDA=ON \
     -DCMAKE_BUILD_TYPE=RelWithDebInfo
-
-make -j$(nproc)
+cmake --build build
```
Aionw, line 70: [P2] The old instructions changed into `build/` before reaching this section, but the new `cmake -B build` flow leaves the shell at the repository root. Consequently, the built modules are under `build/mooncake-integration` and `build/mooncake-common`, while `../mooncake-wheel` now points outside the repository. Please update both the source and destination paths for the new working directory. The UB guide has the same stale relative paths, and the MPComm test snippets similarly still assume that the shell is inside `build/`.

docs/source/design/transfer-engine/sunrise_link_transport.md
```diff
@@ -41,17 +41,16 @@ git clone https://github.com/kvcache-ai/Mooncake.git
 cd Mooncake
 
 # Enable TENT + Sunrise Link
-mkdir build && cd build
-cmake .. -DUSE_TENT=ON -DUSE_SUNRISE=ON
+cmake -B build -G Ninja -DUSE_TENT=ON -DUSE_SUNRISE=ON
 
 # Build
-make -j$(nproc)
+cmake --build
```
Aionw: [P2] `cmake --build` requires a build directory (or a preset); without one this command only prints the usage text and exits with an error. Since the preceding command configures `build`, this should be `cmake --build build`.

Aionw (commented):
Additional issues found while reviewing the documentation conversion to out-of-source CMake commands.

dependencies.sh
ur4t: Currently try keeping `~/.bashrc` untouched to clarify which workflows require `PATH` update.

dependencies.sh
ur4t: Now a subshell will run a script to install toolchains. Toolchain installation process should be a standalone step in workflows in the future.

docs/source/getting_started/build.md
ur4t: All doc typos are supposed to be fixed in the latest revision. One reply to avoid spamming.

dependencies.sh
Aionw: [P2] Clarification after re-checking: `dependencies.sh` itself should not source the venv, because it runs as a child process and cannot update the caller environment. Existing developer and CI images may still compile by finding host-provided CMake/Ninja. The remaining issue is that the clean-host instructions and several workflows neither activate `/opt/mooncake-toolchain` nor invoke its tools through a scoped launcher, so they either fail when host tools are absent or silently bypass the unified toolchain. Please document `source /opt/mooncake-toolchain/bin/activate` or provide a small launcher and use it at those call sites; no `.bashrc` modification is needed.

scripts/install_toolchain.sh
```diff
@@ -0,0 +1,29 @@
...
+
+# ensure python3 with venv module
+command -v apt-get && apt-get update && apt-get install -y python3-venv || :
+command -v yum && yum makecache && yum install -y python3 || :
+command -v python3 || exit 1
+
+python3 -m venv ${MOONCAKE_TOOLCHAIN}
+test -z "$GITHUB_MIRROR" || python3 -m pip config --user set \
+  global.index-url https://mirrors.huaweicloud.com/repository/pypi/simple
+python3 -m pip install --upgrade pip cmake ninja
```
Aionw, line 24: [P2] The shared installer announces a CMake/Ninja/Go toolchain, but this command installs only CMake and Ninja. The default configuration does not require Go, so host environments can hide the omission; however, `USE_ETCD=ON`, `STORE_USE_ETCD=ON`, Go bindings, and P2P Store builds execute `go`, and then use an arbitrary host Go or fail on a clean image. The devcontainer already installs `go-bin==1.27.0`. Please install the same pinned package into this venv, or make every Go-enabled caller provision an explicit version and stop reporting Go as installed here.

scripts/ascend/dependencies_ascend.sh
```diff
@@ -129,12 +127,9 @@ if ! git submodule update --init --recursive; then
     fi
 fi
 rm -rf build
-mkdir -p build
-cd build
-cmake -DCMAKE_POLICY_VERSION_MINIMUM=3.5 ..
-make -j
-make install -j
-cd ..
+cmake -B build -G Ninja -DCMAKE_POLICY_VERSION_MINIMUM=3.5
```
Aionw, line 140: [P2] This remains documented as a one-step build script, but it no longer installs CMake and never installs or activates the new toolchain before invoking CMake with the Ninja generator. A preconfigured Ascend CI image may hide this, while the documented clean-host flow fails as soon as CMake or Ninja is absent. Please bootstrap the shared toolchain and use it through scoped activation/a launcher here, or keep installing both CMake and Ninja in this standalone path.

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.16)
+cmake_minimum_required(VERSION 4.0)
```
Aionw, line 1: [P1] Raising the project minimum to CMake 4.0 conflicts with the primary build guide, which still recommends CMake 3.20.x. A caller following the documented prerequisite is rejected before configuration, independently of the PATH issue. Please update the supported-version documentation and installation flow, or retain a compatible project minimum.

docs/source/getting_started/build.md
```diff
@@ -22,11 +22,9 @@ sudo bash dependencies.sh
...
 
 ```bash
-mkdir build
-cd build
-cmake ..
-make -j
-sudo make install
+cmake -B build -G Ninja
+cmake --build build
+sudo cmake --install build
```
Aionw, line 27: [P1] Invoking `cmake` through `sudo` re-resolves the command using sudo's PATH; on the usual `secure_path` configuration, `/opt/mooncake-toolchain/bin` is absent. Even after the caller activates the new venv, this install either fails on a clean host or silently uses an older system CMake. Please invoke the already-selected executable explicitly (for example, `sudo "$(command -v cmake)" --install build`) or use a user-writable install prefix. The same pattern appears in several workflows.

scripts/ascend/dependencies_ascend_installation.sh
```diff
@@ -108,22 +107,20 @@ elif command -v yum &> /dev/null; then
     clone_repo_if_not_exists "yaml-cpp" https://github.com/jbeder/yaml-cpp.git
     cd yaml-cpp || exit
     rm -rf build
-    mkdir -p build && cd build
-    cmake ..
-    make -j$(nproc)
-    make install
-    cd ../..
+    cmake -B build -G Ninja
```
Aionw: [P2] This script is documented as a standalone deployment artifact, but on the yum path it no longer installs CMake and never installs or activates the shared toolchain, while these new commands require both CMake and Ninja. Running only the copied script on a clean RPM host therefore fails here. Please keep provisioning both tools in this standalone path, or bundle and activate the shared installer.

scripts/install_toolchain.sh
```diff
@@ -0,0 +1,30 @@
...
+python3 -m pip install go-bin==1.27.0
+
+test -z "$AUDIT_GO_WHEEL" && exit
+
+cd $(find ${MOONCAKE_TOOLCHAIN} -type d -name go)
```
Aionw: [P2] `go-bin==1.27.0` contains several directories named `go`, so this command substitution passes multiple operands to `cd` and the audit stops with `cd: too many arguments`. Even after selecting the package root unambiguously, `diff ... | tee` returns `tee`'s status and accepts a mismatch, and the reference tarball is hard-coded to amd64. Please resolve the installed Go root explicitly, select the reference architecture from the host, and preserve `diff`'s failure status.

Aionw (changes requested):
Four additional issues remain on the current head: the documented CMake version is incompatible with the new project minimum, privileged install commands can bypass the venv toolchain, the standalone Ascend installer no longer provisions the tools it invokes on RPM hosts, and the optional Go-wheel audit cannot perform or enforce its comparison. Existing PATH and stale relative-path threads are not duplicated here.

scripts/install_toolchain.sh
ur4t: Now use official Go tarballs instead.

scripts/ascend/dependencies_ascend_installation.sh
ur4t: Temporarily use `cmake` from `yum` to build those packages.

docs/source/getting_started/build.md
ur4t: A slightly hacky soft link is added to `/usr/local/bin/cmake`.

CMakeLists.txt
ur4t: Docs updated.

PR #3660, opened 2026-08-25
[Bugfix][TE] Restore fork state on memory unregister to prevent VMA exhaustion (#3639)

### LINKED ISSUE #3639: [TE] Unconditional ibv_fork_init() makes every register/unregister cycle permanently split VMAs (MADV_DONTFORK never reverted) → vm.max_map_count      exhaustion
## Summary

`RdmaContext` unconditionally calls `ibv_fork_init()` at engine startup (and the new TENT transport does the same). Once verbs fork protection is active, **every `ibv_reg_mr` performed through the transfer engine issues `madvise(range, MADV_DONTFORK)` on the registered range, and the corresponding `unregister` never issues `MADV_DOFORK`**. The kernel therefore tags those pages `VM_DONTCOPY`, which splits the containing VMA and makes it permanently unmergeable with its neighbors.

For workloads that register/unregister buffers at high frequency (per-request buffer registration rather than a registered-once pool), this is a monotonic leak of VMA boundaries: **`vm.max_map_count` is eventually exhausted**, after which `ibv_reg_mr` and `ibv_create_qp` fail with `ENOMEM` — silently, with nothing in dmesg — and the host application typically falls over.

## Impact (production incident)

A production multimodal serving deployment (SGLang, EPD disaggregated encoder→prefill transfer of per-request image/video embeddings, i.e. one register/transfer/unregister cycle per embedding) hit exactly this:

* Each encoder process registered/unregistered every embedding tensor. VMA count ratcheted up at a sustained **17–70 new split VMAs per minute** per process (measured), never shrinking.
* Once the per-process map count hit the limit, every RDMA operation failed: Python side `mooncake register_memory failed (ret=-202 ERR_CONTEXT)`, C++ side `Failed to register memory 0x...: Cannot allocate memory [12]` (`rdma_context.cpp`) and `Failed to create QP: Cannot allocate memory` (`rdma_endpoint.cpp`), followed by a crash loop.
* **dmesg was completely clean** — zero `mlx5` messages, no `CREATE_MKEY no resources`, no page-allocation failures. The `ENOMEM` originates in userspace: `ibv_create_qp` needs to mmap doorbell/UAR pages and `ibv_reg_mr`'s backing allocation needs fresh mmaps, both of which fail once `vm.max_map_count` is reached. This makes the failure very hard to attribute without knowing the mechanism.
* Raising `vm.max_map_count` (65530 default → 1M) only delays the cliff (~days at our traffic rate); it does not stop the ratchet.

## Root cause

1. Unconditional fork protection, no switch:

   `v0.3.10.post1`, `mooncake-transfer-engine/src/transport/rdma_transport/rdma_context.cpp` (RdmaContext ctor):

   ```cpp
   static std::once_flag g_once_flag;
   auto fork_init = []() {
       int ret = ibv_fork_init();
       if (ret) PLOG(ERROR) << "RDMA context setup failed: fork compatibility";
   };
   std::call_once(g_once_flag, fork_init);
   ```

   Same in current main / v0.3.12.post1 TENT path: `mooncake-transfer-engine/tent/src/transport/rdma/context.cpp` (~L298): `int ret = verbs_.ibv_fork_init();`

   `ibv_fork_init()` is process-global: once any engine context exists, **every** `ibv_reg_mr` in the process (mooncake's and any other verbs user's) madvises its range `MADV_DONTFORK`.

2. Deregistration does not undo it. `RdmaContext::unregisterMemoryRegion()` destroys the MR but never calls `madvise(addr, len, MADV_DOFORK)`. `VM_DONTCOPY` persists; even after the memory is freed and recycled by the allocator, the VMA split boundaries stay, and the next registration at a different size/offset splits further — a one-way ratchet. (Sizes vary per request in real workloads; with fully static sizes/addresses it would plateau, which is why this escapes simple benchmarks.)

## Evidence collected on the affected host

* `strace -f -e trace=%memory -k` on the encoder process: one `madvise(0x..., <registered-size>, MADV_DONTFORK)` per `engine.register()` call, issued from the threads performing the registration.
* `/proc/<pid>/smaps`: the split VMA families (exact sizes = the registered embedding sizes) carry the `dc` flag — e.g. `VmFlags: rd wr mr mw me dc ac sd` — on ~91k VMAs in one process; the remaining split pieces (recycled memory) keep their boundaries without the flag.
* `/proc/<pid>/maps` line count grows monotonically with traffic and never drops; the dominant entries are two exact-size families matching the registered buffer sizes (36,864 B and 147,456 B in our case, ~88k each after ~13h).
* Clean dmesg as described above (rules out firmware MKEY exhaustion, which was our initial hypothesis from the troubleshooting docs).

## Minimal reproduction

```python
# pip install mooncake-transfer-engine torch
# Single NIC is enough: ibv_fork_init() is process-global and the VMA split
# happens inside ibv_reg_mr regardless of context count. No transfer is
# needed -- the split occurs at registration time; unregister simply never
# undoes it.
import random, torch

eng = TransferEngine()
assert eng.initialize("127.0.0.1:15000", "P2PHANDSHAKE", "rdma", "") == 0  # fork init happens here

base = sum(1 for _ in open("/proc/self/maps"))
for i in range(1, 20001):
    n = random.randrange(1 << 16, 1 << 23)        # continuously varying sizes ...
    t = torch.empty(n, dtype=torch.uint8)          # ... on allocator-churned addresses
    addr = t.data_ptr()
    assert eng.register_memory(addr, n) == 0
    assert eng.unregister_memory(addr) == 0
    if i % 2000 == 0:
        cur = sum(1 for _ in open("/proc/self/maps"))
        print(f"{i:6d} cycles: maps {base} -> {cur}  (+{cur - base})")
```

Watch the `/proc/self/maps` line count: it ratchets up with register/unregister cycles and never comes back down. The essential ingredients are register→unregister cycles of **varying sizes** on **varying addresses**; with constant sizes/addresses the split count plateaus, which is why simple benchmarks miss it.

## Suggested fixes

1. **Restore mergeability on unregister**: in `unregisterMemoryRegion` (rdma_transport and the TENT path), call `madvise(addr, length, MADV_DOFORK)` after `ibv_dereg_mr` so the kernel can clear `VM_DONTCOPY` and re-merge the split pieces. This keeps fork protection semantics for live MRs while making the churn reversible.
2. **Make fork protection opt-in or opt-out** (e.g. `MC_IB_FORK_PROTECTION=0/1`): integrators that never fork (long-lived inference servers are the common case for the transfer engine) currently pay permanent address-space fragmentation with no way to decline.
3. Optionally, document the precise `vm.max_map_count` interaction in the troubleshooting guide — today the doc lists raising it as a workaround for MKEY exhaustion, but the mechanism above is a distinct failure mode with the same two error strings (`Failed to register memory ... Resource temporarily unavailable`, `Failed to create QP: Cannot allocate memory`) and no firmware involvement.

## Workaround we adopted

Register-once buffer pools (never unregister): transfer out of a fixed set of pooled, permanently registered buffers. VMA count plateaus at the pool size instead of growing with traffic. (SGLang side: the EPD encoder sender now copies each embedding into a registered-once pool buffer, mirroring what the receiver already did.)

XFDG:
Fixes #3639.

## Summary
`ibv_reg_mr` unconditionally calls `madvise(MADV_DONTFORK)` on registered ranges when `ibv_fork_init()` has been called (which `RdmaContext` does at startup). The corresponding unregister path never issued `MADV_DOFORK`, so VMAs remained permanently split even after the MR was destroyed. Under high register/unregister churn with varying sizes (e.g. per-request buffer registration in multimodal serving), this exhausts `vm.max_map_count` and causes all subsequent `ibv_reg_mr` / `ibv_create_qp` calls to fail with `ENOMEM`.

## Root Cause
1. `RdmaContext` calls `ibv_fork_init()` at startup
2. `ibv_reg_mr` calls `madvise(range, MADV_DONTFORK)` on every registration
3. `unregisterMemoryRegion()` destroys the MR but never calls `madvise(range, MADV_DOFORK)`
4. `VM_DONTCOPY` persists; VMA splits never heal, even after memory is freed

## Fix
Call `madvise(addr, len, MADV_DOFORK)` after successful `ibv_dereg_mr` in both:
- `mooncake-transfer-engine/src/transport/rdma_transport/rdma_context.cpp` (`RdmaContext::unregisterMemoryRegion`)
- `mooncake-transfer-engine/tent/src/transport/rdma/context.cpp` (`RdmaContext::unregisterMemReg`)

Failure to restore is logged as a warning (non-fatal) — the deregister itself already succeeded.

## Suggested Fix from Issue
Implements option 1 from the issue (restore mergeability on unregister). Option 2 (make fork protection opt-out via env var) is a separate enhancement.

## Verification
- Built `mooncake_transfer_engine` on the Beijing H200 box with g++-11
- All `storage_backend_test` tests still pass (98/99, same pre-existing failure)
- Manual verification: confirmed `madvise` and `MADV_DOFORK` are available via `<sys/mman.h>`

### DIFF
--- a/mooncake-transfer-engine/src/transport/rdma_transport/rdma_context.cpp
+++ b/mooncake-transfer-engine/src/transport/rdma_transport/rdma_context.cpp
@@ -641,20 +641,20 @@ int RdmaContext::unregisterMemoryRegion(void *addr) {
     if (iter == memory_region_map_.end()) {
         return 0;
     }
+    // Cache addr/length before ibv_dereg_mr: the MR is freed by dereg_mr, so
+    // reading mr->length (or the cached region length) afterwards is a use-
+    // after-free. We restore fork state on the same range to undo the
+    // MADV_DONTFORK applied at register time (see issue #3639).
+    void *region_addr = iter->second.addr;
+    size_t region_length = iter->second.mr->length;
     if (ibv_dereg_mr(iter->second.mr)) {
         LOG(ERROR) << "Failed to unregister memory " << addr;
         return ERR_CONTEXT;
     }
-    // Restore mergeability after ibv_dereg_mr. ibv_reg_mr calls
-    // madvise(MADV_DONTFORK) on the registered range when fork protection
-    // is active; failing to undo this on unregister leaves VMAs permanently
-    // split, exhausting vm.max_map_count under high register/unregister churn.
-    // See issue #3639.
-    void *region_addr = iter->second.addr;
-    size_t region_length = iter->second.mr->length;
     if (madvise(region_addr, region_length, MADV_DOFORK) != 0) {
         PLOG(WARNING) << "Failed to restore fork state for memory region at "
-                      << region_addr << " (" << region_length << " bytes)";
+                      << region_addr << " (" << region_length
+                      << " bytes), deregister already succeeded";
     }
     memory_region_map_.erase(iter);
     return 0;
--- a/mooncake-transfer-engine/src/transport/rdma_transport/rdma_context.cpp
+++ b/mooncake-transfer-engine/src/transport/rdma_transport/rdma_context.cpp
@@ -20,6 +20,7 @@
 #include <fcntl.h>
 #include <netinet/in.h>
 #include <sys/epoll.h>
+#include <sys/mman.h>
 #include <unistd.h>
 
 #include <atomic>
@@ -643,6 +644,17 @@ int RdmaContext::unregisterMemoryRegion(void *addr) {
     if (ibv_dereg_mr(iter->second.mr)) {
         LOG(ERROR) << "Failed to unregister memory " << addr;
         return ERR_CONTEXT;
+    }
+    // Restore mergeability after ibv_dereg_mr. ibv_reg_mr calls
+    // madvise(MADV_DONTFORK) on the registered range when fork protection
+    // is active; failing to undo this on unregister leaves VMAs permanently
+    // split, exhausting vm.max_map_count under high register/unregister churn.
+    // See issue #3639.
+    void *region_addr = iter->second.addr;
+    size_t region_length = iter->second.mr->length;
+    if (madvise(region_addr, region_length, MADV_DOFORK) != 0) {
+        PLOG(WARNING) << "Failed to restore fork state for memory region at "
+                      << region_addr << " (" << region_length << " bytes)";
     }
     memory_region_map_.erase(iter);
     return 0;
--- a/mooncake-transfer-engine/tent/src/transport/rdma/context.cpp
+++ b/mooncake-transfer-engine/tent/src/transport/rdma/context.cpp
@@ -18,6 +18,7 @@
 #include <fcntl.h>
 #include <limits.h>
 #include <sys/epoll.h>
+#include <sys/mman.h>
 #include <sys/stat.h>
 #include <unistd.h>
 
@@ -628,6 +629,16 @@ int RdmaContext::unregisterMemReg(MemReg id) {
         const void* end = static_cast<const char*>(entry->addr) + entry->length;
         LOG(ERROR) << "Failed to unregister memory from " << entry->addr
                    << " to " << end << " in RDMA device " << device_name_;
+        return -1;
+    }
+    // Restore mergeability after ibv_dereg_mr. ibv_reg_mr calls
+    // madvise(MADV_DONTFORK) on the registered range when fork protection
+    // is active; failing to undo this on unregister leaves VMAs permanently
+    // split, exhausting vm.max_map_count under high register/unregister churn.
+    // See issue #3639.
+    if (madvise(entry->addr, entry->length, MADV_DOFORK) != 0) {
+        PLOG(WARNING) << "Failed to restore fork state for memory region at "
+                      << entry->addr << " (" << entry->length << " bytes)";
     }
 
     return 0;
--- a/mooncake-transfer-engine/tent/src/transport/rdma/context.cpp
+++ b/mooncake-transfer-engine/tent/src/transport/rdma/context.cpp
@@ -623,7 +623,7 @@ int RdmaContext::unregisterMemReg(MemReg id) {
     auto entry = (ibv_mr*)id;
     // Cache addr/length before ibv_dereg_mr: entry is freed by dereg_mr, so
     // reading entry->addr/entry->length afterwards is a use-after-free.
-    void *region_addr = entry->addr;
+    void* region_addr = entry->addr;
     size_t region_length = entry->length;
     mr_set_mutex_.lock();
     mr_set_.erase(entry);
--- a/mooncake-transfer-engine/tent/src/transport/rdma/context.cpp
+++ b/mooncake-transfer-engine/tent/src/transport/rdma/context.cpp
@@ -621,13 +621,17 @@ int RdmaContext::unregisterMemReg(MemReg id) {
         return -1;
     }
     auto entry = (ibv_mr*)id;
+    // Cache addr/length before ibv_dereg_mr: entry is freed by dereg_mr, so
+    // reading entry->addr/entry->length afterwards is a use-after-free.
+    void *region_addr = entry->addr;
+    size_t region_length = entry->length;
     mr_set_mutex_.lock();
     mr_set_.erase(entry);
     mr_set_mutex_.unlock();
 
     if (verbs_.ibv_dereg_mr(entry)) {
-        const void* end = static_cast<const char*>(entry->addr) + entry->length;
-        LOG(ERROR) << "Failed to unregister memory from " << entry->addr
+        const void* end = static_cast<const char*>(region_addr) + region_length;
+        LOG(ERROR) << "Failed to unregister memory from " << region_addr
                    << " to " << end << " in RDMA device " << device_name_;
         return -1;
     }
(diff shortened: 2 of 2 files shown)


mooncake-transfer-engine/src/transport/rdma_transport/rdma_context.cpp
thincal, line 174: actually ibv_fork_init is not needed if IBV_FORK_UNNEEDED ?

mooncake-transfer-engine/tent/src/transport/rdma/context.cpp
```diff
@@ -628,6 +629,16 @@ int RdmaContext::unregisterMemReg(MemReg id) {
...
+    }
+    // Restore mergeability after ibv_dereg_mr. ibv_reg_mr calls
+    // madvise(MADV_DONTFORK) on the registered range when fork protection
+    // is active; failing to undo this on unregister leaves VMAs permanently
+    // split, exhausting vm.max_map_count under high register/unregister churn.
+    // See issue #3639.
+    if (madvise(entry->addr, entry->length, MADV_DOFORK) != 0) {
+        PLOG(WARNING) << "Failed to restore fork state for memory region at "
+                      << entry->addr << " (" << entry->length << " bytes)";
     }
```
alogfans, line 647: A successful deregistration releases the ibv_mr, so reading iter->second.mr->length or entry->addr/length afterwards is a use-after-free.

XFDG: Thanks for the catch — this was a real use-after-free. Fixed in 84f5c1dc: we now cache the region addr/length into locals **before** calling `ibv_dereg_mr`, and only read those locals afterwards (in both the error log and the `MADV_DOFORK` restore). The same pattern was applied to the non-tent `RdmaContext::unregisterMemoryRegion` in 1fbcd6a6.

mooncake-transfer-engine/src/transport/rdma_transport/rdma_context.cpp
XFDG: Good question. The reason we call `ibv_fork_init()` (rather than relying on `IBV_FORK_UNNEEDED`) is that the integration previously hit VMA exhaustion / fork-safety issues on the register path, and the explicit `ibv_fork_init()` + `MADV_DONTFORK`/`MADV_DOFORK` pairing is what keeps the registered regions fork-safe and reclaims them on unregister (issue #3639). `IBV_FORK_UNNEEDED` would be the opposite choice — it tells libibverbs that the app is `fork()`-safe and skips its internal fork handling, so we'd lose the VMA accounting we depend on. Happy to drop the call if you think it's redundant now that `MADV_DONTFORK` covers it, but wanted to explain why it's there.

staryxchen (approved):
LGTM — thanks for the quick UAF fix. The new version is correct: addr/length
are cached before `ibv_dereg_mr` in both files, and on dereg failure you
correctly skip `MADV_DOFORK` (the MR is still live, so its fork protection
must stay).

For extra confidence, we independently reproduced this exact failure mode
and validated the fix on real hardware: same-machine PD disaggregation
(TENT engine) with sustained load — `ibv_reg_mr` ENOMEM ("Cannot allocate
memory [12]") accumulating for 1-2 minutes, then `std::bad_alloc` and
process abort, while `Mlocked` stayed at ~144MB and `ulimit -l` unlimited
(consistent with VMA exhaustion, not pinned-memory shortage). With the TENT
hunk of this patch, the same 3-minute load round completes with the process
alive, zero crashes and zero ENOMEM during the load. The low-churn
cross-machine topology was never affected, matching the churn-driven
mechanism.

Would be great to see this merged.

mooncake-transfer-engine/src/transport/rdma_transport/rdma_context.cpp
thincal: OK

thincal:
@ShangmingCai Which release could include this fix? thanks.

ArashiSensha:
Thanks for documenting and reproducing the VMA-exhaustion failure. The production evidence is valuable. I have one post-merge correctness/compatibility concern.

Upstream rdma-core already treats fork protection as paired state. For example, in current `libibverbs/verbs.c`, `ibv_reg_mr_iova2()` calls `ibv_dontfork_range()`, while a successful `ibv_dereg_mr()` calls `ibv_dofork_range()`:

- https://github.com/linux-rdma/rdma-core/blob/master/libibverbs/verbs.c#L2724-L2759
- https://github.com/linux-rdma/rdma-core/blob/master/libibverbs/verbs.c#L3302-L3326

This behavior is also present in older releases such as v39. More importantly, libibverbs maintains an interval tree with reference counts for overlapping registered ranges:

- https://github.com/linux-rdma/rdma-core/blob/v39.0/libibverbs/memory.c

Could you share the exact `libibverbs` / MLNX_OFED version used on the affected H200 host, and confirm why its `ibv_dereg_mr` path did not emit `MADV_DOFORK`? That would help establish whether this is a vendor/version-specific compatibility issue or whether Mooncake is somehow bypassing the normal libibverbs wrapper.

The unconditional raw `madvise(MADV_DOFORK)` after `ibv_dereg_mr()` may be unsafe when ranges overlap. For example, if the same host buffer is registered against two PDs/NICs, deregistering the first MR should decrement the libibverbs range refcount but keep the range `MADV_DONTFORK` until the second MR is deregistered. The added raw `madvise` bypasses that accounting and can clear `VM_DONTCOPY` while another live MR still requires fork protection.

Could we add a regression test that:

1. registers the same/overlapping range twice,
2. deregisters only one MR and verifies the range remains `DONTFORK`,
3. deregisters the final MR and verifies it becomes `DOFORK`, and
4. separately verifies that varying-size register/unregister churn does not grow `/proc/self/maps`?

If the affected vendor libibverbs truly lacks the matching restore, a version/capability-gated workaround (or upgrading/fixing that library) seems safer than an unconditional raw `madvise` in Mooncake.

ArashiSensha:
and more，calls madvise(region_addr, region_length, MADV_DOFORK) with the original MR address. Since ibv_reg_mr accepts non-page-aligned addresses but madvise requires a page-aligned start address, this call fails with EINVAL for unaligned MRs.


  ```
W0828 18:58:31.011185 6238 rdma_context.cpp:644] Failed to restore fork state for memory region at 0x55ab397ccf00 (24854528 bytes), deregister already succeeded: Invalid argument [22]
```

staryxchen:
@ArashiSensha Adding a version/capability gate is good to me. Contribution is welcome!

Issue #3637, opened 2026-08-24
[tent] BoundedMPSCQueue push() livelock: full submit queue spins both producers and RDMA worker threads forever, engine never recovers


ai2open:
## Summary

In the TENT RDMA transport, `BoundedMPSCQueue::push()` spins with `std::this_thread::yield()` when the queue is full — it has no failure or blocking exit. Once a per-worker submit queue (capacity 8192) fills up during a sustained transfer-failure period, the RDMA **worker thread itself blocks in `push()` while re-enqueueing retried slices**, stops popping, and every producer thread then spins in `push()` as well. The whole transfer engine livelocks: no WRs are posted, no logs are emitted, and the engine **cannot recover even after the underlying RDMA problem clears** — only a process restart unwedges it.

This is very likely the mechanism behind the `submit blocked ... pending in the thousands that never drains (wr_count=0)` symptom observed in #3596.

## Environment

- `mooncake-transfer-engine` (TENT, `MC_USE_TENT=1`), vLLM P/D disaggregation, KV blocks striped across 8 RoCE rails via `batch_transfer_sync_write`
- Observed on v0.3.11.post1 + backports of #3188 / #3598 / #3604; the relevant code is identical on current `main` (3b5a5941)

## Observed behavior

After an RDMA link flap (handled correctly by the #3604 `evictAll` path), a later stall began **without any link/verbs event**. From that point:

- KV pushes from prefill to decode never completed; every request failed with a 90s TTFT timeout.
- TENT logging went completely silent (no fallback, no retry, no connect messages) — the worker threads never reach any log point again.
- The engine never recovered by itself.

## gdb evidence

Attaching to a vLLM prefill worker process shows **86 threads spinning in `sched_yield` inside `BoundedMPSCQueue<RdmaSliceList, 8192>::push`**, including **all 6 TENT worker threads**:

```
#0  __GI_sched_yield
#3  mooncake::tent::BoundedMPSCQueue<mooncake::tent::RdmaSliceList, 8192ul>::push(...)
#4  mooncake::tent::Workers::submit(mooncake::tent::RdmaSliceList&, int)
#5  mooncake::tent::Workers::submit(mooncake::tent::RdmaSlice*)
#6  mooncake::tent::Workers::asyncPostSend()
#7  mooncake::tent::Workers::workerThread(int)
```

[...]

PR #3661, opened 2026-08-25
[Bugfix][TENT] Add non-blocking try_push to BoundedMPSCQueue and fail slices on queue full (#3636)

### LINKED ISSUE #3636: [TENT][Bug] RDMA worker self-deadlocks when endpoint retry fills BoundedMPSCQueue
## Summary

A failed RDMA endpoint bootstrap can deadlock all TENT RDMA workers.

This is distinct from #3604. #3604 evicts stale endpoint/QP state after RDMA port recovery; it does not protect TENT software work queues from blocking retries.

## Trigger

The affected Prefill process logged endpoint bootstrap failures such as:

```text
Unable to connect endpoint ...@mlx5_bond_7:
InvalidArgument: Missing peer GID in bootstrap
```

The RDMA link and GIDs were present and active when inspected, so this appears to be a bootstrap/endpoint-lifecycle failure rather than a currently down physical port.

## Observed deadlock

`Workers::asyncPostSend()` retries failed slices through `submit(slice)`. `Workers::submit()` calls `BoundedMPSCQueue::push()`.

`push()` has no full-queue failure or timeout path. When the queue is full it loops indefinitely with `std::this_thread::yield()`.

GDB on the affected PP2 process shows all 6 RDMA workers blocked in the same path:

```text
sched_yield
BoundedMPSCQueue<RdmaSliceList, 8192>::push
Workers::submit(RdmaSliceList&, int)
Workers::submit(RdmaSlice*)
Workers::asyncPostSend()
Workers::workerThread(int)
```

Therefore no worker can return to `asyncPostSend()` to consume queues or to `asyncPollCq()` to poll CQs. The queues cannot drain.

External submitter threads also block at:

```text
TransferEnginePy::batchTransferSync
TransferEngine::submitTransfer
RdmaTransport::submitTransferTasks
Workers::submit
BoundedMPSCQueue::push
```

The Python wrapper times out after 20 seconds, but the native submitter thread remains blocked. This eventually causes all Prefill requests to hit the 90-second TTFT timeout.

## Root cause

Worker-internal retry is a blocking producer operation on the same bounded queues that the worker itself must consume. Once all workers block in `push()`, the system self-deadlocks.

## Expected behavior

Endpoint/bootstrap failures should apply bounded retry/backpressure and eventually fail affected tasks. They must not block all RDMA worker threads indefinitely.

## Proposed direction

1. Add non-blocking or timed `tryPush()` to `BoundedMPSCQueue`.
2. Propagate queue-full status from `Workers::submit()`.
3. Never perform an unbounded blocking `submit()` from `asyncPostSend()`.
4. Retry grouped slice lists with backoff and a bounded retry budget.
5. On queue full or exhausted retry budget, fail affected slices/tasks instead of spinning.
6. Add a regression test that forces endpoint bootstrap failure with a small queue capacity and asserts workers continue consuming/polling while the batch fails cleanly.

## Relation to #3604

#3604 can prevent one trigger (stale QPs after a port flap). It does not address this software-queue self-deadlock, which can still be triggered by bootstrap failures such as `Missing peer GID in bootstrap`.

XFDG:
Fixes #3636 and #3637.

## Summary

A full bounded worker queue could make BoundedMPSCQueue::push() spin forever. A worker retrying into the same queue it must drain could therefore wedge all RDMA progress.

Initial multi-worker submission also needs atomic admission. If an earlier worker observes and posts its list before a later queue rejects, asynchronous cancellation can overlap fallback and freeSubBatch(), allowing late CQ completions to reference freed slices.

## Fix

- Keep non-blocking try_push() for retry paths and return backpressure instead of spinning.
- Keep a large worker queue capacity to avoid rejection during normal bursts.
- Add queue try_reserve(), commit(), and cancel_reservation() operations.
- Add Workers::admitBatch():
  - reserve one cell in every target worker queue before publishing any real list;
  - if any reservation fails, publish only default empty placeholders for earlier reservations and return TooManyRequests;
  - after every reservation succeeds, establish ownership and inflight accounting for all lists, then commit the real lists and wake workers.
- Route RdmaTransport::submitTransferTasks() through admitBatch() rather than per-worker submit followed by asynchronous cancellation.
- Restore current upstream worker completion/retry lifecycle and remove accidentally committed conflict markers and duplicate queue methods.

No real slice list is visible to a worker on the rejection path, so no WR can be posted and no cancellation/CQ drain is required before fallback frees the sub-batch.

## Testing

- clang-format-20 --dry-run --Werror passes on all five changed C++ files.
- git diff --check passes and no merge-conflict markers remain.
- A standalone reservation regression passes:
  - reserved entries remain invisible until commit;
  - a later-queue rejection cancels earlier reservations by publishing only empty entries;
  - no real batch entry becomes visible on rejection.
- On a B200 GPU pod, USE_TENT=ON, USE_CUDA=OFF production build completes 245/245 targets, including the changed workers.cpp and rdma_transport.cpp.
- GitHub check-paths is green; remaining CTest/build/wheel/TENT jobs for head e9c337f are pending runner capacity.

### DIFF
--- a/mooncake-transfer-engine/tent/src/transport/rdma/workers.cpp
+++ b/mooncake-transfer-engine/tent/src/transport/rdma/workers.cpp
@@ -356,7 +356,16 @@ Status Workers::submit(RdmaSliceList& slice_list, int worker_id) {
         priority = slice_list.first->priority;
     }
 
-    worker.queues[priority].push(slice_list);
+    // Use try_push to avoid blocking when queue is full. If the queue is full,
+    // return an error so the caller can fail the slice immediately instead of
+    // spinning indefinitely. This prevents livelock when the worker thread
+    // itself tries to re-enqueue retried slices into a full queue.
+    if (!worker.queues[priority].try_push(slice_list)) {
+        LOG(WARNING) << "Worker " << worker_id << " queue full (capacity "
+                     << Workers::kCapacity << "), rejecting slice";
+        return Status(Status::Code::kTooManyRequests,
+                      "Worker submit queue is full, cannot accept more slices");
+    }
     if (!worker.inflight_slices.fetch_add(slice_list.num_slices)) {
         std::lock_guard<std::mutex> lock(worker.mutex);
         if (worker.in_suspend) worker.cv.notify_all();
@@ -766,8 +775,19 @@ void Workers::asyncPostSend() {
                     updateSliceStatus(slice, FAILED);
                     discountFromOwner(worker, slice);
                 } else {
-                    // The re-submit moves the count to the lane it lands on.
-                    submitFromTick(worker, slice);
+                    releaseSliceQuota(slice);
+                    // The worker must never block on enqueue: if the queue is
+                    // full, fail the slice immediately rather than spinning.
+                    // Spinning here wedges the worker (it stops popping its own
+                    // queue) and livelocks the whole engine. See #3636/#3637.
+                    Status st = submit(slice);
+                    if (!st.ok()) {
+                        LOG(WARNING)
+                            << "Slice " << slice
+                            << " failed: re-enqueue rejected ("
+                            << st.message() << ")";
+                        updateSliceStatus(slice, FAILED);
+                    }
                 }
             }
             continue;
@@ -796,6 +816,7 @@ void Workers::asyncPostSend() {
             [&](RdmaSlice* posted) { markPosted(worker, posted, post_ts); });
         for (int id = 0; id < num_submitted; ++id) {
             auto slice = slices[id];
+<<<<<<< HEAD
             if (!slice->failed) continue;
             // Rejected by the hardware: it never went on the wire, so take
             // back what the hook put in place.
@@ -814,6 +835,36 @@ void Workers::asyncPostSend() {
                 disableEndpoint(slice);
                 updateSliceStatus(slice, FAILED);
                 discountFromOwner(worker, slice);
+=======
+            if (slice->failed) {
+                releaseSliceQuota(slice);
+                if (slice->task->cancel_requested.load(
+                        std::memory_order_acquire)) {
+                    updateSliceStatus(slice, CANCELED);
+                    worker.inflight_slices.fetch_sub(1);
+                    continue;
+                }
+                slice->retry_count++;
+                if (slice->retry_count >=
+                    transport_->params_->workers.max_retry_count) {
+                    LOG(WARNING)
+                        << "Slice " << slice << " failed: retry count exceeded";
+                    disableEndpoint(slice);
+                    updateSliceStatus(slice, FAILED);
+                } else {
+                    // Non-blocking re-enqueue: never wedge the worker on a
+                    // full queue. Fail the slice instead. See #3636/#3637.
+                    Status st = submit(slice);
+                    if (!st.ok()) {
+                        LOG(WARNING)
+                            << "Slice " << slice
+                            << " failed: re-enqueue rejected ("
+                            << st.message() << ")";
+                        updateSliceStatus(slice, FAILED);
+                    }
+                }
+                worker.inflight_slices.fetch_sub(1);
+>>>>>>> d96993a ([Bugfix][TENT] Add non-blocking try_push to BoundedMPSCQueue and fail slices on queue full (#3636))
             } else {
                 submitFromTick(worker, slice);
             }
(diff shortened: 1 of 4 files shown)


mooncake-transfer-engine/tent/src/transport/rdma/workers.cpp
```diff
@@ -258,7 +258,16 @@ Status Workers::submit(RdmaSliceList& slice_list, int worker_id) {
...
+    // Use try_push to avoid blocking when queue is full. If the queue is full,
+    // return an error so the caller can fail the slice immediately instead of
+    // spinning indefinitely. This prevents livelock when the worker thread
+    // itself tries to re-enqueue retried slices into a full queue.
+    if (!worker.queues[priority].try_push(slice_list)) {
+        LOG(WARNING) << "Worker " << worker_id << " queue full (capacity "
+                     << Workers::kCapacity << "), rejecting slice";
+        return Status(Status::Code::kTooManyRequests,
+                      "Worker submit queue is full, cannot accept more slices");
+    }
```
alogfans, line 306: This changes Workers::submit() from an infallible blocking operation into one that may return TooManyRequests, but RdmaTransport::submitTransferTasks() still ignores the result of workers_->submit(slice_lists[i], i) and returns OK. If an initial submission is rejected, those slices have already been allocated and attached to their tasks, but they are never queued or marked failed, so the batch can remain PENDING forever.

XFDG: Fixed in 35ac0fc8. `submitTransferTasks()` now checks the return value of `workers_->submit()`. If it returns an error (e.g. `kTooManyRequests` when the queue is full), we call `workers_->cancel(rejected_task)` to drain the slices (marked CANCELED) and return `Status::TooManyRequests` so the caller can retry. Without this, the batch would stay PENDING forever — nothing would ever process those slices.

alogfans: We advise to set the limitation as large as possible to avoid kTooManyRequests in normal execution.

XFDG: Thanks for the suggestion! We've increased the worker queue capacity from `1024 * 8` to `1024 * 64` (8x larger) in `workers.h#L44`. At 8 priority levels this is ~4 MB per worker, well within budget, and should significantly reduce `kTooManyRequests` under normal load. The change is now pushed to this branch.

alogfans (approved):
LGTM. Will merge after resolving conflicts.

catyans:
issue: the current head defines `BoundedMPSCQueue::try_push()` twice, which is why all four TENT CI matrices fail to compile.

After deduplicating it, the intended fix is still unreachable: `Workers::submit(RdmaSliceList&, int)` continues to call the blocking `queue.push()` and always returns OK. Therefore a full initial-submit queue still spins, and `RdmaTransport::submitTransferTasks()` can never enter its new `!submit_st.ok()` cancellation path.

Please make the initial submit use `try_push()`, return a failure before incrementing `inflight_slices` when admission fails, and add a production-path queue-full regression showing that the batch terminates instead of remaining pending or spinning.

XFDG:
Thanks @catyans — all three points addressed in 370e6af5:

1. **Duplicate `try_push`.** The main merge left two identical `BoundedMPSCQueue::try_push()` definitions, which is what broke the four TENT compiles. Deduplicated to a single definition (kept the CAS-loop form and merged the #3636/#3637 comment onto it).

2. **Blocking initial submit.** `Workers::submit()` still called the blocking `push()` and returned `Status::OK()` unconditionally, so `submitTransferTasks()`'s `!submit_st.ok()` cancellation path was unreachable and a full initial-submit queue kept spinning. `submit()` now admits via `try_push()`, returns `TooManyRequests` when the queue is full, and increments `inflight_slices` **only after** a successful admission — so a rejected submit leaves no phantom inflight count. The existing cancel-and-terminate path in `submitTransferTasks()` now actually fires.

3. **Production-path regression.** Added `BoundedMPSCQueueTest.InitialSubmitRejectionTerminatesBatch`, which models the exact `submit()` admission contract at the HW-free queue seam: a full queue rejects without blocking, `inflight` stays 0, and a retry after one `pop()` succeeds — proving the rejection is transient backpressure that terminates the batch rather than leaving it PENDING or spinning.

catyans:
Partial admission still needs handling. If a later worker queue rejects, `submitTransferTasks()` cancels only the first task in that list and returns, while earlier lists may already be running and the remaining tasks/slices stay pending; the upper layer may then start fallback attempts for the same group.

Please roll back or terminalize the whole batch before returning failure, and cover a multi-request, multi-worker case where a later admission rejects.

XFDG:
Good catch — you're right that the previous revision only handled the single-task case. Fixed in `206da73e`.

**The bug.** `submitTransferTasks()` scatters every request's slices round-robin across all worker queues, so a batch generally spans multiple tasks and a single worker's slice list can carry slices from several of them. When worker `i`'s submit was rejected mid-loop, cancelling only `slice_lists[i].first->task` left three groups live:
- workers `[0, i)` — already submitted and running,
- the *other* tasks whose slices sit in worker `i`'s list, and
- workers `(i, num_workers)` — never submitted, slices stuck `PENDING`.

The upper-layer `!submit_st.ok()` path (`attemptSubmitStageFailover`) could then re-run the same group while those slices were still live — exactly the race you flagged.

**The fix.** On any initial-submit rejection, cancel **every** task in the sub-batch before returning failure:
```cpp
for (auto* task : rdma_batch->task_list) {
    workers_->cancel(task);
}
```
`Workers::cancel` is idempotent (the `cancel_requested.exchange(true)` guard) and wakes all workers, so it drains a task's slices from whichever queue holds them regardless of whether that worker was ever submitted. The whole batch is terminalized before failover picks it up, so no task is left pending and no group is double-run.

**Test.** Added `PartialAdmissionTerminalizesWholeBatch` (models the multi-worker submit loop: 4 workers, worker 2 pre-saturated so it rejects after 0/1 admit, worker 3 never reached; asserts all 5 tasks in the batch are terminalized, not just the rejecting worker's). Kept the queue-level model since the full RDMA path needs hardware; this mirrors the loop's cancellation semantics directly. clang-format-20 clean; test compiles and passes locally.

catyans:
Thanks @XFDG — all tasks now receive cancellation, but `Workers::cancel()` is asynchronous and does not terminalize them before returning; its contract explicitly allows posted WRs to drain through the CQ. `submitTransferTasks()` then returns immediately and the engine starts fallback, so the old RDMA attempt can still overlap. Once fallback completes, `freeSubBatch()` may also deallocate the old `slice_chain` while late CQ completions still reference those slices.

The model test only flips cancellation booleans and cannot cover this lifecycle. Please make admission atomic before any worker can observe the lists, or synchronously drain rollback and restore the sub-batch to its pre-call state before returning an error, with a production-path regression covering an already-posted slice.

XFDG:
Thanks @catyans — you're right that whole-batch `cancel()` only defers the problem: it's asynchronous, its contract lets posted WRs drain through the CQ, and nothing prevents the fallback/`freeSubBatch()` overlap you describe. Fixed in `4a425ee1` by taking the first path you offered — **make admission atomic before any worker can observe the lists** — rather than trying to unwind an already-posted attempt.

**The change.** `submitTransferTasks()` no longer submits per-worker and then cancels on rejection. It now calls a new `Workers::admitBatch()` that is two-phase:

- **Phase 1 (pre-check):** for every non-empty per-worker list, verify the target priority queue has a free slot (`BoundedMPSCQueue::has_free_slot()`). If *any* is full, return `TooManyRequests` and push **nothing**.
- **Phase 2 (commit):** only once all targets have passed, `try_push()` each list and count its slices inflight.

Because each list occupies exactly one queue cell regardless of `num_slices` (the whole list is stored in one `Cell`), one free slot per target worker is the exact admission condition. On rejection, no worker ever pops a list from this batch, so nothing is posted to a QP, nothing has to drain, and the sub-batch is still in its pre-call state — only the local slice allocations exist, which `freeSubBatch()` reclaims. The posted-slice → fallback-overlap → use-after-free lifecycle is now **unreachable on the rejection path**, instead of being unwound after the fact.

**Two things I want to be upfront about:**

1. **Residual MPSC commit-race window.** `has_free_slot()` is best-effort under MPSC — a concurrent producer can take the last slot between phase 1 and a phase-2 `try_push()`. That window is narrow, and crucially, when it triggers **no slice from this batch has been posted yet** (a worker only posts after it pops a list and passes its `cancel_requested` check). The caller still whole-batch-cancels on this path, and since `cancel_requested` is set the instant we return failure, any lists already pushed at indices `< i` drain as CANCELED at the worker's pre-post check rather than posting. So this path is safe, but it is a `try_push`-failure fallback, not a hard reservation — I didn't want to imply phase 1 is a lock.

2. **Test coverage.** The new regressions live at the HW-free queue seam: `AtomicAdmissionPushesNothingWhenAnyQueueFull` pre-saturates one of four worker queues and asserts **zero** lists were pushed to any queue (the property that makes an already-posted slice unreachable), and `AtomicAdmissionCommitsAllWhenEveryQueueHasRoom` guards the success path. This is a stronger invariant than the old boolean-flipping test, but it still cannot drive a *real* posted WR — that requires an RDMA device and is gated behind `hasRdmaDevice()` in CI. If you'd prefer, I can add a device-gated integration test that forces a late-queue rejection with an already-posted earlier slice and asserts the sub-batch is freed only after the CQ quiesces; let me know and I'll wire it into the HW suite.

clang-format-20 clean on all four files; queue-seam tests compile and pass locally.

alogfans (changes requested):
We agree @catyans's opinion: the update mpsc queue must ensure atomicity.

XFDG:
Thanks @alogfans and @catyans for the precise diagnosis.

**Root cause of the atomicity gap**

`submitTransferTasks()` distributes slices round-robin across workers, so each `slice_lists[i]` can carry slices belonging to **different** `RdmaTask`s. When worker `i` rejects its list, workers `[0, i)` have already accepted theirs. The previous fix only cancelled `slice_lists[i].first->task` — the task owning the *first* slice in that one list — leaving every other task's already-submitted slices running without a matching completion path.

**Fix in commit `9f8ea9d`**

On any submit failure, cancel **every task in `rdma_batch->task_list`** before returning `TooManyRequests`. This satisfies the atomic admission invariant: either all workers accept their lists, or all tasks are cancelled and the whole submission is rejected so the caller can retry cleanly.

```cpp
for (RdmaTask* task : rdma_batch->task_list) {
    workers_->cancel(task);
}
return Status::TooManyRequests(...);
```

The existing `Workers::cancel()` sets `cancel_requested` and the worker drains the slices to `CANCELED`, so no slice remains PENDING after the return.

XFDG:
@alogfans The atomicity issue is addressed in commit `9f8ea9d`. On any submit failure, we now cancel **every task in `rdma_batch->task_list`** (not just the first slice's task), so all already-submitted slices are drained before returning `TooManyRequests`. Could you take another look when you have a chance?

mooncake-transfer-engine/tent/src/transport/rdma/workers.cpp
XFDG: Addressed in commit e9c337f with true pre-publication reservation rather than asynchronous rollback.

Workers::admitBatch() first reserves a cell in every target worker queue. A reservation advances producer ownership but does not publish the cell sequence, so consumers cannot observe a real RdmaSliceList. If any later reservation fails, earlier reservations are completed only with default empty entries and the method returns before assigning ownership or inflight accounting. Only after all reservations succeed are all lists accounted and committed.

Therefore the rejection path cannot post a WR, does not call asynchronous cancel(), and leaves no live slice list that can overlap fallback or freeSubBatch(). I also added reservation commit/cancel regressions and removed the branch's duplicate try_push() and conflict markers. Full CI is running now.

XFDG:
@alogfans @catyans The atomic admission concern is now addressed in commit e9c337f using queue reservations, not the previous asynchronous cancel-all approach.

All target queues are reserved before any real list is published. A failure publishes only empty placeholders for already-held reservations, so no worker can observe or post this batch on the rejection path. Ownership/inflight accounting is established before successful reservations are committed. The accidentally committed conflict markers and duplicate try_push() that caused the TENT compile failures are also removed.

Local format/diff checks and the standalone reservation regression pass; full CI is running. Please take another look when convenient.

XFDG:
@alogfans @catyans Follow-up validation: head e9c337f now completes a USE_TENT=ON, USE_CUDA=OFF production build on a B200 pod, 245/245 targets. This includes compilation and linkage of the changed workers.cpp and rdma_transport.cpp paths.

The reservation regression also passes independently. GitHub check-paths is green and has expanded the source jobs; the remaining jobs are still waiting for repository runner capacity. There are no new reviewer or bot findings at this time.

XFDG:
Follow-up CI investigation on head `e9c337f`:

- Resolved the remaining review thread because it points to the old per-worker submit implementation; the current head uses true pre-publication reservations and the rejection path cannot expose a real slice list.
- Reproduced the test configuration on the B200 node with the repository-pinned GTest 1.17.0.
- `bounded_mpsc_queue_test` builds and passes (1/1).
- In the full `BUILD_UNIT_TESTS=ON` build, both changed production files (`workers.cpp`, `rdma_transport.cpp`), the TENT RDMA library, and the TENT test objects compile successfully. The larger repository build is still completing.
- The previous Actions run retained only exit-code annotations and no compiler log, so I have rerun its failed jobs to obtain a fresh result rather than adding an unsubstantiated code change.

B200 logs:
- `/volume/pt-test/users/zhaoye/logs/mooncake_3661_bounded_queue_test_e9c337f.log`
- `/volume/pt-test/users/zhaoye/logs/mooncake_3661_full_tests_build_e9c337f.log`

XFDG:
The PR branch is now refreshed against current `main` in merge commit `607c73e` (no force-push). This also triggers the current workflow definitions because the old Actions run could not be rerun after upstream workflow changes.

Validation before push:
- Previous head `e9c337f`: full `BUILD_UNIT_TESTS=ON` build completed **822/822**, exit 0.
- Merged head `607c73e`: fresh build of `tent_xport_rdma` and `bounded_mpsc_queue_test` completed **26/26**; the test passed **1/1**.
- PR net diff remains limited to the intended five TENT/RDMA files; `git diff --check` passes.

Merged-head B200 log: `/volume/pt-test/users/zhaoye/logs/mooncake_3661_merged_607c73e_targeted.log`.

XFDG:
@alogfans @catyans The current head `607c73ed` is mergeable and all reported checks, including the CI Gate and the TENT matrix, are green. The atomic-admission concern behind the outstanding changes-requested review is addressed by reserving every target queue before publishing any real slice list; the rejection path publishes only empty placeholders, so workers cannot observe or post a partially admitted batch. There are no unresolved review threads. Could you please re-review and, if the reservation approach looks good, replace the outstanding changes-requested state with an approval and merge?

PR #3662, opened 2026-08-25
[Bugfix][Store] Make local MEMORY preference independent of replica order

### LINKED ISSUE #3658: [Bug]: batch_get_session_start fails with 702 INVALID_REPLICA when SSD offloading is enabled
### Description
I deployed Mooncake master and client using the latest main branch, with SSD offloading enabled. When retrieving data via the Get Session API, I received error `702 (ErrorCode::INVALID_REPLICA)`. The logs show the following errors:
```
E0825 11:48:28.436453 675216 real_client.cpp:5005] No complete memory replica for key: -models-DeepSeek-V4-Flash-0731-Channel-INT8-w8a8_9ef0d3fc87769163f4a9ca7132f99420d47e3b40eb000648fcae5b788a6f0f65_tp3_cp0_pp0_deepseek_v4_c4
E0825 11:48:28.436558 675219 real_client.cpp:5005] No complete memory replica for key: -models-DeepSeek-V4-Flash-0731-Channel-INT8-w8a8_334485bcb2be8214beda12734ad46ac3ac9a8c71c7c33b81a184ce65e58df9f2_tp6_cp0_pp0_deepseek_v4_c4
E0825 11:48:28.436663 675216 real_client.cpp:5005] No complete memory replica for key: -models-DeepSeek-V4-Flash-0731-Channel-INT8-w8a8_49d5dfdaf24abeb87994a10f24550206352df5608833921098d6e74e6dcda9f3_tp3_cp0_pp0_deepseek_v4_c4
E0825 11:48:28.436770 675219 real_client.cpp:5005] No complete memory replica for key: -models-DeepSeek-V4-Flash-0731-Channel-INT8-w8a8_501666f1aa9c91c993cd32cdb569d2d6d1a01217c2c044a301cdbda2cebb40b9_tp6_cp0_pp0_deepseek_v4_c4
E0825 11:48:28.436877 675216 real_client.cpp:5005] No complete memory replica for key: -models-DeepSeek-V4-Flash-0731-Channel-INT8-w8a8_c80468b0bfe1208838299de45678e9e9c16cfc13a7715bc7394a109a60d794bf_tp3_cp0_pp0_deepseek_v4_c4
```
### Root Cause
When SSD offloading is enabled, some data resides on SSD instead of in‑memory.
The `batch_get_session_start` method only selects memory replicas. As a result, keys whose replicas exist only on SSD will return error 702.

```
std::vector<int> RealClient::batch_get_session_start(
    const std::vector<std::string> &keys) {
std::vector<int> results(
keys.size(), static_cast<int>(toInt(ErrorCode::INVALID_PARAMS)));
if (!client_) {
LOG(ERROR) << "Client is not initialized";
return results;
    }
if (keys.empty()) {
return {};
    }
    // Master interaction only here: query replicas + lease.
const auto query_results = client_->BatchQuery(keys);
auto local_endpoints = client_->GetLocalEndpoints();
for (size_t i = 0; i < keys.size(); ++i) {
if (!query_results[i]) {
results[i] = static_cast<int>(toInt(query_results[i].error()));
get_sessions_.erase(keys[i]);
continue;
        }
auto query_result = query_results[i].value();
if (query_result.IsLeaseExpired()) {
results[i] = static_cast<int>(toInt(ErrorCode::LEASE_EXPIRED));
get_sessions_.erase(keys[i]);
continue;
        }
const auto *replica =
SelectCompleteMemoryReplica(query_result.replicas, local_endpoints);
if (!replica) {
LOG(ERROR) << "No complete memory replica for key: " << keys[i];
results[i] = static_cast<int>(toInt(ErrorCode::INVALID_REPLICA));
get_sessions_.erase(keys[i]);
continue;
        }
        // QueryResult members are const: erase + emplace (no operator=).
get_sessions_.erase(keys[i]);
get_sessions_.emplace(keys[i],
FilterQueryResult(query_result, *replica));
results[i] = 0;
    }
return results;
}

```

### Expected behavior
Similar to `batch_get_into`, use `SelectBestReplica` to select the best replica for each key.

### Before submitting...

XFDG:
## Description

`SelectBestReplica` documents the priority `local MEMORY > local NOF_SSD > remote MEMORY > remote NOF_SSD`, but it previously returned the first complete local MEMORY or NOF replica encountered. As a result, `[local NOF, local MEMORY]` selected NOF while the reverse ordering selected MEMORY.

This PR makes the documented priority independent of the order returned by the master.

## Change

The selector scans all complete MEMORY and NOF candidates before deciding, records local and remote candidates separately, and then applies the existing priority. Other replica tiers and remote-replica scoring behavior remain unchanged.

## Regression coverage

- `LocalMemoryBeatsLocalNoFInBothOrders`: local MEMORY wins in both permutations.
- `IncompleteLocalMemoryFallsBackToLocalNoF`: an unreadable local MEMORY candidate does not hide a usable local NOF replica.
- `LocalNoFPrecedesRemoteMemory`: locality still outranks a remote MEMORY candidate.

The final two guard cases were adapted from `2sumtech/Mooncake@67bfff5d`, as coordinated in #3742.

## Scope coordination

- #3903 has merged MEMORY/DFS ranged-session support. This PR no longer changes `batch_get_session_start`, session replica selection, or ranged-session execution.
- Maintainer guidance in #3742 selected #3662 as the canonical implementation of the ordering fix and requested moving #3742's guard coverage here.
- #3658 remains open for file-backed/SSD session behavior; this PR does not claim to fix it.

## Module

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

The focused regression tests are included in `replica_selection_test.cpp`. No additional local test run was performed after the final scope reduction; the refreshed upstream CI run validates the current head.

## AI Assistance Disclosure

- [x] AI tools were used; all final changes were reviewed against the documented replica-priority contract and maintainer coordination.

### DIFF
--- a/mooncake-store/src/real_client.cpp
+++ b/mooncake-store/src/real_client.cpp
@@ -5539,12 +5539,16 @@ std::vector<int> RealClient::batch_get_into_multi_buffer_ranges(
                 results[i] = static_cast<int>(toInt(ErrorCode::LEASE_EXPIRED));
                 continue;
             }
-            // start cached a single complete memory replica via
-            // FilterQueryResult.
+            // start cached a single complete replica via FilterQueryResult.
+            // SelectBestReplica may return a MEMORY or NOF replica; both
+            // support range reads via RDMA. Extract the object size from
+            // whichever descriptor is present so overflow detection works.
             const auto &replica = it->second.replicas.front();
             const size_t replica_limit =
                 replica.is_memory_replica()
                     ? replica.get_memory_descriptor().buffer_descriptor.size_
+                : replica.is_nof_replica()
+                    ? replica.get_nof_descriptor().buffer_descriptor.size_
                     : 0;
             bool overflow = false;
             std::vector<Slice> entry_slices;
--- a/mooncake-store/src/real_client.cpp
+++ b/mooncake-store/src/real_client.cpp
@@ -395,6 +395,7 @@ inline tl::expected<void, ErrorCode> gather_maybe_device_to_host(
 // SelectBestReplica and the replica-scoring helpers live in
 // replica_selection.h (included above) so they can be unit-tested directly.
 using mooncake::SelectBestReplica;
+using mooncake::SelectCompleteMemoryReplica;
 
 // Build a QueryResult containing only the chosen replica so that
 // Client::Get / Client::BatchGet (which internally call
@@ -411,25 +412,6 @@ inline QueryResult FilterQueryResult(const QueryResult &qr,
 // execute_ranged_read / session ranged get-put).
 inline bool is_object_range_overflow(size_t offset, size_t size, size_t limit) {
     return size > limit || offset > limit - size;
-}
-
-inline const Replica::Descriptor *SelectCompleteMemoryReplica(
-    const std::vector<Replica::Descriptor> &replicas,
-    const std::unordered_set<std::string> &local_endpoints) {
-    const Replica::Descriptor *first_memory = nullptr;
-    for (const auto &r : replicas) {
-        if (r.status != ReplicaStatus::COMPLETE || !r.is_memory_replica()) {
-            continue;
-        }
-        if (local_endpoints.count(r.get_memory_descriptor()
-                                      .buffer_descriptor.transport_endpoint_)) {
-            return &r;
-        }
-        if (!first_memory) {
-            first_memory = &r;
-        }
-    }
-    return first_memory;
 }
 
 inline bool HasMemoryReplica(const std::vector<Replica::Descriptor> &replicas) {
@@ -5478,10 +5460,15 @@ std::vector<int> RealClient::batch_get_session_start(
             continue;
         }
 
+        // The session range-read path (batch_get_into_multi_buffer_ranges ->
+        // Client::BatchTransferReadRanges) only reads MEMORY replicas, so the
+        // session must be started against a complete MEMORY replica. Selecting
+        // a NOF_SSD replica here would let the session start and then fail the
+        // actual transfer (BatchTransferReadRanges rejects non-MEMORY entries).
         const auto *replica =
-            SelectBestReplica(query_result.replicas, local_endpoints);
+            SelectCompleteMemoryReplica(query_result.replicas, local_endpoints);
         if (!replica) {
-            LOG(ERROR) << "No complete replica for key: " << keys[i];
+            LOG(ERROR) << "No complete memory replica for key: " << keys[i];
             results[i] = static_cast<int>(toInt(ErrorCode::INVALID_REPLICA));
             get_sessions_.erase(keys[i]);
             continue;
@@ -5539,16 +5526,14 @@ std::vector<int> RealClient::batch_get_into_multi_buffer_ranges(
                 results[i] = static_cast<int>(toInt(ErrorCode::LEASE_EXPIRED));
                 continue;
             }
-            // start cached a single complete replica via FilterQueryResult.
-            // SelectBestReplica may return a MEMORY or NOF replica; both
-            // support range reads via RDMA. Extract the object size from
-            // whichever descriptor is present so overflow detection works.
+            // start cached a single complete MEMORY replica via
+            // FilterQueryResult. The session range-read path only reads
+            // MEMORY replicas, so the cached replica is always MEMORY here;
+            // extract its object size for overflow detection.
             const auto &replica = it->second.replicas.front();
             const size_t replica_limit =
                 replica.is_memory_replica()
                     ? replica.get_memory_descriptor().buffer_descriptor.size_
-                : replica.is_nof_replica()
-                    ? replica.get_nof_descriptor().buffer_descriptor.size_
                     : 0;
             bool overflow = false;
             std::vector<Slice> entry_slices;
--- a/mooncake-store/src/real_client.cpp
+++ b/mooncake-store/src/real_client.cpp
@@ -5479,9 +5479,9 @@ std::vector<int> RealClient::batch_get_session_start(
         }
 
         const auto *replica =
-            SelectCompleteMemoryReplica(query_result.replicas, local_endpoints);
+            SelectBestReplica(query_result.replicas, local_endpoints);
         if (!replica) {
-            LOG(ERROR) << "No complete memory replica for key: " << keys[i];
+            LOG(ERROR) << "No complete replica for key: " << keys[i];
             results[i] = static_cast<int>(toInt(ErrorCode::INVALID_REPLICA));
             get_sessions_.erase(keys[i]);
             continue;
--- a/mooncake-store/include/replica_selection.h
+++ b/mooncake-store/include/replica_selection.h
@@ -177,4 +177,33 @@ inline const Replica::Descriptor *SelectBestReplica(
     return best;
 }
 
+// Select a complete MEMORY replica for the session range-read path. The
+// session ranged-get path (Client::BatchTransferReadRanges) only ever reads
+// MEMORY replicas: it rejects every non-MEMORY entry, so selecting a NOF_SSD
+// replica here would let a session start and then fail the actual transfer.
+// Among complete MEMORY replicas, a local one is preferred; otherwise the
+// first complete MEMORY replica is returned (nullptr if there is none).
+//
+// Unlike SelectBestReplica this deliberately does NOT consider NOF_SSD: the
+// session path cannot read it. Local MEMORY is still preferred over a remote
+// MEMORY replica regardless of the order the master returned them in.
+inline const Replica::Descriptor *SelectCompleteMemoryReplica(
+    const std::vector<Replica::Descriptor> &replicas,
+    const std::unordered_set<std::string> &local_endpoints) {
+    const Replica::Descriptor *first_memory = nullptr;
+    for (const auto &r : replicas) {
+        if (r.status != ReplicaStatus::COMPLETE || !r.is_memory_replica()) {
+            continue;
+        }
+        if (local_endpoints.count(r.get_memory_descriptor()
+                                      .buffer_descriptor.transport_endpoint_)) {
+            return &r;
+        }
+        if (!first_memory) {
+            first_memory = &r;
+        }
+    }
+    return first_memory;
+}
+
 }  // namespace mooncake
--- a/mooncake-store/include/replica_selection.h
+++ b/mooncake-store/include/replica_selection.h
@@ -117,12 +117,16 @@ inline const Replica::Descriptor *PickBestRemoteMemory(
 
 // Select the best replica from a list: prefer local MEMORY, local NOF_SSD,
 // remote MEMORY, remote NOF_SSD, LOCAL_DISK, DFS, then DISK. Master may return
-// replicas in any order, so we always scan. When scoring is enabled and there
-// are multiple remote MEMORY replicas, the best-scoring one is chosen instead
-// of the first encountered.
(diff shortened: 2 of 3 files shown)


mooncake-store/src/real_client.cpp
```diff
@@ -5479,9 +5479,9 @@ std::vector<int> RealClient::batch_get_session_start(
         }
 
         const auto *replica =
-            SelectCompleteMemoryReplica(query_result.replicas, local_endpoints);
+            SelectBestReplica(query_result.replicas, local_endpoints);
```
fcczzz: Thanks for working on this. I think this change is incomplete because
`SelectBestReplica` may return a `LOCAL_DISK`, `NOF`, `DISK`, or `DFS` replica, while the downstream session range-read path is still memory-only.

In `batch_get_into_multi_buffer_ranges`, `replica_limit` is set to `0` for a non-memory replica, so the call returns `INVALID_PARAMS`. Moreover, `Client::BatchTransferReadRanges` explicitly rejects non-memory replicas.

As a result, this change may only move the failure from `batch_get_session_start` to `batch_get_into_multi_buffer_ranges`.

Could this PR either:

1. add a `LOCAL_DISK` range-read path using the existing `execute_ranged_read` / offload RPC implementation; or
2. continue rejecting unsupported replica types at session start until their range-read paths are implemented?

It would also be helpful to add an end-to-end test with a LOCAL_DISK-only object covering:

`batch_get_session_start -> batch_get_into_multi_buffer_ranges -> batch_get_session_end`.

bitborne: One additional ordering issue: `SelectBestReplica` does not actually enforce its documented cross-type priority of local MEMORY before local NOF.

I reproduced this against the current PR code:

```text
replicas = [local NOF, local MEMORY] -> selects NOF
replicas = [local MEMORY, local NOF] -> selects MEMORY
```

The previous `SelectCompleteMemoryReplica` selects MEMORY in both cases.

Although initial allocation currently appends MEMORY before NOF, that order is not a lifecycle invariant. `ObjectMetadata` preserves insertion order, and paths such as promotion append a new MEMORY replica to existing metadata. Therefore, a complete MEMORY replica may appear after an existing NOF replica.

Since the session range-read path is memory-only, this can regress a session that still has a usable MEMORY replica. Could we make the type priority order-independent and add a regression test covering both permutations?

XFDG:
@fcczzz @bitborne Fixed both issues:

**1. Priority ordering bug (bitborne):** `SelectBestReplica` now scans *all* replicas before returning. Previously the first local replica (MEMORY or NOF) caused an early `return`, so `[local NOF, local MEMORY]` would incorrectly select NOF. The fix collects `local_memory` and `local_nof` in one pass, then returns `local_memory` first — local MEMORY always beats local NOF regardless of list order.

**2. NOF replica_limit=0 bug (fcczzz):** `batch_get_into_multi_buffer_ranges` now reads the NOF buffer size for overflow detection instead of hardcoding 0. Both MEMORY and NOF replicas support range reads via RDMA, so the limit should come from their respective buffer descriptors.

mooncake-store/src/real_client.cpp
XFDG: Added the regression test `LocalMemoryBeatsLocalNoFInBothOrders` in `replica_selection_test.cpp` — it covers both permutations you reproduced (local NOF before local MEMORY, and local MEMORY before local NOF) and asserts both select the local MEMORY replica. Thanks for the clear repro.

XFDG: This PR takes option 2: `batch_get_session_start` now selects a MEMORY or NOF replica via `SelectBestReplica` (both support range reads over RDMA), and `batch_get_into_multi_buffer_ranges` reads the buffer size from whichever descriptor is present instead of hardcoding 0. Unsupported types are still rejected at session start until their range-read paths land. The LOCAL_DISK end-to-end test you suggested is a good follow-up but needs the full client/server harness, so I'll keep it out of this PR to keep the diff focused.

catyans:
Thanks @fcczzz and @bitborne for digging into the two selection issues, and @XFDG for fixing the order-dependent MEMORY/NoF choice. That part looks good now.

I think the session-read concern is still only halfway resolved, though. `batch_get_session_start()` can now cache a NoF replica, and the next step correctly derives its size, but `Client::BatchTransferReadRanges()` still rejects every non-MEMORY replica at `client_service.cpp:4334`. So a NoF-only object gets past the new size check and then fails the actual transfer; LOCAL_DISK/DFS/DISK selections still fail even earlier with a zero `replica_limit`.

Could we either add the NoF range-transfer path end to end, or keep session selection restricted to the replica types that `BatchTransferReadRanges()` actually supports? A NoF-only `start -> ranged read -> end` regression would make the intended fix especially clear. The two guard cases offered in #3742 would also be worth carrying over.

XFDG:
Thanks @catyans — you're right, and I confirmed it end to end. `batch_get_session_start` was selecting via `SelectBestReplica`, which can return a NOF_SSD replica, but the session range-read path is memory-only: `Client::BatchTransferReadRanges` (`client_service.cpp:4334`) rejects every non-MEMORY entry. So a NOF-only object passed the new size check and then failed the actual transfer, and a LOCAL_DISK/DFS/DISK selection failed even earlier with a zero `replica_limit`. The comment claiming "both support range reads via RDMA" was simply wrong.

I took the second option you offered — **keep session selection restricted to the replica types `BatchTransferReadRanges()` actually supports** — rather than building an end-to-end NOF range-transfer path, which is a much larger transport-layer change and out of scope for this bugfix.

Fixed in `2c009ca`:

- **Session path is MEMORY-only again.** Moved `SelectCompleteMemoryReplica` out of the `real_client.cpp` anonymous namespace into `replica_selection.h` (next to `SelectBestReplica`) so it's unit-testable, and `batch_get_session_start` now uses it. It returns only a complete MEMORY replica (local preferred), or `nullptr` — in which case the session is declined with `INVALID_REPLICA` at start, instead of caching a NOF/disk replica that fails later. That also closes the LOCAL_DISK/DFS/DISK case you noted.
- **Single-object get path is unchanged** — it keeps `SelectBestReplica`, which correctly routes NOF/DISK/DFS through `Client::Get`.
- **The #3658 fix is preserved.** `SelectCompleteMemoryReplica` scans all replicas and prefers a local MEMORY replica, so local MEMORY still beats a later remote MEMORY regardless of master return order. It's orthogonal to the memory-vs-NOF ordering fix in `SelectBestReplica`.
- Removed the dead NOF branch in `batch_get_into_multi_buffer_ranges` and corrected its comment.

New regression `SessionPathRejectsNoFOnlyObject` pins exactly the halves you asked for: a NOF-only object is selected by `SelectBestReplica` (single-object path) but declined by `SelectCompleteMemoryReplica` (session path), and a local MEMORY replica still wins when both MEMORY and NOF are present.

One note on the `#3742` guard cases: those tighten `SelectBestReplica`'s local-NOF-vs-remote scan, which now only affects the single-object path (the session path no longer touches NOF). Happy to carry them over here once #3742 lands, or leave them to #3742 — whichever you prefer.

XFDG:
CI note (not a code change): the one red leg — `test-wheel-ubuntu (ubuntu-22.04, 3.12)` → "Run SSD offload and promotion end-to-end tests" → `test_concurrent_stress` (`test_ssd_offload_in_evict.py:431`, `AssertionError: 108 != 0 : 108 concurrent read(s) failed`) — is a pre-existing flake, not a regression from `2c009ca`.

Evidence:
- The `ubuntu-24.04, 3.10` leg of the **same commit** passes; only `ubuntu-22.04, 3.12` fails. That OS/matrix-leg split, independent of branch content, is the flake signature.
- The identical test/line/assertion fails on several unrelated branches that don't touch replica selection or session-start at all (e.g. `feat/hp-tcp-read-slicing`, `codex/tent-unknown-rpc-error`, `codex/tensor-from-multibuffer-pr`), with only the failure *count* varying (108 / 140 / 640 / 748 / 1860) — nondeterministic concurrency timing under eviction pressure.
- `test_concurrent_stress` exercises the single-object `store.get()` path (`SelectBestReplica`), which this PR leaves unchanged; it never calls `batch_get_session_*`. The session-path change in `2c009ca` is out of its code path.

All logic-relevant checks are green: `CTest unit tests`, `Check code format`, and every `build-wheel` (cuda / cu13 / rocm / efa). A re-run of the flaky leg should clear the gate.

catyans (approved):
Thanks for tracing this all the way through. LGTM.

Restricting session start to complete MEMORY replicas now matches what `BatchTransferReadRanges()` can actually execute, so a NoF-only object is rejected up front instead of failing halfway through the session. The new selector still preserves the order-independent local-MEMORY preference that @fcczzz and @bitborne called out, and the focused test captures both sides of that contract nicely.

The remaining Ubuntu 22.04 failure is in the known SSD concurrent-stress path and doesn't exercise this session-selection change, so I don't see it as a blocker for this fix.

fcczzz:
The current MEMORY-only session selection matches BatchTransferReadRanges, and the order-independent local MEMORY preference looks correct.

However, the session selector and start/read/end behavior are unchanged from the merge base, so SSD-only objects still fail session start with INVALID_REPLICA. Please update the title and description, remove “Fixes #3658”, and coordinate the overlapping priority fix with #3742.

The latest CI run also failed.

catyans:
Good catch, @fcczzz. I agree the current title/body overstate what this head does.

My approval was for the safety correction—keeping the session path MEMORY-only so it cannot cache a replica that `BatchTransferReadRanges()` will reject—and for the separate order-independent local MEMORY preference in `SelectBestReplica`. But the session selector itself remains MEMORY-only, so an SSD-only object still returns `INVALID_REPLICA`; this does not deliver the end-to-end SSD session support requested by #3658.

@XFDG, could you retitle/rewrite the PR around the actual replica-priority fix, remove `Fixes #3658`, and coordinate the overlapping `SelectBestReplica` change with #3742? The code can still be useful, but the issue should stay open until the ranged session read supports SSD/NoF end to end (or the API contract is explicitly changed). Thanks for flagging the distinction.

XFDG:
Thanks @fcczzz and @catyans for the precise review.

Done in this update:

1. **Title and description rewritten** to reflect the actual scope: the PR now describes "restrict `batch_get_session_start` to MEMORY replicas + order-independent local preference", and explicitly states that SSD-only objects still return `INVALID_REPLICA` (full SSD session support is out of scope, tracked in #3658 / #3742).

2. **`Fixes #3658` removed** from the PR body.

The code itself is unchanged — @catyans's approval covers the safety correction (MEMORY-only session start) and the order-independent `SelectBestReplica` fix, both of which remain correct as-is.

XFDG:
Closing as superseded: #3903 has merged the MEMORY/DFS ranged-session support, while #3742 covers the remaining order-independent local MEMORY vs NoF selection fix. This PR no longer has a unique functional scope.

XFDG:
@Icedcoco @2sumtech @catyans I narrowed #3662 in commit `45522cec` following the coordination on #3742. All ranged-session/helper overlap with merged #3903 has been removed; the PR now contains only the order-independent `SelectBestReplica` fix and focused regression coverage. I also adapted both guard cases from `2sumtech/Mooncake@67bfff5d`: incomplete local MEMORY falls back to local NOF, and local NOF still outranks remote MEMORY. Thank you for preparing those cases. The title and description now reflect the reduced scope and no longer claim session or #3658 behavior. Please re-review the refreshed two-file change when convenient.

fcczzz (approved):
LGTM

ykwd (approved):

PR #3669, opened 2026-08-25
[CI] Rebalance PR and nightly test coverage

Aionw:
## Description

Rebalance CI coverage so pull requests get comprehensive pre-merge unit and formatting feedback while resource-intensive end-to-end workloads run nightly.

- Run the complete root CTest suite, Mooncake Store Rust library unit tests, and all Mooncake PG CPU tests in PR CI.
- Keep standard wheel builds on Python 3.10 and 3.12, but reduce the Ubuntu consumer cross-product from four legs to two diagonal legs: Ubuntu 22.04 / Python 3.12 and Ubuntu 24.04 / Python 3.10.
- Fold all Mooncake PG CPU tests into the Ubuntu 22.04 / Python 3.12 consumer leg, where the installation test installs the latest available PyTorch (Mooncake PG supports PyTorch >= 2.13.0).
- Stabilize the elastic all-gather/reduce-scatter recovery test exposed by the expanded PG coverage: complete the healthy round on every rank, use the basic all-reduce path to detect and deactivate the departed rank, then validate the reduced-rank buffer layout and replacement recovery.
- Keep build-flag and CUDA 13 PR validation on Python 3.12; these checks validate build configuration rather than duplicating the standard wheel's Python-version coverage.
- Keep Reshard contract tests in both retained wheel-consumer legs through `scripts/test_installation.sh`, and remove the duplicate standalone two-version Reshard job.
- Add Ruff lint and format checks for Python files changed by a PR, alongside the existing spelling and C/C++ formatting checks.
- Run Store tensor, async, copy/move, drain, safetensor, and RPC smoke coverage once on Ubuntu 22.04 / Python 3.12 instead of repeating it in every wheel-consumer leg.
- Run disk-backed SSD offload-on-evict and promotion-on-hit once in PR CI on Ubuntu 22.04 / Python 3.12; historical test-body medians are about 45 and 41 seconds, so they do not warrant nightly-only placement.
- Keep CXL and the broader Python integration suite nightly, and run the upstream T-One/SGLang integration there with the CUDA 13 x86 / Python 3.12 wheel from that same run; labeled PR E2E remains available through `e2e-ci.yml`.
- Require T-One credentials for artifact-backed nightly/pre-release gates, and make the nightly RPC check fail unless the server becomes ready, the client returns an expected status, and a successful `bandwidth:` result is produced.
- Keep spelling, formatting, and path checks as prerequisites for the existing wheel, CUDA 13, EFA, ROCm, and TENT jobs.
- Reduce a full source-changing PR fan-out from 30 to 23 worker jobs plus the final gate, while retaining the intended version, platform, unit, and smoke coverage.
- Align nightly standard wheel coverage with pre-release/release coverage: 3 variants x 2 architectures x 4 Python versions = 24 wheel jobs.
- Prevent stable EFA release workflows from also building on pre-release tags.
- Fix PR event conditions so opened, synchronized, and reopened PRs run normally while labeled events still require `run-ci`.

Related open work:

- #3735 tracks the pre-existing SSD test assertion gap identified during review; this PR changes where that test runs but intentionally does not modify its implementation.
- #1532 is an older WIP that fetches existing main-branch artifacts for a nightly release; this PR instead changes the current PR/nightly test split and build matrix.
- #3654 touches some of the same workflow files for toolchain unification, but does not address test placement or nightly/release coverage alignment.

## Module

- [x] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [x] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
python <workflow placement, dependency, matrix, lint, Rust-unit, integration-artifact, credential-gate, and RPC-result assertions>
bash -n <extracted Ruff, Store API/RPC smoke, nightly RPC, and T-One integration steps>
MC_FORCE_TCP=true taskset -c 0,1 python mooncake-pg/tests/test_pg_elastic.py TestMooncakePGElasticCPU.test_allgather_reduce_scatter_recovery -v
# Repeated with the Python 3.12 / PyTorch 2.13 wheel environment from the failed CI run.
git diff --check
pre-commit run --files .github/workflows/ci.yml .github/workflows/nightly.yml .github/workflows/integration-test.yml .github/workflows/release-efa-cuda13.yaml .github/workflows/release-efa-non-cuda.yaml .github/workflows/release-efa.yaml
pre-commit run --files mooncake-pg/tests/test_pg_elastic.py
```

**Test results:**
- [ ] Unit tests pass
- [x] Targeted PG CPU recovery test passes repeatedly with affinity restricted to two CPUs, reproducing the constrained-runner condition that previously timed out.
- [x] Manual testing done: parsed the workflow YAML; verified PR job placement and gates, the 23-worker fan-out, both T-One artifact sources and credential requirements, all PG CPU coverage folded into the latest-PyTorch consumer, Reshard coverage in both retained consumers, strict PR/nightly RPC result validation, single-leg Store smoke plus SSD offload/promotion coverage, and the 24-leg nightly/pre-release/release standard wheel matrix; extracted shell steps pass Bash syntax checks.

The complete CTest, Rust unit, full PG CPU suite, wheel-consumer, Store smoke, T-One/SGLang, and nightly workloads require their GitHub runner environments and are left to PR CI/nightly execution.

## Checklist

- [x] I have formatted my code using `./scripts/code_format.sh` (no C/C++ files changed; the corresponding pre-commit hook skipped)
- [x] I have updated the documentation (not applicable; CI-only behavior)
- [x] For changes >500 LOC: I have filed an RFC issue (not applicable; this is not a major architectural change)

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex was used to inspect the existing workflows, implement the CI reorganization, diagnose and stabilize the constrained-runner PG recovery test, and run local static validation. The human submitter reviewed the resulting diff and requested dependency-preservation, smoke-coverage, Python-version, lint, Rust-unit, nightly-integration, PR-runtime, and review-feedback adjustments before submission.


.github/workflows/ci.yml
```diff
@@ -15,6 +15,9 @@ on:
 permissions:
   contents: read
 
+env:
+  MOONCAKE_TEST_TORCH_VERSION: '2.13.0'
+
```
UNIDY2002: FYI, Mooncake PG does not require the CI to pin torch to a specific version. Any version >= 2.13.0 should do, which means that, from PG's perspective, simply installing the latest torch would be the cleanest choice.

UNIDY2002 (approved):
Well done. Just a piece of information from the PG's side.

.github/workflows/ci.yml
Aionw: OK, I'll change to that way.

Aionw: Updated in caf6da17: removed the global PyTorch pin and version assertions. PR wheel consumers and the PG CPU job now install the latest available torch; the Python matrix remains unchanged.

ykwd:
> Move the Store Python API and RPC communicator end-to-end workloads from the default PR workflow to nightly

What’s the motivation for moving these tests to nightly? Do we not need to run them before merging a PR?

Aionw:
Good point. The intent was to remove the resource-intensive suite from four duplicated Ubuntu/Python consumer legs, but moving all API coverage to nightly would weaken pre-merge protection. Updated in 6818403a: PR CI now runs the Store tensor, async, copy/move, drain, safetensor, and RPC smoke tests once on Ubuntu 22.04 / Python 3.12. Only the full suite with SSD eviction/offload, promotion-on-hit, and CXL remains nightly.

Icedcoco:
Thanks for carrying the CI rebalancing work through. I took another look at the latest revision and had a few questions about the intended behavior:

1. In the new PR-side SSD step, `test_ssd_offload_in_evict.py` records empty or mismatched reads in `get_stats`, but `test_put_get_in_evict_operations` does not appear to assert that the failure count is zero. Was this intentionally meant to be a throughput-style smoke test? Since this is now part of the PR coverage for SSD offload, would it make sense to assert that every successfully written key is read back correctly, so a data-loss regression cannot leave the job green?

2. The nightly RPC client still ends with `|| true`. Is this nightly run intentionally best-effort? The PR-side version now accepts only the expected timeout exit code and checks for a `bandwidth:` result. Could we apply the same validation here, so a client startup or transfer failure remains visible to `nightly-gate`?

3. `tone-sglang-integration` is now included in `nightly-gate`, while `integration-test.yml` skips both the trigger and polling steps when `TONE_USER_NAME` is unavailable. Is a successful no-op nightly job without the T-One credentials intentional? If this nightly gate is meant to guarantee that the integration actually runs, could we require `TONE_USER_NAME` and `TONE_USER_TOKEN` when `source_run_id` or `artifact_name` is provided, similar to the TestPyPI path?

I may be missing an intentional distinction between the PR and nightly gates, so I wanted to check the expected behavior before treating these as blockers.

Aionw:
@Icedcoco Thanks for the careful review. I checked each point against `main` and addressed them as follows:

1. The SSD assertion gap is pre-existing; this PR does not modify `test_ssd_offload_in_evict.py`, only where it runs. I filed #3735 to track making empty or mismatched reads fail the unittest (and linked the related real-world empty-read report #3465). I am keeping that test-code change outside this CI-placement PR.

2. Fixed in 6ae1a076. The nightly RPC check no longer uses a blanket `|| true`: it now waits for server readiness, captures client/server logs, rejects unexpected client exit codes, and requires a `bandwidth:` result, matching the PR-side validation.

3. Fixed in 6ae1a076. The reusable T-One workflow now requires both `TONE_USER_NAME` and `TONE_USER_TOKEN` whenever `testpypi_version`, `source_run_id`, or `artifact_name` is supplied. Therefore the artifact-backed nightly job can no longer become a successful no-op when credentials are unavailable.

The updated YAML, extracted nightly RPC shell, `git diff --check`, and pre-commit checks all pass.

Icedcoco (approved):
LGTM

ykwd (approved):

Issue #3556, opened 2026-08-20
[Track]: Mooncake Build, CI, and Release Roadmap


Aionw:
This RFC tracks Mooncake's build, CI, and release evolution. Contributions and feedback are welcome.

## Focus

The work is divided into four directions:

1. **C++ Build** — one explicit and modular native build graph.
2. **Python Build** — one Python source tree and one standard package backend.
3. **CI** — clear PR, nightly, and release CI responsibilities, implemented with small reusable workflows.
4. **Release** — one reviewable process from release branch to published artifacts.

The overall dependency is:

```text
C++ Build + Python Build  -->  Reusable CI  -->  Release
```

Build interfaces come first. CI should consume them instead of implementing another build path, and Release should publish the artifacts already validated by CI.

## 1. C++ Build

**Goal:** use CMake as the single native build graph for standalone components, language bindings, tests, and Python wheels.

### Direction

- **One build graph.** CMake owns feature flags, targets, tests, and install rules for standalone components, language bindings, and Python wheels.
- **Clear component boundaries.** Transfer Engine remains independently buildable; Store exposes one native SDK boundary; and PG/EP native cores do not depend on the PyTorch C++ ABI.
- **One dependency policy.** Prefer maintained, high-performance libraries with few transitive dependencies. Acquire, pin, and link them target-by-target through CMake when practical; keep only toolchains, drivers, and vendor runtimes external with explicit supported versions.

### Tracking

- [ ] #3800 
- [ ] #3181
- [ ] #3553
- [ ] #3550
- [x] #3537
- [x] #3276
- [x] #3206
- [x] #2883

## 2. Python Build

**Goal:** make one tracked Python package and one PEP 517 backend the source of every editable install and wheel variant.

### Direction

- **One source tree.** Code shipped under `import mooncake` has one authoritative location; builds neither copy tracked sources nor modify package metadata.
- **One build interface.** PEP 517 is the public entry point for editable installs and every wheel profile, and its backend consumes CMake install outputs.
- **One install contract.** Package dependencies and public imports are explicit, and every wheel is validated from a clean environment outside the source checkout.

### Tracking

- [ ] #3534
- [ ] #3425

## 3. CI

**Goal:** provide three distinct gates over one shared build and artifact contract: PR CI decides whether a change may merge, Nightly CI reports the health of `main`, and Release CI qualifies one exact candidate for publication.

### Current baseline

[...]

PR #3672, opened 2026-08-25
[CI/Build] Gate pre-releases through TestPyPI

Aionw:
## Description

Turn the pre-release workflow into a TestPyPI-backed release gate for the core wheel matrix.

- Normalize pre-release tags to PEP 440 versions and stamp the existing 24-wheel CUDA, CUDA 13, and non-CUDA matrix with that exact version.
- Validate wheel filenames and metadata, reject TestPyPI version collisions, upload the exact collected artifacts, and wait for the complete indexed matrix.
- Download each exact package version from TestPyPI, install the downloaded wheel with dependencies resolved from production PyPI, and run package smoke checks.
- Feed the indexed CUDA 13 wheel into the existing T-one integration workflow and report publication, index, consumer, and integration results in the workflow summary.
- Keep the reusable workflow declarative by moving GitHub artifact lookup and T-one job creation/polling into a focused, unit-tested Python orchestrator.

Related to #3556.

## Module

- [x] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

Focused workflow and TestPyPI gate regression tests plus pre-commit checks for every touched file.

**Test commands:**
```bash
python -m pytest -q \
  mooncake-wheel/tests/test_release_wheel_tags.py \
  mooncake-wheel/tests/test_testpypi_wheel_gate.py \
  mooncake-wheel/tests/test_tone_integration.py

pre-commit run --files \
  .github/workflows/integration-test.yml \
  .github/workflows/pre-release.yaml \
  mooncake-wheel/tests/test_release_wheel_tags.py \
  mooncake-wheel/tests/test_testpypi_wheel_gate.py \
  scripts/ci/testpypi_wheel_gate.py
```

**Test results:**
- [x] Unit tests pass (22 passed)
- [ ] Integration tests pass (requires GitHub Actions secrets, TestPyPI publication, T-one, architecture-specific runners, and hardware)
- [ ] Manual testing done

## Checklist

- [ ] I have formatted my code using `./scripts/code_format.sh` (not applicable; no C/C++ changes)
- [ ] I have updated the documentation (not applicable; CI-only change)
- [x] For changes >500 LOC: I have filed an RFC issue (#3556)

## AI Assistance Disclosure

- [ ] No AI tools were used
- [x] AI tools were used

Codex helped implement, review, and validate the workflow and focused regression tests. The human submitter reviewed the complete change and directed the final design and fixes.


.github/workflows/pre-release.yaml
```diff
@@ -91,11 +174,173 @@ jobs:
...
+
+          package, expected = sys.argv[1:]
+          installed = importlib.metadata.version(package)
+          if installed != expected:
+              raise SystemExit(
+                  f"installed {package} version {installed}, expected {expected}"
+              )
+
+          import mooncake
+          import mooncake.http_metadata_server
```
ShangmingCai, line 257: **This smoke check can't detect a broken extension on 2 of the 3 packages.**

`scripts/build_wheel.sh` repairs with `--exclude libibverbs.so* --exclude libnuma.so* --exclude libcurl.so*` (and friends), so `engine.so`/`store.so` keep their `DT_NEEDED` on the system libraries — which is exactly why the non-cuda leg above apt-installs `libibverbs1 rdma-core librdmacm1 libnuma1 liburing2`. The CUDA legs install none of them.

And `mooncake/buffer_pool.py`, which `mooncake/__init__.py` imports, is:

```python
try:
    from mooncake.store import BufferPool
except (ImportError, AttributeError):
    BufferPool = None
```

A loader failure on a missing `libibverbs.so.1` raises `ImportError`, so it's swallowed and `import mooncake` succeeds either way. For `mooncake-transfer-engine` and `-cuda13` this step therefore proves METADATA, aiohttp and the pure-Python surface are intact, and nothing at all about the binary. Only the non-cuda leg (`import mooncake.engine, mooncake.store` + `mooncake_master --version`) is real coverage.

Suggest installing the runtime deps on all three legs and importing `mooncake.engine`/`mooncake.store` everywhere. `_build-wheel.yaml`'s own "Smoke test repaired wheel" step already shows the trick for a driver-less runner — symlink the CUDA stub as `libcuda.so.1`.

.github/workflows/pre-release.yaml
```diff
@@ -79,10 +131,41 @@ jobs:
...
+            --version "$PACKAGE_VERSION"
+
       - name: Validate wheels with twine
+        run: twine check mooncake-wheel/dist-release/*.whl
+
+      - name: Reject an existing TestPyPI version
         run: |
-          pip install twine
-          twine check mooncake-wheel/dist-release/*.whl
+          python scripts/ci/testpypi_wheel_gate.py ensure-version-absent \
```
ShangmingCai: **A partial publish permanently burns the tag.**

`ensure-version-absent` combined with deliberately no `--skip-existing` on the upload (and `assert "--skip-existing" not in commands` in the test locking that in) means that if `twine upload` dies after, say, the 7th of 24 wheels, this check fails on every subsequent re-run. The only ways forward are cutting a new tag or manually deleting the release on TestPyPI.

I understand the intent — you don't want to mix two builds under one version — but the current shape has no recovery path for a transient network failure mid-upload. Two options that keep the invariant:

- make the check tolerant when everything already indexed at this version is a subset of the exact set about to be uploaded (same filenames), and add `--skip-existing`; or
- keep it strict and document the "transient publish failure ⇒ cut `-rc2`" recovery in the workflow header, so the next person hitting it isn't stuck.

ShangmingCai (commented):
Reviewed the workflow changes against the surrounding CI, the T-one harness in `scripts/tone_tests/`, and `build_wheel.sh`; applied the branch locally and ran the new tests (22/22 pass).

Good direction, and the gate script is real testable code rather than more inline bash. I ran it against live TestPyPI: `ensure_version_absent` correctly flags an existing nightly version, and `validate_wheel_set` handles auditwheel's compressed platform tags. Moving the T-one credentials out of interpolated command lines into `env:` is a genuine improvement, and passing `version-override` finally stamps pre-release wheels with the tag instead of pyproject's static version.

Comments inline. The ones I'd treat as blocking are the vacuous CUDA-leg smoke check, the lack of a recovery path after a partial upload, and the fact that none of these tests run in CI. The rest are smaller.

For the record, a few things I checked that turned out fine: `pip install twine` does pull `packaging` transitively; `main` has no required status checks, so renaming the job to `test-tone-integration` breaks nothing; the `nightly` environment has no protection rules, so it won't gate on approval; and T-one's `get_whl` globs the first `*.whl` out of the artifact zip, so the single-wheel staging artifact is compatible.

.github/workflows/pre-release.yaml
```diff
@@ -91,11 +174,173 @@ jobs:
...
+            --index-url https://test.pypi.org/simple \
+            --version "$PACKAGE_VERSION" \
+            --attempts 8 \
+            --initial-delay 5 \
+            --max-delay 30
+
+  consume-testpypi:
+    name: Consume ${{ matrix.package }} from TestPyPI
+    needs: [version-stamp, verify-testpypi-index]
+    runs-on: ubuntu-22.04
```
ShangmingCai: **Half the wheel matrix is never actually installed.**

All three `consume-testpypi` legs run on `ubuntu-22.04` (x86_64) with Python 3.12, so `pip download` only ever resolves the cp312 x86_64 wheel. The 12 aarch64 wheels are validated by filename and METADATA only, never downloaded or imported — which is a bit narrower than "Download each exact package version from TestPyPI, install the downloaded wheel with dependencies resolved from production PyPI" in the description.

Worth closing because the aarch64 glibc floor is precisely what regressed in #2858 (published floor moved `manylinux_2_35` → `manylinux_2_39` with no code change), and that's the regression `test_release_wheel_tags.py` exists to guard. `ubuntu-22.04-arm` runners are already in use in `_build-wheel.yaml`, so adding `architecture` to this matrix should be cheap.

.github/workflows/pre-release.yaml
```diff
@@ -15,7 +12,53 @@ on:
       - 'v*-pre*'
 
 jobs:
+  version-stamp:
```
ShangmingCai, line 15: **`pre-release.yaml` loses its fork-safety here.**

There's no `if: github.repository == 'kvcache-ai/Mooncake'` anywhere in this workflow. Before this PR a fork pushing a `v*-rc*` tag built the matrix and validated it successfully; now it builds all 24 wheels and then fails in `publish-testpypi` on the missing `TESTPYPI_API_TOKEN`.

`nightly.yml` guards its dependency-root job exactly this way ("Gate the dependency root so all downstream jobs are skipped in forks") — the same guard on `version-stamp` would restore it.

scripts/ci/tone_integration.py
```diff
@@ -0,0 +1,332 @@
...
+        client = JsonClient()
+        artifact_id = os.environ.get("TESTPYPI_ARTIFACT_ID", "")
+        if artifact_id:
+            print(f"Using TestPyPI wheel artifact id {artifact_id}")
+        else:
+            artifact_id = find_ci_artifact(repository, source_sha, client.github_get)
+
+        branch = ""
+        if os.environ.get("TESTPYPI_VERSION", ""):
+            branch = _required_env("GITHUB_REF_NAME")
```
ShangmingCai: **`BRANCH` is plumbed all the way through but nothing reads it.**

The in-repo T-one harness consumes exactly two variables from `env_info`: `scripts/tone_tests/scripts/run_test.sh` declares `ARTIFACT_ID` and `GIT_REPO`, and `common.sh::get_whl` uses only those two to `gh api .../actions/artifacts/$ARTIFACT_ID/zip`, unzip, and glob the first `*.whl`. Neither `BRANCH` nor `PR_ID` appears anywhere in that tree.

So the new `branch=` parameter on `build_env_info`, this `_required_env` call, and `test_tone_environment_includes_only_the_active_source_context` all exist to ship a value that's dropped on the floor — and `_required_env` turns it into a hard failure mode for something nobody consumes.

If the server-side `mooncake-ci-test` template does read it, worth saying so in a comment — and note that on a tag push `GITHUB_REF_NAME` is a tag, not a branch, so the name is misleading. Otherwise I'd drop `branch` entirely.

scripts/ci/tone_integration.py
```diff
@@ -0,0 +1,332 @@
...
+        if not isinstance(payload, dict):
+            raise ToneIntegrationError(
+                f"expected a JSON object from {request.full_url}"
+            )
+        return payload
+
+    def github_get(self, url: str) -> JsonObject:
+        headers = {
+            "Accept": "application/vnd.github+json",
+            "X-GitHub-Api-Version": GITHUB_API_VERSION,
```
ShangmingCai: Minor, and carried over from the bash this replaces rather than introduced here: these GitHub API calls are unauthenticated, and `find_ci_artifact` polls up to 120 times against a 60-requests/hour unauthenticated limit that is shared per source IP across hosted runners.

Not on the pre-release path (`TESTPYPI_ARTIFACT_ID` short-circuits the lookup), so this only affects the existing `ci.yml`/`e2e-ci.yml` callers. But now that it's a real function instead of inline curl it's a two-line fix — add `Authorization: Bearer ${{ github.token }}` here and `permissions: actions: read` on the job.

mooncake-wheel/tests/test_release_wheel_tags.py
```diff
@@ -1,53 +1,24 @@
...
-runner image. That is not hypothetical: the published aarch64 floor moved from
-``manylinux_2_35`` (0.3.9) to ``manylinux_2_39`` (0.3.10) with no code change,
-which stops ARM Ubuntu 22.04 from resolving any wheel and contradicts the
-"OS: Ubuntu 22.04 LTS+" contract in docs/source/getting_started/build.md.
-
-Running CI and release builds through the same manylinux workflow makes the
-detected glibc a constant of the image instead of a property of the runner. See
-#2858.
-"""
+"""Regression guards for shared wheel builds and the pre-release gate."""
```
ShangmingCai: **None of the tests in this file — or the two new test modules — are executed by CI.**

I checked every entry point:
- `ci.yml` runs `scripts/test_installation.sh`, which copies `mooncake-wheel/tests` into `test_env/` and then runs three named scripts (`test_fast_copy.py`, `test_import_structure.py`, `test_mooncake_config.py`) plus `pytest reshard_tests`.
- `ci.yml` also runs `scripts/run_tests.sh`, which `cd`s into `mooncake-wheel/tests` and invokes named scripts directly.
- There's no `[tool.pytest.ini_options]`/`testpaths` in the root `pyproject.toml`, no `pytest.ini`, no `tox.ini`.

So nothing ever pytest-collects `test_release_wheel_tags.py`, `test_testpypi_wheel_gate.py`, or `test_tone_integration.py`.

That gap is already biting: I ran the **current `main`** version of this file and it fails 7/8 — `_build-wheel.yaml` moved `container` out of the callers' `with:` blocks, so `test_wheel_build_pins_glibc_floor_to_a_container` has been red for a while and nobody noticed. This PR quietly repairs it, which is good and worth a line in the description, but the fix that makes it stay fixed is a pytest step over `mooncake-wheel/tests/test_*.py` run from the repo root (so `_find_workflows_dir()` resolves instead of skipping).

mooncake-wheel/tests/test_release_wheel_tags.py
```diff
@@ -61,57 +32,210 @@ def _find_workflows_dir() -> Path | None:
...
+    version_commands = _commands(version_job)
+    build = jobs["build"]
+
+    assert set(trigger["push"]["tags"]) == {
+        "v*-rc*",
+        "v*-alpha*",
+        "v*-beta*",
+        "v*-pre*",
+    }
+    assert "Version(tag[1:])" in version_commands
```
ShangmingCai: These assertions pin the workflow's literal source text rather than its behaviour — `assert "Version(tag[1:])" in version_commands` here, `assert '"${wheels[0]}"' in commands` at L170, `assert 'mooncake_master" --version' in commands`. They break on any harmless refactor of the YAML and can't catch a real defect.

The structural assertions in the same tests are worth keeping (job graph, `needs`, `environment`, `--skip-existing` absent, relative step ordering) — those encode invariants. It's the source-string matching I'd drop.

`test_testpypi_wheel_gate.py` and `test_tone_integration.py` are the right model, for what it's worth: injected fetchers/sleepers, real retry and collision behaviour exercised. Nice work on those.

ShangmingCai (approved):

Issue #3534, opened 2026-08-19
[RFC]: Unify Mooncake's Python package, build, and release architecture


Aionw:
## Summary

Establish `python/mooncake` as the single source of truth for all tracked Python code distributed as the `mooncake` import package.

Use `scikit-build-core` as the single PEP 517 build backend. It will combine tracked Python sources with native artifacts supplied through explicit CMake install rules, without copying generated files into the source tree.

Use `cibuildwheel` as the standard release-wheel execution layer for supported manylinux variants. Hardware-specific builders may retain specialized runners and toolchains, but must consume the same PEP 517 backend and produce the same validated artifact contract.

This RFC also defines:

- subsystem-oriented Python module boundaries;
- private native extension modules behind stable public Python facades;
- explicit core, optional, development, and hardware dependency groups;
- isolated installed-wheel validation;
- a checked invariant between Git tags, package metadata, and wheel metadata;
- centralized publication to PyPI and GitHub Releases;
- independently retryable hardware-specific release variants.

Distribution names such as `mooncake-transfer-engine` and its hardware variants remain unchanged. The installed import package remains `mooncake`.

Background and a detailed walkthrough are available in [Mooncake Python Packaging Refactor](https://aionw.github.io/posts/mooncake-python-package-refactoring/).

## Relationship to existing RFCs

This proposal overlaps with two existing RFCs:

- #3425, **Refactor the `mooncake-wheel` package layout**, identifies the same fragmented Python source layout and build-time copying problems. Its target architecture keeps `mooncake-wheel/mooncake` as the source root and retains public native module names. This proposal goes further by introducing `python/mooncake` as a source-only package root, private native extensions, `scikit-build-core`, explicit dependency groups, and installed-wheel validation. If accepted, this RFC should supersede the narrower target architecture in #3425.

- #3130, **Release Branch Workflow for Stable Releases**, overlaps with the release-validation part of this proposal. It focuses on release branches and TestPyPI validation, while this RFC defines the package build contract, artifact identity, wheel validation, and centralized publication. The two proposals are complementary: #3130 may define how a release commit is selected and stabilized, while this RFC defines how that commit becomes validated and published Python distributions.

### Python has no single source root

Tracked Python code currently lives under several component directories, including:

- `mooncake-wheel/mooncake`;
- `mooncake-integration`;
- `mooncake-reshard/python/mooncake`;
- EP and PG component directories.

[...]

PR #3673, opened 2026-08-25
[PG] Use JIT for PG torch adapter

UNIDY2002:
## Description

Part of #3534, #3556

We no longer need torch at build time. Also, we will not restrict ourselves to a specific range of torch.

**Potential conflicts:**
- #3623
- #3669

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
```

CI should pass.

**Test results:**
- [x] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex helped me.


Aionw:
Runtime JIT is a reasonable direction for this PR. My concerns are limited to behavior introduced by the current implementation:

- The shipped Docker runtime image does not include the toolchain required by the new import-time compilation path, so `import mooncake.pg` may fail in an officially provided environment.
- A transient compilation failure creates a `.failed` marker that prevents subsequent attempts for the same cache key until the user manually clears it.
- Headless compilation depends on `TORCH_CUDA_ARCH_LIST` because the extension is always loaded with `with_cuda=True`; CI sets this variable, but the user-facing requirement is not documented.
- The MUSA path is selected only through `MOONCAKE_EP_USE_MUSA`, which does not appear to be set automatically by the existing runtime or packaging flow.

Could we address these compatibility and recovery issues, or clearly document the required environment and recovery procedure, as part of this PR?

As a non-blocking follow-up, it may also be useful to consider operational features found in more mature JIT implementations, such as an explicit prebuild/prewarm command or an environment compatibility report. I do not think those need to be implemented in this PR.

UNIDY2002:
@Aionw Thanks for the detailed review. We addressed these compatibility and recovery issues:

- Docker builder images now prewarm the PG adapter and copy the JIT cache into the runtime image. On cache hit, `import mooncake.pg` loads the cached extension directly, so the runtime image does not need Ninja, nvcc, or mcc.
- Failed builds no longer create a permanent `.failed` marker; subsequent imports retry normally.
- Headless CUDA builds require `TORCH_CUDA_ARCH_LIST`, with an actionable error when it is missing. The Docker builder sets it explicitly.
- MUSA is now auto-detected from the installed Torch runtime, while `MOONCAKE_EP_USE_MUSA` remains available as an explicit override.
- We also added `python -m mooncake.pg --prebuild` and `--report`.

The updated path was validated with Torch 2.13 + CUDA 13: cold JIT, warm import, cache-hit import without Ninja/nvcc, and all 14 CUDA PG collective tests passed.

CMakeLists.txt
```diff
@@ -169,63 +169,46 @@ if(WITH_EP)
...
+    file(REMOVE_RECURSE "${EP_PG_STAGING_DIR}")
+    file(MAKE_DIRECTORY "${EP_PG_STAGING_DIR}")
 
-    # Convert semicolon-separated lists to pipe-separated strings so they
-    # survive CMake's COMMAND list-splitting (semicolons are CMake list
-    # separators).
-    string(REPLACE ";" "|" _ep_torch_versions_pipe "${EP_TORCH_VERSIONS}")
-    string(REPLACE ";" "|" _torch_cuda_arch_list_pipe "${TORCH_CUDA_ARCH_LIST}")
+    # Stage the runtime JIT source bundle for the legacy wheel script as well
+    # as the scikit-build install path.  The authoritative files stay under
```
caozhanhao: Does the 'scikit-build install path' actually consume these staged files?

caozhanhao (commented):
Overall LGTM.
One non-blocking observation: the nvcc/mcc requirement appears theoretically removable, since the JIT sources under `mooncake-pg/torch` are host-side C++ adapter code and do not contain device kernels. However, the current cpp extension path (`with_cuda=True`) also handles discovery and injection of the required device include paths, library paths, and link flags. Moving to a pure C++ JIT path would require reproducing that configuration manually and could make compatibility across different environments more brittle. Given that trade-off, keeping the current nvcc/mcc requirement in this PR seems reasonable.

docker/musa.Dockerfile
```diff
@@ -57,6 +50,11 @@ RUN mkdir -p build/mooncake-transfer-engine/nvlink-allocator && \
 # Build the Python wheel from local sources
 RUN OUTPUT_DIR=dist ./scripts/build_wheel.sh
 
+# Warm the runtime JIT cache in the builder image. The cache is copied into the
+# runtime image below, so importing mooncake.pg there does not require mcc.
+RUN python${PYTHON_VERSION} -m pip install --no-cache-dir dist/*.whl && \
```
Aionw: **[P1] Install the wheel from the directory where it is actually built**

`build_wheel.sh` changes into `mooncake-wheel`, so the relative `OUTPUT_DIR=dist` produces `/workspace/mooncake-wheel/dist/*.whl`. This command instead reads `/workspace/dist/*.whl`; the glob does not match and the image build fails before JIT prewarming starts. `docker/mooncake.Dockerfile` has the same issue. I reproduced this in the same CUDA base image: `mooncake-wheel/dist/*.whl` matched, while `dist/*.whl` did not. Please change both call sites to `mooncake-wheel/dist/*.whl`.

docker/mooncake.Dockerfile
```diff
@@ -77,6 +74,9 @@ RUN mkdir -p build && \
     bash build.sh ../../build/mooncake-transfer-engine/nvlink-allocator/ && \
     cd /workspace && \
     OUTPUT_DIR=dist ./scripts/build_wheel.sh && \
+    python${PYTHON_VERSION} -m pip install --no-cache-dir dist/*.whl && \
+    TORCH_CUDA_ARCH_LIST="${TORCH_CUDA_ARCH_LIST}" \
+        python${PYTHON_VERSION} -m mooncake.pg --prebuild && \
```
Aionw: **[P1] Install and pin PyTorch before prebuilding in the CUDA builder**

This builder uses an `nvidia/cuda` image, which does not include `torch`. The new CMake path also removes the old `EP_TORCH_VERSIONS` logic that installed PyTorch, and the Mooncake wheel does not declare `torch` as a runtime dependency. Therefore, even after fixing the wheel path above, `python -m mooncake.pg --prebuild` fails during module import with `ModuleNotFoundError: torch`. I reproduced this with the actual `nvidia/cuda:12.8.1-devel-ubuntu22.04` image, Python 3.10, and the currently declared dependencies. Please explicitly install and pin a compatible PyTorch version in the builder, and ensure the runtime uses the same version so that the copied JIT cache has a stable ABI target.

mooncake-wheel/mooncake/pg.py
```diff
@@ -1,16 +1,394 @@
...
+                _load_torchada()
+                if _mcc_path() is None:
+                    raise ImportError(
+                        "Mooncake PG MUSA JIT requires the MUSA compiler (mcc). "
+                        "Set MUSA_HOME or add mcc to PATH."
+                    )
+            else:
+                from torch.utils.cpp_extension import CUDA_HOME
+
+                if not CUDA_HOME or not (Path(CUDA_HOME) / "bin" / "nvcc").is_file():
```
Aionw, line 320: **[P1] Preserve a working path for the still-supported MACA build**

CMake still supports `WITH_EP=ON` together with `USE_MACA=ON`, and the deleted `setup.py` detected that platform through `MOONCAKE_EP_USE_MACA` or `torch.version.maca`. The new code distinguishes only MUSA from CUDA, so every non-MUSA environment reaches this branch and requires `CUDA_HOME/bin/nvcc`. A normal MACA toolchain provides `mxcc`, which makes the generated wheel fail on its first `import mooncake.pg`. With a mocked `torch.version.maca`, `--report` likewise incorrectly reports `backend: CUDA` and `nvcc: missing`. Please add a MACA JIT path or retain the previous AOT path for MACA builds.

mooncake-wheel/mooncake/pg.py
```diff
@@ -1,16 +1,394 @@
...
+
+def _cache_key(source_dir: Path, core_path: Path, backend: str) -> str:
+    digest = hashlib.sha256()
+    digest.update(torch.__version__.encode())
+    digest.update(str(torch.version.cuda).encode())
+    digest.update(str(getattr(torch.version, "musa", None)).encode())
+    digest.update(str(torch._C._GLIBCXX_USE_CXX11_ABI).encode())
+    digest.update(str(torch.version.git_version).encode())
+    digest.update(backend.encode())
+    digest.update(core_path.read_bytes())
```
Aionw, line 113: **[P2] Include the installation location and Python ABI in the JIT cache identity**

The cache is global to the user, but this key hashes only the contents of the core library, while the linker flags below embed `core_path.parent` as an absolute RPATH. If the same user installs the same wheel in two virtual environments, the second environment reuses the first environment's cached extension. Once the first environment is removed, imports in the second environment keep failing even when its toolchain is complete. A pybind extension also must not be reused across Python ABIs. Please include at least `core_path.resolve()` and `sys.implementation.cache_tag` (or an equivalent SOABI) in the key.

Aionw (commented):
I found several issues that need to be addressed; see the inline comments. The first two Docker issues were reproduced against the corresponding CUDA base image.

CMakeLists.txt
UNIDY2002: Good catch. Fixed in feacda9591a77b77774c2ce4b52517bd1d77657f

mooncake-wheel/mooncake/pg.py
UNIDY2002: Fixed in feacda9591a77b77774c2ce4b52517bd1d77657f

mooncake-wheel/mooncake/pg.py
UNIDY2002: Fixed in feacda9591a77b77774c2ce4b52517bd1d77657f

docker/mooncake.Dockerfile
UNIDY2002: After reconsideration, I decided not to warmup Mooncake PG in the dockerfiles, due to lacking the `torch` dependency.

mooncake-wheel/mooncake/pg.py
```diff
@@ -1,16 +1,428 @@
...
+            elif is_maca:
+                if _mxcc_path() is None:
+                    raise ImportError(
+                        "Mooncake PG MACA JIT requires the MACA compiler (mxcc). "
+                        "Set MACA_HOME or add mxcc to PATH."
+                    )
+            else:
+                from torch.utils.cpp_extension import CUDA_HOME
+
+                if not CUDA_HOME or not (Path(CUDA_HOME) / "bin" / "nvcc").is_file():
```
Aionw: **[P1] Keep the CUDA Docker runtime capable of building the adapter**

Removing the prewarmed cache avoids the missing-`torch` failure in the builder, but it leaves the published CUDA image with no viable path to import `mooncake.pg`. The runtime stage still uses `nvidia/cuda:*runtime*` and installs only the Python interpreter and runtime libraries; it has no `nvcc`, Ninja, C++ compiler, or Python headers, while this cold-cache path requires all of them. I verified the current `nvidia/cuda:12.8.1-runtime-ubuntu22.04` base directly: `nvcc`, `ninja`, `c++`/`g++`, and `Python.h` are all missing. Even if a user installs PyTorch later, the first import therefore fails here. Please either prebuild/copy the adapter using a pinned PyTorch ABI, or provision the complete JIT toolchain in the runtime image (for example by using an appropriate devel base).

docker/mooncake.Dockerfile
Aionw: Removing the prewarm step avoids the missing-`torch` failure in the builder, but it creates a runtime-side blocker instead. The final stage is still based on `nvidia/cuda:${CUDA_VERSION}-runtime-ubuntu${UBUNTU_VERSION}` at `docker/mooncake.Dockerfile:83`, and that stage installs only the Python interpreter and runtime libraries. It has no `nvcc`, Ninja, C++ compiler, or Python headers, so after a user installs PyTorch, the first `import mooncake.pg` still cannot take the cold-cache JIT path. I verified the current `nvidia/cuda:12.8.1-runtime-ubuntu22.04` base directly and all of those tools/headers are absent. Please either restore a prebuilt adapter/cache with a pinned PyTorch ABI or provision the complete JIT toolchain in the final runtime stage.

UNIDY2002: I agree. I'm also wondering whether we should provide torch in our docker runtime. What do you think?

Aionw: Yes—if this Docker image is intended to support the `WITH_EP=ON` PG/EP artifacts it builds, installing a pinned PyTorch version is the cleanest way to keep the image functional. However, installing Torch only in the runtime stage is not sufficient: the builder should install the same pinned Torch build, prewarm the adapter, and copy the cache, while the runtime installs that exact version so the cache key, extension ABI, and embedded Torch library paths all match. This also avoids shipping the full compiler/toolkit stack and paying the cold-build cost at container startup.

If adding Torch to the default image is unacceptable because of image size, I would instead publish a separate PG/EP image variant (or use a devel image with the JIT toolchain) and stop building `WITH_EP=ON` in this lean runtime image. What I would avoid is the current middle state: the image contains PG/EP artifacts but cannot import `mooncake.pg` even after a user installs Torch.

UNIDY2002: Reworked on the dockerfile build process, and tested both cuda and musa cases. They built and passed checks successfully.

PR #3678, opened 2026-08-25
[TENT] Fix IPC relocation to not switch to the peer's device ordinal

staryxchen:
## Description

`NVLinkTransport::relocateSharedMemoryAddress` called
`cudaSetDevice(location.index())` before `cudaIpcOpenMemHandle`, where
`location` is the **peer's** `BufferDesc` location — a device ordinal
relative to the **peer's** `CUDA_VISIBLE_DEVICES`. In the local process
that ordinal may be out of range (when the peer process sees more devices
than we do) or may denote a different physical GPU.

With per-instance `CUDA_VISIBLE_DEVICES` — a common way to partition GPUs
between prefill/decode workers — every transfer targeting the peer's
higher-numbered devices failed (invalid device ordinal) with **zero
diagnostics**: `CHECK_CUDA` returned the error silently, and the broken
device-switch sequence also left the caller's CUDA device state
inconsistent on error paths.

Fix: open the IPC handle on the **caller's current device**.
`cudaIpcMemLazyEnablePeerAccess` handles the mapping and peer access, and
the subsequent copies already run cross-device via P2P (`startTransfer`
never switches devices either).

`MnnvlTransport` had the same pattern; its driver calls are
device-agnostic (`cuMemSetAccess` grants all local devices RW), so the
switch is removed there as well.

Also included, from the same debugging session:

- **`CHECK_CUDA` / `CHECK_CU` now `LOG(ERROR)` before returning** — the
  silent early returns are exactly why this bug produced no diagnostics.
- **`NVLinkTransport::submitTransferTasks` rolls back half-appended task
  entries on synchronous failure**: no I/O has started at that point, so
  stale never-started entries must not linger in the sub-batch (also a
  prerequisite for making submit-stage failover safe).
- New `tent_nvlink_transport_test`: a real IPC handle is exported by a
  forked child (a process cannot open its own exported handle) whose
  `BufferDesc` claims an out-of-range ordinal (`cuda:99`); asserts
  relocation success, data integrity through the mapping, and
  single-mapping reuse under concurrency.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
# new unit test — NOT registered in ctest (needs a real GPU: it opens a
# CUDA IPC handle exported by a forked child). Run manually on a GPU box,
# one case per process (a fork after the parent touched CUDA leaves the
# child's runtime unusable):
./tent_nvlink_transport_test \
    --gtest_filter=NVLinkTransportTest.RelocateToleratesOutOfRangePeerDeviceOrdinal
./tent_nvlink_transport_test \
    --gtest_filter=NVLinkTransportTest.RelocateCachesSingleMappingForConcurrentPeersWithBogusOrdinal

# full TENT suite
```

**Test results:**
- [x] Unit tests pass — new test 2/2 on an 8×H20 machine (each case run in
  its own process; also skips cleanly without a GPU — the forked child
  reports no CUDA device → `GTEST_SKIP`)
- [x] Integration tests pass — full TENT suite 44/44
- [x] Manual testing done (GPU e2e, same-machine PD disaggregation, sglang
  prefill TP1 + decode TP4 with per-instance `CUDA_VISIBLE_DEVICES`
  exposing the ordinal mismatch):
  - Before: first request fails — every NVLink transfer to the peer's
    higher-numbered devices errors out with no log output
  - After: sanity probes 3/3, sustained load transfers 40+ GB over NVLink
    with zero transfer failures; NVLink-mapped memory verified by reading
    back the expected data pattern through the relocation mapping

## Checklist

- [ ] I have updated the documentation (if applicable) — N/A, no doc surface
- [ ] For changes >500 LOC: I have filed an RFC issue — N/A, 425 LOC

## AI Assistance Disclosure

- [ ] No AI tools were used

CodeBuddy (Claude) assisted with root-cause analysis (instrumented
reproducer), the fix, the fork-based test design, and this PR
description. The human submitter has reviewed every changed line and can
defend the change end-to-end.

### DIFF
--- a/mooncake-transfer-engine/tent/include/tent/common/status.h
+++ b/mooncake-transfer-engine/tent/include/tent/common/status.h
@@ -42,21 +42,28 @@
     } while (0)
 
 #ifdef USE_CUDA
+#include <glog/logging.h>
+
 #define CHECK_CUDA(call)                                                      \
     do {                                                                      \
         auto err = call;                                                      \
-        if (err != cudaSuccess)                                               \
-            return Status::InternalError(std::string(#call) + ": " +          \
-                                         cudaGetErrorString(err) + LOC_MARK); \
+        if (err != cudaSuccess) {                                             \
+            auto _msg = std::string(#call) + ": " + cudaGetErrorString(err) + \
+                        LOC_MARK;                                             \
+            LOG(ERROR) << "CUDA error: " << _msg;                             \
+            return Status::InternalError(std::move(_msg));                    \
+        }                                                                     \
     } while (0)
 
-#define CHECK_CU(call)                                                        \
-    do {                                                                      \
-        auto err = call;                                                      \
-        if (err != CUDA_SUCCESS) {                                            \
-            return Status::InternalError(std::string(#call) + ": cuResult " + \
-                                         std::to_string(err) + LOC_MARK);     \
-        }                                                                     \
+#define CHECK_CU(call)                                       \
+    do {                                                     \
+        auto err = call;                                     \
+        if (err != CUDA_SUCCESS) {                           \
+            auto _msg = std::string(#call) + ": cuResult " + \
+                        std::to_string(err) + LOC_MARK;      \
+            LOG(ERROR) << "CUDA driver error: " << _msg;     \
+            return Status::InternalError(std::move(_msg));   \
+        }                                                    \
     } while (0)
 #endif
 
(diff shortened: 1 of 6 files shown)


alogfans (approved):
LGTM

PR #3679, opened 2026-08-25
[TENT] Share the CUDA IPC handle across buffers in one cudaMalloc segment

staryxchen:
## Description

`NVLinkTransport::addMemoryBuffer` deduplicates registrations by cudaMalloc
base address — necessary because caching allocators (e.g. PyTorch)
sub-allocate many logical buffers (KV pool, aux pool, linear-attention
state pool, ...) inside one larger `cudaMalloc` segment, and
`cudaIpcGetMemHandle` works at segment granularity.

The bug: only the **first** registration of a segment stamped the IPC
handle on its `BufferDesc`. Every later `BufferDesc` covering the same
segment was expanded to segment granularity and tagged NVLink-capable,
but carried **no `shm_path`**. Cross-process NVLink submission for those
buffers then failed in `relocateSharedMemoryAddress` with a silent
`NeedsRefreshCache` — no log, no metric label — and the engine failed
every such transfer over to the fallback transport.

Observed on a same-machine PD disaggregation setup: 206k silent
failovers out of 25.8M requests (~0.8%) — exactly the share of transfers
whose destination buffers happened to be the second-and-later
registrations of their segments. That fallback churn also multiplies
fallback-transport endpoint creation.

Fix: cache the serialized handle per base address and stamp it on every
desc covering the segment. Also add rate-limited `LOG(WARNING)` to the
previously silent relocation failure paths (no covering buffer / buffer
without an IPC handle / non-cuda target location), so this class of
misregistration is diagnosable instead of invisible.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
# full TENT suite

# GPU e2e (same-machine PD disaggregation, sustained load):
# metrics: tent_transport_failover_total{from="nvlink",...} before/after
```

**Test results:**
- [x] Unit tests pass — full TENT suite 44/44 (failover / metrics suites
  include submit-stage failure scenarios that exercise the code paths
  around this change)
- [x] Integration tests pass — N/A (no integration surface changed)
- [x] Manual testing done (GPU e2e, same-machine PD, sglang TP1 prefill +
  TP4 decode): NVLink failover count drops from 206k+/3min to **0**;
  every transfer completes on NVLink directly; the rate-limited warnings
  stay silent when registration is correct

## Checklist

- [ ] I have updated the documentation (if applicable) — N/A, no doc surface
- [ ] I have added tests to prove my changes are effective — a unit test
      would need to reproduce a caching allocator's segment sharing (two
      distinct registered ranges inside one allocation) plus a real CUDA
      IPC round trip; covered instead by the GPU e2e measurement above.
      Happy to add one if reviewers prefer.
- [ ] For changes >500 LOC: I have filed an RFC issue — N/A, 37 LOC

## AI Assistance Disclosure

- [ ] No AI tools were used

CodeBuddy (Claude) assisted with root-cause analysis (metrics-based
failure-share attribution), the fix, and this PR description. The human
submitter has reviewed every changed line and can defend the change
end-to-end.


catyans (approved):
Reviewed the current head. Caching the serialized CUDA IPC handle by the underlying `cudaMalloc` base address and copying it into every registered `BufferDesc` that belongs to the same allocation looks correct. The cache synchronization and buffer-removal path also appear consistent, and I did not find a blocking correctness issue.

As a non-blocking follow-up, it would still be useful to add an explicit `<unordered_map>` include and a focused test that registers two ranges from the same allocation and verifies that both receive the same non-empty IPC handle.

The current CI is green. Approving.

staryxchen:
Thanks! Both follow-ups are addressed in 4ec3a69a:

- Added the explicit `<unordered_map>` include. The pre-existing `relocate_map_` member relied on the same transitive include, so this covers both uses.
- Added `NVLinkIpcHandleSharingTest.TwoRangesInOneCudaMallocSegmentShareHandle`: it registers two ranges sub-allocated inside one cudaMalloc segment and asserts both descs receive the same non-empty serialized IPC handle, that both expand to the full segment, and that the handle matches a fresh `cudaIpcGetMemHandle` export. Since the test needs a real CUDA device, its target is built under `USE_CUDA` but not registered with `add_test`, following the same manual-run convention as the NVLink test in #3678. Verified it passes on this change and fails on the parent commit with the original symptom (the second buffer's `shm_path` staying empty).

catyans (approved):
Re-reviewed the current head after the upstream merge. The explicit `<unordered_map>` include and the real-CUDA two-range regression test address the earlier follow-ups, and the conflict resolution preserves the intended NVLink changes. CI is green. Approved.

Non-blocking suggestion: because `tent_nvlink_ipc_handle_sharing_test` is currently built but not registered with `add_test`, please consider wiring it into a GPU-capable CI lane (or factoring the handle-cache behavior behind a testable seam) so this regression is exercised automatically rather than only by manual runs.

alogfans (approved):
LGTM

PR #3677, opened 2026-08-25
[TENT] Cache metrics label cells to remove hot-path lock contention

staryxchen:
## Description

TENT's hot-path metrics (per-transport counters, attempt counters,
latency/size/stage histograms) were recorded through yalantinglibs'
dynamic metrics, whose `inc()`/`observe()` take a mutex-protected map
lookup on every update. The map is sharded by label hash, and TENT
labels come from small fixed enums (`TransportType`, `OpCode`), so all
updates for a given label land on the same shard mutex. After #3648
removed the engine-level global lock, this metrics layer became the
next serialization point: tebench 4K/1 with metrics enabled plateaued
(~170k ops/s flat at t>=4, latency growing with thread count —
47us@t=8, 88us@t=16) while `TENT_METRICS_ENABLED=false` reached ~490k.

Since the label domains are compile-time-known enums, this change
resolves each label's atomic cell once and caches the pointer in a
per-label-index slot, reducing the steady-state update to relaxed
atomic adds:

- New `tent/include/tent/metrics/cached_metric.h`:
  `CachedDynamicCounter<N>` exposes `resolve()` and `incCached()`
  (a relaxed `fetch_add` on the cached cell); `CachedDynamicHistogram<N>`
  is composed of cached-cell bucket counters plus the sum counter,
  replacing ylt's dynamic histograms and the parallel standalone sum
  counters (the sum becomes part of the histogram).
- `TentMetrics` slots N=1 metrics by `TransportType` and N=2 attempt
  metrics by (transport, operation), resolving cells lazily with
  race-free publication (release/acquire; concurrent first-use
  resolutions produce the same cell). Labels are never erased, so
  resolved cells stay valid for the process lifetime. The rare
  failover counter keeps the plain locked path.
- Prometheus/JSON output is unchanged (same custom serializers, now
  walking the histogram's own bucket/sum counters).

Performance, tebench 4K/1 write, local pair (same machine,
back-to-back runs; a background load was present, so the comparison is
relative, not absolute):

| threads | before (metrics on) | after (metrics on) | after (metrics off) | on/off gap |
|---|---|---|---|---|
| 1  | 127k ops/s / 7.9us | 138k / 7.2us | 139k / 7.2us | -0.7% |
| 4  | 142k | 323k | 323k | 0% |
| 8  | 170k / 47.2us | 478k / 16.7us | 488k / 16.4us | -2.0% |
| 16 | 181k / 88.4us | 663k / 24.1us | 682k / 23.5us | -2.8% |

[...]

PR #3680, opened 2026-08-25
[TENT] Add task failure reason counter and in-flight gauges

staryxchen:
## Description

Follow-up to #3677 (stacked on `tent/metrics-cell-cache`, merge after it).
Adds three metrics aimed at production diagnosis, all recorded through the
cached label cells (no locks on the hot path):

- **`tent_task_failures_total{transport, reason}`** — task-level failures
  with reason `submit` / `poll` / `timeout` / `canceled`. Motivation: in the
  CI incident that motivated this, hundreds of failures collapsed into a
  single `write_failures_total{transport="..."}` counter with no way to
  tell whether ops died at submission or in flight without log scraping.
  - `TaskInfo.failure_stage` marks where the **first** failure originated,
    set at the submit-rejection sites (commitPreparedSubmit /
    dispatchQueuedOwner / resubmitTransferTask) and at poll observation
    (updateTaskStatusAfterPoll), first-set-wins. A poll-observed failure
    stays `poll` even when the failover resubmit is synchronously
    rejected — the root cause, not the last bookkeeping event.
  - Transport attribution uses `attempt_type` (captured at attempt start),
    so a submit-stage failure still attributes to the real transport even
    though `task.type` is UNSPEC by then.
  - TIMEOUT/CANCELED outcomes were previously not recorded as failures
    anywhere; they are now.
  - Additive: the legacy `read/write_failures_total` counters are
    unchanged, so existing dashboards keep working.
- **`tent_inflight_attempts{transport}`** (gauge) — incremented on attempt
  submit, decremented on attempt finish. Shows how much transfer work is
  sitting in the engine right now; a persistently high or stuck value is
  the signature of a stalled pipeline (previously required log scraping
  plus Little's law to detect).
- **`tent_registered_buffer_bytes{transport}`** (gauge) — maintained on
  register/unregisterLocalMemory from `desc.transports` (the transports
  that actually registered), pairing with the failure counters to answer
  "registered but not moving".

All labels are small-cardinality enums (4 reasons, transport enum), per
the metrics label-domain constraint.

## Performance

tebench 4K/1 write, local pair, same back-to-back conditions as #3677
(background load present; comparisons are relative):

| threads | this PR, metrics on | this PR, metrics off | on/off gap | #3677 only, metrics on |
|---|---|---|---|---|
| 1  | 137k ops/s / 7.3µs | 137k / 7.3µs | 0%    | 138k / 7.2µs |
| 2  | 171k               | 175k               | -2.3% | 178k |
| 4  | 316k               | 324k               | -2.5% | 323k |
| 8  | 480k / 16.7µs      | 498k / 16.1µs      | -3.6% | 478k / 16.7µs |
| 16 | 669k / 23.9µs      | 669k / 23.9µs      | 0%    | 663k / 24.1µs |

With metrics enabled, this PR is indistinguishable from #3677 alone
(within run-to-run variance), and the metrics on/off gap stays within
the <5% acceptance bound. The only hot-path addition is the in-flight
gauge's pair of relaxed atomic adds per attempt; the failure-reason
counter and registered-bytes gauge are off the hot path.

## Module

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
cmake -S mooncake-transfer-engine -B build-tent -DCMAKE_BUILD_TYPE=Release \
      -DUSE_TENT=ON -DTENT_METRICS_ENABLED=ON -DYLT_ENABLE_IBV=ON -DWITH_STORE=OFF
cmake --build build-tent -j16
for t in build-tent/tent/tests/*_test; do $t || exit 1; done
```

**Test results:**
- [x] Unit tests pass — all 40 tent test binaries pass; two new tests in
  `tent_metrics_recording_test` cover submit-attributed and poll-attributed
  task failures (including the poll-failure-followed-by-rejected-failover
  path, which initially misattributed as submit and is exactly what
  `failure_stage` fixes), in-flight returning to zero after terminal
  states, and registered bytes rising on register and returning to zero on
  unregister
- [x] Manual testing done — reason series asserted via the Prometheus
  text output; gauge serialization covered by the JSON/Prometheus snapshot
  helpers

## Checklist

- [ ] I have updated the documentation (if applicable) — metric names are self-documenting via # HELP
- [ ] For changes >500 LOC: I have filed an RFC issue — 329 added LOC, additive metrics only

## AI Assistance Disclosure

- [ ] No AI tools were used

CodeBuddy (AI coding assistant) assisted with implementation, testing, and
benchmarking under human direction. Every changed line has been reviewed
by the submitter, who can defend the change end-to-end.

### DIFF
--- a/mooncake-transfer-engine/tent/src/metrics/tent_metrics.cpp
+++ b/mooncake-transfer-engine/tent/src/metrics/tent_metrics.cpp
@@ -35,6 +35,20 @@ namespace {
 const char* operationName(Request::OpCode operation) {
     return operation == Request::READ ? "read" : "write";
 }
+
+const char* taskFailureReasonName(TentMetrics::TaskFailureReason reason) {
+    switch (reason) {
+        case TentMetrics::TaskFailureReason::Submit:
+            return "submit";
+        case TentMetrics::TaskFailureReason::Poll:
+            return "poll";
+        case TentMetrics::TaskFailureReason::Timeout:
+            return "timeout";
+        case TentMetrics::TaskFailureReason::Canceled:
+            return "canceled";
+    }
+    return "unknown";
+}
 }  // namespace
 
 Status TentMetrics::initialize(const MetricsConfig& config) {
@@ -195,6 +209,7 @@ void TentMetrics::shutdown() {
     // Clear metric vectors
     counters_.clear();
     histograms_.clear();
+    gauges_.clear();
 
     // Reset bound port so httpPort() returns 0 after shutdown, not a stale
     // port from a previous initialization. Without this, a re-initialize
@@ -221,8 +236,14 @@ void TentMetrics::registerMetrics() {
         &failover_total_,
         &transport_attempts_total_,
         &transport_attempt_failures_total_,
+        &task_failures_total_,
         &deadline_infeasible_total_,
         &quarantined_batches_total_,
+    };
+
+    gauges_ = {
+        &inflight_attempts_,
+        &registered_buffer_bytes_,
     };
 
     // Register the N=1 per-transport histograms for unified serialization.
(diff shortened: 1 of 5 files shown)


mooncake-transfer-engine/tent/src/metrics/tent_metrics.cpp
```diff
@@ -301,6 +322,41 @@ void TentMetrics::recordBatchQuarantined() {
...
+}
+
+void TentMetrics::recordRegisteredBufferBytes(TransportType tp, int64_t delta) {
+    if (!initialized_ || !runtime_enabled_.load(std::memory_order_relaxed))
+        return;
+    registered_buffer_bytes_.incCached(
+        transportSlot(tp),
+        [tp] { return std::array<std::string, 1>{transportTypeName(tp)}; },
+        delta);
+}
```
alogfans, line 360: These gauges become incorrect if `setEnabled()` changes between the matching operations. For example:

- An attempt starts while metrics are enabled, so `tent_inflight_attempts` is incremented.
- Metrics are disabled before the attempt finishes.
- `recordInflightAttemptFinished()` returns without decrementing, leaving the gauge permanently positive.

The reverse transition produces a negative gauge: an attempt starts while disabled, metrics are then enabled, and only the decrement is recorded. The same issue affects `tent_registered_buffer_bytes` when registration and unregistration occur under different enablement states.

staryxchen: Good catch — this was a real bug, and it applies to both gauges (`tent_inflight_attempts` and `tent_registered_buffer_bytes`). Fixed in b3910100.

**Approach**: gauge updates are now maintained whenever initialized and ignore `setEnabled()` entirely. The reasoning is a distinction between the two metric kinds:

- **Counters/histograms are samples** — dropping them while disabled only loses data, which is what the runtime switch is for.
- **Gauges are state** — `tent_inflight_attempts` and `tent_registered_buffer_bytes` reflect engine state through *paired* add/sub operations (attempt start/finish, register/unregister). Neither half of a pair may be skipped, regardless of when the switch toggles, or the value is corrupted.

So the runtime switch now cleanly means "stop sampling", while state tracking stays consistent across any enable/disable transition.

**Tests**: added `InflightGaugeSurvivesEnableToggle` and `RegisteredBytesGaugeSurvivesEnableToggle`, covering both straddle directions (enabled→disabled and disabled→enabled between the two halves of each pair), asserting the gauge returns exactly to baseline.

**Cost while disabled**: maintaining the gauges costs two relaxed atomic adds per op (the register/unregister bookkeeping is cold-path). Measured with tebench 4K/1: the on/off throughput gap stays within run-to-run noise (~2-4% at t=8/16), since the switch still skips the ~21 per-op counter/histogram updates that make up the bulk of the cost. Compile-time disable (`TENT_METRICS_ENABLED=0`) remains fully zero-overhead.

alogfans (approved):
LGTM, thanks!

PR #3682, opened 2026-08-25
[TENT] Recover from synchronous submit failures via transport failover

staryxchen:
## Description

TENT's cross-transport failover only fired from the **completion** stage.
When `submitTransferTasks()` failed synchronously (e.g. a transport-side
IPC relocation error), `commitPreparedSubmit()` / `dispatchQueuedOwner()`
clobbered `task.type` to `UNSPEC`, `pollTaskStatus()` short-circuits
`UNSPEC` to `FAILED`, and the task died **without ever trying the
remaining candidate transports**. On the metrics plane this surfaced as
`tent_transport_attempt_failures{<transport>}=N` followed by
`tent_write_failures{transport="unspec"}=N` — with no log line at the
decision point.

Fix: reuse the existing poll-time failover machinery. On a synchronous
submit failure, walk the remaining candidate transports through
`resubmitTransferTask()`, bounded by `max_failover_attempts_`. The two
hazards documented in `docs/source/design/tent/failover.md` are handled:

- **Merged requests**: only the **owner** task of each merged request is
  resubmitted; derived aliases mirror the owner's recovered route
  (transport, sub-batch slot, status), so the merged transfer is never
  re-posted per alias.
- **Partial enqueue**: relies on transports rolling back half-appended
  entries on synchronous failure (NVLink does; transports that start
  per-request work inside the submit loop may cause one idempotent
  re-post — called out in the design doc).

A task that exhausts all candidates surfaces `FAILED` attributed to the
**last attempted transport** (`recordTaskCompletionMetrics` falls back to
`attempt_type` when `type` is `UNSPEC`), never mislabeled "unspec". The
Known Gaps entry in `failover.md` is updated accordingly.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
```

**Test results:**
- [x] Unit tests pass — full TENT suite 44/44, including 6 new cases:
  - `SubmitStageFailureFailsOverToSecondary` — synchronous submit failure
    recovered on the fallback transport
  - `SubmitStageFailureExhaustsBudget` — both transports reject → terminal
    FAILED, exactly one failover attempt, budget respected
  - `SubmitStageFailoverWithDerivedTasks` — merged request re-posted
    exactly once; both tasks' per-task status agree
  - `QueuedPathSubmitStageFailover` — runtime-queue dispatch path recovers
  - `SubmitStageFailoverRecordsFailoverAndCompletion` —
    `tent_transport_failover_total{from,to}` fires; completion attributed
    to the fallback
  - `SubmitStageExhaustionAttributesToRealTransport` — terminal failure
    attributed to the last attempted transport, `unspec` stays 0
- [x] Manual testing done (GPU e2e, sglang PD disaggregation):
  - **Failover works under production-shaped load**: same-machine
    topology where a transport path fails synchronously at high rate —
    251,904 submit-stage failovers in 2 minutes, every one recovered on
    the fallback transport, zero write failures (before this change the
    first such failure killed the request)
  - **No regression on the healthy path**: 12h cross-machine TP8+TP4
    soak with the 5-commit stack this PR is part of — at the 58-minute
    mark: 22,289 requests / 0 failures, 2.8 TB transferred, p50=1.12s /
    p99=3.20s, `failover=0`, zero `ENOMEM`/`bad_alloc` (final results to
    be appended when the run completes)

## Checklist

- [x] I have updated the documentation (if applicable) —
      `docs/source/design/tent/failover.md` Known Gaps updated
- [ ] For changes >500 LOC: I have filed an RFC issue — N/A, 434 LOC

## AI Assistance Disclosure

- [ ] No AI tools were used

CodeBuddy (Claude) assisted with root-cause analysis, the fix, the test
suite, and this PR description. The human submitter has reviewed every


catyans (changes requested):
The synchronous-failure failover path assumes that a transport returning an error accepted none of the submitted work. That is not guaranteed. For example, a backend submission can accept only a prefix of the requests and then report a short/failed submission. At that point, some operations may already be in flight.

Retrying the entire physical group on the fallback transport can therefore execute the accepted prefix twice. It also stops tracking the original sub-batch, so its completions and resource lifetime may no longer be accounted for. This is especially problematic because a synchronous submit error does not provide an atomic rollback guarantee.

Please make the transport submission contract explicitly all-or-nothing, return the accepted range/ownership to the caller, or cancel and drain any accepted work before failover. A regression transport that accepts the first N requests and then returns an error would help verify that every request executes exactly once and that both the original and fallback attempts settle cleanly.

staryxchen:
Thanks for the careful review — you're right that the failover path silently relied on an all-or-nothing assumption that was never written down, and that one in-tree transport did not actually satisfy it.

**Contract made explicit** (c31802a6): `Transport::submitTransferTasks()` now documents the all-or-nothing contract: on an error return, no request may have been dispatched and the sub-batch must be left exactly as before the call; a transport that starts accepting requests and then hits an error must roll its partial state back before returning. I chose this over returning the accepted range or cancel-and-drain because every in-tree transport already dispatches work only at the end of submit (audit below), so the contract matches reality with a single fix, and it keeps the failover path simple and transport-agnostic.

**Audit of in-tree transports:**

- **RDMA / TCP / GDS / BufIo**: build tasks first, dispatch at the end; all error returns happen before any work is queued (BufIo marks individual tasks FAILED rather than returning mid-loop errors). Already compliant.
- **SHM**: the real violator — `startTransfer()` (a synchronous memcpy) ran inside the per-request loop, so a mid-batch relocation failure left earlier requests already executed. Fixed with a two-phase submit: relocate/validate every request first, then execute any of them.
- **NVLink**: mid-loop failures (e.g. IPC relocation errors) left never-started PENDING tasks in the sub-batch — no double execution, but stale sub-batch state. Fixed by the rollback in #3678.
- **io_uring**: not fully compliant yet; I'll address it as a follow-up.

**Regression tests**, as you suggested:

- `ShmTransportTest.SubmitIsAllOrNothingWhenRelocateFailsMidBatch` — three requests where the second fails relocation: asserts the error return leaves the sub-batch empty and the valid first request's data was never copied.
- `EngineFailoverE2E.SubmitStagePartialAcceptanceExecutesExactlyOnce` — a transport that accepts the first 2 of 4 requests into its sub-batch and then returns an error, rolling the prefix back per the contract: asserts every request executes exactly once on the fallback, the original sub-batch settles empty, and all tasks complete.

**On "stops tracking the original sub-batch"**: under the contract the failed submit leaves nothing accepted in the original sub-batch, so there is no in-flight work to account for. The logical tasks are re-pointed to the fallback transport's sub-batch (derived aliases mirror their owner — covered by `SubmitStageFailoverWithDerivedTasks`), and the failed transport's sub-batch stays owned by the batch and is freed with it at `freeBatch()`.

catyans:
Thanks for documenting the all-or-nothing submission contract, fixing SHM, and adding the rollback/failover tests. Those changes address the issue for contract-compliant transports.

However, I do not think the io_uring violation can be deferred until after this PR merges. `IOUringTransport::submitTransferTasks()` still appends tasks and prepares SQEs incrementally, then treats any `io_uring_submit()` result smaller than `request_list.size()` as an error. A positive short submission means a prefix is already dispatched, but the engine will now rely on the new contract and retry the entire group on the fallback transport. That is the original duplicate-execution and lost-tracking scenario.

The new fake-transport test explicitly rolls its accepted prefix back before returning, so it verifies the compliant path but does not cover the known in-tree violation.

The io_uring implementation does not necessarily have to be fixed in this exact PR, but the dependency must be resolved before this failover change merges. Please either:

- make io_uring satisfy the all-or-nothing contract in this PR or a prerequisite PR; or
- exclude io_uring from synchronous-submit failover until it is compliant.

A regression test that injects a positive short `io_uring_submit()` result would directly cover the failure mode. I would keep the change request open until every transport eligible for this failover path satisfies the contract.

staryxchen:
Hi @catyans io_uring is now addressed. One caveat on verification: none of the environments I have access to can actually run io_uring, but the logic-level paths are covered. Since you're experienced with io_uring, could you help verify that the tests run (and pass) in your environment?

catyans (approved):
Re-reviewed the io_uring changes on `f1b7faf414`. Validation and staging now complete before the sub-batch or SQ ring is mutated, and once the commit phase starts the transport no longer returns an error after a positive short submission. Deferred SQEs remain attached to PENDING tasks and are retried from the poll path, so the engine cannot fail over and re-execute an already-dispatched prefix. Moving the `iovec` into `IOUringTask` also gives it the required completion lifetime. The original all-or-nothing blocker is resolved, and CI is green. Approved.

Non-blocking test suggestion: `ShortSubmitIOUringTransport` currently lets the real kernel submit all SQEs and then under-reports the count with a synthetic deficit. That exercises the transport's accounting loop, but not a ring that genuinely retains unsent SQEs. If practical, a more deterministic submit seam that leaves a suffix pending in the SQ ring would strengthen coverage of the deferred-flush path.

alogfans (approved):

PR #3683, opened 2026-08-25
[TENT] Keep the worker from blocking on its own submit queue

### LINKED ISSUE #3637: [tent] BoundedMPSCQueue push() livelock: full submit queue spins both producers and RDMA worker threads forever, engine never recovers

he-yufeng:
## Description

Fixes #3637.

`BoundedMPSCQueue::push()` spins with `yield()` when the queue is full and has no failure exit. The TENT worker thread calls `push()` on its own priority queues when re-enqueueing retried slices and when promoting entries between levels, so once a queue fills during a sustained failure period the worker wedges itself: it stops popping, every producer then spins in `push()`, and the engine cannot recover even after the underlying RDMA problem clears.

This adds a non-blocking `try_push` and routes every tick-internal re-enqueue through it. A full queue now parks the entry in a per-worker overflow list that the next tick flushes first, so the worker keeps draining and the producer-side spin keeps its exit. Parked entries stay counted as inflight, which keeps the accounting continuous and keeps the worker from suspending while a flush is pending. The external `submit()` path is unchanged; producer backpressure semantics are untouched.

## Module

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

**Test commands:**
```bash
# new deterministic regression test (no RDMA stack needed)
ctest -R bounded_mpsc_queue_test --output-on-failure
```

**Test results:**
- [x] Unit tests pass — `bounded_mpsc_queue_test` (4 cases: full-queue try_push reports instead of spinning, FIFO drain after full, the park/pop/flush overflow pattern, push noop on empty list). Built and run locally on macOS against the header-only queue.
- [ ] Integration tests pass (if applicable) — the RDMA transport does not build on this machine (no rdma-core headers), so the workers.cpp compile is left to CI.

## Checklist

- [x] I have run pre-commit on the files changed in this PR and all hooks pass (the cmake-format hook rewrites unrelated blocks of `tent/tests/CMakeLists.txt` even on pristine main, so those were left out)

## AI Assistance Disclosure

Drafted with Kimi K3 assistance. The invariants this change relies on (the worker never blocks on its own queue; inflight accounting is unchanged: re-enqueue +1 and the call-site -1 net to zero, same as the old submit path) were re-verified against the code before submission.

### DIFF
--- a/mooncake-transfer-engine/tent/include/tent/common/concurrent/bounded_mpsc_queue.h
+++ b/mooncake-transfer-engine/tent/include/tent/common/concurrent/bounded_mpsc_queue.h
@@ -57,11 +57,9 @@ struct BoundedMPSCQueue {
         Cell *cell = &buffer[pos % Capacity];
 
         uint64_t seq = cell->sequence.load(std::memory_order_acquire);
-        intptr_t dif =
-            static_cast<intptr_t>(seq) - static_cast<intptr_t>(pos);
+        intptr_t dif = static_cast<intptr_t>(seq) - static_cast<intptr_t>(pos);
         if (dif != 0) return false;
-        if (!tail.compare_exchange_weak(pos, pos + 1,
-                                        std::memory_order_acq_rel,
+        if (!tail.compare_exchange_weak(pos, pos + 1, std::memory_order_acq_rel,
                                         std::memory_order_relaxed)) {
             return false;
         }
--- a/mooncake-transfer-engine/tent/include/tent/common/concurrent/bounded_mpsc_queue.h
+++ b/mooncake-transfer-engine/tent/include/tent/common/concurrent/bounded_mpsc_queue.h
@@ -43,27 +43,31 @@ struct BoundedMPSCQueue {
     ~BoundedMPSCQueue() = default;
 
     void push(T &slice_list) {
-        if (slice_list.num_slices == 0) return;
-        uint64_t pos;
-        while (true) {
-            pos = tail.load(std::memory_order_relaxed);
-            Cell *cell = &buffer[pos % Capacity];
-
-            uint64_t seq = cell->sequence.load(std::memory_order_acquire);
-            intptr_t dif =
-                static_cast<intptr_t>(seq) - static_cast<intptr_t>(pos);
-            if (dif == 0) {
-                if (tail.compare_exchange_weak(pos, pos + 1,
-                                               std::memory_order_acq_rel,
-                                               std::memory_order_relaxed)) {
-                    cell->data = slice_list;
-                    cell->sequence.store(pos + 1, std::memory_order_release);
-                    return;
-                }
-            } else if (dif < 0) {
-                std::this_thread::yield();
-            }
+        while (!try_push(slice_list)) {
+            std::this_thread::yield();
         }
+    }
+
+    // Non-blocking push: returns false when the queue is full. The worker
+    // thread re-enqueues through this so a full queue parks the entry in a
+    // local overflow instead of wedging the only consumer (issue #3637).
+    bool try_push(T &slice_list) {
+        if (slice_list.num_slices == 0) return true;
+        uint64_t pos = tail.load(std::memory_order_relaxed);
+        Cell *cell = &buffer[pos % Capacity];
+
+        uint64_t seq = cell->sequence.load(std::memory_order_acquire);
+        intptr_t dif =
+            static_cast<intptr_t>(seq) - static_cast<intptr_t>(pos);
+        if (dif != 0) return false;
+        if (!tail.compare_exchange_weak(pos, pos + 1,
+                                        std::memory_order_acq_rel,
+                                        std::memory_order_relaxed)) {
+            return false;
+        }
+        cell->data = slice_list;
+        cell->sequence.store(pos + 1, std::memory_order_release);
+        return true;
     }
 
     T pop() {
--- a/mooncake-transfer-engine/tent/include/tent/common/concurrent/bounded_mpsc_queue.h
+++ b/mooncake-transfer-engine/tent/include/tent/common/concurrent/bounded_mpsc_queue.h
@@ -18,6 +18,7 @@
 #include <atomic>
 #include <thread>
 #include <unordered_set>
+#include <vector>
 
 namespace mooncake {
 namespace tent {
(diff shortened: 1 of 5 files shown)


catyans (changes requested):
Thanks for addressing the worker self-blocking problem. Introducing a non-blocking worker-side requeue path is the right direction, but the current overflow flushing order does not guarantee that parked slices will make progress.

`workerThread()` calls `flushTickOverflow()` before `asyncPostSend()` drains the bounded queues. When a target queue is full, the flush therefore fails; `asyncPostSend()` frees capacity only afterward. External producers already spinning in `push()` can refill those slots before the next worker tick, so under sustained submission pressure an older retry slice may remain in `requeue_overflow` indefinitely. Its task stays pending and its inflight accounting never settles.

The new regression test models `park -> pop -> flush`, while the production loop currently performs `flush -> pop`, so it does not cover this starvation scenario.

I suggest either treating the overflow list as a worker-local queue that is consumed before new shared-queue entries, or retrying the overflow immediately after the worker drains queue capacity in the same tick. A regression test with continuously contending producers would help verify the liveness guarantee.

The code-format check and CI Gate are also currently failing. I would request changes until the overflow liveness issue and CI failures are resolved.

he-yufeng:
@catyans Good catch, and you were right about both holes. Reworked in 032f39360:

- The overflow list is no longer flushed back into the shared queue at all. `asyncPostSend`'s priority drain consumes `requeue_overflow` before shared-queue entries at the same level, so a parked retry makes progress by order rather than by winning a CAS race against refilling producers. `flushTickOverflow` is gone.
- The regression test now models the production order: full queue, parked entry, queue drains, a producer instantly refills, and the next drain must yield the parked entry before the producer's fresh one.

Format check passes on the touched files (`code_format.sh -b upstream/main --check`; clang-format 20 via pip, since this Mac has no toolchain package for it) and the four queue tests pass. The RDMA transport still doesn't build here (no rdma-core on macOS), so the workers.cpp compile is on CI as before.

he-yufeng:
@catyans CI is green now across all legs (28 pass, 1 skip) at 781e5bb0d. The earlier tent-ci failures were a header self-containment issue (`bounded_mpsc_queue.h` used `std::vector` without including it, which my test's include order exposed on libstdc++), fixed in 032759bee. The rework itself (overflow drained before shared queues) is unchanged from what you reviewed.

catyans (approved):
Thanks for the update. I reviewed the changes through `781e5bb0d`.

Draining `requeue_overflow` directly before shared-queue entries at the same priority resolves the starvation scenario I raised: a parked worker retry no longer has to compete with producers to re-enter the bounded queue. The updated regression test now follows the production ordering and verifies that the parked entry is selected before a producer's newly queued entry.

The header self-containment issue is also fixed and CI is green. My previous concerns are resolved. This looks good to me now.

he-yufeng:
This has been sitting approved with green CI for a couple of days, and the merge gate still reports blocked. @catyans since your review cleared both threads, is anything else needed on my side, or is this waiting on a second approver / someone with merge rights?

alogfans (approved):
LGTM

PR #3684, opened 2026-08-25
TENT: flush a peer's pooled RPC connections when one of them fails

xiaodouzi666:
## Description

An idle pooled RPC connection only learns that its peer restarted when it is next used, and that first use fails. Discarding just that one client left every other connection pooled for the same peer equally stale, so the caller paid one failed call per stale connection. `TcpTransport` retries a bounded number of times (`max_retry_count`), so a peer that was already back up could still fail a transfer outright, and notifications, which have no retry above this layer, were dropped.

Drop the whole pool for that address instead, so the next attempt starts from a fresh connection.

This deliberately does not retry the failed call in place. `Pin`, `Unpin`, `Delegate` and `Notify` share this path and mutate remote state, and the error tells us nothing about whether the peer ran the handler: ylt reports both "never sent" and "ran, but the reply was lost" as `io_error`. Recovery stays with the layer that already owns it and knows whether its operation is safe to repeat.

The flush is skipped when the failure came back from the peer as a reply — a handler exception, an unknown function id, a rejected argument. The connection carried a full round trip in those cases, so it is fine, and so are the ones other callers of the same address are holding.

The pool also carries a generation. A client is handed out with the generation it came from, and a failure only flushes if that is still the current one. Several callers notice the same restart at once; without this the second one to fail would throw away the connections the first one has already re-established, and the pool would churn for as long as stale clients keep surfacing.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
cmake -G Ninja -DUSE_TENT=ON -DUSE_HTTP=ON -DBUILD_UNIT_TESTS=ON ..
ninja && ctest --output-on-failure
```

**Test results:**
- [x] Unit tests pass

`rpc_reconnect_test`, all on loopback with no NIC:

- `StalePoolIsFlushedSoTheNextCallSucceeds` — a concurrent burst parks two connections in the pool, the peer restarts on the same port, and the call after the first (unavoidable) failure has to succeed. Fails without the flush.
- `AFailedCallIsNotRetried` — the peer runs a side effect and only then fails; it must have run exactly once. Fails against a version that retries, reporting two executions.
- `HandlerFailureLeavesThePoolAlone`, `DeadPeerStillFails`, `ConcurrentStaleCallsAllRecover` — guards for behaviour that must not change. They pass either way, and their comments say so rather than implying otherwise.

Manual testing on a server with RDMA hardware (8×H20, 5×RoCE): full TENT suite 31/31, the new test 5/5 consecutive runs, and a negative control — this change removed from `rpc.cpp`, only that test rebuilt — fails there as required.

## Checklist

On the pre-commit box: every hook passes on the changed files except `cmake-format`, which rewrites 57 lines of pre-existing content in `tent/tests/CMakeLists.txt` — `main` is already not clean under the pinned v0.6.13. Reformatting it here would bury a 12-line addition in unrelated churn, so I left it alone. The CI format gate (clang-format 20, changed lines only) passes.

## AI Assistance Disclosure

- [ ] No AI tools were used

AI assistance was used while investigating the failure mode and while drafting the tests and this description. I have reviewed every changed line and can defend the change end to end.

### DIFF
--- a/mooncake-transfer-engine/tent/src/rpc/rpc.cpp
+++ b/mooncake-transfer-engine/tent/src/rpc/rpc.cpp
@@ -43,29 +43,52 @@ struct RpcHandlerScope {
 
 class ClientPool {
    public:
-    std::unique_ptr<coro_rpc_client> acquire() {
+    struct Lease {
+        std::unique_ptr<coro_rpc_client> client;
+        // Which generation the client was drawn from. Flushing is keyed on
+        // this: see clearIfCurrent().
+        uint64_t generation = 0;
+    };
+
+    Lease acquire() {
         std::lock_guard<std::mutex> guard(lock_);
         if (!idle_clients_.empty()) {
             auto client = std::move(idle_clients_.back());
             idle_clients_.pop_back();
-            return client;
+            return Lease{std::move(client), generation_};
         }
-        return nullptr;
+        return Lease{nullptr, generation_};
     }
 
+    // A client is only released after a call succeeded on it, which proves it
+    // is alive, so it goes back regardless of which generation it came from.
     void release(std::unique_ptr<coro_rpc_client> client) {
         if (!client) return;
         std::lock_guard<std::mutex> guard(lock_);
         idle_clients_.push_back(std::move(client));
     }
 
+    // Drop every idle client, but only on behalf of a caller that is still
+    // looking at the generation it drew from. Several callers notice a peer
+    // restart at once; without this, the second one to fail would throw away
+    // the connections the first one has already re-established, and the pool
+    // would churn for as long as stale clients keep surfacing.
+    void clearIfCurrent(uint64_t generation) {
+        std::lock_guard<std::mutex> guard(lock_);
+        if (generation != generation_) return;
+        ++generation_;
+        idle_clients_.clear();
+    }
+
     void clear() {
         std::lock_guard<std::mutex> guard(lock_);
+        ++generation_;
         idle_clients_.clear();
     }
 
    private:
     std::mutex lock_;
+    uint64_t generation_ = 0;
     std::vector<std::unique_ptr<coro_rpc_client>> idle_clients_;
 };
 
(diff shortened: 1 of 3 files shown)


catyans (changes requested):
Thanks for addressing recovery from stale pooled RPC connections after a peer restart. Reconnecting after detecting a stale client is the right direction, but the retry is currently applied too broadly in the generic `CoroRpcAgent::callCoroutine()` path.

`!call_result.has_value()` does not mean that the request was never executed. The server may have completed the handler and lost the response, in which case retrying changes the RPC semantics to at-least-once delivery. This affects non-idempotent control-plane operations routed through the same agent:

- `Notify` may invoke the callback twice.
- `Delegate` may execute `transferSync()` twice.
- `Pin` may acquire the same stage buffer twice and leak a reference.
- `Unpin` may release the buffer twice or too early.

Yalantinglibs also represents server-side handler and protocol errors as an unsuccessful `call_result`. The current code therefore clears the pool and retries not only transport failures, but also errors such as `rpc_throw_exception`, where reconnecting cannot fix the problem and may repeat partial side effects.

I suggest making retries opt-in for RPCs known to be idempotent, or adding a request ID with server-side deduplication. The retry decision should also inspect the RPC error code and exclude handler, serialization, and protocol errors.

The pool operations themselves are protected by a mutex, so I do not see a direct data race. However, concurrent failures can repeatedly call `clear()` and remove healthy clients established by another retry, causing unnecessary reconnect churn after a restart. A pool generation/epoch would allow all old leases to be invalidated without later failures clearing clients from the new generation.

The new tests only use idempotent echo handlers, and the multi-client test performs a single call after the restart. Please add coverage for an executed request with a lost response or another non-idempotent handler, an explicit server-side RPC error, and multiple stale calls failing concurrently.

I would request changes until the retry semantics are restricted or deduplicated and the concurrent recovery behavior is covered.

xiaodouzi666:
This is correct, and I've implemented the solutions in code, both the retry semantics and the error code issue have been solved, so the only limitation would be that this method's effect is a reduction of the reconnects, and the `ConcurrentStaleCallsAllRecover` test cannot assert that.

he-yufeng (approved):
Read through the pool logic carefully and the invariants hold. The generation check is what makes this safe under concurrency: the first stale failure bumps the generation and flushes, later failures from the same generation become no-ops, and connections established after the flush belong to the new generation and cannot be churned by a late second failure. The `peerAnswered` gate is the right line too: a handler-level exception says the socket is fine, so only connection-level failures flush, and the failed client itself is discarded either way.

The tests are honest about what they pin and what they don't: the stale-pool regression is pinned end to end (restart, first call fails, second succeeds), the handler-throw and dead-peer guards are labeled as guards rather than regression tests, and the side-effect handler provably runs exactly once, so nothing here moves toward at-least-once retries. CI is green (28 pass, 1 skip) including the new reconnect suite. I could not build the RDMA transport locally (no rdma-core on this machine), so the loopback verification here is read-through plus CI.

catyans (approved):
Thanks for reworking this. I reviewed the updated implementation at `afcf484b`.

Removing the automatic retry from the generic `CoroRpcAgent::callCoroutine()` path preserves the existing at-most-once behavior for `Notify`, `Delegate`, `Pin`, `Unpin`, and other potentially non-idempotent RPCs. Distinguishing peer-generated errors also avoids flushing the pool for handler or protocol failures.

The generation-guarded `clearIfCurrent()` addresses the concurrent stale-client case without allowing a later failure from an old generation to discard connections that another caller has already re-established. Returning a successfully used older-generation client to the pool is safe because the successful round trip proves that connection is alive.

The added side-effect and concurrent-recovery tests cover the important observable semantics. My previous blocking concerns are resolved. This looks good to me now.

staryxchen (approved):
LGTM.

PR #3685, opened 2026-08-25
TENT: add a segment warmup primitive

xiaodouzi666:
## Description

The first transfer towards a freshly opened segment pays for connection setup. On RDMA that is the endpoint and queue-pair bootstrap: 42 – 56 ms for a single NIC pair on the hardware below, and around a second when a whole NIC set has to come up. The classic engine grew `warmupEfaSegment()` for the same reason; TENT had no equivalent, so benchmarks and latency-sensitive callers either ate the stall on their first real transfer or hand-rolled a dummy one.

Adds `TransferEngine::warmupSegment(handle)`, routed to every installed transport through a new optional `Transport::warmupSegment()` that defaults to `NotImplemented`.

- RDMA pre-establishes endpoints from every enabled local context to every RDMA NIC the target advertises, skipping pairs that are already ready. Bootstrapping also brings up the notification queue pair that rides on the same endpoint.
- `NotImplemented` means both "this transport does not warm" and "it has nothing to warm towards this target" — a non-memory segment, or a peer that advertises no RDMA NIC. Neither is a failure, which is how the classic `warmupEfaSegment()` answers the no-devices case too.
- A transport that genuinely failed is reported even when another succeeded. The caller asked for the setup cost to be paid up front, and returning OK would hand back a target whose first transfer still stalls, with only a log line to say why. The failure leaves that transport on its lazy path, so the target stays usable either way.
- `LOCAL_SEGMENT_ID` is a successful no-op that reaches no transport: local transfers take the in-process path and have no connection state.
- C API: `tent_warmup_segment()`.

Only RDMA implements warming. TCP's data path runs on a worker pool, and the RPC agent that owns the connection pool is `thread_local`, so a connection warmed on the caller's thread is not one a transfer would use. Warming it usefully needs its own change, and is left out rather than shipped as a call that looks like it does something.

## Module

- [ ] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
cmake -G Ninja -DUSE_TENT=ON -DUSE_HTTP=ON -DBUILD_UNIT_TESTS=ON ..
ninja && ctest --output-on-failure
```

**Test results:**
- [x] Unit tests pass

`warmup_segment_test` pins the routing and aggregation contract with fake transports: the local segment reaches no transport, every installed transport is offered the target, `NotImplemented` is not a failure, a real failure surfaces alongside another transport's success, and the first of several failures is the one reported. Each case was checked against its own negative control — with the aggregation rule reverted, `AFailureIsReportedEvenIfAnotherTransportSucceeded` fails.

The RDMA implementation needs a NIC and a peer, so it is not covered by the unit tests. It was exercised separately, below.

Manual testing on a server with RDMA hardware (8×H20, 5×RoCE). Full TENT suite 31/31. A standalone harness opens a segment between two engines on the same host and times the first 4 KiB write, pinned to a single NIC, three rounds:

| | first write |
| --- | --- |
| cold, no warmup | 42 – 56 ms |
| after `warmupSegment()` | 0.57 – 0.75 ms |

`warmupSegment()` itself took about 45 ms for that one NIC pair. An independent run on a different day measured 42.4 – 49.0 ms cold, 0.44 – 0.56 ms warm and 41.4 – 43.5 ms for the warmup — close enough that the figures look like a property of the path rather than of one afternoon.

A second harness run without pinning a NIC warms every local context against every NIC the target advertises; that costs around a second, because the per-pair cost is what scales. Those numbers are not quoted here: the third round of that run was disturbed by unrelated load on the shared fabric (both cold and warm first-writes at 3.7 s, warm slower than cold), and the run is reported only for completeness.

The setup cost moves into the explicit call rather than disappearing; that is the point of the primitive. The absolute figure depends on the fabric and is not portable. What the runs establish is that the first-transfer stall is a stable structural cost rather than an occasional hiccup, and that it is gone once the segment is warmed.

## Checklist

On the pre-commit box: every hook passes on the changed files except `cmake-format`, which rewrites 57 lines of pre-existing content in `tent/tests/CMakeLists.txt` — `main` is already not clean under the pinned v0.6.13. Reformatting it here would bury a 12-line addition in unrelated churn, so I left it alone. The CI format gate (clang-format 20, changed lines only) passes.

## AI Assistance Disclosure

- [ ] No AI tools were used

AI assistance was used while investigating the cost of the first transfer and while drafting the tests and this description. I have reviewed every changed line and can defend the change end to end.

### DIFF
--- a/mooncake-transfer-engine/tent/include/tent/transfer_engine.h
+++ b/mooncake-transfer-engine/tent/include/tent/transfer_engine.h
@@ -142,6 +142,12 @@ int tent_open_segment(tent_engine_t engine, tent_segment_id_t* handle,
 
 int tent_close_segment(tent_engine_t engine, tent_segment_id_t handle);
 
+// Eagerly establish connection state towards `handle` on every installed
+// transport that supports it (RDMA endpoint bootstrap, control-plane TCP
+// dial), so the first transfer does not pay the setup cost. Best effort and
+// idempotent; returns 0 when there is nothing to warm. Returns 0 on success.
+int tent_warmup_segment(tent_engine_t engine, tent_segment_id_t handle);
+
 int tent_get_segment_info(tent_engine_t engine, tent_segment_id_t handle,
                           tent_segment_info_t* info);
 
@@ -281,6 +287,13 @@ class TransferEngine {
     Status closeSegment(SegmentID handle);
 
     Status getSegmentInfo(SegmentID handle, SegmentInfo& info);
+
+    // Establishes connection state towards a segment ahead of the first
+    // transfer, so the setup cost is paid where the caller chooses instead of
+    // inside the first request. Best effort and idempotent: a transport with
+    // nothing to warm, including the local segment, succeeds as a no-op, and
+    // a pair that cannot connect yet keeps the lazy path.
+    Status warmupSegment(SegmentID handle);
 
    public:
     Status allocateLocalMemory(void** addr, size_t size,
--- a/mooncake-transfer-engine/tent/include/tent/transfer_engine.h
+++ b/mooncake-transfer-engine/tent/include/tent/transfer_engine.h
@@ -143,9 +143,12 @@ int tent_open_segment(tent_engine_t engine, tent_segment_id_t* handle,
 int tent_close_segment(tent_engine_t engine, tent_segment_id_t handle);
 
 // Eagerly establish connection state towards `handle` on every installed
-// transport that supports it (RDMA endpoint bootstrap, control-plane TCP
-// dial), so the first transfer does not pay the setup cost. Best effort and
-// idempotent; returns 0 when there is nothing to warm. Returns 0 on success.
+// transport that can warm one, so the first transfer does not pay the setup
+// cost. Only RDMA warms today: it brings up every enabled local context
+// against every RDMA NIC the target advertises. TCP is not warmed.
+// Idempotent. Returns 0 when every transport either warmed every pair it
+// attempted or had nothing to warm, and -1 when one of them left a connection
+// cold - a cold pair keeps its lazy path, so the target stays usable.
 int tent_warmup_segment(tent_engine_t engine, tent_segment_id_t handle);
 
 int tent_get_segment_info(tent_engine_t engine, tent_segment_id_t handle,
@@ -290,9 +293,11 @@ class TransferEngine {
 
     // Establishes connection state towards a segment ahead of the first
     // transfer, so the setup cost is paid where the caller chooses instead of
-    // inside the first request. Best effort and idempotent: a transport with
-    // nothing to warm, including the local segment, succeeds as a no-op, and
-    // a pair that cannot connect yet keeps the lazy path.
+    // inside the first request. Idempotent. A transport with nothing to warm,
+    // including the local segment, succeeds as a no-op; a transport that left
+    // any connection it attempted cold reports the first such failure, even
+    // when another transport succeeded. A cold pair keeps its lazy path, so
+    // the target stays usable either way.
     Status warmupSegment(SegmentID handle);
 
    public:
--- a/mooncake-transfer-engine/tent/src/transport/rdma/rdma_transport.cpp
+++ b/mooncake-transfer-engine/tent/src/transport/rdma/rdma_transport.cpp
@@ -765,6 +765,89 @@ std::shared_ptr<RdmaEndPoint> RdmaTransport::getEndpoint(SegmentID target_id,
     return endpoint;
 }
 
+Status RdmaTransport::warmupSegment(SegmentID target_id) {
+    std::string rpc_server_addr, target_seg_name, target_nic_path_name;
+    std::vector<std::string> target_dev_names;
+
+    auto status = metadata_->segmentManager().withCachedSegment(
+        target_id, [&](SegmentDesc* segment) {
+            // withCachedSegment reruns this on NeedsRefreshCache, so start
+            // from a clean list rather than appending to the first pass.
+            target_dev_names.clear();
+            if (segment->type != SegmentType::Memory) {
+                // File and other non-memory segments carry no RDMA endpoint
+                // state. Nothing to warm is not a failure.
+                return Status::NotImplemented(
+                    "Segment type is not Memory" LOC_MARK);
+            }
+            if (target_id != LOCAL_SEGMENT_ID) {
+                rpc_server_addr = segment->rpc_server_addr;
+            }
+            auto topo = &std::get<MemorySegmentDesc>(segment->detail).topology;
+            target_seg_name = segment->name;
+            target_nic_path_name = segment->nicPathServerName();
+            for (size_t id = 0; id < topo->getNicCount(); ++id) {
+                if (topo->getNicType(id) != Topology::NIC_RDMA) continue;
+                auto dev_name = topo->getNicName(id);
+                if (!dev_name.empty()) target_dev_names.push_back(dev_name);
+            }
+            if (target_seg_name.empty()) {
+                return Status::NeedsRefreshCache(
+                    "Empty target segment name" LOC_MARK);
+            }
+            if (target_dev_names.empty()) {
+                // A TCP-only or CPU-only peer advertises no RDMA NIC. There
+                // is nothing to warm towards it, which is not a failure -
+                // and it must not look like a stale cache, or every warmup
+                // of such a peer would refetch the whole segment desc.
+                return Status::NotImplemented(
+                    "Target advertises no RDMA device" LOC_MARK);
+            }
+            return Status::OK();
+        });
+    if (!status.ok()) return status;
+
+    // All enabled local contexts x all remote RDMA NICs: the data plane may
+    // route a slice over any of these pairs, and bootstrapping also brings
+    // up the notification QP that rides on the same endpoint.
+    size_t attempted = 0, ready = 0;
+    Status last_error = Status::OK();
+    for (auto& ctx : context_set_) {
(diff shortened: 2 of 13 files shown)


staryxchen (commented):
Could the application warm up by issuing a meaningless request, rather than having TENT provide a dedicated API?

xiaodouzi666:
It can, and benchmarks do that today, but it costs the caller two things the primitive doesn't. So from my perspective, a dummy transfer has to target a registered remote address the caller is allowed to clobber, so the application reserves scratch space on the peer purely for warming. warmupSegment() touches no memory, and it warms one NIC pair, not the segment, and which pair is up to the transport selector, and the cost scales per pair, that's about 45 ms for one pair on the hardware in the description, in my case, around a second when a whole set has to come up. Warming all of them with dummy transfers means steering the selector, which the public API doesn't let callers do.

staryxchen:
> It can, and benchmarks do that today, but it costs the caller two things the primitive doesn't. So from my perspective, a dummy transfer has to target a registered remote address the caller is allowed to clobber, so the application reserves scratch space on the peer purely for warming. warmupSegment() touches no memory, and it warms one NIC pair, not the segment, and which pair is up to the transport selector, and the cost scales per pair, that's about 45 ms for one pair on the hardware in the description, in my case, around a second when a whole set has to come up. Warming all of them with dummy transfers means steering the selector, which the public API doesn't let callers do.

@xiaodouzi666 Thanks — that make sense to me. A dummy transfer only warms whichever NIC pair the selector happens to pick, needs a clobberable remote buffer, and can't enumerate the pair set the public API doesn't expose. Paying the setup cost as an engine-level warmupSegment(), with each transport implementing (or declining) it, is the right split.

The design looks good. I'll review the implementation in a follow-up.

catyans:
The engine-level aggregation contract is sound, but the RDMA implementation hides partial warmup failures. `RdmaTransport::warmupSegment()` enumerates every enabled local context x every advertised remote RDMA NIC and records `last_error`, but it returns that error only when `ready == 0`. If one pair is ready and the remaining pairs fail, RDMA returns `OK`, so `TransferEngine::warmupSegment()` also reports success.

That conflicts with the PR's documented contract that a genuine transport failure is reported rather than handing back a target whose first transfer can still stall. The data path may later select any of the failed pairs, so even a 1/N success can produce exactly that outcome. It is also inconsistent with the top-level test where one successful transport must not mask another transport's failure.

Could RDMA return an error whenever `ready != attempted` (preferably preserving the first pair failure, like the engine-level aggregation), or explicitly redefine the API as requiring only minimum connectivity and expose the partial result some other way? Since the RDMA loop is not covered by the new unit tests, a small test seam with mixed endpoint outcomes—one already ready or successfully connected, one failing—would verify that partial warmup cannot be silently reported as complete.

mooncake-transfer-engine/tent/src/transport/rdma/rdma_transport.cpp
```diff
@@ -765,6 +765,89 @@ std::shared_ptr<RdmaEndPoint> RdmaTransport::getEndpoint(SegmentID target_id,
...
+        }
+    }
+    if (attempted == 0) {
+        // No enabled local context: nothing to warm from, same as above.
+        return Status::NotImplemented("No enabled RDMA context" LOC_MARK);
+    }
+    LOG(INFO) << "RDMA warmup towards segment " << target_seg_name << ": "
+              << ready << "/" << attempted << " endpoints ready";
+    if (ready == 0) return last_error;
+    return Status::OK();
```
staryxchen: The engine contract is that a genuine warmup failure is reported even if another transport succeeded — returning OK would leave the first real transfer still stalled. The same hole exists inside RDMA: ready > 0 returns OK when some pairs never connected, so a later slice on a cold pair still pays the 40–50 ms bootstrap.

EFA treats any failed pair as `ERR_ENDPOINT`. If a pair cannot come up it can stay on the lazy path, but warmupSegment() should not return OK unless `ready == attempted`.

mooncake-transfer-engine/tent/src/transport/rdma/rdma_transport.cpp
```diff
@@ -765,6 +765,89 @@ std::shared_ptr<RdmaEndPoint> RdmaTransport::getEndpoint(SegmentID target_id,
...
+                last_error = Status::InternalError("Cannot allocate endpoint " +
+                                                   peer_name + LOC_MARK);
+                continue;
+            }
+            if (endpoint->status() == RdmaEndPoint::EP_READY) {
+                ++ready;
+                continue;
+            }
+            auto connect_status =
+                endpoint->connect(target_seg_name, dev_name, rpc_server_addr);
```
staryxchen: These connect() calls run sequentially on the caller’s thread, so wall time is O(local × remote) pair RTTs (~45 ms each here; ~1 s for 5×5, ~11 s for 16×16). Classic EfaTransport::warmupSegment fans the same handshake out with std::async and finishes 256 pairs in ~1.1 s.

For the callers this API is aimed at, a serial full-mesh warmup just moves the stall into warmupSegment(). Overlapping the pair connects would keep the cost closer to one handshake RTT.

mooncake-transfer-engine/tent/tests/CMakeLists.txt
```diff
@@ -451,6 +432,18 @@ target_include_directories(tent_rpc_handler_isolation_test
 add_test(NAME tent_rpc_handler_isolation_test
          COMMAND tent_rpc_handler_isolation_test)
 
+# warmupSegment routing/aggregation semantics plus a real TCP warmup between
+# two loopback engines. RDMA warming needs a NIC; not exercised here.
```
staryxchen: The comment says this test covers “a real TCP warmup between two loopback engines”. warmup_segment_test.cpp only drives fake transports through the engine routing/aggregation contract — there is no TCP loopback, and the RDMA pair path is not exercised either.

Please drop the TCP claim (or add that test). As written it will look like TCP warmup is already covered, which this PR explicitly left out.

mooncake-transfer-engine/tent/include/tent/transfer_engine.h
```diff
@@ -142,6 +142,12 @@ int tent_open_segment(tent_engine_t engine, tent_segment_id_t* handle,
 
 int tent_close_segment(tent_engine_t engine, tent_segment_id_t handle);
 
+// Eagerly establish connection state towards `handle` on every installed
+// transport that supports it (RDMA endpoint bootstrap, control-plane TCP
+// dial), so the first transfer does not pay the setup cost. Best effort and
+// idempotent; returns 0 when there is nothing to warm. Returns 0 on success.
+int tent_warmup_segment(tent_engine_t engine, tent_segment_id_t handle);
```
staryxchen, line 152: The comments and the implementation disagree on the contract.

The C header mentions a control-plane TCP dial, but this PR does not warm TCP. Headers call the call best-effort, while the engine reports the first real transport failure even if another transport succeeded, and RDMA still returns OK when only some pairs connected. Please make the public comments match one rule: what “success” means across transports and across RDMA pairs, and drop the TCP dial.

The try/catch around this C wrapper is also unique. If bootstrap can throw, tent_submit takes the same path and is unguarded — either catch it everywhere it can fire, or don’t imply that only warmup needs it.

staryxchen (commented):
C++ and tent_warmup_segment() are in this PR, but the Python binding is not. TENT already exposes open_segment here, and classic TE has warmup_efa_segment. Python callers still have to fake a transfer.

Please add warmup_segment next to open_segment, or say explicitly that the Python surface is follow-up.

xiaodouzi666:
Fixed in e909a17. A pass now reports the first pair failure it saw unless every
pair it attempted came back ready, I took your first option rather than
redefining the API as minimum-connectivity

catyans (approved):
Thanks, this addresses my concern. LGTM. RDMA warmup now succeeds only when every attempted endpoint pair is ready, preserves the first pair failure, and covers complete, partial, empty, and defensive outcomes in focused tests.

suggestion (non-blocking): @staryxchen's Python-surface point remains open; please either expose `warmup_segment` next to `open_segment` or document the Python binding as a follow-up. The current wheel failure is an unrelated Mooncake PG extension import mismatch; the other 23 checks pass.

mooncake-transfer-engine/tent/src/transport/rdma/rdma_transport.cpp
xiaodouzi666: Done.

mooncake-transfer-engine/tent/tests/CMakeLists.txt
xiaodouzi666: Done.

mooncake-transfer-engine/tent/include/tent/transfer_engine.h
xiaodouzi666: Done.

mooncake-transfer-engine/tent/src/transport/rdma/rdma_transport.cpp
xiaodouzi666: Done.

xiaodouzi666:
Done.

xiaodouzi666:
Merged current main into the branch to pick up #4106; the wheel job's failure was `test_pg_init_functional…test_destroy_and_reinit`, which that PR fixes, and the tent tests are unchanged on the new base, so far so good, ready for another look.

PR #3687, opened 2026-08-26
[Bugfix][Store] Cover HotStandbyService behaviors with deterministic fakes

### LINKED ISSUE #3496: [Bug]: HotStandbyService tests do not exercise claimed etcd behaviors

huangxiaobai-ydmy:
## Description

Fixes #3496.

Several `HotStandbyService` tests were named after etcd-backed standby behaviors but unconditionally `GTEST_SKIP`'d, so CI did not exercise the claimed Start / sync / warm-start / replication / verification paths.

This PR:

1. Lets `SetCatchUpBatchKvBackendForTesting` drive the real `Start()` / replication / promotion catch-up path without a live etcd connection (also works when `STORE_USE_ETCD` is OFF).
2. Replaces the #3496 skip gaps with deterministic fake-backend coverage for Start, state transitions, sync status, warm start, replication metrics/failures, and verification enable/disable.
3. Keeps real-etcd invalid-endpoint / `CONNECTION_FAILED` coverage under `STORE_USE_ETCD`, and names the non-etcd feature-gate path separately.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

Added/rewrote unit coverage in `hot_standby_service_test.cpp` using injected `HaKvBackend` fakes.

Full build/tests not run locally (Windows environment without Mooncake build toolchain); please rely on CI.

**Test commands:**
```bash
# CI: Build & Test (Linux)
# Local (Linux): ctest --test-dir build -R hot_standby_service_test
```

**Test results:**
- [ ] Unit tests pass
- [x] Manual testing done (code review; unit tests added/updated)

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

Cursor AI assisted with issue analysis, test-seam adjustment, unit coverage for #3496, and PR drafting. The submitter reviewed the changes.

### DIFF
--- a/mooncake-common/etcd/etcd_wrapper.go
+++ b/mooncake-common/etcd/etcd_wrapper.go
@@ -327,9 +327,6 @@ func EtcdStoreResetClientWrapper(endpoints *C.char, errMsg **C.char) int {
 	closeAllStoreMaintenanceSessions()
 	cancelAllStoreWatches()
 	cancelAllStorePrefixWatches()
-	// Join prefix-watch goroutines before closing the old client so reset is
-	// closer to fixture-style teardown (cancel + wait + close).
-	waitAllStorePrefixWatches(5 * time.Second)
 
 	storeMutex.Lock()
 	oldClient := storeClient
@@ -636,34 +633,6 @@ func cancelAllStorePrefixWatches() {
 
 	for _, cancel := range cancels {
 		cancel()
-	}
-}
-
-// waitAllStorePrefixWatches waits for prefix-watch goroutines previously
-// cancelled by cancelAllStorePrefixWatches to fully exit (done closed).
-// Returns early if the timeout elapses; callers should still Close the old
-// client afterward so residual RPCs are aborted.
-func waitAllStorePrefixWatches(timeout time.Duration) {
-	storePrefixWatchMutex.Lock()
-	dones := make([]<-chan struct{}, 0, len(storePrefixWatchCtx))
-	for _, watchInfo := range storePrefixWatchCtx {
-		if watchInfo.done != nil {
-			dones = append(dones, watchInfo.done)
-		}
-	}
-	storePrefixWatchMutex.Unlock()
-
-	deadline := time.Now().Add(timeout)
-	for _, done := range dones {
-		remaining := time.Until(deadline)
-		if remaining <= 0 {
-			return
-		}
-		select {
-		case <-done:
-		case <-time.After(remaining):
-			return
-		}
 	}
 }
 
--- a/mooncake-common/etcd/etcd_wrapper.go
+++ b/mooncake-common/etcd/etcd_wrapper.go
@@ -327,6 +327,9 @@ func EtcdStoreResetClientWrapper(endpoints *C.char, errMsg **C.char) int {
 	closeAllStoreMaintenanceSessions()
 	cancelAllStoreWatches()
 	cancelAllStorePrefixWatches()
+	// Join prefix-watch goroutines before closing the old client so reset is
+	// closer to fixture-style teardown (cancel + wait + close).
+	waitAllStorePrefixWatches(5 * time.Second)
 
 	storeMutex.Lock()
 	oldClient := storeClient
@@ -633,6 +636,34 @@ func cancelAllStorePrefixWatches() {
 
 	for _, cancel := range cancels {
 		cancel()
+	}
+}
+
+// waitAllStorePrefixWatches waits for prefix-watch goroutines previously
+// cancelled by cancelAllStorePrefixWatches to fully exit (done closed).
+// Returns early if the timeout elapses; callers should still Close the old
+// client afterward so residual RPCs are aborted.
+func waitAllStorePrefixWatches(timeout time.Duration) {
+	storePrefixWatchMutex.Lock()
+	dones := make([]<-chan struct{}, 0, len(storePrefixWatchCtx))
+	for _, watchInfo := range storePrefixWatchCtx {
+		if watchInfo.done != nil {
+			dones = append(dones, watchInfo.done)
+		}
+	}
+	storePrefixWatchMutex.Unlock()
+
+	deadline := time.Now().Add(timeout)
+	for _, done := range dones {
+		remaining := time.Until(deadline)
+		if remaining <= 0 {
+			return
+		}
+		select {
+		case <-done:
+		case <-time.After(remaining):
+			return
+		}
 	}
 }
 
--- a/mooncake-store/include/hot_standby_service.h
+++ b/mooncake-store/include/hot_standby_service.h
@@ -207,8 +207,9 @@ class HotStandbyService {
     void SetSyncStatusCallback(SyncStatusCallback callback);
 
     /**
-     * @brief Test seam: when set, promotion final catch-up first tries
-     *        batch-record durable prefix/batches from this backend.
+     * @brief Test seam: inject a HaKvBackend used for Start()/replication and
+     *        promotion catch-up. When set, Start() skips live etcd connection
+     *        (and works even when STORE_USE_ETCD is OFF).
      */
     void SetCatchUpBatchKvBackendForTesting(
         std::shared_ptr<HaKvBackend> backend);
--- a/mooncake-store/src/hot_standby_service.cpp
+++ b/mooncake-store/src/hot_standby_service.cpp
@@ -7,7 +7,9 @@
 #include <thread>
 
 #include "etcd_helper.h"
+#ifdef STORE_USE_ETCD
 #include "ha/kv/etcd_ha_kv_backend.h"
+#endif
 #include "ha_metric_manager.h"
 #include "ha/oplog/oplog_applier.h"
 #include "ha/oplog/oplog_batch_standby_reader.h"
@@ -109,8 +111,10 @@ ErrorCode HotStandbyService::Start(const std::string& primary_address,
     cluster_id_ = cluster_id;
 
     if (config_.enable_oplog_following) {
-#ifdef STORE_USE_ETCD
+        // Injected test backends exercise the production Start/replication
+        // path without a live etcd cluster (and without STORE_USE_ETCD).
         if (!catch_up_batch_kv_backend_for_testing_) {
+#ifdef STORE_USE_ETCD
             ErrorCode err =
                 EtcdHelper::ConnectToEtcdStoreClient(oplog_endpoints.c_str());
             if (err != ErrorCode::OK) {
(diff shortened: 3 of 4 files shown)


Icedcoco:
Thanks for adding the deterministic backend coverage. I built and ran the HotStandbyService tests locally in both configurations.

With STORE_USE_ETCD=OFF, the target compiled successfully and all 45 tests completed with 43 passing and 2 expected skips.

With STORE_USE_ETCD=ON, the target also compiled, but the full test suite reported 42 passing, 2 skips, and 2 failures. Before merging, could we revisit the following points?

1. Could we revisit TestStart_InvalidEtcdEndpoints and TestStateTransition_ConnectionFailed?

    In my local ETCD-enabled run, both tests use http://127.0.0.1:1, but Start() returned OK and the service entered WATCHING. The assertions expecting a non-OK return and FAILED state therefore failed.

    From mooncake-common/etcd/etcd_wrapper.go, NewStoreEtcdClient() only validates that the endpoint list is non-empty and creates the etcd client. It does not establish a connection immediately. The actual dial failure appeared later from the replication loop, after the backend Get() timed out.

    Would it make sense to use a syntactically invalid or empty endpoint for synchronous endpoint-validation coverage, and use a separate integration fixture or asynchronous assertion for connection failure? The global etcd client state may also need to be reset between these tests, otherwise test order can affect the result.

2. Could we rework or rename TestStateTransition_SyncFailed?

    The current test constructs a snapshot with duplicate metadata and fails during snapshot restoration. The log shows Snapshot baseline contains duplicate object. This happens before the batch reader or replication loop starts, so the test does not currently exercise a synchronization failure. It also overlaps with the existing SnapshotRestoreRejectsDuplicateObject test.

    Would injecting an unreadable or malformed batch after a successful start be a better way to cover the intended sync-failure path?

3. Could we clarify the intended coverage of the verification-loop tests?

    The production code still states that verification is not implemented and currently only emits a verbose log. The enabled and disabled tests only assert that the service remains in WATCHING and can be stopped, so they would pass even if the verification thread were never started or exited immediately.

    Should these remain skipped until verification has an observable effect, or should they be renamed as thread-lifecycle tests with a more specific assertion?

4. One lifecycle concern about TestReplicationLoop_UpdatesMetrics:

    The callback captures observed_applied by reference using [&]. That variable is local to the test body, while the service remains alive until fixture teardown. After the test body returns, TearDown() calls service_->Stop(), and the stop transition can notify the callback again after the local variable's lifetime has ended.

    I did not see this fail in the ASAN run, but the lifetime is still unsafe. Could we explicitly stop the service before leaving the test body, or store the observed state in an object whose lifetime extends through teardown?

5. Could the deterministic fake-backend cases avoid conditional skips?

    In UsesDurablePrefixLastSeqAsCatchUpTarget and PaginatesBatchRecords, Start() failures are converted into GTEST_SKIP(). Since these tests use injected fake backends and do not depend on an external service, would ASSERT_EQ(ErrorCode::OK, err) be more appropriate? Otherwise a regression in the fake setup or startup path could be reported as a skip rather than a test failure.

The injected backend seam itself looks useful, and the majority of the fake-backed tests passed locally. I would be comfortable revisiting the PR after the ETCD-enabled failures and the coverage/lifetime questions above are resolved.

huangxiaobai-ydmy:
Thanks for the detailed review, @Icedcoco — addressed all five points on this branch (rebased onto latest `main`):

1. **`TestStart_InvalidEtcdEndpoints` / `TestStateTransition_ConnectionFailed`**
   - Invalid endpoints now use an **empty** endpoint string so `NewStoreEtcdClient` fails synchronously (no dial).
   - Connection-failed coverage uses an **injected fake backend** that fails `Get` after `Start()` returns OK, then `WaitForState(FAILED)`. Live `127.0.0.1:1` dial latency is no longer asserted.

2. **`TestStateTransition_SyncFailed`**
   - Reworked to start successfully with a valid batch, then publish a **malformed** batch record and advance the durable prefix so the **replication loop** (not snapshot restore) drives `FAILED`. Duplicate-snapshot coverage remains in `SnapshotRestoreRejectsDuplicateObject`.

3. **Verification tests**
   - Renamed to thread-lifecycle tests and documented that production verification is still unimplemented; they only assert start/stop while enabled/disabled.

4. **`TestReplicationLoop_UpdatesMetrics`**
   - Callback now captures a `shared_ptr<atomic>`, the test explicitly `Stop()`s and clears the callback before leaving the body.

5. **`UsesDurablePrefixLastSeqAsCatchUpTarget` / `PaginatesBatchRecords`**
   - Replaced `GTEST_SKIP()` on `Start()` failure with `ASSERT_EQ(ErrorCode::OK, ...)`.

Happy to adjust further if anything still looks off.

huangxiaobai-ydmy:
Follow-up on points 1 and 3 to match the review more closely:

1. **`TestStateTransition_ConnectionFailed`** now uses the **live etcd path**:
   - `ResetEtcdStoreClient(http://127.0.0.1:1)` to isolate global client state
   - `Start()` is expected to return `OK` (client object exists; no sync dial)
   - `WaitForState(FAILED)` with a longer timeout for the async Get/dial failure
   - restore `http://127.0.0.1:2379` afterward so later etcd tests are not stuck
   - `TestStart_InvalidEtcdEndpoints` remains the synchronous empty-endpoint case

3. **Verification tests** are now `GTEST_SKIP` until checksum verification has an observable effect (as suggested), instead of weak lifecycle-only assertions.

huangxiaobai-ydmy:
Follow-up on the **independent integration fixture** concern for `TestStateTransition_ConnectionFailed`:

True per-process client isolation is not available here: HotStandby's production path can only use the process-global store etcd client (Go singleton via cgo), and forking after the Go runtime is multi-threaded is unsafe.

What we did instead (strongest isolation the current API allows):

1. **`LiveUnreachableEtcdIntegrationFixture`** — named RAII fixture that exclusively owns the global store client for the test:
   - `TryCreate` → `ResetEtcdStoreClient(127.0.0.1:1)`
   - owns the `HotStandbyService` under test
   - `Release` / destructor always `Stop()`s the service **before** restoring `127.0.0.1:2379`

2. **`EtcdStoreResetClientWrapper`** — after canceling keepalives/watches/prefix-watches, now **waits** (up to 5s) for prefix-watch goroutines to exit before swapping/closing the old client, so Reset is closer to fixture-style teardown (cancel + join + close).

Happy to adjust further if you'd prefer a different shape.

Icedcoco:
Sorry for not following up on this PR earlier, and thank you for your patience.

I reviewed the latest changes and wanted to discuss a few remaining concerns before moving forward:

1. LiveUnreachableEtcdIntegrationFixture restores the global etcd client to the hard-coded endpoint http://127.0.0.1:2379. Could this affect tests or components that use a different configured endpoint? Would it be possible to preserve and restore the original endpoint instead?

2. EtcdStoreResetClient() now waits up to five seconds for prefix-watch goroutines to exit. Since this is a shared production API, could the additional blocking affect leader changes or recovery paths? Would it be safer to keep this lifecycle handling within an isolated integration test?

3. The unreachable-etcd test still depends on a process-global client and network timeout behavior, so it may remain slow or order-sensitive. Would an injected backend that fails asynchronously provide sufficiently deterministic coverage, or should this case be moved to a separate integration test?

The fake-backend coverage and the lifecycle fix for the metrics callback look useful. I’m mainly hoping we can resolve the global etcd client isolation and reset behavior before merging. Apologies again for the delayed follow-up, and I’d appreciate your thoughts on these points.

huangxiaobai-ydmy:
Thanks for the follow-up, @Icedcoco — agreed on all three points. Updated on the latest push:

1. **Endpoint restore / live fixture**
   Removed `LiveUnreachableEtcdIntegrationFixture` entirely, so we no longer Reset the process-global store client or hard-code a restore endpoint.

2. **`EtcdStoreResetClient` blocking wait**
   Reverted the production `waitAllStorePrefixWatches(5s)` call from `EtcdStoreResetClientWrapper`. Reset is back to cancel + swap + close, with no extra join in the shared path.

3. **ConnectionFailed coverage**
   `TestStateTransition_ConnectionFailed` now uses an injected fake backend:
   - `Start()` succeeds and reaches `WATCHING` with a valid batch
   - then `SetGetError(ETCD_OPERATION_ERROR)` so a later replication-loop `Get` drives `FAILED`
   - no live dial / network timeout / global-client ownership

Happy to adjust further if you'd like a different shape for the async failure.

huangxiaobai-ydmy:
Rebased onto the latest `main` (`b905925`) to clear merge conflicts.

Resolved one conflict in `FinalCatchUpForPromotionLocked` by keeping main's `PromotionCatchUpPolicy` argument together with the existing `STORE_USE_ETCD` guard.

huangxiaobai-ydmy:
Hi @Icedcoco @Libotry @YiXR @00fish0 @Aionw — gentle ping when you have a moment.

CI is green again after the rebase onto latest `main`. The earlier review points are addressed on the latest push (removed the live-unreachable fixture, kept production `EtcdStoreResetClientWrapper` unchanged, and ConnectionFailed now uses the injected async failure path). If the direction looks good, would you mind taking another look for a CODEOWNER review when convenient? No rush, and happy to address any further feedback. Thanks!

Icedcoco (approved):
LGTM

huangxiaobai-ydmy:
Thanks for the approval, @Icedcoco — really appreciate the careful review and the earlier feedback that helped tighten the tests.

Happy to address any further comments if anything else comes up.

ykwd (approved):

PR #1633, opened 2026-03-08
[TENT] feat: add tcp_hp transport for High-Concurrency Throughput

00fish0:
## Description

This PR reworks the TCP transport in TENT to use a dedicated ASIO-based data plane instead of sending payloads through the control-plane RPC path.

The existing TENT TCP transport routes payload through control-plane RPC helpers (`ControlClient::sendData/recvData`), which adds overhead in the hot path: extra RPC framing, per-transfer connection setup, and tight coupling between control and data planes.

`tcp_hp` is a standalone transport that performs direct socket I/O, bypassing the control plane entirely.

**Changes:**

- **Dedicated data server**: `TcpHpTransport` starts its own TCP acceptor and publishes the data port via `transport_attrs[TCP_HP]`
  - **Lightweight session protocol**: fixed header (`opcode + addr + length`) for async read/write — no RPC framing overhead
  - **Connection pooling**: client-side pool keyed by `host:port`, connections are reused across transfers
  - **Bounce-buffer staging**: pre-allocated bounce buffers for chunked transfers, supporting both CUDA and CANN memory
  - **Inflight control**: configurable max in-flight transfers to avoid overwhelming the receiver
  - **IoContextPool**: dedicated ASIO thread pool for data-plane I/O, decoupled from the control plane

## Module

- [x] Transfer Engine (`mooncake-transfer-engine`)
- [ ] Mooncake Store (`mooncake-store`)
- [ ] Mooncake EP (`mooncake-ep`)
- [ ] Integration (`mooncake-integration`)
- [ ] P2P Store (`mooncake-p2p-store`)
- [ ] Python Wheel (`mooncake-wheel`)
- [ ] PyTorch Backend (`mooncake-pg`)
- [ ] Mooncake RL (`mooncake-rl`)
- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other

## How Has This Been Tested?

```
 ./mooncake-transfer-engine/benchmark/tebench  --xport_type=tcp --seg_type=DRAM --metadata_type=p2p
```

 ## Benchmark

  Tested on 2x H20 nodes connected via 25GbE (eth0), using `tebench --seg_type=DRAM --metadata_type=p2p`:

  | Block Size | Classic TE TCP (GB/s) | TENT tcp_hp (GB/s) | Speedup |
  |---|---|---|---|
[...]

PR #3689, opened 2026-08-26
[TENT]Add correctness-first high-performance TCP foundation

jacklin78911-collab:
## Description

This draft PR adds `hp_tcp` as a standalone, optional high-performance TCP transport for TENT.

The existing `tcp` transport remains unchanged and continues to use `coro_rpc`. The new transport keeps RPC on the control plane while using an independent asynchronous TCP data plane for bulk data movement.

### What is included

- A standalone `HP_TCP` transport type, configuration block, loader path, metadata schema, and build target.
- Persistent per-peer TCP lanes with deterministic worker affinity and connection reuse.
- Per-worker private ASIO I/O contexts and bounded worker mailboxes.
- Atomic admission with hard task, byte, and mailbox bounds.
- An asynchronous client/server data path with connect and progress deadlines.
- A bounded server session executor with completed-session reaping.
- Deterministic shutdown that cancels accepted sockets and waits for active operations to quiesce.
- A fixed binary protocol with network-byte-order encoding and request ID validation.
- READ completion and remote commit acknowledgement for WRITE operations.
- Buffer registration IDs, permission validation, lease-aware unregister, and batch-registration rollback.
- Operation epochs and peer incarnation tracking.
- End-to-end stale-registration recovery:
  - validated remote wire statuses are preserved;
  - stale endpoint and registration metadata is invalidated;
  - the operation is retried at most once within `hp_tcp`;
  - permission and range failures remain terminal;
  - shutdown remains transient and may use normal transport failover.
- Explicit CPU-DRAM-only validation for the v1 implementation.

### Correctness and ownership model

The implementation keeps the following resources bounded:

- admitted tasks and bytes;
- per-worker mailbox entries;
- persistent client lanes;
- accepted server sessions;
- buffer registrations and active leases.

A task publishes its terminal status only after its remote status and completion reason have been stored. Shutdown, timeout, cancellation, and stale-registration recovery retain a single terminal-completion owner.

The detailed ownership, protocol, timeout, buffer-lifetime, and shutdown semantics are documented in the `hp_tcp` design document included in this PR.

### Scope

The current v1 scope is:

- CPU DRAM;
- one advertised endpoint per peer;
- persistent TCP lanes;
- READ and WRITE data movement;
- correctness-first bounded resource management.

The following are intentionally left for separate follow-up work:

- GPU/CANN memory support;
- multi-endpoint and multi-rail routing;
- transfer striping;
- scatter/gather writes;
- additional socket tuning;
- QoS scheduling and cooperative transfer slicing.

The earlier TENT TCP prototype in #1633 remains useful prior work for performance mechanisms such as I/O context pools, bounce buffers, and striping. This PR focuses first on establishing explicit correctness, recovery, ownership, and resource-bound guarantees.

## Module

- [ ] CI/CD
- [x] Docs

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [x] Documentation update
- [x] Performance improvement
- [ ] Other

## How Has This Been Tested?

Local validation on the latest revision:

- Code format and scope:
  - `git diff --check`
  - clang-format 20 dry-run
- Functional tests:
  - related CTest suite: 11/11 passed
  - HP TCP two-process E2E: 5/5 passed
- ASan + UBSan:
  - related tests: 4/4 passed
  - HP TCP E2E: 5/5 passed
  - no sanitizer or leak reports
- TSan:
  - socket tests: 10/10 passed
  - transport tests: 6/6 passed
  - runtime and failover tests: 22/22 passed
  - no data-race reports

The tests cover:

- persistent-lane reuse and bounded session handling;
- connect, progress, zero-byte header, and partial-header timeouts;
- shutdown and socket cancellation;
- buffer permission and lease lifetime;
- atomic batch admission and registration rollback;
- stale registration replacement, metadata refresh, and bounded retry;
- permanent permission/range failures;
- transient shutdown fallback;
- rejection of non-CPU memory.

## Checklist

- [x] The changed files pass the repository's clang-format 20 rules
- [x] I have added tests to prove the changes are effective

## AI Assistance Disclosure

- [ ] No AI tools were used
- [x] AI tools were used

AI tools assisted with implementation review, test construction, and documentation. The resulting code was manually inspected and validated with functional, E2E, and sanitizer tests.


mooncake-transfer-engine/tent/include/tent/transport/hp_tcp/hp_tcp_buffer_registry.h
alogfans, line 1: We'd better to seprate these as `hp_tcp` transport.

mooncake-transfer-engine/tent/src/transport/hp_tcp/hp_tcp_server.cpp
alogfans, line 1: Comments for the current implementation:

- Each accepted connection creates a new std::thread and stores it in sessions_, but completed session threads are never joined or removed until stop(). Since the current client opens a new connection for every transfer, normal traffic will grow this vector and retain joinable thread resources for the entire transport lifetime. This can eventually exhaust memory or thread resources even when every transfer completes successfully.

- `stop()` closes only the acceptor and then joins every session thread. However, a session may be blocked indefinitely in a synchronous header or payload read, and its socket is not reachable from `stop()`. A peer that connects and sends a partial request can therefore prevent transport shutdown forever; after a lease has been acquired, it can also prevent the corresponding buffer from being unregistered.

- `connect_timeout_ms` is parsed, validated as non-zero, and passed to `HighPerformanceTcpClient`, while `progress_timeout_ms` is also exposed in the configuration. However, the transfer path uses blocking resolve/connect/read/write operations and neither timeout is applied. An unreachable or stalled peer can therefore leave a task pending, hold its local buffer lease, and block the assigned worker indefinitely.

Please also add a design document about hp_tcp.

mooncake-transfer-engine/tent/include/tent/transport/hp_tcp/hp_tcp_buffer_registry.h
jacklin78911-collab: Done. I split it out as a standalone `hp_tcp` transport in the latest update, with its own transport type, config block, loader path, metadata, and build target. Standard `tcp` remains unchanged.

mooncake-transfer-engine/tent/src/transport/hp_tcp/hp_tcp_server.cpp
jacklin78911-collab: Thanks for the detailed review. I reworked this part in the latest update:

- sessions now run on a bounded worker executor and are removed after completion;
- `stop()` cancels all accepted sockets and waits for active sessions to quiesce;
- client and server I/O now use asynchronous operations with connect and progress timeouts;
- the client reuses a persistent connection per lane.

I also added regression tests for session reaping, connection-budget reuse, progress timeout, and shutdown behavior. I’ll add an `hp_tcp` design document next to describe the ownership model, resource bounds, timeout/shutdown semantics, buffer leases, and protocol flow.

jacklin78911-collab: `hp_tcp` doc is ready~

catyans (changes requested):
Thanks for addressing the earlier lifecycle findings. The current head now explicitly settles pending mailbox commands, uses asynchronous I/O with connect/progress deadlines, bounds and reaps server sessions, and cancels accepted sockets during shutdown.

One recovery issue remains. The server now preserves specific wire statuses such as `kStaleRegistration`, `kPermissionDenied`, and `kShuttingDown`, but the client still passes every non-`kOk` response through `finishProtocolError()` and completes the task as a generic `FAILED`. The task state therefore loses the wire error before `getTransferStatus()` can expose it.

In particular, after a peer unregisters and re-registers a buffer, reconnecting the TCP lane continues to send the same stale registration ID from cached metadata. Because `kStaleRegistration` cannot trigger metadata invalidation and refresh, retries can repeatedly fail without recovering.

Please propagate the response status in a form the runtime can act on—mapping `kStaleRegistration` to `NeedsRefreshCache` or an equivalent refresh signal, while preserving permanent permission failures and transient shutdown errors. A regression test should replace a peer registration, observe the stale response, refresh metadata, and complete a bounded retry successfully.

I would request changes until the stale-registration recovery path is wired through end to end.

jacklin78911-collab:
@catyans Thanks for the review. Addressed in `88ac27bd`.

The client now preserves validated remote wire statuses, invalidates stale endpoint/registration metadata, and retries once within the same `hp_tcp` transport. Permission and range errors remain terminal and cannot fail over.

I also fixed several related correctness issues found during the review:

- the header progress timeout now covers zero-byte and partial-header stalls;
- failed batch registration rolls back entries created by that call;
- explicit HP TCP registration failures are returned to the caller;
- HP TCP v1 now explicitly rejects non-CPU memory.

The related functional, E2E, ASan/UBSan, and TSan tests pass. Could you please take another look when convenient?

jacklin78911-collab:
I pushed a follow-up that hardens the HP TCP failure lifecycle.

The main changes are:

- Admission accounting now fails closed on an underflow instead of resetting the counters and falsely reporting that all work has drained. Waiting shutdown threads are also notified when accounting enters the failed state.
- If an exception escapes an ASIO handler, the runtime rejects new submissions but keeps the owner event loop alive to retire already committed operations and process teardown cancellation with the original worker affinity. Shutdown joins the workers and reports the failure after tasks, sessions, and buffer leases have drained.
- Client, server, and admission teardown invariants are now checked explicitly.
- `tcp` and `hp_tcp` cannot be enabled together because they would compete for the single control-plane notification callback.
- Runtime health checks are lock-free on the normal path to avoid adding mutex contention to frequent status polling.
- I removed the unnecessary per-worker test state and kept a single sticky runtime failure state.

This does not change the wire format, connection reuse, `TCP_NODELAY`, or the normal socket data path.

Validation completed:

- clang-format 20 and `git diff --check`
- 7/7 core HP TCP tests
- two-process E2E
- 7/7 Debug + ASan tests with leak detection

jacklin78911-collab:
I pushed another correctness-focused update in `cb9fe311`.

This pass addresses the remaining HP TCP failure and lifecycle issues found during a fresh audit:

- Prevented ambiguous post-send WRITE failures from being retried or failed over, preserving the at-most-once remote-write invariant.
- Enforced progress timeouts for initial and partial headers/payloads while allowing pooled idle connections to remain reusable.
- Preserved protocol framing by draining the exact payload of a rejected but valid WRITE request.
- Fixed task-status publication and worker-health ordering so terminal results cannot be masked.
- Hardened buffer registration IDs and active leases against stale-registration ABA and unregister races.
- Corrected notification routing and avoided unsafe retries after ambiguous TCP/HP RPC failures.
- Restored the normal header fast path so a complete header can be consumed in one asynchronous read instead of a forced 1-byte-plus-remainder sequence.

Validation completed:

- clang-format 20 and `git diff --check`
- 11/11 targeted HP TCP, transport, configuration, selector, notification, and two-process E2E tests
- 11/11 tests under ASan/LSan with leak detection enabled
- 60/60 repeated socket/transport/E2E executions
- Documentation build completed without errors
- Two-node A/B performance check: 16 MiB × 16 READ was -0.99%, and WRITE was +2.14%, both within observed run-to-run variance

The wire format is unchanged, and this update does not introduce timed backpressure, HoL experiments, QoS, striping, socket tuning, or other future-performance mechanisms.

staryxchen:
Hi @jacklin78911-collab thanks for your contribution. Could you share some performance data comparing TCP and HP_TCP? You can use tebench to measure the data.

jacklin78911-collab:
@staryxchen Thanks for the suggestion. I ran a small tebench comparison on two H20 nodes connected through the same internal network, using DRAM buffers, batch size 1, and the current PR commit (`4649af9a`). Each point was repeated three times, and the table reports the median.

| Operation | TCP | HP_TCP | HP_TCP / TCP |
|---|---:|---:|---:|
| 4 KiB READ, 1 thread | 0.039 GB/s, 104 us | 0.026 GB/s, 157 us | 0.67x throughput |
| 4 KiB WRITE, 1 thread | 0.036 GB/s, 113 us | 0.026 GB/s, 158 us | 0.72x throughput |
| 1 MiB READ, 4 threads | 3.006 GB/s, 1388 us | 6.641 GB/s, 624 us | 2.21x throughput |
| 1 MiB WRITE, 4 threads | 2.919 GB/s, 1429 us | 7.526 GB/s, 550 us | 2.58x throughput |
| 16 MiB READ, 8 threads | 2.086 GB/s, 64.3 ms | 9.268 GB/s, 14.5 ms | 4.44x throughput |
| 16 MiB WRITE, 8 threads | 2.768 GB/s, 48.5 ms | 10.833 GB/s, 12.4 ms | 3.91x throughput |

HP_TCP provides a substantial improvement for medium and large transfers, reaching about 9.3 GB/s for READ and 10.8 GB/s for WRITE in this setup. It does introduce additional overhead for the 4 KiB single-thread case, so I do not claim that it improves every workload. The intended benefit is higher throughput through persistent connections, fixed lanes, and parallel I/O workers for larger transfers.

staryxchen:
> @staryxchen Thanks for the suggestion. I ran a small tebench comparison on two H20 nodes connected through the same internal network, using DRAM buffers, batch size 1, and the current PR commit (`4649af9a`). Each point was repeated three times, and the table reports the median.
> 
> Operation	TCP	HP_TCP	HP_TCP / TCP
> 4 KiB READ, 1 thread	0.039 GB/s, 104 us	0.026 GB/s, 157 us	0.67x throughput
> 4 KiB WRITE, 1 thread	0.036 GB/s, 113 us	0.026 GB/s, 158 us	0.72x throughput
> 1 MiB READ, 4 threads	3.006 GB/s, 1388 us	6.641 GB/s, 624 us	2.21x throughput
> 1 MiB WRITE, 4 threads	2.919 GB/s, 1429 us	7.526 GB/s, 550 us	2.58x throughput
> 16 MiB READ, 8 threads	2.086 GB/s, 64.3 ms	9.268 GB/s, 14.5 ms	4.44x throughput
> 16 MiB WRITE, 8 threads	2.768 GB/s, 48.5 ms	10.833 GB/s, 12.4 ms	3.91x throughput
> HP_TCP provides a substantial improvement for medium and large transfers, reaching about 9.3 GB/s for READ and 10.8 GB/s for WRITE in this setup. It does introduce additional overhead for the 4 KiB single-thread case, so I do not claim that it improves every workload. The intended benefit is higher throughput through persistent connections, fixed lanes, and parallel I/O workers for larger transfers.

Thanks for the numbers. The large-transfer gains are impressive. Did you also verify data correctness on this path? The functional/E2E suite shows the protocol and recovery cases pass, but I want to confirm the transferred payload itself. In particular, did you run tebench mix --check_consistency (or an equivalent full-buffer compare) over HP_TCP for both small and large sizes? If so, what was the result?

jacklin78911-collab:
@staryxchen Yes. I ran an additional HP_TCP payload-correctness check on the current PR commit (`4649af9a`) using tebench `mix --check_consistency`.

I covered both ends of the tested size range:

- 4 KiB, 1 thread
- 16 MiB, 8 threads

For each iteration, tebench filled the complete source buffer with a pattern, performed a WRITE followed by a READ, and compared the entire returned payload against that pattern. Both runs completed successfully with exit code 0, with no data mismatch, corruption, or transfer errors.

The earlier throughput numbers were collected without consistency checking to avoid including verification overhead; this additional run was specifically for end-to-end payload correctness.

jacklin78911-collab:
I pushed a small upstream-oriented cleanup in `64a9dd7a`: the internal endpoint representation is now truly single-endpoint, and the nonessential user-facing `chunk_size` knob was removed while preserving the metadata wire format and all data-path/lifecycle invariants. clang-format/diff checks, 10 targeted functional/E2E tests, 7 ASan/LSan tests, and the documentation build all pass.

catyans (approved):
Re-reviewed the current head. The earlier stale-registration blocker is resolved end to end: validated remote wire statuses are preserved, `kStaleRegistration` reaches the runtime as `NeedsRefreshCache`, cached metadata is invalidated, and the request is retried once on `hp_tcp`. Permission/range failures and ambiguous post-send WRITE failures remain terminal, so they cannot be replayed through failover.

I also checked the updated task lifecycle: completion is claimed once, buffer leases and admission reservations retire before terminal status publication, and shutdown/runtime-failure paths keep owner-affine callbacks alive long enough to settle committed work. The targeted recovery, socket, E2E, sanitizer, and payload-consistency coverage is convincing, and the current CI is green.

Non-blocking suggestion for a follow-up: combine the real socket stale-registration test and the runtime retry-orchestration test into one full `TransferEngineImpl` E2E case, so the metadata invalidation and bounded retry contract is covered without a fake transport seam.

jacklin78911-collab:
@catyans Thanks for the thorough re-review and approval. Agreed on the follow-up suggestion. I’ll keep this PR scoped to the reviewed HP_TCP v1 and track a separate full `TransferEngineImpl` E2E test covering real-socket stale registration, metadata invalidation, and bounded retry without the fake transport seam.

catyans (approved):
LGTM. The merge resolutions preserve HP_TCP registration and failover semantics while correctly integrating the new metrics accounting and explicit TCP build targets. All checks are green.

alogfans (approved):

catyans (approved):
LGTM. Re-reviewed the current head; the HP_TCP source split consistently updates the include paths, CMake targets, and test registrations without changing the reviewed data-path or lifecycle behavior.

The full CI is still running, so please wait for it to complete before merge.

PR #2199, opened 2026-05-23
[TE] Add ProgressWorker skeleton

ZhenyuePan:
## Summary
- add a default-off TENT ProgressWorker that drives `progressBatch()` from batch-ready notifications
- add engine-level batch lifetime protection so the worker can safely race with `freeBatch()`
- add opt-in `enable_progress_worker` config and a temporary `notifyBatchMaybeReady()` hook for tests/follow-up transport integration

## Context
Follow-up for #2116 after the explicit `progressBatch()` API. This PR only adds the worker skeleton and test hook; transport terminal-event integration and lazy-free behavior changes are intentionally left for follow-up PRs.

## Tests
- Docker (`mooncake-dev:latest`): built `tent_progress_worker_test` and `tent_engine_failover_e2e_test`
- Docker: `tent_progress_worker_test` (5/5 passed)
- Docker: `tent_engine_failover_e2e_test` (16/16 passed)
- Docker: `ctest -R 'tent_(progress_worker|engine_failover_e2e)_test' --output-on-failure` (2/2 passed)
- `git diff --check`

PR #3693, opened 2026-08-26
[TENT] Wake progress worker on RDMA task completion

zbchi:
## Description
  - [#2199](https://github.com/kvcache-ai/Mooncake/pull/2199) introduced `ProgressWorker` to advance batch progress and trigger failover without requiring continuous caller polling.
  - [#2562](https://github.com/kvcache-ai/Mooncake/pull/2562) added a progress  callback to `Transport::SubBatch` and wired the TCP, SHM and io_uring completion paths to invoke it. The callback wakes `ProgressWorker`, which also refills the local runtime queue dispatch window.
  - [#2733](https://github.com/kvcache-ai/Mooncake/pull/2733) reused the same callback for the TPU transport.

RDMA tasks did not retain or invoke this progress callback. Consequently, `ProgressWorker` could not react immediately when an RDMA task completed or  failed.

  This PR propagates the callback from `RdmaSubBatch` to each `RdmaTask` and  invokes it once when the task first reaches a terminal state.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?
The tests verify that a multi-slice task is not notified early, a terminal task is notified once, and repeated status updates do not generate duplicate  notifications.

**Test commands:**
  ```bash
  cmake -S . -B build
 cmake --build build  --target rdma_cancel_test tent_xport_rdma  tent_runtime_queue_dispatch_test tent_progress_worker_test
  ctest --test-dir build --output-on-failure    -R '^(rdma_cancel_test|tent_runtime_queue_dispatch_test|tent_progress_worker_test)$'

```

**Test results:**
- [x] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used
codex was used to update the unit tests.

### DIFF
--- a/mooncake-transfer-engine/tent/include/tent/transport/rdma/slice.h
+++ b/mooncake-transfer-engine/tent/include/tent/transport/rdma/slice.h
@@ -51,6 +51,8 @@ using RdmaTaskStorage = Slab<RdmaTask>;
 struct RdmaTask {
     int num_slices;
     Request request;
+    BatchID progress_batch_id{0};
+    std::function<void(BatchID)> notify_progress;
     // Resolved by TransportSelector and copied from RdmaSubBatch. The
     // per-slice path runs later on worker threads, so it must carry the same
     // device policy as the aggregate allocation path.
@@ -148,7 +150,11 @@ static inline void updateSliceStatus(RdmaSlice* slice,
                 ? COMPLETED
                 : task->first_error;
         if (final_st == PENDING) final_st = FAILED;
-        __sync_bool_compare_and_swap(&task->status_word, PENDING, final_st);
+        if (__sync_bool_compare_and_swap(&task->status_word, PENDING,
+                                         final_st)) {
+            if (task->notify_progress)
+                task->notify_progress(task->progress_batch_id);
+        }
     }
     task->deref();
 }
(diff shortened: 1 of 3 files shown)


catyans (changes requested):
Thanks for wiring RDMA terminal transitions into the ProgressWorker. The once-only notification in `updateSliceStatus()` is locally correct, but it exposes an early-completion race in the surrounding submit lifecycle.

An RDMA task may complete and invoke `notify_progress` while `commitPreparedSubmit()` is still submitting other transport groups. In addition, `submitTransfer()` installs the submit hook only after `commitPreparedSubmit()` returns. The ProgressWorker can therefore process the batch before all `sub_task_id` mappings and the submit hook are ready.

Because each RDMA task emits only one terminal notification, an early wake can be consumed before the hook exists. The hook may then never fire without subsequent caller polling, which conflicts with the event-driven progress goal of this change.

I suggest either installing the submit hook before transport dispatch, or adding a batch commit-ready barrier that records early completions and replays a wake after commit. Please also add a regression test where an RDMA task reaches a terminal state before `submitTransfer()` returns.

The callback propagation itself looks clean and CI is green, but I would request changes for this ordering race before merging.

Issue #2960, opened 2026-07-16
[RFC]: Enhanced RDMA RNIC Failover and Recovery


alogfans:
## Background

Mooncake Transfer Engine supports selecting local and remote RNICs in multi-RDMA-NIC environments through topology and NIC priority matrices. In production, however, a single RNIC may experience link down events, port errors, GID changes, QP errors, or other transient failures.

The current implementation already includes partial recovery mechanisms, such as endpoint reconstruction, GID reprobe, ready ACK, and metadata refresh. However, real RNIC down/up testing still exposed several gaps:

1. When a preferred local RNIC becomes unavailable, queued or newly submitted slices may not smoothly fail over to another local RNIC.
2. Peer RNIC failure and local RNIC failure are not clearly separated, so a local hardware issue may be misclassified as a remote rail issue.
3. After an RNIC transitions from `PORT_ERR` to `PORT_ACTIVE`, immediately publishing it as available may be too early. In real RoCE/mlx5 environments, the data path may still reject connections shortly after link-up.
4. Passive handshake failures do not always return structured error information to the active side, making diagnosis difficult.
5. Reusing old QPs after a pre-connected endpoint handshake failure may conflict with QP information already cached by the peer.

This RFC proposes improving RDMA RNIC failover recovery so that transfers can continue through fallback RNICs when a single RNIC fails, and recovered RNICs are safely republished.

## Goals

- Support handing off unfinished slices to another active local RNIC after a local RNIC failure.
- Distinguish local RNIC failures from peer rail failures.
- Update and publish local topology when a context becomes inactive or recovers.
- Delay reactivating a recovered RNIC after `IBV_EVENT_PORT_ACTIVE`.
- Improve handshake failure diagnostics.
- Preserve compatibility with existing RDMA ready ACK, GID reprobe, and endpoint lifecycle mechanisms.
- Provide a real link down/up validation script.

## Non-Goals

- Do not introduce a new transport protocol.
- Do not change the upper-level Transfer Engine API.
- Do not guarantee lossless recovery from every hardware fatal error.
- Do not refactor Store or TENT failover mechanisms in this RFC.
- Do not change the consistency model of metadata backends.

### 1. Local RNIC Handoff

When a local RDMA context is marked inactive, or when the worker pool detects a local completion failure, the current worker no longer fails the slice immediately. Instead, it tries to hand the slice off to another active local context.

During handoff, the following remote-side information is preserved:

- `target_id`
- `peer_nic_path`
[...]

PR #3555, opened 2026-08-20
[TransferEngine] Delay RDMA redispatch while rails recover

alogfans:
## Description

This PR makes classic RDMA WorkerPool delay and retry redispatch when all currently usable RDMA rails are temporarily unavailable, instead of immediately failing slices.

The change covers transient peer/local rail recovery windows seen during link flap chaos testing. Slices are placed into a delayed redispatch queue with bounded wait time, exponential backoff, jitter, and awareness of existing rail pause deadlines. If rails recover, the slice is redispatched and delivery continues. If recovery never happens within the bounded wait window, the slice fails gracefully.

This also reduces reconnect/handshake pressure during recovery windows by avoiding tight 100ms retry loops.

## Module

- [x] Transfer Engine (`mooncake-transfer-engine`)

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

**Test commands:**
```bash
cmake --build build --target tebench -j 16

python3 mooncake-transfer-engine/tests/chaos/te_chaos.py run \
  --target qjh001 \
  --metadata-mode p2p \
  --suite deterministic \
  --backend classic \
  --fault-kinds link-down \
  --seed 334455 \
  --duration 180 \
  --threads 2 \
  --block-size 65536 \
  --batch-size 16
```

**Test results:**
- [x] Manual testing done
- [x] `tebench` build passed
- [x] TE classic p2p link-down chaos test passed with `rc=0`
- [x] Verified no `AddressNotRegistered`, `corrupted size`, `Failed transfer detected`, `Data mismatch`, or delayed redispatch give-up in the passing run

## Checklist

- [x] I have performed a self-review of my own code
- [x] I have formatted my code using `./scripts/code_format.sh`
- [ ] I have run pre-commit on the files changed in this PR and all hooks pass
- [ ] I have updated the documentation (if applicable)
- [ ] I have added tests to prove my changes are effective
- [ ] For changes >500 LOC: I have filed an RFC issue

## AI Assistance Disclosure

- [x] AI tools were used

Codex helped implement and validate this targeted Transfer Engine RDMA recovery fix. The human submitter should review the changed lines before submission.

PR #3696, opened 2026-08-26
[TE] Add RDMA chaos runner and worker-pool failover retries

alogfans:
## Description

Add a dedicated classic Transfer Engine RDMA chaos runner and harden RDMA worker-pool retry behavior for temporary RNIC/rail unavailability.

This PR treats the chaos runner and WorkerPool reliability changes as one validation package:

- add `rdma_transport_chaos_test` plus `tests/chaos/te_chaos.py` for two-host RDMA chaos runs
- defer slices instead of failing immediately when all peer rails, target RNICs, or local RNIC handoff paths are temporarily unavailable
- handle posted-slice software timeouts in `WorkerPool` by retiring affected endpoints and fencing peer rails when an alternate exists
- keep `preTouchMemory()` from selecting an inactive placeholder context when one RNIC failed during initialization
- add `worker_pool_timeout_test` coverage for posted-slice timeout cleanup

Related: #2960

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
git diff --check
python3 -m py_compile mooncake-transfer-engine/tests/chaos/te_chaos.py
cmake -S . -B build
cmake --build build --target rdma_transport_chaos_test worker_pool_timeout_test rdma_transport_submit_task_test worker_pool_rail_state_test -j2
ctest --test-dir build -R 'worker_pool_timeout_test|rdma_transport_submit_task_test|worker_pool_rail_state_test' --output-on-failure
./scripts/code_format.sh --staged
```

**Test results:**
- [x] Unit tests pass

`rdma_transport_chaos_test` builds successfully. The actual chaos runner requires a two-host RDMA/SSH environment and was not executed in this local PR preparation.

## Checklist

`pre-commit` is not installed in this environment, so it was not run. PR-scoped formatting was run with `./scripts/code_format.sh --staged`.

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex was used to port the chaos runner/WorkerPool changes, trim unrelated edits, run build/test validation, and prepare this PR description. The human submitter should review and own the final changes.

### DIFF
--- a/mooncake-transfer-engine/include/transport/rdma_transport/worker_pool.h
+++ b/mooncake-transfer-engine/include/transport/rdma_transport/worker_pool.h
@@ -68,6 +68,11 @@ class WorkerPool {
     void redispatch(std::vector<Transport::Slice *> &slice_list, int thread_id,
                     bool handoff_to_local_worker = false,
                     bool defer_local_redispatch = false);
+    void delayRedispatch(Transport::Slice *slice, bool handoff_to_local_worker,
+                         const char *reason);
+    void releaseReadyDelayedSlices(int thread_id);
+    bool hasDelayedSlices() const;
+    uint64_t nextDelayedSliceDueNs() const;
 
     void transferWorker(int thread_id);
 
@@ -88,6 +93,8 @@ class WorkerPool {
     void markRailFailed(const std::string &peer_nic_path,
                         bool immediate_pause = false);
     bool isRailAvailable(const std::string &peer_nic_path);
+    uint64_t peerRailPauseRemainingNs(const std::string &peer_nic_path,
+                                      uint64_t now);
 
     // Retry helper: increment retry count and return whether retry is allowed
     static bool shouldRetrySlice(Transport::Slice *slice);
@@ -156,6 +163,14 @@ class WorkerPool {
     // Rail state management: peer_nic_path -> RailState
     std::unordered_map<std::string, RailState> rail_states_;
     std::mutex rail_state_lock_;
+
+    struct DelayedSlice {
+        Transport::Slice *slice = nullptr;
+        uint64_t due_ns = 0;
+        bool handoff_to_local_worker = false;
+    };
+    mutable std::mutex delayed_slices_lock_;
+    std::vector<DelayedSlice> delayed_slices_;
 
     // Rail monitor configuration
     const static int kRailErrorThreshold = 5;  // Errors before pause
--- a/mooncake-transfer-engine/include/transport/rdma_transport/worker_pool.h
+++ b/mooncake-transfer-engine/include/transport/rdma_transport/worker_pool.h
@@ -116,6 +116,11 @@ class WorkerPool {
                             RdmaEndPoint *endpoint = nullptr);
 
     bool tryHandoffToAnotherLocalWorker(Transport::Slice *slice);
+    bool selectAvailablePeerRailAlternative(
+        RdmaTransport::SegmentDesc *peer_segment_desc, int buffer_id,
+        int selected_device_id, uint64_t selection_seed, int &device_id,
+        std::string &peer_nic_path);
+    static uint64_t peerRailSelectionSeed(const Transport::Slice *slice);
 
     // Only direct local completion errors charge the context breaker.
     // Submit-side all-rails-unavailable failures remain peer/rail scoped.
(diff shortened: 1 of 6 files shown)


catyans (changes requested):
Thanks for adding recovery for temporarily unavailable RDMA rails and posted-slice timeouts. I found a race between the timeout monitor and the CQ poller that can allow both paths to process the same slice.

`performPollCq()` polls CQEs before acquiring `posted_slices_mutex_`. Meanwhile, `reapTimedOutPostedSlices()` can acquire the mutex, remove the same slice from `posted_slices_`, release the mutex, and pass its raw pointer to `handleTimedOutPostedSlices()`.

The CQ path then calls `untrackPostedSlices()`, but it ignores whether `erase()` actually removed anything and proceeds to dereference and finalize or redispatch the slice regardless. The timeout path may simultaneously read that slice and retire its endpoint. If the CQ path completes and releases the slice first, the monitor can also be left with a dangling pointer.

The current test only serializes "completion untracks first, timeout reaps afterward," so it does not cover the interleaving where a CQE has already been polled but the monitor claims the slice before `untrackPostedSlices()`.

Please give completion and timeout handling exclusive ownership of each posted slice, for example through an explicit per-slice state/claim protocol. The design also needs to handle a CQE that was already consumed when the timeout path wins, since deleting the endpoint cannot then generate another completion for that WR. A concurrent regression test covering this interleaving would be important.

I would request changes until timeout and CQ completion handling cannot concurrently process the same slice.

alogfans:
@catyans The mentioned problem is partically fixed in #3718. We will resume this PR after merging that.

catyans:
Thanks @alogfans for bringing this back after #3718. I re-read the current head, and the old CQ-vs-timeout ownership race is gone now: the timeout reaper that could claim the same raw slice is no longer part of this diff, and `posted_slices_` is only used for diagnostics. So that specific concern is resolved.

There are still two things I think we should align before calling this ready:

1. The implementation no longer matches the PR's recovery contract. The description says temporarily unavailable peer/local RNIC paths are deferred and that posted slices are reclaimed by a software timeout, but the current diff has no timeout reaper or `worker_pool_timeout_test`. More importantly, `submitPostSend()` marks a slice failed immediately when all peer rails are paused, and the local-RNIC handoff branch also marks it failed when no worker is currently available. Could you either restore the bounded deferral behavior, or narrow the description/tests so the intended fail-fast semantics are explicit? Right now the headline promise and the code say different things.

2. Copilot's remote P2P note still applies. `Host.popen()` writes a remote target's log on that host, while `discover_p2p_segment(log, ...)` reads the same path through the runner's local filesystem. A normal two-host `P2PHANDSHAKE` run will therefore time out unless that directory happens to be shared. Reading the log through `Host.run()` (or streaming it back) would make the advertised two-host runner actually usable.

The smaller runner cleanup notes—retaining/closing the local log handle and replacing the two `malloc()` argument arrays—look non-blocking to me, but they're easy wins while touching this. The rail-selection cleanup itself looks sensible; I mainly want the recovery behavior and validation tool to tell one consistent story.

alogfans:
We have done code revising, and integrate with #3555

catyans:
Thanks @alogfans—the revised scope is much easier to reason about. Removing the unexercised Python runner resolves my second concern, and the new delayed queue restores the intended behavior when peer or local rails are temporarily unavailable.

I found one liveness edge in the bounded-wait implementation, though. `delayRedispatch()` uses `slice->ts` as the start of the 120s deadline, but `redispatch()` clears `slice->ts = 0` whenever it finds a route and re-enqueues the slice. If `performPostSend()` then repeatedly gets `endpoint == nullptr`, or the endpoint setup keeps failing transiently, each trip back through `redispatch()` clears the timestamp and the next `delayRedispatch()` starts a fresh 120s window. `retry_cnt` is not advanced on those direct delay paths either, so a persistent setup failure can leave the task pending indefinitely despite the advertised maximum wait.

Could we keep a dedicated first-deferred timestamp (or otherwise preserve it until terminal completion) and add a deterministic route-found → setup-fails → delayed-again test that reaches the deadline? The existing rail-state test only verifies initial enqueue, so it would not catch the deadline being reset across retry cycles.

Separately, the current code-format gate is red, but that looks mechanical. The recovery direction itself is a good improvement; I just want the “bounded” part to stay bounded through the full retry loop.

PR #3699, opened 2026-08-26
[Store] Allocate client host memory via SPDK hugepages under USE_NOF

gitgaoqian:
## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## Why

In a USE_NOF build, client host memory segments allocated by
`MooncakeHostMemAllocator` (ShmHelper) must be DMA-safe for NVMe-oF
transfers. The previous memfd+mmap path produced plain kernel pages that
SPDK cannot use as NVMe-oF data buffers. NOF now falls back to SPDK
hugepage allocation for these segments.

## What changed

- `ShmHelper::allocate()`: under `USE_NOF`, allocate via
  `SpdkWrapper::GetInstance().Alloc(size, 0x1000, -1)` and record the
  segment with `fd = -1`.
- `ShmHelper::cleanup()` / `free()`: release such segments via
  `SpdkWrapper::GetInstance().Free()`, with correct interaction with the
  existing `USE_ASCEND_DIRECT` fabric (VMM) path.

## Boundaries

- Only active under `USE_NOF`; the memfd+mmap path is untouched
  otherwise.
- Ascend fabric (VMM) allocation keeps precedence.
- SPDK environment init is still performed by the caller, matching
  `memory_alloc.cpp`.

### DIFF
--- a/mooncake-store/src/shm_helper.cpp
+++ b/mooncake-store/src/shm_helper.cpp
@@ -14,6 +14,9 @@
 #if defined(USE_ASCEND_DIRECT)
 #include "ascend_allocator.h"
 #endif
+#ifdef USE_NOF
+#include "spdk/spdk_wrapper.h"
+#endif
 
 #ifndef MOONCAKE_SHM_NAME
 #define MOONCAKE_SHM_NAME "mooncake_shm"
@@ -59,6 +62,12 @@ bool ShmHelper::cleanup() {
                 continue;
             }
 #endif
+#ifdef USE_NOF
+            // SPDK DMA buffers are released through the SPDK allocator.
+            SpdkWrapper::GetInstance().Free(shm->base_addr);
+            shm->base_addr = nullptr;
+            continue;
+#endif
             if (munmap(shm->base_addr, shm->size) == -1) {
                 LOG(ERROR) << "Failed to unmap shared memory: "
                            << strerror(errno);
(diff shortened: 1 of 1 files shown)


fcczzz (commented):
I reproduced a blocking regression on PR head.

I built Mooncake locally in an Ubuntu 22.04 Docker container with `USE_NOF=ON`, `BUILD_UNIT_TESTS=ON`, SPDK v23.01.1, HugeTLB mounted at `/dev/hugepages`, and unlimited memlock.

First, I verified that SPDK DMA allocation works in this environment:

```bash
./build-nof/mooncake-store/tests/client_buffer_test \
  --gtest_filter=ClientBufferTest.SpdkDmaAllocatorDestroysWithSpdkFree
```

This test passed.

I then exercised the existing DummyClient shared-memory path:

```bash
./build-nof/mooncake-store/tests/dummy_client_get_buffer_test \
  --gtest_filter=DummyClientGetBufferTest.GetBuffer_AllocatorFallback
```

This test failed during `SetupStack()` with:

```text
Invalid shm_fd during IPC registration
Failed to register SHM via IPC
Failed to bring up real+dummy stack
```

Under `USE_NOF`, `ShmHelper::allocate()` now records the SPDK allocation with `fd = -1`. However, `DummyClient::setup_dummy()` passes that segment to `register_shm_via_ipc()`, which requires a valid fd so the RealClient process can map the same memory. Therefore, a DummyClient with `local_buffer_size > 0` can no longer initialize.

Simply removing the `fd < 0` check would not fix this: a raw SPDK virtual address from one process is not automatically valid or mapped in another process.

How is this SPDK allocation expected to be shared with the RealClient process? The current DummyClient/RealClient IPC protocol requires an fd-backed mapping, while this change records the allocation with `fd = -1`. Passing only the virtual address would not map the same memory in the other process. I think the cross-process sharing path needs to be addressed before merge.

gitgaoqian:
thanks for review, i will check this .and convert to draft firstly.

PR #3700, opened 2026-08-26
[feat]Add observe-only segment admission controller

### LINKED ISSUE #3688: [Bug]: 华为昇腾910C，4机2P1D/实例。 当一个mooncake master管理 2个实例，其中一个实例停掉、启动时会导致其他所有实例业务slo下降，ttft增加严重，持续很长时间。
### Bug Report
mooncake版本：[v0.3.11.post1](https://github.com/kvcache-ai/Mooncake/releases/tag/v0.3.11.post1)
vllm版本：vllm-ascend 0.23.0

一个mooncake master管理 2个实例（实例A 实例B），其中一个实例A重启：
1、实例A停掉时，不会把实例B拖死--OK，问题修复
2、实例A停掉时，实例B往实例A存KV的请求会发生超时，业务ttft飙升--影响
3、实例A启动初期，实例B往实例A存KV的请求会发生超时，业务ttft飙升--影响大

第3点通过测试观察，应该是因为：实例A启动后，DRAM是空的（没重启的实例的DRAM基本都接近用满），mooncake master会指示所有KV都往实例A存，导致的transfer engine传输压力倍增导致的，hixl传输超时、重传

### Before submitting...

JayLee0701:
## Description
Related issue: https://github.com/kvcache-ai/Mooncake/issues/3688

When one instance managed by Mooncake stops, it may remain eligible for allocation until heartbeat TTL expiration and unmount. Remote writes assigned to that instance can therefore wait for transport timeout and retry.

After the instance restarts, its empty DRAM segment is immediately treated as fully available, while surviving segments may already be nearly full. This can concentrate new writes on the restarted instance, overload the transfer path, trigger HIXL timeout/retransmission, and increase TTFT for other healthy instances.

The planned solution is to add a `SegmentAdmissionController` before the existing allocation strategies. It will:

- maintain `RAMPING`, `ACTIVE`, and `QUARANTINED` states;
- gradually ramp up newly mounted or recovered segments;
- limit per-segment remote-write operations and in-flight bytes;
- quarantine segments after repeated transfer failures;
- use client transfer results and owner heartbeats for recovery;
- exclude unhealthy or overloaded segments during `PutStart`.

This PR implements the first, observe-only stage:

- add the admission controller, configuration, and validation;
- track runtime state across mount, remount, heartbeat, restore, and unmount;
- observe targets selected by the existing allocation path;
- add admission metrics and Admin segment-detail fields;
- add fake-clock, configuration, metrics, Admin, and integration tests.

This PR does not change allocation results, create reservations, reject `PutStart`, or add client result-reporting RPCs. If `enforce` is configured at this stage, the master logs a warning and continues in observe-only mode.

Therefore, this PR provides the foundation and production observability for the complete fix, but does not close #3688 by itself.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?
Tests were added for:
- ramp-up and quarantine state transitions using a fake clock;
- mount/remount/heartbeat/unmount runtime handling;
- configuration propagation through normal and HA paths;
- metrics and Admin fields;
- observe-only decisions not changing preferred placement.

**Test commands:**
```bash
cmake -S . -B build-pr1 -G Ninja \
  -DCMAKE_BUILD_TYPE=RelWithDebInfo \
  -DBUILD_UNIT_TESTS=ON \
  -DBUILD_EXAMPLES=OFF \
  -DWITH_TE=ON \
  -DWITH_STORE=ON \
  -DWITH_STORE_RUST=OFF \
  -DWITH_STORE_GO=OFF \
  -DWITH_P2P_STORE=OFF \
  -DWITH_EP=OFF \
  -DUSE_CUDA=OFF \
  -DUSE_HTTP=ON \
  -DUSE_ETCD=OFF \
  -DSTORE_USE_ETCD=OFF \
  -DWITH_METRICS=ON

cmake --build build-pr1 --parallel "$(nproc)" \
  --target segment_admission_controller_test \
           master_service_config_test \
           master_metrics_test \
           master_admin_server_test

ctest --test-dir build-pr1 \
  --parallel "$(nproc)" \
  --output-on-failure \
  -R '^(segment_admission_controller_test|master_service_config_test|master_metrics_test|master_admin_server_test)$'

```

**Test results:**
- [x] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used
Cursor was used to assist with implementation and test drafting.

### DIFF
--- a/mooncake-store/include/segment_admission_config.h
+++ b/mooncake-store/include/segment_admission_config.h
@@ -1,0 +1,120 @@
+#pragma once
+
+#include <cmath>
+#include <cstdint>
+#include <optional>
+#include <stdexcept>
+#include <string>
+#include <string_view>
+
+namespace mooncake {
+
+enum class SegmentWriteAdmissionMode : uint8_t {
+    DISABLED = 0,
+    OBSERVE = 1,
+    ENFORCE = 2,
+};
+
+inline std::string_view ToString(SegmentWriteAdmissionMode mode) noexcept {
+    switch (mode) {
+        case SegmentWriteAdmissionMode::DISABLED:
+            return "disabled";
+        case SegmentWriteAdmissionMode::OBSERVE:
+            return "observe";
+        case SegmentWriteAdmissionMode::ENFORCE:
+            return "enforce";
+    }
+    return "unknown";
+}
+
+inline std::optional<SegmentWriteAdmissionMode> ParseSegmentWriteAdmissionMode(
+    std::string_view value) noexcept {
+    if (value == "disabled") {
+        return SegmentWriteAdmissionMode::DISABLED;
+    }
+    if (value == "observe") {
+        return SegmentWriteAdmissionMode::OBSERVE;
+    }
+    if (value == "enforce") {
+        return SegmentWriteAdmissionMode::ENFORCE;
+    }
+    return std::nullopt;
+}
+
+struct SegmentAdmissionConfig {
+    SegmentWriteAdmissionMode mode{SegmentWriteAdmissionMode::OBSERVE};
+    uint64_t ramp_up_duration_sec{60};
+    double ramp_initial_ratio{0.05};
+    uint64_t ramp_min_successful_remote_writes{16};
+    uint64_t max_inflight_remote_write_ops{64};
+    uint64_t max_inflight_remote_write_bytes{0};
+    uint64_t failure_window_sec{5};
+    uint64_t failure_threshold{3};
+    uint64_t quarantine_duration_sec{10};
+    uint64_t result_retention_sec{30};
+
+    void Validate() const {
+        if (mode != SegmentWriteAdmissionMode::DISABLED &&
+            mode != SegmentWriteAdmissionMode::OBSERVE &&
+            mode != SegmentWriteAdmissionMode::ENFORCE) {
+            throw std::invalid_argument(
+                "segment_write_admission_mode is invalid");
+        }
+        if (ramp_up_duration_sec == 0) {
+            throw std::invalid_argument(
+                "segment_ramp_up_duration_sec must be greater than 0");
+        }
+        if (!std::isfinite(ramp_initial_ratio) || ramp_initial_ratio <= 0.0 ||
+            ramp_initial_ratio > 1.0) {
+            throw std::invalid_argument(
+                "segment_ramp_initial_ratio must be in (0, 1]");
+        }
+        if (failure_window_sec == 0) {
+            throw std::invalid_argument(
+                "segment_failure_window_sec must be greater than 0");
+        }
+        if (failure_threshold == 0) {
+            throw std::invalid_argument(
+                "segment_failure_threshold must be greater than 0");
+        }
+        if (quarantine_duration_sec == 0) {
+            throw std::invalid_argument(
+                "segment_quarantine_duration_sec must be greater than 0");
+        }
+        if (result_retention_sec == 0) {
+            throw std::invalid_argument(
+                "segment_result_retention_sec must be greater than 0");
+        }
+    }
+};
+
+inline SegmentAdmissionConfig BuildSegmentAdmissionConfig(
+    std::string_view mode, uint64_t ramp_up_duration_sec,
+    double ramp_initial_ratio, uint64_t ramp_min_successful_remote_writes,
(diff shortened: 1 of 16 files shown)


ykwd:
The motivation for this PR is appreciable. There is already another PR addressing the issue of Store nodes going offline: https://github.com/kvcache-ai/Mooncake/pull/2991

As for the issue of a newly added node being overwhelmed by requests, it seems that writes would only be directed to the new node when the other segments are already full. However, limiting writes to the new node in that situation could in turn cause PUT requests to fail. I think we also need to consider things like the conditions that trigger eviction, the case where all nodes are newly added, and so on. This does not seem to be a very straightforward problem.

JayLee0701:
> The motivation for this PR is appreciable. There is already another PR addressing the issue of Store nodes going offline: #2991
> 
> As for the issue of a newly added node being overwhelmed by requests, it seems that writes would only be directed to the new node when the other segments are already full. However, limiting writes to the new node in that situation could in turn cause PUT requests to fail. I think we also need to consider things like the conditions that trigger eviction, the case where all nodes are newly added, and so on. This does not seem to be a very straightforward problem.

Thanks for your reply. The issue mentioned in the issue about write operations being directed to the new node seems to be the situation you described. Situations like triggering eviction conditions when all nodes are new, and so on, definitely need to be considered.

catyans:
I like the idea of rolling this out in observe-only mode first — it gives us a chance to learn from real traffic before admission starts affecting puts. @ykwd’s point about the policy edge cases makes that especially valuable.

One identity edge case looks important to settle before relying on the observations: the controller stores runtime state by segment UUID, but `ObserveRemoteWrite()` resolves a target through a single `segment_name -> UUID` entry. Mooncake permits multiple mounted segments with the same name. Mounting the second one overwrites that entry, and unmounting it erases the entry without restoring the first, so writes to one segment can be charged to another or stop being observed entirely.

Could the allocation path pass the actual segment ID into the controller, or otherwise preserve all same-name registrations and resolve the exact descriptor? A same-name mount plus reverse-unmount regression would be helpful here. This matters even in observe mode because the collected data is what we’ll use to validate a later enforcement policy.

mooncake-store/src/master.cpp
```diff
@@ -323,6 +323,32 @@ DEFINE_string(
     allocation_strategy, "random",
     "Allocation strategy for segments, random | free_ratio_first | cxl | "
     "ssd_free_ratio_first | local_first");
+DEFINE_string(segment_write_admission_mode, "observe",
```
Aionw, line 326: Define all of those admin config here will cause a long cmdline arg list, prefer to use config file instead.

mooncake-store/src/master_admin_service.cpp
```diff
@@ -698,11 +698,25 @@ struct HttpSegmentDetailItem {
     uint64_t allocator_used_bytes{0};
     uint64_t allocator_capacity_bytes{0};
     double allocator_usage_percent{0.0};
+    bool admission_tracked{false};
```
Aionw, line 701: Perfer use a struct to wrap these things.

mooncake-store/src/master_admin_service.cpp
```diff
@@ -698,11 +698,25 @@ struct HttpSegmentDetailItem {
...
+    std::string admission_mode;
```
Aionw, line 702: mode and state can be enum value, there is no type info in string and lead to potentional misuse.

mooncake-store/include/segment_admission_config.h
```diff
@@ -0,0 +1,120 @@
...
+    uint64_t failure_window_sec{5};
+    uint64_t failure_threshold{3};
+    uint64_t quarantine_duration_sec{10};
+    uint64_t result_retention_sec{30};
+
+    void Validate() const {
+        if (mode != SegmentWriteAdmissionMode::DISABLED &&
+            mode != SegmentWriteAdmissionMode::OBSERVE &&
+            mode != SegmentWriteAdmissionMode::ENFORCE) {
+            throw std::invalid_argument(
```
Aionw, line 60: Would be better to return error instead of throw?

Issue #3363, opened 2026-08-10
[SegmentPool] Refactor Memory and CXL management into SegmentPool


Aionw:
Parent RFC: #3360
Depends on: #3362

## Goal

Turn the remaining Memory/CXL segment manager into the region-oriented `SegmentPool` boundary.

## Scope

- Refactor or rename the post-LocalSSD `SegmentManager` into `SegmentPool`.
- Define a mounted pool element as a client-mounted, addressable allocation region.
- Keep Memory and CXL catalog, indexes, allocator ownership, status, and two-phase unmount inside the pool.
- Introduce stateful Memory/CXL region drivers that own physical allocator resources.
- Extract pointer-based placement indexing and concrete replica allocation while preserving all existing policy behavior.
- Separate allocator and pool snapshot codecs while preserving the current wire shape.
- Update internal MasterService call sites and tests while retaining external RPC contracts.

## Out of scope

- Moving NoF state into the pool; that is tracked separately.
- Changing placement-policy behavior, segment-name grouping semantics, locality ordering, or CXL replica representation.
- Generalizing promotion targets.

## Implementation stack

Mechanisms land before the final caller switch:

| Order | PR | Review focus |
| ---: | --- | --- |
| 1 | #3703 | Region-driver resource ownership, import, and rollback |
| 2 | #3704 | Placement pointer stability, policy feedback, and hot-path allocation |
| 3 | #3705 | SegmentPool catalog/index invariants and lifecycle transactions |
| 4 | #3706 | Snapshot validation, allocator adoption, and fork safety |
| 5 | #3707 | MasterService/NoF call-site cutover and legacy-path deletion |

Each PR is based on the preceding PR. Merge in order; do not retarget later layers to `main` until their dependency has landed.

## Acceptance criteria

- `SegmentPool` is the sole owner of Memory/CXL mounted-region and allocator state.
- CXL remains represented as a MEMORY replica and keeps its existing global allocator behavior.
- Same-name segment grouping and existing mount/remount/unmount error behavior are preserved.
- Snapshot wire shape and LocalSSD/NoF ownership boundaries are preserved.
- Release PutStart comparisons remain within 3% single-thread QPS, 5% multi-thread QPS, and 10% p99 regression thresholds.
- Focused driver, placement, SegmentPool, snapshot, NoF mapping, LocalSSD, offload, promotion, and HA tests pass.

## Contribution notes

[...]

Issue #3360, opened 2026-08-10
[RFC]: Refactor Mooncake Store segment management around SegmentPool


Aionw:
## Summary

Mooncake Store currently splits mounted-region ownership between `SegmentManager` and `NoFSegmentManager`, while client-local SSD state is embedded in `SegmentManager` even though it is not an addressable allocation region.

This RFC proposes a behavior-preserving refactor:

- `SegmentPool` owns client-mounted, addressable regions that the Master can allocate from: Memory, CXL, and NoF.
- `LocalSsdManager` owns per-client local SSD registration, capacity, usage, and existing local tier state.
- Existing promotion and offload flows coordinate the two components; they are not owned by `SegmentPool`.

```mermaid
flowchart LR
  subgraph Current
    M1[MasterService] --> SM[SegmentManager]
    SM --> MEM1[Memory]
    SM --> CXL1[CXL]
    SM --> LSSD1[LocalSSD state]
    M1 --> NM[NoFSegmentManager]
    NM --> NOF1[NoF]
  end

  subgraph Proposed
    M2[MasterService] --> SP[SegmentPool]
    SP --> MEM2[Memory regions]
    SP --> CXL2[CXL regions]
    SP --> NOF2[NoF regions]
    M2 --> LS[LocalSsdManager]
    LS --> LSSD2[Client-local object stores]
    LS -. existing offload / promotion .-> SP
  end
```

## Motivation

The current layout duplicates mounted-segment catalogs, indexes, allocators, locking, and lifecycle code across Memory/CXL and NoF. It also couples LocalSSD registration and task state to the memory segment manager.

A resource should belong to `SegmentPool` because it is an addressable allocation region, not because of its physical medium. This keeps NoF in the pool while separating LocalSSD, whose data is object/file based and identified by client.

## Target architecture

For addressable Memory/CXL regions, the conceptual dependency remains `object catalog / request orchestration -> SegmentPool -> RegionDriver`. Replica placement is a stateless execution seam between the object layer and the pool: it borrows a lock-scoped placement view but owns no catalog, target, allocator, or physical resource.

```mermaid
flowchart TB
  subgraph ObjectLayer["Object layer"]
    MS["MasterService<br/>request orchestration"]
    OC["Object catalog<br/>MetadataStore and object shards"]
    MS --> OC
  end

  subgraph PlacementLayer["Stateless replica placement"]
    RP["ReplicaPlacement for SegmentPool<br/>static source binding"]
[...]

PR #3703, opened 2026-08-26
[Store] Introduce stateful region resource drivers

Aionw:
## Description

Part of #3363 and tracked by #3360.

Stack order: **1/5**. The next layer is #3704.

This is layer 1 of the SegmentPool stack. It introduces the physical-resource ownership mechanism without changing `MasterService` call sites:

- adds internal `RegionKind`, `RegionResourceSpec`, and initial allocation state
- introduces stateful Memory and CXL region drivers
- makes prepared resources transactional so abandoned opens/adoptions roll back
- keeps allocators owned by driver resources and exposes stable non-owning allocation targets
- centralizes allocator import validation while preserving descriptor order

No RPC schema, external configuration name/value, snapshot wire shape, or placement behavior changes in this layer.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

```bash
/tmp/mooncake-region-driver-build/mooncake-store/tests/buffer_allocator_test
/tmp/mooncake-region-driver-build/mooncake-store/tests/region_driver_test
git diff --check origin/main..HEAD
```

- [x] Unit tests pass
- [ ] Integration tests pass (not applicable to this mechanism-only layer)

## Checklist

- [x] I have formatted my code
- [x] Documentation is not required because external behavior is unchanged
- [x] I have added focused tests for lifecycle, rollback, and import validation
- [x] For changes >500 LOC: RFC #3360 tracks the change

## AI Assistance Disclosure

- [ ] No AI tools were used


Icedcoco:
Could we keep an exception boundary in CreateBufferAllocator() or at the mount call sites?

The previous MountSegment() path caught allocator-construction exceptions and converted them to INVALID_PARAMS. With the new factory, OffsetBufferAllocator construction and parts of CachelibBufferAllocator construction can still throw, so an allocation failure during MountSegment() or MountNoFSegment() might escape as an exception instead of returning the expected error code.

Is this behavior intentional? If not, would it make sense to catch these exceptions in the factory and map them to ErrorCode::INVALID_PARAMS, together with a focused test for constructor failure?

Icedcoco:
Could we clarify whether BuildRegionLiveAllocations() is expected to preserve the previous alias behavior?

The old BuildRegionInitialState() accepted descriptors whose endpoint matched either spec.transport_endpoint or spec.name, while the new helper only accepts spec.transport_endpoint. The current MasterService::ReMountSegment() path appears to canonicalize the endpoint before calling this helper, so the existing caller may be unaffected.

However, future SegmentPool restore paths may call BuildRegionLiveAllocations() directly. Would it be preferable to retain the alias compatibility, or should the helper explicitly require canonicalized descriptors and document that contract? A small regression test might help make the intended behavior clear.

Aionw:
Thanks for flagging this. The behavior is intentional: caller/parameter-dependent construction failures should be returned as `INVALID_PARAMS`, while allocator metadata exhaustion follows the process-level fail-fast policy rather than being reported as an invalid request.

I pushed e197ae12 to make that boundary explicit. `CachelibBufferAllocator::Create()` now validates the complete parameter-dependent layout, including the slab-count narrowing used by the constructor, and the focused test covers that failure. After the existing mount validation, `OffsetBufferAllocator` has no parameter-dependent throwing construction path; only metadata allocation failures remain able to escape.

Aionw:
@Icedcoco Thanks for pointing this out. The stricter behavior is intentional: `BuildRegionLiveAllocations()` operates on canonicalized descriptors and does not perform alias resolution. Alias compatibility belongs to the recovery component that has the segment/catalog context.

The current `MasterService::ReMountSegment()` path already matches either the segment name or transport endpoint and rewrites the descriptor to the canonical transport endpoint before calling this helper. Future SegmentPool restore paths should follow the same contract. This keeps the region helper consistent and avoids coupling it to historical segment-name aliases. I’ll make the canonicalization requirement explicit in the helper contract and add a focused test.

Icedcoco:
Thanks for the follow-up and for clarifying the intended exception policy. The latest change makes that boundary clearer, and I do not see a blocking issue from this point.

A few small design questions for discussion:

1. BufferAllocatorBase now has a new pure virtual base() method, and CachelibBufferAllocator construction has moved behind Create(). Since allocator.h is referenced in the C++ API documentation, should we treat this as an intentional API evolution, or would lightweight compatibility wrappers be useful?

2. I noticed that CXL capacity metric accounting is still handled by SegmentManager, while CxlRegionDriver now owns the allocator itself. When the later SegmentPool path starts using CreateRegionDrivers() directly, where would you prefer the CXL capacity registration and cleanup to live? A focused metric test might help document that ownership.

3. BuildRegionLiveAllocations() currently requires the canonical transport endpoint, while the previous helper also accepted the segment name as an alias. The current MasterService::ReMountSegment() path appears to canonicalize this before calling the helper. Would it be useful to document that contract explicitly, or should the helper continue accepting both forms for compatibility?

Aionw:
@Icedcoco Thanks for raising these design questions. I pushed a121df28 with the follow-ups.

For the allocator API question, `BufferAllocatorBase` and concrete allocator construction are internal Master implementation details rather than part of the supported Store Client C++ API. The API index was misleading here, so I removed the `BufferAllocatorBase interface` claim from the Client API description. With that boundary clarified, the new `base()` requirement and moving CacheLib construction behind `Create()` are intentional internal API evolution; I did not add compatibility wrappers. In particular, keeping the constructor private prevents bypassing the explicit `INVALID_PARAMS` factory contract and avoids reintroducing a throwing construction path.

I also moved CXL capacity accounting to the lifetime of `CxlRegionDriver`, which owns the global allocator, and added a focused test covering registration and cleanup. Finally, the canonical-endpoint requirement is now documented on `BuildRegionLiveAllocations()`, with a regression test confirming that a segment-name alias must be resolved by the recovery layer first.

Icedcoco (approved):
LGTM. I reviewed the latest head, including the allocator validation, region-driver ownership/lifetime changes, CXL metric accounting, canonical endpoint contract, and the focused tests. I did not find any blocking correctness or compatibility issue in this layer. The latest triage and CodeQL checks are still pending, so merge should wait for the normal CI gate to complete.

ykwd (approved):

PR #3704, opened 2026-08-26
[Store] Introduce placement index and replica allocator

Aionw:
## Description

Part of #3363 and tracked by #3360.

Stack order: **2/5**. Depends on #3703; the next layer is #3705. The GitHub base intentionally points to `codex/region-driver-foundation`.

This is layer 2 of the SegmentPool stack. It introduces placement and replica-allocation mechanisms without switching `MasterService`:

- adds pointer-based `PlacementIndex` with stable logical groups and same-name physical targets
- adds concrete `ReplicaAllocator` with one enum dispatch per request
- keeps policies concrete and uses reusable thread-local scratch storage
- adds the static `ReplicaPlacement<PlacementSource>` hook used by later owners
- preserves preferred/excluded feedback, best-effort replicas, LocalFirst ordering, CXL, and NoF policy mapping

There are no compatibility adapters and no placement-policy behavior change.

## Module

- [ ] Other

## Type of Change

- [x] Refactor
- [x] Performance improvement
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

```bash
/tmp/mooncake-region-driver-build/mooncake-store/tests/replica_allocator_test
/tmp/mooncake-region-driver-build/mooncake-store/benchmarks/allocation_strategy_bench \
  --workload=fillup --num_allocations=10000 --segment_capacity=1024
git diff --check codex/region-driver-foundation..HEAD
```

The focused Random/1-replica/10-group/512KB median-of-five comparison measured +7.52% QPS and -1.54% p99 latency versus the baseline, with 10000/10000 successful allocations.

- [x] Unit tests pass
- [x] Manual benchmark completed

## Checklist

- [x] I have performed a self-review
- [x] Formatting and touched-file pre-commit pass
- [x] External behavior and configuration are unchanged
- [x] Focused policy, fallback, same-name target, and best-effort tests are included
- [x] RFC #3360 tracks the change

## AI Assistance Disclosure

- [ ] No AI tools were used


Icedcoco:
One thing I wanted to clarify about the CXL path. Could a logical group in PlacementIndex contain both CXL and non-CXL targets when they share the same name?

TryAllocateFromGroup() currently attempts all targets in the group, while PreferredOnlyPlacementPolicy uses that helper directly. In that mixed-target case, a CXL-preferred allocation could select the native target first and return a non-CXL descriptor. The previous CXL path explicitly filtered out non-CXL targets.

Maybe mixed-media groups are forbidden by an invariant that is enforced elsewhere. If so, this looks good to me.

Aionw:
Good catch. Mixed-media groups are not currently forbidden, so the refactor could indeed let the CXL path allocate from a native target. I restored explicit CXL classification on placement targets and made the preferred-only CXL path skip non-CXL targets in df48febc. I also added a regression test with native and CXL targets sharing one logical group; when the CXL target cannot allocate, the request now returns NO_AVAILABLE_HANDLE instead of falling back to a native descriptor.

Aionw:
Follow-up: I reworked the initial fix in e4e4f25b so the generic preferred-only policy no longer has implicit CXL semantics. Placement targets expose a typed `PlacementTargetKind`, `PreferredOnlyPlacementPolicy` carries its required kind explicitly, and the group helper applies a general target predicate. Mixed-name groups remain supported, and the regression test now checks both CXL-only rejection and native-only selection.

Aionw:
I tightened this further in 2414e03b: the index now keys placement groups by `(logical name, target kind)`, making every group allocation-equivalent. Native and CXL targets can still share a logical name, but they are never stored in the same placement group. `TryAllocateFromGroup()` is back to the unfiltered implementation; policy resolution selects the correctly typed group up front. This supersedes the predicate-based approach from the previous follow-up.

Icedcoco (approved):
LGTM

ykwd (approved):

PR #3705, opened 2026-08-26
[Store] Introduce SegmentPool catalog and lifecycle

Aionw:
## Description

Part of #3363; architectural work tracked by #3360. The placement foundation from #3704 is already in main. This PR adds the next layer: a `SegmentPool` that coordinates mounted-region metadata, allocator resources, placement eligibility, and lifecycle transactions under one lock. It also simplifies the placement interfaces introduced by that foundation.

For example, preparing an unmount removes a region from allocation immediately while retaining the driver resource for cleanup. An immediate transaction can commit or roll back; a graceful transaction keeps existing replicas readable until finalization. Catalog state, driver activation, and placement membership change together under the Pool write lock.

### Architecture and ownership

```mermaid
flowchart TD
    Caller["Pool caller"] --> W["SegmentPool::WriteAccess<br/>exclusive lock"]
    Caller --> R["SegmentPool::ReadAccess<br/>shared lock"]
    Caller --> A["SegmentPool::AllocateReplicas / AllocateInSegment"]
    A --> RA["ReplicaAllocator + placement policy<br/>shared lock through allocation"]
    subgraph Pool["SegmentPool owns the lock and components"]
        C["RegionCatalog<br/>all mounted regions"]
        P["PlacementIndex<br/>active allocation candidates"]
        D["RegionDriverRegistry<br/>Host Memory / CXL drivers"]
        U["StorageUsageTracker"]
    end
    W --> C
    W --> P
    W --> D
    R --> C
    R --> P
    R --> D
    RA --> P
    D --> RR["RegionResource<br/>owns AllocationCandidate"]
    RR --> BA["BufferAllocator<br/>allocation and allocator lifetime"]
    P -. "borrows candidate references" .-> RR
    BA --> U
```

| Component | Responsibility and boundary |
|---|---|
| `RegionCatalog` | Owns `MountedRegion` records across all lifecycle states. Boost.MultiIndex maintains the unique segment-ID index and non-unique segment-name index over the same records. Owner lookup derives from those records. It has no placement-specific host index. |
| `PlacementIndex` | Contains only active candidates, keyed by `(segment_name, candidate kind)`. Each entry's `flat_set` contains interchangeable allocation candidates; multi-replica selection uses an entry at most once. Inactive regions remain in Catalog, not in this index. |
| `EntryMap` inside PlacementIndex | Owns entries and keeps name lookup and the dense active-entry vector synchronized. Both views describe the same active entries. The dense view supports random selection without scanning the catalog. |
| Host index inside PlacementIndex | Maps host ID to an ordered, indexable `flat_map<segment_name, member_count>`. Counts span regions and kinds, so removing one candidate does not erase a name still represented on that host. Host identity is separate from the transfer endpoint. |
| `RegionDriver` / `RegionResource` | Drivers create, stage, publish, activate, and erase resources. Resources own candidates; candidates delegate allocation to native or CXL allocators. Prepared-resource commit delegates map mutation back to the driver and retains the replaced resource until placement references have been updated. |
| `ReplicaAllocator` and policies | Select eligible entries and allocate replicas. Policies handle random, preferred, local-first, and ranked ordering; candidates provide the concrete allocation operation. Production randomness uses the shared thread-local RNG utilities. |
| Pool access objects | Nested `ReadAccess`/`WriteAccess` express lock scope. Transactions call private WriteAccess operations; driver internals stay inside the driver. Private construction keys restrict creation without `friend` access. |

Local-first placement tries explicit preferences, then visits the maintained host index starting at the writer host (with wraparound). Within each host, the object-key hash rotates the sorted segment names. Names are deduplicated across hosts, and traversal stops once enough replicas have been allocated. This avoids rebuilding host/name lists from all catalog records on every allocation.

### Lifecycle transactions

```mermaid
sequenceDiagram
    participant Caller
    participant W as Pool WriteAccess
    participant T as Lifecycle transaction
    participant D as RegionDriver
    participant I as Catalog and PlacementIndex
    Caller->>W: PrepareMount / PrepareRestore / PrepareAdopt
    W->>D: Stage resource
    W-->>Caller: RegionMountTxn
    Caller->>T: Commit(write access)
    T->>W: PublishMount
    W->>D: Commit prepared resource
    W->>I: Publish record and placement membership
    Note over Caller,I: Dropping an uncommitted mount releases staged resources
    Caller->>W: PrepareUnmount / PrepareGracefulUnmount
    W->>I: Change status and remove placement membership
    W->>D: Deactivate resource
    W-->>Caller: Unmount transaction with lifecycle generation
    Note over Caller,I: Write lock may be released during cleanup or draining
    Caller->>T: Commit / Rollback / Finalize with reacquired write access
    T->>W: Validate IDs, status, and generation
    W->>D: Erase or reactivate resource
    W->>I: Erase record or restore lifecycle state
```

- **Mount / restore / adopt:** prepare stages the resource; commit validates catalog/resource conflicts before publication. Adopted allocators attach to the Pool usage tracker only after successful publication, so abandoned or rejected adoption does not leave phantom capacity.
- **Immediate unmount:** prepare snapshots the previous state and marks the region `UNMOUNTING`. `Commit` removes it; `Rollback` restores it. Completion consumes the move-only transaction, including on failure.
- **Graceful unmount:** prepare marks the region `GRACEFULLY_UNMOUNTING` and stops allocation while retaining readable resources. Dropping the transaction leaves draining in progress; another prepare can resume the same round. `Finalize` consumes the transaction and removes the resource.
- **Stale transactions:** Catalog assigns a new generation on registration and actual status changes; `Clear()` does not reset the counter. Unmount completion checks the captured generation, rejecting old transactions after ID reuse or a new draining round. Repeated preparation of the same graceful round preserves its generation.
- **Usage:** live allocator usage and catalog snapshots serve different purposes. Shared CXL allocator capacity and retained allocator lifetimes must not be counted once per binding. Mount capacity metrics are tracked separately from allocator lifetime usage.

### Scope and reading order

`MasterService` continues to use the existing SegmentManager path in this layer. RPCs, snapshot wire formats, LocalSSD/NoF ownership, and the service cutover are outside this PR.

Suggested reading order: `segment/pool.h` → `segment/catalog.h` and `placement/index.h` → `segment/region_driver.h` → `segment/pool_write_access.h` and `segment/transaction.h` → `segment/pool_mount.cpp` and `segment/pool_write_access.cpp` → `placement/replica_allocator.cpp` → lifecycle/index tests.

## Module

## Type of Change

- [x] Refactor
- [x] Bug fix
- [x] Performance improvement

### Correctness

Validated after merging main at PR head `d9f79610`:

| Suite | Passed |
|---|---:|
| Pool / Catalog | 28 |
| ReplicaAllocator / PlacementIndex | 15 |
| RegionDriver | 10 |
| **Total** | **53** |

Coverage includes mount/remount/restore, publication conflicts, unmount rollback, graceful resource retention and resume, stale transaction rejection after remount/new drain/Clear, abandoned and rejected adoption accounting, shared CXL bindings, host ordering/deduplication, policy fallback, and allocation-versus-unmount synchronization. PR-scoped pre-commit and whitespace checks passed.

- [x] Unit tests pass
- [ ] Full Store integration tests (service cutover is a later layer)

### Benchmark comparison

**Baseline:** `origin/main` at `82fff0432a99c2c523709a804cc0cca3097eafef`, independently built with unchanged production sources.

**Compared code:** the Pool development snapshot used in that paired experiment (not separately committed; it predates the maintained host index and the final lifecycle fixes). These measurements were not rerun on merged head `d9f79610`; the correctness tests above were.

**Setup:** AMD EPYC 7742 (256 logical CPUs); pinned workers; `-O3`, `-fno-tree-slp-vectorize`, assertions and metrics enabled. OFFSET allocator, native/TCP regions of 16 MiB, 16 host IDs, 4 KiB per replica with immediate release. Each worker performs 1,000 warmups. Matrix cases run for 1 second, repeated 3 times; medians are reported, with latency sampled every 64 operations. CPUs are shared, not isolated.

Local-first uses 3 replicas, a fixed writer host/object key, and no lifecycle writer. Random uses 1 or 3 replicas. Mixed random cases use 256 regions and a writer cycling prepare-unmount, commit-unmount, and mount in separate lock scopes, with at least 1 ms between cycle starts. Each paired run alternates model/process order. `Mixed = 1` denotes the lifecycle writer.

#### Independently built main vs Pool

**Local-first:** 54 paired-experiment runs, zero failed requests.

| Segments | Threads | Replicas | Mixed | Main req/s | Pool req/s | Change | Main P99 µs | Pool P99 µs |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 16 | 1 | 3 | 0 | 397,705 | 611,880 | +53.9% | 2.58 | 1.70 |
| 16 | 8 | 3 | 0 | 497,219 | 497,311 | +0.0% | 36.32 | 36.16 |
| 16 | 32 | 3 | 0 | 447,632 | 442,510 | -1.1% | 167.36 | 168.19 |
| 256 | 1 | 3 | 0 | 102,066 | 531,469 | +420.7% | 11.00 | 2.31 |
| 256 | 8 | 3 | 0 | 411,053 | 505,415 | +23.0% | 37.00 | 35.88 |
| 256 | 32 | 3 | 0 | 454,893 | 443,225 | -2.6% | 153.61 | 166.98 |
| 2048 | 1 | 3 | 0 | 13,159 | 247,760 | +1782.8% | 83.35 | 4.15 |
| 2048 | 8 | 3 | 0 | 99,882 | 494,461 | +395.0% | 92.62 | 34.08 |
| 2048 | 32 | 3 | 0 | 341,170 | 460,020 | +34.8% | 127.70 | 159.43 |

**Random:** 126 runs, zero failed requests. Pure-allocation throughput changes range from −0.6% to +6.0%; mixed cases range from +0.8% to +2.5%.

<details>
<summary>Full random-placement comparison with main</summary>

| Segments | Threads | Replicas | Mixed | Main req/s | Pool req/s | Change | Main P99 µs | Pool P99 µs |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 16 | 1 | 1 | 0 | 2,295,800 | 2,280,980 | -0.6% | 2.01 | 2.05 |
| 16 | 1 | 3 | 0 | 820,583 | 857,031 | +4.4% | 2.79 | 2.78 |
| 16 | 8 | 1 | 0 | 3,357,790 | 3,370,250 | +0.4% | 10.19 | 10.09 |
| 16 | 8 | 3 | 0 | 1,066,710 | 1,084,200 | +1.6% | 20.54 | 20.99 |
| 16 | 32 | 1 | 0 | 4,105,600 | 4,232,250 | +3.1% | 29.00 | 29.59 |
| 16 | 32 | 3 | 0 | 1,478,180 | 1,490,780 | +0.9% | 52.77 | 54.84 |
| 256 | 1 | 1 | 0 | 2,056,450 | 2,090,850 | +1.7% | 2.06 | 2.09 |
| 256 | 1 | 3 | 0 | 681,768 | 709,665 | +4.1% | 3.08 | 3.04 |
| 256 | 1 | 3 | 1 | 476,990 | 482,431 | +1.1% | 5.23 | 5.23 |
| 256 | 8 | 1 | 0 | 4,881,070 | 4,934,790 | +1.1% | 4.11 | 4.20 |
| 256 | 8 | 3 | 0 | 1,803,880 | 1,806,730 | +0.2% | 10.08 | 10.34 |
| 256 | 8 | 3 | 1 | 1,680,520 | 1,694,530 | +0.8% | 11.09 | 11.14 |
| 256 | 32 | 1 | 0 | 6,885,120 | 7,297,130 | +6.0% | 11.05 | 10.55 |
| 256 | 32 | 3 | 0 | 2,688,150 | 2,723,590 | +1.3% | 21.94 | 21.60 |
| 256 | 32 | 3 | 1 | 2,668,020 | 2,734,600 | +2.5% | 21.82 | 21.66 |
| 2048 | 1 | 1 | 0 | 1,455,050 | 1,517,170 | +4.3% | 2.28 | 2.29 |
| 2048 | 1 | 3 | 0 | 558,856 | 559,500 | +0.1% | 3.41 | 3.46 |
| 2048 | 8 | 1 | 0 | 4,607,630 | 4,817,360 | +4.6% | 3.61 | 3.60 |
| 2048 | 8 | 3 | 0 | 1,838,210 | 1,835,630 | -0.1% | 8.04 | 8.19 |
| 2048 | 32 | 1 | 0 | 7,142,650 | 7,423,080 | +3.9% | 8.69 | 8.89 |
| 2048 | 32 | 3 | 0 | 2,856,770 | 2,885,340 | +1.0% | 18.64 | 18.83 |

</details>

At 2,048 regions / one worker, local-first throughput was 13,159 → 247,760 req/s and P99 was 83.35 → 4.15 µs. At 32 workers, throughput increased but P99 rose from 127.70 to 159.43 µs. The 256-region/32-worker case measured −2.6% throughput. These results do not establish a non-regression for every workload or latency metric.

These are allocation-component measurements, excluding RPC, payload I/O, retained objects, fragmentation/exhaustion, CacheLib, and CXL performance. Pinning does not isolate CPUs; small differences on the shared host are not a universal performance guarantee. The temporary benchmark source and target have been removed from the PR.

## Checklist

- [x] Code changes reviewed during development
- [x] Changed code formatted with the project formatting hook
- [x] PR-scoped pre-commit checks pass
- [x] Lifecycle, allocation, and resource tests cover the changes
- [x] RFC #3360 tracks the architecture (>500 LOC)

## AI Assistance Disclosure

- [ ] No AI tools were used
- [x] AI tools were used: OpenAI Codex assisted with implementation, regression tests, benchmark execution/analysis, and this description. The human submitter is responsible for reviewing and defending the change.


Icedcoco:
Could we revisit the error returned by EraseUnmountedRegion() here?

At the moment, a client ID mismatch, an unexpected lifecycle status, and a missing driver/resource all return ErrorCode::INTERNAL_ERROR. Would it make sense to distinguish these cases—for example, INVALID_PARAMS for a client mismatch, UNAVAILABLE_IN_CURRENT_STATUS for a stale or unexpected transaction state, and INTERNAL_ERROR only when the driver/resource is actually missing?

Icedcoco:
Thanks for the follow-up changes. I noticed two points that may be worth revisiting:

1. In EraseUnmountedRegion(), a client ID mismatch, an unexpected lifecycle status, and a missing driver/resource still all return ErrorCode::INTERNAL_ERROR. Would it make sense to distinguish these cases—for example, INVALID_PARAMS, UNAVAILABLE_IN_CURRENT_STATUS, and INTERNAL_ERROR, respectively? This would help callers tell stale or invalid requests apart from genuine internal failures.

2. PlacementIndex::active_entries() indexes entries_by_kind_ without validating the enum value, while the other index methods perform bounds checks. Could we either add a guard here or document and enforce the valid-enum precondition consistently?

Would you be open to addressing these points, perhaps with focused regression tests for the lifecycle error paths?

Icedcoco (approved):
LGTM, thanks for addressing the lifecycle error classification and adding the regression coverage.

One minor suggestion: KindIndex() currently relies on assert(), which is disabled in NDEBUG builds. If invalid enum values could ever reach this path, a runtime guard might provide stronger protection in release builds. This is only a minor follow-up suggestion and does not block approval.

Aionw:
@Icedcoco We use debug assert intentionally, enum value used here in controlled scenario not something user input, it's safe here to do so.

ykwd (approved):

PR #3706, opened 2026-08-26
[Store] Add fork-safe SegmentPool snapshot codecs

Aionw:
## Description

Part of #3363 and tracked by #3360. This PR is stacked on `codex/segment-pool-mechanism`.

This is layer 4 of the SegmentPool stack. It adds snapshot boundaries before the final `MasterService` cutover:

- separates allocator snapshot encoding from the mounted-region catalog
- adds an owning `SegmentPoolSnapshot` captured before the child process reads state, so encoding never depends on inherited locks or the source pool's lifetime
- validates complete allocator payloads before adoption
- transfers restored allocators through the driver adoption transaction
- preserves the existing `ma`/`an`/`ms`/`cs`/`ld` wire shape
- preserves CXL snapshot unsupported behavior, LocalSSD handling, NoF exclusion, and detached temporary-reader metrics

`MasterService` remains unchanged in this layer.

## Module

- [ ] Other

## Type of Change

- [x] Refactor
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement

## How Has This Been Tested?

```bash
/tmp/mooncake-region-driver-build/mooncake-store/tests/segment_pool_test
/tmp/mooncake-region-driver-build/mooncake-store/tests/master_snapshot_codec_test
git diff --check codex/segment-pool-mechanism..HEAD
```

- [x] Unit tests pass (`segment_pool_test`, `store_resource_snapshot_codec_test`, `buffer_allocator_test`, `offset_allocator_test`)
- [x] Snapshot codec validation passes

## Checklist

- [x] I have performed a self-review
- [x] Formatting and touched-file pre-commit pass
- [x] Snapshot wire compatibility is preserved
- [x] Restore validation and allocator-adoption tests are included
- [x] RFC #3360 tracks the change

## AI Assistance Disclosure

- [ ] No AI tools were used


Icedcoco:
Thanks for putting this together. The owning, detached snapshot design and the fork coverage look useful for the upcoming MasterService cutover. I had two questions about restore validation that may be worth discussing before merging:

1. Allocator state consistency

   In OffsetAllocator::Restore(), the validation currently checks the layout pointer, multiplier, capacity, and allocated_size, but does not verify that:
    - allocated_num matches the number of used nodes;
    - allocated_size matches the occupied layout;
    - node, bin, neighbor, and free-list indices are internally consistent.

   Could a malformed snapshot therefore be accepted and later cause inconsistent allocation metrics or failures during allocation/deallocation? Since the PR description mentions validating the complete allocator payload before adoption, would it make sense to add these consistency checks, or is the snapshot source considered fully trusted at this layer?

2. Allocation size during deserialization

   In Serializer<__Allocator>::deserialize(), the allocator is constructed using current_capacity and max_capacity before the decompressed node payload size is validated. If a corrupted or unexpectedly large snapshot contains an excessive capacity, this could trigger a large allocation before the input is rejected.

   Would it be safer to validate the capacity relationship and impose a reasonable upper bound before constructing __Allocator? If snapshot data is guaranteed to come only from a trusted internal writer, I would also appreciate documenting that assumption.

These concerns do not change my view that the fork-safe snapshot boundary is useful. I would be happy to re-review once we clarify whether these checks belong here or are intentionally deferred to a later restore-validation change.

Icedcoco:
Thanks for the follow-up changes. The snapshot validation is much more comprehensive now, especially the checks for node topology, accounting, and bounded decompression.

One point I wanted to clarify is the difference between the two allocator restore paths. The new msgpack snapshot path goes through OffsetAllocatorSnapshot::Validate(), while the legacy binary path used by deserialize_from<OffsetAllocator>() still appears to perform only header and capacity checks before installing the deserialized layout.

Is this difference intentional because the legacy binary metadata is considered fully trusted, or should both paths eventually share the same validation semantics? I may be missing an architectural constraint here, so I would appreciate your thoughts on whether this is within the scope of this PR or better handled separately.

Aside from this question, the latest validation changes address the concerns I raised previously and make the snapshot restore boundary substantially clearer.

Aionw:
Yes, that’s correct. The legacy binary restore path is intentionally preserved for compatibility and to avoid changing existing behavior in this PR. Once the refactoring is complete, we’ll migrate its callers to the new implementation with full snapshot validation and remove the legacy path.

Icedcoco (approved):

ykwd (approved):

PR #3710, opened 2026-08-26
[TE] Increase TCP queued transfer admission thresholds

alogfans:
## Description

This PR raises the classic TCP transport's default per-peer queued transfer and pending admission limits.

Previously, TCP used:
- `kDefaultQueuedTransfersPerPeer = 1024`
- `kMaxQueuedTransfersPerPeer = 65535`
- `kDefaultPendingAdmissionsPerPeer = 1024`
- `kMaxPendingAdmissionsPerPeer = 65535`

This change increases them to:
- default queued transfers per peer: `65535`
- max queued transfers per peer: `1048576`
- default pending admissions per peer: `65535`
- max pending admissions per peer: `1048576`

The intent is to reduce premature `QUEUE_FULL` hard rejections under high TCP transfer fanout while preserving bounded admission behavior.

Note: this branch currently also adds `extern/yalantinglibs` as a new gitlink without a `.gitmodules` update. Please confirm whether that submodule pointer is intentional before merging.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [x] Performance improvement
- [ ] Other

## How Has This Been Tested?

Not run yet.

**Test commands:**
```bash
# Not run
```

**Test results:**
- [ ] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

AI assistance was used to inspect the TCP admission behavior and draft this PR description. The human submitter should review and validate all changed lines before merge.

### DIFF
--- a/mooncake-transfer-engine/src/transport/tcp_transport/tcp_transport.cpp
+++ b/mooncake-transfer-engine/src/transport/tcp_transport/tcp_transport.cpp
@@ -300,10 +300,10 @@ TcpTransport::TcpTransport()
     }
 
     constexpr size_t kDefaultLanesPerPeer = 4;
-    constexpr size_t kDefaultQueuedTransfersPerPeer = 1024;
-    constexpr size_t kMaxQueuedTransfersPerPeer = 65535;
-    constexpr size_t kDefaultPendingAdmissionsPerPeer = 1024;
-    constexpr size_t kMaxPendingAdmissionsPerPeer = 65535;
+    constexpr size_t kDefaultQueuedTransfersPerPeer = 65535;
+    constexpr size_t kMaxQueuedTransfersPerPeer = 1048576;
+    constexpr size_t kDefaultPendingAdmissionsPerPeer = 65535;
+    constexpr size_t kMaxPendingAdmissionsPerPeer = 1048576;
     constexpr size_t kDefaultAdmissionTimeoutMs = 1000;
     constexpr size_t kMaxAdmissionTimeoutMs = 600000;
 


catyans (changes requested):
Raising both default per-peer limits from 1,024 to 65,535 is not safe without a byte-based bound or load evidence. `PeerConnectionGroup` retains each `TcpWorkItem` and its associated task/source-buffer lifetime, while the code already notes that the item limit does not bound queued bytes and that a connected payload can stall a lane indefinitely. With this change, one slow or stalled peer can retain 65,535 queued items plus another 65,535 pending admissions by default, turning an early `QUEUE_FULL` signal into potentially very large memory retention and extreme tail latency. The configured maximum of 1,048,576 makes the same failure mode substantially worse.

Please keep the conservative default until admission is also bounded by queued bytes, and add a sustained slow/stalled-peer benchmark reporting queue depth, queued bytes/RSS, rejection rate, and tail latency before increasing it. The PR currently says it has not been tested, so there is no evidence that this 64x increase improves throughput without destabilizing overload behavior.

alogfans:
@catyans Thanks for raising this. I agree that a byte-based admission bound is the proper long-term solution, and this change is not intended to establish the final overload policy.

This is a temporary operational adjustment because the existing 1,024-item threshold is too small for normal TCP workloads: it triggers `QUEUE_FULL` before the TCP transport can maintain its normal pipeline depth, preventing expected throughput even without an abnormal slow or stalled peer. The increase is intended to remove that immediate item-count bottleneck while we implement more appropriate byte-based accounting and collect queue-depth/RSS/tail-latency data.

A follow-up should add queued-byte limits and the slow/stalled-peer measurements you suggested; the default can then be recalibrated based on those results.

catyans (approved):
Thanks for clarifying the intended scope. Re-reviewed the current head: this only adjusts the existing item-count defaults and maxima; it does not change queue ownership, admission timeout, or `QUEUE_FULL` semantics. Given that this is an operational stopgap and all checks are green, LGTM.

suggestion (non-blocking): please follow through with queued-byte admission and slow/stalled-peer measurements covering queue depth, RSS, rejection rate, and tail latency. The larger configurable maxima make those resource bounds and observability especially important.

staryxchen (approved):
LGTM. But I think it's better leaving default value not change.

alogfans:
@staryxchen In some stress workload (e.g. running tebench), the default value is relatively small. Note: If some users report this, they can change the config file to increase the default value in the release.

PR #3711, opened 2026-08-26
[Store] Self-heal dangling LOCAL_DISK replicas in Client::Put

### LINKED ISSUE #3709: [Bug]: store.remove_all break consistency when SSD offload is enabled
### Bug Report

Bug founded in 0.3.12.post1 :  RemoveAll leaves master metadata for lease-protected keys while physically wiping their SSD files, permanently wedging the key.

When RemoveAll (non-forced) is invoked within 5 seconds of a key being read, the master retains that key's metadata (lease not yet expired), but the client-side storage_backend_->RemoveAll() unconditionally deletes all local SSD offload files. The result is a dangling LOCAL_DISK replica: master believes the key exists with a completed disk replica, while the SSD file is already gone (0 bytes).
Afterwards:
- Reads of that key fail with INVALID_KEY (engine sees an L3 miss).
- Writes to that key are permanently rejected with OBJECT_ALREADY_EXISTS, and — worse — Client::Put silently converts this error into success (return {}), so the engine believes the write succeeded while no data was ever stored. The key can never be re-written or recovered, and cache hit rate for it stays at 0 forever.

Should mooncake-master determine which keys to remove from the local SSD (as it knows how to maintain consistency), instead of delete everything? Or at least allow the next actual write to overwrite it when this inconsistency happens?

### Before submitting...

he-yufeng:
## Status

Mitigation, not the root fix. This PR heals the divergence on the Put and BatchPut paths so a wedged key becomes writable again; the RemoveAll/broadcast boundary redesign is tracked in #3715. Read-only keys are not covered by the heal and stay inconsistent until rewritten, by design (a read that evicts would change read semantics).

Fixes #3709.

A non-forced RemoveAll (or the master mailbox broadcast) physically wipes client SSD files without consulting lease state, while master keeps lease-protected metadata. The result is a completed LOCAL_DISK entry whose backing file is gone: reads fail, and every later Put for the key hits `OBJECT_ALREADY_EXISTS`, which `Client::Put` converted into a silent success. The engine then believed data was stored when nothing was ever written, and the key could never be re-written.

When `PutStart` reports `OBJECT_ALREADY_EXISTS` and every completed replica is a LOCAL_DISK replica owned by this client, the client asks the offload file storage whether the backing file still exists (`FileStorage::Exists`): a proven-gone answer evicts the dangling replica through the existing `EvictDiskReplica` RPC and retries `PutStart` once. Any other outcome (file present, or an unknown answer from transport or config trouble) keeps the old idempotent already-exists path, so a healthy replica can never be evicted by mistake. `StartBatchPut` covers the same gap on the batch path: the `OBJECT_ALREADY_EXISTS` subset goes through one batched replica-list query and one batched evict, and only the evicted keys are retried, so a healthy idempotent batch pays one extra RPC and no evictions.

## Module

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

**Test commands:**
```bash
ctest -R file_storage_test --output-on-failure
```

**Test results:**
- [ ] Unit tests pass — new regression test `PutAfterPhysicalWipeHealsDanglingLocalDiskReplica` builds the exact divergence: offload a key so master tracks a LOCAL_DISK replica, wipe the SSD file physically, then Put the same key and assert the write really lands and the new data reads back. The batch twin `BatchPutAfterPhysicalWipeHealsDanglingLocalDiskReplica` wipes two offloaded keys and batches them with a fresh key, then asserts every key really stores and reads back the new data. It needs the store test build, which does not compile on this machine (no rdma-core / Linux headers on macOS), so it runs on CI. I verified the touched code paths by close read against the in-tree type definitions; no local build was possible.
- [ ] Integration tests pass (if applicable) — runs on CI, see above.

## Checklist

## AI Assistance Disclosure

Prepared with Kimi K3 assistance. The divergence mechanism (both physical-wipe paths ignore lease state, the silent already-exists success, and the heal-via-EvictDiskReplica route using only existing RPCs) was re-verified against the source before submission.

### DIFF
--- a/mooncake-store/src/client_service.cpp
+++ b/mooncake-store/src/client_service.cpp
@@ -2017,32 +2017,18 @@ bool Client::healDanglingLocalDiskReplica(const ObjectKey& key) {
         // still serve reads, so there is nothing for this client to heal.
         return false;
     }
-    if (!local_disk || !transfer_submitter_) {
+    if (!local_disk || !local_disk_probe_fn_) {
         return false;
     }
     // RemoveAll wipes client SSD files while master keeps the metadata
     // (issue #3709), leaving a completed entry whose backing file is gone.
-    // Reporting success for a Put on top of it would store nothing. Probe the
-    // last byte through the normal read path: a file-gone error proves the
-    // replica is dangling. Anything else (readable, or a transport error)
-    // keeps the old idempotent path so we never evict a healthy replica.
-    const uint64_t object_size =
-        local_disk->get_local_disk_descriptor().object_size;
-    if (object_size == 0) {
-        return false;
-    }
-    char probe = 0;
-    std::vector<Slice> probe_slice{Slice{&probe, 1}};
-    const ErrorCode probe_err =
-        TransferReadRange(*local_disk, probe_slice, object_size - 1);
-    // The offload read path reports a wiped SSD file as INVALID_KEY (the
-    // storage key is gone from the backend), while other backends surface
-    // OBJECT_NOT_FOUND / FILE_OPEN_FAIL. All three prove the file is gone;
-    // transport/config failures come back as different codes, so a healthy
-    // replica still can never be evicted by mistake.
-    if (probe_err != ErrorCode::OBJECT_NOT_FOUND &&
-        probe_err != ErrorCode::FILE_OPEN_FAIL &&
-        probe_err != ErrorCode::INVALID_KEY) {
+    // Reporting success for a Put on top of it would store nothing. Ask the
+    // installed probe (the offload file storage owned by RealClient) whether
+    // the file is still there. Only a proven-gone answer evicts; present or
+    // unknown (transport/config trouble) keeps the old idempotent path, so a
+    // healthy replica can never be evicted by mistake.
+    const std::optional<bool> file_gone = local_disk_probe_fn_(key);
+    if (file_gone != true) {
         return false;
     }
     auto evicted =
--- a/mooncake-store/src/client_service.cpp
+++ b/mooncake-store/src/client_service.cpp
@@ -1880,6 +1880,11 @@ tl::expected<void, ErrorCode> Client::Put(const ObjectKey& key,
 
     // Start put operation
     auto start_result = master_client_.PutStart(key, slice_lengths, client_cfg);
+    if (!start_result &&
+        start_result.error() == ErrorCode::OBJECT_ALREADY_EXISTS &&
+        healDanglingLocalDiskReplica(key)) {
+        start_result = master_client_.PutStart(key, slice_lengths, client_cfg);
+    }
     if (!start_result) {
         ErrorCode err = start_result.error();
         if (err == ErrorCode::OBJECT_ALREADY_EXISTS) {
@@ -1991,6 +1996,47 @@ tl::expected<void, ErrorCode> Client::Put(const ObjectKey& key,
     }
 
     return {};
+}
+
+bool Client::healDanglingLocalDiskReplica(const ObjectKey& key) {
+    auto replicas = master_client_.GetReplicaList(key);
+    if (!replicas) {
+        return false;
+    }
+    bool has_completed_local_disk = false;
+    for (const auto& replica : replicas->replicas) {
+        if (replica.status != ReplicaStatus::COMPLETE) {
+            continue;
+        }
+        if (replica.is_local_disk_replica() &&
+            replica.get_local_disk_descriptor().client_id == client_id_) {
+            has_completed_local_disk = true;
+            continue;
+        }
+        // A completed replica owned elsewhere (memory, remote disk, DFS) can
+        // still serve reads, so there is nothing for this client to heal.
+        return false;
+    }
+    if (!has_completed_local_disk || !storage_backend_) {
+        return false;
+    }
+    // RemoveAll wipes client SSD files while master keeps the metadata
+    // (issue #3709), leaving a completed entry whose backing file is gone.
+    // Reporting success for a Put on top of it would store nothing.
+    auto exists = storage_backend_->IsExist(key);
+    if (!exists || *exists) {
+        return false;
+    }
+    auto evicted =
+        master_client_.EvictDiskReplica(key, ReplicaType::LOCAL_DISK);
+    if (!evicted) {
+        LOG(WARNING) << "Failed to evict dangling LOCAL_DISK replica for key="
+                     << key << ": " << toString(evicted.error());
+        return false;
+    }
+    LOG(WARNING) << "Evicted dangling LOCAL_DISK replica for key=" << key
+                 << " (backing file already gone)";
+    return true;
 }
 
 tl::expected<void, ErrorCode> Client::Upsert(const ObjectKey& key,
--- a/mooncake-store/src/client_service.cpp
+++ b/mooncake-store/src/client_service.cpp
@@ -2035,8 +2035,14 @@ bool Client::healDanglingLocalDiskReplica(const ObjectKey& key) {
     std::vector<Slice> probe_slice{Slice{&probe, 1}};
     const ErrorCode probe_err =
         TransferReadRange(*local_disk, probe_slice, object_size - 1);
+    // The offload read path reports a wiped SSD file as INVALID_KEY (the
+    // storage key is gone from the backend), while other backends surface
+    // OBJECT_NOT_FOUND / FILE_OPEN_FAIL. All three prove the file is gone;
+    // transport/config failures come back as different codes, so a healthy
+    // replica still can never be evicted by mistake.
     if (probe_err != ErrorCode::OBJECT_NOT_FOUND &&
-        probe_err != ErrorCode::FILE_OPEN_FAIL) {
+        probe_err != ErrorCode::FILE_OPEN_FAIL &&
+        probe_err != ErrorCode::INVALID_KEY) {
         return false;
     }
     auto evicted =
--- a/mooncake-store/src/client_service.cpp
+++ b/mooncake-store/src/client_service.cpp
@@ -2003,28 +2003,40 @@ bool Client::healDanglingLocalDiskReplica(const ObjectKey& key) {
     if (!replicas) {
         return false;
     }
-    bool has_completed_local_disk = false;
+    const Replica::Descriptor* local_disk = nullptr;
     for (const auto& replica : replicas->replicas) {
         if (replica.status != ReplicaStatus::COMPLETE) {
             continue;
         }
         if (replica.is_local_disk_replica() &&
             replica.get_local_disk_descriptor().client_id == client_id_) {
-            has_completed_local_disk = true;
+            local_disk = &replica;
             continue;
         }
         // A completed replica owned elsewhere (memory, remote disk, DFS) can
         // still serve reads, so there is nothing for this client to heal.
         return false;
     }
-    if (!has_completed_local_disk || !storage_backend_) {
+    if (!local_disk || !transfer_submitter_) {
         return false;
     }
     // RemoveAll wipes client SSD files while master keeps the metadata
     // (issue #3709), leaving a completed entry whose backing file is gone.
-    // Reporting success for a Put on top of it would store nothing.
-    auto exists = storage_backend_->IsExist(key);
-    if (!exists || *exists) {
+    // Reporting success for a Put on top of it would store nothing. Probe the
+    // last byte through the normal read path: a file-gone error proves the
+    // replica is dangling. Anything else (readable, or a transport error)
+    // keeps the old idempotent path so we never evict a healthy replica.
+    const uint64_t object_size =
+        local_disk->get_local_disk_descriptor().object_size;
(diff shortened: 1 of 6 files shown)


mooncake-store/tests/file_storage_test.cpp
```diff
@@ -870,4 +871,118 @@ TEST_F(FileStorageTest, NullSsdMetricDoesNotCrash) {
...
+                                          "cpu:0", false, false)
+                    .has_value());
+
+    FileStorageConfig config = FileStorageConfig::FromEnvironment();
+    config.storage_backend_type = StorageBackendType::kFilePerKey;
+    config.storage_filepath = data_path + "/heal_ssd";
+    config.local_buffer_size = 4 * 1024 * 1024;
+    fs::create_directories(config.storage_filepath);
+    FileStorage file_storage(config, client.value(), local_rpc_addr);
+    ASSERT_TRUE(file_storage.storage_backend_->Init());
```
fcczzz: Thanks for adding this regression test. I tried building file_storage_test locally, but it doesn’t compile on my machine. The TEST_F body directly accesses several private FileStorage members; it looks like friend class FileStorageTest applies only to the fixture itself, not the generated test subclass. I also encountered a deleted assignment error at query = client.value()->Query(key). Could you please take a look? Using fixture helper methods for the private accesses and a new variable for the second query may resolve these errors.

fcczzz (commented):
Thanks for working on this. I understand that reusing the existing RPCs keeps the change relatively small. I’m wondering, though, whether the consistency issue would be better addressed at the RemoveAll boundary. The master knows which keys were actually removed after checking their leases, while clients currently wipe all SSD files unconditionally. Healing during a later Put leaves the metadata and files inconsistent until the same key is written again, and it does not cover read-only or batch-write paths.

Would it be safer for the master to return or enqueue the exact keys that each client should delete, or at least avoid the full physical wipe for a non-forced RemoveAll? I understand that this would require an RPC contract change, so if that is intentionally out of scope, perhaps this PR could be described as a mitigation while the root fix remains tracked separately.

mooncake-store/tests/file_storage_test.cpp
fcczzz: Thanks for adding this regression test. I tried building `file_storage_test` locally, but it doesn’t compile on my machine. The `TEST_F` body directly accesses several private `FileStorage` members; it looks like `friend class FileStorageTest` applies only to the fixture itself, not the generated test subclass. I also encountered a deleted assignment error at `query = client.value()->Query(key)`. Could you please take a look? Using fixture helper methods for the private accesses and a new variable for the second query may resolve these errors.

mooncake-store/src/client_service.cpp
```diff
@@ -1993,6 +1998,59 @@ tl::expected<void, ErrorCode> Client::Put(const ObjectKey& key,
...
+    // keeps the old idempotent path so we never evict a healthy replica.
+    const uint64_t object_size =
+        local_disk->get_local_disk_descriptor().object_size;
+    if (object_size == 0) {
+        return false;
+    }
+    char probe = 0;
+    std::vector<Slice> probe_slice{Slice{&probe, 1}};
+    const ErrorCode probe_err =
+        TransferReadRange(*local_disk, probe_slice, object_size - 1);
```
fcczzz: `TransferReadRange()` ultimately calls `TransferSubmitter::submitRangeRead()`, which explicitly rejects both DISK and LOCAL_DISK replicas and returns `nullopt`. That is converted to `TRANSFER_FAIL`, so this probe can never observe `OBJECT_NOT_FOUND` or `FILE_OPEN_FAIL`; the helper always returns `false` and the dangling replica is never healed. Could we use a read/existence path that actually supports LOCAL_DISK and ensure the new regression test is run in CI?

he-yufeng:
Fair push, and I agree the boundary is where the root fix belongs. Filed #3715 to track it, with the three shapes I see and their costs. Short version of why I scoped here first:

The return-exact-keys direction stops the divergence at creation, but it is an RPC contract change (your point), and the offline-client variant means durable per-client queue state on the master that also has to replicate in HA. The mailbox broadcast has no response channel at all, so a return value alone never covers that second wipe path. And any boundary design still races a concurrent Put landing between the lease check and the file delete, which is why a convergent heal stays necessary even after the root fix.

Your read-only point is the honest gap in this PR: a key that is only ever read keeps failing reads until something rewrites it. Healing on read would close that, but it turns a read into an eviction, and changing read semantics felt worse than documenting the gap for a first step.

Happy to take #3715 next if the direction looks right to you. In the meantime I have re-labeled this PR as mitigation in the description, root fix tracked there.

fanganpai:
Thanks for prompt fix. 
However, it still doesn't seem to resolve the original issue. 
The key has a unique local_disk_replica record on the master but the value doesn't actually exist on the SSD, and it can't be read. Even after performing another PUT operation and attempting to read again, the read still fails.
Am I missing something?

```python
>>> store.get_replica_desc("b75f6a2965e1aa010a007901bd2732302ba64a13a84a5b233a1a2cebdfe1f68a__k")
[<mooncake.store.ReplicaDescriptor object at 0x7f8ee249bdb0>]
>>> store.get_replica_desc("b75f6a2965e1aa010a007901bd2732302ba64a13a84a5b233a1a2cebdfe1f68a__k")[0].is_local_disk_replica()
True
>>> o=store.get("b75f6a2965e1aa010a007901bd2732302ba64a13a84a5b233a1a2cebdfe1f68a__k")
E20260828 07:31:56.606736   701 real_client.cpp:6964] Failed to invoke batch_get_offload_object, client_addr = ******, error is: INVALID_KEY
E20260828 07:31:56.606770   701 real_client.cpp:6871] Batch get offload object failed with error: INVALID_KEY
E20260828 07:31:56.606776   701 real_client.cpp:2884] SSD read failed for key 'b75f6a2965e1aa010a007901bd2732302ba64a13a84a5b233a1a2cebdfe1f68a__k': INVALID_KEY
>>> store.put("b75f6a2965e1aa010a007901bd2732302ba64a13a84a5b233a1a2cebdfe1f68a__k","test_str".encode('UTF-8'))
0
>>> o=store.get("b75f6a2965e1aa010a007901bd2732302ba64a13a84a5b233a1a2cebdfe1f68a__k")
E20260828 07:32:27.168946   701 real_client.cpp:6964] Failed to invoke batch_get_offload_object, client_addr = ******, error is: INVALID_KEY
E20260828 07:32:27.168980   701 real_client.cpp:6871] Batch get offload object failed with error: INVALID_KEY
E20260828 07:32:27.168987   701 real_client.cpp:2884] SSD read failed for key 'b75f6a2965e1aa010a007901bd2732302ba64a13a84a5b233a1a2cebdfe1f68a__k': INVALID_KEY
```

he-yufeng:
Thank you for the trace, it pinpoints the hole exactly. Your read path reports the wiped file as `INVALID_KEY` (the bucket backend's `BatchLoad` returns that for a missing storage key), and my probe only accepted `OBJECT_NOT_FOUND` / `FILE_OPEN_FAIL`. So the heal never fired, the Put took the old silent already-exists path, and reads kept failing, matching your log line by line.

2f3a3003 accepts `INVALID_KEY` as proof of a gone backing file as well, and the wipe scenario now runs on both backend variants (per-key file backend and bucket backend). The put you ran (`store.put` returning 0) is precisely the silent-success path this closes: with the new probe it evicts the dangling replica and writes for real.

Could you rerun your repro against this head? Same steps as your comment: confirm the Put actually lands and the following `store.get` returns the new value instead of `INVALID_KEY`.

he-yufeng:
Both findings confirmed in the code, thank you. `submitRangeRead` rejects disk and local_disk replicas outright, so the probe always saw TRANSFER_FAIL and the heal could never fire. And the test indeed never compiled anywhere, which means CI never actually ran it.

Reworked in fc6d80c4 along the line you asked for: the existence check no longer goes through the transfer path at all. `Client` keeps the replica scan and the evict-plus-retry orchestration, and the proof is an installed callback (`SetLocalDiskProbe`) that `RealClient` wires to the offload `FileStorage` it owns, the same read surface that fails with `INVALID_KEY` in fanganpai's trace. A new `FileStorage::Exists` answers existence without allocating read buffers: the per-key backend resolves to a filesystem lookup and the bucket backend to its in-memory index, both of which flip to false after a physical wipe. Present or unknown answers keep the old idempotent path, so a healthy replica still cannot be evicted by mistake.

The test compiles now in principle: the private accesses are friended by test name (`FRIEND_TEST`, since `friend class` does not reach the generated subclasses), the `Query` reassignment is gone (`QueryResult` is not assignable), and the scenario runs on both backend variants, per-key and bucket. I cannot build the store suite on this machine (macOS, no rdma-core), so compile plus run ride CI, and if you have the toolchain handy a local `file_storage_test` build would be a welcome second check.

fcczzz:
Hi, CI failed. Could you take a look?

he-yufeng:
Fixed in 9c6359c. The failure was mine: the FRIEND_TEST include pulled gtest/gtest_prod.h into the production file_storage.h, and every non-test target lost its include path for it. It is now a plain friend declaration on the generated test class names, no gtest include anywhere in the header. Format and the wheel builds are re-running on this head; the format lane is already green on it.

he-yufeng:
Format is green now. The second failure was clang-format 20 wanting the over-length friend declarations broken after `friend class`, which my local check could not see (no clang-format 20 on this machine); fixed in 7d31a95 exactly as the CI formatter rewrites them. Remaining lanes are still queued on the runners.

he-yufeng:
One more latent build error surfaced now that the build actually reaches the test target. `RunPutHealWipeScenario` was a free function in an anonymous namespace, but it touches `storage_backend_`, `offloading_mutex_`, and `enable_offloading_`, all private. Friendship was never going to cover it, so the file could never have compiled in that shape; it only got this far unnoticed because the earlier builds died on the gtest include before reaching this target. Fixed in 94f84eb3 by making the scenario a `FileStorageTest` member, the same pattern the other private-touching helpers in this file already use. Verified locally with the repo's `code_format.sh --check --changed-lines` against upstream main (clang-format 20.1.8), which is clean.

he-yufeng:
All lanes are green now, including CTest (31m). The three failures were layered: the gtest include broke non-test targets, then clang-format 20 wanted the long friend declarations broken, and fixing that finally let the build reach the test file, which had a latent private access from a free helper that friendship could never cover. The scenario is now a `FileStorageTest` member like the other helpers. Ready for another look.

fcczzz:
Thanks, I took another look at 94f84eb3, my two original inline issues are resolved.

I noticed that BatchPut still has the same silent-success gap: StartBatchPut does not attempt the heal for per-key OBJECT_ALREADY_EXISTS results, and CollectResults converts those results to success. The new tests use BatchOffload only to prepare the SSD state and ultimately exercise the single-key Put path.

Do you think BatchPut should be covered by this mitigation as well? It is the same correctness issue, but naively invoking the single-key heal for every affected key could introduce significant per-key RPC and filesystem-probe overhead on a batch hot path. The existing batch RPCs may make a batched query/evict/retry implementation possible, though healthy idempotent batches would still need some performance consideration.

Also, the PR description and the comment in client_service.h still describe the discarded “probe the last byte through the normal read path” implementation; could those be updated to describe the FileStorage::Exists callback?

he-yufeng:
Both points are in 03f2a797.

On BatchPut: yes, it is covered now, and the cost profile matches your constraint. The `OBJECT_ALREADY_EXISTS` subset goes through one `BatchGetReplicaList`, the probe runs locally per candidate key (a filesystem existence check, no RPC), and only the proven-gone keys go into one `BatchEvictDiskReplica` plus one retried `BatchPutStart`. A healthy idempotent batch pays exactly one extra batched query and zero evictions, and a batch with no already-exists keys pays nothing at all: the heal pass only runs when at least one key came back `OBJECT_ALREADY_EXISTS`. Per-key retry semantics mirror the single-Put path, including the race case: if the retried start comes back `OBJECT_ALREADY_EXISTS` again, the key falls back to the idempotent skip instead of being forced.

One structural note: `CollectResults` still converts a surviving `OBJECT_ALREADY_EXISTS` to success, which is now the correct terminal state for anything the probe could not prove gone (healthy replica, or transport/config trouble during the heal). The heal only ever changes the outcome for the proven-dangling case, where the old conversion was the bug.

The header comment and the PR description now describe the `FileStorage::Exists` probe instead of the discarded last-byte read. The new batch twin test wipes two offloaded keys and batches them with a fresh key, then asserts all three really store and read back the new data; it runs on CI like the single-key one.

fcczzz (approved):
LGTM

catyans (approved):
LGTM. The single-key and batch paths only evict after all completed replicas are client-owned `LOCAL_DISK` replicas and `FileStorage` proves the backing data is absent; present or unknown probe results preserve the existing idempotent behavior. The batch retry keeps metadata operations batched, and the tests cover per-key, bucket, and mixed-batch cases. The broader `RemoveAll` boundary fix remains tracked in #3715. CI is green (21 passed, 1 skipped).

ykwd (approved):

PR #3008, opened 2026-07-20
[Store] Add Alibaba Cloud OSS filesystem adapter

stmatengss:
## Description

Add an Alibaba Cloud OSS implementation of `FileSystemAdapter` for Mooncake Store's distributed storage backend.

The adapter uses the OSS REST API with V4 HMAC-SHA256 request signing and supports object writes, reads, range reads, vector I/O, existence and size checks, deletion, and paginated directory-style listing. Static access keys and optional STS security tokens are read from `MOONCAKE_OSS_*` environment variables.

`DistributedStorageBackend` now recognizes `MOONCAKE_DISTRIBUTED_FS_TYPE=oss` and skips POSIX directory creation for flat object-store namespaces. The build enables OSS support automatically when libcurl and OpenSSL are available.

No overlapping open issue or pull request was found for an OSS `FileSystemAdapter`.

## Module

- [ ] Transfer Engine (`mooncake-transfer-engine`)
- [x] Mooncake Store (`mooncake-store`)
- [ ] Mooncake EP (`mooncake-ep`)
- [ ] Mooncake PG (`mooncake-pg`)
- [ ] Integration (`mooncake-integration`)
- [ ] P2P Store (`mooncake-p2p-store`)
- [ ] Python Wheel (`mooncake-wheel`)
- [ ] Common (`mooncake-common`)
- [ ] Mooncake RL (`mooncake-rl`)
- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

The test runs against a local OSS protocol test server and validates V4-authenticated PUT, GET, Range GET, HEAD, DELETE, ListObjectsV2, vector I/O, adapter selection through `CreateStorageBackend`, and the distributed backend health check.

**Test commands:**

```bash
cmake -S . -B build-oss-pr -DBUILD_UNIT_TESTS=ON
cmake --build build-oss-pr --target oss_adapter_test -j8
ctest --test-dir build-oss-pr -R '^oss_adapter_test$' --output-on-failure
SKIP=cmake-format pre-commit run --files <changed-files>
```

**Test results:**

- [x] Unit tests pass
- [x] Integration tests pass
- [x] Manual testing done on `root@mateng04-j`

## Checklist

- [x] I have performed a self-review of my own code
[...]

PR #3714, opened 2026-08-27
[Store] Add object storage adapter and OSS integration

fcczzz:
## Description

This PR ports the Alibaba Cloud OSS implementation from #3008 onto the
object-storage abstraction introduced by #3043 and updates it for the current
`main` branch.

Unlike #3008, this version does not model OSS as a filesystem. It implements
`ObjectStorageAdapter`, so `DistributedStorageBackend` keeps the two storage
models separate:

```text
DistributedStorageBackend
├── FileSystemAdapter       shard / descriptor / offset semantics
└── ObjectStorageAdapter    opaque logical key / whole-object semantics
```

The OSS adapter uses the OSS REST API with V4 HMAC-SHA256 request signing and
supports:

- contiguous and vectored object uploads;
- full and ranged reads;
- existence and size checks;
- object deletion;
- paginated ListObjectsV2 scans; and
- optional end-to-end startup health checks.

Logical Mooncake keys are reversibly encoded under the configured object
prefix. OSS credentials and endpoint settings are read from `MOONCAKE_OSS_*`
environment variables, with the compatibility fallbacks retained from #3008.
The build enables the adapter when libcurl and OpenSSL are available.

The storage factory recognizes `MOONCAKE_DFS_FS_ADAPTER=oss` (with
`MOONCAKE_DISTRIBUTED_FS_TYPE` as the existing fallback). `FileStorage`
selects its control plane from the constructed backend mode: filesystem-backed
distributed storage continues to use DFS, while OSS uses the regular offload
control path.

### Relationship to existing work

- #3008 contains the original provider-specific OSS implementation. This PR
  ports that implementation rather than developing a second OSS client.
- #3043 established the object-storage boundary used here and explicitly
  reserved the OSS provider integration for a follow-up.
- This PR does not change the DFS shard/offset data path or the file-per-key
  storage layout.

### Validation status and open items

The adapter, backend integration, pagination, health-check flow, opaque-key
round trips, and DFS/object-mode selection are covered by local tests. A real
OSS lifecycle test is present but was skipped locally because OSS credentials
are not available in the development environment.

Before marking the PR ready for merge, the provider-specific behavior should
be validated on the exact PR head by the #3008 author or another reviewer with
access to an OSS environment. Review should also confirm the supported encoded
key-length contract, ensure internal health probes cannot appear in `ListKeys`,
and decide whether `CheckHealth` should retain source compatibility through a
default implementation.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] Refactor
- [x] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

`ObjectStorageAdapter::CheckHealth` is currently added as a pure virtual
method, so out-of-tree implementations must implement it. This should be
changed to a default implementation if source compatibility is required.

## How Has This Been Tested?

**Test commands:**

```bash
nproc=$(nproc)
cmake --build build \
  --target oss_adapter_test object_storage_adapter_test \
           file_storage_test file_storage_config_test \
  -j"$nproc"

  -R '^(oss_adapter_test|object_storage_adapter_test|file_storage_test|file_storage_config_test)$'

git diff --check main...HEAD
```

The socket-based tests initially encountered the local sandbox's socket
restriction and passed when rerun with local socket access.

**Test results:**

- [x] Unit tests pass
- [ ] Integration tests pass (real OSS validation has not been run)
- [ ] Manual testing done

The credential-gated real OSS lifecycle case was skipped because no OSS test
environment was available. The mock OSS server tests and the remaining Store
tests passed.

## Checklist

- [ ] I have updated the documentation (required before production-ready OSS support)

No new RFC has been filed. The object-storage architecture was reviewed in
#3043, and the provider implementation originates from #3008.

## AI Assistance Disclosure

- [ ] No AI tools were used

OpenAI Codex assisted with the port, test construction, review, validation,
and preparation of this PR description. The human submitter remains
responsible for reviewing, understanding, and defending every changed line.


mooncake-store/include/storage/distributed/object_storage_adapter.h
```diff
@@ -52,6 +52,10 @@ class ObjectStorageAdapter {
     virtual tl::expected<std::vector<KeyInfo>, ErrorCode> ListKeys() = 0;
 
     virtual tl::expected<void, ErrorCode> Init() = 0;
+
+    // Performs an end-to-end service check after Init(). Implementations may
+    // create a temporary object but must make a best effort to clean it up.
+    virtual tl::expected<void, ErrorCode> CheckHealth() = 0;
     virtual const char* GetName() const = 0;
```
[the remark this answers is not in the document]
fcczzz: Updated CheckHealth() to provide a default implementation returning NOT_SUPPORTED. Existing adapters can continue compiling without an override, while explicitly enabling health checks fails clearly if the adapter does not support them. The OSS implementation retains its end-to-end probe.

mooncake-store/src/storage/distributed/oss_adapter.cpp
```diff
@@ -0,0 +1,620 @@
...
+    curl_slist* headers = nullptr;
+    auto add_header = [&](const std::string& header) {
+        headers = curl_slist_append(headers, header.c_str());
+    };
+
+    const std::string timestamp = OssTimestamp().first;
+    add_header("x-oss-content-sha256: UNSIGNED-PAYLOAD");
+    add_header("x-oss-date: " + timestamp);
+    if (!security_token_.empty())
+        add_header("x-oss-security-token: " + security_token_);
```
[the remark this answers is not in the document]
fcczzz: Updated Init() to trim the security token once, so signing and header construction use the same normalized value. OSS V4 already requires trimming header values during canonicalization, so this is defensive input cleanup rather than a confirmed signature mismatch fix.

mooncake-store/src/storage/distributed/oss_adapter.cpp
```diff
@@ -0,0 +1,620 @@
...
+
+std::string OssObjectStorageAdapter::BuildAuthorization(
+    const std::string& method, const std::string& physical_key,
+    const std::map<std::string, std::string>& query,
+    const std::string& timestamp) const {
+    const std::string date = timestamp.substr(0, 8);
+    const std::string canonical_uri =
+        UriEncode("/" + bucket_ + "/" + physical_key, true);
+    std::string canonical_headers = "x-oss-content-sha256:UNSIGNED-PAYLOAD\n";
+    canonical_headers += "x-oss-date:" + timestamp + "\n";
```
particle128: Could we add x-oss-range-behavior: standard to range GET requests and include this header in V4 signing?

For example, consider a 4 MiB object:
Object:          bytes 0–4194303
Requested range: bytes 2097152–6291455  (4 MiB starting at offset 2 MiB)

By default, OSS may ignore this out-of-bounds range and return 200 OK with the entire 4 MiB object. Both the current status-code and length checks would pass, but the returned data comes from the wrong offset.

With standard, OSS returns 206 Partial Content with the remaining 2 MiB, which the existing length check correctly rejects as a short read.

See the [OSS Range request documentation](https://help.aliyun.com/zh/oss/how-to-obtain-oss-resources-by-segmenting-http-range-requests).

mooncake-store/src/storage/distributed/oss_adapter.cpp
```diff
@@ -0,0 +1,620 @@
...
+    return result;
+}
+
+size_t BodyCallback(char* data, size_t size, size_t count, void* user_data) {
+    auto* body = static_cast<std::string*>(user_data);
+    body->append(data, size * count);
+    return size * count;
+}
+
+size_t HeaderCallback(char* data, size_t size, size_t count, void* user_data) {
```
particle128, line 161: Could we clear the accumulated response headers whenever HeaderCallback receives a new HTTP status line?

libcurl can deliver multiple response header blocks during a single transfer—not just the final response. Examples include informational responses such as 103 Early Hints, authentication negotiation, and proxy CONNECT responses.

Without resetting the header map between response blocks, fields omitted from the final response could retain values from an earlier response. Clearing the map at each new status line would prevent mixing headers from different responses.

See the libcurl documentation for [CURLOPT_HEADERFUNCTION](https://curl.se/libcurl/c/CURLOPT_HEADERFUNCTION.html) and [proxy CONNECT headers](https://curl.se/libcurl/c/CURLOPT_SUPPRESS_CONNECT_HEADERS.html).

image

particle128:
Thanks for adding OSS support! I left two inline comments on range GET behavior and response-header handling. Both should be addressable with small, localized changes: enabling standard OSS range behavior and keeping headers from separate HTTP responses isolated.

mooncake-store/src/storage/distributed/oss_adapter.cpp
```diff
@@ -0,0 +1,621 @@
...
+#include <glog/logging.h>
+
+#include "environ.h"
+
+namespace mooncake {
+namespace {
+
+std::once_flag curl_init_once;
+
+std::string UriEncode(std::string_view value, bool preserve_slash = false) {
```
Aionw, line 30: I think there already some utility to make http request/post much more easier out there, do not write those  url request construct things by yourself.

fcczzz: Thanks. If you mean mooncake::httpGet, it only supports GET and returns the response body for HTTP 200. The OSS adapter also needs custom V4/STS headers, PUT/HEAD/DELETE, range requests, response headers, iovec streaming uploads, and ListObjectsV2 pagination, so httpGet cannot directly replace the request layer.

The full coro_http_client could simplify the HTTP transport, but OSS-specific canonicalization and V4 signing would still be required. Would you prefer using coro_http_client for transport, or the official aliyun-oss-cpp-sdk for the complete OSS protocol handling?

Aionw: coro_http_client for transport. Basically, I prefer use modern cpp http client and json construct instead write them manually.

mooncake-store/src/storage/distributed/oss_adapter.cpp
fcczzz: Good catch. I added x-oss-range-behavior: standard to range requests and included it in the canonical x-oss headers used for V4 signing. GetRange now also requires a 206 response so an ignored Range request cannot be accepted as a successful read.

mooncake-store/src/storage/distributed/oss_adapter.cpp
fcczzz: Agreed. The header callback now clears the accumulated headers whenever a new HTTP status line is received, so only headers from the final response block are retained.

mooncake-store/src/storage/distributed/oss_adapter.cpp
fcczzz: I tried migrating the transport to coro_http_client, but found a hostname-verification issue in the pinned YLT version.

With verify_peer enabled and an explicit hostname supplied, a local TLS test produced these results:
- Trusted CA, matching hostname: accepted.
- Trusted CA, mismatched hostname: also accepted, and the server received the HTTP request.
- Untrusted CA: rejected.

init_ssl() creates the SSL stream before setting the hostname-verification callback on the context. Setting the callback directly on the stream fixes the mismatch case. The current YLT main branch still has the same ordering:
https://github.com/alibaba/yalantinglibs/blob/27b04dd754fa1d9f3ae97dcf48a11ded160f7bf5/include/ylt/standalone/cinatra/coro_http_client.hpp#L258

Would keeping libcurl for this PR be acceptable while we report and resolve this upstream? We can still replace the hand-written XML parsing independently. Alternatively, if you prefer the migration now, we would need to carry a local YLT fix to preserve HTTPS hostname verification.

particle128 (approved):
Thanks for addressing my comments. Both changes look good to me.

ykwd (approved):

PR #3717, opened 2026-08-27
[Bugfix][Store] Register ShmHelper memory with SPDK for NoF zero-copy transfers

### LINKED ISSUE #3736: [Bug]: NoF zero-copy transfer fails with "No translation for ptr" when using MooncakeHostTensorAllocator in SGLang
### Bug Report

When using MooncakeStore as the L3 storage backend for SGLang HiCache with NoF (NVMe-oF) zero-copy transfer enabled, KV cache transfer fails with the following error:

Image

Root Cause
SGLang's MooncakeHostTensorAllocator uses MooncakeHostMemAllocator → ShmHelper (memfd + mmap) to allocate host memory. However, this memory is never registered with SPDK via spdk_mem_register(). SPDK's RDMA transport requires all DMA-capable memory to be pre-registered; otherwise, spdk_rdma_get_translation() fails with No translation for ptr.

### Before submitting...

NUABO:
## Description
fix: https://github.com/kvcache-ai/Mooncake/issues/3736
Implement NOF integration with SGLang, SPDK's RDMA transport can only DMA from SPDK-registered memory. The host pool allocated via ShmHelper (memfd+mmap) was never registered, so NoF (NVMe-oF) put/get passing user buffers directly fails with "spdk_rdma_get_translation: No translation for ptr".

Register the mapping with spdk_mem_register in allocate() (USE_NOF only) and unregister before munmap in free()/cleanup(). Registration failure is non-fatal so non-NoF paths are unaffected. A dedicated ShmSegment::spdk_registered flag tracks SPDK registration, keeping the existing `registered` flag (used by DummyClient for IPC registration) untouched.

This PR fixes an issue where SGLang's MooncakeHostTensorAllocator triggers a No translation for ptr error under Mooncake NoF mode. The allocator uses MooncakeHostMemAllocator → ShmHelper (memfd+mmap), but the allocated memory was not registered with SPDK, so RDMA transport cannot access the buffer. The patch adds spdk_mem_register()/spdk_mem_unregister() to ShmHelper's allocate/free paths to enable NoF zero-copy transfers for SGLang. 

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
```

**Test results:**
- [ ] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

### DIFF
--- a/mooncake-store/src/dummy_client.cpp
+++ b/mooncake-store/src/dummy_client.cpp
@@ -657,7 +657,13 @@ int DummyClient::unregister_device_buffer_for_reconnect(void* buffer) {
 
     auto ret = invoke_rpc<&RealClient::unregister_shm_buffer_internal, void>(
         buffer_addr, client_id_);
-    if (ret.has_value()) {
+    // Same contract as unregister_buffer: the receiver tears the mapping down
+    // even on INTERNAL_ERROR, and INVALID_PARAMS means the server has no such
+    // mapping, so drop the local bookkeeping then; keep it on RPC_FAIL /
+    // RPC_TIMEOUT where the server may not have run.
+    if (ret.has_value() ||
+        (!ret.has_value() && (ret.error() == ErrorCode::INTERNAL_ERROR ||
+                              ret.error() == ErrorCode::INVALID_PARAMS))) {
         std::lock_guard<std::mutex> lock(registered_device_buffers_mutex_);
         registered_device_buffers_.erase(buffer_addr);
     }
@@ -717,17 +723,19 @@ int DummyClient::register_buffer(void* buffer, size_t size) {
         return -1;
     }
     // Check bounds. The buffer must be the segment base and the size must match
-    // the caller's original request (ShmSegment::requested_size). shm->size may
-    // be padded up to the hugepage/2MB boundary for SPDK registration, so it is
-    // NOT the size the caller should pass here.
+    // either the caller's original request (ShmSegment::requested_size) or the
+    // padded mapping size (shm->size, aligned up to the hugepage/2MB boundary
+    // for SPDK registration). Both are valid: IPC always sends shm->size, and
+    // legacy callers may pass the padded size directly.
     if (reinterpret_cast<uint8_t*>(buffer) !=
             reinterpret_cast<uint8_t*>(shm->base_addr) ||
-        size != shm->requested_size) {
+        (size != shm->requested_size && size != shm->size)) {
         LOG(ERROR) << "Invalid buffer address or size for registration: "
                       "Buffer addr: "
                    << buffer << ", need addr: " << shm->base_addr
                    << ", buffer size: " << size
-                   << ", need size: " << shm->requested_size;
+                   << ", need size: " << shm->requested_size
+                   << " (padded: " << shm->size << ")";
         return -1;
     }
 
@@ -776,7 +784,15 @@ int DummyClient::unregister_buffer(void* buffer) {
     }
     auto ret = invoke_rpc<&RealClient::unregister_shm_buffer_internal, void>(
         reinterpret_cast<uint64_t>(buffer), client_id_);
-    if (ret.has_value()) {
+    // The receiver unregisters and unmaps the shm even when its internal
+    // unregisterLocalMemory fails (it reports INTERNAL_ERROR only after
+    // munmap/erase), so the mapping is gone on that side: clear the local flag
+    // for INTERNAL_ERROR (server executed the teardown) and INVALID_PARAMS
+    // (server has no such mapping). Keep it on RPC_FAIL/RPC_TIMEOUT, where the
+    // RPC may never have reached the server and the mapping may still be alive.
+    if (ret.has_value() ||
+        (!ret.has_value() && (ret.error() == ErrorCode::INTERNAL_ERROR ||
+                              ret.error() == ErrorCode::INVALID_PARAMS))) {
         shm->registered = false;
     }
     return to_py_ret(ret);
--- a/mooncake-store/src/dummy_client.cpp
+++ b/mooncake-store/src/dummy_client.cpp
@@ -657,10 +657,11 @@ int DummyClient::unregister_device_buffer_for_reconnect(void* buffer) {
 
     auto ret = invoke_rpc<&RealClient::unregister_shm_buffer_internal, void>(
         buffer_addr, client_id_);
-    // Same contract as unregister_buffer: the receiver tears the mapping down
-    // even on INTERNAL_ERROR, and INVALID_PARAMS means the server has no such
-    // mapping, so drop the local bookkeeping then; keep it on RPC_FAIL /
-    // RPC_TIMEOUT where the server may not have run.
+    // Same contract as unregister_buffer: the receiver removes the mapping on
+    // INTERNAL_ERROR too (unmapped, or quarantined when the SPDK unregister
+    // fails), and INVALID_PARAMS means the server has no such mapping, so drop
+    // the local bookkeeping then; keep it on RPC_FAIL / RPC_TIMEOUT where the
+    // server may not have run.
     if (ret.has_value() ||
         (!ret.has_value() && (ret.error() == ErrorCode::INTERNAL_ERROR ||
                               ret.error() == ErrorCode::INVALID_PARAMS))) {
@@ -784,12 +785,14 @@ int DummyClient::unregister_buffer(void* buffer) {
     }
     auto ret = invoke_rpc<&RealClient::unregister_shm_buffer_internal, void>(
         reinterpret_cast<uint64_t>(buffer), client_id_);
-    // The receiver unregisters and unmaps the shm even when its internal
-    // unregisterLocalMemory fails (it reports INTERNAL_ERROR only after
-    // munmap/erase), so the mapping is gone on that side: clear the local flag
-    // for INTERNAL_ERROR (server executed the teardown) and INVALID_PARAMS
-    // (server has no such mapping). Keep it on RPC_FAIL/RPC_TIMEOUT, where the
-    // RPC may never have reached the server and the mapping may still be alive.
+    // INTERNAL_ERROR means the receiver still executed the teardown and removed
+    // the segment from its active mappings: it unmaps normally, or -- when the
+    // SPDK unregister fails -- quarantines the mapping (retained, never reused
+    // for new registrations). Either way the old mapping is no longer usable on
+    // the receiver, so drop the local flag for INTERNAL_ERROR; also drop it for
+    // INVALID_PARAMS (server has no such mapping). Keep it on RPC_FAIL /
+    // RPC_TIMEOUT, where the RPC may never have reached the server and the
+    // mapping may still be alive.
     if (ret.has_value() ||
         (!ret.has_value() && (ret.error() == ErrorCode::INTERNAL_ERROR ||
                               ret.error() == ErrorCode::INVALID_PARAMS))) {
--- a/mooncake-store/src/dummy_client.cpp
+++ b/mooncake-store/src/dummy_client.cpp
@@ -716,17 +716,18 @@ int DummyClient::register_buffer(void* buffer, size_t size) {
         LOG(ERROR) << "Buffer is not in any registered shared memory";
         return -1;
     }
-    if (shm_helper_->is_hugepage()) {
-        size = align_up(size, get_hugepage_size_from_env());
-    }
-    // Check bounds
+    // Check bounds. The buffer must be the segment base and the size must match
+    // the caller's original request (ShmSegment::requested_size). shm->size may
+    // be padded up to the hugepage/2MB boundary for SPDK registration, so it is
+    // NOT the size the caller should pass here.
     if (reinterpret_cast<uint8_t*>(buffer) !=
             reinterpret_cast<uint8_t*>(shm->base_addr) ||
-        size != shm->size) {
+        size != shm->requested_size) {
         LOG(ERROR) << "Invalid buffer address or size for registration: "
                       "Buffer addr: "
                    << buffer << ", need addr: " << shm->base_addr
-                   << ", buffer size: " << size << ", need size: " << shm->size;
+                   << ", buffer size: " << size
+                   << ", need size: " << shm->requested_size;
         return -1;
     }
 
--- a/mooncake-store/src/dummy_client.cpp
+++ b/mooncake-store/src/dummy_client.cpp
@@ -661,7 +661,7 @@ int DummyClient::unregister_device_buffer_for_reconnect(void* buffer) {
     // INTERNAL_ERROR too (unmapped, or quarantined when the SPDK or transfer
     // engine unregister fails), and INVALID_PARAMS means the server has no such
     // mapping, so drop the local bookkeeping then; keep it on RPC_FAIL /
-    // RPC_TIMEOUT where the server may not have run.
+    // RPC_TIMEOUT where the server may not have run (see unregister_buffer).
     if (ret.has_value() ||
         (!ret.has_value() && (ret.error() == ErrorCode::INTERNAL_ERROR ||
                               ret.error() == ErrorCode::INVALID_PARAMS))) {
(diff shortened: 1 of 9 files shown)


mooncake-store/src/shm_helper.cpp
```diff
@@ -98,42 +176,92 @@ void* ShmHelper::allocate(size_t size) {
...
+    const size_t requested = size;
+
     unsigned int flags = MFD_CLOEXEC;
+    size_t hp_size = 0;  // hugepage size when use_hugepage_ (0 otherwise)
     if (use_hugepage_) {
         bool use_memfd = true;
-        size = align_up(size, get_hugepage_size_from_env(&flags, use_memfd));
+        hp_size = get_hugepage_size_from_env(&flags, use_memfd);
+        if (!(flags & MFD_HUGETLB)) {
+            // get_hugepage_size_from_env() returns 0 and sets no MFD_HUGETLB
```
Enigmo-x: 建议补充说明2MB对齐这个限制是spdk 23.01.1这种旧版本的限制，高版本像26.01已经没有这样的限制了

catyans:
The teardown path is unsafe when `spdk_mem_unregister()` fails: `free()`/`cleanup()` and the RealClient paths still clear `spdk_registered` and `munmap` the region, so SPDK may retain a translation to an unmapped or reused VA. Please only unmap after successful unregister; otherwise retain/quarantine the mapping and surface the failure, with a failure-injection regression test.

NUABO:
> The teardown path is unsafe when `spdk_mem_unregister()` fails: `free()`/`cleanup()` and the RealClient paths still clear `spdk_registered` and `munmap` the region, so SPDK may retain a translation to an unmapped or reused VA. Please only unmap after successful unregister; otherwise retain/quarantine the mapping and surface the failure, with a failure-injection regression test.

Thank you for the review. I will review the code and fix the issue.

NUABO:
> The teardown path is unsafe when `spdk_mem_unregister()` fails: `free()`/`cleanup()` and the RealClient paths still clear `spdk_registered` and `munmap` the region, so SPDK may retain a translation to an unmapped or reused VA. Please only unmap after successful unregister; otherwise retain/quarantine the mapping and surface the failure, with a failure-injection regression test.

 @catyans Thanks for the review — the underlying concern is valid and I've fixed it in commit  ("Retain SPDK-registered shm mappings when spdk_mem_unregister fails").

What changed

All teardown paths (ShmHelper::free()/cleanup(), RealClient::unmap_shm_internal, ascend_unmap_shm_internal, unregister_shm_buffer_internal, tearDownAll_internal) now only munmap/clear spdk_registered after spdk_mem_unregister() succeeds. On failure the mapping is retained (never unmapped, so the VA cannot be reused while SPDK still holds a translation to it) and quarantined: ShmHelper keeps the segment for a later free() retry; RealClient moves it into quarantine_shms_, reports INTERNAL_ERROR, and retries it at each teardown entry point (RetryQuarantinedShmsLocked) so a transient unregister failure does not leak memory until process exit.

Why I did not add a failure-injection regression test

I considered it carefully but it does not fit the current test infrastructure:

No injection seam. SpdkWrapper is a non-virtual singleton that links the real SPDK symbols directly. Forcing spdk_mem_unregister to fail from a test would require adding a test-only fault-injection hook to production code, which I'd rather avoid.

Environment prerequisites. Reaching the guarded state (spdk_registered == true) requires a real allocation path: ShmHelper::allocate under MC_STORE_REGISTER_SPDK=1 forces HugeTLB-backed mappings, and the receiver path needs an fd over IPC. A fault hook can bypass the SPDK calls but not mmap(MAP_HUGETLB); CI runners without a configured hugepage pool cannot run it.

No NoF CI coverage today. There is currently no USE_NOF build/test job in CI (the existing nof_heartbeat_test is also built only locally). A new USE_NOF test would not be exercised by CI, so it would neither prevent regressions nor justify its maintenance cost.

The invariant is instead guaranteed structurally: munmap is only reachable after a successful unregister, and a failed one leaves the mapping alive and quarantined. I verified the change on the Linux + USE_NOF setup by exercising teardown/unmap/unregister paths and confirming no mapping is released while its SPDK registration is still live. If a USE_NOF CI job with hugepages becomes available, I'm happy to add the fault-injection test on top of it.

NUABO:
hi @zxpdemonio @catyans I have already updated the code as required. The CI test errors are unrelated to my modifications. Please review it again, Thank you.

NUABO:
hi @fcczzz could you help review this pull request? thanks

fcczzz:
> hi @fcczzz could you help review this pull request? thanks

Sure, I’m still reviewing it. I’ll leave some feedback once I’m done. Thanks!

NUABO:
> > hi @fcczzz could you help review this pull request? thanks
> 
> Sure, I’m still reviewing it. I’ll leave some feedback once I’m done. Thanks!

Great, thank you for your review! I’ll keep an eye out for your feedback and address it promptly.

mooncake-store/src/real_client.cpp
```diff
@@ -2789,10 +2966,51 @@ tl::expected<void, ErrorCode> RealClient::unregister_shm_buffer_internal(
...
+                        << "Unmapped and cleaned up shared memory: "
+                        << shm_it->shm_name << ", size: " << shm_it->shm_size;
+                }
+            }
+#else
             if (!rc) {
                 LOG(ERROR) << "Failed to unregister memory: "
                            << shm_it->shm_name;
-                return tl::make_unexpected(ErrorCode::INTERNAL_ERROR);
+                unregister_failed = true;
```
fcczzz: Could we retain the mapping when TE unregistration fails instead of continuing to munmap?

For example, RdmaTransport::unregisterLocalMemoryInternal returns early if the metadata update fails, before deregistering the MR. TransferEngineImpl also retains its region record on that error. Continuing here therefore unmaps memory that TE still considers registered; if a later mmap reuses the address, registration can fail with ERR_ADDRESS_OVERLAPPED.

The USE_NOF branch above needs the same protection.

mooncake-store/src/shm_helper.cpp
```diff
@@ -98,42 +183,92 @@ void* ShmHelper::allocate(size_t size) {
...
+                         << base_addr << ", size=" << size
+                         << "; NoF zero-copy transfers to this buffer will be "
+                            "unavailable";
+            // spdk_mem_register() marks g_mem_reg_map REGISTERED before running
+            // its notify callbacks and does NOT roll back on failure (SPDK
+            // v23.01.1, lib/env_dpdk/memory.c). Unregister the range so a later
+            // registration of the same virtual address (e.g. mmap reuse after
+            // free) does not return -EBUSY. The unregister path tolerates
+            // ranges that were never fully registered.
+            SpdkWrapper::GetInstance().UnregisterMemory(base_addr, size);
```
fcczzz: Could we check the rollback result here and retain the mapping if cleanup is incomplete?

spdk_mem_register() can fail after partially registering the range. If this UnregisterMemory() call also fails, spdk_registered remains false even though SPDK may still retain registration state. A later free()/cleanup() therefore skips the unregister/quarantine guard and munmaps the region.

The receiver-side registration failure path in RealClient::map_shm_internal_with_device has the same issue.

mooncake-store/src/real_client.cpp
NUABO: Thanks for the review — you're right, both points are now fixed in 76cce5ac.
  
  For the TE side: a failed unregisterLocalMemory() is now treated as "TE still
  holds this region", so the mapping is retained (never munmapped) and moved into
  quarantine_shms_, and the RPC reports INTERNAL_ERROR. DummyClient already drops
  its local bookkeeping on INTERNAL_ERROR, so the receiver/dummy contract is
  unchanged.
  
  - unmap_shm_internal, unregister_shm_buffer_internal and
    ascend_unmap_shm_internal now only munmap after BOTH the TE and the SPDK
    unregisters succeeded. The USE_NOF arm and the non-USE_NOF arm you anchored on
    now share a single check, so both are covered by the same code path.
  - MappedShm::te_unregister_pending records which side failed, and
    RetryQuarantinedShmsLocked() retries TE -> SPDK -> munmap at every later
    teardown entry point, so a transient failure does not leak the mapping until
    process exit.
  - The retry is gated on that flag rather than unconditional on purpose:
    Client::unregisterLocalMemory() collapses every TE error into INVALID_PARAMS,
    so retrying a range that was never registered could never converge.

mooncake-store/src/shm_helper.cpp
NUABO: Thanks — fixed in 76cce5ac.
  
  Both registration sites now check the rollback result: ShmHelper::allocate() and
  RealClient::map_shm_internal_with_device(). 0 and -EINVAL are treated as
  "cleanup complete" (-EINVAL is the documented "range never reached
  g_mem_reg_map" no-op in SpdkWrapper::UnregisterMemory), while any other error
  means SPDK may still hold (partial) registration state for the range. In that
  case the segment is marked spdk_registered = true, so free()/cleanup()/teardown
  will attempt the unregister again and quarantine the mapping (retained, never
  munmapped) instead of releasing it.

mooncake-store/src/real_client.cpp
NUABO: One more gap in the same area, fixed in 84b32f99: when the TE unregister failed but the SPDK unregister succeeded, the quarantined entry kept spdk_registered = true. RetryQuarantinedShmsLocked() then re-called spdk_mem_unregister() on the already released range — SPDK returns the benign -EINVAL no-op there (no segment is marked REGISTERED), the retry treats any non-zero rc as failure, so the entry never converged: retained forever and retried on every teardown entry. The flag is now set to the actual SPDK-side result, so it still means "SPDK holds a translation for this range."

NUABO:
Hi @fcczzz, I've addressed both of your comments in 76cce5ac and 84b32f99. Could you please take another look when you get a chance? Thanks!

mooncake-store/src/dummy_client.cpp
```diff
@@ -775,7 +785,17 @@ int DummyClient::unregister_buffer(void* buffer) {
...
+    // the segment from its active mappings: it unmaps normally, or -- when the
+    // SPDK or transfer engine unregister fails -- quarantines the mapping
+    // (retained, never reused for new registrations). Either way the old
+    // mapping is no longer usable on the receiver, so drop the local flag for
+    // INTERNAL_ERROR and for INVALID_PARAMS (server has no such mapping). Keep
+    // it on RPC_FAIL / RPC_TIMEOUT, where the RPC may never have reached the
+    // server and the mapping may still be alive.
+    if (ret.has_value() ||
+        (!ret.has_value() && (ret.error() == ErrorCode::INTERNAL_ERROR ||
+                              ret.error() == ErrorCode::INVALID_PARAMS))) {
```
fcczzz, line 801: Could we avoid clearing `shm->registered` for every `INVALID_PARAMS` response? In Ascend agent mode, `unregister_shm_buffer_internal()` also returns this error when `set_context_if_needed()` fails, before unregistering TE memory or removing the mapping. Clearing the flag here then prevents a subsequent `unregister_buffer()` retry from reaching the receiver, since it is rejected locally as “not registered”. Please distinguish “mapping not found” from pre-teardown failures and preserve the flag for the latter.

NUABO: hi @fcczzz You're right, thanks for catching it. Fixed in e1d8ffe5.

unregister_shm_buffer_internal() returns INVALID_PARAMS from three places, and only the last one means the receiver refused:

no shm_contexts_ entry for client_id — nothing was ever mapped;

the address is not in mapped_shms — no such mapping;

set_context_if_needed() failed — before unregisterLocalMemory(), before the SPDK unregister and before the mapped_shms.erase(), so the mapping, its TE region and its SPDK registration are all still live.

The dummy only has ret.error() to go on, so the distinction has to be made on the receiver side: case 3 now returns RPC_FAIL instead of propagating the context error. DummyClient already keeps its bookkeeping for anything outside OK / INTERNAL_ERROR / INVALID_PARAMS, so shm->registered stays set and the next unregister_buffer() reaches the receiver again. As it was, the flag was cleared while the receiver still held the mapping: every later attempt was rejected locally as "not registered", leaking the mapping and its TE/SPDK registrations for the lifetime of the process, and the reconnect loop — which only re-registers segments still flagged as registered — skipped it as well.

INVALID_PARAMS keeps a single meaning ("the receiver holds no such mapping"), so the clearing behaviour for cases 1 and 2, and the meaning callers already assume for it, are unchanged. The same RPC and the same condition are used by unregister_device_buffer_for_reconnect(), so device-buffer bookkeeping is covered too; the contract is now spelled out in the comments at both sites.

I reused RPC_FAIL rather than adding a new code because a handler's return value is transported as a normal response payload and is not conflated with the transport-level RPC_FAIL in invoke_rpc().

mooncake-store/src/shm_helper.cpp
```diff
@@ -98,42 +184,108 @@ void* ShmHelper::allocate(size_t size) {
...
+            // ranges that were never fully registered.
+            const int rollback_rc =
+                SpdkWrapper::GetInstance().UnregisterMemory(base_addr, size);
+            // -EINVAL is the documented benign no-op for a range that never
+            // reached g_mem_reg_map (see SpdkWrapper::UnregisterMemory). Any
+            // other failure means SPDK may still retain (partial) registration
+            // state for this range, so mark it registered: free()/cleanup()
+            // will then attempt the unregister again and quarantine the mapping
+            // instead of munmapping while SPDK still holds a translation to it.
+            if (rollback_rc != 0 && rollback_rc != -EINVAL) {
```
fcczzz: Could we avoid treating every -EINVAL as successful cleanup?

In SPDK v23.01.1 with VFIO/IOVA=VA, registration can establish the DMA mapping and then fail while populating the translation table. The unregister callback can subsequently return -EINVAL because that table is incomplete, before removing the DMA mapping. This condition leaves spdk_registered false, so free()/cleanup() can still munmap the region.

This is an exceptional failure path, but please retain the mapping unless cleanup is confirmed complete. The receiver-side rollback in RealClient::map_shm_internal_with_device needs the same treatment.

NUABO: > Could we avoid treating every -EINVAL as successful cleanup?
> 
> In SPDK v23.01.1 with VFIO/IOVA=VA, registration can establish the DMA mapping and then fail while populating the translation table. The unregister callback can subsequently return -EINVAL because that table is incomplete, before removing the DMA mapping. This condition leaves spdk_registered false, so free()/cleanup() can still munmap the region.
> 
> This is an exceptional failure path, but please retain the mapping unless cleanup is confirmed complete. The receiver-side rollback in RealClient::map_shm_internal_with_device needs the same treatment.

hi @fcczzz Thanks you're right, and it holds up in the source. Really appreciate you taking the time to dig into the SPDK internals with me on this.

In the pinned SPDK (v23.01.1, lib/env_dpdk/memory.c), spdk_mem_register() marks the range in g_mem_reg_map before running its notify callbacks and does not roll back (370-384). In iova=va the vtophys notify installs the IOMMU DMA mapping for the whole range before populating the translation table (1089-1108), so a mid-way failure leaves that mapping live while unregister returns -EINVAL from the buffer_len != len check (1217-1224), before vtophys_iommu_unmap_dma.

Changes:

Only rc == 0 from spdk_mem_unregister() releases the mapping. Any other return leaves spdk_registered set, so teardown retries and quarantines instead of munmapping. Covers ShmHelper::free()/cleanup() and the receiver-side rollback in RealClient::map_shm_internal_with_device.

The normal munmap path is preserved by gating registration on a new SpdkWrapper::IsRegistrableRange(), mirroring SPDK's up-front -EINVAL checks (339-348).

Related: -EBUSY must not roll back either — the range belongs to another owner, and rolling back would clear their registration (358-366, 445-466).

-EINVAL covers both a pre-mapping failure and a post-mapping one and the code cannot distinguish them, so a mapping that could have been safely unmapped may be retained. That is the deliberate trade-off.

Looking forward to your reply~

NUABO: > Could we avoid treating every -EINVAL as successful cleanup?
> 
> In SPDK v23.01.1 with VFIO/IOVA=VA, registration can establish the DMA mapping and then fail while populating the translation table. The unregister callback can subsequently return -EINVAL because that table is incomplete, before removing the DMA mapping. This condition leaves spdk_registered false, so free()/cleanup() can still munmap the region.
> 
> This is an exceptional failure path, but please retain the mapping unless cleanup is confirmed complete. The receiver-side rollback in RealClient::map_shm_internal_with_device needs the same treatment.

fix in https://github.com/kvcache-ai/Mooncake/pull/3717/changes/00179e33fb34c90d920ba3e6ec8cfbcd72c90159

fcczzz:
hi, no further blocking findings from my review. Please resolve the merge conflicts with main.

NUABO:
> hi, no further blocking findings from my review. Please resolve the merge conflicts with main.

Thank you very much! I will resolve the conflict as soon as possible.

NUABO:
> hi, no further blocking findings from my review. Please resolve the merge conflicts with main.

Hi @fcczzz, I've resolved the merge conflicts with main. Could you please take another look when you get a chance? Thanks!

fcczzz (approved):
LGTM

ykwd (approved):

PR #3721, opened 2026-08-27
[Bugfix][Store] Support Dummy external buffer registration

zxpdemonio:
## Description

DummyClient previously accepted only pointers inside its own shared-memory
segments. As a result, `register_buffer(ptr, size)` rejected ordinary host,
pinned-host, and CUDA allocations, and external addresses could not be used
safely by the Dummy read/write APIs.

This PR adds explicit external-buffer registration semantics:

- register host, pinned-host, and CUDA addresses outside Dummy SHM;
- reference-count repeated registration of the exact `(base, size)`;
- reject same-base size changes and partially overlapping ranges;
- validate address/size and hugepage-alignment overflow;
- require exact base addresses for unregister and erase on the final release;
- clear external registrations during teardown;
- validate every external transfer range before staging;
- retain CUDA IPC fast paths and use owned SHM staging only where the Dummy
  RPC boundary requires it.
- include Ascend device and Fabric-host registrations in the same range
  validation path;
- validate CUDA IPC fast paths before export, including post-unregister and
  subrange-capacity checks;
- support `get_into_ranges` for registered external destinations by staging
  the requested prefix and copying it back after the scatter completes.

The change is limited to DummyClient registration and generic external
pointer transfers. Tensor-specific `*_from` multi-buffer routing is covered
by PR-2.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**

```bash
cmake --build build-review-nocuda --target dummy_client_get_buffer_test -j16
ctest --test-dir build-review-nocuda -R dummy_client_get_buffer_test --output-on-failure
```

**Test results:**

- [x] No-CUDA build passed on machine 70.
- [x] `dummy_client_get_buffer_test` passed: 12 test cases (10 passed, 2 skipped), latest run after rebase 81.19 s.
- [x] External host registration lifecycle covered: duplicate registration,
  size mismatch, overlap rejection, overflow rejection, partial transfer, and
  exact unregister, plus transfer rejection after the final unregister.
- [x] External-host ranged read covered, including non-contiguous destination
  ranges, preservation of untouched bytes, and rejection after unregister.
- [ ] CUDA-enabled build: blocked by latest `main` baseline incompatibility
  with the CUDA headers installed on machine 70 (`CUmemFabricHandle`,
  `CU_MEM_HANDLE_TYPE_FABRIC`, and
  `CU_DEVICE_ATTRIBUTE_HANDLE_TYPE_FABRIC_SUPPORTED` are unavailable).
- [ ] Full Python suite — not required for this focused C++ registration PR.

**Review notes:**

- The review follow-up commits (rebased onto latest `main` at `b8f294d4`) add
  309 lines and delete 57 lines across three
  source/test files. The cumulative PR branch diff from the merge-base with
  `main` is 513 additions and 70 deletions across four files (including the
  earlier registration work and its tests).
- `register_buffer` itself performs registration bookkeeping only; it does not
  allocate a staging buffer or copy user data.
- External Dummy transfers may use one operation-scoped SHM staging allocation
  because the RPC handler runs against the Dummy-mapped address space.

### DIFF
--- a/mooncake-store/src/dummy_client.cpp
+++ b/mooncake-store/src/dummy_client.cpp
@@ -605,7 +605,7 @@ std::optional<BufferHandle> DummyClient::allocate_client_buffer(size_t size) {
                                             std::move(release));
 }
 
-bool DummyClient::is_dummy_shm_buffer(void *buffer, size_t size) const {
+bool DummyClient::is_dummy_shm_buffer(void* buffer, size_t size) const {
     if (buffer == nullptr || shm_helper_ == nullptr) return false;
     auto shm = shm_helper_->get_shm(buffer);
     if (!shm) return false;
@@ -615,16 +615,17 @@ bool DummyClient::is_dummy_shm_buffer(void *buffer, size_t size) const {
     return size <= shm->size - (address - base);
 }
 
-bool DummyClient::is_device_buffer(void *buffer) const {
+bool DummyClient::is_device_buffer(void* buffer) const {
     return device::GetAcceleratorRegistry()
                .RuntimeAccelerators()
                .FindDeviceForPointer(buffer) != nullptr;
 }
 
-std::optional<size_t> DummyClient::external_buffer_remaining(void *buffer) const {
+std::optional<size_t> DummyClient::external_buffer_remaining(
+    void* buffer) const {
     const uintptr_t address = reinterpret_cast<uintptr_t>(buffer);
     std::lock_guard<std::mutex> lock(registered_external_buffers_mutex_);
-    for (const auto &[base, registration] : registered_external_buffers_) {
+    for (const auto& [base, registration] : registered_external_buffers_) {
         if (address >= base && address - base <= registration.size) {
             return registration.size - (address - base);
         }
@@ -633,7 +634,7 @@ std::optional<size_t> DummyClient::external_buffer_remaining(void *buffer) const
 }
 
 std::optional<DummyClient::PreparedBuffer> DummyClient::prepare_buffer(
-    void *buffer, size_t size, bool copy_to_staging, bool copy_back) {
+    void* buffer, size_t size, bool copy_to_staging, bool copy_back) {
     if (buffer == nullptr && size != 0) return std::nullopt;
     if (size == 0) return PreparedBuffer{buffer, buffer, 0, nullptr, false};
     if (shm_helper_ != nullptr) {
@@ -653,17 +654,22 @@ std::optional<DummyClient::PreparedBuffer> DummyClient::prepare_buffer(
     if (!allocation) return std::nullopt;
     auto staging = std::make_unique<BufferHandle>(std::move(*allocation));
     if (copy_to_staging) {
-        const auto &runtime = device::GetAcceleratorRegistry().RuntimeAccelerators();
-        if (!runtime.CopyToHost(staging->ptr(), buffer, size)) return std::nullopt;
+        const auto& runtime =
+            device::GetAcceleratorRegistry().RuntimeAccelerators();
+        if (!runtime.CopyToHost(staging->ptr(), buffer, size))
+            return std::nullopt;
     }
     return PreparedBuffer{buffer, staging->ptr(), size, std::move(staging),
                           copy_back || !copy_to_staging};
 }
 
-bool DummyClient::copy_from_staging(const PreparedBuffer &buffer, size_t size) const {
-    if (!buffer.copy_back || buffer.staging == nullptr || size == 0) return true;
+bool DummyClient::copy_from_staging(const PreparedBuffer& buffer,
+                                    size_t size) const {
+    if (!buffer.copy_back || buffer.staging == nullptr || size == 0)
+        return true;
     if (size > buffer.size) return false;
-    const auto &runtime = device::GetAcceleratorRegistry().RuntimeAccelerators();
+    const auto& runtime =
+        device::GetAcceleratorRegistry().RuntimeAccelerators();
     return runtime.CopyFromHost(buffer.original, buffer.staging->ptr(), size);
 }
 
@@ -784,7 +790,8 @@ int DummyClient::register_buffer(void* buffer, size_t size) {
         const uintptr_t base = reinterpret_cast<uintptr_t>(buffer);
         if (size > std::numeric_limits<uintptr_t>::max() - base) return -1;
         std::lock_guard<std::mutex> lock(registered_external_buffers_mutex_);
-        for (const auto &[other_base, registration] : registered_external_buffers_) {
+        for (const auto& [other_base, registration] :
+             registered_external_buffers_) {
             if (base < other_base + registration.size &&
                 other_base < base + size &&
                 (base != other_base || size != registration.size)) {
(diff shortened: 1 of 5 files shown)


zxpdemonio:
Validation update on machine 70: CUDA-enabled build and dummy_client_get_buffer_test passed (1/1, 76.5s). External registration lifecycle and range checks are covered. No source/worktree changes after validation; CI is green except the platform CI Gate still queued.

mooncake-store/src/dummy_client.cpp
```diff
@@ -600,6 +605,74 @@ std::optional<BufferHandle> DummyClient::allocate_client_buffer(size_t size) {
...
+        if (shm != nullptr && !is_device_buffer(buffer)) {
+            if (!is_dummy_shm_buffer(buffer, size)) {
+                return std::nullopt;
+            } else {
+                return PreparedBuffer{buffer, buffer, size, nullptr, false};
+            }
+        }
+    }
+    auto remaining = external_buffer_remaining(buffer);
+    if (!remaining.has_value() || size > *remaining) return std::nullopt;
```
yokinoshitayoki, line 732: Under USE_ASCEND_DIRECT, device buffers are tracked only in registered_device_buffers_, while fabric-host buffers are registered through register_ascend_shm(); neither range is present in registered_external_buffers_. This lookup therefore rejects put_from/get_into/upsert_from immediately after register_buffer() succeeds. Include these registered ranges in validation or preserve their mapped fast path.

mooncake-store/src/dummy_client.cpp
```diff
@@ -1167,7 +1261,19 @@ std::vector<int> DummyClient::batch_put_from(
         return batch_put_from_cuda_ipc(requests, config);
```
yokinoshitayoki: This fast path runs before prepare_buffer(), so an unregistered CUDA allocation, a range larger than the registered subrange, or a buffer used after unregister_buffer() can still be transferred. Validate every buffer against the external registration before exporting it, and apply the same check to the batch_get_into CUDA path.

mooncake-store/src/dummy_client.cpp
yokinoshitayoki, line 1390: register_buffer() now accepts external memory, but get_into_ranges() still forwards the process-local address directly to the real client, which only maps Dummy SHM addresses. Registered external ranged reads therefore return INVALID_PARAMS. Validate and stage the requested ranges, then copy successful ranges back to the original buffers.

zxpdemonio:
Addressed all three review comments in commit `2a95c31b` (rebased onto latest
`main` at `b8f294d4`):

- Ascend device and Fabric-host registrations now participate in external range validation. Fabric-host registrations are reference-counted and are removed from the local registry only after the real-side unmap succeeds.
- CUDA IPC put/get fast paths now require every `(pointer, size)` range to be registered and within capacity before exporting an IPC handle. Unregistered and post-unregister pointers fall back to the checked staging path and are rejected when not registered.
- `get_into_ranges` now validates the nested request shape, stages registered external destinations through Dummy SHM, preserves untouched destination holes, validates the RPC response shape, and copies the completed ranges back.

Validation on machine 70:

- No-CUDA build succeeded.
- `ctest --test-dir build-review-nocuda -R dummy_client_get_buffer_test --output-on-failure`: passed (12 tests: 10 passed, 2 skipped; 83.56 s).
- After rebasing onto the latest `main`, the full `dummy_client_get_buffer_test` rerun also passed (10 passed, 2 skipped; 81.19 s).
- CUDA build remains blocked by the latest `main` baseline with the installed CUDA headers: `CUmemFabricHandle`, `CU_MEM_HANDLE_TYPE_FABRIC`, and `CU_DEVICE_ATTRIBUTE_HANDLE_TYPE_FABRIC_SUPPORTED` are unavailable.

mooncake-store/src/dummy_client.cpp
zxpdemonio: Fixed in `2a95c31b` (rebased onto latest `main` at `b8f294d4`). Ascend device registrations and Fabric-host registrations are now included in external range validation. Fabric-host registrations are reference-counted and removed locally only after the real-side unmap succeeds.

mooncake-store/src/dummy_client.cpp
zxpdemonio: Fixed in `2a95c31b` (rebased onto latest `main` at `b8f294d4`). CUDA IPC put/get fast paths now require every pointer range to be registered and within its registered capacity before exporting an IPC handle; post-unregister and oversized subranges use the checked path and are rejected.

mooncake-store/src/dummy_client.cpp
zxpdemonio: Fixed in `2a95c31b` (rebased onto latest `main` at `b8f294d4`). get_into_ranges now validates the nested request shape, stages registered external destinations through Dummy SHM, preserves untouched holes, validates the response shape, and copies completed ranges back.

mooncake-store/src/dummy_client.cpp
```diff
@@ -600,6 +605,139 @@ std::optional<BufferHandle> DummyClient::allocate_client_buffer(size_t size) {
...
+            }
+        }
+    }
+
+#if defined(USE_ASCEND_DIRECT)
+    // Ascend device allocations are registered through register_ascend_shm()
+    // and tracked separately so they can be restored after reconnect.  Treat
+    // those mappings as registered external buffers for all transfer paths.
+    {
+        std::lock_guard<std::mutex> lock(registered_device_buffers_mutex_);
```
yokinoshitayoki, line 641: This map is now treated as the authoritative registration set, but register_device_buffer_for_reconnect() still stores only base -> size. Registering the same device range twice succeeds without incrementing a reference count, so the first unregister unmaps and erases the buffer while one registration remains; partially overlapping device ranges are also accepted. Apply the same overlap/refcount bookkeeping here and map/unmap the real side only on the first/final reference.

zxpdemonio: Addressed in bc4ebab0. Device-buffer registrations now use the same overlap/refcount ledger as external buffers: exact duplicates retain a reference, partial overlaps and size changes are rejected, the RealClient map is issued only for the first reference, and unmap only for the final reference. Registration and unregistration hold the device ledger lock across the RPC; reconnect replays tracked Fabric and device ranges under their registration locks; and RealClient per-buffer unregister is idempotent so a lost response can be retried safely. Regression coverage now exercises duplicate registration, overlap and overflow rejection, final release, and missing-mapping idempotency. clang-format and dummy_client_get_buffer_test pass in the no-CUDA build; Ascend-specific behavior remains subject to CI and device validation.

yokinoshitayoki (approved):
LGTM.

catyans (approved):
Nice follow-up here. Thanks @yokinoshitayoki for spotting the less obvious edge cases, and @zxpdemonio for patiently working through all of them — buffer registration across process boundaries is exactly the kind of code where the unhappy paths matter.

I took another pass over the latest version and I’m happy with where it landed. Duplicate registrations now have sensible ownership, invalid ranges can’t sneak through the CUDA IPC fast path, and the ranged-read staging/copy-back path looks solid. Making the final unregister retryable after a lost response is a particularly nice touch.

LGTM from me. Nice work!

stmatengss (approved):

PR #3722, opened 2026-08-27
[Bugfix][Store] Route tensor _from writes through multi-buffers

zxpdemonio:
## Description

Serialized tensor objects use `[TensorMetadata][payload]`. This change validates the fixed metadata header and routes the four ordinary raw tensor APIs through ordered metadata/payload parts, allowing DummyClient to handle registered external host, pinned host, and CUDA buffers without wrapper-level concatenation or whole-object host staging.

| API | Write path |
| --- | --- |
| `put_tensor_from` | Single-key `put_from_with_metadata`; DummyClient prepares separate metadata/payload buffers. |
| `batch_put_tensor_from` | `batch_put_from_multi_buffers`. |
| `upsert_tensor_from` | One-key `batch_upsert_tensor_from`, with result-count validation. |
| `batch_upsert_tensor_from` | `batch_upsert_from_multi_buffers`. |

The wrapper copies only the fixed header for validation, retains the caller's original addresses for registration lookup, and omits empty payload slices. It checks metadata, object size, and address arithmetic, including overflow before reading the header. Callers must provide live readable memory and register the complete source region for the duration of the synchronous operation. CUDA buffers require a CUDA-enabled build/runtime.

DummyClient stages external parts in owned SHM for the cross-process RPC boundary. RealClient uses registered source addresses directly (`stage_nonlocal=false` for the multi-buffer path). These APIs are not a zero-copy guarantee across the Dummy/Real process boundary.

External registration also requires working read destinations. `get_into_ranges` and `batch_get_into_multi_buffers` stage reads and copy successful bytes back. Sparse ranged reads now pack fragments into SHM: reading one byte at the end of a registered buffer larger than the local pool allocates only the fragment bytes, rather than the unused prefix. Original destination bounds and physical SHM bounds are checked separately; failed fragments do not copy back. The existing ranged-read RPC signature remains unchanged.

**Scope:** The no-whole-object-staging statement applies to the four ordinary raw APIs above. The TP raw decoding path still copies a device-resident serialized object to host before decoding; this PR also guards TP result counts. This supersedes the earlier cover letter's blanket staging statement. The branch includes the external-registration work originally stacked on `codex/dummy-register-buffer-pr` and now targets `main`.

## Module

- [x] CI/CD

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

Latest head: `966d9596` (implementation fix `3ae43dbb`, CI wiring, and merge of current `main`).

Local checks passed:

- A temporary harness compiled the actual ranged-buffer preparation/copy-back helpers with AddressSanitizer and UndefinedBehaviorSanitizer. The old helper reproduced the sparse-offset allocation failure; the fix passed 10,000 sparse/interior/invalid/zero-length cases, copy bounds, and overlapping-fragment checks. The allocator/runtime were mocked; this is helper-level validation, not RPC or accelerator E2E.

```bash
python3 -m py_compile mooncake-wheel/tests/test_dummy_client.py
```

Regression coverage:

- C++ `ExternalHostSparseRangedReadBeyondPool`: destination larger than the 512 MiB pool, distant fragments, mixed SHM/external destinations, zero-length fragments, invalid source/destination ranges, missing keys, untouched gaps, and interior destination pointers.
- Existing external-host ranged and multi-buffer read tests.
- The Ubuntu 22.04 / Python 3.12 wheel smoke job now starts a bounded Mooncake client and runs the focused `test_dummy_client.py` tensor-from and external-host regressions against the built wheel; its exit status gates the job.
- Python round-trip equality for all four ordinary raw APIs using external host, pinned host, and CUDA source buffers; accelerator-dependent cases skip when unavailable.
- Metadata-only tensors through all four ordinary APIs.
- Malformed sizes, null pointers, and pointer overflow through all four ordinary APIs.

Latest-head [Linux CI run](https://github.com/kvcache-ai/Mooncake/actions/runs/34199678996) is queued awaiting runner assignment; compilation, CTest, and wheel integration results are not yet available. The obsolete implementation-only run was force-cancelled to release the workflow concurrency slot. Both review threads are resolved, and the latest head is mergeable. The previous head `3e588d21` passed all applicable hosted checks. Previous machine-70 runtime reports are historical and are not validation of this fix: that SSH host is currently unreachable. No fresh CUDA runtime execution is claimed.

## Checklist

- [ ] Human submitter has reviewed every changed line and can defend the change end-to-end
- [x] Formatted changed C/C++ lines using `./scripts/code_format.sh`
- [x] Ran pre-commit on the PR's touched files; all applicable hooks pass
- [x] Added regression tests for the requested fixes
- [ ] Latest-head hosted build and test checks pass

## AI Assistance Disclosure

- [x] AI tools were used: OpenAI Codex inspected the review comments and implementation, implemented the sparse-read and tensor-upsert fixes, added regression coverage, and ran the available checks. Human review remains required.


mooncake-store/src/dummy_client.cpp
```diff
@@ -710,14 +811,38 @@ int DummyClient::register_buffer(void* buffer, size_t size) {
         // non-Fabric Host: same rules as shm
     }
 #endif
-    // Find which shm this buffer belongs to
+    if (shm_helper_ == nullptr) return -1;
     auto shm = shm_helper_->get_shm(buffer);
+    if (shm && is_device_buffer(buffer)) shm = nullptr;
     if (!shm) {
```
yokinoshitayoki, line 873: External registration succeeds here, but `get_into_ranges()` and `batch_get_into_multi_buffers()` still forward the original process-local address. RealClient can only map Dummy SHM, so these APIs fail for the newly accepted buffers. Please route these read paths through SHM staging/copy-back as `batch_get_into()` does.

zxpdemonio: Addressed in 3f640848. get_into_ranges() and batch_get_into_multi_buffers() now stage registered process-local host destinations in Dummy-mapped shared memory and copy successful reads back to the original buffers. Ranged reads preserve sparse destination offsets and stage only the touched extent; multi-buffer reads also handle zero-length slices. The existing get_into_ranges_shm_helper RPC signature remains unchanged, with a separate staged helper carrying validated physical capacities. Verified with the Store and mooncake_client builds, dummy_client_get_buffer_test (1/1 passed), the full DummyClient Python suite (13 passed, 4 subtests passed), and pre-commit on all touched files.

catyans:
Thanks @yokinoshitayoki for catching the read-side gap, and @zxpdemonio for extending the fix. I re-checked the current head: the metadata/payload split preserves object ordering, external Dummy buffers are staged per part, and the new ranged-read helper carries validated capacities before copying successful ranges back. I did not find another correctness issue in these paths.

The remaining code-format check is failing on `dummy_client.h` and `dummy_client.cpp`. Could you run the repository formatter and rerun CI? Once that is clean, this looks ready for approval from my side.

yokinoshitayoki:
@zxpdemonio There are some failed checks and conflicts, could you take a look?

zxpdemonio:
Addressed the review feedback and pushed commit `3e588d21681cf23d9a57dde576c35acac30ef9dd`.

Changes:
- Keep single-key `put_tensor_from()` on the single-key `put_from_with_metadata()` path.
- Keep batch tensor-from operations on the multi-buffer APIs, with Dummy/Real dispatch through `store_`.
- Stage device raw tensor objects to host before TP decoding; avoid dereferencing device pointers from C++.
- Preserve metadata-only tensors by omitting zero-sized payload slices in DummyClient.
- Validate TP batch/upsert result counts before indexing or reporting success.
- Add DummyClient coverage for metadata-only external tensor objects.
- Format and pre-commit checks pass.

Validation on 60.205.250.70:
- `store` C++ target: passed
- Dummy Python suite: `12 passed, 3 skipped`
- metadata-only Dummy tensor smoke test: passed
- CUDA runtime tests were not claimed because this build has CUDA disabled.

The PR is currently mergeable with no conflicts; GitHub CI is rerunning for the new head.

yokinoshitayoki:
@zxpdemonio CI checks failed, could you have a look?

mooncake-store/src/dummy_client.cpp
```diff
@@ -744,14 +766,84 @@ std::optional<DummyClient::PreparedBuffer> DummyClient::prepare_buffer(
...
+        for (size_t i = 0; i < row_count; ++i) {
+            const size_t fragment_count =
+                std::min(dst_offsets[i].size(), sizes[i].size());
+            for (size_t j = 0; j < fragment_count; ++j) {
+                const size_t offset = dst_offsets[i][j];
+                const size_t fragment_size = sizes[i][j];
+                if (offset > capacity || fragment_size > capacity - offset) {
+                    continue;
+                }
+                staging_size = std::max(staging_size, offset + fragment_size);
```
yokinoshitayoki: A valid range at a large destination offset makes `staging_size` cover the entire unused prefix. For example, reading one byte at the end of a registered `kLocalBufSize + 4096` buffer requests more staging than the local pool and returns `INVALID_PARAMS`.

mooncake-store/src/dummy_client.cpp
```diff
@@ -714,10 +714,32 @@ int DummyClient::unregister_external_buffer(void* buffer) {
     return 0;
 }
 
+#if defined(USE_ASCEND_DIRECT)
+std::optional<size_t> DummyClient::registered_ascend_buffer_remaining(
+    void* buffer) const {
+    const uintptr_t address = reinterpret_cast<uintptr_t>(buffer);
+    std::lock_guard<std::mutex> lock(registered_device_buffers_mutex_);
+    for (const auto& [base, registration] : registered_device_buffers_) {
```
Aionw, line 722: [P2] Preserve direct mappings for registered Ascend Fabric host buffers

With USE_ASCEND_DIRECT and ASCEND_ENABLE_USE_FABRIC_MEM=1, register_buffer() records external Fabric host allocations in registered_external_buffers_ and maps them into RealClient through register_ascend_shm(). This helper only searches registered_device_buffers_, so it misses those already-mapped host allocations. For allocations outside ShmHelper, prepare_buffer() then falls through to allocate_client_buffer(size) and stages them. Since this PR adds prepare_multi_buffers() to batch_put_from_multi_buffers(), batch_upsert_from_multi_buffers(), and batch_get_into_multi_buffers(), requests that previously used the existing Fabric mapping now fail with INVALID_PARAMS when a part exceeds the local staging pool or its available capacity; smaller requests incur unnecessary allocation and copying.

Please include registered Fabric host mappings in the direct-mapping lookup while retaining staging for ordinary external host/CUDA buffers. A regression test should register an external Fabric host buffer larger than the local pool and exercise the multi-buffer write/read paths. This finding is based on the registration and preparation call paths; I have not run it on Ascend hardware.

stmatengss: Ascend Fabric is unrelated to the current PR.

zxpdemonio:
All required CI checks for this PR have passed, including CI Gate, unit tests, wheel builds/tests, formatting, Ruff, Sphinx, spelling, and path checks. The tent-ci job is skipped rather than failed. The PR is currently only waiting for reviewer approval.

stmatengss (approved):

mooncake-store/src/dummy_client.cpp
zxpdemonio: Thanks for the detailed Ascend-specific analysis. I re-checked the current diff and confirmed that this PR changes the ordinary raw tensor routing and Dummy external-buffer staging/ranged-read copy-back; it does not change the existing Ascend Fabric registration or reconnect path. The Fabric direct-mapping behavior predates this PR and requires Ascend hardware coverage, so I am keeping it as a separate follow-up rather than expanding this PR beyond the pinned/host/CUDA support it targets. The SHM staging fallback remains intentional for ordinary external buffers.

PR #3723, opened 2026-08-27
[Bugfix][Store] Support RealClient CUDA tensor reads

zxpdemonio:
## Description

Enable `get_tensor_into_cuda` and `batch_get_tensor_into_cuda` for RealClient. Read the tensor metadata prefix, validate the destination, then transfer the payload directly into registered destination memory. Empty tensors need no payload transfer; DummyClient retains its CUDA IPC path.

Address the review findings:

- Put the CUDA Fabric compatibility declarations under `tent/include/tent/common`, with a forwarding header for the classic engine. Standalone TENT builds no longer need the classic include directory, and either header include order uses one declaration.
- Preserve negative metadata and payload read errors per item. Return `INTERNAL_ERROR` for malformed result dimensions or an unexpected nonnegative byte count.
- Use one query snapshot for both the metadata and payload reads. This path never refreshes into a newer version; an expired snapshot returns `LEASE_EXPIRED`. Full-object CUDA snapshot reads also bypass the key-based hot cache, which can contain a newer value.
- Preserve leased memory replicas during upsert by allocating replacement storage, including for same-size updates. File-backed upserts return `OBJECT_HAS_LEASE` while readers hold a lease because their key-based storage cannot retain the old version. Unleased same-size updates still reuse storage.

The snapshot guarantee requires the master-side change in this PR. Old allocations are retained for at least the full read TTL and writer grace period, measured after locking metadata. The destination must still be registered before reading:

```python
store.register_buffer(tensor.data_ptr(), tensor.numel() * tensor.element_size())
```

## Module

- [x] CI/CD

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

- PR-scoped pre-commit and clang-format checks pass.
- `python3 mooncake-wheel/tests/test_cuda_fabric_compat.py -v`: both tests pass, covering six toolkit/include-order combinations and standalone MNNVL preprocessing.
- An ASan/UBSan harness using the actual result-mapping loop passes nine cases; the original PR code fails the negative-error case.
- On final head `56b55800`, GitHub CTest passes all 167 test executables, including `pybind_client_test`, `dfs_sync_client_test`, and the master tests.
- A native CUDA regression populates the hot cache with a newer value and verifies that a full-object read still returns the retained snapshot. It skips at runtime without a CUDA device.
- Both Ubuntu wheel jobs (22.04/Python 3.12 and 24.04/Python 3.10) pass all three deterministic tensor snapshot regressions and both CUDA compatibility tests. The snapshot tests cover an upsert between the metadata and payload reads, mixed missing/valid keys, and payload lease-error propagation. The Ubuntu 22.04 job also passes PG CPU tests, Store API/RPC smoke tests, and SSD offload/promotion E2E tests.
- [Final-head GitHub CI](https://github.com/kvcache-ai/Mooncake/actions/runs/34078630600): all 24 workflow jobs pass, including the final CI Gate. CUDA 12.8/13, ROCm, EFA, TENT, and both Ubuntu wheel configurations pass. The CUDA 13 job passed on retry after its first hosted runner lost communication.
- GPU runtime execution was not repeated for these revisions. The new native CUDA regression compiles in CI but skips without a CUDA device; the Python regressions exercised CPU destinations on the hosted runners.

```bash
./scripts/code_format.sh --changed-lines --check --base origin/main
python3 mooncake-wheel/tests/test_cuda_fabric_compat.py -v
# With a running master and matching DEFAULT_KV_LEASE_TTL:
python3 mooncake-wheel/tests/test_tensor_read_snapshot.py -v
```

## Checklist

- [x] Reviewed the changed code
- [x] Formatted the changed C/C++ lines
- [x] Ran pre-commit on the files changed in this PR
- [x] Added regression tests
- [ ] For changes >500 LOC: filed an RFC issue

## AI Assistance Disclosure

- [x] AI tools were used: Codex inspected the review comments, implemented the fixes and regression tests, and ran validation. Human maintainer review remains required.


zxpdemonio:
Validation update on machine 70: CUDA-enabled store target passed; CUDA 12.0 compatibility compile-only check and non-CUDA/CUDA targeted regressions passed. The local Python CUDA integration attempt was blocked before the new read path by the existing metadata service curl URL error; GitHub CI CUDA-off/on checks are green and wheel jobs are still running.

mooncake-transfer-engine/tent/include/tent/transport/mnnvl/mnnvl_transport.h
```diff
@@ -17,6 +17,7 @@
 
 #include <cuda.h>
 #include <cuda_runtime.h>
+#include "cuda_fabric_compat.h"
```
yokinoshitayoki: Standalone TENT exposes only `tent/include`, but this header now unconditionally includes `cuda_fabric_compat.h` from the sibling transfer-engine include directory. This breaks standalone CUDA builds.

mooncake-integration/store/store_py.cpp
```diff
@@ -814,8 +816,68 @@ class MooncakeStorePyWrapper {
...
+            }
+            for (size_t i = 0; i < original_indices.size(); ++i) {
+                const size_t original_index = original_indices[i];
+                if (i < range_results.size() && range_results[i].size() == 1 &&
+                    range_results[i][0].size() == 1 &&
+                    range_results[i][0][0] >= 0 &&
+                    static_cast<size_t>(range_results[i][0][0]) ==
+                        metadata[original_index]->data_bytes) {
+                    results[original_index] = range_results[i][0][0];
+                }
```
yokinoshitayoki, line 919: A negative ranged-read result is ignored here, leaving the item as `INVALID_PARAMS`; this hides errors such as `OBJECT_NOT_FOUND` or `LEASE_EXPIRED` and differs from the DummyClient path. Propagate negative per-item results, and use `INTERNAL_ERROR` only for malformed result shapes or unexpected byte counts.

catyans:
The new RealClient path does not bind the metadata prefix and payload read to the same object version. `batch_get_tensor_metadata_prefixes()` calls `get_into_ranges(..., nullptr)` and releases that query/lease; the payload phase then calls `get_into_ranges(..., nullptr)` again, causing a separate query. An `upsert` of the same key between those phases can therefore validate the destination against version A's metadata but write version B's payload. If the two versions have the same dtype/shape/byte count but different values, the final byte-count check succeeds and the API silently returns a mixed-version result.

Could both phases use one query snapshot/version—for example, batch-query once, retain the resulting `QueryResultCache`/lease, and pass it to both ranged reads, with an explicit failure if that version can no longer be served? A deterministic regression test can pause after the metadata-prefix read, upsert a same-shape tensor with different contents, resume the payload read, and assert the call returns one coherent version (or a retry/error), never old metadata plus new payload reported as success.

This is separate from the currently noted negative-result propagation issue: even with all per-range errors forwarded correctly, a same-sized version change produces a positive byte count and still passes the current loop.

stmatengss:
Addressed all three review findings and updated the branch with `main`:

1. Standalone TENT owns the canonical CUDA Fabric compatibility header; the classic header forwards to it. Old/new toolkit fixtures and both include orders pass.
2. Negative per-item errors are preserved. Malformed result dimensions and unexpected byte counts return `INTERNAL_ERROR`; the result-mapping regression harness passes all nine cases and fails against the original code.
3. Metadata and payload retain one query snapshot, with `LEASE_EXPIRED` instead of a version-changing refresh. The master allocates replacement storage for leased memory replicas and rejects leased file-backed upserts with `OBJECT_HAS_LEASE`. The snapshot guarantee requires the updated master.

A further correctness check found that full-object CUDA snapshot reads could use a newer key-based hot-cache value. Those reads now bypass the hot cache, with a native regression covering this case.

Pushed head: `56b55800087adeae7827f031e5ce426eaf2ed12b`. Both inline threads are resolved and the PR is mergeable.

Validation on this head:

- CTest: all 167 test executables pass.
- Both Ubuntu wheel jobs: all three deterministic tensor snapshot tests and both CUDA compatibility tests pass.
- PG CPU, Store API/RPC smoke, and SSD offload/promotion E2E tests pass.
- PR-scoped pre-commit and changed-lines clang-format checks pass.
- [GitHub CI](https://github.com/kvcache-ai/Mooncake/actions/runs/34078630600): all 24 workflow jobs pass, including CUDA 12.8/13, ROCm, EFA, TENT, both Ubuntu configurations, and the final CI Gate. The CUDA 13 job passed on retry after its first hosted runner lost communication.

GPU runtime validation was not repeated. The native CUDA regression compiles but skips without a CUDA device, and the hosted Python regressions exercised CPU destinations.

stmatengss (approved):

alogfans (approved):
LGTM

mooncake-store/src/master_service.cpp
```diff
@@ -5419,7 +5412,19 @@ auto MasterService::UpsertStart(const UUID& client_id, const std::string& key,
...
+                if (has_read_lease &&
+                    metadata.HasReplica([](const Replica& replica) {
+                        return !replica.is_memory_replica();
+                    })) {
+                    // File-backed reads can resolve storage by key, so they
+                    // cannot retain an old version across replacement.
+                    return tl::make_unexpected(ErrorCode::OBJECT_HAS_LEASE);
+                }
 
+                if (metadata.size == slice_length && !has_read_lease) {
```
yokinoshitayoki, line 5461: A same-size memory object with an active read lease now falls through to Case C, which erases the old metadata before allocating its replacement. If allocation returns NO_AVAILABLE_HANDLE, the failed Upsert leaves the key OBJECT_NOT_FOUND even though its old replicas are retained.

mooncake-store/tests/pybind_client_test.cpp
```diff
@@ -250,6 +250,70 @@ TEST_F(RealClientTest, BatchGetIntoUsesSelectedLocalDiskEndpoint) {
 }
 
 #ifdef MOONCAKE_TEST_CUDA_H2D
+TEST_F(RealClientTest, RangedSnapshotGpuReadBypassesNewerHotCacheValue) {
```
yokinoshitayoki, line 253: MOONCAKE_TEST_CUDA_H2D is set whenever CUDAToolkit is found, including USE_CUDA=OFF. On a GPU host this test runs, but RealClient has no CUDA accelerator registered and TCP attempts to write directly to the device pointer, failing with Bad address/TRANSFER_FAIL. Require USE_CUDA for this test, or skip unless the registry recognizes the destination pointer.

zxpdemonio:
Addressed the two latest review findings and pushed commit `1b814fd68e1ee04e1e30e126e01f534b87e4c645`:

- For an active read lease, `UpsertStart` now allocates replacement replicas before removing the old metadata. Allocation/quota failure leaves the existing object and its snapshot readable.
- Added `MasterServiceTest.LeasedUpsertAllocationFailurePreservesObject`.
- `MOONCAKE_TEST_CUDA_H2D` is now enabled only when `USE_CUDA` is enabled, avoiding CUDA H2D execution in CUDA-toolkit-only CPU builds.

Validation on machine 70:

- PR-scoped pre-commit: passed.
- `master_service_test` build with `USE_CUDA=OFF`: passed.
- `MasterServiceTest.LeasedUpsertAllocationFailurePreservesObject`: passed.
- `pybind_client_test`: 64/65 passed; the single failure was an existing HTTP metadata server port collision in `TestCopyMoveQueryTask`. Rerunning that test alone passed.

The new GitHub workflow run for this head is currently queued. Review threads are intentionally left unresolved for reviewer confirmation.

mooncake-store/src/master_service.cpp
zxpdemonio: Fixed in commit `1b814fd68e1ee04e1e30e126e01f534b87e4c645`. With an active read lease, UpsertStart now charges and allocates replacement replicas before erasing the old metadata. Allocation failure releases only the replacement charge and returns the error, so the old metadata/replicas remain readable. Added `MasterServiceTest.LeasedUpsertAllocationFailurePreservesObject`, which passes on machine 70.

mooncake-store/tests/pybind_client_test.cpp
zxpdemonio: Fixed in commit `1b814fd68e1ee04e1e30e126e01f534b87e4c645`. The `pybind_client_test` CUDA H2D definition and `CUDA::cudart` link are now guarded by `USE_CUDA`, matching the runtime capability instead of mere toolkit discovery. A `USE_CUDA=OFF` compile check on machine 70 confirms `MOONCAKE_TEST_CUDA_H2D` is absent from `pybind_client_test`.

yokinoshitayoki:
@zxpdemonio Seems there still are some failed CI checks, could you take a look?

catyans (approved):
Thanks for working through all of these edge cases. LGTM on the implementation.

I rechecked the snapshot path that originally worried me: both the tensor metadata prefix and payload now consume the same retained `QueryResultCache`, and strict snapshot reads cannot silently refresh to a newer object version. The latest allocation-before-removal change also means a leased upsert that cannot allocate its replacement leaves the old object readable. Together with @yokinoshitayoki's `USE_CUDA` guard fix, the failure behavior now looks coherent rather than partially destructive.

One mechanical item remains before merge: the current `Check code format` job reports diffs in `master_service.h`, `master_service.cpp`, and `master_service_test.cpp`, which is why the downstream jobs were skipped. Please run the repository formatter and let CI complete, but I don't see a remaining correctness blocker. This was a pretty intricate change—nice job keeping the tests focused as the design evolved.

Issue #1749, opened 2026-03-26
[RFC]: Optimizing Mooncake TCP Transport for High-Concurrency Throughput


yxj-97:
### Changes proposed

# 1. Summary
<font style="color:rgb(13, 18, 57);background-color:rgb(246, 247, 248);">This RFC proposes optimizations to Mooncake TCP Transport by introducing multiple io_contexts for parallel processing, a concurrency limiter, asynchronous connection establishment, and CUDA asynchronous copy, in order to improve throughput and GPU memory transfer efficiency under high-concurrency workloads.</font>

# <font style="color:rgb(13, 18, 57);background-color:rgb(246, 247, 248);">2. Motivation</font>
## 2.1 <font style="color:rgb(13, 18, 57);background-color:rgb(246, 247, 248);">Current Architecture Analysis</font>
<font style="color:rgb(13, 18, 57);background-color:rgb(246, 247, 248);">The existing TCP Transport uses a single io_context and worker thread model:</font>

<img width="1372" height="692" alt="Image" src="https://github.com/user-attachments/assets/e299d957-ec01-4cc3-842f-1eb94655e1a7" />

## 2.2 <font style="color:rgb(13, 18, 57);background-color:rgb(246, 247, 248);">Performance Bottlenecks</font>
| **Bottleneck** | **Cause** | **Impact** |
| --- | --- | --- |
| Single-threaded I/O | A single io_contexts is driven by a single worker thread, so CPU multicore resources cannot be used in parallel to process I/O events. Throughput becomes limited under high concurrency. | Low utilization of multicore CPUs |
| Synchronous connection establishment | getConnections waits synchronously for TCP three-way handshake completion. | Blocks the calling thread; connection latency propagates into transfer latency |
| No concurrency limit | Transfer tasks are submitted without restriction. | File descriptor exhaustion, memory pressure, and system instability during traffic spikes |
| Synchronous CUDA copy | cudaMemcpy blocks until the copy completes. | Blocks the I/O thread, preventing it from handling other socket events |
| Temporary buffers | Each CUDA transfer allocates and frees a temporary buffer via new/delete. | <font style="color:rgb(0, 0, 0);background-color:rgb(239, 246, 255);">Memory allocation overhead and fragmentation</font> |

## <font style="color:rgb(13, 18, 57);background-color:rgb(246, 247, 248);">2.3 Current Performance</font>
[...]

PR #3724, opened 2026-08-27
[TE] Shard TCP transport I/O across a pool of io_context threads

anranxia:
## Description

The TCP transfer path currently runs a **single `asio::io_context` on one worker thread** (`TcpTransport::worker` → `io_context.run()`). Every per-peer lane, socket completion, admission timer, and reconnect handler executes on that one thread. On a fast cross-node TCP path (~100 GbE) that single thread saturates at 100% CPU and caps throughput well below line rate — adding more lanes stops helping once they all funnel through the one thread.

This PR introduces **`TcpIoPool`**, a bounded pool of `io_context` threads, and spreads I/O across them while preserving the #2974 bounded-lane model, admission control, reconnect, and exactly-once completion.

Design:
- New `TcpIoPool` owns `N` `io_context` instances, each with its own thread and a work-guard; threads are named `mc-tcp-io-<i>`. `N = MC_NUM_TCP_IO_THREADS` (**default 1 = unchanged behavior**, bounded [1, 32]).
- Each `PeerConnectionGroup`'s coordination state (queue/admission/retry/retirement timers, pump) is pinned to a single shard chosen by hash of the connection key, so its `mutex`-guarded state stays serialized on one executor.
- Each `ConnectionLane`'s socket, and each accepted server socket, are round-robin bound to a shard (not hashed per-peer), so a single hot peer uses multiple cores. `runGroupPump` dispatches a lane's connect/session onto that lane's shard via `asio::dispatch` (runs inline when N=1, so the single-thread path is byte-for-byte the previous behavior).
- Shutdown stops every shard and joins every thread; that join-all is the quiescence barrier.

Correctness rests on the machinery that already existed for the single-thread path — `PeerConnectionGroup::mutex`, the `operation_epoch`/`pump_epoch`/`retry_epoch`/`admission_epoch` stale-handler guards, and the move-only `TerminalAction` + `ClientSession::finalize` exactly-once path. These are mutex/atomic based; they are now genuinely exercised concurrently. Because each socket is pinned to exactly one shard's single-threaded `io_context`, asio's per-socket handler serialization holds without adding strands.

This PR also fixes a **pre-existing data race** (found while adding TSan coverage): `TcpTransport::getTransferStatus` read `task.transferred_bytes` / `success_slice_count` / `failed_slice_count` with plain loads while the I/O thread updates them via `__atomic_fetch_add`. It now uses `__atomic_load_n(..., __ATOMIC_ACQUIRE)`, matching `MultiTransport::getTransferStatus`. This race predates this change (it exists between the single io thread and the polling submission thread).

Relates to RFC #1749; builds on the bounded-lane model from #2974.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix 
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [x] Performance improvement
- [ ] Other

## How Has This Been Tested?

Two cross-node hosts over a ~100 GbE TCP path (data-transfer bytes confirmed to cross the NIC via `/proc/net/dev`). Built with `-DUSE_CUDA=ON -DBUILD_EXAMPLES=ON`. `MC_TCP_ENABLE_CONNECTION_POOL=1` (default).

**Regression — `MC_NUM_TCP_IO_THREADS=1` is unchanged:** cross-node 1 MiB read = 2.49 GB/s, identical to the pre-change baseline; loopback matches too (see below).

**Cross-node scaling (1 MiB, DRAM read):**

| MC_NUM_TCP_IO_THREADS | lanes | submit threads | GB/s |
|---|---|---|---|
| 1 | 8 | 1 | 2.49 |
| 2 | 8 | 1 | 4.61 |
| 4 | 8 | 1 | 7.51 |
| 8 | 16 | 4 | **12.28** (~5×, ≈98 Gbit/s) |

WRITE, `N=8` lanes=16 threads=4: 12.40 GB/s. Per-thread CPU spreads across `mc-tcp-io-*` instead of one thread at 100%.

**Loopback, reproducing the #2974 reviewer config** (`transfer_engine_bench`, threads=4/batch=32/64 KiB/10s):

| config | READ GB/s | WRITE GB/s |
|---|---|---|
| 4-lane pool (baseline, this repo) | 1.70 / 1.72 | 2.11 / 1.12 |
| io-pool N=1, 4-lane | 1.45 / 1.53 | 1.78 / 1.07 |
| io-pool N=4, 4-lane | 2.96 / 2.99 | 2.80 / 2.96 |
| io-pool N=8, 8-lane | 3.11 / 3.28 | 3.25 / 3.25 |

(The baseline 4-lane numbers line up with the #2974 "main 4-lane" measurements, confirming the setup is comparable.)

**Correctness — `transfer_engine_validator`, cross-node, 12 workers:** all 12 workers report `Data validation passed` at both `MC_NUM_TCP_IO_THREADS=4` and `=8`.

**ThreadSanitizer:** built `-fsanitize=thread` and ran the validator (loopback, N=4 and N=8). No data races on the new shared structures (`TcpIoPool` / `PeerConnectionGroup` / `ConnectionLane`) — the existing mutex+epoch model holds under real concurrency. The only transport race TSan reported is the pre-existing `getTransferStatus` counter read (reproducible at N=1), which this PR fixes; after the fix no transport hot-path races remain. (Remaining reports are in the benchmark harness — a `running` stop-flag and buffer reuse — and a `condition_variable` shutdown-tracker report where both sides hold the same mutex.)

**Test commands:**
```bash
cmake -S mooncake-transfer-engine -B build -G Ninja \
  -DUSE_ETCD=OFF -DUSE_CUDA=ON -DBUILD_EXAMPLES=ON -DCMAKE_BUILD_TYPE=Release
cmake --build build -j

# target host
MC_NUM_TCP_IO_THREADS=8 MC_TCP_LANES_PER_PEER=16 \
  ./build/example/transfer_engine_bench --mode=target \
  --metadata_server=P2PHANDSHAKE --protocol=tcp --local_server_name=<ip>:17001
# initiator host (segment_id = the target's "listening on <ip:port>")
MC_NUM_TCP_IO_THREADS=8 MC_TCP_LANES_PER_PEER=16 \
  ./build/example/transfer_engine_bench --mode=initiator \
  --metadata_server=P2PHANDSHAKE --protocol=tcp --segment_id=<ip>:<port> \
  --operation=read --block_size=1048576 --batch_size=32 --threads=4 --duration=10
```

**Test results:**
- [ ] Unit tests pass
- [x] Manual testing done (cross-node bench + validator + TSan, described above)

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

AI (Claude Code) assisted with codebase exploration, drafting the `TcpIoPool` implementation, and running the benchmark/validator/TSan sweeps. The submitter has reviewed every changed line and can defend the change end-to-end.

### DIFF
--- a/mooncake-transfer-engine/include/transport/tcp_transport/tcp_transport.h
+++ b/mooncake-transfer-engine/include/transport/tcp_transport/tcp_transport.h
@@ -35,9 +35,14 @@
 #include <utility>
 #include <vector>
 
+#include <asio/dispatch.hpp>
+#include <asio/executor_work_guard.hpp>
 #include <asio/io_context.hpp>
 #include <asio/ip/tcp.hpp>
+#include <asio/post.hpp>
 #include <asio/steady_timer.hpp>
+#include <glog/logging.h>
+#include <pthread.h>
 
 #include "transfer_metadata.h"
 #include "transport/transport.h"
@@ -48,6 +53,101 @@ class TransferMetadata;
 struct ClientSession;
 class TcpContext;
 class TcpTransport;
+
+// A bounded pool of asio::io_context instances, each driven by its own thread.
+// TCP lanes and accepted server sockets are round-robin distributed across the
+// shards so a single hot peer can use more than one CPU core. With size 1 the
+// behavior is identical to the historical single-io_context transport.
+class TcpIoPool {
+   public:
+    explicit TcpIoPool(size_t num_threads) {
+        if (num_threads < 1) num_threads = 1;
+        contexts_.reserve(num_threads);
+        guards_.reserve(num_threads);
+        for (size_t i = 0; i < num_threads; ++i) {
+            contexts_.push_back(std::make_unique<asio::io_context>());
+            guards_.push_back(
+                std::make_unique<work_guard_t>(contexts_[i]->get_executor()));
+        }
+    }
+
+    ~TcpIoPool() { stop(); }
+
+    size_t size() const { return contexts_.size(); }
+
+    asio::io_context &context(size_t i) {
+        return *contexts_[i % contexts_.size()];
+    }
+
+    asio::io_context::executor_type executor(size_t i) {
+        return contexts_[i % contexts_.size()]->get_executor();
+    }
+
+    // Group coordination (pump/admission/retry/retirement timers) is pinned to
+    // one shard chosen by hash, so a group's serialized state stays on a single
+    // executor while its lanes spread across shards.
+    asio::io_context::executor_type coordinatorExecutor(size_t hash) {
+        return executor(hash % contexts_.size());
+    }
+
+    // Round-robin executor for a new lane / accepted socket.
+    asio::io_context::executor_type nextLaneExecutor() {
+        return executor(rr_.fetch_add(1, std::memory_order_relaxed));
+    }
+
+    size_t nextShardIndex() {
+        return rr_.fetch_add(1, std::memory_order_relaxed) % contexts_.size();
+    }
+
+    // Spawns one thread per shard. before_run(i) runs on shard i's thread
+    // before each io_context.run() (used to arm the acceptor on shard 0 and to
+    // re-arm it after an exception-driven restart).
+    void start(std::function<void(size_t)> before_run) {
+        for (size_t i = 0; i < contexts_.size(); ++i) {
+            threads_.emplace_back([this, i, before_run] {
+#ifdef __linux__
+                std::string name = "mc-tcp-io-" + std::to_string(i);
+                pthread_setname_np(pthread_self(), name.c_str());
+#endif
+                while (!stopping_.load(std::memory_order_acquire)) {
+                    try {
+                        if (before_run) before_run(i);
+                        contexts_[i]->run();
+                        break;
+                    } catch (const std::exception &e) {
+                        LOG(ERROR) << "TcpIoPool shard " << i
+                                   << " encountered an exception during run: "
+                                   << e.what();
+                        // Do not restart once stop() has published its intent:
+                        // re-arming (e.g. shard 0's async_accept) after stop()
+                        // stopped the context would keep run() alive and hang
+                        // join().
+                        if (stopping_.load(std::memory_order_acquire)) break;
+                        contexts_[i]->restart();
+                    }
+                }
+            });
+        }
+    }
+
+    void stop() {
+        stopping_.store(true, std::memory_order_release);
+        guards_.clear();
+        for (auto &c : contexts_) c->stop();
+        for (auto &t : threads_)
+            if (t.joinable()) t.join();
+        threads_.clear();
+    }
+
+   private:
+    using work_guard_t =
+        asio::executor_work_guard<asio::io_context::executor_type>;
+    std::vector<std::unique_ptr<asio::io_context>> contexts_;
+    std::vector<std::unique_ptr<work_guard_t>> guards_;
+    std::vector<std::thread> threads_;
+    std::atomic<size_t> rr_{0};
+    std::atomic<bool> stopping_{false};
+};
 
 class TcpTransport : public Transport {
    public:
@@ -94,8 +194,6 @@ class TcpTransport : public Transport {
     int unregisterLocalMemoryBatch(
         const std::vector<void *> &addr_list) override;
 
-    void worker();
-
     Slice *prepareTransfer(TransferTask *task, const TransferRequest &request);
 
     void startTransfer(Slice *slice,
@@ -111,7 +209,8 @@ class TcpTransport : public Transport {
    private:
     TcpContext *context_;
     std::atomic_bool running_;
-    std::thread thread_;
+    std::unique_ptr<TcpIoPool> io_pool_;
+    size_t num_io_threads_ = 1;
     bool enable_connection_pool_ = true;
 
     // Client-side bounded work queues and fixed connection lanes.
@@ -211,10 +310,16 @@ class TcpTransport : public Transport {
     enum class LaneConnectStage { NONE, RESOLVING, CONNECTING };
 
     struct ConnectionLaneRuntime {
-        explicit ConnectionLaneRuntime(asio::io_context &io_context)
-            : executor(io_context.get_executor()) {}
+        explicit ConnectionLaneRuntime(TcpIoPool &pool_arg) : pool(&pool_arg) {}
 
-        asio::io_context::executor_type executor;
+        asio::io_context::executor_type coordinatorExecutor(size_t hash) {
+            return pool->coordinatorExecutor(hash);
+        }
+        asio::io_context::executor_type nextLaneExecutor() {
+            return pool->nextLaneExecutor();
+        }
+
+        TcpIoPool *pool;
     };
 
     struct ConnectionLaneState;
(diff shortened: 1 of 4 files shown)


mooncake-transfer-engine/include/transport/tcp_transport/tcp_transport.h
```diff
@@ -49,6 +54,94 @@ struct ClientSession;
...
+                    try {
+                        if (before_run) before_run(i);
+                        contexts_[i]->run();
+                        break;
+                    } catch (const std::exception &e) {
+                        LOG(ERROR) << "TcpIoPool shard " << i
+                                   << " encountered an exception during run: "
+                                   << e.what();
+                        contexts_[i]->restart();
+                    }
```
alogfans, line 127: `stop()` does not publish any stopping state to this retry loop. If an I/O handler throws concurrently with shutdown, `stop()` can mark the context stopped and then this catch block can call `restart()` afterward. Shard 0 will subsequently re-arm `async_accept`, while `stop()` has already issued its only stop request, so `join()` may wait indefinitely. Please set an atomic stopping flag before clearing the guards/stopping the contexts, and check it before both `before_run()` and `restart()`.

mooncake-transfer-engine/src/transport/tcp_transport/tcp_transport_lane_impl.h
alogfans: This callback runs on `group->executor`, but it directly calls `ClientSession::cancel()`, `resolver->cancel()`, and socket cancel/close for every lane. With multiple shards, those objects are concurrently used by handlers running on `lane->executor`; in particular, `ClientSession::cancel()` mutates non-atomic timer, epoch, and deadline fields that its completion handlers also access. The old single-`io_context` implementation serialized these operations, but the pool removes that guarantee. Please post each lane's cancellation to `lane->executor` and track those per-lane posts, while keeping the group timers on `group->executor`.

mooncake-transfer-engine/src/transport/tcp_transport/tcp_transport_lane_impl.h
```diff
@@ -858,10 +859,26 @@ void TcpTransport::runGroupPump(
...
-    for (size_t i = 0; i < session_count; ++i)
-        startLaneSession(group, sessions[i].lane, sessions[i].epoch);
+    // Dispatch each lane's connect/session onto that lane's shard executor so
+    // its socket I/O runs on the owning io thread. asio::dispatch runs inline
+    // when already on the target executor (the single-shard/N=1 case stays
+    // identical to the historical inline calls) and hops otherwise.
+    for (size_t i = 0; i < connect_count; ++i) {
+        auto connect_lane = connects[i].lane;
+        uint64_t connect_epoch = connects[i].epoch;
+        asio::dispatch(connect_lane->executor,
```
staryxchen, line 879: **Cross-thread socket `cancel()`/`close()` from the coordinator shard (UB with `MC_NUM_TCP_IO_THREADS > 1`)**

The PR states the safety invariant: *"each socket is pinned to exactly one shard's single-threaded `io_context`, asio's per-socket handler serialization holds without adding strands."* Two teardown paths break that invariant once lanes no longer share the group's executor.

**Path 1 — `runGroupRetirement`** (`tcp_transport_lane_impl.h:460-470`)

It is posted to `group->executor` (the coordinator shard), but the objects it moves out of the lanes — `sessions[i]`, `resolvers[i]`, `sockets[i]` — live on the lanes' round-robin shard executors:

```cpp
for (size_t i = 0; i < session_count; ++i)
    if (sessions[i]) sessions[i]->cancel();          // socket_->cancel(); socket_->close();
for (size_t i = 0; i < resolver_count; ++i) ... resolver->cancel();
for (size_t i = 0; i < socket_count; ++i) closeSocketNoThrow(sockets[i]);
```

`ClientSession::cancel()` (`tcp_transport_session_impl.h:432-440`) calls `socket_->cancel()` and `socket_->close()` directly on the coordinator thread.

**Path 2 — the posted cancellation lambda in `shutdownConnectionLanes`** (`tcp_transport_lane_impl.h:1410-1446`)

It is posted to `group->executor` and then cancels/closes every lane's session, resolver, and socket — again, lane-shard objects touched from the coordinator thread.

**Why this was safe before the PR (and at N=1):** pre-PR, `group->executor` and every lane's socket executor were the same single `io_context` driven by one thread, so these cancel/close calls were serialized with the outstanding async ops by the single-threaded run loop. At N=1 the new code is equivalent (lane executor == coordinator executor), so this only bites when N>1 and a lane lands on a different shard than its group's coordinator.

**Why it's a problem:** asio's documented thread-safety model is "shared objects: unsafe" — concurrent `close()`/`cancel()` from the coordinator thread with an outstanding `async_read`/`async_write`/`async_connect` on the lane's io thread is undefined behavior. Concretely: a peer address change retires a group while one of its lanes is mid-transfer on another shard → the coordinator closes the socket concurrently with the lane thread's reactor servicing that descriptor. Possible outcomes include corrupted descriptor op queues or a completion handler running against an already-closed (and possibly reused) fd.

Note that a clean TSan run doesn't establish safety here: the epoll reactor serializes descriptor access with an internal mutex, so TSan sees locked accesses rather than a data race — the contract violation is at the asio object level, which TSan can't observe.

**Suggested fix:** mirror what this change already does for `startLaneConnect`/`startLaneSession` — the lane now carries its shard `executor`, so in both teardown paths collect `(session, resolver, socket, lane->executor)` triples under the group mutex and `asio::dispatch` each lane's teardown onto its owning shard:

```cpp
asio::dispatch(lane->executor, [session, resolver, socket] {
    if (session) session->cancel();
    if (resolver) { try { resolver->cancel(); } catch (...) {} }
    closeSocketNoThrow(socket);
});
```

The retry/admission timers can stay on the coordinator (they are group-executor objects). For `shutdownConnectionLanes` the existing `LaneCancellationPostTracker` + 2s wait + stop/join barrier already provides the structure to wait for these per-lane posts. At N=1 `asio::dispatch` runs inline, so the regression guarantee is preserved.

mooncake-transfer-engine/include/transport/tcp_transport/tcp_transport.h
anranxia: Good catch, fixed. `TcpIoPool` now has an atomic `stopping_` flag. `stop()` sets it (`store(true, release)`) **before** clearing the work guards and stopping the contexts. Each shard's run loop checks `stopping_` before `before_run()` (it's now the `while` condition) and again before `restart()` in the `catch` block. So if an I/O handler throws concurrently with shutdown, the loop no longer re-arms `async_accept` / re-enters `run()` after `stop()` has issued its stop — `join()` can't hang. At N=1 this is unchanged behavior.

mooncake-transfer-engine/src/transport/tcp_transport/tcp_transport_lane_impl.h
anranxia: Fixed. The `shutdownConnectionLanes` cancellation no longer touches lane objects from `group->executor`. Under `group->mutex` it now collects each lane's `{executor, session, resolver, socket}`, then `asio::dispatch`es that lane's `cancel()`/`close()` onto the lane's own `lane->executor`; the group's retry/admission timers stay on `group->executor`. Every per-lane post is tracked by the existing `LaneCancellationPostTracker`, so `waitUntil()` still waits for all of them before `stop()`/`join()`. At N=1 `asio::dispatch` runs inline, so the single-threaded path is unchanged.

mooncake-transfer-engine/src/transport/tcp_transport/tcp_transport_lane_impl.h
anranxia: Thanks for the precise write-up — both teardown paths are fixed exactly as suggested.

- `runGroupRetirement` and the `shutdownConnectionLanes` cancellation now collect `(session, resolver, socket, lane->executor)` under the group mutex and `asio::dispatch` each lane's teardown onto its **owning shard**. The group's retry/admission timers stay on the coordinator executor (they are group-executor objects).
- Agreed that a clean TSan run does not establish safety here: the epoll reactor serializes descriptor access with an internal mutex, so TSan sees locked accesses rather than a data race — the contract violation is at the asio-object level. This is now avoided **structurally** by only ever touching a socket from its owning io thread, not by relying on the sanitizer.
- At N=1 `asio::dispatch` runs inline, so the regression guarantee is preserved.

staryxchen (approved):
LGTM

anranxia:
> LGTM

All checks have passed. Whenever you have a moment, could you help merge this? Thanks a lot for the reviews!

staryxchen:
> > LGTM
> 
> All checks have passed. Whenever you have a moment, could you help merge this? Thanks a lot for the reviews!

Merged. Thanks for your contribution.

PR #3727, opened 2026-08-27
[TransferEngine][EP] Support Kimi K3 & guard against duplicate glog initialization

kyx1999:
## Description

This PR bundles six related commits from the `kyx1999/add-kimi-k3-support` branch, covering Kimi K3 EP support and a duplicate-glog-initialization fix (plus its cross-version glog compatibility):

1. **Kimi K3 support (EP)** — Add hidden size `3584` to the `SWITCH_HIDDEN` dispatch macro (`case 3584 /* for kimi k3 */`) and raise `kNumMaxTopK` / `kNumMaxTopk` from `11` to `17` in the EP kernel (both the dispatch and combine launch paths) so the expert-parallel path can accommodate Kimi K3's model configuration and routing.
2. **Reproduce duplicate glog initialization** — Add a regression test (`glog_preinitialized_test`) that reproduces the abort caused by re-initializing glog when the host process has already initialized Google logging.
3. **Fix duplicate glog initialization** — Guard `google::InitGoogleLogging` in `loadGlobalConfig` (triggered by the `MC_LOG_DIR` env var) with an "already initialized" check so the transfer engine reuses the host's existing logging setup instead of triggering a duplicate-initialization abort (`"You called InitGoogleLogging() twice!"`).
4. **Support glog < 0.6.0** — `IsGoogleLoggingInitialized()` was only exported at the top level in glog 0.6.0. To keep building against older distros, `FindGLOG.cmake` now detects the glog version (via CONFIG or pkg-config) and defines `MOONCAKE_GLOG_HAS_IS_INITIALIZED=1` only when glog >= 0.6.0. `config.cpp` (and the test) select the right symbol accordingly: the public `google::IsGoogleLoggingInitialized()` on >= 0.6.0, and the forward-declared `google::glog_internal_namespace_::IsGoogleLoggingInitialized()` on older versions, keeping the call site identical.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

A new unit test (`glog_preinitialized_test`) reproduces the duplicate glog initialization scenario: it pre-initializes glog as the host would, then calls `loadGlobalConfig` and verifies logging is reused (no re-init) while `FLAGS_log_dir` is still applied from `MC_LOG_DIR`. The test is version-agnostic and resolves to the correct glog symbol via `MOONCAKE_GLOG_HAS_IS_INITIALIZED`. The Kimi K3 changes were validated by exercising the EP dispatch path with hidden size `3584`.

**Test commands:**
```bash
# Build and run the transfer engine tests (includes glog_preinitialized_test)
ctest --test-dir build -R glog_preinitialized_test
```

**Test results:**
- [x] Unit tests pass
- [x] Manual testing done (EP dispatch verified for hidden size 3584)

## Checklist

- [x] For changes >500 LOC: I have filed an RFC issue (N/A: total ~+202/-27 lines)

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex is used to locate some bugs.

### DIFF
--- a/mooncake-common/FindGLOG.cmake
+++ b/mooncake-common/FindGLOG.cmake
@@ -7,63 +7,64 @@ find_package(glog QUIET CONFIG)
 set(GLOG_DETECTED_VERSION "")
 
 if(TARGET glog::glog)
-    set(GLOG_FOUND TRUE)
-    set(GLOG_TARGET glog::glog)
-    if(DEFINED glog_VERSION)
-        set(GLOG_DETECTED_VERSION "${glog_VERSION}")
-    endif()
+  set(GLOG_FOUND TRUE)
+  set(GLOG_TARGET glog::glog)
+  if(DEFINED glog_VERSION)
+    set(GLOG_DETECTED_VERSION "${glog_VERSION}")
+  endif()
 else()
-    find_package(PkgConfig QUIET)
-    if(PKG_CONFIG_FOUND)
-        pkg_check_modules(PC_GLOG QUIET libglog)
-    endif()
+  find_package(PkgConfig QUIET)
+  if(PKG_CONFIG_FOUND)
+    pkg_check_modules(PC_GLOG QUIET libglog)
+  endif()
 
-    find_path(GLOG_INCLUDE_DIR glog/logging.h
-        HINTS ${PC_GLOG_INCLUDEDIR} ${PC_GLOG_INCLUDE_DIRS}
-        PATHS /usr/include /usr/local/include)
-    
-    find_library(GLOG_LIBRARY glog
-        HINTS ${PC_GLOG_LIBDIR} ${PC_GLOG_LIBRARY_DIRS}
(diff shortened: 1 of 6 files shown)


UNIDY2002 (approved):

alogfans:
@kyx1999 Please aware of the errors from CI.

kyx1999:
> @kyx1999 Please aware of the errors from CI.

OK, I've reviewed it. The CI failure is a glog version mismatch. google::IsGoogleLoggingInitialized() is only exported as a top-level symbol since glog 0.6.0; in older glog (which the CI image uses) the same function lives in the internal namespace google::glog_internal_namespace_, so referencing it unconditionally fails to build there. I added version detection + conditional compilation to pick the right symbol, which fixes the build.

alogfans (approved):
LGTM

PR #3728, opened 2026-08-27
[TENT] Add capability graph path synthesis

alogfans:
## Description

Add a TENT capability graph and path synthesizer for deciding whether a request can use a direct transport path or needs local/remote staging. Transports now expose their memory reachability capabilities, and TransferEngine uses the synthesized path to select the cross-node transport and staging parameters.

This is PR 1 of 4 in the TENT path selection / RDMA latency stack.

## Related Issue

Related issue: #3733

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
cmake -S . -B build
cmake --build build --target tent_capability_graph_test tent_transport_selector_test -j16
ctest --test-dir build -R '^(tent_capability_graph_test|tent_transport_selector_test)$' --output-on-failure
```

**Test results:**
- [x] Unit tests pass

## Checklist

Note: `pre-commit` is not installed in the local environment, so the PR-scoped pre-commit command could not be run locally.

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex was used to split the original branch into focused PRs, resolve rebase conflicts, review diffs, and run local validation. The human submitter should review and be able to defend every changed line.


catyans:
The synthesizer never sees the transports that actually require staging in the production integration. `resolveExecutionRoute()` builds `route_candidates` via `getTransportType(req, i)`, which has already applied reachability to the original source/target memory types, and then enables cross-node transports only from that list.

For example, for CUDA-to-CUDA with RDMA installed but GPUDirect disabled, RDMA has `cross_node_transfer=true` and `dram_to_dram=true`, but `gpu_to_gpu=false`. `getTransportType()` therefore filters RDMA out before `CapabilityPathSynthesizer::synthesize()` runs. If TCP is also available, the graph sees only TCP and selects the direct TCP route instead of the intended CUDA -> host -> RDMA -> host -> CUDA path. The MNNVL CPU-to-CPU staging case is filtered in the same way because MNNVL does not advertise `dram_to_dram`. The unit tests bypass this integration issue by enabling RDMA/MNNVL directly in `CapabilityGraphInput`.

Could the cross-hop candidate set be built from the remote buffer's advertised transports plus locally installed `cross_node_transfer` capabilities, independently of whether each transport can directly reach the original endpoints? The graph can then perform the memory-pair reachability checks for each synthesized cross hop, while transport hints/policy still constrain or rank the authorized set. An integration-level regression through `resolveExecutionRoute()` for RDMA without GPUDirect (with TCP also enabled) would catch this: it should select staged RDMA, not direct TCP.

catyans:
Thanks for the updates. I re-reviewed the current head, and the original integration issue still appears unresolved: `resolveExecutionRoute()` continues to build `route_candidates` through `getTransportType(req, i)`, then enables cross-node graph edges only from that already direct-reachability-filtered list.

The new commits clear the queued task's old QP pool, tighten an isolated graph assertion, and cache the selected queued route, but they do not let staging-only RDMA/MNNVL candidates reach the synthesizer. Could you update the production candidate collection and add the `resolveExecutionRoute()` regression described above?

catyans:
Thanks, the new `InstalledOnly` candidate path and integration tests address the original staging issue.

One routing contract still looks broken: `listCandidates()` preserves the transport-hint/policy order, but `resolveExecutionRoute()` enables every candidate in one graph and the synthesizer then chooses solely by fixed path cost. For example, with staged RDMA cheaper than direct TCP, `transport_hint=TCP` can still resolve to RDMA; a policy ordered as `[TCP, RDMA]` has the same problem. That differs from the existing selector behavior where the hint is pinned first and policy order defines preference.

Could route synthesis preserve candidate order—for example, choose the first candidate with a feasible synthesized path—and add integration cases for a TCP hint and a TCP-first policy when staged RDMA is also feasible?

alogfans:
Thanks for the re-review. This should be addressed in the current head after resetting the branch: `resolveExecutionRoute()` now builds cross-node graph candidates via `TransportSelector::listCandidates(..., TransportReachabilityMode::InstalledOnly)` instead of `getTransportType(req, i)`, so staging-only transports like RDMA without GPUDirect or MNNVL without `dram_to_dram` still enter the capability graph when installed and policy/hint-authorized, with per-hop reachability checked after staging memory is selected. I also added `resolveExecutionRoute()` integration regressions, `ResolveExecutionRoutePrefersStagedRdmaOverDirectTcp` and `ResolveExecutionRouteSeesMnnvlStagingCandidate`, covering cases where direct TCP is available but staged RDMA/MNNVL should win instead of being filtered before synthesis.

catyans:
Thanks @alogfans. `InstalledOnly` addresses staging admission, but not the ordering concern in my latest comment. On `8f493a6`, the ordered `listCandidates()` result is collapsed into boolean `enabled` entries; the synthesizer then enumerates transports and selects solely by fixed cost. A TCP hint or `[TCP, RDMA]` policy can therefore still lose to a cheaper staged RDMA path.

Please preserve candidate rank during path selection (or synthesize candidates in order), with a staged-vs-direct hint/policy regression.

Issue #3477, opened 2026-08-17
[RFC]: Mooncake Store RPC correctness, compatibility, and execution model


Aionw:
## Background

Mooncake Store uses `coro_rpc` for two major communication paths:

- `MasterClient` ↔ `WrappedMasterService`
- `DummyClient` ↔ `RealClient`

An audit found one current correctness problem and several protocol compatibility and availability risks.

This issue records and prioritizes the current problems. It intentionally does not propose a concrete redesign; the expected compatibility guarantees and implementation direction should be discussed with the community first.

## Priority summary

| Priority | Finding | Impact |
|---|---|---|
| P0 | Client/server handler registries have already diverged | Existing RPC calls fail with “function not registered” |
| P1 | The wire contract is implicit and not safely evolvable | Rolling client/server upgrades cannot be guaranteed |
| P1 | Release-version equality is used as protocol negotiation | Both false rejection and false acceptance are possible |
| P1 | Blocking handlers occupy `coro_rpc` I/O threads | Slow requests can delay ping, readiness, and unrelated RPCs |
| P1 | Timed-out mutations have ambiguous outcomes | Clients cannot safely determine whether to retry |
| P2 | Error semantics and connection-state handling are inconsistent | Failures are difficult to classify and recover from |
| P2 | Request sizes and exposed RPC operations are insufficiently bounded | Resource-exhaustion and deployment-security risks |
| P2 | Platform-dependent C++ types are exposed on the wire | Compatibility depends on architecture and source layout |
| P3 | Large payload copies, incomplete metrics, and missing contract tests | Performance and regressions are difficult to quantify |

### Client methods without registered server handlers

The following methods are invoked by clients but are absent from the corresponding server registration lists:

- `WrappedMasterService::CalcCacheStats`
- `RealClient::batch_get_query_results`
- `RealClient::batch_get_replica_desc`
- `RealClient::get_replica_desc`

Relevant code:

- [`MasterClient::CalcCacheStats`](https://github.com/kvcache-ai/Mooncake/blob/main/mooncake-store/src/master_client.cpp#L474)
- [Master RPC registration](https://github.com/kvcache-ai/Mooncake/blob/main/mooncake-store/src/rpc_service.cpp#L1637)
- [DummyClient `batch_get_query_results` call](https://github.com/kvcache-ai/Mooncake/blob/main/mooncake-store/src/dummy_client.cpp#L1122)
[...]

PR #3734, opened 2026-08-27
[Bugfix][Store] Register missing RPC handlers for #3477 P0

huangxiaobai-ydmy:
## Description

Fixes the **P0 functional failures** documented in [#3477](https://github.com/kvcache-ai/Mooncake/issues/3477) — *Mooncake Store RPC correctness, compatibility, and execution model*.

Mooncake Store maintains **two parallel RPC surfaces**:

| Path | Client | Server |
|------|--------|--------|
| Master RPC | `MasterClient` | `WrappedMasterService` via `RegisterRpcService()` |
| Real-client RPC | `DummyClient` (Python / embedding) | `RealClient` via `RegisterClientRpcService()` |

An audit in #3477 found that several **client-side call sites already existed**, and the corresponding **server-side handler implementations already existed**, but the handlers were **never registered** on the production server startup paths. At runtime this surfaces as coro_rpc **"function not registered"** errors rather than a business-logic failure.

This PR intentionally scopes itself to **P0 registration drift only**. It does **not** attempt the broader #3477 work (protocol versioning, `struct_pack::compatible` schema evolution, async handler migration, etc.).

### Root cause

Store RPC registration is **manual and duplicated**:

- `MasterClient` invokes methods via `invoke_rpc<&WrappedMasterService::…>`.
- `DummyClient` invokes methods via `invoke_rpc<&RealClient::…>`.
- Each method must be explicitly listed in `RegisterRpcService()` / `RegisterClientRpcService()`.

There is no compile-time or CI guard ensuring the three lists stay in sync (client calls, server impl, server registration). #3477 lists this drift as direct evidence that the current model is fragile; this PR closes the **currently confirmed gaps** called out under P0.

#### 1. Master path — `CalcCacheStats`

- **Client call:** `MasterClient::CalcCacheStats()` (`master_client.cpp`)
- **Server impl:** `WrappedMasterService::CalcCacheStats()` (`rpc_service.cpp`) — already implemented
- **Missing:** `RegisterRpcService()` did not register the handler

Any caller that reaches the master over RPC (as opposed to calling `WrappedMasterService` in-process) would fail even though `MasterMetricManager::calculate_cache_stats()` itself works.

#### 2. Real-client path — replica / query helpers

These are used by `DummyClient`, which is the bridge for Python bindings and embedded multi-process setups:

| DummyClient method | RealClient RPC target | Used for |
|--------------------|----------------------|----------|
| `batch_get_query_results` | `RealClient::batch_get_query_results` | Cached query result fetch |
| `get_replica_desc` | `RealClient::get_replica_desc` | Single-key replica descriptor |
| `batch_get_replica_desc` | `RealClient::batch_get_replica_desc` | Batch replica descriptors |

- **Client calls:** `dummy_client.cpp`
- **Server impl:** `real_client.cpp`
- **Missing:** `RegisterClientRpcService()` in `real_client_main.cpp` did not register them

Notably, unit tests such as `dummy_client_get_buffer_test.cpp` **manually register** `batch_get_query_results` in a local test helper, which masked the production gap — tests passed while the real `mooncake_client` binary failed.

#### `mooncake-store/src/rpc_service.cpp`

Register `WrappedMasterService::CalcCacheStats` in `RegisterRpcService()` next to `ServiceReady`, aligning the production master server with the existing `MasterClient` RPC name traits.

#### `mooncake-store/src/real_client_main.cpp`

Register the three missing `RealClient` handlers in `RegisterClientRpcService()`:

- `batch_get_query_results`
- `get_replica_desc`
- `batch_get_replica_desc`

No RPC signatures, wire layouts, or business logic are changed — **registration only**.

#### `mooncake-store/tests/rpc_handler_registration_test.cpp`

Add a focused regression test that exercises the **production master registration path**:

1. Start an in-process master via `InProcMaster` (uses `RegisterRpcService()` exactly like production).
2. Connect with `MasterClient`.
3. Call `CalcCacheStats()` over RPC and assert success.

This catches the exact failure mode reported in #3477 P0 (handler exists but is not registered).

Real-client handlers are not covered by a new end-to-end test in this PR because that would require standing up a full `RealClient` + `DummyClient` stack; the registration fix is one line each and mirrors existing patterns. Follow-up from #3477 P3 ("contract tests") could add symmetric coverage later.

### Explicit non-goals (left for future #3477 PRs)

- P1 protocol compatibility (`struct_pack::compatible`, stable method IDs, version negotiation)
- P1 moving blocking handlers off coro_rpc I/O threads
- P1 timed-out mutation retry semantics
- P2/P3 metrics completeness, payload bounds, full registration parity tables

### Compatibility / rollout notes

- **No breaking wire changes** — method signatures are unchanged.
- **Behavior change:** previously failing RPCs now succeed when the underlying implementation would have succeeded.
- Rolling upgrade safe: older clients that never called these methods are unaffected; callers that already attempted these RPCs will start working once master / real-client binaries include this fix.

Closes the P0 portion of #3477. Happy to follow up with broader registration parity tests if maintainers want that in the same issue series.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**

```bash
cmake --build build-standard -j16 --target rpc_handler_registration_test
ctest --test-dir build-standard -R rpc_handler_registration_test --output-on-failure

# Recommended broader Store regression (CI equivalent):
bash scripts/run_ci_test.sh
```

**Test results:**

- [x] Unit tests pass (new `rpc_handler_registration_test` added; full CI pending on GitHub Actions)

Local Windows environment cannot build Mooncake natively; validation is delegated to CI (same workflow as prior Store PRs).

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

Cursor AI was used for issue scoping against RFC #3477, implementation, test scaffolding, and PR description drafting. The submitter has reviewed the diff and can explain each registration change end-to-end.

### DIFF
--- a/mooncake-store/src/real_client_main.cpp
+++ b/mooncake-store/src/real_client_main.cpp
@@ -100,6 +100,9 @@ void RegisterClientRpcService(coro_rpc::coro_rpc_server &server,
     server.register_handler<&RealClient::batch_get_offload_object>(
         &real_client);
     server.register_handler<&RealClient::release_offload_buffer>(&real_client);
+    server.register_handler<&RealClient::batch_get_query_results>(&real_client);
+    server.register_handler<&RealClient::get_replica_desc>(&real_client);
+    server.register_handler<&RealClient::batch_get_replica_desc>(&real_client);
 }
 }  // namespace mooncake
 
--- a/mooncake-store/src/rpc_service.cpp
+++ b/mooncake-store/src/rpc_service.cpp
@@ -1814,6 +1814,8 @@ void RegisterRpcService(
         &wrapped_master_service);
     server.register_handler<&mooncake::WrappedMasterService::ServiceReady>(
         &wrapped_master_service);
+    server.register_handler<&mooncake::WrappedMasterService::CalcCacheStats>(
+        &wrapped_master_service);
     server.register_handler<
         &mooncake::WrappedMasterService::MountLocalDiskSegment>(
         &wrapped_master_service);
--- a/mooncake-store/tests/CMakeLists.txt
+++ b/mooncake-store/tests/CMakeLists.txt
@@ -102,6 +102,7 @@ add_test(
 set_tests_properties(object_checksum_client_test
                      PROPERTIES ENVIRONMENT "MOONCAKE_STORE_CHECKSUM=1")
 add_store_test(rpc_timeout_test rpc_timeout_test.cpp)
+add_store_test(rpc_handler_registration_test rpc_handler_registration_test.cpp)
 if(USE_CXL)
   add_store_test(cxl_client_integration_test cxl_client_integration_test.cpp)
 endif()
--- a/mooncake-store/tests/rpc_handler_registration_test.cpp
+++ b/mooncake-store/tests/rpc_handler_registration_test.cpp
@@ -1,0 +1,42 @@
+// Verifies P0 RPC handlers from RFC #3477 are registered on production paths.
+//
+// These methods already exist on WrappedMasterService / RealClient but were
+// missing from RegisterRpcService / RegisterClientRpcService, causing
+// "function not registered" failures at runtime.
+
+#include <gtest/gtest.h>
+
+#include "master_client.h"
+#include "master_metric_manager.h"
+#include "test_server_helpers.h"
+#include "types.h"
+
+namespace mooncake {
+namespace testing {
+namespace {
+
+class RpcHandlerRegistrationTest : public ::testing::Test {
+   protected:
+    void SetUp() override {
+        InProcMasterConfig config;
+        ASSERT_TRUE(master_.Start(config));
+    }
+
+    void TearDown() override { master_.Stop(); }
+
+    InProcMaster master_;
+};
+
+TEST_F(RpcHandlerRegistrationTest, CalcCacheStatsRpcIsRegistered) {
+    MasterClient client(generate_uuid());
+    ASSERT_EQ(client.Connect(master_.master_address()), ErrorCode::OK);
+
+    auto stats = client.CalcCacheStats();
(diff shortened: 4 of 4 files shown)


Aionw:
Thanks for PR.

The original RPC issue #3477 is mainly about how we can improve the RPC system as a whole, rather than simply exposing more RPC endpoints. While this PR fills in some missing endpoints, there are currently no consumers for them, so I’m not sure this alone addresses the original issue.

huangxiaobai-ydmy:
Hi @Aionw — thanks for the feedback on #3734. I agree that simply registering a few missing handlers does not address the broader #3477 goal.

I'd like to pivot to the systemic RPC improvement track instead of treating this as isolated bugfixes. My proposed incremental plan:

**PR 1 (P3 + P0 root cause):** RPC registration parity guard
- Enforce that every `MasterClient::invoke_rpc` target is registered in `RegisterRpcService()` (and similarly DummyClient ↔ RealClient).
- Consolidate handler lists into a single source of truth to prevent drift with `RpcNameTraits` / metrics metadata.
- Known missing handlers become fix-ups driven by the test, not ad-hoc patches.

I'll hold/close this PR unless you'd prefer keeping it open.

Does registration parity + a single registry sound like the right first step for #3477?

huangxiaobai-ydmy:
Closing in favor of the systemic #3477 track (registration parity guard). See https://github.com/kvcache-ai/Mooncake/issues/3477#issuecomment-5439332605 for the proposed plan.

PR #3738, opened 2026-08-27
[Bugfix] Fail SSD eviction tests on read errors

### LINKED ISSUE #3735: [Bug]: SSD offload correctness test stays green on empty or mismatched reads
### Bug Report

`mooncake-wheel/tests/test_ssd_offload_in_evict.py::TestDistributedObjectStore.test_put_get_in_evict_operations` records empty reads and byte mismatches in `get_stats.failure_count`, but never asserts that the failure count is zero.

During the verification phase:

- a mismatched value calls `get_stats.record_operation(False, ...)` and only prints a message;
- an empty value also calls `get_stats.record_operation(False, ...)`;
- after printing the statistics, the test proceeds to cleanup without a correctness assertion.

As a result, the unittest can remain green even when successfully written keys cannot be read back or return corrupted bytes. This is especially risky for SSD offload regressions such as the empty-read behavior reported in #3465.

Expected behavior:

- every successfully written key must be read back with byte-exact contents;
- any empty or mismatched read must fail the unittest;
- the final statistics should remain available for diagnostics.

A minimal fix could assert `get_stats.failure_count == 0` after the verification loop, with enough context to identify failed reads. A focused regression test should demonstrate that an empty or mismatched read produces a failing test result.

This existing test weakness was identified while reviewing PR #3669, which changes where the SSD test runs but does not modify the test implementation.

### Before submitting...

migarci2:
## Description

Fixes #3735.

The SSD eviction test now fails after cleanup when a read returns empty or corrupted data. The same assertion is applied to the batched batch_get_into path, which had the same false-green behavior. Failure messages retain the count and the first failed keys.

This is separate from #3669: that PR only changes where the test runs and explicitly leaves this assertion gap to #3735.

## Module

- [ ] Transfer Engine
- [ ] Mooncake Store
- [ ] Reshard
- [ ] Mooncake EP
- [ ] Mooncake PG
- [ ] Integration
- [ ] P2P Store
- [x] Python Wheel
- [ ] Common
- [ ] Mooncake RL
- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

Test commands:

    python -m py_compile mooncake-wheel/tests/test_ssd_offload_in_evict.py
    pre-commit run ruff --files mooncake-wheel/tests/test_ssd_offload_in_evict.py
    pre-commit run codespell --files mooncake-wheel/tests/test_ssd_offload_in_evict.py

Test results:

- [x] Unit-level verification passes
- [ ] Integration tests pass
- [x] Manual testing done

A focused harness ran the actual test method against valid, empty, and corrupted batch reads. Valid data passed; both bad-read cases failed after unregistering the buffer and removing all keys. A live SSD-backed Mooncake integration test was not available locally.

## Checklist

- [ ] I have formatted my code using ./scripts/code_format.sh
- [ ] I have run every pre-commit hook on the changed file
- [x] Documentation is not needed for this test-only change
- [x] I have added assertions that prove the change is effective
- [x] The change is below 500 LOC, so no RFC is needed

## AI Assistance Disclosure

- [ ] No AI tools were used
- [x] AI tools were used

Claude was used to review the two readback paths and help draft the assertions. I reviewed every changed line and verified the behavior described above.

### DIFF
--- a/mooncake-wheel/tests/test_ssd_offload_in_evict.py
+++ b/mooncake-wheel/tests/test_ssd_offload_in_evict.py
@@ -207,6 +207,7 @@ class TestDistributedObjectStore(unittest.TestCase):
         """ 
         VALUE_SIZE = 1024 * 1024 
         MAX_REQUESTS = 1000
+        KEY_PREFIX = "put_get_k_"
         reference = {}
         
         # Initialize statistics
@@ -219,7 +220,7 @@ class TestDistributedObjectStore(unittest.TestCase):
         put_stats.start_timer()
         index = 0
         while index < MAX_REQUESTS:
-            key = "k_" + str(index)
+            key = KEY_PREFIX + str(index)
             value = os.urandom(VALUE_SIZE)
             
             # Record operation start time
@@ -254,7 +255,7 @@ class TestDistributedObjectStore(unittest.TestCase):
         failed_keys = []
         index = 0
         while index < MAX_REQUESTS:
-            key = "k_" + str(index)
+            key = KEY_PREFIX + str(index)
             
             op_start = time.perf_counter()
             retrieved = self.store.get(key)
@@ -288,7 +289,7 @@ class TestDistributedObjectStore(unittest.TestCase):
         time.sleep(default_kv_lease_ttl / 1000)
         index = 0
         while index < MAX_REQUESTS:
-            key = "k_" + str(index)
+            key = KEY_PREFIX + str(index)
             self.store.remove(key)
             index = index + 1
         print("Cleanup completed")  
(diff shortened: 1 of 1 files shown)


Aionw (approved):
LGTM

fcczzz (approved):
narrow mitigationThe new post-cleanup assertions correctly make the single and batch readback paths fail on empty or corrupted data.

One adjacent false-green remains in `test_concurrent_stress`: its workers record empty/mismatched reads in the per-thread `get_stats`, but after combining those counters the test only prints them and returns successfully. A focused empty-read harness still lets this method pass with 4,000 failed reads.

This does not block the scoped #3735 fix, but could we add the same post-cleanup `get_stats.failure_count == 0` assertion here as well, or track it in a follow-up?

ykwd (approved):

zxpdemonio (approved):
LGTM.

Issue #3740, opened 2026-08-27
[Bug]: Store clients take ~2 minutes to reattach after a master failover, though promotion itself takes ~10s


gongwei-130:
### Bug Report

On `0.3.13` with `-enable_ha=true -enable_oplog=true -ha_backend_type=etcd`, master failover
itself works well: promotion is fast and **no metadata is lost** — this is a clear improvement
over `0.3.12.post1`. But the pool is unreadable for roughly two minutes afterwards, because the
store clients do not reattach for that long. The master serves an intact 200-object index to
nobody.

Reproduced twice, with near-identical timing both times.

### Setup

- Kubernetes, 1 etcd (v3.5.21), 2 `mooncake_master` (primary + hot standby), 3 `mooncake_client`
  store nodes contributing 200 GiB each (600 GiB pool)
- Image: `docker.io/kvcacheai/mooncake:0.3.13`
- Master: `-enable_ha=true -ha_backend_type=etcd -enable_oplog=true -etcd_endpoints=<etcd>:2379`
- Client: `-master_server_address=etcd://<etcd>:2379 -protocol=rdma -global_segment_size=200GB`
- Workload: 200 objects × 4 MiB, seeded and verified by a small deterministic probe
- Failover triggered with `kubectl delete pod <primary>`

### What works

```
17:49:57.752  Promoting Standby to Primary. Applied seq_id: 206, lag: 0 entries
17:49:57.755  Standby promoted to Primary successfully
17:50:02.755  Starting serve phase...
17:50:02.778  Restored from standby: 200 objects, 2 segments, initial_seq_id=206
```

~4 s from pod deletion to promotion, ~10 s to serving, `lag=0` throughout, and a verify run
after both failovers found **200/200 objects present and intact**.

### What does not

The new primary serves an index it has no clients for:

```
17:50:02.7   new leader serving — Keys: 200, Clients: 0, Mem Storage 0 B / 0 B
17:50:30     verify -> present+intact = 0/200, missing = 200
~17:52       Clients: 2, Mem Storage 800.00 MB / 400.00 GB    (~2 min after the kill)
```

So the metadata survived and the data survived, but for ~2 minutes every read fails.

### First cause: the view is published before the RPC port is bound

The new primary writes its `master_view` entry to etcd when it wins the election, then spends
~5 s in warmup before it binds `:50051`. Since #2484 made the client's view watch event-driven,
clients now wake up *inside* that window and connect to a port nobody is listening on. From a
[...]

PR #3743, opened 2026-08-28
[Store] Avoid blocking HA reattachment on master connection retries

waizuichougou:
## Description

Fix the HA client reattachment stall reported in #3740.

Related to #3740.

The failure has two independent causes that must both be addressed:

1. A newly elected master publishes its address before state restoration and
   before the RPC listener is ready. A client can therefore observe the new
   endpoint during the startup window and receive a connection reset. In the
   reported logs, the leader monitor switched to the new address about 0.73
   seconds before the service started accepting connections.
2. After attaching to a master, the HA client continues to use the initial RPC
   connection policy: a 30-second connect timeout with three retries. If the
   deleted Pod IP silently drops SYN packets, one Ping can block for about 120
   seconds. This matches the four reported failover gaps of approximately
   121.001 seconds.

This change fixes the readiness race while preserving leadership fencing:

- The elected master first writes a non-routable
  `__mooncake_service_warming__:<lease-id>` value into the existing,
  lease-bound `master_view` key.
- The key remains occupied while the master restores state and starts its RPC
  listener, so another master cannot acquire leadership during warmup while
  clients also cannot mistake the warming value for a ready endpoint.
- The routable address is published only after restoration, service
  registration, leadership renewal, and `async_start()` have completed.
- Readiness publication uses an etcd transaction that compares the exact
  warming value and preserves the existing lease. This prevents a stale or
  former leader from publishing itself.
- Publication and leadership-loss handling are serialized. If readiness
  publication fails, the RPC server is stopped and the supervisor returns to
  standby.

This change also separates initial connection behavior from HA runtime
behavior:

- Initial HA discovery and configuration loading retain the existing default
  30-second connection timeout and three retries. This preserves resilience
  while the cluster is initially starting.
- After initial attachment, the client enters HA runtime mode before the
  heartbeat and leader monitor start. Runtime connections use zero retries, no
  retry wait, and a one-second default connection timeout.
- `MC_RPC_CONNECT_TIMEOUT_MS` continues to override the default timeout.
- Reconfiguring the RPC client pool does not invalidate in-flight RPCs.
  Existing calls retain their previous pool instance, while subsequent calls
  use the runtime policy.
- Non-HA clients and store-to-store offload clients keep their existing
  connection behavior.

The change additionally prevents a stale empty `master_view` watch event from
clearing a newer view installed concurrently by the heartbeat path.

The warming value deliberately uses the same lease-bound election key instead
of a separate unprotected readiness key. This keeps election ownership and
readiness as distinct states while preserving the existing fencing and rolling
upgrade behavior.

## Module

- [ ] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [x] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Build:**

```bash
cmake --build build --target rpc_timeout_test high_availability_test -j2
```

**RPC connection-policy tests:**

```bash
build/mooncake-store/tests/rpc_timeout_test \
  --gtest_color=no --gtest_repeat=3 --gtest_break_on_failure
```

Results:

- 4/4 tests passed in each of three post-rebase runs.
- The initial HA connection policy retains its existing retry behavior.
- The same client switches to the bounded runtime policy after pool
  reconfiguration.
- Explicit timeout configuration remains covered.

**HA readiness and fencing tests:**

Run against a real etcd 3.5.21 instance on `127.0.0.1:2379`:

```bash
GTEST_FILTER='HighAvailabilityTest.*' \
GTEST_COLOR=no \
  build/mooncake-store/tests/high_availability_test \
  --etcd_endpoints=127.0.0.1:2379
```

Results:

- 14/14 tests passed.
- The tests cover the warming state, publication after service readiness,
  lease preservation, stale-publication rejection, idempotent publication,
  leadership-loss serialization, legacy contender fencing, and the stale
  empty-view race.

**Repository checks:**

```bash
pre-commit run --from-ref upstream/main --to-ref HEAD
```

- All checks passed locally for the changed files.
- The GitHub Sphinx documentation check passed.

## Checklist

- [x] I have updated the documentation (not applicable; no new user-facing
  option or API is introduced)
- [x] For changes >500 LOC: not applicable; this change is below the threshold

## AI Assistance Disclosure

- [ ] No AI tools were used
- [x] AI tools were used.

OpenAI Codex assistance was used for preparing the minimal change and regression test.


Aionw:
Disabling connection retries cannot fix the underlying problem; it only reduces the stall from multiple attempts to one and may make initial connections less resilient. 
I’d prefer separating election ownership from service readiness and publishing the leader endpoint only after the RPC listener is ready.

waizuichougou:
Thanks @Aionw . I’ll take another look at the code and keep this PR in draft for now. If the reporter shares more details in #3740, I’ll look into them and update the fix accordingly.

Aionw:
Thanks for your quick reply.

catyans:
Thanks @Aionw — the current head now separates election ownership from readiness: the lease-bound warming value fences contenders, and the endpoint is published only after the synchronous RPC bind, serialized with leadership loss.

issue: @waizuichougou, this branch now conflicts with main in `etcd_wrapper.go` and `etcd_helper.cpp`. Main added create-revision transaction compares and maintenance-session APIs; please preserve those while adding `preserve_lease`, then rerun full CI. The current head only has the Sphinx check.

waizuichougou:
@catyans  Thanks for the feedback. I’ve resolved the merge conflicts in `etcd_wrapper.go` and `etcd_helper.cpp`, while preserving the upstream create-revision transaction comparisons and maintenance-session APIs from `main`. The full CI has also passed.

catyans (approved):
LGTM. Re-reviewed the rebased head: it preserves the create-revision comparisons and maintenance-session APIs from `main`, while the fenced warming-to-ready transaction retains the election lease. The restored RPC service is published only after startup and is serialized with leadership loss. All 25 checks pass.

Icedcoco:
Thanks for the updates. The readiness/fencing direction now looks sound, and the new tests cover the core etcd transition well. Before merging, I would like to discuss a few possible behavioral regressions.

1. The HA runtime policy is applied to the shared `MasterClient` pool, so `connect_retry_count = 0` and the one-second connect timeout affect not only `Ping` and `SwitchLeader`, but also foreground RPCs such as `GetReplicaList`, `PutStart`, and `MountSegment`. Those calls do not appear to have the same outer retry loop as the HA control paths. Could we limit the fast-fail policy to HA control-plane reconnects, perhaps through a separate pool, so a transient connection-establishment delay does not immediately fail normal requests?

2. The leader-monitor thread is now started only after `GetStorageConfig()` and the related initialization work complete. If failover happens after the initial connection but before `EnterHaRuntimeMode()`, the client is still using the old leader and the original `4 x 30s` policy, while no leader monitor is running. Could we start leader monitoring immediately after the initial leader is established, while preserving the desired initialization retry behavior?

3. `serve_state_mutex` is held while `PublishServiceReady()` performs an etcd transaction with a timeout of up to five seconds. During that interval, the leadership-loss callback cannot acquire the mutex to stop the RPC server. If keepalive failure and readiness publication overlap, could the server remain listening, or even be activated briefly, after leadership has already been reported lost? Would it be possible to avoid holding the mutex across the remote etcd call and use an explicit serving-state transition or another non-blocking fencing check instead?

4. During warmup, `ReadCurrentView()` maps the lease-bound warming value to `nullopt`, and a newly starting client treats that as `UNAVAILABLE_IN_CURRENT_STATUS` and fails creation immediately. This turns the normal five-second warmup interval into a deterministic client-startup failure window. Could the initial HA connection wait, with a bounded deadline, for the warming view to become ready?

A few smaller behavior changes may also be worth confirming: `Reconfigure()` discards an already healthy pool and forces the next RPC to reconnect under the shorter timeout; Redis and K8s clients also inherit the one-second runtime policy even though their readiness semantics are unchanged; and a transient readiness-publication error now stops the listener and repeats the election/warmup cycle.

These may be acceptable tradeoffs, but I would prefer to clarify or cover the four cases above before merging. The fix for #3740 itself is pointed in the right direction; the concern is mainly avoiding regressions in otherwise healthy startup and request paths.

waizuichougou:
Hi @Icedcoco. Thanks for the detailed review. I agree that these cases could introduce behavioral regressions. The changes are addressed as follows:

1. The HA control-plane RPC paths are now separate from the foreground RPC path. `Ping` uses a fast-fail pool that targets the last confirmed leader, while `ServiceReady` uses a separate candidate-probe pool. Foreground RPCs such as `GetReplicaList`, `PutStart`, and `MountSegment` retain the original retry and timeout policy. A candidate is added to the heartbeat and foreground pools only after its readiness and version checks succeed, so a failed probe cannot redirect heartbeats away from the confirmed leader.

2. Leader monitoring now starts immediately after the initial leader connection is established. Failover can therefore be observed while `GetStorageConfig()` and the remaining client setup are still running. Master-dependent configuration and storage-bootstrap RPCs retain the original retry policy; the fast-fail control pools are enabled after that bootstrap phase and before transfer-engine setup.

3. `PublishServiceReady()` runs outside the local serving-state transition lock, so the leadership-loss callback is not blocked by the etcd transaction timeout and can stop the RPC listener promptly. Listener startup, final activation, and shutdown all use the same serving-state gate. This gives the local transitions a single ordering: a shutdown that wins rejects later activation, while a shutdown arriving after activation triggers deactivation immediately afterward. A publication overlapping leadership loss may still finish, but it cannot leave the service active; the normal cleanup path releases the leadership lease, with lease expiry as the backend cleanup fallback.

4. Initial HA discovery now waits up to 30 seconds when no ready view is visible. This covers both a leader that is still warming and the absence of an elected leader, since `ReadCurrentView()` intentionally exposes both cases as `nullopt`. A new client can observe the ready-value update instead of failing during the normal warmup window, while still returning `UNAVAILABLE_IN_CURRENT_STATUS` if no leader becomes ready before the deadline. The tradeoff is that startup with no leader may now wait up to 30 seconds rather than fail immediately.

`Reconfigure()` has also been removed, so enabling HA no longer discards a healthy foreground pool. The separate control/probe pools are used for HA clients regardless of backend, while foreground RPCs retain their original retry policy. Readiness-publication failures remain conservative: the listener is stopped and the election/warmup cycle is retried, because continuing to serve without successfully publishing and fencing the ready endpoint would be unsafe.

The regression coverage now includes the separated foreground/control retry policies, failed-candidate heartbeat isolation, bounded initial discovery, the warming-to-ready etcd transition, publication overlapping leadership loss, and activation/loss transition ordering.

Icedcoco (approved):
LGTM pending CI

Icedcoco:
Please rebase or merge the latest main and rerun CI. The current Ubuntu 22.04 wheel failure needs a fresh result against the updated base. Once CI is green, this can be merged.

waizuichougou:
Hi @Icedcoco. Thank you very much for the thorough review and helpful suggestions. I’ve rebased the PR onto the latest main and resolved the conflicts.

waizuichougou:
While resolving the rebase conflict, I integrated the upstream OpLog writer fail-stop callback with the PR’s serialized serving-state gate.
I also fixed a teardown lifetime issue found during the integration: the callback referenced the serving-state gate, but the gate could be destroyed before the service and its OpLog worker thread. The declaration order now guarantees that the service stops and joins its worker threads before the gate is destroyed.

catyans (approved):
Thanks @waizuichougou—re-reviewed the current head, and this still looks good to me. LGTM.

The follow-up work addresses @Icedcoco's concerns without weakening foreground traffic: HA heartbeat/candidate probes use their own fast-fail pools, initial discovery waits through the warming window, and readiness publication no longer holds the serving transition lock across etcd I/O. Activation and leadership loss still serialize through the same gate, so a late publication cannot resurrect a stopped service.

I also checked the rebase integration you called out. The OpLog terminal callback goes through that same shutdown gate, and `ServingStateGate` is declared before `WrappedMasterService`, so the service and its writer threads are torn down before the captured gate. That's the right lifetime ordering. With @Icedcoco's approval and all 25 checks green, this feels ready—thanks for taking the review feedback seriously and carrying it through the rebase.

waizuichougou:
@catyans Thanks for your patience and review.

ykwd (approved):

PR #3770, opened 2026-08-29
[P2P-Store] Release batch IDs on all performTransfer exit paths

chlins:
## Description

`performTransfer` allocated a batch ID at the top of each retry iteration but only released it on the successful completion path. Five paths leaked the ID:

- the `location == nil` break (allocated before the location check)
- `openSegment` failure
- `submitTransfer` failure
- `getTransferStatus` failure
- context cancellation

Batch descriptors live on the C++ side of the transfer engine, so a long-running process accumulates them over time.

This PR extracts the single attempt into `performTransferOnce` and releases the batch ID with `defer` on every path. On context cancellation the in-flight task is first drained to a terminal state, because `freeBatchID` returns BatchBusy (and keeps the descriptor) while any task is unfinished; the C API exposes no cancel/abort primitive, so draining is the only route to a freeable state. Original semantics are preserved: a successful transfer followed by a failed free still surfaces the error (named return), and the location is now checked before allocation, so exhausted locations no longer allocate at all.

## Related Components

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

**Test commands:**
```bash
go build ./...
gofmt -l .
go vet ./...
```

**Test results:**

`go build` and `gofmt` clean. `go vet` reports three pre-existing `unsafe.Pointer` warnings in `transfer_engine.go`, unchanged by this PR (verified identical on main).

Added `perform_transfer_test.go` with a fake `batchTransport` that mirrors the engine's BatchBusy semantics (free refused while a task is in flight):
- cancellation regression: task pending + ctx cancelled, asserts the batch is freed and the free is never attempted while busy
- completion path frees the batch
- submit-error path frees the batch

```bash
go test ./...   # locally with CGO_LDFLAGS="-Wl,-undefined,dynamic_lookup" since no native engine is linked
```
Removing the drain makes the cancellation test fail with "batch ID leaked on cancellation".

## Checklist

## AI Assistance Disclosure

AI assisted with the code audit and drafting. The submitter has reviewed, understands, and takes responsibility for all changes.

### DIFF
--- a/mooncake-p2p-store/src/p2pstore/core.go
+++ b/mooncake-p2p-store/src/p2pstore/core.go
@@ -366,48 +366,12 @@ func (store *P2PStore) performTransfer(ctx context.Context, source uintptr, shar
 	retryCount := 0
 	maxRetryCount := max(3, shard.Count())
 	for retryCount < maxRetryCount {
-		batchID, err := store.transfer.allocateBatchID(1)
-		if err != nil {
-			return err
-		}
-
 		location := shard.GetLocation(retryCount)
 		if location == nil {
 			break
 		}
 
-		targetID, err := store.transfer.openSegment(location.SegmentName, retryCount == 0)
-		if err != nil {
-			return err
-		}
-
-		request := TransferRequest{
-			Opcode:       OPCODE_READ,
-			Source:       uint64(source),
-			TargetID:     targetID,
-			TargetOffset: location.Offset,
-			Length:       shard.Length,
-		}
-
-		err = store.transfer.submitTransfer(batchID, []TransferRequest{request})
-		if err != nil {
-			return err
-		}
-
-		var status int
-		for status == STATUS_WAITING || status == STATUS_PENDING {
-			select {
-			case <-ctx.Done():
-				return ctx.Err()
-			default:
-				status, _, err = store.transfer.getTransferStatus(batchID, 0)
-				if err != nil {
-					return err
-				}
-			}
-		}
-
-		err = store.transfer.freeBatchID(batchID)
+		status, err := store.performTransferOnce(ctx, source, shard.Length, location, retryCount == 0)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


he-yufeng (commented):
Walked the five paths against the diff, and they all leak in the old shape: the batch ID was allocated before the `location == nil` break and freed only after the status loop completed, so `openSegment`, `submitTransfer`, `getTransferStatus`, and `ctx.Done()` all returned holding a descriptor. The `performTransferOnce` extraction with the deferred free covers each of them, and the polling semantics are unchanged (same zero-value status init, same `useCache = retryCount == 0`, same retry budget).

One deliberate choice worth calling out for a second opinion: on the completed path `err` is nil, so a `freeBatchID` failure turns a successful transfer into an error. I read that as intended (a leaked descriptor is exactly the bug being fixed, so surfacing it beats hiding it), but it does mean callers can now see a failure for a transfer whose bytes landed fine. If the project would rather log-and-forget there, the `if err == nil` branch is the one line to change.

chlins:
Thanks for the careful path-by-path check!

Yes, surfacing the free failure on the completed path is deliberate, and it also preserves the pre-existing behavior: on main, `freeBatchID` was called before the `status == STATUS_COMPLETED` check, so a free failure already turned a completed transfer into an error. This PR keeps that contract (via the `if err == nil` promotion) rather than introducing a new one.

I'd also lean toward surfacing it, since a silently failed free is exactly the leak this PR fixes. But if maintainers prefer log-and-forget there, happy to flip that one line.

catyans:
The cancellation path still leaks: `ctx.Done()` returns while the task is pending, so the deferred `freeBatchID()` returns `BatchBusy` and leaves the descriptor allocated; the helper only logs that error. Please drain or cancel the task to a terminal state before freeing (or hand it to an async cleanup owner), and add a cancellation regression that verifies the free succeeds.

chlins:
Good catch, thanks. Confirmed in `MultiTransport::freeBatchID`: any task with `!is_finished` returns BatchBusy and the descriptor stays allocated, so the cancellation path was still leaking.

Fixed in 7a66c1d:
- On `ctx.Done()` the task is now drained to a terminal state (`drainTransfer`) before the deferred free runs. The C API has no cancel/abort primitive, so draining is the only way to reach a freeable state; cancellation latency is bounded by the engine's transfer timeout. I went with a synchronous drain rather than an async cleanup owner to keep the batch lifecycle inside the call and avoid racing `Close()`. Happy to switch if you prefer the async shape.
- Extracted a small `batchTransport` interface (satisfied by `*TransferEngine`) so the batch lifecycle can be tested without the native engine.
- Added `perform_transfer_test.go`: a cancellation regression (task pending, ctx cancelled, asserts the free succeeds and is never attempted while busy) plus completion and submit-error cases. Removing the drain makes the cancellation test fail with "batch ID leaked on cancellation".

Side note: the test file avoids the `core_` prefix because the repo's `.gitignore` has a `core_*` rule for core dumps that swallows `core_test.go`.

catyans (approved):
Thanks @chlins, this addresses my concern. LGTM. Re-reviewed `7a66c1d`: cancellation now drains the submitted task to a terminal state before the deferred free, and the regression verifies that the batch is freed without a `BatchBusy` attempt. The completion and submit-error paths remain covered as well. CI is currently queued.

PR #3771, opened 2026-08-29
[Bugfix] Preserve complete objects after snapshot restore

### LINKED ISSUE #3761: [Bug] Snapshot restore reports success but drops every object: ApplySnapshotState evicts on restored (already-expired) leases, and every test skips that path (v0.3.13)
### Bug Report

On v0.3.13, a non-HA master with `enable_snapshot` + `enable_snapshot_restore` writes correct snapshots, reads one back after a restart, logs **`Successfully restored state from snapshot`** — and comes up with **zero objects**. No error, no warning.

The snapshot is not at fault. The objects are decoded and then erased by the post-restore cleanup pass in `ApplySnapshotState`, because restored read leases are wall-clock deadlines that are always expired by the time a master finishes restarting. The only way to skip that pass is an env var that exists **only in the snapshot unit tests**, which is why CI does not catch it.

### Environment

- `docker.io/kvcacheai/mooncake:0.3.13`
- Single master, `-enable_ha=false`, `enable_oplog` off
- `-enable_snapshot=true -enable_snapshot_restore=true -snapshot_interval_seconds=60 -snapshot_object_store_type=local -snapshot_catalog_store_type=embedded`, `MOONCAKE_SNAPSHOT_LOCAL_PATH=/snapshots` on a node-local persistent path
- 3 store clients, 3 x 200 GiB memory segments, `replica_num=1`
- `default_kv_lease_ttl` at its default (10000 ms)

### Repro

1. Start the master with the flags above.
2. Write 300 x 4 MiB objects. `master_key_count` = 300.
3. Wait for at least one snapshot interval. Snapshots are written correctly — `Snapshots: Success=6, Fail=0`, retention honoured, a forked child per snapshot:

```
I local_file_snapshot_object_store.cpp:47]  LocalFileSnapshotObjectStore initialized with path: "/snapshots"
I master_snapshot_manager.cpp:65]           [MasterSnapshotManager] Started
I master_snapshot_manager.cpp:97]           [Snapshot] snapshot_thread started
```

```
/snapshots/mooncake_master_snapshot/<cluster_id>/20260828_074040_060/
    metadata  segments  task_manager  manifest.txt  descriptor.txt
/snapshots/mooncake_master_snapshot/<cluster_id>/latest.txt
```

4. Restart the master and let the store clients reattach.

### What happens

```
I [Restore] Backend info: LocalFileSnapshotObjectStore: base_path=/snapshots
I master_service.cpp:11594]  Restored Replica::next_id_ to 301
I [Restore] Total allocated size after restore: 0
I [Restore] Total capacity size after restore: 644245094400
I [Restore] Successfully restored state from snapshot: 20260828_074040_060
```

`master_key_count` goes **300 -> 0** and stays 0 for the 160 s we watched, with all three clients reattached and full capacity back. Segments restore fine. Only the objects are gone, and the master reports success.

### The objects were in the snapshot and were decoded

`MasterSnapshotCodec` documents its `metadata` payload as "Metadata shards (objects, replicas, tenant state)". Payload sizes make the before/after unambiguous:

```
20260828_074040_060   metadata = 48,462 B   segments = 151,413 B   <- taken at keys=300
20260828_074147_596   metadata =      48 B  segments = 151,503 B   <- taken after, at keys=0
```

The restore read **exactly** `074040_060`, the 48 KB one. `Restored Replica::next_id_ to 301` proves the metadata deserializer ran and saw all 300 replicas. So encode is fine and decode is fine.

### Root cause: the post-restore cleanup evicts everything on expired leases

`MasterService::ApplySnapshotState` (`master_service.cpp:9583`) runs a cleanup pass immediately after decode:

```cpp
const bool skip_cleanup =
    std::getenv("MOONCAKE_MASTER_SERVICE_SNAPSHOT_TEST_SKIP_CLEANUP");
if (!skip_cleanup) {
    auto cleanup_now = now;
    ...
    if (it->second.HasDiffRepStatus(ReplicaStatus::COMPLETE) ||
        it->second.IsLeaseExpired(cleanup_now)) {
        it = EraseMetadata(tenant_state, it, tenant_it->first);
    }
    ...
}
```

`now` is `std::chrono::system_clock::now()`, taken in `RestoreState` at restart time. And `lease_timeout` is persisted as an **absolute epoch timestamp** and restored verbatim — `MetadataSerializer::SerializeMetadata` packs `metadata.lease_timeout.time_since_epoch()` (`master_service.cpp:11763`) and `DeserializeMetadata` reads it straight back (`master_service.cpp:11830`). It is never re-based on restore. The deadline therefore comes from before the snapshot was taken, and `default_kv_lease_ttl` defaults to **10000 ms** (`types.h:84`). A master restart — pod termination, scheduling, bind, snapshot download, decode — takes far longer than 10 seconds in any real deployment, and the snapshot itself is up to `snapshot_interval_seconds` old before that. So **every restored object is lease-expired at `cleanup_now` and every one of them is erased.**

The `Rebuild allocated memory metrics` loop then runs over what survived, which is nothing, hence `Total allocated size after restore: 0`. `RestoreState` sees `ApplySnapshotState` return success and logs `Successfully restored state from snapshot`.

**Why CI does not catch it:** `MOONCAKE_MASTER_SERVICE_SNAPSHOT_TEST_SKIP_CLEANUP` appears in exactly two files in the tree, both tests:

```
mooncake-store/tests/ha/snapshot/master_service_test_for_snapshot_base.h
mooncake-store/tests/ha/snapshot/master_service_ssd_test_for_snapshot.cpp
```

Every snapshot restore test disables the code path that destroys the restored state in production. The feature is only ever tested with the failing step turned off.

### What we would like

1. **Do not apply wall-clock lease expiry to restored objects.** A lease is a liveness hint for readers, not durability state. Re-granting `default_kv_lease_ttl` from `now` on restore, or skipping the lease check during restore entirely, would both give the intended behaviour.
2. **Do not report success when the restore produced zero objects out of a non-empty snapshot.** At minimum log a `WARNING` with decoded-vs-surviving counts; `Successfully restored state from snapshot` alongside `Total allocated size after restore: 0` is actively misleading.
3. **Cover the cleanup path in tests.** A restore test with a snapshot older than `default_kv_lease_ttl` and `skip_cleanup` unset would have caught this. As written, the env var makes the test suite structurally unable to see the bug.

### A second defect underneath

Even with the above fixed, ten seconds after restore `client_ttl=10` expires the restored segments' clients and unmounts all three (`action=unmount_expired_mem_segment`) while the live store pods re-register as new clients. A restored index would need segment identity to survive that churn to be usable.

### Operational note

A non-HA master does not publish `master_view`, so store clients cannot use `-master_server_address=etcd://...` and must be pointed at the master's address directly. Worth a line in the snapshot docs, since the non-HA path is the only one where snapshots are written at all today (see #3561).

migarci2:
## Description

Snapshot restore decoded complete objects correctly, then deleted them when their persisted read lease had expired. This changes the post-restore cleanup to remove only metadata with no replicas or with an incomplete replica. Complete objects remain available, while lease expiry stays an eviction hint rather than becoming a durability decision.

The regression test restores a complete object with a zero-length lease alongside an incomplete object. It verifies that the complete object is still readable and the incomplete one is still cleaned up.

Fixes #3761

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
cmake --build build-snapshot-restore --target master_service_test_for_snapshot -j 12
build-snapshot-restore/mooncake-store/tests/master_service_test_for_snapshot --gtest_filter=MasterServiceSnapshotTest.RestoreKeepsExpiredCompleteObject
build-snapshot-restore/mooncake-store/tests/master_service_test_for_snapshot --gtest_filter='MasterServiceSnapshotTest.PutStartEndFlow:MasterServiceSnapshotTest.RemoveLeasedObject:MasterServiceSnapshotTest.MountUnmountSegmentWithOffsetAllocator'
pre-commit run --files mooncake-store/src/master_service.cpp mooncake-store/tests/ha/snapshot/master_service_test_for_snapshot.cpp
```

The new regression failed against the original cleanup predicate because the complete key was missing after restore, then passed with this change. The three nearby snapshot/lease tests also pass.

**Test results:**
- [x] Unit tests pass
- [x] Manual testing done (fail-before/pass-after regression check described above)

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

An AI coding assistant helped trace the restore path, prepare the focused change, and run the checks. I reviewed the final diff and the test results.

### DIFF
--- a/mooncake-store/src/master_service.cpp
+++ b/mooncake-store/src/master_service.cpp
@@ -9629,21 +9629,23 @@ tl::expected<void, SerializationError> MasterService::ApplySnapshotState(
         segment_access.GetAllSegmentNames(segment_names);
     }
 
-    // Cleanup expired metadata (unless test environment disables it)
+    // Cleanup incomplete metadata (unless test environment disables it)
     {
         const bool skip_cleanup =
             std::getenv("MOONCAKE_MASTER_SERVICE_SNAPSHOT_TEST_SKIP_CLEANUP");
         if (!skip_cleanup) {
-            auto cleanup_now = now;
             for (auto& shard : metadata_shards_) {
                 for (auto tenant_it = shard.tenants.begin();
                      tenant_it != shard.tenants.end();) {
                     auto& tenant_state = tenant_it->second;
                     for (auto it = tenant_state.metadata.begin();
                          it != tenant_state.metadata.end();) {
-                        if (it->second.HasDiffRepStatus(
-                                ReplicaStatus::COMPLETE) ||
-                            it->second.IsLeaseExpired(cleanup_now)) {
+                        auto& metadata = it->second;
+                        const bool has_incomplete_replica =
+                            metadata.HasDiffRepStatus(ReplicaStatus::COMPLETE)
+                                .has_value();
+                        if (has_incomplete_replica ||
+                            metadata.CountReplicas() == 0) {
                             VLOG(1) << "clear metadata key=" << it->first;
                             it = EraseMetadata(tenant_state, it,
                                                tenant_it->first);
--- a/mooncake-store/tests/ha/snapshot/master_service_test_for_snapshot.cpp
+++ b/mooncake-store/tests/ha/snapshot/master_service_test_for_snapshot.cpp
@@ -75,6 +75,55 @@ std::string GenerateKeyForSegment(const UUID& client_id,
     }
 }
 
+TEST_F(MasterServiceSnapshotTest, RestoreKeepsExpiredCompleteObject) {
+    constexpr uint64_t kLeaseTtlMs = 60'000;
+    auto service_config = MasterServiceConfig::builder()
+                              .set_memory_allocator(BufferAllocatorType::OFFSET)
+                              .set_default_kv_lease_ttl(0)
+                              .build();
+    service_.reset(new MasterService(service_config));
+    const auto context = PrepareSimpleSegment(*service_);
+    const std::string complete_key = "restored_expired_lease";
+    const std::string incomplete_key = "incomplete_object";
+
+    ASSERT_TRUE(service_
+                    ->PutStart(context.client_id, complete_key,
+                               TenantId::Default(), 1024, {.replica_num = 1})
+                    .has_value());
+    ASSERT_TRUE(service_
+                    ->PutEnd(context.client_id, complete_key,
+                             TenantId::Default(), ReplicaType::MEMORY)
+                    .has_value());
+    ASSERT_TRUE(service_
+                    ->PutStart(context.client_id, incomplete_key,
+                               TenantId::Default(), 1024, {.replica_num = 1})
+                    .has_value());
+
+    const std::string snapshot_id = GenerateSnapshotId();
+    ASSERT_TRUE(CallPersistState(service_.get(), snapshot_id).has_value());
+
+    ::unsetenv("MOONCAKE_MASTER_SERVICE_SNAPSHOT_TEST_SKIP_CLEANUP");
+    auto restore_config = MasterServiceConfig::builder()
+                              .set_memory_allocator(BufferAllocatorType::OFFSET)
+                              .set_default_kv_lease_ttl(kLeaseTtlMs)
+                              .set_enable_snapshot_restore(true)
+                              .set_snapshot_object_store_type("local")
+                              .build();
+    std::unique_ptr<MasterService> restored_service(
+        new MasterService(restore_config));
+
+    auto replicas =
+        restored_service->GetReplicaList(complete_key, TenantId::Default());
+    ASSERT_TRUE(replicas.has_value());
+    ASSERT_EQ(replicas->replicas.size(), 1);
+    EXPECT_EQ(replicas->replicas.front().status, ReplicaStatus::COMPLETE);
+
+    auto incomplete =
+        restored_service->GetReplicaList(incomplete_key, TenantId::Default());
+    ASSERT_FALSE(incomplete.has_value());
+    EXPECT_EQ(incomplete.error(), ErrorCode::OBJECT_NOT_FOUND);
+}
+
 TEST_F(MasterServiceSnapshotTest, MountUnmountSegmentWithOffsetAllocator) {
     auto service_config = MasterServiceConfig::builder()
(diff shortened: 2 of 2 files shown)


he-yufeng (commented):
Read the diff and chased the lease path on current main. The fix is right, and the test pins it properly.

One thing worth stating precisely, since it decides what state the index is in after this lands: the codec persists the lease deadline as a timestamp (serialize at `master_service.cpp:11962`, `ExtendTo` at :12137), so a restored object keeps its original deadline rather than getting a fresh one. TTLs are seconds to minutes and a restart takes longer than that, which is why the old `IsLeaseExpired` cleanup erased effectively the whole index on any real restart. The `default_kv_lease_ttl(0)` in your test manufactures exactly that state, and on the old branch `complete_key` is erased by the cleanup and the `GetReplicaList` assertion fails, so the test goes red without the fix and green with it.

The cleanup that remains still covers incomplete and zero-replica husks, and expired survivors stay readable (reads re-grant through `lease_->GrantReadLease` on the read path) until normal eviction reclaims them. Sound as far as it goes.

The open question I'd want answered before landing: after this change, every restored COMPLETE object whose deadline passed during the downtime is immediately eviction-eligible. Under memory pressure the whole restored index can be evicted within minutes of coming up, which recreates the cold-cache outcome of #3761 on a delay. Should `ApplySnapshotState` re-grant a fresh lease from restore time for COMPLETE objects, or is immediately-evictable the intended steady state? If intended, a one-line comment at the cleanup site saying so would stop the next reader from filing it as a bug.

Icedcoco:
I recommend closing this PR rather than requesting changes. The core behavior is unsafe: restoring an expired COMPLETE object can expose stale metadata whose memory extent may have been reused, causing silent data corruption. Safely supporting this would require a broader snapshot design change, which is not justified for the legacy path given the #3808 roadmap and planned snapshot removal.

migarci2:
Thanks for the review. This needs a broader snapshot redesign, so I'll close the PR in favor of #3808.

PR #1389, opened 2026-01-16
[Store] Unified parallel tensor IO

zxpdemonio:
## Motivation                                                                                                                                                                      
                                                                                                                                                                                     
  This work is primarily driven by **weight synchronization optimization for RL workloads**. In RL systems, different components often consume the same model weights under different parallel layouts: some hold the full tensor, some only need a TP shard, and some run with mixed layouts such as DP+TP, EP+TP, or PP+TP. A storage path that only supports full tensors or plain TP shards causes unnecessary write amplification, unnecessary read amplification, and limited layout expressiveness. The goal of this update is to make the unified tensor IO layer suitable for RL-style weight sync by supporting richer layout identities while also optimizing both shard-oriented writes and shard/full reads.            
                                                                                                                                                                                     
  ## What is supported now                                                                                                                                                       
                                            
  The unified tensor IO path now supports not only heterogeneous **TP**, but also **DP**, **EP**, **PP**, and their combinations, including exact-match multi-axis shard identities such as **DP+TP**, **PP+TP**, and **EP+TP**. In other words, the system is no longer limited to “TP-only sharded tensors”; it can now persist and resolve multi-axis layout metadata through a single unified API surface.                                                                                                                                     
                                                                                                                                                                                     
  ## Write-path optimization                                                                                                                                                         
                                                                                                                                                                                     
  On the write side, the key optimization for RL weight sync is to avoid having every writer store the full tensor when only a local shard needs to be persisted. This is implemented through `writer_partition` for single-key APIs and `writer_partitions` for batch APIs. In practice, this allows each writer to hold a full tensor locally but write only the shard it owns, directly reducing write amplification and network/storage overhead during synchronization. The `_from` paths were also optimized to use **raw-buffer shard planning** instead of decoding tensor-object buffers into Python tensors and then reshaping or re-sharding them, which removes unnecessary decode, tensor materialization, and extra memcpy/clone work from the hot path.                                                                                                                                               
                                                                                                                                                                                     
  ## Read-path optimization                                                                                                                                                          
                                                                                                                                                                                     
[...]

PR #3772, opened 2026-08-29
[Reshard] Add Store-backed model-weight snapshots

Bo-Vincent:
## Description

This PR adds Store-backed model-weight snapshots for heterogeneous resharding.

`WeightStore` persists a complete, immutable `StoredWeightManifest` only after
all planned payload fragments have been uploaded and verified. The manifest is
address-free and retains the logical tensor/fragment layout required to plan a
later restore into a different target topology. Runtime GPU addresses,
allocation bounds, leases, and allocation guards stay in ephemeral runtime
bindings.

The Store path provides:

- generation-consistent source replica selection and upload planning;
- grouped payload and metadata records through Mooncake Store;
- commit/abort decisions, payload completeness checks, and recovery-safe
  registration cleanup;
- retry-safe manifest publication after a commit decision;
- range-based restore with `get_into_ranges` into validated target bindings;
- an explicit `WeightStoreWriter` lifecycle:
  `begin_weight_snapshot(...) -> write_tensor(...) -> commit()`.

### Breaking Change and Migration

This PR removes the public `*_with_parallelism` API family and the associated
`ParallelAxis`, `TensorParallelism`, and `ReadTarget` helper types.

Model-weight snapshot writers migrate to:

```python
writer = WeightStore(store).begin_weight_snapshot(descriptor, adapter)
writer.write_tensor(tensor_id, tensor)
manifest = writer.commit()
```

The writer stores manifest-managed payload fragments and publishes one
`StoredWeightManifest`; it does not create ordinary Store tensor objects.
After a manifest publication failure that follows the Store commit decision,
the writer preserves its payloads and remains available for `commit()` retry.
Snapshot readers migrate to `load_manifest()`, `plan_load()`, and `load()`
with target placement and runtime binding manifests. The existing single-axis
TP `*_with_tp` APIs remain separate compatibility APIs.

The native Store adapter requires `ReplicateConfig.group_ids`, supplied by the
merged Store group-semantics support. Transfer Engine lowering and live G2G
execution remain in the follow-up PR.

Related RFC: #3111

## Module

- [ ] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [x] Breaking change
- [x] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
PYTHONPATH=mooncake-wheel:mooncake-reshard/python \

# H20-8, with the matching native Mooncake Store build.
MOONCAKE_WEIGHT_GPU_STORE_E2E=1 \
  python -m pytest -q \
  mooncake-reshard/tests/weight_gpu_e2e/test_store_gpu.py \
  mooncake-reshard/tests/weight_all_axes_e2e/test_store.py
```

**Test results:**
- [x] Unit tests pass: `520 passed, 1 skipped`
- [x] Integration tests pass: Store GPU round-trip covered TP split/merge,
  TP/PP/EP/DP topology changes, and L3 restore.
- [x] Manual testing done: H20-8 native Store validation passed on the same
  PR3 source tree.

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex assisted with repository navigation, test orchestration, and draft
editing. The human submitter reviewed the implementation, owns the design
decisions, and is responsible for the submitted change.


zxpdemonio:
I focused this round on whether the `with_parallelism` APIs are actually converging onto the reshard / manifest path. I think the PR still needs to clarify or fix the following points before merge:

1. The existing `with_parallelism` API family has not actually been migrated to the reshard framework yet

   #3772 adds a session-bound facade, `WeightSnapshotSession.store.put_tensor_with_parallelism(...)`, which reuses part of the old call shape. However, the existing `mooncake.store` `put/get/upsert/batch/*_with_parallelism` implementations are not rewired by this PR. The old parallelism spec, tensor metadata/head construction, read reconstruction planner, and range executor still live in `mooncake-integration/store/store_py_internal.h`, `store_py_parallel_read.h`, and `store_py_parallel_write.h`.

   So I do not think this PR can currently claim that the `with_parallelism` family already reuses the reshard framework or removes the duplicated planner/executor/metadata implementation. Please either wire the existing public APIs into the reshard manifest/planner/executor path in this PR, or make the PR description and design docs explicit that this PR only adds a weight snapshot facade and leaves the old `with_parallelism` implementation independent for now.

2. `StoredWeightManifest` should be the single source of truth for weight snapshot reader reconstruction

   I do not think we should require this PR to preserve or reconstruct the exact old #1389 `TensorMetadata` shard-head semantics. The purpose of that metadata was to let the reader stitch shards correctly. After moving to manifest-managed snapshots, that reconstruction contract should belong to the manifest instead of keeping tensor-head metadata as a second authoritative shard source inside the payload.

   The current `StoredWeightManifest` looks sufficient for reshard transfer planning, but the reader reconstruction contract should be documented and tested explicitly: the reader should be able to reconstruct target tensors from the tensor descriptor, such as global shape / dtype / layout / parallel axes, plus fragment information such as `tensor_id`, `global_offset`, `local_shape`, `object_key`, `object_offset`, and `nbytes`. The payload should be treated as manifest-managed raw fragment bytes. Even if compatibility metadata exists, it must not participate in shard planning.

   Please add a test that restores a snapshot without relying on the old tensor head / `TensorMetadata`, proving that the weight snapshot reader can reconstruct solely from the manifest.

3. The new facade is not semantically equivalent to the old Store API and this needs to be explicit

   `WeightSnapshotSession.store.put_tensor_with_parallelism()` is not currently a drop-in equivalent of the old Store API: it only writes inside a snapshot session, returns `UploadReceipt`, requires a final `commit()` to publish the manifest, and rejects `config` / `writer_partition`. The old API writes ordinary Store tensor objects, returns integer status codes, and has the full batch / get / upsert / `_from` API family.

   Without documenting this boundary, users may assume both APIs produce the same kind of object and can be read back by the same `get_tensor_with_parallelism()` path. Please make the docs and tests explicit that this facade writes manifest-managed weight snapshot payloads, not ordinary Store tensor objects. If the goal is to migrate the old `with_parallelism` family onto the reshard framework, that migration should either be completed here or clearly split into a follow-up PR with the boundary documented.

zxpdemonio:
Thanks for the update. The new direction looks much clearer to me: the legacy `*_with_parallelism` public API is no longer exposed, and the snapshot path now goes through the explicit `begin_weight_snapshot(...)` / `write_tensor(...)` / `commit()` lifecycle. That seems fine if this is the intended migration path.

One thing still needs cleanup: the docs, tests, and PR description should describe the same API contract as the implementation.

Current inconsistencies I noticed:

- `mooncake-integration/store/store_py.cpp` removes the `ParallelAxis`, `TensorParallelism`, and `ReadTarget` Python bindings, and also removes all `get/put/upsert/batch/*_with_parallelism` bindings from `MooncakeDistributedStore`.
- The new tests agree with that implementation: `mooncake-wheel/tests/test_weight_snapshot_api.py` and `mooncake-reshard/tests/model_weight_store/test_snapshot_session.py` explicitly assert that the legacy `*_with_parallelism` methods and the old helper types are not exposed.
- But `mooncake-reshard/README.md` still says that the snapshot facade keeps the "parallel tensor write and batch-write method names and parameters unchanged inside the snapshot".
- The PR description also still says `WeightSnapshotSession` accepts the existing `put_tensor_with_parallelism(key, tensor, parallelism)` call shape.

Please update the stale docs / PR text to match the actual implementation. If the intended contract is the new explicit snapshot API, the docs should say that legacy `*_with_parallelism` callers migrate to `begin_weight_snapshot(...)`, `write_tensor(...)`, and `commit()`, rather than implying that the old method names still exist inside the session.

catyans:
issue: @zxpdemonio's API-contract cleanup is still incomplete on the current head. The implementation exposes `begin_weight_snapshot(...) -> write_tensor(...) -> commit()` and the tests assert that the legacy `*_with_parallelism` methods are absent, but the PR description still says the session accepts the existing `put_tensor_with_parallelism(...)` call shape, and `mooncake-reshard/README.md` still says those method names and parameters remain unchanged. The Python API reference also still says dummy clients support the unified parallelism write wrappers.

Please update those remaining descriptions to the actual snapshot lifecycle and mark/document the removal of the public legacy methods as a breaking change, including the migration path. The code/docs contract should be unambiguous before merge.

zxpdemonio:
Thanks, the PR description and the Python API reference now look aligned with the implementation and tests.

One stale sentence still seems to remain in `mooncake-reshard/README.md`: the Store Snapshots section says the snapshot facade keeps the "parallel tensor write and batch-write method names and parameters unchanged inside the snapshot". The implementation and tests now define the contract as `begin_weight_snapshot(...) -> write_tensor(...) -> commit()`, with the public `*_with_parallelism` methods removed.

Could you update that README sentence as well so it matches the new explicit writer lifecycle?

catyans:
Thanks @zxpdemonio and @Bo-Vincent. The current head also updates the remaining Reshard README text to the explicit `begin_weight_snapshot() -> write_tensor() -> commit()` lifecycle and documents the breaking migration from `*_with_parallelism`. This addresses my API-contract concern. The last wheel job is still running, but I have no further concern on this point.

zxpdemonio (approved):
LGTM.

stmatengss (approved):

PR #3776, opened 2026-08-30
fix(transfer-engine): cleanupIdleConnections() only inspected deque tail

CjiW:
## Problem

`cleanupIdleConnections()` only inspected `queue.back()` — the tail of each endpoint's connection deque. The while-loop popped expired idle connections from the back, but **broke immediately** when it hit either:
1. An `in_use` connection at the tail (common — newer connections are pushed to the back and may still be in use)
2. A non-expired idle connection at the tail

This meant any idle-expired connections **earlier in the deque** were never cleaned up. Over time, the pool grew unboundedly.

## Impact

Observed in production (sglang PD disaggregated inference, mooncake_tcp transport, 1P2D cluster):

| Metric | Value |
|--------|-------|
| ESTABLISHED connections (single prefill node) | **27,232** |
| Effective bandwidth | **~1.1 Gbps** (vs 36 Gbps iperf3) |
| KV transfer latency | **14.6s avg** (vs ~2s expected) |
| TTFT (first token) | **~24s** (vs ~3s expected) |

The 27K connections overwhelmed the kernel TCP stack, throttling bandwidth to 3% of available. This caused cascading latency: KV transfer 14.6s → bootstrap queue backup (164 requests) → TTFT 24s.

## Root Cause

```cpp
// BEFORE (buggy): only checks queue.back()
while (!queue.empty()) {
    auto& entry = queue.back();
    if (!entry->in_use && idle > timeout) { queue.pop_back(); }
    else break;  // ← BUG: stops at first non-expired or in_use
}
```

Connections are returned to the back of the deque via `returnConnection()`. When a newer `in_use` connection sits at the tail, the loop terminates immediately — every older idle connection at the front is permanently orphaned.

## Fix

```cpp
// AFTER: iterate ALL entries
for (auto qit = queue.begin(); qit != queue.end();) {
    auto& entry = *qit;
    if (!entry->in_use && idle > timeout) { qit = queue.erase(qit); }
    else ++qit;
}
```

## Performance

O(n) per call where n = pool size per endpoint (~hundreds). At ~10ns per entry comparison, cleanup costs ~microseconds — negligible vs the TCP transfer itself (seconds). The lock (`pool_mutex_`) is held for the same microsecond duration, so no contention impact.

## Additional Change

Made the idle timeout configurable via `MC_TCP_CONNECTION_IDLE_TIMEOUT` env var (default 60s, unchanged). A shorter timeout (e.g. 10s) helps under bursty workloads where the pool grows during peaks and needs to shrink quickly during lulls.

## Testing

- Based on v0.3.11.post1
- Observed in sglang PD with GLM-5.3 (78 layers, 8 TP, 8 CP, 1P2D)
- With the fix + connection pool enabled: ESTAB expected to stabilize at ~6,400 (peak concurrency × connections/request) vs 27K (buggy) or ~600 (pool disabled with 259K TIME_WAIT)

### DIFF
--- a/mooncake-transfer-engine/include/transport/tcp_transport/tcp_transport.h
+++ b/mooncake-transfer-engine/include/transport/tcp_transport/tcp_transport.h
@@ -142,7 +142,8 @@ class TcpTransport : public Transport {
                           std::shared_ptr<asio::ip::tcp::socket> socket);
     void cleanupIdleConnections();
 
-    static constexpr std::chrono::seconds kConnectionIdleTimeout{60};
+    // Configurable via MC_TCP_CONNECTION_IDLE_TIMEOUT env var (default: 60s)
+    std::chrono::seconds connection_idle_timeout_{60};
 };
 }  // namespace mooncake
 
--- a/mooncake-transfer-engine/src/transport/tcp_transport/tcp_transport.cpp
+++ b/mooncake-transfer-engine/src/transport/tcp_transport/tcp_transport.cpp
@@ -487,6 +487,15 @@ TcpTransport::TcpTransport() : context_(nullptr), running_(false) {
             enable_connection_pool_ = true;
         }
     }
+    // Allow tuning the idle timeout via env var (default: 60s).
+    // A shorter timeout reclaims idle sockets faster, reducing pool memory
+    // and fd usage under bursty workloads.
+    if (getenv("MC_TCP_CONNECTION_IDLE_TIMEOUT") != nullptr) {
+        int timeout_val = atoi(getenv("MC_TCP_CONNECTION_IDLE_TIMEOUT"));
+        if (timeout_val > 0) {
+            connection_idle_timeout_ = std::chrono::seconds(timeout_val);
+        }
+    }
 }
 
 TcpTransport::~TcpTransport() {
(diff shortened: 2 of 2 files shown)


catyans:
This patch no longer applies to the current TCP transport architecture. Current `main` has removed the deque-based dynamic `connection_pool_`, `PooledConnection`, `returnConnection()`, and `cleanupIdleConnections()` path entirely. Reusable transfers now go through fixed per-peer connection lanes with bounded work/admission queues, while `getConnection()` is only the pool-disabled one-shot path.

The original tail-only cleanup bug is real for the v0.3.11 implementation this branch was based on, but merging or conflict-resolving this patch must not reintroduce that retired dynamic pool. Could this PR be closed as obsolete if the fixed-lane design already addresses the production deployment, or re-scoped to a concrete retention problem reproducible on current `main`? If an idle-retirement policy is still needed for fixed lanes, it should be designed against `PeerConnectionGroup`/lane lifecycle rather than restoring `cleanupIdleConnections()`.

Separately, a full scan of every endpoint queue on every connection acquisition while holding one global mutex would have made the proposed old-pool fix O(total pooled sockets) on the hot path—27K entries per checkout in the reported deployment—so that implementation should not be carried forward even for a maintenance branch without rate limiting or periodic cleanup.

Issue #3750, opened 2026-08-28
[Bug]: TENT deadline predictors disagree on bandwidth, and the drop predictor multiplies worker-queue wait by the request's slice count


SongOf:
### Bug Report

**Environment:** `main` (observed at 69c22878), TENT RDMA transport with the runtime queue enabled. No hardware needed to reproduce the arithmetic — it is in the code paths below.

**Summary**

A request carrying `deadline_ns` is judged by its *predicted MLU* (predicted completion time / remaining window). Two layers make a decision from that number, and each computes it from a different bandwidth through its own copy of the formula:

| Layer | Decision | Bandwidth used |
|---|---|---|
| `LocalTransferAdmissionQueue::pickForDispatch` → `shouldDrop` (RFC #2519 step 3, #2764/#2816) | drops the owner → `CANCELED` | `RdmaTransport::getEstimatedBandwidth()` = sum of every device's **selection EWMA** (`DeviceSelector::getAggregateEwmaBandwidth`) |
| `Workers::asyncPostSend` deadline arbitration (RFC #2792, #2794) | orders QP posting | the constant `transports/rdma/default_bandwidth_gbps` |

So the same request is ranked on two unrelated scales, and the layer that can cancel it stands on the weaker number. Three concrete defects:

**1. The selection EWMA folds worker-queue wait into "bandwidth", and the drop predictor then multiplies that wait by the request's slice count.**

`Workers::asyncPollCq` feeds `DeviceSelector::release()` with `overall_lat = poll_ts - enqueue_ts` per slice, i.e. queue wait `q` plus wire time. For a slice of length `l` on a link of rate `r` the learned "bandwidth" is `l / (q + l/r)`. `shouldDrop` predicts a request of `L` bytes as `L / bw`:

```
L / bw = L · (q + l/r) / l = (L/l) · q + L/r
```

The wait is charged once per slice, but a request's slices wait behind the same backlog concurrently — they wait once. A 64-slice request is predicted to wait 64×. Large requests are therefore systematically over-estimated and preferentially dropped, and the admission queue's own dispatch window (which is what fills the worker queue) feeds back into its drop decision.

**2. The series is ~99% the latest sample.** `bandwidth_learning_rate` defaults to 0.01 and weights the *old* value (`new = α·old + (1-α)·observed`), so a single slow completion can flip an irreversible drop.

[...]

PR #3777, opened 2026-08-30
[TENT] Prefer the LAG-effective port speed from ibv_query_port_speed

SongOf:
## Description

  `RdmaContext::linkSpeedGbps()` decodes `active_speed × active_width`, the physical link's encoded rate. For a VF over VF LAG / multiport LAG, a PF dropping out of the bond halves the VF's bandwidth while its port stays `ACTIVE` at the same encoding, so the re-read on `IBV_EVENT_DEVICE_SPEED_CHANGE` (#3644) saw no change and never re-seeded the selector. rdma-core v62 added `ibv_query_port_speed()` alongside that event to report the *effective* speed (100 Mb/s units, LAG-aware).

  - `ibPortSpeedGbps()` in `ib_link_speed.h`: effective speed wins when > 0, the encodings decide otherwise.
  - `IbvLoader` resolves `ibv_query_port_speed` as an **optional** symbol (`nullptr` on older libraries; never disables RDMA).
  - `RdmaContext` queries it after every `ibv_query_port` (open and refresh), caches it, and `linkSpeedGbps()` prefers it. A present-but-failing verb records 0 so the fallback applies.
  - `RdmaContext::verbs_` becomes a per-context copy of the loader's table so tests can substitute entries (`RdmaContextTestPeer`), covering the effective-speed and refresh paths without an RNIC.

  No public API or config changes. Refs #3750 (follow-up section).

  ## Module

  - [x] Docs

  ## Type of Change

  - [x] Bug fix

  ## How Has This Been Tested?

  **Test commands:**
  ```bash
  cd build
  cmake --build . --target ib_link_speed_test tent_rdma_transport_test tent_device_selector_test tent_qos_contract_test bw_arbitration_test
  ctest -R 'ib_link_speed_test|tent_rdma_transport_test|tent_device_selector_test|tent_qos_contract_test|bw_arbitration_test' --output-on-failure

  Test results:
  - [x] Unit tests pass — 100% tests passed, 0 tests failed out of 5 (Ubuntu 22.04, rdma-core 39, no RNIC)

  New tests, each observed failing before the change:
  - ib_link_speed_test: effective speed beats the encoded rate (2000 → 200 Gbps over an NDR 4x link); falls back to the encodings when 0.
  - tent_rdma_transport_test: the optional symbol resolves iff this host's libibverbs exports it, with the mandatory verbs still resolved; with fake verbs injected through RdmaContextTestPeer: effective speed preferred, cleared when the verb fails, skipped when absent, refresh sees a renegotiated link, failed ibv_query_port keeps cached values; end-to-end through Workers::applyContextEvent: PORT_ACTIVE at the same speed keeps the learned EWMA, a LAG-halved effective speed re-seeds it to the new rate and the device stays available.

  Not covered without hardware: a real driver reporting a LAG-reduced speed (rdma-core ≥ 62 + Linux ≥ 7.0 + mlx5 + LAG).

  Checklist
  
  - [ ] For changes >500 LOC: I have filed an RFC issue (N/A, ~290 LOC incl. tests)

  AI Assistance Disclosure

  - [ ] No AI tools were used

### DIFF
--- a/mooncake-transfer-engine/tent/include/tent/transport/rdma/context.h
+++ b/mooncake-transfer-engine/tent/include/tent/transport/rdma/context.h
@@ -134,6 +134,18 @@ class RdmaContext {
 
     RdmaParams &params() const { return *params_.get(); }
 
+    // True while linkSpeedGbps() is derived from ibv_query_port_speed()
+    // rather than the encoded speed x width.
+    bool effectiveSpeedKnown() const {
+        return effective_speed_mbps_.load(std::memory_order_relaxed) > 0;
+    }
+
+    // ibv_query_port_speed() errors since the device was opened. The verb
+    // being absent, or reporting 0, is not an error.
+    uint64_t effectiveSpeedQueryFailures() const {
+        return effective_speed_query_failures_.load(std::memory_order_relaxed);
+    }
+
     // PCIe Relaxed Ordering support
     bool isRelaxedOrderingEnabled() const { return relaxed_ordering_enabled_; }
 
@@ -145,7 +157,10 @@ class RdmaContext {
     // Decode one ibv_query_port result into active_speed_/active_width_.
     void recordPortSpeed(const ibv_port_attr &port_attr);
     // Ask ibv_query_port_speed() for the effective speed when the library
-    // has it; records 0 otherwise so linkSpeedGbps() falls back.
+    // has it; records 0 when the verb is absent or reports nothing, so
+    // linkSpeedGbps() falls back. A verb *error* keeps the last known value
+    // instead: on a degraded LAG, falling back would briefly restore the
+    // higher encoded rate. Failures are counted and logged on transition.
     void queryEffectiveSpeed();
 
     // Release every resource currently owned by this context. This is
@@ -177,6 +192,9 @@ class RdmaContext {
     std::atomic<int> active_width_{0};
     // From ibv_query_port_speed(), converted to Mb/s; 0 = unavailable.
     std::atomic<uint64_t> effective_speed_mbps_{0};
+    std::atomic<uint64_t> effective_speed_query_failures_{0};
+    // Whether the last query errored; drives the transition logging.
+    std::atomic<bool> effective_speed_query_failing_{false};
     int gid_index_ = -1;
     ibv_gid gid_;
 
--- a/mooncake-transfer-engine/tent/include/tent/transport/rdma/context.h
+++ b/mooncake-transfer-engine/tent/include/tent/transport/rdma/context.h
@@ -46,6 +46,7 @@ class RdmaTransport;
 class RdmaContext {
     friend class RdmaCQ;
     friend class RdmaEndPoint;
+    friend class RdmaContextTestPeer;
 
    public:
     RdmaContext(RdmaTransport &transport);
@@ -191,7 +192,11 @@ class RdmaContext {
     // PCIe Relaxed Ordering support
     bool relaxed_ordering_enabled_ = false;
 
-    const IbvSymbols &verbs_;
+    // The context's own copy of the loader's verbs table (copied once at
+    // construction, read-only afterwards). A copy rather than a reference so
+    // tests can substitute individual entries and drive the port-attribute
+    // and event paths without an RNIC.
+    IbvSymbols verbs_;
 };
 
 }  // namespace tent
--- a/mooncake-transfer-engine/tent/include/tent/transport/rdma/context.h
+++ b/mooncake-transfer-engine/tent/include/tent/transport/rdma/context.h
@@ -113,7 +113,9 @@ class RdmaContext {
     // The one port this context opened; 0 for a slot that never constructed.
     uint8_t portNum() const { return params_ ? params_->device.port : 0; }
 
-    // Negotiated port speed in Gbps, 0 when it could not be determined.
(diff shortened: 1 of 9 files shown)


catyans (approved):
The implementation matches the upstream verbs contract: `ibv_query_port_speed()` uses the expected `(context, uint32_t port, uint64_t*)` ABI and reports 100 Mb/s units, so dividing by 10 produces Gbps. Resolving it with `dlsym` keeps older libibverbs versions compatible, while a missing, failed, or zero result falls back to the existing encoded speed path.

The runtime integration is also covered well: initial device open and port refresh update the effective value, stale effective data is cleared on query failure, and the event-chain test verifies that a real speed change reseeds the selector while an unchanged rate preserves learned EWMA state. CI is green. Approved.

Non-blocking follow-up suggestion: expose a counter or rate-limited warning for effective-speed query failures/fallbacks, since a transient failure on a degraded LAG can temporarily restore the higher encoded rate and would otherwise be difficult to diagnose.

mooncake-transfer-engine/tent/include/tent/transport/rdma/context.h
```diff
@@ -169,6 +175,8 @@ class RdmaContext {
     // atomic so a reader added elsewhere stays well-defined.
     std::atomic<int> active_speed_{0};
     std::atomic<int> active_width_{0};
+    // From ibv_query_port_speed(), 100 Mb/s units; 0 = unavailable.
+    std::atomic<uint64_t> effective_speed_100mbps_{0};
```
alogfans: What's the difference between this one and active_speed_? Additionally we'd prefer to store values as mbps.

SongOf: Done in 5fb0da4d: the context now stores plain Mb/s (`effective_speed_mbps_`, x100 at the store since the verb reports 100 Mb/s granularity), and `ibPortSpeedGbps()` takes `effective_mbps` (/1000). The fake verb in the tests still returns 100 Mb/s units like the real one, so `EffectiveSpeedPreferredWhenVerbReportsIt` (2000 -> 200.0 Gbps) covers the conversion end to end.

SongOf: To expand on the difference, with the upstream references:

**`active_speed_` vs `effective_speed_mbps_`.** `active_speed_` caches `ibv_port_attr::active_speed` from `ibv_query_port()` - an *encoding* of the per-lane signalling class (1/2/4/.../256 = SDR...XDR), which only becomes a rate after multiplying by the lane count from `active_width` (`ibLinkSpeedGbps()`). `effective_speed_mbps_` caches what `ibv_query_port_speed()` returns: the port's actual effective bandwidth as a number (the verb reports 100 Mb/s granularity; we store Mb/s since 5fb0da4d).

**Why a second verb at all.** From the rdma-core commit that introduced it (linux-rdma/rdma-core#1674, commit 8504597, first released in v62.0): "This is useful for VF LAG and multiport LAG configurations, since **ibv_query_port does not report the combined bandwidth of all aggregated ports**." Three concrete gaps in `ibv_query_port` that the new verb closes:
1. For a VF/port over LAG, `ibv_query_port` reports a single parent port's speed x width; `ibv_query_port_speed` reports the aggregate of the active bond members.
2. When a bond member fails, the VF's bandwidth drops (e.g. halves) while its port stays ACTIVE with the same encoding/width - invisible to `ibv_query_port`; the effective speed changes, and the paired `IBV_EVENT_DEVICE_SPEED_CHANGE` async event fires (we already consume that event to re-seed the selector).
3. Aggregated rates (e.g. 2x200G) have no speed-x-width encoding; the kernel series notes it supports "speeds not expressible as IB speed * width", which is why the verb returns a plain number instead of an encoding.

Kernel side: "[PATCH rdma-next 00/10] Support effective VF bandwidth query in LAG mode" adds `ib_query_port_speed()` and the speed-change event, with mlx5 propagating the aggregated LAG speed to attached vports and updating it on member link changes.

That is also why `linkSpeedGbps()` prefers the effective value and falls back to `active_speed_ x active_width_` only when the verb is absent (< rdma-core 62, resolved as an optional symbol) or reports 0.

SongOf:
Addressed the non-blocking follow-up in cc098853, in this PR:

1. A verb *error* now keeps the last known effective speed instead of reverting to the encoded rate (on a degraded LAG the fallback would briefly restore the higher rate and re-seed the selector at overstated capacity until the next port event). A verb that is absent or reports 0 still falls back as before.
2. Failure episodes are logged once per transition (WARNING with the errno-style rc and the held value; INFO on recovery), and the "link speed X -> Y, re-seeding" warning now names its source (effective speed vs encoded rate) so a fallback is no longer indistinguishable from a genuine renegotiation.
3. Per-device counter: `RdmaContext::effectiveSpeedQueryFailures()`.

Covered hardware-free via the fake-verbs fixture: `EffectiveSpeedHeldWhenVerbFails` (EIO holds 200G over two refreshes, counter 2, recovery at 400G picked up) replaces the old `EncodedRateWhenVerbFails` expectation, and `EncodedRateWhenVerbReportsZero` pins that a zero report is not an error.

catyans (approved):
LGTM. The update safely preserves the last known effective speed on transient query failures and adds useful observability and regression coverage. All checks are green.

alogfans (approved):

Issue #3371, opened 2026-08-10
[RFC]: Generalize `tebench` to Multi-Node Peer Benchmarking


zxpdemonio:
## Summary

This RFC proposes extending `tebench` from a single-target benchmark tool into a general multi-node peer benchmark framework for Mooncake Transfer Engine and TENT.

The core idea is to model every `tebench` process as a **peer**. A peer may publish local memory, initiate transfers to remote peers, or do both. Topologies such as fan-out, incast, M-to-N, all-to-all, and hotspot traffic become different configurations of the same peer model.

This proposal does not introduce application-specific concepts such as provider/consumer. M-to-N is treated as one topology where M initiator peers access N target peers.

## Motivation

Current `tebench` is effective for simple single initiator to single target benchmarking, but it is not sufficient for validating production-like distributed transfer behavior.

Large-scale deployments rarely behave like a single isolated connection. Real workloads involve multiple peers establishing connections at the same time, accessing multiple remote segments, sharing or isolating remote address ranges, and creating topology-specific pressure on both the control plane and data plane.

Several important classes of issues are difficult or impossible to expose with single-target benchmarking:

- RDMA bootstrap storms when many peers connect concurrently.
- Control-plane scalability bottlenecks during simultaneous metadata and connection setup.
- Endpoint lifecycle bugs during peer restart, segment reopen, or teardown.
- Data-plane admission and completion accounting issues under all-to-all traffic.
- Traffic fairness problems when many initiators target the same peer.
- Data correctness risks when multiple initiators share or overlap remote ranges.
- Premature buffer release when one peer exits while other peers still access it.
- Performance cliffs that only appear under fan-out, incast, or hotspot traffic.

Without multi-node benchmarking support in `tebench`, these problems tend to be discovered late by downstream integration tests or production-like fault injection. Adding this capability upstream gives Mooncake a common, reproducible way to validate connection scalability, transfer correctness, and topology-level performance.

## Goals

1. Support multi-target transfers in the `tebench` binary.
2. Support multi-node orchestration through a reusable script.
3. Keep the model generic: peer, role, topology, target plan, range policy.
4. Cover common distributed traffic patterns:
   - fan-out
   - incast
   - M-to-N
   - all-to-all
   - hotspot
   - custom target lists
5. Preserve existing single-target behavior by default.
6. Preserve existing `tebench` capabilities such as TENT backend, classic backend, QoS, deadline-aware workloads, and mixed memory configurations.
[...]

PR #3779, opened 2026-08-30
[TransferEngine] Add per-target metrics to multi-target tebench

jacklin78911-collab:
## What this PR does

This PR builds on the multi-target tebench support introduced in #3391 and adds per-target observability.

- Reports throughput, latency, p99/p999, operation count, and transferred bytes for each target
- Adds optional JSONL output through `--result_output_jsonl`
- Supports both classic Transfer Engine and TENT
- Preserves the existing single-target output
- Includes targets with no assigned operations, making load imbalance visible
- Adds focused unit tests and documentation

## Motivation

Aggregate metrics can hide imbalance between targets in multi-target benchmarks. Per-target results are needed to evaluate target selection and identify slow or underutilized peers before introducing more advanced placement or weighting policies.

## Validation

- Built with TENT enabled and disabled
- Passed existing and new unit tests
- Verified with a local two-target TENT smoke test
- Verified JSONL output
- Passed formatting, documentation, and spelling checks

Part of #3371.

### DIFF
--- a/docs/source/performance/mooncake/tebench.md
+++ b/docs/source/performance/mooncake/tebench.md
@@ -197,6 +197,20 @@ computed from each class's actual transfer size.
 `--qos_classes_json`, and the global `--tent_intent_type`. Non-default
 per-class intents and deadlines require the TENT backend.
 
+### 4.3 Per-Target Metrics
+
+Multi-target runs print one `[target-summary]` line per target. Use
+`--result_output_jsonl=<path>` to also append a schema-versioned JSON record for
+each benchmark configuration. The record keeps the aggregate operation, byte,
+and throughput totals plus each target's segment name, assigned thread count,
+completed operations, transferred bytes, throughput, and latency distribution.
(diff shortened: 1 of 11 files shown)


catyans (approved):
The per-target accounting is consistent with the existing benchmark model. Both backends expose the same modulo target mapping already used for segment and buffer lookup, the per-thread block/batch sizes feed the actual transferred-byte count, and all shared target/stat updates remain under the existing mutex. Per-target and aggregate throughput use the same pooled average worker duration as the current summary table, while unassigned targets remain visible with zero metrics.

The JSONL schema includes the aggregate totals and complete target breakdown, and output errors correctly fail the benchmark configuration. The focused metrics/serialization test and full CI are green. Approved.

Non-blocking follow-up suggestion: add an end-to-end mapping test with more targets than workers (for example, three targets and two threads) to verify the backend target selection and the emitted zero-assignment record together rather than only testing the calculator directly.

alogfans (approved):

Issue #2792, opened 2026-07-08
[RFC]: Deadline-aware NIC bandwidth arbitration within a priority tier (horizontal QoS)


catyans:
### Summary

TENT's QoS today gives strong **vertical** isolation — link-layer SL/TC (#2525/#2526), per-pool QP allocation (#2759), and SLO-layer deadline/EDF/drop (#2618/#2763/#2764) — all keyed on a request's **priority tier**. What's missing is **horizontal** fairness: when multiple flows in the *same* priority tier contend for one NIC, the NIC's bandwidth is split blindly/equally, with no way to give a tight-deadline flow a larger share.

This RFC proposes an opt-in, deadline-aware bandwidth arbitration within a priority tier, reusing the MLU signal already merged in #2618.

### Motivation (measured)

H20 / ConnectX (mlx5_bond_0), RoCEv2, 2-node, `ib_write_bw` 64 KB, varying concurrent QPs on the **same** NIC:

| concurrent flows (QP) | aggregate Gb/s | per-flow Gb/s |
|---|---|---|
| 1  | 194 | 194 |
| 2  | 387 | 193 |
| 4  | 389 | 97  |
| 8  | 389 | 49  |
| 16 | 389 | 24  |

The NIC saturates at ~388 Gb/s (a single QP can't fill it). Once flows exceed the saturation point, per-flow bandwidth drops as `~ceiling / N` — and the split is **equal/blind**. There is currently no mechanism to let a flow that is about to miss its deadline claim a larger slice of that shared 388 Gb/s.

### Current state

- `SharedSlotManager` provides a coarse, cross-process time-slice rotation (slot 0 = HIGH only, slot 1 = +MEDIUM, slot 2 = all), and `canSend(priority)` only sees the priority tier.
- Within one tier, all flows are treated identically on a contended NIC.

### Proposal (for discussion)

At the send-admission point (around `canSend` in the RDMA worker), introduce a **deadline dimension** for flows within the same tier: prioritize send tokens by predicted MLU (predicted transfer time / remaining deadline window), reusing the MLU computation from #2618.

- **Opt-in, default off** → behavior identical to today's equal split.
- Design options to weigh: weighted-fair-queue over flows, per-deadline token bucket, or MLU-threshold preemption.
- The arbitration *logic* is unit-testable; end-to-end bandwidth re-allocation needs a multi-flow RDMA setup (as with the #2519 MLU work).

### Relation to existing lines

[...]

PR #3780, opened 2026-08-30
[Bugfix][TENT] Predict deadline MLU from a metered wire rate and the bytes ahead

### LINKED ISSUE #3750: [Bug]: TENT deadline predictors disagree on bandwidth, and the drop predictor multiplies worker-queue wait by the request's slice count

SongOf:
## Description

Fixes #3750.

The admission queue's deadline-infeasible drop (RFC #2519 step 3) and the RDMA workers' bandwidth arbitration (RFC #2792) each computed "predicted MLU" from a different bandwidth: the drop from the sum of the selection EWMAs, the arbitration from the constant `default_bandwidth_gbps`. The selection EWMA is fed per-slice latency, so it folds queueing into a "bandwidth"; predicting an `N`-slice request from it multiplies that wait by `N` although the slices wait behind the same backlog concurrently, and one slow completion could flip an irreversible drop.

Both layers now predict completion time as `(bytes_ahead + length) / wire_rate` through one definition, `DeadlineMlu` (`tent/runtime/deadline_mlu.h`), restoring the in-flight term RFC #2519 specified. Review of the first version (thanks @alogfans, @catyans) found that neither input was what it claimed to be, and fixing that reached into the per-slice accounting. Three commits, each building and passing on its own:

1. `[Bugfix][TENT] Return every slice's charge and lane count to the counter that holds it` — the per-slice accounting below (retry re-charge, timeout release, shared-QP ownership), nothing the predictors read yet.
2. `[Bugfix][TENT] Meter the NIC's transmit rate and predict the admission drop from it and the bytes ahead` — `DeadlineMlu`, the transmit meter, `dispatching_bytes_`, `getEstimatedBandwidth()`.
3. `[TENT] Score each arbitration slot against the bytes actually posted ahead of it` — the arbitration reads the meter and the posted backlog, greedily per slot.

**Wire rate.** Each device gains a transmit estimate, `ewma_transmit_bps`, with the selection EWMA's update rule, clamp and link-speed seed, but fed from a **meter**: bytes completed over the time the NIC spent with work posted (`posted_bytes` / `busy_ns` / `busy_since`, `maybeSampleTransmit`), one sample per `transmit_meter_interval_ns` (10 ms), smoothed with `transmit_bandwidth_learning_rate` (0.9 ≈ 10 intervals). Per-completion latency cannot serve: work requests are posted in batches whose timestamps are effectively one and a poll pass stamps its completions alike, so a slice's own post→completion grows with batch depth (R, R/2, R/3, …). Busy time rather than elapsed time, because a workload that bursts and waits otherwise reads its duty cycle as the link rate; the sample is taken only on the last completion of a poll pass; a posted slice that ends without moving its bytes (failed, flushed, timed out) resets the interval. `getEstimatedBandwidth()` returns the aggregate transmit estimate; the arbitration reads the local NIC's. The selection EWMA is main's (#3511): a successful attempt's post→completion time.

**Bytes ahead.** The admission queue charges the bytes of drop-eligible owners it has dispatched and not yet completed (`dispatching_bytes_`, plus what the current pick already took); owners on other transports share the queue but not the NIC. The arbitration reads the NIC's **posted bytes** — what has reached the hardware — not the selector's allocation-time `inflight_bytes`, which already held the slices being ordered; `OrderByUrgency` then scores each slot behind the slots chosen before it (greedy for the first 64, one ranked pass for the rest).

**Per-slice accounting.** The meter and the arbitration only mean something if every charge comes back to the counter it was made on, and with `qp_pools` several worker lanes share a queue pair, so a completion or a timeout sweep on one lane routinely pops slices another lane enqueued. Every accounting a slice carries now names the counter it sits in — `charged_dev`, `posted_dev`, `counted_lane` (atomic, −1 = none) — and whichever path takes it out exchanges the field for −1 and pays back exactly what it read, so no path pays twice, none pays the wrong lane or device, and a fallback rewriting `source_dev_id` or a retry moving between lanes cannot misdirect a release. `owner_worker` routes set entries: a sweeper hands another lane's entry to that lane's reclaim list, drained on its next pass together with entries a teardown left behind. Retries are re-charged on the device they actually post on (`rechargeSlice`, charge-then-swap); a timeout returns the charge of the slice and of everything swept with it (`acknowledge` gained a per-slice callback); a slice resolved while it waited for a re-post is dropped rather than posted; everything a shared QP's poller must find (post timestamp, posted device, set entry) is written inside `submitSlices` before the WR reaches the wire. `selectMultiPath` charges each slice its real length, so release by length balances without a recorded charge size. `DeviceInfo` is 64-byte aligned with its hot counters on separate lines.

**Relation to #3511** (merged while this was in review): it fixed the retry/fallback charge with `charged_dev_id` + `charged_bytes` + a release-then-charge `chargeSliceQuota`. This PR keeps its own mechanism (the exchange discipline above is what a pooled QP's races need, and exact charging removes the need for `charged_bytes`) and follows #3511's semantics on the selection-EWMA sample and on baseline mode tracking no inflight. `quota_accounting_test.cpp` keeps #3511's selector-level cases; its slice-level cases live in `RdmaWorkersChargeTest`.

Known and deliberately left out: `submitSlices` and `acknowledge` hold only the endpoint's read guard, so on a pooled layout two lanes can reach one `BoundedSliceQueue` at once. That race predates this PR and needs a per-QP lock (noted in the `slice_queue_` comment); separate change.

No public API or wire-format changes. New config keys: `transports/rdma/transmit_bandwidth_learning_rate`, `transmit_meter_interval_ns`, `transmit_meter_max_interval_ns`.

## Module

- [x] Docs

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

**Test commands:**
```bash
cd build
cmake --build . -j4
ctest -R 'tent_|admission_queue|bw_arbitration' --output-on-failure
```

**Test results:**
- [x] Unit tests pass — 47/47 in the selection above; full `ctest` 94/98, the 4 failures need NUMA sysfs / a network the container lacks and fail identically on `main` (Ubuntu 22.04 container, rdma-core 39, no RNIC)

All new tests run without hardware. Those that pin a specific line were checked by flipping it (the test then fails).

- `deadline_mlu` / `bw_arbitration_test`: `DeadlineMlu` edge cases; each greedy slot is scored behind the ones before it; bytes ahead favour the tighter window.
- `admission_queue_test`: an owner is dispatched on an idle queue, dropped behind 1 MB of eligible work in flight, dispatched again once it completes; other-transport owners do not count; no bandwidth / no deadline never drops.
- `tent_device_selector_test`: the estimate is independent of batch depth (1..64 slices at one timestamp) and slice size; busy time counts only stretches with work posted; idle gaps and stale intervals are not samples; a reset re-baselines. `DeviceSelectorTransmitAccuracyTest` pins the number over ten successive stretches: a steady wire exact, a bursty one within 0.4 %, a duty cycle to 1e-6, a 4x slowdown followed within 478 ms.
- `tent_quota_accounting_test` (#3511's, re-pointed at the selector API): per-slice charge/release balance incl. a short trailing slice, `chargeDevice`, baseline mode keeps inflight at zero, EWMA gating.
- `tent_rdma_transport_test`:
  - `RdmaWorkersArbitrationTest`: `orderByDeadline` with a live selector — allocated contenders are not counted as queue-ahead, posted work outside the batch is.
  - `RdmaWorkersChargeTest`: a retry is re-charged and its completion teaches the selector; a fallback moves the charge; a charge on an untracked device keeps the old one.
  - `RdmaWorkersCompletionTest` (synthesized `ibv_wc` through `handleCompletion`): success feeds the meter; idle between bursts is not charged; a failed slice ends its stretch without teaching; only the last completion of a pass samples; a resolved slice still leaves the lane count; a COMPLETED sweep keeps the interval, a FAILED one abandons it.
  - `RdmaWorkersSharedQpTest` (a real `RdmaEndPoint` on fake verbs, one QP shared by two lanes): a timeout sweep on lane 0 gives lane 1 its slice back with both sets, counters and charges balanced; teardown leftovers are reclaimed by their lane; a retried slice stays counted and its completion is a transmit sample; a retry polled by the other lane moves ownership; a stale set entry is taken out by the lane holding it; a slice resolved while queued is dropped, not posted; a completion sweep keeps the meter interval.
  - `RdmaWorkersTimeoutTest`: a timed-out slice returns its charge with its endpoint gone or no longer holding it.

## Checklist

- [ ] For changes >500 LOC: I have filed an RFC issue (~1.1k LOC excluding tests; #3750 describes the change and the review comments above the rest — happy to file an RFC if wanted)

## AI Assistance Disclosure

- [ ] No AI tools were used

Claude Code was used for drafting code, tests and comments; every changed line was reviewed by the submitter.

### DIFF
--- a/mooncake-transfer-engine/tent/src/transport/rdma/workers.cpp
+++ b/mooncake-transfer-engine/tent/src/transport/rdma/workers.cpp
@@ -288,6 +288,16 @@ Status Workers::submit(RdmaSliceList& slice_list, int worker_id) {
     }
     auto& worker = worker_context_[worker_id];
 
+    // This lane's set and counter now account for the slices, wherever they
+    // are later swept from. First entry only: a slice arriving here has
+    // never been counted, so there is nothing to move.
+    auto* owned = slice_list.first;
+    for (int i = 0; i < slice_list.num_slices && owned; ++i) {
+        owned->owner_worker.store(worker_id, std::memory_order_relaxed);
+        owned->counted_lane.store(worker_id, std::memory_order_relaxed);
+        owned = owned->next;
+    }
+
     // Get priority from first slice (all slices in list have same priority)
     int priority = PRIO_HIGH;
     if (slice_list.first && slice_list.first->task) {
@@ -317,6 +327,22 @@ void Workers::submitFromTick(WorkerContext& worker, RdmaSlice* slice) {
     if (slice && slice->task) {
         priority = slice->priority;
     }
+    // This lane's queue and counter are about to account for the slice, so
+    // move the count here before it is visible on the queue. With a shared
+    // queue pair the lane re-queueing the slice is the one that polled the
+    // completion, not necessarily the one that first enqueued it: whichever
+    // lane still counts it is paid back in the same exchange, so the slice
+    // is never in two counts or in none. A sweep that already discounted it
+    // left -1 behind and there is nothing to pay.
+    if (slice) {
+        const int lane = laneIndex(worker);
+        slice->owner_worker.store(lane, std::memory_order_relaxed);
+        const int prev =
+            slice->counted_lane.exchange(lane, std::memory_order_acq_rel);
+        if (prev >= 0 && prev != lane && prev < (int)num_workers_)
+            worker_context_[prev].inflight_slices.fetch_sub(1);
+        if (prev != lane) worker.inflight_slices.fetch_add(1);
+    }
     // The worker must never block on its own queue (issue #3637): a full
     // queue parks the slice in requeue_overflow and the next tick retries.
     // Either way it stays counted as inflight, which keeps the worker from
@@ -324,7 +350,6 @@ void Workers::submitFromTick(WorkerContext& worker, RdmaSlice* slice) {
     if (!worker.queues[priority].try_push(slice_list)) {
         worker.requeue_overflow.emplace_back(priority, slice_list);
     }
-    worker.inflight_slices.fetch_add(1);
 }
 
 Status Workers::cancel(RdmaTask* task) {
@@ -347,82 +372,125 @@ Status Workers::cancel(RdmaTask* task) {
     return Status::OK();
 }
 
-bool Workers::cancelUnpostedSlice(WorkerContext& worker, RdmaSlice* slice) {
-    if (!slice || !slice->task ||
+void Workers::rechargeSlice(RdmaSlice* slice, int dev_id) {
+    if (!device_selector_ || !slice || dev_id < 0) return;
+    if (slice->charged_dev.load(std::memory_order_relaxed) == dev_id) return;
+    if (auto status = device_selector_->chargeDevice(dev_id, slice->length);
+        !status.ok()) {
+        // A device the selector does not track (a topology/quota mismatch).
+        // The transfer can still go, so the slice is not failed; it keeps
+        // the charge it has, if any, so the release stays balanced -- but
+        // this NIC's load is under-counted, so say so (rate-limited).
+        LOG_EVERY_N(WARNING, 100)
+            << "rechargeSlice: device " << dev_id
+            << " is not tracked by DeviceSelector; slice " << slice
+            << " keeps its previous charge: " << status.ToString();
+        return;
+    }
+    // Charge the new device before letting go of the old one, so a release
+    // racing in between pays back whichever device is on record and never
+    // both, and never one that has not been charged. No latency sample --
+    // this is not a completion.
+    const int prev =
+        slice->charged_dev.exchange(dev_id, std::memory_order_acq_rel);
+    if (prev >= 0) device_selector_->release(prev, slice->length, 0.0);
+}
+
+bool Workers::dropUnpostableSlice(WorkerContext& worker, RdmaSlice* slice) {
+    if (!slice || !slice->task) return false;
+    // Two reasons not to post. The caller cancelled the transfer; or another
+    // lane already resolved this slice while it waited here for a re-post --
+    // with a shared queue pair that lane's timeout sweep can still reach it
+    // through a set entry it has not drained. Posting a resolved slice would
+    // move bytes for a transfer that is already over.
+    if (slice->word == PENDING &&
         !slice->task->cancel_requested.load(std::memory_order_acquire))
         return false;
-    if (slice->word == PENDING) {
-        releaseSliceQuota(device_selector_.get(), slice);
-        updateSliceStatus(slice, CANCELED);
-    }
-    worker.inflight_slices.fetch_sub(1);
+    // Retire it the way a sweep would -- charge, count and set entry all go
+    // back where they came from, each idempotently, since whichever path
+    // resolved the slice may already have run them -- and updateSliceStatus
+    // is a no-op once the slice is terminal.
+    retireSweptSlice(worker, slice, nullptr);
+    updateSliceStatus(slice, CANCELED);
     return true;
 }
 
-void Workers::releaseSliceQuota(DeviceSelector* selector, RdmaSlice* slice,
-                                double latency) {
-    if (!slice || !slice->quota_charged || !selector) return;
-    // Release against the device the bytes were actually charged on, and unwind
-    // exactly the bytes that were charged, not the current routing NIC or the
-    // slice length: a fallback re-route can leave source_dev_id pointing at a
-    // different device, and the allocator's per-slice estimate can differ from
-    // slice->length.
-    selector->release(slice->charged_dev_id, slice->charged_bytes, latency);
-    slice->quota_charged = false;
-    slice->charged_dev_id = -1;
-    slice->charged_bytes = 0;
+void Workers::releaseSliceQuota(RdmaSlice* slice, double latency) {
+    if (!slice || !device_selector_) return;
+    // The field names the device it was charged to, so the release goes
+    // there even if a fallback has since rewritten source_dev_id.
+    const int charged =
+        slice->charged_dev.exchange(-1, std::memory_order_acq_rel);
+    if (charged >= 0)
+        device_selector_->release(charged, slice->length, latency);
 }
 
-void Workers::chargeSliceQuota(DeviceSelector* selector, RdmaSlice* slice) {
-    // Reconcile the inflight charge so it lands on the device the slice will
-    // actually post on (slice->source_dev_id) and reflects the slice's real
-    // length. This keeps DeviceSelector's per-NIC inflight view both symmetric
-    // with releaseSliceQuota and an accurate load signal, across three paths
-    // that would otherwise skew local telemetry:
-    //   - initial allocate: the aggregated allocator charges a per-slice
-    //     estimate (ceil(total/num_slices)); convert it to the exact length so
-    //     inflight equals the bytes really put on the wire.
-    //   - fallback re-route: the charge was made on the originally selected NIC
-    //     but the slice now posts on a different one -> migrate the charge.
-    //   - retry: the previous attempt released the quota, so re-enter the
-    //     inflight view instead of running uncounted.
-    if (!selector || slice->source_dev_id < 0) return;
-    if (slice->quota_charged && slice->charged_dev_id == slice->source_dev_id &&
-        slice->charged_bytes == slice->length)
-        return;  // already charged exactly this slice's length on the right NIC
-    if (slice->quota_charged) {
-        // Stale device and/or estimate -> unwind exactly what was charged
-        // first. Order is deliberate: releasing before charging makes inflight
-        // dip slightly below reality for the instant between the two calls,
-        // whereas charging first would transiently double-count during a
-        // fallback migration. inflight is only a scoring signal (not an
-        // admission gate), so a momentary dip merely perturbs one selection,
-        // while a double-count would over-penalize the NIC -- the dip is the
-        // lesser, intended bias.
-        selector->release(slice->charged_dev_id, slice->charged_bytes, 0.0);
-        slice->quota_charged = false;
-        slice->charged_dev_id = -1;
-        slice->charged_bytes = 0;
+Workers::WorkerContext* Workers::ownerContext(const RdmaSlice* slice) {
+    if (!worker_context_ || !slice) return nullptr;
+    const int lane = slice->owner_worker.load(std::memory_order_relaxed);
+    if (lane < 0 || lane >= (int)num_workers_) return nullptr;
+    return &worker_context_[lane];
+}
+
+int Workers::laneIndex(const WorkerContext& worker) const {
+    if (!worker_context_) return -1;
+    return static_cast<int>(&worker - worker_context_);
+}
+
+bool Workers::routeSetRemoval(WorkerContext& self, RdmaSlice* slice) {
+    // Two sets can hold an entry for one slice: a retry re-queued by another
+    // lane moves ownership while the lane it left still holds the old
+    // entry. So this lane takes out whatever it holds itself, and the
+    // owning lane, if it is somebody else, is handed the slice regardless --
+    // a hand-over for an entry it does not hold costs one erase of nothing.
+    const bool mine = self.inflight_slice_set.count(slice) > 0;
+    auto* owner = ownerContext(slice);
+    if (owner && owner != &self) {
+        std::lock_guard<std::mutex> lock(owner->reclaim_mutex);
+        owner->reclaimed.push_back(slice);
     }
-    // Charge this slice's real length on the routing NIC. Only commit the
-    // bookkeeping if the charge actually succeeded, so a failed charge is never
-    // released later (which would underflow the counter).
-    Status status = selector->chargeDevice(slice->source_dev_id, slice->length);
-    if (status.ok()) {
-        slice->charged_dev_id = slice->source_dev_id;
-        slice->charged_bytes = slice->length;
-        slice->quota_charged = true;
-    } else {
-        // The routing NIC is not tracked by DeviceSelector (a topology/quota
-        // mismatch, e.g. a device that never entered devices_). The data
-        // transfer can still proceed, so we do not fail the slice, but the
-        // slice runs without inflight accounting -- surface it (rate-limited on
-        // the hot path) instead of silently under-counting the device's load.
-        LOG_EVERY_N(WARNING, 100)
-            << "chargeSliceQuota: source_dev_id " << slice->source_dev_id
-            << " is not tracked by DeviceSelector; slice " << slice
-            << " proceeds without inflight accounting: " << status.ToString();
+    return mine || !owner || owner == &self;
+}
+
+void Workers::retireSweptSlice(WorkerContext& self, RdmaSlice* slice,
+                               std::vector<RdmaSlice*>* deferred) {
+    if (!slice) return;
+    releaseSliceQuota(slice);
+    discountFromOwner(self, slice);
+    if (!routeSetRemoval(self, slice)) return;
+    if (deferred)
+        deferred->push_back(slice);
+    else
+        self.inflight_slice_set.erase(slice);
+}
+
+void Workers::discountFromOwner(WorkerContext& self, RdmaSlice* slice) {
+    if (!slice) return;
+    // Several paths can reach one slice -- its own completion, another
+    // lane's sweep of a shared queue pair, this lane's timeout pass -- and
+    // only the first of them takes it out of the count: the field says
+    // which lane's, and comes back -1 for everyone after.
+    const int lane =
+        slice->counted_lane.exchange(-1, std::memory_order_acq_rel);
+    if (lane < 0) return;
+    const bool known = worker_context_ && lane < (int)num_workers_;
+    (known ? worker_context_[lane] : self).inflight_slices.fetch_sub(1);
+}
+
+void Workers::drainReclaimed(WorkerContext& worker) {
+    std::vector<RdmaSlice*> taken;
+    {
+        std::lock_guard<std::mutex> lock(worker.reclaim_mutex);
+        if (worker.reclaimed.empty()) return;
+        taken.swap(worker.reclaimed);
     }
+    for (auto* slice : taken) worker.inflight_slice_set.erase(slice);
+}
+
+void Workers::markPosted(WorkerContext& worker, RdmaSlice* slice,
+                         uint64_t post_ts) {
+    slice->submit_ts = post_ts;
+    worker.inflight_slice_set.insert(slice);
 }
 
 std::shared_ptr<RdmaEndPoint> Workers::getEndpoint(Workers::PostPath path) {
@@ -532,7 +600,7 @@ void Workers::asyncPostSend() {
         if (slice_list.num_slices == 0) continue;
         auto slice = slice_list.first;
         for (int id = 0; id < slice_list.num_slices; ++id) {
-            if (cancelUnpostedSlice(worker, slice)) {
+            if (dropUnpostableSlice(worker, slice)) {
                 slice = slice->next;
                 continue;
             }
@@ -540,13 +608,13 @@ void Workers::asyncPostSend() {
             if (!status.ok()) {
                 LOG(ERROR) << "Failed to generate post path for slice " << slice
                            << ": " << status.ToString();
-                releaseSliceQuota(device_selector_.get(), slice);
+                releaseSliceQuota(slice);
                 updateSliceStatus(slice, slice->task->cancel_requested.load(
                                              std::memory_order_acquire)
                                              ? CANCELED
                                              : FAILED);
-                worker.inflight_slices.fetch_sub(1);
-            } else if (cancelUnpostedSlice(worker, slice)) {
+                discountFromOwner(worker, slice);
+            } else if (dropUnpostableSlice(worker, slice)) {
                 slice = slice->next;
                 continue;
             } else {
@@ -566,7 +634,7 @@ void Workers::asyncPostSend() {
         if (slices.empty()) continue;
         slices.erase(std::remove_if(slices.begin(), slices.end(),
                                     [&](RdmaSlice* slice) {
-                                        return cancelUnpostedSlice(worker,
+                                        return dropUnpostableSlice(worker,
                                                                    slice);
                                     }),
                      slices.end());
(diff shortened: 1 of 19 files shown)


catyans:
The new arbitration `bytes_ahead` is not actually the number of bytes ahead of the candidates being ordered. `DeviceSelector::allocate()` charges every slice before it enters the worker queues, so by the time `asyncPostSend()` calls `getInflightBytes()`, that value already includes every slice in the current `slices` vector (and may include work queued behind it). Passing the full value to `PredictedMlu()` and then adding `flow.length` counts the current slice twice and treats all of its siblings as preceding it. This differs from the admission model, where `dispatching_bytes_` excludes the owner currently being evaluated and grows only after an earlier owner is selected.

There is a complementary undercount on retry: the CQ/error paths call `releaseSliceQuota()` before `submit(slice)`, which clears `quota_charged`, but `source_dev_id` remains non-negative. `selectOptimalDevice()` therefore skips `DeviceSelector::allocate()` on the retry, so the retry is never charged again; its eventual completion also cannot feed either EWMA because `releaseSliceQuota()` sees the flag already clear.

Please make the queue-ahead accounting represent only work that actually precedes a candidate, and re-establish the selector charge for each retried attempt. At minimum, arbitration should exclude the current candidate set from the pre-existing inflight term and then account for ordering explicitly. A production-path regression test should allocate multiple differently sized slices, pass them through worker arbitration, and verify each computed queue-ahead value; a second test should fail and retry a charged slice, then verify inflight remains balanced and the successful retry contributes a transmit sample.

mooncake-transfer-engine/tent/src/transport/rdma/workers.cpp
```diff
@@ -684,7 +719,13 @@ void Workers::asyncPollCq() {
                 (slice->submit_ts - slice->enqueue_ts) / 1000.0;
             double inflight_lat = (poll_ts - slice->submit_ts) / 1000.0;
             double overall_lat_sec = (poll_ts - slice->enqueue_ts) / 1e9;
-            releaseSliceQuota(slice, overall_lat_sec);
+            // Only a successful completion measures wire time; a burst of
+            // flush completions would drag the transmit estimate down for
+            // hundreds of completions.
+            double transmit_lat_sec = wc[i].status == IBV_WC_SUCCESS
+                                          ? (poll_ts - slice->submit_ts) / 1e9
+                                          : 0.0;
```
alogfans: **Blocking: the transmit EWMA still includes queueing within a posted batch**

`transmit_latency = poll_ts - submit_ts` is not a wire-only service-time sample when several WRs are posted together. Their `submit_ts` values are nearly identical, while completions arrive serially, so equal-sized WRs produce samples approximately `R, R/2, R/3, ...`. CQ polling delay is included as well.

Consequently, the learned rate decreases with batch depth and can converge near the lower clamp even on a healthy link. Since `bytes_ahead` is then added separately, this also counts part of the NIC/QP queueing twice and may cause false, irreversible deadline drops.

Could we instead derive the estimate from device-level completed bytes over elapsed time, or from non-overlapping service intervals? A regression test should post/model multiple equal-sized slices at the same timestamp and verify that the estimate remains approximately independent of batch size.

mooncake-transfer-engine/tent/src/transport/rdma/workers.cpp
```diff
@@ -639,34 +655,53 @@ void Workers::promoteTimedOutRequests(WorkerContext& worker) {
...
             worker.inflight_slices.fetch_sub(num_slices);
-            slice_to_remove.push_back(slice);
+            if (num_slices == 0) {
+                // No longer on the QP (a neighbour's retry popped it, with
+                // its count): nothing swept, so fail it here.
+                releaseSliceQuota(slice);
+                updateSliceStatus(slice, TIMEOUT);
+                slice_to_remove.push_back(slice);
+            }
         }
```
alogfans, line 900: **Blocking: timeout cleanup is still attributed to the wrong worker with pooled QPs**

`RdmaEndPoint::acknowledge()` may sweep slices belonging to multiple workers when they share a pooled QP. However, `expireTimedOutSlices(worker, ...)` adds every swept slice to the current worker’s `slice_to_remove` and subtracts the entire returned count from that worker’s `inflight_slices`.

A swept slice owned by another `WorkerContext` can therefore remain in its owner’s `inflight_slice_set`, while the current worker’s counter is over-decremented. If the QP is destroyed before its CQEs are polled, the stale entry may never be removed. Making `quota_charged` atomic prevents a duplicate selector release, but it does not fix worker-local ownership/accounting.

Please route cleanup to each slice’s actual owner, for example by recording the owner worker ID and scheduling removal on that worker. This needs a two-worker regression test where both workers share a QP and a timeout on worker A sweeps a slice owned by worker B; both sets, counters, and selector charges should end balanced.

mooncake-transfer-engine/tent/src/transport/rdma/workers.cpp
```diff
@@ -506,13 +514,20 @@ void Workers::asyncPostSend() {
...
-                                      ? device_selector_->getSchedulingParams()
-                                                .default_bandwidth_gbps *
-                                            1e9 / 8.0
+                                      ? device_selector_->getTransmitBandwidth(
+                                            path.local_device_id)
                                       : 0.0;
+            const size_t bytes_ahead =
+                device_selector_
+                    ? device_selector_->getInflightBytes(path.local_device_id)
+                    : 0;
```
alogfans: **Blocking: `getInflightBytes()` does not represent bytes ahead of these contenders**

The selector charges every slice during allocation, before it is posted. Therefore, when `asyncPostSend()` reads `getInflightBytes(local_device_id)`, the value already includes the slices currently being sorted, other unposted slices, and potentially work that will be scheduled after the candidate. `DeadlineMlu()` then adds the candidate length again.

This is not the documented `(bytes_ahead + length) / wire_rate` model and can materially change the ordering because the common overcharge is divided by different deadline windows.

Please track the posted/WQ backlog separately, or subtract the current contenders with well-defined saturating semantics. The test should exercise the live selector-to-worker path rather than passing a synthetic `bytes_ahead` directly to `PredictedMlu()`.

mooncake-transfer-engine/tent/src/transport/rdma/workers.cpp
SongOf: Fixed in 80d96aa4 (rebased on main; the branch is now three commits). The per-slice latency sample is gone from the transmit estimate; it is a meter over device-level counters: bytes completed divided by the time the NIC spent with work posted (`DeviceInfo::posted_bytes` / `busy_ns` / `busy_since`, `DeviceSelector::maybeSampleTransmit`). Queueing therefore only enters the prediction once, through `bytes_ahead`.

Two details worth knowing:
- Busy time rather than elapsed time, because a bursty workload otherwise reads its duty cycle as the link rate (16 MiB at 8 GB/s every 10 ms settled at 1.68 GB/s with elapsed time). Busy stretches open/close on the `posted_bytes` 0↔non-zero transitions, so the cost is one timestamp per burst edge.
- The sample is offered only on the last completion of a poll pass (`handleCompletion`'s `last_in_pass`): a pass stamps everything it reaps with one timestamp, and an interval closing partway through a pass paid no time for the rest of it (8% high at 2 GB/s with 16-deep polling).

The *selection* EWMA keeps main's per-completion sample as #3511 defined it (a successful attempt's post→completion time); only the deadline predictors read the meter.

Tests: `DeviceSelectorTransmitTest.RateIsIndependentOfBatchDepth` posts 1..64 equal slices at one timestamp and requires the same estimate for every depth; `SliceSizeDoesNotSkewTheEstimate`, `IdleTimeIsNotChargedToTheLink`, `RdmaWorkersCompletionTest.OnlyTheLastCompletionOfAPassSamples` cover the rest, and `DeviceSelectorTransmitAccuracyTest` pins the number over ten successive stretches (steady wire exact, bursty within 0.4%, 4x slowdown followed in 478 ms).

Residuals: a stretch closes when its last completion is *polled*, so that one CQE's polling delay is inside the busy time; and after a failure/timeout the meter restarts its interval (`resetTransmitMeter`), so the first sample after it only re-baselines — a link learns again from the second completion on.

mooncake-transfer-engine/tent/src/transport/rdma/workers.cpp
SongOf: Fixed in dd161d67. Cleanup is routed per slice now; nothing is bulk-subtracted from the sweeping worker any more.

- Every accounting a slice carries names the counter it sits in: `counted_lane` (which lane's `inflight_slices`), `charged_dev`, `posted_dev`; -1 = none. Whichever path takes the slice out — its own completion, another lane's sweep of the shared QP, the timeout pass — exchanges the field for -1 and pays back exactly what it read; a second path reads -1 and does nothing. A retry that lands on another lane moves the count in one exchange (`submitFromTick`).
- `owner_worker` names the lane whose `inflight_slice_set` holds the entry. A sweeper that pops someone else's slice hands it to that lane's `reclaimed` list (`routeSetRemoval`), and the owner erases it on its next pass (`drainReclaimed`). Entries left behind when a QP is destroyed before its CQEs are polled are retired by the owner's pass as well (`expireTimedOutSlices` retires non-PENDING leftovers).

On the rebase this replaces #3511's `charged_dev_id`/`charged_bytes`/`chargeSliceQuota`: that reconcile releases and then re-charges non-atomically, which is exactly the window a pooled QP's poller can hit, and with `selectMultiPath` charging each slice its real length there is nothing left for `charged_bytes` to correct. `quota_accounting_test.cpp` keeps #3511's selector-level cases (per-slice balance, baseline mode tracks no inflight, EWMA gating, charge on an untracked device); the slice-level ones (fallback migration, retry re-charge, failed charge) moved to `RdmaWorkersChargeTest`, where a `Workers` is built without hardware.

Tests (`RdmaWorkersSharedQpTest`, no hardware: a real `RdmaEndPoint` constructed on fake verbs, one QP shared by two lanes): `TimeoutSweepGivesTheOtherLaneItsSliceBack` has lane 0's timeout sweep lane 1's slice and asserts both sets, both counters and the selector's inflight/posted bytes end at zero; `TeardownLeftoversAreReclaimedByTheirLane`, `RetryPolledByAnotherLaneMovesOwnership`, `StaleSetEntryIsTakenOutByTheLaneHoldingIt`, `SliceResolvedWhileQueuedIsDroppedNotPosted` cover the other orderings.

One thing this does not fix, on purpose: `submitSlices` and `acknowledge` hold only the endpoint's read guard, so on a pooled layout two lanes can reach one `BoundedSliceQueue` concurrently. That is a data race in the queue itself and needs a per-QP lock; I've left it out of this PR (noted in the `slice_queue_` comment) and will do it separately.

mooncake-transfer-engine/tent/src/transport/rdma/workers.cpp
SongOf: Fixed in 23d36017 (the posted backlog itself comes with 80d96aa4). The posted backlog is tracked separately: `DeviceInfo::posted_bytes` is charged when a work request reaches the hardware (`Workers::markPosted`, inside `submitSlices` before `ibv_post_send`) and returned when the slice leaves the QP. `orderByDeadline` reads `getPostedBytes()` as `bytes_ahead` — none of the contenders is in it, nor is anything still sitting in a worker queue — and the greedy pass then adds each chosen slice's length before scoring the next slot, so the model is literally `(bytes_ahead + length) / rate` per slot.

Tests go through the worker's own ordering function with a live `DeviceSelector` rather than a synthetic `bytes_ahead`: `RdmaWorkersArbitrationTest.QueueAheadExcludesTheSlicesBeingOrdered` allocates the contenders (so `getInflightBytes` holds them) and checks the order is the one for zero queue-ahead; `QueueAheadCountsWorkOutsideTheBatch` posts unrelated work first and checks the order flips the way the extra bytes predict.

SongOf:
@catyans both halves are fixed: queue-ahead in 23d36017, the retry re-charge in dd161d67 (rebased on main, which picked up #3511 in the meantime — see the note at the end).

**Queue ahead.** Arbitration no longer reads the allocation-time charge. `bytes_ahead` is the device's posted backlog (`getPostedBytes`, charged when a WR reaches the hardware, returned when it leaves the QP), which by construction excludes the candidate set and anything still queued behind it. Ordering is then explicit: the greedy pass scores each slot against `bytes_ahead` plus the lengths of the slots chosen before it, the same `DeadlineMlu` the admission path uses with `dispatching_bytes_`. `RdmaWorkersArbitrationTest.QueueAheadExcludesTheSlicesBeingOrdered` / `QueueAheadCountsWorkOutsideTheBatch` drive differently sized allocated slices through `Workers::orderByDeadline` with a live selector; `BwArbitrationTest.EachSlotIsScoredBehindTheOnesBeforeIt` pins the per-slot accumulation. They assert the resulting order (the comments give the MLUs each slot must have been scored with), not the intermediate values.

**Retry.** Every retry (`retry_count > 0`) goes through `selectFallbackDevice`, which now calls `rechargeSlice` for the device the attempt will actually use, so the charge is re-established each attempt; the release on completion pays back the device recorded in `charged_dev` rather than whatever `source_dev_id` says by then. The transmit estimate itself no longer depends on the charge at all — it is metered from posted bytes over busy time — so a retry's completion is a sample like any other. `RdmaWorkersChargeTest.RetryIsRechargedAndStillLearns` fails and re-charges a slice and asserts the selector's inflight balance and that the release teaches the selection EWMA; `RdmaWorkersSharedQpTest.RetriedSliceIsStillCountedByItsLane` runs the failure→re-queue→re-post→success sequence through `handleCompletion` and asserts the lane count, inflight/posted bytes return to zero and the retry's completion yields the expected transmit sample.

**Relation to #3511.** It fixed the retry/fallback charge with `charged_dev_id` + `charged_bytes` + `chargeSliceQuota` (release, then re-charge). This PR keeps its own mechanism instead — `charged_dev` is exchanged for -1 by whichever path releases it, so a pooled QP's poller cannot race the reconcile, and `selectMultiPath` charges each slice its real length so there is no estimate for `charged_bytes` to correct — and follows #3511 on the two semantics it set: the selection EWMA samples a successful attempt's post→completion time, and baseline mode tracks no inflight. #3511's `quota_accounting_test.cpp` is kept with its cases re-pointed at the selector API; the slice-level ones live in `RdmaWorkersChargeTest`.

catyans (approved):
LGTM. Re-reviewed `ab129170`: queue-ahead now uses the posted NIC backlog, retries are recharged, and shared-QP cleanup is returned to the owning worker. The added tests cover these accounting and rate-meter cases. Thanks @SongOf and @alogfans for the thorough follow-up. Please wait for CI before merge.

alogfans (approved):
LGTM

he-yufeng (commented):
Read through the three commits, mainly to check this against what I verified in #3750 (defects 2 and 3 on current main).

Both defects are closed at the mechanism level:

- The drop predictor no longer folds queueing into bandwidth. `DeadlineMlu` charges queueing as an additive `bytes_ahead` term over the metered wire rate, and `pickForDispatch` passes `dispatching_bytes_`, which already includes the owners picked earlier in the same call. An N-slice request no longer multiplies one shared wait by N.
- The arbitration side now orders against `getTransmitBandwidth(dev_id)` instead of the constant, with `getPostedBytes` as the queue ahead. `Step3DropCountsBytesAlreadyDispatched` pins the exact scenario from the issue: a tight owner drops behind a dispatched 1 MB, and stops dropping once that megabyte completes.

I also traced the new counter for leak paths, since `dispatching_bytes_` is only balanced if every Dispatching exit decrements: `cancel()` rejects owners already dispatching, `retireBatch` refuses non-terminal owners, and `dropOwner` only ever sees Queued owners, so `complete()` is the single exit and the decrement is symmetric with the increment (both gated on `degradation_eligible`). Same care visible in `rechargeSlice`: charging the new device before releasing the old one means a racing release pays back whichever device is on record, never both and never neither.

One thing I would like confirmed rather than changed. The two layers deliberately read different `bytes_ahead`: the drop side counts dispatched-and-not-completed (which includes slices still sitting in worker queues), while the arbitration side counts posted-to-hardware only. The net effect is that the irreversible decision is the conservative one, which is the right way around. But the comment on `DeadlineMlu` says both layers "rank the same request the same way", and with different inputs they only approximately do. Maybe worth one sentence in that header saying the drop side is intentionally the conservative reading, so nobody later "fixes" the asymmetry in the wrong direction.

SongOf:
@he-yufeng confirmed on both.

The asymmetric `bytes_ahead` is intentional: each layer reads the bytes it can actually see ahead of the request. The drop decides before the request reaches a worker, so everything dispatched and not yet completed precedes it; the arbitration decides at posting time, when the worker-queue slices are its own contenders, so only the hardware backlog plus the slots already chosen precede a slot. The drop side therefore reads the wider set, and it is the irreversible decision — the conservatism runs the right way. I'll add a sentence to `deadline_mlu.h` saying so, so the asymmetry is not "fixed" in either direction.

`dispatching_bytes_`: your trace matches mine — `complete()` is the only `Dispatching → Terminal` exit and the only decrement, both sides gated on `degradation_eligible`. `Step3DropCountsBytesAlreadyDispatched` pins the effect but not the invariant; I'll add a unit test that checks `dispatching_bytes_` equals the eligible bytes currently in `Dispatching` after pick / complete / cancel-while-dispatching / drop / retireBatch. Both in the next push.

PR #3782, opened 2026-08-30
[Bugfix][Store] Allow explicit host identity for local-first allocation

### LINKED ISSUE #3762: [Bug] local_first / prefer_alloc_in_same_node never matches on Kubernetes: host_id is the client's own pod IP and cannot be overridden (v0.3.13)
### Bug Report

`-allocation_strategy=local_first` (and `ReplicateConfig.prefer_alloc_in_same_node`) can never match a co-located store segment when Mooncake runs on Kubernetes with a pod network. `host_id` is derived from the client's own `local_hostname`, which for a pod is its **pod IP**, and two pods on the same node have two different pod IPs. So the writer's host is never found in the index, the walk falls through to the deterministic `lower_bound` fallback, and the deployment pays the concentration cost of local-first while getting none of the locality benefit it is for.

Measured on an 8-segment cluster with 3 writers: **three segments held 100% of the data, five held 0 or a single 16 MiB unit, and none of the three was the segment co-located with its writer.** Predicted placement from the source matched observed placement 3/3.

This is the implementation of #2633 / #2638. The RFC's goal 1 is "a Mooncake-native notion of physical host identity" and goal 4 is "prefer segments on the writer's host"; neither holds under Kubernetes. The RFC even lists "a physical host may run multiple Mooncake clients and multiple store segments" as a reason not to make callers pass `Segment.name` — but the chosen host identity has exactly the same problem one layer down.

### Environment

- `docker.io/kvcacheai/mooncake:0.3.13`
- Kubernetes, CNI with per-pod IPs. Store runs as a DaemonSet (one store pod per node); the writers are vLLM pods on those same nodes.
- Store: 8 memory segments x ~1 TiB, `replica_num=1`, `-allocation_strategy=local_first`
- Writers use the default `ReplicateConfig` (`prefer_alloc_in_same_node` left at its default)

### Mechanism, with line numbers

1. `local_first` is not a strategy object at all. `CreateAllocationStrategy` maps `LOCAL_FIRST` to `RandomAllocationStrategy` (`allocation_strategy.h:655`). The locality lives entirely in the `preferred_segments` list built before the call.

2. That list is built from `GetHostOrderedSegments(writer_host_id, key)` (`master_service.cpp:4086`), where

```cpp
if ((allocation_strategy_type_ == AllocationStrategyType::LOCAL_FIRST ||
     config.prefer_alloc_in_same_node) && config.replica_num == 1) {
    writer_host_id = config.host_id.empty() ? GetClientHostId(client_id) : config.host_id;
}
```
(`master_service.cpp:4349-4354`)

3. `config.host_id` is not the caller's to set. `Client::AttachHostId` (`client_service.cpp:534`) **overwrites** it unconditionally with the client's own:

```cpp
ReplicateConfig Client::AttachHostId(const ReplicateConfig& config) const {
    ReplicateConfig client_cfg = config;
    if (!host_id_.empty()) client_cfg.host_id = host_id_;
    return client_cfg;
}
```

4. And `host_id_` is fixed at construction from the client's own bind address:

```cpp
host_id_(ResolveMooncakeHostId(local_hostname)),
```
(`client_service.cpp:416`)

`ResolveMooncakeHostId` is "hostname minus port" (`utils.cpp:601`) — that is, the address the transfer engine binds. In a pod that is the pod IP.

5. The store side is symmetric: the store pod mounts with `segment.host_id = host_id_` (`client_service.cpp:3534`), which the master records as the segment's host (`master_service.cpp:929`). So the index is keyed by **store pod IP**, and the lookup key is the **writer pod IP**. On the same physical node those are two different addresses and the lookup always misses.

6. On a miss, `BuildHostOrderedSegments` (`segment.cpp:54-68`) does:

```cpp
auto start_it = segments_by_host.find(writer_host_id);
if (start_it == segments_by_host.end()) {
    start_it = segments_by_host.lower_bound(writer_host_id);
    if (start_it == segments_by_host.end()) start_it = segments_by_host.begin();
}
```

so each writer is deterministically pinned to the lexicographically next store pod IP and stays there until that segment fills, then rings on to the next. With one segment per host, "rotate the starting segment by key hash" (RFC goal 6) has nothing to rotate over.

We predicted each writer's landing segment from its pod IP and the sorted store pod IP list, and it matched the observed per-segment `segment_allocated_bytes` for all three writers.

### Why this matters beyond imbalance

The concentration itself is by design — RFC goal 3 is to reduce blast radius by keeping one writer's cache together, and we understand that. The problem is that the *benefit* is unreachable here: writes never stay local, so every put crosses the network, and we still pay the concentration. On our cluster that meant 5 of 8 segments sat idle while 3 filled to the allocator ceiling.

There is also no signal. Nothing logs "the writer's host was not found, falling back" — the only hint is a `VLOG(1)` reporting `local_first_preferred_segments=<count>`, which is non-zero in exactly this failure case because the fallback still produces an ordered list. From the outside, a working local-first and a permanently-missing one look identical.

We have switched to `free_ratio_first`, which fixed the balance (segment skew max/mean 1.50 -> 1.00, all segments carrying data). So this is not blocking us — but the feature is unusable on Kubernetes as written, and we suspect anyone who turns it on there is silently getting the fallback.

### What we would like

Any one of these would make the feature usable:

1. **Let the deployment supply `host_id`.** An env var (say `MC_HOST_ID`) or a client config field that is *not* overwritten by `AttachHostId`, so a Kubernetes deployment can pass `spec.nodeName` or the host IP via the downward API to both the store pods and the writers. This is the smallest change and needs no topology awareness inside Mooncake.
2. **Default to a real host identity** rather than the bind address — e.g. the kernel hostname, or the node IP — with the bind address as fallback.
3. Failing both, **log a warning once per client** when `local_first`/`prefer_alloc_in_same_node` is active and the writer's `host_id` is absent from the segment index, and **document** that `host_id` is the client's bind address so container deployments know what they are enabling.

Happy to test a patch on our cluster; we have the 3-writer / 8-segment reproduction standing.

zupengwang:
## Description

Fixes #3762.

Kubernetes deployments commonly use each pod IP as `local_hostname`. A writer pod and a colocated store pod therefore report different derived host IDs, so `local_first` / `prefer_alloc_in_same_node` never recognizes them as colocated.

This change:

- lets the C++ client use `MOONCAKE_HOST_ID` as an explicit deployment-level host identity;
- preserves the existing `local_hostname`-derived fallback when the variable is unset, empty, or whitespace-only;
- normalizes endpoint-shaped overrides before rejecting loopback and wildcard identities, matching the existing derived-host safety behavior; and
- documents a Kubernetes Downward API setup that keeps `status.podIP` as the routable transfer endpoint while using `spec.nodeName` as the shared placement identity.

The change does not modify the RPC/wire format, allocation algorithm, or transfer transports.

## Module

- [ ] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [x] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Exact-base regression:**

On `main@73471dc2803f0de7e42484cfa1e2f44c09dc1de5` with the original PR implementation applied, `ResolveMooncakeHostIdNormalizesEndpointOverride` failed because the explicit override `kubernetes-node-a:5000` was returned unchanged. With the review fix applied, the same test passes after normalizing the endpoint to `kubernetes-node-a`.

**Test commands:**

```bash
cmake --build build-store --target master_service_test -j6

./build-store/mooncake-store/tests/master_service_test \
  --gtest_filter='MasterServiceTest.ResolveMooncakeHostId*' \
  --gtest_repeat=20 --gtest_break_on_failure

./build-store/mooncake-store/tests/master_service_test \
  --gtest_break_on_failure

clang-format --dry-run --Werror \
  mooncake-store/src/utils.cpp \
  mooncake-store/tests/master_service_test.cpp

pre-commit run --files \
  mooncake-store/AGENTS.md \
  mooncake-store/src/utils.cpp \
  mooncake-store/tests/master_service_test.cpp \
  docs/source/deployment/mooncake-store-deployment-guide.md \
  docs/source/design/store/mooncake-store.md

cd docs && make html
git diff --check
```

**Test results:**

- [x] Unit tests pass on `main@73471dc2803f0de7e42484cfa1e2f44c09dc1de5`: 5 focused tests × 20 repetitions; full `master_service_test` 114/114.
- [x] Sphinx documentation build succeeds for all 96 pages on the original PR base.

The code path is transport-independent. No RDMA or GPU performance claim is made, and a live Kubernetes deployment was not available in this validation environment.

## Checklist

- [x] I have formatted my code using `clang-format` 20
- [x] For changes >500 LOC: I have filed an RFC issue (not applicable; this change is under 500 LOC, and the original local-first RFC is #2633)

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex was used to audit the relevant issue/PR history, implement the focused code, tests, and documentation changes, and run validation. The final diff and evidence were reviewed end-to-end before publication.


yokinoshitayoki:
This direction looks promising to me. In PR #2638, I thought adding another env varient is inconvient for users so I chose the simpler approach. That said, I do agree that supporting an explicit host identity while preserving the existing behavior when the variable is unset, empty, or whitespace-only strikes a better balance.

mooncake-store/src/utils.cpp
```diff
@@ -598,22 +598,27 @@ int64_t time_gen() {
...
+           !AsciiCaseInsensitiveEquals(host_id, "localhost") &&
+           host_id != "127.0.0.1" && host_id != "0.0.0.0" && host_id != "::1" &&
+           host_id != "[::1]" && host_id != "::" && host_id != "[::]";
+}
+
 std::string ResolveMooncakeHostId(const std::string &local_hostname) {
+    const std::string configured_host_id(
+        TrimAsciiWhitespace(Environ::GetString("MOONCAKE_HOST_ID", "")));
+    if (!configured_host_id.empty()) {
+        return IsUsableMooncakeHostId(configured_host_id) ? configured_host_id
```
yokinoshitayoki: Normalize or reject endpoint-shaped overrides before validation. `localhost:5000`, `127.0.0.1:5000`, and `[::]:5000` are currently accepted despite the documented loopback/wildcard rejection, which can give unrelated pods the same invalid placement identity. Strip/reject ports first and add regression cases.

yokinoshitayoki:
Could we also update this on AGENTS.md?

mooncake-store/src/utils.cpp
zupengwang: Fixed in `ce37f6fc`. Non-empty overrides now go through the existing endpoint normalization before loopback/wildcard validation. I added regression coverage for `localhost:5000`, `127.0.0.1:5000`, `0.0.0.0:5000`, `[::1]:5000`, and `[::]:5000`, plus a valid endpoint normalization case. On `main@73471dc`, the focused resolver suite passed 5 tests × 20 repetitions and the full `master_service_test` passed 114/114.

zupengwang:
Thanks for the feedback. Updated in `ce37f6fc`: unset, explicitly empty, and whitespace-only `MOONCAKE_HOST_ID` values continue to fall back to the existing `local_hostname` behavior, with separate regression coverage for the set-but-empty case. I also added the requested local-first host identity guidance to `mooncake-store/AGENTS.md`, including the placement-vs-transfer identity boundary and endpoint normalization rule. Validation on `main@73471dc`: 5 focused resolver tests × 20 repetitions and 114/114 full `master_service_test`.

yokinoshitayoki (approved):
LGTM.

ykwd (approved):

PR #3783, opened 2026-08-31
[P2P-Store] Unregister buffers on GetReplica failure paths

chlins:
## Description

The replica-fetch path registers caller buffers with the transfer engine but never unregisters them on failure:

- In `doGetReplica`, if `memory.Add` fails partway through the buffer list, the buffers registered in earlier iterations stay registered. A transfer failure after all `Add` calls leaks every registration.
- In `GetReplica`, once `doGetReplica` has registered all buffers, a failure in the metadata recheck or in `updatePayloadMetadata` returns without any cleanup. `catalog.Add` only runs on full success, so a later `DeleteReplica` cannot release them either (`ErrPayloadNotOpened`).

`Register` already handles its own failure paths with `unregisterBuffers`; this PR applies the same pattern to the replica path: `doGetReplica` tracks what it registered and rolls back on its own failures, and `GetReplica` rolls back on post-transfer failures via a small `buffersFromLists` helper.

Note: this covers the error paths only. The pre-existing behavior where a repeated metadata recheck loop can bump registration refcounts more than once on success is unchanged and can be looked at separately if there is interest.

## Related Components

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

**Test commands:**
```bash
go build ./...
gofmt -l .
go vet ./...
```

**Test results:**

`go build` and `gofmt` clean; `go vet` shows only the three pre-existing `unsafe.Pointer` warnings in `transfer_engine.go` (identical on main). The package has no unit tests; the change adds rollback on error paths only, and the success path is untouched.

## Checklist

## AI Assistance Disclosure

AI assisted with the code audit and drafting. The submitter has reviewed, understands, and takes responsibility for all changes.

### DIFF
--- a/mooncake-p2p-store/src/p2pstore/core.go
+++ b/mooncake-p2p-store/src/p2pstore/core.go
@@ -267,6 +267,9 @@ func (store *P2PStore) doGetReplica(ctx context.Context, payload *Payload, addrL
 		addr, size := addrList[i], sizeList[i]
 		err := store.memory.Add(addr, size, maxShardSize, "cpu:0")
 		if err != nil {
+			// Wait for transfers already started in earlier iterations:
+			// unregistering memory that in-flight transfers still use is unsafe.
+			wg.Wait()
 			store.unregisterBuffers(registered, maxShardSize)
 			return err
 		}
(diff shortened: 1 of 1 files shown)


catyans (approved):
LGTM. The current head addresses both Copilot findings: partial-Add rollback waits for already-started transfers, and failure cleanup unregisters every successful recheck round before catalog ownership exists. CI is green.

suggestion (non-blocking): please track the acknowledged multi-round success refcount leak in a follow-up; a successful GetReplica still records only one catalog cleanup for multiple registration rounds.

PR #3784, opened 2026-08-31
[Bugfix][TENT] TENT store warmup failure on multi-NUMA RDMA hosts

Colors-111:
## Description

On machines with multiple RDMA NICs spanning different NUMA nodes, starting a mooncake store in TENT mode (`MC_USE_TENT=1`) and connecting sglang via `--hicache-storage-backend mooncake` consistently fails during `MooncakeStore.warmup()`:

```
transfer_engine_impl.cpp:1652] Unable to find registered buffer for request: opcode 1 source 0x... target_id 1 target_offset 0x... length 4096 transport_hint unspec
transfer_task.cpp:800] Batch ... completed with task failures: task_ids=[0]AssertionError  # self.store.put(warmup_key, warmup_value) == 0
```

The same configuration works under the classic Transfer Engine (classic TE).

## Root Cause

When the store detects multiple NICs spanning NUMA nodes, it allocates NUMA-segmented memory for the segment and produces a location string of the form `"segments:4096:0,1"` via `buildSegmentsLocation()` (`mooncake-transfer-engine/src/memory_location.cpp`), passing it into TENT through `registerLocalMemory`.

TENT's `registerLocalMemory`, when `options.location` is not the wildcard, uses it to override the `"cpu:N"` probed by `getLocation()`, writes it into `BufferDesc.location`, and publishes it to metadata (`transfer_engine_impl.cpp`).

However, when the peer TENT resolves that buffer's location in `getTransportType()` → `TransportSelector::select()` → `isTransportAvailable()`:

- `getTypeEnum("segments")` returns `MTYPE_UNKNOWN`  (`transfer_engine_impl.cpp`, which only recognizes `"cpu"`/`"*"`)
- `memTypeFromLocation("segments:...")` likewise returns `MEM_UNKNOWN`  (`topology.h`)

As a result `remote_memory_type = MTYPE_UNKNOWN`, which matches none of the CPU↔CPU capability branches in `isTransportAvailable`, so `select()` returns `UNSPEC` and the transfer fails.

## Trigger Conditions and Reproduction

Multiple RDMA NICs distributed across different NUMA nodes (`seg_numa_nodes` non-empty) + TENT mode. On single-NIC or single-NUMA machines `seg_numa_nodes` is empty, location falls back to `kWildcardLocation` (`"*"`), and the issue does not trigger — which is why it is rarely seen in common single-node setups.

Exact trigger conditions (all three required):

1. **TENT mode**: `MC_USE_TENT=1`
2. **RDMA protocol**: `--protocol=rdma` (or sglang `hicache-storage-backend` configured with `protocol: rdma`)
3. **≥2 RDMA NICs on different NUMA nodes**: e.g. `mlx5_0` on NUMA 0, `mlx5_1` on NUMA 1

**Minimal reproduction**: two machines meeting the hardware conditions above. Start a TENT store on one (`mooncake_client` + `MC_USE_TENT=1`), and use a TENT client on the other (sglang `--hicache-storage-backend mooncake`) to write to it. The store's segment desc contains a 100GB buffer with location `"segments:4096:0,1"`; the client-side `select()` returns `UNSPEC` and `put` fails.

## Fix

The root cause is the location override in `registerLocalMemory`: any non-wildcard caller location replaces the probe with no validation, and `"segments:"` is merely the illegal value that happened to hit this path. The fix validates the caller location before overriding, keeping the probe as the source of truth for transport selection:

```cpp
if (options.location != kWildcardLocation && !options.location.empty()) {
    auto probed_type = getTypeEnum(LocationParser(desc.location).type());
    auto caller_type = getTypeEnum(LocationParser(options.location).type());
    if (caller_type == MTYPE_UNKNOWN) {
        // unknown type (e.g. "segments:...") → ignore caller, keep probe + WARN
    } else if (caller_type != probed_type) {
        // type conflict (e.g. cpu vs cuda) → ignore caller, keep probe + WARN
    } else {
        desc.location = options.location;   // same-type refinement (cpu:0 → cpu:1)
    }
}
```

This avoids polluting TENT's type system with a classic TE `"segments"` dialect and blocks any future illegal location, not just `"segments"`. The cross-NUMA layout is not lost — it is already recorded in `BufferDesc.regions` (`coalesceRegions(entries)`), which the RDMA worker uses for affinity; that is orthogonal to the `location` field (used only
for transport-type selection).

## Tests

- `TransferEngineConfigOverrideTest.RegisterLocalMemoryIgnoresIncompatibleCallerLocation`:
  starts a TCP-only engine, registers a host buffer with
  `options.location = "segments:4096:0,1"`, and asserts via `getSegmentInfo`
  that `buffer.location` keeps the probed `"cpu:..."` form — the caller's
  `"segments:"` does not leak in.

## Files Changed

- `mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp` — validate caller location before overriding in `registerLocalMemory` 
- `mooncake-transfer-engine/tent/tests/transfer_engine_config_override_test.cpp` — adds `RegisterLocalMemoryIgnoresIncompatibleCallerLocation`

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
```

**Test results:**
- [X] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used


staryxchen (commented):
Thanks for the detailed analysis. The warmup failure chain is real: on multi-NUMA, multi-NIC hosts the store writes "segments:..." into registerLocalMemory, TENT uses it to overwrite the probed "cpu:N", the peer’s getTypeEnum("segments") returns MTYPE_UNKNOWN, and selection becomes UNSPEC. Classic TE survives because it never uses this string to pick a transport; it only decodes it to "cpu:N" for HCA binding inside RDMA.

I would not treat "segments" as MTYPE_CPU / MEM_HOST. That string is classic TE’s NUMA-segment encoding, not a TENT memory type. TENT selection should only understand types like cpu / cuda / *. Teaching getTypeEnum a TE dialect means the next illegal location will fail the same way.

The deeper issue is the override itself:
```
  if (options.location != kWildcardLocation)
      desc.location = options.location;
```

  Any non-"*" caller string replaces the probe with no validation. "segments:" is just the illegal value that happened to hit this path. TENT cannot assume the location from upper layers is always valid.

registerLocalMemory observes existing memory; the probe is the source of truth. allocateLocalMemory is the API that should trust location (it is an allocation target). TENT already records cross-NUMA layout in regions, and the RDMA worker uses that for affinity. What gets overwritten is the location field used for transport selection.

Suggestion: validate the caller location before overriding.

  1. "*" / empty: keep the probe.
  2. Caller type unknown (segments:, garbage) or conflicts with the probe (cpu vs cuda): ignore the caller, log a WARNING, keep the probe.
  3. Same-type refinement only (e.g. probe "cpu:0", caller "cpu:1").

Then the store does not need a TENT-specific location path, and TENT’s type system does not need to learn a TE dialect. The MOONCAKE_LOCAL_HOSTNAME mapping is a separate change and should land on its own PR.

Colors-111:
> Thanks for the detailed analysis. The warmup failure chain is real: on multi-NUMA, multi-NIC hosts the store writes "segments:..." into registerLocalMemory, TENT uses it to overwrite the probed "cpu:N", the peer’s getTypeEnum("segments") returns MTYPE_UNKNOWN, and selection becomes UNSPEC. Classic TE survives because it never uses this string to pick a transport; it only decodes it to "cpu:N" for HCA binding inside RDMA.
> 
> I would not treat "segments" as MTYPE_CPU / MEM_HOST. That string is classic TE’s NUMA-segment encoding, not a TENT memory type. TENT selection should only understand types like cpu / cuda / *. Teaching getTypeEnum a TE dialect means the next illegal location will fail the same way.
> 
> The deeper issue is the override itself:
> 
> ```
>   if (options.location != kWildcardLocation)
>       desc.location = options.location;
> ```
> 
> Any non-"*" caller string replaces the probe with no validation. "segments:" is just the illegal value that happened to hit this path. TENT cannot assume the location from upper layers is always valid.
> 
> registerLocalMemory observes existing memory; the probe is the source of truth. allocateLocalMemory is the API that should trust location (it is an allocation target). TENT already records cross-NUMA layout in regions, and the RDMA worker uses that for affinity. What gets overwritten is the location field used for transport selection.
> 
> Suggestion: validate the caller location before overriding.
> 
> 1. "*" / empty: keep the probe.
> 2. Caller type unknown (segments:, garbage) or conflicts with the probe (cpu vs cuda): ignore the caller, log a WARNING, keep the probe.
> 3. Same-type refinement only (e.g. probe "cpu:0", caller "cpu:1").
> 
> Then the store does not need a TENT-specific location path, and TENT’s type system does not need to learn a TE dialect. The MOONCAKE_LOCAL_HOSTNAME mapping is a separate change and should land on its own PR.

Thanks for the detailed review! I've updated the PR to follow your recommended approach — validating the caller-supplied location before overriding the probe in `registerLocalMemory`, instead of teaching `getTypeEnum` the "segments" dialect. The `MOONCAKE_LOCAL_HOSTNAME` mapping has also been split out to a separate PR. Please take another look when you have a chance.

catyans (approved):
Reviewed `730f6124b5`. The revised fix addresses the failure at the correct boundary: `registerLocalMemory()` keeps the probed memory type as the source of truth, ignores unknown caller encodings such as `segments:...` and known-type conflicts, and still permits same-type location refinement. The independently recorded `regions` remain available for RDMA affinity, so rejecting the classic-TE location dialect does not discard the multi-NUMA routing information. The unrelated `MOONCAKE_LOCAL_HOSTNAME` change has also been removed from this PR.

The regression test exercises the original metadata corruption path and verifies that `segments:...` cannot leak into the published buffer location. I found no blocking issue and approve this direction.

Non-blocking follow-up: add small cases for the other two validation branches—a known conflicting type is ignored, while a valid same-type refinement is accepted. The current test directly covers only the unknown-type branch. Please also let the newly triggered CI complete before merge.

staryxchen:
@Colors-111 There are some fail job, plz check it out.

staryxchen (commented):
The unknown-type path (segments:...) is covered. Please add two small cases on the same fixture:

1. Conflicting known type: register host memory with cuda:0 and assert the published location stays cpu:....
2. 2. Same-type refinement: register with cpu:1 (different from the probe) and assert the published location is the caller value.

That exercises the mismatch-ignore and same-type-accept branches; "*" / empty can stay untested.

catyans (approved):
LGTM. Re-reviewed the current head; the follow-up only removes the duplicated `/metadata` suffix from the test endpoint, with no production-code change. The TENT matrices pass.

The remaining PG elastic recovery failure is unrelated to this change; please rerun it before merge.

alogfans (approved):
LGTM

PR #3787, opened 2026-08-31
[Store] Increase offset allocator size-class precision

982945902:
## Description

Increase `OffsetAllocator`'s `SmallFloat` mantissa precision from 3 bits to 4 bits, doubling the number of leaf size classes from 256 to 512.

This reduces size-class rounding waste for KV workloads while preserving compatibility with allocator state serialized by the previous 3-bit layout. Legacy raw and MsgPack snapshots are detected during deserialization and their free-bin lists are rebuilt into the new layout.

For the DSA pair used by the allocation benchmark:
- KV `3,274,752 B`: normalized size changes from `3,407,872 B` to `3,276,800 B`
- Indexer `658,432 B`: normalized size changes from `720,896 B` to `688,128 B`

## Module

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [x] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Unit tests:**
```bash
/build/release/mooncake-store/tests/offset_allocator_test
/build/release/mooncake-store/tests/serializer_test
```

Results:
- OffsetAllocator: 54/54 passed
- Serializer: 7/7 passed
- Storage allocator persistence subset: 9/9 passed
- HA snapshot tests: 69/69 passed
- Legacy raw 3-bit layout migration test passed
- Legacy MsgPack 256-bin layout migration test passed
- Cross-version manual test passed: an unmodified `origin/main` 3-bit binary wrote raw and MsgPack allocator snapshots, and the 4-bit binary loaded both snapshots and successfully continued allocating

**Complete DSA allocation benchmark:**
```bash
taskset -c 2 allocation_strategy_bench \
  --workload=dsa \
  --dsa_segment_capacity=16384 \
  --dsa_warmup_multiplier=1 \
  --dsa_capacity_multiplier=3 \
  --dsa_max_allocs=2000000
```

Method:
- Same `origin/main` baseline and build environment
- Full 96-case matrix per run: Random/FreeRatioFirst, KV-only/paired KV+Indexer, homogeneous/skewed, replicas 1/2/3, segments 1/2/4/8/16
- Baseline and change run in alternating order on the same pinned CPU
- Two complete runs per version

Aggregate result across both complete runs:

| Workload | 3-bit utilization | 4-bit utilization | Change | Throughput change |
|---|---:|---:|---:|---:|
| KV-only | 96.0833% | 99.9242% | +3.8409 pp | +0.24% |
| KV + Indexer | 95.0501% | 98.9827% | +3.9326 pp | +0.10% |

The throughput deltas are effectively neutral; the reproducible gain is reduced size-class rounding and higher achievable utilization.

- [x] Unit tests pass
- [x] Manual testing done (complete benchmark described above)

## Checklist

- [x] I have formatted my code using the repository's clang-format configuration
- [ ] I have run pre-commit on the files changed in this PR and all hooks pass (the `pre-commit` executable is unavailable in the current environment)
- [x] I have updated the documentation (not applicable; no user-facing interface changes)
- [x] For changes >500 LOC: I have filed an RFC issue (not applicable; 266 insertions / 90 deletions)

## AI Assistance Disclosure

- [ ] No AI tools were used
- [x] AI tools were used (Codex assisted with implementation, tests, benchmark execution, and PR preparation; the submitter is responsible for reviewing and understanding the changes.)


982945902:
Added explicit legacy MsgPack migration coverage in b9dfa531. I also ran a real cross-version check: an unmodified origin/main (3-bit) binary generated both raw and MsgPack allocator snapshots; the PR build loaded both artifacts, preserved the allocation/free-space metrics, and successfully continued allocating. The full OffsetAllocator suite now passes 54/54.

ykwd:
Thanks for working on this. Increasing the size-class precision looks useful, and I am supportive of this direction.

One thing that may be worth considering is whether we should go directly to `MANTISSA_BITS = 5` instead of stopping at 4. I shared this PR with some other users, and one of them mentioned that they had encountered a similar issue when using Mooncake for RL data-plane offloading. Compared with KV cache workloads, where allocation sizes are relatively stable, RL workloads can have much more diverse allocation sizes. In their case, increasing `MANTISSA_BITS` to 5 was the smallest change that provided the most noticeable improvement in terms of fragmentation.

That said, I think the more important point is that the workload characteristics here may be quite different from the KV/DSA workloads currently covered by the benchmark. It might therefore be useful to first profile representative RL allocation traces and add an RL-like workload to `allocation_strategy_bench` (e.g., with a wider or more skewed allocation-size distribution). We could then quantitatively compare 3/4/5-bit precision with different workload and segment sizes, and also verify whether increasing from 3 to 5 bits introduces any hot-path overhead.

Of course, the above questions go beyond the scope of this PR. We will proceed with reviewing and testing this PR as it stands, while also welcome anyone interested in exploring the questions above.

982945902:
Thanks for suggesting the 5-bit comparison. I ran a local 4-bit vs. 5-bit experiment without pushing any 5-bit code.

### Method

- Same Release build environment and benchmark binary source
- Complete 96-case DSA matrix per run:
  - Random and FreeRatioFirst
  - KV-only and paired KV+Indexer
  - homogeneous and skewed capacities
  - replicas 1/2/3
  - segments 1/2/4/8/16
- 16 GiB per base Segment
- 1x-capacity warmup + 3x-capacity measurement
- 5% fail-triggered eviction, maximum 2M allocations per case
- Both binaries pinned to the same CPU
- Four complete runs per version, in alternating order

### Results

| Workload | Metric | 4-bit | 5-bit | 5-bit vs. 4-bit |
|---|---:|---:|---:|---:|
| KV-only | Achievable utilization | 99.924163% | 99.924163% | +0.0000 pp |
| KV-only | Allocation throughput | 972,846/s | 968,511/s | -0.45% |
| KV-only | P99 | 2,725 ns | 2,754 ns | +1.04% |
| Paired KV+Indexer | Achievable utilization | 98.987256% | 99.387264% | +0.4000 pp |
| Paired KV+Indexer | Allocation throughput | 892,222/s | 878,943/s | -1.49% |
| Paired KV+Indexer | P99 | 2,793 ns | 2,974 ns | +6.49% |

The utilization result matches the size-class rounding:

- KV `3,274,752 B`: both 4-bit and 5-bit normalize to `3,276,800 B`, so 5-bit provides no additional KV-only capacity benefit.
- Indexer `658,432 B`: 4-bit normalizes to `688,128 B`; 5-bit normalizes to `671,744 B`, which explains the paired-workload improvement.

The latency percentiles were noisier than utilization. For paired mode, Random's aggregate P99 increased while FreeRatioFirst's decreased, so I would not treat the aggregate P99 delta as a stable regression without longer allocator-only profiling. However, the current data does not show a hot-path advantage for 5-bit.

### Implementation and metadata cost

Moving from 4 to 5 bits is small mechanically, but it is not only a constant change:

- leaf bins increase from 512 to 1,024;
- each top-bin bitmap must widen from `uint16_t` to `uint32_t`;
- `m_binIndices` grows from 2,048 B to 4,096 B;
- leaf bitmaps grow from 64 B to 128 B;
- total bin-index metadata grows from 2,112 B to 4,224 B per OffsetAllocator (+2,112 B), and the serialized bin-index payload grows correspondingly.

The lookup remains constant-time and uses native 32-bit bit operations, but the lookup-table footprint doubles again.

For the current KV/DSA workload, my interpretation is that 4-bit captures most of the available gain: 5-bit adds about 0.40 pp only for the paired case, provides no KV-only gain, and doubles bin metadata again. I therefore plan to keep this PR at 4 bits.

I agree that the decision may be different for RL offloading with a much wider allocation-size distribution. A representative RL trace or RL-like benchmark would be the right way to evaluate 3/4/5 bits separately, rather than extrapolating from the KV/DSA sizes used here.

982945902:
Rebased onto the latest `main` (`ebe38646`) and resolved the conflict with the new `ImportOffsetBufferAllocator` / `LiveAllocation` interface. The precision-independent normalization-boundary test now uses the new import API while retaining its original intent.

Local Release validation after the rebase:
- `offset_allocator_test`: 54/54 passed
- `buffer_allocator_test`: 19/19 passed
- `serializer_test`: 7/7 passed

GitHub now reports the PR as mergeable; the new CI run is queued.

catyans (approved):
LGTM. The 4-bit bin layout is applied consistently across allocation, free, and serialization, while legacy raw and MsgPack states rebuild their free-bin chains under the new mapping. Migration tests, benchmarks, and CI are green.

mooncake-store/include/offset_allocator/offset_allocator.h
```diff
@@ -382,49 +388,61 @@ void __Allocator::serialize_to(T& serializer) const {
...
+        const size_t node_data_size = static_cast<size_t>(m_current_capacity) *
+                                      (sizeof(Node) + sizeof(NodeIndex));
+        const size_t common_tail_size = sizeof(m_freeOffset) + node_data_size;
+        const size_t legacy_tail_size =
+            sizeof(m_usedBinsTop) + NUM_TOP_BINS * sizeof(uint8) +
+            LEGACY_NUM_LEAF_BINS * sizeof(NodeIndex) + common_tail_size;
+        const size_t current_tail_size =
+            sizeof(m_usedBinsTop) + sizeof(m_usedBins) + sizeof(m_binIndices) +
+            common_tail_size;
+        const size_t remaining_size = serializer.remaining_size();
```
yokinoshitayoki, line 434: The layouts differ by 1,056 bytes, exactly 33 serialized nodes. Corrupting a legacy snapshot's node count from 1,501 to 1,468 makes this accept it as the new layout; allocate(64) then segfaults, while the baseline rejects it.

982945902: Thanks, you are right. The raw deserializer was inferring the 3-bit versus 4-bit layout solely from the remaining byte count, so reducing the serialized node count by 33 could exactly hide the 1,056-byte layout difference and make a corrupted legacy snapshot enter the new-layout path. I updated the raw format in e6d7c951 to prefix new snapshots with an explicit magic and version. Snapshots without the magic are now parsed only as the legacy 3-bit layout, with its exact expected payload size; there is no longer a length-based choice between the two layouts. I also added the 1,501 -> 1,468 regression case, which is now rejected during deserialization before allocate() can be called. The full offset_allocator_test suite passes 55/55 and pre-commit passes.

yokinoshitayoki:
@982945902 Could we have a look at the failed checks?

982945902:
Fixed the failed `Check code format` job in dca44f03 by applying clang-format 20 to the two reported locations. I also reran the PR-scoped pre-commit checks and `./scripts/code_format.sh --check --changed-lines --base origin/main`; both pass locally.

yokinoshitayoki:
@982945902 I think RFC #3965 investigates allocation failures and segment placement in RL workloads. The RL workload seems useful for revisiting the 4-bit vs. 5-bit trade-off. Could you take a look?

982945902:
Thanks for pointing me to RFC #3965. I reviewed its analysis, and it clarifies that the RL allocation failures are primarily caused by segment-placement fragmentation rather than size-class precision.

Since reproducing the full RFC experiment requires its production traces and supporting tooling, I ran a smaller allocator-only synthetic sanity check based on the workload shape described in the RFC:

- 16 × 4 GiB segments and 8 simulated DP ranks;
- 12–16 mixed-size objects plus one 640–1920 MiB object per rank and step;
- objects released after two steps, with no eviction;
- 12 fixed seeds and 172,483 allocation attempts for each configuration.

| Placement | 4-bit large-object failures | 5-bit large-object failures |
|---|---:|---:|
| Random | 160 (1.3889%) | 161 (1.3976%) |
| FreeRatioFirst | 521 (4.5226%) | 509 (4.4184%) |
| Best-fit model | 0 | 0 |

Moving from 4 to 5 bits reduced measured size-class rounding overhead from approximately 0.65% to 0.26%, but did not materially reduce large-object allocation failures. At failure time, the cluster still had approximately 17.5–18.7 GiB of total free space on average, while the largest contiguous region was only approximately 1.5 GiB.

This synthetic test is not intended to reproduce the RFC’s production traces, but its result is consistent with the RFC’s conclusion: higher size-class precision reduces rounding waste, while placement policy determines whether large contiguous regions survive.

yokinoshitayoki:
This direction looks pretty good to me, I'll test this PR locally later.

catyans (approved):
LGTM. Re-reviewed the current head after the snapshot-format update. The explicit magic/version removes the ambiguous legacy/current layout inference, while legacy raw and MsgPack snapshots still take the compatibility path and the new regression covers the previously ambiguous node count. CI is green.

Also, thanks @yokinoshitayoki for connecting this with #3965, and @982945902 for running the extra workload-shaped sanity check. The result makes the boundary here much clearer: this PR improves allocator rounding precision; preserving large contiguous regions remains a placement-policy problem.

yokinoshitayoki:
@982945902 I compared the 3-bit baseline (ebe38646) with this PR (dca44f03) on a local A100 host with AMD EPYC 7742 CPUs. Both used identical Release build settings (GCC 11.4, -O3, metrics enabled). These benchmarks exercise CPU allocation paths, excluding GPU transfers and RPC. Each version ran four times in alternating order with fixed CPU affinity.
The complete 96-case DSA matrix reproduced the expected capacity improvement:
```
taskset -c 2 ./allocation_strategy_bench \
  --workload=dsa --dsa_segment_capacity=16384 \
  --dsa_warmup_multiplier=1 --dsa_capacity_multiplier=3 \
  --dsa_max_allocs=2000000
```
Peak utilization before eviction increased from 96.083% to 99.924% for KV-only and 95.047% to 98.987% for KV+Indexer. Aggregate throughput was essentially unchanged (−0.05% for both workloads).
However, we observed a repeatable slowdown with eight threads allocating and freeing against the same segment through the Put allocation layer. An isolated follow-up used:
```
taskset -c 2-9 ./offset_allocator_concurrency_bench \
  --threads=8 --iterations=500000 --warmup_iterations=20000 \
  --success_request_kb=3198 --layer=put_allocation --scenario=success
```
Across four paired runs, throughput decreased 6.3% geometrically, while P99 increased 8.5%. One run had a larger slowdown; the median paired throughput change was −3.0%. Repeating with --success_request_kb=4 showed −3.1% throughput and +8.3% P99. Both request sizes showed lower throughput in every pair, with zero unexpected allocation results.
The fixed metadata cost also matched expectations: +1,056 bytes per internal allocator and +1,068 bytes per Raw snapshot.
Could you reproduce the contention case and investigate whether this slowdown is an expected trade-off? The DSA results look good, but this additional workload seems worth including in the performance evaluation.

982945902:
Thanks for reporting this and for providing the exact command. I tested the same baseline (`ebe38646`) and PR head (`dca44f03`) using the command from your comment:

```bash
taskset -c 2-9 ./offset_allocator_concurrency_bench \
  --threads=8 --iterations=500000 --warmup_iterations=20000 \
  --success_request_kb=3198 \
  --layer=put_allocation --scenario=success
```

An earlier four-pair run showed a slowdown in the same direction as your result. However, the result was not stable in subsequent runs, so I ran another 20 paired comparisons (40 executions), randomizing whether the baseline or PR version ran first within each pair.

The per-pair PR-vs-baseline throughput changes were:

| Pair | Delta | Pair | Delta |
|---:|---:|---:|---:|
| 1 | +3.58% | 11 | -5.34% |
| 2 | -1.07% | 12 | +0.54% |
| 3 | +7.91% | 13 | +2.85% |
| 4 | -1.51% | 14 | +1.93% |
| 5 | -2.84% | 15 | +5.17% |
| 6 | +0.53% | 16 | -4.26% |
| 7 | +0.14% | 17 | -1.93% |
| 8 | -5.28% | 18 | -4.83% |
| 9 | +2.06% | 19 | +0.34% |
| 10 | +2.78% | 20 | -0.12% |

Across all 20 pairs:

- geometric-mean throughput change: **-0.03%**
- median paired throughput change: **+0.24%**
- PR faster in **11/20** pairs and slower in **9/20**
- approximate 95% interval for the throughput change: **[-1.57%, +1.54%]**
- when the PR ran first: **-1.04%**
- when the PR ran second: **+1.00%**

P99 was also unstable. It increased by 5.34% across all pairs, but this was concentrated in the initial pairs; across pairs 5–20, the geometric-mean P99 change was only +0.08%.

I also compared the versions with `perf stat` and `perf record`. I did not find a stable increase in hot-path work:

- instructions: **-0.03%**
- cycles: **+0.33%**
- combined allocator hotspot share: **13.59% → 13.64%**
- `pthread_mutex_lock`: **7.04% → 7.00%**
- `SmallFloat::*`: **0.76% → 0.83%**
- `__Allocator::allocate`: **1.17% → 1.15%**

Unfortunately, my current host is heavily shared, with a sustained load average around 275–305 and substantial scheduler activity. I also attempted a stricter CPU/cpuset isolation run, but persistent kernel-side activity still caused millions of context switches, so I discarded those measurements as unreliable. I do not currently have access to a sufficiently quiet host.

My current conclusion is therefore that I can observe occasional regressions, but cannot reproduce a stable slowdown or identify a new allocator/locking bottleneck caused by the 4-bit layout. This does not invalidate the consistent regression you observed on the EPYC 7742 host.

Would it be possible to repeat this benchmark on one or two additional hosts, or share the per-run output and `perf stat` counters from the EPYC host? That would help determine whether the effect is architecture-specific or caused by the benchmark environment.

yokinoshitayoki (approved):
@982945902 Thanks for the additional measurements. I repeated the contention benchmark on the same EPYC 7742 host, reusing the exact baseline (<code dir="ltr">ebe38646</code>) and PR (<code dir="ltr">dca44f03</code>) binaries, with their SHA-256 hashes verified against the previous run.</p><p>The benchmark parameters remained unchanged: CPU 2–9, 8 threads, 500,000 iterations per thread, 20,000 warmup iterations, and <code dir="ltr">--layer=put_allocation --scenario=success</code>.</p><p>This time I ran <strong>12 paired comparisons per request size, totaling 48 executions</strong>. The execution order was randomized and balanced: the PR ran first in six pairs and second in six pairs for each size. All runs were included.</p><div><div><div>
Request size | Geometric-mean throughput change | Geometric-mean P99 change | Pairs with lower throughput / higher P99
-- | -- | -- | --
KV: 3,274,752 B | −1.83% | +5.05% | 12/12 / 12/12
4 KiB | −3.08% | +5.95% | 12/12 / 12/12

</div></div></div><p>The throughput change remained negative under both execution orders. The P99 increase also persisted beyond the initial pairs: across pairs 5–12, it was +4.07% for KV and +5.85% for 4 KiB. All executions completed with zero unexpected allocation results.</p><p>I also recorded CPU utilization every second, thread-level <code dir="ltr">pidstat</code> output, and <code dir="ltr">/usr/bin/time -v</code> statistics. The SMT siblings of the benchmark CPUs averaged approximately 0.3% non-idle time. However, the host remained shared, CPU frequency scaling was enabled, and <code dir="ltr">perf stat</code> was denied by the current <code dir="ltr">perf_event_paranoid=4</code> setting, so environmental effects are not fully excluded.</p><p>These results strengthen the evidence for a small, repeatable slowdown in this setup. The KV regression is smaller than our earlier −6.3% aggregate, which should not be treated as a fixed cost. The underlying cause remains open; I have retained the per-run outputs and monitoring logs for further comparison.</p>

Given the reproducible utilization gains in the KV/DSA workloads and the essentially unchanged throughput in the full DSA matrix, I consider this trade-off acceptable. But still, we should leave root-cause analysis as follow-up work. I would not block this PR solely on the small regression observed in this benchmark.

yokinoshitayoki:
@982945902 Also, please resolve the branch conflicts here.

982945902:
Rebased onto the latest `main` and resolved the conflicts while preserving the newer snapshot validation checks. I also fixed the 4-bit bitmap validation and made the rounding-bound test precision-independent.

Validation: `offset_allocator_test` — 60/60 passed; PR-scoped pre-commit hooks passed.

yokinoshitayoki:
@982945902 Please have a look at the failed CI checks.

982945902:
Fixed in `d9acf2a6`.

Root cause: the MsgPack serializer still iterated the widened `m_usedBins` array as `unsigned char`, truncating the upper 8 leaf-bin bits. This corrupted restored allocator bitmaps and caused the snapshot-related CTest failures. The serializer now preserves the full `uint16_t` value, and a focused high-bit MsgPack round-trip regression test was added.

Validation:
- `offset_allocator_test`: 61/61 passed
- all 9 allocator/snapshot-focused tests: passed
- exact CI format command with clang-format 20: passed
- pre-commit: passed
- unrelated `ShmTransportTest.RelocateCopySurvivesConcurrentCap`: passed 20/20 local reruns (the prior CI failure appears flaky)

A new CI run has been triggered.

982945902:
The latest CTest failure was a compile error in `batch_oplog_snapshot_gc_test`: `FakeBackend` became abstract after `HaKvBackend::DeleteRange` was added. This was fixed upstream by #4131.

I merged the latest `main` (including #4131) into this branch and pushed `579c1b77`. Local validation passed:

- `batch_oplog_snapshot_gc_test`: 2/2 passed
- `offset_allocator_test`: 61/61 passed
- `./scripts/code_format.sh --check --changed-lines --base origin/main`
- `pre-commit run --files <PR files>`

A fresh CI run is now in progress.

PR #3788, opened 2026-08-31
[Bugfix][Store] Fix tenant-quota eviction treadmill and watermark collision

### LINKED ISSUE #3741: [Bug]: Over-quota tenants silently lose writes — RFC #2153's rejection path is unreachable because eviction always succeeds
### Bug Report

With `-enable_multi_tenants=true` on `0.3.13`, a tenant that exceeds its effective quota gets
**`rc=0` on every write while the master discards the data on arrival**. No rejection, no
`over_quota` flag, nothing in the logs, and nothing in the global eviction counters. From the
caller's side it is indistinguishable from a successful write followed by a cache miss.

RFC #2153, which this feature implements, specifies the opposite:

> If this fails, Mooncake may run tenant-scoped eviction for that tenant. **If eviction cannot
> create enough headroom, admission fails with a tenant quota error.**

> On memory shrink: **New writes are rejected until usage returns below effective quota.**

We think the rejection path is not merely unused but **structurally unreachable**: tenant-scoped
eviction picks victims from the same tenant, and a tenant that is actively writing always has
recently-written evictable objects. So "eviction cannot create enough headroom" never becomes
true, and admission never fails — eviction just consumes the tenant's own new data.

### Reproduction

Pool: 600 GiB across 3 store nodes. Two tenants registered through
`PUT /api/v1/tenant_quotas`:

```
default  requested = 1024 TiB   effective = 599.99 GiB
probe2   requested =   10 GiB   effective =   5.9 MB
```

(The tiny effective quota is just proportional scaling doing its job — see the note at the
bottom, it is not the bug.)

A client with `tenant_id="probe2"` then writes 256 × 1 MiB objects and reads back what is
present:

```
wrote_ok=256  rejected=0  of 256 MiB      <- every put returned rc=0
still_present=5/256
after_45s_present=5/256                   <- stable, not a transient

probe2:  charged_bytes=5242880  over_quota=false  admission_closed=false
mooncake_tenant_evict_bytes_total{tenant_id="probe2"} 275775488
```

251 of 256 objects were accepted and thrown away. `charged_bytes` sits exactly at the quota, so
accounting is correct — the tenant is simply being trimmed to its share on every write.

### Three separate problems

**1. Writes report success for data that is immediately discarded.**
Per RFC #2153 this should be `TENANT_QUOTA_EXCEEDED` once eviction cannot make room, but
eviction can always make room by evicting what was just written. A caller has no way to
distinguish "stored" from "accepted and dropped", so there is no signal to back off, shrink the
working set, or alert. For a cache this is a survivable semantic, but it should be an explicit
one rather than an accident of eviction ordering.

RFC #2153 leaves this as an open question:

> Should over-quota eviction run synchronously on mount/unmount, asynchronously, or only on
> future writes?

Our request is narrower: whatever the eviction policy, there should be a way to make an
over-quota write **fail** instead of silently evicting. A flag would be enough
(`-tenant_quota_reject_on_exceed`), or a rule that eviction may not select objects written by
the request currently being admitted.

**2. `mooncake_tenant_quota_reject_total` is declared but never exported.**
RFC #2153 specifies `mooncake_tenant_quota_reject_total{tenant_id,reason}`. The string is
present in the binary, but the metric does not appear in `/metrics` at all — before, during, or
after the run above. Anyone building an alert on it gets permanent silence. (This may simply
follow from problem 1: if the counter is only registered on first increment and rejection never
happens, it can never appear. Either way it is not usable as an alarm.)

**3. Tenant-quota eviction is invisible in the global eviction counters.**
Throughout the 263 MiB evicted above:

```
master_evicted_key_count      0
master_evicted_size_bytes     0
master_successful_evictions_total 0
```

Only `mooncake_tenant_evict_bytes_total` moved. A dashboard watching the documented global
eviction metrics shows a completely idle cluster while a tenant loses 98% of its writes. These
are real evictions and we would expect them to be counted as such, or for the relationship
between the two counter families to be documented.

### Environment

- `docker.io/kvcacheai/mooncake:0.3.13`, Kubernetes, 3 store nodes × 200 GiB, RDMA
- Master: `-enable_multi_tenants=true -tenant_quota_connector_type=etcd
  -tenant_quota_connector_uri=<etcd>:2379 -eviction_high_watermark_ratio=0.9 -eviction_ratio=0.05`
- Client: `MooncakeDistributedStore.setup(..., tenant_id="probe2")`

### Expected

At least one of:

- an over-quota write fails with `TENANT_QUOTA_EXCEEDED`, as RFC #2153 describes; or
- it is documented that over-quota is enforced by eviction and never by rejection, with
  `mooncake_tenant_quota_reject_total` removed rather than left dead, and tenant eviction
  reflected in the global eviction counters so it is at least observable.

### Aside, not the bug

The 5.9 MB effective quota above comes from proportional scaling working as designed:
`10 GiB / (1024 TiB + 10 GiB) × 600 GiB`. Setting one tenant to a very large requested value —
the natural way to express "this tenant gets the whole pool, whatever size it is today" — makes
every subsequently added tenant a rounding error. That is arguably correct arithmetic, but it is
a sharp edge, and it interacts badly with problem 1: the newly added tenant is starved *and*
silent. A warning when a tenant's effective quota falls below some fraction of its requested
value would have saved us an afternoon.

- [x] Searched existing issues and read the documentation. #2153 (RFC, closed) and #2612 (the
      implementing PR, closed) are the design and implementation this report is measured
      against; #3157 is a separate refactor of quota charge/release.

huangxiaobai-ydmy:
## Description

Fixes #3741.

Implements the production-side guidance from @gongwei-130 on the two tenant-quota regimes:

1. **Semantic fix**
   - `EvictTenantMemoryForQuota` now frees `deficit + eviction_ratio * effective_quota` headroom (not exactly the deficit), to reduce the concurrent admission treadmill.
   - On tenant-quota shortfall during PutStart/UpsertStart, set `need_mem_eviction_` so `BatchEvict` can still run when `used_ratio` sits exactly on `eviction_high_watermark_ratio`.
   - Optional `-tenant_quota_reject_on_exceed` to reject immediately instead of evicting.
   - Startup/recompute warnings when effective quota `<=` watermark bytes, or effective quota is far above charged bytes.

2. **Observability**
   - Keep `mooncake_tenant_*` counters.
   - Also increment global `inc_eviction_success` / `inc_eviction_fail` (and mem variants) from the tenant eviction path so `master_attempted_evictions_total` is not silent for tenant eviction.

3. **Tests (both required cases)**
   - Silent-eviction: 256 Puts succeed with only a small residual set under quota.
   - Watermark collision: effective quota == watermark * capacity, hard-pinned fill, next Put rejects and arms `need_mem_eviction_`.

@gongwei-130 — if you can replay this patch on the production collision cluster, that would be very helpful.

## Module

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

Added unit coverage in `master_service_tenant_quota_test.cpp`. Full build not run locally (Windows without Mooncake toolchain); please rely on CI.

**Test commands:**
```bash
# CI: Build & Test (Linux)
# Local: ctest --test-dir build -R master_service_tenant_quota_test
```

**Test results:**
- [ ] Unit tests pass
- [x] Manual testing done (code review of eviction/admission paths; unit tests added)

## Checklist

## AI Assistance Disclosure

Cursor AI assisted with issue analysis against #3741 discussion, implementation of headroom / need_mem_eviction_ / metrics / warnings, regression tests, and PR drafting. The submitter reviewed the changes.

### DIFF
--- a/mooncake-store/include/master_config.h
+++ b/mooncake-store/include/master_config.h
@@ -157,6 +157,8 @@ struct MasterConfig {
     bool enable_disk_eviction;
     uint64_t quota_bytes;
     bool enable_multi_tenants = false;
+    // When true, over-quota Puts reject immediately instead of tenant eviction.
+    bool tenant_quota_reject_on_exceed = false;
     std::string tenant_quota_connector_type = "file";
     std::string tenant_quota_connector_uri;
 
@@ -286,6 +288,8 @@ class MasterServiceSupervisorConfig {
     bool enable_disk_eviction = true;
     uint64_t quota_bytes = 0;
     bool enable_multi_tenants = false;
+    // When true, over-quota Puts reject immediately instead of tenant eviction.
+    bool tenant_quota_reject_on_exceed = false;
     std::string tenant_quota_connector_type = "file";
     std::string tenant_quota_connector_uri;
     uint32_t max_total_finished_tasks = DEFAULT_MAX_TOTAL_FINISHED_TASKS;
@@ -456,6 +460,7 @@ class MasterServiceSupervisorConfig {
         enable_disk_eviction = config.enable_disk_eviction;
         quota_bytes = config.quota_bytes;
         enable_multi_tenants = config.enable_multi_tenants;
+        tenant_quota_reject_on_exceed = config.tenant_quota_reject_on_exceed;
         tenant_quota_connector_type = config.tenant_quota_connector_type;
         tenant_quota_connector_uri = config.tenant_quota_connector_uri;
 
@@ -617,6 +622,8 @@ class WrappedMasterServiceConfig {
     bool enable_disk_eviction = true;
     uint64_t quota_bytes = 0;
     bool enable_multi_tenants = false;
+    // When true, over-quota Puts reject immediately instead of tenant eviction.
+    bool tenant_quota_reject_on_exceed = false;
     std::string tenant_quota_connector_type = "file";
     std::string tenant_quota_connector_uri;
 
@@ -842,6 +849,7 @@ class WrappedMasterServiceConfig {
         enable_disk_eviction = config.enable_disk_eviction;
         quota_bytes = config.quota_bytes;
         enable_multi_tenants = config.enable_multi_tenants;
+        tenant_quota_reject_on_exceed = config.tenant_quota_reject_on_exceed;
         tenant_quota_connector_type = config.tenant_quota_connector_type;
         tenant_quota_connector_uri = config.tenant_quota_connector_uri;
         put_start_discard_timeout_sec = config.put_start_discard_timeout_sec;
@@ -915,6 +923,7 @@ class MasterServiceConfigBuilder {
     bool enable_disk_eviction_ = true;
     uint64_t quota_bytes_ = 0;
     bool enable_multi_tenants_ = false;
+    bool tenant_quota_reject_on_exceed_ = false;
     std::string tenant_quota_connector_type_ = "file";
     std::string tenant_quota_connector_uri_;
     uint64_t put_start_discard_timeout_sec_ = DEFAULT_PUT_START_DISCARD_TIMEOUT;
@@ -1098,6 +1107,11 @@ class MasterServiceConfigBuilder {
     MasterServiceConfigBuilder& set_allocation_strategy_type(
         AllocationStrategyType type) {
         allocation_strategy_type_ = type;
+        return *this;
+    }
+
+    MasterServiceConfigBuilder& set_tenant_quota_reject_on_exceed(bool enable) {
+        tenant_quota_reject_on_exceed_ = enable;
         return *this;
     }
 
@@ -1332,6 +1346,8 @@ class MasterServiceConfig {
     bool enable_disk_eviction = true;
     uint64_t quota_bytes = 0;
     bool enable_multi_tenants = false;
+    // When true, over-quota Puts reject immediately instead of tenant eviction.
+    bool tenant_quota_reject_on_exceed = false;
     std::string tenant_quota_connector_type = "file";
     std::string tenant_quota_connector_uri;
 
@@ -1425,6 +1441,7 @@ class MasterServiceConfig {
         enable_disk_eviction = config.enable_disk_eviction;
         quota_bytes = config.quota_bytes;
         enable_multi_tenants = config.enable_multi_tenants;
+        tenant_quota_reject_on_exceed = config.tenant_quota_reject_on_exceed;
         tenant_quota_connector_type = config.tenant_quota_connector_type;
         tenant_quota_connector_uri = config.tenant_quota_connector_uri;
         put_start_discard_timeout_sec = config.put_start_discard_timeout_sec;
@@ -1505,6 +1522,7 @@ inline MasterServiceConfig MasterServiceConfigBuilder::build() const {
     config.enable_disk_eviction = enable_disk_eviction_;
     config.quota_bytes = quota_bytes_;
     config.enable_multi_tenants = enable_multi_tenants_;
+    config.tenant_quota_reject_on_exceed = tenant_quota_reject_on_exceed_;
     config.tenant_quota_connector_type = tenant_quota_connector_type_;
     config.tenant_quota_connector_uri = tenant_quota_connector_uri_;
     config.enable_snapshot_restore = enable_snapshot_restore_;
--- a/mooncake-store/include/master_service.h
+++ b/mooncake-store/include/master_service.h
@@ -1037,6 +1037,9 @@ class MasterService {
     };
     TenantQuotaEvictionResult EvictTenantMemoryForQuota(
         const TenantId& tenant_id, uint64_t target_bytes);
+    // Warn when tenant effective quotas collide with the pool eviction
+    // watermark or dwarf steady-state charge (operator UX; #3741).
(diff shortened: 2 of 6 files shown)


huangxiaobai-ydmy:
cc @gongwei-130 — patch is up in this PR if you can validate on the collision cluster. Thanks again for the detailed production measurements.

he-yufeng (commented):
Read the full #3741 thread and the diff, then walked the eviction paths on current main to check the arming semantics. The direction matches what the production data points to: freeing exactly the deficit is the treadmill, and the two new tests pin the two regimes from the issue table. The headroom target with overflow saturation, the global counter coverage, and `tenant_quota_reject_on_exceed` all line up with what the reporter asked for.

One design question on the `need_mem_eviction_` arming in PutStart/UpsertStart. The flag forces a `BatchEvict` pass even when `used_ratio` sits far below the watermark: `EvictionThreadFunc` only needs `need_mem_eviction_ && eviction_ratio_ > 0.0`, and BatchEvict picks victims globally with no tenant weighting. In the collision regime from gongwei-130's measurement (tenant effective quota near the watermark, that tenant owning most of the pool) this is exactly right, since the bulk pass releases the constrained tenant's own charge through the quota ledger release in `evict_replicas`, and it is what absorbed the 20x write rate in their cluster.

But the arming fires on every non-final over-quota attempt, including the opposite regime: a tenant with a small effective quota (the probe2 shape from the original report) going over quota while the pool is mostly empty and owned by other tenants. There the bulk pass cannot relieve the constrained tenant, because its charged bytes only drop when its own objects are picked, so each armed pass evicts roughly eviction_ratio of the other tenants' hot objects below the watermark and the Put still rejects after retries. A tenant pushing a steady stream of over-quota writes would keep the flag armed most of the time.

Two ways to scope it, both keep the collision fix:

1. Arm only when the tenant's effective quota can plausibly push the pool over the watermark (say `effective_quota_bytes >= watermark_bytes`), which is precisely the regime where the bulk path helps.
2. Arm only when `EvictTenantMemoryForQuota` reports it could not free enough, matching the RFC's "if eviction cannot create enough headroom" language. This one misses the case gongwei-130 measured, where tenant eviction frees 25-198 MB/s continuously yet the treadmill persists, so the first option seems safer.

Not a blocker if the broad arming is a deliberate tradeoff, but the choice is worth making explicitly rather than by accident.

Scope note: read-only review against current main plus the diff; I did not build or run the new tests, and CI was still queued when I looked.

huangxiaobai-ydmy:
Thanks @he-yufeng — agreed the previous unconditional arming was too broad.

Adopted your option 1:

- `HandleTenantQuotaAdmissionEviction` now arms `need_mem_eviction_` **only when** `effective_quota_bytes >= eviction_high_watermark_ratio * capacity`
- PutStart / UpsertStart both call that helper (no duplicated arming logic)
- Added `SmallTenantQuotaExceedDoesNotArmBulkEviction` so the small-quota regime stays on tenant eviction / reject without waking global `BatchEvict`

Collision case (`TenantQuotaWatermarkCollisionArmsNeedMemEvictionOnReject`) still expects the flag to be armed.

catyans (approved):
LGTM. Re-reviewed the current head after @he-yufeng's comment: bulk eviction is now armed only when a tenant's effective quota can pin the pool, while deficit-plus-headroom eviction, the reject switch, warnings, and global metrics cover the two #3741 regimes. CI is green.

suggestion (non-blocking): @huangxiaobai-ydmy @gongwei-130, when integrating #3805, please make this arming conditional on the per-tenant watermark being disabled (or otherwise choose one trigger). Dropping it unconditionally would leave existing default-off configurations exposed to the collision; keeping both when enabled can trigger an extra global eviction pass.

huangxiaobai-ydmy:
Thanks for the LGTM, @catyans.

Agreed on the follow-up with #3805: when that lands, we should keep a single trigger — e.g. only arm need_mem_eviction_ when the per-tenant watermark is disabled (or otherwise pick one path). Dropping this arming unconditionally would leave default-off configs exposed to the collision; enabling both could cause an extra global BatchEvict pass.

Happy to help wire that guard when #3805 is integrated.

huangxiaobai-ydmy:
Rebased onto the latest `main` (`b905925`) to clear merge conflicts.

Resolved the conflict in the snapshot test base by keeping main's `AssertRestoredClientAffiliations` check and retaining the restored-service eviction freeze from this PR.

huangxiaobai-ydmy:
Hi @ykwd @stmatengss @XucSh @YiXR — gentle ping when you have a moment.

CI is green again after the rebase onto latest `main`, and @catyans has already approved. This PR only touches the tenant-quota eviction / watermark path (plus a small snapshot-test TearDown race fix). If the direction looks good, would you mind taking a look for a CODEOWNER review when convenient? No rush, and happy to address any further feedback. Thanks!

catyans (approved):
LGTM. Re-reviewed the rebased head. The tenant-quota fix is unchanged by the rebase, and the snapshot conflict resolution keeps main's restored-client affiliation check while freezing the restored eviction worker before comparing state. That closes the test race without changing the production eviction path. CI is green again.

Thanks for calling out exactly how the conflict was resolved — that made the re-review pleasantly straightforward. My earlier note about choosing a single trigger when #3805 is integrated remains non-blocking.

huangxiaobai-ydmy:
Thanks for re-reviewing after the rebase, @catyans — really appreciate the careful look.

PR #3790, opened 2026-08-31
[TENT]: Map MOONCAKE_LOCAL_HOSTNAME to TENT rpc_server_hostname

Colors-111:
## Description

`MOONCAKE_LOCAL_HOSTNAME` is the standard environment variable used by the
classic Transfer Engine and the store to specify the local RPC bind address
(read at store startup; sglang's `MooncakeStore` config reads it too). TENT
did not previously read it — the RPC bind address could only come from the
`rpc_server_hostname` field of the `MC_TENT_CONF` JSON or be passed
explicitly via the constructor. When neither is set, TENT falls back to
`discoverLocalIpAddress()` to auto-discover the local IP.

Under container / hostNetwork multi-NIC environments, `discoverLocalIpAddress`
takes the first non-lo IPv4 from `getifaddrs`, which can select a container
network IP (e.g. `10.154.0.1`) instead of the RDMA-network IP
(e.g. `10.15.56.196`). That IP is written into `SegmentDesc.rpc_server_addr`
and published to metadata, so the peer's RDMA handshake goes over the wrong
network and cross-node transfers time out / fail.

## Change

Map `MOONCAKE_LOCAL_HOSTNAME` to TENT's `rpc_server_hostname` in
`ConfigHelper::loadFromEnv`, consistent with how `MC_RDMA_BIND_ADDRESS` and
`MC_TE_FILTERS` reuse classic env vars:

```cpp
setConfig(config, "MOONCAKE_LOCAL_HOSTNAME", "rpc_server_hostname");
```

After the mapping, a single `MOONCAKE_LOCAL_HOSTNAME` works for both the
classic and TENT engines — no need to duplicate `rpc_server_hostname` in the
TENT JSON or maintain a separate env set for TENT.

## Precedence

`loadFromEnv` runs before `restoreExplicitTransferEngineConfig`; the store's
TENT branch in `TransferEngine::init` only passes `local_segment_name` /
`metadata_type` / `metadata_servers` explicitly, not `rpc_server_hostname`,
so the env-set value is not overwritten by restore. Same as
`MC_RDMA_BIND_ADDRESS`: env wins over `MC_TENT_CONF` JSON, which suits
per-pod injection.

## Tests

- `TransferEngineConfigOverrideTest.LocalHostnameEnvLoadsIntoRpcServerHostname`:
  sets `MOONCAKE_LOCAL_HOSTNAME` and asserts `rpc_server_hostname` is populated
- `TransferEngineConfigOverrideTest.LocalHostnameEnvOverridesMcTentConf`:
  JSON sets one value + env sets another; asserts env wins

## Files Changed

- `mooncake-transfer-engine/tent/src/common/config.cpp` — add `MOONCAKE_LOCAL_HOSTNAME` → `rpc_server_hostname` mapping in `loadFromEnv`
- `mooncake-transfer-engine/tent/tests/transfer_engine_config_override_test.cpp` — two regression tests

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
```

**Test results:**
- [X] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

### DIFF
--- a/mooncake-transfer-engine/tent/src/common/config.cpp
+++ b/mooncake-transfer-engine/tent/src/common/config.cpp
@@ -135,6 +135,7 @@ Status ConfigHelper::loadFromEnv(Config& config) {
     }
 
     // Legacy keys for backward compatibility (MC_* env vars)
+    setConfig(config, "MOONCAKE_LOCAL_HOSTNAME", "rpc_server_hostname");
     setConfig(config, "MC_RDMA_BIND_ADDRESS", "transports/rdma/bind_address");
     setConfig(config, "MC_NUM_CQ_PER_CTX",
               "transports/rdma/device/num_cq_list");
--- a/mooncake-transfer-engine/tent/tests/transfer_engine_config_override_test.cpp
+++ b/mooncake-transfer-engine/tent/tests/transfer_engine_config_override_test.cpp
@@ -299,9 +299,8 @@ TEST(TransferEngineConfigOverrideTest,
 }
 
 TEST(TransferEngineConfigOverrideTest, LocalHostnameEnvOverridesMcTentConf) {
-    EnvVarGuard conf_guard(
-        "MC_TENT_CONF",
-        R"({"rpc_server_hostname":"10.0.0.1"})");
+    EnvVarGuard conf_guard("MC_TENT_CONF",
+                           R"({"rpc_server_hostname":"10.0.0.1"})");
     EnvVarGuard host_guard("MOONCAKE_LOCAL_HOSTNAME", "10.0.0.2");
 
     Config config;
--- a/mooncake-transfer-engine/tent/tests/transfer_engine_config_override_test.cpp
+++ b/mooncake-transfer-engine/tent/tests/transfer_engine_config_override_test.cpp
@@ -283,6 +283,35 @@ TEST(TransferEngineConfigOverrideTest,
     EXPECT_EQ(config.get("transports/rdma/bind_address", ""), "10.0.0.2");
 }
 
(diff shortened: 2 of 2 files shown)


staryxchen (commented):
loadFromEnv mapping is the wrong layer, and it is not how classic TE works. Classic TE never reads MOONCAKE_LOCAL_HOSTNAME; Store parses local_hostname with parseHostNameWithPort and passes (hostname, port) into TransferEngine::init.

On the default Store P2P path, classic TE uses that host as the RPC bind address. The TENT init branch drops ip_or_host_name / rpc_port, keeps only local_segment_name, and falls back to discoverLocalIpAddress(). That is the misalignment: TE trusts Store’s host, TENT takes the first non-lo IPv4.

Please forward ip_or_host_name to rpc_server_hostname in the TENT init path so it matches TE. Forward rpc_port only when the caller actually wants a bind port; P2P TE still picks a free port itself. Do not copy the raw env into rpc_server_hostname — values like localhost:17814 will fail checkLocalIpAddress().

JSON/setup() hostnames with no env var also need this init path; the current tests only cover a portless IP.

catyans (approved):
LGTM. This correctly maps `MOONCAKE_LOCAL_HOSTNAME` into TENT while preserving explicit caller configuration precedence. The override paths are covered and all checks are green.

alogfans (approved):

PR #3282, opened 2026-08-04
[TENT] Add native UB foundation and URMA resource management

zchuango:
## Description

This PR introduces the foundation for a TENT-native UB transport backed by the openEuler UMDK URMA APIs.

It is the first implementation stage of RFC #3093 and establishes the native URMA abstraction, resource ownership model, UB topology representation, build integration, and deterministic mock-based validation required by the later endpoint and data-path stages.

This PR supersedes #2828 as part of the new implementation series. Unlike #2828, which adapted TENT to the existing Classic Transfer Engine UB data path, this series implements UB as a TENT-native transport. TENT owns transport-level policy and lifecycle, while the URMA adapter is limited to native URMA operations and resource management.

Related:

- RFC: #3093
- Superseded implementation: #2828

#### 1. Add UB identities to TENT

This PR adds UB as a distinct TENT transport and topology type:

- `TransportType::UB`
- `Topology::NIC_UB`
- `"ub"` transport-name parsing and round-trip conversion
- `TRANSPORT_UB` in the public C API
- UB transport exposure through Python bindings
- Default UB configuration entries

Existing transport and topology enum values remain unchanged. UB is appended as a new value to preserve the serialized numeric values used by existing transports.

#### 2. Extend topology metadata for native UB devices

The TENT topology model now supports the identity fields required by URMA device discovery:

- Native device name
- Device index
- EID index
- EID
- Active state
- PCI bus ID
- NUMA node

UB discovery is optional and adapter-backed. The existing one-argument `Topology::discover()` behavior is preserved, while UB-aware discovery must be explicitly enabled.

Topology JSON parsing remains backward-compatible: topology descriptions that do not contain the new fields continue to load with safe defaults. Serialization adds the new optional identity fields for all NIC entries.

#### 3. Add an injectable URMA adapter

This PR introduces a TENT-owned URMA adapter boundary covering low-level native operations such as:

- URMA runtime initialization and shutdown
- Device and EID discovery
- Context creation and destruction
- JFC and JFCE lifecycle management
- Local segment registration
- Remote segment import
- Jetty creation, binding, reset, quiesce, and destruction
- READ/WRITE work-request posting
- Completion polling

The adapter does not contain TENT scheduling, retry, transport selection, admission-control, or receiver-credit policy.

The injectable interface allows the resource lifecycle to be tested without UB hardware or a real `liburma` runtime.

#### 4. Add native context, JFC, and buffer resource management

This PR adds the resource-management components required by the later native UB transport:

[...]

Issue #3093, opened 2026-07-24
[RFC]: TENT-Native UB/URMA Transport


zchuango:
## Summary

This RFC proposes adding a **TENT-native UB transport** backed by the openEuler UMDK URMA APIs.

The goal is to make UB a first-class TENT transport whose runtime-facing interfaces and architectural responsibilities are aligned with the existing TENT RDMA transport.

The intended data path is:

```text
TENT Request
    → UbTask / UbSlice
    → TENT UB Workers
    → UbContext / UbEndpoint / UbJfc
    → URMA Adapter
    → liburma
```

TENT owns task scheduling, slicing, endpoint lifecycle, telemetry, cancellation, and failure recovery. The URMA adapter is limited to native resource management and URMA API translation.

The initial hardware scope is **host DRAM-to-DRAM transfer**. Additional memory types should be enabled only after their registration, topology, and access semantics have been validated.

---

## Related Work

UB/URMA support was previously introduced for the classic Mooncake Transfer Engine in #1773, #1805, and #1855. That work established the underlying URMA device, memory-registration, endpoint, build, mock, and testing capabilities.

This RFC builds on that foundation but focuses on a different architectural scope: implementing UB as a TENT-native transport with TENT-owned tasks, slices, workers, endpoint lifecycle, scheduling signals, cancellation, and failure recovery.

---

## Motivation

TENT provides transport-independent runtime capabilities such as:

- transport selection;
  
- task and slice lifecycle management;
  
- admission control and priority scheduling;
  
- cancellation;
  
- receiver-side credits;
  
- runtime bandwidth estimation;
  
- path failover and health monitoring.
  

To participate fully in this model, UB should expose the same logical capabilities as TENT RDMA.

A native implementation allows the TENT scheduler to observe UB runtime information such as:

- UB device availability;
  
- Jetty and remote-path health;
  
- JFC completion pressure;
  
- outstanding requests and inflight bytes;
  
- estimated bandwidth;
  
- endpoint generations;
  
- timeout and error statistics.
  

---

## Proposed Architecture

```mermaid
flowchart TD
    APP[Application / Store] --> RT[TENT Runtime]

    RT --> AQ[Admission Queue]
    AQ --> TS[Transport Selector]
    TS --> RC[Receiver Credit Gate]

    RC --> UB[UbTransport]
    UB --> WK[UbTask / UbSlice / Workers]

    WK --> PS[Device and Path Selection]
    PS --> CTX[UbContext]
[...]

PR #3796, opened 2026-08-31
[TENT] Add native UB endpoint bootstrap and READ/WRITE data path

zchuango:
## Description

This PR continues the TENT-native UB/URMA work from RFC #3093 and follows
#3282 (foundation + URMA resource management).

It adds the native UB control-plane bootstrap path and the READ/WRITE data
path under TENT, and includes bonding-oriented device selection / runtime
fixes needed for dual-node UB stability.

Related:

- RFC: #3093
- Foundation: #3282
- Prior adapter-oriented attempt (superseded by the native series): #2828

#### 1. UB control-plane bootstrap

- Dedicated `UbBootstrapDesc` (not RDMA QP fields)
- `ControlClient::bootstrapUb()` / `ControlService::onBootstrapUb()`
- Injectable UB bootstrap callback registration

#### 2. Native UB transport data path

- `UbTransport`, endpoint / endpoint store, workers, slices
- Local/remote path setup through the URMA adapter boundary
- Mock-backed unit coverage for core and native data-path behavior

#### 3. Bonding device selection and runtime hardening

- Prefer bonding devices when `device_filter` is empty
- Bonding-mode / CTP priority / multi-path jetty configuration in the adapter
- Unified bonding detection and stronger segment teardown

#### 4. CI

- Keep / restore the `tent-ci (ub-mock)` matrix coverage
- Add a small `tebench` CLI help smoke check for UB flags under `ub-mock`

### Compatibility and impact

- `USE_UB=OFF`: native UB target is not built; existing transports unchanged
- `USE_UB=ON`: TENT can select UB, bootstrap endpoints, and run READ/WRITE
  through the native path (hardware required for real dual-node validation)
- RDMA bootstrap validation (`decodeBootstrapResponse`) from main is preserved
  alongside the UB bootstrap APIs

### Out of scope

- Full Phase 5 reliability stack (advanced failover / telemetry / credit
  integration beyond what is required for this data path)
- Replacing Classic TE Kunpeng UB
- Claiming production readiness without dual-node hardware sign-off

### Review focus

1. Bootstrap protocol and control-plane boundary vs RDMA
2. Endpoint generation / teardown safety
3. Bonding selection behavior when filters are empty
4. Mock tests vs what still requires real UB hardware

## Module

- [x] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

### Mock / CI

Validated on LinQuickDev against latest synced `main`:

- Internal CI PR: https://github.com/LinQuickDev/Mooncake/pull/40 (all checks green, including `tent-ci (ub-mock)`)

### Unit tests

```bash
# USE_TENT=ON USE_UB=ON Debug / mock adapter (as in tent-ci ub-mock)
ctest --test-dir build-tent/mooncake-transfer-engine/tent/tests --output-on-failure
```

### Dual-node (real UB)

Validated on a dual-node Kunpeng UB setup with bonding devices.

- Build: `-DUSE_TENT=ON -DUSE_UB=ON`
- Transport: TENT + UB (`--backend=tent --xport_type=ub`)
- Path: READ/WRITE over UB bonding
- Result: end-to-end data transfer completed successfully on both nodes

**Test results:**

- [x] Unit tests / mock CI pass
- [x] Manual dual-node testing done

## Checklist

- [x] For changes >500 LOC: I have filed an RFC issue (#3093)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -945,6 +945,17 @@ jobs:
       run: ./scripts/ci/run_transfer_engine_rust_smoke.sh
       shell: bash
 
+    - name: Smoke test TENT UB benchmark CLI
+      if: matrix.name == 'ub-mock'
+      run: |
+        cd build-tent
+        help_output="$(./mooncake-transfer-engine/benchmark/tebench \
+          --backend=tent --xport_type=ub --tent_transport_hint=ub \
+          --help 2>&1 || true)"
+        grep -q 'iouring|ub|sunrise_link' <<< "${help_output}"
+        grep -q 'ascend|ub|sunrise_link' <<< "${help_output}"
+      shell: bash
+
     - name: Run sccache stat for check
       if: ${{ env.SCCACHE_PATH != '' }}
       shell: bash
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -707,6 +707,17 @@ jobs:
       run: ./scripts/ci/run_transfer_engine_rust_smoke.sh
       shell: bash
 
+    - name: Smoke test TENT UB benchmark CLI
+      if: matrix.name == 'ub-mock'
+      run: |
+        cd build-tent
+        help_output="$(./mooncake-transfer-engine/benchmark/tebench \
+          --backend=tent --xport_type=ub --tent_transport_hint=ub \
+          --help 2>&1)"
+        grep -q 'iouring|ub|sunrise_link' <<< "${help_output}"
+        grep -q 'ascend|ub|sunrise_link' <<< "${help_output}"
+      shell: bash
+
     - name: Run sccache stat for check
       if: ${{ env.SCCACHE_PATH != '' }}
       shell: bash
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -949,17 +949,6 @@ jobs:
       run: ./scripts/ci/run_transfer_engine_rust_smoke.sh
       shell: bash
 
-    - name: Smoke test TENT UB benchmark CLI
-      if: matrix.name == 'ub-mock'
-      run: |
-        cd build-tent
-        help_output="$(./mooncake-transfer-engine/benchmark/tebench \
-          --backend=tent --xport_type=ub --tent_transport_hint=ub \
-          --help 2>&1 || true)"
-        grep -q 'iouring|ub|sunrise_link' <<< "${help_output}"
-        grep -q 'ascend|ub|sunrise_link' <<< "${help_output}"
-      shell: bash
-
     - name: Run sccache stat for check
       if: ${{ env.SCCACHE_PATH != '' }}
       shell: bash
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -713,7 +713,7 @@ jobs:
         cd build-tent
         help_output="$(./mooncake-transfer-engine/benchmark/tebench \
           --backend=tent --xport_type=ub --tent_transport_hint=ub \
-          --help 2>&1)"
+          --help 2>&1 || true)"
         grep -q 'iouring|ub|sunrise_link' <<< "${help_output}"
         grep -q 'ascend|ub|sunrise_link' <<< "${help_output}"
       shell: bash
--- a/mooncake-transfer-engine/tent/src/runtime/topology.cpp
+++ b/mooncake-transfer-engine/tent/src/runtime/topology.cpp
@@ -108,11 +108,6 @@ void Topology::print() const {
 }
 
 Status Topology::discover(const std::vector<Platform*>& platforms) {
-    return discover(platforms, false);
-}
-
-Status Topology::discover(const std::vector<Platform*>& platforms,
-                          bool discover_ub) {
     clear();
     for (auto& entry : platforms) {
         CHECK_STATUS(entry->probe(nic_list_, mem_list_));
@@ -121,7 +116,7 @@ Status Topology::discover(const std::vector<Platform*>& platforms,
     // UB discovery is intentionally adapter-backed instead of inferring UB
     // devices from verbs/sysfs names. One topology NIC is emitted per EID and
     // carries both the globally serialized identity and the native URMA name.
-    auto adapter = discover_ub ? ub::createDefaultUrmaAdapter() : nullptr;
+    auto adapter = ub::createDefaultUrmaAdapter();
     if (adapter && adapter->available()) {
         auto status = adapter->initialize();
         if (status.ok()) {
@@ -182,7 +177,6 @@ Status Topology::discover(const std::vector<Platform*>& platforms,
         }
     }
 #endif
-    (void)discover_ub;
     return Status::OK();
 }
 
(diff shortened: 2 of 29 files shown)


mooncake-transfer-engine/tent/src/runtime/topology.cpp
```diff
@@ -365,7 +366,7 @@ Status Topology::loadFromConfig(const Config& conf,
             LOG(WARNING) << "Failed to parse topology/priority_matrix: "
                          << status.ToString()
                          << ", falling back to auto-discover";
-            return discover(platforms);
+            return discover(platforms, discover_ub);
```
alogfans: I’m not convinced Topology::discover() should take a transport-enable driven flag like discover_ub.
Topology discovery should ideally describe hardware/runtime facts, while whether UB is enabled feels like a higher-layer policy that belongs in transport loading/selection.

Would it make more sense to keep discover(platforms) transport-agnostic, always discover UB when the runtime/platform supports it, and then gate actual UbTransport activation in the loader with transports/ub/enable?

If the concern is probe cost or provider side effects, I’d still prefer expressing that as a neutral discovery policy/options object, rather than plumbing a transport feature toggle into the Topology API.

alogfans:
@ascend-direct-dev Are there any suggestions?

catyans:
I found one liveness/accounting gap in the missing-completion recovery path.

After `quiesce()` succeeds, `scanTimeouts()` resolves the logical attempt and may enqueue a retry, but a token for which the provider returned no completion remains in `inflight_`. Consequently, `releaseInflight()` is never called for that attempt, so its endpoint outstanding count and quota reservation remain charged until `stop()`. Repeated lost completions can therefore exhaust endpoint/path capacity and stall later transfers even though each successful fence has already proved that the old WR can no longer DMA.

The current `MissingCompletionIsFencedBeforeRetryAndEventuallyCompletes` test does not cover this case because `FakeUrmaAdapter::quiesceJetty()` synthesizes an `ENDPOINT_ERROR` completion for every pending WR, which removes the token through `handleCompletion()`.

Could the successful-fence path explicitly remove and release any still-present timed-out token before scheduling its retry, and add a mock case where quiesce succeeds without returning that token? The regression should verify that `inflightCount()` and quota/outstanding accounting return to zero and that subsequent transfers still make progress.

mooncake-transfer-engine/tent/src/runtime/topology.cpp
zchuango: Thanks for the careful look — I agree with the layering concern. discover_ub does leak a transport-level policy into what should be a hardware/runtime fact query.

I'll rework this so that discover(platforms) stays transport-agnostic and always reports UB when the platform supports it, and gate UbTransport activation in the loader via transports/ub/enable instead. If probe cost becomes a concern we can revisit with a neutral discovery options object, but for now the loader gate should be enough.

I'll update the PR shortly — happy to hear @ascend-direct-dev's thoughts as well.

zchuango:
> I found one liveness/accounting gap in the missing-completion recovery path.
> 
> After `quiesce()` succeeds, `scanTimeouts()` resolves the logical attempt and may enqueue a retry, but a token for which the provider returned no completion remains in `inflight_`. Consequently, `releaseInflight()` is never called for that attempt, so its endpoint outstanding count and quota reservation remain charged until `stop()`. Repeated lost completions can therefore exhaust endpoint/path capacity and stall later transfers even though each successful fence has already proved that the old WR can no longer DMA.
> 
> The current `MissingCompletionIsFencedBeforeRetryAndEventuallyCompletes` test does not cover this case because `FakeUrmaAdapter::quiesceJetty()` synthesizes an `ENDPOINT_ERROR` completion for every pending WR, which removes the token through `handleCompletion()`.
> 
> Could the successful-fence path explicitly remove and release any still-present timed-out token before scheduling its retry, and add a mock case where quiesce succeeds without returning that token? The regression should verify that `inflightCount()` and quota/outstanding accounting return to zero and that subsequent transfers still make progress.

Great catch — you're right. In the successful-fence path the timed-out token can stay in inflight_ if the provider never returns its completion, so releaseInflight() is never called and the endpoint outstanding count + quota reservation stay charged until stop(). Repeated lost completions would indeed exhaust capacity.

I'll fix it by explicitly removing and releasing any still-present timed-out token after a successful quiesce() and before scheduling the retry (safe, since the fence proves the old WR can no longer DMA; the CAS in releaseInflight() keeps it idempotent against a racing completion). I'll also add a regression test where quiesce() succeeds without returning the token, asserting inflightCount() and quota/outstanding accounting return to zero and subsequent transfers still make progress.

Thanks for the detailed analysis — the note about FakeUrmaAdapter::quiesceJetty() synthesizing completions explains exactly why the existing test missed it.

zchuango:
I'll fix it by explicitly removing and releasing any still-present timed-out token after a successful quiesce() and before scheduling the retry — safe, since the fence proves the old WR can no longer DMA, and the CAS in releaseInflight() keeps it idempotent against a racing completion. I'll also add a regression test where quiesce() succeeds without returning the token, asserting inflightCount() and quota/outstanding accounting return to zero and subsequent transfers still make progress.

Thanks for the detailed analysis — the note about FakeUrmaAdapter::quiesceJetty() synthesizing completions explains exactly why the existing test missed it.

catyans (approved):
Thanks, this addresses my concern. LGTM. After a successful fence, the timed-out token is removed and endpoint/quota accounting is released exactly once before retry; the late-completion race remains idempotent. The new no-completion regression verifies zero inflight/outstanding/reservations and successful follow-up progress. CI is green.

zchuango:
@alogfans @catyans  @ascend-direct-dev The latest update includes the UB recovery safeguards and regression coverage; all CI checks are green. Could you please take a look when you have a moment?

catyans (approved):
LGTM. Re-reviewed current head `e54826c`: the new slice cap rejects oversized requests before task allocation, and incomplete endpoint retirement is now quarantined instead of replaced or reused. The `main` merge also preserves both UB and HP-TCP configuration/test coverage. No new correctness blocker found; CI is currently queued.

alogfans (approved):
LGTM. @ykwd

.github/workflows/ci.yml
```diff
@@ -949,6 +949,17 @@ jobs:
       run: ./scripts/ci/run_transfer_engine_rust_smoke.sh
       shell: bash
 
+    - name: Smoke test TENT UB benchmark CLI
```
Aionw: This smoke test only verifies the static gflags help strings. Since `--help` exits during flag parsing, neither the TENT configuration nor the UB transport mapping is exercised. The `|| true` also discards the process exit status.

The `ub-mock` build and CTest coverage already exercise the actual UB code paths, so I think this step can be removed.

zchuango: Thanks for catching this — agreed, the help-string grep doesn't exercise the TENT config or the UB transport mapping, and || true masks the exit status. The step has been removed in ; the ub-mock build and CTest coverage remain the real gate for the UB paths.

PR #3797, opened 2026-08-31
[TENT]: Deduplicate TentMetrics singleton via shared tent_metrics library

Colors-111:
## Description

`TentMetrics` is a process-wide singleton — `TentMetrics::instance()` holds a function-local `static TentMetrics instance`, and `initialized_` is guarded by `compare_exchange_strong` so each process initializes it at most once and starts a single 30s periodic report thread (`metric_report_thread_`, the source of the recurring `TENT Metrics: ...` log line).

`tent_metrics` was previously a **STATIC** library: it was packaged into the static `transfer_engine`, which was then **copied** into both the `engine.so` and `store.so` pybind extensions at link time. The result was two mutually invisible `static TentMetrics instance` objects plus two `initialized_` guards in the same process. When sglang `import mooncake.engine` (TENT KV migration) and `import mooncake.store` (`--hicache-storage-backend mooncake`) together, each scheduler process would:

1. start **two** report threads, logging `TENT Metrics:` twice per interval  (TP4 = 8 lines/interval instead of 4);
2. have both singletons bind the default HTTP port `9100` — **only one succeeds**; the other logs `HTTP endpoint unavailable ... continuing with log-only metrics` and silently degrades, so Prometheus scrapes only one engine's metrics and the other is lost.

## Trigger conditions

The duplicate-singleton problem only appears when all three hold:

1. **`TENT_METRICS_ENABLED=ON`** (compile time). An OFF build strips the entire metrics code path via `#if`; record macros are no-ops and no HTTP server is created, so there is no double-thread / port issue to begin with.
2. **`USE_TENT=ON`**. `transfer_engine` links `tent_metrics` only under this flag; non-TENT `engine.so` / `store.so` carry no metrics code and are unaffected.
3. **The same process loads both `mooncake.engine` and `mooncake.store` pybind extensions**. Each `.so` statically embeds its own `TentMetrics`; the singleton guard only prevents re-entry into the *same* instance, not the *other* copy.

The typical trigger is sglang with hierarchical cache backed by mooncake:

```
python -m sglang.launch_server --tp 4 \
  --enable-hierarchical-cache \
  --hicache-storage-backend mooncake \
  --hicache-storage-backend-extra-config '{"master_server_address": "etcd://...", ...}'
```

`mooncake.engine` serves TENT KV migration and `mooncake.store` serves the `--hicache-storage-backend mooncake` storage backend; both are imported inside the same scheduler process, so each rank ends up with two metrics instances.

**Does not trigger**: loading only `mooncake.engine` (TENT migration without a mooncake storage backend) or only `mooncake.store`; `TENT_METRICS_ENABLED=OFF`; `USE_TENT=OFF`. These deployments load metrics code once, so the singleton is naturally unique.

## Root-cause evidence

`nm -D` shows `engine.so` and `store.so` **each export** their own `TentMetrics::instance` (different addresses), confirming the static code was duplicated:

```
engine.so  00000000005043c0 T _ZN8mooncake4tent11TentMetrics8instanceEv
store.so   0000000000e0d0b0 T _ZN8mooncake4tent11TentMetrics8instanceEv
```

Mapping TID → PID (`ps -eLo pid,tid,comm`) shows 4 scheduler processes with 2 report threads each, not 8 processes — ruling out "multiple sglang instances" and confirming "two singletons in one process".

## Change

Switch `tent_metrics` from STATIC to SHARED:

```cmake
add_library(tent_metrics SHARED ${TENT_METRICS_SOURCES})
```

As a shared library, `engine.so` / `store.so` no longer embed the metrics code; they take a single `DT_NEEDED` dependency on `libtent_metrics.so`. The dynamic loader loads that library once (by soname), so both `.so`s reference the same mapping and share the one process-wide `TentMetrics` singleton — 8 threads collapse back to 4, the port clash disappears, and Prometheus sees the full picture.

## Mechanism

- **PIC**: `tent/CMakeLists.txt` sets `CMAKE_POSITION_INDEPENDENT_CODE ON` globally, so OBJECT libraries such as `tent_common` are already compiled PIC and link into the shared object with no extra change.
- **Symbol visibility**: after the fix, `TentMetrics::instance` is exported by `libtent_metrics.so` (`T`) and appears as `U` (undefined, resolved at load time from `DT_NEEDED`) inside `engine.so` / `store.so` / `libtent_shared.so`.
- **RPATH**: `engine.so` / `store.so` use `INSTALL_RPATH "$ORIGIN"`, so installing `libtent_metrics.so` alongside them makes it discoverable — the same pattern `asio_shared` already uses (installed to both `lib/` and the python package directory).

## Guards

- `transfer_engine` links `tent_link_group` (which includes `tent_metrics`) only under `if(USE_TENT)`, so `engine.so` / `store.so` depend on `libtent_metrics.so` only when `USE_TENT` is on.
- The install is guarded by `if(TARGET tent_metrics)`, aligning naturally with `USE_TENT`; non-TENT builds produce no `libtent_metrics.so` and incur no such dependency.

## Tests

- **Symbol check** (root cause): `nm -D engine.so` / `store.so` shows `TentMetrics::instance` as `T` (defined) before the fix and `U` (imported from `libtent_metrics.so`) after.
- **Runtime**: before the fix each scheduler process had 2 report threads (TP4 = 8 lines/interval); after, 1 (4 lines/interval). The `HTTP endpoint unavailable` warning disappears and port 9100 binds uniquely.
- **Existing tests**: `metrics_config_loader_test` / `tent_metrics_http_server_test` / `tent_metrics_recording_test` link `tent_metrics` directly; the SHARED switch does not change their behavior and they keep passing.

## Files Changed

- `mooncake-transfer-engine/tent/src/metrics/CMakeLists.txt` — switch `tent_metrics` from STATIC to SHARED and install it to `lib/`.
- `mooncake-integration/CMakeLists.txt` — install `libtent_metrics.so` to the python package directory (`MOONCAKE_PYTHON_INSTALL_DIR`) alongside `engine.so` / `store.so` for `$ORIGIN` resolution.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
```

**Test results:**
- [ ] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

### DIFF
--- a/mooncake-transfer-engine/rust/build.rs
+++ b/mooncake-transfer-engine/rust/build.rs
@@ -176,7 +176,6 @@ fn main() {
             build_dir.join("mooncake-common/src"),
             build_dir.join("mooncake-common/etcd"),
             build_dir.join("mooncake-transfer-engine/tent/src"),
-            build_dir.join("mooncake-transfer-engine/tent/src/metrics"),
             build_dir.join("src"),
             build_dir.join("src/common/base"),
         ] {
@@ -201,7 +200,6 @@ fn main() {
         manifest_dir.join("../../build/mooncake-common"),
         manifest_dir.join("../../build/mooncake-common/etcd"),
         manifest_dir.join("../../build/mooncake-transfer-engine/tent/src"),
-        manifest_dir.join("../../build/mooncake-transfer-engine/tent/src/metrics"),
         manifest_dir.join("../tent/build/src"),
     ] {
         push_dir(&mut search_dirs, dir);
--- a/mooncake-transfer-engine/rust/build.rs
+++ b/mooncake-transfer-engine/rust/build.rs
@@ -176,6 +176,11 @@ fn main() {
             build_dir.join("mooncake-common/src"),
             build_dir.join("mooncake-common/etcd"),
             build_dir.join("mooncake-transfer-engine/tent/src"),
+            // libtent_metrics.so is a SHARED lib built under tent/src/metrics.
+            // collect_tent_archives skips it (only .a), and has_library's
+            // /usr/local/lib fallback only gates the -l emit, not the -L search
+            // path — so the linker still needs this directory explicitly.
+            build_dir.join("mooncake-transfer-engine/tent/src/metrics"),
             build_dir.join("src"),
             build_dir.join("src/common/base"),
         ] {
@@ -200,6 +205,7 @@ fn main() {
         manifest_dir.join("../../build/mooncake-common"),
         manifest_dir.join("../../build/mooncake-common/etcd"),
         manifest_dir.join("../../build/mooncake-transfer-engine/tent/src"),
+        manifest_dir.join("../../build/mooncake-transfer-engine/tent/src/metrics"),
         manifest_dir.join("../tent/build/src"),
     ] {
         push_dir(&mut search_dirs, dir);
--- a/mooncake-transfer-engine/rust/build.rs
+++ b/mooncake-transfer-engine/rust/build.rs
@@ -176,6 +176,7 @@ fn main() {
             build_dir.join("mooncake-common/src"),
             build_dir.join("mooncake-common/etcd"),
             build_dir.join("mooncake-transfer-engine/tent/src"),
+            build_dir.join("mooncake-transfer-engine/tent/src/metrics"),
             build_dir.join("src"),
             build_dir.join("src/common/base"),
         ] {
@@ -200,6 +201,7 @@ fn main() {
         manifest_dir.join("../../build/mooncake-common"),
         manifest_dir.join("../../build/mooncake-common/etcd"),
         manifest_dir.join("../../build/mooncake-transfer-engine/tent/src"),
+        manifest_dir.join("../../build/mooncake-transfer-engine/tent/src/metrics"),
         manifest_dir.join("../tent/build/src"),
     ] {
         push_dir(&mut search_dirs, dir);
@@ -268,6 +270,9 @@ fn main() {
         }
         if has_mooncake_common {
             println!("cargo:rustc-link-arg=-lmooncake_common");
+        }
+        if has_library(&search_dirs, "tent_metrics") {
+            println!("cargo:rustc-link-arg=-ltent_metrics");
         }
         println!("cargo:rustc-link-arg=-Wl,--end-group");
         println!("cargo:rustc-link-lib=dl");
--- a/mooncake-transfer-engine/tent/src/metrics/CMakeLists.txt
+++ b/mooncake-transfer-engine/tent/src/metrics/CMakeLists.txt
@@ -1,6 +1,18 @@
 # TENT_METRICS_ENABLED option is declared in tent/CMakeLists.txt.
 file(GLOB TENT_METRICS_SOURCES "*.cpp")
-add_library(tent_metrics STATIC ${TENT_METRICS_SOURCES})
+# Build as a SHARED library so the TentMetrics singleton (the function-local
+# `static TentMetrics instance` in TentMetrics::instance()) is not duplicated
+# across DSOs. When statically linked, the metrics code is copied into every
+# consumer .so (engine.so and store.so), giving each its own TentMetrics
+# instance: a process that loads both modules ends up with two independent
+# metrics singletons, two periodic report threads ("TENT Metrics:" logged
+# twice per interval), and a metrics HTTP port clash (both default to 9100,
+# so only one binds and the other silently degrades to log-only). As a shared
+# library, both modules take a single DT_NEEDED on libtent_metrics.so and
+# share the one process-wide singleton. PIC is already on for the whole tent
+# tree (tent/CMakeLists.txt sets CMAKE_POSITION_INDEPENDENT_CODE ON), so the
+# tent_common OBJECT dependency links cleanly into the shared object.
+add_library(tent_metrics SHARED ${TENT_METRICS_SOURCES})
 # Link yalantinglibs::yalantinglibs for metrics and HTTP server (ylt headers +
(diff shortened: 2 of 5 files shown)


alogfans:
Seems the build of TENT fails.

catyans:
The shared-library direction makes sense, but the Rust TENT link path was not updated. All four TENT matrix failures reach the Rust smoke link and then fail on unresolved `TentMetrics` / `MetricsConfigLoader` symbols.

`mooncake-transfer-engine/rust/build.rs` discovers only TENT `.a` archives. After this change, the link command has neither the nested metrics build directory in its search/RPATH list nor `-ltent_metrics`, so `libtent.a`'s metrics references remain unresolved. Please add the shared metrics library to the TENT Rust link path (including standalone-build discovery) and keep the Rust smoke test covering this configuration.

It would also be worth validating a custom-prefix install of `libtent_shared.so`: CI shows its install RPATH is empty, so it should either carry an `$ORIGIN` lookup for the sibling `libtent_metrics.so` or have another tested runtime-resolution mechanism.

mooncake-transfer-engine/tent/src/metrics/CMakeLists.txt
```diff
@@ -1,6 +1,21 @@
...
+# twice per interval), and a metrics HTTP port clash (both default to 9100,
+# so only one binds and the other silently degrades to log-only). As a shared
+# library, both modules take a single DT_NEEDED on libtent_metrics.so and
+# share the one process-wide singleton. PIC is already on for the whole tent
+# tree (tent/CMakeLists.txt sets CMAKE_POSITION_INDEPENDENT_CODE ON), so the
+# tent_common OBJECT dependency links cleanly into the shared object.
+add_library(tent_metrics SHARED ${TENT_METRICS_SOURCES})
+set_target_properties(
+  tent_metrics
+  PROPERTIES INSTALL_RPATH "$ORIGIN" BUILD_WITH_INSTALL_RPATH TRUE)
```
staryxchen: INSTALL_RPATH "$ORIGIN" is needed so the installed libtent_metrics.so finds sibling libasio.so. Is BUILD_WITH_INSTALL_RPATH TRUE necessary?

mooncake-transfer-engine/tent/src/metrics/CMakeLists.txt
```diff
@@ -20,3 +35,13 @@ else()
...
+
+# Install to lib/ for libtent_shared.so / system consumers. The python C
+# extensions (engine.so, store.so) resolve it via $ORIGIN, so a second
+# install lands libtent_metrics.so next to them — see
+# mooncake-integration/CMakeLists.txt.
+install(
+  TARGETS tent_metrics
+  LIBRARY DESTINATION lib
+  ARCHIVE DESTINATION lib
+  RUNTIME DESTINATION bin)
```
staryxchen: This is a Linux SHARED library. Only LIBRARY DESTINATION lib is used. `ARCHIVE DESTINATION lib` is for import libraries (Windows) and `RUNTIME DESTINATION bin` is for DLLs. They do nothing here and suggest a Windows layout this target does not have. Please drop them.

mooncake-transfer-engine/rust/build.rs
```diff
@@ -269,6 +271,9 @@ fn main() {
         if has_mooncake_common {
             println!("cargo:rustc-link-arg=-lmooncake_common");
         }
+        if has_library(&search_dirs, "tent_metrics") {
+            println!("cargo:rustc-link-arg=-ltent_metrics");
+        }
```
staryxchen, line 280: `-ltent_metrics` is the necessary Rust fix. These two hardcoded .../tent/src/metrics entries are not.

collect_tent_archives already walks the TENT build tree; it just only accepts .a. After cmake --install, has_library also finds libtent_metrics.so under /usr/local/lib. Hardcoding two nested CMake output dirs duplicates that, and still misses the standalone layout you already list as ../tent/build/src (the .so is in ../tent/build/src/metrics).

catyans:
Thanks, the Rust link path and sibling `$ORIGIN` lookup are fixed. The remaining TENT failures now come from `metrics_config_loader_test`: `libtent_metrics.so` is built with `BUILD_WITH_INSTALL_RPATH TRUE`, so its build-tree RPATH is only `$ORIGIN` and it cannot find `libasio.so` in the separate build directory.

Please keep the normal build-tree RPATH and apply `$ORIGIN` only at install time (or otherwise make `asio_shared` directly resolvable by this test), then rerun the TENT matrix.

Colors-111:
@catyans @staryxchen All three points addressed:

Dropped `BUILD_WITH_INSTALL_RPATH` TRUE — build-tree RPATH now resolves `libasio.so` during ctest; `$ORIGIN` applies only at install time.
Trimmed `ARCHIVE/RUNTIME` install destinations (Linux SHARED lib).
Added `-ltent_metrics` to the Rust `--start-group`.
The two `.../tent/src/metrics` search dirs must stay, though.` collect_tent_archives` only gathers .a (skips the .so), and has_library's `/usr/local/lib` fallback only gates the -l emit — it does not add a -L to the linker. Removing them caused the previous CI run to fail with `unable to find library -ltent_metrics`. They're load-bearing for the link step, kept with an explanatory comment.

catyans (approved):
Thanks @Colors-111, this addresses my remaining concern. LGTM. The build tree now keeps CMake's normal RPATH for `libasio.so`, installed artifacts retain `$ORIGIN`, and Rust finds and links `tent_metrics` inside the rescan group. All 25 checks are green.

alogfans (approved):
LGTM

Issue #3523, opened 2026-08-19
[TENT] RDMA link-flap recovery gaps: hung WRs wedge QPs silently, swallowed bootstrap rejection, no PORT_ACTIVE fallback


ai2open:
# [TENT] RDMA link-flap recovery gaps: hung WRs wedge QPs silently (no error WC, no software timeout, no reconnect), plus swallowed bootstrap rejection and missing PORT_ACTIVE fallback

## Scenario

Prefill/Decode disaggregation (KV cache transfer) over RoCE with the TENT backend
(`MC_USE_TENT=1`), RDMA NICs are bonded (`mlx5_bond_*`).

**Observed behavior:** after PD transfer timeouts, or an RDMA NIC link down/up,
subsequent transfers never recover on their own. The classic (non-TENT) engine does
not show this behavior in the same environment.

## Controlled experiment (single-host repro)

`ip link set <netdev of mlx5_bond_4> down`, wait ~4 min, `ip link set up`.

Timeline (all from logs):

- `08:18:06` — link down; all 8 engine processes correctly log `Action: mlx5_bond_4 down`.
- `08:19:03-08:19:54` — `fallback device selection` floods (~780k reroute attempts);
  throughput degrades but does not drop to zero (other NICs still serve).
- `08:19:35+` — application-level 20 s timeouts (`batch_transfer_sync_write TIMEOUT`,
  ret=-1).
- `08:21:54` — link up; all 8 processes correctly log `Action: mlx5_bond_4 up`
  (so PORT_ACTIVE handling worked in this run).
- After link up: **partial recovery only**. All 4 pp_ranks submitted RDMA writes
  (25 each), but pp_rank 1/2 succeeded 25/25 while **pp_rank 0/3 failed 25/25** —
  and stayed broken until the end of the log (16+ min later).

The crucial part: for the failing ranks after link-up, the TENT layer logged
**absolutely nothing**:

- 0 × `Detected error WQE` (no error completions)
- 0 × `transfer timeout (software)` (slice-level software timeout never fired)
- 0 × `Unable to connect endpoint` (no reconnect attempts/failures)
- 0 × `Endpoint marked for destruction`

Slices were submitted at the app level and then vanished: never completed, never
errored, never reported by any TENT diagnostic path.

### Defect A (primary, confirmed by experiment): hung WRs during link-down wedge the QP/endpoint forever

WRs posted to QPs on the downed NIC hang silently: while the port is down the
[...]

PR #3799, opened 2026-08-31
[Bugfix][TENT] Drain the async event queue and recover contexts whose PORT_ACTIVE was lost

TTThanos:
## Description

**Problem**
Issue #3523 defect C. A TENT RDMA context that was paused by IBV_EVENT_PORT_ERR only ever leaves DEVICE_PAUSED from one place: the IBV_EVENT_PORT_ACTIVE branch of Workers::applyContextEvent(). Nothing revalidates the port afterwards, so if that one event is missed the context stays paused for the rest of the process' life, Workers::getEndpoint() rejects every transfer on that NIC, and no log line ever explains why.

That event is easy to miss. joinNonblockingPollList() registers the async fd with EPOLLIN | EPOLLET, while monitorThread() calls epoll_wait(..., maxevents=1, 100) and handleContextEvents() consumed exactly one ibv_async_event per wakeup. Under edge-triggered semantics epoll does not report readiness again until a new event arrives, so anything queued behind the event that woke us is stranded. Bursts are routine — IBV_EVENT_COMM_EST fires once per connection — so a PORT_ACTIVE sitting behind one waits for an unrelated future event that may never come. The classic engine hit exactly this and fixed it in WorkerPool::doProcessContextEvents(); TENT still had the one-event-per-wakeup read.

**Fix**
Workers::handleContextEvents() now drains the queue: loop until EAGAIN/EWOULDBLOCK, retry on EINTR, and propagate a genuine read error instead of spinning on a broken fd. Mirrors the classic engine.
Workers::resumePausedContexts() is a safety net on monitorThread()'s existing 1 Hz tick: for contexts in DEVICE_PAUSED it reads the port state and, when the hardware reports IBV_PORT_ACTIVE, runs the same recovery the event path runs. Intermediate states (INIT/ARMED/ACTIVE_DEFER) are left alone and re-checked on the next tick.
The recovery actions (resume() + re-seed bandwidth + setDeviceAvailable) are factored into Workers::activateContext() so the event path and the poll cannot drift apart.
New RdmaContext::queryPortState() reads the port state without touching the cached speed/width, because refreshLinkSpeed() diffs against them — refreshing there would hide a renegotiated link rate from the selector.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Tests**
New tent/tests/rdma_async_event_drain_test.cpp (5 cases), scripting the event source with -Wl,--wrap=ibv_get_async_event, so it needs no RDMA device: full drain of a burst, a handled event queued behind unhandled ones, empty queue is not an error, EINTR is retried, a real read error aborts. All 5 fail on main and pass with this change.
tent/tests/rdma_transport_test.cpp: RecoveryPollLeavesInertContextsAlone (the poll must not hand an inert slot back to the selector), QueryOnInertContextIsRejected, and PollActivatesPausedContextWithLivePort — the regression itself, on a real NIC (skipped when no device is present).

**Test commands:**
```bash
```

**Test results:**
- [x] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

I use AI tools to help summarise the PR content, assist and review on my code.

### DIFF
--- a/mooncake-transfer-engine/tent/include/tent/transport/rdma/context.h
+++ b/mooncake-transfer-engine/tent/include/tent/transport/rdma/context.h
@@ -123,6 +123,15 @@ class RdmaContext {
     // open or the query fails.
     int refreshPortAttributes();
 
+    // Read the port's current state from the hardware. The monitor thread
+    // polls this for paused contexts so a link that came back without a
+    // usable IBV_EVENT_PORT_ACTIVE is still noticed. Deliberately leaves the
+    // cached speed/width alone: refreshLinkSpeed() diffs against them, so
+    // refreshing here would hide a renegotiated rate from the selector.
+    // Returns -1 and leaves *state untouched if no device is open or the
+    // query fails.
+    int queryPortState(ibv_port_state *state) const;
+
     int eventFd() const { return event_fd_; }
 
     RdmaCQ *cq(int index);
--- a/mooncake-transfer-engine/tent/include/tent/transport/rdma/workers.h
+++ b/mooncake-transfer-engine/tent/include/tent/transport/rdma/workers.h
@@ -84,6 +84,9 @@ class Workers {
 
     // dev_id is the NicID (context_set_ index), which is also the
     // DeviceSelector id, so port events can flip that device's availability.
+    // Drains the whole async event queue: the async fd is edge-triggered and
+    // ibv_get_async_event() dequeues one record per call, so an event left
+    // behind waits for an unrelated later event to release it.
     int handleContextEvents(int dev_id, std::shared_ptr<RdmaContext>& context);
 
     // The decision half of handleContextEvents, kept apart from
@@ -94,10 +97,23 @@ class Workers {
     void applyContextEvent(int dev_id, RdmaContext& context,
                            const ibv_async_event& event);
 
+    // Everything a recovered port needs: resume the context, re-seed its
+    // bandwidth and make it selectable again. Shared by the
+    // IBV_EVENT_PORT_ACTIVE path and by resumePausedContexts().
+    void activateContext(int dev_id, RdmaContext& context);
+
     // Re-read the link speed after a port event and re-seed the selector if
     // it changed; a link that returns at the same speed keeps what it
     // learned.
     void refreshLinkSpeed(int dev_id, RdmaContext& context);
+
+    // 1 Hz heartbeat from monitorThread(): safety net for a lost
+    // IBV_EVENT_PORT_ACTIVE. Nothing else ever leaves DEVICE_PAUSED, so a
+    // context whose recovery event was dropped (edge-triggered fd, event
+    // queue overflow, ...) would fail every transfer on that NIC forever.
+    // Polls the port state of paused contexts and activates the ones the
+    // hardware reports as up.
+    void resumePausedContexts();
 
     Status generatePostPath(RdmaSlice* slice);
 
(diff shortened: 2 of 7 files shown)


he-yufeng (commented):
Read the full diff and chased the invariants I was worried about; this is a careful fix for a real defect class.

Verified on my read:

- The starvation mechanism is real: `joinNonblockingPollList` registers the async fd EPOLLET and the old loop consumed one event per wakeup, so anything queued behind it only surfaces when a new event arrives. COMM_EST bursts make that routine, and a stranded PORT_ACTIVE is a silent permanent outage for that NIC.
- The drain loop gets the errno handling right for a non-blocking fd (EAGAIN/EWOULDBLOCK drains, EINTR retries, anything else propagates).
- Stale-event resurrection is not possible: the queue is FIFO per device, so a PORT_ACTIVE queued before a later PORT_ERR is always processed in that order, and `resume()` is CAS-guarded from DEVICE_PAUSED, so a double activation is a no-op rather than a crash or a state rewrite.
- Poll and event handling both run on monitorThread, so the 1 Hz net and the drain cannot race each other.
- `queryPortState` leaving the cached speed/width alone is the right call; `refreshLinkSpeed` diffs against the cache to detect renegotiation, and refreshing at query time would hide exactly that.
- The wrap-based test is honestly constructed: no device needed, the Apple linker guard is correct, and scripting the drained errno covers EAGAIN/EINTR/real-error plus handled/unhandled event mix.

One operational nit, not a blocker: every drained event still logs at WARNING, and before this change most COMM_EST events were never dequeued at all (which was the bug, but also meant they were never logged). After the drain, a connection burst produces a WARNING line per event, so on a busy cluster the log will get loud fast. Worth demoting COMM_EST to INFO or VLOG in a follow-up.

Scope note: read-only review against the PR head; I did not build or run the tests here. Leaving the formal approve to a second pair of eyes since this is the RDMA context path.

alogfans (approved):
LGTM. However I advise to avoid mass logs from high-frequent async events such as IBV_EVENT_COMM_EST.

catyans (approved):
LGTM. The edge-triggered async-event queue is now drained with the correct `EAGAIN`/`EINTR` handling, and the recovery poll only reactivates paused contexts whose port is fully active while preserving link-speed re-seeding. The regression coverage matches these paths.

The remaining `MasterServiceSSDSnapshotTest.EvictObject` failure appears unrelated to this RDMA-only change; please rerun it before merge.

catyans (approved):
LGTM. Re-reviewed the current head; the new commit only demotes the high-frequency `IBV_EVENT_COMM_EST` message to `VLOG(1)`, while preserving event handling and acknowledgement behavior.

The PR is currently conflicting with the base branch and only the triage check is present, so please resolve the conflict and run the full CI before merge.

alogfans:
It seems OK but you may check your code format.

catyans:
issue: the merge resolution dropped the closing brace for `RdmaContext::queryPortState()` after `return 0;`, so `queryEffectiveSpeed()` now starts inside that function and the current head cannot compile. Please restore the `}` before the next member-function definition.

catyans (approved):
Thanks, this addresses my concern. LGTM. The current head restores the missing closing brace after `queryPortState()`; the remaining incremental changes are formatting-only, and the async-event drain/recovery logic is unchanged. CI is green.

Issue #3769, opened 2026-08-29
[RFC]:  A non-leasing existence probe (`ProbeKey` / `BatchProbeKey`)


duhaode520:
> Related PR: #3599.

## Motivation

`ExistKey` / `BatchExistKey` grant a read lease on every hit (`master_service.cpp:2780-2787` and `:2851`). This is the right behavior when an existence check means "I am about to read this object". It is the wrong behavior when the check is a speculative scan.

Speculative scans are common in KV-cache reuse: to determine how much of a cached prefix it can reuse, a client probes a set of candidate keys and then reads back only the portion it needs. The probe set is substantially larger, and issued far more frequently, than the set that is actually read — yet every probed hit currently acquires a lease.

Leases only extend, and are refreshed at half the TTL (`master_service.h:1302-1313`), so with the default 10s TTL any key re-probed within ~5s remains leased indefinitely. Eviction skips every leased object (`master_service.cpp:10250`). Together, these mean a scan-heavy client can pin effectively its entire working set.

On a small MEMORY tier with a big SSD tier, this is more than a hit-rate regression. Eviction is also the demotion path to SSD, so nothing drains, MEMORY remains full, and `Put` begins to fail — despite ample free capacity on the SSD tier behind it.

## Proposal

Provide a way to ask whether a key exists without acquiring a retention lease:

> **Point-in-time existence check.** A `true` result means the object existed at the time of the call. It may be evicted before a subsequent `Get`; callers must treat a following miss as normal.

Reads remain protected in either case — `GetReplicaList` grants its own lease (`master_service.cpp:3678`, `:3865`), so the read-in-flight window is unaffected.

There are two plausible ways to expose this, and I do not have a strong enough case to settle it unilaterally — input on the trade-off would be welcome.

### Option A — new `ProbeKey` / `BatchProbeKey` RPC

A twin of `ExistKey` / `BatchExistKey` that shares the lookup path and skips the lease.

This is purely additive: existing clients continue to call `ExistKey`, new clients opt into `ProbeKey`, and rolling upgrades remain safe in both directions. `ExistKey`'s contract is untouched, and the two operations can be documented separately with distinct guarantees.

[...]

PR #3801, opened 2026-08-31
[Store] Add ProbeKey/BatchProbeKey RPCs for lease-free existence probes

duhaode520:
## Description

Implements Option A of #3769: add `ProbeKey` / `BatchProbeKey`, lease-free twins of `ExistKey` / `BatchExistKey`, for point-in-time existence checks.

Unlike `ExistKey`, a successful probe grants no read lease, so speculative scans (e.g. probing candidate keys for KV-cache prefix reuse) can no longer pin the working set against eviction and block the demotion path to SSD. A `true` result only means the object existed at the time of the call; callers must treat a following miss as normal. Reads remain protected since `GetReplicaList` grants its own lease.

The new RPCs share the same lookup path as `ExistKey` (refactored into a `grant_lease`-parameterized impl) and cross all client layers: `WrappedMasterService` (coro_rpc), `MasterClient`, `Client`, `RealClient` / `DummyClient` (`PyClient`), and the Python binding (`probe_key` / `batch_probe_key`), plus the Python API docs. `store_c.h` and the generated Rust bindings are intentionally untouched, per the RFC. Purely additive: existing clients are unaffected and rolling upgrades are safe in both directions.

## Module

- [x] Docs

## Type of Change

- [x] New feature

## How Has This Been Tested?

**Test commands:**
```bash
./build/mooncake-store/tests/master_service_test --gtest_filter='*ProbeKey*:*ExistKeyLeasesPinSegment*'
./build/mooncake-store/tests/pybind_client_test
```

**Test results:**
- [x] Unit tests pass

Added two tests:
- `ExistKeyLeasesPinSegmentButProbeKeyDoesNot`: on a full segment, `ExistKey` on every object makes the next `PutStart` fail with `NO_AVAILABLE_HANDLE` (leased objects are skipped by eviction), while with `ProbeKey` the allocation eventually succeeds by evicting the probed objects.
- `BatchProbeKeyReportsPointInTimeExistence`: batch semantics for existing / missing / duplicate keys.

Regression: full `master_service_test` 112/112, `pybind_client_test` 64/64, `master_service_group_test` 13/13, `master_metrics_test` 14/14.

## Checklist

## AI Assistance Disclosure

AI-assisted implementation (CodeBuddy), reviewed and verified line-by-line by the human submitter.


catyans:
Thanks for following up on #3769 and keeping the existing `ExistKey` contract intact. The explicit lease-free API is a much clearer direction.

issue: `Client::ProbeKey()` and `Client::BatchProbeKey()` are declared in `client_service.h`, and `RealClient` calls them, but there are no corresponding definitions in `client_service.cpp`. This is why the current CTest and wheel builds fail to link with undefined references to both methods. Please add the forwarding implementations; the batch path should also preserve the response-size validation used by `BatchIsExist()`.

duhaode520:
Thanks for catching this! The two definitions were accidentally dropped. Both methods are added back, now forward to MasterClient, and the batch path preserves the same response-size validation as BatchIsExist().

mooncake-store/src/dummy_client.cpp
```diff
@@ -928,6 +928,36 @@ std::vector<int> DummyClient::batchIsExist(
     return results;
 }
 
+int DummyClient::probeKey(const std::string& key) {
+    auto result = invoke_rpc<&RealClient::probeKey_internal, bool>(key);
```
fcczzz, line 932: Please register `RealClient::probeKey_internal` and `RealClient::batchProbeKey_internal` in `RegisterClientRpcService` (`real_client_main.cpp`). Neither handler is registered, so when using `setup_dummy()`, both new APIs fail with `function_not_registered` before reaching the master, even for existing keys. A test using the production RPC registration path would catch this; the new MasterService tests bypass that boundary.

ykwd:
This PR might also help with a case we're currently looking into.

In our multi-turn K3 workload, Mooncake lookup refreshes the LRU/lease state for all matched keys. Since we need to query all candidate hashes first to determine which positions have Mamba cache and then select the longest hit, this ends up refreshing Mamba cache entries from previous turns as well—even though we only actually load one of them.

We're currently investigating this behavior and looking into whether the observe API introduced in this PR could help with this case.

mooncake-store/src/dummy_client.cpp
duhaode520: Thanks — both handlers are now registered in `RegisterClientRpcService`.

  I also added `dummy_client_probe_key_test`, which drives `probeKey` / `batchProbeKey` through `DummyClient` over the real RPC boundary (verified it fails if either registration is
  dropped).

staryxchen (commented):
`ProbeKey` / `BatchProbeKey` as implemented here are the right additive split from `ExistKey`, but they don't cover the lookup pattern @ykwd described (and the one we have internally): probe every candidate hash, then load only the longest hit.
 `BatchExistKey` refreshes a lease on every hit, so unused prefix / Mamba entries stay pinned and block eviction. `BatchProbeKey` avoids that, but then existence and Get are no longer atomic — a `true` result is only point-in-time, and the key we actually want to load can be evicted before Get.
A boolean `grant_lease` (or `lease_select`) on the existing `ExistKey` / `BatchExistKey` RPCs is not viable: `coro_rpc` encodes arguments by method signature, so a new parameter breaks mixed-version rolling upgrades in both directions. That contract should stay “every hit is leased”.
Proposal: keep this PR's new RPCs, and extend **`BatchProbeKey` only** with a `lease_select` policy the client chooses up front (by key order + enum), not a callback after seeing results. Master runs scan → select → `GrantReadLease` on the 
  selected keys in one RPC:
  - `None` (default): current ProbeKey, no lease
  - `LongestPrefix`: keys are ordered by prefix length; lease `keys[0..k]`
  - `LongestPrefixLastOnly`: lease only `keys[k]`

If a selected key comes back `true`, it is already leased. Callers that will Get the whole prefix use `LongestPrefix`; callers that load a single object (e.g. one Mamba cache) use `LongestPrefixLastOnly`.
 This need not block the lease-free APIs in this PR, it can land as a follow-up. @ykwd does this match what we need for the K3 / Mamba lookup case?

yokinoshitayoki:
> `ProbeKey` / `BatchProbeKey` as implemented here are the right additive split from `ExistKey`, but they don't cover the lookup pattern @ykwd described (and the one we have internally): probe every candidate hash, then load only the longest hit. `BatchExistKey` refreshes a lease on every hit, so unused prefix / Mamba entries stay pinned and block eviction. `BatchProbeKey` avoids that, but then existence and Get are no longer atomic — a `true` result is only point-in-time, and the key we actually want to load can be evicted before Get. A boolean `grant_lease` (or `lease_select`) on the existing `ExistKey` / `BatchExistKey` RPCs is not viable: `coro_rpc` encodes arguments by method signature, so a new parameter breaks mixed-version rolling upgrades in both directions. That contract should stay “every hit is leased”. Proposal: keep this PR's new RPCs, and extend **`BatchProbeKey` only** with a `lease_select` policy the client chooses up front (by key order + enum), not a callback after seeing results. Master runs scan → select → `GrantReadLease` on the selected keys in one RPC:
> 
> * `None` (default): current ProbeKey, no lease
> * `LongestPrefix`: keys are ordered by prefix length; lease `keys[0..k]`
> * `LongestPrefixLastOnly`: lease only `keys[k]`
> 
> If a selected key comes back `true`, it is already leased. Callers that will Get the whole prefix use `LongestPrefix`; callers that load a single object (e.g. one Mamba cache) use `LongestPrefixLastOnly`. This need not block the lease-free APIs in this PR, it can land as a follow-up. @ykwd does this match what we need for the K3 / Mamba lookup case?

It's a good suggestion. But for K3 situation, selecting the last existing KDA checkpoint can lease an unusable state when its required KV prefix is missing, while leaving the earlier checkpoint actually selected by the engine unprotected. Maybe we can select the longest jointly restorable KV/KDA boundary and lease all keys required to restore it? But it's kind of complicated, do you have any better idea? Thanks.

staryxchen:
@yokinoshitayoki That's a fair concern. Our view is that matching KV prefixes with KDA checkpoints does not need to be something the store understands — that pairing belongs in the inference framework.

`BatchProbeKey` and `BatchExistKey` share the same lookup path. The only difference is what gets pinned: Probe pins nothing, Exist pins every hit. `lease_select` is just the middle ground — pin a subset of those hits in the same RPC. Whether that subset is actually a jointly restorable KV/KDA boundary is up to the caller: it prepares key order (and the policy) so the selected prefix is the set it will load. The store only applies the policy to `exist[]`; it does not know which keys are KV or KDA.

For the K3 case, that means using `LongestPrefix` (lease `keys[0..k]`) rather than `LongestPrefixLastOnly`, and putting the KV+KDA keys the engine will actually restore into that prefix. A later KDA whose KV prefix is missing would be pinned only if the caller lined it up that way — same as today, where `BatchExistKey` also does not check that pairing.

Does that line up with how you see the K3 lookup path?

yokinoshitayoki:
@staryxchen Thanks a lot for your clarification. As I understand it, LongestPrefix would still lease earlier KDA checkpoints if they are included in the selected prefix, even though only the final checkpoint is loaded. That may leave the original issue unresolved for K3.

Would a two-stage lookup work instead?

1. Use batch_is_exist to check and lease the KV blocks, determining the longest contiguous KV prefix.
2. Restrict the KDA candidates to that boundary, then use LongestPrefixLastOnly to select and lease only the last complete KDA checkpoint.

In K3 workload, KDA checkpoints are much larger than KV blocks, so some over-leasing on the KV side is acceptable; the main issue we want to avoid is retaining unused KDA checkpoints. Does this approach sound feasible to you?

If this two-stage approach matches what you have in mind, I think LastHitOnly would describe the KDA policy more accurately than LongestPrefixLastOnly: KDA checkpoints do not need to exist consecutively. We want to select and lease the last complete checkpoint within the already-protected KV boundary.

staryxchen:
@yokinoshitayoki Agreed — the two-stage lookup is a good fit for K3 on the framework side: `batch_is_exist` on KV first, then select KDA inside that boundary, so unused KDA checkpoints are not retained. The pairing stays in the inference framework, which is where it belongs.

For the store, we only need to offer the `lease_select` policies callers compose. From this thread there are already four:

- `None` (default): current ProbeKey, pin nothing
- `LongestPrefix`: lease `keys[0..k]`
- `LongestPrefixLastOnly`: lease only `keys[k]`
- `LastHitOnly`: lease the last hit in the list (need not be consecutive)

That seems enough to start with; further policies can be added later if new lookup shapes show up.

Does that work as a first cut?

yokinoshitayoki:
@staryxchen That makes sense. However, ProbeKey / BatchProbeKey are meant to be lease-free. Would it be better to keep this PR’s current semantics and add LongestPrefix, LongestPrefixLastOnly, and LastHitOnly as optional policies for ExistKey / BatchExistKey?

staryxchen:
@yokinoshitayoki Agreed on the design: `ProbeKey` / `BatchProbeKey` should stay lease-free, as in this PR. Pinning policies belong on `ExistKey` / `BatchExistKey` — that is the API whose job is “exist and lease.” `None` stays Probe; `All` is today’s Exist; `LongestPrefix`, `LongestPrefixLastOnly`, and `LastHitOnly` are partial pin on the same lookup path.

The part we need to be careful about is compatibility. `coro_rpc` encodes arguments by method signature, so adding `lease_select` onto the existing `ExistKey` / `BatchExistKey` RPCs would break mixed-version rolling upgrades in both directions. A C++ default argument does not change that on the wire.

So we are aligned on putting the policies on ExistKey, but the follow-up should keep the current RPC signatures and introduce the new policies in a compatibility-safe way (for example a new RPC, with the C++/Python Exist APIs dispatching to it only when the caller opts into a non-default policy).

Happy to discuss that shape in a follow-up so this PR can stay focused on the lease-free APIs.

yokinoshitayoki:
@staryxchen Sorry, I overlooked the rolling upgrade compatibility issue. Thanks a lot for pointing out. We can revisit this in a follow-up and work out how to introduce the new policies safely.

yokinoshitayoki (approved):
This PR looks good to me on its own, thanks a lot for your contribution!

staryxchen (approved):
LGTM, thanks!

ykwd (approved):

PR #3818, opened 2026-09-01
[Store] Complete KV event publication

Misak2333:
Finish the master-side KV cache event publisher so external cache-aware
indexers can consume a correct availability stream.

* Carry replica availability per medium instead of collapsing an object's metadata into a single medium string, so a tier is never dropped from the stream and a partial eviction retracts only the tier that went away.
* Announce evictions that leave the object alive, and pick a single publisher per removal so a retraction is not emitted twice.
* Add a tenant-wide clear event, tied to the real outcome of a removal.
* Stamp the configured model context on every event instead of nil.
* Add publisher and master-service test coverage, and document the wire format and the subscriber contract.

Co-Authored-By: lolo-pop <lolopop.rong@gmail.com>

## Description
The KV event publisher was wired up but the availability stream it produced was not usable by a cache-aware router: an object living on two tiers was announced as one, evictions that left the object alive were silent, tenant-wide clears had no event type, and every event carried a nil model context. This PR completes the master-side publisher so an external indexer can track replica availability from the stream alone.
### 1. Per-medium replica availability

An object's metadata was collapsed into a single medium string, so a two-tier object was announced as living on one tier. Availability is now carried per medium: `PublishCommitted` announces every medium a commit made available, and `SyncObjectState` diffs the medium set captured before a metadata mutation against the set after it. This makes tier transitions expressible — an offload emits `removed(cpu)` + `stored(disk)`, which reads as a demotion rather than a disappearance.
### 2. Partial eviction is announced

Evicting one replica of a multi-tier object produced no event at all, so subscribers kept routing to a tier that no longer held the data. Eviction now reports the tier it reclaimed, and a single publisher is chosen per removal so a retraction is not emitted twice for the same `(object, medium)`.

### 3. Tenant-wide `cleared` event

Clearing a tenant emitted one `removed` per object, forcing subscribers to infer "the tenant is empty" from a burst of retractions. There is now a `cleared` event type (legacy alias `AllBlocksCleared`), tied to the real outcome: a tenant is announced as cleared only when it held objects and none remain, so a partial failure does not tell subscribers to drop a tenant that still has live data.

### 4. Model context on every event

`model_name`, `block_size`, `lora_name` and `dp_rank` were emitted as nil, so subscribers keying on model identity had nothing to key on. They are now stamped from the Master configuration on every event; one publisher serves one model /block-size / LoRA / DP context.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
  -R '^(kv_event_publisher_test|master_service_test|master_service_test_for_snapshot)$'
```

**Test results:**
- [x] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used


catyans:
issue: `cleared` is not synchronized with concurrent writes. Both `RemoveAll` overloads scan and release one shard at a time while holding only the shared `snapshot_mutex_`. A Put can therefore commit into an already-scanned shard and enqueue `stored`; `RemoveAll` then publishes `cleared` from its stale `saw/skipped` state, so subscribers observe `stored -> cleared` even though the new object is live.

Please serialize tenant writes across the scan and publication, or use an epoch/barrier that prevents a clear from superseding a newer commit. A deterministic regression that pauses `RemoveAll` after the target shard, commits a new key there, and then resumes would cover the ordering. The current code-format check and CI Gate are also failing.

catyans:
Follow-up on `610fb7e` after the force-push: the race is still present. Both `RemoveAll` overloads hold only a shared `snapshot_mutex_` while locking shards one at a time; `PutStart`/`PutEnd` can therefore commit to an already-scanned shard before `PublishKvCleared()` runs. The publisher rewrite adds no epoch or barrier to prevent the resulting live `stored -> cleared` ordering.

@Misak2333, please serialize writes across scan-and-publish or add a tenant epoch/barrier, plus the deterministic concurrent regression described above. The code-format check is also currently failing.

Misak2333:
> Follow-up on `610fb7e` after the force-push: the race is still present. Both `RemoveAll` overloads hold only a shared `snapshot_mutex_` while locking shards one at a time; `PutStart`/`PutEnd` can therefore commit to an already-scanned shard before `PublishKvCleared()` runs. The publisher rewrite adds no epoch or barrier to prevent the resulting live `stored -> cleared` ordering.
> 
> @Misak2333, please serialize writes across scan-and-publish or add a tenant epoch/barrier, plus the deterministic concurrent regression described above. The code-format check is also currently failing.

Get it.  I am planning fix this case by adding a barrier in removeall function, therefore it can avoid send clear-event after new kv-cache meta creation.

New commit just fix this bug and added the deterministic regression ut test. @catyans  
I add a dedicated kv_tenant_epoch_mutex_ guards a fixed 1024-slot array of per-tenant write epochs. On the commit side, every store event bumps its tenant's epoch value under that lock, before the stored is enqueued, then releases it. On the RemoveAll side, the scan takes a baseline snapshot of the array before it starts, and at publication time does not read-then-publish — it re-checks and publishes cleared event while the epoch value is exactly same as the snapshot value.

catyans (approved):
Thanks, this addresses the ordering race. LGTM. Commits bump the tenant epoch before enqueueing `stored`, while `cleared` rechecks and enqueues under the same lock, so a concurrent commit either suppresses the clear or follows an already-enqueued clear. Both RemoveAll overloads have deterministic regression coverage. CI has 20 checks passed and 1 queued.

ykwd (approved):

Issue #3670, opened 2026-08-25
[RFC]: Tenant-First Object-Level Metadata Locking


Aionw:
# [RFC][Store]: Tenant-First Object-Level Metadata Locking

## Summary

Move metadata ownership from global shards to tenants and move the mutation
boundary from one shard lock to one object lock.

```mermaid
flowchart LR
    subgraph CURRENT["Current: shard-level ownership"]
        S["Metadata shards: 1024"] --> L["One lock per shard"]
        L --> M["Mixed tenant objects and task maps"]
    end

    subgraph PROPOSED["Proposed: tenant-first ownership"]
        R["TenantRegistry: COW snapshots"] --> T["TenantState"]
        T --> Q["TenantQuotaHandle"]
        T --> G["GroupDirectory"]
        G --> I["ObjectRouteIndex: key to object handle"]
        G --> MG["Explicit GroupEntry: 1..N objects"]
        G --> SG["Implicit singleton group"]
        MG --> O["ObjectEntry: object lock and state"]
        SG --> O
        I --> O
    end

    M --> CHANGE["Change: invert ownership and narrow the mutation lock"]
    CHANGE --> R
```

| Concern | Current | Proposed |
| --- | --- | --- |
| Ownership | Global shards contain fragments of many tenants | `TenantState -> GroupDirectory -> ObjectEntry` |
| Mutation | One shard lock covers unrelated objects | One lock per `ObjectEntry` |
| Group | Separate object-to-group and member maps | One tenant-local `GroupDirectory` |
| Lookup | Shard routing is also the lifetime boundary | One flat object route pins the owning object/group |
| Eviction | Scan and mutate under broad shard locks | Handle census, group-aware commit, matched unlink |

The expected architectural win is isolation and a clear ownership model. The
simple locked route defined here is the correctness baseline. Epoch/COW route
optimization is specified separately in
[Group Directory Object-Route Read-Path Optimization](https://github.com/kvcache-ai/Mooncake/issues/3671).

## Scope

| Goals | Non-goals |
| --- | --- |
| Put `TenantState` at the outer semantic boundary | Lock-free object business state |
| Make `GroupDirectory` the only tenant-local object container | Changing the external `group_id` contract |
| Allow unrelated objects to mutate concurrently | Concurrent conflicting mutations of one object |
| Make erase/recreate and async work lifetime-safe | Dynamic route-shard resizing |
[...]

PR #3822, opened 2026-09-01
[Store] Tenant-first object-level metadata locking (flat group)

CAICAIIs:
## Description

Implements **object-level metadata locking** for the Mooncake Store (issue #3670, flat-group model from #3589).

Today a single coarse **shard lock** serializes all writes and eviction on a shard, so unrelated objects contend on one lock. This PR replaces it with a **per-object lock** (`ObjectEntry::mutex`) plus a **per-tenant object route**: an object is excluded only while it is itself being read, written, or evicted. Group semantics keep the routing decoupling of #3589 in a thinner form — a group is a membership table plus one shared `Lease` (`GroupIndex`), consulted only by eviction for all-or-none decisions; the read path never touches the group table.

New modules under `mooncake-store/include/tenant/`:

- `MetadataCatalog` — boundary between MasterService and the tenant module; owns the tenant registry and lifecycle rules.
- `TenantDirectory` — copy-on-write `TenantId` → `TenantCatalog` map; lock-free reads, atomic get-or-create.
- `TenantCatalog` — per-tenant aggregate: quota handle, `ObjectIndex`, `GroupIndex`.
- `ObjectIndex` — flat key → `shared_ptr<ObjectEntry>` route with identity-checked erase (`EraseIf`), publication generations for remove/recreate races, and the dynamic-replication lease table.
- `ObjectEntry` — per-object mutex guarding the metadata envelope and all per-key task state that previously lived in parallel shard maps.
- `ObjectMetadata` / `object_replica_list.h` — extracted metadata envelope (identity, replicas, lease, soft-pin, quota ledger).

Concurrency rules: route locks are never held together with object locks (strong handles bridge the gap), teardown is claimed once per entry (`is_torn_down`), and a route slot is erased only against the exact entry being torn down, so a concurrent remove + re-create cannot drop a live object.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [x] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Unit / regression suites** (all pass): `master_service_test` (69), `master_service_ha_test` (93), `tenant_store_test` (15), `tenant_directory_test` (4), `object_entry_test` (1), plus new `group_index_test`, `object_index_test`, `tenant_catalog_test`, `tenant_directory_test`.

**Performance A/B** (Debug builds, same machine and compiler, same workload). baseline = pre-PR master (`ca09c18d`, shard locks), this PR = `7960b555`. 2x4GiB segments, 1MiB values, fixed arrival, groups of 5 consecutive put keys share one group_id, eviction driven to the 0.8 high-watermark on both.

Metadata plane (client→master RPC, `master_synthetic_bench`), 60s per run:

| workload | ref | p50 (ms) | p99 (ms) | p999 (ms) | item_qps | rpc_fail |
|---|---|---:|---:|---:|---:|---:|
| medium | baseline | 1.608 | 58.68 | 104.10 | 2173.7 | 148 |
| medium | **this PR** | **1.595** | **8.87** | **10.79** | 2173.7 | 148 |
| high | baseline | 1.743 | 61.39 | 106.95 | 4347.5 | 390 |
| high | **this PR** | **1.503** | **8.63** | **10.95** | 4347.5 | 390 |

p99 -84.9% / -85.9%, p999 -89.6% / -89.8%; p50 -0.8% / -13.7%. `rpc_failure_events` align on both sides (148 vs 148, 390 vs 390) and item_qps are identical, so the comparison is same-load. The shard-lock baseline tail is also unstable across runs (medium p99 varied 58-79 ms between rounds); this PR stays in a 9-13 ms band.

**Isolation under churn.** One client writes 1MiB objects at full speed to keep eviction continuously active; a second client runs a mixed read/write workload against the same master. Victim p99: baseline 52.9 → 85.9 ms under churn; this PR 28.6 → 46.4 ms. Under churn, this PR's victim tail stays below even the baseline's uncontended p99 (46.4 vs 52.9 ms).

**2x load, 16 workers** (metadata plane, fixed arrival at ~8695 item/s on both): baseline p99 105.5 ms, this PR 12.4 ms; `rpc_failure_events` identical (876 both). This PR's p99 stays in the 9-13 ms band from 1x to 2x load.

**Memory cost.** Master RSS measured over a 90s linear small-object write (~1M objects): baseline ~0.8 KB/object, this PR ~1.2 KB/object (+~0.4 KB, about +50%). The increase is the per-object lock, generation and task state folded into `ObjectEntry`, plus the route key stored next to `ObjectMetadata::user_key` — roughly 4 GB of master RSS at 10M objects. This is the cost of the per-object locking model; compaction (single key storage, on-demand task allocation) is left as follow-up work.

End-to-end runs (`store_kv_bench`, real data movement through the transfer engine, eviction continuously active) show the same direction: throughput 1800 → 2788 MiB/s and p99 52.9 → 28.6 ms. Note the store bench is closed-loop, so those two runs are not same-load; the metadata-plane tables above are the strict comparison. Raw logs available on request.

The branch is rebased onto current `upstream/main`; tenant module headers live under `include/tenant/`.

## Checklist

- [x] For changes >500 LOC: this is derived from the tenant-first design RFC (#3670/#3589)

## AI Assistance Disclosure

- [ ] No AI tools were used
- [x] AI tools were used (see issue #3670; comments reviewed line-by-line)


mooncake-store/include/tenant/tenant_directory.h
```diff
@@ -0,0 +1,91 @@
...
+    // while a strong snapshot is held, so a concurrent COW publish cannot
+    // invalidate it; it must NOT call Upsert/Remove on this directory. Callers
+    // dereference `handle` (a strong owning handle) and lock the target's own
+    // mutex before mutating it.
+    template <typename Fn>
+    void Visit(Fn&& fn) const {
+        auto frame =
+            std::atomic_load_explicit(&snapshot_, std::memory_order_acquire);
+        for (auto& kv : *frame) {
+            std::forward<Fn>(fn)(kv.first, kv.second);
```
Aionw, line 107: Could we expose a `shared_ptr<const Frame>` snapshot instead? The current `Visit() const` allows a callback taking `Handle&` to replace entries in an already-published frame and bypass the COW invariant. A const frame would enforce immutability of the mapping, though not of the pointed-to tenant state. If we keep `Visit`, it should also iterate an immutable frame.

mooncake-store/src/master_service.cpp
```diff
@@ -1413,15 +1437,11 @@ tl::expected<TenantId, ErrorCode> MasterService::ResolveTenantIdForWriteLocked(
...
 
 bool MasterService::TenantHasObjects(const TenantId& tenant_id) const {
-    for (size_t i = 0; i < kNumShards; ++i) {
-        MetadataShardAccessorRO shard(this, i);
-        auto tenant_it = shard->tenants.find(tenant_id);
-        if (tenant_it != shard->tenants.end() &&
-            !tenant_it->second.metadata.empty()) {
-            return true;
-        }
+    auto handle = tenant_directory_.Lookup(tenant_id);
```
Aionw: Non-blocking cleanup: this helper now has only one production call site and performs a lookup followed by an object-count check. Could we inline it unless we intend to keep it as a tenant-lifecycle abstraction?

mooncake-store/src/master_service.cpp
```diff
@@ -1516,14 +1536,28 @@ void MasterService::RecomputeTenantEffectiveQuotas() {
...
+        }
+        tenant_directory_.Upsert(tenant_id, handle);
     }
-    return it->second;
+    return handle;
+}
+
+MasterService::TenantState& MasterService::GetOrCreateTenantState(
+    const TenantId& tenant_id) {
+    return *GetOrCreateTenantStateHandle(tenant_id);
```
Aionw: Could we remove the reference-returning overload and keep the `shared_ptr`-returning API? Callers need an owning handle because the directory can remove or replace a tenant concurrently. Returning a reference drops that lifetime protection. The reference-returning overload also appears to have no remaining callers.

mooncake-store/include/tenant/object_entry.h
```diff
@@ -0,0 +1,100 @@
...
+    // its enclosed lease/soft-pin state; a path that touches both holds this
+    // mutex and then takes the metadata lock (never the reverse).
+    mutable std::shared_mutex mutex;
+
+    // Metadata is NON-movable / NON-copyable and self-locking, so it is owned
+    // through a pointer here. A live routed object has metadata wired in; an
+    // entry that has not yet materialized metadata (teardown, or a module-level
+    // unit test exercising only task state) has a null metadata_.
+    ObjectMetadata* metadata() const { return metadata_.get(); }
+    bool has_metadata() const { return metadata_ != nullptr; }
```
Aionw: Why do published entries need to support missing metadata? Production construction paths appear to attach metadata before publishing the entry. Could we enforce that invariant instead of repeating `has_metadata()` checks throughout the service? Object liveness should be tracked separately: an entry can retain metadata after being removed from the index, so `has_metadata()` alone cannot validate a pinned entry.

mooncake-store/src/master_service.cpp
```diff
@@ -1645,49 +1698,29 @@ MasterService::ObjectOperationLock MasterService::AcquireObjectOperationLock(
...
-    std::vector<std::string> member_keys;
-    GroupDomainAccessorRO shard(this);
-    auto it = shard->groups.find(tenant_id.MakeScopedKey(group_id));
-    if (it != shard->groups.end()) {
-        member_keys.assign(it->second.member_keys.begin(),
-                           it->second.member_keys.end());
-    }
-    return member_keys;
+    const TenantState& tenant_state, const std::string& group_id) const {
+    return tenant_state.object_route.Members(group_id);
```
Aionw: Could we remove this forwarding helper and call `Members()` directly? It no longer adds synchronization, validation, or business semantics. The nearby `RegisterGroupMember()` also appears to have no remaining callers.

mooncake-store/src/master_service.cpp
```diff
@@ -2334,12 +2401,14 @@ MasterService::EraseMetadata(
             }
             break;
     }
-    auto next = tenant_state.metadata.erase(it);
-    if (had_completed_disk && shard) {
-        shard->OnDiskReplicaRemoved(had_completed_disk);
+    // Release the object lock BEFORE the route mutation, so the route erase
+    // (EraseObject, route_lock_ unique) never runs under an entry lock.
+    object_lock.unlock();
+    tenant_state.EraseObject(key);
```
Aionw: Could we consolidate the `EraseMetadata` forwarding overloads using default arguments or an options struct, while preserving the distinction between deleting the current object by key and deleting a previously pinned entry? More importantly, this entry-based implementation erases by key after releasing the object lock, without checking that the index still points to the same entry. A stale entry must not remove a replacement object with the same key. The core deletion operation should validate instance identity and ensure teardown happens only once.

mooncake-store/src/master_service.cpp
```diff
@@ -2365,120 +2434,76 @@ void MasterService::ReleaseLocalDiskUsage(
...
-                }
+    tenant_directory_.Visit([&](const TenantId& tenant_id,
+                                const std::shared_ptr<TenantState>& handle) {
+        auto& tenant_state = *handle;
+        // Collect handles, then lock each per-object (there is no tenant
+        // level lock).
+        std::vector<std::shared_ptr<mooncake::tenant::ObjectEntry>> objs;
+        tenant_state.VisitObjects(
+            [&](const std::shared_ptr<mooncake::tenant::ObjectEntry>& entry) {
+                objs.push_back(entry);
```
Aionw: `VisitObjects()` already collects strong handles and releases the route lock before invoking the callback. This caller then copies those handles into another vector just to iterate them. Could we process entries directly in the callback, or expose a `SnapshotObjects()` API that returns the collected vector? The first collection protects lifetime without holding the route lock; the second seems unnecessary here. This is separate from the two-pass deadline aggregation and lease-wiring logic, which should retain its semantics.

mooncake-store/src/master_service.cpp
```diff
@@ -2365,120 +2434,76 @@ void MasterService::ReleaseLocalDiskUsage(
...
-                        return r.is_local_disk_replica() && r.is_completed();
-                    })) {
-                    shard->disk_object_count++;
-                }
-            }
-        }
-    }
+    // Every object already lives in its own tenant's container (routing is
+    // decoupled from groups), so re-routing objects across shards is a no-op.
+    return;
```
Aionw: Can we remove this no-op, its declaration, and its call sites? Deserialization already routes restored objects into their tenant's container regardless of the legacy shard index. Compatibility tests should exercise restoration of legacy snapshots rather than invocation of an empty migration function.

mooncake-store/src/master_service.cpp
```diff
@@ -1516,14 +1536,28 @@ void MasterService::RecomputeTenantEffectiveQuotas() {
...
+std::shared_ptr<MasterService::TenantState>
+MasterService::GetOrCreateTenantStateHandle(const TenantId& tenant_id) {
+    auto handle = tenant_directory_.Lookup(tenant_id);
+    if (!handle) {
+        handle = std::make_shared<TenantState>();
+        if (enable_multi_tenants_) {
+            handle->quota_account =
+                tenant_quota_table_.GetOrCreateTenantHandle(tenant_id);
+        }
+        tenant_directory_.Upsert(tenant_id, handle);
```
Aionw: The lookup and insertion are separate operations. Two concurrent requests for different keys in the same absent tenant can each create a `TenantState`, and the second `Upsert()` replaces the first. Both callers retain valid handles, but only one state remains reachable through the directory. The per-object striped operation locks do not serialize all keys within a tenant. Could the directory provide an atomic get-or-create operation that returns the winning handle to every caller, with a regression test covering concurrent first writes to distinct keys?

mooncake-store/include/master_service.h
```diff
@@ -2433,60 +1760,91 @@ class MasterService {
...
+                lock_ = std::unique_lock<std::shared_mutex>(entry_->mutex);
+            }
         }
 
         void MaybeEraseEmptyTenant() NO_THREAD_SAFETY_ANALYSIS {
             if (tenant_state_ == nullptr || !tenant_state_->Empty()) {
                 return;
             }
-            shard_guard_->tenants.erase(tenant_it_);
+            service_->tenant_directory_.Remove(object_id_.tenant_id);
```
Aionw: The emptiness check and directory removal are not atomic with object insertion. Another request can populate this tenant after `Empty()` returns true, and this path will still remove it from the directory. A strong handle prevents destruction but does not preserve reachability. Could we avoid eager empty-tenant removal, or introduce a retirement protocol coordinated with new writes? Checking handle identity alone would not close the race where the same tenant instance becomes non-empty. Other `Empty()`/`Remove()` call sites need the same treatment.

Aionw (commented):
The main concern is that the move from shard-level locking to per-object locking needs an explicit lifecycle protocol, not just finer-grained locks and strong handles. A handle keeping a tenant/object alive does not mean that it is still the current instance reachable through the directory/index. The inline comments on get-or-create, empty-tenant removal, and entry-based deletion address this distinction; these correctness concerns should take priority over the API cleanup suggestions.

### Module boundaries and directory layout

The directory layout seems to follow runtime ownership rather than responsibility. Objects belong to a tenant, but object metadata, object locking, and replication/promotion state are not inherently tenant-management concerns. Could we clarify these boundaries before deciding where the extracted types should live?

- `tenant/object_entry.h` would fit more naturally alongside `object_metadata.h` in an object/metadata module.
- `tenant/object_entry_types.h` contains replication, offloading, and promotion state records. A responsibility-oriented name such as `object_runtime_state.h` would describe its purpose better than its consumer.
- `TenantState::object_route` is actually a `TenantStore` managing object lookup, group membership, and dynamic-replication leases. Could we distinguish the object index from the tenant-level aggregate, with group and lease management exposed through their own interfaces? A per-tenant object index is reasonable; the issue is the mismatch between the name and its responsibilities.
- The actual tenant-level aggregate, `TenantState`, is still nested inside `MasterService`, while object-level types have moved into `tenant/`. Could we make the tenant aggregate an explicit module boundary and keep `MasterService` focused on orchestration?

I would avoid a purely mechanical file move here: clarify responsibilities and dependency direction first, then let the directory layout reflect those decisions. Unrelated quota-file reorganization can remain a follow-up.

Aionw:
Could we converge this change toward a concrete `MetadataCatalog` structure, rather than exposing `TenantDirectory`, `TenantStore`, and raw object entries directly to `MasterService`?

The target I have in mind is **one authoritative object record, a primary lookup index, and an optional eviction index**:

```text
MetadataCatalog
└── TenantId → TenantCatalog
    ├── ObjectIndex
    │   └── hash map: ObjectKey → shared_ptr<ObjectEntry>
    │
    ├── EvictionIndex [OPTIONAL]
    │   └── e.g. radix tree: EvictionKey → ObjectRef{key, generation}
    │
    └── GroupIndex
        └── GroupId → {shared lease, member references}

ObjectEntry
├── metadata / replicas
├── per-object task state
├── lifecycle state / generation
└── object mutex
```

Point operations would use the hash map. An **optional** `EvictionIndex` could support a different traversal structure, such as a radix tree, without changing the point-access representation. It would reference authoritative entries rather than duplicate their metadata. When disabled, eviction could fall back to scanning the primary index through the Catalog API.

**Implementing a radix tree is not a requirement for this PR.** The goal is to establish an architectural boundary that allows a specialized eviction index to be added later. Its key layout and mutable-priority update semantics would need to be defined separately.

The Catalog should own the lifecycle and consistency rules:

- Atomic tenant get-or-create and safe reclamation.
- Complete initialization before object publication.
- Coordinated updates to lookup, group membership, and the eviction index **when enabled**.
- Retirement and generation checks so stale handles or eviction candidates cannot act on replacement objects.
- RAII read/write handles, keeping raw entry access and mutexes private.

For example:

```cpp
auto object = catalog.GetForUpdate(object_id);
if (!object) {
    return ErrorCode::OBJECT_NOT_FOUND;
}
object->ReplaceReplicas(std::move(replicas));
```

Eviction would request candidates through a Catalog API, independently of whether they come from the optional index or a primary-index scan. Before execution, each candidate must be revalidated under its object lock, including generation, lease, pin, and replica eligibility.

This does **not** imply one global mutex. The Catalog can retain structural locks and per-object locks internally, while `MasterService` owns business policy, allocation, RPCs, and HA orchestration.

The current tenant-first hierarchy looks reusable for this. The intended direction is to centralize index consistency and object lifecycle guarantees, rather than requiring every caller to reconstruct them from individual map operations and manually acquired locks.

mooncake-store/include/tenant/group_index.h
```diff
@@ -0,0 +1,89 @@
...
+#include <shared_mutex>
+#include <string>
+#include <unordered_map>
+#include <unordered_set>
+#include <vector>
+
+#include "lease.h"
+
+namespace mooncake {
+namespace tenant {
```
Aionw: It's better here to rename `tenant` into `metadata` or index.

mooncake-store/include/tenant/group_index.h
```diff
@@ -0,0 +1,89 @@
...
+
+// GroupIndex: group_id -> {shared Lease, member keys}. A group is not a
+// container of objects: it is a thin membership table plus a single shared
+// Lease. The lease is the all-or-none unit consulted by eviction (the
+// per-member ObjectMetadata entry points at it); the read path extends that
+// shared lease on a member hit without touching this index.
+class GroupIndex {
```
Aionw, line 22: Are there any benchmarks for this struct? In the common case, we add multiple objects to a group, and concurrent modifications — typically inserting different keys — end up contending on the same mutex, which isn't great for performance.

mooncake-store/include/tenant/metadata_catalog.h
```diff
@@ -0,0 +1,67 @@
...
+namespace mooncake {
+namespace tenant {
+
+class MetadataCatalog {
+   public:
+    MetadataCatalog() = default;
+
+    // Atomic tenant get-or-create: concurrent first writers for the same
+    // tenant all observe the one winning TenantCatalog.
+    template <typename Factory>
```
Aionw: Unnecessary template type `Factory`, Tenant value here is fixed into `TenantCatalog`, we can save factory as class member.

mooncake-store/include/tenant/metadata_catalog.h
```diff
@@ -0,0 +1,67 @@
...
+    std::shared_ptr<TenantCatalog> GetOrCreateTenant(const TenantId& tenant_id,
```
Aionw: Could we bind the tenant factory once in the `MetadataCatalog` constructor and store it privately, so `GetOrCreateTenant()` only takes `tenant_id`?

The result type is fixed (`std::shared_ptr<TenantCatalog>`), and we do not appear to need a different creation/initialization policy for each call. A constructor-injected factory taking `const TenantId&` would make that policy consistent across the catalog without exposing a templated factory parameter on every call.

This should preserve lazy creation inside the atomic get-or-create path, including quota initialization before publishing the tenant.

mooncake-store/include/tenant/object_entry.h
```diff
@@ -0,0 +1,97 @@
...
+    // reverse.
+    std::unique_lock<std::shared_mutex> LockUnique() const {
+        return std::unique_lock<std::shared_mutex>(mutex);
+    }
+    std::shared_lock<std::shared_mutex> LockShared() const {
+        return std::shared_lock<std::shared_mutex>(mutex);
+    }
+    // Non-blocking probe: true when the mutex was free and is now locked for
+    // this caller (released again immediately on return).
+    bool TryLockUnique() const {
```
Aionw: Could we return the std::unique_lock instead of bool? The temporary lock is destroyed at the end of the return expression, releasing any acquired lock before the caller resumes.

 Suggested implementation:

 ```cpp
   [[nodiscard]]
   std::unique_lock<std::shared_mutex> TryLockUnique() const {
       return std::unique_lock<std::shared_mutex>(mutex, std::try_to_lock);
   }
 ```

 The caller can then retain the lock for the entire protected operation:

 ```cpp
   if (auto lock = TryLockUnique(); lock.owns_lock()) {
       // Access protected data while holding the exclusive lock.
   } // Automatically unlocks when the lock goes out of scope.
 ```

 This preserves RAII-based unlocking without requiring manual release.

mooncake-store/include/tenant/object_index.h
```diff
@@ -0,0 +1,239 @@
...
+            if (it->second.key == key) {
+                it = dynamic_replication_leases.erase(it);
+            } else {
+                ++it;
+            }
+        }
+    }
+
+    // Remove every lease that has expired (expire_at_ms_epoch < now_ms).
+    void EraseExpiredDynamicReplicationLeases(int64_t now_ms) {
```
Aionw: Prefer using std::chrono::time_point here.

mooncake-store/include/master_service.h
```diff
@@ -56,6 +60,8 @@
 
 namespace mooncake {
 
+using TenantCatalog = tenant::TenantCatalog;
```
Aionw: Unnecessary alias.

mooncake-store/src/master_service.cpp
```diff
@@ -1717,128 +1706,86 @@ MasterService::ObjectOperationLock MasterService::AcquireObjectOperationLock(
...
-    // group_id) and read WITHOUT a metadata shard lock.
+    // All members of a tenant live in a single tenant container, so look up
+    // that tenant and read its group membership from its own
+    // object_index.
+    auto tenant_handle = catalog_.Lookup(tenant_id);
+    if (!tenant_handle) {
+        return result;
+    }
+    TenantCatalog& tenant_accessor = *tenant_handle;
+    auto& tenant_state = *tenant_handle;
```
Aionw, line 1728: Duplicate value

mooncake-store/src/master_service.cpp
```diff
@@ -1717,128 +1706,86 @@ MasterService::ObjectOperationLock MasterService::AcquireObjectOperationLock(
...
+
     std::vector<std::string> member_keys =
-        GetGroupMemberKeys(tenant_id, group_id);
+        tenant_state.group_index.Members(group_id);
```
Aionw: Why not make this a public function of tenant state, so you do not expose internal state of tenant_state and safe to control boundary.

mooncake-store/include/tenant/object_index.h
```diff
@@ -0,0 +1,239 @@
...
+class ObjectIndex {
+   public:
+    ObjectIndex() = default;
+    ~ObjectIndex() = default;
+
+    // The object route is a flat map key -> strong ObjectEntry handle. Read
+    // pins the entry under the shared route lock (fast), releases it, then
+    // takes the per-object lock; the strong handle keeps the entry alive across
+    // that handoff.
+    std::shared_ptr<ObjectEntry> Pin(const std::string& key) const {
```
Aionw: Prefer using Get, using `Pin` here is confused.

mooncake-store/include/master_config.h
```diff
@@ -939,6 +940,10 @@ class MasterServiceConfigBuilder {
     std::string cxl_path_ = DEFAULT_CXL_PATH;
     size_t cxl_size_ = DEFAULT_CXL_SIZE;
     bool enable_cxl_ = false;
+    // Injected hook invoked when PutStart enters the snapshot section (after
```
Aionw: Do not pu it into config, since this member is only used for test. It could be misused.

mooncake-store/src/master_service.cpp
```diff
@@ -1996,20 +1945,19 @@ void MasterService::FinalizeRemovedReplicasAfterDurable(
...
-    if (tenant_it == shard->tenants.end()) {
+    auto tenant_handle = catalog_.Lookup(tenant_id);
+    if (!tenant_handle) {
         return;
     }
-    auto& tenant_state = tenant_it->second;
-    auto metadata_it = tenant_state.metadata.find(durable_entry.object_key);
-    if (metadata_it == tenant_state.metadata.end()) {
+    TenantCatalog& tenant_accessor = *tenant_handle;
+    auto& tenant_state = *tenant_handle;
```
Aionw, line 1955: Same as above. Duplicate

mooncake-store/src/master_service.cpp
```diff
@@ -2296,39 +2234,67 @@ tl::expected<void, ErrorCode> MasterService::PersistStaleHandleCleanupForHA(
...
+    MasterMetricManager::instance().inc_promotion_cancelled();
 }
 
-std::unordered_map<std::string, MasterService::ObjectMetadata>::iterator
-MasterService::EraseMetadata(
-    TenantState& tenant_state,
-    std::unordered_map<std::string, ObjectMetadata>::iterator it,
-    const TenantId& tenant_id, QuotaEraseMode quota_mode) {
-    return EraseMetadata(tenant_state, it, tenant_id, quota_mode, nullptr);
+void MasterService::EraseMetadata(TenantCatalog& tenant_state,
```
Aionw: Keep one EraseMetadata here is enough

Aionw (commented):
Reviewed the new per-object locking model on `f989d484`. I verified the 8 findings below independently: each one is deterministically reproducible (held-entry-lock / unlock-gate probes) and the existing tests still pass, so none of them is covered today. All of them are regressions relative to the previous shard-lock behavior.

Scope by configuration:
- Default path (no feature flags needed): `master_service.cpp:9882`, `:5435`, `:7327`, `master_service.h:1617`, `:1771`
- Requires HA + oplog: `master_service.cpp:1954`; requires offload/local-disk: `master_service.h:1650`; requires groups: `tenant_catalog.h:38`

Beyond these 8, a spot check of the ~30 `Pin()` sites in `master_service.cpp` shows several more entry-state accesses outside the entry lock (e.g. `offloading_task`, `dynamic_replication_pending`, `promotion_task` paths). It would be worth stating the rule explicitly — every access to entry state (metadata, tasks, flags) requires `entry->mutex` — and auditing against it, since `NO_THREAD_SAFETY_ANALYSIS` currently disables compiler help.

The repro tests are deterministic; happy to turn them into a test PR if useful.

mooncake-store/src/master_service.cpp
```diff
@@ -10105,315 +9763,297 @@ void MasterService::EvictionThreadFunc() {
...
+                auto settle_result =
+                    SettlePrimaryWriteQuotaIfReady(tenant_state, metadata);
+                if (settle_result) {
+                    entry->is_processing = false;
+                }
             }
         }
+    }
+    for (const auto& key : erase_keys) {
+        EraseMetadata(tenant_state, key, tenant_id, QuotaEraseMode::kFull,
```
Aionw: **[P1] Deferred cleanup re-resolves the key and can erase a replacement object.**

`erase_keys` is collected while each entry's lock is held, but this pass runs after all those locks have been released and re-resolves the key through the key-based `EraseMetadata`. If the invalidated entry is erased and a new object is published under the same key in that window, this call tears down the *replacement*; the `EraseObjectIf(key, entry.get())` inside `EraseMetadata` then compares the route against that same replacement, so it does not protect it either.

Reproduced deterministically: the reaper drops the last replica of key K and queues K, then blocks on another entry's lock; a PutStart/PutEnd for K succeeds; when the reaper resumes the new object is gone, although both Put calls returned OK.

Fix: keep the pinned `shared_ptr<ObjectEntry>` handles and call the entry-based overload (the one documented for "the caller may have extra state wired to it"). The same pattern applies to `replication_erase_keys` below and to `keys_to_erase` in `ClearStaleHandles`.

mooncake-store/src/master_service.cpp
```diff
@@ -5693,52 +5422,50 @@ auto MasterService::UpsertStart(const UUID& client_id, const std::string& key,
...
                 ChargeTenantQuota(GetBoundTenantQuotaHandle(tenant_state), 0);
             if (!admission_result) {
-                if (tenant_state.Empty()) {
-                    shard->tenants.erase(object_id.tenant_id);
-                }
                 return tl::make_unexpected(admission_result.error());
             }
 
-            auto it = tenant_state.metadata.find(key);
+            auto object_entry = tenant_state.Pin(key);
```
Aionw: **[P1] `UpsertStart` mutates entry state without holding `entry->mutex`.**

`tenant_state.Pin(key)` only keeps the entry alive; everything after it (stale-handle cleanup, `PopReplicas`/`mark_processing`, `client_id`/`put_start_time`, soft-pin bookkeeping, quota settlement, and the Case B/C rewrite) runs without the per-object lock. `object_operation_lock` only excludes the other `PutStart`/`UpsertStart` on the same key, and `snapshot_mutex_` is taken shared, so a concurrent `GetReplicaList`/`PutEnd`/`Remove`/copy that does hold the lock is not excluded at all.

Reproduced deterministically: while another thread holds `entry->LockUnique()`, `UpsertStart` still modifies metadata and returns success.

Fix: hold the entry lock for the whole read-modify-write region and use a teardown variant that assumes the lock is already held (the entry lock and the route lock must still never be held together).

mooncake-store/src/master_service.cpp
```diff
@@ -1996,20 +1945,19 @@ void MasterService::FinalizeRemovedReplicasAfterDurable(
...
+    auto tenant_handle = catalog_.Lookup(tenant_id);
+    if (!tenant_handle) {
         return;
     }
-    auto& tenant_state = tenant_it->second;
-    auto metadata_it = tenant_state.metadata.find(durable_entry.object_key);
-    if (metadata_it == tenant_state.metadata.end()) {
+    TenantCatalog& tenant_accessor = *tenant_handle;
+    auto& tenant_state = *tenant_handle;
+    auto object_entry = tenant_state.Pin(durable_entry.object_key);
```
Aionw: **[P1] Durable finalizer mutates replica state without the entry lock.**

This pins the entry and then runs `PopReplicasWithCacheTotalAccounting`, `RecordDynamicReplicaRemoval`, quota release / `is_processing` update and `CancelPromotionTaskForRemovedReplicas` under only the shared `snapshot_mutex_`. The previous code held a `MetadataShardAccessorRW` here, which excluded everything else on the shard; now a request that holds the entry lock concurrently (Get/PutEnd) races with these vector and ledger mutations.

Reproduced deterministically: with the entry lock held by another thread, the finalizer still removes a REMOVED replica and returns.

Fix: take `entry->LockUnique()` before touching metadata and use a locked teardown helper.

mooncake-store/src/master_service.cpp
```diff
@@ -7582,54 +7301,37 @@ auto MasterService::BatchRemove(const std::vector<std::string>& keys,
...
 
-            if (it == tenant_state.metadata.end()) {
+            if (!object_entry) {
                 VLOG(1) << "key=" << key << ", error=object_not_found";
                 results[original_idx] =
                     tl::make_unexpected(ErrorCode::OBJECT_NOT_FOUND);
                 continue;
             }
 
+            ObjectMetadata& metadata = object_entry->metadata();
```
Aionw, line 7424: **[P1] `BatchRemove` checks and mutates under no entry lock.**

After `Pin`, the stale-handle plan, `IsValid`, `IsLeaseExpired`, `AllReplicas` and (in the HA branch) `mark_removed()` all run without `entry->mutex`, and the final key-based `EraseMetadata` cannot protect the earlier checks. A concurrent writer can change the state between the check and the erase.

Reproduced deterministically: with the entry lock held, `BatchRemove` still reads a PROCESSING replica and returns `REPLICA_IS_NOT_READY` — i.e. the lock is not consulted at all.

Fix: same as above. `RemoveByRegex` (the loop that pins each key) and the second scan of `BatchEvict` have the same pattern and should be fixed together.

mooncake-store/include/master_service.h
```diff
@@ -2377,80 +1558,79 @@ class MasterService {
...
             NO_THREAD_SAFETY_ANALYSIS {
-            service_->EraseMetadata(*tenant_state_, it_, object_id_.tenant_id,
-                                    QuotaEraseMode::kFull, &shard_guard_,
-                                    previous_media_hint);
-            it_ = tenant_state_->metadata.end();
-            MaybeEraseEmptyTenant();
+            // Release the per-object lock first; EraseMetadata re-locks it for
+            // teardown and releases it before the route mutation. The two locks
+            // are never held together.
+            lock_ = std::unique_lock<std::shared_mutex>();
```
Aionw: **[P1] `Erase()` drops the entry lock before teardown, so the caller's checks are not atomic with the delete.**

`Remove` validates lease / all-completed / no-replication-task while the accessor holds the lock, then `Erase()` releases it and lets `EraseMetadata` re-acquire it without redoing any of those checks. A request that arrives in that window (or is already blocked on the entry lock) can be granted a fresh lease and then see the object deleted; `CopyStart` can likewise start a task that teardown then cleans up.

Reproduced deterministically with a pthread unlock gate: pause `Remove` exactly at this unlock; a concurrent `GetReplicaList` succeeds and refreshes the lease; non-forced `Remove` still deletes the object.

Fix: claim teardown (and do the cleanup) inside the original lock region; release the entry lock only immediately before the route mutation, keeping the identity-checked route erase.

mooncake-store/include/master_service.h
```diff
@@ -2559,61 +1757,47 @@ class MasterService {
...
+              lock_(entry_ != nullptr ? entry_->LockShared()
+                                      : std::shared_lock<std::shared_mutex>()) {
+        }
 
         // Check if metadata exists
         bool Exists() const NO_THREAD_SAFETY_ANALYSIS {
-            return tenant_state_ != nullptr &&
-                   it_ != tenant_state_->metadata.end() &&
-                   it_->second.IsValid();
+            return tenant_state_ != nullptr && entry_ != nullptr &&
```
Aionw, line 1814: **[P1] A torn-down entry is still readable: `is_torn_down` is not checked on the read path.**

`EraseMetadata` sets `is_torn_down` and intentionally keeps metadata wired, then releases the entry lock and only afterwards erases the route. `Pin()` does not filter torn-down entries (only `IsCurrent()` does, and read paths don't use it), so `Exists()` / `GetReplicaList` / `ExistKey` can see `IsValid() == true` on an object whose quota/disk accounting has already been released and whose KV remove event was already published; they will also grant a new lease on it.

Reproduced deterministically: take the exact handoff state (torn-down, route still published) — `ExistKey` and `GetReplicaList` both succeed. Note that a reader which pins before the teardown and waits on the entry lock reaches this state in normal operation, so it is not a tiny-window race.

Fix: include `!entry->is_torn_down` in the accessor `Exists()` checks, and audit the manual `Pin()` paths (batch/regex/scan) for the same condition. A strong handle keeping the entry alive must not be treated as "the object still exists".

mooncake-store/include/master_service.h
```diff
@@ -2461,61 +1641,78 @@ class MasterService {
...
-                std::forward_as_tuple(object_id_.user_key),
-                std::forward_as_tuple(
+            EnsureTenantCatalog();
+            auto entry = std::make_shared<mooncake::tenant::ObjectEntry>(
+                std::make_unique<ObjectMetadata>(
                     client_id, now, total_length, std::move(replicas),
                     std::nullopt, enable_hard_pin, data_type, group_id,
                     object_id_.tenant_id, object_id_.user_key));
-            it_ = result.first;
+            if (!tenant_state_->InsertObject(object_id_.user_key, entry)) {
```
Aionw, line 1685: **[P1] `Create()` publishes an empty entry before taking the entry lock.**

`AddReplicaForRetainedClient` calls `Create(..., std::vector<Replica>{})`; `InsertObject` places the entry in the route first and the entry lock is only taken at line 1671. In that window a normal RW request (e.g. `Remove`, or any RW accessor whose constructor auto-cleans invalid metadata) sees the published empty envelope as invalid and tears it down; `AddReplica` then attaches its disk replica to an already-unpublished entry and returns `true`, while the key no longer exists.

Reproduced deterministically with a route-lock gate: publish, then `Remove` (auto-cleanup) removes the key, then `AddReplica` returns success and `ContainsObject(key)` is false.

Fix: have the entry fully initialized (replica included) before publishing, or at least take the entry lock before `InsertObject`. Note `AllocateAndInsertMetadata` writes the quota ledger, pending soft-pin and `is_processing` after publication without the lock as well, and should be fixed the same way.

mooncake-store/include/tenant/tenant_catalog.h
```diff
@@ -0,0 +1,176 @@
...
+    // Publication transaction: wire the group lease and register membership
+    // before publishing, rolling membership back if the publish fails — a
+    // concurrently-pinned entry never observes a half-wired grouped member.
+    bool InsertObject(std::string key, std::shared_ptr<ObjectEntry> entry) {
+        if (key != entry->key()) {
+            return false;
+        }
+        const std::string group_id = entry->group_id();
+        if (!group_id.empty()) {
+            entry->metadata().SetLease(group_index.LeaseFor(group_id));
```
Aionw: **[P1] `LeaseFor()` + `AddMember()` are not one atomic group operation.**

Both calls take the `GroupIndex` lock separately, so the group can be destroyed and recreated in between (`RemoveMember` erases the group once its last member leaves, and `LeaseFor` then creates a *new* lease). The entry ends up registered in the new group while its metadata keeps the old lease: reads of that member extend the orphaned lease, and the group's shared TTL/eviction protection no longer covers all members. If no one recreates the group, `AddMember` returns false and the failed publish is reported as `OBJECT_ALREADY_EXISTS` (`AllocateAndInsertMetadata`), which is misleading for the client.

Reproduced deterministically: `LeaseFor` (L1) → last member removed (group erased) → `LeaseFor` recreates (L2) → both members are added, but the first still holds L1; reading L1 does not protect L2.

Fix: one `GroupIndex` operation that materializes the group, inserts the member and returns the matching lease under a single lock.

Aionw:
Could we retain a small, sparse promotion-candidate index in `TenantCatalog` as a minimal fix for this PR?

The previous `promotion_candidates` map provided both state storage and sparse enumeration. Moving the authoritative state into `ObjectEntry` is reasonable, but it should not require `RunPromotionCandidateRetry()` to snapshot and lock every object just to find active candidates. Even a single candidate waiting for backoff currently triggers that full scan.

A narrowly scoped approach would be:

- Keep `ObjectEntry::promotion_candidate` as the single source of truth.
- Add a catalog-owned index containing only candidate object references, without duplicating their task state. Use weak references or identity/generation-checked tokens.
- Maintain membership through the existing candidate creation/removal helpers, including object teardown and reload cleanup. Refreshing a candidate or changing its backoff would not require changing index membership.
- Have the retry loop enumerate this index instead of `SnapshotObjects()`. Release the index lock before acquiring object locks, then revalidate the current candidate state.

This restores scan cost proportional to the number of candidates rather than the total object population.

mooncake-store/src/master_service.cpp
```diff
@@ -2089,21 +2039,13 @@ void MasterService::FinalizeMetadataEraseAfterDurable(
...
+    auto tenant_handle = catalog_.Lookup(tenant_id);
+    if (!tenant_handle) {
         return;
     }
-    EraseMetadata(tenant_state, metadata_it, tenant_id, quota_mode, &shard);
-    if (tenant_state.Empty()) {
-        shard->tenants.erase(tenant_it);
-    }
+    auto& tenant_state = *tenant_handle;
+    EraseMetadata(tenant_state, tenant_state.Get(durable_entry.object_key),
```
Aionw: **[P1] Durable metadata-erase finalizer still erases by key, not by entry identity.**

This callback runs on the oplog writer's callback thread after durability. `ClearStaleHandles` queues it against a specific invalidated entry (`master_service.cpp:2694`), but here the key is resolved again at callback time and whatever entry is current gets erased. If the old entry is gone by then and a new object has been published under the same key (a retry Put after the stale-handle sweep is the natural flow), this silently deletes the replacement.

Reproduced deterministically: with a fresh valid entry under key K, invoking the finalizer with an old REMOVE entry for K erases it (`ContainsObject(K) == false`).

Fix: carry the entry identity (`ObjectEntry::generation()` captured at scheduling, or a pinned handle) and revalidate (`IsCurrent` / generation match) before erasing; only use the key-based fallback when the recorded entry is no longer the published one.

mooncake-store/src/master_service.cpp
```diff
@@ -10105,317 +9808,295 @@ void MasterService::EvictionThreadFunc() {
...
-                            ? [this, source_id,
-                               target_ids = std::move(target_ids),
-                               dynamic_lease_id, dynamic_version_epoch,
-                               ttl](const OpLogEntry& durable_entry) {
-                                  FinalizeExpiredReplicationTaskAfterDurable(
-                                      durable_entry, source_id, target_ids,
-                                      dynamic_lease_id, dynamic_version_epoch,
-                                      ttl);
+                            ? [this, ttl](const OpLogEntry& durable_entry) {
+                                  FinalizeExpiredProcessingReplicasAfterDurable(
```
Aionw, line 9933: **[P1] Expired-processing finalizer pops whatever entry the key resolves to now.**

This schedules `FinalizeExpiredProcessingReplicasAfterDurable` for the expired entry by key, but the callback re-resolves the key with a fresh accessor and pops *all* PROCESSING replicas of whatever entry it finds — it never checks that this is the same entry, nor even that its `put_start_time` matches the TTL that triggered the cleanup. If the old entry was preempted in the meantime (a `PutStart` after `put_start_release_timeout`, or an `UpsertStart`, both erase and re-create the key), the finalizer discards the new object's in-flight replicas and can erase it.

Reproduced deterministically: a fresh PROCESSING entry under the key plus the finalizer for an older entry -> `CountReplicas() == 0` and the key is gone.

Fix: pass the entry identity/generation (or a pinned handle) into the callback and skip when the current entry is not the expired one. The REMOVE variant two branches down needs the same treatment.

mooncake-store/src/master_service.cpp
```diff
@@ -3300,54 +3102,35 @@ std::vector<tl::expected<bool, ErrorCode>> MasterService::BatchExistKey(
...
-        if (tenant_it == shard->tenants.end()) {
-            for (const size_t i : key_indices) {
-                VLOG(1) << "key=" << keys[i]
-                        << ", tenant_id=" << normalized_tenant
-                        << ", info=object_not_found";
-                results[i] = false;
-            }
+        auto entry_lock = entry->LockShared();
+        const ObjectMetadata& metadata = entry->metadata();
+        if (!metadata.IsValid() || !HasReadableReplica(metadata)) {
```
Aionw, line 3144: **[P1] Torn-down entries are still served by the manual read paths.**

The accessors now check `!entry->is_torn_down`, but these paths open-code the check with `metadata.IsValid()` only, so they can serve an entry whose teardown already released quota/disk accounting and published the KV remove event — and `BatchExistKey` refreshes its lease a few lines below:

- `BatchExistKey` (`master_service.cpp:3125`)
- `GetReplicaListByRegex` (`:4060`)
- `GetReplicaListForAdmin` (`:4243`, `:4369`)
- `GetAllKeys` (`:3150`) lists torn-down keys (minor)

Reproduced deterministically: with `is_torn_down` set on a published entry, `ExistKey`/`GetReplicaList` are correctly rejected, but `BatchExistKey` returns true and `GetReplicaListByRegex` still returns the key.

Fix: include `!entry->is_torn_down` in these manual checks (or route them through the accessors), so every read path agrees that a claimed teardown is not readable.

mooncake-store/include/master_service.h
```diff
@@ -2461,61 +1668,81 @@ class MasterService {
...
+            // because no path takes the route lock and then an entry lock.
+            lock_ = entry->LockUnique();
+            if (!tenant_state_->InsertObject(object_id_.user_key, entry)) {
+                // A concurrent writer already inserted this key. Re-pin the
+                // existing entry and use it instead of the orphan we built.
+                // Drop the orphan's lock first, or the same thread would
+                // re-lock a mutex it already holds.
+                lock_ = std::unique_lock<std::shared_mutex>();
+                entry_.reset();
+                entry_ = tenant_state_->Get(object_id_.user_key);
```
Aionw, line 1692: **[P1] `Create()` can adopt a torn-down entry, and `AddReplica` then acknowledges a registration for a key that is already gone.**

`Exists()` returns false for a torn-down entry, so `Create()` builds and locks a new envelope. If the route still points at the torn-down entry at that moment, `InsertObject` fails and this fallback re-resolves the key and adopts whatever it finds — including an entry that has already been claimed for teardown but whose route slot has not been erased yet. `AddReplicaForRetainedClient` never re-checks `Exists()` after `Create()`, so it wires the disk replica into the torn-down entry and returns `true`; the teardown then erases the route.

Reproduced deterministically with a gate at the teardown's entry-unlock: `AddReplica` returns success while `ContainsObject(key) == false`.

Fix: if the fallback entry is torn down, retry the create (the route erase is imminent) or return a retryable error; and re-check `accessor.Exists()` after `Create()` before wiring.

mooncake-store/include/master_service.h
```diff
@@ -110,41 +114,26 @@ class MasterServiceSSDTest;
...
+ * 4. per-object ObjectEntry::mutex (via metadata::TenantCatalog::object_index
+ * Get)
  * 5. tenant_quota_recompute_mutex_
  * 6. ShardedTenantQuotaTable internal mutex or segment_mutex_
  * 7. soft_pin_deadline_index_ mutex
  *
- * Strict tenant admission and policy mutation paths that need both
- * tenant_quota_policy_mutex_ and snapshot_mutex_ must acquire the tenant
- * policy mutex first, then snapshot_mutex_.
+ * The per-object lock is never held together with the object route lock
```
Aionw: **[P2] The documented lock-order invariant is no longer true.**

This (newly added) comment says the per-object lock and the route lock are never held together, but `Create()` (`master_service.h:1678`) and `AllocateAndInsertMetadata` (`master_service.cpp:4643`) now take `entry->LockUnique()` *before* `InsertObject`, which acquires the route lock — i.e. entry -> route nesting. The call-site comment at `:1679` documents the weaker, accurate claim ("no path takes the route lock and then an entry lock"), and I could not find that order violated in the current tree, but the general invariant here and the same wording in `EvictGroupOrObject` (`:1200`) should be updated — otherwise future route -> entry code (or the HA route-lock gating tests) will rely on a rule the tree no longer follows.

Aionw (commented):
Re-verified on `3f530134`. Most of the previous review is fixed — I re-ran the deterministic probes plus the repo suites.

Fixed (probes now pass):
- `UpsertStart`, `FinalizeRemovedReplicasAfterDurable`, `BatchRemove` claim teardown under the entry lock (R2a/R2b/R2c).
- `Erase()` hands its held lock to `EraseMetadataLocked`; a Get arriving during the teardown handoff is no longer served (R3).
- The reaper / `ClearStaleHandles` / snapshot cleanup keep pinned handles instead of re-resolving keys (R1, synchronous paths).
- `Create` / `AllocateAndInsertMetadata` lock the envelope before publishing (R5 empty-envelope race).
- `GroupIndex::AddMember` materializes + registers + returns the lease atomically (R6).
- Repo suites: `master_service_test` 57/57, `master_service_group_test` 11/11, catalog tests 20/20.

Still open (probes fail):
1. The HA/oplog deferred finalizers are still key-addressed and act on whatever entry is current (`FinalizeMetadataEraseAfterDurable`, `FinalizeExpiredProcessingReplicasAfterDurable`).
2. Manual read paths (`BatchExistKey`, `GetReplicaListByRegex`, `GetReplicaListForAdmin`, `GetAllKeys`) do not check `is_torn_down`.
3. `Create()`'s insert-failure fallback can adopt a torn-down entry, so `AddReplica` can acknowledge a registration for a key that is already gone.
4. (doc) The new lock-order invariant comment is contradicted by the entry -> route nesting in `Create`/`AllocateAndInsertMetadata`.

Details in the inline comments. Happy to share the probe file if useful.

mooncake-store/src/master_service.cpp
```diff
@@ -7457,7 +7224,10 @@ long MasterService::RemoveAll(bool force) {
 long MasterService::RemoveAll(const TenantId& tenant_id, bool force) {
     long removed_count = 0;
     int64_t total_freed_size = 0;
-    std::shared_lock<std::shared_mutex> shared_lock(snapshot_mutex_);
+    // There is no per-tenant mutex; serialize RemoveAll with an exclusive
+    // snapshot lock so concurrent RemoveAll calls do not double-count/tear
+    // down.
+    std::unique_lock<std::shared_mutex> shared_lock(snapshot_mutex_);
```
Aionw: **[P2] Tenant-scoped RemoveAll now blocks metadata requests across every tenant.**

This changes the snapshot barrier from shared to exclusive and holds it for the entire tenant scan and teardown loop. Normal `GetReplicaList`, `ExistKey`, `PutStart`, and `PutEnd` requests also acquire `snapshot_mutex_`, so clearing a large tenant A prevents even requests for unrelated tenant B from proceeding until A's cleanup finishes. The baseline held the snapshot barrier shared and serialized mutations through the individual shard locks; this introduces a global pause rather than merely changing the object-lock granularity.

Avoiding duplicate teardown/counting is necessary, but could we keep the snapshot barrier shared and enforce that through the per-entry teardown claim and its return value, or separate appropriately scoped serialization? The global `RemoveAll(bool)` overload has the same new exclusive-lock pattern and should be considered alongside this one.

A regression test could pause tenant A's cleanup after it enters the scan and verify that a point read/write against tenant B can complete before A is resumed. This would explicitly protect the cross-tenant isolation that the object-level locking change is intended to improve.

CAICAIIs:
Benchmark numbers for the current head (18390d5e) vs the pre-tenant-first baseline (ca09c18d), measured with `master_synthetic_bench` (#3147) against a Release-built `mooncake_master`. Both binaries were built from the same tree and flags on an idle 256-core machine.

Workload: 8 segments x 400 GiB, 448 KiB values (the sglang-derived defaults), `group_size=5`, fixed arrival, 60 s per run. Medium is the benchmark's default per-segment rate (exist 6.34 / put 3.14 / get 2.02 qps), high is 2x. Two interleaved rounds per side; ~8.7k (medium) and ~17.4k (high) items/s.

| load | build | p50 (ms) | p99 (ms) | p999 (ms) | failed RPCs |
|---|---|---|---|---|---|
| medium | baseline | 0.732 / 0.766 | 8.106 / 7.911 | 14.249 / 15.883 | 0 / 2 |
| medium | this PR | 0.899 / 0.912 | 7.054 / 7.284 | 14.056 / 14.728 | 0 / 0 |
| high | baseline | 0.734 / 0.742 | 8.511 / 8.628 | 15.430 / 18.180 | 0 / 3 |
| high | this PR | 0.877 / 0.805 | 7.377 / 6.836 | 15.780 / 15.090 | 0 / 0 |

Each cell is round 1 / round 2.

What the numbers show:

- Tail latency improves: p99 is 10-17% lower and p999 4-8% lower, consistent across rounds and loads.
- Stability: zero failed RPCs in all PR runs; the baseline had 5 failures total.
- Median latency is ~14-21% higher, and per-request CPU is ~1.3-1.4x the baseline. We know where this comes from: the tenant-first metadata structures allocate more per request (string-keyed maps storing metadata by value, plus group-membership key copies). 78df7e56 tried a lookup-side fix (transparent hashing + initial reserve) and was reverted in 7bcc6c4d; we plan to revisit it, since it should recover most of the p50 gap while keeping the tail wins.
- The correctness fixes in this PR do not move the numbers: with and without them, measurements are the same within round-to-round noise.

stmatengss (changes requested):
There are a lot of code changes. Please split the extensive code changes into multiple smaller PRs.

CAICAIIs:
Benchmark numbers for the current head vs the pre-tenant-first baseline (ca09c18d), measured with `master_synthetic_bench` (#3147) against a Release-built `mooncake_master`. Both binaries were built from the same tree and flags on an idle 256-core machine.

Workload: 8 segments x 400 GiB, 448 KiB values (the sglang-derived defaults), `group_size=5`, fixed arrival, 60 s per run. Medium is the benchmark's default per-segment rate (exist 6.34 / put 3.14 / get 2.02 qps), high is 2x. Two interleaved rounds per side; ~8.7k (medium) and ~17.4k (high) items/s.

| load | build | p50 (ms) | p99 (ms) | p999 (ms) | failed RPCs |
|---|---|---|---|---|---|
| medium | baseline | 0.732 / 0.766 | 8.106 / 7.911 | 14.249 / 15.883 | 0 / 2 |
| medium | this PR | 0.899 / 0.912 | 7.054 / 7.284 | 14.056 / 14.728 | 0 / 0 |
| high | baseline | 0.734 / 0.742 | 8.511 / 8.628 | 15.430 / 18.180 | 0 / 3 |
| high | this PR | 0.877 / 0.805 | 7.377 / 6.836 | 15.780 / 15.090 | 0 / 0 |

Each cell is round 1 / round 2.

What the numbers show:

- Tail latency improves: p99 is 10-17% lower and p999 4-8% lower, consistent across rounds and loads.
- Stability: zero failed RPCs in all PR runs; the baseline had 5 failures total.
- Median latency is ~14-21% higher, and per-request CPU is ~1.3-1.4x the baseline.

**Update (after the upstream/main merge):**

Profiling on an idle machine showed the head executes 1.54x the instructions of the baseline with no elevated cache-miss or stall rates, so the gap is straight-line work, not memory stalls. The largest single contributor was `SnapshotAllocatorManager()` deep-copying the segment->allocator registry on every put/allocate call. It is now a shared copy-on-write snapshot (rebuilt only on mount/unmount) — 58996ff4, on top of the upstream merge. Measured effect vs the previously tabulated head, same harness: CPU per request -6% (medium) to -12% (high), p50 -3~4%, p99 ~-4%, across 8/8 interleaved runs. Remaining p50 gap tracks the per-request instruction volume of the new metadata structures and is follow-up work; the tail-latency and stability wins are unchanged.

mooncake-store/tests/master_service_test.cpp
```diff
@@ -731,59 +731,6 @@ TEST_F(MasterServiceTest, RemoveAllKeepsObjectWhenOpLogReservationFails) {
         << "the object survived, so the tenant was never emptied";
 }
 
-// RemoveAll holds one shard lock at a time, so a commit can land in a shard the
-// scan already passed. The scan's own bookkeeping cannot see it, and publishing
-// `cleared` would order it after that commit's `stored` — telling subscribers
-// to drop an object that is live. Pause the scan right after the shard the new
-// key belongs to, commit there, and the clear must be withheld.
-TEST_F(MasterServiceTest, ConcurrentCommitDuringScanSuppressesClear) {
```
rishabhsinha17, line 739: Deleting this and `TenantScopedRemoveAllSuppressesClearOnRace` removes the only coverage of the suppression branch in `PublishKvClearedIfEpochUnchanged`, and that branch is still live. Both `RemoveAll` overloads read their epoch baseline before the walk and decide per tenant afterwards. The race stays as reachable as it was under shards. Only the uncontended mirror test remains, and it pins the publish side, which leaves `kv_cleared_suppressed_by_epoch_` free to regress silently. Your shard hook is gone, but re-expressing this does not need a new hook on MasterService. Your own gate technique from `TenantRemoveAllDoesNotBlockOtherTenants` reproduces the race deterministically. Park a forced `RemoveAll` on a held entry lock, then commit a fresh key into the same tenant while the sweep waits on the gate. That key misses the walk's `SnapshotObjects` snapshot and its `PutEnd` bumps the epoch past the baseline. Release the gate, join, and expect suppressed 1, published 0, and the racer still readable. One more copy of that gate covers the tenant scoped overload.

mooncake-store/tests/master_service_test_fixture.h
```diff
@@ -229,20 +212,55 @@ class MasterServiceTest : public ::testing::Test {
...
-    // deterministically.
-    size_t ShardIndexForKey(MasterService& service, const std::string& key,
-                            const TenantId& tenant_id = TenantId::Default()) {
-        return service.getShardIndex(tenant_id, key);
+    // Regression: MetadataAccessorRW::Create() used to ignore InsertObject()'s
+    // return value. If a concurrent writer inserted the key first, Create()
+    // would bind to an orphan entry not in the route. It must re-resolve the
+    // existing entry instead. Lives on the fixture (a MasterService friend) so
+    // it can reach the private MetadataAccessorRW.
+    void AccessorCreateRePinsWinnerEntry(MasterService& service) {
```
rishabhsinha17, line 262: No TEST_F invokes this, so the orphan rebind regression it documents goes unpinned. A one line TEST_F in master_service_test.cpp calling it closes that.

mooncake-store/tests/master_service/dsl/scenario.cpp
```diff
@@ -1071,33 +1071,22 @@ MasterScenario& MasterScenario::When(ExpireAtAction action) {
...
-        metadata_it->second.lease_->SetDeadline(action.lease_timeout);
-        metadata_it->second.soft_pin_timeout = action.soft_pin_timeout;
-        return true;
-    };
-
-    const size_t routed = service_->getShardIndex(tenant, action.key);
-    if (update(routed)) {
+    // All of a tenant's keys live in one tenant container, so a single
+    // directory lookup finds the object (there is no per-shard probe).
+    auto tenant_handle = service_->catalog_.Lookup(tenant);
```
rishabhsinha17, line 1078: Faithful rewrite from where I sit as the DSL owner. Both miss paths keep the old failure text, the same two fields change, and the setters take the spinlock the old inline code took. Dropping the all shards probe is right, it existed only for placements the flat index cannot produce. I will take two follow-ups on my side once this merges, renaming `GroupOnDifferentShard` and a contract test pinning `ExpireAt`'s failure report. Nothing for you to do here.

rishabhsinha17 (commented):
Reviewed the test and DSL lane on 1c1835d1, wearing the MasterScenario DSL author hat (#3847, #3927). I read the full diff against merge base 6ae74b35 and ran the affected suites locally at the head, numbers at the end. Aionw's locking findings stand on their own and I did not re-verify them.

The DSL side is in good shape. This PR leaves `scenario.h` and `scenario_contract_test.cpp` untouched, so the grammar and its pins hold. Rewriting the `ExpireAt` engine preserved the verb's contract. Both miss paths keep the old failure text, the rewrite mutates the same two fields, and the new ForTesting setters take the same spinlock the old inline code took. One `catalog_.Lookup` plus the entry lock keeps the white box surface as narrow as before, and follows the locking rules this PR introduces. Our #3927 scenario suites run unmodified. Flat group all or none eviction matches what `EvictsWholeGroupTogether` and `ActiveGroupMemberBlocksWholeGroup` pin, so the model and the scenarios agree.

Most of the test deletions are sound and worth putting on record. `ShrinkBucketsIfSparse` and its two tests go away with the helper, and the PR body prices the deferred compaction. `ReRouteRestoredObjectsMovesStaleShardObjects` tested a migration that cannot exist without placement. `QueueLimitRejectsCrossShard` collapses into `QueueLimitRejectsBeyondCap` once shards are gone. `GroupedRoutingUsesHashOfTenantAndKeyOnly` pinned shard hash math, and `group_index_test` now covers its membership half. And the renamed `ClearInvalidHandlesSweepPreservesCoverage` keeps the erase all, survive all assertions.

Not every deletion is free, though. Dropping both commit versus clear ordering race tests left the suppression branch they pinned uncovered while it is still live, and the fixture gained `AccessorCreateRePinsWinnerEntry` with no TEST_F invoking it. Both are cheap to fix, details inline.

On the additions, `ConcurrentFirstWritesToDistinctKeysShareReachableTenant` pins get or create atomicity, `CandidateIndexMirrorsEntryState` pins the sparse candidate index against drift, and `TenantRemoveAllDoesNotBlockOtherTenants` proves the shared barrier claim with a held entry lock gate. `object_index_test` pins `EraseIf` identity and generation ordering, the index level form of the removal invariants the concurrent scenarios exercise. Good coverage to have.

Local runs at head 1c1835d1, same Ubuntu 22.04 container as my #3927 runs, RelWithDebInfo, gcc 11, USE_ETCD=OFF. The thirteen DSL suites come back green, master_scenario_test 44/44, evict 22/22, concurrent 10/10, offload 25/25, ssd 9/9, base scenario 32/32, group 13/13, placement 12/12, replication 16/16, segment 15/15, task 9/9, tenant 10/10, dfs 6/6. Suites this PR edits pass as well, master_service_test 59/59, group_test 11/11, double_erase 1/1, tenant_quota_test 31/31, promotion_on_hit_test 55/55, and the five new unit suites come out 20 for 20. I did not run the HA or client integration matrix.

A rebase is due again. Current main conflicts in `master_service.h`, `master_service.cpp` and `master_service_tenant_quota_test.cpp`, from #4052 and #3805. Worth folding the #3805 per tenant eviction watermark into `TenantCatalog` during that rebase, it fits this model naturally.

Two DSL follow-ups are mine to take once this merges, renaming `GroupOnDifferentShard` since the name encodes the removed sharding model, and a contract test pinning `ExpireAt`'s failure report. Please keep both out of this PR given the split request already standing.

mooncake-store/tests/ha/master_service_ha_test.cpp
```diff
@@ -6,6 +6,7 @@
 #include <algorithm>
 #include <atomic>
 #include <chrono>
+#include <atomic>
```
rishabhsinha17: Duplicate, `<atomic>` is already included two lines up.

mooncake-store/tests/master_service_processing_key_double_erase_test.cpp
```diff
@@ -85,21 +85,14 @@ class MasterServiceProcessingKeyDoubleEraseTest : public ::testing::Test {
...
 
-    // Friend access: find a key that routes to the SAME metadata shard as
-    // `key` (getShardIndex hashes tenant+key, so a naive second key
-    // lands in a different shard's TenantState and cannot keep THIS shard's
-    // tenant non-empty).
+    // Friend access: any second live key in the same tenant (Default) keeps
+    // the TenantCatalog non-empty. All of a tenant's keys live in a single
+    // TenantCatalog regardless of shard (there is no per-key shard to share),
+    // so a simple suffix suffices instead of a shard-index probe.
     std::string FindKeyOnSameShard(MasterService& service,
```
rishabhsinha17: The helper name and the SAME shard wording here and in the file header describe the old model, while the body is now a plain suffix. Worth renaming to something like MakeKeepaliveKey while you are in this file.

CAICAIIs:
> Reviewed the test and DSL lane on [1c1835d](https://github.com/kvcache-ai/Mooncake/commit/1c1835d1af44721627cbe9319804a6d325d797d9), wearing the MasterScenario DSL author hat (#3847, #3927). I read the full diff against merge base [6ae74b3](https://github.com/kvcache-ai/Mooncake/commit/6ae74b3512c4d0ce70c4a7c22bc4f33aeb7d3ff8) and ran the affected suites locally at the head, numbers at the end. Aionw's locking findings stand on their own and I did not re-verify them.
> 
> The DSL side is in good shape. This PR leaves `scenario.h` and `scenario_contract_test.cpp` untouched, so the grammar and its pins hold. Rewriting the `ExpireAt` engine preserved the verb's contract. Both miss paths keep the old failure text, the rewrite mutates the same two fields, and the new ForTesting setters take the same spinlock the old inline code took. One `catalog_.Lookup` plus the entry lock keeps the white box surface as narrow as before, and follows the locking rules this PR introduces. Our #3927 scenario suites run unmodified. Flat group all or none eviction matches what `EvictsWholeGroupTogether` and `ActiveGroupMemberBlocksWholeGroup` pin, so the model and the scenarios agree.
> 
> Most of the test deletions are sound and worth putting on record. `ShrinkBucketsIfSparse` and its two tests go away with the helper, and the PR body prices the deferred compaction. `ReRouteRestoredObjectsMovesStaleShardObjects` tested a migration that cannot exist without placement. `QueueLimitRejectsCrossShard` collapses into `QueueLimitRejectsBeyondCap` once shards are gone. `GroupedRoutingUsesHashOfTenantAndKeyOnly` pinned shard hash math, and `group_index_test` now covers its membership half. And the renamed `ClearInvalidHandlesSweepPreservesCoverage` keeps the erase all, survive all assertions.
> 
> Not every deletion is free, though. Dropping both commit versus clear ordering race tests left the suppression branch they pinned uncovered while it is still live, and the fixture gained `AccessorCreateRePinsWinnerEntry` with no TEST_F invoking it. Both are cheap to fix, details inline.
> 
> On the additions, `ConcurrentFirstWritesToDistinctKeysShareReachableTenant` pins get or create atomicity, `CandidateIndexMirrorsEntryState` pins the sparse candidate index against drift, and `TenantRemoveAllDoesNotBlockOtherTenants` proves the shared barrier claim with a held entry lock gate. `object_index_test` pins `EraseIf` identity and generation ordering, the index level form of the removal invariants the concurrent scenarios exercise. Good coverage to have.
> 
> Local runs at head [1c1835d](https://github.com/kvcache-ai/Mooncake/commit/1c1835d1af44721627cbe9319804a6d325d797d9), same Ubuntu 22.04 container as my #3927 runs, RelWithDebInfo, gcc 11, USE_ETCD=OFF. The thirteen DSL suites come back green, master_scenario_test 44/44, evict 22/22, concurrent 10/10, offload 25/25, ssd 9/9, base scenario 32/32, group 13/13, placement 12/12, replication 16/16, segment 15/15, task 9/9, tenant 10/10, dfs 6/6. Suites this PR edits pass as well, master_service_test 59/59, group_test 11/11, double_erase 1/1, tenant_quota_test 31/31, promotion_on_hit_test 55/55, and the five new unit suites come out 20 for 20. I did not run the HA or client integration matrix.
> 
> A rebase is due again. Current main conflicts in `master_service.h`, `master_service.cpp` and `master_service_tenant_quota_test.cpp`, from #4052 and #3805. Worth folding the #3805 per tenant eviction watermark into `TenantCatalog` during that rebase, it fits this model naturally.
> 
> Two DSL follow-ups are mine to take once this merges, renaming `GroupOnDifferentShard` since the name encodes the removed sharding model, and a contract test pinning `ExpireAt`'s failure report. Please keep both out of this PR given the split request already standing.

thanks your review, and I will split this large PR into multiple smaller ones and take your suggestions into account.

PR #3825, opened 2026-09-02
[Build] conductor: build skeleton and common layer

Chase-Rong:
## Description

First slice of the Mooncake Conductor upstreaming series. This series splits it into 7 dependency-
ordered slices; this PR is the skeleton so the later slices are pure additions.

What lands here:

- `mooncake-conductor/CMakeLists.txt` building the `conductor_cpp_core`
  static library, with a `GLOBAL_CONFIG` guard so the component also
  configures standalone (outside the top-level build).
- Common layer: the `ServiceConfig` / `HashProfileConfig` /
  `ResolvedHashProfile` data model (`common/types.h`) and the
  environment-variable and log-level helpers (`common/utils.{h,cpp}`).
- `tests/CMakeLists.txt` with the `conductor_test` gtest target, covering
  the env helpers.
- `WITH_CONDUCTOR` option, default `OFF`, wired into the top-level build
  and documented in `build.md`.

No conductor binary is produced yet — `src/main.cpp` depends on the KV
event manager and arrives with that slice. The remaining slices are, in
order: prefix index types + hash strategy, tiered prefix indexer, ZMQ
subscriber + wire protocol, KV event manager + HTTP API, then e2e tests +
cache-aware proxy example + docs.

Note on CI wiring: the conductor tests are picked up by enabling
`-DWITH_CONDUCTOR=ON` in the existing unified `Configure project with unit
tests` step, so `conductor_test` flows through the current CTest run and
its JUnit report and failure diagnostics. This is one line rather than a
separate build-and-run step.

Related: the store-side KV event publisher changes are in #3818. There is
no compile dependency between the two — conductor includes no
`mooncake-store` headers — so they can land in either order.

## Module

- [x] CI/CD
- [x] Docs
- [x] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

```bash
cmake -S mooncake-conductor -B build -G Ninja -DBUILD_UNIT_TESTS=ON
cmake --build build -j8
```
Result: configure and build clean, conductor_test passes — 9 tests
across 4 suites. Verified on this slice alone, with no later-slice files
present, to confirm it builds independently.

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

AI-assisted. The conductor sources are carried over unmodified from the
existing conductor branch; AI was used to determine the dependency-ordered
split, trim the CMake scaffolding to only what this slice needs, and adapt
the CI wiring to mainline's current unified CTest layout. Every changed
line has been reviewed by the submitter.


catyans (approved):
LGTM. The opt-in build wiring works in both top-level and standalone configurations, the environment parsing is strict and covered, and the uint64 JSON boundary tests preserve the intended values.

mooncake-conductor/include/conductor/common/types.h
```diff
@@ -0,0 +1,74 @@
+#pragma once
+
+#include <cstdint>
+#include <optional>
+#include <string>
+#include <string_view>
+
+namespace conductor {
```
Aionw: Can we using conductor as as sub namesapce for mooncake, like

```cpp
namespace mooncake::conductor::common {  }
```

mooncake-conductor/include/conductor/common/utils.h
```diff
@@ -0,0 +1,27 @@
...
+#include <string>
+
+namespace conductor {
+namespace common {
+
+enum class LogLevel { kDebug, kInfo, kWarn, kError };
+
+// Reads CONDUCTOR_LOG_LEVEL (DEBUG/INFO/WARN/ERROR, case-insensitive).
+// Empty -> INFO; invalid values warn and fall back to INFO.
+LogLevel ParseLogLevel();
```
Aionw: I'm wondering why we need to create a new file for functionality that already exists in mooncake store. Can we reuse the existing utility instead, so we don't have to maintain two similar codebases for basically the same functionality?

One approach is to make those function a common file include by those project who need it.

mooncake-conductor/include/conductor/common/types.h
Chase-Rong: > Can we using conductor as as sub namesapce for mooncake, like
> 
> ```c++
> namespace mooncake::conductor::common {  }
> ```

done

mooncake-conductor/include/conductor/common/utils.h
Chase-Rong: Agreed on the principle, we'll reuse rather than duplicate.  `ToUpper` and `AtoiStrict` now call `mooncake::AsciiToLower` and `mooncake::TryParseInteger` from mooncake-common. Both headers are header-only, so this adds an include path without linking `mooncake_common`. 

Regarding log levels, nothing can be reused. Mooncake-common does not have glog logic (`init_ylt_log_level()` drives YLT’s internal easylog). Each module (Mooncake storage/transfer engine) initializes its own logging: every entry point calls `InitGoogleLogging`, and the case or level names used in these calls are inconsistent (`transfer-engine/src/config.cpp:422`, `tent/.../ transfer_engine_impl.cpp:286`, `mooncake-store/src/master.cpp:1402`, `mooncake-common/src/environ.cpp:177`—the last of which is not called by any module). Consequently, there are no shared helper functions that can be called for the log levels themselves.

catyans (approved):
LGTM. Re-reviewed the current head: the Conductor namespace now sits under `mooncake`, and the environment parsing reuses the shared ASCII/integer helpers while preserving the strict parsing contract and focused tests. Thanks @Aionw and @Chase-Rong for tightening this up. The remaining CI jobs should complete before merge.

ykwd (approved):

Issue #1058, opened 2025-11-14
[RoadMap] Mooncake Transfer Engine NEXT


alogfans:
# Mooncake Transfer Engine NEXT (TENT) Roadmap

This document outlines the detailed roadmap for TENT, following the release of its technical preview. We categorize the roadmap into two phases:

- **Phase 1**: Make it production-ready
- **Phase 2**: Advanced feature development

> The roadmap is informed by the TE+ system architecture described in the FAISys 25 talk _"TENT: A Declarative Slice Spraying Engine for Performant and Resilient Data Movement in Disaggregated LLM Serving."_ 
> TENT aims to implement declarative orchestration, dynamic path scheduling, and resilience in a modular and community-driven way.

---

### Transport Backend Support
- [x] NVIDIA (GDR, GDS, NVLink, MNNVL)
- [ ] AMD/RoCM GPU (via ROCm APIs, port from mainstream version)
- [x] Huawei Ascend NPU (via UB/HIXL, completed)
    - [ ] Direct file I/O
- [ ] Moorethread NPUs (requires vendor interface, port from mainstream version)
- [ ] Multi-rail TCP (with parallel stream slicing and connection pooling)
- [ ] More NPUs and SmartNICs (as thin transport plugins)
- [ ] Backport to the mainstream
    - [ ] NVLink (w/o fabric)

### Configuration Handling
- [x] Load configuration from file
- [ ] Support override via environment variables or runtime parameters
- [ ] Hot-reload or staged application of config changes

### Logging
- [ ] Adjustable log levels (INFO/DEBUG) to suppress noise in prod
- [ ] Write logs to rotating file instead of stdout
- [ ] Non-blocking async file writing (community friendly)

### Python Integration
- [X] Compatible fallback via `MC_USE_TEV1=1`
- [ ] Clean high-level declarative Python API (`transfer(src, dst)` style)
- [ ] Runtime integration with Mooncake Store (replace build-time coupling)
- [X] Integrate with Mooncake Store at build type

### Packaging & Usability
- [ ] Provide Python wheel with compiled transport backends
- [ ] Enable runtime dynamic loading without manual config
- [ ] Extract TENT into a standalone service or library
- [ ] CLI tool for isolated transfers & diagnostics (community friendly)
- [ ] Docs & guide

---

### Declarative Slice Scheduler
- [ ] Avoid starvation problem
- [ ] Considering from both senders and receivers during scheduling
[...]

PR #3834, opened 2026-09-02
[TENT] Route HP TCP lanes across configured rails

jacklin78911-collab:
## Description

Follow-up to #3689 and related to #1058.

This PR adds optional static multi-rail routing to `hp_tcp`:

- `rail_addresses` publishes an ordered endpoint list.
- Lane `i` uses rail `i % rail_count` and binds to the matching local source address.
- Existing lane-to-worker affinity and persistent connection reuse remain unchanged.
- An empty `rail_addresses` list preserves the original single-endpoint behavior.
- Both peers must configure the same rail count; multiple rails require a wildcard listener.

This PR routes complete persistent lanes across rails. It does not add per-transfer striping, rail failover, adaptive scheduling, socket tuning, GPU/CANN support, or QoS. Unlike #1633, an individual transfer remains on one lane and one rail.

## Module

- [x] Docs

## Type of Change

- [x] New feature
- [x] Documentation update
- [x] Performance improvement

## How Has This Been Tested?

Added focused coverage for:

- single-endpoint and multi-endpoint metadata;
- `rail_addresses` configuration parsing;
- source-address binding;
- multi-rail endpoint publication;
- two-process READ/WRITE over two loopback rails with full payload comparison.

Previous dual-host validation of the same routing logic:

- all four configured source/destination rail pairs were used;
- small- and large-buffer payload consistency checks passed;
- 64 MiB exploratory results improved from approximately 23.73 to 25.61 GB/s for READ and from 22.90 to 24.09 GB/s for WRITE.

These performance results are specific to the tested four-NIC environment. This PR does not claim linear scaling.

Local checks:

- `git diff --check`
- clang-format 20
- independent full-diff and data-path review

Final validation will also use the GitHub CI results from this PR.

## Checklist

- [ ] I have performed a final human review of every changed line
- [x] I have formatted the changed C++ code with clang-format 20
- [x] I have added focused tests
- [x] The production change is below the 500-line RFC threshold

## AI Assistance Disclosure

- [x] AI tools were used

### DIFF
--- a/docs/source/design/tent/hp-tcp.md
+++ b/docs/source/design/tent/hp-tcp.md
@@ -2,8 +2,8 @@
 
 `hp_tcp` is a standalone TENT transport for CPU DRAM transfers over
 data-center TCP. Standard `tcp` remains the RPC-based compatibility path.
-The first version intentionally excludes GPU memory, TLS, multi-endpoint
-routing, multi-NIC striping, transparent replay after an ambiguous WRITE and
+The first version intentionally excludes GPU memory, TLS, per-transfer
+striping, rail failover, transparent replay after an ambiguous WRITE and
 dynamic lane scheduling.
 
 ## Architecture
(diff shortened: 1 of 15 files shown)


catyans:
issue: a single-rail configuration can publish an address that the server is not listening on. `validateParams()` requires a wildcard listener only when `rail_addresses.size() > 1`, so `bind_address=10.0.0.1` with `rail_addresses=["10.1.0.1"]` is accepted; `install()` then listens on the former but advertises the latter, making every peer connection fail.

Please require a non-wildcard `bind_address` to equal the sole rail address (or require wildcard binding whenever rails are configured), and add a validation regression for the mismatched case.

catyans (approved):
Thanks, this addresses my concern. LGTM. The validation now requires a specific listener to match the sole rail address, while retaining wildcard binding for multiple rails; the mismatch regression and documentation cover the contract. CI has just restarted.

alogfans (approved):
Overall, this looks good to me. 

One suggestion: the new multi-rail E2E test verifies payload correctness, but it would still pass if all traffic accidentally used only one rail. It would be valuable to record the accepted connections’ peer addresses and assert that both configured source addresses are observed. This could be a follow-up if it is difficult to add now.

Also, this should be treated as the static-routing foundation for multi-rail TCP rather than completion of the roadmap item involving per-transfer parallel stream slicing.

jacklin78911-collab:
@alogfans Thanks for the review and merge. Agreed—#3834 is the static lane-to-rail routing foundation, not the per-transfer parallel stream slicing roadmap item.

The current socket test verifies configured source binding, while the E2E verifies payload correctness. It does not yet prove that both rails are exercised within the same E2E, so I’ll keep the accepted-peer-address assertion as a focused follow-up rather than add test-only instrumentation to this foundation PR.

Issue #3833, opened 2026-09-02
[RFC]: TENT Versioned Configuration Lifecycle and Staged Runtime Application


bugkeep:
## Summary

The TENT roadmap calls for **hot-reload or staged application of configuration changes** (#1058), but the current runtime has no configuration lifecycle contract that makes such updates safe.

This RFC proposes a TENT-specific configuration model with:

- a conservative split between immutable bootstrap configuration and versioned runtime configuration;
- validated, immutable whole-configuration snapshots;
- an explicit staged `prepare -> publish -> commit/rollback` update protocol;
- well-defined consistency boundaries for in-flight requests and batches;
- explicit ownership rules for process-wide resources and multiple `TransferEngine` instances.

The first milestones establish the lifecycle and validation foundation. They do **not** enable arbitrary hot reload or change current runtime behavior.

## Motivation

TENT already supports loading configuration from `MC_TENT_CONF`, environment overrides, and caller-provided `Config` values. However, configuration is currently a mutable JSON object whose individual `get()` and `set()` operations are protected by a mutex. That prevents a data race on the JSON object, but it does not provide an atomic, coherent runtime update.

During `TransferEngineImpl::construct()`, many values are copied into engine members or used to construct long-lived components:

- RPC address, port, and server thread count;
- progress worker and admission queue limits;
- topology and platform loader;
- transport instances and transport selector;
- metrics server configuration;
- failover and request-merging behavior.

Changing the underlying `Config` after construction therefore has inconsistent effects: some code may read a new value, while already-constructed components continue using old values.

There is also process-wide state:

- `Platform::getLoader()` is initialized with `std::call_once` and silently retains the first engine's configuration;
- `TentMetrics` is a process-wide singleton;
- logging level and several allocator/prober facilities are process-wide.

This makes staged updates and multiple engine instances ambiguous unless ownership and conflict behavior are defined first.

## Goals

1. Define one canonical, validated configuration representation after file/environment/API inputs are merged.
2. Classify fields conservatively as bootstrap-only, runtime-updatable, or unsupported for live update.
3. Publish runtime configuration as immutable snapshots with monotonically increasing generations.
4. Validate complete candidate snapshots, including cross-field and build-capability constraints.
5. Prevent partial visibility: readers observe either the old generation or the new generation.
6. Define safe update boundaries for requests, batches, workers, selectors, and transports.
7. Define `prepare`, `commit`, rejection, and rollback behavior for runtime consumers.
[...]

PR #3836, opened 2026-09-02
[TENT] Add configuration lifecycle foundation

bugkeep:
## Description

TENT currently reads startup settings and policy values through the same mutable `Config`. This PR adds lifecycle views over one immutable capture of the already-merged effective configuration, plus a generation-tagged runtime snapshot.

- Use two classifications: `kBootstrapOnly` and `kRuntimeCandidate`. An explicit runtime-candidate allowlist uses exact/subtree matching with longest-match precedence; unmatched paths default to bootstrap-only.
- Freeze the effective configuration once and share that backing object between `BootstrapConfig` and `RuntimeConfig`.
- Restrict reads to the view's lifecycle and reject mixed-lifecycle subtree/object reads.
- Require non-empty read paths with no empty slash-delimited segments. The view rejects leading, trailing, and repeated slashes before lifecycle classification and descendant checks. This closes the `/merge_requests`, `metrics/`, and `/` aliases caused by legacy `Config` skipping empty segments. Rejected reads return the supplied default or `false`, and `dumpSubtree` leaves its output unchanged.
- Preserve nested-before-flat lookup for canonical paths and the configuration precedence established by the caller.
- Retain invalid-root diagnostics and focused allowlist, lifecycle-boundary, snapshot, and precedence tests.

`kRuntimeCandidate` records eligibility for future runtime updates. Runtime publication, component apply handlers, and configuration typo checking remain follow-up work.

Refs #3833

## Module

## Type of Change

- [x] New feature

## How Has This Been Tested?

```bash
cmake --build build/tent-config-lifecycle --target config_lifecycle_test -j4
ctest --test-dir build/tent-config-lifecycle -R '^config_lifecycle_test$' --output-on-failure -V
pre-commit run --files \
  mooncake-transfer-engine/tent/include/tent/common/config_lifecycle.h \
  mooncake-transfer-engine/tent/src/common/config_lifecycle.cpp \
  mooncake-transfer-engine/tent/tests/config_lifecycle_test.cpp
```

- All three new path-alias regression tests failed against the implementation at `7753b6da` before the fix. They cover the owner's three cases, repeated internal slashes, scalar/object/contains access, both views, and unchanged output on rejection.
- All 15 lifecycle unit tests pass with the fix, including the existing canonical path, mixed-subtree, snapshot immutability, nested-before-flat lookup, and environment precedence tests.
- Pre-commit passes for the three files changed by the path-alias fix, including the repository's clang-format 20 hook.
- A separate six-file PR-wide pre-commit attempt reported an existing CMake formatting difference in the unrelated `tent_local_memory_lifecycle_test` block. That automatic rewrite was excluded from this focused fix; a complete PR-wide pre-commit pass is not claimed.
- The local shell helper was restored to the repository's LF line endings so the C++ formatting hook could execute under WSL. This introduces no tracked script change.
- GitHub CI results must be evaluated on the latest head; earlier green runs do not validate this revision.

## Checklist

- [ ] I have performed a self-review of my own code (human line-by-line review pending).
- [x] I have formatted the path-alias fix using `./scripts/code_format.sh` through the pre-commit hook.
- [ ] I have run pre-commit on the files changed in this PR and all hooks pass (see scoped results above).
- [ ] I have updated the documentation (not applicable to this foundation slice).
- [x] This work has an associated RFC: #3833.

## AI Assistance Disclosure

- [ ] No AI tools were used.
- [x] AI tools were used: OpenAI Codex assisted with repository analysis, implementation, tests, and review. The human submitter is responsible for reviewing every changed line and understanding and defending the change end-to-end before marking this PR ready.


mooncake-transfer-engine/tent/src/common/config_lifecycle.cpp
```diff
@@ -0,0 +1,233 @@
...
+#include "tent/common/config_lifecycle.h"
+
+namespace mooncake {
+namespace tent {
+namespace {
+
+// Transport namespaces are extension points and therefore use subtree
+// entries. Runtime-eligible exceptions below them are listed as exact fields;
+// classifyConfigPath() resolves those using longest-match semantics.
+constexpr ConfigFieldSpec kConfigFields[] = {
```
alogfans, line 24: Can we add a safeguard to ensure that any newly added config key is also updated in the lifecycle inventory?

catyans:
Thanks for establishing the lifecycle boundary. Building on @alogfans's inventory-safeguard point, subtree/object reads currently cross that boundary: `bootstrap.dumpSubtree("metrics")` or `bootstrap.get<json>("transports/rdma", ...)` is allowed by the parent classification, but returns runtime-candidate children such as `metrics/report_interval_seconds` or `transports/rdma/strict_local_numa`. Could mixed-lifecycle subtree reads be filtered or rejected, with a regression covering one of these cases?

bugkeep:
Thanks for catching this. Addressed in e0ea554d.
Mixed-lifecycle subtree/object reads are now rejected as a whole. get, dumpSubtree, and contains verify that every configured descendant belongs to the view’s lifecycle before exposing the requested value.
I added regression coverage for both reported cases:
- bootstrap.dumpSubtree("metrics")
- bootstrap.get<json>("transports/rdma", ...)

The tests also verify that direct leaf reads and homogeneous subtrees continue to work. All related tests pass locally.

staryxchen (changes requested):
Thanks for your work. I believe two classes, `kBootstrapOnly` and `kRuntimeCandidate`, are enough for this PR.

Please drop `kDerived`. Bound ports, discovered topology, and similar computed state live on engine objects, not in `Config`. The inventory has no derived paths; putting this value on the enum implies JSON may contain read-only computed keys.

Please drop `kUnsupported` and default unmatched paths to `kBootstrapOnly`. `Config` has no schema today: unknown or stale keys are silently ignored (e.g. stale metrics buckets). Unknown-key diagnostics belong to typo checking / `check-config`, not to lifecycle. Live-update safety only needs a runtime allowlist: an unlisted key treated as bootstrap may require a restart, but it will not be hot-applied by mistake.

With that default, a full-key inventory audit is no longer needed. If we keep a test, it should only guard the runtime-candidate allowlist (longest-match overrides).

jacklin78911-collab:
Hi @bugkeep, I'm working on a follow-up that pins failover policy at transfer submission and builds on this foundation.

I can help address @staryxchen's feedback: keep only bootstrap/runtime-candidate classification, default unmatched keys to bootstrap-only, and remove unsupported-key diagnostics and the full source audit while retaining focused lifecycle tests.

Are you already working on these changes? If helpful, I can send a small PR against your branch.

bugkeep:
> Hi @bugkeep, I'm working on a follow-up that pins failover policy at transfer submission and builds on this foundation.
> 
> I can help address @staryxchen's feedback: keep only bootstrap/runtime-candidate classification, default unmatched keys to bootstrap-only, and remove unsupported-key diagnostics and the full source audit while retaining focused lifecycle tests.
> 
> Are you already working on these changes? If helpful, I can send a small PR against your branch.

Thanks for offering! I haven’t started these changes yet, so a small PR against my  branch would be welcome.

staryxchen (commented):
The two-class model looks good to me now. Two small leftovers:

Please update the PR description — it still mentions derived/unsupported classification and unsupported-field diagnostics.

`ConfigFieldSpec::lifecycle` is unused in practice: every allowlist entry is `kRuntimeCandidate`, and unmatched paths already default to bootstrap. Consider dropping the field and storing only path + match.

bugkeep:
Thanks @staryxchen! Both follow-up points are addressed:

- Updated the PR description to reflect the two-class model and bootstrap-only default.
- Removed `ConfigFieldSpec::lifecycle` in [7753b6da](https://github.com/kvcache-ai/Mooncake/pull/3836/commits/7753b6da865720dd7b43221b34067a728112e2f9). Entries now store only `path` and `match`; the classifier returns `kRuntimeCandidate` for matches and `kBootstrapOnly` otherwise. All 31 allowlist entries and their match modes are unchanged.

The inventory test now checks actual classification through `classifyConfigPath`. All 12 lifecycle tests and clang-format 20 checks pass locally. CI for the new commit is pending.

staryxchen (commented):
Thanks for updating the description and dropping ConfigFieldSpec::lifecycle.

The path-alias leak you called out should be fixed in this PR: Config skips empty segments, so after unmatched paths default to bootstrap, bootstrap.get("/merge_requests"), dumpSubtree("metrics/"), and dumpSubtree("/") can cross the lifecycle boundary. Please canonicalize (or reject) noncanonical paths in the view, and add regressions for those three cases.

bugkeep:
> Thanks for updating the description and dropping ConfigFieldSpec::lifecycle.
> 
> The path-alias leak you called out should be fixed in this PR: Config skips empty segments, so after unmatched paths default to bootstrap, bootstrap.get("/merge_requests"), dumpSubtree("metrics/"), and dumpSubtree("/") can cross the lifecycle boundary. Please canonicalize (or reject) noncanonical paths in the view, and add regressions for those three cases.

ok

staryxchen (approved):
LGTM, thanks.

PR #3837, opened 2026-09-02
[TE] Add metadata versioning

alogfans:
## Description

Add lightweight metadata versioning to Transfer Engine metadata to reduce stale metadata usage.

This PR adds a `metadata_version` field to both `SegmentDesc` and `RpcMetaDesc`. The version is a wall-clock microsecond timestamp, and logs/dumps render it as a readable local time for easier incident recovery.

The cache refresh path now:

- accepts newer metadata versions
- rejects older fetched metadata as stale
- rejects same-version-but-different metadata and logs it as a possible concurrent publisher or missing version bump
- invalidates cached RPC metadata only after the segment metadata update passes version checks

Publishing paths update the version before writing centralized metadata, while P2P handshake keeps legacy-compatible version `0`.

## Module

## Type of Change

- [x] Bug fix
- [x] New feature

## How Has This Been Tested?

**Test commands:**
```bash
clang-format-20 --dry-run --Werror mooncake-transfer-engine/include/common.h mooncake-transfer-engine/include/transfer_metadata.h mooncake-transfer-engine/src/transfer_metadata.cpp mooncake-transfer-engine/src/transfer_metadata_dump.cpp mooncake-transfer-engine/tests/tcp_write_visibility_test.cpp mooncake-transfer-engine/tests/transfer_metadata_test.cpp

git diff --check -- mooncake-transfer-engine/include/common.h mooncake-transfer-engine/include/transfer_metadata.h mooncake-transfer-engine/src/transfer_metadata.cpp mooncake-transfer-engine/src/transfer_metadata_dump.cpp mooncake-transfer-engine/tests/tcp_write_visibility_test.cpp mooncake-transfer-engine/tests/transfer_metadata_test.cpp

cmake --build build --target transfer_metadata_test tcp_write_visibility_test -j

build/mooncake-transfer-engine/tests/transfer_metadata_test --gtest_filter='TransferMetadataVersionTest.*:TransferMetadataPollingTest.*:TransferMetadataPublicationTest.*:TransferMetadataValidationTest.*' --logtostderr=1

build/mooncake-transfer-engine/tests/tcp_write_visibility_test --gtest_filter='TcpWriteVisibilityTest.EndpointRefreshRoutesNewWorkToNewTcpEndpoint:TcpWriteVisibilityTest.RpcMetadataRefreshInvalidatesCachedHost:TcpWriteVisibilityTest.ConcurrentEndpointRefreshNeverCombinesHostAndPortSnapshots' --logtostderr=1

build/mooncake-transfer-engine/tests/tcp_write_visibility_test --gtest_filter='TcpWriteVisibilityTest.CompletedWriteIsVisibleToSubsequentRead' --logtostderr=1
```

**Test results:**
- [x] Unit tests pass
- [x] Integration tests pass
- [x] Manual testing done

Notes:
- `transfer_metadata_test`: 9 tests passed.
- `tcp_write_visibility_test`: centralized HTTP metadata refresh/transfer tests passed.
- P2P handshake TCP write/read smoke test passed.
- A local etcd smoke test was attempted, but this build does not include the etcd metadata plugin (`USE_ETCD`), so it failed during plugin initialization before exercising this change.
- `pre-commit` was not run because it is not installed in the local environment.

## Checklist

- [x] I have formatted my code using `clang-format-20`

## AI Assistance Disclosure

Codex was used to help implement the metadata versioning change, add focused tests, and run local validation. The human submitter should review and own all changed lines before submitting.


catyans:
Thanks for adding the version guards and addressing the earlier review comments. I found one remaining edge case in `syncSegmentCache()`: because `SegmentDesc::operator==` excludes `metadata_version`, the `!changed` branch does not install a newer version when the payload is unchanged. A cache at v1 can observe v3(A), remain at v1, and later accept stale v2(B). Could we advance the cached version on version-only updates and add a v1(A) -> v3(A) -> v2(B) regression test?

alogfans:
@catyans Thanks, we will retrieve latest version when finding inconsistent version.

staryxchen (commented):
RPC invalidation is tied to payload change: the new !changed path updates the segment watermark but does not erase rpc_meta_map_. getRpcMetaEntry() never force-refreshes, so the RPC version checks do not run on a cache hit. An RPC-only host/port update (same buffers) can leave handshake/notify/probe on the stale address.

Previously any successful segment fetch during sync dropped the RPC cache. Could we invalidate RPC when a newer segment version is accepted—including version-only updates—or refresh RPC by its own version, and add a test where RPC host changes while the segment payload stays the same?

alogfans:
@staryxchen Updated with minimal changes: when `syncSegmentCache()` accepts a newer version-only segment update, it now also invalidates the cached RPC metadata so the next `getRpcMetaEntry()` refreshes host/port using the RPC metadata version checks.

staryxchen (approved):
LGTM

PR #3841, opened 2026-09-03
[Store] Add bounded standby promotion handoff

Icedcoco:
## Description

Depends on #3811.

Implement PR-N07 bounded standby promotion for the batch-OpLog snapshot path:

- move the complete `StandbyMetadataStore` into a move-only promotion context instead of exporting a second full object vector;
- carry the applied batch/sequence cursor, producer view, segments, and maximum live `ReplicaID` through the handoff;
- drain and install primary metadata in bounded object chunks while preserving cross-chunk validation and allocator accounting;
- use `batch_oplog_retry_timeout_sec` as a no-progress timeout during final catch-up instead of a fixed 30-second total deadline;
- keep the legacy snapshot vector path unchanged and keep RPC registration behind the complete restore gate.

GitHub cannot use `dev/oplog-ha-prs/I00` as the upstream base because that branch exists only in the contributor fork. #3811 is now merged into `main`; this branch has been rebased onto the resulting `main` and is N07-only. The N07 commit is [d3d418584](https://github.com/Icedcoco/Mooncake/commit/d3d418584c57538735b3fba89cb02a4c63fb5584).

#3806 also edits `RestoreFromStandbySnapshot`, but for different semantics: it tolerates and discards malformed objects. This PR bounds ownership and conversion memory while retaining the current R01 fail-closed behavior. If #3806 lands first, the shared restore implementation will need a semantic rebase rather than duplicating either path.

## Module

## Type of Change

- [x] New feature
- [x] Performance improvement

## How Has This Been Tested?

**Test commands:**
```bash
cmake --build /tmp/mooncake-n07-build --target hot_standby_service_test hot_standby_snapshot_bootstrap_test standby_metadata_store_test batch_oplog_snapshot_provider_test batch_oplog_snapshot_promotion_test master_service_ha_test batch_oplog_promotion_bench -j8
ctest --test-dir /tmp/mooncake-n07-build --output-on-failure -R '(^hot_standby_service_test$|^hot_standby_snapshot_bootstrap_test$|^standby_metadata_store_test$|^batch_oplog_snapshot_provider_test$|^batch_oplog_snapshot_promotion_test$|^master_service_ha_test$)'
/tmp/mooncake-n07-build/mooncake-store/benchmarks/batch_oplog_promotion_bench --objects=10000 --chunk_objects=128
```

The build used `STORE_USE_ETCD=ON` so the final catch-up tests were compiled.

**Test results:**
- [x] Six related test targets passed, including the complete `master_service_ha_test` suite
- [x] Snapshot-only provider fallback and empty batch namespace promotion regressions passed
- [x] Eight `PromotionCatchUpTest` cases passed
- [x] Synthetic 10,000-object promotion completed in 108 ms with 70.9 MB peak RSS in this debug build
- [x] C++ formatting, whitespace, conflict, large-file, spelling, and CMake formatting hooks passed; unrelated pre-existing CMake reformatting was left out per `AGENTS.md`

## Checklist

- [x] I have formatted my code using the project formatter
- [ ] I have run pre-commit on the files changed in this PR and all hooks pass (see the scoped CMake note above)
- [x] Documentation update is not applicable; N08 owns configuration and production wiring
- [x] RFC #3167 covers this change

## AI Assistance Disclosure

- [x] AI tools were used (Codex implemented the change, tests, benchmark, and local verification)

The human submitter must review every changed line and be able to defend the change end-to-end before merge.

### DIFF
--- a/mooncake-store/benchmarks/CMakeLists.txt
+++ b/mooncake-store/benchmarks/CMakeLists.txt
@@ -52,6 +52,10 @@ target_link_libraries(
   batch_evict_bench PRIVATE mooncake_store cachelib_memory_allocator
                             gflags::gflags glog::glog pthread)
 
+add_executable(batch_oplog_promotion_bench batch_oplog_promotion_bench.cpp)
+target_link_libraries(batch_oplog_promotion_bench PRIVATE mooncake_store
+                                                          gflags::gflags)
+
 if(STORE_USE_ETCD)
   add_executable(oplog_batch_bench oplog_batch_bench.cpp)
   target_link_libraries(
--- a/mooncake-store/benchmarks/batch_oplog_promotion_bench.cpp
+++ b/mooncake-store/benchmarks/batch_oplog_promotion_bench.cpp
@@ -1,0 +1,84 @@
+#include <gflags/gflags.h>
+#include <glog/logging.h>
+
+#include <chrono>
+#include <fstream>
+#include <iostream>
+#include <limits>
+#include <string>
+
+#include "ha/snapshot/batch_oplog/promotion.h"
+#include "master_service.h"
+
+DEFINE_uint64(objects, 100000, "Synthetic standby object count");
+DEFINE_uint64(chunk_objects, 1000, "Objects drained per chunk");
+
+namespace {
+
+size_t PeakRssBytes() {
+    std::ifstream status("/proc/self/status");
+    std::string name;
+    size_t rss_kib = 0;
+    while (status >> name) {
+        if (name == "VmHWM:") {
+            status >> rss_kib;
+            break;
+        }
+        status.ignore(std::numeric_limits<std::streamsize>::max(), '\n');
+    }
+    return rss_kib * 1024;
+}
+
+}  // namespace
+
+int main(int argc, char** argv) {
+    google::InitGoogleLogging(argv[0]);
+    FLAGS_logtostderr = true;
+    gflags::ParseCommandLineFlags(&argc, &argv, true);
+    if (FLAGS_objects == 0 || FLAGS_chunk_objects == 0 ||
+        FLAGS_objects == std::numeric_limits<uint64_t>::max() ||
+        FLAGS_chunk_objects > std::numeric_limits<size_t>::max()) {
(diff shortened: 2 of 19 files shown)


catyans:
Nice direction—the move-only handoff removes the second full object vector, while restore keeps cross-chunk duplicate, range, and capacity validation, and RPC registration remains behind the completed restore. One non-blocking question: `producer_view_version` and `applied_cursor.batch_id` are carried through the handoff and supervisor, but `RestoreFromBatchOpLogPromotion()` currently consumes only `last_seq` and `max_replica_id`. IIUC the new primary already claims its own producer view and the writer reloads the durable prefix from storage, so are these two fields intentionally informational for N08? If so, documenting that contract (or asserting the expected relationship) would keep them from looking like unenforced safety state.

Icedcoco:
Yes, this is intentional. producer_view_version is diagnostic only: the new primary claims and uses its own leader view, so it must not reuse the standby’s view. applied_cursor.batch_id is retained for cursor identity and audit; the batch reader already validates batch/sequence continuity, while the new primary writer reloads the authoritative full DurablePrefix from the backend. Therefore the restore path only needs the sequence portion and max_replica_id; no runtime change is required. A short comment can make this contract explicit. @catyans

Icedcoco:
Fixed in `347f1a2c9`:

- Legacy `RestoreFromStandbySnapshot()` keeps the contiguous `vector<pair<...>> + sort` range validation path; only bounded batch promotion uses the cross-chunk `std::map` index.
- `HotStandbyService::Start()` recreates a detached `StandbyMetadataStore` before constructing `OpLogApplier`, so a failed primary restore can restart standby and bootstrap again instead of passing a null store.
- Added regressions for snapshot-only fallback, empty batch namespace zero cursor, and restart after batch detach.

Validation: scoped pre-commit passed; promotion tests 7/7, final catch-up tests 10/10, and restore-related HA tests 16/16 passed. The complete `master_service_ha_test` suite also passed in the same ETCD-enabled build.

Aionw (approved):
LGTM

ykwd (approved):

PR #3845, opened 2026-09-03
[Store/RDMA] Fix ibv_reg_mr failures with 2MB hugepage SHM (memfd + MTT alignment)

1998zxn:
## Description

When `MC_STORE_USE_HUGEPAGE=1` is set, starting the Mooncake store with a large KV cache buffer causes `ibv_reg_mr` to fail with **EINVAL (-600)**, preventing RDMA registration of the host memory pool.

Observed on **GLM-5.2-FP8, TP=8**: each TP worker allocates four ShmHelper pools totalling ~11–12 GB per worker (~90 GB across 8 workers). Representative pool sizes (one worker, 2 MB hugepages; the physical segment is what `ShmHelper::allocate()` actually maps, rounded up to hugepage granularity):

| Pool | Requested size | Physical segment (hugepage-rounded) | Rounding overhead |
|------|---------------|-------------------------------------|-------------------|
| swa | 3,779,571,220 B (3,604.5 MB) | 3,781,165,056 B (1,803 × 2 MB) | +1.5 MB |
| deepseek_v4_c4 | 4,434,553,733 B (4,229.9 MB) | 4,435,476,480 B (2,115 × 2 MB) | +0.9 MB |
| indexer | 998,579,896 B (952.3 MB) | 1,000,341,504 B (477 × 2 MB) | +1.7 MB |
| state | 773,094,113 B (737.3 MB) | 773,849,088 B (369 × 2 MB) | +0.7 MB |

None of the four requested sizes is a 2 MB multiple, so every pool carries the hazard below; the two largest surfaced first: `swa` (3,604 MB, registered with its logical size, whose end lands inside the last huge page of the mapping) and `deepseek_v4_c4` (4,230 MB, above the 4 GiB threshold where the parallel preTouch path trial-registers each thread's block).

The rounding overhead itself is negligible (≤ 2 MB per pool, ~5 MB per worker, ~40 MB across 8 workers) and — importantly — **this PR does not change allocation at all**: `ShmHelper` already rounds the mapping up to hugepage granularity. The fix is to make registration agree with the mapping instead of reshaping the mapping to fit the registration.

### Root cause

strace-level tracing pinpointed the failure origin, and it never reaches the mlx5 driver. After `ibv_fork_init()` is enabled, every `ibv_reg_mr` first runs `ibv_dontfork_range()`, which issues `madvise(MADV_DONTFORK)` over the memory range. The kernel refuses to split a hugetlb VMA at a non-hugepage-aligned boundary, so any registered range whose start or end falls inside a huge page fails with **EINVAL** — before the driver is ever reached. For 2 MB hugepages this means every registered range must start and end on 2 MB boundaries (and for 1 GB hugepages, on 1 GB boundaries).

This requirement is violated at two points in the call chain, both of which must be fixed:

### Bug 1: register_buffer passes the logical size (`real_client.cpp`)

`register_buffer()` is called from Python with the original requested size (`tensor.nbytes`), which is not a huge-page multiple. Because `ShmHelper::allocate()` rounds the physical mapping up to hugepage granularity, the registered range ends inside the last huge page of the mapping — exactly where the kernel refuses to split the VMA.

### Bug 2: preTouch splits large buffers into fixed 4 KB blocks (`memory_location.cpp` / `rdma_transport.cpp`)

For buffers ≥ 4 GiB, `preTouchMemory()` splits the buffer into per-thread blocks (`length / num_threads`, 4 KB granular) and trial-registers each block on the RDMA path. Any block boundary that is not hugepage-aligned triggers the same EINVAL, even when the buffer as a whole is properly aligned.

Bug 1 controls what length is registered; Bug 2 controls the intermediate ranges touched during preTouch. Both must respect hugepage boundaries, and both must agree with the physical mapping produced by `ShmHelper`.

### Testing

- New end-to-end suite `real_client_hugepage_register_test` (4 tests, real client + in-proc master, `MC_STORE_USE_HUGEPAGE=1`). Run with `PROTOCOL=rdma DEVICE_NAME=mlx5_0` for full discriminating power:

  | Test | Discriminating input | Unfixed | Fixed |
  |---|---|---|---|
  | SegmentBaseRegistersFullSegment | 8MB segment, base registered with logical size 3MB+512 | EINVAL on rdma | succeeds (widened to 8MB) |
  | PreTouchPageAlignedBlocksOnLargeSegment | 4GiB+1MB → 2049-page segment, not divisible by 8/16 threads | preTouch trial-registration EINVAL | succeeds with page-aligned blocks |
  | HotCacheShmRegistersAlignedSegment | 33MB cache / 3MB blocks (not a 2MB multiple) | hot cache silently disabled (EINVAL) | `IsHotCacheEnabled()` true |
  | MisalignedInteriorSubRangeRegistersExactly | interior ptr base+1MB, len 3MB+512 | EINVAL | still fails on rdma (by design); succeeds on tcp |

  Note: the last test intentionally stays last — a deliberately failed `ibv_reg_mr` leaves residue in libibverbs' process-global fork-protection bookkeeping (rdma-core does not roll back DONTFORK on failure), which can poison the next registration in the same process.
- `client_local_hot_cache_test` extended with 4 block-bookkeeping cases (39 total).
- Full-store verification on an mlx5_0 host: 4098 MB and 24.3 GB buffers now register reliably (80/80 passes). Suite results: hugepage suite 4/4 PASSED (rdma), hot cache suite 39/39 PASSED.

```bash
cmake -B build -S . -DCMAKE_BUILD_TYPE=Release -DUSE_RDMA=ON \
      -DUSE_HTTP=ON -DWITH_STORE=ON -DUSE_NVMEOF=OFF -DUSE_SPDK=OFF
cmake --build build --target real_client_hugepage_register_test \
      client_local_hot_cache_test --parallel 32
PROTOCOL=rdma DEVICE_NAME=mlx5_0 \
  ./build/mooncake-store/tests/real_client_hugepage_register_test
./build/mooncake-store/tests/client_local_hot_cache_test
```

Requires `vm.nr_hugepages > 0` (the suite SKIPs otherwise). Pass criterion: gtest prints `[  PASSED  ] N tests.`; the widening is also directly visible in the log line `Registering HugeTLB shared-memory segment: base=…, segment_size=…, logical_size=…`.

## Module

- [*] Mooncake Store (`mooncake-store`)

## Type of Change

- [*] Bug fix

## Checklist

- [*] I have performed a self-review of my own code

## AI Assistance Disclosure

- [*] AI tools were used (specify below)

Claude (Kiro) assisted with writing the code changes and PR description. Changes were tested by the submitter on v0.3.12.post1 and manually reviewed before submission.


catyans:
Thanks for tracking the allocated size through registration. One edge case: `get_shm(buffer)` returns the segment containing `buffer`, not only an exact base-address match. If a caller registers `shm->base_addr + offset`, this keeps the interior start address but replaces the requested length with the full `shm->size`, so the MR extends past the mapping by `offset` and can still fail registration. Could we use the padded size only when `buffer == shm->base_addr` (or register from the base with matching unregister bookkeeping), and add an interior-subrange regression test?

1998zxn:
> Thanks for tracking the allocated size through registration. One edge case: `get_shm(buffer)` returns the segment containing `buffer`, not only an exact base-address match. If a caller registers `shm->base_addr + offset`, this keeps the interior start address but replaces the requested length with the full `shm->size`, so the MR extends past the mapping by `offset` and can still fail registration. Could we use the padded size only when `buffer == shm->base_addr` (or register from the base with matching unregister bookkeeping), and add an interior-subrange regression test?

Thank you. This is a great suggestion. Although I haven't noticed this kind of edge situation so far, I think adding `buffer == shm->base_addr` as a criterion for judgment is a good suggestion, which can focus more on this modification.

Icedcoco:
Thanks for the fix, and also for addressing the interior-pointer case with `buffer == shm->base_addr`.

While tracing the other allocation and registration paths, I found a few points that may be worth discussing before merging. I may be missing some context, so please correct me if any of these assumptions are inaccurate.

1. `LocalHotCache` may still register the logical size rather than the padded allocation size.

    `ShmHelper::allocate()` now rounds the mapping size up to `16 * hugepage_size`, but `LocalHotCache` keeps `bulk_memory_size_` based on the original logical cache size. `Client::InitLocalHotCache()` then passes `GetTotalSize()` directly to `transfer_engine_->registerLocalMemory()`.

    For example, with 2 MB hugepages, a 48 MB logical cache would result in a 64 MB mapping but still register 48 MB. Would it make sense for the hot cache registration path to use `GetShmSegment()->size` (or otherwise track the actual mapped size) when SHM and hugepages are enabled?

2. Could the new allocation size also affect the `DummyClient` path?

    `DummyClient::register_buffer()` still aligns the caller-provided size to one hugepage and then compares it with `shm->size`. With the new allocation rule, something like `alloc_from_mem_pool(1 MB)` would create a 32 MB SHM segment, while `register_buffer(ptr, 1 MB)` would only be adjusted to 2 MB and fail the exact-size check.

    Should this path also use the recorded `ShmSegment::size`, or should the API explicitly require callers to pass the padded size? It might be useful to add a regression test for `alloc_from_mem_pool()` followed by `register_buffer()`.

3. Could we also verify the interaction with MR chunking?

    `RdmaTransport::registerLocalMemory()` may split a large buffer according to `max_mr_size`, but the resulting `chunk_len` is not necessarily aligned to `16 * hugepage_size`. If `max_mr_size` is not itself a 32 MB multiple for 2 MB hugepages, a later chunk could still be passed to `ibv_reg_mr()` with an unaligned length.

    Should the chunk limit be rounded down to the required alignment for hugepage-backed buffers, or is there an existing guarantee that the device limit is always aligned this way?

4. `ShmHelper` is transport-agnostic, but the new 16-page alignment is mlx5/RDMA-specific.

    The helper is also used by `MooncakeHostMemAllocator`, `DummyClient`, and SHM-based `LocalHotCache`, including configurations that may use TCP or other transports. With 1 GB hugepages, even a small allocation would be rounded up to 16 GB.

    Would it be safer to keep this policy in the affected RDMA registration path, or should we document and accept this larger allocation granularity for all hugepage-backed SHM users?

Could we add focused coverage for the hot-cache path, the dummy-client path, and a non-aligned MR chunk limit? That would help verify that the physical allocation size, the registered MR size, and the logical buffer size remain consistent across all callers.

I also noticed that the current Ascend CI job is red because the fork checkout was rejected by the `pull_request_target` security rule. Could this job be rerun with the appropriate safe-checkout configuration, or otherwise be noted as an infrastructure/check-out issue?

1998zxn:
> Thanks for the fix, and also for addressing the interior-pointer case with `buffer == shm->base_addr`.
> 
> While tracing the other allocation and registration paths, I found a few points that may be worth discussing before merging. I may be missing some context, so please correct me if any of these assumptions are inaccurate.
> 
> 1. `LocalHotCache` may still register the logical size rather than the padded allocation size.
>    `ShmHelper::allocate()` now rounds the mapping size up to `16 * hugepage_size`, but `LocalHotCache` keeps `bulk_memory_size_` based on the original logical cache size. `Client::InitLocalHotCache()` then passes `GetTotalSize()` directly to `transfer_engine_->registerLocalMemory()`.
>    For example, with 2 MB hugepages, a 48 MB logical cache would result in a 64 MB mapping but still register 48 MB. Would it make sense for the hot cache registration path to use `GetShmSegment()->size` (or otherwise track the actual mapped size) when SHM and hugepages are enabled?
> 2. Could the new allocation size also affect the `DummyClient` path?
>    `DummyClient::register_buffer()` still aligns the caller-provided size to one hugepage and then compares it with `shm->size`. With the new allocation rule, something like `alloc_from_mem_pool(1 MB)` would create a 32 MB SHM segment, while `register_buffer(ptr, 1 MB)` would only be adjusted to 2 MB and fail the exact-size check.
>    Should this path also use the recorded `ShmSegment::size`, or should the API explicitly require callers to pass the padded size? It might be useful to add a regression test for `alloc_from_mem_pool()` followed by `register_buffer()`.
> 3. Could we also verify the interaction with MR chunking?
>    `RdmaTransport::registerLocalMemory()` may split a large buffer according to `max_mr_size`, but the resulting `chunk_len` is not necessarily aligned to `16 * hugepage_size`. If `max_mr_size` is not itself a 32 MB multiple for 2 MB hugepages, a later chunk could still be passed to `ibv_reg_mr()` with an unaligned length.
>    Should the chunk limit be rounded down to the required alignment for hugepage-backed buffers, or is there an existing guarantee that the device limit is always aligned this way?
> 4. `ShmHelper` is transport-agnostic, but the new 16-page alignment is mlx5/RDMA-specific.
>    The helper is also used by `MooncakeHostMemAllocator`, `DummyClient`, and SHM-based `LocalHotCache`, including configurations that may use TCP or other transports. With 1 GB hugepages, even a small allocation would be rounded up to 16 GB.
>    Would it be safer to keep this policy in the affected RDMA registration path, or should we document and accept this larger allocation granularity for all hugepage-backed SHM users?
> 
> Could we add focused coverage for the hot-cache path, the dummy-client path, and a non-aligned MR chunk limit? That would help verify that the physical allocation size, the registered MR size, and the logical buffer size remain consistent across all callers.
> 
> I also noticed that the current Ascend CI job is red because the fork checkout was rejected by the `pull_request_target` security rule. Could this job be rerun with the appropriate safe-checkout configuration, or otherwise be noted as an infrastructure/check-out issue?

Thanks for all the detailed review comments, they're all critical points. I'll address them as soon as possible, including auditing for any other ShmHelper usage paths that may have been missed.

For the Ascend CI job: I think I don't have permission to re-run it with a different checkout config, and the failure is documented as an infrastructure/check-out issue in the PR description.

ykwd:
Re-ran the Ascend CI job

ykwd:
For the following failed CI, we may need to rebase to the latest main branch: [Build & Test (Linux) / test-wheel-ubuntu (ubuntu-22.04, 3.12) (pull_request)](https://github.com/kvcache-ai/Mooncake/actions/runs/33737926451/job/100676476789?pr=3845)

1998zxn:
> For the following failed CI, we may need to rebase to the latest main branch: [Build & Test (Linux) / test-wheel-ubuntu (ubuntu-22.04, 3.12) (pull_request)](https://github.com/kvcache-ai/Mooncake/actions/runs/33737926451/job/100676476789?pr=3845)

sure, We have further explored this issue and addressed it from a more fundamental perspective. We will be rebasing while supplementing the code later today.

1998zxn:
> Thanks for the fix, and also for addressing the interior-pointer case with `buffer == shm->base_addr`.
> 
> While tracing the other allocation and registration paths, I found a few points that may be worth discussing before merging. I may be missing some context, so please correct me if any of these assumptions are inaccurate.
> 
> 1. `LocalHotCache` may still register the logical size rather than the padded allocation size.
>    `ShmHelper::allocate()` now rounds the mapping size up to `16 * hugepage_size`, but `LocalHotCache` keeps `bulk_memory_size_` based on the original logical cache size. `Client::InitLocalHotCache()` then passes `GetTotalSize()` directly to `transfer_engine_->registerLocalMemory()`.
>    For example, with 2 MB hugepages, a 48 MB logical cache would result in a 64 MB mapping but still register 48 MB. Would it make sense for the hot cache registration path to use `GetShmSegment()->size` (or otherwise track the actual mapped size) when SHM and hugepages are enabled?
> 2. Could the new allocation size also affect the `DummyClient` path?
>    `DummyClient::register_buffer()` still aligns the caller-provided size to one hugepage and then compares it with `shm->size`. With the new allocation rule, something like `alloc_from_mem_pool(1 MB)` would create a 32 MB SHM segment, while `register_buffer(ptr, 1 MB)` would only be adjusted to 2 MB and fail the exact-size check.
>    Should this path also use the recorded `ShmSegment::size`, or should the API explicitly require callers to pass the padded size? It might be useful to add a regression test for `alloc_from_mem_pool()` followed by `register_buffer()`.
> 3. Could we also verify the interaction with MR chunking?
>    `RdmaTransport::registerLocalMemory()` may split a large buffer according to `max_mr_size`, but the resulting `chunk_len` is not necessarily aligned to `16 * hugepage_size`. If `max_mr_size` is not itself a 32 MB multiple for 2 MB hugepages, a later chunk could still be passed to `ibv_reg_mr()` with an unaligned length.
>    Should the chunk limit be rounded down to the required alignment for hugepage-backed buffers, or is there an existing guarantee that the device limit is always aligned this way?
> 4. `ShmHelper` is transport-agnostic, but the new 16-page alignment is mlx5/RDMA-specific.
>    The helper is also used by `MooncakeHostMemAllocator`, `DummyClient`, and SHM-based `LocalHotCache`, including configurations that may use TCP or other transports. With 1 GB hugepages, even a small allocation would be rounded up to 16 GB.
>    Would it be safer to keep this policy in the affected RDMA registration path, or should we document and accept this larger allocation granularity for all hugepage-backed SHM users?
> 
> Could we add focused coverage for the hot-cache path, the dummy-client path, and a non-aligned MR chunk limit? That would help verify that the physical allocation size, the registered MR size, and the logical buffer size remain consistent across all callers.
> 
> I also noticed that the current Ascend CI job is red because the fork checkout was rejected by the `pull_request_target` security rule. Could this job be rerun with the appropriate safe-checkout configuration, or otherwise be noted as an infrastructure/check-out issue?

Digging one level deeper, we found the underlying mechanism: after `ibv_fork_init()` is enabled, every `ibv_reg_mr` first calls `madvise(MADV_DONTFORK)` for fork protection, and the kernel refuses to split a hugetlb VMA at a non-huge-page-aligned boundary. Knowing this, a more elegant approach became possible: instead of reshaping the allocation, make the registration agree with the mapping. The updated description above covers the details.
We also checked every other HugeTLB registration path for the same problem:
- `DummyClient` — not affected. It always registers the exact whole segment, so it can never hit this.
- `LocalHotCache` (SHM mode) — had the same problem; now fixed to register the full segment size instead of the logical cache size.
Regression tests have been added: `real_client_hugepage_register_test` (4 tests) and 4 new cases in `client_local_hot_cache_test` (39 total).

Icedcoco:
Thanks for the update and for tracing the failure to the HugeTLB/DONTFORK interaction. Keeping the existing allocation granularity and making registration respect the mapping boundaries makes sense. The hot-cache change also addresses the earlier concern.

Looking through 379eb7f5, I noticed a few remaining cases that may be worth discussing:

1. Alignment after MR chunking

   The outer registration path still splits buffers directly at max_mr_size, which may not be a hugepage multiple. The new pre-touch logic then adds length % page_size to the last thread’s block, potentially leaving its end unaligned.

   For example, with 2 MiB hugepages, a 4 GiB + 2 MiB buffer and MC_MAX_MR_SIZE=4293918720 (4 GiB − 1 MiB), the first chunk ends halfway through a hugepage. This appears to retain the original failure under the DONTFORK conditions described in the PR.

   Could we align the outer MR chunk boundaries as well and add a regression case with a non-aligned MR limit? This looks like an existing path the fix does not yet cover, rather than a new regression.

2. The interior-subrange test assumes RDMA registration must fail

   MisalignedInteriorSubRangeRegistersExactly currently expects failure whenever PROTOCOL=rdma. However, libibverbs can report IBV_FORK_UNNEEDED, and RDMAV_HUGEPAGES_SAFE can also change the alignment handling. In those environments, successful registration would not necessarily mean the range was widened.

   Could we verify that the original address and length reach the registration layer unchanged, and make the EINVAL expectation conditional on the relevant fork-protection behavior? It would also be good to unregister whenever registration succeeds; the current RDMA branch skips that cleanup.

3. Zero-length registration now bypasses validation

   For a HugeTLB segment’s base address, size == 0 satisfies size < shm->size, so it is expanded to the full segment before reaching CheckRegisterMemoryParams(). Previously, that check rejected zero-length requests. The bookkeeping would still record the original zero length.

   Could we preserve the zero-length rejection before widening the registration, with a small regression test?

For verification context, I checked the page-size detection and the kernel’s alignment behavior using a HugeTLB mapping, but have not rerun the full RDMA suite. The current CTest job also appears to have stopped because the runner received a shutdown signal (exit 143), so I would not attribute that failure to this change.

1998zxn:
> Thanks for the update and for tracing the failure to the HugeTLB/DONTFORK interaction. Keeping the existing allocation granularity and making registration respect the mapping boundaries makes sense. The hot-cache change also addresses the earlier concern.
> 
> Looking through [379eb7f](https://github.com/kvcache-ai/Mooncake/commit/379eb7f5fb4405e0d8ba36968fa2cfd152a46886), I noticed a few remaining cases that may be worth discussing:
> 
> 1. Alignment after MR chunking
>    The outer registration path still splits buffers directly at max_mr_size, which may not be a hugepage multiple. The new pre-touch logic then adds length % page_size to the last thread’s block, potentially leaving its end unaligned.
>    For example, with 2 MiB hugepages, a 4 GiB + 2 MiB buffer and MC_MAX_MR_SIZE=4293918720 (4 GiB − 1 MiB), the first chunk ends halfway through a hugepage. This appears to retain the original failure under the DONTFORK conditions described in the PR.
>    Could we align the outer MR chunk boundaries as well and add a regression case with a non-aligned MR limit? This looks like an existing path the fix does not yet cover, rather than a new regression.
> 2. The interior-subrange test assumes RDMA registration must fail
>    MisalignedInteriorSubRangeRegistersExactly currently expects failure whenever PROTOCOL=rdma. However, libibverbs can report IBV_FORK_UNNEEDED, and RDMAV_HUGEPAGES_SAFE can also change the alignment handling. In those environments, successful registration would not necessarily mean the range was widened.
>    Could we verify that the original address and length reach the registration layer unchanged, and make the EINVAL expectation conditional on the relevant fork-protection behavior? It would also be good to unregister whenever registration succeeds; the current RDMA branch skips that cleanup.
> 3. Zero-length registration now bypasses validation
>    For a HugeTLB segment’s base address, size == 0 satisfies size < shm->size, so it is expanded to the full segment before reaching CheckRegisterMemoryParams(). Previously, that check rejected zero-length requests. The bookkeeping would still record the original zero length.
>    Could we preserve the zero-length rejection before widening the registration, with a small regression test?
> 
> For verification context, I checked the page-size detection and the kernel’s alignment behavior using a HugeTLB mapping, but have not rerun the full RDMA suite. The current CTest job also appears to have stopped because the runner received a shutdown signal (exit 143), so I would not attribute that failure to this change.

Thanks for the review, and for validating the page-size detection — all three make sense.

(2) Interior sub-range test. Right, the EINVAL check is environment-dependent. What matters is that the interior pointer isn't silently widened, not that registration fails — will accept success too (and unregister). Fixing here.

(3) Zero-length registration. Good catch. size == 0 gets widened before the length == 0 check; will add size > 0 plus a test. Fixing here.

(1) Outer MR chunk alignment. Agreed. For hugepage buffers, will round the chunk limit down to the hugepage size before splitting, and add a test with a non-aligned MC_MAX_MR_SIZE.

One question here: if MC_MAX_MR_SIZE is set below one hugepage (e.g. 1 MiB with 2 MiB pages), rounding down gives 0, and a hugepage buffer can't be split below one hugepage anyway. My plan is to round up to one hugepage and log a warning — it's already capped to the device max, so no truncation, and the workload keeps running instead of failing on a bad config. Does that sound reasonable to you, or would you prefer we reject the registration in that case? Happy to go either way.

Also agreed the CTest failure is the runner shutdown (exit 143), not this change. Full suite passes locally on mlx5_0 (hugepage 4/4, hot cache 39/39); will re-trigger CI once these land.

Icedcoco:
Thanks for explaining the edge case. My preference would be to round the MR chunk limit down to the hugepage size and return a clear registration error if the result is zero.

For example, with a 1 MiB effective MR limit and 2 MiB hugepages, we cannot create a page-aligned chunk within that limit. Raising only the local chunk limit to 2 MiB would still hit the globalConfig().max_mr_size check in RdmaContext::registerMemoryRegionInternal(). Raising the effective limit itself would need additional care, since it incorporates both configuration and device constraints; being capped before rounding does not guarantee that the rounded value remains within the device limit.

For this PR, preserving the existing limit and reporting the conflict seems like the simplest consistent approach. The error could include both values and suggest checking MC_MAX_MR_SIZE against the device capability and hugepage configuration.

That said, I’m not very familiar with this module or all the RDMA registration details, so I may be missing an existing convention or compatibility consideration. I’d appreciate input from developers who know this path better, particularly on whether this limit is intended to be strict or whether automatically relaxing a user-configured limit is an established behavior.

mooncake-transfer-engine/src/transport/rdma_transport/rdma_transport.cpp
alogfans, line 1: Could we round `chunk_limit` down to `detectBufferPageSize(addr)` before splitting the buffer? Otherwise, a non-page-aligned effective `max_mr_size` can still produce a HugeTLB-unaligned chunk boundary and cause the same `MADV_DONTFORK` / `EINVAL` failure during pre-touch or final MR registration.

1998zxn: Good point! Rounding down chunk_limit before splitting solves this cleanly for both preTouch and MR registration at once, as they stem from the exact same alignment issue.

Looking at the logic in rdma_transport.cpp:418, since preTouchMemory is called per-chunk post-split, aligning chunk_limit upstream automatically guarantees that no misaligned boundaries reach madvise.

So fixing chunk_limit up front handles everything, and no changes inside preTouch are needed. Does that sound good to you?

1998zxn:
> Thanks for explaining the edge case. My preference would be to round the MR chunk limit down to the hugepage size and return a clear registration error if the result is zero.
> 
> For example, with a 1 MiB effective MR limit and 2 MiB hugepages, we cannot create a page-aligned chunk within that limit. Raising only the local chunk limit to 2 MiB would still hit the globalConfig().max_mr_size check in RdmaContext::registerMemoryRegionInternal(). Raising the effective limit itself would need additional care, since it incorporates both configuration and device constraints; being capped before rounding does not guarantee that the rounded value remains within the device limit.
> 
> For this PR, preserving the existing limit and reporting the conflict seems like the simplest consistent approach. The error could include both values and suggest checking MC_MAX_MR_SIZE against the device capability and hugepage configuration.
> 
> That said, I’m not very familiar with this module or all the RDMA registration details, so I may be missing an existing convention or compatibility consideration. I’d appreciate input from developers who know this path better, particularly on whether this limit is intended to be strict or whether automatically relaxing a user-configured limit is an established behavior.

Thanks for the great suggestion!

I rounded chunk_limit down to the buffer's page size prior to splitting and added an ERR_INVALID_ARGUMENT check if the limit falls below one page. A corresponding test has also been added.

1998zxn:
Hi @alogfans @Icedcoco @ykwd @catyans — I think every point from the last rounds is now addressed and the branch is green (mergeable, CI clean): Local suite test on an mlx5_0 host: hugepage 4/4, hot cache 39/39.

@alogfans because `preTouchMemory` runs per-chunk after the split, rounding `chunk_limit` down up front keeps every chunk (and therefore every pre-touch block) page-aligned, so no separate change inside preTouch is needed. Does that match your expectation, or would you prefer the alignment also be applied inside preTouch? Happy to adjust either way.

Otherwise, glad to hear any remaining suggestions before merge — thanks for the thorough reviews!

Icedcoco:
Thanks for the update. Rounding chunk_limit down before splitting looks consistent with alogfans’s suggestion, and the new check for limits smaller than one page makes the conflict explicit.

I had one question about the alignment check at rdma_transport.cpp:327. It appears to run even when the buffer is smaller than chunk_limit and no splitting is needed. For example, with a regular 4 KiB-page buffer, max_mr_size set to 1 KiB, and a 512-byte registration, the aligned limit becomes zero and the registration is rejected, although the requested MR is within the configured limit.

Could we limit this alignment check to cases where chunking is actually needed and page-aligned boundaries are required? A small test for a sub-page registration on a regular-page mapping might help confirm the intended behavior. I may be missing a constraint in the RDMA path, so I’d appreciate your thoughts.

1998zxn:
> Thanks for the update. Rounding chunk_limit down before splitting looks consistent with alogfans’s suggestion, and the new check for limits smaller than one page makes the conflict explicit.
> 
> I had one question about the alignment check at rdma_transport.cpp:327. It appears to run even when the buffer is smaller than chunk_limit and no splitting is needed. For example, with a regular 4 KiB-page buffer, max_mr_size set to 1 KiB, and a 512-byte registration, the aligned limit becomes zero and the registration is rejected, although the requested MR is within the configured limit.
> 
> Could we limit this alignment check to cases where chunking is actually needed and page-aligned boundaries are required? A small test for a sub-page registration on a regular-page mapping might help confirm the intended behavior. I may be missing a constraint in the RDMA path, so I’d appreciate your thoughts.

Thanks for the review. Addressed in the latest commit:

The chunk-limit round-down and the sub-page rejection are now confined to HugeTLB-backed buffers. The gate changed from buffer_page_size > 0 (always true, so it also fired on regular 4 KiB pages) to buffer_page_size > static_cast<size_t>(pagesize), i.e. it only triggers when the buffer's KernelPageSize is larger than the base page — exactly the HugeTLB case that hits the MADV_DONTFORK EINVAL when a VMA is split at a non-hugepage boundary.

Regular pages (and THP, which reports the 4 KiB base page) are untouched: the kernel splits ordinary VMAs at any boundary, so the previous code was (a) falsely rejecting a registration when MC_MAX_MR_SIZE was below one base page, and (b) forcing needless extra chunking. Both are gone now.

Regression tests, verified on RDMA (mlx5_0):

- Hugepage (real_client_hugepage_register_test): SubHugepageMaxMrSizeIsRejected (2 MiB page / max_mr_size=1 MiB / register 0.5 MiB → rejected), MisalignedMaxMrSizeChunksStayPageAligned (5 MiB → aligned down to 4 MiB → OK), 7/7 pass.
- Regular page (new real_client_regular_page_register_test): SubPageMaxMrSizeNotRejected (4 KiB page / max_mr_size=1024 → not rejected, splits into 8 chunks), NonPageMultipleMaxMrSizeRegisters (max_mr_size=5000 → OK), 2/2 pass.

Issue #3808, opened 2026-09-01
[RoadMap]: Mooncake Store HA and OpLog Roadmap


Icedcoco:
**Last reviewed:** 2026-08-31
**Code baseline:** [`main` at `73471dc28`](https://github.com/kvcache-ai/Mooncake/commit/73471dc28)
**Scope:** durable batch OpLog, HA leadership, standby recovery, snapshot migration, replica lifecycle, entity ownership, observability, codecs, and conditional Segment persistence work.

This is the roadmap for the work tracked under [`oplog-ha-prs`](../specs/oplog-ha-prs/README.md). It consolidates the PR-level specifications and the earlier RFC direction into one delivery view. It is broader than the snapshot implementation: snapshots are the recovery and retention track, while writer correctness, fencing, lifecycle, identity, ownership, metrics, and codec work are independent but connected workstreams.

## End state

Mooncake Store should converge on one production HA state model:

- a fenced, ordered batch OpLog is the durable mutation history;
- standby masters apply that history and periodically publish immutable materialized snapshots;
- a new or promoted node restores a verified snapshot and replays its suffix before serving;
- replica lifecycle events, resource ownership, Segment state, and access recency have stable identities and deterministic replay semantics;
- OpLog and snapshot growth are bounded by a reader-visible compaction floor and tested etcd operations;
- the legacy primary snapshot writer and other transitional writers are removed after rollout evidence, not merely after compilation.

The program does not introduce multiple authorities for the same state. During migration, old readers may remain for explicit import or historical compatibility, but production selection is explicit and never based on bucket/catalog listing.

## Program invariants

1. **Durability precedes visibility.** A mutation or lifecycle transition is externally visible only after its required durable record is committed. No lock is held across an asynchronous durable wait.
2. **Fencing is fail-closed.** A stale producer view cannot append a batch. Fencing, corruption, and non-retryable storage failures stop admission and eventually remove serving, rather than silently falling back to an unfenced writer.
3. **Recovery is deterministic.** Restore validates identity, cursor, checksum, ordering, and resource descriptors. Unknown history, malformed state, or infrastructure failures never become an empty-cluster success.
4. **One identity follows an entity.** `ObjectIdentity`, `ReplicaID`, and the future `created_at` generation field are reused by asynchronous callbacks, lifecycle records, restore, promotion, and eviction. No global registry or operation-ID framework is required for the first version.
[...]

PR #3860, opened 2026-09-03
[Store] Stabilize HA under OpLog pressure

tensorcopy:
Related to #3808

## Why

Sustained OpLog pressure exposed two independent HA failures. OpLog queue
saturation could request repeated eviction passes. The leader also shared one
etcd client with Store traffic, so resetting that client closed the leadership
keep-alive stream.

## What

- Stop retrying a full eviction pass after OpLog queue saturation.
- Give each leadership lease keep-alive its own etcd client.
- Keep active lease streams alive when the general Store etcd client resets.
- Update HA tests to require reset isolation.

## How

Batch eviction consumes one pending allocation-pressure request. It restores
that request only when the pass made no progress, ended without an OpLog error,
and still found evictable objects.

The Go etcd wrapper creates one client for each blocking lease keep-alive.
General Store reset still replaces the Store client and cancels its watches and
maintenance sessions. It does not cancel leadership keep-alive. Explicit lease
cancel and coordinator shutdown remain the normal stop paths.

## Test

- The reset-isolation regression failed before the fix because Store client
  reset closed the keep-alive channel.
- The two focused leadership tests passed 20 consecutive runs.
- The full etcd HA test binary passed all 14 tests.
- Go tests with the race detector passed.
- The queue-saturation regression passed 20 consecutive runs.
- Eight related eviction tests passed.
- `MasterServiceTest.PutStartExpiringTest` passed.
- One full HA test run passed all 93 tests.
- Pre-commit and `git diff --check` passed.

A sustained downstream deployment test confirmed that the target cache
collapse did not recur. Active-master deletion promoted the standby, and the
replacement standby caught up. Environment, image, traffic, hardware, and
performance details are intentionally omitted.

The Ubuntu 22.04 wheel job failed in
`test_ssd_offload_in_evict.py::test_concurrent_stress`. The unchanged base
commit failed in the same job and test. All other CI jobs passed.

This PR does not address the separate durable mutation barrier in #3808. It
must not be treated as the complete production-readiness gate for that roadmap.

## Ablation

- Existing shared-client behavior failed because reset closed the lease stream.
- A dedicated client alone failed because reset still canceled the lease
  context.
- Removing reset cancellation alone failed because closing the shared client
  still closed the stream.
- The combined lease design passed the reset and leadership checks.
- Always clearing the eviction request stopped the retry, but it failed
  `PutStartExpiringTest`. Conditional restore passed that test and the
  queue-saturation test.

## Module

- [x] Mooncake Store
- [x] Common

## Type of change

- [x] Bug fix

## Checklist

- [x] The related roadmap issue is #3808

## AI assistance disclosure

- [x] AI tools were used. Codex helped diagnose the failure, implement the
  change, and run tests. The human submitter must review every changed line
  before merge.

### DIFF
--- a/mooncake-common/etcd/etcd_wrapper.go
+++ b/mooncake-common/etcd/etcd_wrapper.go
@@ -324,7 +324,6 @@ func EtcdStoreResetClientWrapper(endpoints *C.char, errMsg **C.char) int {
 		return -1
 	}
 
-	cancelAllStoreKeepAlives()
 	closeAllStoreMaintenanceSessions()
 	cancelAllStoreWatches()
 	cancelAllStorePrefixWatches()
@@ -593,20 +592,6 @@ func cancelAndDeleteWatch(k string) int {
 	return -1
 }
 
-func cancelAllStoreKeepAlives() {
-	storeKeepAliveMutex.Lock()
-	cancels := make([]context.CancelFunc, 0, len(storeKeepAliveCtx))
-	for leaseId, cancel := range storeKeepAliveCtx {
-		cancels = append(cancels, cancel)
-		delete(storeKeepAliveCtx, leaseId)
-	}
-	storeKeepAliveMutex.Unlock()
-
-	for _, cancel := range cancels {
-		cancel()
-	}
-}
-
 func closeAllStoreMaintenanceSessions() {
 	storeMaintenanceMutex.Lock()
 	sessions := make([]*maintenanceSession, 0, len(storeMaintenanceSessions))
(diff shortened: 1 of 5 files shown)


Icedcoco (approved):
LGTM

Icedcoco (approved):
LGTM from my side. The HA behavior is correct: resetting the general Store etcd client no longer drops a healthy leadership lease, while an actual keep-alive loss still stops serving. The eviction change also removes the self-sustaining retry request when the OpLog reservation queue is saturated.

Please rebase or merge the latest `main` and rerun CI. The current Ubuntu 22.04 wheel failure needs a fresh result against the updated base. Once CI is green, this can be merged.

This is a useful stability follow-up for #3808, but it does not replace the roadmap's writer fail-stop or durable-mutation gates.

ykwd (approved):
Thanks for the great work

PR #3861, opened 2026-09-04
[TENT] Slice large HP TCP reads across rails

jacklin78911-collab:
## Description

Build on the static lane-to-rail routing introduced by #3834 and add minimal per-transfer slicing for large HP TCP READs.

A READ is split into one contiguous slice per configured rail when each rail receives at least one 1 MiB I/O step. The slices reuse the existing persistent lanes and complete as one logical TENT task.

Small READs, all WRITEs, and single-rail configurations keep the existing single-operation path. This change adds no new configuration, dynamic scheduling, hashing policy, failover, or WRITE striping.

## Module

- [ ] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] Refactor
- [ ] Breaking change
- [x] Documentation update
- [x] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Focused local validation:**

```bash
git diff --check
clang-format-20 --dry-run --Werror <changed C++ files>
cmake --build <build-dir> --target tent_hp_tcp_transport_test tent_hp_tcp_workers_test
ctest --test-dir <build-dir> --output-on-failure -R '^tent_hp_tcp_(workers|transport)_test$'
```

Both focused test targets built successfully and both tests passed.

**Dual-host benchmark:**

Two hosts were connected through independently routed `eth1` and `eth2` interfaces, each reporting 400 Gb/s. Both cases used the same Release binary and workload:

- 64 MiB READ
- batch size 1
- one benchmark thread
- 16 HP TCP workers
- 4 persistent lanes per peer
- 10-second measurement

| Configuration | Throughput | Avg transfer latency | P99 |
|---|---:|---:|---:|
| 1 rail | 2.526 GB/s | 26.56 ms | 35.33 ms |
| 2 rails | 4.917 GB/s | 13.65 ms | 17.72 ms |

The two-rail configuration improved single-transfer throughput by 94.6%.

Payload correctness was verified separately with `tebench --op_type=mix --check_consistency=true` using 4 MiB transfers over two rails. The full-buffer comparison completed without mismatch.

**Test results:**

- [x] Unit tests pass
- [x] Integration tests pass
- [x] Manual dual-host testing done

### Follow-up Store validation — 2026-09-11

On two H20 VMs, Store host-memory READs benefited from the existing static multi-rail implementation with **4 HP workers per endpoint and 4 total persistent lanes per peer** in both configurations. This follow-up changes no runtime mechanism or defaults.

Three interleaved repetitions per configuration, with a 2-second warmup and a 20-second timed phase. Values below are medians; throughput brackets contain the full minimum–maximum range. GB/s is decimal.

| Request / batch | Concurrency | HP one rail, GB/s [min, max] | HP two rails, GB/s [min, max] | Throughput change | Client CPU cores, one → two |
|---|---:|---:|---:|---:|---:|
| 8 MiB | 1 | 2.267 [2.237, 2.283] | 3.458 [3.429, 3.481] | 52.5% | 1.665 → 2.004 |
| 8 MiB | 4 | 6.518 [6.516, 6.542] | 10.838 [10.717, 10.915] | 66.3% | 5.098 → 6.739 |
| 4 KiB | 1 | 0.02706 [0.02655, 0.02721] | 0.02690 [0.02619, 0.02741] | -0.6% | 0.906 → 0.913 |
| 4 × 512 KiB | 4 | 6.251 [6.086, 6.298] | 6.394 [6.251, 6.408] | 2.3% | 5.570 → 5.576 |

For 8 MiB, client p99 fell from 4.863 to 3.076 ms at concurrency 1 and from 7.171 to 4.928 ms at concurrency 4. These are medians of each run's sampled quantile, not pooled quantiles. Client CPU increased by 20.3% and 32.2%, respectively, despite keeping the configured worker and lane counts fixed. CPU cores are client process CPU seconds divided by measured wall time; they exclude the holder, master and whole-host interrupt work. Small READs showed no material benefit, and the 4 × 512 KiB batch ranges overlap. Three repetitions do not establish statistical significance or universal scaling.

**Correctness and rail usage:**

- All 33 timed Store cases completed without request failures or misses. Each case first verified all 32 objects byte-for-byte; byte verification was disabled during timing.
- All 24 HP cases had exact payload-byte agreement between successful Store reads and per-process TCP_INFO deltas after subtracting protocol headers. All six dual-rail 8 MiB cases split payload 50% / 50% between the two rails.
- The #3987 / #3988 follow-up check used four logical peers, each with four lanes (two per rail): client connections went 16 → 0 → 16 across a 2-second idle threshold and reconnection. All 64 patterned READs before and after idle passed. The four server engines shared one server process; this checks connection behavior rather than multi-server throughput scaling.

**Workload and fixed conditions:**

- Reused the `mooncake-store/benchmarks/store_kv_bench.py` `get_into` / `batch_get_into` read path through a local validation driver for separate prepare, verify, warmup and timed phases. The dataset was synthetic resident KV in a remote 1 GiB holder segment. The client had no global segment, a 128 MiB local buffer, 32 objects, I/O depth 1 per job, and 1 or 4 jobs.
- HP source-bound routes used one or two interfaces in NUMA 0. Both processes were allowed CPUs 0–89 in NUMA 0; this affinity is not a four-core CPU cap. RPC threads stayed at four. No system route or NIC tuning was changed. The H20 GPUs were unused.
- Ubuntu 22.04 / GCC 11 / Python 3.10, Release without LTO; CUDA and TENT metrics disabled. The entire matrix used the same binaries. Standard TCP reference cases were included in the 33-case total, but used a different outbound interface, so the performance claim above compares HP one rail with HP two rails.
- The three repetition orders were TCP → HP1 → HP2, HP2 → HP1 → TCP, and HP1 → TCP → HP2. The batch case compared HP1 / HP2 only. Latency was sampled every seven requests.

**Source provenance:** This is a later combined validation candidate, `ea02043fc9cf0f5d65c67d1d1ee8e4cbf0b2b05c`, based on `32c4b93880eeaa63606fcb0dee78ac38364bc27e` (already containing #3987 and #3988), with changes from #4001, #4002, #4005, #4006 and #4007 plus the two-line Python guard fix subsequently submitted as #4051. This is not a measurement of this PR's original head or unmodified upstream main. The candidate was held fixed between the one- and two-rail cases.

The results support large remote Store host-memory READs in this setup. They do not measure GPU-memory transfers, model-inference TTFT or production workload gains. The full raw archive remains local; all 34 raw result files (33 Store cases plus the multi-peer check) were matched to the controller results. Archive SHA-256: `4d9f13747ef353c2b531af061849c7ba4363483f0571e5b38441a78bf69f3e50`.

### Current upstream main validation — 2026-09-14

Retested **unmodified upstream main `202ad9c89e1814f0273c605cc12b3ccb1c8117c6`** on the two H20 VMs. The runtime code and defaults were unchanged for this validation. As above, both configurations used four HP workers per endpoint, four total lanes per peer, four RPC threads, the Store host-memory `get_into` / `batch_get_into` path, 32 resident objects, a 2-second warmup and a 20-second timed phase. Three interleaved repetitions compared HP1→HP2, HP2→HP1, then HP1→HP2 using the same binaries. GPUs were unused.

With both endpoints allowed CPUs 0–89:

| Request / batch | Concurrency | One rail, GB/s [min, max] | Two rails, GB/s [min, max] | Change | Client CPU cores, one → two |
|---|---:|---:|---:|---:|---:|
| 8 MiB | 1 | 2.206 [2.119, 2.207] | 3.293 [3.266, 3.334] | +49.3% | 1.675 → 2.014 |
| 8 MiB | 4 | 6.205 [6.136, 6.405] | 10.709 [10.313, 10.790] | +72.6% | 5.092 → 6.847 |
| 4 KiB | 1 | 0.02668 [0.02641, 0.02724] | 0.02666 [0.02649, 0.02678] | −0.1% | 0.913 → 0.915 |
| 4 × 512 KiB | 4 | 6.210 [6.158, 6.241] | 6.197 [6.018, 6.216] | −0.2% | 5.746 → 5.675 |

For 8 MiB, median sampled p99 was 5.071→3.601 ms at C1 and 7.738→5.096 ms at C4. Client CPU means process CPU seconds / wall seconds, excluding the holder, master and whole-host interrupt work.

**CPU budget boundary:** six cases restricted both endpoints to CPUs 0–3. The guest topology reports two cores with two SMT threads each. C4 8 MiB throughput was 3.653 [3.650, 3.689]→3.661 [3.646, 3.729] GB/s (+0.2%). A six-case follow-up kept four logical CPUs but spread them across four distinct guest core IDs (0,2,4,6, one thread each): 4.332 [4.095, 4.521]→4.249 [4.249, 4.345] GB/s (−1.9%). Neither restricted configuration showed a clear two-rail throughput benefit; ranges overlap. Guest core IDs do not establish exclusive physical-core allocation. The wide-affinity gain requires CPU headroom in this setup; fixed worker counts alone do not impose a four-CPU process budget.

**Validation:** all 14 relevant CTest targets and all 36 timed Store cases passed. Every case first verified all 32 objects byte-for-byte. All 36 HP TCP_INFO payload checks matched successful Store bytes, and all 12 dual-rail 8 MiB cases split payload equally. The four-logical-peer check again passed 64 patterned READs and connection retirement/reconnection 16→0→16. As above, these peers shared one server process.

All 37 raw result JSONs matched their controller records, and both endpoints' actual CPU affinity was checked. Raw archives remain local:

- Main matrix: `53f94913b8314ed0225aec5f6859eda44b47d0667fa351db362ead4f99d5611e`.
- Four-distinct-guest-core follow-up: `dce9eb1b6c41b0efbf68a9a36107c8130b165a233631155dccf0fa2f7960a849`.

These results support large remote Store CPU-memory READs under the measured conditions. Small and CPU-constrained workloads showed no clear throughput gain. This is not a cross-version regression comparison, a GPU-memory transfer measurement or an inference TTFT claim.

## Store buffer-layout replay (2026-09-16)

Completed 18 additional two-host cases on the same frozen, unmodified main `202ad9c89e1814f0273c605cc12b3ccb1c8117c6`, using the previously validated binaries. This replays host KV buffer shapes from the audited SGLang path, with a selected BF16/page64 DeepSeek-V3 object of 4,497,408 bytes (4.289 MiB). It is a Store API replay, not a model inference run.

Both endpoints used 4 workers, 4 total lanes per peer, affinity 0–89, C1 and 32 resident objects. Each point verified all 32 objects, warmed up for 2 seconds and measured for 20 seconds. Three interleaved repetitions used HP1→HP2, HP2→HP1, HP1→HP2.

| Destination layout | One rail GB/s [min,max] | Two rails GB/s [min,max] | Change | Client CPU cores, one → two | Sampled p99 ms, one → two |
|---|---:|---:|---:|---:|---:|
| Page-first, one contiguous buffer | 2.777 [2.715,2.785] | 4.621 [4.602,4.793] | +66.4% | 1.820 → 2.417 | 2.150 → 1.330 |
| Layer-first, 61 scattered 72 KiB buffers | 1.291 [1.286,1.294] | 1.274 [1.260,1.277] | −1.3% | 1.772 → 1.740 | 4.818 → 4.866 |
| Contiguous 61-buffer control | 2.669 [2.613,2.697] | 4.504 [4.496,4.569] | +68.8% | 1.809 → 2.352 | 2.274 → 1.308 |

Numbers are medians of three runs; GB/s is decimal. CPU covers the client process only. Latency samples one in seven requests. This used a different host pair from September 14, so absolute rates are not a cross-version comparison.

All 18 payload/wire-count checks passed, and every raw result matched its controller record. Page-first and contiguous multi-buffer emitted 1→2 wire READs per Store call; scattered layer-first emitted 61 with either rail count. Dual-rail contiguous payload split equally, and scattered payload differed by at most one 72 KiB request between rails. Thus the scattered workload did use both rails without a throughput benefit. All 36 endpoint configurations and CPU affinities were independently checked, as were the seven runtime file hashes on each host.

This confirms the benefit depends on the contiguous request shape, not just total object size. The contiguous multi-buffer control retains the API form while restoring adjacency/merging; the scattered layout changes both locality and request count, so the experiment does not isolate a single bottleneck. No packing, scheduling, split-threshold or other production change was made. No GPU transfer or TTFT improvement is claimed; changing application layout still needs its conversion costs evaluated.

Raw archive remains local, SHA256 `c9b817c3653d70ddaf2a220a7a7260d3afc8ae84bfd28d5f74313b3b6efdd9be`. Both endpoints were cleaned up and reservations released.

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used
- [x] AI tools were used


mooncake-transfer-engine/tent/src/transport/hp_tcp/hp_tcp_task.cpp
```diff
@@ -5,6 +5,37 @@
...
+    }
+
+    if (remaining_slices_.fetch_sub(1, std::memory_order_acq_rel) != 1) {
+        return false;
+    }
+
+    terminal = slice_status_.load(std::memory_order_acquire);
+    return completeOnce(terminal, terminal == COMPLETED ? reserved_bytes_ : 0);
+}
+
```
staryxchen, line 38: Now a failed slice cannot fail the task until every other rail finishes: planTask does not call cancelRequest when a slice fails, and operation.complete only aggregates. Should we take the “first error cancels the rest” way?

staryxchen (commented):
Please add a test cover the new failure/cancel/retry surface

jacklin78911-collab:
@staryxchen Thanks for the review. I added a focused regression test covering the new failure/cancel/retry path: one slice returns stale-registration, the active sibling slice is canceled, and the request succeeds after metadata refresh and retry.

catyans (approved):
LGTM — I enjoyed reviewing this one. The change adds useful multi-rail bandwidth without making the design harder to follow: a large READ gets one clean, contiguous slice per rail, while WRITEs deliberately stay on a single rail to avoid ambiguous partial-write recovery. The shared completion and cancellation path also keeps the new lifecycle nicely contained.

Thanks for following up on @staryxchen's review with the stale-registration/cancel/retry test. That was the part I was most curious about, and seeing a failed slice cancel its sibling and then succeed after a metadata refresh makes the behavior much easier to trust.

I see the TENT checks are green as well. The remaining wheel failure comes from the unrelated Store SSD concurrent-read test, so it doesn't change my view here. Nice work!

mooncake-transfer-engine/tent/tests/hp_tcp_transport_test.cpp
```diff
@@ -328,17 +460,29 @@ TEST(HighPerformanceTcpTransportTest,
...
+    if (!entered) release_blocker.store(true);
+    ASSERT_TRUE(entered);
+
+    const Status submitted = client.submitTransferTasks(batch, {request});
+    if (!submitted.ok()) release_blocker.store(true);
+    ASSERT_TRUE(submitted.ok()) << submitted.ToString();
+    const bool sibling_active = WaitUntil([&] {
+        return HighPerformanceTcpTransportTestPeer::connectionsCreated(
+                   client) == 1;
+    });
```
staryxchen: `blockLane()` busy-loops an ASIO worker, so this test only works if lane 1 is on the other thread. `worker_count` is 2 and that split is never asserted, so a same-worker hash (or a 2-core CI box starving the sibling connect) fails at `ASSERT_TRUE(sibling_active)` without exercising cancel/retry.
Please don't block an ASIO worker to pause a slice. A dedicated stalled acceptor thread (as in the earlier SliceFailureCancelsSiblingRead test) or a cancelable timer/strand gate would pin the "stale slice fails while sibling is in-flight" timing without depending on affinityOwner or a busy-loop. If we keep two workers, assert they are distinct before submitting.

jacklin78911-collab: Thanks for catching this. I removed the ASIO worker busy-loop and the affinity dependency. The test now uses a raw loopback peer: it accepts both slice connections and reads both request headers before returning stale-registration on one, then verifies that the other in-flight slice is canceled and that the retry succeeds after metadata refresh. It also runs with a single worker, so the timing no longer depends on worker assignment or CPU count.

staryxchen (approved):
LGTM, thanks.

Issue #3856, opened 2026-09-03
[RFC]: Log-structured LOCAL_DISK backend with transparent tiered compaction


982945902:
### Summary

This RFC proposes a new, opt-in, log-structured `LOCAL_DISK` storage backend for Mooncake Store.

The backend is a separate implementation and an independent on-disk format. It does not reuse the existing FilePerKey, Bucket, or OffsetAllocator data layout. Foreground writes append self-describing records to holder-local Segment files; a persistent logical index maps each object version to its current physical record; background compaction rewrites live records and reclaims obsolete bytes without changing Master metadata.

```text
Master metadata:
  tenant/key/(object incarnation) -> LOCAL_DISK holder

Holder-local metadata:
  tenant/key/incarnation -> segment_id + record_offset + record_length
```

The Master owns logical placement and visibility. The Mooncake client process that mounts the Local SSD owns physical placement. Segment IDs and offsets never cross the Master protocol, so holder-local compaction is transparent to the Master.

This RFC deliberately separates two delivery stages:

- **V1 in this RFC:** independent log format, restart-safe local index, Put/Get, callback rollback, reclaim-only and size-tiered compaction, capacity bounds, throttling, recovery, metrics, and tests.
- **Durable per-object deletion:** enabled only after the object-incarnation plus delete-task/ACK protocol from #3221, or an equivalent protocol, is available. V1 must not advertise durable tombstone capability before that dependency exists.

## Motivation

Existing Local SSD backends optimize for different layouts:

- FilePerKey is simple but creates one file per object.
- Bucket groups batches into immutable files and tracks Bucket metadata.
- OffsetAllocator places values into reusable in-place extents inside a preallocated file.

A log-structured backend targets workloads with frequent replacement, uneven object sizes, small foreground batches, and a preference for sequential writes. It turns random free-space management into sequential append plus background reclamation.

The expected benefits are:

- sequential foreground writes;
- no allocation search on the Put path;
- cheap logical replacement of an existing key;
- holder-local physical reorganization;
- bounded, observable background compaction;
- crash recovery from checksummed records, WAL, checkpoints, and manifests.

The cost is temporary space and write amplification during compaction. The backend therefore remains opt-in and must enforce explicit capacity and throttling limits.

## Relationship to existing work

[...]

PR #3869, opened 2026-09-04
[Store] Add log-structured local SSD backend

982945902:
## Description

Implements #3856 by adding an opt-in log-structured `LOCAL_DISK` backend with an independent on-disk format.

The backend provides:

- append-only foreground writes into versioned Segment files, with short-lock offset reservation and payload I/O outside the Store lock;
- a holder-local version index with prepare/commit/abort semantics;
- WAL, checkpoint, manifest, and post-checkpoint recovery;
- crash-safe reclaim-only and size-tiered compaction;
- capacity reservation, compaction throttling, and backend metrics;
- concurrent batched payload submission through io_uring, with same-key striped serialization and an ordered publication barrier;
- concurrent reads with pinned Segment file descriptors, allowing in-flight reads to survive compaction unlink;
- direct-to-destination value reads without an intermediate decoded-record buffer;
- a shared read barrier that preserves the `complete_handler -> local commit` visibility protocol while allowing concurrent readers.

The Master continues to store only the logical `LOCAL_DISK` replica location. Segment IDs and offsets remain private to the holder, so compaction does not require a Master protocol change.

This is a separate backend and disk format. Existing FilePerKey, Bucket, OffsetAllocator, DFS, NoF, and NVMe-KV backends are unchanged.

### Integrity model

Complete records are checksummed and fully validated during recovery, Segment scanning, and compaction. The normal read path trusts the already validated committed physical index, validates its bounds, and reads the value directly into the caller buffer. It does not recompute the payload CRC on every cache hit. This matches the existing OffsetAllocator policy of performing full CRC validation during recovery rather than on the hot value-read path.

### Production-readiness hardening

The latest update (`94d70194`) addresses the concrete correctness and bounded-state issues identified during review:

- checkpoints retain prepared entries, while the post-checkpoint WAL uses its file generation as the replay boundary rather than incorrectly rejecting commit/abort records whose object sequence is already represented by the checkpoint;
- successful checkpoints prune obsolete, aborted, tombstoned, and reclaimable version history while preserving current committed and prepared versions;
- WAL append or sync failures fence the store into recovery-required state;
- callback exceptions abort prepared batches through an RAII guard;
- capacity rejection seals a non-empty active Segment so reclaim-only compaction can make progress;
- tiered compaction revalidates fan-in after input-budget truncation;
- recovery and compaction reuse one reader per source Segment, and compaction reads live records in physical-offset order;
- the existing FileStorage `use_uring` switch now controls log-structured payload writes.

### Current limitation

Durable delayed per-object deletion remains dependent on the object-incarnation and delete ACK protocol described by #3221. This backend does not advertise stronger deletion guarantees before that dependency is available.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [x] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
ctest --test-dir /tmp/Mooncake-rfc3856-build --output-on-failure \
  -R 'log_structured_.*test'

ctest --test-dir /tmp/Mooncake-rfc3856-build --output-on-failure \
  -R '^(file_storage_test|file_storage_config_test|ssd_metrics_test)$'

cargo test --offline --manifest-path mooncake-store/rust/Cargo.toml --lib
```

**Test results:**
- [x] All available `log_structured_*` test binaries pass (WAL 7/7, Store 46/46, Segment 14/14, Index 11/11, Metadata 9/9, StorageDirectory 5/5); the concurrent durable-append test also passes 100/100 repetitions.
- [x] Rust static-link unit tests pass: 8 passed, 0 failed.
- [x] `file_storage_test`, `file_storage_config_test`, and `ssd_metrics_test` pass.
- [x] Concurrent active-Segment reads, concurrent `BatchLoad`, compaction publication, reader-after-unlink, callback visibility, crash recovery, and capacity boundaries are covered.
- [ ] Integration tests pass (not run against a dedicated physical SSD host).

### Durable write benchmark

`storage_backend_bench`, buffered filesystem setup with equivalent durability policies:

- OffsetAllocator: `MOONCAKE_OFFSET_PERSIST_MODE=strict`
- Log Structured: `MOONCAKE_LOG_SYNC_POLICY=record`
- Three interleaved runs per point; medians shown; all operations completed with zero errors.
- Both backends were run with the same requested `use_uring` setting. The `io_uring` rows are not a pure layout-only comparison because OffsetAllocator's uring path opens its data file with `O_DIRECT`, while the log backend uses its batched uring writer on normally opened Segment files.

Results for `94d70194`, measured on September 7, 2026:

| Value / batch | Requested I/O | OffsetAllocator strict | Log Structured record-sync | Ratio |
| --- | --- | ---: | ---: | ---: |
| 4 KiB / 32 | pwrite | 244.39 MiB/s | 90.19 MiB/s (QD 4) | 0.37x |
| 4 KiB / 32 | io_uring | 145.41 MiB/s | 77.53 MiB/s (QD 4) | 0.53x |
| 128 KiB / 32 | pwrite | 345.98 MiB/s | 887.91 MiB/s (QD 16) | 2.57x |
| 128 KiB / 32 | io_uring | 341.74 MiB/s | 817.11 MiB/s (QD 16) | 2.39x |
| 4 MiB / 8 | pwrite | 358.59 MiB/s | 829.45 MiB/s (QD 32) | 2.31x |
| 4 MiB / 8 | io_uring | 360.20 MiB/s | 897.50 MiB/s (QD 32) | 2.49x |

This replaces the earlier 3–4x table, which unintentionally mixed I/O engines: the log backend used io_uring unconditionally while the OffsetAllocator benchmark used its default non-uring path. The corrected results show a clear medium/large-value benefit and a real 4 KiB regression caused by the log/WAL durability fixed cost. No small-value speedup is claimed.

The concurrent-append optimization is included in commit `9fe82b31`. Commit `3eb29922` embeds the implementation objects directly in `libmooncake_store.a`, preserving the Rust/Python/C static-link contract. Commit `94d70194` adds the recovery, bounded-state, capacity-progress, and configuration hardening described above.
Commit `ce1bcbc5` adds cross-request WAL group commit, releases the Store state mutex while waiting for WAL durability, and serializes transitions only per logical key so independent keys can share a durability boundary without reordering same-key state changes.

### WAL group-commit benchmark

The group-commit benchmark runs the real `PreparePut` + `CommitPut` path with `sync_data=true`, `sync_wal=true`, one operation per key, and no artificial batching delay. It compares the previous per-request WAL durability path at `8ff17433` with `ce1bcbc5`. Three interleaved runs were collected per point; medians are shown.

The shared 384-CPU host was heavily loaded during the run (`load average 278-289`), so these results are directional rather than stable hardware claims.

| Value | Writers | Baseline ops/s | Group commit ops/s | Speedup | WAL requests/fsync |
| --- | ---: | ---: | ---: | ---: | ---: |
| 4 KiB | 1 | 2,054 | 2,051 | 1.00x | 1.00 |
| 4 KiB | 8 | 3,114 | 8,138 | 2.61x | 3.26 |
| 4 KiB | 16 | 4,120 | 15,842 | 3.85x | 6.14 |
| 4 KiB | 32 | 3,517 | 21,026 | 5.98x | 13.01 |
| 128 KiB | 1 | 1,343 | 1,298 | 0.97x | 1.00 |
| 128 KiB | 8 | 3,341 | 3,961 | 1.19x | 2.52 |
| 128 KiB | 16 | 2,409 | 8,736 | 3.63x | 4.63 |
| 128 KiB | 32 | 2,116 | 11,954 | 5.65x | 9.36 |

Single-writer performance remains effectively neutral because no group forms. Under concurrency, fewer `fdatasync` calls account for the improvement; the benchmark does not attribute this gain to payload layout or io_uring.

### Read benchmark

The final implementation was also exercised for three interleaved rounds against OffsetAllocator for 4 KiB, 128 KiB, and 4 MiB values, using single-key and batched reads with one and eight reader threads. All valid-capacity runs completed with zero errors.

The shared host remained heavily loaded during this run (`load average 254-319` on 384 CPUs), so absolute and highly concurrent throughput numbers are not presented as stable performance claims. Representative three-run medians were:

| Workload | OffsetAllocator | Log Structured |
| --- | ---: | ---: |
| Single 4 KiB, 1 reader | 1410 MB/s | 1550 MB/s |
| Batch 128 KiB, 8 readers | 3908 MB/s | 3755 MB/s |
| Batch 4 MiB, 8 readers | 4407 MB/s | 4466 MB/s |

## Checklist

- [x] I have formatted my code using `./scripts/code_format.sh` (`--check --changed-lines --base origin/main` passes)
- [x] I have run pre-commit on the files changed in this PR and all hooks pass (`uvx pre-commit run --files ...`)
- [x] I have updated the documentation (RFC #3856 describes the design; the hot-read integrity clarification is included above)
- [x] For changes >500 LOC: I have filed an RFC issue (#3856)

## AI Assistance Disclosure

- [ ] No AI tools were used
- [x] AI tools were used (OpenAI Codex assisted with implementation, tests, review, and benchmark scripting; the submitter reviewed the resulting changes and test evidence.)

### DIFF
--- a/mooncake-store/benchmarks/storage_backend_bench.cpp
+++ b/mooncake-store/benchmarks/storage_backend_bench.cpp
@@ -875,6 +875,9 @@ void PrintBackendStorageStats(
     std::cout << "  Compaction bytes: " << stats->compaction_input_bytes
               << " input, " << stats->compaction_output_bytes << " output, "
               << stats->compaction_reclaimed_bytes << " reclaimed\n";
+    std::cout << "  Compaction C/C/E: " << stats->compaction_conflicts << "/"
+              << stats->compaction_cancellations << "/"
+              << stats->compaction_errors << "\n";
     std::cout << "  WAL/checkpoint:   " << stats->wal_sequence << "/"
               << stats->checkpoint_sequence << "\n";
 }
--- a/mooncake-store/benchmarks/storage_backend_bench.cpp
+++ b/mooncake-store/benchmarks/storage_backend_bench.cpp
@@ -15,7 +15,7 @@
 /**
  * @file storage_backend_bench.cpp
  * @brief Comprehensive benchmark for storage backends (OffsetAllocator, Bucket,
- * FilePerKey)
+ * FilePerKey, LogStructured)
  *
  * TESTS AVAILABLE:
  *   - init: Backend initialization time
@@ -92,6 +92,7 @@
 
 #include "gflags/gflags.h"
 #include "glog/logging.h"
+#include "storage/local/log_structured/log_structured_backend.h"
 #include "storage_backend.h"
 
 namespace fs = std::filesystem;
@@ -101,8 +102,10 @@ namespace fs = std::filesystem;
 // ============================================================================
 
 // === Core Parameters ===
-DEFINE_string(backend, "offset_allocator",
-              "Backend type: offset_allocator, bucket, file_per_key, or all");
+DEFINE_string(
+    backend, "offset_allocator",
+    "Backend type: offset_allocator, bucket, file_per_key, log_structured, or "
+    "all");
 DEFINE_uint64(value_size, 128 * 1024, "Value size in bytes (default: 128KB)");
 DEFINE_uint64(batch_size, 32, "Batch size for operations (default: 32)");
 DEFINE_uint64(num_operations, 1000,
@@ -242,7 +245,12 @@ AccessPattern StringToAccessPattern(const std::string& str) {
 // Backend Types
 // ============================================================================
 
-enum class BackendType { OFFSET_ALLOCATOR, BUCKET, FILE_PER_KEY };
+enum class BackendType {
+    OFFSET_ALLOCATOR,
+    BUCKET,
+    FILE_PER_KEY,
+    LOG_STRUCTURED
+};
 
 std::string BackendTypeToString(BackendType type) {
     switch (type) {
@@ -252,6 +260,8 @@ std::string BackendTypeToString(BackendType type) {
             return "bucket";
         case BackendType::FILE_PER_KEY:
             return "file_per_key";
+        case BackendType::LOG_STRUCTURED:
+            return "log_structured";
     }
     return "unknown";
 }
@@ -260,6 +270,7 @@ BackendType StringToBackendType(const std::string& str) {
     if (str == "offset_allocator") return BackendType::OFFSET_ALLOCATOR;
     if (str == "bucket") return BackendType::BUCKET;
     if (str == "file_per_key") return BackendType::FILE_PER_KEY;
+    if (str == "log_structured") return BackendType::LOG_STRUCTURED;
     LOG(FATAL) << "Unknown backend type: " << str;
     return BackendType::OFFSET_ALLOCATOR;
 }
@@ -830,6 +841,12 @@ std::shared_ptr<mooncake::StorageBackendInterface> CreateBackend(
             fpk_config.enable_eviction = false;
             return std::make_shared<mooncake::StorageBackendAdaptor>(
                 config, fpk_config);
+        }
+        case BackendType::LOG_STRUCTURED: {
+            config.storage_backend_type =
+                mooncake::StorageBackendType::kLogStructured;
+            return std::make_shared<mooncake::LogStructuredStorageBackend>(
+                config);
         }
     }
     return nullptr;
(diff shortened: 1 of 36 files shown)


982945902:
Marking this PR as Draft while we evaluate the RocksDB-based alternative in #3932.

The custom implementation remains useful as a reference design and A/B baseline, especially for understanding the cost of a purpose-built Segment/WAL/checkpoint/compaction stack. However, #3932 directly addresses the RFC question about reusing an established storage engine and currently offers a substantially smaller implementation surface.

For now, the proposed review direction is:

1. prioritize review of #3932 as the candidate production path;
2. keep this PR open as an experimental/reference implementation;
3. use the same workload matrix to compare durability, write throughput, hot-read performance, recovery, and long-term compaction behavior before deciding whether custom machinery is justified.

982945902:
Closing this experimental implementation after completing the Bucket, custom-log, and RocksDB comparison.

The corrected no-sync benchmark shows that the existing Bucket backend remains substantially faster for medium and large immutable batches. The custom backend adds WAL, checkpoint, recovery, and compaction complexity without demonstrating enough benefit for Mooncake's common KV-cache and RL lifecycle patterns. It also introduces write-amplification and read-tail risks that require production workload evidence before they are justified.

The implementation and benchmark remain useful as design evidence, but the better direction is to improve Bucket around explicit batch/epoch grouping, whole-Bucket retirement, and a well-defined durability contract rather than introduce a parallel storage format.

PR #3877, opened 2026-09-04
[Store] Fix get_batch returning unwritten buffer for duplicate keys

### LINKED ISSUE #3876: [Bug]: get_batch with a duplicate key returns the first occurrence's unwritten buffer as success (stale data)
### Bug Report

**Environment**
- Mooncake `main` @ `408b831`
- Component: `mooncake-store` central store client (`RealClient::batch_get_buffer_internal` → `Client::BatchGet`), reachable from the Python `get_batch` API
- Platform-independent (logic bug)

**Description**

When a batch read contains the **same key more than once**, the first occurrence is returned as a *successful* buffer that was never actually filled — it contains uninitialized/recycled client-allocator memory instead of the object's data. No error is raised and the checksum check does not catch it.

Root cause: the destination slice map is keyed by string, so duplicate keys collapse.

In `mooncake-store/src/real_client.cpp` (`batch_get_buffer_internal`, ~L3272-3300):
```cpp
std::unordered_map<std::string, std::vector<Slice>> batch_slices;
for (auto &op : valid_ops) {
    batch_keys.push_back(op.key);              // both duplicates kept
    batch_query_results.push_back(op.query_result);
    batch_slices[op.key] = op.slices;          // second op OVERWRITES the first
}
auto batch_get_results = client_->BatchGet(batch_keys, batch_query_results, batch_slices);
for (size_t i = 0; i < valid_ops.size(); ++i) {
    if (batch_get_results[i]) {
        final_results[op.original_index] =
            std::make_shared<BufferHandle>(std::move(*op.buffer_handle));  // first buffer unwritten
    }
}
```
Each op gets its **own** allocated buffer (`allocator->allocate` earlier in the loop), but only the last op's slices survive in `batch_slices`. In `Client::BatchGet` (`mooncake-store/src/client_service.cpp:1655`) the destination is resolved by `slices.find(key)`, so **both** duplicate indices transfer into the *same* (last) buffer, and the first op's buffer is never a transfer target. The checksum in `BatchGet` (`client_service.cpp:1778`) is computed against `slices.find(key)->second` — the filled buffer — for both indices, so it passes for the first index too. Both indices are reported `SUCCESS`, and `batch_get_buffer_internal` returns each op's own handle — including the unwritten first one.

The client buffer allocator does not zero memory on `allocate` (`client_buffer.cpp`, `offset_allocator.cpp`) and recycles freed regions, so the first handle surfaces bytes of previously fetched objects.

**Impact**

Silent data-integrity failure: a caller requesting object `k` at the first index gets `SUCCESS` with wrong/stale bytes, no error, no checksum catch. Duplicate keys occur naturally in real workloads (e.g. prefix-cache block hashes repeated within one batch), so this is reachable without adversarial input.

**Steps to reproduce**
1. Put an object under key `k`.
2. Call `get_batch(["k", "k"])` (or the C++ `batch_get_buffer` with a repeated key).
3. Index 1 holds `k`'s data; **index 0 is reported successful but contains uninitialized/recycled buffer memory**, not `k`'s data.

**Notes**

Duplicate-key handling is explicitly tested for the write and delete paths — `ClientIntegrationTest.BatchPutDuplicateKeys`, `BatchRemoveTest.EmptyAndDuplicate`, `RealClientTest.ErrDuplicatePutSameKey` — but there is no equivalent coverage or guard on the batch-get read path.

**Suggested fix**

Handle duplicate keys in `batch_get_buffer_internal` so each occurrence's own buffer is filled — e.g. send only the first occurrence of each key through the batch and satisfy subsequent occurrences via the existing single-key `get_buffer` path (or key the destination by index rather than by string).

elbourne12345:
## Description

Fixes #3876.

`RealClient::batch_get_buffer_internal` built `Client::BatchGet`'s destination
map keyed by object key (`batch_slices[op.key] = op.slices`). When the same key
appeared more than once in a single batch, those entries collapsed onto one
destination: in `Client::BatchGet` the destination is resolved via
`slices.find(key)`, so every occurrence transferred into the last op's buffer,
while the other occurrences' separately-allocated buffers were never written yet
still reported successful — surfacing uninitialized/recycled allocator memory as
valid data (the checksum is computed against the filled buffer, so it does not
catch it). Duplicate keys occur naturally in real workloads (e.g. repeated
prefix-cache block hashes within one batch).

This change processes the ops in rounds where each round holds at most one
occurrence of any key, so every occurrence's own buffer is filled. The common,
duplicate-free case runs in a single round and behaves exactly as before.

## Module

- [ ] Other

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

Added `DummyClientGetBufferTest.BatchGetBufferDuplicateKeysAllFilled`
(`mooncake-store/tests/dummy_client_get_buffer_test.cpp`): puts an object under
one key, requests that key three times in a single `batch_get_buffer` call, and
asserts every occurrence's own buffer holds the object's bytes (not stale/zero
memory) and that occurrences do not share a buffer. The key is deliberately not
warmed into the hot cache so all occurrences flow through the allocator-fallback
path (`batch_get_buffer_internal`) where duplicate keys used to collapse.

Built and run on a clean Ubuntu 22.04 machine (`sudo bash dependencies.sh -y`
then `cmake -B build -DUSE_HTTP=ON -DUSE_CXL=ON -DUSE_ETCD=OFF -DUSE_UB=OFF
-DUSE_CUDA=OFF -DWITH_P2P_STORE=OFF -DBUILD_UNIT_TESTS=ON` then
`cmake --build build --target dummy_client_get_buffer_test`):

- **With the fix — PASS:**
  ```
  [ RUN      ] DummyClientGetBufferTest.BatchGetBufferDuplicateKeysAllFilled
  [       OK ] DummyClientGetBufferTest.BatchGetBufferDuplicateKeysAllFilled (5034 ms)
  [  PASSED  ] 1 test.
  ```
- **Reverting only the fix hunk in `real_client.cpp` (pre-fix) — FAIL**, confirming the test catches the bug:
  ```
  Duplicate occurrence 0 ... was reported successful but its buffer holds stale data
    (first mismatch at byte 0: got 0, expected 7). ...
  Duplicate occurrence 1 ... was reported successful but its buffer holds stale data
    (first mismatch at byte 0: got 0, expected 7). ...
  [  FAILED  ] DummyClientGetBufferTest.BatchGetBufferDuplicateKeysAllFilled
  ```
  (Occurrences 0 and 1 return zeroed/unwritten memory; only the last, surviving
  buffer is filled — exactly the reported bug.)

**Formatting:** verified with `clang-format` 20.1.7 (LLVM 20, per
`CONTRIBUTING.md`); `scripts/code_format.sh --staged` reports no changes.

- [x] Unit tests pass
- [x] Manual testing done (fixed=PASS, pre-fix=FAIL on Ubuntu 22.04, output above)

## Checklist

- [x] I have formatted my code using `./scripts/code_format.sh` (clang-format LLVM 20)
- [ ] I have updated the documentation (if applicable) — *N/A*
- [x] For changes >500 LOC: I have filed an RFC issue — *N/A, small change*

## AI Assistance Disclosure

An AI coding assistant was used to analyze the affected code path, draft the fix,
write the regression test, and run the build/test verification described above.
The submitter has reviewed the change.

### DIFF
--- a/mooncake-store/src/real_client.cpp
+++ b/mooncake-store/src/real_client.cpp
@@ -3291,8 +3291,8 @@ RealClient::batch_get_buffer_internal(
         std::vector<std::string> batch_keys;
         std::vector<QueryResult> batch_query_results;
         std::unordered_map<std::string, std::vector<Slice>> batch_slices;
-        std::vector<size_t> round_ops;      // valid_ops indices in this round
-        std::vector<size_t> deferred_ops;   // duplicate keys for a later round
+        std::vector<size_t> round_ops;     // valid_ops indices in this round
+        std::vector<size_t> deferred_ops;  // duplicate keys for a later round
         batch_keys.reserve(pending_ops.size());
         batch_query_results.reserve(pending_ops.size());
         round_ops.reserve(pending_ops.size());
--- a/mooncake-store/src/real_client.cpp
+++ b/mooncake-store/src/real_client.cpp
@@ -3332,62 +3332,91 @@ RealClient::batch_get_buffer_internal(
 
     // 5. Execute batch get for LOCAL_DISK replicas via SSD RPC
     if (!disk_ops.empty()) {
-        // Group by transport endpoint
-        std::unordered_map<std::string,
-                           std::unordered_map<std::string, std::vector<Slice>>>
-            offload_objects;
-        // Build key -> disk_ops index for result lookup
-        std::unordered_map<std::string, size_t> disk_key_to_idx;
-
-        for (size_t idx = 0; idx < disk_ops.size(); ++idx) {
-            auto &op = disk_ops[idx];
-            // Find the LOCAL_DISK replica — replicas may be in any order.
-            const Replica::Descriptor *replica_ptr = nullptr;
-            for (const auto &r : op.query_result.replicas) {
-                if (r.is_local_disk_replica()) {
-                    replica_ptr = &r;
-                    break;
-                }
-            }
-            if (!replica_ptr) {
-                LOG(ERROR) << "No LOCAL_DISK replica found for key: " << op.key;
-                continue;
-            }
-            const auto &replica = *replica_ptr;
-            offload_objects[replica.get_local_disk_descriptor()
-                                .transport_endpoint]
-                .emplace(op.key, std::vector<Slice>{
-                                     {op.buffer_handle->ptr(), op.total_size}});
-            disk_key_to_idx[op.key] = idx;
+        // Same duplicate-key hazard as the memory path above: offload_objects
+        // (emplace keeps the first occurrence's buffer) and disk_key_to_idx
+        // (assignment keeps the last occurrence's index) are both keyed by
+        // object key, so a key repeated within one batch collapses — the SSD
+        // read fills the first occurrence's buffer while the result is resolved
+        // to the last occurrence, whose own buffer is returned unwritten.
+        // Process the ops in rounds where each round holds at most one
+        // occurrence of any key, so every occurrence's own buffer is filled.
+        // The common, duplicate-free case runs in a single round.
+        std::vector<size_t> pending_disk_ops;
+        pending_disk_ops.reserve(disk_ops.size());
+        for (size_t i = 0; i < disk_ops.size(); ++i) {
+            pending_disk_ops.push_back(i);
         }
 
-        for (auto &[endpoint, objects] : offload_objects) {
-            if (objects.empty()) continue;
-            auto read_result =
-                batch_get_into_offload_object_internal(endpoint, objects);
-            for (auto &[key, slices] : objects) {
-                auto idx_it = disk_key_to_idx.find(key);
-                if (idx_it == disk_key_to_idx.end()) continue;
-                auto &op = disk_ops[idx_it->second];
-                if (read_result) {
-                    auto checksum_result = client_->VerifyObjectChecksum(
-                        key, slices, op.total_size,
-                        op.query_result.object_checksum);
-                    if (!checksum_result) {
-                        LOG(ERROR)
-                            << "SSD checksum verification failed for key '"
-                            << key
-                            << "': " << toString(checksum_result.error());
-                        continue;
+        while (!pending_disk_ops.empty()) {
+            // Group by transport endpoint
+            std::unordered_map<
+                std::string,
+                std::unordered_map<std::string, std::vector<Slice>>>
+                offload_objects;
+            // key -> disk_ops index for this round's result lookup
+            std::unordered_map<std::string, size_t> disk_key_to_idx;
+            std::vector<size_t> deferred_disk_ops;
+
+            for (size_t idx : pending_disk_ops) {
+                auto &op = disk_ops[idx];
+                // A key already claimed this round waits for the next round so
+                // it gets its own destination buffer.
+                if (disk_key_to_idx.find(op.key) != disk_key_to_idx.end()) {
+                    deferred_disk_ops.push_back(idx);
+                    continue;
+                }
+                // Find the LOCAL_DISK replica — replicas may be in any order.
+                const Replica::Descriptor *replica_ptr = nullptr;
+                for (const auto &r : op.query_result.replicas) {
+                    if (r.is_local_disk_replica()) {
+                        replica_ptr = &r;
+                        break;
                     }
-                    final_results[op.original_index] =
-                        std::make_shared<BufferHandle>(
-                            std::move(*op.buffer_handle));
-                } else {
-                    LOG(ERROR) << "SSD read failed for key '" << key
-                               << "': " << toString(read_result.error());
+                }
+                if (!replica_ptr) {
+                    LOG(ERROR)
+                        << "No LOCAL_DISK replica found for key: " << op.key;
+                    continue;
+                }
+                const auto &replica = *replica_ptr;
+                offload_objects[replica.get_local_disk_descriptor()
+                                    .transport_endpoint]
+                    .emplace(op.key,
+                             std::vector<Slice>{
+                                 {op.buffer_handle->ptr(), op.total_size}});
+                disk_key_to_idx[op.key] = idx;
+            }
+
+            for (auto &[endpoint, objects] : offload_objects) {
+                if (objects.empty()) continue;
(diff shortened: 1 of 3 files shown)


he-yufeng:
Read the full diff and traced the surrounding code on current main. The rounds approach is sound where it applies: each round claims at least the first pending op, so it always terminates; key/query/slices stay pushed together, so `batch_get_results[j]` correctly maps to `round_ops[j]`; and the duplicate-free case is one round with index-identical behavior to before. The regression test is well aimed (allocator-fallback path, position-dependent payload, distinct-pointer check).

One uncovered path though: step 5, the LOCAL_DISK SSD-RPC branch just below, has the same collapse and this PR doesn't touch it (`real_client.cpp`, the `disk_ops` loop). Two maps disagree there:

```cpp
offload_objects[endpoint].emplace(op.key, {op.buffer_handle->ptr(), op.total_size});
// unordered_map::emplace keeps the FIRST occurrence's slices: the transfer
// targets the first op's buffer.
disk_key_to_idx[op.key] = idx;
// assignment keeps the LAST occurrence's index.
```

With the same key twice in one batch, both best-replica=LOCAL_DISK on the same endpoint: the read fills the first op's buffer, the result loop resolves the LAST op via `disk_key_to_idx`, the checksum verifies the filled (first op's) buffer and passes, and `final_results` gets the last op's never-written buffer handle. So the last occurrence reports success over stale allocator memory, while the first occurrence's filled buffer is orphaned and its slot comes back nullptr. Same silent-corruption class as #3876, reachable once an object is offloaded to SSD and only the local-disk replica remains.

Suggest handling disk_ops per occurrence the same way (rounds, or keying `offload_objects`/`disk_key_to_idx` by op index instead of by key), plus a regression case that forces the LOCAL_DISK replica so the disk path is pinned too.

elbourne12345:
Great catch, thank you — you're right, and my initial fix missed it. The `disk_ops` / LOCAL_DISK SSD-RPC branch had the same collapse: `offload_objects` (emplace → first occurrence's buffer) and `disk_key_to_idx` (assignment → last occurrence's index) disagree on a repeated key, so the SSD read fills the first occurrence's buffer, the result resolves to the last occurrence, the checksum verifies the already-filled buffer and passes, and the last occurrence's own never-written buffer is returned as success while the first occurrence's slot comes back `nullptr`.

Pushed `19b4a53`: the `disk_ops` branch now runs in rounds mirroring the memory-path fix — `offload_objects`/`disk_key_to_idx` are rebuilt per round with unique keys, so each occurrence of a repeated key gets its own destination buffer filled. Duplicate-free batches stay a single round, unchanged.

I'm verifying the build now and will add a regression case that forces a LOCAL_DISK-only replica (offload + evict the memory replica) so the SSD path is pinned too.

elbourne12345:
Follow-up: added the LOCAL_DISK regression test in `ca8cbbb` — `RealClientTest.BatchGetBufferDuplicateKeysLocalDiskAllFilled` (in `pybind_client_test.cpp`, whose `RealClientTest` fixture already has the offload/SSD scaffolding). It enables offload, waits for the LOCAL_DISK replica, clears the MEMORY replica so only the disk replica remains, then requests the same key 3× in one `batch_get_buffer` — so every occurrence routes through `disk_ops` — and asserts each occurrence's own buffer holds the correct bytes with distinct pointers.

Verified on a clean Ubuntu 22.04 build: it passes with the fix, and fails on the pre-fix `disk_ops` (occurrence 0 comes back `nullptr`/stale), so it genuinely guards this path. The memory-path test and the rest of the store CTest suite still pass. Thanks again for catching the disk path.

he-yufeng:
Verified 19b4a53a line by line against the invariants from my earlier note. The disk-path fix closes the mirrored collapse correctly:

- Round-local `disk_key_to_idx` means each key in a round maps to exactly one op, so `offload_objects[key]` slices and the result lookup now agree on the same occurrence's buffer. The deferral check sits before the replica lookup, which is the right place: duplicates share one query_result and therefore one endpoint, so per-key deferral covers the cross-endpoint case too.
- Termination holds (first pending op per key is always claimed), the duplicate-free case is a single round with unchanged behavior, and the `!replica_ptr` path still drops the op to a nullptr result exactly as before.
- The result loop is unchanged code operating on round-local maps, so the checksum-then-move ordering is preserved per occurrence.

I'm holding approval for a store maintainer since this is the data plane, but from my read both halves of the duplicate-key collapse are now covered, and the new LOCAL_DISK regression pins the path that was missing.

he-yufeng:
3e515c5f checks out too. The test forces the disk_ops route (waits for the LOCAL_DISK replica, then clears the MEMORY one) and asserts all three duplicate occurrences come back filled, byte-identical to the source, and in distinct buffers, which is exactly the collapse this PR fixes. CTest is green on this branch.

The remaining red lane is the known #3867 main failure, unrelated to this change. Once #3870 lands and this rebases onto it, CI should go green and this is ready to go in.

he-yufeng:
#3870 landed on main, so rebasing this onto it should turn the test-wheel-ubuntu lane green for the first time in days. Nothing else is holding this from my side.

Aionw:
Could we deduplicate the keys before querying metadata and fetching data, then copy each successfully fetched and verified object into the other occurrences’ independent buffers?

Repeated occurrences of the same key should return identical data within one batch. The round-based approach performs N sequential reads for a key repeated N times, and repeatedly scanning deferred entries adds O(N²) bookkeeping in the all-duplicate case.

A key-to-original-indices mapping would let us fetch each unique key once, preserve result order and independent buffer ownership, and propagate a fetch failure to all corresponding entries. This should work for both the regular and SSD offload paths, using the appropriate copy operation for the allocator’s memory type.

For example, for `get_batch([A, B, A])`:

1. Deduplicate the keys into `[A, B]`, keeping track of their original positions: `A → [0, 2]`, `B → [1]`.
2. Call `BatchGet` once with `[A, B]`, writing the results into `buffer[0]` and `buffer[1]`.
3. Once A’s read and checksum verification succeed, copy `buffer[0]` into an independently allocated `buffer[2]`.
4. Return the buffers in the original order: `[buffer[0], buffer[1], buffer[2]]`.

This fetches each unique key only once while preserving a separate buffer for every occurrence. If a fetch fails, mark all occurrences of that key as failed.

PR #3879, opened 2026-09-05
[Conductor] conductor: prefix index types and hash strategy

Chase-Rong:
## Description

Second slice of the Mooncake Conductor upstreaming series, on top of the skeleton and common layer (#3825). Adds the hashing foundation the prefix index is built on; still a pure addition — nothing outside `mooncake-conductor` changes.

What lands here:

- `prefixindex/types.h`: `ContextKey` (the four fields that identify a cache context), `ProjectedPrefix`, `StorageTier`, the engine/shared owner types, and the registration/mutation/clear payload shapes, plus `std::hash` specializations for the two lookup keys.
- `prefixindex/hash_strategy.{h,cpp}`: the `HashChain` / `HashStrategy` interfaces and their vLLM and SGLang implementations, profile resolution and validation (`ResolveHashProfile` / `ValidateHashProfile` / `CreateHashStrategy`), and `DigestToHex`. The chain is lazily evaluated: asking for block *i* hashes and caches only up to *i*, so a prefix walk whose cursors have all stalled never pays for the untouched tail.
- SHA-256 backend selection in CMake: probe for the low-level `SHA256_*` API, which hashes through a stack context instead of allocating an EVP context per call, and fall back to EVP when those declarations are hidden by an OpenSSL
  no-deprecated build. FIPS deployments can force EVP with `-DCONDUCTOR_FORCE_EVP_SHA256=ON`.
- Golden-vector fixtures for both hash profiles, the generator script, and the vLLM cross-check script, wired into `conductor_test` through  `CONDUCTOR_TEST_FIXTURE_DIR`.
   * `hash_golden_vectors.json` -- generated vectors for the `sha256_cbor` profile (`hash_block_tokens` semantics with cbor2 6.1.1 `canonical=True`): 2 seed roots and 11 cases totalling 14 block digests. 
   * `hash_golden_vectors_sha256.json` -- generated vectors for the `sha256` profile (SHA-256 over `pickle.dumps(value, protocol=5)`): 4 seed roots and 19 cases totalling 25 block digests. Cases cover UTF-8 multibyte lora and salt
names, the 255/256-byte string-length boundaries, signed 32-bit token boundaries, CPython's 64 KiB Pickle frame boundary and its large-payload path, incomplete tail blocks, and inputs with no complete block.
    * `generate_hash_golden_vectors.py` -- regenerates `hash_golden_vectors_sha256.json` with the real CPython pickler as the normative oracle. It writes in place, so copy it to a scratch directory before running it to keep the checked-in fixture intact if it raises partway through. Regeneration is reproducible apart from the `generator_metadata.python_version` line.
    * `verify_fixture_against_vllm.py` -- re-checks `hash_golden_vectors_sha256.json`, loading
`vllm/utils/hashing.py` directly so that the upstream `sha256` function is the oracle. It needs an external checkout, so it is run by hand, not in CI.

The fixture scripts pin encoding and line ending explicitly rather than inheriting the platform default: the vectors carry non-ASCII cache salts, which the default encoding is not guaranteed to represent, and a fixed LF keeps regeneration byte-identical so the documented regenerate-and-diff check reports only real changes.

Remaining slices, in order: tiered prefix indexer, ZMQ subscriber + wire protocol, KV event manager + HTTP API, then e2e tests + cache-aware proxy example + docs.

## Module

- [ ] CI/CD
- [ ] Docs
- [x] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
# Standalone component build
cmake -S mooncake-conductor -B build -G Ninja -DBUILD_UNIT_TESTS=ON
cmake --build build -j8

# EVP SHA-256 backend (FIPS path)
cmake -S mooncake-conductor -B build-evp -G Ninja -DBUILD_UNIT_TESTS=ON \
  -DCONDUCTOR_FORCE_EVP_SHA256=ON
cmake --build build-evp -j8 && ctest --test-dir build-evp --output-on-failure

# Top-level build, the path CI takes
cmake -S . -B build-top -G Ninja -DBUILD_UNIT_TESTS=ON -DWITH_CONDUCTOR=ON
cmake --build build-top -j8 --target conductor_test
```

**Test results:**
- [x] Unit tests pass

`conductor_test` passes — 38 tests across 11 suites — under both the standalone and top-level configurations, and both SHA-256 backends (low-level and `-DCONDUCTOR_FORCE_EVP_SHA256=ON`) produce identical digests. Golden vectors
regenerate byte-identically, and the vLLM cross-check script agrees with the checked-in vectors.

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

AI-assisted. The conductor sources are carried over from the existing conductor branch; AI was used to determine the dependency-ordered split, scope the CMake increment to this slice, replace the hand-rolled integer parsing with `mooncake-common`'s `TryParseInteger` and differentially test the substitution, and make the fixture scripts platform-independent. Every changed line has been reviewed by the submitter.

### DIFF
--- a/.typos.toml
+++ b/.typos.toml
@@ -18,4 +18,8 @@ extend-exclude = [
   # DeepEP-derived elastic kernel headers keep upstream identifiers such as
   # `ue8m0x4`; exclude the imported header block from spelling checks.
   "mooncake-ep/include/elastic/*",
+  # Conductor hash golden vectors are generated digests and pickle byte
+  # dumps; hex runs collide with dictionary words. The generator scripts
+  # stay checked.
+  "mooncake-conductor/tests/fixtures/*.json",
 ]
--- a/mooncake-conductor/CMakeLists.txt
+++ b/mooncake-conductor/CMakeLists.txt
@@ -13,8 +13,10 @@ if(NOT GLOBAL_CONFIG)
 endif()
 
 find_package(Threads REQUIRED)
+find_package(OpenSSL REQUIRED COMPONENTS Crypto)
 
-add_library(conductor_cpp_core STATIC src/common/utils.cpp)
+add_library(conductor_cpp_core STATIC src/common/utils.cpp
+                                      src/prefixindex/hash_strategy.cpp)
 
 # mooncake-common headers are consumed directly instead of linking
 # mooncake_common: the helpers conductor needs (ascii_string.h,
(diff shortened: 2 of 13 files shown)


UNIDY2002:
We should specify codeowners for `mooncake-conductor/` first.

ykwd (approved):
LGTM

PR #3880, opened 2026-09-05
[Store] Restore caller thread affinity after SPDK env init

### LINKED ISSUE #3865: [Bug]:  NoF initialization unexpectedly changes the caller thread's CPU affinity
## Description

When Mooncake is built with `USE_NOF` and used as an in-process library, SPDK initialization can change the host application's thread affinity and cause CPU contention.

`RealClient::setup_internal()` calls `SpdkWrapper::InitializeEnv()`, which executes the following code on the calling thread:

```cpp
struct spdk_env_opts opts;
spdk_env_opts_init(&opts);
opts.name = "mooncake";
int rc = spdk_env_init(&opts);
```

The initialization path does not explicitly configure SPDK's core mask or main core. With the default `core_mask="0x1"`, SPDK pins the calling thread to CPU0 and makes it the DPDK main lcore. Mooncake then returns to the caller with its CPU affinity changed.

[Relevant code](https://github.com/kvcache-ai/Mooncake/blob/e58ad95342158514e273ad287e3859271f079572/mooncake-store/src/spdk/spdk_wrapper.cpp#L111)

## Impact

In an integration such as SGLang with an in-process Mooncake NoF client, initialization can override the scheduler thread's existing CPU/NUMA affinity. Threads subsequently created by that thread may inherit the restricted affinity, potentially causing inference scheduling, prefetch, and NoF polling to contend for CPU0.

CPU0 may also already be busy with unrelated tasks. The current initialization path does not expose SPDK core selection, so users cannot choose a suitable CPU based on workload and NUMA topology.

Mooncake's NoF polling workers are separate worker threads with their own affinity handling. Configuring the SPDK core mask alone would not automatically configure these workers.

## Expected behavior

Initializing Mooncake's NoF backend should not unexpectedly change the host application's business-thread CPU affinity.

## Possible improvements

- Consider performing SPDK initialization on a dedicated thread managed by Mooncake, with coordinated initialization entry points, completion notification, error propagation, and lifecycle management.
- Expose configuration for SPDK's core mask/main core and coordinate it with the CPU affinity of NoF polling workers, allowing users to avoid busy CPUs and account for NUMA topology.

These are potential approaches; the implementation can be discussed separately.

### Before submitting...

he-yufeng:
## Description

Fixes #3865. `SpdkWrapper::InitializeEnv()` runs `spdk_env_init()` on the caller's thread, and under DPDK's default core mask that thread is pinned to CPU0 as the main lcore. Mooncake is linked in-process, so the host application (an inference scheduler, for example) silently keeps running with that restricted mask, and threads it spawns afterwards inherit it. This saves the caller's affinity before the init and restores it right after, on both the success and failure paths. Linux only; other platforms are untouched.

## Module

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

No CI lane builds `USE_NOF` and there is no SPDK/DPDK setup available to me here, so a full-file compile against real SPDK headers and a live `spdk_env_init` were not possible. Verified what is verifiable without the hardware:

- Extracted the patched `InitializeEnv()` body verbatim and compiled it with `g++ 11.4 -std=c++17 -Wall -Wextra -Werror` (Ubuntu 22.04 container) against an SPDK stub whose `spdk_env_init` pins the calling thread to CPU0, mirroring DPDK's default core mask behavior.
- Stub sanity probe: without the restore, the thread really ends pinned to CPU0.
- Success path: returns true, caller's original mask restored.
- Failure path (`spdk_env_init` returns nonzero): returns false, affinity still restored.
- `clang-format-14 --dry-run --Werror` clean on the touched file.

A behavior check on a real NoF host is welcome; the change region is self-contained around the init call.

**Test results:**
- [ ] Unit tests pass

See above; no in-repo harness exists for the NoF path.

## Checklist

- [x] I have formatted my code using `./scripts/code_format.sh` (clang-format 14 dry-run clean)
- [x] I have run pre-commit on the files changed in this PR and all hooks pass (format hook equivalent; see testing notes)

## AI Assistance Disclosure

Prepared with AI assistance (Kimi K3), including the container verification harness. The submitter reviewed every changed line and can defend the change end-to-end.

### DIFF
--- a/mooncake-store/src/spdk/spdk_wrapper.cpp
+++ b/mooncake-store/src/spdk/spdk_wrapper.cpp
@@ -6,6 +6,11 @@
 #include <cstring>
 #include <cstdlib>
 #include <thread>
+
+#if defined(__linux__)
+#include <pthread.h>
+#endif
+
 #include "spdk/spdk_wrapper.h"
 
 namespace mooncake {
@@ -122,7 +127,26 @@ bool SpdkWrapper::InitializeEnv() {
     spdk_env_opts_init(&opts);
     opts.name = "mooncake";
 
+    // spdk_env_init pins the calling thread to the first core in core_mask
+    // (CPU0 by default) as the DPDK main lcore. Mooncake runs in-process, so
+    // that silently overrides the host application's own affinity; hand the
+    // caller's mask back once DPDK is done with the thread.
+#if defined(__linux__)
+    cpu_set_t caller_mask;
+    const bool have_caller_mask =
+        pthread_getaffinity_np(pthread_self(), sizeof(caller_mask),
+                               &caller_mask) == 0;
+#endif
+
     int rc = spdk_env_init(&opts);
+
+#if defined(__linux__)
+    if (have_caller_mask) {
+        pthread_setaffinity_np(pthread_self(), sizeof(caller_mask),
+                               &caller_mask);
+    }
+#endif
+
     if (rc != 0) {
         fprintf(stderr, "SPDK init failed: %d\n", rc);
         return false;


he-yufeng:
Status note for reviewers: the red test-wheel-ubuntu lane here is the known #3867 main failure, not this change. Same step, same assertion: `test_concurrent_stress` in `test_ssd_offload_in_evict.py` fails with 1116 empty concurrent reads, and every main push since #3738 fails this lane the same way. #3870 is the in-flight fix for the lane itself.

fcczzz:
Cross-linking #3904, which implements the same affinity save/restore fix for the SGLang issue reported in #3905.

Could we coordinate the two PRs and keep a single affinity fix? This PR already keeps the scope focused and includes Linux guards; the restore-failure warning from #3904 would be a useful addition. The SGLang/NoF setup from #3905 could also help validate the fix beyond the stub tests.

he-yufeng:
Closing this: #3904 landed the same caller-thread affinity save/restore across SPDK env init and is merged on main as of yesterday. Verified the merged version covers the path this one took (the env-init wrapper now preserves the calling thread's CPU set). Thanks @fcczzz for cross-linking. If any edge of my variant turns out to matter (the restore-on-init-failure path), I'll bring it back as a follow-up against the merged code.

PR #3885, opened 2026-09-05
[Store] Fail PutEnd closed on OpLog admission errors

200lz:
## Description

On the HA whole-object OpLog path, `PutEnd` updated live replicas to `COMPLETE` and exposed the object before attempting to append the corresponding `PUT_END` record.

If OpLog reservation or `Commit` failed, the failure was only logged and `PutEnd` still returned success. This could leave live state ahead of the recoverable OpLog state.

This change builds the post-`PutEnd` replica descriptors first, reserves an OpLog slot, and commits that payload before modifying live metadata. Reservation and commit errors are now returned to the caller without applying the `COMPLETE` transition.

The resulting invariant is:

```text
no OpLog admission/Commit
    => no live COMPLETE transition
    => no successful PutEnd
```

The existing live completion path is otherwise unchanged.

Related: #3808. PR #3860 addresses separate OpLog-pressure and leadership keep-alive issues.

### Scope

This is intentionally narrower than the durable-mutation barrier tracked in #3808.

It does **not**:

* wait for durable-prefix advancement;
* implement W02 `AwaitDurable`;
* introduce RL01-RL04 lifecycle records;
* change the OpLog record format;
* add a synchronous etcd wait to `PutEnd`;
* change multi-tenant quota-settlement ordering.

A `PUT_END` accepted by the writer can therefore still be lost if the leader fails before backend durability. That remains separate roadmap work.

## Module

* [ ] Transfer Engine (`mooncake-transfer-engine`)
* [x] Mooncake Store (`mooncake-store`)
* [ ] Reshard (`mooncake-reshard`)
* [ ] Mooncake EP (`mooncake-ep`)
* [ ] Mooncake PG (`mooncake-pg`)
* [ ] Integration (`mooncake-integration`)
* [ ] P2P Store (`mooncake-p2p-store`)
* [ ] Python Wheel (`mooncake-wheel`)
* [ ] Common (`mooncake-common`)
* [ ] Mooncake RL (`mooncake-rl`)
* [ ] CI/CD
* [ ] Docs
* [ ] Other

## Type of Change

* [x] Bug fix
* [ ] New feature
* [ ] Refactor
* [ ] Breaking change
* [ ] Documentation update
* [ ] Performance improvement
* [ ] Other

## How Has This Been Tested?

Two deterministic HA regressions cover the failure windows:

* `MasterServiceHATest.PutEndFailsBeforeMutationWhenBatchReservationUnavailable`
* `MasterServiceHATest.PutEndCommitFailureLeavesReplicasProcessing`

Before the fix, both fault injections demonstrated the same invariant violation: `PutEnd` returned success and exposed a `COMPLETE`/readable object even though its OpLog operation had not been admitted successfully.

After the fix, both paths return the injected error, leave the replica `PROCESSING` and unreadable, and permit the original write to complete successfully after the fault is removed.

```bash
cmake --build build-putend --target master_service_ha_test --parallel 1

./build-putend/mooncake-store/tests/master_service_ha_test \
  --gtest_filter='MasterServiceHATest.*PutEnd*'

./build-putend/mooncake-store/tests/master_service_ha_test \
  --gtest_filter='MasterServiceHATest.PutEndFailsBeforeMutationWhenBatchReservationUnavailable:MasterServiceHATest.PutEndCommitFailureLeavesReplicasProcessing' \
  --gtest_repeat=20

./build-putend/mooncake-store/tests/master_service_ha_test

ctest --test-dir build-putend --output-on-failure \
  -R '^(master_service_test|master_service_tenant_quota_test|master_service_ha_test|ordered_oplog_writer_test|oplog_batch_storage_test)$'
```

Results:

* baseline HA suite: 93/93 passed;

* both new regressions failed against the unchanged implementation as expected;

* focused post-fix `PutEnd` tests: 4/4 passed;

* new regressions repeated 20 times each: 40/40 passed;

* complete post-fix HA suite: 95/95 passed;

* relevant CTest set: 283/283 tests passed across 5 targets, with no skips;

* pre-commit, changed-line formatting, and `git diff --check` passed.

* [x] Unit tests pass

* [ ] Integration tests pass (not run; the change is covered by deterministic HA fault-injection tests)

* [ ] Manual testing done

## Checklist

* [x] I have performed a self-review of my own code
* [x] I have formatted my code using `./scripts/code_format.sh`
* [x] I have run pre-commit on the files changed in this PR and all hooks pass
* [x] Documentation update is not applicable; this is an internal HA correctness fix with no new API or configuration
* [x] I have added tests to prove my changes are effective
* [x] RFC is not required; the production change is well below the >500 LOC threshold

## AI Assistance Disclosure

* [ ] No AI tools were used
* [x] AI tools were used

GPT-6 Astra / Codex assisted with investigation, implementation, regression tests, validation, and PR drafting. The submitter reviewed the final diff and is responsible for understanding and defending the change end-to-end.

### DIFF
--- a/mooncake-store/src/master_service.cpp
+++ b/mooncake-store/src/master_service.cpp
@@ -4748,6 +4748,33 @@ auto MasterService::PutEnd(const UUID& client_id, const ObjectMeta& object_meta,
         return tl::make_unexpected(ErrorCode::INVALID_WRITE);
     }
 
+    if (enable_oplog_ && ordered_oplog_writer_) {
+        // Admit the post-state before changing live metadata. Commit can
+        // still fail after Reserve; durability remains asynchronous.
+        std::vector<Replica::Descriptor> post;
+        post.reserve(metadata.GetAllReplicas().size());
+        for (const auto& replica : metadata.GetAllReplicas()) {
+            auto descriptor = replica.get_descriptor();
+            if (replica.is_processing() && is_target_replica(replica)) {
+                descriptor.status = ReplicaStatus::COMPLETE;
+            }
+            post.push_back(std::move(descriptor));
+        }
+
+        auto reservation = ReserveBatchOpLogSlot();
+        if (!reservation) {
+            return tl::make_unexpected(reservation.error());
+        }
+        auto persist_result = AppendReservedOpLogWithDurableFinalize(
+            std::move(reservation.value()), OpType::PUT_END,
+            object_id.tenant_id.value(), key,
+            SerializeMetadataForOpLogFromReplicaDescriptors(metadata, post),
+            nullptr);
+        if (!persist_result) {
+            return tl::make_unexpected(persist_result.error());
+        }
+    }
+
     const bool had_completed_replica =
         metadata.HasReplica(&Replica::fn_is_completed);
     bool completed_pending_replica = false;
@@ -4827,15 +4854,6 @@ auto MasterService::PutEnd(const UUID& client_id, const ObjectMeta& object_meta,
     metadata.GrantReadLease(std::chrono::milliseconds::zero());
     PublishKvStored(key, metadata, object_id.tenant_id);
 
-    if (enable_oplog_ && ordered_oplog_writer_) {
-        std::string payload = SerializeMetadataForOpLog(metadata);
-        auto result = AppendOpLogVisibleBeforeDurable(
-            OpType::PUT_END, object_id.tenant_id.value(), key, payload);
-        if (!result) {
-            LOG(WARNING) << "PutEnd: OpLog queue failed for key=" << key
-                         << ", err=" << static_cast<int>(result.error());
-        }
-    }
     return {};
 }
 
--- a/mooncake-store/tests/ha/master_service_ha_test.cpp
+++ b/mooncake-store/tests/ha/master_service_ha_test.cpp
@@ -3240,6 +3240,124 @@ TEST_F(MasterServiceHATest, PutEndWritesBatchRecordOpLog) {
     EXPECT_EQ(2u, prefix.last_seq);
 }
 
+TEST_F(MasterServiceHATest,
+       PutEndFailsBeforeMutationWhenBatchReservationUnavailable) {
+    const std::string cluster_id = "test_put_end_reservation_failure";
+    auto backend = std::make_shared<FakeBatchHaKvBackend>();
+    auto service_config = MasterServiceConfig::builder()
+                              .set_enable_ha(true)
+                              .set_enable_oplog(true)
+                              .set_cluster_id(cluster_id)
+                              .set_oplog_batch_max_entries(1)
+                              .build();
+    MasterService service(service_config);
+    ASSERT_EQ(ErrorCode::OK, service.SetBatchOpLogBackendForTesting(backend));
+
+    auto mounted = PrepareSimpleSegment(service, "put_end_reservation_segment");
+    OpLogBatchStorage storage(cluster_id, *backend);
+    OpLogBatchRecord batch;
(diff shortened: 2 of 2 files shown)


200lz:
CI note: the failing `test-wheel-ubuntu (ubuntu-22.04, 3.12)` check appears unrelated to this change.

The only failure is:
`test_ssd_offload_in_evict.py::TestDistributedObjectStore::test_concurrent_stress`
at line 431 (`concurrent read(s) failed`).

The exact PR base, `2b3cefbac4d571bcb001db3aa702772b41b8c647`, fails in the same step/test/assertion in upstream run `33876771587` / job `101047822486`. PR #3860 also shows the same failure signature.

In the failing wheel test, HA and OpLog are disabled (`enable_ha=0, enable_oplog=0`). The separate CTest job passed 167/167 entries, including `master_service_ha_test` and the two new `PutEnd` regression tests.

I have not changed or rerun CI based on this failure.

he-yufeng (approved):
Approved. Reordering PutEnd to commit the PUT_END record before mutating live metadata is the right call for the live-state-ahead-of-OpLog hole.

I verified the load-bearing invariant myself rather than trusting the description: the post-state descriptors are built with exactly the mutation loop's predicate (`is_processing() && is_target_replica`), and `mark_complete()` only flips `status_` with no timestamps or side fields, so the committed record matches the state that gets applied. The reservation-failure and commit-failure tests both prove no mutation happens on the fail path and that the write becomes retryable once capacity returns, which is the behavior that matters operationally.

The red test-wheel lane is the known #3867 main failure (360 empty concurrent reads in test_concurrent_stress, same step and assertion), not this change; builds and CTest are green.

Icedcoco:
Thank you for addressing the PutEnd/OpLog ordering window. However, I do not think this change should be merged in its current form.

The existing behavior is intentional. Mooncake is a cache, and its HA contract permits losing a small amount of recently acknowledged data during failover. PutEnd therefore updates the live metadata and returns without waiting for durable-prefix advancement. OpLog submission is best effort: if admission or commit fails, we record the failure but keep the write available on the primary. This preserves PutEnd latency and prevents temporary OpLog pressure from turning successful cache writes into client-visible failures.

This PR changes that contract:

OpLog admission failure
  -> PutEnd fails
  -> the replica remains PROCESSING

That creates a new availability and recovery problem. The public Client::Put() path does not retry PutEnd; retrying the whole Put() can encounter OBJECT_ALREADY_EXISTS, which is currently treated as success even though the object may still be unreadable. The added tests retry the internal MasterService::PutEnd() directly, so they do not cover this user-visible path.

Also, on the healthy path both implementations still perform Reserve + Commit without waiting for durable persistence. The main effect of this PR is therefore changing admission failure from best effort to fail-closed, rather than improving durability or materially reducing blocking.

Given the intended cache semantics, I recommend keeping the current fail-open behavior and improving observability for OpLog admission failures. A fail-closed design would require a separate end-to-end API and retry/cleanup design for PROCESSING writes. I recommend closing or withdrawing this PR rather than merging it as-is.

200lz:
Thanks for clarifying the intended current cache semantics and the public `Client::Put()` retry behavior.

I agree that the direct `MasterService::PutEnd()` retry in my tests does not cover the user-visible recovery path, and that making OpLog admission fail-closed in isolation changes the current availability contract without providing the end-to-end retry/cleanup semantics needed for a `PROCESSING` object.

I’ll withdraw this change rather than expand its scope into the lifecycle/API work. The admission-pressure evidence may still be useful for a narrower observability follow-up, but I’ll keep that separate.

Thanks for the detailed review.

200lz:
Closing as discussed above. Thanks again for the clarification.

PR #3888, opened 2026-09-06
[TE] feat(efa): move AWS Neuron (Trainium) HBM directly over EFA

whn09:
## Description

On a Trainium (AWS Neuron) instance the EFA transport could only move host DRAM, so a Neuron deployment had to stage every buffer through the CPU. This adds direct Neuron HBM transfers: `neuron:<N>` becomes a first-class memory location, and the same `USE_EFA` binary still runs unchanged on hosts with no Neuron device.

libfabric's EFA provider already implements `FI_HMEM_NEURON` — it is how the Neuron collectives stack moves HBM between instances — so most of the work was asking for it correctly:

* **`FI_HMEM=neuron`.** The variable takes exactly one interface, not a list, so it is set from a runtime probe rather than unconditionally.
* **`FI_MR_HMEM` in `mr_mode`**, otherwise the provider never looks at the iface we pass.
* **`fi_mr_regattr()`** with `iface = FI_HMEM_NEURON` and `device.neuron` set to the ordinal the buffer lives on.
* **libnrt mapped before the first `fi_getinfo()`.** libfabric reaches the runtime with `dlopen("libnrt.so.1")` while initialising its HMEM interfaces, but the Neuron SDK installs libnrt under `/opt/aws/neuron/lib` without putting that directory on the loader path — so on a default install that dlopen fails and the only symptom is `fi_mr_regattr()` returning `-FI_ENOSYS` much later. `neuronAvailable()` dlopens it with `RTLD_GLOBAL` (adopting an already-mapped copy first, and never calling `nrt_init()`), which satisfies libfabric's later request by soname. Setting `LD_LIBRARY_PATH` from inside the process cannot work — glibc parses it once, at startup.

Neuron HBM reaches user space as a shared mmap of `/dev/neuron<N>`, so both questions the transport needs answered — *is this pointer device memory* and *which device is it on* — fall out of `/proc/self/maps`. That keeps the new file free of Neuron SDK headers and of any link-time dependency on libnrt.

The location string has to be `neuron:<N>` rather than a `cpu:N` hint, because a `cpu` prefix is what opts a buffer into `preTouchMemory()`'s host stores. The probe therefore overrides even an explicit caller hint, and logs when it does.

**PCIe-switch NIC affinity is not a micro-optimisation here.** Neuron HBM registers fine on any NIC, but a NIC one switch away from the device it serves measured **2.5 GB/s against 19.2 GB/s** for a switch-local one. Topology discovery therefore groups each Neuron device with the EFA NICs behind its PCIe bridge, using `nec_get_device_pci_bdf()` — necessary because the driver presents neuron devices as *virtual* class devices with no link back to their PCI function, and the ordinals are not in BDF order (on trn2.48xlarge `neuron0` is `0000:cc:00.0` while `neuron1` is `0000:b5:00.0`). That call needs no device allocation and no `nrt_init()`, so it cannot claim a NeuronCore out from under a framework sharing the process.

`Topology::resolve()` now numbers HCAs in **sorted storage-type order**. A device index is not private to one process: a peer rebuilds the matrix with `parse()` and then indexes the device/rkey arrays we published with the numbers *its* `resolve()` produced. `matrix_` is an `unordered_map`, so the two ends walked it in different orders (discovery order vs. JSON member order) and any affinity preference was silently defeated on the remote side. This is not a correctness bug — rkeys and device names stay consistently indexed — but it costs the whole benefit above.

One small change on the CUDA side: in `registerMemoryRegionInternal()` the `ScopedCudaContext` is still declared for the whole registration block (it has to stay current across `fi_mr_regattr()`), but it is now only *bound* when `iface == FI_HMEM_CUDA`. `device_ordinal` is a CUDA device only when the iface says so, and binding a CUDA primary context to a Neuron ordinal would fail the registration outright on a hypothetical build that has both enabled.

Also: `transfer_engine_bench`'s EFA branch now honours `--nic_priority_matrix`, which it previously ignored (it always called `discover({})`). That is what makes the affinity-off control experiment below possible.

Behaviour on non-Neuron hosts is unchanged: `neuronAvailable()` returns false on the first `access("/dev/neuron0")`, and `MC_EFA_DISABLE_NEURON=1` turns the whole path off.

### Old libfabric headers, and what the published wheels compile against

`FI_HMEM_NEURON` and `fi_mr_attr::device.neuron` arrived together in **libfabric 1.15** (checked tag by tag: 1.14.0 no, 1.15.0 yes) and are absent from older headers — including Ubuntu 22.04's `libfabric-dev`, which is 1.11. The two lines that name them therefore sit behind a configure-time `check_cxx_source_compiles()` probe, `MOONCAKE_HAVE_FI_HMEM_NEURON`, following the existing `HAVE_IBV_ACTIVE_SPEED_EX` precedent. When the probe fails, `neuronAvailable()` returns false with a message saying why, rather than letting Neuron HBM be registered as host memory — which would be silently wrong, and slow, since a `cpu` location pre-touches every page.

**The probe alone would be a trap**: it makes CI green while shipping published wheels with Neuron off. So `_build-efa-wheel.yaml` now **builds libfabric from source** (every provider `--disable`d, headers and a link target are all that is wanted) instead of installing `libfabric-dev`, pinned at **1.18.2** — the oldest release the transport actually claims to support, since it already asks `fi_getinfo()` for `FI_VERSION(1, 18)`, which jammy's 1.11 could never have satisfied anyway. Measured cost on a trn2: curl 0.3s + configure 7.8s + `make -j8` 4.7s ≈ 15s, so no caching is needed. Nothing changes at runtime — `scripts/build_wheel.sh` has an unconditional `--exclude libfabric.so*`, so the wheel still resolves to the EFA installer's copy under `/opt/amazon/efa` on the host. One edit covers everything: `ci_efa.yml` and all three `release-efa*.yaml` call this reusable workflow.

Bumping the runner to ubuntu-24.04 for libfabric 1.17 was rejected: it would raise the wheel's glibc floor.

A Neuron end-to-end CI job is not possible — GitHub-hosted runners have no Trainium hardware, which is why `ci_ascend.yml` / `ci_rocm.yml` are build-only for the same reason. CI coverage here is compile-only, and the hardware validation below is manual.

### Same-host transfers, and why `FI_HMEM` is deliberately not in `caps`

`hints_->caps` gains `FI_HMEM` only on CUDA/HIP builds, where a comment records that `FI_MR_HMEM` in `mr_mode` alone is not enough: without `FI_HMEM` in caps the SHM sub-provider initialises its copy callbacks in plain-host mode and does a host `memcpy()` into a device VA during SAR ([ofiwg/libfabric#12328](https://github.com/ofiwg/libfabric/issues/12328)). The Neuron path sets `mr_mode |= FI_MR_HMEM` and does **not** add `FI_HMEM` to caps, which was originally just an untested asymmetry. It is now a deliberate one, because the same-host path has been measured and the CUDA analogy does not hold:

* Intra-node Neuron-to-Neuron **works and is correct**, but at **0.07 GB/s** on trn2 and **0.09 GB/s** on trn1 — 200x+ off the wire rate, with every byte intact.
* The reason is that libfabric's `shm` sub-provider has no protocol that can *read* a Neuron buffer: the `FI_HMEM_NEURON` op table in `src/hmem.c` implements `copy_to_hmem` only, `copy_from_hmem` is a `FI_WARN_ONCE(...); return -FI_ENOSYS;` stub, and every IPC entry is `ofi_hmem_no_*`, so `smr_ep.c`'s `use_ipc` is always false. Left alone, shm `memcpy`s the `/dev/neuron` mapping as host memory — across an uncached PCIe BAR, hence the 0.07.
* So adding `FI_HMEM` to caps would be a **regression, not a fix**: `efa_shm.c` keys `FI_MR_HMEM` off exactly that bit, and an HMEM-aware shm fails outright on the ENOSYS above instead of merely crawling. (CUDA gets a SIGSEGV where Neuron gets slowness for the same reason: a CUDA device VA is not CPU-mappable at all, a `/dev/neuron` BAR mapping is.)
* Fixed instead by keeping shm out of the way on Neuron hosts: `setenv("FI_EFA_ENABLE_SHM_TRANSFER", "0", 0)` before the first `fi_getinfo()`. Same-host traffic then loops back over the EFA device at **19.23 GB/s** (trn2) and **14.21 GB/s** (trn1) — in both cases the same rate one device gets *between* two hosts, i.e. no loss. Overwrite is `0`, so an explicit environment value still wins, which is how the slow row above was measured with the same binary.

## Module

- [ ] Mooncake Master (mooncake-master)
- [ ] Integration (vLLM/SGLang/etc.)
- [x] Documentation
- [x] CI/CD
- [ ] Other (please describe):

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Refactor (no functional change)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Other (please describe):

## How Has This Been Tested?

Two `trn2.48xlarge` instances (us-east-2), 16 Neuron devices and 16 EFA NICs each, Neuron runtime 2.34.10.0 / driver 2.30.2.0, libfabric 2.4.0 from the EFA installer. All numbers are `transfer_engine_bench`. Run-to-run spread is ±2%, so only larger differences are read as real.

**Tuned shape.** Swept over block size, threads and batch at 16 devices × 4 GB buffers; `--threads` is the dominant knob and peaks well short of the core count, while block size is flat from 512 KB up (at 16 threads: 256 KB 233.87, 512 KB 247.09, 1 MB 243.04, 2 MB 248.44 GB/s):

| `--threads` | `--batch_size` | Write throughput |
| --- | --- | --- |
| 8 | 128 | 123.83 GB/s |
| 16 | 32 | 259.93 GB/s |
| **32** | **128** | **266–271 GB/s** |
| 64 | 32 | 249.38 GB/s |
| 128 | 16 | 255.26 GB/s |

**PCIe-switch affinity A/B** — identical runs at that shape apart from `--nic_priority_matrix`:

| Configuration | Throughput |
| --- | --- |
| Affinity on (this PR) | **266.58 GB/s** |
| Affinity off (all 16 NICs per device) | 23.48 GB/s |
| Host DRAM control, same fabric (CPU-to-CPU) | 113.16 GB/s |

Correctly routed Neuron HBM is ~2.4x faster than host DRAM on the same fabric; getting the routing wrong puts it ~4.8x *below* it. (An earlier sweep at 1 GB buffers / 256 KB blocks / 16 threads read 231.16 vs. 12.31 GB/s — same conclusion, the ratio just depends on the shape.)

**Verified on the wire, not just self-reported.** Cross-checked against `/sys/class/infiniband/*/ports/1/hw_counters/tx_bytes`, sampled in windows triggered on counter movement rather than on a log line — a window overlapping the MR-registration phase dilutes the rate and reads low:

| Run | Summed NIC counters | Bench self-report |
| --- | --- | --- |
| All 16 devices | 279.54 / 279.57 GB/s | 278.97 GB/s |
| One device | 19.26 / 19.27 / 19.27 GB/s | 19.26 GB/s |

The single-device run also puts **every** byte on exactly two NICs — `rdmap201s0` and `rdmap202s0`, the pair behind that device's switch — with the other fourteen flat at zero, which is the affinity claim measured from outside the process.

**Per-device throughput is the ceiling, not the fabric.** One device sustains 19.3 GB/s and does not move between 2 and 16 threads; 16 devices reach ~270, roughly linear at ~17 GB/s each. EFAv3 on this instance is 3,200 Gbps = 400 GB/s aggregate and each device has two 25 GB/s NICs behind its switch, so neither the NICs nor HBM (46.4 TB/s) nor NeuronLink-v3 (1,024 GB/s/chip) is what binds. It is also not the logical-core count: spreading buffers over all 64 logical NeuronCores (`--neuron_core_stride=1 --neuron_device_count=64`) is worth only +3% (279 vs. 270 GB/s) for twice the HBM footprint, and the same thread count on the 16-buffer shape gives 255, so that +3% comes from the core spread rather than the threads.

**The bytes actually land in HBM.** Device memory cannot go through the existing `memcmp` validator, so the two sides fill with different bytes (`--neuron_fill_byte`) and the target reports on shutdown how much of its HBM the peer overwrote. Target filled `0x00`, initiator `0xAB`; the target reported `524288/1048576 bytes ... differ from the fill byte 0x0, first new value 0xab` on both buffers — exactly the expected coverage for the block size used.

**Affinity mapping verified** against ground truth: all 16 devices resolve into 8 PCIe switches x 2 EFA NICs, matching the two `lspci`-confirmed pairs.

**Regressions:** `topology_test` and `efa_transport_test` both pass (13/13 and 11/11 on this base; 8/8 and 10/10 on the branch the work was originally developed on), and the CPU-to-CPU path is unchanged at 113.16 GB/s (256 KB blocks, 32 threads) with the sorted `resolve()` in place.

The rebased commit was rebuilt and re-validated on the same two hosts. Both instances were shared with another workload at that point, so the re-run was limited to the free NeuronCores (`NEURON_RT_VISIBLE_CORES=56-63`, 2 devices, 256 KB blocks, 8 threads): **write 46.07 GB/s, read 48.71 GB/s**, consistent with the 2-device point of the scaling curve, with the correct ordinals discovered (`neuron:14`, `neuron:15`) and the `0xAB` read-back confirmed again (524288/1048576 bytes of the first MiB of each addressed buffer).

### Also validated on two `trn1.32xlarge` (Trainium1, a second PCIe layout)

Added because trn1 is laid out differently — 4 root buses × (4 Neuron + 2 EFA), no PCIe switch in between, 8 EFA × 100 Gbps — and it found two things trn2 could not:

* **`pciParentBus()` recognised only the switch-port form** of a sysfs parent (`0000:b9:02.1`). trn1's parent is the host bridge, `pci0000:10`, so grouping returned nothing for every device, all 8 NICs were dropped, and each Neuron device fell back to using all NICs: **0.19 GB/s against 82.71 GB/s** with the grouping. Worse than a wrong switch on trn2 because trn1's groups are two NUMA nodes apart and NIC enumeration lists the far-NUMA NICs first. Also the "no device could be matched" warning sat *after* the early return this took, so the hole was completely silent; it now warns there too.
* **The `--neuron_core_stride` auto-detect (`core_count / 2`) hardcoded trn2's LNC=2 core fusion.** trn1 reports `core_count=2` and fuses nothing, so the step came out as `1` and `--neuron_device_count=16` stacked two buffers on each of devices 0–7, never touching 8–15 — with ~10% throughput as the only symptom. Replaced by a probe: allocate a page on each logical core, ask which `/dev/neuron<N>` the pointer belongs to, place one buffer on the first core of each distinct device. 86.45 GB/s over 16 devices against 80.02 before, matching an explicit `--neuron_core_stride=2` (87.33) within 1%.

Sweep at 16 buffers × 8 GB: every knob is inside the ±2% spread above 16 threads, including `--block_size` all the way down to 256 KB — the ceiling is not the wire. Peak **88.15 GB/s = 88% of the 100 GB/s line rate**; the device ladder is 14.20 (1 device) / 40.24 (4) / 82.67 (8) / 86.5–89.6 (16), which tracks how many of the four root buses the buffers reach rather than the device count itself. The **CPU-to-CPU control is 97.2 GB/s** (97% of line rate, flat over five shapes), so unlike trn2 host DRAM is slightly *faster* than the HBM path here — the case for moving HBM directly on trn1 is avoiding the `nrt_tensor_read()` staging copy, its buffer and its CPU time, not raw throughput.

The benchmark also refuses to start rather than guess: a `--nic_priority_matrix` it cannot open or parse is fatal (it used to fall through to a fabricated matrix naming `--device_name`, i.e. a NIC that exists on no Trainium host), and so is `--neuron_device` with any protocol other than `efa` -- these buffers are registered under the wildcard location on purpose, so the location string cannot be the guard.

New bench flags: `--neuron_device` (logical NeuronCore, `-1` disables), `--neuron_device_count`, `--neuron_core_stride` (0, the default, probes for one core per physical device — see the trn1 section above), `--neuron_fill_byte`. Both sides must agree on `--neuron_device_count` and `--neuron_core_stride`, since the initiator indexes the *remote* buffer list with its own buffer count — the same constraint the GPU path already has.

**Documentation:** `docs/source/design/transfer-engine/efa_transport.md` gains an AWS Neuron section covering the build (nothing Neuron-specific is needed — a stock `-DUSE_EFA=ON -DUSE_CUDA=OFF` prints `Neuron (FI_HMEM_NEURON): 1`), the PCIe affinity result on both layouts and how to confirm it from the NIC counters, the same-host caveat, the benchmark invocation at the tuned shape, and per-instance tuning sections with the sweeps above. `trn1.32xlarge` is listed as a supported instance type.

## Checklist

- [x] I have formatted my own code using `./scripts/code_format.sh` before submitting
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/mooncake-transfer-engine/example/transfer_engine_bench.cpp
+++ b/mooncake-transfer-engine/example/transfer_engine_bench.cpp
@@ -12,18 +12,22 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <dlfcn.h>
 #include <gflags/gflags.h>
 #include <glog/logging.h>
 #include <signal.h>
 #include <sys/time.h>
 
+#include <algorithm>
 #include <cmath>
+#include <cstdint>
 #include <cstdlib>
 #include <fstream>
 #include <iomanip>
 #include <memory>
 #include <sstream>
 #include <unordered_map>
+#include <vector>
 
 #include "common.h"
 #include "common/base/status.h"
@@ -117,10 +121,243 @@ DEFINE_int32(gpu_id, 0,
              "GPU/NPU ID to use, -1 for all GPUs, not supported for NPUs");
 #endif
 
+#ifdef USE_EFA
+DEFINE_int32(neuron_device, -1,
+             "Allocate buffers from AWS Neuron (Trainium/Inferentia) HBM on "
+             "this logical NeuronCore instead of DRAM; -1 disables. Valid if "
+             "protocol=efa");
+DEFINE_int32(
+    neuron_device_count, 1,
+    "Number of Neuron devices to spread buffers over, starting at the "
+    "one holding --neuron_device. One buffer per device, since only two "
+    "EFA NICs sit on any one device's PCIe switch and a single device "
+    "therefore cannot use the whole fabric");
+DEFINE_int32(
+    neuron_fill_byte, 0xCC,
+    "Byte to prefill Neuron HBM with. Give the two sides different "
+    "values and --mode=target reports, on shutdown, how much of its "
+    "HBM the peer actually overwrote -- the only end-to-end check that "
+    "the bytes landed in device memory, since the host cannot memcmp() "
+    "it");
+DEFINE_int32(neuron_core_stride, 0,
+             "Logical NeuronCores per physical device, i.e. how far to step "
+             "between the buffers of --neuron_device_count. 0 auto-detects. "
+             "Check the \"Registering Neuron HBM ... as neuron:N\" lines: the "
+             "ordinals must all differ");
+#endif
+
 using namespace mooncake;
+
+#ifdef USE_EFA
+// Neuron HBM allocation for the benchmark.
+//
+// libnrt is reached through dlopen() rather than linked, so a stock build of
+// this benchmark still starts on hosts with no Neuron SDK -- the same reason
+// the transport itself takes no build flag for Neuron (see efa_neuron.h).  Note
+// that no location string is derived from FLAGS_neuron_device here: that is a
+// *logical* NeuronCore index, which NEURON_RT_VISIBLE_CORES renumbers and which
+// several cores share per physical device.  Naming is left to the transport,
+// which reads the real device ordinal back out of the pointer.
+namespace {
+struct nrt_tensor;
+
+struct NeuronRuntime {
+    int (*init)(int, const char*, const char*) = nullptr;
+    int (*tensor_allocate)(int, int, size_t, const char*,
+                           struct nrt_tensor**) = nullptr;
+    void* (*tensor_get_va)(struct nrt_tensor*) = nullptr;
+    void (*tensor_free)(struct nrt_tensor**) = nullptr;
+    int (*tensor_write)(struct nrt_tensor*, const void*, size_t,
+                        size_t) = nullptr;
+    int (*tensor_read)(struct nrt_tensor*, void*, size_t, size_t) = nullptr;
+};
+
+// Tensor handles are kept keyed by VA because allocateMemoryPool() hands its
+// caller a bare pointer and freeMemoryPool() only ever gets that pointer back.
+std::unordered_map<void*, struct nrt_tensor*> neuron_tensors;
+
+const NeuronRuntime* neuronRuntime() {
+    static NeuronRuntime rt;
+    static bool ok = [&]() -> bool {
+        // Prefer an already-loaded libnrt over a second independent copy of the
+        // device runtime in one process.
+        void* handle = dlopen("libnrt.so.1", RTLD_LAZY | RTLD_NOLOAD);
+        if (!handle) handle = dlopen("libnrt.so.1", RTLD_LAZY);
+        if (!handle)
+            handle = dlopen("/opt/aws/neuron/lib/libnrt.so.1", RTLD_LAZY);
+        if (!handle) {
+            LOG(ERROR) << "--neuron_device requires libnrt: " << dlerror();
+            return false;
+        }
+        rt.init = (decltype(rt.init))dlsym(handle, "nrt_init");
+        rt.tensor_allocate =
+            (decltype(rt.tensor_allocate))dlsym(handle, "nrt_tensor_allocate");
+        rt.tensor_get_va =
+            (decltype(rt.tensor_get_va))dlsym(handle, "nrt_tensor_get_va");
+        rt.tensor_free =
+            (decltype(rt.tensor_free))dlsym(handle, "nrt_tensor_free");
+        rt.tensor_write =
+            (decltype(rt.tensor_write))dlsym(handle, "nrt_tensor_write");
+        rt.tensor_read =
+            (decltype(rt.tensor_read))dlsym(handle, "nrt_tensor_read");
+        if (!rt.init || !rt.tensor_allocate || !rt.tensor_get_va ||
+            !rt.tensor_free) {
+            LOG(ERROR) << "libnrt is missing the tensor API";
+            return false;
+        }
+        // NRT_FRAMEWORK_TYPE_NO_FW.  Idempotent, so calling it here is safe
+        // even when a framework in the same process has already initialised the
+        // runtime.
+        int rc = rt.init(1, "", "");
+        if (rc != 0) {
+            LOG(ERROR) << "nrt_init failed with NRT_STATUS " << rc
+                       << "; check NEURON_RT_VISIBLE_CORES (the range must be "
+                          "contiguous) and that no other process owns the core";
+            return false;
+        }
+        return true;
+    }();
+    return ok ? &rt : nullptr;
+}
+
+// How far to step, in nrt_tensor_allocate() core indices, to land on the next
+// physical device.
+//
+// The step is not the physical core count sysfs reports: those indices are
+// *logical* cores, and trn2 fuses cores in pairs by default (LNC=2), so its 8
+// physical cores per device are 4 logical ones.  The fusion factor is a runtime
+// setting with no clean query -- nrt_get_visible_vnc_count() segfaults when
+// called with the signature its symbol suggests -- so it is inferred from the
+// two numbers that can be read reliably: how many logical cores the runtime
+// hands out in total, and how many devices exist.
+int neuronCoresPerDevice() {
+    static const int stride = []() -> int {
+        if (FLAGS_neuron_core_stride > 0) return FLAGS_neuron_core_stride;
+
+        int physical_cores = 0;
+        std::ifstream f(
+            "/sys/devices/virtual/neuron_device/neuron0/core_count");
+        if (!(f >> physical_cores) || physical_cores <= 0) {
+            LOG(WARNING)
+                << "Cannot read Neuron core_count from sysfs, stepping "
+                   "one core per buffer; pass --neuron_core_stride if "
+                   "the buffers land on the same device";
+            return 1;
+        }
+        // LNC=2 is the trn2 default and the only fusion factor shipped so far.
+        const int stride = physical_cores >= 2 ? physical_cores / 2 : 1;
+        LOG(INFO) << "Neuron: " << physical_cores
+                  << " physical cores per device, stepping " << stride
+                  << " logical core(s) per buffer";
+        return stride;
+    }();
+    return stride;
+}
+
+void* allocateNeuronMemory(size_t size, int buffer_id) {
+    const NeuronRuntime* rt = neuronRuntime();
+    if (!rt) return nullptr;
+
+    // One buffer per device, so step a whole device's worth of cores.
+    const int core = FLAGS_neuron_device + buffer_id * neuronCoresPerDevice();
+
+    struct nrt_tensor* tensor = nullptr;
+    // 0 == NRT_TENSOR_PLACEMENT_DEVICE, i.e. HBM rather than pinned host
+    // memory, which is the whole point of the exercise.
+    int rc = rt->tensor_allocate(0, core, size, "mooncake_bench", &tensor);
+    if (rc != 0) {
+        LOG(ERROR) << "nrt_tensor_allocate(" << size << " bytes on NeuronCore "
+                   << core << ") failed with NRT_STATUS " << rc;
+        return nullptr;
+    }
+    void* va = rt->tensor_get_va(tensor);
+    if (!va) {
+        rt->tensor_free(&tensor);
+        LOG(ERROR) << "nrt_tensor_get_va returned NULL";
+        return nullptr;
+    }
+    // The buffer is filled through the runtime, never with memset(): host
+    // stores into device HBM are exactly what the transport's Neuron path
+    // exists to avoid.  Skipped silently if libnrt predates nrt_tensor_write.
+    if (rt->tensor_write) {
+        std::vector<uint8_t> pattern(std::min<size_t>(size, 1ull << 20),
+                                     (uint8_t)FLAGS_neuron_fill_byte);
+        for (size_t off = 0; off < size; off += pattern.size()) {
+            size_t len = std::min(pattern.size(), size - off);
+            rc = rt->tensor_write(tensor, pattern.data(), off, len);
+            if (rc != 0) {
+                LOG(WARNING) << "nrt_tensor_write at offset " << off
+                             << " failed with NRT_STATUS " << rc
+                             << ", buffer left uninitialized";
+                break;
+            }
+        }
+    }
+    LOG(INFO) << "Allocated " << size << " bytes of Neuron HBM at " << va
+              << " on logical NeuronCore " << core;
+    neuron_tensors[va] = tensor;
+    return va;
+}
+
+// Reports how much of each Neuron buffer no longer holds the fill byte, i.e.
+// how much of it the peer's transfers actually landed in HBM.
+//
+// This is the only end-to-end data check available on the Neuron path: the
+// buffer is device memory, so the validator's memcmp() approach cannot look at
+// it, and a completion queue entry only says the NIC believed it wrote
+// somewhere.  Reading it back through the runtime is what distinguishes "the
+// transport reported success" from "the bytes are in HBM".
+void reportNeuronOverwrite(const std::vector<void*>& addr) {
+    const NeuronRuntime* rt = neuronRuntime();
+    if (!rt || !rt->tensor_read) {
+        LOG(WARNING) << "libnrt has no nrt_tensor_read, cannot check what the "
+                        "peer wrote into Neuron HBM";
+        return;
+    }
+
+    const size_t kWindow = 1ull << 20;
+    for (size_t i = 0; i < addr.size(); ++i) {
+        auto it = neuron_tensors.find(addr[i]);
+        if (it == neuron_tensors.end()) continue;
+
+        std::vector<uint8_t> readback(kWindow);
+        int rc = rt->tensor_read(it->second, readback.data(), 0, kWindow);
+        if (rc != 0) {
+            LOG(WARNING) << "nrt_tensor_read on buffer " << i
+                         << " failed with NRT_STATUS " << rc;
+            continue;
+        }
+        size_t changed = 0;
+        int first_new_value = -1;
+        for (size_t off = 0; off < kWindow; ++off) {
+            if (readback[off] == (uint8_t)FLAGS_neuron_fill_byte) continue;
+            if (first_new_value < 0) first_new_value = readback[off];
+            ++changed;
+        }
+        LOG(INFO) << "Neuron buffer " << i << " at " << addr[i] << ": "
+                  << changed << "/" << kWindow
+                  << " bytes of the first MiB differ from the fill byte 0x"
+                  << std::hex << (FLAGS_neuron_fill_byte & 0xff)
+                  << ", first new value 0x"
+                  << (first_new_value < 0 ? 0 : first_new_value) << std::dec;
+    }
+}
+
+void freeNeuronMemory(void* addr) {
+    const NeuronRuntime* rt = neuronRuntime();
+    auto it = neuron_tensors.find(addr);
+    if (!rt || it == neuron_tensors.end()) return;
+    rt->tensor_free(&it->second);
+    neuron_tensors.erase(it);
+}
+}  // namespace
+#endif  // USE_EFA
 
 static void* allocateMemoryPool(size_t size, int buffer_id,
                                 bool from_vram = false) {
+#ifdef USE_EFA
+    if (FLAGS_neuron_device >= 0) return allocateNeuronMemory(size, buffer_id);
+#endif
 #if defined(USE_CUDA) || defined(USE_MUSA) || defined(USE_HIP) ||    \
     defined(USE_MACA) || defined(USE_HYGON) || defined(USE_COREX) || \
     defined(USE_UBSHMEM) || defined(USE_SUPA) || defined(USE_SUNRISE)
@@ -194,6 +431,12 @@ static void* allocateMemoryPool(size_t size, int buffer_id,
 }
 
 static void freeMemoryPool(void* addr, size_t size) {
+#ifdef USE_EFA
+    if (FLAGS_neuron_device >= 0) {
+        freeNeuronMemory(addr);
+        return;
+    }
+#endif
 #if defined(USE_CUDA) || defined(USE_MUSA) || defined(USE_HIP) ||    \
     defined(USE_MACA) || defined(USE_HYGON) || defined(USE_COREX) || \
     defined(USE_UBSHMEM) || defined(USE_SUPA) || defined(USE_SUNRISE)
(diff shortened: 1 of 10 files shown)


Aionw (approved):
LGTM for build and ci part.

mooncake-transfer-engine/src/transport/efa_transport/efa_neuron.cpp
```diff
@@ -0,0 +1,327 @@
...
+                   "loaded ("
+                << dlerror()
+                << "); libfabric will refuse to register Neuron HBM "
+                   "(fi_mr_regattr -> ENOSYS). Add the Neuron SDK library "
+                   "directory (usually /opt/aws/neuron/lib) to "
+                   "LD_LIBRARY_PATH.";
+        }
+        LOG(INFO) << "Neuron device detected, EFA transport will register "
+                     "Neuron HBM with FI_HMEM_NEURON";
+        return true;
```
alogfans, line 278: In this case, if `dlopenNeuronRuntime()` fails, why it still returns true? Does any undef behavior happen?

whn09: Good question, and it made me look properly. Short answer: `true` is deliberate, no undefined behaviour follows from it — but the line right below it *was* UB. Both are addressed in 9d13bd03.

**Why it still returns true.** This answer is cached for the lifetime of the process (the `static` in the lambda), and of the two possible mistakes `false` is the more dangerous one. `neuronAvailable() == false` makes `neuronProbeAddress()` report Neuron HBM as ordinary host memory, and a `cpu` location prefix is exactly what opts a buffer into `EfaTransport::preTouchMemory()`'s CPU-side stores — i.e. host writes into a device VA for every buffer of 4 GiB or more. Since `neuronAvailable()` first runs during transport install, a framework that loads libnrt itself and allocates its tensors afterwards would be caught by that. Returning `true` keeps both the hints and the per-buffer classification correct; its only cost, in the case where libfabric's own `dlopen` of libnrt also failed, is a loud `fi_mr_regattr -> -FI_ENOSYS` at registration time, which is what the warning immediately above predicts in words.

**No undefined behaviour downstream.** Nothing on this path dereferences anything obtained from libnrt — the `dlsym`'d pointers are only ever called after a successful load, and the device-ordinal probe reads `/proc/self/maps`, not the runtime. Registration either succeeds or fails with that errno. The reasoning was only in my head, so it is now in the comment.

**Your question did point at real UB, one line down**: `LOG(WARNING) << ... << dlerror()`. `dlerror()` may return `NULL` — notably right after an `RTLD_NOLOAD` probe, which is exactly what `dlopenNeuronRuntime()` tries first — and streaming a `NULL const char*` into an ostream is undefined. Guarded now in both places that did it (`efa_neuron.cpp` and `transfer_engine_bench.cpp`).

Same commit also fixes a smaller thing in that function: `"Neuron device detected, EFA transport will register Neuron HBM with FI_HMEM_NEURON"` was logged even when the load had just failed, i.e. directly after the warning saying registration will not work. It is now only on the success path.

whn09:
Pushed four commits on top of the reviewed `a29bb055` (`91f0d8a8`). One is @alogfans's review point; the other three come out of validating this on a **second instance type, `trn1.32xlarge`** (Trainium1), which has a different PCIe layout and found two real defects. Details are in the updated description; in short:

| Commit | What |
| --- | --- |
| `9d13bd03` | guard a `NULL dlerror()` (real UB) and stop logging "Neuron device detected" after a failed libnrt load — @alogfans's comment |
| `6450c414` | keep libfabric's `shm` sub-provider off Neuron HBM (`FI_EFA_ENABLE_SHM_TRANSFER=0`). This settles the open item that used to be in the description: same-host Neuron-to-Neuron *works* but at 0.07–0.09 GB/s, because shm's `FI_HMEM_NEURON` table has no `copy_from_hmem`. Adding `FI_HMEM` to `caps`, which the description previously floated, would have turned that into an outright ENOSYS failure. With shm off: 19.23 GB/s intra-node on trn2, 14.21 on trn1 — the same as one device gets between hosts. |
| `b855c7f3` | group EFA NICs by **root bus** as well as by switch port. trn1 hangs NICs and Neuron devices straight off the root complex (`pci0000:10`, one colon), so grouping silently returned nothing and every device used all NICs: **0.19 GB/s vs 82.71**. The relevant warning was also unreachable on that path. |
| `91f0d8a8` | benchmark places Neuron buffers by **probing** which `/dev/neuron<N>` backs each core, instead of stepping `core_count / 2` cores (that hardcoded trn2's LNC=2 fusion; on trn1 it silently used 8 of 16 devices, −10%). 86.45 GB/s over 16 devices vs 80.02 before. Plus the trn1 sweep, its CPU-to-CPU baseline and the doc updates. |

Nothing changes for trn2: the two transport fixes are additive (a warning, a second sysfs shape, one `setenv` with overwrite `0`), and the earlier trn2 numbers still stand. `trn1.32xlarge` is now listed as a supported instance type in the doc, with its own tuning section.

mooncake-transfer-engine/src/topology.cpp
```diff
@@ -797,8 +844,23 @@ int Topology::resolve() {
...
+    // rkey arrays we published with the numbers *its* resolve() produced.
+    // matrix_ is an unordered_map, so its iteration order depends on insertion
+    // order, and the peer inserts in a different one (JSON member order, not
+    // discovery order) -- which silently hands transfers to a NIC other than
+    // the preferred one.  Sorting by storage type makes the numbering a
+    // function of the matrix contents alone, so both ends agree.
+    std::vector<std::string> storage_types;
+    storage_types.reserve(matrix_.size());
+    for (const auto &entry : matrix_) storage_types.push_back(entry.first);
+    std::sort(storage_types.begin(), storage_types.end());
```
staryxchen, line 859: The sort is the right fix — peer `parse()` + `selectDevice()` indexes the published `rkey[]`/`devices[]`, so numbering has to be a function of the matrix contents, not `unordered_map` walk order. That contract is shared with RDMA/CXI, not just EFA.

Please add a test that `parse()` of the same matrix with keys in reverse order yields an identical `getHcaList()`. Today `topology_test` only checks set membership, so a revert to `for (auto &entry : matrix_)` would stay green and silently drop preferred-HCA again.

mooncake-transfer-engine/example/transfer_engine_bench.cpp
```diff
@@ -509,8 +861,15 @@ static Transport* installTransportFromFlags(TransferEngine* engine) {
...
-        // would auto-install RDMA transport. Manually discover instead.
-        engine->getLocalTopology()->discover({});
+        // would auto-install RDMA transport. Manually discover instead --
+        // unless an explicit matrix was given, which is then the only way to
+        // pin EFA transfers to a subset of the NICs (discovery always takes
+        // all of them).
+        if (FLAGS_nic_priority_matrix.empty()) {
+            engine->getLocalTopology()->discover({});
+        } else {
+            engine->getLocalTopology()->parse(loadNicPriorityMatrix());
```
staryxchen: If `--nic_priority_matrix` is set but the file cannot be opened, `loadNicPriorityMatrix()` silently fabricates a matrix around `FLAGS_device_name` (default `mlx5_2`). RDMA already did that; wiring it to EFA is new. A typo no longer falls back to `discover({})` — it tries a NIC that does not exist on Trainium.

Please fail if the file is missing or `parse()` != 0. Also worth a doc note that a custom matrix must include `neuron:<N>` keys or affinity is lost (376× on trn1).

staryxchen (commented):
Public-surface notes only (topology numbering + bench flags). EFA transport internals look good after the trn1 follow-ups.

mooncake-transfer-engine/example/transfer_engine_bench.cpp
```diff
@@ -117,10 +128,331 @@ DEFINE_int32(gpu_id, 0,
              "GPU/NPU ID to use, -1 for all GPUs, not supported for NPUs");
 #endif
 
+#ifdef USE_EFA
+DEFINE_int32(neuron_device, -1,
+             "Allocate buffers from AWS Neuron (Trainium/Inferentia) HBM on "
+             "this logical NeuronCore instead of DRAM; -1 disables. Valid if "
+             "protocol=efa");
```
staryxchen, line 135: Help text says “Valid if protocol=efa” but `allocateMemoryPool()` only checks `neuron_device >= 0`. With `--protocol=rdma`/`tcp` the bench still allocates Neuron HBM and registers it as `*`, so the host path can pre-touch device memory.

Please `LOG(FATAL)` when `neuron_device >= 0 && protocol != "efa"`.

mooncake-transfer-engine/src/topology.cpp
whn09: Added as `ToplogyTest.TestHcaNumberingIndependentOfInsertionOrder` in b854f73a, and you are right that the existing set-membership checks would not have noticed a revert.

One thing to flag about the exact form you asked for: **two parses of the same matrix with the keys in reverse order cannot fail on their own.** jsoncpp returns `getMemberNames()` sorted, so `parse()` walks the members — and therefore inserts into `matrix_` — in the same order either way. Both parses are in the test to state the invariant, but what actually catches a revert is pinning the numbering to the sorted keys: six locations (`cpu:0/1`, `cuda:0/1`, `neuron:0/1`), one preferred NIC each, and `selectDevice()` must return each location's sort position.

Verified the way you framed it — reverted `resolve()` to `for (auto &entry : matrix_)`, rebuilt, and the test fails:

```
  from_forward.selectDevice(entries[i].first)  Which is: 1
  static_cast<int>(i)                          Which is: 4
neuron:0
... and neuron:1 comes out 0 instead of 5
[  FAILED  ] ToplogyTest.TestHcaNumberingIndependentOfInsertionOrder
```

With the sort restored, `topology_test` is 14/14 — both in a `USE_EFA=ON` build and in a `USE_EFA=OFF USE_CUDA=OFF` one.

Agreed the contract is shared with RDMA/CXI, not just EFA, so the test is written in those terms and does not mention any transport. Worth noting for anyone reading later that `resolve()` runs only from `discover()` and `parse()` — `disableDevice()` deliberately does not re-run it, to keep indexes stable for exactly the reason above — so the sort costs one `std::sort` over at most a few dozen device names at startup, and nothing on the transfer path.

mooncake-transfer-engine/example/transfer_engine_bench.cpp
whn09: Correct, and fixed in 4c6c9c10.

`loadNicPriorityMatrix()` now fails when `--nic_priority_matrix` names a file it cannot open, instead of falling through to the fabricated matrix — that removes the same trap for `rdma`/`barex`, where a typo silently substituted `--device_name`. The EFA branch also checks `parse()`'s return value, which it discarded:

```
F ... transfer_engine_bench.cpp:823] Cannot open --nic_priority_matrix=/tmp/does_not_exist.json
F ... transfer_engine_bench.cpp:879] Cannot parse --nic_priority_matrix=/tmp/bad_matrix.json, rc -103
```

Doc note added in the PCIe affinity section: a custom matrix **replaces** the discovered mapping rather than adding to it, so it has to key on `neuron:<N>` or every Neuron buffer ends up with no preferred NIC — the second row of that table. It also restricts the transport to the NICs it names, which is worth stating because it is how that row was measured; a control run on a `trn1.32xlarge` with a matrix listing only `rdmap16s25`/`rdmap16s26` brings up exactly those two NICs and registers Neuron HBM normally.

mooncake-transfer-engine/example/transfer_engine_bench.cpp
whn09: Correct, and fixed in 4c6c9c10 — checked once in `main()` right after flag parsing, so it covers both `--mode` values and both backends:

```
F ... transfer_engine_bench.cpp:1233] --neuron_device requires --protocol=efa, got --protocol=tcp
```

Your reasoning about the pre-touch is exactly right, and it is worth spelling out why nothing further down would have caught this: those buffers are registered under the wildcard location **on purpose**, because a `cpu` prefix is what opts a buffer into `preTouchMemory()`'s host stores. So the location string cannot double as the guard here — the protocol has to be.

Also compile-checked a `USE_EFA=OFF` build, since the new check references `FLAGS_neuron_device`: builds clean, and `--help` lists no `neuron` flags there.

whn09:
Both of @staryxchen's review points are addressed, plus the missing test for the `resolve()` sort. Head is now `4c6c9c10`.

| Commit | What |
| --- | --- |
| `b854f73a` | `topology_test`: HCA numbering must follow the matrix contents, not `unordered_map` walk order. Verified by reverting `resolve()` — the test fails, with `neuron:0` numbered 1 instead of 4. |
| `4c6c9c10` | `transfer_engine_bench`: fail on a `--nic_priority_matrix` that cannot be opened or parsed, and on `--neuron_device` with a protocol other than `efa`. Plus the doc note that a custom matrix replaces the discovered mapping and so must key on `neuron:<N>`. |

Validated on one of the `trn1.32xlarge` instances:

* `USE_EFA=ON`: builds clean, `topology_test` 14/14, all three new guards fire with the messages quoted in the threads, and a control run with a valid matrix still comes up on the NICs it names and registers Neuron HBM as `neuron:0`.
* `USE_EFA=OFF USE_CUDA=OFF`: builds clean, `topology_test` 14/14, no `neuron` flags in `--help` — the new check sits behind `#ifdef USE_EFA` and references `FLAGS_neuron_device`, so this configuration is the one it could have broken.
* Both files are clean under clang-format-20, the version CI uses.

No throughput A/B this round: every changed line runs before the first transfer (one `is_open()` on a file that was already being opened, one `parse()` return check, one flag comparison in `main()`), and `resolve()` is reached only from `discover()` and `parse()` — `disableDevice()` deliberately does not re-run it. The instances are also running someone else's Neuron workload at the moment, so a full-instance benchmark would neither be fair to them nor meaningful.

Issue #3884, opened 2026-09-05
[RFC]: Self-heal dangling LOCAL_DISK replicas on the read path


he-yufeng:
### Changes proposed

**The gap.** When a COMPLETE LOCAL_DISK replica's backing file physically disappears (manual wipe, RemoveAll broadcast, disk loss), the key stays listed and every read of it fails the same quiet way: the offload RPC read in `real_client.cpp` returns nullptr and Python `get()` returns `b""` with no error. #3711 self-heals exactly this on the Put path (probe file existence, `EvictDiskReplica`, clean retry), but read-only keys never heal by design, so for them the failure is permanent and invisible. This is the still-open half of what #3465 reported.

**Proposal.** On a LOCAL_DISK read failure, probe whether the replica's backing file still exists; if it is gone, evict the dangling replica via `EvictDiskReplica` (idempotent, master-side, same call #3711 already uses), then fail over to another replica if one exists. If none remains, the get fails visibly and the key stops being advertised, instead of returning empty bytes forever.

**Boundaries.** Failure path only, so healthy reads pay nothing. No change to the success path, to replica registration, or to the Python API's success semantics. Multi-replica failover ordering stays as it is today.

**Verification plan.** The in-proc master harness used by #3877's LOCAL_DISK regression test: put, wait for offload, wipe the backing file, then assert the get fails visibly and the key is no longer listed; plus a multi-replica case asserting failover serves the data.

I can implement this if the direction looks right.

Prepared with AI assistance (Kimi K3); the mechanism was verified by reading current main (2b3cefba), including the #3711 Put-path healing boundary and the read failure chain into the Python binding.

he-yufeng:
Implemented in #3889. The read path now runs the same probe-plus-evict heal the Put path has: a failed SSD read proves the backing file gone, evicts the dangling replica, and retries once so a survivor (or a fresh re-put) can serve the read; otherwise the master stops advertising the key and the miss surfaces honestly. Regression test included; it fails without the fix and passes with it in the in-proc master harness.

PR #3889, opened 2026-09-06
[Store] Self-heal dangling LOCAL_DISK replicas on the read path

he-yufeng:
## Description

A COMPLETE LOCAL_DISK replica whose backing file physically disappears (manual wipe, RemoveAll broadcast, disk loss) stays registered forever, and every `batch_get` against it just logs "SSD read failed" and hands back an empty buffer. #3711 already self-heals exactly this on the Put path (probe file existence, `EvictDiskReplica`, clean retry); the read path had no equivalent, which is the still-open half of #3465 and what #3884 tracks.

On a failed SSD read in `batch_get_buffer_internal`, the same heal now runs: the offload-file probe proves the backing file gone, the dangling replica is evicted master-side via `EvictDiskReplica`, and the key is retried once so a surviving replica (or a fresh re-put) can serve it instead of returning empty. When nothing survives, the master stops advertising the key and the miss surfaces honestly. The retry only ever heals distinct dead replicas, so it terminates. Checksum failures keep their existing path (the file exists, just corrupt, so no heal fires). Healthy reads are untouched; this is failure-path only.

`healDanglingLocalDiskReplica` moves from private to public in `client_service.h` so the RealClient read path can share the exact Put-side logic instead of growing a second copy.

Refs #3884, #3465, #3711.

## Module

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

**Test commands:**

```bash
cmake .. -DCMAKE_BUILD_TYPE=Release && make -j4 pybind_client_test
./mooncake-store/tests/pybind_client_test --gtest_filter='*HealsDangling*:*LocalDisk*:*Offload*:*BatchGet*'
./mooncake-store/tests/pybind_client_test --gtest_filter='*HealsDangling*'
```

**Test results:**

- [x] Unit tests pass

Docker (ubuntu 22.04, gcc) with the in-proc master harness: the new regression test `BatchGetBufferHealsDanglingLocalDiskReplica` (put, wait for the LOCAL_DISK replica, clear MEMORY, wipe the SSD root, then batch_get) fails on the unpatched tree (the dead replica stays advertised) and passes with the fix (honest miss, replica evicted from metadata). Log sequence observed end to end: `SSD read failed ... FILE_OPEN_FAIL` -> `Evicted dangling LOCAL_DISK replica ... (backing file already gone)` -> `Object not found`. The neighboring local-disk / offload / batch-get tests pass (3/3). The full CTest suite was not run locally; it runs in CI.

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

Prepared with AI assistance (Kimi K3). The investigation, implementation, and tests were reviewed line by line, and the verification above was run in Docker as described.


fcczzz:
I missed this when reviewing #3711: the existence probe receives the raw object key, but offload storage uses tenant-scoped keys, including for the default tenant. A healthy object can therefore be reported as missing.

This PR exposes that existing bug on reads. I reproduced it locally with two disk-only keys, A and B, in separate bucket files on the same owner:

- Delete only A's file and verify that B still reads correctly.
- Call batch_get_buffer({A, B}). The SSD batch fails, and healing runs for both keys.
- B's replica is evicted from master metadata even though its file is byte-for-byte unchanged. Subsequent reads of B fail.

The new counterexample fails on this PR and passes when I remove the added read-healing block. The PR's existing regression test passes in the same build.

Could we fix the probe before extending its use? It needs both the correct tenant-scoped storage key and a physical-file existence check: BucketStorageBackend::IsExist currently checks only the in-memory index, so fixing the key alone would still miss externally deleted files. Please also add a mixed healthy/missing-key regression to ensure a failed batch preserves healthy replicas.

he-yufeng:
@fcczzz Fixed, and your counterexample is now a regression test. Three changes:

**1. The probe now asks with the tenant-scoped key.** The install site in `RealClient::setup_internal` wraps the raw object key with `TenantId(tenant_id).MakeScopedKey(...)` before calling `FileStorage::Exists`. One wrap covers every call site of `healDanglingLocalDiskReplica`, so the Put, BatchPutStart, and batch-get paths are all fixed by the same edit rather than per-caller.

**2. `BucketStorageBackend::IsExist` now verifies the physical file.** Index hit plus `fs::exists(GetBucketDataPath(bucket_id))`; a miss on either side answers false. The ordering invariant makes this safe for fresh writes: `BatchOffload` commits the bucket file before the index insert, so an index hit can never precede its file. The per-key backend already answered physically (`fs::exists` on the resolved path), so the two backends now agree.

**3. Your mixed-key scenario is the new regression test** (`BatchGetBufferWipedSiblingPreservesHealthyReplica` in pybind_client_test.cpp): A and B land in separate bucket files, A's files are wiped, and `batch_get_buffer({A, B})` must evict A's replica, surface a real miss for A, and return B's data with B's replica untouched.

That test exposed one more real bug in my original read-path code, which I'd rather admit than bury: the SSD batch read dies wholesale on A's missing bucket file, so B's read failed too, and since B's replica was (correctly) not evicted, B's result stayed null. Healing preserved the replica but the caller still got a false miss for a healthy key. So `healDanglingLocalDiskReplica` now returns a 4-state result (`kNotDangling / kUnknown / kPresent / kEvicted`) instead of a bool, and the batch-get path retries a key when the file is proven *present* as well as after an eviction. The Put path contract is unchanged (retries only on kEvicted).

Verified in the dev container on this head: both heal tests pass, and the new test fails on the pre-fix code (B's replica evicted, B's data lost), passes with the fix.

he-yufeng:
Merged current main to clear the conflict (91d8e781). The disk-read section moved to op_status accounting upstream, so the heal retry now records the batch failure first and writes final_results only on a healed read, leaving the duplicate fan-out and handle assembly untouched. Both heal tests re-pass on the merged tree.

he-yufeng:
Four days since the last note, so a short status check rather than a bump: this still merges cleanly on current main (GitHub reports MERGEABLE) and CI is green end to end after the op_status rebase.

The uncovered half it closes is still real on main: a COMPLETE LOCAL_DISK replica whose backing file is gone stays registered, and every batch_get just logs the SSD failure and returns an empty buffer. The fix mirrors the heal #3711 already runs on the Put path (probe, EvictDiskReplica, one clean retry), and the tests pin both the healed-hit and the honest-miss outcomes. Happy to adjust the shape if you would rather see the heal live somewhere other than batch_get_buffer_internal.

mooncake-store/src/real_client.cpp
```diff
@@ -3423,6 +3428,30 @@ RealClient::batch_get_buffer_internal(
...
+                    // provably exists is retried as-is: its failure came from
+                    // a sibling's wiped bucket file, not from its own data.
+                    // The retry allocates through the allocator and lands in
+                    // final_results directly; op_status stays an error so the
+                    // duplicate fan-out and handle assembly below ignore it.
+                    if (client_) {
+                        const auto heal =
+                            client_->healDanglingLocalDiskReplica(key);
+                        if (heal == Client::DiskReplicaHealResult::kEvicted ||
+                            heal == Client::DiskReplicaHealResult::kPresent) {
```
fcczzz, line 3446: Could we explicitly bound this retry? kPresent establishes that the file exists, but not that a standalone read will succeed. For example, a truncated bucket file returns FILE_READ_FAIL while the existence probe keeps returning kPresent. The recursive call then takes this same branch again, retaining the previous buffer and allocating another on each attempt until allocation fails or the stack is exhausted.

mooncake-store/src/real_client.cpp
```diff
@@ -3423,6 +3428,30 @@ RealClient::batch_get_buffer_internal(
...
+                            auto healed = batch_get_buffer_internal(
+                                {key}, client_buffer_allocator);
+                            if (!healed.empty() && healed[0]) {
```
fcczzz, line 3449: The successful retry also needs to reach duplicate-key results. In the new mixed-file scenario, request {A, B, B}, with A's bucket file deleted and B healthy on the same owner. The initial batch fails, and B's standalone retry succeeds, but only the first B's final_results entry is updated. Its op_status remains an error, so FanOutDuplicates skips the copy and the second B still returns nullptr.

Could we feed the retried buffer and its success status into the duplicate fan-out path? Updating the status alone would not be enough, since the original buffer is not the buffer populated by the retry. Extending the mixed-key regression to {A, B, B} would cover this.

fcczzz (commented):
Sorry for the delayed follow-up — this PR slipped my mind over the past few days.

Thanks for fixing the scoped-key probe and physical-file check. I've now reviewed the updated code and left two inline comments.

The direction makes sense for #3884, and both fixes should fit within the current scope.

mooncake-store/src/real_client.cpp
he-yufeng: Both real, thanks. Fixed in 7fa1ce66.

The retry now runs once and bounded: the recursive call passes `heal_dangling_disk_replica=false`, so a kPresent-but-unreadable replica (your truncated-bucket case) fails plainly instead of recursing per attempt. Pinned by `BatchGetBufferTruncatedBucketFileSurfacesMiss`: the file is truncated to 1 KiB so the existence probe keeps answering kPresent while every read fails, and the run log shows exactly two read attempts (initial batch + one retry) before the miss surfaces, with the replica left advertised. One honest caveat: in the test's small arena the old recursion would still have terminated via allocation failure, so this test pins the contract rather than reproducing the stack burn — with a production-sized arena the old code would have kept going much longer, as you said.

mooncake-store/src/real_client.cpp
he-yufeng: Fixed in 7fa1ce66, exactly your scenario. The retry now copies the healed bytes into the op's own buffer with `CopyMaybeDevice` (the buffer was carved for this object by the original batch, so capacity matches; a size equality check guards it) and flips `op_status` to success. The duplicate fan-out then copies primary to duplicate as usual, and the handle assembly moves the buffer like any other successful read — no more direct `final_results` write on the heal path.

Regression test `BatchGetBufferHealedKeyFeedsDuplicateFanOut` runs your {A, B, B} case with A's bucket wiped: red on the pre-fix code (the duplicate B comes back null), green now with both B entries byte-identical to the source. The other two heal tests still pass, and `BatchGetBufferTruncatedBucketFileSurfacesMiss` pins the bounded retry from the comment above.

he-yufeng:
Heads-up on the moving SHA: the head is now 5328bb7bc0 — same two fixes as 7fa1ce66, amended for clang-format-20 conformance (the CI format gate flagged one string continuation and one stray blank line; no behavior change).

mooncake-store/include/real_client.h
```diff
@@ -787,7 +774,11 @@ class RealClient : public PyClient {
     std::vector<std::shared_ptr<BufferHandle>> batch_get_buffer_internal(
         const std::vector<std::string> &keys,
         const std::shared_ptr<ClientBufferAllocator> &client_buffer_allocator =
-            nullptr);
+            nullptr,
```
fcczzz, line 792: Could we cover the single-key path as well? Python `get()` calls `get_buffer()`, whose LOCAL_DISK branch still returns `nullptr` immediately after an SSD read failure. This new flag and healing logic only affect `batch_get_buffer_internal`, so the original `store.get(key)` empty-value symptom remains and the dangling replica stays advertised.

mooncake-store/src/real_client.cpp
```diff
@@ -3903,6 +3429,41 @@ RealClient::batch_get_buffer_internal(
...
+                    // The retry runs with healing off so a replica whose file
+                    // exists but keeps failing (e.g. a truncated bucket)
+                    // surfaces its error instead of recursing until the stack
+                    // or the allocator gives out. On success the bytes land in
+                    // this op's own buffer and op_status flips to success, so
+                    // the duplicate fan-out and handle assembly below treat
+                    // the healed key like any other successful read.
+                    if (client_ && heal_dangling_disk_replica) {
+                        const auto heal =
+                            client_->healDanglingLocalDiskReplica(key);
```
fcczzz: This reuses a Put-specific helper whose eligibility is too restrictive for reads. `healDanglingLocalDiskReplica` returns `kNotDangling` unless every COMPLETE replica is a LOCAL_DISK replica owned by the reading client. Consequently, a remote reader can never heal the failed owner’s replica, and a key with a healthy DFS/DISK fallback cannot fail over either—`SelectBestReplica` chooses LOCAL_DISK first, then this guard rejects healing because another COMPLETE replica exists. Could the read-side protocol ask the replica owner to verify/evict the specific failed replica and then re-query, with regressions for a remote reader and a LOCAL_DISK-plus-DFS/DISK key?

mooncake-store/include/real_client.h
he-yufeng: Covered in 1325e0a3. `get_buffer_internal` takes the same heal flag now: a failed LOCAL_DISK read heals before returning, then retries once with healing off so a present-but-unreadable file still surfaces a plain miss.

Regression test `GetBufferRemoteReaderHealsOwnersDanglingLocalDiskReplica`: a second client with no SSD of its own drives `get()` against the owner's wiped replica; the read misses and the replica gets evicted.

mooncake-store/src/real_client.cpp
he-yufeng: Reworked in 1325e0a3 along the lines you suggested. The read side no longer touches the Put-path helper. The reader asks the replica owner over the offload RPC (`verify_disk_replica`) to check the backing file, and the owner evicts its own replica when the file is proven gone.

The owner side turned out to be the only place the eviction can happen: `MasterService::EvictDiskReplica` scopes LOCAL_DISK eviction to the calling client_id, so a reader-side call reports success while removing nothing. The response is tri-state (present/evicted/undetermined) so a storage-layer error never evicts a healthy replica.

Remote readers are pinned by two regression tests, one through `get()` and one through `batch_get_buffer()`. After eviction the retry re-queries and re-selects, so a key with a surviving fallback is served by it; the fixture cannot spin up a real DFS/DISK replica cheaply, so the tests assert the miss-and-evict semantics directly.

mooncake-store/src/real_client.cpp
```diff
@@ -7974,6 +7251,75 @@ bool RealClient::release_offload_buffer(uint64_t batch_id) {
...
+    // batch_get_offload_object: the lambda captures only a raw pointer, every
+    // owning value lives in the heap state.
+    struct CallState {
+        std::vector<std::string> raw_keys;
+        std::vector<std::string> scoped_keys;
+        std::shared_ptr<FileStorage> file_storage;
+        std::shared_ptr<Client> client;
+    };
+    auto state = std::make_unique<CallState>();
+    const TenantId tenant_id(client_->tenant_id());
```
fcczzz: Could we carry the reader's tenant through this RPC and use it for both the storage probe and eviction? An SSD owner can hold objects from other tenants, so client_->tenant_id() here is not necessarily the failed object's tenant. The EvictDiskReplica overload below also defaults to the owner's tenant.

I reproduced this with a default-tenant SSD owner holding a tenant-a object. After deleting its backing file, both get() and batch_get_buffer() logged "Owner evicted", but tenant-a's dangling replica remained advertised: the master received an eviction for tenant=default, returned OBJECT_NOT_FOUND, and that was treated as success.

he-yufeng: Good catch, and thanks for the precise repro. The write path scopes offload files by the object's recorded tenant (the master fills OffloadTaskItem.tenant_id from the object identity), while the handler probed and evicted under the owner's own tenant. For a tenant-a object on a default-tenant owner the probe could never find the scoped file, and the eviction went out for tenant=default, so the master answered OBJECT_NOT_FOUND and we logged success over a replica that stayed advertised.

VerifyDiskReplicaRequest now carries the reader's tenant, and both the storage probe and the EvictDiskReplica call go through it. A reader can only discover the replica through its own tenant's master records, so the request tenant always matches the scope the write path used.

The regression covers the same mismatch with the roles flipped (tenant-a SSD owner, default-tenant object and caller, which keeps the memory replica clearable through the default-scoped BatchReplicaClear RPC): after the backing files are wiped, the caller's miss evicts the dead replica for real and the replica list comes back empty. On the old code the eviction goes out for tenant-a and the default record stays advertised. One note: the mismatch only exists with tenant isolation on, since the master resolves every tenant to default otherwise, so the test enables multi-tenancy through new InProcMasterConfig plumbing. All six earlier heal regressions still pass.

fcczzz: Thanks, the original cross-tenant repro now passes. However, I found a related regression when tenant isolation is disabled:

With enable_multi_tenants=false, the master maps every request to default, but the reader still sends its configured tenant. Probing tenant-a therefore reports a missing file even when the actual default-scoped file is healthy. The subsequent eviction is normalized to default by the master and removes that healthy replica.

Could we ensure the probe and eviction use the same master-resolved tenant as the stored object?

he-yufeng: Sharp catch, and exactly the hole in the assumption behind the third fix: "the reader's tenant matches the write scope" only holds when tenant isolation is on. With it off, the master resolves everything to "default", the write lands under the default scope, and the scoped probe then reported the healthy file missing, after which the master normalized the eviction to "default" and removed that healthy replica.

The probe and eviction now go through the same master-resolved scope without the handler needing to know the deployment mode: on a scoped miss it also probes the default scope and answers undetermined instead of evicting on the ambiguity. Only a miss in both scopes proceeds to eviction, which the master still resolves to the record the reader actually saw. That covers the four shapes: isolation-off healthy replica survives (your repro), a different tenant's object sharing the raw key is never evicted by mistake, a genuinely wiped replica still heals, and the default-tenant request path is unchanged.

Regression test added: isolation disabled, tenant-a reader against a healthy default-scoped replica; its verify must leave the replica advertised (b4fc706e). The full heal suite passes 10/10, including the earlier cross-tenant case.

PR #3891, opened 2026-09-06
[TransferEngine] Clear pending notify when freeing batch

### LINKED ISSUE #3890: [Bug][TransferEngine] Failed notify batches retain stale notification state
### Description

Classic Transfer Engine keeps notify metadata after a failed transfer batch is successfully freed.

`TransferEngineImpl::submitTransferWithNotify()` stores the target and `NotifyDesc` in `notifies_to_send_`, keyed by `BatchID`. `getBatchTransferStatus()` erases that entry only when the whole batch reaches `COMPLETED`. On `FAILED` or `TIMEOUT`, callers can later release the batch through `freeBatchID()`, but that method only delegates to `MultiTransport::freeBatchID()`.

`MultiTransport::freeBatchID()` deletes the underlying `BatchDesc`, while the `notifies_to_send_` entry remains alive with the same pointer-derived key and its notification strings.

### Runtime path

The Python synchronous transfer paths exercise this lifecycle:

```text
TransferEnginePy::transferSync / batchTransferSync
  -> allocateBatchID()
  -> submitTransferWithNotify()
       -> notifies_to_send_[batch_id] = notify
  -> getTransferStatus() / getBatchTransferStatus()
       -> FAILED
  -> freeBatchID()
       -> delete BatchDesc
       -> notification entry remains
```

The same side table is shared by the multi-protocol notify submission path.

### Impact

The immediate deterministic impact is lifetime retention: each failed notify batch can retain an `unordered_map` node, target ID, notification name, and arbitrary notification payload until the engine is destroyed.

There is also an ABA correctness risk because `BatchID` is the integer representation of the allocated `BatchDesc*`. If the allocator later reuses that address for a different non-notify batch, a successful status query for the new batch can find the stale entry and send the old notification for unrelated data.

### Expected behavior

When `freeBatchID()` successfully releases a batch, it should also remove any pending notification associated with that batch. A failed `freeBatchID()` (for example, `BatchBusy`) must keep the notification because the batch is still live and may later complete successfully.

### Reproduction

A deterministic unit test can use the existing `TransferEngineImplTestPeer` and fake transports in `mooncake-transfer-engine/tests/transport_uint_test.cpp`:

1. Submit a notify transfer through a fake transport that reaches terminal `FAILED`.
2. Confirm the batch status is `FAILED`.
3. Successfully call `freeBatchID(batch_id)`.
4. Inspect the pending-notify side table through the test peer.

Current behavior leaves one entry after step 3. The expected count is zero. A companion case should verify that `BatchBusy` leaves the entry intact.

### Relevant code

- `mooncake-transfer-engine/include/transfer_engine_impl.h`
  - `submitTransferWithNotify()`
  - `mp_submitTransferWithNotify()`
  - `freeBatchID()`
  - `getBatchTransferStatus()`
- `mooncake-transfer-engine/src/multi_transport.cpp`
  - `allocateBatchID()`
  - `freeBatchID()`
  - `getBatchTransferStatus()`
- `mooncake-integration/transfer_engine/transfer_engine_py.cpp`
  - `transferSync()`
  - `batchTransferSync()`

I plan to add hardware-independent regression coverage for both successful release and `BatchBusy`, then make `TransferEngineImpl::freeBatchID()` erase the side-table entry only after the underlying batch is actually freed. I expect to have a reviewed implementation ready within one day.

This issue was identified with AI assistance; the contributor independently traced the current `upstream/main` call path and collision-checked existing issues and pull requests.

CorgiBoyG:
## Description

Fixes #3890.

Classic Transfer Engine stores notify metadata in `notifies_to_send_`, keyed by the pointer-derived `BatchID`. The entry was removed only when a batch reached `COMPLETED`. A failed batch could therefore be successfully freed while its notification payload remained in the side table, retaining memory and allowing a later batch that reused the same address to observe stale notification state.

This change aligns the notification side table with the batch lifetime:

- `MultiTransport::freeBatchID()` keeps its existing public one-argument ABI and gains a private internal overload with a `before_delete` callback.
- The callback runs only after every task is confirmed finished, so `BatchBusy` preserves the live batch and its notification.
- On a successful release, pending notification state is erased before `BatchDesc` is deleted and its pointer-derived ID can be reused.
- The callback returns before `BatchDesc`, tasks, slices, and transport cleanup callbacks are destroyed, keeping potentially expensive destruction outside the global notification spin lock.
- A completed notification is atomically moved out of the side table under the lock, then sent through RDMA/RPC after the lock is released.

The regression coverage is hardware-independent and verifies:

- a terminal failed batch clears its pending notification after successful release;
- a busy batch keeps its notification until it can be successfully released;
- the internal pre-delete callback completes before batch and slice cleanup begins.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Checks run locally:**

```bash
git diff --check

pre-commit run --files \
  mooncake-transfer-engine/include/multi_transport.h \
  mooncake-transfer-engine/include/transfer_engine_impl.h \
  mooncake-transfer-engine/src/multi_transport.cpp \
  mooncake-transfer-engine/tests/transport_uint_test.cpp
```

The applicable trailing-whitespace, end-of-file, merge-conflict, large-file, clang-format 20, and codespell hooks passed.

The updated C++ tests were not executed locally. The available host is macOS and CMake configuration stops while resolving Linux project dependencies (`yaml-cpp`, followed by the RDMA/Linux stack). The tests are hardware-independent and are expected to run in the repository's Linux CI:

```bash
ctest --test-dir build-ci-local -R transport_uint_test --output-on-failure
```

The previous PR head passed its Linux CTest lane, but that result is not claimed for this updated head; the new CI run remains authoritative.

**Test results:**

- [ ] Unit tests pass (pending Linux CI for the updated head)
- [ ] Manual testing done

## Checklist

- [x] I have run the applicable pre-commit hooks on all changed files
- [x] I have updated the documentation (not applicable: internal lifecycle correction)
- [x] For changes >500 LOC: not applicable

## AI Assistance Disclosure

- [ ] No AI tools were used

AI tools assisted with code-path analysis, implementation, regression-test design, and static review. The human contributor reviewed every changed line of commit `107113c8`, understands the implementation, and can defend the change end-to-end. Validation limitations are stated above; no unexecuted test is represented as passing.

### DIFF
--- a/mooncake-transfer-engine/include/multi_transport.h
+++ b/mooncake-transfer-engine/include/multi_transport.h
@@ -15,14 +15,17 @@
 #ifndef MULTI_TRANSPORT_H_
 #define MULTI_TRANSPORT_H_
 
+#include <functional>
 #include <unordered_map>
 
 #include "transport/transport.h"
 
 namespace mooncake {
+class TransferEngineImpl;
 class TransferEngineImplTestPeer;
 
 class MultiTransport {
+    friend class TransferEngineImpl;
     friend class TransferEngineImplTestPeer;
 
    public:
@@ -85,6 +88,9 @@ class MultiTransport {
     void *getBaseAddr();
 
    private:
+    Status freeBatchID(BatchID batch_id,
+                       const std::function<void()> &before_delete);
+
     Status submitTransfer(BatchID batch_id,
                           const std::vector<TransferRequest> &entries,
                           std::vector<size_t> *task_sizes);
--- a/mooncake-transfer-engine/include/transfer_engine_impl.h
+++ b/mooncake-transfer-engine/include/transfer_engine_impl.h
@@ -263,14 +263,12 @@ class TransferEngineImpl {
     }
 
     Status freeBatchID(BatchID batch_id) {
-        // BatchID is pointer-derived and may be reused immediately after free.
-        // Keep release and notify cleanup atomic with notify table access.
-        RWSpinlock::WriteGuard guard(send_notifies_lock_);
-        auto status = multi_transports_->freeBatchID(batch_id);
-        if (status.ok()) {
+        return multi_transports_->freeBatchID(batch_id, [this, batch_id] {
+            // BatchID is pointer-derived. Remove side-table state before the
+            // descriptor is deleted and its address can be reused.
+            RWSpinlock::WriteGuard guard(send_notifies_lock_);
             notifies_to_send_.erase(batch_id);
-        }
-        return status;
+        });
     }
 
     int getNotifies(std::vector<TransferMetadata::NotifyDesc>& notifies);
@@ -366,16 +364,19 @@ class TransferEngineImpl {
         }
 #endif
         if (result.ok() && status.s == TransferStatusEnum::COMPLETED) {
-            // send notify
-            RWSpinlock::WriteGuard guard(send_notifies_lock_);
-            if (!notifies_to_send_.count(batch_id)) return result;
-            auto value = notifies_to_send_[batch_id];
-            auto rc = sendNotifyByID(value.first, value.second);
+            std::pair<SegmentID, TransferMetadata::NotifyDesc> value;
+            {
+                RWSpinlock::WriteGuard guard(send_notifies_lock_);
+                auto notify = notifies_to_send_.find(batch_id);
+                if (notify == notifies_to_send_.end()) return result;
+                value = std::move(notify->second);
+                notifies_to_send_.erase(notify);
+            }
+            auto rc = sendNotifyByID(value.first, std::move(value.second));
             if (rc) {
                 LOG(ERROR) << "Failed to send notify message, error code: "
                            << rc;
             }
-            notifies_to_send_.erase(batch_id);
         }
         return result;
     }
--- a/mooncake-transfer-engine/include/transfer_engine_impl.h
+++ b/mooncake-transfer-engine/include/transfer_engine_impl.h
@@ -263,7 +263,14 @@ class TransferEngineImpl {
     }
 
     Status freeBatchID(BatchID batch_id) {
-        return multi_transports_->freeBatchID(batch_id);
+        // BatchID is pointer-derived and may be reused immediately after free.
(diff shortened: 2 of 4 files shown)


he-yufeng (commented):
Read the full diff and checked the invariants this has to preserve:

- The callback runs after the all-tasks-done gate and before `delete &batch_desc`, which is the only safe point for state keyed by a pointer-derived BatchID. `BusyBatchKeepsPendingNotify` pins the busy door, and `BatchCleanupRunsAfterBeforeDeleteCallback` pins the ordering against slice cleanup callbacks. Both are the right probes.
- The completed path now does find -> move -> erase inside one short lock scope, then sends outside the lock. Erasing before send matches the old behavior on send failure (state is dropped either way), so no semantic change there, just a smaller critical section.
- The old single-arg `freeBatchID` still compiles to the empty callback, so existing callers are unaffected.

One thing worth a sentence in the commit message or a comment: a failed batch freed via `freeBatchID` now silently drops its notify state, which is exactly the fix, but it also means a reader of `notifies_to_send_` can no longer distinguish "notify sent" from "batch freed mid-flight". That seems fine for the current call sites; flag it if a future debugging path needs the distinction.

CTest green on CI. Looks right to me; leaving the merge to a second pair of eyes per our convention.

CorgiBoyG:
Thanks for the careful review. Agreed on the observability point: `notifies_to_send_` represents pending-delivery state, not an outcome record. Its absence already does not distinguish a successful send from a failed send, since the completed path erases the entry before `sendNotifyByID()`. This change adds intentional drop-on-release as another terminal transition. If sent/send-failed/dropped-on-free attribution becomes useful later, separate counters or events would be safer than retaining pointer-keyed state. I’ll keep this PR scoped to the lifecycle fix.

staryxchen (approved):
LGTM

PR #3893, opened 2026-09-06
[Reshard] Execute bound model-weight transfers through Transfer Engine

Bo-Vincent:
## Description

This PR adds Transfer Engine execution for bound heterogeneous model-weight
reshard plans.

The new resource-neutral `mooncake.reshard.transfer_engine` layer owns
physical transfer batches, registration leases, completion fencing, and
pending-resource quarantine. The model-weight adapter lowers validated N-D
regions into bounded flat batches and provides both source-initiated writes
and target-initiated reads through `MooncakeTransferEngineSink` and
`MooncakeTransferEngineReader`.

Execution revalidates the selected placement, runtime binding, generation,
lease, allocation range, and fragment evidence after framework allocation
guards pin the participating allocations. Unknown native completion retains
registrations and allocation tokens until an explicit terminal drain. The
current Python TE binding uses `batch_transfer_sync_read/write`; C++ scatter
lowering remains a later performance integration.

Related RFC: https://github.com/kvcache-ai/Mooncake/issues/3111

## Module

- [ ] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [x] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
PYTHONPATH=mooncake-wheel:mooncake-reshard/python:mooncake-reshard/tests \

python3 -m pre_commit run --files $(git diff --name-only origin/main...HEAD)

```

**Test results:**
- [x] Unit tests pass: `630 passed, 1 skipped`
- [x] Integration tests pass: exact-SHA H20 RDMA managed-memory TE E2E passed
  for TP2 -> TP1 with 2 MiB byte-for-byte validation.
- [x] Manual testing done: built the exact branch's native `engine` on H20-8
  and verified that the Python binding exposes the current flat batch API.
  A CUDA allocation RDMA-registration attempt reached the native TE boundary
  and was blocked by the host GDR environment (`Bad address [14]`); this is
  recorded as an environment validation boundary rather than a passing GPU
  G2G result.

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

AI assistance was used for implementation support, test harness preparation,
and review. The human submitter reviewed the design, code, and validation
results and remains responsible for this change.

### DIFF
--- a/mooncake-reshard/python/mooncake/reshard/transfer_engine/completion.py
+++ b/mooncake-reshard/python/mooncake/reshard/transfer_engine/completion.py
@@ -1,0 +1,465 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+from threading import Lock
+from typing import Any, Sequence, Union
+from uuid import uuid4
+
+from .lifetime import (
+    AllocationLifetimeToken,
+    TerminalTransferState,
+    release_tokens_after_terminal,
+)
+
+
+class TransferEngineError(RuntimeError):
+    pass
+
+
+class TransferCompletionUnknownError(TransferEngineError):
+    def __init__(self, message: str, *, pending_transfer_id: str) -> None:
+        super().__init__(message)
+        self.pending_transfer_id = pending_transfer_id
+
+
+class TransferCompletionFailedError(TransferEngineError):
+    """Native transfer reached a known terminal failure state."""
+
+
+class TransferRegistrationCleanupPendingError(TransferEngineError):
+    """Registration cleanup is quarantined until retry reaches a terminal state."""
+
+    def __init__(self, message: str, *, pending_transfer_id: str) -> None:
+        super().__init__(message)
+        self.pending_transfer_id = pending_transfer_id
+
+
+class _CompletionUnknown(RuntimeError):
+    def __init__(self, ticket: Any) -> None:
+        super().__init__("transfer completion remains unknown")
+        self.ticket = ticket
+
+
+class _CompletionWaitInterrupted(BaseException):
+    def __init__(self, ticket: Any, interruption: BaseException) -> None:
+        super().__init__(str(interruption))
+        self.ticket = ticket
+        self.interruption = interruption
+
+
+class _PendingCompletionWaitInterrupted(BaseException):
+    def __init__(
+        self,
+        pending_transfer_id: str,
+        interruption: BaseException,
+    ) -> None:
+        super().__init__(str(interruption))
+        self.pending_transfer_id = pending_transfer_id
+        self.interruption = interruption
+
+
+class _UndrainableCompletionUnknownTicket:
+    status = "COMPLETION_UNKNOWN"
+    restart_required = True
+
+    def drain(self, timeout_ms: int) -> str:
+        return self.status
+
+
+class _TerminalCleanupTicket:
+    restart_required = False
+
+    def __init__(self, status: str) -> None:
+        self.status = status
+
+    def drain(self, timeout_ms: int) -> str:
+        return self.status
+
+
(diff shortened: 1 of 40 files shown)


mooncake-reshard/python/mooncake/reshard/transfer_engine/completion.py
```diff
@@ -0,0 +1,398 @@
...
+    engine: Any,
+    *,
+    ticket_method_name: str,
+    legacy_method_name: str,
+    arguments: tuple[Any, ...],
+    max_drain_attempts: int,
+    drain_timeout_ms: int,
+) -> Union[int, str]:
+    ticket_method = getattr(engine, ticket_method_name, None)
+    if not callable(ticket_method):
```
zxpdemonio, line 152: **P1: The legacy fallback does not provide a completion fence.**

The current Python binding only exposes `batch_transfer_sync_read/write` (the PR description also says this), so production enters this branch because no `*_with_ticket` methods are implemented by the binding. The legacy C++ `batchTransferSync` returns `-1` on its timeout path even though its own code comments that `freeBatchID()` may run while tasks are still waiting (`mooncake-integration/transfer_engine/transfer_engine_py.cpp:675-684`). `_completion_status_name(-1)` maps that result to `FAILED_DRAINED`; the executor then takes the known-failure path and the registration context releases the caller allocation.

I reproduced this on the test machine with a legacy engine returning `-1` while marking the transfer in-flight: the code called `unregister_memory` while the transfer was still marked in-flight, returned no pending transfer ID, and released both allocation guards as `FAILED_DRAINED`. This defeats the completion-fencing guarantee and can release an allocation still referenced by DMA. Please either add/use a drainable native ticket API in the same change, or treat the legacy `-1` path as completion-unknown and quarantine/restart instead of releasing resources.

mooncake-reshard/python/mooncake/reshard/transfer_engine/registration.py
```diff
@@ -0,0 +1,398 @@
...
+    ):
+        raise _handoff_pending_resources(
+            pending_owner,
+            primary_error,
+            registrations=owned,
+            resources=resources,
+            lifetime_tokens=lifetime_tokens,
+        )
+
+    failures: list[tuple[int, Union[str, int]]] = []
```
zxpdemonio, line 349: **P1: Failed `unregister_memory` is neither retried nor quarantined.**

This path (and the mirrored target path below) collects unregister failures and immediately raises. The outer reader/sink `finally` still releases the allocation tokens, while `_pending_transfers` receives no entry containing the failed address. A transient unregister failure therefore leaves a live TE registration with no recovery path; later address reuse can collide with the stale registration.

I reproduced this on the test machine with `unregister_memory` returning `-7`: `pending_ids=()` and the allocation token was released as `ABORTED`. This is inconsistent with `mooncake-reshard/python/mooncake/reshard/weight/_store/registration.py:193-215`, which quarantines failed addresses and retains the lifetime token until a drain/retry succeeds. Please apply the same quarantine/retry mechanism here (including `registered_targets`) before releasing the allocation.

zxpdemonio (changes requested):
I reviewed and exercised this PR on the test machine at commit a9449d525d2b2af3e662832524cd971c211e4c5a. The focused TE suites pass (109 passed). The full reshard suite could not be collected because the test machine is missing the `hypothesis` package. I also ran focused probes for the legacy completion and registration-cleanup paths; the two issues below are reproducible and affect the lifetime/completion guarantees described by this PR.

Bo-Vincent:
I’ll close this PR and split it into two more focused PRs with clearer responsibilities.

Issue #3641, opened 2026-08-24
[RFC]: DFS replica: extract GlobalAllocator interface and add ImmutableBucketAllocator


yangzhxdxfl:
## 1. Background

The current DFS uses a **fixed-shard layout** managed by `OffsetAllocator`. While stable, this design has two limitations:

- **Inflexible capacity scaling**: The number of shards is fixed at initialization. Scaling total capacity requires rebuilding the entire DFS instance, which is disruptive in production environments.
- **Suboptimal I/O pattern**: `OffsetAllocator` performs **in-place updates** (cover writes) within shard files. This results in random-write I/O  compared to an append-only model.

By **abstracting the allocation logic into a `GlobalAllocatorInterface`**, we can introduce alternative allocation strategies without modifying core DFS replica logic. This lays the groundwork for a bucket-based, append-only allocator that addresses both scalability and I/O efficiency.

## 2. Design

We extract the allocation logic into a pluggable `GlobalAllocatorInterface` and introduce a new backend, `ImmutableBucketAllocator`. This enables DFS to support both the existing shard-based layout and a new append-only bucket model, addressing shard-mode limitations in capacity scaling and write I/O efficiency.

The design of `ImmutableBucketAllocator` is inspired by the bucket-based allocation model already used in Mooncake's *local SSD replica* implementation. That model has demonstrated good write amplification characteristics and simple lifecycle management in local storage. This PR extends the same concept to DFS, adapting it to a distributed setting with persistent metadata and bucket-level LRU eviction.

> **Note:** Mode selection is controlled via the `MOONCAKE_DFS_ALLOCATOR_TYPE` configuration option (`shard` or `bucket`).

### 2.1 Architecture Overview

```mermaid
flowchart TB
    subgraph Master["Master Service"]
        subgraph GAI["GlobalAllocatorInterface (abstract)"]
            direction TB
            GAI_methods["Allocate(key,size)→Descriptor<br/>Free(key,descriptor)<br/>UpdateAccess(key,descriptor)<br/>PrepareEviction()→PendingEviction<br/>CommitEviction(pending)<br/>AbortEviction(pending)<br/>IsInitialized()"]
        end

        subgraph DGA["DfsGlobalAllocator (Shard mode)"]
            DGA_files["dfs_shard_00.data<br/>dfs_shard_01.data<br/>..."]
        end

        subgraph IBA["ImmutableBucketAllocator (Bucket mode)"]
[...]

Issue #3819, opened 2026-09-01
[RFC]: [Store] Persist and Recover DFS Allocator Metadata


fcczzz:
### Changes proposed

Related: [PR #2683: DFS replica support](https://github.com/kvcache-ai/Mooncake/pull/2683)

## Summary

PR #2683 added DFS replicas backed by Master-allocated ranges in shared shard files. `DfsGlobalAllocator` currently keeps allocation state only in memory, so after a Master restart it may reuse an offset still referenced by restored metadata.

This RFC adds one checkpoint and one append-only WAL per shard:

```text
dfs_shard_<index>.data  # object bytes
dfs_shard_<index>.meta  # allocator checkpoint
dfs_shard_<index>.wal   # ALLOC/RELEASE records
```

The allocator is recovered first and remains sealed until it is reconciled with a standalone Master snapshot. This RFC covers single-Master restart recovery only. DFS remains incompatible with HA, OpLog recovery, and Standby promotion until a follow-up RFC.

This protects allocator metadata, not object bytes: a successful DFS `WriteAt` still does not imply data-file `fsync` durability.

## Motivation

Without allocator recovery:

```text
key A gets extent E
  -> snapshot records A's descriptor
  -> Master restarts with an empty allocator
  -> key B also gets E and overwrites A
```

A snapshot cannot rebuild the allocator exactly. A descriptor contains the exposed aligned offset and object size, but not the raw allocator node, normalized reservation size, pending frees, or allocations newer than the snapshot.

The two metadata sources therefore have different roles:

| State | Authority |
|---|---|
| Whether an extent may be reused | Allocator checkpoint + WAL |
| Whether an object or DFS replica is visible | Master snapshot |

## Goals

- Persist an allocation before returning its descriptor.
- Persist a release before making its extent reusable.
- Recover allocator state after Master restart.
- Reconcile recovered allocations with standalone snapshots.
- Fail closed on corrupt or incompatible sidecars.
- Preserve the Phase-1 descriptor and client data path.

## Non-goals

- HA, OpLog, Standby promotion, or automatic failover.
- Data-file `fsync` or durable object storage.
- Automatic migration of Phase-1 DFS roots.
- Multi-tenant DFS, DFS-only placement, or corruption repair.

### Files and compatibility fingerprint

Every shard requires `.data`, `.meta`, and `.wal`. If none exist, the allocator creates a fresh shard. If only a subset exists, startup fails. An existing data file must have exactly the configured capacity.

[...]

PR #3902, opened 2026-09-07
[Store] Add immutable bucket allocator for DFS replicas

wp-bc:
## Description

This PR implements the DFS immutable Bucket allocator proposed in:

- RFC: https://github.com/kvcache-ai/Mooncake/issues/3641

It introduces a pluggable `GlobalAllocatorInterface` for DFS replica allocation while preserving the existing `DfsGlobalAllocator` as the default `shard` implementation.

A new `ImmutableBucketAllocator` can be selected with:

```bash
MOONCAKE_DFS_ALLOCATOR_TYPE=bucket
```

The Bucket allocator stores objects as append-only immutable entries. Bucket files are created on demand, sealed metadata is persisted for standalone restart recovery, and physical space is reclaimed by evicting complete Buckets in LRU order.

The existing `DistributedFSDescriptor` wire format is unchanged. In Bucket mode, the existing `shard_idx` field carries the Bucket ID, preserving descriptor serialization compatibility.

#### Pluggable DFS allocator

- Extract `GlobalAllocatorInterface` as the Master-facing DFS allocation contract.
- Adapt the existing fixed-shard `DfsGlobalAllocator` to implement the interface.
- Keep `shard` as the default allocator with its existing behavior.
- Add `ImmutableBucketAllocator` following the terminology used by RFC #3641.

#### Immutable Bucket allocation

- Create Bucket files dynamically as capacity is needed.
- Store each key and value as one aligned, append-only entry.
- Never overwrite or reuse an entry in place.
- Allow a batch allocation to be packed across multiple Buckets.
- Roll back incomplete batch reservations without publishing invalid descriptors.
- Reclaim Bucket space when allocation is exhausted.

#### Metadata and recovery

- Persist one metadata snapshot for each sealed Bucket.
- Restore committed entries from sealed Bucket metadata after a standalone Master restart.
- Exclude pending and tombstoned entries from recovery.
- Remove the former append-only metadata log and rename-based publication flow.
- Conservatively handle corrupt metadata, orphan files, interrupted eviction, and incompatible configuration.

The active Bucket is intentionally not persisted. After an abrupt restart, it is discarded, while committed entries in sealed Buckets are recovered.

#### Bucket-level LRU eviction

- Track access at Bucket granularity.
- Exclude the active Bucket and Buckets containing processing entries from eviction.
- Validate leases and pin state before removing any replica.
- Remove all eligible replicas from Master metadata before deleting Bucket files.
- Abort the entire Bucket eviction if any entry is still protected.
- Trigger reclamation both at configured watermarks and on allocation exhaustion.

#### Batch I/O and direct reads

- Add packed Bucket `BatchWrite`.
- Group `BatchRead` requests by Bucket and process independent Buckets in parallel.
- Optionally merge adjacent entries into bounded reads.
- Add POSIX `O_DIRECT` reads with aligned pooled staging buffers.
- Fall back to buffered POSIX reads when the filesystem rejects `O_DIRECT`.
- Treat short reads, zero progress, EOF before the requested range, and partial failures as errors.
- Preserve the existing HF3FS adapter-specific read path.

#### Master and Client integration

- Integrate allocator selection into Master and Client initialization.
- Add Bucket batch reservation and asynchronous write completion.
- Copy or stage caller-owned buffers before asynchronous writes so they can be safely reused.
- Keep DFS replicas in `PROCESSING` until the write succeeds and the Master commits them.
- Revoke failed writes so incomplete data is never published as readable.
- Drain queued writes during Client shutdown.

#### Configuration and administration

The following settings are introduced or documented:

```text
MOONCAKE_DFS_ALLOCATOR_TYPE=shard|bucket
MOONCAKE_DFS_BUCKET_CAPACITY
MOONCAKE_DFS_MAX_BUCKET_COUNT
MOONCAKE_DFS_BATCH_READ_THREADS
MOONCAKE_DFS_BATCH_READ_MERGE_ENABLED
MOONCAKE_DFS_DIRECT_READ_ENABLED
```

The active Bucket limit can be changed online through:

```http
PUT /api/v1/dfs/max_bucket_count
Content-Type: application/json

{"max_bucket_count": 512}
```

The endpoint validates the JSON type and range. Values must be in `[1, INT32_MAX]`, and the endpoint returns a conflict response when the active allocator is not the Bucket allocator.

Lowering the limit below the current Bucket count does not delete existing Buckets immediately. It prevents additional Bucket creation until the count falls below the configured limit.

### Compatibility

- `shard` remains the default allocator.
- Existing shard allocation semantics are preserved.
- `DistributedFSDescriptor` serialization is unchanged.
- This PR builds on the DFS replica support already present in the community tree; it does not reintroduce the original DFS POSIX implementation.
- DFS prefetch, session object cache, monitoring metrics, persistent asynchronous H2D streams, Transfer Engine changes, and hardware-specific adaptations are intentionally out of scope.

### Known limitations

- Recovery covers sealed Buckets only; the active Bucket is discarded after an abrupt restart.
- Bucket recovery is for a standalone Master reopening the same DFS root.
- DFS remains incompatible with Master snapshot restore, oplog recovery, standby restore, and HA failover.
- Asynchronous Bucket writes do not currently provide a persistent retry queue, request cancellation, DFS-specific timeout, or `fsync` durability guarantee.
- A complete object must fit within one Bucket.

## Module

- [ ] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] Refactor
- [ ] Breaking change
- [x] Documentation update
- [x] Performance improvement
- [ ] Other

## How Has This Been Tested?

Validation was performed in a Linux build environment. The host contains vendor-modified RDMA headers, so an upstream-compatible verbs include path was supplied through:

```bash
CPLUS_INCLUDE_PATH=/tmp/mooncake-standard-rdma/include
```

**Build command:**

```bash
cmake --build build-no-shca --target \
  mooncake_store \
  dfs_immutable_bucket_allocator_test \
  dfs_immutable_bucket_backend_test \
  dfs_immutable_bucket_client_test \
  master_service_test \
  master_admin_server_test \
  dfs_sync_client_test
```

**Test results:**

- `dfs_immutable_bucket_allocator_test`: 51/51 passed
- `dfs_immutable_bucket_backend_test`: 50/50 passed
- `dfs_immutable_bucket_client_test`: 7/7 passed
- `dfs_posix_test`: 27/27 passed
- Master DFS regression filter: 9/9 passed
- Master admin DFS filter: 2/2 passed
- `dfs_sync_client_test`: 7 passed, 1 checksum test skipped because `MOONCAKE_STORE_CHECKSUM` was disabled
- `git diff --check 2b3cefbac4d571bcb001db3aa702772b41b8c647...HEAD`: passed

Coverage includes:

- Shard allocator regression behavior
- Bucket allocation and cross-Bucket packing
- Metadata persistence and restart recovery
- LRU eviction and allocation-exhaustion recovery
- Dynamic `max_bucket_count` updates and validation
- BatchWrite and BatchRead
- POSIX direct-read alignment and staging reuse
- Short reads and partial error propagation
- Master and Client integration
- Asynchronous write completion, revoke, and shutdown behavior

Not yet run in the current environment:

- Service-backed 3FS/HF3FS integration tests
- Documentation HTML build
- Pre-commit hooks
- LLVM 20 formatting check

The required 3FS/HF3FS service, Sphinx toolchain, pre-commit, and `clang-format-20` are not available on the validation host.

- [x] Unit tests pass
- [x] Integration tests pass (in-process POSIX integration)

## Checklist

RFC: https://github.com/kvcache-ai/Mooncake/issues/3641

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex was used to audit the internal implementation against the community baseline, identify already-upstream functionality, port and adapt the Bucket-specific changes, prepare tests and documentation, and assist with local validation.

The human submitter is responsible for reviewing, understanding, and defending every changed line before this PR is marked ready for review.


fcczzz:
I think this PR combines four substantial changes:

1. Introducing the bucket backend, including data layout, allocation, and reclamation.
2. Adding bucket metadata persistence and recovery.
3. Introducing asynchronous DFS writes with deferred PutEnd/PutRevoke.
4. Adding O_DIRECT reads and related read optimizations.

Would you be open to splitting these into separate PRs? We could start with a functional bucket backend using synchronous writes and regular reads, then follow up with separate PRs for recovery, asynchronous writes, and direct I/O optimizations.

For recovery, there is an existing RFC: https://github.com/kvcache-ai/Mooncake/issues/3819. Could we discuss the alternatives there and agree on which approach to adopt, particularly what the allocator should restore versus what the Master should restore?

Asynchronous writes change what a successful Put means and would benefit from a separate discussion. Ideally, this behavior should be independent of the allocator type. O_DIRECT and the other read optimizations could also be evaluated separately with correctness tests and benchmarks.

This would make each change easier to review and validate. Are there dependencies between these pieces that you think require them to land together?

wp-bc:
> I think this PR combines four substantial changes:
> 
> 1. Introducing the bucket backend, including data layout, allocation, and reclamation.
> 2. Adding bucket metadata persistence and recovery.
> 3. Introducing asynchronous DFS writes with deferred PutEnd/PutRevoke.
> 4. Adding O_DIRECT reads and related read optimizations.
> 
> Would you be open to splitting these into separate PRs? We could start with a functional bucket backend using synchronous writes and regular reads, then follow up with separate PRs for recovery, asynchronous writes, and direct I/O optimizations.
> 
> For recovery, there is an existing RFC: #3819. Could we discuss the alternatives there and agree on which approach to adopt, particularly what the allocator should restore versus what the Master should restore?
> 
> Asynchronous writes change what a successful Put means and would benefit from a separate discussion. Ideally, this behavior should be independent of the allocator type. O_DIRECT and the other read optimizations could also be evaluated separately with correctness tests and benchmarks.
> 
> This would make each change easier to review and validate. Are there dependencies between these pieces that you think require them to land together?

Thanks for the detailed feedback.   That makes sense, and we are happy to split this PR along the boundaries you suggested.
Our proposed sequence is:
1.   A minimal functional bucket backend: the allocator interface, bucket data layout, append-only allocation, whole-bucket reclamation and eviction, together with the minimum Master/Client integration.   This version will use synchronous DFS writes and regular buffered reads.
2.   Bucket metadata persistence and restart recovery, after aligning on the recovery model in #3819.
3.   Asynchronous DFS writes and deferred completion as a separate, allocator-independent change.
4.   O_DIRECT, merged reads, parallel batch-read optimizations, and their correctness tests and benchmarks as a separate performance PR.

Regarding #3819, is anyone currently implementing it?   If so, could you share the current progress or an implementation branch?   We would be glad to align the bucket recovery work with it and avoid introducing a competing recovery model.
On asynchronous writes, the current code keeps the DFS replica in PROCESSING state while the background write is running.   It calls PutEnd(DFS) only after BatchWrite succeeds, and calls PutRevoke(DFS) if the write fails, so an incomplete DFS replica is not exposed as readable.

However, you are right that the caller-visible semantics still change: BatchPut may return success after the synchronous replica has completed but before the deferred DFS write and its finalization have completed.   Therefore, a successful return no longer necessarily means that all requested replica writes have finished.   We will remove this behavior from the initial bucket PR and discuss it separately as an allocator-independent feature, including whether it should use an explicit async API/configuration and how background failures, retries, cancellation, and shutdown should be reported.

We will prepare the first reduced PR around the synchronous, buffered-I/O bucket backend and follow up with the other pieces separately.   Thanks again for helping us make the review scope clearer.

fcczzz:
@wp-bc Thanks, the proposed split sounds good.

Regarding #3819, yes—we have an implementation in progress in draft PR #4018. It covers shard allocator metadata persistence and reconciliation with the Master snapshot.

The bucket backend may need a different persistence mechanism from the shard allocator's checkpoint/WAL. Let's discuss the alternatives in #3819 and agree on the recovery semantics, particularly how to handle entries present on only one side.

wp-bc:
> @wp-bc Thanks, the proposed split sounds good.
> 
> Regarding #3819, yes—we have an implementation in progress in draft PR #4018. It covers shard allocator metadata persistence and reconciliation with the Master snapshot.
> 
> The bucket backend may need a different persistence mechanism from the shard allocator's checkpoint/WAL. Let's discuss the alternatives in #3819 and agree on the recovery semantics, particularly how to handle entries present on only one side.

Following the submission plan we discussed, I’ve opened PR #4124 covering the basic Bucket functionality. When you have a chance, could you please help review it?

fcczzz:
Thanks for opening PR #4124! I’ll take a look as soon as I can. Since the split is now tracked in #4124, could you close this PR and link the follow-ups?

PR #3903, opened 2026-09-07
[Bugfix][Store] Support DFS replicas in ranged get sessions

wp-bc:
## Description

Related to [#3658](https://github.com/kvcache-ai/Mooncake/issues/3658).

This PR extends the existing ranged get session APIs to support DFS-backed replicas, while preserving the existing MEMORY fast path.

The MEMORY replica path remains unchanged and is still preferred when available. `LOCAL_DISK`, `DISK`, and `NOF` replicas remain unsupported by ranged get sessions.

`batch_get_session_start` performs a single Master `BatchQuery`, selects one complete MEMORY or DFS replica, and caches the filtered `QueryResult` together with its lease. Subsequent `batch_get_into_multi_buffer_ranges` calls reuse the cached metadata and do not query the Master again.

For DFS replicas:

- Requests use a request-scoped staging arena backed by the existing `PinnedBufferPool`.
- Pinned host memory is used when available. If it is unavailable, the synchronous path falls back to pageable host memory.
- DFS objects are read through batched `Client::BatchGet`.
- Requested ranges are scattered from staging into registered host or device destination buffers.
- Duplicate keys in one request are read once and scattered to each requested destination.
- Staging is released after the synchronous read and scatter operation completes.
- No session object cache or persistent asynchronous H2D stream is introduced.

The Python API documentation has been updated to describe MEMORY/DFS replica selection, staging lifetime, and pageable-memory fallback behavior.

## Module

- [ ] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [x] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

The DFS session path was built and tested in a Linux container using the TCP transport and POSIX DFS adapter.

Commands used for DFS validation:

```bash
  mooncake_store \
  pybind_client_test \
  client_integration_test \
  dfs_posix_test \
  dfs_sync_client_test \
  -j8

./build/mooncake-store/tests/pybind_client_test \
  --gtest_filter='RealClientTest.SessionRangesReadDfsAndPropagateShortRead'

ctest --test-dir build \
  -R '^(dfs_posix_test|dfs_sync_client_test)$' \
  --output-on-failure

cd docs
make html
```

The latest scope-reduction commit only removes the unverified `LOCAL_DISK` and `DISK` session paths. No runtime tests were rerun after this removal.

Static validation performed for the scope-reduction commit:

```bash
git diff --check
```

The remaining session implementation was also inspected to confirm that:

- replica selection contains only MEMORY and DFS;
- the session read plan contains only MEMORY and DFS request groups;
- the `LOCAL_DISK` and `DISK` session executors and declarations were removed;
- the corresponding `LOCAL_DISK` and `DISK` session tests were removed.

## Checklist

- [x] I have performed a self-review of every changed line
- [ ] I have run pre-commit on the files changed in this PR
- [x] I have added tests covering the DFS session read path
- [ ] For changes over 500 LOC, I have filed an RFC issue

## AI Assistance Disclosure

- [x] AI tools were used


fcczzz:
Could you clarify why this path introduces a separate buffer pool instead of reusing `client_buffer_allocator_`, as the existing DFS read paths do?

wp-bc:
> Could you clarify why this path introduces a separate buffer pool instead of reusing `client_buffer_allocator_`, as the existing DFS read paths do?

Thanks for raising this. The separate pool was intentional because ranged sessions are primarily used in L1–L3 direct-transfer workflows, for example transferring data from a DFS-backed replica into an SGLang GPU buffer.
For a DFS replica, the data still has to pass through host staging before being scattered to the registered destination buffers. client_buffer_allocator_ provides Store/Transfer-Engine registered memory, but it does not necessarily guarantee accelerator-pinned host memory. When the destination is GPU memory, staging in pinned host memory can significantly improve the H2D transfer efficiency compared with pageable host memory.
The dedicated PinnedBufferPool also avoids consuming the fixed-size client_buffer_allocator_ arena, which may already be shared by other Store operations. Its lifetime is still request-scoped from the session-read perspective: the staging handle is released after the synchronous DFS read and scatter, while the underlying allocation may be retained by the pool for reuse. If accelerator-pinned allocation is unavailable, it falls back to pageable host memory.
That said, I agree that always using a separate pool introduces another allocation policy and differs from the existing DFS read paths. A reasonable refinement would be to make the staging strategy configurable, or to use pinned staging only when the destination includes device memory and reuse client_buffer_allocator_ for host-only reads. I’m happy to adjust the implementation in that direction if that better matches the project’s preferred design.

fcczzz:
> > Could you clarify why this path introduces a separate buffer pool instead of reusing `client_buffer_allocator_`, as the existing DFS read paths do?
> 
> Thanks for raising this. The separate pool was intentional because ranged sessions are primarily used in L1–L3 direct-transfer workflows, for example transferring data from a DFS-backed replica into an SGLang GPU buffer. For a DFS replica, the data still has to pass through host staging before being scattered to the registered destination buffers. client_buffer_allocator_ provides Store/Transfer-Engine registered memory, but it does not necessarily guarantee accelerator-pinned host memory. When the destination is GPU memory, staging in pinned host memory can significantly improve the H2D transfer efficiency compared with pageable host memory. The dedicated PinnedBufferPool also avoids consuming the fixed-size client_buffer_allocator_ arena, which may already be shared by other Store operations. Its lifetime is still request-scoped from the session-read perspective: the staging handle is released after the synchronous DFS read and scatter, while the underlying allocation may be retained by the pool for reuse. If accelerator-pinned allocation is unavailable, it falls back to pageable host memory. That said, I agree that always using a separate pool introduces another allocation policy and differs from the existing DFS read paths. A reasonable refinement would be to make the staging strategy configurable, or to use pinned staging only when the destination includes device memory and reuse client_buffer_allocator_ for host-only reads. I’m happy to adjust the implementation in that direction if that better matches the project’s preferred design.

Thanks for clarifying. `ClientBufferAllocator` can also manage externally allocated accelerator-pinned memory—the LOCAL_DISK pinned restore arena already uses this pattern. Could we reuse that approach here with a fixed-capacity arena, rather than introduce a separate allocation policy?

wp-bc:
> > > Could you clarify why this path introduces a separate buffer pool instead of reusing `client_buffer_allocator_`, as the existing DFS read paths do?
> > 
> > 
> > Thanks for raising this. The separate pool was intentional because ranged sessions are primarily used in L1–L3 direct-transfer workflows, for example transferring data from a DFS-backed replica into an SGLang GPU buffer. For a DFS replica, the data still has to pass through host staging before being scattered to the registered destination buffers. client_buffer_allocator_ provides Store/Transfer-Engine registered memory, but it does not necessarily guarantee accelerator-pinned host memory. When the destination is GPU memory, staging in pinned host memory can significantly improve the H2D transfer efficiency compared with pageable host memory. The dedicated PinnedBufferPool also avoids consuming the fixed-size client_buffer_allocator_ arena, which may already be shared by other Store operations. Its lifetime is still request-scoped from the session-read perspective: the staging handle is released after the synchronous DFS read and scatter, while the underlying allocation may be retained by the pool for reuse. If accelerator-pinned allocation is unavailable, it falls back to pageable host memory. That said, I agree that always using a separate pool introduces another allocation policy and differs from the existing DFS read paths. A reasonable refinement would be to make the staging strategy configurable, or to use pinned staging only when the destination includes device memory and reuse client_buffer_allocator_ for host-only reads. I’m happy to adjust the implementation in that direction if that better matches the project’s preferred design.
> 
> Thanks for clarifying. `ClientBufferAllocator` can also manage externally allocated accelerator-pinned memory—the LOCAL_DISK pinned restore arena already uses this pattern. Could we reuse that approach here with a fixed-capacity arena, rather than introduce a separate allocation policy?

Thanks for the suggestion. I’ve updated the PR to reuse the existing fixed-capacity pinned restore arena, which is managed through ClientBufferAllocator, and removed the separate PinnedBufferPool from the session path.
For DFS ranged-session reads targeting device memory, staging now first attempts to allocate from the pinned arena configured by MC_STORE_PINNED_RESTORE_ARENA_SIZE_BYTES. If the arena is unavailable or exhausted, it falls back to the regular client_buffer_allocator_. Host-only destinations use client_buffer_allocator_ directly.
This keeps the allocation and lifetime management consistent with the existing LOCAL_DISK restore path: sub-allocations are thread-safe and released through BufferHandle RAII, while the pinned backing memory remains owned by FileStorage.

fcczzz:
Hi, CI is currently failing on the code format check.  Could you apply the formatting fix?

wp-bc:
> Hi, CI is currently failing on the code format check. Could you apply the formatting fix?

Thanks for pointing this out. I’ve applied the formatting fix in commit 77926ac, and the code format check now passes.
The remaining failure is in CTest unit tests (Python 3.12). It is unrelated to this PR: the same failure occurs on the current main branch because FakeBackend in batch_oplog/gc_test.cpp does not implement the newly added pure virtual HaKvBackend::DeleteRange method. This PR does not modify those HA-related files.

fcczzz (approved):
Overall, LGTM — MEMORY-preferred/DFS-fallback replica selection, request-scoped staging released after the synchronous scatter, and full-object slices that keep checksum verification intact.

Before merge, please follow up with pinned-arena/device-destination coverage plus a doc note that unsupported replica types return INVALID_REPLICA at session start.

fcczzz:
I missed #3916 when I reviewed this — flagging the overlap now, since neither has landed.

The two PRs both rewrite `batch_get_session_start` and the ranged-session read path. They touch four of the same files; the substantive overlapping edits are in `real_client.cpp`, `real_client.h`, and `docs/source/api-reference/python/mooncake-store.md` (`pybind_client_test.cpp` is also touched, in different test areas).

Scope-wise, this PR adds MEMORY + DFS session reads. #3916 also rewrites the same path and adds LOCAL_DISK/DISK support, along with its own DFS handling. Since #3658 is specifically the SSD-offload case, `Related to #3658` is the right framing here: this PR does not by itself close it because LOCAL_DISK remains unsupported.

Proposal: land this first, then have #3916 rebase on top, preserve the DFS path from here, and layer its LOCAL_DISK/DISK support onto it. That avoids ending up with two competing implementations of the DFS session path.

ykwd (approved):

PR #3907, opened 2026-09-07
[TENT] Say why the drop and arbitration MLU read different bytes_ahead

SongOf:
## Description

  Comment-only follow-up promised in the #3780 review. `deadline_mlu.h` said the admission drop predictor and the RDMA arbitration "must rank the same request the same way", but the two layers intentionally pass different `bytes_ahead`: admission passes `dispatching_bytes_` (dispatched and not yet completed, including slices still in worker queues), arbitration passes the posted hardware backlog (`getPostedBytes`) plus the slots already ordered ahead. The header now states that each layer counts the bytes it can actually see at its decision point, that the drop side reading the wider set is the intended conservative direction (it is the irreversible decision), and asks that the necessity be weighed carefully before changing the asymmetry either way. No code change.

  ## Module

  ## Type of Change

  - [x] Documentation update

  ## How Has This Been Tested?

  Comment-only change in a header; no behavior affected.

  **Test commands:**
  ```bash
  pre-commit run --files mooncake-transfer-engine/tent/include/tent/runtime/deadline_mlu.h

  Test results:
  - [x] Manual testing done (pre-commit hooks pass; no code changed)
  
  Checklist

  AI Assistance Disclosure

### DIFF
--- a/mooncake-transfer-engine/tent/include/tent/runtime/deadline_mlu.h
+++ b/mooncake-transfer-engine/tent/include/tent/runtime/deadline_mlu.h
@@ -14,9 +14,22 @@
 //
 // The one definition of a request's predicted deadline feasibility (MLU),
 // shared by the admission queue's drop predictor (RFC #2519 step 3) and the
-// RDMA workers' bandwidth arbitration (RFC #2792). Both must rank the same
-// request the same way from the same bandwidth series, or the admission
-// layer can drop a flow the NIC layer just promoted.
+// RDMA workers' bandwidth arbitration (RFC #2792). Both must apply the same
+// formula to the same bandwidth series, or the admission layer can drop a
+// flow the NIC layer just promoted.
+//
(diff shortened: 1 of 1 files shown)


staryxchen (commented):
Maybe we could add a specify doc for this feature.

SongOf:
Agreed. Today the deadline machinery is only described in passing inside `docs/source/design/tent/slice-spraying.md`, which is really the device-selection document. I will follow up with a dedicated design doc covering the admission queue (EDF ordering, promotion, the deadline-infeasible drop and its `runtime_queue/*` knobs) and the RDMA bandwidth arbitration, and set both side by side with the device-selection implementation: which bandwidth series each layer reads (selection EWMA vs. transmit estimate), which `bytes_ahead` each one sees, and why the two predictors are intentionally not identical — the point this header comment makes. I'd keep this PR as the comment-only fix so the header stops contradicting the code, and land the doc separately.

staryxchen (approved):
LGTM

PR #3910, opened 2026-09-07
[TENT] Pin dispatching_bytes_ to the eligible bytes in Dispatching

SongOf:
## Description

  Test follow-up promised in the #3780 review. `dispatching_bytes_` is the queue-ahead term of the step-3 drop prediction and is maintained by hand (`+=` on pick, `-=` on complete, both gated on `degradation_eligible`); existing tests only pin the drop behaviour at specific thresholds, so a leaked or asymmetric update would go unnoticed and make drops progressively more aggressive. This adds a `dispatchingBytes()` accessor and one test that asserts, after every owner state transition (pick, complete, cancel, drop, retireBatch, and the refused variants), that the counter equals the bytes of the eligible owners currently in Dispatching.

  ## Module

  ## Type of Change

  - [x] Other (test)

  ## How Has This Been Tested?

  **Test commands:**
  ```bash
  cmake --build build-tent --target admission_queue_test -j4
  ./build-tent/mooncake-transfer-engine/tent/tests/admission_queue_test

  Test results:
  - [x] Unit tests pass (41/41; removing the -= in complete() makes the new test fail at 1300 != 300)
  
  Checklist

  AI Assistance Disclosure

### DIFF
--- a/mooncake-transfer-engine/tent/include/tent/runtime/admission_queue.h
+++ b/mooncake-transfer-engine/tent/include/tent/runtime/admission_queue.h
@@ -163,6 +163,9 @@ class LocalTransferAdmissionQueue {
 
     size_t outstandingBytes() const;
 
+    // Bytes of degradation-eligible owners currently Dispatching.
+    size_t dispatchingBytes() const;
+
    private:
     enum class QueueState {
         Queued,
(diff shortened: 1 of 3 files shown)


alogfans (approved):
LGTM

PR #3916, opened 2026-09-07
[Bugfix][Store] Read file-backed replicas in session API

### LINKED ISSUE #3658: [Bug]: batch_get_session_start fails with 702 INVALID_REPLICA when SSD offloading is enabled

ascend-direct-dev:
## Description

Fixes #3658.

After #3903, get sessions already support MEMORY (zero-copy) and DFS (staging + scatter). SSD-only / file-backed keys still failed session start with `702 INVALID_REPLICA` because `SelectSessionReplica` and the classify/execute pipeline only handled MEMORY and DFS.

This PR extends that same prepare → classify → execute pipeline (no separate packing layer):

1. **`SelectSessionReplica`**: after MEMORY / DFS, also accept a complete `LOCAL_DISK` or `DISK` replica. NOF-only keys still return `INVALID_REPLICA`.
2. **`classify_session_range_read_requests`**: route `LOCAL_DISK` → `local_disk_requests`, `DISK` → `disk_requests`.
3. **LOCAL_DISK** (`execute_session_local_disk_range_reads`): group by owner endpoint; restore each unique key once via shared `restore_offload_objects(..., allow_pinned=false)` (RPC into the owner's TE-registered client buffer — never the unregistered pinned arena); then `Client::BatchTransferReadOffloadRanges` from restored pointer + `src_offset` into already `register_buffer`'d destinations. The reader does **not** need a setup local buffer (`local_buffer_size=0` works). Same-node GPU destinations also go through TE (no offload `kLocalAddress` memcpy on the session path).
4. **DISK** (`execute_session_disk_range_reads`): one `Client::BatchGet` into a temp host buffer, then scatter by `src_offset`. Keys fail if the client has no allocator.
5. **Shared restore helper**: sequential `batch_get_into_offload_object_internal` now uses the same `restore_offload_objects` / `OffloadRestoreLease` (default `allow_pinned=true` so memcpy into the pinned arena remains available).

Duplicate keys restore once and scatter many times. Invalid ranges do not block valid entries. Lease is checked locally after transfer. Offload GC TTL rewrites only successful scatters to `OBJECT_HAS_LEASE`. No session-level object pin.

## Relation to #3903

#3903 landed the DFS session path and the shared pipeline. This branch was rebased onto main: redundant DFS / `MultiBufferRangeRequest` work from earlier revisions was dropped; only LOCAL_DISK / DISK session reads and the shared offload restore helper remain.

## Module

- [x] Docs

## Type of Change

- [x] Bug fix
- [x] Documentation update

## How Has This Been Tested?

Lima VM `hixl`, `USE_ASCEND_DIRECT=ON`, `MC_FORCE_TCP=1` (TCP-only; Ascend transport install is skipped).

**Test commands:**
```bash
source /Users/pkgs/Ascend/cann/bin/setenv.bash
cd build
make pybind_client_test
MC_FORCE_TCP=1 ./mooncake-store/tests/pybind_client_test \
  --gtest_filter='*Session*:*Ssd*:*SSD*:*Offload*:*LocalDisk*'
```

**Test results:**
- [x] Unit tests pass

Includes `RealClientTest.TestGetSessionRangesFromSsdOffload` (reader `local_buffer_size=0`, mixed SSD+MEMORY, duplicate-key restore-once, invalid/empty ranges, lease expiry) and existing session tests.

Not run: RDMA, CUDA `PinnedSsdRestoreReadsNonTailRangeIntoGpu` (`MOONCAKE_TEST_CUDA_H2D` off), DISK/DFS session e2e (no-allocator path covered by fail-on-allocate).

## Checklist

Tracking issue is #3658 (no separate RFC). clang-format 20 was applied to the touched C++ files.

## AI Assistance Disclosure

Cursor / Claude assisted with the rebase onto #3903, the LOCAL_DISK / DISK session execute paths, shared `restore_offload_objects`, tests/docs, and this PR description. The human submitter is responsible for reviewing and defending the changes.

### DIFF
--- a/mooncake-store/src/real_client.cpp
+++ b/mooncake-store/src/real_client.cpp
@@ -11,9 +11,15 @@
 #include <dlfcn.h>  // for dlsym (Python detection)
 #include <cstdlib>  // for atexit
 #include <algorithm>
+#include <atomic>
+#include <condition_variable>
 #include <functional>
 #include <limits>
+#include <mutex>
 #include <optional>
+#include <span>
+#include <unordered_map>
+#include <unordered_set>
 #include <vector>
 
 #include "real_client.h"
@@ -414,35 +420,59 @@ inline bool is_object_range_overflow(size_t offset, size_t size, size_t limit) {
     return size > limit || offset > limit - size;
 }
 
-inline const Replica::Descriptor *SelectCompleteMemoryReplica(
-    const std::vector<Replica::Descriptor> &replicas,
-    const std::unordered_set<std::string> &local_endpoints) {
-    const Replica::Descriptor *first_memory = nullptr;
-    for (const auto &r : replicas) {
-        if (r.status != ReplicaStatus::COMPLETE || !r.is_memory_replica()) {
-            continue;
-        }
-        if (local_endpoints.count(r.get_memory_descriptor()
-                                      .buffer_descriptor.transport_endpoint_)) {
-            return &r;
-        }
-        if (!first_memory) {
-            first_memory = &r;
-        }
-    }
-    return first_memory;
-}
-
-inline bool HasMemoryReplica(const std::vector<Replica::Descriptor> &replicas) {
-    for (const auto &r : replicas) {
-        if (r.is_memory_replica()) {
-            return true;
-        }
-    }
-    return false;
-}
-
 }  // namespace
+
+class RealClient::OffloadRestoreLease {
+   public:
+    OffloadRestoreLease(
+        RealClient *owner, std::string endpoint, bool local_batch,
+        std::optional<FileStorage::LocalBatchResult> local_owner,
+        BatchGetOffloadObjectResponse response,
+        std::chrono::steady_clock::time_point start_time)
+        : owner_(owner),
+          endpoint_(std::move(endpoint)),
+          local_batch_(local_batch),
+          local_owner_(std::move(local_owner)),
+          response_(std::move(response)),
+          start_time_(start_time) {}
+
+    ~OffloadRestoreLease() {
+        if (owner_ == nullptr || local_batch_ ||
+            owner_->client_requester_ == nullptr) {
+            return;
+        }
+        owner_->client_requester_->release_offload_buffer(endpoint_,
+                                                          response_.batch_id);
+    }
+
+    OffloadRestoreLease(const OffloadRestoreLease &) = delete;
+    OffloadRestoreLease &operator=(const OffloadRestoreLease &) = delete;
+    OffloadRestoreLease(OffloadRestoreLease &&) = delete;
+    OffloadRestoreLease &operator=(OffloadRestoreLease &&) = delete;
+
+    bool local_batch() const { return local_batch_; }
+    BatchGetOffloadObjectResponse &response() { return response_; }
+    const BatchGetOffloadObjectResponse &response() const { return response_; }
+
+    uint64_t ElapsedMs() const {
+        return static_cast<uint64_t>(
+            std::chrono::duration_cast<std::chrono::milliseconds>(
+                std::chrono::steady_clock::now() - start_time_)
+                .count());
+    }
+
+    bool LeaseExpired() const {
+        return !local_batch_ && ElapsedMs() >= response_.gc_ttl_ms;
+    }
+
+   private:
+    RealClient *owner_;
+    std::string endpoint_;
+    bool local_batch_;
+    std::optional<FileStorage::LocalBatchResult> local_owner_;
+    BatchGetOffloadObjectResponse response_;
+    std::chrono::steady_clock::time_point start_time_;
+};
 
 PyClient::~PyClient() {}
 
@@ -5502,9 +5532,8 @@ std::vector<int> RealClient::batch_get_session_start(
         return {};
     }
 
-    // Master interaction only here: query replicas + lease.
     const auto query_results = client_->BatchQuery(keys);
-    auto local_endpoints = client_->GetLocalEndpoints();
+    const auto local_endpoints = client_->GetLocalEndpoints();
 
     std::lock_guard<std::mutex> lock(session_mutex_);
     for (size_t i = 0; i < keys.size(); ++i) {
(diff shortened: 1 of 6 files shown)


mooncake-store/src/real_client.cpp
```diff
@@ -442,6 +427,54 @@ inline bool HasMemoryReplica(const std::vector<Replica::Descriptor> &replicas) {
     return false;
 }
 
+class SessionScatterBuilder {
```
Aionw: Could we move this out of `real_client.cpp`, this file is large enough.

Aionw (approved):

fcczzz:
I missed this while reviewing #3903 — flagging the overlap now, since neither has landed.

The two PRs both rewrite `batch_get_session_start` and the ranged-session read path. They touch four of the same files, with substantive overlapping edits in `real_client.cpp`, `real_client.h`, and `docs/source/api-reference/python/mooncake-store.md`.

Proposal: land #3903 first, then rebase this on top and preserve its DFS session path while adding the LOCAL_DISK/DISK coverage here. The LOCAL_DISK path in this PR is the part that directly addresses the SSD-offload failure in #3658, so please keep that coverage.

One correctness issue before merge: at this PR's base, `SelectBestReplica` returns immediately on the first local MEMORY or local NoF replica. That means `[local NoF, local MEMORY]` selects the NoF replica. Since `batch_get_session_start` now uses `SelectBestReplica`, it can cache that NoF replica; the ranged-read path later accepts only MEMORY, LOCAL_DISK, DISK, or DFS, so it returns `INVALID_REPLICA` even though a usable local MEMORY replica exists.

#3662 fixes the selection ordering so local MEMORY wins regardless of replica order and adds a regression test. Please either rebase on that fix before merge or carry the equivalent ordering fix and regression coverage here.

ascend-direct-dev:
@fcczzz thanks for the heads-up.

We've already rebased onto current `main` after #3903 and #3662 landed, and updated this PR to match the latest session pipeline / replica selection. The LOCAL_DISK/DISK session coverage for #3658 is kept on top of that; the earlier overlapping rewrite is gone.

mooncake-store/src/real_client.cpp
```diff
@@ -6599,6 +6665,194 @@ void RealClient::execute_session_dfs_range_reads(
...
+                    results[request->result_index] =
+                        static_cast<int>(toInt(ErrorCode::INTERNAL_ERROR));
+                }
+            }
+            continue;
+        }
+        for (size_t k = 0; k < transfer.size(); ++k) {
+            const auto *request = entry_requests[k];
+            if (transfer[k]) {
+                if (lease.LeaseExpired()) {
```
fcczzz: This checks the restored buffer's GC TTL, but unlike the MEMORY and DFS session paths, it does not recheck the Get Session lease after the transfer. If that lease expires during the read, LOCAL_DISK can return a successful byte count while the other paths return LEASE_EXPIRED. Is this difference intentional? If so, could we document the expected behavior?

PR #3920, opened 2026-09-07
[Fix] TENT gid lid refresh

TTThanos:
## Description

Related to #3523, Defect D only. 

TENT currently reads an RNIC's LID, GID, and GID index once in RdmaContext::openDevice() and reuses those cached values for later bootstrap replies and QP address-handle setup. If the address changes after a link flap or network reconfiguration, peers can receive stale Segment metadata or bootstrap data and build QPs against an obsolete address, causing RTR failures, transfer timeouts, or reconnect loops.

This PR adds runtime RDMA address refresh and makes recovery fail closed:

1. Handle IBV_EVENT_GID_CHANGE, IBV_EVENT_LID_CHANGE, and IBV_EVENT_PORT_ACTIVE by temporarily pausing the context and removing the NIC from DeviceSelector.
2. Require the hardware port to be fully IBV_PORT_ACTIVE, then re-query its LID, GID, and GID index. Auto-GID mode repeats the initial selection policy; an explicitly configured GID index remains fixed while its value is refreshed.
3. Update the local Segment metadata before exposing the new address locally. If the NIC was omitted during startup because its port was down, recovery upserts it even when the address value itself did not change.
4. Keep the NIC paused if querying or metadata publication fails, preserving the previous local address. The local metadata update is rolled back when registry synchronization reports failure.
5. Retire all endpoints built with the previous address before changing the context back to DEVICE_ENABLED and making it selectable again.

The refresh sequence is:

GID_CHANGE / LID_CHANGE / PORT_ACTIVE
  -> pause context and remove NIC from DeviceSelector
  -> verify port is ACTIVE
  -> query current LID, GID, and gid_index
  -> publish/upsert the address in Segment metadata
  -> retire endpoints using the old address
  -> set context ENABLED and re-enable NIC selection

**Address-generation consistency**
A refresh can race with bootstrap and endpoint setup, so reading LID, GID, and GID index independently is unsafe. This PR introduces RdmaAddressSnapshot and protects address reads with a mutex, ensuring each snapshot contains values from one generation.

Each RdmaEndPoint captures one immutable local address snapshot when it is constructed. Bootstrap descriptors, data QP setup, and notification QP setup all use that same snapshot and GID index. A context refresh evicts the complete endpoint before the NIC becomes selectable again, preventing a new address from being paired with QPs created from the previous generation.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

The focused coverage verifies:

GID change refreshes both the context and published Segment metadata.
LID change refreshes both the context and published Segment metadata.
Address-change events for another port are ignored.
A GID/LID event while the port is still down does not re-enable the NIC.
A query failure preserves the old address and keeps the NIC paused.
A NIC omitted from startup metadata is inserted on recovery even when its address is unchanged.
PORT_ACTIVE refreshes and publishes the address before re-enabling the NIC.
An endpoint keeps the address generation captured at construction time.
The existing real-RNIC recovery test passes on mlx5/CX8 hardware.

**Test commands:**
```bash
```

**Test results:**
- [ ] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

### DIFF
--- a/mooncake-transfer-engine/tent/include/tent/transport/rdma/context.h
+++ b/mooncake-transfer-engine/tent/include/tent/transport/rdma/context.h
@@ -43,6 +43,18 @@ class RdmaEndPoint;
 class EndpointStore;
 class RdmaTransport;
 
+struct RdmaAddressSnapshot {
+    uint16_t lid = 0;
+    std::string gid;
+    int gid_index = -1;
+};
+
+enum class RdmaAddressRefreshResult {
+    UNCHANGED = 0,
+    CHANGED = 1,
+    FAILED = 2,
+};
+
 class RdmaContext {
     friend class RdmaCQ;
     friend class RdmaEndPoint;
@@ -101,11 +113,22 @@ class RdmaContext {
     const std::string name() const { return device_name_; }
 
    public:
-    uint16_t lid() const { return lid_; }
+    // Take lid/gid/index under one lock so a bootstrap never mixes values from
+    // two address generations while the monitor thread refreshes the port.
+    RdmaAddressSnapshot address() const;
 
-    std::string gid() const;
+    uint16_t lid() const { return address().lid; }
 
-    int gidIndex() const { return gid_index_; }
+    std::string gid() const { return address().gid; }
+
+    int gidIndex() const { return address().gid_index; }
+
+    // Re-query the port's LID and GID. Auto-GID mode repeats the initial
+    // selection; an explicit gid_index keeps that index and refreshes its
+    // value. The new address is published before it becomes visible locally.
+    RdmaAddressRefreshResult refreshAddress(
+        RdmaAddressSnapshot *previous = nullptr,
+        RdmaAddressSnapshot *current = nullptr);
 
     ibv_context *nativeContext() const { return native_context_; }
 
@@ -193,6 +216,11 @@ class RdmaContext {
     size_t num_comp_channel_ = 0;
     std::vector<ibv_comp_channel *> comp_channel_;
 
+    // The monitor thread may refresh these while bootstrap handlers read
+    // them. address_mutex_ keeps each lid/gid/index snapshot self-consistent;
+    // address_refresh_mutex_ serializes hardware reprobes and publication.
+    mutable std::mutex address_mutex_;
+    std::mutex address_refresh_mutex_;
     uint16_t lid_ = 0;
     // Set by openDevice() and refreshed by refreshPortAttributes() on the
     // monitor thread. Today every runtime reader is that same thread;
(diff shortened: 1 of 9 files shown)


mooncake-transfer-engine/tent/include/tent/transport/rdma/context.h
```diff
@@ -193,6 +216,11 @@ class RdmaContext {
     size_t num_comp_channel_ = 0;
     std::vector<ibv_comp_channel *> comp_channel_;
 
+    // The monitor thread may refresh these while bootstrap handlers read
+    // them. address_mutex_ keeps each lid/gid/index snapshot self-consistent;
+    // address_refresh_mutex_ serializes hardware reprobes and publication.
+    mutable std::mutex address_mutex_;
+    std::mutex address_refresh_mutex_;
```
alogfans, line 223: Why require two types of mutex? This seems a standard rwlock.

mooncake-transfer-engine/tent/src/transport/rdma/workers.cpp
```diff
@@ -1266,12 +1282,44 @@ void Workers::applyContextEvent(int dev_id, RdmaContext& context,
...
+    // resume() evicts every endpoint built with the previous address before
+    // the NIC becomes selectable again. This is deliberately conservative for
+    // a spurious change event whose queried value is unchanged.
+    if (context.resume() != 0) return false;
     // The link may have renegotiated while down: re-seed before the device
     // becomes selectable so no worker scores it on the old rate.
     refreshLinkSpeed(dev_id, context);
     if (device_selector_) device_selector_->setDeviceAvailable(dev_id, true);
+    return true;
 }
```
alogfans, line 1331: activateContext() currently ignores the result of pause() and proceeds with the hardware query and metadata publication. Under the normal event path the context should be either enabled or already paused, but if this races with teardown or is called in another state, it could publish a device address even though resume() will necessarily fail. Please verify that the resulting state is DEVICE_PAUSED before calling refreshAddress(). Note that the current pause() return-value logic also appears inverted for a successful compare-exchange, so checking the state directly—or fixing that return contract—would be safer.

alogfans (approved):
LGTM. Please refer the above non-blocking comment.

mooncake-transfer-engine/tent/include/tent/transport/rdma/context.h
TTThanos: The two mutexes protect different scopes: address_mutex_ only guards the short critical section for reading or updating the LID/GID/index snapshot, while address_refresh_mutex_ serializes the full refresh transaction, including the verbs queries and metadata publication. Using a single rwlock would either require holding the write lock across potentially blocking metadata I/O, or release it early and allow concurrent refresh transactions. The split keeps address reads short while ensuring only one refresh is in flight.

mooncake-transfer-engine/tent/src/transport/rdma/workers.cpp
TTThanos: Solved

PR #3922, opened 2026-09-07
[Store] Optimize CRC-64 checksum calculation with slice-by-8

SnowLegend-star:
## Description
CrcCheckSum::Update processed one byte per iteration, and every table index depended on the  previous iteration's register value. That serial dependency chain costs `~3.13` ns/byte(about `320MB/s` per thread)on an Ice Lake Xeon 8350C, which makes the optional object checksum be the throughout bottleneck well below the nic line rate.

Fold eight byte steps into a single iteration using the standard slicing-by-8 tables. The eight lookups are independent, so they work in parallel:0.65 ns/byte, about 4.8x faster. Tables grow from `2KiB` to `16 KiB`, still comfortably inside L1d. 

The computed checksum is unchanged, so sorted checksums remain verifiable and mixed-version deployments are safe. The tests now compare against an independent bit-at-a-time reference implementation across every tail length, every streaming split point, and unaligned  inputs.

Measured end to end with `MOONCAKE_STORE_CHECKSUM=1` on a 100 Gb RoCE loopback(16 clients,1 MiB objects): read throughput loss versus checksum disabled drops from `61.8%` to `0.8%`, with 8 MiB objects, form `61.0%` to `0.6%`

## Future work: Intel ISA-L

Slicing-by-8 removes the checksum bottleneck at moderate concurrency, but it is
still ~8x away from the memory bandwidth of a single core. Intel ISA-L ships a
PCLMULQDQ folding implementation of the same polynomial (`crc64_ecma_norm`),
which measures 0.088 ns/byte (11.4 GB/s) per thread on the same Ice Lake Xeon
8350C -- 7.4x faster than slicing-by-8 and 35x faster than the current
byte-at-a-time loop. At that speed the checksum is no longer compute bound but
memory bound: the same benchmark reports 0.041 ns/byte (24 GB/s) when the buffer
stays resident in cache.

Two consequences are worth noting.

First, the number of client threads required to keep a 100 Gb NIC saturated
while checksumming drops sharply. Using the per-operation latency model
(fixed overhead + transfer + checksum, at 1 MiB objects):

| implementation | ns/byte | per-thread throughput | threads to saturate 12.6 GB/s |
|---|---|---|---|
| byte-at-a-time (before) | 3.13 | 0.31 GB/s | ~41 |
| slicing-by-8 (this PR) | 0.65 | 1.28 GB/s | ~10 |
| ISA-L `crc64_ecma_norm` | 0.088 | 4.74 GB/s | ~3 |

Second, and probably more relevant in production: sustaining line rate costs
roughly 39 cores of CRC work with the current implementation, 8.2 cores with
slicing-by-8, and 1.1 cores with ISA-L. Mooncake clients are typically colocated
with inference workers, so CPU cycles spent on checksums are taken directly from
the model server, not from idle capacity. That saving applies even where
throughput is already at line rate and the loss figures above read 0%.

I verified that `~crc64_ecma_norm_by8(~crc, buf, len)` reproduces the value
produced by this PR's implementation bit for bit, so adopting ISA-L would also
preserve stored checksums.

The caveats are why this is future work rather than part of this PR:

- it adds a third-party dependency;
- the accelerated path requires nasm >= 2.11.01 (or yasm >= 1.2.0) at build time
  and is x86-64 only. ISA-L's portable fallback (`crc64_ecma_norm_base`) is
  slower than slicing-by-8, so a build without an assembler would regress unless
  the fallback selects this PR's implementation;
- AArch64 would need a separate PMULL path to get comparable numbers.

A reasonable shape would be an optional `WITH_ISAL` build flag that falls back to
the slicing-by-8 implementation added here. I am happy to prepare that as a
follow-up if maintainers are interested.

Note on measurement scope: the ISA-L numbers above come from a standalone
micro-benchmark of the checksum routine, not from a build with ISA-L wired into
Mooncake. The per-thread throughput and thread-count columns are projections
from the same latency model that predicted the slicing-by-8 end-to-end results
within 7% at 1-4 clients (it was 29% optimistic at 8 clients, where queueing at
the NIC begins). The end-to-end figures reported earlier in this PR are all
measured.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [x] Other

## How Has This Been Tested?

**Test commands:**
```bash
# Unit
cmake --build /workspace/mooncake-build --target crc_checksum_test -j
/workspace/mooncake-build/mooncake-common/tests/crc_checksum_test

# End-to-end A/B/C: checksum off / byte-at-a-time / slicing-by-8
# same binary, switched by MOONCAKE_CRC_IMPL, RDMA (mlx5_1)
bash run_slice8_ab.sh
```

**Test results:**
- [x] Unit tests pass

## Checklist

## AI Assistance Disclosure

- [ ] No AI tools were used

### DIFF
--- a/mooncake-common/src/crc_checksum.cpp
+++ b/mooncake-common/src/crc_checksum.cpp
@@ -54,7 +54,7 @@ inline uint64_t LoadBigEndian64(const uint8_t* p) {
 }
 
 inline uint64_t UpdateByteAtATime(uint64_t crc, const uint8_t* bytes,
-                                 size_t size) {
+                                  size_t size) {
     for (size_t i = 0; i < size; ++i) {
         const auto index = static_cast<uint8_t>((crc >> 56) ^ bytes[i]);
         crc = kCrc64EcmaTable[index] ^ (crc << 8);
@@ -67,10 +67,11 @@ inline uint64_t UpdateByteAtATime(uint64_t crc, const uint8_t* bytes,
 void CrcChecksum::Update(const void* data, size_t size) {
     const auto* bytes = static_cast<const uint8_t*>(data);
 
-    // The byte-at-a-time loop is latency bound: every table index depends on the
-    // previous iteration's register value. Slicing-by-8 folds eight byte steps
-    // into one, and the eight lookups are independent, so the CPU can issue them
-    // in parallel. Measured ~4.8x on an Ice Lake Xeon (3.13 -> 0.65 ns/byte).
+    // The byte-at-a-time loop is latency bound: every table index depends on
+    // the previous iteration's register value. Slicing-by-8 folds eight byte
+    // steps into one, and the eight lookups are independent, so the CPU can
+    // issue them in parallel. Measured ~4.8x on an Ice Lake Xeon (3.13 -> 0.65
+    // ns/byte).
     uint64_t crc = crc_;
     while (size >= 8) {
         crc ^= LoadBigEndian64(bytes);
--- a/mooncake-common/src/crc_checksum.cpp
+++ b/mooncake-common/src/crc_checksum.cpp
@@ -8,29 +8,84 @@ namespace {
 
 constexpr uint64_t kCrc64EcmaPolynomial = 0x42F0E1EBA9EA3693ULL;
 
-constexpr std::array<uint64_t, 256> MakeCrc64EcmaTable() {
-    std::array<uint64_t, 256> table{};
-    for (size_t i = 0; i < table.size(); ++i) {
+// Advance the register by one byte with a zero input byte. Feeding one message
+// byte is `(crc << 8) ^ table[0][(crc >> 56) ^ byte]`, so composing this
+// function k times yields the table for a byte that still has k more byte
+// positions to travel.
+constexpr uint64_t AdvanceOneByte(uint64_t crc, const uint64_t* base_table) {
+    return (crc << 8) ^ base_table[crc >> 56];
+}
+
+// Slicing-by-8 tables. tables[0] is the classic byte-at-a-time table; tables[k]
+// additionally advances the value by k more bytes. Total size is 16 KiB, which
+// fits comfortably in L1d.
+constexpr std::array<std::array<uint64_t, 256>, 8> MakeCrc64EcmaTables() {
+    std::array<std::array<uint64_t, 256>, 8> tables{};
+    for (size_t i = 0; i < 256; ++i) {
         uint64_t crc = static_cast<uint64_t>(i) << 56;
         for (int bit = 0; bit < 8; ++bit) {
             crc = (crc & (1ULL << 63)) != 0 ? (crc << 1) ^ kCrc64EcmaPolynomial
                                             : crc << 1;
         }
-        table[i] = crc;
+        tables[0][i] = crc;
     }
-    return table;
+    for (size_t k = 1; k < 8; ++k) {
+        for (size_t i = 0; i < 256; ++i) {
+            tables[k][i] = AdvanceOneByte(tables[k - 1][i], tables[0].data());
+        }
+    }
+    return tables;
 }
 
-constexpr auto kCrc64EcmaTable = MakeCrc64EcmaTable();
+constexpr auto kCrc64EcmaTables = MakeCrc64EcmaTables();
+constexpr const auto& kCrc64EcmaTable = kCrc64EcmaTables[0];
+
+// CRC-64/ECMA is defined MSB-first, so the first byte of the message must land
+// in the most significant byte of the register. Written as explicit shifts to
+// stay endian-independent; compilers fold this into a load plus a byte swap.
+inline uint64_t LoadBigEndian64(const uint8_t* p) {
+    return (static_cast<uint64_t>(p[0]) << 56) |
+           (static_cast<uint64_t>(p[1]) << 48) |
+           (static_cast<uint64_t>(p[2]) << 40) |
+           (static_cast<uint64_t>(p[3]) << 32) |
+           (static_cast<uint64_t>(p[4]) << 24) |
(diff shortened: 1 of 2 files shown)


he-yufeng (commented):
Read the full diff. The slicing-by-8 fold is implemented correctly: the eight tables are derived by repeated single-byte advancement (`tables[k] = advance(tables[k-1])` from the classic table), the big-endian load is written as explicit shifts so it stays endian-independent, the group loop only runs while at least 8 bytes remain, and the tail falls back to the byte-at-a-time loop. The streaming case is also sound: folding eight bytes is valid from any register state, so a mid-group split followed by another `Update` still composes.

The tests are the right shape for this change — an independent bit-at-a-time reference (so a table bug can't be masked by a shared bug), every length modulo 8, every split point, and unaligned input. That is the coverage this optimization actually needs.

CI was still queued when I read it; the code itself is correct as written.

Prepared with AI assistance (Kimi); I verified the fold and the table derivation against the diff myself.

SnowLegend-star:
> Read the full diff. The slicing-by-8 fold is implemented correctly: the eight tables are derived by repeated single-byte advancement (`tables[k] = advance(tables[k-1])` from the classic table), the big-endian load is written as explicit shifts so it stays endian-independent, the group loop only runs while at least 8 bytes remain, and the tail falls back to the byte-at-a-time loop. The streaming case is also sound: folding eight bytes is valid from any register state, so a mid-group split followed by another `Update` still composes.
> 
> The tests are the right shape for this change — an independent bit-at-a-time reference (so a table bug can't be masked by a shared bug), every length modulo 8, every split point, and unaligned input. That is the coverage this optimization actually needs.
> 
> CI was still queued when I read it; the code itself is correct as written.
> 
> Prepared with AI assistance (Kimi); I verified the fold and the table derivation against the diff myself.

Could you re-run the failed CI job when you get a chance?

`master_service_ssd_test_for_snapshot` failed (81/175), but it looks unrelated to
this change:

- This PR only touches `mooncake-common/src/crc_checksum.cpp` and its test. The
  only caller of `CrcChecksum` is `mooncake-store/src/client_service.cpp` on the
  client object-verification path; this test covers master-side snapshot and
  restore and does not reference the checksum code at all.
- I rebuilt and ran the test locally with this patch applied and all 8 cases
  pass (`[  PASSED  ] 8 tests`, 61 s on an otherwise idle machine).
- The test waits on real lease expiry (`sleep_for(kv_lease_ttl)` with a 2000 ms
  margin in `EvictObject`, and 1 s polling loops in `PutStartExpires`), so it
  looks timing sensitive under a loaded CI runner. It takes 44 s on its own here,
  while the full 175-test suite finished in 517 s, so tests appear to run in
  parallel and contend for CPU.

Happy to dig in if it reproduces on the re-run -- in that case, could you point
me at the failing assertion? The ctest summary in the log does not show which
`EXPECT_*` tripped.

Aionw:
ISA-L v2.31.1 already includes an AArch64 PMULL implementation of `crc64_ecma_norm`: https://github.com/intel/isa-l/blob/v2.31.1/crc/aarch64/crc64_ecma_norm_pmull.S

So the AArch64 caveat in the description appears outdated; we should be able to reuse the existing implementation rather than write a separate PMULL path.

SnowLegend-star:
> ISA-L v2.31.1 already includes an AArch64 PMULL implementation of `crc64_ecma_norm`: https://github.com/intel/isa-l/blob/v2.31.1/crc/aarch64/crc64_ecma_norm_pmull.S
> 
> So the AArch64 caveat in the description appears outdated; we should be able to reuse the existing implementation rather than write a separate PMULL path.

You're right, and thanks for the pointer -- that caveat was my mistake. I tested
against an internal mirror of ISA-L 2.18.0, which predates the AArch64 work
(there is no `crc/aarch64/` directory at all in that tree), and I wrote the
caveat from what I had in front of me instead of checking current upstream.
Please read that bullet as withdrawn.

ISA-L is indeed the follow-up I have in mind. I would still like to land
slicing-by-8 first, for two reasons:

1. It needs no new dependency. At realistic concurrency the remaining gap is
   already small: with 16 clients and 1 MiB objects the read throughput loss
   versus checksum-disabled is 0.8%, and 0.6% at 8 MiB. ISA-L's marginal value
   is concentrated at low concurrency (1-8 clients, where a single thread cannot
   both saturate the NIC and checksum) and in CPU time freed for colocated
   inference workers -- both worth having, but neither is blocking.

2. This implementation is what an ISA-L integration would fall back to anyway.
   ISA-L's portable C path (`crc64_ecma_norm_base`) is bit-at-a-time, i.e.
   slower than what this PR replaces, so a `WITH_ISAL=OFF` build or an
   architecture without an accelerated path would regress unless the fallback is
   something like this. Landing it now means the ISA-L PR can be purely additive
   and gated behind a build flag, rather than having to carry a portable
   implementation of its own.

So I see the two as complementary rather than alternatives. Happy to start on the
ISA-L integration once this is in -- and given the AArch64 implementation already
exists upstream, that follow-up is simpler than I had assumed.

yokinoshitayoki (approved):
Looks good to me, I think we can merge it after the CI checks passed.

SnowLegend-star:
> Looks good to me, I think we can merge it after the CI checks passed.

Could you re-run the failed CI job when you get a chance? I think these two failed jobs were not caused by my code, maybe the system pipeline's problem :)

SnowLegend-star:
> Looks good to me, I think we can merge it after the CI checks passed.

All checks passed and approved, ready to merge.

SnowLegend-star:
@yokinoshitayoki If you could merge my code as soon as you get a chance, I'd really appreciate it—then I can start integrating the ISA-L library for further optimization.

PR #3923, opened 2026-09-07
[Store] Fail-stop supervisor on OpLog writer terminal failure

Icedcoco:
## Description

When the batch OpLog writer reaches a terminal failure, the HA supervisor now fail-stops serving: mutation admission is closed through admin availability, the RPC delegate and leader label are removed, runtime state becomes standby, and the RPC server is stopped. Terminal notifications are delivered outside writer locks and duplicate notifications are ignored.

Related: Oplog HA PR-F02 design.

## Module

- [x] Mooncake Store

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

**Test commands:**
```bash
pre-commit run --files mooncake-store/include/ha/oplog/ordered_oplog_writer.h mooncake-store/src/ha/oplog/ordered_oplog_writer.cpp mooncake-store/include/master_service.h mooncake-store/src/master_service.cpp mooncake-store/include/rpc_service.h mooncake-store/src/rpc_service.cpp mooncake-store/src/ha/leadership/master_service_supervisor.cpp
git diff --check
```

**Test results:**
- [x] Pre-commit passed
- [ ] Unit tests pass (build directory unavailable in this worktree)
- [ ] Integration tests pass
- [ ] Manual testing done

## Checklist

- [x] Self-review completed
- [x] Formatting and pre-commit checks passed
- [x] AI tools were used; the submitter must review every changed line

## AI Assistance Disclosure

Codex assisted with implementation and validation.

### DIFF
--- a/mooncake-store/include/ha/oplog/ordered_oplog_writer.h
+++ b/mooncake-store/include/ha/oplog/ordered_oplog_writer.h
@@ -83,6 +83,7 @@ class OrderedOpLogWriter {
     bool IsAccepting() const;
     ErrorCode LastError() const;
     std::optional<OrderedOpLogWriterTerminalState> GetTerminalState() const;
+    void SetTerminalCallback(TerminalCallback callback);
     void Start();
     virtual void Stop();
 
--- a/mooncake-store/include/master_service.h
+++ b/mooncake-store/include/master_service.h
@@ -204,6 +204,8 @@ class MasterService {
     ErrorCode SetBatchOpLogBackendForTesting(
         std::shared_ptr<HaKvBackend> backend);
     void SetBatchOpLogWriterFactoryForTesting(BatchOpLogWriterFactory factory);
+    void SetBatchOpLogTerminalCallback(
+        OrderedOpLogWriter::TerminalCallback callback);
 
     /**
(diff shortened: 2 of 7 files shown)


he-yufeng (commented):
Read end to end and verified locally. The four invariants I cared about all hold: the terminal callback fires outside the writer lock in `EnterTerminalState`, the setter's late-join path fires immediately for an already-terminal writer (also outside the lock), double notification is filtered twice (the `terminal_state` early-return plus the supervisor's `exchange`), and `serving_gate` closes the activate-after-terminal race in `RunSupervisorLoop`.

Built this head (ef8447a9) in a Linux container and ran both suites against it: master_service_ha_test 93/93 and master_service_test 72/72, all green.

One consistency note from the adjacent restore path: #3806's repair block treats a writer terminal failure as a restore failure (the durable finalize errors, so the candidate never serves), which composes with this fail-stop — a candidate that cannot durably repair does not serve, and a leader whose writer dies now steps down here.

Leaving the approval to another maintainer since this changes the serving gate, but from where I sit this is ready.

Aionw (approved):
LGTM

ykwd (approved):

PR #3924, opened 2026-09-07
[Bugfix][TENT] Serialize lanes sharing a queue pair on its slice queue

SongOf:
## Description
  
  A `qp_pools` layout with fewer QPs than lanes folds several worker lanes onto one queue pair. `submitSlices` and `acknowledge` then reach the same `BoundedSliceQueue` under the endpoint's shared read guard, which does not exclude them from each other, and the queue itself is unsynchronized: concurrent push/pop tears the ring (crash), and a post that interleaves with another lane's post/push puts the queue out of the QP's posting order, so `acknowledge()` — which retires everything up to a completed slice — stamps a later slice with an earlier one's outcome. The comment on `slice_queue_` already named this race; #3780's shared-QP accounting fixes assumed the queue was safe.

  This adds one `TicketLock` per QP (`queue_lock_list_`, allocated and freed with `wr_depth_list_`). `submitSlices` holds it from `ibv_post_send` through the enqueue loop so the queue mirrors the hardware order; `acknowledge` holds it across the queue walk. The lock is uncontended in the default one-lane-per-QP layout; the error log is kept outside it. The default path is otherwise unchanged. Follow-up left open by #3780.

  ## Module

  ## Type of Change

  - [x] Bug fix

  ## How Has This Been Tested?

  Two new hardware-free tests in `RdmaWorkersSharedQpTest` run the real `submitSlices`/`acknowledge` against fake QPs (a fake `ibv_post_send` via `qp->context->ops`, which also records the hardware posting order):
  - `TwoLanesOnOneQueuePairKeepTheSliceQueueIntact`: one lane posts 20k slices while another acknowledges them; every slice completes once, queue / WR depth / inflight end at zero. Segfaults 3/3 without the fix.
  - `TwoPostersOnOneQueuePairKeepPostingOrder`: two lanes post concurrently; replaying completions in the recorded post order must pop exactly one slice each time. With `ibv_post_send` moved outside the lock it fails 3/3
  (19923–19939 of 20000) without crashing — the ordering bug on its own.

  **Test commands:**
  ```bash
  cmake --build build-tent --target tent_rdma_transport_test tent_endpoint_lifecycle_test tent_endpoint_store_test tent_rdma_async_event_drain_test rdma_cancel_test -j4
  ./build-tent/mooncake-transfer-engine/tent/tests/tent_rdma_transport_test

  Test results:
  - [x] Unit tests pass (tent_rdma_transport_test 55 passed / 2 skipped (hardware); endpoint lifecycle 20, endpoint store 8, async event drain 5, rdma_cancel 2 all pass)
  - [x] Manual testing done (uncontended lock+unlock measured at ~3 ns in the build container)
  
  Checklist

  AI Assistance Disclosure

### DIFF
--- a/mooncake-transfer-engine/tent/include/tent/transport/rdma/endpoint.h
+++ b/mooncake-transfer-engine/tent/include/tent/transport/rdma/endpoint.h
@@ -23,6 +23,7 @@
 #include <vector>
 
 #include "context.h"
+#include "tent/common/concurrent/ticket_lock.h"
 
 namespace mooncake {
 namespace tent {
@@ -32,6 +33,8 @@ class RdmaEndPoint : public std::enable_shared_from_this<RdmaEndPoint> {
         uint64_t padding[7];
     };
 
+    // Not synchronized: every access goes through queue_lock_list_[qp_index]
+    // (submitSlices / acknowledge) or the endpoint write lock (teardown).
     struct BoundedSliceQueue {
         size_t head, tail, capacity, count;
         std::vector<RdmaSlice*> entries;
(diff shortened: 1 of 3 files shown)


alogfans (approved):
LGTM

PR #3925, opened 2026-09-07
[TENT] Spin with pause before yielding in TicketLock

SongOf:
## Description

  `TicketLock::lock()` called `std::this_thread::yield()` on every spin, so a waiter paid a `sched_yield` syscall per iteration even when the holder was running and about to release. Wait with `PAUSE()` for up to 1000 spins first and only then yield — the same policy `RWSpinlock` already uses. Contended hand-off drops to the cost of a cross-core cache-line transfer; the yield path is kept for a preempted holder. Adds `tent_ticket_lock_test`. No callers change.

  Two-thread microbenchmark, build container (`-O2`, glibc futex mutex as reference):

  | lock+unlock | before | after |
  |---|---|---|
  | uncontended | 2.9 ns | 1.6–2.3 ns |
  | 2 threads contending, empty critical section | 123 ns/op | 40–65 ns/op |
  | (`std::mutex`, same test, unfair) | 17–39 ns/op | — |

  ## Module

  ## Type of Change

  - [x] Performance improvement

  ## How Has This Been Tested?

  **Test commands:**
  ```bash
  cmake --build build-tent --target tent_ticket_lock_test tent tent_xport_rdma tentpy -j4
  ./build-tent/mooncake-transfer-engine/tent/tests/tent_ticket_lock_test

  Test results:
  - [x] Unit tests pass (tent_ticket_lock_test: mutual exclusion under 8 threads, hand-off in arrival order, hand-off after a hold past the spin budget; 4/4, stable over 5 runs)
  - [x] Manual testing done (microbenchmark above; all tent* libraries and tentpy build clean)

  Checklist

  - [x] I have run pre-commit on the files changed in this PR and all hooks pass (cmake-format wants to reflow an unrelated pre-existing block in tests/CMakeLists.txt; left out per AGENTS.md)

  AI Assistance Disclosure

### DIFF
--- a/mooncake-transfer-engine/tent/tests/ticket_lock_test.cpp
+++ b/mooncake-transfer-engine/tent/tests/ticket_lock_test.cpp
@@ -1,0 +1,144 @@
+// Copyright 2026 KVCache.AI
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Unit tests for TicketLock: mutual exclusion under contention, hand-off in
+// arrival order, and hand-off after a hold long enough to push waiters past
+// the spin budget into the yield path.
+
+#include "tent/common/concurrent/ticket_lock.h"
+
+#include <gtest/gtest.h>
+
+#include <atomic>
+#include <chrono>
+#include <mutex>
+#include <thread>
+#include <vector>
+
+namespace mooncake {
+namespace tent {
+
+// Tickets are the lock's only observable state, and a test that wants to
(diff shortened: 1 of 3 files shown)


mooncake-transfer-engine/tent/tests/ticket_lock_test.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+        lock.lock();
+        std::thread b([&] {
+            b_arriving.store(true);
+            lock.lock();
+            order.push_back('B');
+            lock.unlock();
+        });
+        while (!b_arriving.load()) std::this_thread::yield();
+        std::this_thread::sleep_for(20ms);
```
alogfans: In rare conditions, thread b may not obtain the lock after 20ms, especially if scheduling is delayed.

SongOf: Right — the 20 ms gaps only made the order likely, not certain. Replaced them with a `TicketLockTestPeer` (same friend pattern as `RdmaEndPointTestPeer`) that reads `next_ticket_`: the test now starts C only after it sees B holding ticket 1, and releases only after C holds ticket 2, so the arrival order is a fact rather than a race. The other test waits for the waiter's ticket the same way before sleeping past the spin budget. Passes 30/30 pinned to one core with a CPU hog on the same core (`taskset -c 0`), which is the delayed-scheduling case you describe. Pushed as 9c829734.

alogfans (approved):
LGTM

PR #3929, opened 2026-09-07
[Store] Fan verified bytes out to duplicate keys in batch gets

### LINKED ISSUE #3876: [Bug]: get_batch with a duplicate key returns the first occurrence's unwritten buffer as success (stale data)

he-yufeng:
## Description

Fixes #3876: a batch read containing the same key more than once returned the first occurrence's buffer as SUCCESS without writing it. The destination slice maps are keyed by string, so duplicates collapse: every occurrence transfers into the last op's buffer, the per-index checksum validates against that one filled buffer, and the first op's own freshly allocated (never zeroed, recycled) buffer is handed back with stale bytes.

The change transfers each unique key exactly once (first occurrence as primary), then fans the verified bytes out to each duplicate's own destination locally. The dedup plan and result propagation live in one shared layer (`batch_read_fanout.h`: `PlanDuplicateKeys`, `FanOutDuplicates`), so every read path behaves identically: duplicates of a failed primary inherit the primary's error instead of a pre-set success, and a size mismatch between same-key destinations fails the duplicate loudly everywhere.

The fan-out copy is device-aware: `RuntimeAccelerator` gains `CopyAuto` (host or device on either side, memcpy fallback for plain host pairs), and `CopySlicesMaybeDevice` walks scatter-gather lists whose chunking may differ. MEMORY/LOCAL_DISK duplicates with GPU-resident caller buffers no longer go through a plain memcpy.

Covered paths, all in `real_client.cpp`:

- `batch_get_buffer_internal` memory path (the issue's repro) and its LOCAL_DISK offload path (where the overwrite order additionally made the last op win the unwritten buffer).
- `batch_get_into_internal` memory path, its DISK/DFS temp-buffer path (which also mixed up which temp handle gets read back), and its LOCAL_DISK offload path.
- `batch_get_into_multi_buffers_internal` memory / LOCAL_DISK / DISK-DFS paths.

## Module

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

- New unit suite `batch_read_fanout_test` (10 tests): dedup plan ordering/pairing, slice walker across mismatched chunking and zero-size slices, underflow rejection, and the fan-out contract (primary failure propagates, size mismatch fails loudly without copying, copy failure recorded).
- `runtime_accelerator_test` gains 4 `CopyAuto` cases (host fallback, destination-device context, source-device D2H, device failure reporting).
- Existing integration test `BatchGetDuplicateKeysFanOut` in `client_integration_test.cpp` still covers end-to-end: put a 4KB pattern, batch-read the same key twice with the first destination pre-poisoned with 0xAA, assert both buffers return the payload.

**Test commands:**
```bash
cmake --build build --target batch_read_fanout_test runtime_accelerator_test client_integration_test
./mooncake-store/tests/batch_read_fanout_test
./mooncake-store/tests/runtime_accelerator_test
./mooncake-store/tests/client_integration_test --gtest_filter="*Duplicate*:*BatchGet*"
```

**Test results:**
- [x] Unit tests pass (10/10 fanout, 8/8 accelerator)
- [x] Integration tests pass (duplicate + batch-get cases, in a Linux container on Ubuntu 22.04)

## Checklist

- [x] I have formatted my code (pre-commit run on every touched file; all hooks pass)
- [x] I have updated the documentation (not applicable; no user-facing behavior change beyond returning correct bytes)
- [ ] For changes >500 LOC: I have filed an RFC issue (N/A: bug fix; the shared fan-out layer is a direct response to review feedback on this PR)

## AI Assistance Disclosure

- [ ] No AI tools were used

Prepared with AI assistance (Kimi Code): the duplicate-collapse mechanism was verified by reading `real_client.cpp` / `client_service.cpp` on current main, the fix, refactor and tests were drafted by the agent, and I reviewed the complete diff before submitting. Verification was actually run as described above (container build + unit/integration runs on this head).


Aionw:
The duplicate-key handling repeats the same bookkeeping, primary-result checks, size validation, copying, and result propagation across several read paths. These implementations already differ: some check sizes while others do not, and MEMORY/LOCAL_DISK use plain `memcpy` even though caller buffers may reside on a GPU.

Could we handle deduplication and result propagation once at the batch-operation level, and share a device-aware copy helper for contiguous and scatter-gather buffers? That would keep replica-specific paths focused on reading each unique key and make the duplicate-key behavior consistent across storage types.

he-yufeng:
@Aionw Reworked along exactly those lines on head 084c8d62.

Dedup and result propagation now live in one shared layer, `mooncake-store/include/batch_read_fanout.h`:

- `PlanDuplicateKeys(ops)` decides who transfers (unique keys, first occurrence as primary) and who waits for a fan-out copy. All three BatchGet call sites build their batch arrays from the plan's primaries, so the "slice map is keyed by string" bookkeeping exists exactly once.
- `FanOutDuplicates(jobs, results)` owns the whole propagation contract: a duplicate of a failed primary inherits the primary's error, a size mismatch between same-key destinations fails the duplicate loudly, and the copy's own error is recorded as-is. Every path now does the size check, not just some.

The copy is device-aware end to end:

- `RuntimeAccelerator` gains `CopyAuto(dst, src, size)`: host or device on either side, run in the destination device's context, memcpy fallback for plain host pairs.
- `CopySlicesMaybeDevice(dst_slices, src_slices, total_size, ctx)` walks scatter-gather lists whose chunking may differ, so the MEMORY/LOCAL_DISK duplicate paths no longer plain-memcpy buffers that may be GPU-resident, and multi-buffer destinations with different per-caller carving copy by byte position.

What stays per-path is only the part that is genuinely replica-specific: where the primary's verified bytes sit (own slices, a temp BufferHandle, or the offload RPC result), expressed as one copy lambda per duplicate job.

Tests: new `batch_read_fanout_test` unit suite (10 tests: plan ordering, slice walker across mismatched chunking / zero-size slices / underflow, and each fan-out contract arm), 4 new `CopyAuto` cases in `runtime_accelerator_test`, and the existing `BatchGetDuplicateKeysFanOut` integration test still passes end-to-end on this head. Pre-commit clean on all touched files.

he-yufeng:
Note on the red leg: `build-wheel-rocm (3.10)` failed after 56 minutes with no failed step and no retrievable job log, while the same run's rocm 3.12 build passed and my change touches only arch-neutral C++ (runtime_accelerator.cpp has no HIP-dependent code). This matches the runner-death signature the ROCm lane has shown before (exit 137 on a previous PR). A re-run of that one job should clear it; flag if it recurs and I will dig.

ykwd:
Reran the failed job

Aionw (approved):
LGTM

ykwd (approved):

PR #3883, opened 2026-09-05
[TENT] Disable Nagle on HP TCP client sockets

jacklin78911-collab:
## Description

Small HP TCP WRITEs can stall between the separately sent request header and payload. Enable `TCP_NODELAY` after the client socket connects successfully, before sending the first header. Socket-option failures follow the existing I/O error path. The shared callback covers both source-bound and ordinary connections.

This is a seven-line fix in `hp_tcp_client.cpp`. It preserves the wire format, completion semantics, buffer ownership and lane scheduling.

Related-work check: [TCP_NODELAY search](https://github.com/kvcache-ai/Mooncake/issues?q=TCP_NODELAY). #2314 applied socket tuning to the classic Transfer Engine TCP implementation; this PR covers the separate TENT HP TCP client introduced in #3689 and tested at the #3861 merge commit.

## Module

- [x] Transfer Engine (`mooncake-transfer-engine`)
- [ ] Mooncake Store (`mooncake-store`)
- [ ] Reshard (`mooncake-reshard`)
- [ ] Mooncake EP (`mooncake-ep`)
- [ ] Mooncake PG (`mooncake-pg`)
- [ ] Integration (`mooncake-integration`)
- [ ] P2P Store (`mooncake-p2p-store`)
- [ ] Python Wheel (`mooncake-wheel`)
- [ ] Common (`mooncake-common`)
- [ ] Mooncake RL (`mooncake-rl`)
- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [x] Performance improvement
- [ ] Other

## How Has This Been Tested?

Release Transfer Engine build with `USE_TENT=ON`, `USE_CUDA=OFF` passed. Manual tests used the compiled Python binding and registered CPU buffers on two x86 hosts, forced TCP over one Ethernet interface, four workers/four lanes and eight pinned CPU cores per process. The base was `2b3cefbac4d571bcb001db3aa702772b41b8c647`; the intervention was only this socket option.

For a 4-byte WRITE, each entry below is the p50 from one independent process run with three warmup requests and 40 measured requests:

| Version | Three independent p50 values (ms) |
| --- | --- |
| Base HP TCP | 44.017 / 43.981 / 43.996 |
| With this change | 0.174 / 0.135 / 0.142 |

[...]

PR #3931, opened 2026-09-08
[TENT] Fix HP TCP coalescing limits and short READ stalls

jacklin78911-collab:
## Description

Follow-up to #3834, #3861 and #3883. This fixes the existing `hp_tcp`; it does not introduce a transport, GPU path or routing/scheduling mechanism as proposed in the older #1633 work.

- Keep runtime coalescing within both local and advertised remote HP TCP transfer-size limits, including unequal limits. Adjacent individually valid requests could otherwise merge into an oversized request and fail. Reuse existing boundary resolution; single-request submissions incur no extra metadata decoding.
- Complete #3883 on the server side: set TCP_NODELAY on accepted HP TCP sockets. #3883 already covers client sockets and small WRITEs. Separate READ response-header/payload writes could otherwise incur delayed-ACK stalls; three two-host repeats reduced 4 KiB READ p99 from about 48 ms to 94–99 µs.
- Observe actual source/destination addresses and completed slice ranges in test-local relays, with payload/guard/reuse checks and stalled/healthy peers sharing one client worker. Run all three fork-based E2E cases separately; use unequal endpoint limits to cover coalescing.
- Strengthen HP TCP CPU data only with `--check_consistency=true`, so reordered slices cannot pass a constant-byte check. Ordinary mix, other backends and write_seed/read_verify retain their existing patterns; a focused benchmark regression checks the gate. Shorten existing usage documentation.

Wire format, WRITE ACK/single-lane semantics, ownership, admission and bounded retry remain unchanged. No failover, WRITE slicing, new policy or public configuration is added.

## Module

- [x] Docs

## Type of Change

- [x] Bug fix
- [x] Documentation update

## How Has This Been Tested?

```bash
ctest --test-dir build \
  -R '^(request_merge_test|transfer_engine_config_override_test|tent_hp_tcp_transport_test|tent_hp_tcp_e2e_.*|tent_tcp_datapath_roundtrip_test_.*|tent_transport_selector_test|tebench_target_metrics_test)$' \
  --output-on-failure
```

- [x] Final-revision Release: 10/10 targeted CTest entries, 88 GTest cases with no skips, including unequal-limit E2E, single-worker peer isolation, merge, normal TCP, selector and benchmark pattern gating. `tebench` also builds.
- [x] Final-revision ASan+UBSan: 6/6 targeted CTest entries, 20 GTest cases with no skips; leak detection and halt-on-error enabled. No sanitizer errors. TSan and two-host performance were not rerun during the second-pass correction.
- [x] Prior revision `4682e076`: Release 16/16; HP TCP/request-merge ASan+UBSan (including leak detection) 10/10; TSan 10/10. These previous full gates are not claimed as rerun on the final revision.
- [x] Two-host `tebench --op_type=mix --check_consistency=true --xport_type=hp_tcp`: non-constant CPU data, 4 KiB and 64 MiB, full comparison passed.
- [x] Three interleaved two-host performance repeats and same-engine/same-peer mixed READ tests. With four workers/four total lanes, single 64 MiB READ improved from 3.22 to 6.46 GB/s with two rails. Four concurrent READs showed no additional gain (11.13 vs 10.96 GB/s). Both interfaces carried payload-direction traffic; this is not a claim of universal or purely physical-NIC scaling.
- [x] `git diff --check`, clang-format 20 / LLVM changed-line check, and documentation HTML build. Other applicable pre-commit hooks passed; the Windows C++ shell entry was replaced by the equivalent direct LLVM check.

## Checklist

- [x] Human submitter has reviewed every changed line and the measured limitations.
- [x] Documentation and targeted regression coverage updated.
- [x] No protocol or public API change; no new production abstraction.

## AI Assistance Disclosure

- [x] AI tools were used

### DIFF
--- a/docs/source/design/tent/hp-tcp.md
+++ b/docs/source/design/tent/hp-tcp.md
@@ -15,6 +15,11 @@ never moves between workers, and operations on a lane are FIFO. ASIO provides
 the event queue; process-wide task and byte admission limits bound all accepted
 work, including callbacks waiting in that queue.
 
+Both ends disable Nagle's algorithm on these persistent sockets. Request and
+response headers are sent separately from their payloads; waiting for a
+delayed ACK between those writes can otherwise add tens of milliseconds to
+small transfers, even without queue pressure.
+
 The server uses the same worker pool. Accepted sockets are assigned to workers
 and stored in worker-owned session sets. A global connection limit bounds live
 sessions; closing a session removes it immediately rather than retaining one
@@ -42,6 +47,17 @@ persistent lane. A READ large enough to contribute at least one internal I/O
 step per configured rail is split into one contiguous slice per rail. The
 slices reuse the persistent lanes and complete as one TENT task. `hp_tcp` does
 not rebalance traffic or fail over between rails.
+
+The current internal step is 1 MiB: with two rails, READs of at least 2 MiB
+are sliced; smaller READs and single-rail READs are not. Remainder bytes are
+assigned to the first slices, so uneven lengths still cover the request
+exactly. Multiple independent WRITEs can use different rails, but each WRITE
+stays on one lane and waits for its own remote completion ACK.
+
+`connections_per_peer` is the **total** lane budget, not a per-rail multiplier.
+For example, four lanes and two rails give two lanes per rail. Worker count is
+independent of rail count: several lanes may share an owner worker. Increasing
+lanes on a single rail does not split a single READ into multiple streams.
 
 ## Protocol and memory safety
 
@@ -76,6 +92,13 @@ deadline resumes as soon as the next header begins. A timeout cancels the
 resolver or socket; terminal completion is published only after the
 corresponding callback retires.
 
+A failed READ slice cancels its siblings; the logical task settles only after
+all slice callbacks retire. A stale-registration result can trigger the
+existing bounded metadata refresh/retry, not migration onto another rail.
+Independent peers can continue while a peer is waiting for its progress
+timeout. FIFO sharing within one peer's lanes can still delay small requests
+behind large ones; slicing is not a priority or preemption mechanism.
+
 Shutdown closes admission and the listener, drains queued dispatch callbacks,
 cancels every client lane and server session on its owner, waits for operations
 and leases, then stops and joins worker threads. This makes shutdown bounded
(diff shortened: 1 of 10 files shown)


mooncake-transfer-engine/tent/tests/hp_tcp_transport_test.cpp
```diff
@@ -241,11 +242,105 @@ Status WaitForTransportResult(HighPerformanceTcpTransport& transport,
...
+
+    asio::io_context io_;
+    asio::ip::tcp::acceptor acceptor_;
+    asio::ip::tcp::endpoint backend_;
+    std::exception_ptr error_;
+};
+
+// As in the socket tests, the kernel orders the remote WRITE before its ACK,
+// but that ordering is not a C++ happens-before edge visible to TSan.
+__attribute__((no_sanitize("thread"))) bool SocketOrderedEqual(
```
alogfans: `__attribute__((no_sanitize("thread")))`: Is this necessary?

alogfans (approved):
LGTM. But before approving, please merge the main branch and make sure CI check is all-green.

docs/source/design/tent/hp-tcp.md
```diff
@@ -106,10 +127,87 @@ The transport is configured under `transports.hp_tcp`:
...
+
+On two Xeon 8457C virtual machines, with four workers and four total lanes,
+three interleaved runs (1-second warmup, 3-second measurement) gave median
+single-task 64 MiB READ throughput of 3.22 GB/s on one rail and 6.46 GB/s on
+two. Four concurrent READs showed no additional gain (11.13 vs 10.96 GB/s).
+Small 4 KiB READs averaged 72 vs 78 microseconds. Both interfaces carried
+payload-direction traffic, but their underlying resource independence is not
+guaranteed. A same-pool 4 KiB/64 MiB closed-loop mix still delayed small tasks
+behind large ones: static slicing offers neither latency isolation nor
+universal bandwidth scaling.
```
staryxchen, line 219: Presenting the test results in a table would make them easier to understand.

mooncake-transfer-engine/tent/tests/hp_tcp_transport_test.cpp
jacklin78911-collab: I initially followed the existing socket-test pattern because TSan may not recognize the ordering provided by the TCP ACK. I overlooked that we could reuse the existing worker barrier to make the synchronization explicit. Thanks for questioning this—I’ve removed the suppression, and all 7 transport tests pass under TSan.

docs/source/design/tent/hp-tcp.md
jacklin78911-collab: OK. Thanks for your suggestion and I'll take it.

staryxchen (approved):
LGTM

PR #3932, opened 2026-09-08
[Store] Add optional RocksDB storage backend

982945902:
## Description

Adds an opt-in RocksDB-backed `LOCAL_DISK` storage backend as an alternative implementation for the design space discussed in #3856.

The backend uses RocksDB transactions and BlobDB storage to provide:

- durable prepare payload writes followed by callback-controlled commit or rollback;
- atomic logical-key to committed-version publication;
- overwrite cleanup while preserving the previous value when publication is rejected;
- batched reads through `MultiGet`;
- restart recovery, `ScanMeta`, `IsExist`, and `RemoveAll`;
- an isolated on-disk directory and backend descriptor: `rocksdb_storage_backend`.

This PR does **not** implement the custom Segment/WAL/checkpoint format proposed by #3856. Instead, it evaluates whether Mooncake should reuse a mature storage engine for the same holder-local physical-placement boundary. Master metadata remains unchanged and stores only the logical `LOCAL_DISK` replica location.

RocksDB support is disabled by default through `MOONCAKE_USE_ROCKSDB=OFF`, so default builds do not acquire a RocksDB dependency. Blob files are enabled by default. Direct blob writes remain disabled by default for conservative durability semantics and can be enabled explicitly for controlled benchmarking.

This is not a DFS implementation and does not address #3641.

### Local-SSD baseline and durability correction

The previous four-backend table was not an apples-to-apples write comparison. The current Bucket POSIX write path explicitly syncs neither its data file nor its metadata file, while the prior Custom Log run used `MOONCAKE_LOG_SYNC_POLICY=record` and RocksDB used `MOONCAKE_ROCKSDB_SYNC_WRITES=true`. The earlier numbers are therefore not used as the primary performance evidence below.

The write benchmark was rerun as a fully interleaved five-round comparison with the three relevant implementations using no explicit sync:

- Bucket used its current buffered POSIX write path;
- Custom Log used `MOONCAKE_LOG_SYNC_POLICY=none`, compaction disabled, and payload write parallelism `4/16/32` for `4 KiB/128 KiB/4 MiB` values;
- RocksDB 11.8.1 used WAL enabled, `MOONCAKE_ROCKSDB_SYNC_WRITES=false`, BlobDB, and opt-in direct blob writes;
- all backends used the same Release binary, buffered I/O, ext4 NVMe-backed volume, workload shape, and readback verification;
- all 45 runs completed with zero operation or verification errors;
- the shared 384-CPU host was heavily loaded (`load average 303-313`), so five-run medians and full ranges are reported.

| Scenario | Bucket | Custom Log | RocksDB |
| --- | ---: | ---: | ---: |
| Write 4 KiB × 32 | 342.54 MB/s | 182.20 MB/s | **386.19 MB/s** |
| Write 128 KiB × 32 | **2971.79 MB/s** | 1148.90 MB/s | 1619.35 MB/s |
| Write 4 MiB × 8 | **4505.00 MB/s** | 1261.87 MB/s | 851.40 MB/s |

| Scenario | Bucket p99 | Custom Log p99 | RocksDB p99 |
| --- | ---: | ---: | ---: |
| Write 4 KiB × 32 | 1.68 ms | 1.43 ms | **1.34 ms** |
| Write 128 KiB × 32 | **2.93 ms** | 8.75 ms | 21.21 ms |
| Write 4 MiB × 8 | **12.83 ms** | 39.25 ms | 50.49 ms |

Five-run throughput ranges:

- Bucket: 4 KiB `312.31-372.98`, 128 KiB `2763.63-3219.74`, 4 MiB `4018.41-4838.22 MB/s`.
- Custom Log: 4 KiB `175.70-190.40`, 128 KiB `1047.78-1256.52`, 4 MiB `1192.39-1325.61 MB/s`.
- RocksDB: 4 KiB `384.16-459.79`, 128 KiB `1016.91-1656.54`, 4 MiB `804.08-873.32 MB/s`.

This correction materially changes the interpretation:

- RocksDB is the fastest implementation for the 4 KiB workload when sync is disabled for all compared backends.
- Bucket remains substantially faster for 128 KiB and 4 MiB values because one `BatchOffload` becomes a packed immutable file with minimal indexing and no WAL/LSM maintenance.
- Custom Log improves substantially without per-record durability work, but it still does not beat Bucket in this short append-only workload.
- The earlier large Bucket advantage over RocksDB at 4 KiB was primarily a durability-configuration artifact.

The hot-read measurements from the preceding run are unaffected by the write-sync configuration and remain useful as a read-path reference: Bucket `6802.20 MB/s`, Custom Log `5380.42 MB/s`, and RocksDB `2187.27 MB/s` for 128 KiB × 32 hot reads. OffsetAllocator was excluded from the corrected no-sync write table because its existing persistence modes are not equivalent to the simple no-sync switch used by Custom Log and RocksDB.

These are buffered/page-cache microbenchmarks, not physical-media durability or SSD-endurance measurements. They do not cover power-loss recovery, steady-state compaction/Blob GC, capacity turnover, or host/device write amplification. A separate durable comparison requires a common contract—ideally one durable commit per `BatchOffload`—and a corrected Bucket data/metadata publication protocol.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [x] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
# Default dependency-free configuration
cmake ... -DMOONCAKE_USE_ROCKSDB=OFF
cmake --build ... --target mooncake_store storage_backend_bench

# RocksDB 11.8.1 configuration
cmake ... -DMOONCAKE_USE_ROCKSDB=ON \
  -DROCKSDB_INCLUDE_DIR=<rocksdb>/include \
  -DROCKSDB_LIBRARY=<rocksdb>/librocksdb.so
cmake --build ... --target rocksdb_backend_test storage_backend_bench
ctest --test-dir ... --output-on-failure -R '^rocksdb_backend_test$'

uvx pre-commit run --files <changed-files>
```

**Test results:**
- [x] Default `MOONCAKE_USE_ROCKSDB=OFF` build passes.
- [x] RocksDB-enabled build passes with RocksDB 11.8.1 and GCC 11.4.
- [x] `rocksdb_backend_test` passes, covering batch write/read, restart plus `ScanMeta`, `RemoveAll`, factory creation, and rejected-overwrite rollback.
- [x] `storage_backend_bench` builds in both configurations.
- [x] Three-run A/B benchmark matrix completed.
- [ ] Integration tests pass (not run against a dedicated physical SSD host).

## Checklist

- [x] I have updated the documentation (the implementation choice and tradeoffs are documented here and in RFC #3856 discussion)
- [x] For changes >500 LOC: I have filed an RFC issue (#3856)

## AI Assistance Disclosure

- [ ] No AI tools were used
- [x] AI tools were used (OpenAI Codex assisted with implementation, tests, review, and benchmark scripting; the submitter reviewed the resulting changes and evidence.)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,6 +30,7 @@ option(
   OFF)
 option(WITH_EP "build mooncake with expert parallelism support" OFF)
 option(USE_NOF "build mooncake store with NoF SSD pool support" OFF)
+option(MOONCAKE_USE_ROCKSDB "build Mooncake Store with RocksDB backend" OFF)
 option(MOONCAKE_ENABLE_TEST_FAILPOINTS
        "Enable file-handshake failpoints for integration tests" OFF)
 if(MOONCAKE_ENABLE_TEST_FAILPOINTS)
--- a/mooncake-store/benchmarks/storage_backend_bench.cpp
+++ b/mooncake-store/benchmarks/storage_backend_bench.cpp
@@ -92,6 +92,9 @@
 
 #include "gflags/gflags.h"
 #include "glog/logging.h"
+#ifdef MOONCAKE_USE_ROCKSDB
+#include "storage/local/rocksdb/rocksdb_backend.h"
+#endif
 #include "storage_backend.h"
 
 namespace fs = std::filesystem;
@@ -102,7 +105,8 @@ namespace fs = std::filesystem;
 
 // === Core Parameters ===
 DEFINE_string(backend, "offset_allocator",
-              "Backend type: offset_allocator, bucket, file_per_key, or all");
+              "Backend type: offset_allocator, bucket, file_per_key, rocksdb, "
+              "or all");
 DEFINE_uint64(value_size, 128 * 1024, "Value size in bytes (default: 128KB)");
 DEFINE_uint64(batch_size, 32, "Batch size for operations (default: 32)");
 DEFINE_uint64(num_operations, 1000,
@@ -242,7 +246,14 @@ AccessPattern StringToAccessPattern(const std::string& str) {
 // Backend Types
 // ============================================================================
 
-enum class BackendType { OFFSET_ALLOCATOR, BUCKET, FILE_PER_KEY };
+enum class BackendType {
+    OFFSET_ALLOCATOR,
+    BUCKET,
+    FILE_PER_KEY,
+#ifdef MOONCAKE_USE_ROCKSDB
+    ROCKSDB
+#endif
+};
 
 std::string BackendTypeToString(BackendType type) {
     switch (type) {
@@ -252,6 +263,10 @@ std::string BackendTypeToString(BackendType type) {
             return "bucket";
         case BackendType::FILE_PER_KEY:
             return "file_per_key";
+#ifdef MOONCAKE_USE_ROCKSDB
+        case BackendType::ROCKSDB:
+            return "rocksdb";
+#endif
     }
     return "unknown";
 }
@@ -260,6 +275,9 @@ BackendType StringToBackendType(const std::string& str) {
     if (str == "offset_allocator") return BackendType::OFFSET_ALLOCATOR;
     if (str == "bucket") return BackendType::BUCKET;
     if (str == "file_per_key") return BackendType::FILE_PER_KEY;
+#ifdef MOONCAKE_USE_ROCKSDB
+    if (str == "rocksdb") return BackendType::ROCKSDB;
+#endif
     LOG(FATAL) << "Unknown backend type: " << str;
     return BackendType::OFFSET_ALLOCATOR;
 }
@@ -831,6 +849,13 @@ std::shared_ptr<mooncake::StorageBackendInterface> CreateBackend(
             return std::make_shared<mooncake::StorageBackendAdaptor>(
                 config, fpk_config);
         }
+#ifdef MOONCAKE_USE_ROCKSDB
+        case BackendType::ROCKSDB: {
+            config.storage_backend_type =
+                mooncake::StorageBackendType::kRocksDb;
+            return std::make_shared<mooncake::RocksDBStorageBackend>(config);
+        }
+#endif
     }
     return nullptr;
 }
@@ -2282,7 +2307,12 @@ struct BenchmarkResult {
 void RunAllBenchmarks(const std::string& storage_path, size_t capacity) {
     std::vector<BackendType> backends = {BackendType::OFFSET_ALLOCATOR,
                                          BackendType::BUCKET,
-                                         BackendType::FILE_PER_KEY};
+                                         BackendType::FILE_PER_KEY
+#ifdef MOONCAKE_USE_ROCKSDB
+                                         ,
+                                         BackendType::ROCKSDB
+#endif
+    };
 
     std::vector<BenchmarkResult> results;
 
--- a/mooncake-store/include/storage/local/rocksdb/rocksdb_backend.h
+++ b/mooncake-store/include/storage/local/rocksdb/rocksdb_backend.h
@@ -1,0 +1,75 @@
+#pragma once
+
+#include <array>
+#include <atomic>
+#include <functional>
+#include <memory>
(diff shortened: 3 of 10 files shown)


ykwd:
Thanks for the PR and the RFC. I agree that the existing local-disk backends are not perfect and there is definitely room for improvement.

That said, I have some concerns about whether a log-structured design with background compaction is the right tradeoff for our common workloads. In typical use cases such as KV cache and RL, the workload is usually write-heavy and read-light, while still being quite sensitive to read latency. The write amplification introduced by compaction could therefore become a significant cost, especially when storage bandwidth and SSD endurance are important.

For the Bucket backend, while it is indeed not ideal for workloads with frequent in-place updates, I am not sure frequent updates are actually common in our major use cases. For KV cache, the value associated with a given key is generally immutable once written. For RL workloads, a more common pattern may be to delete a whole batch of old values at the end of each iteration and then insert a new batch, rather than repeatedly updating the same keys.

Because of this, I think it may be better to optimize around concrete workloads and application patterns rather than introducing a more general storage design with additional complexity and write amplification. In particular, I would prefer that we first explore whether the existing backends can be improved for these target workloads before adding a completely new storage format and compaction subsystem.

982945902:
Correction to the earlier benchmark comparison: after checking the current Bucket write path, I found that its normal POSIX path explicitly syncs neither the bucket data file nor the metadata file. The previous comparison therefore used different durability settings (`Bucket` buffered writes versus Custom Log record sync and RocksDB `sync=true`).

I reran a fully interleaved five-round no-sync comparison using the same Release binary and buffered ext4 volume:

| Scenario | Bucket | Custom Log (`sync_policy=none`) | RocksDB (`sync=false`, WAL enabled) |
| --- | ---: | ---: | ---: |
| 4 KiB × 32 | 342.54 MB/s | 182.20 MB/s | **386.19 MB/s** |
| 128 KiB × 32 | **2971.79 MB/s** | 1148.90 MB/s | 1619.35 MB/s |
| 4 MiB × 8 | **4505.00 MB/s** | 1261.87 MB/s | 851.40 MB/s |

All 45 runs completed with zero operation or verification errors. The PR description now contains p99 values, five-run ranges, the exact settings, and the corrected interpretation.

The result is more nuanced: RocksDB leads the 4 KiB case, while Bucket remains substantially faster for larger immutable batches. These buffered tests still do not answer power-loss durability, steady-state GC/compaction, capacity turnover, or SSD write-amplification questions; those require a separate workload and a common durable-commit contract.

982945902:
Thanks for the detailed feedback. After adding the existing Bucket backend as the baseline and correcting the durability mismatch in the initial benchmark, I agree with your conclusion.

The current Bucket POSIX path explicitly syncs neither the data file nor the metadata file. Once Custom Log and RocksDB are tested with equivalent no-sync settings, RocksDB only leads the 4 KiB case, while Bucket remains substantially faster for the 128 KiB and 4 MiB immutable-batch workloads. RocksDB also retains additional indexing, WAL/LSM, Blob GC, and compaction complexity, without a demonstrated benefit for the dominant large immutable-object workload.

The custom log design has the same fundamental issue: it improves update and space-reclamation generality, but introduces a new WAL/checkpoint/compaction subsystem and write amplification before we have established that frequent in-place updates are an important production workload.

Given the workload characteristics you described, I agree that this direction is not sufficiently justified. A better next step would be to improve the existing Bucket backend around explicit batch/epoch lifecycle, whole-Bucket retirement, and a clearly defined durability contract, then benchmark that against representative KV-cache and RL lifecycle traces.

I am closing this PR, the custom implementation PR #3869, and RFC #3856. Thank you for steering the discussion toward the actual workload requirements.

ykwd:
Thanks a lot for the thoughtful work on this, and especially for taking the time to revisit the benchmark setup and rerun the comparison with consistent durability settings.

PR #3934, opened 2026-09-08
[TENT] Pin failover policy at transfer submission

jacklin78911-collab:
## Description

Focused follow-up to the configuration lifecycle foundation merged in #3836. Supersedes the closed #3857, now based directly on `main` with only this change (five files).

- Capture the runtime generation, failover-attempt limit and auto-failover-on-poll flag once at transfer submission.
- Copy that policy into direct and runtime-queued tasks, including task copies/moves, and retain it across cross-transport retries.
- Use the pinned policy during status polling while preserving explicit/background progress semantics and the current HP TCP metadata-refresh/permanent-error handling.
- Use shared failover defaults and parse the two policy values when building the snapshot, keeping configuration lookup off the submission hot path.
- Add two regression cases that change each policy field independently: an in-flight transfer retains the old policy, while a new transfer observes the published policy.

The initial snapshot is installed before fallible setup steps. Publication is exposed only through a test hook, which rejects null snapshots without replacing the active policy. This does **not** add a production live-update API, configuration provider, rollback mechanism, or propagation of a parent's generation into staging children.

Refs #3833.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

Ubuntu 24.04 (WSL), GCC 13.3, RelWithDebInfo, CUDA disabled. The engine regressions use fake transports and local P2P metadata; no GPU or two-host performance result is claimed for this revision.

**Test commands:**
```bash
cmake --build build/failover --target config_lifecycle_test tent_engine_failover_e2e_test tent_runtime_queue_dispatch_test tent_progress_worker_test -j4
ctest --test-dir build/failover --parallel 1 --output-on-failure -R '^(config_lifecycle_test|tent_engine_failover_e2e_test|tent_runtime_queue_dispatch_test|tent_progress_worker_test)$'
pre-commit run --files mooncake-transfer-engine/tent/include/tent/common/config_lifecycle.h mooncake-transfer-engine/tent/include/tent/runtime/transfer_engine_impl.h mooncake-transfer-engine/tent/src/common/config_lifecycle.cpp mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp mooncake-transfer-engine/tent/tests/engine_failover_e2e_test.cpp
```

**Test results:**
- [x] Unit tests pass: configuration lifecycle, 15 tests.
- [x] Integration tests pass: engine failover, 30; progress worker, 23; runtime queue dispatch, 12.
- [ ] Manual hardware testing done.

All four test executables passed (80 tests total), including both pinned-policy cases and their null-publication rejection checks. All applicable PR-scoped pre-commit hooks passed, including changed-line formatting with clang-format 20.

## Checklist

- [x] I have updated the documentation (not applicable: no new user-facing API)

## AI Assistance Disclosure

- [ ] No AI tools were used

### DIFF
--- a/mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
+++ b/mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
@@ -338,6 +338,9 @@ Status TransferEngineImpl::setupLocalSegment() {
 
 Status TransferEngineImpl::construct() {
     CHECK_STATUS(ParseHpTcpTransportConfig(*conf_, &hp_tcp_transport_config_));
+    std::atomic_store_explicit(&runtime_config_snapshot_,
+                               buildTentConfigBundle(*conf_).runtime,
+                               std::memory_order_relaxed);
     auto metadata_type = conf_->get("metadata_type", "p2p");
     auto metadata_servers = conf_->get("metadata_servers", "");
 
@@ -358,9 +361,6 @@ Status TransferEngineImpl::construct() {
     CHECK_STATUS(getRpcServerThreadsFromConfig(*conf_, rpc_threads_default,
                                                rpc_server_threads));
     merge_requests_ = conf_->get("merge_requests", true);
-    max_failover_attempts_ = conf_->get("max_failover_attempts", 3);
-    enable_auto_failover_on_poll_ =
-        conf_->get("enable_auto_failover_on_poll", true);
     enable_progress_worker_ = conf_->get("enable_progress_worker", false);
     runtime_queue_config_.enabled = conf_->get("enable_runtime_queue", false);
     if (runtime_queue_config_.enabled) enable_progress_worker_ = true;
@@ -1502,6 +1502,7 @@ struct TransferEngineImpl::PreparedSubmit {
     };
 
     std::chrono::steady_clock::time_point submit_time{};
+    LogicalTransferRuntimePolicy runtime_policy;
     std::vector<Task> tasks;
     std::vector<Owner> owners;
 };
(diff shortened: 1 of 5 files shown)


mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
```diff
@@ -1804,6 +1805,13 @@ Status TransferEngineImpl::prepareSubmit(
     }
 
     prepared = PreparedSubmit{};
+    auto runtime_config = std::atomic_load_explicit(&runtime_config_snapshot_,
+                                                    std::memory_order_acquire);
+    prepared.runtime_policy.config_generation = runtime_config->generation;
```
staryxchen, line 1811: runtime_config_snapshot_ is null until the second statement of construct(). If ParseHpTcpTransportConfig fails first, the snapshot is never stored, but submitTransfer still dereferences it in prepareSubmit. That is a crash regression: the old scalars were in-class initialized and could not be null here. publishRuntimeConfigForTest(nullptr) has the same hole.

Initialize the member to a non-null default snapshot (or fill it before construct()), so this load is always defined.

staryxchen (commented):
3 / true now live in three places: RuntimeConfigSnapshot, LogicalTransferRuntimePolicy, and the get() fallbacks in buildTentConfigBundle. A later default change has to touch all three, and a TaskInfo that misses the policy copy silently falls back to 3/true instead of the user’s configured value.

Pull them into named constants (or parse once in buildTentConfigBundle and copy from the snapshot) so the defaults cannot drift.

mooncake-transfer-engine/tent/src/runtime/transfer_engine_impl.cpp
jacklin78911-collab: Thanks for the review. Both points are addressed

staryxchen (approved):
LGTM

PR #3935, opened 2026-09-08
[TransferEngine] Add Intel XPU (Level Zero) support

siju-samuel:
## Description

Adds Intel XPU (oneAPI Level Zero) support to the Transfer Engine, so XPU device
memory can be registered and moved over the existing RDMA (dma_buf) and TCP
transports. XPU reuses those transports through a cuda-alike shim layer rather
than adding a new transport.

Intel XPUs use Level Zero, a different stack from CUDA/HIP/MUSA. This follows the
same `gpu_vendor/` + `cuda_alike.h` seam the MUSA and MACA ports already use, so
no transport or engine code needed restructuring:

- `USE_XPU` cmake option with Level Zero auto-detection
- `gpu_vendor/xpu.h` / `xpu.cpp` — cuda-alike surface over Level Zero
  (ZeRuntime singleton, device alloc/free, H2D/D2H memcpy, pointer attributes,
  PCI bus id, allocation-base query, dma_buf export)
- `USE_XPU` added to the existing GPU-vendor guards in `cuda_alike.h`,
  `memory_location.cpp`, `topology.cpp` and the TCP staging path
- `rdma_context.cpp` — dma_buf export for XPU device memory
- `scripts/build_xpu.sh` for conda-based XPU builds

Scope note: this is the first phase of Intel XPU support — transports and memory
registration only. The MoE EP layer is a separate, stacked PR. The full phased
design is in the linked RFC issue rather than a file in the tree.

## Module

- [x] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

Hardware: 8x Intel Arc Pro B60 (Battlemage), oneAPI 2026.1, Level Zero 1.26,
glibc 2.42, GCC 15. This host has no InfiniBand, so the TCP path is covered
end to end and the RDMA/dma_buf path is covered only at the export level (see
Caveats).

**Test commands:**
```bash
# Prerequisites: source /opt/intel/oneapi/setvars.sh, the Level Zero
# development package (level-zero-devel / libze-dev), and the usual Mooncake
# dependencies from dependencies.sh. No conda env is required; if one is
# active it is added to the CMake search paths. Level Zero is looked up in
# $LEVEL_ZERO_ROOT, then $CONDA_PREFIX, then /usr and /usr/local.
#
# Note: do NOT put a conda gtest on the include path — the project
# FetchContents googletest 1.17.0 (mooncake-common/FindGTest.cmake) and a
# second gtest causes an ABI mismatch at link time. yalantinglibs needs no
# action; FindYLT.cmake fetches it.
./scripts/build_xpu.sh build-xpu

# Unit tests
./build-xpu/mooncake-transfer-engine/tests/xpu_transport_test

# Wheel
BUILD_DIR=build-xpu ./scripts/build_wheel.sh 3.12 dist-xpu
pip install --no-deps mooncake-wheel/dist-xpu/mooncake_transfer_engine-*.whl

# CI gates
```

**Test results:**
- [x] Unit tests pass — 21/21 `xpu_transport_test` cases on real hardware
      (init, set-device, device/host alloc, pointer attributes, H2D+D2H memcpy
      roundtrip, isDeviceMemory, dma_buf export, caller-owned fd, allocation
      base/offset, multi-alloc, GPU_PREFIX, plus 8 async/stream cases). All
      GTEST_SKIP cleanly on hosts with no Intel GPU.
- [x] Integration tests pass — `engine.so` initialises with
      `protocol=tcp, metadata=P2PHANDSHAKE`; `register_memory` /
      `transfer_sync_write` return 0; topology discovery reports
      "Level Zero XPU runtime initialized".
- [x] Manual testing done — end-to-end transfer of **XPU device memory**
      between two processes over TCP: a `torch.xpu` tensor is registered with
      `register_memory()`, written with `transfer_sync_write()`, and the
      receiving process reads back the expected byte pattern from its own XPU
      tensor. Also verified `pip install mooncake-transfer-engine` then
      importing and initialising from outside the build tree.

Fixes found by that first build/run, each with the failure it produced:

| Fix | Failure without it |
|---|---|
| `build_xpu.sh`: pass `-DWITH_STORE_RUST=OFF` | configure aborts — `WITH_STORE_RUST=ON requires WITH_STORE=ON` |
| `build_xpu.sh`: only apply `-isystem /usr/include` for the conda cross-compiler | `stdlib.h: No such file or directory` on GCC 15 (breaks libstdc++ `#include_next`) |
| `build_wheel.sh`: guard the `transfer_engine_bench` copy | wheel build aborts under `set -e` when `BUILD_EXAMPLES=OFF` |
| `rdma_context.cpp`: add `USE_XPU` to the `ibv_reg_dmabuf_mr` guard | exported dma_buf fd discarded; XPU device memory silently fell back to `ibv_reg_mr` |
| `rdma_context.cpp`: resolve the allocation base instead of assuming offset 0 | a tensor at a nonzero offset in a larger block registers the wrong bytes (the normal case with PyTorch's caching allocator) |
| `xpu.cpp`: return a `dup()` and call `zeMemPutIpcHandle()` | one leaked driver fd per export, and the caller's `close()` freed a descriptor the driver still owned |

### Asynchronous copies

`cudaMemcpyAsync` is a real asynchronous copy rather than a synchronous
fallback, with the minimum stream API to use it safely
(`cudaStreamCreate`/`CreateWithFlags`/`Destroy`/`Synchronize`/`Query` and
`cudaDeviceSynchronize`). A `cudaStream_t` is an in-order asynchronous Level
Zero immediate command list plus the device it was created on.

Three findings from bringing this up, each of which would have corrupted data
silently:

| Finding | Consequence |
|---|---|
| Level Zero immediate command lists are **not** in-order by default | A small copy overtakes a preceding large copy to the same address. Fixed with `ZE_COMMAND_QUEUE_FLAG_IN_ORDER`; `StreamPreservesSubmissionOrder` fails without it. |
| `zeMemFree` does no in-flight checking, unlike `cudaFree`'s implicit device sync | Freeing a buffer with a copy in flight corrupts silently once copies are truly async. `cudaFree`/`cudaFreeHost` now drain first. |
| `zeCommandListAppendMemoryCopy` must not be called concurrently on one command list | The per-device list is shared by every thread calling `cudaMemcpy`; the multi-threaded TCP transport would have hit this. Now locked. |

The ordering test is a real detector, not a smoke test: it needs a
write-after-write to one address, asymmetric sizes (large then <=4 KiB), and
pinned host sources. Weaken any of the three and it passes against the broken
implementation — measured, and recorded in a comment so it is not "simplified"
later.

`cudaEvent_t` is deliberately not implemented (needs an event pool, no consumer
in the XPU build yet) — callers get a link error, not silent misbehaviour.

## Checklist

- [x] I have updated the documentation (if applicable) — README protocol and
      accelerator lists
- [ ] For changes >500 LOC: I have filed an RFC issue — **TODO before merge**

## Caveats / reviewer notes

- **The RDMA dma_buf path is not verified on hardware.** This host has no
  InfiniBand, so `exportDmabuf()` is unit-tested (a valid fd is produced,
  caller-owned, with the right base/offset) but `ibv_reg_dmabuf_mr` and an
  actual GPU-direct RDMA transfer are untested. Only the TCP path is covered
  end to end. Happy to run this on a Mellanox host if someone can point me at one.
- `exportDmaBufFd()` extracts the fd from the first `sizeof(int)` bytes of
  `ze_ipc_mem_handle_t.data`. That is how the xe/i915 driver lays it out today,
  but it is not guaranteed by the Level Zero spec — flagging it as the most
  driver-version-sensitive assumption here.
- `build-xpu/` added to `.gitignore` — the default build dir of the new
  `scripts/build_xpu.sh`.

## AI Assistance Disclosure

- [ ] No AI tools were used

Claude Code was used to bring up and debug the build on Intel hardware, to audit
the never-before-compiled Level Zero code (which is how the three dma_buf
defects above were found), and to draft this description. Every changed line has
been reviewed by the submitter, and every claim in the "How Has This Been
Tested" section corresponds to a command that was actually run on the hardware
described.


alogfans (approved):
LGTM.

PR #3943, opened 2026-09-08
[Store] Drain in-flight RPCs and keep pools alive across client teardown

### LINKED ISSUE #3909: [Performance]: Segmentation fault in `asio` `epoll_reactor` when frequently recreating read-only RealClient under extreme network timeout
### Describe your performance question

# Overview
In large‑scale high‑throughput read‑only Mooncake workload, when simulating poor network with extremely short RPC timeout, tearing‑down and recreating read‑only RealClient after consecutive RPC timeouts triggers SIGSEGV null‑pointer crash. The service runs stable with normal MC_RPC_TIMEOUT_MS=400.
Hypothesis: Race between in‑flight coro‑rpc TCP async coroutine requests and client close / object destruction. According to coro_rpc doc, only send_request is thread‑safe; connect / close / object destruction must not run concurrently with unfinished RPC. Short timeout amplifies the race window. Resumed coroutine access already‑freed socket descriptor_data and causes segfault.
# Deployment and Configuration
Cluster has 200 Mooncake storage nodes. Dataset is split into 200 partitions and fully persisted on local SSD without in‑memory read cache. Each storage node runs one `mooncake_master` plus one storage‑side RealClient; master only manages local RealClient. SSD offload & disk eviction enabled, promotion‑on‑hit disabled.

Read‑side host with 8 GPUs, per rank process:

1. One shared `TransferEngine`
2. 200 read‑only `RealClient` instances, one‑to‑one connect to 200 remote storage nodes, read‑only workload, `global_segment_size=0`
3. Business logic: use `batch_get_into` for batch read. After **5 consecutive rpc\_timeout failures**: invoke `tearDownAll()`, reset client object, later rebuild via `setup_internal`.

Per host scale: 8 ranks, total `200 * 8` read‑only RealClient, high‑throughput random read pattern.

## Storage node: mooncake_master launch args
``` bash
mooncake_master \
--rpc_port=8001 \
--rpc_thread_num=8 \
--rpc_enable_tcp_no_delay=true \
--metrics_port=8007 \
--default_kv_lease_ttl=5000 \
--allow_evict_soft_pinned_objects=true \
--eviction_ratio=0.45 \
--eviction_high_watermark_ratio=0.6 \
--cluster_id=flash_mooncake_cluster \
--allocation_strategy=free_ratio_first \
--quota_bytes=3298534883328 \
--enable_ha=false \
--enable_offload=true \
--offload_on_evict=true \
--offload_force_evict=false \
--enable_disk_eviction=true \
--promotion_on_hit=false
```
## Storage‑side RealClient env & launch
``` bash
MC_IB_PCI_RELAXED_ORDERING                         = 1
MOONCAKE_LOCAL_BUFFER_SIZE                         = 6442450944
MOONCAKE_OFFLOAD_BUCKET_EVICTION_POLICY            = lru
MOONCAKE_OFFLOAD_BUCKET_KEYS_LIMIT                 = 128
MOONCAKE_OFFLOAD_BUCKET_MAX_TOTAL_SIZE             = 3113851289600
MOONCAKE_OFFLOAD_BUCKET_SIZE_LIMIT_BYTES           = 268435456
MOONCAKE_OFFLOAD_CLIENT_BUFFER_GC_INTERVAL_SECONDS = 2
MOONCAKE_OFFLOAD_FILE_STORAGE_PATH                 = /ssd_file/mooncake_cache/
MOONCAKE_OFFLOAD_HEARTBEAT_INTERVAL_SECONDS        = 1
MOONCAKE_OFFLOAD_LOCAL_BUFFER_SIZE_BYTES           = 21474836480
MOONCAKE_OFFLOAD_STORAGE_BACKEND_DESCRIPTOR        = bucket_storage_backend
MOONCAKE_OFFLOAD_TOTAL_SIZE_LIMIT_BYTES            = 3221225472000
MOONCAKE_OFFLOAD_USE_URING                         = 1
mooncake_client \
--master_server_address=<local_node_ip>:8001 \
--host=<local_node_ip> \
--protocol=rdma \
--port=8005 \
--global_segment_size=35GB \
--enable_offload=true \
--threads=8 \
--metadata_server=P2PHANDSHAKE
```
## Read‑only RealClient env
``` bash
MC_STORE_CLIENT_MIN_PORT=12000
MC_STORE_CLIENT_MAX_PORT=30000
MC_STORE_CLIENT_SETUP_RETRIES=2
MC_RPC_TIMEOUT_MS=20 # Simulate extremely poor network
mooncake_client \
--master_server_address=<master_ip>:8001 \
--host=<master_ip> \
--protocol=rdma \
--port=8005 \
--global_segment_size=0 # read‑only client, no local storage
```
## Key code snippet
``` C++
// Establish read‑only client connection
auto rc = real_client_->setup_internal(
    transport_.local_hostname(),
    transport_.metadata_server(),
    0,        // global_segment_size: read‑only client
    0,        // local_buffer_size
    "rdma",
    "",       // rdma devices: auto‑discovery
    remote_addr_,
    shared_transfer_engine,
    "",       // ipc_socket_path unused
    0,        // local rpc port unused
    false,    // enable_ssd_offload disabled
    false,    // start offload rpc server disabled
    "",       // ssd offload path
    "",       // tenant_id
    false,    // enable_client_http_server
    8003      // client_http_port
);

// Read workflow
real_client_->register_buffer(recv_buf_.get(), buf_size_);
std::vector<void*>  bufs(chunk_bufs_.begin(), chunk_bufs_.begin() + batch_sz);
std::vector<size_t> sizes(batch_sz, chunk_size_);
auto ret = real_client_->batch_get_into(batch, bufs, sizes);

// After 5 consecutive rpc_timeout: tear down and recreate
real_client_->unregister_buffer(recv_buf_.get());
real_client_->tearDownAll();
real_client_.reset();
remote_addr_.clear();
// Later call setup_internal() to rebuild client

```
# Crash Backtrace
Crash occurs in TCP coro‑rpc path `BatchGetReplicaList`. Direct cause: `descriptor_data == nullptr`, null‑pointer dereference inside asio epoll\_reactor, SIGSEGV.
``` bash 
#0  asio::detail::epoll_reactor::start_op
      (this=0x7d1083a40380, op_type=1, descriptor=8739,
       descriptor_data=@0x7d0bf2892850: 0x0,   ← nullptr dereference
       op=0x7d0c6fe1e960, is_continuation=<optimized out>)
      at /usr/local/include/asio/detail/impl/epoll_reactor.ipp:245
      descriptor_lock = {mutex_=@0x7d0ca9ee37a8, locked_=true}

#1  asio::detail::reactive_socket_service_base::async_send (…)
      at /usr/local/include/asio/detail/reactive_socket_service_base.hpp:313
      is_continuation=false; slot={handler_=0x0}
      p={h=0x7d0fdf9f8550, v=0x7d0c6fe1e960, p=0x7d0c6fe1e960}

#2  asio::basic_stream_socket<tcp,any_io_executor>::initiate_async_send::operator() (…)
      at /usr/local/include/asio/basic_stream_socket.hpp:1141
#3  asio::detail::completion_handler_async_result<…>::initiate (…)
      at /usr/local/include/asio/async_result.hpp:481
#4  asio::async_initiate<…> (…)
      at /usr/local/include/asio/async_result.hpp:895
#5  asio::basic_stream_socket<…>::async_write_some (…)
      at /usr/local/include/asio/basic_stream_socket.hpp:970
#6  asio::detail::write_op<…>::operator() (ec=<optimized out>, start=1, bytes_transferred=0, this=0x7d0fdf9f8550)
      at /usr/local/include/asio/impl/write.hpp:344
      max_size=65536
#7  asio::detail::start_write_op<…> (…)
      at /usr/local/include/asio/impl/write.hpp:460
#8  asio::detail::initiate_async_write<…>::operator() (…)
      at /usr/local/include/asio/impl/write.hpp:492
#9..#11  asio::async_result / async_initiate / async_write (…)
      at /usr/local/include/asio/async_result.hpp:481 / :895 / impl/write.hpp:568
#12 coro_io::async_write<…> lambda::operator() (socket=<optimized out>, buffer=<optimized out>)
      at /usr/local/include/ylt/coro_io/coro_io.hpp:439
#13 coro_io::async_io<…> lambda::operator() (…)
      at /usr/local/include/ylt/coro_io/coro_io.hpp:281
#14 coro_io::callback_awaitor_base<…>::await_suspend (…)
      at /usr/local/include/ylt/coro_io/coro_io.hpp:81
#15 coro_io::async_io(…Frame*) (…)
      at /usr/local/include/ylt/coro_io/coro_io.hpp:280
#16 operator() (frame_ptr=0x7d0c3f969f00)   ← coroutine resume point
      at /usr/local/include/async_simple/coro/Lazy.h:422
#17 async_simple::coro::detail::LazyBase<vector<expected<GetReplicaListResponse,…>>>::start<…>::operator() (…)
      at /usr/local/include/async_simple/coro/Lazy.h:422
#18 async_simple::coro::detail::LazyBase<…>::start<…> (…)
      at /usr/local/include/async_simple/coro/Lazy.h:427
#19 async_simple::coro::syncAwait<…> (executor=0x0, cond={_M_counter=0})
      at /usr/local/include/async_simple/coro/SyncAwait.h:45
#20 mooncake::MasterClient::invoke_batch_rpc<&WrappedMasterService::BatchGetReplicaList,…>
      (this=0x7d0c9ab46c30, input_size=<optimized out>)
      at /path/to/Mooncake/mooncake-store/src/master_client.cpp:416
      pool = shared_ptr<coro_io::client_pool<…>> (use count 2) = {get()=0x7d0c9b500010}
#21 mooncake::MasterClient::BatchGetReplicaList
      (this=0x7d0c9ab46c30, object_keys=vector(1), tenant_id="default")
      at /path/to/Mooncake/mooncake-store/src/master_client.cpp:559
#22 mooncake::Client::BatchQuery
      (this=0x7d0c9ab46c00, object_keys=vector(1), tenant_id="default")
      at /path/to/Mooncake/mooncake-store/src/client_service.cpp:1198
#24 mooncake::RealClient::batch_get_into_internal
      (this=0x7d0fdc40b700, keys=vector(1), buffers=vector(1), sizes=vector(1))
      at /path/to/Mooncake/mooncake-store/src/real_client.cpp:4870
#27 mooncake::RealClient::batch_get_into
      (this=0x7d0fdc40b700, keys=…, buffers=vector(1), sizes=vector(1))
      at /path/to/Mooncake/mooncake-store/src/real_client.cpp:4519
#28 MooncakeBackend::batch_get_into
      (this=0x7d1875a4e1c0, keys=…, result=…)
      at /path/to/cache_client.cpp:626

```
## Reproduction condition
- Reproduce: `MC_RPC_TIMEOUT_MS=20`
- Stable‑ok: `MC_RPC_TIMEOUT_MS=400`

## Analysis & Questions
Reference coro_rpc doc:
https://github.com/alibaba/yalantinglibs/blob/main/website/docs/zh/coro_rpc/coro_rpc_client.md#%E7%BA%BF%E7%A8%8B%E5%AE%89%E5%85%A8

Only partial APIs are thread‑safe for coro\_rpc\_client. `connect / close / object destruction` must not run concurrently with in‑flight `send_request`.

Race flow in our workload:

1. `MC_RPC_TIMEOUT_MS=20`, many RPC timeout rapidly.
2. Upper layer detect consecutive timeouts, immediately call `tearDownAll()` and reset RealClient object.
3. Some coro‑rpc coroutines are still suspended and in‑flight.
4. Coroutine resumes and operates already‑freed socket reactor resource → null‑pointer crash.

# Questions:

1. Is this crash caused purely by race between in‑flight RPC and client destruction?
2. For read‑only `RealClient` (`global_segment_size=0`), should Mooncake add in‑flight request counter / guard to prevent destroying client with outstanding requests?
3. What is the recommended safe pattern to tear‑down & recreate read‑only client on RPC timeout?

he-yufeng:
## Description

Fixes #3909: tearing down a read-side client after consecutive RPC timeouts segfaults in asio's `epoll_reactor`. Two lifetime violations, both verified by reading current main against ylt's documented contract (only `send_request` is thread-safe; close/destruction must not race in-flight RPC):

1. **In-flight requests.** `MasterClient`/`DummyClient`/`ClientRequester` destructors just dropped the client pool while `co_await pool->send_request(...)` coroutines could still be suspended. A shared `RpcDrainGuard` (mooncake-common) stops admitting new calls once draining and makes the destructor wait for in-flight ones, bounded at 30s with a loud error rather than silently proceeding.

2. **Background reconnect coroutines.** ylt's `client_pool` reconnect loop (`client_pool.hpp`) holds references into pool storage and resumes across sleeps, so no amount of request draining makes pool destruction safe mid-retry. `RpcClientPool` now hands out pools from a process-wide registry keyed by address instead of freeing them on teardown or address flap. Pools are one-per-distinct-address per process and deliberately process-lifetime.

## Module

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

**Test commands:**
```bash
cmake --build build-linux --target rpc_drain_guard_test rpc_timeout_test master_service_test master_service_ha_test client_integration_test
```

**Test results:**
- [x] Unit tests pass

Linux container (Ubuntu 22.04): new `rpc_drain_guard_test` (3 tests: drain waits for in-flight and refuses new calls, empty drain returns immediately, timeout reports false) passes. `rpc_timeout_test` gains `TeardownDrainsInFlightTimeoutRequests`: 16 concurrent `ExistKey` calls in flight against a black-hole master at client teardown, 10 rounds, green with the fix. `master_service_test` 69/69 and `master_service_ha_test` 95/95 pass (one earlier run flaked once on an unrelated timing test under container load; two clean runs followed).

Honest limitation: the reporter's crash is timing-dependent at 200-node scale, and my in-process harness did not crash pre-fix in this shape, so I am not claiming a local red-to-green crash repro. The fix is verified by construction against ylt's documented thread-safety contract plus the guard semantics tests above.

## Checklist

- [x] I have formatted my code using `./scripts/code_format.sh` (clang-format plus pre-commit on every touched file; all hooks pass)
- [ ] For changes >500 LOC: I have filed an RFC issue (N/A: +322/-35 incl. tests)

## AI Assistance Disclosure

- [ ] No AI tools were used

Prepared with AI assistance (Kimi Code): the mechanism chain (drain + reconnect-storage) was traced on current main by reading, the fix and tests were drafted by the agent, and I reviewed the complete diff before submitting. All verification runs above were actually executed, and the no-deterministic-crash-repro limitation is stated as observed.

## Update (2026-09-11, head f9844e4c)

Reviewer questions closed:

- `ClientRequester`'s offload pool collection is no longer freed at teardown at all (`detail::KeepClientPoolsAlive`): the ylt reconnect loop references pool storage whether or not a user call is in flight, so freeing was the original UAF class on every teardown, not just on drain timeout. The 30s drain now covers response correctness only; on timeout, in-flight calls lose their answers, but no late resume touches freed storage. Verified against the offload RPC call sites (`invoke_rpc<&RealClient::batch_get_offload_object>` and `release_offload_buffer`), which both enter through the guarded `invoke_rpc`.
- The pool registry stores the first configuration's salient knobs (`max_connection`, connect/request timeouts) and warns loudly on a later config mismatch for the same address instead of silently keeping first-wins.
- New tests: `RegistryKeepsFirstConfigForSameAddress`, `KeepClientPoolsAliveRetainsCollection`. rpc_client_io_context_test 5/5, rpc_drain_guard_test 3/3, rpc_timeout_test 6/6, all in the Linux container.


he-yufeng:
Follow-up on head dbc39a03: the CTest lane caught `RpcClientIoContextPoolTest.ReplacesPoolWhenTargetChanges`, which pinned the old pool-freeing contract (`EXPECT_TRUE(old_pool.expired())` after release). That contract is exactly what this PR changes by design — pools are process-lifetime now because a freed pool's background reconnect coroutines still reference its storage. The test now asserts the new contract (same address same pool, different address different pool, released pool stays alive). Green locally in the container.

he-yufeng:
The CTest leg failed on `CloseDmabufExport.ClosesLiveFdAndClearsIt` ("fd 8 should be closed", actual false) in `dmabuf_export_test`, a transfer-engine binary that does not link the code this PR touches (the RPC pool registry lives in mooncake-store's client teardown path). The assertion checks an fd right after `close()` in a shared CI process, so an fd-number reuse under parallel load fits the observed "Actual: false" shape. Could a maintainer re-run the job? If it recurs on rerun I will dig into the test's fd bookkeeping instead.

Icedcoco:
Thanks for putting this together. The underlying issue is important, and the drain guard is a reasonable direction. I had a few questions about the remaining lifecycle and resource semantics before we can consider merging:

1. Drain timeout and pool release

   In the destructors, the pool is still released after drain_for(30s) returns false. Could this still leave a suspended RPC accessing the pool after teardown, preserving the original UAF risk in the timeout or cancellation-disabled cases?

   Would it be safer to keep the pool alive until all in-flight calls have completed, or otherwise explicitly cancel and join those calls before releasing the client state?

2. Process-wide pool registry

   The new registry keeps one pool permanently for every address ever used by the process. Could this grow without bound in deployments where peer addresses or ports are dynamic?

   Also, since the first pool created for an address determines its configuration, how should we handle two users selecting different timeout, protocol, or IO-context settings for the same address? Is the process-wide “first configuration wins” behavior intentional and guaranteed by the surrounding APIs?

3. ClientRequester coverage

   The guard is added to ClientRequester::invoke_rpc(), but could we add teardown tests covering all relevant offload paths, including batch_get_offload_object() and release_offload_buffer()? This would help confirm that no public path can outlive client_pools_.

The fix is valuable, but I would prefer to resolve these points or document the intended trade-offs explicitly before giving approval.

he-yufeng:
@Icedcoco Good catches — all three points were real gaps, and the new head (will note the SHA after push) closes them:

**1. Drain timeout / pool release.** You were right, and it was worse than the drain-failure case: the ylt pool's background reconnect loop references pool storage whether or not a user call is in flight, so freeing `client_pools_` at teardown was the original UAF class on every teardown, drain or not. `~ClientRequester` no longer frees the collection at all — it is parked process-wide (`detail::KeepClientPoolsAlive`, collections are one per RealClient, so few in practice). The 30s drain now only covers response correctness: if it times out, in-flight calls lose their answers, but no late resume ever touches freed storage. That matches how the MasterClient read path already works (the process-wide registry), so both RPC planes now share one lifetime rule.

**2. Registry config.** Also fair. The registry now stores the first pool's salient knobs (`max_connection`, connect/request timeouts) and warns loudly on a later mismatch for the same address instead of silently keeping the first config. I deliberately kept first-wins rather than keying by config fingerprint: two pool configurations for one address in one process is a configuration smell, and a loud warning surfaces it instead of multiplying pools. If you would rather fingerprint the key, it is a three-line change from here.

**3. Offload path coverage.** Verified end to end: both `ClientRequester::batch_get_offload_object` and `release_offload_buffer` route through `ClientRequester::invoke_rpc`, which takes the `RpcDrainGuard::ScopedCall` on entry (real_client.cpp around the `invoke_rpc<&RealClient::batch_get_offload_object>` and `release_offload_buffer` call sites). So the RPC side of both paths is covered by the same guard. The byte plane (TransferEngine `openSegment`/transfer requests) is a different subsystem with no pool-backed coroutines and is out of this PR's scope by design.

New tests pin 1 and 2: `RegistryKeepsFirstConfigForSameAddress` (same address, differing config, same pool object back) and `KeepClientPoolsAliveRetainsCollection` (collection outlives its owner). The rpc_timeout and drain-guard suites stay green.

he-yufeng:
@Icedcoco Merged current main into the branch (424c49ee). The conflicts with your HA connection-policy split (#3743) and the RpcProtocolConfig restructure (#3976) are resolved by keeping both sides, with one placement change worth calling out: the drain guard moved from `invoke_rpc` down into `invoke_rpc_with_client_pool`, so the foreground, HA control, and HA probe paths all drain through one choke point.

The merge also caught a real interaction between this PR and yours, and I want to flag it explicitly because it changes the registry behavior you reviewed. `HaControlPolicyPreservesForegroundRetryPolicy` failed on the merged tree: the HA probe to a refused address took 4299ms where the test expects a fast fail under 750ms. The cause was my shared-pool registry keying pools by address alone. The test's first (foreground) Connect creates the resilient pool for that address; the later HA probe on the same address then got handed the resilient pool, retry budget and all. The config-mismatch warning fired correctly, but "keep the first configuration" is wrong when two policies legitimately coexist on one address, which is exactly what your HA split does.

So the registry key now carries the behavioral knobs (connection count, retry count, reconnect wait, connect and request timeouts, socket flavor) instead of the bare address. Identical configurations still share one pool, which is the case the keep-alive exists for, and different policies get their own pools. The mismatch warning is gone because a mismatch can no longer alias two policies; `RegistryKeepsFirstConfigForSameAddress` became `RegistrySharesPoolOnlyForIdenticalConfig` to pin that.

Verified on the merged head in the dev container: `rpc_timeout_test` 8/8 (both your HA tests and the drain loop), `rpc_client_io_context_test` 5/5, `rpc_drain_guard_test` 3/3, and `scripts/code_format.sh --check` clean.

Icedcoco:
Thanks for the follow-up and for addressing the earlier lifecycle and HA-policy concerns. I rechecked the current head and confirmed that the latest CI and relevant tests are green.

I have two remaining questions about teardown behavior:

1. Lifetime of RpcDrainGuard after the 30-second timeout

   If drain_for(30s) times out, the client object and its rpc_drain_ member can still be destroyed while an existing ScopedCall is running. When that RPC eventually returns, ScopedCall::~ScopedCall() calls guard_.leave(), which accesses the guard’s counters and synchronization objects.

   Could this leave a use-after-free in the guard itself, even though the underlying pool is kept alive? Would it be possible for ScopedCall to own a shared state object, or otherwise keep the drain state alive until all in-flight calls have exited?

2. DummyClient teardown RPCs after draining starts

   DummyClient::~DummyClient() calls drain_for() before tearDownAll(). Since drain_for() permanently sets stopping_, the cleanup RPC issued by tearDownAll() through unregister_shm() will be rejected by RpcDrainGuard.

   Is this intentional? If remote shared-memory cleanup is still expected during teardown, should it run before draining, or have a separate teardown path? If the cleanup is deliberately best effort after draining, could we document that behavior and add a focused test?

The registry and HA policy changes look substantially better now, and I am no longer treating CI as an issue. These questions are specifically about the remaining timeout and teardown semantics.

he-yufeng:
@Icedcoco Both were real bugs, thanks for pushing on the teardown semantics. Fixed in 0279884c.

**1. Guard lifetime after a timed-out drain**

You were right, the pool surviving was only half the problem. `ScopedCall` held a reference to the guard, so once `drain_for(30s)` timed out and the owner finished tearing down, the late `~ScopedCall() -> guard_.leave()` ran on freed memory. The comment even admitted the pool side of the risk while the guard side sat there unprotected.

The counters now live in a shared `State` held by `shared_ptr` from both the guard and each `ScopedCall`. A call that outlives its guard still finds valid counters, and `drain_for` semantics are unchanged otherwise.

I gave the regression test teeth by running it under ASAN:

- Pre-fix: `ScopedCallSurvivesGuard` dies with `heap-use-after-free` in `__atomic_base::fetch_sub`, freed stack pointing at the guard destruction. Exactly your scenario.
- Post-fix: clean, and the full suite passes (5/5, including a second new test that pins counter consistency across a timed-out drain followed by a late leave).

**2. DummyClient teardown RPCs after draining starts**

Not intentional, and worse than you described. `unregister_shm()` goes through `invoke_rpc`, so the old order rejected the unmap unconditionally. On top of that the ping thread is only joined inside `tearDownAll()`, so during the 30s drain the rejected pings flipped it into the reconnection loop, which re-registers shm over UDS while the client is trying to tear down. The pre-drain destructor was just `tearDownAll()`, so my ordering managed to break cleanup and add spurious re-registrations at the same time.

`~DummyClient` now runs `tearDownAll()` first while the guard is still open (unmap goes through, ping thread joins), then `drain_for(30s)` as the last step before members die. `MasterClient` and `ClientRequester` destructors only drain and issue no RPCs afterwards, so they are unaffected by the same trap.

Verification: `rpc_drain_guard_test` 5/5, `rpc_timeout_test` 8/8, `rpc_client_io_context_test` 5/5, store objects rebuild clean, pre-commit format gate green.

Icedcoco:
Thanks for addressing the previous lifecycle concerns and for adding the regression coverage. I rechecked the latest head, and the current CI is green, including the CTest and CI Gate jobs.

I have one remaining question about the DummyClient teardown order:

~DummyClient() now calls tearDownAll() before rpc_drain_.drain_for(30s). This preserves the unregister_shm() RPC, which makes sense, but it also means that teardown can reset client_, allocators, shared-memory contexts, and related state while previously admitted RPCs are still running.

Could we confirm that those in-flight RPCs only retain and access their independent client-pool state, and never touch the DummyClient members being cleared by tearDownAll()? If that invariant is guaranteed, documenting it or adding a focused regression test would make the ordering easier to validate.

If the in-flight calls can access any of those members, would it be possible to split teardown into phases—for example, handle the required unmap operation, prevent further activity, wait for existing calls, and then release the local resources?

This is a clarification about the remaining teardown invariant; the earlier guard lifetime, pool lifetime, HA policy, and CI concerns appear to be addressed in the latest revision.

he-yufeng:
@Icedcoco Good question — I walked every member access on the in-flight path to answer it.

**What an in-flight RPC actually is.** The drain only ever waits for `RpcDrainGuard::ScopedCall` holders, and those live inside `invoke_rpc` / `invoke_batch_rpc` (dummy_client.cpp:186, 223). The guard is RAII across the entire `syncAwait`, so an "in-flight RPC" is exactly the code inside that scope, nothing else.

**What that scope touches.** Reading every access inside the guard: `client_accessor_.GetClientPool()` (the pool is kept alive process-wide by the registry — that is the #3909 design), `connected_.load()` (an atomic; `tearDownAll()`'s store to it is race-safe), by-value arguments, and the drain's shared `State` (a `shared_ptr`, kept valid past the guard's own destruction precisely for a late `leave()`). That is the complete set.

**What it never touches.** `shm_helper_`, `registered_external_buffers_`, `local_buffer_base_`, `hot_cache_base_`/`hot_cache_fd_`, and the ping members are only ever accessed synchronously on caller threads, outside the drain's scope. A caller that races `~DummyClient` on those is a usage-level data race that no teardown order can fix — it is not a path the drain can or should serialize.

So the invariant holds, and a phased split would gate on a hazard that does not exist on this path. The reverse order remains impossible for the reason you already validated: `unregister_shm()` is itself an RPC and the ping thread is only joined inside `tearDownAll()`, so draining first rejects the unmap at the gate and the ping failures spin the reconnection loop for the whole wait.

Documented the invariant at `~DummyClient()` in 30d2c867 (comment-only change) so the ordering is checkable from the code rather than from this thread.

Icedcoco (approved):
LGTM

ykwd (approved):

PR #3944, opened 2026-09-08
[TE] Metadata cache invalidation on rdma failure

alogfans:
## Description

This PR improves Transfer Engine metadata handling for node replacement scenarios.

** Code are fully revised: ** #3837 -> #3072 -> #3944

When an RDMA failure suggests that cached remote metadata may be stale, the worker now triggers metadata cache invalidation. The invalidation keeps existing segment IDs as handles for in-flight retry paths, but marks remote segment descriptors stale and clears cached RPC metadata so the next lookup reloads fresh metadata from the metadata backend.

The change intentionally does not proactively disconnect all cached RDMA endpoints. Existing endpoint failure cleanup remains responsible for the failed endpoint, while metadata refresh is handled lazily on the next lookup.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
git diff --check
cmake --build build --target transfer_metadata_test -j2
cmake --build build --target worker_pool_rail_state_test -j2
./build/mooncake-transfer-engine/tests/transfer_metadata_test --gtest_filter='TransferMetadataInvalidationTest.*:TransferMetadataVersionTest.*:TransferMetadataPollingTest.*'
./build/mooncake-transfer-engine/tests/worker_pool_rail_state_test
```

**Test results:**
- [x] Unit tests pass

Note: full `transfer_metadata_test` was not run successfully in this environment because legacy etcd-backed tests require metadata storage configuration. The focused metadata invalidation/version/polling tests pass.

## Checklist

Note: `pre-commit` is not installed in the local environment used for this commit.

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex was used to inspect the existing Transfer Engine metadata/RDMA paths, implement the scoped cache invalidation change, add regression coverage, and run the focused build/test commands.

### DIFF
--- a/mooncake-transfer-engine/src/transport/rdma_transport/worker_pool.cpp
+++ b/mooncake-transfer-engine/src/transport/rdma_transport/worker_pool.cpp
@@ -32,8 +32,7 @@
 #include "transport/rdma_transport/rdma_endpoint.h"
 #include "transport/rdma_transport/rdma_transport.h"
 
-// Experimental: Per-thread SegmentDesc & EndPoint Caches
-// #define CONFIG_CACHE_SEGMENT_DESC
+// Experimental: Per-thread EndPoint Cache
 // #define CONFIG_CACHE_ENDPOINT
 
 namespace mooncake {
@@ -276,41 +275,16 @@ WorkerPool::~WorkerPool() {
 
 int WorkerPool::submitPostSend(
     const std::vector<Transport::Slice *> &slice_list) {
-#ifdef CONFIG_CACHE_SEGMENT_DESC
-    thread_local uint64_t tl_last_cache_ts = getCurrentTimeInNano();
-    thread_local std::unordered_map<SegmentID,
-                                    std::shared_ptr<RdmaTransport::SegmentDesc>>
-        segment_desc_map;
-    uint64_t current_ts = getCurrentTimeInNano();
-
-    if (current_ts - tl_last_cache_ts > 1000000000) {
-        segment_desc_map.clear();
-        tl_last_cache_ts = current_ts;
-    }
-
-    for (auto &slice : slice_list) {
-        auto target_id = slice->target_id;
-        if (!segment_desc_map.count(target_id)) {
-            segment_desc_map[target_id] =
-                context_.engine().meta()->getSegmentDescByID(target_id);
-            if (!segment_desc_map[target_id]) {
-                segment_desc_map.clear();
-                LOG(ERROR) << "Cannot get target segment description #"
-                           << target_id;
-                return ERR_INVALID_ARGUMENT;
-            }
-        }
-    }
-#else
+    // Cache only within this submission: a thread-local descriptor cache would
+    // bypass refresh requests consumed by getSegmentDescForTransfer().
     std::unordered_map<SegmentID, std::shared_ptr<RdmaTransport::SegmentDesc>>
         segment_desc_map;
     for (auto &slice : slice_list) {
         auto target_id = slice->target_id;
         if (!segment_desc_map.count(target_id))
             segment_desc_map[target_id] =
-                context_.engine().meta()->getSegmentDescByID(target_id);
+                context_.engine().meta()->getSegmentDescForTransfer(target_id);
     }
-#endif  // CONFIG_CACHE_SEGMENT_DESC
 
     SliceList prepared_slice_list;
     uint64_t submitted_slice_count = 0;
@@ -330,8 +304,9 @@ int WorkerPool::submitPostSend(
         if (selectPeerDevice(peer_segment_desc.get(), slice->rdma.dest_addr,
                              slice->length, context_.deviceName(), buffer_id,
                              device_id)) {
-            peer_segment_desc = context_.engine().meta()->getSegmentDescByID(
-                slice->target_id, true);
+            peer_segment_desc =
+                context_.engine().meta()->getSegmentDescForTransfer(
+                    slice->target_id, true);
             if (!peer_segment_desc) {
                 LOG(ERROR) << "Cannot reload target segment #"
                            << slice->target_id;
@@ -645,6 +620,11 @@ void WorkerPool::performPostSend(int thread_id) {
                         break;
                     }
                 }
+                if (!local_context_inactive) {
+                    for (auto *slice : entry.second)
+                        context_.engine().meta()->requestSegmentRefresh(
+                            slice->target_id);
+                }
                 LOG(WARNING) << "Worker: Cannot make connection for endpoint: "
                              << entry.first
                              << (has_peer_alternative
@@ -678,6 +658,9 @@ void WorkerPool::performPostSend(int thread_id) {
         }
         if (!endpoint->readyToSend()) {
             if (endpoint->readyAckTimedOut()) {
+                for (auto *slice : entry.second)
+                    context_.engine().meta()->requestSegmentRefresh(
+                        slice->target_id);
                 LOG(ERROR) << "Worker: Timed out waiting for RDMA ready ACK "
                            << "for endpoint: " << entry.first
                            << ", deleting endpoint";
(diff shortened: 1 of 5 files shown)


staryxchen:
This PR and #3072 / #3944 look complementary rather than duplicate: both fix stale remote metadata in Transfer Engine, but from different sides.

- #3072 covers “peer is gone”: `syncSegmentCache` currently never invalidates, so a remote that unmounted or expired stays cached and keeps being used. The poller-side `kNotFound` vs `kUnavailable` distinction is the right way to drop that entry only on an authoritative absence.
- #3944 covers “peer was replaced”: on RDMA failure, mark remote metadata stale, keep the `SegmentID` for in-flight retries, and force the next lookup to reload — instead of waiting for the next poll.

Dumping the two diffs together would leave two invalidation models (erase vs mark-stale). Would you be open to co-working on one cache lifecycle instead? Something like Fresh / Stale / Gone, with the poller driving Gone only on authoritative `kNotFound`, and transport failures driving per-peer Stale (keep the ID, reload on next lookup, rather than invalidating every remote on any RDMA error). Happy to help sync on the design if that’s useful.

alogfans:
@staryxchen Thanks for the referring. We will first review and merge #3072, then compact this PR to satisfy that goal.

mooncake-transfer-engine/src/transport/rdma_transport/worker_pool.cpp
```diff
@@ -895,6 +924,10 @@ void WorkerPool::processCompletions(int thread_id,
                 }
                 retry_list = &local_failed_slice_list;
             } else {
+                invalidateRemoteCachesAfterFailure(
+                    context_, std::string("RDMA completion failure for ") +
```
jfeng18: After invalidation, the first `getSegmentDescByID` on the completion path (e.g. inside `hasAvailablePeerRailAlternative`, called right after `invalidateRemoteCachesAfterFailure` in the same stack) now performs a synchronous metadata fetch on the poll-CQ worker thread. When the failure that triggered the invalidation also affects the metadata backend — which the setup-failure trigger explicitly allows, since the comment at L662 acknowledges the local RNIC may be the failed side — how long can this stall that worker's CQ polling for other slices? Was an async/off-thread refresh considered for the first post-invalidation lookup?

mooncake-transfer-engine/src/transport/rdma_transport/worker_pool.cpp
```diff
@@ -632,6 +656,9 @@ void WorkerPool::performPostSend(int thread_id) {
             }
             int setup_ret = setup_result.ret;
             if (setup_ret) {
+                invalidateRemoteCachesAfterFailure(
+                    context_,
```
jfeng18: Three scope questions on the trigger design:

1. The active-setup-failure trigger fires on any handshake failure, including cases where the local RNIC is the failed side (the ambiguity the comment here acknowledges) — in that case remote metadata isn't stale, yet all healthy peers' caches get dropped and each pays a full re-fetch on next lookup.
2. For a dead peer with no alternative rail, `markRailFailed` is skipped (guarded by `hasAvailablePeerRailAlternative`), so under sustained traffic the invalidation — and the resulting full re-fetch across *all* peers — recurs every 100ms cooldown window. That looks like a periodic global metadata storm while one peer is down.
3. The cooldown `static` inside `invalidateRemoteCachesAfterFailure` is process-global across all engines. For multi-engine processes (e.g. the store pool split in #3450), engine A's invalidation suppresses engine B's within the window.

Were a narrower trigger set (e.g. only REMOTE_ACCESS_ERR-class WC failures), a per-peer scope, or a per-engine cooldown considered?

mooncake-transfer-engine/src/transport/rdma_transport/worker_pool.cpp
```diff
@@ -895,6 +924,10 @@ void WorkerPool::processCompletions(int thread_id,
                 }
                 retry_list = &local_failed_slice_list;
             } else {
+                invalidateRemoteCachesAfterFailure(
```
jfeng18: Since `peer_nic_path` is server_name + device name and both are typically unchanged across node replacement, the cached endpoint for a replaced node stays `connected()` and serves the first post-replacement slice over a dead QP before the WC failure retires it — and each per-context endpoint instance pays that failure separately, each one then re-triggering the global invalidation (compounding with the cooldown question above).

This is the acknowledged trade-off of not proactively disconnecting. Did you consider retiring the failed server's endpoints across all contexts as part of invalidation, or tying endpoint validity to the refreshed descriptor, so the first retry after replacement doesn't have to fail first?

jfeng18 (commented):
Read through the invalidation flow end-to-end (triggers → `invalidateRemoteCaches` → the lazy-reload paths in `getSegmentDescByName`/`getSegmentDescByID`). The core semantics look clean: the invalid-set check happens under the same read lock as the cache hit, the else-branch re-check closes the TOCTOU window, the two-phase invalidation uses independent locks, and `was_invalid → nullptr` is the right conservative call versus returning a suspect stale descriptor. No correctness blockers from me.

Three design-level questions below — the first two are about failure amplification under sustained faults, the third is about the convergence cost of not coupling with the endpoint cache. Plus a small test-coverage nit.

(Context: these follow from the review I left on #3016 before it was closed; the clear()-predecessor concern from there doesn't apply here since this PR's two-phase design has no cross-lock assumptions.)

mooncake-transfer-engine/tests/transfer_metadata_test.cpp
```diff
@@ -495,6 +495,58 @@ TEST(TransferMetadataPollingTest, PollingRefreshesCachedRemoteSegmentDesc) {
     FAIL() << "TE metadata refresh polling did not refresh cached descriptor";
 }
 
+TEST(TransferMetadataInvalidationTest,
```
jfeng18: Nit: the new `was_invalid → nullptr` branches are unreachable in the current test setup — P2P mode never bumps `metadata_version`, so version stays 0 and `shouldInstallFetchedSegmentDesc` always installs. Could the test set an explicit version (as TransferMetadataVersionTest does) to cover the version-regression-under-invalidation path, plus the `getSegmentDescByName` invalid path?

alogfans:
We revised this PR to focus on the metadata recovery needed following merged PR #3072 for single-node replacement under the same segment name. The patch marks failed peers for refresh on subsequent transfers, preserves refresh requests across lookup failures, and refreshes the affected peer’s RPC address after handshake failures.

jfeng18:
Reviewed the revision end-to-end. It addresses all four points from my earlier round, in ways cleaner than what I suggested:

- The completion path now records failures via `requestSegmentRefresh` with no network I/O, and rail selection keeps serving from the cached descriptor, so a metadata-backend outage can no longer stall CQ polling. `CompletionBurstDoesNotFetchMetadataForRailSelection` pins this down nicely with the reads counter.
- Scope is now per-segment: healthy peers' caches are untouched, local/flush WC failures are excluded, and the process-global static cooldown is gone entirely.
- The consume-before-fetch / re-arm-on-failure semantics in `getSegmentDescForTransfer` (with the `after_get` interleaving tests) is a cleaner answer to the lost-refresh-during-lookup-failure window than anything I had proposed.
- RPC address refresh on handshake failure, with version checks preserved, closes the loop for same-name node replacement.

One CI note: the only failures (`CI Gate`, `tent-ci (ub-mock)`) trace to `tent_runtime_queue_dispatch_test` (RuntimeQueue.RejectsCancellationForUnsupportedTransport, "runtime queue dispatch window must be non-zero"), which doesn't overlap this change's classic-TE metadata paths — #4066 touches that exact test file.

No further comments from me; this looks ready.

alogfans:
@staryxchen What about your opinion?

staryxchen (approved):
LGTM overall. Two non-blocking notes:

1. On P2P, and whenever `metadata_version == 0`, `shouldInstallFetchedSegmentDesc` accepts every fetched descriptor. `submitPostSend` and `redispatch` can force-fetch the same id at the same time, and each clears the refresh bit before the fetch. A slower fetch of the pre-replacement rkey can overwrite the newer one, and later submits keep using it until the next remote completion failure. Versioned etcd/HTTP metadata rejects that older writeback. A local generation checked at install time would close this.

2. `sendHandshake` refreshes RPC metadata on every non-P2P failure via `getRpcMetaEntryInternal(..., true)`, which holds `rpc_meta_lock_` across the backend read. One handshake failure then stalls other peers' RPC lookups for that round trip. Fetching outside the lock would keep the recovery behavior. A rejection or a `notify_rq_depth` decode failure means the address was reachable and does not need a refresh.

PR #3951, opened 2026-09-08
[Store] Interleave the stale-handle sweep with concurrent RPCs

### LINKED ISSUE #3940: [Bug][Store]: Lease-expired client cleanup blocks all master RPC for minutes under HA + SSD Offload
### Bug Report

**Summary**

When a store's lease expires (client process stalls but doesn't crash — misses pings until lease TTL), the master's `ClientMonitorThreadFunc` synchronously runs `ClearInvalidHandles` → `ClearStaleHandles`: an O(total keys) full sweep of all shards, holding each shard's write lock. Because the sweep visits **every** key (not just the expired client's), cleaning a client that owns a large replica set (tens of millions, a fraction of total keys) takes minutes. During that time **all** RPCs — including `Ping` and `FetchTasks` — are blocked (`Total=0.00` in admin metrics), so the master appears hung even though the process is alive and `view_version` is unchanged.

> Reproduced on a deployment with SSD Offload enabled and **HA on, OpLog off**. So the bottleneck below is **not** OpLog/etcd throughput — it holds with pure in-memory deletion.

**Symptoms (observed in production)**

- ~7-minute "unresponsive" window (10:24:25 → 10:31:20).
- Admin metrics keep printing every 10s, `role=leader`, `view_version` unchanged (no HA re-election) → the master process is alive.
- During the window `PutStart/Get/Ping/FetchTasks` all show `Total=0.00`; `Clients` 8→7→8 (self-recovered at 10:31:46).
- `Mem Storage` pinned at 4.41 TB / 5.47 TB (80.7%) — unchanged, because the cleanup targets `local_disk`/SSD replicas, not memory.
- `SSD Storage` dropped 5.32 TB (37.47 → 32.15 TB); `Keys` dropped ~33.7M (234.1M → 200.5M).
- All requests from the inference engine (sglang) were blocked — get/put throughput was 0 for 7 minutes.
- All admin-metric counters (`Eviction/Discard/Promotion/Snapshot`) stayed 0 / unchanged, because `ClearStaleHandles` doesn't update any of them — making the work nearly invisible during diagnosis.

**Root cause analysis** — three compounding factors:

1. **The lease-expire path is unconditionally synchronous.** `ClientMonitorThreadFunc` calls `ClearInvalidHandles(alive_clients)` directly ([`master_service.cpp:11576`](https://github.com/kvcache-ai/Mooncake/blob/main/mooncake-store/src/master_service.cpp#L11576)), unlike the normal unmount path which can go async via `replica_cleanup_worker_.Schedule()` ([`master_service.cpp:2858`](https://github.com/kvcache-ai/Mooncake/blob/main/mooncake-store/src/master_service.cpp#L2858)). The lease-expire path doesn't even consult `enable_async_segment_cleanup_`.

2. **`ClearStaleHandles` holds each shard's write lock for the entire shard sweep** ([`master_service.cpp:2729-2798`](https://github.com/kvcache-ai/Mooncake/blob/main/mooncake-store/src/master_service.cpp#L2729)). `MetadataShardAccessorRW shard(this, i)` scopes the whole shard loop; `PutStart`/`Get`/`Ping` all need a read lock on the shard → all blocked for the sweep duration.

3. **`ClearStaleHandles` is O(total keys), not O(stale keys).** The sweep iterates **every** key in **every** shard/tenant and calls `BuildStaleHandleCleanupPlan` on each ([`master_service.cpp:2738`](https://github.com/kvcache-ai/Mooncake/blob/main/mooncake-store/src/master_service.cpp#L2738)), which walks that key's replicas and runs `is_stale` — and `has_stale_local_disk_client` does an `alive_clients.find()` hash lookup per replica ([`replica.h:465`](https://github.com/kvcache-ai/Mooncake/blob/main/mooncake-store/include/replica.h#L465)). So even though only ~33.7M of 234M keys were stale, **all 234M keys are visited and judged**. Stale keys then go through `EraseMetadata` ([`master_service.cpp:2252`](https://github.com/kvcache-ai/Mooncake/blob/main/mooncake-store/src/master_service.cpp#L2252)), which per key does several hash lookups/erasures (offloading_tasks, processing_keys, replication_tasks, soft-pin index). The whole pass is a large, cache-miss-bound in-memory traversal. This is independent of OpLog: with `enable_oplog_=false` the deletion path is pure in-memory `EraseMetadata` (no etcd write), yet the sweep still took ~7 min. OpLog, when enabled, adds a per-replica `Reserve`+`Commit` on top ([`master_service.cpp:2205-2226`](https://github.com/kvcache-ai/Mooncake/blob/main/mooncake-store/src/master_service.cpp#L2205)); `Commit` is non-blocking but `Reserve` returns `TASK_PENDING_LIMIT_EXCEEDED` at 1024 in-flight slots ([`ordered_oplog_writer.cpp:189-191`](https://github.com/kvcache-ai/Mooncake/blob/main/mooncake-store/src/ha/oplog/ordered_oplog_writer.cpp#L189)), which can only make things slower.

**Proposed approaches**

- **A:** Make `ClearStaleHandles` release the shard write lock periodically (time-sliced, e.g. every ~50ms) and resume from a key cursor (`lower_bound(last_key)`). While the lock is released, other RPCs may insert, delete, or modify keys in that shard — on resume, those keys are re-judged against `is_stale`: newly inserted keys don't belong to the expired client and are skipped, already-deleted keys are skipped by `lower_bound`, so releasing the lock mid-sweep is consistent. When OpLog is enabled, `ClearStaleHandles` only enqueues the OpLog entry + `mark_removed` and the actual `EraseMetadata` runs in the durable callback, which likewise supports releasing the lock mid-sweep. This keeps the master responsive regardless of cleanup volume — the "hung" symptom is fixed even if cleanup still takes minutes.
- **B:** Maintain a client→replica reverse index so lease-expire cleanup scans only that client's replicas (O(its replicas) vs O(total keys)). This directly attacks factor 3 — turning a 234M-key sweep into a 33.7M-key one. Larger change; suggest as a follow-up.
- **C (only when OpLog is enabled):** Increase `max_entries_per_batch` (1024 → larger) and/or batch `ClearStaleHandles`' OpLog submissions. Doesn't apply when OpLog is off (as in this report); when on, combine with A.

**Discussion**

- Is there a correctness reason `enable_async_segment_cleanup_` is disabled under HA, or is it conservative? If it can be lifted, the lease-expire path could use `replica_cleanup_worker_` directly.
- Which approach is preferred? Or if there's a better one, please share.
- Any concerns with the time-sliced lock release in `ClearStaleHandles` (cursor validity, intermediate consistency)?

### Before submitting...

he-yufeng:
## Description

Fixes #3940. A lease-expiry cleanup ran `ClearStaleHandles` while holding each shard's write lock for the entire shard sweep — every tenant, every key, with the HA persist path inline per key. At tens of millions of keys the sweep blocked all RPC on that shard for minutes (the production trace in the issue: `Total=0.00` for Put/Get/Ping/FetchTasks, master alive, no re-election).

The sweep now works at a different lock cadence, with identical cleanup semantics:

- Per shard, the key set is snapshotted under one short read lock, then the stale predicate is evaluated in 256-key chunks, releasing the read lock between chunks.
- Each candidate is committed under its own short write lock, re-verified against current state (a candidate that moved out of stale between passes is skipped).
- A `yield` between candidates lets a waiting writer win the lock.

What gets cleaned, the HA persist path, and the tenant-empty erase rule are unchanged — only when the locks are held.

## Module

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

**Test commands:**
```bash
cmake --build build-linux --target master_service_test master_service_ha_test
./mooncake-store/tests/master_service_test
./mooncake-store/tests/master_service_ha_test
```

**Test results:**
- [x] Unit tests pass

New regression test `StaleHandleSweepInterleavesConcurrentWrites`: 600 keys seeded on the writer's shard, a 10ms slow stale predicate stretches the sweep to several seconds, and a `PutStart` issued mid-sweep must return in far less than the sweep's remaining time. It fails on current main (the write waits out the whole sweep: 13.6s of a 14.1s sweep in the measured red run) and passes here. In a Linux container (Ubuntu 22.04): `master_service_test` 70/70 and `master_service_ha_test` 95/95 pass.

The existing cleanup correctness coverage (the HA persist paths, tenant erase, quota accounting) is unchanged and green; the predicate is re-evaluated under the write lock before every commit, so a key whose state moved between passes is left alone.

## Checklist

- [x] I have formatted my code (clang-format) and run pre-commit on the touched files; all hooks pass
- [ ] For changes >500 LOC: I have filed an RFC issue (N/A)

## AI Assistance Disclosure

- [ ] No AI tools were used

Prepared with AI assistance (Kimi Code): the lock chain was traced on current main by reading, the fix and tests were drafted by the agent, and I reviewed the complete diff before submitting. The container runs and the red/green evidence above are the ones actually executed.

### DIFF
--- a/mooncake-store/src/master_service.cpp
+++ b/mooncake-store/src/master_service.cpp
@@ -2728,73 +2728,117 @@ void MasterService::ClearLocalDiskHandlesOwnedBy(const UUID& owner) {
 void MasterService::ClearStaleHandles(
     const std::function<bool(const Replica&)>& is_stale) {
     for (size_t i = 0; i < kNumShards; i++) {
-        MetadataShardAccessorRW shard(this, i);
-        for (auto tenant_it = shard->tenants.begin();
-             tenant_it != shard->tenants.end();) {
-            auto& tenant_state = tenant_it->second;
-            auto it = tenant_state.metadata.begin();
-            while (it != tenant_state.metadata.end()) {
-                const auto cleanup_plan =
-                    BuildStaleHandleCleanupPlan(it->second, is_stale);
-                if (!cleanup_plan.removed_ids.empty()) {
-                    if (enable_ha_) {
-                        if (enable_oplog_) {
-                            auto persist_result =
-                                PersistStaleHandleCleanupForHA(
-                                    "ClearStaleHandles", tenant_it->first,
-                                    it->first, it->second, cleanup_plan);
-                            if (!persist_result) {
-                                ++it;
-                                continue;
-                            }
-                            ++it;
-                            continue;
-                        }
-                    }
-                    if (CleanupStaleHandles(it->first, tenant_it->first,
-                                            tenant_state, it->second, is_stale,
-                                            &shard)) {
-                        it = EraseMetadata(tenant_state, it, tenant_it->first,
-                                           QuotaEraseMode::kFull, &shard);
-                    } else {
-                        ++it;
-                    }
-                } else if (!it->second.IsValid()) {
-                    if (enable_ha_) {
-                        if (enable_oplog_) {
-                            auto persist_result =
-                                AppendOpLogWithDurableFinalize(
-                                    OpType::REMOVE, tenant_it->first.value(),
-                                    it->first, {},
-                                    [this](const OpLogEntry& durable_entry) {
-                                        FinalizeMetadataEraseAfterDurable(
-                                            durable_entry,
-                                            QuotaEraseMode::kFull);
-                                    });
-                            if (!persist_result) {
-                                LOG(WARNING)
-                                    << "ClearStaleHandles(last replica)"
-                                    << ": REMOVE persist failed for key="
-                                    << it->first << ", err="
-                                    << static_cast<int>(persist_result.error());
-                                ++it;
-                                continue;
-                            }
-                            ++it;
-                            continue;
-                        }
-                    }
-                    it = EraseMetadata(tenant_state, it, tenant_it->first,
-                                       QuotaEraseMode::kFull, &shard);
-                } else {
-                    ++it;
+        // Pass 1 (read-only): collect the (tenant, key) pairs whose current
+        // state needs cleanup. The sweep used to hold the shard write lock
+        // for the whole shard and run the HA persist path inline per key,
+        // which blocked every RPC on that shard for minutes at large key
+        // counts (#3940). Candidates are committed one key at a time below,
+        // each under its own short write lock, so other work interleaves.
+        // Pass 1 (read-only, chunked): snapshot the shard's keys under one
+        // short read lock, then evaluate the stale predicate in bounded
+        // chunks, releasing the lock between chunks so writes interleave
+        // during the scan itself.
+        std::vector<std::pair<TenantId, std::string>> shard_keys;
+        {
+            MetadataShardAccessorRO shard(this, i);
+            for (const auto& [tenant_id, tenant_state] : shard->tenants) {
+                for (const auto& [key, metadata] : tenant_state.metadata) {
+                    shard_keys.emplace_back(tenant_id, key);
                 }
             }
-            if (tenant_state.Empty()) {
-                tenant_it = shard->tenants.erase(tenant_it);
-            } else {
-                ++tenant_it;
+        }
+        constexpr size_t kScanChunkSize = 256;
+        std::vector<std::pair<TenantId, std::string>> candidates;
+        for (size_t start = 0; start < shard_keys.size();
+             start += kScanChunkSize) {
+            const size_t end =
+                std::min(start + kScanChunkSize, shard_keys.size());
+            std::this_thread::yield();
+            MetadataShardAccessorRO shard(this, i);
+            for (size_t k = start; k < end; ++k) {
+                const auto& [tenant_id, key] = shard_keys[k];
+                auto tenant_it = shard->tenants.find(tenant_id);
+                if (tenant_it == shard->tenants.end()) {
+                    continue;
+                }
+                auto it = tenant_it->second.metadata.find(key);
+                if (it == tenant_it->second.metadata.end()) {
+                    continue;
+                }
+                const auto cleanup_plan =
+                    BuildStaleHandleCleanupPlan(it->second, is_stale);
+                if (!cleanup_plan.removed_ids.empty() ||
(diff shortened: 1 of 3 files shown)


he-yufeng:
Rebased onto current main (06261401). The conflict was in master_service_test.cpp: #3927 migrated the eviction/hard-pin scenarios to the MasterScenario DSL and deleted the block my interleave test was appended to. The rebased tree keeps upstream's deletions verbatim and places the sweep test under its own section header. Full master_service_test suite green on this head (55/55, container Ubuntu 22.04), including StaleHandleSweepInterleavesConcurrentWrites.

ykwd:
Thanks for the thorough write-up and the regression test — the lock-hold analysis on ClearStaleHandles matches what we see as well.

We are currently replacing the per-shard metadata lock with a per-group lock, which should remove this stall at the lock-granularity layer rather than by time-slicing the existing shard sweep.

cc.  @CAICAIIs @Aionw

he-yufeng:
@ykwd Understood, and that is the better layer to fix it at. A per-group lock removes the stall structurally, while this PR only makes the existing shard sweep interruptible.

Two ways to play it, your call:

- If the group-lock rework is landing soon, I am happy to close this and let the structural fix carry it. The regression test (600-key shard, slow predicate, concurrent writer) would translate cleanly onto the new locking and is the part worth keeping.
- If there is room for both, this PR composes with it: even with per-group locks, a full sweep of one group still holds that group's lock for the whole scan, and the chunked-commit-plus-yield shape keeps long sweeps from parking the group's RPCs either way. It is semantics-preserving on the cleanup path, so it does not complicate the rework.

Tell me which you prefer; I will close or rebase accordingly.

he-yufeng:
Closing as superseded: #3593 ("Sweep stale handles in bounded write-lock batches", landed 09-09 on main) implements the same mechanism this PR carried — a shared-lock pick pass followed by bounded write-lock batches with re-lookup per key — which fixes the #3940 stall by capping each exclusive hold at 64 keys and releasing between batches, so waiting RPCs get in. The remaining difference (a yield between candidates) is not worth a second implementation of the same sweep.

@ykwd the per-group locking you mentioned is now partly pre-empted by this landed batching; if the group-level granularity still goes in, nothing in the sweep should need to change.

PR #3952, opened 2026-09-08
[TransferEngine] Prefer THP for shared-segment mmap without a HugeTLB pool

ascend-direct-dev:
## Description

`create_shared_segment(mmap=True)` previously used POSIX shm (`shm_open` + a name in `/dev/shm`) and only applied `MADV_HUGEPAGE` after the mapping existed. That left large HostRegister MRs on 4KiB pages unless the kernel happened to collapse them.

This switches the mmap backend to an unnamed `memfd` (no `MFD_HUGETLB`, no `vm.nr_hugepages`) and asks for THP **before** the first fault:

1. Reserve a 2MiB-aligned `PROT_NONE` window
2. `mmap` the memfd there with `MAP_SHARED|MAP_FIXED`
3. `madvise(MADV_HUGEPAGE)` then prefault one byte per PMD
4. `madvise(MADV_DONTFORK)`

If THP cannot allocate, the kernel keeps 4KiB pages. TP ranks still share the same physical pages through `/proc/<pid>/fd/<n>`.

Relates to the original shared-segment work in #3285. No overlapping open PR for this mmap path.

## Module

- [ ] CI/CD
- [ ] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [x] Performance improvement
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
# gtest (CI)
# mooncake-transfer-engine/tests/shared_segment_test.cpp
#   OwnerHandleIsProcFd, TwoProcessesSharePages, HostRegisterTwoRanksSharePages

# local Ascend HixlCS check (not in this PR): 32GiB slices, nr_hugepages=0
# register+transfer through 256GiB; two-rank share + transfer
```

**Test results:**
- [x] Unit tests added (proc-fd handle, fork share, HostRegister two ranks)

On an Ascend host with THP `always` / shmem `advise` and `HugePages_Total=0`:

- smaps `ShmemPmdMapped` equaled the full mapping size (2MiB through 256GiB)
- HixlCS `RegisterMem(MEM_DEVICE)` + `TransferSync` succeeded for 1×–8×32GiB
- Two ranks shared one 32GiB segment and transferred correctly

## Checklist

- [x] I have formatted my code using `./scripts/code_format.sh` (`clang-format-20` on the rewritten mmap backend; `git-clang-format` was not available on the build host)
- [x] I have run pre-commit on the files changed in this PR and all hooks pass (`mooncake-code-format` skipped: no `git-clang-format`)
- [x] For changes >500 LOC: I have filed an RFC issue (N/A; under 500 LOC excluding tests)

## AI Assistance Disclosure

- [ ] No AI tools were used

Cursor Grok 4.6 drafted the mmap rewrite and tests. The submitter reviewed the memfd/THP/madvise sequence and the sharing handle.

### DIFF
--- a/mooncake-transfer-engine/src/shared_segment/shared_segment.cpp
+++ b/mooncake-transfer-engine/src/shared_segment/shared_segment.cpp
@@ -235,7 +235,7 @@ bool SharedSegment::Supported(bool mmap, bool host_register) {
         return false;
 #endif
     }
-    // mmap without HostRegister only needs POSIX shm.
+    // mmap without HostRegister only needs an anonymous memfd.
     return true;
 }
 
--- a/mooncake-integration/shared_segment.py
+++ b/mooncake-integration/shared_segment.py
@@ -73,9 +73,12 @@ def create_shared_segment(
     1. ``tp_group`` is a deprecated alias of ``comm_group``. ``device_id``
     defaults to the rank's current accelerator.
 
-    ``mmap`` (default True) uses POSIX shm. Set ``mmap=False`` for the platform
-    VMM fabric path. ``host_register`` (default False) HostRegister's mmap pages
-    for ``device_id`` so ``tensors().data_ptr()`` is a device VA suitable for TE
+    ``mmap`` (default True) uses an unnamed memfd and asks the kernel for THP
+    (``MADV_HUGEPAGE``) so the span can be 2MiB pages without a HugeTLB pool;
+    4KiB pages if THP cannot allocate. Ranks share those pages through the
+    memfd. Set ``mmap=False`` for the platform VMM fabric path.
+    ``host_register`` (default False) HostRegister's mmap pages for
+    ``device_id`` so ``tensors().data_ptr()`` is a device VA suitable for TE
     ``location=\"npu\"`` ROCE D2rH; requires ``mmap=True``. Ascend VMM
     (``mmap=False``) also exposes NPU tensors: the host SVM VA is already
     device-accessible after ``MemSetAccess`` (owner ``MallocMem``, peer
@@ -127,9 +130,7 @@ def create_shared_segment(
     except RuntimeError as exc:
         create_error = str(exc)
 
-    _raise_if_any_rank_failed(
-        create_error, world_size, rank_id, comm_group, "create"
-    )
+    _raise_if_any_rank_failed(create_error, world_size, rank_id, comm_group, "create")
     if segment is None:
         raise SharedSegmentError("Shared segment create returned no segment")
 
--- a/mooncake-transfer-engine/include/shared_segment/shared_segment.h
+++ b/mooncake-transfer-engine/include/shared_segment/shared_segment.h
@@ -56,7 +56,9 @@ struct SharedSegmentOptions {
     uint32_t owner_rank = 0;
     // Accelerator that is granted access to the host pages (D2H/H2D).
     int32_t device_id = 0;
-    // true: POSIX shm_open + mmap on the same host.
+    // true: unnamed memfd + mmap. Prefer THP (MADV_HUGEPAGE) so the kernel
+    // can back the span with 2MiB pages; 4KiB pages if THP cannot allocate.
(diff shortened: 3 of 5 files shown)


mooncake-transfer-engine/src/shared_segment/shared_segment.cpp
```diff
@@ -235,7 +235,7 @@ bool SharedSegment::Supported(bool mmap, bool host_register) {
         return false;
 #endif
     }
-    // mmap without HostRegister only needs POSIX shm.
+    // mmap without HostRegister only needs an anonymous memfd.
     return true;
 }
```
alogfans, line 240: What if memfd_create not supported?

alogfans (commented):
The exported handle is encoded as `/proc/<pid>/fd/<fd>`, while the memfd is created with `MFD_CLOEXEC`:

```cpp
syscall(__NR_memfd_create, "mcss", MFD_CLOEXEC);
```

This means the handle is only usable while the owner process keeps the original file descriptor open. If the owner process calls `exec()` before all peer ranks import the segment, the descriptor is automatically closed and the `/proc/<pid>/fd/<fd>` path becomes invalid. A peer will then fail here:

```cpp
int fd = open(name.c_str(), O_RDWR);
```

For example, a launcher may create the shared segment, perform the all-gather, and then `exec()` or replace the worker process. In that case, a peer that imports the handle after the `exec()` will receive `ENOENT` even though the shared segment was created successfully.

This is a behavioral change from the previous `shm_open` implementation, where the named shared-memory object remained accessible independently of the owner’s file descriptor lifetime.

Could we either:

1. avoid `MFD_CLOEXEC` and explicitly document that the owner must keep the fd open until all peers have imported it; or
2. use a proper fd-passing mechanism such as a Unix domain socket with `SCM_RIGHTS`, instead of serializing a `/proc/<pid>/fd/<fd>` path?

ascend-direct-dev:
@alogfans on the `/proc/<pid>/fd/<n>` handle and `MFD_CLOEXEC`:

The intended use is already-running TP ranks: `Create` → all-gather the blob → `Complete` in those same processes. The owner `SharedSegment` keeps the memfd open for its lifetime, and the existing contract is that the owner's segment must outlive every peer mapping. There is no launcher that creates the segment and then `exec()`s into the worker before peers import.

We would like to keep `MFD_CLOEXEC`. If the training process later `fork`+`exec`s (data loaders, etc.), the huge memfd should not be inherited. Dropping `CLOEXEC` would only help the create-then-exec case, which this API does not do.

Passing the fd with `SCM_RIGHTS` also does not fit this design: the export blob is a fixed-size byte string all-gathered through the framework process group (typically torch), not a side-channel Unix socket. `/proc/<pid>/fd/<n>` is the handle that can travel in that blob while the owner process still holds the fd.

So we are not planning to drop `CLOEXEC` or switch to `SCM_RIGHTS` in this change. The owner must keep the `SharedSegment` object alive; peers import before that object is destroyed, and nobody `exec()`s the owner in between.

mooncake-transfer-engine/src/shared_segment/shared_segment.cpp
ascend-direct-dev: On the Linux hosts this path is meant to run on, `memfd_create` is present: the syscall has existed since 3.17, and we invoke it via `__NR_memfd_create` rather than a glibc wrapper. Mooncake Store's `ShmHelper` already depends on the same syscall with no `shm_open` fallback.

`Supported(mmap=true)` stays a build-time check (same as before for POSIX shm). If the syscall is missing at runtime, `Create` fails with `memfd_create failed: ...` instead of mapping a different backing. We are not adding a named-shm fallback here; that would reintroduce `/dev/shm` objects and a different handle lifetime. Pre-3.17 kernels and non-Linux are out of scope for this backend.

alogfans (approved):
LGTM

PR #3963, opened 2026-09-09
[EP] Support NCCL ElasticBuffer recovery after rank replacement

akhillanger:
## Description

Enable NCCL-backed Mooncake EP `ElasticBuffer` instances to recover after Mooncake PG restores a failed process into the same fixed logical rank slot.

- Keep an existing NCCL generation on NCCL during group-wide transport consensus, including when a replacement process requested `auto`.
- Require every process to observe the complete fixed logical membership before starting NCCL bootstrap.
- Exchange a fresh unique ID, then build and atomically publish a complete communicator/device/window generation. Survivors update their existing buffers while replacement ranks construct matching buffers.
- Add a rank-local abort path so obsolete NCCL device communicators, windows, allocations, streams, and GIN resources can be retired without participation from the failed process.
- Tag EP handles with their NCCL generation and reject stale handles after recovery.
- Document the quiescent-boundary protocol, temporary two-generation resource requirement, and fixed-world limitations.

This complements Mooncake PG failure detection and rank replacement. It does not execute EP operations with missing ranks or resume an operation interrupted by a failure; callers retry that work after the full logical world has recovered.

## Module

- [ ] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [x] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

**Test commands:**
```bash
pre-commit run --files <changed files>
cmake --build "${BUILD_DIR}" --target _ep mooncake_ep_device_stage mooncake_pg_ext device_backend_api_compatibility_test
python -m pytest -q python/tests/ep/test_mooncake_elastic_buffer_transport.py
torchrun --standalone --nnodes=1 --nproc-per-node=2 python/tests/ep/test_elastic_buffer.py --transport nccl --quick --reconfigure --route alltoall
WITH_NVIDIA_PEERMEM=0 python -m pytest -q -s --maxfail=1 mooncake-pg/tests/test_pg_elastic.py::TestMooncakePGElasticCUDA::test_nccl_elastic_buffer_recovery
make -C docs html
```

**Test results:**
- [x] Unit tests pass: 34 transport, consensus, membership, and generation tests.
- [x] Integration tests pass: two consecutive healthy NCCL rebuilds preserved dispatch/combine correctness and rejected a stale handle.
- [x] Manual testing done: on two GB300 GPUs with CUDA 13.1 and NCCL 2.30.6, rank 1 exited abruptly, Mooncake PG detected and replaced its logical slot, both ranks rebuilt NCCL state, the stale handle was rejected, and post-recovery dispatch/combine passed.
- [x] Native NCCL EP, device-stage, PG extension, and device API compatibility targets build successfully.
- [x] Documentation builds successfully.

## Checklist

- [x] RFC not required (fewer than 500 non-test lines added)

## AI Assistance Disclosure

- [ ] No AI tools were used
- [x] AI tools were used

This PR was generated with help of AI. The submitter reviewed the implementation and test results.


python/mooncake/mooncake_elastic_buffer.py
```diff
@@ -384,15 +453,61 @@ def __init__(
...
     def _active_ranks_mask(self) -> list:
         # `mooncake.ep.get_active_ranks` is a Mooncake PG helper and performs a
-        # native static cast to MooncakeBackend. ElasticBuffer transport
-        # bootstrap can also be driven by a regular NCCL/Gloo ProcessGroup; in
-        # that case every rank in the supplied group is active by definition.
-        if "Mooncake" not in type(self.backend).__name__:
+        # native static cast to MooncakeBackend. PyTorch exposes custom process
+        # groups through its base ProcessGroup type, so query the registered
+        # backend name instead of relying on the Python wrapper's type name.
+        if dist.get_backend(self.backend) != "mooncake":
```
UNIDY2002: One nit: the backend could be "mooncake-cpu". That's why we typically use the `"mooncake" in` pattern.

alogfans (approved):
LGTM in Mooncake TE

python/mooncake/mooncake_elastic_buffer.py
```diff
@@ -567,12 +682,50 @@ def _exchange_nccl_unique_id(self, ep: Any) -> List[int]:
...
+        # NCCL cannot patch one peer in place. Join the same bootstrap sequence
+        # used by a replacement rank's constructor and publish a complete new
+        # communicator/device/window generation at a quiescent EP boundary.
+        # Restore the device captured at construction because applications may
+        # change the thread's current CUDA device between EP calls.
+        with torch.cuda.device(self._device_index):
+            torch.cuda.synchronize()
+            from mooncake import ep
+
+            has_nccl_device_support = getattr(ep, "has_nccl_device_support", None)
```
UNIDY2002: `has_nccl_device_support` always exists as an attribute. Maybe change from `getattr` to a direct call?

mooncake-pg/tests/test_pg_elastic.py
```diff
@@ -1508,6 +1727,44 @@ def configure_for_cuda_device_count(cls, device_count: int) -> None:
             return
         cls.world_size = min(device_count, 4)
 
+    def test_nccl_elastic_buffer_recovery(self) -> None:
```
UNIDY2002: Although it reuses the MooncakePGWorkerContext, the case primarily targets EP, so I think the optimal place for this test should be under the ep-related modules. Could you do a reframe on this test if it is feasible?

akhillanger:
@alogfans I have a question:

Currently, NCCL recovery rebuilds the communicator, device context, and communication allocation. A graph captured before recovery still references the old resources, so it cannot safely be replayed afterward. The Python handle-generation check does not protect graph replay because replay bypasses Python.

  There are two possible approaches:

  1. Require recapture after recovery. Document and test a recovery sequence that stops GPU work, rebuilds NCCL state, recreates EP handles, and recaptures affected graphs. This keeps the backend change small, but applications/frameworks retaining graphs on surviving workers would need recovery-path changes.
     If EP is part of a whole-model graph, that containing graph needs recapture.

  2. Preserve existing graphs across recovery. Keep graph-visible buffer/scratch addresses stable and introduce a stable device-side descriptor through which kernels read the current NCCL state. This requires a deeper allocation/lifetime redesign, validation against NCCL window/signal requirements, and
     benchmarking any additional device-side loads. Updating only the communicator pointer would not be sufficient.

  IBGDA impact: Either approach can remain NCCL-specific, without changing existing IBGDA behavior. IBGDA retains its buffer and updates connection/IPC tables in place, so it does not inherently have the same allocation-replacement problem. However, we have not verified graph replay across IBGDA recovery
  with a dedicated test.

  Do current elastic EP integrations retain and replay graphs on surviving workers after update_ep_member(), or already recapture them? Is an explicit recapture requirement acceptable initially, or should preserving existing graphs be a requirement for NCCL recovery?

python/mooncake/mooncake_elastic_buffer.py
akhillanger: Updated the membership check to use the "mooncake" in pattern, with coverage for both backend names. I also fixed NCCL bootstrap to use CPU metadata tensors for mooncake-cpu. An end-to-end test now verifies worker replacement and post-recovery dispatch/combine using a CPU control group with NCCL GPU transfers.

python/mooncake/mooncake_elastic_buffer.py
akhillanger: Replaced the getattr/None checks with direct ep.has_nccl_device_support() calls in both construction and recovery. The binding already returns False when NCCL device support is disabled, so the existing transport-selection behavior is preserved.

mooncake-pg/tests/test_pg_elastic.py
akhillanger: Moved the tests and EP-specific helpers to python/tests/ep/test_elastic_buffer_recovery.py, reusing the existing PG worker harness. The PG elastic test file now matches its pre-PR contents.

UNIDY2002:
@akhillanger I'll take on this question. Thanks for your deep investigation. Although current elastic EP integration assumes preservable graphs across recovery, I think this feature is just a bonus. We can go well with requiring a re-capture initially, and research on more advanced features in follow-up PRs.

UNIDY2002 (approved):

Issue #3425, opened 2026-08-13
[RFC]: Refactor the `mooncake-wheel` package layout


ykwd:
## Summary

Make `mooncake-wheel` the source of truth for every Python module that ships
in the Mooncake wheel, and group that Python by subsystem instead of leaving
it as a flat directory that also receives build-time copies of native
artifacts.

## Motivation

Today the installed Python package is one wheel, but the source layout does
not match that fact.

As a general packaging principle, Python source that ships as part of a
single wheel should live, as much as practical, under one complete and
self-contained source tree. A developer inspecting that tree should be able
to understand what Python code belongs to the package without first reading
the wheel build script or reconstructing which files are copied from other
top-level directories.

This is particularly important for an open-source project. The repository
layout is part of the project's interface to contributors: it affects code
discovery, local development, testing, static analysis, IDE behavior, and
the cost of understanding where a change should be made. When the wheel is
assembled from Python sources scattered across several unrelated directory
trees, the repository no longer represents the package that users actually
install. Build-time copying also creates two notions of package structure:
the source layout seen by contributors and the assembled layout seen by
users.

Mooncake currently has exactly this mismatch:

* `mooncake-reshard/` is a top-level tree whose only distribution path is the
  combined wheel. `scripts/build_wheel.sh` copies
  `mooncake-reshard/python/mooncake/reshard` into
  `mooncake-wheel/mooncake/reshard` and deletes the copy on exit. Developers
  need `PYTHONPATH=mooncake-wheel:mooncake-reshard/python` to test it.
* `mooncake-wheel/mooncake/` is a flat mix of unrelated Python: Store
  helpers, Transfer Engine CLI, vLLM connectors, SSD/SPDK scripts, RL
  DataProto transfer (`structured_object_store.py`, ~6.5k lines), and EP
  loaders. Native `.so` files and binaries are copied into the same
  directory at build time.
* Some Python files are tracked in `mooncake-integration/` and copied into
  the wheel (`async_store.py`, `allocator.py`,
  `allocator_ascend_npu.py`, `fabric_allocator_utils.py`). The wheel is the
  install surface, but not the source of truth.
* The wheel already uses `pkgutil.extend_path` so that an out-of-tree
  `mooncake.reshard` can join the `mooncake` namespace. That split exists
  only because reshard source does not live in the wheel.

The result is extra staging, multiple places to look for Python, and no
obvious home for the next Python-only subsystem. It also makes the package
[...]

PR #3970, opened 2026-09-09
[CI/Build] Unify release wheels on the root scikit-build backend

Aionw:
## Description

Move release wheel assembly to the repository-root scikit-build-core backend used by source installs.

- Replace `scripts/build_wheel.sh` and the legacy setuptools project with a release configuration/metadata adapter plus an isolated auditwheel repair step.
- Migrate CUDA, CPU, EFA, ROCm, MUSA, NPU, nightly, and Docker callers; preserve variant names, output conventions, and configured native build profiles.
- Preserve CUDA EP/PG artifacts during repair, include allocator helpers and Store-only etcd runtime dependencies, and make root version metadata authoritative (0.3.13).
- Add packaging/workflow regression coverage and document migration of removed build commands.

Related RFC: #3534 (also related to #3425 and build roadmap #3556). This implements the release-backend cutover/legacy assembly removal portion, not the complete RFC. Existing subsystem migration PRs such as #3628 and #3623 move Python modules; this PR instead changes wheel construction and its callers. It does not add cibuildwheel, centralize publication, or complete private native-module facades.

## Module

- [x] CI/CD
- [x] Docs

## Type of Change

- [x] Refactor
- [x] Bug fix
- [x] Breaking change
- [x] Documentation update

Breaking change: legacy local packaging commands are removed; see the build migration documentation for replacements.

## How Has This Been Tested?

**Test commands:**
```bash
/tmp/mooncake-review-venv/bin/python -m pytest -q -rs \
  python/tests/packaging \
  mooncake-wheel/tests/test_release_wheel_tags.py \
  mooncake-wheel/tests/test_testpypi_wheel_gate.py
bash -n scripts/repair_wheel.sh
git diff --check
# Used prek because pre-commit is unavailable; run only against changed files.
prek run --files <changed files>
```

**Test results:**
- [x] Unit tests pass: 65 passed, 1 skipped.
- [x] Local installed-wheel integration tests pass: see the CUDA/non-CUDA A/B validation below. The optional pytest smoke test in the earlier command remained skipped because `MOONCAKE_TEST_WHEEL` was not supplied.
- [x] Manual testing done: shell syntax and diff whitespace checks pass.

The scoped hook run passed after auto-fixes. Unrelated pre-existing Python/CMake formatting rewrites were subsequently reverted to keep the PR focused, so the final tree is not claimed to pass all formatting hooks. Regression tests were rerun after those reversions and still pass. Local CUDA/non-CUDA native builds and installed-wheel checks were subsequently completed as described below; the full hardware release matrix remains unvalidated.

### Local before/after wheel validation

Compared `140e1f43` (before the packaging migration) with `ce6460bc` using independently built **CUDA 12.8 and non-CUDA wheels** in the same Ubuntu 22.04 / x86_64 / CPython 3.12.13 environment. The old revision used the legacy setuptools release script; the new revision used scikit-build-core and the new repair script. CUDA builds included EP/PG for one representative Torch ABI (2.13.0).

For each variant, tested a fresh old installation, a fresh new installation, and an in-place `pip install --upgrade` from old to new. The new wheels used a test-only `0.3.14.dev0` version override to exercise pip's upgrade behavior; this does not change the proposed release version.

| Check | non-CUDA | CUDA 12.8 |
|---|---|---|
| Install, pip check, core imports, CLI checks | Pass in all three states | Pass in all three states |
| fast-copy, import, CLI and native BufferPool tests | 25/25 in each state | 25/25 in each state |
| Store two-client TCP put/get/remove (1 B, 4 KiB, 1 MiB) | Pass | Pass |
| TE registered CPU-memory TCP write/read | Pass | Pass |
| Real GPU-memory TCP write/read (1 MiB, RTX 5080) | N/A | Pass in all three states |

This is **150 pytest test executions** across six installation states, not 150 distinct tests. Store smoke tests disabled the same-host direct-copy shortcut with `MC_STORE_MEMCPY=0`.

Additional checks:
- Old/new/upgraded non-CUDA wheels passed real Store/TE smoke tests in a separate container without GPU, driver or CUDA libraries, after installing the required OS libraries. Neither non-CUDA wheel had CUDA DT_NEEDED dependencies.
- Public engine/store symbol and class-member sets, six console entry-point declarations, and effective base dependencies on Python 3.12 matched. Common ELF DT_NEEDED sets matched after normalizing ABI suffixes and auditwheel hashes; duplicate vendored libasio/libetcd_wrapper copies were eliminated.
- All four wheel RECORDs passed hash/size verification. Each CUDA wheel's EP/PG staged payloads were byte-identical to its own original build artifacts after repair. EP and PG 2.13 imports passed with Torch 2.13.0+cu130 and an explicitly supplied CUDA 12 runtime; this was not a distributed-kernel test.
- Missing OS/CUDA runtime dependencies and mixed CUDA/non-CUDA uninstall breakage reproduced on both revisions, so these are not new regressions.

**Compatibility observations:** Standard imports and the tested operations showed no regression. The switch to the root package removes the legacy `pkgutil.extend_path` behavior and changes generic native filenames: non-CUDA no longer includes `engine.so`, and neither variant includes `store.so` (ABI-tagged extensions are installed instead). Constructed split-namespace imports and hardcoded native-path loads therefore differ. No actual downstream dependency on these behaviors was identified; they are compatibility boundaries, not demonstrated failures of normal wheel usage.

**Limits:** These local wheels target manylinux_2_35, not the production manylinux_2_28 baseline. This A/B test does not cover CUDA 13, other Python versions, ARM64, RDMA/GPUDirect, multi-GPU/NVLink, distributed EP/PG kernels, the full Torch ABI matrix, or publication to PyPI. Local harnesses, artifacts and logs are retained under `build/pip-ab/` (not committed).

## Checklist

- [x] For changes >500 LOC: I have filed an RFC issue (existing RFC #3534)

## AI Assistance Disclosure

- [ ] No AI tools were used

OpenAI coding assistant in pi inspected the working-tree changes and related issues/PRs, ran checks, applied scoped formatting fixes, prepared the commit and PR, and performed the local before/after wheel builds and validation described above. Human line-by-line review and end-to-end ownership are still required; this PR is opened as a draft pending that review and full release-matrix validation.


Icedcoco:
Thanks for the update. I reviewed the latest CI results. The root backend migration and the CUDA, CUDA13, EFA, wheel installation, CTest, formatting, and static checks are now passing, so the main migration path appears to be working as intended.

I still have a few points I would like to clarify before merging:

1. ROCm builds are currently failing

   Both of the following jobs are red:
    - build-wheel-rocm (3.10)
    - build-wheel-rocm (3.12)

   Since ROCm is one of the release paths migrated by this PR, could we check the failure logs and confirm whether this is caused by the new backend/repair flow or by the runner environment? In either case, would it be possible to rerun both jobs after the issue is addressed, including a basic installation/import check for the generated wheels?

2. Wheel selection in repair_wheel.sh

   The repair script expects exactly one wheel in the output directory. The normal clean-runner paths are working, but could a retry or a reused workspace leave an older wheel behind and cause the repair step to fail? Would it be reasonable to add a small regression test for this case, or document the guarantee that all callers provide an empty output directory?

3. Build reproducibility

   build_release_wheel.py installs several build tools dynamically, while some backend dependencies remain open-ended. Is this intentional? If reproducible release builds are a goal, pinning these versions in the release environment or workflow might be worth considering.

These are mostly clarification and release-readiness points rather than concerns about the core design. Once the ROCm jobs are passing and the intended platform coverage is made explicit, I would be comfortable reassessing the PR for merge.

ascend-direct-dev (approved):

alogfans (approved):
LGTM

PR #2330, opened 2026-06-04
[TransferEngine] Fix: TENT TCP transport implicitly creates CUDA context on GPU 0

gogongxt:
## Problem

When using TENT's TCP transport to transfer GPU memory, processes that only use a specific GPU (e.g., GPU 7) or only DRAM would spuriously acquire a primary CUDA context on **GPU 0**, as observed via `nvidia-smi --query-compute-apps`.

This was reproduced using the script shared in PR #2307 comments by @chestnut-Q.

## Root Causes

Two categories of CUDA runtime calls silently trigger GPU 0 context creation on any thread with no prior `cudaSetDevice()`:

1. **`cudaPointerGetAttributes()`** — used in `cuda_free()`, `getMemoryType()`, `getLocation()`, and `cuda_query_location()` to classify host/device pointers. Calling it on a host pointer initializes the CUDA runtime with GPU 0 as the default device.

2. **`CudaPlatform::copy()`** — called `getStreamFromPool()` with `kCurrentDevice = -1`, which internally called `cudaGetDevice()` on a worker thread with no prior `cudaSetDevice()`, returning 0 and silently initializing GPU 0.

## Fix

- **New `cuda_utils.h`**: introduces `getCudaDeviceForPtr()` using the CUDA **driver API** (`cuPointerGetAttribute`). Unlike the runtime API, the driver API returns an error code for host pointers without initializing any context, making it safe to call from any thread.
- **`CudaPlatform::copy()`**: for host-to-host transfers (both pointers classify as non-device), falls back to `::memcpy` entirely — no CUDA involved. For GPU transfers, calls `cudaSetDevice(dev)` before any runtime API call to establish the correct device on the calling thread.
- **All `cudaPointerGetAttributes` call sites** in `cuda_allocator.cpp`, `cuda_probe.cpp`, and `cuda_plugin.cpp` are replaced with `getCudaDeviceForPtr()`.
- **`plugins/cuda/CMakeLists.txt`**: adds `CUDA::cuda_driver` to link the driver API.

## Verification

Tested with the two-process reproduction script (`test_tent_repro_2307.py`) on an 8×H200 machine:

**Before fix**: after 50× `transfer_sync_write` from initiator (GPU 7) to target (DRAM), both processes showed a compute context on GPU 0 in `nvidia-smi`.

**After fix**:
```
=== COMPUTE APPS ===
<initiator_pid>, GPU-...<GPU7_uuid>, 584 MiB   ← only GPU 7, correct
=== GPU MEM ===
[...]

PR #2307, opened 2026-06-03
[TransferEngine] Fix: TCP transport implicitly creates CUDA context on GPU0

gogongxt:
## Summary

When TCP transport transfers GPU memory, asio worker threads (which never call `cudaSetDevice`) execute `cudaMemcpy`, causing the CUDA Runtime to implicitly initialize a primary context on the default device (GPU0). This wastes ~520 MiB of GPU0 memory per worker process on H200 clusters.

## Root Cause

`isCudaMemory()` only checks whether an address is CUDA device memory but does not return the device ordinal. Without knowing which device owns the memory, the code cannot call `cudaSetDevice()` before `cudaMemcpy()`, so the CUDA Runtime defaults to GPU0.

## Fix

- Replace `isCudaMemory()` with `getCudaDevice()` which returns the CUDA device ordinal (or -1 for non-device memory)
- Call `cudaSetDevice(cuda_device)` before every `cudaMemcpy` so the Runtime uses the correct GPU
- All `is_cuda_memory` boolean checks replaced with `cuda_device >= 0`

## Verification

Verified on H200 cluster: TCP transfer +521 MiB on GPU0 before fix, +0 MiB after fix.

**before:**

```
Thu Jun  4 11:34:21 2026
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 570.148.08             Driver Version: 570.148.08     CUDA Version: 12.9     |
|-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA H200                    On  |   00000000:18:00.0 Off |                    0 |
| N/A   31C    P0            119W /  700W |  143054MiB / 143771MiB |    100%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+
|   1  NVIDIA H200                    On  |   00000000:2A:00.0 Off |                    0 |
| N/A   32C    P0            120W /  700W |  140122MiB / 143771MiB |     99%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+
|   2  NVIDIA H200                    On  |   00000000:3A:00.0 Off |                    0 |
[...]

PR #3973, opened 2026-09-09
[TENT] fix: TCP transport implicitly creates CUDA contexts

gogongxt:
## Description

TENT's TCP transport implicitly creates CUDA primary contexts in processes that never touched CUDA, or on GPUs the transfer never uses — several hundred MiB per process (~520 MiB on H200, ~414 MiB on A800). In per-rank deployments (`CUDA_VISIBLE_DEVICES=<rank>`), every rank's TCP traffic burns ~0.5 GB of GPU memory on its own card.

Three independent mechanisms, each verified on current `main` (2×A800, CUDA 12.9, TENT-enabled wheel, per-process nvidia-fd tracking at staged checkpoints):

1. **Host-to-host copies go through the CUDA stream pool.** `CudaPlatform::copy()` classifies both pointers as "current device", `CUDAStreamPool::acquire()` then calls `cudaGetDevice()` (device 0 on a fresh RPC worker thread) and `cudaStreamCreateWithFlags()` creates a primary context on GPU 0. This is reached on **every TCP WRITE** via the server-side `ControlService::onSendData()` (`control_plane.cpp:516` — no `MTYPE_CPU` fast path, unlike `sendData()`/`onRecvData()`) and on **every TCP READ** via `ControlClient::recvData()` (`control_plane.cpp:148`).
2. **Runtime-API pointer classification.** `cudaPointerGetAttributes()` on a thread with no prior `cudaSetDevice()` implicitly initializes the calling thread's current device (default GPU 0).
3. **`cudaMemcpyAsync()` validates the stream against the current device's context.** When a thread with no context (current device 0) copies through a stream belonging to another GPU, the runtime creates a primary context on GPU 0 — even though both buffer and stream live elsewhere. Confirmed with a minimal C-level repro (all other calls in the sequence are innocent).

### Fix

- New `tent/platform/cuda_utils.h`: `ensureCudaDriverInit()` (idempotent `cuInit`, creates no context — the driver API does not initialize itself lazily) and `getCudaDeviceForPtr()`, which classifies pointers via `cuPointerGetAttribute` and reads the driver's global pointer table without creating any context or implicitly initializing a device.
- `CudaPlatform::getMemoryType()` / `getPointerDeviceId()` / `getLocation()` / `free()` and the CUDA device plugin now classify pointers through the driver API.
- `CudaPlatform::copy()`: host-to-host copies use plain `::memcpy`; device-involved copies pin the buffer's device for the whole acquire/copy/synchronize sequence (restoring the caller's binding only when the thread already had a context), so a context is only ever created on the GPU that actually owns the buffer.

This supersedes #2330 (authored against a June tree; `copy()` has since been reworked by #3476 and the probe guards by #3261/#3945, so that diff no longer applies — and it did not cover mechanism 3 or the missing `cuInit`). It is the TENT counterpart of #2307, which fixed the classic (non-TENT) `TcpTransport`.

## Module

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

**Test commands:** two-process transfer over loopback TCP (`MC_USE_TENT=1`, tcp-only `MC_TENT_CONF`, GPU buffers allocated via ctypes `libcudart` so the test itself never pulls in torch), observing `nvidia-smi --query-compute-apps` plus per-process `/proc/<pid>/fd` nvidia-fd counts (driver init vs. context creation) at staged checkpoints; round-trip data-integrity checks in both directions. Also ran the exact repro script shared by @chestnut-Q in the #2307 discussion.

**Test results:** (2×A800-SXM4-80GB, CUDA 12.9, wheel built from this branch)

| Scenario (pure TCP, no torch) | Before | After |
|---|---|---|
| DRAM→DRAM WRITE | CPU-only **target** gets a 414 MiB context on GPU 0 | no context |
| DRAM→DRAM READ | CPU-only **initiator** gets a 414 MiB context on GPU 0 | no context |
| same + `CUDA_VISIBLE_DEVICES=1` | 414 MiB context on the rank's own GPU (the reported ~0.5 GB/card symptom) | no context; round-trip data verified |
| `cuda:1` → DRAM WRITE | initiator: contexts on GPU 0 **and** GPU 1; target polluted on GPU 0 | initiator: GPU 1 only (its own allocation); target clean; round-trip data verified |
| @chestnut-Q's repro script (#2307) | extra ~320 MiB on GPU 0 in the initiator (per their comment on #2330) | GPU 0 completely clean; only the initiator's own GPU 1 allocation remains; 50× WRITE completes |

- [ ] Unit tests pass (deferred to CI)

## Checklist

- [x] I have formatted my code using `./scripts/code_format.sh` (clang-format clean on all changed files)
- [ ] I have updated the documentation (if applicable) — N/A
- [ ] I have added tests to prove my changes are effective (manual repro matrix above; happy to add a unit test if maintainers prefer)
- [x] For changes >500 LOC: I have filed an RFC issue — N/A (~100 LOC)

## AI Assistance Disclosure

Claude Code assisted with root-cause analysis (fd-tracking repro, staged isolation, minimal C-level repro), implementing the fix, and running the verification matrix. The human submitter reviewed every changed line and can defend the change end-to-end.

### DIFF
--- a/mooncake-transfer-engine/tent/include/tent/platform/cuda_utils.h
+++ b/mooncake-transfer-engine/tent/include/tent/platform/cuda_utils.h
@@ -1,0 +1,53 @@
+// Copyright 2026 KVCache.AI
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef TENT_PLATFORM_CUDA_UTILS_H
+#define TENT_PLATFORM_CUDA_UTILS_H
+
+#include <cuda.h>
+
+#include <mutex>
+
+namespace mooncake {
+namespace tent {
+
+// cuInit() is idempotent and creates no context; the driver API does not
+// initialize itself lazily, so it must run before any driver query.
+inline bool ensureCudaDriverInit() {
+    static std::once_flag flag;
+    static CUresult result = CUDA_ERROR_NOT_INITIALIZED;
+    std::call_once(flag, []() { result = cuInit(0); });
+    return result == CUDA_SUCCESS;
+}
+
+// Device ordinal owning `ptr`, or -1 for host memory. Unlike the runtime
+// cudaPointerGetAttributes(), this never implicitly initializes the calling
+// thread's current device and its primary context (default GPU 0).
+inline int getCudaDeviceForPtr(const void* ptr) {
+    if (!ensureCudaDriverInit()) return -1;
+    unsigned int mem_type = 0;
+    CUresult res = cuPointerGetAttribute(
+        &mem_type, CU_POINTER_ATTRIBUTE_MEMORY_TYPE, (CUdeviceptr)ptr);
+    if (res != CUDA_SUCCESS || mem_type != CU_MEMORYTYPE_DEVICE) return -1;
+    unsigned int dev_ordinal = 0;
+    res = cuPointerGetAttribute(
+        &dev_ordinal, CU_POINTER_ATTRIBUTE_DEVICE_ORDINAL, (CUdeviceptr)ptr);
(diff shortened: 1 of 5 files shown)


alogfans:
Could you please check the CI failure since it seems related to your changes?

gogongxt:
Checked — the previous `build-wheel-rocm (3.12)` failure was unrelated to this change. I pushed an empty commit to re-trigger CI, and the new run passes all checks, including `build-wheel-rocm (3.12)`.

alogfans (approved):
LGTM

PR #3980, opened 2026-09-09
[TENT] Let the admission drop probe past a depressed bandwidth estimate

SongOf:
## Description

  Follow-up to #3780 (RFC #2519 step 3). The admission drop divides by the transmit estimate, and the estimate only learns from traffic that was dispatched, so the two form a loop. On a queue-depth-bound link an estimate depressed by a spell of contention admits too little to ever measure more: the admitted owners achieve exactly what the estimate says, and the owners that would prove it wrong are the ones dropped. The first commit pins this with a link model and with the real `DeviceSelector` meter — a 4-of-8 plateau that 60 rounds never leave, every drop after the contention lifted being a transfer the link could have carried; alpha does not matter, theta only moves the basin. The second adds opt-in exploration:  `runtime_queue/mlu_probe_owners` (default 0, behaviour unchanged) dispatches up to N owners per pick that the drop would reject, in EDF order and capped by `runtime_queue/mlu_probe_ceiling_factor` (default 2.0) so only owners misjudged by a depressed estimate are probed, not ones infeasible at line rate. One probe per pick leaves the plateau by round 16 and stops by itself once nothing is infeasible. Recommended value is 1; the probe's cost when the estimate is right is measured in the tests, and `tent_deadline_probe_total{outcome="met"|"missed"}` exports the verdict so a deployment can tell which case it is in.

  ## Module

  ## Type of Change

  - [x] New feature

  ## How Has This Been Tested?

  Hardware-independent. `admission_queue_test` drives the queue against a
  queue-depth-bound link model; `tent_admission_feedback_test` closes the same loop
  over the real `DeviceSelector` transmit meter; `tent_metrics_recording_test` covers
  the probe counter. Built with `-DTENT_METRICS_ENABLED=ON`.

  **Test commands:**
  ```bash
  cmake --build build-tent --target admission_queue_test tent_admission_feedback_test tent_metrics_recording_test
  ./admission_queue_test && ./tent_admission_feedback_test && ./tent_metrics_recording_test
  pre-commit run --files <files changed in this PR>

  Test results:
  - [x] Unit tests pass (55 / 8 / 26)
  
  Checklist                                                                                                                                                                                           

  - [ ] For changes >500 LOC: I have filed an RFC issue — the change is RFC #2519 step 3; ~150 LOC of production code, the rest is tests

  AI Assistance Disclosure

### DIFF
--- a/mooncake-transfer-engine/tent/include/tent/metrics/tent_metrics.h
+++ b/mooncake-transfer-engine/tent/include/tent/metrics/tent_metrics.h
@@ -118,6 +118,13 @@ class TentMetrics {
     // distinguishable from genuine MLU samples in the histogram above.
     void recordDeadlineInfeasible(TransportType tp);
 
+    // Record how an owner the admission queue dispatched as a probe -- past
+    // a drop predicate that called it infeasible (runtime_queue/
+    // mlu_probe_owners) -- ended: met its deadline, or missed it. The ratio
+    // says whether the drop threshold is over-conservative (probes mostly
+    // meet) or the link is genuinely full (probes mostly miss).
+    void recordDeadlineProbe(bool met_deadline);
+
     // Record a batch abandoned by lazyFreeBatch after repeated failed reclaim
     // attempts. Such a batch stays on the freelist without further retries and
     // is only reclaimed at engine teardown, so a nonzero value means resources
@@ -226,6 +233,8 @@ class TentMetrics {
                                                                   "operation"};
     static inline const std::array<std::string, 2> kTaskFailureLabels{
         "transport", "reason"};
+    static inline const std::array<std::string, 1> kOutcomeLabel{"outcome"};
+    static constexpr size_t kOutcomeDomain = 2;  // met, missed
 
     // Hot-path metrics record through pre-resolved label cells (see
     // cached_metric.h). Label domain sizes: TransportType has
@@ -290,6 +299,11 @@ class TentMetrics {
         "tent_deadline_infeasible_total",
         "Transfers whose deadline was already in the past at submit",
         kTransportLabel, kTransportDomain};
+    metrics::CachedDynamicCounter<1> deadline_probe_total_{
+        "tent_deadline_probe_total",
+        "Owners the admission drop dispatched as probes, by whether they met "
+        "their deadline",
+        kOutcomeLabel, kOutcomeDomain};
     // Label-less: quarantine is a batch-lifecycle event, not a per-transport
     // one (a batch may hold SubBatches on several transports).
     ylt::metric::counter_t quarantined_batches_total_{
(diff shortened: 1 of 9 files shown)


SongOf:
Closing this. Measured on real hardware (two-node tebench), and the premise does not hold — the plateau I built this on is an artifact of my test harness's link model, not something the shipped meter produces.

**What the run showed.** Tightening `mlu_local_threshold` cuts dispatch and *raises* the on-time rate, and dispatch per round is a straight line in theta:

| theta | dispatched / 8 | on time | dispatched / theta |
|---|---|---|---|
| 1.0 | 4.5 | 93% | 4.5 |
| 0.75 | 3.5 | 98% | 4.67 |
| 0.5 | 2.3 | 100% (35/35) | 4.6 |
| 0.3 | 1.5 | 100% (23/23) | 5.0 |

That last column holding at 4.5–5.0 across four settings is the signature of a **static cumulative-MLU cutoff**, not a feedback loop: within one pick the i-th owner is judged behind the i−1 before it, so MLU rises linearly with position and the admitted count is theta divided by the per-owner increment (~0.22 here). Had the estimate been depressed, that increment would have grown and the ratio would have drifted. It did not — the estimate was stable and correct in all four runs. Dropping also *lowers* `dispatching_bytes_` and therefore lowers MLU for the survivors, which is negative feedback. The loop this PR is named for cannot close.

**The 95% probe hit rate at theta 0.5 is arithmetic, not discovery.** MLU is the fraction of its window a transfer needs, so MLU < 1 means it fits. theta = 0.5 rejects everything needing more than half its window — owners with MLU in [0.5, 1.0), all of which the predicate itself expects to finish in time. Dispatching them and watching them succeed confirms the definition. At theta = 1.0 the probe met 1 of 40 (2.5%), which is exactly what `ProbeOnACorrectEstimateIsPaidEveryRound` and `AtLineRateOverloadTheProbeMeasuresNothing` in this PR predicted: against an accurate predicate the probe is pure cost.

**Where my model went wrong.** `admission_feedback_test.cpp`'s `serve(n, concurrent)` charges the whole span as busy time and makes the rate *while busy* fall with concurrency (`line × concurrent / 8`). The shipped meter does the opposite on purpose — #3780 made it charge bytes to busy time precisely so that low occupancy reads as less busy time at the same rate, not as a slow link. My harness reintroduced, in its link model, the bias that change removed from the meter. It is also wrong at the sizes tested: a couple of 2 MB work requests saturate a modern RNIC, so "4 in flight achieve half line rate" does not hold. Queue-depth dependence is a small-message phenomenon.

**What the run did turn up, and what I would rather do instead.** `mlu_local_threshold` below 1.0 is a misconfiguration that drops transfers the predicate itself says will succeed, and nothing warns about it. A range check plus a WARNING is a much smaller change than this PR and addresses the failure actually observed. Separately, the met/missed ratio was a genuinely good diagnostic — 95% versus 2.5% separated "theta too strict" from "predicate accurate" cleanly — but to keep it without the probe it would have to judge *dropped* owners in the shadow rather than dispatched probes. I will open those separately if they look worth it.

**Not settled:** one host pair, one workload shape, large blocks. The small-message / high-concurrency regime where queue depth genuinely bounds throughput was not exercised, nor the path where `resetTransmitMeter` rebuilds the estimate from very short intervals after flushed work requests. If the plateau exists anywhere, it is there, and I would reopen this with the automatic back-off (stop after k consecutive misses) so the probe is inert at theta >= 1.

Thanks for pushing on the performance question — that is what got this measured.

PR #3981, opened 2026-09-09
[Bugfix][TENT] Plan RDMA slices from the block size: no empty slices, and fold short tails

SongOf:
## Description

Two defects in how `submitTransferTasks` cuts a request into slices. Both come from the same place: the slice count and the block size were decided separately, in the wrong order.

**Zero-length slices.** The count came from `ceil(length / block_size)` capped at `max_slice_count` (32 for writes and CUDA sources, 64 otherwise), the block from `roundup(ceil(length / num_slices), block_size)`. The loop then ran the full count with `min(remaining, block_size)` per slice, so whenever the round up actually rounded, the surplus came out as zero-length slices. A 10 MB read was cut into 192 KB blocks, 54 of which cover it, and the other 10 were empty. It needs the cap to bite (over 2 MB for writes and CUDA, over 4 MB otherwise) or the tail merge to fire, and how much is wasted depends on whether `ceil(length / cap)` is already block-aligned, so it jumps around with request size: 8 of 32 empty for a 3 MB write, none for 6 MB. Not a correctness bug — a zero-length RDMA op is legal and the task accounting is consistent — but each empty slice still takes a work request, a CQE of the CQ quota, a full `generatePostPath` device selection and a completion settlement.

**A tail merge that widens the block instead of folding the tail.** The merge meant to keep a short last slice from being posted on its own, and did it by asking for one slice fewer. The block is chosen after, so one slice fewer only rounds the block up to the next whole one: the short tail comes back as its own slice regardless, and the wider block spreads the request over half as many slices. Its `tail` was `length % block_size`, which once the cap has widened the block is not the remainder of anything, so it could also fire for no reason. For 520 KB at a 64 KB block — the shape the merge exists for — it produced a 128 KB block, 5 slices, and the 8 KB tail still on its own.

**Fix.** `planRdmaSlices()` in `slice.h` computes the block as before and then takes the count *from* the block, so no slice is empty; since `block_size >= ceil(length / requested)` the count never exceeds the cap. The fold happens after that: when the last slice would be shorter than `merge_ratio` of a block, the count drops by one and the last slice takes what remains, one block plus the tail. The loop gives the last slice the remainder and every other slice exactly one block. `merge_ratio` is a defaulted parameter (0.25 as before, 0 never folds) rather than a constant at the call site, which loses its pre-merge block entirely and passes `max_slice_count` straight through. 520 KB now plans a 64 KB block, 8 slices, last one 72 KB. Extracted as a pure function because the arithmetic had no test coverage and `submitTransferTasks` cannot be driven without an RNIC.

**The accounting the fold breaks.** The last slice can now exceed `block_size` by up to one block, and `DeviceSelector::selectMultiPath()` reconstructed each slice's length as `min(slice_bytes, remaining)` to charge `inflight_bytes`. So the fold was charged one block but released its real length, and `inflight_bytes` is unsigned: it drifts down by the folded tail per transfer and wraps to ~2^64, which then reads as an impossibly loaded NIC in device scoring and in the deadline arbitration's queue-ahead term. Both charging walks now give the last slice what remains, and `allocate()`'s declaration states the partition so the rule is written where the two sides can be held to it. Nothing downstream constrains a slice to the block — it becomes one work request with one SGE.

## Module

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

Ten `RdmaSlicePlanTest` cases in `rdma_transport_test.cpp` and two in `quota_accounting_test.cpp`, all hardware-free. The plan: the contract over a sweep of lengths against both caps (the slices cover the request exactly, none is empty, the cap holds, the block is a whole number of configured blocks, every slice but the last is exactly a block), the block staying unchanged below the cap, the regressions above, the cap bounding the count while the block grows, a short tail joining the slice before it without widening the block, a tail worth its own slice keeping one, a zero ratio never folding, folding never dropping the last slice, and degenerate inputs.

The accounting: a last slice longer than a block is charged its real length, and real `planRdmaSlices()` plans walked through `allocate()` and `release()` bring the inflight bytes back to zero over a sweep of lengths against both caps.

Red-checked four ways: restoring the old count fails the sweep with "empty slice"; removing the fold fails four cases; restoring the fold-before-the-block form reproduces exactly the 128 KB block and 5 slices described above; and restoring the `min(slice_bytes, remaining)` charge leaves 18446744073709551516 inflight bytes.

**Test commands:**
```bash
cmake --build build-tent --target tent_rdma_transport_test
./tent_rdma_transport_test
ctest   # in build-tent
pre-commit run --files <files changed in this PR>
```

**Test results:**
- [x] Unit tests pass (`tent_rdma_transport_test` 76/76, `tent_quota_accounting_test` 9/9; full ctest green apart from four failures that reproduce on an unmodified `main`: `tcp_transport_test`, `transfer_metadata_test`, `memory_location_test`, `transfer_engine_config_override_test`)

## Checklist

## AI Assistance Disclosure

Claude Code helped analyse the slicing arithmetic, write the fix and the tests. Every changed line has been reviewed by the submitter.

### DIFF
--- a/mooncake-transfer-engine/tent/include/tent/transport/rdma/quota.h
+++ b/mooncake-transfer-engine/tent/include/tent/transport/rdma/quota.h
@@ -188,7 +188,15 @@ class DeviceSelector {
 
     // Allocate devices for a request (new API)
     // slice_bytes: pre-calculated slice size from rdma_transport to ensure
-    // consistency
+    // consistency.
+    //
+    // The slices partition the request: the first num_slices - 1 carry
+    // slice_bytes each and the last carries what is left, which is a block
+    // plus a folded-in tail when planRdmaSlices() folded one -- so it can be
+    // longer than slice_bytes. Each device is charged what its slices
+    // actually carry, so release(), which returns a slice's real length,
+    // balances. Charging every slice slice_bytes would leave the folded tail
+    // released but never charged, and inflight_bytes is unsigned.
     Status allocate(uint64_t total_length, uint32_t num_slices,
                     uint64_t slice_bytes, const std::string &location,
                     std::vector<int> &slice_dev_ids, int priority = PRIO_HIGH,
--- a/mooncake-transfer-engine/tent/include/tent/transport/rdma/slice.h
+++ b/mooncake-transfer-engine/tent/include/tent/transport/rdma/slice.h
@@ -41,6 +41,36 @@ struct RdmaSliceList {
     int num_slices = 0;
 };
 
+// `count` slices of `block_size` bytes, the last holding what remains.
+struct RdmaSlicePlan {
+    uint64_t block_size = 0;
+    uint64_t count = 0;
+};
+
+// Cut `length` into at most `max_slices` slices, each a whole number of
+// `base_block` bytes. Rounding the block up covers the request in fewer
+// slices than were asked for, so the count comes from the block and not the
+// other way round: an empty slice would still cost a work request, a CQE, a
+// path selection and a completion. Zero length keeps one slice -- task
+// accounting counts slices, and a task with none never reaches a terminal
+// status.
+inline RdmaSlicePlan planRdmaSlices(uint64_t length, uint64_t base_block,
+                                    uint64_t max_slices) {
+    if (base_block == 0) base_block = 1;
+    if (max_slices == 0) max_slices = 1;
+    if (length == 0) return {base_block, 1};
+
+    uint64_t count = (length + base_block - 1) / base_block;
+    if (count > max_slices) count = max_slices;
+
+    const uint64_t per_slice = (length + count - 1) / count;
+    const uint64_t block_size = (per_slice % base_block == 0)
+                                    ? per_slice
+                                    : (per_slice / base_block + 1) * base_block;
+
+    return {block_size, (length + block_size - 1) / block_size};
+}
+
 // Forward declarations
 class RdmaEndPoint;
 struct RdmaTask;
--- a/mooncake-transfer-engine/tent/include/tent/transport/rdma/slice.h
+++ b/mooncake-transfer-engine/tent/include/tent/transport/rdma/slice.h
@@ -41,7 +41,10 @@ struct RdmaSliceList {
     int num_slices = 0;
 };
 
-// `count` slices of `block_size` bytes, the last holding what remains.
+// `count` slices: the first count - 1 of `block_size` bytes and the last
+// holding what remains, which is a short tail folded in when one was worth
+// folding -- so up to `block_size` more than a block, never less than one
+// byte.
 struct RdmaSlicePlan {
     uint64_t block_size = 0;
     uint64_t count = 0;
@@ -51,11 +54,21 @@ struct RdmaSlicePlan {
 // `base_block` bytes. Rounding the block up covers the request in fewer
 // slices than were asked for, so the count comes from the block and not the
 // other way round: an empty slice would still cost a work request, a CQE, a
-// path selection and a completion. Zero length keeps one slice -- task
-// accounting counts slices, and a task with none never reaches a terminal
-// status.
+// path selection and a completion.
+//
+// A last slice shorter than `merge_ratio` of a block is given to the slice
+// before it instead of being posted on its own, trading a fraction of a
+// block of extra work on that one slice for a work request, a CQE and a
+// completion. This has to happen after the block is chosen: asking for one
+// slice fewer up front only makes the block round up to the next whole one,
+// which leaves the short tail as its own slice anyway and spreads the
+// request over half as many slices. A ratio of 0 never folds.
+//
+// Zero length keeps one slice -- task accounting counts slices, and a task
+// with none never reaches a terminal status.
 inline RdmaSlicePlan planRdmaSlices(uint64_t length, uint64_t base_block,
-                                    uint64_t max_slices) {
+                                    uint64_t max_slices,
+                                    double merge_ratio = 0.25) {
     if (base_block == 0) base_block = 1;
     if (max_slices == 0) max_slices = 1;
     if (length == 0) return {base_block, 1};
@@ -68,7 +81,12 @@ inline RdmaSlicePlan planRdmaSlices(uint64_t length, uint64_t base_block,
                                     ? per_slice
                                     : (per_slice / base_block + 1) * base_block;
 
-    return {block_size, (length + block_size - 1) / block_size};
+    count = (length + block_size - 1) / block_size;
+    if (count > 1) {
+        const uint64_t tail = length - (count - 1) * block_size;
+        if (static_cast<double>(tail) < merge_ratio * block_size) --count;
+    }
+    return {block_size, count};
 }
 
 // Forward declarations
--- a/mooncake-transfer-engine/tent/src/transport/rdma/quota.cpp
+++ b/mooncake-transfer-engine/tent/src/transport/rdma/quota.cpp
@@ -142,8 +142,11 @@ Status DeviceSelector::allocate(uint64_t total_length, uint32_t num_slices,
                 int dev_id = tl_eligible[tl_rr_counter % tl_eligible.size()];
                 tl_rr_counter++;
                 slice_dev_ids.push_back(dev_id);
+                // Last slice takes what is left; see the header.
                 uint64_t this_slice_bytes =
-                    std::min(slice_bytes, total_length - offset);
+                    (i + 1 == num_slices)
+                        ? total_length - offset
+                        : std::min(slice_bytes, total_length - offset);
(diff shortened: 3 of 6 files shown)


alogfans (commented):
Please fix the accounting mismatch caused by the folded tail: `planRdmaSlices()` can make the last slice larger than `block_size`, while `DeviceSelector::selectMultiPath()` still charges only `slice_bytes`, causing `allocate()` and `release()` to become unbalanced.

SongOf:
@alogfans Confirmed, and worse than an under-report: `inflight_bytes` is unsigned and `releaseInflight()` is an unguarded `fetch_sub`, so the folded tail is released but never charged and the counter wraps to ~2^64 — which then reads as an impossibly loaded NIC in device scoring and in the arbitration's queue-ahead term.

Fixed in b6ddad48: both charging walks give the last slice what remains, and `allocate()`'s declaration now states the partition instead of leaving it implicit in two loops. Tests in `quota_accounting_test.cpp` cover a long last slice and real `planRdmaSlices()` plans walked through allocate/release; restoring the old charge leaves 18446744073709551516 inflight.

alogfans (approved):
LGTM

alogfans:
Additionally, do you have any performance comparision results?

SongOf:
No — I have not run a performance comparison, and I should say so plainly rather than imply the change is free. Everything behind this PR is arithmetic reasoning plus hardware-free unit tests; I do not have an RNIC pair to measure on.

What I can say about the expected direction, so the ask below is concrete:

- **Commit 1 (empty slices)** should be a strict win and the easiest to see. The data slicing is bit-for-bit what it was; only slices carrying no bytes disappear. A 10 MB read goes from 64 work requests to 54, a 3 MB write from 32 to 24. Each one removed was still taking a WR slot, a CQE of the CQ quota, a full `generatePostPath` device selection and a completion settlement, so I would expect a small drop in per-request CPU and CQ pressure with bandwidth unchanged.
- **Commit 2 (tail fold)** is the one that could go either way, and the one I would most want measured. For a request with a short tail the block no longer doubles: 520 KB at a 64 KB configured block used to plan a 128 KB block over 5 slices carrying data, and now plans a 64 KB block over 8. More slices and smaller blocks means more spread across NICs and queue pairs but more work requests, so the sign depends on where the bottleneck is.
- **Commit 3 (charging)** does not touch the data path, but it does correct `inflight_bytes`, which feeds device scoring — so selection behaviour changes, and on a multi-NIC host the distribution may shift.

Advice I would appreciate:

1. Which `tebench` configuration would you consider representative here? The interesting sizes are the ones where the cap bites or a short tail exists — with the shipped 64 KB block that is roughly 2 MB and up for writes, 4 MB and up otherwise — so something like `--start_block_size`/`--max_block_size` spanning 1 MB to 32 MB rather than the default 4 KB start. Is there a canonical sweep the project uses for transfer-engine changes?
2. Is bandwidth the metric you want, or would per-request CPU / completions-per-byte show commit 1's effect more clearly? Bandwidth on a saturated link may not move at all even though the overhead is real.
3. Is there a machine or CI lane with an RNIC pair I can run this on? If not, I would rather land commits 1 and 3, which have no plausible downside, and hold commit 2 until the fold can be measured — happy to split the PR that way if you prefer.

If none of that is available I will state the absence of measurements in the PR description instead of leaving it implied.

SongOf:
@alogfans 
I now have an RNIC pair, so here are the measurements I owed this PR.

**TL;DR** — the saved work requests are real on hardware (a 5 MB READ plans 64 slices on `main` vs 40 here). Throughput is unchanged when the link saturates, and converts into up to **+34.7%** throughput when work-request submission — rather than link bandwidth — is the limiting factor. Controls where the plan is unchanged stay within ±1%.

## Test environment

| | |
|---|---|
| Nodes | 2× Alibaba Cloud `ecs.gn8is-2x.8xlarge` (32 vCPU, 256 GiB) |
| NIC | eRDMA `erdma_0`, `PORT_ACTIVE`, RoCE; instance baseline bandwidth **32 Gbit/s** |
| Path | DRAM ↔ DRAM, `-protocol=rdma`, `-backend=tent`, `-metadata_server=P2PHANDSHAKE` |
| Build | `-DUSE_TENT=ON -DCMAKE_BUILD_TYPE=Release`, identical flags for both builds |
| Baseline | `9cfb3381` (this PR's merge base) vs all three commits |
| Tool | `transfer_engine_bench`, `-threads=4 -batch_size=16 -duration=8`, 4–6 reps per point, base/PR interleaved |

Two sizes act as **controls**: at 4 MB and 8 MB the old and new planners produce an identical plan (64 slices), so any movement there is noise, not the patch.

## Why a CPU-bound configuration was needed

At full speed both builds sit at **4.14–4.15 GB/s** — that is the 32 Gbit/s instance cap, and even `-threads=1 -batch_size=1` reaches it. With the link saturated, removing work requests has nowhere to show up. To measure what this PR actually changes I constrained the initiator's CPU (`systemd-run --scope -p CPUQuota=N%`) until throughput fell well below the cap:

| CPUQuota | throughput |
|---|---|
| unconstrained | 4.14 GB/s (link-bound) |
| 100% | 2.65 GB/s |
| 50% | 1.71 GB/s |
| 30% | 1.10 GB/s |

Throughput tracks the quota almost linearly, confirming CPU — not the link — is the limiter in that regime.

## Results

### 1. Slice plans, measured on hardware

Both builds instrumented to print the plan (counting only; no logic touched). Same request, 5 MB READ:

```
main:     len=5242880 block=131072 num_slices=64
this PR:  len=5242880 block=131072 num_slices=40
```

64 → 40, exactly the −37.5% the arithmetic predicts. An exhaustive sweep of the planner (4 KB…64 MB in 4 KB steps, both slice caps, 32768 cases) gives: **63.5% of cases on `main` produce at least one zero-length slice, 0% here**; byte totals are conserved in both.

### 2. Link-saturated — no change, no regression

48 runs, READ and WRITE, 4/5/9 MB:

| op | size | predicted WR saving | main | PR | Δ |
|---|---|---|---|---|---|
| read | 4 MB | 0% *(control)* | 4.14 | 4.14 | +0.0% |
| read | 5 MB | 37.5% | 4.14 | 4.14 | +0.0% |
| read | 9 MB | 25.0% | 4.14 | 4.14 | +0.1% |
| write | 4 MB | 0% *(control)* | 4.15 | 4.14 | −0.1% |
| write | 5 MB | 15.6% | 4.14 | 4.15 | +0.2% |
| write | 9 MB | 9.4% | 4.14 | 4.14 | +0.0% |

Every run landed on 4.14 or 4.15 GB/s. Consistent with the "bandwidth unchanged" expectation in the description.

### 3. CPU-bound — the saved work requests convert into throughput

`CPUQuota=50%`, READ, 4 reps each:

| size | predicted WR saving | main | PR | measured gain | gain / prediction |
|---|---|---|---|---|---|
| **4 MB** | 0% *(control)* | 1.485 | 1.490 | **+0.3%** | — |
| **8 MB** | 0% *(control)* | 2.435 | 2.410 | **−1.0%** | — |
| 5 MB | 37.5% | 1.760 | 2.370 | **+34.7%** | 0.92× |
| 6 MB | 25.0% | 1.990 | 2.320 | +16.6% | 0.66× |
| 9 MB | 25.0% | 2.650 | 3.380 | **+27.5%** | 1.10× |
| 10 MB | 15.6% | 2.900 | 3.340 | **+15.2%** | 0.97× |
| 11 MB | 7.8% | 3.160 | 3.370 | **+6.6%** | 0.85× |

Gains track the predicted work-request reduction across sizes (Pearson r = 0.933, n = 5), while both controls stay within ±1%.

Tightening the constraint keeps the effect and separates the distributions completely — at `CPUQuota=30%`, 5 MB gives 1.115 → 1.325 (**+18.8%**) with observed ranges 1.06–1.17 vs 1.25–1.38 (no overlap), against −1.6% on the 4 MB control whose ranges do overlap.

## Conclusion

- **No regression** anywhere measured.
- The work-request reduction is **real on hardware** and, where WR submission/completion is the bottleneck, converts into throughput at roughly the predicted ratio.
- Whether a given deployment sees the gain depends on where its bottleneck sits: link-bound traffic sees nothing, CPU/CQ-bound traffic sees it directly.

## Caveats

1. These numbers are for **all three commits together**; commit 2 was not measured in isolation. Arithmetically it is a constant −1 WR relative to commit 1, changing the plan in ~22.7% of READ cases (20.6% for WRITE), at the cost of a last slice of up to 1.25 blocks — so it trades one work request for a slightly less even spread, and the numbers above already include that trade.
2. The 6 MB point (0.66× of prediction) sits off the trend the other four follow (0.85–1.10×); I have left it in rather than drop it, but it is worth a re-run.
3. `tebench` defaults to `-backend=classic`, which does not exercise this code at all — an earlier batch of my runs was wasted on it. Everything above uses `-backend=tent`.

PR #3984, opened 2026-09-09
[Store] Support online DFS shard capacity expansion

zupengwang:
## Description

The default DFS shard allocator fixes capacity when the Master starts. Increasing that capacity currently requires a restart, which loses the in-memory cache metadata. This change lets an operator append shard files through the existing Master HTTP admin service while upgraded clients keep using existing descriptors.

Related RFC: [#3812 — DFS replica dynamic capacity scaling](https://github.com/kvcache-ai/Mooncake/issues/3812), including the [maintainer's lifecycle and compatibility requirements](https://github.com/kvcache-ai/Mooncake/issues/3812#issuecomment-5490632735).

The admin interface is:

```http
GET /api/v1/dfs/shard_count
PUT /api/v1/dfs/shard_count
Content-Type: application/json

{"shard_count":128}
```

The PUT value is the desired total count. Repeating the current count succeeds; shrinking and invalid values return HTTP 400. Disabled DFS or an already active expansion returns HTTP 409; an unavailable or standby service returns HTTP 503. Filesystem preparation runs outside the HTTP I/O threads. A work guard retains the connection executor until the accepted operation completes, including during Stop; completion returns to that executor before responding.

The implementation addresses the RFC requirements as follows:

- Prepare all new files and allocator states, then atomically publish one immutable topology snapshot. Existing shard states remain shared by allocation, free, access tracking, and prepared eviction transactions. Failed preparation leaves the published count unchanged and rollback only removes files created by that request.
- Make new filenames independent of the total count, preserving the exact paths of discovered legacy padded files.
- Start allocation at the hash-selected shard and try the other ready shards when it has no suitable extent, so empty appended shards provide usable capacity.
- Open client shard handles only when consuming a published descriptor. Validate the path, index, and capacity before caching the handle; client initialization does not create or open shard files.
- Discover the contiguous existing shard layout on Master startup and retain at least the configured initial capacity. Reject ambiguous indices, gaps, and capacity mismatches.

### Compatibility and limits

Upgrade the Master and every DFS client before expanding. Older clients can reject descriptors for added shards. All processes must use the same shared root, filesystem adapter, shard capacity, and alignment. Only one active Master may manage a root, and shard files must not be created, renamed, truncated, or deleted externally.

Expansion only adds shards. It does not resize individual files, move existing objects, change the root, or stripe an object across files. Operators must provision the backing filesystem space.

Startup discovery preserves the capacity layout only. It does not recover keys or allocation ownership, add snapshot/OpLog/HA support, or provide a new data durability guarantee. Existing single-tenant and DFS replica placement limits remain.

[#3902](https://github.com/kvcache-ai/Mooncake/pull/3902) adds a separate immutable bucket allocator and an online bucket-count limit; it preserves default shard behavior. [#3849](https://github.com/kvcache-ai/Mooncake/pull/3849) adds configured multiple roots. This change addresses online growth of the default fixed-shard allocator.

## Module

- [ ] CI/CD
- [x] Docs
- [ ] Other

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Breaking change
- [x] Documentation update
- [ ] Performance improvement
- [ ] Other

## How Has This Been Tested?

Validated on a remote Linux node with GCC 11.5, a Release build, native TCP transport and the POSIX filesystem adapter. The final patch is based on `b9059252aa8db150ada8220241707d47e27ebb79`.

| Target | Result |
| --- | --- |
| `dfs_posix_test` | 43 passed |
| `dfs_sync_client_test` | 8 passed; checksum-only case skipped in the default configuration |
| `dfs_batch_read_checksum_test` | 1 passed with `MOONCAKE_STORE_CHECKSUM=1`, covering that skipped case |
| `master_admin_server_test` | 59 passed |
| `master_service_dfs_scenario_test` | 6 passed |

All five CTest entries passed, covering 117 distinct passing test cases. Coverage includes an already initialized client writing and reading a newly added shard, old data remaining readable, failed preparation and retry, legacy names, concurrent allocation/eviction and lazy opening, request validation, HTTP responsiveness during queued expansion, and Stop draining the accepted operation.

The baseline at `1caf8f8f9e800886debf9013a22b3403faf4ecfb` passed its 27 existing POSIX tests. The added baseline-compatible `DfsBackendTest.OpensNewShardDescriptorAfterClientInitialization` fails on that baseline with `Invalid DFS shard_idx 4`. That is a separate earlier baseline; it is not claimed as an A/B run against the newer final base.

**Test commands:**

```bash
cmake --build build --target dfs_posix_test dfs_sync_client_test \
  master_admin_server_test master_service_dfs_scenario_test -j4
ctest --test-dir build -V -j1 --timeout 180 \
  -R '^(dfs_posix_test|dfs_sync_client_test|dfs_batch_read_checksum_test|master_admin_server_test|master_service_dfs_scenario_test)$'
git diff --cached --name-only -z | xargs -0 pre-commit run --files
(cd docs && make html)
```

Build options included `USE_CUDA=OFF`, `USE_TCP=ON`, `USE_ETCD=OFF`, `USE_HTTP=ON`, `USE_3FS=OFF`, `WITH_STORE=ON`, `WITH_STORE_RUST=OFF`, and `BUILD_UNIT_TESTS=ON`. Changed-line clang-format 20.1.8 and all applicable pre-commit hooks passed. Documentation built successfully, and the generated sections were inspected.

No GPU/RDMA, live HF3FS deployment, full-library sanitizer, or performance result is claimed.

**Test results:**

- [x] Unit tests pass
- [x] Native TCP Master/client and POSIX integration tests pass
- [ ] Manual deployment testing done

## Checklist

- [x] For changes >500 LOC: covered by existing RFC #3812

Existing design RFC: [#3812](https://github.com/kvcache-ai/Mooncake/issues/3812), opened by @sun-baoshan-sunshine. The submitter has reviewed the final diff.

## AI Assistance Disclosure

- [ ] No AI tools were used

Codex assisted with upstream screening, implementation, tests, documentation, and independent static review. The submitter has reviewed the final diff and is responsible for understanding and defending the changes.

### DIFF
--- a/mooncake-store/src/storage/distributed/distributed_storage_backend.cpp
+++ b/mooncake-store/src/storage/distributed/distributed_storage_backend.cpp
@@ -79,7 +79,7 @@ DistributedStorageBackend::DistributedStorageBackend(
 }
 
 DistributedStorageBackend::~DistributedStorageBackend() {
-    for (auto& [index, shard] : shard_files_) {
+    for (auto& [_, shard] : shard_files_) {
         if (shard && shard->fd >= 0 && fs_adapter_) {
             fs_adapter_->CloseFile(shard->fd);
             shard->fd = -1;
@@ -136,18 +136,33 @@ DistributedStorageBackend::GetOrOpenShard(
         return tl::make_unexpected(ErrorCode::INVALID_PARAMS);
     }
 
-    std::lock_guard cache_lock(shard_files_mutex_);
-    auto existing = shard_files_.find(descriptor.shard_idx);
-    if (existing != shard_files_.end()) {
-        if (existing->second->path != path.string()) {
+    ShardFile* shard;
+    {
+        std::lock_guard cache_lock(shard_files_mutex_);
+        auto existing = shard_files_.find(descriptor.shard_idx);
+        if (existing == shard_files_.end()) {
+            existing = shard_files_
+                           .emplace(descriptor.shard_idx,
+                                    std::make_unique<ShardFile>())
+                           .first;
+        }
+        shard = existing->second.get();
+    }
+
+    // File validation and opening may block. Serialize them only with this
+    // shard's initialization and I/O, leaving other cached shards available.
+    std::lock_guard shard_lock(shard->mutex);
+    auto file_path = path.string();
+    if (shard->fd >= 0) {
+        if (shard->path != file_path) {
             return tl::make_unexpected(ErrorCode::INVALID_PARAMS);
         }
-        return existing->second.get();
+        return shard;
     }
 
     // The master prepares new shards before publishing their descriptors.
     // OpenFile may create a file, so reject missing/unprepared shards first.
-    auto file_size = fs_adapter_->GetFileSize(path.string());
+    auto file_size = fs_adapter_->GetFileSize(file_path);
     if (!file_size) return tl::make_unexpected(file_size.error());
     if (*file_size != distributed_config_.shard_capacity) {
         return tl::make_unexpected(ErrorCode::INVALID_PARAMS);
@@ -161,14 +176,13 @@ DistributedStorageBackend::GetOrOpenShard(
         canonical_path.filename() != path.filename()) {
         return tl::make_unexpected(ErrorCode::INVALID_PARAMS);
     }
-    auto fd = fs_adapter_->OpenFile(path.string());
+    auto fd = fs_adapter_->OpenFile(file_path);
     if (!fd) return tl::make_unexpected(fd.error());
-    auto shard = std::make_unique<ShardFile>();
-    shard->path = path.string();
+    // Failed attempts keep fd == -1 so a later request can retry
+    // initialization.
+    shard->path = std::move(file_path);
     shard->fd = *fd;
-    auto* result = shard.get();
-    shard_files_.emplace(descriptor.shard_idx, std::move(shard));
-    return result;
+    return shard;
 }
 
 tl::expected<int64_t, ErrorCode> DistributedStorageBackend::BatchOffload(
--- a/mooncake-store/src/storage/distributed/distributed_storage_backend.cpp
+++ b/mooncake-store/src/storage/distributed/distributed_storage_backend.cpp
@@ -3,13 +3,35 @@
 #include <algorithm>
 #include <filesystem>
 #include <limits>
+#include <optional>
+#include <string_view>
 
-#include "storage/distributed/dfs_global_allocator.h"
 #include "types.h"
 
 namespace mooncake {
 
 namespace {
+
+std::optional<int> ParseShardFileName(std::string_view name) {
+    constexpr std::string_view prefix = "dfs_shard_";
+    constexpr std::string_view suffix = ".data";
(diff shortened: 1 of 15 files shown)


mooncake-store/src/storage/distributed/distributed_storage_backend.cpp
```diff
@@ -92,26 +114,61 @@ tl::expected<void, ErrorCode> DistributedStorageBackend::Init() {
...
+    if (root.filename().empty()) root = root.parent_path();
+    auto parent = path.parent_path();
+    if (parent.empty()) parent = ".";
+    const auto index = ParseShardFileName(path.filename().string());
+    if (descriptor.shard_idx < 0 || !index || *index != descriptor.shard_idx ||
+        parent != root) {
+        return tl::make_unexpected(ErrorCode::INVALID_PARAMS);
+    }
+
+    std::lock_guard cache_lock(shard_files_mutex_);
```
fcczzz: Could we avoid holding shard_files_mutex_ across GetFileSize, canonical, and OpenFile? Cache hits also acquire this mutex, so a slow first open of a new shard blocks subsequent reads/writes to already-open shards in the same backend.

I reproduced this by pausing OpenFile for one shard: a cached read from another shard could not complete until the open was released. Please keep the cache lock limited to lookup/publication, use per-shard synchronization for initialization.

zupengwang: Fixed in [464892d](https://github.com/kvcache-ai/Mooncake/commit/464892dcddd1b375ebe69e0eee92ac5e77d2b17c). The global cache mutex now only protects lookup/insertion; file-size validation, canonical-path checks, and opening run under the individual shard's mutex. Initialization and positional I/O are serialized per shard, and failed opens remain retryable.

Added a deterministic regression that pauses OpenFile and GetFileSize in separate runs. It fails on a043ec4 and passes with the fix: cached reads/writes and another shard's first I/O complete while initialization is blocked. It also checks that concurrent first access to the same shard opens the file only once. A separate test covers retry after an open failure.

Remote Linux/POSIX validation: 54 tests passed across the DFS backend, TCP client, and checksum suites; three initialization/concurrency regressions passed 10 repetitions each. Applicable pre-commit checks passed.

mooncake-store/src/storage/distributed/distributed_storage_backend.cpp
```diff
@@ -57,7 +79,7 @@ DistributedStorageBackend::DistributedStorageBackend(
 }
 
 DistributedStorageBackend::~DistributedStorageBackend() {
-    for (auto& shard : shard_files_) {
+    for (auto& [index, shard] : shard_files_) {
         if (shard && shard->fd >= 0 && fs_adapter_) {
```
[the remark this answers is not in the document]
zupengwang: Fixed in [464892d](https://github.com/kvcache-ai/Mooncake/commit/464892dcddd1b375ebe69e0eee92ac5e77d2b17c): changed the unused destructor binding from `index` to `_`, following the existing convention. Applicable pre-commit checks passed.

fcczzz (approved):
LGTM

ykwd (approved):

